PR #6297, opened 2018-06-06
[18.05] add static content for vis plugins

martenson:
is this what is needed to make visualizations work on release branch? It is not even compressed so we shouldn't merge as is :/

ping @guerler

PR #5810, opened 2018-03-30
Plugin directory staging

guerler:

PR #6344, opened 2018-06-15
[18.05] Visualization static staging

dannon:
In #5810, we shifted to staging visualization plugins into the predefined static route.  This happened at client build time, for simplicity in implementation and use.  Unfortunately we don't *do* a client build yet as a part of the release-branch startup, so 18.05 isn't staging this stuff.

Instead of requiring the node environment, I put together a python script that does the same globbing as the gulp task, but uses symlinks instead of copying files, to keep it fast.  If people have concerns about this, I can easily swap it to a `copytree()`, but if this works well we may want to just drop the gulp plugin staging as well (or convert it to do the same thing).  This stages both repo-included plugins and those locally deployed, something #6297 did not do.

One note is that after this has run, the gulp plugin staging won't overwrite these symlinks once they're in place.  Seems to be 100% fine, since they do effectively the same way, but it should be cleaned up down the road.

### DIFF
--- a/scripts/plugin_staging.py
+++ b/scripts/plugin_staging.py
@@ -2,13 +2,13 @@ import glob
 import os
 import shutil
 
-GXY_ROOT = os.path.abspath(os.path.join(os.path.dirname(os.path.realpath(__file__)), '../'))
+GXY_ROOT = os.path.abspath(os.path.join(os.path.dirname(os.path.realpath(__file__)), os.pardir))
 
 
 def link_up_static(f):
     src = os.path.join(GXY_ROOT, 'config/plugins', f)
     dest = os.path.join(GXY_ROOT, 'static/plugins', f)
-    dest_parent = os.path.abspath(os.path.join(dest, "../"))
+    dest_parent = os.path.abspath(os.path.join(dest, os.pardir))
     if os.path.exists(dest):
         # We have to clear out the old staged or linked static to relink.
         if os.path.islink(dest):
--- a/scripts/plugin_staging.py
+++ b/scripts/plugin_staging.py
@@ -1,0 +1,30 @@
+import glob
+import os
+import shutil
+
+GXY_ROOT = os.path.abspath(os.path.join(os.path.dirname(os.path.realpath(__file__)), '../'))
+
+
+def link_up_static(f):
+    src = os.path.join(GXY_ROOT, 'config/plugins', f)
+    dest = os.path.join(GXY_ROOT, 'static/plugins', f)
+    dest_parent = os.path.abspath(os.path.join(dest, "../"))
+    # We have to clear out the old staged or linked static to relink.
+    if os.path.exists(dest):
+        if os.path.islink(dest):
+            os.unlink(dest)
+        else:
+            shutil.rmtree(dest)
+    elif not os.path.exists(dest_parent):
+        os.makedirs(dest_parent)
+    os.symlink(src, dest)
+
+
+if __name__ == "__main__":
+    # This is not awesome, but it's temporary, and it supports two-tier plugin static.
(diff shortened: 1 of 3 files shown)


scripts/plugin_staging.py
```diff
@@ -0,0 +1,28 @@
+import glob
+import os
+import shutil
+
+GXY_ROOT = os.path.abspath(os.path.join(os.path.dirname(os.path.realpath(__file__)), '../'))
```
nsoranzo: Replace `'../'` with `os.pardir`

dannon:
WIP for a sec to fix a uwsgi-related static-security-detection issue (uwsgi notices the symlink and won't serve it, since it's outside the static-map).  (thanks @guerler)

scripts/plugin_staging.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def link_up_static(f):
+    src = os.path.join(GXY_ROOT, 'config/plugins', f)
+    dest = os.path.join(GXY_ROOT, 'static/plugins', f)
+    dest_parent = os.path.abspath(os.path.join(dest, "../"))
```
nsoranzo: Replace `'../'` with `os.pardir`

dannon:
Ok, added a uwsgi-specific fix in 8364ebe, to allow uwsgi to navigate/serve the symlinked static paths (paste didn't care).  We can get rid of this if we decide the best path down the road is to continue to use gulp to pre-stage once per build, without symlinks.

Another option (for this PR) would be to swap back from symlinks to full static dir copies (`shutil.copytree` instead of the link) the same as gulp does it.  It's not much to copy, but of course it's slightly slower and may be a concern, with it happening every time run.sh is executed.

guerler:
👍

nsoranzo (commented):
What about creating symlinks only for `config/plugins/visualizations/` and `config/plugins/interactive_environments/` ? Worked for me in my limited testing and should save some unnecessary unlink/relink at each server restart.

dannon:
@nsoranzo Regarding the two toplevel symlinks vs the specific ones -- we don't want to serve the python files and other contents -- we only want to serve the plugin-specific static contents directories.

scripts/plugin_staging.py
dannon: Thanks, done!

scripts/plugin_staging.py
dannon: Done.

nsoranzo:
> Regarding the two toplevel symlinks vs the specific ones -- we don't want to serve the python files and other contents -- we only want to serve the plugin-specific static contents directories.

Ah, that's right, sorry! For some reason I thought the code were symlinking also the files inside the `static` subdirs, but it's symlinking only the dirs, so all good!

nsoranzo (approved):

martenson:
@dannon please re-target the branch

dannon:
@martenson Sure, thanks for checking it out.

martenson (approved):

martenson:
@dannon Thank you for working on this.

I am not sure how the symlinks will play together with upgrading to a new release where we want to actually copy the files to there but it sounds like a solvable problem( for future us).

nsoranzo:
@martenson Or we can keep symlinking also in the next releases?

Connected to that, I was thinking we could wipe `static/plugins/` at the start of the script to prevent stale subdirs for removed plugins.

dannon:
@nsoranzo We can keep symlinking, though the bypass I had to add with uwsgi doesn't feel super clean to me.  I'd rather just copy-on-client-build in the future, I think (exactly as I implemented originally via `gulp plugins`), for an isolated non-linked static serving location.  Ideally this doesn't run every time you run.sh, and is attached to a client build -- that it's a part of run.sh is mostly just a fix to work with the no-deployment-builds-required constraints we have right now.

PR #6342, opened 2018-06-15
[18.01][WIP] Fix to speedup, reduce memory of large number of JobToInput associations.

jmchilton:
Should speed up creation large collection jobs and reduce memory footprint.

PR #5538, opened 2018-02-14
Rev SQLAlchemy from 1.0.15 to 1.2.2.

jmchilton:
It'd be nice to use [selectinload](http://docs.sqlalchemy.org/en/latest/orm/loading_relationships.html#sqlalchemy.orm.selectinload) when optimizing postgres calls.

xref https://github.com/galaxyproject/starforge/pull/191

PR #6348, opened 2018-06-15
[18.05] Fix to speedup, reduce memory of large number of JobToInput associations.

jmchilton:
Should speed up creation large collection jobs and reduce memory footprint.

Seems to be incompatible in a fails transiently sort of way with 18.01 - see #6342.

### DIFF
--- a/lib/galaxy/tools/actions/__init__.py
+++ b/lib/galaxy/tools/actions/__init__.py
@@ -491,7 +491,7 @@ class DefaultToolAction(object):
         # results back and update the association objects.
         bulk_job_input_assocs = None
         handler = None
-        if len(job.input_datasets) > 0:
+        if len(job.input_datasets) > 10:
             bulk_job_input_assocs = job.input_datasets
             job.input_datasets = []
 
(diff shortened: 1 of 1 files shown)


lib/galaxy/tools/actions/__init__.py
```diff
@@ -495,10 +509,25 @@ def handle_output(name, output, hidden=None):
...
+                sa_session.bulk_save_objects(bulk_job_input_assocs)
+                t.commit()
+            except Exception:
+                t.rollback()
+                raise
+
+            job.handler = handler
+            trans.sa_session.flush()
+
+        log.info("Flushed transaction for job %s %s %s" % (job.log_str(), job_flush_timer, job.handler))
```
jmchilton: Some debugging here to remove. The 0 cutoff above as well.

nsoranzo:
Thanks a lot for working on this! I can test it if it's ready.

jmchilton:
Do you have an 18.05 test case to throw at this? I wasn't able to do this with 18.01? You could apply this along with https://github.com/galaxyproject/galaxy/pull/5538/files to 18.01 though I suppose.

jmchilton:
We should probably be exposing the use_batch_mode mode parameter in our sqlalchemy engine creation - http://docs.sqlalchemy.org/en/latest/dialects/postgresql.html#psycopg2-batch-mode-fast-execution.

jmchilton:
Closed because in actual testing in running Galaxy instances I was not able to observe speed ups due to this with postgres (the general approach does speed up sqlite in testing - but this isn't enough to make this change I think).

Issue #1765, opened 2016-02-19
objectstore: _call_method failed: get_filename


bgruening:
In rare cases I get this in my log. I can not really reproduce it but sometimes it happes during copying histories.

```
galaxy.objectstore WARNING 2016-02-19 09:03:43,522 The backend object store ID (None) for Dataset object with ID 92807 is invalid
galaxy.datatypes.interval ERROR 2016-02-19 09:03:43,523 Exception caught attempting to generate viewport for dataset '104946'
Traceback (most recent call last):
  File "/usr/local/galaxy/galaxy-dist/lib/galaxy/datatypes/interval.py", line 169, in get_estimated_display_viewport
    fh = open( dataset.file_name )
  File "/usr/local/galaxy/galaxy-dist/lib/galaxy/model/__init__.py", line 1853, in get_file_name
    return self.dataset.get_file_name()
  File "/usr/local/galaxy/galaxy-dist/lib/galaxy/model/__init__.py", line 1658, in get_file_name
    filename = self.object_store.get_filename( self )
  File "/usr/local/galaxy/galaxy-dist/lib/galaxy/objectstore/__init__.py", line 477, in get_filename
    return self._call_method('get_filename', obj, ObjectNotFound, True, **kwargs)
  File "/usr/local/galaxy/galaxy-dist/lib/galaxy/objectstore/__init__.py", line 498, in _call_method
    % ( method, str( obj ), str( kwargs ) ) )
ObjectNotFound: objectstore, _call_method failed: get_filename on <galaxy.model.Dataset object at 0x20995250>, kwargs: {}
galaxy.objectstore WARNING 2016-02-19 09:03:44,272 The backend object store ID (None) for Dataset object with ID 92811 is invalid
galaxy.objectstore WARNING 2016-02-19 09:03:44,620 The backend object store ID (None) for Dataset object with ID 92812 is invalid
galaxy.objectstore WARNING 2016-02-19 09:03:45,478 The backend object store ID (None) for Dataset object with ID 92816 is invalid
galaxy.objectstore WARNING 2016-02-19 09:03:45,953 The backend object store ID (None) for Dataset object with ID 92817 is invalid
galaxy.objectstore WARNING 2016-02-19 09:03:46,277 The backend object store ID (None) for Dataset object with ID 92818 is invalid
galaxy.objectstore WARNING 2016-02-19 09:03:46,581 The backend object store ID (None) for Dataset object with ID 92819 is invalid
[...]

PR #6337, opened 2018-06-14
Return the return code produced by the tool script so the resource ma…

dpryan79:
…nager can see it.

This allows resource managers (e.g., slurm) to see the exit code produced by jobs and set the job state accordingly. This is nice in cases where Galaxy needs to run `scontrol` to check on a job state. Currently, it sees that failed jobs finished successfully...which means that it then looks for their non-existent output files. Half the time this then freezes the job handlers, which produce a bunch of errors like:

    ObjectNotFound: objectstore, _call_method failed: get_filename on <galaxy.model.Dataset object at 0x7f1e436eead0>, kwargs: {}

I mentioned this on gitter and I suspect that this change will prevent the above from happening. I'm testing this at the moment and will report back.

PR #6305, opened 2018-06-08
Fix default job script by moving env setup

rhpvorderman:
The part where pythonpath was evaluated totally ignored any variables set in the env setup. That is now fixed. All credits go to @mvdbeek for finding the cause of the issue and proposing the solution.

PR #6349, opened 2018-06-15
Prevent throwing an exception if a dataset doesn't exist (and never w…

dpryan79:
…ill) due to a job error. This is an alternative (and better) solution to #6337 when jobs set metadata externally.

In short, if a job throws an error and thus doesn't create its output job, then any call to `dataset.dataset.file_name` will result in an error, at least if you use a nested data object store. You can see the error message caused by this in #6337 near the bottom of the comments. It's problematic that `log.warning()` line that I'm changing in this PR uses that function call, since it (A) prevents checking if the file is only absent due to NFS lag and (B) if the job truly did finish in an error then `Unable to finish job` is printed in the error field in the Galaxy history item rather than the actual error message printed to stderr.

It'd be nice if this and #6305 could get backported to 18.05 :)

### DIFF
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1258,7 +1258,7 @@ class JobWrapper(HasResourceParameters):
                             trynum = self.app.config.retry_job_output_collection
                         except (OSError, ObjectNotFound) as e:
                             trynum += 1
-                            log.warning('Error accessing %s, will retry: %s', dataset.dataset.file_name, e)
+                            log.warning('Error accessing dataset with ID %i, will retry: %s', dataset.dataset.id, e)
                             time.sleep(2)
                 if getattr(dataset, "hidden_beneath_collection_instance", None):
                     dataset.visible = False


mvdbeek:
That looks pretty uncontroversial and probably also fixes what is left of https://github.com/galaxyproject/galaxy/issues/1765

mvdbeek (approved):

dpryan79:
I don't think this will fix #1765, but I think that has the same cause (namely you get an error if a job has failed and files don't exist when you use an object store).

dpryan79:
I take that back, it'll at least fix the latter error that @bgruening reported in #1765.

mvdbeek:
@galaxybot test this

mvdbeek:
The other question is whether the nested object store should be raising an exception if the file does not exist. The other object stores don't, and having this divergent behaviour makes things more difficult IMO

nsoranzo:
@mvdbeek It seems `get_filename()`also in other object stores (e.g. s3, rods) can raise `ObjectNotFound`.

PR #6360, opened 2018-06-19
[18.05] Fix size handling in legacy modal view

### LINKED ISSUE #6352: When there are too many issues in a workflow you cannot click continue
auswahl_039

- I could zoom out but this is sub-optimal.
- Yes I know this should never happen.

guerler:
Fixes #6352

### DIFF
--- a/client/galaxy/scripts/layout/modal.js
+++ b/client/galaxy/scripts/layout/modal.js
@@ -80,14 +80,7 @@ $.extend(Modal.prototype, {
         this.$body.css("min-width", this.$body.width());
         // Set max-height so that modal does not exceed window size and is in middle of page.
         // TODO: this could perhaps be handled better using CSS.
-        this.$body.css(
-            "max-height",
-            $(window).height() -
-                this.$footer.outerHeight() -
-                this.$header.outerHeight() -
-                parseInt(this.$dialog.css("padding-top"), 10) -
-                parseInt(this.$dialog.css("padding-bottom"), 10)
-        );
+        this.$body.css("max-height", $(window).height() / 1.5);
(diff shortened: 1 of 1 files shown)


martenson:
I think changing the modal default height is potentially problematic since various places display it for various purposes and this would affect all of them.

dannon:
Agreed dropping it to 50% window height is probably more than some modal-using interfaces are going to handle correctly right now.  Fixing it correctly to use the calculations as intended, though (as in the first changeset, if I read it correctly), will fix existing modals that run off the screen and we should do that.

guerler:
Its not changing the default height. This PR exposes the existing maximum height handling which limits the modal size by the window height. I set the max to 67% of the window height because its more robust and the existing calculation is flawed i.e. the modal footer overlaps with the window bottom. This legacy modal is only used in pages, the workflow editor and sweepster. I checked the use cases and the modals look fine.

dannon:
@guerler Maybe I missed a rebase?  I could swear it was cutting it down more than that.  In any event, if you've tested all the locations that's okay with me.

guerler:
Oh I upped it to 67% after you raised concerns regarding 50%. (: Thanks for the review.

dannon:
@guerler Ok, whew, I'm not going crazy.  Yet.  Sure, anytime man!

jmchilton (approved):
Tested out various collection modals and they looked fine.

dannon:
Sorry, I forgot to explicitly +1 yesterday.  This looks good to me.

dannon:
I'm going to merge #6364, and will rebuild the client during that, which will incorporate these changes.

PR #6361, opened 2018-06-19
[18.05] Add fasta datatype to msa viewer

### LINKED ISSUE #6359: Please add support for the fasta datatype to charts
@guerler 

We have made extensive use of the Charts viz for the fasta datatype with the PlantTribes tools, and we are hoping that you will agree to support fasta in Charts on Galaxy test and main.  Here is a snapshot of an example of what we do using the Charts viz with some of the PlantTribes tools.  Again, we do a lot more than just this.

I have the Charts viz configured in a custom way using an older version of Galaxy where I compiled the Javascript for Charts after adding the fasta datatype along with the txt datatype (in the older version, Charts was configured only for txt).  This custom setup is running on a Galaxy environment that I’m maintaining for a local lab.

With the latest version of Galaxy, Charts is integrated into the code, so I’m no longer sure how to add support for the fasta datatype.  

Can you please let us know if you think we can add support for fasta to charts?

image

@martenson @davebx after sending the email regarding this, I figured I should have just opened an issue.  The old days are still too fresh in my mind - sorry!  ;)

guerler:
Fixes #6359. ping @gregvonkuster.

### DIFF
--- a/config/plugins/visualizations/msa/config/msa.xml
+++ b/config/plugins/visualizations/msa/config/msa.xml
@@ -6,6 +6,7 @@
         <data_source>
             <model_class>HistoryDatasetAssociation</model_class>
             <test test_attr="ext" result_type="datatype">txt</test>
+            <test test_attr="ext" result_type="datatype">fasta</test>
             <to_param param_attr="id">dataset_id</to_param>
         </data_source>
     </data_sources>
--- a/config/plugins/visualizations/msa/config/msa.xml
+++ b/config/plugins/visualizations/msa/config/msa.xml
@@ -7,6 +7,7 @@
             <model_class>HistoryDatasetAssociation</model_class>
             <test test_attr="ext" result_type="datatype">txt</test>
             <test test_attr="ext" result_type="datatype">fasta</test>
+            <test test_attr="ext" result_type="datatype">phylip</test>
             <to_param param_attr="id">dataset_id</to_param>
         </data_source>
     </data_sources>


davebx:
@guerler I have @gregvonkuster requesting the phylip format added as well

dannon:
@guerler Descendants of classes don't count for the matching?  I thought Fasta would work as-is, with just the Text entry there -- "Text" implying text and descendant classes.

davebx:
@dannon I recently tested and found that I had to change the datatype to `txt` before the MSA viewer would recognize it.

dannon:
@guerler @davebx Yeah, looks like we should just swap this to:
```
<test type="isinstance" test_attr="datatype" result_type="datatype">sequence.Sequence</test>
```
?

(edit: updated it to 'sequence')

davebx:
@dannon that _sounds_ reasonable, but I'm not familiar enough with the internals of visualizations to speak _ex cathedra_, as it were

martenson:
> "Text" implying text and descendant classes

I understood it the same. 

But I would also like to propose that if this is (or would be) the case almost nothing should have `txt` since we want to offer users reasonable visualizations for their data formats, not overwhelm them with everything we have.

guerler:
Looks like `sequence.Sequence` does not work because its not listed in the datatypes_conf.xml i.e. has not extension by itself.

dannon:
To xpost from IRC, it looks like there's an additional bug here preventing `sequence.Sequence` inheritance working correctly, though we *should* be able to specify it as a type here.

guerler:
Should we merge this as it is to get it working for fasta files and then revisit this issue separately.

martenson (approved):

dannon:
@guerler Sure, that sounds fine to me.  This is certainly an improvement, in any event.

gregvonkuster:
Thanks very much for this everyone!  I'm glad to now get confirmation on the subclass intent here as that is what @guerler and I discussed way back when I was initially configuring fasta support in a custom way.  I really appreciate this temporary fix for the 2 datatypes we needed for an upcoming workshop in Kenya.  Much appreciated!  ;)

PR #6363, opened 2018-06-20
changed default location of integrated_tool_panel.xml

bernt-matthias:
I guess it should be in config/ instead of the Galaxy root

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,7 +61,7 @@ migrated_tools_conf.xml
 shed_tool_conf.xml
 tool_data_table_conf.xml
 tool_sheds_conf.xml
-config/integrated_tool_panel.xml
+integrated_tool_panel.xml
 openid_conf.xml
 shed_tool_data_table_conf.xml
 job_conf.xml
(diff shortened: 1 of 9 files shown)


jmchilton:
Unit test failures appear to be legitimate, marking as WIP for that reason. @natefoo / @mvdbeek  are you okay with moving this? I mean we put mutable stuff in both Galaxy root and config currently - I'm not sure I find a compelling reason to prefer one over the other.

If there is a consensus around making this change though I can fix the unit tests and get this merged.

mvdbeek:
Sure, seems fine to me. Should we recover `integrated_tool_panel.xml` from the root dir if it exists there but not in `config/` ?

natefoo:
I wasn't even aware that it wasn't in `config/`. I think somewhere in `database/` (say `database/config/`?) actually makes more sense (I call this the "mutable" config/data dir in the Ansible stuff), because ideally `config/` is only for stuff an admin writes. The other Galaxy-written configs should also go wherever `integrated_tool_panel.xml` goes, but those probably aren't something we should change except for new installs.

nsoranzo:
This was moved to `managed_config_dir` long ago.

PR #6365, opened 2018-06-20
Add period to the list of allowed characters in user names

moskalenko:
A lot of email addresses contain periods. The usernames created from them currently have periods replaced by dashes, which creates problems for real user jobs. Let's add period to the list of characters allowed in the usernames.


jmchilton (approved):

jmchilton:
@galaxybot test this

PR #6367, opened 2018-06-20
Optimizations for dynamic output file collection.

jmchilton:
During the set hid and update quota parts of this operation - skip a bunch of extra flushes and skip checking for extra files since these cannot be created in conjunction with dynamic file discovery currently.

Cut this part of that operation from about 50 seconds to 7 when creating a collection with 1000 elements on my laptop against a local postgres database.

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -1403,7 +1403,6 @@ class History(HasTags, Dictifiable, UsesAnnotations, HasName):
             if quota and self.user:
                 disk_usage = sum([d.get_total_size() for d in datasets])
                 self.user.adjust_total_disk_usage(disk_usage)
-
             sa_session.add_all(datasets)
             if flush:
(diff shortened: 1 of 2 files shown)


lib/galaxy/model/__init__.py
```diff
@@ -1981,10 +1980,17 @@ def get_size(self, nice_size=False):
...
 
-    def set_size(self):
-        """Sets the size of the data on disk"""
+    def set_size(self, no_extra_files=False):
+        """Sets the size of the data on disk.
+
+        If the caller is sure there are no extra files, pass no_extra_files as True to optimize subsequent
+        calls to get_total_size or set_total_size - potentially avoiding both a database flush and check against
+        the file system.
+f        """
```
mvdbeek: The starting `f` is a typo, right ?

jmchilton: Yes... thanks for catching that.

PR #6368, opened 2018-06-20
[WIP] Optimize permission loading and checking collection parameters.

jmchilton:
We were checking permissions twice it seems like to me (once when collecting the inputs into a big odict and once right before setting up the jobtoinput associations), so now only do it once. Also - avoid loading dataset permissions for each individual dataset in a collection - work with summary sets of role ids.

I still need to collect data but this really better speed up large collection operation job submissions, I'll try to get some data.


jmchilton:
I can't make this speed up anything quite yet, closing it out. It is still in the mega branch with some other changes as well.

PR #6369, opened 2018-06-20
[18.05] Make sure that the workflow step has a proper state available

### LINKED ISSUE #6353: workflow in a Galaxy Page stops rendering with an error
reported here: https://biostar.usegalaxy.org/p/28436/

looks like this: 
screenshot 2018-06-18 10 47 02

the workflow: https://usegalaxy.org/u/mwolfien/p/trapline---manual
It has more than 100 steps but the rest is not rendered

error when editing after import: 
screenshot 2018-06-18 10 45 52

guerler:
Fixes #6353. This PR fixes a regression, by allowing the displaying controller endpoints to produce complete workflow states despite missing tools or step state errors. The shared controller mixin is used in pages and workflows.

### DIFF
--- a/lib/galaxy/web/base/controller.py
+++ b/lib/galaxy/web/base/controller.py
@@ -1211,7 +1211,7 @@ class UsesStoredWorkflowMixin(SharableItemSecurityMixin, UsesAnnotations):
         module_injector = WorkflowModuleInjector(trans)
         for step in stored_workflow.latest_workflow.steps:
             try:
-                module_injector.inject(step)
+                module_injector.inject(step, exact_tools=False)
             except exceptions.ToolMissingException:
                 pass
 


templates/webapps/galaxy/workflow/display.mako
```diff
@@ -104,7 +104,9 @@
                 %if tool:
                   <div class="toolFormTitle">Step ${int(step.order_index)+1}: ${tool.name | h}</div>
                   <div class="toolFormBody">
-                    ${do_inputs( tool.inputs, step.state.inputs, "", step )}
+                    %if hasattr(step, "state"):
```
mvdbeek: Can you add a comment on when a step has no state attribute?
Is that when a workflow step (or is it a workflow module ?) references a tool parameter that doesn't exist anymore ?
If that's the case, should we render a message that explains this, e.g "This step references an unknown parameter. Please update this step." ?

guerler:
Thanks for the review @mvdbeek. While looking more into it I found a more consistent solution for this issue.

martenson (approved):

Issue #6350, opened 2018-06-15
Duplicate history entries


guerler:
Possibly related to #3818. I noticed multiple times recently duplicate entries in the history. Created this to keep track of it.

![duplicates](https://user-images.githubusercontent.com/2105447/41490826-d1e4e4b4-70c3-11e8-8900-1d5c3f34fa7e.png)

mvdbeek:
Have you been using the collection operation tools on a collection containing those datasets ? I have seen this happening in these circumstances, I belive this happens when we add new HDAs to a history [here](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/__init__.py#L2630)

guerler:
Interesting. Thanks @mvdbeek, and yes, it occurs in relation to collection operations.

jmchilton:
I wasn't even trying to fix this but I think https://github.com/galaxyproject/galaxy/pull/6370/files will fix it.

mvdbeek:
Yeah, we solved this a while back I think.

PR #6370, opened 2018-06-21
More correct, more efficient collection operations.

jmchilton:
We were copying datasets and not assigning them new HIDs in some cases (the newer collection operations apply_rules, tag_from_file were doing the right thing here but it looks like we didn't go back and fix older operations).

With the uniform handling of dataset copying in place - I used the much more efficient history.add_datasets instead of history.add_dataset.

This also optimizes the actual HDA copying by eliminating some extra flushes. hda.copy() would flush to get an HDA before copying metadata but this is only needed if there are FileParameter metadata types (e.g. bams) so I made that flush conditional as well as the final one. So for most datatypes - this eliminates two extra flushes per dataset copied.

### DIFF
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -2523,7 +2523,6 @@ class FilterDatasetsTool(DatabaseOperationTool):
         new_elements = odict()
         for dce in elements_to_copy:
             element_identifier = dce.element_identifier
-            copied_value = dce.element_object.copy()
             if getattr(dce.element_object, "history_content_type", None) == "dataset":
                 copied_value = dce.element_object.copy(force_flush=False)
                 copied_value.visible = False
@@ -2811,7 +2810,6 @@ class FilterFromFileTool(DatabaseOperationTool):
             in_filter_file = element_identifier in filtered_identifiers
             passes_filter = in_filter_file if how_filter == "remove_if_absent" else not in_filter_file
 
-            copied_value = dce_object.copy()
             if getattr(copied_value, "history_content_type", None) == "dataset":
                 copied_value = dce_object.copy(force_flush=False)
                 copied_value.visible = False
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -16,7 +16,7 @@ from xml.etree import ElementTree
 import packaging.version
 from mako.template import Template
 from paste import httpexceptions
-from six import string_types
+from six import itervalues, string_types
 from six.moves.urllib.parse import unquote_plus
 
 import tool_shed.util.repository_util as repository_util
@@ -2378,6 +2378,15 @@ class DatabaseOperationTool(Tool):
 
             map(check_dataset_instance, input_dataset_collection.dataset_instances)
 
+    def _add_datasets_to_history(self, history, elements):
+        datasets = []
+        for element_object in elements:
+            if getattr(element_object, "history_content_type", None) == "dataset":
+                datasets.append(element_object)
+
+        if datasets:
+            history.add_datasets(self.sa_session, datasets, set_hid=True)
+
     def produce_outputs(self, trans, out_data, output_collections, incoming, history):
         return self._outputs_dict()
 
@@ -2400,10 +2409,7 @@ class UnzipCollectionTool(DatabaseOperationTool):
         assert collection.collection_type == "paired"
         forward_o, reverse_o = collection.dataset_instances
         forward, reverse = forward_o.copy(copy_tags=tags), reverse_o.copy(copy_tags=tags)
-
-        # TODO: rename...
-        history.add_dataset(forward, set_hid=True)
-        history.add_dataset(reverse, set_hid=True)
+        self._add_datasets_to_history(history, [forward, reverse])
 
         out_data["forward"] = forward
         out_data["reverse"] = reverse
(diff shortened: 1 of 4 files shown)


mvdbeek:
The API test is failing because it relies on the number of datasets in the history. https://github.com/jmchilton/galaxy/pull/69 should fix that

mvdbeek (approved):

lib/galaxy/tools/__init__.py
```diff
@@ -2378,6 +2378,20 @@ def check_dataset_instance(input_dataset):
...
             map(check_dataset_instance, input_dataset_collection.dataset_instances)
 
+    def _add_datasets_to_history(self, history, elements_dict_or_list):
+        datasets = []
+        if isinstance(elements_dict_or_list, list):
+            for element_object in elements_dict_or_list:
+                if getattr(element_object, "history_content_type", None) == "dataset":
+                    datasets.append(element_object)
+        else:
+            for element_object in elements_dict_or_list.values():
```
nsoranzo: Wouldn't it be cleaner if this method would accept only lists, i.e. we pass `.values()` of the dict instead of the dict?
Or start this method with `if not isinstance(elements_dict_or_list, list): elements_dict_or_list = elements_dict_or_list.values()`

jmchilton: Yes, good point - I'll fix this up.

lib/galaxy/tools/__init__.py
```diff
@@ -2791,13 +2814,18 @@ def produce_outputs(self, trans, out_data, output_collections, incoming, history
 
             copied_value = dce_object.copy()
```
nsoranzo: This line is not needed any more, I think.

lib/galaxy/tools/__init__.py
```diff
@@ -2378,6 +2378,15 @@ def check_dataset_instance(input_dataset):
...
             map(check_dataset_instance, input_dataset_collection.dataset_instances)
 
+    def _add_datasets_to_history(self, history, elements):
+        datasets = []
+        for element_object in elements:
+            if getattr(element_object, "history_content_type", None) == "dataset":
+                datasets.append(element_object)
+
+        if datasets:
+            history.add_datasets(self.sa_session, datasets, set_hid=True)
```
nsoranzo, line 2388: Should `set_hid` be a parameter for this method? Here is always `True`, but `False` was passed before in some cases (`ZipCollectionTool`, `MergeCollectionTool`, `FilterDatasetsTool`, `FlattenTool`, `SortTool`, `RelabelFromFileTool`, `TagFromFileTool`, `FilterFromFileTool`).
Not sure if that's connected with the failed API test: https://jenkins.galaxyproject.org/job/docker-api/12151/testReport/junit/api.test_tools/ToolsTestCase/test_map_over_collection_structured_like/

jmchilton: It being False before was a bug I would say - we need to either copy the HDAs and assign them new HIDs or use the original HDAs. There have been some reports of duplicate HIDs #6350(?)  - and collection operations have probably been the problem.

I just clicked through on that issue and realized @mvdbeek knew this at least several days before me 😆.

lib/galaxy/tools/__init__.py
jmchilton: Indeed!

lib/galaxy/tools/__init__.py
jmchilton: Oh this test failed because I rebased your changes and lost @mvdbeek's fix 😢.

lib/galaxy/tools/__init__.py
nsoranzo: Still present.

lib/galaxy/tools/__init__.py
nsoranzo: > It being False before was a bug I would say - we need to either copy the HDAs and assign them new HIDs or use the original HDAs.

It seems to me it should be possible to use the original HDAs for `ZipCollectionTool`, `MergeCollectionTool`, `FilterDatasetsTool`, `FlattenTool`, `SortTool`, `RelabelFromFileTool` and `FilterFromFileTool` classes.

jmchilton: If you delete the new collection + contents - it deletes the old HDAs. You have a green collection with deleted HDAs and the user has no clue why. This was part of the original collection design that I really regret - the behavior makes sense from an API design perspective but it doesn't lead to a good UX. Collection contents shouldn't be HDAs - but if they are HDAs they should at least be in exactly one collection to make the UI work correctly. We've been on a path toward making collections more cohesive and I think it has been good.

We do not duplicate the datasets, just the HDAs - so we aren't duplicating space on disk for this or anything (expect BAI files - but the fix for that isn't reusing the HDAs I don't think). Just trying to improve the user experience.

If you want to open an issue to add an option to not duplicate the HDAs or something that is fine - we could discuss that more I think but certainly copying them and not giving them a new HID is a problem right?

nsoranzo: That makes sense, I've just tried to explore the "or use the original HDAs." part of your previous reply.
If you remove the line mentioned in https://github.com/galaxyproject/galaxy/pull/6370/files#r197497120 , I'll merge this.

lib/galaxy/tools/__init__.py
jmchilton: There was another one in this rebase at 2526 - must have been a repeated copy and paste mistake and just removed one instance. I'll rebase again and try to get them all.

lib/galaxy/tools/__init__.py
```diff
@@ -2789,15 +2810,19 @@ def produce_outputs(self, trans, out_data, output_collections, incoming, history
             in_filter_file = element_identifier in filtered_identifiers
             passes_filter = in_filter_file if how_filter == "remove_if_absent" else not in_filter_file
 
-            copied_value = dce_object.copy()
             if getattr(copied_value, "history_content_type", None) == "dataset":
```
nsoranzo: s/copied_value/dce_object/

nsoranzo (approved):

PR #6382, opened 2018-06-22
[WIP] Modify grt export

atyryshkina:
Changed to use the standard python library csvwriter to write to tsv. https://docs.python.org/2/library/csv.html This may be more robust to unusual input. The csv writer automatically quotes those fields which contain special characters such as delimiters, quote characters, and line terminators.

### DIFF
--- a/scripts/grt/export.py
+++ b/scripts/grt/export.py
@@ -11,6 +11,7 @@ import sys
 import tarfile
 import time
 from collections import defaultdict
+import csv
 
 import yaml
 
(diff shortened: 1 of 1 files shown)


scripts/grt/export.py
```diff
@@ -419,6 +439,7 @@ def annotate(label, human_label=None):
     with open(CHECK_POINT_FILE, 'w') as handle:
         handle.write(str(end_job_id))
 
+    print('done')
```
hexylena, line 442: Should use logging here

scripts/grt/export.py
```diff
@@ -289,43 +292,46 @@ def annotate(label, human_label=None):
...
-            handle_datasets.write(hdas[hda_id][1])
-            handle_datasets.write('\t')
-            handle_datasets.write(str(datasets[dataset_id][0]))
-            handle_datasets.write('\t')
-            handle_datasets.write(job[2])
-            handle_datasets.write('\t')
-            handle_datasets.write(filetype)
-            handle_datasets.write('\n')
+            result = []
+            for i in range(len(record)):
```
hexylena, line 319: maybe we just make a function for doing this?

PR #6383, opened 2018-06-22
Rewrite admin data types as vue component

scrathat:
Take the DataTables component as an example and do the same for data types.

- Add DataTypes vue component which reuses BaseGrid and Message
- Users have the option to show predefined or all columns supplied by the api
- Add data_types_list web api which returns all datatypes and their attributes
- Add client route for admin/data_types and reflect the change in admin router


bgruening:
Cool thanks @scrathat! Keep them coming.

dannon:
@scrathat Thanks for this!  I pushed one more commit removing the old template and endpoint. 

There's a minor display bug when toggling the full column display, but we can revisit that when we overhaul GridBase.vue, now that we're going to be using bootstrap-vue everywhere.

PR #6394, opened 2018-06-25
Optimize setting large numbers of HIDs for collection outputs.

jmchilton:
For tools that output explicit collections of known type (e.g. lists with ``structured_like``), this produces  output lists with 1000 elements a bit faster with this patch than lists with 500 elements without it in my testing. This eliminates many flushes when creating these datasets and sets HIDs all together in one transaction instead of separate HID generation transaction per dataset.

### DIFF
--- a/lib/galaxy/tools/actions/__init__.py
+++ b/lib/galaxy/tools/actions/__init__.py
@@ -402,6 +402,7 @@ class DefaultToolAction(object):
                     collections_manager = app.dataset_collections_service
                     element_identifiers = []
                     known_outputs = output.known_outputs(input_collections, collections_manager.type_registry)
+                    created_element_datasets = []
                     # Just to echo TODO elsewhere - this should be restructured to allow
                     # nested collections.
                     for output_part_def in known_outputs:
@@ -428,21 +429,19 @@ class DefaultToolAction(object):
 
                         effective_output_name = output_part_def.effective_output_name
                         element = handle_output(effective_output_name, output_part_def.output_def, hidden=True)
+                        created_element_datasets.append(element)
                         # TODO: this shouldn't exist in the top-level of the history at all
                         # but for now we are still working around that by hiding the contents
                         # there.
                         # Following hack causes dataset to no be added to history...
                         child_dataset_names.add(effective_output_name)
-
-                        history.add_dataset(element, set_hid=set_output_hid, quota=False)
                         trans.sa_session.add(element)
-                        trans.sa_session.flush()
-
                         current_element_identifiers.append({
                             "__object__": element,
                             "name": output_part_def.element_identifier,
                         })
 
+                    history.add_datasets(trans.sa_session, created_element_datasets, set_hid=set_output_hid, quota=False, flush=True)
                     if output.dynamic_structure:
                         assert not element_identifiers  # known_outputs must have been empty
                         element_kwds = dict(elements=collections_manager.ELEMENTS_UNINITIALIZED)


lib/galaxy/tools/actions/__init__.py
```diff
@@ -428,21 +429,19 @@ def handle_output(name, output, hidden=None):
...
-                        history.add_dataset(element, set_hid=set_output_hid, quota=False)
                         trans.sa_session.add(element)
-                        trans.sa_session.flush()
-
                         current_element_identifiers.append({
                             "__object__": element,
                             "name": output_part_def.element_identifier,
                         })
 
+                    history.add_datasets(trans.sa_session, created_element_datasets, set_hid=True, quota=False, flush=True)
```
nsoranzo: Just to double check: you changed `set_hid` to be always `True` for the same reason explained in https://github.com/galaxyproject/galaxy/pull/6370/files#r197464679 , right?

jmchilton: Good question - this time it wasn't intentional. I'll have to look into this more and see when set_hid would be false - hopefully it would never take this path in that case but I'll verify.

Update: Note to self added in https://github.com/galaxyproject/galaxy/commit/a7ea9357f20fe19dbb6a8f8f216570b41cf281cc.

jmchilton: So I think this is only used for implicit converters which don't define explicit output collections so set_hid would never be false here. I guess I'll throw in an assert and a comment unless you have different suggestion. Good catch though - thanks!

nsoranzo: > So I think this is only used for implicit converters which don't define explicit output collections so set_hid would never be false here.

Because we haven't implemented implicit converters for collections yet?

Changing it back to `set_output_hid` feels safer to me.

jmchilton: > Because we haven't implemented implicit converters for collections yet?

No because implicit converters produce datasets - even when they work over collections - they are many individual jobs each producing a dataset. Converters can't produce collections - but I will change it back.

nsoranzo:
Thanks @jmchilton! It's clear that this will make collection output creation faster, but I was wondering if it could make the memory problems worse?

jmchilton:
I don't know if this little slice of things is a memory problem at all - I doubt it is and if it is I think this would shift the problematic memory consumption into a smaller window of time? So I'd guess on balance if anything it improves things a bit memory-wise but I can't really speculate in an educated way in either direction.

What I will say is that it further consolidates things into ``__add_datasets_optimized`` (in ``models/__init__.py``) that is becoming a quite critical path that we could optimize further to reduce memory and runtime of by just rewriting it to use dataset IDs and building low level insert, update statements with SA core. So... if the problem is huge flushes with a bunch of individual SA inserts/updates consuming a bunch of memory for some reasons... this puts us in a better position to address it with lower level database interactions that don't require SA book keeping.

If you're unconvinced though we can keep everything in #6380 until I get a handle on the memory issue and I can profile it all together.

nsoranzo:
Basically my doubt is that we are creating a list `created_element_datasets` which could contain several thousand of `HistoryDatasetAssociation` objects.

jmchilton:
The objects are in memory already - so it is just the list backbone that is being stored in memory for a very brief amount of time vs the unpatched version. I expect that the list backbone memory consumption per element is much smaller than the full object it holds a reference to, its entry in current_element_identifiers, the SA stuff that gets attached to it, etc..

The list backbone itself should probably be the same size in memory as a list of integers? 

```python
>>> sys.getsizeof(list(xrange(100000)))
900120
```

So for 100,000 elements that is an extra 900K.

nsoranzo:
I was hoping these objects are garbage collected when a new value is assigned to `element`, but maybe there are other references to them.

jmchilton:
@nsoranzo there are other references in a dict that can't be garbage collected until after ``created_element_datasets`` can be - see https://github.com/galaxyproject/galaxy/pull/6394/files#diff-bfb557a99c1f7d646d4968d8d680b885R440.

Should I work on making this more explicit - I'll do some Googling but I don't think we're chasing the right problems here.

nsoranzo:
Thanks @jmchilton !

PR #6400, opened 2018-06-25
Client library upgrades.

dannon:
Upgrade vue-loader to v15.  It wasn't backwards compatible and needed minor changes -- this silences the new warnings on client build like `[1] No parser and no filepath given, using 'babylon' the parser now but this will throw an error in the future. Please specify a parser or a filepath so one can be inferred.`

(and then I started pulling the upgrade string)

Upgrades underscore, babel, gulp, prettier, and more.


dannon:
Actually, WIP for a second because I'm going to make a pass upgrading more client libs.

Issue #6396, opened 2018-06-25
Enable resource requests for data_manager tools


moskalenko:

moskalenko:
This is from a discussion at the galaxy admins training session. It seems that data_manager tools do not have the resource request interface hook, so it's not possible to use resources appropriate to the genome size in questions, but a sometimes very high water mark has to be set based on the largest genome as a default instead. It would be great if we could actually set cpus, memory, and time when running data managers. All our admins are real users and have SLURM allocations.

mvdbeek:
So we're limiting the resource selection at https://github.com/galaxyproject/galaxy/blob/65c1cd761c48ba121a847d4d501eb8a97a911e70/lib/galaxy/jobs/__init__.py#L383 to regular tools. I don't think there's a reason why we couldn't include data manager tools. We can look at that during the hackathon, or if you get a chance for testing this I'm happy to change this into `if tool_id and tool_type in ('default', 'manage_data'):`

moskalenko:
I just tested it. I have a bowtie2 builder job running with a custom resource request. Yay! This is great.

Thank you!

Alex

PR #6402, opened 2018-06-26
Allow resource selection for data managers

mvdbeek:
We used to restrict this to standard tools,
but there's no reason why this shouldn't work
for data manager tools. Fixes https://github.com/galaxyproject/galaxy/issues/6396


jmchilton (approved):

moskalenko:
Thank you!

Issue #1715, opened 2016-02-11
Python 3 support.


nsoranzo:
Python 2.7 [will not be maintained past 2020](http://pythonclock.org/).

Moreover, some Galaxy dependencies dropped Python2 support: cachetools, cmd2, cwltool, numpy, schema-salad

Add support for Python >= 3.5 while maintaining support for Python 2.7.

xref.: https://trello.com/c/dZcCVf9I/2702-migrate-to-python-3

### Useful tools and documentation

https://docs.python.org/3/howto/pyporting.html
https://docs.python.org/2/library/2to3.html
http://python-future.org/
https://six.readthedocs.io/
http://python3porting.com/

### Dependencies which need to be ported, dropped or updated
- [x] [bx-python](https://github.com/bxlab/bx-python): https://github.com/bxlab/bx-python/pull/7, https://github.com/bxlab/bx-python/pull/25, https://github.com/galaxyproject/galaxy/pull/5492
- [x] Cheetah: replaced with [Cheetah3 fork](https://github.com/CheetahTemplate3/cheetah3) in https://github.com/galaxyproject/galaxy/pull/5359
- [x] Fabric: fabric/fabric#1424, replaced with [Fabric3](https://github.com/mathiasertl/fabric/) fork in https://github.com/galaxyproject/galaxy/pull/5359
- [x] [feedparser](https://github.com/kurtmckee/feedparser): was copied inside https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/util/sanitize_html.py , removed in https://github.com/galaxyproject/galaxy/pull/5642
- [x] [mercurial](https://www.mercurial-scm.org/): https://www.mercurial-scm.org/wiki/Python3, we should use the command line interface through `subprocess` instead of using the [Mercurial API](https://www.mercurial-scm.org/wiki/MercurialApi) https://github.com/galaxyproject/galaxy/pull/9065
- [x] [NoseHTML](https://github.com/galaxyproject/nosehtml): PR #https://github.com/galaxyproject/nosehtml/pull/1
- [x] [paste](https://github.com/cdent/paste)
- [x] [pulsar](https://github.com/galaxyproject/pulsar): https://github.com/galaxyproject/pulsar/issues/152, https://github.com/galaxyproject/pulsar/commit/5b663c3027ec31ef2ce2c58335333bcdf593b5f8, https://github.com/galaxyproject/pulsar/pull/156, https://github.com/galaxyproject/galaxy/pull/5494
[...]

PR #6403, opened 2018-06-26
Update dev dependencies

nsoranzo:
In particular for NoseHTML 0.4.5, which will enable unit testing under Python 3.

xref. #1715

### DIFF
--- a/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
+++ b/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
@@ -1,7 +1,7 @@
 amqp==2.2.2
 appdirs==1.4.3
 asn1crypto==0.24.0
-babel==2.5.3
+babel==2.6.0
 bagit==1.6.4
 bcrypt==3.1.4
 bdbag==1.3.0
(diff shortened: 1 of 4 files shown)


nsoranzo:
I can't reproduce the unit test failures on my workstation, ideas welcome!

mvdbeek:
Looks like pysam issues: `[E::bgzf_uncompress] Call to inflateInit2 failed: invalid parameter/compression level, or inconsistent stream state`. ~~Maybe the official pysam wheels for python3 are not compatible with what's available on travis ? https://github.com/pysam-developers/pysam/blob/master/buildwheels.sh looks like it should be portable though :/~~
Nevermind, it's the py27-unit tests ... d'oh.

mvdbeek:
Re-ran this one and https://travis-ci.org/galaxyproject/galaxy/jobs/397055540 for the last dev commit to make sure this isn't some new travis change. Very strange.

nsoranzo:
I've fixed it by changing lxml back to version 4.2.1 . I think this is due to https://github.com/lxml/lxml/pull/265 , the wheel for lxml 4.2.2 ships a `libz-a147dcb0.so.1.2.3` which seems to interfere with the one from pysam.

mvdbeek (approved):

mvdbeek:
Cool, thanks for working this all out!

PR #6406, opened 2018-06-27
Add py34-unit (allowed to fail) job to Travis

nsoranzo:
Also speedup the py34-first_startup TravisCI job by skipping most database migrations, as for py27-first_startup.

xref. #1715


jmchilton:
Thanks @nsoranzo !

PR #6409, opened 2018-06-28
WIP: Expose workflow versions

mvdbeek:
There is no UI-element so far, but you can append &version=<N>
to the editor URL to load any arbitrary version of a workflow.


bgruening:
Neat! And the Codefest has not even started! :)

mvdbeek:
Kept me busy on the flight :)

mvdbeek:
Oops, closed that unintentionally, I do want to get that into 18.09

PR #267, opened 2015-05-18
Allow config of shed tool conf used with ``run_tests.sh -installed``.

jmchilton:
Set with `GALAXY_TEST_SHED_TOOL_CONF`.

This is required to implement galaxyproject/planemo#176.

PR #6413, opened 2018-06-29
Replace login-related makos with vue components

guerler:
This PR replaces the login mako with a vue component and connects it with api-like login endpoint in the user controller for Galaxy. It also replaces the change password, and forgotten password options, and the open id login form. Legacy code and makos have been moved to the toolshed. It simplifies the upcoming replacement of the registration mako. 
 
Todos:
 - [x] Redirect for change password view
 - [x] Add test cases
 - [x] Fix/check activation link


bgruening:
Nice!

lib/galaxy/managers/users.py
```diff
@@ -32,39 +32,30 @@ class UserManager(base.ModelManager, deletable.PurgableManagerMixin):
     # TODO: incorp CreatesApiKeysMixin
```
nsoranzo: I think you can now remove the comment line 31.

dannon:
I'm still concerned about the duplication of code here, xref #6512, which has been grafted into this branch.  I'd prefer the existing login route take an `async` argument in the request and return a json blob instead of the redirect, leaving the rest of the handling intact.

guerler:
Yeah, the arguments in #6512 are still valid. I am fine with reducing the toolshed code path separation within this PR.

client/galaxy/scripts/components/login/Login.vue
```diff
@@ -0,0 +1,117 @@
...
+                </b-form-group>
+                <b-button type="submit">Login</b-button>
+            </b-card>
+        </b-form>
+        <br>
+        <b-form v-if="openid_providers && openid_providers.length > 0" @submit.prevent="submit('openid')">
+            <b-card header="OpenID login">
+                <b-form-group label="Enter a URL">
+                    <b-form-input type="text" v-model="url"/>
+                </b-form-group>
```
VJalili: Not many users know what is an OpenID URL, or where to find it. Also, there is not different URLs for an identity provider. We would still be safe relying on the URLs we know; i.e., using the following drop-down where user can choose an OpenID. Therefore, I suggest removing this textbox and show the dropdown only.

client/galaxy/scripts/components/login/Login.vue
```diff
@@ -0,0 +1,117 @@
...
+                <b-form-group label="Password">
+                    <b-form-input type="password" v-model="password"/>
+                    <b-form-text>Forgot password? Click here to <a @click="reset" href="#">reset</a> your password.</b-form-text>
+                </b-form-group>
+                <b-button type="submit">Login</b-button>
+            </b-card>
+        </b-form>
+        <br>
+        <b-form v-if="openid_providers && openid_providers.length > 0" @submit.prevent="submit('openid')">
+            <b-card header="OpenID login">
```
VJalili: Would it be useful to add a note here saying something like "this feature will be deprecated in 19.05 release"?

client/galaxy/scripts/components/login/Login.vue
```diff
@@ -0,0 +1,117 @@
...
+            axios
+                .post(`${Galaxy.root}user/reset_password`, {email: this.login})
+                .then(response => {
+                    this.messageVariant = "info";
+                    this.messageText = response.data.message;
+                })
+                .catch(error => {
+                    this.messageVariant = "danger";
+                    let message = error.response.data && error.response.data.err_msg;
+                    this.messageText = message || "Password reset failed for unkown reason.";
```
VJalili: 1- "unkown" -> "unknown". 
2- "for unknown reason." -> "for an unknown reason."

client/galaxy/scripts/components/login/Login.vue
```diff
@@ -0,0 +1,117 @@
...
+                    } else if (response.data.redirect) {
+                        window.location = response.data.redirect;
+                    } else {
+                        window.location = `${Galaxy.root}`;
+                    }
+                })
+                .catch(error => {
+                    this.messageVariant = "danger";
+                    let message = error.response.data && error.response.data.err_msg;
+                    this.messageText = message || "Login failed for unkown reason.";
```
VJalili: 1- "unkown" -> "unknown".
2- "for unknown reason." -> "for an unknown reason."

client/galaxy/scripts/components/login/Password.vue
```diff
@@ -0,0 +1,66 @@
...
+                    password: this.password,
+                    confirm: this.confirm
+                })
+                .then(response => {
+                    window.location = `${Galaxy.root}`;
+                })
+                .catch(error => {
+                    this.messageVariant = "danger";
+                    let message = error.response && error.response.data && error.response.data.err_msg;
+                    this.messageText = message || "Login failed for unkown reason.";
```
VJalili: 1- "unkown" -> "unknown".
2- "for unknown reason." -> "for an unknown reason."

lib/galaxy/auth/__init__.py
```diff
@@ -99,12 +108,12 @@ def check_change_password(self, user, current_password):
                 auth_result = provider.authenticate_user(user, current_password, options)
                 if auth_result is True:
                     if string_as_bool(options.get("allow-password-change", False)):
-                        return (True, '')  # accept user
+                        return
```
VJalili, line 101: If I read this correct, you return `void` when password change is allowed, and return a message when it is not. I think you decided not to use boolean, because you have two different reasons for "False". However, I think returning a tuple (`boolean, message`) is intuitive. Would it be a challenge to keep it as it is now?

lib/galaxy/auth/__init__.py
```diff
@@ -125,6 +134,111 @@ def active_authenticators(self, email, username, password):
             log.exception("Active Authenticators Failure")
             raise
 
+    def change_password(self, trans, password=None, confirm=None, token=None, id=None, current=None, **kwd):
```
VJalili: Is the `kwd` used here?

lib/galaxy/webapps/galaxy/controllers/user.py
```diff
@@ -82,68 +71,56 @@ def build_initial_query(self, trans, **kwd):
...
     user_openid_grid = UserOpenIDGrid()
     installed_len_files = None
 
-    @web.expose
-    def openid_auth(self, trans, **kwd):
+    def __init__(self, app):
+        super(User, self).__init__(app)
+        self.user_manager = users.UserManager(app)
+
+    def __openid_auth(self, trans, payload=None, **kwd):
```
VJalili: I think `kwd` is not used. If so, could you plz remove it? If you remove it, you may need to update line #267 as well.

lib/galaxy/webapps/galaxy/controllers/user.py
```diff
@@ -807,116 +536,50 @@ def create(self, trans, cntrller='user', redirect_url='', refresh_frames=[], **k
...
-        if activation_token is None:
-            activation_token = hash_util.new_secure_hash(str(random.getrandbits(256)))
-            user.activation_token = activation_token
-            trans.sa_session.add(user)
-            trans.sa_session.flush()
-        return activation_token
+        else:
+            # User activation is OFF, proceed without sending the activation email.
+            message = 'Now logged in as %s.<br><a target="_top" href="%s">Return to the home page.</a>' % (escape(user.email), url_for('/'))
+        return message, status, user, status is None
```
VJalili, line 566: You're still inferring a boolean "status" by `status is None`.  Therefore, I think the combination of (boolean)`success` + (string)`message` was more intuitive than the new combination of (string)`status` + (string)`message`.

lib/galaxy/webapps/galaxy/controllers/user.py
```diff
@@ -554,75 +334,61 @@ def __autoregistration(self, trans, login, password, status, kwd, no_password_ch
...
-        """Validates numerous cases that might happen during the login time."""
-        status = kwd.get('status', 'error')
-        login = kwd.get('login', '')
-        password = kwd.get('password', '')
-        referer = trans.request.referer or ''
-        redirect = kwd.get('redirect', referer).strip()
-        success = False
+    def __validate_login(self, trans, payload=None, **kwd):
+        '''Handle Galaxy Log in'''
+        payload = payload or {}
```
VJalili: Would not it be better to default the payload to `{}`?

lib/galaxy/webapps/galaxy/controllers/user.py
```diff
@@ -952,171 +615,76 @@ def activate(self, trans, **kwd):
                 return trans.show_error_message("You are using an invalid activation link. Try to log in and we will send you a new activation email. <br><a href='%s'>Go to login page.</a>") % web.url_for(controller='root', action='index')
         return
 
-    @web.expose
-    def reset_password(self, trans, email=None, **kwd):
+    @expose_api_anonymous_and_sessionless
+    def change_password(self, trans, payload={}, **kwd):
```
VJalili, line 603: I think a documentation of arguments, payload in particular, would be quite useful here.

lib/galaxy/webapps/galaxy/controllers/user.py
```diff
@@ -952,171 +615,76 @@ def activate(self, trans, **kwd):
...
+        Allows to change own password.
+        """
+        user, message = trans.app.auth_manager.change_password(trans, **payload)
+        if user is None:
+            return self.message_exception(trans, message)
+        trans.handle_user_login(user)
+        return {"message": "Password has been changed."}
+
+    @expose_api_anonymous_and_sessionless
+    def reset_password(self, trans, payload={}, **kwd):
```
VJalili, line 622: I think a documentation of arguments, payload in particular, would be quite useful here. 

Also, I think you're not using `**kwd`; if so, would you mind please remove it?

lib/galaxy/webapps/galaxy/controllers/user.py
```diff
@@ -952,171 +615,76 @@ def activate(self, trans, **kwd):
...
+                                    "Please contact your local Galaxy administrator.")
+        email = payload.get("email")
+        if not email:
+            return self.message_exception(trans, "Please provide your email.")
+        message = validate_email(trans, email, check_dup=False)
+        if message:
+            return self.message_exception(trans, message)
+        else:
+            # Default to a non-userinfo-leaking response message
+            message = ("Your reset request for %s has been received.  "
```
VJalili: Is this `message` used anywhere?!

lib/galaxy/webapps/galaxy/controllers/user.py
```diff
@@ -952,171 +615,76 @@ def activate(self, trans, **kwd):
...
-        return trans.fill_template('/user/reset_password.mako',
-                                   message=message,
-                                   status=status)
+                    trans.log_event('User reset password: %s' % email)
+                except Exception as e:
+                    log.debug(body)
+                    return self.message_exception(trans, "Failed to submit email. Please contact the administrator: %s" % str(e))
+        return {"message": "Reset link has been sent to your email."}
 
     def __validate(self, trans, params, email, password, confirm, username):
```
VJalili: `params` is not used, if it is not needed, I would recommend removing it.

lib/galaxy/webapps/galaxy/controllers/user.py
```diff
@@ -952,171 +615,76 @@ def activate(self, trans, **kwd):
...
-            trans.user.preferences[pref_name] = pref_value
-            trans.sa_session.flush()
-
-    @web.expose
-    def log_user_action_async(self, trans, action, context, params):
-        """ Log a user action asynchronously. If user is not logged in, do nothing. """
-        if trans.user:
-            trans.log_action(trans.get_user(), action, context, params)
-
     def __get_redirect_url(self, redirect):
```
VJalili, line 635: Would it be better to default `redirect` to `None`? 
and at line 683 check for `None` instead of `"None"`?

templates/user/openid_manage.mako
```diff
@@ -1,6 +1,29 @@
...
 <%def name="body()">
     <h2>Associate more OpenIDs</h2>
     ${render_openid_form( redirect, True, openid_providers )}
-</%def>
\ No newline at end of file
+</%def>
+
+<%def name="render_openid_form( redirect, auto_associate, openid_providers )">
+    <div class="toolForm">
+        <div class="toolFormTitle">OpenID Login</div>
```
VJalili, line 9: Do you think putting a text such as "this feature will be deprecated in 19.05 release" would be useful?

templates/user/openid_manage.mako
```diff
@@ -1,6 +1,29 @@
...
+        <form name="openid" id="openid" action="${h.url_for( controller='user', action='login' )}" method="post" target="_parent" >
+            <div class="form-row">
+                <label>OpenID URL:</label>
+                <input type="hidden" name="method" value="redirect"/>
+                <input type="text" name="openid_url" size="60" style="background-image:url('${h.url_for( '/static/images/openid-16x16.gif' )}' ); background-repeat: no-repeat; padding-right: 20px; background-position: 99% 50%;"/>
```
VJalili: Similar to my other comment, I think not many users know what is an OpenID URL, or where to find it. Also, there is not different URLs for an identity provider (it is usually only one). We would still be safe relying on the URLs we know; i.e., using the following drop-down where user can choose an OpenID. Therefore, I suggest removing this textbox and show the dropdown only.

VJalili (changes requested):

VJalili:
@guerler, I very much liked that you moved all the logics from the API to their manager. Thanks.

This looks good to me. I added few minor comments. 

One point about how it looks on screen: (I don't know if it is on my end only, or it is the intended behavior) I see text boxes span as wide as window size (see the following screenshot). I think, it would be more elegant to centralize them horizontally and vertically, and limit their width to a much smaller width. 

screen shot 2018-07-30 at 12 05 59 pm

client/galaxy/scripts/components/login/Login.vue
guerler: Done.

client/galaxy/scripts/components/login/Login.vue
guerler: Done.

client/galaxy/scripts/components/login/Login.vue
guerler: Done.

client/galaxy/scripts/components/login/Login.vue
guerler: Done.

client/galaxy/scripts/components/login/Password.vue
guerler: Done.

lib/galaxy/auth/__init__.py
guerler: This function is only called once within the same module. It's trivial to change it but I am not sure how useful it would be.

lib/galaxy/webapps/galaxy/controllers/user.py
guerler: Done.

lib/galaxy/webapps/galaxy/controllers/user.py
guerler: Done.

templates/user/openid_manage.mako
guerler: Done.

templates/user/openid_manage.mako
guerler: Done.

lib/galaxy/auth/__init__.py
guerler: Done.

lib/galaxy/webapps/galaxy/controllers/user.py
guerler: Done.

lib/galaxy/webapps/galaxy/controllers/user.py
guerler: Done.

lib/galaxy/webapps/galaxy/controllers/user.py
guerler: This is temporary, in the follow-up of this PR the Registration mako will be replaced and this status parameter will be dropped.

lib/galaxy/webapps/galaxy/controllers/user.py
guerler: Done.

lib/galaxy/auth/__init__.py
VJalili: I think it would make it more intuitive; however, it I would leave it up to you to decide.

lib/galaxy/webapps/galaxy/controllers/user.py
VJalili: Sounds good. Thanks.

guerler:
ping @VJalili

lib/galaxy/managers/users.py
```diff
@@ -236,6 +247,150 @@ def tags_used(self, user, tag_models=None):
...
 
+    def change_password(self, trans, password=None, confirm=None, token=None, id=None, current=None):
+        """
+        Allows to change a user password with a token.
+        """
+        if not token and not id:
+            return None, "Please provide a token or a user and password."
+        if token:
+            token_result = trans.sa_session.query(trans.app.model.PasswordResetToken).get(token)
+            if not token_result or not token_result.expiration_time > datetime.utcnow():
```
VJalili, line 265: IMO the second `not` is not required.

lib/galaxy/managers/users.py
```diff
@@ -236,6 +247,150 @@ def tags_used(self, user, tag_models=None):
...
+                # Save new password
+                user.set_password_cleartext(password)
+                # Invalidate all other sessions
+                if trans.galaxy_session:
+                    for other_galaxy_session in trans.sa_session.query(trans.app.model.GalaxySession) \
+                                                     .filter(and_(trans.app.model.GalaxySession.table.c.user_id == user.id,
+                                                                  trans.app.model.GalaxySession.table.c.is_valid == true(),
+                                                                  trans.app.model.GalaxySession.table.c.id != trans.galaxy_session.id)):
+                        other_galaxy_session.is_valid = False
+                        trans.sa_session.add(other_galaxy_session)
```
VJalili, line 305: Since you're updating an existing record, you may not need to `add` it to the database, though it needs to be tested. The following `flush` might be enough to commit your updates.

VJalili:
- As I mentioned in the previous comment, the login page spans whole browser width; maybe it would be better to limit its width and align them horizontally to the center of screen. 
- In the register page, some UI elements have a maximum width, but some span throughout whole window (see the following screenshot); I think all elements should have a maximum width and centered horizontally. 

screen shot 2018-08-24 at 3 35 20 pm

- Would not it be useful to show new password strength when changing a current password?

guerler:
@VJalili thanks for the review and your comments. I revised the page to limit the width for input elements through columns. The login page now looks as shown belown. The latter two comments are related to the registration mako replacement PR #6621 and I will revisit them there once this PR is merged. I think we should merge this PR after the upcoming freeze, early in the release cycle.

loginpage

VJalili:
@guerler Thanks for updating this. It looks much nicer now. 

Also, I agree that this PR and #6621 need to be merged as soon as possible (ideally for the next release), since we do have some parallel work that depend on the functionality you implemented in these two PRs.

VJalili (approved):

dannon:
@VJalili I think @guerler was suggesting the opposite, that we merge this *after* the freeze and not before.

VJalili:
Thats correct. Indeed I was wondering if that would be possible to merge this soon so that we can have it in the next release.

VJalili:
@guerler is this ready to be reviewed and merged?

PR #6414, opened 2018-06-29
Implement authnz API endpoints

VJalili:
Authnz was initially implemented as a controller; however, to adhere with the design decisions of the new login architecture, the controllers are replaced by API endpoints. 

The core functionality remains intact; however, some interactions are changed. For instance, previously the `login` controller was redirecting to the given IdP, while the new `login` API returns a URL to be called to initiate authentication. 

ping @guerler @dannon

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/authnz.py
+++ b/lib/galaxy/webapps/galaxy/api/authnz.py
@@ -108,7 +108,7 @@ class OIDC(BaseAPIController):
         trans.handle_user_login(user)
         return {"message": message}
 
-    @web.expose
+    @web.expose_api
     @web.require_login("authenticate against the selected identity provider")
     def disconnect(self, trans, provider, **kwargs):
         """
--- a/lib/galaxy/webapps/galaxy/api/authnz.py
+++ b/lib/galaxy/webapps/galaxy/api/authnz.py
@@ -8,12 +8,12 @@ import logging
 
 from galaxy import web
 from galaxy.web import url_for
-from galaxy.web.base.controller import BaseUIController
+from galaxy.web.base.controller import BaseAPIController
 
 log = logging.getLogger(__name__)
 
 
-class OIDC(BaseUIController):
+class OIDC(BaseAPIController):
 
     @web.expose
     @web.require_login("list third-party identities")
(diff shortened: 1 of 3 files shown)


lib/galaxy/webapps/galaxy/api/authnz.py
```diff
@@ -0,0 +1,140 @@
...
+        :param trans: Galaxy web transaction.
+
+        :type  provider: string
+        :param provider: The name of the provider which should be used to authenticate a user.
+
+        :rtype:  string
+        :return: A URL that should be to initiated a user authentication process.
+        """
+        if not trans.app.config.enable_oidc:
+            raise MessageException(
```
jmchilton, line 58: Should probably throw a ConfigDoesNotAllowException here instead. https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/exceptions/__init__.py#L133

lib/galaxy/webapps/galaxy/api/authnz.py
```diff
@@ -0,0 +1,140 @@
...
+
+        :param kwargs: keyword arguments returned from the OIDC identity provider. It must contain
+        `state` and `code` arguments, among others.
+
+        :rtype:  string
+        :return: a success/error message as a result of handling the callback.
+        """
+        user = trans.user.username if trans.user is not None else 'anonymous'
+        if not bool(kwargs):
+            raise MessageException(
```
jmchilton, line 84: RequestParameterMissingException or RequestParameterInvalidException instead of MessageException I think.

lib/galaxy/webapps/galaxy/api/authnz.py
```diff
@@ -0,0 +1,140 @@
...
+
+        :type  provider: string
+        :param provider: The name of the provider which should be used to authenticate a user.
+
+        :param kwargs:
+
+        :rtype:
+        :return:
+        """
+        if trans.user is None:
```
jmchilton, line 130: If you want to check this, I'd just do ``assert trans.user``. This has a ``require_login`` already, so this isn't going to be reached right? Thus an assertion as a sanity check should be used if anything at all.

jmchilton:
Looks good overall, I think refactoring this into an API endpoint makes a lot of sense!

VJalili:
I think for now we only return json object (see #6517 - ping @dannon ), and maybe implement this as an API later. Therefore, I'm closing it now, and we'll reopen if we decide to use an API-based implementation.

PR #6418, opened 2018-06-29
Fixes for uploading empty files to Galaxy.

jmchilton:
Don't enable it by default, but if check content is off it should be allowed. Also if an empty file is sniffed - don't hang indefinitely - which is what happen now a proteomics datatype sniffer.

#GCCBOSC #CoFEST


bgruening (approved):

jmchilton:
Thanks for the merge @bgruening - I'll be sure to loudly thank you for contributing to the CWL implementation effort at the codefest update ;).

PR #6421, opened 2018-06-30
Fix panel drag resize.

jxtx:
The mouse position and dragging were out of sync, sometimes even
resizing in the wrong direction (when dragging slowly). This switches
from a relative calculation to an absolute one which solves the
problem. Might have been events out of order or dropped.


martenson (approved):
The funky dragging seems to be gone with this PR.

PR #6422, opened 2018-06-30
Initial work on updating styles

jxtx:
Still WIP, working on some potential changes for default styles

- System fonts
- Eliminate "lavender"
- Improve default colors
- Fewer borders
- More flex (resizeable) components
- Better / more consistent use of whitespace (far from complete)


jxtx:
image

bgruening:
I think this looks a lot cleaner then the current interface. Nice work!

bgruening:
@jxtx this is great!
A few things I have found:

grafik

grafik

* the history boxes do have too much padding imho
* the peek-preview of datasets looks too small and lost
* the click-target of tool categories could be larger

martenson:
Adding this tentatively to the 19.01 milestone. This will require a lot of testing but if it is identified as a priority we can get it done I believe.

mvdbeek:
This is really awesome, it already looks very mature and is a huge improvement.
One small itch is the grey help text on white background, a little more contrast here would be good imo.

screen shot 2018-09-07 at 16 21 10

Maybe black and italic would be good enough ?

guerler:
@mvdbeek Thanks I fixed that by restoring the default text-muted/gray color. Let me know if you notice other things.

mvdbeek:
Yeah, that looks better. Should I do the full review or do you want to work more on this ?
One other thing I noticed is that validator looks a bit too big for my taste (and maybe the text could be bigger and/or bold?):

screen shot 2018-09-07 at 21 17 16

guerler:
@mvdbeek Yeah that sounds great. Thank you. Since this PR is already pretty big my goal would be to look for major issues which need to be fixed and then follow-up with a PR improving minor issues and style changes. Thanks again!

mvdbeek:
Awesome, I will do that!

mvdbeek:
Alright, so these things I have found:

- [x] in multi-select mode the checkboxes are slightly cut off:
screen shot 2018-09-11 at 10 23 15

- [ ] The done button for history and HDA annotations is too big:
screen shot 2018-09-11 at 10 24 28

- [x] The placement of the integer box here seems off (and the tool header space could be used better)
screen shot 2018-09-11 at 10 26 21

- [ ] Tool form preview in the workflow editor still uses the old style
screen shot 2018-09-11 at 10 31 27

- [ ] View workflow also still uses the old style
screen shot 2018-09-11 at 10 38 25

- [ ] Share workflow uses old style:
screen shot 2018-09-11 at 10 39 22

- [ ] Visualizations partly use old style
screen shot 2018-09-11 at 10 40 16

- [ ] Importing workflow uses old style:
screen shot 2018-09-11 at 10 41 52

-  [ ] Various forms in the admin interface use the old style
screen shot 2018-09-11 at 10 42 31
screen shot 2018-09-11 at 10 43 08
screen shot 2018-09-11 at 10 43 28
screen shot 2018-09-11 at 10 43 33

None of these are showstoppers IMO, bu we should get the selenium tests working before merging.

guerler:
@mvdbeek Thanks for collecting these issues, I created a separate issue to keep track of them since this PR is already pretty big. However I fixed two issues which obviously did not look right.

test/selenium_tests/test_jupyter.py
```diff
@@ -1,77 +0,0 @@
-from base.populators import flakey
```
jmchilton: well that isn't right... are these just failing locally (that is kind of expected)

guerler: I don't know, they sometimes seem to fail on both and I keep getting a varying number of failing tests since the last 2-3 commits for the last few days. Now I am trying to disable all failing tests first to get an overview.

client/galaxy/scripts/components/RuleCollectionBuilder.vue
```diff
@@ -206,7 +206,7 @@
                                 </column-selector>
                             </div>
                             <div class="buttons rule-edit-buttons d-flex justify-content-end">
-                                <button v-b-tooltip.hover :title="titleAddColumnDefinition" type="button" class="dropdown-toggle btn btn-primary mr-1" data-toggle="dropdown" v-if="unmappedTargets.length > 0">
```
jmchilton, line 209: We should do a better job here - find a way to make that disappear, appear in a different place, or drop the title property from the Vue model also so we don't have an unused variable?

jmchilton:
I can't write to this branch - but here is a commit (https://github.com/jmchilton/galaxy/commit/e40bbfd93a99800ea61a7686fb8fd5e70ce5adcb) that fixes the tour tests and the existing tours for the changes in this branch.

jmchilton:
A set of screenshots from across the GUI for a recent Selenium run on this branch if anyone wants to review - https://jenkins.galaxyproject.org/job/docker-selenium/3622/artifact/3622-test-screenshots/.

client/galaxy/scripts/components/RuleCollectionBuilder.vue
guerler: Yes, ideally we find a way to hide the tooltip in the selenium tests, before clicking on the item. I can look into it. I tried different orientations of the tooltip to workaround it but the test still failed.

mvdbeek (approved):
That's a huge step forward, I love this!

jmchilton:
I think this increases the memory footprint of Galaxy by a significant amount based on the changes needed to make the Selenium suite pass - but I think it is best to get this merged early in the cycle and before it stales - we can resolve the remaining issues in dev hopefully.

jmchilton (approved):

dannon:
I'd like a chance to go through this again with the more recent changes, if you're fine with it sitting for another day @guerler?

guerler:
This has not changed in the last week, we only worked on getting the selenium tests running. Having said this, of course it is fine to merge this tomorrow if you want to go through it today. Thanks a lot for all the help @jmchilton and @mvdbeek.

dannon:
@guerler I meant recent as in the last 30-40 commits or so -- I was tracking it early on but then quit to wait for it to cool off, and then have been busy with release and other stuff the past week :)

And, cool, will plan on merging it tomorrow after review/changes, I have a few in mind.

guerler:
@dannon sounds good. Also feel free to add to #6706 if you have ideas for a follow-up.

dannon:
The outcome looks great.

The implementation, however, has a lot I'm really worried about in terms of throwing away our abstraction layer that sits on top of whatever UI library we're using.  That we don't have a "Galaxy UI Elements" abstraction layer that is *rigidly* adhered to is exactly how we ended up with the CSS sprawl and style divergence that we have now.  The shift straight to bootstrap *does* obviously clean up a lot by unifying styles under a certain set of elements, but I think we really need to rethink the double-down, and get our abstraction layer back.

All that said, @guerler convinced me that this is two steps forward and one back, and not the other way around, so I'm going to just go ahead and merge and we'll follow up afterward.  We've got a whole (long) release cycle to get it worked on before 19.01.

jxtx:
I'm always the one arguing for abstraction, but we can always override the base styles as much as we want. So I'm not really concerned about this. I appreciate the uniformity and simplification this gives us.

martenson:
Great job @guerler! Nice styles @jxtx!

bgruening:
:fireworks:

dannon:
@jxtx Yeah, don't get me wrong, I definitely appreciate the simplification, love the refreshed look, and so on.  You mean overriding the bootstrap base styles, or something else?

I do think it's important to keep an abstraction layer between us and bootstrap -- in the sense that Galaxy's `gx-alert` is just an extension of bootstrap's `alert`, with whatever customizations/overrides there, but it's always the `gx-alert` that we're using.  That leaves us free to swap in Bulma or something else down the road without terrible effort.  Even if we never use anything else, it'd also make bootstrap upgrades easier -- a lot of the pain points from the last upgrade were where we'd used classes directly and had to grep around to find out what was broken.  One (admittedly minor, but the latest thing I had to fix) example is the bootstrap 3->4 change from `label`s to `badge`s -- if we'd had a `ui-tag` that extended the appropriate class, we'd have had to change exactly one line to get it all right, instead of grepping all over the codebase, missing one, and so on.

guerler:
I think that this PR is a step forward and simplifies our style related code-base and that it will make it easier to add a consistent abstraction layer if we decide to do so in the future. The customizations in ui.scss, base.scss and other module specific files have gone too far and folding them back will actually allow us to regain control and apply an appropriate strategy moving forward.

PR #6427, opened 2018-06-30
Citation style work

dannon:
Got busted somehow in recent work (in dev), so I took the opportunity to use bs-vue and refresh the styles and formatting a bit.


bgruening:
Cool, thanks @dannon!

martenson:
The style of this is based on https://github.com/galaxyproject/galaxy/pull/6422 which I suspect will not get merged that soon since it requires a lot of testing. Do we want to drift freely towards that with isolated components?

screenshot 2018-07-01 12 51 27


A minor thing is that now the bibtex is not rendered within editable textarea, which makes it harder to select.

PR #6429, opened 2018-06-30
Python 3 fixes from GCC-BOSC-2018

johnbradley:
Several fixes where I added assert instead of relying on doctest have trouble comparing values.
Changes in the returned values by python causing issues with doctest.
Changes in IOString properties.
Also includes a work around for importing maketrans.

Helps with #1715


mvdbeek:
@galaxybot test this

nsoranzo:
@galaxybot test this

nsoranzo (approved):
Brilliant fixes, thanks for working on this @johnbradley!

PR #6435, opened 2018-07-01
introduce ability to favorite a tool

martenson:
* Create tool favoriting API that can be easily expanded to other objects.
* Create button on every toolform to favorite/unfavorite tool.

~~I wanted to use `axios.delete` but had trouble with Routes parsing the wildcard patterns when given the tool_id encoded in the url. Would still be better API I think if anyone has an idea how to do it. ping @mvdbeek~~


lib/galaxy/webapps/galaxy/buildapp.py
```diff
@@ -628,6 +628,18 @@ def populate_api_routes(webapp, app):
...
+                          '/api/users/{id}/favorites/tools',
+                          controller='users',
+                          action='set_favorite_tool',
+                          conditions=dict(method=["PUT"]))
+
+    webapp.mapper.connect('remove_favorite_tool',
+                          '/api/users/{id}/favorites/tools',
+                          controller='users',
+                          action='remove_favorite_tool',
+                          conditions=dict(method=["POST"]))
```
mvdbeek: DELETE should work after https://github.com/galaxyproject/galaxy/pull/6467 is merged, since it upgrades to recent WebOb, which introduced DELETE with a body in version 1.7 https://docs.pylonsproject.org/projects/webob/en/stable/changes.html. But wouldn't it be better to have the route be `/api/users/{id}/favorites/tools/{tool_id}` ? Than POST and DELETE would feel pretty natural to me.

martenson: >trouble with Routes parsing the wildcard patterns when given the tool_id encoded in the url

lib/galaxy/webapps/galaxy/api/users.py
```diff
@@ -510,6 +510,38 @@ def set_information(self, trans, id, payload={}, **kwd):
         trans.log_event('User information added')
         return {'message': 'User information has been saved.'}
 
+    @expose_api
+    def set_favorite_tool(self, trans, id, payload={}):
```
guerler: I think we should generalize this slightly such that we wont have to add individual endpoints for other `favorite` items e.g. workflows, visualizations etc. by creating an endpoint `set_favorites` instead.

martenson: Sounds good to me, I can do that.

martenson: This is now in.

lib/galaxy/webapps/galaxy/buildapp.py
martenson: tried few more things and it seems `/{object_id:.*?}` works, so I will refactor this as well

martenson: This is now in.

martenson:
superseded by https://github.com/galaxyproject/galaxy/pull/7209

PR #6436, opened 2018-07-01
fix a few tool_deps test errors

bgruening:
Leftover changes from yesterday.

xref. #1715

### DIFF
--- a/test/unit/tools/test_tool_deps.py
+++ b/test/unit/tools/test_tool_deps.py
@@ -499,7 +499,7 @@ def __test_base_path():
 
 
 def test_parse():
-    with __parse_resolvers('''<dependency_resolvers>
+    with __parse_resolvers(b'''<dependency_resolvers>
   <tool_shed_packages />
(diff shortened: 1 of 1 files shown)


test/unit/tools/test_tool_deps.py
```diff
@@ -499,23 +499,23 @@ def __test_base_path():
 
 
 def test_parse():
-    with __parse_resolvers('''<dependency_resolvers>
+    with __parse_resolvers(b'''<dependency_resolvers>
```
nsoranzo, line 502: I think these additions of `b` are wrong, the proper solution is changing line 634 of this file to use `tempfile.NamedTemporaryFile('w+')` to open the temporary file in text mode.

@bgruening Can you open a PR to fix it?

nsoranzo: Addressed in https://github.com/galaxyproject/galaxy/pull/6455 .

PR #6437, opened 2018-07-01
declutter the history 'cog' menu

martenson:
- [x] move the 'import from file' to the saved histories page

### DIFF
--- a/client/galaxy/scripts/mvc/history/options-menu.js
+++ b/client/galaxy/scripts/mvc/history/options-menu.js
@@ -202,16 +202,6 @@ var menu = [
         html: _l("Export History to File"),
         href: "history/export_archive?preview=True",
         anon: true
-    },
-
-    {
-        html: _l("Other Actions"),
-        header: true
-    },
-    {
-        html: _l("Import from File"),
-        href: "histories/import",
-        target: "_top"
     }
 ];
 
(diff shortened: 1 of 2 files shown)


dannon:
@martenson Looks like a reasonable first step forward -- I'm not sure how to best approach most of the other stuff.  Want to just get this merged and take the rest of it in separate PRs?

martenson:
ready for review

client/galaxy/scripts/mvc/history/options-menu.js
```diff
@@ -172,16 +172,6 @@ var menu = [
...
         href: "history/export_archive?preview=True",
         anon: true
-    },
-
-    {
-        html: _l("Other Actions"),
-        header: true
-    },
-    {
-        html: _l("Import from File"),
```
nsoranzo, line 212: What about putting it together with the previous section, to be renamed from `Downloads` to `Import/Export` ?

martenson: I still think it does not belong in the current history 'context' menu at all.

nsoranzo: We have "Create New" under the "Current History" header...

martenson: That should imo go away too, along with other options.

nsoranzo: Thanks for #7565 ! IMO import and export should be together, but it's a -0.

jmchilton (approved):

martenson:
thank you @nsoranzo and @jmchilton

PR #6438, opened 2018-07-01
fixes test/unit/jobs/test_job_output_checker.py; strange default?

caleb-easterly:
Unclear on the implications of this. 

In `lib/galaxy/tools/parser/interface.py`, the default for `ToolStdioRegex.error_level` was `"fatal"`. The following test passes under Python2 but fails under Python3: 

https://github.com/galaxyproject/galaxy/blob/01ba96a7acda5f321e3f3f0975909615122a7880/test/unit/jobs/test_job_output_checker.py#L62-L68

This is because, in the following

https://github.com/galaxyproject/galaxy/blob/01ba96a7acda5f321e3f3f0975909615122a7880/lib/galaxy/jobs/error_level.py#L21-L26

`'fatal' > 0 and 'fatal' <= StdioErrorLevel.MAX` is `false` in Python 2 but throws an error in Python 3. It seems like these shouldn't be compared at all...but I might not have fixed it in the right way.

xref. #1715


martenson:
@galaxybot test this

mvdbeek (approved):
Thanks @caleb-easterly!

dannon:
This is the correct fix, but we should probably use the constant `StdioErrorLevel.FATAL` (which is 3).  Want to push one more changeset, or should I?  (and you can drop the TODO as well, since it's suggesting to use the constant here)

caleb-easterly:
I probably wouldn't get to it for an hour or so - feel free to make the necessary changes if you're looking at it now!

dannon:
@caleb-easterly Sure, I'll do it now, thanks!

lib/galaxy/tools/parser/interface.py
```diff
@@ -5,6 +5,8 @@
 
 import six
 
+from galaxy.jobs.error_level import StdioErrorLevel
```
mvdbeek: I was discussing that with @jmchilton yesterday, and we didn't want to import from galaxy.jobs to keep this separate. We could move the error_level code into tools though ?

dannon: @mvdbeek Sure, that sounds good to me -- happy to move it wherever, I just want to use the constant.

jmchilton: Pushed a fix for this, the refactoring I was hoping for actually eliminates some need for delayed imports and such so we closer to untangling circular imports in Galaxy I think. Thanks for working on this all!

nsoranzo:
I've rebased and updated @jmchilton's last commit to also update the import in `test/unit/jobs/test_job_output_checker.py`

nsoranzo (approved):

PR #6441, opened 2018-07-01
Add zoom support to workflow editor.

jxtx:
Zoom levels are discrete from a pre-defined set. Uses css transforms
and... math.


martenson:
Seems to work nice in my testing on Chrome/macos. The only thing I noticed is that when on maximum zoom you lose the ability to drag canvas/elements and you are stuck until you click `-` which does not zoom you out but enables dragging.

dannon (approved):

PR #6443, opened 2018-07-01
Allows DatasetMatcher to use multiple filters

ebolyen:
In particular this enables multiple add_value filters to be used with options's option_filter_attribute.

This will be useful for filtering input based on QIIME 2 semantic types without requiring a unique format class/registration for every format backing a QIIME 2 artifact.

Thanks to @blankenberg for the assistance!

First time contributing, let me know what needs to be changed! :smiley:

### DIFF
--- a/lib/galaxy/tools/parameters/dataset_matcher.py
+++ b/lib/galaxy/tools/parameters/dataset_matcher.py
@@ -99,13 +99,14 @@ class DatasetMatcher(object):
         self.trans = trans
         self.param = param
         self.tool = param.tool
-        filter_value = None
+        filter_values = set()
         if param.options and other_values:
             try:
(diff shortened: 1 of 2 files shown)


mvdbeek:
@galaxybot test this

ebolyen:
FYI: it's possible this has broken the ability to use `<configfiles>` and `<param type="data"><options><filter>`, but not certain yet. See https://github.com/qiime2/q2galaxy/issues/1 for details.

ebolyen:
I don't think this patch has anything to do with the above situation, configfiles and params of type data have just never composed.

jmchilton:
Looks great, thanks for this and thanks for the tests!

PR #6447, opened 2018-07-02
Changing menu tabs without hard refresh

almahmoud:
**Changes:**
-Menu clicks don't hard refresh
-Some user preferences don't hard refresh
-Saved visualizations highlight User tab

**Not fixed:**
-Visualization tabs still don't get highlighted when their options are clicked

**Problem:**
-Analysis tab doesn't become active when switching


dannon:
@almahmoud Thanks!  This is a nice step forward.

bgruening:
@almahmoud this is super cool! We have many users that are annoyed by the reloads :)
Keep the PRs coming!

PR #6449, opened 2018-07-02
CI - ignore node_modules anywhere

dannon:
Since we use them in multiple places and not always 'client' (proxy, plugins, etc).


nsoranzo:
Good catch!

PR #6450, opened 2018-07-02
Fix all remaining unit tests under Python 3

mvdbeek:
xref. #1715

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,6 @@ matrix:
     - env: TOX_ENV=py34-unit
       addons: *py3_addons
   allow_failures:
-    - env: TOX_ENV=py34-unit
     - env: TOX_ENV=check_python_dependencies
 
 
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1546,7 +1546,7 @@ class JobWrapper(HasResourceParameters):
         return None
 
     def has_limits(self):
-        has_output_limit = self.app.job_config.limits.output_size > 0
+        has_output_limit = self.app.job_config.limits.output_size and self.app.job_config.limits.output_size > 0
         has_walltime_limit = self.app.job_config.limits.walltime_delta is not None
         return has_output_limit or has_walltime_limit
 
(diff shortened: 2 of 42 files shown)


mvdbeek:
I think all unit tests should be passing now, and I even managed to run a couple of tools successfully!

lib/galaxy/jobs/__init__.py
```diff
@@ -1546,7 +1546,7 @@ def check_limits(self, runtime=None):
         return None
 
     def has_limits(self):
-        has_output_limit = self.app.job_config.limits.output_size > 0
+        has_output_limit = self.app.job_config.limits.output_size and self.app.job_config.limits.output_size > 0
```
nsoranzo, line 1549: Need to do the same at line 1533

mvdbeek:
Hmm, the failing tests in https://travis-ci.org/mvdbeek/galaxy/jobs/399571214 pass fine locally, but I'm running 3.6 instead of 3.4 ... do you think it's worth switching travis to 3.6 ?

lib/galaxy/jobs/__init__.py
mvdbeek: Good catch, done that in aaf968c

lib/galaxy_ext/metadata/set_metadata.py
```diff
@@ -26,7 +26,7 @@
 from galaxy.util import stringify_dictionary_keys
 
 # ensure supported version
-assert sys.version_info[:2] >= (2, 6) and sys.version_info[:2] <= (2, 7), 'Python version must be 2.6 or 2.7, this is: %s' % sys.version
+assert sys.version_info[:2] >= (2, 7) and sys.version_info[:2] <= (3, 7), 'Python version must be at least 2.7, this is: %s' % sys.version
```
nsoranzo: Can we just drop the `sys.version_info[:2] <= ()` ?

nsoranzo:
@natefoo Can you help with the last remaining [unit test error](https://travis-ci.org/galaxyproject/galaxy/jobs/399960334) which involves `lib/galaxy/containers/docker_swarm.py` ?

nsoranzo (approved):

nsoranzo:
All green!

.travis.yml
```diff
@@ -34,7 +34,6 @@ matrix:
     - env: TOX_ENV=py34-unit
       addons: *py3_addons
   allow_failures:
-    - env: TOX_ENV=py34-unit
```
martenson, line 37: 🎉

PR #6451, opened 2018-07-02
Fix droptarget:drop trigger (drag/drop in the multiview)

### LINKED ISSUE #6393: drag and drop of data sets between histories in the multiple histories view seems broken
when dragging a data set to the first column the box for dropping appears but the data set is not copied. 

affects my dev instance, but on works on 18.05

dannon:
Wasn't passing the panel after some recent refactoring.

Fixes #6393


martenson:
thanks @dannon

PR #6452, opened 2018-07-02
py3: fixes test_DatasetManager.py:DatasetManagerTestCase

caleb-easterly:
Looks like there was a plan to cast/extract to list anyway?

xref. #1715

### DIFF
--- a/lib/galaxy/managers/rbac_secured.py
+++ b/lib/galaxy/managers/rbac_secured.py
@@ -87,16 +87,15 @@ class DatasetRBACPermission(RBACPermission):
     def by_dataset(self, dataset):
         self.__assert_action()
         all_permissions = self._all_types_by_dataset(dataset)
-        return filter(lambda p: p.action == self.action_name, all_permissions)
+        return list(filter(lambda p: p.action == self.action_name, all_permissions))
 
-    # TODO: list?
     def by_roles(self, dataset, roles):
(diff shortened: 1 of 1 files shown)


lib/galaxy/managers/rbac_secured.py
```diff
@@ -87,16 +87,15 @@ def __assert_action(self):
...
-    # TODO: list?
     def by_roles(self, dataset, roles):
         permissions = self.by_dataset(dataset)
-        return filter(lambda p: p.role in roles, permissions)
+        return list(filter(lambda p: p.role in roles, permissions))
 
     def by_role(self, dataset, role):
         permissions = self.by_dataset(dataset)
-        found = filter(lambda p: p.role == role, permissions)
+        found = list(filter(lambda p: p.role == role,  permissions))
```
mvdbeek: The extra space will make the linting fail

mvdbeek (approved):

lib/galaxy/managers/rbac_secured.py
caleb-easterly: good eye, thanks!

mvdbeek:
I had a similar change in https://github.com/galaxyproject/galaxy/pull/6450 and the tests passed there, so I'll merge this now.

nsoranzo:
For future reference, list comprehensions would make the code clearer IMO.

PR #6453, opened 2018-07-02
py3: Fix the dynamic tool destination unit test

mvdbeek:
Strings are greater than Integers in py2, but in py3 the comparison is invalid, so I assume setting the upper bound to float('inf') should be correct. The records rule also accepts file size units, so I guess passing this to `str_to_bytes` is fine.
Does that make sense @ericenns ?

xref. #1715


ericenns:
Looking at the code, we are setting valid_rule to false in that block so setting lower_bound to float('inf') is correct as in that context the rule will never be evaluated.

Issue #5894, opened 2018-04-17
Access to shared histories


joachimwolff:
I want to share a history via link. If I do so, the data items are not viewable. (Permissions on first dataset were manually changed.)

![image](https://user-images.githubusercontent.com/13435136/38861342-9f772786-4232-11e8-9975-5ba85bf749d3.png)

The only way to change this is to change the access permissions in my history for every item by hand (!!!). It would be nice if either

- the default option that if I share my history via link, all items are viewable or
- that checkboxes with `make all items viewable`  and / or `make all items editable` are offered. 

Thanks for considering.

hexylena:
This just happened to me to. Super mind numbing to to edit dataset permissions individually. Sharing by link should definitely give a checkbox to remove any access permissions.

hexylena:
Update: there is also no API for this so it isn't possible to automate with bioblend.

hexylena:
This happened again to us and it's really tough to explain to our users: "No sharing doesn't work like all of the other systems, that 'NOTE' in the management screen for access permissions is the most important note and you *need* to read + understand it to do sharing right." It isn't a user problem, it's a user experience/expectations problem from the other web based systems they use. I think there is sufficient evidence that our sharing interface is "clunky" and would like to propose the following (requiring API + UI changes):

I'd like to propose we try and align our interface (where possible) like Google Drive to better match what users will be familiar with. A history is like a folder, and then inside there are datasets. Honestly I think we should copy the dialog wholesale:

![history-sharing](https://user-images.githubusercontent.com/458683/44459052-c24afe80-a5f7-11e8-9861-c260b4ecedb3.png)

And then re-use *exactly* the same sharing form for dataset permissions, just like gdrive does for sharing folders vs datasets. If necessary, put in a "Old sharing form" button at the bottom which shows the old sharing form?

hexylena:
Problem continues with creating pages. From @eschen42 on gitter

>  If I create and make "accessible" Galaxy pages that reference particular datasets (through the page editor, not by pasting links), will this make them not public?

bgruening:
@dannon @guerler what do you think about this? (Easy) sharing is for our users currently broken :(

dannon:
This isn't something we recently broke, right, it has always been like this?

(regardless of the answer to that, yes, we should improve the behavior here somehow -- sharing by a link being worthless is pretty annoying)

blankenberg:
[...]

PR #6461, opened 2018-07-03
Implement dataset permissions api

hexylena:
Copies the code mostly from

https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/controllers/dataset.py#L486-L511

except exchanging the returned plain messages for more appropriate exceptions. In support of https://github.com/galaxyproject/galaxy/issues/5894

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/datasets.py
+++ b/lib/galaxy/webapps/galaxy/api/datasets.py
@@ -93,6 +93,55 @@ class DatasetsController(BaseAPIController, UsesVisualizationMixin):
             trans.response.status = 500
         return rval
 
+    @web._future_expose_api
+    def set_permissions(self, trans, dataset_id, payload, **kwargs):
+        """
+        POST /api/datasets/{encoded_dataset_id}/permissions
+        Updates permissions of a dataset.
+
+        , hda_ldda='hda', manage_ids=[], access_ids=[]
+
+        :rtype:     dict
+        :returns:   dictionary containing new permissions
+        """
+        hda_ldda = kwargs.get('hda_ldda', 'hda')
+        try:
+            hda = self.get_hda_or_ldda(trans, hda_ldda=hda_ldda, dataset_id=dataset_id)
+        except Exception as e:
+            return str(e)
+
+        access = payload.get('access', [])
+        manage = payload.get('manage', [])
+        params = {
+            'DATASET_MANAGE_PERMISSIONS': manage,
+            'DATASET_ACCESS': access,
+        }
+
+        if trans.app.security_agent.can_manage_dataset(trans.get_current_user_roles(), hda.dataset):
+            payload_permissions = {}
+            for action in trans.app.model.Dataset.permitted_actions.keys():
+                payload_permissions[action] = [trans.security.decode_id(role_id) for role_id in util.listify(params.get(action))]
+            # The user associated the DATASET_ACCESS permission on the dataset with 1 or more roles.  We
+            # need to ensure that they did not associate roles that would cause accessibility problems.
+            permissions, in_roles, error, message = \
+                trans.app.security_agent.derive_roles_from_access(trans, hda.dataset.id, 'root', **payload_permissions)
+            if error:
+                # Keep the original role associations for the DATASET_ACCESS permission on the dataset.
+                access_action = trans.app.security_agent.get_action(trans.app.security_agent.permitted_actions.DATASET_ACCESS.action)
+                permissions[access_action] = hda.dataset.get_access_roles(trans)
+                trans.sa_session.refresh(hda.dataset)
+                raise galaxy_exceptions.RequestParameterInvalidException(message)
+            else:
+                error = trans.app.security_agent.set_all_dataset_permissions(hda.dataset, permissions)
+                trans.sa_session.refresh(hda.dataset)
+                if error:
+                    raise galaxy_exceptions.RequestParameterInvalidException(error)
+                else:
+                    return self.hda_serializer.serialize_to_view(hda, view='detailed', user=trans.user, trans=trans)['permissions']
+        else:
+            raise galaxy_exceptions.InsufficientPermissionsException('You are not authorized to change this dataset\'s permissions.')
+
+
     def _dataset_state(self, trans, dataset, **kwargs):
         """
         Returns state of dataset.
--- a/lib/galaxy/webapps/galaxy/api/datasets.py
+++ b/lib/galaxy/webapps/galaxy/api/datasets.py
@@ -103,10 +103,7 @@ class DatasetsController(BaseAPIController, UsesVisualizationMixin):
         :returns:   dictionary containing new permissions
         """
         hda_ldda = kwargs.get('hda_ldda', 'hda')
-        try:
-            hda = self.get_hda_or_ldda(trans, hda_ldda=hda_ldda, dataset_id=dataset_id)
-        except Exception as e:
-            return str(e)
+        hda = self.get_hda_or_ldda(trans, hda_ldda=hda_ldda, dataset_id=dataset_id)
 
         access = payload.get('access', [])
         manage = payload.get('manage', [])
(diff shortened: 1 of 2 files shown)


lib/galaxy/webapps/galaxy/api/datasets.py
```diff
@@ -93,6 +93,55 @@ def show(self, trans, id, hda_ldda='hda', data_type=None, provider=None, **kwd):
             trans.response.status = 500
         return rval
 
+    @web._future_expose_api
+    def set_permissions(self, trans, dataset_id, payload, **kwargs):
+        """
+        POST /api/datasets/{encoded_dataset_id}/permissions
+        Updates permissions of a dataset.
+
+        , hda_ldda='hda', manage_ids=[], access_ids=[]
```
nsoranzo: Need to finish the documentation?

lib/galaxy/webapps/galaxy/buildapp.py
```diff
@@ -262,6 +262,7 @@ def populate_api_routes(webapp, app):
     webapp.mapper.resource('dataset', 'datasets', path_prefix='/api')
     webapp.mapper.resource('tool_data', 'tool_data', path_prefix='/api')
     webapp.mapper.connect('/api/tool_data/{id:.+?}/fields/{value:.+?}/files/{path:.+?}', action='download_field_file', controller="tool_data")
+    webapp.mapper.connect('/api/histories/published', action='published', controller="histories", conditions=dict(method=["GET"]))
```
martenson: is this line here by accident?

static/style/blue/base.css
```diff
@@ -1039,8 +1039,7 @@ pre {
   margin-bottom: 1rem;
   background-color: transparent; }
   .table th, .ui-table-striped th,
-  .table td,
-  .ui-table-striped td {
+  .table td, .ui-table-striped td {
```
martenson: this whole file has been added to the commit accidentally I think

martenson (commented):
This is worth checking, we might be able to somehow unite these? https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/api/library_datasets.py#L230

It would probably result in an extra API call for libraries though :/

lib/galaxy/webapps/galaxy/api/datasets.py
```diff
@@ -93,6 +93,53 @@ def show(self, trans, id, hda_ldda='hda', data_type=None, provider=None, **kwd):
...
+                error = trans.app.security_agent.set_all_dataset_permissions(hda.dataset, permissions)
+                trans.sa_session.refresh(hda.dataset)
+                if error:
+                    raise galaxy_exceptions.RequestParameterInvalidException(error)
+                else:
+                    return self.hda_serializer.serialize_to_view(hda, view='detailed', user=trans.user, trans=trans)['permissions']
+        else:
+            raise galaxy_exceptions.InsufficientPermissionsException('You are not authorized to change this dataset\'s permissions.')
+
+
```
martenson: lint does not like the extra empty line

static/style/blue/base.css
hexylena: Ugh how did that get there :(

lib/galaxy/webapps/galaxy/buildapp.py
hexylena: Yes. No idea where they came from

hexylena:
I'm not sure how we'd go about unifying that in a clean way? One has three permissions and does a bunch of library calls, the other has two and does dataset calls?

martenson:
@erasche yeah, I do not have a marvelous idea here, maybe it is not worth trying

lib/galaxy/webapps/galaxy/api/datasets.py
```diff
@@ -93,6 +93,52 @@ def show(self, trans, id, hda_ldda='hda', data_type=None, provider=None, **kwd):
             trans.response.status = 500
         return rval
 
+    @web._future_expose_api
+    def set_permissions(self, trans, dataset_id, payload, **kwargs):
+        """
+        POST /api/datasets/{encoded_dataset_id}/permissions
```
jmchilton: This doesn't match the route added in buildapp right?

lib/galaxy/webapps/galaxy/api/datasets.py
```diff
@@ -93,6 +93,52 @@ def show(self, trans, id, hda_ldda='hda', data_type=None, provider=None, **kwd):
...
+        Updates permissions of a dataset.
+
+        :rtype:     dict
+        :returns:   dictionary containing new permissions
+        """
+        hda_ldda = kwargs.get('hda_ldda', 'hda')
+        try:
+            hda = self.get_hda_or_ldda(trans, hda_ldda=hda_ldda, dataset_id=dataset_id)
+        except Exception as e:
+            return str(e)
```
jmchilton: This is not correct, better just to let the exception bubble up in an API controller and let the _future_expose_api framework exception handling code deal with it.

hexylena: Ah, ok, didn't know future_expose_api did that, awesome. That will save some boilerplate. Should I do that for *all* exceptions or just the ones where there is not some more informative exception message to return?

jmchilton:
I'm writing some tests for this and trying to polish it a bit when I hit this:

```
  File "/Users/john/workspace/galaxy/lib/galaxy/web/framework/decorators.py", line 281, in decorator
  File "/Users/john/workspace/galaxy/lib/galaxy/webapps/galaxy/api/datasets.py", line 136, in set_permissions
    return self.hda_serializer.serialize_to_view(hda, view='detailed', user=trans.user, trans=trans)['permissions']
KeyError: 'permissions'
```

I'll open a PR or push the changes if I figure out what I'm doing wrong.

hexylena:
Huhh. That's interesting. (Also, thank you *so* much for writing tests, sorry, clearly should've done that.)

I was mostly being lazy and trying to obtain permissions, maybe there is a better way to do that? I wanted to avoid returning whatever the user sent and instead returning the actual effective permissions since that's a lot more useful. Not sure.

jmchilton:
Some thoughts:

- The tests aren't going well because the tests don't create datasets with an initial manage permission. So the test user cannot manage its own datasets.
- The permissions key error was a problem because I had added in a check that would allow admins to modify permissions - but admins don't have permission to serialize the HDA fully.

Neither of these things are problems that should prevent this from merging I guess, but I'd really like to get this API working for writing security tests - of which we have nearly none.

What I would like to see changed is the endpoint. Playing around with this - it is pretty clear this doesn't work with datasets in general - only HDA datasets. So I'd change the endpoint from:

```
POST /api/datasets/{encoded_dataset_id}/set_permissions
```

to

```
POST /api/histories/{history_id}/contents/datasets/{dataset_id}/permissions
```

I'd also merge the parameter handling with ``update_permissions`` in the library controller mentioned by @martenson. Maybe something that would be backward compatible with the awkward library syntax but still also work well for this:

```
util.listify(kwd.get("access_ids", None) or kwd.get('access_ids[]', None))
util.listify(kwd.get("manage_ids", None) or kwd.get('manage_ids[]', None))
util.listify(kwd.get("modify_ids", None) or kwd.get('modify_ids[]', None))  # only in library
```

I'll let you know if I make any more progress with the tests - I just wanted to lay out my initial concerns.

martenson:
>awkward library syntax 

🤣 sorry mate, this is one of the first API controllers I wrote for Galaxy

jmchilton:
> 🤣 sorry mate, this is one of the first API controllers I wrote for Galaxy

I wasn't trying to critique it - I had assumed there was some good legacy or form interface reason to have it that way. 

I made some progress on getting the tests to work - though I'm still having odd issues with the permissions key stuff and getting a 200 status code when I think it should be 200 - but the behavior change is at least detected.

https://github.com/jmchilton/galaxy/commit/68b8ac079a4e742c44b1485c082769b4f3d39bf9

hexylena:
Closed in favour of https://github.com/galaxyproject/galaxy/pull/6568

PR #6466, opened 2018-07-05
Let data managers also use conda python.

rhpvorderman:
Since data managers only return a json object, there is no need for them to use the galaxy python.


lib/galaxy/tools/__init__.py
```diff
@@ -508,7 +508,7 @@ def requires_galaxy_python_environment(self):
         """Indicates this tool's runtime requires Galaxy's Python environment."""
         # All special tool types (data source, history import/export, etc...)
         # seem to require Galaxy's Python.
-        if self.tool_type != "default":
+        if self.tool_type not in ["default","manage_data"]:
```
nsoranzo: Can you add a space after the comma for PEP-8 style?

mvdbeek (approved):

PR #6467, opened 2018-07-05
Make upload tool python3 compatible

mvdbeek:
I couldn't get make update_dependencies to run so that still needs to be done.

xref. #1715


lib/galaxy/web/framework/base.py
```diff
@@ -346,7 +347,8 @@ def read_lines(self):
             self.read_lines_to_eof()
 
 
-cgi.FieldStorage = FieldStorage
+webob.compat.cgi_FieldStorage = FieldStorage
+webob.request.cgi_FieldStorage = FieldStorage
```
nsoranzo: I don't think `webob.request.cgi_FieldStorage = FieldStorage` is needed:
```
>>> from webob.compat import cgi_FieldStorage as cgi_FieldStorage_compat
>>> from webob.request import cgi_FieldStorage as cgi_FieldStorage_request
>>> cgi_FieldStorage_compat == cgi_FieldStorage_request
True
>>> cgi_FieldStorage_request.__module__
'webob.compat'
```

nsoranzo:
There are 2 more instances of `cgi.FieldStorage`, if you want to change them too:
- `config/plugins/webhooks/demo/tour_generator/__init__.py`
- `lib/galaxy/webapps/tool_shed/api/repositories.py`

lib/galaxy/web/framework/base.py
mvdbeek: It was needed when I tested it (this took me quite a while to figure out). I think this is because we import `webob.exc` which import from `webob.request` and  `cgi_FieldStorage` is imported in the [`webob.request` module](https://github.com/Pylons/webob/blob/master/src/webob/request.py#L38), and if not patching this explicitly is being used from the local scope in various places in https://github.com/Pylons/webob/blob/master/src/webob/request.py#L223. Does that make sense ?

mvdbeek: You can test this by removing `webob.request.cgi_FieldStorage = FieldStorage` and running `sh run_tests.sh -api test/api/test_dataset_collections.py:DatasetCollectionApiTestCase.test_upload_collection`

nsoranzo: I see! Then I think the best way to proceed would be to patch the `cgi_FieldStorage` class, instead of subclassing it and then replacing it, see https://tryolabs.com/blog/2013/07/05/run-time-method-patching-python/

I'll do that, test, and push to your branch, shout if you think that's not good!

mvdbeek:
Perfectly fine, thanks @nsoranzo

nsoranzo:
Rebased to fix a merge conflict.

nsoranzo (approved):

jmchilton:
Wonderful, thanks a bunch!

PR #6468, opened 2018-07-05
Allow including paths in inputs json files.

jmchilton:
See https://github.com/qiime2/q2galaxy/issues/1#issuecomment-402419124 for context.


hexylena:
Omg! Yes! I've wanted this forever for the JBrowse tool.

jmchilton:
Thanks for the testing all and thanks for the fix @blankenberg!

PR #6478, opened 2018-07-09
GRT: fix the upload logging init

martenson:
- properly initialiaze the logger
- change level for some of the logs
- add some debug and critical logs

### DIFF
--- a/scripts/grt/grt.yml.sample
+++ b/scripts/grt/grt.yml.sample
@@ -1,7 +1,7 @@
 grt:
     # Register at https://telescope.galaxyproject.org to obtain an Instance ID and API key
-    instance_id: 
-    api_key: 
+    instance_id:
+    api_key:
 
     # Galaxy Project offers a public galactic-radio-telescope instance, however
     # you are free to run your own if you need. We would love it if you were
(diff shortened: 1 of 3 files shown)


martenson:
added a commit to expand default sanitation, I think data source tools are safe to ignore

scripts/grt/grt.yml.sample
```diff
@@ -21,6 +21,23 @@ sanitization:
     tools:
         - __SET_METADATA__
         - upload1
+        - ucsc_table_direct1
```
bgruening: @martenson I find these information super useful, any reason to blacklist those?
https://stats.usegalaxy.eu/d/SDduH5Zik/grt-tool-usage?refresh=15m&orgId=1&var-tool_id=ucsc_table_direct1&var-instance_id=All&var-date_interval=30d

scripts/grt/upload.py
```diff
@@ -24,14 +24,16 @@ def main(argv):
                         default=default_config)
     parser.add_argument("-l", "--loglevel", choices=['debug', 'info', 'warning', 'error', 'critical'],
                         help="Set the logging level", default='warning')
+
     args = parser.parse_args()
+    logging.getLogger().setLevel(getattr(logging, args.loglevel.upper()))
```
hexylena, line 29: Oh. Duh. Thanks.

hexylena (approved):
Could you also make the default for max-records in [export.py](https://github.com/martenson/galaxy/blob/390aedf8c8c28db739a78dc7c41e9275a735369b/scripts/grt/export.py#L138) something like 5M?

scripts/grt/grt.yml.sample
martenson: Hmm, I guess the execution numbers could be useful. I was more thinking in terms of parameters, runtime, metrics etc. which are nonexistent/useless. I will revert this.

martenson:
@erasche set to 5M
@bgruening reverted the sanitation

thanks for review!

Issue #5033, opened 2017-11-17
Feature Request: List mulled environment in Job Dependencies section


hexylena:
Trying to debug some failed jobs and I don't know which mulled environment corresponds to a set of dependencies. Some tools (e.g. bowtie) show this in the tool_version information. 

![utvalg_318](https://user-images.githubusercontent.com/458683/32942031-ac05a128-cb7f-11e7-9db8-e2717ef25976.png)

Others do not show it at all. I'd love something like:

![utvalg_319](https://user-images.githubusercontent.com/458683/32942025-a6ec58e4-cb7f-11e7-851d-c074a8793c5e.png)

mvdbeek:
We could predict the environment that would be used, or we could record this with the JobParameters in the database (other ideas?). I'm not sure what would be best. You can also see which environment is being used in the admin panel, under Manage tool dependencies, but that's a lot of clicks from a failed job ...

bgruening:
I fear that the reverse lookup will not be possible. But we can provide the mulled encoding for an arbitrary set of requirements. Code for this should be in galaxy-lib.

mvdbeek:
Yes, you can't go from mulled env to individual requirements without a reversible hash. But the job records the tool, the tool record the requirements, and from there you can get the mulled environment. Which is what we already do in the Manage tool dependencies page.

nsoranzo:
As you point out above, that may be different with respect to what was used when the job was actually run (e.g. if the dependencies were installed later).

mvdbeek:
Right, but the initial question `which mulled environment corresponds to a set of dependencies`, we can answer unambiguously (as long as the combination of requirements is still present in one of the tools).

It would also be great to either record the mulled environment, or even to record the contents of the mulled environment, which we could collect together with the job metadata.

Issue #5134, opened 2017-12-06
Enhance Job Metrics Framework and Integration


jmchilton:
Two job metrics Trello cards existed and had good points that never seemed to make it to Github.

https://trello.com/c/uAcQYz5I/1606-job-metrics-reports-and-interpretation covered integration of job metrics with the reports app.
- [ ] Add access to job metric information for particular jobs to the reports interface.
- [ ] Add access to aggregate information about runtime to the reports interface (might now overlap with GRT - we should find a way to make this all work together in harmony).
- [ ] Figure out the best way to express cluster utilization breakdown from collected job metrics.

https://trello.com/c/XsQdqliU/1607-job-metrics-technical-enhancements
- [ ] Add API for setting metrics (getting is now possible).
- [ ] Add plugin for slurm data collection (formatter only).
- [ ] Create infrastructure for cron-based dumping of slurm accounting logs into Galaxy (use API end-point).

Newer Ideas:
- [ ] Reproducibility metrics from also recorded in #4191
  - [ ] Record the full list of packages (including versions and sources) used in Conda for a given job.
  - [ ] Record every container identifier and some sort of hash (Docker or Singularity).
- [ ] Get collectl running on usegalaxy.org and figure out what it takes to make it useful.

hexylena:
Maybe to add to this: custom site-specific metric plugins loaded from a directory?

We have some extra facets of information that we could collect that aren't something that would be generally useful. Specific example, we store a 'build tag' in /etc/vgcn-release (indicating the version of the VM that the job is running on) that we could collect. It might be useful to have this tagged to the job for debugging purposes

nsoranzo:
@erasche For your specific example, you can probably use the `<env />` plugin if you export that information as an environment variable.

hexylena:
@nsoranzo huh, interesting idea. I'll test that! Thanks

hexylena:
Ah, bit by an old 'bug'.

Looks like that's not an option in my case, htcondor cleans the environment before running a job. It's something we've noticed earlier, and causes problems for us in other cases as well, HTCondor overrides TMPDIR settings leading to us [manually patching the upload tool](https://github.com/usegalaxy-eu/galaxy/commit/07f2aec016bf9eec603a0c2b1d9b6a22f7cf0c3b#diff-00554a62c3dc28eed1126a6a2d2c9114). Please excuse the exasperated commit message.

example of the environment a condor job gets by by default (though galaxy does set the option to pass through its environment)

[...]

Issue #4191, opened 2017-06-14
Improve Dependency Reproducibility Robustness


jmchilton:
In four sections going from "we are asking for disaster if don't do this" to "wow, that would be pretty cool".

For the community on depot for we should back up all:

 - [ ] Conda artifacts used by known tools. https://github.com/maxpoint/conda-mirror
 - [ ] Quay.io containers under the biocontainers namespace.
 - [ ] Singularity containers matching the biocontainers.

On Galaxy instances we should record:

- [ ] Record the full list of packages (including versions and sources) used in Conda for a given job.
- [ ] Record every container identifier and some sort of hash.
- [ ] Provide instructions and potentially infrastructure for local Docker repositories to store and serve containers.

On Galaxy instances we should provide the option to backup:

- [ ] All Conda packages used for every job.
- [ ] All containers used for every job.

On Galaxy instances we should provide APIs for:

- [ ] Fetching commands and package lists for rebuilding Conda environments for given jobs outside of Galaxy.
- [ ] Downloading the container used for a given job.
- [ ] Build containers with packages matching those used natively for non-container jobs.

PR #6480, opened 2018-07-10
Add more information to the dataset template

hexylena:
Add more information to the `show_params` template to help debug.

We still really need https://github.com/galaxyproject/galaxy/issues/5033 for debugging but there is unresolved discussion there.

### DIFF
--- a/templates/show_params.mako
+++ b/templates/show_params.mako
@@ -276,12 +276,10 @@ ${ job.command_line | h }</pre>
     <table class="tabletip">
         <thead>
         <tr>
-            <th>Key</th>
-            <th>Value</th>
+            <th>Runner</th><th>${ job.job_runner_name }</th>
         </tr>
         </thead>
         <tbody>
-            <tr><td>Runner</td><td>${ job.job_runner_name }</td></tr>
             <tr><td>Runner Job ID</td><td>${ job.job_runner_external_id }</td></tr>
             <tr><td>Handler</td><td>${ job.handler }</td></tr>
             %if job.destination_params:
@@ -297,7 +295,6 @@ ${ job.command_line | h }</pre>
                 </tr>
             %endfor
             %endif
-
         </tbody>
     </table>
 %endif
--- a/templates/show_params.mako
+++ b/templates/show_params.mako
@@ -274,17 +274,13 @@ ${ job.command_line | h }</pre>
 %if trans.user_is_admin():
 <h3>Destination Parameters</h3>
     <table class="tabletip">
-        <thead>
-        <tr>
-            <th>Runner</th><th>${ job.job_runner_name }</th>
-        </tr>
-        </thead>
         <tbody>
-            <tr><td>Runner Job ID</td><td>${ job.job_runner_external_id }</td></tr>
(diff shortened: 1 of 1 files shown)


hexylena:
Huh, checked a random job and mulled paths did show up.

auswahl_115

In that case maybe this works.

templates/show_params.mako
```diff
@@ -271,6 +271,37 @@ ${ job.command_line | h }</pre>
...
+                        %if str(k) in ('nativeSpecification', 'rank', 'requirements'):
+                        <pre style="white-space: pre-wrap; word-wrap: break-word;">${ v | h }</pre>
+                        %else:
+                        ${ v | h }
+                        %endif
+                    </td>
+                </tr>
+            %endfor
+            %endif
+
```
nsoranzo: Remove empty line

templates/show_params.mako
```diff
@@ -271,6 +271,37 @@ ${ job.command_line | h }</pre>
...
+<h3>Destination Parameters</h3>
+    <table class="tabletip">
+        <thead>
+        <tr>
+            <th>Key</th>
+            <th>Value</th>
+        </tr>
+        </thead>
+        <tbody>
+            <tr><td>Runner</td><td>${ job.job_runner_name }</td></tr>
```
nsoranzo: What about using `<th>` for `Runner`, `Runner Job ID`, ... and drop `Key`, `Value` above?

hexylena: Yeah that could work!

nsoranzo: Thanks for considering, I think I didn't make myself very clear, I've pushed a commit to your branch, which is probably easier. Feel free to remove it if you don't like it.

hexylena: Oh, interesting, I had never seen that construct before. Looks fine to me!

jmchilton (approved):

jmchilton:
This appears to be super useful! I really like this. Thanks.

templates/show_params.mako
```diff
@@ -271,6 +271,30 @@ ${ job.command_line | h }</pre>
...
+        <tbody>
+            <tr><th scope="row">Runner</th><td>${ job.job_runner_name }</td></tr>
+            <tr><th scope="row">Runner Job ID</th><td>${ job.job_runner_external_id }</td></tr>
+            <tr><th scope="row">Handler</th><td>${ job.handler }</td></tr>
+            %if job.destination_params:
+            %for (k, v) in job.destination_params.items():
+                <tr><th scope="row">${ k | h }</th>
+                    <td>
+                        %if str(k) in ('nativeSpecification', 'rank', 'requirements'):
+                        <pre style="white-space: pre-wrap; word-wrap: break-word;">${ v | h }</pre>
```
bgruening, line 286: @jmchilton the default-singularity container is reported but not the actually used Singularity container. Any idea why this is the case?

jmchilton: Yeah - I don't think we actually record that unfortunately. We have a couple different issues tracking that deficit: https://github.com/galaxyproject/galaxy/issues/4191 (bullet 2.2) / #5134 (bullet 3.1.2).

PR #6481, opened 2018-07-10
Update common_startup_functions.sh

nagoue:
Not to override tool_shed config file.

### DIFF
--- a/scripts/common_startup_functions.sh
+++ b/scripts/common_startup_functions.sh
@@ -149,7 +149,7 @@ find_server() {
             exit 1
         fi
         [ "$server_config" != "none" ] && arg_getter_args="-c \"$server_config\""
-        [ -n "$server_app" ] && arg_getter_args="--app $server_app"
+        [ -n "$server_app" ] && arg_getter_args="$arg_getter_args --app $server_app"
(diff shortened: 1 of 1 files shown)


martenson:
@nagoue could you please describe in a bit more depth what are you fixing here?

nsoranzo:
Code originally introduced in commit 09feaf6f4d46509c7401970013b5a8097a5d4c4e

nsoranzo:
@galaxybot test this

nagoue:
Hi,

I encountered a problem when starting running a local toolshed on our server with a specified IP instead of the localhost name. The fact is that our tool_shed.yml config file was not considered. Previously in the code:
l151: arg_getter_args returned the correct value
l152:  arg_getter_args returned default value crushing the previous one

I am currently using Galaxy 18.01.

nsoranzo (approved):

jmchilton (approved):

jmchilton:
@nsoranzo Nate is out this week I think, so I'm just going to merge it. I think it makes a lot of sense. Thanks so much for the contribution @nagoue!

martenson:
@nagoue I see. Thanks for the fix!

PR #6486, opened 2018-07-10
Fix for show params server error

blankenberg:
for https://github.com/galaxyproject/galaxy/pull/6480


hexylena:
Thanks. I didn't have the else when I originally wrote it and added it in because someone would've surely asked for it. Sorry, should've tested that change too.

blankenberg:
No worries, its always these last minute 'Someone is gonna ask for this, so I better add it' things that get me too. You even had a sweet screenshot and everything in the original PR. 

We should blame the lack of a previously existing test for the minor inconvenience.

martenson:
Or lack of mako syntax check tests.

PR #6489, opened 2018-07-11
Use bootstrap vue classes for workflow import form

guerler:


static/style/blue/base.css
```diff
@@ -1039,8 +1039,7 @@ pre {
   margin-bottom: 1rem;
   background-color: transparent; }
   .table th, .ui-table-striped th,
-  .table td,
-  .ui-table-striped td {
+  .table td, .ui-table-striped td {
```
martenson: is this whole file here by accident?

guerler: Yes, thanks for catching this. I am not sure why this happens but it seems that make client listener might be running a different version.

jmchilton:
Pushed fixes for the Selenium tests broken with these changes, hope that is okay.

jmchilton (approved):

PR #6490, opened 2018-07-12
Small improvement to the tool xsd docs regarding ``discovered_dataset``

bernt-matthias:
With hindsight this seems clear .. but it cost me some hours to find this out... :)

### DIFF
--- a/lib/galaxy/tools/xsd/galaxy.xsd
+++ b/lib/galaxy/tools/xsd/galaxy.xsd
@@ -1358,6 +1358,8 @@ provides a demonstration of using this tag.
 </test>
 ```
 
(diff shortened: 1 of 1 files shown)


lib/galaxy/tools/xsd/galaxy.xsd
```diff
@@ -1358,6 +1358,8 @@ provides a demonstration of using this tag.
 </test>
```
nsoranzo, line 1363: Maybe you can also add the `<outputs>` section of the tool to the above snippet.

bernt-matthias: good idea

jmchilton (approved):

nsoranzo (approved):

PR #5462, opened 2018-02-06
Tool that adds/sets tags for collection elements from a file

mvdbeek:
This tool can set or add tags from a file. I wrote this for https://github.com/galaxyproject/galaxy/pull/5457, but I think it may be useful for other applications as well.

PR #5457, opened 2018-02-04
Allow selecting subset of collection by group tag

mvdbeek:
This would allow selecting tagged subsets of (nested) collections. To use this you need to tag the individual collection elements with `group:<some group name>`. All groups defined this way will be selectable in the tool form.

Together with [this patch against the deseq2 wrapper](https://gist.github.com/mvdbeek/195522e708ede83debf147384c756604) one can setup a deseq2 experiment from an abitrarily nested collection like this:
screen shot 2018-02-04 at 12 02 02

That's a fairly simple way to do this, I'd be curious to know if people think that something like this could work out ?
xref #740

PR #5403, opened 2018-01-27
[18.01] More robust nametag discovery and propagation.

jmchilton:
- Nametags weren't being discovered on collection inputs - so for instance tools that take in a single collection and produce a dataset or file weren't propagating nametags (unless the collection contents had tags). To see this create two unnamed datasets without tags, create a pair from them, add a tag to the pair, and run the pair through the "paired_test" test tool.
- Nametags weren't being propagated to the outputs of collection operations - to either dataset or collection outputs.

This PR fixes both problems I think.

Ping @nekrut

PR #3979, opened 2017-04-26
Hashtags

dannon:
Purely clientside implementation of hashtag shortcut for nametags.  These are translated to 'name:nametag_value' prior to saving and are persisted in that form.

PR #6491, opened 2018-07-12
Implement ``group:`` tags (toward multi-factor analysis with group tagging)

jmchilton:
See #5457 for an example of how tool authors may consume such tags and how users might define them. This PR simply makes sure they will auto-propagate through an analysis the way ``name:`` tags do.

xref #3979 
xref #5462
xref Many long conversations about the future of multi-factor analyses in Galaxy at the GCC that we didn't take any notes or create any issues out of.

Adds an API test case to verify both name tagging and group tagging auto propagate for simple tools (more tests could be added #5403).


nsoranzo (approved):

mvdbeek (commented):
Wonderful, that looks great!

Issue #6492, opened 2018-07-12
ftp_upload_purge flag in galaxy.yml is broken


Takadonet:
I am attempting to get the desire behavior where files that are uploaded from the ftp_dir are then cleaned up from the directory. In release 17.09 that was the default behavior but not so in 18.05 .

I have set [ftp-upload-purge](https://docs.galaxyproject.org/en/latest/admin/options.html?highlight=ftp_upload_purge#ftp-upload-purge) to True in galaxy.yml and still does not clean uploaded ftp files.

After taking a look at the [upload.py](https://github.com/galaxyproject/galaxy/blob/release_18.05/tools/data_source/upload.py#L87) , it seems that is happening  ftp_upload_purge from galaxy.yml is not being used when determining if we are setting the in_place flag to True or false. 
When in_place is set to False, later on in [upload](https://docs.galaxyproject.org/en/latest/_modules/galaxy/datatypes/sniff.html#handle_uploaded_dataset_file_internal) , it will create a new temporary converted_path file instead of using the original one. That temporary file will be removed but not the original.

Takadonet:
May look like that we have to modify this [line ](https://github.com/galaxyproject/galaxy/blob/release_18.05/tools/data_source/upload.py#L144)

`shutil.move(converted_path or dataset.path, output_path)`

because need to remove both converted_path **and** dataset.path when we have ftp-upload-purge set

Takadonet:
Thanks for merging!

Question, is there any plans for back porting this regression into release_18.05 branch?

martenson:
@Takadonet I will open the backport shortly

Takadonet:
awesome! 
Thanks again.

PR #6494, opened 2018-07-12
Fix ftp_upload_purge in galaxy.yml

Takadonet:
Pull request to fix issue #6492 where uploaded ftp files were not being purged even when being set in galaxy.yml with flag [ftp_upload_purge](https://docs.galaxyproject.org/en/latest/admin/options.html?highlight=ftp_upload_purge#ftp-upload-purge) .

New code  covers case where both purge_source and converted_path are set to True hence need to move the converted_path to 'real' path AND remove the original dataset from FTP dir.

### DIFF
--- a/tools/data_source/upload.py
+++ b/tools/data_source/upload.py
@@ -141,7 +141,12 @@ def add_file(dataset, registry, output_path):
         # Move the dataset to its "real" path. converted_path is a tempfile so we move it even if purge_source is False.
         if purge_source or converted_path:
             try:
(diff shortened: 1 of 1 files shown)


martenson (approved):
lgtm, thanks for the fix!

nsoranzo:
Does this need to be backported?

martenson:
@nsoranzo https://github.com/galaxyproject/galaxy/pull/6497

Issue #6202, opened 2018-05-23
Batch Upload API and UI - The Next Iteration


jmchilton:
- [x] Integrate tags in the upload 2.0 API and rule definition language. (in 18.09)
- [x] Integrate hashes in the upload 2.0 API and rule definition language.
- [ ] New "Archive" upload tab in the UI.
  - [ ] MVP: Direct upload or paste URL to a BagIt, Zip, or YAML description to create a collection of dataset instances using the upload 2.0 API.
  - [ ] Extension 1: Allow selecting FTP files.
  - [ ] Extension 2: Allow dumping contents to a data library folder.
  - [ ] Extension 3: Allow dumping contents to a new data library.
  - [ ] Extension 4: Dump contents into the rule builder as an optional second step after this first step is implemented.

neoformit:
@jmchilton Can we edit the MVP to include tar please?

Issue #5822, opened 2018-04-01
Rules Widget - Post Merge Smaller Tweaks


jmchilton:
- [X] Upload widget needs a  "Close" button for the Rule-based tab, same as on the other tabs (Regular/Composite/Collection). (Suggested by @mblue9 on #5365). (in https://github.com/galaxyproject/galaxy/pull/5823)
- [X] Upload widget needs reset button as well. (Suggested by @mblue9) (in #5823)
- [X] Instead of "Okay" button call it "Apply"  when editing rules / mapping. (Suggested by @mblue9) (#5823)
- [X] X next to mapping to remove mapping in edit mode (this is only present in the outer display of the mappings I think). (Suggested by @mblue9) (in #5823)
- [ ] Prevent "Build" until some data is present (reported by @martenson below).
- [x] Reconsider the formatting of rule definitions (namely alignment, select width) (reported by @martenson below).
- [ ] Only allow selecting text formats or tabular formats in the dataset selection list in the upload widget  (reported by @martenson below).
- [x] Make genome selector wider (reported by @martenson below).
- [ ] Better handling of the textfield as datatype of the upload widget changes (reported by @martenson below).
- [x] Allow name tags for collections:
   - [x] Textbox by name if just one collection specified, read from table like name if multiple collections.
- [x] Add drop target for text datasets. (Suggested by @mvdbeek.)
- [x] Allow reading in tags - name tags and normal tags - from rule data for individual datasets.
- [ ] New filter - contains value.
- [ ] New column adder - split on character. 
- [ ] Add checkbox requiring multiple collections to be matching - currently one can create two collections with the same structure but there is no way to force that they are the same structure and warn if they aren't. Would then implement #3916.
- [ ] Allow setting path for uploads (for admins).
- [x] Integrate with library import folder (nearly identical to FTP directory I think).
- [x] Scrolling is a mess #6100 (also reported on https://github.com/galaxyproject/galaxy/issues/6220).
- [ ] The last set of ruled uploadings might force the closure of rules currently being worked on (reported in https://github.com/galaxyproject/galaxy/issues/6220).
- [x] Add button and browser state to allow reloading last rules.

Some migrated from #5380.

martenson:
a few observations from latest dev

* we could add validation and do not allow to click `build` until some data is present
* when switching `Load tabular data from:` the textfields should be re-initialized. e.g. now when switching from value `FTP directory` the actual contents of FTP dir remain in the field
[...]

PR #6499, opened 2018-07-13
Allow setting tags on targets & contents in data fetch API.

jmchilton:
Backend to power better metadata collection when importing large amounts of structured data - with tests.

I'll follow up with rule builder enhancements that allow setting name tags on uploaded collections and datasets with specific support for group tags, name tags, and general purpose tags.

xref #6202
xref #5822
xref #6491
xref #5457
xref #3979
xref #5462
xref Many long conversations about the future of multi-factor analyses in Galaxy at the GCC that we didn't take any notes of or create any issues out of.

### DIFF
--- a/lib/galaxy/managers/collections.py
+++ b/lib/galaxy/managers/collections.py
@@ -46,11 +46,11 @@ class DatasetCollectionManager(object):
         self.tag_manager = tags.GalaxyTagManager(app.model.context)
         self.ldda_manager = lddas.LDDAManager(app)
 
-    def precreate_dataset_collection_instance(self, trans, parent, name, structure, implicit_inputs=None, implicit_output_name=None):
+    def precreate_dataset_collection_instance(self, trans, parent, name, structure, implicit_inputs=None, implicit_output_name=None, tags=None):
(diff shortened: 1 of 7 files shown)


lib/galaxy/managers/collections.py
```diff
@@ -46,11 +46,11 @@ def __init__(self, app):
         self.tag_manager = tags.GalaxyTagManager(app.model.context)
         self.ldda_manager = lddas.LDDAManager(app)
 
-    def precreate_dataset_collection_instance(self, trans, parent, name, structure, implicit_inputs=None, implicit_output_name=None):
+    def precreate_dataset_collection_instance(self, trans, parent, name, structure, implicit_inputs=None, implicit_output_name=None, tags=[]):
```
mvdbeek: I guess we'll want to avoid tags being mutable here.

jmchilton: Opps... definitely. Thanks for the catch.

jmchilton:
Thanks for the typo fixes @dannon and the merge!

PR #6500, opened 2018-07-13
Integrate tagging into the rule builder.

jmchilton:
- Allow specifying name tags, group tags (added in https://github.com/galaxyproject/galaxy/pull/6491), and "general" or generic tags on stand-alone datasets and collection content datasets.
- Allow reusing the collection name as a name tag during upload for the whole collection. In this context I think the name tag is especially useful and will typically be the same name as the collection so the UI is kept simple by making it a checkbox.

Ideas tracked on https://github.com/galaxyproject/galaxy/issues/5822.

Builds on #6499.


dannon:
+1

PR #6502, opened 2018-07-14
Fix bam to bigwig error detection.

pvanheus:
As per the [documentation](https://docs.galaxyproject.org/en/master/dev/schema.html#tool-command) leaving the tool without `detect_errors` and without a `profile` set to greater than or equal to `16.04` invokes the old behaviour where any output to `stderr` denotes failure. When provided with a BAM format input file and the `-g` option, `bedtools` outputs a warning to `stderr` stating that it will ignore the provided genome, triggering a tool failure.


bernt-matthias:
I guess this should be handled by the tool using the `<stdio>` tag functionality.

jmchilton:
Updated the title to yield better release notes, thanks for the fix @pvanheus !

pvanheus:
@bernt-matthias [best practice](http://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#exit-code-detection) recommends using `detect_errors` rather than `<stdio>`.

bernt-matthias:
OK. Cool. Got it. Still need time to remember all this :)

jmchilton:
Thanks a bunch!

PR #6509, opened 2018-07-17
[WIP] Explicitly free memory in sprawling tool action execute method.

jmchilton:
Should give Python the freedom to garbage collect more data more quickly and there are scenarios where a lot of time is spent in this execute method.


jmchilton:
@nsoranzo I'm not sure this is a good idea, but if you wanted to profile it and see if it helps I'm not opposed to doing something like this either.

PR #6510, opened 2018-07-17
[18.05] Fix infinite loop in uploads

nsoranzo:
If the upload get somehow interrupted and an empty file is created in `new_file_path`, `upload.py` is stuck in an infinite 100%-CPU loop due to the following `Ms2.sniff_prefix()` bug.


dannon:
xref #6418 for the original (dev) PR containing this fix.

jmchilton (approved):

PR #6511, opened 2018-07-17
GNU sort based sorter 

bernt-matthias:
GNU sort is much more efficient than the python based version. also gives more possibilities. xml (repeat) has been simplifies

- [ ] fix header
- [ ] is there a conda package with the parallel option? do we want a conda dependency here?


bgruening:
Before we had conda we agreed to keep this tool as it is and not introduce external dependencies. Maybe this has changed. But we came up with this tool: https://github.com/bgruening/galaxytools/blob/master/tools/text_processing/text_processing/sort.xml and associated tools.

bernt-matthias:
This is exactly what I was looking for .. and has already all points solved that are open in my PR.  Problem was that these nice tools did not pop up in the toolshed when searching for "sort".

So I guess we can close this one. But we could think about replacing/merging some of the text processing tools in the galaxy core by their more efficient counterparts.

bgruening:
@bernt-matthias these tools are very old and predates the multiplexing-functionality of planemo. Maybe we can at some point create separate repos for every tools and keep tool-linage intact.

mblue9:
If these tools are very old and there are better options, then why can't the old tools just be deprecated? 

Having multiple similar tools causes confusion for admins and users. Personally I wish there was only 1 Sort tool, 1 Cut tool etc...

bernt-matthias:
Yep. So how about replacing/merging (carefully .. maintaining compatibility) these tools. Would this be an option?

PR #6512, opened 2018-07-18
Remove user create mixin

guerler:
The Toolshed login mechanism is interwined with Galaxy but uses its own User model and several custom functions and if/else statements to handle differences. In order to safely remove the create user mixin and use the manager instead the Toolshed code has to be separated from Galaxy.


jmchilton:
This introduces 700 lines of duplicated code to eliminate a few conditional branches? I'm real skeptical that eliminating the maintenance burden of having those conditionals is worth the maintenance burden of duplicating all that code for the tool shed. Is there a larger context about why eliminating this mixin is so important that can help sway me this is a good idea?

guerler:
Yes there is a larger context. I want to move Galaxy's registration related code to the auth manager and then revise it such that it works in an api-like fashion for vue components. This Mixin is just at the very bottom of the code path. We cannot remove the makos and the backend-redirects in Galaxy without separating the too. Instead of separating the two entities one function at a time I decided to do it in a single PR and opened it separately to discuss the concerns you are raising. I tried to separate the two in a more piece meal fashion but there are too many interdependencies i.e. OpenID, Toolshed, Galaxy are interwined with advanced login features. Thanks for the review and the comments @jmchilton.

guerler:
Closing in favor of #6413.

PR #6517, opened 2018-07-19
Return a json blob containing redirect URI in OIDC login controller 

VJalili:
The `login` function of the OIDC controller redirects a user to authentication end-point of an IdP. Due to recent re-architecture of the login mako, the `login` function is returns a json object containing the redirect uri.


VJalili:
Thanks @dannon for merging this.

PR #6519, opened 2018-07-20
Improvements to Cloud Storage Manager

VJalili:
To copy an object from cloud, the cloud manager currently stages the object first, then pass the staged object to the upload tool. Hence, the manager's copy function would return only after a given object is staged (downloaded). This design can cause request timeout when copying big files from a cloud-based storage to Galaxy. 

This PR solves that issue by passing a _singed url_ of the object to copied from cloud, to upload tool. Hence, the copy process is handled as a job, avoiding request timeout.


VJalili:
@afgane Thanks for merging this.

PR #6520, opened 2018-07-20
Fix failing qunit tests in dev.

jmchilton:


jmchilton:
Broken with https://github.com/galaxyproject/galaxy/pull/6447 I assume.

PR #6521, opened 2018-07-20
Rebuild CSS against latest dev

jmchilton:
If we're going to keep this file in the tree it should be updated against current sources then I believe right? Otherwise it is sort of a constant nuisance.

### DIFF
--- a/static/style/blue/base.css
+++ b/static/style/blue/base.css
@@ -1039,8 +1039,7 @@ pre {
   margin-bottom: 1rem;
   background-color: transparent; }
   .table th, .ui-table-striped th,
-  .table td,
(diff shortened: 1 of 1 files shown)


nsoranzo:
Not sure it still makes sense to treat this file differently than the rest of the client.

dannon:
@nsoranzo It doesn't other than that there are a lot of extra artifacts (sprite sheets, images, etc.) that go into the style build, so it isn't as easy to remove from the tree.

That said, the webpack 4 branch includes webpack-based style builds and will eliminate the need for this.

jmchilton:
Thanks for the merge @dannon!

PR #6524, opened 2018-07-21
Use cached database identity to avoid many DB requests when creating jobs.

jmchilton:
When creating thousands of jobs for large database transactions - previously Galaxy would query the database thousands of times for the same history id, user id, and session id. Extensive comments in code and new test case asserting the sqlalchemy behavior described therein.

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -160,6 +160,32 @@ class UsesCreateAndUpdateTime(object):
         return (galaxy.model.orm.now.now() - create_time).total_seconds()
 
 
+def cached_id(galaxy_model_object):
+    """Get model object id attribute without a firing a database query.
+
+    Useful to fetching the id of a typical Galaxy model object after a flush,
+    where SA is going to mark the id attribute as unloaded but we know the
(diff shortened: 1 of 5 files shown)


lib/galaxy/model/__init__.py
```diff
@@ -160,6 +160,28 @@ def seconds_since_created(self):
...
+
+    Useful to fetching the id of a typical Galaxy model object after a flush,
+    where SA is going to mark the id attribute as unloaded but we know the
+    id is immutable and so we can use the database identity to fetch.
+
+    With Galaxy's default SA initialization - any flush marks all attributes
+    as unloaded - even objects completely unrelated to the flushed changes
+    and even attributes we know to be immutable like id. See test_galaxy_mapping.py
+    for verification of this behavior. This method is a workaround that
+    uses the fact we know all Galaxy objects use the id attribute as identity
```
bgruening: Is here a `that` missing?

bgruening (commented):
Insane what you can learn from a PR. Thanks John!

jxtx:
Nice.

PR #6528, opened 2018-07-23
[18.05] Fix warning in CONVERTER_bam_to_bigwig_0 and CONVERTER_sam_to_bigwig_0

nsoranzo:
Since v2.27.0 (see https://github.com/arq5x/bedtools2/commit/4d188435777cfbd6ada7394ee78fde37f07c9a75 ), `bedtools genomecov` emits this warning:
```
*****
*****WARNING: Genome (-g) files are ignored when BAM input is provided.
*****
```
if both `-ibam` and `-g` option are specified, which causes the converters to always fail.

Also:
- Add `detect_errors="aggressive"` to `<command/>` to prevent future warnings from creating other issues.
- Fix input format in `CONVERTER_sam_to_bigwig_0`.

### DIFF
--- a/lib/galaxy/datatypes/converters/bam_to_bigwig_converter.xml
+++ b/lib/galaxy/datatypes/converters/bam_to_bigwig_converter.xml
@@ -4,23 +4,22 @@
         <requirement type="package" version="357">ucsc-bedgraphtobigwig</requirement>
         <requirement type="package" version="2.27.1">bedtools</requirement>
     </requirements>
-    <command><![CDATA[
-        bedtools genomecov -bg -split -ibam '$input' -g '$chromInfo'
+    <command detect_errors="aggressive"><![CDATA[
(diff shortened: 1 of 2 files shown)


jmchilton:
xref #6502 for similar fix to dev from @pvanheus - we changed the version of the tool in that case - though I'm indifferent as to whether we do that or not.

nsoranzo:
I'd regard the changes to `CONVERTER_bam_to_bigwig_0` (partially fixed also by Peter) as bug fixes not modifying the tool inputs/outputs/requirements.

The input format change to `CONVERTER_sam_to_bigwig_0` instead deserves a version bump, even if the tool was broken since its introduction in 2013, so I've rebased the PR including this change.

dannon (approved):

jmchilton (approved):

PR #6531, opened 2018-07-24
Optimizations for history copying.

jmchilton:
Use only one unqualified flush to prevent constant repeated invalidation off all SA objects just to get a single ID for a new object.

Some random timings:

Before optimization:

10 datasets - with a MetadataFile metadata parameter
history copied (390.457 ms)
history copied (398.501 ms)
history copied (360.997 ms)

10 datasets  - without a MetadataFile metadata parameter
history copied (291.972 ms)
history copied (276.479 ms)
history copied (272.323 ms)

after optimization:

10 datasets - with a MetadataFile metadata parameter
history copied (230.171 ms)
history copied (218.215 ms)
history copied (226.205 ms)

10 datasets - without a MetadataFile metadata parameter
history copied (66.360 ms)
history copied (70.126 ms)
history copied (65.166 ms)

Also fixes HDCA annotations which seem to have been completely non-functional on the backend.

### DIFF
--- a/lib/galaxy/managers/hdcas.py
+++ b/lib/galaxy/managers/hdcas.py
@@ -34,7 +34,7 @@ class HDCAManager(
     foreign_key_name = 'history_dataset_collection_association'
 
     tag_assoc = model.HistoryDatasetCollectionTagAssociation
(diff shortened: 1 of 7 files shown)


dannon:
Nice!

Issue #4572, opened 2017-09-07
[RFC] DFARS / Restricted Data support


hexylena:
At $oldjob we discussed this a bit, and I've recently responded to an inquiry about the topic as well. Some universities are coming under DFARS regulations and have new requirements for data sanitization and access controls.

I think that these features can be built into Galaxy and allow those groups to meet DFARS requirements and also to continue using their local Galaxy instance that they're using for the rest of their data.

There are a couple major points:

Point | Resolution
--- | ---
Restricted data must be securely deleted | This can be accomplished by replacing most `os.unlink` calls with `subprocess.check_call(['srm', ...])` (and obviously requiring the admin to install this.)
The admins must be able to produce logs of who had access to data | This is a simple auditing requirement. It could be accomplished either via adding a table, or via just logging all changes to a backend store (file, append-only-log, some service like kafka, etc), and then processing these offline to generate any requested reports of "Alice granted Bob access at 19:00 UTC to these datasets"
Restricted data must be encrypted at rest | This one is a minor nightmare for most people. This could probably be accomplished via 1) implementing galaxy-hashicorp vault support (discussed in issue ...), 2) encrypting uploaded datasets before they hit disk using a key stored in vault, 3) issuing a token to the job runner (possibly pulsar could be strictly required) for 1 time access of data, downloading that data + decrypting it, processing, and then encrypting the result / sending key to hashicorp / securely wiping any files generated.

I mentioned above that "they could still use their existing galaxy" for which I propose the following:

- admins can say "all data in this galaxy should be treated as restricted," or "all data from these users / roles should be treated as restricted," or "users may decide what is restricted.
- at the user level, users can tag specific datasets as restricted. This would opt those datasets in, and permanently tag all derived artefacts as being restricted. (The admin would choose a tag to mean restricted, and that tag would not be able to be removed from datasets / histories / libraries / etc. If an admin wants to say "in my galaxy the tags restricted, secure, CUI, NOFORN, etc.." all mean secure, they can, and users can easily tag things. We would give special CSS rules to these tags.) Thanks to the awesome propagation of dataset tags, we can accomplish most of this.

After thinking about this a bit at $oldjob, I moved. My colleague there @moffmade is still stuck implementing some of these. At $newjob I've discussed it with @bgruening and he mentioned some other groups who might be interested in this sort of feature.

[...]

PR #6539, opened 2018-07-25
Convert tool parameter XML attribute 'hidden' into the correct boolean Python type

andreyto:
Previously it was converted into a string, and in the logical predicates
any non-empty value would always evaluate to True.
This fix allows using the 'hidden' attribute in parameterized XML
macros with 'true' or 'false' passed as a macros parameter. This
in turns allows hiding some form fields (after setting default values)
based on choices picked in conditional switches.

This would still require documenting the 'hidden' attribute in the Galaxy Tool XML Wiki.
The need for this feature is related to #4572. The idea here is that for some parameters
certain default values have been "validated" for certain choices of a "master" parameter switch
used in a conditional, and the validated parameters must be frozen for the user.

It would be even better to have an attribute 'disabled' that would make the form field visible
but read-only like many UI frameworks allow to do. That will require more work though.

### DIFF
--- a/lib/galaxy/tools/parameters/basic.py
+++ b/lib/galaxy/tools/parameters/basic.py
@@ -94,7 +94,7 @@ class ToolParameter(Dictifiable):
         self.argument = input_source.get("argument")
         self.name = self.__class__.parse_name(input_source)
         self.type = input_source.get("type")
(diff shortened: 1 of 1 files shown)


jmchilton:
Looks great to me, thanks for the contribution @andreyto!

PR #5819, opened 2018-03-30
Allow rule based operations on existing collections.

jmchilton:
Allow the rules DSL & GUI component (#5365) to operate on existing collections - this enables very flexible filtering, sorting, relabelling, grouping, flattening, and general re-organization of existing collections (e.g. the outputs of tools). This is implemented as a collection operation tool so that it should be executable interactively in the tool form and in a batch fashion as part of workflow executions.

For this to be tracked properly as a tool execution and to work properly in the tool form, I've implemented a new tool framework parameter type (including tool form integration) called "rules". Right now there is just one tool that uses this parameter type - but one can imagine other operations using the rule builder - e.g. creating collections from samplesheets as the first step in an automated workflow using the tool framework instead of the upload API.

This can thought of as a more GUI friendly alternative to some of my proposed collection operations (#2496 , #1313) in the past that consumed JavaScript expressions. This alternative approach is much easier to visualize for GUI based manipulation and is simpler than learning JavaScript if one does not know how to program, but the ubiquity of JavaScript would provide its owns benefits. Hopefully in the long run both approaches prove useful within the Galaxy ecosystem.

- [X] ~Outline the collection operation tool and parameter plumbing. Backend basics. (1 Day)~
- [X] ~Implement simplest tool form parameter. (1 Day)~
- [X] ~Connect tool form parameter to rule based builder component, including callback. (1 Day)~
- [X] ~Simplest rule working in API to create flat and nested collections. (1 Days)~
- [X] ~Simplest rule working GUI.~ (1/2 Day)
- [X] ~Preload collection identifiers as columns the first time the rules widget is loaded.~
- [x] ~Selenium test case for no-op tool. (1 Day)~
- [X] ~Flatten use case working API (1/2 Day)~
- [X] ~Flatten use case - GUI + Selenium. (1/2 Day)~
- [x] ~Swap the frontend to use Python-like regexes so these two approaches give the same results. (2 Days)~
- [x] ~Develop specification framework for rules DSL that can be used to test frontend JS and backend Python.~
- [x] Backend implementation of remaining rule types: (2 Days)
  - [x] add_column_basename
  - [x] add_column_rownum
  - [x] add_column_value
  - [x] add_column_metadata
  - [x] add_column_regex
  - [x] add_column_concatenate
  - [x] add_column_substr
  - [x] remove_columns
  - [x] add_filter_regex
  - [x] add_filter_count
  - [x] add_filter_empty
  - [x] add_filter_matches
  - [x] add_filter_compare
[...]

PR #6545, opened 2018-07-26
Allow Consuming Tags in the Apply Rules Tool

jmchilton:
A few small polish, bug fix commits and then two primary commits to allow using tags more effectively in the Apply Rules tool.

The previously merged PR https://github.com/galaxyproject/galaxy/pull/6500 allowed setting tags from metadata when importing new collections using the rule builder. 4d83746 included in this PR then lets dataset tags already present to be loaded in when modifying the collection with the "Apply Rules" tool (added in 18.05 as part of https://github.com/galaxyproject/galaxy/pull/5819). This allows filtering a collection based on tags and re-organizing the nesting structure of a collection based on tags.

18.09 will introduce a new kind of tags - group tags (added in https://github.com/galaxyproject/galaxy/pull/6491) that like name tags will be auto-propagated but won't be used for naming but instead of categorizing datasets in a factorial analysis. #6500 added support for populating datasets with such tags when importing them using the rule builder and #5457 will add support for tools to effectively consume these group tags. 5807230 included in this PR now lets the user very easily use these tags with the "Apply Rules" tool. This allows certain kinds of filtering and re-organization using group tags specifically that are technically possible after the previous commit be implemented by users in one step instead of several. For instance, if datasets in a collection have different group tags such as ``group:type:single`` and ``group:type:paired-end`` - a single rule can now be used to populate a column containing either ``single`` or ``paired-end``. This will likely be the most common tagging operation with the Apply Rules tool and would require 3 rules without this change.

### DIFF
--- a/client/galaxy/scripts/components/RuleCollectionBuilder.vue
+++ b/client/galaxy/scripts/components/RuleCollectionBuilder.vue
@@ -1182,6 +1182,7 @@ export default {
                         metadataOptions["identifier" + index] = _l("Paired Identifier");
                     }
                 }
+                metadataOptions["tags"] = _l("Tags");
             } else if (this.elementsType == "ftp") {
                 metadataOptions["path"] = _l("Path");
             } else if (this.elementsType == "library_datasets") {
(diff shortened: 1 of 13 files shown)


mvdbeek (approved):
Very nice, I like it!

Issue #4959, opened 2017-11-08
Increase Job Throughput


jmchilton:
With metadata handled as part of the remote process - Galaxy still struggles a bit when creating and finishing hundreds of jobs a time. I've spent a good amount of time working on this problem in 2016 as part of scaling up collections but I think I need to take some more passes. Here are some things I think should be done:

TODO:

- ``job_wrapper.change_state`` should just be one database statement execute - at least in some threads sometimes. There is probably no need to load multiple objects and do things in SA. (EDIT: done in #9722)
- Explore faster non-SA object batch insert methods - https://stackoverflow.com/questions/12095176/optimizing-batch-inserts-sqlite.
- Implement a number of pre-created threads to work on creating jobs for a single workflow thread - this is almost ready but we should implement the last few things, have test cases,  and document the entry point. https://github.com/galaxyproject/galaxy/issues/3903
- ``job.mark_deleted`` should be made an atomic database method when ``change_state`` above is - but I guess it isn't used in the main creation and finish threads.
- https://stackoverflow.com/questions/8134602/psycopg2-insert-multiple-rows-with-one-query

DONE:
- Pre-allocate all HIDs need for output implicit collections to eliminate a few database flushes in that process,
- Flush all collection mapping jobs at once. Last optimization left in https://github.com/galaxyproject/galaxy/pull/6563. Done in https://github.com/galaxyproject/galaxy/pull/9985
- https://github.com/galaxyproject/galaxy/pull/6367 (could be back ported to at least 18.01 safely I believe).
- Optimize object store interaction to not require so much DB flushing when creating jobs.
- Optimize checking permissions for large collections - keeping permissions objects in the DB and not loading.
- Optimize checking stage for large collections -  use newer dataset matchers.

FAILED PRS:

- https://github.com/galaxyproject/galaxy/pull/6348
- https://github.com/galaxyproject/galaxy/pull/6343
- https://github.com/galaxyproject/galaxy/pull/6342
- #1957

mvdbeek:
I'm gonna call this one complete (even though we didn't do batch inserts yet). I think the next step is to break down pieces into tasks we can schedule with celery, where we're not limited by the GIL. Job handlers are often close to 100% in CPU utilization, so I don't think threading is going to help much there. It'll also be much easier to scale celery tasks based on load, and they can be retried and resumed more easily.

Issue #4974, opened 2017-11-10
[RFC] Temporarily marking storage backends as "temporarily down"


hexylena:
We need to have a temporary downtime for one backend in our object store configuration that looks something like the following:

```xml
<object_store type="distributed" id="primary" order="0">
    <backends>
        <backend id="files1" type="disk" weight="10">
            <files_dir path="/data/1/" />
        </backend>
        <backend id="files2" type="disk" weight="0">
            <files_dir path="/data/2/" />
        </backend>
    </backends>
</object_store>
```

I'd like to propose a feature whereby we add an additional configuration option to the `<backend />` structure, such as `offline="true"` which behaves as follows:

- datasets are greyed out (like when permissions are wrong)
- the users cannot click "view"
- jobs cannot be launched on these datasets

bgruening:
Obviously I like this!

VJalili:
What happens to the shared data persisted on offline storages?

hexylena:
It is still there, just temporarily inaccessible to all users (and jobs can't be run on it, etc.) The idea is eventually the object store will be re-attached and then the datasets will be available again.

natefoo:
Could this be done dynamically, possibly with some sort of health check? Unless Galaxy can be made to reload the object store on signal (or... something else, with uWSGI).

hexylena:
If it could, a healthcheck sounds brilliant, better than forcing a restart to update it. I would not know how to implement it in a dynamic way though. There is I chance I have the ability to implement it in a static config-y way. (In case someone wants to give me some pointers.)

PR #6550, opened 2018-07-27
Configure object store for datasets in job handler instead of web thread.

### LINKED ISSUE #6513: Don't create job output datasets on disk in web threads
The job setup code in the job handler should be doing the makedirs and stuff but it appears to happen in job creation code in response to tool form submission. This means a slower user experience and also means disk slow downs could prevent jobs from being created and hinder user experience even further and probably hang web threads as well. After https://github.com/galaxyproject/galaxy/pull/6373 is merged, this functionality is called from ``ObjectStorePopulator.create_datasets()`` in ``galaxy.tools.actions`` but it really needs to be moved to the job handler I think (#6373 doesn't make this worse - it actually separates things out a little better and hopefully will allow us to improve this more easily).

xref #4959

jmchilton:
This is good for a few reasons:

* It speeds up the bottleneck that is the job creation code in the web thread (the job handler is much easier to parallelize). Skipping the flush for all those datasets and skipping the disk write per dataset should really speed things up nicely when submitting large numbers of jobs and jobs with a large number of inputs. Fixes https://github.com/galaxyproject/galaxy/issues/6513 and important progress on #4959.
* This prevents a disk slow down from grinding the web response to a halt. We don't want the handler to freeze up either - but if something is going to freeze it is a better user experience and easier to recover from the job handler freezing while their are disk problems.
* This allows per-job-handler object store configurations - simply configure different job handlers with different object store configurations. This is also a huge step toward per-job-destination object stores (see #6552 that builds on this PR to add that feature).

Updated: Rebased with some fixes and removed the refactoring toward per-job-destination commit since it has been merged into a single commit that actually adds that feature as part of #6552.

### DIFF
--- a/lib/galaxy/objectstore/__init__.py
+++ b/lib/galaxy/objectstore/__init__.py
@@ -808,3 +808,24 @@ def _create_object_in_session(obj):
         object_session(obj).flush()
     else:
         raise Exception(NO_SESSION_ERROR_MESSAGE)
+
+
+class ObjectStorePopulator(object):
+    """ Small helper for interacting with the object store and making sure all
+    datasets from a job end up with the same object_store_id.
+    """
+
+    def __init__(self, app):
+        self.object_store = app.object_store
+        self.object_store_id = None
+
+    def set_object_store_id(self, data):
+        # Create an empty file immediately.  The first dataset will be
+        # created in the "default" store, all others will be created in
+        # the same store as the first.
+        data.dataset.object_store_id = self.object_store_id
+        try:
+            self.object_store.create(data.dataset)
+        except ObjectInvalid:
+            raise Exception('Unable to create output dataset: object store is full')
+        self.object_store_id = data.dataset.object_store_id  # these will be the same thing after the first output
--- a/config/galaxy.yml.sample
+++ b/config/galaxy.yml.sample
@@ -296,6 +296,20 @@ galaxy:
   # as the 'watch_tools' option.
   #watch_job_rules: false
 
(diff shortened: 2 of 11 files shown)


mvdbeek:
That is awesome, I also have the vague feeling that this could allow some cool extensions. I suppose a dynamic rule could be implemented that routes jobs to user or tool specific object stores? That would be really cool!

bgruening:
> This is cool and important and I can imagine some awesome applications but I will wait for them to materialize.

Indeed! Super cool, don't wait :) ... xref: https://github.com/galaxyproject/galaxy/issues/4974

jmchilton:
The new test case in #6546 is already paying dividends - I didn't write it for this PR but I included it here and it caught a very subtle problem with this PR that none of the other tests would have.

lib/galaxy/objectstore/__init__.py
```diff
@@ -808,3 +808,24 @@ def _create_object_in_session(obj):
...
+    def set_object_store_id(self, data):
+        # Create an empty file immediately.  The first dataset will be
+        # created in the "default" store, all others will be created in
+        # the same store as the first.
+        data.dataset.object_store_id = self.object_store_id
+        try:
+            self.object_store.create(data.dataset)
+        except ObjectInvalid:
+            raise Exception('Unable to create output dataset: object store is full')
+        self.object_store_id = data.dataset.object_store_id  # these will be the same thing after the first output
```
VJalili, line 831: @jmchilton so you expect the `data.dataset.object_store_id` to change after calling the `create` method?

PR #6551, opened 2018-07-28
Redirect unidentified routes in admin panel

guerler:
Potential follow-up for #6447 to redirect missing routes from the admin to the analysis app. Currently, analysis routes in the masthead are unavailable in the admin app.


dannon:
@guerler Pushed one changeset demonstrating how this can be problematic with multiple routers -- I'm testing good solutions now.

guerler:
Thanks @dannon for catching that. Thats awesome.

PR #6554, opened 2018-07-30
Fix linting issue introduced in #6553.

jmchilton:


dannon:
D'oh.  Thanks.  I actually fixed that locally, too, and then clicked merge on github instead of my local branch...

PR #6555, opened 2018-07-30
Restore log_destination = stdout configuration option.

jmchilton:
Implementation in 17.09 was here https://github.com/galaxyproject/galaxy/blob/release_17.09/lib/galaxy/config.py#L910. This option set to stdout is used by the test framework output Galaxy logs to stdout. As far as I can tell log_destination was removed in 18.01 and the effect was to sort of default to stdout - but when we fixed log_destination for files and backported recently https://github.com/galaxyproject/galaxy/commit/bc95f2563601635cfa1daa686b3d5c4e3ebb51ea - test output became written to a file called "stdout". This commit should restore the 17.09 behavior for both files and stdout configuration of this logging.

### DIFF
--- a/lib/galaxy/config.py
+++ b/lib/galaxy/config.py
@@ -723,7 +723,16 @@ class Configuration(object):
                 'qualname': 'COMPLIANCE'
             }
 
-        if kwargs.get("log_destination", None):
+        log_destination = kwargs.get("log_destination", None)
+        if log_destination == "stdout":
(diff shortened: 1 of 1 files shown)


mvdbeek (approved):

mvdbeek:
Had been wondering what happened to the log output, thanks @jmchilton. This is ready, right ?

jmchilton:
Yup!

PR #6561, opened 2018-07-31
Fix all py3 api tests

mvdbeek:
~~We're almost there, locally only the history import/export fails, and the test_workflow_stability test.~~
I'm running tests using `GALAXY_TEST_TOOL_CONF=config/tool_conf.xml.sample,test/functional/tools/samples_tool_conf.xml pytest -v test/api`.
Also fixes some remaining issues with webob/paste and mako. This feels almost functional!

xref: #1715


lib/galaxy/tools/imp_exp/unpack_tar_gz_archive.py
```diff
@@ -64,7 +64,7 @@ def unpack_archive(archive_file, dest_dir):
     """
     Unpack a tar and/or gzipped archive into a destination directory.
     """
-    archive_fp = tarfile.open(archive_file, mode='r')
+    archive_fp = tarfile.open(archive_file, mode='rb')
```
nsoranzo: `rb` is not listed among the allowed modes in https://docs.python.org/3/library/tarfile.html#tarfile.open , is that necessary?

mvdbeek: Thanks, it is not, there were and are still other issues with the import/export , some of which I fixed in https://github.com/galaxyproject/galaxy/pull/6561/commits/025167546202591524a41ae1f8ceddb859cbac37

tools/filters/random_lines_two_pass.py
```diff
@@ -7,16 +7,69 @@
...
+                j = int(random.random() * n)
+                while j in selected:
+                    j = int(random.random() * n)
+                selected_add(j)
+                result[i] = population[j]
+        except (TypeError, KeyError):   # handle (at least) sets
+            if isinstance(population, list):
+                raise
+            return sample(tuple(population), k)
+    return result
```
jmchilton, line 62: Gross... but needed - thanks for figuring that out!

lib/galaxy/workflow/steps.py
```diff
@@ -35,7 +35,7 @@ def order_workflow_steps(steps):
     if position_data_available:
         steps.sort(key=lambda _: math.sqrt(_.position['left'] ** 2 + _.position['top'] ** 2))
     try:
-        edges = edgelist_for_workflow_steps(steps)
+        edges = sorted(edgelist_for_workflow_steps(steps))
```
mvdbeek, line 38: @jmchilton does this make sense to you ? I'm a little concerned that by updating the test workflows I have bent the test to pass for the sake of passing ...

mvdbeek:
This is ready for review (while I set up a jenkins job for python 3).

lib/galaxy/tools/imp_exp/imp_history_from_archive.xml
```diff
@@ -1,7 +1,11 @@
 <tool id="__IMPORT_HISTORY__" name="Import History" version="0.1" tool_type="import_history">
   <type class="ImportHistoryTool" module="galaxy.tools"/>
   <action module="galaxy.tools.actions.history_imp_exp" class="ImportHistoryToolAction"/>
-  <command>python '$__tool_directory__/unpack_tar_gz_archive.py' '${ str( $__ARCHIVE_SOURCE__ ).encode( 'base64' ) }' '${ str( $__DEST_DIR__ ).encode( 'base64' ) }' --$__ARCHIVE_TYPE__ --encoded</command>
+  <command>#from base64 import b64encode#
+python '$__tool_directory__/unpack_tar_gz_archive.py'
+'${ b64encode(bytearray(str($__ARCHIVE_SOURCE__), encoding="utf-8")).decode('utf-8')}'
```
mvdbeek, line 6: The need to wrestle with b64encode, bytearray and string decoding unfortunately made we realise that of course @jmchilton was right at the GCC , we probably need to keep supporting python 2 for a very long time if we want to remain reproducible. I think it's a small minority of tools though that will fail if not updated:(.

lib/galaxy/jobs/runners/state_handlers/_safe_eval.py
```diff
@@ -19,7 +19,9 @@
 VALID_FUNCTIONS = BUILTIN_AND_MATH_FUNCTIONS + STRING_AND_LIST_METHODS
 
 
-def _check_name(ast_node, allowed_variables=[]):
+def _check_name(ast_node, allowed_variables=None):
+    if allowed_variables is None:
+        allowed_variables = []
```
jmchilton, line 24: I need to stop doing that - sorry. Thanks for the fix!

lib/galaxy/webapps/galaxy/api/tools.py
```diff
@@ -174,6 +174,8 @@ def test_data(self, trans, id, **kwd):
         def json_encodeify(obj):
             if isinstance(obj, odict):
                 return dict(obj)
+            elif isinstance(obj, map):
+                return list(obj)
```
jmchilton, line 178: The Java programmer in me was confused by this - but I guess ``map`` is the iterable result type of mapping over a list so this makes sense.

lib/galaxy/workflow/steps.py
jmchilton: Hmm... such a good question. The fact that sorted is needed means the previous test was somehow depending on the particular hashing or random lib behavior of Python 2.7 I assume - which is less than ideal. I'd like to establish a "correct order" for the steps in a workflow so ``order_index`` is preserved for structurally similar workflows across saves - the test still passes so it is preserved - its just different than it was previously. ``topsort`` using ``random`` so I guess I shouldn't be surprised I'm failing to establish a correct order - the test probably stinks.

jmchilton:
Thanks a million @mvdbeek - amazing!

lib/galaxy/workflow/steps.py
mvdbeek: I don't think `topsort` itself is the issue. I tried this in isolation using the `pairlist` I intercepted form python 2, it gives consistent results between py2 and py3. I think something is different in `edgelist_for_workflow_steps`, perhaps the order in which `step.input_connections` are listed ?

jmchilton:
Saw this in the Selenium test logs for dev, related?

```
  File "/galaxy_venv/local/lib/python2.7/site-packages/paste/httpserver.py", line 1085, in process_request_in_thread
    self.finish_request(request, client_address)
  File "/usr/lib/python2.7/SocketServer.py", line 331, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib/python2.7/SocketServer.py", line 652, in __init__
    self.handle()
  File "/galaxy_venv/local/lib/python2.7/site-packages/paste/httpserver.py", line 459, in handle
    BaseHTTPRequestHandler.handle(self)
  File "/usr/lib/python2.7/BaseHTTPServer.py", line 340, in handle
    self.handle_one_request()
  File "/galaxy_venv/local/lib/python2.7/site-packages/paste/httpserver.py", line 454, in handle_one_request
    self.wsgi_execute()
  File "/galaxy_venv/local/lib/python2.7/site-packages/paste/httpserver.py", line 304, in wsgi_execute
    self.wsgi_start_response)
  File "/galaxy_venv/local/lib/python2.7/site-packages/paste/urlmap.py", line 216, in __call__
    return app(environ, start_response)
  File "/galaxy/lib/galaxy/web/framework/middleware/static.py", line 48, in __call__
    app.cache_control(max_age=int(self.cache_seconds))
AttributeError: 'FileApp' object has no attribute 'cache_control'
```

mvdbeek:
Yes, I'll have a look!

PR #6562, opened 2018-08-01
More Selenium test fixes.

jmchilton:
A lot broke while we weren't running the tests against PRs.


dannon:
@jmchilton Were you thinking we could re-enable these, now?  I worry that, while they cause noise, if we leave them off they'll just keep getting broken worse.

PR #6563, opened 2018-08-01
[WIP] More Job Scheduling Optimizations

jmchilton:
Work on #4959, like #6380 & #6548  but a better set of optimizations - dropped a few things that didn't pan out and added some nice newer optimizations that are better.

Below are some timings for a collection mapping job that creates 100 datasets (10 collections of 10). This branch is getting very close to resulting in a 10x speed up in creating jobs. They are a bit interconnect though still and I want tests for the permission handling changes - so I think I want #6461 in before getting too far with this. Obviously though my initial results are that #6550 is a really nice optimization in its own right and gets us half way there.

dev

galaxy.tools.execute DEBUG 2018-08-01 14:35:00,126 Executed 10 job(s) for tool create_10 request: (13434.709 ms)

dev with populate object store in job handler (PR #6550):

galaxy.tools.execute DEBUG 2018-08-01 14:37:53,418 Executed 10 job(s) for tool create_10 request: (6131.476 ms)

dev + PR#6550 + permission optimization:

galaxy.tools.execute DEBUG 2018-08-01 14:42:08,823 Executed 10 job(s) for tool create_10 request: (5202.416 ms)

dev + PR#6550 + permission optimization + add_datasets query opts:

galaxy.tools.execute DEBUG 2018-08-01 14:33:10,409 Executed 10 job(s) for tool create_10 request: (4712.302 ms)

dev + PR#6550 + permission optimization + add_datasets query opts + flush once:

galaxy.tools.execute DEBUG 2018-08-01 15:14:11,361 Executed 10 job(s) for tool create_10 request: (1549.116 ms)

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -246,6 +246,7 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
         self.database_create_tables = string_as_bool(kwargs.get("database_create_tables", "True"))
         self.database_query_profiling_proxy = string_as_bool(kwargs.get("database_query_profiling_proxy", "False"))
         self.database_encoding = kwargs.get("database_encoding", None)  # Create new databases with this encoding.
+        self.database_log_query_counts = string_as_bool(kwargs.get("database_log_query_counts", False))
         self.thread_local_log = None
         if string_as_bool(kwargs.get("enable_per_request_sql_debugging", "False")):
             self.thread_local_log = threading.local()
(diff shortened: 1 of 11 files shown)


jmchilton:
The only optimization left in here is the one that flushes all jobs for a collection mapping step at once (36c5ba8385790f0e82870a97c8a45297702ce33b). It is a small-big change that in the tests above cut the submission runtime time of a collection mapping step to a third of what it was without the change. The runtime becomes dominated by one really large SA flush after that change (still a lot to be done after that point - but it becomes about figuring out how to get SA to write raw mappings without re-populating object attributes we don't need instead of just avoiding flushes - which has been the main technique to optimize things up until now).

PR #6568, opened 2018-08-03
Unified dataset permission API.

jmchilton:
There was an existing library dataset permission API and one proposed for HDAs in #6461. This tries to reconcile the two approaches and move both approaches toward managers.

- Proper treatment of "master_api_key" throughout the manager layers related to this.
- Use PUT instead of POST, since that is more appropriate for updates I believe.
- Use a similar URL pattern for library datasets, history contents, and dataset APIs.
- Generalize the dataset API to allow HDA or LDDA inputs (the proposal for the new API in #6461 added an HDA-only option to the dataset API - this isn't how that API layer is meant to be used I believe).
- Add an HDA only endpoint in the history contents API for this functionality to mirror the library dataset API.
- Use consistent and backward compatible payload parsing for these endpoints.
- Use consistent serialization for the result of these changes.
- Generalize the HDA variant of this to allow different actions - such as make public and make private that were available in the library dataset variant.
- Lots of new tests - for this functionality as well as for dataset permission checking when running tools and creating collections and guess_derived permissions for tool output datasets.

### DIFF
--- a/lib/galaxy/managers/datasets.py
+++ b/lib/galaxy/managers/datasets.py
@@ -222,57 +222,6 @@ class DatasetSerializer(base.ModelSerializer, deletable.PurgableSerializerMixin)
         return permissions
 
 
-class DatasetDeserializer(base.ModelDeserializer, deletable.PurgableDeserializerMixin):
-    model_manager_class = DatasetManager
-
-    def __init__(self, app):
-        super(DatasetDeserializer, self).__init__(app)
-        # TODO: this manager may make more sense inside rbac_secured
-        self.role_manager = roles.RoleManager(app)
-
-    def add_deserializers(self):
-        super(DatasetDeserializer, self).add_deserializers()
-        # not much to set here besides permissions and purged/deleted
-        deletable.PurgableDeserializerMixin.add_deserializers(self)
-
-        self.deserializers.update({
-            'permissions' : self.deserialize_permissions,
-        })
-
-    def deserialize_permissions(self, dataset, key, permissions, user=None, **context):
-        """
-        Create permissions for each list of encoded role ids in the (validated)
-        `permissions` dictionary, where `permissions` is in the form::
-
-            { 'manage': [ <role id 1>, ... ], 'access': [ <role id 2>, ... ] }
-        """
-        self.manager.permissions.manage.error_unless_permitted(dataset, user)
-        self._validate_permissions(permissions, **context)
-        manage = self._list_of_roles_from_ids(permissions['manage'])
-        access = self._list_of_roles_from_ids(permissions['access'])
-        self.manager.permissions.set(dataset, manage, access, flush=False)
-        return permissions
-
-    def _validate_permissions(self, permissions, **context):
-        self.validate.type('permissions', permissions, dict)
-        for permission_key in ('manage', 'access'):
-            if(not isinstance(permissions.get(permission_key, None), list)):
-                msg = 'permissions requires "{0}" as a list of role ids'.format(permission_key)
-                raise exceptions.RequestParameterInvalidException(msg)
-
-        # TODO: push down into permissions?
-        manage_permissions = permissions['manage']
-        if len(manage_permissions) < 1:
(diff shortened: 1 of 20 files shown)


hexylena:
Thanks @jmchilton this looks way better than #6461. Love it!

lib/galaxy/managers/datasets.py
```diff
@@ -365,6 +315,85 @@ def extra_files(self, dataset_assoc):
...
+        rval = dict(access_dataset_roles=access_dataset_role_list, manage_dataset_roles=manage_dataset_role_list)
+        if library_dataset is None:
+            modify_roles = set(trans.app.security_agent.get_roles_for_action(library_dataset, trans.app.security_agent.permitted_actions.LIBRARY_MODIFY))
+            modify_item_role_list = [(modify_role.name, trans.security.encode_id(modify_role.id)) for modify_role in modify_roles]
+            rval["modify_item_roles"] = modify_item_role_list
+        return rval
+
+    def update_permissions(self, trans, dataset_assoc, **kwd):
+        action = kwd.get('action', None)
+        if action not in ['remove_restrictions', 'make_private', 'set_permissions']:
```
hexylena: Maybe `make_public` would be more explicit than `remove_restrictions` for end users?

lib/galaxy/webapps/galaxy/api/datasets.py
```diff
@@ -93,6 +94,26 @@ def show(self, trans, id, hda_ldda='hda', data_type=None, provider=None, **kwd):
             trans.response.status = 500
         return rval
 
+    @web._future_expose_api
+    def set_permissions(self, trans, dataset_id, payload, **kwd):
+        """
+        POST /api/datasets/{encoded_dataset_id}/permissions
```
hexylena: should this be `PUT` here?

hexylena (approved):

lib/galaxy/managers/datasets.py
```diff
@@ -365,6 +314,85 @@ def extra_files(self, dataset_assoc):
...
+        rval = dict(access_dataset_roles=access_dataset_role_list, manage_dataset_roles=manage_dataset_role_list)
+        if library_dataset is not None:
+            modify_roles = set(trans.app.security_agent.get_roles_for_action(library_dataset, trans.app.security_agent.permitted_actions.LIBRARY_MODIFY))
+            modify_item_role_list = [(modify_role.name, trans.security.encode_id(modify_role.id)) for modify_role in modify_roles]
+            rval["modify_item_roles"] = modify_item_role_list
+        return rval
+
+    def update_permissions(self, trans, dataset_assoc, **kwd):
+        action = kwd.get('action', 'set_permissions')
+        if action not in ['remove_restrictions', 'make_private', 'set_permissions']:
```
hexylena, line 340: What was your feeling re: `make_public` (or `make_public_by_link` instead of `remove_restrictions`?)?

jmchilton: I thought we could add that as an alias but I wanted to maintain the previous action name also for backward compatibility in data libraries. I'll make this change!

hexylena: Backwards compat is a great reason! If that's what devs/clients are expecting to need, definitely no reason to add something else we'll accept.

jmchilton: Fair, thanks!

lib/galaxy/managers/users.py
```diff
@@ -103,12 +103,20 @@ def by_email_like(self, email_with_wildcards, filters=None, order_by=None, **kwa
...
+
+        Do not pass trans to simply check if an existing user object is an admin user,
+        pass trans when checking permissions.
         """
         admin_emails = self._admin_emails()
-        return user and admin_emails and user.email in admin_emails
+        if user is None:
+            # Anonymous session or master_api_key used, if master_api_key is detected
+            # return True.
+            log.info("in is_admin.... %s" % trans)
```
nsoranzo, line 116: @jmchilton Just saw this in my logs, debug leftover?

PR #6577, opened 2018-08-08
improved grouping tool

bernt-matthias:
- replaces non 'floatable' values to a user supplied default value
- do not strip lines because this removes empty cells at the ends
- added some comments to python code

Would be nice to keep column names .. any ideas? Also I could not run the test manually with planemo since I could not find the test data (results). Can we uncomment the last test?

### DIFF
--- a/tools/stats/grouping.xml
+++ b/tools/stats/grouping.xml
@@ -1,4 +1,4 @@
-<tool id="Grouping1" name="Group" version="2.1.1">
+<tool id="Grouping1" name="Group" version="2.1.2">
   <description>data by a column and perform aggregate operation on other columns.</description>
   <command interpreter="python">
     grouping.py 
@@ -8,9 +8,7 @@
       "${ignorecase}"
       "${ignorelines}"
       #for $op in $operations
-       '${op.optype}
-        ${op.opcol}
-        ${op.opround}'
+        '${op.optype},${op.opcol},${op.opround},${op.opdefault}'
       #end for
   </command>
   <inputs>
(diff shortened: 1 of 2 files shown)


tools/stats/grouping.xml
```diff
@@ -19,7 +17,7 @@
     <param name="ignorecase" type="boolean" truevalue="1" falsevalue="0">
       <label>Ignore case while grouping?</label>
     </param>
-    <param name="ignorelines" type="select" display="checkboxes" multiple="True" label="Ignore lines beginning with these characters" help="lines beginning with these are not grouped">
+    <param name="ignorelines" type="select" multiple="True" label="Ignore lines beginning with these characters" help="lines beginning with these are not grouped">
```
jmchilton: Why change checkboxes?

jmchilton:
Gotta bump tool version I think.

tools/stats/grouping.xml
bernt-matthias: restored the checkboxes

bernt-matthias:
> Gotta bump tool version I think.

Done

mvdbeek:
We've recently updated the tool for py3 compatibility, can you rebase this against dev ?

bernt-matthias:
hmm. apparently I did something wrong in the git action that was intended to be a rebase. But at least the conflict is gone. The question is if its OK to open a named temporary read only and then let an external program (in this case sort) write in to it. Technically this works -- just a bit nasty.

nsoranzo:
@bernt-matthias I've rebased the PR and I had to squash all your commits in 1, I hope that's fine. Can you double check that the changes look correct to you?

bernt-matthias:
looks good

bernt-matthias:
I guess the failing integration test is unrelated? seems to be fastqc failing.

mvdbeek:
@galaxybot test py3

jmchilton:
Thanks a bunch, looks like a good enhancement!

PR #6578, opened 2018-08-08
Fix last consistently failing Selenium test.

jmchilton:
Target HID here changed when we went to building lists of pairs directly with upload 2.0 instead of from individual pairs and the HDCA API.


dannon:
Verified by running selenium tests locally.

Issue #5142, opened 2017-12-06
History Panel Style Access to Workflow Invocations


jmchilton:
We talk about this a lot but it doesn't seem to have an issue as far as I can tell. Running workflows need to be present somewhere the user can readily see them in the analysis view. We need to be able to provide feedback on workflow scheduling status and runtime errors that are encountered, we need to be able to navigate to outputs, see inputs, track recorded parameters, etc.... There are cool things implemented in the backend such as pause to review steps that can't really be used without this feature as well.

I don't know if they should be a separate tab in the history panel, if they should be alongside datasets like collections are, or if they should be their own thing somewhere else - but they need to be somewhere.

blankenberg:
This is greatly needed. 

I can 'run' a workflow and every UI indication shows that it worked: green success message in UI, and a new history is created. But several hours later I just have my new target history that is empty. 

Is it working really hard? Did it fail and give up? What is going on? If it failed, it should at least let me know to stop waiting and hoping. 

Failures due to scheduling used to get displayed to the user, but now that it is asynchronous they do not.

xref https://github.com/galaxyproject/galaxy/issues/5265#issuecomment-365964018

jmchilton:
As a short term fix we could also monitor the invocation from the workflow execution landing page where we have a lot more freedom - at least then it is a regression of behavior relative to the synchronous case.

blankenberg:
I think something is way better than nothing and this should be a pretty quick change?

Maybe show a paginated list of the last 5-10 workflows and their status + cancel, etc (if running) buttons would be a great start.

martenson:
xref https://github.com/galaxyproject/galaxy/pull/8699

mvdbeek:
We've got something like this with https://github.com/galaxyproject/galaxy/issues/5179 and the invocation details component. We should maybe make those navigable directly from the history, but that seems like a different project.

PR #6580, opened 2018-08-08
Add API endpoint to fetch a history's workflow invocations.

jmchilton:
Refactor toward managers to allow unified serialization of invocations and such. Add test case.

In service of #5142 I suppose, builds on #6559 (cherry-picked into dev).

Next up... I want to add a jobs_summary API that matches collections (https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/api/history_contents.py#L183) - this should allow effective summary of job states for fully scheduled workflows. One can then imagine a progress bar for whats steps are scheduled and below it a progress bar for what job states are - this second one built on this new API I'm planning and that could probably completely reuse the collection summary code on the client.

### DIFF
--- a/lib/galaxy/managers/workflows.py
+++ b/lib/galaxy/managers/workflows.py
@@ -181,22 +181,49 @@ class WorkflowsManager(object):
         trans.sa_session.flush()
         return workflow_invocation_step
 
-    def build_invocations_query(self, trans, decoded_stored_workflow_id):
+    def build_invocations_query(self, trans, stored_workflow_id=None, history_id=None, user_id=None):
         """Get invocations owned by the current user."""
-        stored_workflow = trans.sa_session.query(
-            self.app.model.StoredWorkflow
-        ).get(decoded_stored_workflow_id)
-        if not stored_workflow:
-            raise exceptions.ObjectNotFound()
-        invocations = trans.sa_session.query(
-            model.WorkflowInvocation
-        ).filter_by(
-            workflow_id=stored_workflow.latest_workflow_id
-        )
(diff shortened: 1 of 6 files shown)


bgruening (approved):
Thanks this is great! Displaying all this information to the user at the end will be a big improvement.

dannon:
@jmchilton My only question is about getting *all* of a user's active invocations.  Would we want another endpoint, additional query parameters here, or multiple requests for all histories?

(I ask, because this is a feature I started working on at the hackathon and then had this exact issue with -- displaying all active invocations on the workflows landing page)

jmchilton:
I can add ``/api/invocations`` to sort of mirror ``/api/jobs`` - should be pretty trivial. Would that be enough?

dannon:
@jmchilton Yeah, that's what I was getting at.  You think the second endpoint is better than having `/api/invocations` defaulting to all invocations, filtering with a `history_id` parameter or other query params?  (like workflow id, etc)

jmchilton:
You know me - I'd just add both 😄. But I'm totally fine dropping the ``/api/histories`` view of this and just adding the ``/api/invocations`` with ``history_id`` parameter. I'll rework this, thanks for the feedback @dannon.

dannon:
@jmchilton Thanks!  I think it'll be a little more flexible in the long run, and I'm excited to use it to finish off this active invocations view!

jmchilton:
I think the changes are good for review - ``/api/invocations`` is now good to go. It can take a ``history_id``, ``workflow_id``, or neither.

dannon:
@jmchilton Thanks!  Checking it out.

PR #5903, opened 2018-04-17
Cloud Authorization

VJalili:
This PR implements the logic of accessing a user's cloud-based resources without asking for their credentials, leveraging [CloudAuthz](https://github.com/galaxyproject/cloudauthz). In general, this PR implements the following: 

1. defines a `cloudauthz` class in Galaxy model;
2. implements migration script to create a table in Galaxy database to store instances of the new class;
3. updates mappings by required joins;
4. creates API and manager so that user can define a `cloudauthz` instance;
5. integrates [`cloudauthz`](https://github.com/galaxyproject/cloudauthz) with `AuthnzManager`;
6. replaces `credentials` with a `cloudauthz` ID in the  cloud `upload` API;
7. updates `CloudManager` to leverage `cloudauthz` to get credentials.
8. updates `CloudManager` to adhere with `v1.0.0` cloudbridge interface.  

This PR requires the following to updated: 
- [x] Version 1.0.0 of CloudBridge; #6660 
- [x] Update [`cloudauthz`](https://github.com/galaxyproject/cloudauthz) to resolve dependency conflicts with Galaxy;
- [x] Some of the raised exceptions cause 500 error (e.g., exceptions in `galaxy/api/cloudauthz.py`).

PR #3383, opened 2017-01-04
Add extension point for user preferences

bgruening:
This pull request will implement an optional extension point to extend the user-preferences in a dynamic, admin-controlled way.

The admin defines in a yaml files allowed settings (name, type ...). A user can enter these settings and it will be saved in the user-object as external user preferences. These settings will become available inside the User object and with this accessible everywhere else.

### Password Parameter Implementation (https://github.com/galaxyproject/galaxy/pull/3121)

A tool needs a password or any user-specific information, like @erasche Apollo stuff. Such a tool just need to assume the variable that it needs as environment variable ($MY_APOLLO_SERVER). An admin uses the `dynamic_job_destinations` to write a small function to map an arbitrary user-preference to this environment variable. 

* no tool syntax changes needed
* no information is leaked during tool installation
* central point to store and secure sensible data
* a user only needs to enter once a configuration and reuse them whenever it is needed (workflow etc ...)

### Webhooks can use personalized data

We have a overlay-search in development and it would be great if plugins can collect user-preferences, like the `n` in give me the `n` most used tools.
An other plugin we have developed is a simple RSS reader that shows interesting papers or science news after tool submission. To make this configurable we need such an extension point.

### Bjoern's Blue Sky Idea

I would like to run a pulsar VM in our OpenStack cloud and register this specific VM in my account so that a few of my jobs can be calculated in my personal VM ... enabling me to bring my own compute resources.

I promise to write some documentation about the connection to `dynamic_job_destinations` and how you can abuse this very flexible idea if I get a green light.

PR #6583, opened 2018-08-08
Send data to cloud

VJalili:
This PR implements the logic of [downloading](https://github.com/galaxyproject/galaxy/blob/5d85c32da24396a4853d09cf5fccb12481f1ccee/lib/galaxy/managers/cloud.py#L236) data from Galaxy to a cloud-based storage, as a tool, hence it can be run as job. 

This tool can be invoked using the following two methods:

1. API call, send a payload as the following to `/api/cloud/storage/download`

```python
payload = {
    "history_id": "...",
    "bucket": "...",
    "authz_id": "..."
    }}
```

2. Graphical interface: 

screen shot 2018-10-11 at 11 59 29 am

This PR leverages [CloudAuthz](https://github.com/galaxyproject/cloudauthz) that PR #5903 integrates it with Galaxy. Accordingly, user **does not provide Galaxy with access credentials to their account** (see the above screenshot and API payload). 

See PR #5903 on defining a CloudAuthz configuration (to obtain `authz_id`).


tools/cloud/download.xml
```diff
@@ -0,0 +1,86 @@
...
+        -->
+        <conditional name="connection">
+            <param name="provider" type="select" label="Cloud-Based Storage Provider">
+                <option value="aws">Amazon Simple Storage Service (S3)</option>
+                <option value="azure">Microsoft Azure Blob Storage</option>
+                <option value="openstack">OpenStack Object Storage (Swift)</option>
+            </param>
+            <when value="aws">
+                <param name="access_key" label="Access key" type="text"/>
+                <param name="secret_key" label="Secret key" type="text"/>
```
jmchilton: So when a user clicks view info on these datasets they can see the secret key and such right? These things cannot be tool inputs IMO, am I missing something? 

I think you want to start with a special class of user preferences for these tools (probably based on https://github.com/galaxyproject/galaxy/pull/3383/files) and then add a tool parameter type that takes a credential "name" that the user has configured through their parameters and then use these preferences associated with the user to build the input file? We can then build from there to create per-cloud inputs on the tool form that update the preferences and to make the UX more smooth.

VJalili: @jmchilton thanks for looking into this. I agree it is not a good practice to put user credentials in tool input. Also, the method you linked sounds great. However, we're working on another method that enables accessing cloud-based resource __without__ asking user for their credentials (see #5903 and https://github.com/galaxyproject/cloudauthz, which I'm going to work on it next).

jgoecks:
Tested and works as advertised. +1

Some suggestions:
1. Perhaps “Download to cloud” should be “Copy to Cloud”?
2. Perhaps tool log should include more logging? All I get now is a warning.
3. Could you get around parameter ordering by using keyword parameters?

VJalili:
@jgoecks Thanks for testing this. 

> Perhaps “Download to cloud” should be “Copy to Cloud”?

We have the following naming options: 
 - pull/get/copy/upload data __from__ cloud;
 - push/put/copy/download data __to__ cloud.

Given that we have the _upload_ tool, and the widely used upload/download terms, IMHO, the upload/download terminology could be more intuitive than other options. However, this can be changed if upload/download are misleading here. 

> Could you get around parameter ordering by using keyword parameters?

Are you referring to the following?

https://github.com/VJalili/galaxy/blob/897d024ef7301403ec1ead14452493b2fda62bc1/tools/cloud/download.xml#L38-L44

If so, I think that is related to `__current_case__`, which I think it shall be assigned with order number of an item that we select via API/manager (see [this line](https://github.com/VJalili/galaxy/blob/897d024ef7301403ec1ead14452493b2fda62bc1/lib/galaxy/managers/cloud.py#L314)).

VJalili:
ping @jmchilton now this tool neither asks nor stores user credentials. Any suggestion on the tool and api/manager? 


ping @jgoecks 
> 2. Perhaps tool log should include more logging? All I get now is a warning.
> 3. Could you get around parameter ordering by using keyword parameters?

Both addressed.

tools/cloud/download.xml
```diff
@@ -26,7 +26,7 @@
         <param name="overwrite_existing" type="boolean" label="overwrite_existing" />
     </inputs>
     <outputs>
-        <data name="output" type="data" label="Output log" hidden="false"/>
+        <data name="output" type="data" label="Download to cloud log" hidden="false"/>
```
dannon: I saw the commit message and got excited, but alas -- we're still downloading to the cloud.  What can I do to persuade you to rename this from "Download to cloud" to something like "Send to cloud"?  As we all talked about at the last meeting, "download" has a strong meaning for most people that involves the device they're currently on -- not cloud storage.

VJalili:
@galaxybot test this

VJalili:
@galaxybot test this

VJalili:
@galaxybot test this

nuwang:
As a user, how would I obtain an authz id? Is it single use? Would it be an option to automatically list the buckets?

tools/cloud/download.xml
martenson: 'Transfer' is a word that has no direction implication and may be useful here.

VJalili:
> how would I obtain an authz id?

Go to `api/cloud/authz` and you can get a list of all authz you have defined. (Check https://github.com/galaxyproject/galaxy/pull/5903 on how to create a cloudauthz config.)

> Is it single use?

Cloudauthz configs are defined by a user, and can be used by that user only. 

> Would it be an option to automatically list the buckets?

Yes, it will be in a future work.

tools/cloud/download.xml
VJalili: @dannon refactored to `Send`. Do you have any other comments/concerns about this PR?

VJalili: ping @jgoecks ^ (I think you had similar naming concerns).

afgane:
Overall, the tool is operational and really cool! We did find a couple of issues, one when selecting multiple files in the UI and a usability issue w/ an output log being created for each file being uploaded, which would potentially generate double the number of datasets in the history. Just summarizing those there while they're being worked on.

nsoranzo:
@galaxybot test py3

VJalili:
@galaxybot test this

tools/cloud/send.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0"?>
+
+<tool id="send_to_cloud" name="Send to cloud" version="0.1.0" workflow_compatible="false">
+    <requirements>
+        <requirement type="package" version="0.3.3">cloudbridge</requirement>
```
nsoranzo, line 5: @VJalili This requirement is presently listed in the "Admin" -> "Manage dependencies" interface as missing, but is not installable because there is no conda package for `cloudbridge`.
Since this is not actually needed (cloudbridge comes from the Galaxy virtualenv since this tool is listed in `GALAXY_LIB_TOOLS_UNVERSIONED`), I'd suggest to remove or comment out (`<!-- ... -->`) the requirement.

VJalili: @nsoranzo agree, please check https://github.com/galaxyproject/galaxy/pull/7727

Issue #6585, opened 2018-08-09
ObjectNotFound when extra_files_path doesn't exist (with data manager)


abretaud:
Hi,
I'm playing with [data_manager_manual](https://github.com/galaxyproject/tools-iuc/tree/master/data_managers/data_manager_manual), and I get this error on one of my two Galaxy instances (the only one using objectstore):

```
Traceback (most recent call last):
  File "lib/galaxy/jobs/runners/__init__.py", line 690, in finish_job
    self._finish_or_resubmit_job(job_state, stdout, stderr, exit_code)
  File "lib/galaxy/jobs/runners/__init__.py", line 433, in _finish_or_resubmit_job
    job_state.job_wrapper.finish(stdout, stderr, exit_code, check_output_detected_state=check_output_detected_state)
  File "lib/galaxy/jobs/__init__.py", line 1415, in finish
    self.tool.exec_after_process(self.queue.app, inp_data, out_data, param_dict, job=job)
  File "lib/galaxy/tools/__init__.py", line 2293, in exec_after_process
    data_manager.process_result(out_data)
  File "lib/galaxy/tools/data_manager/manager.py", line 329, in process_result
    if os.path.exists(ref_file.extra_files_path):
  File "lib/galaxy/model/__init__.py", line 2128, in extra_files_path
    return self.dataset.extra_files_path
  File "lib/galaxy/model/__init__.py", line 1931, in get_extra_files_path
    return self.object_store.get_filename(self, dir_only=True, extra_dir=self._extra_files_path or "dataset_%d_files" % self.id)
  File "lib/galaxy/objectstore/__init__.py", line 506, in get_filename
    return self._call_method('get_filename', obj, ObjectNotFound, True, **kwargs)
  File "lib/galaxy/objectstore/__init__.py", line 527, in _call_method
    % (method, str(obj), str(kwargs)))
ObjectNotFound: objectstore, _call_method failed: get_filename on , kwargs: {'dir_only': True, 'extra_dir': 'dataset_89476_files'}
```

It seems like a ObjectNotFound is thrown when `lib/galaxy/tools/data_manager/manager.py` is just getting the directory path with `ref_file.extra_files_path`
The directory doesn't exist, but it shouldn't be a problem as there is a `os.path.exists` check after getting the path.

bgruening:
See here: https://github.com/galaxyproject/galaxy/issues/1765

abretaud:
Thanks @bgruening,
[...]

PR #6588, opened 2018-08-09
[18.05] Use object store to check for existence of extra_files_path

mvdbeek:
This should fix https://github.com/galaxyproject/galaxy/issues/6585 for now.
I will make the normal DiskObjectStore also raise ObjectNotFound in dev and figure out if anything else needs fixing.


abretaud:
Cool, thanks! I tested it, it works great

PR #6594, opened 2018-08-13
Respect unavailable display peek

mvdbeek:
This fixes the following exception when running the Filter1 tool via the API

```
galaxy.datatypes.tabular ERROR 2018-08-12 22:53:42,939 make_html_peek_rows failed on HDA 64
  File "/Users/mvandenb/src/galaxy/lib/galaxy/datatypes/tabular.py", line 186, in make_html_peek_rows
    dataset.set_peek()
  File "/Users/mvandenb/src/galaxy/lib/galaxy/model/__init__.py", line 2264, in set_peek
    return self.datatype.set_peek(self)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/datatypes/tabular.py", line 57, in set_peek
    super(TabularData, self).set_peek(dataset, line_count=line_count, WIDTH=WIDTH, skipchars=skipchars, line_wrap=False)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/datatypes/data.py", line 813, in set_peek
    dataset.peek = get_file_peek(dataset.file_name, WIDTH=WIDTH, skipchars=skipchars, line_wrap=line_wrap)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/datatypes/data.py", line 1029, in get_file_peek
    with compression_utils.get_fileobj(file_name, "U") as temp:
  File "/Users/mvandenb/src/galaxy/lib/galaxy/util/compression_utils.py", line 29, in get_fileobj
    return get_fileobj_raw(filename, mode, compressed_formats)[1]
  File "/Users/mvandenb/src/galaxy/lib/galaxy/util/compression_utils.py", line 58, in get_fileobj_raw
    return compressed_format, io.open(filename, mode, encoding='utf-8')
FileNotFoundError: [Errno 2] No such file or directory: '/private/var/folders/df/6xqpqpcd7h73b6jpx9t6cwhw0000gn/T/tmpan_4qaow/tmpfzus9jhv/tmplqjtbp5c/database/files/000/dataset_40.dat'
```

### DIFF
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1375,11 +1375,13 @@ class JobWrapper(HasResourceParameters):
                             return path
 
                         dataset.metadata.from_JSON_dict(output_filename, path_rewriter=path_rewriter)
+                    line_count = context.get('line_count', None)
                     try:
-                        assert context.get('line_count', None) is not None
-                        dataset.set_peek(line_count=context['line_count'])
-                    except Exception:
-                        dataset.set_peek()
+                        # Certain datatype's set_peek methods contain a line_count argument
+                        dataset_assoc.dataset.set_peek(line_count=line_count)
+                    except TypeError:
+                        # ... and others don't
+                        dataset_assoc.dataset.set_peek()
                 else:
                     # Handle purged datasets.
                     dataset.blurb = "empty"
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -54,7 +54,6 @@ from galaxy.util.sanitize_html import sanitize_html
 from galaxy.web.form_builder import (AddressField, CheckboxField, HistoryField,
                                      PasswordField, SelectField, TextArea, TextField, WorkflowField,
                                      WorkflowMappingField)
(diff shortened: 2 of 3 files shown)


lib/galaxy/jobs/__init__.py
```diff
@@ -1375,11 +1375,13 @@ def path_rewriter(path):
...
                         dataset.metadata.from_JSON_dict(output_filename, path_rewriter=path_rewriter)
+                    line_count = context.get('line_count', None)
                     try:
-                        assert context.get('line_count', None) is not None
-                        dataset.set_peek(line_count=context['line_count'])
-                    except Exception:
-                        dataset.set_peek()
+                        # Certain datatype's set_peek methods contain a line_count argument
+                        dataset_assoc.dataset.set_peek(line_count=line_count)
+                    except TypeError:
```
dannon, line 1382: Would it be worth forcing a standard interface for set_peek down the road?  Just food for thought -- not sure I have a strong opinion one way or the other, but that'd be an alternative to having to catch and re-invoke here.

mvdbeek: absolutely, I think we should do that.

Issue #6566, opened 2018-08-03
Workflow download from workflow editor


bebatut:
Hi,

Would it possible to add the possibility to export a workflow from the workflow editor?
Thanks

Bérénice

bgruening:
@anuprulez could you look at this maybe?

anuprulez:
@bgruening I can have a look at it.

jmchilton:
Implemented in #6596

PR #6596, opened 2018-08-13
Add download option to workflow editor menu

anuprulez:
The PR adds a "download" workflow option to the dropdown menu (on the right) of the workflow editor screen. It fixes the feature request in https://github.com/galaxyproject/galaxy/issues/6566.

download_opt

### DIFF
--- a/client/galaxy/scripts/mvc/workflow/workflow-view.js
+++ b/client/galaxy/scripts/mvc/workflow/workflow-view.js
@@ -322,6 +322,10 @@ export default Backbone.View.extend({
                     self.workflow.clear_active_node();
                 },
                 "Auto Re-layout": layout_editor,
+                Download: {
+                    url: `${Galaxy.root}api/workflows/${self.options.id}/download?format=json-download`,
+                    action: function() {}
+                },
                 Close: close_editor
             });
         }


anuprulez:
@bgruening If someone tries to download an unsaved workflow, the browser already gives a warning whether to "stay" on page or "leave". A user should save the workflow and then download.

hexylena:
@anuprulez would the button maybe be more user-friendly as "save and download" then?

anuprulez:
@erasche To have such a feature, a new API needs to be registered which should call two APIs (to update a workflow and then download). Every time a user clicks on "save and download", both the APIs are called (which may be unnecessary if the workflow is already saved). Due to these reasons, I kept the "download" feature separate.

hexylena:
@anuprulez couldn't you just call the existing save API from the client, and then when that returns successfully, redirect to the download API? I did not imagine that it would need a new server side API.

And an unnecessary save call is never harmful. But if you were worried about that, isn't there a dirty flag on the workflow, or similar that could let you tell if a save was needed?

Edit: just call `save_current_workflow` before redirecting? That even includes the check for an unnecessary save.

dannon:
I'm actually in favor of two explicitly separate actions here, leaving this feature as currently implemented by @anuprulez, but maybe I'm in the minority?

To me, this way is simpler, and on the off chance a user has edited but not saved, the existing 'warn before doing anything' paradigm is totally fine IMO.

(I was going to merge as-is, but intended to dig into the need for the empty action function)

hexylena:
No, makes sense, small minority won't have saved. Works for me. Let's merge :)

dannon:
Sounds good, then!  @anuprulez Do you remember why you needed to have the empty action function?  That seemed like something that we should be able to simplify.

anuprulez:
@dannon I needed to create an `anchor` tag in the dropdown menu and in the file `/galaxy/client/galaxy/scripts/ui/popupmenu.js` requires an `action` attribute (`line 47`). Otherwise, it throws console error.

dannon:
@anuprulez Hrmm.  I thought action was intended to be optional, but anyway, I can look into that later.  Thanks for the contribution!

PR #6606, opened 2018-08-15
Refresh an expired OIDC ID token

VJalili:
This PR implements the following:
1- makes sure an OAuth2.0 _refresh token_  is provided when authenticating a user;
2- when returning an OIDC _ID tokne_, first checks if token has not expired, and it is, it refreshes the token, and returns a valid token.


VJalili:
@jmchilton Thanks!

PR #6609, opened 2018-08-16
Order data parameters before unlinked multiplication

mvdbeek:
The 2 failed tests [here](https://jenkins.galaxyproject.org/job/docker-api-py3/30/testReport/junit/) fail randomly, and at least the first one relies in the order of implicit collections.

### DIFF
--- a/lib/galaxy/managers/jobs.py
+++ b/lib/galaxy/managers/jobs.py
@@ -203,7 +203,7 @@ class JobSearch(object):
                     return []
 
         for k, v in wildcard_param_dump.items():
-            wildcard_value = json.dumps(v).replace('"id": "__id_wildcard__"', '"id": %')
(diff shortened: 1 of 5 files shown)


mvdbeek:
Still finding more of these ..

mvdbeek:
This should really be an issue with dictionary ordering -- I can reproduce the failures on python 3.5, but not 3.6, which has stable ordering ...

mvdbeek:
Setting the PYTHONHASH to 3 makes the test always pass, 1 or 2 always fail ...

mvdbeek:
Alright, the last 10 runs all passed (https://jenkins.galaxyproject.org/job/docker-api-py3/), this is ready for review.

jmchilton:
Thanks you so much for deep diving into this!

PR #6611, opened 2018-08-17
Raise an exception if a history with decoded id does not exist.

VJalili:
In the upload to cloud manager, raise an exception if a history with decoded id does not exist.

### DIFF
--- a/lib/galaxy/managers/cloud.py
+++ b/lib/galaxy/managers/cloud.py
@@ -171,7 +171,7 @@ class CloudManager(sharable.SharableModelManager):
         :param trans:       Galaxy web transaction
 
         :type  history_id:  string
-        :param history_id:  the (encoded) id of history to which the object should be uploaded to.
+        :param history_id:  the (decoded) id of history to which the object should be uploaded to.
 
         :type  provider:    string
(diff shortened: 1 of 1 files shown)


lib/galaxy/managers/cloud.py
```diff
@@ -221,6 +221,8 @@ def upload(self, trans, history_id, provider, bucket, objects, credentials, inpu
             params = Params(self._get_inputs(obj, key, input_args), sanitize=False)
             incoming = params.__dict__
             history = trans.sa_session.query(trans.app.model.History).get(history_id)
+            if not history:
```
mvdbeek, line 224: Not related to this PR, but I think the docstring on line 173 should not say `(encoded)`, right ?

mvdbeek:
This looks fine to me, is it ready ?

lib/galaxy/managers/cloud.py
VJalili: Thats correct. I updated that.

VJalili:
@mvdbeek Thanks for checking this; yes it is ready.

mvdbeek (approved):

PR #6613, opened 2018-08-17
add dcd datatype for molecular dynamics

chrisbarnettster:
xref: https://www.charmm.org/charmm/documentation/by-version/c42b1/params/doc/io/
xref: http://www.ks.uiuc.edu/Research/vmd/plugins/molfile/dcdplugin.html

A datatype for molecular dynamics simulations run using CHARMM and NAMD.


lib/galaxy/datatypes/binary.py
```diff
@@ -1907,6 +1907,59 @@ def sniff(self, filename):
...
+        Binary.__init__(self, **kwd)
+        self._magic_number = b'CORD'
+
+    def sniff(self, filename):
+        # The second struct of any dcd file contain 'CORD'
+        # First struct is not sufficient, velocity files will match.
+        # Not checking for endianness
+        try:
+            intsize = 4
+            header = open(filename, 'rb')
```
bgruening: can we use `with open() as header:` here?

bgruening (commented):
Thanks @chrisbarnettster!

lib/galaxy/datatypes/binary.py
chrisbarnettster: yes, updating :)

lib/galaxy/datatypes/binary.py
```diff
@@ -1907,6 +1907,59 @@ def sniff(self, filename):
...
+    >>> from galaxy.datatypes.sniff import get_test_fname
+    >>> fname = get_test_fname('test_glucose_vacuum.dcd')
+    >>> Dcd().sniff(fname)
+    True
+    >>> fname = get_test_fname('interval.interval')
+    >>> Dcd().sniff(fname)
+    False
+    """
+    file_ext = "dcd"
+    edam_format = "format_3842"  # molecular simulation data
```
bgruening: @hmenager does this look good for you?
If so should we also add the same ID for the trr datatype?

chrisbarnettster: Link to the ontology - https://www.ebi.ac.uk/ols/ontologies/edam/terms?iri=http%3A%2F%2Fedamontology.org%2Fdata_3842

chrisbarnettster:
Thanks @bgruening for review. As suggested on gitter chat, I have used seek and included a CHARMM vel datatype. The edam format may not be appropriate for the velocity file

hmenager:
@chrisbarnettster @bgruening, unfortunately the EDAM annotation needs corrections:
- `edam_format = "format_3842"` should become `edam_data = "data_3842"` because "Molecular simulation data" is a data annotation, not the specific DCD format used to serialize this kind of information
- we cannot currently add the `edam_format` for the DCD format, as it appears it is not in EDAM. What about we create the term in EDAM and you add the annotation when this comes through the next release?
Thanks for making the effort to annotate datatypes with EDAM, this is very much appreciated ;)

bgruening:
@hmenager thanks for looking into it! Always good to have an expert around! Lets add DCD and also TRR to EDAM if possible.

chrisbarnettster:
@hmenager thank you, sounds like a good plan. 
I will remove the annotation for now and update on next release.

lib/galaxy/datatypes/binary.py
```diff
@@ -1920,7 +1920,6 @@ class Dcd(Binary):
     False
     """
     file_ext = "dcd"
-    edam_format = "format_3842"  # molecular simulation data
```
bgruening: You can really use `edam_data` it is supported by galaxy. See here: https://github.com/galaxyproject/galaxy/blob/30e3658b8b0e2f6b975dc6ccccb0cc8cc040247c/lib/galaxy/datatypes/annotation.py#L15

hmenager:
regarding trr files, same applies: the format is not in EDAM. 
Regarding both, can I suggest you request these formats using the shiny-shiny term request forms we now have on github?
These are available here:
https://github.com/edamontology/edamontology/issues/new/choose
thanks ;)

lib/galaxy/datatypes/binary.py
chrisbarnettster: updating 💯

bgruening:
Thanks all!

blankenberg:
Does it make sense to have a separate molecular_dynamics.py file for these datatypes?

PR #6627, opened 2018-08-22
Test tools shipped with Galaxy

mvdbeek:
This fixes tests for a lot of the tools we ship with galaxy. I've set up a Jenkins job for these as well.

xref. #1715

PR #6631, opened 2018-08-23
Fix metadata elements reset when creating Metadata files

mvdbeek:
This is a fallout out from 9780d069c8646fb4f02de848d62159680d95ae79

I noticed that some metadata elements from MAF files we're missing while working on https://github.com/galaxyproject/galaxy/pull/6627

### DIFF
--- a/lib/galaxy/model/metadata.py
+++ b/lib/galaxy/model/metadata.py
@@ -167,6 +167,13 @@ class MetadataCollection(object):
                 raise ValueError("json_dict must be either a dictionary or a string, got %s." % (type(json_dict)))
         else:
             raise ValueError("You must provide either a filename or a json_dict")
+
+        # We build a dictionary for metadata name / value pairs
+        # because when we copy MetadatTempfile objects we flush the datasets'
+        # session, but only include the newly created MetadataFile object.
+        # If we were to set the metadata elements in the first for loop we'd
+        # lose all previously set metadata elements
+        metadata_name_value = {}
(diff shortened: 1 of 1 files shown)


jmchilton:
Sorry for breaking this. I've read your comment a few times and I'm still unclear about what is happening. Any clue if this introduces a new flush in there or is this about getting SA to recognize objects that have changed value and it wasn't detecting as changed before?

mvdbeek:
I'm not entirely sure what's going on either ... so what I'm seeing is that inside the loop that sets `dataset._metadata[name]` as soon as we flush here https://github.com/galaxyproject/galaxy/blob/5c889cd5e18e60818a922816884f56cb35262a2e/lib/galaxy/model/metadata.py#L566

we're losing all previously set metadata elements. So my interpretation is that when we flush only certain objects, the "dirty" ? objects are reset. Does that make any sense ?

jmchilton:
> So my interpretation is that when we flush only certain objects, the "dirty" ? objects are reset. Does that make any sense ?

Yuk, I wasn't assuming it worked that way - I need to write a test case and verify that theory - that would be really gross and make what I was doing error prone.

mvdbeek:
I mean, this isn't what's documented in the sqlalchemy docs either ... http://docs.sqlalchemy.org/en/latest/orm/session_api.html?highlight=flush#sqlalchemy.orm.session.Session.flush

jmchilton:
Thanks for the fix!

PR #6632, opened 2018-08-23
Testing framework improvements

mvdbeek:
Couple of things I pulled out of https://github.com/galaxyproject/galaxy/pull/6627 that make testing more convenient

### DIFF
--- a/lib/galaxy/jobs/output_checker.py
+++ b/lib/galaxy/jobs/output_checker.py
@@ -15,6 +15,9 @@ DETECTED_JOB_STATE = Bunch(
 )
 
 
+ERROR_PEAK = 2000
+
+
 def check_output(tool, stdout, stderr, tool_exit_code, job):
     """
     Check the output of a tool - given the stdout, stderr, and the tool's
@@ -130,11 +133,14 @@ def check_output(tool, stdout, stderr, tool_exit_code, job):
             # log.debug( "Tool did not define exit code or stdio handling; "
             #          + "checking stderr for success" )
             if stderr:
-                peak = stderr[0:250]
-                log.debug("Tool produced standard error failing job - [%s]" % peak)
                 state = DETECTED_JOB_STATE.GENERIC_ERROR
             else:
                 state = DETECTED_JOB_STATE.OK
+
+        if DETECTED_JOB_STATE != DETECTED_JOB_STATE.OK and stderr:
+            if stderr:
+                peak = stderr[0:ERROR_PEAK]
+                log.debug("job failed, standard error is - [%s]" % peak)
 
     # On any exception, return True.
     except Exception:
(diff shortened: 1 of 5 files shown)


test/base/driver_util.py
```diff
@@ -172,6 +172,9 @@ def setup_galaxy_config(
     galaxy_data_manager_data_path = tempfile.mkdtemp(prefix='data_manager_tool-data', dir=tmpdir)
 
     tool_conf = os.environ.get('GALAXY_TEST_TOOL_CONF', default_tool_conf)
+    conda_auto_install = conda_auto_install or os.environ.get('GALAXY_TEST_CONDA_AUTO_INSTALL', False)
+    conda_auto_init = conda_auto_init or os.environ.get('GALAXY_TEST_CONDA_AUTO_INIT', False)
+    conda_prefix = os.environ.get('GALAXY_TEST_CONDA_PREFIX', None)
```
nsoranzo, line 177: `, None` is not needed.

test/base/driver_util.py
```diff
@@ -236,6 +240,8 @@ def setup_galaxy_config(
         # Used by shed's twill dependency stuff - todo read from
         # Galaxy's config API.
         os.environ["GALAXY_TEST_TOOL_DEPENDENCY_DIR"] = tool_dependency_dir
+    if conda_prefix:
+        config['conda_prefix'] = conda_prefix
```
nsoranzo, line 244: This 2 lines seem to not be necessary, this is already set at line 198.

lib/galaxy/jobs/output_checker.py
```diff
@@ -130,12 +133,15 @@ def check_output(tool, stdout, stderr, tool_exit_code, job):
...
             #          + "checking stderr for success" )
             if stderr:
-                peak = stderr[0:250]
-                log.debug("Tool produced standard error failing job - [%s]" % peak)
                 state = DETECTED_JOB_STATE.GENERIC_ERROR
             else:
                 state = DETECTED_JOB_STATE.OK
 
+        if DETECTED_JOB_STATE != DETECTED_JOB_STATE.OK and stderr:
+            if stderr:
```
bgruening, line 141: Is this additional if needed?

mvdbeek: Depends what stderr is. I suppose it might be an empty string, in which case we'd produce a meaningless log message. If it's actually `None` it'd fail at `stderr[0:ERROR_PEAK]`

test/base/driver_util.py
```diff
@@ -172,6 +172,9 @@ def setup_galaxy_config(
     galaxy_data_manager_data_path = tempfile.mkdtemp(prefix='data_manager_tool-data', dir=tmpdir)
 
     tool_conf = os.environ.get('GALAXY_TEST_TOOL_CONF', default_tool_conf)
+    conda_auto_install = conda_auto_install or os.environ.get('GALAXY_TEST_CONDA_AUTO_INSTALL', False)
+    conda_auto_init = conda_auto_init or os.environ.get('GALAXY_TEST_CONDA_AUTO_INIT', False)
```
nsoranzo, line 176: I think these should be:
```python
    conda_auto_install = os.environ.get('GALAXY_TEST_CONDA_AUTO_INSTALL', conda_auto_install)
    conda_auto_init = os.environ.get('GALAXY_TEST_CONDA_AUTO_INIT', conda_auto_init)
```

mvdbeek: you're right, I'll follow up on this

PR #6636, opened 2018-08-24
Fix for workflow extraction cleanup routine.

blankenberg:
Workflow extraction would be incorrect if you had parameters ordered and named as e.g. "somename" and "somename_2" in the input root. When these parameters are both dataset inputs, this would cause a server error.

### DIFF
--- a/lib/galaxy/workflow/extract.py
+++ b/lib/galaxy/workflow/extract.py
@@ -322,6 +322,7 @@ def __cleanup_param_values(inputs, values):
     if 'dbkey' in values:
         del values['dbkey']
     root_values = values
+    root_input_keys = inputs.keys()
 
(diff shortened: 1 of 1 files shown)


jmchilton:
Just noticed the branch name - 11yowffix - so I guess that comment "FIXME: MUST REMOVE SOON" meant sometime in the next couple decades.

martenson:
So we are leaving that comment in? :)

blankenberg:
This commit is definitely a patch on a patch, and the comment is still valid, probably. A look at solving/confirming the underlying issues (pollution, etc.) and removing the need for this cleanup, should happen 'soon'.

jmchilton:
I definitely wasn't implying I expected a more comprehensive fix or anything - I should have put a lol in there to make my tone clear.

PR #6637, opened 2018-08-25
Add conda to PATH when executing conda commands

mvdbeek:
This helps with errors like

```
ModuleNotFoundError: No module named 'conda.cli'
```

if only referencing the conda binary. This fixes local
conda when using the testing framework.

I'm not quite sure why this only affects local framework tests, but it also doesn't seem to harm anything.


nsoranzo:
How to reproduce the issue? Which version of conda?

mvdbeek:
`sh run_tests.sh -framework` on dev, so 4.3.33

mvdbeek:
So now I can't reproduce this anymore, but it has been plaguing me ever since we introduced conda into the framework tests. I'll look some more to figure out what's going on here.

PR #6638, opened 2018-08-25
Move tests to pytest setup

mvdbeek:
This moves unit, framework, data manager tools, migrated tools to use pytest, and includes some fixes necessary for tests to pass reliably.
Also contains some enhancements for using pytest to run tool tests:
  - We mark tools that should be skipped instead of not collecting them at all
  - We mark tools or data managers tools, so that they can be selected using the `-m` switch
And finally I've added an additional test target for tools in tool_conf.xml.sample and tool_conf.xml.main (And I've fixed those tools in https://github.com/galaxyproject/galaxy/pull/6627)

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -273,6 +273,7 @@ DOCKER_DEFAULT_IMAGE='galaxy/testing-base:18.09.0'
 
 test_script="./scripts/functional_tests.py"
 report_file="run_functional_tests.html"
+coverage_arg=""
 xunit_report_file=""
 structured_data_report_file=""
 with_framework_test_tools_arg=""
@@ -368,6 +369,7 @@ do
               api_script="./test/api"
               shift 1
           fi
+          coverage_file="api_coverage.xml"
           ;;
       -selenium|--selenium)
           with_framework_test_tools_arg="-with_framework_test_tools"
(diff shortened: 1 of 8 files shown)


run_tests.sh
```diff
@@ -627,12 +649,26 @@ if [ -n "$structured_data_report_file" ]; then
 else
     structured_data_args=""
 fi
-if [ -n "$with_framework_test_tools_arg" ]; then
-    GALAXY_TEST_TOOL_CONF="config/tool_conf.xml.sample,test/functional/tools/samples_tool_conf.xml"
+if [ "$with_framework_test_tools_arg" ]; then
```
nsoranzo: I think you can simplify the code by getting rid of `with_framework_test_tools_arg` and just set `GALAXY_TEST_TOOL_CONF` directly in the `case` above.

jmchilton:
It works with Planemo?

mvdbeek:
Yes, I think this is still passing through test/functional/test_toolbox.py, which shouldn't be affected by these changes.

jmchilton:
And the custom noise plugin that generates the json blob planemo consumes? I don’t get how this can work, but I’ll trust you and we can find new solutions if this crop up. Thanks for the progress.

PR #6639, opened 2018-08-26
[18.05] row ids that are all numeric become ints...

bwlang:
Since the "F" prefix is checked with substring, row_id must always be a string


martenson:
@bwlang good catch, I have never seen a numeric id myself yet. Thank you for the fix.

As of the backport to 18.05 - Given 18.09 should be out within a month, and that some static-building dependencies have changed their behaviour (which is causing this PR to have 235 files changed) I would like to avoid this sizable PR and just fix it in dev instead.

Would that be fine @bwlang @mvdbeek ?

mvdbeek:
I don't know, if it's broken we should fix it IMHO, unfortunately most deployers like to play it "safe" by staying behind.

bwlang:
I'll leave the wait vs now decision up to you - i doubt many people get burned by it and I've worked around it locally (thanks to @mvdbeek's static generation for - still not sure why that doesn't work locally for me). Hmm -what's the probability of generating an all numeric id of n characters - might be a fun  question to work out ;)  for one digit it's 10/36 right? so is that (10/36)^n ??

martenson:
@mvdbeek Feel free to merge.

martenson (approved):

mvdbeek:
We've had someone at the GCC who also saw all numeric IDs and it was causing problems when we had to guess if the ID was encoded or not, so it's not that unlikely. 
Also they are hex values, so chance of numeric is (10/16)^n, and who knows what biases we have in our hashing scheme.

Issue #6334, opened 2018-06-13
Get Data > EBI SRA is assigning fastqsanger to fastqsanger.gz output


jennaj:
Causes downstream tool problems such as this one: https://github.com/galaxyproject/usegalaxy-playbook/issues/144

ping @davebx @martenson

jennaj:
**Workaround until fixed:**

Reassign the datatype metadata to be `fastqsanger.gz` for your fastq inputs. The data is compressed but is not being labeled that way -- we'll be fixing that. Please follow this ticket for updates.

How to change metadata is in the Galaxy FAQs: https://galaxyproject.org/support/

 - How do I find, adjust, and/or correct metadata? https://galaxyproject.org/support/metadata/

Or, if your accession is at NCBI, the tool **NCBI SRA Tools > Download and Extract Reads in FASTA/Q format from NCBI SRA** can be used instead. That tool sets the metadata correctly depending on the output datatype selected on the form.

jennaj:
Duplicate with more scope: https://github.com/galaxyproject/galaxy/issues/6472

jennaj:
Ping again @natefoo @davebx @martenson 

This is still an issue and _triggers errors with any downstream tool used_ -- until the user updates to the correct datatype themselves. Most don't know what is going wrong with those errors, so this is coming up often still in bug reports/biostars questions.

Is there a priority/plan for the fix?

jennaj:
Another bug with the tool: **if the data is really in "fastqcssanger.gz" format, it is assigned "fastqsanger"**. The datatype "fastqsanger.gz" cannot be automatically applied to all outputs from this source.  Example: SRR033795

**Fastq Groomer** will convert correctly if the datatype is first changed to fastqcssanger.gz. Manipulate Fastq will also do the conversion. Test history: https://usegalaxy.org/u/jen/h/test-history-ebi-sra-solid-fastq

martenson:
jen:
>the issue is that 1) fastqsanger.gz is given the fastqsanger datatype when imported to Galaxy 2) related but less important, fastqcssanger.gz is given fastqsanger. so the user has to change to fastqcssanger.gz, then use the groomer
If only the first was fixed that will address all but one reported problem. item 1 is reported daily (again today already)

dan:
>It seems like the real issues are 1.) compressed datatypes are uncompressed for sniffing/detection and ext assignment, but are then finalized in their compressed form. 2.) the fastqsanger sniffer is not correct

jmchilton
[...]

PR #6641, opened 2018-08-27
document config for sniffing compressed files, enable by default

martenson:
Me and @natefoo think that it is time to switch the default of this config option to True. At least for Main.

In addition I suspect there is a bug in data_source tools that prevent/avoid uncompressing uploaded archives and thus causing issues like https://github.com/galaxyproject/galaxy/issues/6334.

ping @jmchilton

### DIFF
--- a/test/api/test_tools_upload.py
+++ b/test/api/test_tools_upload.py
@@ -9,6 +9,7 @@ from base.constants import (
 from base.populators import (
     DatasetPopulator,
     skip_without_datatype,
+    uses_test_history,
 )
 
 from galaxy.tools.verify.test_data import TestDataResolver
@@ -91,9 +92,14 @@ class ToolsUploadTestCase(api.ApiTestCase):
         assert details["state"] == "ok"
         assert details["file_ext"] == "fastqsanger.gz", details
 
-    def test_fetch_compressed_auto_decompress_target(self):
+    @uses_test_history(require_new=True)
+    def test_fetch_compressed_auto_decompress_target(self, history_id):
         fastqgz_path = TestDataResolver().get_filename("1.fastqsanger.gz")
-        details = self._upload_and_get_details(open(fastqgz_path, "rb"), api="fetch", assert_ok=False, auto_decompress=True)
+        details = self._upload_and_get_details(open(fastqgz_path, "rb"),
+                                               api="fetch",
+                                               history_id=history_id,
+                                               assert_ok=False,
+                                               auto_decompress=True)
         assert details["state"] == "ok"
         assert details["file_ext"] == "fastqsanger", details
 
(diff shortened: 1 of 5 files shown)


natefoo:
Just to walk it back slightly - I think we should do this, but I think we ought to check the proportion of jobs over time running tools with fastq inputs and see how many of those are tools that support native `fastqsanger.gz` inputs first. If the majority of such jobs are still using non-`fastqsanger.gz`-supporting tools, this would probably be a net loss for usegalaxy.org (but we should work quickly to get any remaining tools updated to supporting `fastqsanger.gz`).

martenson:
It could technically be a short run net loss (more space consumed because of the compressed archive presence in history after implicit conversion) but it is the path forward we want to promote I believe. 

It will motivate tool authors to write tools that accept compressed datatypes as first-class citizens and also improve our storage efficiency in the long run. 

>If the majority of such jobs are still using non-fastqsanger.gz-supporting tools

I wouldn't give this ratio too much importance, since we are now being inefficient (on Main) by unpacking all compressed formats during upload that are not specifically set to compressed datatype (and setting compressed datatype is rarely used I believe). By stopping this practice (with enabling sniffing of compressed datatypes) we will save space in cases that will never run implicit conversion afterwards. My guess is that this saved space outweighs the extra space consumed by preservation of the compressed copies after implicit conversion (for the analyses that use it).

lib/galaxy/datatypes/registry.py
```diff
@@ -141,7 +141,7 @@ def __import_module(full_path, datatype_module, datatype_class_name):
                 auto_compressed_types = galaxy.util.listify(elem.get('auto_compressed_types', ''))
                 sniff_compressed_types = galaxy.util.string_as_bool_or_none(elem.get("sniff_compressed_types", "None"))
                 if sniff_compressed_types is None:
-                    sniff_compressed_types = getattr(self.config, "sniff_compressed_dynamic_datatypes_default", True)
+                    sniff_compressed_types = getattr(self.config, "sniff_compressed_dynamic_datatypes", True)
```
jmchilton: This can be configured/overridden on a per-datatype basis in the datatypes_conf.xml - so I think the old config value name is more appropriate.

martenson: I removed the renaming and added the note about configuration to the description. Thanks for pointing this out.

jmchilton:
Did you mess up your gitfu here - I don't see the original changes anymore. I'd use git reflog to find the old head if you no longer have them in a named branch.

martenson:
@jmchilton this is now a documentation PR only. I see it fine I think.

martenson:
I have added a commit that switches the default for sniffing compressed to `True`.

jmchilton:
I was super confused and remember the original PR being bigger and switching the default, I'm... old I guess.

Thanks for switching the default - I think all contemporary tools have been written with compressed files in mind and this will definitely save us space - but regardless you points about the ratio and being future facing are absolutely true regardless of that debate.

nsoranzo:
@galaxybot test py3

nsoranzo:
@martenson API test failures seem legit.

martenson:
I have added commit with adjustments to the compressed upload API tests. I see strange results locally so let's see what jenkins sees.

test/api/test_tools_upload.py
```diff
@@ -102,7 +101,7 @@ def test_upload_decompresses_auto_by_default(self):
         bedgz_path = TestDataResolver().get_filename("4.bed.gz")
```
nsoranzo: This test also need a name change.

martenson:
Tests now pass for me locally. Hopefully they make sense.

martenson:
the travis build is all green, not sure what caused it to not update here

jennaj:
This change will help for issues with fastqsanger data, however, if the data is fastqcssanger, probably not. That type of data is being sniffed incorrectly in two ways (.gz compression not recognized **and** base fastq type is wrong) -- not sure at what step that is done. 

**In short, EBI SRA "Send to Galaxy" data cannot be assumed to be fastqsanger.** Example and details here: https://github.com/galaxyproject/galaxy/issues/6334#issuecomment-419205445

martenson:
@jennaj I don't think this affects the current PR. Please provide examples of these wrongly sniffed `fastqcssanger` files and create a new issue for it.

nsoranzo (approved):

PR #6565, opened 2018-08-02
Setup py3 test job

mvdbeek:
@jmchilton if the Dockerfile looks OK to you, could you build a new image ?

xref. #1715

PR #6642, opened 2018-08-27
Allow running scripts/galaxy-main without Galaxy on sys.path

natefoo:
Fixes an issue that arose in #6565 from adding a galaxy lib import.

Also fix a bug where attributes in the config builder were accessed before they were set.

### DIFF
--- a/scripts/galaxy-main
+++ b/scripts/galaxy-main
@@ -170,6 +170,8 @@ class GalaxyConfigBuilder(object):
     """
 
     def __init__(self, args=None, **kwds):
+        self.config_file = None
+        self.config_section = None
         self.app_name = kwds.get("app") or (args and args.app) or DEFAULT_CONFIG_SECTION
         config_file = kwds.get("config_file", None) or (args and args.config_file)
         # If given app_conf_path - use that - else we need to ensure we have a
--- a/scripts/galaxy-main
+++ b/scripts/galaxy-main
@@ -34,11 +34,6 @@ from logging.config import fileConfig
 from six.moves.configparser import ConfigParser
 
 try:
-    from galaxy.util import unicodify
-except ImportError:
-    unicodify = None
-
-try:
     from daemonize import Daemonize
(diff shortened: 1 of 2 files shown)


scripts/galaxy-main
```diff
@@ -33,7 +33,10 @@ from logging.config import fileConfig
 
 from six.moves.configparser import ConfigParser
 
-from galaxy.util import unicodify
+try:
+    from galaxy.util import unicodify
+except ImportError:
+    unicodify = None
```
nsoranzo: Can't we simply add `sys.path.insert(1, os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir, 'lib')))` before the import, as we do in most other scripts?

natefoo: I'm not sure - we do actually add it to `sys.path` already, but not until `main()`. I don't know why that's not done at the top, but the fact that `lib/galaxy/main.py` is a symlink to this may be the reason? A look through blame shows that it was implemented that way from the start (by @jmchilton).

nsoranzo: We could resolve the symlink if needed, I think.

jmchilton: Copied from Pulsar and where it is meant to be pip installable possibly - pip installed stuff should not be mucking with the Python path like that unconditionally? We need to do something like this with all the scripts - but we can do that all at once at some point in the future? I'm indifferent on what is done here for this script.

natefoo: Some day Galaxy will be installable and then we won't need to do any path muckery. Until then...

nsoranzo: @natefoo I've added a commit to your branch to simplify this, let me know what you think. I haven't had time to add a test for this yet.

nsoranzo: I've also added an integration test now, which passes for me also under Python 3.

mvdbeek (approved):

PR #6644, opened 2018-08-28
Small Optimizations to Adding Datasets to History

jmchilton:
Refactored out of https://github.com/galaxyproject/galaxy/pull/6563, the timings I had on that PR was that eliminating these queries reduced the time to execute 10 jobs each creating 10 datasets by about 10%. Also eliminate an ancient note about optimization stuff that is almost certainly no longer correct.

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -1455,7 +1455,9 @@ class History(HasTags, Dictifiable, UsesAnnotations, HasName):
         set_genome = genome_build not in [None, '?']
         for i, dataset in enumerate(datasets):
             dataset.hid = base_hid + i
-            dataset.history = self
+            # Don't let SA manage this.
+            delattr(dataset, "history")
+            dataset.history_id = cached_id(self)
             if set_genome:
                 self.genome_build = genome_build
         for dataset in datasets:
--- a/lib/galaxy/model/mapping.py
+++ b/lib/galaxy/model/mapping.py
@@ -2541,10 +2541,10 @@ def db_next_hid(self, n=1):
     trans = conn.begin()
     try:
         if "postgres" not in session.bind.dialect.name:
-            next_hid = select([table.c.hid_counter], table.c.id == self.id, for_update=True).scalar()
+            next_hid = select([table.c.hid_counter], table.c.id == model.cached_id(self), for_update=True).scalar()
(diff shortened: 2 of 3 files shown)


lib/galaxy/model/__init__.py
```diff
@@ -1455,7 +1455,9 @@ def __add_datasets_optimized(self, datasets, genome_build=None):
         set_genome = genome_build not in [None, '?']
         for i, dataset in enumerate(datasets):
             dataset.hid = base_hid + i
-            dataset.history = self
+            # Don't let SA manage this.
+            delattr(dataset, "history")
```
mvdbeek, line 1459: Do you know why this speeds up things ?
Is it possible this prevents an HDA from being marked dirty, which results in a HDA copy ?
I've been looking at unnecessary HDA version bumps, I suspect something like
```
--- a/lib/galaxy/model/base.py
+++ b/lib/galaxy/model/base.py
@@ -7,8 +7,12 @@ from inspect import (
     isclass
 )

-from sqlalchemy import event
+from sqlalchemy import (
+    event,
+    inspect
+)
 from sqlalchemy.orm import (
+    class_mapper,
     scoped_session,
     sessionmaker
 )
@@ -52,7 +56,20 @@ class ModelMapping(Bunch):
 def versioned_objects(iter):
     for obj in iter:
         if hasattr(obj, '__create_version__'):
-            yield obj
+            if object_has_changed(obj):
+                yield obj
+
+
+def object_has_changed(myobj):
+    """We check wether the objbect itself has changed."""
+
+    inspr = inspect(myobj)
+    attrs = class_mapper(myobj.__class__).column_attrs  # exclude relationships
+    for attr in attrs:
+        hist = getattr(inspr.attrs, attr.key).history
+        if hist.has_changes():
+            return True
+    return False
```
might result in a similar speedup ?

jmchilton: That isn't a speed up - the speed up is setting ``history_id`` instead of ``history`` - the way I do other places (https://github.com/galaxyproject/galaxy/pull/6524/files#diff-bfb557a99c1f7d646d4968d8d680b885R645). Usually that trick just seems to work, here it didn't because the dataset already had dataset.history = None set on it in its constructor. So SA was opting to respect the null history instead of using the history_id that I was trying to set.

dannon:
Toolshed test that failed is unrelated to this.

Issue #6617, opened 2018-08-21
Saved Pages malformed id in usegalaxy.org


toniher:
Sorry if this is not the proper location to report it, since yesterday's migration, it seems that many pages in https://usegalaxy.org/pages/list_published are getting a malformed id error.
As example I created one a few days ago https://usegalaxy.org/u/toniher/p/deeptools-pergola , I can continue editing it, but I cannot view its content.
Let me know if I should report it somewhere else.

Thanks!

dannon:
@toniher Thank you for reporting this, I'll take a look!

dannon:
So the problem here is that pages internally save the contents hashes instead of raw ids.  I'm working on ideas for fixes, but it's going to take a little work.

toniher:
It seems it's already working! Thanks for solving this!

dannon:
@toniher It is!  Sorry I forgot to update this issue, we took care of it last week.

PR #6647, opened 2018-08-28
[18.05] Pages - fixes for encoded ids in the database

dannon:
xref #6617, when we updated the secret for usegalaxy.org, we broke all the internal hashed identifiers that pages were saving.

This PR converts the pages to read both encoded and unencoded ids out of the database (for backwards compatibility), display to the user using encoded ids (both in view and edit modes), but always save using unencoded ids.  It also actually validates page contents on save, now, including items.

It additionally includes a script for admins to fix 'mangled' pages, where the secret has been changed and the page hashes are no longer valid.  It uses beautifulsoup to quickly rewrite identifiers in an encoded form.


dannon:
Looks like the contents pruning was too aggressive and we actually need some of that placeholder stuff for good editor display.  Reworking a tad.

dannon:
I added a bonus bugfix for some busted javascript that was causing embed insertions like this, which failed to correctly pull in the item names:

image

dannon:
@natefoo Were you able to test this?

natefoo:
Tested and run on Main, thanks @dannon!

martenson:
post mortem 👍

PR #6655, opened 2018-08-29
Adding 4 new GROMACS datatypes: cpt, xtc, top, and gro

simonbray:
Some MD datatypes - @bgruening could you have a look please?


lib/galaxy/datatypes/tabular.py
```diff
@@ -1161,3 +1161,47 @@ def get_chunk(self, trans, dataset, chunk):
...
+
+@dataproviders.decorators.has_dataproviders
+@build_sniff_from_prefix
+class Gro(Tabular):
+    file_ext = 'gro'
+
+    def __init__(self, **kwd):
+        """Initialize gro datatype"""
+        super(Gro, self).__init__(**kwd)
+
```
bgruening: here is an unneeded linebreak I think

lib/galaxy/datatypes/tabular.py
```diff
@@ -1161,3 +1161,47 @@ def get_chunk(self, trans, dataset, chunk):
         ck_data_body = re.sub('[ ]+', '\t', ck_data_body)
 
         return dumps({'ck_data': util.unicodify(ck_data_header + "\n" + ck_data_body), 'ck_index': ck_index + 1})
+
+
+@dataproviders.decorators.has_dataproviders
```
bgruening: the dataprovider should probably be removed

lib/galaxy/datatypes/tabular.py
```diff
@@ -1161,3 +1161,47 @@ def get_chunk(self, trans, dataset, chunk):
...
+    def __init__(self, **kwd):
+        """Initialize gro datatype"""
+        super(Gro, self).__init__(**kwd)
+
+
+    def display_peek(self, dataset):
+        """Returns formated html of peek"""
+        return self.make_html_table(dataset)
+
+    def sniff_prefix(self, file_prefix):
```
bgruening: This sniffer seems to me not to strict enough. If we can not reliably sniff a GRO files we should not do it and just subclass the GRO datatype from tabular in the datatypes_conf.xml file. So this python class is not needed.

lib/galaxy/datatypes/text.py
```diff
@@ -573,3 +573,37 @@ def sniff_prefix(self, file_prefix):
...
+        else:
+            dataset.peek = 'file does not exist'
+            dataset.blurb = 'file purged from disk'
+
+    def get_mime(self):
+        """Returns the mime type of the datatype"""
+        return 'text/top'
+
+    def sniff_prefix(self, file_prefix):
+        """
```
bgruening: This one as well. The sniffer seems to be not strong enough. Everything with `atoms` will be recognized as Top file.
Just subclass from text in the datatypes_conf.xml file instead.
The downside is that users during upload need to specify the file type, which I think is ok in this case.

lib/galaxy/datatypes/binary.py
```diff
@@ -776,7 +776,88 @@ def display_peek(self, dataset):
...
+    True
+    >>> fname = get_test_fname('em.trr')
+    >>> Xtc().sniff(fname)
+    False
+    """
+    file_ext = "xtc"
+
+    def __init__(self, **kwd):
+        Binary.__init__(self, **kwd)
+        self._magic_number = 1995
```
bgruening: Could you just add a comment with a link to a documentation where you have found this magic number?
Some for the other datatype, thanks :)

bgruening (commented):
Thanks @simonbray!

simonbray:
Thanks @bgruening - working on the fixes

blankenberg:
Can we shrink the datatype test files, preferably as minimal as possible, but to less than a couple of kb each, and squash?

lib/galaxy/datatypes/binary.py
```diff
@@ -776,7 +776,87 @@ def display_peek(self, dataset):
...
+    >>> fname = get_test_fname('md.xtc')
+    >>> Xtc().sniff(fname)
+    True
+    >>> fname = get_test_fname('md.trr')
+    >>> Xtc().sniff(fname)
+    False
+    """
+    file_ext = "xtc"
+
+    def __init__(self, **kwd):
```
mvdbeek: Can we reduce the boilerplate by making the magic number and the peek text a class attribute ?
Those classes all follow the same pattern, so then you'd just have a base class that implements sniff and set_peek, and the child classes just set the magic number and peek text

simonbray: Thanks @mvdbeek, I've made this change in the most recent commit.

lib/galaxy/datatypes/binary.py
```diff
@@ -736,26 +736,18 @@ def display_peek(self, dataset):
...
-    True
-    >>> fname = get_test_fname('interval.interval')
-    >>> Trr().sniff(fname)
-    False
+    Base class for GROMACS binary files - xtc, trr, cpt
     """
-    file_ext = "trr"
 
-    def __init__(self, **kwd):
+    def __init__(self, magic_number, file_ext, **kwd):
```
mvdbeek: That's about it, but you can just make `magic_number` a class attribute that the child classes overwrite, and then you don't need any `__init__` at all.

lib/galaxy/datatypes/binary.py
```diff
@@ -776,7 +768,64 @@ def display_peek(self, dataset):
...
+
+    >>> from galaxy.datatypes.sniff import get_test_fname
+    >>> fname = get_test_fname('md.trr')
+    >>> Trr().sniff(fname)
+    True
+    >>> fname = get_test_fname('interval.interval')
+    >>> Trr().sniff(fname)
+    False
+    """
+    file_ext = "trr"
```
mvdbeek: So far instance here you put `magic_number = 1993` and then you don't need any `def __init__()`

lib/galaxy/datatypes/binary.py
```diff
@@ -736,37 +736,30 @@ def display_peek(self, dataset):
...
-    >>> Trr().sniff(fname)
-    True
-    >>> fname = get_test_fname('interval.interval')
-    >>> Trr().sniff(fname)
-    False
+    Base class for GROMACS binary files - xtc, trr, cpt
     """
-    file_ext = "trr"
+
+    magic_number = 0  # variables to be overwritten in the child class
```
bgruening: None might be a better default.

lib/galaxy/datatypes/binary.py
```diff
@@ -736,37 +736,30 @@ def display_peek(self, dataset):
...
-    True
-    >>> fname = get_test_fname('interval.interval')
-    >>> Trr().sniff(fname)
-    False
+    Base class for GROMACS binary files - xtc, trr, cpt
     """
-    file_ext = "trr"
+
+    magic_number = None  # variables to be overwritten in the child class
+    file_ext = ""
```
mvdbeek: You don't need the dunder init here either.

bgruening (approved):

mvdbeek:
We need to squash the commit that introduced the test files with the commit that reduced the size, @simonbray can you do that or should I do that for you ?

(Or just squash them all using the github interface, if you don't care about the inidividual commits)

simonbray:
@mvdbeek, I squashed the commits together, but I fear I've messed up the history. Sorry - I should have left it to you.

mvdbeek:
No worries, you almost had it, all that was needed is fetching the current galaxy remote (if that is origin you'd do `git fetch origin`) and then `git rebase origin/dev`. I've done that now. Thanks a lot for the contribution!

mvdbeek:
@galaxybot test this

blankenberg:
Bad Merge/squash? Both classes for Top and Gro are missing, and the final commit doesn't have text.py and tabular.py shown as changed in the final PR diff.

```
galaxy.datatypes.registry ERROR 2018-09-11 17:45:04,937 [p:25538,w:0,m:0] [MainThread] Error importing datatype module galaxy.datatypes.text
  File "lib/galaxy/datatypes/registry.py", line 235, in load_datatypes
    datatype_class = getattr(module, datatype_class_name)
AttributeError: 'module' object has no attribute 'Top'
galaxy.datatypes.registry DEBUG 2018-09-11 17:45:04,938 [p:25538,w:0,m:0] [MainThread] Retrieved datatype module galaxy.datatypes.binary:Xtc from the datatype registry.
galaxy.datatypes.registry DEBUG 2018-09-11 17:45:04,938 [p:25538,w:0,m:0] [MainThread] Retrieved datatype module galaxy.datatypes.binary:Cpt from the datatype registry.
galaxy.datatypes.registry ERROR 2018-09-11 17:45:04,938 [p:25538,w:0,m:0] [MainThread] Error importing datatype module galaxy.datatypes.tabular
  File "lib/galaxy/datatypes/registry.py", line 235, in load_datatypes
    datatype_class = getattr(module, datatype_class_name)
AttributeError: 'module' object has no attribute 'Gro'
g
```

mvdbeek:
I think they were removed in https://github.com/galaxyproject/galaxy/pull/6655/commits/56d55399ebe2ecd614a7ba5a146739f58e917541, we probably just need to make them tabular subclasses in the datatypes config.

PR #6657, opened 2018-08-29
Added support for setting snpeff_version of a snpeffdb manually uploaded to a user's history.

hepcat72:
Prior to this change, the snpeff_version was always set to SnpEff4.0 because the only files the version check was made on were those in the data_dir that houses galaxy's snpeff databases. Thus, if a user uploaded a 4.3 .bin file, the version would be defaulted to 4.0 and not show up in the history dropdown of the snpeff eff tool.

Files affected: lib/galaxy/datatypes/text.py


hepcat72:
Just realized I accidentally developed on the last stable release. I'll switch to dev and copy changes over & re-test.

PR #6658, opened 2018-08-30
Support for manual snpeffdb history uploads

hepcat72:
The snpeffdb snpeff_version was being set to SnpEff4.0 regardless of the version in the file.  The code assumed that the file was in the galaxy maintained data_dir location and not in the user's history.  Also had to add processing of the text file because it gets decompressed in the user's history.

Now, if a user uploads a snpEffectPredictor.bin file, it gets a snpeff version that was parsed from the file.

### DIFF
--- a/lib/galaxy/datatypes/text.py
+++ b/lib/galaxy/datatypes/text.py
@@ -415,7 +415,16 @@ class SnpEffDb( Text ):
                 snpeff_version = m.groups()[0] + m.groups()[1]
             fh.close()
         except:
-            pass
+            try:
+                # In case this was a decompressed file manually uploaded to a user's history
+                with open(path, 'r') as fh:
+                    buf = fh.read(100)
+                    lines = buf.splitlines()
+                    m = re.match('^(SnpEff)\s+(\d+\.\d+).*$', lines[0].strip())
+                    if m:
+                        snpeff_version = m.groups()[0] + m.groups()[1]
+            except Exception:
+                pass
         return snpeff_version
 
     def set_meta( self, dataset, **kwd ):
@@ -461,6 +470,11 @@ class SnpEffDb( Text ):
                         fh.write("regulations: %s\n" % ','.join(regulations))
             except:
                 pass
+        # If this was a file uploaded manually by the user, get the snpeff version from it instead
+        if os.path.isfile(dataset.file_name):
+            snpeff_version = self.getSnpeffVersionFromFile(dataset.file_name)
+            if snpeff_version:
+                dataset.metadata.snpeff_version = snpeff_version
 
 
 class SnpSiftDbNSFP( Text ):
--- a/lib/galaxy/datatypes/text.py
+++ b/lib/galaxy/datatypes/text.py
@@ -413,7 +413,6 @@ class SnpEffDb(Text):
     def getSnpeffVersionFromFile(self, path):
         snpeff_version = None
         try:
-<<<<<<< HEAD
             with gzip.open(path, 'rb') as fh:
                 buf = fh.read(100)
                 lines = buf.splitlines()
@@ -421,16 +420,6 @@ class SnpEffDb(Text):
                 if m:
                     snpeff_version = m.groups()[0] + m.groups()[1]
         except Exception:
-            pass
-=======
-            fh = gzip.open(path, 'rb')
-            buf = fh.read(100)
-            lines = buf.splitlines()
-            m = re.match('^(SnpEff)\s+(\d+\.\d+).*$', lines[0].strip())
-            if m:
-                snpeff_version = m.groups()[0] + m.groups()[1]
-            fh.close()
-        except:
             try:
                 # In case this was a decompressed file manually uploaded to a user's history
                 with open(path, 'r') as fh:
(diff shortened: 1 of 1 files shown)


martenson:
@galaxybot test this

nsoranzo:
@hepcat72 Can you share a link to a file that exhibits the problem?

hepcat72:
Any SnpEff database file that’s not version 4.0 will demonstrate the issue, but I was testing with the one for Bacillis subtilis subsp str 168 I downloaded from the 4.3 databases on sourceforge. I’m not in the office at the moment, but when I get in, I’ll paste a link.

hepcat72:
Incidentally, the loop in set_meta that was walking through data_dir was never calling the ‘get version from file’ method. I had put a debug print both in the loop and in the method, but since the code was running in a thread, those messages weren't going to the log.  I eventually learned that those messages would show up in the downloaded file’s metadata (as viewed on galaxy) because once I added the call sending the file located in the history, I started seeing the debug message I'd put in the "...from file" method and was getting an error because the file wasn't gzipped. Then I added the text file (not gzip) processing and the version started updating.

hepcat72:
OK, here's the file I was testing with:

https://www.dropbox.com/s/f94l0atwsfmnxu4/snpEffectPredictor.bin?dl=0

hepcat72:
I just realized that "genome_version" is not set in this use-case either. Additionally, when you run `SnpEff download` and supply the genome version as "Bacillus_subtilis_subsp_subtilis_str_168", the resultant version in the metadata is "Bacillus_subtilis_subsp_subtilis_str_168_gca_000789275", which is a separate database.  I intend to address this and update the PR.

lib/galaxy/datatypes/text.py
```diff
@@ -420,7 +420,16 @@ def getSnpeffVersionFromFile(self, path):
                 if m:
                     snpeff_version = m.groups()[0] + m.groups()[1]
         except Exception:
-            pass
+            try:
+                # In case this was a decompressed file manually uploaded to a user's history
+                with open(path, 'r') as fh:
```
nsoranzo, line 425: You should be able to use `with compression_utils.get_fileobj(dataset.file_name, compressed_formats=['gzip']) as fh:` at line 416, instead of adding this code block.

I'd also need to add the corresponding import, by changing:
```python
from galaxy.util import nice_size, string_as_bool
```
to
```python
from galaxy.util import (
    compression_utils,
    nice_size,
    string_as_bool
)
```
at the start of this file.

hepcat72: Lance and I were just talking over the over-arching issue and I'm thinking that this PR is the wrong way to go, and I intend to close it after I've had a chance to make sure I'm not missing something else.

So let me back up... Using snpeff on galaxy has never worked for me and I didn't remember why because it had been too long ago when I gave up trying to use it... but it turns out that the code in text.py that sets the metadata sets it incorrectly because it assumes that a downloaded snpeff database zip file contains 1 directory whose name is the genome version, but I work mostly with microbes that are always in zip files with multiple other genome directories.  The code in text.py ends up always setting the "last" genome version, which for me, was always wrong.

I'm still not sure why when you upload a snpEffectPredictor.bin file (set to datatype snpeffdb), it doesn't get any genome version or the default snpeff_version, but Lance pointed out that regardless of that, the file you get in your history via the snpeff download method versus the manual upload method are completely different.  The snpeff download method puts a 2-line file with snpeff and genome versions in it.  That's the file type specified by datatype 'snpeffdb'.  To assign 'snpeffdb' to a full-on real snpEffectPredictor.bin file is technically incorrect because galaxy elsewhere will treat it like the 2-line file. It only worked for me because I had mucked around with the snpeff.xml file when I couldn't figure out how to make things work how I expected them to.

So the real problem is that the set_meta method in text.py is setting the wrong genome_version metadata because the download includes multiple directories, but assumes only 1.  `snpeff download` is downloading the zip and expanding it to multiple directories and set_meta is marching through all of them and setting the genome_version at each step.  The last one to be processed is the one that sticks.

So, the real solution, that I intend to implement is to have `snpeff download` download to a temp directory and then move out the one for the genome version the user specified.  That makes the assumption made in set_meta in text.py correct.

As far as uploading individual snpEffectPredictor.bin file, the main reason a user might attempt that is to circumvent this whole problem in the first place.  True, they might have their own that they've built on the command line, but as far as that's concerned, it's too complicated to figure out how to make that fit into the current architecture, so I'm going to drop it.

One last thing... the new download on demand option appears to put the downloaded database into a snpeff default location (or whatever is in the snpeff config file - I don't know if galaxy manipulates that, but I presume it doesn't). There's no -dataDir option specified in that case in the snpeff eff tool, so presumably the database is placed somewhere in the conda environment, probably near where the snpeff jar file lives.  If that's a new & codified possibility, it seems like all the other storage and metadata efforts might be unnecessary.  You can take all that out and just let users specify the genome version and it will just work...  Either way, they either have to specify it in `snpeff eff`, `snpeff download`, or tell a galaxy admin the genome version.

nsoranzo: I also experimented a bit with the datasets created by `snpEff donwload` and I can confirm what you said above (apart from the download on demand option mentioned in the last paragraph, which I haven't looked into yet).
The issue was already reported more than a year ago in https://github.com/galaxyproject/tools-iuc/issues/1354.
I would also add that presently when a genome is extracted from a multi-genome bundle, `snpEff download` pollutes the dataset directory with all the other `snpEffectPredictor.bin` files for no apparent reason.

hepcat72:
I have fixed this issue via the snpeff download tool in the tools-iuc repo.  Though the user still cannot manually upload a snpeffdb file (because it doesn't get a correct snpeff_version and doesn't get a genome_version assigned in the metadata), that use-case will be unnecessary with the tool fix.

PR #6659, opened 2018-08-30
Fix get_fileobj() call in Eland.set_meta()

nsoranzo:
Broken in commit 59f44cf2be48abd4bed0f7c6ddf7cc0aee5ad1f5 .


mvdbeek (approved):

mvdbeek:
Thanks @nsoranzo, we'll also have to backport this, right ?

nsoranzo:
I thought about this, is Eland used a lot? I don't think there was any report of this. I could easily do it any way.

mvdbeek:
I did have to look up Eland ... I guess it's not necessary but you'd have my 👍 again ...

PR #6660, opened 2018-08-31
Upgrade CloudBridge to v1.0.1

VJalili:
This PR updates the cloudbridge version to latest current release. Since there has been interface change from `v0.3.3` to `v1.0.0`, this PR updates object store to adhere with latest interface changes.

### DIFF
--- a/lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
+++ b/lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
@@ -2,16 +2,16 @@
 --extra-index-url https://pypi.python.org/simple
 alabaster==0.7.11
 argh==0.26.2
-atomicwrites==1.1.5
-attrs==18.1.0
+atomicwrites==1.2.1; python_version != '3.3.*'
+attrs==18.2.0
 babel==2.6.0
-certifi==2018.8.13
+certifi==2018.8.24
 chardet==3.0.4
 commonmark==0.5.4
 docutils==0.14
 funcsigs==1.0.2; python_version < '3.3'
 idna==2.7
-imagesize==1.0.0
+imagesize==1.1.0; python_version != '3.3.*'
 jinja2==2.10
 lxml==4.2.4
 markupsafe==1.0
@@ -25,14 +25,14 @@ pathlib2==2.3.2; python_version < '3.6'
 pathtools==0.1.2
 pbr==4.2.0
 pluggy==0.7.1; python_version != '3.3.*'
-py==1.5.4; python_version != '3.3.*'
+py==1.6.0; python_version != '3.3.*'
 pygithub3==0.5.1; python_version < '3'
 pygments==2.2.0
 pyparsing==2.2.0
 pytest-html==1.19.0
 pytest-metadata==1.7.0
-pytest-pythonpath==0.7.2
-pytest==3.7.2
+pytest-pythonpath==0.7.3
+pytest==3.8.0
 pytz==2018.5
 pyyaml==3.13
 recommonmark==0.4.0
@@ -44,8 +44,8 @@ snowballstemmer==1.2.1
 sphinx-rtd-theme==0.4.1
 sphinx==1.6.7
 sphinxcontrib-websupport==1.1.0; python_version != '3.3.*'
-testfixtures==6.2.0
+testfixtures==6.3.0
 twill==0.9.1; python_version < '3'
-typing==3.6.4; python_version < '3.5'
+typing==3.6.6; python_version < '3.5'
 urllib3==1.23; python_version >= '2.6'
-watchdog==0.8.3
+watchdog==0.9.0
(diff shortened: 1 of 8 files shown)


lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
```diff
@@ -2,16 +2,16 @@
...
+atomicwrites==1.2.1; python_version != '3.3.*'
+attrs==18.2.0
 babel==2.6.0
-certifi==2018.8.13
+certifi==2018.8.24
 chardet==3.0.4
 commonmark==0.5.4
 docutils==0.14
-funcsigs==1.0.2; python_version < '3.3'
+funcsigs==1.0.2; python_version == '2.6'
```
nsoranzo: This (and others below) looks like a mistake from pipenv which should be automatically fixed by `lib/galaxy/dependencies/pipfiles/update.sh` (which is what `make update-dependencies` run).

VJalili: You may also noticed that running `make update-dependencies` has down-graded both cloudbridge and genomespace to their previous versions (those already on dev). Any thoughts?

nsoranzo: That's because `python-genomespaceclient` is pinned to `<1.0` in lib/galaxy/dependencies/pipfiles/default/Pipfile ,see https://github.com/galaxyproject/galaxy/pull/6668/files#diff-3c6aba54d3aab2581826ac1eef590665L72

VJalili: Thanks! then I think it might be better hold this PR till #6668 is updated and merged.

lib/galaxy/managers/cloud.py
```diff
@@ -203,7 +203,7 @@ def upload(self, trans, history_id, provider, bucket, objects, credentials, inpu
 
         connection = self._configure_provider(provider, credentials)
         try:
-            bucket_obj = connection.object_store.get(bucket)
+            bucket_obj = connection.storage.buckets.get(bucket)
```
nuwang: Just a minor nit, but it might be better to call this bucket instead of bucket_obj, just so it doesn't cause any confusion with BucketObject.

nuwang (commented):
+1 LGTM. Seems like a matter of regenerating the dependencies and merging after https://github.com/galaxyproject/galaxy/pull/6668

lib/galaxy/managers/cloud.py
VJalili: `bucket` is the variable that contains the name of bucket. Hence to avoid confusion I named the bucket object `bucket_obj`.

nsoranzo:
@VJalili Unfortunately the results of the rebase and the `make update-dependencies` don't look correct, I fixed them locally, but I cannot push to your branch. Can you allow collaborators to push to your fork?

lib/galaxy/managers/cloud.py
afgane: Perhaps renaming `bucket` to `bucket_name` and `bucket_obj` to `bucket` (or something other than `bucket_obj`)? `bucket_obj` feels very close to an _object_ in a bucket that can it easily cause confusion for a someone reading the code.

VJalili:
@nsoranzo Thanks for checking this. Could you please try pushing again?

nsoranzo:
@VJalili Worked this time, thanks! Can you check that the changes look correct to you?

VJalili:
@nsoranzo I am running this on a mac, and I am getting the following error: 

```shell
Collecting pyinotify==0.9.6 (from -r requirements.txt (line 106))
  Using cached https://files.pythonhosted.org/packages/e3/c0/fd5b18dde17c1249658521f69598f3252f11d9d7a980c5be8619970646e1/pyinotify-0.9.6.tar.gz
    Complete output from command python setup.py egg_info:
    inotify is not available on macosx-10.13-x86_64
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/hq/dpy7w16x2ll7fkzhr18hq1rc0000gn/T/pip-install-o37Z7d/pyinotify/
```
Maybe related to [this line](https://github.com/galaxyproject/galaxy/pull/6660/files#diff-aeb9dee0c2605f413ed70e295969d45aR106); any thoughts?

nsoranzo:
@VJalili Yep, it's a new requirement from cloudbridge dependency `oslo.log`, but the line should be:
```
pyinotify==0.9.6; sys_platform != "win32" and sys_platform != "darwin" and sys_platform != "sunos5"
```
I'll add a fix to your branch in a minute.

nsoranzo:
@VJalili Should be solved now.

VJalili:
@nsoranzo Thanks indeed! I pushed some refactoring. It should be good now.

VJalili:
@dannon Thanks for merging this!

PR #6661, opened 2018-09-03
Allow group tag selection also for multiple="true" inputs

mvdbeek:
One small adjustment to allow group tags for multiple="true" data inputs.


jmchilton (approved):

jmchilton:
Awesome, thanks!

PR #6664, opened 2018-09-04
Implement runtime parameters for subworkflows.

jmchilton:
Includes tests for singly nested workflows and parameters outside of conditionals. I think parameters in conditionals would work also, probably not doubly nested workflows - but we don't have any tests for those at this time at any level I don't think?

Implementation: When generating the form for subworkflow steps - the whole workflow is walked and checked for non-data runtime parameters. The resulting parameter inputs are wrapped using a proxy that modifies their to_dict to provide the workflow context. The API has been modified so that runtime parameters passed to subworkflow steps have their outer order_index chopped off and their inner index then remapped against the nested workflow. This also requires tracking more workflow inputs in the database - this a good thing for reproducibility and we were loosing some of this before.

### DIFF
--- a/lib/galaxy/workflow/modules.py
+++ b/lib/galaxy/workflow/modules.py
@@ -31,6 +31,7 @@ from galaxy.tools.parameters.basic import (
     BooleanToolParameter,
     DataCollectionToolParameter,
     DataToolParameter,
+    is_runtime_value,
     parameter_types,
     RuntimeValue,
     SelectToolParameter,
@@ -380,6 +381,40 @@ class SubWorkflowModule(WorkflowModule):
         state = DefaultToolState()
         state.inputs = dict()
         return state
+
+    def get_runtime_inputs(self, connections=None):
+        inputs = {}
+        for step in self.subworkflow.steps:
+            if step.type == "tool":
+                tool = step.module.tool
+                tool_inputs = step.module.state
+
+                def callback(input, prefixed_name, prefixed_label, value=None, **kwds):
+                    # All data parameters are represented as runtime values, skip them
+                    # here.
+                    if input.type in ['data', 'data_collection']:
+                        return
+
+                    if is_runtime_value(value):
+                        input_name = "%d|%s" % (step.order_index, prefixed_name)
+                        inputs[input_name] = InputProxy(input, input_name)
+
+                visit_input_values(tool.inputs, tool_inputs.inputs, callback)
+
+        return inputs
+
+
+class InputProxy(object):
+    """Provide InputParameter-interfaces over inputs but renamed for workflow context."""
+
+    def __init__(self, input, prefixed_name):
+        self.input = input
+        self.prefixed_name = prefixed_name
+
+    def to_dict(self, *args, **kwds):
+        as_dict = self.input.to_dict(*args, **kwds)
+        as_dict["name"] = self.prefixed_name
+        return as_dict
 
 
 class InputModule(WorkflowModule):
@@ -1175,16 +1210,23 @@ class WorkflowModuleInjector(object):
         # Any connected input needs to have value DummyDataset (these
         # are not persisted so we need to do it every time)
         module.add_dummy_datasets(connections=step.input_connections, steps=steps)
+
+        # Populate subworkflow components
+        if step.type == "subworkflow":
+            subworkflow_param_map = step_args or {}
+            unjsonified_subworkflow_param_map = {}
+            for key, value in subworkflow_param_map.items():
+                unjsonified_subworkflow_param_map[int(key)] = value
+
+            subworkflow = step.subworkflow
+            populate_module_and_state(self.trans, subworkflow, param_map=unjsonified_subworkflow_param_map)
+
         state, step_errors = module.compute_runtime_state(self.trans, step_args)
         step.state = state
 
         # Fix any missing parameters
         step.upgrade_messages = module.check_and_update_state()
 
-        # Populate subworkflow components
-        if step.type == "subworkflow":
-            subworkflow = step.subworkflow
-            populate_module_and_state(self.trans, subworkflow, param_map={})
         return step_errors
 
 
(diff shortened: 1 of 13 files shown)


lib/galaxy/workflow/modules.py
```diff
@@ -381,6 +382,39 @@ def get_runtime_state(self):
         state.inputs = dict()
         return state
 
+    def get_runtime_inputs(self, connections=None):
+        print(self.subworkflow)
```
mvdbeek: That's a debug remnant, right ?

lib/galaxy/workflow/modules.py
```diff
@@ -381,6 +382,39 @@ def get_runtime_state(self):
...
+        inputs = {}
+        for step in self.subworkflow.steps:
+            if step.type == "tool":
+                tool = step.module.tool
+                tool_inputs = step.module.state
+
+                def callback(input, prefixed_name, prefixed_label, value=None, **kwds):
+                    if input.type in ['data', 'data_collection']:
+                        return
```
mvdbeek: Don't we want to raise an exception here and prevent running these workflows ?

lib/galaxy/workflow/run_request.py
```diff
@@ -100,6 +101,16 @@ def _normalize_step_parameters(steps, param_map, legacy=False, already_normalize
             param_dict = param_map.get(str(step.order_index), {})
         else:
             param_dict = _step_parameters(step, param_map, legacy=legacy)
+        if step.type == "subworkflow" and param_dict:
+            if not already_normalized:
+                raise exceptions.RequestParameterInvalidException("subworkflow step parameters must normalized currently")
```
mvdbeek: Maybe `subworkflow step parameters must be normalized` ?

lib/galaxy/workflow/run_request.py
```diff
@@ -436,6 +453,9 @@ def workflow_request_to_run_config(work_request_context, workflow_invocation):
                 use_cached_job = (parameter.value == 'true')
         elif parameter_type == param_types.RESOURCE_PARAMETERS:
             resource_params[parameter.name] = parameter.value
+        elif parameter_type == param_types.STEP_PARAMETERS:
+            import json
```
mvdbeek: json is alredy imported at this point, right ?

lib/galaxy/workflow/modules.py
```diff
@@ -1196,7 +1237,7 @@ def populate_module_and_state(trans, workflow, param_map, allow_tool_state_corre
         module_injector = WorkflowModuleInjector(trans, allow_tool_state_corrections)
     for step in workflow.steps:
         step_args = param_map.get(step.id, {})
-        step_errors = module_injector.inject(step, step_args=step_args)
+        step_errors = module_injector.inject(step, step_args=step_args, param_map=step_args)
```
mvdbeek: Passing step_args twice looks a bit funny here, would it be problematic to always use `step_args` on line 1215 ?

mvdbeek (approved):
Incredible, that is so awesome! I played around with this a little, this is a huge improvement.

lib/galaxy/workflow/modules.py
jmchilton: I think data parameters with input connections are represented in the state as RuntimeValues still at this point - I guess because the data parameter is filled in later in the process. I had to put this check in for a perfectly valid workflow otherwise. This deserves a comment at least though for sure.

lib/galaxy/workflow/modules.py
jmchilton: Let me see how else it is used - but yeah that is definitely a good idea.

lib/galaxy/workflow/run_request.py
jmchilton: Yeah - made this change while "polishing" PR yesterday. I should have reviewed more closely - this is kind of a mess. Thanks - I'll fix this also.

lib/galaxy/workflow/modules.py
mvdbeek: Ah, that makes sense. We can definitely address this at a later point if you prefer. I do think it'd be nice to error out ASAP though for the user.

lib/galaxy/workflow/modules.py
jmchilton: So ``WorkflowProgress.remaining_steps()`` called inject without step_args previously - so making this change will change that - so this line will change:

```
        state, step_errors = module.compute_runtime_state(self.trans, step_args)
        step.state = state
```

I don't know if that will introduce bugs or fix bugs - but probably one of them. We will see what the tests say - but the UI exercises this API in ways that the tests don't cover. The new test ``test_run_runtime_parameters_after_pause`` still passes with this change - I guess that seems like the kind of thing most likely to break.

lib/galaxy/workflow/modules.py
jmchilton: I'm confused - this isn't an error - perfectly fine connected data inputs that are fine are represented as RuntimeValues. And the UI doesn't allow data parameters to be specified as RuntimeValues.

I guess you're suggesting we should produce proper errors on disconnected inputs - this I agree with but probably that should happen somewhere else? I suspect if we just stuck the disconnected data inputs in here they would work but I hate that a lot.

mvdbeek: I was more thinking that for runtime data inputs inside subworkflows we'll want to raise an error, but maybe I'm misunderstanding what's going on here.

mvdbeek: Nevermind, I think I misunderstood this.

mvdbeek: > I guess you're suggesting we should produce proper errors on disconnected inputs - this I agree with but probably that should happen somewhere else?

Right, I thought that would be a good place, but I see now that we don't know which inputs are disconnected at this point.

> I suspect if we just stuck the disconnected data inputs in here they would work but I hate that a lot.

I was curious and tried that out, yes it does, but I agree that's not a good idea.

mvdbeek:
Fantastic, this should be a highlight in the release notes.

PR #6668, opened 2018-09-04
Upgraded genomespaceclient to newer version

nuwang:
This fixes issue reported in: https://github.com/galaxyproject/usegalaxy-playbook/issues/100
Also related to PR: https://github.com/galaxyproject/galaxy/pull/6660

### DIFF
--- a/lib/galaxy/dependencies/pipfiles/default/Pipfile
+++ b/lib/galaxy/dependencies/pipfiles/default/Pipfile
@@ -69,7 +69,7 @@ svgwrite = "*"
 pyparsing = "*"
 "Fabric3" = "*"
 paramiko = "*"
-python-genomespaceclient = "<1.0"
+python-genomespaceclient = "<2.0"
 social_auth_core = {version = "==1.5.0", extras = ['openidconnect']}
 
 [requires]
--- a/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
+++ b/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
@@ -93,7 +93,7 @@ pyperclip==1.6.4
 pysam==0.14.1
 python-cinderclient==2.0.1
 python-dateutil==2.7.3
-python-genomespaceclient==0.2
+python-genomespaceclient==1.0.0
 python-glanceclient==2.6.0; python_version != '3.3.*'
 python-keystoneclient==3.10.0; python_version != '3.3.*'
 python-lzo==1.12
--- a/tools/genomespace/genomespace_importer.py
+++ b/tools/genomespace/genomespace_importer.py
@@ -148,7 +148,7 @@ def determine_file_type(input_url, output_filename, metadata, json_params, sniff
     Determine the Galaxy data format for this file.
     """
     # Use genomespace metadata to map type
-    file_format = metadata.dataFormat.name if metadata.dataFormat else None
+    file_format = metadata.data_format.name if metadata.data_format else None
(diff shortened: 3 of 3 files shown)


VJalili:
Also related to #5903

lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
```diff
@@ -93,7 +93,7 @@ pyperclip==1.6.4
 pysam==0.14.1
 python-cinderclient==2.0.1
 python-dateutil==2.7.3
-python-genomespaceclient==0.2
+python-genomespaceclient==1.0.0
```
nsoranzo, line 96: Not a blocker for this PR, but can you do a 1.0.1 release and push the universal wheel on PyPI? That would save us to build it on Python 2.7 (see https://travis-ci.org/galaxyproject/galaxy/jobs/424522196#L905 ).

nuwang: Sure, thanks for reminding me, I'd forgotten that the wheel hadn't been built already. We were planning to push out a new version with an updated cloudbridge dependency anyway, so the change you made to build universal wheels will be included with it.

nsoranzo: There's a wheel on PyPI, but only for Python 3 (because it predates https://github.com/gvlproject/python-genomespaceclient/pull/8 ).

nsoranzo (approved):

VJalili:
@nuwang does the new version of Genomespace depend on the cloudbridge v1? 

I see `install_requires=['cloudbridge>=0.3,<0.4', 'requests'],`, which I think will cause dependency conflict with Cloudbridge v1.

nuwang:
@VJalili Yes. All this does is to update the genomespace client. The cloudbridge version update will have to be made as a minor patch after it's released, hopefully tomorrow or Friday.

VJalili:
If the new Genomespace functionality depends on CloudBridge v1, then maybe it would be better if we update Genomespace and cloudbridge together with one PR #6660

nuwang:
@VJalili The gs-client update required some changes in the importer too, so I think that it's a separate issue conceptually and should remain a separate PR, although it's just one extra line of code. Once CB v1 is out, we can just include all version bumps separately with https://github.com/galaxyproject/galaxy/pull/6660?

VJalili:
As long as these changes are functional without requiring cloudbridge v1, I agree this should stay a separate PR.

nuwang:
@VJalili Yes, they are functional and do not require cloudbridge v1.

VJalili:
Sounds good. Thanks.

lib/galaxy/dependencies/pipfiles/default/Pipfile
```diff
@@ -69,7 +69,7 @@ svgwrite = "*"
 pyparsing = "*"
 "Fabric3" = "*"
 paramiko = "*"
-python-genomespaceclient = "<1.0"
+python-genomespaceclient = "<2.0"
```
VJalili, line 72: why this is not `1.0.0`?

nsoranzo: To allow minor release upgrades, e.g. the already released 1.1.0 . Is there a problem?

VJalili: I might be missing something here; is not this preventing upgrading cloudbridge version to v1 in #6660?

PR #6669, opened 2018-09-05
Increase robustness and accuracy of job search / cache use

mvdbeek:
The major point here is that we record every change to a
HistoryDatasetAssociation in a HistoryDatasetAssociationHistory, and we
update the actually used version once the job inputs are ready. This
prevents issues if a dataset is being updated while the job is new.

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -2570,7 +2570,8 @@ class HistoryDatasetAssociation(DatasetInstance, HasTags, Dictifiable, UsesAnnot
         state = inspect(self)
         changes = {}
 
-        for attr in state.attrs:
+        for attr in state.mapper.columns:
+            # We only create a new version if columns of the HDA table have changed, and ignore relationships.
             hist = state.get_history(attr.key, True)
 
             if not hist.has_changes():
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -34,7 +34,11 @@ from sqlalchemy import (
     type_coerce,
     types)
 from sqlalchemy.ext import hybrid
-from sqlalchemy.orm import aliased, joinedload, object_session
+from sqlalchemy.orm import (
+    aliased,
+    joinedload,
+    object_session,
+)
 from sqlalchemy.schema import UniqueConstraint
 
 import galaxy.model.metadata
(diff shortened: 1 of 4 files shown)


jmchilton:
I was already vaguely nervous about this performance of this but I've rerun a couple of these tests that failed, I'm concerned they are potentially timing out randomly because this slows things down. Probably just confirmation bias and everything is fine but I was wondering if you had an opinion.

In particular, I don't see a  ``@current_dataset_version.expression`` - do you need one of those - I thought you needed one with hybrid properties to cause that query evaluation to happen on the database server?

mvdbeek:
> I was already vaguely nervous about this performance of this but I've rerun a couple of these tests that failed, I'm concerned they are potentially timing out randomly because this slows things down.

Yeah, I think that is probably what is happening, although for instance the integration tests rarely do any tool tests and yet they slow down a lot.

Ahh, but the API test are also slow, so that's probably it.

mvdbeek:
I think I was overcomplicating it here, the current_dataset_version property shouldn't be necessary, so I've dropped that in https://github.com/galaxyproject/galaxy/pull/6669/commits/d063933799266ee2415d62376f057f6a56df231e

lib/galaxy/model/__init__.py
```diff
@@ -1110,7 +1115,11 @@ class JobToInputDatasetAssociation(object):
     def __init__(self, name, dataset):
         self.name = name
         self.dataset = dataset
-        self.dataset_version = dataset.version if dataset else None
+        self.dataset_version = 0
+
+    @hybrid.hybrid_property
+    def current_dataset_version(self):
+        return self.dataset_version or self.dataset.version
```
nsoranzo: Should this be removed too?

mvdbeek: Absolutely, forgot to git add that

lib/galaxy/model/__init__.py
```diff
@@ -2572,7 +2572,7 @@ def __create_version__(self, session):
 
         changes = {}
 
-        for attr in class_mapper(HistoryDatasetAssociation).column_attrs:
+        for attr in state.attrs:
```
mvdbeek: Looks like that was causing the issues, at least the integration tests passed in time.

PR #6676, opened 2018-09-07
Tell pytest not to execute Python code in node_modules for web proxy.

jmchilton:
Causes test execution to fail for me.


mvdbeek (approved):

mvdbeek:
Thanks, I think this was already on the ignore list when I made the switch, but I dropped all the suff that either worked or could be made to work.

PR #6677, opened 2018-09-07
Serialize select field options, fixes beta install

mvdbeek:
This fixes the empty select list in the beta install interface.


mvdbeek:
For reference this is what I see without this fix.
screen shot 2018-09-07 at 14 29 51
And that then fails during install.

Issue #6584, opened 2018-08-09
Should keep old tools if reloading new tool failed


mvdbeek:
That's a bit annoying with planemo serve and development. If you break the XML, the tool disappears, and hence isn't being watched anymore. Either we only load valid new tools, or we somehow keep watching.

```
galaxy.tools.toolbox.base INFO 2018-08-09 10:57:04,513 Parsing the tool configuration /private/var/folders/df/6xqpqpcd7h73b6jpx9t6cwhw0000gn/T/tmpicezz6xb/tool_conf.xml
galaxy.tools.toolbox.base ERROR 2018-08-09 10:57:04,519 Error reading tool from path: /Users/mvandenb/src/tools-iuc/tools/fastp/fastp.xml
Traceback (most recent call last):
  File "/private/var/folders/df/6xqpqpcd7h73b6jpx9t6cwhw0000gn/T/tmpicezz6xb/galaxy-dev/lib/galaxy/tools/toolbox/base.py", line 567, in _load_tool_tag_set
    tool = self.load_tool(concrete_path, use_cached=False)
  File "/private/var/folders/df/6xqpqpcd7h73b6jpx9t6cwhw0000gn/T/tmpicezz6xb/galaxy-dev/lib/galaxy/tools/toolbox/base.py", line 751, in load_tool
    tool = self.create_tool(config_file=config_file, repository_id=repository_id, guid=guid, **kwds)
  File "/private/var/folders/df/6xqpqpcd7h73b6jpx9t6cwhw0000gn/T/tmpicezz6xb/galaxy-dev/lib/galaxy/tools/__init__.py", line 270, in create_tool
    raise e
KeyError: 'adapter_trimming_options'
```

mblue9:
Oh if this can be changed that would be great! This has bitten me many times! Would be great not to have to restart planemo serve and reload data in every time this happens (or is there a workaround?), that would save a chunk of time.

PR #6681, opened 2018-09-08
Keep newly broken tool in toolbox

mvdbeek:
That's handy when developing tools and you messed up something so bad
the tool isn't valid anymore. We retrieve the non-broken previous
version from the tool cache.

Still needs some visual feedback in the tool form, so that you're not left wondering why nothing happens when you change the tool ...

Closes https://github.com/galaxyproject/galaxy/issues/6584

### DIFF
--- a/client/galaxy/scripts/mvc/tool/tool-form-base.js
+++ b/client/galaxy/scripts/mvc/tool/tool-form-base.js
@@ -86,6 +86,12 @@ export default FormBase.extend({
                     .append(this._footer())
             );
         }
+        options.tool_errors &&
+            this.message.update({
+                status: 'danger',
+                message: options.tool_errors,
+                persistent: true
+            });
         this.show_message &&
             this.message.update({
                 status: "success",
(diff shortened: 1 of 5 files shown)


mvdbeek:
That's how it looks like if you just broke a tool:
screen shot 2018-09-08 at 15 33 25

client/galaxy/scripts/mvc/tool/tool-form-base.js
```diff
@@ -86,6 +86,12 @@ export default FormBase.extend({
                     .append(this._footer())
             );
         }
+        options.tool_errors &&
+            this.message.update({
+                status: 'warning',
```
martenson: maybe we could use `error` to be more prominent (or is it `alert`?)

mvdbeek: Might be alert, error didn't do anything

mvdbeek: Both error and alert don't do anything:
screen shot 2018-09-08 at 16 20 04

martenson: it is probably based on bootstrap, so try`danger`

mvdbeek: That was it, thanks @martenson !
screen shot 2018-09-08 at 17 58 48

mvdbeek: I've rebased the last commit with this change.

PR #6682, opened 2018-09-08
Trim whitespace from data manager conf paths

pvanheus:
The templates used for `<target>` and `<value_translation>` tags should be `.strip()` to make it easier to write sophisticated templates that still yield a single line.


martenson:
Thanks @pvanheus !

PR #6684, opened 2018-09-10
 Deprecate `size` attribute of `<param/>` and remove it from tools 

nsoranzo:
Also:
- Remove an unused and broken tool `tools/data_source/bed_convert.xml`
- `dos2unix test/functional/tools/for_workflows/head.xml`
- Single-quote text and data params in `<command/>`
- Remove deprecated `interpreter` attribute of `<command />`


jmchilton:
I'm -0 on this. Clearly the docs should be updated to reflect this is unused. But in the abstract it seems like valid display metadata the tool author might want to capture - especially in the case of text area inputs. Galaxy not actually using it currently is pretty irrelevant in my mind - Galaxy tools shouldn't be artifacts designed to only be used by one specific runtime and shouldn't only capture metadata that runtime can currently consume intelligently. I understand other people feel differently about these points and I'm fine with this being merged.

nsoranzo:
> Galaxy tools shouldn't be artifacts designed to only be used by one specific runtime and shouldn't only capture metadata that runtime can currently consume intelligently.

@jmchilton I think a reason to drop `size` is that it is actually a presentational detail which doesn't fit in the "abstract" tool description you support. In fact, there's no equivalent of `size` in the CWL command line tool description standard, AFAIK.

jmchilton:
Galaxy tools have many presentational aspects, this isn't a problem in my opinion and I think it is likely a strength of them versus other approaches. The counter argument against CWL made to me frequently is that it doesn't capture any presentation details - I am told it is a deficit of that specification and I find the argument not unpersuasive. Admittedly you've never made that argument to me, but others certainly have. Regardless, I'm not trying to make Galaxy tools into CWL tools - I'm trying to make them suitably abstract and detached from the specific implementation that is Galaxy - form input type, labels, and help seem appropriate and at roughly the same level at preferred text box size and shape.

nsoranzo:
> form input type, labels, and help seem appropriate and at roughly the same level at preferred text box size and shape.

I generally support making Galaxy tools as abstract descriptions as possible, but I don't agree on the above sentence: text box size and shape are at the same level of text colour and font, IMO. In fact, input type and label are in the CWL tool description.

guerler:
In my view the size attribute is/was basically the size property of the HTML input dom field. It is a somewhat arbitrary UI suggestion and I'd be fine with removing it. It is true that tool configs contain representational properties for inputs like e.g. the display property of select fields or the hidden option. But these properties are more abstract and do not go into the detail of specifying a particular numeric value like a certain length, width or color.

dannon:
Since there seems to be discussion to be had here, and it's not an urgent change, I'm going to re-milestone this for 19.01 so we can freeze.

jmchilton:
Merged with serious reservations. Thanks though for the contribution, it is appreciated even if we don't agree.

PR #6338, opened 2018-06-14
Allow for fatal_oom detection by regex also when exit_code tests are present

bernt-matthias:
The current behavior of the stdio error handling is to apply regexes only if no (fatal) exit code was found: 

https://github.com/galaxyproject/galaxy/blob/c78a23cb873cf5cbcf177cfa7052ae6dad506eea/lib/galaxy/jobs/output_checker.py#L68

Thus, because fatal_oom > fatal currently fatal_oom can be detected if and only if
- they are indicated by an exit_code
- indicated by output to stderr/stdio but no other (fatal) exit_code rule is found.

This assumes that programmers use exit_codes to indicate different error conditions -- which is in my opinion a very optimistic assumption that virtually never applies (I can not remember a single program). C programmers, for instance, usually just use libc's EXIT_SUCCESS and EXIT_FAILURE macros and output a more specific message to stderr. Hence in my opinion exit codes are in practice (unfortunately) unspecific. 

Furthermore, the current implementations of galaxy tools test for non zero exit code only anyway (I grepped in all iuc,  devteam, and bgruening tools and only found a single one). 

The PR is changes this behavior by checking for regexes also if a FATAL exit_code was found (and also all specified regexes are tested). 

Furthermore: 
- As a suggestion I implemented the test line wise in reverse order. Intuitively I would expect this to be faster. The original code is still available as additional function - so this is easy to revert. 
- I introduced a function for checking a regex against a stream (which was a TODO in the code).

PR #6685, opened 2018-09-10
[WIP] Track job failures in a structured way.

jmchilton:
Still working on this - built from @bernt-matthias's branch https://github.com/galaxyproject/galaxy/pull/6338.


jmchilton:
@bernt-matthias So this breaks all the new tests in error_detect.xml. Looking over those tests, ``failure_reasons`` might not be a great name for this property I added either because there are info level messages.

Should I should maybe this redo this and call the property ``runtime_messages`` or ``job_messages`` and those tests need to be modified since we are no longer modifying the programs stdout or stderr - which I still think is a good thing. I'll think about this some more, sorry for mucking things up.

bernt-matthias:
Fine with any of the suggested names. 

Regarding the test: currently the fatal_oom regex is tested only in conjunction with fatal(_oom) exit codes. But we should also test it alone, i.e. without an exit code. 

I'm a bit lost in the tests, can I see somewhere the the reported levels for each of the tests.

lib/galaxy/jobs/output_checker.py
```diff
@@ -88,32 +114,16 @@ def check_output(tool, stdout, stderr, tool_exit_code, job):
                     #   - If it matched, then determine the error level.
                     #       o If it was fatal, then we're done - break.
                     # Repeat the stdout stuff for stderr.
-                    # TODO: Collapse this into a single function.
+                    # TODO could test for stderr first? Reason: I would expect it to be smaller and contain the errors
```
bernt-matthias: I agree. lets just reverse order

bernt-matthias:
Hi @jmchilton any progress on this. Just let me know if there is anything I can help.

lib/galaxy/jobs/output_checker.py
jmchilton: will do

jmchilton:
Rebased and took a crack at polishing this.

Barring test failures I think what is left is to restore/improve the UX around reporting why jobs failed.

jmchilton:
Visual of job messages no longer being pre-pending to job stdout/stderr - now appear as structured data in the show parameters UI:

screen shot 2018-12-11 at 9 14 00 am

jmchilton:
Closing in lieu of #7095

PR #6693, opened 2018-09-11
Add utility function to create history entries

guerler:
This PR provides a utility function enabling UI plugins to create history datasets.
```
Galaxy.data.create({
     url_paste: "Example Content",
     success: (response) => {window.console.log(response)}
});
```

### DIFF
--- a/client/galaxy/scripts/layout/data.js
+++ b/client/galaxy/scripts/layout/data.js
@@ -47,7 +47,7 @@ export default class Data {
                         "files_0|space_to_tab": options.space_to_tab ? "Yes" : null,
                         "files_0|to_posix_lines": options.to_posix_lines ? "Yes" : null,
                         "files_0|dbkey": options.genome || null,
-                        "files_0|file_type": options.extension || null,
+                        "files_0|file_type": options.extension || "auto",
                         "files_0|url_paste": options.url_paste
                     })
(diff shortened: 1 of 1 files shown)


client/galaxy/scripts/layout/data.js
```diff
@@ -16,4 +16,42 @@ export default class Data {
...
+                    history_panel.refreshContents();
+                }
+                if (options.success) {
+                    options.success(response);
+                }
+            },
+            error: options.error,
+            data: {
+                payload: {
+                    tool_id: "upload1",
```
mvdbeek, line 42: could this be done with the fetch tool instead ? Isn't that intended to replace upload1 in the future ?

guerler: Interesting, I have not used that tool yet in that fashion. Do we have an example? It should be straightforward to redirect the post if needed, either now or in the future.

guerler: @mvdbeek I would prefer to keep using the current upload tool until the new tool is fully functional and ready to replace it.

martenson:
@mvdbeek I think this is what we currently do in many places, the data_fetch tool could be used I think, but that does not make this wrong or obsolete imho.

PR #6695, opened 2018-09-12
[18.09] Make gro and top datatypes tabular subclasses

mvdbeek:
I think that was the intention in https://github.com/galaxyproject/galaxy/commit/56d55399ebe2ecd614a7ba5a146739f58e917541

@simonbray can you confirm this is correct ?


simonbray:
Yes - I removed the sniffers but didn't update the datatypes_conf file. Thanks!

nsoranzo (approved):

PR #5925, opened 2018-04-19
[18.01] jobscript - check memory later

hexylena:
This tried to write to cwd before cd'ing and our upload jobs thus failed due to RO filesystem

PR #5880, opened 2018-04-12
Move the memory statement below _galaxy_setup_environment

moskalenko:
This is done  to ensure that a non-standard PATH to, for example, a SLURM install will be applied if set in environment_setup_file for 'scontrol' call in the job shell script. This was discovered by me and a fix was proposed by mvdbeek.

PR #6697, opened 2018-09-12
jobscript order (memory statements before environment setup)

bernt-matthias:
This moves the memory statement before the environment setup. As a minor change I moved the export of the slots variable into the slots statement. 

Rationale: If the memory variable is set before the environment then the environment can be configured 
dynamically based on memory and slots. I'm thinking here for instance of setting java's -Xmx parameter dynamically based on the available memory. In particular when submitting with dynamic slot ranges (e.g. grid engine allows to submit jobs with 8-32 processors) the memory depends on the actual number of processors and can not be set statically.


bernt-matthias:
Damn 6258cc6  and 53f066f need to be removed .. somehow.

bernt-matthias:
Understood why the test is not running. Could either relocate the export (but then I would do the analogous for the memory statement) or if possible just insert the actual template. Any ideas?

nsoranzo:
Unfortunately this is in direct conflict with what was done in https://github.com/galaxyproject/galaxy/pull/5880 , ping @moskalenko.

bernt-matthias:
Hmm, I see. Is there maybe an environment variable set by slurm that gives the available memory, as in SGE?

moskalenko:
This would break the currently working SLURM environments. The memory_statement could set both java options and the tightly-coupled grid environment options. Why is there a need to move the memory_statement up in the order?

bernt-matthias:
I was thinking to call something similar to:

`_JAVA_OPTIONS=-Xmx ((GALAXY_MEMORY_MB * 2/3))`

in my environment setup. I guess that there will not be a consensus on the actual computation since this depends on the configuration of the cluster.

bernt-matthias:
I'm just to lazy to maintain this value in each of the destinations that I'm using.

jmchilton:
I think given the rationale on #5880 this shouldn't be merged. If they are your destinations, can you use the underlying DRM set variables instead of the standardized Galaxy ones in your common setup for things like ``_JAVA_OPTIONS``? Presumably you know your cluster better than Galaxy.

For this reason I'm going to mark this as WIP and kick the milestone to the next release. Let me know if I'm crazy though and we can discuss things more.

hexylena:
Yeah, agreed with @jmchilton, I hit it in https://github.com/galaxyproject/galaxy/pull/5925 and made a similar change as in #5880, for us 

> This tried to write to cwd before cd'ing and our upload jobs thus failed due to RO filesystem

So it is simply impossible that either of those come before changing directory unfortunately :(

bernt-matthias:
Agreed.

PR #6702, opened 2018-09-12
Fix wrong version being returned if requested version doesn't exist

mvdbeek:
Includes a unit test that fails without the fix and an API test that tests showing a tool by version (but that wouldn't actually fail without the fix for a local tool).

### DIFF
--- a/lib/galaxy/tools/toolbox/base.py
+++ b/lib/galaxy/tools/toolbox/base.py
@@ -446,6 +446,9 @@ class AbstractToolBox(Dictifiable, ManagesIntegratedToolPanelMixin):
                 for tool in self._tools_by_id.values():
                     if tool.old_id == tool_id:
                         rval.append(tool)
+                # if we don't have a lineage_map for this tool we need to sort by version,
+                # so that the last tool in rval is the newest tool.
+                rval.sort(key=lambda t: t.version)
             if rval:
                 if get_all_versions:
                     return rval
--- a/lib/galaxy/tools/toolbox/base.py
+++ b/lib/galaxy/tools/toolbox/base.py
@@ -455,8 +455,8 @@ class AbstractToolBox(Dictifiable, ManagesIntegratedToolPanelMixin):
                         for tool in rval:
                             if tool.version == tool_version:
                                 return tool
-                    # No tool matches by version, simply return the first available tool found
-                    return rval[0]
+                    # No tool matches by version, simply return the newest matching tool
+                    return rval[-1]
             # We now likely have a Toolshed guid passed in, but no supporting database entries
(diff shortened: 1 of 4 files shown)


lib/galaxy/tools/toolbox/base.py
```diff
@@ -455,8 +455,8 @@ def get_tool(self, tool_id, tool_version=None, get_all_versions=False, exact=Fal
                         for tool in rval:
                             if tool.version == tool_version:
                                 return tool
-                    # No tool matches by version, simply return the first available tool found
-                    return rval[0]
+                    # No tool matches by version, simply return the newest matching tool
+                    return sorted(rval, key=lambda x: x.version, reverse=True)[0]
```
nsoranzo: I think `rval[-1]` should be a faster equivalent, since `tool_lineage.get_versions()` returns ToolLineageVersion objects from oldest to newest.

mvdbeek: I had forgotten that!

martenson:
this could aim at 18.05 I think

mvdbeek:
probably earlier than that, need to figure out when we moved away from the toolshed versioning scheme

lib/galaxy/webapps/galaxy/api/tools.py
```diff
@@ -97,7 +97,8 @@ def show(self, trans, id, **kwd):
         """
         io_details = util.string_as_bool(kwd.get('io_details', False))
         link_details = util.string_as_bool(kwd.get('link_details', False))
-        tool = self._get_tool(id, user=trans.user)
+        version = kwd.get('version')
```
nsoranzo: Can you add `version` to the API documentation above?

test/api/test_tools.py
```diff
@@ -142,8 +142,11 @@ def test_data_sources_block_file_parameters(self):
             assert output_details["state"] == "error", output_details
             assert "has not sent back a URL parameter" in output_details["misc_info"], output_details
 
-    def _show_valid_tool(self, tool_id):
-        tool_show_response = self._get("tools/%s" % tool_id, data=dict(io_details=True))
+    def _show_valid_tool(self, tool_id, version=None):
+        url = "tools/%s" % tool_id
+        if version:
+            url += "?version=%s" % version
+        tool_show_response = self._get(url, data=dict(io_details=True))
```
nsoranzo: I think this should be:
```python
        data = dict(io_details=True)
        if version:
            data['version'] = version
        tool_show_response = self._get("tools/%s" % tool_id, data=data)
```

mvdbeek:
I've rebased this with the changes, should be ready to go!

nsoranzo (approved):

nsoranzo (approved):

PR #6293, opened 2018-06-06
[18.05] Strip a spurious Slurm warning from job stderr

nsoranzo:
Reason for the Slurm warning
----------------------------

The Linux kernel memory controller (responsible for the memory cgroup) may run out of cgroup subsystem state (CSS) IDs. When a cgroup is created it is assigned a CSS ID to manage its state. Upon removal of the cgroup the corresponding state information (e.g. cache entries) may still exist and therefore the CSS ID is still held. When multiple frequent short-lived jobs are run on a cluster node, the number of available CSS IDs becomes exhausted and creating a new memory cgroup results in a ENOSPC ("No space left on device"), which is reported by SLURM with the message `unable to add task[pid=<PID>] to memory cg '(null)'` (even if jobs run fine).

There is a bugfix for the kernel that releases the CSS ID upon cgroup destruction, but it appears to be only available after Linux 4.4. We use CentOS 7, which is based on Linux 3.10. The only temporary solutions is to reboot the affected cluster node.

Thanks to @tuxtobin for the detailed analysis above.

Reason for this patch
---------------------

Many tools rely on an empty stderr to determine if the job was successful, either because they were never updated to use `<stdio>`/`detect_errors`, or because the underlying tool returns a non-zero exit code when successful, e.g. `tranalign` from https://toolshed.g2.bx.psu.edu/view/devteam/emboss_5/832c20329690 . Even using a `<regex>` inside `<stdio>` would not work because the Slurm warning contains the word `error`.

PR #6705, opened 2018-09-12
[18.05] Catch additional permutations of Slurm cgroup creation warnings

natefoo:
Newer versions of Slurm produce these messages in different ways, see #6293 for details.

A handy tool to test:

```xml
<tool id="slurm_warn" name="Generate Slurm warnings">
    <command><![CDATA[
        {
        true ;
        #for $warning in str($warnings).split(','):
            #if $warning == "add":
                echo "slurmstepd: error: task/cgroup: unable to add task[pid=123] to memory cg '(null)'" ;
            #elif $warning == "create":
                echo "slurmstepd: error: xcgroup_instantiate: unable to create cgroup '/sys/fs/cgroup/foo/bar' : No space left on device" ;
            #elif $warning == 'instanciate':
                echo 'slurmstepd: error: jobacct_gather/cgroup: unable to instanciate job 345 memory cgroup' ;
            #end if
        #end for
        #for $i in $range($stderr_lines):
            echo "additional stderr line $i" ;
        #end for
        } >&2
    ]]></command>
    <inputs>
        <param name="stderr_lines" type="integer" label="stderr lines to generate" value="0" default_value="0" min="0" max="10"/>
        <param name="warnings" type="select" label="Warnings to include" multiple="true" display="checkboxes">
            <option value="add">task/cgroup: unable to add task</option>
            <option value="create">xcgroup_instantiate: unable to create cgroup</option>
            <option value="instanciate">jobacct_gather/cgroup: unable to instanciate job</option>
        </param>
    </inputs>
    <outputs>
        <data name="output1" format="txt" />
    </outputs>
</tool>
```

### DIFF
--- a/lib/galaxy/jobs/runners/slurm.py
+++ b/lib/galaxy/jobs/runners/slurm.py
@@ -29,6 +29,13 @@ SLURM_MEMORY_LIMIT_EXCEEDED_PARTIAL_WARNINGS = [': Exceeded job memory limit at 
                                                 ': Exceeded step memory limit at some point.']
 SLURM_MEMORY_LIMIT_SCAN_SIZE = 16 * 1024 * 1024  # 16MB
 SLURM_UNABLE_TO_ADD_TASK_TO_MEMORY_CG_MSG_RE = re.compile(r"""slurmstepd: error: task/cgroup: unable to add task\[pid=\d+\] to memory cg '\(null\)'$""")
+SLURM_UNABLE_TO_CREATE_CGROUP_MSG_RE = re.compile(r"""slurmstepd: error: xcgroup_instantiate: unable to create cgroup '[^']+' : No space left on device$""")
+SLURM_UNABLE_TO_INSTANCIATE_JOB_MSG_RE = re.compile(r"""slurmstepd: error: jobacct_gather/cgroup: unable to instanciate job \d+ memory cgroup$""")
+SLURM_TOP_WARNING_RES = (
+    SLURM_UNABLE_TO_ADD_TASK_TO_MEMORY_CG_MSG_RE,
+    SLURM_UNABLE_TO_CREATE_CGROUP_MSG_RE,
+    SLURM_UNABLE_TO_INSTANCIATE_JOB_MSG_RE
+)
 
 # These messages are returned to the user
 OUT_OF_MEMORY_MSG = 'This job was terminated because it used more memory than it was allocated.'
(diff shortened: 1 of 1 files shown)


jmchilton (approved):

dannon:
@nsoranzo Did you want to look at this one?  It looks good to me, but I don't want to merge with the outstanding review request.

lib/galaxy/jobs/runners/slurm.py
```diff
@@ -201,3 +202,34 @@ def __check_memory_limit(self, efile_path):
...
         return False
+
+
+def _remove_spurious_top_lines(rfh, ajs, maxlines=2):
+    bad = []
+    putback = None
+    i = 0
+    while i < maxlines:
+        line = rfh.readline()
+        log.debug('#### checking line: %s', line)
```
nsoranzo: Maybe `log.trace()` here?

lib/galaxy/jobs/runners/slurm.py
```diff
@@ -29,6 +29,13 @@
                                                 ': Exceeded step memory limit at some point.']
 SLURM_MEMORY_LIMIT_SCAN_SIZE = 16 * 1024 * 1024  # 16MB
 SLURM_UNABLE_TO_ADD_TASK_TO_MEMORY_CG_MSG_RE = re.compile(r"""slurmstepd: error: task/cgroup: unable to add task\[pid=\d+\] to memory cg '\(null\)'$""")
+SLURM_UNABLE_TO_CREATE_CGROUP_MSG_RE = re.compile(r"""slurmstepd: error: xcgroup_instantiate: unable to create cgroup '[^']+' : No space left on device$""")
+SLURM_UNABLE_TO_INSTANCIATE_JOB_MSG_RE = re.compile(r"""slurmstepd: error: jobacct_gather/cgroup: unable to instanciate job \d+ memory cgroup$""")
```
nsoranzo, line 33: @natefoo By looking at the code below, it seems you see both the new lines for each job, correct?

lib/galaxy/jobs/runners/slurm.py
```diff
@@ -201,3 +202,34 @@ def __check_memory_limit(self, efile_path):
             log.exception('Error reading end of %s:', efile_path)
 
         return False
+
+
+def _remove_spurious_top_lines(rfh, ajs, maxlines=2):
+    bad = []
+    putback = None
+    i = 0
+    while i < maxlines:
```
nsoranzo: `for i in range(maxlines):`

lib/galaxy/jobs/runners/slurm.py
```diff
@@ -201,3 +202,34 @@ def __check_memory_limit(self, efile_path):
...
+            if bad:
+                # no match found on this line so line is now a good line, but previous bad lines are found, so it needs to be put back
+                putback = line
+            # no match on this line, stop looking
+            break
+        # check next line
+        i += 1
+    if bad:
+        with tempfile.NamedTemporaryFile('w', delete=False) as wfh:
+            if putback:
```
nsoranzo: I'd use `if putback is not None:` since line could have been `''` (which should be copied back nonetheless).

lib/galaxy/jobs/runners/slurm.py
natefoo: Whoops, that wasn't even supposed to be in there, but I like the idea.

lib/galaxy/jobs/runners/slurm.py
natefoo: Yes, although I wasn't sure if it was a good idea to assume that would always be true.

natefoo:
Thanks as always for the suggestions @nsoranzo!

nsoranzo (approved):

natefoo:
Apparently it's possible to have all 3 messages, plus more. One job that failed due to some network maintenance that caused nodes to go offline had the following stderr:

```
slurmstepd: error: task/cgroup: unable to add task[pid=1852] to memory cg '(null)'
slurmstepd: error: xcgroup_instantiate: unable to create cgroup '/sys/fs/cgroup/memory/slurm/uid_819800/job_14794258' : No space left on device
slurmstepd: error: jobacct_gather/cgroup: unable to instanciate job 14794258 memory cgroup
slurmstepd: error: get_exit_code task 0 died by signal
```

I think the last error may be legitimate, but it looks like we should at least default `maxlines` to `3`.

PR #6707, opened 2018-09-12
[18.05] unquote the tool_version in the tools API

### LINKED ISSUE #6701: incorrect version of the tool loaded
https://usegalaxy.org/root?tool_id=toolshed.g2.bx.psu.edu%2Frepos%2Fiuc%2Fsnpeff%2FsnpEff_build_gb%2F4.3%2BT.galaxy3 displays `SnpEff build: database from Genbank record (Galaxy Version 4.3+T.galaxy1)` even though all 3 versions are present in the toolbox

reproduced on .eu, see https://usegalaxy.eu/?tool_id=toolshed.g2.bx.psu.edu%2Frepos%2Fiuc%2Fsnpeff%2FsnpEff_build_gb%2F4.3%2BT.galaxy3&version=4.3%2BT.galaxy3&__identifer=db21p7a1en

ping @mvdbeek who seemed interested a bit

martenson:
if the tool version includes '+', which is a recommended way
of versioning per https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#tool-versions
this would fail the comparison down the line in the toolbox.get_tool() method

this commit also downgrades the unquote_plus to unquote for tool_id
for similar reasons since it also contains the version

ping @mvdbeek 

fixes #6701

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/tools.py
+++ b/lib/galaxy/webapps/galaxy/api/tools.py
@@ -1,7 +1,7 @@
 import logging
 from json import dumps
 
-from six.moves.urllib.parse import unquote_plus
+from six.moves.urllib.parse import unquote
 
 import galaxy.queue_worker
 from galaxy import exceptions, managers, util, web
@@ -512,7 +512,9 @@ class ToolsController(BaseAPIController, UsesVisualizationMixin):
     # -- Helper methods --
     #
     def _get_tool(self, id, tool_version=None, user=None):
-        id = unquote_plus(id)
+        id = unquote(id)
+        if tool_version:
+            tool_version = unquote(tool_version)
         tool = self.app.toolbox.get_tool(id, tool_version)
         if not tool:
             raise exceptions.ObjectNotFound("Could not find tool with id '%s'." % id)
--- a/lib/galaxy/webapps/galaxy/api/workflows.py
+++ b/lib/galaxy/webapps/galaxy/api/workflows.py
@@ -9,7 +9,7 @@ import os
 
 import requests
 from markupsafe import escape
-from six.moves.urllib.parse import unquote_plus
+from six.moves.urllib.parse import unquote
 from sqlalchemy import desc, false, or_, true
 from sqlalchemy.orm import joinedload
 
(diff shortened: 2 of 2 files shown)


mvdbeek:
oh, yeah, that makes sense!

mvdbeek (approved):

martenson:
@mvdbeek any idea why we would `unquote_plus` tool_id before?

mvdbeek:
nope, that just seems wrong to me. We're doing that for > 4 years, didn't blame until the beginning of time. but we also do this in https://github.com/galaxyproject/galaxy/blob/3e4bbfbbc29f84ae30cd5905e7a43376915cbf47/lib/galaxy/webapps/galaxy/api/workflows.py#L561, so we should change that there too or unify the helper

blankenberg:
Are we sure we shouldn't be doing unquote_plus? Any plus should be encoded into `%2B` before being used.

mvdbeek:
These are requests coming in from the API, so we want to work with `+`, whether it's encoded or not.
```
In [1]: from six.moves.urllib.parse import unquote_plus

In [2]: from six.moves.urllib.parse import unquote

In [3]: test = 'snpEff_build_gb%2F4.3%2BT.galaxy3'

In [4]: unquote_plus(test)
Out[4]: 'snpEff_build_gb/4.3+T.galaxy3'

In [5]: unquote(test)
Out[5]: 'snpEff_build_gb/4.3+T.galaxy3'

In [6]: unquote_plus(unquote(test))
Out[6]: 'snpEff_build_gb/4.3 T.galaxy3'

In [7]: unquote(unquote(test))
Out[7]: 'snpEff_build_gb/4.3+T.galaxy3'
```
So unquote_plus replaces the unencoded plus with a whitespace (6), which is problematic.
Now I wonder if we even need to do any unquoting at all ...

mvdbeek:
OK, so I traced this back up until webob.compat's parse_qsl_text, which does the unquoting for us
https://github.com/Pylons/webob/blob/18a9b5298038e7bf1aa047d4b539cf6a00f323b7/src/webob/compat.py#L105

So this is really high up in the stack, I don't think we need to do any unquoting at all.

mvdbeek:
@martenson I opened a PR against your branch that drops unquoting of the tool id altogether https://github.com/martenson/galaxy/pull/30

martenson:
@mvdbeek That PR does not work, I have tried that before. Does that load the proper tool for you?

martenson:
Pushed a commit that does the same for get_tool helper method in workflows API

mvdbeek (approved):

PR #6711, opened 2018-09-13
[18.09] Add self help to error reporting interface

dannon:
As requested by @nekrut and @jennaj.  This could be configurable in the future, but right now gives us a good way to link out to a standard location with more help that we can update separately from the codebase.

### DIFF
--- a/client/galaxy/scripts/mvc/dataset/dataset-error.js
+++ b/client/galaxy/scripts/mvc/dataset/dataset-error.js
@@ -2,6 +2,12 @@ import _l from "utils/localization";
 import Utils from "utils/utils";
 import Ui from "mvc/ui/ui-misc";
 import Form from "mvc/form/form-view";
+import Backbone from "backbone";
+import * as _ from "underscore";
+
+/* global Galaxy */
+/* global $ */
+
 /** Dataset edit attributes view */
 var View = Backbone.View.extend({
     initialize: function() {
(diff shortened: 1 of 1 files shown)


martenson:
Since this was identified as high priority should it go to 18.09?

dannon:
If we want to make an exception for it, sure, I'm fine with that.  Anyone not want to do that?

(I went ahead and did it, but if anyone doesn't want the change made, I'll retarget again)

jmchilton:
Exception? No exception needed - just retitle - "Bugfix - users not presented with essential debug information".

bugs

dannon:
@jmchilton Works for me!

client/galaxy/scripts/mvc/dataset/dataset-error.js
```diff
@@ -57,12 +63,23 @@ var View = Backbone.View.extend({
...
+            <h3>Troubleshoot This Error</h3>
+            <p>
+                There are a number of help resources to self diagnose and
+                correct problems.
+                Start here: <a
+                href="https://galaxyproject.org/support/tool-error/"
+                target="_blank"> My job ended with an error. What can I do?</a>
+            </p>
+
+            <h3>Report This Error</h2>
```
martenson: the `<h3>` here is not closed

dannon: D'oh, good catch.  Copied it from before and only updated one end!

dannon: (fixed)

PR #6723, opened 2018-09-17
[18.09] Add Loom datatype and remove rdata.seurat

blankenberg:


blankenberg:
@galaxybot test this

blankenberg:
@galaxybot test this

jxtx:
image

bgruening (approved):
Thanks @blankenberg! Looks good to me.
@mblue9 any objections to remove rdata.seurat in favor of loom for this release?

Not sure why tests are failing :(

mblue9:
Fine by me 👍

dannon:
Rerunning the integration tests to see what's up.

mvdbeek:
The previous integration passed but didn't shut down correctly. The error now is unrelated.

dannon:
@mvdbeek Yeah, I dug through, too, and it looked unrelated to me.  Queued it for another run just to see what happens with that test, though

lib/galaxy/datatypes/binary.py
```diff
@@ -736,6 +736,97 @@ def display_peek(self, dataset):
...
+                dataset.metadata.description = loom_file.attrs.get('description', None)
+                dataset.metadata.url = loom_file.attrs.get('url', None)
+                dataset.metadata.doi = loom_file.attrs.get('doi', None)
+                dataset.metadata.loom_spec_version = loom_file.attrs.get('LOOM_SPEC_VERSION', None)
+                dataset.creation_date = loom_file.attrs.get('creation_date', None)
+                dataset.metadata.shape = tuple(loom_file['matrix'].shape)
+
+                tmp = list(loom_file['layers'].keys())
+                dataset.metadata.layers_count = len(tmp)
+                dataset.metadata.layers_names = tmp
```
mvdbeek, line 809: If loom_file['layers'] and friends aren't some kind of OrderedDict we should probably sort here

blankenberg: Its not technically a dictionary at all. Its an object that implements several bits of the dictionary style interface. My understanding is that h5py returns things in the same order as the standard tools (alphanumerically sorted), which is probably the desired behavior here, but I haven't done any extensive testing here. Presumably, if a specific tool needs a different ordering, then a sort filter could be applied within the parameter dropdown. Although if we really think we need to manually resort things, we can.

mvdbeek: I'm thinking about tool tests and equality of metadata when caching jobs here. If some of the metadata elements are returned with a random order that might be a little annoying. If it's the same order it's fine for me, and we can always fix it later (if it's not).

mvdbeek (approved):

dannon:
The data manager test that failed before passed this time, too!  We'll have to track that down some other time.

blankenberg:
Off topic, but worth burning an [hour](https://www.youtube.com/watch?v=z1aVDael-KM) of time...

blankenberg:
Thanks all for the comments, reviews, and getting the tests to finally pass, I had started to give up after the 3rd-or-so unrelated failure.

PR #6726, opened 2018-09-17
[18.09] Badge default fix

### LINKED ISSUE #6715: 'deprecated' label in toolpanel is missing style
As seen on Main.

screenshot 2018-09-14 13 18 29

dannon:
Minor fix to correct the default presentation of tool badges without a specially defined color.  Includes a doc fix and client format.

Fixes #6715

### DIFF
--- a/client/galaxy/docs/bootstrap.md
+++ b/client/galaxy/docs/bootstrap.md
@@ -135,8 +135,8 @@ Disabled
 
 ```
 <div>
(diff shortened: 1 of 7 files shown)


jmchilton:
I've seen that new unit test fail a couple times now - I think it needs a loop and a sleep instead of just a sleep. I'll see what I can do.

PR #6728, opened 2018-09-17
[18.09] remove html tags and newlines from the help text

martenson:
this should improve the toolsearch for certain keywords
and also make the search index smaller

### DIFF
--- a/lib/galaxy/tools/search/__init__.py
+++ b/lib/galaxy/tools/search/__init__.py
@@ -6,6 +6,7 @@ import logging
 import re
 import tempfile
 
+from bleach import clean
 from whoosh import analysis
 from whoosh.analysis import StandardAnalyzer
 from whoosh.fields import (
@@ -25,7 +26,6 @@ from galaxy.util import ExecutionTimer
 from galaxy.web.framework.helpers import to_unicode
 
 log = logging.getLogger(__name__)
-HTML_CLEANER = re.compile("<[^<]+?>")
 
 
 class ToolBoxSearch(object):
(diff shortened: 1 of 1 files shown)


lib/galaxy/tools/search/__init__.py
```diff
@@ -25,6 +25,7 @@
 from galaxy.web.framework.helpers import to_unicode
 
 log = logging.getLogger(__name__)
+HTML_CLEANER = re.compile("<[^<]+?>")
```
martenson: poor man's beautiful soup/lxml, please offer better alternatives if you have

nsoranzo: https://bleach.readthedocs.io/en/latest/ could be an option, since we already use it elsewhere.

martenson: Do you have any experience with its speed?

martenson: so it does not really behave as I want it to:

```python
>>> import bleach
>>> bleach.clean('test<strong>sdsd</strong>', strip=True)
u'test<strong>sdsd</strong>'
>>> bleach.clean('test<span></span>', strip=True)
u'test'
```

martenson: it seems I need to define an empty list of allowed tags, then it works well, commit inc.

nsoranzo (approved):

martenson:
just for fun: previously you could search for html tag contents, eg `strong` and you would get the tools that use the most bold font :D

screenshot 2018-09-18 09 48 12

Issue #6569, opened 2018-08-05
Issues with Nested Workflows which take Collections as Input


alexlenail:
I have a workflow I'd like to assemble from two subworkflows. Let's call the top-level workflow A and the subworkflows B and C. 

B takes a collection of pairs of fastqs and returns a collection of bams. 
C takes a collection of bams and returns a collection of small files and plots. 

I'd like to build workflow A which takes a collection of fastqs and returns a collection of small files and plots. I'd like workflow A to be very simple: 
```
input collection -----> B -----> C -----> done. 
```
However, I can't drag a noodle from B to C. The noodle I drag from the input collection to B is also not a "multi"-noodle. It's clear the workflow UI thinks the output of B is a single file, though I know it isn't from running B individually. 

----

Bonus issue: 

When I add subworkflows to a workflow, e.g. in the workflow editor UI for A, I select B from the side panel, it appears in the workflow UI with its old name. That is, I have renamed that workflow, it shows up on the sidebar with its new name, but once I click it, a tool shows up in the editor UI with the old name. 

(cc @bgruening)

mvdbeek:
The non-bonus part should work after #6732, the bonus part should be a separate issue with a simple example.

Issue #6514, opened 2018-07-18
Filter Failed does not take a collection output from a sub-worfklow as an acceptable input


mtekman:
![out2](https://user-images.githubusercontent.com/20641402/42897089-2f196b40-8abf-11e8-8118-915bc65c7be2.gif)

The sub-workflow takes a list:paired collection as input (oddly, this is only represented as a single noodle) and outputs a collection of text files (also only represented as a single noodle).

The list:paired input *does* connect to Filter Failed, but as you can see from the Gif above, the output collection from the sub-workflow does not.

mtekman:
Related to https://github.com/galaxyproject/galaxy/issues/6012

jmchilton:
Do you have any idea if your Galaxy includes (https://github.com/galaxyproject/galaxy/pull/6178) - that might fix the problem or might not.

hexylena:
Yes that is included, and we have rebuilt the client as of this morning due to an unrelated issue. @mtekman could you try it again now that we've rebuilt the client?

mtekman:
Still no change on my side, the noodle fails to connect on either input or output as shown in the gif previously

bgruening:
The same workflow has an additional problem as it seems, so I think this might be related. If it is executed all jobs from the subworkflow are scheduled and are running, the last step, the one after the subworkflow (column join) is not even scheduled and does not appear in the history.

The input in the subworkflow are two files and one paired-list, the expected output is one file. Now we are feeding a collection of paired-lists into the subworkflow so the expectation would be a list as output. But this seems to be broken as it seems. 
I would also expect a fat-noodle from paired-collections to the subworkflow and from the subworkflow to the column-join tool.

The workflow from above is here: https://usegalaxy.eu/u/mehmet-tekman/w/celseq2-new-super-workflow
The subworkflow is here: https://usegalaxy.eu/u/mehmet-tekman/w/celseq2-nodemulti-todo-update-featurecounts-include-umitools-count

mvdbeek:
I'm working on this -- there are multiple problems, one of which is that we don't look up the `collection_type` if `collection_type_source` source is specified. This works OK for regular inputs, but not for instance `multiple="true"` inputs. In the same direction we need to update the collection type description when we connect and disconnect tools.

mvdbeek:
Should be done in https://github.com/galaxyproject/galaxy/pull/6732 -- it'll be in 18.09

Issue #6012, opened 2018-04-30
Filter Failed tool in workflow coerces wrong noodle type when connecting collection data


mtekman:
Hi, I'm not sure whether this is really a bug with a workflow, or an issue with these specific tool inputs.

1 .Connecting a Collection output to a tool expecting multiple inputs:

  * Tool1 → {1:N} → Tool2 → {N:N} → ToolN.

![edited_1](https://user-images.githubusercontent.com/20641402/39426510-78fe946a-4c80-11e8-9a07-4245b98ce838.gif)

2. Connecting a Collection output to `Filter_Failed` and then connecting to a tool expecting multiple inputs:

  * Tool1 → {1:1} → FilterFailed → {1:N} → Tool2 → {N:N} → ToolX
  * (This is what I expect to be working fine)

![edited_2](https://user-images.githubusercontent.com/20641402/39427341-b302f5cc-4c83-11e8-913e-e1da41b1a5f9.gif)

3. However, the workflow does not actually like the above arrangement because any attempts to create more output noodles are refused (causing the tool to be dragged instead).

  * Tool1 → {1:1} → FilterFailed →  {1:N} → Tool2  → {N:N} → Stuck!

![edited_3](https://user-images.githubusercontent.com/20641402/39426827-a508da10-4c81-11e8-922a-d5b9003914f6.gif)

4. Attempting to rectify this by linking downstream tools first via 1:1 noodles, and then linking Filter Failed tool does not work:

 * Tool1 → {1:1} → FilterFailed → {Stuck!} → Tool2 → {1:1} → ToolN

![edited_4](https://user-images.githubusercontent.com/20641402/39426949-2503016e-4c82-11e8-9279-583b2462dd38.gif)

 * Tool1 → {Stuck!} → Tool2 → {1:1} → ToolN 

![edited_5](https://user-images.githubusercontent.com/20641402/39427062-a2b5e52c-4c82-11e8-87d5-deb2b77c4897.gif)

5. Unlinking all and re-linking as per our initial setup works fine:

    * Tool1 → {1:N} → Tool2 → {N:N} → ToolN.
![edited_6](https://user-images.githubusercontent.com/20641402/39427116-e47a34ae-4c82-11e8-8580-5707dbd42719.gif)

guerler:
Possibly related to #5205

mvdbeek:
I should have a fix for this soon. xref https://github.com/galaxyproject/galaxy/issues/6514

[...]

Issue #1889, opened 2016-03-08
Multiple inputs do not work across subworkflows


bgruening:
The right workflow shows a concat tool that can take `n` input files. If I use this workflow as a subworkflow (the left image) I can not insert multiple input files. The output from `get_pos_data` is not able to connect to `selex_dataset_postprocessing`.

This would work if I define `n` inputs in the workflow on the right, but usually `n` is not known in advance and you would like to include as many inputs as needed - also in a subworkflow.

![multiple_inputs_subworkflows](https://cloud.githubusercontent.com/assets/469983/13597442/f0bd7c22-e518-11e5-83d2-d5859e89e19b.png)

xref: https://github.com/galaxyproject/galaxy/issues/1387
ping @jmchilton @guerler

jmchilton:
One can imagine solving this by introducing a multi-input module (or a flag on the input module) or by introducing module to build a list from inputs and switching the input on the right to a collection input. The problem with the former is that it would be very unclear to users when to use muliti-inputs versus collection inputs. The problem with the latter is that collection is unnatural and seemingly superfluous. I guess I prefer the second problem - it is closer to what we have now (less to implement) and more importantly it doesn't introduce a concept we may later regret. So maybe when I re-open a PR for collection operations - I can include a "build list" from multiple input operation?

bgruening:
Sounds like a plan!

descostesn:
I still have this problem, any update?

PR #6732, opened 2018-09-18
Look up collection_type if collection_type_source is set

mvdbeek:
We do this in the wf-editor and in the backend, so this works for
subworkflows. This should make quite some tools more usable in a
workflow scenario and fixes the noodle being of the wrong type 
when connecting subworkflow outputs. Should fix
- https://github.com/galaxyproject/galaxy/issues/6514
- https://github.com/galaxyproject/galaxy/issues/6012
- https://github.com/galaxyproject/galaxy/issues/6569
- https://github.com/galaxyproject/galaxy/issues/1889

and perhaps more issues related to this.

### DIFF
--- a/client/galaxy/scripts/mvc/workflow/workflow-connector.js
+++ b/client/galaxy/scripts/mvc/workflow/workflow-connector.js
@@ -1,3 +1,5 @@
+import * as Toastr from "libs/toastr";
+
 function Connector(handle1, handle2) {
     this.canvas = null;
     this.dragging = false;
@@ -30,8 +32,11 @@ $.extend(Connector.prototype, {
         }
         $(this.canvas).remove();
     },
-    destroyIfInvalid: function() {
+    destroyIfInvalid: function(warn) {
         if (this.handle1 && this.handle2 && !this.handle2.attachable(this.handle1)) {
+            if (warn) {
+                Toastr.warning("Destroying a connection because collection type has changed.");
+            }
             this.destroy();
         }
     },
--- a/client/galaxy/scripts/mvc/workflow/workflow-terminals.js
+++ b/client/galaxy/scripts/mvc/workflow/workflow-terminals.js
@@ -131,6 +131,7 @@ var Terminal = Backbone.Model.extend({
         if (this.node) {
             this.node.markChanged();
             this.resetMappingIfNeeded();
+            this.resetCollectionTypeSource();
         }
     },
     redraw: function() {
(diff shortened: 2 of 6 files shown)


client/galaxy/scripts/mvc/workflow/workflow-terminals.js
```diff
@@ -438,6 +438,14 @@ var InputCollectionTerminal = BaseInputTerminal.extend({
         var other = connector.handle1;
         if (!other) {
             return;
+        } else {
+            let node = this.node;
+            _.each(node.output_terminals, function(output_terminal) {
```
mvdbeek, line 456: Still need to do the opposite when removing the input connection!

mvdbeek: Done with 714481cc3201a65d6974cc221f713409a81e4a51

client/galaxy/scripts/mvc/workflow/workflow-connector.js
```diff
@@ -30,8 +32,11 @@ $.extend(Connector.prototype, {
         }
         $(this.canvas).remove();
     },
-    destroyIfInvalid: function() {
+    destroyIfInvalid: function(warn) {
         if (this.handle1 && this.handle2 && !this.handle2.attachable(this.handle1)) {
+            if (warn) {
+                Toastr.warning("Destroying a connection because collection type has changed.");
+            }
```
jmchilton: Very nice!

lib/galaxy/managers/workflows.py
```diff
@@ -591,15 +591,30 @@ def callback(input, prefixed_name, **kwargs):
...
-    def _workflow_to_dict_export(self, trans, stored=None, workflow=None, version=None):
-        """ Export the workflow contents to a dictionary ready for JSON-ification and export.
+    def _resolve_collection_type_source(self, steps):
         """
-        if workflow is None:
-            assert stored is not None
-            workflow = stored.get_internal_version(version)
+        Fill in collection type for step outputs that infer the collection type via collection_type_source.
 
+        This information is only needed in the workflow editor.
```
jmchilton, line 601: I caught on this at first, but the commit comment is good. The editor will do this also and needs to because the source type can change from the form change, but it can't infer that for subworkflows (which luckily also can't change in the editor). Very nice.

jmchilton:
This is so important and looks really stellar. I imagined this would be more frontend code. Really, profoundly super awesome - thanks a million as always @mvdbeek!

PR #6737, opened 2018-09-19
[18.05] Decode tool id and version in the client, not in the API 

nsoranzo:
using `decodeUriComponent` from [decode-uri-component](https://github.com/SamVerschueren/decode-uri-component) JS package, which also decodes `+` to a space.

Revert #6707 .

Remove unused `_get_tool()` method in `lib/galaxy/webapps/galaxy/api/workflows.py` .

Rebuild client with:
```
make client-production-maps
```

Also:
- Remove (again) the unused `lib/galaxy/dependencies/requirements.txt` file.
- Fix symlinking during plugin staging by removing also broken symlinks before creating the new ones (needed for local Selenium testing).


nsoranzo:
The selenium test failure is legit, working on it.

nsoranzo:
All green!

mvdbeek:
Looks good in my tests too!

mvdbeek (approved):

mvdbeek:
@dannon can you confirm that we should do `make client-production-maps` or is it `make client-production`

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -545,15 +544,6 @@ def build_module(self, trans, payload={}):
     #
     # -- Helper methods --
     #
-    def _get_tool(self, id, tool_version=None, user=None):
-        id = unquote(id)
-        if tool_version:
-            tool_version = unquote(tool_version)
-        tool = self.app.toolbox.get_tool(id, tool_version)
-        if not tool or not tool.allow_user_access(user):
```
jmchilton, line 553: This allow_user_access check happens elsewhere?

jmchilton: Oh I see it was already unused.

dannon:
@mvdbeek I noticed that too, in a recent 18.05 commit I had to make including a rebuild.  We'd agreed to use just `production` for 18.05, but at some point (maybe even at the start, and I just didn't notice) `production-maps` entered the codebase and it's not worth changing gears on it mid-release with the big diffs.  Might as well stick with `make client production-maps` for the rest of 18.05.

lib/galaxy/webapps/galaxy/api/workflows.py
nsoranzo: Yep, this method is never called since commit 3e3eb4d1397effc8b46c913bb80f3c02773ac506 .

nsoranzo:
@dannon That's also what `client/README.md` still says:
```
Important Note: The development branch of Galaxy does not include client script
artifacts, and these should not be committed.  When issuing a PR to a stable
branch, please run "make client-production-maps", and include those artifacts.
Or, if you'd rather, include only the /client source changes and build
artifacts can be added by maintainers on merge.
```

dannon:
@nsoranzo Good point!  I'll remove that; per conversation on gitter right now I think we're ready to pull off the bandaid and stop commiting them (even for release branches) moving forward.

jmchilton (approved):

martenson (approved):

martenson:
Thanks @nsoranzo and @mvdbeek . I hope we finally squished this one! 🧐

PR #6740, opened 2018-09-19
[18.05] Backport tool version select fixes

mvdbeek:
That's a backport of https://github.com/galaxyproject/galaxy/pull/6702 and https://github.com/galaxyproject/galaxy/pull/6710.
Without this you would get the oldest tool version if the tool version didn't match, instead of the newest as intended

### DIFF
--- a/test/api/test_tools.py
+++ b/test/api/test_tools.py
@@ -140,8 +140,11 @@ class ToolsTestCase(api.ApiTestCase):
             assert output_details["state"] == "error", output_details
             assert "has not sent back a URL parameter" in output_details["misc_info"], output_details
 
-    def _show_valid_tool(self, tool_id):
-        tool_show_response = self._get("tools/%s" % tool_id, data=dict(io_details=True))
+    def _show_valid_tool(self, tool_id, version=None):
+        data = dict(io_details=True)
+        if version:
+            data['version'] = version
+        tool_show_response = self._get("tools/%s" % tool_id, data=data)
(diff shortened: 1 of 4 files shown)


test/api/test_tools.py
```diff
@@ -436,6 +439,13 @@ def test_run_by_versions(self):
             output1_content = self.dataset_populator.get_history_dataset_content(history_id, dataset=output1)
             self.assertEqual(output1_content.strip(), "Version " + version)
 
+    @skip_without_tool("multiple_versions")
+    @uses_test_history(require_new=False)
```
nsoranzo: You need to add `uses_test_history` to the imports.

mvdbeek: That didn't exist in 18.05, but I can use the old style populator

mvdbeek: Uh, I don't even need a history here

test/api/test_tools.py
```diff
@@ -440,7 +440,6 @@ def test_run_by_versions(self):
             self.assertEqual(output1_content.strip(), "Version " + version)
 
     @skip_without_tool("multiple_versions")
-    @uses_test_history(require_new=False)
     def test_show_with_wrong_tool_version_in_tool_id(self, history_id):
```
jmchilton: Well now that history_id will not have a value in this method definition.

mvdbeek: yeah, that's silly ...

nsoranzo (approved):

Issue #6675, opened 2018-09-07
Cannot submit bug report as anonymous user (18.05)


loraine-gueguen:
Having the same issue as https://github.com/galaxyproject/galaxy/issues/5195 in 17.09 (white page displaying instead of "report error" form and 403 error in logs), I have updated to 18.01, but I had the same issue. I have then updated to 18.05: the "report error" form is now displayed, but there is another 403 error in logs when clicking on the report button after filling the form in (see below), and an error message appears on the top of the Galaxy window "Error occured while saving. Please fill all the required fields and try again."

"POST /api/jobs/30858b0ba227edd6/error HTTP/1.1" 403 - "http://webtools.sb-roscoff.fr/datasets/error?dataset_id=xxxxx&__identifer=yyyy" "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:56.0) Gecko/20100101 Firefox/56.0"

mvdbeek:
Thanks for the report, that is indeed unfortunate. We fixed this with https://github.com/galaxyproject/galaxy/pull/6742, which includes a test that will hopefully prevent us form breaking this again in the future.

loraine-gueguen:
Thanks for the fix!

PR #6742, opened 2018-09-19
[18.05] Fix and test anonymous error reports

mvdbeek:
Fixes https://github.com/galaxyproject/galaxy/issues/6675, this time with tests.

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/jobs.py
+++ b/lib/galaxy/webapps/galaxy/api/jobs.py
@@ -320,7 +320,7 @@ class JobController(BaseAPIController, UsesLibraryMixinItems):
                 jobs.append(job)
         return [self.encode_all_ids(trans, single_job.to_dict('element'), True) for single_job in jobs]
 
-    @expose_api
+    @expose_api_anonymous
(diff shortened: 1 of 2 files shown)


jmchilton (approved):

jmchilton:
Failing tests look certainly unrelated. +1, thanks for the awesome fix and tests.

lib/galaxy/webapps/galaxy/api/jobs.py
```diff
@@ -343,10 +343,17 @@ def error(self, trans, id, **kwd):
         # Get job
         job = self.__get_job(trans, id)
         tool = trans.app.toolbox.get_tool(job.tool_id, tool_version=job.tool_version) or None
+        email = kwd.get('email')
+        if not email and not trans.anonnymous:
```
martenson: `trans.anonymous` I think (extra `n` here)

mvdbeek: Oops, thanks, I've rebased the commit with the fix and I'm testing that route as well now.

PR #6744, opened 2018-09-19
[18.09] dispose of dataset tooltip when the delete is being called

### LINKED ISSUE #6322: history dataset tooltip visible after element removed
seen on Main

screenshot 2018-06-12 14 59 25

martenson:
fixes #6322


dannon:
Tested, this works and cleans up the delete tooltips.

PR #5807, opened 2018-03-29
Bootstrap 4 !!!!

dannon:
Since this is a very large project, tracking progress here: https://github.com/galaxyproject/galaxy/projects/10

This branch upgrades us to bootstrap 4.  All styles have been converted from .less to .scss, and build infrastructure has been implemented.  All of our client grunt tasks with the exception of spritesmith have been migrated to the simpler, faster gulp tasks.  Building the client is still just 'make client'.  Other tasks have been updated like 'client-watch' and 'style-watch'.

At this point, I'm mostly tracking down and fixing styles that didn't upgrade correctly (backwards compatability was significantly broken with bootstrap4).  I've been trying to modernize everything I can along the way, as well -- there's a *lot* of manually styled cruft that we need to fix or we're going to be carrying technical debt forever.

**styleguide:** `yarn styleguide` will serve a styleguide generated from markdown files under `client/galaxy/docs` with live reloading. You can use this for editing stylesheets. Add any Galaxy markup that is not working to `galaxy-styles.md` and then edit the bootstrap variables or scss to make things work.

I've been working on this for a while now, and am opening the PR to solicit help.  If you're even remotely interested in this effort, please ping me and I'd be happy to loop you in if you're willing to work on tracking down and fixing remaining migration issues (which I'm working on documenting now).

PR #6745, opened 2018-09-19
[18.09] Generic tooltip hide for onclick faIconButtons

dannon:
@martenson After I merged #6744 I looked at it more -- this change should make the core faIconButton click handler responsible for hiding tooltips on click, so we don't have to have individual onclick functions doing it.  This fixes other stuck tooltips that were using faIconButtons, too.

I was torn about before/after the click handler, but before makes it that much snappier and the interface *feels* better to me.

### DIFF
--- a/client/galaxy/scripts/mvc/dataset/dataset-li-edit.js
+++ b/client/galaxy/scripts/mvc/dataset/dataset-li-edit.js
@@ -104,7 +104,6 @@ var DatasetListItemEdit = _super.extend(
                 faIcon: "fa-times",
                 classes: "delete-btn",
                 onclick: function() {
-                    self.$el.find(".icon-btn.delete-btn").tooltip('dispose');
                     self.model["delete"]();
                 }
             });
--- a/client/galaxy/scripts/mvc/dataset/dataset-li-edit.js
+++ b/client/galaxy/scripts/mvc/dataset/dataset-li-edit.js
@@ -104,6 +104,7 @@ var DatasetListItemEdit = _super.extend(
                 faIcon: "fa-times",
                 classes: "delete-btn",
                 onclick: function() {
+                    self.$el.find(".icon-btn.delete-btn").tooltip("dispose");
                     self.model["delete"]();
                 }
             });
--- a/client/galaxy/scripts/ui/fa-icon-button.js
+++ b/client/galaxy/scripts/ui/fa-icon-button.js
@@ -1,6 +1,5 @@
 import jQuery from "jquery";
 import * as _ from "underscore";
-("use_strict");
 
 var $ = jQuery;
 //============================================================================
@@ -43,12 +42,11 @@ var faIconButton = options => {
     ].join("");
     var $button = $(html).tooltip(options.tooltipConfig);
     if (_.isFunction(options.onclick)) {
-        let wrapped = _.wrap(options.onclick, function(func) {
+        $button.click(function(ev){
             $button.tooltip("hide");
             $button.parent().attr('tabindex', -1).focus();
-            func.apply(this, arguments);
+            options.onclick.apply(this, arguments);
         });
-        $button.click(wrapped);
     }
     return $button;
 };
--- a/client/galaxy/scripts/ui/fa-icon-button.js
+++ b/client/galaxy/scripts/ui/fa-icon-button.js
@@ -42,9 +42,12 @@ var faIconButton = options => {
     ].join("");
     var $button = $(html).tooltip(options.tooltipConfig);
     if (_.isFunction(options.onclick)) {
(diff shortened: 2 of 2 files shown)


martenson:
IMHO the standard `onclick` is not supposed to hide the tooltips I think, that is the point where you appreciate the tooltip being visible the most. 

Also on delete we should be calling `destroy/dispose` on the tooltip, since just `hide` might leave the DOM element orphaned without a reference.

dannon:
@martenson Do you have examples of leaving tooltips shown on button click (in other websites)?  That's very counterintuitive to me -- the tooltip is for *before* I've clicked, not after.  I thought we always hid them before, it was only broken with the bs4 upgrades.

So, more digging, ARIA-implemented tooltips hide as soon as you click, which is the behavior I've implemented here.  I believe this PR is a correct change.

martenson:
This is how they are designed in BS4: https://getbootstrap.com/docs/4.0/components/tooltips/

few quick examples of tooltips that I consider useful when they stay:

screenshot 2018-09-19 14 26 04
screenshot 2018-09-19 14 28 28
screenshot 2018-09-19 14 28 10
screenshot 2018-09-19 14 32 15

dannon:
When I first updated us to Bootstrap 4 folks hated the sticky-on-focus tooltips and thought they were a bug, and I had to argue that they're designed to keep the tooltip up on focus as an accessibility feature -- e.g. for users who cannot use a mouse, when tabbing through inputs the tooltip will display and stay until `blur()` happens.  This, I agree with (and as stated, argued for, before), though once you've clicked a button (unless it's a toggle, in which case the toggle behavior happens and stays -- though I think the tooltip should still go) it should blur and the tooltip should go away.

This fixes us having to do stuff like this:  https://github.com/galaxyproject/galaxy/pull/5807/commits/187fde6be5ce10a18cd494bc043b93a37b8eeaec, which breaks accessibility concerns (tooltips unavailable on focus *and* hover, which is the new default, just because we didn't want them being persistent on click)

Maybe I don't understand the screenshot examples above, but all of those tooltips seem like they should go away on click to me.  If there's more contextual information about an interaction that we need to communicate to a user, I don't think a sticky profusion of tooltips is the right way about it.

mvdbeek:
I hate that the tooltips stick and I hate the tooltips, but I can tolerate them if they're not in my way. If I need to click to make them go away that just slows me down, and I'm under the impression that doesn't work consistently at this time anyway (if this PR fixes that: 💯 ). The examples that @martenson posted should IMO go away once I'm not hovering anymore.

dannon:
@mvdbeek Yes, that's exactly what this PR makes steps towards fixing -- anywhere using the faIconButton abstraction.  The reason they 'stick' is generally that the button retains focus on clicks, and bootstrap4's tooltips are by default activated on focus *and* hover.   We *want* both of those things, for accessibility in tooltips when tabbing through (so, focus but not hover) and when just hovering (no focus) -- we just want them to go away after the interaction.

mvdbeek:
> We _want_ both of those things, for accessibility in tooltips when tabbing through (so, focus but not hover) and when just hovering (no focus) -- we just want them to go away after the interaction.

I don't know about that either. How many times do I need to know that `x` deletes a dataset ?
I'd argue we're annoying every Galaxy user that knows the interface. But I get that's not the point of the PR, so 👍 on the intention here.

martenson:
I don't understand how removing focus from an element (which user just focused on) is helping usability. The focus is there so the next TAB can move user to the next focusable element. Remove it and you break the path.

I could probably be persuaded that the tooltip behavior should be consistent across the site and that the consistent choice should be hiding it. But not for the cost of a focus.

dannon:
@mvdbeek Yeah, I agree with you there, we may be overusing tooltips.  When I was spending way too much time on this small change trying to sort out best practices from other sites, most folks just don't have near as many tips as we do.  Maybe we should add more of a delay on tooltip display or something?

@martenson I don't know how common (or even realistic) that particular scenario is where you tab to something, click it using enter, then need to continue tabbing in the same taborder to click something else.  It seems like a mixed bag between resetting the tab order being the right thing (because some new interface shows up when you click the button) and between wanting to continue.  In any event, even if I was only using the keyboard, I'd be more annoyed by the sticky focused buttons and tooltips than I would be by having to tab again.  If you're very worried about this, I can figure out a way to keep the tabindex correctly and still blur on click -- just let me know.

(For what it's worth, the tab order already resets after hitting delete since it's effectively a blur, with the object being deleted)

mvdbeek (approved):

dannon:
(the API test that failed on a previous run is completely unrelated)

dannon:
@mvdbeek Thank you.

PR #6746, opened 2018-09-19
Fill in needed state defaults when importing manually crafted workflows.

jmchilton:
This makes format 2 workflows more palatable since unset default values in conditionals/repeats don't need to be specified when imported this way (xref https://github.com/galaxyproject/planemo/blob/master/tests/data/wf4-distro-tools.gxwf.yml#L18).


mvdbeek (approved):
Awesome, that looks very nice to me.

PR #6748, opened 2018-09-20
[18.09] Make newer unit tests more robust with a looped wait...

jmchilton:
... instead of raw sleep.

This test also fails on Mac OS X, but that is a different problem I think (no watchdog wheel?).

### DIFF
--- a/test/unit/tools/test_toolbox.py
+++ b/test/unit/tools/test_toolbox.py
@@ -212,19 +212,38 @@ class ToolBoxTestCase(BaseToolBoxTestCase):
         tool_path = tool.config_file
         with open(tool.config_file, 'w') as out:
             out.write('certainly not a valid tool')
-        time.sleep(1.5)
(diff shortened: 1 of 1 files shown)


mvdbeek (approved):

mvdbeek:
This test shouldn't be using watchdog, this should be a manual poll IIRC. I did write the test on OSX (and I just ran the test again), so I'm not sure what the cause of the failure could be ? Is there a traceback or some other log that would indicate something ?

nsoranzo:
> This test also fails on Mac OS X, but that is a different problem I think (no watchdog wheel?).

There definitely are watchdog wheels for OS X: https://wheels.galaxyproject.org/simple/watchdog/

nsoranzo (approved):

jmchilton:
I saw this fail again even with this change, I'm not sure what is up.

PR #6749, opened 2018-09-20
[18.09] Use mapped-over collectionType when inferring collectionType through …

mvdbeek:
…format_source

Followup to https://github.com/galaxyproject/galaxy/pull/6741 -- once I updated the production server I noticed that the usecase I was trying to enable was still not working because the input was mapped over.

### DIFF
--- a/client/galaxy/scripts/mvc/workflow/workflow-terminals.js
+++ b/client/galaxy/scripts/mvc/workflow/workflow-terminals.js
@@ -131,7 +131,7 @@ var Terminal = Backbone.Model.extend({
             this.node.markChanged();
             this.resetMappingIfNeeded();
             if (!connector.dragging) {
-                this.resetCollectionTypeSource();
+                connector.handle2.resetCollectionTypeSource();
             }
         }
     },
--- a/client/galaxy/scripts/mvc/workflow/workflow-terminals.js
+++ b/client/galaxy/scripts/mvc/workflow/workflow-terminals.js
@@ -456,7 +456,11 @@ var InputCollectionTerminal = BaseInputTerminal.extend({
             _.each(node.output_terminals, function(output_terminal) {
                 if (output_terminal.attributes.collection_type_source && !connector.dragging) {
                     if (other.isMappedOver()) {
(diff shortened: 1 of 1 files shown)


mvdbeek:
This is the frontend fix, so in the same workflow you can now start with `list` -> `cat1` -> `filter failed` (or any tool that uses `type_source`). I'd also need to do this on the backend if you want to include such a workflow in a subworkflow, but that looks like it'd be a bit trickier.

client/galaxy/scripts/mvc/workflow/workflow-terminals.js
```diff
@@ -455,7 +455,11 @@ var InputCollectionTerminal = BaseInputTerminal.extend({
             let node = this.node;
             _.each(node.output_terminals, function(output_terminal) {
                 if (output_terminal.attributes.collection_type_source && !connector.dragging) {
-                    output_terminal.attributes.collection_type = other.attributes.collection_type;
+                    if (other.isMappedOver()) {
+                        output_terminal.attributes.collection_type = other.terminalMapping.mapOver.collectionType;
```
jmchilton: Does this work if you're mapping over a collection output type? ... and if there is subcollection mapping on the inputs? There is a lot of cases to handle here but this change is at very least a step in the right direction so I'm definitely 👍. There are some client unit test cases trying to capture the map/reduce logic for these things - I should probably update those with collection_type_source outputs to simulate some of these more complicated cases - but the tests are a bit of a mess.

mvdbeek: I haven't tried that yet, it might not. Good point about the tests, I'm essentially flying blind here.

jmchilton (approved):

client/galaxy/scripts/mvc/workflow/workflow-terminals.js
mvdbeek: You're right, in that case I think it should be
```
other.terminalMapping.mapOver.append(other.collectionType)
```

mvdbeek:
I think this works OK now, the backend fix for subworkflows is probably more work.

PR #6750, opened 2018-09-20
[18.09] Undo auto-propagation of group tags.

jmchilton:
@mvdbeek and I discussed this in the Galaxy project gitter a few weeks ago and decided the propagation is too aggressive and should likely be done manually at the end of a workflow for now - using for instance the set tags collection operation tool.


mvdbeek (approved):
Thanks @jmchilton!

PR #6751, opened 2018-09-20
[WIP] Direct LSF runner

pcm32:
This PR introduces a direct LSF runner that doesn't require DRMAA. It is based in the condor runner.

### DIFF
--- a/lib/galaxy/jobs/runners/lsf.py
+++ b/lib/galaxy/jobs/runners/lsf.py
@@ -88,7 +88,6 @@ class LSFJobRunner(AsynchronousJobRunner):
             executable=executable,
             output=ljs.output_file,
             error=ljs.error_file,
-            user_log=ljs.user_log,
             query_params=query_params,
         )
 
--- a/lib/galaxy/jobs/runners/lsf.py
+++ b/lib/galaxy/jobs/runners/lsf.py
@@ -90,12 +90,13 @@ class LSFJobRunner(AsynchronousJobRunner):
         submit_file = os.path.join(cluster_directory, 'galaxy_%s.lsf.sh' % galaxy_id_tag)
         executable = ljs.job_file
 
+        query_params.update({'working_dir': os.path.abspath(job_wrapper.working_directory)})
+
         build_submit_params = dict(
             executable=executable,
             output=ljs.output_file,
             error=ljs.error_file,
             user_log=ljs.user_log,
-            working_dir=os.path.abspath(job_wrapper.working_directory),
             query_params=query_params,
         )
 
--- a/lib/galaxy/jobs/runners/lsf.py
+++ b/lib/galaxy/jobs/runners/lsf.py
@@ -72,7 +72,7 @@ class LSFJobRunner(AsynchronousJobRunner):
                 # HTCondor needs the image as 'docker_image'
                 query_params.update({'docker_image': container.container_id})
 
-        galaxy_slots = query_params.get('request_cpus', None)
+        galaxy_slots = query_params.get('cores', None)
         if galaxy_slots:
             galaxy_slots_statement = 'GALAXY_SLOTS="%s"; export GALAXY_SLOTS_CONFIGURED="1"' % galaxy_slots
         else:
@@ -167,6 +167,8 @@ class LSFJobRunner(AsynchronousJobRunner):
             job_id = cjs.job_id
             galaxy_id_tag = cjs.job_wrapper.get_id_tag()
             try:
+                # TODO have a way of buffering requests to LSF, so that
+                # it doesn't get flooded.
                 #if os.stat(cjs.user_log).st_size == cjs.user_log_size:
                 #    new_watched.append(cjs)
                 #    continue
--- a/lib/galaxy/jobs/runners/lsf.py
+++ b/lib/galaxy/jobs/runners/lsf.py
@@ -115,9 +115,8 @@ class LSFJobRunner(AsynchronousJobRunner):
 
         cleanup_job = job_wrapper.cleanup_job
         try:
-            open(submit_file, "w").write(submit_file_contents)
-            st = os.stat(submit_file)
-            os.chmod(submit_file, st.st_mode | stat.S_IEXEC)
+            with open(submit_file, "w") as sub:
+                sub.write(submit_file_contents)
         except Exception:
             if cleanup_job == "always":
                 ljs.cleanup()
@@ -145,8 +144,6 @@ class LSFJobRunner(AsynchronousJobRunner):
                 ljs.cleanup()
             job_wrapper.fail("LSF bsub failed", exception=True)
             return
-
-        os.unlink(submit_file)
 
         log.info("(%s) queued as %s" % (galaxy_id_tag, external_job_id))
 
@@ -195,6 +192,8 @@ class LSFJobRunner(AsynchronousJobRunner):
                     if external_metadata:
                         self._handle_metadata_externally(cjs.job_wrapper, resolve_requirements=True)
                     log.debug("(%s/%s) job has completed" % (galaxy_id_tag, job_id))
+                    # TODO add deletion of submit file here
+                    # os.unlink()
                     self.work_queue.put((self.finish_job, cjs))
                 continue
             if job_failed:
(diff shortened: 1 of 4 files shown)


bgruening:
You are amazing @pcm32! :)

bernt-matthias:
I was thinking often if the [qsb]sub, [qsb]acct, [qsb]stat, [qsb]del based runners can be unified into one runner. 

The executed command lines are highly similar. The biggest difference if the evaluation of the finished jobs (detecting errors, etc). I imagine that it might be sufficient to have specialized command line templates and a specialized finish function for each runner.

Pro: 
- The python drmaa library can not be used when jobs run as real user anyway. So the slurm and drmaa runners need  to implement the corresponding del, stat, and acct commands anyway.
- Having a single runner reduces maintenance, because users of the different runners can collaborate on a single runner. 
- Might enforce all runners have the same functionality. Currently they seem to diverge, eg. some allow submitting jobs as real user others don't.

Con: 
- Its more effort in the beginning. 

Anyway, maybe I can encourage you to check in the end for the differences in the condor and your LSF runner and maybe to come  up with a generalized runner and derived condor and LSF runners. 

Then, grid engine, torque, and slurm could follow ... and in the end python drmaa would not be required ... eventually.

pcm32:
I totally agree that having a properly unified setup is the way forward... my only problem is that I don't have the man power for it currently :-(, but if others do, I'm all onboard with helping on the side. I can initially provide this quick and dirty one (which is still work in progress), but surely the process of doing it will give us a good understanding of what is general and what is particular towards a more object oriented approach.

bernt-matthias:
@pcm32 Agreed. I completely understand. 

I will follow the progress and see what I can do in the end ..

pcm32:
Question: where should I handle the resources consumed by the job? I'm thinking within the `check_watched_items` when the job is detected to be finished, or is there another standard place to do that? Is there an standard place where this data (CPU usage, RAM usage, swap usage) would end up (where do I feed what I can get from LSF for these?). I have been looking at the condor, pbs and drmaa runner, but I don't seem to find what I'm after. I would like to record into the job metadata the consumption of RAM and CPU essentially, for later mining/reporting. Maybe @bgruening or @jmchilton could give me some directions?

bernt-matthias:
As far as I remember `check_watched_item(s)` is the correct place. You only need set the job state there, i.e. check if to much run time / memory was used. As far as I know the cpu/mem usage is not stored. See also: https://docs.galaxyproject.org/en/master/dev/build_a_job_runner.html?highlight=runner

pcm32:
@bgruening pointed me to https://docs.galaxyproject.org/en/release_18.05/admin/special_topics/job_metrics.html which provides me with that I need I think! Thanks! Trying it now.

pcm32:
So far this seems to be working at the following levels:

- Job submission.
- Set memory/cpu allowance through destinations.
- Job wrap-up when finished.
- Job recovery after restart (this could fail eventually if the job finished a long time ago in the LSF and the cluster decided to expire the id of the job).
- Job failure (I think).

Sometimes (well, has happened twice), when I restart a job, galaxy doesn't seem to update the running status well (when it goes from PEND to RUN), and I had had to restart the instance to make it become aware the the job is either running or finished. However, I haven't been able to replicate this voluntarily, so I'm not sure. Once it just worked through refreshing the history, the other time I had to restart the instance. I haven't seen it for jobs started out of scratch.

One other thing that I would like to add is something to reduce the chattiness with the LSF cluster in terms of asking for updates. Maybe something like what condor does through the user_log file. I also think I might be using stdout and user_log wrongly (issue being that apparently LSF only produces one of the two. Or should I be doing something like `executable > user_log` within `build_submit_description`? Any directions opinion welcomed.

lib/galaxy/jobs/runners/lsf.py
```diff
@@ -0,0 +1,240 @@
...
+            container = self._find_container(job_wrapper)
+            if container:
+                # HTCondor needs the image as 'docker_image'
+                query_params.update({'docker_image': container.container_id})
+
+        galaxy_slots = query_params.get('cores', None)
+        if galaxy_slots:
+            galaxy_slots_statement = 'GALAXY_SLOTS="%s"; export GALAXY_SLOTS_CONFIGURED="1"' % galaxy_slots
+        else:
+            galaxy_slots_statement = 'GALAXY_SLOTS="1"'
```
jmchilton, line 71: Does https://github.com/galaxyproject/galaxy/commit/4cdc3eb93e1438b8d796b3dec1dc912c357806fc not work for this? I'd rather populate the cores at job runtime in on the actual cluster - I think it is more robust that way.

jmchilton: I see, this is in the Condor runner also. I'd try to get away without it if you can.

lib/galaxy/jobs/runners/lsf.py
```diff
@@ -0,0 +1,240 @@
...
+        # get configured job destination
+        job_destination = job_wrapper.job_destination
+
+        # wrapper.get_id_tag() instead of job_id for compatibility with TaskWrappers.
+        galaxy_id_tag = job_wrapper.get_id_tag()
+
+        # get destination params
+        query_params = submission_params(prefix="", **job_destination.params)
+        container = None
+        universe = query_params.get('universe', None)
```
jmchilton: This is very Condor specific and should be dropped I think.

jmchilton:
> I was thinking often if the [qsb]sub, [qsb]acct, [qsb]stat, [qsb]del based runners can be unified into one runner.

Yeah - so the thing is we have that and I feel like it would radically simplify what is here. It is the CLI runner (https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/jobs/runners/cli.py) - all you need to do is add an LSF plugin here (https://github.com/galaxyproject/galaxy/tree/dev/lib/galaxy/jobs/runners/util/cli/job). There are slurm, torque, and slurm PBS examples there already. It would eliminate a lot of the boilerplate here and let you focus on the LSF specific details. The resulting plugin would be usable with Pulsar managers as well and the Galaxy CLI runner has some bonus goodies like supporting SSH'ing into a host to run the commands.

I think the CLI runner supports batch checking job statuses which would dramatically reduce the chattiness like you are hoping also I believe. If that is still not enough - it we could program some backoff logic in there - that would then be shared across all CLI runner plugins in a non-LSF specific manner.

jmchilton:
I can take a stab at turning what is here into a CLI plugin - but I don't have a test bed but let me know.

lib/galaxy/jobs/runners/lsf.py
pcm32: will drop the universe, I'm using query_params to build the execution file for LSF with its options.

pcm32:
@jmchilton thanks for the review! I would be very happy to test whatever you generate at the CLI plugin side.

pcm32:
Cli plugin looks quite accessible actually... I guess I just looked at the initial example someone proposed to me.

jmchilton:
> I guess I just looked at the initial example someone proposed to me.

Yeah - if the CLI stuff wasn't there this would have been a fine approach. I've opened a PR that adds a note at the top of condor.py (https://github.com/galaxyproject/galaxy/pull/6783) saying that it is a fine approach in production but it shouldn't be emulated for new development. Hopefully this would have prevented this confusion.

jmchilton:
> Cli plugin looks quite accessible actually

Does that mean you want to take a crack at it and let me know how it goes?

pcm32:
I'm bored on the train, I'll give it a go ;-). I'll create another PR for that.

pcm32:
So, besides the weird (and unacceptable) restarting behaviour, this is sort of working. But will leave it a stand-by while I test the other setup (cli-plugin).

jmchilton:
I'm closing this in lieu of the now merged #6788 that I think is a better approach - thanks so much for the effort on both PRs. If there is still a desire to have this one merged despite that other being merged, please re-open and we can discuss. Thanks again!

pcm32:
I agree that is better to only use the other one, so all fine with closing this one. Thanks for the reviews and advice, I'll try to get hold of you John in the beginning next time I want to write another runner :-). I see in the near future at least something for Kubernetes that can be used through pulsar.

PR #6754, opened 2018-09-21
Add backrefs to mapped objects pointing at Dataset

natefoo:
These are useful for something I'm writing to do some dataset ownership/age analysis.


dannon:
These are always one-to-one right now, right?  We may want to add a unique constraint elsewhere since it looks like these generate an array by default.  Unless there are actual performance concerns though, it may not be worth it.

natefoo:
Should be 1-to-1, yeah.

PR #6755, opened 2018-09-21
[18.09] Catch any failures encountered while pausing/stopping jobs whose inputs are terminal and non-ok

natefoo:
So that it doesn't interrupt processing other jobs. That was discovered because attempting to delete purged job outputs on failure was raising `ObjectNotFound` since dataset object store assignment is done later now (as of #6550). So also, catch that `ObjectNotFound`.

There's another case (hence WIP): while checking jobs at startup, an attempt is made to write the per-job registry file. This accesses the job wrapper's `working_directory` method, which raises an exception if the job's `object_store_id` is not set. This was causing one of our handlers to cycle (there's no exception handling above the startup job check so it crashes the entire application) after upgrading to 18.09. As per [the comment in __write_registry_file_if_absent](https://github.com/galaxyproject/galaxy/blob/7481d97a33ba7fb2a003ae2ad51414c04ae826b0/lib/galaxy/jobs/handler.py#L113) we should just remove this anyway, but maybe there would be additional problems elsewhere?

@jmchilton was there a specific reason you weren't using the dataset's delete method? The only thing I can think is that you only want to delete the local file and not the dataset in the object store. But wouldn't you want to delete it in the object store anyway?

### DIFF
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1851,9 +1851,10 @@ class JobWrapper(HasResourceParameters):
                 # to the object store from jobs - be sure that file is cleaned up. This
                 # is a bit of hack - our object store abstractions would be stronger
                 # and more consistent if tools weren't writing there directly.
-                target = dataset.file_name
-                if os.path.exists(target):
-                    os.remove(target)
+                try:
+                    dataset.full_delete()
+                except ObjectNotFound:
+                    pass
 
     def __link_file_check(self):
         """ outputs_to_working_directory breaks library uploads where data is
--- a/lib/galaxy/jobs/handler.py
+++ b/lib/galaxy/jobs/handler.py
@@ -415,11 +415,17 @@ class JobHandlerQueue(Monitors):
             pause_message = ", ".join(jobs_to_pause[job_id])
             pause_message = "%s. To resume this job fix the input dataset(s)." % pause_message
             job, job_wrapper = self.job_pair_for_id(job_id)
-            job_wrapper.pause(job=job, message=pause_message)
+            try:
+                job_wrapper.pause(job=job, message=pause_message)
+            except Exception:
+                log.exception("(%s) Caught exception while attempting to pause job.", job_id)
         for job_id in sorted(jobs_to_fail):
             fail_message = ", ".join(jobs_to_fail[job_id])
             job, job_wrapper = self.job_pair_for_id(job_id)
-            job_wrapper.fail(fail_message)
+            try:
+                job_wrapper.fail(fail_message)
+            except Exception:
+                log.exception("(%s) Caught exception while attempting to fail job.", job_id)
         jobs_to_ignore.update(jobs_to_pause)
         jobs_to_ignore.update(jobs_to_fail)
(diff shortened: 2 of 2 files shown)


lib/galaxy/jobs/__init__.py
```diff
@@ -1851,9 +1851,10 @@ def __update_output(self, job, hda, clean_only=False):
                 # to the object store from jobs - be sure that file is cleaned up. This
                 # is a bit of hack - our object store abstractions would be stronger
                 # and more consistent if tools weren't writing there directly.
-                target = dataset.file_name
-                if os.path.exists(target):
-                    os.remove(target)
+                try:
```
jmchilton, line 1854: xref https://github.com/galaxyproject/galaxy/commit/009e2cbf5e9592fd8af782b2cde36d8d1687c003

jmchilton:
> As per the comment in __write_registry_file_if_absent we should just remove this anyway, but maybe there would be additional problems elsewhere?

That should only affect upgrading from an early 17.XX Galaxy to a 18.09+ Galaxy - I feel like we should just drop that whole method if it is a problem and see what happens.

lib/galaxy/jobs/handler.py
```diff
@@ -415,11 +415,17 @@ def __filter_jobs_with_invalid_input_states(self, jobs):
             pause_message = ", ".join(jobs_to_pause[job_id])
             pause_message = "%s. To resume this job fix the input dataset(s)." % pause_message
             job, job_wrapper = self.job_pair_for_id(job_id)
-            job_wrapper.pause(job=job, message=pause_message)
+            try:
+                job_wrapper.pause(job=job, message=pause_message)
+            except Exception:
```
jmchilton, line 420: So should this be ObjectNotFound ideally?

jmchilton:
I'll PR this change to drop the datatypes file.

lib/galaxy/jobs/handler.py
natefoo: ObjectNotFound should probably be caught at a lower level. I just wanted to make sure that *any* exception in these method don't kill the monitor loop, which effectively kills the handler until you correct whatever the problem is, possibly in the database, by hand.

lib/galaxy/jobs/__init__.py
natefoo: Not sure I'm seeing your reasoning there for why you're not using `.delete()`?

jmchilton: That was for my own tracking purposes, I'm sure what you have right now is better.

lib/galaxy/jobs/handler.py
jmchilton: Well I'll defer to you. I don't have opinion on this. If you take this out of WIP I'll merge.

natefoo:
> As per the comment in __write_registry_file_if_absent we should just remove this anyway, but maybe there would be additional problems elsewhere?

What I meant by this is there may be other places in the startup job check that attempt to access `JobWrapper.working_directory` before `Job.object_store_id` is set. I didn't check, but we can give this a shot and see if we run in to problems.

PR #5013, opened 2017-11-15
Improved Collection and Workflow State with Applications

jmchilton:
## Overview

This is a fairly expansive reworking on the concept of state for both workflows and dataset collections and many new features and fixes these state changes allow. These features include displaying more information with less noise to users of collections and making both collections and workflows simultaneously more correct and more efficient.

(Sorry for the unwieldy PR in terms of both scope and code lines - I've broken many smaller PRs off this branch but the core changes each involve modifying the same database migration repeatedly and so are inter-dependent despite being seemingly disjoint.)

## Features

The main features are as follows and described in detail below.

- Display collection state in the GUI.
- Optimize dataset collection APIs and UX interactions.
- Hide outputs of collection mapping immediately.
- Allow empty collections and empty collection mapping.
- Explicitly track workflow invocation outputs - provide via API.
- Explicitly track workflow invocation step outputs - provide via API.
- Add concept of workflow invocation step state - allow re-scheduling same step multiple times.
- Many tests covering workflow and collection applications.

###  Display collection state in the GUI.

To do this I implement a model abstraction called ``ImplicitCollectionJobs``. These aggregate all the jobs contributing to an implicitly created ``HistoryDatasetCollectionAssociation``. This PR includes model enhancements that allow this and API endpoints for fetching a single job summary or set of job summaries for a history.

With the addition of ``ImplicitCollectionJobs ``, there are now three types of collections - those generated from a single tool (job_source_type = "Job"), those generated from many jobs via mapping (job_source_type = "ImplicitCollectionJobs"), and those just created via the GUI. The following screenshots demonstrate state representations for each for each of these combinations:

We don't have the connections to properly produce this data efficiently for existing collections - so legacy collections (collections created prior to 18.01) will have the improved display of collection counts and such but will not have proper gray, yellow, green, red states. New collections however will.

Fixes #4820.
Fixes #4735.
Fixes #4245.
Fixes #1244.
Progress toward #1810.

### Optimize dataset collection APIs and UX interactions.

I've eliminate ``populated`` from the default dataset collection information in the API. For large and/or nested collections this could cause hundreds of thousands of database queries and joins to compute. I should have never been included - my bad. This is a breaking change but I think it is important. If anyone complains I'd be happy to add a Galaxy config option or something to restore it - but it would be better to fetch it from the detailed variants of those APIs.

[...]

PR #6757, opened 2018-09-21
Give all model objects with an id a __repr__ that includes the id

natefoo:
Useful when developing, but it also helps when an error message includes the string representation of an object.

Before:
```
<galaxy.model.Job object at 0x7f7ee5b71f10>
```

After:
```
<galaxy.model.Job(None)>
<galaxy.model.Job(1)>
```

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -78,6 +78,11 @@ JOB_METRIC_SCALE = 7
 AUTO_PROPAGATED_TAGS = ["name"]
 
 
+class RepresentById(object):
+    def __repr__(self):
+        return '<galaxy.model.%s(%s)>' % (self.__class__.__name__, self.id)
+
+
 class NoConverterException(Exception):
     def __init__(self, value):
         self.value = value
@@ -244,7 +249,7 @@ class JobLike(object):
         return "%s[%s,tool_id=%s]" % (self.__class__.__name__, extra, self.tool_id)
 
 
-class User(Dictifiable):
+class User(Dictifiable, RepresentById):
     use_pbkdf2 = True
     """
     Data for a Galaxy user or admin and relations to their
@@ -517,23 +522,23 @@ class BaseJobMetric(object):
         self.metric_value = metric_value
 
 
-class JobMetricText(BaseJobMetric):
+class JobMetricText(BaseJobMetric, RepresentById):
     pass
 
 
-class JobMetricNumeric(BaseJobMetric):
+class JobMetricNumeric(BaseJobMetric, RepresentById):
     pass
 
 
-class TaskMetricText(BaseJobMetric):
+class TaskMetricText(BaseJobMetric, RepresentById):
     pass
 
 
-class TaskMetricNumeric(BaseJobMetric):
+class TaskMetricNumeric(BaseJobMetric, RepresentById):
     pass
 
 
-class Job(JobLike, UsesCreateAndUpdateTime, Dictifiable):
+class Job(JobLike, UsesCreateAndUpdateTime, Dictifiable, RepresentById):
     dict_collection_visible_keys = ['id', 'state', 'exit_code', 'update_time', 'create_time']
     dict_element_visible_keys = ['id', 'state', 'exit_code', 'update_time', 'create_time']
 
@@ -933,7 +938,7 @@ class Job(JobLike, UsesCreateAndUpdateTime, Dictifiable):
         return config_value
 
 
-class Task(JobLike):
+class Task(JobLike, RepresentById):
     """
     A task represents a single component of a job.
     """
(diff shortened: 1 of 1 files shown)


lib/galaxy/model/__init__.py
```diff
@@ -4544,12 +4550,12 @@ def to_dict(self, view='collection', value_mapper=None):
         return rval
 
 
-class WorkflowInvocationStepJobAssociation(Dictifiable):
+class WorkflowInvocationStepJobAssociation(Dictifiable, RepresentById):
```
nsoranzo: I just found out this class can be removed, it was introduced in commit d5f98f49045bca351e982f17cac9386308089eeb , but its use was removed in commit e82012bde154c456f24743f3b61c0ed9ffc7eca2 , both part of PR #5013.

lib/galaxy/model/__init__.py
```diff
@@ -4544,12 +4550,12 @@ def to_dict(self, view='collection', value_mapper=None):
...
     dict_collection_visible_keys = ('id', 'job_id', 'workflow_invocation_step_id')
     dict_element_visible_keys = ('id', 'job_id', 'workflow_invocation_step_id')
 
 
-class WorkflowRequest(Dictifiable):
+class WorkflowRequest(Dictifiable, RepresentById):
```
nsoranzo: Also this class can be removed, introduced in commit d1cd4ab4d28c72f365852b75e2f2026c6b76ebbd and never used.

lib/galaxy/model/__init__.py
```diff
@@ -4587,45 +4593,45 @@ def __init__(self, workflow_step=None, name=None, value=None):
...
 
-class WorkflowRequestToInputDatasetCollectionAssociation(Dictifiable):
+class WorkflowRequestToInputDatasetCollectionAssociation(Dictifiable, RepresentById):
     """ Workflow step input dataset collection parameters.
     """
     dict_collection_visible_keys = ['id', 'workflow_invocation_id', 'workflow_step_id', 'dataset_collection_id', 'name']
 
 
-class WorkflowRequestInputStepParmeter(Dictifiable):
+class WorkflowRequestInputStepParmeter(Dictifiable, RepresentById):
```
nsoranzo: Uh, this is misspelled (s/Parmeter/Parameter/) both in this file and in lib/galaxy/model/mapping.py

lib/galaxy/model/__init__.py
```diff
@@ -5089,18 +5095,18 @@ def __init__(self, state=None, path=None, info=None, pid=None, socket=None, para
...
         self.parent_id = parent_id
         self.name = name
 
     def __str__(self):
-        return "Tag(id=%s, type=%i, parent_id=%s, name=%s)" % (self.id, self.type, self.parent_id, self.name)
+        return "Tag(id=%s, type=%i, parent_id=%s, name=%s)" % (self.id, self.type or -1, self.parent_id, self.name)
 
 
-class ItemTagAssociation(Dictifiable):
+class ItemTagAssociation(Dictifiable, RepresentById):
```
nsoranzo: That's a base class as `BaseJobMetric` and `ItemRatingAssociation`, so I think this can be reverted.

lib/galaxy/model/__init__.py
```diff
@@ -5176,7 +5182,7 @@ def __init__(self, id=None, user=None, tool_id=None, tag_id=None, user_tname=Non
         self.user_value = None
 
 
-class WorkRequestTagAssociation(ItemTagAssociation):
+class WorkRequestTagAssociation(ItemTagAssociation, RepresentById):
```
nsoranzo: Another class that can be removed, also introduced in commit d1cd4ab and never used.

lib/galaxy/model/__init__.py
```diff
@@ -78,6 +78,11 @@
 AUTO_PROPAGATED_TAGS = ["name", "group"]
 
 
+class RepresentById(object):
+    def __repr__(self):
+        return '<galaxy.model.%s(%s)>' % (self.__class__.__name__, self.id)
```
jmchilton: This will fire a DB query after a flush for every log statement that includes a model object. Consider using cached_id in this file instead (https://github.com/galaxyproject/galaxy/pull/6524/files#diff-5805bf58afdc31ee4a9c4e788301b11dR163).

jmchilton:
I'm -0 on this, it feels a bit too heavy and log messages should be more carefully crafted. Job has a ``log_str`` method you can use instead that produces a nicer, more useful log line IMO.

natefoo:
They should be, but in cases where they're not, it really helps as an admin to know what object is causing the problems you're experiencing.

nsoranzo:
This PR needs to be rebased.

martenson:
@jmchilton I assume nothing changed for the worse on your -0 with the latest commit: https://github.com/galaxyproject/galaxy/pull/6757/commits/414ce0a0eb18ad224d31136ad886079843814b66 therefore I am merging this.

PR #6758, opened 2018-09-22
Fixes for the upload row template

guerler:
This PR adds indentions to the upload row template and aligns the style of the dataset name text field. It also fixes the cursor style handling for icons in the upload row. Additionally the PR contains a minor fix for the button title/icon margin in the upload modal.

after


martenson:
Thanks for improving this.

PR #6759, opened 2018-09-22
[WIP] Re-use test_driver across session

mvdbeek:
That should speed up tests ...

Let's see what I'm breaking here.


mvdbeek:
(the first 20 minutes we're uninstalling and installing dependencies, that's not optimal ...)

mvdbeek:
It doesn't really change much, probably not worth it (for now).

PR #6765, opened 2018-09-25
[18.09] Don't attempt len(None), ensure stdout/stderr are strings in Job(Like).set_streams().

natefoo:
Pulsar jobs that finish ok but with no stdout are failing with the following exception on usegalaxy.org:

```pytb
  File "/cvmfs/main.galaxyproject.org/galaxy/lib/galaxy/jobs/runners/pulsar.py", line 514, in finish_job
    remote_metadata_directory=remote_metadata_directory,
  File "/cvmfs/main.galaxyproject.org/galaxy/lib/galaxy/jobs/__init__.py", line 1445, in finish
    job.set_streams(job.stdout, job.stderr)
  File "/cvmfs/main.galaxyproject.org/galaxy/lib/galaxy/model/__init__.py", line 227, in set_streams
    if (len(stdout) > galaxy.util.DATABASE_MAX_STRING_SIZE):
TypeError: object of type 'NoneType' has no len()
```

Possibly due to #6565 although I'm not sure how exactly.

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -222,8 +222,8 @@ class JobLike(object):
         return self.text_metrics + self.numeric_metrics
 
     def set_streams(self, stdout, stderr):
-        stdout = galaxy.util.unicodify(stdout)
-        stderr = galaxy.util.unicodify(stderr)
+        stdout = galaxy.util.unicodify(stdout) or u''
+        stderr = galaxy.util.unicodify(stderr) or u''
         if (len(stdout) > galaxy.util.DATABASE_MAX_STRING_SIZE):
             stdout = galaxy.util.shrink_string_by_size(stdout, galaxy.util.DATABASE_MAX_STRING_SIZE, join_by="\n..\n", left_larger=True, beginning_on_size_error=True)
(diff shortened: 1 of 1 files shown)


lib/galaxy/model/__init__.py
```diff
@@ -222,8 +222,8 @@ def metrics(self):
         return self.text_metrics + self.numeric_metrics
 
     def set_streams(self, stdout, stderr):
-        stdout = galaxy.util.unicodify(stdout)
-        stderr = galaxy.util.unicodify(stderr)
+        stdout = galaxy.util.unicodify(stdout) or u''
```
dannon, line 225: Might read cleaner to set default parameters on the function, any reason not to do that?

(Does unicodify return `None` for empty strings, instead of an empty unicode string?  That's the only reason I could think of)

nsoranzo: `unicodify()` returns `None` for `None`, not for `''`

dannon: @nsoranzo Got it, yeah.  I just realized my suggestion probably didn't make sense, but that's what I was getting at -- instead of having the logic in consumers like this it might be good to standardize behavior at a single point.  That's for future refactoring though, I guess, no worries about it here.

natefoo:
@martenson Thanks!

PR #6770, opened 2018-09-26
Display input and output tool labels in the workflow editor

### LINKED ISSUE #6764: Display output labels in workflow editor
from @mblue9 in https://github.com/galaxyproject/tools-iuc/pull/2103#issuecomment-424235825
screen shot 2018-09-25 at 3 10 24 pm

> I received some user feedback that in the editor view the featurecounts outputs are a bit confusing, see below (e.g. it's not very obvious to users what `output_short`, `output_medium` and `output_full` are).

This should be straightforward, we can display the label (maybe on hover, to not cram too much text in there ?)

mvdbeek:
I think this makes it a lot easier to choose the correct variable.

Closes #6764

### DIFF
--- a/client/galaxy/scripts/mvc/workflow/workflow-forms.js
+++ b/client/galaxy/scripts/mvc/workflow/workflow-forms.js
@@ -191,10 +191,23 @@ function _visit(head, head_list, output_id, options) {
     }
 }
 
+function _makeRenameHelp(name_labels) {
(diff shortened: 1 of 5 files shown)


dannon:
@mvdbeek Thanks, this is a lot easier to read!

mblue9:
Brilliant! Thanks a lot @mvdbeek !

scholtalbers:
This makes renaming outputs and connecting the right stuff so much easier for those tools where the wrapper author likes short variables!! 🎉

Issue #6773, opened 2018-09-26
Workflow-in-workflow: workflow renaming does not show in editing UI


alexlenail:
When I add subworkflows to a workflow, e.g. in the workflow editor UI for workflow A, I select workflow B from the side panel, it appears in the workflow UI with its old name (prior to renaming). 

Repro: 
- Create workflow B. (add some stuff)
- Rename workflow B to C
- Create workflow A
- Edit workflow A
- Sidebar -> workflows -> C
--> Workflow C shows up in the editing UI under the name "B" 

(cc @mvdbeek)

PR #6774, opened 2018-09-26
[18.05] Update name of Workflow when updating StoredWorkflow name

mvdbeek:
Fixes https://github.com/galaxyproject/galaxy/issues/6773.

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/workflows.py
+++ b/lib/galaxy/webapps/galaxy/api/workflows.py
@@ -481,8 +481,15 @@ class WorkflowsAPIController(BaseAPIController, UsesStoredWorkflowMixin, UsesAnn
         workflow_dict = payload.get('workflow') or payload
         if workflow_dict:
             new_workflow_name = workflow_dict.get('name') or workflow_dict.get('name')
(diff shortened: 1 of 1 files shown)


dannon:
Interesting.  Do you know the reason for why we have `name` for both `Workflow` and `StoredWorkflow`?  (same with `user`, I guess -- is this always redundant, and should we fix the glitch in the model?)

mvdbeek:
I think this is actually nice, this way keep track of metadata changes, so a user can go back and see what a previous version looked like.

dannon:
@mvdbeek Storing those with the workflow revisions is totally fine, but then we should dump it in `StoredWorkflow`, right?

(well, `name` at least -- I'm not sure `user` can ever change without a new `StoredWorkflow`, can it?)

mvdbeek:
Yeah, I suppose StoredWorkflow.name could be a property pointing to latest_workflow.name ? But let's do that in dev

dannon:
@mvdbeek Yep, exactly.  And, I'm fine with fixing it in dev, sure.

jmchilton (approved):

PR #6778, opened 2018-09-26
Release issue scaffold tweaks

dannon:
Minor reminders that would have been nice (for me, at least) to have on the release scaffold.  More the first than the second, but it's something to be mindful of in any event.


nsoranzo (approved):
Thanks for taking the time to formalise this!

PR #6779, opened 2018-09-27
Added singleton application accessor and window.Galaxy monitor

Nerdinacan:
So.... the goal here is to start removing global references to window.Galaxy as we migrate the primary application into a single entry point. I've removed manual window.Galaxy checking from all the python templates, and installed a javascript accessor using Object.defineProperty which will allow us to monitor scripts that are making references to window.Galaxy.

The intent is to continue support old scripts as we migrate toward a more modular application build and provide development tooling through the console so that we can see when random scripts are accessing the global window.Galaxy.

Other changes include addition of a configuration stage to the initialization pipeline. The intent here is to loadConfiguration data from an external source (currently just printed to the page by python) then later initialize the app using a promise. (e.g. loadConfig().then(config => initApp(config));

Most of the magic happens in galaxy.js. An example of endpoint initialization is in analysis.js


dannon:
@Nerdinacan Thanks for the work on this.

For everyone else, this is one of the important steps we identified previously -- having a single Galaxy object that's accessible everywhere, and always accessed the same way, and it will help a lot moving the client rewrite forward.

dannon:
@galaxybot test this

martenson:
@Nerdinacan This is awesome!!! 🤩

client/galaxy/scripts/galaxy.js
```diff
@@ -293,10 +292,75 @@ GalaxyApp.prototype.toString = function toString() {
...
+
+
+
+// Install temporary code to detect when window.Global is being accessed
+// We'll use this to hunt down all the global Galaxy references and replace
+// them with getGalaxyInstance()
+
+Object.defineProperty(window, 'Galaxy', {
+    enumerable: true,
+    get() {
```
jmchilton: Clever - thanks for teaching me something! 😃

Nerdinacan:
Sorry sorry sorry, writing tests now. I know it means I'm a bad kid, but I was looking to see if anybody had objections to the plan before I went all-in. :)

dannon:
@Nerdinacan No worries, it's good step directly in the direction we'd talked about -- it looks to me like we just need to update the tests, totally expected.

client/galaxy/scripts/galaxy.js
Nerdinacan: If you like that, check out ES6 Proxies. :)

dannon:
@galaxybot test this

dannon:
@galaxybot test this

martenson:
@Nerdinacan this has gotten a bit outdated and conflicted, would you mind doing a merge/rebase please?

Nerdinacan:
> @Nerdinacan this has gotten a bit outdated and conflicted, would you mind doing a merge/rebase please?

Absolutely. I just did a merge with a fresh copy of dev, and had to force push the branch. You're right, it's diverged a lot from when I opened the PR. Let me know if you'd rather I just opened a separate one.

Nerdinacan:
Something's seriously screwy here, all these selenium tests are running clean on my machine. Yet most of them fail on the jenkins build.

Nerdinacan:
I suspect something went foul when I force pushed this branch. I only did that to preserve the conversation. I've created a new PR with a fresh copy of the branch organized into cleaner commits so it's easy to review/complain about.

https://github.com/galaxyproject/galaxy/pull/7005

PR #6780, opened 2018-09-27
[18.09] All extracted datasets should be visible

mvdbeek:
HDAs in a HDCA are hidden, but if we extract them we do want to see them.


jmchilton:
It is almost like I just write tests and never actually try out this Galaxy thing. Thanks for this fix.

PR #6783, opened 2018-09-27
Warn job runner plugin writers not to emulate condor.py.

jmchilton:
Better to write a CLI plugin I think (https://github.com/galaxyproject/galaxy/pull/6751#issuecomment-425079201).


mvdbeek (approved):
I did propose this very recently on gitter as well.

jmchilton:
> I did propose this very recently on gitter as well.

Didn't see that, but now I know I am right 😄.

PR #6786, opened 2018-09-27
Add PB and EB to util.size_to_bytes

natefoo:
Also, reorder match check so that the sizes more likely to be encountered are first (super awesome optimization!!!!1).

These aren't too likely to be encountered in Galaxy itself but I wrote a script that used this function and it crashed trying to convert PB. Seemed reasonable to add both PB and EB since they are both supported in `nice_size`.


nsoranzo (approved):
I think the failed API tests are unrelated.

PR #6788, opened 2018-09-27
CLI plugin for LSF

pcm32:
As suggested by @jmchilton at #6751, here is a first attempt at a Cli plugin for LSF (untested yet). One thing I couldn't figure out is how to attach this to job_conf, couldn't find any cli-plugins examples. Do I need to add the new LSF plugin class somewhere? Thanks!

### DIFF
--- a/lib/galaxy/jobs/runners/util/cli/job/lsf.py
+++ b/lib/galaxy/jobs/runners/util/cli/job/lsf.py
@@ -62,7 +62,7 @@ class LSF(BaseJobExec):
         return "bsub <%s | awk '{ print $2}' | sed 's/[<>]//g'" % script_file
 
     def delete(self, job_id):
-        return 'bstop %s' % job_id
+        return 'bkill %s' % job_id
 
     def get_status(self, job_ids=None):
         return "bjobs -a -o \"id stat\" -noheader"  # check this
--- a/lib/galaxy/jobs/runners/util/cli/job/lsf.py
+++ b/lib/galaxy/jobs/runners/util/cli/job/lsf.py
@@ -65,7 +65,7 @@ class LSF(BaseJobExec):
         return 'bstop %s' % job_id
 
     def get_status(self, job_ids=None):
-        return "bjobs -o \"id stat delimiter='^'\" -noheader"  # check this
+        return "bjobs -a -o \"id stat\" -noheader"  # check this
 
     def get_single_status(self, job_id):
         return "bjobs -o stat -noheader " + job_id
@@ -73,8 +73,8 @@ class LSF(BaseJobExec):
     def parse_status(self, status, job_ids):
         # Get status for each job, skipping header.
         rval = {}
-        for line in status.splitlines()[1:]:
-            id, state = line.split('^')
+        for line in status.splitlines():
+            id, state = line.split()
             if id in job_ids:
                 # map job states to Galaxy job states.
                 rval[id] = self._get_job_state(state)
--- a/lib/galaxy/jobs/runners/util/cli/job/lsf.py
+++ b/lib/galaxy/jobs/runners/util/cli/job/lsf.py
@@ -74,10 +74,10 @@ class LSF(BaseJobExec):
         # Get status for each job, skipping header.
         rval = {}
         for line in status.splitlines():
-            id, state = line.split()
-            if id in job_ids:
+            job_id, state = line.split()
+            if job_id in job_ids:
                 # map job states to Galaxy job states.
-                rval[id] = self._get_job_state(state)
+                rval[job_id] = self._get_job_state(state)
         return rval
 
     def parse_single_status(self, status, job_id):
--- a/lib/galaxy/jobs/runners/util/cli/job/lsf.py
+++ b/lib/galaxy/jobs/runners/util/cli/job/lsf.py
@@ -84,6 +84,7 @@ class LSF(BaseJobExec):
     def _get_job_state(self, state):
         # based on:
         # https://www.ibm.com/support/knowledgecenter/en/SSETD4_9.1.3/lsf_admin/job_state_lsf.html
+        # https://www.ibm.com/support/knowledgecenter/en/SSETD4_9.1.2/lsf_command_ref/bjobs.1.html
         try:
             return {
                 'EXIT': job_states.ERROR,
@@ -92,7 +93,10 @@ class LSF(BaseJobExec):
                 'DONE': job_states.OK,
                 'PSUSP': job_states.ERROR,
                 'USUSP': job_states.ERROR,
-                'SSUSP': job_states.ERROR
+                'SSUSP': job_states.ERROR,
+                'UNKWN': job_states.ERROR,
+                'WAIT' : job_states.QUEUED,
+                'ZOMBI' : job_states.ERROR
             }.get(state)
         except KeyError:
             raise KeyError("Failed to map LSF status code [%s] to job state." % state)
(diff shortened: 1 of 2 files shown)


lib/galaxy/jobs/runners/util/cli/job/lsf.py
```diff
@@ -0,0 +1,102 @@
...
+                log.warning('Unrecognized long argument passed to LSF CLI plugin: %s' % k)
+
+        # Generated template.
+        template_scriptargs = ''
+        for k, v in scriptargs.items():
+            template_scriptargs += '#BSUB %s %s\n' % (k, v)
+        return dict(headers=template_scriptargs)
+
+    def submit(self, script_file):
+        return 'bsub <%s' % script_file
```
pcm32: Not sure if this is going to play nicely with Popen upstream though (I mean the internal redirect).

lib/galaxy/jobs/runners/util/cli/job/lsf.py
```diff
@@ -0,0 +1,102 @@
...
+        for line in status.splitlines()[1:]:
+            id, state = line.split('^')
+            if id in job_ids:
+                # map job states to Galaxy job states.
+                rval[id] = self._get_job_state(state)
+        return rval
+
+    def parse_single_status(self, status, job_id):
+        if "is not found" in status:
+            raise Exception("Job %s not found in LSF" % job_id)
```
pcm32: Not sure if I should be handling the 'not found' case here though through an exception. Please advise.

lib/galaxy/jobs/runners/util/cli/job/lsf.py
```diff
@@ -0,0 +1,102 @@
...
+        # https://www.ibm.com/support/knowledgecenter/en/SSETD4_9.1.3/lsf_admin/job_state_lsf.html
+        try:
+            return {
+                'EXIT': job_states.ERROR,
+                'RUN': job_states.RUNNING,
+                'PEND': job_states.QUEUED,
+                'DONE': job_states.OK,
+                'PSUSP': job_states.ERROR,
+                'USUSP': job_states.ERROR,
+                'SSUSP': job_states.ERROR
```
pcm32: There are a number of suspended states that don't match nicely to QUEUED in Galaxy, they shouldn't be reached normally, but I guess the most correct interpretation for them would be ERROR. Please advise.

pcm32:
I would be very interested in trying this with pulsar to send stuff to our LSF cluster from outside of that cluster from our Kubernetes setup.

pcm32:
Ok, found docs for how to use it.

lib/galaxy/jobs/runners/util/cli/job/lsf.py
jmchilton: https://www.ibm.com/support/knowledgecenter/en/SSETD4_9.1.2/lsf_command_ref/bsub.1.html 

Looks like if it is just the last argument to bsub you don't need to pipe it into stdin?

jmchilton: If that is still a problem though - we can make submit return something more structured - we probably shouldn't be so relaxed with argument handling and such here anyway. Returning something with a list of command and arguments as well as options for stdin/stdout/etc... would be probably be good.

lib/galaxy/jobs/runners/util/cli/job/lsf.py
jmchilton: The other plugins assume the job has completed and is no longer being tracked by the DRM:

```
        # no state found, job has exited
        return job_states.OK
```

Stick with that? If you are confident the LSF is going to reserve and track that job_id for a long time and so this indicates something definitely went wrong you could return job_states.ERROR instead - this would be better than throwing an exception I think (this is used in a loop that doesn't catch and handle that exception https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/jobs/runners/cli.py#L154).

lib/galaxy/jobs/runners/util/cli/job/lsf.py
jmchilton: ¯\\\_(ツ)\_/¯ ... ... maybe start with error and if at some point you find there are jobs being suspended that would resume you could add a job destination parameter to handle the suspending jobs differently - like resetting them to queued?

lib/galaxy/jobs/runners/util/cli/job/lsf.py
pcm32: No worries, this is all working like this, so I wouldn't give it any importance. On the other LSF runner implementation I tried not using stdin and it never worked. But this is working as it is, so no worries here.

lib/galaxy/jobs/runners/util/cli/job/lsf.py
pcm32: I guess you are right, after many years of not using LSF, what I noticed now is that it only keeps the id IFF the job fails. I think I have added -a as well, which will show all jobs as well (at the expense maybe of showing too many jobs). Will change to the behaviour that you suggest.

pcm32:
What I haven't been able to figure out @jmchilton is how the Cli plugin sets the env var that is expected at https://github.com/galaxyproject/galaxy/blob/ed6f5d45a9ecbf0cf8a85341e0213b82a9132a46/lib/galaxy/jobs/runners/util/job_script/CLUSTER_SLOTS_STATEMENT.sh (so that GALAXY_SLOTS gets set)
I don't see the slurm job cli plugin setting anything for instance so that `SLURM_CPUS_ON_NODE` is set. Thanks!

jmchilton:
@pcm32 Slurm sets the SLURM_CPUS_ON_NODE environment variable - that isn't something set in Galaxy. If LSF isn't setting https://github.com/galaxyproject/galaxy/blob/ed6f5d45a9ecbf0cf8a85341e0213b82a9132a46/lib/galaxy/jobs/runners/util/job_script/CLUSTER_SLOTS_STATEMENT.sh#L24 can you print a dump of what it is setting and find the right variable?

pcm32:
Great, yes, LSF is setting it (LSB_DJOB_NUMPROC), thanks for the hint (I though I was supposed to inject something in the environment).

jmchilton (approved):

jmchilton:
Happy to merge this as is and we can refine it if you hit bugs - is it ready to go?

pcm32:
This is working well and I'm happy to merge. I only find this a bit nasty https://github.com/galaxyproject/galaxy/pull/6788/files#diff-11a6329dd180df8eaa9deaf64b10ce60R62 but it works. Up to you.

lib/galaxy/jobs/runners/util/cli/job/lsf.py
```diff
@@ -0,0 +1,110 @@
...
+        return "bjobs -a -o \"id stat\" -noheader"  # check this
+
+    def get_single_status(self, job_id):
+        return "bjobs -o stat -noheader " + job_id
+
+    def parse_status(self, status, job_ids):
+        # Get status for each job, skipping header.
+        rval = {}
+        for line in status.splitlines():
+            id, state = line.split()
```
mvdbeek: Some of the other runners can output something more structured, like xml or json, that might be an alternative if it exists ?  If not can you change `id` to `job_id` to not mask the builtin ?

pcm32: As you might see from lines 68 and 71 above I'm defining the output structure already... I would say that for one and two tokens that need to be generated in each case (only status for single case; job_id and status for the multiple case), this should be more than fine and I think is unnecessary to get any json/xml handling logic for such simple cases. Will definitely change id to job_id as suggested, thanks for spotting that.

PR #6791, opened 2018-09-27
[18.09] Flex-related style fixes

### LINKED ISSUE #6777: cannot scroll on a published history
found on Main

when you expand multiple datasets the scrollbar is never offered

example: https://usegalaxy.org/u/zhangxi/h/circrna

dannon:
Fixes published object titlebars that look like this:
image

And all pages that won't scroll (page editor, published history display, etc.) xref, fixes #6777 

This also includes yarn integrity hashes for dependencies that we'll update as we update node deps moving forward.


dannon:
Ok, this is ready to go now.  It fixes #6791 and the titlebars that were noticed to be wonky, as well.

bgruening:
@galaxybot test this

PR #6792, opened 2018-09-27
Flex style fixes

dannon:
Fixes published object titlebars that look like this:
image

And annotated history contents that look like this:
image


dannon:
Well that was weird.  I was editing #6791 and this showed up.

PR #6793, opened 2018-09-27
[18.09] allow making all datasets within history public, when sharing

martenson:
addresses part of https://github.com/galaxyproject/galaxy/issues/5894

This is written to be minimalistic and targets 18.09 as it has been identified as major bug by @nekrut.

notes: 
- the UI is horrendous, but not much worse than it was before
- this is limited to histories at the moment
- this is a one-off operation so notable behavior is that if you create a new dataset in the shared history, it still won't be public when there are default permissions set on history
- this does not handle the reverse (when unpublishing it is not making your datasets restricted, since we do not know what were their permissions before. Should it just make it private?)
- naively works with collections, needs checking
- needs tests

cc @bgruening @jmchilton @mvdbeek

### DIFF
--- a/client/galaxy/scripts/components/Sharing.vue
+++ b/client/galaxy/scripts/components/Sharing.vue
@@ -1,7 +1,7 @@
 <template>
     <div v-if="ready">
         <h2>Share or Publish {{model_class}} `{{item.title}}`</h2>
-        <b-alert variant="danger" :show="err_msg">
+        <b-alert :show="show_danger" variant="danger" dismissible>
             {{ err_msg }}
         </b-alert>
         </br>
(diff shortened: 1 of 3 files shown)


dannon:
To solve two problems at once (UI and unpublishing without making assumptions), instead of the checkboxes next to each button we could have a single toggle below that's always available, and an indicator there showing whether or not the contents are accessible or not.  Anyway, future work, this is a step forward for the release and it works for me.  Thanks.

dannon:
(test failure is unrelated)

bgruening:
Awesome! Thanks a lot. This was bugging our users a lot lately.

PR #6796, opened 2018-09-28
[16.01] Drop unused peek method

dannon:
The `TODO: unused?` is correct.  No need to fix it to use hashed identifiers, just dropping it in 16.01 and merging forward.


martenson (approved):

jmchilton:
Thanks!

Issue #6182, opened 2018-05-21
When processing collections in batch mode, $input.element_identifier seems to revert to $input.name instead


lparsons:
Example is [RSeQC `geneBody_coverage.xml`](https://github.com/galaxyproject/tools-iuc/blob/master/tools/rseqc/geneBody_coverage.xml). When running in ["merge" mode](https://github.com/galaxyproject/tools-iuc/blob/2bd06c2b43c295fb4cf172c4f156fed5475855a4/tools/rseqc/geneBody_coverage.xml#L17), `$input.element_identifier` is set to the element id in the collection. However, when run in ["batch" mode](https://github.com/galaxyproject/tools-iuc/blob/2bd06c2b43c295fb4cf172c4f156fed5475855a4/tools/rseqc/geneBody_coverage.xml#L31) `$input.element_identifier` get's set to the filename (which is not desired).

jmchilton:
I don't think this is true in general because we have tests I think (https://github.com/galaxyproject/galaxy/blob/dev/test/api/test_tools.py#L912), but we don't have tests for mapping over a parameter inside a conditional - the conditional might be what is breaking this. We also have tests for element identifiers inside conditionals but they are for multiple=true parameters not single parameters like this tool.

lparsons:
Feel free to edit the issue to make the description more precise. I only know this one case atm.

jmchilton:
I have a test case that puts a single data identifier inside a repeat and a conditional and the identifier is still used for me - at least in 18.05. What version of Galaxy did you observe this in?

lparsons:
I saw it on 17.09 I believe. I'm in the process of updating to 18.01 (or possibly straight to 18.05), but didn't want to push it out before the long weekend, so I'll do it next week and let you know if things looks resolved or not. Thanks for looking into it, I appreciate it.

lparsons:
@jmchilton I'm running 18.05 and this is still happening.

lparsons:
Any updates on this? It makes the RNASeQC GeneBodyCoverage output in MultiQC much less useful. Has anyone come up with a workaround? I'd be happy to patch the tools to fix this.

mvdbeek:
I think the tool here might be doing something unexpected. Can you try adding the conditional prefix to input ?  https://github.com/galaxyproject/tools-iuc/blob/2bd06c2b43c295fb4cf172c4f156fed5475855a4/tools/rseqc/geneBody_coverage.xml#L32

[...]

PR #6798, opened 2018-09-28
[18.09] Fix element identifier for non multiple true mapover in conditionals

mvdbeek:
Fixes https://github.com/galaxyproject/galaxy/issues/6182


dannon:
@mvdbeek Legitimate test failures?  (looks likely)

mvdbeek:
yep, I'll have a look

PR #6802, opened 2018-09-29
[18.05] Cast rule.length to int when parsing rules

### LINKED ISSUE #6800: Bug: Apply Rules - New Column - Substring function broken
Attempting to use the function to create a column by keeping or stripping a substring suffix or prefix always results in the following exception:
```
Error executing tool: Rule of type [add_column_substr] does not contain correct value type for key [length].
```

mvdbeek:
Fixes #6800

To rebuild the client I did `make client-production-maps`, but I didn't add the changes in static/style/blue/base.css, those changes seem unrelated ?


jmchilton (approved):

dannon:
@mvdbeek build looked fine to me; merged forward.  The good news is that past 18.05, we won't ever be committing those again, so it'll all just work!

PR #6805, opened 2018-09-30
[18.09] Allow starting Galaxy under gunicorn when using galaxy.ini

mvdbeek:
and starting Galaxy with `APP_WEBSERVER=gunicorn sh run.sh`.
Additional gunicorn cmdline arguments can be passed using
`GUNICORN_CMD_ARGS`. In a first pass this will allow
running planemo {serve, test} under python 3, as Paste
has some problems under python 3:

```
galaxy.web.stack INFO 2018-09-29 22:16:00,216 Galaxy server instance 'main' is running
  File "./scripts/paster.py", line 27, in <module>
    serve.run()
  File "/tmp/tmpv5tow6c8/galaxy-dev/lib/galaxy/util/pastescript/serve.py", line 1055, in run
    invoke(command, command_name, options, args[1:])
  File "/tmp/tmpv5tow6c8/galaxy-dev/lib/galaxy/util/pastescript/serve.py", line 1061, in invoke
    exit_code = runner.run(args)
  File "/tmp/tmpv5tow6c8/galaxy-dev/lib/galaxy/util/pastescript/serve.py", line 226, in run
    result = self.command()
  File "/tmp/tmpv5tow6c8/galaxy-dev/lib/galaxy/util/pastescript/serve.py", line 676, in command
    serve()
  File "/tmp/tmpv5tow6c8/galaxy-dev/lib/galaxy/util/pastescript/serve.py", line 653, in serve
    server(app)
  File "/tmp/tmpv5tow6c8/galaxy-dev/lib/galaxy/util/pastescript/loadwsgi.py", line 232, in server_wrapper
    **context.local_conf)
  File "/tmp/tmpv5tow6c8/galaxy-dev/lib/galaxy/util/pastescript/loadwsgi.py", line 90, in fix_call
    val = callable(*args, **kw)
  File "/home/travis/.planemo/gx_venv_3_release_18.09/lib/python3.6/site-packages/paste/httpserver.py", line 1353, in server_runner
    for name, value in kwargs.items():
RuntimeError: dictionary changed size during iteration
```

That said gunicorn is a nice pure-python alternative to paste going
forward.

It's a bit of a stretch to call this a bug and include it in 18.09, 
but this should go long ways in verifying that tools are python 3 compatible.

### DIFF
--- a/lib/galaxy/dependencies/pipfiles/default/Pipfile
+++ b/lib/galaxy/dependencies/pipfiles/default/Pipfile
@@ -41,6 +41,7 @@ bleach = "*"
 "bz2file" = {version = "*", markers = "python_version < '3.3'"}
 ipaddress = {version = "*", markers = "python_version < '3.3'"}
 boltons = "*"
+gunicorn = "*"
 Paste = "*"
 docutils = "*"
 Routes = "*"
--- a/lib/galaxy/dependencies/pipfiles/default/Pipfile
+++ b/lib/galaxy/dependencies/pipfiles/default/Pipfile
@@ -9,6 +9,7 @@ verify_ssl = true
 name = "pypi"
 
 [dev-packages]
+gunicorn = "*"
 lxml = "!=4.2.2"
 mock = "*"
 NoseHTML = "*"
@@ -41,7 +42,6 @@ bleach = "*"
 "bz2file" = {version = "*", markers = "python_version < '3.3'"}
 ipaddress = {version = "*", markers = "python_version < '3.3'"}
 boltons = "*"
-gunicorn = "*"
 Paste = "*"
 docutils = "*"
 Routes = "*"
--- a/lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
+++ b/lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
@@ -10,6 +10,7 @@ chardet==3.0.4
 commonmark==0.5.4
 docutils==0.14
 funcsigs==1.0.2; python_version < '3.3'
+gunicorn==19.9.0
 idna==2.7
 imagesize==1.1.0; python_version != '3.3.*'
 jinja2==2.10
--- a/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
+++ b/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
@@ -54,7 +54,6 @@ functools32==3.2.3.post2; python_version == '2.7'
 future==0.16.0
 futures==3.2.0; python_version == '2.6' or python_version == '2.7'
 galaxy-sequence-utils==1.1.3
-gunicorn==19.9.0
 h5py==2.8.0
 html5lib==1.0.1
 idna==2.7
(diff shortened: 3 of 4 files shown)


mvdbeek:
xref https://github.com/galaxyproject/planemo/pull/874

jmchilton:
Planemo can readily target dev... but I guess the goal is to Python 3 test all of the big repos in their main travis setup so you want this in master quicker?

jmchilton (approved):

mvdbeek:
Yes, ideally we'd do one round of cleanup and then switch to testing against python3.

mvdbeek:
On the other hand we do only need this for `serve`, it seems that `test` works fine in the default setup.

jxtx:
This looks fine, but is it at odds with our goal of doing uwsgi as the default web server for Galaxy? We have discussed using some uwsgi specific features (e.g. proxy stuff) and basically requiring uwsgi for simplicity.

mvdbeek:
Yeah, it is, but I was thinking about this as an alternative to paste, which would allow an easy switch now in planemo. I definitely wouldn't argue for setting up a standalone / production Galaxy instance using gunicorn, but for planemo and live-debugging (which is certainly possible with uwsgi, just haven't figured out how) this is nice.

nsoranzo (approved):
This seems worth if the goal is to get rid of Paste from the dependencies.

mvdbeek:
> This seems worth if the goal is to get rid of Paste from the dependencies.

I think that's definitely something we can achieve in the next cycle, there's just a few places where we still use Paste, and there seem to be replacements readily available. But to be honest we don't need gunicorn for this, so I'd consider that independent.

mvdbeek:
Keep in the mind that gunicorn doesn't fully read the [server] sections from paste-style ini files, so it wouldn't be a drop-in replacement for setups that still use Paste.

mvdbeek:
Alright, planemo serve is passing tests against this branch: https://travis-ci.org/galaxyproject/planemo/jobs/435572552#L519

martenson:
Will this PR make everybody download gunicorn? Can we make it dev-requirement if so?

mvdbeek:
I've moved it to the dev-requirements.

lib/galaxy/dependencies/pipfiles/default/Pipfile
```diff
@@ -41,7 +41,6 @@ bleach = "*"
 "bz2file" = {version = "*", markers = "python_version < '3.3'"}
 ipaddress = {version = "*", markers = "python_version < '3.3'"}
 boltons = "*"
-gunicorn = "*"
```
nsoranzo: @mvdbeek This should actually be moved in the `[dev-packages]` section above, not deleted.

mvdbeek: And I was surprised that I didn't find a Pipfile in develop ...

nsoranzo: The `develop` subdir is gone in 18.09, commit a2d99e30c727e7050c1fd6218603b7b2941c8c4d .

dannon: @mvdbeek Same thing confused me briefly when I swapped nodeenv around to a core requirement!

PR #6776, opened 2018-09-26
Allow API import/export of format 2 workflows.

jmchilton:
This builds on a set of Galaxy PRs #6746, #6807, #6811 and a series of [gxformat2](https://github.com/jmchilton/gxformat2) releases.

In addition to bringing in the latest changes to gxformat2 to allow these syntax changes (described below) and updating hundreds of lines of test workflows - this PR finally enables such workflows to work directly with the Galaxy API (switched on ``enable_beta_workflow_format``). When enabled, workflows imported from dictionaries will now be checked and if they look like a format2 workflow they will be pre-converted to a native format before import - uploading YAML wrapped in JSON to avoid dictionary handling changes is allowed also. Likewise, there is a couple new styles of download that attempt to extract format 2 workflows from the native representation (using new functionality added to recent versions of gxformat2). These can be downloaded as JSON or as YAML string content wrapped up in JSON in order to preserve the pretty YAML formatting and dictionary ordering implemented in gxformat2.

If the JSON wrapped YAML-in-strings seems odd - consider also an important future direction of this work is likely to store the original supplied YAML alongside the native representation of the workflow and then allow the workflow editor to operate a series of deltas to both in parallel. This can be done with a round trip aware YAML parsing/writer such as [ruamel.yaml](https://yaml.readthedocs.io/en/latest/) so that user comments and formatting as well as extraneous data in the YAML are preserved.

This PR continues to refine the workflow syntax toward a more concise and CWL compatible syntax. The following two code blocks are a before and after example demonstrating the syntax changes.

Before this series of PRs:

```yaml
class: GalaxyWorkflow
inputs:
  - id: input_fastqs
    type: collection
  - id: reference
outputs:
  - id: pileup_output
    source: pileup#out_file1
steps:
  - label: map_over_mapper
    tool_id: mapper
    state:
      input1:
        $link: input_fastqs
      reference:
        $link: reference
  - label: pileup
    tool_id: pileup
    state:
      input1:
        $link: map_over_mapper#out_file1
      reference:
        $link: reference
```

My recommended best practice after this PR is merged:

```yaml
class: GalaxyWorkflow
inputs:
  input_fastqs: collection
  reference: data
outputs:
  pileup_output:
    outputSource: pileup/out_file1
steps:
  map_over_mapper:
    tool_id: mapper
    in:
      input1: input_fastqs
      reference: reference
  pileup:
    tool_id: pileup
    in:
      input1: map_over_mapper/out_file1
      reference: reference
```

[...]

PR #6807, opened 2018-09-30
Support for Enhancements to gxformat2.

jmchilton:
Builds on https://github.com/galaxyproject/galaxy/pull/6746, to provide even better support for Galaxy Format 2 workflows.

A few changes:

- Import the gxformat2 module as an external dependency instead of using code from within the test/ directory. This isn't strictly needed yet but will be in subsequent PRs (#6776) that bring Format 2 support into Galaxy core directly. I had initially done this by moving the code into ``lib/`` but I decided it would be better to maintain one canonical source and skip the complexity of sync'ing stuff the way we do with galaxy-lib.
- Revs the version of gxformat2 which contains some fixes, but most importantly adds a CWL-style input connection handling with an ``in`` keyword. A minor syntax difference but it should go a long way toward making gxformat2 workflows - proper CWL workflows with extensions for Galaxy style tools.

``in`` requires skipping the special special link syntax with tool state. The embedded state syntax using ``$link`` still works fine though but it is very particular to tool steps.

```diff
   - tool_id: cat1
     label: first_cat
-    state:
-      input1:
-        $link: outer_input
+    in:
+      input1: outer_input
```

For non-tool steps though the ``in`` syntax looks a lot like the ``connect`` syntax which already existed. Just using ``/`` to separate step ID and output name.

```diff
     label: nested_workflow
-    connect:
-      inner_input: first_cat#out_file1
+    in:
+      inner_input: first_cat/out_file1
```

All three syntaxes are still supported, I suppose at very least we should say ``in`` is preferred to ``connect`` going forward and drop support for ``connect`` as a keyword.

- Finally, Galaxy workflows will work if connections are references in conditionals and at the top level but don't have empty tool state values - but within repeats if there wasn't an empty repeat block in the tool state and an input connection was specified that would require a repeat instance - that wouldn't work because the repeat instance that is required wouldn't be visited by the tool parameter visit code. This tweaks workflow import to create empty repeat instances needed for input connections that are specified without empty tool state blocks. There is a long discussion in the doc strings on https://github.com/galaxyproject/galaxy/commit/c685b6bada3c08d121fb7fa9c6b0532381c04bdf and a test case. The tl;dr is that this makes it feasible to define steps with just connections (``in`` statements) and not require ``state`` statements for tools.


mvdbeek:
This is conflicting now that I merged https://github.com/galaxyproject/galaxy/pull/6746, but it does look good to me, I think the syntax is a lot cleaner, and moving gxformat2 to an external dep seems like the right thing to do. Maybe we could have a more descriptive package name though, like `galaxy-workflow-lib`, `lib-galaxy-workflow` or something like that ?

nsoranzo:
@jmchilton This is breaking selenium tests on dev: https://jenkins.galaxyproject.org/job/docker-selenium/3811/console

```
$ ack workflows_format_2
test/manual/workflows_scaling.py
27:from base.workflows_format_2.converter import python_to_workflow  # noqa: I100

test/selenium_tests/framework.py
23:from base.workflows_format_2 import (  # noqa: I100
```

jmchilton:
I see - thanks @nsoranzo  - I'll get a fix right in.

PR #6810, opened 2018-10-01
[18.09] do not rule out numbers that are accessed with shift modifier

### LINKED ISSUE #5370: Unable to enter value in numerical input with keyboards that use 'shift' for numbers
Seem to be related to #507 

Impossible to enter a numerical value with a french keyboard mapping (using the numeric keys above azerty line) on numerical inputs. I spot it on a local Galaxy instance and can reproduce it on main usegalaxy ( Tool text Manipulation/Select first lines from a dataset  , "Number of lines" box)

 It is nevertheless Ok with numpad, or Ok if I switch my keyboard mapping to US/English .

martenson:
closes #5370
I think there are more problems with this code:
 - it assumes too much about the keyboard used and puts
unnecessary restrictions on the user
 - the implementation should be refactored to use KeyboardEvent.key
instead of the deprecated charCode

edit: the problems are listed as 'future work', I do not think they stop this PR


mvdbeek:
Merci from my 🇫🇷 coworkers!

PR #6811, opened 2018-10-01
Support for Enhancements to gxformat2 (part II)

jmchilton:
Further extends #6807 and #6746 with more Format 2 workflow tweaks.

This:

 - Switches most format 2 workflows to use test_data separation and ``in`` keyword for connections introduced in #6807.
 - Uses new id map based workflow input and output definitions.
 - Revs gxformat2 to allow these enhancements and to handle subworkflows, post job actions, and pause steps.

The id map based approach replaces arrays for inputs and outputs. Like CWL both are allowed, but also like CWL the newer id map based approach is probably preferred.

For an input definition, the following diff describes the change:

```diff
 inputs:
-  - id: input1
-    type: data_collection_input
+  input1:
+    type: collection
     collection_type: list
```

For simpler data inputs without extra modifiers, a simpler map can be used:

```diff
 inputs:
-  - id: input1
+  input1: data
```

For an output definition, the following diff describes the change:

```diff
 workflow_outputs:
-  - id: outer_output
-    source: second_cat#out_file1
+  outer_output:
+    outputSource: second_cat/out_file1
```

While not particularly more concise, it does bring the syntax inline with CWL - which uses ``outputSource`` on workflow steps instead of ``source``.

### DIFF
--- a/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
+++ b/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
@@ -55,7 +55,7 @@ functools32==3.2.3.post2; python_version == '2.7'
 future==0.16.0
 futures==3.2.0; python_version == '2.6' or python_version == '2.7'
 galaxy-sequence-utils==1.1.3
-gxformat2==0.3.0
+gxformat2==0.6.0
 h5py==2.8.0
 html5lib==1.0.1
 idna==2.7
--- a/test/api/test_workflows.py
+++ b/test/api/test_workflows.py
@@ -77,8 +77,8 @@ class: GalaxyWorkflow
 inputs:
   - id: outer_input
 outputs:
-  - id: outer_output
-    source: second_cat#out_file1
+  outer_output:
+    outputSource: second_cat/out_file1
 steps:
   - tool_id: cat1
     label: first_cat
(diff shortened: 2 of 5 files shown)


lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
```diff
@@ -55,6 +55,7 @@ functools32==3.2.3.post2; python_version == '2.7'
 future==0.16.0
 futures==3.2.0; python_version == '2.6' or python_version == '2.7'
 galaxy-sequence-utils==1.1.3
+gxformat2==0.6.0
```
mvdbeek, line 58: Something unexpected is going on here, after merging https://github.com/galaxyproject/galaxy/pull/6807 there should be a merge conflict here, but there isn't ?!

jmchilton: This PR had that PRs exact commits - so they don't conflict.  I changed this line twice - once in commits in that PR and again in a commit only in this PR.

mvdbeek: Alright, I was assuming the context in the github interface would at least show a bump from 0.3.1 to 0.6.0 and not show the commits that are already integrated -- although I guess this is a nice "feature" for stale PRs.

jmchilton: I'll rebase - I'm sure it will generate a cleaner diff against dev then.

mvdbeek (approved):

PR #6814, opened 2018-10-04
[18.05] Fix apply rules bug in filter first/last N rows.

shiltemann:
When using the `Apply rule to collection` tool, and using the `Filter first or last N rows` filter, I would get the following error: 

```
Error executing tool: Rule of type [add_filter_count] does not 
contain correct value type for key [count].
```

this fixed it for me


hexylena (approved):

mvdbeek (approved):

hexylena (approved):
Thanks @shiltemann, looks great!

jmchilton:
There are unit tests and specification tests of the rules and the processing on the frontend - but they don't actually use a text form for input, they use JSON definitions - so this is why a few of these type errors have popped up. Thanks for the fix @shiltemann!

PR #6818, opened 2018-10-05
Add more detail about container tags

pvanheus:
Explain that singularity container tags are now supported and add some detail about how to use them.

### DIFF
--- a/lib/galaxy/tools/xsd/galaxy.xsd
+++ b/lib/galaxy/tools/xsd/galaxy.xsd
@@ -233,8 +233,8 @@ For more information, see https://planemo.readthedocs.io/en/latest/writing_advan
 This is a container tag set for the ``requirement`` and ``container`` tags
 described in greater detail below. ``requirement``s describe software packages
 and other individual computing requirements required to execute a tool, while
-``container``s describe Docker containers that should be able to serve as
-complete descriptions of the runtime of a tool.
+``container``s describe Docker or Singularity containers that should be able to
+serve as complete descriptions of the runtime of a tool.
 
 ]]></xs:documentation>
     </xs:annotation>
@@ -312,9 +312,21 @@ resolver.
     <xs:annotation>
       <xs:documentation xml:lang="en"><![CDATA[
 This tag set is contained within the 'requirements' tag set. Galaxy can be
-configured to run tools within [Docker](https://www.docker.com/) containers -
-this tag allows the tool to suggest possible valid Docker containers for this
-tool.
+configured to run tools within [Docker](https://www.docker.com/) or [Singularity](https://www.sylabs.io/singularity/)
+containers - this tag allows the tool to suggest possible valid containers for this tool. The contents of the tag should
+be a container image identifier appropriate to the particular container runtime being used, e.g. 
+``quay.io/biocontainers/fastqc:0.11.2--1`` for Docker or ``docker://quay.io/biocontainers/fastqc:0.11.2--1``
+(or alternatively ``/opt/containers/fastqc.simg`` if your Galaxy installation will be loading the image from a filesystem path)
+for Singularity. The requirements tag can contain multiple ``container`` tags describing suitable container options, in
+which case the first container that is found by the Galaxy container resolver at runtime will be used.
+
+Example:
+
+```xml
+<requirements>
+  <container type="docker">quay.io/biocontainers/fastqc:0.11.2--1</container>
+<requirements>
+```
 
 Read more about configuring Galaxy to run Docker jobs
 [here](https://galaxyproject.org/admin/tools/docker/).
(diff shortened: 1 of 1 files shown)


lib/galaxy/tools/xsd/galaxy.xsd
```diff
@@ -312,9 +312,21 @@ resolver.
     <xs:annotation>
       <xs:documentation xml:lang="en"><![CDATA[
 This tag set is contained within the 'requirements' tag set. Galaxy can be
-configured to run tools within [Docker](https://www.docker.com/) containers -
-this tag allows the tool to suggest possible valid Docker containers for this
-tool.
+configured to run tools within [Docker](https://www.docker.com/) or [Singularity](https://www.sylabs.io/singularity/)
+containers - this tag allows the tool to suggest possible valid containers for this tool. The contents of the tag should
+be a container image identifier appropriate to the particular container runtime being used, e.g.
```
nsoranzo: "appropriate to the particular" -> "appropriate for the particular" sounds better, I think.

lib/galaxy/tools/xsd/galaxy.xsd
```diff
@@ -312,9 +312,21 @@ resolver.
...
 This tag set is contained within the 'requirements' tag set. Galaxy can be
-configured to run tools within [Docker](https://www.docker.com/) containers -
-this tag allows the tool to suggest possible valid Docker containers for this
-tool.
+configured to run tools within [Docker](https://www.docker.com/) or [Singularity](https://www.sylabs.io/singularity/)
+containers - this tag allows the tool to suggest possible valid containers for this tool. The contents of the tag should
+be a container image identifier appropriate to the particular container runtime being used, e.g. 
+``quay.io/biocontainers/fastqc:0.11.2--1`` for Docker or ``docker://quay.io/biocontainers/fastqc:0.11.2--1``
+(or alternatively ``/opt/containers/fastqc.simg`` if your Galaxy installation will be loading the image from a filesystem path)
+for Singularity. The requirements tag can contain multiple ``container`` tags describing suitable container options, in
```
nsoranzo: "requirements tag" -> "\`\`requirements\`\` tag"

nsoranzo (approved):

lib/galaxy/tools/xsd/galaxy.xsd
```diff
@@ -312,9 +312,21 @@ resolver.
...
 This tag set is contained within the 'requirements' tag set. Galaxy can be
-configured to run tools within [Docker](https://www.docker.com/) containers -
-this tag allows the tool to suggest possible valid Docker containers for this
-tool.
+configured to run tools within [Docker](https://www.docker.com/) or [Singularity](https://www.sylabs.io/singularity/)
+containers - this tag allows the tool to suggest possible valid containers for this tool. The contents of the tag should
+be a container image identifier appropriate for the particular container runtime being used, e.g. 
+``quay.io/biocontainers/fastqc:0.11.2--1`` for Docker or ``docker://quay.io/biocontainers/fastqc:0.11.2--1``
+(or alternatively ``/opt/containers/fastqc.simg`` if your Galaxy installation will be loading the image from a filesystem path)
+for Singularity. The ``requirements`` tag can contain multiple ``container`` tags describing suitable container options, in
```
jmchilton, line 320: Does this work today? If yes, I don't think we should document it - I think instead we should turn IDs that aren't paths and that don't contain a ``://`` into a ``docker://<id>`` when resolving Docker type containers for singularity. That way these tools work the same between Docker and Singularity and the tool definition doesn't need to depend on singularity?

lib/galaxy/tools/xsd/galaxy.xsd
```diff
@@ -324,7 +336,7 @@ Read more about configuring Galaxy to run Docker jobs
       <xs:extension base="xs:string">
         <xs:attribute name="type" type="ContainerType" use="required">
           <xs:annotation>
-            <xs:documentation xml:lang="en">This value describes the type of container that the tool may be executed in and currently must be ``docker``.</xs:documentation>
+            <xs:documentation xml:lang="en">This value describes the type of container that the tool may be executed in and currently may be ``docker`` or ``singularity``.</xs:documentation>
```
jmchilton, line 339: +1 on this change though

lib/galaxy/tools/xsd/galaxy.xsd
pvanheus: @jmchilton yes this works today and in fact we are using it for a project where we want to pre-populate a server with tools that are not under an open license. I.e. in our use case, images are placed in a particular path and found using explicit `<container>` tags.

In terms of not depending on singularity - the `type="XXX"` attribute of the `<container>` tag is tested against the list of enabled container types [here](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/deps/container_resolvers/explicit.py#L23) thus there does seem to be a dependency between tool definition and container type?

PR #6820, opened 2018-10-05
Add docs about running jobs in containers

pvanheus:
Explain how jobs can be run in containers (something currently documented in the `job_conf.xml.sample_advanced` file)


martenson:
Thanks @pvanheus !

PR #6827, opened 2018-10-07
Clarify code used in output <filter> tags

pvanheus:
Add extra detail to the explanation of how the Python code to ``<filter>`` tags works in outputs.


lib/galaxy/tools/xsd/galaxy.xsd
```diff
@@ -3901,7 +3901,12 @@ tool ultimately creates. If the code, when executed, returns ``True``,
 the output dataset is retained. In these code blocks the tool parameters appear
 as Python variables and are thus referred to without the $ used for the Cheetah
 template (used in the ``<command>`` tag). Variables that are part of
-conditionals are accessed using a hash named after the conditional.
+conditionals are accessed using a dictionary named after the conditional. Boolean
+parameters appear as booleans, not the value of their ``truevalue`` and
+``falsevalue`` attributes. In the example below, ``options["selection_mode"`` would
```
bgruening: `]` are missing.

PR #1306, opened 2015-12-11
New Workflow Modules - Input Parameters and Subworkflows

jmchilton:
## Overview

This PR contains two new workflow modules to bring in concepts required to implement CWL workflows in Galaxy - nested workflows and input parameters. I developed these in the CWL fork of Galaxy but I am confident they will be very useful for vanilla Galaxy workflows.
## Process

If people generally agree to this PR - I'd like to merge e4c9ddb and 22f06b7 which have two database migrations into one before it is merged. I'm keeping them separated for now in case people object to one of these workflow modules but not the other.
## Nested Workflows
- Add a new workflow module describing subworkflows.
- Update workflow, workflow step, and workflow invocation models to track subworkflow connections and execution.
- Extend workflow outputs with concepts of labels (and UUIDs while I'm there) to match workflow inputs. This allow us to have something to label outputs with in the workflow editor and to reference in the format 2 workflow description language.
- Extend workflow editor UI to allow labelling workflow outputs (and enforce that these are unique across a workflow).
- Extend workflow invocation and progress tracking to allow invoking a subworkflow as part of another workflow invocation.
- Extend workflow import and export code to allow a nested representation of workflows.
- Update format 2 workflow description to allow testing nested workflows.

```
./run_tests.sh -api test/api/test_workflows.py:WorkflowsApiTestCase.test_run_subworkflow_simple
./run_tests.sh -api test/api/test_workflows_from_yaml.py:WorkflowsFromYamlApiTestCase.test_subworkflow_simple
./run_tests.sh -api test/api/test_workflows_from_yaml.py:WorkflowsFromYamlApiTestCase.test_outputs
nosetests test/unit/test_galaxy_mapping.py
nosetests test/unit/workflows/test_workflow_progress.py
```

Following screenshot demonstrates two ways to utilize workflows within another workflow:
- the link icon preserves the subworkflow structure and treats it as a single unit
- the copy icon copies the other workflow into the current workflow node for node

subworkflow_types
## Workflow Input Parameters (Experimental)
- Implement a input parameter module that mirrors data and collection input modules but has a type that can currently be one of text, integer, float, color, and boolean.
- Allow connections between these and tool step inputs.
- Extend model to support this.
- Add new input types for format 2 workflow definitions for various types that all map to this kind of step. Typed inputs such as this match well with CWL workflow inputs.

[...]

PR #6829, opened 2018-10-08
[WIP] Add workflow step parameters to workflow editor

mvdbeek:
This should allow specifying text, integer, float, boolean and color Parameters that can be re-used across workflow nodes. This build on https://github.com/galaxyproject/galaxy/pull/1306/commits/45962f32344a00162f5d04b94096c8a570f2adf1

### DIFF
--- a/client/galaxy/scripts/mvc/workflow/workflow-view-terminals.js
+++ b/client/galaxy/scripts/mvc/workflow/workflow-view-terminals.js
@@ -161,6 +161,17 @@ var InputTerminalView = BaseInputTerminalView.extend({
     }
 });
 
+var InputParameterTerminalView = BaseInputTerminalView.extend({
+    terminalMappingClass: Terminals.TerminalMapping,
+    terminalMappingViewClass: InputTerminalMappingView,
+    terminalForInput: function(input) {
+        return new Terminals.InputParameterTerminal({
+            element: this.el,
+            input: input
+        });
+    }
+});
+
 var InputCollectionTerminalView = BaseInputTerminalView.extend({
     terminalMappingClass: Terminals.TerminalMapping,
     terminalMappingViewClass: InputTerminalMappingView,
@@ -278,9 +289,25 @@ var OutputCollectionTerminalView = BaseOutputTerminalView.extend({
     }
 });
 
+var OutputParameterTerminalView = BaseOutputTerminalView.extend({
+    terminalMappingClass: Terminals.TerminalMapping,
+    terminalMappingViewClass: TerminalMappingView,
+    terminalForOutput: function(output) {
+        var collection_type = output.collection_type;
+        var collection_type_source = output.collection_type_source;
+        var terminal = new Terminals.OutputCollectionTerminal({
+            element: this.el,
+            type: output.type,
+        });
+        return terminal;
+    }
+});
+
 export default {
     InputTerminalView: InputTerminalView,
+    InputParameterTerminalView: InputParameterTerminalView,
     OutputTerminalView: OutputTerminalView,
+    OutputParameterTerminalView: OutputParameterTerminalView,
     InputCollectionTerminalView: InputCollectionTerminalView,
     OutputCollectionTerminalView: OutputCollectionTerminalView
 };
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -4179,14 +4179,10 @@ class WorkflowStepConnection(object):
         self.input_step_id = None
         self.input_name = None
 
-    def set_non_data_connection(self):
-        self.output_name = WorkflowStepConnection.NON_DATA_CONNECTION
-        self.input_name = WorkflowStepConnection.NON_DATA_CONNECTION
-
     @property
     def non_data_connection(self):
-        return (self.output_name == WorkflowStepConnection.NON_DATA_CONNECTION and
-                self.input_name == WorkflowStepConnection.NON_DATA_CONNECTION)
+        return (self.output_name == self.input_name == WorkflowStepConnection.NON_DATA_CONNECTION or
+                self.output_step and self.output_step.type == 'parameter_input')
 
     def copy(self):
         # TODO: handle subworkflow ids...
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -4181,8 +4181,7 @@ class WorkflowStepConnection(object):
 
     @property
     def non_data_connection(self):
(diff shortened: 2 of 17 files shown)


mvdbeek:
A good fraction of the UI is there and seems to work OK-ish, but the actual non-data parameter replacement only works for new-style workflow API submissions with `inputs`. The workflow run form still uses a different format for this. I'll see if it is sufficient to just move `input_parameter` to inputs, but IIRC there was a longer-term plan to unify UI and API. That isn't happening anytime soon, right ?

lib/galaxy/workflow/modules.py
```diff
@@ -1244,27 +1256,19 @@ def load_module_sections(trans):
...
+            },
+            {
+                "name": "parameter_input",
+                "title": "Parameter Input",
+                "description": "Simple inputs used for workflow logic"
+            },
         ],
     }
 
-    if trans.app.config.enable_beta_workflow_modules:
```
jmchilton, line 1250: My two cents is that I'd keep the option and just promote parameter_input to an un-beta parameter.

jmchilton:
I'll play with this and see if I can figure out an opinion on the API unification and fix the tests.

jxtx:
Oooooh. Screenshots?

mvdbeek:
Nothing fancy yet:
screen shot 2018-10-11 at 17 50 32
screen shot 2018-10-11 at 17 51 42

We should separate the data and non-data inputs in the workflow node (a bar, and maybe different color noodle), and the workflow inputs should move to the top where we also render the old workflow replacement parameters. And I'm sure there are plenty of bugs in there, just trying to figure this out.

jmchilton:
> The workflow run form still uses a different format for this. I'll see if it is sufficient to just move input_parameter to inputs, but IIRC there was a longer-term plan to unify UI and API. That isn't happening anytime soon, right ?

:( indeed it does. I'll start with the input parameters and see if I can separate them out, after that if I can separate out the data parameters so we are tracking them with database integrity using proper tracking with  ``WorkflowRequestToInputDatasetAssociation`` / ``WorkflowRequestToInputDatasetCollectionAssociation`` the way standardized API calls with generate.

jmchilton:
Working on ripping parameter inputs from the raw param map produced by the tool form and I encountered another problem - it looks like this branch breaks the parameter_inputs in some way. In particular the following test case that works in dev:

```
./run_tests.sh -api test/api/test_workflows.py::WorkflowsApiTestCase::test_run_with_text_connection
```

No longer works, the key line in the logs I think is seeing that there is an unmatched input connection.

```
galaxy.workflow.modules WARNING 2018-10-12 10:31:42,351 Failed to use input connections for inputs [set([u'seed_source|seed'])]
```

I'm going to see if I can figure out what caused this - none of the backend changes are obvious problems at first glance to me.

lib/galaxy/model/__init__.py
```diff
@@ -4179,14 +4179,10 @@ def __init__(self):
...
-    def set_non_data_connection(self):
-        self.output_name = WorkflowStepConnection.NON_DATA_CONNECTION
-        self.input_name = WorkflowStepConnection.NON_DATA_CONNECTION
-
     @property
     def non_data_connection(self):
-        return (self.output_name == WorkflowStepConnection.NON_DATA_CONNECTION and
-                self.input_name == WorkflowStepConnection.NON_DATA_CONNECTION)
+        return (self.output_name == self.input_name == WorkflowStepConnection.NON_DATA_CONNECTION or
+                self.output_step and self.output_step.type == 'parameter_input')
```
jmchilton: Reverting this diff fixes the test case. I guess there aren't docs for this - but I don't like this change. The parameter_input is transferring data to the step - it isn't an implicit connection so this should return False IMO. I see why you made the change here - for when building the editor dict:

```diff
                 # FIXME: this removes connection without displaying a message currently!
+                input_connections = [conn for conn in input_connections if conn.input_name in data_input_names]
-                input_connections = [conn for conn in input_connections if conn.input_name in data_input_names or conn.non_data_connection]
```

Maybe we should bite the bullet and just not do that filtering at all - push all the connections to the editor? This way the editor can tell the user if there are invalid connections found and we can eliminate that FIXME?

mvdbeek: Sounds good to me!

client/galaxy/scripts/mvc/workflow/workflow-view-terminals.js
```diff
@@ -253,10 +264,10 @@ var OutputTerminalView = BaseOutputTerminalView.extend({
     terminalMappingClass: Terminals.TerminalMapping,
     terminalMappingViewClass: TerminalMappingView,
     terminalForOutput: function(output) {
-        var type = output.extensions;
-        var terminal = new Terminals.OutputTerminal({
+        var type = output.type;
+        var terminal = new Terminals.OutputParameterTerminal({
             element: this.el,
-            datatypes: type
```
jmchilton: Should these changes be reverted - given the class below?

mvdbeek: Oh yeah, sure, should have cleaned this up before.

jmchilton:
@mvdbeek any chance I can take this over this week and rebase #6776 underneath it? I'd like to add some Selenium tests and see if I can implement default values, but I can wait on you and limit the scope at how it stands now if you'd like.

mvdbeek:
Sure, go ahead, happy to get this rolling!

jmchilton:
@mvdbeek My new change is still creating more issues then solving and the GUI is terrible - but this is an outline of separating RuntimeValue into RuntimeValue and ConnectedValue (https://github.com/jmchilton/galaxy/commit/15e667cdf9b566b7167d53954f8da9d57b54d10f).

I guess my next step is to try to make data and non-data parameters more similar in terms of the tool state and module interface. The data parameters should have ConnectedValue as their tool state value and then we can uniformly have inputs with ConnectedValues appear with connections and eliminate the distinction between parameters and inputs. This will be good from an interface perspective but is also required for CWL where a parameter that can take an integer or a File for instance is totally possible. I'll probably merge https://github.com/galaxyproject/galaxy/pull/6911 into that to provide that consistent interface on the backend, I'll rework the frontend a bit - I think you have the right classes I just want to change how their instantiated a bit I think.

client/galaxy/scripts/mvc/workflow/workflow-view-terminals.js
jmchilton: Going to remove this via rebase on this commit and force push to your branch - hope that is okay.

mvdbeek: sure

PR #6832, opened 2018-10-08
Fix/slurm out of memory warning exit code 0:125

scholtalbers:
We receive these warning messages (OUT_OF_MEMORY but with exit code 0:125) a lot on our cluster and the jobs are to be considered successful 
https://bugs.schedmd.com/show_bug.cgi?id=3820#c62

### DIFF
--- a/lib/galaxy/jobs/runners/slurm.py
+++ b/lib/galaxy/jobs/runners/slurm.py
@@ -134,7 +134,7 @@ class SlurmJobRunner(DRMAAJobRunner):
                         return
                     except Exception:
                         ajs.fail_message = "This job failed due to a cluster node failure, and an attempt to resubmit the job failed."
-                elif slurm_state in ('OUT_OF_MEMORY') and exit_code in ("0:125", None):
+                elif slurm_state in ('OUT_OF_MEMORY') and exit_code == '0:125':
                     log.debug("(%s/%s) SLURM reported job OUT_OF_MEMORY with exit code 0:125. We are assuming success!",
                         ajs.job_wrapper.get_id_tag(), ajs.job_id)
                     drmaa_state = self.drmaa_job_states.DONE
--- a/lib/galaxy/jobs/runners/slurm.py
+++ b/lib/galaxy/jobs/runners/slurm.py
@@ -48,7 +48,7 @@ class SlurmJobRunner(DRMAAJobRunner):
 
     def _complete_terminal_job(self, ajs, drmaa_state, **kwargs):
         def _get_slurm_state_with_sacct(job_id, cluster):
-            cmd = ['sacct', '-n', '-o', 'state%-32']
+            cmd = ['sacct', '-n', '-o', 'state%-32,exit']
             if cluster:
                 cmd.extend(['-M', cluster])
             cmd.extend(['-j', job_id])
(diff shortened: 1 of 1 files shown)


lib/galaxy/jobs/runners/slurm.py
```diff
@@ -60,12 +60,12 @@ def _get_slurm_state_with_sacct(job_id, cluster):
                     log.warning('SLURM accounting storage is not properly configured, unable to run sacct')
                     return
                 raise Exception('`%s` returned %s, stderr: %s' % (' '.join(cmd), p.returncode, stderr))
-            # First line is for 'job_id'
+            # First line is for 'job_id  exit'
```
nsoranzo: This is not a description of the fields in the line, just that the first line contains info for `jod_id`. Revert?

lib/galaxy/jobs/runners/slurm.py
```diff
@@ -48,7 +48,7 @@ class SlurmJobRunner(DRMAAJobRunner):
 
     def _complete_terminal_job(self, ajs, drmaa_state, **kwargs):
         def _get_slurm_state_with_sacct(job_id, cluster):
-            cmd = ['sacct', '-n', '-o', 'state%-32']
+            cmd = ['sacct', '-n', '-o', 'state%-32,exit']
```
nsoranzo: We should probably use the [official field names](https://slurm.schedmd.com/sacct.html), i.e. `State%-32,ExitCode`

lib/galaxy/jobs/runners/slurm.py
```diff
@@ -60,12 +60,12 @@ def _get_slurm_state_with_sacct(job_id, cluster):
                     log.warning('SLURM accounting storage is not properly configured, unable to run sacct')
                     return
                 raise Exception('`%s` returned %s, stderr: %s' % (' '.join(cmd), p.returncode, stderr))
-            # First line is for 'job_id'
+            # First line is for 'job_id  exit'
             # Second line is for 'job_id.batch' (only available after the batch job is complete)
             # Following lines are for the steps 'job_id.0', 'job_id.1', ... (but Galaxy does not use steps)
-            first_line = stdout.splitlines()[0]
+            status, exit_code = stdout.splitlines()[0].split()
```
nsoranzo: s/status/state/ (also below)

lib/galaxy/jobs/runners/slurm.py
```diff
@@ -134,6 +134,10 @@ def _get_slurm_state():
                         return
                     except Exception:
                         ajs.fail_message = "This job failed due to a cluster node failure, and an attempt to resubmit the job failed."
+                elif slurm_state in ('OUT_OF_MEMORY') and exit_code == '0:125':
+                    log.debug("(%s/%s) SLURM reported job OUT_OF_MEMORY with exit code 0:125. We are assuming success!",
+                        ajs.job_wrapper.get_id_tag(), ajs.job_id)
+                    drmaa_state = self.drmaa_job_states.DONE
```
nsoranzo: I'd move this inside the next `elif`

nsoranzo:
@scholtalbers If I read https://bugs.schedmd.com/show_bug.cgi?id=3820#c62 correctly, with SLURM 17.11.3 this will not do the right thing any more, since `OUT_OF_MEMORY` state with ExitCode `0:125` will identify the real OOM-killed jobs.

nsoranzo:
@galaxybot test this

scholtalbers:
we have 17.11.6 and it does not correctly identify oom killed jobs. I don't know the exact details, but I as far as I understood if an event like this is detected on the node (for whatever reason) this will become the job state, even though the job was not affected.

natefoo:
I'm unsure on this it looks like code `0:125` should only be happening if processes were really killed. You're sure nothing's been killed when you're seeing this?

scholtalbers:
Positive. However, if in doubt, we can also make this a configurable option?

natefoo:
Yeah, that'd be ideal I think.

PR #6837, opened 2018-10-08
Don't create jobs for inactive users

natefoo:
This mostly avoids the accumulation of jobs in the `new` state that will never run because the user never activates their account. It'd still be possible on servers that allow anonymous jobs because you could create a job before logging in and then log in before it gets picked up by a handler. But this should take care of nearly all such jobs.

This is the backend only, although the UI is acceptable for history import/export. Uploading yields a 400 in the upload modal with no indication as to the cause of failure and running regular jobs returns a modal and the response json, and the error message is in there but it's easy to miss:

> The server could not complete the request. Please contact the Galaxy Team if this error persists. Error executing tool: Action requires account activation.

Could a UI person take a look at making this display nicer? Also, for uploads, it'd be good if the client outright prevented uploading so that users don't waste a bunch of time uploading a file only to have the job creation fail.

### DIFF
--- a/test/unit/tools/test_actions.py
+++ b/test/unit/tools/test_actions.py
@@ -257,6 +257,10 @@ class MockTrans(object):
 
     user_is_active = property(get_user_is_active, set_user_is_active)
 
+    def check_user_activation(self):
+        if not self.user_is_active:
+            raise UserActivationRequiredException()
+
     def db_dataset_for(self, input_db_key):
         return None
 
--- a/test/unit/tools/test_actions.py
+++ b/test/unit/tools/test_actions.py
@@ -3,6 +3,7 @@ import unittest
 from xml.etree.ElementTree import XML
 
 from galaxy import model
+from galaxy.exceptions import UserActivationRequiredException
 from galaxy.tools.actions import (
     DefaultToolAction,
     determine_output_format,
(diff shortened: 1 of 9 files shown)


nsoranzo:
Unit tests failure are legit: `AttributeError: 'MockTrans' object has no attribute 'user_is_active'`

jmchilton:
https://github.com/natefoo/galaxy/pull/9

natefoo:
Seems like the obviously better solution now... thanks @jmchilton.

test/unit/tools/test_actions.py
```diff
@@ -117,6 +118,14 @@ def test_handler_set(self):
         job, _ = self._simple_execute()
         assert job.handler == TEST_HANDLER_NAME
 
+    def test_inactive_user_job_create_failure(self):
+        self.trans.user_is_active = False
```
nsoranzo, line 122: Isn't a bit misleading that this is something you cannot do in the real `trans`?

No big issue anyway.

natefoo: Yeah and so I was a little wary of doing it. I did find some other places where it's done though.

Is there a better way to do this for unit tests? It all gets a little ugly since there are so many fake objects anyway, which need to be kept at least somewhat in sync with the functionality of the real ones.

natefoo:
@jmchilton you broke my new test! :grin:

nsoranzo (approved):

natefoo:
Thanks @nsoranzo @jmchilton, and @martenson for creating the followup issue.

PR #6826, opened 2018-10-06
Remove `trans` from OIDC-related types

VJalili:
In oidc-related types (i.e., `PSAAssociation`, `PSACode`, `PSANonce`, `PSAPartial`, and `UserAuthnzToken`), `trans` is used only for database access (i.e., `trans.sa_session`). Hence this PR replaces `trans` with `sa_session` in these types.

ping @jmchilton

PR #6838, opened 2018-10-08
Replace trans with session and request in psa Storage

VJalili:
This PR replaces `trans` in the `Strategy` constructor with `request` and `session`. This change makes it possible to instantiate a `Strategy` when `trans` is not available/exposed; e.g., from within a tool wrapper. 

ping @jmchilton

### DIFF
--- a/lib/galaxy/authnz/managers.py
+++ b/lib/galaxy/authnz/managers.py
@@ -228,7 +228,7 @@ class AuthnzManager(object):
             log.exception(msg)
             return False, msg, None
 
-    def get_cloud_access_credentials(self, cloudauthz, request, sa_session, user_id):
+    def get_cloud_access_credentials(self, cloudauthz, sa_session, user_id, request=None):
         """
         This method leverages CloudAuthz (https://github.com/galaxyproject/cloudauthz)
         to request a cloud-based resource provider (e.g., Amazon AWS, Microsoft Azure)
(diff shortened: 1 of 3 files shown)


jmchilton:
Have you figured out how to get what you need from the request to the tool wrapper?

VJalili:
With the changes of this PR and #6826, the [`get_cloud_access_credentials`](https://github.com/galaxyproject/galaxy/blob/1f389ae3e77352920566f09c4f1be1a1e40ff862/lib/galaxy/authnz/managers.py#L231) function can be called without `trans`. Its signature is changed as the following: 

```python
# From:
def get_cloud_access_credentials(self, trans, cloudauthz):

# To:
def get_cloud_access_credentials(self, cloudauthz, request, sa_session, user_id):
```

In tool wrapper, I have `sa_session` and `user_id`, but I may not have `request`. However, my tests show that for cloudauthz purpose, `request` can safely be `None`, though it needs further tests.

jmchilton:
> However, my tests show that for cloudauthz purpose, request can safely be None, though it needs further tests.

This would be ideal for these operations, hopefully that holds up! Awesome.

VJalili:
@galaxybot test this

VJalili:
I ran a number of tests, and it turns out `request` can be safely set to `None` for the purpose of CloudAuthz. Hence, I made it optional as the following: 

```python
def get_cloud_access_credentials(self, cloudauthz, sa_session, user_id, request=None)
```

VJalili:
@jmchilton, thank you!

PR #6839, opened 2018-10-08
[18.09] Unified panel style bottom padding fix

dannon:
This fixes the bug @jennaj noticed in the history panel, the bottom ~25px being masked.

I really wanted to generalize this with something like:
```
div.unified-panel:has(> .unified-panel-footer) {
    .unified-panel-body {
        padding-bottom: $panel_footer_height;
        overflow: auto;
    }
}
```

But I couldn't make it cooperate, if anyone cares to follow up.


jennaj:
Retested at ORG running pre-release 18.09. Scroll displays complete history/dataset content. Thanks!

PR #6841, opened 2018-10-09
 Replace deprecated assertEquals() method 

nsoranzo:
Fix warnings during py34-unit tests like:

```
/home/travis/build/galaxyproject/galaxy/test/unit/workflows/test_extract_summary.py:46: DeprecationWarning: Please use assertEqual instead.
  self.assertEquals(job_dict[hda.job], [('out1', derived_hda_2)])
```

See e.g. https://travis-ci.org/galaxyproject/galaxy/jobs/438829277

Also remove unneeded parentheses in lib/galaxy/web/base/controller.py

### DIFF
--- a/test/api/test_api_batch.py
+++ b/test/api/test_api_batch.py
@@ -29,7 +29,7 @@ class ApiBatchTestCase(api.ApiTestCase):
         response = self._post_batch(batch)
         response = response.json()
         self.assertIsInstance(response, list)
-        self.assertEquals(len(response), 3)
+        self.assertEqual(len(response), 3)
 
     def test_unallowed_route(self):
         batch = [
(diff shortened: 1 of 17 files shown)


test/api/test_api_batch.py
```diff
@@ -29,7 +29,7 @@ def test_simple_array(self):
         response = self._post_batch(batch)
         response = response.json()
         self.assertIsInstance(response, list)
-        self.assertEquals(len(response), 3)
+        self.assertEqual(len(response), 3)
```
mvdbeek, line 32: Even better would be plain assert statements IMO. This would allow us to get rid of unittest. I'm not sure that's a road we've decided on.

mvdbeek (approved):

jmchilton (approved):

PR #6842, opened 2018-10-09
Fix unit test that uses fake trans.user_is_admin

natefoo:
for the change to a property made in #6835.

The test passed but made the same error of using an uncalled function as a boolean that was to be avoided by turning it in to a property (it is a fake attribute in the test).


natefoo:
@galaxybot test this

PR #6844, opened 2018-10-09
Infrastructure/API Support for Composite Uploads/Downloads

jmchilton:
Extends the upload tool to allow attaching arbitrary composite files to datasets on upload. Adds an API endpoint to list a dataset's composite files and directories. An API for downloading such content already exists - but the API consumer would need to know the files exist ahead of time (as is the case say with tool tests that leverage this API). Add a Directory datatype that can represent the decompressed contents of a tar dataset and the needed converter. I'll follow up with similar converters for zip files if this is merged.

### DIFF
--- a/lib/galaxy/datatypes/data.py
+++ b/lib/galaxy/datatypes/data.py
@@ -927,6 +927,10 @@ class Text(Data):
         return dataproviders.line.RegexLineDataProvider(dataset_source, **settings)
 
 
+class Directory(Data):
+    """Class representing a directory of files."""
+
+
 class GenericAsn1(Text):
     """Class for generic ASN.1 text format"""
     edam_data = "data_0849"
--- a/test/api/test_tools.py
+++ b/test/api/test_tools.py
@@ -221,10 +221,10 @@ class ToolsTestCase(api.ApiTestCase):
         test_data = test_data_response.json()
         assert len(test_data) == 3
 
-    def test_upload_composite_as_tar(self):
-        with self.dataset_populator.test_history() as history_id:
-            tar_path = self.test_data_resolver.get_filename("testdir.tar")
-            tar_f = open(tar_path, "rb")
+    @uses_test_history(require_new=False)
+    def test_upload_composite_as_tar(self, history_id):
+        tar_path = self.test_data_resolver.get_filename("testdir.tar")
+        with open(tar_path, "rb") as tar_f:
             payload = self.dataset_populator.upload_payload(history_id, "Test123",
                 extra_inputs={
                     "files_1|file_data": tar_f,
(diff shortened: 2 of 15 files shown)


test/api/test_tools.py
```diff
@@ -221,6 +221,59 @@ def test_test_data_yaml_tools(self):
         test_data = test_data_response.json()
         assert len(test_data) == 3
 
+    def test_upload_composite_as_tar(self):
+        with self.dataset_populator.test_history() as history_id:
+            tar_path = self.test_data_resolver.get_filename("testdir.tar")
+            tar_f = open(tar_path, "rb")
```
mvdbeek: We can use a `with` statement here

test/api/test_tools.py
```diff
@@ -221,6 +221,59 @@ def test_test_data_yaml_tools(self):
         test_data = test_data_response.json()
         assert len(test_data) == 3
 
+    def test_upload_composite_as_tar(self):
+        with self.dataset_populator.test_history() as history_id:
```
mvdbeek: We could also use the history_id decorator here, right ?

test/api/test_tools.py
```diff
@@ -221,6 +221,59 @@ def test_test_data_yaml_tools(self):
...
+                assert path in expected_contents
+                assert extra_file["class"] == expected_contents[path]
+                found_files.add(path)
+
+            assert len(found_files) == 5, found_files
+
+    def test_upload_composite_from_bad_tar(self):
+        with self.dataset_populator.test_history() as history_id:
+            tar_path = self.test_data_resolver.get_filename("unsafe.tar")
+            tar_f = open(tar_path, "rb")
```
mvdbeek: Can use `with` here as well.

tools/data_source/upload.py
```diff
@@ -172,32 +174,65 @@ def add_file(dataset, registry, output_path):
 
 
 def add_composite_file(dataset, output_path, files_path):
+
+    def to_path(path_or_url):
+        is_url = path_or_url.find('://') != -1  # todo fixme
```
mvdbeek, line 187: I suppose you still want to do something about this ?

mvdbeek (commented):
I like it, and I think this has been requested many times before.

lib/galaxy/datatypes/data.py
```diff
@@ -927,6 +927,10 @@ def regex_line_dataprovider(self, dataset, **settings):
         return dataproviders.line.RegexLineDataProvider(dataset_source, **settings)
 
 
+class Directory(Text):
```
mvdbeek: What's the reason to inherit from Text ?

tools/data_source/upload.py
jmchilton: Well you can see the code in red below. I'm not sure this was a @jmchilton fixme - I'd have used ``TODO`` instead of ``todo`` and misspelled something.

mvdbeek: how could I have missed that 😜

lib/galaxy/datatypes/data.py
jmchilton: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/datatypes/text.py#L24 HTML felt like the closest thing and it descends from ``Text`` also. Do you have an alternative suggestion?

mvdbeek: plain Data, I think composite datatypes wouldn't have to be restricted to text files ? Or maybe they are ?

jmchilton:
Tried rebasing on top of ISA-TAB changes but it seems something went awry and the ISA-TAB tests broke - moving back into WIP.

mvdbeek (approved):
This looks good to me, but I have the vague feeling `force_composite` should be documented a little more. I'm not sure what the right place for that would be, maybe as help attribute in the upload tool ?

jmchilton:
> I have the vague feeling force_composite should be documented a little more.

I don't think anyone can see the upload.xml help stuff, and I'm not sure I think this is the greatest entry point for general purpose usage. It allows me to do some stuff downstream in the CWL branch and might allow others to do cool stuff by attaching extra data to datasets also - but I don't think it is quite ready to advertise. If I was going to document upload API stuff I'd write up something on upload 2.0 - I think that would be a better use of time. For now I think the tests are going to be the docs, if someone asks about this at least we have something to point at. Thanks for the approve - I'm going to abuse it and just merge.

Issue #6831, opened 2018-10-08
problem when using the ShellJobRunner


Chabrier:
Hi,
i'm trying tho use the ShellJobRunner. The job submission does work, the cluster targetted does execute the job.
But the job always stay in the "grey" state (This job is waiting to run).
The job does not seem to be monitored.

The same job executed with the local job runner is correctly monitored.

Am i missing something?

My job_conf.xml lok like this:

```xml
<?xml version="1.0"?>
<!-- A sample job config that explicitly configures job running the way it is configured by default (if there is no explicit config). -->
<job_conf>
    <plugins>
        <plugin id="local" type="runner" load="galaxy.jobs.runners.local:LocalJobRunner" workers="4"/>
	<plugin id="cli" type="runner" load="galaxy.jobs.runners.cli:ShellJobRunner" workers="4"/>
    </plugins>
    <destinations default="ssh_slurm">
        <destination id="local" runner="local"/>
        <destination id="ssh_slurm" runner="cli">
            <param id="shell_plugin">SecureShell</param>
            <param id="job_plugin">Slurm</param>
            <param id="shell_username">me</param>
            <param id="shell_hostname">thecluster.fr</param>
            <param id="job_time">2:00:00</param>
            <param id="job_ncpus">4</param>
            <param id="job_partition">defq</param>
        </destination>
    </destinations>
    <tools>
	<tool id="runtask" destination="ssh_slurm" />
    </tools>
</job_conf>
```

Thanks fo the help

Cordially.

mvdbeek:
How did you set up your job handlers ?

Chabrier:
I did keep the default setting, by just copying workflow_schedulers_conf.xml.sample to  workflow_schedulers_conf.xml. I knew i was missing something. If you have any documentation entry point or configuration example, it would be helpfull.

Chabrier:
I did keep the default setting, by just copying workflow_schedulers_conf.xml.sample to workflow_schedulers_conf.xml. I knew i was missing something. If you have any documentation entry point or configuration example, it would be helpfull.

mvdbeek:
you shouldn't need to change `workflow_schedulers_conf.xml.sample`, but you should setup a `job_conf.xml` file. If you're running the current stable release (18.05) I'd suggest to start with the all-in-one job handling described here: https://docs.galaxyproject.org/en/master/admin/scaling.html#uwsgi-all-in-one-job-handling

Chabrier:
[...]

PR #6845, opened 2018-10-09
[18.05] Fix slurm cli on secure shell

mvdbeek:
Includes tests for the CLI Slurm and Torque plugins using the ParamikoShell and the SecureShell runners.

torque and slurm docker images are being built on [dockerhub](https://hub.docker.com/r/mvdbeek/galaxy-integration-docker-images/builds/)

Should fix https://github.com/galaxyproject/galaxy/issues/6831 in combination with https://github.com/galaxyproject/galaxy/pull/6836

### DIFF
--- a/lib/galaxy/jobs/runners/util/cli/shell/rsh.py
+++ b/lib/galaxy/jobs/runners/util/cli/shell/rsh.py
@@ -15,20 +15,23 @@ __all__ = ('RemoteShell', 'SecureShell', 'GlobusSecureShell', 'ParamikoShell')
 
 class RemoteShell(LocalShell):
 
-    def __init__(self, rsh='rsh', rcp='rcp', hostname='localhost', username=None, **kwargs):
+    def __init__(self, rsh='rsh', rcp='rcp', hostname='localhost', username=None, options=None, **kwargs):
         super(RemoteShell, self).__init__(**kwargs)
         self.rsh = rsh
         self.rcp = rcp
         self.hostname = hostname
         self.username = username
+        self.options = options
         self.sessions = {}
 
     def execute(self, cmd, persist=False, timeout=60):
         # TODO: implement persistence
-        if self.username is None:
-            fullcmd = '%s %s %s' % (self.rsh, self.hostname, cmd)
-        else:
-            fullcmd = '%s -l %s %s %s' % (self.rsh, self.username, self.hostname, cmd)
+        fullcmd = [self.rsh]
+        if self.options:
+            fullcmd.extend(self.options)
+        if self.username:
+            fullcmd.extend(["-l", self.username])
+        fullcmd.extend([self.hostname, cmd])
         return super(RemoteShell, self).execute(fullcmd, persist, timeout)
 
 
@@ -37,15 +40,13 @@ class SecureShell(RemoteShell):
 
     def __init__(self, rsh='ssh', rcp='scp', private_key=None, port=None, strict_host_key_checking=True, **kwargs):
         strict_host_key_checking = "yes" if strict_host_key_checking else "no"
-        rsh += " -oStrictHostKeyChecking=%s -oConnectTimeout=60" % strict_host_key_checking
-        rcp += " -oStrictHostKeyChecking=%s -oConnectTimeout=60" % strict_host_key_checking
+        options = ["-oStrictHostKeyChecking=%s" % strict_host_key_checking]
+        options.append("-oConnectTimeout=60")
         if private_key:
-            rsh += " -i %s" % private_key
-            rcp += " -i %s" % private_key
+            options.extend(['-i', private_key])
         if port:
-            rsh += " -p %s" % port
-            rcp += " -p %s" % port
-        super(SecureShell, self).__init__(rsh=rsh, rcp=rcp, **kwargs)
+            options.extend(['-p', str(port)])
+        super(SecureShell, self).__init__(rsh=rsh, rcp=rcp, options=options, **kwargs)
 
 
 class ParamikoShell(object):
--- a/lib/galaxy/jobs/runners/util/cli/shell/rsh.py
+++ b/lib/galaxy/jobs/runners/util/cli/shell/rsh.py
@@ -40,8 +40,8 @@ class SecureShell(RemoteShell):
 
     def __init__(self, rsh='ssh', rcp='scp', private_key=None, port=None, strict_host_key_checking=True, **kwargs):
         strict_host_key_checking = "yes" if strict_host_key_checking else "no"
-        options = ["-oStrictHostKeyChecking=%s" % strict_host_key_checking]
-        options.append("-oConnectTimeout=60")
+        options = ["-o", "StrictHostKeyChecking=%s" % strict_host_key_checking]
+        options.extend(["-o", "ConnectTimeout=60"])
         if private_key:
             options.extend(['-i', private_key])
         if port:
--- a/test/integration/test_cli_runners.py
+++ b/test/integration/test_cli_runners.py
@@ -4,25 +4,12 @@ import shlex
 import string
 import subprocess
 import tempfile
-import time
 import unittest
 
-import paramiko
 from base import integration_util
 from base.populators import skip_without_tool
 
 from .test_job_environments import BaseJobEnvironmentIntegrationTestCase
-
-
-def deploy_key(key, server, username, password, port=22):
-    client = paramiko.SSHClient()
-    client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
-    client.connect(server, username=username, password=password, port=port)
-    client.exec_command('rm -R .ssh/')
-    client.exec_command('mkdir -p .ssh/')
-    client.exec_command('echo "%s" > .ssh/authorized_keys' % key)
-    client.exec_command('chmod 644 .ssh/authorized_keys')
-    client.exec_command('chmod 700 .ssh/')
 
 
 def generate_key(key_path):
@@ -32,10 +19,13 @@ def generate_key(key_path):
 
 @integration_util.skip_unless_docker()
 def start_ssh_docker(container_name, jobs_directory, port=10022, image='agaveapi/slurm'):
+    key_file = tempfile.NamedTemporaryFile(suffix='_slurm_integration_ssh_key', delete=True).name
+    generate_key(key_file)
+    pub_key = "%s.pub" % key_file
     START_SLURM_DOCKER = ['docker',
                           'run',
                           '-h',
-                          'docker.example.com',
+                          'localhost',
                           '-p',
                           '{port}:22'.format(port=port),
                           '-d',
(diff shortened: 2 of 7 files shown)


lib/galaxy/jobs/runners/util/cli/shell/rsh.py
```diff
@@ -26,9 +26,9 @@ def __init__(self, rsh='rsh', rcp='rcp', hostname='localhost', username=None, **
     def execute(self, cmd, persist=False, timeout=60):
         # TODO: implement persistence
         if self.username is None:
-            fullcmd = '%s %s %s' % (self.rsh, self.hostname, cmd)
+            fullcmd = '%s %s "%s"' % (self.rsh, self.hostname, cmd)
         else:
-            fullcmd = '%s -l %s %s %s' % (self.rsh, self.username, self.hostname, cmd)
+            fullcmd = '%s -l %s %s "%s"' % (self.rsh, self.username, self.hostname, cmd)
```
nsoranzo: What about using a list instead of a string?

mvdbeek: I don't think this works because we construct a ssh command here, that's why we need shell="true" as well in https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/jobs/runners/util/cli/shell/local.py#L47

mvdbeek: actually it might work, I had tried something else before.

nsoranzo: Yep, you can set `shell=True` only if `cmd` is a string.

jmchilton:
Holy crap these tests are amazing!

jmchilton (approved):

lib/galaxy/jobs/runners/util/cli/shell/rsh.py
mvdbeek: Hmm, is this what you had in mind?
```
diff --git a/lib/galaxy/jobs/runners/util/cli/shell/local.py b/lib/galaxy/jobs/runners/util/cli/shell/local.py
index a5b570abaa..7e6b65c521 100644
--- a/lib/galaxy/jobs/runners/util/cli/shell/local.py
+++ b/lib/galaxy/jobs/runners/util/cli/shell/local.py
@@ -44,7 +44,7 @@ class LocalShell(BaseShellExec):
 
     def execute(self, cmd, persist=False, timeout=DEFAULT_TIMEOUT, timeout_check_interval=DEFAULT_TIMEOUT_CHECK_INTERVAL, **kwds):
         outf = TemporaryFile()
-        p = Popen(cmd, shell=True, stdin=None, stdout=outf, stderr=PIPE)
+        p = Popen(cmd, stdin=None, stdout=outf, stderr=PIPE)
         # poll until timeout
 
         for i in range(int(timeout / timeout_check_interval)):
diff --git a/lib/galaxy/jobs/runners/util/cli/shell/rsh.py b/lib/galaxy/jobs/runners/util/cli/shell/rsh.py
index d1b3b5f74f..c3e370dfc1 100644
--- a/lib/galaxy/jobs/runners/util/cli/shell/rsh.py
+++ b/lib/galaxy/jobs/runners/util/cli/shell/rsh.py
@@ -26,9 +26,11 @@ class RemoteShell(LocalShell):
     def execute(self, cmd, persist=False, timeout=60):
         # TODO: implement persistence
         if self.username is None:
-            fullcmd = '%s %s "%s"' % (self.rsh, self.hostname, cmd)
+            fullcmd = [self.rsh, self.hostname, "'%s'" % cmd]
         else:
-            fullcmd = '%s -l %s %s "%s"' % (self.rsh, self.username, self.hostname, cmd)
+            fullcmd = [self.rsh, '-l', self.username, self.hostname, "'%s'" % cmd]
         return super(RemoteShell, self).execute(fullcmd, persist, timeout)
```

This fails with `OSError: [Errno 2] No such file or directory`

mvdbeek: Ah, this is because of the additional options, got it.

mvdbeek: Awesome, no need for this hackish fix then! 🏅 to 🦅 👁 !

lib/galaxy/jobs/runners/util/cli/shell/rsh.py
```diff
@@ -37,15 +40,13 @@ class SecureShell(RemoteShell):
...
+        options.append("-oConnectTimeout=60")
         if private_key:
-            rsh += " -i %s" % private_key
-            rcp += " -i %s" % private_key
+            options.extend(['-i', private_key])
         if port:
-            rsh += " -p %s" % port
-            rcp += " -p %s" % port
-        super(SecureShell, self).__init__(rsh=rsh, rcp=rcp, **kwargs)
+            options.extend(['-p', str(port)])
```
nsoranzo, line 48: The `str()` casting shouldn't be needed.

lib/galaxy/jobs/runners/util/cli/shell/rsh.py
```diff
@@ -37,15 +40,13 @@ class SecureShell(RemoteShell):
 
     def __init__(self, rsh='ssh', rcp='scp', private_key=None, port=None, strict_host_key_checking=True, **kwargs):
         strict_host_key_checking = "yes" if strict_host_key_checking else "no"
-        rsh += " -oStrictHostKeyChecking=%s -oConnectTimeout=60" % strict_host_key_checking
-        rcp += " -oStrictHostKeyChecking=%s -oConnectTimeout=60" % strict_host_key_checking
+        options = ["-oStrictHostKeyChecking=%s" % strict_host_key_checking]
+        options.append("-oConnectTimeout=60")
```
nsoranzo: I think that `-o Keyword=Value` is more common, what about:
```python
        options = ["-o", "StrictHostKeyChecking=%s" % strict_host_key_checking]
        options.extend(["-o", "ConnectTimeout=60"])
```

test/integration/test_cli_runners.py
```diff
@@ -0,0 +1,154 @@
...
+    @classmethod
+    def tearDownClass(cls):
+        stop_ssh_docker(container_name=cls.container_name)
+        super(BaseCliIntegrationTestCase, cls).tearDownClass()
+
+    @classmethod
+    def handle_galaxy_config_kwds(cls, config, ):
+        config["jobs_directory"] = cls.jobs_directory
+        config["file_path"] = cls.jobs_directory
+        config["job_config_file"] = cli_job_config(remote_connection=cls.remote_connection, shell_plugin=cls.shell_plugin)
```
nsoranzo: Missing ` , job_plugin=cls.job_plugin` at the end of this line?

lib/galaxy/jobs/runners/util/cli/shell/rsh.py
mvdbeek: It is for the unit test, overall I think it's sensible for a `port` variable to accept integer (as paramiko for instance does).

lib/galaxy/jobs/runners/util/cli/shell/rsh.py
mvdbeek: sure, looks better

test/integration/test_cli_runners.py
mvdbeek: oh yeah

lib/galaxy/jobs/runners/util/cli/shell/rsh.py
nsoranzo: I mean that `Popen` does the casting already.

mvdbeek:
```
In [3]: subprocess.Popen(['echo', 5])
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-6bc2fd563ae2> in <module>()
----> 1 subprocess.Popen(['echo', 5])

/usr/local/Cellar/python@2/2.7.15_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.pyc in __init__(self, args, bufsize, executable, stdin, stdout, stderr, preexec_fn, close_fds, shell, cwd, env, universal_newlines, startupinfo, creationflags)
    392                                 p2cread, p2cwrite,
    393                                 c2pread, c2pwrite,
--> 394                                 errread, errwrite)
    395         except Exception:
    396             # Preserve original exception in case os.close raises.

/usr/local/Cellar/python@2/2.7.15_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.pyc in _execute_child(self, args, executable, preexec_fn, close_fds, cwd, env, universal_newlines, startupinfo, creationflags, shell, to_close, p2cread, p2cwrite, c2pread, c2pwrite, errread, errwrite)
   1045                         raise
   1046                 child_exception = pickle.loads(data)
-> 1047                 raise child_exception
   1048
   1049

TypeError: execv() arg 2 must contain only strings
```

nsoranzo: Ops, I assumed that was the case for what I remembered from lib/galaxy/jobs/runners/slurm.py , but evidently `job_id` must always be a string there. Sorry for the noise!

mvdbeek: No worries, definitely appreciate the review.

lib/galaxy/jobs/runners/util/cli/shell/rsh.py
```diff
@@ -37,15 +40,13 @@ class SecureShell(RemoteShell):
 
     def __init__(self, rsh='ssh', rcp='scp', private_key=None, port=None, strict_host_key_checking=True, **kwargs):
         strict_host_key_checking = "yes" if strict_host_key_checking else "no"
-        rsh += " -oStrictHostKeyChecking=%s -oConnectTimeout=60" % strict_host_key_checking
-        rcp += " -oStrictHostKeyChecking=%s -oConnectTimeout=60" % strict_host_key_checking
+        options = ["-o" , "StrictHostKeyChecking=%s" % strict_host_key_checking]
```
nsoranzo: There's an extra space before the comma.

nsoranzo (approved):

test/integration/test_cli_runners.py
```diff
@@ -0,0 +1,150 @@
...
+    os.remove(remote_connection.public_key)
+
+
+def cli_job_config(remote_connection, shell_plugin='ParamikoShell', job_plugin='Slurm'):
+    job_conf_template = string.Template("""<job_conf>
+    <plugins>
+        <plugin id="cli" type="runner" load="galaxy.jobs.runners.cli:ShellJobRunner" workers="1"/>
+    </plugins>
+    <destinations default="ssh_slurm">
+        <destination id="ssh_slurm" runner="cli">
```
nsoranzo, line 66: Not really important, but the destination id should probably be more generic.

PR #6848, opened 2018-10-10
Indicate language in client docs to enable syntax highlighting.

jdavcs:
Enables syntax highlighting for code blocks in .md files. (unless there's a reason why we don't use this).
Also, added indent in galaxy-menus.md.


hexylena (approved):

dannon (changes requested):
@ic4f Thanks for taking a look at this.  To your question, these are primarily designed to be viewed as a part of the styleguide (`yarn run styleguide`), and not viewed as .md.  In fact, specifying the type here breaks the styleguide in my hands, though if you can find a way to satisfy both styleguide and .md, that'd be great.

jxtx:
Just from looking at: https://raw.githubusercontent.com/vue-styleguidist/vue-styleguidist/master/examples/basic/src/components/Button/Readme.md

I think if you use \```vue instead of \```html it should work in styleguide and github?

jdavcs:
I didn't know about the styleguide. Will take a quick look (if it doesn't work for both styleguide and md, will leave as is.)
Both jsx and vue are not supported by my chrome md viewer extension, but they are highlighted correctly on github. 
Checking about styleguide....

jdavcs:
Found it. @jxtx is right: changing html to vue (or jsx/js/javascript) fixes the issue: everything is rendered correctly. As per vue-styleguidist docs: "Examples with all other languages are rendered only as highlighted source code, not an actual component" (which is different from React's original styleguidist). I'll change the code accordingly and will resumbit.

Question: since these .md files are used as input to a tool, should we have a very brief README in this folder alerting people not to modify the files?

I'll submit the fix for the svg file as a separate PR. For now, to compile the styleguide, simply comment out line 401 in client/galaxy/style/scss/base.scss.

jdavcs:
I added a README to the docs directory. I tried adding comments to the .md files instead, but styleguidist renders them as html.

dannon:
@ic4f Thanks!

PR #6855, opened 2018-10-11
[DONOTMERGE] Revert selective metadata flushing - seems to break unrelated dataset state changes.

jmchilton:


martenson:
merged here: https://github.com/galaxyproject/galaxy/pull/6858

PR #6857, opened 2018-10-11
update paramiko dep

martenson:


nsoranzo:
We will get this by merging forward.

PR #6865, opened 2018-10-13
Share ssh key generation across tests

### LINKED ISSUE #6860: combine ssh key methods
One in https://github.com/galaxyproject/galaxy/blob/dev/test/unit/test_remote_shell.py#L55, another one in https://github.com/galaxyproject/galaxy/blob/release_18.05/test/integration/test_cli_runners.py#L16

mvdbeek:
Closes https://github.com/galaxyproject/galaxy/issues/6860

(and also fixes the integration test on python 3)

### DIFF
--- a/test/base/ssh_util.py
+++ b/test/base/ssh_util.py
@@ -1,0 +1,19 @@
+import collections
+import tempfile
+
+from Crypto.PublicKey import RSA
+
+
+def generate_ssh_keys():
+    """Returns a named tuple with private and public key and their paths."""
+    key = RSA.generate(2048)
+    private_key = key.export_key()
+    public_key = key.publickey().export_key(format='OpenSSH')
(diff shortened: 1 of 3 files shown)


lib/galaxy/util/__init__.py
```diff
@@ -38,6 +38,7 @@
     default_enter,
     remap,
 )
+from Crypto.PublicKey import RSA
```
nsoranzo: I think @jmchilton would like to keep the requirements for galaxy-lib to a minimum, so maybe use try/except or put the method somewhere in test/ ?

mvdbeek: I've moved the function to test/base/ssh_util.py and rebased the commit

nsoranzo (approved):

PR #6867, opened 2018-10-13
[18.09] Keep xlsx files compressed

### LINKED ISSUE #6849: sniffing of xlsx file as blastxml
reported on galaxy-dev, example file: [Book1.xlsx](https://github.com/galaxyproject/galaxy/files/2465937/Book1.xlsx)
screenshot 2018-10-10 14 18 04

mvdbeek:
Fixes #6849

### DIFF
--- a/lib/galaxy/datatypes/binary.py
+++ b/lib/galaxy/datatypes/binary.py
@@ -1552,6 +1552,7 @@ class GAFASQLite(SQlite):
 class Xlsx(Binary):
     """Class for Excel 2007 (xlsx) files"""
     file_ext = "xlsx"
+    compressed = True
 
     def sniff(self, filename):
         # Xlsx is compressed in zip format and must not be uncompressed in Galaxy.
--- a/lib/galaxy/datatypes/upload_util.py
+++ b/lib/galaxy/datatypes/upload_util.py
@@ -30,6 +30,7 @@ def handle_upload(
 ):
     stdout = None
     converted_path = None
+    multi_file_zip = False
 
     # Does the first 1K contain a null?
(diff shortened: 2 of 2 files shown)


nsoranzo:
LGTM, should we have a unit test for this?

martenson:
@nsoranzo I started one earlier today, I will finish it and push to @mvdbeek branch

mvdbeek:
I think it'd be nice to test the shipped test data in an integration test case with and without auto_decompress, but I'd target dev with that.

nsoranzo:
> I think it'd be nice to test the shipped test data in an integration test case with and without auto_decompress, but I'd target dev with that.

Agreed, but for 18.09 we could just add a unit test to verify that xlsx are sniffed correctly.

mvdbeek:
It's not that I don't want to add a unit test for sniffing, but it wouldn't have caught the issue here, since the sniffing worked fine before, unless that would be a more intricate unit test that tests the  upload_util module.

nsoranzo:
Isn't https://github.com/galaxyproject/galaxy/issues/6849 about xlsx sniffed as blastxml ?

mvdbeek:
Yes, but it's not the sniffer's fault, if you have a datatype that is not marked as compressed the first file in the archive will be unpacked and sniffed, which in the case of xlsx files is a xml manifest.

martenson:
@mvdbeek The manifest should still not be sniffed as blastxml given the process described by @peterjc. However given the work in https://github.com/galaxyproject/galaxy/pull/6869 I think unit tests is not getting us much here and we are pretty unlikely to touch this in release branch I guess.

Issue #6706, opened 2018-09-12
List of style update issues


guerler:
Follow-ups for #6422.

from @jxtx:

- System fonts
- Eliminate "lavender"
- Improve default colors
- Fewer borders
- More flex (resizeable) components
- Better / more consistent use of whitespace (far from complete)

from @mvdbeek:

Alright, so these things I have found:

- [x] in multi-select mode the checkboxes are slightly cut off:
![screen shot 2018-09-11 at 10 23 15](https://user-images.githubusercontent.com/6804901/45347591-c6da5580-b5ac-11e8-9b3b-e931df20eef2.png)

- [x] The done button for history and HDA annotations is too big:
![screen shot 2018-09-11 at 10 24 28](https://user-images.githubusercontent.com/6804901/45347649-ecfff580-b5ac-11e8-9cfe-dfa7da7af332.png)

- [x] The placement of the integer box here seems off (and the tool header space could be used better)
![screen shot 2018-09-11 at 10 26 21](https://user-images.githubusercontent.com/6804901/45347824-4536f780-b5ad-11e8-836b-d81195c165b6.png)

- [x] Tool form preview in the workflow editor still uses the old style
![screen shot 2018-09-11 at 10 31 27](https://user-images.githubusercontent.com/6804901/45348074-f047b100-b5ad-11e8-9f16-a9100fa919f6.png)

- [x] View workflow also still uses the old style
![screen shot 2018-09-11 at 10 38 25](https://user-images.githubusercontent.com/6804901/45348418-dfe40600-b5ae-11e8-9d36-f2896ac9432c.png)

- [x] Share workflow uses old style:
![screen shot 2018-09-11 at 10 39 22](https://user-images.githubusercontent.com/6804901/45348464-03a74c00-b5af-11e8-906c-10a6412edfbc.png)

- [x] Visualizations partly use old style
![screen shot 2018-09-11 at 10 40 16](https://user-images.githubusercontent.com/6804901/45348495-22a5de00-b5af-11e8-8e99-ce90e840d52f.png)

- [x] Importing workflow uses old style:
![screen shot 2018-09-11 at 10 41 52](https://user-images.githubusercontent.com/6804901/45348571-57b23080-b5af-11e8-9e34-5a3156d47fa9.png)

Various forms in the admin interface use the old style

-  [x] Data tables
[...]

PR #6868, opened 2018-10-13
Use card-styles for display applications, job manager and other admin views

guerler:
Follow-up on style revision #6706. Replaces legacy forms with bootstrap cards.


martenson:
Most of these pages would benefit from a redesign, but given our approach to admins: 'functionality fast, styles last' this is an improvement, thanks @guerler

martenson (approved):

PR #6869, opened 2018-10-14
Datatype upload integration test and fixes

mvdbeek:
Hopefully catches issues like #6849 earlier.

This also tightens up fastqsanger sniffing, introduces a fastqcssanger sniffer (which is unambiguously not fastqsanger), fixes the memepsp sniffer that was fairly broken, fixed a segfault when uploading bcf.uncompressed, fixes datatypes that set binary text or numpy integer/floats on MetadataElement objects on python 3, and skips newline / space2tab conversion when check_binary returns a false negative but the datatype was detected as being binary.

To test additional new datatypes just drop a test file into lib/galaxy/datatypes/test and it will be uploaded and checked. This should catch sniffers with too many false positives.


mvdbeek:
Already caught issues in bcf_uncompress, meme_psp and fastqcssanger, and I'm sure validating the set metadata will catch more still.

lib/galaxy/datatypes/sniff.py
```diff
@@ -359,7 +359,7 @@ def guess_ext(fname, sniff_order, is_binary=False):
     >>> fname = get_test_fname('3unsorted.bam')
     >>> guess_ext(fname, sniff_order)
     'unsorted.bam'
-    >>> fname = get_test_fname('test.idpDB')
+    >>> fname = get_test_fname('test.idpdb')
```
nsoranzo, line 363: This is the filename, which was correct before. Revert?

mvdbeek: No, this is intentional, to map the file extension to the datatype extension. I've also moved the file.

mvdbeek:
@galaxybot test py3

mvdbeek:
@galaxybot test py3

mvdbeek:
@galaxybot test py3

mvdbeek:
@galaxybot test py3

mvdbeek:
https://jenkins.galaxyproject.org/job/docker-api-py3/163/testReport/test.api.test_workflows/WorkflowsApiTestCase/test_workflow_run_zip_collections/ is probably unrelated but needs fixing

lib/galaxy/datatypes/binary.py
```diff
@@ -657,7 +657,8 @@ def set_meta(self, dataset, overwrite=True, **kwd):
                                        '__dataset_%d_%s' % (dataset.id, os.path.basename(index_file.file_name)))
         os.symlink(dataset.file_name, dataset_symlink)
         try:
-            pysam.bcftools.index(dataset_symlink)
+            cmd = ['python', '-c', "import pysam.bcftools; pysam.bcftools.index('%s')" % (dataset_symlink)]
```
nsoranzo: Can you add a comment along the lines of the message for commit cc177b550f7d04c2d40c6ec16d82115bf86d60fc ?

lib/galaxy/datatypes/binary.py
```diff
@@ -657,7 +657,8 @@ def set_meta(self, dataset, overwrite=True, **kwd):
...
+            subprocess.check_call(cmd)
```
nsoranzo, line 661: This is not needed any more after commit e561267bfc6e36853b4f56ebabda7acbdfa9eeb, right?

mvdbeek: It's still good to do it this way because you may set metadata on an invalid bcf file, causing a segfault. If metadata is being set externally it's not a huge issue, but the internal metadata setting would just kill the interpreter.

mvdbeek:
@galaxybot test this

mvdbeek:
Not sure what's going on with the selenium tests, but at least the failing API test will be fixed by ordering the workflow conversion as done in gxformat2 >= 0.7.0

mvdbeek:
@galaxybot test py3

mvdbeek:
@galaxybot test py3

mvdbeek:
The selenium python 3 docker tests pass fine on my laptop, I guess that the disk is full on the node that runs the tests. So all of the remaining test failures are false negatives and the flakey python 3 API tests would be fixed by upgrading gxformat2.

lib/galaxy/datatypes/test/2.txt
```diff
@@ -1 +1 @@
-a	2
\ No newline at end of file
+a	2
```
jmchilton, line 1: Is this changing the newline used? Why?

mvdbeek: Yes, so that we can verify that a upload-> download cycle still results in the same file (using a hash). We could also allow some diff, or skip that test.

mvdbeek: Or test without newline conversion

mvdbeek:
@galaxybot test py3

mvdbeek:
@galaxybot test py3

mvdbeek:
@galaxybot test py3

nsoranzo:
Need to rebase again unfortunately.

mvdbeek:
@galaxybot test py3

mvdbeek:
@galaxybot test py3

jmchilton:
Wonderful!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

martenson:
This is damn great, thank you so much @mvdbeek !

PR #6871, opened 2018-10-15
prevent double sanitizing history name when renaming

### LINKED ISSUE #169: Renaming histories that contain ' quotes are problematic
> ` 

does not cause a problem

> ' 

does cause a problem

Everything before the END was the original history name, everthing after is what I added during renaming.

screen shot 2018-10-12 at 12 42 43 pm

almahmoud:
This is intended to fix: https://github.com/galaxyproject/usegalaxy-playbook/issues/169 / https://github.com/galaxyproject/galaxy/issues/6873

I was originally trying to "unescape" the name whenever it was displayed instead of removing the escape altogether, but given that names for workflows are only sanitized and not escaped, and that the history name is not escaped when changed from the panel on the right, I thought it might just be that it didn't need to be escaped in the first place.
If I am wrong and it does need to be escaped, let me know and I can instead try to convert it to its value whenever returned to the UI, while storing it escaped


martenson:
Thanks @almahmoud !

PR #6876, opened 2018-10-15
Specify POST and GET actions for cloudauthz API actions

VJalili:


martenson:
I think the suggestion was that the endpoint will be the same but the method will be different depending on the action (which is Galaxy 'standard' REST). This implementation has two methods, but also two endpoints.

edit: Also if the previous PR made it to the release this needs to target that, so we do not offer insta-deprecated API.

VJalili:
Thanks for the clarification @martenson. 

This functionality is not available in 18.09, so I guess we can safely target `dev`.

martenson:
@VJalili Looks good, please update the docs of this endpoint to reflect the change.

VJalili:
@martenson thanks for pointing that out; it it updated now.

martenson:
thanks @VJalili

PR #6878, opened 2018-10-15
Use small button for editable text fields in history annotation

guerler:
Fixes issue listed in #6706.

image


dannon:
@guerler While you're at this, I think it'd be an improvement for the "Done" button to be to the right of the box, not the left.

PR #6884, opened 2018-10-16
Improve resilience on handling sharing attributes

guerler:


martenson:
is this broken for some items? do we need to backport?

guerler:
It caused issues with sharing visualizations in dev.

lib/galaxy/web/base/controller.py
```diff
@@ -1407,17 +1407,17 @@ def sharing(self, trans, id, payload=None, **kwd):
             action = payload.get("action")
             if action == "make_accessible_via_link":
                 self._make_item_accessible(trans.sa_session, item)
-                if item.has_possible_members and payload.get("make_members_public", False):
+                if hasattr(item, "has_possible_members") and payload.get("make_members_public", False):
```
martenson:
```suggestion
                if hasattr(item, "has_possible_members"):
                    if item.has_possible_members and payload.get("make_members_public", False):
```

martenson: Trying new github feature here. :)

The point is that it should still check the truthiness of the attribute if it exists.

guerler: Makes sense.

guerler:
@galaxybot test this.

PR #6885, opened 2018-10-17
[18.05] Create composite file archives in new_file_path instead of TMP

mvdbeek:
This hopefully fixes the issue that @lparsons encountered when using xsendfile, where the Apache user can't access the created file/stream.


nsoranzo (approved):

martenson:
@lparsons can you please test?

lparsons:
Looks like it works (once I had the sysadmins fiddle with SELinux). One concern I have is that things will get cleaned up. What is the mechanism to clean up the `database/tmp` dir?

PR #6898, opened 2018-10-18
prevent sanitized history names being saved in database

martenson:
followup to https://github.com/galaxyproject/galaxy/pull/6871

Sanitizing/escaping strings before we put them in the database is not buying us anything I think. Before #6871 this used to sanitize/escape the strings already sanitized/escaped and stored in the DB which creates a huge mess. 

ping @almahmoud

### DIFF
--- a/lib/galaxy/webapps/galaxy/controllers/history.py
+++ b/lib/galaxy/webapps/galaxy/controllers/history.py
@@ -1207,8 +1207,7 @@ class HistoryController(BaseUIController, SharableMixin, UsesAnnotations, UsesIt
                     messages.append('History \'%s\' does not appear to belong to you.' % cur_name)
                 # skip if it wouldn't be a change
                 elif new_name != cur_name:
-                    # sanitize, set, and log the change
-                    h.name = sanitize_html(new_name)
+                    h.name = new_name
                     trans.sa_session.add(h)
                     trans.sa_session.flush()
                     trans.log_event('History renamed: id: %s, renamed to: %s' % (str(h.id), new_name))
--- a/lib/galaxy/webapps/galaxy/controllers/history.py
+++ b/lib/galaxy/webapps/galaxy/controllers/history.py
@@ -17,7 +17,6 @@ from galaxy.model.item_attrs import (
 from galaxy.util import listify, Params, parse_int, sanitize_text
 from galaxy.util.create_history_template import render_item
 from galaxy.util.odict import odict
-from galaxy.util.sanitize_html import sanitize_html
 from galaxy.web import url_for
 from galaxy.web.base.controller import (
     BaseUIController,


nsoranzo:
> Sanitizing/escaping strings before we put them in the database is not buying us anything I think.

If this will allow users to put HTML tags in the database, I don't think it's a good idea. I was actually arguing for using `escape()` to strip all HTML tags instead of some.

martenson:
@nsoranzo I realize this is different from what you suggested. But this is what I ended up proposing. I am slightly in favor of not stripping html tags.

afaik `markupsafe.escape()` does not strip tags (just escapes) and storing html-escaped values seems bad. 

Edit: this PR is is how it works in the other renaming endpoint, and with other objects too: 
screenshot 2018-10-18 15 44 00

nsoranzo:
> I am slightly in favor of not stripping html tags. afaik markupsafe.escape() does not strip tags (just escapes) and storing html-escaped values seems bad.

What about
```python
from bleach import clean
...
h.name = clean(new_name, tags=[], strip=True)
```

as in `lib/galaxy/tools/search/__init__.py` ?

martenson:
@nsoranzo Yes, that is doable, but as I said, I prefer (albeit slightly) not to modify user-entered values here.

hexylena:
Most of the advice I've seen tends towards sanitization of outputs being the preferable solution:

- https://security.stackexchange.com/questions/95325/input-sanitization-vs-output-sanitization
- https://www.owasp.org/index.php/XSS_(Cross_Site_Scripting)_Prevention_Cheat_Sheet#RULE_.236_-_Sanitize_HTML_Markup_with_a_Library_Designed_for_the_Job
- https://www.owasp.org/index.php/Injection_Prevention_Cheat_Sheet_in_Java

My only concern was that *some* weird person (don't know [anyone like that](https://github.com/galaxyproject/tools-iuc/blob/master/tools/jbrowse/jbrowse.xml#L90)) is doing something sub-optimal with the history name. But I've just tested that and it looks fine since all the `<`/`>` were escaped.

Could we maybe add a short test case for that scenario? Just in case it's ever changed, I don't want to expose users to possible issues. Maybe something like this?

```
<tool id="test_history_sanitization" name="asdf" version="0.0.1">
    <command>
    echo "${useless.history.get_display_name()}" > $out_file1
    </command>
    <inputs>
        <param name="useless" type="integer" value="0" />
    </inputs>
    <outputs>
        <data format="tabular" name="out_file1"/>
    </outputs>
    <tests>
        <test>
            <output name="out_file1" file="__lt__script__gt__alert(1)__lt__/script__gt__" />
        </test>
    </tests>
</tool>
```

mvdbeek:
@galaxybot test py3

jmchilton:
I prefer cleaning to escaping. I'm not sure about whether to clean (@nsoranzo's solution) or not (@martenson's solution). Certainly if this was a well tested application and I had a lot of confidence about how the value was going to be used in every context I'd say @martenson's solution is much more correct. Is it that and am I confident? No. Does my lack of confidence mean we shouldn't do the right thing? No.

Another thing we could do is check the output of clean and throw a validation exception if the name is modified. That would seem more correct than changing the user input silently and it would be safer than not touching the input.

It occurs to me though if this name had never been cleaned in the first place - we wouldn't think about cleaning it here. We're not going in and cleaning everything else - so why spend the time on this and not just merge as is.

Long winded +0 for whoever wants it I guess 😆.

jmchilton (approved):

martenson:
@nsoranzo please let me know if you want to discuss this further, merging

PR #5787, opened 2018-03-27
Feature/isa data type

pkrog:
New data types ISA-tab and ISA-json, proposed by the [PhenoMeNal](https://github.com/phnmnl) team. These new types have been successfully used inside the PhenoMeNal project, in tools such as 
[mtbls-dwnld](https://github.com/workflow4metabolomics/mtbls-dwnld) and [isa2w4m](https://github.com/workflow4metabolomics/isa2w4m).

The uploader has been modified in order to be able to update the name of the composite dataset after uploading and opening the archive.

[isa-rwval](https://github.com/ISA-tools/isa-rwval) has been added as a new dependency. This package is required in order to read ISA archive content and display a summary to the user. It triggers the installation of packages `networkx` and `six`.

This PR is in relation with issue #3134 .

PR #6906, opened 2018-10-22
add isa-rwal to default pipfile

martenson:
it is in pinned already

followup to https://github.com/galaxyproject/galaxy/pull/5787


nsoranzo:
Some of the isa-rwal dependencies are missing from requirements.txt, I'll push a fix to your branch if that's OK.

martenson:
@nsoranzo that is very ok

nsoranzo (approved):

mvdbeek:
There are some other things that are not quite right in https://github.com/galaxyproject/galaxy/pull/5787, is anyone on this already ?

nsoranzo:
@mvdbeek I'm not.

martenson:
@mvdbeek I don't think so.

nsoranzo:
@mvdbeek Small follow-up in commit https://github.com/galaxyproject/galaxy/pull/6921/commits/21189b213f10687ca0e211bd5fc379f5316e2769 .

PR #6911, opened 2018-10-23
Refactor collection mapping workflows toward independence from tools.

jmchilton:
Refactor workflow module method ``get_data_inputs`` into ``get_all_inputs`` and use the resulting dictionaries to reason about if collection mapping should occur during invocation of tools. Using these dictionaries instead of explicit tool input objects should allow reuse within other module types since they may produce the same interface.

### DIFF
--- a/lib/galaxy/tools/parameters/history_query.py
+++ b/lib/galaxy/tools/parameters/history_query.py
@@ -16,10 +16,7 @@ class HistoryQuery(object):
         return HistoryQuery(**kwargs)
 
     @staticmethod
-    def from_parameter(param, collection_type_descriptions):
-        """ Take in a tool parameter element.
-        """
-        collection_types = param.collection_types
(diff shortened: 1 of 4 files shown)


jmchilton:
Here (https://github.com/galaxyproject/galaxy/pull/6919/commits/6d7632e8e457216b0de8431e6775257e113a8970) you can see this refactoring enabling a nice simplification/unification of data and non-data parameters in the step parameter branch corresponding to #6919. And here (https://github.com/common-workflow-language/galaxy/pull/47/commits/7598bb5bd73cc2c8f1834dbd234a454173bf1b94#diff-5cab48d5f01fec3fa30d17f4bf5d1848R22) you can see how it enables a bunch of features in the main CWL commit in that branch - defaults, expression inputs, field parameter types, etc..

mvdbeek:
@galaxybot test py3

mvdbeek (approved):

PR #6087, opened 2018-05-08
Fix JobWrapper.get_destination_configuration() so that it actually checks the JobDestination params

natefoo:
I believe this was the original intent, but I could be wrong. Currently, when this method is called it only checks the persisted `destination_params` on the `Job` object and falls back to the app config. But I assume it should also be looking at the job destination's params. Because the params are persisted to the `Job` fairly late, you can't really use destination params for things like `outputs_to_working_directory` as intended.

With this change, it will now check the persisted params, followed by the destination params, and finally, the app config.

PR #6912, opened 2018-10-23
[18.09] Fix stopping DRMAA jobs

natefoo:
I broke them in #6087. Also improve logging around this process.

### DIFF
--- a/lib/galaxy/jobs/runners/drmaa.py
+++ b/lib/galaxy/jobs/runners/drmaa.py
@@ -310,21 +310,27 @@ class DRMAAJobRunner(AsynchronousJobRunner):
 
     def stop_job(self, job):
         """Attempts to delete a job from the DRM queue"""
+        dest_params = {}
+        try:
+            # fully dynamic destinations may not be defined in the job config
+            dest_params = self.app.job_config.get_destination(job.destination_id).params
+        except KeyError:
+            pass
         try:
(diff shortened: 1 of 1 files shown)


nsoranzo:
@natefoo Wrong target branch?

natefoo:
Yep thanks.

nsoranzo:
This looks good as a fix for 18.09, but I was wondering if `stop_job()` in `BaseJobRunner` and subclasses should actually receive a `JobWrapper` object as parameter instead of a `Job` object.
This should simplify this code and put the `stop_job()` method in line with the rest of the `BaseJobRunner` methods.

natefoo:
Yeah, that makes sense to me. In fact, all the ones receiving both the `Job` and `JobWrapper` should probably just receive `JobWrapper`. I think in the past the idea was that things that didn't need a `JobWrapper` should just get a `Job` since it would be quicker not to create a `JobWrapper` when it was unneeded.

jmchilton:
> In fact, all the ones receiving both the Job and JobWrapper should probably just receive JobWrapper. I think in the past the idea was that things that didn't need a JobWrapper should just get a Job since it would be quicker not to create a JobWrapper when it was unneeded.

Perhaps it was a premature optimization - but I've sometimes passed a job parameter directly to something that had access to a job_wrapper to prevent the SA interactions in ``get_job``.

PR #6914, opened 2018-10-24
Allow ``remove_value`` filter on data tables

nsoranzo:
The `RemoveValueFilter.filter_options()` method was expecting `options` to be a list of lists, where each inner list contains exactly 3-elements (name, value, selected), which is the format returned by the `data_meta` filter. In fact, the only known use of the `remove_falue` filter is in the tool `tools/maf/maf_to_interval.xml`, to filter dataset metadata.

This commit generalises the filter to look for a column of the dynamic options named `value`, and use it during the filtering.

Fix https://github.com/galaxyproject/tools-iuc/issues/2151

### DIFF
--- a/lib/galaxy/tools/parameters/dynamic_options.py
+++ b/lib/galaxy/tools/parameters/dynamic_options.py
@@ -388,7 +388,7 @@ class RemoveValueFilter(Filter):
         self.ref_name = elem.get("ref", None)
         self.meta_ref = elem.get("meta_ref", None)
         self.metadata_key = elem.get("key", None)
-        assert self.value is not None or ((self.ref_name is not None or self.meta_ref is not None) and self.metadata_key is not None), ValueError("Required 'value' or 'ref' and 'key' attributes missing from filter")
+        assert self.value is not None or self.ref_name is not None or (self.meta_ref is not None and self.metadata_key is not None), ValueError("Required 'value', or 'ref', or 'meta_ref' and 'key' attributes missing from filter")
         self.multiple = string_as_bool(elem.get("multiple", "False"))
         self.separator = elem.get("separator", ",")
 
--- a/lib/galaxy/tools/parameters/dynamic_options.py
+++ b/lib/galaxy/tools/parameters/dynamic_options.py
@@ -388,7 +388,7 @@ class RemoveValueFilter(Filter):
         self.ref_name = elem.get("ref", None)
         self.meta_ref = elem.get("meta_ref", None)
         self.metadata_key = elem.get("key", None)
-        assert self.value is not None or ((self.ref_name is not None or self.meta_ref is not None)and self.metadata_key is not None), ValueError("Required 'value' or 'ref' and 'key' attributes missing from filter")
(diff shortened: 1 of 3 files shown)


lib/galaxy/tools/parameters/dynamic_options.py
```diff
@@ -388,7 +388,7 @@ def __init__(self, d_option, elem):
         self.ref_name = elem.get("ref", None)
         self.meta_ref = elem.get("meta_ref", None)
         self.metadata_key = elem.get("key", None)
-        assert self.value is not None or ((self.ref_name is not None or self.meta_ref is not None)and self.metadata_key is not None), ValueError("Required 'value' or 'ref' and 'key' attributes missing from filter")
+        assert self.value is not None or ((self.ref_name is not None or self.meta_ref is not None) and self.metadata_key is not None), ValueError("Required 'value' or 'ref' and 'key' attributes missing from filter")
```
mvdbeek: I think the docstring also needs updating here, right ? Now it reads as if key was required.

mvdbeek:
I've pushed a commit with a test, hope that's OK.

test/functional/tools/remove_value.xml
```diff
@@ -0,0 +1,31 @@
...
+    </outputs>
+    <tests>
+        <test>
+            <output name="out_file">
+                <assert_contents>
+                    <has_line line="hg19" />
+                </assert_contents>
+            </output>
+        </test>
+        <test expect_failure="true">
```
mvdbeek, line 28: So this test actually passes (instead of producing a failure!) and produces an output file with `hg19`. It's not a huge  deal as that can't happen in the UI, but we should probably fix this anyway.

lib/galaxy/tools/parameters/dynamic_options.py
nsoranzo: I just added a space here, but I think it is wrong anyway, it should be:
```suggestion
        assert self.value is not None or self.ref_name is not None or (self.meta_ref is not None and self.metadata_key is not None), ValueError("Required 'value', or 'ref', or 'meta_ref' and 'key' attributes missing from filter")
```

test/functional/tools/remove_value.xml
nsoranzo: I think this happens because only validators (not filters) are checked at tool execution, so it's kind of expected. Maybe comment this test out and open a separate issue, if you think that's not the correct behaviour?

lib/galaxy/tools/parameters/dynamic_options.py
mvdbeek: Yes, that looks more correct

test/functional/tools/remove_value.xml
mvdbeek: Arg, that's probably just me not knowing how to write a galaxy tool 😠. Missing the name attribute.

mvdbeek: yep

mvdbeek (approved):

jmchilton (approved):

jmchilton:
Very nice, thanks!

PR #6916, opened 2018-10-24
Do not log exception info twice

nsoranzo:
`logging.exception()` already adds the the exception information after the supplied message.

Also: the `message` attribute of `Exception` class has been dropped in Python 3, always use `str(e)`.

xref. https://github.com/galaxyproject/galaxy/issues/1715


mvdbeek (approved):

mvdbeek:
Thanks @nsoranzo !

Issue #6139, opened 2018-05-14
Job metascheduling across heterogenous clusters and resource types


natefoo:
I have mulled this over myself quite a bit and with other admins, and am throwing it out here for wider discussion. As far as I know, I'm the only person in this situation, but maybe there are others who could benefit from a solution to this problem.

Galaxy currently makes it fairly easy to schedule across multiple types of compute resources, from local clusters, remote HPC, and cloud instances. However, it lacks certain metascheduling functionality that make distributing jobs in an automated way somewhat difficult.

## The scheduling dilemma

Consider the following scenario, which currently exists on usegalaxy.org:

- Multicore jobs are currently able to run on 3 different, but partially connected clusters:
  1. A local bare metal cluster, where datasets are on a filesystem shared between the Galaxy server VMs and the compute nodes.
  2. A remote OpenStack-based cluster with Pulsar and an NFS server for the nodes running on the "head node"
  3. A largely identical copy of 2. on a different OpenStack cloud.
- All 3 clusters run Slurm and are interconnected using Slurm's multi-cluster functionality, but they do not share filesystems amongst each other
- Jobs can be directly submitted to any of the 3 clusters via Slurm on the Galaxy server VMs
- Jobs sent to cluster 1 do not require staging, jobs sent to 2 and 3 do

The desire is to distribute jobs across these 3 clusters balanced by load. Galaxy currently makes this very easy to do (with static or dynamic mappings) by tool, user, datasets, etc., but not by load of the compute resource. Ideally I could just submit these to Slurm and let it make that decision. Indeed, Slurm has the ability to do this, but in practice, for Galaxy, it is not possible: to submit on clusters 2 and 3, the inputs must be staged (by Pulsar) and the job script is written differently (by Pulsar).

In light of this problem, we currently "balance" jobs across these 3 clusters with an increasingly ugly [dynamic rule](https://github.com/galaxyproject/usegalaxy-playbook/blob/2ec2d506443f4194da4794bc7e4a13b43147f4b2/env/common/files/galaxy/dynamic_rules/multi_dynamic_walltime.py) that submits a "test job" to Slurm with `sbatch`'s `--test-only` flag to see which cluster Slurm would select, and then submitting the real job to the appropriate destination.

[...]

PR #6920, opened 2018-10-24
Allow dynamic runner rules to return None to indicate that mapping should be retried on the next iteration of the job readiness check

natefoo:
This makes it possible to do some very basic metascheduling since it can defer the decision on where to route jobs into the future (i.e. check multiple destinations and only assign one once it doesn't have waiting jobs).

The performance could be quite poor depending on what your rule does since it will run once for every ready job on every iteration. Some sort of caching in the rule would be recommended.

xref #6139

### DIFF
--- a/lib/galaxy/jobs/handler.py
+++ b/lib/galaxy/jobs/handler.py
@@ -481,6 +481,8 @@ class JobHandlerQueue(Monitors):
                 log.debug("Intentionally failing job with message (%s)" % failure_message)
             job_wrapper.fail(failure_message)
             return JOB_ERROR, job_destination
+        if job_destination is None:
+            return JOB_WAIT, None
         # job is ready to run, check limits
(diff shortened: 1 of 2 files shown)


jmchilton:
Can already throw JobNotReadyException right? Is this better?

afgane:
Yes, I think this is better because an automatic attempt will be made for the job to be rescheduled in the next scheduling cycle vs. having to deal with the exception.

I just tested this and works as described, cycling through the (re)scheduling process until a destination is available.

natefoo:
@jmchilton Totally forgot about it. Blah thanks.

@afgane Raising `JobNotReadyException` should automatically attempt rescheduling/"redestinationing" as well. Try it and see if it doesn't do the same thing as this PR. It does for me.

afgane:
Yep, raising `from galaxy.jobs.mapper import JobNotReadyException` seems to behave the same. Good team work!

PR #6924, opened 2018-10-25
Be more precise with the job runner worker thread join timeout and improve documentation

natefoo:
**Before:** Wait at most `monitor_thread_join_timeout` seconds on each work thread. Unlucky threads would only get ~5 seconds to complete their work, lucky ones would get `5 * nworkers`.

**After:** Wait at most `monitor_thread_join_timeout` seconds for *all* work threads. All threads get 30 seconds.

This makes the timing more consistent.

@erasche @Slugger70 probably relevant to your interests (but you don't have to wait for this PR, you should be able to increase `monitor_thread_join_timeout` and set the relevant supervisor/uWSGI options and see some improvement today on handler restart safety).

EDIT: [The change to lib/galaxy/jobs/handler.py](https://github.com/galaxyproject/galaxy/pull/6924/files#diff-bab4b69cf3ddd83cc0cc83f123692e0a) is probably worth backporting to 18.09 since that's a bug.

### DIFF
--- a/lib/galaxy/jobs/handler.py
+++ b/lib/galaxy/jobs/handler.py
@@ -965,5 +965,12 @@ class DefaultJobDispatcher(object):
             job_wrapper.fail(DEFAULT_JOB_PUT_FAILURE_MESSAGE)
 
     def shutdown(self):
-        for runner in self.job_runners.values():
-            runner.shutdown()
+        for name, runner in self.job_runners.items():
+            failures = []
+            try:
+                runner.shutdown()
+            except Exception:
+                failures.append(name)
+                log.exception("Failed to shutdown runner %s", name)
+            if failures:
+                raise Exception("Failed to shutdown runners: %s" % ', '.join(failures))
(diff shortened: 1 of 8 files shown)


lib/galaxy/jobs/handler.py
```diff
@@ -965,8 +965,12 @@ def recover(self, job, job_wrapper):
             job_wrapper.fail(DEFAULT_JOB_PUT_FAILURE_MESSAGE)
 
     def shutdown(self):
-        for runner in self.job_runners.values():
+        for name, runner in self.job_runners.items():
+            failures = []
```
bgruening: Was the idea to collect all names fo all job_runners before raising the Exception? If so should this not be outside the for loop?

bgruening (commented):
Thanks @natefoo! I hope this will make me less scared every time I restart the handlers :)

lib/galaxy/jobs/handler.py
natefoo: Indeed, thanks!

PR #6925, opened 2018-10-25
Add workflow step parameters to workflow editor (try 3)

jmchilton:
An alternative to #6919, which was an alternative of #6829.

Differences from #6829:

* Keep beta workflow modules config option, just move step parameters out of beta.
* Rather than making all non-data parameters that are specified as RuntimeValue parameters - have separate states for RuntimeValue parameters and ConnectedValue parameters. I think this change is pretty isolated to 35260f2 if we want to undo it, I think on balance it is the right thing to do but I could be convinced I'm wrong I suspect.
* Rather than defining "parameters" and "inputs" separately like in #6829, this brings in work from the CWL branch (standalone as #6911) to generalize inputs/outputs in the workflow module and API layer to non-data inputs. I like the consistency of this interface better and it starts us on the road to allow for complex field parameters in the CWL branch.

Differences from #6919:

* Defines some simple workflow editor test cases.
* ~Brings in workflow API test definition improvements #6776 to define some new tests.~ (already merged into dev and this has been rebased)
* Fixes handling of manually specified workflow definitions with non-data connections that have a connection specified but not a ConnectedValue in the tool state.

### DIFF
--- a/lib/galaxy/workflow/run_request.py
+++ b/lib/galaxy/workflow/run_request.py
@@ -265,6 +265,17 @@ def build_workflow_run_configs(trans, workflow, payload):
             # inputs with referential integrity if parameters are already normalized (coming from tool form).
             normalized_inputs = {}
 
+        if param_map:
+            # disentangle raw parameter dictionaries into formal request structures if we can
+            # to setup proper WorkflowRequestToInputDatasetAssociation, WorkflowRequestToInputDatasetCollectionAssociation
+            # and WorkflowRequestInputStepParameter objects.
+            for step in workflow.steps:
+                normalized_key = step.id
+                if step.type == "parameter_input":
+                    if normalized_key in param_map:
+                        value = param_map.pop(normalized_key)
+                        normalized_inputs[normalized_key] = value["input"]
+
         steps_by_id = workflow.steps_by_id
         # Set workflow inputs.
         for key, input_dict in normalized_inputs.items():
--- a/client/galaxy/scripts/mvc/form/form-data.js
+++ b/client/galaxy/scripts/mvc/form/form-data.js
@@ -84,7 +84,9 @@ export var Manager = Backbone.Model.extend({
                             if (field && field.value) {
                                 value = field.value();
                                 if (input.ignore === undefined || input.ignore != value) {
-                                    if (field.collapsed && input.collapsible_value) {
+                                    if (field.collapsed && field.connected) {
+                                        value = { __class__: "ConnectedValue" };
+                                    } else if (field.collapsed && input.collapsible_value) {
                                         value = input.collapsible_value;
                                     }
                                     add(flat_id, input.id, value);
--- a/client/galaxy/scripts/mvc/form/form-input.js
+++ b/client/galaxy/scripts/mvc/form/form-input.js
@@ -25,6 +25,7 @@ export default Backbone.View.extend({
         this.$collapsible = this.$(".ui-form-collapsible");
         this.$collapsible_text = this.$(".ui-form-collapsible-text");
         this.$collapsible_icon = this.$(".ui-form-collapsible-icon");
+        this.$connected_icon = this.$(".ui-form-connected-icon");
         this.$title = this.$(".ui-form-title");
         this.$title_text = this.$(".ui-form-title-text");
         this.$error_text = this.$(".ui-form-error-text");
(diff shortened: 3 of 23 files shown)


lib/galaxy/workflow/run_request.py
```diff
@@ -265,6 +265,17 @@ def build_workflow_run_configs(trans, workflow, payload):
             # inputs with referential integrity if parameters are already normalized (coming from tool form).
             normalized_inputs = {}
 
+        if param_map:
+            # disentangle raw parameter dictionaries into formal request structures if we can
+            # to setup proper WorkflowRequestToInputDatasetAssociation, WorkflowRequestToInputDatasetCollectionAssociation
+            # and WorkflowRequestInputStepParameter objects.
+            for step in workflow.steps:
+                normalized_key = step.id
+                if step.type == "parameter_input":
```
mvdbeek, line 274: This still trips up when there are `data_input` parameters in the workflow:
```
galaxy.workflow.run ERROR 2018-11-05 17:28:53,706 Failed to schedule Workflow[id=287,name=NonDataInputModule], problem occurred on WorkflowStep[index=1,type=data_input].
  File "/Users/mvandenb/src/galaxy/lib/galaxy/workflow/run.py", line 190, in invoke
    incomplete_or_none = self._invoke_step(workflow_invocation_step)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/workflow/run.py", line 266, in _invoke_step
    use_cached_job=self.workflow_invocation.use_cached_job)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/workflow/modules.py", line 535, in execute
    progress.set_outputs_for_input(invocation_step, step_outputs)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/workflow/run.py", line 393, in set_outputs_for_input
    raise ValueError(message)
ValueError: Step with id 1008 not found in inputs_step_id ({1007: 'new label'})
galaxy.workflow.run ERROR 2018-11-05 17:28:53,708 Failed to execute scheduled workflow.
  File "/Users/mvandenb/src/galaxy/lib/galaxy/workflow/run.py", line 83, in __invoke
    outputs = invoker.invoke()
  File "/Users/mvandenb/src/galaxy/lib/galaxy/workflow/run.py", line 190, in invoke
    incomplete_or_none = self._invoke_step(workflow_invocation_step)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/workflow/run.py", line 266, in _invoke_step
    use_cached_job=self.workflow_invocation.use_cached_job)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/workflow/modules.py", line 535, in execute
    progress.set_outputs_for_input(invocation_step, step_outputs)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/workflow/run.py", line 393, in set_outputs_for_input
    raise ValueError(message)
ValueError: Step with id 1008 not found in inputs_step_id ({1007: 'new label'})
```

mvdbeek:
```
diff --git a/lib/galaxy/workflow/run_request.py b/lib/galaxy/workflow/run_request.py
index ff49501943..1dd955dd9d 100644
--- a/lib/galaxy/workflow/run_request.py
+++ b/lib/galaxy/workflow/run_request.py
@@ -271,10 +271,13 @@ def build_workflow_run_configs(trans, workflow, payload):
             # and WorkflowRequestInputStepParameter objects.
             for step in workflow.steps:
                 normalized_key = step.id
-                if step.type == "parameter_input":
+                if step.type in ["parameter_input", 'data_input']:
                     if normalized_key in param_map:
                         value = param_map.pop(normalized_key)
-                        normalized_inputs[normalized_key] = value["input"]
+                        if step.type == 'data_input':
+                            normalized_inputs[normalized_key] = value["input"]['values'][0]
+                            normalized_inputs[normalized_key] = value["input"]

         steps_by_id = workflow.steps_by_id
         # Set workflow inputs.
```

this let's me run the workflow -- but I'm guessing this will fall flat in many scenarios :/

mvdbeek:
@jmchilton with this https://github.com/jmchilton/galaxy/pull/70 I have successfully run a couple of workflows with input parameters, I'm not sure if that's the right way to go.

lib/galaxy/workflow/run_request.py
jmchilton: But the tests were passing and we weren't touching those inputs at all in this loop before, I don't get it. Would it only fail on workflows that had both data and parameter inputs?

jmchilton: I'll write some tests and try to verify this is needed.

jmchilton: Added a test case, I can verify you aren't crazy.

jmchilton: Okay - I think the problem is here (https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/workflow/run.py#L393). inputs_by_id is now populated by only partially, but if outputs is supplied and has an 'output' value we can skip the if for that step. I can't make the change tonight, but I'll do it in the morning. I think this is a lighter touch than the PR - is that okay?

mvdbeek: Yeah, that's more than reasonable, I did play around this but didn't realize we could just skip if the output is there ... which makes a lot of sense.

mvdbeek:
Alright, this runs now, but there are some connectivity issues in the workflow editor. For instance if you toggle runtime inputs connections are not destroyed, and occasionally the input terminal either doesn't show up or isn't being destroyed properly. I guess this is probably one and the same issue, I'll have a look if I can figure this out.

mvdbeek:
Ah, that's actually working fine. I just didn't fully realise that we have separate buttons for setting a Parameter at runtime and connecting a Parameter, sorry about that. Maybe we should have a text that explains this when you hover over the buttons ?

mvdbeek (approved):
Fantastic stuff, thanks a lot @jmchilton !

mvdbeek:
@galaxybot test py3

mvdbeek:
Test failure is unrelated, I'll open a fix later.

guerler:
This is pretty cool. Thanks for the contribution @jmchilton. Sorry for the late review but I noticed a few minor issues. Clicking on a tool parameter label in the run workflow form does not collapse the input parameters anymore although the 'hand' cursor is shown. Another minor issue is that the collapsible and connected icons in the workfow editor are very close to each other and share the same tooltip which makes it difficult to select the correct one.

PR #6927, opened 2018-10-26
Allocate more memory for selenium Python 3 client.

jmchilton:
Matching Python 2 fix in 4054ffc31165b533cb292c58edafac49b59de9de.


mvdbeek:
@galaxybot test py3

jmchilton:
This was opened pre #6928, so the Travis tests will fail but that is fine.

mvdbeek:
Awesome, all green.

PR #6929, opened 2018-10-26
Small testing tweaks from the CWL branch.

jmchilton:


mvdbeek (approved):

test/base/populators.py
```diff
@@ -1187,7 +1188,7 @@ class GiPostGetMixin(object):
     """Mixin for adapting Galaxy testing populators helpers to bioblend."""
 
     def _get(self, route, data={}):
```
mvdbeek, line 1190: Do you want also drop the mutable default ?

jmchilton: I should indeed, I'll merge this and open a follow up.

PR #6931, opened 2018-10-27
tests need to count data sets and collections

bernt-matthias:
Otherwise `expect_num_outputs` seems not to work as expected IMHO.

https://github.com/galaxyproject/tools-iuc/pull/2127 is an example that does not work without this change.

### DIFF
--- a/lib/galaxy/tools/verify/interactor.py
+++ b/lib/galaxy/tools/verify/interactor.py
@@ -748,7 +748,7 @@ def _verify_outputs(testdef, history, jobs, tool_id, data_list, data_collection_
     maxseconds = testdef.maxseconds
     if testdef.num_outputs is not None:
         expected = testdef.num_outputs
-        actual = len(data_list)
+        actual = len(data_list) + len(data_collection_list)
         if expected != actual:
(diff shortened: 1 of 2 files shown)


jmchilton (approved):

martenson:
could use a squash/rebase

bernt-matthias:
> could use a squash/rebase

@martenson wonderful idea. but as you see I failed horribly :( still struggling with git and could need some help here.

mvdbeek:
I think in your case it would be easiest to just cherry-pick b22ba0f and 0198164 into a clean new branch. In practive I'd do `git checkout -b clean_new_branch origin/dev` (assuming origin is galaxyproject), `git cherry-pick b22ba0f 0198164`, push and make sure all is good. Then delete the local branch `git branch -D bernt-matthias:topic/test-colouts` and rename `clean_new_branch` to `bernt-matthias:topic/test-colouts` (`git checkout -b bernt-matthias:topic/test-colouts`).

Then you just need to force-push and the PR will only contain the commits in the cleaned-up branch.

bernt-matthias:
Thanks @mvdbeek.

martenson (approved):
thanks @bernt-matthias !

PR #6937, opened 2018-10-29
Default shed tool installation to database/shed_tools

natefoo:
...instead of `../shed_tools`.

The old path was historical and due to tool repositories being Mercurial repositories, which could not be contained within another Mercurial repository (Galaxy itself, at the time).

This does not affect any existing installations since `shed_tool_conf.xml` is only initialized once.


bernt-matthias:
Just stumbled also over this issue. Wondering in how far this interferes with shed_tool_data_path in galaxy.yml (which defaults to `tool-data`).

Issue #6936, opened 2018-10-29
Publication of Galaxy Release v 19.01


dannon:
- [X] **Prep**

    - [X] ~~Create this release issue ``make release-issue RELEASE_CURR=19.01``.~~
    - [X] ~~Set freeze date (2019-01-07).~~

- [x] **Branch Release (on or around 2019-01-07)**

    - [x] Ensure all [blocking milestone PRs](https://github.com/galaxyproject/galaxy/pulls?q=is%3Aopen+is%3Apr+milestone%3A19.01) have been merged, delayed, or closed.

          make release-check-blocking-prs RELEASE_CURR=19.01
    - [x] Merge the latest release into dev and push upstream.

          make release-merge-stable-to-next RELEASE_PREVIOUS=release_18.09
          make release-push-dev

    - [x] Create and push release branch:

          make release-create-rc RELEASE_CURR=19.01 RELEASE_NEXT=19.05

    - [x] Open PRs from your fork of branch ``version-19.01`` to upstream ``release_19.01`` and of ``version-19.05.dev`` to ``dev``.
    - [x] Update ``next_milestone`` in [P4's configuration](https://github.com/galaxyproject/p4) to `19.05` so it properly tags new PRs.
    - [x] Set the ``release_19.01`` branch in GitHub [settings](https://github.com/galaxyproject/galaxy/settings/branches) as protected.

- [x] **Issue Review Timeline Notes**
    - [x] Ensure any security fixes will be ready prior to 2019-01-07 + 1 week, to allow time for notification prior to release.
    - [x] Ensure ownership of outstanding bugfixes and track progress during freeze.

- [x] **Deploy and Test Release**

    - [x] Update test.galaxyproject.org to ensure it is running a dev at or past branch point (2019-01-07 + 1 day).
    - [x] Update testtoolshed.g2.bx.psu.edu to ensure it is running a dev at or past branch point (2019-01-07 + 1 day).
    - [x] Deploy to usegalaxy.org (2019-01-07 + 1 week).
    - [x] Deploy to toolshed.g2.bx.psu.edu (2019-01-07 + 1 week).
    - [x] [Update BioBlend CI testing](https://github.com/galaxyproject/bioblend/commit/b74b1c302a1b8fed86786b40d7ecc3520cbadcd3) to include a ``release_19.01`` target: add ``- TOX_ENV=py27 GALAXY_VERSION=release_19.01`` to the ``env`` list in ``.travis.yml`` .

[...]

PR #6942, opened 2018-10-30
added global monitor for window.Galaxy

Nerdinacan:
Still waiting on another PR before I can release most of what I've been working on, but this is a little diagnostic tool I've been using that can exist without webpack 4. 

It's a simple proxy monitor that watches window variables. I've set it up to watch window.Galaxy, some of the other horrible globals defined in onload.js. Actual functionality of the site should be unchanged. All this does is spam the console when the appropriate messages are toggled-on.

I've also used webpack's Provide plugin to define "Galaxy" as the resulting proxy object in all the webpack bundled scripts. (See webpack config, Provide plugin). This means it effectively watches all changes to window.Galaxy or Galaxy from the legacy scripts, and is intended to be used as a diagnostic tool while we refactor.

**To get a list of currently tracked window properties:**
showMonitoredProperties()

**To toggle the console messages on or off:**
toggleMonitorMessages("Galaxy", true|false)

**To watch a new window variable:**
import { installMonitor } from "utils/installMonitor";
installMonitor("SomeGlobalVariable", [optionalFallbackValue])

Then in the chrome/firefox/whatever console:
toggleMonitorMessage("SomeGlobalVariable", true)

And behold the spam.


Nerdinacan:
Hmmm, looks like the karma configs are screwy. Will update those too.

test/galaxy_selenium/navigates_galaxy.py
```diff
@@ -141,6 +141,9 @@ def wait_length(self, wait_type):
 
     def sleep_for(self, wait_type):
         time.sleep(self.wait_length(wait_type))
+    
+    def sleep_for_exactly(self, duration):
+        time.sleep(duration)
```
jmchilton: Intentional? Seems unused. I'm fine if you are using it downstream though, but do you want to change the definition of sleep_for to use sleep_for_exactly - removes some duplication and then the entry point is being used in the current code also?

```python
    def sleep_for(self, wait_type):
        self.sleep_for_exactly(self.wait_length(wait_type))
```

I'd also prefer the name ``sleep_for_duration`` or ``sleep_for_seconds`` but the current name is fine if you'd rather not change it.

Nerdinacan: Ahh I use it periodically during busted test diagnosis. I can rename it if you like. sleep_for_seconds sounds fine to me.

Nerdinacan:
This is happy now that the unit tests can deal with a javascript Proxy. Merge whenever you like.

dannon:
@Nerdinacan Awesome, thanks, let me check it out now.

dannon:
@Nerdinacan Looks like we crossed updates here.  I'd merged locally without checking back in the PR window, so I didn't see your last merge from dev here.  It was just a merge with no conflicts, and no rebase, right?

(looks like it, I have an idea for a fix that'll mark this as merged -- will ping offline)

dannon:
Well.  Github didn't update this to reflect what we wanted, but it's all merged and in there.  The only bug is that #6936 should make sure to include this in the release notes.

PR #6943, opened 2018-10-30
Collection builder: add option to strip url prefix from pair names

shiltemann:
This adds the option to remove a url prefix from pair names when building a list of pairs collection, similar to the option to remove the file extension. 

This streamlines things a lot in [one of my training manuals](https://galaxyproject.github.io/training-material/topics/metagenomics/tutorials/mothur-miseq-sop/tutorial.html), where participants upload data from Zenodo, and create a collection. Since the downstream tools use the pair names as sample names, it was necessary for participants to manually edit every pair name, but now this should happen automatically.

e.g. before:

galaxy_strip_urlprefix_before

after:
galaxy_strip_urlprefix_after

ping @jmchilton

### DIFF
--- a/client/galaxy/scripts/mvc/collection/list-of-pairs-collection-creator.js
+++ b/client/galaxy/scripts/mvc/collection/list-of-pairs-collection-creator.js
@@ -256,6 +256,9 @@ var PairedCollectionCreator = Backbone.View.extend(baseMVC.LoggableMixin)
             this.removeExtensions = true;
             //this.removeExtensions = false;
 
+             /** remove file extensions (\.*) from created pair names? */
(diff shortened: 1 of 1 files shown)


jmchilton:
Thanks so much for the contribution this makes a lot of sense. Do we want to just do this without an option - this should be the default behavior IMO? I'd say the pair name should never contain a /. (I'm a little mortified it is even valid but I can't find any code that would prevent it 😢.) Unless there are objections, can you simplify the PR to just do this unconditionally?

shiltemann:
@jmchilton no objections from my side at least, definitely can do :)

jmchilton:
Wonderful, thanks for the contribution @shiltemann !

PR #6948, opened 2018-10-30
Use recently introduced overflow utility instead of overwriting card defaults

guerler:
This PR fixes admin related views e.g. Display Applications and the Whitelist by using the recently introduced BS4 overflow utility instead of overwriting card-style defaults. This is less invasive, restores the default and prevents the cropping of border shadows of elements which are placed close to the card body borders.

### DIFF
--- a/client/galaxy/style/scss/ui.scss
+++ b/client/galaxy/style/scss/ui.scss
@@ -24,18 +24,17 @@
     transform: rotate(45deg);
 }
 
+// utility class from: https://github.com/twbs/bootstrap/pull/27268
+.overflow-auto {
+    overflow: auto !important;
+}
+
 // default margins
 $ui-margin-vertical: $margin-v * 0.25;
 $ui-margin-vertical-large: $margin-v * 0.5;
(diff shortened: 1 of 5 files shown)


guerler:
There were no changes. I just resolved some merge conflicts.

client/galaxy/style/scss/ui.scss
```diff
@@ -24,19 +24,18 @@
     transform: rotate(45deg);
 }
 
+// utility class from: https://github.com/twbs/bootstrap/pull/27268
+.overflow-auto {
```
martenson, line 28: do we have to redefine the class given it was implemented in the PR mentioned above? https://github.com/twbs/bootstrap/pull/27268/files#diff-afd6c72eb64ef2c8a2d963958d6991d1R4

martenson: ah, it is unreleased yet, so I assume the plan is to remove it when they release

guerler: Its not in our version yet. It will be available with the next BS4 revision.

PR #6951, opened 2018-10-31
[18.09] Backport: tests need to count data sets and collections

bernt-matthias:
Would like to backport this since IUC uses 18.09 in their CI tests.

### DIFF
--- a/test/api/test_workflows.py
+++ b/test/api/test_workflows.py
@@ -2200,7 +2200,7 @@ steps:
     state:
       produce_out_1: False
       filter_text_1: '1'
-      produce_collection: True
+      produce_collection: False
 test_data: {}
     """, history_id=history_id, wait=False)
             self.wait_for_invocation_and_jobs(history_id, run_object.workflow_id, run_object.invocation_id)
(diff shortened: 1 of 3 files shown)


mvdbeek:
https://jenkins.galaxyproject.org/job/docker-api/13595/testReport/junit/test.api.test_workflows/WorkflowsApiTestCase/test_optional_workflow_output/ looks like we need to update the state in https://github.com/galaxyproject/galaxy/blob/release_18.09/test/api/test_workflows.py#L2202. That might be handled automatically on dev now, which would explain why the tests passed there but not against the release branch.

bernt-matthias:
My guess is that the new parameter of the tool needs to be specified. I tried this with my latest commit.

test/api/test_workflows.py
```diff
@@ -2200,6 +2200,7 @@ def test_optional_workflow_output(self):
     state:
       produce_out_1: False
       filter_text_1: '1'
+      produce_collection: True
```
mvdbeek: I think you need to set this to `False` here. This is the default, but if `state` is provided in gxformat2 workflows missing parameters aren't updated on 18.09.

mvdbeek (approved):
Thanks @bernt-matthias !

PR #6952, opened 2018-10-31
Quick 'make all datasets private' implementation.  

dannon:
Takes a single history or operates on all of a user's histories, making the datasets within that the user has 'management' permissions of private.  Intentionally doesn't use an API, rather a simple webcontroller endpoint, because I think we'll want to handle this with group operations and the like down the road, and will enhance existing APIs.

@natefoo You'd mentioned library datasets, and I was thinking about collections, too, but I didn't see at first glance how those were shared/exposed via RBAC, so if someone with more knowledge wanted to take a poke, that'd be great.

### DIFF
--- a/client/galaxy/scripts/apps/analysis.js
+++ b/client/galaxy/scripts/apps/analysis.js
@@ -150,10 +150,9 @@ window.app = function app(options, bootstrapped) {
                 new GridView({
                     url_base: `${Galaxy.root}workflow/list_published`,
                     active_tab: "shared",
-                    url_data:
-                        {
-                            'f-username': ( userFilter == null ) ? "" : userFilter
-                        }
+                    url_data: {
+                        "f-username": userFilter == null ? "" : userFilter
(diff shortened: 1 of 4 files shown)


martenson:
It is the beginning of release dev cycle, I think we should aspire to enhance the APIs that we want to enhance, instead of implementing custom controller non-API that behaves like one and that we intend to replace asap.

dannon:
WIP for a sec to address library dataset concern.

dannon:
@natefoo Sorry to ping yet again, but I addressed the library dataset concern.  We should be good to go now.

natefoo (approved):

natefoo:
The changes to base.css seem unintentional?

dannon:
@natefoo Yeah, let me amend that.

Okay, done.  (commit -a -m ...)

natefoo:
Thanks @dannon !

PR #6953, opened 2018-11-01
Update shed_tool_conf.xml.sample

bernt-matthias:
With the current default tools are installed to:

`GALAXY_ROOT/../shed_tools/`

I'm not sure in how far this interferes with `shed_tool_data_path` in galaxy.yml (which defaults to `tool-data`).

Also, I guess the same should be applied to `migrated_tools_conf.xml`


nsoranzo:
Duplicate of https://github.com/galaxyproject/galaxy/pull/6937 ?

bernt-matthias:
Indeed.

PR #6955, opened 2018-11-01
[18.09] Fix for workflow parameter collection in workflows view.

### LINKED ISSUE #6918: Workflow Editor Stops working when post_job_actions is set
I have isolated this issue to workflow step 21, specifically, the "post_job_actions" in that step.  I was editing the cut tool outfile attributes, specifying which column was the chromosome, start, stop, strand, and name and the editor hung when I tried to save it.  After that, the workflow would not load.  I downloaded the workflow and on a hunch, removed the "post_job_actions" stuff from step 21 (cut tool) and the workflow loaded fine.  I noted that there was only 1 setting in the post_job_actions section to set the chromosome column.  So that must be where the first error occurred.

My intent at the time was to be able to connect the output of cut to the input of closestBed.  I *should* have changed the output type, but being somewhat inexperienced, I first came across the setting of these columns, thinking that would allow me to connect the 2 tools...

Here's a gist of the workflow that does not load:

[galaxy-workflow-chip-seq_single_end_-_step_2__peaks_and_motifs-ga](https://gist.github.com/hepcat72/f1d9765d6a543666f211ede27592f977#file-galaxy-workflow-chip-seq_single_end_-_step_2__peaks_and_motifs-ga)

I have recovered my workflow and fixed the issue by manually editing the json of the workflow, so I'm not waiting on a fix.  Thus, this is a low priority issue.  I just wanted to report it so that you could correct the issue for anyone else that might run into it.

dannon:
This fixes #6918.

I didn't track down when exactly this broke, but the problem was that the internal representation of columns in the data is now (?) correctly (?) an integer.  Anyway, we just skip matching for params against those.

### DIFF
--- a/client/galaxy/scripts/mvc/workflow/workflow-view.js
+++ b/client/galaxy/scripts/mvc/workflow/workflow-view.js
@@ -734,9 +734,11 @@ export default Backbone.View.extend({
                 $.each(node.post_job_actions, (k, pja) => {
                     if (pja.action_arguments) {
                         $.each(pja.action_arguments, (k, action_argument) => {
-                            var arg_matches = action_argument.match(parameter_re);
-                            if (arg_matches) {
-                                matches = matches.concat(arg_matches);
+                            if (typeof action_argument === "string") {
+                                var arg_matches = action_argument.match(parameter_re);
+                                if (arg_matches) {
+                                    matches = matches.concat(arg_matches);
+                                }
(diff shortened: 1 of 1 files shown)


jmchilton (approved):

bgruening (approved):

bgruening:
Thanks @dannon

PR #6958, opened 2018-11-01
Kubernetes job runner integration test and enhancements

mvdbeek:
- pykube is now a conditional requirement and will be installed if a kubernetes runner is set up
- Submit ajs.job_file instead of tool_script.sh, which allows using standard Galaxy features, like setting env variables per destination, reading stdout/stderr from file, collecting job measurements
- Fixes state changes
- Allows selecting the shell to use for all job runners
- Python 3 fixes for container resolver
- Writes stdout and stderr to standard file locations
- Passes GALAXY_SLOTS and GALAXY_MEMORY_MB and GALAXY_MEMORY_MB_PER_SLOT to wrappers
- Includes integration tests for running jobs, cancelling jobs, executing jobs in mulled containers, detecting exit code, stdout, stderr and passing GALAXY_SLOTS and GALAXY_MEMORY_MB.

### DIFF
--- a/lib/galaxy/jobs/command_factory.py
+++ b/lib/galaxy/jobs/command_factory.py
@@ -15,6 +15,7 @@ from galaxy.jobs.runners.util.job_script import (
 log = getLogger(__name__)
 
 CAPTURE_RETURN_CODE = "return_code=$?"
+YIELD_CAPTURED_CODE = 'sh -c "exit $return_code"'
 SETUP_GALAXY_FOR_METADATA = """
 [ "$GALAXY_VIRTUAL_ENV" = "None" ] && GALAXY_VIRTUAL_ENV="$_GALAXY_VIRTUAL_ENV"; _galaxy_setup_environment True
 """
@@ -268,6 +269,8 @@ tee -a stderr.log < "$err" >&2 &""",
             self.append_command(CAPTURE_RETURN_CODE)
 
     def build(self):
+        if self.return_code_captured:
+            self.append_command(YIELD_CAPTURED_CODE)
         return self.commands
 
 
--- a/lib/galaxy/jobs/command_factory.py
+++ b/lib/galaxy/jobs/command_factory.py
@@ -30,6 +30,8 @@ def build_command(
     create_tool_working_directory=True,
     remote_command_params={},
     metadata_directory=None,
+    stdout_file=None,
+    stderr_file=None,
 ):
     """
     Compose the sequence of commands necessary to execute a job. This will
@@ -103,6 +105,8 @@ def build_command(
     if include_work_dir_outputs:
         __handle_work_dir_outputs(commands_builder, job_wrapper, runner, remote_command_params)
 
+    if stdout_file and stderr_file:
+        commands_builder.capture_stdout_stderr(stdout_file, stderr_file)
     commands_builder.capture_return_code()
 
     if include_metadata and job_wrapper.requires_setting_metadata:
@@ -227,23 +231,36 @@ class CommandsBuilder(object):
         # the last thing that happens is an exit with return code.
         self.return_code_captured = False
 
-    def prepend_command(self, command):
+    def prepend_command(self, command, sep=";"):
         if command:
-            self.commands = u"%s; %s" % (command,
+            self.commands = u"%s%s %s" % (command,
+                                         sep,
                                          self.commands)
         return self
 
     def prepend_commands(self, commands):
         return self.prepend_command(u"; ".join(c for c in commands if c))
 
-    def append_command(self, command):
+    def append_command(self, command, sep=';'):
         if command:
-            self.commands = u"%s; %s" % (self.commands,
-                                         command)
+            self.commands = u"%s%s %s" % (self.commands,
+                                          sep,
+                                          command)
         return self
 
     def append_commands(self, commands):
         self.append_command(u"; ".join(c for c in commands if c))
+
+    def capture_stdout_stderr(self, stdout_file, stderr_file):
+        self.prepend_command("""out="${TMPDIR:-/tmp}/out.$$" err="${TMPDIR:-/tmp}/err.$$"
+mkfifo "$out" "$err"
+trap 'rm "$out" "$err"' EXIT
+tee -a stdout.log < "$out" &
+tee -a stderr.log < "$err" >&2 &""",
+                             sep="")
+        self.append_command("> '{stdout_file}' 2> '{stderr_file}'".format(stdout_file=stdout_file,
+                                                                          stderr_file=stderr_file),
+                            sep="")
 
     def capture_return_code(self):
         if not self.return_code_captured:
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -61,7 +61,7 @@ class KubernetesJobRunner(AsynchronousJobRunner):
             k8s_default_requests_memory=dict(map=str, default=None),
             k8s_default_limits_cpu=dict(map=str, default=None),
             k8s_default_limits_memory=dict(map=str, default=None),
-            k8s_pod_retrials=dict(map=int, valid=lambda x: int > 0, default=3))
+            k8s_pod_retrials=dict(map=int, valid=lambda x: int >= 0, default=3))
 
         if 'runner_param_specs' not in kwargs:
             kwargs['runner_param_specs'] = dict()
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -263,14 +263,6 @@ class KubernetesJobRunner(AsynchronousJobRunner):
         k8s_container = {
             "name": self.__get_k8s_container_name(ajs.job_wrapper),
             "image": self._find_container(ajs.job_wrapper).container_id,
-            "env": [{
-                "name": "GALAXY_VIRTUAL_ENV",
-                "value": "None"
-            }, {
-                "name": "GALAXY_LIB",
-                "value": "None"
-            }
-            ],
             # this form of command overrides the entrypoint and allows multi command
             # command line execution, separated by ;, which is what Galaxy does
             # to assemble the command.
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -151,10 +151,6 @@ class KubernetesJobRunner(AsynchronousJobRunner):
         ajs.job_id = k8s_job_name
         self.monitor_queue.put(ajs)
 
-        # external_runJob_script can be None, in which case it's not used.
-        external_runjob_script = None
-        return external_runjob_script
-
     def __get_pull_policy(self):
         if "k8s_pull_policy" in self.runner_params:
             if self.runner_params['k8s_pull_policy'] in ["Always", "IfNotPresent", "Never"]:
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -96,7 +96,18 @@ class KubernetesJobRunner(AsynchronousJobRunner):
         if not self.prepare_job(job_wrapper, include_metadata=False, modify_command_for_container=False):
             return
 
-        job_destination = job_wrapper.job_destination
+        ajs = AsynchronousJobState(files_dir=job_wrapper.working_directory,
+                                   job_wrapper=job_wrapper,
+                                   job_destination=job_wrapper.job_destination)
+
+        # fill in the DRM's job run template
+        script = self.get_job_file(job_wrapper, exit_code_path=ajs.exit_code_file, shell=job_wrapper.shell)
+        try:
+            self.write_executable_script(ajs.job_file, script)
+        except Exception:
+            job_wrapper.fail("failure preparing job script", exception=True)
+            log.exception("(%s) failure writing job script" % job_wrapper.get_id_tag())
+            return
 
         # Construction of the Kubernetes Job object follows: http://kubernetes.io/docs/user-guide/persistent-volumes/
         k8s_job_name = self.__produce_unique_k8s_job_name(job_wrapper.get_id_tag())
@@ -110,7 +121,7 @@ class KubernetesJobRunner(AsynchronousJobRunner):
                     "namespace": self.runner_params['k8s_namespace'],
                     "labels": {"app": k8s_job_name}
             },
-            "spec": self.__get_k8s_job_spec(job_wrapper)
+            "spec": self.__get_k8s_job_spec(ajs)
         }
 
         # Checks if job exists and is trusted, or if it needs re-creation.
@@ -140,8 +151,7 @@ class KubernetesJobRunner(AsynchronousJobRunner):
             job.create()
 
         # define job attributes in the AsyncronousJobState for follow-up
-        ajs = AsynchronousJobState(files_dir=job_wrapper.working_directory, job_wrapper=job_wrapper,
-                                   job_id=k8s_job_name, job_destination=job_destination)
+        ajs.job_id = k8s_job_name
         self.monitor_queue.put(ajs)
 
         # external_runJob_script can be None, in which case it's not used.
@@ -202,23 +212,23 @@ class KubernetesJobRunner(AsynchronousJobRunner):
             instance_id = self._galaxy_instance_id + "-"
         return "galaxy-" + instance_id + galaxy_internal_job_id
 
-    def __get_k8s_job_spec(self, job_wrapper):
+    def __get_k8s_job_spec(self, ajs):
         """Creates the k8s Job spec. For a Job spec, the only requirement is to have a .spec.template."""
-        k8s_job_spec = {"template": self.__get_k8s_job_spec_template(job_wrapper)}
+        k8s_job_spec = {"template": self.__get_k8s_job_spec_template(ajs)}
         return k8s_job_spec
 
-    def __get_k8s_job_spec_template(self, job_wrapper):
+    def __get_k8s_job_spec_template(self, ajs):
         """The k8s spec template is nothing but a Pod spec, except that it is nested and does not have an apiversion
         nor kind. In addition to required fields for a Pod, a pod template in a job must specify appropriate labels
         (see pod selector) and an appropriate restart policy."""
         k8s_spec_template = {
             "metadata": {
-                "labels": {"app": self.__produce_unique_k8s_job_name(job_wrapper.get_id_tag())}
+                "labels": {"app": self.__produce_unique_k8s_job_name(ajs.job_wrapper.get_id_tag())}
             },
             "spec": {
-                "volumes": self.__get_k8s_mountable_volumes(job_wrapper),
-                "restartPolicy": self.__get_k8s_restart_policy(job_wrapper),
-                "containers": self.__get_k8s_containers(job_wrapper)
+                "volumes": self.__get_k8s_mountable_volumes(ajs.job_wrapper),
+                "restartPolicy": self.__get_k8s_restart_policy(ajs.job_wrapper),
+                "containers": self.__get_k8s_containers(ajs)
             }
         }
         # TODO include other relevant elements that people might want to use from
@@ -252,26 +262,26 @@ class KubernetesJobRunner(AsynchronousJobRunner):
         }
         return [k8s_mountable_volume]
 
-    def __get_k8s_containers(self, job_wrapper):
+    def __get_k8s_containers(self, ajs):
         """Fills in all required for setting up the docker containers to be used, including setting a pull policy if
            this has been set.
         """
         k8s_container = {
-            "name": self.__get_k8s_container_name(job_wrapper),
-            "image": self._find_container(job_wrapper).container_id,
+            "name": self.__get_k8s_container_name(ajs.job_wrapper),
+            "image": self._find_container(ajs.job_wrapper).container_id,
             # this form of command overrides the entrypoint and allows multi command
             # command line execution, separated by ;, which is what Galaxy does
             # to assemble the command.
-            # TODO possibly shell needs to be set by job_wrapper
-            "command": ["/bin/bash", "-c", job_wrapper.runner_command_line],
-            "workingDir": job_wrapper.working_directory,
+            "command": [ajs.job_wrapper.shell],
+            "args": ["-c", ajs.job_file],
+            "workingDir": ajs.job_wrapper.working_directory,
             "volumeMounts": [{
                 "mountPath": self.runner_params['k8s_persistent_volume_claim_mount_path'],
                 "name": self._galaxy_vol_name
             }]
         }
 
-        resources = self.__get_resources(job_wrapper)
+        resources = self.__get_resources(ajs.job_wrapper)
         if resources:
             k8s_container['resources'] = resources
 
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -3,6 +3,7 @@ Offload jobs to a Kubernetes cluster.
 """
 
 import logging
+import math
 import os
 import re
 from time import sleep
@@ -286,7 +287,7 @@ class KubernetesJobRunner(AsynchronousJobRunner):
                 if 'memory' in limits:
                     envs.append({'name': 'GALAXY_MEMORY_MB', 'value': str(ByteSize(limits['memory']).to_unit('M', as_string=False))})
                 if 'cpu' in limits:
-                    envs.append({'name': 'GALAXY_SLOTS', 'value': limits['cpu']})
+                    envs.append({'name': 'GALAXY_SLOTS', 'value': str(int(math.ceil(float(limits['cpu']))))})
             k8s_container['resources'] = resources
             k8s_container['env'] = envs
 
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -446,9 +446,14 @@ class KubernetesJobRunner(AsynchronousJobRunner):
                 return job_state
             elif failed > max_pod_retrials:
                 return self._handle_job_failure(job, job_state)
-            # We should not get here
-            log.error("Kubernetes job '%s' not classified as succ., active or failed. Full Job object: \n%s", job.name, job.obj)
-            return job_state
+            elif job_state.job_wrapper.get_job().state == model.Job.states.DELETED:
+                # Job has been deleted via stop_job, cleanup and remove from watched_jobs by returning `None`
+                if job_state.job_wrapper.cleanup_job in ("always", "onsuccess"):
+                    job_state.job_wrapper.cleanup()
+                return None
+            else:
+                # We really shouldn't reach this point, but we might if the job has been killed by the kubernetes admin
+                log.info("Kubernetes job '%s' not classified as succ., active or failed. Full Job object: \n%s", job.name, job.obj)
 
         elif len(jobs.response['items']) == 0:
             # there is no job responding to this job_id, it is either lost or something happened.
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -5,7 +5,6 @@ Offload jobs to a Kubernetes cluster.
 import logging
 import os
 import re
-import string
 from time import sleep
 
 from galaxy import model
@@ -495,50 +494,6 @@ class KubernetesJobRunner(AsynchronousJobRunner):
             return True
 
         return False
-
-    def fail_job(self, job_state):
-        """
-        Kubernetes runner overrides fail_job (called by mark_as_failed) to rescue the pod's log files which are left as
-        stdout (pods logs are the natural stdout and stderr of the running processes inside the pods) and are
-        deleted in the parent implementation as part of the failing the job process.
-
-        :param job_state:
-        :return:
-        """
-
-        # First we rescue the pods logs
-        try:
-            with open(job_state.output_file, 'r') as outfile:
-                stdout_content = outfile.read()
-        except IOError:
-            stdout_content = "<Failed to recuperate log for job {} >".format(job_state.job_id)
-            log.exception("Failed to get stdout for job %s", job_state.job_id)
-
-        if getattr(job_state, 'stop_job', True):
-            self.stop_job(job_state.job_wrapper)
-        self._handle_runner_state('failure', job_state)
-        # Not convinced this is the best way to indicate this state, but
-        # something necessary
-        if not job_state.runner_state_handled:
-            job_state.job_wrapper.fail(
-                message=getattr(job_state, 'fail_message', 'Job failed'),
-                stdout=stdout_content, stderr='See stdout for pod\'s stderr.'
-            )
-            if job_state.job_wrapper.cleanup_job == "always":
-                job_state.cleanup()
-
-    def __produce_log_file(self, job_state):
-        pod_r = Pod.objects(self._pykube_api).filter(selector="app=" + job_state.job_id)
-        log_string = ""
-        for pod_obj in pod_r.response['items']:
-            try:
-                pod = Pod(self._pykube_api, pod_obj)
-                log_string += "\n\n==== Pod " + pod.name + " log start ====\n\n"
-                log_string += pod.logs(timestamps=True)
-                log_string += "\n\n==== Pod " + pod.name + " log end   ===="
-            except Exception as detail:
-                log.info("Could not write log file for pod %s due to HTTPError %s",
-                         pod_obj['metadata']['name'], detail)
 
     def stop_job(self, job_wrapper):
         """Attempts to delete a dispatched job to the k8s cluster"""
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -263,10 +263,22 @@ class KubernetesJobRunner(AsynchronousJobRunner):
     def __get_k8s_containers(self, ajs):
         """Fills in all required for setting up the docker containers to be used, including setting a pull policy if
            this has been set.
+           $GALAXY_VIRTUAL_ENV is set to None to avoid the galaxy virtualenv inside the tool container.
+           $GALAXY_LIB is set to None to avoid changing the python path inside the container.
+           Setting these variables changes the described behaviour in the job file shell script
+           used to execute the tool inside the container.
         """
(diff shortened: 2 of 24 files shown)


mvdbeek:
Alright, this is ready now. @pcm32 @ilveroluca can you guys give this a try ? I only have a toy kubernetes cluster to play with.

lib/galaxy/jobs/command_factory.py
```diff
@@ -229,24 +233,37 @@ def __init__(self, initial_command=u''):
...
 
     def append_commands(self, commands):
         self.append_command(u"; ".join(c for c in commands if c))
 
+    def capture_stdout_stderr(self, stdout_file, stderr_file):
+        self.prepend_command("""out="${TMPDIR:-/tmp}/out.$$" err="${TMPDIR:-/tmp}/err.$$"
+mkfifo "$out" "$err"
+trap 'rm "$out" "$err"' EXIT
+tee -a stdout.log < "$out" &
+tee -a stderr.log < "$err" >&2 &""",
```
jmchilton, line 260: Not to self - confirmed minimal biocontainers have these utils.

jmchilton (approved):

jmchilton:
It looks good to me, I was going to merge before I saw the above comment. If it makes testing easier let me know and I'll merge it, looks ready to go to me.

mvdbeek:
Yeah, next step is a set of re-usable runner tests. I want all runners to make sure they know about Galaxy-isms like GALAXY_SLOTS, stderr etc ... . If bugs arise I'm happy and in a good position to fix them.

pcm32:
I have set up some testing for Galaxy-Kubernetes on travis [here](https://github.com/ebi-gene-expression-group/container-galaxy-sc-tertiary/blob/develop/.travis.yml). We could put variations of this if needed somewhere on the Galaxy tests. It is missing though to have a job execution test (so it is currently more of a test that Galaxy spins up, but should be easy to extend for sending a job and waiting for it). It takes a little while though, in case you have alternatives to travis.

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -252,41 +259,42 @@ def __get_k8s_mountable_volumes(self, job_wrapper):
...
         }
 
-        resources = self.__get_resources(job_wrapper)
+        resources = self.__get_resources(ajs.job_wrapper)
         if resources:
+            envs = []
+            if 'limits' in resources:
+                limits = resources['limits']
+                if 'memory' in limits:
+                    envs.append({'name': 'GALAXY_MEMORY_MB', 'value': str(ByteSize(limits['memory']).to_unit('M', as_string=False))})
```
pcm32, line 281: limits normally come provided as Gi normally, are you sure that this part will work fine? Or ByteSize finds out the unit in which the require/limits are given?

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -252,41 +259,42 @@ def __get_k8s_mountable_volumes(self, job_wrapper):
...
+                if 'cpu' in limits:
+                    envs.append({'name': 'GALAXY_SLOTS', 'value': limits['cpu']})
```
pcm32: Please note that cpu values can be fractions of CPU, not sure if that is going to play nicely with the usage of $GALAXY_SLOTS in other places. I would take the ceiling of the value of cpu limits to put on GALAXY_SLOTS.

lib/galaxy/jobs/runners/kubernetes.py
mvdbeek: Right, it can take all the valid Kubernetes strings from https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#meaning-of-memory and turn them into bytes

lib/galaxy/jobs/runners/kubernetes.py
mvdbeek: that's a good point, I'll do that.

mvdbeek:
I think with this travis setup you should be able to run the integration tests here, which run a couple of jobs that use important Galaxy features. In the best case you'd just have to `sh run_tests.sh -integration test/integration/test_kubernetes_runner.py`

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -336,48 +344,23 @@ def __get_cpu_request(self, job_wrapper):
...
     def __transform_memory_value(self, mem_value):
-        """Transforms memory value
-
-           If the value is 0 and not a string, then None is returned.
-           If the value has a decimal part, then it is multiplied by 1000 and expressed as Megabytes.
-           If the value is an integer, then it is truncated and expressed as Gigabytes.
-           If it is an already formatted string, it is returned as it was.
         """
-        if not isinstance(mem_value, str) and float(mem_value) == 0:
-            return None
```
pcm32, line 375: We don't need to handle this corner case anymore?

mvdbeek: I don't think this could have worked as this was always going to be a string after parsing the destination params.

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -501,63 +484,6 @@ def __job_failed_due_to_low_memory(self, job_state):
...
-                job_state.cleanup()
-
-    def __produce_log_file(self, job_state):
-        pod_r = Pod.objects(self._pykube_api).filter(selector="app=" + job_state.job_id)
-        log_string = ""
-        for pod_obj in pod_r.response['items']:
-            try:
-                pod = Pod(self._pykube_api, pod_obj)
-                log_string += "\n\n==== Pod " + pod.name + " log start ====\n\n"
-                log_string += pod.logs(timestamps=True)
```
pcm32, line 543: So if we are removing this, where are we now retrieving the stdout of the tools when running on k8s? I couldn't find in the PR any other part where this is happening.

lib/galaxy/jobs/runners/util/job_script/CLUSTER_SLOTS_STATEMENT.sh
```diff
@@ -23,6 +23,9 @@ elif [ -f "$PBS_NODEFILE" ]; then
     GALAXY_SLOTS=`wc -l < $PBS_NODEFILE`
 elif [ -n "$LSB_DJOB_NUMPROC" ]; then
     GALAXY_SLOTS="$LSB_DJOB_NUMPROC"
+elif [ -n "$GALAXY_SLOTS" ]; then
+    # kubernetes runner injects GALAXY_SLOTS into environment
+    $GALAXY_SLOTS=$GALAXY_SLOTS
```
pcm32: This should be `GALAXY_SLOTS=$GALAXY_SLOTS`, you have an extra $.

lib/galaxy/jobs/runners/kubernetes.py
mvdbeek: This is now part of the command line generation -- see `capture_stdout_stderr`. This will intercept the job script output, but continue printing out, so you can still use the kubernetes pod logs if necessary.

pcm32 (changes requested):
This is a great improvement, many thanks! I have some questions regarding how certain things are handled now, most notably the logs of the tool (stdout and stderr, which you normally retrieve through kubectl logs). However I found a typo somewhere, so I thought I'd made it a proper review. Unfortunately I made most of the comments outside the review. So my concerns are not blocking of course (but would appreciate some clarification), but the typo at least it is I guess.

lib/galaxy/jobs/runners/util/job_script/CLUSTER_SLOTS_STATEMENT.sh
mvdbeek: hah, good one. it still worked because we avoid the `else`, I'll fix that in a rebase.

lib/galaxy/jobs/runners/kubernetes.py
mvdbeek: Alright, [I've set the CPU to 1.9 in the job_conf.xml](https://github.com/galaxyproject/galaxy/pull/6958/commits/676bdb5b38ed3032dea836d022ab299014d0df4a) setup and we were already testing that GALAXY_SLOTS is 2.

lib/galaxy/jobs/command_factory.py
pcm32: mkfifo and tee are on alpine:3.8, but not trap :-(. We need a replacement for it or handle it somehow. I tried with `docker run -it alpine:3.8 sh` and checked inside, otherwise this would make Galaxy incompatible with tools built on top of alpine.

pcm32: well I searched by `which` but I think trap could be a shell built-in? Then my comment wouldn't apply of course....

mvdbeek: Are you sure about that? This script works for me on `alpine:3.8`
```
#!/bin/sh

out="${TMPDIR:-/tmp}/out.$$" err="${TMPDIR:-/tmp}/err.$$"
mkfifo "$out" "$err"
trap 'rm "$out" "$err"' EXIT
tee -a stdout.log < "$out" &
tee -a stderr.log < "$err" >&2 &
bwa >"$out" 2>"$err"
echo $? > exit_code
```

also trap seems to be there: 
```
/ # command -v trap
trap
/ #
```

ilveroluca: `trap` should be ok.  It's a shell command compatible with both `sh` and `bash`

```
17:05 [pireddu@slynx ~] docker run -it alpine:3.8 sh
/ # trap 'echo "bye bye"' EXIT
```
Enter Ctrl+d...
```
/ # bye bye
```

nsoranzo: FYI you can use `command -V trap` to find out if it's a builtin, alias or normal executable:
```
$ command -V trap
trap is a shell builtin
```

ilveroluca: Scienziato! :bowing_man: :smile:

mvdbeek:
Oh, and if you want to test this using docker for mac you have to set the TEMP variable to something different than `/var/private`, since that can't be mounted by docker. I'm just doing `TMPDIR=~/temp sh run_tests.sh -integration test/integration/test_kubernetes_runner.py`

pcm32 (approved):
LGTM! Thanks again! I should be able to test this after Thursday if that is fine.

mvdbeek:
Sure, I'm not in a hurry for this.

pcm32:
On second thoughts, I think that galaxy slots should be equal to the floor of cpu-limits instead of it's ceiling, otherwise we risk having the job killed by kubernetes for trespassing the max amount of CPUs allocated.

pcm32:
Handling of course the case where this would lead to zero slots, which might produce weird behaviour in the tool.

mvdbeek:
Is that how kubernetes handles this ? I'd assume it'd just throttle, otherwise I can't really see how fractions of a cpu would work ?

pcm32:
Yes, you're right, it throttles for CPU, it kills for memory, so yes, no change needed on that!

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -435,44 +418,44 @@ def check_watched_item(self, job_state):
 
             # This assumes jobs dependent on a single pod, single container
             if succeeded > 0:
-                self.__produce_log_file(job_state)
```
pcm32, line 438: Unfortunately in the new setup I fail to see the stdout and stderrr of jobs in Galaxy after execution. Any ideas @mvdbeek? I can confirm that this is viewable before this PR (as I was testing the same tool at both sides.

pcm32 (changes requested):
I have found that stdout and stderr are not visible now from the Galaxy UI on this branch when running tools with k8s.

lib/galaxy/jobs/runners/kubernetes.py
pcm32: (I'm not sure the cuprit is actually there, I just pointed to any potentially related piece of the code)

mvdbeek: Are you using a specific tool to check this? I'm a bit surprised by this since the integration test does test visibility of stdout and stderr messages

test/integration/test_kubernetes_runner.py
```diff
@@ -0,0 +1,229 @@
...
+            "job_properties",
+            inputs,
+            self.history_id,
+            assert_ok=False,
+        )
+        result = self.dataset_populator.wait_for_tool_run(run_response=running_response, history_id=self.history_id, assert_ok=False).json()
+        details = self.dataset_populator.get_job_details(result['jobs'][0]['id'], full=True).json()
+
+        assert details['state'] == 'error', details
+        assert details['stdout'] == 'The bool is not true\n', details
```
mvdbeek, line 228: This test tests the presence of messages on stdout and stderr, so I'd assume this works ?

lib/galaxy/jobs/runners/kubernetes.py
pcm32: Well I saw it on a specific tool we are developing to bring data from our archive... but I had just run the same tool and seen its stdout at the older branch, so I know that the tool generates stdout (this is a tool we are developing here: https://github.com/ebi-gene-expression-group/container-galaxy-sc-tertiary/blob/develop/tools/tertiary-analysis/data/retrieve-scxa.xml). I don't think is related to the tool though. I will check with another one very soon.

pcm32: I see the lack of stdout with other tools as well (haven't been able to see stdout on any tools I have tried).

pcm32:
I also get some strange errors with python tools:
```
python: error while loading shared libraries: libz.so.1: cannot open shared object file: No such file or directory
```

I'm inclined to blame the call of `_galaxy_setup_environment` that now makes it to the script executed inside the container. I don't think that we should meddle in any way with the environment inside the container (ie python env active in this case), besides mere environment variables of course.

Would it be possible to make the script avoid that `_galaxy_setup_environment` or not call it?

lib/galaxy/jobs/runners/kubernetes.py
mvdbeek: I can't reproduce this, this is working fine for me (see the red items, this is the text from stderr):
screen shot 2018-11-23 at 17 01 50

As in other job runner we only produce a message on stdout/stderr if there is actually something logged, the previous implementation was always producing a message with the pod logs I think.
The pod logs for instance shouldn't appear in a tools' stdout/stderr, if necessary the admin can look at those. If the pod fails we get `More pods failed than allowed. See stdout for pods details.`

mvdbeek:
I agree, some things in `_galaxy_setup_environment` aren't necessary in a kubernetes setup, I'll see if we can modularize this. Activating the virtual env would indeed be problematic.

lib/galaxy/jobs/runners/kubernetes.py
pcm32: ...interesting... do you manage to see stdout? I couldn't see that either. I did see the Fatal errors like you show when using python tools. So maybe we were just misusing the stderr. The logs were used to produce the stdout, but I agree is much better to be able to separate err and out (logs produced both).

jmchilton:
> I'll see if we can modularize this

So this is back in WIP?

mvdbeek:
yes

pcm32:
Just tried the env vars based fix, but python tools are still throwing an error:

```
Fatal error: Exit code 127 ()
```

with no more details. The container does see the GALAXY_VIRTUAL_ENV and GALAXY_LIB set to None.

pcm32:
...errors are different... maybe setting GALAXY_LIB to None was exaggerated... will try without that..

pcm32:
No, on revision I think is correct to set GALAXY_LIB to None, otherwise we might end up adding virtualenv paths to the containers python path, wrecking python execution (which should depend only on the python setup of the tool container). I'm however confused by:

https://github.com/galaxyproject/galaxy/blob/9b2f3af6f3967ab172cb5bbaa527ee622dd8fcf5/lib/galaxy/jobs/runners/util/job_script/__init__.py#L44

this gives me the impression that already those variables where set to None by default. The other issue is that apparently I should be setting the lowercase version of the variables to None, not the upper case ones... will try that better.

pcm32:
Error remains despite changes to env vars to lowercase.

pcm32:
Non-python tools show no issues... it is with a python tool (that otherwise works on 18.09 without these changes on the same setup).

pcm32:
Ok, the problem seems to be that even though I set env variables at the k8s job level, the variables used to write the script happen much earlier and the main script of the job still gets GALAXY_VIRTUAL_ENV and GALAXY_LIB set to values that are different to None. I went into a failed job and reproduce it but changing manually those variables in the script and that fixed it. So question is, how can the runner control variables GALAXY_VIRTUAL_ENV and GALAXY_LIB (probably the lowercase ones) before they get injected into the main execution script? Setting the job_wrapper.galaxy_lib_dir to None?

pcm32:
Besides setting those variables somehow to None, the other issue is that the function `_galaxy_setup_environment()` gets as well a rogue activation of the galaxy virtualenv which cannot be  turned off in any way:

```
_galaxy_setup_environment() {                                                                                                    
    local _use_framework_galaxy="$1"                                                                                             
    _GALAXY_JOB_HOME_DIR="/export/galaxy-central/database/job_working_directory/000/4/home"                                      
    _GALAXY_JOB_TMP_DIR=$([ ! -e '/export/galaxy-central/database/job_working_directory/000/4/tmp' ] || mv '/export/galaxy-centra
                                                                                                                                 
. "/export/venv/bin/activate"                                                                                                    
    if [ "$GALAXY_LIB" != "None" -a "$_use_framework_galaxy" = "True" ]; then                                                    
...
```

pcm32:
@jmchilton @mvdbeek what would be the correct way forward? I was thinking on adding the following:

- In the job template that generates the bash script for execution I would add some overrides for GALAXY_LIB and GALAXY_VIRTUAL_ENV of the type:

```
GALAXY_LIB=${OVERRIDE_GALAXY_LIB:-"$galaxy_lib"}
GALAXY_VIRTUAL_ENV=${OVERRIDE_GALAXY_VIRTUAL_ENV:-''$galaxy_virtual_env"}
```

so that you can override with either "None" or a value of choice by setting the optional $OVERRIDE_* vars.

- A clause to avoid the `. "/export/venv/bin/activate"` line via a:

```
if [ -z ${IGNORE_GALAXY_VIRTUAL_ENV+x} ]; then
   $env_setup_commands
fi
```  

which would only have an effect if $IGNORE_GALAXY_VIRTUAL_ENV is defined. I would set these variables in particular through the k8s job definition.

Would this be fine @jmchilton @mvdbeek or is there a more elegant approach? I don't see one, and other runners using containers don't seem to be handling this, but they are manually handling the command line generation, as it was the case before for k8s. Please let me know so that we can have this working.

mvdbeek:
@pcm32 I'll have a look, there is probably a flag missing for containerization somewhere.

pcm32:
I would suggest as well that the integration tests try to incorporate the case of a python tool being scheduled, to account for this case.

mvdbeek:
Actually the single mount point thing makes it difficult to test tools with a wrapper living next to the tool ... :/

mvdbeek:
Alright, @pcm32 https://github.com/galaxyproject/galaxy/pull/6958/commits/7e329770ec66cb41518fdebdd873008b33616b4b should fix your issue with GALAXY_VIRTUAL_ENV bleeding into the container. I didn't have any problem also without the commit, but that's because I am only mounting in the test directory and the tool directory, so the virtualenv couldn't be activated. That's fixed now, I hope this works for you as well. Note that I changed the syntax for mounting in PVCs in https://github.com/galaxyproject/galaxy/pull/6958/commits/65f33dfea6fe16d79388e835043f4f33ff6fe363, let me know what you think about that.

pcm32:
Thanks, I'll try it as soon as I can. I have some concerns on the multiple mount points (made the comment on that commit), but great that we can have this working!

pcm32:
Unfortunately after the last rebase the images don't manage to spin up galaxy correctly, we must have grabbed a bug in dev. Images build correctly (see docker-galaxy-stable/compose build-orchestration-images.sh), but fail to start with this error:

```
galaxy.jobs DEBUG 2019-02-21 14:36:18,901 [p:1953,w:0,m:0] [MainThread] Done loading job configuration
  File "lib/galaxy/webapps/galaxy/buildapp.py", line 49, in app_factory
    app = galaxy.app.UniverseApplication(global_conf=global_conf, **kwargs)
  File "lib/galaxy/app.py", line 128, in __init__
    self.watchers = ConfigWatchers(self)
  File "lib/galaxy/webapps/galaxy/config_watchers.py", line 34, in __init__
    self.tool_data_watcher = get_tool_data_dir_watcher(self.app.tool_data_tables, config=self.app.config)
  File "lib/galaxy/tools/toolbox/watcher.py", line 57, in get_tool_data_dir_watcher
    observer_class = get_observer_class(config_value, default="False", monitor_what_str="tool-data directory")
  File "lib/galaxy/tools/toolbox/watcher.py", line 46, in get_observer_class
    raise Exception(message)
Exception: Watchdog library unavailable, cannot monitor tool-data directory.
```

as such I cannot yet test this :-( . I will investigate what is the problem next week and try to fix.

Happy to go with the additional mount points, but will probably stay without it for a while in a branch until we move to 19.05, to avoid changing the helm charts (I maintain a number of different instances at different galaxy versions, and I don't have man power to deal with that now).

mvdbeek:
This seems to be missing the watchdog library, so you can either run through common_startup.sh or disable watch_tool_data_dir https://github.com/galaxyproject/galaxy/blob/dev/config/galaxy.yml.sample#L402 . I don't think this has changed on the Galaxy side.

pcm32:
Yes, strange, as I haven't changed either the default setting, so it should be turned off (for watchdog). I tried a completely clean start with a fresh PV, but still.

pcm32:
I had to go inside the running container and install watchdog... unfortunately the container fails with the same error:
```
$ kubectl logs galaxy-mk-1809-int-test-2-dwxbm
python: error while loading shared libraries: libz.so.1: cannot open shared object file: No such file or directory
```

if I inspect the script file, generated for the failed job I see:
```
#!/bin/bash



_galaxy_setup_environment() {
    local _use_framework_galaxy="$1"
    _GALAXY_JOB_HOME_DIR="/export/galaxy-central/database/job_working_directory/000/2/home"
    _GALAXY_JOB_TMP_DIR=$([ ! -e '/export/galaxy-central/database/job_working_directory/000/2/tmp' ] || mv '/export/galaxy-central/database/job_working_directory/000/2/tmp' '/export/galaxy-central/database/job_working_directory/000/2'/tmp.$(date +%Y%m%d-%H%M%S) ; mkdir '/export/galaxy-central/database/job_working_directory/000/2/tmp'; echo '/export/galaxy-central/database/job_working_directory/000/2/tmp')
    
. "/export/venv/bin/activate"
    if [ "$GALAXY_LIB" != "None" -a "$_use_framework_galaxy" = "True" ]; then
        if [ -n "$PYTHONPATH" ]; then
            PYTHONPATH="$GALAXY_LIB:$PYTHONPATH"
        else
            PYTHONPATH="$GALAXY_LIB"
        fi
        export PYTHONPATH
    fi
    # These don't get cleaned on a re-run but may in the future.
    [ -z "$_GALAXY_JOB_TMP_DIR" -a ! -f "$_GALAXY_JOB_TMP_DIR" ] || mkdir -p "$_GALAXY_JOB_TMP_DIR"
    [ -z "$_GALAXY_JOB_HOME_DIR" -a ! -f "$_GALAXY_JOB_HOME_DIR" ] || mkdir -p "$_GALAXY_JOB_HOME_DIR"
    if [ "$GALAXY_VIRTUAL_ENV" != "None" -a -f "$GALAXY_VIRTUAL_ENV/bin/activate" \
         -a "`command -v python`" != "$GALAXY_VIRTUAL_ENV/bin/python" ]; then
        . "$GALAXY_VIRTUAL_ENV/bin/activate"
    fi
}


# The following block can be used by the job system
# to ensure this script is runnable before actually attempting
# to run it.
if [ -n "$ABC_TEST_JOB_SCRIPT_INTEGRITY_XYZ" ]; then
    exit 42
fi

export GALAXY_SLOTS_CONFIGURED="1"
if [ -n "$SLURM_CPUS_ON_NODE" ]; then
    # This should be valid on SLURM except in the case that srun is used to
    # submit additional job steps under an existing allocation, which we do not
    # currently do.
    GALAXY_SLOTS="$SLURM_CPUS_ON_NODE"
elif [ -n "$SLURM_NTASKS" ] || [ -n "$SLURM_CPUS_PER_TASK" ]; then
    # $SLURM_CPUS_ON_NODE should be set correctly on SLURM (even on old
    # installations), but keep the $SLURM_NTASKS logic as a backup since this
    # was the previous method under SLURM.
    #
    # Multiply these values since SLURM_NTASKS is total tasks over all nodes.
    # GALAXY_SLOTS maps to CPUS on a single node and shouldn't be used for
    # multi-node requests.
    GALAXY_SLOTS=`expr "${SLURM_NTASKS:-1}" \* "${SLURM_CPUS_PER_TASK:-1}"`
elif [ -n "$NSLOTS" ]; then
    GALAXY_SLOTS="$NSLOTS"
elif [ -n "$NCPUS" ]; then
    GALAXY_SLOTS="$NCPUS"
elif [ -n "$PBS_NCPUS" ]; then
    GALAXY_SLOTS="$PBS_NCPUS"
elif [ -f "$PBS_NODEFILE" ]; then
    GALAXY_SLOTS=`wc -l < $PBS_NODEFILE`
elif [ -n "$LSB_DJOB_NUMPROC" ]; then
    GALAXY_SLOTS="$LSB_DJOB_NUMPROC"
elif [ -n "$GALAXY_SLOTS" ]; then
    # kubernetes runner injects GALAXY_SLOTS into environment
    GALAXY_SLOTS=$GALAXY_SLOTS
else
    GALAXY_SLOTS="1"
    unset GALAXY_SLOTS_CONFIGURED
fi

export GALAXY_SLOTS
GALAXY_VIRTUAL_ENV="None"
_GALAXY_VIRTUAL_ENV="None"
PRESERVE_GALAXY_ENVIRONMENT="False"
GALAXY_LIB="/export/galaxy-central/lib"
```

so that means that while GALAXY_VIRTUAL_ENV is being correctly set to None, there is still a direct call to `. "/export/venv/bin/activate"` that is not being controlled by any environment variable, messing up the tools python environment. :-(

mvdbeek:
I am not seeing this, any chance this is caused by a non-default setting in your job_conf.xml file ?

mvdbeek:
e.g https://github.com/galaxyproject/galaxy/blob/dev/config/job_conf.xml.sample_advanced#L605 / https://github.com/galaxyproject/galaxy/blob/dev/config/job_conf.xml.sample_advanced#L606

mvdbeek:
So this is what I see:
```
#!/bin/sh



_galaxy_setup_environment() {
    local _use_framework_galaxy="$1"
    _GALAXY_JOB_HOME_DIR="/Users/mvandenb/temp/tmpw_906fen/000/2/home"
    _GALAXY_JOB_TMP_DIR=$([ ! -e '/Users/mvandenb/temp/tmpw_906fen/000/2/tmp' ] || mv '/Users/mvandenb/temp/tmpw_906fen/000/2/tmp' '/Users/mvandenb/temp/tmpw_906fen/000/2'/tmp.$(date +%Y%m%d-%H%M%S) ; mkdir '/Users/mvandenb/temp/tmpw_906fen/000/2/tmp'; echo '/Users/mvandenb/temp/tmpw_906fen/000/2/tmp')
    
SOME_ENV_VAR="42"; export SOME_ENV_VAR
    if [ "$GALAXY_LIB" != "None" -a "$_use_framework_galaxy" = "True" ]; then
        if [ -n "$PYTHONPATH" ]; then
            PYTHONPATH="$GALAXY_LIB:$PYTHONPATH"
        else
            PYTHONPATH="$GALAXY_LIB"
        fi
        export PYTHONPATH
    fi
    # These don't get cleaned on a re-run but may in the future.
    [ -z "$_GALAXY_JOB_TMP_DIR" -a ! -f "$_GALAXY_JOB_TMP_DIR" ] || mkdir -p "$_GALAXY_JOB_TMP_DIR"
    [ -z "$_GALAXY_JOB_HOME_DIR" -a ! -f "$_GALAXY_JOB_HOME_DIR" ] || mkdir -p "$_GALAXY_JOB_HOME_DIR"
    if [ "$GALAXY_VIRTUAL_ENV" != "None" -a -f "$GALAXY_VIRTUAL_ENV/bin/activate" \
         -a "`command -v python`" != "$GALAXY_VIRTUAL_ENV/bin/python" ]; then
        . "$GALAXY_VIRTUAL_ENV/bin/activate"
    fi
}


# The following block can be used by the job system
# to ensure this script is runnable before actually attempting
# to run it.
if [ -n "$ABC_TEST_JOB_SCRIPT_INTEGRITY_XYZ" ]; then
    exit 42
fi

export GALAXY_SLOTS_CONFIGURED="1"
if [ -n "$SLURM_CPUS_ON_NODE" ]; then
    # This should be valid on SLURM except in the case that srun is used to
    # submit additional job steps under an existing allocation, which we do not
    # currently do.
    GALAXY_SLOTS="$SLURM_CPUS_ON_NODE"
elif [ -n "$SLURM_NTASKS" ] || [ -n "$SLURM_CPUS_PER_TASK" ]; then
    # $SLURM_CPUS_ON_NODE should be set correctly on SLURM (even on old
    # installations), but keep the $SLURM_NTASKS logic as a backup since this
    # was the previous method under SLURM.
    #
    # Multiply these values since SLURM_NTASKS is total tasks over all nodes.
    # GALAXY_SLOTS maps to CPUS on a single node and shouldn't be used for
    # multi-node requests.
    GALAXY_SLOTS=`expr "${SLURM_NTASKS:-1}" \* "${SLURM_CPUS_PER_TASK:-1}"`
elif [ -n "$NSLOTS" ]; then
    GALAXY_SLOTS="$NSLOTS"
elif [ -n "$NCPUS" ]; then
    GALAXY_SLOTS="$NCPUS"
elif [ -n "$PBS_NCPUS" ]; then
    GALAXY_SLOTS="$PBS_NCPUS"
elif [ -f "$PBS_NODEFILE" ]; then
    GALAXY_SLOTS=`wc -l < $PBS_NODEFILE`
elif [ -n "$LSB_DJOB_NUMPROC" ]; then
    GALAXY_SLOTS="$LSB_DJOB_NUMPROC"
elif [ -n "$GALAXY_SLOTS" ]; then
    # kubernetes runner injects GALAXY_SLOTS into environment
    GALAXY_SLOTS=$GALAXY_SLOTS
else
    GALAXY_SLOTS="1"
    unset GALAXY_SLOTS_CONFIGURED
fi

export GALAXY_SLOTS
GALAXY_VIRTUAL_ENV="None"
_GALAXY_VIRTUAL_ENV="None"
PRESERVE_GALAXY_ENVIRONMENT="False"
GALAXY_LIB="/Users/mvandenb/src/galaxy/lib"
GALAXY_PYTHON=`command -v python`
cd /Users/mvandenb/temp/tmpw_906fen/000/2
if [ -n "$SLURM_JOB_ID" ]; then
    GALAXY_MEMORY_MB=`scontrol -do show job "$SLURM_JOB_ID" | sed 's/.*\( \|^\)Mem=\([0-9][0-9]*\)\( \|$\).*/\2/p;d'` 2>memory_statement.log
fi
if [ -n "$SGE_HGR_h_vmem" ]; then
    GALAXY_MEMORY_MB_PER_SLOT=`echo "$SGE_HGR_h_vmem" | sed 's/G$/ * 1024/' | bc | cut -d"." -f1` 2>memory_statement.log
fi

if [ -z "$GALAXY_MEMORY_MB_PER_SLOT" -a -n "$GALAXY_MEMORY_MB" ]; then
    GALAXY_MEMORY_MB_PER_SLOT=$(($GALAXY_MEMORY_MB / $GALAXY_SLOTS))
elif [ -z "$GALAXY_MEMORY_MB" -a -n "$GALAXY_MEMORY_MB_PER_SLOT" ]; then
    GALAXY_MEMORY_MB=$(($GALAXY_MEMORY_MB_PER_SLOT * $GALAXY_SLOTS))
fi
[ "${GALAXY_MEMORY_MB--1}" -gt 0 ] 2>>memory_statement.log && export GALAXY_MEMORY_MB || unset GALAXY_MEMORY_MB
[ "${GALAXY_MEMORY_MB_PER_SLOT--1}" -gt 0 ] 2>>memory_statement.log && export GALAXY_MEMORY_MB_PER_SLOT || unset GALAXY_MEMORY_MB_PER_SLOT

echo "$GALAXY_SLOTS" > '/Users/mvandenb/temp/tmpw_906fen/000/2/__instrument_core_galaxy_slots' 
echo "$GALAXY_MEMORY_MB" > '/Users/mvandenb/temp/tmpw_906fen/000/2/__instrument_core_galaxy_memory_mb' 
date +"%s" > /Users/mvandenb/temp/tmpw_906fen/000/2/__instrument_core_epoch_start
out="${TMPDIR:-/tmp}/out.$$" err="${TMPDIR:-/tmp}/err.$$"
mkfifo "$out" "$err"
trap 'rm "$out" "$err"' EXIT
tee -a stdout.log < "$out" &
tee -a stderr.log < "$err" >&2 & rm -rf working; mkdir -p working; cd working; /Users/mvandenb/temp/tmpw_906fen/000/2/tool_script.sh > '/Users/mvandenb/temp/tmpw_906fen/000/2/galaxy_2.o' 2> '/Users/mvandenb/temp/tmpw_906fen/000/2/galaxy_2.e'; return_code=$?; sh -c "exit $return_code"
echo $? > /Users/mvandenb/temp/tmpw_906fen/000/2/galaxy_2.ec
date +"%s" > /Users/mvandenb/temp/tmpw_906fen/000/2/__instrument_core_epoch_end
```

pcm32:
You rock @mvdbeek! It must be this line:
https://github.com/ebi-gene-expression-group/container-galaxy-sc-tertiary/blob/develop/config/job_conf.xml#L39

I'll rebuild removing that and try again as soon as I can.

pcm32:
Yes, after that it works. Thanks @mvdbeek for all the follow ups and patiently going through all my concerns. This PR is fine for me.

test/functional/tools/samples_tool_conf.xml
```diff
@@ -60,6 +60,7 @@
   <tool file="job_environment_default.xml" />
   <tool file="job_environment_default_legacy.xml" />
   <tool file="job_environment_explicit_shared_home.xml" />
+  <tool file="galaxy_slots_and_memory.xml" />
```
nsoranzo, line 63: @mvdbeek Did you forget to add the `galaxy_slots_and_memory.xml` file? Loading of this tool fails and the corresponding `test_slots_and_memory` integration test is skipped.

PR #6959, opened 2018-11-01
Switch from PhantomJS to Chrome Headless for unit tests

Nerdinacan:
I ran into a problem with another PR because we're using PhantomJS (which is no longer maintained) to run our unit tests and it doesn't play well with certain modern javascript features which now have pretty broad browser acceptance. (And are non polyfillable, such as Proxy). 

Furthermore, I figured I'd take this opportunity to clean up the karma configs so we can run both the legacy Qunit tests as well as a 2nd karma run of mocha tests. I would like to see more unit testing, which will encourage more rigorous separation of concerns. (Unit testing a file with 25 dependencies is a headache, but testing a smaller, focused object or class is pretty straightforward.)

I'd like to switch over to Chrome headless anyway as it's kind of the industry standard and it's actually maintained. If this goes well, I'd like to use Chrome headless for selenium as well, but I'll deal with that in a separate PR.

I may need some help making ChromeHeadless available to the Jenkins build server, as that falls a little outside my experience. But I figured I'd throw this up just to see what fails. :)


jmchilton:
The unit tests run on Travis - that environment can be tweaked with .travis.yml (https://github.com/galaxyproject/galaxy/blob/dev/.travis.yml). We don't run that with sudo - so there may be issues.

The Selenium tests run in a Docker container built with this Dockerfile (https://github.com/galaxyproject/galaxy/blob/dev/test/docker/base/Dockerfile), so in theory Jenkins changes should not be needed - just a Dockerfile update and rebuild. I can help republish a new image if you PR changes to the Dockerfile.

Nerdinacan:
Thanks John. I don't have a whole lot of experience with either Travis or Jenkins, but I can do some homework.

Nerdinacan:
> The unit tests run on Travis - that environment can be tweaked with .travis.yml (https://github.com/galaxyproject/galaxy/blob/dev/.travis.yml). We don't run that with sudo - so there may be issues.

Ok, still playing catch-up, but I did notice that the chrome implementation described in the travis docs wants sudo: required. So that is presumably the difficulty you're referencing. So two questions:

1. For my education, Is there a specific reason that we are not running sudo which prohibits us from switching it over? (I assume there must be a good reason.)

2. If switching to sudo: required is problematic, is there anything stopping me from making a separate travis job?

Nerdinacan:
Well, I got headless chrome running in the qunit build with the exception of a couple of the legacy tests that are dumpster fires for other reasons. 

I would be more concerned about those legacy tests failing if I didn't know they were failing for the exact reasons that the app needs to have its init chain and jquery dependencies refactored, which I am already addressing in a separate PR.

But hey! At least the thing runs.

Nerdinacan:
Think I'm done with this. Please merge whenever you get a chance to evaluate it. Thanks!

jmchilton (approved):

martenson (approved):

PR #6962, opened 2018-11-01
add RMA6 data type (MEGAN6 read-match archive)

bernt-matthias:
see manual http://ab.inf.uni-tuebingen.de/data/software/megan6/download/manual.pdf

discussion on magic numbers: http://megan.informatik.uni-tuebingen.de/t/rma-file-technical-question/972/3

tools are in preparation: 
https://github.com/bernt-matthias/mb-galaxy-tools/tree/master/tools/megan_tools
https://github.com/bernt-matthias/mb-galaxy-tools/tree/master/data_managers/data_manager_megan_tools/data_manager


bernt-matthias:
Ups: test failed because ` [Errno 28] No space left on device`.

mvdbeek:
I've put that node offline and retriggered the test

lib/galaxy/datatypes/binary.py
```diff
@@ -2168,6 +2168,29 @@ def sniff(self, filename):
             return f.read(8) == self._magic
 
 
+class RMA6(Binary):
+    """
+    Class describing an RMA6 (MEGAN6 read-match archive) file
+    >>> from galaxy.datatypes.sniff import get_test_fname
+    >>> fname = get_test_fname('diamond.rma')
```
mvdbeek: The new datatype tests requires the test data to end with `file_ext`, if you rename `'diamond.rma'` to `'diamond.rma6'` the test file will be uploaded, check that it will be sniffed correctly (in the presence of all other known datatypes) and downloaded again to verify the file is still intact.
(And I'm planning to extend this to all download all generated metadata)

bernt-matthias: Thanks for the explanation. Changed the extension.

PR #6967, opened 2018-11-05
Refactor workflow editor styles and move from mako to css

guerler:
This PR moves workflow editor styles from the mako template to the scss style sheet. It also removes redundant, unused or conflicting style definitions and improves the specificity of css selectors. This PR prepares the adaption of the new style theme for the workflow editor.


dannon:
(the one failing test is unrelated and was fixed in #6982)

PR #6729, opened 2018-09-18
Vue rewrite of admin data manager (local data) in steps

scrathat:
Contains the same changes as #6558 in standalone working steps. Every commit contains selenium tests for the rewritten component.

Index page
* Added Alert component: Light bootstrap-vue b-alert wrapper.
 - Hides automatically when given message is empty.
 - Translates galaxy state keywords into bootstrap ones.
* Added vue-router integration: Display (future) Data Manager Jobs and Job
  components inside router-views.
 - Added router-view component DataManagerView.
* Rewrite (old mako, new json interface, new vue component)
 - Index page (index.mako, data_managers_list, DataManager)
* Added Selenium test for DataManager component (index page)

Manager page
* Rewrite (old mako, new json interface, new vue component)
 - Manager page (manage_data_manager.mako, jobs_list, DataManagerJobs)
* Added Selenium test for DataManagerJobs component (manager page)

View job page
* Rewrite (old mako, new json interface, new vue component)
 - View job page (view_job.mako, job_info, DataManagerJob)
* Added Selenium test for DataManagerJob component (view job page)

Manage data table page
* Rewrite (old mako, new json interface, new vue component)
 - Manage data table page (manage_data_table.mako, tool_data_table_info, DataManagerTable)
* Added Selenium test for DataManagerTable component (Manage data table page)

PR #6970, opened 2018-11-06
Normalized jquery bundle

Nerdinacan:
So we KIND OF had a solution for this. Previously we jammed all jquery plugins into a script that frequently gets included (onload.js) in entry points that are rendered by webpack. However, that can fail because of [webpack's tree-shaking](https://webpack.js.org/guides/tree-shaking/): explicitly when you render an entry point that does not require onload.js.

This PR modifies the webpack configs to guarantee the same version of jquery with the same attached plugins regardless of whether that particular onload.js script got included. This was something that could occasionally happen in iframe documents.

This is accomplished pretty easily with webpack's module aliasing rules. The "jQuery" module now points to a file called "jquery.custom.js" which builds on "jqueryVendor" which points at the npm or local lib version. Then we use the expose-loader to put the jquery variable on every page rendered by webpack regardless of whether or not jquery was included in the source script.

The end result is the same version of jquery with the same plugins everywhere in the site. So no more weird ".select2 is not a function" errors.

### DIFF
--- a/client/galaxy/scripts/libs/jquery.custom.js
+++ b/client/galaxy/scripts/libs/jquery.custom.js
@@ -1,0 +1,24 @@
+/**
+ * jQuery and all its horrible plugins. Bundled together and repackaged into a
+ * single module, then aliased by webpack as "jquery";
+ */
+
+var jQuery = require("jqueryVendor");
+
+require("imports-loader?jQuery=jqueryVendor!libs/jquery/jquery.autocomplete");
+require("imports-loader?jQuery=jqueryVendor!libs/jquery/jquery.event.hover");
+require("imports-loader?jQuery=jqueryVendor!libs/jquery/jquery.event.drag");
+require("imports-loader?jQuery=jqueryVendor!libs/jquery/jquery.event.drop");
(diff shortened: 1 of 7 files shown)


Nerdinacan:
Found a general bug in one of the selenium tests. Don't think it's related to this specific PR, but I can address it.

test_import_dataset_from_history
This test fails if there's more than one saved history in the dropdown. It looks like the test always assumed that there would be only one saved history on the dropdown, so it didn't bother selecting the one it cared about, but now there's an "API history" or something in there, probably from some other test.

dannon:
(to document it here, I think it's the test from https://github.com/galaxyproject/galaxy/pull/6729/files#diff-aaafd90514e4845df6eaed99c3ce8be1R67 that's likely having a side effect in the recent selenium test failures, causing this problem)

dannon:
To clarify a little, this isn't addressing a bug in the current `dev` codebase.  It breaks all our jquery plugins out of onload, which we won't want to include in all entrypoints necessarily moving forward (though we do now).

We're going to follow up on the selenium fix (which is unrelated to this) in a subsequent PR.

PR #6971, opened 2018-11-06
biom1 datatype changes to add table column headers to the metadata

Slugger70:
This PR adds an element to the biom1 datatype metadata:

```python
MetadataElement(name="table_column_metadata_headers", default=[], desc="table_column_metadata_headers", param=MetadataParameter, readonly=True, visible=True, optional=True, no_value=[])
```
It populates the array with the column header labels found in the `columns` field of the biom table.

This is so we can use the column header labels in tool wrappers.

e.g. 
```xml
<param format="biom1" name="input" type="data" label="Input File"/>
<param name="subsetColumn" type="select" label="Select a column for subsetting">
    <options>
        <filter type="data_meta" ref="input" key="table_column_metadata_headers" />
    </options>
</param>
```

This is really useful for a bunch of tools we're developing for NMDS style visualisation of metagenomic data.

### DIFF
--- a/lib/galaxy/datatypes/text.py
+++ b/lib/galaxy/datatypes/text.py
@@ -12,6 +12,8 @@ import tempfile
 
 from six.moves import shlex_quote
 
+from time import sleep
+
 from galaxy.datatypes.data import get_file_peek, Text
 from galaxy.datatypes.metadata import MetadataElement, MetadataParameter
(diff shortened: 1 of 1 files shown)


Slugger70:
Argh, I had an old text.py. Apologies. :(

PR #6972, opened 2018-11-06
Added a metadata element to the biom1 datatype

Slugger70:
This PR adds an element to the biom1 datatype metadata:

```python
MetadataElement(name="table_column_metadata_headers", default=[], desc="table_column_metadata_headers", param=MetadataParameter, readonly=True, visible=True, optional=True, no_value=[])
```
It populates the array with the column header labels found in the `columns` field of the biom table.

This is so we can use the column header labels in tool wrappers.

e.g. 
```xml
<param format="biom1" name="input" type="data" label="Input File"/>
<param name="subsetColumn" type="select" label="Select a column for subsetting">
    <options>
          <filter type="data_meta" ref="input" key="table_column_metadata_headers" />
    </options>
</param>
```
This is really useful for a bunch of tools we're developing for NMDS style visualisation of metagenomic data.

### DIFF
--- a/lib/galaxy/datatypes/text.py
+++ b/lib/galaxy/datatypes/text.py
@@ -255,7 +255,7 @@ class Biom1(Json):
                         metadata_value = json_dict.get(b_name, None)
                         if b_name == "columns" and metadata_value:
                             mks = metadata_value[0]['metadata'].keys()
-                            keep_columns = { mk: None for mk in mks }
+                            keep_columns = {mk: None for mk in mks}
                             for column in metadata_value:
                                 for k in column['metadata']:
                                     if column['metadata'][k] is not None:
--- a/lib/galaxy/datatypes/text.py
+++ b/lib/galaxy/datatypes/text.py
@@ -272,7 +272,7 @@ class Biom1(Json):
                             metadata_value = b_transform[b_name](metadata_value)
                         setattr(dataset.metadata, m_name, metadata_value)
                     except Exception:
-                        log.exception("Something in the metadata detection for biom1 went wrong: " + str(Exception))
+                        log.exception("Something in the metadata detection for biom1 went wrong")
                         pass
 
 
--- a/lib/galaxy/datatypes/text.py
+++ b/lib/galaxy/datatypes/text.py
@@ -184,6 +184,7 @@ class Biom1(Json):
     MetadataElement(name="table_type", default="", desc="table_type", param=MetadataParameter, readonly=True, visible=True, optional=True, no_value="")
     MetadataElement(name="table_id", default=None, desc="table_id", param=MetadataParameter, readonly=True, visible=True, optional=True, no_value=None)
     MetadataElement(name="table_columns", default=[], desc="table_columns", param=MetadataParameter, readonly=True, visible=False, optional=True, no_value=[])
+    MetadataElement(name="table_column_metadata_headers", default=[], desc="table_column_metadata_headers", param=MetadataParameter, readonly=True, visible=True, optional=True, no_value=[])
 
     def set_peek(self, dataset, is_multi_byte=False):
         super(Biom1, self).set_peek(dataset)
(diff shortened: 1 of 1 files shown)


lib/galaxy/datatypes/text.py
```diff
@@ -252,10 +253,26 @@ def _transform_dict_list_ids(dict_list):
...
+                            for k in keep_columns:
+                                if not keep_columns[k] is None:
+                                    final_list.append(k)
+                            final_list.sort()
+                            dataset.metadata.table_column_metadata_headers = final_list
                         if b_name in b_transform:
                             metadata_value = b_transform[b_name](metadata_value)
                         setattr(dataset.metadata, m_name, metadata_value)
                     except Exception:
+                        log.exception("Something in the metadata detection for biom1 went wrong: " + str(Exception))
```
nsoranzo: `log.exception()` already logs the latest traceback. This won't work anyway because you would need to assign the `Exception` to a variable.

```suggestion
                        log.exception("Something in the metadata detection for biom1 went wrong")
```

lib/galaxy/datatypes/text.py
```diff
@@ -252,10 +253,26 @@ def _transform_dict_list_ids(dict_list):
                                          ('table_columns', 'columns')]:
                     try:
                         metadata_value = json_dict.get(b_name, None)
+                        if b_name == "columns" and metadata_value:
+                            mks = list(metadata_value[0]['metadata'].keys())
```
nsoranzo: I think `list()` is not needed.

lib/galaxy/datatypes/text.py
```diff
@@ -252,10 +253,26 @@ def _transform_dict_list_ids(dict_list):
...
+                            keep_columns = {}
+                            for mk in mks:
+                                keep_columns[mk] = None
```
nsoranzo: You can use a [dict comprehension](https://www.python.org/dev/peps/pep-0274/):

```python
                            keep_columns = {mk: None for mk in mks}
```

lib/galaxy/datatypes/text.py
```diff
@@ -252,10 +253,26 @@ def _transform_dict_list_ids(dict_list):
...
+                            for column in metadata_value:
+                                for k in column['metadata']:
+                                    if not column['metadata'][k] is None:
```
nsoranzo:
```suggestion
                                    if column['metadata'][k] is not None:
```

lib/galaxy/datatypes/text.py
```diff
@@ -252,10 +253,26 @@ def _transform_dict_list_ids(dict_list):
...
+                                        keep_columns[k] = 1
+                            final_list = []
+                            for k in keep_columns:
+                                if not keep_columns[k] is None:
```
nsoranzo:
```suggestion
                                if keep_columns[k] is not None:
```

lib/galaxy/datatypes/text.py
Slugger70: Done! Thanks Nicola

lib/galaxy/datatypes/text.py
```diff
@@ -252,10 +253,24 @@ def _transform_dict_list_ids(dict_list):
                                          ('table_columns', 'columns')]:
                     try:
                         metadata_value = json_dict.get(b_name, None)
+                        if b_name == "columns" and metadata_value:
+                            mks = metadata_value[0]['metadata'].keys()
+                            keep_columns = { mk: None for mk in mks }
```
nsoranzo: Can you remove the extra spaces to make flake8 happy? https://travis-ci.org/galaxyproject/galaxy/jobs/451325707

```suggestion
                            keep_columns = {mk: None for mk in mks}
```

Slugger70: Done! Cheers again @nsoranzo

nsoranzo:
@Slugger70 I've pushed a commit to your branch which simplifies the code a bit, can you double check if it's correct?

Slugger70:
@nsoranzo, It works fine. Thanks for the clean up. :)

nsoranzo (approved):

PR #6977, opened 2018-11-06
changed exit code and regex parsing

bernt-matthias:
... Just a little idea that I just had ...

I thought it could be handy to be able to combine exit codes and regexes that are derived from 
the command tag and those from the stdio tag. 

For instance one could make use of the detect_errors == "aggressive" regexes and add tool specific 
regexes via the stdio tag. 

Haven't thought about compatibility. Also documentation is missing and probably some other pieces of code (tests) need updates.

A possible alternative would be to add a property/element to the stdio tag that allows to load predefined sets of regexes and exit codes. I'm thinking of 
- language specific errors (e.g. the annoying zoo of JAVA oom messages)
- user defined error mesages (that may depend on the local computing platform)

Maybe this would be even nicer.

### DIFF
--- a/lib/galaxy/tools/parser/xml.py
+++ b/lib/galaxy/tools/parser/xml.py
@@ -348,6 +348,20 @@ class XmlToolSource(ToolSource):
         return output
 
     def parse_stdio(self):
+        """
+        parse error handling from command and stdio tag
+
+        returns list of exit codes, list of regexes
+        - exit_codes contain all non-zero exit codes (:-1 and 1:) if
+          detect_errors is default (if not legacy), exit_code, or aggressive
+        - the oom_exit_code if given and detect_errors is exit_code
+        - exit codes and regexes from the stdio tag
+          these are prepended to the list, i.e. are evaluated prior to regexes
+          and exit codes derived from the properties of the command tag.
+          thus more specific regexes of the same or more severe error level
+          are triggered first.
+        """
+
         command_el = self._command_el
         detect_errors = None
         if command_el is not None:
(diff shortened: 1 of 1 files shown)


mvdbeek:
@bernt-matthias there are some unit tests in https://github.com/galaxyproject/galaxy/blob/84597a12356aaa5b5d59ba95c30691739fed2300/test/unit/tools/test_parsing.py#L187, I think it would be straightforward to extend those.

bernt-matthias:
@mvdbeek thanks for the link. Wanted to see if this change would is considered useful and acceptable by the community .. or if the possible alternative I discussed at the end might be even better.

mvdbeek:
This is definitely useful and reduces clutter by allowing the combination of stdio regex and `detect_errors="exit_code"`, I don't think there are any downsides. 

I am not convinced about the suggested alternatives though, it seems to me that a grid manager is supposed to kill the OOM jobs, resulting in a sigkill. If for some reason galaxy doesn't register the exit code we can work on this, but it definitely does for torque, which I am using and I suspect (but have no proof) that this is the case for the majority of job runners.

The example we discussed in https://github.com/galaxyproject/tools-iuc/pull/2141#discussion_r230790163 in that light is not exactly what I would expect to happen IRL. Once I have a little more time I can set up an integration test case for kubernetes or the CLI runners.

In essence I don't think we should tailor tools to whatever weird thing the DRM is doing. If needed we can override things in the job runner code.

bernt-matthias:
ping @jmchilton @natefoo @bgruening

bernt-matthias:
Does somebody know something about the failing selenium test?

mvdbeek:
You can ignore that, it's been fixed in another PR

jmchilton:
Totally solid enhancement, thanks a bunch! More or less agree with @mvdbeek's review on future directions note with the caveat that "language specific errors" makes sense to me as a good idea even if "computing platform errors" should be avoided.

PR #6982, opened 2018-11-06
Update test/selenium_tests/test_library_contents.py

Nerdinacan:
Fixed a minor bug in "test_import_dataset_from_history".

This test used to break when the test context had other saved histories making it very state-dependent. Now it explicitly selects the test history that was created as part of the test before moving on. 

The issue of cleaning up the other saved histories is probably worth another PR.


Nerdinacan:
This was just a quick test fix. Ready when you guys are.

martenson:
thanks for fixing this @Nerdinacan !

PR #4720, opened 2017-09-28
Add logging during for script integrity exceptions

hexylena:
xref https://github.com/galaxyproject/galaxy/pull/4604#discussion_r141545068

PR #6990, opened 2018-11-08
Don't mount external /tmp for tests.

jmchilton:
Nate wants the file busy errors back ;).


mvdbeek (approved):

mvdbeek:
xref https://github.com/galaxyproject/galaxy/pull/4720 -- we should be more robust to those by now, and the python 3 tests don't mount /tmp either.

PR #6993, opened 2018-11-08
Allow rules module override and refactor submodule imports

nuwang:
This PR allows the rules_module to be specified per destination, instead of only at the plugin level, allowing for greater flexibility WRT to where dynamic rules can come from.
For example:
```
        <destination id="pulsar_djr" runner="dynamic">
            <param id="function">cloudlaunch_pulsar_burst</param>
            <param id="rules_module">galaxycloudrunner.rules</param>
        </destination>
```
In addition, this PR also does the following:
1. Updates filesystem watchers to also watch for per destination rules_module changes
2. Refactors `galaxy.util.submodule` import handling, preferring importlib over `__import__` statements
3. Removes some code duplication in the job mapper by using galaxy.util.submodule
4. Adds a test for rules_module overriding

### DIFF
--- a/lib/galaxy/containers/__init__.py
+++ b/lib/galaxy/containers/__init__.py
@@ -22,7 +22,7 @@ from six import string_types, with_metaclass
 from six.moves import shlex_quote
 
 from galaxy.exceptions import ContainerCLIError
-from galaxy.util.submodules import submodules
+from galaxy.util.submodules import import_submodules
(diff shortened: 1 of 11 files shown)


mvdbeek:
@galaxybot test py3

jmchilton:
Awesome, this looks good to me thanks!

PR #6994, opened 2018-11-09
Replaced jStorage with store

Nerdinacan:
jStorage is giving me some grief in another branch. 

Ostensibly It's just a local storage library... but it was written as a plugin for jQuery, which is kind of like hiring a tax attorney for your cat. It's also poorly-written and problematic to import with webpack due to basic incompetence issues and frequent references to window.$ instead of a passed jQuery parameter in a module wrapper.

Oh yeah! We also, only use it in about 1 place, so we get all the hassle with virtually no payoff. 

(Ok I'm done venting.)

[Store](https://www.npmjs.com/package/store) is a standard npm multi-storage wrapper conforming to the normal storage javascript API, works with local/session/whatever. It's basic. Like all xStorage wrappers it's barely necessary, but still offers backwards compatibility for older browsers while conforming to a modern javascript API. 

Its biggest selling point for me is that it's not bolted Frankenstein-style to the side of jQuery.

I'm open to different libs if somebody has strong yearnings, so long as it isn't tied to jQuery or Backbone or some other travesty, and more or less conforms to the standard [Web Storage API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API)


Nerdinacan:
Anybody know how many chickens I have to sacrifice to get that last Travis job to pass? It's passing with this exact build over on my Travis instance, and locally, but chokes here. Seems strange that my little client-side change would affect a server-side unit test
.
https://travis-ci.org/Nerdinacan/galaxy/builds/452701586

martenson:
@Nerdinacan it seems that was the right number of chickens. Save the rest.

dannon:
To follow up here, we're actually removing this completely and using localstorage/etc directly as we do in the rest of the app.  Since this has already been merged, we'll follow up in subsequent PRs to finish it.

PR #4857, opened 2017-10-25
A new runner for DRMAA (currently UNIVA)

bernt-matthias:
Reimplementation of the DRMAA runner inspired by the SLURM runner.
Currently tested only for the UNIVA grid engine.

This solves the problem that the current DRMAAJobRunner does
not work when jobs are submitted as real user (because jobs that are
started in a different drmaa session can not be accessed from the
session that is open in galaxy):
- this is done by resorting to command line tools qstat and qacct if
the drmaa library can not be used to check the job status and to get run
time information.
- this has the additional advantage that if the drmaa library
functions are not working (DRMAAJobRunner had implemented a repeated
checking to handle this problem) the runner can still use the command
line tools.

Furthermore (in contrast to the original drmaa runner) the new one
tests for run time and memory violations:
- memory violations are determined by checking the stderr output and
by comparing the used and the requested memory
- run time violations are determined by checking the signal that
killed the job and by comparing the used and the requested run time
Where the used memory and time are determined with drmaa.wait() or
qacct

Open (or better perspective):
- adaptions to other grid engines. the current implementation (the
  command line calls and result parsing) might be specific for the
  Univa grid engine. to include other GEs one could determine the
  GE (+ version) and make the calls and result parsing depending
  on this.

In addition that the job template is loaded in the external runner before
the user is changed. otherwise the cluster_files_directory needs to be
readable by all users.

PR #6995, opened 2018-11-09
external_runner: load json template first

bernt-matthias:
If the JSON is loaded first the cluster_files_directory does not need to be accessible by the real user (ie all users). 

This change is from https://github.com/galaxyproject/galaxy/pull/4857. I decided to split this PR (a few) parts, which makes it easier for me to manage -- and might speed up integration.


mvdbeek (approved):
Thanks, looks good!

Issue #4308, opened 2017-07-12
DRMAAJobRunner as real user fails when jobs are resubmitted


bernt-matthias:
For resubmission (I guess only the automatic resubmission due to run time or memory violations) the owner ship of the job directory is not changed back to the galaxy user. Which leads to an error. 

```
Traceback (most recent call last):
  File "/gpfs1/data/galaxy_server/galaxy-dev/lib/galaxy/jobs/runners/__init__.py", line 366, in _handle_runner_state
    handler(self.app, self, job_state)
  File "/gpfs1/data/galaxy_server/galaxy-dev/lib/galaxy/jobs/runners/state_handlers/resubmit.py", line 75, in failure
    _handle_resubmit_definitions(resubmit_definitions, app, job_runner, job_state)
  File "/gpfs1/data/galaxy_server/galaxy-dev/lib/galaxy/jobs/runners/state_handlers/resubmit.py", line 116, in _handle_resubmit_definitions
    job_state.job_wrapper.clear_working_directory()
  File "/gpfs1/data/galaxy_server/galaxy-dev/lib/galaxy/jobs/__init__.py", line 937, in clear_working_directory
    shutil.move( self.working_directory, arc_dir )
  File "/global/apps/bioinf/galaxy/bin/Python-2.7.13/lib/python2.7/shutil.py", line 299, in move
    copytree(src, real_dst, symlinks=True)
  File "/global/apps/bioinf/galaxy/bin/Python-2.7.13/lib/python2.7/shutil.py", line 208, in copytree
    raise Error, errors
Error: [('/gpfs1/data/galaxy_server/galaxy-dev/jobs_dir/000/76/working/core.104128', '/gpfs1/data/galaxy_server/galaxy-dev/jobs_dir/_cleared_contents/000/76/20170712-133512/working/core.104128', "[Errno 13] Permission denied: '/gpfs1/data/galaxy_server/galaxy-dev/jobs_dir/000/76/working/core.104128'"), ('/gpfs1/data
/galaxy_server/galaxy-dev/jobs_dir/000/76/set_metadata_apjSIM.py', '/gpfs1/data/galaxy_server/galaxy-dev/jobs_dir/_cleared_contents/000/76/20170712-133512/set_metadata_apjSIM.py', "[Errno 13] Permission denied: '/gpfs1/data/galaxy_server/galaxy-dev/jobs_dir/000/76/set_metadata_apjSIM.py'")]

```

[...]

PR #6996, opened 2018-11-09
reclaim ownership for failed jobs

bernt-matthias:
Fixes the first bug described in https://github.com/galaxyproject/galaxy/issues/4308. Note that the error described in the issue stems from `_handle_runner_state`. There are two ways the code may have ended up in this function: 

- `_handle_runner_state() <- finish_or_resubmit() <- finish_job()`
- `_handle_runner_state() <- fail_job()`

The former already calls `reclaim_ownership` in `finish_job`, but for `fail_job()` seems to be missing. 

Note that, I'm using the drmaa runner calling reclaim_ownership in `_handle_runner_state()` for a long time now (https://github.com/galaxyproject/galaxy/pull/4857), but as @jmchilton suggested this might not be the best place for the call (https://github.com/galaxyproject/galaxy/pull/4857#discussion_r182831418) -- also the `reclaim_ownership` would be called twice in when `finish_job()` is called.  

Still a bit unsure if this is the optimal place for reclaiming owner ship, but certainly ownership needs to be changed for all jobs.


bernt-matthias:
Given the CI errors, the place might really be wrong..

jmchilton (approved):

jmchilton:
👍 from me, I'll merge after these tests run.

PR #6997, opened 2018-11-09
drmaa: soft catch of external runner fails

bernt-matthias:
IMHO failures of the external runner should not doom the complete runner.

### DIFF
--- a/lib/galaxy/jobs/runners/drmaa.py
+++ b/lib/galaxy/jobs/runners/drmaa.py
@@ -208,6 +208,9 @@ class DRMAAJobRunner(AsynchronousJobRunner):
             filename = self.store_jobtemplate(job_wrapper, jt)
             self.userid = pwent[2]
             external_job_id = self.external_runjob(external_runjob_script, filename, pwent[2]).strip()
+            if external_job_id is None:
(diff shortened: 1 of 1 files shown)


lib/galaxy/jobs/runners/drmaa.py
```diff
@@ -208,6 +208,9 @@ def queue_job(self, job_wrapper):
             filename = self.store_jobtemplate(job_wrapper, jt)
             self.userid = pwent[2]
             external_job_id = self.external_runjob(external_runjob_script, filename, pwent[2]).strip()
+            if external_job_id is None:
```
mvdbeek, line 211: This doesn't do what you think it would.
`external_job_id is None` is only `True` if `external_job_id ` is `None`,  if `external_job_id ` is an empty string you wouldn't enter this condition. If `self.external_runjob(external_runjob_script, filename, pwent[2])` does return `None` you will get an exception because `None` doesn't have a `strip()` function, so effectively you can never enter this `if` branch.
My guess is that you want `if not external job:`

bernt-matthias: Correct. I changed `self.external_runjob` to return already stripped stdio.

PR #6998, opened 2018-11-09
Change join two dataset output format to tabular

qiagu:
Try to alleviate datatype sniffing issue https://github.com/galaxyproject/usegalaxy-playbook/issues/46

Now the datatype is sniffed to either `tabular` or `interval` depending the order of two input datasets.


tools/filters/joiner.xml
```diff
@@ -85,7 +85,7 @@ ${json.dumps( __fill_options )}
    </param>
   </inputs>
   <outputs>
-     <data format_source="input1" name="out_file1" metadata_source="input1" />
+     <data auto_format="true" name="out_file1" />
```
mvdbeek: can you just make this `tabular` ? 
```suggestion
     <data format="tabular" name="out_file1" />
```

This would be correct in all instances, while joining subclasses of tabular, like gtf or gff are not valid gff or gtf.

qiagu: I was swinging between a fixed one, `tabular`, and auto. Other people also mentioned `tabular` in previous discussion. There will still be discrepancy existing though as uploaded dataset could be sniffed to other format, like `interval`, by autodetect. Anyway, I'm totally fine with `tabular`. @jennaj  what do you think?

qiagu: @mvdbeek  I made the change. There is another issue associated with this one. I'll have it in a try. Please keep this PR open. Thanks.

nekrut:
@mvdbeek @qiagu ready to merge?

mvdbeek:
Looks fine to me, but I understood there's something else still going on ?

qiagu:
We were planning to make the `sniff` button for datatypes. That will take a lot of time. I plan to work on it later this week, but I'm open to discuss whether we should put that in priorities.

mvdbeek (approved):

jennaj:
Do we plan to backport the change to at least 18.09? So it can be implemented on the public servers now?

PR #6975, opened 2018-11-06
Use virtualenv when conda is on PATH

mvdbeek:
This uses our proper pinned requirements and is significantly easier to set up and maintain.

PR #6999, opened 2018-11-09
Fix several typos in admin docs.

dannon:
xref comments in https://github.com/galaxyproject/galaxy/pull/6975, with two bonus typo as well.


martenson:
Thanks @dannon for picking it up.

PR #7000, opened 2018-11-10
App singleton 2

Nerdinacan:
I force pushed a branch over the same name, this is a test to see if that is the reason for the screwy problems with the main PR.


Nerdinacan:
Nope, still failing. closing this little experiment.

PR #7005, opened 2018-11-12
Replace global Galaxy references with galaxy singleton accessor/setter

Nerdinacan:
Previously: https://github.com/galaxyproject/galaxy/pull/6779

This is the same code as in app_singleton, but the selenium tests are not running on the original PR. (They fail on jenkins, but run clean locally, and I cannot recreate the error.) I suspect something happened when I force-pushed a later rebuild of the feature. I only did that to preserve the name of the branch and the associated PR.

This is the same code organized into tighter commits so it's easier to review, and built directly off a recent copy of dev, so let's see if it passes selenium.

Instead of referring to Galaxy globally or through window.Galaxy, please write all future client-side code using the following functions:

```javascript
import { getGalaxyInstance, setGalaxyInstance } from "app";
import { getAppRoot } from "onload/loadConfig";


// Return the current galaxy app object

let Galaxy = getGalaxyInstance();

// Will set galaxy singleton instance, should only be done in new entry points
// The parameter to setGalaxyInstance is a factory function which will receive
// the GalaxyApp constructor as an argument

let Galaxy = setGalaxyInstance(GalaxyApp => {
     return new GalaxyApp(someOptions, bootstrappedStuff);
});

// If all you need is the galaxy root, please use the special getAppRoot function
// In future revisions, we will be making a distinction between configuration
// variables and application state

let galaxyRoot = getAppRoot();
```

If you need to debug legacy code which is still accessing window.Galaxy or Galaxy, please use the monitor functions I installed in this PR:
https://github.com/galaxyproject/galaxy/pull/6942


Nerdinacan:
We were just able to attribute the selenium test failures to differences between the production and development versions of the client as produced by webpack.

In particular I was doing some sloppy type-checking using constructor.name instead of the more appropriate instanceof operator. (When the code was uglified constructor.name no longer matched the expectation.)

Nerdinacan:
This is finally ready whenever it can get reviewed.

dannon:
@Nerdinacan Nice, I'll check it out tonight!

client/galaxy/scripts/app/galaxy.js
```diff
@@ -292,9 +298,6 @@ GalaxyApp.prototype.toString = function toString() {
     return `GalaxyApp(${userEmail})`;
 };
 
-// This is not great.  Rework to be a singleton-style Galaxy app accessible/created by import
```
dannon, line 295: Nice to check this one off the list!

Nerdinacan (commented):
getAppRoot is actually a function. Making it one delays execution to the point that its value is required to avoid having semi-global constants. (Which is the biggest problem with Backbone classes)

martenson:
@Nerdinacan great stuff, thanks for this!

PR #7006, opened 2018-11-12
Added support for chaining dynamic destinations

nuwang:
Currently, a dynamic destination cannot in turn return another dynamic destination. This PR adds support for chaining dynamic destinations, allowing a more compositional approach to configuring dynamic destinations.
e.g.
```
    <destinations default="burst_if_queued">
        <destination id="burst_if_queued" runner="dynamic">
            <param id="type">burst</param>
            <param id="from_destination_ids">local</param>
            <param id="to_destination_id">burst_if_size</param>
            <param id="num_jobs">2</param>
            <param id="job_states">queued</param>
        </destination>
        <destination id="burst_if_size" runner="dynamic">
            <param id="function">to_destination_if_size</param>
            <param id="rules_module">galaxycloudrunner.rules</param>
            <param id="max_size">1g</param>
            <param id="to_destination_id">pulsar</param>
            <param id="fallback_destination_id">local</param>
        </destination>
    </destinations>
```
will first check whether to burst through the built-in `burst` stock rule, and then channel it to `burst_if_size`, a custom rule which will in turn conditionally route it to a remote pulsar node if the input size is below a certain threshold, or route it to `local` if not.


afgane:
Looks like the test failure is relevant here. I reran the test wondering if it might have been a server load issue but the result was the same both times:
```
test/integration/test_job_resubmission.py::JobResubmissionIntegrationTestCase::test_condition_any_failure Traceback (most recent call last):
  File "/galaxy_venv/local/lib/python2.7/site-packages/paste/httpserver.py", line 1090, in process_request_in_thread
Build timed out (after 80 minutes). Marking the build as aborted.
Build was aborted
```

A corner case with recursion perhaps?

nuwang:
Thanks. I didn't realise that the cached_job_destination was in fact a public property, and not just an internal cache value, and so the minor refactoring I did to cache handling broke it. I've documented that it is in fact, a public property and added a test to prevent a future regression.

mvdbeek:
Can you add the example in the PR description to `job_conf.xml.sample_advanced` ?
Bonus points are up for grabs if you could set up an integration test case that exercises the chaining.
We already have an integration test for job resubmissions in https://github.com/galaxyproject/galaxy/blob/626195335ee70dbc2d0568603e0627744d07d529/test/integration/test_job_resubmission.py
That test is completely optional of course, the unit test is already good.

nuwang:
Thanks for the review Marius. I've made the modifications. I used the tmpdir as a proxy for determining whether the chain had been traversed because I couldn't think of any other way, is that ok?

config/job_conf.xml.sample_advanced
```diff
@@ -652,6 +652,30 @@
...
+            <param id="num_jobs">2</param>
+            <param id="job_states">queued</param>
+        </destination>
+        <destination id="burst_if_size" runner="dynamic">
+            <param id="type">python</param>
+            <param id="function">to_destination_if_size</param>
+            <!-- Also demonstrates a destination level override of the
+            rules_module. This rules_module will take precedence over the
+            plugin level rules module when resolving the dynamic function -->
+            <param id="rules_module">galaxycloudrunner.rules</param>
```
nuwang, line 674: I just realised that this should actually be a part of PR: https://github.com/galaxyproject/galaxy/pull/6993, but this seems like an appropriate place to document both. Should I leave this in, or wait till that's reviewed? Also, we should be able to modify this existing integration test for chained rules to also test rules_module overriding in the same go, if that's ok?

jmchilton:
Merged that PR if you want to rebase?

nuwang:
Thanks John. Have rebased and modified the chained destination integration test to also test that rules_module overriding happens as expected.

config/job_conf.xml.sample_advanced
mvdbeek: Sure, that would be great.

test/integration/test_chained_dynamic_destinations.py
```diff
@@ -0,0 +1,32 @@
+"""Integration tests for the Pulsar embedded runner."""
```
mvdbeek:
```suggestion
"""Integration tests for chained dynamic job destinations."""
```

mvdbeek (approved):
Fantastic work, I think this has been requested a couple of time before, that'll be very helpful.

test/integration/test_chained_dynamic_destinations.py
nuwang: Thanks for catching that!

nuwang:
Thanks for the feedback everyone! All tests are passing now, so it seems good to go.

PR #7007, opened 2018-11-13
Fix handling of forgotten LSF jobs in cli

pcm32:
Previous case assumed that "is not found" was part of the STDOUT when in fact is in the STDERR (and STDOUT is empty).

### DIFF
--- a/lib/galaxy/jobs/runners/util/cli/job/lsf.py
+++ b/lib/galaxy/jobs/runners/util/cli/job/lsf.py
@@ -81,8 +81,13 @@ class LSF(BaseJobExec):
         return rval
 
     def parse_single_status(self, status, job_id):
-        if "is not found" in status:
+        if not status:
             # Job not found in LSF, most probably finished and forgotten.
+            # lsf outputs: Job <num> is not found -- but that is on the stderr
+            # Note: a very old failed job job will not be shown here either,
+            # which would be badly handled here. So this only works well when Galaxy
+            # is constantly monitoring the jobs. The logic here is that DONE jobs get forgotten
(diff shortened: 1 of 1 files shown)


lib/galaxy/jobs/runners/util/cli/job/lsf.py
```diff
@@ -81,8 +81,13 @@ def parse_status(self, status, job_ids):
...
 
     def parse_single_status(self, status, job_id):
-        if "is not found" in status:
+        if not status:
             # Job not found in LSF, most probably finished and forgotten.
+            # lsf outputs: Job <num> is not found -- but that is on the stderr
+            # Note: a very old failed job job will not be shown here either,
+            # which would be badly handled here. So this only works well when Galaxy
+            # is constantly monitoring the jobs. The logic here is that DONE jobs get forgotten
+            # faster than failed jobs.
```
mvdbeek, line 90: That'll make other status parsing issues very hard to debug though. I assume this doesn't happen very often, maybe log a warning here ?
```suggestion
            # faster than failed jobs.
            log.warning("Job id '%s' not found LSF status check" % job_id)
```

pcm32: Yes, good suggestion! I guess it will happen often if you turn off and on the instance frequently with large pauses in the middle and you have jobs that change status in between (like in a development env). It will also depend as well on the settings of the local LSF (ie. how quickly it forgets DONE jobs). But besides your suggestion, I'm not sure there is much to be done, unless that outright the jobs would be set to failed.

mvdbeek (approved):

PR #7018, opened 2018-11-15
Add datatype sniff button to edit dataset attributes page

qiagu:


lib/galaxy/webapps/galaxy/controllers/dataset.py
```diff
@@ -18,6 +18,7 @@
     web
 )
 from galaxy.datatypes.display_applications.util import decode_dataset_user, encode_dataset_user
+from galaxy.datatypes import sniff
```
nsoranzo, line 20: This needs to be moved one line above, see our import order style: https://github.com/PyCQA/flake8-import-order/blob/master/tests/test_cases/complete_smarkets.py

lib/galaxy/webapps/galaxy/controllers/dataset.py
```diff
@@ -459,6 +461,24 @@ def __ok_to_edit_metadata(dataset_id):
...
+                    is_binary = check_binary(path)
+                    datatype = sniff.guess_ext(path, trans.app.datatypes_registry.sniff_order, is_binary=is_binary)
+                    trans.app.datatypes_registry.change_datatype(data, datatype)
+                    trans.sa_session.flush()
+                    trans.app.datatypes_registry.set_external_metadata_tool.tool_action.execute(
+                        trans.app.datatypes_registry.set_external_metadata_tool, trans, incoming={'input1': data},
+                        overwrite=False)  # overwrite is False as per existing behavior
+                    message = 'Sniffed and changed the type to %s.' % datatype
+            else:
+                return self.message_exception(trans, 'Changing datatype %s is not allowed.' % (data.extension))
```
nsoranzo:
```suggestion
                return self.message_exception(trans, 'Changing datatype "%s" is not allowed.' % (data.extension))
```

qiagu: @nsoranzo Thanks

client/galaxy/scripts/mvc/dataset/dataset-edit-attributes.js
```diff
@@ -172,6 +172,14 @@ var View = Backbone.View.extend({
                     onclick: function() {
                         self._submit("datatype", form);
                     }
+                }),
+                submit_datatype_sniff: new Ui.ButtonIcon({
+                    tooltip: _l("Sniff the datatype and change it."),
+                    title: _l("Sniff datatype"),
```
martenson: Can we go with something like 'Detect datatype' instead? Sniff could be confusing.

qiagu: Absolutely. Changed every 'sniff' to 'detect'. Thanks.

qiagu:
@galaxybot test this

jennaj:
Passes tests but presented a display problem for me in a default local inside a virtualenv. 

I'm not sure why. Does python need to be updated so the virtualenv is created correctly?

This is how the display problem presented: https://github.com/galaxyproject/usegalaxy-playbook/issues/46#issuecomment-439551830

jennaj:
Passes tests in a local using this branch. Details: https://github.com/galaxyproject/usegalaxy-playbook/issues/46#issuecomment-443334407

This we should merge

qiagu:
@jennaj Thanks!

martenson:
@qiagu looks very good, thanks!

qiagu:
@martenson  Thanks!

PR #6850, opened 2018-10-10
Track workflow step input definitions in our model.

jmchilton:
We don't track workflow step inputs in any formal way in our model currently. This has resulted in some current hacks and prevents future enhancements. This commit splits WorkflowStepConnection into two models WorkflowStepInput and WorkflowStepConnection - normalizing the previous table workflow_step_connection on input step and input name.

In terms of current hacks forced on it by restricting all of tool state to be confined to a big JSON blob in the database - we have problems distinguishing keys and values when walking tool state. As we store more and more JSON blobs inside of the giant tool state blob - the worse this problem gets. Take for instance checking for runtime parameters or the rules parameter values - these both use JSON blobs that aren't simple values, so it is hard to tell looking at the tool state blob in the database or the workflow export to tell what is a key or what is a value. Tracking state as normalized inputs with default values and explicit attributes runtime values should allow much more percise state definition and construction.

This variant of the models would also potentially allow defining runtime values with non-tool default values (so default values defined for the workflow but still explicitly settable at runtime). The combinations of overriding defaults and defining runtime values were not representable before.

In terms of future enhancements, there is a lot we cannot track with the current models - such as map/reduce options for collection operations (https://github.com/galaxyproject/galaxy/issues/4623#issuecomment-389544980). This should enable a lot of that. Obviously there are a lot of attributes defined here that are not yet utilized, but I'm using most (all?) of them downstream in the CWL branch. I'd rather populate this table fully realized and fill in the implementation around it as work continues to stream in from the CWL branch - to keep things simple and avoid extra database migrations. But I understand if this feels like speculative complexity we want to avoid despite the implementation being readily available for inspection downstream.

PR #7019, opened 2018-11-15
File-based Workflow Editor

jmchilton:
- Allow workflows to be uploaded via path using the GUI and and API for admins.
- Track the path of such uploads and re-sync workflows on save (both .ga and format 2 workflows) - this will allow Planemo to leverage the Galaxy workflow editor to interactively save workflows (e.g. ``planemo workflow_edit <workflow.(ga|yml)>``).

Builds on #6850

### DIFF
--- a/test/base/populators.py
+++ b/test/base/populators.py
@@ -556,6 +556,14 @@ class BaseWorkflowPopulator(object):
         workflow = self.load_workflow(name)
         return self.create_workflow(workflow, **create_kwds)
 
+    def import_workflow_from_path(self, from_path):
+        data = dict(
+            from_path=from_path
+        )
+        import_response = self._post("workflows", data=data)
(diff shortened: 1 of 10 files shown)


test/base/populators.py
```diff
@@ -625,6 +633,21 @@ def download_workflow(self, workflow_id, style=None):
         api_asserts.assert_status_code_is(response, 200)
         return response.json()
 
+    def update_workflow(self, workflow_id, workflow_object):
+        data = dict(
+            workflow=workflow_object
+        )
+        raw_url = 'workflows/%s' % workflow_id
+        # url = self.galaxy_interactor._api_url(raw_url, use_key=True)
```
bgruening: This line should be removed?

martenson:
Seems to work as expected in my testing, thanks @jmchilton

PR #7020, opened 2018-11-16
Expose published workflows through API

frederikcoppens:
Published workflows are already exposed anonymously through https://usegalaxy.org/workflows/list_published, but the API endpoint /api/workflows requires authentication. This PR is to be able to list all published workflows anonymously through the API

I exposed /api/workflows anonymous and sessionless. The user authentication is respected, showing only published workflows. 
The show method required an extra change to deal with the fact that user is None.

I changed the default for show_published to True to avoid having to add /?show_published=True

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/workflows.py
+++ b/lib/galaxy/webapps/galaxy/api/workflows.py
@@ -123,7 +123,7 @@ class WorkflowsAPIController(BaseAPIController, UsesStoredWorkflowMixin, UsesAnn
         :param  missing_tools:       if True, include a list of missing tools per workflow
         :type   missing_tools:       boolean
         """
-        show_published = util.string_as_bool(kwd.get('show_published', 'False'))
+        show_published = util.string_as_bool(kwd.get('show_published', 'True'))
         missing_tools = util.string_as_bool(kwd.get('missing_tools', 'False'))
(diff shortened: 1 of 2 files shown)


bgruening:
Thanks for working on this @frederikcoppens!

nsoranzo:
It would be nice to have a test :wink:

nsoranzo:
@galaxybot test this

jmchilton:
I think we need to change the default of show_published back to False, I'd be open to having it default to true if ``trans.user is None`` though to preserve the old behavior and get your desired behavior in the sessionless case.

I can add a test also @nsoranzo - give me some time.

nsoranzo:
> I think we need to change the default of show_published back to False, I'd be open to having it default to true if trans.user is None though to preserve the old behavior and get your desired behavior in the sessionless case.

Completely agree on this, I was going to suggest the same.

frederikcoppens:
Implemented the default to True if user is None, False otherwise. Not sure where this behaviour needs to be documented.

jmchilton:
@galaxybot test this

jmchilton:
Perfect, I pushed a test case and merged! Thanks for the contribution @frederikcoppens  - another great Biohackathon outcome.

PR #7027, opened 2018-11-19
Use solid background for text field autocomplete options

guerler:
Thanks for reporting @erasche.


hexylena (approved):
grafik

This is an improvement, though I think the original Looked a bit nicer?

photo_2018-11-20_10-01-14

guerler:
@erasche, I revised this PR by restoring the jquery style import.

hexylena:
grafik

Oh yes, that looks much better, way more readable! Thanks!

PR #7029, opened 2018-11-20
Follow-up on #6850

nsoranzo:
Only migrate distinct (input_step_id, input_name) pairs, otherwise the UNIQUE constraint on the `workflow_step_input` table is violated.

Reported by @mvdbeek.

Also a small simplification of SQLAlchemy session handling in 2 files (found while trying to run the migration step in a transaction and failing).


mvdbeek (approved):
Thanks @nsoranzo!

jmchilton:
Thanks!

PR #7031, opened 2018-11-20
Provide element_identifier attribute in DatsetCollectionWrapper

mvdbeek:
This allows accessing `$input1.element_identifier` for an `input1` of
collection_type `list:pair`.

I think we do want to encourage all tool authors to use `element_identifier` always and everywhere, and this should help, @bernt-matthias brought this up on gitter.


jmchilton (approved):
Thanks @mvdbeek, I'd be +1 on a backport of this as a bug fix as well. Dev is a good start though.

bernt-matthias:
maybe mildly related: https://github.com/galaxyproject/galaxy/issues/7034

PR #7036, opened 2018-11-22
Fix error when building images 

scholtalbers:
The default was passed as a string that got joined by `','` resulting in `-c 'c' -c 'o' -c 'n'` etc.


nsoranzo:
Thanks @scholtalbers , looks good to me! I've opened https://github.com/galaxyproject/galaxy/pull/7038 , a backport of this for Galaxy 18.09, where this problem was first introduced in commit b99fedefa81b3ca59c27eb64ea9a7e9865cd9db8 . I've added also another small related bugfix there.

martenson:
thanks @scholtalbers !

PR #7047, opened 2018-11-27
Replace login mako with Vue component

guerler:
This replaces #6413. This PR replaces the login mako with a vue component and connects it with api-like login endpoint in the user controller for Galaxy. It also replaces the change password, and forgotten password options, and the open id login form. Legacy code and makos have been moved to the toolshed. It simplifies the upcoming replacement of the registration mako.


guerler:
@galaxybot test this.

Nerdinacan:
I like the look of this Sam, but there's a couple formatting suggestions I'd like to offer. We can do these later if you like, but I'd like to float them here to see what you think.

Firstly, I think any important component should have its own directory and be built as if it were a small module (which it really is). I like to separate their chunks using the src attribute that's available in vue syntax as follows:

```
<!--- components/LoginFormOrWhateverYouNameIt/index.vue -->
<template>
    <!-- your markup goes here -->
</template>
<style lang="scss" src="./styles.scss"></style>
<script src="./YourLoginComponentObject.js"></script>
```

I've already enabled index.vue as a module index in webpack so you can import this component exactly the same way you currently do, even when it's broken up:

```
import Login from "components/LoginFormOrWhateverYouNameIt"; // still works
```

The main advantage here is ease of unit testing your Vue script since your Vue controller is just a plain javascript object now. Putting the component in its own folder also makes an obvious place for me to look for the unit tests for that component.

Secondly, I think it's important for us to start separating concerns, in this instance by separating out the data calls into their own file. As I've been refactoring the existing code it's blatantly apparent that people here have been making the same maintainability mistakes that developers everywhere make:

1. They don't separate concerns and therefore testing is hard.
2. Because testing is hard they don't do it.
3. They use conditional javascript to do what CSS can already do for free.

It's as easy as making a separate file in that folder and putting a couple exported functions in it for use in your login component

```
// components/Login/loginServices.js

import axios from "axios";

export function authenticate(username,password) {
   // return a promise for your component
}

export function resetPassword() {}

// etc
```

Now your component looks like this:

```
import { authenticate, resetPassword } from "./loginServices.js"

export default {

   /// stuff

    methods: {
        submit(method) {
              authenticate(username, password)
                  .then(doSomethingWithResult)
                  .catch(err => blah());
        }
    }
}
```

Now when we decide that login works differently the component doesn't care because you separated your concerns, and the unit tests that you are going to write for your data access services (which are now just plain simple javascript functions with minimal mocked dependencies) will continue to be relevant.

I realize this is probably overkill for a simple login component but because this component is so easy this is also the perfect opportunity to start generating maintainable, testable code with a solid example for the rest of the community.

Thoughts?

P.S. Also we could totally add all this after the fact. I'm just happy we're writing components. :)

dannon:
We've talked about it a good bit and I wholeheartedly agree with the second suggestion @Nerdinacan has regarding separation of concerns.  Building out modules that contain interfaces/models for our single file components to use is something we're working on and a lot of this logic should *leave* SFCs completely and be tested elsewhere.

I don't agree with the first point, though.  It seems onerous and an abuse of what SFCs are intended to be.  The concern is *testing*, right?  Let's use SFCs as they are intended, single file components, and test them like this:  https://vue-test-utils.vuejs.org/guides/#testing-single-file-components-with-mocha-webpack

This, coupled with a lot of this logic migrating *out* of the SFCs into other, separately unit tested layers, seems a lot more appropriate.

Nerdinacan:
I can live with that so long as we start separating data services, and unit testing our components and services as a habit. If you're going to jam everything into one file, that file better be small. :)

dannon:
> I can live with that so long as we start separating data services, and unit testing our components and services as a habit. If you're going to jam everything into one file, that file better be small. :)

Totally agreed.  SFCs should contain *only* inherently coupled code/style, which should be very small as we continue to move functionality to other (independently unit tested) interfaces.

guerler:
Ok, cool sounds good.

dannon:
Resolved conflicts and merged locally after testing.  There's some future work here I think (styling of the registration form (which is in a separate related PR) and making sure all that works together smoothly, but it's a nice step forward and it'll be good to go ahead and get this in dev.

PR #7049, opened 2018-11-27
doc: updated error parsing docs

bernt-matthias:
according to https://github.com/galaxyproject/galaxy/pull/6977


martenson:
Thanks @bernt-matthias !

PR #7052, opened 2018-11-28
Add styled loader vue component

inkuzmin:
Hi! This is a loader (wrapped as a Vue component) developed during the short hacking session after the EGD2018 in Freiburg.

In its core it is HTML + CSS with no JS, it could be checked at https://jsfiddle.net/inkuzmin/pnthykug/

Vue wrapper provides preliminary API for different use cases I came up with. Component could be initialised with a few different predefined styles and in a range of sizes.

I didn't find a good place to add it on my local Galaxy installation (everything is too fast), but perhaps it may be useful as a spinner for some slow transitions I haven't noticed?

### DIFF
--- a/client/galaxy/scripts/components/GalaxyLoader.vue
+++ b/client/galaxy/scripts/components/GalaxyLoader.vue
@@ -1,0 +1,155 @@
+<template>
+    <div class="galaxy-loader"
+         :class="[`galaxy-loader_${style}`, {'galaxy-loader_center': center}]"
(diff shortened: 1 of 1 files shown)


martenson:
@inkuzm very nice! We should be able to make a good use of it on Main and other big servers.

bgruening:
Yeah pretty cool! Workflow loading could be one target.

martenson:
thanks again @inkuzm

PR #7053, opened 2018-11-28
Added resubmit handlers on Pulsar connection failures

nuwang:
Currently, the Pulsar runner does not have support for resubmit handlers. This PR adds support for resubmit handlers to the Pulsar runner if an unknown error occurs during job preparation or finalisation. For example:
```
        <destination id="initial_pulsar" runner="pulsar_rest">
            <param id="url">https://<some_pulsar_server>:8913/</param>
            <param id="private_token">some_token</param>
            <resubmit condition="unknown_error" destination="local" />
        </destination>
```

In addition, it also introduces a referrer parameter for dynamic destinations. The referrer parameter is useful for finding out the resubmit conditions of a dynamic destination's referrer, or any of its parameters.
For example:
```
        <destination id="galaxycloudrunner" runner="dynamic">
            <param id="function">cloudlaunch_pulsar_burst</param>
            <param id="fallback_destination">local</param>
            <resubmit condition="unknown_error and attempt &lt;= 7" destination="galaxycloudrunner" />
        </destination>
```

With the referrer parameter being accessible within the dynamic function as:
```
def cloudlaunch_pulsar_burst(app, referrer):
    print(referrer.id) // should print galaxycloudrunner
```


jmchilton:
Can reproduce the test error locally, trying to figure it out though...

nuwang:
Thanks for looking into this. I was hoping that you could help, because I can't immediately see a connection between the changes and why the tool output is empty, as the resubmission works as expected when I run against the [galaxycloudrunner](https://galaxycloudrunner.readthedocs.io).

jmchilton:
I think the problem is finish being called there? I can see you getting a command_line in production but the transport failing so that you'd get a chance to call fail_job without finish. Regardless of the specifics, I feel like the approach I pushed is a bit more uniform and it seems to cause the now simplified test of resubmission to pass. Locally this test takes a minute or two to pass though - I'm a little concerned about including it in the suite of tests that run with each PR for that reason. 🤔

nuwang:
@jmchilton I totally missed that - thanks for tracking it down. Perhaps the test will run faster if we use an invalid dns name instead of a reserved IP perhaps? I can try that next.

jmchilton:
I'm just going to assume this works, if it is broken in production it should be close - hopefully it is just a small bug fix PR to fix it. galaxycloudrunner looks awesome, keep me updated on the amazing things you do with it!

PR #7060, opened 2018-11-29
[18.09] Eliminate dependency on make for automatic client build

dannon:
xref gitter this morning, @olisteadman mentioned that the client build blew up due to lacking make.

It's pretty historical at this point -- it was useful before because we relied on cli-based grunt/webpack, but I've since pushed all the logic into the package scripts executed via yarn, which we install automatically.  This shifts the client auto-building logic to prefer using yarn directly, eliminating the dependency on make.  I left the make targets because they're handy shortcuts for developers, but we should prefer the actual package scripts here.

Also retabs (to spaces) a block earlier in the file where I saw tabs mixed in.

### DIFF
--- a/scripts/common_startup.sh
+++ b/scripts/common_startup.sh
@@ -241,10 +241,16 @@ if [ $SKIP_CLIENT_BUILD -eq 0 ]; then
     fi
 
     # Build client
-    if ! make client-production-maps; then
-        echo "ERROR: Galaxy client build failed. See ./client/README.md for more information, including how to get help."
-        exit 1
+    pushd client
+    if yarn install --network-timeout 120000 --check-files; then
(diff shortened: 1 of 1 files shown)


jmchilton (approved):

jmchilton (commented):
https://travis-ci.org/galaxyproject/galaxy/jobs/461359144 Failure looks related - pushd not available.

dannon:
@jmchilton Argh, thanks for noticing that, I'll try to make this even more vanilla sh.

nsoranzo (approved):

jmchilton:
Thanks @dannon !

dannon:
Thanks for the merge!

PR #7054, opened 2018-11-28
[18.09] add new item to the help menu

martenson:
part of https://github.com/galaxyproject/usegalaxy-playbook/issues/179

PR #7061, opened 2018-11-29
[18.09] make the menu transition non-instrusive

martenson:
followup to https://github.com/galaxyproject/galaxy/pull/7054, sorry for the extra PR

@jmchilton @bgruening

### DIFF
--- a/config/galaxy.yml.sample
+++ b/config/galaxy.yml.sample
@@ -698,7 +698,7 @@ galaxy:
   #logo_url: /
 
   # The URL linked by the "Galaxy Help" link in the "Help" menu.
-  #helpsite_url: 'https://help.galaxyproject.org/'
+  #helpsite_url: ''
 
   # The URL linked by the "Wiki" link in the "Help" menu.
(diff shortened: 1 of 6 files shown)


config/galaxy.yml.sample
```diff
@@ -1358,9 +1358,6 @@ galaxy:
   # of Galaxy's stable API.
   #enable_beta_workflow_modules: false
 
-  # Enable import and export of workflows as Galaxy Format 2 workflows.
-  #enable_beta_workflow_format: false
-
```
jmchilton, line 1363: Yeah - this should be rolled back also - must have been generated against dev.

jmchilton (approved):

config/galaxy.yml.sample
martenson: @jmchilton yep 😖

nsoranzo: In fact, it seems that a `make config-rebuild` was missed in https://github.com/galaxyproject/galaxy/pull/6776

nsoranzo (approved):

PR #7067, opened 2018-12-02
Improved parsing of sizes (bytes)

bernt-matthias:
Before floats without byte extension were not recognized (eg 1.0)

We could also drop the try except block, since the new regex covers all cases. Don't know if it is faster with the try except block.

### DIFF
--- a/lib/galaxy/util/__init__.py
+++ b/lib/galaxy/util/__init__.py
@@ -1383,6 +1383,7 @@ def size_to_bytes(size):
     else:
         raise ValueError("Unknown multiplier '%s' in '%s'" % (multiple, size))
 
+
 def send_mail(frm, to, subject, body, config, html=None):
     """
     Sends an email.
--- a/lib/galaxy/util/__init__.py
+++ b/lib/galaxy/util/__init__.py
@@ -1351,6 +1351,8 @@ def size_to_bytes(size):
 
     >>> size_to_bytes('1024')
     1024
+    >>> size_to_bytes('1.0')
+    1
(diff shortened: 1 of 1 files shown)


nsoranzo:
I'm fine with removing the try/except block, if you like.

Also, `assert size_match is not None` is not very useful since the method would fail at the next instruction (maybe replace with a `if size_match is None: raise Exception("Some useful message")` ?).

lib/galaxy/util/__init__.py
```diff
@@ -1360,16 +1362,16 @@ def size_to_bytes(size):
...
         pass
     # Otherwise it must have non-numeric characters
-    size_re = re.compile(r'([\d\.]+)\s*([eptgmk]b?|b|bytes?)$')
+    size_re = re.compile(r'(?P<size>[\d\.]+)\s*(?P<multiple>[eptgmk]?(b|bytes?)?)?$')
     size_match = size_re.match(size.lower())
     assert size_match is not None
-    size = float(size_match.group(1))
-    multiple = size_match.group(2)
-    if multiple.startswith('b'):
+    size = float(size_match.group("size"))
```
nsoranzo: Maybe call this `float_size` (also below) instead of reusing `size`, so that at the end we can add:
```python
    else:
        return size
```

bernt-matthias: agreed (made it `number`). But the else branch should raise an exception which I added (the case of a size without a multiplier is treated in the 1st if branch).

nsoranzo (approved):

PR #7070, opened 2018-12-03
downgrade bdbag, 1.5.0 requires python:>=2.7.9 

martenson:
which centos7 does not have

### DIFF
--- a/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
+++ b/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
@@ -20,7 +20,7 @@ azure-storage-nspkg==3.0.0
 babel==2.6.0
 bagit==1.7.0
 bcrypt==3.1.4
-bdbag==1.5.0
+bdbag==1.4.1
 beaker==1.10.0
 bioblend==0.11.0
 bleach==3.0.2


nsoranzo:
Centos 7 has python 2.7.5, see http://mirror.centos.org/centos/7/os/x86_64/Packages/

Would it make sense to comment on https://github.com/fair-research/bdbag/commit/86517b9ba89524c3e1328eea7f4537552f0af82e to ask the reason for the minimum patch version?

martenson:
@nsoranzo We have very light usage of that library, so I think downgrading is very low risk. That said contacting them might be a good thing to do community-wise.

nsoranzo:
@martenson I left a comment.

nsoranzo:
I'll follow up on this, the Pipfile needs to be pinned.

natefoo:
Oh, sorry, I didn't even notice this was in the generated requirements. Thanks @nsoranzo.

martenson:
And I have no idea how this works.

jmchilton:
Filing this under additional evidence our primary target platform should be Conda and vanilla Python should be supported but only with additional constraints - like having a minimum version and GCC. Hopefully we get a fix bdbag though - otherwise I will fork it.

Issue #999, opened 2015-10-28
NGS: RNA-seq: gffread > errors when loading tool form at http://usegalaxy.org


jennaj:
Click on the tool to see error: 

<img width="337" alt="screen shot 2015-10-28 at 10 39 57 am" src="https://cloud.githubusercontent.com/assets/8999735/10797517/43fdd02e-7d60-11e5-8792-252c887b22d6.png">

martenson:
also happens on test; the exception in the log is:

```
==> /srv/galaxy/test/log/uwsgi.log <==
galaxy.web.framework.decorators ERROR 2015-10-29 20:40:16,159 Uncaught exception in exposed API method:
Traceback (most recent call last):
  File "lib/galaxy/web/framework/decorators.py", line 258, in decorator
    rval = func( self, trans, *args, **kwargs)
  File "lib/galaxy/webapps/galaxy/api/tools.py", line 97, in build
    return tool.to_json(trans, kwd.get('inputs', kwd))
  File "lib/galaxy/tools/__init__.py", line 2391, in to_json
    populate_state(trans, self.inputs, state_inputs, state_errors, params.__dict__)
  File "lib/galaxy/tools/__init__.py", line 2267, in populate_state
    state[input.name] = input.get_initial_value(trans, context, history=history)
  File "lib/galaxy/tools/parameters/grouping.py", line 620, in get_initial_value
    current_case = self.get_current_case( test_value, trans )
  File "lib/galaxy/tools/parameters/grouping.py", line 569, in get_current_case
    raise ValueError( "No case matched value:", self.name, str_value )
ValueError: ('No case matched value:', 'gffs', None)
```

jennaj:
Duplicate? https://github.com/galaxyproject/galaxy/issues/980

martenson:
yes @jennaj , this is a duplicate of #980
closing

PR #777, opened 2015-09-23
Add docker command wrapper.

scholtalbers:
For example, to change the current group to docker, which is needed when the galaxy user belongs to (too) many groups.

PR #7072, opened 2018-12-04
Make tags more colourful

hexylena:
**EDIT**

After much discussion we've settled on [these Options](https://github.com/galaxyproject/galaxy/pull/7072#issuecomment-444218991). 

**Original Post**

Today I was running through the peaks2genes tutorial and found that my eyes struggled to recognise if I was looking at something tagged `#peaks` or `#genes`, and this wasn't great, they just all looked the same when scanning across a history. (I've added an example at the end.) So I've implemented coloured tags. The colours are done automatically from a hash of the string so they're stateless and consistent.

grafik

Whole UI:

grafik

It just hashes the tag, and then pulls out some bits of that number. I've done a `%10` which means that the lightest colour that can theoretically be generated is which doesn't have super amazing legibility but it isn't that bad (tags are already bad due to small size for that, this isn't making it that much worse.) 

grafik

I chose a hash function at random by googling, if there's something more appropriate (or already in codebase) let me know. You could probably optimise the colour distributions, I think the dark colours do not have a high perceptual difference but I'm not sure what to do about this. You could probably find a better mapping.

Light colours are also an option / nice?
grafik

Original | New
-- | --
These colours I found difficult when scanning my history | With this patch it's much easier to scan! (Faked screenshot because I'm lazy and didn't want to backport it to 18.05)
grafik | grafik

### DIFF
--- a/client/galaxy/scripts/utils/utils.js
+++ b/client/galaxy/scripts/utils/utils.js
@@ -378,38 +378,55 @@ function contrastingColor(r, g, b) {
     return (o > 125) ? 'black' : 'white';
 }
 
-export function hashToTagColor(tagValue) {
-    var hash = hashFnv32a(tagValue);
-    var r = ((hash & 0x00f) >> 0),
-        g = ((hash & 0x0f0) >> 4),
-        b = ((hash & 0xf00) >> 8);
+/**
+ * Converts an HSL color value to RGB. Conversion formula
+ * adapted from http://en.wikipedia.org/wiki/HSL_color_space.
+ * Assumes h, s, and l are contained in the set [0, 1] and
+ * returns r, g, and b in the set [0, 1].
+ *
+ * @param   {number}  h       The hue
+ * @param   {number}  s       The saturation
+ * @param   {number}  l       The lightness
+ * @return  {Array}           The RGB representation
+ */
+function hslToRgb(h, s, l){
+    var r, g, b;
 
-    // Pastels
-    var r2 = 15 - (r % 6),
-        g2 = 15 - (g % 6),
-        b2 = 15 - (b % 6);
+    if(s == 0){
+        r = g = b = l; // achromatic
+    }else{
+        var hue2rgb = function hue2rgb(p, q, t){
+            if(t < 0) t += 1;
+            if(t > 1) t -= 1;
+            if(t < 1/6) return p + (q - p) * 6 * t;
+            if(t < 1/2) return q;
+            if(t < 2/3) return p + (q - p) * (2/3 - t) * 6;
+            return p;
+        }
 
-    // All colours
-    //var r2 = r % 0xf,
-        //g2 = g % 0xf,
-        //b2 = b % 0xf;
+        var q = l < 0.5 ? l * (1 + s) : l + s - l * s;
+        var p = 2 * l - q;
+        r = hue2rgb(p, q, h + 1/3);
+        g = hue2rgb(p, q, h);
+        b = hue2rgb(p, q, h - 1/3);
+    }
 
-    // Not horribly dark
-    //var r2 = r % 0xa,
-        //g2 = g % 0xa,
-        //b2 = b % 0xa;
-
-    return [r2, g2, b2];
+    return [r, g, b];
 }
 
-export function generateTagColor(tag) {
-    var [r, g, b] = hashToTagColor(tag)
-    return `#${r.toString(16)}${g.toString(16)}${b.toString(16)}`;
-}
+export function generateTagStyle(tag) {
+    var hash = hashFnv32a(tag);
+    var hue = (hash >> 4) & 360;
+    //var lightnessOffset = 20; // dark
+    var lightnessOffset = 75; // pastel
+    var lightness = lightnessOffset + (hash & 0xf);
+    var bgColor = `hsl(${hue}, 100%, ${lightness}%)`;
+    var brColor = `hsl(${hue}, 100%, 30%)`; // darker
 
-export function generateTagColorFg(tag) {
-    var [r, g, b] = hashToTagColor(tag)
-    return contrastingColor(r / 0xf, g / 0xf, b / 0xf)
+    var [r, g, b] = hslToRgb(hue, 1.0, lightness / 100);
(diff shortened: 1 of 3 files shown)


hexylena:
Earlier screenshot with this patch (but faked because I'm lazy, but, it gets the point across.)

grafik

client/galaxy/scripts/utils/utils.js
```diff
@@ -349,6 +349,40 @@ export function setWindowTitle(title) {
...
+
+    for (i = 0, l = str.length; i < l; i++) {
+        hval ^= str.charCodeAt(i);
+        hval += (hval << 1) + (hval << 4) + (hval << 7) + (hval << 8) + (hval << 24);
+    }
+    return hval >>> 0;
+}
+
+export function generateTagColor(tagValue) {
+    //var r = (15 - ((hash & 0xf) % 6)).toString(16),
```
martenson: drop?

hexylena: This was the 'light' version, left it in so I would not forget what I did, but should be dropped before merge.

martenson:
Is it possible that these generated colors will be unreadable/colorblind unfriendly?

hexylena:
**Readability**

The lightest they can go is `#999`, as seen [here](https://user-images.githubusercontent.com/458683/49456914-c800d480-f7e1-11e8-81ba-8f74fbf336dc.png). This does not pass WAVE AA or AAA for small text. We would have to cap it `#777` for that. This is OK, but has the unfortunate side effect of becoming murkier colours and giving us a smaller set of colours to use.

grafik

Alternatively we can just eliminate the modulus operation, and use the [W3C's algorithm](https://www.w3.org/TR/AERT/#color-contrast) ([js version](https://stackoverflow.com/questions/11867545/change-text-color-based-on-brightness-of-the-covered-background-area)) to change the foreground color to provide sufficient contrast. This produces some more wild colour combinations.

grafik

But comes at the risk we generate colours that blend into the background
grafik (maybe file under colourblind)

Or we just go back to origin %10 colours + W3C algo (but the black tags stick out like a sore thumb)
grafik


**Colourblind**

- There is a chance that tags will be generated with highly similar colours, but this is not a negative change from current situation. 
- Contrast against most job states is OK, even if we artificially choose the worst colour, none of our job states are that dark to start with.

martenson:
> % 10, even if it permits some light colours + W3C algo for font fix.

💯

hexylena:
Hah, ok, sounds good. :) (Sorry for editing, just felt like it could be read with a negative tone and didn't intend that.) 

That said, pastels are growing on me, so maybe we leave open for a while and see what people say.

grafik

They just don't stand out well against running datasets, the red + green should be dark enough to provide contrast.

grafik

And that could also be solved with a border
grafik

martenson:
@erasche border is exactly what I was just exploring after @jxtx suggested it, light grey 1px border looks very good on these I think

hexylena:
@martenson agreed, it's pretty nice! I like it.

So Galaxy community:

Option 1 | Option 2
--- | --- 
 |

jxtx:
I like option 2 but maybe the border should be the same hue as the bg but darker.

hexylena:
@jxtx sure, that's possible. Give me a few minutes.

dannon:
I was worried about 'too many lines' cluttering up the UI there with borders, but this looks nice.  Same color but darker is a great idea.

hexylena:
Dunno, can't tell much difference from grey lines. But I have imperfect colour vision (red-green blindness). Personally prefer grey lines but yeah.

grafik

Another example. I've converted the code to use HSL (so the 'darker' could be implemented easily,) and it's taking some tries to find similarly good distributions to what I saw in RGB pastels.

grafik

hexylena:
A little bit better

**Pastel**

grafik

**Dark**
grafik

martenson:
I think @jxtx meant the background color of the dataset not the background of the tag.

hexylena:
@martenson oh, hmm. (thanks for clarification)

hexylena:
based on dataset state? I don't love it

grafik

nsoranzo:
Pastel with gray border (i.e. Option 2) looks best to me.

hexylena:
Ok created some nice fake histories with lots of tags to show it off. Here are two we can put to the community :)

Option Null | Option 1 | Option 2
--- | --- | ---
grafik | grafik | grafik

shiltemann:
this is cool, will be really nice for histories with many tags :+1: 

My vote goes to Option2 :)

jxtx:
> I think @jxtx meant the background color of the dataset not the background of the tag.

No, I meant the color of the tag. I liked that one.

hexylena:
@jxtx Yeah, ok, when I zoom in like 500% I can see it, I'm looking at it again and it's ok. It looks a bit funny on yellow but the others are nice? Less grey is always good. I'll play around with it a bit and just find one that isn't too contrasting that certain ones stand out too much, but it's better than grey. Anything is better than grey.

martenson:
@jxtx my apologies

We have a poll now. :) https://help.galaxyproject.org/t/what-colors-of-galaxy-tags-do-you-like-the-most/101 I will add the color-border option. @erasche can you please get me comparable screenshot?

hexylena:
@martenson I'd say don't bother updating poll due to:

- the poll is already started, when I updated the option names to be more descriptive it wiped out previous votes
- it's pretty subtle? I do not think anyone will notice or say "I definitely do not want pastels because of the border colour", or would be able to tell the difference between that + grey due to how much stuff is going on in the UI, there are boxes open, datasets, tool menus, etc. I will claim PR author's discretion over those 100 pixels that would change.

martenson:
@erasche You got it.

> when I updated the option names to be more descriptive it wiped out previous votes

This is nice behavior actually.

hexylena:
> This is nice behavior actually.

Agreed, I can immediately imagine a bunch of bad uses for polls if it didn't. I like it!

hexylena:
Final version, I'm pretty happy :) Here it is zoomed 

grafik

Full screen version

grafik


Ready for review / merge.

martenson:
Thanks a bunch!

PR #7073, opened 2018-12-04
Client-side initialization queue

Nerdinacan:
So....  this is another big ugly PR.

The primary intent of these changes is to put in some more tools for the transition to a unified component-based app which me and Dannon will be tackling very soon.

**Initialization Chain**

The main thing that we're trying to accomplish here is to control the order in which initializations occur. Currently, initialization functions are written to the page via the python templates, and because of the  different loading mechanisms we have (notably iframes), the timing in which they run is not guaranteed.

What I've created here is an initialization cycle that first loads configs, waits for configs to stop loading (via a debounced observable), then generates the galaxy application singleton, then finally runs through a queue of registered init functions.

Both the init functions and config variables can be assigned at any time because of the nature of the debounced observables, so we've bypassed the "which thing got written to the page first?" problem.

The "config.set" function merges all configs together with a set of defaults before creating the Galaxy app, and the "config.addInitilization" function appends a new function onto the end of a queue of initilizations that will run after the app is built. Those init functions receive 2 arguments: the galaxy instance, and the merged configs.

Even if initialization functions are registered VERY late (i.e. a slow ajax call or iframe load), it will not rebuild the galaxy application because of the way we're combining observables, it'll just keep running init functions whenever they get registered. If you want to know more about how this works, checkout RxJS and/or ping me. Vue has observables kind-of built-in, so we won't need these specific tools when we switch over to a unified Vue app container.

_The most relevant changes to this functionality an be seen in the scripts/onload directory._

The new config object is used as follows:

```
// I live in some python template

config.set({
    options: options,
    bootstrapped: bootstrapped,
    form_input_auto_focus: ${h.to_js_bool(form_input_auto_focus)}
});

config.addInitialization(function(galaxyInstance, mergedConfigs) {
     // do your worst
});

config.addInitialization(function(galaxyInstance, mergedConfigs) {
     // do more evil, and it doesn't matter if it shows up later
     // I still won't run until after the app has been instantiated
});
```

**Removal of most global variables.**

I've also removed almost all global variables from the site. (This is why so many files were changed.) This includes nearly all window.Galaxy references, as well as other random window variables assigned by other scripting. There should now only be 2 global objects on window: bundledEntries and config.

Every webpack-rendered javascript object that is intended to be used by a python init script lives exclusively in window.bundledEntries now. (That was kind of the plan before, but I made it a little more strict.) All these objects should fade away into their respective components when the app is converted.

**Minor restructure of python templates**

Minor restructuring of the python templates was required to point all loaded pages to a single webpack-generated entry point. We are including about 3 javascript files on every page, but there needs to be a single entry point or webpack will execute its module loads more than once. Usually this point is "scripts/entry/generic.js", but there are exceptions for the admin, analysis and login sections.


Nerdinacan:
I think I need to separate this into two separate PRs, one for the big window global replacement and another for the initialization stuff. Going to close this one.

PR #7035, opened 2018-11-21
Remove ``sudo: false`` from .travis.yml, it will stop being supported soon

martenson:
https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

PR #7078, opened 2018-12-06
[18.09] Fix set_peek method for copied datasets

mvdbeek:
dataset_assoc is the job to ouptut dataset association that references
the HDA created by a job. If the HDA is copied the copies need to be
updated as well, so here we need to call `set_peek` on dataset, `not`
`dataset_assoc.dataset`.

Broken in f4343d876f13e55e92ca228582b4430dcd01e4aa

Thanks for tracking this down @jmchilton!

### DIFF
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1399,10 +1399,10 @@ class JobWrapper(HasResourceParameters):
                     line_count = context.get('line_count', None)
                     try:
                         # Certain datatype's set_peek methods contain a line_count argument
-                        dataset_assoc.dataset.set_peek(line_count=line_count)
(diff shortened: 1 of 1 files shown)


jmchilton (approved):

jmchilton:
Thanks for the fix, I hadn't noticed any errors surfacing from it but I was refactoring things in #7058 and noticed the incongruity - so this will break my refactoring when merged forward but is the right thing to do. Thanks for the fix @mvdbeek !

martenson:
the travis fail is probably same as https://github.com/galaxyproject/galaxy/pull/7035#issuecomment-441315077

martenson:
also this has failed before: https://travis-ci.org/galaxyproject/galaxy/jobs/464104614

Issue #7059, opened 2018-11-29
ToolShed tool installation error :This repository is not installed correctly


jvanbraekel:
Hello,

Wathever toolshed tool I try to install into a local test version of Galaxy 18.05, I get a repository error. 
Full message : "This repository is not installed correctly (see the Repository installation error below). Choose Reset to install from the Repository Actions menu, correct problems if necessary and try installing the repository again."
Sadly, there is no "repository installation error" at all : nothing above,below,aside nor into any logs. 
The only thing present into the logs that may be related to the problem is :

```
galaxy.tools.toolbox.base ERROR 2018-11-29 14:54:03,300 [p:115693,w:0,m:2] [MainThread] Error reading tool configuration file from path 'evolution/codingSnps.xml': [Errno 2] No such file or directory: '/home/galaxy/galaxy/tool-data/codingSnps.loc'
galaxy.tools.toolbox.base ERROR 2018-11-29 14:54:03,301 [p:115693,w:0,m:2] [MainThread] Error reading tool configuration file from path 'evolution/add_scores.xml': [Errno 2] No such file or directory: '/home/galaxy/galaxy/tool-data/add_scores.loc'
galaxy.tools.toolbox.base ERROR 2018-11-29 14:54:03,303 [p:115693,w:0,m:2] [MainThread] Error reading tool configuration file from path 'phenotype_association/sift.xml': [Errno 2] No such file or directory: '/home/galaxy/galaxy/tool-data/sift_db.loc'
```
For instance, with the last version of Fastqc, the installed location should be : /home/galaxy/shed_tools/toolshed.g2.bx.psu.edu/repos/devteam/fastqc/3e1cdf5406db/fastqc, but the folder home/galaxy/shed_tools remainded empty.   I have checked the permissions and owner : every folder under /home/galaxy is own by galaxy:galaxy and has 775 permissions.  The related config file : "shed_tool_conf.xml" look like : 

```
<?xml version="1.0"?>
<toolbox tool_path="/home/galaxy/shed_tools">
</toolbox>
```

Additionnaly, manual installation of other custom tools trough "tool_conf.xml" work as attented.  

Update : Configuration for a dedicated toolshed sqlite database is used ,  the sqlite file is in the right path and not empty ;  the connection string in galaxy.yaml is :

`install_database_connection: sqlite:////home/galaxy/galaxy/database/toolsheddb.sqlite?isolation_level=IMMEDIATE`

jvanbraekel:
[...]

PR #7080, opened 2018-12-06
Fix toolshed repository error display

jvanbraekel:
The template display was broken in case of error. 'in_error_state' is never defined it should be 'repository.in_error_state' instead. This was leading to some wrong field type display and not displaying the details about the error. Discover during attempts to solve #7059 .

This is how the template look like before :
galaxy _toolshed_administration_before

After the changes : 
galaxy _toolshed_ administration_after

### DIFF
--- a/templates/admin/tool_shed_repository/manage_repository.mako
+++ b/templates/admin/tool_shed_repository/manage_repository.mako
@@ -36,7 +36,7 @@ ${render_galaxy_repository_actions( repository )}
             </div>
             <div class="form-row">
                 <label>Description:</label>
-                %if in_error_state:
+                %if repository.in_error_state:
                     ${description|h}
                 %else:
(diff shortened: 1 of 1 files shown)


mvdbeek (approved):
Oh, looks like this has always been broken, but that's hard to say with those `*` imports.
Thanks a lot @jvanbraekel

mvdbeek:
@galaxybot test this

martenson:
Thank you @jvanbraekel !

martenson:
Thank you @jvanbraekel !

nsoranzo:
> Oh, looks like this has always been broken, but that's hard to say with those * imports.

Actually that was broken in commit bbea71ac7d29fb96be6a3f3ffb23d5e878a10a15 in 2013, so "just" 5 years ago!

Issue #7050, opened 2018-11-27
Containerized, Database-Free Job Creation and Completion


jmchilton:
Establish internal APIs consuming and producing JSON and detached from the database for every part of command execution. Rewrite these functions that can consume and produce database
models directly as currently or an alternative JSON-ified representation of the relevant models.

Job Setup:

- Command-line generation.
- Environment construction.
- Input staging from object store.
  - Simple datasets.
  - Composite datatypes.
  - Collections.
- Dependency resolution.

Job Completion:

- Job failure detection.
- Dynamic dataset discovery.
- Metadata collection.
- Output writing to object store.

The medium term goal would be placing all these operations into two containerized jobs (setup and completion) and allow splitting Galaxy job execution into a three containerized jobs. The first consuming a JSON-ified description of job inputs and parameters (including input dataset metadata, composite dataset paths) and generate a command-line for the second stage tool described container (typically a biocontainers container), and then a final job completion container.

If done well, one can then imagine a variant of Pulsar (or some other task execution API) that would stage its own files from the object store, generate its own command-line, and unstage files back to the object store by itself, and then just send a much more expansive version of galaxy.json back to Galaxy containing just the data to serialize to the database. This would allow data to live near compute without having to be loaded back to Galaxy workers for metadata collection or through Galaxy workers for input staging. This could dramatically decrease network transfers for large files and could prevent Galaxy from needing to directly touch say protected data to perform computation on it (as long as there is some container-aware service closer to the data that can be used).

While not the end goal, even this medium term goal could help a number of issues with Galaxy jobs.
- Job handlers could potentially not have to do dynamic file discovery - which can sometimes take hours and freeze job handlers for large jobs. https://github.com/galaxyproject/galaxy/issues/1854
- Pulsar could handle dynamic job discovery and staging of galaxy.json defined outputs correctly. https://github.com/galaxyproject/pulsar/issues/93
- We can imagine finally making progress on Pulsar caching of outputs. https://github.com/galaxyproject/pulsar/issues/48
- We can start to imagine a real library for Galaxy tool execution (the piece sitting inside the container) - that could be used by Tool building GUIs, for tool testing without requiring a whole Galaxy server, or by other platforms to leverage Galaxy tools without a Galaxy runtime.
- We can imagine a real library for Galaxy datatypes and metadata handling, sniffing, etc...

[...]

PR #7082, opened 2018-12-07
Remove legacy job handling hacks.

jmchilton:
These were left in at various times to let Galaxy jobs complete after version upgrades, but the plan was always to remove them eventually.

Modifying these bits heavily in #7058 / #7050 so it is a good time to clean things up.


natefoo (approved):
I was going to suggest that we mention in the release docs that running jobs will break if you upgrade from a version older than X.Y but I think a more reasonable blanket policy is just to say we attempt to support live upgrades from the past 2 releases and anything beyond that you're taking your chances on.

PR #7083, opened 2018-12-07
Move model/Tool interactions out of galaxy.jobs.output_checker.

jmchilton:
This means output_checker can be placed into galaxy-lib without dependency problems, and we can evaluate if a tool passed without access to a job object or a Tool object as long as we have a description of the job outputs and stdio objects (already in galaxy-lib).

This is a prelude to being able to do job output checking in Pulsar without Galaxy internals.

xref #7050 / #7058

### DIFF
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1527,7 +1527,17 @@ class JobWrapper(HasResourceParameters):
         self.cleanup(delete_files=delete_files)
 
     def check_tool_output(self, stdout, stderr, tool_exit_code, job):
-        return check_output(self.tool, stdout, stderr, tool_exit_code, job)
+        job_id_tag = "<unknown job id>"
+        if job is not None:
+            job_id_tag = job.get_id_tag()
+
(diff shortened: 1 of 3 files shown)


lib/galaxy/jobs/__init__.py
```diff
@@ -1527,7 +1527,17 @@ def path_rewriter(path):
         self.cleanup(delete_files=delete_files)
 
     def check_tool_output(self, stdout, stderr, tool_exit_code, job):
-        return check_output(self.tool, stdout, stderr, tool_exit_code, job)
+        job_id_tag = "<unknown job id>"
+        if job is not None:
+            job_id_tag = job.get_id_tag()
+
+        state, trimmed_stdout, trimmer_stderr = check_output(self.tool.stdio_regexes, self.tool.stdio_exit_codes, stdout, stderr, tool_exit_code, job_id_tag)
```
jmchilton, line 1534: What a random typo... also I don't think this actually trims those outputs. I need work on this.

jmchilton:
I guess this should be merged with https://github.com/galaxyproject/galaxy/pull/6685. I'll work on that this week.

PR #7085, opened 2018-12-07
De-couple objectstore abstraction from XML.

jmchilton:
- Allow loading objectstores for Python dicts and YAML configurations.
- New tests for XML parsing of azure, cloud, pithos, and S3 object stores configs.
- New tests for YAML parsing of all these object stores plus the vanilla disk and nested ones.

Sets up both loading object store configs right in Galaxy's galaxy.yml or Pulsar's app.yml as well as writing per-job object store configs out for #7050 so that jobs can know how to stage and unstage their own files on the potentially remote worker.

xref #7050 / #7058

### DIFF
--- a/lib/galaxy/objectstore/__init__.py
+++ b/lib/galaxy/objectstore/__init__.py
@@ -13,6 +13,7 @@ import threading
 import time
 from xml.etree import ElementTree
 
+import yaml
 try:
     from sqlalchemy.orm import object_session
 except ImportError:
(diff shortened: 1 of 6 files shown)


jmchilton:
Delaying this to 19.05 but it is ready to go, if someone wants to be nice and merge it before the release is branched just move the tag back.

hexylena:
I'll do that @jmchilton :)

hexylena (approved):

hexylena:
I'll try an add an XML test during the freeze period (that's what freezers are for, right?)

PR #7086, opened 2018-12-07
Localize version strings to the working directory w/predictable path.

jmchilton:
``new_file_path`` is awful and we shouldn't be using it. Making the job directory more self contained should make remote execution more plausible, simple.

xref #7050 / #7058


natefoo:
> `new_file_path` is awful and we shouldn't be using it

+:100:

Also +1 from me once the unit tests are fixed.

PR #7087, opened 2018-12-07
Add datatype for Mash Sketch files.

dfornika:
https://mash.readthedocs.io/en/latest/sketches.html

I'm working on adding a `mash screen` tool to the tools-iuc repo:

https://github.com/galaxyproject/tools-iuc/pull/2205

This pull request adds a datatype for mash sketch files.

### DIFF
--- a/config/datatypes_conf.xml.sample
+++ b/config/datatypes_conf.xml.sample
@@ -538,6 +538,7 @@
     <datatype extension="biom2" type="galaxy.datatypes.binary:Biom2" mimetype="application/octet-stream" display_in_upload="true">
       <converter file="biom2_to_biom1.xml" target_datatype="biom1"/>
     </datatype>
+    <datatype extension="msh" type="galaxy.datatypes.binary:MashSketch" display_in_upload="True" />
     <!-- Strand-specific Coordinate Count Datatype used by the Center for Eukaryotic Gene Regulation labs at Penn State -->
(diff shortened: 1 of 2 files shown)


bgruening:
@dfornika do you know if this file format as a magic-bit (https://en.wikipedia.org/wiki/List_of_file_signatures) set? Or can we `sniff` this format in any way?

dfornika:
I took a look at the docs and couldn't find any mention of a magic-bit. We could try contacting the developer(s) to find out how to sniff `.msh` files.

dfornika:
@ondovb is there any way to 'sniff' a mash sketch file for a specific magic bit or some other binary signature to detect the file type? I'd like to add a mash sketch datatype to Galaxy.

jmchilton:
I'm going to merge as is and we can follow up if there is a sniffer we can add?

jmchilton (approved):

dfornika:
Sounds good to me. Thanks @jmchilton

ondovb:
> @ondovb is there any way to 'sniff' a mash sketch file for a specific magic bit or some other binary signature to detect the file type? I'd like to add a mash sketch datatype to Galaxy.

Not exactly, but they are [Cap'n Proto](capnproto.org) serializations, so there may be a way to check for that. Capnp schemas also have globally unique IDs, which for the Mash schema is `@0xc4c8b1ada05e7704`. I don't know how to actually query this, or if it is even stored in the serialized files, but hopefully this is a good place to start looking at least. For what it's worth, our tools just look for a `.msh` extension :).

Issue #5148, opened 2017-12-07
Improve utility of configuration specification


jmchilton:
This issue a follow up to https://github.com/galaxyproject/galaxy/pull/5105 to improve what is there beyond a MVP.

- [ ] Allow more annotations of Galaxy options in config_schema.yml and use when generating documentation.
  - [ ] Specify an optional minimum Galaxy version that implements the options.
  - [ ] Specify how common it would be to modify the options - so can build a version of the docs that hides un-common options.
  - [ ] Specify unsafe / beta options and automatically warn on them.
  - [ ] Specify paths and path defaults more intelligently.
- [ ] Pre-process the options when loading into Galaxy to not need to duplicate type information like converting strings to integers when parsing ini options.
- [ ] Implement a modality of Galaxy that prevents loading non-lint passing configurations.
- [ ] Rework various other configuration files and allow them to be embedded in a hierarchal way into galaxy.yml.
  - [ ] job_conf.yml
  - [ ] per-IE options
  - [ ] datatypes
  - [ ] dependency resolvers (can already by yaml)
  - [ ] job metrics (can already by yaml)
  - [ ] containers (is already yaml)

natefoo:
John and I worked on a specification for a YAML job configuration and came up with the following. There's some incomplete stuff in there in regards to templating but we didn't want to lose this since a fair amount of thought went in to it.

```yaml
job_config:
  runners:
    local:
      class: galaxy.jobs.runners.local:LocalJobRunner
      workers: 4
      default: true
  handling:
    default: handlers
    method:
      - uwsgi_mule_message
      - db_preassign
    processes:
      handler0:
        tags: 
          - handlers
      handler1:  # no config
  execution:
    default: local
    environments:
      # could also be a dict keyed on ID, list is used to support the templating idea
      - id: local
        runner: local
      - common:
          native_spec: '-cores $cores -mem $mem' 
          mem: 12
          runner: drmaa
        environemnts:
          - id: bigmem
            mem: 45
      # dict version
      local:
          runner: local
  tools:
    # one option for specifying mapping defaults, instead of in "handling" and "execution"
    defaults:
      environment: foo
      handler: local
    mappings:
      cat1:
        environment: bar
  tools:
    # alternative tools implementation
    cat1:
      environment: bar
```

jmchilton:
@natefoo 

I think I want to make:

```yaml
runners:
  local:
      load: 'galaxy.jobs.runners.local:LocalJobRunner'
```

[...]

PR #7091, opened 2018-12-11
Configurable job/workflow handler assignment methods 

natefoo:
What started as a quick fix to the way we assign handlers (in order to support some new assignment methods) turned in to an overhaul. It's now possible to configure them explicitly, and the logic of how to configure them implicitly is now more centralized and clear.

This work isn't super exciting on its own, but it paves the way for new assignment methods (DB-based "message queue", AMQP messaging, etc.) that will be coming next.

## Assignment methods and configuration

Assignment methods are configured via attributes on the job config's `<handlers>` tag, the attributes and methods (none of which are now) are explained below (text pulled from `job_conf.xml.sample_advanced):

- `assign_with` - How jobs should be assigned to handlers. The value can be a single method or a comma-separated list that will be tried in order. The default depends on whether any handlers and a job handler "pool" (such as uWSGI mules in a `job-handlers` farm) are configured. Valid methods are:
     - `mem-self` - Jobs are assigned to the web worker that received the tool execution request from the user via an internal in-memory queue. If a tool is configured to use a specific handler, that configuration is ignored; the process that creates the job *always* handles it. This can be slightly faster than `db-self` but only makes sense in single process environments without dedicated job handlers. This option replaces the former `track_jobs_in_database` option in galaxy.yml.
     - `db-self` - Like `mem-self` but assignment occurs by setting a new job's 'handler' column in the database to the process that created the job at the time it is created. Additionally, if a tool is configured to use a specific handler (ID or tag), that handler is assigned (tags by `db-preassign`). This is the default if no handlers are defined and no `job-handlers` pool is present (the default for a completely unconfigured Galaxy).  
     - `db-preassign` - Jobs are assigned a handler by selecting one at random from the configured tag or default handlers. This occurs by setting a new job's 'handler' column in the database to the chosen handler ID (hence "preassign"ment). This is the default if handlers are defined and no `job-handlers` pool is present.
     - `uwsgi-mule-message` - Jobs are assigned a handler via uWSGI mule messaging (see uWSGI documentation). A mule in the `job-handlers` (if sent to the default tag) or `job-handlers.<tag>` farm will recieve the message and assign itself. This the default if a `job-handlers` pool is present and no handlers are configured.

  In the event that both a `job-handlers` pool is present and handlers are configured, the default is `uwsgi-mule-message,db-preassign`. At present, only `uwsgi-mule-message` is capable of deferring handler assignment to a later method (which would occur in the event that a tool is configured to use a tag for which there is not a matching pool).

  In all cases, if a tool is configured to use a specific handler (by ID, not tag), configured assignment methods are ignored and that handler is directly assigned in the job's 'handler' column at job creation time.

 - `default` - An ID or tag of the handler(s) that should handle any jobs not assigned to a specific handler (which is probably most of them). If unset, the default is any untagged handlers plus any handlers in the `job-handlers` (no tag) pool.

   Note that in the event such a mixed configuration environment exists (both a `job-handlers` pool (farm) and untagged handlers are configured), the default value of `assign_with="uwsgi-mule-message,db-preassign"` would prevent any of the configured handlers from being assigned since `uwsgi-mule-message` is the preferred assignment method.

Comma-separated XML attribute values is not the best way to do a list, but after some discussion with @jmchilton and our combined aversion to making any significant extensions to the job_conf.xml schema since the hope is to switch to YAML anyway (#5148) and the proposed structure there didn't translate to any XML schema we liked, this was considered to be reasonable temporary compromise, especially since it matches the way we do tags in `job_conf.xml` already.

## Improvements

- `get_handler()` is replaced with `assign_handler()`. When I added uWSGI messaging, there was a lot of special casing in `get_handler()` and spread out elsewhere for handler assignment.
- The concept of a asynchronous worker "pool" is clearer and more generalized now. I have separate work that refactors stack messaging to combine it with the existing kombu messaging that should finish things up and make worker pools usable by any part of the code.
- A `default` attribute on `<handlers>` is no longer required in any case, the default is simply any untagged handlers that are configured, plus the `job-handlers` pool (farm), if it exists. This is explained fully in additions to `job_conf.xml.sample_advanced`.
- Pool members (aka mules in a farm) can now be assigned the old fashioned way (`db-preassign`) rather than being forced to use mule messaging (`uwsgi-mule-message`). This feature will be more useful for the new assignment mechanisms and allowing the mixture of static (mule) and dynamic (webless) pool members.

### Job execution improvements

- The job of assigning handlers and queuing jobs for execution is moved from the tool action to `JobManager.enqueue()`. The job queue and job stop queues are no longer directly exposed in the manager. The manager's `enqueue()` and `stop()` methods are now the high level interfaces to job execution.
- You can now create "subpools" matching handler tags to separate handlers for different tools/purposes but still use mule messaging for assignment, and this is done without any explicit `<handlers>` configuration, you simply use a `<tool>` in `job_conf.xml`:

    ```xml
    <tools>
        <tool id="foo" handler="bar"/>
    </tools>
    ```

    and then create a pool for tag `bar` with:

    ```console
    $ uwsgi ... --mule=lib/galaxy/main.py --mule=lib/galaxy/main.py --farm=job-handlers.bar:1,2
    ```

###  Workflow scheduling improvements

- You can now create a `workflow-schedulers` pool (uWSGI farm) and automatically separate workflow scheduling from job handlers. Previously scheduling would automatically occur in the `job-handlers` pool (if it existed) or in the configured default handlers (if not).
- In my initial uWSGI farm implementation for workflow schedulers I'd used assignment by mule messaging, and only recently discovered in conversation with @jmchilton that this is not correct, since it can allow for multiple workflows to be scheduled simultaneously in the same history. After lengthy discussions as to the correct implementation, I decided to simply switch workflow scheduling to use the `db-preassign` method. However, creating a `workflow-schedulers` (or if not that, `job-handlers`) pool does automatically make its member mules eligible for assignment via `db-preassign` (which was not the case before). It is still possible to use mule messages by creating a `workflow_schedulers_conf.xml` and setting `<handlers assign_with="uwsgi-mule-message">`.

----

The unit tests only test for correct parsing of the job config, additional testing of execution with the various assignment methods would be good. I will probably work on this in the next PR (which will add new assignment types) though.

### DIFF
--- a/config/galaxy.yml.sample
+++ b/config/galaxy.yml.sample
@@ -1563,9 +1563,8 @@ galaxy:
   # failing jobs are just failed outright.
   #default_job_resubmission_condition: null
 
-  # In multiprocess configurations, notification between processes about
-  # new jobs must be done via the database.  In single process
-  # configurations, this can be done in memory, which is a bit quicker.
+  # This option is deprecated, use the `mem-self` handler assignment
+  # option in the job configuration instead.
   #track_jobs_in_database: true
 
   # This enables splitting of jobs into tasks, if specified by the
--- a/config/job_conf.xml.sample_advanced
+++ b/config/job_conf.xml.sample_advanced
@@ -332,24 +332,76 @@
              used by Galaxy is no environment variable of the specified name is found.
         -->
     </plugins>
-    <handlers default="handlers">
-        <!-- Additional job handlers - the id should match the name of a
-             [server:<id>] in galaxy.ini.
+    <handlers>
+        <!-- Job handler processes - for a full discussion of job handlers, see the documentation at:
+             https://docs.galaxyproject.org/en/latest/admin/scaling.html
+
+             The <handlers> container tag takes two optional attributes:
+
+               <handlers assign_with="method" default="id_or_tag"/>
+
+               - `assign_with` - How jobs should be assigned to handlers. The value can be a single method or a
+                 comma-separated list that will be tried in order. The default depends on whether any handlers and a job
+                 handler "pool" (such as uWSGI mules in a `job-handlers` farm) are configured. Valid methods are:
+
+                   - `mem-self` - Jobs are assigned to the web worker that received the tool execution request from the
+                     user via an internal in-memory queue. If a tool is configured to use a specific handler, that
+                     configuration is ignored; the process that creates the job *always* handles it. This can be
+                     slightly faster than `db-self` but only makes sense in single process environments without
+                     dedicated job handlers. This option replaces the former `track_jobs_in_database` option in
+                     galaxy.yml.
+
+                   - `db-self` - Like `mem-self` but assignment occurs by setting a new job's 'handler' column in the
+                     database to the process that created the job at the time it is created. Additionally, if a tool is
+                     configured to use a specific handler (ID or tag), that handler is assigned (tags by
+                     `db-preassign`). This is the default if no handlers are defined and no `job-handlers` pool is
+                     present (the default for a completely unconfigured Galaxy).
+
+                   - `db-preassign` - Jobs are assigned a handler by selecting one at random from the configured tag or
+                      default handlers. This occurs by setting a new job's 'handler' column in the database to the
+                      chosen handler ID (hence "preassign"ment). This is the default if handlers are defined and no
+                      `job-handlers` pool is present.
+
+                   - `uwsgi-mule-message` - Jobs are assigned a handler via uWSGI mule messaging (see uWSGI
+                     documentation). A mule in the `job-handlers` (if sent to the default tag) or `job-handlers.<tag>`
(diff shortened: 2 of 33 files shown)


natefoo:
Test failures are legit, I'm working on them.

jmchilton:
This breaks handler_runner_plugins - that variable never gets set anymore because you are no longer calling _parse_handler 🤔.

jmchilton:
I'll fix it and add a test case.

natefoo:
Hey cool, I noticed that it was broken a while ago but didn't realize it was me who broke it. :sunglasses: 

IIRC the reason I added that feature was so not all my handlers would listen to AMQP. Normally it's only 2, which are the ones that all Pulsar jobs are handled by. However, I haven't noticed any issues with the other handlers also handling those messages.

Thanks!

PR #7093, opened 2018-12-11
Allow for missing trailing \t in GFF records.

innovate-invent:
Some tools do not output a trailing \t when writing a GFF record with an empty group.
The current code assumes .split("\t") will create an array of size 9, this fixes that.

### DIFF
--- a/lib/galaxy/datatypes/util/gff_util.py
+++ b/lib/galaxy/datatypes/util/gff_util.py
@@ -42,7 +42,7 @@ class GFFInterval(GenomicInterval):
         self.score = self.fields[self.score_col]
 
         # GFF attributes.
-        self.attributes = parse_gff_attributes(fields[8])
+        self.attributes = parse_gff_attributes(fields[8] if 8 < self.nfields else "")
 
     def copy(self):
         return GFFInterval(self.reader, list(self.fields), self.chrom_col, self.feature_col, self.start_col,


nsoranzo:
@innovate-invent Thanks for the contribution!

It seems to me that these GFF files are not compliant with the specification https://github.com/The-Sequence-Ontology/Specifications/blob/master/gff3.md , maybe fixing the tools producing them is the best approach?

innovate-invent:
Would it be better to `raise MissingFieldError("No field for group column.")` from a bounds check?

nsoranzo:
@innovate-invent That's surely a valid option. But isn't the 9th column named "attributes" (not "group")?

innovate-invent:
@nsoranzo That depends on if it is supposed to be gff2 or gff3:
http://gmod.org/wiki/GFF2

nsoranzo:
Well, GFF2 has been deprecated for years, and both GFF3 and [GTF](https://www.ensembl.org/info/website/upload/gff.html) call the 9th column "attributes", so I'd use that in the error message.

innovate-invent:
Before committing to throwing a exception, is that really a better option than quietly handling the missing trailing \t?

nsoranzo:
I think it is, together with opening an issue for the tool, because GFF is quite a messy set of formats. But maybe others have a different opinion.

jmchilton:
Enough committers are uncomfortable with supporting these older files, so we're going to close this. Sorry but thanks for the contribution.

nsoranzo:
@innovate-invent I've implemented the suggested exception raising in #7708.

PR #7095, opened 2018-12-11
More Structured Job I/O Tracking

jmchilton:
- Track messages generated tool stdio tags in a JSON blob and don't prepend them to the tool's reported stdout/stderr. Report them separately in the UI.
- Track job script stdout and stderr separately from the tool's stdout and stderr in the database. Should make our ability to report information about dependency resolution and metadata generation and lot better.

Builds on #6685 #6338. Pulled out the tracking branch #7058 for issue #7050.


jmchilton:
Delaying this to 19.05 but it is ready to go, if someone wants to be nice and merge it before the release is branched just move the tag back.

lib/galaxy/tools/verify/interactor.py
```diff
@@ -817,11 +817,36 @@ def register_exception(e):
         "stdout": "Standard output of the job",
         "stderr": "Standard error of the job",
     }
+    # TODO: Only hack the stdio like this for older profkle, for newer tool profiles
```
mvdbeek: s/profkle/profile

lib/galaxy/jobs/output_checker.py
```diff
@@ -14,11 +14,45 @@
...
+    """
+    regex_match = re.search(regex.match, stream, re.IGNORECASE)
+    if regex_match:
+        reason = __regex_err_msg(regex_match, stream_name, regex)
+        job_messages.append(reason)
+        return max(max_error_level, regex.error_level)
+    return max_error_level
+
+
+def check_output_regex_byline(job_id_tag, regex, stream, stream_append, max_error_level):
```
mvdbeek: This  isn't used AFAICT, drop it ?

jmchilton:
I don't think this quite works at all like I expected... putting it back into WIP. Might be good to have tests that somehow proved the streams were being separately properly.

lib/galaxy/jobs/__init__.py
```diff
@@ -993,7 +993,7 @@ def _fix_output_permissions(self):
             if os.path.exists(path):
                 util.umask_fix_perms(path, self.app.config.umask, 0o666, self.app.config.gid)
 
-    def fail(self, message, exception=False, stdout="", stderr="", exit_code=None):
+    def fail(self, message, exception=False, tool_stdout="", tool_stderr="", exit_code=None, job_stdout=None, job_stderr=None):
```
mvdbeek, line 990: Need to update `job_wrapper.fail(stdout=...)` kubernetes.py and pulsar.py for this change as well.

lib/galaxy/jobs/output_checker.py
```diff
@@ -14,11 +14,45 @@
...
+        regex_match = re.search(regex.match, line, re.IGNORECASE)
+        if regex_match:
+            rexmsg = __regex_err_msg(regex_match, regex)
+            log.info("Job %s: %s" % (job_id_tag, rexmsg))
+            stream_append.append(rexmsg)
+            return max(max_error_level, regex.error_level)
+    return max_error_level
+
+
+def check_output(stdio_regexes, stdio_exit_codes, stdout, stderr, tool_exit_code, job_id_tag):
```
mvdbeek, line 37: `state = DETECTED_JOB_STATE.OK` on line 69 has no effect, we could drop it.

lib/galaxy/jobs/output_checker.py
```diff
@@ -149,32 +180,35 @@ def check_output(tool, stdout, stderr, tool_exit_code, job):
                     "assuming tool was successful: " + tb)
         state = DETECTED_JOB_STATE.OK
 
-    # Store the modified stdout and stderr in the job:
-    if job is not None:
-        job.set_streams(stdout, stderr)
-
-    return state
+    return state, stdout, stderr, job_messages
```
mvdbeek, line 156: Can we bump up the log.warning above to log.exception ?
This seems like something serious that we should be fixed if it ever really happens (plus we can drop the `traceback.format_exc()` part).

lib/galaxy/jobs/output_checker.py
```diff
@@ -14,11 +14,45 @@
...
+    """
+    check a single regex against a stream line by line, since errors
+    are expected to appear in the end of the stream we start with
+    the last line
+    returns the max of the error_level of the regex and the given max_error_level
+    """
+    for line in reversed(stream.split("\n")):
+        regex_match = re.search(regex.match, line, re.IGNORECASE)
+        if regex_match:
+            rexmsg = __regex_err_msg(regex_match, regex)
```
mvdbeek: we don't use it, but it should be `__regex_err_msg(regex_match, stream_name, regex)` if we don't drop it

lib/galaxy/jobs/output_checker.py
```diff
@@ -149,32 +180,35 @@ def check_output(tool, stdout, stderr, tool_exit_code, job):
...
+    if mend - mstart > 256:
+        match_str = match.string[mstart : mstart + 256] + "..."
+    else:
+        match_str = match.string[mstart: mend]
+
     # If there's a description for the regular expression, then use it.
     # Otherwise, we'll take the first 256 characters of the match.
     if None is not regex.desc:
-        err_msg += regex.desc
+        desc += regex.desc
```
mvdbeek, line 177: I guess `if regex.desc is not None` would read a bit better ?

lib/galaxy/jobs/runners/__init__.py
```diff
@@ -437,22 +437,62 @@ def mark_as_resubmitted(self, job_state, info=None):
...
+    def _job_io_for_db(self, stream):
+        return shrink_stream_by_size(stream, DATABASE_MAX_STRING_SIZE, join_by="\n..\n", left_larger=True, beginning_on_size_error=True)
+
+    def _finish_or_resubmit_job(self, job_state, job_stdout, job_stderr, job_id=None, external_job_id=None):
+        job_wrapper = job_state.job_wrapper
+        try:
+            job = job_state.job_wrapper.get_job()
+            exit_code = job_state.read_exit_code()
+
+            tool_stdout = tool_stderr = None
```
mvdbeek: This can't be reached, and `None` would blow up in `check_output_regex`.

lib/galaxy/model/__init__.py
```diff
@@ -252,6 +263,27 @@ def log_str(self):
...
+        raise NotImplementedError("Attempt to set stdout, must set tool_stdout or job_stdout")
+
+    def get_stderr(self):
+        stderr = self.tool_stderr
+        if self.job_stderr:
+            stderr += "\n" + self.job_stderr
+        return stderr
+
+    def set_stderr(self, stderr):
+        raise NotImplementedError("Attempt to set stdout, must set tool_stderr or job_stderr")
```
mvdbeek, line 282: We need to change https://github.com/galaxyproject/galaxy/blob/355b48c361e68a7d9fbe9fe86de7aa2313e5d608/lib/galaxy/webapps/galaxy/controllers/admin.py#L1647 and https://github.com/galaxyproject/galaxy/blob/355b48c361e68a7d9fbe9fe86de7aa2313e5d608/lib/galaxy/jobs/handler.py#L302 too, right ?

mvdbeek (commented):
This is awesome, but I think there's some minor things that need fixing.

lib/galaxy/tools/verify/interactor.py
jmchilton: will rebase with a fix, thanks

lib/galaxy/jobs/output_checker.py
jmchilton: will rebase with a fix, thanks

lib/galaxy/jobs/__init__.py
jmchilton: indeed - how did I miss that. will rebase with a fix, thanks

lib/galaxy/jobs/output_checker.py
jmchilton: I like removing the ``else``s inside the main ``if`` more? It is more clear this way what the default value is and that it always has a default value. I'm going to do that instead - hope that is okay.

lib/galaxy/jobs/output_checker.py
jmchilton: will do in a rebase when fixing up default state, thanks!

lib/galaxy/jobs/output_checker.py
jmchilton: Will fix in a rebase, thanks.

lib/galaxy/jobs/runners/__init__.py
jmchilton: ... will fix in a rebase - thanks

mvdbeek:
> Might be good to have tests that somehow proved the streams were being separately properly.

How about a integration test that uses a custom DEFAULT_JOB_FILE_TEMPLATE to inject some messages on stdout/stderr ?

jmchilton:
@mvdbeek Thanks for the detailed review, I'll admit without that test this was pretty half baked. I've added a test now and I think it is working?

mvdbeek:
Awesome, yeah, looks great. I have a feeling that we might also be able to get rid of `capture_stdout_stderr` introduced for kubernetes.

mvdbeek:
Thanks again @bernt-matthias @jmchilton, that was much needed!

PR #7097, opened 2018-12-12
Replace display applications mako with vue component

guerler:
This PR replaces the display applications mako with a vue component and provides a new display applications api endpoint.


lib/galaxy/webapps/galaxy/api/display_applications.py
```diff
@@ -13,6 +13,7 @@
 class DisplayApplicationsController(BaseAPIController):
 
     @expose_api
+    @require_admin
```
dannon: Out of curiosity, why do we require admin for these?  Is there anything privileged in the response, or is it just historical?

Seems like this is info we'd expose just like the /api/tools index.

(obviously the reload/post below should require admin)

guerler: Historical. The legacy endpoint which previously produced this display was in the admin controller. I preserved the same level of access requirements. I did this to avoid a discussion, although I generally agree that this information should be public. I removed the requirement below.

client/galaxy/scripts/components/admin/DisplayApplications.vue
```diff
@@ -0,0 +1,124 @@
+<template>
+    <div>
+        <div v-if="messageVisible" :class="messageClass">
```
dannon: Did you try using b-alert here?  I'd like to standardize towards using that component for alerts and messages like this.

guerler: Replaced it. Thanks.

dannon:
@guerler Thanks for the tweaks, looks good to me!  I think future enhancements could really be made to the datatype lists, but I'm not sure what'd work best there right now -- when something is applicable to a ton of datatypes it doesn't get displayed well at all (but this, of course, isn't something new with this PR).

PR #7101, opened 2018-12-12
Fix workflow output rectification

### LINKED ISSUE #7055: Cannot undo use of snowflakes in workflow
On 18.09 and usegalaxy.org the effect of snowflakes is not correctly undone. Once a workflow had one(or more) snowflake set (and saved) the hide post action stay even after removing all snowflakes. I believe this tends to happen after adding a new step and then removing the snowflakes. 

Example: https://usegalaxy.org/u/_jelle/w/no-more-snowflakes (downloading the workflow will show the post actions)

dannon:
This should fix #7055.  It cleans up the logic that determines whether a workflow is 'using workflow outputs' (via stars) to only count 'tool steps'.  It also 'cleans' PJAs inadvertently added to previous versions of the workflow by dropping the tool step constraint during that processing.

So, going forward, workflows will be correctly saved with outputs/stars.

### DIFF
--- a/client/galaxy/scripts/layout/menu.js
+++ b/client/galaxy/scripts/layout/menu.js
@@ -221,9 +221,9 @@ var Collection = Backbone.Collection.extend({
         }
         if (options.helpsite_url) {
             helpTab.menu.unshift({
-                    title: _l("Galaxy Help"),
-                    url: options.helpsite_url,
-                    target: "_blank"
+                title: _l("Galaxy Help"),
+                url: options.helpsite_url,
+                target: "_blank"
             });
         }
         this.add(helpTab);
--- a/client/galaxy/scripts/mvc/workflow/workflow-manager.js
+++ b/client/galaxy/scripts/mvc/workflow/workflow-manager.js
@@ -92,10 +92,10 @@ class Workflow {
         var using_workflow_outputs = false;
         var has_existing_pjas = false;
         $.each(this.nodes, (k, node) => {
-            if (node.workflow_outputs && node.workflow_outputs.length > 0) {
+            if (node.type === "tool" && node.workflow_outputs && node.workflow_outputs.length > 0) {
                 using_workflow_outputs = true;
             }
-            $.each(node.post_job_actions, (pja_id, pja) => {
+            $.each(node.post_job_actions, (pja_using_workflow_outputsid, pja) => {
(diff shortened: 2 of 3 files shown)


mvdbeek:
@dannon is this ready to review ?

dannon:
@mvdbeek Uhh, good question.  Certainly wouldn't hurt to take a look, if you want, but I did not ever circle back to check subworkflow handling.

mvdbeek:
This is working well for me. One downside is that you'll probably start composing your workflow, all without stars. Now save and reload --> all tools have stars, which is OK IMO. But if you now add additional tools you need to star them all, which might be surprising to users ("Why are my new outputs not appearing ?").
So I guess if by default a new tool state could always have all outputs starred I think that'd be the least surprising.

dannon:
@mvdbeek Thanks for taking a look.  Yeah, I like the idea of the default state being explicitly made an output after this change, I'll drop that in when I have a sec.

mvdbeek:
Hmm, the last commit doesn't quite work for me. If I add a new tool it isn't starred. if I save and reload the workflow it becomes starred.

dannon:
@mvdbeek Weird.  Any error?  And the client rebuild was successful with the new changeset?  I'm seeing this work in my hands for single and multi output tools, etc.

mvdbeek:
One thing I did was merging this forward into dev (didn't want to do the db downgrade ...). I've thested the PR as is now and I get the correct behavior, so the PR is definitely fine. We may just have to double-check after merging forward that this still works as intended.

mvdbeek (approved):

dannon:
@mvdbeek Ahh, sure, I understand the aggravation with that all too well :)  I will double check it after we get it merged and merged forward.

mvdbeek:
@dannon I think I'd be more comfortable merging this into dev, and then maybe merging backwards if there are no problems ?

dannon:
@mvdbeek I opened this PR so long ago I completely forgot about it, assumed we'd merged it already.  Yeah, if you want I can change the base, I don't care.

This should have been included in 19.01 based on the milestone, which is a bummer.

mvdbeek:
@galaxybot test this

mvdbeek:
Thanks again @dannon and sorry for slowing this down. failing travis tests are unrelated.

PR #7110, opened 2018-12-13
Add to_dict() to ObjectStores.

jmchilton:
Dict-ify an object store configuration back to its ``config_dict`` description. More library support from #7058 that will be useful downstream in supporting #7050. Arguably this is speculative complexity but this patch has I believe 100% test coverage and merging now will help ensure other people enhancing and implementing object stores implement them in a way that is compatible with this serialization.

xref #7085

### DIFF
--- a/lib/galaxy/objectstore/__init__.py
+++ b/lib/galaxy/objectstore/__init__.py
@@ -220,6 +220,16 @@ class ObjectStore(object):
         config_dict = clazz.parse_xml(config_xml)
         return clazz(config, config_dict, **kwd)
 
+    def to_dict(self):
+        extra_dirs = []
+        for extra_dir_type, extra_dir_path in self.extra_dirs.items():
+            extra_dirs.append({"type": extra_dir_type, "path": extra_dir_path})
+        return {
+            'config': config_to_dict(self.config),
+            'extra_dirs': extra_dirs,
+            'type': self.store_type,
+        }
+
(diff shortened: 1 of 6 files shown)


bgruening (approved):

lib/galaxy/objectstore/__init__.py
```diff
@@ -220,6 +220,16 @@ def from_xml(clazz, config, config_xml, **kwd):
         config_dict = clazz.parse_xml(config_xml)
         return clazz(config, config_dict, **kwd)
 
+    def to_dict(self):
+        extra_dirs = []
```
mvdbeek, line 224: Do you think there'd be an advantage of listing attributes that should be serialised as a class variable ?
i.e 
```Python
class SomeObjectStore(object):

    serializable_attributes = ('config', 'type', 'extra_dirs',)

    def to_dict(self):
        return {attr: getattr(attr, self) for attr in self.serializable_attributes}
```

mvdbeek:
LGTM, want to take it out of WIP ?

lib/galaxy/objectstore/__init__.py
jmchilton: I'll consider this - I'm not sure it would simplify much in the subclasses. It would be closer to what we do other places in the app though I'll ready admit. Thanks for the review!

PR #7112, opened 2018-12-13
Allow "installation" of container images, display images in Manage Dependencies

mvdbeek:
- I have modified the `api/tools/{tool_id}/dependencies` endpoint by adding an additional `resolver_type` parameter, which can be set to "mulled" or "conda" for now. If set, that resolver will be used to install (in the case of Conda) or pull (for docker, and singularity soon) dependencies. If not set the old rules apply, i.e resolvers are tried according to the hierarchical order defined by the admin.
- I've also hacked the ContainerDependencies into the regular DependencyManager, which allows us the handle things like the status display, installation and uninstallation via the DependencyResolversView abstraction. That means you can do GET, POST and DELETE to `api/tools/{tool_id}/dependencies` and it'll do the corresponding action.
- Unresolved dependencies (aka NullDependency) are displayed with their version in the Manage Dependencies menu

screen shot 2018-12-13 at 16 06 17

This has grown pretty large by now, if someone wants to review this I'd appreciate it.
Next step is to pull and convert mulled docker images for singularity, and then I'd like to move the Manage Dependencies menu to the API and use Vue to overhaul the display.


mvdbeek:
I think this should be merged as part of https://github.com/galaxyproject/galaxy/pull/7125, or, if we decide that this isn't the right abstraction, stripped from  https://github.com/galaxyproject/galaxy/pull/7125.

PR #7113, opened 2018-12-13
Fix `make client-watch` extra error message.

dannon:
Client watch referred to an old gulp task that is no longer necessary.


martenson:
thanks @dannon!

Issue #6330, opened 2018-06-13
Enhancement: Enable tag/anno edits for queued, executing, failed jobs


jennaj:
Would it be possible to add these in? Would make it possible to annotate jobs right after being launched, to better keep track of data, rather than needing to go back in after and figure out what was done. Inherited tags are not always enough, and annotation is not inherited.

Alternatively, could add in a function to tools that allow tag/annotation manipulations. Could be placed on all forms right above "submit". Bigger project yet is a better function when doing runs in batch.

For failed jobs, would be useful to add tag/anno in for reruns, troubleshooting, experimenting with params, etc. 

<img width="249" alt="screen shot 2018-06-13 at 10 55 11 am" src="https://user-images.githubusercontent.com/8999735/41369059-a7bed41e-6ef8-11e8-81b3-e5d1b14768af.png">

<img width="255" alt="screen shot 2018-06-13 at 10 49 52 am" src="https://user-images.githubusercontent.com/8999735/41369075-afb651ba-6ef8-11e8-8b54-ec503b675e45.png">

<img width="248" alt="screen shot 2018-06-13 at 10 55 25 am" src="https://user-images.githubusercontent.com/8999735/41369079-b33a0c8c-6ef8-11e8-9677-ee188e5884b4.png">

jennaj:
@ic4f This was assigned to you by @nekrut in https://github.com/galaxyproject/usegalaxy-playbook/projects/2#card-14475567

We were unable to assign the issue ticket to you in GH using "@ic4f" for some reason. Could you check why and assign yourself? Thanks!

jdavcs:
Can't self-assign - already tried: self-assign option not available. (will ask on slack)
But to acknowledge, yes, this issue is on me.

dannon:
@jennaj @ic4f I took a look and fixed it up, @ic4f can now be assigned issues.

jennaj:
@ic4f Thanks for taking this on. Doing this for collections might be tricky but we would want to include that.

jennaj:
@ic4f Adding tags is now possible. I did notice a few things, not sure what can be done about any though.

 - If the dataset state changes while I am creating a tag, the GUI freezes for a few seconds and the tag is lost.

[...]

PR #7115, opened 2018-12-13
Enable tagging and annotation of jobs regardless of state

jdavcs:
See issue #6330 

As far as I can tell, as long as a model exists (and is not purged or deleted - which is checked), the model's state is inconsequential. All states (states.js) seem to have no relevance to the creation of a record in the db association tables (history_dataset_association_tag/annotation_association).

I did not add any new tests for this change, so I'll what Travis says about this...

### DIFF
--- a/client/galaxy/scripts/mvc/dataset/dataset-li-edit.js
+++ b/client/galaxy/scripts/mvc/dataset/dataset-li-edit.js
@@ -120,12 +120,12 @@ var DatasetListItemEdit = _super.extend(
             var state = this.model.get("state");
 
             if (!this.model.isDeletedOrPurged()) {
-              //Enable tagging+annotation regardless of job state (see issue #6330) 
+                //Enable tagging+annotation regardless of job state (see issue #6330)
                 this._renderTags($details);
                 this._renderAnnotation($details);
 
                 if (_.contains([STATES.OK, STATES.FAILED_METADATA], state)) {
-                  this._makeDbkeyEditLink($details);
+                    this._makeDbkeyEditLink($details);
                 }
             }
 
--- a/client/galaxy/scripts/mvc/dataset/dataset-li-edit.js
+++ b/client/galaxy/scripts/mvc/dataset/dataset-li-edit.js
@@ -83,7 +83,7 @@ var DatasetListItemEdit = _super.extend(
                 }
 
                 // disable if still uploading or new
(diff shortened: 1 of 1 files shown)


mvdbeek:
At first I was thinking this is problematic because tools can branch on tags and metadata, but I guess it's not a big problem as we update the job parameters once we build a jobs' command line, so at least we record the provenance correctly. And if the API allows this anyway there's even less of a point to not allow it.

jdavcs:
> At first I was thinking this is problematic because tools can branch on tags and metadata ...

Thank you for looking into this, @mvdbeek . And yes, that's exactly the kind of pitfalls I'm hoping to avoid (..learning the ropes here..) Do I understand correctly that a tool can make (procedural?) decisions based on user-assigned metadata? So, assigning tag X to a dataset could result in some tool trying to access/use that dataset regardless of its state? (sorry if I misunderstood)

mvdbeek:
> Do I understand correctly that a tool can make (procedural?) decisions based on user-assigned metadata?

Tools template out a command line using Cheetah once all input datasets are ready. This can include things like `if metdata.dbkey == 'bla':` or similar. My concern was that you could launch a workflow and then modify the the input dataset of a step waiting to run, thereby modifying the result of a dependent step. But I think that's OK.

> So, assigning tag X to a dataset could result in some tool trying to access/use that dataset regardless of its state? (sorry if I misunderstood)

No that can't happen, I think.

dannon:
@ic4f Made two minor tweaks, but this looks good to me, thanks!

jdavcs:
@mvdbeek - thank you for clarifying - this helps, I understand now!
@dannon - thanks for looking+editing! Yes, tabstop=4, I noticed that comment line after submitting. My .vimrc was set to ts=2 for js files - I've changed that to =4 to stay consistent with the team, but that one line fell through. And the second edit - yes, absolutely! Wanted to minimize my footprint in the original code, so used statement as is. Thanks!

dannon (approved):

dannon:
@ic4f Regarding spacing and other changes, you can always run `make client-format`, which runs https://prettier.io with our minimal set of rules.

The other bit of refactoring I did was just a general recommendation moving forward to replace jquery, underscore, etc., with vanilla JS when you can, but it's also totally safe to use constructs that exist in the surrounding code as you did.

Looked like selenium tests blew up, so I'm rerunning those and will merge this after.

PR #7116, opened 2018-12-14
Fix activation of ``_galaxy_`` conda env

nsoranzo:
- Use `conda_activate`, which works also under dash
- Run `conda_activate` also when `$GALAXY_CONDA_ENV` already exists
- Run `virtualenv` when the conda env is active but `$GALAXY_VIRTUAL_ENV`   does not exist
- Update the default for `GALAXY_CONDA_ENV` in scripts/common_startup_functions.sh

Follow-up on https://github.com/galaxyproject/galaxy/pull/6975 .

### DIFF
--- a/scripts/common_startup.sh
+++ b/scripts/common_startup.sh
@@ -106,8 +106,7 @@ fi
 # GALAXY_CONDA_ENV is not set here because we don't want to execute the Galaxy version check if we don't need to
 
 if [ $SET_VENV -eq 1 -a $CREATE_VENV -eq 1 ]; then
-    if [ ! -d "$GALAXY_VIRTUAL_ENV" ]
-    then
+    if [ ! -d "$GALAXY_VIRTUAL_ENV" ]; then
         # Locate `conda` and set $CONDA_EXE (if needed). If `python` is Conda Python and $GALAXY_VIRTUAL_ENV does not
         # exist, virtualenv will not be used. setup_python calls this as well but in this case we need it done
         # beforehand.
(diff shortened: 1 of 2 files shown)


mvdbeek (approved):

scripts/common_startup.sh
```diff
@@ -121,12 +120,12 @@ if [ $SET_VENV -eq 1 -a $CREATE_VENV -eq 1 ]; then
                     echo "Creating Conda environment for Galaxy: $GALAXY_CONDA_ENV"
                     echo "To avoid this, use the --no-create-venv flag or set \$GALAXY_CONDA_ENV to an"
                     echo "existing environment before starting Galaxy."
-                    $CONDA_EXE create --yes --name "$GALAXY_CONDA_ENV" 'python=2.7' 'pip>=9' 'virtualenv>=16' -c 'conda-forge'
-                    source activate "$GALAXY_CONDA_ENV"
-                    virtualenv "$GALAXY_VIRTUAL_ENV"
+                    $CONDA_EXE create --yes --override-channels --channel conda-forge --channel defaults --name "$GALAXY_CONDA_ENV" 'python=2.7' 'pip>=9' 'virtualenv>=16'
                     unset __CONDA_INFO
                 fi
+                conda_activate
```
mvdbeek, line 126: We're not activating the `_galaxy_` environment here, so the `virtualenv` is not the special virtualenv installed into `$GALAXY_CONDA_ENV`

nsoranzo: Not sure I understand your comment, we are activating the `$GALAXY_CONDA_ENV` conda environment, which by default is `_galaxy_` (line 117).

mvdbeek: Where is that happening ?

mvdbeek: Ah, OK, conda_activate is not what conda ships, it's a custom function.  OK, I see.

nsoranzo: Yes, in `scripts/common_startup_functions.sh`

PR #7117, opened 2018-12-14
Client initialization pipeline

Nerdinacan:
Okaaay. Sorry for the big PR, but most of it is just one-line file edits, adding a reference to the Galaxy singleton accessor.

**This PR:**

1. Adds an initialization queue to the main application. Configuration variables and initialization functions are added via an exposed config object on the window, allowing initialization functions to be defined anywhere (including python templates) but still allow us to decide when they get executed because I've written us an initialization pipeline (See client/galaxy/scripts/onload/).

2. Because we're using webpack, we must now have a single javascipt entry point for each page which uses the webpack bundles. I've made minor modifications to the python templates to allow for an overridable javascript_entry mako function and provided a generic entry point that works for most situations that calls the initialization queue I described in #1 (above). The expected lifetime for this feature is pretty short given that Dannon and I are about to rework the application container, but still it's there if you need to override it for some reason, (and we do have a few pages that require specialized entry points).

3. I've done some minor clarifications with the webpack build. window.bundledEntries now come from "bundledEntries.js" instead of "extended.js". Entry points explicitly separate out polyfills from main application code.

4. The bulk of the remaining files changed in this PR were simple replacements of the global Galaxy object, which is now obtained by using the getGalaxyInstance function.

**Why the hell did we do this?**

In order to transition from Backbone code to a component-based framework, we need to make sure that models and views are referencing the right galaxy application properties. Unfortunately, the way our backbone code is written, a lot of the scripts reference Galaxy directly and peel a few literal variables off a global galaxy reference, which might be anything given that Backbone definitions execute as soon as the script hits the page, but the existing application writes any number of things to the Galaxy instance depending on when an iframe loaded, or what order a function written directly to the page decides to execute, etc.

Now we've controlled how the central Galaxy variable is initialized and accessed and we can be sure that when some model asks for the Galaxy instance, that the instance has been properly initialized and configured before the model starts querying it for properties.

Initializing existing views via functions does another important thing for us: it enables us to create functional abstractions that can be unit-tested today, and then we can replace their contents with something more modern tomorrow. (And I mean literally tomorrow, December 14th.)

Check out what I did with the masthead code (components/Masthead). We can now register an initialization function (via config.addInitialization() )  that generates the masthead, and in that initialization I have decided to wrap up the existing masthead code into a shallow Vue component container.

It's the dumbest possible Vue component and basically it just runs a bunch of existing Backbone code in a lifecycle method, but this means we now have a separate Vue component that we can unit test and later re-implement it with whatever we want. This allows us to box-up existing code into smaller units, write tests, and replace what's inside those boxes with code we'd rather have.

Which is what we're doing next.


Nerdinacan:
Hmmm, that integration test that's failing ran fine on my nerdinacan/galaxy instance and locally.

nsoranzo:
It's the `ToolBoxTestCase.test_tool_reload_for_broken_tool` unit test which fails randomly ~40% of the times.

@jmchilton @mvdbeek Maybe we could increase the range in `_try_until_no_errors()`?

Nerdinacan:
You are my new hero Nicola. Thanks for triggering that.

jmchilton:
@nsoranzo I think increasing the time won't help - I think the reload is not being triggered and the problem is more subtle and the test case is telling us something useful. But I'll open a PR to try to increase that timeout

Nerdinacan:
Yay it's passing again! I can't really change any of the tags that go on this thing, but this work is no longer in-progress.

dannon:
(already pinged Mason about this on gitter, but for anyone else -- I'm working through a few issues and we will hopefully get this merged tomorrow)

bgruening:
Awesome!!!! Thanks a lot @Nerdinacan and  @dannon!

Issue #7089, opened 2018-12-10
Rendering errors for collections in shared history


jxtx:
Regular datasets render at full width, but collections only fill the width partially.

![image](https://user-images.githubusercontent.com/79973/49757894-85c90e80-fc8b-11e8-9070-894503c4a446.png)

dannon:
Does this happen against dev, too?  I thought I already fixed this in a separate issue, maybe we didn't backport the fix to 18.09?

jxtx:
This is on test which appears to be running dev from Friday or so.

dannon:
Hrmm, definitely wasn't since Friday -- I looked at this a month or more ago when we had all those various positioning issues related to flex.  @guerler I can take a look and dig up what I did before, if you want.

guerler:
Sure, that works for me. I too remember that this issue occured a few weeks ago.

PR #7122, opened 2018-12-17
Fixes state-change animations causing 'block' display 

dannon:
This was appearing in various places due to collection job state fetches.  Includes some import/export refactoring and general client cleanup.

Fixes https://github.com/galaxyproject/galaxy/issues/7089.


dannon:
@guerler Thanks!

PR #7124, opened 2018-12-17
Adds a vue component to display tool errors in the admin panel

guerler:


lib/galaxy/webapps/galaxy/api/tools.py
```diff
@@ -404,6 +405,15 @@ def fetch(self, trans, payload, **kwd):
         create_payload.update(files_payload)
         return self._create(trans, create_payload, **kwd)
 
+    @expose_api
+    @web.require_admin
+    def error_stack(self, trans, **kwd):
+        """
+        POST /api/tools/error_stack
```
dannon: Is this actually a post?  Seems like it should be a GET.

guerler: Yep, thats supposed to be a GET. Thanks.

dannon: Cool, thought so!

jmchilton:
screen shot 2019-01-02 at 10 03 16 am

Works for me!

PR #7125, opened 2018-12-17
Use singularity build / exec with biocontainers

mvdbeek:
This makes it possible to pull a container from
quay.io/biocontainers and to convert it to singularity format
for later execution (via the CachedMulledSingularityContainerResolver),
or to run a container straight from the quay.io registry.

This is nicely integrated with the DependencyManager view, so
you can (as for conda or docker) do:
```
requests.post(galaxy_url+'/api/tools/deseq2/dependencies', {'key': API_KEY, 'resolver_type': 'mulled_singularity'})
```

This is quite lightweight, the resulting image is significantly smaller
than the conda env (since it is compressed), and the install process
is much faster.


lib/galaxy/config.py
```diff
@@ -1054,6 +1056,9 @@ def _configure_toolbox(self):
             involucro_auto_init=self.config.involucro_auto_init,
         )
         self.container_finder = containers.ContainerFinder(app_info)
+        self.toolbox.dependency_manager.resolver_classes.update(self.container_finder.container_registry.resolver_classes)
+        self.toolbox.dependency_manager.dependency_resolvers.extend(self.container_finder.container_registry.container_resolvers)
```
mvdbeek: This will append enabled container resolvers to the regular dependency resolvers and therefore determine the order in which resolvers are tried. Do we think this is good enough or do we want finer-grained control, and if yes, should we maybe merge the container_resolvers_conf.xml file into the dependency_resolver_conf.xml file ?

Of note the regular dependency resolver will pick a dependency, but the container resolver runs after that and might overwrite the chosen dependency (this is pre-existing, but I guess that would be also up to debate ?).

I think for my own needs this is fine as is, so let me know.

lib/galaxy/tools/deps/container_classes.py
```diff
@@ -0,0 +1,357 @@
+import os
```
jmchilton, line 1: Good idea to move this.

jmchilton:
Merging these two concepts (container resolution and dependency resolution) together makes me really uncomfortable. Will the use case where Conda auto-install of resolved requirement into a specified container still work? I'd also like to make progress toward dependency resolvers and container resolvers being configurable on a per-destination basis - does this make any assumptions that set us back there? I doubt it does I'm just a bit confused by the DependencyManager taking in a set of container types it can resolve against - it feels like those could be entirely dynamic within the destination and variable across destinations - so I don't understand yet how that works? I need to sit with it for longer, I'm sure you're four steps ahead of me here and I just need to catch up.

mvdbeek:
> Merging these two concepts (container resolution and dependency resolution) together makes me really uncomfortable.

I'm hoping merging this makes it simpler to understand for admins -- the fact that one is a container and the other is not is just a implementation detail, IMO. I think it'd be common for admins wanting to have dependencies for one tool satisfied via Conda / Modules etc. and another via Docker or Singularity.

> Will the use case where Conda auto-install of resolved requirement into a specified container still work?

Yes, that should still work fine, I'll add a test for this.

> I'd also like to make progress toward dependency resolvers and container resolvers being configurable on a per-destination basis - does this make any assumptions that set us back there?

I don't think there'd be any issue, you could extend the Dependency Manager by passing in the destination info object once it is available. I'm doing this little hack of just passing the first available destination info into the dependency manager to get an idea of what is enabled and with which parameters, but we could be smarter here.

> it feels like those could be entirely dynamic within the destination and variable across destinations

I agree that this isn't well taken care of via the DependencyManager at this point, but it doesn't have to be IMO. The container resolver still kicks in as before. If this is completely dynamic you're probably not going to micro-manage what is installed and what is not installed, and that's the primary reason for adding container resolvers to the list of dependency resolvers in the DependencyManager.

I'm not so sure I have thought about everything, but I think this should not block any future developments.

mvdbeek:
@jmchilton I've added some more integration tests, but the `test_build_mulled` test can't run on Jenkins because it needs Docker. If all works as it does locally the other singularity tests though should run on Jenkins.

mvdbeek:
Hmm, looks like something's going wrong when running on Jenkins:
```
|  #x1B[91mERROR  : Singularity is not running with appropriate privileges!
|  #x1B[0m#x1B[91mERROR  : Check installation path is not mounted with 'nosuid', and/or consult manual.
|  #x1B[0m#x1B[31mABORT  : Retval = 255
```

This is working fine on my mac, I wonder if this could be a docker version issue ?
I've got:
`Docker version 18.06.1-ce, build e68fc7a`
while the infrastructure playbook is installing docker.io, which should be 1.6.2

config/container_resolvers_conf.xml.sample
```diff
@@ -21,15 +21,23 @@
...
        mapping. Should be the method of choice if mulled resolution is required
-       but the container execution is handled by a container orchestration 
-       instead of docker directly (ie. Galaxy has no access to a docker client 
+       but the container execution is handled by a container orchestration
+       instead of docker directly (ie. Galaxy has no access to a docker client
        binary). It works of course with direct docker engine tool executions.
        -->
 
+  <!-- <mulled_singularity /> -->
+  <!-- mulled: only resolves container URI through bioconda to mulled automatic
```
nsoranzo:
```suggestion
  <!-- mulled_singularity: only resolves container URI through bioconda to mulled automatic
```

config/container_resolvers_conf.xml.sample
```diff
@@ -21,15 +21,23 @@
...
+       mapping. Should be the method of choice if mulled resolution is required
+       but the container execution is handled by a container orchestration
+       instead of singularity directly (ie. Galaxy has no access to a singularity client
+       binary). It works of course with direct singluarity tool executions.
```
nsoranzo:
```suggestion
       binary). It works of course with direct singularity tool executions.
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -414,7 +414,7 @@ mapping:
         default: database/dependencies/involucro
         required: false
         desc: |
-          involucro is a tool used to build Docker containers for tools from Conda
+          involucro is a tool used to build Docker ir Singularity containers for tools from Conda
```
```suggestion
          involucro is a tool used to build Docker or Singularity containers for tools from Conda
```

mvdbeek:
So a simple `docker run --privileged --entrypoint /usr/bin/singularity mvdbeek/testing-base:19.01.0 exec docker://busybox hostname` works on jenkins. Which makes me think there's something weird going on when we create the `galaxy` user during startup. Very strange though that this works fine on OSX (version 18.06.1-ce) and Ubuntu 16.04 (version 18.09.0) ... maybe 17.05.0-ce that is on Jenkins is too old ?

mvdbeek:
Hmm, a little bit of progress, turns out that on OSX and the Ubuntu server it worked fine because singularity was falling back to user namespaces and wasn't using the setuid mechanism. Which is pretty cool, but a new and consistently implemented feature, so this is the error I saw on Jenkins:
```
Failed invoking the NEWUSER namespace runtime: Operation not permitted
```

So a manual build from sources brought this a little further, but still fails with
```
ERROR   [U=999,P=39]       priv_escalate()               Failed to set effective UID to 0
```

I still have no idea why this doesn't work on Jenkins but works locally.

bgruening:
@mvdbeek sorry haven't had time to look at it atm, but from the error above you need to to the following.

`find $PREFIX/libexec/singularity -type f -name '*-suid' -exec chown root:root {} \; -exec chmod u+s {} \;`

mvdbeek:
I'll try, but I think that's what `sudo make install` does.

mvdbeek:
That wasn't it, same error.

mvdbeek:
I've added some additional debugging to singularity, and it comes out with:

```
Failed to set effective UID to 0: Operation not permitted
```

Again, this is working fine locally -- I wonder if there is some seccomp or apparmor profile that blocks this (despite passing --privileged to docker run)

mvdbeek:
OK I give up, @natefoo do you have any idea what could be blocking `seteuid` on the Jenkins nodes inside docker ?

mvdbeek:
(the error is coming from here: https://github.com/mvdbeek/singularity/blob/master/cmd/starter/c/starter.c#L121)

mvdbeek:
For the record the issue was that `/var/lib/docker` was mounted with the `nosuid` option, thanks for the help @natefoo !

mvdbeek:
At this point I think what is left to implement is better control over what dependency/container resolver will be used. I have pulled or built almost all dependencies that I need on our server using the MulledSingularity and BuildMulledSingularity resolvers, but there are a couple of dependencies that should remain Conda (or ToolShed ?) packages (for instance because the Conda package doesn't specify all dependencies, and so the containers are broken but work fine if used as a regular Conda package).

More importantly I think most admins wouldn't want to go through the exercise of moving all dependencies to docker/singularity for existing and known working dependencies, and would instead like to get containers only for newly installed or updated tools. As it is now the container resolver runs regardless of the dependency resolution status, and if the Non-Cached resolver finds a match it will pull or build a container for that dependency. And repeatedly attempting to build a broken container I absolutely want to avoid.

I'd like to be able to control the order of resolution as I do in the `dependency_resolvers_conf.xml` file, with Conda first for the few dependencies that need it, followed by CachedMulledSingularity (uses container images from the filesystem), MulledSingularity (pulls containers from quay.io) and BuildMulledSingularity (uses involucro to build containers from conda packages).

I'd be tempted to include container resolvers in the `dependency_resolvers_conf.xml` file so that I can have full control over the order, but that's just because I can't see a better solution here.

Do you have any suggestions / wishes @jmchilton @bgruening @erasche @natefoo ?

EDIT:
Another way to go about this would be to introduce an `auto_install` flag on each resolver, so I could turn this off for the non-cached variants. That'd probably the most straightforward ...

mvdbeek:
Alright, I've implemented a `auto_install` flag for the resolvers than can pull or build images. This defaults to `True`, since that's how it worked up until this point. If you set it to `False` you can still use the API (and hopefully soon the UI) to trigger a pull or build.
That satisfies my requirements, as long as I don't have a cached singularity image for a dependency the installed Conda package will be used.

mvdbeek:
Yay, green! This is ready for review.

jmchilton:
This is impressive, amazing, and a little terrifying - I trust you though @mvdbeek. Thanks for the mountains of work that went into this - this is quite huge.

hexylena:
> More importantly I think most admins wouldn't want to go through the exercise of moving all dependencies to docker/singularity for existing and known working dependencies

Not true for us? We would prefer to have everything on singularity, for the ability to apply restrictions to mounts/paths available to executed tools, even if this means going through some work to get dependencies working in singularity.

mvdbeek:
Sure, I assume the time and expertise is there for a couple of sites, it's just not great for admins that want to take the one at a time route for new dependencies. I also think it'll be very time-consuming with little payoff if you want to do this for old revisions (unless that one huge container idea works / or mounting in packages into the container). 

The last commit introduced the `auto_install` flag, which makes this a moot point. Set it to `false` and put the container resolver below Conda / TS and you can happily mix dependencies (i.e new tools containers, don't touch old tools).

jmchilton:
I'm bumping into a lot of problems with this when trying to implement https://github.com/galaxyproject/galaxy/pull/8741... perhaps just in my head. I can replicate the current grid using the API and VueUJS but I dislike the merging of container resolvers and dependency resolvers into a single concept here - it frequently is simply incorrect. It ignores certain kind of container resolution, the requirements of a tool influence the container resolved but not exclusively and that seems to be ignored, containers will resolve before dependencies but this isn't enforced here (right?), no job config information is used here so containers are displayed even if they are only used by a single destination, tools with no requirements but having a container don't show resolution at all, the way it infers valid container types is an incorrect approximation given dynamic job destinations, ...

I came back here to re-read the comments and my reservations still hold and have been made more concrete. I really just want to undo the merging of the resolvers and provide separate interfaces for containers and dependencies so we can provide "correct" information. The part that gives me pause is:

> I'm hoping merging this makes it simpler to understand for admins -- the fact that one is a container and the other is not is just a implementation detail, IMO. 

It is hard to disagree with this... but the grid right now is so incorrect I just don't feel it has accomplished this. It seems most admins will have an all container Galaxy or an all Conda Galaxy and for the admins that have mixed setups - they are more sophisticated and being able to summarize correct, actionable information is even more important.

PR #7127, opened 2018-12-18
Update help text as per suggestion from CWunder

pvanheus:
This [thread](https://help.galaxyproject.org/t/clearer-documentation-for-filter-list/231) suggests that the help text for this tool is not clear. This PR attempts to address that weakness.

### DIFF
--- a/lib/galaxy/tools/filter_from_file.xml
+++ b/lib/galaxy/tools/filter_from_file.xml
@@ -81,9 +81,11 @@
     
 .. class:: infomark
 
-        This tool will take an input list and a text file with identifiers to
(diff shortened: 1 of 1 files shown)


martenson:
thanks @pvanheus !

nsoranzo:
Not sure if replacing `identifier` with `name` if correct, ping @jmchilton.

jmchilton:
Yeah - this is wrong in my opinion. I think it was more precise and correct before (ping @mvdbeek for a second opinion).

pvanheus:
I think "identifier" can be used in place of "name" while the rest of the edit contributes to clarifying what this tool does. If "identifier" is used perhaps it can be explained in context?

PR #7128, opened 2018-12-18
[18.05] Fix for use of `log.trace` in Slurm runner

natefoo:
Not sure why this hasn't cropped up before but I was hitting it in development.

### DIFF
--- a/lib/galaxy/jobs/runners/slurm.py
+++ b/lib/galaxy/jobs/runners/slurm.py
@@ -1,7 +1,6 @@
 """
 SLURM job control via the DRMAA API.
 """
-import logging
 import os
 import re
 import shutil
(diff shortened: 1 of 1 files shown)


dannon (approved):

bernt-matthias:
SInce I would like to learn a bit I have to ask: why is  using `logging.getLogger` a problem? There are so many places where this is used...

natefoo:
@bernt-matthias It's only a problem if you want to use `log.trace()`, which is a non-standard extension that's only available through `galaxy.util.logging`. It's hardly used at the moment, unfortunately.

bernt-matthias:
cool. thanks

bgruening:
Second :+1: from me. Feel free to merge.

PR #7134, opened 2018-12-19
Remove unneeded session flushes and adds when discovering outputs.

jmchilton:
Should be slightly more efficient, it is more about refactoring the direct sa_session interactions out downstream.


dannon:
For anyone else checking, the `primary_data` add already happens in `_new_hda`, so it's not necessary again here.

PR #7139, opened 2018-12-20
Fix authentication provider exception not logged

BioinformaticsPlatformWIV-ISP:
Fix the not logged exception during an external provider authentication  ->  log and reraise the exception properly instead of ignoring it when the provider raise one.

### DIFF
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -1241,6 +1241,13 @@ class Tool(Dictifiable):
             self.__ensure_help()
         return self.__help_by_page
 
+    @property
+    def raw_help(self):
+        # may return rst (or Markdown in the future)
+        tool_source = self.__help_source
+        help_text = tool_source.parse_help()
+        return help_text
+
     def __ensure_help(self):
         with HELP_UNINITIALIZED:
             if self.__help is HELP_UNINITIALIZED:
--- a/lib/galaxy/tools/search/__init__.py
+++ b/lib/galaxy/tools/search/__init__.py
@@ -6,7 +6,6 @@ import logging
 import re
 import tempfile
 
-from bleach import clean
 from whoosh import analysis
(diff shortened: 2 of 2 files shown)


lib/galaxy/auth/__init__.py
```diff
@@ -66,7 +66,8 @@ def check_auto_registration(self, trans, login, password, no_password_check=Fals
                     try:
                         auth_return = parse_auth_results(trans, auth_results, options)
                     except Conflict:
-                        break
+                        log.exception(Conflict)
```
jmchilton: This is in a for loop, I'd rework this. Also the exception should have a message not just log a class name. I'm going to mark this as "work in progress" waiting on those changes. We're (the committers) unsure about whether this should be logged at all though.

jvanbraekel: Maybe I should mention that the 'to string'  ("_ _str_ _") operator is overridden in the base 'MessageException' class which 'Conflict' derive from.  Thus the error message is logged, not the class name as it might seems at first.

nsoranzo: `Conflict` is still only the class, not the actual exception object, which you can get with e.g. `except Conflict as e:`

jvanbraekel: Good catch. I have updated the code.

nsoranzo:
This PR probably needs to be rebased for CircleCI tests to run.

BioinformaticsPlatformWIV-ISP:
We have tried to rebase it, as result all the changes were lost inside ours local branch, and the PR closed.  So we reverted the changes inside our repository but we can't re-open the PR.

nsoranzo:
@BioinformaticsPlatformWIV-ISP It seems that (per https://github.com/isaacs/github/issues/361 ) it is easier to open a new pull request from your branch.

PR #7144, opened 2018-12-20
Remove collect_outputs_from option.

jmchilton:
And just disable dynamic discovery of outputs from new_file_path for legacy profile tools.

This is part of an ongoing effort to kill new_file_path and simplify output discovery.

This option was added by @natefoo 10 years ago and he switched main to disable new_file_path discovery and noone ever complained. So lets just do this for all Galaxies - if main hasn't run tools that require this functionality in a decade - I very much doubt any Galaxies require this.

### DIFF
--- a/config/galaxy.yml.sample
+++ b/config/galaxy.yml.sample
@@ -518,14 +518,6 @@ galaxy:
   # store this information.
   #citation_cache_lock_dir: database/citations/lock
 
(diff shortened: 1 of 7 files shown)


dannon:
Sounds good to me.  @natefoo?

natefoo:
Belated +1

PR #7145, opened 2018-12-20
Use regular js-event handling instead of jquery for uploader

### LINKED ISSUE #7138: Drag into upload broken in dev
If you try dragging a file into the upload field the browser just displays the file.
The workflow drag handler works fine though.

guerler:
Fixes #7138.


jmchilton:
Verified bug and that this fixes it, thanks so much @guerler !

PR #7147, opened 2018-12-21
make job script actually return tool exit code

bernt-matthias:
Problem: Galaxy job scripts currently look like this:

```
...
JOB_WORKING_DIR/tool_script.sh; 
return_code=$?; 
...
sh -c "exit $return_code"
echo $? > JOB_WORKING_DIR/galaxy_JOBID.ec
date +"%s" > JOB_WORKING_DIR/__instrument_core_epoch_end
```

Since the `sh -c "exit $return_code"` is not the last command it has no effect (on the exit code of the job script) and the job_script returns the exit code of `date` (ie 0). More precisely the part 

```
sh -c "exit $return_code"
echo $? > JOB_WORKING_DIR/galaxy_6831.ec
```

has the same effect as `echo $return_code > JOB_WORKING_DIR/galaxy_JOBID.ec`

The problem then is that the job runner can not detect the potentially non-zero exit code of the tool script from the exit code of the job script. Job runners that get the exit code from galaxy_JOBID.ec are not affected, all others are, currently this is the univa job runner and it also seems that some cli runners are affected. So all features of those runners that depend of the runners ability to obtain the tool exit code don't work at the moment. 

My solution idea is to record the tool exit code always (should not hurt) and exit with the tool exit code... 

TODOs: 
- [ ] some code comments may need updates

### DIFF
--- a/lib/galaxy/jobs/command_factory.py
+++ b/lib/galaxy/jobs/command_factory.py
@@ -15,7 +15,6 @@ from galaxy.jobs.runners.util.job_script import (
 log = getLogger(__name__)
 
 CAPTURE_RETURN_CODE = "return_code=$?"
-YIELD_CAPTURED_CODE = 'sh -c "exit $return_code"'
 SETUP_GALAXY_FOR_METADATA = """
 [ "$GALAXY_VIRTUAL_ENV" = "None" ] && GALAXY_VIRTUAL_ENV="$_GALAXY_VIRTUAL_ENV"; _galaxy_setup_environment True
 """
@@ -253,8 +252,6 @@ class CommandsBuilder(object):
             self.append_command(CAPTURE_RETURN_CODE)
 
     def build(self):
-        if self.return_code_captured:
-            self.append_command(YIELD_CAPTURED_CODE)
         return self.commands
 
 
--- a/lib/galaxy/jobs/runners/util/job_script/DEFAULT_JOB_FILE_TEMPLATE.sh
+++ b/lib/galaxy/jobs/runners/util/job_script/DEFAULT_JOB_FILE_TEMPLATE.sh
@@ -37,5 +37,7 @@ cd $working_directory
 $memory_statement
 $instrument_pre_commands
 $command
-echo $? > $exit_code_path
+echo $return_code > $exit_code_path
 $instrument_post_commands
+# exit with the exit code captured for the tool_script
+sh -c "exit $return_code"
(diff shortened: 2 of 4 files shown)


jmchilton:
This is a good simplification, but...

>Job runners that get the exit code from galaxy_JOBID.ec are not affected

This I think should be the standard behavior and the source of truth. I'd fixed the other runners to respect this is possible rather than counting on the exit code of the job script.

bernt-matthias:
Thanks. Was thinking about if getting the exit code from the exit code file could/should be done also for the univa job runner. The main point that bugs me is that UNIVA encodes more information in the exit code, e.g., if the job was killed by UNIVA. I documented this here: 

https://github.com/galaxyproject/galaxy/blob/11bbcfa87d014f80f10153a2fcca7a4f15951cf8/lib/galaxy/jobs/runners/univa.py#L260

So I guess in this case I should leave it as is and use the exit code from UNIVA which is now correct.

mvdbeek:
> This I think should be the standard behavior and the source of truth. I'd fixed the other runners to respect this is possible rather than counting on the exit code of the job script.

For that reason I'm going to revert this commit in #6958. It also breaks exit code and stdout/stderr and job failure detection in the CLI runner.

bernt-matthias:
OK. Thanks for letting me know. Then I need to review the UNIVA runner. 

Isn't the default content of 0 (https://github.com/galaxyproject/galaxy/blob/6778fa47ed899eefe1418139f43832f225760e0f/lib/galaxy/jobs/runners/__init__.py#L698) a bit "optimistic". I can only imagine that this line triggers if the job was canceled during the execution of `tool_script.sh` .. on the other hand these cases should be handled by the runner.

mvdbeek:
It is, but I don't think that exit code ever gets used. If the job runner really killed the job you'd not get there, you'd call fail directly.

mvdbeek:
See https://github.com/galaxyproject/galaxy/blob/6778fa47ed899eefe1418139f43832f225760e0f/lib/galaxy/jobs/runners/drmaa.py#L240  / https://github.com/galaxyproject/galaxy/blob/6778fa47ed899eefe1418139f43832f225760e0f/lib/galaxy/jobs/runners/pbs.py#L408 /  https://github.com/galaxyproject/galaxy/blob/936679ea817151438ff76d2e14ac2557c4caa3a0/lib/galaxy/jobs/runners/kubernetes.py#L459

PR #7148, opened 2018-12-21
Correct code error in 'build job runner' tutorial

jdavcs:
When  __init__ is called, **kwargs (most likely) will not contain
the 'runner_param_specs' key. And if so, calling update() on
kwargs['runner_param_specs'] will cause an error. Check if key exists,
then update or create.

### DIFF
--- a/doc/source/dev/build_a_job_runner.rst
+++ b/doc/source/dev/build_a_job_runner.rst
@@ -120,10 +120,7 @@ B: Update the dictionary structure in kwargs.
 
 ::
 
-    if 'runner_param_specs' in kwargs:
-        kwargs['runner_param_specs'].update(runner_param_specs)
-    else:
-        kwargs['runner_param_specs'] = runner_param_specs
+    kwargs.update({'runner_param_specs': runner_param_specs})
 
 C: Now call the parent constructor to assign the values.
 
--- a/doc/source/dev/build_a_job_runner.rst
+++ b/doc/source/dev/build_a_job_runner.rst
@@ -120,7 +120,10 @@ B: Update the dictionary structure in kwargs.
 
 ::
 
-    kwargs['runner_param_specs'].update(runner_param_specs)
+    if 'runner_param_specs' in kwargs:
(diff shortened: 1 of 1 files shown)


dannon:
Admittedly haven't tried the actual job runner tutorial, but update should create *or* update as necessary, right?

```
yoplait@malleus:~/work/galaxy python                                                                                                                   ± dev
Python 2.7.15 (default, Jun 17 2018, 12:46:58)
[GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> z = {"a": 1}
>>> z.update({"b":2})
>>> z
{'a': 1, 'b': 2}
```

jdavcs:
> Admittedly haven't tried the actual job runner tutorial, but update should create _or_ update as necessary, right?
> 

Yes. At least that's my understanding - I'm currently working through it.

dannon:
~Should maybe just be `kwargs.update(runner_param_specs)`~  Nevermind, it's a nested dict.  Yeah, I'd say to tinker with it and see what's up.

jdavcs:
> Should maybe just be `kwargs.update(runner_param_specs)`

I don't think so: then it would add runner_param_specs as key/value pairs (it is a dict) to kwargs; but it needs to add *one* key/value pair, with 'runner_param_specs' as the key. I think.

dannon:
@ic4f Yeah, that's what I was saying wouldn't work in the edit; you're going to want something like:  `kwargs.update({"runner_param_specs": runner_param_specs})`

(in any event, thank you for fixing up docs! :) )

jdavcs:
I see, yes, that makes sense. Let me double check, and I'll change it.

jdavcs:
@dannon Thanks. Updated.

natefoo:
Sorry for the late reply - this does do one unintended thing: it completely replaces the previous value (a dict) of `kwargs['runner_param_specs']`. The old example would fail if the `runner_param_specs` key didn't exist, but it would succeed and only *update* that value if it did.

PR #7149, opened 2018-12-21
added store library to exported libs in bundleEntries

Nerdinacan:
window.bundleEntries is missing an npm resource in common.mako. This is just a localstorage wrapper, but it's used by common.mako in existing on-page scripting.


dannon:
xref https://github.com/galaxyproject/galaxy/pull/6994/files; looks like there was a bad merge somewhere that dropped this back out of extended.js

PR #7150, opened 2018-12-23
Follow-up on handling connectable and collapsible parameters in workflows

guerler:
Connecting basic input parameters in the workflow editor is an awesome feature. Thanks a lot for adding it @jmchilton. This PR fixes a few remaining issues with the handling of connected and collapsible value options in the workflow editor. Currently connected values are not always updated after a click i.e. the terminal is not added to the module if the value was collapsed before. This required to re-click the option, to uncollapse and then re-trigger the connection request. This PR resolves this by fixing the refresh trigger and yielding priority to the connected option over the collapsible option i.e. if a parameter is connected it is connected regardless of it's prior collapsible state. Another addition ensures that only collapsible values can be connected. This avoids that e.g. data parameters are shown with a connectable option although they are always connected. Additionally, this PR fixes the tooltips which currently have two issues. They do not contain the correct text and they disappear after an option is selected. The latter being an issue of the BS4 update. Finally the pointer cursor is removed from the title text since it is not clickable anymore and margins are added between the option icons.

### DIFF
--- a/client/galaxy/scripts/mvc/form/form-data.js
+++ b/client/galaxy/scripts/mvc/form/form-data.js
@@ -12,7 +12,7 @@ export var Manager = Backbone.Model.extend({
             var id = $(this).attr("id");
             var field = self.app.field_list[id];
             if (field) {
-                sum += `${id}:${JSON.stringify(field.value && field.value())}:${field.collapsed};`;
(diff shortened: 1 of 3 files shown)


jmchilton:
Wonderful - thanks for fixing up my mess @guerler - sorry for not getting to it sooner.

PR #7154, opened 2018-12-26
Allow storing objects in the object store by UUID.

jmchilton:
This is the most obvious way to allow creating new datasets during a job creation or finish context without a database available (see #7050 / #7058 for some context).

I'm working on the job completion context first here and in this context the fixed known outputs will have database IDs but other datasets such as discovered primary datasets and dynamic collection contents will not have IDs. So if the job or Pulsar or whatever is going to write these datasets to the object store they need to be keyed on something other than ID (at least initially).

This PR builds on #7110.

### DIFF
--- a/config/galaxy.yml.sample
+++ b/config/galaxy.yml.sample
@@ -530,6 +530,11 @@ galaxy:
   # it overrides any other objectstore settings.
   #object_store_config_file: config/object_store_conf.xml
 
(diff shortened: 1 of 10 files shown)


mvdbeek (approved):

mvdbeek:
@galaxybot test py3

PR #7156, opened 2018-12-26
Stronger ToolProvidedMetadata abstraction.

jmchilton:
- Move to its own file (goal is to be importable from set_metadata without baggage from tools and jobs modules). See #7058 for the downstream implementation of this (e.g. https://github.com/galaxyproject/galaxy/pull/7058/commits/9300191fe944fce388b9f06fdf6ed32ce1c4d98a).
- Proper interface with some basic documentation.

Work from #7058  / #7080.


mvdbeek:
There's a merge conflict here.

lib/galaxy/tools/provided_metadata.py
```diff
@@ -0,0 +1,185 @@
...
+import os
+
+
+log = logging.getLogger(__name__)
+
+
+def parse_tool_provided_metadata(meta_file, provided_metadata_style=None, job_wrapper=None):
+    """Return a ToolProvidedMetadata object for specified file path.
+
+    If path is absent, return a NullToolProvidedMetadata. If type is None
```
mvdbeek:
```suggestion
    If meta_file is absent, return a NullToolProvidedMetadata. If provided_metadata_style is None
```

lib/galaxy/tools/provided_metadata.py
```diff
@@ -0,0 +1,185 @@
...
+        return ToolProvidedMetadata(meta_file)
+
+
+def _guess_tool_provided_metadata_style(path):
+    try:
+        with open(path, "r") as f:
+            metadata = json.load(f)
+        metadata_type = metadata.get("type", None)
+        return "legacy" if metadata_type in ["dataset", "new_primary_dataset"] else "default"
+    except ValueError:
```
mvdbeek, line 34: Could you move this up until after `metadata = json.load(f)` ?

lib/galaxy/tools/provided_metadata.py
```diff
@@ -0,0 +1,185 @@
...
+        return found_failed
+
+    def get_unnamed_outputs(self):
+        return []
+
+
+class ToolProvidedMetadata(object):
+
+    def __init__(self, meta_file):
+        with open(meta_file, 'r') as f:
```
mvdbeek, line 139: if this code is going to run on containers with broken UTF-8 defaults (many/most of them are broken unfortunately) this will fall back to assuming ASCII encoding -- and I guess that'll fail. Probably best to open this and other places where we read metadata in utf-8 using `with io.open(meta_file, mode='r', encoding='utf-8')`. Though @nsoranzo might have a better suggestion here.

lib/galaxy/tools/provided_metadata.py
```diff
@@ -0,0 +1,185 @@
...
+        """For a discovered dataset, get the corresponding metadata entry.
+
+        The discovery may have been from explicit listing in this file (returned
+        from get_new_datasets) or via file regex, either way the basename of the
+        file is used to index the fetching of the metadata entry.
+        """
+        return {}
+
+    def get_unnamed_outputs(self):
+        """Return unnamed outputs dataset structured introduced for upload 2.0.
```
mvdbeek: Should this be 
```suggestion
        """Return unnamed outputs dataset introduced for upload 2.0.
```
?

mvdbeek (approved):
Nice, just got some minor comments -- the UTF-8 stuff is not a new problem in any way. I'm actually surprised this has passed py3 tests now that I think about it.

mvdbeek:
@galaxybot test py3

jmchilton:
Like said above, these are pre-existing issues. I have more PRs on this same code coming so I will try to address the UTF-8 issue there - I've opened an issue for these (https://github.com/galaxyproject/galaxy/issues/7362) and put it on the remote execution board I'm working on.

lib/galaxy/tools/provided_metadata.py
nsoranzo: @mvdbeek Completely agree that your suggested code is the optimal Python 2&3-compatible way to open a text file which may contain Unicode characters.

jmchilton: Hmm.. I created an issue for this but revisiting it this code doesn't actually run in the biocontainers though right? The metadata stuff will always need to be ran from a side car container? I'm not sure we need to worry about this - but if either of y'all want me to make the change I can.

PR #7157, opened 2018-12-26
Do not use default values for runtime values in workflow execution form

### LINKED ISSUE #1888: Set parameter at runtime shows the default parameter
With the new form the default value of a parameter is shown, but it should be empty or not-set. Usually this feature is used to require a user action before running a workflow, but with the current behaviour a default value is used which can yield unexpected results.

guerler:
Fixes #1888.


jmchilton:
I think this broke test_workflow_run.py::test_runtime_parameters_simple ... but not consistently? I'm confused because the test is in this branch and green - but it has started failing other places now pretty consistently. I'm confused... maybe we should just disable this test for now.

guerler:
Ok, I'll debug it a little bit and then will issue a PR for that test case.

jmchilton:
I think it is fixed with https://github.com/galaxyproject/galaxy/pull/7188

PR #7158, opened 2018-12-27
Cleanup, test, and fix metadata collection code.

jmchilton:
There are a trio of important metadata commits in #7058 for #7050. One [rearranges set_metadata.py to use ToolProvidedMetadata instead of custom parsing of galaxy.json](https://github.com/galaxyproject/galaxy/commit/e7b188f61395210775c614d47446e6a9624378d0), one [introduces a new method of metadata generation that is portable and more robust](https://github.com/galaxyproject/galaxy/commit/1a73696652ce923780e93e7f9839b3919d2744ad), and one [extends this new method to capture much more of the job finish code as part of metadata collection](https://github.com/galaxyproject/galaxy/commit/33daa6a68c0ed86d49fcc7a8a7e9f1c03370e031).

This PR is refactors, introduce extension points and unit tests, and fixes bugs that set the stage for doing all of this confidently and cleanly in the that branch. I debated whether the a unit test for metadata collection was needed because it isn't very isolated - but given that I've already caught a bug that was covered up by job finish code handling metadata errors fairly robustly - I've been convinced this testing is very valuable despite the large suite of API/framework tests we have. The API/framework tests both mask metadata collection errors but also make it hard to debug metadata problems due to all the levels of indirection with these errors propagating back up through the job finish code and pytest test reporting.

### DIFF
--- a/lib/galaxy/datatypes/metadata.py
+++ b/lib/galaxy/datatypes/metadata.py
@@ -11,7 +11,6 @@ from galaxy.model.metadata import (
     DBKeyParameter,
     DictParameter,
     FileParameter,
-    JobExternalOutputMetadataWrapper,
     ListParameter,
     MetadataCollection,
     MetadataElement,
@@ -42,5 +41,4 @@ __all__ = (
     "PythonObjectParameter",
     "FileParameter",
     "MetadataTempFile",
-    "JobExternalOutputMetadataWrapper",
 )
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -24,11 +24,12 @@ from pulsar.client.staging import COMMAND_VERSION_FILENAME
 
 import galaxy
 from galaxy import model, util
-from galaxy.datatypes import metadata, sniff
+from galaxy.datatypes import sniff
 from galaxy.exceptions import ObjectInvalid, ObjectNotFound
 from galaxy.jobs.actions.post import ActionBox
 from galaxy.jobs.mapper import JobRunnerMapper
 from galaxy.jobs.runners import BaseJobRunner, JobState
+from galaxy.metadata import get_metadata_compute_strategy
 from galaxy.objectstore import ObjectStorePopulator
 from galaxy.util import safe_makedirs, unicodify
 from galaxy.util.bunch import Bunch
@@ -702,7 +703,7 @@ class JobWrapper(HasResourceParameters):
         self.output_hdas_and_paths = None
         self.tool_provided_job_metadata = None
         # Wrapper holding the info required to restore and clean up from files used for setting metadata externally
-        self.external_output_metadata = metadata.JobExternalOutputMetadataWrapper(job)
+        self.external_output_metadata = get_metadata_compute_strategy(self.app, job.id)
         self.job_runner_mapper = JobRunnerMapper(self, queue.dispatcher.url_to_destination, self.app.job_config)
         self.params = None
         if job.params:
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1363,23 +1363,7 @@ class JobWrapper(HasResourceParameters):
                             not self.external_output_metadata.external_metadata_set_successfully(dataset, self.sa_session)):
                         dataset._state = model.Dataset.states.FAILED_METADATA
                     else:
-                        # load metadata from file
-                        # we need to no longer allow metadata to be edited while the job is still running,
-                        # since if it is edited, the metadata changed on the running output will no longer match
-                        # the metadata that was stored to disk for use via the external process,
-                        # and the changes made by the user will be lost, without warning or notice
-                        output_filename = self.external_output_metadata.get_output_filenames_by_dataset(dataset, self.sa_session).filename_out
-
-                        def path_rewriter(path):
-                            if not path:
-                                return path
-                            normalized_remote_metadata_directory = remote_metadata_directory and os.path.normpath(remote_metadata_directory)
-                            normalized_path = os.path.normpath(path)
-                            if remote_metadata_directory and normalized_path.startswith(normalized_remote_metadata_directory):
-                                return normalized_path.replace(normalized_remote_metadata_directory, self.working_directory, 1)
-                            return path
-
-                        dataset.metadata.from_JSON_dict(output_filename, path_rewriter=path_rewriter)
+                        self.external_output_metadata.load_metadata(dataset, dataset_assoc.name, self.sa_session, working_directory=self.working_directory, remote_metadata_directory=remote_metadata_directory)
                     line_count = context.get('line_count', None)
                     try:
                         # Certain datatype's set_peek methods contain a line_count argument
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1752,9 +1752,14 @@ class JobWrapper(HasResourceParameters):
         if datatypes_config is None:
             datatypes_config = os.path.join(self.working_directory, 'registry.xml')
             self.app.datatypes_registry.to_xml_file(path=datatypes_config)
-        command = self.external_output_metadata.setup_external_metadata([output_dataset_assoc.dataset for
-                                                                         output_dataset_assoc in
-                                                                         job.output_datasets + job.output_library_datasets],
+
+        output_datasets = {}
+        for output_dataset_assoc in job.output_datasets + job.output_library_datasets:
+            output_name = output_dataset_assoc.name
+            assert output_name not in output_datasets
+            output_datasets[output_dataset_assoc.name] = output_dataset_assoc.dataset
+
+        command = self.external_output_metadata.setup_external_metadata(output_datasets,
                                                                         self.sa_session,
                                                                         exec_dir=exec_dir,
                                                                         tmp_dir=tmp_dir,
--- a/lib/galaxy/metadata/__init__.py
+++ b/lib/galaxy/metadata/__init__.py
@@ -1,0 +1,216 @@
+"""Define abstraction for capturing the metadata of job's output datasets."""
+
+import json
+import os
+import shutil
+import tempfile
+from logging import getLogger
+from os.path import abspath
+
+from six.moves import cPickle
+
+import galaxy.model
+from galaxy.model.metadata import FileParameter, MetadataTempFile
+from galaxy.util import in_directory
+
+log = getLogger(__name__)
+
+
+def get_metadata_compute_strategy(app, job_id):
+    return JobExternalOutputMetadataWrapper(job_id)
+
+
+class JobExternalOutputMetadataWrapper(object):
+    """
+    Class with methods allowing set_meta() to be called externally to the
+    Galaxy head.
+    This class allows access to external metadata filenames for all outputs
+    associated with a job.
+    We will use JSON as the medium of exchange of information, except for the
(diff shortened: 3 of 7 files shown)


test/unit/tools/test_metadata.py
```diff
@@ -0,0 +1,171 @@
...
+        }
+        command = self.metadata_command(output_datasets)
+        self._write_galaxy_json("""{"type": "dataset", "dataset_id": "%s", "name": "my dynamic name", "ext": "fasta", "info": "my dynamic info"}""" % output_dataset.dataset.id)
+        self._write_output_dataset_contents(output_dataset, ">seq1\nGCTGCATG\n")
+        self.exec_metadata_command(command)
+        metadata_set_successfully = self.metadata_compute_strategy.external_metadata_set_successfully(output_dataset, sa_session)
+        assert metadata_set_successfully
+        output_dataset.extension = "fasta"  # gets done in job finish...
+        self.metadata_compute_strategy.load_metadata(output_dataset, "out_file1", sa_session, working_directory=self.job_working_directory)
+        print(output_dataset.metadata)
```
bgruening: Should this be removed?

test/unit/tools/test_metadata.py
```diff
@@ -0,0 +1,171 @@
...
+                                    cwd=self.job_working_directory,
+                                    env=_environ,
+                                    stdout=stdout_file,
+                                    stderr=stderr_file)
+            ret = proc.wait()
+        self.print_command_output()
+        assert ret == 0
+        return ret
+
+    def print_command_output(self):
```
bgruening: If this should stay, maybe some more debug lines are useful to know what is actually printed.

test/unit/tools/test_metadata.py
```diff
@@ -0,0 +1,171 @@
...
+        output_dataset = self._create_output_dataset(
+            extension="auto",
+        )
+        sa_session = self.app.model.session
+        sa_session.flush()
+        output_datasets = {
+            "out_file1": output_dataset,
+        }
+        command = self.metadata_command(output_datasets)
+        self._write_galaxy_json("""{"type": "dataset", "dataset_id": "%s", "name": "my dynamic name", "ext": "fasta", "info": "my dynamic info"}""" % output_dataset.dataset.id)
```
bgruening: should that be `auto`?

test/unit/tools/test_metadata.py
```diff
@@ -0,0 +1,171 @@
...
+        output_dataset = self._create_output_dataset(
+            extension="fasta",
+        )
+        sa_session = self.app.model.session
+        sa_session.flush()
+        output_datasets = {
+            "out_file1": output_dataset,
+        }
+        command = self.metadata_command(output_datasets)
+        # self._write_galaxy_json("""{"type": "dataset", "dataset_id": "%s", "name": "my dynamic name", "ext": "fasta", "info": "my dynamic info"}""" % output_dataset.id)
```
bgruening: remove?

test/unit/tools/test_metadata.py
jmchilton: yes...

test/unit/tools/test_metadata.py
jmchilton: Really good point, I'll rebase this with some context!

test/unit/tools/test_metadata.py
jmchilton: No - setting it to auto during setup results in the extension being set in this file being respected. So it should be auto above and actual extension here. I'll leave a comment to this point - because I agree it reads funny.

test/unit/tools/test_metadata.py
jmchilton: definitely, thanks!

lib/galaxy/metadata/__init__.py
```diff
@@ -0,0 +1,277 @@
...
+                except Exception as e:
+                    log.debug('Failed to cleanup external metadata file (%s) for %s: %s' % (key, dataset_key, e))
+
+    def set_job_runner_external_pid(self, pid, sa_session):
+        for metadata_files in sa_session.query(galaxy.model.Job).get(self.job_id).external_output_metadata:
+            metadata_files.job_runner_external_pid = pid
+            sa_session.add(metadata_files)
+            sa_session.flush()
+
+    def load_metadata(self, dataset, dataset_name, sa_session, working_directory, remote_metadata_directory=None):
```
mvdbeek: The signature in the metaclass is `self, dataset, name, sa_session, working_directory, remote_metadata_directory=None`, probably good to keep that consistent.

lib/galaxy/metadata/__init__.py
```diff
@@ -0,0 +1,277 @@
...
+        # load metadata from file
+        # we need to no longer allow metadata to be edited while the job is still running,
```
mvdbeek, line 260: off-topic, but we should fix this (is that even enforced anywhere?). Since we record HDA versions now this shouldn't be an issue anymore.

lib/galaxy/jobs/__init__.py
```diff
@@ -1758,9 +1743,14 @@ def setup_external_metadata(self, exec_dir=None, tmp_dir=None,
...
             self.app.datatypes_registry.to_xml_file(path=datatypes_config)
-        command = self.external_output_metadata.setup_external_metadata([output_dataset_assoc.dataset for
-                                                                         output_dataset_assoc in
-                                                                         job.output_datasets + job.output_library_datasets],
+
+        output_datasets = {}
+        for output_dataset_assoc in job.output_datasets + job.output_library_datasets:
+            output_name = output_dataset_assoc.name
+            assert output_name not in output_datasets
+            output_datasets[output_dataset_assoc.name] = output_dataset_assoc.dataset
```
mvdbeek, line 1760: Is `.name` guaranteed to be unique ? If not I think this would break things. My gut feeling is it's not unique. Would it hurt to `enumerate` here for the dictionary keys ?

lib/galaxy/metadata/__init__.py
```diff
@@ -0,0 +1,277 @@
...
+    def setup_external_metadata(self, datasets_dict, sa_session, exec_dir=None,
+                                tmp_dir=None, dataset_files_path=None,
+                                output_fnames=None, config_root=None,
+                                config_file=None, datatypes_config=None,
+                                job_metadata=None, compute_tmp_dir=None,
+                                include_command=True, max_metadata_value_size=0,
+                                kwds=None):
+        kwds = kwds or {}
+        if tmp_dir is None:
+            tmp_dir = MetadataTempFile.tmp_dir
```
mvdbeek: This (and below) looks so weird and unnecessary to me. I know it's old code, but Is it worth putting at least a `# FIXME` in there ?

lib/galaxy_ext/metadata/set_metadata.py
```diff
@@ -103,7 +103,7 @@ def set_metadata():
             try:
                 line = stringify_dictionary_keys(json.loads(line))
                 if line['type'] == 'dataset':
-                    existing_job_metadata_dict[line['dataset_id']] = line
+                    existing_job_metadata_dict[int(line['dataset_id'])] = line
```
mvdbeek, line 106: Why is the `dataset_id` a string here, that seems weird ...  but good catch. Should we also log the exception on line 109 ?

mvdbeek (commented):
~~Nice cleanup, I'm just a little worried about keying on `dataset.name`~~
Since this isn't an HDA, but a JobToOutputDatasetAssociation it's fine.

lib/galaxy/jobs/__init__.py
jmchilton: So this is the name of the tool output (like the name in the ``data`` tag - ``<data format="txt"  name="output">`` ), not the name of the dataset. I'm really hoping it is unique because I'm keying on it over and over downstream. Could a tool really have two outputs with the same output name? I've already started keying on it for like ToolProvidedMetadata (galaxy.json) in recent releases of Galaxy - because it is a lot more easy to read and write consistently than the dataset id - which requires transferring knowledge about the runtime of the output context to the tool.

lib/galaxy_ext/metadata/set_metadata.py
jmchilton: We can't log exceptions in set_metadata.py yet right - might break the job or whatever? Maybe after #7095?

mvdbeek: Yeah, https://github.com/galaxyproject/galaxy/pull/7095 should do it.

lib/galaxy/jobs/__init__.py
mvdbeek: oh, missed that this isn't a HDA. Yes, that's a totally reasonable assumption.

jmchilton: Extremely easy to miss given the misleading name.

lib/galaxy/metadata/__init__.py
jmchilton: That helps... but imagine the tool gives an output a name, the job runs, the user changes it, the tool runtime provided metadata gives a different name - despite having a record of HDA versions we don't use it to enforce the name provided by the tool has been changed by the user right? The tool will still overwrite the user changes because we're not checking this information.

lib/galaxy/metadata/__init__.py
jmchilton: Yeah... this seems bad... if it is used at all it is almost worse - it would seem conflicting jobs might overwrite each other here...

lib/galaxy/metadata/__init__.py
mvdbeek: Right, Im not sure why I thought this wouldn't be an issue anymore.

lib/galaxy/metadata/__init__.py
jmchilton: Going to let the tests run on https://github.com/galaxyproject/galaxy/pull/7213 and see what happens, I'm not sure we have any tests that would indicate if there is a problem here though.

lib/galaxy/metadata/__init__.py
jmchilton: Done in a rebase, thanks!

mvdbeek (approved):

lib/galaxy/datatypes/metadata.py
```diff
@@ -42,5 +41,4 @@
     "PythonObjectParameter",
     "FileParameter",
     "MetadataTempFile",
-    "JobExternalOutputMetadataWrapper",
```
nsoranzo, line 45: Can this removal break some ToolShed tool?

jmchilton: That would perverse... this shouldn't be used by datatypes the way these other things might be. This class never really belonged with the other ones in my opinion and I don't understand what the point of importing it from TS code would have been.

lib/galaxy/metadata/__init__.py
```diff
@@ -0,0 +1,275 @@
...
+from six.moves import cPickle
+
+import galaxy.model
+from galaxy.model.metadata import FileParameter, MetadataTempFile
+from galaxy.util import in_directory
+
+log = getLogger(__name__)
+
+
+def get_metadata_compute_strategy(app, job_id):
```
nsoranzo, line 21: For anyone noticing that `app` is unused: it is needed for follow-up PR https://github.com/galaxyproject/galaxy/pull/7058/commits/d8d43e9e238a0b28f5a4582fcf0fea12e4151dc8#diff-07df768c6572ad942b290901fe1424b3

PR #7160, opened 2018-12-30
Unprivileged remote tool tests

mvdbeek:
This would allow users to run tool tests against a Galaxy server.
Previously admins could ask for the path to test datasets on the Galaxy server via the API,
but access was only possible if the test files were available from where the test was run.

Instead of revealing the test data path we can also allow downloading the listed files directly.
I've changed the test interactor to now download the data from Galaxy instead of reading them from the file system. (One could also think about `paste_path` or `paste_url`, but for `paste_path` that option would need to be activated, and `paste_url` by default doesn't allow requests to localhost/127.0.01).

### DIFF
--- a/lib/galaxy/tools/verify/interactor.py
+++ b/lib/galaxy/tools/verify/interactor.py
@@ -4,6 +4,8 @@ from __future__ import print_function
 import os
 import re
 import sys
+import tarfile
+import tempfile
 import time
 from collections import OrderedDict
 from json import dumps
@@ -18,7 +20,11 @@ try:
     import requests
 except ImportError:
     requests = None
-from six import StringIO, text_type
+from six import (
+    BytesIO,
+    StringIO,
+    text_type,
+)
 
 from galaxy import util
 from galaxy.tools.parser.interface import TestCollectionDef, TestCollectionOutputDef
@@ -155,13 +161,13 @@ class GalaxyInteractorApi(object):
 
     def verify_output_dataset(self, history_id, hda_id, outfile, attributes, tool_id):
         fetcher = self.__dataset_fetcher(history_id)
-        test_data_path_builder = self.__test_data_path_builder(tool_id)
+        test_data_downloader = self.__test_data_downloader(tool_id)
         verify_hid(
             outfile,
             hda_id=hda_id,
             attributes=attributes,
             dataset_fetcher=fetcher,
-            test_data_path_builder=test_data_path_builder,
+            test_data_downloader=test_data_downloader,
             keep_outputs_dir=self.keep_outputs_dir
         )
         self._verify_metadata(history_id, hda_id, attributes)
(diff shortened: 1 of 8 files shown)


lib/galaxy/tools/verify/interactor.py
```diff
@@ -251,10 +257,16 @@ def new_history(self):
...
     @nottest
-    def test_data_path(self, tool_id, filename):
-        response = self._get("tools/%s/test_data_path?filename=%s" % (tool_id, filename), admin=True)
+    def test_data_download(self, tool_id, filename, mode='file'):
+        response = self._get("tools/%s/test_data_download?filename=%s" % (tool_id, filename), admin=True)
         assert response.status_code == 200
-        return response.json()
+        if mode == 'file':
+            return response.content
+        elif mode == 'directory':
```
mvdbeek, line 266: I'm not sure what the syntax for this in tool tests is, if this isn't documented we could also drop this. That would simplify this a lot.

mvdbeek: This is in the xsd here https://github.com/galaxyproject/Galaxy-XSD/blob/master/galaxy.xsd#L1277 and an example is here: https://galaxyproject.org/admin/tools/writing-tests/#directory-comparison

We don't seem to have any test tool that exercises this, but I guess I could add one.

lib/galaxy/webapps/galaxy/api/tools.py
```diff
@@ -129,6 +130,41 @@ def test_data_path(self, trans, id, **kwd):
         else:
             raise exceptions.ObjectNotFound("Specified test data path not found.")
 
+    @expose_api_raw_anonymous_and_sessionless
+    def test_data_download(self, trans, id, **kwd):
+        """
+        GET /api/tools/{tool_id}/test_data_path?tool_version={tool_version}&={filename}
```
martenson: the URL here seems incorrect, should it be `test_data_download`  and `&filename={filename}`?

mvdbeek: Yep, thanks!

mvdbeek:
~~The new test in https://github.com/galaxyproject/galaxy/pull/7160/commits/5490a6db7c5fa5db431c64b51998f923110106a6 depends on https://github.com/galaxyproject/galaxy-test-data/pull/2~~

jmchilton:
I was imaging teaching the upload tool custom ``tool_data://tool_id/?filename`` Galaxy URIs for efficiency but we can totally implement that later if needed given that the API endpoint added here is totally reasonable and great to have regardless. My approach could easily be considered over-engineering. My reading is that this would not expose data on the Galaxy server if filename is not contained in the tool data directory (e.g. ``filename?=../../../etc/allpasswords`` or ``filename?=/etc/allpasswords``) given the ``in_directory`` test in ``__walk_test_data``. Do you agree? A test would make me feel more comfortable but happy to merge as is - this is wonderful.

jmchilton (approved):

mvdbeek:
> Do you agree? A test would make me feel more comfortable but happy to merge as is - this is wonderful.

That's my understanding, but I can add a test for the new route. I think `test_test_data_filepath_security` in test/api/test_tools.py already tests the relevant code, but doesn't hurt to test this explicitly for the new route.

jmchilton:
Well this is utterly wonderful, thanks so much @mvdbeek

lib/galaxy/tools/verify/test_data.py
```diff
@@ -43,10 +43,16 @@ def get_filename(self, name):
             if not resolver.exists(name):
                 continue
             filename = resolver.path(name)
-
             if filename:
                 return os.path.abspath(filename)
 
+    def get_filecontent(self, name):
+        filename = self.get_filename(name=name)
+        return open(filename, mode='rb')
```
nsoranzo, line 51: @mvdbeek Shouldnt' this be:
```python
        with open(filename, mode='rb') as f:
            return f.read()
```

mvdbeek: Yep, you're right.

martenson:
@luke-c-sargent we should adjust our WF testing pipeline to use this

PR #7167, opened 2019-01-03
Remote tool testing enhancements

mvdbeek:
I'll collect some things as I encounter them.

- [x] Allow passing in a test history. Otherwise you'll end up with one history per test, meaning > 1000 for a decent sized server if you're testing all tools
- [x] Fall back to local test data if Galaxy version isn't compatible with remote test data fetching/uploading
~~- [ ] Hash and check if we already used the test data, avoids re-uploading same file over and over (not sure if in interactor or in ephemeris ...)~~

### DIFF
--- a/lib/galaxy/tools/verify/interactor.py
+++ b/lib/galaxy/tools/verify/interactor.py
@@ -12,6 +12,7 @@ from collections import OrderedDict
 from json import dumps
 from logging import getLogger
 
+from packaging.version import parse as parse_version, Version
 try:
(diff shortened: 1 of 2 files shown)


mvdbeek:
This is ready for review, and should be backported to 19.01, otherwise targeting external Galaxy < 19.01 with the galaxy interactor will fail.

martenson (approved):

jmchilton:
Switched tag to 19.01 since it was backported.

PR #7071, opened 2018-12-04
[18.09] Fix process key

mvdbeek:
This should fix #7048

PR #7168, opened 2019-01-04
Misc fixes

nsoranzo:
Cleaning up my stashed changes...

### DIFF
--- a/lib/galaxy/tools/errors.py
+++ b/lib/galaxy/tools/errors.py
@@ -75,7 +75,7 @@ error_report_template_html = """
 
 <h3>User Provided Information</h3>
 
-The user <a href="mailto:${email_str}"><span style="font-family: monospace;">${email_str}</span></a> provided the following information:
+The user <span style="font-family: monospace;">${email_str}</span> provided the following information:
 
 <pre style="white-space: pre-wrap;background: #eeeeee;border:1px solid black;padding:1em;">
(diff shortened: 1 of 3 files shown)


mvdbeek (approved):

nsoranzo:
@mvdbeek I think the failing py34-unit test (which I can reproduce locally) may be due to https://github.com/galaxyproject/galaxy/pull/7071/files#r239664690

mvdbeek:
Yes, but I don't think it matters, the repeat elements may be in any order for `expand_meta_parameters`. But I agree this isn't made clear in any way, I can fix it.

nsoranzo:
In fact I was not sure what should be the content of those dictionaries, should I just strip the commit?

mvdbeek:
Sure, I can add a separate PR with the commit and and the fix.

nsoranzo:
@mvdbeek I've stripped the commit.

It seems that https://github.com/galaxyproject/galaxy-test-data/pull/2 broke `test/api/test_tools.py::ToolsTestCase::test_test_data_download_composite` though.

mvdbeek:
yeah, it's 5 items now, can you fix the test?

mvdbeek:
I've fixed the API test in #7171

mvdbeek:
Thanks @nsoranzo !

PR #7170, opened 2019-01-04
bump pyyaml due to CVE-2017-18342

martenson:
see https://nvd.nist.gov/vuln/detail/CVE-2017-18342


martenson:
After a short review I learned that we've already moved all relevant usage of `yaml.load` to `yaml.safe_load` and given there is no newer _stable_ release of pyyaml than what we use this PR is unwanted I believe. Closing.

PR #7174, opened 2019-01-07
Document ``message_box_class`` option

nsoranzo:
Reported by @vazovn.


hexylena (approved):

dannon:
Should certainly be documented, but we should probably just deprecate this and use info as the default, allowing style overrides via scss like anything else.

PR #7179, opened 2019-01-07
WIP: library_upload_dir: improvements

bernt-matthias:
**major improvement**: when using upload_from_galaxy_filesystem the PATH needed to be given absolute and the script generates the complete hierarchy in Galaxy library (i.e. starting from root). Now the script has an additional parameter `--root_folder ROOT` which allows to upload ROOT/PATH into the library folder PATH.

**minor improvements**:
- added `--library_name` which allows to specify the library by name
- fixed `--folder` which now works as documented, ie the root folder of the library is used if not given
- added preserve dir and tag using files names
- also started a bit with documentation

Questions: 
- maybe there is also a better way to add filesystem folders to a library folder
- maybe a mechanism like the one I implemented with --root should be in the API .. then this would also be possible with `parsec`
- should this script stay here or is it better to integrate into (e.g.) ephemeris ... ?

### DIFF
--- a/scripts/api/library_upload_dir.py
+++ b/scripts/api/library_upload_dir.py
@@ -34,10 +34,17 @@ class Uploader(object):
         self.gi = galaxy.GalaxyInstance(url=url, key=api)
         libs = self.gi.libraries.get_libraries(library_id=library_id,
                                         name=library_name)
+
+        # TODO libs also contains deleted libraries even if it should not
+        # https://github.com/galaxyproject/bioblend/issues/239 fixed by
+        # https://github.com/galaxyproject/bioblend/pull/273
+        # remove if bioblend>0.12 is released
+        libs = [d for d in libs if not d['deleted']]
+
         if len(libs) == 0:
             raise Exception("Unknown library [%s,%s]" % (library_id, library_name))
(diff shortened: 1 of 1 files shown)


jmchilton:
> should this script stay here or is it better to integrate into (e.g.) ephemeris ... ?

This script since it is here for demonstration purposes - I'm not sure it is designed to actually be used. I' rather see enhancements to https://github.com/galaxyproject/ephemeris/blob/master/src/ephemeris/setup_data_libraries.py or a second data library script there. But I'm not going to abandon this so I'll merge the PR.

jmchilton:
Actually, there are linting problems - will you fix those?

> ./scripts/api/library_upload_dir.py:25:56: W291 trailing whitespace

bernt-matthias:
I'm fine with adding the script to ephemeris. I thought that it would fit better there. I will test this a bit more and then PR to ephemeris.

martenson:
@bernt-matthias how do you see the future of this PR?

mvdbeek:
I've dropped the milestone here, but I also agree with John that this should probably go as an enhancement to ephemeris (if that didn't already happen?).

PR #6321, opened 2018-06-12
Remove reset handler from ruler input element

guerler:
This PR attempts to revise the reset handling mechanism used in the ruler input element. The PR is not complete yet. ping @jmchilton

PR #7183, opened 2019-01-07
Add error listener for input elements

### LINKED ISSUE #7172: Tool IU - parameter mins no longer expressed to users and are silently enforced
Integer and floating parameter types with a min specified is not properly relayed to user. Values entered by a user are 'silently' changed (replaced without notice), see attached gif. 

Previously this would cause a validation error to be displayed to the user, and tool would not execute until user corrected.

screen recording 2019-01-04 at 4 55 55 pm

May also be helpful to confirm that all validation failures are still being properly displayed to the user and not automatically corrected and executed.

guerler:
This PR adds an error listener for input elements, allowing them to individually trigger custom errors or warnings. Warning messages have been added to the slider element to highlight out-of-range inputs. This fixes #7172. This feature can be re-used in  #6321.


mvdbeek:
Is this ready for review @guerler  ?

mvdbeek (approved):
Works well, thanks @guerler!

PR #7185, opened 2019-01-07
Add python to Count1 tool requirements

mvdbeek:
... the version is kind of arbitrary, but I guess we'll want to move away from 2.

Without python the tool will fail if run containerised.
There are more tools that use just python or perl, do we also want to add requirements there ?

### DIFF
--- a/tools/filters/uniq.xml
+++ b/tools/filters/uniq.xml
@@ -1,7 +1,8 @@
-<tool id="Count1" name="Count" version="1.0.2">
+<tool id="Count1" name="Count" version="1.0.3">
   <description>occurrences of each record</description>
     <requirements>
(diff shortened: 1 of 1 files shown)


tools/filters/uniq.xml
```diff
@@ -1,7 +1,8 @@
-<tool id="Count1" name="Count" version="1.0.2">
+<tool id="Count1" name="Count" version="1.0.3">
   <description>occurrences of each record</description>
     <requirements>
         <requirement type="package" version="8.30">coreutils</requirement>
+        <requirement type="package" version="3.6">python</requirement>
```
nsoranzo: There's already 3.7 on conda-forge if we want to go for the latest version.

PR #7188, opened 2019-01-08
[19.01] Fix workflow parameter test case?

jmchilton:


nsoranzo (approved):

test/selenium_tests/test_workflow_run.py
```diff
@@ -211,8 +211,10 @@ def _set_num_lines_to_3(self, tour_id):
         # for random_lines num_lines parameter as runtime parameter in workflow form.
         div = self.tool_parameter_div(tour_id)
         input_element = div.find_element_by_css_selector("input")
-        initial_value = input_element.get_attribute("value")
-        assert initial_value == "1", initial_value
```
guerler, line 215: Alternatively we could just test for the initial_value to be "" instead of "1".

nsoranzo: @guerler That also seemed a good idea, are you going to open a PR for that?

jmchilton: This would break ``test_subworkflow_runtime_parameters`` presumably, though is that an indication the other PR should somehow be applied to subworkflows also?

guerler: Right, those two test cases would have to be distinguished. The answer to the question depends on how subworkflows currently process runtime values. This is a good point. It might cause an inconsistency.

Issue #7108, opened 2018-12-13
Shared Libraries: Unable to change permissions


scholtalbers:
On 18.09 (and maybe before that) I cannot change permissions on datasets anymore. See below for the exception that is raised from a url like https://galaxy.embl.de/api/libraries/datasets/64d168807f4b3765/permissions?scope=current

https://github.com/galaxyproject/galaxy/blob/49693814a9702a855d7df166d27c7a6f4f1faba1/lib/galaxy/webapps/galaxy/api/library_datasets.py#L158
```
Traceback (most recent call last):  
File "lib/galaxy/web/framework/decorators.py", line 283, in decorator    rval = func(self, trans, *args, **kwargs)  File "lib/galaxy/webapps/galaxy/api/library_datasets.py", line 125, in show_roles    return self._get_current_roles(trans, library_dataset) 
File "lib/galaxy/webapps/galaxy/api/library_datasets.py", line 158, in _get_current_roles    return self.serialize_dataset_association_roles(library_dataset)AttributeError: 'LibraryDatasetsController' object has no attribute 'serialize_dataset_association_roles'
```

PR #7190, opened 2019-01-08
[18.09] fix library dataset permission API

martenson:
@jmchilton Could you please have a look whether this is how it was meant to be used?

broken in refactoring at https://github.com/galaxyproject/galaxy/commit/49693814a9702a855d7df166d27c7a6f4f1faba1#diff-55311aee30c3c44eb2dc3159363be503

closes https://github.com/galaxyproject/galaxy/issues/7108

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/library_datasets.py
+++ b/lib/galaxy/webapps/galaxy/api/library_datasets.py
@@ -22,7 +22,8 @@ from galaxy.managers import (
     folders,
     lddas,
     library_datasets,
-    roles
+    roles,
+    hdas
 )
 from galaxy.tools.actions import upload_common
 from galaxy.tools.parameters import populate_state
@@ -45,6 +46,7 @@ class LibraryDatasetsController(BaseAPIController, UsesVisualizationMixin, Libra
         self.role_manager = roles.RoleManager(app)
         self.ld_manager = library_datasets.LibraryDatasetsManager(app)
         self.ldda_manager = lddas.LDDAManager(app)
+        self.hda_manager = hdas.HDAManager(app)
 
     @expose_api_anonymous
     def show(self, trans, id, **kwd):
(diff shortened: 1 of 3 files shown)


lib/galaxy/webapps/galaxy/api/library_datasets.py
```diff
@@ -22,7 +22,8 @@
     folders,
     lddas,
     library_datasets,
-    roles
+    roles,
+    hdas
```
nsoranzo: This import should go after `folders,` above.

lib/galaxy/webapps/galaxy/api/library_datasets.py
```diff
@@ -155,7 +157,7 @@ def _get_current_roles(self, trans, library_dataset):
         :rtype:     dictionary
         :returns:   dict of current roles for all available permission types
         """
-        return self.serialize_dataset_association_roles(library_dataset)
+        return self.hda_manager.serialize_dataset_association_roles(trans, library_dataset)
```
nsoranzo: I think this should be `ldda_manager` (added in the same PR, see https://github.com/galaxyproject/galaxy/commit/49693814a9702a855d7df166d27c7a6f4f1faba1#diff-55311aee30c3c44eb2dc3159363be503R47 ) instead of `hda_manager`.

jmchilton: Exactly what I think I was thinking - it should inherit the same method as the HDA one (https://github.com/galaxyproject/galaxy/commit/49693814a9702a855d7df166d27c7a6f4f1faba1#diff-1c4ba52684d803d42ae20734665ea77cR317). Let me know if that doesn't work?

lib/galaxy/webapps/galaxy/api/library_datasets.py
martenson: img_20190108_135406896

lib/galaxy/webapps/galaxy/api/library_datasets.py
martenson: So something like now?

nsoranzo:
@martenson You may want to also solve https://github.com/galaxyproject/galaxy/commit/49693814a9702a855d7df166d27c7a6f4f1faba1#r31873472 , which is related.

martenson:
@nsoranzo sure, I've added that

nsoranzo (approved):

jmchilton:
Thanks for cleaning up my mess @martenson

PR #7193, opened 2019-01-09
Simplify history import to fix extra files and ObjectStore interactions.

jmchilton:
This usage counting that I am eliminating here made sense in (a94bf20abfd15f21c3cc380ee92896d0f107bb23) but it was broken with the transition to ObjectStore compatibility (ea89dd8f149778b6c2f0f3f4a34c8b21f7033af7) and it is broken with extra files handling (9ea0fee48aa57361147e9145be8deaaafb2a0960).

### DIFF
--- a/lib/galaxy/tools/imp_exp/__init__.py
+++ b/lib/galaxy/tools/imp_exp/__init__.py
@@ -101,16 +101,6 @@ class JobImportHistoryArchiveWrapper(UsesAnnotations):
                     provenance_attrs = load(open(provenance_file_name))
                     datasets_attrs += provenance_attrs
 
-                # Get counts of how often each dataset file is used; a file can
-                # be linked to multiple dataset objects (HDAs).
-                datasets_usage_counts = {}
-                for dataset_attrs in datasets_attrs:
-                    temp_dataset_file_name = \
-                        os.path.realpath(os.path.join(archive_dir, dataset_attrs['file_name']))
-                    if (temp_dataset_file_name not in datasets_usage_counts):
-                        datasets_usage_counts[temp_dataset_file_name] = 0
-                    datasets_usage_counts[temp_dataset_file_name] += 1
-
                 # Create datasets.
                 for dataset_attrs in datasets_attrs:
                     metadata = dataset_attrs['metadata']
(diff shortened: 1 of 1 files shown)


lib/galaxy/tools/imp_exp/__init__.py
```diff
@@ -150,26 +140,22 @@ def get_tag_str(tag, value):
...
-                                if file_list:
-                                    for extra_file in file_list:
-                                        self.app.object_store.update_from_file(
-                                            hda.dataset, extra_dir='dataset_%s_files' % hda.dataset.id,
-                                            alt_name=extra_file, file_name=os.path.join(archive_dir, dataset_extra_files_path, extra_file),
-                                            create=True)
-                        else:
-                            datasets_usage_counts[temp_dataset_file_name] -= 1
-                            shutil.copyfile(temp_dataset_file_name, hda.file_name)
+                        self.app.object_store.update_from_file(hda.dataset, file_name=temp_dataset_file_name, create=True)
```
mvdbeek, line 143: This was broken because it creates HDAs with associated `Dataset` instances without actually creating the file for the `Dataset` instance in the object store, right ? So would it make sense to do something like:

```
imported_datasets = {}
...
if not temp_dataset_file_name in imported_datasets:
   <all the necessary stuff>
    imported_datasets[temp_dataset_file_name] = hda.dataset
else:
    hda.dataset = imported_datasets[temp_dataset_file_name]
```
?

That would avoid importing HDA copies multiple times, I think.

jmchilton: But with if ``extra_files_path`` doesn't match - I feel like this never happens and we should just simplify the code for now with this fix and make this optimization down the road with tests to verify everything works.

mvdbeek: Ah, right. Fair enough!

mvdbeek (approved):

lib/galaxy/tools/imp_exp/__init__.py
jmchilton: Thanks @mvdbeek, I'm building out tests and abstractions downstream that should make me confident enough to do this - I will try to incorporate this request.

PR #7195, opened 2019-01-09
Custos integration

machristie:
*Note: initially this PR aimed to add generic OIDC integration for any Keycloak instance. It has since been changed to add integration specific to Custos so as to simplify configuration and to keep ownership of the integration details with the Custos project.*

This pull request adds the ability to do OIDC authentication with Custos. The integration builds on the earlier python-social-auth integration support in Galaxy.


machristie:
**Note: these instructions no longer work. See below for how to test locally with Custos.**

Some notes on how to test locally with Keycloak:
1. Run Keycloak in Docker.
    ````
    docker run -d -p 9000:8080 -e KEYCLOAK_USER=admin -e KEYCLOAK_PASSWORD=admin jboss/keycloak
    ```
2. Log into the Keycloak console at http://localhost:9000/auth with the username and password specified above.
3. Create a new client in the master realm: http://localhost:9000/auth/admin/master/console/#/realms/master/clients
    - specify a Client ID
    - click Save
    - set Access Type to confidential
    - set the Valid Redirect URI. The value will depend on what you named the "provider" in oidc_backends_config.xml. In this example, the provider is called "galaxy-auth": `http://localhost:8000/authnz/galaxy-auth/oidc/callback*`.
    - click Save
    - go to the Credentials tab and copy the client secret into oidc_backends_config.xml
4. Create a user to test with at http://localhost:9000/auth/admin/master/console/#/realms/master/users. You can set a non-temporary password after creating the user.
5. Set the well_known_oidc_config_uri in oidc_backends_config.xml. With this setup it would be: `http://localhost:9000/auth/realms/master/.well-known/openid-configuration`. The configuration will look something like this:
  ```xml
   <provider name="Galaxy-auth" implementation="OIDC">
        <client_id>galaxy-web</client_id>
        <client_secret>d2fac42f-8de0-4750-9e93-11ab7f5a9aed</client_secret>
        <redirect_uri>http://localhost:8000/authnz/galaxy-auth/oidc/callback</redirect_uri>
        <well_known_oidc_config_uri>http://localhost:9000/auth/realms/master/.well-known/openid-configuration</well_known_oidc_config_uri>
    </provider>
   ```

machristie:
@VJalili this is the Keycloak integration pull request. Let me know if you have any questions.

I think the only question I have now is whether `disconnect()` needs to be implemented and if so, what should it do.

lib/galaxy/authnz/managers.py
```diff
@@ -106,26 +111,48 @@ def _parse_google_config(self, config_xml):
...
+        rtv = {
+            'client_id': config_xml.find('client_id').text,
+            'client_secret': config_xml.find('client_secret').text,
+            'redirect_uri': config_xml.find('redirect_uri').text,
+            'well_known_oidc_config_uri': config_xml.find('well_known_oidc_config_uri').text}
+        if config_xml.find('idp_hint') is not None:
+            rtv['idp_hint'] = config_xml.find('idp_hint').text
+        return rtv
+
+    def _unify_provider_implementation_names(self, provider, implementation):
```
VJalili: Not all code paths return a value. For instance, if `v != provider` for all of enumerated items, the func returns `None`, which is an unexpected value for the caller of this function.

config/oidc_backends_config.xml.sample
```diff
@@ -22,4 +22,20 @@
          login to Galaxy using their Google account.
         -->
     </provider>
+    <provider name="Keycloak-auth" implementation="Keycloak">
```
VJalili: I guess it would be better to replace `keycloak-auth` with a provider name (e.g., `Google`) and set the `idp_hint` automatically accordingly.

config/oidc_backends_config.xml.sample
```diff
@@ -22,4 +22,20 @@
...
+        <client_id> ... </client_id>
+        <client_secret> ... </client_secret>
+        <redirect_uri>http://localhost:8000/authnz/keycloak-auth/keycloak/callback</redirect_uri>
+        <!-- The .well-known configuration URL is typically of the form:
+        https://<KEYCLOAK_SERVER_NAME>/auth/realms/<REALM_NAME>/.well-known/openid-configuration
+        For example, if the Keycloak server is on host
+        mykeycloak.org and the realm name is testrealm, then the URL is:
+        https://mykeycloak.org/auth/realms/testrealm/.well-known/openid-configuration
+        -->
+        <well_known_oidc_config_uri>https://.../auth/realms/.../.well-known/openid-configuration</well_known_oidc_config_uri>
```
VJalili: Maybe it would be more intuitive to replace this attribute with something like `Keycloack_server` and `realm`, and generate `well_known_oidc_config_uri` in the backend.

lib/galaxy/authnz/keycloak_authnz.py
```diff
@@ -0,0 +1,147 @@
...
+    def _validate_nonce(self, trans, nonce_hash):
+        nonce_cookie = trans.get_cookie(name=NONCE_COOKIE_NAME)
+        # Delete the nonce cookie
+        trans.set_cookie('', name=NONCE_COOKIE_NAME, age=-1)
+        nonce_cookie_hash = self._hash_nonce(nonce_cookie)
+        if nonce_hash != nonce_cookie_hash:
+            raise Exception("Nonce mismatch!")
+
+    def _load_well_known_oidc_config(self, well_known_uri):
+        return requests.get(well_known_uri).json()
```
VJalili: This function raises the following unhandled exception.
`SSLError: HTTPSConnectionPool(host='localhost', port=9000): Max retries exceeded with url: /auth/realms/master/.well-known/openid-configuration (Caused by SSLError(SSLError("bad handshake: Error([('SSL routines', 'ssl3_get_record', 'wrong version number')],)",),))`

(I've set the `well_known_oidc_config_uri` to `https://localhost:9000/auth/realms/master/.well-known/openid-configuration`.)

lib/galaxy/authnz/managers.py
```diff
@@ -106,26 +111,48 @@ def _parse_google_config(self, config_xml):
...
             except Exception as e:
-                log.exception('An error occurred when loading PSAAuthnz')
+                log.exception('An error occurred when loading {}'.format(authnz_backend_class.__name__))
                 return False, str(e), None
         else:
             msg = 'The requested identity provider, `{}`, is not a recognized/expected provider'.format(provider)
             log.debug(msg)
             return False, msg, None
 
+    def _get_authnz_backend_class(self, implementation):
```
VJalili: `static` method?

lib/galaxy/authnz/managers.py
```diff
@@ -188,17 +215,19 @@ def try_get_authz_config(sa_session, user_id, authz_id):
             raise exceptions.ItemAccessibilityException(msg)
         return qres
 
-    def authenticate(self, provider, trans):
+    def authenticate(self, provider, implementation, trans):
```
VJalili: If I use `provider = 'psa'` and `implementation = 'google'`, as a result of calling `authnz/psa/google/login` instead of `authnz/google/psa/login`, I get a `redirect_uri: null`. Maybe a more self-explanatory message would be better instead.

VJalili (changes requested):

VJalili:
@machristie Thank you for this integration. I added some inline comments. 

Regarding your notes on how to set this up for local testing: 
- I don't see ` Access Type` option at client creation, I see it in client settings tab after it is created.
- The `Credentials` tab appear after `Access type` is set to `confidential`. 

While testing this locally I am getting `Max retries exceeded with url` error, I added an inline comment. 

> I think the only question I have now is whether disconnect() needs to be implemented and if so, what should it do.

Since we're implementing this for psa, I guess for consistency between backends, it would be better to implement it for keycloak as well. 

Disconnect is the process of disassociating a Galaxy user and a third-party authnz. In other words, it is the process of removing any access and/or ID tokens of a user.

lib/galaxy/authnz/keycloak_authnz.py
machristie: The URL should be "http" not "https".  See if that works.

lib/galaxy/authnz/managers.py
machristie: Good catch. I'll change this to return `None, None` if there are no matches so it always returns a tuple.

lib/galaxy/authnz/keycloak_authnz.py
VJalili: Thats correct; however, it might be better to handle such an exception here, and also update the example well_known url.

config/oidc_backends_config.xml.sample
machristie: Well, the idp_hint has to be the alias of the identity provider, which could be whatever a user provides when configuring the identity provider in Keycloak, so we can't figure out the idp_hint from the name of the provider.

Or are you saying to flesh out the sample entry here as an example of how to add Google via Keycloak? If so, yeah that makes sense.

config/oidc_backends_config.xml.sample
machristie: Yeah that would probably make it more user-friendly. I was trying to make it more forward compatible safe since Keycloak could decide at some point to change how they construct the .well-known url, but the risk of that seems small.  I'll change it to what you suggested.

lib/galaxy/authnz/managers.py
machristie: Yes

lib/galaxy/authnz/managers.py
machristie: Yeah, will do

machristie:
Thanks @VJalili . I've updated the local testing instructions now.

I think it will be simple to implement the disconnect functionality. However, quick question, with the PSA implementation is there a way to connect an existing Galaxy user to a Google login? From what I could tell it looks like a PSA Google login would just create a new Galaxy user.

lib/galaxy/authnz/keycloak_authnz.py
VJalili: Also with http I am getting `null` redirect uri, and checking terminal I see the following error:

```shell
  File "lib/galaxy/authnz/managers.py", line 233, in authenticate
    return True, "Redirecting to the `{}` identity provider for authentication".format(provider), backend.authenticate(trans)
  File "lib/galaxy/authnz/keycloak_authnz.py", line 44, in authenticate
    base_authorize_url, **extra_params)
  File "../.venv/lib/python2.7/site-packages/requests_oauthlib/oauth2_session.py", line 153, in authorization_url
    **kwargs), state
  File "../.venv/lib/python2.7/site-packages/oauthlib/oauth2/rfc6749/clients/web_application.py", line 86, in prepare_request_uri
    redirect_uri=redirect_uri, scope=scope, state=state, **kwargs)
  File "../.venv/lib/python2.7/site-packages/oauthlib/oauth2/rfc6749/parameters.py", line 71, in prepare_grant_uri
    raise InsecureTransportError()
InsecureTransportError: (insecure_transport) OAuth 2 MUST utilize https.
```
While this error makes a total sense, as it is per explicit requirements of OAuth 2.0 (and OIDC inherently) to TLS/SSL secure connection, but it can be tricky for local testing. Maybe this would better bond with the `VERIFY_SSL` setter `oidc_config.xml`. Those configurations should apply to all providers. So, while `VERIFY_SSL` is set to `False`, the backend should not throw exception/warning complaining of insecure connection.

config/oidc_backends_config.xml.sample
VJalili: I guess I was confused with "how keycloak knows that it should redirect the user to Google, for instance"; because I don't see any reference to IdP, so I assumed it is the `idp_hint` specifying this backend (as you also . put it in the commend). I guess it could be intuitive (and consistent with PSA config) if it reads as `<provider name="google" implementation="Keycloak">`.

If you leave it as `<provider name="Keycloak-auth" implementation="Keycloak">` and then you want to implement login with X and Y idps, then we might end-up in configs such as the following: 

```xml
    <provider name="Keycloak-auth" implementation="Keycloak">
        ...
        <idp_hint>X</idp_hint>
    </provider>
    <provider name="Keycloak-auth" implementation="Keycloak">
        ...
        <idp_hint>Y</idp_hint>
    </provider>
```
or: 
```xml
    <provider name="Keycloak-auth" implementation="Keycloak">
        ...
        <idps>
                <idp_hint>X</idp_hint>
                <idp_hint>Y</idp_hint>
        </idp_hint>
    </provider>
```
but I guess the following would be more intuitive: 
```xml
    <provider name="X" implementation="Keycloak">
        ...
    </provider>
    <provider name="Y" implementation="Keycloak">
        ...
    </provider>
```

lib/galaxy/authnz/managers.py
VJalili: Thanks. And probably also handling `None, None` and throwing proper exception.

lib/galaxy/authnz/managers.py
VJalili: Would it make sense to make it static then? i.e., add `@staticmethod` decorator and remove `self`.

VJalili:
You can associate an existing Galaxy user with a third-party identity via PSA, and I guess this attribute should preserve with Keycloak as well. 

(if you're logged in with a Galaxy user account, and then you head to `/authnz/google/login` and complete the process, your Galaxy account will be associated with your Google account.)

One of the functions PSA leverages for this association is the following: 

https://github.com/galaxyproject/galaxy/blob/f954dc45fe94706f2ea04accc9fc92bcbcea892f/lib/galaxy/model/__init__.py#L5024-L5026

VJalili:
Consider the following function: 

https://github.com/galaxyproject/galaxy/blob/f954dc45fe94706f2ea04accc9fc92bcbcea892f/lib/galaxy/authnz/psa_authnz.py#L247-L250

This is currently implemented as part of PSA pipeline. And I guess a similar functionality should implement for Keycloak as well. Maybe it would be better to abstract this function, define a base that defines Galaxy's minimum requirement for both PSA and keycloak (e.g., `id token` must be available), then extend it as needed for keycloak and PSA.

config/oidc_backends_config.xml.sample
machristie: Yes, that last one is what I had in mind, you would create two provider entries that could have different idp_hints.

lib/galaxy/authnz/keycloak_authnz.py
machristie: Did you create the environment variable: `export OAUTHLIB_INSECURE_TRANSPORT=1` ?

lib/galaxy/authnz/managers.py
machristie: Yes, I'll update it.

machristie:
Going to try to keep track of some TODOs:
- [x] rename from Keycloak to OIDC, since the integration is generic and need not be dependent on Keycloak. This will require a couple small refactors to remove some Keycloak assumptions.
- [x] add support for associating an existing Galaxy user with an external identity.
- (to be implemented in separate pull request) when logging in through identity provider try to associate with an existing Galaxy user based on email address.
  - but need to send an email to the user to verify that they do in fact own this email address
  - see proposal below for details
- [x] implement `disconnect()`
- [x] Handle unknown provider/implementation in `_unify_provider_implementation_names`
- [x] better error message if provider/implementation are not found or mixed up in URL
- [x] simplify local testing, currently environment variable needs to be set
- [x] unit tests aren't running in travis

machristie:
Here's my proposal for email verification and associating a Galaxy user with an existing Galaxy user by email:
- I think oidc_authnz should respect the user activation setting in Galaxy. If it isn't enabled then 1) no email verification should take place and 2) no linking to existing Galaxy user accounts based on email address will take place
  - Discussed with @VJalili - if oidc is enabled then user activation will be required to be enabled as well. The reason is that we want to discourage the possibility of a user ending up with two Galaxy user accounts with the same email address which would happen if user activation is disabled and a user creates a Galaxy user account and then also logs in through OIDC with the same email address.  This does mean that to enable OIDC, either in production or locally for development, one will have to enable user activation [and anything else that user activation requires such as an smtp server and admin email](https://galaxyproject.org/admin/user-accounts/).
- if user activation is turned on, then the user will be sent an account activation link just as if they had registered with Galaxy.
- when the user verifies their email address and clicks on the activate URL, we can display a list of existing Galaxy user accounts that are also active and have the same email address. The user has the option of linking to one of those accounts or not. If they choose not to link then a new account is created or activated.
- we could also check if the `email_verified` claim is returned in the userinfo response. If so, oidc_authnz could bypass email verification and just activate the user directly and associate with the first existing Galaxy user account that has the same email address if any exist.

These changes will be made in a separate pull request.

machristie:
@VJalili I updated the email verification proposal above with what we discussed yesterday. Can you please review it and let me know if it still looks good and if I left anything out? Thanks!

machristie:
**Note: this PR no longer has generic OIDC support so this no longer works. To test this PR, see notes below on how to test the Custos integration.**

To test locally with Google OIDC, follow [the instructions here](https://galaxyproject.org/admin/authentication/config/) and use the following in `oidc_backends_config.xml`:
```xml
    <provider name="Google" implementation="OIDC">
        <client_id> ... </client_id>
        <client_secret> ... </client_secret>
        <redirect_uri>http://localhost:8080/authnz/google/callback</redirect_uri>
        <well_known_oidc_config_uri>https://accounts.google.com/.well-known/openid-configuration</well_known_oidc_config_uri>
        <userinfo_claim_mappings>
            <username>email</username>
        </userinfo_claim_mappings>
    </provider>
```

machristie:
@VJalili This PR is ready for a re-review whenever you have a chance. FYI, I'm out of the office until Monday.

bgruening:
@galaxybot test this

config/oidc_backends_config.xml.sample
```diff
@@ -22,4 +22,33 @@
          login to Galaxy using their Google account.
         -->
     </provider>
+    <provider name="Google" implementation="OIDC">
```
VJalili: `OIDC` refers to all the providers, so I guess it would be better to name this `implementation="keyclock"` instead.

VJalili (changes requested):

VJalili:
> * if oidc is enabled then user activation will be required to be enabled as well. The reason is that we want to discourage the possibility of a user ending up with two Galaxy user accounts with the same email address which would happen if user activation is disabled and a user creates a Galaxy user account and then also logs in through OIDC with the same email address.  This does mean that to enable OIDC, either in production or locally for development, one will have to enable user activation [and anything else that user activation requires such as an smtp server and admin email](https://galaxyproject.org/admin/user-accounts/).

Following our conversation, I agree we should require user activation enabled for OIDC to work in order to avoid user impersonation. However, I guess this maybe perceived an over-requirement to setup a development environment.

lib/galaxy/model/mapping.py
```diff
@@ -123,6 +123,22 @@
     Column('lifetime', Integer),
     Column('assoc_type', VARCHAR(64)))
 
+model.OIDCAccessToken.table = Table(
```
VJalili: Since in addition to access token, we have id and refresh token among other values, I guess it would be better to have a more generic class name; e.g., `OIDCToken`.

VJalili: Similarly, I guess table name can be more generic; e.g., `oidc_token`.

VJalili (changes requested):

lib/galaxy/model/mapping.py
machristie: Sounds good 👍

machristie:
How to test the Custos integration locally:

1. To test you'll use the production SciGaP Keycloak installation at https://iam.scigap.org/auth/. For that to work, however, you need to download the CA certificate that iam.scigap.org uses for its SSL certificate so that the SSL connection can be verified.
```
curl -O https://gist.githubusercontent.com/machristie/52a9f2715140df5fe88ea763dbd50b1a/raw/82aebbd7ea45df72e9f5757771e5272f8ea17498/incommon_rsa_server_ca.pem
mv incommon_rsa_server_ca.pem /path/to/galaxy/config/
```
2. Add the following `provider` config to oidc_backends_config.xml:
```xml
    <provider name="Custos">
        <url>https://iam.scigap.org/auth</url>
        <ca_bundle>/path/to/galaxy/config/incommon_rsa_server_ca.pem</ca_bundle>
        <client_id>local-galaxy</client_id>
        <client_secret>3b1e1256-d284-4cb6-b9bb-882f7bf8f06b</client_secret>
        <redirect_uri>http://localhost:8000/authnz/custos/callback</redirect_uri>
        <realm>galaxy-test</realm>
    </provider>
```
3. Start up Galaxy and go to http://localhost:8000/authnz/Custos/login. Copy and paste the redirect uri into the browser.
4. Sign in with CILogon.

machristie:
@VJalili This PR is ready for a re-review. See the previous comment for how to test the Custos integration.

config/oidc_backends_config.xml.sample
```diff
@@ -31,4 +31,16 @@
 
         <prompt>consent</prompt>
     </provider>
+
+    <provider name="Custos">
+        <client_id> ... </client_id>
```
VJalili, line 37: I guess we should also have another required config for the Keycloak endpoint, named `keycloak_endpoint` for instance. Here you're using https://iam.scigap.org (see [this line](https://github.com/galaxyproject/galaxy/pull/7195/files#diff-74566421da44253f075336826614cf9cR194)) as the default endpoint; however, I guess it is better not to assume any defaults for this base endpoint and make it a required configuration.

lib/galaxy/authnz/custos_authnz.py
```diff
@@ -0,0 +1,212 @@
...
+        well_known_oidc_config = self._load_well_known_oidc_config(self.config['well_known_oidc_config_uri'])
+        self.config['authorization_endpoint'] = well_known_oidc_config['authorization_endpoint']
+        self.config['token_endpoint'] = well_known_oidc_config['token_endpoint']
+        self.config['userinfo_endpoint'] = well_known_oidc_config['userinfo_endpoint']
+
+    def _get_well_known_uri_for_provider_and_realm(self, provider, realm):
+        # TODO: Look up this URL from a Python library
+        if provider == 'custos':
+            self.config['ca_bundle'] = CA_CERTFILE
+            return "https://iam.scigap.org/auth/realms/{}/.well-known/openid-configuration".format(realm)
```
VJalili: I guess the endpoint part of the URL (i.e., https://iam.scigap.org) can be defined in the config. (See my comment on the config sample.)

lib/galaxy/authnz/incommon_rsa_server_ca.pem
```diff
@@ -0,0 +1,68 @@
+-----BEGIN CERTIFICATE-----
```
VJalili: I guess this file is committed accidentally.

lib/galaxy/authnz/custos_authnz.py
```diff
@@ -0,0 +1,212 @@
...
+
+from galaxy import util
+from galaxy.model import CustosAuthnzToken, User
+from ..authnz import IdentityProvider
+
+log = logging.getLogger(__name__)
+STATE_COOKIE_NAME = 'custos-state'
+NONCE_COOKIE_NAME = 'custos-nonce'
+BASE_DIR = os.path.dirname(os.path.abspath(__file__))
+CA_CERTFILE = os.path.join(BASE_DIR, "incommon_rsa_server_ca.pem")
```
VJalili: I guess Galaxy's `config` folder could be a better place for storing/retrieving RSA certificates; because with code, the certificates are expected to be located at `galaxy/lib/authnz`, hence making config and code colocated, which is an uncommon setup in Galaxy.

config/oidc_backends_config.xml.sample
```diff
@@ -31,4 +31,16 @@
...
+        <client_secret> ... </client_secret>
+        <redirect_uri>http://localhost:8000/authnz/custos/callback</redirect_uri>
+
+        <realm> ... </realm>
+        <!-- (Optional) Override the default Custos well-known URL to point to a different instance -->
+        <!-- <well_known_oidc_config_uri>https://.../.well-known/openid-configuration</well_known_oidc_config_uri> -->
```
VJalili, line 46: Maybe we can expect admins to provide only the `.well-known` part of the URL for this setter (e.g., `.well-known/openid-configuration</well_known_oidc_config_uri`), and construct the full URL in the `custos_authnz` backend by concatenating this setter with the base URL (i.e., `https://.../`; see `keycloak_endpoint` in my previous comment).

In other words, we can assume the URLs are made as `<KEYCLOAK_ENDPOINT> + <.WELL-KNOWN>`, where `KEYCLOAK_ENDPOINT`, set using `keycloak_endpoint` setter is mandatory, and `WELL-KNOWN` defined by the `well_known_oidc_config_uri` setter is optional with default values.

lib/galaxy/authnz/custos_authnz.py
```diff
@@ -0,0 +1,212 @@
...
+            return False, "Failed to disconnect provider {}: {}".format(provider, str(e)), None
+
+    def _create_oauth2_session(self, state=None, scope=None):
+        client_id = self.config['client_id']
+        redirect_uri = self.config['redirect_uri']
+        if (redirect_uri.startswith('http://localhost')
+                and os.environ.get("OAUTHLIB_INSECURE_TRANSPORT", None) != "1"):
+            log.warn("Setting OAUTHLIB_INSECURE_TRANSPORT to '1' to "
+                     "allow plain HTTP (non-SSL) callback")
+            os.environ['OAUTHLIB_INSECURE_TRANSPORT'] = "1"
```
VJalili, line 134: Do you think it would make any sense to set `OAUTHLIB_INSECURE_TRANSPORT` leveraging the [VERIFY_SSL](https://github.com/galaxyproject/galaxy/blob/31a5fc24c1ce1c9d902535eb06292bedfdb6efd4/config/oidc_config.xml.sample#L16) config?

lib/galaxy/authnz/custos_authnz.py
```diff
@@ -0,0 +1,212 @@
...
+                                  verify=self._get_verify_param()).json()
+
+    def _get_custos_authnz_token(self, sa_session, user_id, provider):
+        return sa_session.query(CustosAuthnzToken).filter_by(
+            external_user_id=user_id, provider=provider).one_or_none()
+
+    def _get_current_user(self, trans):
+        return trans.user if trans.user else None
+
+    def _create_user(self, sa_session, username, email):
```
VJalili, line 163: Do you think using [this create_user](https://github.com/galaxyproject/galaxy/blob/31a5fc24c1ce1c9d902535eb06292bedfdb6efd4/lib/galaxy/model/__init__.py#L5082-L5089) method instead would make any sense?

lib/galaxy/model/__init__.py
```diff
@@ -5099,6 +5099,22 @@ def create_social_auth(cls, user, uid, provider):
...
+        # 'user' is a backref to model.User
+        self.user = user
+        self.external_user_id = external_user_id
+        self.provider = provider
+        self.access_token = access_token
+        self.id_token = id_token
+        self.refresh_token = refresh_token
+        self.expiration_time = expiration_time
+        self.refresh_expiration_time = refresh_expiration_time
+        self.raw_token = raw_token
```
VJalili: I guess now that you're storing `access_token`, `id_token`, `refresh_token`, `expiration_time`, and etc., which are all part of the `raw_token`, you would not need to persist the `raw_token` it self.

test/unit/authnz/test_custos_authnz.py
```diff
@@ -0,0 +1,474 @@
...
+class CustosAuthnzTestCase(unittest.TestCase):
+
+    _create_oauth2_session_called = False
+    _fetch_token_called = False
+    _get_userinfo_called = False
+    _raw_token = None
+
+    def setUp(self):
+        self.orig_requests_get = requests.get
+        requests.get = self.mockRequest("https://iam.scigap.org/auth/realms/test-realm/.well-known/openid-configuration", {
```
VJalili: My comment about the config setter also applies to this endpoint.

VJalili (changes requested):

lib/galaxy/authnz/custos_authnz.py
machristie: No because VERIFY_SSL is used to control whether to verify SSL of the authorization server whereas OAUTHLIB_INSECURE_TRANSPORT is used to enable insecure callback on the client server.

machristie:
Hi @VJalili . I've integrated your changes (thanks!) and I've also removed the raw_token column and made the ca_bundle a config option and removed the CA certificate from the PR. Can you please re-review when you have a chance?

I'll update the instructions on how to test locally.

VJalili (approved):

VJalili:
@machristie LGTM, +1, Thanks. 

In a different PR, I think we should work on the following function, and make sure our tests are not tied to that specific endpoint.

```python
def _get_base_idp_url(self):
    # it would be ideal is we can use a URI as the following:
    # https://test_base_uri/auth
    return 'https://iam.scigap.org/auth'
```

afgane:
Nice work @machristie - thank you a bunch for pushing this through!

machristie:
Thanks @afgane and thanks to @VJalili for his patience and dedication to making this PR a success!

PR #6917, opened 2018-10-24
Unified webpack4

dannon:
WIP, still, but this swaps us to webpack 4, which is an important step towards more intelligent client shipping.  This swaps webpack to build our styles, and bundle assets, as well.  Will document more about the changes, but I'm working on getting tests working first -- still have a few loose ends to fix, hopefully today, but I want to get the real tests running on jenkins now.

- [x] 4 workflow tests are currently expected to fail, and that's what I'm working to address first
- [x] image mapping/staging for old endpoints like trackster, etc., which do interesting asset management.
- [x] jquery rating styles

Notes:  With the dropped grunt/gulp tasks and other upgrades, this will allow us to unpin from node 9 (10 works, in my hands).  I still want to upgrade karma and qunit, but I did the minimum to get it working for this pass.  Qunit of the version we'll need does *not* work under node 9, though, so we'll want to do that carefully and re-pin against 10, maybe.

PR #7196, opened 2019-01-09
[19.01] Re-enable script staging

dannon:
This PR (temporarily) re-enables independent script transpiling/staging.  We do want to drop this long term as I did in #6917, but I'd be more comfortable leaving it in for 19.01.  It shouldn't change the behavior of anything in the 'core', but it is necessary for some plugin functionality (like IEs) to work without some additional refactoring (which I'm also working on, but will target the post-freeze -dev with).

We'll want to deprecate this now, with 19.01 being the last release supporting any of this sort of access -- it'll be disabled in 19.05 completely xref #6936

### DIFF
--- a/client/galaxy/scripts/galaxy.interactive_environments.js
+++ b/client/galaxy/scripts/galaxy.interactive_environments.js
@@ -19,7 +19,7 @@ export function clear_main_area() {
 
 export function display_spinner() {
     $("#main").append(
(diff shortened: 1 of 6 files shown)


dannon:
@jxtx This should hopefully fix up your local IE launching.  I am more comfortable with this approach for 19.01 anyway, saving the refactoring on top of the init work for dev as mentioned above.

jxtx (approved):

PR #7197, opened 2019-01-09
Apply metadata setter for default to conditionals

pvanheus:
The existing code only expects the case like:

```
<actions>
                <action name="column_names" type="metadata" default="Geneid,${alignment.element_identifier}" />
</actions>
```

whereas a conditional can be inserted within `<actions>` yielding a case like:

```
        <actions>
            <conditional name="format">
                <when value="something_good">
                    <action type="metadata" name="column_names" default="one,two,three" />
                </when>
            </conditional>
        </actions>
```

The fact that this is not handled causes the display logic to fail when `column_names` is set for a tabular dataset inside such a conditional. This patch provides a solution for that case.

### DIFF
--- a/lib/galaxy/tools/actions/__init__.py
+++ b/lib/galaxy/tools/actions/__init__.py
@@ -743,10 +743,25 @@ class DefaultToolAction(object):
                 <action name="column_names" type="metadata" default="${','.join([input.name for input in $individual_inputs ])}" />
             </actions>
         </data>
+
+        or
+        
+        <actions>
+            <conditional name="format">
+                <when value="something_good">
+                    <action type="metadata" name="column_names" default="one,two,three" />
+                </when>
+            </conditional>
+        </actions>
         """
         if output.actions:
+            action_list = output.actions.actions
             for action in output.actions.actions:
-                if action.tag == "metadata" and action.default:
+                if action.tag == "conditional":
+                    for case in action.cases:
+                        action_list.extend(case.actions)
+            for action in action_list:
+                if action.tag == "metadata" and hasattr(action, "default") and action.default:
                     metadata_new_value = fill_template(action.default, context=params).split(",")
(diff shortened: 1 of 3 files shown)


lib/galaxy/tools/actions/__init__.py
```diff
@@ -743,10 +743,25 @@ def set_metadata_defaults(self, output, dataset, tool, on_text, trans, incoming,
                 <action name="column_names" type="metadata" default="${','.join([input.name for input in $individual_inputs ])}" />
             </actions>
         </data>
+
+        or
+
```
nsoranzo: `./lib/galaxy/tools/actions/__init__.py:748:1: W293 blank line contains whitespace`

test/functional/tools/metadata_column_names_conditional.xml
```diff
@@ -0,0 +1,30 @@
...
+    ]]></command>
+    <inputs>
+        <param name="input" type="data" multiple="false" />
+	<param name="variable" type="integer" value="1" size="1" />
+    </inputs>
+    <outputs>
+        <data format="tabular" name="output">
+            <actions>
+       	        <conditional name="variable">
+		    <when value="1">
```
bgruening: @pvanheus can you fix the indentation here and below?

pvanheus: I've committed a version with new indentation - is this better?

jmchilton:
Maybe I'm confused but there doesn't seem to be anything conditional about this logic. This commit (https://github.com/jmchilton/galaxy/commit/98f273fd722902d1518ce6592ea6895b284d44c4) on top of your branch adds a second test case in a conditional and causes the first test to fail.

jmchilton:
I'm going to put this into WIP until that test passes, is that okay?

PR #7198, opened 2019-01-10
Fix galaxy instance access in visualizations

guerler:
Follow-up for #7117. Properly queues the initialization for visualizations.


dannon:
@guerler Thanks for the fix!

PR #7199, opened 2019-01-10
[19.01] Backport #7167 - interactor improvements 

mvdbeek:
This is important for targeting Galaxy < 19.01 with the new API interactor.


martenson:
@mvdbeek I was about to backport, thanks for doing so!

PR #7200, opened 2019-01-10
Assorted component bugfixes

dannon:
Commits for more details, but the second two of these are harmless in that modern browsers work around them, but I've fixed them here for correctness.  The first one, I'm not sure if it was actually causing a bug or not, will check (and backport, if necessary).


nsoranzo (approved):

dannon:
Looks like the first one is worked around by the browser, too, in my testing -- so I'm not going to backport any of this.

PR #7202, opened 2019-01-10
Eslint

dannon:
Adds a fairly lenient eslint configuration and updates the client codebase to pass.  I've included almost all of our javascript, and all of our vue components.  Includes a prettier version update and rerun (in a single commit, so it's easy to filter out and look at changes independently), which failed prior to a few fixes included here.

There's a lot I'd like to tighten up in terms of enforcement (working towards a nearer subset of https://www.npmjs.com/package/eslint-config-airbnb-base, as mentioned in the codebase), but this is a solid step forward and has caught a few errors already.

This also adds a job to travis running eslint (which passes now on the whole codebase), and overhauls the javascript-based travis jobs in general, removing a duplicate qunit job and streamlining execution, not relying on our python stack for building.


dannon:
Overhauled javascript travis runs, they work correctly now (and we don't run qunit twice anymore -- this was incorrectly set up in the matrix before).  The failure currently shown is expected (it's the last two eslint fails), and now that testing is working I'll get them resolved and this will be ready to go.

.travis.yml
```diff
@@ -38,6 +34,27 @@ matrix:
       # No version of Python is available via virtualenv on OS X workers, see https://github.com/travis-ci/travis-ci/issues/2312
       language: generic
       env: TOX_ENV=py27-first_startup
+    - env: js
```
nsoranzo: You may want to use `name` instead of `env` (which is used to also export environment variables).

.travis.yml
```diff
@@ -38,6 +34,27 @@ matrix:
...
+      node_js:
+          - 10
+      before_install:
+        - cd client
+      install:
+        - yarn
+        - yarn run build
+      script:
+        - yarn run test
+    - env: js
```
nsoranzo: Maybe this job can be merged with the previous one?

Makefile
```diff
@@ -166,7 +166,11 @@ _client-test-mocha:  ## Run mocha tests via karma
 _client-test-qunit:  ## Run qunit tests via karma
```
nsoranzo: Remove the now unused `_client-test-mocha` and `_client-test-qunit` Makefile targets?

.travis.yml
dannon: Sure, let me try that now.

Makefile
dannon: Sure.

.travis.yml
dannon: We *could*, but I think I'd prefer to keep it separate, like the python jobs.  I'm going to update the name though to indicate lint vs test.  (now pushed, we have `js-unit` and `js-lint` -- sound good?)

nsoranzo: Fair enough, thanks!

dannon:
Ok, this is ready for review and everything should pass.  I created an issue for the one trackster `no-undef` problem at #7224 and added an inline exception to make lint pass, because that's a much larger problem than the one undef and it's totally unrelated to this particular PR (has been broken quite some time...).

dannon:
Immediately conflicted.  But all green.  I'll resolve the conflict now, and it'd be great to merge this sooner rather than later.

dannon:
@guerler Thanks!

PR #7205, opened 2019-01-11
group: don't remove empty cells at the begin/start of lines

bernt-matthias:
Currently group removes empty cells at the begin and end of lines. If needed single cells could be stripped.

Additionally the sort command line is changed (hoping that this looks a bit more obvious)


bernt-matthias:
failing tests seem unrelated

mvdbeek:
LGTM, I wonder why this tool doesn't seem to be tested in the framework tests ... .
There's a couple of other things in this tool that could use an update. The requirement uses an unknown type and is in the wrong position, the interpreter attribute had been deprecated etc.
We probably need to bump the tool version too, since it possibly changes the output.

bernt-matthias:
> LGTM, I wonder why this tool doesn't seem to be tested in the framework tests ... .

Don't know

> There's a couple of other things in this tool that could use an update. The requirement uses an unknown type and is in the wrong position, the interpreter attribute had been deprecated etc.
> We probably need to bump the tool version too, since it possibly changes the output.

Indeed, I will take care of it.

bernt-matthias:
Maybe framework tests don't run because the test data was absent? This might be the case in more tools...

mvdbeek:
The test data should come from https://github.com/galaxyproject/galaxy-test-data, that seems unlikely as a cause.

bernt-matthias:
Ah, OK. I always assumed test data is in the test-data directory..

bgruening:
This will potentially break many workflows. Cleaning up these tools probably make sense, but I would not touch the functionality. We should either migrate them or use the text processing tools imho.

bgruening:
This will potentially break many workflows. Cleaning up these tools probably make sense, but I would not touch the functionality. We should either migrate them or use the text processing tools imho.

bernt-matthias:
I totally understand your concern, but tabs should not be removed from tab delimited data. IMHO I think that (intended/documented) functionality is unchanged. 

The only case that I can think of is tables that have a fixed number (x) of leading empty cells on each row and workflows that actually shift column indices by the offset of x. But this borderline case could be documented in the release notes. 

- Removing `strip` for lines only prevents the removal of empty leading or training cells. If this tool was used on data with empty 1st or last cells then the results were wrong anyway. As a compromise one could reintroduce `strip([' '])`, ie only remove leading and trailing spaces. 
- Using `exit(1)` instead of `exit()` (which defaults to `exit(0)`) only ensures that errors are correctly reported by exit code.

bernt-matthias:
`strip` was introduced here: https://github.com/galaxyproject/galaxy/commit/82dde6e37e70b84d9a721905bffd57e6bd431fc7#diff-2a19392373ee09918603fe641e729512 . Before that `line.rstrip( '\r\n' )` was used. This (or strip(' \r\n')) might seem sensible. 

But, since the strip is called anyway for each (considered cell) this is not necessary.

PR #7209, opened 2019-01-12
introduce the ability to favorite tools

martenson:
# favorite tools you like
galaxy

# display only tools you have favorited
galaxy

### DIFF
--- a/client/galaxy/scripts/entry/panels/tool-panel.js
+++ b/client/galaxy/scripts/entry/panels/tool-panel.js
@@ -6,6 +6,7 @@ import _l from "utils/localization";
 import _ from "libs/underscore";
 import { getGalaxyInstance } from "app";
 import { getAppRoot } from "onload";
+import iconButton from "mvc/ui/icon-button";
 
 var ToolPanel = Backbone.View.extend({
     initialize: function(page, options) {
@@ -42,13 +43,25 @@ var ToolPanel = Backbone.View.extend({
             default_extension: config.default_extension
         });
 
+        this.favorite_button = new iconButton.IconButtonView({
+            model: new iconButton.IconButton({
+                title: _l("Show favorites"),
+                icon_class: "fa fa-star-o",
+                href: "javascript:void(0)",
+                on_click: e => {
+                    $("#tool-search-query").val("favorites").trigger("change");
+                },
+            })
+        });
+        this.favorite_button.render();
+
         // add uploader button to Galaxy object
         Galaxy.upload = this.upload_button;
 
         // components for panel definition
         this.model = new Backbone.Model({
             title: _l("Tools"),
-            buttons: [this.upload_button]
+            buttons: [this.upload_button, this.favorite_button]
         });
 
         // build body template
--- a/client/galaxy/scripts/entry/panels/tool-panel.js
+++ b/client/galaxy/scripts/entry/panels/tool-panel.js
@@ -6,7 +6,7 @@ import _l from "utils/localization";
 import _ from "libs/underscore";
 import { getGalaxyInstance } from "app";
 import { getAppRoot } from "onload";
-import iconButton from "mvc/ui/icon-button";
+import Buttons from "mvc/ui/ui-buttons";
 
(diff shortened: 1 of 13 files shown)


bgruening:
@martenson what about having a "star" or a link next to the search bar indicating favorites instead of searching for them? Similar to the `deleted` datasets link on the history maybe?

martenson:
@bgruening I am open to that but I do not think a button is necessary. I want to make tool search more powerful without overloading the interface as it happened with histories - Imo we should devise a way how to hint these extra/reserved searches. Something like a `?` popover with hints and link to docs.

nsoranzo:
@martenson If the goal is to make access to favourite tools faster, than a button makes more sense also for me than having to type something in the search box (IIUC).

martenson:
@bgruening @mvdbeek @nsoranzo @erasche you got it

martenson:
@guerler The upload button behaves very non-standard in the standard panel button menu, any ideas?

galaxy

martenson:
thanks a lot for helping @guerler !

martenson:
ready for review I think

jmchilton:
I'm going to approve this because it has languished and I'm not sure we need to spend more time on it. I would just say that despite my +1 I think we should not use user preferences for this - we should store the data in the database in a more structured way. For instance a ToolFavorite class with a tool ID that is linked to the User for instance. User preferences are completely untyped and unstructured and are a hack to provide an extension point for installers - we don't need to avoid creating a table because this is not optional functionality and not an extension, we can do a better job with the modeling as a result.

jmchilton (approved):

bgruening:
@jmchilton you are correct. Worth to open an issue about this I guess.

hexylena:
in case anyone was curious :)

```
(venv) galaxy@sn04:~$ psql -qAt -c "select value::json->'tools' from user_preference where name = 'favorites'" | jq -s add | jq '. | @tsv' -r | sed 's/\t/\n/g' | sort | uniq -c | sort -nr | head -n 20
     20 toolshed.g2.bx.psu.edu/repos/devteam/fastqc/fastqc/0.72+galaxy1
     12 toolshed.g2.bx.psu.edu/repos/iuc/multiqc/multiqc/1.7
     11 toolshed.g2.bx.psu.edu/repos/iuc/hisat2/hisat2/2.1.0+galaxy4
     11 toolshed.g2.bx.psu.edu/repos/iuc/deseq2/deseq2/2.11.40.6
      8 toolshed.g2.bx.psu.edu/repos/crs4/prokka/prokka/1.13
      8 Filter1
      7 toolshed.g2.bx.psu.edu/repos/iuc/featurecounts/featurecounts/1.6.4
      6 toolshed.g2.bx.psu.edu/repos/pjbriggs/trimmomatic/trimmomatic/0.36.5
      6 toolshed.g2.bx.psu.edu/repos/nml/spades/spades/3.12.0+galaxy1
      6 toolshed.g2.bx.psu.edu/repos/iuc/volcanoplot/volcanoplot/0.0.3
      5 toolshed.g2.bx.psu.edu/repos/iuc/rgrnastar/rna_star/2.6.0b-2
      5 toolshed.g2.bx.psu.edu/repos/bgruening/trim_galore/trim_galore/0.4.3.1
      4 toolshed.g2.bx.psu.edu/repos/pjbriggs/trimmomatic/trimmomatic/0.32.3
      4 toolshed.g2.bx.psu.edu/repos/iuc/trinity/trinity/2.8.4
      4 toolshed.g2.bx.psu.edu/repos/iuc/stringtie/stringtie/1.3.6
      4 toolshed.g2.bx.psu.edu/repos/iuc/edger/edger/3.24.1
      4 toolshed.g2.bx.psu.edu/repos/devteam/fastq_groomer/fastq_groomer/1.1.1
      4 Cut1
      4 cat1
      3 toolshed.g2.bx.psu.edu/repos/lparsons/htseq_count/htseq_count/0.9.1
```

martenson:
@erasche awesome, people are starting to use it!

PR #7210, opened 2019-01-12
univa: code cleanup

bernt-matthias:
some debug messages somehow survived

### DIFF
--- a/lib/galaxy/jobs/runners/univa.py
+++ b/lib/galaxy/jobs/runners/univa.py
@@ -1,7 +1,7 @@
 """
-Kind of fail save job control via the DRMAA API / qstat and qacct.
+Job control via the DRMAA API / qstat and qacct.
 
-Known to work on the UNIVA grid engine. 
+Known to work on the UNIVA grid engine.
 
 known bugs/problems:
 - if a job runs longer than the time limits of the queue two things happen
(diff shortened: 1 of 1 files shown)


lib/galaxy/jobs/runners/univa.py
```diff
@@ -86,21 +79,21 @@ def _complete_terminal_job(self, ajs, drmaa_state, **kwargs):
             mem_wasted = extinfo["memory_wasted"]
             slots = extinfo["slots"]
 
-            # log.debug("UnivaJobRunner:_complete_terminal_job ({jobid}) memviolation {mv}".format(jobid=ajs.job_id, mv=memviolation))
+                # log.debug("UnivaJobRunner:_complete_terminal_job ({jobid}) memviolation {mv}".format(jobid=ajs.job_id, mv=memviolation))
```
mvdbeek: Can you just drop the comment ?
Else you need to drop the extra indentation here.

lib/galaxy/jobs/runners/univa.py
```diff
@@ -1,6 +1,7 @@
 """
-Kind of fail save job control via the DRMAA API.
-Tested for UNIVA grid engine
+Kind of fail save job control via the DRMAA API / qstat and qacct.
```
mvdbeek: fail-safe ?

mvdbeek (commented):
You also need to drop the galaxy.model.Job import (https://travis-ci.org/galaxyproject/galaxy/jobs/478768574)

lib/galaxy/jobs/runners/univa.py
bernt-matthias: Will drop the extra indentation. If debugging is necessary its easier to renable. I guess that some debugging might be necessary if others start using it. 

Will do it everywhere.

lib/galaxy/jobs/runners/univa.py
bernt-matthias: Don't know how this got there :) .. consider it gone.

mvdbeek (approved):

mvdbeek:
Thanks @bernt-matthias !

Issue #6674, opened 2018-09-06
'Import from path'  in data libraries skip filenames with '+' character


ashvark:
For some reasion, "import files from path" in data libraries does not accept filenames having '+' character in it.
Actually, I am not getting any error message. Galaxy is simply ignoring/skipping the files. Even, I could not see any errors in the galaxy logs.

ashvark:
#7211 has some fix to this issue

mvdbeek:
Fixed by https://github.com/galaxyproject/galaxy/pull/7432.

PR #7211, opened 2019-01-14
Data libraries - Import from paths - encoding special characters in the path

ashvark:
Fix for the issue #6674


mvdbeek:
@galaxybot test this

mvdbeek:
@nsoranzo @martenson this is similar to the spaces and `+` in tool id / versions in https://github.com/galaxyproject/galaxy/pull/6737, do we like this ?

martenson:
per @mvdbeek suggestion I have made an alternative fix that should resolve this (and similar situations for other importing code paths) : https://github.com/galaxyproject/galaxy/pull/7432

mvdbeek:
I'm gonna close this, https://github.com/galaxyproject/galaxy/pull/7432 is merged and should fix this as well. Thanks alot for taking a look @ashvark!

PR #7213, opened 2019-01-14
Fix global var badness in setup_external_metadata

jmchilton:


lib/galaxy/model/metadata.py
```diff
@@ -771,6 +768,7 @@ def __get_filename_override():
                 for meta_key, spec_value in dataset.metadata.spec.items():
                     if isinstance(spec_value.param, FileParameter) and dataset.metadata.get(meta_key, None) is not None:
                         metadata_temp = MetadataTempFile()
+                        metadata_temp.tmp_dir = tmp_dir
```
mvdbeek, line 771: As far as I can tell it's only used in https://github.com/galaxyproject/galaxy/blob/cb5fc4bb16dc075094f00b59a363fa8ec1e2d8c2/lib/galaxy/metadata/parameters.py#L605, should we maybe make `tmp_dir` a regular instance variable ?

mvdbeek (approved):

lib/galaxy/model/metadata.py
jmchilton: We're setting this as a regular instance variable I think...

```python
>>> class A():
...    x = 7
...
>>> a = A()
>>> A.x
7
>>> a.x
7
>>> a.x = 5
>>> a.x
5
>>> A.x
7
```

But I guess we should initialize it that way as well.

PR #4532, opened 2017-08-31
Export collections when exporting an history

FredericBGA:
#4345 

Here are some first commits.

Collections are exported and their attributes and files are now present in the archive.

I have to deal now with import. I hope that I have exported enough stuff within the archive

Exports works within my install (I've add a column in one table of the database: job_export_history_archive). 
It's not heavily tested (two small histories, one with only one collection, another one with two collections)

Collections are type "list", I haven't tried type "pair" yet

I used pep8 in order to avoid common errors of style

I don't know what to do here: 

> lib/galaxy/model/migrate/versions/

PR #7214, opened 2019-01-14
Cleanup History Export CLI Call

jmchilton:
Also makes it future proof to add more model attribute files for export (collections for instance).

Variants of these commits have been in #4532 for a long time.

Cherry-picked from https://github.com/galaxyproject/galaxy/pull/7367

### DIFF
--- a/lib/galaxy/tools/imp_exp/export_history.py
+++ b/lib/galaxy/tools/imp_exp/export_history.py
@@ -98,17 +98,24 @@ def main():
     # Parse command line.
     parser = optparse.OptionParser()
     parser.add_option('-G', '--gzip', dest='gzip', action="store_true", help='Compress archive using gzip.')
+    parser.add_option('--galaxy-version', dest='galaxy_version', help='Galaxy version that initiated the command.', default=None)
     (options, args) = parser.parse_args()
+    galaxy_version = options.galaxy_version
+    if galaxy_version is None:
+        galaxy_version = "19.01" if len(args) == 4 else "19.05"
+
     gzip = bool(options.gzip)
-    if len(args) == 2:
+    if galaxy_version == "19.01":
+        # This job was created pre 18.0X with old argument style.
+        history_attrs, dataset_attrs, job_attrs, out_file = args
+    else:
+        assert len(args) >= 2
         # We have a 19.0X directory argument instead of individual arguments.
-        temp_directory, out_file = args
+        temp_directory = args[0]
+        out_file = args[1]
         history_attrs = os.path.join(temp_directory, 'history_attrs.txt')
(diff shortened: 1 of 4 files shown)


lib/galaxy/tools/imp_exp/export_history.py
```diff
@@ -100,7 +100,15 @@ def main():
     parser.add_option('-G', '--gzip', dest='gzip', action="store_true", help='Compress archive using gzip.')
     (options, args) = parser.parse_args()
     gzip = bool(options.gzip)
-    history_attrs, dataset_attrs, job_attrs, out_file = args
+    if len(args) == 2:
```
mvdbeek: Could we future proof it by adding a `galaxy_version` flag to the argument parser ?

jmchilton: I'll work on this - but this seems like overkill we don't apply to other scripts. What if we just used named arguments for everything instead?

mvdbeek: Sure, that'd be fine.

mvdbeek (approved):

nsoranzo:
According to `git bisect`, commit db78c26c74a29b02f0e5dad48409cd1ac4d73f0a in this PR is causing a timeout of BioBlend Travis builds during the `test_download_history` test at this point https://github.com/galaxyproject/bioblend/blob/master/bioblend/_tests/TestGalaxyHistories.py#L154 

This is due to the fact that the Galaxy histories API keeps replying with 202: https://github.com/galaxyproject/bioblend/blob/master/bioblend/galaxy/histories/__init__.py#L542

See https://travis-ci.org/nsoranzo/bioblend/builds/504466566

I have looked at the commit and related code for a few hours, but I still don't understand why it should cause this issue and why I cannot reproduce it locally, help welcome!

jmchilton:
Any chance to see the Galaxy logs on that? Waiting on the 202 makes sense for a bit as the export is readied but obviously it should switch at some point. I'm guessing the job is failing over and over and new export jobs are being created - I'd hope the problem would be clear from the logs but perhaps it wouldn't be.

jmchilton:
Galaxy doesn't have a test for an empty history export... it'd be good to have one to match bioblend. Then if we could synchronize the download endpoint - I'll have to research more because it isn't obvious if they are or not currently.

nsoranzo:
> Any chance to see the Galaxy logs on that?

Here is the error in the Galaxy logs: https://travis-ci.org/nsoranzo/bioblend/builds/504995632#L4527

Not sure how `./database/tmp` ends up there.

nsoranzo:
> Not sure how ./database/tmp ends up there.

Should be fixed by #7517.

PR #7216, opened 2019-01-15
[18.09] Update configuration samples

mvdbeek:
This is all properly documented in https://docs.galaxyproject.org/en/master/admin/nginx.html#serving-galaxy-at-a-url-prefix, but missing and/or confusing in the galaxy.yml.sample file.

The way i've hacked in the commented `#mount` is probably not the right way to do it.

### DIFF
--- a/config/galaxy.yml.sample
+++ b/config/galaxy.yml.sample
@@ -98,12 +98,6 @@ uwsgi:
 
 galaxy:
 
-  # If not running behind a proxy server, you may want to enable gzip
-  # compression to decrease the size of data transferred over the
-  # network.  If using a proxy server, please enable gzip compression
-  # there instead.
-  #filter-with: gzip
-
   # If you serve Galaxy at a URL prefix and you're running more than one
   # Galaxy instance behind one hostname, you will want to set this to
   # the same path as the prefix in the filter above.  This value becomes
--- a/config/galaxy.yml.sample
+++ b/config/galaxy.yml.sample
@@ -98,11 +98,11 @@ uwsgi:
 
 galaxy:
 
-  # If Galaxy is served at a URL prefix and you are running more than
-  # one Galaxy instance behind one hostname, you will want to set this
-  # to the same path as the prefix in the mount uWSGI configuration
-  # above. This value becomes the "path" attribute set in the cookie so
-  # the cookies from each instance will not clobber each other.
+  # When running multiple Galaxy instances under separate URL prefixes
+  # on a single hostname, you will want to set this to the same path as
+  # the prefix set in the uWSGI "mount" configuration option above. This
+  # value becomes the "path" attribute set in the cookie so the cookies
+  # from one instance will not clobber those from another.
   #cookie_path: ''
 
   # By default, Galaxy uses a SQLite database at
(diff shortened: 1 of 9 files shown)


lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -40,8 +40,8 @@ mapping:
         required: false
         desc: |
           If running behind a proxy server and Galaxy is served from a subdirectory,
-          enable the proxy-prefix filter and set the prefix in the
-          [filter:proxy-prefix] section above.
+          enable the proxy-prefix filter, uncomment the 'mount: /galaxy=galaxy.webapps.galaxy.buildapp:uwsgi_app()'
```
natefoo: Probably also need to drop `enable the proxy-prefix filter` here?

natefoo (approved):
The commented out option as a key in `config_manage.py` seems a little hacky but I don't have any real complaints with this. Thanks for adding this to the samples.

lib/galaxy/webapps/galaxy/config_schema.yml
martenson: @natefoo does this still stand? ( `#filter-with: proxy-prefix` is still relevant, no?)

natefoo: Not with uWSGI

martenson:
I think the following commit should make this more consistent in wording and remove the old Paste option: https://github.com/galaxyproject/galaxy/commit/d72f82de98f464ace59db312fe9db398793c3494

martenson:
opened a PR https://github.com/mvdbeek/galaxy/pull/11

mvdbeek:
Thanks @martenson!

config/galaxy.yml.sample
```diff
@@ -90,16 +98,17 @@ uwsgi:
...
-  # If proxy-prefix is enabled and you're running more than one Galaxy
-  # instance behind one hostname, you will want to set this to the same
-  # path as the prefix in the filter above.  This value becomes the
-  # "path" attribute set in the cookie so the cookies from each instance
-  # will not clobber each other.
+  # If not running behind a proxy server, you may want to enable gzip
+  # compression to decrease the size of data transferred over the
+  # network.  If using a proxy server, please enable gzip compression
+  # there instead.
+  #filter-with: gzip
```
natefoo: This is also a Paste-only option.

doc/source/admin/galaxy_options.rst
```diff
@@ -3,10 +3,11 @@
 ~~~~~~~~~~~~~~~
```
natefoo: We can probably just remove this entire section

config/galaxy.yml.sample
mvdbeek: So we drop it ?

natefoo: Yep

martenson:
This is ready I think.

nsoranzo (approved):

nsoranzo:
I've added a commit to remove `filter-with` from reports config/docs and update `cookie_path` also for reports and Tool Shed.

martenson (approved):

config/reports.yml.sample
```diff
@@ -80,34 +80,30 @@ uwsgi:
...
-  # datasets, this should be shared through the cluster
+  # Where dataset files are saved. Temporary storage for additional
+  # datasets, this should be shared through the cluster.
   #file_path: database/files
 
-  # Where dataset files are saved Temporary storage for additional
-  # datasets, this should be shared through the cluster
+  # Where dataset files are saved. Temporary storage for additional
+  # datasets, this should be shared through the cluster.
   #new_file_path: database/tmp
```
natefoo, line 104: Not the fault of this PR, but the comments for `file_path` and `new_file_path` are duplicated and not correct for either. I'll push a correction.

nsoranzo (approved):

natefoo (approved):

martenson:
good job everybody, soon we will have enough +1s that we can change the repo ownership :D

PR #7217, opened 2019-01-15
Fix history import and export logic related to deleted datasets.

jmchilton:
At some point they stop being exported all together (during some of Carl's initial work toward managers). When imported, they were always imported as undeleted, even though I think it makes more sense that they are imported and restored, but with the deleted flag intact. This fixes both of those problems and brings in various testing enhancements from downstream work on getting collection export working properly.


jmchilton:
Closed in lieu #7441 - that contains better versions of all these commits.

PR #7220, opened 2019-01-15
Verify/update .venv's node version 

dannon:
...is in line with what we've specified as being locked, rather than just checking if node exists.

As it was, if .venv ever had a node, we would *not* update it.  This would result in problems down the road.  With this change, we verify the node version is the currently required/locked version, and install that as appropriate.  Not sure if this will/should be backported, yet.

### DIFF
--- a/scripts/common_startup.sh
+++ b/scripts/common_startup.sh
@@ -18,6 +18,7 @@ CREATE_VENV=1
 REPLACE_PIP=$SET_VENV
 COPY_SAMPLE_FILES=1
 SKIP_CLIENT_BUILD=${GALAXY_SKIP_CLIENT_BUILD:-0}
+NODE_VERSION="10.13.0"
 
 for arg in "$@"; do
     [ "$arg" = "--skip-eggs" ] && FETCH_WHEELS=0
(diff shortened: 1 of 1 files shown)


scripts/common_startup.sh
```diff
@@ -18,6 +18,7 @@ CREATE_VENV=1
 REPLACE_PIP=$SET_VENV
 COPY_SAMPLE_FILES=1
 SKIP_CLIENT_BUILD=${GALAXY_SKIP_CLIENT_BUILD:-0}
+NODE_VERSION="10.13.0"
```
jmchilton, line 21: As soon as I merged it, I realized I think this should be more readily configurable. What do you think about?

```
NODE_VERSION=${GALAXY_NODE_VERSION:-"10.13.0"}
```

Not a big deal either way though.

dannon: Hrmm, maybe?  Can't hurt, but I don't know that we really want people configuring this one.  If they want to run their own node/yarn and try different client builds, they already can.

The more I think about this, I want to backport it to 18.09 and merge forward with pinned node versions starting there, so I can add it when I do that if you want.

Issue #7140, opened 2018-12-20
Saving a broken workflow hangs forever and requires a page refresh


hexylena:
Seen on dev with @shiltemann, if you upload a 'broken' workflow with missing tools/etc, and try to save it, the safe dialog hangs and renders the interface unresponsive. You must refresh the page and lose your changes. It should just show the error message / toast that the saving failed / something, and let you regain control of the interfac.e

guerler:
Can you share a workflow which causes that?

guerler:
This happens to any workflow with missing tools. It seems to be caused by #6741.

mvdbeek:
Should be fixed by https://github.com/galaxyproject/galaxy/pull/7227, please re-open if you see it again.

PR #7227, opened 2019-01-17
[18.09] Properly deal with missing tools in workflows

mvdbeek:
If tools were missing and you tried to save a workflow the client would fail
when accessing `node.config_form.version`.

If a subworkflow is missing tools the entire parent workflow couldn't
be loaded into the editor, because we raised an exception. This is now
only logged as an error. On the user side we now prevent running and
saving (until the offending modules are removed) such workflows by actually implementing the `get_errors` method
for the SubWorkflowModule class.

This should fix https://github.com/galaxyproject/galaxy/issues/7140.

### DIFF
--- a/lib/galaxy/exceptions/__init__.py
+++ b/lib/galaxy/exceptions/__init__.py
@@ -103,6 +103,11 @@ class ToolMissingException(MessageException):
     status_code = 400
     err_code = error_codes.USER_TOOL_MISSING_PROBLEM
 
+    @property
+    def tool_id(self):
+        if 'tool_id' in self.extra_error_info:
+            return self.extra_error_info['tool_id']
+
 
 class RequestParameterInvalidException(MessageException):
     status_code = 400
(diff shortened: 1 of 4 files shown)


mvdbeek:
This is what loading and saving a workflow w/ subworkflows and missing tools looks like now:
screen shot 2019-01-17 at 13 35 10
screen shot 2019-01-17 at 13 35 19
screen shot 2019-01-17 at 13 36 24

mvdbeek:
the selenium error is relevant, I think I got a fix for it.

lib/galaxy/exceptions/__init__.py
```diff
@@ -103,6 +103,11 @@ class ToolMissingException(MessageException):
     status_code = 400
     err_code = error_codes.USER_TOOL_MISSING_PROBLEM
 
+    @property
+    def tool_id(self):
+        if 'tool_id' in self.extra_error_info:
+            return self.extra_error_info['tool_id']
```
nsoranzo: What about overloading `__init__` instead?

```python
    def __init__(self, err_msg=None, type="info", tool_id=None, **extra_error_info):
        super(ToolMissingException, self).__init__(err_msg, type, **extra_error_info)
        self.tool_id = tool_id
```

mvdbeek: Sure, that works for me.

nsoranzo (approved):

jmchilton:
Wonderful as always!

PR #7233, opened 2019-01-17
[19.01] Drop event handling fixes

### LINKED ISSUE #7230: Dropping workflow files into the workflow list panel broken
Instead of uploading the workflow the browser displays the workflow content.
Probably same root cause as #7138 #7137.

dannon:
Fixes #7230.

It looks like `jquery.event.drop`, now that we strap it onto jquery everywhere via our singular `jquery.custom`, is the culprit breaking backbone drop events.  This PR strips jquery.event.drop completely -- I think we only ever actually used this in the workflow editor, and now those terminals are Backbone.Views with that event handling integrated.  

(this is going to be WIP for a while since I need to replace the functionality in the editor, etc.)


dannon:
This turned out to be a larger project and I'm not going to target 19.01 with it since it touches so much.  Going to try to get in a more isolated fix for #7230.

PR #7235, opened 2019-01-17
remove underline from history datasets

martenson:
I don't think we are losing any information with the ink used for these

galaxy_and_galaxy

### DIFF
--- a/client/galaxy/style/scss/list-item.scss
+++ b/client/galaxy/style/scss/list-item.scss
@@ -80,7 +80,6 @@ $vertical-gap: 8px;
         .title {
             display: inline;
             font-weight: bold;
-            text-decoration: underline;
             word-wrap: break-word;
             word-break: break-all;
             line-height: 16px;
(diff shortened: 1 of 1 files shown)


guerler:
We should probably add an underline when hovering over a history dataset. What do you think?

martenson:
@guerler given that the clicktarget is not the text but the whole dataset 'rectangle' and that we properly change the cursor, I am hoping we can save that too.

jmchilton (approved):

guerler:
I think we should underline the title when hovering over it. It helps with the contrast and clarifies the behavior. It also keeps it consistent with the behavior of tool sections in the tool panel. When expanded, clicking on the title collapses the section/dataset.

martenson:
@guerler I will make the change

martenson:
@guerler I have added the requested change

martenson:
@dannon thanks for the merge

nsoranzo:
Looks nice, thanks all!

PR #7251, opened 2019-01-22
[WIP] Fix rule builder data handling

guerler:


Nerdinacan:
I just came here to do this exact fix. You win the race!

nsoranzo:
The js-lint Travis job failed: https://travis-ci.org/galaxyproject/galaxy/jobs/482729358

dannon:
Yeah, this one's on me, I guess, though maybe it's actually demonstrating the underlying problem here -- I actually made this change in #7202 because computed properties shouldn't have any external side effects.  They're cached and may or may not update unexpectedly, etc.  If this method is changing internal data (intentionally), it *should* be a method and not a computed property, and there should be something else that actually gets cached if necessary.

What I don't get is how the change where I swapped this actually *did* pass the selenium tests, if you look at that previous PR.  All tests were passing at https://github.com/galaxyproject/galaxy/pull/7202/commits/a482c5ce4cd70b208f07e79eb0f5c08ec1e67292, which is the last substantive changeset before the merge.

Nerdinacan:
@Dannon Yeah I get your motivation. If this was going to be a computed it would have to be broken into more pieces to populate the many things that are set in that one big method.

As it is, if you want to call it as a method, it obviously just needs some parentheses added here and there.

If only there was some kind of technology that could observe the starting value, filter on the properties, compose intermediate immutable processing stages and emit one or more multiple output values in several places via a tree-like processing pipeline...  :)

dannon:
@Nerdinacan @guerler I'm running the tests locally to see what the problem is, but I can take a different stab at a correct fix for this that doesn't have side effects today (especially since I apparently broke it...)

guerler:
Yeah, it's surprising that the test passed in an earlier iteration. Closing this PR in favor of #7256.

Issue #7250, opened 2019-01-21
Tool installation breaks shed_tool_conf.xml


loraine-gueguen:
Each time a new tool or revision is installed through the toolshed, the shed_tool_conf.xml that is updated with the new tool / revision is broken because line breaks are added inside the tag values (see below) and the shed_tool_conf.xml is not loaded correctly when galaxy is restarted (see logs below). This can be corrected by removing the break lines.

This happens since we have updated Galaxy from 17.05 to 18.09
Do you have an idea why this is happening and what can be done to solve this problem ?

Extract of updated shed_tool_conf.xml:

>     <section id="fasta_fastq_manipulation" name="Fasta Fastq Manipulation" version="">
>         <tool file="toolshed.g2.bx.psu.edu/repos/peterjc/seq_filter_by_id/fb1313d79396/seq_filter_by_id/tools/seq_filter_by_id/seq_filter_by_id.xml" guid="toolshed.g2.bx.psu.edu/repos/peterjc/seq_filter_by_id/seq_filter_by_id/0.2.5">
>             <tool_shed>
>                 toolshed.g2.bx.psu.edu
>             </tool_shed>
>             <repository_name>
>                 seq_filter_by_id
>             </repository_name>
>             <repository_owner>
>                 peterjc
>             </repository_owner>
>             <installed_changeset_revision>
>                 fb1313d79396
>             </installed_changeset_revision>
>             <id>
>                 toolshed.g2.bx.psu.edu/repos/peterjc/seq_filter_by_id/seq_filter_by_id/0.2.5
>             </id>
>             <version>
>                 0.2.5
>             </version>
>         </tool>
>     </section>
> 

Extract of error logs when Galaxy is restarted:

> galaxy.tools.toolbox.base DEBUG 2019-01-21 21:37:22,733 [p:27522,w:0,m:0] [MainThread] Error while loading tool toolshed.g2.bx.psu.edu/repos/peterjc/seq_filter_by_id/fb1313d79396/seq_filter_by_id/tools/seq_filter_by_id/seq_filter_by_id.xml
> galaxy.tools.toolbox.base ERROR 2019-01-21 21:37:22,917 [p:27522,w:0,m:0] [MainThread] Error reading tool from path: toolshed.g2.bx.psu.edu/repos/peterjc/seq_filter_by_id/fb1313d79396/seq_filter_by_id/tools/seq_filter_by_id/seq_filter_by_id.xml
> Traceback (most recent call last):
[...]

PR #7252, opened 2019-01-22
[18.09] strip() text when parsing xml

mvdbeek:
This should make parsing shed_tool_conf.xml files (and others) more robust.
Should fix https://github.com/galaxyproject/galaxy/issues/7250.

### DIFF
--- a/lib/galaxy/util/__init__.py
+++ b/lib/galaxy/util/__init__.py
@@ -222,6 +222,11 @@ def parse_xml(fname):
     tree = ElementTree.ElementTree()
     try:
         root = tree.parse(fname, parser=ElementTree.XMLParser(target=DoctypeSafeCallbackTarget()))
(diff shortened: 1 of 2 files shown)


dannon (commented):
Looks good to me.  Another improvmeent would be swapping to lxml, which has a 'strip whitespace' parser option to make it cleaner.  We already have lxml as a dep.

dannon (approved):

mvdbeek:
I looked at that (`etree.XMLParser(remove_blank_text=True)`), but it only strips "safe" whitespaces.
If `https://github.com/galaxyproject/galaxy/issues/7250` strikes you'll still have whitespaces inside the tags.

mvdbeek:
So uhm, this might remove intentional leading or trailing whitespaces. Which is fine, I think ?

Issue #5619, opened 2018-02-28
Installing repo in new state fails in admin panel


mvdbeek:
These are the query parameters, which look wrong to me ...
```
operation=install&id=05fbbc97ac1631ca?async=false&sort=name&page=1&show_item_checkboxes=true&advanced_search=false&operation=install&id=05fbbc97ac1631ca
```
![screen shot 2018-02-28 at 12 20 16](https://user-images.githubusercontent.com/6804901/36785034-c7560424-1c81-11e8-9390-144b7d7cd2a6.png)

The traceback is

```
TypeError: Non-hexadecimal digit found
  File "galaxy/web/framework/middleware/sentry.py", line 40, in __call__
    iterable = self.application(environ, start_response)
  File "/bioinfo/guests/mvandenb/galaxy/.venv/local/lib/python2.7/site-packages/paste/recursive.py", line 85, in __call__
    return self.application(environ, start_response)
  File "/bioinfo/guests/mvandenb/galaxy/.venv/local/lib/python2.7/site-packages/paste/httpexceptions.py", line 640, in __call__
    return self.application(environ, start_response)
  File "galaxy/web/framework/base.py", line 136, in __call__
    return self.handle_request(environ, start_response)
  File "galaxy/web/framework/base.py", line 215, in handle_request
    body = method(trans, **kwargs)
  File "galaxy/web/framework/decorators.py", line 98, in decorator
    return func(self, trans, *args, **kwargs)
  File "galaxy/webapps/galaxy/controllers/admin_toolshed.py", line 591, in manage_repository
    repository = repository_util.get_installed_tool_shed_repository(trans.app, repository_id)
  File "tool_shed/util/repository_util.py", line 399, in get_installed_tool_shed_repository
    repository_ids = [app.security.decode_id(i)for i in id]
  File "galaxy/web/security/__init__.py", line 110, in decode_id
    return int(unicodify(id_cipher.decrypt(codecs.decode(obj_id, 'hex'))).lstrip("!"))
  File "encodings/hex_codec.py", line 42, in hex_decode
    output = binascii.a2b_hex(input)
```

`input` in this instance is `u'05fbbc97ac1631ca?async=false'`

nsoranzo:
Is that only on dev?

mvdbeek:
No, I noticed this on 18.01 first and then tried again on dev, so the traceback is from dev.

PR #7253, opened 2019-01-22
[18.09] Fix install operation in installed_repository_grid

mvdbeek:
The incoming url keywords look like this:
```
operation=install&id=ebfb8f50c6abde6d?async=false&sort=name&page=1&show_item_checkboxes=true&advanced_search=false&operation=install&id=ebfb8f50c6abde6d
```
and the id parameter gets parsed into a list:
```
['ebfb8f50c6abde6d?async=false', 'ebfb8f50c6abde6d']
```

Obviously there's something wrong in the grid or mako,
but this works fine.

Fixes https://github.com/galaxyproject/galaxy/issues/5619


jmchilton (approved):
Approving but there is a linting issue, thanks for the fix @mvdbeek !

mvdbeek:
Oops, thanks, fixed that in a rebase.

PR #7254, opened 2019-01-22
Determine collection_type for all workflow steps

mvdbeek:
This allows us to get the correct collection type for subworkflows, which is a pretty major aspect if you want to use multiple subworkflows or want to link subworkflow outputs to collection inputs. This probably doesn't return the correct nesting level for unlinked, multiplied inputs (but that also doesn't work for regular workflows at this time, and I think it can't). But this already enables things like this:
screen shot 2019-01-22 at 19 41 53

### DIFF
--- a/client/galaxy/scripts/mvc/workflow/workflow-terminals.js
+++ b/client/galaxy/scripts/mvc/workflow/workflow-terminals.js
@@ -451,6 +451,7 @@ var InputCollectionTerminal = BaseInputTerminal.extend({
     update: function(input) {
         this.multiple = false;
         this.collection = true;
+        this.collection_type = input.collection_type;
         this.datatypes = input.extensions;
         var collectionTypes = [];
         if (input.collection_types) {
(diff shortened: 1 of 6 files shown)


jmchilton (approved):

jmchilton:
Unit tests are failing on Travis. Looks really good though - this is obviously really needed and thanks for doing it.

mvdbeek:
Thanks @jmchilton! I noticed another small issue when actually using this, we should only fill in `collection_type` for subworkflows. If we do this for the parent we nest way too deep, because the client also attempts to figure out the collection type. And I guess this is the right behaviour.

mvdbeek:
5ac0802 fixes the subworkflow's input data collection appearing in the workflow outputs (see 0:output in the list:list input to list:list:list example in the PR description) and c5726c3 fills in the collection_type for collection input modules, which fixes the mapping over calculation.

I think this is ready again.

nsoranzo:
Unit tests are failing: https://travis-ci.org/galaxyproject/galaxy/jobs/485428749

mvdbeek:
Ah, that's probably relevant -- we don't force a `collection_type` to be set. I'll make this more defensive.

mvdbeek:
@galaxybot test py3

Issue #7249, opened 2019-01-21
After logging out, the "User" menu still displays email address, back button stalls and reveals last active history


jennaj:
The logout behavior has changed. **Should we change it back?** 

This changed in maybe release `18.05`? Tried to get used to it but think it is problematic. 

1. Is cumbersome to "click twice" for what used to be a single action (click home page, wait for reload, then access "login or register).
2. Users likely expect that "logging out" removes any personal info from being displayed further. Yet the GUI is exposing the logged-out-account's email address.

How to reproduce:

**Before**: after logging out, the prior user was completely anonymous in the GUI. The login or register menu was made available in the masthead, with no "User menu".

**Now**: after logging out, the prior user's email is still visible in the GUI under the "User" menu. Actions on history menu stall out, sometimes exposing the last history the prior user had loaded. One needs to click on the link to "go to the homepage" to completely reset the GUI. 

Screenshot for **Now**

<img width="958" alt="logout_gui" src="https://user-images.githubusercontent.com/8999735/51499514-70153180-1d7f-11e9-9c99-3b992d111b12.png">

Nerdinacan:
Happy to take a peek!

Nerdinacan:
This is ready to go, soon as we fix the problems with the current dev build I'll put up a PR.

martenson:
xref https://github.com/galaxyproject/galaxy/issues/6853
https://github.com/galaxyproject/galaxy/issues/7194

Nerdinacan:
Associated PR:
https://github.com/galaxyproject/galaxy/pull/7255

dannon:
Fixed in #7255

jennaj:
I don't think this is quite working as expected. I noticed this earlier, sorry didn't write back sooner!

Logout leaves the prior user's quota usage displayed, test with both admin and non-admin accounts at https://usegalaxy.org:

<img width="1145" alt="screen shot 2019-02-28 at 5 25 10 pm" src="https://user-images.githubusercontent.com/8999735/53609926-f0e4fb80-3b7d-11e9-813b-8453a2a7cd14.png">


I expected logout to take me all the way back here, without needing to click on anything else (how it used to work):

<img width="1010" alt="screen shot 2019-02-28 at 5 21 07 pm" src="https://user-images.githubusercontent.com/8999735/53609822-8df36480-3b7d-11e9-990c-dd1693ece709.png">

jennaj:
[...]

PR #7255, opened 2019-01-22
Changed logout behavior to automatically redirect home

Nerdinacan:
Modifies logout to wipe local session data first, then redirect to home page following logout redirect.

A cleaner solution would be to implement a logout API endpoint and do a promise-based ajax request that clears the local session after the API call was successful, then redirects home, but this will fix the immediate issue.

Addresses:
https://github.com/galaxyproject/galaxy/issues/7249


jmchilton (approved):

dannon:
This works well for me.  Ran related tests manually, looked good.

PR #7256, opened 2019-01-22
Fix hotdata method access; 

dannon:
this still needs a lot of love/refactoring (in general, after this PR) but this makes it work again, I think, and without a potential caching issue?  (opening to rerun selenium tests...)

xref #7251

### DIFF
--- a/client/galaxy/scripts/components/RuleCollectionBuilder.vue
+++ b/client/galaxy/scripts/components/RuleCollectionBuilder.vue
@@ -1385,7 +1385,7 @@ export default {
                 data = sources.map(el => []);
                 columns = [];
             } else if (this.elementsType == "collection_contents") {
-                const collection = this.initialElements.slice();
+                const collection = this.initialElements;
                 if (collection) {
                     const obj = this.populateElementsFromCollectionDescription(
                         collection.elements,
--- a/client/galaxy/scripts/components/RuleCollectionBuilder.vue
+++ b/client/galaxy/scripts/components/RuleCollectionBuilder.vue
@@ -1257,8 +1257,9 @@ export default {
             return targets;
         },
         colHeaders() {
-            const data = this.hotData["data"];
-            const columns = this.hotData["columns"];
+            let hotData = this.hotData();
+            const data = hotData["data"];
+            const columns = hotData["columns"];
             return RuleDefs.colHeadersFor(data, columns);
         },
         colHeadersDisplay() {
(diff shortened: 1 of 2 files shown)


Nerdinacan:
That's basically what I did. There's a place in the template you're going to have to update as well

```
<hot-table
                        id="hot-table"
                        ref="hotTable"
                        :data="hotData['data']"
                        :colHeaders="colHeadersDisplay"
                        :readOnly="true"
                        stretchH="all"
                    >
```

Nerdinacan:
So I took a careful look at this, and there was indeed a side-effect in the hotData computed, but it was completely unnecessary. Some processing was being bundled in an imported module in a function applyRules which mutated a passed array, triggering the eslint warning.

Looking carefully at what happens in that applyRules function, there was no real reason for it to be mutating the passed parameter, so I just made it an immutable input and generated a new computed value from the result of that processing and I think we're good.

client/galaxy/scripts/components/RuleCollectionBuilder.vue
```diff
@@ -1029,7 +1029,7 @@ export default {
         }
         return {
             rules: rules,
-            colHeadersPerRule: [],
+            // colHeadersPerRule: [],
```
guerler: We probably should just remove this line.

Nerdinacan: Yup

dannon:
@Nerdinacan Thanks for taking a look; I hadn't had a chance to dig deeper and figure out if the side effect (resetting `this.colHeadersPerRule` on access) was intentional and necessary or not.  Good to know it's not!

Nerdinacan:
@dannon Yeah I got all the tests passing locally, but the entire selenium suite looks like it's not running at all. Not sure why yet.

mvdbeek:
I've fixed the test setup in https://github.com/galaxyproject/galaxy/pull/7278, problem was an incompatible version of node in the client builder image.

nsoranzo (approved):
With this PR, Selenium test errors go down from 26 to 13, with no new error.

nsoranzo:
The failed API test should be fixed by https://github.com/galaxyproject/galaxy/pull/7279 .

dannon:
I'm checking the remaining failures manually and they're all passing on my local machine, looking for other configuration differences that'd explain it.

In any event, this is ready to merge.

mvdbeek:
I see the same errors locally, this one for instance is also in the selenium logs:
screen shot 2019-01-31 at 15 39 21

dannon:
@mvdbeek The last 13?  Yeah, I reproduced them yesterday in a virtual machine, working on fixes this morning.  Out of curiosity, what's your environment that reproduced them?  I couldn't get it to happen on my macbook.

mvdbeek:
So weird, I tried the same thing just now again and it worked without error.

dannon:
@mvdbeek I think it's a render timing thing, and fast machines won't reproduce it.  If I bump up the waits, tests pass, but I don't really like doing that so I'm trying a few more things.

PR #7258, opened 2019-01-23
[18.09] Not all workflow outputs have a UUID

mvdbeek:
(Majorly!) Broken in https://github.com/galaxyproject/galaxy/pull/7227

### DIFF
--- a/lib/galaxy/workflow/modules.py
+++ b/lib/galaxy/workflow/modules.py
@@ -353,8 +353,10 @@ class SubWorkflowModule(WorkflowModule):
                     label = workflow_output['label']
                     if not label:
                         label = "%s:%s" % (order_index, workflow_output['output_name'])
+                    workflow_output_uuid = workflow_output.get('uuid') or object()
                     for data_output in data_outputs:
-                        if data_output['name'] == workflow_output['output_name'] or data_output.get('uuid', object()) == workflow_output.get('uuid', object()):
+                        data_output_uuid = data_output.get('uuid') or object()
+                        if data_output['name'] == workflow_output['output_name'] or data_output_uuid == workflow_output_uuid:
                             data_output['label'] = label
                             data_output['name'] = label
                             # That's the right data_output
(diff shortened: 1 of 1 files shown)


nsoranzo:
Was it failing with `KeyError`?

Can either `data_output['uuid']` or `workflow_output['uuid']` be `''` or `None` ?

jmchilton (approved):

mvdbeek:
> Can either `data_output['uuid']` or `workflow_output['uuid']` be `''` or `None` ?

Yes, good point!

mvdbeek:
(https://github.com/galaxyproject/galaxy/blob/c228760e1f067111fff0ecf9cf69ed3b019313a2/lib/galaxy/managers/workflows.py#L715)

nsoranzo:
Which means that:
```
data_output.get('uuid', object()) == workflow_output.get('uuid', object())
```
can become:
```
None == None
```
which is `True`.

mvdbeek:
Thanks @nsoranzo, I think the last commit should fix that.

lib/galaxy/workflow/modules.py
```diff
@@ -353,8 +353,11 @@ def get_data_outputs(self):
                     label = workflow_output['label']
                     if not label:
                         label = "%s:%s" % (order_index, workflow_output['output_name'])
+                    workflow_output_uuid = object()
+                    if workflow_output.get('uuid'):
+                        workflow_output_uuid = workflow_output['uuid']
```
nsoranzo: You can do the same in one line:
```python
workflow_output_uuid = workflow_output.get('uuid') or object()
```
For consistency, I'd also define here `data_output_uuid` in a similar way.

mvdbeek: Thanks, I've rebase the commit with your suggestions.

lib/galaxy/workflow/modules.py
```diff
@@ -353,8 +353,11 @@ def get_data_outputs(self):
                     label = workflow_output['label']
                     if not label:
                         label = "%s:%s" % (order_index, workflow_output['output_name'])
+                    workflow_output_uuid = object()
```
nsoranzo: This line is not needed anymore.

mvdbeek: Of couse 🤦‍♂️ ..., thanks again!

nsoranzo (approved):

Issue #7261, opened 2019-01-23
Saving imported subworkflow (from different instance) fails


mvdbeek:
```
Traceback (most recent call last):
  File "/Users/mvandenb/src/galaxy/lib/galaxy/web/framework/decorators.py", line 283, in decorator
    rval = func(self, trans, *args, **kwargs)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/webapps/galaxy/api/workflows.py", line 545, in update
    **from_dict_kwds
  File "/Users/mvandenb/src/galaxy/lib/galaxy/managers/workflows.py", line 362, in update_workflow_from_raw_description
    **kwds
  File "/Users/mvandenb/src/galaxy/lib/galaxy/managers/workflows.py", line 417, in _workflow_from_raw_description
    self.__load_subworkflows(trans, step_dict, subworkflow_id_map, **kwds)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/managers/workflows.py", line 1069, in __load_subworkflows
    trans, step_dict, subworkflow_id_map, **kwds
  File "/Users/mvandenb/src/galaxy/lib/galaxy/managers/workflows.py", line 1151, in __load_subworkflow_from_step_dict
    trans, subworkflow_id
  File "/Users/mvandenb/src/galaxy/lib/galaxy/managers/workflows.py", line 105, in get_owned_workflow
    self.check_security(trans, workflow, check_ownership=True)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/managers/workflows.py", line 128, in check_security
    stored_workflow = has_workflow.top_level_stored_workflow
  File "/Users/mvandenb/src/galaxy/lib/galaxy/model/__init__.py", line 4063, in top_level_stored_workflow
    return self.top_level_workflow.stored_workflow
  File "/Users/mvandenb/src/galaxy/lib/galaxy/model/__init__.py", line 4052, in top_level_workflow
    assert len(self.parent_workflow_steps) == 1
AssertionError
```

Issue #6032, opened 2018-05-01
When attempting to edit a workflow with subworkflows: `Uncaught exception in exposed API method:`


lparsons:
```galaxy.web.framework.decorators ERROR 2018-05-01 16:56:04,487 Uncaught exception in exposed API method:                                                                                                                                                                            
Traceback (most recent call last):                                                                                                                                                                                                                                                 
  File "/galaxy/galaxy-app/lib/galaxy/web/framework/decorators.py", line 281, in decorator                                                                                                                                                                                         
    rval = func(self, trans, *args, **kwargs)                                                                                                                                                                                                                                      
  File "/galaxy/galaxy-app/lib/galaxy/webapps/galaxy/api/workflows.py", line 469, in update                                                                                                                                                                                        
    workflow_dict,                                                                                                                                                                                                                                                                 
  File "/galaxy/galaxy-app/lib/galaxy/managers/workflows.py", line 267, in update_workflow_from_dict                                                                                                                                                                               
    name=stored_workflow.name,                                                                                                                                                                                                                                                     
  File "/galaxy/galaxy-app/lib/galaxy/managers/workflows.py", line 310, in _workflow_from_dict                                                                                                                                                                                     
    self.__load_subworkflows(trans, step_dict)                                                                                                                                                                                                                                     
  File "/galaxy/galaxy-app/lib/galaxy/managers/workflows.py", line 833, in __load_subworkflows                                                                                                                                                                                     
    trans, step_dict                                                                                                                                                                                                                                                               
  File "/galaxy/galaxy-app/lib/galaxy/managers/workflows.py", line 905, in __load_subworkflow_from_step_dict                                                                                                                                                                       
    trans, subworkflow_id                                                                                                                                                                                                                                                          
  File "/galaxy/galaxy-app/lib/galaxy/managers/workflows.py", line 84, in get_owned_workflow                                             
    self.check_security(trans, workflow, check_ownership=True)      
  File "/galaxy/galaxy-app/lib/galaxy/managers/workflows.py", line 107, in check_security                                                
    stored_workflow = has_workflow.top_level_stored_workflow        
  File "/galaxy/galaxy-app/lib/galaxy/model/__init__.py", line 3776, in top_level_stored_workflow                                        
    return self.top_level_workflow.stored_workflow                  
  File "/galaxy/galaxy-app/lib/galaxy/model/__init__.py", line 3765, in top_level_workflow                                               
    assert len(self.parent_workflow_steps) == 1                     
[...]

PR #7262, opened 2019-01-23
Fix saving imported subworkflows

mvdbeek:
whose child-workflow is included more than once.
Fixes https://github.com/galaxyproject/galaxy/issues/7261 and probably https://github.com/galaxyproject/galaxy/issues/6032


nsoranzo (approved):

mvdbeek:
Should we backport this ? This was probably an issue from the start, and https://github.com/galaxyproject/galaxy/issues/6032 dates back a while.

PR #7270, opened 2019-01-25
Tagging component

Nerdinacan:
Tagging component, unit testing examples

**templates/tagging_common.mako**
This is a replacement for the code that ran in the old tagging_common.mako. It is a simple Vue wrapper around a tagging component with temporary housings for the community and individual tag scenarios that were previously rendered to the page by python and run by Backbone.

**components/Tags**
I've tried to flesh out concrete examples of how to unit test and mock imported service modules as well as the Vue components themselves. I've illustrated a dependency mocking technique which allows us to write fast unit tests by mocking out low-level dependencies such as axios.

**grid-view.js**
I've introduced a mechanism that demonstrates how to couple the Vuex store (which is the central state housing for the future application) with pre-existing backbone views via an exposed .watch() method. Essentially this allows backbone to subscribe to changes in the Vuex store as if it were an observable. The goal is to allow our new components to continue to affect state in the backbone views until such time as the backbone views are phased-out.

I'm still fiddling with the CSS and double-checking all the various scenarios where the component gets mounted. Several of the old parameters to the python rendering functions have been removed as they were either doing nothing, or doing something that should have been handled internally by the client-side rendering code.

If this works out, the next thing I'll do will be to replace select2 with this same component.


Nerdinacan:
All those failed selenium tests are passing locally. Still need to figure out why the selenium tests are failing only on jenkins.

templates/tagging_common.mako
```diff
@@ -1,259 +1,124 @@
...
-    from random import random
-    from math import floor
-    import six
-    from galaxy.model import Tag, ItemTagAssociation
-    from galaxy.web.framework.helpers import iff
-    from galaxy.util import unicodify
+
+from cgi import escape
+import six
+## from galaxy.model import Tag, ItemTagAssociation
```
nsoranzo: Remove?

Nerdinacan: Yeah, looks unused.

dannon:
@Nerdinacan I'll check this out, thanks for the ping.  I reproduced those failing selenium tests in a local VM and I think it's just a testing wait/timing issue that I hope to get a fix in for shortly, as well.

Nerdinacan:
@dannon Awesome, I'm glad somebody could reproduce them. Let me know if you need anything.

Nerdinacan:
Awww yeah, this is finally ready if you guys have no beef with the way I built it. Thanks @dannon for fixing those broken selenium tests!

dannon:
@Nerdinacan Cool, I'll review it soon (probably tomorrow morning)!

dannon:
@Nerdinacan Sorry to leave this in the lurch for a bit.  The community tagging portions aren't working for me, haven't dug into why yet, but that's on my list for today.

Nerdinacan:
> @Nerdinacan Sorry to leave this in the lurch for a bit. The community tagging portions aren't working for me, haven't dug into why yet, but that's on my list for today.

@dannon Ahhh ok. I'll take a peek too. This PR was only intended to remove tagging done by .mako files, there's another PR that's ready that takes care of the select2 instances elsewhere in the app. Let me check the community thingie.

UPDATE: Ok, I see what was wrong. Community tags should have been display-only. Also there was a bug on the Pages tagging interface that forced me to look more carefully at the way in which the tag data was being delivered to the components, and I've updated the .mako functions to process the incoming data appropriately.

PR #7272, opened 2019-01-27
Add Azure and GCP support for ObjectStore Cloud backend 

VJalili:


lib/galaxy/objectstore/cloud.py
```diff
@@ -92,15 +89,121 @@ def _initialize(self):
...
+
+            if len(missing_config) > 0:
+                msg = "The following configuration required for {} cloud backend " \
+                      "are missing: {}".format(provider, missing_config)
+                log.error(msg)
+                raise Exception(msg)
+            else:
+                return config
+        except Exception:
+            log.exception("TMalformed ObjectStore Configuration XML -- unable to continue")
```
jmchilton: TMalformed

jmchilton:
Can you add some more serialization tests for the new stuff (https://github.com/galaxyproject/galaxy/blob/dev/test/unit/test_objectstore.py#L543), test coverage for configuring these things is getting better I think we should not backtrack here.

VJalili:
Agree, I am working on the tests now. First updating to adhere with new changes, and will then extend to assert new vendors.

jmchilton (approved):

config/object_store_conf.xml.sample
```diff
@@ -46,22 +46,44 @@
...
+        -->
+
+        <!-- Cloud ObjectStore: Google Compute Engine (GCE) -->
+        <!--
+        <object_store type="cloud" provider="google" order="0">
+            <auth credentials_file="..." />
+            <bucket name="..." use_reduced_redundancy="False" />
+            <cache path="database/object_store_cache" size="1000" /> <!-- size is in gigabytes -->
+            <extra_dir type="job_work" path="database/job_working_directory_gce"/>
+            <extra_dir type="temp" path="database/tmp_gce"/>
```
afgane: Perhaps better to go with the Google Cloud **Platform** (GCP) terminology vs. GCE so it's consistent with CloudBridge (and Google). xref https://github.com/CloudVE/cloudbridge/pull/189

afgane (approved):
LGTM with just the suggestion or renaming GCE to GCP throughout the PR.

config/object_store_conf.xml.sample
VJalili: Done.

VJalili:
@afgane Thank!

Issue #7273, opened 2019-01-27
Tool migration checks at startup do not read the correct Galaxy config file when not using PasteDeploy


natefoo:
The Galaxy config file to read for determining the `database_connection` for tool migration is pulled from the `global_conf` kwarg (it's a dict) key `__file__` in `app.py`, but `global_conf` is automatically passed by PasteDeploy and does not exist under uWSGI. So the tool migration check is working from the default config and attempting to open `database/universe.sqlite`, which does not exist and it cannot create when running with privilege separation.

### Example of how to reproduce

- Galaxy owned by root in `/srv/galaxy/server`
- Galaxy config owned by root in `/srv/galaxy/config/galaxy.yml`
- Start Galaxy for the first time in cwd `/srv/galaxy/server` with `uwsgi --yaml ../config/galaxy.yml`

### Workarounds

- Disable the `check_migrate_tools` Galaxy config opt
- Maybe (untested, but no `global_conf` is I believe why I added these options) in the `uwsgi` section of `galaxy.yml`:
    ```yaml
    uwsgi:
      set: galaxy_config_file=/srv/galaxy/config/galaxy.yml
      set: galaxy_root=/srv/galaxy/server
    ```
- Maybe (untested, should create `database/universe.sqlite` and satisfy the bug though):
    ```console
    sudo -iu <privsep user>
    cd /srv/galaxy/server
    . <path to venv>/bin/activate
    sh create_db.sh
    sh manage_db.sh upgrade tools'`
    ```

In addition to fixing the bug, I propose we disable `check_migrate_tools` by default in 19.01, this is code that nobody needs to run at this point and as far as I know we have no more tool migrations planned.

### Traceback

```pytb
Traceback (most recent call last):
  File "/srv/galaxy/server/lib/galaxy/webapps/galaxy/buildapp.py", line 49, in app_factory
    app = galaxy.app.UniverseApplication(global_conf=global_conf, **kwargs)
  File "/srv/galaxy/server/lib/galaxy/app.py", line 86, in __init__
    self._configure_models(check_migrate_databases=True, check_migrate_tools=check_migrate_tools, config_file=config_file)
  File "/srv/galaxy/server/lib/galaxy/config.py", line 1143, in _configure_models
    verify_tools(self, install_db_url, config_file, install_database_options)
  File "/srv/galaxy/server/lib/tool_shed/galaxy_install/migrate/check.py", line 65, in verify_tools
[...]

PR #7275, opened 2019-01-27
[18.09] Stop tool migrations checking by default...

jmchilton:
They are not needed in modern Galaxies and even problematic in some ways (e.g. #7273 and having the possibility of slowing Galaxy down at startup).


mvdbeek (approved):

davebx (approved):

davebx:
Thanks, @jmchilton. I should have realized this was the cause of @erasche's migration issues the other week.

PR #7278, opened 2019-01-29
Update node for selenium tests

mvdbeek:
AFAICT this is at least part of the problem:
```
yarn install v1.3.2
[1/4] Resolving packages...
[2/4] Fetching packages...
info fsevents@1.2.4: The platform "linux" is incompatible with this module.
info "fsevents@1.2.4" is an optional dependency and failed compatibility check. Excluding it from installation.
error eslint@5.12.0: The engine "node" is incompatible with this module. Expected version "^6.14.0 || ^8.10.0 || >=9.10.0".
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
Makefile:144: recipe for target 'node-deps' failed
make: *** [node-deps] Error 1
```

### DIFF
--- a/.ci/jenkins/selenium-py3/run_tests.sh
+++ b/.ci/jenkins/selenium-py3/run_tests.sh
@@ -2,7 +2,7 @@
 
 # Enable retries on tests to reduce chances of transient failures.
 : ${GALAXY_TEST_SELENIUM_RETRIES:=1}
-: ${GALAXY_TEST_CLIENT_BUILD_IMAGE:='node:9.4.0'}
+: ${GALAXY_TEST_CLIENT_BUILD_IMAGE:='node:11'}
 
 # If in Jenkins environment, use it for artifacts.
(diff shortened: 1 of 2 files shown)


mvdbeek:
Yay, that was it, it seems: https://jenkins.galaxyproject.org/job/docker-selenium/4932/console

mvdbeek:
The remaining selenium failures are probably bugs which should to be fixed separately. I'm guessing https://github.com/galaxyproject/galaxy/pull/7256 might fix some of those.

nsoranzo:
Thanks @mvdbeek for finding what was the issue! At the moment we are using `node` 10.13.0 in `scripts/common_startup.sh` (not 11). I think having to remember to sync the node version in 2 places is going to be a recurring problem, so I'm trying a different fix in https://github.com/galaxyproject/galaxy/pull/7280.

PR #7279, opened 2019-01-29
Fix flakey history import metadata test

mvdbeek:
We need to wait for the BAM dataset to be final before we can export.
(At least that solved it locally for me).


jmchilton:
Thanks so much you two!

PR #7280, opened 2019-01-29
Do not skip the client build in dockerized selenium tests

nsoranzo:
This allows the removal of the `docker run` of `make
client-production-maps`, which was run on a docker image for the
wrong version of `node`.

Alternative to https://github.com/galaxyproject/galaxy/pull/7278 with the advantage of not having to remember to sync the node version in 2 places.

### DIFF
--- a/.ci/jenkins/selenium-py3/run_tests.sh
+++ b/.ci/jenkins/selenium-py3/run_tests.sh
@@ -2,7 +2,6 @@
 
 # Enable retries on tests to reduce chances of transient failures.
 : ${GALAXY_TEST_SELENIUM_RETRIES:=1}
-: ${GALAXY_TEST_CLIENT_BUILD_IMAGE:='node:9.4.0'}
 
 # If in Jenkins environment, use it for artifacts.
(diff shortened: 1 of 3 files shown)


mvdbeek (approved):

nsoranzo:
@galaxybot test py3

dannon:
@mvdbeek @nsoranzo Argh, good catch -- thank you both.  I was chasing a red herring looking at the chromedriver startup failures seen in the logs!  I had completely forgotten that we set up this separate client-docker-build pipeline for selenium (we needed it in the days before Galaxy managed its own node/client stuff), and we do not need it anymore -- and it's best to test against the single pinned version specified in the galaxy common startup bit anyway.

nsoranzo:
Same number of failed Selenium tests here and in #7278, should be good to go.

dannon:
@nsoranzo Yeah, that's what I was just looking at.  Many of those pass locally for me, but this is a clear improvement and I can follow up looking into that separately.  Going to go ahead and merge.

PR #7287, opened 2019-01-31
Fix remaining selenium tests and add headless testing option.

dannon:
This was a race condition with the new logout redirect, I believe.  On fast machines everything works fine because the logout click was happening before the redirect kicked in, so the tests all passed.  On jenkins or on my intentionally slow testing VM, I was able to reproduce the delayed click trying to happen after/during redirect.

I'm going to lump a headless selenium testing option that was very useful for me while testing/reproducing (for at least chrome) into this PR as well.  At least one more commit coming to make it configurable, off by default, and maybe try to get geckodriver doing it as well.


nsoranzo:
Seems to work! :tada:

The first 2 failing tests passed.

guerler:
Is this good to go? Thanks @dannon

dannon:
@guerler Not quite; I went down the rabbit hole trying to get firefox/geckodriver working to set up (native) headless compatibility for that but it seems to be a mess with firefox 48+ and marionette or whatever.  I'm going to push a tweak for chrome's headless mode to make it configurable and write some documentation and then it'll be ready.  I'll switch it out of WIP then.

mvdbeek (approved):
Awesome! Thanks @dannon !

dannon:
Thanks for the merge.  @guerler your chrome-based selenium testing should now be natively headless, too, by default without the extra display requirements.

guerler:
Thanks @dannon.

PR #7290, opened 2019-01-31
Add a method to assert if user can assume an authz record with given id.

VJalili:


VJalili:
One can do same assertion using `try_get_authz_config`.

PR #7294, opened 2019-02-01
[20.01] Fix S3 issue discovered during admin training

hexylena:
Keys are being generated with `/...` but it seems to work fine. Galaxy even throws an error for old style exceptions here.

Discovered by @sanman-99

### DIFF
--- a/lib/galaxy/objectstore/s3.py
+++ b/lib/galaxy/objectstore/s3.py
@@ -378,8 +378,6 @@ class S3ObjectStore(ObjectStore, CloudConfigMixin):
         except S3ResponseError:
             log.exception("Trouble checking existence of S3 key '%s'", rel_path)
             return False
-        if rel_path[0] == '/':
-            raise
         return exists
 
     def _in_cache(self, rel_path):
--- a/lib/galaxy/objectstore/s3.py
+++ b/lib/galaxy/objectstore/s3.py
@@ -123,9 +123,10 @@ class CloudConfigMixin(object):
                 'conn_path': self.conn_path,
             },
             'cache': {
-                'size': -1,  # disable cache cleaning when starting an exported object store in set_metadata.py
+                'size': self.cache_size,
                 'path': self.staging_path,
-            }
+            },
+            'enable_cache_monitor': False,
         }
 
 
@@ -146,6 +147,7 @@ class S3ObjectStore(ObjectStore, CloudConfigMixin):
         bucket_dict = config_dict['bucket']
         connection_dict = config_dict.get('connection', {})
         cache_dict = config_dict['cache']
+        self.enable_cache_monitor = config_dict.get('enable_cache_monitor', True)
 
         self.access_key = auth_dict.get('access_key')
         self.secret_key = auth_dict.get('secret_key')
@@ -190,8 +192,16 @@ class S3ObjectStore(ObjectStore, CloudConfigMixin):
 
         self._configure_connection()
         self._bucket = self._get_bucket(self.bucket)
+        self.start_cache_monitor()
+        # Test if 'axel' is available for parallel download and pull the key into cache
+        if which('axel'):
+            self.use_axel = True
+        else:
+            self.use_axel = False
+
+    def start_cache_monitor(self):
         # Clean cache only if value is set in galaxy.ini
-        if self.cache_size != -1:
+        if self.cache_size != -1 and self.enable_cache_monitor:
             # Convert GBs to bytes for comparison
             self.cache_size = self.cache_size * 1073741824
             # Helper for interruptable sleep
@@ -199,11 +209,6 @@ class S3ObjectStore(ObjectStore, CloudConfigMixin):
             self.cache_monitor_thread = threading.Thread(target=self.__cache_monitor)
             self.cache_monitor_thread.start()
             log.info("Cache cleaner manager started")
-        # Test if 'axel' is available for parallel download and pull the key into cache
-        if which('axel'):
-            self.use_axel = True
-        else:
-            self.use_axel = False
 
     def _configure_connection(self):
         log.debug("Configuring S3 Connection")
--- a/lib/galaxy/objectstore/s3.py
+++ b/lib/galaxy/objectstore/s3.py
@@ -189,7 +189,7 @@ class S3ObjectStore(ObjectStore, CloudConfigMixin):
                          'conn_path': self.conn_path}
 
         self._configure_connection()
-        self.__bucket = self._get_bucket(self.bucket)
+        self._bucket = self._get_bucket(self.bucket)
         # Clean cache only if value is set in galaxy.ini
         if self.cache_size != -1:
             # Convert GBs to bytes for comparison
@@ -356,7 +356,7 @@ class S3ObjectStore(ObjectStore, CloudConfigMixin):
 
     def _get_size_in_s3(self, rel_path):
         try:
-            key = self.__bucket.get_key(rel_path)
+            key = self._bucket.get_key(rel_path)
             if key:
                 return key.size
         except S3ResponseError:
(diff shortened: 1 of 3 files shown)


VJalili:
Does it also affect `cloud` backend?
 
https://github.com/galaxyproject/galaxy/blob/645031578d31801eee84ac79013d2da358212078/lib/galaxy/objectstore/cloud.py#L270-L271

hexylena:
Probably, based on your snippet.

VJalili:
Any idea on how I can reproduce this issue?

hexylena:
@sanman-99 could you provide the configuration which you got working?

sanman-99:
The working config involves commenting out the 'raise' that was in the conditional statement "if rel_path[0] == '/'" in the _key_exists function.  Should be at line number 381 in /lib/galaxy/objectstore/s3.py.  Let me know if you need more detail.

```python
def _key_exists(self, rel_path):
        exists = False
        try:
            # A hackish way of testing if the rel_path is a folder vs a file
            is_dir = rel_path[-1] == '/'
            if is_dir:
                keyresult = self.bucket.get_all_keys(prefix=rel_path)
                if len(keyresult) > 0:
                    exists = True
                else:
                    exists = False
            else:
                key = Key(self.bucket, rel_path)
                exists = key.exists()
        except S3ResponseError:
            log.exception("Trouble checking existence of S3 key '%s'", rel_path)
            return False
        #if rel_path[0] == '/':
            #raise
        return exists
```

If these lines are kept in, when I run a job with S3 as an object store, I get the following error:

```

  File "/srv/galaxy/server/lib/galaxy/jobs/handler.py", line 319, in __monitor_step 

    self.dispatcher.put(self.job_wrappers.pop(job.id)) 

  File "/srv/galaxy/server/lib/galaxy/jobs/handler.py", line 920, in put 

    self.job_runners[runner_name].put(job_wrapper) 

  File "/srv/galaxy/server/lib/galaxy/jobs/runners/__init__.py", line 122, in put 

    job_wrapper.enqueue() 

  File "/srv/galaxy/server/lib/galaxy/jobs/__init__.py", line 1197, in enqueue 

    self._set_object_store_ids(job) 

  File "/srv/galaxy/server/lib/galaxy/jobs/__init__.py", line 1222, in _set_object_store_ids 

    self._setup_working_directory(job=job) 

  File "/srv/galaxy/server/lib/galaxy/jobs/__init__.py", line 904, in _setup_working_directory 

    working_directory = self._create_working_directory(job) 

  File "/srv/galaxy/server/lib/galaxy/jobs/__init__.py", line 937, in _create_working_directory 

    job, base_dir='job_work', dir_only=True, obj_dir=True) 

  File "/srv/galaxy/server/lib/galaxy/objectstore/__init__.py", line 712, in create 

    self.backends[0].create(obj, **kwargs) 

  File "/srv/galaxy/server/lib/galaxy/objectstore/s3.py", line 441, in create 

    if not self.exists(obj, **kwargs): 

  File "/srv/galaxy/server/lib/galaxy/objectstore/s3.py", line 415, in exists 

    in_s3 = self._key_exists(rel_path) 

  File "/srv/galaxy/server/lib/galaxy/objectstore/s3.py", line 282, in _key_exists 

    raise 

TypeError: exceptions must be old-style classes or derived from BaseException, not NoneType

```

hexylena:
@sanman-99 could you share the configuration from the `object_store_conf.xml`? (with access/secret redacted of course)

sanman-99:
The configuration is:

<?xml version="1.0"?>
<object_store type="hierarchical">
    <backends>
        <object_store type="s3" id="primary" order="0">
             <auth access_key="REMOVED" secret_key="REMOVED" />
             <bucket name="cfsan-galaxy-objectstore-poc" use_reduced_redundancy="False" />
             <cache path="/srv/galaxy/var/cache/object_store_cache" size="1" />
             <extra_dir type="job_work" path="/data2/job_work_dir2"/>
             <extra_dir type="temp" path="/tmp"/>
        </object_store>
        <object_store type="disk" id="olddata" order="1">
             <files_dir path="/data"/>
             <extra_dir type="job_work" path="/data/job_work_dir"/>
        </object_store>
    </backends>
</object_store>

sanman-99:
`<?xml version="1.0"?>
<object_store type="hierarchical">
    <backends>
        <object_store type="s3" id="primary" order="0">
             <auth access_key="REMOVED" secret_key="REMOVED" />
             <bucket name="cfsan-galaxy-objectstore-poc" use_reduced_redundancy="False" />
             <cache path="/srv/galaxy/var/cache/object_store_cache" size="1" />
             <extra_dir type="job_work" path="/data2/job_work_dir2"/>
             <extra_dir type="temp" path="/tmp"/>
        </object_store>
        <object_store type="disk" id="olddata" order="1">
             <files_dir path="/data"/>
             <extra_dir type="job_work" path="/data/job_work_dir"/>
        </object_store>
    </backends>
</object_store>`

sanman-99:
```xml
<object_store type="hierarchical">
    <backends>
        <object_store type="s3" id="primary" order="0">
             <auth access_key="REMOVED" secret_key="REMOVED" />
             <bucket name="cfsan-galaxy-objectstore-poc" use_reduced_redundancy="False" />
             <cache path="/srv/galaxy/var/cache/object_store_cache" size="1" />
             <extra_dir type="job_work" path="/data2/job_work_dir2"/>
             <extra_dir type="temp" path="/tmp"/>
        </object_store>
        <object_store type="disk" id="olddata" order="1">
             <files_dir path="/data"/>
             <extra_dir type="job_work" path="/data/job_work_dir"/>
        </object_store>
    </backends>
</object_store>
```

VJalili:
I am not able to reproduce this issue. I am testing on a fresh branch (i.e., latest dev) with the same configuration as your provided, and as shown in the following figure, Galaxy successfully uploads datasets to S3. 

screen shot 2019-02-05 at 4 34 05 pm

Any thoughts on how I can reproduce the issue? 

Few points worth noting:
- **always** use one bucket per Galaxy instance, never point two different Galaxy instances to use a common bucket;
- if you have used a bucket on Galaxy instance, then deleted the database of that instance, and want started over, then make sure you have emptied the bucket before starting the Galaxy, or use a different bucket. 
- do not modify Galaxy's bucket; e.g., do not try to rename `000/dataset_*.dat` structure.

(These points generally apply to all the paths Galaxy persists data at.)

sanman-99:
Thanks, I will try again on a fresh Galaxy install using the latest and a new bucket to see if I am able to get it to work.

Takadonet:
Having same issue as well on a fresh Galaxy on 18.09 branch.
Stack-trace : 
https://gist.github.com/Takadonet/aac97d08f7e23dc13e7676791e8ee8d4

nsoranzo:
I suppose that's due to having the path starting with `/` in `extra_dir` elements, e.g. `<extra_dir type="job_work" path="/data2/job_work_dir2"/>` .

sanman-99:
ok, thanks that did the trick.  So I assume then absolute paths cannot be used for these parameters?

VJalili:
Can we close this PR then?

hexylena:
I'd rather not close this PR, it should be reworked to support absolute paths, it's an issue that Galaxy can't support that. Additionally it's an issue that there's still an old style exception raised there, that suppresses the real error message.

mvdbeek:
Is there any harm in merging the PR as is ? It did come up again in https://github.com/galaxyproject/galaxy/issues/7649

nsoranzo:
I don't think the code is supposed to work with absolute paths at the moment, so removing this check could be dangerous. But I'm not an object store expert, ping @dannon

lib/galaxy/objectstore/s3.py
```diff
@@ -187,7 +189,7 @@ def _initialize(self):
                          'conn_path': self.conn_path}
 
         self._configure_connection()
-        self.bucket = self._get_bucket(self.bucket)
+        self.__bucket = self._get_bucket(self.bucket)
```
nsoranzo: Is the use of name mangling due to [double leading underscore](https://dbader.org/blog/meaning-of-underscores-in-python) intentional? I think it may break `SwiftObjectStore` which inherits from this class.

mvdbeek: We are testing `SwiftObjectStore` so that should be fine. But single underscore will be fine too.

nsoranzo:
py27-unit test failure is unrelated

hexylena:
:tada:

PR #7296, opened 2019-02-01
Add datatype for wiff_archive for AB SCIEX mass spec data

jj-umn:


lib/galaxy/datatypes/binary.py
```diff
@@ -2013,6 +2013,69 @@ def display_peek(self, dataset):
...
+        False
+        """
+        try:
+            if filename and zipfile.is_zipfile(filename):
+                with zipfile.ZipFile(filename, 'r') as tempzip:
+                    for name in tempzip.namelist():
+                        if name.lower().endswith('.wiff'):
+                            return True
+                return False
+        except Exception as e:
```
mvdbeek, line 2060: It's not necessary to do this, exceptions are caught automatically, and I don't think we want to see failed sniffing attempts in the logs, as that's a very normal thing.

mvdbeek (commented):
Looks good, can you drop a small test file ending in `.wiff_archive` into lib/galaxy/datatypes/test ?

chambm:
What does `AB SCIEX archive` mean? Has Sciex started officially exporting zip files of its wiff/scan pairs and calling them archives? [I am working on a similar approach using tar for directory-based sources](https://github.com/galaxyproject/galaxy/compare/dev...chambm:feature/rawtar-datatypes) like Bruker, Agilent, and Waters use (also working on the msconvert on Wine/Docker changes to go with it).

Assuming this archive format is not an official thing, I figure WIFF/SCAN pairs can go here too. I chose TAR (no gz/bz/xz) instead of ZIP because some testing on Jetstream showed that I could extract a 4 gigabyte tar file in less than 2 seconds, whereas the same directory in a ZIP file (store-mode only, no compression) took over a minute to extract. I was unable to reproduce this difference on AWS, where the ZIP and TAR extracted in essentially the same amount of time. Both formats share the same main disadvantages:
1. How do users select a bunch of wiff/scan pairs or directory-based sources and create the raw-tars (or raw-zips) for them in order to upload them to Galaxy?
2. Unless the raw-tars/raw-zips are created, uploaded, and deleted one at a time, it takes up twice the amount of space to upload data.

Some nice GUI is needed that lets users easily select directory-based sources and WIFF files to upload, and it could perhaps even tar on-the-fly when uploading rather than having to create the tar on the disk first.

jj-umn:
@chambm  This is not an AB SCIEX format,  I was just trying to make it easier for users to upload a single archive of the directory.   I think it would be best to make this consistent with your other directory based rawtar datatypes.   I chose zip on the advice of a MSI proteomics consultant;  He said the users he knew would figure out winZip, but be baffled by tar.  I personally prefer tar, and would like them to be able to be stored in gz or bz similar to fastq files to reduce storage space.  

Would you be willing to add AB SCIEX to your rawtar types?  If so, I'll close this pull request.   Thanks.

chambm:
Yes I considered the built-in zip capability of Windows as an advantage for zip, but it would still be a major pain in the ass to do that for multiple wiff/scan pairs.

I made a VBS script as a short-term solution. A user would just drop it in the directory you want to tar up everything in, and double click it. It downloads 7za.exe and creates tars for all candidates. The downloading 7za.exe part is a bit of a security risk I suppose (but no different than running somebody's custom binary). If there's a better way to do that I'm happy to replace it.

And yeah, I'll add WIFF/SCAN to my PR. It's a little different than the others but should be nearly the same to handle.

```
Set objFSO = Createobject("Scripting.FileSystemObject")

If Not objFSO.Fileexists("7za.exe") Then
  Set objXMLHTTP = CreateObject("MSXML2.ServerXMLHTTP.6.0")
	objXMLHTTP.open "GET", "https://drive.google.com/uc?export=download&id=1y8mtPb9lt8536Jfz4mkzVxnlOxADH9lE", false
	objXMLHTTP.setRequestHeader "If-Modified-Since", "Sat, 1 Jan 2000 00:00:00 GMT"
	objXMLHTTP.send()
	
	If objXMLHTTP.Status = 200 Then
	  Set objADOStream = CreateObject("ADODB.Stream")
	  objADOStream.Open
	  objADOStream.Type = 1 'adTypeBinary
	
	  objADOStream.Write objXMLHTTP.ResponseBody
	  objADOStream.Position = 0    'Set the stream position to the start
	
	  objADOStream.SaveToFile "7za.exe"
	  objADOStream.Close
	  Set objADOStream = Nothing
	Else
		WScript.Echo "Error downloading: " + objXMLHTTP.Status
	End if
End if
	
Set objShell = CreateObject("WScript.Shell")

For Each oFile In objFSO.GetFolder(objFSO.GetAbsolutePathName(".")).SubFolders
	ext = UCase(objFSO.GetExtensionName(oFile.Name))
	If Not objFSO.FileExists(oFile.Name + ".tar") And (ext = "RAW" Or ext = "D") Then
	  cmd = "7za.exe a -ttar """ + oFile.Name + ".tar"" """ + oFile.Name + """"
		'WScript.Echo cmd
		objShell.Run cmd, 1, true
  End if
Next
```

jj-umn:
@chambm   Thanks, Matt.   I'm quite ignorant of anything Windows related, so it's nice to have an expert step in.

PR #7297, opened 2019-02-02
Add new test syntax specific for HDF5 files

bgruening:
HDF5 files can be really large and can not be `diff`ed. Moreover, they tend to be slightly different depending on the target environment on which they were produced. This makes it very hard to write tool test that involve HDF5 files.

This PR adds two new test methods to the tool-test syntax. `hdf5_keys` can be used to check the outer container names/keys in a HDF5 file and `hdf5_attrs` can be used to check arbitrary attributes a file should have.

Example:
```xml
<output name="output" hdf5_keys="bins, chroms,indexes,pixels" \
   hdf5_attrs="nbins:33754, nchroms:15" />
```

This should be helpful for HiCExplorer (@joachimwolff), HiFive (@jxtx), scanpy (@bebatut) and various other tools, including ecology (@yvanlebras).


bgruening:
Alternatively, we could try to add a bunch more metadata to the HDF5 main class and can filter for that? I'm not sure we can test for metadata without specifying an output file at the moment.

blankenberg:
I have been thinking about a similar output testing problem (testing QIIME2 .qza/v files). These are zip archives containing multiple files (so the argument to uncompress the output file and only get first file doesn't help us) with content that needs to be tested. This results in the same problem as with HDF5 datasets: there is a file that can't be diffed, and contains multiple 'datasets', or other extractable fields.

I would propose approaching this problem a bit differently than this PR, which adds hdf5 specific options to the tool xml parser. 

Can we define a set of generic 'extract/filter/groom' options (maybe just call it all 'filters') in the standard tool xml parser.
Then for each datatype in its class in `lib/galaxy/datatypes/xyz.py`, can we define what it actually means to filter/extract/groom the output dataset for testing.
Then we could have multiple testing elements (like with composite datasets, but instead of ` <extra_files/>`) in the tool.xml's `<test><output_1><element name="INTERNAL_OBJECT_ID" element_type="file">`.

In the case of a zipfile, we could grab and test specific files (with all of the same options available to a standard/flat output dataset).

For HDF5 case, you provide the internal path to the dataset/metadata/whatever, extract those contents to a file, then compare against the known test file. Can allow e.g. `element_type=structure` to get a JSON/YAML/whatever representation of the internal structure of the HDF5 file with attributes like size.

jmchilton:
The assertion testing framework is pluggable and already has vaguely datatype aware testing (e.g. tabular tests). https://github.com/galaxyproject/galaxy/tree/dev/lib/galaxy/tools/verify/asserts

I feel like this is the correct entry point to add tests like this.

bgruening:
@jmchilton thanks! I hope this one here is better: https://github.com/galaxyproject/galaxy/pull/7301

thermokarst:
@blankenberg - I have opened a PR [here](https://github.com/galaxyproject/galaxy/pull/7400) related to zip-file testing, as you mentioned above. It does not set up the generic filtering you described, rather, it defines a "context-aware" tag, for nesting tests for specific members of zip archives (it attempts to emulate the changes introduced in #7301). Thanks!

PR #7300, opened 2019-02-05
 Do not remove tools from other installed revisions when uninstalling TS repo

nsoranzo:
Multiple installed revisions of a repository can provide the same version of a tool, e.g. for the repository

https://toolshed.g2.bx.psu.edu/view/nilesh/rseqc

only one tool out of 22 has a different version in revision 57 with respect to revision 56.

In this case, when uninstalling one repository, the tools with the same guid were removed also from the other repository revisions in the `shed_tool_conf` file. This can be prevented by checking also the tool path.

Includes also 2 other fixes related to tool installation, see commits for details.


mvdbeek (approved):
Thanks @nsoranzo !

martenson:
very nice bug squish! thanks @nsoranzo

PR #7301, opened 2019-02-05
[19.01] Add new test syntax specific for HDF5 files

bgruening:
HDF5 files can be really large and can not be `diff`ed. Moreover, they tend to be slightly different depending on the target environment on which they were produced. This makes it very hard to write tool test that involve HDF5 files.

This PR adds two new test methods to the tool-test syntax. 

Example:
```xml
        <test>
            <param name="input" value="matrix.cool" />
            <output name="output">
                <assert_contents>
                    <has_keys keys="bins, chroms,indexes,pixels" />
                    <has_attr name="nbins" value="33754" />
                    <has_attr name="nchroms" value="15" />
                </assert_contents>
            </output>
        </test>
```

This should be helpful for HiCExplorer (@joachimwolff), HiFive (@jxtx), scanpy (@bebatut) and various other tools, including ecology (@yvanlebras). 

replaces https://github.com/galaxyproject/galaxy/pull/7297


jmchilton:
Closer - thanks for modifying this. What do you think about https://github.com/bgruening/galaxy/pull/41?

bgruening:
@jmchilton !!! The same questions that I had during my bike ride to work today. Actually there is one more, do you prefer `name` or should this also be `key` (key:value)?

Thanks John!

jmchilton:
Yeah - I guess we should get this right and rename name to key. Thanks!

bgruening:
@jmchilton done and green :)

Issue #7293, opened 2019-02-01
Workflow map-over not mapping over


mvdbeek:
I'm not sure at what step mapping over should occur when I have a Workflow that defines a `list` input and I use a `list:list` input. If I connect a multi data tool the `list:list` gets completely reduced, which is unfortunate if you want to scale up a workflow by nesting samples.

This works fine if I do this through the regular tool interface, but the workflow code never runs through `expand_meta_parameters`, where the magic seems to happen.

mvdbeek:
Fixed by https://github.com/galaxyproject/galaxy/pull/7302

PR #7302, opened 2019-02-05
[19.01] Don't reduce list:list in multi data input workflow step

mvdbeek:
There is some unfortunate divergence between the mapping over code in tool runs vs workflow runs.
Without this patch a list:list input to a multiple="true" data input would get fully reduced in a workflow, while tool runs would just reduce the inner collection (and I think that's the correct behavior).

Fixes https://github.com/galaxyproject/galaxy/issues/7293

### DIFF
--- a/test/api/test_workflows.py
+++ b/test/api/test_workflows.py
@@ -788,7 +788,6 @@ steps:
                 }
                 invocation_id = self.__invoke_workflow(history_id, workflow_id, inputs)
                 self.wait_for_invocation_and_jobs(history_id, workflow_id, invocation_id)
-                self.dataset_populator.wait_for_history(history_id, assert_ok=True)
                 output_collection = self.dataset_populator.get_history_collection_details(history_id, hid=6)
                 assert output_collection['collection_type'] == 'list'
                 assert output_collection['job_source_type'] == 'ImplicitCollectionJobs'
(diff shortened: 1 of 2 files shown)


jmchilton (approved):

test/api/test_workflows.py
```diff
@@ -766,6 +766,33 @@ def test_workflow_resume_with_mapped_over_input(self):
...
+""")
+            with self.dataset_populator.test_history() as history_id:
+                hdca_id = self.dataset_collection_populator.create_list_of_list_in_history(history_id).json()
+                self.dataset_populator.wait_for_history(history_id, assert_ok=True)
+                inputs = {
+                    '0': self._ds_entry(hdca_id),
+                }
+                invocation_id = self.__invoke_workflow(history_id, workflow_id, inputs)
+                self.wait_for_invocation_and_jobs(history_id, workflow_id, invocation_id)
+                self.dataset_populator.wait_for_history(history_id, assert_ok=True)
```
nsoranzo: This seems unnecessary since `wait_for_invocation_and_jobs()` calls `wait_for_history_jobs()` which in turn calls `wait_for_history()`.

Same for `test_workflow_run_output_collection_mapping`.

nsoranzo (approved):

PR #7303, opened 2019-02-05
Define doctype method on XMLParser target

mvdbeek:
That should silence some Deprecation Warnings.

### DIFF
--- a/lib/tool_shed/util/xml_util.py
+++ b/lib/tool_shed/util/xml_util.py
@@ -1,7 +1,6 @@
 import io
 import logging
 import os
-import sys
 import tempfile
 from xml.etree import ElementTree as XmlET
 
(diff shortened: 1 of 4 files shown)


lib/tool_shed/util/xml_util.py
```diff
@@ -82,18 +85,10 @@ def parse_xml(file_name):
         return None, "File does not exist %s" % str(file_name)
 
     with open(file_name, 'r') as fobj:
-        if using_python_27:
```
nsoranzo, line 85: You can now remove the definition of `using_python_27` at line 13.

nsoranzo (approved):

mvdbeek:
Hmm, looks like there's one left:
`  /usr/lib/python2.7/xml/etree/ElementTree.py:1638: DeprecationWarning: This method of XMLParser is deprecated.  Define doctype() method on the TreeBuilder target.
`

Could be coming from one of the dependencies though.

nsoranzo:
@mvdbeek https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/util/__init__.py#L219

mvdbeek:
Hmm, that does define `doctype`, so I think it should be fine ?
```
import warnings
warnings.simplefilter('always', DeprecationWarning)
from galaxy.util import parse_xml
parse_xml('../config/tool_conf.xml')
```
doesn't report any warning

PR #7305, opened 2019-02-05
Simplify SafeStringWrapper.__new__

mvdbeek:
This removes an attempt at sanitizing values before passing values on to the super class constructor. This doesn't seem necessary as far as I can tell (see the doctest for my best guess as to why this was done).
This was introduced in commit 581977f112d86c9b01d951aa75ed98696ab064d8, which fixed a security issue.

Removing this silences a lot of `object() takes no parameters` deprecation warnings on python 2.


mvdbeek:
@blankenberg can you have a look at this ? Not 100% sure this wouldn't re-open the vulnerability that you fixed in https://github.com/galaxyproject/galaxy/commit/581977f112d86c9b01d951aa75ed98696ab064d8

mvdbeek:
OK, that'll need more work ...

PR #7309, opened 2019-02-05
[19.01] More contrast in CSS

hexylena:
Per discussion with @Slugger70 where he complained of similar lack of contrast issues as I have, so we're filing this last minute PR for that:

image

and

image

Another option would be a box shadow?

image

### DIFF
--- a/client/galaxy/style/scss/base.scss
+++ b/client/galaxy/style/scss/base.scss
@@ -199,6 +199,7 @@ body {
     background: $panel-bg-color;
     display: flex;
     flex-direction: column;
+    box-shadow: $gray-700 0px 0px 10px;
 }
 
 #left,
@@ -405,6 +406,7 @@ div.unified-panel-body-background {
     margin-bottom: 0px;
     background-color: $base-color-1;
     height: $masthead-height;
+    z-index: 250; // Above #left/#right to cover up 'glow' from box-shadow
     .navbar-nav {
         min-height: 100%;
         > li {
--- a/client/galaxy/style/scss/base.scss
+++ b/client/galaxy/style/scss/base.scss
@@ -623,7 +623,7 @@ div.toolFormRow {
         @extend .py-1;
         @extend .px-2;
         cursor: move;
-        background: $panel-bg-color;
+        background: $base-color-3;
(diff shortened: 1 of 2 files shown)


Slugger70:
Great idea. The few users I showed with the low/no contrast were confused and didn't like it.

mvdbeek:
There are some test failures in travis (https://travis-ci.org/galaxyproject/galaxy/jobs/489285867#L1215) and you need to target the release_19.01 branch.
Otherwise I think my preference is 
- shadow
- base-color-3
- as it is now

martenson:
this probably needs a rebase, the base branch change did not do it

hexylena:
Yeah I'm reworking it a bit.

`base-color-3` is problematic, as it suffers from depending too heavily on the top bar color. If your masthead is black, the panels are quite dark, if your masthead is red the panels are white. So think maybe need to just opt for gray-200

hexylena:
gray-200 for panels, and I've toned down the box shadow as well.

image
image

martenson:
I am for a contrast increase but not a fan of the shadow, our path seemed to be towards minimalism and plenty of space, shade does not fit there imho.

dannon:
@martenson Totally agree on the shadow, would prefer space and contrast as necessary.  Shadows aren't part of our style anywhere else (and personally I prefer it to stay that way).

hexylena:
@dannon box shadows are used in numerous places so it didn't seem like a huge stretch.

Some examples / list of places where shadows can be removed, if that is not the intended path:

image
image
image
image

dannon:
@erasche I'd forgotten they were in the history multiview.  The others are select2 and toastr specific styles (which we are getting rid of) and not galaxy-based styles.

hexylena:
@dannon personally I love careful use of drop shadows, but, if removing them is the way, can do that.

dannon:
@erasche I just think it's at odds with the minimal/clean/flat look of the bootstrap-based styling we've always gone for, which is exactly why the toastr popups seem jarring (at least to me).  I could be wrong though.

hexylena:
Sounds fine, done!

Issue #7075, opened 2018-12-05
unable to add emoji hashtags to Test Galaxy


martenson:
works on usegalaxy.eu

on test.galaxyproject.org after saving the `#🐟` you will get `name` or `##`

![screenshot 2018-12-05 15 23 19](https://user-images.githubusercontent.com/1814954/49541826-e3cbaf80-f8a1-11e8-80ab-eb56c9321c15.png)
![screenshot 2018-12-05 15 24 03](https://user-images.githubusercontent.com/1814954/49541827-e3cbaf80-f8a1-11e8-924b-07e28a33d723.png)

hexylena:
Yeah! Now that I've discovered I can just #:bee: for my apiMel3 samples and all my :hankey: samples can be easily labeled... I can't live without it. ;)

![grafik](https://user-images.githubusercontent.com/458683/49542145-a4668880-f8cc-11e8-8e4c-fe6bafe3319c.png)

martenson:
I guess somewhere between 18.05 and dev we started treating tag names differently?

hexylena:
Update, cannot see emojis in 18.09, @bgruening also noticed that they stopped working. So between 18.05 and 09.

mvdbeek:
Fixed by https://github.com/galaxyproject/galaxy/pull/7310

martenson:
🎉

PR #6572, opened 2018-08-07
Add tagging rule to history level rule builder, copy HDAs when building collection

mvdbeek:
Tagging is useful also when you already have a history with datasets that contain for example group information in the name. To avoid adding tags to the source HDA, which might be confusing we now copy HDAs when building collections from HDAs in the UI, but default to not doing this in the API for backwards compatibility.

PR #7310, opened 2019-02-05
[19.01] Fix emoji tags

hexylena:
Hey @mvdbeek apparently #6572 caused the regression in #7075. Would this patch sit well with you?

I definitely get blocking the other `C.` classes of characters, but `Cn` includes a lot of emoji that I'd like to have, and shouldn't include any whitespace which I think are just in Cc/Cf?

### DIFF
--- a/lib/galaxy/util/__init__.py
+++ b/lib/galaxy/util/__init__.py
@@ -1058,7 +1058,7 @@ def smart_str(s, encoding=DEFAULT_ENCODING, strings_only=False, errors='strict')
 
 def strip_control_characters(s):
     """Strip unicode control characters from a string."""
-    return "".join(c for c in unicodify(s) if unicodedata.category(c)[0] != "C")
+    return "".join(c for c in unicodify(s) if unicodedata.category(c)[0] != "C" or unicodedata.category(c) == "Cn")
 
 
 def strip_control_characters_nested(item):
--- a/lib/galaxy/util/__init__.py
+++ b/lib/galaxy/util/__init__.py
@@ -1058,7 +1058,7 @@ def smart_str(s, encoding=DEFAULT_ENCODING, strings_only=False, errors='strict')
 
 def strip_control_characters(s):
     """Strip unicode control characters from a string."""
-    return "".join(c for c in unicodify(s) if unicodedata.category(c)[0] != "C" or unicodedata.category(c) == "Cn")
+    return "".join(c for c in unicodify(s) if unicodedata.category(c) != "Cc")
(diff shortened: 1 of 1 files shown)


martenson:
@erasche @mvdbeek what did you use as a source for the decision what to filter?

mvdbeek:
@martenson 
https://github.com/galaxyproject/galaxy/pull/6572#issuecomment-412562841

So if `Cn` doesn't cause this issue it's fine with me.

lib/galaxy/util/__init__.py
```diff
@@ -1058,7 +1058,7 @@ def smart_str(s, encoding=DEFAULT_ENCODING, strings_only=False, errors='strict')
 
 def strip_control_characters(s):
     """Strip unicode control characters from a string."""
-    return "".join(c for c in unicodify(s) if unicodedata.category(c)[0] != "C")
+    return "".join(c for c in unicodify(s) if unicodedata.category(c)[0] != "C" or unicodedata.category(c) == "Cn")
```
mvdbeek: Had to look twice at this to understand that `unicodedata.category(c)[0]` and `unicodedata.category(c)` are not the same.
On a sidenote it also looks like this is only an issue on python2: `unicodedata.category('💩')` is `So` on python 3.
How about this?
```suggestion
    return "".join(c for c in unicodify(s) if unicodedata.category(c) != "Cc")
```
That handles the one case I wanted to deal with in https://github.com/galaxyproject/galaxy/commit/366701cfb23d8de31eac3c4312c6accca2e74ef1

hexylena: No concerns for Cf? I couldn't quite tell what it was (their documentation for "what is in the group" is rubbish. I'm happy to just make an exception for Cn since I know that's a high range / private use area.

hexylena:
@martenson there's some info [on this ftp server](ftp://unicode.org/Public/5.1.0/ucd/) but I haven't found any good listing of "These are the codepoints in this group"

lib/galaxy/util/__init__.py
mvdbeek: I think it's only NUL (0x00) characters that are problematic, as far as I understand it might even be restricted to `\0`

hexylena: Ok, sounds good, I'll just apply your patch then.

mvdbeek (approved):

martenson:
xref https://www.unicode.org/reports/tr44/#GC_Values_Table

PR #7313, opened 2019-02-06
 Build tool XML schema docs using ``sphinx_markdown_tables`` 

nsoranzo:
instead of converting the Markdown to reStructuredText with pandoc.

This fixes the broken tables in `dev/schema.html` , but unfortunately code inside tables is not coloured as in the rest of the document.

Also:
- Remove architecture slides, now available at https://galaxyproject.github.io/training-material/topics/dev/tutorials/architecture/slides.html
- Disable auto toctree for Markdown docs, which was breaking lists of links
- Change cross references to use the `.md` extension instead of `.html`, which also works on GitHub
- Various other doc fixes


jmchilton:
Thanks for all the work and 😢.

PR #7315, opened 2019-02-06
Remove unused keyword in resend email helper

guerler:


nsoranzo (approved):
Introduced in commit a7bb8db3a966f87c2f472c3bc34768fe135cb6f1 , thanks for fixing it!

PR #7316, opened 2019-02-06
Support loading/using tools that are not in the install database

natefoo:
Note that such tools do not appear in the admin interface, but the presumed use case is for tools that you are not managing in this particular Galaxy instance (e.g. they are in CVMFS) anyway. The only downside is you won't be able to use the admin UI to view and fix dependencies (which we also expect to be installed in CVMFS, but you could imagine wanting to resolve them locally instead), but that might be fixable.

### DIFF
--- a/lib/tool_shed/tools/tool_validator.py
+++ b/lib/tool_shed/tools/tool_validator.py
@@ -291,16 +291,10 @@ class ToolValidator(object):
             tool = create_tool_from_source(config_file=full_path, app=self.app, tool_source=tool_source, repository_id=repository_id, allow_code_files=False)
             valid = True
             error_message = None
-        except KeyError as e:
-            tool = None
-            valid = False
-            error_message = 'This file requires an entry for "%s" in the tool_data_table_conf.xml file.  Upload a file ' % str(e)
-            error_message += 'named tool_data_table_conf.xml.sample to the repository that includes the required entry to correct '
-            error_message += 'this error.  '
-        except Exception as e:
-            tool = None
-            valid = False
-            error_message = str(e)
+        except Exception:
+            # As best I can tell, error messages will no longer get back to the UI, so raise exceptions instead so that
+            # at least something will be logged.
+            raise
         return tool, valid, error_message
 
     def load_tool_from_tmp_config(self, repo, repository_id, ctx, ctx_file, work_dir):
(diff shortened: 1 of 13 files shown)


mvdbeek:
> admin UI to view and fix dependencies (which we also expect to be installed in CVMFS, but you could imagine wanting to resolve them locally instead), but that might be fixable.

That is only an issue for toolshed packages, so not a concern I think. For everything else there's the Manage dependencies menu, that lists everything.

lib/tool_shed/tools/tool_validator.py
```diff
@@ -291,16 +291,10 @@ def load_tool_from_config(self, repository_id, full_path):
...
-            valid = False
-            error_message = 'This file requires an entry for "%s" in the tool_data_table_conf.xml file.  Upload a file ' % str(e)
-            error_message += 'named tool_data_table_conf.xml.sample to the repository that includes the required entry to correct '
-            error_message += 'this error.  '
-        except Exception as e:
-            tool = None
-            valid = False
-            error_message = str(e)
+        except Exception:
+            # As best I can tell, error messages will no longer get back to the UI, so raise exceptions instead so that
```
mvdbeek: That doesn't look right to me, there's some additional logic in the calling code that checks if a tool is valid and, if it makes sense, displays `error_message`. Otherwise no need to even return `valid` and `error_message`, I think.

natefoo:
> For everything else there's the Manage dependencies menu, that lists everything.

Awesome, I didn't know that feature didn't use the install DB.

lib/tool_shed/tools/tool_validator.py
natefoo: So the situation I was encountering was an exception due to a change to the method signature of `Tool.__init__()`. The clone was succeeding, the tool was "installed" as far as the install DB was concerned, but it wasn't being added to `shed_tool_conf.xml`. There was no error reported in the UI or log file. So although the error was of my own making, figuring it out was a real nightmare. Presumably there are other non-development errors that would be missed by this as well?

mvdbeek: I'm sure there are, but this breaks all code that uses the current logic. How about just logging the exception here?

mvdbeek:
> Awesome, I didn't know that feature didn't use the install DB.

We should make this more prominent in the admin panel, I think it's the most frequent advice I send out to the mailing list / gitter. And it could use some UI love ...

mvdbeek:
@natefoo I've added some changes in https://github.com/natefoo/galaxy/pull/10, with those the tests should pass

natefoo:
Thanks! I hadn't had time to get back to this.

mvdbeek (approved):

mvdbeek:
@galaxybot test py3

PR #7317, opened 2019-02-07
[WIP] jobs_summary API for workflow invocations.

jmchilton:
API backbone used to generate the progress bar for collections but applied to a workflow invocation. Should be helpful in quickly summarizing all the job states in a workflow invocation.


jmchilton:
This API test passes just fine locally in isolation. Maybe an sqlalchemy vs. postgres problem or maybe something about running the full suite and having jobs of different states...

mvdbeek:
@galaxybot test this

PR #7318, opened 2019-02-07
Fix popupmenu creation

### LINKED ISSUE #7314: "Repository Actions" action menu exploded
Branch affected: dev (future 19.05), 19.01 works fine

How to reproduce: in the admin interface, click on "Manage tools" and then on one repository name; the "Repository Actions" menu on the top right is exploded in multiple buttons

screenshot_2019-02-06 galaxy installed tool shed repositories

`git bisect` says the cause is commit 21b2dd8 .

dannon:
fixes #7314 

With the init queue changes, make_popupmenus was receiving new, unexpected arguments and attempting scope the popupmenu-setting-queries on them, resulting in no matches (and the repo menus not being stitched together, as observed).

(we probably want to just convert all these to the standard bootstrap4 equivalents, but that's more involved)

### DIFF
--- a/client/galaxy/scripts/layout/menu.js
+++ b/client/galaxy/scripts/layout/menu.js
@@ -9,7 +9,6 @@ import { CommunicationServerView } from "layout/communication-server-view";
 import Webhooks from "mvc/webhooks";
 import Utils from "utils/utils";
 
-
 function logoutClick() {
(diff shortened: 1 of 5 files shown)


davebx (approved):

nsoranzo:
Thanks for the quick fix @dannon !

PR #7321, opened 2019-02-07
Remove overlapping tooltip from dropdown button in library toolbar

guerler:
Having a tooltip in addition to a dropdown with title+subtitle is somewhat redundant and leads in this case to an overlap for auto-positioned tooltips.


nsoranzo:
Thanks for the fast fix @guerler !

martenson:
Ideally our tests wouldn't fail on this, but the tooltips are notoriously problematic when hovered over so I can see this as a good change even though it removes usage help.

PR #7322, opened 2019-02-07
Refactor dataset controller to reuse input validation checks between get/set

jmchilton:


test/api/test_history_contents.py
```diff
@@ -203,6 +203,17 @@ def test_update(self):
...
+        data = {
+            "dataset_id": hda1["id"],
+            "name": "moocow",
+            "dbkey": "?",
+            "annotation": None,
+            "info": "my info is",
+            "operation": "attributes"
+        }
+        update_response = self._set_edit_update(data)
+        assert update_response.status_code == 400
```
nsoranzo: I must be missing something obvious, but what's wrong with this PUT request that makes it return 400? Maybe a comment would help.

PR #7329, opened 2019-02-07
Remove logout mako, redirect to login page

guerler:
This PR removes the logout mako, restores session checks, and fixes some routing and redirecting behavior.


Nerdinacan:
I am glad that we're doing this!

I would make two quick suggestions for the client-side implementation.

If we're making an API call, it seems like the client session data should be wiped following the successful logout, in the .then() part of the promise chain:

```
callYourLogoutAPI().then(result => {
   if (result.is(cool)) {
       user.killsession()
   }
));
```

Also, the value you are looking for in Galaxy.root is now obtained explicitly from getAppRoot() in the event that Galaxy is not yet initialized (Admittedly not much of a concern here.)

Thanks for doing this Sam.

guerler:
@galaxybot test this

jmchilton (approved):

Nerdinacan:
Looks structurally good, one last question, on the logout click is this where you intended to go, or did you want to head to the home page?

```
window.top.location.href = `${galaxy.root}login`;
```

Nerdinacan (approved):

dannon:
This looks good to me, was just waiting on that specific review requested.  I can't find the comment here, but I'll second that it might make more sense to just redirect back to home/welcome rather than login.  We can do that in a followup, though.

PR #7331, opened 2019-02-07
[19.01] Discriminate data managers and allow managed testing

mvdbeek:
I am not a big fan of the heuristic in which we fall back
to loading the data manager tool relative to the data_manager_conf.xml
file, but I don't really see a better way here.

It's late, but I'd like to get this into 19.01 so we can start testing data managers in the IUC before transitioning them to python 3 and / or testing them containerized.


jmchilton:
Thanks @mvdbeek - how exciting!

Issue #7326, opened 2019-02-07
Show versions of duplicate data managers in new local data admin view


Slugger70:
When looking at the Local Data/Data Managers page of the Admin menu, data managers with more than one version installed are shown for each version in the left hand column. However, no version numbers are displayed so I have no idea which is which. See picture below for details.

![screen shot 2019-02-07 at 4 41 14 pm](https://user-images.githubusercontent.com/1740254/52444750-438b4480-2af7-11e9-84ed-259ba701fb0b.png)

Note the two different bwa_mem data managers.

![screen shot 2019-02-07 at 4 41 34 pm](https://user-images.githubusercontent.com/1740254/52444758-48e88f00-2af7-11e9-8b71-09cc61ed496f.png)

Note the two versions of the SnpEff Databases, SnpEff Download and TwoBit data managers.

Would it be possible to either: 1. Display the data manager tool version next to the tool button OR 2. have one tool displayed and default to the latest with the older version accessible via a drop down button in the tool display as in normal Galaxy tools?

Slugger70:
@scrathat @dannon @mvdbeek

PR #7332, opened 2019-02-07
Include version in data_manager_list API and show it in DataManager vue component.

scrathat:
#7326


bgruening:
@scrathat thanks! Can we group/merge the same tool IDs into one entry as proposed by @Slugger70?

bgruening:
Should this target 19.01?

mvdbeek:
@galaxybot test this

scrathat:
> @scrathat thanks! Can we group/merge the same tool IDs into one entry as proposed by @Slugger70?

Different data manager versions are merged into one entry now.

However, for some instances there seems to be a difference between the version info saved in a data manager instance (backend) and the version at the end of a constructed tool url with that same data manager id, e.g. `web.url_for(controller='root', tool_id=data_manager.tool.id)`. Is this normal?

martenson:
@galaxybot test this

martenson:
This PR seems to be showing the version from the `data_manager_conf.xml` i.e. from `<data_manager tool_file="data_manager/bowtie_index_builder.xml" id="bowtie_index_builder" version="0.0.1">` instead of the actual version of the tool xml e.g.`<tool id="bowtie_index_builder_data_manager" name="Bowtie index" tool_type="manage_data" version="1.2.0">` from `bowtie_index_builder.xml`

screenshot 2019-02-15 13 57 36

martenson:
I've pushed this to 19.09 due to outstanding comments

bgruening:
@scrathat you had some questions to @martenson wasn't it?

scrathat:
@martenson i added a commit showing both the tool and the data manager version. As needed i can remove one or the other depending on which version is more relevant.

dm version

nsoranzo:
@galaxybot test this

martenson:
I have installed two revisions of `data_manager_bowtie_index_builder` that have the following versions:
Galaxy___Installed_tool_shed_repositories
Galaxy___Installed_tool_shed_repositories

But I only see the older version in the UI:
Galaxy

scrathat:
I installed multiple revisions of `data_manager_bowtie_index_builder` and `trans.app.data_managers.data_managers.values()` only returns
`"toolVersion": "0.0.2"` and data manager `"version": "0.0.1"` and no other revisions.

Works fine with multiple revisions of `data_manager_build_kraken_database`.

bgruening:
@martenson do you have an idea here? Is that a TS bug?

PR #7333, opened 2019-02-08
Fix rerunning a tool via API

nsoranzo:
When sending an API request `POST /api/tools` with payload:

```python
{
    'action': 'rerun',
    'history_id': history_id,
    'inputs': {},
    'target_dataset_id': target_dataset_id
    'tool_id': tool_id,
}
```

the following exception was raised:

```
  File "lib/galaxy/web/framework/decorators.py", line 283, in decorator
  File "lib/galaxy/webapps/galaxy/api/tools.py", line 456, in create
    return self._create(trans, payload, **kwd)
  File "lib/galaxy/webapps/galaxy/api/tools.py", line 462, in _create
    return self._rerun_tool(trans, payload, **kwd)
  File "lib/galaxy/webapps/galaxy/api/tools.py", line 710, in _rerun_tool
    for action in tool.trackster_conf.actions:
AttributeError: 'NoneType' object has no attribute 'actions'
```

and after fixing this, another exception was raised:

```
  File "lib/galaxy/web/framework/decorators.py", line 283, in decorator
  File "lib/galaxy/webapps/galaxy/api/tools.py", line 456, in create
    return self._create(trans, payload, **kwd)
  File "lib/galaxy/webapps/galaxy/api/tools.py", line 462, in _create
    return self._rerun_tool(trans, payload, **kwd)
  File "lib/galaxy/webapps/galaxy/api/tools.py", line 822, in _rerun_tool
    dataset_dict['track_config'] = self.get_new_track_config(trans, output_dataset)
  File "lib/galaxy/web/base/controller.py", line 1017, in get_new_track_config
    "filters": {'filters' : track_data_provider.get_filters()},
AttributeError: 'NoneType' object has no attribute 'get_filters'
```

Also add basic `get_filters()` method to `BaseDataProvider` class.

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/tools.py
+++ b/lib/galaxy/webapps/galaxy/api/tools.py
@@ -613,9 +613,6 @@ class ToolsController(BaseAPIController, UsesVisualizationMixin):
         decoded_dataset_id = self.decode_id(payload.get('target_dataset_id'))
         original_dataset = self.hda_manager.get_accessible(decoded_dataset_id, user=trans.user)
         original_dataset = self.hda_manager.error_if_uploading(original_dataset)
-        msg = self.hda_manager.data_conversion_status(original_dataset)
-        if msg:
-            return msg
 
         # Set tool parameters--except non-hidden dataset parameters--using combination of
         # job's previous parameters and incoming parameters. Incoming parameters
--- a/lib/galaxy/webapps/galaxy/api/tools.py
+++ b/lib/galaxy/webapps/galaxy/api/tools.py
@@ -625,13 +625,13 @@ class ToolsController(BaseAPIController, UsesVisualizationMixin):
         tool = trans.app.toolbox.get_tool(original_job.tool_id)
         if not tool or not tool.allow_user_access(trans.user):
             return trans.app.model.Dataset.conversion_messages.NO_TOOL
-        tool_params = dict([(p.name, p.value) for p in original_job.parameters])
+        tool_params = {p.name: p.value for p in original_job.parameters}
 
         # TODO: rather than set new inputs using dict of json'ed value, unpack parameters and set using set_param_value below.
         # TODO: need to handle updates to conditional parameters; conditional
         # params are stored in dicts (and dicts within dicts).
         new_inputs = payload['inputs']
-        tool_params.update(dict([(key, dumps(value)) for key, value in new_inputs.items() if key in tool.inputs and value is not None]))
+        tool_params.update({key: dumps(value) for key, value in new_inputs.items() if key in tool.inputs and value is not None})
         tool_params = tool.params_from_strings(tool_params, self.app)
 
         #
@@ -707,17 +707,18 @@ class ToolsController(BaseAPIController, UsesVisualizationMixin):
 
         # Set parameters based tool's trackster config.
         params_set = {}
-        for action in tool.trackster_conf.actions:
-            success = False
-            for joda in original_job.output_datasets:
-                if joda.name == action.output_name:
-                    set_param_value(tool_params, action.name, joda.dataset)
-                    params_set[action.name] = True
-                    success = True
-                    break
+        if tool.trackster_conf:
+            for action in tool.trackster_conf.actions:
+                success = False
+                for joda in original_job.output_datasets:
+                    if joda.name == action.output_name:
+                        set_param_value(tool_params, action.name, joda.dataset)
+                        params_set[action.name] = True
+                        success = True
+                        break
 
-            if not success:
-                return trans.app.model.Dataset.conversion_messages.ERROR
+                if not success:
+                    return trans.app.model.Dataset.conversion_messages.ERROR
 
         #
         # Set input datasets for tool. If running on regions, extract and use subset
--- a/lib/galaxy/visualization/data_providers/basic.py
+++ b/lib/galaxy/visualization/data_providers/basic.py
@@ -54,6 +54,15 @@ class BaseDataProvider(object):
         """
         raise Exception("Unimplemented Function")
 
+    def get_filters(self):
+        """
+        Returns filters for provider's data. Return value is a list of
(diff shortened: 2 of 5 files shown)


lib/galaxy/webapps/galaxy/api/tools.py
```diff
@@ -625,13 +625,13 @@ def _rerun_tool(self, trans, payload, **kwargs):
         tool = trans.app.toolbox.get_tool(original_job.tool_id)
         if not tool or not tool.allow_user_access(trans.user):
             return trans.app.model.Dataset.conversion_messages.NO_TOOL
-        tool_params = dict([(p.name, p.value) for p in original_job.parameters])
+        tool_params = dict((p.name, p.value) for p in original_job.parameters)
```
mvdbeek:
```suggestion
        tool_params = {p.name: p.value for p in original_job.parameters}
```

lib/galaxy/webapps/galaxy/api/tools.py
```diff
@@ -625,13 +625,13 @@ def _rerun_tool(self, trans, payload, **kwargs):
...
 
         # TODO: rather than set new inputs using dict of json'ed value, unpack parameters and set using set_param_value below.
         # TODO: need to handle updates to conditional parameters; conditional
         # params are stored in dicts (and dicts within dicts).
         new_inputs = payload['inputs']
-        tool_params.update(dict([(key, dumps(value)) for key, value in new_inputs.items() if key in tool.inputs and value is not None]))
+        tool_params.update(dict((key, dumps(value)) for key, value in new_inputs.items() if key in tool.inputs and value is not None))
```
mvdbeek:
```suggestion
        tool_params.update({key: dumps(value) for key, value in new_inputs.items() if key in tool.inputs and value is not None})
```

mvdbeek:
Do we know when this was introduced ? Does this happen to all tools ? Should we add a test case if we don't have one ?

nsoranzo:
@mvdbeek I think this API call was not used except for Trackster, and it may be still working when used from Trackster.

I was surprised that we don't have a tool rerun method in BioBlend and started to look at the Galaxy API. This is all we have at the moment, which I think it's missing some of the features available through the UI, but worked for the simplest case with this small modifications to BioBlend:

```diff
diff --git a/bioblend/galaxy/tools/__init__.py b/bioblend/galaxy/tools/__init__.py
index 4915593..2ccf22f 100644
--- a/bioblend/galaxy/tools/__init__.py
+++ b/bioblend/galaxy/tools/__init__.py
@@ -96,7 +96,7 @@ class ToolClient(Client):
         params['link_details'] = link_details
         return self._get(id=tool_id, params=params)
 
-    def run_tool(self, history_id, tool_id, tool_inputs):
+    def run_tool(self, history_id, tool_id, tool_inputs, **kwargs):
         """
         Runs tool specified by ``tool_id`` in history indicated
         by ``history_id`` with inputs from ``dict`` ``tool_inputs``.
@@ -119,6 +119,14 @@ class ToolClient(Client):
         payload = {}
         payload["history_id"] = history_id
         payload["tool_id"] = tool_id
+        action = kwargs.get('action')
+        if action:
+            payload['action'] = action
+            if action == 'rerun':
+                target_dataset_id = kwargs.get('target_dataset_id')
+                if not target_dataset_id:
+                    raise ValueError("target_dataset_id must be specified when action='rerun'")
+                payload['target_dataset_id'] = target_dataset_id
         try:
             payload["inputs"] = tool_inputs.to_dict()
         except AttributeError:
```

So, to answer your Qs:

> Do we know when this was introduced ? Does this happen to all tools ?

I think this never worked outside of Trackster.

> Should we add a test case if we don't have one ?

Yes, I can try to add one.

mvdbeek:
> I was surprised that we don't have a tool rerun method in BioBlend and started to look at the Galaxy API. This is all we have at the moment, which I think it's missing some of the features available through the UI, but worked for the simplest case with this small modifications to BioBlend:

yeah, the UI first fetches the parameters by ~~posting~~ GET to api/job/<job_id>/build_for_rerun and then submits a new POST, that is probably preferable ?

> Yes, I can try to add one.

🥇

lib/galaxy/webapps/galaxy/api/tools.py
nsoranzo: Fixed in a rebase, thanks!

nsoranzo:
> > I was surprised that we don't have a tool rerun method in BioBlend and started to look at the Galaxy API. This is all we have at the moment, which I think it's missing some of the features available through the UI, but worked for the simplest case with this small modifications to BioBlend:
> 
> yeah, the UI first fetches the parameters by posting GET to api/job/<job_id>/build_for_rerun and then submits a new POST, that is probably preferable ?

Not sure what's the way to go here, maybe @jmchilton or @guerler have a better informed opinion.

> > Yes, I can try to add one.
> 
> :1st_place_medal:

Added.

mvdbeek (approved):

jmchilton:
@jgoecks gave me permission to remove ``_rerun_tool`` over a year ago, I just haven't gotten around to it. I don't think it is an active path in the code and I'd prefer to invest in building up whatever API the UI uses.

nsoranzo:
@jmchilton That's fine by me, the present API is anyway missing some features I'd like to see, namely having the option of replacing the failed datasets and collections in the history, and restarting jobs which were waiting on those.

Also, it returns a single dataset instead of the re-run job info.

bgruening:
@nsoranzo @jmchilton should we get something similar in before we have the new API?

nsoranzo:
I think there was a consensus to just have a new, better API endpoint, both @jmchilton and @mvdbeek offered to work on that :wink:

bgruening:
Upps, this bit me again.

mvdbeek:
This should be possible with https://github.com/galaxyproject/galaxy/pull/9802

PR #7234, opened 2019-01-17
Use es6 literal for library toolbar header, use icon buttons, left align

guerler:
Suggestion for library toolbar header appearance.

PR #7337, opened 2019-02-08
ui fixes to datalibs

martenson:
* fix template broken in #7234
* remove bs tooltip, re-add normal title


guerler:
Thanks @martenson

Issue #7335, opened 2019-02-08
CORS blocking TTS access


martenson:
so nobody can install tools from there

reported by @rdvelazquez

![galaxy___administration](https://user-images.githubusercontent.com/1814954/52452900-f74dfd80-2b12-11e9-9959-0767c5f77948.png)

Nerdinacan:
Ahhh yeah. The old tool installer runs inside an iframe and the init buffer was being installed in window.top, which resulted in a CORS error. The beta installer should work, but I'll see what happens if I store that init buffer in the window instead.

dannon:
Xref the previous fix for this: https://github.com/galaxyproject/galaxy/pull/7135

So the init buffer broke it again?

Nerdinacan:
Last time it was the singleton I believe? This time it was the init buffer.

```
// initQueue.js
const queue = (window.top._initQueue = new BehaviorSubject([]));
```

dannon:
@Nerdinacan Ahh, yeah, was the singleton.  Should we hang the initqueue off Galaxy instead of window, since we've already narrowed it to having a single Galaxy singleton?

Nerdinacan:
@dannon Well the point of the init queue was to collect initializations before galaxy was built, right? I just stored it in window.

dannon:
@Nerdinacan Yeah, chicken and egg I guess.  Was trying to trim a global, but it might not be worth the gymnastics.

Nerdinacan:
@dannon I agree, it's a bit distasteful, but there's probably some kind of graph theory that proves that we need to have a global to aggregate global behavior, haha. 

I was talking to Jeremy about the tool servers, more generally. How hard would it be to just turn the tool repos into an API instead of a site hosting its own UI?

dannon:
@Nerdinacan That's exactly what the beta installer is working towards.

Nerdinacan:
@dannon Good news then!

Nerdinacan:
@dannon You know what, the only point of the window storage of that buffer was to store initializations across iframes in the same domain. If we're not bothering with that any more, it can probably just be a closure variable inside the module. So at least it's not on window. Let me try that.

dannon:
Ahh, good thinking, I like that!

martenson:
the new installer uses only API, I will spend some time this release to polish it and we can roll it out as default in 19.05

martenson:
[...]

PR #7338, opened 2019-02-08
Fix CORS issue in legacy test tool frame installation

Nerdinacan:
Addresses
https://github.com/galaxyproject/galaxy/issues/7335

The new initQueue stores a buffer of operations in window.top. In legacy code that executes from inside an iframe (and therefore in a different domain) this will throw a CORS error.

Moving the buffer into the local window. I think it's harmless enough.

### DIFF
--- a/client/galaxy/scripts/onload/initQueue.js
+++ b/client/galaxy/scripts/onload/initQueue.js
@@ -17,7 +17,7 @@
 import { BehaviorSubject } from "rxjs";
 import { filter } from "rxjs/operators";
 
-const queue = (window._initQueue = new BehaviorSubject([]));
+const queue = new BehaviorSubject([]);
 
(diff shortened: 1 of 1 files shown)


martenson:
This needs to target 19.01 so we don't break MTS in the release.

nsoranzo:
This shouldn't have been merged without any other committer approval (and an unresolved comment), right?
ref. https://github.com/galaxyproject/galaxy/blob/dev/doc/source/project/organization.rst#handling-pull-requests

Nerdinacan:
Ahhh sorry. It was my first time! Me and Dannon were talking about it in the issue thread. It was just a quick bug fix with minimal impact, but I'll check out the merge guidelines you linked.

dannon:
Post-facto +1 on this, so we don't have to revert it.

PR #7340, opened 2019-02-09
Fix ``test_import_metadata_regeneration`` API test

nsoranzo:
Should fix this recurrent test failure: https://jenkins.galaxyproject.org/job/docker-api/14480/testReport/junit/test.api.test_histories/HistoriesApiTestCase/test_import_metadata_regeneration/

### DIFF
--- a/test/api/test_histories.py
+++ b/test/api/test_histories.py
@@ -1,4 +1,6 @@
 # -*- coding: utf-8 -*-
+import time
+
 from requests import (
     get,
     post,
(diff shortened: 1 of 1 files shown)


nsoranzo:
An alternative may be to move the `galaxy.tools.imp_exp.JobImportHistoryArchiveWrapper()` call from the `cleanup()` to the `finish()` method in `JobWrapper`, which should keep the import job around until all the metadata regeneration jobs are scheduled.

jmchilton:
> An alternative may be to move the `galaxy.tools.imp_exp.JobImportHistoryArchiveWrapper()` call from the `cleanup()` to the `finish()` method in `JobWrapper`, which should keep the import job around until all the metadata regeneration jobs are scheduled.

I'm actually in process of doing this in a round about way I think... I'd like history import to be a not special case of job finish. Finish should collect model objects created by the job for https://github.com/galaxyproject/galaxy/projects/11 and import should just write these objects and rely on normal job finish stuff. It is a ways off though so this is fine and so would be refactoring cleanup as you suggested.

Issue #700, opened 2015-09-07
Test framework: do not call _bam_to_sam for compare="contains"


peterjc:
The test framework will try to show a diff when the output file does not match, and for the special case of BAM files Galaxy will attempt to convert to SAM prior to generating the diff (see function `_bam_to_sam` in `test/base/twilltestcase.py`).

Example test:

```
<test>
    ...
    <param name="bam_wanted" value="true"/>
    <output name="out_fasta" file="U13small_m.mira4_de_novo.fasta" ftype="fasta" />
    <output name="out_bam" file="empty_file.dat" compare="contains" ftype="bam" />
    <output name="out_maf" file="U13small_m.mira4_de_novo.mira" ftype="mira" />
    <output name="out_log" file="empty_file.dat" compare="contains" ftype="txt" />
</test>
```

I'm being a little sneaky here using `compare="contains"` and an empty file - essentially all I test is that the file exists.

In any case, the very fact I am using `compare="contains"` means that the reference file is not expected to be a full BAM file, and therefore attempting to call `samtools view` on it is likely to fail.

On this test run https://travis-ci.org/peterjc/pico_galaxy/jobs/78430789 for what ever reason this test fails - yet Galaxy attempts to run `samtools view` on the partial BAM file (here literally an empty file):

```
======================================================================
FAIL: MIRA v4.0 de novo assember ( mira_4_0_de_novo ) > Test-1
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/peterjc/pico_galaxy/galaxy-dev/test/functional/test_toolbox.py", line 289, in test_tool
    self.do_it( td )
  File "/home/travis/build/peterjc/pico_galaxy/galaxy-dev/test/functional/test_toolbox.py", line 73, in do_it
    raise e
JobOutputsError: History item  different than expected, difference (using contains):
( /home/travis/build/peterjc/pico_galaxy/galaxy-dev/test-data/empty_file.dat v. /tmp/tmpfQXvtU/tmpUpgWsf/new_files_path_Kp4f1Q/tmpS8u1Wdempty_file.dat )
[...]

PR #7342, opened 2019-02-09
Fall back to binary comparison if BAM conversion fails

bernt-matthias:
the so far undocumented conversion of bam to sam files should be optional.\

one might think about leaving the conversion the default and skipping it if decompress==false

ping @peterjc https://github.com/galaxyproject/galaxy/issues/700

I stumbled over this since stacks produced bam files that can currently not converted https://github.com/pysam-developers/pysam -- so it should be possible to deactivate.

### DIFF
--- a/lib/galaxy/tools/verify/__init__.py
+++ b/lib/galaxy/tools/verify/__init__.py
@@ -107,7 +107,7 @@ def verify(
                     local_fh, temp_name = _bam_to_sam(local_name, temp_name)
                     local_name = local_fh.name
                 except Exception as e:
-                    log.warning( "Conversion BAM to SAM failed for local or temp file. Will compare BAM files")
+                    log.warning( "%s. Will compare BAM files" %e)
             if compare == 'diff':
                 files_diff(local_name, temp_name, attributes=attributes)
             elif compare == 're_match':
@@ -153,11 +153,13 @@ def _bam_to_sam(local_name, temp_name):
     try:
         pysam.view('-h', '-o%s' % temp_local.name, local_name)
     except Exception as e:
-        raise Exception("Converting local (test-data) BAM to SAM failed: %s" % e)
+        msg = "Converting local (test-data) BAM to SAM failed: %s" % e
+        raise Exception(msg)
     try:
         pysam.view('-h', '-o%s' % temp_temp, temp_name)
     except Exception as e:
-        raise Exception("Converting history BAM to SAM failed: %s" % e)
+        msg = "Converting history BAM to SAM failed: %s" % e
+        raise Exception(msg)
     os.remove(temp_name)
     return temp_local, temp_temp
 
--- a/lib/galaxy/tools/verify/__init__.py
+++ b/lib/galaxy/tools/verify/__init__.py
@@ -103,9 +103,11 @@ def verify(
         try:
             compare = attributes.get('compare', 'diff')
             if attributes.get('ftype', None) in ['bam', 'qname_sorted.bam', 'qname_input_sorted.bam', 'unsorted.bam']:
-                if attributes.get("decompress", None):
+                try:
                     local_fh, temp_name = _bam_to_sam(local_name, temp_name)
                     local_name = local_fh.name
+                except Exception as e:
+                    log.warning( str(e), "will compare BAM files")
             if compare == 'diff':
                 files_diff(local_name, temp_name, attributes=attributes)
             elif compare == 're_match':
--- a/lib/galaxy/tools/verify/__init__.py
+++ b/lib/galaxy/tools/verify/__init__.py
@@ -107,7 +107,7 @@ def verify(
                     local_fh, temp_name = _bam_to_sam(local_name, temp_name)
                     local_name = local_fh.name
                 except Exception as e:
-                    log.warning( str(e), "will compare BAM files")
+                    log.warning( "Conversion BAM to SAM failed for local or temp file. Will compare BAM files")
             if compare == 'diff':
                 files_diff(local_name, temp_name, attributes=attributes)
             elif compare == 're_match':
--- a/lib/galaxy/tools/verify/__init__.py
+++ b/lib/galaxy/tools/verify/__init__.py
@@ -103,8 +103,9 @@ def verify(
         try:
             compare = attributes.get('compare', 'diff')
             if attributes.get('ftype', None) in ['bam', 'qname_sorted.bam', 'qname_input_sorted.bam', 'unsorted.bam']:
-                local_fh, temp_name = _bam_to_sam(local_name, temp_name)
-                local_name = local_fh.name
+                if attributes.get("decompress", None):
+                    local_fh, temp_name = _bam_to_sam(local_name, temp_name)
+                    local_name = local_fh.name
             if compare == 'diff':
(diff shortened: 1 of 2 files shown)


lib/galaxy/tools/verify/__init__.py
```diff
@@ -103,8 +103,9 @@ def verify(
         try:
             compare = attributes.get('compare', 'diff')
             if attributes.get('ftype', None) in ['bam', 'qname_sorted.bam', 'qname_input_sorted.bam', 'unsorted.bam']:
-                local_fh, temp_name = _bam_to_sam(local_name, temp_name)
-                local_name = local_fh.name
+                if attributes.get("decompress", None):
```
jmchilton: Shouldn't the default be True here?

bernt-matthias: Good question!

I copied it from here: https://github.com/galaxyproject/galaxy/blob/c1a6a2a48f4c8bdf6682b8ab842a0cd391d40ca9/lib/galaxy/tools/verify/__init__.py#L187 and my intuition was mislead (I guessed that there None means do not decompress, but it does). 

The default value of the decompress attribute seems to be not explicitly documented in the tool xsd file .. but I would interpret it as `False`. As far as I can see also the parser seems to get a default of `False`: https://github.com/galaxyproject/galaxy/blob/2376d7f2fd57fcf9ae00264c4244ba06d63b8906/lib/galaxy/tools/parser/xml.py#L523

So we should make this consistent here .. and add more explicit documentation. I think a default of True should work in most cases and seems to be more convenient for tool developers. 

What do you think?

jmchilton: So the problem I guess is that ``decompress`` is ``False`` by default in general but this bam->sam behavior is ``True`` by default. I'm not necessarily eager to change either default without a profile version flag and who wants to add that complexity...

Can we catch the exception in the sam conversion and just compare them as bams if that fails? I guess that has issues since the semantics of the other attributes changes a bit.

bernt-matthias: I hope my commit is what you meant.

lib/galaxy/tools/verify/__init__.py
```diff
@@ -103,8 +103,11 @@ def verify(
         try:
             compare = attributes.get('compare', 'diff')
             if attributes.get('ftype', None) in ['bam', 'qname_sorted.bam', 'qname_input_sorted.bam', 'unsorted.bam']:
-                local_fh, temp_name = _bam_to_sam(local_name, temp_name)
-                local_name = local_fh.name
+                try:
+                    local_fh, temp_name = _bam_to_sam(local_name, temp_name)
+                    local_name = local_fh.name
+                except Exception as e:
+                    log.warning( "%s. Will compare BAM files" %e)
```
mvdbeek:
```suggestion
                    log.warning("%s. Will compare BAM files" %e)
```

lib/galaxy/tools/xsd/galaxy.xsd
```diff
@@ -1218,9 +1218,7 @@ it may be inconvenient to upload the entiry file and this can be used instead.
     <xs:attribute name="decompress" type="PermissiveBoolean">
       <xs:annotation>
         <xs:documentation xml:lang="en"><![CDATA[
-
-If ``compare`` is set to ``diff``, attempt to decompress both produced output and expected output files if needed before performing the diff. This flag is useful for testing gzipped outputs that are non-deterministic despite having deterministic decompressed contents. This is available in Galaxy since release 17.05 and was introduced in [pull request #3550](https://github.com/galaxyproject/galaxy/pull/3550).
-
+This flag is useful for testing gzipped outputs that are non-deterministic despite having deterministic decompressed contents. Gzipped files (local and produced) are attempted to be decompressed if needed if ``compare`` is set to ``diff``. The default value is False. This is available in Galaxy since release 17.05 and was introduced in [pull request #3550](https://github.com/galaxyproject/galaxy/pull/3550).
```
mvdbeek:
```suggestion
This flag is useful for testing compressed outputs that are non-deterministic despite having deterministic decompressed contents. Compressed files are attempted to be decompressed if needed if ``compare`` is set to ``diff``. The default value is False. This is available in Galaxy since release 17.05 and was introduced in [pull request #3550](https://github.com/galaxyproject/galaxy/pull/3550).
```

peterjc:
On reflection, I think the default should be to decompress for comparison, and in the context of BAM, convert to SAM, see https://github.com/galaxyproject/galaxy/issues/700#issuecomment-465525690

mvdbeek:
That's the default, the title is misleading.

lib/galaxy/tools/verify/__init__.py
```diff
@@ -103,8 +103,11 @@ def verify(
         try:
             compare = attributes.get('compare', 'diff')
             if attributes.get('ftype', None) in ['bam', 'qname_sorted.bam', 'qname_input_sorted.bam', 'unsorted.bam']:
-                local_fh, temp_name = _bam_to_sam(local_name, temp_name)
-                local_name = local_fh.name
+                try:
+                    local_fh, temp_name = _bam_to_sam(local_name, temp_name)
+                    local_name = local_fh.name
+                except Exception as e:
+                    log.warning("%s. Will compare BAM files" %e)
```
mvdbeek:
```suggestion
                    log.warning("%s. Will compare BAM files" % e)
```

mvdbeek (approved):

lib/galaxy/tools/xsd/galaxy.xsd
```diff
@@ -1218,9 +1218,7 @@ it may be inconvenient to upload the entiry file and this can be used instead.
     <xs:attribute name="decompress" type="PermissiveBoolean">
       <xs:annotation>
         <xs:documentation xml:lang="en"><![CDATA[
-
-If ``compare`` is set to ``diff``, attempt to decompress both produced output and expected output files if needed before performing the diff. This flag is useful for testing gzipped outputs that are non-deterministic despite having deterministic decompressed contents. This is available in Galaxy since release 17.05 and was introduced in [pull request #3550](https://github.com/galaxyproject/galaxy/pull/3550).
-
+This flag is useful for testing compressed outputs that are non-deterministic despite having deterministic decompressed contents. Compressed files are attempted to be decompressed if needed if ``compare`` is set to ``diff``. The default value is False. This is available in Galaxy since release 17.05 and was introduced in [pull request #3550](https://github.com/galaxyproject/galaxy/pull/3550).
```
nsoranzo:
```suggestion
When this attribute is true and ``compare`` is set to ``diff``, try to decompress files if needed. This flag is useful for testing compressed outputs that are non-deterministic despite having deterministic decompressed contents. By default, only files compressed with bz2, gzip and zip will be automatically decompressed. This is available in Galaxy since release 17.05 and was introduced in [pull request #3550](https://github.com/galaxyproject/galaxy/pull/3550).
```

mvdbeek: @bernt-matthias if this change is OK to you can you commit this ?

nsoranzo:
Could use a rebase to run lint and unit tests (they are not run on either Travis or CircleCI at the moment).

mvdbeek:
I rebased locally and ran lint and unit tests, all good.

PR #7343, opened 2019-02-11
LDAP search member of filter + Username/email generation

BioinformaticsPlatformWIV-ISP:
Hello galaxy team,

- This add  the possibillity to authenticate by username or email and generate the missing one from the LDAP results.
- The second addition is a member check conditional authentication : a  new feature that allow to manage users by groups (at any LDAP directory node level).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,6 +84,7 @@ job_metrics_conf.xml
 workflow_schedulers_conf.xml
 config/*
 config/plugins/interactive_environments/**/*.ini
+config/plugins/**/.cache
 !config/plugins
 static/welcome.html.*
 static/welcome.html
--- a/scripts/tool_shed/build_ts_whoosh_index.py
+++ b/scripts/tool_shed/build_ts_whoosh_index.py
@@ -2,7 +2,7 @@
 Build indexes for searching the Tool Shed.
 Run this script from the root folder, example:
 
-$ python scripts/tool_shed/build_ts_whoosh_index.py -c config/tool_shed.ini
+$ python scripts/tool_shed/build_ts_whoosh_index.py -c config/tool_shed.ini --hgweb var/hgweb_config_dir/
 
 Make sure you adjusted your config to:
  * turn on searching via toolshed_search_on
@@ -14,7 +14,10 @@ from __future__ import print_function
 
 import logging
 import os
+import shutil
 import sys
+import tempfile
+from distutils.dir_util import copy_tree
 from optparse import OptionParser
 
 from mercurial import hg, ui
@@ -46,14 +49,19 @@ def build_index(sa_session, whoosh_index_dir, path_to_repositories, hgweb_config
     Build the search indexes. One for repositories and another for tools within.
     """
     #  Rare race condition exists here and below
+    tool_index_dir = os.path.join(whoosh_index_dir, 'tools')
     if not os.path.exists(whoosh_index_dir):
         os.makedirs(whoosh_index_dir)
-    tool_index_dir = os.path.join(whoosh_index_dir, 'tools')
-    if not os.path.exists(tool_index_dir):
         os.makedirs(tool_index_dir)
+        work_repo_dir = whoosh_index_dir
+        work_tool_dir = tool_index_dir
+    else:
+        # Index exists, prevent in-place index regeneration
+        work_repo_dir = tempfile.mkdtemp(prefix="tmp-whoosh-repo")
+        work_tool_dir = tempfile.mkdtemp(prefix="tmp-whoosh-tool")
 
-    repo_index_storage = FileStorage(whoosh_index_dir)
-    tool_index_storage = FileStorage(tool_index_dir)
+    repo_index_storage = FileStorage(work_repo_dir)
+    tool_index_storage = FileStorage(work_tool_dir)
 
     repo_index = repo_index_storage.create_index(repo_schema)
     tool_index = tool_index_storage.create_index(tool_schema)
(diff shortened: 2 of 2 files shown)


lib/galaxy/auth/providers/ldap_ad.py
```diff
@@ -205,12 +219,32 @@ def authenticate(self, email, username, password, options):
...
+                register_username = params['displayName']
+                register_email = _get_subs(options, 'auto-register-email', params)
+            else:
+                register_username = _get_subs(options, 'auto-register-username', params)
+                register_email = params['mail']
+        else:
+	    register_email = _get_subs(options, 'auto-register-email', params)
+            register_username = _get_subs(options, 'auto-register-username', params)
+
+        register_username = _clean_register_username(register_username)
```
bernt-matthias: Is it really a good idea to modify the username? I would have assumed that the username is checked letter by letter by the auth provider...

lib/galaxy/auth/providers/ldap_ad.py
```diff
@@ -151,6 +156,15 @@ def ldap_search(self, email, username, options):
                 # setup search
                 attributes = [_.strip().format(**params)
                               for _ in options['search-fields'].split(',')]
+                if string_as_bool(options.get('ldap-autogenerate', False)):
+                    if str(username).lower() == 'none':
+		        attributes.append('displayName')
+                    else:
+                        attributes.append('mail')
+
+                if 'search-memberof-filter' in options:
```
bernt-matthias: Question is if this (a specialized search filter) is actually needed or could be covered with the already present possibility of search filters.

I'm not sure of the setup you have in mind. For my case its a Galaxy that only certain LDAP users can use. They will be autoregistered and con login only if they are found using the search filters.

lib/galaxy/auth/providers/ldap_ad.py
BioinformaticsPlatformWIV-ISP: Done for consistency with the rule display into account creation panel : "Public names must be at least three characters in length and contain only lower-case letters, numbers, dots, underscores, and dashes ('.', '_', '-').".  This is the public name used by Galaxy ( in case of auto-registration) and return by the provider, not the username used by the provider to check the credentials (this one remained unmodified) .

lib/galaxy/auth/providers/ldap_ad.py
BioinformaticsPlatformWIV-ISP: In our case the user can login only if he belongs to the right group in addition to being present in our Active Directory. That allow us to manage several instances of Galaxy with a central authentication provider  and dynamically allow ours users to log into specific instances (with the same credentials) by adding or removing them from customs groups.

lib/galaxy/auth/__init__.py
```diff
@@ -66,7 +66,8 @@ def check_auto_registration(self, trans, login, password, no_password_check=Fals
                     try:
                         auth_return = parse_auth_results(trans, auth_results, options)
                     except Conflict:
-                        break
+                        log.exception(Conflict)
+                        raise
```
nsoranzo: What about:
```python
log.exception(Conflict)
break
```
?

lib/galaxy/auth/providers/ldap_ad.py
```diff
@@ -151,6 +156,15 @@ def ldap_search(self, email, username, options):
                 # setup search
                 attributes = [_.strip().format(**params)
                               for _ in options['search-fields'].split(',')]
+                if string_as_bool(options.get('ldap-autogenerate', False)):
+                    if str(username).lower() == 'none':
+		        attributes.append('displayName')
```
nsoranzo: You are mixing tabs and spaces, please never use tabs to indent Galaxy Python code, see e.g. https://www.python.org/dev/peps/pep-0008/#indentation

We run the code linter flake8 as part of our CI testing, can you please fix the errors for your pull request?: https://travis-ci.org/galaxyproject/galaxy/jobs/491564860

config/auth_conf.xml.sample
```diff
@@ -93,7 +93,15 @@
 <!--        <search-user>cn=jsmith,ou=People,dc=domain,dc=com</search-user>
             <search-password>mysecret</search-password>
 -->
-
+            <!-- Generate username or email from the LDAP results if they are not available from the filter -->
+<!--        <ldap-autogenerate>False</ldap-autogenerate>
```
nsoranzo: This doesn't seem necessary to me if you add the proper fields to `<search-fields/>` above and use them in `<auto-register-username/>` and `<auto-register-email/>` below.

lib/galaxy/auth/providers/ldap_ad.py
BioinformaticsPlatformWIV-ISP: Done. Thank you for yours  detailed explanations.

lib/galaxy/auth/__init__.py
BioinformaticsPlatformWIV-ISP: Why would you hide the exception to the uppers layers , instead of allowing them to handle it, for instance with a custom error page ? In this specific case, a red flash banner with an authentication failure message is gently display to the user.

nsoranzo: Without the `raise` I think you should get a "No such user or invalid password." message, what do you get with the `raise` instead?

BioinformaticsPlatformWIV-ISP: A blue banner with the  'Conflict' message (self.err_msg) is display , most likely the : "Cannot make unique username" raise in 'parse_auth_results'.  A 'break' indeed give a red banner with a "No such user or invalid password" message.  The 409 error state of the 'Conflict' exception does'nt seems to matter in both cases.

config/auth_conf.xml.sample
BioinformaticsPlatformWIV-ISP: Indeed, this seems to be the result of some confusion on our end. This entry and the corresponding code have been removed.

nsoranzo:
@BioinformaticsPlatformWIV-ISP Can you rebase this pull request (or extract the commits that are still relevant), now that https://github.com/galaxyproject/galaxy/pull/8424 has been merged?

PR #7344, opened 2019-02-11
Update ``area`` issue/PR labels

nsoranzo:
- Add some labels that were added on GitHub, but not here:
  - authentication
  - client-build
  - compliance
- Rename `system` to `scripts`, since this was the original (I think) and `system` is too generic
- Remove unused `roadmap`, we use GitHub `Projects` now


martenson:
authorization and authentication folks like to use `authnz` or `authz` I think

jmchilton (approved):

PR #7346, opened 2019-02-11
Refactor item_attrs for reuse without mixin inheritance.

jmchilton:
Functionality that doesn't require instance variables should be accessible without mixin inheritance.


nsoranzo:
@galaxybot test this

jmchilton:
Thanks @nsoranzo

PR #7203, opened 2019-01-11
Replace custom UI colors with equivalent theme colors, consolidate color handling

guerler:
This PR follows-up on improving the color handling by removing duplicates, fixing relationships between colors and resolving conflicts. Colors are derived from the 10 base/brand colors listed in https://github.com/galaxyproject/galaxy/pull/7203/files#diff-083a815db1387e189109dae6cce99198L26.

PR #7350, opened 2019-02-12
Merge 19.01 into dev

nsoranzo:
@erasche scss files had merge conflicts, can you please check?


client/galaxy/style/scss/base.scss
```diff
@@ -608,7 +608,7 @@ div.toolFormRow {
         @extend .py-1;
         @extend .px-2;
         cursor: move;
-        background: $panel-bg-color;
+        background: $base-color-3;
```
nsoranzo: Mmh, this doesn't exist any more: https://travis-ci.org/galaxyproject/galaxy/jobs/492133195

hexylena:
@nsoranzo scss LGTM

client/galaxy/style/scss/base.scss
nsoranzo: It was removed in https://github.com/galaxyproject/galaxy/pull/7203/files#diff-083a815db1387e189109dae6cce99198L7

dannon:
In resolving the conflict (use of a now-nonexistent color var) there was a necessary change to the appearance of the workflow editor titles, seen here:  
image

They're not quite as dark as was introduced in #7309, but I think they still achieve the desired goal, and now they use a standard color.

client/galaxy/style/scss/base.scss
hexylena: Argh, only checked against my PR, and looked fine against that, didn't realise it was later removed. Sorry, should have rebuilt the client.

PR #7355, opened 2019-02-13
Separate Toolshed and Galaxy

guerler:


mvdbeek:
Will this break old instances' access to the tool shed ? If yes, are we globally OK with that ? I think I am, but this is probably a huge deal for many deployers, and our reproducibility claims. Is there a route for old instances to get tool updates beyond updating Galaxy ? Manual installs ?

dannon:
@mvdbeek The goal here is just to move toolshed stuff to a separate, unlinked repo.  Everything should continue to be able to talk to it.

davebx:
@dannon is right. We've made a commitment to preserving interoperability regardless of versions, and our goal is to maintain that commitment as best we can.

PR #7356, opened 2019-02-13
Add --force_path_paste argument to tool test verify script.

jmchilton:
Reverts small changes in https://github.com/galaxyproject/galaxy/pull/7160 and brings in option added in https://github.com/galaxyproject/galaxy-lib/pull/137.

Though #7160 made huge strides in usability of this stuff, I still like making progress toward pasting paths since I believe this is the long term direction this code should move in (https://github.com/galaxyproject/galaxy/pull/7160#issuecomment-450751074). Some people (@bgruening) don't even think Galaxy should have a copy of the test data - so that will even more encourage pasting URIs.

### DIFF
--- a/lib/galaxy/tools/verify/interactor.py
+++ b/lib/galaxy/tools/verify/interactor.py
@@ -73,7 +73,7 @@ class OutputsDict(OrderedDict):
             return item
 
 
-def stage_data_in_history(galaxy_interactor, tool_id, all_test_data, history):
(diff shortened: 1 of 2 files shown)


martenson:
Thanks for resolving this @jmchilton.

bgruening:
(at least test-data should be something optional on the Galaxy server)

Thanks John!

PR #7357, opened 2019-02-13
Separate client node version requirement into individual dotfile for …

dannon:
…easier reading/use in ansible playbooks

@natefoo This should be all you need,  right?


natefoo:
Yep. Thanks!

:+1:

natefoo (approved):

PR #7359, opened 2019-02-14
test: added a test check for counting paired collections

bernt-matthias:
Following up for a question on iuc gitter: 

> Does anyone know how expect_num_outputs are counted for paired collections?

(ping @bgruening) I decided to extent the test .. just to be sure. So the answer all data sets and 1st level collections are counted. Nested collections are not counted. 

Side note: For dynamic nested collections there seems to no way at the moment to check the count of included elements (only 1st level via the count attribute of output_collection).

See also https://github.com/galaxyproject/galaxy/pull/6931 and https://github.com/galaxyproject/galaxy/pull/6951

### DIFF
--- a/test/functional/tools/output_filter.xml
+++ b/test/functional/tools/output_filter.xml
@@ -1,15 +1,21 @@
 <tool id="output_filter" name="output_filter" version="1.0.0">
+  <description>test for output filtering and expect_num_outputs</description>
   <command>
     echo "test" > 1;
     echo "test" > 2;
     echo "test" > 3;
     echo "test" > 4;
     echo "test" > 5;
+    echo "test" > p1.forward;
+    echo "test" > p1.reverse;
+    echo "test" > p2.forward;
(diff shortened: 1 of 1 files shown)


bgruening:
Cool, thanks!

mvdbeek (approved):

mvdbeek:
Thanks @bernt-matthias!

bgruening:
Just a question, is this also what we expect? Counting only the top-level boxes in the history?

I think I get different results here: https://github.com/galaxyproject/tools-iuc/pull/2298 I think it is counting the files created.

Which can be a true positive ... just checking.

mvdbeek:
At this point there's a syntax error: https://travis-ci.org/galaxyproject/tools-iuc/jobs/492831949#L606

bernt-matthias:
@mvdbeek wrong PR ?

mvdbeek:
What do you mean ?

PR #7365, opened 2019-02-15
Remove pin on Sphinx. Update all dependencies

nsoranzo:
with `make update-dependencies` using `pipenv` 2018.11.26 .

Important updates:
- Cheetah3 3.2.0 has now x86_64 wheels on PyPI
- `sphinx-rtd-theme` 0.4.3 has integrated Google Analytics support that we can use for the docs website
- Fix environment markers for `functools32` and `more-itertools`

Also:
- Use `mysqlclient` instead of `mysql-python` as conditional dependency. MySQL-python is legacy and unmaintained. xref. galaxyproject/starforge-recipes#22
- Update documentation about dependencies


nsoranzo:
I had to revert the `uWSGI` to `pyuwsgi` migration because on Python 3.5 it was crashing with `Fatal Python error: GC object already tracked`.

Any way, it is now green (failed selenium test is unrelated).

mvdbeek (approved):

Issue #6120, opened 2018-05-11
[GDPR] History import bugs


hexylena:
I produced [this script](https://github.com/galaxyproject/bioblend/issues/262) which downloaded my history (note the explicit request for include_hidden, include_deleted.)

And I got this out:

![auswahl_126](https://user-images.githubusercontent.com/458683/39927648-7b093350-5522-11e8-9cd7-aa21afe57a06.png)

- [ ] Deleted datasets were dropped completely
- [ ] History tags were not imported (included in `history_attrs.txt`)
- [ ] Dataset tags were not imported (included in `dataset_attrs.txt`)
- [ ] "Size on disk" is zero/wrong
- [ ] Job metrics are completely missing :(
- [ ] API does not give back a future-history-id so I'm unable to automatically correct the name. (Also frustrating that I can't tell if it's "done" or not (assuming active use of the account) so I guess I'll just tell the target galaxy to handle 300 histories at once? (Sure, I could do some sort of check for how many histories now vs when started, but that's not optimal under active use scenario)
- [ ] Failed datasets were marked as successful (despite including `state: "error"` in `job_attrs.txt`)

    ![auswahl_125](https://user-images.githubusercontent.com/458683/39927698-9ef2b782-5522-11e8-85fa-e6e6a5f841fb.png)

- [ ] I tried adding some history + dataset annotations + dataset tags to see if those would work, but galaxy immediately returned the new history archive, not recognising that things were different.
- [ ] Deleting successful datasets also does not permit a new archive download

xref #6082

mvdbeek:
This has changed a ton, I think we can close this

Issue #987, opened 2015-10-26
History export does not include dataset collections


hexylena:
I recently downloaded a history and it did not include any of the dataset collection datasets. (This was an attempt to get around #740 until that's fixed)

hexylena:
Seems to still be an issue.

hexylena:
Happy belated 3rd birthday! :cake: 

Found this because it bit @shiltemann as well. Xref #6120

hexylena:
This history fails: https://usegalaxy.eu/u/saskia/h/gmt-benchmarking---tutorial-samples-20

With an internal server error on 18.05 @jmchilton:

```
w.x.y.z - - [20/Nov/2018:13:52:45 +0200] "GET /history/export_archive?preview=True HTTP/1.1" 500 - "https://usegalaxy.eu/histories/list" "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:63.0) Gecko/20100101 Firefox/63.0"
Traceback (most recent call last):
  File "lib/galaxy/web/framework/middleware/batch.py", line 75, in __call__
    return self.application(environ, start_response)
  File "lib/galaxy/web/framework/middleware/request_id.py", line 15, in __call__
    return self.app(environ, start_response)
  File "lib/galaxy/web/framework/middleware/xforwardedhost.py", line 23, in __call__
    return self.app(environ, start_response)
  File "lib/galaxy/web/framework/middleware/translogger.py", line 71, in __call__
    return self.application(environ, replacement_start_response)
  File "lib/galaxy/web/framework/middleware/error.py", line 166, in __call__
    response = self.exception_handler(exc_info, environ)
  File "lib/galaxy/web/framework/middleware/error.py", line 181, in exception_handler
    get_vars = wsgilib.parse_querystring(environ)
  File "/opt/galaxy/venv/lib/python2.7/site-packages/paste/request.py", line 107, in parse_querystring
    strict_parsing=False)
  File "/usr/lib64/python2.7/urlparse.py", line 408, in parse_qsl
    pairs = [s2 for s1 in qs.split('&') for s2 in s1.split(';')]
  File "lib/galaxy/web/framework/middleware/error.py", line 154, in __call__
    app_iter = self.application(environ, sr_checker)
  File "/opt/galaxy/venv/lib/python2.7/site-packages/paste/recursive.py", line 85, in __call__
    return self.application(environ, start_response)
[...]

PR #7367, opened 2019-02-15
Rework and Generalize History Import and Export

jmchilton:
The tracking PR for remote execution work (https://github.com/galaxyproject/galaxy/pull/7058) bumped into the problem of needing to serialize/deserialize Galaxy models to/from workers. Pickling was okay initially but doesn't solve a lot of the merging things back problem. In this standing branch I'm building up a library for dealing with model object serialization and deserialization effectively. Rather than write this from scratch - I'm re-worked history import/export to be a natural first use case of this framework.

Currently this branch includes the following:

- Allow import and export collections.
- Fix job I/O for separated tool and job script streams (merged in #7095 during 19.05)
- Fix job state handling so that all imported jobs have terminal state.
- Imported job parameters were not usable I think (based on workflow extraction testing), they now work and no longer require the toolbox on import or export (so if tools get added later after the import or removed before the export - I believe everything should still work).
- Fixes to eliminate HID handling for associations (objects were previously tracked by HID but inputs/outputs in history may not be in history - so HID is not at all appropriate, problem was amplified by collections).
- Fix job output mapping - was using the HDA name instead of job output name during import (old exports don't have this info - only new ones do).
- Fixes assigning HIDs to datasets during import and fixes to hid_counter - in many ways this was very broken for imported histories.
- Introduce cohesive model object import/export abstractions including a lot of new, well tested functionality.
  - Allow import/export of datasets and collections stand-alone regardless of history.
  - Allow export and re-import of datasets and collections (so for instance datasets can be created for a job and read in at the end of a job and merged back into the database). This is needed for extended metadata collection.
  - Separate implementation for import abstraction that allow import of legacy archives without complicating logic for current archive format.
  - Implementation of import and export abstraction for [BagIt](https://datatracker.ietf.org/doc/rfc8493/) using [bdbag](https://github.com/fair-research/bdbag).
  - Context manager support and clean APIs for all of this including import and export of directories, archives, and bags.
- A lot more integration testing - in particular most of our workflow extraction tests now export and reimport the history to ensure reproducibility survives history import/export.
- A lot more unit testing.

### DIFF
--- a/lib/galaxy/model/store/__init__.py
+++ b/lib/galaxy/model/store/__init__.py
@@ -1,0 +1,1321 @@
+import abc
+import contextlib
+import datetime
+import os
+import shutil
+import tarfile
+import tempfile
+from json import dump, dumps, load
+
+import six
+from bdbag import bdbag_api as bdb
+from boltons.iterutils import remap
+from sqlalchemy.orm import eagerload_all
+from sqlalchemy.sql import expression
+
+from galaxy.exceptions import MalformedContents, ObjectNotFound
+from galaxy.security.idencoding import IdEncodingHelper
+from galaxy.util import FILENAME_VALID_CHARS
+from galaxy.util import in_directory
+from galaxy.util.bunch import Bunch
+from galaxy.version import VERSION_MAJOR
+from ..item_attrs import add_item_annotation, get_item_annotation_str
+from ... import model
+
+ATTRS_FILENAME_HISTORY = 'history_attrs.txt'
+ATTRS_FILENAME_DATASETS = 'datasets_attrs.txt'
+ATTRS_FILENAME_JOBS = 'jobs_attrs.txt'
+ATTRS_FILENAME_IMPLICIT_COLLECTION_JOBS = 'implicit_collection_jobs_attrs.txt'
+ATTRS_FILENAME_COLLECTIONS = 'collections_attrs.txt'
+ATTRS_FILENAME_EXPORT = 'export_attrs.txt'
+ATTRS_FILENAME_LIBRARIES = 'libraries_attrs.txt'
+
+
+class ImportOptions(object):
+
+    def __init__(self, allow_edit=False, allow_library_creation=False, allow_dataset_object_edit=None):
+        self.allow_edit = allow_edit
+        self.allow_library_creation = allow_library_creation
+        if allow_dataset_object_edit is None:
+            allow_dataset_object_edit = allow_edit
+        self.allow_dataset_object_edit = allow_dataset_object_edit
+
+
+class SessionlessContext(object):
+
+    def __init__(self):
+        self.objects = []
+
+    def flush(self):
+        pass
+
+    def add(self, obj):
+        self.objects.append(obj)
+
+    def query(self, model_class):
+
+        def find(obj_id):
+            for obj in self.objects:
+                if isinstance(obj, model_class) and obj.id == obj_id:
+                    return obj
+            return None
+
+        return Bunch(find=find)
+
+
+@six.add_metaclass(abc.ABCMeta)
+class ModelImportStore(object):
+
+    def __init__(self, import_options=None, app=None, user=None, object_store=None):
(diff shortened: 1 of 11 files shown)


hexylena:
Some old history import/export tickets we can close after this :) https://github.com/galaxyproject/galaxy/issues/6120 https://github.com/galaxyproject/galaxy/issues/987

FredericBGA:
> 
> 
> Some old history import/export tickets we can close after this :) #6120 #987

and this PR as well #4532

mvdbeek:
The unit test failures look relevant
`sqlalchemy.orm.exc.FlushError: New instance <HistoryDatasetCollectionAssociation at 0x7ff75f540550> with identity key (<class 'galaxy.model.HistoryDatasetCollectionAssociation'>, (1,), None) conflicts with persistent instance <HistoryDatasetCollectionAssociation at 0x7ff75f211630>`

jmchilton:
This was green before I rebased some more stuff into here this morning. I'll put this back in WIP until I make some progress on this.

jmchilton:
Back out of WIP.

mvdbeek:
@galaxybot test py3

lib/galaxy/model/store/__init__.py
```diff
@@ -0,0 +1,1321 @@
...
+
+    def __init__(self, export_directory, app=None, for_edit=False, serialize_dataset_objects=None, export_files=None):
+        if not os.path.exists(export_directory):
+            os.makedirs(export_directory)
+
+        if app is not None:
+            self.app = app
+            self.security = app.security
+        else:
+            self.security = IdEncodingHelper(id_secret="randomdoesntmatter")
```
mvdbeek, line 937: Is this for future work ? In the current state we don't test or use this, right ?

lib/galaxy/model/store/__init__.py
```diff
@@ -0,0 +1,1321 @@
...
+            add = os.symlink
+        elif self.export_files == "copy":
+            def add(src, dest):
+                if os.path.isdir(src):
+                    shutil.copytree(src, dest)
+                else:
+                    shutil.copyfile(src, dest)
+
+        export_directory = self.export_directory
+        if dataset.id in self.included_datasets:
```
mvdbeek, line 967: is there a case where this is not true ? Those would fail at `if file_name:` on line 998 I think

mvdbeek (approved):
This is awesome, it's working well and is so much cleaner than what we had before, plus a great set of tests! I've added some minor stuff in https://github.com/jmchilton/galaxy/pull/72

mvdbeek:
Feel free to merge, I can also add https://github.com/jmchilton/galaxy/pull/72 as a separate PR

lib/galaxy/model/store/__init__.py
jmchilton: Definitely used extensively in extended metadata collection #7058 but #7684 which is closer uses this (https://github.com/galaxyproject/galaxy/pull/7684/commits/65f747338a75ce205f6375ba4d7249884b29c8ba#diff-c96649080f6e1af45ac25198b92a4d68R28). So unit tests are coming quickly ideally.

lib/galaxy/model/store/__init__.py
jmchilton: So it would seem, good catch. I've eliminated the conditional in https://github.com/galaxyproject/galaxy/pull/7684 (which already had a bunch of model store fixes/enhancements).

jmchilton:
@mvdbeek I've included jmchilton#72 as part of #7684 (which already had a bunch of model store fixes/enhancements), is that sufficient?

jmchilton:
Seems yes, thank you for the merge 😄. I'll write some tests for that tag handling stuff - it was on the TODO list anyway.

mvdbeek:
Yep, I saw you pulled it into https://github.com/galaxyproject/galaxy/pull/7684. Thanks again for spending time on this, I think this was much needed!

PR #7372, opened 2019-02-16
Data set and file name fixes

bernt-matthias:
- when uploading a file the data set name should not contain a path component. this happens for instance if planemo is used and the input data is in subfolders within test-data
- when testing https://github.com/galaxyproject/galaxy/pull/7352 i realised that this breaks for test-data with subfolders since these don't exist (I guess this is how the call to basename ended up there). hence is is now recursively created

second part solves https://github.com/galaxyproject/planemo/issues/901

lets see what CI says about this .. I have no idea about possible side effects

I guess these need backporting (at least to the Galaxy version now used in planemo by default)

### DIFF
--- a/lib/galaxy/tools/verify/__init__.py
+++ b/lib/galaxy/tools/verify/__init__.py
@@ -91,10 +91,9 @@ def verify(
         # if the server's env has GALAXY_TEST_SAVE, save the output file to that dir
         if keep_outputs_dir:
             ofn = os.path.join(keep_outputs_dir, filename)
-            try:
-                os.makedirs(os.path.dirname(ofn))
-            except OSError:
-                pass
+            ofd = os.path.dirname(ofn)
+            if not os.path.isdir(ofd):
+                os.makedirs(os.path.dirname(ofd))
             log.debug('keep_outputs_dir: %s, ofn: %s', keep_outputs_dir, ofn)
             try:
                 shutil.copy(temp_name, ofn)
--- a/lib/galaxy/tools/verify/__init__.py
+++ b/lib/galaxy/tools/verify/__init__.py
@@ -93,7 +93,7 @@ def verify(
             ofn = os.path.join(keep_outputs_dir, filename)
             ofd = os.path.dirname(ofn)
             if not os.path.isdir(ofd):
-                os.makedirs(os.path.dirname(ofd))
+                os.makedirs(ofd)
             log.debug('keep_outputs_dir: %s, ofn: %s', keep_outputs_dir, ofn)
             try:
                 shutil.copy(temp_name, ofn)
--- a/lib/galaxy/tools/verify/__init__.py
+++ b/lib/galaxy/tools/verify/__init__.py
@@ -91,6 +91,10 @@ def verify(
         # if the server's env has GALAXY_TEST_SAVE, save the output file to that dir
         if keep_outputs_dir:
             ofn = os.path.join(keep_outputs_dir, filename)
+            try:
+                os.makedirs(os.path.dirname(ofn))
+            except OSError:
+                pass
             log.debug('keep_outputs_dir: %s, ofn: %s', keep_outputs_dir, ofn)
             try:
                 shutil.copy(temp_name, ofn)
--- a/lib/galaxy/tools/actions/upload_common.py
+++ b/lib/galaxy/tools/actions/upload_common.py
@@ -354,7 +354,7 @@ def create_paramfile(trans, uploaded_datasets):
                 uploaded_dataset.type = 'ftp_import'
             params = dict(file_type=uploaded_dataset.file_type,
                           ext=uploaded_dataset.ext,
-                          name=uploaded_dataset.name,
+                          name=os.path.basename(uploaded_dataset.name),
                           dataset_id=data.dataset.id,
                           dbkey=uploaded_dataset.dbkey,
                           type=uploaded_dataset.type,


lib/galaxy/tools/verify/__init__.py
```diff
@@ -91,6 +91,10 @@ def verify(
         # if the server's env has GALAXY_TEST_SAVE, save the output file to that dir
         if keep_outputs_dir:
             ofn = os.path.join(keep_outputs_dir, filename)
+            try:
+                os.makedirs(os.path.dirname(ofn))
+            except OSError:
+                pass
```
nsoranzo: What about:
```python
            # filename can contain a subdirectory part
            ofd = os.path.dirname(ofn)
            if not os.path.isdir(ofd):
                os.makedirs(ofd)
```
so that `OSError`s different from "directory already exists" are not silenced.

bernt-matthias: good idea

lib/galaxy/tools/verify/__init__.py
```diff
@@ -91,6 +91,8 @@ def verify(
         # if the server's env has GALAXY_TEST_SAVE, save the output file to that dir
         if keep_outputs_dir:
             ofn = os.path.join(keep_outputs_dir, filename)
+            if not os.path.isdir(ofn):
+                os.makedirs(os.path.dirname(ofn))
```
nsoranzo: `ofn` is a filename, so `os.path.isdir(ofn)` is always False, it should be:
```python
            # filename can contain a subdirectory part
            ofd = os.path.dirname(ofn)
            if not os.path.isdir(ofd):
                os.makedirs(ofd)
```

bernt-matthias: oh .. sure .. my fault

lib/galaxy/tools/verify/__init__.py
```diff
@@ -91,6 +91,9 @@ def verify(
         # if the server's env has GALAXY_TEST_SAVE, save the output file to that dir
         if keep_outputs_dir:
             ofn = os.path.join(keep_outputs_dir, filename)
+            ofd = os.path.dirname(ofn)
+            if not os.path.isdir(ofd):
+                os.makedirs(os.path.dirname(ofd))
```
nsoranzo:
```suggestion
                os.makedirs(ofd)
```

nsoranzo (approved):

mvdbeek:
> * when uploading a file the data set name should not contain a path component. this happens for instance if planemo is used and the input data is in subfolders within test-data

I'm a little worried about just doing `os.path.basename` here. If you have subfolders `treatment` and `control`, and you upload a file called `replicate1`, it was previously clear what file this is. Is there a way to know this now ? I assume this is the same code that's also used for uploads by path, where this might be less of a contrived example. Maybe we'd just replace `os.pardir` with `_` ?

nsoranzo:
>  If you have subfolders `treatment ` and `control`, and you upload a file called `replicate1`, it was previously clear what file this is.

If you upload from your computer, that would not apply. Not sure if it makes sense to special case the uploads by path.

It would be nice to have some tests, which could clarify the implications and prevent regressions. Not sure if `galaxy.tools.verify` has any unit test, or if unit tests are the best way to test these issues.

bernt-matthias:
I guess we are talking only about uploads in a `planemo test` run?

Maybe the underlying problem is that its currently not possible to give names to data set. Currently for 

```
<param name="input_type|fqinputs" value="demultiplexed/PopA_01.1.fq,demultiplexed/PopA_02.1.fq" ftype="fastqsanger" />
```

the data set gets the name `demultiplexed/PopA_01.1.fq` which made problems when linking the data to the working dir. And I think the path from the test-data folder should not end up in the data set name.  (I could have done `mkdir $(dirname FILE)`  but with my change it just gets `PopA_01.1.fq`)

If I use a collection I can give names to the elements: 

```
            <param name="input_type|fqinputs">
                <collection type="list">
                    <element name="PopA_01" value="demultiplexed/PopA_01.1.fq" ftype="fastqsanger" />
                    <element name="PopA_02" value="demultiplexed/PopA_02.1.fq" ftype="fastqsanger"/>
                </collection>
            </param>
```

but here then its not possible to give a name to the collection .. 

So for tools that consume the data set (or collection) name (or identifiers), such a thing would be really useful.

mvdbeek:
> If you upload from your computer, that would not apply. Not sure if it makes sense to special case the uploads by path

Sure, I agree, but that's how we've been doing this, and I think this would be a surprise to users that know the current behavior.

> the data set gets the name `demultiplexed/PopA_01.1.fq` which made problems when linking the data to the working dir.

Users are free to name their datasets with `/` in them, so if this causes issues I'd naively say that either the tool needs to do some sanitization or Galaxy itself (a sanitized `.name` / `.element_identifier` variant would probably be nice for tool authors)

bernt-matthias:
> (a sanitized `.name` / `.element_identifier` variant would probably be nice for tool authors)

Indeed. All characters that should not appear in a file name should be replaced. 

But, still, this would not help with my test example where I would need to give names to the uploaded data sets.

mvdbeek:
> But, still, this would not help with my test example where I would need to give names to the uploaded data sets.

I don't understand what isn't possible there. Maybe a complete but minimal example might enlighten things ?

bernt-matthias:
for instance here: https://github.com/galaxyproject/tools-iuc/blob/4e87a14a5479800df9675c1cbcdbe1b11f63653b/tools/stacks2/stacks_denovomap.xml#L121

the elements of `$input_type` have an `.element_identifier` "demultiplexed/PopA_01.1.fq" and "demultiplexed/PopA_02.1.fq". Here the "demultiplexed/" part is just becaue the test-data is structured in subdirectories. There is currently no way so specify a dataset name name/identifier in `<param>`, or?

If I use a data collection as input https://github.com/galaxyproject/tools-iuc/blob/4e87a14a5479800df9675c1cbcdbe1b11f63653b/tools/stacks2/stacks_denovomap.xml#L138 then the elements of the collection can be named, but the collection itself can not be named in the test (which is possible/necessary in the UI).

mvdbeek:
> There is currently no way so specify a dataset name name/identifier in `<param>`, or?

right

> but the collection itself can not be named in the test

Right, but I still don't understand why you need to do that ?

mvdbeek:
> (which is possible/necessary in the UI).

maybe that's what I'm not understanding ? So the input dataset needs to be named in certain way in the history ? I don't think we should be forcing users to do that.

bernt-matthias:
Nah (:)) what I mean is that the user need to give a name to a collection when she creates one.

bernt-matthias:
> Right, but I still don't understand why you need to do that ?

In my example 
1. I have two tools that have some identical outputs .. and I test for identity. 
2. Furthermore I test with collection input and manually selected multiple inputs .. also these cases should have identical output.

For the two cases of 2) I currently get different identifiers, because the data sets names differ, which results in non-identical outputs, ie. failing tests.

bgruening:
ping @joachimwolff as he stumbled upon the very same issue.

bernt-matthias:
The changes to `lib/galaxy/tools/verify/__init__.py` seem to be in dev already due to another PR .. https://github.com/galaxyproject/galaxy/pull/7519 (wanted to suggest to accept this part anyway). 

Since there is currently no consent on the other change I guess we might 
- just close this PR or
- implement an alternative, eg a save element_identifier function that can be used to create file names in tools (which is done quite often).

nsoranzo:
> implement an alternative, eg a save element_identifier function that can be used to create file names in tools (which is done quite often).

I think this was already proposed somewhere (probably on tools-iuc), so :+1:

bernt-matthias:
Just started to think again about "path safe" element identifiers / names and tried to disentangle the actual problems. Maybe we can get some consent (or create some interesting ideas) before I think about starting to implement something. Sorry for the long text... hope it's worth reading :)

**Dataset/collection names sometimes matter**: Many programs use the filename to encode metadata, e.g. sample name or read direction. A good example are demultiplexed reads which are nicely represented as paired collection in Galaxy where the names of the pairs are the sample names and the names of the contained read files are the read direction. So it seems obvious for the tool developer to derive the file names from data set or collection names (tags would be another option). These names are usually automatically generated by upstream tools or during upload.

One problem is that the Galaxy **tool syntax does not allow to name input data sets or collections in tests**. Instead the name of data sets is derived from the path specified in by `value` which contains slashes if the test data is structured in subdirectories. Collections can't be named at all. This sometimes makes it difficult to formulate tests that emulate what upstream tools do or users do in the Galaxy UI (like naming data sets and collections). 

The other problems are **technical**: First of all **problems in testing**: 
- the verify part is now solved in https://github.com/galaxyproject/galaxy/pull/7519
- the upload part (element identifiers are equal to the paths which contain slashes) is still open; _question is if this can be solved in planemo / ephemeris..._?

Next there are problems in the tool development due to special characters in metadata which needs to be mapped to filenames. 

From the tool perspective, most special chars like `$` or spaces are handled already by using single quotes around input and output file names. Only `/` is not treated by this since it can't be used at all in file names on Linux and Windows systems (https://stackoverflow.com/questions/9847288/is-it-possible-to-use-in-a-filename, https://stackoverflow.com/questions/10708334/how-can-i-create-files-on-windows-with-embedded-slashes-using-python). Even with quoted special chars the actual programs might still have problems with non alphanumeric characters in the file names. A feasible solution is something along the lines of `re.sub('[^\w\-_.]', '_', input.element_identifier)` as already implemented in some tools (question is if its necessary to provide a convenience function like safe_element_identifier). The downside is that these changes of the metadata which might lead to side effects in downstream tools that are unexpected by the user. An interesting alternative might be _validators for data set names..._?

mvdbeek:
> tool syntax does not allow to name input data sets or collections in tests

That is fairly easy to implement. You just need to specify the upload name in verify.py.

> safe_element_identifier

Again, that should be straightforward to implement, I think you just need to add another property as https://github.com/galaxyproject/galaxy/blob/d35a07689891b5cff03976ca36e928cf10827594/lib/galaxy/tools/wrappers.py#L449 and https://github.com/galaxyproject/galaxy/blob/d35a07689891b5cff03976ca36e928cf10827594/lib/galaxy/tools/wrappers.py#L252

If you want to implement this I'll be happy to help and review.

mvdbeek:
We decided to push this to 20.01, but my intention is to open a PR that lets the tool author specify the element_identifier / dataset name. This would probably solve the same issue without causing any of the potential downsides.

Issue #7336, opened 2019-02-08
data libraries cannot sniff datatype or autodetect metadata


martenson:
on dev

RequestParameterInvalidException: This Galaxy does not recognize the datatype of: auto

![screenshot 2019-02-08 10 12 18](https://user-images.githubusercontent.com/1814954/52486810-218ed200-2b8a-11e9-9dc0-d25eca1fbbeb.png)

martenson:
it seems this never worked actually, I will look into implementing it.

martenson:
implemented in https://github.com/galaxyproject/galaxy/pull/7379

PR #7379, opened 2019-02-18
implement sniffing of datatypes for lddas

martenson:
implements https://github.com/galaxyproject/galaxy/issues/7336

### DIFF
--- a/lib/galaxy/managers/datasets.py
+++ b/lib/galaxy/managers/datasets.py
@@ -338,7 +338,7 @@ class DatasetAssociationManager(base.ModelManager,
         return rval
 
     def detect_datatype(self, trans, dataset_assoc):
-        """Sniff and assign the datatype to a given dataset associsation (ldda or hda)"""
+        """Sniff and assign the datatype to a given dataset association (ldda or hda)"""
         def __ok_to_edit_metadata(dataset_id):
             # prevent modifying metadata when dataset is queued or running as input/output
             # This code could be more efficient, i.e. by using mappers, but to prevent slowing down loading a History panel, we'll leave the code here for now
--- a/lib/galaxy/managers/datasets.py
+++ b/lib/galaxy/managers/datasets.py
@@ -12,6 +12,7 @@ from galaxy import (
     exceptions,
     model
 )
+from galaxy.datatypes import sniff
 from galaxy.managers import (
     base,
     deletable,
@@ -19,6 +20,7 @@ from galaxy.managers import (
     secured,
     users
 )
+from galaxy.util.checkers import check_binary
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


martenson:
the https://github.com/galaxyproject/galaxy/pull/7379/commits/3f64fef118939155cff92007d73059d2e5554c23 commit maybe justifies backport

lib/galaxy/managers/datasets.py
```diff
@@ -335,6 +337,37 @@ def serialize_dataset_association_roles(self, trans, dataset_assoc):
             rval["modify_item_roles"] = modify_item_role_list
         return rval
 
+    def detect_datatype(self, trans, dataset_assoc):
+        """Sniff and assign the datatype to a given dataset associsation (ldda or hda)"""
```
nsoranzo:
```suggestion
        """Sniff and assign the datatype to a given dataset association (ldda or hda)"""
```

nsoranzo (commented):
Is the failed API test relevant? https://jenkins.galaxyproject.org/job/docker-api/14609/testReport/junit/test.api.test_libraries/LibrariesApiTestCase/test_update_dataset_in_folder/

martenson:
@nsoranzo I think so, will check it out tomorrow

martenson:
With ebd8c44 in place Galaxy is creating jobs in db without a session_id and that seems to work. I think this means that anonymous set_metadata jobs will forever be orphaned, since they won't have user or session_id. Which is probably fine (@natefoo ?). Thanks @jmchilton.

natefoo:
> this means that anonymous set_metadata jobs will forever be orphaned, since they won't have user or session_id

This doesn't seem great - I feel like I'm missing some understanding here though?

jmchilton:
Tools API controller:

```
        history_id = payload.get('history_id', None)
        if history_id:
            decoded_id = self.decode_id(history_id)
            target_history = self.history_manager.get_owned(decoded_id, trans.user, current_history=trans.history)
        else:
            target_history = None
```

Tool actions:

```
        if hasattr(trans, "get_galaxy_session"):
            galaxy_session = trans.get_galaxy_session()
            # If we're submitting from the API, there won't be a session.
            if type(galaxy_session) == trans.model.GalaxySession:
                job.session_id = model.cached_id(galaxy_session)
        if trans.user is not None:
            job.user_id = model.cached_id(trans.user)
        job.history_id = model.cached_id(history)
        job.tool_id = tool.id
```

Clearly these things can be null already, there is no problem here IMO.

martenson:
the last commit was wrongly cherry picked here and is also part of https://github.com/galaxyproject/galaxy/pull/7397, it should create no conflict I think

lib/galaxy/tools/actions/metadata.py
```diff
@@ -18,8 +18,11 @@ def execute(self, tool, trans, incoming={}, set_output_hid=False, overwrite=True
         Execute using a web transaction.
         """
         trans.check_user_activation()
-        job, odict = self.execute_via_app(tool, trans.app, trans.get_galaxy_session().id,
-                                          trans.history.id, trans.user, incoming, set_output_hid,
+        session = trans.get_galaxy_session()
+        session_id = session and session.id
```
mvdbeek, line 22: If this is supposed to be a boolean `has_session_id` might be a better variable name, but I think this should probably be `session_id = session.id if session else None` or `session_id = getattr(session, 'id', None)`?

lib/galaxy/managers/library_datasets.py
```diff
@@ -115,7 +121,7 @@ def _validate_and_parse_update_payload(self, payload):
                 validated_payload[key] = val
             if key in ('file_ext'):
                 datatype = self.app.datatypes_registry.get_datatype_by_extension(val)
-                if datatype is None:
+                if datatype is None and val not in ("auto"):
```
nsoranzo:
```suggestion
                if datatype is None and val not in ("auto",):
```

lib/galaxy/tools/actions/metadata.py
nsoranzo: I think this already equivalent to `session_id = session.id if session else None`

lib/galaxy/managers/library_datasets.py
martenson: How did this pass lint? Thanks.

lib/galaxy/tools/actions/metadata.py
martenson: @nsoranzo that is my understanding also, I like @mvdbeek's explicit proposal though

lib/galaxy/managers/library_datasets.py
nsoranzo: Because it's equivalent to `val not in "auto"` :
```
In [5]: "auto" in "auto"
Out[5]: True
```

martenson: Python tryin to be helpful, nice.

lib/galaxy/tools/actions/metadata.py
mvdbeek: Learning something new every day
```python
>>> 'a' and 'b'
'b'
```
I would have expected `True` here.

martenson:
@mvdbeek I've added a basic API test that triggers the datatype detection. It is fairly primitive, I could add something that actually checks the eventually sniffed datatype but that seemed out of scope for an API test. We test sniffing elsewhere.

Edit: when put like this the test actually does not test much... I guess I will add the check for sniffed extension.

martenson:
@mvdbeek I upgraded the test, now it actually tests the sniffing (for one datatype only, just the process)

mvdbeek (approved):
Thanks @martenson, looks good to me.

PR #7381, opened 2019-02-18
Fix upload button style

guerler:


martenson:
works for me, thanks @guerler !

PR #7382, opened 2019-02-18
[19.01] Fix HDF5 test assert docs

nsoranzo:
The content of `TestAssertions` is special and cannot contain `complexType`s.

Introduced in https://github.com/galaxyproject/galaxy/pull/7301 , they are breaking `make docs`, see for example:

https://jenkins.galaxyproject.org/job/latest-Sphinx-Docs/2223/console


jmchilton:
I think https://github.com/galaxyproject/galaxy/pull/7384 is a better approach, the rest of the assertions should look like the new ones not vice versa. 

https://github.com/bgruening/galaxy/pull/41

> * Stricter validation of XSD (the other assertions don't have this but that is problematic on their part), I'm not sure if the docs will render unevenly now though - we may want to revisit that and do this for all the assertions.

PR #7390, opened 2019-02-20
[19.01] Report line number mismatch in sort collection tool

mvdbeek:
Otherwise this would just fail because sorted_elements is undefined.
I don't think there can be a HDA without metadata, so I've dropped
that check.

### DIFF
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -2701,7 +2701,8 @@ class SortTool(DatabaseOperationTool):
             sorted_elements = [x[1] for x in sorted(presort_elements, key=lambda x: x[0])]
         if sorttype == 'file':
             hda = incoming["sort_type"]["sort_file"]
-            if hda.metadata and hda.metadata.get('data_lines', 0) == len(elements):
+            data_lines = hda.metadata.get('data_lines', 0)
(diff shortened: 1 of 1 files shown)


martenson:
>I don't think there can be a HDA without metadata

This is imo grey zone and I would leave the check in. Ideally there wouldn't be such hdas, but I've seen many cases of hdas/lddas without metadata in local/test galaxies and I bet there are such cases on Main too.

mvdbeek:
> This is imo grey zone and I would leave the check in. Ideally there wouldn't be such hdas, but I've seen many cases of hdas/lddas without metadata in local/test galaxies and I bet there are such cases on Main too.

But they should still have the metadata association object, it'd just be empty I think (see https://github.com/galaxyproject/galaxy/blob/9bac3c44faf51e0e4c396d7f1d3470f7aa4a588e/lib/galaxy/model/mapping.py#L194) ... and also they should be in non-OK (FAILED_METADATA) state and not be selectable anyway. If they are that'd be another, serious IMO, bug.
Also there are plenty of places in the code where we don't check first.

martenson:
@mvdbeek fair enough

>be in non-OK (FAILED_METADATA) state

often they are stuck in SETTING_METADATA

PR #7394, opened 2019-02-20
Restore admin client router handling

guerler:


guerler:
Duplicate.

dannon:
@guerler Was working on larger refactoring separately, but if you want to reopen this I'll just merge the individual change to get the minimal bug fix in if that's fine.

dannon:
@guerler First commit didn't quite fix it; I followed up with a little more of the refactoring that should get the job done.

I'm going through and checking what `kwd` we actually want passed through and I think I'm going to trim that down, though.  I think we intentionally filtered those off in the past.

PR #7396, opened 2019-02-21
Fix pie chart and portlet height in flexboxes

guerler:
Some pie chart settings are not correctly parsed to the underlying plugin. This PR fixes it and also fixes the form height in the published visualizations view.

image


dannon:
@guerler Thanks for the ping, I'll check this out in a few.

dannon:
@guerler Works well for me!

guerler:
Thanks for the review @dannon.

PR #7397, opened 2019-02-21
specify generic exceptions more precisely

### LINKED ISSUE #7393: login/registration form js console errors
after attempting to log in, create account, or reset password with an invalid form (no email address) I receive an error message but also the following javascript console errors:

`xhr.js:178 POST http://127.0.0.1:8080/user/reset_password 400 (Bad Request)`
`xhr.js:178 POST http://127.0.0.1:8080/user/login 400 (Bad Request)`
`xhr.js:178 POST http://127.0.0.1:8080/user/create 400 (Bad Request)`

I think those should not be 'Bad Request's, ideally they wouldn't be thrown at all. In super-ideal world the requests would never actually be fired since empty field cannot be valid here.

martenson:
in user api controller

closes #7393

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/users.py
+++ b/lib/galaxy/webapps/galaxy/api/users.py
@@ -871,7 +871,7 @@ class UserAPIController(BaseAPIController, UsesTagsMixin, CreatesApiKeysMixin, B
                     build_dict['len'] = new_len.id
                     build_dict['linecount'] = new_linecount.id
                 except Exception:
-                    raise exceptions.InternalServerError('Failed to convert dataset.')
+                    raise exceptions.ToolExecutionError('Failed to convert dataset.')
             dbkeys[key] = build_dict
             user.preferences['dbkeys'] = json.dumps(dbkeys)
             trans.sa_session.flush()
--- a/lib/galaxy/webapps/galaxy/api/users.py
+++ b/lib/galaxy/webapps/galaxy/api/users.py
@@ -817,7 +817,7 @@ class UserAPIController(BaseAPIController, UsesTagsMixin, CreatesApiKeysMixin, B
         if len_type not in ['file', 'fasta', 'text'] or not len_value:
             raise exceptions.RequestParameterInvalidException('Please specify a valid data source type.')
         if not name or not key:
-            raise exceptions.RequestParameterInvalidException('You must specify values for all the fields.')
+            raise exceptions.RequestParameterMissingException('You must specify values for all the fields.')
         elif key in dbkeys:
             raise exceptions.Conflict('There is already a custom build with that key. Delete it first if you want to replace it.')
         else:
--- a/lib/galaxy/webapps/galaxy/api/users.py
+++ b/lib/galaxy/webapps/galaxy/api/users.py
@@ -819,7 +819,7 @@ class UserAPIController(BaseAPIController, UsesTagsMixin, CreatesApiKeysMixin, B
         if not name or not key:
             raise exceptions.RequestParameterMissingException('You must specify values for all the fields.')
         elif key in dbkeys:
-            raise exceptions.Conflict('There is already a custom build with that key. Delete it first if you want to replace it.')
+            raise exceptions.DuplicatedIdentifierException('There is already a custom build with that key. Delete it first if you want to replace it.')
         else:
             # Have everything needed; create new build.
             build_dict = {'name': name}
--- a/lib/galaxy/webapps/galaxy/api/users.py
+++ b/lib/galaxy/webapps/galaxy/api/users.py
@@ -19,10 +19,7 @@ from galaxy import (
     util,
     web
 )
-from galaxy.exceptions import (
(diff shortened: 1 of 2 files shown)


jmchilton (approved):

dannon:
@martenson I'm not seeing a change in the response related to #7393 here; were there more changes coming?  (not that this isn't a good idea, just trying to verify what it's changing related to those 4xx responses)

martenson:
@dannon Yeah, the new registration form actually does not use the API, it still uses the old controller with old decorators. I am less concerned about that surface being HTTP precise. I guess technically this does not 'close' the issue, but I think I will close it nevertheless.

dannon:
@martenson Yep, exactly, that's why I was asking if there was more to come :)   No worries, and I agree on just closing that one out.

lib/galaxy/webapps/galaxy/api/users.py
```diff
@@ -598,7 +595,7 @@ def set_password(self, trans, id, payload={}, **kwd):
         """
         user, message = self.user_manager.change_password(trans, id=id, **payload)
         if user is None:
-            raise MessageException(message)
+            raise exceptions.AuthenticationRequired(message)
```
nsoranzo, line 598:
```suggestion
            raise exceptions.AuthenticationFailed(message)
```

lib/galaxy/webapps/galaxy/api/users.py
```diff
@@ -818,11 +815,11 @@ def add_custom_builds(self, trans, id, key, payload={}, **kwd):
         len_type = payload.get('len|type')
         len_value = payload.get('len|value')
         if len_type not in ['file', 'fasta', 'text'] or not len_value:
-            raise MessageException('Please specify a valid data source type.')
+            raise exceptions.RequestParameterInvalidException('Please specify a valid data source type.')
         if not name or not key:
-            raise MessageException('You must specify values for all the fields.')
+            raise exceptions.RequestParameterInvalidException('You must specify values for all the fields.')
```
nsoranzo:
```suggestion
            raise exceptions.RequestParameterMissingException('You must specify values for all the fields.')
```

lib/galaxy/webapps/galaxy/api/users.py
```diff
@@ -818,11 +815,11 @@ def add_custom_builds(self, trans, id, key, payload={}, **kwd):
...
         elif key in dbkeys:
-            raise MessageException('There is already a custom build with that key. Delete it first if you want to replace it.')
+            raise exceptions.Conflict('There is already a custom build with that key. Delete it first if you want to replace it.')
```
nsoranzo: Maybe
```suggestion
            raise exceptions.DuplicatedIdentifierException('There is already a custom build with that key. Delete it first if you want to replace it.')
```
?

lib/galaxy/webapps/galaxy/api/users.py
```diff
@@ -874,7 +871,7 @@ def add_custom_builds(self, trans, id, key, payload={}, **kwd):
                     build_dict['len'] = new_len.id
                     build_dict['linecount'] = new_linecount.id
                 except Exception:
-                    raise MessageException('Failed to convert dataset.')
+                    raise exceptions.InternalServerError('Failed to convert dataset.')
```
nsoranzo:
```suggestion
                    raise exceptions.ToolExecutionError('Failed to convert dataset.')
```

nsoranzo (commented):
I'd also update an error message:
```patch
diff --git a/lib/galaxy/exceptions/error_codes.json b/lib/galaxy/exceptions/error_codes.json
index 1190851..a1b1d08 100644
--- a/lib/galaxy/exceptions/error_codes.json
+++ b/lib/galaxy/exceptions/error_codes.json
@@ -27,7 +27,7 @@
    {
     "name": "USER_OBJECT_ATTRIBUTE_MISSING",
     "code": 400005,
-    "message": "Attempted to create object without required attribute."
+    "message": "Attempted to create or update object without required attribute."
     },
    {
     "name": "USER_SLUG_DUPLICATE",
```

lib/galaxy/webapps/galaxy/api/users.py
martenson: I specifically want 403 here, 401 implies you can authenticate using headers I think

nsoranzo: Uh, you are right, 401 should be used only if the client needs to authenticate using HTTP Authentication, which we don't use. Should we remove this class and corresponding error code to prevent misuse?

martenson: Maybe we could just change it to 403? But I would do it in a different PR since that will probably break a few things. E.g. some of our client javascript checks specific error codes.

nsoranzo: Filed an issue: https://github.com/galaxyproject/galaxy/issues/7407

Issue #4452, opened 2017-08-18
Activating the "modules" dependency resolver prevent a Galaxy instance to start properly


arbernard:
If you try to uncomment the "**modules**" dependency resolver in the **dependency_resolvers_conf.xml** file and try to start the Galaxy instance then it will fail with the following errors:

```
Traceback (most recent call last):
  File "./scripts/manage_tool_dependencies.py", line 87, in <module>
    main()
  File "./scripts/manage_tool_dependencies.py", line 26, in main
    action_func(args)
  File "./scripts/manage_tool_dependencies.py", line 33, in _init_if_needed
    _build_dependency_manager_no_config(kwargs)
  File "./scripts/manage_tool_dependencies.py", line 76, in _build_dependency_manager_no_config
    dependency_manager = DependencyManager( **dependency_manager_kwds )
  File "/home/arbernard/galaxy_project/galaxy/lib/galaxy/tools/deps/__init__.py", line 84, in __init__
    self.dependency_resolvers = self.__build_dependency_resolvers( conf_file )
  File "/home/arbernard/galaxy_project/galaxy/lib/galaxy/tools/deps/__init__.py", line 195, in __build_dependency_resolvers
    return self.__parse_resolver_conf_xml( plugin_source )
  File "/home/arbernard/galaxy_project/galaxy/lib/galaxy/tools/deps/__init__.py", line 210, in __parse_resolver_conf_xml
    return plugin_config.load_plugins( self.resolver_classes, plugin_source, extra_kwds )
  File "/home/arbernard/galaxy_project/galaxy/lib/galaxy/util/plugin_config.py", line 36, in load_plugins
    return __load_plugins_from_element(plugins_dict, source, extra_kwds)
  File "/home/arbernard/galaxy_project/galaxy/lib/galaxy/util/plugin_config.py", line 48, in __load_plugins_from_element
    plugin = plugins_dict[ plugin_type ]( **plugin_kwds )
  File "/home/arbernard/galaxy_project/galaxy/lib/galaxy/tools/deps/resolvers/modules.py", line 47, in __init__
    self.module_checker = AvailModuleChecker(self, self.modulepath, prefetch, self.default_indicator)
  File "/home/arbernard/galaxy_project/galaxy/lib/galaxy/tools/deps/resolvers/modules.py", line 118, in __init__
    for module in self.__modules():
[...]

PR #7398, opened 2019-02-21
Environment modules mapping

FredericBGA:
Based on what @arbernard (thank you!) has done for lmod, here is the same capabilities for module (http://modules.sourceforge.net/) dependency resolver.
I have added a sample file, see #4452

### DIFF
--- a/lib/galaxy/tools/deps/resolvers/modules.py
+++ b/lib/galaxy/tools/deps/resolvers/modules.py
@@ -36,6 +36,7 @@ DEFAULT_MODULECMD_PATH = "modulecmd"  # Just check path
 DEFAULT_MODULE_PATH = '/usr/share/modules/modulefiles'
 DEFAULT_INDICATOR = '(default)'
 DEFAULT_MODULE_PREFETCH = "true"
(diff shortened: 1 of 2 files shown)


lib/galaxy/tools/deps/resolvers/modules.py
```diff
@@ -44,6 +45,8 @@ class ModuleDependencyResolver(DependencyResolver, MappableDependencyResolver):
     resolver_type = "modules"
 
     def __init__(self, dependency_manager, **kwds):
+        # Mapping file management
+        self._set_default_mapping_file(kwds)
```
jmchilton, line 49: The next line should take care of this ☹️ , but the LMOD resolver does the same thing.

jmchilton (approved):

lib/galaxy/tools/deps/resolvers/modules.py
FredericBGA: @jmchilton what do you mean ? Do you want me to improve this. I can try. My contribution here has been only copy/paste :) so I can do better maybe !

PR #7399, opened 2019-02-21
[19.01] Bump yarn network tolerance

dannon:
From 2 minutes to 5 minutes -- 2 minutes was too low for some larger dependencies as debugged with @AjitPS in gitter.


nsoranzo (approved):

martenson:
here is to hope we never need more

PR #7401, opened 2019-02-21
Place oidc login button under regular login form

guerler:
image

Previously:

image

### DIFF
--- a/client/galaxy/scripts/components/login/Login.vue
+++ b/client/galaxy/scripts/components/login/Login.vue
@@ -24,12 +24,10 @@
                         </b-card-footer>
                     </b-card>
                 </b-form>
+                <b-button id="oidc" class="mt-3" @click="submitOIDCLogin()">
+                    <icon class="fa fa-google"/> Sign in with Google
+                </b-button>
             </div>
-            <b-card header="Other Login Options">
-                <b-form id="oidc" @submit.prevent="submitOIDCLogin()">
-                    <b-form-group> <b-button type="submit">Log in with Google</b-button> </b-form-group>
-                </b-form>
-            </b-card>
             <div v-if="show_welcome_with_login" class="col">
                 <b-embed type="iframe" :src="welcome_url" aspect="1by1" />
             </div>
--- a/client/galaxy/scripts/components/login/Login.vue
+++ b/client/galaxy/scripts/components/login/Login.vue
@@ -24,7 +24,7 @@
                         </b-card-footer>
                     </b-card>
                 </b-form>
-                <b-button id="oidc" class="mt-3" @click="submitOIDCLogin()">
+                <b-button v-if="oidcShow" class="mt-3" @click="submitOIDCLogin()">
                     <icon class="fa fa-google"/> Sign in with Google
                 </b-button>
             </div>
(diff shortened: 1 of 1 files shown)


dannon:
When I added this I intentionally segmented it off into the "Other Login Options" column on equal footing (not below it on the page) to point out that there are two equal pathways for logging in.  It's in a column because we anticipate supporting multiple options here (so, multiple buttons in that right column stacked on top of one another).

(the change you made that only displays the contents when there are external login options obviously makes sense, but I don't like the layout change)

guerler:
It produces a lot of unused space as of right now and I think it looks better under the regular login form. When we add new OIDC providers we should probably just add their colored OIDC provider icons as links and replace the current button and title.

dannon:
> It produces a lot of unused space as of right now. It looks better under the regular login form.

@guerler completely disagree -- it's actually *more* compact to make use of columns like I did, and the intent was to present 'OR''d login options on equal footing on the page.  (and, since it's obviously subjective I'll present it as such -- it looks better the way it was...)
```
This
OR
This
```

Vs

```
This OR This
```

Viewing options the second way is more natural, I feel like.

guerler:
Actually it is "This or This" instead of "This or This or This". The first column is for login options and the second column shows the welcome page. Instead of having the first two for login and then a third one for the welcome page. In the latter case, the welcome page just becomes too squeezed in imo.

dannon:
The welcome page isn't a choice the user is making to move forward and log in, that's not what I'm talking about.  To log in you have two pathways, either internal or external, so the intent was to present them side by side.  Maybe the welcome should move.

I really don't like just hanging random buttons off the bottom of a card :/

guerler:
That is fine, we can revisit this and move the welcome page or add new providers and overhaul the whole thing, but right now, it just does not look right as it is.

I was thinking of something like: 
image

dannon:
The simplest possible welcome works when shunted into the column like that, but nothing else does.  Welcome should be beneath the two login options.  (obviously your second screenshot edited in up top looks janky)

guerler:
The welcome page is designed for the center column. I think there is plenty of space in the right column of the login page to accommodate the welcome page in a similar way and that space can easily be further increased.

dannon:
@guerler Right, and the center column is *HUGE* compared to half of the 1140 container width here.  I think the button just hanging out below the card looks worse, and that the whole 'shove the welcome in an equal column' only works with the simplest possible default splash we have now.

(this is probably something we'd both be better off just drawing out on a whiteboard instead of trying to figure out on a PR -- if you *really* like that random button I'll just merge this and we can figure it out down the road, it's not like I'm going to -1 it)

guerler:
Yeah, I prefer merging this because as of right now it clearly looks better and its a minimal change.

dannon:
@guerler You keep saying things like "it clearly looks better", but please consider other folks have opinions too.  It's clear you're designing around the 'show_welcome_with_login' option, but that's disabled by default and I'd bet almost nobody is using it right now.

guerler:
Ok, fair enough. It is definitely a subjective topic. Are you still fine with merging this and revisiting this another time?

PR #7404, opened 2019-02-22
Automatically set cookie_path using url_for()

natefoo:
This sets it with a trailing slash, the docs previously did not suggest that but it's probably more correct.

WIP because I'll update docs once #7216 is merged forward.


jmchilton:
Nate - do you mind if I rebase and merge this?

natefoo:
@jmchilton not at all

jmchilton:
Thanks @natefoo !

PR #7418, opened 2019-02-27
[19.01] Fix dependency installation in UI

mvdbeek:
We need to pass the install kwarg here.
Broken in https://github.com/galaxyproject/galaxy/commit/2d3ab4ce1205aec24664225d74b2a851cb94b2b0

### DIFF
--- a/lib/galaxy/tools/deps/views.py
+++ b/lib/galaxy/tools/deps/views.py
@@ -104,6 +104,7 @@ class DependencyResolversView(object):
         return list(removed_environments)
 
     def install_dependencies(self, requirements, **kwds):
+        kwds['install'] = True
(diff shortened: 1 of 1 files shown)


martenson:
thanks @mvdbeek !

nsoranzo:
@mvdbeek We can then remove `kwds['install'] = True` from https://github.com/galaxyproject/galaxy/commit/2d3ab4ce1205aec24664225d74b2a851cb94b2b0#diff-d67d082895c846a6e64465fea59344e7R256 , I suppose.

mvdbeek:
Yeah, that's right (and also why API installs didn't have the issue). Good enough to do it on dev only ?

nsoranzo:
Of course!

davebx:
Thanks @mvdbeek. This has been applied to and solved the issue on usegalaxy.org

PR #7419, opened 2019-02-27
Added data type gpr: Class and sniffer, Updated data type config file.

bensellak:


lib/galaxy/datatypes/test/test.gpr
```diff
@@ -0,0 +1,13 @@
+ATF       1.0
+8         5
+"Type=GenePix ArrayList V1.0"
```
nsoranzo, line 3: It looks like this is a GAL file, not GPR.

nsoranzo:
@bensellak No need to close a pull request if you just need to push more commits to it.

PR #7420, opened 2019-02-27
Adding Gpr data type : class Gpr and sniffer

bensellak:


lib/galaxy/datatypes/binary.py
```diff
@@ -126,6 +126,35 @@ def sniff(self, filename):
             return False
 
 
+class Gpr(Binary):
+
+    """Binary data in GPR, Axon Text File (ATF) format."""
```
nsoranzo, line 131: Can you add `Described at http://mdc.custhelp.com/app/answers/detail/a_id/18883#gpr` ?

lib/galaxy/datatypes/binary.py
```diff
@@ -126,6 +126,35 @@ def sniff(self, filename):
             return False
 
 
+class Gpr(Binary):
```
nsoranzo, line 129: By looking at the test file and the online description of the file format, it seems this can be a subclass of `Tabular` or at least `Text`, or am I missing something?

lib/galaxy/datatypes/test/test.gpr
```diff
@@ -0,0 +1,13 @@
+ATF       1.0
+8         5
+"Type=GenePix ArrayList V1.0"
```
nsoranzo, line 3: It looks like this is a GAL file, not GPR.

bensellak: GPR and GAL are quite the same, as well as ATF.

bensellak: For submission to microarrays databases like geo they use gpr format (http://mdc.custhelp.com/app/answers/detail/a_id/18883/kw/18883/session/L2F2LzEvdGltZS8xNTUxMjY5NTc2L3NpZC9MajhQX3E4bw%3D%3D#gpr). I quote from the link "ATF is a tab-delimited text file format that can be read by typical spreadsheet programs such as Microsoft Excel. It is used for GenePix Array List (GAL) files, and GenePix Results (GPR) files.".

lib/galaxy/datatypes/binary.py
bensellak: For sure, where? in the class or the data types config file? I put it in binary because it's the same as cell, both describe a microarray experiment result.

lib/galaxy/datatypes/test/test.gpr
nsoranzo: As far as I understanda, GPR and GAL are "subclasses" of ATF, but are not equivalent. For example, `BlockCount` is a header only for GAL, not GPR.

bensellak: You are right. The test file is there just for sniffing the header for the type. ATF is a format used to describe both GPR and GAL. The most used one is GPR and the only one my tools will be needing.A good GPR is joined, I didn't put it in the tests folder for the size sake.
[GSM103772_1.gpr.gz](https://github.com/galaxyproject/galaxy/files/2911034/GSM103772_1.gpr.gz)

lib/galaxy/datatypes/binary.py
nsoranzo: I meant in the class comment, something like:
```python
    """
    GenePix Results format, described at
    http://mdc.custhelp.com/app/answers/detail/a_id/18883#gpr
    """
```

> I put it in binary because it's the same as cell, both describe a microarray experiment result.

The `cel` is format is clearly binary, see e.g. `lib/galaxy/datatypes/test/test.cel`, while `gpr` is not. So, you can put the `Gpr` class either together with its parent class (which, as suggested above, should be `Tabular` or `Text`) or in a separate file for microarray datatypes.

lib/galaxy/datatypes/test/test.gpr
bensellak: Got it, I will close this pull, add your suggested modifications and pull back.
Thank you very much

Issue #7421, opened 2019-02-27
Extract workflow failed when a dataset has unicode character in its name


vloux:
Galaxy version 18.09.

To reproduce :
- Put a special character like "é" or "è" in the name of a dataset
- Extract the workflow from the history menu. The workflow is extracted without error
- When trying to edit the workflow an error ("Loading workflow failed") occurs

Removing the special character from the dataset name and re-extracting the workflow solve the problem.

Error in the logs :
```
File '/project/proteore/galaxy/lib/galaxy/managers/workflows.py', line 1014 in __set_default_label
if str(default_label).lower() not in ['input dataset', 'input dataset collection']:
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe9' in position 16: ordinal not in range(128)
```

Thanks !

mvdbeek:
Thanks @vloux, it looks like your environment doesn't have LC_ALL set (or set to ASCII). If that is the case python will fall back to ASCII as the default encoding, and that'll break a lot of things. You may want to set up the locales to `en_US.UTF-8` or any other UTF-8 encoding.

mvdbeek:
(if you want to check it works you can do `python -c "print(u'\xe9')"` and you should see:
```
é
```
You can simulate that error with `LC_ALL="ASCII" python -c "print(u'\xe9')"`:                                                                                                                                                          
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe9' in position 0: ordinal not in range(128)
```

mvdbeek:
Well, we can also force UTF-8 when loading the step dict, but you should really investigate the environment.

vloux:
Hello Marius,
thank you for your quick answer. Indeed the LC_ALL environment variable wasn't set on our Galaxy server environment. I tried to set it (via the GALAXY_LOCAL_ENV_FILE ), but it doesn't seem to resolve the problem ?
For information, I can reproduce the same bug on usegalaxy.org

mvdbeek:
[...]

PR #7422, opened 2019-02-27
[18.09] Fix workflow loading on systems with ASCII locale

mvdbeek:
Fixes https://github.com/galaxyproject/galaxy/issues/7421,
though it's very likely Galaxy won't play well when python
is falling back to ASCII as default encoding.

WorkflowStep.label is backed by a 255 character unicode string,
but tool_inputs is JSONType, which might have integer or float values
(in principle), so we play it safe by casting to unicode.

### DIFF
--- a/lib/galaxy/managers/workflows.py
+++ b/lib/galaxy/managers/workflows.py
@@ -1012,7 +1012,7 @@ class WorkflowContentsManager(UsesAnnotations):
         if not module.label and module.type in ['data_input', 'data_collection_input']:
             new_state = safe_loads(state)
             default_label = new_state.get('name')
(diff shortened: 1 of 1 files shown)


mvdbeek:
Surprisingly I could only reproduce https://github.com/galaxyproject/galaxy/issues/7421 by setting my LC_ALL to ASCII ... but on my production instance (as well as usegalaxy.*) this fails as well and they are correctly set up. MIght be worth looking into what's going on there.

nsoranzo (approved):

martenson:
thanks a bunch @mvdbeek !

PR #7428, opened 2019-02-28
Allow configuration location of user preferences extra configuration

hexylena:
Additionally moves the yaml parsing from runtime to boottime since it is not useful to re-load it every time the API is accessed.

### DIFF
--- a/lib/galaxy/config.py
+++ b/lib/galaxy/config.py
@@ -17,6 +17,7 @@ import sys
 import tempfile
 import threading
 import time
+import yaml
 from datetime import timedelta
 
 from six import string_types
@@ -642,6 +643,13 @@ class Configuration(object):
         self.biostar_never_authenticate = string_as_bool(kwargs.get('biostar_never_authenticate', False))
         self.pretty_datetime_format = expand_pretty_datetime_format(kwargs.get('pretty_datetime_format', '$locale (UTC)'))
         self.user_preferences_extra_config_file = kwargs.get('user_preferences_extra_conf_path', None)
+        try:
+            with open(self.user_preferences_extra_config_file, 'r') as stream:
+                self.user_preferences_extra = yaml.safe_load(stream)
+        except Exception:
+            log.warning('Config file (%s) could not be found or is malformed.' % self.user_preferences_extra_config_file)
+            self.user_preferences_extra = {'preferences': {}}
+
         self.default_locale = kwargs.get('default_locale', None)
         self.master_api_key = kwargs.get('master_api_key', None)
         if self.master_api_key == "changethis":  # default in sample config file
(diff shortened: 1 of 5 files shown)


martenson:
Probably also want to drop this then? https://github.com/galaxyproject/galaxy/pull/7428/files#diff-5e5bbef6f4c68752f8c1e486e8266d94R64

lib/galaxy/config.py
```diff
@@ -641,6 +641,14 @@ def __init__(self, **kwargs):
         self.biostar_enable_bug_reports = string_as_bool(kwargs.get('biostar_enable_bug_reports', True))
         self.biostar_never_authenticate = string_as_bool(kwargs.get('biostar_never_authenticate', False))
         self.pretty_datetime_format = expand_pretty_datetime_format(kwargs.get('pretty_datetime_format', '$locale (UTC)'))
+        self.user_preferences_extra_config_file = kwargs.get('user_preferences_extra_conf_path', None)
```
nsoranzo:
```suggestion
        self.user_preferences_extra_config_file = kwargs.get('user_preferences_extra_conf_path', 'config/user_preferences_extra_conf.yml')
```

nsoranzo (approved):

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -936,6 +936,14 @@ mapping:
           - $iso8601 (complete format string as specified by ISO 8601 international
             standard).
 
+      user_preferences_extra_conf_path:
+        type: str
+        default: 'config/user_preferences_extra_conf.yml'
+        required: false
+        desc: |
+          Location of the user_preferences_extra_conf.yml file, if not in the
+          standard configuration directory.
```
nsoranzo: I was wondering if here we should replace `user_preferences_extra_conf.yml` with a description, e.g.: `Location of the configuration file containing extra user preferences.`

And I think we can drop `if not in the standard configuration directory.`

hexylena: sounds good

nsoranzo (approved):

hexylena:
Circle-ci seems to have failed because it couldn't find a circle ci config in my ancient branch, not surprising.

nsoranzo:
> Circle-ci seems to have failed because it couldn't find a circle ci config in my ancient branch, not surprising.

@erasche See https://github.com/galaxyproject/galaxy/pull/7451 .

martenson:
>because it couldn't find a circle ci config in my ancient branch

The tests should run on a merge afaik (for all of our tests - travis, jenkins, circleci). If they don't that is a bug so please keep it in mind and report when it behaves wrongly.

Thanks @nsoranzo ! And thanks @erasche for the user pref PR.

PR #7430, opened 2019-02-28
Added Gal and Gpr data types

bensellak:
Gal and Gpr are both Tabular. Gpr contains a header name ("Rgn R²" utf-8 encode issue) that make set_meta fail. Fixed by inheriting from Binary.


bensellak:
@nsoranzo  I think it should be working fine, don't you?

nsoranzo:
@galaxybot test this

lib/galaxy/datatypes/binary.py
```diff
@@ -126,6 +126,34 @@ def sniff(self, filename):
             return False
 
 
+class Gpr (Binary):
+    """ Gpr File format described at:
+    http://mdc.custhelp.com/app/answers/detail/a_id/18883/kw/18883/session/L2F2LzEvdGltZS8xNTUxMjY5NTc2L3NpZC9MajhQX3E4bw%3D%3D#gpr
```
```suggestion
    http://mdc.custhelp.com/app/answers/detail/a_id/18883/#gpr
```

lib/galaxy/datatypes/tabular.py
```diff
@@ -1161,3 +1161,41 @@ def get_chunk(self, trans, dataset, chunk):
         ck_data_body = re.sub('[ ]+', '\t', ck_data_body)
 
         return dumps({'ck_data': util.unicodify(ck_data_header + "\n" + ck_data_body), 'ck_index': ck_index + 1})
+
+
+class Gal(Tabular):
+    """ Gal File format described at:
+        http://mdc.custhelp.com/app/answers/detail/a_id/18883/kw/18883/session/L2F2LzEvdGltZS8xNTUxMjY5NTc2L3NpZC9MajhQX3E4bw%3D%3D#gal
```
nsoranzo:
```suggestion
        http://mdc.custhelp.com/app/answers/detail/a_id/18883/#gal
```

lib/galaxy/datatypes/binary.py
```diff
@@ -126,6 +126,34 @@ def sniff(self, filename):
             return False
 
 
+class Gpr (Binary):
```
```suggestion
class Gpr(Binary):
```

nsoranzo: Also, not sure why this cannot also be a subclass of `Tabular`?

bensellak: It causes set_meta function to fail. There is a header name that causes this in the gpr format  as in the first commit comment. When i rename the header and do it in the tabular way it works just fine. But i will no be tempted to change some technology fromat. Multiple packages are using this format. It is not very satisfactory but works.

nsoranzo: Do you have a traceback for the set_meta fail?
Also, can you provide a link to the commit you mentioned?

bensellak: Check screenshots for the error  using upload tool from main panel: 
The error body:
  File "lib/galaxy/jobs/runners/local.py", line 152, in queue_job
    self._finish_or_resubmit_job(job_state, stdout, stderr, exit_code)
  File "lib/galaxy/jobs/runners/__init__.py", line 428, in _finish_or_resubmit_job
    job_state.job_wrapper.finish(stdout, stderr, exit_code, check_output_detected_state=check_output_detected_state)
  File "lib/galaxy/jobs/__init__.py", line 1374, in finish
    dataset.datatype.set_meta(dataset, overwrite=False)
  File "lib/galaxy/datatypes/tabular.py", line 337, in set_meta
    line = dataset_fh.readline()
  File "/opt/galaxy/.venv/lib/python2.7/codecs.py", line 314, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf8' codec can't decode byte 0xb2 in position 384: invalid start byte
screenshot from 2019-03-01 13-33-40
screenshot from 2019-03-01 13-32-59




[Test-Files.zip](https://github.com/galaxyproject/galaxy/files/2919436/Test-Files.zip)

lib/galaxy/datatypes/binary.py
```diff
@@ -126,6 +126,34 @@ def sniff(self, filename):
...
+
+        >>> fname = get_test_fname('test.gal')
+        >>> Gpr().sniff(fname)
+        False
+        """
+
+        try:
+            source = open(filename, 'rb').read(2000)
+            return b'ATF' in source and b'GenePix' in source and b'BlockCount' not in source
+        except Exception:
```
nsoranzo: No need to deal with exceptions in `sniff()`/`sniff_prefix()` methods, any exception raised is caught by the framework and is equivalent to returning `False`.

lib/galaxy/datatypes/binary.py
```diff
@@ -126,6 +126,34 @@ def sniff(self, filename):
...
 
+class Gpr (Binary):
+    """ Gpr File format described at:
+    http://mdc.custhelp.com/app/answers/detail/a_id/18883/kw/18883/session/L2F2LzEvdGltZS8xNTUxMjY5NTc2L3NpZC9MajhQX3E4bw%3D%3D#gpr
+    """
+    edam_format = "format_3829"
+    edam_data = "data_3110"
+    file_ext = "gpr"
+
+    def sniff(self, filename):
```
nsoranzo: I'd suggest you to use `sniff_prefix(self, file_prefix)` instead. Then you can deal with the already opened file `file_prefix` (of type `FilePrefix`), which should deal well with UTF-8 characters and has various convenient methods. See `lib/galaxy/datatypes/molecules.py` for many examples.
Please note that you would need to decorate your class with `@build_sniff_from_prefix` in that case.

bensellak: The problem is UTF-8 as shown in my previous message.

nsoranzo: To clarify my message above, I think you are seeing the UTF-8 error because of how you are opening the file. The appropriate way would be either:
```python
            source = io.open(filename, 'r', encoding='utf-8').read(2000)
```
or changing this method to `sniff_prefix()`, as described above.

bensellak:
@nsoranzo Thank you, now they are declared in a different file and as data.text.

bensellak:
@nsoranzo Should it be rebuilt?

bensellak:
@bgruening Added data types. Can this pull go through building again?

bensellak:
@galaxybot test this

bensellak:
@nsoranzo Need your help here please.

bensellak:
@nsoranzo can this be approved? I followed your suggestion and  made a py file specific to microarrays with Gpr and Gal inside. They are declared as Text and added more metadata to the dataset.

PR #4470, opened 2017-08-21
Nicer tool click targets for base + "plain" theme

hexylena:
Something quite useless I worked on over the weekend.

- nicer click targets for base theme. There's a slightly darker hover colour, and clicking anywhere in that area results in the tool link being activated. This is significantly nicer than trying to click a short tool name (sometimes 3-4 characters) vs a large, unclickable description (10-100 chars)
auswahl_032
- plain theme screenshots follow. Can be enabled by uncommenting the theme in galaxy_variables.less

auswahl_025
auswahl_027
auswahl_029
auswahl_030

PR #7437, opened 2019-03-01
Add tool hover

hexylena:
Originally in https://github.com/galaxyproject/galaxy/pull/4470 I added some CSS changes to enlarge the click targets and make hovering over tools darken their backgrounds a bit. (I know it switches to a hand, but with larger click targets not always clear which tool you're hovering over.) Apparently those css changes were never included, not even in the original PR, so I'm re-proposing them now.

image

the padding is moved from the tool title to the tool link to ensure there's no vertical space that doesn't lead to a tool.

image

### DIFF
--- a/client/galaxy/style/scss/base.scss
+++ b/client/galaxy/style/scss/base.scss
@@ -1545,8 +1545,6 @@ div.toolPanelLabel {
 }
 
 div.toolTitle {
-    padding-top: 5px;
-    padding-bottom: 5px;
     margin-left: 10px;
     margin-right: 10px;
     display: block;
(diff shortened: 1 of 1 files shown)


mvdbeek:
I like it!

hexylena:
The only open question, maybe @guerler has some ideas, is that if you set `$tool-panel-bg` to a dark colour, then the `darken()` operation is the wrong one to use. I've never seen anyone with a 'dark' themed galaxy, but I wouldn't want to rule out the possibility in the future.

martenson:
We have many places where we use `darken()`, and I assume most of these places would be broken if the base was dark. 

```
galaxy/style/scss/base.scss
406:                background: darken($brand-dark, 10%);
756:    border-color: darken(rgba(82, 168, 236, 0.8), 15%);
1199:    border-color: darken($state-default-border, 30%);
1201:    background: darken($state-default-bg, 30%);
1568:        background: darken($panel-bg-color, 5%);
1912:    color: darken($border-default-color, 20%);

galaxy/style/scss/ui.scss
66:            background: darken($state-success-bg, 30%);
```

This change is a good QoL improvement. Thanks @erasche !

Issue #7269, opened 2019-01-25
LocalShell runner fails on "No such file or directory"


vholer:
When `LocalShell` is directly used to run Slurm CLI commands, the execution of the commands fails on "no such file or directory"

```
galaxy.jobs.runners.cli DEBUG 2019-01-25 19:13:53,789 [p:3868,w:1,m:0] [ShellRunner.work_thread-0] (5) submitting file: /opt/galaxy/database/jobs_directory/000/5/galaxy_5.s
h
galaxy.jobs.runners ERROR 2019-01-25 19:13:53,964 [p:3868,w:1,m:0] [ShellRunner.work_thread-0] (5) Unhandled exception calling queue_job
Traceback (most recent call last):
  File "lib/galaxy/jobs/runners/__init__.py", line 113, in run_next
    method(arg)
  File "lib/galaxy/jobs/runners/cli.py", line 98, in queue_job
    returncode, stdout = self.submit(shell, job_interface, ajs.job_file, galaxy_id_tag, retry=MAX_SUBMIT_RETRY)
  File "lib/galaxy/jobs/runners/cli.py", line 130, in submit
    cmd_out = shell.execute(job_interface.submit(job_file))
  File "lib/galaxy/jobs/runners/util/cli/shell/local.py", line 47, in execute
    p = Popen(cmd, stdin=None, stdout=outf, stderr=PIPE)
  File "/usr/lib/python2.7/subprocess.py", line 394, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1047, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
```

as the `Popen` function call inside `LocalShell.execute` gets whole command including arguments as a single string `cmd` (e.g.: `'sbatch /opt/galaxy/database/jobs_directory/000/6/galaxy_6.sh'`):

https://github.com/galaxyproject/galaxy/blob/0b95564789c37e7b0aba2d32b9442e96d7cb786d/lib/galaxy/jobs/runners/util/cli/shell/local.py#L45-L47

PR #7438, opened 2019-03-01
[18.09] Fix LocalShellRunner

mvdbeek:
All the `cmd` coming from the plugins are strings, so I'm just doing
shlex.split here, which should be sufficient to bring this on par
with the Remote Shell plugins (which consume strings).
Should fix https://github.com/galaxyproject/galaxy/issues/7269.

### DIFF
--- a/lib/galaxy/jobs/runners/util/cli/shell/local.py
+++ b/lib/galaxy/jobs/runners/util/cli/shell/local.py
@@ -6,6 +6,8 @@ from subprocess import (
 from tempfile import TemporaryFile
 from time import sleep
 
+import six
+
 from ..shell import BaseShellExec
 from ....util import (
     Bunch,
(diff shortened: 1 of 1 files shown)


lib/galaxy/jobs/runners/util/cli/shell/local.py
```diff
@@ -37,12 +40,16 @@ class LocalShell(BaseShellExec):
...
     True
+    >>> shell.execute('echo hi').stdout == "hi\\n"
+    True
     """
 
     def __init__(self, **kwds):
         pass
 
     def execute(self, cmd, persist=False, timeout=DEFAULT_TIMEOUT, timeout_check_interval=DEFAULT_TIMEOUT_CHECK_INTERVAL, **kwds):
+        if isinstance(cmd, (six.string_types, six.text_type)):
```
nsoranzo:
```suggestion
        if isinstance(cmd, six.string_types):
```

lib/galaxy/jobs/runners/util/cli/shell/local.py
```diff
@@ -37,12 +40,16 @@ class LocalShell(BaseShellExec):
...
+            cmd = shlex.split(cmd)
         outf = TemporaryFile()
         p = Popen(cmd, stdin=None, stdout=outf, stderr=PIPE)
```
nsoranzo: Maybe replace these 4 lines with:
```python
        is_cmd_string = isinstance(cmd, six.string_types)
        outf = TemporaryFile()
        p = Popen(cmd, stdin=None, stdout=outf, stderr=PIPE, shell=is_cmd_string)
```

mvdbeek: Thanks, that is much better, I've rebased this with your suggestion.

nsoranzo (approved):

PR #7440, opened 2019-03-01
Allow webless toolbox watching.

jmchilton:
(for galaxy-lib, it can't import galaxy.web.stack stuff now that it imports a bunch of sqlalchemy stuff)

### DIFF
--- a/lib/galaxy/tools/toolbox/watcher.py
+++ b/lib/galaxy/tools/toolbox/watcher.py
@@ -15,7 +15,13 @@ except ImportError:
     can_watch = False
 
 from galaxy.util.hash_util import md5_hash_file
-from galaxy.web.stack import register_postfork_function
+try:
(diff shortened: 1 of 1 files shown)


lib/galaxy/tools/toolbox/watcher.py
```diff
@@ -15,7 +15,13 @@
...
 
 from galaxy.util.hash_util import md5_hash_file
-from galaxy.web.stack import register_postfork_function
+try:
+    from galaxy.web.stack import register_postfork_function
+except ImportError:
+
+    # noop, replacement for non-web environments
+    def register_postfork_function(f):
+        pass
```
mvdbeek, line 24: Shouldn't this match the signature of `register_postfork_function`, which is `f, *args, **kwargs`. Next question, shouldn't we immediately execute the function ?

jmchilton:
These are good points, I'll figure this out.

PR #7441, opened 2019-03-01
Ready History Import/Export for Revamp

jmchilton:
Simplifications, fixes, and refactoring to get ready for the big main commit in #7367. See individual commits for details.


jmchilton:
Some of these commits are in https://github.com/galaxyproject/galaxy/pull/7217 but there is one seemingly unrelated API test in that PR that I just can't get to pass on Jenkins but the much bigger branch #7367 doesn't have that problem so I'm curious to see what the tests look like for this.

Issue #7447, opened 2019-03-02
All workflow work stuck and constant Uncaught exception in exposed API method


alpapan:
Hello 

version: dev, 1 week old but tried with current before submitting ticket (just checked out ).

Part 1:
A couple of hours ago I tried to run a workflow that uses Trinity Abundance estimation with a collection 24 datasets. It never started the job but it stayed in the 'creating jobs' state:

![image](https://user-images.githubusercontent.com/608484/53682964-5e945180-3d4f-11e9-99c4-c4088fb90fd1.png)

This was a few hours ago. I tried selecting delete and then stopping the galaxy server process and restarted it but there was no difference no matter what I did.

Part 2:
Then i noticed that i cannot copy the history that contains this run. It reads : 

```
Uncaught exception in exposed API method
```

Part 3: 
Then i realised that I can run jobs but I cannot run or save any workflows for **any** user.
Here is an empty one with just the input dataset:

![image](https://user-images.githubusercontent.com/608484/53682986-b03cdc00-3d4f-11e9-9b0e-de627decb85d.png)

Importantly, no user can run any workflows. Restarting the server doesn't fix it.

Here is some part of the log file:
```
220.233.79.26 - - [03/Mar/2019:00:51:47 +1100] "GET /api/webhooks HTTP/1.1" 200 - "https://galaxy.gpi.westernsydney.edu.au/workflow/editor?id=79f88a7654569b51" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/72.0.3626.119 Safari/537.36"
galaxy.web.framework.decorators ERROR 2019-03-03 00:51:51,469 Uncaught exception in exposed API method:
Traceback (most recent call last):
  File "/scratch_cory/sysadmin/galaxy/lib/galaxy/web/framework/decorators.py", line 283, in decorator
    rval = func(self, trans, *args, **kwargs)
  File "/scratch_cory/sysadmin/galaxy/lib/galaxy/webapps/galaxy/api/workflows.py", line 545, in update
    **from_dict_kwds
  File "/scratch_cory/sysadmin/galaxy/lib/galaxy/managers/workflows.py", line 362, in update_workflow_from_raw_description
    **kwds
  File "/scratch_cory/sysadmin/galaxy/lib/galaxy/managers/workflows.py", line 420, in _workflow_from_raw_description
[...]

PR #7448, opened 2019-03-02
[18.09] Fix loading workflows with steps without default label

mvdbeek:
Broken in https://github.com/galaxyproject/galaxy/pull/7422.
Prior to https://github.com/galaxyproject/galaxy/pull/7422
we'd call str(default_label), so if `default_label` was None
we'd get 'None'. Fixes part of https://github.com/galaxyproject/galaxy/issues/7447


nsoranzo (approved):

martenson:
thanks for the quick followup @mvdbeek !

PR #7452, opened 2019-03-03
Restore get_filename to galaxy.tools.verify:verify()

jmchilton:
xref https://github.com/galaxyproject/galaxy-lib/pull/144

PRing here also to run Galaxy tests, manually verified galaxy-lib changes fix bugs in Planemo as a result of this being dropped.


mvdbeek (approved):
Looks good, thanks for figuring this out!

PR #7453, opened 2019-03-04
Fix doc for statsd conf for uwsgi?

abretaud:
Hi,
So I'm (finally) trying to add more monitoring to our galaxy server.
Looking at the doc, I think there is a small mistake in the uwsgi conf to push stats to statsd, so here's a little patch.
However, this doesn't work here since I get this error at startup:
`unable to find "statsd" stats_pusher`
I guess I need to do something like https://github.com/unbit/uwsgi/issues/773 but not sure what's the good way to do it with galaxy, should I manually uninstall+reinstall uwsgi weel?

### DIFF
--- a/doc/source/admin/special_topics/performance_tracking.rst
+++ b/doc/source/admin/special_topics/performance_tracking.rst
@@ -24,7 +24,22 @@ Or a StatsD server via:
       socket: ...
       stats-push: statsd:127.0.0.1:8125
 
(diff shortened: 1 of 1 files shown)


hexylena:
I found the setup through uwsgi's pusher too difficult, and just included it in [`gxadmin`](https://github.com/usegalaxy-eu/gxadmin#uwsgi-stats_influx) but if this fixes installation that's good!

abretaud:
Ah yes thanks, much easier with gxadmin, it works!
What about this change to mention it?

nsoranzo:
Not related, but would be nice to remove the mention about `paste` in the previous paragraph.

dannon:
Thanks @abretaud!

PR #7446, opened 2019-03-02
Use a lock on cleanup, expire_tool and cache_tool

mvdbeek:
This should fix https://github.com/galaxyproject/galaxy/issues/7444.

PR #7457, opened 2019-03-04
Adding microarrays data types Gpr and Gal 

bensellak:
Class definition in microarray.py file.


bensellak:
@martenson I closed the other pull. I didn't understand why it was failing. I reduced the data size to 2k.

lib/galaxy/datatypes/microarrays.py
```diff
@@ -0,0 +1,152 @@
...
+                    optional=True, no_value=0)
+
+    def set_peek(self, dataset, is_multi_byte=False):
+        if not dataset.dataset.purged:
+            if dataset.metadata.number_of_blocks == 1:
+                dataset.blurb = "1 block %s recodrs" % dataset.metadata.number_of_records
+                dataset.peek = get_n_first_lines(dataset.file_name, n=3)
+            else:
+                dataset.blurb = "%s blocks %s recodrs" % (dataset.metadata.number_of_blocks, dataset.metadata.number_of_records)
+                dataset.peek = get_n_first_lines(dataset.file_name, n=3)
```
nsoranzo:
```suggestion
            dataset.peek = get_file_peek(dataset.file_name)
```

lib/galaxy/datatypes/microarrays.py
```diff
@@ -0,0 +1,152 @@
...
+    MetadataElement(name="number_of_blocks", default=1, desc="Number of blocks", readonly=True, visible=True,
+                    optional=True, no_value=0)
+    MetadataElement(name="number_of_records", default=1, desc="Number of records", readonly=True, visible=True,
+                    optional=True, no_value=0)
+
+    def set_peek(self, dataset, is_multi_byte=False):
+        if not dataset.dataset.purged:
+            if dataset.metadata.number_of_blocks == 1:
+                dataset.blurb = "1 block %s recodrs" % dataset.metadata.number_of_records
+                dataset.peek = get_n_first_lines(dataset.file_name, n=3)
```
nsoranzo: This line is repeated in the else branch, so it can be removed if you apply my suggestion below.

lib/galaxy/datatypes/microarrays.py
```diff
@@ -0,0 +1,152 @@
...
+        Rendering n first lines in order to populate peek.
+    """
+    try:
+        cmd = ["head", '-' + str(n)]
+        cmd.extend([filename])
+        out = subprocess.Popen(cmd, stdout=subprocess.PIPE)
+        return out.communicate()[0]
+    except Exception:
+        pass
+    return 0
```
nsoranzo: I think you can remove this function and use `get_file_peek` , which can be imported as:
```
from galaxy.datatypes.data import get_file_peek
```

lib/galaxy/datatypes/microarrays.py
```diff
@@ -0,0 +1,152 @@
...
+        >>> fname = get_test_fname('test.gal')
+        >>> Gal().sniff(fname)
+        True
+        >>> fname = get_test_fname('test.gpr')
+        >>> Gal().sniff(fname)
+        False
+        """
+        found_genepix = count_special_lines("GenePix", filename) >= 1
+        found_atf = count_special_lines("ATF", filename) >= 1
+        found_blockcount = count_special_lines("BlockCount", filename) >= 1
```
nsoranzo: I fear that forking three grep subprocesses when sniffing is going to be a big performance hit.

Can you please have a look at how sniffing is done using the `sniff_prefix` method in `lib/galaxy/datatypes/molecules.py` ?

nsoranzo (commented):
Thanks for your work on this!

lib/galaxy/datatypes/microarrays.py
bensellak: I did look at molecules file very deeply. The get_peek doesn't support well, mixed encoding files. The case of Gpr, it contains a special character that io.string fails to deal with. I committed the changes you suggested, but the test will probably fail.

bensellak: @nsoranzo Gpr is an ISO-8859 and Gal is ASCII.

nsoranzo:
@galaxybot test this

bensellak:
@nsoranzo As predicted UnicodeDecodeError: 'utf8' codec can't decode byte 0xb2 in position 1303: invalid start byte. And file_prefix wont do the job either. And the famouse 0xb2  comes from ² in "Rgn R²". This is clearly a galaxy issue for not handling this kind of stuff. You can take a sample txt, add ² to it and it wont be uploaded. get peek is trying to convert it to ascii/utf-8 and fails.

bensellak:
I think get_file_peek is poorly coded. Doesn't take in count mode or multibyte files.

bensellak:
The test will pass but the peek will be "binary file".

bensellak:
Thank you @nsoranzo  for your help, very instructive. Would it be merged?

martenson:
@galaxybot test this

lib/galaxy/datatypes/microarrays.py
nsoranzo: ISO-8859 is a legacy encoding, I've locally converted the GPR test file from ISO-8859-1 to UTF-8 using `recode` (see e.g. https://linux-tips.com/t/converting-files-between-various-character-sets-and-usages/161 ) and this fixes the tests.

Do you think that's something acceptable for your use case? If yes, I can upload here the converted test file.

bensellak:
@nsoranzo Is it fine? I would be happy to see it merged. I have a bunch of tools to upload to iuc and this is blocking the process.

lib/galaxy/datatypes/microarrays.py
bensellak: Better leave it as it is. The tools i built will not convert, they need just this type to be detected by galaxy while filtering inputs. But great idea thank you

bensellak:
@martenson @martenson I don't get the meaning of the error. Any explanation will be very appreciated.

mvdbeek:
You can ignore this.

bensellak:
Thank you @mvdbeek

lib/galaxy/datatypes/microarrays.py
mvdbeek: Then the tests won't pass at this point, and we can't merge it. It is an issue we'll think about and work on, but I'd suggest converting the test files to a standard encoding for the time being. I also agree with @nsoranzo, we'll not want to run external commands during sniffing.

bensellak: @nsoranzo has already converted the files (Thanks for that). I can get rid of the external then. thank you both @mvdbeek  and @nsoranzo.

mvdbeek: Are you sure? test.gpr still contains "Rgn R�" on line 29, which looks like a non-UTF8 character.

nsoranzo: I've converted it on my laptop, I'll push it later.

nsoranzo:
@galaxybot test this

lib/galaxy/datatypes/microarrays.py
nsoranzo: > I fear that forking three grep subprocesses when sniffing is going to be a big performance hit.
>
> Can you please have a look at how sniffing is done using the sniff_prefix method in `lib/galaxy/datatypes/molecules.py` ?

This is still an issue, can you have another look now that the test file is fixed?

bensellak: I will update it tomorrow morning

bensellak:
@nsoranzo I think we are good now.

lib/galaxy/datatypes/microarrays.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def set_meta(self, dataset, **kwd):
+        """
+        Set metadata for Gal file.
+        """
+        super(Gal, self).set_meta(dataset, **kwd)
+        headers = iter_headers(dataset.file_name, sep="\t", count=5)
+        for count, line in enumerate(headers):
+            if count == 0:
+                dataset.metadata.file_format = str(line[0])
```
nsoranzo: I think you can remove the cast to `str` here and below (6 times).

bensellak: Done

bensellak: @nsoranzo any idea why the get code and test fails?.

nsoranzo:
@galaxybot test this

lib/galaxy/datatypes/microarrays.py
nsoranzo: It's an old issue that has been fixed recently in https://github.com/galaxyproject/galaxy/pull/7446 , which was merged after you opened this pull request. Bottom line is: don't worry about it.

nsoranzo:
@galaxybot test py3

mvdbeek:
@galaxybot test this

bensellak:
@nsoranzo Could we merge it.

nsoranzo (approved):

nsoranzo:
@galaxybot test this

mvdbeek:
Thanks a lot @bensellak and @nsoranzo!

bensellak:
Thanks guys @nsoranzo,  @mvdbeek and @martenson .

PR #7400, opened 2019-02-21
ENH: multifile zip testing

thermokarst:
This changeset introduces the ability to individually test zip archive members --- useful for working with multi-file zip archives (_ahem_, looking at you, @qiime2).

```xml
<has_archive_member path="testdir1/dir1/file3">
    <has_line_matching expression="subdirfile" />
</has_archive_member>
```

This PR attempts to emulate the changes introduced in https://github.com/galaxyproject/galaxy/pull/7301 (detailed testing of HDF5 attributes).

I am open to [other alternatives](https://github.com/galaxyproject/galaxy/pull/7297#issuecomment-459973607) for solving this problem, but given the discussions in #7297 & #7301, it _seems_ like this might be the "right" way to go. Anyway, thanks for entertaining this idea, look forward to chatting more about it!

ccing @ebolyen

PR #7462, opened 2019-03-04
fix syntax of xsd tool schema

martenson:
found by @nsoranzo


martenson:
@nsoranzo I suspect circleci did not test this since it tested this commit for `martenson/galaxy` repo (https://circleci.com/workflow-run/c8e35b49-5fb2-43f2-b82b-75e7cb6080ba). Still annoying though since it does not show the results here.

nsoranzo:
@martenson See https://bioconda.github.io/faqs.html#circlecimacos , I think we should disable building on our forks.

martenson:
@nsoranzo I disabled builds on my fork. For this reason: https://twitter.com/drazisil/status/1102704259718672386

Maybe circleci will be generous to us as well, we should ask them for macos allocation.

thermokarst:
Thanks @nsoranzo & @martenson! I didn't realize that I needed to specify my new `complexType` as a top-level type, rather than inline (as done in #7400). Thanks for fixing this!

martenson:
@thermokarst fwiw I think your spec was xsd-valid, just our parsing does expect it like this

PR #7463, opened 2019-03-05
Update to PSA 3.1 and support login with Globus identity

VJalili:
- Updates to current latest PSA version (3.1.0);
- Make some backend configuration generic;
- Support three new OIDC-based backends: Globus, Elixir, and PixelPin

### DIFF
--- a/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
+++ b/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
@@ -1,7 +1,7 @@
 -i https://wheels.galaxyproject.org/simple
 --extra-index-url https://pypi.python.org/simple
 adal==1.2.1
-amqp==2.4.1
+amqp==2.4.2
 appdirs==1.4.3
 asn1crypto==0.24.0
 azure-common==1.1.14
@@ -21,10 +21,10 @@ babel==2.6.0
 bagit==1.6.4
 bcrypt==3.1.6
 bdbag==1.4.1
-beaker==1.10.0
+beaker==1.10.1
 bioblend==0.12.0
 bleach==3.1.0
-boltons==19.0.1
+boltons==19.1.0
 boto3==1.7.84
 boto==2.49.0
 botocore==1.10.84
@@ -33,7 +33,7 @@ bx-python==0.8.2
 bz2file==0.98 ; python_version < '3.3'
 cachetools==3.1.0
 certifi==2018.11.29
-cffi==1.12.1
+cffi==1.12.2
 chardet==3.0.4
 cheetah3==3.2.0
 cliff==2.14.0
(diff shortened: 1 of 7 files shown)


lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
```diff
@@ -82,34 +83,31 @@ netifaces==0.10.9
...
-pbr==5.1.2
+pbr==5.1.3
 prettytable==0.7.2
 psutil==5.5.1
 pulsar-galaxy-lib==0.8.3
 pyasn1==0.4.5
 pycparser==2.19
 pycryptodome==3.7.3
-pycryptodomex==3.7.3
-pyinotify==0.9.6 ; sys_platform != 'win32' and sys_platform != 'darwin' and sys_platform != 'sunos5'
```
nsoranzo: This line should **not** be removed (pyinotify is still needed on Linux). But that's not what causing the build failure, I'm not sure what's going on there.

VJalili: I guess that was removed by the update-dependencies command; not sure what caused it though.

nsoranzo: I guess it's because you run `make update-dependencies` under macOS, where it's not needed.

Unfortunately, due to the fact that historically the list of Python dependencies installed by `pip` can be modified in `setup.py`, `pipenv` has to install them all to generate the `requirements.txt` files. So these files may change under different operating system or Python versions.

VJalili: So you recommend not running `make update-dependencies` under macOS?

nsoranzo: I'd not go that far, but deletions have to be checked.

VJalili: Sounds good, thank you for the clarification.

VJalili:
@nsoranzo I think `sqlalchemy==1.3.0` was the change causing tests failure; maybe some backward incompatible change in newer version of sqlalchemy?! I reverted this change and tests pass locally.

VJalili:
There are issues with the Elixir backend of PSA as it does not include ID token in its authentication response; hence, it will be supported after that problem is resolved (probably from within PSA). 

See: https://github.com/python-social-auth/social-core/issues/327

martenson:
thanks @VJalili

VJalili:
Thanks for reviewing this @martenson

nsoranzo:
> @nsoranzo I think sqlalchemy==1.3.0 was the change causing tests failure; maybe some backward incompatible change in newer version of sqlalchemy?! I reverted this change and tests pass locally.

Thanks for the heads-up!

lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
```diff
@@ -82,34 +83,32 @@ netifaces==0.10.9
...
-paste==3.0.6
+paste==3.0.7
 pastedeploy==2.0.1
-pastescript==3.0.0
-pbr==5.1.2
+pastescript==3.1.0
+pbr==5.1.3
 prettytable==0.7.2
-psutil==5.5.1
+psutil==5.6.0
```
nsoranzo, line 106: This one needs new wheels: https://github.com/galaxyproject/starforge-recipes/pull/40

Issue #5661, opened 2018-03-08
Feature request: greyout conditional outputs in workflow editor


scholtalbers:
Some tools have conditional outputs, however all are displayed (with for users cryptic names). It would help a lot to hide/greyout outputs that will not be generated. Currently workflow steps on non generated outputs will just silently be ignored and the user has to guess what was wrong. A particular example is [FeatureCounts](https://github.com/galaxyproject/tools-iuc/blob/master/tools/featurecounts/featurecounts.xml), where 4 different output formats can be selected - resulting in >4 selectable outputs in the workflow editor: 

<img width="753" alt="screen shot 2018-03-08 at 10 59 45" src="https://user-images.githubusercontent.com/102411/37144961-dc530068-22bf-11e8-9eb6-0ac47af9a7d8.png">

hexylena:
xref my original https://github.com/galaxyproject/galaxy/issues/1418 and your bug report that looks like it as well https://github.com/galaxyproject/galaxy/issues/5322

scholtalbers:
@erasche thanks for making those connections! (sometimes one uses the wrong search terms to locate duplicate issues :))

hexylena:
No problem! I knew I'd filed a similar issue but couldn't find it under the same search terms either, ended up going through all issues I'd filed. I should've used 'conditional', like, anywhere in the text. I end up commenting on that issue every time it happens to me which is starting to be a lot...

hexylena:
The eventual workflow scheduling error also shows up as a bug report in sentry. E.g.https://sentry.denbi.uni-freiburg.de/share/issue/47ca0a18767a44bda21dfb057b101126/ This is pretty annoying as it is a user-facing error rather than something a sysadmin should see. Whenever we fix this/1418/5322 we should be sure that that exception is caught and displayed to the user rather than logged in sentry. I've got 300 events for this error, surely main has even more.

mvdbeek:
We're now (tip of 19.01) filtering out outputs that aren't going to be produced in the workflow editor, so while you can still run into this with old workflows it should be harder to create broken workflows now. If a workflow doesn't run now you can open them up in the workflow editor and check that all steps are properly connected. The plan is still to prevent submission of these workflows, at which point we could actually grey out conditional outputs but still offer to connect outputs that are conditional, for workflows that are intentionally designed to have these optional branches of execution.
Thanks a lot for reporting this @scholtalbers and @erasche, I hope this will cause much less pain now.

Issue #1724, opened 2016-02-13
Workflows silently drop steps due to connection issues


hexylena:
```
galaxy.workflow.run ERROR 2016-02-13 11:13:49,193 Failed to schedule Workflow[id=884,name=BICH464 PAP 2016 Part C - Functional Prediction + WA], proble
m occurred on WorkflowStep[index=29,type=tool].
Traceback (most recent call last):
  File "/galaxy/lib/galaxy/workflow/run.py", line 162, in invoke
    jobs = self._invoke_step( step )
  File "/galaxy/lib/galaxy/workflow/run.py", line 228, in _invoke_step
    jobs = step.module.execute( self.trans, self.progress, self.workflow_invocation, step )
  File "/galaxy/lib/galaxy/workflow/modules.py", line 1159, in execute
    visit_input_values( tool.inputs, execution_state.inputs, callback, no_replacement_value=NO_REPLACEMENT )
  File "/galaxy/lib/galaxy/tools/parameters/__init__.py", line 31, in visit_input_values
    visit_input_values( input.inputs, d, callback, new_name_prefix, new_label_prefix, no_replacement_value=no_replacement_value )
  File "/galaxy/lib/galaxy/tools/parameters/__init__.py", line 31, in visit_input_values
    visit_input_values( input.inputs, d, callback, new_name_prefix, new_label_prefix, no_replacement_value=no_replacement_value )
  File "/galaxy/lib/galaxy/tools/parameters/__init__.py", line 37, in visit_input_values
    visit_input_values( input.cases[current].inputs, values, callback, new_name_prefix, label_prefix, no_replacement_value=no_replacement_value )
  File "/galaxy/lib/galaxy/tools/parameters/__init__.py", line 47, in visit_input_values
    prefixed_label=label_prefix + input.label )
  File "/galaxy/lib/galaxy/workflow/modules.py", line 1152, in callback
    replacement = progress.replacement_for_tool_input( step, input, prefixed_name )
  File "/galaxy/lib/galaxy/workflow/run.py", line 274, in replacement_for_tool_input
    replacement = [ self.replacement_for_connection( c ) for c in connection ]
  File "/galaxy/lib/galaxy/workflow/run.py", line 303, in replacement_for_connection
    raise Exception( message )
Exception: Workflow evaluation problem - failed to find output_name gff3_output_file in step_outputs {'output': <galaxy.model.HistoryDatasetAssociation
 object at 0x7f9021efbf10>}
[...]

PR #7466, opened 2019-03-05
Don't display outputs that aren't going to be produced in workflow editor

mvdbeek:
This obviously doesn't work (yet) with filters that work on the input dataset, but this already removes a lot of outputs that are not going to be produced based on static options. Since exceptions default to producing the dataset this isn't much harm.

output_filter

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/workflows.py
+++ b/lib/galaxy/webapps/galaxy/api/workflows.py
@@ -24,6 +24,7 @@ from galaxy.managers import (
 )
 from galaxy.model.item_attrs import UsesAnnotations
 from galaxy.tools.parameters import populate_state
+from galaxy.tools.parameters.basic import workflow_building_modes
 from galaxy.util.sanitize_html import sanitize_html
 from galaxy.web import _future_expose_api as expose_api
 from galaxy.web import _future_expose_api_anonymous_and_sessionless as expose_api_anonymous_and_sessionless
@@ -561,6 +562,7 @@ class WorkflowsAPIController(BaseAPIController, UsesStoredWorkflowMixin, UsesAnn
         Builds module models for the workflow editor.
         """
         inputs = payload.get('inputs', {})
+        trans.workflow_build_mode = workflow_building_modes.ENABLED
         module = module_factory.from_dict(trans, payload)
         if 'tool_state' not in payload:
             module_state = {}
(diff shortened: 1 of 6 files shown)


mvdbeek:
Fixes https://github.com/galaxyproject/galaxy/issues/5661 and https://github.com/galaxyproject/galaxy/issues/1724, except for when the filter is determined using an input dataset.

dannon:
Great idea, and way more concise to implement than I'd have thought.

mvdbeek:
Yeah, I was also expecting that it might be harder than that ...

jmchilton:
Marvelous!

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -561,6 +562,7 @@ def build_module(self, trans, payload={}):
         Builds module models for the workflow editor.
         """
         inputs = payload.get('inputs', {})
+        trans.workflow_build_mode = workflow_building_modes.ENABLED
```
nsoranzo, line 565: @mvdbeek Did you mean `trans.workflow_building_mode` ?

test/api/test_workflows.py
```diff
@@ -491,6 +491,50 @@ def test_export_editor(self):
                 'label',
             )
 
+    @skip_without_tool('output_filter')
```
nsoranzo, line 494:
```suggestion
    @skip_without_tool('output_filter_with_input')
```

test/api/test_workflows.py
```diff
@@ -491,6 +491,50 @@ def test_export_editor(self):
...
+        assert outputs[1]['name'] == 'out_3'
+        workflow_id = self._upload_yaml_workflow(template.format(produce_out_1='true', filter_text_1='foo'))
+        downloaded_workflow = self._download_workflow(workflow_id, style="editor")
+        outputs = downloaded_workflow['steps']['0']['outputs']
+        assert len(outputs) == 3
+        assert outputs[0]['name'] == 'out_1'
+        assert outputs[1]['name'] == 'out_2'
+        assert outputs[2]['name'] == 'out_3'
+
+    @skip_without_tool('output_filter_with_input')
```
nsoranzo, line 525:
```suggestion
    @skip_without_tool('output_filter_exception_1')
```

lib/galaxy/webapps/galaxy/api/workflows.py
mvdbeek: Exactly, and was half expecting this wouldn't work because we may also use this when building the workflow run form. Which it may not after all. Thanks for checking @nsoranzo !

PR #7459, opened 2019-03-04
set_metadata "unit" tests

jmchilton:
Baseline for a lot of testing for new and branched behavior in https://github.com/galaxyproject/galaxy/pull/7058.

PR #4437, opened 2017-08-16
Enhanced tool options for dataset discovery

jmchilton:
## Changes 

- Redo the structure of ``galaxy.json`` for profile >= ``17.09`` tools with two new test tools to demonstrate the new structure - for both output dataset metadata and discovered dataset metadata ([tool_provided_metadata_5.xml](https://github.com/jmchilton/galaxy/blob/0df820586b83b00cde01220e8eb2386a9c4b6794/test/functional/tools/tool_provided_metadata_5.xml) and [tool_provided_metadata_6.xml](https://github.com/jmchilton/galaxy/blob/0df820586b83b00cde01220e8eb2386a9c4b6794/test/functional/tools/tool_provided_metadata_6.xml) respectively). a73a0275a9d4ae2afa628c0fafcffb06d56b16b0
- Allow newer tools to specify the older style format for galaxy.json can be used - with XSD update and a test tool to demonstrate this ([tool_provided_metadata_8.xml](https://github.com/jmchilton/galaxy/blob/0df820586b83b00cde01220e8eb2386a9c4b6794/test/functional/tools/tool_provided_metadata_8.xml)). This can be specified by setting the ``provided_metadata_style`` on the ``outputs`` element for a tool - valid values include ``legacy`` and ``default``. 0df820586b83b00cde01220e8eb2386a9c4b6794
- When using the new ``galaxy.json structure`` - allow discovering datasets directly from ``galaxy.json`` instead of needing to specify file name patterns - including XSD update and test tools to demonstrate for new primary datasets and for dynamic collection contents ([tool_provided_metadata_7.xml](https://github.com/jmchilton/galaxy/blob/ec375376fde1fd57ce328e196455a28632e4f7f5/test/functional/tools/tool_provided_metadata_7.xml) and [collection_creates_dynamic_nested_from_json.xml](https://github.com/jmchilton/galaxy/blob/ec375376fde1fd57ce328e196455a28632e4f7f5/test/functional/tools/collection_creates_dynamic_nested_from_json.xml) respectively). ec375376fde1fd57ce328e196455a28632e4f7f5
[...]

PR #7470, opened 2019-03-05
Implement portable metadata generation.

jmchilton:
This PR starts by implementing a new metadata generation interface (in both the narrow and broad sense of this word). This variant of external metadata generation that doesn't use command-line arguments or absolute paths and simply takes operates on the current, job working directory and knows what to do.

- Uses relative paths allowing the directory to be staged elsewhere via Pulsar, Condor, or Docker for instance.
- Avoids DB objects for hard-coded paths making everything more robust.
- Avoids huge command-lines which for huge collection reduction jobs might hit system limits for command-line length.
- Uses JSON dicts to communicate parameters - meaning we can avoid awkward argument parsing as this evolves (see old ``max_metadata_value_size`` handling in ``set_metadata.py`` for instance).
- Uses output names instead of database IDs to key files - so file naming is consistent and human readable.

This builds on the new abstraction around metadata generation introduced in #7158 to create a new implementation of ``MetadataCollectionStrategy`` called ``PortableDirectoryMetadataGenerator``. This also builds on unit testing introduced in #7459.

The second commit in this PR improves the correctness of metadata generation in the context of tool provided metadata by ensuring ``set_metadata.py`` uses the ``ToolProvidedMetadata`` abstraction introduced in #4437 and enhanced for this context in #7156. There were (and still are) things in ``set_metadata.py`` that assume the old style ``galaxy.json`` but using this abstraction sets the stage for proper external metadata collection of newer style ``galaxy.json`` files.

This PR also brings in a fix for the legacy ``ToolProvidedMetadata`` implementation for behavior defined previously only in ``set_metadata.py``.

All of this work is from https://github.com/galaxyproject/galaxy/pull/7058 (tracked on board https://github.com/galaxyproject/galaxy/projects/11). In that context, this new metadata approach is extended even further to deal with job finish tasks including object store interactions as part of the Galaxy job script - setting the stage for data outputs that are truly remote from Galaxy.

### DIFF
--- a/lib/galaxy/metadata/__init__.py
+++ b/lib/galaxy/metadata/__init__.py
@@ -99,6 +99,7 @@ class PortableDirectoryMetadataGenerator(MetadataCollectionStrategy):
                                 job_metadata=None, compute_tmp_dir=None,
                                 include_command=True, max_metadata_value_size=0,
                                 kwds=None):
+        assert job_metadata, "setup_external_metadata must be supplied with job_metadata path"
         kwds = kwds or {}
         tmp_dir = _init_tmp_dir(tmp_dir)
 
--- a/lib/galaxy/metadata/__init__.py
+++ b/lib/galaxy/metadata/__init__.py
@@ -17,9 +17,15 @@ from galaxy.util import in_directory
 
 log = getLogger(__name__)
 
+SET_METADATA_SCRIPT = 'from galaxy_ext.metadata.set_metadata import set_metadata; set_metadata()'
+
 
 def get_metadata_compute_strategy(app, job_id):
-    return JobExternalOutputMetadataWrapper(job_id)
+    metadata_strategy = app.config.metadata_strategy
+    if metadata_strategy == "legacy":
+        return JobExternalOutputMetadataWrapper(job_id)
+    else:
+        return PortableDirectoryMetadataGenerator(job_id)
 
 
 @six.add_metaclass(abc.ABCMeta)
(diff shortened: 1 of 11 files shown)


mvdbeek:
https://jenkins.galaxyproject.org/job/docker-framework/14480/testReport/junit/test.functional/test_toolbox_pytest/test_tool_output_auto_format_test_1_/ looks related

jmchilton:
Indeed - with some more debugging I got:

```
ERROR:galaxy_ext.metadata.set_metadata:Problem sniffing datatype.
  File "/Users/john/workspace/galaxy/lib/galaxy_ext/metadata/set_metadata.py", line 48, in set_meta_with_tool_provided
    extension = sniff.handle_uploaded_dataset_file(dataset_instance.dataset.external_filename, datatypes_registry)
  File "/Users/john/workspace/galaxy/lib/galaxy/datatypes/sniff.py", line 720, in handle_uploaded_dataset_file
    return handle_uploaded_dataset_file_internal(*args, **kwds)[0]
  File "/Users/john/workspace/galaxy/lib/galaxy/datatypes/sniff.py", line 757, in handle_uploaded_dataset_file_internal
    guessed_ext = guess_ext(converted_path, sniff_order=datatypes_registry.sniff_order, is_binary=is_binary)
AttributeError: 'NoneType' object has no attribute 'sniff_order'
```

Looks like that isn't getting initialized properly anymore.

jmchilton:
Rebased with a fix for that error.

lib/galaxy/metadata/__init__.py
```diff
@@ -217,20 +287,17 @@ def __get_filename_override():
             # return command required to build
             fd, fp = tempfile.mkstemp(suffix='.py', dir=tmp_dir, prefix="set_metadata_")
             metadata_script_file = abspath(fp)
-            os.fdopen(fd, 'w').write('from galaxy_ext.metadata.set_metadata import set_metadata; set_metadata()')
+            os.fdopen(fd, 'w').write(SET_METADATA_SCRIPT)
```
mvdbeek: I think we'll want to close the descriptor here.

lib/galaxy_ext/metadata/set_metadata.py
```diff
@@ -70,8 +71,65 @@ def set_meta_with_tool_provided(dataset_instance, file_dict, set_meta_kwds, data
...
+        filename_kwds = os.path.join("metadata/metadata_kwds_%s" % output_name)
+        filename_out = os.path.join("metadata/metadata_out_%s" % output_name)
+        filename_results_code = os.path.join("metadata/metadata_results_%s" % output_name)
+        override_metadata = os.path.join("metadata/metadata_override_%s" % output_name)
+        dataset_filename_override = output_dict["filename_override"]
+
+        # Same block as below...
+        set_meta_kwds = stringify_dictionary_keys(json.load(open(filename_kwds)))  # load kwds; need to ensure our keywords are not unicode
+        try:
+            dataset = cPickle.load(open(filename_in, 'rb'))  # load DatasetInstance
```
mvdbeek, line 112: Unrelated comment: I think pickling might bite us if we really start doing portable metadata generation (plus it's a performance hit I think). We just need very little from the dataset instance, it just needs to be compatible for the set_meta methods in the dataype framework, right ? We could mock this out, which would also allow us to do metadata setting without the model / with a very minimal model.

mvdbeek (approved):

lib/galaxy_ext/metadata/set_metadata.py
jmchilton: Once my history import/export rewrite branch that introduces the concept of a JSON model store merges with my remote metadata branch - we will be using json, not pickles for remote metadata.

jmchilton: I guess this is a slightly different context though - still here I agree it would be good to do this. I’ll try to keep this in mind.

lib/galaxy/metadata/__init__.py
jmchilton: Added to tracking branch (https://github.com/galaxyproject/galaxy/pull/7058/commits/c4da5abbe3fbce7eb058cc3ba9aeb415899b3a35).

mvdbeek:
Hmm, there's still a relevant error in https://jenkins.galaxyproject.org/job/docker-api/14864/testReport/junit/test.api.test_histories/HistoriesApiTestCase/test_import_metadata_regeneration/, it looks like:
```
galaxy.jobs ERROR 2019-03-07 19:58:51,543 Unable to cleanup job 35
  File "/galaxy/lib/galaxy/jobs/__init__.py", line 1523, in cleanup
    galaxy.tools.imp_exp.JobImportHistoryArchiveWrapper(self.app, self.job_id).cleanup_after_job()
  File "/galaxy/lib/galaxy/tools/imp_exp/__init__.py", line 159, in cleanup_after_job
    hda, new_history, jiha.job
  File "/galaxy/lib/galaxy/tools/__init__.py", line 2266, in regenerate_imported_metadata_if_needed
    history.id, job.user, incoming={'input1': hda}, overwrite=False
  File "/galaxy/lib/galaxy/tools/actions/metadata.py", line 93, in execute_via_app
    kwds={'overwrite': overwrite})
  File "/galaxy/lib/galaxy/metadata/__init__.py", line 131, in setup_external_metadata
    "job_metadata": job_relative_path(job_metadata),
  File "/galaxy/lib/galaxy/metadata/__init__.py", line 109, in job_relative_path
    path_relative = os.path.relpath(path, tmp_dir)
  File "/galaxy_venv/lib/python2.7/posixpath.py", line 428, in relpath
    raise ValueError("no path specified")
ValueError: no path specified
```

PR #7476, opened 2019-03-06
correct virtualenv download with no wget/curl

hmenager:
This is a small correction to the embedded python code in a startup script that downloads virtualenv. This code could not work in the no curl/wget available case, because of a smallish python error.
Obviously, to test it you need to try installing it with a very bare bones installation (no wget nor curl)...


dannon:
Good catch!

PR #7477, opened 2019-03-06
added option to pass in individual globs to unit test watcher

Nerdinacan:
Just a little tiny addition to the client side unit-testing script.

This lets you watch individual tests as you're developing by passing them in a command line arg to the npm test-watch script.

```
npm run test-watch watch-only="Tags/*.test.js,store/tagStore.test.js"
```
... will run karma and watch your tests with the following file globs (only).

**/Tags/*.test.js
**/store/tagStore.test.js

### DIFF
--- a/client/karma/karma.config.mocha.js
+++ b/client/karma/karma.config.mocha.js
@@ -4,42 +4,56 @@
  * Individual files can be run by passing in a comma-delimited list
  * of globs for the karma config like this:
  * 
- *    env KARMA_TESTS="Tags.test.js,something.js,doodads.js" npm run test-watch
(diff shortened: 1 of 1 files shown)


dannon:
Great idea.  Not a major issue, but it seems might be more consistent with the rest of Galaxy's test framework to swap this from using an env var to cli args to specify test globs.  What do you think?

Nerdinacan:
> Great idea. Not a major issue, but it seems might be more consistent with the rest of Galaxy's test framework to swap this from using an env var to cli args to specify test globs. What do you think?

Ahh sure I can play with that a bit, I'll see how easy it is to make NPM eat explicit args in its scripts.

dannon:
Nice, it's working well for me!

Issue #5860, opened 2018-04-05
improve UX of the tour list


martenson:
We have 3 tours on Main, but nobody can tell without reading the whole page.

![screenshot 2018-04-05 15 04 48](https://user-images.githubusercontent.com/1814954/38386461-bd6c209c-38e2-11e8-99db-90b289d8f94c.png)

dannon:
Yeah, agreed.  We started out with a simple list, added this interface for category sorting with the intent to follow up on it.  Not sure if there's already another issue, though, so thanks for making this one!

selten:
I'll take a stab at it. I'm kinda responsible for this mess anyway.

jxtx:
Simple list with tags and "faceted filtering" seems most appropriate here?

dannon:
@jxtx Yep, exactly what we'd talked about in IRC a bit, +1

selten:
And possibly a search function as well to search by the tour name?

selten:
@dannon can this issue now be closed?

dannon:
Yep!

PR #7478, opened 2019-03-07
Update of tours overview according to #5860

selten:
Default view:
afbeelding

Filtered view:
afbeelding


dannon:
@selten Thanks for taking a poke at this, it's definitely going to be an improvement.  The logic for toggling tags looks like it's not applied quite right, though.  Given the default tours here:
image

If you do something like uncheck "UI", then uncheck "History", then recheck "UI" both related tours come back into the list.  I'd probably have the displayed list logic be such that tours are displayed if they match *any* of the top tags.

selten:
Altered behaviour, now all tags that belong to a tour need to be set by the buttons.

afbeelding

afbeelding

dannon:
@galaxybot test this

dannon:
@galaxybot test this

dannon:
@selten Works well for me.  I re-triggered the tests and will merge when they pass.

Issue #3947, opened 2017-04-20
Changing datatype in workflow not working for collections


mvdbeek:
It looks like changing the datatype in a workflow doesn’t work for collections. I have a collection of tabular files, I run the the `regex_replace` tool over that collection (which produces txt files), and I change the datatype to tabular in the workflow editor. Unfortunately tools that require tabular input will not accept the workflow output, even though the individual elements are tabular now.

[This](https://gist.github.com/mvdbeek/d228a65e4966f0f54d856e3bf15ea7f1) is a simply workflow that demonstrates the problem.

Not sure if this is a bug or if that has never been implemented.

jmchilton:
> Not sure if this is a bug or if that has never been implemented.

I think it is both - I've never implemented a test case for this anyway so I'm not surprised it doesn't work.

mvdbeek:
Done with https://github.com/galaxyproject/galaxy/pull/7480

PR #7480, opened 2019-03-07
Fix post job actions for Database Operation tools, implement change_datatype for collection output

### LINKED ISSUE #7473: Workflow collection operation "Build list": no hide; no delete intermediate data
Tool: Build List from one or more datasets (Galaxy Version 1.0.0)

At least two problems, might be related:

1. The tool job itself ends up in the history output -- not following the hidden/show toggle or the post job action "delete intermediate datasets".
1. The datasets the tool uses as inputs are copied back into the same history and are not hidden either. 

History: https://usegalaxy.org:/u/jen/h/test-buildlist-wf-dup-data

Workflow:  https://usegalaxy.org/u/jen/w/testbuildlist

ping @dannon @jmchilton

mvdbeek:
The rename action works, the hide action fails.
Building this to fix https://github.com/galaxyproject/galaxy/issues/7473.

### DIFF
--- a/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
+++ b/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
@@ -56,7 +56,7 @@ functools32==3.2.3.post2 ; python_version < '3.2'
 future==0.17.1
 futures==3.2.0 ; python_version == '2.6' or python_version == '2.7'
 galaxy-sequence-utils==1.1.3
-gxformat2==0.8.0
+gxformat2==0.8.1
 h5py==2.9.0
 idna==2.8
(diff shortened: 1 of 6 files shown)


dannon:
@mvdbeek Thanks for this -- I should have added tests for these years ago, would have saved some pain!

mvdbeek:
Looks like it's working, but I think I'll want to test all of the possible PJAs.

mvdbeek:
This will also fix https://github.com/galaxyproject/galaxy/issues/3947.

jmchilton:
I assume we should backport this to at least 19.01?

mvdbeek:
Yep, I wouldn't go further back than 19.01, it is more of a new feature than a bug I'd say.

Issue #7162, opened 2019-01-02
Issues with prefix path (/galaxy)


ashvark:
I am using galaxy 18.09 version with old galaxy.ini format. 

I am getting javascript error in the multiple history view. When I debug the issue i realised that galaxy request url does not contains the prefix (/galaxy) . for example http://localhost:8082/api/histories/0a248a1f62a0cc04/contents/datasets/e90facaf332c3038 . This issue happens only in the multiple history view.

![image](https://user-images.githubusercontent.com/7712157/50595796-36e55200-0ea2-11e9-8270-8dc32a1019cb.png)

Also, several links in the top menu seems to be broken. In this case the prefix path is duplicated
http://localhost:8082/galaxy/galaxy/workflows/list?__identifer=h0cm1bzxuc
http://localhost:8082/galaxy/galaxy/visualizations?__identifer=lt0el2s7baq
http://localhost:8082/galaxy/galaxy/user?__identifer=39ze7z0ldbc
http://localhost:8082/galaxy/galaxy/custom_builds?__identifer=juay9tburl8

if would be great if someone can provide a solution to fix this issue.

ashvark:
All the links (Workflow, Create Visualize, Shared Histories ) are working fine when accessing from the shared data libraries page whereas the same links are broken ("/galaxy/galaxy/*****) when accessing from the home page.

ashvark:
Also the issue with multi history panel seems to work with the latest dev version, I think  #7005 should have some fixes to this issue.

ashvark:
I made some dirty fixes in my galaxy application to overcome the issue with broken top menu links. I made following changes in the file ` client/galaxy/scripts/layout/menu.js`
```javascript
                    if (options.target == "__use_router__" && typeof Galaxy.page != "undefined") {
                        //Galaxy.page.router.push(options.url);
                        Galaxy.page.router.push(options.url.replace("/galaxy/","/"));
                    } else {
                        options.url=String(options.url).startsWith("/galaxy")? options.url : "/galaxy" + options.url;
                        Galaxy.frame.add(options);
                    }
```

```javascript
                    if (model.attributes.target == "__use_router__" && typeof Galaxy.page != "undefined") {
                        Galaxy.page.router.push(model.attributes.url.replace("/galaxy/","/"));
                        //Galaxy.page.router.push(model.attributes.url);
                    } else {
[...]

Issue #7474, opened 2019-03-06
Serving Galaxy with a url prefix broken in 19.01


abretaud:
Hello!
As I was testing 19.01 version of the docker image I find out that running Galaxy with an url prefix seems to be broken.

Here's a minimal test setup, using the PROXY_PREFIX option of the docker image (which worked fine with 18.09) : https://github.com/abretaud/galaxy_proxy

`docker-compose up -d` then go to http://localhost:8700/foo/bar/: 
-  some assets get a 404 error, e.g. it tries to download http://localhost:8700/static/scripts/bundled/assets/af7ae505a9eed503f8b8e6982036873e.woff2 instead of http://localhost:8700/foo/bar/static/scripts/bundled/assets/af7ae505a9eed503f8b8e6982036873e.woff2
- some pages don't work, e.g. go to data libraries -> it makes an xhr request to http://localhost:8700/api/libraries?deleted=false instead of http://localhost:8700/foo/bar/api/libraries?deleted=false

I suspect the first thing is related to this line https://github.com/galaxyproject/galaxy/blob/release_19.01/client/webpack.config.js#L100 but I'm not a webpack expert
Any help?

mvdbeek:
Can you check whether this is synced up with https://docs.galaxyproject.org/en/master/admin/nginx.html?highlight=proxy%20prefix#serving-galaxy-at-a-url-prefix ?
Especially
```
    mount: /galaxy=galaxy.webapps.galaxy.buildapp:uwsgi_app()
    manage-script-name: true
```

abretaud:
Yep it is:
```
uwsgi:
  socket: unix:///srv/galaxy/var/uwsgi.sock
  mount: /foo/bar/=galaxy.webapps.galaxy.buildapp:uwsgi_app()
  manage-script-name: true
```
And here's the nginx conf in the container
```
$ cat /etc/nginx/conf.d/uwsgi.conf 
# pass to uWSGI by default
location / {
    uwsgi_pass 127.0.0.1:4001;
    include uwsgi_params;
    uwsgi_param SCRIPT_NAME /foo/bar/;
    # forward the remote_user header in case it is set by a previous proxy
    uwsgi_param HTTP_REMOTE_USER $remote_user if_not_empty;
}
```

mvdbeek:
I'm not sure you'll want to have the trailing slash in there, but I don't think that's the issue.
[...]

PR #7481, opened 2019-03-07
[19.01] Use a relative output.publicPath to avoid 404 errors when serving with a url prefix

abretaud:
This should fix the first problem from #7474
The changelog is smaller than I thought, but it seems to be enough :)
ping @dannon

### DIFF
--- a/client/webpack.config.js
+++ b/client/webpack.config.js
@@ -97,7 +97,7 @@ let buildconfig = {
                     loader: "file-loader",
                     options: {
                         outputPath: "assets",
-                        publicPath: "/static/scripts/bundled/assets/"
+                        publicPath: "../scripts/bundled/assets/"
                     }
                 }
             },
--- a/templates/webapps/galaxy/galaxy.panels.mako
+++ b/templates/webapps/galaxy/galaxy.panels.mako
@@ -140,6 +140,8 @@
             | ${self.galaxy_config['title']}
             %endif
         </title>
+        ## relative href for site root
+        <link rel="index" href="${ h.url_for( '/' ) }"/>
 
         ${self.stylesheets()}
         ${self.javascripts()}


client/webpack.config.js
```diff
@@ -97,7 +97,7 @@ let buildconfig = {
                     loader: "file-loader",
                     options: {
                         outputPath: "assets",
-                        publicPath: "/static/scripts/bundled/assets/"
+                        publicPath: "../scripts/bundled/assets/"
```
nsoranzo, line 100:
```suggestion
                        publicPath: path.join(__dirname, "../scripts/bundled/assets/")
```

dannon:
@abretaud Nice!  Yeah, I was trying to set it on the fly in the client based on the configuration Galaxy shipped with the whole `__webpack_public_path__` var, but it didn't want to cooperate.  I didn't even think of just setting it to a relative path at build time -- I'll check and make sure this doesn't have any unintended side effects, but this looks good at first glance!

client/webpack.config.js
dannon: Unless I misread the suggestion, this will not work.  This defeats the purpose of it being relative and will generate a full path like `/Volumes/work_cs/galaxy/static/style`

nsoranzo: I saw this syntax used in the rest of the file, but feel free to disregard my suggestion.

dannon: Right, so for actually resolving paths for consumption when building we do want to use the concrete paths.  The publicPath is what actually gets injected into the resultant bundle for resolving resources as deployed, so we won't want the full thing.

abretaud:
Yeah I tried __webpack_public_path__ too until I had this stupid idea hehe

dannon:
@abretaud It's not stupid if it works!  I'm going to test it out in a sec, but I'm hopeful.

abretaud:
I've just pushed a fix for the second problem of #7474 : it seems to work on my test setup, but I haven't checked everything (I'm curious to know if it would fix #7162 too)
I'm leaving now (big week-end!), so feel free to edit the PR as you will
Cheers

dannon:
Doesn't look like the second fix is addressing the link/routing issues in the client, to me.  The first one, though, I think works fine and I haven't found anywhere that it breaks.

abretaud:
The last commit is because this link element is used by some Galaxy js code to determine the approot, without this the libraries page tries to make xhr calls without the URL prefix and gets 404

abretaud:
Ok, back to business! Anything I can help on this?

dannon:
@abretaud Hey, welcome back!  The relative path changes work well in my testing.  I'm hoping to push another fix to this branch this morning to address more of the second problem here, and then we can get it in and backport as appropriate.

dannon:
@abretaud The other prefix issue I was wrangling is different enough that I think I'm going to separate it out and just merge this now.  Thank you!

jmchilton (commented):
+1

abretaud:
Ok cool, thanks for working on this!

PR #7483, opened 2019-03-07
Log exception in set_metadata.py.

jmchilton:
We can finally do this as a result of #7095 I believe!


jmchilton:
xref https://trello.com/c/Nrwodu9d 😆

martenson:
> I would suggest writing the metadata exceptions to a file (rather than reusing the existing job streams). The framework could inject a logging stream into the datatypes to allow this.

? :)

mvdbeek:
Cool! Should we test it as well ?

martenson:
@mvdbeek As a separate PR maybe? We are testing sniffing already, not sure if another test here helps.
@jmchilton I'll leave you the honors to close the trello ticket.

PR #7484, opened 2019-03-07
Replace select2 instances with Tag component

Nerdinacan:
This is still a WIP as I wrap up the unit test coverage and make some minor style adjustments.

This PR replaces the old select2 instances with reactive Tagging components and unifies the design with the previous PR that replaced the mako-rendered Tagging elements.

**components/Tags**
Persists tag changes to the Vuex store, handles initialization and event handling. This is the tag you most likely want to use. 

**components/Tags/StatelessTags**
A stateless UI-only component. Changes to the styling of the tags themselves (to make it look more like the existing bootstrap elements or whatever is desired) should occur here.

**components/Tags/mounts**
2 mount functions for instantiating the components in non-Vue environments like the python rendered pages, or from within existing Backbone views. 

**components/Tags/tagService**
The previous implementations saved tags to a variety of confusing API endpoints locations. In some cases, an entire Dataset or History object would be saved, in others it would hit a specific endpoint which only saved the tags themselves. I've structured the Tags component so that it accepts a data service as a property so we can easily swap this behavior if necessary.

**store/tagStore**
Central Vuex store holding the current tags for a particular object. The main advantage we're getting with this is we're starting to see reactivity across different regions of the application. If you change the history tags in one place, it will update everywhere else those tags are used. (Without a page reload.)


Nerdinacan:
This history is getting pretty ugly. I've made a new clean commit for this and am opening a separate PR.

PR #7485, opened 2019-03-07
Git providers error reporting plugins

selten:
The GitLab error reporter looks up the Git repository by using the ToolShed (if applicable, it otherwise defaults to another project) and uses this information to submit an issue to the selected project (default or tool specific). It's based off of the GitHubPlugin.

### DIFF
--- a/config/error_report.yml.sample
+++ b/config/error_report.yml.sample
@@ -59,6 +59,8 @@
 # version and include all of the information the normal emailed bug reports
 # include. If you use a private Github Enterprise deployment, you can set
 # github_base_url='https://...' and github_api_url='https://api.....' as shown below.
+# The 'github_default_repo_only' flag restores the previous behaviour. When this is set
+# to true it will automatically only submit to the default git repository.
 # - type: github
 #   verbose: false
 #   user_submission: true
@@ -67,6 +69,7 @@
 #   github_api_url: https://api.github.com
 #   github_default_repo_owner: galaxyproject
 #   github_default_repo_name: galaxy
+#   github_default_repo_only: true
 
 # GitLab error reporting backend. You will need to `pip install python-gitlab`
 # in the galaxy virtualenv. This will create a new issue if none exists, and
@@ -76,7 +79,8 @@
 # gitlab_base_url='https://...'. It supports creating an issue on the git
 # repository of the tool by querying the ToolShed where the tool comes from
 # (if applicable). Set verbose to true if you want the message to be displayed
-# to the user.
+# to the user. The 'gitlab_default_repo_only' flags ensures all errors are
+# submitted to the default repository only.
 # - type: gitlab
 #   verbose: false
 #   user_submission: true
@@ -84,3 +88,4 @@
 #   gitlab_private_token: 00000000000
 #   gitlab_default_repo_owner: galaxyproject
 #   gitlab_default_repo_name: galaxy
+#   gitlab_default_repo_only: true
--- a/config/error_report.yml.sample
+++ b/config/error_report.yml.sample
@@ -64,3 +64,18 @@
 #   github_oauth_token: 00000000000
 #   github_repo_owner: galaxyproject
 #   github_repo_name: galaxy
+
+# GitLab error reporting backend. You will need to `pip install python-gitlab`
+# in the galaxy virtualenv. This will create a new issue if none exists, and
+# comment on existing, open issues. The issues are labelled based on tool ID /
+# version and include all of the information the normal emailed bug reports
+# include. If you use a private GitLab deployment, you can set
+# gitlab_base_url='https://...'. This is based off of the Github error
+# reporting backend. It supports creating an issue on the git repository of
+# the tool by querying the ToolShed where the tool comes from (if applicable).
+# - type: gitlab
+#   user_submission: true
+#   gitlab_base_url: https://gitlab.com
+#   gitlab_private_token: 00000000000
+#   gitlab_default_repo_owner: galaxyproject
+#   gitlab_default_repo_name: galaxy
--- a/config/error_report.yml.sample
+++ b/config/error_report.yml.sample
@@ -73,7 +73,9 @@
 # gitlab_base_url='https://...'. This is based off of the Github error
 # reporting backend. It supports creating an issue on the git repository of
 # the tool by querying the ToolShed where the tool comes from (if applicable).
+# Set verbose to true if you want the message to be displayed to the user.
 # - type: gitlab
+#   verbose: false
 #   user_submission: true
 #   gitlab_base_url: https://gitlab.com
 #   gitlab_private_token: 00000000000
--- a/config/error_report.yml.sample
+++ b/config/error_report.yml.sample
@@ -58,12 +58,15 @@
 # comment on existing, open issues. The issues are labelled based on tool ID /
 # version and include all of the information the normal emailed bug reports
 # include. If you use a private Github Enterprise deployment, you can set
-# github_base_url='https://...'
+# github_base_url='https://...' and github_api_url='https://api.....' as shown below.
 # - type: github
+#   verbose: false
 #   user_submission: true
 #   github_oauth_token: 00000000000
-#   github_repo_owner: galaxyproject
-#   github_repo_name: galaxy
+#   github_base_url: https://github.com
+#   github_api_url: https://api.github.com
+#   github_default_repo_owner: galaxyproject
+#   github_default_repo_name: galaxy
 
 # GitLab error reporting backend. You will need to `pip install python-gitlab`
 # in the galaxy virtualenv. This will create a new issue if none exists, and
--- a/config/error_report.yml.sample
+++ b/config/error_report.yml.sample
@@ -73,10 +73,10 @@
 # comment on existing, open issues. The issues are labelled based on tool ID /
 # version and include all of the information the normal emailed bug reports
 # include. If you use a private GitLab deployment, you can set
-# gitlab_base_url='https://...'. This is based off of the Github error
-# reporting backend. It supports creating an issue on the git repository of
-# the tool by querying the ToolShed where the tool comes from (if applicable).
-# Set verbose to true if you want the message to be displayed to the user.
+# gitlab_base_url='https://...'. It supports creating an issue on the git
+# repository of the tool by querying the ToolShed where the tool comes from
+# (if applicable). Set verbose to true if you want the message to be displayed
+# to the user.
 # - type: gitlab
 #   verbose: false
 #   user_submission: true
--- a/lib/galaxy/tools/error_reports/plugins/github.py
+++ b/lib/galaxy/tools/error_reports/plugins/github.py
@@ -32,6 +32,7 @@ class GithubPlugin(BaseGitPlugin):
         self.github_api_url = kwargs.get('github_api_url', 'https://api.github.com')
         self.git_default_repo_owner = kwargs.get('github_default_repo_owner', False)
         self.git_default_repo_name = kwargs.get('github_default_repo_name', False)
+        self.git_default_repo_only = string_as_bool(kwargs.get('github_default_repo_only', True))
 
         try:
             import github
@@ -70,8 +71,8 @@ class GithubPlugin(BaseGitPlugin):
             ts_repourl = self._get_gitrepo_from_ts(job, ts_url)
 
             # Determine the GitLab project URL and the issue cache key
-            github_projecturl = urlparse.urlparse(ts_repourl).path[1:] if ts_repourl else \
-                "/".join([self.git_default_repo_owner, self.git_default_repo_name])
+            github_projecturl = urlparse.urlparse(ts_repourl).path[1:] if (ts_repourl and not self.git_default_repo_only) \
+                else "/".join([self.git_default_repo_owner, self.git_default_repo_name])
             issue_cache_key = self._get_issue_cache_key(job, ts_repourl)
 
             # Connect to the repo
--- a/lib/galaxy/tools/error_reports/plugins/github.py
+++ b/lib/galaxy/tools/error_reports/plugins/github.py
@@ -96,10 +96,9 @@ class GithubPlugin(BaseGitPlugin):
             log.info(error_title in self.issue_cache[issue_cache_key])
             if error_title not in self.issue_cache[issue_cache_key]:
                 # Create a new issue.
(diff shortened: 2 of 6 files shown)


lib/galaxy/tools/error_reports/plugins/gitlab.py
```diff
@@ -0,0 +1,144 @@
...
+            else:
+                import urllib.parse as urllib
+                urlparse = urllib
+
+            log.info(job.tool_id)
+            log.info(job.tool_version)
+            log.info(tool.tool_shed)
+
+            # Determine the ToolShed
+            ts_url_request = requests.get('http://' + str(tool.tool_shed))
```
hexylena:
```suggestion
            ts_url_request = requests.get('https://' + str(tool.tool_shed))
```

lib/galaxy/tools/error_reports/plugins/gitlab.py
```diff
@@ -0,0 +1,144 @@
...
+            gitlab_urlencodedpath = urllib.quote_plus(gitlab_projecturl)
+
+            # Make sure we are always logged in, then retrieve the GitLab project
+            self.gitlab.auth()
+            gl_project = self.gitlab.projects.get(gitlab_urlencodedpath)
+
+            # Get a list of all the issues, parse them for de-duplication later
+            log.info(gl_project.issues.list())
+
+            issue_cache = {}
```
hexylena: Have you considered modifying the issue cache to be as part of `__init__`? Then you could have sub-caches per `gitlab_projecturl`, which might speed up error reporting for tools with large bug lists (and help decrease the chance that you hit API ratelimits).

lib/galaxy/tools/error_reports/plugins/gitlab.py
```diff
@@ -0,0 +1,144 @@
...
+                    "api",
+                    "v4",
+                    "projects",
+                    gitlab_urlencodedpath,
+                    "issues",
+                    str(issue_cache[error_title]),
+                    "notes"
+                ])
+                self.gitlab.http_post(gl_url, post_data={'body': error_message})
+            return ('Submitted error report to GitLab. Your issue number is %s. You can view the issue on <a href="%s/%s/issues/%s" target="_blank">GitLab</a>.' % (issue_cache[error_title], self.gitlab_base_url, gitlab_projecturl, issue_cache[error_title]), 'success')
```
hexylena:
```suggestion
            return ('Submitted error report to GitLab. Your issue number is <a href="%s/%s/issues/%s" target="_blank">#%s</a>.' % (self.gitlab_base_url, gitlab_projecturl, issue_cache[error_title], issue_cache[error_title]), 'success')
```

lib/galaxy/tools/error_reports/plugins/gitlab.py
```diff
@@ -0,0 +1,144 @@
...
+                gitlab_projecturl = urlparse.urlparse(ts_repourl).path[1:]
+            else:
+                gitlab_projecturl = "/".join([self.gitlab_default_repo_owner,
+                                              self.gitlab_default_repo_name])
+
+            gitlab_urlencodedpath = urllib.quote_plus(gitlab_projecturl)
+
+            # Make sure we are always logged in, then retrieve the GitLab project
+            self.gitlab.auth()
+            gl_project = self.gitlab.projects.get(gitlab_urlencodedpath)
```
hexylena: If the repository URL is not valid (i.e. same base as `gitlab_base_url`), will this step fail causing this bug reporter to fail (gracefully)?

hexylena (commented):
Hi @selten 

I like the idea of submitting bug reports to the relevant repositories! That's an interesting idea, I'll have to consider adding that to the github vesion as well. Looks good overall

lib/galaxy/tools/error_reports/plugins/gitlab.py
selten: @erasche the idea behind this piece of code is to get Python to determine whether the website is HTTP or HTTPS. As not all ToolSheds will have HTTPS, this piece of code actually detects the HTTP redirect. So for instance for the testtoolshed, it goes to HTTPS. It just removes the requirement for needing a SSL certificate on the ToolShed for people their ToolShed deployments.

lib/galaxy/tools/error_reports/plugins/gitlab.py
selten: Possibly a clean issue cache on `__init__` which gets filled on the first bug report for a specific tool would make most sense in this case. I'll add this.

lib/galaxy/tools/error_reports/plugins/gitlab.py
selten: It should throw an error then. It won't fail gracefully. That's still to be done.

lib/galaxy/tools/error_reports/plugins/gitlab.py
```diff
@@ -47,98 +47,124 @@ def __init__(self, **kwargs):
...
+                        "v4",
+                        "projects",
+                        gitlab_urlencodedpath,
+                        "issues",
+                        str(issue_cache[error_title]),
+                        "notes"
+                    ])
+                    self.gitlab.http_post(gl_url, post_data={'body': error_message})
+                return ('Submitted error report to GitLab. Your issue number is <a href="%s/%s/issues/%s" target="_blank">#%s</a>.' % (self.gitlab_base_url, gitlab_projecturl, issue_cache[error_title], issue_cache[error_title]), 'success')
+            except ImportError:
```
hexylena: these should all be available. Don't need to be part of try/except.

lib/galaxy/tools/error_reports/plugins/gitlab.py
```diff
@@ -0,0 +1,230 @@
...
+            self.gitlab = None
+
+    def submit_report(self, dataset, job, tool, **kwargs):
+        """Submit the error report to GitLab
+        """
+        log.info(self.gitlab)
+
+        if self.gitlab:
+            # Import GitLab here for the error handling
+            import gitlab
```
hexylena, line 78: Does this need to be done again when it was done in import? I would have thought it stayed loaded.

lib/galaxy/tools/error_reports/plugins/gitlab.py
```diff
@@ -0,0 +1,230 @@
...
+            try:
+                # Import the necessary libraries
+                import sys
```
hexylena: Move all of this to the top, all of these will be available

lib/galaxy/tools/error_reports/plugins/gitlab.py
```diff
@@ -0,0 +1,230 @@
...
+                import requests
+                if sys.version_info[0] < 3:
+                    import urllib as urllib
+                    import urlparse as urlparse
+                else:
+                    import urllib.parse as urllib
+                    urlparse = urllib
+
+                log.info("GitLab error reporting - submit report - job tool id: " + job.tool_id)
```
hexylena: Could you collapse into a single logging message? Also if you use the syntax

```
log.info("some %s details %s", a, b)
```

Then logging library doesn't have to construct the string unless it will be printed (saving a tiiiiny amount of time if log level is above the level of the message.)

lib/galaxy/tools/error_reports/plugins/gitlab.py
```diff
@@ -0,0 +1,230 @@
...
+                log.info("GitLab error reporting - submit report - job tool version: " + job.tool_version)
+                log.info("GitLab error reporting - submit report - tool tool_shed: " + tool.tool_shed)
+
+                # Determine the ToolShed url, initially we connect with HTTP and if redirect to HTTPS is set up,
+                # this will be detected by requests and used further down the line. Also cache this so everything is
+                # as fast as possible
+                if tool.tool_shed in self.ts_urls:
```
hexylena: I (personally, maybe irrationally) dislike deep indentation and might have written as:

```
if tool.tool_shed not in self.ts_urls:
  ...
  self.ts_urls[tool.tool_shed] = ts_url

ts_url = self.ts_urls[tool.tool_shed]
```

but personal preference, feel free to ignore this suggestion.

lib/galaxy/tools/error_reports/plugins/gitlab.py
```diff
@@ -0,0 +1,230 @@
...
+                    ts_url = self.ts_urls[tool.tool_shed]
+                else:
+                    ts_url_request = requests.get('http://' + str(tool.tool_shed))
+                    ts_url = ts_url_request.url
+                    self.ts_urls[tool.tool_shed] = ts_url
+                log.info("GitLab error reporting - Determined ToolShed is " + ts_url)
```
hexylena:
```suggestion
                log.info("GitLab error reporting - Determined ToolShed is %s", ts_url)
```

lib/galaxy/tools/error_reports/plugins/gitlab.py
```diff
@@ -0,0 +1,230 @@
...
+
+                    for changeset, repoinfo in ts_repo_request_data.items():
+                        if isinstance(repoinfo, dict):
+                            if 'repository' in repoinfo.keys():
+                                if 'remote_repository_url' in repoinfo['repository'].keys():
+                                    ts_repourl = repoinfo['repository']['remote_repository_url']
+
+                    self.ts_repo_cache[job.tool_id] = ts_repourl
+
+                log.info("GitLab error reporting - Determine ToolShed Repository URL: " + ts_repourl)
```
hexylena:
```suggestion
                log.info("GitLab error reporting - Determine ToolShed Repository URL: %s", ts_repourl)
```

lib/galaxy/tools/error_reports/plugins/gitlab.py
```diff
@@ -0,0 +1,230 @@
...
+                    ])
+                    self.gitlab.http_post(gl_url, post_data={'body': error_message})
+
+                return ('Submitted error report to GitLab. Your issue number is <a href="%s/%s/issues/%s" '
+                        'target="_blank">#%s</a>.' % (self.gitlab_base_url, gitlab_projecturl,
+                                                      self.issue_cache[issue_cache_key][error_title],
+                                                      self.issue_cache[issue_cache_key][error_title]), 'success')
+
+            except ImportError as e:
+                log.error("GitLab error reporting - Please ensure that requests, urllib and urlparse are available. "
```
hexylena: I believe you can replace this with:

```
except ImportError:
  log.exception("Your message")
```

lib/galaxy/tools/error_reports/plugins/gitlab.py
```diff
@@ -0,0 +1,230 @@
...
+                          "Exception information: " + e.message)
+                return ('Error occured while submitting error report to GitLab. Report could not be submitted due to '
```
hexylena: I'd say that since these are the messages that bubble up to the user, you don't need to (and maybe shouldn't) expose that much debugging information. I'd say that you just write all of the exceptions and then at the end write `return ('Internal Error', 'danger')` and be done.

hexylena (commented):
Some more comments :)

lib/galaxy/tools/error_reports/plugins/gitlab.py
hexylena: I would expect the `tool.tool_shed` to be correct, but I can recognise that if it is written `http` in the tool, while the server actually responds on HTTPS, you don't want to have to update your tool for the bug reporter to work correctly.

OK, sounds fine. (But people should really write the correct tool shed URL in their tool shed repo metadata.)

lib/galaxy/tools/error_reports/plugins/gitlab.py
selten: Seemingly doesn't stay loaded (or at least not available for submit_report)

lib/galaxy/tools/error_reports/plugins/github.py
```diff
@@ -20,73 +26,126 @@ def __init__(self, **kwargs):
...
-            return ('Submitted bug report to Github. Your issue number is %s' % self.issue_cache[error_title].number, 'success')
+                self._append_issue(issue_cache_key=issue_cache_key, error_title=error_title, error_message=error_message)
+            return ('Submitted error report to Github. Your issue number is <a href="%s/%s/issues/%s" '
+                    'target="_blank">#%s</a>.' % (self.github_base_url, github_projecturl,
+                                                  self.issue_cache[issue_cache_key][error_title].number,
+                                                  self.issue_cache[issue_cache_key][error_title].number), 'success')
+
+    def _create_issue(self, **kwargs):
+        # Create a new issue.
+        self.issue_cache[kwargs.get('issue_cache_key')][kwargs.get('error_title')] = kwargs.get('gh_project').create_issue(
```
hexylena: Would you mind just make these explicit function parameters rather than pulling from kwargs, since all of the arguments are required? Same for `_append_issue`

config/error_report.yml.sample
```diff
@@ -58,9 +58,29 @@
...
+#   github_api_url: https://api.github.com
+#   github_default_repo_owner: galaxyproject
+#   github_default_repo_name: galaxy
+
+# GitLab error reporting backend. You will need to `pip install python-gitlab`
+# in the galaxy virtualenv. This will create a new issue if none exists, and
+# comment on existing, open issues. The issues are labelled based on tool ID /
+# version and include all of the information the normal emailed bug reports
+# include. If you use a private GitLab deployment, you can set
+# gitlab_base_url='https://...'. This is based off of the Github error
```
hexylena: No need to say based off of GH backend

hexylena (changes requested):
Great changes @selten, the refactoring looks quite nice.

I'm sorry to ask for another feature so late in the review, but would you mind adding a flag, maybe `git{hub,lab}_submit_only_to_default: true`? (Open to better naming) Now that you've refactored to include github, this will change the behaviour of the github error reporter, potentially unexpectedly.

I worry that someone would configure this plugin, and have it:

1. start spamming other projects with bug reports (especially if `user_submission: false`)
2. release internal details about the site's configuration that the admin did not necessarily intend to release.

Specifically if the plugin allows authenticating to GitHub, to submit bugs to that sites's private issue list, it would be permitted to post bugs to tools-iuc and many other repositories.

hexylena (approved):
Looks great! Lots of good improvements and cleanups to the github plugin and the gitlab plugin can be really useful for some of us.

dannon:
@galaxybot test this

dannon:
Nice, thanks @selten!

PR #4659, opened 2017-09-20
Dataset Checksum

VJalili:
This PR calculates an MD5-based checksum for every uploaded dataset, and stores it as an attribute of a dataset in `dataset` table. 

### Why are we doing this? 
We are planning to enable data mounting, i.e., plug-and-play any storage/folder such as Amazon S3 bucket and Microsoft Azure blob. One of the basic requirements of this goal is to check and maintain data integrity. In other words, make sure a previously called `/a/path/someFile.bed` is the very same file the next time you access it (and probably, allow user to relocated a renamed and/or moved file). A common approach for data integrity assessment is using __checksum__--a (cryptographic) hash key generated based on a given file content. 

### What's next? 
Calculating checksum could be an expensive process as the entire file should be streamed in memory, which is a bottleneck in particular with large files. To avoid such challenges, in this PR we are streaming a file in memory in smaller chunks. However, a reasonable optimization would be parallelizing the upload and checksum calculation. 

Additionally, we also need to remotely compute checksum for remote files (e.g., hosted on S3). You may track a discussion on this topic [here](https://github.com/gvlproject/cloudbridge/issues/76) .

### Which datasets will be "assigned" a checksum ?
- Automatic: 
-- all datasets uploaded to or downloaded from a cloud-based storage (e.g., Amazon S3).
- Optional (user preference): 
-- all the dataset persisted on a local storage. 
- Excluded: 
-- all temporary datasets (e.g., workflow/tool execution related tmp files).

### When a checksum is computed by Galaxy ? 
- For data hosted on cloud: we first try to fetch a hash computed by the provider (e.g., boto ETag), if available we use it (most probably it is always available), if not Galaxy computes the checksum. 
- For data hosted on local storage: Galaxy computes the checksum, if set to (see above).

### If a checksum should be computed, which hash function is used ? 
- For data hosted on cloud: the hash function which is supported/required by the provider.
- For data hosted on a local storage: user-preference.

PR #7487, opened 2019-03-07
Implement dataset source and hash tracking.

jmchilton:
Generalization of #4659, dataset checksum should also indicate if they are the checksum for the top-level data or data within extra files path. These models get closer to accounting for that. Extend this work to also allow tracking the source a dataset (or part of a dataset) was generated from as well as the hash of those objects. Includes initial support in the Upload 2.0 API to track the source of certain URLs coming in, the hash of certain things coming in, and an option to validate against those.

This can be the basis of doing a lot of cool stuff in the rule based uploader, dealing with remote data (dev/users customizing these things as part of model store imports - see #7367 for instance), cloud stuff, etc.. not sure what a minimum viable product people would want to see is.


lib/galaxy/util/hash_util.py
```diff
@@ -17,11 +17,25 @@
...
+    "SHA-256": sha256,
+    "SHA-512": sha512,
+}
+HASH_NAMES = list(HASH_NAME_MAP.keys())
+
+
+def memory_bound_hexdigest(hash_func=None, hash_func_name=None, path=None, file=None):
+    if hash_func is None:
+        assert hash_func_name is not None
+        assert hash_func_name in HASH_NAME_MAP
```
nsoranzo: This assert is not needed.

test/api/test_libraries.py
```diff
@@ -135,19 +135,42 @@ def test_fetch_zip_to_folder(self):
 
     def test_fetch_single_url_to_folder(self):
         history_id, library, destination = self._setup_fetch_to_folder("single_url")
-        items = [{"src": "url", "url": "https://raw.githubusercontent.com/galaxyproject/galaxy/dev/test-data/4.bed"}]
+        items = [{"src": "url", "url": "https://raw.githubusercontent.com/galaxyproject/galaxy/dev/test-data/4.bed", "MD5": "37b59762b59fff860460522d271bc111"}]
```
nsoranzo: Split in multiple lines?

test/api/test_libraries.py
```diff
@@ -135,19 +135,42 @@ def test_fetch_zip_to_folder(self):
...
         }
         self.dataset_populator.fetch(payload)
         dataset = self.library_populator.get_library_contents_with_path(library["id"], "/4.bed")
         assert dataset["file_size"] == 61, dataset
 
+    def test_fetch_failed_validation(self):
+        # Exception handling is really rough here - we should be creating a dataset in error instead
+        # of just failing the job like this.
+        history_id, library, destination = self._setup_fetch_to_folder("single_url")
+        items = [{"src": "url", "url": "https://raw.githubusercontent.com/galaxyproject/galaxy/dev/test-data/4.bed", "MD5": "37b59762b59fff860460522d271bc112"}]
```
nsoranzo: Split in multiple lines?

lib/galaxy/model/migrate/versions/0148_add_checksum_table.py
```diff
@@ -0,0 +1,64 @@
...
+        log.exception("Adding dataset source and hash tables failed.")
+
+
+def downgrade(migrate_engine):
+    metadata.bind = migrate_engine
+    metadata.reflect()
+    try:
+        dataset_source_table.create()
+        dataset_hash_table.create()
+        dataset_source_hash_table.create()
```
nsoranzo: Tables should be dropped here, not created.

lib/galaxy/model/mapping.py
```diff
@@ -223,6 +223,32 @@
...
+    Column("dataset_id", Integer, ForeignKey("dataset.id"), index=True),
+    Column("hash_function", TEXT),
+    Column("hash_value", TEXT),
+    Column("extra_files_path", TEXT),
+)
+
+model.DatasetSourceHash.table = Table(
+    "dataset_source_hash", metadata,
+    Column("id", Integer, primary_key=True),
+    Column("dataset_id", Integer, ForeignKey("dataset_source.id"), index=True),
```
nsoranzo: Should this column be called `dataset_source_id` to avoid confusion (also below)?

jmchilton: Definitely - that is a copy and paste problem - sorry about that.

martenson:
One of the cool things we could do is to use hashes to deduplicate our storage (i.e. store every dataset only once).

martenson:
xref https://github.com/galaxyproject/galaxy/issues/7520

PR #2385, opened 2016-05-19
[wip] Dynamic proxy using uWSGI. 

jxtx:
A dynamic proxy implementation using uWSGI. A RPC function written in python is used to map the session to the appropriate backend.

Why?
1. We use uwsgi for Galaxy main. With this configuration we can proxy directly to interactive environments, no double proxy needed. 
2. We've talked about making uwsgi the default webserver for all Galaxy's. If we do that we can proxy directly without needing additional processes or ports.
3. uwsgi is fast, flexible, and well tested.
4. I was having trouble getting the node based proxy running.

ping @jmchilton @natefoo @bgruening @erasche

Issue #4651, opened 2017-09-19
GIE Hardening


natefoo:
After a few attempts to teach workshops (mainly overseas) we've discovered some robustness problems with GIEs on usegalaxy.org. Here's a list of identified issues and potential resolutions, as well as planning for enhancements, in order of priority:

- [x] Re-fix autoloading launch dataset @natefoo (broken by #4458, fix in #4760)
- [x] Client alive timeouts don't work for slow connections: client side functions (the readiness test and keepalive) use a 500ms timeout which is easily hit. Also these functions can be updated to handle general failures and timeouts differently. The timeout count should be reset when the connection succeeds @natefoo #4680 
- [x] Increase the in-container timeout (this is ~~trivially easy~~not supported with the containers lib =/) @natefoo #4740
- [ ] Monitoring: ensure usegalaxy.org GIEs are working, probably w/ Jenkins/Selenium @jmchilton #4694 
- [ ] Reconnecting to an existing running container @martenson 
  - [ ] Reconnection UI
  - [ ] Indicator of whether you have a running container
- [x] More feedback on progress while waiting for launch, remove spinner and display permanent error message on failure @natefoo #4680
- [ ] Asynchronous dataset `get()` on launch
- [x] Missing green "save" button in 17.05.1 Jupyter image
  - [x] Fix and build new image @bgruening bgruening/docker-jupyter-notebook#23
  - [x] Deploy new image on Test/Main @natefoo 
- [ ] Prevent launch on configurably large datasets
- [ ] uWSGI proxy @natefoo 
- [ ] Keep terminated containers for a configurable period and allow them to be relaunched
  - [ ] Most recent container only
  - [ ] Older containers
- [ ] Multiple containers per user/session
- [x] swarmscale bug fixes @natefoo galaxyproject/usegalaxy-playbook@dd9ab853788639b2b2916a215a8d955258d55e98
- [x] Implement Docker API containers interface @natefoo #5861
- [ ] Improve the launch UI

natefoo:
ping @nekrut

nekrut:
@erasche @bgruening - we need to solve the mystery of the green button. @davebx volunteered to help so can you guide him?

hexylena:
We're happy to help, but this isn't a good week for us. Next week will be better for me at least, I think @bgruening will still be at conferences/meetings then though.

blankenberg:
Would be neat to enable checkpointing and restoring, including filesystem changes.

[...]

Issue #1923, opened 2016-03-15
IE Provevance


nekrut:
To truly take advantage of IEs such as Jupyter `get` and `put` functions need to be modified so that input and output datasets are trackable in history. This would also allow inclusion of IEs in workflows. An enormous advantage here is that you can quickly write, say, a statistical test function in Jupyter and then use it as a tool. (e.g., see [this](http://googleresearch.blogspot.com/2014/08/doing-data-science-with-colaboratory.html)).

hexylena:
Closing in favour of the bullet point in the existing GIE roadmap.

dannon:
@erasche The bullet on the IE roadmap is linked here.  Should this not remain open as a way to discuss and plan for this particular issue, similar to how other roadmap issues are linked?

hexylena:
@dannon that would possibly be reasonable. How much discussion is warranted on this feature?

There aren't many way to implement, we have evidence that people who are writing notebooks are capable of using the linked google research syntax, and we'll turn that into a tool somehow.

martenson:
ITs can be used in workflows now, I think this issues has been resolved.

PR #1688, opened 2016-02-05
Overhaul of Tools and Jobs for 16.04

jmchilton:
This is a fairly substantial cleanup of tool and job handling.
- Introduce `profile="16.04"` attribute tools, the resulting tools have all new defaults that should really simplify tool development. https://github.com/galaxyproject/galaxy/issues/1609
  - `#set -e` on tool executions.
  - Use exit code by default for error detection.
  - Eliminate some buggy features - `interpreter` attribute for instance.
- Make bash the default shell with 16.04 so conda will work out of the box.
- Ensure tool execution working directories are empty. https://github.com/galaxyproject/galaxy/issues/1575
- Enable per-destination configuration of many job properties (makes Galaxy much more usable for multi-cluster deployments).  https://github.com/galaxyproject/galaxy/pull/1573, https://trello.com/c/6w8bples
- Small cleanups of legacy features and duplicated job runner code.

PR #7494, opened 2019-03-08
Add InteractiveTools.

blankenberg:
Basic idea: launch a container-backed Galaxy Tool and get access to content inside in real time.

- Give user access via uwsgi (inspired by https://github.com/galaxyproject/galaxy/pull/2385), or another proxy.
- Access is based upon a key, key_type, token mapping to a host, port. You can have and access any number of RealTimeTools at a time. 
- Port and entry url-path snippet are specified in standard tool.xml files. 
- You can specify any number of ports, so a single RealTimeTool can give access to multiple running applications.
- RealTimeTools can be added to and installed from the ToolShed.
- Currently working for docker in local runner.


lib/galaxy/datatypes/realtime.py
```diff
@@ -0,0 +1,77 @@
+"""Binary classes"""
+from __future__ import print_function
+
+import logging
+import os
+import sys
+
+from text import Html
```
nsoranzo:
```suggestion
from .text import Html
```

jmchilton:
Committers meeting sort of got hung up on this, we need more context here I think @blankenberg. Can you provide an example in tree and outline the compelling use case driving this development?

jmchilton:
I'm going to push this to 19.09 because @jxtx marked himself as needing to review it, I'm likewise not ready to merge it. We have a week, it may still get merged.

jmchilton:
I'm 👍 on the idea but I'm not quite sold on the implementation yet, however I will be happy to do the leg work to address my own concerns. My intention is:

- Integrate this Pulsar - develop an internal API to publish hostname/port back to Galaxy.
- Implement a notebook tool.
- Add a Selenium test case.

This is to convince myself this generalizes beyond the local job runner, we can support it in an ongoing fashion, we have the right data model, and to see if I can avoid persisting those docker commands in the job state.

I'd also like to see the proxies merged, but that seems unlikely at first glance. I'd also like to see it hidden behind ``enable_beta_tool_formats`` for one release, but I won't insist on it.

If I have not made substantial progress on these goals in the next month (ping me May 16th), I will merge it as is barring comments from other reviewers.

natefoo:
I'd also like to see if we can get uWSGI to manipulate the paths of proxied cookies, this will help make them work out of the box without needing a DNS wildcard and a wildcard SSL cert.

lib/galaxy/jobs/runners/local.py
```diff
@@ -170,6 +181,7 @@ def stop_job(self, job_wrapper):
                 return
         else:
             log.warning("stop_job(): %s: PID %d refuses to die after signaling TERM/KILL" % (job.id, pid))
+        self.__kill_container(job_wrapper)
```
bgruening: `self._kill_container(job_wrapper)`

lib/galaxy/jobs/runners/local.py
```diff
@@ -213,12 +225,70 @@ def _check_pid(self, pid):
...
+            if cont:
+                if cont.container_type == 'docker':
+                    # Handle getting RealTimeTool ports, if applicable
+                    eps = job.realtimetool_entry_points
+                    if eps:
+                        max_command_attempts = 10 + 1
+                        ports_raw = None
+                        for i in range(1, max_command_attempts):
+                            with tempfile.TemporaryFile() as stdout_file:
+                                exit_code = subprocess.call(cont.container_info['commands']['port'],
```
bgruening: Why not `_run_command`?

bgruening (commented):
Hi Dan, I added a bunch of comments and

* documentation
* condor-runner integration, yeah it works! :)
* 2 IEs that demonstrate the usage, Jupyter and Single-Cell

https://github.com/blankenberg/galaxy/pull/4/files

Everything rough, sorry we do not have that much time at the moment :(

lib/galaxy/jobs/runners/local.py
blankenberg: Nice catch, thanks!

lib/galaxy/jobs/runners/local.py
blankenberg: We need to capture and parse `stdout` to get the port info here, whereas in the other cases, we just let the stream pass through. Don't want the overhead of making a temp file when we don't need it. Could make it flag-able in the method, but it might be messier due to the with.

blankenberg:
> Hi Dan, I added a bunch of comments and
> 
> * documentation
> * condor-runner integration, yeah it works! :)
> * 2 IEs that demonstrate the usage, Jupyter and Single-Cell
> 
> https://github.com/blankenberg/galaxy/pull/4/files
> 
> Everything rough, sorry we do not have that much time at the moment :(

Awesome news. Will take a look soon. Thanks!

hexylena:
Hi @blankenberg, you suggested we continue the discussion here

> > I think we should stick to the "Interactive" name and keep that branding in the community and migrate all internal IEs that are feasible to those tools, this could be a nice hackathon project 
>
> I disagree, these are Galaxy RealTimeTools. GIEs are something else, as described in 10.1371/journal.pcbi.1005425.

Likewise, I disagree. You implement the core ideas of GIEs and things that were on our roadmap, while positioning this as something novel. It's features we've discussed previously that we either wanted to implement or wanted to discuss with the devteam what the best implementation would be.

- [where it was raised that we should run these as tools](https://github.com/galaxyproject/galaxy/issues/1923) which we merged into our GIE roadmap but had temporarily removed from our high priority GIE fixes
- [managing multiple running containers](https://github.com/galaxyproject/galaxy/issues/4651) and long running containers
- [ability to shut down IEs when done](https://github.com/galaxyproject/galaxy/issues/1586)

In the most optimistic interpretation of your actions, how this was done was a bit tone deaf and without considerations for optics or community contributors.

I'm struggling not to have a more negative interpretation, perhaps you can clear this up for me?

nekrut:
It pains me greatly to see this level of deliberations related to naming. We can probably dig into the exact history from the start here using github, but I'm pretty sure the first GIEs were developed by Bjorn and Helena. They probably weren't the first one who though of doing interactive analyses in Galaxy, but they did implement this (I certainly though of creating infinite energy using cold fusion, but it is unlikely that my name will be associated with it, which is perfectly fine with me). This got all of us excited and this is how the GIE manuscript came about. The future of Galaxy is very much dependent on ability to seamlessly switch between Galaxy proper and *ad hoc* analysis environments such as Jupyter, RStudio, and anything else, as well as interactive visualizations. This was a great breakthrough, but as time went by we realized that just as piston engines were not good for supersonic flight the original GIEs could be improved on and better integrated with Galaxy. This is true for any first implementation. Dan rethought this and developed iteration 2 of this. It may be mechanistically different, but ultimately it achieves the same goal. So a good compromise will be to just call it something like "interactive tools" and get this settled. This does not in any possible way diminish one's ability to publish on that. In fact, the paper will be significantly better if it embraces all community members who care about GIEs.

bgruening:
Hi all,

I would like to keep this PR technical, lets move the historical and political aspect to a chat with all interesting parties and figure out what is our next plans.
I think there is and was a huge amount of miscommunication involved here and I would like to propose an IE/RRT meeting. Would you all have time on Tuesday at 16.00 CEST? We can use gitter as some indicated that text would be better than a video call.

Would be great if you all can make it.

jmchilton:
The two things that I grate on most here when digging through the code are writing command-lines to the database and the nested loops and conditionals in the runner code that are tied to particular container types and these commands to kill the container.

I had a thought about a different, potentially more general approach. I feel like the command builder is already dealing with container details, what if we wrote the stop/kill commands into the job script with a trap https://www.shellscript.sh/trap.html function. Would that solve the problem with the stop/kill command for all the runners and for all container execution? Does anyone know off hand if that is going to work (or not) before I try it out?

I'm not suggesting that this should necessarily block the merge of this - because I'm willing to accept containerized tools are things I committed and were more or less broken by design, but if we're going to fix it I wonder if we should fix it more generally and fixing it would make this more general and make more more comfortable with what is here.

test/functional/tools/realtimetool_cellxgene.xml
```diff
@@ -0,0 +1,27 @@
...
+    <requirements>
+        <container type="docker">quay.io/galaxy/cellxgene-galaxy-ie:ie2</container>
+    </requirements>
+    <realtime>
+        <entry_point name="cellxgene_single_cell_visualisation">
+            <port>80</port>
+        </entry_point>
+    </realtime>
+    <command><![CDATA[
+        cp '${input}' '/tmp/galaxy_cellxgene_${input.element_identifier}.h5ad'
```
jmchilton: I'm sure you hate the best practice for dealing element_identifier Dan 😓, but please clean it the way @bgruening's tool does.

xref https://planemo.readthedocs.io/en/latest/writing_advanced.html#processing-identifiers
xref #7937

blankenberg: I mean sure, I will fix this, but this was from the code that @bgruening PR'd, just cleaned up a little bit with adding quotes after the merge conflict.

jmchilton:
Cherry-pick https://github.com/galaxyproject/galaxy/commit/3d2b6bdec54fb94c1480777b39e3f4ab55b5220e?

I really like JobContainerAssociation and I like the idea of just assigning all containers a name.

lib/galaxy/tools/deps/container_classes.py
```diff
@@ -263,12 +273,17 @@ def containerize_command(self, command):
             volumes_from=volumes_from,
             env_directives=env_directives,
             working_directory=working_directory,
-            net=self.prop("net", "none"),  # By default, docker instance has networking disabled
+            net=self.prop("net", None),  # By default, docker instance has networking disabled
```
jmchilton: This change seems wrong? Won't this disable disabling network access by default or was there a bug here already?

blankenberg: @bgruening was fixing a bug?

blankenberg:
> Cherry-pick [3d2b6bd](https://github.com/galaxyproject/galaxy/commit/3d2b6bdec54fb94c1480777b39e3f4ab55b5220e)?

Done and done, also you have far too many branches...

test/functional/tools/realtimetool_cellxgene.xml
blankenberg: re.sub'd in 87fed7e8eede87a6ab97e1f0862a9727d9a19880

jmchilton:
I really like tracking the container job association and thanks for cherry picking what I sent you Dan. That said, I think I dislike writing ``container_info``, modifying it, re-writing it. It seems to violate abstractions at that high level description, and digging into it - ``containerize_command`` modifying ``container_info`` seems odd and a violation also. That method should not be mutating state in my opinion and that side effect is non-obvious. It seems like as long as we make sure ``docker_host_props`` are preserved (maybe call it ``container_target_props`` for compatibility with what that would look like in Singularity)  on that container association we can get away without doing all of that:

So instead of pre-computing commands we may need for every job and violating the abstractions:

```python
        self.container_info['commands'] = {}
        for cmd in ['inspect', 'port', 'kill', 'stop']:
            self.container_info['commands'][cmd] = docker_util.build_docker_simple_command(cmd, container_name=self.container_name, **docker_host_props)
```

We could just add something like this to job_wrapper:

```python
     def container_command_ports(self):
          self._container_command('ports')

     def _container_command(self, command):
           container_assoc = self.get_job().container  # container_assoc
           return docker_util.build_docker_simple_command(command, container_assoc.container_name, container_assoc.container_target_props)
````

I'm nervous this following line was really confusing and involved changing a property with complex state already:

```python
        stored_container = job_wrapper.get_job().container
        if container and stored_container:
            if container.container_info:
                stored_container.container_info = container.container_info
```

It would become the following, which is clearer to me and which avoids re-writing container_info. We're writing a different property that didn't previously have state and only gets written in one place.

```python
        # Make sure target container is written for command generation.
        stored_container = job_wrapper.get_job().container
        if container and stored_container:
              stored_container.container_target_props = container.container_target_props
```

Before actually trying this out I was wondering how do these abstraction tweaks sound to you?

blankenberg:
> Before actually trying this out I was wondering how do these abstraction tweaks sound to you?

Abstractions are good :) Certainly makes sense to build those container interaction commands on the fly.

lib/galaxy/tools/deps/container_classes.py
bgruening: I think this fixed a bug in my testing. The former yielded to `--net=None` I think and not to ` ` which was expected.

test/functional/tools/realtimetool_cellxgene.xml
bgruening: Please treat all what I did as experimental. I just wanted to demonstrate that this kind of works for Jupyter, RStudio and where are limitations.

jmchilton:
https://github.com/galaxyproject/galaxy/commit/2876d48a88d1d0552bacbf64f39eabbaa9819218 Integration tests for this... too much depends on web controllers instead of API controllers to make more progress without either...

- Adding a new test suite that blends per-testcase Galaxy configs of integration testing with configuration of Selenium environment (a project I've wanted to do for a long time - found this WIP branch from a year ago - https://github.com/jmchilton/galaxy/tree/selenium_integration - one of those way too many branches). This test suite would really let us test whole new classes of awesome things we just cannot test currently and is very important.
- ***or*** Refactoring more this PR to work with the API controllers. Again this is probably a good idea anyway - we've wanted to avoid adding new non-API controllers for years now and generally push back against them.

I'll try to make some progress on one or the other.

jmchilton:
Following WIP commit fetches that entry point URI target from the API instead of generating it in the controller. I didn't quite understand what the index controller was doing and overshot the amount of controller code that can be deleted... but it is an approach I think we should work toward.

https://github.com/jmchilton/galaxy/commit/8b048095734b7170ddc9f3deb4268c6bac6c8f8d

Error handling will be more SPA-y and more functionality will be accessible to developers via the API - hopefully these are both appealing.

Tomorrow I'll see if I can get it working and then get that integration test working based on it, and then try to apply some refactorings I discussed above and some I have stuffed in local branches.

jmchilton:
VueJs component for polling status of RTT right on the execution result page, progress away from backbone and web controllers and toward Vue and API controllers...

https://github.com/jmchilton/galaxy/commit/7bcb1dfcc82f632b2c133a032465091cd1b9d326

blankenberg:
Awesome @jmchilton. Especially like the updating message on the tool submit success form (clicking did not work for me, however).

One thing that would be nice would be if the links were 'real' and allowed optionally opening into new tabs, etc with ctrl- or right-click.

jmchilton:
So... this only works in Chrome for me? I can generate one of your ... interesting... realtime proxy URLs and then use it in that tab or in an incognito browser for Chrome. But I can't use these URLs with wget, Python requests, Firefox, or Safari. @natefoo was skeptical about this localhost wildcard prefix trick - but maybe it only works in Chrome? I guess with a willcard actual DNS record this would work regardless of the Chrome magic?

Updated:

Work around for Python testing...

```
>>> requests.get("http://localhost:8080", headers={"Host": "realtimetoolentrypoint.9ce617df390851fc.b529bca87f604f559994b1e0b0848435.realtime.localhost:8080"}).content
'moo cow\n'```

**Comment by natefoo** [comment]
Wildcard DNS is going to break most people's SSL setups though. See my comment at the top about cookie path manipulation.

**Comment by blankenberg** [comment]
> So... this only works in Chrome for me? I can generate one of your ... interesting... realtime proxy URLs and then use it in that tab or in an incognito browser for Chrome. But I can't use these URLs with wget, Python requests, Firefox, or Safari. @natefoo was skeptical about this localhost wildcard prefix trick - but maybe it only works in Chrome? I guess with a willcard actual DNS record this would work regardless of the Chrome magic?
> 
> Updated:
> 
> Work around for Python testing...
> 
> ```
> >>> requests.get("http://localhost:8080", headers={"Host": "realtimetoolentrypoint.9ce617df390851fc.b529bca87f604f559994b1e0b0848435.realtime.localhost:8080"}).content
> 'moo cow\n'```
> ```

Well, (I think) it should work in all:
https://tools.ietf.org/html/rfc2606
```
      The ".localhost" TLD has traditionally been statically defined in
      host DNS implementations as having an A record pointing to the
      loop back IP address and is reserved for such use.  Any other use
      would conflict with widely deployed code which assumes this use.
```
And there is: https://tools.ietf.org/html/rfc6761
And https://tools.ietf.org/html/draft-west-let-localhost-be-localhost-06

But in just testing now, it worked for me in Opera and Chrome, but not Firefox or Safari.

For subdomains with wildcard A records, it was working for each browser that I tested. I did not test SSL certs, but it is possible to get wildcard SSL certs.

**Comment by blankenberg** [comment]
FWIIW: This is the change where I first swapped from using a path-based approach to subdomain: https://github.com/galaxyproject/galaxy/pull/7494/commits/a7c9ac042f41c9b0b416875b8cc1c676b5af30b9

It solved a bunch of issues I was having when testing features of various RealTimeTool example cases, and also seemed like a potentially good way to help with distributed-routing in the future.

With just a little more effort the path-based approach could likely be made to work well.

**Comment by jmchilton** [comment]
> One thing that would be nice would be if the links were 'real' and allowed optionally opening into new tabs, etc with ctrl- or right-click.

Ugh... yeah that makes sense but I can't do that cleanly from a click handler. I guess I'll make the grid active and poll the entry points and only enable the link when it becomes active.

**Comment by jmchilton** [comment]
1126d68fedc32c1644915addce951a945496d776 reworking toward that polling, pre-computing those links and placing them in the main API response.

I don't know about these links to be honest, they make me quite nervous. This proxy doesn't require the Galaxy cookie so those links feel less secure than GIE approach - I'm not sure if that is just a feeling though of if there is something to that 🤔. To get the user's API key from a GIE notebook, you need both the link to the session and the cookie I think... not just some permanent link that might be saved in a browser history or obtained from an XSS attack. Less relevant for tools that just produce single outputs though obviously...

**Comment by jmchilton** [comment]
- Proper container killing that should work across all job runners, removing the local runner code and container_info hacking I disliked the most  https://github.com/galaxyproject/galaxy/pull/7494/commits/70989ef32694dfbbd4bc3e63b0ac72624f004fb8.
- Delegate detecting ports to the job script, should make it much more easy to generalize this to other runners https://github.com/galaxyproject/galaxy/pull/7494/commits/0dfe1343681db04223c4f5696649a2a5dca0c9a4

**Comment by rdallet** [comment]
Hi Dan,

I have developed several Galaxy Interactive Environments and I am therefore very interested in Interactive Tools. I tested the latest version of your Pull Request but I get this error.

```
cp: cannot stat ‘/home/rdallet/galaxy/lib/galaxy/tools/test/functional/tools/default_notebook.ipynb’: No such file or directory
```

I have this file in `/home/rdallet/galaxy/test/functional/tools/default_notebook.ipynb` but not in `lib/galaxy/tools/...`. How can I fix this ?

**Comment by blankenberg** [comment]
> Hi Dan,
> 
> I have developed several Galaxy Interactive Environments and I am therefore very interested in Interactive Tools. I tested the latest version of your Pull Request but I get this error.
> 
> ```
> cp: cannot stat ‘/home/rdallet/galaxy/lib/galaxy/tools/test/functional/tools/default_notebook.ipynb’: No such file or directory
> ```
> 
> I have this file in `/home/rdallet/galaxy/test/functional/tools/default_notebook.ipynb` but not in `lib/galaxy/tools/...`. How can I fix this ?

Hi Romain,

Is this error when you are trying to use the example Jupyter notebook InteractiveTool found at `test/functional/tools/realtimetool_juypter_notebook.xml`, or your own notebook?

Thanks,

Dan

**Comment by rdallet** [comment]
> > Hi Dan,
> > I have developed several Galaxy Interactive Environments and I am therefore very interested in Interactive Tools. I tested the latest version of your Pull Request but I get this error.
> > ```
> > cp: cannot stat ‘/home/rdallet/galaxy/lib/galaxy/tools/test/functional/tools/default_notebook.ipynb’: No such file or directory
> > ```
> > 
> > 
> > I have this file in `/home/rdallet/galaxy/test/functional/tools/default_notebook.ipynb` but not in `lib/galaxy/tools/...`. How can I fix this ?
> 
> Hi Romain,
> 
> Is this error when you are trying to use the example Jupyter notebook InteractiveTool found at `test/functional/tools/realtimetool_juypter_notebook.xml`, or your own notebook?
> 
> Thanks,
> 
> Dan

The Jupyter notebook InteractiveTool example. (`test/functional/tools/realtimetool_juypter_notebook.xml`)

**Comment by blankenberg** [comment]
> The Jupyter notebook InteractiveTool example. (`test/functional/tools/realtimetool_juypter_notebook.xml`)

And you are loading the tool directly, or have you copied it?

From the xml, it calls
`cp '$__tool_directory__/default_notebook.ipynb' ./ipython_galaxy_notebook.ipynb &&`

where `$__tool_directory__` refers to the directory where the xml file is located, which should resolve to `test/functional/tools/default_notebook.ipynb`.

**Comment by rdallet** [comment]
> And you are loading the tool directly, or have you copied it?

I cloned your PR and pulled the latest version. I didn't change anything exept the config files (config/galaxy.yml , config/job_conf.xml, config/tool_conf.xml).

> From the xml, it calls
> `cp '$__tool_directory__/default_notebook.ipynb' ./ipython_galaxy_notebook.ipynb &&`
> 
> where `$__tool_directory__` refers to the directory where the xml file is located, which should resolve to `test/functional/tools/default_notebook.ipynb`.

OK, so `default_notebook.ipynb` is in `/home/rdallet/galaxy/test/functional/tools/`, but it's like `$__tool_directory__` refers to `/home/rdallet/galaxy/lib/galaxy/tools/test/functional/tools/` instead of `/home/rdallet/galaxy/test/functional/tools/`. Can be a configuration problem?

**Comment by blankenberg** [comment]
> OK, so `default_notebook.ipynb` is in `/home/rdallet/galaxy/test/functional/tools/`, but it's like `$__tool_directory__` refers to `/home/rdallet/galaxy/lib/galaxy/tools/test/functional/tools/` instead of `/home/rdallet/galaxy/test/functional/tools/`. Can be a configuration problem?

Maybe a general configuration problem? How are you loading the test tool? I am somewhat lazy, so in the `tool_conf.xml` file in my dev instance I just have:
```
  <section id="functest" name="Functional Test Tools">
    <tool_dir dir="/path/to/git/galaxy/test/functional/tools" />
  </section>
```
which loads all of the test tools and the jupyter InteractiveTool starts up OK for me.

**Comment by rdallet** [comment]
OK so it was a configuration problem. I used the documentation in `doc/source/admin/special_topics/` and I added `<tool file="../test/functional/tools/realtimetool_juypter_notebook.xml" />` in `tool_conf.xml`.
The problem was solved when I changed the tool file path to an absolute path. But I don't understand why the relative path doesn't work.

And now I have a second issue... This seems to be a docker one.
```
/opt/conda/lib/python3.5/site-packages/IPython/paths.py:61: UserWarning: IPython dir '/home/jovyan/.ipython' is not a writable location, using a temp directory.
  " using a temp directory.".format(ipdir))
  File "/opt/conda/bin/jupyter-trust", line 6, in <module>
    sys.exit(nbformat.sign.TrustNotebookApp.launch_instance())
  File "/opt/conda/lib/python3.5/site-packages/jupyter_core/application.py", line 267, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "/opt/conda/lib/python3.5/site-packages/traitlets/config/application.py", line 657, in launch_instance
    app.initialize(argv)
  File "<decorator-gen-6>", line 2, in initialize
  File "/opt/conda/lib/python3.5/site-packages/traitlets/config/application.py", line 87, in catch_config_error
    return method(app, *args, **kwargs)
  File "/opt/conda/lib/python3.5/site-packages/jupyter_core/application.py", line 243, in initialize
    self.migrate_config()
  File "/opt/conda/lib/python3.5/site-packages/jupyter_core/application.py", line 169, in migrate_config
    migrate()
  File "/opt/conda/lib/python3.5/site-packages/jupyter_core/migrate.py", line 241, in migrate
    with open(os.path.join(env['jupyter_config'], 'migrated'), 'w') as f:
PermissionError: [Errno 13] Permission denied: '/home/jovyan/.jupyter/migrated'
/opt/conda/lib/python3.5/site-packages/IPython/paths.py:61: UserWarning: IPython dir '/home/jovyan/.ipython' is not a writable location, using a temp directory.
  " using a temp directory.".format(ipdir))
  File "/opt/conda/bin/jupyter-notebook", line 6, in <module>
    sys.exit(notebook.notebookapp.main())
  File "/opt/conda/lib/python3.5/site-packages/jupyter_core/application.py", line 267, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "/opt/conda/lib/python3.5/site-packages/traitlets/config/application.py", line 657, in launch_instance
    app.initialize(argv)
  File "<decorator-gen-7>", line 2, in initialize
  File "/opt/conda/lib/python3.5/site-packages/traitlets/config/application.py", line 87, in catch_config_error
    return method(app, *args, **kwargs)
  File "/opt/conda/lib/python3.5/site-packages/notebook/notebookapp.py", line 1290, in initialize
    super(NotebookApp, self).initialize(argv)
  File "<decorator-gen-6>", line 2, in initialize
  File "/opt/conda/lib/python3.5/site-packages/traitlets/config/application.py", line 87, in catch_config_error
    return method(app, *args, **kwargs)
  File "/opt/conda/lib/python3.5/site-packages/jupyter_core/application.py", line 243, in initialize
    self.migrate_config()
  File "/opt/conda/lib/python3.5/site-packages/jupyter_core/application.py", line 169, in migrate_config
    migrate()
  File "/opt/conda/lib/python3.5/site-packages/jupyter_core/migrate.py", line 241, in migrate
    with open(os.path.join(env['jupyter_config'], 'migrated'), 'w') as f:
PermissionError: [Errno 13] Permission denied: '/home/jovyan/.jupyter/migrated'
```

**Comment by blankenberg** [comment]
> OK so it was a configuration problem. I used the documentation in `doc/source/admin/special_topics/` and I added `<tool file="../test/functional/tools/realtimetool_juypter_notebook.xml" />` in `tool_conf.xml`.
> The problem was solved when I changed the tool file path to an absolute path. But I don't understand why the relative path doesn't work.

This does sound like a bug. Would you mind opening an issue explaining the cause?

> And now I have a second issue... This seems to be a docker one.
> 

Yes, this looks to be a user-permission issue between the docker container and the docker runner / environment. If you modify the job destination to set the docker user to _empty_ (`<param id="docker_set_user"></param>`) then it should work. I've added a change to the `job_conf.xml.realtime` example config https://github.com/galaxyproject/galaxy/pull/7494/commits/e04e03d772dbe82eab9abf83c26562df84471760 that should fix this. 

A better/real fix is likely needed in the underlying docker container dependency for this tool.

Please let me know how it goes.

**Comment by rdallet** [comment]
> Yes, this looks to be a user-permission issue between the docker container and the docker runner / environment. If you modify the job destination to set the docker user to _empty_ (`<param id="docker_set_user"></param>`) then it should work. I've added a change to the `job_conf.xml.realtime` example config [e04e03d](https://github.com/galaxyproject/galaxy/commit/e04e03d772dbe82eab9abf83c26562df84471760) that should fix this.

It solves the problem but another permission issue appear.
```mkdir: cannot create directory ‘./jupyter’: Permission denied```

**Comment by jmchilton** [comment]
Can you add ``requires_domain`` (defaulting to ``False``) or something like that to the tools (``entry_point``) to keep a path-base proxy approach (instead of sub-domain based proxy) a potential default requirement for the future?

File: test/functional/tools/interactive_tool_juypter_notebook.xml [hunk 6]
```diff
@@ -9,14 +9,9 @@
...
     <environment_variables>
-        <environment_variable name="HISTORY_ID" strip="True">
-            #if $jupyter_notebook:
-                ${__app__.security.encode_id($jupyter_notebook.history_id)}
-            #else:
-                ${__app__.security.encode_id($out_file1.history_id)}
-            #end if
-        </environment_variable>
-        <environment_variable name="REMOTE_HOST">TODO</environment_variable>
+        <environment_variable name="HISTORY_ID" strip="True">${__app__.security.encode_id($jupyter_notebook.history_id)}</environment_variable>
```
jmchilton: We really need a way to inject this and something comparable to ``galaxy_infrastructure_url`` without ``__app__``. Willing to implement higher-level primitives for encoded history ID and inject ``galaxy_infrastructure_url`` with a per-job destination override option (https://github.com/galaxyproject/galaxy/pull/1688/commits/81b5ca564ac7cdd855cb74dad8e98abc83a8efdc). Or at least a warning in every sample that uses ``__app__`` that we will break backward compatibility in the near future

blankenberg:
> Can you add `requires_domain` (defaulting to `False`) or something like that to the tools (`entry_point`) to keep a path-base proxy approach (instead of sub-domain based proxy) a potential default requirement for the future?

Added in 41d7155fb6defa3ec83fdaaa93182b9935b6f53d

test/functional/tools/interactive_tool_juypter_notebook.xml
blankenberg: Added warning in 725016230ef56f1e9f25d3018f3643dbfcc539cf and 34ec3bfd3a76a852dcbc9e637af395501e23a5ba

jmchilton:
We should use ``false`` not ``False`` in XML probably (https://www.w3schools.com/xml/schema_dtypes_misc.asp), but yeah - looks great, thanks for adding ``requires_domain``!

This and the warnings and them being off by default really assuage a lot of my fears. I think I'm very close to being +1 on this for 19.09. I really am sorry for saying I would merge it sooner when I wasn't ready to and didn't have the time I thought I did to put into it.

Update: Trying to figure out the tests - are they transient problems?

Update 2: Last laugh is yours though because you cause me physical pain when you demo these being used in workflows or tell people to put them on the ToolShed. If I could somehow make my +1 contingent on you not doing those two things - I would - but I don't think that would be feasible. Despite that though - really fantastic work - I think many people are really excited for these.

jmchilton (approved):

bgruening:
Thanks John for merging.

lizhi16:
The Interactive Tool is an amazing function. Our team is very interested in it. But we don't find any tutorial about how to deploy the InteractiveTools or how to use it. Could you please provide some advice for us to instruct how to use it. Thanks!

bgruening:
@lizhi16 we added some initial docs here: https://github.com/blankenberg/galaxy/blob/aea9ed2a219f35ac8b94d11b51ca52ec73cdcb48/doc/source/admin/special_topics/realtimetools.rst

Let us know if this does not help.

huster16:
> @lizhi16 we added some initial docs here: https://github.com/blankenberg/galaxy/blob/aea9ed2a219f35ac8b94d11b51ca52ec73cdcb48/doc/source/admin/special_topics/realtimetools.rst
> 
> Let us know if this does not help.

Thanks for your help. Following this doc, we found `<tool file="../test/functional/tools/realtimetool_juypter_notebook.xml" />
<tool file="../test/functional/tools/realtimetool_cellxgene.xml" />` are not contained in the `../test/functional/tools/`. Only is `realtimetool_ethercalc.xml` in this dict. 

We add <tool file="../test/functional/tools/realtimetool_ethercalc.xml" /> in `config/tool_conf.xml` and enable `tool_config_file: config/tool_conf.xml,config/shed_tool_conf.xml` in galaxy.yml.

But when we run the galaxy with run.sh, we still don't find any realtime tools in website.

May I ask is that all tools in galaxy can run on a container, or just the customized tools can start with a container? such as realtimetool_cellxgene and realtimetool_juypter_notebook ...

Thanks!

PR #7497, opened 2019-03-09
Add Okta as OIDC authentication option

selten:
Basic idea: add (more/multiple) OIDC IdPs to Galaxy.

-  Changes to Python backend to more easily enable adding new OIDC IdPs.
-  Changes to Vue login template for having multiple OIDC IdPs.
-  Specialisations added for Okta which is however dependent on https://github.com/python-social-auth/social-core/pull/333.


VJalili:
Would you please build this PR based on the latest dev? PSA version is upgraded in https://github.com/galaxyproject/galaxy/pull/7463 so you may base backend implementation based off that.

config/oidc_backends_config.xml.sample
```diff
@@ -31,4 +31,15 @@
...
+
+    <provider name="Okta">
+        <client_id> ... </client_id>
+        <client_secret> ... </client_secret>
+        <redirect_uri>http://localhost:8080/authnz/okta/callback</redirect_uri>
+        <!-- Okta API URL, based on 2nd URL here: https://developer.okta.com/docs/api/resources/oidc#2-okta-as-the-identity-platform-for-your-app-or-api
+        This has subsequently had the '/v1/authorize' removed. In productive deployments, this will likely resemble:
+        https://${company}.okta.com/oauth2/${authServerId}/
+         -->
+        <api_url>https://dev-000000.oktapreview.com/oauth2/default</api_url>
```
VJalili: We're not expecting it to work with this endpoint, correct? if so, then I would not set it here, instead, I would replace it with `...` and provide a documentation on where/how they can get it.

lib/galaxy/authnz/managers.py
```diff
@@ -102,8 +105,14 @@ def _parse_idp_config(self, config_xml):
             'client_id': config_xml.find('client_id').text,
             'client_secret': config_xml.find('client_secret').text,
             'redirect_uri': config_xml.find('redirect_uri').text}
+
         if config_xml.find('prompt') is not None:
             rtv['prompt'] = config_xml.find('prompt').text
+        if config_xml.find('api_url') is not None:
+            rtv['api_url'] = config_xml.find('api_url').text
+        if config_xml.find('url') is not None:
```
VJalili, line 126: Could you please elaborate on what is this URL?
- it is not set it in the config sample;
- can we compose `api_url` given the `url`?

lib/galaxy/managers/configuration.py
```diff
@@ -61,6 +61,7 @@ def _defaults_to(default):
             'allow_user_creation'               : _defaults_to(False),
             'use_remote_user'                   : _defaults_to(None),
             'enable_oidc'                       : _defaults_to(False),
+            'oidc'                              : _defaults_to(self.app.config.oidc),
```
VJalili: probably a more intuitive name for this variable would be `oidc_idps`, or `oidc_backends`, or `oidc_identity_providers`.

VJalili (changes requested):

lib/galaxy/authnz/managers.py
selten: That's for possible future compatibility. In PSA e.g. GitHub Enterprise could make use of this. It's not used right now, but I don't see that it hurts either.

Adding the flexibility however, does seem like a good idea to me.

VJalili: Is `url` a substring of `api_url`? e.g., `url = http://xyz.com/` and `api_url = http://xyz.com/oauth2/default`?

selten: Not necessarily. Okta isn't using url. It depends on a per IdP basis.

VJalili: Could you please elaborate a bit more?

selten: I opted to create more generic code for adding more IdPs in the future. Unlike. E.g. Google where the domain and path are consistent, this doesn't have to be the case for commercial IdPs. Therefore allowing these values to be set allows more flexibility for possible future expansion of the IdPs if the arises. So this could also be used for GitHub (Enterprise), which for regular github uses 'github.com', but for the API uses 'api.github.com'. So by trying to automatically determine this, we'd need to write code for this flexibility for every IdP that would qualify. So in order to create flexibility, having both of these as configuration instead would result in the most flexible backend with the least amount of code required for specialisation towards a certain IdP.

selten:
@VJalili could you please review this again? The Okta side was finally implemented on the side of PSA so this PR can finally move from WIP to actual PR.

VJalili:
@selten  LGTM, thank you. 
Could you please add documentation to the following page? https://github.com/galaxyproject/galaxy-hub/tree/master/src/authnz

Please note that there are `use` (docs intended for users) and `config` (docs intended for admins) folders.

selten:
I'll write the documentation for this later, thanks for pointing it out to me

VJalili (approved):

config/oidc_backends_config.xml.sample
```diff
@@ -1 +0,0 @@
-../lib/galaxy/config/sample/oidc_backends_config.xml.sample
```
VJalili: Could you please revert this change?

VJalili (approved):

mvdbeek:
@selten can you rebase this ?

selten:
@mvdbeek i merged the changes from dev into this branch, is that also fine?

mvdbeek:
That's fine, but there are style issues in Login.vue, can you resolve those?

selten:
@mvdbeek should be fixed now.

mvdbeek:
Thank you @selten !

PR #7500, opened 2019-03-10
Add new metadata and data table tool validators

bernt-matthias:
added
- dataset_metadata_not_in_data_table
- value_in_data_table
- value_not_in_data_table

and a bit of documentation

changes rely in parts on my intuition of the code, ie might be wrong. some testing for all metadata/value vs data table validators would be nice .. ? 

ping @mvdbeek


jmchilton:
Thanks for the enhancements - they look good to me. Renamed PR for release notes, try to name PRs as release note lines.

PR #7507, opened 2019-03-11
Legacy History Import Testing

jmchilton:
API and unit tests for importing 19.01 exports, important for maintaining backward compatibility as newer export formats are added in #7367.

Breaking up #7217 (part of #7441) into even smaller pieces to try to understand which of them fails on Jenkins (they all pass locally for me). Bisection by PR ☹️ - this builds on the passing #7505.


mvdbeek (approved):
Looks good, but can you drop the commented code if you don't need it?

jmchilton:
It is uncommented in a follow up commit (https://github.com/galaxyproject/galaxy/pull/7441/commits/dd09af9fc4916ab4bff0a6795f0743c190777328#diff-30fa180161da339dabdecb66b3ac1852). All that testing code should be valid now but there is a bug fix needed to enable that and have everything green.

PR #7510, opened 2019-03-11
remove biostar integrations

martenson:
we stopped using it and removing it will lower our maintenance burden


dannon:
Is it an option to update the functionality to use help.galaxyproject.org instead of just dropping it?  Or is it not worth it?  I'm not informed enough to have an opinion on this one, but I'm curious -- I don't know how often this was actually used.

martenson:
> Is it an option to update the functionality to use help.galaxyproject.org

We run hosted. So this would at minimum require transfer to our own hosting and then subsequent code modification and then maintenance onward. It is of course doable but I doubt beneficial.

In my understanding the Galaxy->biostar integration was not used a lot. After few months from launching the help.halaxyproject.org seems to get similar amount of traffic and more new questions.

I am open to enhancing our support paths/integration but biostar is not the way for us anymore, which is what this PR is about.

dannon:
@martenson Thanks for the opinion, and, yeah I totally agree that our old biostar isn't what we want to be using right now, obviously.

I get that this PR is about stripping biostar integration, but much if not most of this could easily be more general purpose help integration regardless of target, I think.  That said, and what my question was getting at, was that if nobody was using this *style* of help then I'm totally happy to just dump it.  If it was valuable we should consider just keeping it disabled until we can update it.

nsoranzo:
Deprecate it in 19.05, drop it in 19.09 if no one screams?

jxtx:
Probably if we want this we need to redo it. We often (at usegalaxy.org) talk about how the bug reports often include questions that could be better handled at help.galaxy... I think that was the hope with this integration but that never really took off. We've talked about automagically searching the help forum on the report bug page and suggesting answers, etc, but that's not been done.

So, if this is not being used, I'm fine with purging it entirely because we probably need to do something else completely different.

dannon:
> if this is not being used ... we probably need to do something else completely different.

@jxtx Yeah, that's what I was trying to get a little more context on to help sort out the path forward.  I did a big of digging through our issues and it does seem like that's the case here, and a fresh approach down the road is best.

@martenson Thanks for cleaning this up!

dannon:
> if this is not being used ... we probably need to do something else completely different.

@jxtx Yeah, that's what I was trying to get a little more context on to help sort out the path forward.  I did a big of digging through our issues and it does seem like that's the case here, and a fresh approach down the road is best.

@martenson Thanks for cleaning this up!

martenson:
@nsoranzo I am sorry you did not get a response. How strongly do you feel about this going into 19.05? Personally I think this is was such a niche feature (without API) that it does not need deprecation cycle but I can very easily be persuaded otherwise. Just say the word.

nsoranzo:
@martenson No worries, I was mainly trying to mediate.

dannon:
@nsoranzo Thanks :)   I figured that was the case, didn't intend to jump the gun on the merge there.

PR #4840, opened 2017-10-20
User-based ObjectStore

VJalili:
_(1) Initially started at https://github.com/galaxyproject/galaxy/pull/4314; (2) all the commits of this PR were squashed into a single commit on Dec 11, 2019, the history of the changes are preserved via [this branch](https://github.com/VJalili/galaxy/tree/ubos_backup)._
## Introduction
This PR extends Galaxy's `ObjectStore` to enable users to _bring-their-own-resources_: users can _plug a media_ (e.g., Amazon S3 bucket) on which Galaxy will persist their datasets.

## Motivations
-	**unlimited storage:** users on Galaxy instances with limited storage resources (e.g., storage quota) can potentially have an unlimited storage by plugging their own (cloud-based) storage to Galaxy;
-	**data sharing:** having datasets generated by Galaxy stored on user’s cloud-based storage makes it easier sharing analysis results with collaborators;
-	**flexible persistence location:** members of different labs using a common Galaxy instance hosted at their institute can have their data stored on their lab’s network attached storage (NAS).

## Highlights
-	For users without a plugged storage media, Galaxy will continue to use an instance-wide configuration for their data storage needs;
-	A user's storage media (e.g., an S3 bucket) will be used for their data storage needs only, and will not be accessed for other user's storage needs; 
-	A _storage media_ can be a _local_ path, or an Amazon S3 bucket;
-	Users can plug multiple media (e.g., two different local path, and three Amazon S3 buckets), assign an `order` and `quota` attribute to each, and Galaxy will use them based on the given `order` and will fall from one to another if their quota limit is reached;
-	Leveraging the `order` attribute of _storage media_, users can use both instance-wide storage and their own media. For instance, they can direct Galaxy to use the instance-wide storage until their quota limit is reached (e.g., 250GB on Galaxy _Main_), then use their own media for the rest of their data storage needs. 
-	_Storage media_ are defined leveraging [Galaxy’s cloud authorization model]( https://doi.org/10.1093/bioinformatics/btz472), hence Galaxy does not ask for user’s credentials.
-	This PR implements all the necessary models, managers, functions, and APIs; and there will be a separate PR for UI;
-	The functionality is leveraging `Hierarchical` ObjectStore; hence, it is functional only if `Hierarchical` ObjectStore is configure. However, the hierarchy is applied instance-wide only, and does not affect user’s plugged media configuration;
[...]

PR #4314, opened 2017-07-14
User-based ObjectStore

VJalili:
This PR is in accordance with the _Federated user-based ObjectStore_ goal. The updates introduced in this PR are mock-up of the _Federated user-based ObjectStore_, and can generally be grouped as: 

1. **User-based ObjectStore**. Current ObjectStore reads its configuration (e.g., provider, bucket name, access and secret keys) from `config/object_store_conf.xml` and applies them Galaxy instance-wide. In other words, the data uploaded/created/processes by any of the users are written-to or read-from a common bucket to which the Galaxy instance has full access (accordingly, implicitly accessible by all the users of the galaxy instance). This PR introduces **user-based objectstore**, where each user defines his/her own `PluggedMedia` (e.g., AWS S3 bucket(s), Azure Blob, and etc.) where a `dataset` is read/written to that `PluggedMedia`.

2. **Provider agnostic ObjectStore**. This PR replaces the `S3ObjecStore` module with `cloud` module. The former uses `boto` and provides access to AWS S3 only; while, the latter leverages on `CloudBridge` which provides an interface for unified access to various providers (including S3 and Swift).

What is expected in future related PRs: 
- at this PR, only `cloud` module is user-based; other modules such as `DistributedObjectStore` are not (user should be able to define how to distribute his/her own data on the `PluggedMedia`s he/she has defined using a user-based `DistributedObjectStore`).

- some UI elements will be defined to allow user (a) define a `PluggedMedia`, (b) choose a `PluggedMedia` when uploading/creating a new dataset. At this PR, to define a `PluggedMedia`, one should manually add it to the database (`PluggedMedia` table). Additionally, the code has two temporary code-snippets (marked with `TEMP BLOCK`) which choose the first available `PluggedMedia` of the user.

PR #7512, opened 2019-03-12
Ensure metadata.properties entry exists for fastg.

VJalili:
ping @dannon @jmchilton 

Originally proposed by Dannon; see: https://github.com/galaxyproject/galaxy/pull/4840/commits/d584fee04c96f16d20fab6167e836a9360d3cc39


VJalili:
I think this patch may not be necessary, as there could be other reasons for the failing tests @#4314

PR #7514, opened 2019-03-12
Redefining Cel data type

bensellak:


bensellak:
@nsoranzo Here we go for the last round (For now) related to microarrays data types.

bensellak:
@galaxybot test this

bensellak:
@martenson can you check if it's good please :)

bensellak:
@jmchilton is it mergeable? thanks

martenson:
@galaxybot test this

martenson:
@galaxybot test this

bensellak:
@martenson thank you for your help, it is always the same issue (Files with mixed encoding) that tests can't support. The uploading is working locally  just fine, I removed the v4 file for the time being and converted to a supported encoding.

bensellak:
@martenson can we give it a try? thanks

bensellak:
@mvdbeek can you test this one? thanks

mvdbeek:
@galaxybot test this

bensellak:
@mvdbeek Thank you for running the tests. The code for sniffing the cell is quite simple and it is working locally for autodetection during upload. Any idea why is it failing here? I think the fact that the version 3 of the cel file is in text format (Defined as binary because of the two other versions 4 and agcc). Your help is very appreciated.

lib/galaxy/datatypes/test/affy_v_3.CEL
```diff
@@ -0,0 +1,74 @@
+[CEL]
```
mvdbeek: This file needs to end with `.cel ` for the test to work. Make sure to also adjust the doctests for the new filename.

bensellak: Done. Awesome thank you

mvdbeek:
@galaxybot test this

lib/galaxy/datatypes/test/affy_v_3.CEL
bensellak: Thanks @mvdbeek. Hope it will make it to 19.05.

bensellak:
@mvdbeek  is it fine? Is there something i should change?

bensellak:
@nsoranzo can you check if we can merge this one. Thank you

mvdbeek:
We will look at this when we have the time to do that, it isn't necessary to ping us.

bensellak:
Got it, No more pings. Thank you guys for your time (I know that you are very busy).

jmchilton:
So we've effectively moved this definition into microarrays.py, this is a better home but it has the real possibility of breaking people's datatypes_conf.xml with existing modifications. It'd be safer to move class definition back into binary.py and then import the class into microarrays.py.

```
# Keeping definition in binary.py for backward compatibility, but ref in datatypes_conf from here.
from galaxy.datatypes.binary import Cel
```

We've done this for other classes like (https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/datatypes/images.py#L243).

lib/galaxy/datatypes/microarrays.py
```diff
@@ -125,3 +126,66 @@ def set_meta(self, dataset, **kwd):
...
+        True
+        >>> fname = get_test_fname('test.gal')
+        >>> Cel().sniff(fname)
+        False
+        """
+        handle = open(filename, 'rb')
+        found_cel_4 = False
+        found_cel_3 = False
+        found_cel_agcc = False
+        header_bytes = handle.read(5)
```
nsoranzo: Can you use the `with` statement to open the file, please?
```python
        with open(filename, 'rb') as handle:
            header_bytes = handle.read(5)
```

lib/galaxy/datatypes/microarrays.py
```diff
@@ -125,3 +126,66 @@ def set_meta(self, dataset, **kwd):
...
+        elif header_bytes.decode("utf8", errors="ignore") == '[CEL]':
+            found_cel_3 = True
+        return found_cel_3 or found_cel_4 or found_cel_agcc
+
+    def set_meta(self, dataset, **kwd):
+        """
+        Set metadata for Cel file.
+        """
+        handle = open(dataset.file_name, 'rb')
+        header_bytes = handle.read(5)
```
nsoranzo: Can you use the `with` statement here too, please?

lib/galaxy/datatypes/microarrays.py
```diff
@@ -125,3 +126,66 @@ def set_meta(self, dataset, **kwd):
...
+        False
+        """
+        handle = open(filename, 'rb')
+        found_cel_4 = False
+        found_cel_3 = False
+        found_cel_agcc = False
+        header_bytes = handle.read(5)
+        if header_bytes.decode("utf8", errors="ignore")[0] == '@':
+            found_cel_4 = True
+        elif header_bytes.decode("utf8", errors="ignore")[0] == ';':
```
nsoranzo: Can you keep the previous test, which seems more specific?
```suggestion
        elif header_bytes == b';\x01\x00\x00':
```

lib/galaxy/datatypes/microarrays.py
```diff
@@ -125,3 +126,66 @@ def set_meta(self, dataset, **kwd):
...
+    edam_format = "format_1638"
+    edam_data = "data_3110"
+    MetadataElement(name="version", default="3", desc="Version", readonly=True, visible=True,
+                    optional=True, no_value="3")
+
+    def sniff(self, filename):
+        """
+        Try to guess if the file is a Cel file.
+        >>> from galaxy.datatypes.sniff import get_test_fname
+        >>> fname = get_test_fname('affy_v_agcc.cel')
```
nsoranzo: Is `lib/galaxy/datatypes/test/affy_v_agcc.cel` actually needed? It seems of the same subtype of `test.cel` .

Can you instead add a test file for the version 4?

lib/galaxy/datatypes/microarrays.py
```diff
@@ -125,3 +126,66 @@ def set_meta(self, dataset, **kwd):
...
+        False
+        """
+        handle = open(filename, 'rb')
+        found_cel_4 = False
+        found_cel_3 = False
+        found_cel_agcc = False
+        header_bytes = handle.read(5)
+        if header_bytes.decode("utf8", errors="ignore")[0] == '@':
+            found_cel_4 = True
+        elif header_bytes == b';\x01\x00\x00':
```
nsoranzo: I've researched this a bit more and found this: https://www.affymetrix.com/support/developer/powertools/changelog/gcos-agcc/generic.html

It seems the correct test is:
```suggestion
        elif struct.unpack("<bb", header_bytes[:2]) == (59, 1):
```

or the equivalent:
```suggestion
        elif header_bytes[:2] == b';\x01':
```

lib/galaxy/datatypes/microarrays.py
```diff
@@ -1,6 +1,7 @@
 import logging
```
nsoranzo, line 1: Need to add `import struct` here

bensellak: Done, I didn't want to surcharge sniffing with bytes unpacking. I will go with your suggestion.

nsoranzo: Thanks! You need to keep the empty line separating Standard Library imports from the others.

lib/galaxy/datatypes/microarrays.py
bensellak: Done

lib/galaxy/datatypes/microarrays.py
bensellak: Done

bensellak:
Cel first format was data text. The old defined cel only defined the agcc format wich is binary for sure. If I understand your comment, even if the tests passe, we still can break something (Workflow, tool)?. I am open to any suggestion or changes.

nsoranzo:
@galaxybot test this

bensellak:
Can you guys merge this one? any suggestions or modifications needed? thank you for your time.

mvdbeek:
Can you address @jmchilton's comment please ?

bensellak:
Sorry for tagging @jmchilton, but I don't understand very well your suggestion (I commented above but you didn't see it). Should I keep the Cel definition in binary (I mean the new definition) and keep the old ref in datatype_conf?, If so why import it in microarrays.py, if it's not used?.

mvdbeek:
The idea is to move `class Cel(Binary) ...` back into binary.py, and inside microarrays.py you import the class (`from .binary import Cel`). This means the datatype will be available for reference from both files. The rest of the PR you can leave as is.

lib/galaxy/datatypes/microarrays.py
```diff
@@ -1,6 +1,10 @@
...
 
 from galaxy.datatypes import data
+from galaxy.datatypes.binary import (
```
mvdbeek:
```suggestion
from galaxy.datatypes.binary import (  # noqa: F401
```

This is needed to silence the linter warnings.

lib/galaxy/datatypes/binary.py
```diff
@@ -98,32 +98,66 @@ def sniff(self, filename):
...
+    MetadataElement(name="version", default="3", desc="Version", readonly=True, visible=True,
+                    optional=True, no_value="3")
 
     def sniff(self, filename):
         """
-        Try to guess if the file is a CEL file.
-
+        Try to guess if the file is a Cel file.
         >>> from galaxy.datatypes.sniff import get_test_fname
-        >>> fname = get_test_fname('test.cel')
```
nsoranzo, line 112: I think you can remove the `test.cel` file from `lib/galaxy/datatypes/test/`

lib/galaxy/datatypes/binary.py
```diff
@@ -98,32 +98,66 @@ def sniff(self, filename):
...
         """
-        try:
-            header = open(filename, 'rb').read(4)
-            if header == b';\x01\x00\x00':
-                return True
-            return False
-        except Exception:
-            return False
+        with open(filename, 'rb') as handle:
+            header_bytes = handle.read(5)
```
```suggestion
            header_bytes = handle.read(8)
```

lib/galaxy/datatypes/binary.py
```diff
@@ -98,32 +98,66 @@ def sniff(self, filename):
...
+            found_cel_3 = True
+        return found_cel_3 or found_cel_4 or found_cel_agcc
+
+    def set_meta(self, dataset, **kwd):
+        """
+        Set metadata for Cel file.
+        """
+        with open(dataset.file_name, 'rb') as handle:
+            header_bytes = handle.read(5)
+        if header_bytes.decode("utf8", errors="ignore")[0] == '@':
```
```suggestion
        if struct.unpack("<ii", header_bytes[:9]) == (64, 4):
```

See http://media.affymetrix.com/support/developer/powertools/changelog/gcos-agcc/cel.html#V4

bensellak: In the first version, I used the unpack (big/little endian). Was worried about python2/3 byte to int conversion, so i kept what was working for both. Thank you for your time, very appreciated.

lib/galaxy/datatypes/binary.py
```diff
@@ -98,32 +99,66 @@ def sniff(self, filename):
...
+        with open(filename, 'rb') as handle:
+            header_bytes = handle.read(8)
+        found_cel_4 = False
+        found_cel_3 = False
+        found_cel_agcc = False
+        if struct.unpack("<ii", header_bytes[:9]) == (64, 4):
+            found_cel_4 = True
+        elif struct.unpack(">bb", header_bytes[:2]) == (59, 1):
+            found_cel_agcc = True
+        elif header_bytes[:5].decode("utf8", errors="ignore") == '[CEL]':
```
```suggestion
        elif header_bytes.decode("utf8", errors="ignore").startswith('[CEL]'):
```

lib/galaxy/datatypes/binary.py
```diff
@@ -98,32 +99,66 @@ def sniff(self, filename):
...
+        """
+        Set metadata for Cel file.
+        """
+        with open(dataset.file_name, 'rb') as handle:
+            header_bytes = handle.read(8)
+        if struct.unpack("<ii", header_bytes[:9]) == (64, 4):
+            dataset.metadata.version = "4"
+        elif struct.unpack(">bb", header_bytes[:2]) == (59, 1):
+            dataset.metadata.version = "agcc"
+        elif header_bytes[:5].decode("utf8", errors="ignore") == '[CEL]':
```
```suggestion
        elif header_bytes.decode("utf8", errors="ignore").startswith('[CEL]'):
```

lib/galaxy/datatypes/binary.py
```diff
@@ -26,6 +26,7 @@
 from galaxy.util import nice_size, sqlite
 from galaxy.util.checkers import is_bz2, is_gzip
 from . import data, dataproviders
+from galaxy.datatypes.data import get_file_peek
```
nsoranzo: This line should go between:
```python
from galaxy.datatypes import metadata
```
and:
```python
from galaxy.datatypes.metadata import DictParameter, ListParameter, MetadataElement, MetadataParameter

```

bensellak: Thanks.

nsoranzo:
@galaxybot test this

nsoranzo:
@galaxybot test py3

lib/galaxy/datatypes/binary.py
```diff
@@ -98,32 +99,66 @@ def sniff(self, filename):
...
+            dataset.metadata.version = "4"
+        elif struct.unpack(">bb", header_bytes[:2]) == (59, 1):
+            dataset.metadata.version = "agcc"
+        elif header_bytes.decode("utf8", errors="ignore").startswith('[CEL]'):
+            dataset.metadata.version = "3"
+
+    def set_peek(self, dataset, is_multi_byte=False):
+        if not dataset.dataset.purged:
+            dataset.blurb = "Cel version: %s" % dataset.metadata.version
+            dataset.peek = get_file_peek(dataset.file_name)
```
nsoranzo, line 158: For binary files (i.e. version 4 and agcc), this is probably not a good idea.

lib/galaxy/datatypes/microarrays.py
```diff
@@ -1,6 +1,10 @@
 import logging
 
 from galaxy.datatypes import data
+from galaxy.datatypes.binary import (  # noqa: F401
+    Binary,
+    Cel
+)
```
nsoranzo: You don't need `Binary` any more, so this could be changed to:
```python
from galaxy.datatypes.binary import Cel  # noqa: F401
```

lib/galaxy/datatypes/binary.py
bensellak: It will just put the default peek for Binary, I think.

nsoranzo: Ah, `get_file_peek()` returns `"binary file"` in that case, sorry for the noise.

nsoranzo (approved):

mvdbeek:
@galaxybot test this

nsoranzo:
Thanks @bensellak!

bensellak:
Thank you guys @nsoranzo  @mvdbeek  @jmchilton

PR #7516, opened 2019-03-12
Replace backbone tagging views with components

Nerdinacan:
This PR replaces the old backbone tagging instances with reactive Tagging components and unifies the design with the previous PR that replaced the mako-rendered Tagging elements.

**components/Tags**
Persists tag changes to the Vuex store, handles initialization and event handling. This is the tag you most likely want to use. 

**components/Tags/StatelessTags**
A stateless UI-only component. Changes to the styling of the tags themselves (to make it look more like the existing bootstrap elements or whatever is desired) should occur here.

**components/Badges**
I recreated the existing pill badge display to listen to the changes to the central tag store. They should update as tags are changed just like the tags do.

**components/Tags/mounts**
2 mount functions for instantiating the components in non-Vue environments like the python rendered pages, or from within existing Backbone views. 

**components/Tags/tagService**
The previous implementations saved tags to a variety of confusing API endpoints locations. In some cases, an entire Dataset or History object would be saved, in others it would hit a specific endpoint which only saved the tags themselves. I've structured the Tags component so that it accepts a data service as a property so we can easily swap this behavior if necessary.

**store/tagStore**
Central Vuex store holding the current tags for a particular object. The main advantage we're getting with this is we're starting to see reactivity across different regions of the application. If you change the history tags in one place, it will update everywhere else those tags are used. (Without a page reload.)


Nerdinacan:
This is actually ready. I can't edit PRs anymore, or I'd dress it up.

dannon:
@Nerdinacan I'll take a look today.

Nerdinacan:
You're the man

dannon:
I pushed one change mirroring https://github.com/galaxyproject/galaxy/pull/7270/commits/d6836c5998d0d63156fd9043cf3a714d8718aa27 from the previous PR.

This looks great overall, the last thing I'm trying to figure out is why the styles bloat quite so much, and if there's anything we can do about it -- it's about a 55% increase in the size of the built CSS module, an extra 350k (compressed) on the initial pageloads.

dannon:
Test is unrelated, passes locally.  Not sure what's going on with that one.  Thanks for fixing that last selenium test, looks good!

martenson:
@dannon it was probably running old code before @mvdbeek and @nsoranzo fixed it (since the circleci config present here does not run on merge). It should be good for new PRs..

PR #7517, opened 2019-03-12
Make sure ``temp_output_dir`` path is absolute

nsoranzo:
Fix `test_download_history` test during BioBlend Travis builds, see:

https://github.com/galaxyproject/galaxy/pull/7214#issuecomment-471558460

https://travis-ci.org/galaxyproject/bioblend/jobs/505250097

Note that we already do the same in other 3 places in the codebase.


jmchilton (approved):

jmchilton:
Thanks so much for tracking this down @nsoranzo.

PR #7518, opened 2019-03-13
Fix narrow style for favorites button in tool form header

guerler:


martenson:
thanks @guerler !

martenson:
thanks @guerler !

PR #7519, opened 2019-03-13
Create directory before copying file there

pvanheus:
`shutil.copy(temp_name, ofn)` causes an error if the directory containing `ofn` does not exists. This patch ensures that that directory exists before the copy is done.

### DIFF
--- a/lib/galaxy/tools/verify/__init__.py
+++ b/lib/galaxy/tools/verify/__init__.py
@@ -6,6 +6,7 @@ import hashlib
 import io
 import logging
 import os
+import os.path
 import re
 import shutil
 import tempfile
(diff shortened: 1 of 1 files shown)


lib/galaxy/tools/verify/__init__.py
```diff
@@ -97,6 +98,11 @@ def get_filename(filename):
         # if the server's env has GALAXY_TEST_SAVE, save the output file to that dir
         if keep_outputs_dir:
             ofn = os.path.join(keep_outputs_dir, filename)
+            out_dir = os.path.dirname(ofn)
+            try:
+                os.makedirs(out_dir)
+            except OSError as e:
+                log.debug('error creating {}, possibly because directory exists: {}'.format(out_dir, str(e)))
```
martenson: What about doing `if not os.path.exists(out_dir)` instead of try-except and logging?

pvanheus: Make sense - addressed in latest version of the patch.

nsoranzo:
@galaxybot test this

nsoranzo (approved):

PR #7521, opened 2019-03-13
Eliminate import dependency of datatypes on galaxy.web (url_for).

jmchilton:
Use app to find url_for at runtime. app objects created outside a web context can stub this out and throw an exception if display applications are used in that context (which wouldn't make sense anyway).

This is a step toward being able to break datatypes out into their own package and toward faster loading of datatype modules.

### DIFF
--- a/lib/galaxy/datatypes/genetics.py
+++ b/lib/galaxy/datatypes/genetics.py
@@ -97,7 +97,7 @@ class GenomeGraphs(Tabular):
                                                       dataset_id=dataset.id,
                                                       action='display_at',
                                                       filename='ucsc_' + site_name)
-                    display_url = "%s%s/display_as?id=%i&display_app=%s&authz_method=display_at" % (base_url, url_for(controller='root'), dataset.id, type)
+                    display_url = "%s%s/display_as?id=%i&display_app=%s&authz_method=display_at" % (base_url, app.url_for(controller='root'), dataset.id, type)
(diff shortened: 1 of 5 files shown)


lib/galaxy/datatypes/genetics.py
```diff
@@ -94,10 +93,10 @@ def ucsc_links(self, dataset, type, app, base_url):
...
                     site_url = site_url.replace('/hgTracks?', '/hgGenome?')  # for genome graphs
-                    internal_url = "%s" % url_for(controller='dataset',
-                                                  dataset_id=dataset.id,
-                                                  action='display_at',
-                                                  filename='ucsc_' + site_name)
+                    internal_url = "%s" % app.url_for(controller='dataset',
+                                                      dataset_id=dataset.id,
+                                                      action='display_at',
+                                                      filename='ucsc_' + site_name)
                     display_url = "%s%s/display_as?id=%i&display_app=%s&authz_method=display_at" % (base_url, url_for(controller='root'), dataset.id, type)
```
nsoranzo:
```suggestion
                    display_url = "%s%s/display_as?id=%i&display_app=%s&authz_method=display_at" % (base_url, app.url_for(controller='root'), dataset.id, type)
```

nsoranzo (approved):

mvdbeek (approved):

PR #7523, opened 2019-03-13
[19.01] Fix routing when serving Galaxy at a prefix

### LINKED ISSUE #7162: Issues with prefix path (/galaxy)

dannon:
More galaxy-at-prefix issues fixed w/ backbone router.  Fixes #7162 (and more)
Fixed a CORS issue when welcome was external.
Prettier URLs by dropping the `__identifier` cruft after navigation.  Still annoying, but this makes nicer URLs until we have the vue router infra in.

### DIFF
--- a/client/galaxy/scripts/layout/router.js
+++ b/client/galaxy/scripts/layout/router.js
@@ -19,10 +19,12 @@ var Router = Backbone.Router.extend({
             .toString(36)
             .substr(2);
         url += url.indexOf("?") == -1 ? "?" : "&";
-        url += $.param(data, true);
+        let bustParam = $.param(data, true);
+        url += bustParam;
         let Galaxy = getGalaxyInstance();
         Galaxy.params = data;
         this.navigate(url, { trigger: true });
+        window.history.replaceState(window.history.state, "", window.location.pathname.replace(bustParam, ""));
     },
 
     /** override to parse query string into obj and send to each route */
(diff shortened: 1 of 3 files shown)


dannon:
I think the CircleCI build failure here is unrelated.  It occurs on all pre-dev PRs, right?

martenson:
@dannon correct, circleci atm tests head of fork's branch so older branches won't have the config

dannon:
@martenson Ok, cool, thanks for confirming that.

jmchilton (approved):

client/galaxy/scripts/layout/router.js
```diff
@@ -12,17 +13,27 @@ var Router = Backbone.Router.extend({
         this.options = options;
     },
 
+    executeUseRouter: function(url) {
```
martenson, line 16: I am not sure about the naming here, what is a 'use router'?

dannon: It refers to the `__use_router__` navigation definition we use, this function executes the navigation intended by those targets.

PR #7524, opened 2019-03-13
Decompose Galaxy into Python libraries in-tree.

### LINKED ISSUE #1396: Break project into namespaces
I'm going to let natefoo break it down but I guess it might make sense to start with something like galaxy-lib and galaxy-app with the further intention to break galaxy-app into galaxy-business-logic (but with a better name), galaxy-app, and galaxy-web.

jmchilton:
Alternative to galaxy-lib codebase synchronization - via script and manual fixes to divergences.

Fixes https://github.com/galaxyproject/galaxy/issues/1396.

This initial PR creates three new Python "projects" pairing new ``setup.py`` and ``requirements.txt`` files with symlink-ed directories in the core Galaxy project. The three projects are galaxy-util (just ``galaxy.util`` and ``galaxy.exceptions``), galaxy-objectstore (galaxy.objectstore), and galaxy-data (``galaxy.datatypes``, ``galaxy.model``, and ``galaxy.security``). The goal would be someday decompose ``galaxy.datatypes`` and ``galaxy.model`` into separate projects I think... but that is hard to do because datatype definitions depend on the metadata stuff and the metadata stuff depends on the models.

Decoupling those two has been a project I've tried to tackle before and sort of failed at. With  https://github.com/galaxyproject/galaxy/commit/5756f6059720d39a75e5786283b232787a60f14e#diff-396ad58d05f7d2d61c8d9e107633db37 I tried to make it clear that datatypes depended on the models - because there used to be all that complex model logic JobExternalOutputMetadataWrapper. But JobExternalOutputMetadataWrapper never belonged in either of these directories really and has been refactored out all together (https://github.com/galaxyproject/galaxy/commit/8570db46fd5cf9669315376041e80bc733255025#diff-07df768c6572ad942b290901fe1424b3) - so now it is really looking like we should do that opposite and just declare models depend on datatypes. Perhaps undoing 5756f6059720d39a75e5786283b232787a60f14e, conditionally import ``galaxy.model`` in metadata and modify that ``FileParameter`` to behave differently if galaxy.model.MetadataFile is unavailable.

Maybe it is not worth creating this galaxy-data project if we really want it to be two separate projects before we even start...

TODO (Not necessarily in this PR):
- [ ] tox setups for unit testing and flaking these things as units (ensure dependency structure is maintained for each new PR)
- [ ] Pipfiles?
   - [ ] Figure out how to make these depend on each other? (maybe... galaxy-data -> galaxy-util>19.5,<19.9).
- [ ] docs for each project? (needed - when would it be useful and when wouldn't it)?
- [ ] Refactor galaxy.tools subset in galaxy-lib into ``galaxy.toolutil`` and then create a ``galaxy-toolutil`` project.
- [ ] Refactor galaxy.jobs.metrics into ``galaxy.jobmetrics`` and then create a ``galaxy-jobmetrics`` project.
- [ ] Create a ``galaxy-containers`` project for ``galaxy.containers``.
- [ ] Bring in ``Makefile`` for publishing these, get exceptions in Galaxy project governance for tagging these things and pushing new versions (restrict changes to things in the ``project/`` folder - so all code still needs to be reviewed).


projects/datatypes/requirements.txt
```diff
@@ -0,0 +1,5 @@
+galaxy-lib
+pysam
+numpy
+bx
+h5py
```
nsoranzo: Sort alphabetically?

projects/util/requirements.txt
```diff
@@ -0,0 +1,7 @@
+boltons
+six>=1.9.0
+docutils
+pyyaml
+packaging
+markupsafe
+
```
nsoranzo: Sort and remove empty line?

projects/datatypes/requirements.txt
jmchilton: Sort alphabetically except put the galaxy-* dependencies on top? This is what I did now, let me know if you think that is okay.

mvdbeek:
> Maybe it is not worth creating this galaxy-data project if we really want it to be two separate projects before we even start...

I think we do want separate `galaxy.data` into `galaxy.datatypes` and `galaxy.model` at some point, but if that's not going to happen in this release cycle I'd be in favor of moving ahead with the current structure. Having `galaxy.data` releases should already make external metadata setting a lot easier to set up in cluster and container environments by just adding a dependency on this via our normal dependency resolution.

mvdbeek (approved):

nsoranzo:
Should we use `packages` as directory name, since the goal is to upload them to the Python Package Index?

jmchilton:
Renamed directory as requested, good catch 🦅👀

.gitignore
```diff
@@ -150,6 +150,11 @@ client/styleguide
 # webpack stats
 client/webpack-stats.json
 
+# Projects
+projects/*/build
+projects/*/dist
+projects/*/*.egg-info
+
```
natefoo, line 157: Dropping by to post-merge bikeshed... these should be `packages` now?

lib/galaxy/util/__init__.py
```diff
@@ -1524,7 +1524,10 @@ def safe_str_cmp(a, b):
 
 
 def galaxy_directory():
-    return os.path.abspath(galaxy_root_path)
+    root_path = os.path.abspath(galaxy_root_path)
+    if os.path.basename(root_path) == "projects":
```
martenson, line 1528: and here

.gitignore
martenson: https://github.com/galaxyproject/galaxy/pull/7656

PR #7526, opened 2019-03-14
try running on merge rather than head of PR's fork

martenson:


martenson:
this should have targeted my fork, sorry for the noise

Issue #7425, opened 2019-02-28
Idea: separate out user release notes from dev/deployer/admin


hexylena:
I think there's probably a use case for two sets of release notes:

- things admins / devs / power users care about (e.g. internal workflow enhancements for some bug they'd encountered)
- things end users care about, with lots of screenshots / gifs / etc. More like [github's blog post](https://github.blog/2019-02-14-introducing-draft-pull-requests/)

I'd propose that we include an additional file in the releases, `xx.yy_announce_user`, which would have contents like the following, an example based on 19.01's. I envision it as something admins can link to from their galaxy home pages or blogs or newsletters that they send out to their users after an upgrade, covering the highlights in a visual manner, that would be most accessible for users.

It would take some energy to do this, I don't have a concrete plan for making that easier. In the short term I can volunteer since I have a vested interest in making sure my users know about cool new features with gifs and pictures and things that will help them understand the new feature. E.g. workflow inputs are super cool! But the user is left to their own devices to figure out where in the interface that could be located. (Ok one of the PRs potentially included screenshots but those images are broken now.) Something short of having to develop a full training on how to use this, just a couple of screenshots or a quick gif explaining the feature, as needed, all centralised in a single document admins could share with their communities.

# January 2019 Galaxy Release (v 19.01)

**Revised UI Style**

  The Galaxy stylesheet has been substantially reworked resulting in the largest visual refresh to Galaxy in years.

![image](https://user-images.githubusercontent.com/458683/53549988-9459f000-3b2d-11e9-9916-9e060b388e19.png)

**More Colourful Tags**
  Whenever you use dataset name tags (`#tag`), they will be assigned a random color consistently, to increase visual distinction between tags, making it easier to track data flow in complex histories.

![image](https://user-images.githubusercontent.com/458683/53550095-cff4ba00-3b2d-11e9-9d74-567feed77ffe.png)

**Extensive Workflow Enhancements**

The workflow editor now features explicit step parameter nodes for non-file data (integers, strings, booleans, etc) and the ability to connect these to all tool inputs - not just files.

The workflow editor now features "Simple inputs for workflow logic" under the "Inputs" section, which allows creating "tools" that can act as number, boolean, text, or color inputs.

![image](https://user-images.githubusercontent.com/458683/53550272-3aa5f580-3b2e-11e9-833b-dc5ab678e7a7.png)

The workflow editor allows you to specify that specific inputs should be exposed, and can thus be connected to the new workflow inputs.

[...]

PR #7527, opened 2019-03-14
[19.01] User facing release notes

hexylena:
Fixes https://github.com/galaxyproject/galaxy/issues/7425 :)

I've made an attempt at updating `bootstrap_history.py` but I am ... quite unconfident in my changes. :(

I have not linked to the user announcements from anywhere, not sure where that would happen. I volunteer to write these for the foreseeable future.


dannon:
I definitely like the presentation here, seems more useful for the majority of readers than the other format.

I initially thought we might want to come up with some other approach for images, but I don't know that it's (~200k a few times per year) worth worrying about to try to do anything other than include them in the repo as you've done here.

hexylena:
yeah, just a different (and maybe larger) target audience, I still find utility in the full release notes, I skim them every time for interesting things that have been included. Hopefully something easily consumable/redistributable by admins and community managers.

I ran optipng on all of these, given the width of the documentation could potentially shave a few more bytes off but yeah maybe not worth it to either do that or to explore s3 type options.

hexylena:
@dannon any more comments or are you OK with this? :)

dannon:
@erasche Oh, yeah, sorry that wasn't clear -- I like them.  I have not tried to run your changes to the bootstrapper, but I can check that out later if you'd like.

hexylena:
Sure, if you have time, that'd be great. If not, I can drop my changes to the bootstrap script? I'd like to be able to link people to the notes when we do manage to upgrade at least.

jmchilton:
These are no doubt nice and I understand the appeal of targeting a different audience. Selfishly, I just don't love creating more work indefinitely. Is Freiburg willing to commit to writing these going forward or does this fall on the release manager? If this is merged, we can automate picking out the datatype changes and tool changes using tags and pre-populate some of that stuff.

hexylena:
> Is Freiburg willing to commit to writing these going forward or does this fall on the release manager?

I am willing to commit to writing the highlights, if we can automate the rest that would be brilliant. Even for the this PR, it was nice to have a reason to explore the WF parameters, I might not have for another few months otherwise.

jmchilton:
If you do that write up, I'll do the rest and automate what I can for 19.05. You can have my +1 I suppose, thanks!

jmchilton (approved):

dannon:
@erasche Pushed one bug fix and one bit of cleanup, looks like it'll work now to me.  (But I guess @jmchilton will take care of it when 19.05 rolls around if not ;))

hexylena:
Thanks for the fixes @dannon!

martenson:
How is this going to be presented? A link from the general release notes?

hexylena:
@martenson per comments in OP, no idea. We could link from general, I thought about something like "[Release notes]() to send to your users" but didn't have a good place to put that that wouldn't either be completely skipped over, or way too prominent.

hexylena:
Maybe we leave it alone this release, and next release if I write some nice notes we can add it as a 4th highlighted feature? And then in future just a small link somewhere of "user facing release notes"

martenson:
I wouldn't be afraid of giving it prominent spot - as a highlight or link above the highlights.

PR #7528, opened 2019-03-14
Remove galaxy.web.buildapp legacy.

jmchilton:
I think it would take a very old ini to reference this, I think everyone is pointing at a newer entrypoint.


jmchilton:
Looks like it was the default as late as 16.07, nevermind...

PR #7532, opened 2019-03-14
Static plugin staging as a part of client build instead of startup

### LINKED ISSUE #7231: Move client plugin staging back to to client build process
Having it in common_startup was a temporary move to avoid requiring node.js, but since we now require node, and since common_startup isn't actually common to all deployments, we want to move this back to client build infra which *is* now common to all deployments.

Bonus points for handling externally located and/or multiple configurable viz plugin dirs.

(this could also potentially tie into my plans for next-gen viz plugin installation via npm, etc)

dannon:
Fixes #7231, makes a few other incremental upgrades.

Plugin static staging has been removed from common_startup.  It's now a part of the gulp build stages of the standard `make client{-variant}`.

It's also accessible at `yarn run gulp plugins`, if developers or deployers need an isolated staging run.  It's intentional that plugin_staging.py` is staying, for now, until I'm sure it's not needed in the docker containers, etc.


dannon:
(yeah, sure enough, even closing and reopening you can't put something back in 'GitHub Draft' form -- was worth a test)

dannon:
Not sure why this toolshed test is stuck in running, but it should be completely unrelated to this PR.

PR #7536, opened 2019-03-15
Small object store fixes.

jmchilton:
Both are follow up on #7154.  Found in work on #7058.

### DIFF
--- a/lib/galaxy/objectstore/__init__.py
+++ b/lib/galaxy/objectstore/__init__.py
@@ -374,6 +374,7 @@ class DiskObjectStore(ObjectStore):
                     rel_path = os.path.join(rel_path, extra_dir)
             path = os.path.join(base, rel_path)
         if not dir_only:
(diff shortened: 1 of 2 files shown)


lib/galaxy/objectstore/__init__.py
```diff
@@ -374,6 +374,7 @@ def _construct_path(self, obj, old_style=False, base_dir=None, dir_only=False, e
                     rel_path = os.path.join(rel_path, extra_dir)
             path = os.path.join(base, rel_path)
         if not dir_only:
+            assert obj_id is not None, "Effective dataset id consumed by object store must be set before filename used"
```
mvdbeek: before filename can be used / determined / returned / is usable ?

jmchilton: ``"The effective dataset identifier consumed by object store [%s] must be set before a path can be constructed." % (self.store_by)``?

mvdbeek: Perfect!

PR #7539, opened 2019-03-15
run circleCI on merge rather than head of PR's fork

martenson:
this is taking advantage of github providing refs to preliminary merges of PRs

this is a missing feature in circleci: https://circleci.com/ideas/?idea=CCI-I-431


martenson:
well circleCI won't build it since it already built the hash for my fork when I was testing this (but it does not display results here)

PR #7540, opened 2019-03-15
History Import/Export Fixes and Enhancements Ahead of New Format

jmchilton:
See individual atomic commits for details. Focus is on testing and simplifying and fixing import and export of history within the context of the current format.

### DIFF
--- a/lib/galaxy/tools/imp_exp/__init__.py
+++ b/lib/galaxy/tools/imp_exp/__init__.py
@@ -429,15 +429,15 @@ class JobExportHistoryArchiveWrapper:
             job_attrs['update_time'] = job.update_time.isoformat()
 
             # Get the job's parameters
+            params_dict = {}
             try:
                 params_objects = job.get_param_values(self.app)
+                for name, value in params_objects.items():
+                    params_dict[name] = value
             except Exception:
                 # Could not get job params.
-                continue
+                pass
 
-            params_dict = {}
-            for name, value in params_objects.items():
-                params_dict[name] = value
             job_attrs['params'] = params_dict
 
             # -- Get input, output datasets. --
--- a/lib/galaxy/tools/imp_exp/__init__.py
+++ b/lib/galaxy/tools/imp_exp/__init__.py
@@ -258,23 +258,24 @@ class JobExportHistoryArchiveWrapper:
     archive.
     """
 
-    def __init__(self, job_id):
+    def __init__(self, app, job_id):
+        self.app = app
         self.job_id = job_id
+        self.sa_session = self.app.model.context
 
-    def get_history_datasets(self, trans, history):
+    def get_history_datasets(self, history):
         """
         Returns history's datasets.
         """
-        query = (trans.sa_session.query(trans.model.HistoryDatasetAssociation)
-                 .filter(trans.model.HistoryDatasetAssociation.history == history)
+        query = (self.sa_session.query(model.HistoryDatasetAssociation)
+                 .filter(model.HistoryDatasetAssociation.history == history)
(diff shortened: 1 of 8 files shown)


lib/galaxy/tools/imp_exp/__init__.py
```diff
@@ -202,36 +206,29 @@ def as_hda(obj_dct):
...
+                        if isinstance(obj, dict) and obj.get('__HistoryDatasetAssociation__', False):
+                            return (k, hdas_by_key[obj[object_key]].id)
+                        return (k, obj)
 
-                        def default(self, obj):
-                            """ Encode an HDA, default encoding for everything else. """
-                            if isinstance(obj, model.HistoryDatasetAssociation):
-                                return obj.id
-                            return json.JSONEncoder.default(self, obj)
+                    from boltons.iterutils import remap
```
mvdbeek: This should be imported at the top, right ?

lib/galaxy/tools/imp_exp/export_history.py
```diff
@@ -90,16 +89,18 @@ def create_archive(history_attrs_file, datasets_attrs_file, jobs_attrs_file, out
 
         # Status.
         print('Created history archive.')
+        return 0
     except Exception as e:
         print('Error creating history archive: %s' % str(e), file=sys.stderr)
+        return 1
```
mvdbeek, line 95: What do you think about being explicit here and doing `sys.exit("Error creating history archive: %s" % str(e))` ?
This will print to stderr and exit with 1

mvdbeek:
@galaxybot test py3

lib/galaxy/tools/imp_exp/export_history.py
jmchilton: I prefer non-explicit exits, feels like an implementation detail that should be left to the interpreter. The way it is structured now makes it easier to use (and test) as a library function - I think the unit tests would probably break if I did this?

lib/galaxy/tools/imp_exp/__init__.py
jmchilton: Yup - I'll rebase with a fix for that and resolve this.

test/unit/tools/test_history_imp_exp.py
```diff
@@ -101,6 +109,131 @@ def test_history_import_abspath_in_metadata():
...
+                                         compressed=True)
+    wrapper = JobExportHistoryArchiveWrapper(app, 1)
+    wrapper.setup_job(jeha)
+
+    from galaxy.tools.imp_exp import export_history
+    ret = export_history.main(["--gzip", jeha.temp_directory, dest_export])
+    assert ret == 0
+
+    _, imported_history = import_archive(dest_export, app=app)
+    print(_)
```
mvdbeek, line 232: Do you want to keep this in ?

mvdbeek (approved):

test/unit/tools/test_history_imp_exp.py
jmchilton: Nope, but I'll rebase a fix into #7367 - this PR is so green and approved 😆.

mvdbeek: Sounds good.

yvanlebras:
Just a comment following discussion on gitter with Martin. It appears to me that it can be of interest to propose several options to export datasets from a history from a Galaxy instance 1 and import to another Galaxy instance. For example, if people "just" want to export datasets and import it to another instance, one way can be to extract something like a text file with dataset names + URL like shown in the capture here
Capture

PR #7541, opened 2019-03-15
Refactor output_collect.py toward a database session-less modality.

jmchilton:
Dynamic output discovery/collection (output_collect.py) happens in the job finish code on the Galaxy server, but we'd like the option to run this code and publish results to the object store as part of the job script on the worker node. On the worker node, no database session would be configured - so we need a modality of this logic that works even if a database session is unavailable. This PR does not do that but it contains a series of atomic refactorings that isolate all the session requiring code nicely - allowing downstream functionality to be added that allows this cleanly.

The huge change downstream that builds on this PR looks something like (https://github.com/galaxyproject/galaxy/pull/7058/commits/21ed487153b6b61d4412c2f4078c079a74bb0ab3) currently. It reuses the abstractions here for discovering datasets and library folders but plugs in an alternative implementation for "persisting" the resulting model objects. A model export store will back that implementation but needs to be introduced as part of the history import/export work (https://github.com/galaxyproject/galaxy/pull/7058/commits/a88ecb291b3b315b1be00097e8b62ddd680179f9).

### DIFF
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -1696,6 +1696,7 @@ class Tool(Dictifiable):
             permission_provider,
             metadata_source_provider,
             input_dbkey,
+            object_store=tool.app.object_store,
         )
         collected = output_collect.collect_primary_datasets(
             job_context,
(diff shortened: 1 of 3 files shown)


mvdbeek (approved):
Very nice!

PR #7542, opened 2019-03-15
Drop d3 from publicly provided libs

dannon:
This works towards a goal for 19.05 where we're no longer providing these for arbitrary external consumers.  The one external consumer in the codebase was aequatus, so I've frozen the libs version that is known compatible and moved it there.  This reduces the bundle size while still providing functionality for aequatus as it moves to the new packaging approach we're now using.

(side note: there are going to be several of these PRs, but I'm keeping them atomic to make testing easier so that things don't get lost)


nsoranzo:
Ping @anilthanki

dannon:
@nsoranzo Thanks for that ping!  Yeah, this should continue to work exactly as it was for Aequatus, it'll just be using a (minified) d3 directly out of its own plugin directory instead of out of the deprecated scripts/libs dir.  Please let me know if it doesn't seem to, in your hands, and I'll fix it up.

config/plugins/visualizations/aequatus/templates/aequatus.mako
```diff
@@ -21,11 +21,8 @@
         ${h.javascript_link( app_root + "aequatus-vis/scripts/jquery/js/jquery-migrate-1.4.1.min.js" )}
         ${h.javascript_link( app_root + "aequatus-vis/scripts/scriptaculous/prototype.js" )}
 
-
-## install shared libraries
-        ${h.js( 'libs/d3')}
-
 ## aequatus-vis
+        ${h.javascript_link( app_root + "aequatus-vis/scripts/d3.min.js" )}
```
anilthanki: Hi Dannon,

I think you forgot to add this d3.min.js

Thanks

Anil

dannon: @anilthanki It should be there, let me doublecheck.  Ahh!  Looks like the path isn't quite right, let me fix it.  Thanks for taking a look.

dannon:
@anilthanki Thanks for taking a look, everything working the same as it was now?

anilthanki:
Yes, it's working now...
Thanks...

PR #7544, opened 2019-03-15
Client dep upgrades 19 03

dannon:
Full pass on dependency upgrades.  Everything was fairly straightforward, though bootstrap 4.3 has complications w/ popovers that I'm going to try to upgrade/address separately with a few other new bootstrap features.


dannon:
This should be good to go now.

martenson:
Do you think squashing this is reasonable?

dannon:
I'd rather not, it's intentionally split out because it made it easier to find/fix bugs with particular libs (using bisect and a script that ran the client tests repeatedly, already came in handy).

Though, I guess, if the number of commits is a concern (though I don't think it ever has been before -- everyone seems to have their own style and preference for this on the team), I can try to merge more of the 'safer' upgrades together moving forward.

dannon:
Rebased to integrate, but it looks like that unearthed a new, exciting dependency conflict.  Fixed now in
 04f59cb, should be good to go.

dannon:
@galaxybot test this

dannon:
Hrmm.  Travis doesn't want to run again on this?  It was passing previously, but in the rebase shuffle it no longer wants to rerun.

(tests are running now, just slow, I guess -- or someone kicked a box)

martenson:
thanks @dannon !

dannon:
@martenson Thank you!

PR #7545, opened 2019-03-16
Allow Admins to Import Dynamic JSON-based Tools

jmchilton:
- Introduce models and a API for creating tools dynamically. No toolbox is really needed, after import these tools are loaded from the database as needed for jobs and workflows.
 - Allow referencing such tools via a UUID established by Galaxy on import. 
 - Allow tools without ids for such use cases.
 - Extend Format 2 workflow definitions to allow embedding tools directly into workflows, either directly or using a CWL-style ``@import`` syntax.

Test cases included:
 - Running a tool as a regular user imported by an admin both via tool id and via UUID.
 - Exporting and re-importing workflows with such tools works properly (for admins).
 - Using the format2 embedded tool and ``@import`` syntax. 
 - Security tests that ensure only admins can import tools - either stand-alone as as part of workflows.


mvdbeek:
This looks very very interesting, is this primarily for running CWL workflows or do you see this as a valid way of shipping small-ish, one-shot Galaxy tools as well ?

jmchilton:
I would say this is the basis for that, absolutely. I think it is really an atomic and concrete step more than a final product though.

My sales pitch would be something mirroring the Anton sales pitch for [Jupyter GIEs](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1005425). Typically some percent of an analysis is doable with Galaxy tools - maybe it is 75%, maybe 95% - but for most novel analyses there are bits at the end and perhaps in the middle that require seemingly ad-hoc bits of computation. Jupyter notebooks are one approach, but another approach using actual tools is possible as well - building more closely on our traditional strengths. If we can transition Galaxy to allow user installation of tools (this in return requiring personal toolboxes, forced containers, sandbox evaluation of "expressions" within tools) and we can make it easier to search for tools or easier to write and bundle new tools with workflows - we can fill in these gaps and allow Galaxy workflows to more easily capture more of the typical analyses in the biomedical realm. This is a step in that direction, later this week I will [open a PR](https://github.com/galaxyproject/galaxy/compare/dev...jmchilton:expression_tools?expand=1) for "Expression Tools" as well - these are also a step in that direction.

Both of these will be slightly niche until Galaxy allows users to load "arbitrary" tools in safely - but that niche is perhaps not as small as it seems. Projects like Galaksio and Refinery control the workflows they offer so presumably restricting such workflows to admin installation is not a problem at all. Likewise, personal Galaxy servers like the typical use case for docker-galaxy-stable should not have any problem finding maximal utility for such workflows. Obviously CWL is another use case - even CWL support in Galaxy when it finally lands will be a bit niche until we allow users to arbitrarily load tools.

These workflows and tools could also be useful in development - I'm trying to push TDD development of YAML workflows in Galaxy (https://planemo.readthedocs.io/en/latest/test_format.html) and developers could build out workflows with tool stubs embedded in workflows and use them that way locally. When it comes time to publish the workflow - they'd have very clearly boundaries for the tool they need to write and a good start.

lib/galaxy/webapps/galaxy/api/dynamic_tools.py
```diff
@@ -0,0 +1,67 @@
...
+
+        return dynamic_tool.to_dict()
+
+    @web.require_admin
+    @expose_api
+    def create(self, trans, payload, **kwd):
+        """
+        POST /api/dynamic_tools
+
+        If ``tool_id`` appears in the payload this executes tool using
```
mvdbeek:
```suggestion
        If ``tool_id`` appears in the payload this executes the tool using
```

lib/galaxy/webapps/galaxy/api/dynamic_tools.py
```diff
@@ -0,0 +1,67 @@
...
+        specified inputs and returns tool's outputs. Otherwise, the payload
```
mvdbeek:
```suggestion
        specified inputs and returns the tool's outputs. Otherwise, the payload
```

But wouldn't it be cleaner to POST directly to the regular tools endpoint to run the tool? This seems to be at odds with the class docstring.

mvdbeek (commented):
This looks like a nice first step, but maybe we should control loading of dynamic tools with a config setting ?
I think admins really need to be aware that they are possibly importing broken or malicious tools that users can run if the admin imports a workflow containing a dynamic tool.

What do you think about appending a list of loaded dynamic tools to a tool conf file ? 
This would have the advantage of neatly listing all dynamic tools and could then be used to disable a dynamic tool if it's not needed anymore.

jmchilton:
> This looks like a nice first step, but maybe we should control loading of dynamic tools with a config setting ?

I assumed because these were using YamlToolSource this would be hidden behind ``enable_beta_tool_formats`` but it appears not to be, would fixing that and updating the docs be sufficient or do you want a second config flag?

jmchilton:
> What do you think about appending a list of loaded dynamic tools to a tool conf file ?
This would have the advantage of neatly listing all dynamic tools and could then be used to disable a dynamic tool if it's not needed anymore.

I'm against this... I don't even know how to reference these tools in that file and we don't do this with other model objects. These tools are loaded as they are referenced so there is no need to cull them - they can just sit in the database. We don't remove workflows no longer being used from the database. I could imagine there being more CRUD operations on these entries.

I added an ``active`` flag to the model/db but don't actually use it - one could imagine an API endpoint for listing the tools, deactivating them, maybe an admin interface? Feels unneeded at this point, but I can add the API endpoints with some API tests at the very least if you'd like me to. At least then we could have a plausible story for how to deactivate rogue tools installed this way.

lib/galaxy/webapps/galaxy/api/dynamic_tools.py
jmchilton: Yeah - the tool create endpoint used to be overloaded to do this - I undid that but haven't fixed the docstring. Let me do that and rebase...

mvdbeek:
> At least then we could have a plausible story for how to deactivate rogue tools installed this way.

Yep, that's my main concern. If we just have a small admin only API endpoint to do that I'm 👍! 
Although I can imagine that being a little tricky if you have a bunch of badly named tools ...

mvdbeek (approved):

mvdbeek (approved):

jmchilton:
How exciting - here is a version of this commit with the proper 2015 date (https://github.com/common-workflow-language/galaxy/commit/42252ddd46f843bbde2d73c53f96720320a61bc5) 😄. Fixes common-workflow-language/galaxy#28 opened in Feb of 2016.

PR #7547, opened 2019-03-18
Extend reload test case logging

### LINKED ISSUE #7444: Prevent concurrent access to some ``ToolCache`` class attributes
Both `cleanup()` and `cache_tool()` methods of `ToolCache` access the following class attributes when a tool is reloaded:

- `_hash_by_tool_paths`
- `_mod_time_by_path`
- `_new_tool_ids`
- `_tools_by_path`
- `_tool_paths_by_id`

This causes the failure of unit test `ToolBoxTestCase.test_tool_reload_for_broken_tool`, see e.g.: https://travis-ci.org/galaxyproject/galaxy/jobs/476364519

mvdbeek:
Hopefully fixes https://github.com/galaxyproject/galaxy/issues/7444 ... again.

### DIFF
--- a/lib/galaxy/webapps/galaxy/config_watchers.py
+++ b/lib/galaxy/webapps/galaxy/config_watchers.py
@@ -20,7 +20,7 @@ from galaxy.util.watcher import (
 class ConfigWatchers(object):
     """Contains ToolConfWatcher, ToolWatcher and ToolDataWatcher objects."""
 
-    def __init__(self, app):
+    def __init__(self, app, start_thread=True):
         self.app = app
         # ToolConfWatcher objects will watch the tool_cache if the tool_cache is passed into get_tool_conf_watcher.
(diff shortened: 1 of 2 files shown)


mvdbeek:
OK, my current hypothesis is that setting the reload callback (`app.watchers.tool_config_watcher.reload_callback = lambda: reload_callback(test_case)`) may or may not be propagated to the running thread depending on how long it takes to start the thread ??
So now we only start the thread after setting the reload callback. The first tests have passed, so I guess this is worth a shot.

nsoranzo:
What about instatiating the toolbox as part of `BaseToolBoxTestCase.setUp()` instead? Just an idea.

mvdbeek:
There are a couple of other tests that use `BaseToolBoxTestCase`, I guess it'd be better to just populate the `self._toolbox` with this `SimplifiedToolBox` variant here, but no strong opinion here either.

Issue #5601, opened 2018-02-23
server_name templating and mangling issues


natefoo:
With the new uWSGI features it was necessary to do more intensive post-fork `server_name` mangling. In addition, with multiple ways to start the server (Paste w/ PasteDeploy, uWSGI w/ PasteDeploy, plain uWSGI, and webless `scripts/galaxy-main`), there are multiple different ways to set the base value of `server_name` pre-fork (which is needed for things like job handlers).

Complicating this is the fact that the default `server_name` under PasteDeploy is `main`, whereas under uWSGI with the YAML config files, it's the app name, which is the key to the config dict, i.e. `galaxy`.

My intent was to make it possible to configure a base `server_name` in `galaxy.{ini,yml}`, for that to be templated with `galaxy.util.facts`, and for it to be overrideable e.g. with `paster.py --server-name=` or `scripts/galaxy-main --server-name=` etc. It appears these have not all worked as intended and so they need to be revisited.

I discovered this in the course of figuring out how some of the uWSGI workers on my front end VM 1 were receiving `rebuild_toolbox_search_index` async tasks via the message queue before they had themselves reloaded the toolbox and queued the task; it's because the workers on VM 2 had already done it, 2's workers had queued the tasks, and because they *both* have workers with `server_name`s of `main.web.{0..7}`, 1 was picking up those tasks that 2 had intended for itself. So it should be documented somewhere that `server_name` will need to include a hostname-unique element if you use the same internal message queue with multiple hosts. Alternatively, we can automatically include some kind of host identifier when those exchanges are set up. Thoughts @dannon?

Milestoned to 18.01 because we need to get a fix in 18.01, but it's a bug, so that's fine - this should *not* block release.

martenson:
amongst the things that this causes is unreliable search response since some workers may have outdated search index

martenson:
pushed to 18.09

martenson:
Pushing to 19.05

natefoo:
#7712 will fix the specific problem I described with overlapping `server_names` on multiple VMs, but I'm not sure exactly what I was referring to when I said:

> It appears these have not all worked as intended and so they need to be revisited.

I think there's still a bug with one of the multitude of ways you can specify `server_name`, but I don't recall the specifics, so until it pops up again, #7712 can close this issue. I'll probably try to find it again whenever I get to #7289 .

mvdbeek:
I've cleared the milestone in case you'll want to revisit the actual server name config.

PR #7548, opened 2019-03-18
More robust task messaging

### LINKED ISSUE #7564: Data managers/tables - uWSGI
I have data managers and they (obviously) fill a few data tables.
Now I have my uWSGI set up with multiple processes.
What seems to happen is that some of these processes have outdated versions of the data table and therefore users are experiencing in some cases up to date and in some cases out of date data.

mvdbeek:
This implements a database heartbeat and allows the QueueWorker to receive and send tasks to local and remote processes as long as they are connected to the database. This should immediately improve things like the data table reload that is triggered after running a data manager tool and therefore fixes https://github.com/galaxyproject/galaxy/issues/7564 and https://github.com/galaxyproject/galaxy/issues/7568

I'm also implementing a way to wait and get the results of a task, which I will use for waiting for these reloads (in a new PR). Longer term we could use this to manage tool dependencies for handlers that run remotely and scheduling the config watching to a single process (instead of every process doing that separately). 
Everything implemented here has unit tests for amqp connections via sqlite and postgresql as well as a real message broker.

### DIFF
--- a/lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
+++ b/lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
@@ -34,7 +34,7 @@ pyparsing==2.3.1
 pytest-html==1.20.0
 pytest-metadata==1.8.0
 pytest-pythonpath==0.7.3
-pytest-pgsql==1.1.1 ; python_version > '3.3'
+pytest-postgresql==1.3.4
 pytest==4.3.1
 pytz==2018.9
 pyyaml==5.1
--- a/lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
+++ b/lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
@@ -34,6 +34,7 @@ pyparsing==2.3.1
 pytest-html==1.20.0
 pytest-metadata==1.8.0
 pytest-pythonpath==0.7.3
+pytest-pgsql==1.1.1 ; python_version > '3.3'
 pytest==4.3.1
 pytz==2018.9
 pyyaml==5.1
--- a/lib/galaxy/model/migrate/versions/0151_add_worker_process.py
+++ b/lib/galaxy/model/migrate/versions/0151_add_worker_process.py
@@ -1,0 +1,53 @@
+"""
+Add table for worker processes
+"""
+from __future__ import print_function
+
+import logging
+
+from sqlalchemy import (
+    Column,
+    DateTime,
+    MetaData,
+    Table,
+    TEXT,
+)
+
+from galaxy.model.orm.now import now
+
+log = logging.getLogger(__name__)
+metadata = MetaData()
+
+
+WorkerProcess_table = Table(
+    'worker_process',
+    metadata,
+    Column('server_name', TEXT, primary_key=True),
+    Column("update_time", DateTime, default=now, onupdate=now),
+)
+
+
+def upgrade(migrate_engine):
+    metadata.bind = migrate_engine
+    print(__doc__)
+    metadata.reflect()
+    _create(WorkerProcess_table)
+
+
+def downgrade(migrate_engine):
+    metadata.bind = migrate_engine
+    _drop(WorkerProcess_table)
+
+
+def _create(table):
+    try:
+        table.create()
+    except Exception:
+        log.exception("Creating %s table failed.", table.name)
+
+
+def _drop(table):
+    try:
+        table.drop()
+    except Exception:
+        log.exception("Dropping %s table failed.", table.name)
--- a/lib/galaxy/model/migrate/versions/0151_add_worker_process.py
+++ b/lib/galaxy/model/migrate/versions/0151_add_worker_process.py
@@ -13,6 +13,7 @@ from sqlalchemy import (
     TEXT,
 )
 
+from galaxy.model.migrate.versions.util import create_table, drop_table
 from galaxy.model.orm.now import now
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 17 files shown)


mvdbeek:
I remember now what the problem is, no fanout / broadcast for the sqlachemy (aka in-memory) transport, so queues / exchanges need to be declared up front, which makes this not very dynamic when spinning up new handlers. For uwsgi that's not too bad because we call `UWSGIApplicationStack.worker()`, which should give us all the uwsgi workers, but classic webless handlers need to be registered manually or can't be reached via control tasks.

Now I guess workers could also report themselves with an alive message in the database, which would close that gap. I'll see if that could be done.

mvdbeek:
We did actually use postgres for the integration tests, it seems ...

lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
```diff
@@ -34,6 +34,7 @@ pyparsing==2.3.1
 pytest-html==1.20.0
 pytest-metadata==1.8.0
 pytest-pythonpath==0.7.3
+pytest-postgresql==1.3.4
```
nsoranzo, line 37: You need to add this to `lib/galaxy/dependencies/pipfiles/default/Pipfile`

mvdbeek: Thanks @nsoranzo, I've rebased that into https://github.com/galaxyproject/galaxy/pull/7548/commits/e7e8283560ef061450a96c96d7cb2f61a6d41ba4

mvdbeek:
It's all green and ready to review again!

lib/galaxy/model/migrate/versions/0151_add_worker_process.py
```diff
@@ -0,0 +1,53 @@
...
+    metadata.reflect()
+    _create(WorkerProcess_table)
+
+
+def downgrade(migrate_engine):
+    metadata.bind = migrate_engine
+    _drop(WorkerProcess_table)
+
+
+def _create(table):
```
nsoranzo: FYI, I'm refactoring these methods in https://github.com/galaxyproject/galaxy/pull/7621 to avoid the code duplications.

dannon:
@mvdbeek Ran into a snag testing the administrative job locking under uwsgi, digging deeper to see if it's a wonky configuration on my end or something else.

mvdbeek:
Just tried this too, it locks the jobs but the UI doesn't show the lock, so you can't unlock, right ?
fwiw I tested this with the tool whitelist, which was working fine. I can look at that and also use the new methods from #7621

mvdbeek:
I think I fixed it with https://github.com/galaxyproject/galaxy/pull/7548/commits/4493d126d6fe1c99c32ab77f87b873e598680f25 -- moving that up in the init from `bind_and_start` was one of the last things I did ...

dannon:
@mvdbeek Cool, thanks, I'll check it out again.

mvdbeek:
@galaxybot test py3

mvdbeek:
@galaxybot test py3

dannon (approved):
@mvdbeek Thanks for this; the changes look great and it's working well in my testing so far -- let's get it into more hands! :)

natefoo:
@mvdbeek I think this does not address the issue in #5601 but correct me if I'm wrong?

> I discovered this in the course of figuring out how some of the uWSGI workers on my front end VM 1 were receiving `rebuild_toolbox_search_index` async tasks via the message queue before they had themselves reloaded the toolbox and queued the task; it's because the workers on VM 2 had already done it, 2's workers had queued the tasks, and because they _both_ have workers with `server_name`s of `main.web.{0..7}`, 1 was picking up those tasks that 2 had intended for itself. So it should be documented somewhere that `server_name` will need to include a hostname-unique element if you use the same internal message queue with multiple hosts. Alternatively, we can automatically include some kind of host identifier when those exchanges are set up.

(I'm not saying this PR should have fixed that issue, just wanted to make sure I should not close it).

Since servers now register themselves via heartbeat and my server names overlap there may be new issues? But I will finally fix things so I don't have overlapping server names.

mvdbeek:
Oh yeah, that's right. I could also add a uuid-postfix or hostname or really anything else to the task queue if that would make things easier ?

natefoo:
Hmm. Something host-specific (`hostid(1)`?) would work for my case, I'm trying to think of any other cases where we might need something unique to some other property and can't think of any.

Isn't this specific to the silly way we send messages meant for all processes though? In reality we should use a fanout exchange instead of generating server names, I'd think. So it'd probably make more sense if I just make my base server names unique until we implement that?

mvdbeek:
It is, but fanout exchanges don't work with in-memory transports like sqlalchemy. I think docker-based handlers will also benefit from including the hostname by default.

natefoo:
I can't think of a downside, so if you're volunteering to spend a few minutes to add it, be my guest, but it's certainly not a dire need. :smile:

mvdbeek:
I don't wanna bring main down (again) so I'll do it 😆

natefoo:
Thanks!

blankenberg:
I had an instance that I had to comment out `pytest-postgresql==1.3.4` to get to start up. The machine it is on doesn't have `pg_config` , and pytest-postgresql was causing a new version of `psycopg2-binary==2.8.1` be attempted to be built (which requires `pg_config`). 

The venv already has a working version of `psycopg2-binary==2.7.4` that is being used (and is required for Galaxy as an optional dependency).

mvdbeek:
We can update to 1.4 which removed the hard requirement on psycopg2, but why is your instance pulling in dev-requirements ? Are you setting DEV_WHEELS ?

blankenberg:
Not sure exactly, it is started with
`DRMAA_LIBRARY_PATH=/usr/local/lib/libdrmaa.so sh run.sh --daemon`

mvdbeek:
k, I think https://github.com/galaxyproject/galaxy/pull/7703 should fix that additional dependency, if you wanna give it a try.

nsoranzo:
@blankenberg Then dev wheels should get installed only if you are on the `dev` branch.

blankenberg:
Yes, definitely on a dev-based branch there (realtimetools).

blankenberg:
Will try the change shortly, once I can restart the server again.

PR #7556, opened 2019-03-18
Implement expression tools and non-data tool outputs.

jmchilton:
PR #6925 introduced a GUI for connecting non-data (e.g. integer, boolean, color, etc..) workflow input parameter to tool input parameters (the backend for this was originally added in #1306). That ideally was just the beginning of work toward using such values in structured ways in workflows.

This PR extends tool output handling to allow producing of non-data parameters. These can serve as a source for non-data values in workflows the same work workflow input parameters can. To make such values more easy to produce, this PR also introduces Galaxy expression tools - mirroring functionality regularly used in CWL. These small JavaScript-based tools that consume inputs just like a regular Galaxy tool but that produce dictionary of non-data values.

I think these expressions will be maximally useful when paired with format 2 workflows once we allow users to load arbitrary tools (I make the case more in full [here](https://github.com/galaxyproject/galaxy/pull/7545#issuecomment-473894424)), but I outline important near term potential uses there as well.

Because there is always a checklist in my PR descriptions:

- Tool definition language and plumbing and datatype for expressing expressions as jobs.
- Allow connecting expression tools to parameters in workflows, will delay evaluation of workflow so calculated value
- Example test expression tools for testing and demonstration.


lib/galaxy/tools/__init__.py
```diff
@@ -102,6 +104,12 @@
 
 log = logging.getLogger(__name__)
 
+REQUIRES_JS_RUNTIME_MESSAGE = ("The tool [%s] requires a nodejs runtime to execute "
+                               "but node nor nodejs could be found on Galaxy's PATH and "
+                               "no runtime was configured using the nodejs_path option in "
+                               "galaxy.ini.")
```
mvdbeek: This is a user-facing message, right ? I'd keep it a bit simpler
```
REQUIRES_JS_RUNTIME_MESSAGE = ("The tool [%s] requires a nodejs runtime to execute "
                               "but node or nodejs could not be found. Please contact the Galaxy adminstrator")
```

lib/galaxy_ext/expressions/handle_job.py
```diff
@@ -0,0 +1,45 @@
...
+
+from galaxy.tools.expressions import evaluate
+
+logging.basicConfig()
+log = logging.getLogger(__name__)
+
+
+def run(environment_path=None):
+    if expression is None:
+        raise Exception("Python library cwltool must available to evaluate expressions.")
```
mvdbeek:
```suggestion
        raise Exception("Python library cwltool must be available to evaluate expressions.")
```

lib/galaxy/tools/parameters/wrapped_json.py
```diff
@@ -62,8 +62,14 @@ def _data_input_to_path(v):
...
-        else:
-            raise NotImplementedError()
+        elif handle_files == "OBJECT":
+            if value:
+                if isinstance(value, list):
+                    value = value[0]
+                return _hda_to_object(value)
+            else:
+                return None
+        raise NotImplementedError()
```
mvdbeek:
```suggestion
        else:
            raise NotImplementedError()
```
This broke the inputs_as_json_with_paths tool, I think

mvdbeek (commented):
Absolutely awesome, can't wait to use this!

lib/galaxy/tools/__init__.py
jmchilton: And ``galaxy.ini`` can you tell I've had some of these commits around since 2015 😄. I'll make this change - thanks!

jmchilton:
And I force pushed over the Github changes :). Sorry, going to do that without credit.

mvdbeek:
This is going to be so awesome:
expression_tool

We should probably ship with a tool that just takes a text file and parses that into integer / float/ text / bool/ color, this would be useful for workflows that have to compute parameters, and I think it's not too difficult for users.

mvdbeek:
I opened [a PR](https://github.com/jmchilton/galaxy/pull/71) against your branch with some fixes I found while testing this

jmchilton:
It is green 😄 - and a million thanks for the workflow test case fix - no wonder it was sort of consistently failing during my history import/export testing but in such an odd way.

mvdbeek (approved):
Fantastic, 19.05 is going to be a great release!

lib/galaxy/tools/xsd/galaxy.xsd
```diff
@@ -2781,6 +2782,33 @@ Prior to Galaxy release 19.01 the stdio block has only been used for non-legacy
...
+      </xs:extension>
+    </xs:simpleContent>
+  </xs:complexType>
+
+  <xs:simpleType name="ExpressionType">
+    <xs:annotation>
+      <xs:documentation xml:lang="en"></xs:documentation>
+    </xs:annotation>
+    <xs:restriction base="xs:string">
+      <xs:enumeration value="emca5.1" />
```
martenson: should this be `ecma5.1` ?

scripts/common_startup.sh
```diff
@@ -224,14 +224,23 @@ if [ $SKIP_CLIENT_BUILD -eq 0 ]; then
     fi
 fi
 
+# Install node if not installed
+if [ -n "$VIRTUAL_ENV" ]; then
+    if ! in_venv "$(command -v node)" || [ "$(node --version)" != "v$NODE_VERSION" ]; then
+        echo "Installing node into $VIRTUAL_ENV with nodeenv."
+        nodeenv -n $NODE_VERSION -p
+    fi
+elif [ -n "$CONDA_DEFAULT_ENV" -a -n "$CONDA_EXE" ]; then
```
martenson, line 233: This change seems unrelated to the PR and I am a bit reluctant, since the things we do with startup and node has shown to be troublesome for various different-than-ours environments. Maybe we could extract it to a separate PR for more eyes to see? I might be overstating the risk though.

martenson (approved):
very exciting, I've noted two minor points

lib/galaxy/tools/xsd/galaxy.xsd
jmchilton: Yes, thanks I'll fix this and the tools.

scripts/common_startup.sh
jmchilton: Is this in regards to the conda piece specifically or moving the whole block including the nodeenv thing outside the build client block below?

martenson: The conda part mostly.

Trying to install node every time with no way for the admin/user to opt out (unless you stop using common_startup/virtualenv) could also be tricky based on the number of reports/feedback we got with client builds lately but that is our new requirement after this PR and I am fine with that and it is early enough in the dev cycle to introduce it

lib/galaxy/tools/xsd/galaxy.xsd
jmchilton: Fixed and the validation/default handling has been updated to reflect the XSD docs.

scripts/common_startup.sh
jmchilton: It is @mvdbeek's commit but I like it as it is and 👍 it. I feel like this would be the correct way to handle installing node in a Conda environment and that is pretty experimental so I'm not too worried about breaking it - seems more likely to fix something than break it. Unrelated sure, but who would comment on Conda dependency installation - probably only @mvdbeek anyway 😅.

As for this not being hidden behind a flag, I suspect at some point someone will ask for a flag (given https://github.com/galaxyproject/galaxy/commit/6e865a3363587582165788ea7b604cda084dff4f it will probably be me) but I'd wait until someone asks to add that logic because we could be wrong and hopefully are.

martenson: I see, I did not notice that `CONDA_EXE` is only when we use conda to install deps.  resolved!

mvdbeek: Yeah, just wanted to make sure that by default we'd get node installed. Not having node in the unit tests was causing the circle-ci builds to fail. If you guys want to take this one out and instead raise a skip in the unit test if there's no node that's fine with me as well.
We still usually get node through the client build, so for use outside of unit tests this doesn't matter too much.

PR #7558, opened 2019-03-19
Improve linking between job conf handler assignment and documentation

hexylena:
My initial reaction was to attempt to search the docs site for `mem-self` and other of the values, expecting that they would be used in the documentation. I did not immediately make the connection whilst skimming the linked documentation from the conf, so I've additionally linked to the specific subsection of interest


natefoo:
Thanks @erasche!

PR #7561, opened 2019-03-19
[19.01] Correct db-self handler assignment method

hexylena:
When testing this, I saw some crashes due to no such attribute, it looks like an underscore was dropped here.


jmchilton (approved):

nsoranzo:
:+1: Introduced in commit 355b48c361e68a7d9fbe9fe86de7aa2313e5d608 .

natefoo:
@erasche thanks!

PR #7565, opened 2019-03-19
move the 'create new history' out of the history context menu

martenson:
cc @nsoranzo 
Galaxy

xref https://github.com/galaxyproject/galaxy/pull/6437

### DIFF
--- a/client/galaxy/scripts/entry/panels/history-panel.js
+++ b/client/galaxy/scripts/entry/panels/history-panel.js
@@ -36,6 +36,15 @@ var HistoryPanel = Backbone.View.extend({
                 self.historyView.loadCurrentHistory();
             }
         });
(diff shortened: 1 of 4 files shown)


hexylena:
Like it! That's a commonly used button.

hexylena (approved):

martenson:
@erasche what do you think about moving it one position to the right, to avoid context-changing mis-clicking when you are trying to refresh? (on the other hand refresh should be a very rare action, history should be updating automagically)

hexylena:
Either way you're at risk of a context-changing misclick. There you can create a new one when you wanted to refresh, moving right you could open (maybe slow) multi history view when you just wanted a new history. I'm not sure there is a better position, and with limited horizontal space you can't add more whitespace to help discourage misclicks easily. :/

PR #7569, opened 2019-03-21
Add mtx datatype

bebatut:
Hi, 

This datatype is used in single-cell (output of STAR-solo, input for scanpy/anndata, etc), but not only
Doc: https://math.nist.gov/MatrixMarket/formats.html

Bérénice


lib/galaxy/datatypes/tabular.py
```diff
@@ -1161,3 +1161,93 @@ def get_chunk(self, trans, dataset, chunk):
...
+        lines = handle.readlines()
+
+        if not lines[0].startswith('%%MatrixMarket matrix coordinate real general'):
+            return False
+
+        for line in lines:
+            if not line.startswith('%'):
+                split_line = line.split('\t')
+                if len(split_line) != 3:
+                    print("not 3 columns")
```
bgruening: this should be removed

lib/galaxy/datatypes/tabular.py
```diff
@@ -1161,3 +1161,93 @@ def get_chunk(self, trans, dataset, chunk):
...
+    file_ext = "mtx"
+
+    def __init__(self, **kwd):
+        super(MatrixMarket, self).__init__(**kwd)
+
+    def sniff_prefix(self, file_prefix):
+        count = 0
+        nonzeros = 0
+        handle = file_prefix.string_io()
+        lines = handle.readlines()
```
bgruening: can we use `with open() as ` here?

lib/galaxy/datatypes/test/1.mtx
```diff
@@ -0,0 +1,31 @@
...
+%=================================================================================
+5	5	8
+1	1	1.000e+00
+2	2	1.050e+01
+3	3	1.500e-02
+1	4	6.000e+00
+4	2	2.505e+02
+4	4	-2.800e+02
+4	5	3.332e+01
+5	5	1.200e+01
```
mvdbeek: The hash comparison in the integration test is failing here because there's no newline at the end of the file, and we add one during the import.

lib/galaxy/datatypes/tabular.py
bebatut: I removed it. 
Thanks for spotting it (remainder of local tests 😄)

lib/galaxy/datatypes/tabular.py
bebatut: Updated

lib/galaxy/datatypes/test/1.mtx
bebatut: I added a newline at the end of the file. Thanks

bebatut:
I have one issue. When the input file is delimited by space and not tab, it is now correctly sniffed as mtx. The metadata and view in the history look ok:

Screenshot 2019-03-21 at 15 25 54

But in the view, the data are not split into the columns (as in the quick view)

Screenshot 2019-03-21 at 15 20 10

Any idea how I can fix that?

mvdbeek:
~I think you'd have to implement (or inherit) a custom `get_chunk` method. It looks like the ConnectivityTable datatype is doing the same in https://github.com/galaxyproject/galaxy/blob/24f97f6917da606f8b9c1fc1fb4faef5c54f1933/lib/galaxy/datatypes/tabular.py#L1142~

dannons's idea is better, and then we can do away with the custom code in ConnectivityTable as well.

dannon:
@bebatut Right now the 'fancy' data display there just splits on `\t` by default; the easiest way to address this is probably add a datatype-specific tabular separator set.  So, normal tabular files would remain `\t`, but you could use `['\t',\ ']` for this datatype to extend the functionality.

The splitting happens here:  https://github.com/galaxyproject/galaxy/blob/dev/client/galaxy/scripts/mvc/dataset/data.js#L242

So you'd just need to find a way to get the extra separators (in your case, a space character) to this display method.

bebatut:
@mvdbeek @dannon thanks for the information
I also figure out that the cutting on 1st column (using `cut` tool) returns the whole lines (not only the 1st column)

lib/galaxy/datatypes/tabular.py
```diff
@@ -1161,3 +1161,92 @@ def get_chunk(self, trans, dataset, chunk):
...
+    file_ext = "mtx"
+
+    def __init__(self, **kwd):
+        super(MatrixMarket, self).__init__(**kwd)
+
+    def sniff(self, filename):
+        count = 0
+        nonzeros = 0
+        with open(filename, 'r') as handle:
+            lines = handle.readlines()
```
jmchilton: This does unbounded reads - this can crash the Galaxy server. Isn't that header line enough to say this is that kind of file?

jmchilton: Using the newer sniff_prefix handling (just from ``@build_sniff_from_prefix`` on the class definition), I feel like the following would be a fine sniffer:

```python
def sniff_prefix(self, file_prefix):
    return file_prefix.startswith('%%MatrixMarket matrix coordinate real general')
```

There are advantages to sniff_prefix - it avoids unneeded I/O and keeps memory consumption bounded.

bebatut: I made the change as you suggested @jmchilton 
Thanks 😄

bebatut:
@dannon I tried (in probably an uggly way) as you suggested. It works!
I also added support for csv support as it was also not rendered.
Let me know if/how I can improve the code (as it is currently stupid copy/paste)

bgruening (approved):
Thanks @bebatut!

@dannon is that ok for you as well?

dannon:
@bgruening @bebatut Sorry I missed the ping before when you made the changes!  I'll take a quick look now.

Yeah, this looks like it'll work great.  We can do more specific binding of separators to datatypes later to be a little more efficient, but this is nice and flexible.

PR #7574, opened 2019-03-21
trim down travis config, only run osx first startup

martenson:
we use circleci in stead

https://travis-ci.org/martenson/galaxy/builds/509581696

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,82 +1,19 @@
 language: python
 python: 2.7
-os:
-  - linux
-env:
-  - TOX_ENV=py27-lint
-  - TOX_ENV=py27-unit
-  - TOX_ENV=py27-first_startup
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,82 +1,19 @@
...
-
-before_install:
-  # Workaround for https://github.com/travis-ci/travis-ci/issues/7940
-  - sudo rm -f /etc/boto.cfg
-
 install:
   - set -e
   - pip install tox
   - |
-    if [ "$TOX_ENV" == "py27-first_startup" ]; then
```
nsoranzo: I'd keep this `if`, since we may want to add more macOS test envs (e.g. `py35-first_startup`).

martenson: done

nsoranzo (approved):

PR #7577, opened 2019-03-21
[19.01] Backport workflow output rectification

mvdbeek:
@dannon I don't need bfc0e65950978901b743070c3b9be1e58181c53f for 19.01, right ?
I didn't see any issue in testing, but I'm not sure how to know if `Galaxy` is globally available.


dannon:
@mvdbeek I think it'll actually *work* without it, but it's more correct to include it.  19.01 does have 'getGalaxyInstance' for accessing the singleton Galaxy.  19.01 doesn't have eslint, though, so it's not going to complain about it in the testing.

mvdbeek:
Thanks, then I'll include it.

PR #7585, opened 2019-03-22
Format authnz-related api responses as JSON objects

VJalili:


lib/galaxy/webapps/galaxy/api/cloud.py
```diff
@@ -233,6 +234,6 @@ def send(self, trans, payload, **kwargs):
                                                authz_id=authz_id,
                                                dataset_ids=dataset_ids,
                                                overwrite_existing=payload.get("overwrite_existing", False))
-        return {'sent_dataset_labels': sent,
-                'failed_dataset_labels': failed,
-                'bucket_name': bucket}
+        return json.dumps({'sent_dataset_labels': sent,
```
dannon: The expose_api used here (and in the above method) should automatically be converting to json -- did you find it was not?

lib/galaxy/webapps/galaxy/controllers/authnz.py
```diff
@@ -35,7 +35,7 @@ def index(self, trans, **kwargs):
         rtv = []
         for authnz in trans.user.social_auth:
             rtv.append({'id': trans.app.security.encode_id(authnz.id), 'provider': authnz.provider})
-        return rtv
+        return json.dumps(rtv)
```
dannon: Instead of this approach, can you try using the @web.json decorator?

lib/galaxy/webapps/galaxy/api/cloud.py
VJalili: That is what I guessed when implementing this in the first place, however, testing it with @Nerdinacan it turned out its returned value is not exactly a JSON object.

dannon: Interesting.  Can you debug what's going wrong with the expose_api decorator?

lib/galaxy/webapps/galaxy/controllers/authnz.py
VJalili: Thanks, I agree that is cleaner.

lib/galaxy/webapps/galaxy/api/cloud.py
VJalili: The API returns an object like the following: 

```
{
    "sent_dataset_labels": [],
    "bucket_name": "vahid-objectstore-tests",
    "failed_dataset_labels": []
}
```

This looks a valid JSON to me, so, I guess maybe there was unrelated issue on the branch we were looking at.

dannon: Ok, cool.  Let me know if you need help tracking it down, but I was going to be surprised/worried if the decorators were busted :)

VJalili: Sure, thanks. As things stand, we're good for now :)

lib/galaxy/webapps/galaxy/api/cloud.py
```diff
@@ -132,7 +133,7 @@ def get(self, trans, payload, **kwargs):
         rtv = []
         for dataset in datasets:
             rtv.append(self.datasets_serializer.serialize_to_view(dataset, view='summary'))
-        return rtv
+        return json.dumps(rtv)
```
dannon, line 136: @VJalili Can you revert and/or rebase over this change as well, since @web.expose_api should be handling this? (if it isn't, we should look into that)

VJalili: Sure, I created a different PR: https://github.com/galaxyproject/galaxy/pull/7592

VJalili:
a cleaner branch PRed: https://github.com/galaxyproject/galaxy/pull/7592

dannon:
@VJalili Thanks!

PR #7587, opened 2019-03-23
Add ImgtJson datatype for IMGT immune system libraries

jj-umn:
Immunogenetics libraries are used as a reference for aligning immunome
sequence data.  The mixcr application uses these as references.
https://milaboratory.com/software/mixcr/
IMGT libraries are available at:
https://github.com/repseqio/library-imgt/releases
The ImgtJson datatype stores the available taxonIDs in the IMGT library
so that a taxonID can be selected from a galaxy tool from a select
parameter.


jmchilton:
Thanks, looks good!

martenson:
thanks @jj-umn !

PR #7588, opened 2019-03-23
Refactor galaxy.dataset_collections into galaxy.model.dataset_collections.

jmchilton:
Originally what is now galaxy.managers.collections was in galaxy.dataset_collections with all the app and trans centralized there. Pre-managers I think that made sense - app and trans logic doesn't belong in galaxy.model. But now that the former collection "service" has been refactored out into a manager, everything that is left is very abstract logic including model manipulation logic and I think is a perfect fit for galaxy.model.

Put another way, I think what is left in there would belong in a galaxy-models package and so might as well be relocated into galaxy.model to formalize those dependencies, etc..


dannon:
Yep, makes sense to have this live in model to me.

PR #7589, opened 2019-03-25
Add ``py35-first_startup`` macOS job to Travis

nsoranzo:
Use conda to install Python 3.5 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,10 @@
-language: python
-python: 2.7
+os: osx
+# No version of Python is available via virtualenv on OS X workers, see https://github.com/travis-ci/travis-ci/issues/2312
+language: generic
 
-matrix:
-  include:
-    - os: osx
-      # No version of Python is available via virtualenv on OS X workers, see https://github.com/travis-ci/travis-ci/issues/2312
-      language: generic
-      env: TOX_ENV=py27-first_startup
+env:
+  - TOX_ENV=py27-first_startup
+  - TOX_ENV=py35-first_startup
 
 install:
   - set -e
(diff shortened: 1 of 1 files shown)


mvdbeek:
How about running the unit tests as well ? I think that'd be quite valuable.

nsoranzo:
> How about running the unit tests as well ? I think that'd be quite valuable.

py27-unit ?

mvdbeek:
I'd probably opt for both unless we think it's too much load for our travis resources ?

nsoranzo:
macOS builds used to be an issue, but now the situation seems much better. Maybe we can start with the 2 startup and the `py27-unit` jobs, and add `py35-unit` if it works well.

mvdbeek:
Sounds good to me.

mvdbeek:
Also if it does become a resource issue we can also bundle the unit tests with the startup test, I assume what is slow is booting the machines.

mvdbeek:
Hmm, I see this is still 2/3 pending for more than 30 minutes. Should we maybe try to run everything as a single job ?

nsoranzo:
@mvdbeek Unit tests failed twice for both reload tests :cry:

mvdbeek:
Hmm, the callback hasn't been triggered, so I assume the thread failed to start in time ?

nsoranzo:
> Hmm, the callback hasn't been triggered, so I assume the thread failed to start in time ?

Maybe, not sure to be honest. I don't have a Mac, so I cannot debug locally. Should we leave `py27-unit` for a separate PR, or do you want to try to fix it as part of this PR?

martenson:
I would stick to the startup tests for now - those are the most important for macos I think.

mvdbeek:
Different issue, let's not start fixing things in this PR. I do think the unit tests are as important as the startup tests, but we can talk about that when someone has figured out what is going wrong there.

martenson (approved):

PR #7601, opened 2019-03-26
Add extra verbosity in client build startup logic

dannon:
Simply adds a few extra echoes in common startup to explain why a client build is being skipped.  Ran into a situation trying to help figure out what was going on with @VJalili's local environment and this would have provided the info I needed.

### DIFF
--- a/scripts/common_startup.sh
+++ b/scripts/common_startup.sh
@@ -210,7 +210,7 @@ if [ $SKIP_CLIENT_BUILD -eq 0 ]; then
         # If git is not used and static/client_build_hash.txt is present, next
         # client rebuilds must be done manually by the admin
         if [ "$GIT_BRANCH" = "0" ]; then
-            echo "Skipping Galaxy client build because git is not in use and the client build state cannot be compared against local changes.  If you have made local modifications manual client builds will be required."
+            echo "Skipping Galaxy client build because git is not in use and the client build state cannot be compared against local changes.  If you have made local modifications, then manual client builds will be required."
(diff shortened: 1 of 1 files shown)


scripts/common_startup.sh
```diff
@@ -210,10 +210,12 @@ if [ $SKIP_CLIENT_BUILD -eq 0 ]; then
         # If git is not used and static/client_build_hash.txt is present, next
         # client rebuilds must be done manually by the admin
         if [ "$GIT_BRANCH" = "0" ]; then
+            echo "Skipping Galaxy client build because git is not in use and the client build state cannot be compared against local changes.  If you have made local modifications manual client builds will be required."
```
nsoranzo:
```suggestion
            echo "Skipping Galaxy client build because git is not in use and the client build state cannot be compared against local changes.  If you have made local modifications, manual client builds will be required."
```

scripts/common_startup.sh
```diff
@@ -222,6 +224,8 @@ if [ $SKIP_CLIENT_BUILD -eq 0 ]; then
     else
         echo "The Galaxy client has not yet been built and will be built now."
     fi
+else
+    echo "The Galaxy client build is being skipped due to the SKIP_CLIENT_BUILD env var."
```
nsoranzo:
```suggestion
    echo "The Galaxy client build is being skipped due to the SKIP_CLIENT_BUILD environment variable."
```

nsoranzo (approved):

PR #7602, opened 2019-03-26
Fix styleguide

dannon:
Didn't track down when this broke, but it has been a while.   To test this, you can run `yarn run styleguide` and then look at the local webserver.  

We should be making better use of the styleguide and using it in development.  I'm hoping to include vue components (like the recent badges/tags, etc) as well in the near future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -145,6 +145,7 @@ doc/source/api/api.rst
 doc/source/api/ts_api.rst
 doc/source/dev/schema.md
 doc/source/dev/schema.rst
+client/styleguide
 
 # webpack stats
 client/webpack-stats.json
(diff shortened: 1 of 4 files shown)


martenson:
damn, 2.5k lines yarn.lock change for a styleguideist update

dannon:
Yep.  It has lots of dependencies, and this was two major version bumps.

The good news is that it works again, pretty nicely.  I've started experimenting adding Vue components to it and it's going to be very helpful moving forward.

dannon:
Looks like TravisCI completed, it's just not updating here?  This is good to go.

dannon:
@martenson Thanks!

PR #7606, opened 2019-03-26
History panel header button organization, tweaks.

dannon:
It'd be nice for consistency to keep the 'cog' always at the far right (last entry) in our menu button lists.  This also only presents the 'create new history' (`+`) button when it's an option for the user (i.e. -- not anons; currently if anonymous and you click it it just pops up a big red alert).  This aligns handling of this with many other options we similarly mask if the user is anonymous.


martenson:
Thanks! This might interest you: https://github.com/galaxyproject/galaxy/pull/7607

dannon:
@martenson Yeah, that's cool, thanks for the pointer!  It'll be great to gather hard stats for better decision making about these things moving forward.

mvdbeek (approved):
Cool, works well, thanks dannon!

PR #7607, opened 2019-03-26
[usegalaxy] gather click stats in the history menu

martenson:
this aims to be deployed on Main only for a limited amount of time so we can get insights about how the menu is used and then streamline it


martenson:
It works locally in my testing so far. 

The 'purge deleted datasets' behaves differently though, before it would open the page with 'green' message in the middle pane, now it opens in top (regardless whatever window.* it is given) - help would be appreciated. Frankly, this should be a toast.

This uses a fairly heavy handed approach, two main reasons for it:

- it will only run on Main, and only for a limited time (a month?)
- we have about 4 means of handling menu items in this single menu, so writing this in a generic way would require changes in many places

PR #7612, opened 2019-03-27
Fix history options menu href navigation

### LINKED ISSUE #7608: 'purge deleted datasets' history menu item does nothing
this is on `dev`, can be reproduced on https://test.galaxyproject.org, works on release_19.01 (can be seen on Main)

xref: https://github.com/galaxyproject/galaxy/pull/7607

dannon:
~Leaving this wip for a sec while I try to understand what exactly changed here, why 'galay_main' is ending up as a string hanging off the parent object instead of being the frame target, but this simplifies the logic a little (and works, for me).~

I still don't quite get what happened here, but this works well in my testing, so I'm moving on.

fixes #7608


mvdbeek:
Thanks dannon, works well for me too.

martenson:
thanks @dannon !

PR #7613, opened 2019-03-27
Add when_empty attribute to tool outputs that can 'hide' or 'delete' …

blankenberg:
…the item if it is empty at job completion.

### DIFF
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -1716,6 +1716,28 @@ class Tool(Dictifiable):
     def exec_after_process(self, app, inp_data, out_data, param_dict, job=None):
         pass
 
+    def handle_outputs_when_empty(self, app, out_data, out_collections):
+        for key, value in self.outputs.items():
+            if value.when_empty:
+                data = out_data.get(key, None)
+                if data and data.get_total_size() == 0:
+                    if value.when_empty == 'hide':
+                        data.visible = False
+                        log.debug('Hiding empty output dataset: %s', data)
+                    if value.when_empty == 'delete':
+                        data.deleted = True
+                        log.debug('Deleting empty output dataset: %s', data)
+        for key, value in self.output_collections.items():
+            if value.when_empty:
+                collection = out_collections.get(key, None)
+                if collection and not collection.dataset_instances:
+                    if value.when_empty == 'hide':
+                        collection.visible = False
+                        log.debug('Hiding empty output collection: %s', collection)
+                    if value.when_empty == 'delete':
+                        collection.deleted = True
+                        log.debug('Deleting empty output collection: %s', collection)
+
     def job_failed(self, job_wrapper, message, exception=False):
         """
         Called when a job has failed
--- a/lib/galaxy/tools/xsd/galaxy.xsd
+++ b/lib/galaxy/tools/xsd/galaxy.xsd
@@ -3755,7 +3755,7 @@ pipes or periods (e.g. ``.``).]]></xs:documentation>
     <xs:attribute name="when_empty">
       <xs:annotation>
         <xs:documentation xml:lang="en"><![CDATA[Action to take when output
-is empty. Can currently be ``hide`` or ``delete`` elements.]]></xs:documentation>
+is empty. Can currently be ``hide`` or ``delete``.]]></xs:documentation>
       </xs:annotation>
       <xs:simpleType>
         <xs:restriction base="xs:string">
@@ -3764,7 +3764,6 @@ is empty. Can currently be ``hide`` or ``delete`` elements.]]></xs:documentation
         </xs:restriction>
       </xs:simpleType>
     </xs:attribute>
-
   </xs:attributeGroup>
 
   <xs:attributeGroup name="OutputDataAttributes">
--- a/lib/galaxy/tools/xsd/galaxy.xsd
+++ b/lib/galaxy/tools/xsd/galaxy.xsd
@@ -3752,6 +3752,19 @@ path in ``command`` and ``configfile`` elements. The name should not contain
 pipes or periods (e.g. ``.``).]]></xs:documentation>
       </xs:annotation>
     </xs:attribute>
+    <xs:attribute name="when_empty">
+      <xs:annotation>
+        <xs:documentation xml:lang="en"><![CDATA[Action to take when output
+is empty. Can currently be ``hide`` or ``delete`` elements.]]></xs:documentation>
+      </xs:annotation>
+      <xs:simpleType>
+        <xs:restriction base="xs:string">
+          <xs:enumeration value="hide"/>
+          <xs:enumeration value="delete"/>
+        </xs:restriction>
+      </xs:simpleType>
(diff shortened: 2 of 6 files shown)


mvdbeek:
I think this will be difficult to predict in the scheduling of workflows and queueing up of tool runs. Say you have 2 matched collections with elements A B C, and you run them against each other, if one of the inputs got deleted we may either stop evaluating the workflow here or switch from matched to multiplied mode in the tool form, both of which is probably going to be a surprising event to the user.

This is not so much a concern if just hiding instead of deleting the output, but it still feels like something a user should have full control over.

lib/galaxy/tools/xsd/galaxy.xsd
```diff
@@ -3752,6 +3752,18 @@ path in ``command`` and ``configfile`` elements. The name should not contain
 pipes or periods (e.g. ``.``).]]></xs:documentation>
       </xs:annotation>
     </xs:attribute>
+    <xs:attribute name="when_empty">
+      <xs:annotation>
+        <xs:documentation xml:lang="en"><![CDATA[Action to take when output
+is empty. Can currently be ``hide`` or ``delete``.]]></xs:documentation>
```
mvdbeek, line 3758: Would be more accurate to say `when the output file size is 0.`

blankenberg: You can also use the flag on collections to hide empty collections. *edit: or delete

blankenberg: > I think this will be difficult to predict in the scheduling of workflows and queueing up of tool runs. Say you have 2 matched collections with elements A B C, and you run them against each other, if one of the inputs got deleted we may either stop evaluating the workflow here or switch form matched to multiplied mode in the tool form, both of which is probably going to be a surprising event to the user.
> 
> This is not so much a concern if just hiding instead of deleting the output, but it still feels like something a user should have full control over.

I was just going to call it hide_when_empty as a boolean, but then maybe there are other things a tool designer wants to do when empty? First thought was delete. 

With a current tool I am working on, I am more concerned with allowing hiding (because of the reasons you listed, and others). But there could be a case where deleting is more desirable to a tool designer?

mvdbeek: > You can also use the flag on collections to hide empty collections.

Which is a different thing actually, this should probably be a separate attribute with a different name ? You can still have a collection with elements having no data when the attribute is set, which is different from how it works for dataset outputs.

> But there could be a case where deleting is more desirable to a tool designer?

I hope not. It'll sound appealing to tool authors, but it will be a big pain when you try to do anything complex with such a tool. And as a user you will have no indication that this is going to happen.

blankenberg: To expand a bit on the current use-case. There is a tool that creates between 1 and 110 (!, I know...) outputs (mix of datasets and collections; outputs are different file types as well), but which ones are going to be created cannot be readily determined before job run (so no prefiltering, unfortunately). 

For the datasets, they are using `from_work_dir`, for collections, it is using `discover_datasets`.

blankenberg: > > You can also use the flag on collections to hide empty collections.
> 
> Which is a different thing actually, this should probably be a separate attribute with a different name ? You can still have a collection with elements having no data when the attribute is set, which is different from how it works for normal datasets.
> 

Don't think I agree that they are different. There is an OutputObject, it can have an action performed when_empy. For datasets this is file size, for collection this is content. 

Presumably, you might want to also apply this flag to the contents (datasets) within the collection, but that would happen to the dataset OutputObjects, not the collection OutputObject.


> > But there could be a case where deleting is more desirable to a tool designer?
> 
> I hope not. It'll sound appealing to tool authors, but it will be a big pain when you try to do anything complex with such a tool. And as a user you will have no indication that this is going to happen.

Perhaps, but I am not entirely convinced that this shouldn't be possible if a tool designer wants to do it. I suppose that I am -0 on removing delete functionality, but still think the `when_empty` attribute setup makes sense for potential future enhancements.

nsoranzo: @blankenberg Maybe the easiest solution is to fix the tool to not create the empty datasets in the first place? Or is there a use case for the empty datasets?

blankenberg: I agree that not creating the datasets would be ideal, but unfortunately there is not a perfect way to determine the exact subset of files that will be created by the underlying binary for a particular set of inputs before job execution.

Another potential approach to handle this would be to use `discover_datasets` for all of the tool wrapper's outputs, but that would greatly decrease the utility of the tool, and, I would argue, be objectively worse than creating unused outputs (hidden or not). 

A third approach could be to fall back to a tool specific code file to perform the hiding, but I'd much rather avoid that. 


(Thanks to each of you for the reviews and discussions, it is appreciated)

lib/galaxy/tools/__init__.py
```diff
@@ -1716,6 +1716,28 @@ def exec_before_job(self, app, inp_data, out_data, param_dict={}):
...
+                if data and data.get_total_size() == 0:
+                    if value.when_empty == 'hide':
+                        data.visible = False
+                        log.debug('Hiding empty output dataset: %s', data)
+                    if value.when_empty == 'delete':
+                        data.deleted = True
+                        log.debug('Deleting empty output dataset: %s', data)
+        for key, value in self.output_collections.items():
+            if value.when_empty:
+                collection = out_collections.get(key, None)
```
jmchilton, line 1732: If you're mapping over this tool will collection be a DC (DatasetCollection) here instead of an HDCA (HistoryDatasetCollectionAssociation)? The operations below wouldn't make sense or be possible in that context I believe. I'd strongly recommend (insist on?) adding an API test test_tools.py that maps over an example of such a tool with an output collection and clarify what the behavior should be in that context.

jmchilton:
I'd recommend breaking this up into a series of PRs of increasing controversy so we can weigh in on the issues individually. Whether an output dataset is hidden or visible is completely controllable from the tool provided metadata (galaxy.json) - so I'd have no issue adding the hide functionality here for datasets as a simplification of what is already possible.

Likewise hiding collections would be fine by me also - I'd change the tag though to signify it is a suggestion and wouldn't be applied if a collection is being mapped over. ``when_dataset_empty='hide'`` and ``when_collection_empty="suggest_hide"``.

jmchilton:
> Presumably, you might want to also apply this flag to the contents (datasets) within the collection, but that would happen to the dataset OutputObjects, not the collection OutputObject.

There are times where there is a single output tag for a collection - no discover datasets or explicit outputs nested in the XML file (https://github.com/galaxyproject/galaxy/blob/dev/test/functional/tools/collection_creates_list.xml#L12). If I added this tag there - it would seem to be ambiguous if it applied to the collection or the datasets. So I agree with Marius suggestion to use different attribute names I think. Seems more clear regardless.

jmchilton:
Delay because of existing comments.

PR #7615, opened 2019-03-28
Adding new composite datatype Shapefile

Alanamosse:
The idea is to add a new generic GIS class.
It will gather Geographic Information System files. 
 
The first and most used datatype is the shapefile, a composite with at least 3 files.
I was unable to use the optional tag when adding a composite element. (see the .prj commented)
All elements are non-optional at upload.

Some others datatype to come.

Thank you galaxy team for a review. Sorry about the likely many mistakes. first time working on datatypes..

@bgruening


lib/galaxy/datatypes/gis.py
```diff
@@ -0,0 +1,123 @@
...
+        except Exception:
+            if verbose:
+                gal_Log.debug('@@@shapefile set_meta failed %s - dataset %s has no shp ?' % (sys.exc_info()[0], dataset.name))
+            return False
+        if len(flist) == 0:
+            if verbose:
+                gal_Log.debug('@@@shapefile set_meta failed - %s shp %s is empty?' % (dataset.name, shp))
+            return False
+        self.regenerate_primary_file(dataset)
+        if not dataset.info:
```
mvdbeek: Info and blurb shouldn't be written to in `set_meta`, you can do this in a `set_peek` method.

lib/galaxy/datatypes/gis.py
```diff
@@ -0,0 +1,123 @@
...
+        except Exception:
+            if verbose:
+                gal_Log.debug('@@@shapefile set_meta failed %s - dataset %s has no shp ?' % (sys.exc_info()[0], dataset.name))
+            return False
+        try:
+            flist = os.listdir(shp)
+        except Exception:
+            if verbose:
+                gal_Log.debug('@@@shapefile set_meta failed %s - dataset %s has no shp ?' % (sys.exc_info()[0], dataset.name))
+            return False
```
mvdbeek: What is the reason for returning `False` here ? We don't consume `set_meta` anywhere. If you want to indicate that setting metadata failed you can just not catch the exception or reraise it.

lib/galaxy/datatypes/gis.py
```diff
@@ -0,0 +1,123 @@
...
+        """
+        GIS.set_meta(self, dataset, **kwd)
+        if not kwd.get('overwrite'):
+            if verbose:
+                gal_Log.debug('@@@ shapefile set_meta called with overwrite = False')
+            return True
+        try:
+            shp = dataset.extra_files_path
+        except Exception:
+            if verbose:
```
mvdbeek: `dataset.extra_files_path` should be defined for all valid `dataset`s, so I'd remove this exception handling code.

lib/galaxy/datatypes/gis.py
```diff
@@ -0,0 +1,123 @@
...
+                gal_Log.debug('@@@shapefile set_meta failed %s - dataset %s has no shp ?' % (sys.exc_info()[0], dataset.name))
+            return False
+        try:
+            flist = os.listdir(shp)
```
mvdbeek: Again I don't think this can be reached, the directory will be created as soon as you touch the `extra_files_path` property, I'd remove the exception handling.

lib/galaxy/datatypes/gis.py
```diff
@@ -0,0 +1,123 @@
...
+gal_Log = logging.getLogger(__name__)
+
+
+# The base class for all GIS data
+class GIS(data.Data):
+    """
+    base class to use for gis datatypes
+    """
+
+    def __init__(self, **kwd):
```
mvdbeek: You can skip this definition.

lib/galaxy/datatypes/gis.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def regenerate_primary_file(self, dataset):
+        """
+        cannot do this until we are setting metadata
+        """
+        shp = dataset.extra_files_path
+        flist = os.listdir(shp)
+        rval = ['<html><head><title>Files for Composite Dataset %s</title></head><body><p/>Composite %s contains:<p/><ul>' % (dataset.name, dataset.name)]
+        for i, fname in enumerate(flist):
+            sfname = os.path.split(fname)[-1]
```
mvdbeek:
```suggestion
            sfname = os.path.basename(fname)
```

lib/galaxy/datatypes/gis.py
```diff
@@ -0,0 +1,123 @@
...
+            f, e = os.path.splitext(fname)
+            rval.append('<li><a href="%s">%s</a></li>' % (sfname, sfname))
+        rval.append('</ul></body></html>')
+        with open(dataset.file_name, 'w') as f:
+            f.write("\n".join(rval))
+            f.write('\n')
+
+    def set_meta(self, dataset, **kwd):
```
mvdbeek:
```suggestion
    def set_meta(self, dataset, overwrite=True, **kwd):
```

lib/galaxy/datatypes/gis.py
```diff
@@ -0,0 +1,123 @@
...
+        """
+
+        """
+        GIS.set_meta(self, dataset, **kwd)
+        if not kwd.get('overwrite'):
```
mvdbeek:
```suggestion
        if not overwrite:
```

lib/galaxy/datatypes/gis.py
```diff
@@ -0,0 +1,123 @@
...
+        try:
+            flist = os.listdir(shp)
+        except Exception:
+            if verbose:
+                gal_Log.debug('@@@shapefile set_meta failed %s - dataset %s has no shp ?' % (sys.exc_info()[0], dataset.name))
+            return False
+        if len(flist) == 0:
+            if verbose:
+                gal_Log.debug('@@@shapefile set_meta failed - %s shp %s is empty?' % (dataset.name, shp))
+            return False
```
mvdbeek, line 97: If it failed you should raise an Exception here.

mvdbeek (commented):
Thanks for working on this @Alanamosse, it looks like a good start. Three things in addition to what I mentioned in the review:
  - do you have a small test file that you could upload somewhere ? 
  - could you please drop all of the logging ?
  - What is the issue you're seeing with the optional Coordinate system informations file ?

Alanamosse:
Thank @mvdbeek  for your help !

Here are some shapefiles you can use to test uploads : https://github.com/Alanamosse/Shapefile_eg_data
You can see there are always 3 non-optional files (.shp .shx .dbf) + a various number of optional.
However, I was unable to left optional files empty in the galaxy interface during upload : Can't click on "Start" button when all files are not selected.
I saw the same behavior with openbabel datatype (obfs defined in molecule.py).

This work is strongly inspired (aka copy/pasta) from the Rgenetics class defined in genetics.py, some of your comments may also be legit there.

I tried to apply your advices, but I'm not sure to fully understand how the ``set_meta`` works.
Thks again

mvdbeek:
Ah, this is going to conflict with the PR we merged into 19.05, I'll revert this

PR #7616, opened 2019-03-28
[19.01] Improve string escaping and changeset validation on toolshed

martenson:


nsoranzo:
This should target 18.05, I think.

Also, I'm going to push a commit that should fix https://jenkins.galaxyproject.org/job/docker-toolshed/13508/testReport/junit/shed_functional.functional.test_0000_basic_repository_features/TestBasicRepositoryFeatures/test_0130_verify_handling_of_invalid_characters/

martenson:
the previous similar PR from January went to 18.09, but I am not against backporting

jmchilton:
We shouldn't support tool sheds and haven't really promised to anywhere - let alone 18.05 tool sheds - I feel like targeting 19.01 is sufficient.

nsoranzo:
@jmchilton Since we release the Tool Shed source code together with Galaxy's, I don't see how the Security Policy doesn't apply to it. Obviously we could amend the policy, if we decide to.

jmchilton:
The security policy over and over mention's Galaxy's version, Galaxy instances, etc... it doesn't mention "code" or "code shipped with Galaxy" or alternative servers, or the Tool Shed. I think my reading of it is entirely reasonable.

nsoranzo:
There has never been a separate Tool Shed repository or separate Tool Shed releases, Galaxy source code imports TS and viceversa. This clearly indicates to me that "Galaxy" in the Sec. policy refers to the entire "project".

martenson:
Unless there are big obstacles/conflicts I volunteer to backport it.

nsoranzo:
@martenson I've fixed the tests, can you double check that this still fix the original issue?

martenson:
from gitter

> @nsoranzo I suspect there are more than 3 statuses
we might use 'warning'?
and I don't think we use 'info'
info seems like a reasonable default though, so let's just add warning?

nsoranzo:
Thanks @martenson , I've rebased the latest commit adding `warning`.

martenson:
@nsoranzo bleach seems to work well in my testing of these endpoints. This is ready I think.

nsoranzo (approved):

mvdbeek (approved):

nsoranzo:
Merged forward.

martenson:
backport: https://github.com/galaxyproject/galaxy/pull/7626

PR #5421, opened 2018-01-30
include `hidden` into the galaxy.xsd

bgruening:
The `chart` tool for example is using the tag.

PR #7619, opened 2019-03-28
[PROCEDURES] exempt documentation (docs, release notes) and tests from mandatory review

martenson:


nsoranzo:
What's the rationale for this exceptions?

martenson:
It speeds up things, helps to avoid context-switching, and has no downside. We don't need typo-free release notes and if something is missing in the docs/notes you can just add it anyways. I think we've never had internal conflict or -1 about what should or should not go to docs or release notes. 

This also smoothes out the way for https://github.com/galaxyproject/galaxy/pull/7524 which would otherwise add a substantial overhead given the number of sub projects with independent releases.

The tests are somewhat similar - modifying tests does not affect anything functionally. They should still be passing but I do not think we need to overdo our procedures to control contributions at that level.

jmchilton (approved):

jmchilton:
Rationale well articulated @martenson - this would eliminate my reservations around #7524 completely I believe.

bgruening (approved):

doc/source/project/organization.rst
```diff
@@ -97,10 +97,12 @@ Direct Commit Access
 
 A *committer* may only commit directly to the Galaxy repository (i.e. outside of
 a pull request and not following the procedures described below) the following
-two categories of changes:
+four categories of changes:
 
 * Patches for serious security vulnerabilities.
 * Cherry-picking and/or merging of existing approved commits to other branches.
+* Documentation-only changes.
```
mvdbeek, line 104: Documentation is one of the most important things (we don't do enough), and if we add it we should at least have another set of eyes look over it, because writing decent documentation is harder than coding, I think. If the docs are misleading or wrong it's worse than having no documentation. Direct commits can go completely unnoticed, and I don't think we want that for important changes, which documentation is IMO.

My proposal would be `Release notes and changelogs.`, and amend line 141 to say that documentation-only changes to released branches require a single +1.

martenson: @mvdbeek is your concern only for release branches?

mvdbeek: No, if it goes unnoticed into dev it will go unnoticed into release.

mvdbeek: I edited the my comment to say +1 (instead of +2) for doc-only changes to release branches.

martenson: I am having hard time finding a PR that had wrong or misleading documentation and review helped with it. I agree we need more documentation but refusing to lower the bar to add it is not helping the case I feel.

mvdbeek: There's probably a lot more, some of them are not only, but mostly documentation-only.

https://github.com/galaxyproject/galaxy/pull/2096
https://github.com/galaxyproject/galaxy/pull/2700
https://github.com/galaxyproject/galaxy/pull/7148
https://github.com/galaxyproject/galaxy/pull/7342
https://github.com/galaxyproject/galaxy/pull/5421

IMO we don't alter the bar either way, external contributions still need a single +1, and with my proposition they will also just need a single +1 for released branches. I doubt anyone would be deterred from writing documentation just because it'll go through review. Personally I'd feel more confident knowing someone has seen the changes and understands what I've written, because I know that many things that seem obvious to me are not.

martenson: Only 2 of those 5 you listed would be affected by the change. And the one from Bjoern would not be wrong if it was merged.

> I doubt anyone would be deterred from writing documentation just because it'll go through review.

But this is exactly how it works in my opinion. Compare changing the docs with one click in the github interface to get editor and then one click to commit with creating a pull request from your fork and waiting for reviews.

nsoranzo: I'm with @mvdbeek, breaking the doc building is quite easy, and I already advocated for having a Jenkins job for docs on all PRs.

mvdbeek: > And the one from Bjoern would not be wrong if it was merged.
> 

Not wrong but unfortunate. We can debate this, I also just spent 5 minutes going through doc PRs, but there's many more that are mixed with code but have important changes to docs that were not easily digestible at the first attempt. No reason to think that this couldn't or wouldn't happen to doc only changes.

> But this is exactly how it works in my opinion. Compare changing the docs with one click in the github interface to get editor and then one click to commit with creating a pull request from your fork and waiting for reviews.

I disagree there. If this was about lowering the bar for non-committers I'd buy this as an argument, but I don't for comitters.
I think we should be playing by the same rules as much as possible as all non-committers. I'm not going to veto this, but I think this will lead to unfortunate situations down the road.

jmchilton: This is about requiring reviews - I think people would still generally submit PRs and want them green?

I rarely write docs for Galaxy and write many more docs for Planemo. I suspect all the context switching and heavy processes around Galaxy contributes to that. Docs are important that doesn't mean they "need" reviews. We introduced mandatory reviews to prevent people from breaking things in ways we couldn't fix them - none of the docs rise anywhere near that level.

Also about the difference between release branch and dev - we should drop the second +1 all together for release branch fixes. I care more about slowing down the addition of new APIs we're committed to than slowing down bug fixes.

nsoranzo: > Also about the difference between release branch and dev - we should drop the second +1 all together for release branch fixes. I care more about slowing down the addition of new APIs we're committed to than slowing down bug fixes.

I'd support that.

jmchilton: > I think we should be playing by the same rules as much as possible as all non-committers.

This speaks to me of course - my counter is that you're implicitly a committer +1. Committers actually have a higher bar in some ways because they need two committers to agree to the changes instead of one. Many times I'd have merged an external PR and opened a new PR with fixes but didn't because it is easier to work with the open PR.

mvdbeek: That is not unreasonable @jmchilton, and I get that there's more context switching if you need to come back and re-word the documentation. But I'd say there's only cost to this if there was actually something you had to get back to. Minor typos we can fix with the `suggestion` feature, so that's just a click of a button.

dannon:
So were there any forthcoming changes to this?  It can't be merged as-is (discussion in https://github.com/galaxyproject/galaxy/pull/974#issuecomment-153405955, and https://github.com/galaxyproject/galaxy/pull/1024 -- this is something I've tried to change before), so if we think the dust is settled we should probably open a new "clean" PR for voting.

martenson:
new iteration here: https://github.com/galaxyproject/galaxy/pull/7706

PR #7620, opened 2019-03-28
Generate OIDC callback URL on the fly.

VJalili:
With this PR, we auto-generate OIDC callback URLs, which are generally admin-set endpoints. 
These OIDC callback endpoints are composed as: 

```
<client_base_url>/authnz/<provider>/callback
```

for instance:
```
http://localhost:8080/authnz/google/callback
```

Admins still need to be aware of these endpoints, because they need to enter them for configuring their Galaxy server with an IdP. However, since there is no second way of setting this URL (e.g., you cannot set it `<client_base_url>/oidc/<provider>/`), it makes it ideal for an auto-configuration, which this PR implements.

ping @natefoo @martenson


config/oidc_backends_config.xml.sample
```diff
@@ -1,9 +1,61 @@
...
+
+
+Please mind `http` and `https`.
+
+-->
 <OIDC>
     <provider name="Google">
         <client_id> ... </client_id>
         <client_secret> ... </client_secret>
-        <redirect_uri>http://localhost:8080/authnz/google/callback</redirect_uri>
```
martenson, line 6: Why was this dropped? This tag stayed for the globus config.

config/oidc_backends_config.xml.sample
```diff
@@ -1,9 +1,61 @@
...
+
+    <host_url>/authnz/<provider>/callback
+
+
+Some examples:
+
+- using localhost to authenticate with Google:
+    http://localhost:8080/authnz/google/callback
+
+- using localhost to authenticat with Globus:
```
martenson:
```suggestion
- using localhost to authenticate with Globus:
```

lib/galaxy/authnz/managers.py
```diff
@@ -100,8 +100,7 @@ def _parse_oidc_backends_config(self, config_file):
     def _parse_idp_config(self, config_xml):
         rtv = {
             'client_id': config_xml.find('client_id').text,
-            'client_secret': config_xml.find('client_secret').text,
-            'redirect_uri': config_xml.find('redirect_uri').text}
```
martenson, line 104: Are we positive the on-the-fly URL generation will never fail? - proxies, mirrors etc. Because here we are removing the option for admins to configure it.

martenson: Also are we confident that Admin will use the same `redirect_uri` for IdP as the one we will generate?

VJalili: I have not tested `trans.request.host_url` with proxies or mirrors. If anyone have proxy/mirror setup in place, I would appreciate if this they can test this. Otherwise, I can set up proxy and mirror.

config/oidc_backends_config.xml.sample
VJalili: I guess I missed it, I will remove it for Globus as well. We're dropping it mainly because admins do not have any "freedom" setting up this setter (e.g., they cannot change `/authnz/google/callback` to `/authentication/another_callback/google`. We had it user defined, because (a) they must know this and set it correctly when they're registering their Galaxy with the idp, and (b) it was not clear what is the best method for getting the base URL (i.e., the `http://localhost:8080/` part of the sample). 

However, in this PR, we're updating docs making sure admins know how to set this url with the idp (reg. (a)); and we think `trans.request.host_url` is our best option for getting the base URL (reg. (b)). 

Hence, I think we can drop this setter, and make config easier.

VJalili:
It seems `request.host_url` is not guaranteed to get the actual url when the server is hosted behind some setups, such as proxy (see [this topic](https://stackoverflow.com/questions/690025/how-to-tell-if-a-request-is-coming-from-a-proxy) for instance). 

Also, if we make the `redirect_uri` optional, admins may not set it, and they would not know if it is working until their users cannot login (e.g., see error messages on the IdP complaining redirect URL in the request does not match with the URL registered for that client). 

Therefore, I would recommend closing this PR now, and we can get back to it when we have a guaranteed method for getting the URL at which a Galaxy instance is hosted.

martenson:
@VJalili I believe the improvements to the configuration docs should still make it, can you please open a new PR with that?

VJalili:
@martenson Yes, I will create a separate PR for that.

PR #7621, opened 2019-03-29
Large refactoring of model migrations

nsoranzo:
Put common code in `lib/galaxy/model/migrate/versions/util.py` .
The refactoring is still incomplete, but already results in 600 less lines
of code.


dannon:
This looks great, thanks!

PR #7628, opened 2019-03-30
Add tool directory to dynamic_options global namespace

qiagu:
Maybe this is not good idea.
Basically, I want to the `code` tag file able to access the tool directory. Because of the current implementation of `code` file , it seems impossible to take advantage of the `__file__` attribute from inside of the `code` file. In the other hand, it's also not easy to pass the `__tool_directory__` to dynamic_options as an argument in tool wrapper. Instead, this PR pushes the `__tool_directory__` to the global namespace. Then `code` function can directly call the variable without passing it as an argument.
If there is a better solution, please let me know. 
Thanks.


qiagu:
I'll find a new way for my tool wrapper.

Issue #983, opened 2015-10-23
Don't mark discovered added datasets as complete until all are added


hexylena:
This is a problem only while:
- dataset collections are still being introduced and are not ubiquitous
- #740 isn't complete (so I can't switch all of my tools to producing/consuming DCs)

Dynamically discovered datasets are marked as complete even when they're not all discovered yet. This is counterintuitive and I feel to be incorrect behaviour. The job is not complete if all of the outputs haven't been discovered yet. Users have been _conditioned_ to accept "green = job complete" so dynamic discovery breaks this expectation. Users are not informed in **any** way that they're supposed to wait on outputs to show up.

I think I've mentioned this on IRC once, but I'm now filing a proper bug report (on behalf of my PI), as he emailed me a couple times this morning being very confused that not all of his datasets had shown up. When he went to run the tool again (thinking it failed in some weird way) discovered that they were all there.

hexylena:
Oh, I misread this one. Sorry for the noise, thought it was about collections and was solved by the new progress bar.

mblue9:
Would this be the same issue as what I encountered below? Collection appears green but the jobs are all stuck in a paused state. Because the collection was green I thought they had worked and had started trying to run other tools on the output. Would be good if the collection turned green only when the jobs were finished.

![screen shot 2018-07-09 at 4 10 03 pm](https://user-images.githubusercontent.com/5591821/42433466-d469065c-8392-11e8-81bd-b5f8ca973da4.png)

mvdbeek:
Should be fixed by https://github.com/galaxyproject/galaxy/pull/7633

PR #7633, opened 2019-03-31
[19.01] Delay workflow step execution for discovered & mapped-over input

### LINKED ISSUE #5867: Flatten Collection runs before job that discovers collection elements
I am trying to use the 'Flatten Collection' tool to flatten a nested list within a workflow, but it produces an empty list.

When used outside a workflow, it works fine. Also, if the nested list is a workflow input connected to 'Flatten Collection', it works. But if the nested list is generated by another tool as part of the workflow and feeds into 'Flatten Collection', it won't work.

Thanks
Anil

mvdbeek:
We should only finalize collections up until the map over depth,
as indicated by the TODO that is now obsolete.
Fixes #5867 and probably
a bunch of other issues where workflows don't run to completion.

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -3445,14 +3445,13 @@ class DatasetCollection(Dictifiable, UsesAnnotations, RepresentById):
         self.populated_state = DatasetCollection.populated_states.FAILED
         self.populated_state_message = message
 
-    def finalize(self):
+    def finalize(self, collection_type_description):
         # All jobs have written out their elements - everything should be populated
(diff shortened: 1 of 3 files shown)


mvdbeek:
The one failed API test is not relevant, it's because we download something form usegalaxy.org which is down for maintenance atm.

jmchilton (approved):
Amazing, thanks!

mvdbeek:
Also fixes https://github.com/galaxyproject/galaxy/issues/983.

qiagu:
@mvdbeek Thanks. This is so great. It solves one of my workflow issues permanently.

Issue #7431, opened 2019-02-28
Workflow Editor - validation and propagation of singular/collection input swaps


dannon:
![image](https://user-images.githubusercontent.com/155398/53590345-88872180-3b5f-11e9-8f2e-85332126e971.png)

Given the above workflow, if you want to swap from a singular input to a collection you must first disconnect the link between the two random lines steps.  We should be able to validate and propagate an input change from singular to collection on the fly.

(didn't see one at first glance, but if there is an open issue for this already feel free to link it up and close out)

jennaj:
Related ticket https://github.com/galaxyproject/galaxy/issues/5632

jennaj:
Example: Thread today at Galaxy Help that links to another related thread (same issue from an end-user perspective): https://help.galaxyproject.org/t/workflow-is-not-operating-on-a-collection/586/2

jennaj:
**Note for end-users:** If your workflow is not executing correctly, perhaps with no inputs, or seems to "stop" at some point even with inputs included, please try the following (until this ticket closes):

* Make sure that there are at least one of the "Inputs" included at the start of your workflow.
* Every tool should have an "input" connected to it, or the output from a prior tool connected to it.
* **If an "input" is added or modified in any way** (including the collection type), or if the input/output for a tool is modified, in many cases you must: 1) disconnect all existing noodles, 2) reconnect all noodles. This resets the metadata internal to the workflow.
* It is important to *start connecting noodles from the "input" through downstream tools* in the order of data flow/execution. 
* Be aware of what constitutes an internal workflow metadata change. Many tools accept more than one input type/datatype. Some tools include parameter selections that inpact the output datatype. And "post job actions" permit user-defined output datatype modifications. _Any of these changes might result in conflicts resulting in execution problems._

martenson:
This is another report: https://help.galaxyproject.org/t/workflow-does-not-show-in-history/934 And it does not involve collections.

I think this issue is a big problem. There is no intuitive way to solve it, the fact you have to connect tools “from left to right” in order to be able to execute workflow is illogical.

jmchilton:
In re:

> the fact you have to connect tools “from left to right” in order to be able to execute workflow is illogical


[...]

PR #7638, opened 2019-04-01
[WIP] Provide feedback on why workflow editor nodes won't connect.

jmchilton:


dannon:
Even though I think there are more things I'd like to see re: #7431, as mentioned on gitter, feedback like this is definitely a good step forward.

Toasts don't seem a natural option for displaying this, here, though.  I'd probably expect a tooltip on the drop destination when I hovered over an incompatible map -- think you could swing that?

PR #7639, opened 2019-04-01
Delete cloudauthz record from database instead of marking it as deleted

VJalili:
@martenson can we modify this migration script? I guess no newer script exists, and previous script was not included in 19.01 release.


martenson:
@VJalili there is no migration-consistency check, so instances running `dev` will not re-run the migration and will be stuck with the column. Which is probably fine.

VJalili:
Agree.

jmchilton:
Agree also, merging fast to minimize the damage though. Thanks!

VJalili:
Thanks @jmchilton

PR #5652, opened 2018-03-07
[WIP] Bower elimination

dannon:
WIP, opening PR to get tests running.

This PR updates a few libraries but is primarily focused on swapping currently repo-based libs to being used out of node_modules and works towards completely removing the formerly-bower-staged libs.

PR #7640, opened 2019-04-01
Toastr debowerization, refactoring.

dannon:
This pulls a chunk out of #5652 and updates it, swapping from bower-based (in-repo) toastr to node_modules, and follows up on a little additional refactoring.  This leaves a single ui/toast.js integration point from which we can stub in whatever toast library we'd like.  Right now I'm looking into options that'll make use of https://getbootstrap.com/docs/4.2/components/toasts/, since we're all-in on bootstrap now.


dannon:
Sample use at https://github.com/galaxyproject/galaxy/commit/ac87bfcb96cd23719601dbfd4670f68a8737cd48, debated pulling it into this PR but it needs more styling and I didn't know if I wanted to commit to vue-toasted.

mvdbeek (approved):
Nice, thanks dannon.

PR #7641, opened 2019-04-01
Continue gracefully in set_meta if unable to count number of dataset lines.

blankenberg:
Possibly caused by datatype inconsistency (e.g. binary set as text), which is corrected elsewhere, but would cause otherwise unnecessary job failure here.

### DIFF
--- a/lib/galaxy/datatypes/data.py
+++ b/lib/galaxy/datatypes/data.py
@@ -802,10 +802,13 @@ class Text(Data):
         """
         data_lines = 0
         with compression_utils.get_fileobj(dataset.file_name) as in_file:
-            for line in in_file:
-                line = line.strip()
-                if line and not line.startswith('#'):
-                    data_lines += 1
+            try:
+                for line in in_file:
+                    line = line.strip()
+                    if line and not line.startswith('#'):
+                        data_lines += 1
+            except Exception:
+                pass
         return data_lines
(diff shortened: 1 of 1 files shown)


martenson:
lgtm. Should we even have upper bound of how many lines we care about?

lib/galaxy/datatypes/data.py
```diff
@@ -802,10 +802,13 @@ def count_data_lines(self, dataset):
         """
         data_lines = 0
         with compression_utils.get_fileobj(dataset.file_name) as in_file:
-            for line in in_file:
-                line = line.strip()
-                if line and not line.startswith('#'):
-                    data_lines += 1
+            try:
```
mvdbeek, line 808: Is that an encoding issue ? I.e if you upload something that doesn't have a utf-8 equivalent ?
The change is fine with me, but if that's the case can you add a comment so that we can take this out if / when we have found a better fix ?

blankenberg: Yes, this is essentially the problem, with the specific issue I encountered being unable to determine how to split the file data into lines. 

Also agree that we should have a better fix. Would be great to be able to log the issue, and not have the set_meta process cause a job failure. Will add fixme note.

mvdbeek (approved):

PR #7642, opened 2019-04-01
Provide feedback on why workflow editor nodes won't connect.

jmchilton:
Redone to use tooltips instead of Toastr based on https://github.com/galaxyproject/galaxy/pull/7638#issuecomment-478640571.

### DIFF
--- a/client/galaxy/scripts/mvc/workflow/workflow-terminals.js
+++ b/client/galaxy/scripts/mvc/workflow/workflow-terminals.js
@@ -569,7 +569,7 @@ var InputCollectionTerminal = BaseInputTerminal.extend({
                 return new ConnectionAcceptable(false, "Incompatible collection type(s) for attachment.");
             }
         } else {
-            return new ConnectionAcceptable(false, "Cannot attach a data output to a collection output.");
+            return new ConnectionAcceptable(false, "Cannot attach a data output to a collection input.");
         }
     }
 });
--- a/client/galaxy/scripts/mvc/workflow/workflow-terminals.js
+++ b/client/galaxy/scripts/mvc/workflow/workflow-terminals.js
@@ -5,8 +5,10 @@ import Backbone from "backbone";
 // TODO; tie into Galaxy state?
 window.workflow_globals = window.workflow_globals || {};
 
-const COLLECTION_MAPPING_INCOMPATIBLE_OUTPUT = "This connection would change how this tool is being mapped over, but mapping is already constrained in an incompatible way by outputs. Disconnecting target step's outputs may fix this.";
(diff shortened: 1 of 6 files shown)


mvdbeek:
This is almost perfect!
workflow notification

Two minor issues: The tooltip is sticky (see when I move the input collection out of the way so I can connect to bwa) and can get in the way, and we probably need to make the input selection trigger a change when we leave the focus (the last one has of course always been a problem and is not related to this PR, it's just more obvious now that we see why we can't connect).

client/galaxy/scripts/mvc/workflow/workflow-terminals.js
```diff
@@ -531,20 +552,25 @@ var InputCollectionTerminal = BaseInputTerminal.extend({
...
                 if (mappingConstraints.every(effectiveMapOver.canMatch)) {
                     return this._producesAcceptableDatatype(other);
+                } else {
+                    return new ConnectionAcceptable(false, COLLECTION_MAPPING_INCOMPATIBLE_INPUT);
                 }
+            } else {
+                return new ConnectionAcceptable(false, "Incompatible collection type(s) for attachment.");
             }
+        } else {
+            return new ConnectionAcceptable(false, "Cannot attach a data output to a collection output.");
```
nsoranzo: Noticed in @mvdbeek's video:

```suggestion
            return new ConnectionAcceptable(false, "Cannot attach a data output to a collection input.");
```

jmchilton:
The over-sticky property is simultaneously easy and hard to produce - like those old stick "X"s. I pushed an adjustment (1a2a7bc1d7649258647d206fb6ddf6a05f735ddd) that I think fixes it - but I have no confidence obviously since I can't consistently produce it to start with.

The side form has always been a bit touchy given there is no obvious save button or anything, seems out of scope for this PR?

dannon:
@jmchilton Thanks for swapping it over to tooltips, this is much better.  I was able to reproduce the sticky tips pretty reliably last night, so will check in a sec.

It's also minor, but did you look for other existing in-ui colors to use for the no-connection noodle interior?

dannon:
Tooltips are good to go now, they reliably dispose.

The other thing I noticed was this:

image

In this case, it's actually the *outputs* that need to be disconnected, right?

jmchilton:
> In this case, it's actually the outputs that need to be disconnected, right?

Yeah - I originally had these messages inverted based on my understanding of the code and found a case where they were wrong and swapped them and it seemed right empirically though I was confused why. I'll try to work through more combinations and see if I can refine this.

dannon:
@jmchilton Yeah, l saw that change in 79c61f8e65fb5983e797026d7731b8a22548c8a0 and tried to track through but you're right -- the logic gets complicated fast.

jmchilton:
Re-branched the new branching for precision - so that is why there are so many unit tests. I could potentially branch some more to to make the language ``(and potentially outputs)`` unconditional and use named constants and update the unit tests.

dannon:
This looks great and the failing test is unrelated.  Thanks @jmchilton.

PR #7644, opened 2019-04-02
Allow purging toolshed repository white ghosts using API.

blankenberg:


lib/galaxy/webapps/galaxy/api/tool_shed_repositories.py
```diff
@@ -704,6 +704,45 @@ def repair_repository_revision(self, trans, payload, **kwd):
         # Display the list of repaired repositories.
         return tool_shed_repositories
 
+    @expose_api
+    @web.require_admin
+    def purge_repository(self, trans, id=None, **kwd):
+        """
+        Purge a "white ghost" repository from the database.
```
mvdbeek, line 711: I know that's what we have in the equivalent controller method, but "white ghost" is not great as description.
```suggestion
        Purge a repository that is stuck in "NEW" state from the database.
```

mvdbeek (approved):

lib/galaxy/webapps/galaxy/api/tool_shed_repositories.py
```diff
@@ -704,6 +704,45 @@ def repair_repository_revision(self, trans, payload, **kwd):
...
+
+        GET /api/tool_shed_repositories/{id}/purge
+
+        :param id:  encoded repository id. Either id or name, owner, changeset_revision and tool_shed_url need to be supplied
+        :param kwd: 'name'   : Repository name
+                    'owner'  : Repository owner
+                    'changeset_revision': Changeset revision to uninstall
```
dannon, line 718:
```suggestion
                    'changeset_revision': Changeset revision to purge
```

lib/galaxy/webapps/galaxy/buildapp.py
```diff
@@ -960,6 +960,12 @@ def populate_api_routes(webapp, app):
                           action='status',
                           conditions=dict(method=["GET"]))
 
+    webapp.mapper.connect('tool_shed_repository',
+                          '/api/tool_shed_repositories/:id/purge',
+                          controller='tool_shed_repositories',
+                          action='purge_repository',
+                          conditions=dict(method=["GET", "POST"]))
```
dannon, line 967: Probably shouldn't allow purges via GET -- can we swap this to POST only?

lib/galaxy/webapps/galaxy/api/tool_shed_repositories.py
jmchilton: +1

nsoranzo:
Can't we just modify the `ToolShedRepositoriesController.uninstall_repository()` method to call `irm.purge_repository()` instead of `irm.uninstall_repository()` if `repository.is_new` is True?

That would mean using the correct REST method, i.e. `DELETE`.

blankenberg:
I think these are all great points, and making DELETE work as indicated sounds like a smart direction to move in. I likely wont have time to do this in the near future however, so please feel free to close this PR. 

(Also, I've already used these changes to delete the ~200 or so items that were 'annoying' me on one particular instance, so there is no real rush from my end)

It might be useful to sketch out this API in general, before any additional implementation.

dannon:
I really like @nsoranzo's suggestion, too.  I guess let's close this one out as you suggest @blankenberg and revisit that approach.

PR #7645, opened 2019-04-02
Adjust test_run_with_numeric_input_connection to extra line

mvdbeek:
introduced in https://github.com/galaxyproject/galaxy/pull/7634


jmchilton (approved):
Thanks so much!

PR #7646, opened 2019-04-02
[19.01] Fix incorrect lddas being used as collection members

martenson:
Adjust library folder api to return ldda_id on ld items
and use it for the obect passed to colelction creater
instead of the ld_id.

This bug would only materialize if you have Galaxy isntance that
has a library dataset with multiple versions. Which is the reason why
it eluded us for some time. That and missing tests.


jmchilton (approved):
Thanks for the fix, sorry about that.

martenson:
@jmchilton I broke that 😭

PR #7647, opened 2019-04-02
Fix genomespace_importer tool

blankenberg:


mvdbeek (approved):
Should this be targeting 19.01 ?

blankenberg:
> Should this be targeting 19.01 ?

I didn't double check today, but IIRC the underlying changes that broke this part happened after 19.01.

(Version inconsistency with cloudbridge to a new version, when lower version was needed by python-genomespace-client; plus changing of default Tool handling.)

dannon:
Yeah, I just confirmed it actually does work in 19.01, and that this fixes genomespace in dev.

In both cases, the exporter was throwing a `WARNING:cloudbridge.providers.aws.services:AWS Bucket genomespace-prod already exists but user doesn't have enough permissions to access the bucket` causing the job to appear to fail, even though it does work, but that's yet another separate issue I believe.

nuwang:
The cloudbridge warning is usually caused by the user not having list permissions on the bucket. I've changed the warning to be less misleading: https://github.com/CloudVE/cloudbridge/commit/a181fcce4a3f1a6c83362b28cd9db55a3d4a21ae

dannon:
@nuwang Yeah, the clarified error is a good idea, and the file does upload just fine, but it's still going to cause the resultant dataset to be in 'error' state in Galaxy instead of 'ok'.  What we probably need to do here is update the tool xml to ignore stderr and look for an exit code.

https://github.com/galaxyproject/galaxy/pull/7647#issuecomment-481208063 might do it?

PR #7650, opened 2019-04-03
Fix composite data and metadata file handling for UUID object stores

jmchilton:
https://github.com/galaxyproject/galaxy/pull/7154 added the ability to store objects in an object store by UUID. This was needed for working with "discovered" datasets in the context of remote workers writing to the object store. In that case, these objects will not have an ID yet and no database should assume to be present, so the remote code would have no way of knowing where to store the data if keyed on ID. UUIDs offer a mechanism to let the remote code decide where to put the data.

#7154 enabled some very simple tests to pass on the remote data and execution branch (#7058), but running the full suite of tool framework tests - it became quickly apparent composite and metadata files need to be handled also.

### DIFF
--- a/lib/galaxy/datatypes/data.py
+++ b/lib/galaxy/datatypes/data.py
@@ -372,7 +372,9 @@ class Data(object):
             return smart_str(data)
         if filename and filename != "index":
             # For files in extra_files_path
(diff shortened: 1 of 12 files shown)


mvdbeek (approved):

jmchilton:
Thanks for the merge. I've refactored the migration in here to use the new migration utilities in #7670.

mvdbeek:
Ah, I missed that, thanks for following up.

PR #7651, opened 2019-04-03
Implement functional dict-ifiable tool outputs and error detection.

jmchilton:
These objects can be completely recovered from serialized representation now. This is needed downstream for projects#11 to serialize data needed for extended metadata collection on worker nodes.

### DIFF
--- a/lib/galaxy/tools/parser/output_collection_def.py
+++ b/lib/galaxy/tools/parser/output_collection_def.py
@@ -58,7 +58,9 @@ def dataset_collector_descriptions_from_list(discover_datasets_dicts):
 
 
 def dataset_collection_description(**kwargs):
-    if asbool(kwargs.get("from_provided_metadata", False)):
+    from_provided_metadata = asbool(kwargs.get("from_provided_metadata", False))
+    discover_via = kwargs.get("discover_via", "tool_provided_metadata" if from_provided_metadata else "pattern")
+    if discover_via == "tool_provided_metadata":
         for key in ["pattern", "sort_by"]:
             if kwargs.get(key):
                 raise Exception("Cannot specify attribute [%s] if from_provided_metadata is True" % key)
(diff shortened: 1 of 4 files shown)


jmchilton:
No test changes here but the tool framework tests applied to the remote metadata branch exercise every line of this I believe.

lib/galaxy/tools/parser/output_collection_def.py
```diff
@@ -79,6 +81,17 @@ def __init__(self, **kwargs):
         self.directory = kwargs.get("directory", None)
         self.recurse = False
 
+    def to_dict(self):
+        return {
+            'discover_via': self.discover_via,
```
bgruening, line 86: discover_via is not part of the class DatasetCollectionDescription isn't it?

jmchilton: The two concrete subclasses that are actually instantiated both define this I think. I can ``@six.add_metaclass(abc.ABCMeta)`` / ``@abc.abstractproperty`` if you'd like me to.

lib/galaxy/tools/parser/output_objects.py
```diff
@@ -232,6 +277,26 @@ def collection_prototype(self, inputs, type_registry):
             collection_prototype.collection_type = collection_type
         return collection_prototype
 
+    def to_dict(self):
+        return {
+            'collection_type': self.collection_type,
+            'collection_type_source': self.collection_type_source,
+            'collection_type_from_rules': self.collection_type_from_rules,
+            'structured_like': self.structured_like,
+            'discover_datasets': list(map(lambda d: d.to_dict(), self.dataset_collector_descriptions)),
```
mvdbeek:
```suggestion
            'discover_datasets': [d.to_dict() for d in self.dataset_collector_descriptions],
```
is easier on my eyes but completely fine either way. Probably not worth re-running the tests for.

mvdbeek (approved):
Cool!

PR #7657, opened 2019-04-03
Refactor IntegrationTestCase to allow integration test functions.

jmchilton:
I like what @mvdbeek did with test_datatype_upload.py in turning a test class into a fixture, but I wanted to formalize it a bit and eliminate the hackiness around re-defining ``runTest`` and instantiating a TestCase just to run a test method outside the test case.

I want to be able to make more integration tests like test_pulsar_embedded.py, where we just define a configuration and then some tool tests to run. In particular this will be useful for extended metadata testing of the most interesting tools.

### DIFF
--- a/test/base/integration_util.py
+++ b/test/base/integration_util.py
@@ -7,6 +7,8 @@ tessting configuration.
 import os
 from unittest import skip, SkipTest, TestCase
 
+import pytest
+
 from galaxy.tools.deps.commands import which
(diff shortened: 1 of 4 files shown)


mvdbeek (approved):
Is it still WIP ? looks great to me

jmchilton:
Great, thanks for the review @mvdbeek !

PR #7659, opened 2019-04-04
Allow import of format2 workflows by default for 19.05.

jmchilton:
Setup forcing export of format2 workflows by default in next release. There should be a least one release between enabling them to be imported by default and exporting them by default anyway I think - so 19.09 workflows can still be used on slightly older Galaxy instances.

### DIFF
--- a/config/galaxy.yml.sample
+++ b/config/galaxy.yml.sample
@@ -1350,8 +1350,10 @@ galaxy:
   # of Galaxy's stable API.
   #enable_beta_workflow_modules: false
 
-  # Enable import and export of workflows as Galaxy Format 2 workflows.
-  #enable_beta_workflow_format: false
+  # Export of workflows as Galaxy Format 2 instead of GA workflows by
+  # default. The default value of this configuration variable will
+  # switch to True in 19.09.
+  #enable_beta_export_format2_default: false
 
   # Following options only apply to workflows scheduled using the legacy
   # workflow run API (running workflows via a POST to /api/workflows).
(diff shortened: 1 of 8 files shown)


nsoranzo:
Can we allow the user to choose the format? Then we can just change the default to format2 directly, and have the format1 is still available as an option in the UI.

jmchilton:
It is certainly an option, and if you insist I can try to work on that but I'd prefer not to I think. I think those are details we don't want users to care about? Also I'm not sure I'm ready for export by default yet. So I prefer this two-stage rollout.

nsoranzo:
> I think those are details we don't want users to care about?

Since the main reason for an export is allowing the import of the workflow on another Galaxy instance, we cannot ignore that there are loads of instances around that are stuck on very old releases.
Also, in the future we will hopefully have the option of exporting in CWL format, so we are going to need this user-visible option any way.
This is not needed for this PR to be merged, though, only when we switch the default, I think.

> Also I'm not sure I'm ready for export by default yet. So I prefer this two-stage rollout.

That's surely fine.

jmchilton:
> This is not needed for this PR to be merged, though, only when we switch the default, I think.

Let's table this then and pick that conversation back up in 19.09. I'm not totally sold and I'm tempted to remove the parts of this that promise the default will switch in 19.09 but perhaps it is too small a point to get so hung up on.

Thanks for the input @nsoranzo, it was well articulated and much appreciated.

mvdbeek (approved):
I think allowing the import should be fine, whether or not we do the default switch I think we can discuss later ? In any case format2 exports are still a little fragile (completely unrelated to the PR).

PR #7665, opened 2019-04-04
Authnz UI and associated api

Nerdinacan:
This is a new user panel for managing cloud authorization keys.

@VJalili implemented the api endpoint, I cherry-picked his commits into this PR so we can evaluate it as a whole.

Since we don't have a site-wide router yet and the number of fields is small for each saved key, I built it as an accordion UI, so it's contained to just one view, which is mounted from the user preferences page.

**components/User/CloudAuth**
Main listing component, this is what gets mounted inside the existing backbone user prefs view.

**CloudAuthItem, CredentialForm, CredentialConfig**
Sub-components for individual items, the static form, and the polymorphic configuration variables. (Config fields change depending on the selected identity provider.)


Nerdinacan:
Probably could use some more test coverage, but otherwise this is ready for review.

jmchilton:
Can you rebase or merge with latest changes?

jmchilton:
Looks like there are linting errors according to the CI? Are those valid?

dannon:
Not totally sure about some of the interactions and transitions here (just in terms of consistency in the app), but that's certainly something we can follow up on down the road.  I merged with dev and pushed a few tweaks/fixes several hours ago, and this looks good to me now.  Thanks!

PR #7667, opened 2019-04-04
Fix shutdown for Pulsar MQ Runner

jmchilton:


natefoo:
Main will be on 19.05 shortly, but should we backport this to 19.01 anyway?

PR #7669, opened 2019-04-05
Workflow parameter via API typing issues

dannon:
Add simple test (that currently fails but should not), demonstrating workflow replacement parameter typing issue.

Currently this results in this error:
```
  File "/Volumes/work_cs/galaxy/lib/galaxy/jobs/runners/__init__.py", line 223, in prepare_job
    job_wrapper.prepare()
  File "/Volumes/work_cs/galaxy/lib/galaxy/jobs/__init__.py", line 863, in prepare
    tool_evaluator.set_compute_environment(compute_environment, get_special=get_special)
  File "/Volumes/work_cs/galaxy/lib/galaxy/tools/evaluation.py", line 78, in set_compute_environment
    visit_input_values(self.tool.inputs, incoming, validate_inputs)
  File "/Volumes/work_cs/galaxy/lib/galaxy/tools/parameters/__init__.py", line 167, in visit_input_values
    callback_helper(input, input_values, name_prefix, label_prefix, parent_prefix=parent_prefix, context=context)
  File "/Volumes/work_cs/galaxy/lib/galaxy/tools/parameters/__init__.py", line 130, in callback_helper
    new_value = callback(**args)
  File "/Volumes/work_cs/galaxy/lib/galaxy/tools/evaluation.py", line 76, in validate_inputs
    value = input.from_json(value, request_context, context)
  File "/Volumes/work_cs/galaxy/lib/galaxy/tools/parameters/basic.py", line 357, in from_json
    raise ValueError("An integer is required")
```

The UI works because the client actually does this parameter substitution prior to submission, even though it also sends along the dict.

### DIFF
--- a/test/api/test_workflows.py
+++ b/test/api/test_workflows.py
@@ -3067,7 +3067,6 @@ steps:
             inputs = {
                 '0': self._ds_entry(hda),
             }
-            workflow_request = {}
             workflow_request = {"replacement_params": dumps(dict(numlines=3))}
             invocation_id = self.__invoke_workflow(history_id, workflow_id, inputs=inputs, request=workflow_request)
 
--- a/test/api/test_workflows.py
+++ b/test/api/test_workflows.py
@@ -3048,6 +3048,38 @@ steps:
         self.__assert_lines_hid_line_count_is(history_id, 3, 5)
 
     @skip_without_tool("random_lines1")
+    def test_integer_param_replacement(self):
+        workflow_id = self._upload_yaml_workflow("""
+class: GalaxyWorkflow
+inputs:
+  test_input: data
+steps:
+  first:
+    tool_id: random_lines1
+    state:
+      num_lines: ${numlines}
+      input:
+        $link: test_input
+""", round_trip_format_conversion=True)
+        with self.dataset_populator.test_history() as history_id:
+            hda = self.dataset_populator.new_dataset(history_id, content="\n".join(str(_) for _ in range(10)))
+            self.dataset_populator.wait_for_history(history_id)
+            inputs = {
+                '0': self._ds_entry(hda),
+            }
+            workflow_request = {}
+            workflow_request = {"replacement_params": dumps(dict(numlines=3))}
+            invocation_id = self.__invoke_workflow(history_id, workflow_id, inputs=inputs, request=workflow_request)
+
+            time.sleep(2)
+            self.dataset_populator.wait_for_history(history_id)
+            self.__review_paused_steps(workflow_id, invocation_id, order_index=2, action=True)
+
+            self.workflow_populator.wait_for_workflow(workflow_id, invocation_id, history_id)
+            time.sleep(1)
+            self.__assert_lines_hid_line_count_is(history_id, 2, 3)
+
+    @skip_without_tool("random_lines1")
     def test_run_replace_params_by_uuid(self):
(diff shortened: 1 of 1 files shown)


dannon:
@mvdbeek This should concretely demonstrate the problem; I don't yet know where in the stack that's getting converted to a string, but dumps there should be doing the right thing and ending up with 
`'{"numlines": 4}'` instead of `'{"numlines": "4"}'`, which I think was your original concern?

test/api/test_workflows.py
```diff
@@ -3047,6 +3047,37 @@ def test_run_replace_params_by_tool(self):
...
+    def test_integer_param_replacement(self):
+        workflow_id = self._upload_yaml_workflow("""
+class: GalaxyWorkflow
+inputs:
+  test_input: data
+steps:
+  first:
+    tool_id: random_lines1
+    state:
+      num_lines: ${numlines}
```
mvdbeek, line 3060:
```suggestion
      num_lines: {numlines}
```

test/api/test_workflows.py
```diff
@@ -3047,6 +3047,37 @@ def test_run_replace_params_by_tool(self):
...
+      input:
+        $link: test_input
+""", round_trip_format_conversion=True)
+        with self.dataset_populator.test_history() as history_id:
+            hda = self.dataset_populator.new_dataset(history_id, content="\n".join(str(_) for _ in range(10)))
```
mvdbeek, line 3065: This needs to be a raw string, and then it actually works

test/api/test_workflows.py
dannon: Ahh, is this a yaml-format specific thing?  When you download a workflow .ga it's the same `${numlines}` as you'd use in the UI.

test/api/test_workflows.py
dannon: Really?  I snagged this straight from another test.

mvdbeek: Argh, I thought so, but for some reason num_lines falls back to 1 in this test, making me think that's the issue

dannon: If I drop the replacement attempt completely, and just set '3' lines in the yaml, this works fine, so I don't think there is a problem here.

test/api/test_workflows.py
mvdbeek: just to recap, the way you're doing it is correct, I got confused by another test

test/api/test_workflows.py
mvdbeek: agreed, it actually looks like the replacement isn't happening at all in this case

mvdbeek:
I think we ultimately figured out that the client needs to fill the param dict, and that's always been the case. We do have a better way to do this using explicitly typed workflow parameters. With the new expression tools we can also create parameters that can replace composed parameters like `#{param} another text`, so this is the way to go IMO.
So can we close this as won't fix ?

dannon:
I dunno, this is a bummer.  It seems like it'd be a nice-to-fix, just for consistency.  We *have* the intended payload, it just needs to be coerced to types like we already must do somewhere, right?

Realistically, I'm not going to find the time to do this one, though, so if nobody else cares I guess it's dead in the water.

dannon:
We've discussed this and decided that we're not going to try to fix these old style parameters to work via the API.  The path forward is through explicit inputs, we just need to improve their usability.

PR #7670, opened 2019-04-05
Track basename at the time of creation for datasets.

jmchilton:
This came up in the context of objectstores at the last devteam meeting and the consensus seemed to be to try to get this in. There was some appetite for at least offering modalities where were treat datasets more like files on a disk. Like sources and hashes added this release cycle, this PR is a bit API/infrastructure-y without a super obvious purpose expressed in the current code base. But tool, tool testing, and web API interfaces for consuming and producing these values are here and more or less tested. Going a bit further, I think it could be arguably be thought of as good data to save for capturing data flow and making things accessible and transparent.

My real purpose is tracking this and getting the column/interfaces in for CWL. All CWL tools and workflows consume data using this field in the CWL branch of Galaxy. This is also the last database change in the CWL branch not yet in Galaxy.

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -2510,6 +2510,16 @@ class DatasetInstance(object):
         """Detects whether there is any data"""
         return self.dataset.has_data()
 
+    def get_created_from_basename(self):
+        return self.dataset.created_from_basename
+
+    def set_created_from_basename(self, created_from_basename):
+        if self.dataset.created_from_basename is not None:
+            raise Exception("Underlying dataset already has a created_from_basename set.")
+        self.dataset.created_from_basename = created_from_basename
+
+    created_from_basename = property(get_created_from_basename, set_created_from_basename)
+
     def get_raw_data(self):
         """Returns the full data. To stream it open the file_name and read/write as needed"""
         return self.datatype.get_raw_data(self)
@@ -3393,6 +3403,7 @@ class LibraryDataset(RepresentById):
                     state=ldda.state,
                     name=ldda.name,
                     file_name=ldda.file_name,
+                    created_from_basename=ldda.created_from_basename,
                     uploaded_by=ldda.user.email,
                     message=ldda.message,
                     date_uploaded=ldda.create_time.isoformat(),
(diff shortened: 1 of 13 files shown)


lib/galaxy/model/migrate/versions/0152_created_from_basename.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def downgrade(migrate_engine):
+    metadata.bind = migrate_engine
+    metadata.reflect()
+
+    __drop_column("created_from_basename", "dataset", metadata)
+
+
+def __add_column(column, table_name, metadata, **kwds):
```
dannon: Probably want to use the new utilities in galaxy.model.migrate.versions.util -- add_column/drop_column?

jmchilton: Hadn’t realized that had merged, sounds great. I’ll redo this and give those a shot.

jmchilton:
Rebased due to conflicted database migration from #7650.

lib/galaxy/model/migrate/versions/0152_add_metadata_file_uuid.py
```diff
@@ -20,13 +21,8 @@ def upgrade(migrate_engine):
...
     metadata_file_table = Table("metadata_file", metadata, autoload=True)
-
-    try:
-        uuid_column = Column('uuid', UUIDType())
-        uuid_column.create(metadata_file_table)
-        assert uuid_column is metadata_file_table.c.uuid
-    except Exception:
-        log.exception("Adding column 'uuid' to `MetadataFile` table failed.")
+    uuid_column = Column('uuid', UUIDType())
+    add_column(uuid_column, metadata_file_table)
```
nsoranzo:
```suggestion
    add_column(uuid_column, 'metadata_file', metadata)
```

and drop line 23. Similarly for `drop_column` and migration 153.

jmchilton: Done, thanks for the catch. Very nice!

jmchilton:
Rebased due to another conflicted database migration 😆.

lib/galaxy/model/__init__.py
```diff
@@ -2495,6 +2495,16 @@ def has_data(self):
         """Detects whether there is any data"""
         return self.dataset.has_data()
 
+    def get_created_from_basename(self):
+        return self.dataset.created_from_basename
+
+    def set_created_from_basename(self, created_from_basename):
+        if self.dataset.created_from_basename is not None:
+            raise Exception("Underlying dataset already has a cwlfilename set.")
```
dannon: Minor, but s/cwlfilename/basename if this is to be used outside the context of the cwl branch?

jmchilton: Of course, I rebased this with that fix. Thanks for the catch.

mvdbeek:
@jmchilton the migration is conflicting again, it should be 154 now

jmchilton:
It is 154 in here @mvdbeek right? There are some fixes for 0152 also but they are just fixing the existing migration.

mvdbeek:
Ha, just skimmed this, I'll review now.

mvdbeek (approved):

PR #7674, opened 2019-04-06
[19.01] Bugfix: when removing tags from a dataset, only the first tag can be removed

sveinugu:
When removing tags from a dataset, e.g. as part of a workflow, only the first tag in a list is currently possible to remove. The cause is that a space character is erroneously added in front of the other tags.

I have added a test that fails with the current code, and then fixed the issue. This might also fix other tag-related bugs.

New version of #7673, now based upon release_19.01. Had some issues with the other PR.

### DIFF
--- a/lib/galaxy/managers/tags.py
+++ b/lib/galaxy/managers/tags.py
@@ -198,7 +198,7 @@ class TagManager(object):
             if tag.value is not None:
                 tag_str += ":" + tag.user_value
             tags_str_list.append(tag_str)
-        return ", ".join(tags_str_list)
+        return ",".join(tags_str_list)
 
     def get_tag_by_id(self, tag_id):
         """Get a Tag object from a tag id."""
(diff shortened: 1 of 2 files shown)


sveinugu:
Not sure who should review. @dannon?

If it is a bug, it is old, but might not have caused any issues until recently, when get_tags_str() has been more used in places where it is assumed that the string is ","-joined, not ", "-joined. So the bugs may also be those assumptions. Removing space in get_tags_str() still seems the correct way to fix the issue, but might have consequences other places.

Also:
 - [ ] Should be fixed according to #7530 when merged to dev.

mvdbeek (approved):
This looks correct to me, many thanks for the fix and the test case.

mvdbeek:
@galaxybot test this

dannon:
I haven't dug into it yet, but this doesn't seem like quite the right fix to me.  I'll try to do some digging.  Thanks for the ping!

PR #7677, opened 2019-04-06
Drop .pyc files before attempting to migrate

mvdbeek:
One small thing that is kinda annoying to remember

### DIFF
--- a/manage_db.sh
+++ b/manage_db.sh
@@ -11,4 +11,5 @@ cd `dirname $0`
 
 setup_python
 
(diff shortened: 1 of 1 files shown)


manage_db.sh
```diff
@@ -11,4 +11,5 @@ cd `dirname $0`
 
 setup_python
 
+find lib/galaxy/model/migrate/versions -name *.pyc -delete
```
nsoranzo, line 14: `*.pyc` should be quoted to prevent globbing by the shell.

Issue #7661, opened 2019-04-04
Cut tool not working in new workflows


hepcat72:
We've got the latest version of your @bgruening's cut tool (1.1.0) and when I include it in a workflow, it fails with the error:

`Parameter list requires a value, but has no legal values defined.`

despite all required values being filled in.

I have not tried old workflows with this tool incorporated, so I can only attest to new workflows with the tool.  Here are some relevant screenshots:

The tool in the workflow editor:

<img width="228" alt="cutbox" src="https://user-images.githubusercontent.com/2300532/55564914-e4951680-56c6-11e9-84ac-2ca451725d6f.png">

It's set values in the workflow editor:

<img width="267" alt="cutentries" src="https://user-images.githubusercontent.com/2300532/55564943-ee1e7e80-56c6-11e9-8b6b-b4979a0a8530.png">

Its shown values in the form when running the workflow:

<img width="1196" alt="cut original should be 2 - untouched run" src="https://user-images.githubusercontent.com/2300532/55565024-14441e80-56c7-11e9-91ff-025eec602085.png">

The info on a failed output:

<img width="885" alt="cuterr" src="https://user-images.githubusercontent.com/2300532/55565070-2625c180-56c7-11e9-89a7-37f518aead3d.png">

Further, I went to view the tool in the admin settings and I get the following error:

`Error displaying tool, probably due to a problem in the tool config. The exception is: list indices must be integers, not str.`

This started happening after upgrading to 19.01.

I created a test workflow with the 2 cut tools we have.  The other succeeds:

https://galaxy.princeton.edu/u/rleach/w/cut-test

hepcat72:
I believe that 'Sort' (Galaxy Version 1.1.1) is affected by this same issue.  Sort says it completes successfully but it's output is 0 bytes and it has this in the stderr:

```
sort: invalid number at field start: invalid count at start of `None,Nonenr'
grep: write error
```

If you omit the "on column" field entry in the workflow or select "Nothing Selected", it produces output with no error.

hepcat72:
The command that's run when the sort tool is run via the workflow is:

```
python '/galaxy/galaxy-app/tools/filters/sorter.py'  --input='/galaxy/files/000/426/dataset_426213.dat' --output='/galaxy/files/000/426/dataset_426225.dat'    --key=,nr   --header 0
```

and apparently, not entering a column in the workflow does not work-around the issue.  That only works if you run the tool manually.

[...]

PR #7679, opened 2019-04-06
[19.01] Detect errors using exit_code for sort tool

mvdbeek:
Fixes another issue mentioned in #7661. Not bumping the version here since results don't change for valid inputs, but could do that if people think that'd better.

### DIFF
--- a/tools/filters/sorter.xml
+++ b/tools/filters/sorter.xml
@@ -5,40 +5,40 @@
         <requirement type="package" version="2.14">grep</requirement>
         <requirement type="package" version="4.4">sed</requirement>
     </requirements>
-    <command detect_errors="aggressive">
-        python '$__tool_directory__/sorter.py'
+    <command detect_errors="exit_code">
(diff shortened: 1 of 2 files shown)


jmchilton (approved):

tools/filters/sorter.xml
```diff
@@ -5,40 +5,40 @@
         <requirement type="package" version="2.14">grep</requirement>
         <requirement type="package" version="4.4">sed</requirement>
     </requirements>
-    <command detect_errors="aggressive">
-        python '$__tool_directory__/sorter.py'
+    <command detect_errors="exit_code">
```
nsoranzo, line 8: Just to clarify, I think this is not what fixes the tool (in fact [aggressive](https://docs.galaxyproject.org/en/latest/dev/schema.html#error-detection) also fails the job if the exit code is not zero), but the above changes from `subprocess.call()` to `subprocess.check_call()` do.

PR #7680, opened 2019-04-06
[19.01] Tags remove_tags_from_list fix update

dannon:
Followup to #7674.  As @sveinugu notes, there are other consumers of `get_tags_str`, so this fix has fewer side-effects.

Thank you again @sveinugu for the original report and fix!

### DIFF
--- a/lib/galaxy/managers/tags.py
+++ b/lib/galaxy/managers/tags.py
@@ -50,7 +50,7 @@ class TagManager(object):
 
     def remove_tags_from_list(self, user, item, tag_to_remove_list):
         tag_to_remove_set = set(tag_to_remove_list)
-        tags_set = set(self.get_tags_str(item.tags).split(','))
+        tags_set = set(self.get_tags_str(item.tags).split(', '))
         if item.tags:
             tags_set -= tag_to_remove_set
         return self.set_tags_from_list(user, item, tags_set)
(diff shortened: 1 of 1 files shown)


lib/galaxy/managers/tags.py
```diff
@@ -198,7 +198,7 @@ def get_tags_str(self, tags):
             if tag.value is not None:
                 tag_str += ":" + tag.user_value
             tags_str_list.append(tag_str)
-        return ",".join(tags_str_list)
+        return ", ".join(tags_str_list)
```
mvdbeek, line 201: Hmm, I think I don't agree with this. There are many instances where we call `get_tags_str ` followed by `split(',')`. If this is for display purposes the downstream code should probably handle nicer formatting or get the tags as a list ? It's true that

mvdbeek: many places in the code should probably just receive the tags as a list or set, but that wouldn't be the minimal fix anymore and is something we should do in dev.

dannon: @mvdbeek We shouldn't be composing a string and splitting it back apart multiple times in any event, this should be passed around as an array and formatted on display.  My goal here was to minimize potential side effects (for a stable PR), which this does.  It's all due for an overhaul though.

mvdbeek: sounds good

mvdbeek (approved):

lib/galaxy/managers/tags.py
sveinugu: Maybe a new issue might be in order for refactoring this in dev? As @mvdbeek writes, there are other consumers that splits on comma and potentially other bugs. Agree this fix should be minimal.

PR #7681, opened 2019-04-07
Quote *.pyc in the find/cleanup command for manage_db.sh

dannon:
xref https://github.com/galaxyproject/galaxy/pull/7677#discussion_r272813014

@nsoranzo Both versions worked fine in my testing, but I'm using zsh.  Which shells did the unquoted version break or potentially have problems with, out of curiosity?

### DIFF
--- a/manage_db.sh
+++ b/manage_db.sh
@@ -11,5 +11,5 @@ cd `dirname $0`
 
 setup_python
 
-find lib/galaxy/model/migrate/versions -name *.pyc -delete
+find lib/galaxy/model/migrate/versions -name '*.pyc' -delete
 python ./scripts/manage_db.py $@


nsoranzo:
@dannon Any shell, as long as you have a `.pyc` file in the galaxy root directory.

```console
[galaxy]$ touch foo.pyc
[galaxy]$ find lib/galaxy/model/migrate/versions -name *.pyc
[galaxy]$ find lib/galaxy/model/migrate/versions -name '*.pyc'
lib/galaxy/model/migrate/versions/0149_dynamic_tools.pyc
lib/galaxy/model/migrate/versions/0047_job_table_user_id_column.pyc
...
```

dannon:
@nsoranzo Ahh!  Ok, I thought this was some shell-specific expansion; makes sense that it'd expand in the root directory (where there probably shouldn't ever be *.pyc, hence me not seeing it), but you're definitely right that it's best to be explicit here.  Thanks for the info!

Issue #7622, opened 2019-03-29
Incorrect output format detected when using `format_source` for `fastqsanger.gz` dataset for the 1st run


bebatut:
Hi,

@wm75 and I identified a strange "bug" with output format detection using `format_source`

**Issue**: incorrect output format detected when using `format_source` for `fastqsanger.gz` (as fastqsanger) datasets run for the 1st run
**Usecase**: SortMeRNA on usegalaxy.eu but also using Planemo
**How to reproduce it**:
1. Auto-detect (same issue if `fastqsanger.gz` as format) 
	1. Upload a fastq.gz file with auto-detection of the format
	2. Run SortMeRNA with the uploaded file as input file (`(as fastqsanger)`)
	3. Check the generated file: `format: data`
    4. Re-run SortMeRNA
	5. Check the generated file: `format: fastqsanger` (what is expected)
2. Force `fastqsanger` format on upload
	1. Upload the same fastq.gz file by forcing `fastqsanger` format
	2. Run SortMeRNA with the uploaded file as input file (no `(as fastqsanger)` displayed)
	3. Check the generated file: `format: fastqsanger`

It seems that the format detection for the output using `format_source` of the input is made before the correct metadata of the input are assigned 
We echoed `ext` in the command section: it is always `fastqsanger`. So the output format detection is made before the command section.

Any idea how we could fix that?

Thanks a lot

Bérénice

mvdbeek:
> 3\. Check the generated file: `format: data`

So is that where it is going wrong ? This should have been fastqsanger ?

martenson:
is `sniff_compressed_dynamic_datatypes_default: true` on *.eu?

hexylena:
@martenson https://github.com/usegalaxy-eu/infrastructure-playbook/blob/master/group_vars/gxconfig.yml#L332

mvdbeek:
I can reproduce this, this is unrelated to any eventual sniffing. https://github.com/galaxyproject/galaxy/pull/7682 should fix this.

mvdbeek:
Alright, this should be fixed by https://github.com/galaxyproject/galaxy/pull/7682, it'll just need to be deployed on usegalaxy.*

PR #7682, opened 2019-04-07
[19.01] Fix format_source for implicit conversion inputs and fasta.gz upload

mvdbeek:
While writing the test I also noticed that fasta.gz uploads were broken, so this also fixes that.
I also noticed that implicit conversion doesn't happen for collection inputs (they work fine if mapped over), but since we don't have any reports for that and it'll probably be a bigger change I'll reserve that for dev

Fixes https://github.com/galaxyproject/galaxy/issues/7622


test/functional/tools/sample_datatypes_conf.xml
```diff
@@ -7,7 +7,7 @@
     <datatype extension="csv" type="galaxy.datatypes.tabular:CSV" display_in_upload="true" />
     <datatype extension="interval" type="galaxy.datatypes.interval:Interval" display_in_upload="true" description="File must start with definition line in the following format (columns may be in any order)." >
     </datatype>
-    <datatype extension="fasta" type="galaxy.datatypes.sequence:Fasta" auto_compressed_types="gz,bz2" display_in_upload="true">
```
martenson: Why is bz2 going away? And if so should we remove it for fastqsanger too?

martenson: oh it was restored later, nvm

lib/galaxy/model/mapping.py
```diff
@@ -1926,6 +1926,9 @@ def simple_mapping(model, **kwds):
     implicitly_converted_datasets=relation(model.ImplicitlyConvertedDatasetAssociation,
         primaryjoin=(model.ImplicitlyConvertedDatasetAssociation.table.c.ldda_parent_id ==
                      model.LibraryDatasetDatasetAssociation.table.c.id)),
+    implicitly_converted_parent_datasets=relation(model.ImplicitlyConvertedDatasetAssociation,
+        primaryjoin=(model.ImplicitlyConvertedDatasetAssociation.table.c.ldda_id ==
+                     model.LibraryDatasetDatasetAssociation.table.c.id)),
```
nsoranzo: Alternatively you could add a backref at line 1608-1610.

mvdbeek: Alright, done that and also replaced `implicitly_converted_parent_datasets` for the HDA relation in the same way.

jmchilton (approved):

martenson:
thanks @mvdbeek !

PR #7683, opened 2019-04-07
Implicit conversion for input collection parameters

mvdbeek:
The last 2 commits add implicit conversion support for input collections, while the first commits include #7682. This had never been implemented, so I think we don't need to target 19.01 for this (though it is a bug given the UI lets you do this, and if you try it'll fail ...).


mvdbeek:
This is ready again, rebased on top of current dev that contains https://github.com/galaxyproject/galaxy/pull/7682

PR #7689, opened 2019-04-08
Plugin static staging fix

dannon:
This fixes a bug where previously staged plugins (via symlink in the python version of this staging) would not be overwritten with the same real file contents.  This would cause 404s for plugin static under uwsgi serving due to the removal of `config/plugins/etc` from the static-safe map.

Includes minor gulpfile cleanup and dropping a few now-unused dependencies.


guerler:
Thanks @dannon

PR #7690, opened 2019-04-08
Fix side panel styling, adjust appearance in reports

guerler:
This PR fixes the appearance of the side panel in the reports app and performs a clean-up on the side/tool panel style definition overall.

### DIFF
--- a/client/galaxy/style/scss/theme/blue.scss
+++ b/client/galaxy/style/scss/theme/blue.scss
@@ -71,15 +71,15 @@ $font-family-sans-serif: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, 
     "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
 $font-family-monospace: Monaco, Menlo, Consolas, "Courier New", monospace;
 $font-family-base: $font-family-sans-serif;
-$font-size-base: 0.8rem;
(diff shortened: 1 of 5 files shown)


client/galaxy/style/scss/theme/blue.scss
```diff
@@ -71,15 +71,15 @@ $font-family-sans-serif: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto,
     "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
 $font-family-monospace: Monaco, Menlo, Consolas, "Courier New", monospace;
 $font-family-base: $font-family-sans-serif;
-$font-size-base: 0.8rem; // default 1rem
```
dannon: @guerler Can we leave these comments in (and the ones below), for future documentation?  It helps (at least me) remember we're downsizing from the bootstrap defaults, and by how much.

guerler: Sure.

hexylena (approved):

martenson:
I was under impression this aimed to bring the style on par with how tool panel looks like, but the change is supposed to look just like this, right?

Galaxy___Reports_and_Galaxy___Reports

edit: I think this is a reasonable fix, thanks a bunch @guerler!

PR #7691, opened 2019-04-08
Add docs to OIDC backends configuration file.

VJalili:
ping @martenson 

These are the docs from https://github.com/galaxyproject/galaxy/pull/7620


martenson:
@VJalili thanks!

PR #7696, opened 2019-04-09
Patch for metadata setting of minimal BIOM1 files

### LINKED ISSUE #7625: Error with BIOM upload and generation on 19.01
Hi, 

The upload and generation of BIOM file are raising errors on 19.01, not on 18.09.

**Issue**: 
```
ERROR:galaxy.datatypes.text:Something in the metadata detection for biom1 went wrong.
  File "/opt/galaxy/server/lib/galaxy/datatypes/text.py", line 259, in set_meta
    for k, v in column['metadata'].items():
AttributeError: 'NoneType' object has no attribute 'items'
```

**Usecase**: 
- Upload biom file on usegalaxy.eu but also using Planemo (no error but same info in stdout)
- MetaPhan2 on usegalaxy, but also using Planemo (error)

**How to reproduce it**:
- usegalaxy.eu (19.01)
	- Upload biom file: no error but same info in stdout
    - Run MetaPhan2: tool failing with the message
- 19.01 using planemo
    - Run MetaPhan2: tool failing with the message
- 18.09 using planemo
    - Run MetaPhan2: tool not failing

The piece of code pointed by the error has been modified in [6dd098f747c89ba](https://github.com/galaxyproject/galaxy/commit/6dd098f747c89ba3494b64bb5a64dc078bb3d5db#diff-29996c0d5f723a702ea671841bacefa3) on Nov 7, 2018. 
@nsoranzo do you think your changes may cause the error or it is a previous change (looks like there were several changes on the same portion in the days before)?
Do you have any idea how we can fix that?
I will investigate later otherwise

Thanks a lot

Bérénice

bebatut:
Hi,

As reported in #7625, the upload and generation of BIOM file are raising errors on 19.01, not on 18.09. This issue happens only with minimal BIOM files (not rich ones).

The minimal BIOM1 files may have `column = {u'id': u'sth', u'metadata': None}`, raising then the following error:

```
ERROR:galaxy.datatypes.text:Something in the metadata detection for biom1 went wrong.
  File "/opt/galaxy/server/lib/galaxy/datatypes/text.py", line 259, in set_meta
    for k, v in column['metadata'].items():
AttributeError: 'NoneType' object has no attribute 'items'
```

This PR checks if `column['metadata']` is not None before the for loop. It should fix #7625.

@nsoranzo @Slugger70 could you check this patch? Should I add some tests? If yes, how/where?

Thanks a lot

Bérénice

---

See [BIOM1 specs](http://biom-format.org/documentation/format_versions/biom-1.0.html) for examples of minimal and rich BIOM1 files

### DIFF
--- a/lib/galaxy/datatypes/text.py
+++ b/lib/galaxy/datatypes/text.py
@@ -280,9 +280,10 @@ class Biom1(Json):
                         if b_name == "columns" and metadata_value:
                             keep_columns = set()
                             for column in metadata_value:
-                                for k, v in column['metadata'].items():
-                                    if v is not None:
-                                        keep_columns.add(k)
+                                if column['metadata'] is not None:
+                                    for k, v in column['metadata'].items():
+                                        if v is not None:
+                                            keep_columns.add(k)
                             final_list = sorted(list(keep_columns))
                             dataset.metadata.table_column_metadata_headers = final_list
                         if b_name in b_transform:


lib/galaxy/datatypes/text.py
```diff
@@ -280,9 +280,10 @@ def _transform_dict_list_ids(dict_list):
                         if b_name == "columns" and metadata_value:
                             keep_columns = set()
                             for column in metadata_value:
-                                for k, v in column['metadata'].items():
-                                    if v is not None:
-                                        keep_columns.add(k)
+                                if column['metadata'] is not None:
+                                    for k, v in column['metadata'].items():
```
hexylena, line 284: you could also do `column.get('metadata', {}).items():`. I'm not sure, @nsoranzo you have opinions on these things :)

nsoranzo (approved):

lib/galaxy/datatypes/text.py
nsoranzo: Both are probably fine, @bebatut's version is a bit more strict, since it requires `'metadata'` to be a key in the dict.

bgruening:
Should that target 19.01?

bgruening (approved):

Slugger70 (approved):

nsoranzo:
Feel free to open a backport PR to 19.01 if that's worth.

vebaev:
Hi guys,

I'm using galaxy-docker 19.01 and I see the above error. My output from the qiime tools are red, but seems inside they are ok. Is there a fix or just wait for  19.05?
Unfortunately, when the output datasets are red I cannot continue with my analysis...

vebaev:
Hi guys,

I'm using galaxy-docker 19.01 and I see the above error. My output from the qiime tools are red, but seems inside they are ok. Is there a fix?
(Unfortunately when datasets are red I cannot continue with my analysis)

nsoranzo:
@vebaev This has been already backported to 19.01, maybe you need to update Galaxy to the latest commit of the `release_19.01` branch inside your docker container?

vebaev:
@nsoranzo Strange, I have pulled the latest 19.01 for my docker flavor build 5 days ago, but I got the error? (FROM bgruening/galaxy-stable:19.01) which was build before the fix probably?

PR #7697, opened 2019-04-09
Fix genomespace exporter tool

dannon:
Specify profile and look for exit codes for genomespace tool -- warnings cause failures otherwise.

Someone more familiar with the tool -- is this going to work like I expect, or is it going to potentially fail without returning a nonzero exit code?

And, we may need to apply this to the other genomespace tools using cloudbridge as well.

xref https://github.com/galaxyproject/galaxy/pull/7647#issuecomment-481208063


mvdbeek (approved):

mvdbeek:
Yes, that should fix it, it looks like `GenomeSpaceClient` just raises the proper exceptions, so the python interpreter will quit with a non-zero exit code.

mvdbeek (approved):

blankenberg:
Any volunteers to target this to 19.01 as well? (updated usegalaxy.org has this issue currently)

dannon:
@blankenberg Sure.

PR #7699, opened 2019-04-09
[19.01] Only check for exit code on actual sort command

mvdbeek:
The grep command that collects headers will exit with a non-zero exit code if there is no header, so we can't use `subprocess.check_call` there. You can see the test is actually failing here: https://jenkins.galaxyproject.org/view/All/job/docker-main-tools/218/testReport/

And you can manually test the tool with planemo:
```
All 4 test(s) executed passed.
sort1[0]: passed
sort1[1]: passed
sort1[2]: passed
sort1[3]: passed
```

### DIFF
--- a/tools/filters/sorter.py
+++ b/tools/filters/sorter.py
@@ -42,9 +42,7 @@ def main():
             # sed header
             if header_lines > 0:
                 sed_header = ['sed', '-n', "1,%dp" % header_lines, input]
-                exit_code = subprocess.call(sed_header, stdout=out)
-                if exit_code:
-                    stop_err('Capturing header lines failed')
+                subprocess.check_call(sed_header, stdout=out)
 
             # grep comments
             grep_comments = ['grep', '^#', input]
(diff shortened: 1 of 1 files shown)


jmchilton (approved):

tools/filters/sorter.py
```diff
@@ -45,7 +45,7 @@ def main():
 
         # grep comments
         grep_comments = "(grep '^#' %s) >> %s" % (input, output)
-        subprocess.check_call(grep_comments, shell=True)
+        subprocess.call(grep_comments, shell=True)
```
nsoranzo: You could capture the return value of `subprocess.call()` and call `stop_err()` if it is not 0 or 1.

Also, I'd suggest to:
- pass the 2 above commands as lists instead of strings, and remove `shell=True`
- replace the output redirects with the `stdout` parameter of `call`/`check_call`

bernt-matthias (approved):
looks good to me

nsoranzo (approved):

PR #7702, opened 2019-04-09
Relax the 1st line sniffing for mtx (not only real coordinate)

bebatut:
Hi, 

The sniffing of the 1st line of mtx file is currently too strict.
This PR should fix that

Bérénice


hexylena (approved):

jmchilton:
Thanks, looks good to me!

PR #7703, opened 2019-04-09
Update pytest-posgresql to 1.4.0, which removes a hard dependency on psycopg2

mvdbeek:


nsoranzo (approved):

blankenberg:
Fixes the issue for me, thanks!

blankenberg (approved):

PR #7708, opened 2019-04-09
Improve error message if GFF file is missing attribute col

nsoranzo:
Alternative to #7093.

### DIFF
--- a/lib/galaxy/datatypes/util/gff_util.py
+++ b/lib/galaxy/datatypes/util/gff_util.py
@@ -42,7 +42,7 @@ class GFFInterval(GenomicInterval):
         self.score = self.fields[self.score_col]
 
         # GFF attributes.
-        if 8 >= self.nfields:
+        if self.nfields < 9:
             raise MissingFieldError("No field for attribute column (8)")
         self.attributes = parse_gff_attributes(fields[8])
 
--- a/lib/galaxy/datatypes/util/gff_util.py
+++ b/lib/galaxy/datatypes/util/gff_util.py
@@ -42,6 +42,8 @@ class GFFInterval(GenomicInterval):
         self.score = self.fields[self.score_col]
 
         # GFF attributes.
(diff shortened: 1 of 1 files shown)


lib/galaxy/datatypes/util/gff_util.py
```diff
@@ -42,6 +42,8 @@ def __init__(self, reader, fields, chrom_col=0, feature_col=2, start_col=3, end_
         self.score = self.fields[self.score_col]
 
         # GFF attributes.
+        if 8 >= self.nfields:
```
mvdbeek:
```suggestion
        if self.nfields < 9:
```

lib/galaxy/datatypes/util/gff_util.py
```diff
@@ -42,6 +42,8 @@ def __init__(self, reader, fields, chrom_col=0, feature_col=2, start_col=3, end_
...
+            raise MissingFieldError("No field for attribute column (8)")
```
mvdbeek, line 46:
```suggestion
            raise MissingFieldError("No field for attribute column (9)")
```
It's the ninth column we're validating, right ?

lib/galaxy/datatypes/util/gff_util.py
nsoranzo: I was using the same coding style of the code above, but we could probably update all conditions as you suggested.

lib/galaxy/datatypes/util/gff_util.py
nsoranzo: Also in this case I'm using the 0-indexed column number as in the code above and as used also in bx-python. Not sure it's useful to fix only in this method.

mvdbeek: I think if it's a user-facing exception it'd be clearer to use 1-based counting, but either way is fine.

mvdbeek (approved):

Issue #7710, opened 2019-04-09
/api/workflows/{}/invocations not accessible


innovate-invent:
Same problem as https://github.com/galaxyproject/galaxy/issues/4401

I am not sure how your model queries work but check_security might be used as a filter here: [workflows.py#L217](https://github.com/galaxyproject/galaxy/blob/d05ee64e64601afaea631b435fad3a5fe1f2b932/lib/galaxy/managers/workflows.py#L217)

This won't work though as check_security throws an exception rather than return false.

mvdbeek:
Thanks for the report, should be fixed with https://github.com/galaxyproject/galaxy/pull/7711

PR #7711, opened 2019-04-10
[19.01] Restrict workflow invocation index to current user

mvdbeek:
If the user is not an admin.
Fixes https://github.com/galaxyproject/galaxy/issues/7710

An alternative would be to skip Accessibility / Ownership
exceptions, but given that we can just restrict the query
this seems like the better approach to me.


jmchilton (approved):

blankenberg:
+1

PR #7712, opened 2019-04-10
Add hostname to task message queues

### LINKED ISSUE #5601: server_name templating and mangling issues

mvdbeek:
This prevents processes with the same server_name on different
hosts to execute processes not destined for them.
This time I think it closes https://github.com/galaxyproject/galaxy/issues/5601

### DIFF
--- a/lib/galaxy/model/migrate/versions/0151_add_worker_process.py
+++ b/lib/galaxy/model/migrate/versions/0151_add_worker_process.py
@@ -8,9 +8,10 @@ import logging
 from sqlalchemy import (
     Column,
     DateTime,
+    Integer,
     MetaData,
     Table,
-    TEXT,
+    Text,
 )
 
 from galaxy.model.migrate.versions.util import create_table, drop_table
@@ -23,7 +24,9 @@ metadata = MetaData()
 WorkerProcess_table = Table(
     'worker_process',
     metadata,
-    Column('server_name', TEXT, primary_key=True),
+    Column("id", Integer, primary_key=True),
+    Column("server_name", Text, index=True),
+    Column("hostname", Text),
     Column("update_time", DateTime, default=now, onupdate=now),
 )
 
--- a/lib/galaxy/model/migrate/versions/0151_add_worker_process.py
+++ b/lib/galaxy/model/migrate/versions/0151_add_worker_process.py
@@ -12,6 +12,7 @@ from sqlalchemy import (
     MetaData,
     Table,
     Text,
+    UniqueConstraint,
 )
 
 from galaxy.model.migrate.versions.util import create_table, drop_table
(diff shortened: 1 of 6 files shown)


lib/galaxy/model/migrate/versions/0151_add_worker_process.py
```diff
@@ -23,7 +24,9 @@
 WorkerProcess_table = Table(
     'worker_process',
     metadata,
-    Column('server_name', TEXT, primary_key=True),
+    Column("id", Integer, primary_key=True),
+    Column("server_name", Text, index=True),
+    Column("hostname", Text),
     Column("update_time", DateTime, default=now, onupdate=now),
```
nsoranzo, line 31: I'd add:
```python
    UniqueConstraint('server_name', 'hostname'),
```

lib/galaxy/model/migrate/versions/0151_add_worker_process.py
```diff
@@ -23,7 +24,9 @@
 WorkerProcess_table = Table(
     'worker_process',
     metadata,
-    Column('server_name', TEXT, primary_key=True),
+    Column("id", Integer, primary_key=True),
+    Column("server_name", Text, index=True),
```
nsoranzo: Maybe `String(255)` is enough? Also for `hostname`.

mvdbeek:
Thanks, those are good suggestions, I added them.

natefoo (approved):

jmchilton (approved):

Issue #7714, opened 2019-04-10
Using separate uWSGI and Galaxy ini files with Python 3 raises TypeError from configparser


natefoo:
`config/uwsgi.ini`:
```ini
[uwsgi]
http = :8080
processes = 1
threads = 2
single-interpreter = true
pythonpath = lib
module = galaxy.webapps.galaxy.buildapp:uwsgi_app_factory()
set = galaxy_config_file=/path/to/galaxy/config/galaxy.ini
set = galaxy_root=/path/to/galaxy
static-map = /static/style=style/blue
static-map = /static=static
```

`config/galaxy.ini`:
```ini
[app:main]
```

run:
```console
$ . ./.venv/bin/activate
(.venv)$ python -V
Python 3.5.3
(.venv)$ uwsgi --ini config/uwsgi.ini
*** Starting uWSGI 2.0.18 (64bit) on [Wed Apr 10 09:52:36 2019] ***
compiled with version: 6.3.0 20170516 on 09 April 2019 19:31:42
os: Linux-4.15.0-0.bpo.2-amd64 #1 SMP Debian 4.15.11-1~bpo9+1 (2018-04-07)
nodename: example.org
machine: x86_64
clock source: unix
pcre jit disabled
detected number of CPU cores: 8
current working directory: /path/to/galaxy
detected binary path: /path/to/galaxy/.venv/bin/uwsgi
*** WARNING: you are running uWSGI without its master process manager ***
your processes number limit is 127816
your memory page size is 4096 bytes
detected max file descriptor number: 1024
building mime-types dictionary from file /etc/mime.types...554 entry found
lock engine: pthread robust mutexes
thunder lock: disabled (you can enable it with --thunder-lock)
uWSGI http bound on :8192 fd 4
spawned uWSGI http 1 (pid: 20244)
uwsgi socket 0 bound to TCP address 127.0.0.1:39171 (port auto-assigned) fd 3
Python version: 3.5.3 (default, Sep 27 2018, 17:25:39)  [GCC 6.3.0 20170516]
Python main interpreter initialized at 0x5557ddf0f650
python threads support enabled
your server socket listen backlog is limited to 100 connections
your mercy for graceful operations on workers is 60 seconds
mapped 166752 bytes (162 KB) for 4 cores
*** Operational MODE: preforking+threaded ***
added lib/ to pythonpath.
INFO:rdflib:RDFLib Version: 4.2.2
```
```pytb
Traceback (most recent call last):
[...]

PR #7715, opened 2019-04-10
Fix uwsgi startup with separate ini file under python 3

mvdbeek:
Fixes https://github.com/galaxyproject/galaxy/issues/7714


natefoo (approved):
Thanks!

mvdbeek:
@nsoranzo has a better fix for this.

PR #7720, opened 2019-04-10
OIDC login buttons based on configuration

machristie:
Instead of a hardcoded Google login button, display a login button for each OIDC backend based on the configured OIDC backends.

The approach of adding the OIDC backends to the Galaxy config object and then displaying a login button for each was taken from PR #7497. One difference is that I changed the value of the `oidc` field of the config object to be a list instead of a dictionary, since a list would preserve the order of the backends in oidc_backends_config.xml and give the admin control over the order of the login buttons.

### DIFF
--- a/client/galaxy/scripts/components/login/Login.vue
+++ b/client/galaxy/scripts/components/login/Login.vue
@@ -24,8 +24,8 @@
                         </b-card-footer>
                     </b-card>
                 </b-form>
-                <b-button v-if="enable_oidc" class="mt-3" @click="submitOIDCLogin()">
-                    <icon class="fa fa-google" /> Sign in with Google
+                <b-button v-for="idp in oidc_idps" :key="idp" class="d-block mt-3" @click="submitOIDCLogin(idp)">
+                    <icon v-bind:class="oidc_idps_icons[idp]" /> Sign in with {{ idp.charAt(0).toUpperCase() + idp.slice(1) }}
                 </b-button>
             </div>
             <div v-if="show_welcome_with_login" class="col">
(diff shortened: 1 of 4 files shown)


machristie:
I also removed the `encodeURI()` of the redirect_uri in commit https://github.com/galaxyproject/galaxy/pull/7720/commits/671f2b65cf69f5a5df2dc416b57b9f9000a718ee. The reason for this is that the Custos redirect_uri was getting double encoded, apparently because request-oauthlib (which is what is used for Custos) already encodes the query parameters. However, I tested and removing `encodeURI()` worked for the PSA generated Google redirect_uri.

Here's what I was seeing. With `encodeURI()` the redirect_uri is:
```
https://iam.scigap.org/auth/realms/galaxy-test/protocol/openid-connect/auth?response_type=code&client_id=local-galaxy&redirect_uri=http%253A%252F%252Flocalhost%253A8000%252Fauthnz%252Fcustos%252Fcallback...
```
You can see the double encoding in the redirect_uri query parameter: `redirect_uri=http%253A%252F%252Flocalhost%253A8000%252Fauthnz...`.

Without `encodeURI()` the redirect_uri is 
```
https://iam.scigap.org/auth/realms/galaxy-test/protocol/openid-connect/auth?response_type=code&client_id=local-galaxy&redirect_uri=http%3A%2F%2Flocalhost%3A8000%2Fauthnz%2Fcustos%2Fcallback...
```

machristie:
@VJalili Please review when you have a chance, thanks

VJalili:
@machristie LGTM, +1 for me, Thank you!

afgane:
Thanks again @machristie!

PR #7730, opened 2019-04-12
Fix assert_command is not executed

bernt-matthias:
`<assert_command>` seems not to work. I added a test that negates the conditions of the first test. So the second test should fail. 

Maybe a starting point for a fix? Any ideas? 

Additionally, can we express in `<test>` that the test (not the job) is expected to fail? Would be handy here.

### DIFF
--- a/lib/galaxy/tool_util/verify/interactor.py
+++ b/lib/galaxy/tool_util/verify/interactor.py
@@ -1064,7 +1064,7 @@ class ToolTestDescription(object):
         self.exception = processed_test_dict.get("exception", None)
 
         self.output_collections = map(TestCollectionOutputDef.from_dict, processed_test_dict.get("output_collections", []))
-        self.command_line = processed_test_dict.get("command", None)
+        self.command_line = processed_test_dict.get("command_line", None)
         self.stdout = processed_test_dict.get("stdout", None)
         self.stderr = processed_test_dict.get("stderr", None)
         self.expect_exit_code = processed_test_dict.get("expect_exit_code", None)
--- a/test/functional/tools/job_properties.xml
+++ b/test/functional/tools/job_properties.xml
@@ -32,11 +32,14 @@
         </collection>
     </outputs>
     <tests>
+        <!-- test for assert_command, -stdout, and stderr -->
         <test expect_exit_code="0">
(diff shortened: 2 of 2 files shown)


bernt-matthias:
One test seems to fail as expected. The other raises `TypeError: assert_has_text_matching() takes exactly 2 arguments (1 given)`

mvdbeek:
Yes, it should be 
```
diff --git a/test/functional/tools/job_properties.xml b/test/functional/tools/job_properties.xml
index 74cb66737c..2f1080579a 100644
--- a/test/functional/tools/job_properties.xml
+++ b/test/functional/tools/job_properties.xml
@@ -39,7 +39,7 @@
             <assert_command>
                 <has_text text="really" />
                 <not_has_text text="REALLY" />
-                <has_text_matching text="r.a.ly" />
+                <has_text_matching expression="r.a.ly" />
             </assert_command>
             <assert_stdout>
                 <has_line line="The bool is true" />
@@ -55,7 +55,7 @@
             <assert_command>
                 <not_has_text text="really" />
                 <has_text text="REALLY" />
-                <has_text_matching text="R.A.LY" />
+                <has_text_matching expression="R.A.LY" />
             </assert_command>
 <!--            <assert_stdout>-->
 <!--                <not_has_text text="The bool is true" />-->
```

I was tempted to just add `expect_failure="true"`, but that refers to the galaxy job itself, it'll complain with
```
>           raise JobOutputsError(found_exceptions, job_stdio)
E           galaxy.tools.verify.interactor.JobOutputsError: Expected job to fail but Galaxy indicated the job successfully completed.
E           Command produced by the job different than expected
E           Output file contains unexpected text 'really'
```
which does make sense. Maybe we can drop the test and just add the fix ?
This is a failure to parse the test, which we could unit test instead.

bernt-matthias:
OK reverted the first commit. Unit test would be nice, but is beyond my capabilities :(

mvdbeek:
Can you also rebase this against dev ? There's a conflict now.

bernt-matthias:
Done. But git history looks really ugly now.

mvdbeek:
Thanks, I'll keep the unit tests in mind. I guess we should backport this to 19.05

PR #7740, opened 2019-04-15
Python dependency upgrades

dannon:
This also swaps to docker-based dependency updating as the default Makefile target.

Sqlalchemy was intentionally held back at 1.2.18

### DIFF
--- a/lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
+++ b/lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
@@ -1,3 +1,5 @@
+-i https://wheels.galaxyproject.org/simple
+--extra-index-url https://pypi.python.org/simple
 alabaster==0.7.12
 argh==0.26.2
 atomicwrites==1.3.0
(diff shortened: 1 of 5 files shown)


lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
```diff
@@ -1,5 +1,3 @@
--i https://wheels.galaxyproject.org/simple
```
dannon: Hrmm, looks like these got dropped when I ran, simply `make update-dependencies` -- @nsoranzo is the expectation that these are manually added to the file, or has something changed since last time we ran this?

nsoranzo: That doesn't happen for me under Linux, or in Docker.

nsoranzo (approved):

nsoranzo:
@martenson @VJalili If you want to test this, you should now be able to correctly run `make update-dependencies` under macOS (provided you have Docker installed).

martenson (approved):
the docker update-dependencies runs well for me with this, thanks!

VJalili:
Thanks @nsoranzo & @dannon

PR #7745, opened 2019-04-15
Extend dataset search and add search API endpoint

mvdbeek:
The goal is to be able to search for datasets and dataset collections via the API, without knowing which history a dataset is in. This also extends the `HistoryContentsFilters` in a way that allows searching for datasets or collections with a given tag or annotation.

### DIFF
--- a/test/api/test_datasets.py
+++ b/test/api/test_datasets.py
@@ -86,7 +86,7 @@ class DatasetsApiTestCase(api.ApiTestCase):
         payload = {'limit': 10, 'offset': 0, 'q': ['history_content_type', 'tag-invalid_op'], 'qv': ['dataset', 'notag']}
         index_response = self._get("datasets", payload)
         self._assert_status_code_is(index_response, 400)
-        assert index_response.json()['err_msg'] == 'bad filter'
+        assert index_response.json()['err_msg'] == 'bad op in filter'
 
     def test_show(self):
         hda1 = self.dataset_populator.new_dataset(self.history_id)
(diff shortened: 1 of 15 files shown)


martenson:
Would this replace the current approach in view_multiple histories page?

mvdbeek:
I think it might, but I haven't yet looked at what the client-side code currently does. This should definitely be re-usable, and filtering by tag or datatype for instance should be easy to add to the current standard history view.

mvdbeek:
I think this works OK now, but it looks like the most interesting searches (in the history and multi-history views) are client-only searches ? The saved dataset search is still using controllers, so I'm not sure if this is worth investing energy in. Are we OK with having this exist as a API-only function at this point ? 
I'd really like to have this because I occasionally need to find some old dataset buried in a collection, and being able to search by tool id would save me a lot of time.

mvdbeek:
@galaxybot test py3

test/api/test_datasets.py
```diff
@@ -3,21 +3,93 @@
...
 
 from base import api
-from base.populators import TestsDatasets
+from base.populators import (
+    DatasetCollectionPopulator,
+    DatasetPopulator,
+)
 
 
-class DatasetsApiTestCase(api.ApiTestCase, TestsDatasets):
```
jmchilton, line 9: Awesome, been meaning to do that forever.

jmchilton:
It is absolutely fine IMO to have API-only functionality. The client-side multi-history search I think is a dead-end and I wouldn't worry too much about it.

martenson:
@jmchilton I agree with both of those statements. @mvdbeek thanks!

Issue #4707, opened 2017-09-27
Tool Panel - Advanced Data Selection


jmchilton:
We've discussed this at team meetings a couple different times but I can't find a formal issue corresponding to our discussion other than a stray comment on #4013 ("Redesign tool inputs with an "advanced mode" drop down that allows greater flexibility in selecting inputs."). Various parts of this are mentioned in other contexts but not the specific idea of a pop-up on the tool form corresponding to an advanced mode.

New mode should allow various selection types for data / data_collection parameters including:
- [x] Selecting datasets from collections - https://github.com/galaxyproject/galaxy/issues/740.
- [x] Selecting collections from collections - https://github.com/galaxyproject/galaxy/issues/4324
- [ ] Advanced filtering - matching query language available in the history panel, access to tags, etc....
- [ ] More configurable mapping / reduction specification - https://github.com/galaxyproject/galaxy/issues/4623.
- [x] Directly from libraries (https://github.com/galaxyproject/galaxy/issues/4667).
- [ ] Directly from new uploads (https://github.com/galaxyproject/galaxy/issues/4668).
- [X] ~Slicing collections - #4013~ (I've become convinced this isn't so much a tool form problem, explicit collection reorganization is better than more complicated tool form.). 
- [x] Creating collections on-the-fly - from library folders, filtering, expressions.
- [ ] Provide interface for linked vs unlinked multi-selection (e.g. cross product vs dot product) - the backend allows this specification. https://github.com/galaxyproject/galaxy/issues/5224
- [ ] GUI for letting users choose how mapping of inputs occurs https://github.com/galaxyproject/galaxy/issues/4379

martenson:
- selecting as collection from library

peterjc:
Cross reference https://github.com/galaxyproject/galaxy/issues/4667 for accessing files directly from libraries

jmchilton:
xref https://github.com/galaxyproject/galaxy/issues/4379 - I completely forgot that this is an important part of that.

martenson:
xref https://github.com/galaxyproject/galaxy/pull/7553

PR #7746, opened 2019-04-15
Enable the selection of library datasets in the tool form

### LINKED ISSUE #4667: Access DataLibrary files from tool form
I would like to submit some ideas that could help users (these ideas are based on comments from the users of our Galaxy server).
Discussions have started here : http://dev.list.galaxyproject.org/Displaying-tool-select-list-according-to-user-td4671049.html

It could be great to be able to access to a file stored in a library, without the the import to current history step

Like @peterjc has pointed I really don't know how it could be straightforward or a nightmare to implement.

As these are requests from my own users, I would be glad to help!

And fell free to make any change in this issue in order to make it clearer...

thank you.

guerler:
Follow-up for #7553. Fixes #4667 and #740. This PR activates the data dialog option, and allows users to select library datasets directly in the tool form. On job submission, the selected library dataset is added to the current history before the job is executed. Library dataset selection is only allowed for single dataset inputs.


martenson:
xref https://github.com/galaxyproject/galaxy/issues/4707

jmchilton (approved):

martenson:
Thanks a bunch @guerler !

FredericBGA:
I will try this, thank you @guerler

PR #7748, opened 2019-04-15
Replace ``enable_beta_export_format2_default`` with ``default_workflow_export_format``

nsoranzo:
This allows future different formats to be included.

Also remove promises about next release.

Follow-up on https://github.com/galaxyproject/galaxy/pull/7659 .


jmchilton:
Thanks a bunch, this is a lot better!

PR #7750, opened 2019-04-15
[19.05] Workflow editor accessibility fixes

dannon:
Simply adds a screen-reader only reference from workflow node to details pane to aid in navigation there.


mvdbeek:
I think the force push may have gone wrong here ?

dannon:
@mvdbeek Ahh, yeah, that's what I get for not checking the PR after.  I rebased to resolve a conflict and pushed, but I rebased against -dev and not 19.05.  Thanks!

mvdbeek (approved):
NIce! Didn't know about sr-only.

PR #7751, opened 2019-04-15
Sanitize html in messages using galaxy.util.sanitize_html 

dannon:
... instead of bleach directly.

`galaxy.util.sanitize_html` in turn calls bleach with our 'standard' configuration

Fixes stuff like: 
image


nsoranzo:
Do we need to backport this to 18.09 and 19.01 ?

PR #7754, opened 2019-04-16
Fix Kubernetes integration tests for Docker on Mac.

jmchilton:


test/integration/test_kubernetes_runner.py
```diff
@@ -125,7 +125,8 @@ def setUp(self):
 
     @classmethod
     def setUpClass(cls):
-        cls.jobs_directory = tempfile.mkdtemp()
+        # realpath for docker deployed in a VM on Mac, also done in driver_util.
+        cls.jobs_directory = os.path.realpath(tempfile.mkdtemp())
```
mvdbeek, line 129: I don't think this works on a default setup of docker for mac, at least not for me. 
tempfile.mkdtemp() creates folders in  '/var/folders/df/6xqpqpcd7h73b6jpx9t6cwhw0000gn/T/tmp5e8fm88k', which is not in the default mount paths.
Screen Shot 2019-04-16 at 14 52 26
Adding `/var/folders` seems to have no effect, probably due to the way /private overlaps with `/var/folders`.
None of the docker based tests actually work for me out of the box, I always need to prefix test commands with something like `TMPDIR=~/temp`. That said this doesn't break anything, so it'd be OK for me.

mvdbeek: The actual error with this PR and without the workaround is:
```
galaxy.jobs.runners ERROR 2019-04-16 14:59:13,998 (2/galaxy-gx-short-id-2) Job output not returned from cluster: [Errno 2] No such file or directory: '/private/var/folders/df/6xqpqpcd7h73b6jpx9t6cwhw0000gn/T/tmpMZVTFk/000/2/galaxy_2.o'
```

jmchilton: I don't know.

```python
% python
Python 2.7.13 (default, Jan 31 2018, 22:17:47)
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> import tempfile
>>> tempfile.mkdtemp()
'/var/folders/78/zxz5mz4d0jn53xf0l06j7ppc0000gp/T/tmpkszpsV'
>>> os.path.realpath(tempfile.mkdtemp())
'/private/var/folders/78/zxz5mz4d0jn53xf0l06j7ppc0000gp/T/tmpRs3O7Z'
```

For me the realpath causes the use of ``private`` in the integration tests and this causes things to work since that is a path known about inside the Docker VM and mounted.

mvdbeek: Well, I agree it should work, it also works if I mount things in manually via `docker run -v $(realpath $TMPDIR):$(realpath $TMPDIR)` ... just not in the test setup. But it's completely possible something is weird in my setup.

mvdbeek: OK, mystery solved, I had some old jobs with the old config lingering, `kubectl delete jobs --all` makes the tests pass.

mvdbeek (approved):

test/integration/test_kubernetes_runner.py
jmchilton: This happened to me a few times also, I wonder if we should generate a new job prefix and persistence claim name for each test run.

jmchilton:
Thanks for testing this @mvdbeek !

test/integration/test_kubernetes_runner.py
mvdbeek: I'm a bit hesitant creating a new prefix each time, if these are not cleaned up properly they can consume quite some resources. The cleanup routine though should always run, but I don't think it does now on `<ctrl-c>`, which is how got into that state.

mvdbeek:
I don't think the integration test failure is relevant (looks like a quay.io problem) and the remaining tests are not relevant.

PR #7755, opened 2019-04-16
Report a couple common errors in dataset/job report.

jmchilton:
Report if empty inputs were used or if the same input was used more than once, these were identified as common sources of potential errors across many tools. Neither of these are definite problems, but both could be depending on the tool and both happen frequently. Outline a few more ideas as TODOs in the code.

These were identified as action items for easing Galaxy Help/Support at the Galaxy team meeting in April of 2019.

The first commit here also updates the report for more fine grain standard error reporting and job message handling added in https://github.com/galaxyproject/galaxy/pull/7095/files (regardless of the other slightly more controversial commit, this commit should be in 19.05).

Screen Shot 2019-04-16 at 10 24 46 AM

### DIFF
--- a/client/galaxy/scripts/mvc/dataset/dataset-error.js
+++ b/client/galaxy/scripts/mvc/dataset/dataset-error.js
@@ -32,6 +32,7 @@ var View = Backbone.View.extend({
                     url: job_url,
                     success: job => {
                         this.render_error_page(dataset, job);
+                        this.find_common_problems(job);
                     },
                     error: response => {
                         var error_response = {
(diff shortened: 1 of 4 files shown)


martenson:
Cool stuff, ping @jennaj.

jennaj:
Very very useful!! Thank you both so much!

@jmchilton I would suggest updating the reported text for these two. Gives a bit more actionable advice, fixes duplicated wording and reads well with "google translate" (tested French + Chinese, we could check others). The messages are using the passive voice (could add "by gremlins" ;), but I think that is appropriate for our use case -- as these are suggested, but not confirmed, remedies.

Original1:

The tool was executed with one or more empty input datasets, this isn't
always a problem and may not have been but frequently this is a source
of error.

Change1:

The tool was executed with one or more empty input datasets. This frequently 
results in tool errors due to problematic input choices.

Original2: 

The tool was executed with one or more duplicate input datasets, this isn't
always a problem and may not have been but frequently this is a source
of error.

Change2:

The tool was executed with one or more duplicate input datasets. This frequently
results in tool errors due to problematic input choices.

*Updated to make it even shorter/clearer, plus works for people using workflows (not directly using "tool forms").*

PR #7770, opened 2019-04-18
Create repository clone path if it doesn't exist.

davebx:
Previously, if an admin deleted empty directories under the `tool_path/<toolshed>/repos/<owner>` path after uninstalling repositories and attempted to reinstall one, the installation would fail. This checks for and recreates the path before cloning a repository, just in case that happened.

### DIFF
--- a/lib/tool_shed/util/hg_util.py
+++ b/lib/tool_shed/util/hg_util.py
@@ -44,6 +44,9 @@ def clone_repository(repository_clone_url, repository_file_dir, ctx_rev=None):
     if ctx_rev:
         cmd.extend(['-r', ctx_rev])
     cmd.extend([repository_clone_url, repository_file_dir])
+    # Make sure the destination path actually exists before attempting to clone
+    if not os.path.exists(repository_file_dir):
+        os.makedirs(repository_file_dir)
     try:
         subprocess.check_output(cmd, stderr=subprocess.STDOUT)
         return True, None


lib/tool_shed/util/hg_util.py
```diff
@@ -44,6 +44,9 @@ def clone_repository(repository_clone_url, repository_file_dir, ctx_rev=None):
     if ctx_rev:
         cmd.extend(['-r', ctx_rev])
     cmd.extend([repository_clone_url, repository_file_dir])
+    # Make sure the destination path actually exists before attempting to clone
+    if not os.path.exists(repository_file_dir):
```
mvdbeek, line 48: Hmm, if I try this with my system installed mercurial it works fine:
```
hg clone https://bitbucket.org/brmtron/mathlibcythontest/ /tmp/blurgas/xscsa
```
If older versions don't support this I guess it doesn't harm to do it your way, but shouldn't we only create the parent dir ?

davebx: This is only a problem if an admin deleted empty directories under `tool_path` after uninstalling repositories. At least, that's how I ran into this issue.

mvdbeek: Thanks, got it.

mvdbeek (approved):

mvdbeek:
I'm still puzzled by why this is necessary, but I guess there's no harm either.

davebx:
As am I, because of what you said on the diff. But with this change, test.galaxyproject let me install the broken repositories, so...

PR #7772, opened 2019-04-19
Fix raise syntax error

mvdbeek:
I'm not sure if this is really an error, but my IDE tells me this
is invalid.


dannon:
@mvdbeek Interesting, what IDE/linting?  AFAIK it's valid (unless this is another 3.x python change), and that 'raise' without arguments just uses the last exception.  That said, since we do have an explicit handle to the exception, it certainly doesn't hurt to just use it here.

mvdbeek:
It's pycharm, which is usually pretty accurate with these things. Not sure what the underlying library is. It might just not be able to inspect the calling context. It doesn't complain about other places in the code where the exception handling is not in its own function.

dannon:
@mvdbeek Yeah that makes sense -- pycharm just isn't sure that this function is always called from the except block with appropriate context, so it errs on the side of caution.

PR #7773, opened 2019-04-19
Add GeoJSON datatype

bgruening:
Needed by our ecology community.


bgruening:
Mh, does the failing test mean that Galaxy is modifying the uploaded file?

martenson:
@bgruening hmm, yes, the temp file has different md5 sum :/

lib/galaxy/datatypes/test/gis.geojson
```diff
@@ -0,0 +1,44 @@
...
+          ]
+        ]
+      },
+      "properties": {
+        "prop0": "value0",
+        "prop1": { "this": "that" }
+      }
+    }
+  ]
+}
```
nsoranzo: It may be the missing end of line here.

martenson: correct, that is it

bgruening:
Merci!

PR #7786, opened 2019-04-19
[19.01] Fix the use of JobMappingException

natefoo:
Currently, if a dynamic rule raises `JobMappingException`, this causes the readiness loop to call the job wrapper's `fail()` method. In `fail()`, a call to `self.get_destination_configuration()` accesses `self.job_destination` which raises `JobMappingException` again, preventing the `fail()` method from completing. The exception is caught further up and logged, but the job state remains `new` and the process repeats with each loop of the job readiness check.


jmchilton:
Did you figure out what broke this, was it me?

jmchilton (approved):

natefoo:
I didn't look. It looks like it was probably 183d84e5dc8ce4530835852292f3d20318dc9980. Maybe instead of this we should revert that change?

natefoo:
@nsoranzo I'd be interested in your opinion on this one.

PR #7790, opened 2019-04-20
Vue rewrite of admin jobs page

scrathat:
This is a rewrite of the admin jobs page with vue.


bgruening:
Thanks @scrathat! There is a linting issue:

```
/home/circleci/repo/client/galaxy/scripts/components/admin/Jobs.vue
  136:60  error  'data' is defined but never used  vue/no-unused-vars
```

client/galaxy/scripts/components/admin/Jobs.vue
```diff
@@ -0,0 +1,371 @@
...
+.code {
+    background: black;
+    color: white;
+    padding: 1em;
+}
+.break-word {
+    white-space: pre-wrap;
+    word-break: break-word;
+}
+</style>
```
guerler, line 370: Can we remove the custom styles here, particularly those which specify colors and spacing? There are some alternative code-related style tags available: https://getbootstrap.com/docs/4.0/content/code/, maybe also https://getbootstrap.com/docs/4.0/utilities/colors/.

client/galaxy/scripts/components/admin/Jobs.vue
```diff
@@ -0,0 +1,371 @@
+<template>
+    <div>
+        <h2 class="mb-3">
+            <span id="jobs-title">Jobs</span>
+        </h2>
+        <Alert :message="message" :variant="status" />
```
guerler: Should this be `b-alert`?

client/galaxy/scripts/components/admin/Jobs.vue
```diff
@@ -0,0 +1,371 @@
...
+                Unfinished and recently finished jobs are displayed on this page. The 'cutoff' input box will do two
+                things -- it will limit the display of unfinished jobs to only those jobs that have not had their job
+                state updated recently, and it will limit the recently finished jobs list to only displaying jobs that
+                have finished since the cutoff.
+            </p>
+            If any jobs are displayed, you may choose to stop them. Your stop message will be displayed to the user as:
+            "This job was stopped by an administrator: <strong>&lt;YOUR MESSAGE&gt;</strong> For more information or
+            help, report this error".
+        </b-alert>
+        <Alert v-if="loading" message="Waiting for data" variant="info" />
```
guerler: Same here.

client/galaxy/scripts/components/admin/Jobs.vue
```diff
@@ -0,0 +1,371 @@
...
+                            {{ showCommandLine ? "Hide" : "Show" }} Command Line
+                        </b-button>
+                    </b-col>
+                    <b-col>
+                        <b-card
+                            v-if="jobsItemsComputed.length"
+                            header="Stop Selected Jobs"
+                            header-bg-variant="info"
+                            header-text-variant="white"
+                            border-variant="info"
```
guerler: Can we reduce the customization here?

client/galaxy/scripts/components/admin/Jobs.vue
guerler: I think we should avoid using `Alert` if we can use `b-alert` instead. It might be easier to adjust the response in the python-controller.

guerler:
Thank you @scrathat for working on this! It looks great.

lib/galaxy/webapps/galaxy/controllers/admin.py
```diff
@@ -1600,14 +1600,25 @@ def manage_roles_and_groups_for_user(self, trans, payload=None, **kwd):
...
         deleted = []
-        msg = None
-        status = None
+        message = kwd.get('message', '')
+        status = kwd.get('status', 'info')
         job_ids = util.listify(stop)
         if job_ids and stop_msg in [None, '']:
-            msg = 'Please enter an error message to display to the user describing why the job was terminated'
+            message = 'Please enter an error message to display to the user describing why the job was terminated'
             status = 'error'
```
guerler: Alternatively you could use `return self.message_exception(trans, message)` and then catch the error message on the client in the exception of the `axios` request. This is how the api triggers error messages.

bgruening:
@guerler green! :)

guerler:
@scrathat thanks again for working on this. I noticed that when I lock jobs and then reload the page, the checkbox field does not remain checked. I think it would be better to remove the "Job dispatching is currently unlocked." label and the "Update" button and just show the checkbox field in the correct state. The checkbox should always reflect the actual lock state and change the lock state upon checking/unchecking the field.

guerler:
This looks great. Thank you so much. There might be a few minor things which we could address in a follow-up. The label-for="cutoff-seconds" should probably be just “cutoff”. When selecting a job and then clicking the back-button the admin side panel is duplicated. This is related to targeting galaxy_main when accessing the job_info. I wonder if we need “galaxyKwdToBootstrap” here? Same goes for nodeName != “A”?

scrathat:
`galaxyKwdToBootstrap` comes from the Alert component.
I think it's nice to have one b-alert to display whatever message and status the backend returns. Maybe it should be moved to another file so it doesn't get duplicated.
`e.target.nodeName != "A"` checks for a click on the job id link (info_url) so the row details don't also show when clicking on the job id.

jmchilton:
I think this was merged without the Selenium tests running and I think it broke them, I'll try to find a fix.

martenson:
none of the jenkins tests were triggered

PR #7791, opened 2019-04-22
[WIP] Add new job states for better feedback to users

natefoo:
This adds 7 new job and dataset states:

- **waiting**: The dataset has been picked up by a handler (only used by the `db-skip-locked` and `db-transaction-isolation` handler assignment methods) but its input datasets are not ready yet.
- **limited**: The job is ready to run but is waiting due to concurrency limits.
- **dispatched**: The job has been dispatched to a runner plugin.
- **submitted**: The job has been submitted to the cluster (or Pulsar) but is not yet queued.
- **stagein**: The job's inputs are staging in (Pulsar runner only).
- **stagein**: The job's outputs are staging out (Pulsar runner only).
- **finishing**: Job execution is complete and the job finish method is running.

The meaning of the **new** state has changed: If using the old handler assignment methods, it retains two of its old meanings: either it hasn't been picked up by a handler or its inputs are not ready. If using one of the new methods, it only means that it hasn't been picked up by a handler. In all assignment configs, limited jobs are now assigned their own state (before they were left in **new**).

With this change, the **queued** state is *only* set once the job has been reported as queued by the DRM.

In a healthy Galaxy you should rarely see some of these states: **new** (unless using the old assignment methods) indicates handler job "grabber" problems, and **dispatched** and **submitted** indicate handler runner plugin or DRM problems (**submitted** only if the destination is not Pulsar). This should help with using e.g. Grafana to detect problems with Galaxy handlers without having to run your own test jobs.

I've also made the messages more informative, but am very open to ideas on better language for these. I want it to be clear to the user what they should do (mostly: wait).

Thanks for the client work, @dannon!

TODO (maybe in another PR):

- [x] Add **finishing** state
- [ ] Make Pulsar differentiate between submitted and queued
- [x] Handle submitted vs. queued states in the Pulsar runner
- [x] Add **staging-in** and **staging-out** states for Pulsar


natefoo:
Moved to WIP - I forgot I wanted to add a **finishing** state.

natefoo:
@dannon I added a few more states. Currently **stagein** uses the same style as **submitted**, and both **stageout** and **finishing** use the same style as **running**. I am kinda ambivalent on staging, maybe we could give them a file-transfery icon, or maybe they are fine as-is. For **finishing** can we increase the speed of the spinner maybe?

jmchilton:
If we're just adopting Pulsar states outright in Galaxy - I'm not sure the benefit of renaming them. I trust you though @natefoo 👍

natefoo:
@jmchilton I did consider it...

natefoo:
@dannon thanks!

natefoo:
The latest API failure wasn't failing until my most recent commit and doesn't fail for me locally. Few of these tests I've been fixing have ever failed for me locally. So I'm not sure if I broke something or not.

Also I guess I broke the scratchbook selenium test but I can't figure out how? This one does at least fail for me locally but looking at the error files I can't see why the scratchbook modal isn't appearing. The dataset is green and the eye is clicked.

mvdbeek:
I've triggered a re-run for the failed API test. I haven't seen this test fail in a while, though this is a tricky one. It runs a workflow, and then runs it again with the job cache turned on. It tests that a equivalent job is found and no new jobs will actually be run, and that's the part that failed.

You may need to add all new non-error states to https://github.com/galaxyproject/galaxy/blob/fe2c14d3299c863e8d623a23f972944eebb93210/lib/galaxy/managers/jobs.py#L112 for this test to pass reliably

mvdbeek:
... which you already did, but I think we also need a couple more, basically all that are not error or deleted

lib/galaxy/managers/jobs.py
```diff
@@ -109,7 +109,9 @@ def replace_dataset_ids(path, key, value):
             conditions.append(
                 model.Job.state.in_([model.Job.states.NEW,
                                      model.Job.states.QUEUED,
+                                     model.Job.states.NOT_READY,
```
mvdbeek, line 139: I think
```
~model.Jobs.state.in_([model.Job.state.ERROR, model.Job.state.FAILED])
```
should work and fix the API test.

mvdbeek: I've opened a PR against your branch in https://github.com/natefoo/galaxy/pull/11 that implements this. That should make the test more robust.

natefoo: Do we also need `model.Job.states.DELETED` if we are looking to exclude terminal jobs? We could also just use `model.Job.terminal_states` which has the value `[states.OK, states.ERROR, states.DELETED]` or `model.Job.non_ready_states` which includes all the others (except `model.Job.states.DELETED_NEW`, which maybe it should).

In fact, what about:

```python
class Job(...):
    states = Bunch(...)  # same as it is now
    terminal_states = set([states.OK, states.ERROR, states.DELETED, states.FAILED])
    non_terminal_states = set(states.values()).difference(terminal_states)
```

And maybe a few more sets as class attributes describing other conditions such as "pre-running" (everything that used to be queued), "running" (running and later but non-terminal), etc? If everything used these sets instead of building their own we should run in to fewer problems in the future?

mvdbeek: Didn't see that we also had `model.Job.states.DELETED`, yes, that should also go in there.
terminal or non_terminal would both not be correct, we want everything that hasn't failed yet (or is OK).

mvdbeek: so something like `potentially_ok_states = self.non_terminal_states.union({states.OK})`

natefoo: Ok, I'll work on it.

qiagu:
Is it possible to make APIs for tool wrappers to insert progress bar just for the running state?

martenson:
@qiagu can you elaborate? (maybe in a different issue)

qiagu:
I just opened a new ticket.

natefoo:
#7822 is the new issue for @qiagu's question, just for future reference.

dannon:
@natefoo Fix one selenium test error and another one pops up.  I'll take a look at that one now.

natefoo:
A thought occurred to me that we should probably consider before adding all these new states: Should we really be adding these states as both dataset and job states, or should we have essentially one "non-terminal" dataset state (or keep "new", "queued", and "running" just for backwards compatibility) and add the the job state to the history and history update APIs? The client could then use the job state instead of the dataset state (or at least use the job state while the dataset state is non-terminal). Using the job state here directly is more correct than simply passing the job state down as the dataset state all the time. It might also cause less breakage in the future for things that watch the dataset state from an expected list of states (as most of these bugs have been) - only the client really cares about these new states, since they are mainly for user feedback.

I vaguely recall a conversation that maybe putting job states in the history and history update APIs might be too expensive. So maybe correctness isn't worth it.

Thoughts?

dannon:
Regarding 'one non-terminal' dataset state -- these additional states may or may not be the right vector for conveying it, but the additional info we're presenting to the user at least in terms of the state 'blurb' -- `this dataset has gone to the upside down` or whatever is definitely a plus, here.  And I don't know of a better way off the top of my head to get that to the user.

natefoo:
I might’ve been confusing so just in case: we’d sill have all these fancy
new states, we just would only set them on the Job, not the Job and Dataset
as we do now. This would require a fair amount of client and API changes to
allow the client to use the job state.

Sticking with Dataset state is certainly the easy path since that’s how we
already do it. Just wanted to get some opinions before I started polluting
everyone’s databases.

dannon:
@natefoo Ahh!  Yeah, I misunderstood what you meant; I do like the idea of just using the job state directly.  I don't know if it's going to be best to add the job state directly to the history/contents API or simply progressively load the details via a job_id handle on the dataset that we then use to query the jobs API, but that's something we could experiment with to see how expensive it is.

dannon:
The selenium test that's failing is completely unrelated to the changes here, and is also failing in a separate PR (https://github.com/galaxyproject/galaxy/pull/7749).  The one related selenium error was fixed in 59e4371. I don't know why it's failing yet, and it doesn't fail locally, but it shouldn't prevent merging this.

That said, @natefoo, think we want to bump this to 19.09 given the conversation above?

natefoo:
Yep, definitely want to bump this. Thanks!

lib/galaxy/jobs/__init__.py
```diff
@@ -1177,10 +1177,10 @@ def get_destination_configuration(self, key, default=None):
             dest_params, self.app.config, key, default
         )
 
-    def enqueue(self):
+    def mark_as_dispatched(self):
```
nsoranzo, line 1387: I think this should be changed also in the comment on line 917.

lib/galaxy/jobs/runners/__init__.py
```diff
@@ -772,3 +774,26 @@ def mark_as_finished(self, job_state):
 
     def mark_as_failed(self, job_state):
         self.work_queue.put((self.fail_job, job_state))
+
+    def _recover_async_job_state(self, job, job_wrapper, cls=AsynchronousJobState):
+        """Recovers jobs still in the submitted/queued/running state when Galaxy started"""
```
nsoranzo:
```suggestion
        """Recovers job still in the submitted/queued/running state when Galaxy started"""
```

lib/galaxy/jobs/runners/chronos.py
```diff
@@ -151,34 +152,21 @@ def stop_job(self, job_wrapper):
...
-        elif job.state == model.Job.states.QUEUED:
-            LOGGER.debug(msg.format(
-                name=job.id, runner=job.job_runner_external_id,
-                state='queued'))
-            ajs.old_state = model.Job.states.QUEUED
-            ajs.running = False
+        ajs = self._recover_async_job_state(job, job_wrapper)
+        if getattr(ajs, 'fail_job', False):
+            log.error("(%s) Failing job due to job state recovery error",
+                      job.id, job.state)
```
nsoranzo, line 165: `job.state` is not used in the error message (also other 6 times below).

lib/galaxy/model/__init__.py
```diff
@@ -1967,8 +1991,15 @@ def __init__(self, id, **kwargs):
...
                    UPLOAD='upload',
+                   WAITING='waiting',
+                   LIMITED='limited',
+                   DISPATCHED='dispatched',
+                   SUBMITTED='submitted',
+                   STAGEIN='stagein',
                    QUEUED='queued',
                    RUNNING='running',
+                   STAGEOUT='stageout',
+                   FINISHING='finishing',
```
nsoranzo, line 2189: I think all these new states should be added to the `non_ready_states` tuple below.

lib/galaxy/tools/actions/history_imp_exp.py
```diff
@@ -30,7 +30,7 @@ def execute(self, tool, trans, incoming={}, set_output_hid=False, overwrite=True
         job.tool_id = tool.id
         job.user_id = trans.user.id
         start_job_state = job.state  # should be job.states.NEW
-        job.state = job.states.WAITING  # we need to set job state to something other than NEW, or else when tracking jobs in db it will be picked up before we have added input / output parameters
+        job.state = job.states.NOT_READY    # we need to set job state to something other than NEW, or else when tracking jobs in db it will be picked up before we have added input / output parameters
```
nsoranzo, line 34:
```suggestion
        job.state = job.states.NOT_READY  # we need to set job state to something other than NEW, or else when tracking jobs in db it will be picked up before we have added input / output parameters
```

lib/galaxy/tools/actions/history_imp_exp.py
```diff
@@ -99,7 +99,7 @@ def execute(self, tool, trans, incoming={}, set_output_hid=False, overwrite=True
             # If this is an actual user, run the job as that individual.  Otherwise we're running as guest.
             job.user_id = trans.user.id
         start_job_state = job.state  # should be job.states.NEW
-        job.state = job.states.WAITING  # we need to set job state to something other than NEW, or else when tracking jobs in db it will be picked up before we have added input / output parameters
+        job.state = job.states.NOT_READY    # we need to set job state to something other than NEW, or else when tracking jobs in db it will be picked up before we have added input / output parameters
```
nsoranzo, line 104:
```suggestion
        job.state = job.states.NOT_READY  # we need to set job state to something other than NEW, or else when tracking jobs in db it will be picked up before we have added input / output parameters
```

lib/galaxy/tools/actions/metadata.py
```diff
@@ -66,7 +66,7 @@ def execute_via_app(self, tool, app, session_id, history_id, user=None,
         except AttributeError:
             job.tool_version = "1.0.1"
         job.dynamic_tool = tool.dynamic_tool
-        job.state = job.states.WAITING  # we need to set job state to something other than NEW, or else when tracking jobs in db it will be picked up before we have added input / output parameters
+        job.state = job.states.NOT_READY    # we need to set job state to something other than NEW, or else when tracking jobs in db it will be picked up before we have added input / output parameters
```
nsoranzo, line 75:
```suggestion
        job.state = job.states.NOT_READY  # we need to set job state to something other than NEW, or else when tracking jobs in db it will be picked up before we have added input / output parameters
```

jmchilton (dismissed):

lib/galaxy/jobs/runners/__init__.py
```diff
@@ -728,6 +728,8 @@ def finish_job(self, job_state):
         """
         galaxy_id_tag = job_state.job_wrapper.get_id_tag()
         external_job_id = job_state.job_id
+        log.debug("(%s/%s) Job execution complete, performing job finish tasks", galaxy_id_tag, external_job_id)
+        job_state.job_wrapper.change_state(model.Job.states.FINISHING)
```
jmchilton, line 732: Ugh - this line is going to be a big performance hit for relatively little gain from the user perspective 😟.

jmchilton:
Ugh... I'm sorry to throw you under the bus but ``change_state`` is so expensive and this adds new ones for states that will only exist for a very brief period of time. My intuition is this will significantly reduce Galaxy's already crummy throughput for large numbers of jobs (e.g. collection operations, batch workflows, etc..). I'm -0 on this PR in this form - if we could refactor ``change_state`` into some sort of compound SQL statement or eliminate 90% of state tracking on datasets and offload them to job state tracking (or ideally both) I'd be much more eager to see this in.

If you want me to produce some throughput data to justify my position I will, but I'd rather spend the time coming up with a more optimal solution. Any idea how to implement delegation of dataset states to job states?

nsoranzo:
> Any idea how to implement delegation of dataset states to job states?

An SQL trigger maybe?

jmchilton:
> An SQL trigger maybe?

Better than a Python loop (assuming we don't need to reload all the dataset association data into sql alchemy anyway...) but we still would have to update potentially thousands of rows in the transaction. It'd be better if there was just some link in the dataset that said "here is the job that has my state, unless I have a metadata error". 

As a stop gap though - a trigger might really be a huge help and could probably be implemented independently of this PR.

natefoo:
So, moving the state tracking to the job was exactly why this PR stalled (albeit not for performance reasons, but yeah, I see how this is potentially problematic). That said, I am not sure that the reverse (including job state in the query used to display history contents) will perform well either.

jmchilton:
What if every dataset association had a direct link to the job that specifies its state? Our in-python ``creating_job`` logic is definitely not going to cut it.

mvdbeek:
Yes please, that would be great!

natefoo:
Yeah, I think that was what I was waffling over - whether HDAs, LDDAs, and all other dataset instances should get a `creating_job_id` column, although this duplicates the `job_to_output_*` tables. IIRC there is already dataset-referencing table that does this so it didn't seem like a terrible change? I forget which one, though.

mvdbeek:
> Our in-python `creating_job` logic is definitely not going to cut it.

We could probably use a column property to move this into the sqlalchemy layer. That should probably be more performant ?

jmchilton:
> although this duplicates the job_to_output_* tables

It is encoding slightly different information (in lieu of stuff like copied datasets, etc..) - so I wouldn't really call it duplicating those at all. It would also bring collections and datasets more inline - since collections get their state this way. I might not call it ``creating_job_id`` though - but something like that.

natefoo:
> since collections get their state this way

Ah yeah, that was what I was thinking of.

mvdbeek:
https://github.com/galaxyproject/galaxy/pull/10222 being merged might mean we can make some progress here ... but pushing to 21.09 for now

PR #7796, opened 2019-04-23
Add general purpose expression tool

mvdbeek:
This tool takes as input a dataset and will create an output parameter (currently text, integer, float or boolean) that can be connected to any tool parameter of that type in a workflow.

### DIFF
--- a/lib/galaxy/tools/parameters/wrapped_json.py
+++ b/lib/galaxy/tools/parameters/wrapped_json.py
@@ -66,7 +66,11 @@ def _json_wrap_input(input, value, handle_files="skip"):
             if value:
                 if isinstance(value, list):
                     value = value[0]
-                return _hda_to_object(value)
+                json_value = _hda_to_object(value)
+                if input.load_contents:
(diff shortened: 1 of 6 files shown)


lib/galaxy/tools/parameters/wrapped_json.py
```diff
@@ -66,7 +66,11 @@ def _data_input_to_path(v):
             if value:
                 if isinstance(value, list):
                     value = value[0]
-                return _hda_to_object(value)
+                json_value = _hda_to_object(value)
+                if input.load_contents:
+                    with open(str(value), mode='rb') as fh:
+                        json_value['contents'] = fh.read(input.load_contents).decode('utf-8', errors='replace')
```
jmchilton, line 72: Should we throw an error if the file is too big or something? This will occasionally crash Galaxy processes.

jmchilton: Nvm, that is bytes not a boolean - very cool.

PR #7797, opened 2019-04-23
Hide ExpressionTools in main analysis app

### LINKED ISSUE #7795: Expression tools should be hidden if not in workflow editor
If you try to load them up in the analysis app it'll just say `Tool execution through the `tool_runner` requires a `runtool_btn` flag or `URL` parameter.`. Since the expression tools are not very useful outside of workflows (at the moment at least) we should probably only display them in the workflow editor.

mvdbeek:
This shows expression tools only via the API or the workflow editor.
Resolves https://github.com/galaxyproject/galaxy/issues/7795


jmchilton:
Wonderful thank you 😄

PR #7798, opened 2019-04-23
Don't clean up jobs when running tests if `$GALAXY_TEST_NO_CLEANUP` is set

natefoo:
`$GALAXY_CONFIG_OVERRIDE_CLEANUP_JOB` can be used to override this behavior.

### DIFF
--- a/test/base/driver_util.py
+++ b/test/base/driver_util.py
@@ -172,6 +172,8 @@ def setup_galaxy_config(
             default_data_manager_config = data_manager_config
     data_manager_config_file = "%s,test/functional/tools/sample_data_manager_conf.xml" % default_data_manager_config
     master_api_key = get_master_api_key()
+    cleanup_job = 'never' if ("GALAXY_TEST_NO_CLEANUP" in os.environ or
+                              "TOOL_SHED_TEST_NO_CLEANUP" in os.environ) else 'onsuccess'
 
     # Data Manager testing temp path
     # For storing Data Manager outputs and .loc files so that real ones don't get clobbered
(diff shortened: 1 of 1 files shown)


mvdbeek (approved):

test/base/driver_util.py
```diff
@@ -172,6 +172,8 @@ def setup_galaxy_config(
             default_data_manager_config = data_manager_config
     data_manager_config_file = "%s,test/functional/tools/sample_data_manager_conf.xml" % default_data_manager_config
     master_api_key = get_master_api_key()
+    cleanup_job = 'never' if ("GALAXY_TEST_NO_CLEANUP" in os.environ or
+                              "TOOL_SHED_TEST_NO_CLEANUP" in os.environ) else 'onsuccess'
```
nsoranzo, line 176: What about?
```python
   cleanup_job = os.environ.get('GALAXY_TEST_CLEANUP_JOB', 'onsuccess')
```

natefoo: In that case would you need to set both `$GALAXY_TEST_NO_CLEANUP` and `$GALAXY_TEST_CLEANUP_JOB`? The goal was to have `cleanup_job` follow from `$GALAXY_TEST_NO_CLEANUP`.

nsoranzo: Ah, I didn't realise we already use these 2 env variables. Their original aim looks a bit different though (i.e. keep Galaxy/TS files instead of job working dirs) and also this doesn't follow our more-or-less standard pattern `GALAXY_TEST_` followed by uppercased config option.

What's the use case?

natefoo: `GALAXY_TEST_NO_CLEANUP` keeps the temporary directory structure used for running a Galaxy server for tests - the database, datasets, etc. I was trying to diagnose a test failure that seemed to be the result of a job failure and wanted to look at the job exit code file and script, but they were removed due to the default value of cleanup_job. It seemed logical to me that if you're explicitly instructing the test framework to keep the test dir to inspect the test artifacts, you'd likely be interested in the jobs as well.

martenson: 👍 to that, I've hacked around this in the past few times

natefoo:
@nsoranzo thanks!

PR #7806, opened 2019-04-24
Return HTTP 400 for `GET /api/datasets/nonexistent_id`

nsoranzo:
Previously it was returning HTTP status code 200 and the exception message as content.

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/datasets.py
+++ b/lib/galaxy/webapps/galaxy/api/datasets.py
@@ -126,10 +126,7 @@ class DatasetsController(BaseAPIController, UsesVisualizationMixin):
         Displays information about and/or content of a dataset.
         """
         # Get dataset.
-        try:
-            dataset = self.get_hda_or_ldda(trans, hda_ldda=hda_ldda, dataset_id=id)
-        except Exception as e:
-            return str(e)
+        dataset = self.get_hda_or_ldda(trans, hda_ldda=hda_ldda, dataset_id=id)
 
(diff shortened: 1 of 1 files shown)


lib/galaxy/webapps/galaxy/api/datasets.py
```diff
@@ -126,10 +126,7 @@ def show(self, trans, id, hda_ldda='hda', data_type=None, provider=None, **kwd):
         Displays information about and/or content of a dataset.
         """
         # Get dataset.
-        try:
-            dataset = self.get_hda_or_ldda(trans, hda_ldda=hda_ldda, dataset_id=id)
-        except Exception as e:
-            return str(e)
+        dataset = self.get_hda_or_ldda(trans, hda_ldda=hda_ldda, dataset_id=id)
```
mvdbeek, line 129: Can you also try replacing `legacy_expose_api_anonymous` and use `expose_api_anonymous` ?
I think that should work fine here.

nsoranzo: I don't think the way exceptions are dealt with in the try/except block below is how non-legacy API calls are supposed to work.
Can we postpone that to a 19.09 `enhancement` PR?

mvdbeek (approved):

lib/galaxy/webapps/galaxy/api/datasets.py
mvdbeek: sure

PR #7807, opened 2019-04-24
Historian feature added

KyleLamoureux:
A new feature for galaxy. It adds, a new button to the history options drop down. Which when clicked takes you to a mako form which will collect history information and use it to create a write up about the history. It works by creating a new workflow and then extracting the needed data from both this new workflow and history to generate the write up. It then creates a new zipped directory for this extracted data and then it will downloaded the zipped directory.


mvdbeek:
Thanks for your contribution @KyleLamoureux, this is a good idea and I'm sorry for the late feedback. I think we've made and are making quite some progress with refactoring the history export and re-working Galaxy Pages, which also allow you to describe an analysis. I think that the approach taken here doesn't fit in well with our current work. If you do want to continue working on this feel free to send an email to the galaxy mailing list or discuss with us in the gitter channel and we can try to outline a plan to work on or open an issue with the goals you'd like to achieve.

Issue #1270, opened 2015-12-08
Bed to gff3


hexylena:
https://biostar.usegalaxy.org/p/2083/ apparently this is the feature request which spawned our IRC channel ... 5 years ago (thanks @natefoo). It also hasn't been solved and I recently tried to feed the output of this tool to a gff3 tool and realised what was going on. A little googling later revealed that nothing has been done yet.

souravsingh:
I would like to work on this.Where do I start?

hexylena:
@souravsingh the existing bed2gff tool is https://github.com/galaxyproject/galaxy/blob/dev/tools/filters/bed_to_gff_converter.py

This would involve:
- writing an updated version of that tool that outputs gff3
- making a PR to galaxy 
  - adding it to the filters folder, alongside the original
  - adding it as a converter in the datatypes_conf.xml (https://github.com/galaxyproject/galaxy/blob/dev/config/datatypes_conf.xml.sample#L21)

Thanks for your interest! :D

souravsingh:
@erasche  Would it be alright if I make use of tools from BioPython for the conversion or do I need to modify the existing file?

hexylena:
Hi @souravsingh it would be easiest to get the contribution in if you didn't use any of the external dependencies, I'm afraid.

apcamargo:
In case anyone is still interested, I modified the Galaxy script to convert BED files to GFF3, instead of GFF2. I updated the script to Python 3 and included argparse for the CLI interface.

https://gist.github.com/apcamargo/50edadc6ab13bfafd05eb4b887f8dd6d

nsoranzo:
@apcamargo I think there would be interest for a pull request.

See also https://github.com/galaxyproject/galaxy/issues/7771

jennaj:
Having a BED > GFF3 converter would be great. Sorry if naive question -- but we plan to keep both conversion options, correct? We need both conversions available, or just GTF, for the reasons below.

Many tools do not consume GFF3 (only actual GTF content, sometimes given a "GFF" datatype). Some accept both GTF/GFF3. Tools can be unclear in the input datatype filter area and filter for just "gff" -- which could mean any of GFF, GTF (GFF2), or GFF3. 

[...]

PR #7808, opened 2019-04-24
Remove GFF headers during conversion.

### LINKED ISSUE #7771: CONVERTER bed2gff1 creates headers in the result, causing many downstream tools to fail
**Galaxy Tool ID:** bed2gff1
**Galaxy Tool Version:** 2.0.0

Headers added are:
```
##gff-version 2
##bed_to_gff_converter.py
```

**Problem:** Any tool that uses GFF/GTF input, resulting from a data format conversion using this utility, that does not "grep -v" the headers out fails for odd formatting reasons that users do not understand.

Tools that use implicitly converted BED-to-GFF can fail, same reason.

This is packaged into a tool panel "tool" converter, an edit attributes "covert" converter function, and an implicit data converter.

**Suggested solution:** It would be better to **not** add in a header and keep the file in the most conservative format, so the result will work with any tool/function. Headers in GFF/GFT files are a very common cause of end-user reported problems. We cannot control what is published by public data sources, but we can help to format data already inside of (or produced by) Galaxy in a way that is more likely to work successfully with our tools.

**WORKAROUND for End-Users** 

* If your BED data is being implicitly converted to GFF format by Galaxy (at runtime), convert the data first using the tool `BED-to-GFF` or the `Edit Attributes` (pencil icon) "Convert" function. Once in GFF format, run the tool `Select` with the option "NOT Matching" and regular expression `^#`. This will remove all header lines. 
* If your BED data is already converted to GFF, and has headers, then just run the `Select` tool as described above.
* For either case, input the filtered "header removed" GFF result into the tool presenting with the problem.

**Example of an error (IUC tool given bed input, implicitly converted to GFF)** There are many others (most Bioconductor tools, etc).

```
Dataset Error
An error occured while running the tool toolshed.g2.bx.psu.edu/repos/iuc/cwpair2/cwpair2/1.0.0.

Tool execution generated the following messages:

Unable to parse file "/galaxy-repl/main/files/031/018/dataset_31018984.dat".
  File "/cvmfs/main.galaxyproject.org/shed_tools/toolshed.g2.bx.psu.edu/repos/iuc/cwpair2/d4db13c9dd7f/cwpair2/cwpair2_util.py", line 280, in perform_process
    chromosomes = parse_chromosomes(input)
  File "/cvmfs/main.galaxyproject.org/shed_tools/toolshed.g2.bx.psu.edu/repos/iuc/cwpair2/d4db13c9dd7f/cwpair2/cwpair2_util.py", line 96, in parse_chromosomes
    cname, junk, junk, start, end, value, strand, junk, junk = line
ValueError: need more than 1 value to unpack
```

jmchilton:
Fixes #7771.

### DIFF
--- a/tools/filters/bed_to_gff_converter.py
+++ b/tools/filters/bed_to_gff_converter.py
@@ -13,8 +13,6 @@ def __main__():
     skipped_lines = 0
     first_skipped_line = 0
     out = open(output_name, 'w')
-    out.write("##gff-version 2\n")
-    out.write("##bed_to_gff_converter.py\n\n")
     i = 0
     for i, line in enumerate(open(input_name)):
         complete_bed = False


nsoranzo:
I don't think that is the appropriate fix, the cwpair2 tool is at fault for not skipping the comment and empty lines, which are part of GFF "standard": https://github.com/The-Sequence-Ontology/Specifications/blob/master/gff3.md#other-syntax

jmchilton:
Ping @jennaj - can you respond to this?

nsoranzo:
I've been working today to update the cwpair2 tool to fix this and other bugs, will open a PR soon.

nsoranzo:
See https://github.com/galaxyproject/tools-iuc/pull/2382

jennaj:
The tool needs a GTF format, not GFF3, right? Or is it one of the tools that accept both? 

Screen Shot 2019-05-02 at 12 45 08 PM

Screen Shot 2019-05-02 at 12 44 23 PM

GTFs can and do have header lines from public sources, but it is not in spec. Some tools know how to strip that out, most do not (and our wrappers do not address it -- just some examples are ~~all~~ **MOST** of the Bioconductor tools.

HTseq looks for "gff", finds all, but really only accepts GTF (or a "GFF" that has something in the 9th field that is formatted as if it was in GTF format). 


Screen Shot 2019-05-02 at 12 57 29 PM

Screen Shot 2019-05-02 at 12 57 43 PM


Format specs, are many more, and everyone seems wary of the public sources but also very specific about what their particular tools expect. Google for "GFT GFF3" and find loads more all over the place.

https://software.broadinstitute.org/software/igv/GFF
http://mblab.wustl.edu/GTF22.html
http://genome.ucsc.edu/goldenPath/help/GTF.html

GTF headers are a problem across tools, not just this one. **I think using the datatype "gff" is problematic for many reasons. It is too generic.** Tools either actually use GTF content (expect "attributes" in 9th field, not just blank or "comments") or GFF3 content. Some accept both. When the wrong one is given, and it is out of spec, tools error for odd reasons that users find difficult to understand.

See my comments in here for details. This impact implicit conversion. https://github.com/galaxyproject/galaxy/issues/1270#issuecomment-488796227. That also links to this: galaxyproject/usegalaxy-playbook#45

I help up to 10 people a day to figure out why their GFF/GTF/GFF3 data isn't working. Be good to get some solution in place, and that might mean more tool wrapper fixes, or something else.

Few examples that show the complicated ways using just "gff" as an acceptable input datatype becomes a problem for users. I can share more (!!)

https://help.galaxyproject.org/t/featurecounts-to-analyse-mir-rna-seq-data/492
https://help.galaxyproject.org/t/htseq-count-shows-fatal-error-every-time/753
https://help.galaxyproject.org/t/rna-star-and-hg38-gtf-reference-annotation/749

jennaj:
And here is featurecounts, which also fails with GFF3 input (shows why user in https://help.galaxyproject.org/t/featurecounts-to-analyse-mir-rna-seq-data/492 was confused about the error). Look at the input select filter, what it finds, versus the help, versus what actually happens (tool fails if given GFF3).

Screen Shot 2019-05-02 at 12 48 01 PM

Screen Shot 2019-05-02 at 12 48 12 PM

Screen Shot 2019-05-02 at 12 47 51 PM

We even contradict in the select versus help on this one:

Screen Shot 2019-05-02 at 1 09 21 PM

jennaj:
I put out a lot of info. Not focused on a particular fix. Anything that makes the input select correct is good (do not show GFF3 when it is not accepted, and the tool will always fail). 

If all tools can be changed to work with GTF that have headers, that would be good too (just seems like a giant project).

And get rid of the generic "gff" select filter -- I think it causes many many problems. Be specific -- GTF or GFF3 or both, but only when both can actually be used by the tool.

nsoranzo:
@jennaj 

> The tool needs a GTF format, not GFF3, right? Or is it one of the tools that accept both?

cwpair2 does not use the 9th column, so works on both formats.

> GTFs can and do have header lines from public sources, but it is not in spec.

Agreed.

> Some tools know how to strip that out, most do not (and our wrappers do not address it -- just some examples are all MOST of the Bioconductor tools.

If an underlying tool doesn't support the comment lines, we should either fix the underlying tool or workaround the issue in the tool wrapper. But we would need specific issues in the relevant tool repositories (e.g. tools-iuc) to track them separately.

> And get rid of the generic "gff" select filter -- I think it causes many many problems. Be specific -- GTF or GFF3 or both, but only when both can actually be used by the tool.

I agree that tools should be as specific as possible in the `format` attribute. Unfortunately I don't think we can remove `gff` from the available formats without breaking a consistent number of tools. Also, it serves as a fallback for [GFF2](http://gmod.org/wiki/GFF2) (the deprecated format, not GTF).

A bit unrelated: when looking at `config/datatypes_conf.xml.sample` I noticed that we have:
```xml
    <sniffer type="galaxy.datatypes.interval:Gtf"/>
    <sniffer type="galaxy.datatypes.interval:Gff"/>
    <sniffer type="galaxy.datatypes.interval:Gff3"/>
```

Do you have any evidence of GFF3 files wrongly detected as `gff` when uploaded?

jennaj:
@nsoranzo 

Thanks for feedback!

> Do you have any evidence of GFF3 files wrongly detected as gff when uploaded?

Not that I have seen, if the gff3 header is there, "gff3" is the assigned datatype. If a GFF3 does not have a header, or the user "converted spaces to tabs" in upload, breaking the expected header format, then it is assigned the datatype "txt". What happens with that, usually, is they try to fix the "txt" problem by manually assigning "gtf" then other things go wrong (tool fails because data is not actually GTF), and that is what ends up being reported. I don't think we can do anything about that.

I have seen GFF3 with "extra" comment lines internally (forget who publishes these) and those always cause tool errors. Any sequence content at the end also reliably causes tool errors. Including sequences at the end is in spec, technically. Not sure if extra comment lines _internal to the file_ are in spec (but probably not, from what I can tell).

SO, it is just GTF's with headers that get assigned GFF. Happens quite a bit. GTFs from Gencode are one example. These always need to be "fixed up" first (headers removed). That is where the "break out the header into a separate dataset upon upload" idea came from. It can be informative, just breaks tools.

> If an underlying tool doesn't support the comment lines, we should either fix the underlying tool or workaround the issue in the tool wrapper. But we would need specific issues in the relevant tool repositories (e.g. tools-iuc) to track them separately.

I don't think I know of a tool that DOES accepts a GTF with any header lines, except if that is what you did for the cwpair2 tool now (?) or was the change to convert to GFF3 instead? Wasn't sure.

Might be able to make a tool list just by querying out all tools that DO accept GTF (at all, even if also accept GFF3) and make sure they handle potential headers already (probably not) and strip them out in the wrapper (actually, James suggested that a few weeks ago).

**All that said, I still think the bed-to-gtf should not be adding in a header.** It adds no useful info, creates an out of spec GTF, and why create potential problems in a built-in function used by explicit conversion, the stand-alone tool, and the edit attributes convert functionality? Plus, if someone downloads the Galaxy version of a bed-to-gtf, it is out of spec and would fail with whatever other tools they are using line-command or in other web apps. 

Not all tools can be updated -- legacy, not iuc or devteam, etc. So even if we fix the tools we can to strip out headers (in public GTFs), still seems like it just creates trouble to create out-of-spec GTF headers from scratch ourselves. If we were adding in some kind of useful metadata (data source, etc), that's different -- but we are not in this case -- and never would be able to anyway with just a random bed file as input. We are not publishing these out as reference annotation (is why public sources are probably including the headers, to add source/metadata info).

Pretty sure we all agree there are problems ... just need to come up with the best solutions. I think that means 1) we don't create GTF headers ourselves 2) come up with some better way to deal with external GTFs with headers. In upload, or in tool wrappers, or maybe both ... but do something so GTFs are no longer leading to so many usage issues. 

And ... Mixing up GFF/GTF/GFF3 on tool forms .. needs some solution. Are so many mix-ups by users. I get that this isn't simple and we've been using "gff" as a catch-all. All kinds of hybrid formats are out there. We can't control that -- but we can control how we sort them out as acceptable inputs when we already know what a tool expects. Maybe start with the Bioconductor tools (eliminate gff3 as an accepted input)? Those are heavily used, would be worth it.

jennaj:
thank you!

PR #7810, opened 2019-04-24
Requirejs template purge

dannon:
This drops the (now nonexistent) requirejs lib from our templates.  This has been deprecated for some time now, and I grepped around and didn't see anything, but -- does anyone know of any plugins which are still using require-based entrypoints?

(marking as a bug, since these files don't exist anymore and throw 404s when we attempt to access them)

### DIFF
--- a/templates/base.mako
+++ b/templates/base.mako
@@ -51,7 +51,6 @@
 <%def name="javascripts()">
     ## TODO: remove when all libs are required directly in modules
     ${h.js(
-        'libs/require',
         'bundled/libs.chunk',
         'bundled/base.chunk'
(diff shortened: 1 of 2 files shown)


mvdbeek:
let's go ahead, we'll figure out if plugins still rely on this sooner or later anyway.

dannon:
@mvdbeek Thanks!  I should have updated the original text -- we actually were using require in IE templates but I'm working on overhauling that (in a separate branch) to fix that bug.  Either way, this PR was fine to merge since requirejs doesn't exist anymore, so no worries there.

PR #7811, opened 2019-04-24
Data Manager interface - use standard card header styles without extra markup

dannon:
(which gets escaped)


scrathat:
The current header markup can be kept if the header slot is used. I've included the correction in #7332 .

dannon:
@scrathat Thanks, but part of the intent here was to use the default card header styles as well, so that card headers have a consistent look across the application instead of having varying sized headers.

client/galaxy/scripts/components/admin/DataManager/DataManager.vue
```diff
@@ -6,18 +6,10 @@
         <div v-else-if="dataManagers && !dataManagers.length">
             <Alert variant="primary">
```
guerler, line 7: Minor but could use a regular `b-alert` instead of `Alert` here.

dannon: @guerler Agreed; I didn't add this one.  I'd support separately making a pass removing the custom thin Alert wrapper and standardizing on normal b-alert, but that's not refactoring I'm doing in this PR.  When I looked at this when it first got added I think in a few places there were responses from the server that we needed to update to use the correct class, but that's what we should do instead of having the extra wrapper.

martenson:
This branch is in (seemingly trivial) conflict, could you please resolve it?

dannon:
Conflict resolved, fixed prefix bug from #8193 as well.

mvdbeek (approved):
Thanks @dannon !

PR #7817, opened 2019-04-25
Run tests for Natefoo job state granularity

mvdbeek:


mvdbeek:
Alright, this worked, just the one selenium test that's still failing

dannon:
@mvdbeek I'll take care of that one.  This is a different test than was failing that I fixed in 59e4371 (but likely the same issue)

Issue #7412, opened 2019-02-26
Analyze 75 upload does not work in 19.01 version


foellmelanie:
Hi, 

I tried to upload Analayze 75 files to usegalaxy.org and I got the following error message: 

```
Fatal error: Exit code 1 ()
Traceback (most recent call last):
  File "/cvmfs/main.galaxyproject.org/galaxy/tools/data_source/upload.py", line 329, in <module>
    __main__()
  File "/cvmfs/main.galaxyproject.org/galaxy/tools/data_source/upload.py", line 320, in __main__
    metadata.append(add_composite_file(dataset, registry, output_path, files_path))
  File "/cvmfs/main.galaxyproject.org/galaxy/tools/data_source/upload.py", line 243, in add_composite_file
    stage_file(name, composite_file_path, value.is_binary)
  File "/cvmfs/main.galaxyproject.org/galaxy/tools/data_source/upload.py", line 223, in stage_file
    sniff.convert_newlines(dp, tmp_dir=tmpdir, tmp_prefix=tmp_prefix)
  File "/cvmfs/main.galaxyproject.org/galaxy/lib/galaxy/datatypes/sniff.py", line 122, in convert_newlines
    for i, line in enumerate(io.open(fname, mode="U", encoding='utf-8')):
  File "/cvmfs/main.galaxyproject.org/venv/lib/python2.7/codecs.py", line 314, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf8' codec can't decode byte 0x80 in position 0: invalid start byte
```

Planemo with 18.09 is working with Analyze 75 files but not Planemo with 19.01 version: 
https://github.com/galaxyproteomics/tools-galaxyp/pull/350

mvdbeek:
Thanks for the report @foellmelanie. The datatype for analyze75 files says that the hdr file is not binary:
```
        """The header file. Provides information about dimensions, identification, and processing history."""
        self.add_composite_file(
            'hdr',
            description='The Analyze75 header file.',
            is_binary=False)
```
While the test data in the tool that fails is binary https://github.com/galaxyproteomics/tools-galaxyp/blob/f127be2141cf22e269c85282d226eb16fe14a9c1/tools/cardinal/test-data/Analyze75.hdr

[...]

Issue #5326, opened 2018-01-17
Composite dataset - error with optional file during upload


Alanamosse:
For composite datatypes, composite file should be optional with the specification "optional=True" set in the class definition.
Here's the OBFS type definition with 5 optional composites files :

```
class OBFS(Binary):
    """OpenBabel Fastsearch format (fs)."""
    file_ext = 'obfs'
    composite_type = 'basic'
    allow_datatype_change = False
    MetadataElement(name="base_name", default='OpenBabel Fastsearch Index',
                    readonly=True, visible=True, optional=True,)

    def __init__(self, **kwd):
        """
            A Fastsearch Index consists of a binary file with the fingerprints
            and a pointer the actual molecule file.
        """
        Binary.__init__(self, **kwd)
        self.add_composite_file('molecule.fs', is_binary=True,
                                description='OpenBabel Fastsearch Index')
        self.add_composite_file('molecule.sdf', optional=True,
                                is_binary=False, description='Molecule File')
        self.add_composite_file('molecule.smi', optional=True,
                                is_binary=False, description='Molecule File')
        self.add_composite_file('molecule.inchi', optional=True,
                                is_binary=False, description='Molecule File')
        self.add_composite_file('molecule.mol2', optional=True,
                                is_binary=False, description='Molecule File')
        self.add_composite_file('molecule.cml', optional=True,
                                is_binary=False, description='Molecule `File')
```

However this option doesn't seem to work during upload, every composite file of composite type are needed to validate upload even with the optional tag.

mvdbeek:
This should be easy to fix, I'll have a look

PR #7819, opened 2019-04-25
[19.05] Add shapefile datatype

bgruening:
ping @yvanlebras, @Alanamosse and @annefou


mvdbeek:
Is this replacing https://github.com/galaxyproject/galaxy/pull/7615 ?

bgruening:
@mvdbeek yes. Greetings from the Tool dev hackathon!

bgruening:
@mvdbeek if you have a minute. Thanks!

pcm32:
Great to see shapefile and gis here! are there any viewers for shapefiles on Galaxy?

bgruening:
Not yet, but it's on the ToDo list, at first a few tools than the Vis ... :)

test/functional/tools/composite_shapefile.xml
```diff
@@ -0,0 +1,35 @@
+<tool id="composite" name="composite" version="1.0.0">
```
mvdbeek: To test the tool you need to add it to `test/functional/tools/samples_tool_conf.xml` and the datatype to `test/functional/tools/sample_datatypes_conf.xml`.

mvdbeek:
```suggestion
<tool id="shapefile_composite" name="composite shapefile tool" version="1.0.0">
```

bgruening:
Putting it as WIP, I would like to get the second output working as well.

bgruening:
```
  File "/galaxy/test/functional/tools/upload.py", line 326, in <module>
    __main__()
  File "/galaxy/test/functional/tools/upload.py", line 317, in __main__
    metadata.append(add_composite_file(dataset, registry, output_path, files_path))
  File "/galaxy/test/functional/tools/upload.py", line 240, in add_composite_file
    stage_file(name, composite_file_path, value.is_binary)
  File "/galaxy/test/functional/tools/upload.py", line 220, in stage_file
    sniff.convert_newlines(dp, tmp_dir=tmpdir, tmp_prefix=tmp_prefix)
  File "/galaxy/lib/galaxy/datatypes/sniff.py", line 122, in convert_newlines
    for i, line in enumerate(io.open(fname, encoding='utf-8')):
  File "/galaxy_venv/lib/python2.7/codecs.py", line 314, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf8' codec can't decode byte 0xe8 in position 28: invalid continuation byte
```

This looks familiar. Probably related to https://github.com/galaxyproject/galaxy/issues/7412

annefou:
That is similar to what I am getting with my Galaxy tool when we have special characters such as icelandic characters (è, etc.)

yvanlebras:
hmm, maybe we have to recode example dataset into UTF8 no ?

mvdbeek:
I don't think this is the same issue as #7412, that one is about a non-UTF8 text file. The only text file in the test data can be read with UTF-8. On a first glance I'd say this more looks like the composite upload doesn't respect the `is_binary` flag

test/functional/tools/composite_shapefile.xml
```diff
@@ -18,21 +18,21 @@
     <tests>
         <test>
             <param name="input" value="" ftype="shp" >
-                <composite_data value="shapefile/shapefile.dbf"/>
-                <composite_data value="shapefile/shapefile.prj"/>
                 <composite_data value="shapefile/shapefile.shp"/>
                 <composite_data value="shapefile/shapefile.shx"/>
+                <composite_data value="shapefile/shapefile.dbf"/>
+                <composite_data value="shapefile/shapefile.prj"/>
```
mvdbeek, line 24: And I was just looking at why the prj file looked like a binary file after the upload ...

mvdbeek:
```
        if found_exceptions:
>           raise JobOutputsError(found_exceptions, job_stdio)
E           galaxy.tools.verify.interactor.JobOutputsError: Test file (shapefile/output.html) is missing. If you use planemo try --update_test_data to generate one.
```

mvdbeek:
I've added a fix for python 3 in https://github.com/bgruening/galaxy/pull/42

test/functional/tools/composite_shapefile.xml
bgruening: Do you know if this recently changed? It wired that you need to stick to the ordering of the datatype definition ... No tool dev should need to know this :(

bgruening:
Thanks @mvdbeek ... let's see if this passes now.

mvdbeek:
@galaxybot test py3

bgruening:
@galaxybot test py3

test/functional/tools/composite_shapefile.xml
```diff
@@ -29,10 +29,10 @@
...
             </output>
             <output name="output_shapefile" file="shapefile/output.html">
-                <extra_files type="file" name="shapefile.shp" value="shapefile/shapefile.shp" compare="sim_size" />
-                <extra_files type="file" name="shapefile.shx" value="shapefile/shapefile.shx" compare="sim_size" />
-                <extra_files type="file" name="shapefile.dbf" value="shapefile/shapefile.dbf" compare="sim_size" />
-                <extra_files type="file" name="shapefile.prj" value="shapefile/shapefile.prj" compare="sim_size" />
+                <extra_files type="file" name="shapefile.shp" value="shapefile/shapefile.shp" md5="ae4648a8a6be706e910e46607f261317" />
+                <extra_files type="file" name="shapefile.shx" value="shapefile/shapefile.shx" md5="6321ef787b7826afe06e9cf3eb20e656" />
+                <extra_files type="file" name="shapefile.dbf" value="shapefile/shapefile.dbf" md5="d65d87e1570510af8a81d7d174ea361a" />
+                <extra_files type="file" name="shapefile.prj" value="shapefile/shapefile.prj" lines_diff="2" />
```
mvdbeek, line 35: It's the html file that is different, I think you need to add lines_diff there (as well ?)

test/functional/tools/composite_shapefile.xml
mvdbeek: It's probably always been that way, otherwise there's a good chance existing tests would have broken.
I don't see anywhere in the code that we match up the filename to the composite file (but that's just skimming over the code)

mvdbeek:
If you don't want to wait for jenkins you can also run the tests using `sh run_tests.sh -framework -id shapefile_composite`.

mvdbeek:
```
collected 219 items / 218 deselected / 1 selected

test/functional/test_toolbox_pytest.py::test_tool[shapefile_composite_test_1] PASSED                                                                   [100%]
```

mvdbeek (approved):

lib/galaxy/datatypes/gis.py
```diff
@@ -0,0 +1,68 @@
+"""
+GIS classes
+"""
+
+import logging
+
+from galaxy.datatypes.binary import Binary
+
+log = logging.getLogger(__name__)
```
mvdbeek: This is unused I think.

lib/galaxy/datatypes/gis.py
```diff
@@ -6,8 +6,6 @@
 
 from galaxy.datatypes.binary import Binary
 
-log = logging.getLogger(__name__)
-
```
mvdbeek: You also have to drop the logging import

mvdbeek (approved):

mvdbeek:
Do you want to target 19.05 ?

bgruening:
> Do you want to target 19.05?

That would be ideal!

annefou:
That would be fantastic! Thank you.

bgruening:
Green and against 19.05. Thanks @mvdbeek!

hexylena (approved):

hexylena:
Awesome! Can't wait to see the visualisations

anuprulez:
shapefile_upload

I am not able to upload `shapefiles`. I have added the three mandatory files for the composite `shp` filetype. The `Start` button is not enabled. Am I missing something? Thanks!

yvanlebras:
hmm, this make me remember the issue we had with @Alanamosse one year ago.. https://github.com/galaxyproject/galaxy/issues/5326

If I well remember, we were thinking about a bug from the Galaxy side as we encounter the same issue with other composite datatypes who can have optional files (babel related things if I well remember..)

yvanlebras:
So is https://github.com/galaxyproject/galaxy/pull/7872 the origin of the issue ?

mvdbeek:
https://github.com/galaxyproject/galaxy/issues/5326 is the issue, https://github.com/galaxyproject/galaxy/pull/7872 should fix it, yes.

yvanlebras:
You're right! Thank you Marius to be so limpid!

Issue #7758, opened 2019-04-16
Excessive WSGI processes created with ObjectStore multipart upload


tmcgowan:
We are using an S3 object store as the primary means of data storage. When a file is split for multipart upload ~16 new WSGI processes are created at the same time. At times, we can have a dozen split jobs running with hundreds of WSGI processes. This is bringing the server down.

Has anyone reported or observed this behavior? Have I screwed up a configuration?  I really do not want to turn off the object store, but this behavior can't continue.

We are running 18.09 with the following YML:

```
uwsgi:
 http: 127.0.0.1:8080 
 processes: 5
 threads: 2
 offload-threads: 1
 static-map: /static/style=static/style/blue
 static-map: /static=static
 master: true
 virtualenv: .venv
 module: galaxy.webapps.galaxy.buildapp:uwsgi_app()
 die-on-term: true
 hook-master-start: unix_signal:2 gracefully_kill_them_all
 hook-master-start: unix_signal:15 gracefully_kill_them_all
 py-call-osafterfork: true
 enable-threads: true
 thunder-lock: true
 http-raw-body: True
```

And an object_store_conf.xml of:

```
<?xml version="1.0"?>
<object_store type="hierarchical">
    <backends>
       <object_store type="swift" id="primary" order="0">
         <auth access_key="" secret_key="" />
            <bucket name="galaxy" use_reduced_redundancy="True" />
            <connection host="s3.msi.umn.edu"  port="443"/>
            <cache path="database/files" size="500" />
            <extra_dir type="job_work" path="database/job_working_directory"/>
            <extra_dir type="temp" path="database/tmp"/>
       </object_store>
       <object_store type="disk" id="secondary" order="1">
            <files_dir path="/panfs/roc/galaxy/PRODUCTION/database/files"/>
        </object_store>
    </backends>
</object_store>
```

I've attached a PS output with an example of 14 split jobs and hundreds of WSGI processes. 

This is a simple example:

Process 9402 is one of the original WSGI processes created by run.sh start. The split job, process 14863 is created by 9402. At the same time 16 new WSGI processes are also started by 9402. Once the split job ends, the WSGI processes end as well. 

`
[...]

PR #7824, opened 2019-04-25
[19.01] Possible fix for WSGI process issue #7758

tmcgowan:
PR originating from issue #7758 

The use of multiprocessing.Pool and creating as many WSGI processes as CPUs was crashing our server when users uploaded multiple files. Our server has 16 vCPUs, so at times hundreds of WSGI processes were created to upload files to the object store.
Since this code is executing on the hosting server and from within a WSGI process, spawning full-weight processes to write files is too much of a resource drain on the hosting server. 

I made a minimal change and moved to lighter weight threads instead of processes. I have been running this on our dev instance and it is working.

As a side note, in 18.09 the lib/galaxy/jobs/__init__.py file called the s3 multipart code twice in succession. This greatly added to the server load. Looks like this issue has been eliminated in 19.01

18.09
```
                if 'uuid' in context:
                    dataset.dataset.uuid = context['uuid']
                # Update (non-library) job output datasets through the object store
                if dataset not in job.output_library_datasets:
                    self.object_store.update_from_file(dataset.dataset, create=True)
                self.__update_output(job, dataset)
```
19.01
```
                if 'uuid' in context:
                    dataset.dataset.uuid = context['uuid']
                self.__update_output(job, dataset)
 
```

### DIFF
--- a/lib/galaxy/objectstore/s3_multipart_upload.py
+++ b/lib/galaxy/objectstore/s3_multipart_upload.py
@@ -5,26 +5,16 @@ This parallelizes the task over available cores using multiprocessing.
 Code mostly taken form CloudBioLinux.
 """
 
-import contextlib
-import functools
 import glob
-import multiprocessing
 import os
 import subprocess
-from multiprocessing.pool import IMapIterator
+import threading
 
 try:
     import boto
     from boto.s3.connection import S3Connection
 except ImportError:
     boto = None
-
-
-def map_wrap(f):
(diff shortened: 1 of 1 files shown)


mvdbeek:
@galaxybot test this

mvdbeek (approved):
Thanks a lot, this looks good to me. It just seems like a surprising choice to use multiprocessing here in the first place.

dannon:
Thanks for this @tmcgowan!

PR #7827, opened 2019-04-26
Add option to wait for db to become available

jdavcs:
Configurable option to wait for the database to become available instead
of failing immediately.

_configure_models() seems to be the earliest/reasonable place for it.


config/galaxy.yml.sample
```diff
@@ -163,6 +163,15 @@ galaxy:
   # not recommended for production use.
   #database_auto_migrate: false
 
+  # Wait for database to become available instead of failing immediately.
+  #database_wait: false
+
+  # Number of attempts before failing.
+  #database_wait_attempts: 60
+
+  # Sleep period between attepmts (seconds; can be a floating point number).
```
nsoranzo:
```suggestion
  # Sleep period between attempts (seconds; can be a floating point number).
```

nsoranzo:
`config/galaxy.yml.sample` should not be modified directly, you should update `lib/galaxy/webapps/galaxy/config_schema.yml` and then execute `make config-rebuild` .

lib/galaxy/config.py
```diff
@@ -1196,3 +1206,15 @@ def _configure_models(self, check_migrate_databases=False, check_migrate_tools=F
     def _configure_signal_handlers(self, handlers):
         for sig, handler in handlers.items():
             signal.signal(sig, handler)
+
+    def _wait_for_database(self, url):
+        from time import sleep
```
nsoranzo: I'd move the standard library import to the top of the file (line 21).

jdavcs: I would too, no question about that! Was trying to be consistent with the rest of the file (and tread softly). So yes, will do.

config/galaxy.yml.sample
jdavcs: and thanks! :)

lib/galaxy/config.py
jdavcs: We can do without it: time module is already imported, so just time.sleep(pause).
I'm leaving the sqlalchemy_utils import in the method.

nsoranzo: > We can do without it: time module is already imported, so just time.sleep(pause).

Surely better!

jdavcs: Done. Thank you for reviewing!

jdavcs: Correction: not done. I've just noticed your comment (that I missed) on generating config/galaxy.yml.sample. Working on it...

nsoranzo (approved):

nsoranzo:
@ic4f Still WIP?

lib/galaxy/config.py
```diff
@@ -1196,3 +1206,14 @@ def _configure_models(self, check_migrate_databases=False, check_migrate_tools=F
...
         for sig, handler in handlers.items():
             signal.signal(sig, handler)
+
+    def _wait_for_database(self, url):
+        from sqlalchemy_utils import database_exists
+        attempts = self.config.database_wait_attempts
+        pause = self.config.database_wait_sleep
+        for i in range(1, attempts):
+            try:
+                database_exists(url)
```
nuwang, line 1216: Does this need a break here so the loop doesn't continue if the database exists?

jdavcs: Of course! (yikes... missed this one!) Thank you! Pushing...
The rest - done, no longer WIP (config schema updated, make config rebuild done).

nsoranzo (approved):

dannon:
Selenium error is unrelated.  Thanks @ic4f!

PR #7829, opened 2019-04-26
Fix tool search function

mvdbeek:
This broke in https://github.com/galaxyproject/galaxy/pull/7797
and was causing the Selenium test to fail (which weirdly enough it didn't in that PR).


dannon:
@mvdbeek Thanks!

PR #7831, opened 2019-04-26
Make class-scope variables global if accessed from list comprehension on python 3

mvdbeek:
This will allow tools to work that access variables from the enclosing scope in a list/dict/tuple comprehension.
WIP while I add some tests and figure out what we need to use for generator and dict comprehensions

### DIFF
--- a/lib/galaxy/util/template.py
+++ b/lib/galaxy/util/template.py
@@ -1,10 +1,19 @@
 """Entry point for the usage of Cheetah templating within Galaxy."""
+from __future__ import absolute_import
+
+import logging
+import sys
+import traceback
+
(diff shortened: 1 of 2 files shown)


mvdbeek:
Meh, this is not going to work, if I make the variable global it isn't going to be updated and stays `None`. I tried a couple of other things; what would work is injecting a `#var = $var` before the comprehension statement and changing `$var` to `var`, but I haven't figured out how to get the line number for the template line that causes the issue. And even if we got that one we couldn't just inject this blindly as we may be in a multiline statement.
We may need to have a sandbox after all for these tools ...

Issue #7820, opened 2019-04-25
Old job unexpected finishing triggers deletion of newer job on multiple resubmissions of the same job


pcm32:
Using the cli runner with LSF I have noticed the following failure scenario when a job gets resubmitted twice (sequentially). 

- Job A runs with external identifier 1 and fails on a resubmit-able condition
- Job A is resubmitted with external identifier 2 on a different destination and fails on a resubmit-able condition
- Job A is resubmitted with external identifier 3. On Start of external identifier 3, check_watched_items enquires about the state Job A with external identifier 1, it doesn't find it, and marks Job A/1 as done, which in turn signals the finishing of Job A/3 which only started running. Because it is a user cancelation, Job A/3 gets marked as failed. Job A/2 also gets the same treatment, again triggering a finish signal for A/3.

You can see this from these logs, where A/1 is 2292/6149125, A/2 is 2292/6152223 and A/3 is 2292/6160157.
```
galaxy.jobs.runners.cli INFO 2019-04-25 01:26:48,515 [p:35980,w:2,m:0] [ShellRunner.work_thread-3] (2292) queued with identifier: 6160157
galaxy.jobs DEBUG 2019-04-25 01:26:48,515 [p:35980,w:2,m:0] [ShellRunner.work_thread-3] (2292) Persisting job destination (destination id: dynamic-lsf-large)
galaxy.jobs.runners.cli DEBUG 2019-04-25 01:26:49,932 [p:35980,w:2,m:0] [ShellRunner.monitor_thread] (2292/6160157) state change: from new to queued
[pid: 35975|app: 0|req: 2292/4578] 10.7.35.87 () {38 vars in 630 bytes} [Thu Apr 25 01:38:43 2019] GET /api/histories/c333314861e68c5c?key=d6bfcf542103ec809d5f7c230b12788b => generated 2844 bytes in 198 msecs (HTTP/1.1 200) 3 headers in 124 bytes (1 switches on core 3)
[pid: 35980|app: 0|req: 2292/4588] 10.7.35.87 () {38 vars in 630 bytes} [Thu Apr 25 01:40:25 2019] GET /api/histories/c333314861e68c5c?key=d6bfcf542103ec809d5f7c230b12788b => generated 2844 bytes in 305 msecs (HTTP/1.1 200) 3 headers in 124 bytes (1 switches on core 0)
[...]

Issue #7118, opened 2018-12-14
Issue with resubmission + dynamic destination?


hexylena:
Hi @jmchilton since we debugged last time I'm having an issue with the resubmission and maybe you can shed some light on it. Since I can't really pass around parameters I've settled on having three destinations: 1 1,5 and 2x memory destinations.

I have the following dynamic destination:

```python
from galaxy.jobs import JobDestination

def testway(tool_id, user, memory_scale=1.0, next_dest=None):
    resubmit = []
    if next_dest:
        resubmit = [{
            'condition': 'any_failure and attempt <= 3',
            'destination': next_dest
        }]

    return JobDestination(
        id='blah',
        runner='condor',
        params={'description': 'aaaa', 'request_memory': '%sG' % memory_scale},
        resubmit=resubmit,
    )

def gateway_1x(tool_id, user):
    return testway(tool_id, user, memory_scale=1, next_dest='gateway_1_5x')

def gateway_1_5x(tool_id, user):
    return testway(tool_id, user, memory_scale=1.5, next_dest='gateway_2x')

def gateway_2x(tool_id, user):
    return testway(tool_id, user, memory_scale=2)
```

And the following job conf:

```xml
<job_conf>
    <plugins workers="4">
        <plugin id="condor" type="runner" load="galaxy.jobs.runners.condor:CondorJobRunner" />
    </plugins>
    <destinations default="gateway_1x">

        <destination id="gateway_1x" runner="dynamic">
            <param id="function">gateway</param>
            <param id="tmp_dir">True</param>
            <param id="type">python</param>
        </destination>

        <destination id="gateway_1_5x" runner="dynamic">
            <param id="function">gateway_1_5x</param>
            <param id="tmp_dir">True</param>
            <param id="type">python</param>
        </destination>

        <destination id="gateway_2x" runner="dynamic">
            <param id="function">gateway_2x</param>
            <param id="tmp_dir">True</param>
            <param id="type">python</param>
        </destination>

    </destinations>
</job_conf>
```

With the expectation that it will submit to first, then second, then third.

In practice I see that it is only resubmitted once:

<details>
<summary>Galaxy logs</summary>
<pre>
galaxy.tools DEBUG 2018-12-14 10:25:18,113 [p:60895,w:1,m:0] [uWSGIWorker1Core1] Validated and populated state for tool request (12.920 ms)
[...]

PR #7832, opened 2019-04-26
Fix failure in resubmitting job more than once

pcm32:
This addresses the problem of sequential resubmissions failing as described in #7820. It does so by avoiding to add failed jobs to the watched items.


mvdbeek (approved):
Yep, thanks for figuring this out!

hexylena:
I wonder if this addresses my issue in https://github.com/galaxyproject/galaxy/issues/7118

pcm32:
@erasche I'm afraid this fix will only affect runners under the cli implementation, and it seems to me that condor is a direct runner (based on what I see from your destination code at least).

pcm32:
But it might be worth checking if the condor runner takes care of removing failed jobs from the watched items list (which was the problem in this case).

hexylena:
Ahh you're right. made a note here mostly to remind myself to check. Thanks!

PR #7834, opened 2019-04-26
Remove recursion from ``expand_nested_tokens()``

nsoranzo:
Also add test for recursive nested token.

Follow-up on https://github.com/galaxyproject/galaxy/pull/7823 .


mvdbeek (approved):
Thanks for thinking this through, I was definitely making this too complicated.

Issue #7816, opened 2019-04-24
Pulsar for Remote Data Iteration


jmchilton:
Implement Pulsar Side of a Two-Container Pod Job:

- Docker Container pops up.
- Reads MQ message BASE64 encoded from command-line (mirroring Mesos work).
  - https://github.com/galaxyproject/pulsar/blob/master/pulsar/scripts/submit_util.py#L22
  - ?
- Stage files - "preprocess".
- Writes command-line to a place.
- Waits for exit.
- Unstages... completes.
- Do we need a pseudo "Manager" that just writes the file and waits for an exit code written somewhere or something like that? That might make the above three operations more natural.

Implement Galaxy Runner for Two-Container Pod Job:

- Base on Pulsar MQ Runner. Build up message.
- Base 64 encode the message and create a job with a 2 container POD. The biocontainer pod with command that just waits for a file and then executes the command and the Pulsar pod with argument just being the BASE64 encoded message I suppose.

Implement Extended Metadata Collection with Pulsar:

- Seems entirely testable here...
- WIP on extended metadata for other runners (https://github.com/galaxyproject/galaxy/pull/7058).

xref #7058  #4030

jmchilton:
Mostly resolved and in 20.01 currently.

PR #7835, opened 2019-04-28
Kubernetes 2-container pod staging.

### LINKED ISSUE #4030: Kubernetes Staging
I've been thinking about how to use Kubernetes to schedule containers instead of scheduling cluster jobs. I know @pcm32 has already put a good deal of work into this - but it relies on having a shared file system available inside of the Kubernetes cluster. There are good use cases for that - but I think there are likely cases where such volumes are unavailable or impractical.

Pulsar is very good at taking a big json blob produced by Galaxy, communicating with Galaxy to stage the data, run a job, and stage out. However - Pulsar isn't available in containers generally - so we cannot leverage it from inside of the containers. So I think that leaves two paths forward - we either have the container we schedule do the staging and then schedule another container or we have separate containers for staging and running and instruct Kubernetes to process them in serial.

I have not found any APIs that seemed geared toward doing the first of these options. I don't see a general way to create some disk and then dynamically create a pod or job to attach to that disk and wait for it. Maybe I'm missing something though and this can be done cleanly - any ideas?

The second route seems more promising and it seems like we could pretty easily leverage the existing work on the Kubernetes runner. My idea is to use the ``batch/v1`` endpoint still - but we break the execution into two "init containers" and one normal container. The first init container would stage the data in, the second init container would run the job on the data, and the normal container would then stage the data out. The staging containers are pretty straight forward and can leverage a lot of the work that we have already done on the Mesos runner mode for Pulsar - but it is much simpler since we don't have to worry about state as much - Kubernetes takes care of that - this just looks like a normal job.

Both jobs themselves and init containers are kind of newish additions to Kubernetes but I did the following little example with my minikube setup to see if they work together:

```
$ cat galaxy-job.yml
apiVersion: batch/v1
kind: Job
metadata:
  name: galaxy-job-uuid-here
  labels:
    jobgroup: galaxy-job
spec:
  template:
    metadata:
      name: galaxy-job
      labels:
        jobgroup: galaxy-job
    spec:
      restartPolicy: Never
      initContainers:
      - name: pulsarstagin
        image: busybox
        command: ["sh", "-c", "echo 'staging in'; echo 'input' > /work-dir/input"]
        volumeMounts:
        - name: workdir
          mountPath: "/work-dir"
      - name: runcommand
        image: busybox
        command: ["sh", "-c", "echo 'running tool'; cat /work-dir/input; echo 'output' > /work-dir/output"]
        volumeMounts:
        - name: workdir
          mountPath: "/work-dir"
      containers:
      - name: pulsarstageout
        image: busybox
        command: ["sh", "-c", "echo 'staging out'; ls /work-dir/"]
        volumeMounts:
        - name: workdir
          mountPath: "/work-dir"
      volumes:
      - name: workdir
        emptyDir: {}
```

And the final output was:

```
2017-05-05T13:32:49.164483895Z staging out
2017-05-05T13:32:49.164545285Z input
2017-05-05T13:32:49.164550002Z output
```

So it seemed like these features worked together and saw each other disk and such.

Ping @pcm32 - does this seem like the right approach to you for staging? Is there a more obvious way to have one container schedule another with disk setup by the first?

jmchilton:
The tests included here are really very manual and brittle tests, but I'm not sure how to do much better without significantly more work. See documentation in the file docstring for more information.

Pulsar work on this (part of Pulsar 0.10.0) includes:
- https://github.com/galaxyproject/pulsar/pull/176
- https://github.com/galaxyproject/pulsar/pull/178

Fixes #4030. Implements first two big sections outlined in https://github.com/galaxyproject/galaxy/issues/7816.  Builds on #7854.

### DIFF
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -14,21 +14,17 @@ from galaxy.jobs.runners import (
     AsynchronousJobState,
     JobState
 )
+from galaxy.jobs.runners.util.pykube_util import (
+    DEFAULT_JOB_API_VERSION,
+    ensure_pykube,
+    Job,
+    job_object_dict,
+    Pod,
+    produce_unique_k8s_job_name,
+    pull_policy,
+    pykube_client_from_dict,
+)
 from galaxy.util.bytesize import ByteSize
-
-# pykube imports:
-try:
-    from pykube.config import KubeConfig
-    from pykube.http import HTTPClient
-    from pykube.objects import (
-        Job,
-        Pod
-    )
-except ImportError as exc:
-    KubeConfig = None
-    K8S_IMPORT_MESSAGE = ('The Python pykube package is required to use '
-                          'this feature, please install it or correct the '
-                          'following error:\nImportError %s' % str(exc))
 
 log = logging.getLogger(__name__)
 
@@ -43,15 +39,15 @@ class KubernetesJobRunner(AsynchronousJobRunner):
 
     def __init__(self, app, nworkers, **kwargs):
         # Check if pykube was importable, fail if not
-        assert KubeConfig is not None, K8S_IMPORT_MESSAGE
+        ensure_pykube()
         runner_param_specs = dict(
-            k8s_config_path=dict(map=str, default=os.environ.get('KUBECONFIG', None)),
+            k8s_config_path=dict(map=str, default=None),
             k8s_use_service_account=dict(map=bool, default=False),
             k8s_persistent_volume_claims=dict(map=str),
             k8s_namespace=dict(map=str, default="default"),
             k8s_galaxy_instance_id=dict(map=str),
             k8s_timeout_seconds_job_deletion=dict(map=int, valid=lambda x: int > 0, default=30),
-            k8s_job_api_version=dict(map=str, default="batch/v1"),
+            k8s_job_api_version=dict(map=str, default=DEFAULT_JOB_API_VERSION),
             k8s_supplemental_group_id=dict(map=str),
             k8s_pull_policy=dict(map=str, default="Default"),
             k8s_run_as_user_id=dict(map=str, valid=lambda s: s == "$uid" or s.isdigit()),
@@ -71,11 +67,7 @@ class KubernetesJobRunner(AsynchronousJobRunner):
         """Start the job runner parent object """
         super(KubernetesJobRunner, self).__init__(app, nworkers, **kwargs)
 
-        if "k8s_use_service_account" in self.runner_params and self.runner_params["k8s_use_service_account"]:
-            self._pykube_api = HTTPClient(KubeConfig.from_service_account())
-        else:
-            self._pykube_api = HTTPClient(KubeConfig.from_file(self.runner_params["k8s_config_path"]))
-
+        self._pykube_api = pykube_client_from_dict(self.runner_params)
         self._galaxy_instance_id = self.__get_galaxy_instance_id()
 
         self._run_as_user_id = self.__get_run_as_user_id()
@@ -122,18 +114,11 @@ class KubernetesJobRunner(AsynchronousJobRunner):
 
         # Construction of the Kubernetes Job object follows: http://kubernetes.io/docs/user-guide/persistent-volumes/
         k8s_job_name = self.__produce_unique_k8s_job_name(job_wrapper.get_id_tag())
-        k8s_job_obj = {
-            "apiVersion": self.runner_params['k8s_job_api_version'],
-            "kind": "Job",
-            "metadata": {
-                    # metadata.name is the name of the pod resource created, and must be unique
-                    # http://kubernetes.io/docs/user-guide/configuring-containers/
-                    "name": k8s_job_name,
-                    "namespace": self.runner_params['k8s_namespace'],
-                    "labels": {"app": k8s_job_name}
-            },
-            "spec": self.__get_k8s_job_spec(ajs)
-        }
+        k8s_job_obj = job_object_dict(
+            self.runner_params,
+            k8s_job_name,
+            self.__get_k8s_job_spec(ajs)
+        )
 
         # Checks if job exists and is trusted, or if it needs re-creation.
         job = Job(self._pykube_api, k8s_job_obj)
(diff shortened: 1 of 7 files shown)


test/integration/test_kubernetes_staging.py
```diff
@@ -0,0 +1,117 @@
+"""Integration tests for Kubernetes pod staging.
+
+This case is way more brittle than the other integration tests distributed with Galaxy.
+This is because it requires Kubernetes, RabbitMQ, and the test itself needs to know what
```
mvdbeek, line 4: Probably makes some sense for the test to spin up a rabbitmq docker container ?

jmchilton: @mvdbeek Indeed, I started down that road but I got worried about something... whether the Kubernetes containers would be able to talk to the MQ container maybe... but that seems odd because you'd probably bind its ports to a port on the host. It is a good idea and would make this more robust I think.

@VJalili did you make any progress on MQ tests for Galaxy/Pulsar? If you're still working on or planning to work on that task, it might be worth shifting focus to this context and make this change here as @mvdbeek suggests. I'm not sure if this is a better start than what I had previously pointed you at or if it just more moving pieces to get right.

test/unit/jobs/job_conf.sample_advanced.yml
```diff
@@ -125,6 +125,13 @@ runners:
     # higher value (in seconds) (or `None` to use blocking connections).
     #amqp_consumer_timeout: None
 
+  pulsar_k8s:
+    # Use a two-container Kubernetes to run jobs - one for staging and one
+    # for tool execution.
+    load: galaxy.jobs.runners.pulsar:PulsarKubernetesCoexecutionJobRunner
+    galaxy_url: http://docker.for.mac.localhost:8080
```
mvdbeek, line 132: Since we figured that out for IEs as well, it seems that the recommendation for modern docker for mac (starting 18.03) is `host.docker.internal` (https://docs.docker.com/docker-for-mac/networking/)
```suggestion
    galaxy_url: http://host.docker.internal:8080
```

test/integration/test_kubernetes_staging.py
```diff
@@ -0,0 +1,117 @@
...
+the test host IP address will be relative to a container inside Kubernetes in order to
+communicate job status updates back.
+
+For this reason, this test will only work out of the box currently with Docker for Mac,
```
mvdbeek, line 8: Surely not needed for this PR, but https://kind.sigs.k8s.io/ looks interesting for testing purposes.

test/integration/test_kubernetes_staging.py
```diff
@@ -0,0 +1,117 @@
...
+import tempfile
+
+from base import integration_util  # noqa: I100,I202
+from base.populators import skip_without_tool
+from .test_containerized_jobs import MulledJobTestCases  # noqa: I201
+from .test_job_environments import BaseJobEnvironmentIntegrationTestCase  # noqa: I201
+
+TOOL_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir, os.pardir, 'tools'))
+AMQP_URL = os.environ.get("GALAXY_TEST_AMQP_URL", "amqp://guest:guest@localhost:5672//")
+GALAXY_TEST_KUBERNETES_INFRASTRUCTURE_HOST = os.environ.get("GALAXY_TEST_KUBERNETES_INFRASTRUCTURE_HOST", "docker.for.mac.localhost")
```
mvdbeek:
```suggestion
GALAXY_TEST_KUBERNETES_INFRASTRUCTURE_HOST = os.environ.get("GALAXY_TEST_KUBERNETES_INFRASTRUCTURE_HOST", "host.docker.internal")
```

jmchilton:
Rebased and then updated with some improvements:

- Rev pulsar-galaxy-lib for improved error messages.
- Rev to a newer container - one that actually logs.
- More use of host.docker.internal to improve testing robustness.
- Add documentation for overriding container to use.

mvdbeek (approved):
Works like a charm! Can you just resolve the conflict ?

jmchilton:
Rebased with a conflict fix, thanks @mvdbeek. I have WIP on a generalization of this that just uses a single container with embedded Pulsar and normal Conda resolution if no container is configured. It might change the interface here a little but we are early in the release cycle and it would be good to get this merged before the GCC I think so I think this should still be merged. Thanks again!

mvdbeek:
Sounds good! I think it's actually quite nice to have this split, it'd potentially allow targeting the staging to a lower resource / cheaper target, right ?

jmchilton:
@mvdbeek Yeah... so there are really three options here:

- Stage + BioContainer as separate containers. (This PR)
- Stage, Dependency Resolution in same container. (Next PR)
- Stage + Dependency Resolution in separate containers.

I'm not sure that the third option would work with this WIP as is... but Pulsar certainly does have a nauseating number of knobs that can be tweaked and may be possible. We don't actually use init containers yet - so both containers would come online immediately and you'd be charged for the more expensive container the whole time. If we could do stage-in in an init-container and stage-out / biocontainer in separate containers though that might make sense to save staging in/out costs.

Update: xref init containers (https://kubernetes.io/docs/concepts/workloads/pods/init-containers/)

mvdbeek:
Yeah, that makes sense, I was guessing that was your idea after writing my comment ....

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -4,7 +4,6 @@
 
 import logging
 import math
-import os
```
mvdbeek: os is used again https://github.com/galaxyproject/galaxy/blob/5598305f07ddde4873a0943c9e5cea9a1f32f94f/lib/galaxy/jobs/runners/kubernetes.py#L161, which was added in the meantime.

VJalili:
@jmchilton could you please elaborate how data are staged?

jmchilton:
@VJalili 

Pulsar has many options for data staging - they can be configured on a per file basis - based on file type (input, output, config, reference, tool file, etc..) and/or based on regexes, globs, and prefixes. Each target specification has a corresponding transfer type - see https://pulsar.readthedocs.io/en/latest/galaxy_conf.html#data-staging.

The transfer types that expect a permanent Pulsar server (``transfer``) or that expect the target directory to exist after a Galaxy initiated setup (``copy``) will not work - but everything that is done by Pulsar should work:
- ``remote_transfer`` - Pulsar fetches files from the Galaxy job files API.
- ``remote_rsync_transfer`` - Pulsar uses rsync to copy the files. (untested in this context). 
- ``remote_copy`` - Pulsar can copy the file if K8S is somehow configured to have the target directory mounted.
- ``rewrite`` - just rewrite the paths - again assumes K8S is somehow configured to have the target directory mounted.

The default and easiest to use is ``remote_transfer`` though. This is what the test cases use.

PR #7837, opened 2019-04-29
Use markupsafe.escape instead of cgi.escape

mvdbeek:
This silences some deprecation warnings on python 3.


mvdbeek:
@galaxybot test py3

nsoranzo (approved):

mvdbeek:
@galaxybot test py3

Issue #7849, opened 2019-04-29
tool_script.sh contains unnecessary replicas of conda activation blocks and conda creation failure


qiagu:
A typical Conda Env activation block looks like below:
```
[ "$CONDA_DEFAULT_ENV" = ".../database/dependencies/_conda/envs/mulled-v1-..." ] ||
 .....
done ;
```
One `tool_script.sh` in my local Galaxy contains 9 replicas, which take ~30 s for just running them, whereas the real python script takes < 2 s to be finished (a simple script). After removing all the duplicates, the tool_script still runs great and, of course, much faster.

mvdbeek:
This is an artefact of how we build dependencies for tools that contain multiple requirements,
and how `$CONDA_DEFAULT_ENV` is being set.
We used to actually generate a script that looks like 
`[ "$CONDA_DEFAULT_ENV" = "mulled-v1-..." ] || ...`
which would only evaluate to false if the environment hadn't been activated yet (activating a conda environment sets `$CONDA_DEFAULT_ENV` to the environment name).

qiagu:
Thanks @mvdbeek 
Another issue was noticed during the process of creating conda env.
If failure occurs, conda tries to create multiple ENVs for each package, but those ENVs don't help the tool.
For example,
If `conda create -y --override-channels --channel iuc --channel bioconda --channel conda-forge --channel defaults --name mulled-v1-13a015120bc6b0cdf6195aaf6c27ae8c8572b2cf8c6d30faaac86e34ea04d736 python=3.6 pandas=0.23.4 numpy=1.15.1` fails,
Galaxy tries to create `__python@3.6__`, `__pandas@0.23.4__` and `__numpy@1.15.1__` following the failure.

PR #7853, opened 2019-04-30
[19.01] Use just the env name when checking ``CONDA_DEFAULT_ENV``

mvdbeek:
This will still write the conda activate part out multiple times,
but fix the check that would otherwise lead to repeated activation
of the conda env. Fixes the slowness reported in https://github.com/galaxyproject/galaxy/issues/7849


nsoranzo:
s/CONDA_DEFAULT_NAME/CONDA_DEFAULT_ENV/ in the PR title?

mvdbeek:
Yep, corrected it now.

nsoranzo (approved):

PR #7854, opened 2019-04-30
Allow YAML/JSON job configuration.

jmchilton:
- Allow loading job configuration from YAML/JSON files.
- Allow specifying job configuration from the ``galaxy.yml`` via a nested dictionaries.
- Actual validation of YAML specifications (either stand-alone or in galaxy.yml) via pykwalify.
- Fix up some terminology (at least from a user facing standpoint) that @natefoo and I have wanted to cleanup for a while. In particular they are runners not plugins now (since everything is a plugin) and they are environments not destinations (since destination sounds like a physical location, when really you are defining a lot of context besides where the job runs - you define how it runs also and how it is configured - so environment feels more appropriate).

XML remains the default and better documented format but there is work on the documentation piece here. I've translated hundreds and hundreds of lines job_conf.xml.sample_advanced to YAML and written/refactored a bunch of tests around that to ensure both XML and YAML are tested. But that sample file is too large and I'd like to get a commitment from the team that this is the right approach before I continue to translate that file (it has TODOs about what is left). @natefoo has committed to updating the admin focused docs if this is merged (no time table). I'd say once that is done we should replace the XML sample file with the YAML one in progress but I don't think there is any rush.

Rather than mixing XML parsing and object configuration logic - I've separated JobConfiguration into a two step process. The first step is loading the XML into a new dictionary-based format (outlined by @natefoo and myself here https://github.com/galaxyproject/galaxy/issues/5148#issuecomment-444208836) and the second step is to load the configuration from that. I then extended that process to allow loading that dictionary from a YAML file or from the application's configuration directly.

I needed this for #7816, referencing external YAML files in the job configuration XML is gross (see file_actions for instance right now), I don't want to do more of that. #7816 requires embedding a full (though light-weight) pulsar configuration into a job destination - it'd be better to do this with nested dictionaries in YAML than trying to rewrite the XML into YAML.


lib/galaxy/jobs/__init__.py
```diff
@@ -171,54 +300,61 @@ def __init__(self, app):
...
-            self.__parse_job_conf_xml(tree)
+            if 'job_config' in self.app.config.config_dict:
+                job_config_dict = self.app.config.config_dict["job_config"]
+            else:
+                job_config_file = self.app.config.job_config_file
+                if '.xml' in job_config_file:
+                    tree = load(job_config_file)
+                    job_config_dict = self.__parse_job_conf_xml(tree)
+                else:
+                    import yaml
```
mvdbeek: That should move to the top I guess

test/integration/delay_job_conf.yml
```diff
@@ -0,0 +1,17 @@
+runners:
```
mvdbeek, line 1: Much nicer!

mvdbeek (commented):
Definitely like this a lot. I'm just wondering if we could move the actual parsing into jobs.parser.xml / jobs.parser.yml as we do for tools ? Not necessarily for this PR, but might be nice for consistency.

lib/galaxy/jobs/__init__.py
jmchilton: Rebased with this fix, thanks for the catch.

mvdbeek (approved):

jmchilton:
> I'm just wondering if we could move the actual parsing into jobs.parser.xml / jobs.parser.yml as we do for tools ? Not necessarily for this PR, but might be nice for consistency.

jobs.parser.yml would just be yaml.safe_load... but I do agree. Especially because there is really just too much going on in galaxy.jobs - queue definitions, job configuration parsing and setup, JobWrapper/TaskWrapper.

mvdbeek:
> jobs.parser.yml would just be yaml.safe_load

Awesome :).

jmchilton:
That was easier than I thought to get in, thanks for the merge @mvdbeek!

PR #7855, opened 2019-04-30
[19.01] Fix workflow extraction for jobs whose JobToDatasetOutputAssociation references a discovered dataset

mvdbeek:
To test this run a tool that generates discovered data sets twice, with the second run taking the first output as input.


natefoo:
Thanks @mvdbeek !

PR #7857, opened 2019-04-30
[19.01] Fix the job config's shared _get_default() method

natefoo:
[As noticed](https://github.com/galaxyproject/galaxy/commit/355b48c361e68a7d9fbe9fe86de7aa2313e5d608#r33361410) by @jmchilton, these because both the handler and destination configs use this method, the changes in #7091 allowed Galaxy to start without a default destination, which would raise an exception when attempting to run jobs.

### DIFF
--- a/lib/galaxy/util/handlers.py
+++ b/lib/galaxy/util/handlers.py
@@ -62,8 +62,7 @@ class ConfiguresHandlers(object):
                         [x.strip() for x in handler.get('tags', self.DEFAULT_HANDLER_TAG).split(',')]
                     )
             self.default_handler_id = self._get_default(
-                    self.app.config, config_element, list(self.handlers.keys()),
-                    required=self.deterministic_handler_assignment)
+                self.app.config, config_element, list(self.handlers.keys()), required=False)
 
     def _init_handler_assignment_methods(self, config_element=None):
         self.__is_handler = None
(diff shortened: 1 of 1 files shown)


jmchilton (approved):

lib/galaxy/util/handlers.py
```diff
@@ -61,7 +61,8 @@ def _init_handlers(self, config_element):
                         handler_id,
                         [x.strip() for x in handler.get('tags', self.DEFAULT_HANDLER_TAG).split(',')]
                     )
-            self.default_handler_id = self._get_default(self.app.config, config_element, list(self.handlers.keys()))
+            self.default_handler_id = self._get_default(
+                    self.app.config, config_element, list(self.handlers.keys()), required=False)
```
nsoranzo:
```suggestion
                self.app.config, config_element, list(self.handlers.keys()), required=False)
```

lib/galaxy/util/handlers.py
```diff
@@ -115,7 +116,7 @@ def _set_default_handler_assignment_methods(self):
     def _parse_handler(self, handler_id, handler_def):
         pass
 
-    def _get_default(self, config, parent, names, auto=False):
+    def _get_default(self, config, parent, names, auto=False, auto_include=None, required=True):
```
nsoranzo: `auto_include` is not used?

natefoo: Whoops, thanks

bgruening (approved):

PR #7858, opened 2019-04-30
Remove close option from workflow gear menu

guerler:
This PR removes the 'Close' option from the workflow gear dropdown menu. This option is redundant and not-placed well, users can navigate away either way, in which case a similar warning message is shown. It would be good to follow-up and improve the generic warning message but that can be done in another PR.


martenson:
Awesome. Shrink the menus, drop the clutter!

dannon:
@guerler I like this, thanks.  Wherever we can minimize these additional options and just make 'the right thing' obvious and/or part of the normal navigation we should do that, just like this.

PR #7859, opened 2019-04-30
Fix flag for library dataset option in data selector

guerler:


client/galaxy/scripts/mvc/ui/ui-select-content.js
```diff
@@ -7,16 +7,17 @@ import Ui from "mvc/ui/ui-misc";
 import Select from "mvc/ui/ui-select-default";
 import { getGalaxyInstance } from "app";
 
-/** Batch mode variations */
-var Batch = { DISABLED: "disabled", ENABLED: "enabled", LINKED: "linked" };
+/** Batch mode letiations */
+let Batch = { DISABLED: "disabled", ENABLED: "enabled", LINKED: "linked" };
```
dannon: We should prefer `const` for situations like this where there is no rebinding of the variables.

martenson:
Why are we restricting anons here?

client/galaxy/scripts/mvc/ui/ui-select-content.js
guerler: Thanks @dannon.

guerler:
@martenson, as of right now, `http://127.0.0.1:8080/api/libraries/f2db41e1fa331b3e/contents` is not exposed for anon-users. It triggers a forbidden exception.

martenson:
We should fix that instead of restricting this feature. I will have a look
tomorrow.

client/galaxy/scripts/mvc/ui/ui-select-content.js
dannon: @guerler Sure, no worries.  I just opened https://github.com/galaxyproject/galaxy/pull/7860 that'll fix the existing occurrences of this and will lint it moving forward, too.

guerler:
@martenson, I removed that particular setting from this PR. The remaining PR is still relevant, it fixes the overall handling of that option. Thank you for the review.

martenson:
~~You might need to rebase because of https://github.com/galaxyproject/galaxy/pull/7860~~

You actually need to switch everything to `const` :)

PR #7866, opened 2019-05-01
[19.05] Favorite filtering only for logged-in users

martenson:
we are already hiding the favorite button on tools, so this makes no sense to render


client/galaxy/scripts/entry/panels/tool-panel.js
```diff
@@ -43,26 +43,29 @@ var ToolPanel = Backbone.View.extend({
             default_genome: config.default_genome,
             default_extension: config.default_extension
         });
+        let panel_buttons = [this.upload_button];
```
dannon:
```suggestion
        const panel_buttons = [this.upload_button];
```

PR #7870, opened 2019-05-01
[19.01] Fix bug preventing handler runner plugin handling.

jmchilton:
Broken in https://github.com/galaxyproject/galaxy/pull/7091 somewhere.

With some more job config test cases from https://github.com/galaxyproject/galaxy/pull/7854 that would have covered this.

### DIFF
--- a/test/unit/jobs/test_job_configuration.py
+++ b/test/unit/jobs/test_job_configuration.py
@@ -230,7 +230,7 @@ class JobConfXmlParserTestCase(unittest.TestCase):
 
         assert env_dest.env[3]["execute"] == "module load javastuff/2.10"
 
-    def test_hanlder_runner_plugins(self):
+    def test_handler_runner_plugins(self):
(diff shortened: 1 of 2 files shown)


test/unit/jobs/test_job_configuration.py
```diff
@@ -230,6 +230,16 @@ def test_env_parsing(self):
 
         assert env_dest.env[3]["execute"] == "module load javastuff/2.10"
 
+    def test_hanlder_runner_plugins(self):
```
bgruening:
```suggestion
    def test_handler_runner_plugins(self):
```

jmchilton:
Thanks @bgruening !

bgruening (approved):

natefoo:
Just a thought for the future, we should probably add a flag to `galaxy-main` so that you can set this option for dynamic handlers. Not important at the moment, however.

PR #7694, opened 2019-04-08
Fix pulsar remote metadata with metadata_strategy == 'directory'.

jmchilton:

PR #7873, opened 2019-05-02
Add file missing from #7694

jmchilton:


mvdbeek:
Want to target 19.05, since that's where the test was added ?

martenson:
I tried to change the base via github but it won't work here, need force push or sth. ;)

jmchilton:
I'll backport - no big deal.

PR #7878, opened 2019-05-02
Look and Feel in Dataset Libraries

### LINKED ISSUE #7764: Fix Look and Feel in Dataset Libraries
Although we made progress aligning the Data Library's Look and Feel to the rest of Galaxy, there are still some remaining issues. Here, our goal is to remove most if not all custom-styles and hard-coded style tags, and use Bootstrap4 classes instead.

I recommend to start by fixing the code format of the template functions by inserting indentions and using multi-line strings (``) as shown in #7234, instead of joining string arrays. This will restore readability of the template functions. Remember to run `make client-watch` to automatically repack the client-code after making changes.

Here is a list:

 - [ ] Fix `Create New Library` modal. Currently inputs are lined up in one row, should be replaced with a regular input form e.g. like login or tool form.
See: https://getbootstrap.com/docs/4.0/components/forms/

- [ ] Replace Toastr messages with regular alert messages. Toastr messages overlap with underlying options and in some cases produce a third layer i.e. Library View + Modal + Toastr.
See: https://getbootstrap.com/docs/4.0/components/alerts/

- [ ] Use regular small buttons for `Save`, `Cancel`, `Delete`, `Edit` and `Manage`.
See: https://getbootstrap.com/docs/4.0/components/buttons/

 - [ ] Fix `Adding datasets from your history` modal content, `Select all`, `Unselect all` spacing, use table for list, use default style select fields, fix overall appearance.

- [ ] Fix `Create History Collection from Datasets`, use regular input elements e.g. radio button and select field, remove custom styling.

- [ ] Remove `How many libraries per page do you want to see?` alert and use an editable field instead of using the `change` link.

- [ ] Use regular fonts and highlighting for table-rows in main library list view

ping @bgruening, @martenson.

inkuzmin:
I create this pull request to initiate a discussion on the "Fix Look and Feel in Dataset Libraries" issue (fix #7764). The following changes reflect my interpretation of the original tasks and might be not complete/correct. Please, let me know what could be improved.

> I recommend to start by fixing the code format of the template functions by inserting indentions and using multi-line strings

- [x] Templates were refactored with the use of template literals (` `` `).
- [x] Indentation was introduced.
- [x] JS comment were replaced with HTML comments.
- [x] Except for a few special cases lines were wrapped at 120 characters length.

> Fix Create New Library modal.

- [x] Regular Bootstrap 4 form styling was used.

Concerns:
- Placeholders repeat labels, it might be better to use hints/examples as a placeholders.

> Replace Toastr 

Skipped for the moment due to the concerns originally posted by @martenson in https://github.com/galaxyproject/galaxy/issues/7764#issuecomment-485824102

> Use regular small buttons for `Save`, `Cancel`, `Delete`, `Edit` and `Manage`.

- [x] `btn`, `btn-secondary`, `btn-sm ` classes were used.

> Fix `Adding datasets from your history` modal content, `Select all`, `Unselect all` spacing, use table for list, use default style select fields, fix overall appearance.

- [x] Add a note when selected history is empty.
- [x] Use regular Bootstrap 4 component styling for select inputs.
- [x] Use table with selectable rows for the list (no sorting, would it be useful?).
- [x] Remove `Select all`, `Unselect all`.
- [x] Use a checkbox in the table's header to toggle selection.

> Fix `Create History Collection from Datasets`, use regular input elements e.g. radio button and select field, remove custom styling.

- [x] Use Bootstrap 4 component styling for inputs.
- [x] Use definitions list for list with types explanation.
- [x] Use regular `select` inputs.

> Remove How many libraries per page do you want to see? alert and use an editable field instead of using the change link.

- [x] Add number input
- [x] Remove change link
- [x] Adjust margins

The same changes were introduced for the `library/list#folders`.

> Use regular fonts and highlighting for table-rows in main library list view

- [x] Not sure I didn't misinterpret this task, but I noticed that the text at the table at `library/list#folders` was modified in its styling from the initial after selecting and unselecting the row it belongs to. That was fixed.


martenson:
@inkuzmin Thanks a bunch! Could you please resolve the conflict in library-foldertoolbar-view.js?

inkuzmin:
@martenson sure!

martenson:
@galaxybot test this

martenson:
This is very, very nice! 

What do you think about this feedback @inkuzmin ?

* I think rows should have constant height, no matter what the content.
Galaxy___Data_Libraries
* item per page selector does not respond to mouse well
item-select
* Searching should not affect the toolbar layout (watch the width)
width

Stretch goal:
* Create libraries (and folders) 'inline' instead of a in a modal (this helps e.g. visually impaired people and keeps context focused)

martenson:
This is a relevant failing selenium test: https://jenkins.galaxyproject.org/job/docker-selenium/6291/testReport/junit/selenium_tests.test_library_contents/LibraryContentsTestCase/test_import_dataset_from_history/

inkuzmin:
@martenson, thank you for the review!

> I think rows should have constant height, no matter what the content.

There are three columns that might affect row's height currently: `name`, `description` and `state`.
I can fix the width of the last one (or prevent line breaks in some other way), so that all buttons would be at the same line.
But how would you recommend to handle names and descriptions, would ellipsizing be a solution?

> item per page selector does not respond to mouse well

Indeed, I can reproduce it. I'll fix it.

> Searching should not affect the toolbar layout (watch the width)

Good point! Also just noticed, that it doesn't look nice if I would resize the window to be narrower.

> Create libraries (and folders) 'inline' instead of a in a modal (this helps e.g. visually impaired people and keeps context focused)

Great idea! I can try it out.

martenson:
>how would you recommend to handle names and descriptions

I was thinking of targeting some reasonable width (1200px?), subtracting the 'fixed columns' and then forcing the text to appear as `this is a long line of text that will be sho...` when it is too long. 

We are already doing something along those lines here: https://github.com/galaxyproject/galaxy/blob/dev/client/galaxy/scripts/mvc/library/library-libraryrow-view.js#L244

Ideally we would have fluid layout and change it based on viewport but that would require much more work and frankly libraries are not an important target for this I think.

inkuzmin:
@martenson yes, I get it now! I'll commit a fixed version.

guerler:
Thanks a lot for working on this @inkuzm. The dataset selection modals look so much better now and it is easier to select datasets. Thank you. A few things I noticed are that when selecting a dataset the corresponding table row is highlighted in yellow/brown. Can you change that to the success color (~green/primary-success)? Another thing I noticed is that the 'Add datasets' button wraps in 2-lines and the add datasets/create folder icons are not consistent. Maybe we should use `+(plus-icon) Datasets` and `+(plus-icon) Folder` instead.
Another thing to look at is the usage of text in the button-top-bar e.g. 'items per page, 1 libraries shown, 1 total' and the checkboxes followed by text. Can we can experiment with trimming/removing, or using text-muted for the 'items-per-page' info. We can also make sure that the checkboxes are always the last items listed in the toolbar and not inbetween buttons.
Thanks again for this prompt and very useful contribution.

inkuzmin:
@guerler thank you for your feedback! I added items to my task list.

Could you explain this idea, please, I am not sure I've got what you mean:
>  we can experiment with trimming/removing

inkuzmin:
@guerler do you know if we have a reference table to reproduce?

There are some examples I found:

1. A table at `admin/data_types`. It has stripped rows, darker header (#d8d8d8) and no lines between rows.
2. A table at `admin/display_applications`. It has stripped (but with different colors) rows, transparent header and lines between rows.
3. A table at `admin/users`. No stripes, lavender "current" row, lighter header (#f8f9fa), lines between rows, a footer with actions. Hovered or selected rows are not highlighted. Click on the row wouldn't select the row.
4. A table at `Install new tools`. The same as 3, but header color is #e9ecef.
5. A table at `Manage whitelist`. No lines, transparent header, selected/hovered rows are not highlighted, no action on click on a row.

All other tables I noticed fall into one of the examples' categories listed above. Styles of the 3rd example look like the most commonly adopted (and the closest to the sample at `docs/galaxy-tables.md`). The questions this arises are:

- Do we want rows to be highlighted on hover (when they are selectable)?
- Do we want rows to be selectable on click (when they *are* selectable)?
- Do we want rows to be highlighted when selected (if yes, then should we use bootstrap class or custom one (`current`?); if bootstrap, should it be [`table-active`, `table-primary` or `table-success`](https://getbootstrap.com/docs/4.0/content/tables/#contextual-classes))?
- Should we replace sorting icon (`<span class="fa fa-sort-alpha-asc"></span>`) to `<span class="sort-arrow">↑</span>`?
- Should we capitalise header titles?

martenson:
my 2c:

>The questions this arises are: 

* Yes.
* Yes.
* Yes. The specific class should be designed in such a way that the color tetrad (unselected/unselected:hover/selected/selected:hover) is working well together.
* I actually think we should use `fa-sort-alpha-asc` with alphabetical sorting everywhere else instead. It is superior to just an arrow imho.
* Yes.

inkuzmin:
@martenson,

> The specific class should be designed in such a way that the color tetrad (unselected/unselected:hover/selected/selected:hover) is working well together.

I put together a [proposal with bootstrap classes](https://jsfiddle.net/inkuzmin/8Lbv7kwh/), do you think it would be fine to have it like this?

guerler:
@inkuzmin the example looks pretty good and I like the blue, but please keep in mind however that colors are specified in https://github.com/galaxyproject/galaxy/blob/dev/client/galaxy/style/scss/theme/blue.scss. Variations of these colors reached through darken/lighten operations are ok too.

martenson:
@inkuzmin like Sam says, the goal here is to bring the style closer to Galaxy's brand and even though there are no such exact UI elements (like library) used in Galaxy yet, the color palette should be closer to what Sam linked

inkuzmin:
This is how the same classes would style tables within the application. The `table-primary` color  #87a0b6 is inferred from `primary` by [`theme-color-level` function](https://github.com/twbs/bootstrap/blob/7626f8b17d59fcb65b1f3e9f5ea4f4c8522d3f40/scss/_functions.scss#L80).

table-colors

And here it is with `table-success` (not sure if it is semantically accurate to use `success` color to highlight selected rows).

table-success

martenson:
I like it (the blue/gray version). Maybe the text in the highlighted row could have been light for higher contrast, but that might be just a personal preference when to do that.

martenson:
@inkuzmin Very nice job with the inline library and folder creation and the color theme. I think it is safe to remove the commented out modal code.

inkuzmin:
@martenson thanks! I will remove it.

A note to myself:
- [x] Have to handle the case, when "Add library/folder" button is clicked multiple times.
- [x] Access control icons' colors are now inconsistent (gray in the list of libraries and black in the list of folders).

Also a pagination control is at the top of the list of libraries and at the bottom of the list of folders. Would it make sense to have it at the same place across all views?

Also, an issue with a flickering tooltip. Would it be fine to use `pointer-events: none` to prevent it?

UPD. "Close pull request" button was clicked accidentally, never mind, please.

inkuzmin:
@martenson,

> I think rows should have constant height, no matter what the content.
- [x] Rows heights' variability was minimized. Description is capped now, buttons are aligned in one line.

> item per page selector does not respond to mouse well
- [x] Pagination control was granularized, now input is not rerendered on change.

> Searching should not affect the toolbar layout (watch the width)
- [x] Pagination control was moved to the bottom, so nothing affects toolbar's layout now.
- [x] Filter icon was added to the filter input.

@guerler,

> Maybe we should use +(plus-icon) Datasets and +(plus-icon) Folder
- [x] Done.

> Another thing to look at is the usage of text in the button-top-bar e.g. 'items per page, 1 libraries shown, 1 total'
- [x] Nouns are inflected appropriately now.

> using text-muted for the 'items-per-page' info. 
- [x] `text-muted` class was added.

> We can also make sure that the checkboxes are always the last items listed in the toolbar and not inbetween buttons.
- [x] Checkboxes were moved.

mvdbeek:
@galaxybot test this

martenson:
the errors in linting form circleci (you can get to them if you click on 'detail')

```
/home/circleci/repo/client/galaxy/scripts/mvc/library/library-librarytoolbar-view.js
  4:8   error  '_l' is defined but never used                 no-unused-vars
  5:10  error  'Toast' is defined but never used              no-unused-vars
  6:8   error  'mod_library_model' is defined but never used  no-unused-vars
```

martenson:
@inkuzm looks very good, I pushed one commit with small changes


for future reference when in in `client/` you can run `make client-format` to apply our styles to things

guerler:
I agree it looks really good. Thanks a lot for fixing this view so quickly. A last thing I noticed is that the footer text "n items per page, k shown, m total" should probably just be "n per page, m total", since the smaller of the two numbers is always equivalent to k if I understand this label correct. Removing k will remove some redundancy.

martenson:
@guerler that is a good point

~~The changing of items per page is a bit tricky though, since it does not change on lost focus and neither on `Enter`. Try entering the value with keyboard.~~

edit: on second testing it triggers in both these cases

I think better usability could be reached if the arrows change the value, but the value gets applied on lost focus/`Enter` instead.

edit: this would also address the element 'running away' if you click the up/down ticks on the page size input

martenson:
@guerler the only case where that is not true is the last page, but I am willing to sacrifice it
Galaxy___Data_Libraries

nsoranzo:
@martenson There shouldn't be a page 4 on the last image you posted, right?

martenson:
@nsoranzo it always renders the next page button so we do not have 'jumping' interface. It is 'disabled' though. In these new styles it is less visible :/

nsoranzo:
I see (more or less)...

martenson:
I pushed another commit to remove the overrides used for pagination styling.

inkuzmin:
@martenson,

> edit: on second testing it triggers in both these cases

No, you are right, there was an error at the folders' pagination control implementation (`click` event was used instead of `change` event). Thanks for noticing.

> I think better usability could be reached if the arrows change the value, but the value gets applied on lost focus/Enter instead.

Fixed to work this way.

@guerler,

> "n per page, m total"

Good point, fixed.

martenson:
the failing selenium tests passed locally, restarting them

martenson:
I am not sure what is up here. @inkuzmin you can check the test artifacts including screenshots and tracebacks here: https://jenkins.galaxyproject.org/job/docker-selenium/6381/artifact/6381-test-errors/

However the only maybe-relevant thing I found there is this in the console:
```
 {
  "source": "javascript", 
  "message": "http://127.0.0.1:8569/static/scripts/bundled/generic.bundled.js?v=1557398798 0:383193 Uncaught TypeError: Cannot read property 'focus' of undefined", 
  "timestamp": 1557400650970, 
  "level": "SEVERE"
 }
```

inkuzmin:
Ah! Got it finally, thanks @martenson!

martenson:
@inkuzmin hehe, that was a helpful test :)

martenson (approved):
Great job. Thanks for accommodating all our review comments!

inkuzmin:
My pleasure. Thank you for the feedback and insightful comments!

bgruening:
@inkuzmin you simply rock!!!

PR #7880, opened 2019-05-02
[19.05] Allow typing e in float param field

### LINKED ISSUE #7836: Cannot type `e` and `E` inside float param fields in tool form
Which means scientific notation cannot be used. Copy/pasting a string in scientific notation into the field works though, so I suppose it's a JavaScript issue.

WilliamHolden:
Fixes #7836. Add keycode to allow typing "e" in float param. This will allow for typing in scientific notation.


martenson:
@WilliamHolden would you like to try writing a test for this? (or for float param field in general)

WilliamHolden:
@martenson Yes, I will look into it. I also plan on making a larger change to how the key events are handled. I am going to switch to use the "keypress" event instead of "keydown". 

"keydown" gets the keycode which is unreliable since different browsers use different key codes. "keypress" is more reliable since it gets the character code which is the same across all browsers.

For example, The "-" key is allowed in chrome because there is a check for that key code (189). However the code for "-" is different in firefox (173) so there is a bug that you can't type the minus key right now for float parameters if you're using firefox.

There are other bugs I noticed too such as these characters being allowed "@#$%^&*(". These characters have the same key code as numeric values so they are allowed when they shouldn't be.

Switching to "keypress" should fix most of these issues, but I will have to be careful not to break anything since the code is old.

guerler:
@WilliamHolden thanks for looking into this issue. It would be nice to make this input field more resilient and consistent in a follow-up. Unfortunately it looks like "keypress" has been deprecated. https://stackoverflow.com/questions/52882144/replacement-for-deprecated-keypress-dom-event

dannon:
Seems like it might be most robust to filter `oninput` or `onchange`?

WilliamHolden:
Thanks for the information @guerler. That is unfortunate. It looks like we are also using Keyboard​Event​.which which is also deprecated so we will have to remove that from the codebase as well: 
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which 

@dannon I vote for ```oninput``` since ```onchange``` only fires after the input loses focus. Let me know if anyone opposes this or has any other input.

dannon:
@WilliamHolden Sounds good to me, give it a shot and see how well it works in practice IMO.  An added benefit here is that it should actually now validate on pasted-in values, etc., as well.

WilliamHolden:
If you guys have any free time, please review this pull request into my own repo. I have tested the code and it is working, but I still need to do a lot more testing before it's ready for a pull request into the base repo. Any feedback is appreciated

https://github.com/WilliamHolden/galaxy/pull/1/files

martenson:
I think this is an OK fix for the release.  Thanks @WilliamHolden!

Could you please open https://github.com/WilliamHolden/galaxy/pull/1/files against dev of this repo?So more people can see it, we have tests etc.

PR #7883, opened 2019-05-03
Allow type_source to reference a collection input in a repeat 

mvdbeek:
I was reviewing the local-only commits I had on our prod server, this one allows `type_source="inputs_0|input"` for 
```
    <repeat name="inputs" title="Input Collections" min="2">
        <param name="input" type="data_collection" label="Input Collection" />
    </repeat>
```
The current workaround would be to create one static input, but that seems unnecessary. 

WIP while I add a test.

### DIFF
--- a/lib/galaxy/tools/execute.py
+++ b/lib/galaxy/tools/execute.py
@@ -192,9 +192,17 @@ class ExecutionTracker(object):
             for key, value in input_dict.items():
                 if key == input_name:
                     return value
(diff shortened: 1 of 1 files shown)


jmchilton:
https://github.com/galaxyproject/galaxy/pull/7885 will hopefully fix this failing integration test, sorry about that.

martenson:
integration test restarted

PR #6891, opened 2018-10-17
Visualization for free style image annotation

anuprulez:
This pull request adds an image annotator for `.jpg` and `.png` files. It can be used to draw freestyle shapes using different coloured pens. It comes with several other options like `redo`, `undo`, `erase`, `download` and so on. The downloaded images include the annotated image and an another image with just the annotations (both as `.png`). This visualization is built [paperjs](https://github.com/paperjs/paper.js) and jquery.

image

Annotated (downloaded) image:

original-with-annotations 4

Only annotations (downloaded):

only-annotations 4

PR #7888, opened 2019-05-03
OpenLayers maps as Galaxy visualisation

anuprulez:
This pull request adds a visualisation based on [OpenLayers](https://openlayers.org/) for showing `GeoJSON` and `Shapefile` datatypes. The visualisation provides the following features:

- Zoom in/out using buttons and a slider
- Fullscreen mode
- Map scale
- Show grid lines and longitude and latitude information
- Draw and edit lines, polygons and circles
- Color picker
- Export as PNG

geojson_image

shapefile_image

I am not sure if this can be ported to 19.05 branch.

Edit: This viz can be developed as a "charts" plugin but a chart visualisation is refreshed after every option change which would not be optimal for this interactive map visualisation. If it is possible not to refresh the viz with each option change, I can do this viz as a charts plugin.

Thank you!

ping @bgruening @annefou @yvanlebras @Alanamosse


bgruening:
@anuprulez I guess you should rename this vis to `openlayer` instead of `ol`. If I'm not mistaken @dannon had some new client build magic so that we do not need to include all the JS libraries into the repo. Dannon can you help here?

guerler:
@dannon should we wait until the new build system is in place before merging this PR?

dannon:
I don't mind merging this now and then repackaging it in the next few weeks prior to GCC as an example, if you'd want to do that.

hexylena:
TODO (me): minify JS

martenson:
@erasche the webpack based approach is a nice cookie cutter: https://github.com/galaxyproject/galaxy/pull/8227/files#diff-88b6b89437beb3e93df42225658e8c8e

hexylena:
Thanks!

martenson:
My points from https://github.com/galaxyproject/galaxy/pull/6891#pullrequestreview-279727039 apply here as well. @anuprulez

dannon:
This looks good to me, and should build for Galaxies on changes now.  I know there were a couple of issues mentioned here and I'll commit to addressing those bugs during the freeze.

(there is no difference between the previous and final squashed/forcepushed 
8a66345  other than dropping the >1mb build artifact)

bgruening (approved):

bgruening:
Thanks @dannon and @anuprulez! GIS people will be so happy :)

annefou:
Yes I confirm! We are looking forward to use it. Thank you so much!!!

yvanlebras:
Sure! Groovy! Thank you so much @anuprulez @dannon and all helpers!

anuprulez:
@dannon @bgruening Please wait for this PR to be finished before merging. I still need to fix a couple of issues like exporting the annotated `shapefile` map. I have not been able to fix it. In the older version (with `mako` file) of the viz, all the features I added were working but with this new version as `charts`, few features seem to break. I tried to fix them but was not able to. 

Reload of the viz file with the change of every option is another issue I am not sure how to solve. For example, every option change invokes the reload of the viz. Because of this, users will not be able to annotate with more than one geometry shape as the previous annotations would be lost and the same file will be loaded in the viz.

One good thing is now the viz works with the latest packages of `OpenLayer`, `Jszip`, `Proj4` ...

dannon:
@anuprulez My plan was to merge the core visualization and then, as mentioned, I will address the bugs next week so that we could go ahead and freeze, while getting this into 19.09.  Is that okay with you?

anuprulez:
@dannon then it is ok :)

PR #7891, opened 2019-05-03
Raise detailed exception for broken citation

nsoranzo:
For an empty citation, this will result in logging:
```
Exception: Invalid citation of type 'doi' with content 'None': 'NoneType' object has no attribute 'strip'
```

instead of:
```
AttributeError: 'NoneType' object has no attribute 'strip'
```

xref https://github.com/galaxyproject/planemo/issues/929

### DIFF
--- a/lib/galaxy/managers/citations.py
+++ b/lib/galaxy/managers/citations.py
@@ -65,7 +65,11 @@ def parse_citation(elem, directory, citation_manager):
     if not citation_class:
         log.warning("Unknown or unspecified citation type: %s" % citation_type)
         return None
-    return citation_class(elem, directory, citation_manager)
+    try:
+        citation = citation_class(elem, directory, citation_manager)
+    except Exception as e:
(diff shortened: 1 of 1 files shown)


mvdbeek (approved):

jmchilton (approved):

nsoranzo:
@jmchilton While working to enhance galaxy/tools/linters/citations.py in galaxy-lib , I noticed that in commit cb9fc095fc435250e09b21ac7dafd2cb2f85447f you added the possibility for a bibtex citation to use a `file` attribute, but this is not allowed in the [XSD](https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-citations-citation) and there are no tests for it.
I'm not aware of any tool using this attribute, so should we remove the code allowing it or fix the XSD?

jmchilton:
Hmm... I'm fine with removing the option. Seems undocumented at best.

nsoranzo:
Thanks, I've opened https://github.com/galaxyproject/galaxy/pull/7911 to remove the attribute.

PR #7892, opened 2019-05-03
Fix Kubernetes test for job I/O separation.

jmchilton:
I assumed that is what caused this to change?


mvdbeek (approved):
Makes a lot of sense!

PR #7894, opened 2019-05-04
Fix docs for `expect_num_outputs` and add `count` attribute for test `output` and `element` tags

bernt-matthias:
If a tool returns a paired collection the test for the expected number of outputs seems to be wrong if the forward and reverse element are assigned explicitly. It seems that the contents of the pair are counted as additional data elements (the reported number is larger by 2 than the expected number). Surprisingly, it works for automatically detected pair elements. Also because of this inconsistency I think that this needs fixing.

I added a test to show this. 

I started to debug a bit (but it seems to me that the problem is deeper than just counting .. API?). In this line https://github.com/galaxyproject/galaxy/blob/98b84d6750ff5ed38dd4fc49b718119a9360efbd/lib/galaxy/tools/verify/interactor.py#L828

`data_list` has 4 elements (should be 2) and `data_collection_list has 1 (which is correct)

The superficial elements of the former are 
- `(u'pair_output|__part__|forward', {u'visible': False, u'create_time': u'2019-05-04T10:45:02.818074', u'file_size': 0, u'file_ext': u'txt', u'id': u'4b455cfa602b54cb', u'misc_info': None, u'hda_ldda': u'hda', u'state': u'new', u'metadata_dbkey': u'?', u'misc_blurb': u'queued', u'peek': None, u'update_time': u'2019-05-04T10:45:02.908499', u'data_type': u'galaxy.datatypes.data.Text', u'tags': [], u'deleted': False, u'history_id': u'cb227cec8ca83994', u'genome_build': u'?', u'hid': 1, u'metadata_data_lines': None, u'uuid': u'2beffaeb-6f48-42ba-bdd2-1c33cb32de7c', u'history_content_type': u'dataset', u'name': u'output_filter', u'model_class': u'HistoryDatasetAssociation', u'output_name': u'pair_output|__part__|forward', u'purged': False}), ` and 
- `(u'pair_output|__part__|reverse', {u'visible': False, u'create_time': u'2019-05-04T10:45:02.818495', u'file_size': 0, u'file_ext': u'txt', u'id': u'9564811bfb69c81e', u'misc_info': None, u'hda_ldda': u'hda', u'state': u'new', u'metadata_dbkey': u'?', u'misc_blurb': u'queued', u'peek': None, u'update_time': u'2019-05-04T10:45:02.908506', u'data_type': u'galaxy.datatypes.data.Text', u'tags': [], u'deleted': False, u'history_id': u'cb227cec8ca83994', u'genome_build': u'?', u'hid': 2, u'metadata_data_lines': None, u'uuid': u'1aade6a4-321d-437e-8ae5-9e4cb9fca2a3', u'history_content_type': u'dataset', u'name': u'output_filter', u'model_class': u'HistoryDatasetAssociation', u'output_name': u'pair_output|__part__|reverse', u'purged': False}), `

Also I don't understand why in the following line accesses `$forward` and not `$paired_output.forward`

https://github.com/galaxyproject/galaxy/blob/98b84d6750ff5ed38dd4fc49b718119a9360efbd/test/functional/tools/collection_creates_pair.xml#L3

This would fail for multiple output pairs...


mvdbeek:
> it works for automatically detected pair elements

I think it's the other way round, automatically detected pairs are not behaving "correctly".
Discovered datasets are not included in the tool response (the API response to a POST to /api/tools) and only linked to the job after the job completed running (it could be any number of discovered datasets after all). What is being verified here is the output datasets and dataset collections in the tool response, so this doesn't include discovered outputs.

> `data_list` has 4 elements (should be 2)

I am counting 2. The format is a tuple with output name and output. So 2 is what you were expecting ?

> Also I don't understand why in the following line accesses `$forward` and not `$paired_output.forward`

It's a shortcut, no harm in that ?

> This would fail for multiple output pairs...

With multiple you mean list:pair output ? True, but you can always use the full prefix (that's clearer anyway).

I guess my suggestion would be to not count the data / dataset collection count from the tool response, but instead to wait for the job that is referenced in the tool_response, and then get the jobs' output datasets. And I'd also separate out verifying the number of datasets from the number of dataset collections. That should make it much clearer what went wrong.

bernt-matthias:
> > it works for automatically detected pair elements
> 
> I think it's the other way round, automatically detected pairs are not behaving "correctly".
> Discovered datasets are not included in the tool response (the API response to a POST to /api/tools) and only linked to the job after the job completed running (it could be any number of discovered datasets after all). What is being verified here is the output datasets and dataset collections in the tool response, so this shouldn't include discovered outputs.

I guess my wording was wrong here. What I meant is that it works for discovered data sets as expected (as I expect it :) ).

> > `data_list` has 4 elements (should be 2)
> 
> I am counting 2. The format is a tuple with output name and output. So 2 is what you were expecting ?

The list that I posted contains only the the 2 elements that I would not expect in the list of data sets (`data_list`), in this case the two contents of the pair. 

> > Also I don't understand why in the following line accesses `$forward` and not `$paired_output.forward`
> 
> It's a shortcut, no harm in that ?
> > This would fail for multiple output pairs...
> 
> With multiple you mean list:pair output ? True, but you can always use the full prefix (that's clearer anyway).

Fine with me :) 

> I guess my suggestion would be to not count the data / dataset collection count from the tool response, but instead to wait for the job that is referenced in the tool_response, and then get the jobs' output datasets. And I'd also separate out verifying the number of datasets from the number of dataset collections. That should make it much clearer what went wrong.

I don't understand the technical details of this part of your answer. The main point is the the count should be equal to the number of (top level) outputs, i.e. the outputs and output_collections of the output section, that pass the filters. Currently for pairs  this fails sometimes because elements of paired collections are also counted. 

The main point of the PR is to create awareness of the problem (it seems we agree that there is one), but unfortunately I don't know how to fix it.

mvdbeek:
> The main point is the the count should be equal to the number of (top level) outputs, i.e. the outputs and output_collections of the output section, that pass the filters

I disagree, I think it should be the number of datasets produced. If they are top-level or not doesn't matter, just like it doesn't matter if they're hidden or not. Otherwise there's no way to ascertain the correct number of datasets is present in a collection (The element count is a bit ambiguous in that context, I think it's important to know exactly how many datasets have been produced).

bernt-matthias:
From the tool xml documentation: 

- `expect_num_outputs`: Assert the number of outputs this test should produce, this is useful to ensure ``filter`` directives are implemented correctly.

Since the meaning of "outputs" is undefined here the first part of the sentence is OK with your and my interpretation. But I think only my interpretation can test for the correctness of filters. 

Furthermore the number of elements in a (1st level) collection can be ensured with the count parameter of `output_collections`

- `count`: Number of elements in output collection

For completeness see also here: 
- https://github.com/galaxyproject/galaxy/pull/6931
- https://github.com/galaxyproject/galaxy/pull/7359

mvdbeek:
> But I think only my interpretation can test for the correctness of filters.

Why ? My interpretation is how Galaxy works today.

> Furthermore the number of elements in a (1st level) collection can be ensured with the count parameter of `output_collections`

So there's no way to tell how many datasets are in a nested collection, right ? 
That's why I think we need to count all datasets.

bernt-matthias:
> > But I think only my interpretation can test for the correctness of filters.
> 
> Why ? My interpretation is how Galaxy works today.

Sorry, maybe I was wrong. 

I often forget the there are only data sets and some are hidden when they get included in collections. From my intuition  it seemed unnecessary to count them again since the number of elements of collections can be tested separately. 

I guess both ways are equivalently useful -- but currently its inconsistent. 

Just as a side note, a real test for the correctness of a specific filter would be possible if we could test for the absence of outputs (as far as I know this is not possible at the moment?). 

Good to have this discussion (I learned / remembered something and maybe we have some new ideas).

> > Furthermore the number of elements in a (1st level) collection can be ensured with the count parameter of `output_collections`
> 
> So there's no way to tell how many datasets are in a nested collection, right ?

As far as I know.

mvdbeek:
> but currently its inconsistent.

I think the inconsistency is with not counting discovered outputs, which is a bug to me. That should be easy to fix, and we have to do this independently of whether we want to include hidden datasets or not.

I guess my proposition here is to discriminate between visible and not-visible datasets, so you don't have to specify the hidden datasets if you don't want to.
Sth. like
```
<test datasets_visible="2" datasets_hidden="2">
```
? 
Which also means we can leave the current counting scheme as is and not break tests that aren't actually broken.

> Just as a side note, a real test for the correctness of a specific filter would be possible if we could test for the absence of outputs (as far as I know this is not possible at the moment?).

I also think that's not possible at this point, but we can get all outputs from the API, we can implement this.
Do you have a specific tool test syntax in mind ?

bernt-matthias:
I think I got the idea of your proposition. The hidden data sets are all that are within collections (at any nesting level). Together with checking the count for each collection this seems useful. 

So I checked the IUC for tools that have `output_collections` and use `expect_num_outputs`. As far as I see 4 out of 5 would fail:

- tools/egsea/egsea.xml
- tools/gffcompare/gffcompare.xml
- tools/limma_voom/limma_voom.xml
- tools/fastp/fastp.xml

it seems that those follow my intuition (ie count the top level items). As it seems (to me) that both interpretations are equally expressive (but yours is closer to what Galaxy actually does) the big question is which path to follow. 

> I also think that's not possible at this point, but we can get all outputs from the API, we can implement this. Do you have a specific tool test syntax in mind ?

Maybe just allow

`<output name="..." compare="absent">`

`<output name="..." compare="present">`

i.e. also allow empty `value`.

jmchilton:
```
<output name="..." compare="absent">

<output name="..." compare="present">
```

Something like this would be really nice!

I really created a mess when I added this feature with such an unspecific behavior. I apologize to everyone for the mess.

bernt-matthias:
I added a more systematic test tool that helped me to understand what is going on and suggest to change the docs for `expect_num_outputs`. 

So it appears to me that the only thing missing is a possibility to assert the count of discovered datasets in the following cases:

1. discovered data sets
2. discovered elements of a nested collection

My first thought was to add a `count` attribute to `output` and `element` tags. For collections something like a `total_count` attribute might be nice as well.

bernt-matthias:
I guess the current state could be in 22.01. For the followup an issue could be opened.

bernt-matthias:
I added a `count` attribute for `output` and `element`

- `test` - `output`
- `test` - `collection` - `element` (also nested `element`)

that allows to check the number of discovered elements. For collection elements the number of contained child collections/datasets is counted -- ie. like for output_collection the count considers only direct children.

works also for statically defined outputs and collection elements (where the count is 1).

TODO:

- [ ] more documentation in xsd doc
- [ ] Lint for `discover_outputs` where the corresponding test elements do not specify `count`

bernt-matthias:
Since we are at it now .. we could also think about useful extensions, e.g. allowing `delta`, `min`, `max` either 

- by adding `count_delta`, `count_min`, `count_max` attributes or
- by allowing also `min:max` notation for the count attribute (or count+-delta)

On hindsight this might have been more elegant (and convenient for the tool developer) for the all the other places where we recently introduced min,max,delta... :(

lib/galaxy/tool_util/verify/interactor.py
```diff
@@ -174,10 +176,14 @@ def verify_output(self, history_id, jobs, output_data, output_testdef, tool_id,
             raise AssertionError(f"Output {name}: {str(e)}")
 
         primary_datasets = attributes.get('primary_datasets', {})
-        if primary_datasets:
-            job_id = self._dataset_provenance(history_id, hid)["job_id"]
-            outputs = self._get(f"jobs/{job_id}/outputs").json()
-
+        job_id = self._dataset_provenance(history_id, hid)["job_id"]
+        outputs = self._get(f"jobs/{job_id}/outputs").json()
+        found_outputs = 0
```
mvdbeek: I'd rename this to found_datasets, to discriminate a dataset from a tool output.
```suggestion
        found_datasets = 0
```

lib/galaxy/tool_util/verify/interactor.py
```diff
@@ -174,10 +176,14 @@ def verify_output(self, history_id, jobs, output_data, output_testdef, tool_id,
...
+        for output in outputs:
+            if output["name"] == name or output["name"].startswith(f'__new_primary_file_{name}|'):
+                found_outputs += 1
```
mvdbeek:
```suggestion
                found_datasets += 1
```

lib/galaxy/tool_util/verify/interactor.py
```diff
@@ -174,10 +176,14 @@ def verify_output(self, history_id, jobs, output_data, output_testdef, tool_id,
...
+        if expected_count is not None and expected_count != found_outputs:
```
mvdbeek:
```suggestion
        if expected_count is not None and expected_count != found_datasets:
```

lib/galaxy/tool_util/verify/interactor.py
```diff
@@ -174,10 +176,14 @@ def verify_output(self, history_id, jobs, output_data, output_testdef, tool_id,
...
+            raise AssertionError(f"Output '{name}': expected to have '{expected_count}' elements, but it had '{found_outputs}'")
```
mvdbeek:
```suggestion
            raise AssertionError(f"Output '{name}': expected to have '{expected_count}' datasets, but it had '{found_datasets}'")
```

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -1095,8 +1095,16 @@ specified and instead a series of assertions is made about the output.
     </xs:attribute>
     <xs:attribute name="expect_num_outputs" type="xs:integer">
       <xs:annotation>
-        <xs:documentation xml:lang="en">Assert the number of outputs this test
-should produce, this is useful to ensure ``filter`` directives are implemented correctly.</xs:documentation>
+        <xs:documentation xml:lang="en">Assert the number of statically defined outputs this test
```
```suggestion
        <xs:documentation xml:lang="en">Assert the number of statically defined items (datasets and collections) this test
```

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -1095,8 +1095,16 @@ specified and instead a series of assertions is made about the output.
...
+should produce. Where each `data` or `collection` tag that is listed in the
```
```suggestion
should produce. Each `data` or `collection` tag that is listed in the
```

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -1095,8 +1095,16 @@ specified and instead a series of assertions is made about the output.
...
+outputs section is a statically defined output and adds one to this count.  For
+instance a statically defined pair adds 3: 1 for the collection and 1 for each
```
```suggestion
instance a statically defined pair adds a count of 3; 1 for the collection and 1 for each
```

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -1095,8 +1095,16 @@ specified and instead a series of assertions is made about the output.
...
+of the two data elements.  Dynamically defined outputs (``discover_datasets``)
```
```suggestion
of the two datasets.  Dynamically defined output datasets (using ``discover_datasets`` tag)
```

mvdbeek (approved):
Thanks for documenting that so clearly. `expect_num_outputs` is absolutely insane and we can't really change it, but it's great that at least we've spelled out how this is counted.

bernt-matthias:
One thing that I was wondering about: `__parse_test_attributes` (which now parses `count`) is not only called from the functions 
that parse `element` and `output` tags, but also from `__parse_extra_files_elem`.

Would the count be of use for this as well? 
.

lib/galaxy/tool_util/verify/interactor.py
```diff
@@ -194,6 +194,8 @@ def verify_output(self, history_id, jobs, output_data, output_testdef, tool_id,
         attributes = output_testdef.attributes
         name = output_testdef.name
 
+        expected_count = attributes.get("count", None)
+
```
mvdbeek:
```suggestion
        expected_count = attributes.get("count")
```

lib/galaxy/tool_util/verify/interactor.py
```diff
@@ -964,13 +972,13 @@ def get_element(elements, id):
...
     def verify_elements(element_objects, element_tests):
         expected_sort_order = {}
-
         eo_ids = [_["element_identifier"] for _ in element_objects]
         for element_identifier, element_test in element_tests.items():
             if isinstance(element_test, dict):
                 element_outfile, element_attrib = None, element_test
             else:
                 element_outfile, element_attrib = element_test
+            expected_count = element_attrib.get("count", None)
```
mvdbeek:
```suggestion
            expected_count = element_attrib.get("count")
```

lib/galaxy/tool_util/verify/interactor.py
```diff
@@ -981,10 +989,16 @@ def verify_elements(element_objects, element_tests):
 
             element_type = element["element_type"]
             if element_type != "dataset_collection":
+                found_outputs = 1
```
mvdbeek: Maybe rename this to element_count ?

lib/galaxy/tool_util/verify/interactor.py
```diff
@@ -981,10 +989,16 @@ def verify_elements(element_objects, element_tests):
...
                 verify_dataset(element, element_attrib, element_outfile)
             else:
                 elements = element["object"]["elements"]
+                found_outputs = len(elements)
                 verify_elements(elements, element_attrib.get("elements", {}))
+            if expected_count is not None and expected_count != found_outputs:
+                raise AssertionError(
+                    f"Element '{element_identifier}': expected to have '{expected_count}' elements, but it had '{found_outputs}'"
```
mvdbeek:
```suggestion
                    f"Element '{element_identifier}': expected to have {expected_count} elements, but it had {found_outputs}"
```
I don't think we need to quote numbers.

test/functional/tools/expect_num_outputs.xml
```diff
@@ -0,0 +1,326 @@
+<tool id="expect_num_outputs" name="expect_num_outputs" version="1.0.0">
```
mvdbeek:
```suggestion
<tool id="expect_num_outputs" name="expect_num_outputs" version="1.0.0" profile="22.05">
```
Would be good to do that consistently for new tools.

mvdbeek (approved):

PR #7896, opened 2019-05-06
[19.05] Add local namespaced resolver to default container resolvers

mvdbeek:
The mulled container resolvers are building containers that are
namespaced to "local", so I think it makes a lot of sense to
include local in the defaults. This prevents planemo from
building the same image over and over again when run with
`--mulled_containers` / `--biocontainers`.


mvdbeek:
As a sidenote I think we should use these options when testing IUC tools, but disable the `biocontainer` namespace so that we always test the latest version available in Conda.

jmchilton:
I think this makes sense, thanks!

PR #7904, opened 2019-05-06
[19.05] Fix application of history default permissions to anonymous histories carried over upon login

natefoo:
This was originally addressed many years ago when the behavior on login was updated in ca6ba6e01128aecb970efdefe50492dc5bc47826. I believe the intent was to ensure that permission defaults would only be set on histories that were created while anonymous but as it was, they would only be applied to users who had no previous sessions, which would only happen on registration/creation.

Histories created upon login have their permission defaults set in the new history creation method, histories that are already owned by the user are not modified.


martenson:
Can we target 19.05 with this? Or we can backport, but I think it is worth it.

natefoo:
Retargeted to 19.05.

PR #7905, opened 2019-05-07
Implement "xref" tag on the tool structure (with visualisation)

ValentinMrcn:
## Purpose:

We (I and @hmenager ) make this proposal to add a "**Tool Reference**" tag on the tool structure.
The tool reference aims to add unique reference to the tool description as contained in a registry, here we use the ELIXIR bio.tools registry.

We had in mind to add "bio.tools" id, but we generalized this to a tool reference.
So the XML will be like:
```
 <tool_refs>
    <tool_ref type="bio.tools">my_tool_uniq_bt_id</tool_ref>
 </tool_refs>
```

We also add a visualization of the reference(s) with a link to the tool reference website (only bio.tools is managed for the moment). 
We don't propose here a new field visualization, we used the same as "Citations".

Finally we have exposed tool_refs ("content" and "reftype") to the API. 

---

### Comments and questions are more than welcome!
 
---
## Changed Files:

- client/galaxy/scripts/components/ToolRefs.vue **[Add]**
_To print toolrefs list and links (only available for type bio.tools)_
- client/galaxy/scripts/mvc/tool/tool-form-base.js **[Edit]**
_To call Toolrefs.vue_
- doc/schema_template.md **[Edit]**
_To add "Toolrefs" to the doc_ 
- lib/galaxy/managers/tool_refs.py **[Add]**
_New manager object for Toolref_
- lib/galaxy/config.py **[Edit]**
_Create the manager object from corresponding tool_refs.py_
- lib/galaxy/tools/_ _init__.py **[Edit]**
_Add the Toolref parser_
- lib/galaxy/tools/xsd/galaxy.xsd **[Edit]**
_Add tool_refs element and ToolRefs type in xsd shema_
- lib/galaxy/tools/parser/xml.py & yaml.py **[Edit]**
_Add "tool_refs" management in the xml and yaml parsers_
- lib/galaxy/webapps/galaxy/api/tools.py **[Edit]**
- lib/galaxy/webapps/galaxy/buildapp.py **[Edit]**
_To expose tool_refs in API_

---
**To test** the changes you can add `<tool_refs>` tags to any xml.


nsoranzo:
See also https://github.com/bgruening/galaxytools/pull/856#discussion_r280836782 for a possible future extension to get useful info from bio.tools .

hmenager:
Thanks @nsoranzo. Should we rather discuss this in the issue or here?
Anyway, I agree with https://github.com/bgruening/galaxytools/pull/856#discussion_r280873302. This PR aims exactly at getting an xref in the tools, so that we can retrieve other tool metadata from other repositories/registries if needed.
This is of course just a suggestion, any change in the technical details is fine as long as we get the crossrefs eventually...

nsoranzo:
> Should we rather discuss this in the issue or here?

Probably on a new issue, but I just wanted to point out that basically the addition of such a feature was independently discussed just a couple of days ago.

bgruening:
That is super cool! One item on my list that I can kill off :)

The only question I have is the name. Should we name it `xref`? So we could also cite/reference data?

What about `types` that are not know to Galaxy and can not be rendered nicely? Can we implement a fallback, a simple list `type_name: type_value` that is displayed in such cases? This would enable us to annotate our tools in the future with new refs and old Galaxy will still see it.

The from @nsoranzo referenced license-tag idea should maybe be a separate tag, or do we envision something like:

```xml
 <xrefs>
    <xref type="licsense">https://opensource.org/licenses/BSD-3-Clause</tool_ref>
 </xrefs>
```

ping @galaxyproject/iuc, please provide input here!

Thanks a lot for working on this!

@hmenager can I remind you on the EDAM PRs that are still open :) Maybe we can renew them and include bio.tools ids.

ValentinMrcn:
Thanks for your kind feedback!

I think ping @galaxyproject/iuc is not working

wm75:
This looks really promising. Two quick thoughts:
- should we include a `version` attribute in the `xref`/`tool-ref` (or whatever tag), e.g.:

```
 <tool_refs>
    <tool_ref version=@TOOL_VERSION@ type="bio.tools">my_tool_uniq_bt_id</tool_ref>
 </tool_refs>
```
similar to what we have in a `requirement` tag?

- license could become an attribute, too, if you think of it as something that's not constant for a tool.
  Otherwise, I would argue that the license info should be captured in the referred to tool index rather than in the tool wrapper.

jmchilton:
I agree with @bgruening about changing the tag name to xrefs in general. I'd promote license to a top-level thing and probably try to align with CWL and schema.org (https://www.commonwl.org/user_guide/17-metadata/index.html) in whatever way we can.

ValentinMrcn:
Thanks a lot @bgruening @nsoranzo @jmchilton @wm75 for your feedback!

Would you all agree to the syntax proposal illustrated with the example below? If so I will change the code of this PR accordingly. One important point for discussion is that the version attribute might not be relevant when it comes to bio.tools, because bio.tools does not have tool version-specific information.

Thank you for your advice!

```xml
<tool id="Summary_Statistics1" name="Summary Statistics" version="1.1.2">
  <description>for any numerical column</description>
  <edam_topics>
    <edam_topic>topic_2269</edam_topic>
  </edam_topics>
  <xrefs>
    <xref type="bio.tools" version="1.1">summary_statistics</xref>
    <xref type="license">https://opensource.org/licenses/BSD-3-Clause</xref>
  </xrefs>
  <requirements>
    <requirement type="package" version="2.9.4">rpy2</requirement>
  </requirements>
  ...
```

Thumbs up and comment are welcome.

bwlang:
This seems reasonable to me. 
Should there be some kind of automated link checking on the server before rendering the link?
(no link is better than a broken link?)
I'm imagining some breaking URL change in the future that requires a ton of tool updates to fix.
Maybe that's an argument for a dedicated license "else branch" like the bio.tools bit - then an update to the vue fixes all affected tools. Might also end up being more consistent. Are there other potentially common xref targets?

I'd vote agains having the version number in the xref unless it's necessary.  It's already in the tool tag and more places = potential for inconsistency (though that could be mediated with a macro)

nsoranzo:
I think that for the license @jmchilton meant something like:

```xml
<tool id="Summary_Statistics1" name="Summary Statistics" version="1.1.2">
  <description>for any numerical column</description>
  <license>https://opensource.org/licenses/BSD-3-Clause</license>
  <edam_topics>
    <edam_topic>topic_2269</edam_topic>
  </edam_topics>
  <xrefs>
    <xref type="bio.tools" version="1.1">summary_statistics</xref>
  </xrefs>
  <requirements>
    <requirement type="package" version="2.9.4">rpy2</requirement>
  </requirements>
  ...
```

John?

Also, when speaking about the license, we need to make clear if we are speaking about the Galaxy tool (the "wrapper") or the license of the (possibly multiple and diversely licensed) underlying tools (the requirements).
I would interpret the one specified above as the wrapper license, while the licenses of the requirements could probably be recovered automatically from bio.tools.

bgruening:
I like your proposal @ValentinMarcon.

Let's have the license discussion in another thread, it would be great if @bwlang could lead this discussion. I just brought that up because of you :) - as a representant for all industry partners. I know licensing of tools, in general, is a huge pain point in the industry.

> I would interpret the one specified above as the wrapper license. I'm not sure the wrapper needs while the licenses of the requirements could probably be recovered automatically from bio.tools.

The one I brought up was more the latter. I'm personally not interested in licensing of the wrapper, but for a specific user-base it is hugely important that a tool exposes its underlying license, also in workflows.
I do not think Galaxy should use bio.tools for this.
* bio.tools is not for all scientific domains where Galaxy is used.
* Licenses needs to be exposed in isolated networks as well.

I do think our tooling should use bio.tools to get license information where available and explicitly annotate tools. We will have soon a github repo, where the community (bio.tools, containers, bioconda etc) collects those kind of metadata and this will make tooling also easier, especially if we have this bio.tools identifier.

bwlang:
I agree that we should take up the license question elsewhere if it needs more discussing.
I think this xref implementation is not a bad idea (brittleness questions I raised above aside). It would a bummer to have this get delayed due to the tangential license questions that have been raised.
I have started a [new topic on that subject](https://github.com/galaxyproject/galaxy/issues/8006).

nsoranzo:
@galaxybot test this

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -47,7 +47,8 @@ A ``data_source`` tool contains a few more relevant attributes.
         <xs:element name="macros" type="Macros" minOccurs="0"/>
         <xs:element name="edam_topics" type="EdamTopics" minOccurs="0"/>
         <xs:element name="edam_operations" type="EdamOperations" minOccurs="0"/>
-        <xs:element name="requirements" type="Requirements" minOccurs="0"/>
+        <xs:element name="xrefs" type="xrefs" minOccurs="0" />
+	<xs:element name="requirements" type="Requirements" minOccurs="0"/>
```
nsoranzo: I think the change of indentation here was by mistake.

lib/galaxy/webapps/galaxy/api/tools.py
```diff
@@ -403,6 +403,14 @@ def citations(self, trans, id, **kwds):
             rval.append(citation.to_dict('bibtex'))
         return rval
 
+    @expose_api_anonymous_and_sessionless
+    def xrefs(self, trans, id, **kwds):
+        tool = self._get_tool(id, user=trans.user)
+        rval = []
+        for xref in tool.xrefs:
+            rval.append(xref.to_dict())
```
nsoranzo: You can use a list comprehension instead of lines 409-411:
```python
        rval = [_.to_dict() for _ in tool.xrefs]
```

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -5581,14 +5582,62 @@ A tool can have any number of EDAM operation references.
...
+      <xs:sequence>
+        <xs:element name="xref" type="xref" minOccurs="0" maxOccurs="unbounded">
+      </xs:element>
+    </xs:sequence>
+  </xs:complexType>
+  <xs:complexType name="xref">
+    <xs:annotation>
+      <xs:documentation xml:lang="en">Each xrefs element can contain one or
+more ``xref`` tag elements - each of which specifies reference information 
+according to a catalog or an other.
```
nsoranzo: I'd simplify the text to: `The ``xref`` element specifies reference information 
according to a catalog.`

lib/galaxy/tools/__init__.py
```diff
@@ -1079,6 +1080,24 @@ def _parse_citations(self, tool_source):
...
+        root = tool_source.root
+        xrefs = []
+        xrefs_elem = root.find("xrefs")
+        if xrefs_elem is None:
+            return xrefs
+
+        for xref_elem in xrefs_elem:
+            if xref_elem.tag != "xref":
+                pass
+            if hasattr(self.app, 'xrefs_manager'):
```
nsoranzo: Not sure we need a manager for xrefs, probably it's enough to parse `xref_elem` directly here.

lib/galaxy/tool_util/xsd/galaxy.xsd
ValentinMrcn: You're right  : corrected in commit 17807f4

lib/galaxy/webapps/galaxy/api/tools.py
ValentinMrcn: Thanks for the advice : corrected in commit 17807f4  

I made [shorter](https://github.com/galaxyproject/galaxy/pull/7905/commits/17807f4ddd9499b71b5be7724fa10085a466f8b8#diff-d67d082895c846a6e64465fea59344e7R409):
`return [_.to_dict() for _ in tool.xrefs]`

lib/galaxy/tool_util/xsd/galaxy.xsd
ValentinMrcn: You're right, I had copied the 'citation' documentation : corrected in commit 17807f4

ValentinMarcon:
@galaxybot test this

lib/galaxy/tools/__init__.py
ValentinMrcn: Something like that should work instead?
```
for xref_elem in xrefs_elem:
            if xref_elem.tag != "xref":
                pass
            xref = dict(
                reftype=xref_elem.text.strip(),
                content=xref_elem.attrib.get('type', None)
            )
            if xref:
                xrefs.append(xref) 
```

ValentinMarcon:
@galaxybot test this

lib/galaxy/tools/__init__.py
nsoranzo: Almost correct:
```python
        for xref_elem in xrefs_elem:
            if xref_elem.tag != "xref":
                pass
            xref = dict(
                content=xref_elem.text.strip(),
                reftype=xref_elem.attrib.get('type', None)
            )
            xrefs.append(xref)
```

ValentinMrcn:
I removed the (useless) xref manager
- lib/galaxy/config.py : changes removed
- xrefs.py : file removed
- __ init__.py : manager removed

https://github.com/galaxyproject/galaxy/pull/7905#discussion_r296825217

lib/galaxy/tools/__init__.py
ValentinMrcn: https://github.com/galaxyproject/galaxy/pull/7905#issuecomment-505405772

lib/galaxy/tools/__init__.py
```diff
@@ -1079,6 +1080,24 @@ def _parse_citations(self, tool_source):
...
+        xrefs = []
+        xrefs_elem = root.find("xrefs")
+        if xrefs_elem is None:
+            return xrefs
+        for xref_elem in xrefs_elem:
+            if xref_elem.tag != "xref":
+                pass
+            xref = dict(
+                reftype=xref_elem.text.strip(),
+                content=xref_elem.attrib.get('type', None)
```
nsoranzo: These are still inverted:
```python
                content=xref_elem.text.strip(),
                reftype=xref_elem.attrib.get('type', None)
```

ValentinMrcn: Right, I missed that point

lib/galaxy/tools/__init__.py
```diff
@@ -808,6 +808,7 @@ def parse(self, tool_source, guid=None, dynamic=False):
         self.containers = containers
 
         self.citations = self._parse_citations(tool_source)
+        self.xrefs = self._parse_xrefs(tool_source)
```
nsoranzo: Shouldn't this be:
```suggestion
        self.xrefs = tool_source.parse_xrefs()
```
? Otherwise the methods in lib/galaxy/tool_util/parser/xml.py and lib/galaxy/tool_util/parser/yaml.py are unused.

nsoranzo (commented):
It would be nice to have tests e.g. in test/unit/tools/test_parsing.py

ValentinMrcn (commented):
I guess something like that must be sufficient?

https://github.com/galaxyproject/galaxy/pull/7905/files#diff-950cd8e18178aa3cac1982751f588291

nsoranzo:
@ValentinMarcon You need to add the xrefs also to `TOOL_YAML_1` in `test/unit/tools/test_parsing.py`, see the failed job: https://circleci.com/gh/galaxyproject/galaxy/23213

lib/galaxy/tools/__init__.py
```diff
@@ -1079,6 +1080,24 @@ def _parse_citations(self, tool_source):
                     citations.append(citation)
         return citations
 
+    def _parse_xrefs(self, tool_source):
```
nsoranzo: This method is unused now, I think you need to merge its implementation into `XmlToolSource.parse_xrefs()` in `lib/galaxy/tool_util/parser/xml.py` .

ValentinMrcn (commented):
> @ValentinMarcon You need to add the xrefs also to `TOOL_YAML_1` in `test/unit/tools/test_parsing.py`, see the failed job: https://circleci.com/gh/galaxyproject/galaxy/23213

I tried that

lib/galaxy/tools/__init__.py
ValentinMrcn: I'am not sure what to do there

lib/galaxy/tool_util/parser/yaml.py
```diff
@@ -40,6 +40,13 @@ def parse_edam_operations(self):
     def parse_edam_topics(self):
         return self.root_dict.get("edam_topics", [])
 
+    def parse_xrefs(self):
+        xrefs = []
+        for i, xref_dict in enumerate(self.root_dict.get("xrefs", [])):
```
nsoranzo:
```suggestion
        for xref_dict in self.root_dict.get("xrefs", []):
```

lib/galaxy/tool_util/parser/xml.py
```diff
@@ -95,6 +95,12 @@ def parse_edam_topics(self):
             return []
         return [edam_topic.text for edam_topic in edam_topics.findall("edam_topic")]
 
+    def parse_xrefs(self):
+        xrefs = self.root.find("xrefs")
+        if xrefs is None:
+            return []
+        return [xref.text for xref in xrefs.findall("xref") if xref.get("type") in {'bio.tools'}]
```
```suggestion
        return [dict(content=xref.text.strip(), reftype=xref_elem.attrib['type']) for xref in xrefs.findall("xref") if xref.get("type")]
```

Here I have used `content` and `reftype` as keys, but maybe `type` and `value` (which you used in the yaml test) may make more sense and be more consistent. The Vue file would also need to use whatever keys you decide to use.

test/unit/tools/test_parsing.py
```diff
@@ -311,6 +317,10 @@ def test_tests(self):
         assert attributes1["compare"] == "sim_size"
         assert attributes1["lines_diff"] == 4
 
+    def test_xrefs(self):
+        xrefs = self._tool_source.parse_xrefs()
+        assert xrefs[0] == "bwa"
```
nsoranzo:
```suggestion
        assert xrefs == [{'content': 'bwa', 'reftype': 'bio.tools'}]
```

lib/galaxy/tools/__init__.py
nsoranzo: I think that with the other changes I just suggested, this method can be removed.

lib/galaxy/tool_util/parser/yaml.py
```diff
@@ -40,6 +40,13 @@ def parse_edam_operations(self):
...
+            if xref_dict["type"] == "bio.tools":
+                xrefs.append(xref_dict["value"])
+        return xrefs
```
nsoranzo: I think this method should also return a list of dicts like what I've suggested for lib/galaxy/tool_util/parser/xml.py .

lib/galaxy/tool_util/parser/xml.py
ValentinMrcn: Finally I choosed `reftype` and `value`

ValentinMrcn:
Thank you a lot @nsoranzo for your advices !

lib/galaxy/webapps/galaxy/api/tools.py
```diff
@@ -403,6 +403,11 @@ def citations(self, trans, id, **kwds):
             rval.append(citation.to_dict('bibtex'))
         return rval
 
+    @expose_api_anonymous_and_sessionless
+    def xrefs(self, trans, id, **kwds):
+        tool = self._get_tool(id, user=trans.user)
+        return [_ for _ in tool.xrefs]
```
nsoranzo:
```suggestion
        return tool.xrefs
```

lib/galaxy/tool_util/parser/yaml.py
```diff
@@ -40,6 +40,12 @@ def parse_edam_operations(self):
     def parse_edam_topics(self):
         return self.root_dict.get("edam_topics", [])
 
+    def parse_xrefs(self):
+        xrefs = self.root_dict.get("xrefs", [])
+        if xrefs is None:
+            return []
```
nsoranzo: Lines 45-46 are not needed (`xrefs` will never be `None`).

ValentinMarcon: Why?

lib/galaxy/webapps/galaxy/api/tools.py
ValentinMrcn: Saw that too late sorry

lib/galaxy/tools/__init__.py
```diff
@@ -1989,6 +1990,7 @@ def to_json(self, trans, kwd={}, job=None, workflow_building_mode=False):
             'id'            : self.id,
             'help'          : tool_help,
             'citations'     : bool(self.citations),
+            'xrefs'         : bool(self.xrefs),
```
nsoranzo: What about adding:
```python
        tool_dict["xrefs"] = self.xrefs
```
around line 1880 and remove this line?

ValentinMrcn: Ok, but I am not sure I understand the usage of the tool_model

nsoranzo:
@galaxybot test this

nsoranzo:
Looks good to me for the backend part, can someone review the UI part?

hmenager:
Thank you so much for your work @nsoranzo and @ValentinMarcon !

bgruening:
@dannon do you have a minute? Thanks!!

ValentinMrcn:
Thank you again @nsoranzo !

dannon:
@bgruening Sorry, just saw this ping in my backlog!  I do have some changes that I'd like to make to the client side of this prior to merge.  I'll work on those now.  Everyone is on board with the schema changes and internals, right?

nsoranzo:
@dannon We discussed this at the IUC meeting at GCC2019 and everyone was happy with the idea, feel free to merge if the UI part is ready!

dannon:
@nsoranzo Cool, yeah, it's good to go!

hmenager:
thank you very much @dannon and @nsoranzo for your hard work!

bgruening:
Great! Thanks all!

PR #7911, opened 2019-05-08
Remove ``file`` attribute of bibtex citation

nsoranzo:
Introduced in commit cb9fc095fc435250e09b21ac7dafd2cb2f85447f but never documented, tested or used in any known tool.

### DIFF
--- a/lib/galaxy/managers/citations.py
+++ b/lib/galaxy/managers/citations.py
@@ -1,6 +1,5 @@
 import functools
 import logging
-import os
 
 import requests
 from beaker.cache import CacheManager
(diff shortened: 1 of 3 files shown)


jmchilton:
I was thinking back to that hackathon and I think I added this for khmer and those tools do not use it so I think this is especially fine. https://github.com/galaxyproject/tools-iuc/blob/master/tools/khmer/macros.xml. +1 from me.

jmchilton (approved):

nsoranzo:
@jmchilton Thanks for the additional feedback!

Out of WIP now.

PR #7913, opened 2019-05-08
Datatype Validation and Tool-based QC 

jmchilton:
These concepts are completely separable so if either is problematic - I'll split them out.

## Validation

Add HDA/LDDA validation state and validation message columns, add a framework for validating whole files, API endpoints for running validation, and a PJA for validating outputs during workflow runs. I'd say it is different than sniffing since I don't really think it is important sniffers don't have false positives or false negatives - it is more that the whole validation system works together correctly. Whereas validation should be much more precise and should check the whole file. One of the biggest problems @jennaj says Galaxy users have is truncated files and sniffers pretty explicitly shouldn't be catching those.

There was an attempt to do this over a decade ago that didn't take off:
- xref 925aec1
- xref f4f9e1d

## QC

New stdio type for capturing QC output messages from jobs. 

## Context

I'm planning some work on automated reporting for workflow invocations - my thinking was one of the sections (or maybe tabs if iReport is leaned on) would be dataset validation and QC. I could scrape the workflow and capture this data. QC might also be actual file/dataset outputs - so I would combine this job QC with maybe checking for a "qc" tag on outputs to summarize information across both potential sources.

## Future Directions

(If I were reviewing this, I might insist on some of these before merging 😄.)

- [ ] Include validation information in current dataset details.
- [ ] Include QC information in current dataset details.
- [ ] Workflow editor UI for validation PJA.
- [ ] Check validation information in dataset error reports and add links to generate validation information if not present (follow up on #7755).
- [ ] Consider an option for not executing tools with inputs that are marked as "invalid". This seems like it has a real opportunity to piss off users if things get marked as invalid incorrectly - so I'm not sure it is a wise idea in general - but maybe as part of specific workflows or tools?
- [ ] Workflow invocation reports.

### DIFF
--- a/lib/galaxy/datatypes/binary.py
+++ b/lib/galaxy/datatypes/binary.py
@@ -25,6 +25,7 @@ from galaxy.datatypes import metadata
 from galaxy.datatypes.data import (
     DatatypeValidation,
     get_file_peek,
+)
 from galaxy.datatypes.metadata import DictParameter, ListParameter, MetadataElement, MetadataParameter
 from galaxy.util import nice_size, sqlite
 from galaxy.util.checkers import is_bz2, is_gzip
--- a/lib/galaxy/datatypes/binary.py
+++ b/lib/galaxy/datatypes/binary.py
@@ -22,7 +22,9 @@ from bx.seq.twobit import TWOBIT_MAGIC_NUMBER, TWOBIT_MAGIC_NUMBER_SWAP, TWOBIT_
 
 from galaxy import util
 from galaxy.datatypes import metadata
-from galaxy.datatypes.data import DatatypeValidation, get_file_peek
+from galaxy.datatypes.data import (
+    DatatypeValidation,
+    get_file_peek,
 from galaxy.datatypes.metadata import DictParameter, ListParameter, MetadataElement, MetadataParameter
 from galaxy.util import nice_size, sqlite
 from galaxy.util.checkers import is_bz2, is_gzip
--- a/lib/galaxy/datatypes/binary.py
+++ b/lib/galaxy/datatypes/binary.py
@@ -22,7 +22,7 @@ from bx.seq.twobit import TWOBIT_MAGIC_NUMBER, TWOBIT_MAGIC_NUMBER_SWAP, TWOBIT_
 
 from galaxy import util
 from galaxy.datatypes import metadata
-from galaxy.datatypes.data import get_file_peek
+from galaxy.datatypes.data import DatatypeValidation, get_file_peek
 from galaxy.datatypes.metadata import DictParameter, ListParameter, MetadataElement, MetadataParameter
 from galaxy.util import nice_size, sqlite
 from galaxy.util.checkers import is_bz2, is_gzip
@@ -284,6 +284,10 @@ class BamNative(CompressedArchive):
         Binary.init_meta(self, dataset, copy_from=copy_from)
 
     def sniff(self, filename):
+        return BamNative.is_bam(filename)
+
+    @classmethod
+    def is_bam(cls, filename):
         # BAM is compressed in the BGZF format, and must not be uncompressed in Galaxy.
         # The first 4 bytes of any bam file is 'BAM\1', and the file is binary.
         try:
@@ -422,6 +426,13 @@ class BamNative(CompressedArchive):
                                        column_number=column_number,
                                        column_names=column_names,
                                        column_types=column_types)
+
+    def validate(self, dataset, **kwd):
+        if not BamNative.is_bam(dataset.file_name):
+            return DatatypeValidation.invalid("This dataset does not appear to a BAM file.")
+        elif self.dataset_content_needs_grooming(dataset.file_name):
+            return DatatypeValidation.invalid("This BAM file does not appear to have the correct sorting for declared datatype.")
+        return DatatypeValidation.validated()
 
 
 @dataproviders.decorators.has_dataproviders
--- a/lib/galaxy/datatypes/data.py
+++ b/lib/galaxy/datatypes/data.py
@@ -48,6 +48,36 @@ DOWNLOAD_FILENAME_PATTERN_DATASET = "Galaxy${hid}-[${name}].${ext}"
 DOWNLOAD_FILENAME_PATTERN_COLLECTION_ELEMENT = "Galaxy${hdca_hid}-[${hdca_name}__${element_identifier}].${ext}"
 
 
+class DatatypeValidation(object):
+
+    def __init__(self, state, message):
+        self.state = state
+        self.message = message
+
+    @staticmethod
+    def validated():
+        return DatatypeValidation("ok", "Dataset validated by datatype validator.")
+
+    @staticmethod
+    def invalid(message):
+        return DatatypeValidation("invalid", message)
+
+    @staticmethod
+    def unvalidated():
+        return DatatypeValidation("unknown", "Dataset validation unimplemented for this datatype.")
+
+    def __repr__(self):
+        return "DatatypeValidation[state=%s,message=%s]" % (self.state, self.message)
+
+
+def validate(dataset_instance):
+    try:
+        datatype_validation = dataset_instance.datatype.validate(dataset_instance)
+    except Exception as e:
+        datatype_validation = DatatypeValidation.invalid("Problem running datatype validation method [%s]" % str(e))
+    return datatype_validation
+
+
 class DataMeta(abc.ABCMeta):
     """
     Metaclass for Data class.  Sets up metadata spec.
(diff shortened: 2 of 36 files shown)


jennaj:
Example Galaxy Help Q&A from today that shows how data formatting problems trip up users (just for reference, I realize we all already agree integrating format QC will help ;). Check out how many tools/steps they need to use to find out what is going wrong. All to hunt for what is probably at least one blank line somewhere in the data... .. and I didn't even include the full how-to. Will do that as a follow up if they need it/can't figure out how to use the Select tool.
https://help.galaxyproject.org/t/usegalaxy-chipseq-trim-galore-error/1253

cc @ic4f (is working on other fastq data checks, eg: this is another problematic format use-case/potential test dataset to create)

bgruening:
ping @wm75, this seems to be related to our general `report` work/plans.

lib/galaxy/jobs/actions/post.py
```diff
@@ -65,6 +65,23 @@ def get_short_str(cls, pja):
             return "Email the current user when this job is complete."
 
 
+class ValidateOutputsAction(DefaultJobAction):
+    """
+    This action sends an email to the galaxy user responsible for a job.
```
mvdbeek:
```suggestion
    This action validates the produced outputs against the expected datatype.
```

lib/galaxy/datatypes/tabular.py
```diff
@@ -729,6 +730,19 @@ def merge(split_files, output_file):
         if exit_code != 0:
             raise Exception("Error merging VCF files: %s" % stderr)
 
+    def validate(self, dataset, **kwd):
+        # with tempfile.NamedTemporaryFile() as t:
```
mvdbeek: we can drop this, right ?

lib/galaxy/datatypes/data.py
```diff
@@ -743,6 +769,9 @@ def dataprovider(self, dataset, data_format, **settings):
             return self.dataproviders[data_format](self, dataset, **settings)
         raise dataproviders.exceptions.NoProviderAvailable(self, data_format)
 
+    def validate(self, dataset, **kwd):
+        return DatatypeValidation.unvalidated()
```
mvdbeek, line 773: Would it make sense to perhaps do a minimum check and make sure that that datatypes' sniffer doesn't return False / raise an exception ?

jmchilton: I've got problems with this... although I know we do this in some ways already. I wish the sniffer had a way of saying - I don't know. Because the way it is now I think it is totally fine for a sniffer to report false positives or false negatives - it is an estimate. The validation wouldn't be perfect of course, but I'd like it to be considered a bug if the validation reports false positives or negatives but not the sniffer. I want to start with the assumption that validation is a stronger claim, and I'd like to work toward the sniffer being a weaker claim.

@nsoranzo recommended the same thing though... maybe I'm wrong and we should just make that change and then go through and do the work of tweaking it toward my goals from that basis.

jennaj: I vote for full checks with the validator. Otherwise, won't help as much. 

No one runs _production quality_ line-command bioinfo analysis without fully checking the inputs first -- is the top way to prevent bad data from presenting as "tool errors" when is really an input content problem. It should be possible to do that in Galaxy, too.

Super annoying and wastes resources to discover that after investing a bunch of time/effort/compute to complete upstream work, that sends the result to a downstream tool that is more "picky" about same inputs (reused) or entire new inputs (ref annotation, etc), and fails out later. The solution is to fix the problematic input and rerun the entire thing again. IF the user is even able to figure out that is the root problem (many don't -- and the fixes are usually pretty simple).

SO, investing in complete, early detection of input problems is worth it imo, and the primary goal that at least I had in mind when discussing why to add this in to begin with.  

But maybe I am missing something? Am looking at this from a "how to make analysis more likely to be successfull for the end user" perspective.

lib/galaxy/datatypes/binary.py
```diff
@@ -22,7 +22,7 @@
 
 from galaxy import util
 from galaxy.datatypes import metadata
-from galaxy.datatypes.data import get_file_peek
+from galaxy.datatypes.data import DatatypeValidation, get_file_peek
```
```suggestion
from galaxy.datatypes.data import (
    DatatypeValidation,
    get_file_peek,
```

lib/galaxy/datatypes/genetics.py
```diff
@@ -20,7 +20,7 @@
 from six.moves.urllib.parse import quote_plus
 
 from galaxy.datatypes import metadata
-from galaxy.datatypes.data import Text
+from galaxy.datatypes.data import DatatypeValidation, Text
```
mvdbeek:
```suggestion
from galaxy.datatypes.data import (
    DatatypeValidation,
    Text,
)
```

lib/galaxy/datatypes/interval.py
```diff
@@ -286,9 +286,9 @@ def validate(self, dataset):
                 try:
                     next(reader)
                 except ParseError as e:
-                    errors.append(e)
+                    return DatatypeValidation.invalid(str(e))
```
mvdbeek:
```suggestion
                    return DatatypeValidation.invalid(unicodify(e))
```

jmchilton:
I think this is ready to go again. Thanks for the review!

lib/galaxy/datatypes/sequence.py
```diff
@@ -795,16 +796,36 @@ def quality_check(lines):
...
             if match:
                 start, end = match.span()
-                if (end - start) == len(headers[1][0]):
+                if (end - start) == len(block[1][0]):
                     return True
         return False
 
+    def validate(self, dataset, **kwd):
+        headers = iter_headers(dataset.file_name, sep='\n', count=-1)
+        # check to see if the base qualities match
```
bgruening, line 815: Can we add here a TODO ... this is currently not implemented, isn't it?
We will work on this in the next release cycle.

lib/galaxy/webapps/galaxy/api/history_contents.py
```diff
@@ -673,6 +673,29 @@ def update(self, trans, history_id, id, payload, **kwd):
...
+        * PUT /api/histories/{history_id}/contents/{id}/validate
+            updates the values for the history content item with the given ``id``
+
+        :type   history_id: str
+        :param  history_id: encoded id string of the items's History
+        :type   id:         str
+        :param  id:         the encoded id of the history item to validate
+
+        :rtype:     dict
+        :returns:   TODO
```
bgruening, line 689: Should that be fixed? Should this return an error object?

lib/galaxy/webapps/galaxy/buildapp.py
```diff
@@ -236,6 +236,11 @@ def populate_api_routes(webapp, app):
                           controller="history_contents",
                           action="update_permissions",
                           conditions=dict(method=["PUT"]))
+    webapp.mapper.connect("history_contents_validate",
```
bgruening, line 239: Do we need a library_contents_validate as well?

bgruening (approved):
Thanks, John only a few minor, non-blocking things inline.

lib/galaxy/webapps/galaxy/api/history_contents.py
jmchilton: It is asynchronous - because it launches an external set metadata job effectively - so we won't have that yet.

jmchilton: ~I guess ``PUT`` s should just return empty bodies and use status codes to indicate what happened.~

Update: Even if some RESTful guides recommend this, we don't do this currently anywhere.

lib/galaxy/datatypes/sequence.py
jmchilton: This was a copy and pasted comment, but isn't this what quality_check does in the nextline?

bgruening: The function is defined as ...
```
def quality_check(lines):
        return True
```

maybe I miss something.

lib/galaxy/webapps/galaxy/api/history_contents.py
bgruening: Ok, maybe I misread the update_batch() function.

lib/galaxy/datatypes/sequence.py
jmchilton: The subclasses implement it I thought - at least FastqSanger does:

```python
    @staticmethod
    def quality_check(lines):
        """Presuming lines are lines from a fastq file, return True if the qualities are compatible with sanger encoding"""
        for line in islice(lines, 3, None, 4):
            if not all(_ >= '!' and _ <= 'S' for _ in line[0]):
                return False
        return True
```

Looks like we could definitely do more validation for sure though.

lib/galaxy/webapps/galaxy/buildapp.py
jmchilton: At some point - yes. I don't know how to do that yet though.

PR #7919, opened 2019-05-08
change keyboard input for floats and switch to es6 features

WilliamHolden:


martenson:
@WilliamHolden please resolve the conflict so we can run tests

PR #7921, opened 2019-05-08
WIP: Allow building client in GALAXY_STATIC_ROOT/static

mvdbeek:


mvdbeek:
not sure what the point was, but since we got the installable client now we probably don't need this anymore

PR #7924, opened 2019-05-08
More structured Pulsar path handling.

jmchilton:
- Track metadata files and transfer them properly (fixing  https://github.com/galaxyproject/pulsar/issues/169 along the way hopefully).
- For metadata inputs and extra files inputs, only transfer them if they are referenced in the tool command block. (Track actual references the way we do for "unstructured" - i.e. loc file - paths.)

In addition to more optimized, correct, and structured file staging this also enables loading files directly from object stores from Pulsar. The newer file actions on the Pulsar to target these kinds of files and to utilize the object store can be found here at least https://github.com/galaxyproject/pulsar/commit/d130b11b095f8354e9b0e81d488e63355d58d57a.


jmchilton:
Merged as part of #8625

PR #7925, opened 2019-05-09
[19.05] Fix default image

hexylena:


dannon:
@erasche Thanks, this fetches the right image for me.

PR #7927, opened 2019-05-09
change keyboard input for floats and switch to es6 features

WilliamHolden:
reopened this with new branch since I was having issues with the TravisCI build test on https://github.com/galaxyproject/galaxy/pull/7919.


dannon:
@WilliamHolden Code is fetching to CircleCI fine now, it's just an eslint failure: https://circleci.com/gh/galaxyproject/galaxy/14944

client/galaxy/scripts/mvc/ui/ui-slider.js
```diff
@@ -26,52 +24,25 @@ var View = Backbone.View.extend({
...
+            .on("change", e => {
+                this.value(e.currentTarget.value);
             })
-            .on("keyup", e => {
-                pressed[e.which] = false;
-            })
-            .on("keydown", function(e) {
-                var v = e.which;
-                pressed[v] = true;
-                if (self.model.get("is_workflow") && pressed[16] && v == 52) {
```
dannon, line 40: I think what this logic was trying to do was to allow `${workflow_parameter}` entries when this widget is used in a workflow context, and it looks like the input replacement below might restrict this.

WilliamHolden: I'll do some more testing to make sure this works in a workflow context. I'm not super familiar with how the workflow works but my though process was that this if statement would cover the use case you described
```
if (this._isParameter(input.value)) {
            return;
}
```

dannon: @WilliamHolden Yep, that handles it, just tested it.  Thanks for checking, and simplifying the logic here!

martenson:
This one could really use a test.

dannon:
@martenson Yep, agreed.  All of our tool form input widgets should have input validation tests; my hope/thinking right now is that we can write those using vue-test-utils based testing as we rewrite them as a set of vue components instead of trying to fight with this old backbone-based cruft more than we have to.

martenson:
I didn't like the infinity-if but this old cruft was time-tested. I fear we will see more of https://github.com/galaxyproject/galaxy/issues/7836 because of this switch.

PR #7929, opened 2019-05-09
Use transparent background in default welcome page

guerler:
Improves contrast and matches appearance to main.


guerler:
@galaxybot test this

Issue #7879, opened 2019-05-02
data dialog improvements & feedback


martenson:
Overall design

* It is unclear to me what am I able to select. Can I select whole folder? Whole collection? Whole library? Three datasets from the same collection? A data library dataset and a history dataset?
* I am unable to navigate 'folder' structure by clicking on folder names, which I think is the standard for browsing such systems. Instead I have to find the 'view' link in the right. Maybe swapping the 'view' for 'select' and allowing navigation by clicking on the name would be better and could help a bit with the point above.
- [x] Related: I am not sure that the 'OK' button has any function, since the interface seems to be exclusively 'click-to-select'.
* Similarly the interface could use the "folder" structure navigation similar to how data libraries do it (i.e. providing an element at the top of the folder to go to the parent, instead of an extra "back" button in the very bottom left)
* The UI does not indicate clearly what are the types of the rows that I am seeing. Data libraries row, folder row and collection row all look exactly the same.
* Given the structure complexity collection/libraries can have it would be nice to have the ability to resize the modal and render more items. 
* Similarly showing location breadcrumbs when digging in folders would improve the experience.

Architecture
* The library browsing should go through the one-level-at-time `folders/{id}/contents` endpoint instead of the old library endpoint that loads all library levels at once.

Extra or tiny things

* Time stamp could be made user-readable (4 days ago)
* Search could include data libraries
- [x] Instead of search "clear" button we use `X` in other interfaces.

Big Picture

Now we have **five** ways to select datasets for input: select2 input element/single dataset/multiple datasets/one collection/data dialog(single dataset). I was hoping the data dialog would lead to unification of the ways we select dataset inputs and would deprecate at least the three previous ways. Is that possible? Is that wanted/needed?

xref https://github.com/galaxyproject/galaxy/pull/7553 https://github.com/galaxyproject/galaxy/pull/7746

guerler:
Thanks for starting this card to keep track of this. Here are some additional items:
- [ ] Time stamp could be made user-readable (4 days ago) (from list above)
- [ ] Pagination should always be visible and not scroll out
- [ ] Table header should always be visible and not scroll out
- [ ] Show bread crumbs at the top for navigation, remove “back” button (from list above)

PR #7935, opened 2019-05-11
Remove view button, fix library flag in data dialog

guerler:
This PR removes the current "view" button from the data dialog row and allows users to click on the entire row instead. Additionally, it fixes a bug regarding the library flag. The seach clearing label is replaced by an x-icon. xref #7879.


martenson (approved):

martenson:
Thanks @guerler!

PR #7936, opened 2019-05-11
CLI tools for installing/testing Docker and Singularity containers

simonbray:
Previously in a PR to the galaxy-lib repo: https://github.com/galaxyproject/galaxy-lib/pull/87


.travis.yml
```diff
@@ -1,11 +1,30 @@
...
 language: generic
+sudo: required
 
 env:
   - TOX_ENV=py27-first_startup
   - TOX_ENV=py35-first_startup
 
+before_install:
+  # Install Singularity
+  - sudo apt install -y build-essential libssl-dev uuid-dev libgpgme11-dev squashfs-tools wget git
```
mvdbeek: We are not using travis for these tests anymore. The docker testing image comes pre-installed with singularity. For a simple test I'd suggest you create something unit-test like in `test/integration`, which will be picked up automatically.

lib/galaxy/tools/deps/mulled/get_tests.py
```diff
@@ -0,0 +1,304 @@
...
+    for container in containers:
+        tests = test_search(container, recipes_path, deep, anaconda_channel, github_repo)
+        package_tests['commands'] += tests.get('commands', [])  # not a very nice solution but probably the simplest
+        for imp in tests.get('imports', []):
+            package_tests['imports'].append("%s 'import %s'" % (tests['import_lang'], imp))
+
+    return package_tests
+
+
+doctest.testmod()
```
mvdbeek: I'd suggest avoiding doctests. We're currently using pytest, but you can also use unittest tests if you like.

lib/galaxy/tools/deps/mulled/mulled_list.py
```diff
@@ -0,0 +1,168 @@
...
+                f.write('%s\n' % container)
+    else:
+        print(containers)
+
+
+if __name__ == '__main__':
+    # main()
+
+    import doctest
+    doctest.testmod()
```
mvdbeek: Again, those doctests should probably live as unit tests in test/integration, or if you want, test/mulled

mvdbeek:
Let me know if you'd like me to re-create some of the doctests as a unit test

simonbray:
Thanks for the offer; I'll let you know if I need a hand.

simonbray:
@mvdbeek would you be able to convert the doctests? Or perhaps at least a few examples so I have an idea of how to do the others.

mvdbeek:
I'll try to get something started this week

simonbray:
Thanks! Much appreciated.

mvdbeek:
@simonbray I converted 2 tests to unittest style tests here: https://github.com/simonbray/galaxy/pull/1

simonbray:
Hi @mvdbeek, I've got most of the tests passing now, with one exception. This PR needs the jinja2 library and therefore I added it to `pinned-requirements.txt` in the most recent commit. However, the test_galaxy_packages test still fails with `ImportError: No module named 'jinja2'`. Do you have an idea why?

mvdbeek:
You'd have to add jinja2 to `packages/tool_util/requirements.txt`

simonbray:
@jmchilton, I'm happy to finish this, but would really appreciate some guidance about how much of this PR is useful or in scope. For example, we already used some of the code here: https://github.com/BioContainers/singularity-build-bot; I don't know if you think it's worth duplicating it in the Galaxy codebase?

Thanks!

jmchilton (approved):

mvdbeek:
@galaxybot test this

mvdbeek:
Awesome, thanks a lot @simonbray!

simonbray:
Oh, I missed this. Thanks for the merge @mvdbeek!

Should I now make a PR to the galaxy-lib repo (which was the original goal)? Or will this be done automatically?

mvdbeek:
galaxy-lib will be phased out, the mulled utils live in galaxy-tool-utils. You should be able to install galaxy-tool-util (`pip install galaxy-tool-util`) and that should place the mulled scripts on your PATH.

PR #7944, opened 2019-05-13
[19.05] Allow mapping over of expression tool params

mvdbeek:
This allows feeding a collection of parameters values to a parameter
value. This enables [galaxyproject/tools-iuc#2404](https://github.com/galaxyproject/tools-iuc/pull/2404#issuecomment-491740393)
for instance, where we normalize a collection based on a computed
parameter.

### DIFF
--- a/lib/galaxy/workflow/run.py
+++ b/lib/galaxy/workflow/run.py
@@ -376,6 +376,8 @@ class WorkflowProgress(object):
                 if not replacement.is_ok:
                     raise modules.CancelWorkflowEvaluation()
             else:
+                if not replacement.collection.populated:
+                    raise modules.DelayedWorkflowEvaluation()
                 pending = False
                 for dataset_instance in replacement.dataset_instances:
                     if dataset_instance.is_pending:
--- a/lib/galaxy/workflow/run.py
+++ b/lib/galaxy/workflow/run.py
@@ -1,4 +1,3 @@
-import json
 import logging
 import uuid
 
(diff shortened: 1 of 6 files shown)


lib/galaxy/workflow/run.py
```diff
@@ -369,15 +368,22 @@ def replacement_for_connection(self, connection, is_data=True):
...
+        data_inputs = (model.HistoryDatasetAssociation, model.HistoryDatasetCollectionAssociation, model.DatasetCollection)
+        if not is_data and isinstance(replacement, data_inputs):
+            if isinstance(replacement, model.HistoryDatasetAssociation):
+                if replacement.is_pending:
+                    raise modules.DelayedWorkflowEvaluation()
+                if not replacement.is_ok:
+                    raise modules.CancelWorkflowEvaluation()
             else:
-                raise modules.CancelWorkflowEvaluation()
+                pending = False
```
jmchilton, line 381: should this check if ``replacement.collection.populated`` and delay if not or is that happening somewhere else?

mvdbeek: good point!

jmchilton (approved):
I'm a bit scared this doesn't match semantics I've established in the CWL branch... but it seems to make sense in this context. If things are incompatible I'll just deal with that downstream with different mapping types or something. 👍 and thanks @mvdbeek.

mvdbeek:
Do you have some commits I can look at, maybe there's another way to do this ?

jmchilton:
I wouldn't worry about it for now, thanks though!

PR #7945, opened 2019-05-13
cellxgene GIE: use API to get datasets.

jxtx:
No longer requires volume mounts, so useable on main. Also updates to
latest version of cellxgene in container.


bgruening (approved):
Looks good!

jxtx:
Thanks!

Issue #5785, opened 2018-03-26
history peeks for compressed datatypes are unreliable


martenson:
We should fix them or remove them since they are confusing to users.

* number of lines in a sequence is wrong most/all of the time
* the peek has extra whitespace that is not present when uncompressed

martenson:
examples: 

same fastqsanger file with gz results in different peek than uncompressed
![screenshot 2018-03-26 14 48 10](https://user-images.githubusercontent.com/1814954/37926227-c4d179d0-3104-11e8-90e0-425ac9af648a.png)

genbank seems different, but this looks just like a formatting diff, unlike the fastqsanger that looks like different data
![screenshot 2018-03-27 16 01 05](https://user-images.githubusercontent.com/1814954/37991814-2bdbf898-31d8-11e8-81b9-8c2a830bc261.png)

martenson:
xref https://github.com/galaxyproject/galaxy/issues/2694

hexylena:
Looks good now :)

PR #7952, opened 2019-05-13
Fix extra newlines in peek code.

jmchilton:
Probably related to https://github.com/galaxyproject/galaxy/issues/5785


lib/galaxy/datatypes/data.py
```diff
@@ -1062,6 +1064,8 @@ def get_file_peek(file_name, is_multi_byte=False, WIDTH=256, LINE_COUNT=5, skipc
                     skip_line = True
                     break
             if not skip_line:
+                if line.endswith('\n'):
+                    line = line[:-1]
```
nsoranzo: You can remove it from here and rework lines 1053-1060 instead, as follows:
```python
            if line.endswith('\n'):
                line = line[:-1]
            elif not line_wrap:
                while True:
                    i = temp.read(1)
                    if not i or i == '\n':
                        break
```

lib/galaxy/datatypes/data.py
```diff
@@ -1062,6 +1064,8 @@ def get_file_peek(file_name, is_multi_byte=False, WIDTH=256, LINE_COUNT=5, skipc
...
                 lines.append(line)
                 count += 1
-    return '\n'.join(lines)
+    return '\n'.join(lines) + '\n'
```
nsoranzo: This change makes the `test_get_file_peek` unit test fail. Reverting it may break some other tests, but I'd prefer not to add newlines that may not be at the end of the original file.

jmchilton: I was thinking since they are just used in HTML display it wasn't a big deal, but you're right I guess - we shouldn't be adding extra newlines. 🤔

jmchilton:
Another shot at cleaning up get_file_peek.

Behavior changes:
 - Integrated @nsoranzo suggestions.
 - Now includes a trailing new line if the input contained a trailing newline.
 - Now returns only as many lines as are present if the file contains fewer lines than LINE_COUNT.

@nsoranzo still hate it? Maybe I should have started with a failing test case to demonstrate the poor past behavior I didn't want to retain and ask you advice on fixing it.

PR #7953, opened 2019-05-13
Fix populators.py workflow testing for non-text inputs.

jmchilton:


mvdbeek:
@galaxybot test py3

mvdbeek:
Thanks @jmchilton !

PR #7960, opened 2019-05-14
Adding support for yaml files

astrovsky01:


lib/galaxy/datatypes/text.py
```diff
@@ -726,3 +727,38 @@ def sniff_prefix(self, file_prefix):
...
+            try:
+                item = yaml.load(file_prefix.contents_header)
+                # exclude simple types, must set format in these cases
+                assert isinstance(item, (list, dict))
+                return True
+            except Exception:
+                return False
+        else:
+            start = file_prefix.string_io().read(100).strip()
+            if start:
```
jmchilton: I feel like this will include too many non-YAML files. I'd rather you just return False but I understand that has downsides.

astrovsky01: I had thought of that, and my rationale was that it had to be labeled as a yaml in the filename first to trigger the sniffer. If it is bad after that, it will break and throw an error, but yaml are too open in their structure to do much else. I can't get to the computer that has this on it right now, but if you would prefer, I'll remake the pr tomorrow morning and return false if the file is too big.

lib/galaxy/datatypes/text.py
```diff
@@ -726,3 +727,38 @@ def sniff_prefix(self, file_prefix):
...
+@build_sniff_from_prefix
+class Yaml(Text):
+    """Yaml files"""
+    file_ext = "yaml"
+
+    def sniff_prefix(self, file_prefix):
+        """
+            Try to load the string with the yaml module. If successful it's a yaml file.
+        """
+        return self._looks_like_yaml(file_prefix)
```
bernt-matthias, line 872: Why what is the reason for the extra function call?

lib/galaxy/datatypes/text.py
bernt-matthias: Some random remarks from the top of my head:

- I guess there are loads of test files in the Galaxy sources that we could try. 
- Wouldn't it be sufficient to check for "^---" is the first non comment line?

lib/galaxy/datatypes/text.py
```diff
@@ -726,3 +727,38 @@ def sniff_prefix(self, file_prefix):
...
+        return 'application/yaml'
+
+    def _looks_like_yaml(self, file_prefix):
+        # Pattern used by SequenceSplitLocations
+        if file_prefix.file_size < 50000 and not file_prefix.truncated:
+            # If the file is small enough - don't guess just check.
+            try:
+                item = yaml.load(file_prefix.contents_header)
+                # exclude simple types, must set format in these cases
+                assert isinstance(item, (list, dict))
```
bernt-matthias, line 884: why exclude them?

lib/galaxy/datatypes/text.py
bernt-matthias: - Question is also how fast the yaml reader will find out that it is not a yaml file. My guess is that there will be only a few cases where the whole file is read. Maybe we can give the while prefix/file to the loader and it will fail fast in almost all cases. 

Btw. is the file/prefix that is given to the parser complete lines or can there be truncated lines?

lib/galaxy/datatypes/text.py
astrovsky01: I had been working with the json sniffer as a template. I had the same thought, but my lack of experience sold me it might be there for a good reason that I hadn't understood.

lib/galaxy/datatypes/text.py
astrovsky01: The prefix is based on Galaxy's build from prefix function called at the top, and I had understood it to be full lines. I don't think, especially with the shape of yaml files, that full lines are a problem here, and in fact truncating them could cause big problems.
In regards to the other comment, looking at some files online, I had a bit of a misunderstanding as the the purpose of the "---". Since it doesn't always start the document, just separating directives, I didn't think to use it, but the case where it won't happen in the first couple hundred lines would be rare, and it would be valid as a sniff. I'll switch it out.

lib/galaxy/datatypes/text.py
astrovsky01: Remnant of Json sniffer that I failed to take out

lib/galaxy/datatypes/text.py
```diff
@@ -726,3 +727,37 @@ def sniff_prefix(self, file_prefix):
...
+            try:
+                item = yaml.load(file_prefix.contents_header)
+                assert isinstance(item, (list, dict))
+                return True
+            except Exception:
+                return False
+        else:
+            start = file_prefix.string_io().read(100).strip()
+            if start:
+                # If file is too big,
```
bgruening: We could be a little bit more explicit here: 
Read line by line and split after 10.000 lines or so on a line that doesn't start with space. You should be able to load the first chunk into the yaml parser, imho. There are some cases, like multi-line lists and so, but they are rare imho.

bgruening (commented):
It would be great to add doc-tests as well for yaml, json and some random txt files.

mvdbeek:
@galaxybot test this

mvdbeek:
That's not really necessary IMO, the upload integration test is a much better true/false positive test.

mvdbeek:
But we should have a yaml file in `lib/galaxy/datatypes/test`.

mvdbeek:
@astrovsky01 can you resolve the conflicts and address outstanding comments?

mvdbeek:
Pushed it to 21.05, but if you want to get this in ping me.

lib/galaxy/datatypes/text.py
```diff
@@ -856,3 +858,40 @@ def sniff_prefix(self, file_prefix):
...
+
+    def _looks_like_yaml(self, file_prefix):
+        # Pattern used by SequenceSplitLocations
+        if file_prefix.file_size < 5000 and not file_prefix.truncated:
+            # If the file is small enough - don't guess just check.
+            try:
+                item = yaml.load(file_prefix.contents_header)
+                assert isinstance(item, (list, dict))
+                return True
+            except Exception:
```
bernt-matthias: Maybe we should catch `YAMLError`

lib/galaxy/datatypes/text.py
```diff
@@ -856,3 +858,40 @@ def sniff_prefix(self, file_prefix):
...
+    def get_mime(self):
+        """Returns the mime type of the datatype"""
+        return 'application/yaml'
+
+    def _looks_like_yaml(self, file_prefix):
+        # Pattern used by SequenceSplitLocations
+        if file_prefix.file_size < 5000 and not file_prefix.truncated:
+            # If the file is small enough - don't guess just check.
+            try:
+                item = yaml.load(file_prefix.contents_header)
```
bernt-matthias: Maybe we should use `safe_load`?

lib/galaxy/datatypes/text.py
```diff
@@ -856,3 +858,40 @@ def sniff_prefix(self, file_prefix):
...
+        """Returns the mime type of the datatype"""
+        return 'application/yaml'
+
+    def _looks_like_yaml(self, file_prefix):
+        # Pattern used by SequenceSplitLocations
+        if file_prefix.file_size < 50000 and not file_prefix.truncated:
+            # If the file is small enough - don't guess just check.
+            try:
+                item = yaml.safe_load(file_prefix.contents_header)
+                assert isinstance(item, (list, dict))
```
bernt-matthias, line 884: Is the assertion necessary, or just  return false

astrovsky01: Good idea

astrovsky01: I guess without that, it passes safe_load as a string. It seems to have been necessary in the JSON sniffer as well.

lib/galaxy/config/sample/datatypes_conf.xml.sample
```diff
@@ -247,6 +247,7 @@
     </datatype>
     <datatype extension="directory" type="galaxy.datatypes.data:Directory">
     </datatype>
+    <datatype extension="yaml" type="galaxy.datatypes.text:Yaml" display_in_upload="true" />
```
elichad, line 251: If this list is what users search through to assign datatypes in the UI, we should also include the `yml` extension here.

astrovsky01: Good point. I'll add it in the last commit. Thanks!

nsoranzo: I don't think that's desirable, this will require a datatype subclass and will unnecessarily inflate the (already long) datatype list.

bernt-matthias: I agree with @nsoranzo . But a short description would be nice.

bgruening: xref: https://yaml.org/faq.html

elichad: Is there not a way to have 2 extensions link to a single datatype? I'm not familiar with the intricacies of datatypes here, and having had a brief look, I don't understand how the subclasses work... but in this case it probably wouldn't matter if a `.yml` file was converted to `Yaml` for all internal/tool uses.

`.yml` may not be officially encouraged by the YAML developers, but it is still widely used – and I would guess that most users of YAML files don't know which is the preferred extension. I've been using YAML for a couple of years and I only discovered `.yaml` is preferred when I looked it up before commmenting this yesterday.

I guess whether this is appropriate depends on whether users (believe they) search datatypes by file _extension_ or file _type_. So far I've only searched by extension, so I would naturally search for `yml` when uploading a `.yml` file, and I might assume YAML wasn't supported if I didn't find any result for that.

bernt-matthias: With the availability of a sniffer I would hope that users would only need to choose the datatype manually in corner cases.

elichad: Oh, that's true, I forgot about that. I guess this is okay in that case

bgruening (approved):

PR #7962, opened 2019-05-14
tensorflow checkpoint file support

astrovsky01:
Adds support for binary checkpoint files used in tensorflow

### DIFF
--- a/config/datatypes_conf.xml.sample
+++ b/config/datatypes_conf.xml.sample
@@ -736,6 +736,7 @@
     <datatype extension="dada2_mergepairs" type="galaxy.datatypes.binary:RData" subclass="true" display_in_upload="true" />
     <datatype extension="dada2_sequencetable" type="galaxy.datatypes.tabular:Tabular" mimetype="application/text" subclass="true" display_in_upload="true" />
     <datatype extension="dada2_uniques" type="galaxy.datatypes.tabular:Tabular" mimetype="application/text" subclass="true" display_in_upload="true" />
+    <datatype extension="ckpt" type="galaxy.datatypes.binary:Tensorflow" display_in_upload="true" />
   </registration>
   <sniffers>
     <!--
--- a/config/datatypes_conf.xml.sample
+++ b/config/datatypes_conf.xml.sample
@@ -736,7 +736,7 @@
     <datatype extension="dada2_mergepairs" type="galaxy.datatypes.binary:RData" subclass="true" display_in_upload="true" />
     <datatype extension="dada2_sequencetable" type="galaxy.datatypes.tabular:Tabular" mimetype="application/text" subclass="true" display_in_upload="true" />
     <datatype extension="dada2_uniques" type="galaxy.datatypes.tabular:Tabular" mimetype="application/text" subclass="true" display_in_upload="true" />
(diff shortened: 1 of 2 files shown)


astrovsky01:
The travis ci log is rather confusing. I actually don't know what to fix here...

config/datatypes_conf.xml.sample
```diff
@@ -736,6 +736,7 @@
     <datatype extension="dada2_mergepairs" type="galaxy.datatypes.binary:RData" subclass="true" display_in_upload="true" />
     <datatype extension="dada2_sequencetable" type="galaxy.datatypes.tabular:Tabular" mimetype="application/text" subclass="true" display_in_upload="true" />
     <datatype extension="dada2_uniques" type="galaxy.datatypes.tabular:Tabular" mimetype="application/text" subclass="true" display_in_upload="true" />
+    <datatype extension="ckpt" type="galaxy.datatypes.binary:Tensorflow" display_in_upload="true" />
```
nsoranzo: Since the class in binary.py it's basically empty, you can remove it and use `subclass="true"` like the `cxb` datatype.

astrovsky01: Awesome, thanks!

config/datatypes_conf.xml.sample
```diff
@@ -736,6 +736,7 @@
     <datatype extension="dada2_mergepairs" type="galaxy.datatypes.binary:RData" subclass="true" display_in_upload="true" />
     <datatype extension="dada2_sequencetable" type="galaxy.datatypes.tabular:Tabular" mimetype="application/text" subclass="true" display_in_upload="true" />
     <datatype extension="dada2_uniques" type="galaxy.datatypes.tabular:Tabular" mimetype="application/text" subclass="true" display_in_upload="true" />
+    <datatype extension="ckpt" type="galaxy.datatypes.binary:Tensorflow" subclass="true" display_in_upload="true" />
```
nsoranzo:
```suggestion
    <datatype extension="ckpt" type="galaxy.datatypes.binary:Binary" subclass="true" display_in_upload="true" />
```

and remove the changes to lib/galaxy/datatypes/binary.py

Sorry my first comment wasn't very clear.

lib/galaxy/datatypes/binary.py
```diff
@@ -2407,6 +2407,13 @@ def get_type(self):
         return "Sciex WIFF/SCAN archive"
 
 
+class Tensorflow(Binary):
```
nsoranzo, line 2410: Still need to remove this unused class.

astrovsky01:
@nsoranzo Anything else I'm missing?

lib/galaxy/datatypes/binary.py
bgruening: ping @astrovsky01, such a class is not needed as you specify subclass="True" in the datatypes_conf file.

mvdbeek: This class still needs to be removed @astrovsky01, this isn't resolved yet.

astrovsky01: @mvdbeek I was waiting while I discussed with Bjorn the security aspects of the checkpoint files, as they contain code that is run inside tensorflow. I can close this for now and reopen it when it is all resolved, if you would prefer

PR #7971, opened 2019-05-15
Unit test Galaxy packages during CI.

jmchilton:
All the same unit tests run in the context of a full Galaxy but this will ensure the tests also run with their dependencies as would be configured via pip-installed variants. This will catch dependency problems and ensure the package dependency structure remains correct.

Update: So I redid this on top of the new job metrics and tool_util packages and on top of #8051. I can't get galaxy-data or galaxy-tool-util quite tested right but testing this other stuff has already caught dependencies so I think this PR represents a good step forward and I'm taking it out of WIP.

xref #7524

### DIFF
--- a/lib/galaxy/util/web_compat.py
+++ b/lib/galaxy/util/web_compat.py
@@ -1,0 +1,10 @@
+"""Work around for gross circular dependency between galaxy.util and galaxy.web.stack.
+
+Provide a function that will delay to forking in a uwsgi environment but run immediately
+otherwise.
+"""
+try:
+    from galaxy.web.stack import register_postfork_function
+except ImportError:
+    def register_postfork_function(f, *args, **kwargs):
+        f(*args, **kwargs)
--- a/packages/test.sh
+++ b/packages/test.sh
@@ -17,9 +17,9 @@ PACKAGE_DIRS="
     util
     objectstore
     job_metrics
-    tool_util
 "
-# 'data' package doesn't yet work.
+# tool_util seems to need containers...
+# 'data' package doesn't yet work - quota, unit test problems, tool shed install database dependencies...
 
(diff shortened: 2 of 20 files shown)


packages/test.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+
+set -e
+
+TEST_PYTHON=${TEST_PYTHON:-"python"}
+TEST_ENV_DIR=${TEST_ENV_DIR:-`mktemp -d -t gxpkgtestenv`}
```
martenson: `XXXXXX` ?

mvdbeek (approved):

lib/galaxy/util/web_compat.py
```diff
@@ -0,0 +1,10 @@
+"""Work around for gross circular dependency between galaxy.util and galaxy.web.stack.
+
+Provide a function that will delay to forking in a uwsgi environment but run immediately
+otherwise.
+"""
+try:
+    from galaxy.web.stack import register_postfork_function
+except ImportError:
+    def register_postfork_function(f, *args, **kwargs):
```
mvdbeek, line 9: That'll go away with 2b6bccdcf7

mvdbeek (approved):
Great! Like 🦅 👀  I think we should start moving those tests over to py3 sooner rather than later :).

lib/galaxy/util/web_compat.py
mvdbeek: ah, we're still using it in lib/galaxy/util/monitors.py. Still, no problem.

jmchilton:
I'll open a PR to convert this to Python 3 - I don't even think about these things... obviously 😆... my bad.

Issue #7963, opened 2019-05-14
Workflow connection is broken when tool param or label updated


innovate-invent:
version 19.05
This occurs when a collection is sent to a multiple=true input of a tool. The tool is not mapped over so when I try to remake the connection without breaking the output I get 
![Screenshot_2019-05-14_14-12-27](https://user-images.githubusercontent.com/1768820/57732892-f917e780-7652-11e9-82a6-d116125f6314.png)

mvdbeek:
So this is intentional, and not so easy to fix. See https://github.com/galaxyproject/galaxy/issues/7431#issuecomment-478714812

martenson:
This seems like a duplicate of the issue linked above, let's discuss there please, closing.

innovate-invent:
@mvdbeek The issue isn't that I can't recreate the connection, the issue is that the mapping of the tool is unstable and is reevaluated any time you change a tool parameter. The code that is reevaluating the mapping appears to draw a different conclusion than the code that originally made the mapping, invalidating and breaking existing links.

mvdbeek:
And is that wrong ? If yes a reproducible example would be good.

innovate-invent:
Why would changing the tool label invalidate the tools connections?

mvdbeek:
it doesn't for me, that's why I am asking for an example.

mvdbeek:
OK, I see it, need to connect a list to a multiple=True and something to the output.

innovate-invent:
I can't readily locate a tool with a single input and a multiple=true input so you will have to patch the awk tool https://raw.githubusercontent.com/bgruening/galaxytools/bc9dc2fc8a806b312d679466d1ea581f3ded35ee/tools/text_processing/text_processing/awk.xml

![image](https://user-images.githubusercontent.com/1768820/57792716-8b23fc80-76f4-11e9-9d4f-c224c02deca6.png)


[Galaxy-Workflow-Tool_Connection_Break.ga.tar.gz](https://github.com/galaxyproject/galaxy/files/3183436/Galaxy-Workflow-Tool_Connection_Break.ga.tar.gz)

Creating this example it appears that it initially tries to map over a multiple=true input but once the connection is made it falls back to not mapped over

martenson:
Thanks for clarification and sorry for closing, I misread this as a duplicate.

mvdbeek:
Alright, this should be fixed by https://github.com/galaxyproject/galaxy/pull/7979

PR #7979, opened 2019-05-15
[19.05] Fix attachable for multiple="true" list input

mvdbeek:
This fixes https://github.com/galaxyproject/galaxy/issues/7963


jmchilton:
Awesome - thanks as always!

PR #7980, opened 2019-05-16
output_checker: fix superficial condition

bernt-matthias:
Follow up to https://github.com/galaxyproject/galaxy/pull/7966


bernt-matthias:
Should I also backport it to 19.05 (as follow up to https://github.com/galaxyproject/galaxy/pull/7969)

mvdbeek (approved):
The selenium test was fixed in https://github.com/galaxyproject/galaxy/pull/7986

PR #7983, opened 2019-05-16
[19.05] Fixed wrong example for amqp_ack_republish_time in job_conf.xml.sample_advanced 

AndreasSko:
"amqp_republish_time" should be "amqp_ack_republish_time" according to: https://github.com/galaxyproject/galaxy/blob/96d3f85372f074c71e73fea3bc822bd200f1127f/lib/galaxy/jobs/runners/pulsar.py#L110


dannon:
@AndreasSko Thanks!  I opened https://github.com/galaxyproject/pulsar/pull/185 to follow up on the pulsar side.

Issue #4813, opened 2017-10-16
Workflow Post Job Action of Change Datatype ignored when connecting tools in editor


blankenberg:
Tool 1 (FLASH) outputs 'fastq'. Tool 2 (FASTQ joiner) accepts fastqsanger (and others). 

Can extract a workflow from history with these two tools connected:
![image](https://user-images.githubusercontent.com/1231190/31634781-a3849cb2-b292-11e7-86b5-4db1106d14d7.png)

As soon as you add post job action of changing datatype to fastqsanger (which is actually accepted) the connections are destroyed and cannot be recreated. Same behavior when trying to create from scratch:
![image](https://user-images.githubusercontent.com/1231190/31634830-c4dcea7c-b292-11e7-9d43-c0e59ce06b56.png)

mvdbeek:
Funny enough it seems to work for the first output (merged_reads).

mvdbeek:
Should be fixed by https://github.com/galaxyproject/galaxy/pull/7989 / https://github.com/galaxyproject/galaxy/pull/8040

PR #7989, opened 2019-05-16
[19.05] Update tool output datatype in workflow editor if datatype changes

### LINKED ISSUE #7965: Changing output format on extract dataset tool has no effect in subworkflows
version 19.05
It seems to have no effect when the tool is flagged as the output of a subworkflow.
The subworkflow reports the output type as `data` despite being set to `newick` in the tool.
This prevents connections to other tools.

mvdbeek:
This adds the `force_dataype` key value pair into the step dict on the backend, which means that subworkflows outputs and tools that have a change datatype PJA will display the correct output datatype.
The client now also sets `force_datatype` when changing the PJA and updates the node's output label to indicate the new output datatype.

fixes #7965 and #4813


jmchilton:
Well this is pretty exciting, thanks a bunch @mvdbeek !

PR #7991, opened 2019-05-16
remove href from a tag to prevent backbone page reload and change som…

### LINKED ISSUE #7893: issues with history sharing
this is from Main

I am unable to edit URL of a shared history

cannot-edit

js console has `GET https://usegalaxy.org/api/histories/undefined/sharing 400 (Bad Request)`

bonus: the url has an extra `:` in the link name
Galaxy

**extra extra bonus** get rid of the custom yellow 'pencil icon' and replace with standard font awesome

WilliamHolden:
fixes #7893

### DIFF
--- a/client/galaxy/scripts/components/Sharing.vue
+++ b/client/galaxy/scripts/components/Sharing.vue
@@ -25,7 +25,7 @@
                         <span id="item-url-text" style="display: none">
                             {{ item_url_parts[0] }}<span id="item-identifier">{{ item_url_parts[1] }}</span>
                         </span>
-                        <a href="#" id="edit-identifier"><img :src="pencil_url"/></a>
+                        <a href="javascript:void(0)" id="edit-identifier"><img :src="pencil_url" alt="Edit Share Url"/></a>
                     </blockquote>
                     <div v-if="item.published">
                         <p>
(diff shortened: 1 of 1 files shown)


client/galaxy/scripts/components/Sharing.vue
```diff
@@ -25,7 +25,7 @@
                         <span id="item-url-text" style="display: none">
                             {{ item_url_parts[0] }}<span id="item-identifier">{{ item_url_parts[1] }}</span>
                         </span>
-                        <a href="#" id="edit-identifier"><img :src="pencil_url"/></a>
+                        <a id="edit-identifier"><img :src="pencil_url" alt="Edit Share Url"/></a>
```
martenson:
```suggestion
                        <a href="javascript:void(0)" id="edit-identifier"><img :src="pencil_url" alt="Edit Share Url"/></a>
```

martenson: links that don't have href are often styled differently

WilliamHolden: Good Point! Updated.

martenson:
@WilliamHolden you can save some force pushing work next time by just clicking on the commit button next to my suggestion. Thanks for incorporating!

blankenberg:
This should probably be cherry-picked to at least 19.05

PR #7993, opened 2019-05-16
Add mcool datatype and extend h5 tool output testing

msauria:
This PR does two things.
1. It adds a datatype for a multi-resolution cooler file (the cooler datatype is already in main).
2. It also adds a new assertion element for testing tool output. Currently, h5 output files can be tested by looking at attributes name and values, and keys, but only in the base group. This new test, 'contains_h5_keys' would allow specification of a subset of keys (currently the passed set of keys must exactly match all keys in the output) and allow drilling into groups so multilayered h5 files could be tested (e.g. 'base/subkey').


lib/galaxy/web/proxy/js/lib/proxy.js
```diff
@@ -91,6 +94,10 @@ DynamicProxy.prototype.findSession = function(request) {
 DynamicProxy.prototype.handleProxyRequest = function(req, res) {
     var othis = this;
     var target = this.targetForRequest(req);
+    console.log(req.method);
```
bgruening: This should be removed I think?

test/functional/tools/validation_hdf5.xml
```diff
@@ -13,7 +13,8 @@
             <param name="input" value="matrix.cool" />
             <output name="output">
                 <assert_contents>
-                    <has_h5_keys keys="bins, chroms,indexes,pixels" />
```
bgruening, line 16: The space was intended to test the stripping.

lib/galaxy/tools/xsd/galaxy.xsd
```diff
@@ -1652,6 +1652,8 @@ module.
       </xs:element>
       <xs:element name="has_h5_keys" type="AssertHasH5Keys">
       </xs:element>
+      <xs:element name="contains_h5_keys" type="AssertContainsH5Keys">
```
bgruening: has_ and contains_ is for a user probably the same. Could we extend the current has_ functionality?

lib/galaxy/datatypes/binary.py
```diff
@@ -1107,6 +1107,63 @@ def display_peek(self, dataset):
             return "Cool (HDF5) file (%s)." % (nice_size(dataset.get_size()))
 
 
+class MCool(H5):
```
bgruening, line 1110: Should this derive from Cool?

bgruening (commented):
Thanks @msauria. Small review inline.

ping @joachimwolff

lib/galaxy/web/proxy/js/lib/proxy.js
msauria: Agreed

test/functional/tools/validation_hdf5.xml
msauria: Ahh, that makes sense.

lib/galaxy/tools/xsd/galaxy.xsd
msauria: I suppose so. I wanted to avoid needing to specify a large number of keys for complex h5 files, but given that this is only invoked for tests, I suppose test output can be controlled to keep this list reasonable.

lib/galaxy/datatypes/binary.py
msauria: That is unclear to me. The structure of the mcool file is a series of cool files nested as groups within the mcool file, meaning that the only function specific to the cool class (the sniffer) doesn't apply to the mcool class. But for the sake of suggesting the relationship, I guess that makes sense.

lib/galaxy/tools/xsd/galaxy.xsd
msauria: Also, do you mean altering the 'has_h5_keys' test? Doing so would definitely break any tests currently using it if they have groups in the h5 files.

jxtx: I was reading this as change `has_h5_keys` to not require a complete exact match. Which makes sense, "has" does not mean "has only".

bgruening: What @jxtx said is what I wanted to say :) Sorry.

msauria: Done

lib/galaxy/datatypes/binary.py
```diff
@@ -1107,6 +1107,63 @@ def display_peek(self, dataset):
...
+        >>> fname = get_test_fname('matrix.mcool')
+        >>> MCool().sniff(fname)
+        True
+        >>> fname = get_test_fname('test.mz5')
+        >>> MCool().sniff(fname)
+        False
+        >>> fname = get_test_fname('wiggle.wig')
+        >>> MCool().sniff(fname)
+        False
+        >>> fname = get_test_fname('biom2_sparse_otu_table_hdf5.biom2')
```
bgruening, line 1132: can we add here a cool test what will eval to False?

msauria: Yes. Done

lib/galaxy/datatypes/binary.py
msauria: Turns out this should not derive from Cool as this causes a failure of the sniffer

bgruening (approved):
Thanks Mike!

PR #8001, opened 2019-05-17
Don't close popup menu if header item is clicked

### LINKED ISSUE #7105: Section names in history dropdown should not be clickable
grafik

"Current History" there is clickable + closes the menu. it probably should not be clickable and should not close the menu if you accidentally click it.

WilliamHolden:
Also upgrade to es6 features and removed unused functions.

fixes #7105


hexylena:
Thanks! @WilliamHolden

PR #8009, opened 2019-05-20
Change username to public name

### LINKED ISSUE #4359: Standardize "Public Name" label on all Galaxy GUI forms
**Username** is the label for the value on the account login form.

**Public name** is the label for the value on the account registration form.

**Public name** is the label for the value once logged in and User > Preferences is reviewed and/or modified.

**All should be "Public name"** - to clearly distinguish between "username" that is common nomenclature in FTP clients where the account email address is the expected input. There may be other places in the Galaxy GUI where "[U/u]sername" is present - the intention of this change is to make all consistently labeled as "[P/p]ublic name"

WilliamHolden:
fixes #4359


dannon:
Thanks for standardizing this and for the cleanup.

Issue #7957, opened 2019-05-14
sniffer problem in 19.05?


jmchilton:
From internal bug list:

```
> Attempt to upload GTF from pseudomonas.com (genome resource) failed with
> the error above. Not sure how to proceed:
> Fatal error: Exit code 1 ()
> Traceback (most recent call last):
>   File "/cvmfs/main.galaxyproject.org/galaxy/tools/data_source/upload.py",
> line 326, in <module>
>     __main__()
>   File "/cvmfs/main.galaxyproject.org/galaxy/tools/data_source/upload.py",
> line 319, in __main__
>     metadata.append(add_file(dataset, registry, output_path))
>   File "/cvmfs/main.galaxyproject.org/galaxy/tools/data_source/upload.py",
> line 128, in add_file
>     convert_spaces_to_tabs=dataset.space_to_tab,
>   File "/cvmfs/
> main.galaxyproject.org/galaxy/lib/galaxy/datatypes/upload_util.py", line
> 54, in handle_upload
>     convert_spaces_to_tabs=convert_spaces_to_tabs,
>   File "/cvmfs/main.galaxyproject.org/galaxy/lib/galaxy/datatypes/sniff.py",
> line 770, in handle_uploaded_dataset_file_internal
>     line_count, _converted_path = convert_fxn(converted_path,
> in_place=in_place, tmp_dir=tmp_dir, tmp_prefix=tmp_prefix)
>   File "/cvmfs/main.galaxyproject.org/galaxy/lib/galaxy/datatypes/sniff.py",
> line 122, in convert_newlines
>     for i, line in enumerate(io.open(fname, mode="U", encoding='utf-8')):
>   File "/cvmfs/main.galaxyproject.org/venv/lib/python2.7/codecs.py", line
> 314, in decode
>     (result, consumed) = self._buffer_decode(data, self.errors, final)
> UnicodeDecodeError: 'utf8' codec can't decode byte 0xa0 in position 690:
> invalid start byte
```

mvdbeek:
The problem is that we do the newline / space to tab conversion for text datatypes, but that fails with non-utf8 characters. So we could try reading this in `rb` mode instead, but there's a good chance downstream tools would fail. Automatic recoding isn't perfect either. Any ideas here @nsoranzo  ?

mvdbeek:
[...]

PR #8010, opened 2019-05-21
[19.05] Use byte comparison for contains and files_re_match

mvdbeek:
I don't think there's any drawback opening this in byte mode. ~~WIP because I want to add some tests before merging. Also the code looks more complicated than necessary.~~

Fixes https://github.com/galaxyproject/galaxy/issues/7957#issuecomment-493954901

### DIFF
--- a/lib/galaxy/tools/verify/__init__.py
+++ b/lib/galaxy/tools/verify/__init__.py
@@ -3,7 +3,6 @@
 import difflib
 import filecmp
 import hashlib
-import io
 import logging
 import os
 import os.path
@@ -268,9 +267,9 @@ def files_diff(file1, file2, attributes=None):
 
 def files_re_match(file1, file2, attributes=None):
     """Check the contents of 2 files for differences using re.match."""
-    local_file = io.open(file1, encoding='utf-8').readlines()  # regex file
-    history_data = io.open(file2, encoding='utf-8').readlines()
-    assert len(local_file) == len(history_data), 'Data File and Regular Expression File contain a different number of lines (%d != %d)\nHistory Data (first 40 lines):\n%s' % (len(local_file), len(history_data), ''.join(history_data[:40]))
+    local_file = open(file1, 'rb').readlines()  # regex file
+    history_data = open(file2, 'rb').readlines()
+    assert len(local_file) == len(history_data), b'Data File and Regular Expression File contain a different number of lines (%d != %d)\nHistory Data (first 40 lines):\n%s' % (len(local_file), len(history_data), b''.join(history_data[:40]))
     if attributes is None:
         attributes = {}
     if attributes.get('sort', False):
@@ -279,24 +278,24 @@ def files_re_match(file1, file2, attributes=None):
     line_diff_count = 0
     diffs = []
     for i in range(len(history_data)):
-        if not re.match(local_file[i].rstrip('\r\n'), history_data[i].rstrip('\r\n')):
+        if not re.match(local_file[i].rstrip(b'\r\n'), history_data[i].rstrip(b'\r\n')):
             line_diff_count += 1
-            diffs.append('Regular Expression: %s\nData file         : %s' % (local_file[i].rstrip('\r\n'), history_data[i].rstrip('\r\n')))
+            diffs.append(b'Regular Expression: %s\nData file         : %s' % (local_file[i].rstrip(b'\r\n'), history_data[i].rstrip(b'\r\n')))
         if line_diff_count > lines_diff:
-            raise AssertionError("Regular expression did not match data file (allowed variants=%i):\n%s" % (lines_diff, "".join(diffs)))
+            raise AssertionError(b"Regular expression did not match data file (allowed variants=%i):\n%s" % (lines_diff, b"".join(diffs)))
 
 
 def files_re_match_multiline(file1, file2, attributes=None):
     """Check the contents of 2 files for differences using re.match in multiline mode."""
-    local_file = io.open(file1, encoding='utf-8').read()  # regex file
+    local_file = open(file1, 'rb').read()  # regex file
     if attributes is None:
         attributes = {}
     if attributes.get('sort', False):
-        history_data = io.open(file2, encoding='utf-8').readlines()
+        history_data = open(file2, 'rb').readlines()
         history_data.sort()
-        history_data = ''.join(history_data)
+        history_data = b''.join(history_data)
     else:
-        history_data = io.open(file2, encoding='utf-8').read()
+        history_data = open(file2, 'rb').read()
     # lines_diff not applicable to multiline matching
     assert re.match(local_file, history_data, re.MULTILINE), "Multiline Regular expression did not match data file"
 
--- a/lib/galaxy/tools/verify/__init__.py
+++ b/lib/galaxy/tools/verify/__init__.py
@@ -3,6 +3,7 @@
 import difflib
 import filecmp
 import hashlib
+import io
 import logging
 import os
 import os.path
@@ -15,6 +16,7 @@ try:
 except ImportError:
     pysam = None
 
+from galaxy.util import unicodify
 from galaxy.util.compression_utils import get_fileobj
 from .asserts import verify_assertions
 from .test_data import TestDataResolver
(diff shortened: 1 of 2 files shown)


lib/galaxy/tools/verify/__init__.py
```diff
@@ -279,37 +278,37 @@ def files_re_match(file1, file2, attributes=None):
...
-    local_file = io.open(file1, encoding='utf-8').readlines()  # regex file
+    local_file = open(file1, 'rb').readlines()  # regex file
     # TODO: allow forcing ordering of contains
-    history_data = io.open(file2, encoding='utf-8').read()
+    history_data = open(file2, 'rb').read()
     lines_diff = int(attributes.get('lines_diff', 0))
     line_diff_count = 0
     while local_file:
-        contains = local_file.pop(0).rstrip('\n\r')
+        contains = local_file.pop(0).rstrip(b'\n\r')
```
nsoranzo: This `while` looks weird, I'd replace the above 2 lines with:
```python
    for contains in local_file:
        contains = contains.rstrip(b'\n\r')
```

nsoranzo (commented):
I'm not sure I like this approach, in particular the bytes exception messages. Can we try to read the files as Unicode and if that fails, resort to binary mode?
So the exception messages can remain proper strings.

mvdbeek:
> I'm not sure I like this approach, in particular the bytes exception messages. Can we try to read the files as Unicode and if that fails, resort to binary mode ?

I get the hesitation, but reading data as unicode and then wrapping this with a byte fallback seems equally silly and makes for yet another path through the code that can fail.

> So the exception messages can remain proper strings.

If that's the concern we could also use unicodify on all Exceptions, do you think that'd be better ?

mvdbeek:
@nsoranzo I changed the file reading to `errors="replace"`, what do you think about that ?

nsoranzo:
> I changed the file reading to `errors="replace"`, what do you think about that ?

I guess it depends which problem are we trying to solve: if we want to allow proper checking of binary files (contain, regexps), then this won't work.

mvdbeek:
I personally only want to fix https://github.com/galaxyproject/galaxy/issues/7957#issuecomment-493954901, but of course covering whatever makes sense. 

How about keeping contains and regex comparisons in bytes but do `lines_diff` in unicode ? I don't think `difflib.unified_diff` would work with non-utf8 characters (https://docs.python.org/3/library/difflib.html#difflib.unified_diff) which we could reasonably expect in some non-binary files (like pdf). Running `unicodify` would strip those characters and we'd still get a meaningful result.

I tried this in https://github.com/galaxyproject/galaxy/pull/8010/commits/a398127ede922d504ec7f1c39e66df22a9f0a16d

peterjc:
Unicode when looking at differences in lines makes sense (in this context should not have a binary file).

Raw bytes seems best when looking at file contains in a test (as used in the PDF example).

Also if we offered it, raw bytes for file starts with (that was the intent of my the PDF example - equally useful in PNG, SFF or something where there is a standard magic header).

nsoranzo:
@mvdbeek @peterjc I've pushed a commit with my idea of how to try to read the files first as Unicode and resort to binary mode if that fails with `UnicodeDecodeError`. I don't think it increases the code complexity too much, and it allows to do all the various checks for both text and binary files.

Let me know what do you think!

nsoranzo (approved):

lib/galaxy/tools/verify/__init__.py
```diff
@@ -205,12 +206,21 @@ def get_lines_diff(diff):
...
+                local_file = fh.readlines()
+        except UnicodeDecodeError:
+            if file1.endswith('.pdf') or file2.endswith('.pdf'):
+                is_pdf = True
+                # Replace non-Unicode characters using unicodify(),
+                # difflib.unified_diff doesn't work on list of bytes
+                history_data = [unicodify(l) for l in get_fileobj(file2, mode='rb', compressed_formats=compressed_formats)]
+                local_file = [unicodify(l) for l in get_fileobj(file1, mode='rb', compressed_formats=compressed_formats)]
+            else:
+                raise AssertionError("Binary data detected, not displaying diff")
```
mvdbeek, line 223: This wouldn't work for the last example in the unit test if the extension isn't `.pdf`. This effectively makes non-utf8 encoded text uncomparable using `lines_diff`.  And it used to be comparable prior to our py3 porting.

lib/galaxy/tools/verify/__init__.py
```diff
@@ -263,9 +273,21 @@ def get_lines_diff(diff):
 
 def files_re_match(file1, file2, attributes=None):
     """Check the contents of 2 files for differences using re.match."""
-    local_file = open(file1, 'rb').readlines()  # regex file
-    history_data = open(file2, 'rb').readlines()
-    assert len(local_file) == len(history_data), 'Data File and Regular Expression File contain a different number of lines (%d != %d)\nHistory Data (first 40 lines):\n%s' % (len(local_file), len(history_data), ''.join(unicodify(history_data[:40])))
+    join_char = ''
+    to_strip = os.linesep
```
mvdbeek, line 277: This isn't correct, we want to always strip carriage returns here, `\n` is already dealt with using `readlines`.

lib/galaxy/tools/verify/__init__.py
mvdbeek: Actually that's OK, we already failed the lines diff at this point ...

lib/galaxy/tools/verify/__init__.py
peterjc: Do we care about Windows where ``os.linesep`` would be ``\r\n`` (CR LF)? Probably not.

However, more realistically, we probably should care about files using Windows new lines - the Python default universal text mode ought to take care of that case for us (although we'd have to double check Python 2 vs 3 there), that might be a good reason not to use binary mode for the line-based ``re`` check?

peterjc:
Noted as a line diff, but I would worry about Unix (CR) vs DOS/Windows (CR LF) line endings if you want to handle line-based regular expressions in binary mode.

Using text mode and letting Python handle the cross-platform new line conventions could be simpler?

lib/galaxy/tools/verify/__init__.py
mvdbeek: We actually want to always strip `\r\n` here, independent of what platform we're on. If old test data ends with `\r\n` and new data produced by a tool run is `\n` we don't want to fail IMO. We haven't in the past, and I don't think we should introduce this now.

mvdbeek: Actually `\r\n` is being stripped by readlines, it'll just end up as `\n`, so that's actually the correct change. ...

peterjc:
With the explicit inclusion of both Unix and DOS/Windows new lines in the tests, I'm happy. Thanks!

mvdbeek:
@nsoranzo thanks for the commit, I was still making some wrong assumptions there. It's OK for me now.

nsoranzo:
@mvdbeek Thanks for double-checking and adding the carriage return test! I'll merge when tests pass.

peterjc:
Will we need another pull request to get this work onto the ``dev`` branch (which is what I've been testing against with the Venn Diagram tool producing a PDF file)?

mvdbeek:
@nsoranzo merged 19.05 into dev in https://github.com/galaxyproject/galaxy/commit/0e14a47e6c29f78c8cfca478feb05c1f0d6d4ab9, so if you run the test it should hopefully pass now

peterjc:
Strange - I can see the changes now, but checked earlier and didn't. Most likely my mistake (although I saw a brief GitHub outage earlier today). Thanks.

mvdbeek:
Yeah, github's been showing some outdated changes yesterday as well.
I just ran the tests for https://github.com/peterjc/pico_galaxy/pull/33:
```
All 3 test(s) executed passed.
venn_list[0]: passed
venn_list[1]: passed
venn_list[2]: passed
```

So maybe that was just a intermittent issue ?

peterjc:
Fingers crossed - rerunning the TravisCI check now (it grabs Galaxy from the specified branch on GitHub).

peterjc:
Worked this time - thank you 👍

Issue #6156, opened 2018-05-16
Relabel List Identifiers from contents of a file on collection fails without telling user why


blankenberg:
Get a job submission error message, which is good. But it doesn't give any feed back to the user on why, so that they might fix it.

![image](https://user-images.githubusercontent.com/1231190/40140528-2fbfef3a-5921-11e8-89c4-6816f9f5d128.png)

From log, some reasons may be:

```
galaxy.tools DEBUG 2018-05-16 15:48:42,977 [p:1970348,w:4,m:0] [uWSGIWorker4Core2] Validated and populated state for tool request (3112.794 ms)
galaxy.tools ERROR 2018-05-16 15:48:49,352 [p:1970348,w:4,m:0] [uWSGIWorker4Core2] Exception caught while attempting tool execution:
Traceback (most recent call last):
  File "lib/galaxy/tools/__init__.py", line 1416, in handle_single_execution
    completed_job=completed_job,
  File "lib/galaxy/tools/__init__.py", line 1497, in execute
    return self.tool_action.execute(self, trans, incoming=incoming, set_output_hid=set_output_hid, history=history, **kwargs)
  File "lib/galaxy/tools/actions/model_operations.py", line 57, in execute
    self._produce_outputs(trans, tool, out_data, output_collections, incoming=incoming, history=history, tags=preserved_tags)
  File "lib/galaxy/tools/actions/model_operations.py", line 71, in _produce_outputs
    tool.produce_outputs(trans, out_data, output_collections, incoming, history=history, tags=tags)
  File "lib/galaxy/tools/__init__.py", line 2641, in produce_outputs
    new_labels_dict = {source: new_label for source, new_label in source_new_label}
  File "lib/galaxy/tools/__init__.py", line 2641, in <dictcomp>
    new_labels_dict = {source: new_label for source, new_label in source_new_label}
ValueError: too many values to unpack
galaxy.tools.execute WARNING 2018-05-16 15:48:49,359 [p:1970348,w:4,m:0] [uWSGIWorker4Core2] There was a failure executing a job for tool [__RELABEL_FROM_FILE__] - Error executing tool: too many values to unpack
galaxy.tools.execute DEBUG 2018-05-16 15:48:49,359 [p:1970348,w:4,m:0] [uWSGIWorker4Core2] Executed 1 job(s) for tool __RELABEL_FROM_FILE__ request: (6382.382 ms)
```
[...]

PR #8012, opened 2019-05-21
Update help text for Relabel List Identifiers

sneumann:
This addresses https://github.com/galaxyproject/galaxy/issues/6156#issuecomment-494290527.
The rest of #6156 might require python-hacking. Yours, Steffen

### DIFF
--- a/lib/galaxy/tools/relabel_from_file.xml
+++ b/lib/galaxy/tools/relabel_from_file.xml
@@ -147,6 +147,9 @@
         in the collection. In that case only matching list identifiers will be
         relabeled.
 
+        Valid identifiers must contain only characters (a-z, A-Z), numbers (0-9), 
+        dash and underscore (-, _). Punctuation or whitescape are not allowed. 
+
         This tool will create new history datasets from your collection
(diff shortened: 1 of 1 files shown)


bernt-matthias:
Maybe we can use a regexp similar to the one used here: https://github.com/galaxyproject/galaxy/pull/7937

lib/galaxy/tools/relabel_from_file.xml
```diff
@@ -147,6 +147,9 @@
         in the collection. In that case only matching list identifiers will be
         relabeled.
 
+        Valid identifiers must contain only characters (a-z, A-Z), numbers (0-9), 
+        dash and underscore (-, _). Punctuation or whitescape are not allowed.
```
mvdbeek, line 151:
```suggestion
        dash and underscore (-, _). Punctuation or whitespace are not allowed. 
```

mvdbeek (approved):

mvdbeek:
We've merged this as part of https://github.com/galaxyproject/galaxy/pull/8038, thanks a lot @sneumann!

Issue #8011, opened 2019-05-21
Galaxy not always recognizes that job from Pulsar is finished


AndreasSko:
Hi,
I'm relatively new to Galaxy and right now I'm working on getting Galaxy (Version 19.01) to work with Pulsar (Version 0.11.0) and RabbitMQ (https://github.com/AndreasSko/GalaxyToPulsarInfrastructure). It already works most of the time but I have the problem, that Galaxy arbitrarily doesn't recognize, that a Job is already finished (I start the same job multiple times, some jobs succeed without any issue, some stay in waiting-status and some stay in running-status)..
I logged the messages in RabbitMQ and it seems that Pulsar sends the complete-message and Galaxy receives and even acknowledges it, but Galaxy still says that its "waiting to run" or "currently running" - so it kind of just ignores some status updates. 

I tried to dig deeper into the problem and I think that  https://github.com/galaxyproject/galaxy/blob/3dbab3dab23732f6b88088a5a2964e5b0540827b/lib/galaxy/jobs/runners/pulsar.py#L764 is not always called after a new message arrived. So could it have something to do with the threading in https://github.com/galaxyproject/pulsar/blob/master/pulsar/client/manager.py ? Something else? Or could it be just a wrong configuration on my side? (https://github.com/AndreasSko/GalaxyToPulsarInfrastructure/blob/master/ansible/files/pulsar/app.yml and https://github.com/AndreasSko/GalaxyToPulsarInfrastructure/blob/master/ansible/files/galaxy/job_conf.xml).

PR #8017, opened 2019-05-22
Fixed missing job_state-updates for a complete job in pulsar.py

AndreasSko:
The two statements were missing. This fixes https://github.com/galaxyproject/galaxy/issues/8011

### DIFF
--- a/lib/galaxy/jobs/runners/pulsar.py
+++ b/lib/galaxy/jobs/runners/pulsar.py
@@ -253,6 +253,8 @@ class PulsarJobRunner(AsynchronousJobRunner):
 
     def _update_job_state_for_status(self, job_state, pulsar_status, full_status=None):
         if pulsar_status == "complete":
+            job_state.running = False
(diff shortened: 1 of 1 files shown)


lib/galaxy/jobs/runners/pulsar.py
```diff
@@ -253,6 +253,8 @@ def check_watched_item(self, job_state):
 
     def _update_job_state_for_status(self, job_state, pulsar_status, full_status=None):
         if pulsar_status == "complete":
+            job_state.running = False
+            job_state.job_wrapper.change_state(model.Job.states.OK)
```
jmchilton, line 257: I don't know why the Galaxy finish code isn't being called in your cases, but we really can't set OK here. There is a lot of code in `jobs/runners/__ini__.py` that still needs to run for these jobs - Galaxy needs to determine if they ran okay, needs to collect metadata, etc..

Maybe put some debugging in ``run_next`` in job runners to see if the jobs really do get queued for completion in the following line?

PR #8018, opened 2019-05-22
Fix tool run uses ellipses slightly too soon

### LINKED ISSUE #7984: Tool run uses ellipses slightly too soon
With 4 items probably makes sense to print all four since there's room for 3 + ellipsis? And only if ≥5 then use ellipses

image

WilliamHolden:
fixes #7984

### DIFF
--- a/client/galaxy/scripts/mvc/tool/tool-form.js
+++ b/client/galaxy/scripts/mvc/tool/tool-form.js
@@ -67,15 +67,15 @@ const View = Backbone.View.extend({
                                     );
                                 } else {
                                     Galaxy.modal &&
-                                    Galaxy.modal.show({
-                                        title: _l("Tool request failed"),
-                                        body: error_message,
-                                        buttons: {
-                                            Close: () => {
-                                                Galaxy.modal.hide();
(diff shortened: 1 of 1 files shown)


client/galaxy/scripts/mvc/tool/tool-form.js
```diff
@@ -68,15 +68,15 @@ var View = Backbone.View.extend({
                                     );
                                 } else {
                                     Galaxy.modal &&
```
dannon: @guerler Do you remember why we needed to have this guard in here?  Is it irrelevant now, and safe to drop?

martenson:
switching to WIPdue to https://github.com/galaxyproject/galaxy/issues/7984#issuecomment-499227590

client/galaxy/scripts/mvc/tool/tool-form.js
guerler: Yes, it should be safe to remove that safe guard now.

martenson:
Galaxy

dannon (approved):

Issue #8015, opened 2019-05-21
Loading tools '_sa_instance_state' AttributeError


almahmoud:
In the context of running Galaxy on K8S (on GKE) and loading tools from the CVMFS `main.galaxyproject.org` repository, I switched from `19.01` to `19.05` to be able to load tools without the install DB (which worked in that I moved past the "Attempted to load from database" error), but this error came up for most of the tools (repeated roughly 2000 times in the log):
```
galaxy.tools.toolbox.base ERROR 2019-05-21 18:57:13,924 [p:1,w:0,m:0] [MainThread] Error reading tool from path: toolshed.g2.bx.psu.edu/repos/devteam/sam_pileup/890d97772e2a/sam_pileup/sam_pileup.xml
Traceback (most recent call last):
  File "/galaxy/server/lib/galaxy/tools/toolbox/base.py", line 605, in _load_tool_tag_set
    tool_shed_repository = self.get_tool_repository_from_xml_item(item, path)
  File "/galaxy/server/lib/galaxy/tools/toolbox/base.py", line 652, in get_tool_repository_from_xml_item
    installed_changeset_revision=installed_changeset_revision)
  File "/galaxy/server/lib/galaxy/tools/__init__.py", line 339, in _get_tool_shed_repository
    installed_changeset_revision=installed_changeset_revision
  File "/galaxy/server/lib/tool_shed/util/repository_util.py", line 374, in get_installed_repository
    repository_id=repository_id)
  File "/galaxy/server/lib/galaxy/tools/cache.py", line 165, in get_installed_repository
    repository_id=repository_id)
  File "/galaxy/server/lib/galaxy/tools/cache.py", line 182, in _get_installed_repository
    repos = [repo for repo in self.tool_shed_repositories if repo.tool_shed == tool_shed and repo.owner == owner and repo.name == name]
  File "/galaxy/server/lib/galaxy/tools/cache.py", line 150, in tool_shed_repositories
    if repositories and not repositories[0]._sa_instance_state._attached:
AttributeError: 'ToolConfRepository' object has no attribute '_sa_instance_state'
```
Full log dumped here, in case it's needed/relevant: https://gist.github.com/almahmoud/6284906a7d1be4fa7c785b08630a9e92

Following a conversation with @natefoo , I verified that we are using the correct version (image built by using ansible-galaxy role with mostly default settings and commit_id `release_19.05`), and tagging @mvdbeek  

Let me know if you need more info!

[...]

PR #8021, opened 2019-05-23
[19.05] Fix repo cache when no install database is present

mvdbeek:
Fixes https://github.com/galaxyproject/galaxy/issues/8015

### DIFF
--- a/lib/galaxy/tools/cache.py
+++ b/lib/galaxy/tools/cache.py
@@ -147,7 +147,7 @@ class ToolShedRepositoryCache(object):
         except AttributeError:
             self.rebuild()
             repositories = self.cache.repositories
-        if repositories and not repositories[0]._sa_instance_state._attached:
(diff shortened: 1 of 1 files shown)


lib/galaxy/tools/cache.py
```diff
@@ -147,7 +147,7 @@ def tool_shed_repositories(self):
         except AttributeError:
             self.rebuild()
             repositories = self.cache.repositories
-        if repositories and not repositories[0]._sa_instance_state._attached:
+        if repositories and any(_ for _ in repositories if hasattr(_, '_sa_instance_state') and not _._sa_instance_state.attached):
```
```suggestion
        if repositories and any(_ for _ in repositories if hasattr(_, '_sa_instance_state') and not _._sa_instance_state._attached):
```

or we can use a more stable API like https://docs.sqlalchemy.org/en/13/orm/session_state_management.html#getting-the-current-state-of-an-object

nsoranzo (approved):

natefoo (approved):

PR #7994, opened 2019-05-17
[WIP] "Galaxy Flavored" Markdown & Workflow Invocation Report Plugin Framework

jmchilton:
Screen Shot 2019-05-19 at 9 33 09 PM

Screen Shot 2019-05-19 at 9 33 19 PM

Screen Shot 2019-05-19 at 9 33 24 PM

Screen Shot 2019-05-19 at 9 33 30 PM

## Done so far:

- [X] Controller to generate markdown for invocations and send it to the client.
- [X] Database migration to add report configuration to workflow model.
- [X] Plugin plumbing on the backend to allow potentially other approaches.
- [X] Add client library to render markdown with plugins (markdown-it and markdown-it-container).
- [X] Implement a VueJS component that utilizes markdown-it and defines Galaxy-specific extensions for embedding Galaxy objects.
- [X] Implement a VueJS component and AnalysisRouter entry point for fetching report Markdown and rendering it using the above VueJS component.
- [X] Load report configuration from workflow imports (just top-level ``report:`` key in Galaxy Format 2 workflows or in ``.ga`` dict).
- [x] Implement Galaxy flavored markdown - custom Markdown containers (directives that start with ``:::``) that hook into components for display of Galaxy model objects. Keep it in markdown so PDF and archives can be built outside context of Vue and front-end. ``ref`` is ``history_dataset_id=encoded_id``, ``history_dataset_collection_id=encoded_id``, ``workflow_id=encoded_id``, or ``job_id=encoded_id``.
  - [X] ``::: history_dataset_display ref`` (reuses embedded object rendering code from pages)
  - [X] ``::: history_dataset_as_image ref``
  - [X] ``::: history_dataset_peek ref``
  - [X] ``::: history_dataset_info ref``
  - [X] ``::: workflow_display workflow_id=id`` (reuses embedded object rendering code from pages)
  - [x] ``::: history_dataset_collection_display ref``
  - [X] ``::: job_metrics job_id=id``
  - [x] ``::: tool_stdout ref``
  - [x] ``::: tool_stderr ref``
[...]

PR #8022, opened 2019-05-23
Rework dataset job properties tables into components+APIs

jmchilton:
Move job metrics and tool parameters tables out of mako and into reusable VueJS components, with matching APIs.

Both the APIs and the VueJS interfaces can operate using either job_id or a dataset_id (explained in the API docstrings - but we authorize jobs and datasets a little bit differently so it makes sense in my mind to allow specifying either). The show_params.mako usage of these components uses the dataset variant of these, https://github.com/galaxyproject/galaxy/pull/7994 demonstrates reusing these components using a job id instead. Not demonstrated here, but I think this component would work with library datasets now also as part of these generalizations.

I made some small visual improvements using the greater structure and reactivity provided by VueJS. For instance, the "note for rerun" column takes up a lot of real estate in this form and very rarely ever has any content. That column is now only rendered at all if any of the rows of the table actually contain a note.

The show parameters API endpoint is marked as unstable, this is because it is deeply tied to client rendering. This isn't great, but I think it is clearly a step forward from the mako that it replaces in a few ways (the Python code is being linted, the JS code is being linted, less mako overall, reusable component).

---

Allows reuse in #7994 for workflow invocation report generation:

Screen Shot 2019-05-23 at 7 25 47 AM

### DIFF
--- a/client/galaxy/scripts/components/JobParameters/mount.js
+++ b/client/galaxy/scripts/components/JobParameters/mount.js
@@ -1,0 +1,20 @@
+/**
+ * Endpoint for mounting job metrics from non-Vue environment.
+ */
+import $ from "jquery";
+import Vue from "vue";
+import JobParameters from "./JobParameters.vue";
+
+export const mountJobParameters = propsData => {
+    $(".job-parameters").each((index, el) => {
+        const component = Vue.extend(JobParameters);
+        const jobId = $(el).attr("job_id");
+        const datasetId = $(el).attr("dataset_id");
+        const datasetType = $(el).attr("dataset_type") || "hda";
+        propsData = propsData || {};
+        propsData["jobId"] = jobId;
+        propsData["datasetId"] = datasetId;
+        propsData["datasetType"] = datasetType;
+        return new component({ propsData: propsData }).$mount(el);
+    });
+};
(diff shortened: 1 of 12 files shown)


client/galaxy/scripts/components/JobParameters/mount.js
```diff
@@ -1,20 +1,19 @@
 /**
- * Endpoint for mounting job metrics from non-Vue environment.
+ * Endpoint for mounting job parameters from non-Vue environment.
```
jmchilton, line 2: Thanks @dannon - looks great!

dannon: No, thank you!

inkuzmin:
@jmchilton wow, fantastic work! It was quite educational for me to check it.

Just to clarify for myself some adopted conventions:

- https://github.com/galaxyproject/galaxy/pull/8022/files#diff-97fa54345b7ba851960e6a1b26ccc0beR14 -- would it make sense to use ES6 default parameters for this?

- https://github.com/galaxyproject/galaxy/pull/8022/files#diff-97fa54345b7ba851960e6a1b26ccc0beR18 -- do we capitalize constructors?

- https://github.com/galaxyproject/galaxy/pull/8022/files#diff-c7d20ce6d8ae40bc11c6d254245ad4dfR114 -- may be it's better to use component style scoping?

I didn't really understand this method:

- https://github.com/galaxyproject/galaxy/pull/8022/files#diff-c7d20ce6d8ae40bc11c6d254245ad4dfR91 -- wouldn't it always return `false` (`0 ^ x = 0`)?

jmchilton:
I'm not the person to ask about client conventions, I'm sure I got them wrong. I didn't even know component style scoping was an option and that property is definitely broken, I'll try to open a PR to fix it.

martenson:
Component style scope is very very good.

PR #8023, opened 2019-05-23
Fix column assign bug

mvdbeek:
If columns are specified as integers this would fail currently.
Reported by Peter Briggs on the [mailing list](https://lists.galaxyproject.org/archives/list/galaxy-dev@lists.galaxyproject.org/thread/PXLCVKIZ5LW3MLI6333PFMANYGOCDLD2/)

Test depends on https://github.com/jmchilton/gxformat2/pull/8


martenson:
Thanks!

Issue #5389, opened 2018-01-25
Archival and replication of user analyses


mvdbeek:
This was brought up on gitter by @lparsons, and we also had questions about this at the admin workshop in Oslo.

I am currently seeing 2 possible routes to archive an analysis in Galaxy, and they should be documented better, probably in the admin section of the GTN.

To me the ideal archival strategy would be to work exclusively with workflows and to keep input datasets, a workflow, and something resembling the workflow invocation, so that the analysis can be faithfully reproduced, including workflow parameters. 

Another possibility is exporting the entire history, which currently has the drawback of not supporting collection (there's some work in https://github.com/galaxyproject/galaxy/pull/4532),
and not all analyses are confined to a single history.

jennaj:
This is a good idea. Similiar to what I keep on a wish-list. Maybe some of the ideas would be useful: https://github.com/jennaj/support-known-issues/wiki/History-usability-ideas

mblue9:
This is a great idea. I really need a way for users to export a record of what was actually run (versions, parameters, input dataset names...). Is this the issue for that?

Being able to extract a record of the workflow invocation would be good but if some steps are run outside the workflow then being able to extract from the history (aswell) would be good.

These ideas in Jen's list sound like what I'm looking for:

>Extract history structure similar to how extract workflow works now. Elements clickable same as in workflow canvas flavored view. Note: History menu > Show structure results in a server error (and has for a while)

This would be great!:
>Output the history content - datasets, tool version/paramaters, status, etc - into a tab-delimited file.

Takadonet:
I have hired a four month co-op student to work on ability to export a record(provenance) of a history in both a machine and human readable formats.

mblue9:
OMG @Takadonet that is some great news!!!! 🙌 🎉

hexylena:
Any results from them @Takadonet? :)

Takadonet:
We had a prototype working by end of the term. It did not work well with collections which is most of our workflows that we run at our lab.

Believe the route we are leading toward will be using [John's Workflow invocation](https://github.com/galaxyproject/galaxy/pull/8030)

jennaj:
This came up at Galaxy Help the other day.

Even a PDF type of output would be some improvement. But I really really like the option of outputting into some type of formatted tab file that could be parsed. 

Versioning would be an important consideration here, too ... 

https://help.galaxyproject.org/t/can-i-download-a-printable-version-of-a-workflow/2076

PR #8030, opened 2019-05-23
[WIP] Implement workflow invocation reports.

jmchilton:
Implement markdown backend and frontend components as well generator plugin framework to allow customizable workflow invocation reports. The following screenshots demonstrate the idea more and the workflow report featured describes the feature in more detail.

### Example Report for a Run:

Screen Shot 2019-05-23 at 7 30 27 PM

### Report Template in Workflow Editor

Screen Shot 2019-05-23 at 7 30 53 PM

### Report Template in Workflow YAML

Screen Shot 2019-05-23 at 7 30 33 PM

### Demonstrating all the components available

Screen Shot 2019-05-23 at 7 31 15 PM

Screen Shot 2019-05-23 at 7 31 22 PM

Screen Shot 2019-05-23 at 7 31 28 PM

Screen Shot 2019-05-23 at 7 31 45 PM

```yaml
class: GalaxyWorkflow
name: My Cool Workflow
inputs:
  input_1: data
  image_input: data
  input_list: collection
outputs:
  output_1:
    outputSource: first_cat/out_file1
  output_image:
    outputSource: image_cat/out_file1
steps:
  first_cat:
    tool_id: cat
    in:
      input1: input_1
  image_cat:
    tool_id: cat
    in:
      input1: image_input
  qc_step:
    tool_id: qc_stdout
    state:
      quality: 9
    in:
      input: input_1
report:
  markdown: |
    ## About This Report

    This report is generated from markdown content in the workflow YAML/JSON.
    Workflow invocation reports provide a customizable way for workflow authors
    to summarize the results of a workflow execution. There is a default markdown
    template if a workflow does not define one (right now it just shows the inputs,
    outputs, and workflow - but this will evolve).

    This report is written in "Galaxy Workflow Flavored Markdown". This workflow
    variant of "Galaxy Flavored Markdown" contains workflow-relative references.
    Dataset, collections, and jobs are referenced by step labels, input labels, and output
    labels - rather than by object IDs. The Galaxy invocation report generator plugin
    translates this to "Galaxy Flavored Markdown" at the time the report is viewed and
    that format references actual object IDs (by database ID internally, and encoded ID
    when exported to the client via the API).

    An upshot of translating the workflow markdown to this second neutral format that
    has no concept of the workflow invocation is that client side rendering (and much
    of the backend processing) is completely general and not tied to workflows or
    invocations. The same markdown components could potentially be used to render pages,
    history annotations, describe library folders, etc..

    The next two sections demonstrate the auto generated inputs and outputs sections
    in the default workflow invocation report template.

    ```
    ## Workflow Inputs
    ::: invocation_inputs
    :::

    ## Workflow Outputs
    ::: invocation_outputs
    :::
    ```

    ## Workflow Inputs
    ::: invocation_inputs
    :::

    ## Workflow Outputs
    ::: invocation_outputs
    :::

    The auto-generated sections could be hand-crafted from workflow markdown also,
    listing out each input and output explicitly. The auto generated sections are merely
    a convenience to avoid needing to that by hand and for supplying a default report
    for workflows that don't define a custom one report.

    ## More Custom Content

    The rest of this report demonstrates more the directives allowed in the report
    markdown. ``invocation_outputs`` and ``invocation_inputs`` are not allow in
    "Galaxy Flavored Markdown" (the non-workflow invocation specific format), but
    the remainder of these are allowed - they would just need to reference object
    IDs instead of inputs, outputs, and steps.

    Once can reference an output and embed a display of it as follows:

    ```
    ::: history_dataset_display output=output_1
    :::
    ```

    ::: history_dataset_display output=output_1
    :::

    Inputs can be referenced and displayed the same way:

    ```
    ::: history_dataset_display input=input_1
    :::
    ```

    ::: history_dataset_display input=input_1
    :::

    ---

    Images can be embedded directly into the report as follows:

    ```
    ::: history_dataset_as_image output=output_image
    :::
    ```

    ::: history_dataset_as_image output=output_image
    :::

    ---

    Dataset peek content can be displayed to quickly provided an embedded
    summary of an input or output:

    ```
    ::: history_dataset_peek output=output_1
    :::
    ```

    ::: history_dataset_peek output=output_1
    :::

    ---

    Dataset "info" content can be displayed as well:

    ```
    ::: history_dataset_info input=input_1
    :::
    ```

    ::: history_dataset_info input=input_1
    :::

    ---

    Collections can be displayed:

    ```
    ::: history_dataset_collection_display input=input_list
    :::
    ```

    ::: history_dataset_collection_display input=input_list
    :::

    ---

    The whole workflow can be embedded to provide some context and display
    annotations and steps.

    ```
    ::: workflow_display
    :::
    ```

    ::: workflow_display
    :::

    ---

    Job parameters can be summarized:

    ```
    ::: job_parameters step=qc_step
    :::
    ```

    ::: job_parameters step=qc_step
    :::

    ---

    Job metrics can be summarized as well:

    ```
    ::: job_metrics step=image_cat
    :::
    ```

    ::: job_metrics step=image_cat
    :::

    ---

    Tool standard out and error are also available for steps.

    ```
    ::: tool_stdout step=qc_step
    :::
    ```

    ::: tool_stdout step=qc_step
    :::

    ```
    ::: tool_stderr step=qc_step
    :::
    ```

    ::: tool_stderr step=qc_step
    :::

    ---

    *fin*
```

Simplified, more fully backed version of the idea from https://github.com/galaxyproject/galaxy/pull/7994

### Old-style embedded objects vs components

In theory, this could make really nice use of excellent VueJS components for dataset display, dataset collection display, workflow display, etc.. These aren't available yet, we don't even really have Backbone stuff that exist very well outside history panels, so this re-uses "components" from Galaxy Pages for datasets and reuses the collection display for history panels as a stand-alone display for collections (I got this trick from DIsplayStructured.vue). This isn't ideal and I know that but at least reusing things this way will provide a path forward for migrating both pages and this new Markdown language (which could easily replace or co-exist with pages HTML someday) together seamlessly as the real modern components become available and doesn't increase the overall work needed to integrate newer style components. In fact, this might even be the impetus for creating and polishing more of these components.

### markdown-it vs markdown-it-vue

I tried this with markdown-it-vue also but it added very little in terms of reducing client code, obscured entirely how to attach plugins to the markdown rendering process, and brought in many, many more extra packages that we don't need or want.

## Testing this

Given there is no representation of invocations in the GUI and not editor for reports config this is a bit challenging still. But here is goes:

- Source Galaxy's virtualenv.
- Login to a user and grab an API key, set in the following command:
- ``GALAXY_TEST_EXTERNAL=http://localhost:8080 GALAXY_TEST_USER_API_KEY=38175dfc69d9009992b69efcca8b209b pytest test/api/test_workflows.py::WorkflowsApiTestCase::test_workflow_invocation_report_custom``
- In the web browser go to http://localhost:8080/api/invocations and grab the latest invocation ID. Replace it in the follow query string.
- Navigate to http://localhost:8080/workflows/invocations/report?id=c887f1d0da42bdfe


bgruening:
John this is a "Geniestreich"! It works flawlessly for me. The only thing I needed to change is to make the png filetype visible during upload (https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/sample/datatypes_conf.xml.sample#L312) but this might be my setup. The test pass for me and are very educational. It was a little bit hard to test it on a real workflow that actually does something - but all I tested worked so far. 

I would like to propose that we get this into 19.09 as it is, with the kind of hidden user exposure. Freiburg will use the next release cycle to add reports to a few key training workflows and write proper documentation and an own training for invocation-reports. We announce it properly to all users in 20.01. What do you think?  

¡muchas gracias!

jmchilton:
I love the eagerness and I'm disappointed in myself that I lost momentum on this a bit against GCC and haven't been able to find time to pick it back up. @jxtx made the point the syntax would feel more comfortable if it mirrored RMarkdown (https://rmarkdown.rstudio.com/lesson-4.html). I think I tend to agree but I don't have a sense yet how hard it would be to replace.

So... I think I'm fine with merging behind a flag for 19.09, if you promise to merge a cleanup that I will attempt to write essentially completely replaces the embedded code syntax for 20.01. If we're going to replace the syntax I don't know the benefit of merging now... maybe so that we have something in there in case the syntax re-write never happens.

I'll rebase #7913 today but that should be merged very quickly here if we really want this PR to be in 19.09. I think I had convinced @mvdbeek that we shouldn't just re-use sniffers as-is for validation...

mvdbeek:
> I'll rebase #7913 today but that should be merged very quickly here if we really want this PR to be in 19.09. I think I had convinced @mvdbeek that we shouldn't just re-use sniffers as-is for validation...

Indeed, plus we discussed that writing validators might be nice beginner and/or hack projects.
Additionally we can still decide on a per datatype basis that the sniffer is good enough and just re-use that. So I'm fine with #7913.

bgruening:
@jmchilton I also considered briefly to use `rst` for people that want to use it. I see the argument that you have more level of freedom in `rst` which might be beneficial in a report. So what about adding a tag, where the user can choose between different flavors of Markdown and `rst`?

jmchilton:
@bgruening I don't think rst is worth the effort. We should focus on one language and really push it hard and make it as easy and robust and rich as possible. I'm not -1 on the idea if you want to push it - but it isn't how I'd allocate resources.

bgruening:
I guess my point was to annotate this properly from the beginning so we can enhance/change the syntax whenever it is time.

> So... I think I'm fine with merging behind a flag for 19.09, if you promise to merge a cleanup that I will attempt to write essentially completely replaces the embedded code syntax for 20.01.

I promise :)

jmchilton:
> I guess my point was to annotate this properly from the beginning so we can enhance/change the syntax whenever it is time.

The workflow syntax does key the report on ``markdown:`` so I think we are safe at that user-facing level - and that is the stored pretty much directly as is as JSON in the database so we could easily support ``rst`` instead of ``markdown`` I think.

jmchilton:
Rebased on the validation PR. I think this is solid and ready to go but I did introduce an alternative syntax in https://github.com/galaxyproject/galaxy/pull/8511. Something easy to bike shed about this finally 🤣.

martenson:
I believe superseded by https://github.com/galaxyproject/galaxy/pull/8543

jennaj:
Could that view include some kind of "export to PDF" button? So people could have a visual record, put in a paper, etc?

If already included, skip this comment. If needs a different ticket (almost any view/page in Galaxy could use a "export to PDF" button) could make a new ticket.

So -- if that on the radar anywhere yet? See comment here: https://github.com/galaxyproject/galaxy/issues/5389#issuecomment-530547182

PR #8034, opened 2019-05-24
es6 upgrades in entry folder

WilliamHolden:


martenson (approved):
Thank you @WilliamHolden !

PR #8035, opened 2019-05-24
enhance TS repo search capabilities

martenson:
* adds `repo_lineage` to the index and results - this should allow us to quickly identify what repo needs updating and bring utility to our tool installation interface. 
* Includes changes to API to return the new attribute. 
* Also lowers the search limit for repos and tools to 1 character.
* include categories as keywords for repository index
* allow filtering on category/owner using github-like wildards "category:name"
* only log metadtaa like matched terms, do not return in http response
* refactoring

cc @guerler


guerler:
Looks good, thanks for working on this!

PR #8042, opened 2019-05-27
Extend ObjectStore integration tests

VJalili:
This PR does the following: 

1. moves all ObjectStore integration tests, and their related files, to a dedicated package;
2. adds a test asserting the content of the files that ObjectStore creates as a result of running the test tool, i.e., `create_10`. Ideally, this test should be implemented as a separate test method; however, with the current setup of the integration test, test methods within a test case are not executed independently. For instance, with `n` test method in a test case each producing `10` datasets running `create_10` tool, ObjectStore will have `10n` datasets in the output because all the test method use one instance of Galaxy with a common ObjectStore setup. It may worth it to work on making sure there is one Galaxy instance with independent setup per test method of a test class that can be initialized and cleaned-up using `setUp` and `tearDown` methods.


VJalili:
@jmchilton could you please have a look at this?

test/integration/objectstore/test_jobs.py
```diff
@@ -71,6 +71,25 @@ def test_tool_simple_constructs(self):
...
+        According to the ObjectStore configuration given in the
+        `DISTRIBUTED_OBJECT_STORE_CONFIG_TEMPLATE` variable, datasets
+        can be stored on three backends, named:
+            -   primary/files1;
+            -   primary/files2;
+            -   secondary/files3.
+
+        Objectstore _randomly_ distributes tools outputs on
+        `primary/files1` and `primary/files2`, and will use
+        `secondary/files3` and both `primary` backends fail.
```
mvdbeek:
```suggestion
        `secondary/files3` if both `primary` backends fail.
```

test/integration/objectstore/test_jobs.py
```diff
@@ -79,17 +98,39 @@ def test_tool_simple_constructs(self):
...
         # and not actually testing object store behavior.
         assert (files_1_count <= 11) and (files_2_count <= 11)
         assert (files_1_count >= 0) and (files_2_count >= 0)
 
+        # TODO: ideally the following assertion should be separated in a different test method.
+        contents = []
+        path1_files = _get_datasets_files_in_path(self.files1_path)
+        path2_files = _get_datasets_files_in_path(self.files2_path)
+        path3_files = _get_datasets_files_in_path(self.files3_path)
+        for filename in set().union(path1_files, path2_files, path3_files):
```
mvdbeek:
```suggestion
        for filename in path1_files + path2_files + path3_files:
```

test/integration/objectstore/test_jobs.py
```diff
@@ -79,17 +98,39 @@ def test_tool_simple_constructs(self):
...
+            with open(filename) as f:
+                content = f.read().strip('\n').strip('\t')
```
mvdbeek:
```suggestion
                content = f.read().strip()
```
I don't see where the tab should be coming from.

VJalili: no tab, but better to strip the line break to simplify content assertions.

mvdbeek: line break (and tabs for that matter) are considered whitespace, so they are removed by `strip()`:
```
In [1]: s = ''

In [2]: ?s.strip
Docstring:
S.strip([chars]) -> str

Return a copy of the string S with leading and trailing
whitespace removed.
If chars is given and not None, remove characters in chars instead.
Type:      builtin_function_or_method

In [3]: s = '\n\t'

In [4]: s.strip()
Out[4]: ''
```

VJalili: Good point, thanks.

mvdbeek:
Thanks @VJalili !

VJalili:
Thanks @mvdbeek

PR #8044, opened 2019-05-27
Update JobMetrics component to use a Vuex Store + vuex-cache...

jmchilton:
... to cache API requests.

I'm trying to rewrite my Markdown generation of workflow reports to not pre-build data needed as part of a mammoth request. I think using a vuex store to allow various components to reuse the same data is the right approach, but I will need a way to ensure various components don't cause the same data to be fetched repeatedly from different contexts. vuex-cache seems to be a clean enough solution to this.

This may be overkill for this single component - but I'm intending to build several components over other parts of the job API and there I think having an approach agreed on in this context will be very helpful.

### DIFF
--- a/client/galaxy/scripts/components/JobMetrics/JobMetrics.vue
+++ b/client/galaxy/scripts/components/JobMetrics/JobMetrics.vue
@@ -16,8 +16,8 @@
 </template>
 
 <script>
-import { getAppRoot } from "onload/loadConfig";
-import axios from "axios";
+import { mapCacheActions } from "vuex-cache";
+import { mapGetters } from "vuex";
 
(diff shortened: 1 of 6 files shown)


martenson:
thanks @jmchilton !

dannon:
@jmchilton Sorry I missed this when I was out last week.  Yes, using a central vuex store like this with caching for anything accessed from multiple places in the application is exactly what we want to do.

jmchilton:
Thanks for the merge @martenson - thanks for the review @dannon - glad this is on the right track. I can see this pattern being really helpful and it was good to learn vuex by getting my hands dirty.

PR #8051, opened 2019-05-28
Prepare all packages/ for actual releases.

jmchilton:
Pre-releases based on this PR are already available:
- https://pypi.org/project/galaxy-util/
- https://pypi.org/project/galaxy-job-metrics/
- https://pypi.org/project/galaxy-objectstore/
- https://pypi.org/project/galaxy-tool-util/
- https://pypi.org/project/galaxy-data/

Changes:

- Update setup.py to reflect Galaxy no longer supports Python 3.4.
- Move PROJECT_DESCRIPTION into a variable imported into setup.py dynamically, reducing the differences in setup.py across packages.
- Fix some project descriptions (copy-paste problems).
- Fix versions and HISTORY.rst for a release this cycle, didn't get any 19.5.0 packages out.
- Add development requirements for publishing, linting build artifacts
- Add Makefile(s) for building projects (all synchronized via symbolic links).
- Extend Makefile template from other projects to support development releases (a bit anyway).
- Fix bug job metrics package listing.
- Build universal wheels (for Python 2/3 compat.)
- Update versions to reflect pre-releases already made.

### DIFF
--- a/.ci/flake8_blacklist.txt
+++ b/.ci/flake8_blacklist.txt
@@ -2,6 +2,10 @@
 .tox
 .venv
 .venv3
+packages/*/.venv
+packages/*/build
+packages/*/dist
+packages/venv
 node_modules
 database
 doc/build
--- a/packages/build_scripts/commit_version.py
+++ b/packages/build_scripts/commit_version.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python
+# Modify version...
+import datetime
+import os
+import re
+import subprocess
+import sys
+
+
+PROJECT_DIRECTORY = os.path.join(os.path.dirname(__file__), "..")
+PROJECT_DIRECTORY_NAME = os.path.basename(os.path.abspath(PROJECT_DIRECTORY))
+PROJECT_MODULE_FILENAME = "galaxy_%s.py" % PROJECT_DIRECTORY_NAME
+PROJECT_NAME = PROJECT_DIRECTORY_NAME.replace("_", "-")
+
+
+def main(argv):
+    source_dir = argv[1]
+    version = argv[2]
+    history_path = os.path.join(PROJECT_DIRECTORY, "HISTORY.rst")
+    history = open(history_path, "r").read()
+    today = datetime.datetime.today()
+    today_str = today.strftime('%Y-%m-%d')
+    history = history.replace(".dev0", " (%s)" % today_str)
+    open(history_path, "w").write(history)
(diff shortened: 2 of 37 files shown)


mvdbeek (approved):
nice!

mvdbeek:
Do you have any plans on managing dependencies for these packages ?

jmchilton:
> Do you have any plans on managing dependencies for these packages ?

It is a good question, I've thought a lot about it and I don't know what I think will work best. There are a lot of things to balance here.

Dependencies on galaxy-lib have been kind of a mess since it doesn't really have real versioning - just random snapshots of development releases in some ways.  Frequently a galaxy-lib "minor" release will break Planemo - but Planemo doesn't pin an upper bound on galaxy-lib either - so part of the problem is Planemo. 

I think within the repo, the versions should probably be pinned something like ``galaxy-objectstore`` depends on ``galaxy-util>=19.9,<19.12`` where we would increase that lower pin if there are obvious incompatibilities that arise during a release cycle. I think for 19.09 - I'll probably do some real releases before Galaxy freezes - but for future releases if we didn't do "real" releases of these things until Galaxy freezes - we'd have a real shot at library interface level semantic versioning. Real versions would be nice - but under this plan Planemo, Pulsar, Ephemeris would probably be perpetually targeting development releases so what would be the point I suppose?

It is hard to imagine more frequent releases of stable things being derived from something unstable. Perhaps we could do a hybrid approach - instead of ``galaxy-util>=19.9,<19.12`` we could do ``galaxy-util>=19.9,<19.10`` and if we know we're breaking something in a semvar perspective in development of Galaxy - we could bump the versions of these libraries accordingly and do releases in-between the Galaxy releases (19.10, 19.11, 19.12, etc..).

All of this assumes dependent projects on these libraries will always depend on a single major release of Galaxy's version of the dependencies (e.g. 19.9.X). This may not be an assumption we want to impose - presumably galaxy-objectstore 19.9 will work with galaxy-util 20.1 - why prohibit project X from depending on these both concurrently. So we could instead have ``galaxy-objectstore`` depend on ``galaxy-util>=19.9`` and just do another  minor release of all dependent projects if we break the interface of ``galaxy-util`` during development that sets a maximum pin for the dependency.

martenson:
> So we could instead have galaxy-objectstore depend on galaxy-util>=19.9 and just do another minor release of all dependent projects if we break the interface of galaxy-util during development that sets a maximum pin for the dependency.

this makes a lot of sense to me


Thanks for working on this John!

PR #8058, opened 2019-05-29
Refactor jstree.py out of galaxy.util to simplify library dependencies.

jmchilton:
This utility is only used in one file - remote_files.py - so just place it beside that file to keep galaxy-util dependencies simpler.

Needed for #7971.

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -515,7 +515,7 @@ do
       -u|-unit|--unit)
           report_file="run_unit_tests.html"
           test_script="pytest"
-          unit_extra='--doctest-modules --ignore lib/galaxy/web/proxy/js/node_modules/ --ignore lib/galaxy/webapps/tool_shed/controllers --ignore lib/galaxy/jobs/runners/chronos.py --ignore lib/galaxy/webapps/tool_shed/model/migrate --ignore lib/galaxy/util/jstree.py'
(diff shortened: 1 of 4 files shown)


run_tests.sh
```diff
@@ -515,7 +515,7 @@ do
       -u|-unit|--unit)
           report_file="run_unit_tests.html"
           test_script="pytest"
-          unit_extra='--doctest-modules --ignore lib/galaxy/web/proxy/js/node_modules/ --ignore lib/galaxy/webapps/tool_shed/controllers --ignore lib/galaxy/jobs/runners/chronos.py --ignore lib/galaxy/webapps/tool_shed/model/migrate --ignore lib/galaxy/util/jstree.py'
+          unit_extra='--doctest-modules --ignore lib/galaxy/web/proxy/js/node_modules/ --ignore lib/galaxy/webapps/tool_shed/controllers --ignore lib/galaxy/jobs/runners/chronos.py --ignore lib/galaxy/webapps/tool_shed/model/migrate'
```
martenson, line 518: this means we have to lint it, forcing us to modify something we took verbatim

martenson:
I am fine with moving this, but could we move it somewhere other than `api/`?  That module has special treatment and jstree does not fit there imho.

run_tests.sh
jmchilton: We've already modified it though - I feel like we own it at this point.

martenson: fair enough

jmchilton:
If we don't want here - might as well keep it as a utility. I need to work on optional dependencies for ``galaxy.util.template`` anyway.

martenson:
We could maybe release it under our pypi account and use as a proper requirement? Dunno what is best, but I am offering help.

PR #8065, opened 2019-05-30
[19.05] Fix metadata output collection on Pulsar with new metadata strategy.

jmchilton:
Output fix corresponding to input fix in https://github.com/galaxyproject/galaxy/pull/7694/files. I think I had been testing with metadata retry on and didn't realize the outputs weren't returning to the correct directory.

### DIFF
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1231,6 +1231,7 @@ class JobWrapper(HasResourceParameters):
         job_stderr=None,
         check_output_detected_state=None,
         remote_metadata_directory=None,
+        job_metrics_directory=None,
     ):
         """
         Called to indicate that the associated command has been run. Updates
@@ -1470,7 +1471,7 @@ class JobWrapper(HasResourceParameters):
         job.set_final_state(final_job_state)
         if not job.tasks:
             # If job was composed of tasks, don't attempt to recollect statisitcs
-            self._collect_metrics(job)
+            self._collect_metrics(job, job_metrics_directory)
         self.sa_session.flush()
         log.debug('job %d ended (finish() executed in %s)' % (self.job_id, finish_timer))
         if job.state == job.states.ERROR:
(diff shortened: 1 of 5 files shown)


bgruening:
I get the following in my rather complex local setup.

```
  File "lib/galaxy/jobs/runners/__init__.py", line 130, in run_next
    method(arg)
  File "lib/galaxy/jobs/runners/pulsar.py", line 517, in finish_job
    self._handle_metadata_externally(job_wrapper, resolve_requirements=True)
  File "lib/galaxy/jobs/runners/__init__.py", line 346, in _handle_metadata_externally
    kwds={'overwrite' : False})
  File "lib/galaxy/jobs/__init__.py", line 1964, in setup_external_metadata
    **kwds)
  File "lib/galaxy/metadata/__init__.py", line 107, in setup_external_metadata
    os.mkdir(metadata_dir)
OSError: [Errno 17] File exists: '/home/bag/projects/code/galaxy/database/jobs_directory/000/95/metadata'
```

gmauro:
With this patch and cgroups enabled both sides (glx and pulsar), I got this error:
```
galaxy.web.framework.decorators ERROR 2019-05-31 14:40:36,297 [p:21824,w:1,m:0] [uWSGIWorker1Core1] Uncaught exception in exposed API method:
  File "/srv/galaxy/server/lib/galaxy/web/framework/decorators.py", line 282, in decorator
  File "/srv/galaxy/server/lib/galaxy/webapps/galaxy/api/job_files.py", line 103, in create
    shutil.move(input_file.name, path)
  File "/usr/lib/python2.7/shutil.py", line 316, in move
    copy2(src, real_dst)
  File "/usr/lib/python2.7/shutil.py", line 144, in copy2
    copyfile(src, dst)
  File "/usr/lib/python2.7/shutil.py", line 97, in copyfile
    with open(dst, 'wb') as fdst:
IOError: [Errno 2] No such file or directory: u'/srv/galaxy/jobs/000/334/metadata/__instrument_cgroup__metrics'
```

jmchilton:
Yes - so this fixes metadata for the default metadata_strategy but it does break metrics. I think @natefoo and I know this would prefer to make the fix anyway to get main working. Is this a regression on your end though? Does the job work fine and set metadata without this patch?

gmauro:
Yes, without this patch the job works fine.

bgruening:
> Does the job work fine and set metadata without this patch?

@gmauro we did not get metadata setting to work on pulsar yet, right?

jmchilton:
Oh I see, this would break jobs with metrics and without remote metadata... crap.

bgruening:
We are trying to get remote metadata working, so don't bother with this just because of us.

jmchilton:
Are you experiencing scaling problems that make remote metadata appealing?

I think the fact this patch breaks the default Pulsar configuration (no remote metadata) is a real problem.

bgruening:
> Are you experiencing scaling problems that make remote metadata appealing?

No, not yet. But we thought this is the recommended way.

> I think the fact this patch breaks the default Pulsar configuration (no remote metadata) is a real problem.

Yes, it would mean changing the default as well, which I think is ok, if it is the recommended way for production Galaxy instances.

jmchilton:
> No, not yet. But we thought this is the recommended way.

It isn't documented anywhere and I don't think it is worth the complexity in terms of configuration. It is a last resort IMO until we get it right in 19.09. I'd rather see Pulsar running at scale and have actual profile data on this being a problem - and then decide if just fixing the problematic datatypes are better or switching the server to remote metadata.

bgruening:
Ok, we can work around this on our side. Whatever you do, we can patch it on our side to make it work. Thanks John.

jmchilton:
Latest stuff should fix job metrics regression from this branch?

natefoo:
Appears to work for me!

PR #8067, opened 2019-05-31
Add BAM test files reference hg38 patch 12 

aloraine:
Focused on region chr17:41,713,943-41,744,899 encompassing MEOX1, alternatively spliced gene encoding a homeobox and highly expressed in heart.


bgruening:
Hi @aloraine! This seems to be an interesting region! What do you want to test with that BAM file?

mvdbeek:
I suppose this was a mistargeted PR ? Feel free to re-open with some explanation of what the PR does or shows if not.

PR #8077, opened 2019-06-01
Call tool_script.sh with shell

mvdbeek:
The way we currently invoke the tool_script.sh script requires that we can set the executable bit on the script. That's not the case in our environment (i've worked around this with a terrible hack so far), and I don't think there's any harm in including the shell explicitly. The shell can be configured via the destination and the container resolver, so I think this should be fine in all situations.


jmchilton:
Thanks for the fix!

PR #8078, opened 2019-06-03
Delegate config file watching to single handler

mvdbeek:
This should reduce IO especially on instances that use a lot of
handlers and/or where the config files are on a remote file system.

The process that watches the config files will now
be the active process that alphabetically compares highest.
If that process goes away or is superseded by a process that
compares higher now that process will take over.
If a change has been found a control task will be sent via
the kombu messaging system.

~~Overall this works nicely, but while stopping the current
config watcher there is a small break during which changes
to the config don't trigger a reload, this could be handled
more gracefully.~~
When stopping the current config watcher the shutdown
method in the DatabaseHeartbeat class will remove itself
from the `WorkerProcess`es and trigger a reconfiguration
of the watcher processes.

This also includes some cleanup in the actual watchers,
and uses the more generally useful `get_watcher` and
`EventHandler` instead of custom factories and Event
Handler classes


jmchilton:
Thanks a lot @mvdbeek !

PR #8080, opened 2019-06-03
Filter all pykube queries by namespace

### LINKED ISSUE #8079: k8s runner not filtering by namespace
The k8s runner does not always respect the runner parameter "k8s_namespace", and throws an exception from the `__job_failed_due_to_low_memory` method when a namespace other than "default" is used.

nuwang:
Fixes: https://github.com/galaxyproject/galaxy/issues/8079


pcm32:
LGTM, does it work well?

nuwang:
Yes, works as expected.

PR #8087, opened 2019-06-04
Move pykube from conditional requirements to pinned requirements.

VJalili:
Since pykube is necessary for some tests now (e.g., https://github.com/galaxyproject/galaxy/pull/7835), and might be required for more tests in the future, hence maybe it might be easier to install it by default.

ping @jmchilton @nsoranzo

### DIFF
--- a/lib/galaxy/dependencies/__init__.py
+++ b/lib/galaxy/dependencies/__init__.py
@@ -143,9 +143,6 @@ class ConditionalDependencies(object):
     def check_pbs_python(self):
         return "galaxy.jobs.runners.pbs:PBSJobRunner" in self.job_runners
 
-    def check_pykube(self):
-        return "galaxy.jobs.runners.kubernetes:KubernetesJobRunner" in self.job_runners
-
     def check_chronos_python(self):
         return "galaxy.jobs.runners.chronos:ChronosJobRunner" in self.job_runners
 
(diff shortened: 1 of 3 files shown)


nsoranzo:
This seems premature, are there tests not working without this?

Also, it looks like that [pykube is abandoned](https://github.com/kelproject/pykube/issues/166), time to move to something else? A quick Google search suggests [kubernetes](https://pypi.org/project/kubernetes/) (developed at https://github.com/kubernetes-client/python ).

VJalili:
> This seems premature, are there tests not working without this?

Yes; https://github.com/galaxyproject/galaxy/pull/7835

> Also, it looks like that pykube is abandoned, time to move to something else? A quick Google search suggests kubernetes (developed at https://github.com/kubernetes-client/python ).

Indeed, it would be ideal to replace this abandoned library, however, currently the aforementioned test,  Pulsar, and some k8s job runners leverage this library.

nsoranzo:
> > This seems premature, are there tests not working without this?

> Yes; https://github.com/galaxyproject/galaxy/pull/7835

I think that for integration tests, the existing mechanism of conditional dependencies should install pykube when needed, but is not used on purpose? Ping @jmchilton

VJalili:
It does not install, because the condition is not checking for the test; hence, I need to install it manually.

mvdbeek:
You're right in that you'll have to install it manually at the moment. I don't think the solution is adding dependencies to Galaxy. I'd suggest adding an additional tox environment for integration tests.

Running through the conditional dependencies isn't really an option for integration tests, they can be different for each test, and they need to be installed before collecting the tests.

VJalili:
Sounds good.

PR #8089, opened 2019-06-04
Added support for runAsUser and runAsGroup to k8srunner

nuwang:
This PR adds support for specifying the docker user and group when running the job pod (defaulting to the same uid and group as the job handler), so that file ownership propagates as expected. Without it, the job pods may run as root, depending on who the default container user is, and incorrectly sets file ownership on the shared filesystem, which makes it problematic for Galaxy to delete files without setgid, umask and other trickery.

### DIFF
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -54,6 +54,8 @@ class KubernetesJobRunner(AsynchronousJobRunner):
             k8s_job_api_version=dict(map=str, default="batch/v1"),
             k8s_supplemental_group_id=dict(map=str),
             k8s_pull_policy=dict(map=str, default="Default"),
+            k8s_run_as_user_id=dict(map=int, default=os.getuid()),
+            k8s_run_as_group_id=dict(map=int, default=app.config.gid),
             k8s_fs_group_id=dict(map=int),
             k8s_default_requests_cpu=dict(map=str, default=None),
             k8s_default_requests_memory=dict(map=str, default=None),
@@ -238,16 +240,19 @@ class KubernetesJobRunner(AsynchronousJobRunner):
         }
         # TODO include other relevant elements that people might want to use from
         # TODO http://kubernetes.io/docs/api-reference/v1/definitions/#_v1_podspec
-
-        if self._supplemental_group and self._supplemental_group > 0:
-            k8s_spec_template["spec"]["securityContext"] = dict(supplementalGroups=[self._supplemental_group])
-        if self._fs_group and self._fs_group > 0:
-            if "securityContext" in k8s_spec_template["spec"]:
-                k8s_spec_template["spec"]["securityContext"]["fsGroup"] = self._fs_group
-            else:
-                k8s_spec_template["spec"]["securityContext"] = dict(fsGroup=self._fs_group)
-
+        k8s_spec_template["spec"]["securityContext"] = self.__get_k8s_security_context()
         return k8s_spec_template
+
+    def __get_k8s_security_context(self):
+        security_context = {
+            "runAsUser": int(self.runner_params["k8s_run_as_user_id"]),
+            "runAsGroup": int(self.runner_params["k8s_run_as_group_id"])
+        }
+        if self._supplemental_group and self._supplemental_group > 0:
+            security_context["supplementalGroups"] = [self._supplemental_group]
+        if self._fs_group and self._fs_group > 0:
+            security_context["fsGroup"] = self._fs_group
+        return security_context
 
     def __get_k8s_restart_policy(self, job_wrapper):
         """The default Kubernetes restart policy for Jobs"""


lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -54,6 +54,8 @@ def __init__(self, app, nworkers, **kwargs):
             k8s_job_api_version=dict(map=str, default="batch/v1"),
             k8s_supplemental_group_id=dict(map=str),
             k8s_pull_policy=dict(map=str, default="Default"),
+            k8s_run_as_user_id=dict(map=int, default=os.getuid()),
```
nuwang, line 57: @natefoo Is this ok in terms of caching? The assumption is that the job runner is instantiated only once and therefore, this will get cached? The alternative could be to add an app.config.uid property.

natefoo: You're correct about it instantiating once, so this method looks fine to me.

natefoo:
Per discussion on the dev gitter channel yesterday, we were wondering how people using the k8s runner are currently handling cleaning up container outputs without this. @pcm32?

natefoo (approved):
This looks good to me, hopefully it does not cause problems for others' setups or we may need to make the runAs* specs optional.

pcm32:
We have never had issues with this on different kubernetes setup (AWS, GCP, OpenStack-Kubespray, OpenStack-Kubenow, Rancher, etc) and with different shared file systems provisioned (GlusterFS, NFS, CephFS, local minikube, FSx on AWS, etc). Do consider that each file system has certain ways of dealing with permissions differently. I can confirm as well that we never leave files behind (as I always need to turn on the feature to keep working directories for debugging and most of our long term running instances would run out of disk very quickly otherwise). Whether that is an artifact of the umask done on docker-galaxy-stable (@nuwang is using a container that I haven't tested) or other intermediate setting, I cannot confirm it because I have never dig on this as it hasn't been an issue on any of our different setups. 

I would prefer if this is left as optional and configurable, as older versions of Kubernetes didn't always support runAs* and sometimes you had to run the cluster with different config flags to expose this functionality (runAs* support). Not sure at what version of kubernetes this stopped being the case, but I think that with little effort we could avoid any potential backward incompatibility. I for once have some long term running k8s clusters on openstack that sometimes run on older versions.

Thanks!

nuwang:
@pcm32 The reason for doing this came out of a discussion with Nate on the gitter dev channel. Specifically, the issue is that ownership of the files generated by the tools will belong to whatever user the container runs as (could be root) and if root squash is used, whatever it's mapped to. Thus, the files generated by the tools are not owned by Galaxy, and Galaxy can only manipulate them through group ownership. Nate was of the opinion that this should not be the case, and Galaxy should own these files.

We followed the docs that you gave regarding your config: https://github.com/crs4/container-galaxy-elixir-is-varcall/blob/master/docs/making_galaxy_and_tool_containers_share_data.md and it appears to confirm that it's done through group ownership, not user ownership. 

Can you let pls double check your setup and let us know which user owns the files generated by Galaxy job pods?

With regard to backward compatibility, runAsUser appears to have been introduced in Kube 1.2, since 2016. https://github.com/kubernetes/kubernetes/issues/22179
Do you need to suppose users running versions prior to that? If not, I think it would be preferable to specify the correct user so that ownership rights to the files can be configured easily, without resorting to setgid and umask, both from a consistency perspective (files owned by root on ebs  volumes and nobody in root-squashed nfs) and an ease-of-use perspective.

nuwang:
Just wanted to add that it's not a problem to change the defaults to be backward compatible if your prior configurations are unlikely to run with this new default. The idea was to make it easier to stick to how Galaxy owns files in general. Just let me know.

pcm32:
@nuwang I remember running across things like this https://stackoverflow.com/questions/50366494/kubenates-runasuser-is-forbidden at some point when implementing the fs_group and supplementary_groups. Some clusters might be configured not to allow it, and hence forcing it might mean that certain users who don't control the k8s deployment might be left out. The above is for a 1.10 cluster and I have some clusters running on 1.9, so I would appreciate being able to turn it on/off in case it produces issues.

I don't doubt that the user being used might be messy, but as I say this works correctly on a number of setups (whether this is due to groups being used in a favourable way, umasking or other, would need to be investigated). I'll try to send jobs and be looking at the file system on one of the deployments to check which user is being used - but I don't think that this exercise's result changes the fact that this should be left as optional.

nuwang:
Thanks, yes, that'd be good to know. In the meantime, I'll change this to be optional, instead of using defaults.

pcm32:
To clarify, I think that is great that we can have this additional control (runAs*), I don't oppose that in anyway, I hope that my concerns for the optionality didn't sound too negative.

nuwang:
Not at all, given that runAs may not be available in all cases, it makes complete sense to make this optional.

pcm32:
When I look on a job working directory when forcing it not to delete files I see:

```
# ls -l ../46
total 52
-rw-r--r-- 1 root   root      0 Jun  4 20:19 galaxy_46.e
-rw-r--r-- 1 root   root      2 Jun  4 20:19 galaxy_46.ec
-rw-r--r-- 1 root   root      0 Jun  4 20:19 galaxy_46.o
-rwxr-xr-x 1 galaxy galaxy 5906 Jun  4 20:19 galaxy_46.sh
drwxr-xr-x 2 root   root      6 Jun  4 20:19 home
-rw-r--r-- 1 root   root     11 Jun  4 20:19 __instrument_core_epoch_end
-rw-r--r-- 1 root   root     11 Jun  4 20:19 __instrument_core_epoch_start
-rw-r--r-- 1 root   root      1 Jun  4 20:19 __instrument_core_galaxy_memory_mb
-rw-r--r-- 1 root   root      2 Jun  4 20:19 __instrument_core_galaxy_slots
-rw-r--r-- 1 root   root   7792 Jun  4 20:19 __instrument_cpuinfo_cpuinfo
-rw-r--r-- 1 root   root   1279 Jun  4 20:19 __instrument_meminfo_meminfo
-rw-r--r-- 1 root   root    113 Jun  4 20:19 __instrument_uname_uname
-rw-r--r-- 1 root   root      0 Jun  4 20:19 memory_statement.log
drwxr-xr-x 2 root   root      6 Jun  4 20:19 tmp
-rwxr-xr-x 1 galaxy galaxy  418 Jun  4 20:19 tool_script.sh
-rw-r--r-- 1 root   root      0 Jun  4 20:19 tool_stderr
-rw-r--r-- 1 root   root    256 Jun  4 20:19 tool_stdout
drwxr-xr-x 2 root   root     61 Jun  4 20:19 working
```

sending a new job when asking it to delete the working files, I see, temporarily:

```
# ls -ltr
total 8
-rw-r--r-- 1 root root   0 Jun  4 20:44 tool_stderr
drwxr-xr-x 2 root root  61 Jun  4 20:44 working
-rw-r--r-- 1 root root 256 Jun  4 20:44 tool_stdout
-rw-r--r-- 1 root root  11 Jun  4 20:44 __instrument_core_epoch_end
```

So as you suspected nearly everything is root, how things are being deleted (as you can see the second case has less files), it escapes any quick explanation I could provide (maybe the container is being run as root? although I think that supervisor runs the process as the user...). But certainly better not to get stuff written as root, and possibly on some setups this could lead to have lagging files, so great that we have this now.

On my logs, the closest I see to the error mentioned is:

```
galaxy.util WARNING 2019-06-04 20:44:16,291 [p:425,w:2,m:0] [KubernetesRunner.work_thread-1] Unable to honor primary group (grp.struct_group(gr_name='galaxy', gr_passwd='x', gr_gid=1450, gr_mem=[])) for /export/galaxy-central/database/files/000/dataset_78.dat, group remains grp.struct_group(gr_name='root', gr_passwd='x', gr_gid=0, gr_mem=[]), error was: [Errno 1] Operation not permitted: '/export/galaxy-central/database/files/000/dataset_78.dat'
```

but I don't seem to find there the same error that @nuwang reports.

nuwang:
Thanks for the details. Yes, that would be the most plausible explanation - I just tried logging into that container and the default container user is root. Couldn't quite determine how the supervisor conf is overridden, but the file permissions suggest that the galaxy job handlers are being run as root. The reason we are seeing a different error is because in the minimal container, all processes run as user "galaxy", and being non-root, that user is unable to access the files generated by the k8s runner.

I'll close this PR and reissue it against 19.05, because the k8s runner's default ownership rights are undesirable and it's better to provide a way to fix this (will leave it backward compatible).

PR #8090, opened 2019-06-04
Convert repository client to vue

guerler:
Screen Shot 2019-06-08 at 3 04 40 PM

- [x] Determine which attributes to display in the revisions section (see comments below, example of available data shown at https://gist.github.com/guerler/8764b3ee917f17c34989b153cfa1fcbf)
- [x] Determine which attributes to display in the installations modal
- [x] Fix naive installation state listener
- [x] Combine status and button state columns
- [x] Collapse tool names in repository details view


lib/galaxy/webapps/galaxy/api/repositories.py
```diff
@@ -0,0 +1,260 @@
+import json
```
mvdbeek: There is a lot of overlap with `lib/galaxy/webapps/galaxy/api/tool_shed_repositories.py`, was this supposed to be a rename ? All the new methods here seem to be available in that file as well.

guerler: Good point. tool_shed_repositories.py and toolshed.py (and repositories.py) have several overlapping functions but right now tools are installed by making at least 2 individual calls see: https://github.com/galaxyproject/galaxy/blob/dev/client/galaxy/scripts/mvc/toolshed/repository-view.js#L303. prepareInstall and doInstall. One requesting from an api endpoint and the latter from the admin toolshed controller. There are ~5 methods in tool_shed_repositories.py which seem to install repositories. As of right now the new repositories api is an attempt to collect and refactor required features while reducing the dependency on toolshed modules. Overall the goal is to deprecate/consolidate tool_shed_repositories.py and tool_shed.py endpoints.

mvdbeek: That isn't necessary for installations though; I don't know why the UI does what it does.
You can certainly install repositories with a single POST, that's how bioblend and by extension ephemeris install repos. I'm not saying the current API is particularly easy to use, but it's widely used and functional.

Happy to see where you're taking it with this, once it's ready we can probably merge the methods.

guerler: Yes that sounds good!

jmchilton:
Very cool screenshot - I'm just going to bike shed it for a second and I realize that and I apologize ahead o time. Please remove workflows and datatypes "checkmark/X" from the GUI - we don't want to highlight or encourage those - they are deprecated tool shed "features" that don't work.

lib/galaxy/webapps/galaxy/api/repositories.py
guerler: I removed the repositories api, using existing endpoints for now regardless of their dependency on toolshed modules.

guerler: Thank you!

mvdbeek:
This is working very well. I think we should definitely remove the the legacy interface after merging this. I think we may also want to remove the beta interface, which is very clunky. The only feature I'd miss from the beta interface is the ability to install missing tools from workflows, though the implementation isn't practical if you have more than a handful of workflows anyway.

lib/galaxy/webapps/galaxy/api/toolshed.py
```diff
@@ -310,3 +310,19 @@ def search(self, trans, **kwd):
...
         response = json.loads(util.url_get(tool_shed_url, params=dict(q=q), pathspec=['api', 'repositories']))
         return response
+
+    @expose_api
+    @web.require_admin
+    def request(self, trans, **params):
+        """
+        GET /api/tool_shed/request
+        """
+        tool_shed_url = params.pop("tool_shed_url")
```
jmchilton, line 321: I would prefer it if you were keying on something and recovering the URL server side - shouldn't allow the client to make such arbitrary queries on behalf of the server... but I guess it is hidden behind a ``@require_admin`` at least.

guerler:
@jmchilton thanks for the comment. Can you recommend alternative attributes which we could show alongside revisions? Here is a list of available attributes: https://gist.github.com/guerler/8764b3ee917f17c34989b153cfa1fcbf

lib/galaxy/webapps/galaxy/api/toolshed.py
jmchilton: Actually if you can just verify this value is in the configured tool shed URLs (from tool_sheds_conf.xml - however that gets configured) that would be enough I think to greatly reduce the exploit surface of this.

guerler: I agree with you, this is a temporary endpoint and will either be replaced by an explicit one or the toolshed service will be setup for cors.

guerler: I followed up on your recommendation. The endpoint is now validating the toolshed url. Thanks.

martenson:
@guerler I would actually go the other way and also recommend dropping 'visible' and maybe even 'dependencies'. Nowadays everything has a dependency basically, and visible just gives you no actionable information.

Consider adding link to the toolshed page of the repo (maybe link to development url also/instead), and maybe the time since last repo update.

Having a 'copy to clipboard' on repo's changeset (which we can otherwise not display in the UI) would also be nice I think.

nsoranzo:
> Consider adding link to the toolshed page of the repo (maybe link to development url also/instead)

The TS page has a link to development URL, so I'd prefer the first or both.

guerler:
@mvdbeek suggested to display/list the tools contained in a repository. I think that could be helpful. There is already a link to the toolshed repository accessible by clicking on 'Show additional details and dependencies.'

martenson:
I suspect the list of available tools would complicate the UI substantially, since we'll need to design it for repos that have 1 tool or 40 tools.

mvdbeek:
> I suspect the list of available tools would complicate the UI substantially, since we'll need to design it for repos that have 1 tool or 40 tools.

A section / dropdown would be fine ? We do that in the legacy UI as well.

martenson:
@mvdbeek The legacy ui is something we want to improve on, the list of tools is displayed there on a page dedicated to a repository, as opposed here, where everything is a list of (inifinite scrollling) search results with just the important details, to keep the UI simple.

The too list is still one click away (on the TS repo detail page).

mvdbeek:
> The legacy ui is something we want to improve on

If you can't see the tools contained in a repo that would be a serious step back though.
The section shouldn't be expanded by default, but it really should be there.

martenson:
> The section shouldn't be expanded by default, but it really should be there.

So it is one click away, same as going to the TS repo detail page. Does it make that much of a difference?

guerler:
If the information on tools and versions contained in a repository is useful, it might be worth trying to display it in the ui. Other useful attributes might include the number of tools and the number of test cases. Thanks for all the comments!

guerler:
@galaxybot test

lib/galaxy/web/base/controller.py
```diff
@@ -314,7 +314,9 @@ def _get_extended_config(self, trans):
             'toolbox'                       : trans.app.toolbox.to_dict(trans, in_panel=False),
             'toolbox_in_panel'              : trans.app.toolbox.to_dict(trans),
             'message_box_visible'           : trans.app.config.message_box_visible,
-            'show_inactivity_warning'       : trans.app.config.user_activation_on and trans.user and not trans.user.active
+            'show_inactivity_warning'       : trans.app.config.user_activation_on and trans.user and not trans.user.active,
+            'tool_shed_urls'                : trans.app.tool_shed_registry.tool_sheds.values() if trans.app.tool_shed_registry else [],
```
mvdbeek:
```suggestion
            'tool_shed_urls'                : list(trans.app.tool_shed_registry.tool_sheds.values()) if trans.app.tool_shed_registry else [],
```

Otherwise this fails to serialize on python 3 because the map object can't be json encoded (https://github.com/galaxyproject/galaxy/blob/0fa287f76a675c49c2702681ac4404f050b77130/lib/galaxy/util/odict.py#L72). That'll go away fortunately with https://github.com/galaxyproject/galaxy/pull/8104 but until then that'll be necessary.

client/galaxy/scripts/entry/panels/admin-panel.js
```diff
@@ -88,7 +88,13 @@ const AdminPanel = Backbone.View.extend({
                 title: _l("Tool Management"),
                 items: [
                     {
-                        title: _l("Install new tools"),
+                        title: _l("Install or Uninstall"),
```
mvdbeek, line 91: How do you feel about `Install Repositories` ? It's cool that you can uninstall repos here, along with seeing the various revisions, but I think most of the time if an admin want to uninstall a repo it's easier to look at a list of installed repos and then decide on those that should be uninstalled

mvdbeek (approved):
Beautiful, thanks so much @guerler, this works really well!
There are still a couple of things that don't work a 100% right, but that are not the fault of the PR:

- the last updated date is incorrect
- browsing by category misses a lot of repos
- searching for repos doesn't always find the repo, even if it exists

We should fix those too, but I think the fault is with the toolshed search. Similarly we should drop the beta install ASAP.

mvdbeek:
Oh, and another thing I noticed is that suites don't ask for the tool panel section that they should be installed into, that is also pretty important.

client/galaxy/scripts/components/Toolshed/RepositoryDetails.vue
```diff
@@ -0,0 +1,177 @@
...
+        },
+        onOk: function() {
+            this.showSettings = false;
+        },
+        onHide: function() {
+            this.showSettings = false;
+        },
+        setupRepository: function(details) {
+            this.selectedChangeset = details.changeset_revision;
+            this.selectedRequiresPanel = details.includes_tools_for_display_in_tool_panel;
```
mvdbeek: How about
```suggestion
            this.selectedRequiresPanel = details.includes_tools_for_display_in_tool_panel || details.repository.type == "repository_suite_definition";
```
? That would display the panel selection when installing a suite.

client/galaxy/scripts/entry/panels/admin-panel.js
martenson: I think the goal is to gravitate towards a single UI for install/manage/uninstall _all_ tools/repos. So ultimately this is going to be something like `Tools` or `Tool Repositories`.

mvdbeek: So you mean you would have another tab/view/whatever that would be equivalent to the `Manage tools` view that we have now ?

martenson: yes, broadly equivalent in function

PR #8091, opened 2019-06-04
Added option to delete or preserve k8s pods after completion

nuwang:
This PR adds an option to delete job pods after running to completion. It follows the conventions of the `cleanup_job` option in the galaxy config, and provides options between `always`, `onsuccess` and `never`.

### DIFF
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -57,7 +57,7 @@ class KubernetesJobRunner(AsynchronousJobRunner):
             k8s_default_requests_memory=dict(map=str, default=None),
             k8s_default_limits_cpu=dict(map=str, default=None),
             k8s_default_limits_memory=dict(map=str, default=None),
-            k8s_cleanup_job=dict(map=str, default="always"),
+            k8s_cleanup_job=dict(map=str, valid=lambda s: s in set(["onsuccess", "always", "never"]), default="always"),
             k8s_pod_retries=dict(map=int, valid=lambda x: int >= 0, default=3),
             k8s_pod_retrials=dict(map=int, valid=lambda x: int >= 0, default=3),
             k8s_walltime_limit=dict(map=int, valid=lambda x: int(x) >= 0, default=172800))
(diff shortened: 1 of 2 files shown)


mvdbeek:
Thanks a lot @nuwang , I've added some minor things in https://github.com/nuwang/galaxy/pull/3, with those in place I'm +1 on the PR!

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -465,9 +474,25 @@ def _handle_job_failure(self, job, job_state, reason=None):
                 job_state.fail_message = "More pods failed than allowed. See stdout for pods details."
         job_state.running = False
         self.mark_as_failed(job_state)
-        job.scale(replicas=0)
+        self.__cleanup_k8s_job(job)
         return None
 
+    def __cleanup_k8s_job(self, job):
+        k8s_cleanup_job = self.runner_params['k8s_cleanup_job']
+        job_failed = (job.obj['status']['failed'] > 0
```
nuwang, line 491: @mvdbeek The `job_failed` test on this line only checks whether the k8s job pod has failed, not whether the galaxy job has failed. I suppose we should also leave the k8s job behind if the Galaxy job fails? Also, another, related, issue is that the k8s pods do not have any stdout or stderr, as the output is being redirected to files. Therefore, pods that are left behind offer little in terms of debugging potential. It also kind of goes against the grain of k8s, where pod logs could be coupled with a logstash/kibana combo for tracking job failures. Of course, I suppose the job logs could be obtained from Galaxy and sent to logstash/kibana too. What are your thoughts on this?

mvdbeek: > Also, another, related, issue is that the k8s pods do not have any stdout or stderr, as the output is being redirected to files.

Are you sure about that ? I had modified the runner specifically for that. If they're being swallowed now we can surely recover them.

nuwang: So far I've not seen any output at all on any of the pods, regardless of success or failure. I assumed that was because the galaxy job script was redirecting output to files, and I didn't spot a `tee` or anything that redirected output to both. Is there some option I need to set?

mvdbeek: It should happen here: https://github.com/galaxyproject/galaxy/blob/bf31260192eede738e5e9e41b1cd1b490449333b/lib/galaxy/jobs/command_factory.py#L257 but we may need to adapt that to the new stdout/stderr collection strategy. I can add a test for this.

nuwang: That's great, I thought I saw somewhere that was supposed to happen, so that clarifies things. Should I also add an explicit check to avoid cleanup if galaxy determines the job as failed? I kind of recall that some tools incorrectly write errors to stdout which Galaxy explicitly detects?

mvdbeek: I'd say no. If the job hasn't been killed by the runner and the error is instead inferred by some other condition like presence of stderr I don't think there'd be any value in keeping the kubernetes job around.

nuwang:
@martenson @mvdbeek I just realised that this PR was never merged into 19.09 although it was reviewed and approved because of some conflicts. I've rebased and fixed the conflicts, can we still merge this into 19.09?

martenson:
I'll let Marius re-evaluate but see no problem in getting this in. Changed milestone to 19.09.

mvdbeek (approved):

mvdbeek:
Nice, thanks @nuwang!

PR #8101, opened 2019-06-05
Remove unresolved function causing console error

WilliamHolden:
Line 113 calls a function hoverImage which does not exist. This logic was changed in the past and the hoverImage function was removed. I removed the function for now since it's throwing a console error. Let me know if you prefer that I reimplement the function instead of just removing it. Old code can be seen here: 
https://github.com/galaxyproject/galaxy/blob/release_18.09/client/galaxy/scripts/mvc/workflow/workflow-view-data.js


dannon:
@WilliamHolden No need to replace that ancient function, thanks for this cleanup.  Failing test is unrelated.

PR #8103, opened 2019-06-05
[19.05] Fix k8s runasuser

nuwang:
Based on the discussion here: https://github.com/galaxyproject/galaxy/pull/8089 I'm reopening this as a bug fix, as the k8srunner's job pods run as the default container user, which may be root, and therefore, incorrectly sets file ownership on the shared filesystem.

This PR allows for specifying the docker user and group when running the job pod, so that file ownership propagates as expected. The integer values for the uid and gid can be specified, and it also allows for the special values $uid and $gid that will be replaced by the job handler's uid and gid respectively.


afgane (approved):

pcm32:
Great! is this going to be in time for 19.05? Would be great to have it there.

martenson:
@pcm32 It target the release branch, so yeah!

PR #8104, opened 2019-06-06
Swap out galaxy.util.odict for core python collections.OrderedDicts

### LINKED ISSUE #8093: Wrap display_application cheetah fill_template in try/except
as we do for tools.
Otherwise starting without a datatype_conf.xml file (or a very empty one) could lead to
```
galaxy.web.framework.decorators ERROR 2019-06-04 16:38:43,484 [p:74,w:1,m:0] [uWSGIWorker1Core0] Uncaught exception in exposed API method:
File "/galaxy/server/lib/galaxy/web/framework/decorators.py", line 282, in decorator
File "/galaxy/server/lib/galaxy/webapps/galaxy/api/history_contents.py", line 82, in index
return self.__index_v2(trans, history_id, **kwd)
File "/galaxy/server/lib/galaxy/webapps/galaxy/api/history_contents.py", line 891, in __index_v2
user=trans.user, trans=trans, view='detailed', **serialization_params))
File "/galaxy/server/lib/galaxy/managers/base.py", line 692, in serialize_to_view
return self.serialize(item, all_keys, **context)
File "/galaxy/server/lib/galaxy/managers/hdas.py", line 369, in serialize
return super(HDASerializer, self).serialize(hda, keys, user=user, **context)
File "/galaxy/server/lib/galaxy/managers/datasets.py", line 604, in serialize
serialized = super(DatasetAssociationSerializer, self).serialize(dataset_assoc, keys, **context)
File "/galaxy/server/lib/galaxy/managers/base.py", line 605, in serialize
returned[key] = self.serializers[key](item, key, **context)
File "/galaxy/server/lib/galaxy/managers/hdas.py", line 376, in serialize_display_apps
for display_app in hda.get_display_applications(trans).values():
File "/galaxy/server/lib/galaxy/model/__init__.py", line 2756, in get_display_applications
return self.datatype.get_display_applications_by_dataset(self, trans)
File "/galaxy/server/lib/galaxy/datatypes/data.py", line 551, in get_display_applications_by_dataset
value = value.filter_by_dataset(dataset, trans)
File "/galaxy/server/lib/galaxy/datatypes/display_applications/application.py", line 309, in filter_by_dataset
if link_value.filter_by_dataset(data, trans):
File "/galaxy/server/lib/galaxy/datatypes/display_applications/application.py", line 95, in filter_by_dataset
if fill_template(filter_elem.text, context=context) != filter_elem.get('value', 'True'):
File "/galaxy/server/lib/galaxy/util/template.py", line 91, in fill_template
raise first_exception or e
NotFound: cannot find ‘site_id’
```

dannon:
Testing the waters (i.e. the full test suite) in swapping our odict for real OrderedDicts

WIP, fixes https://github.com/galaxyproject/galaxy/issues/8093, with an explanation why at https://github.com/galaxyproject/galaxy/issues/8093#issuecomment-499328125


dannon:
Well, there's an optimistic refactoring first pass, time to look at a few of these more closely.

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -29,7 +29,7 @@ def visit_input_values(inputs, input_values, callback, name_prefix='', label_pre
 
     >>> from xml.etree.ElementTree import XML
     >>> from galaxy.util.bunch import Bunch
-    >>> from galaxy.util.odict import odict
+    >>> from collections import OrderedDict
```
nsoranzo: We don't check for import order in doctests, but for consistency move this to first?

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -270,7 +270,7 @@ def populate_state(request_context, inputs, incoming, state, errors={}, prefix='
     Populates nested state dict from incoming parameter values.
     >>> from xml.etree.ElementTree import XML
     >>> from galaxy.util.bunch import Bunch
-    >>> from galaxy.util.odict import odict
+    >>> from collections import OrderedDict
```
nsoranzo: Move to first import.

lib/galaxy/tools/toolbox/panel.py
```diff
@@ -112,7 +112,7 @@ def to_dict(self, **kwds):
         return super(ToolSectionLabel, self).to_dict()
 
 
-class ToolPanelElements(odict, HasPanelItems):
+class ToolPanelElements(OrderedDict, HasPanelItems):
```
nsoranzo: This is the only class using a method that exists only for `odict` but not for `OrderedDict`, i.e. `insert()` which in used in `update_or_append()`, `replace_tool` and `insert_tool`.

lib/galaxy/util/topsort.py
```diff
@@ -36,7 +36,7 @@
 many methods to help analyze and break the cycles.  This requires
 a good deal more code than topsort itself!
 """
-from galaxy.util.odict import odict as OrderedDict
+from collections import OrderedDict as OrderedDict
```
nsoranzo:
```suggestion
from collections import OrderedDict
```

lib/galaxy/webapps/galaxy/api/users.py
```diff
@@ -689,7 +689,7 @@ def _add_filter_inputs(self, factory, filter_types, inputs, errors, filter_type,
             inputs.append({'type': 'section', 'title': filter_title, 'name': filter_type, 'expanded': True, 'inputs': filter_inputs})
 
     def _get_filter_types(self, trans):
-        return odict([('toolbox_tool_filters', {'title': 'Tools', 'config': trans.app.config.user_tool_filters}),
+        return OrderedDict([('toolbox_tool_filters', {'title': 'Tools', 'config': trans.app.config.user_tool_filters}),
                       ('toolbox_section_filters', {'title': 'Sections', 'config': trans.app.config.user_tool_section_filters}),
                       ('toolbox_label_filters', {'title': 'Labels', 'config': trans.app.config.user_tool_label_filters})])
```
nsoranzo: Indentation

lib/galaxy/workflow/modules.py
```diff
@@ -681,7 +681,7 @@ def get_inputs(self):
                 # item 0 is option description, item 1 is value, item 2 is "selected"
                 option[2] = True
                 input_parameter_type.static_options[i] = tuple(option)
-        return odict([("parameter_type", input_parameter_type),
+        return OrderedDict([("parameter_type", input_parameter_type),
                       ("optional", BooleanToolParameter(None, Element("param", name="optional", label="Optional", type="boolean", value=optional)))])
```
nsoranzo: Indentation

test/unit/tools/test_execution.py
```diff
@@ -185,7 +185,7 @@ def execute(self, tool, trans, **kwds):
             if num_calls > self.error_message_after_excution:
                 return None, "Test Error Message"
 
-        return galaxy.model.Job(), odict(dict(out1="1"))
+        return galaxy.model.Job(), OrderedDict(dict(out1="1"))
```
nsoranzo:
```suggestion
        return galaxy.model.Job(), OrderedDict(out1="1")
```

lib/galaxy/util/topsort.py
dannon: Batched this with the other fixes, thanks!

test/unit/tools/test_execution.py
dannon: Batched with d83ceb4

lib/galaxy/tools/toolbox/panel.py
nsoranzo: We discussed this on Gitter, these were the main points:
- The `insert()` method was added to `odict` in commit https://github.com/galaxyproject/galaxy/commit/7853e71dad4cfbde60a847197586769b91d2def3 .
- There is no alternative implementation that solves the tool panel use case and is as efficient as `odict`. Especially given that Python 3.6 has `OrderedDict` implemented in C, so we cannot add an `insert` method to it any more.
- If we end up keeping `odict` for this use case, it may make sense to leave it where it is since it has been distributed as part of `galaxy-lib` and now `galaxy-util` packages.
- We can remove the unused `reverse()` method of `odict`.

nsoranzo:
@dannon Why is `odict` needed for `MetadataSpecCollection` ? It doesn't use `insert()` I think, so I may be missing something.

dannon:
@nsoranzo I swapped it back because it was the other place where we rely on extending odict, and the swap didn't seem like it was going to be a straightforward one (https://gist.github.com/76c7890eecdd69b8bf5af286821de526 doesn't work, and I didn't want/care to dig deeper on this one right now -- bleh, that was the wrong gist, but it gets the point across; in another stash I'd correctly called super/__init__, but I don't have the traceback in front of me).

If you want to add a change that addresses that one, that'd be great.

nsoranzo:
> If you want to add a change that addresses that one, that'd be great.

Done! I've also rebased to include the integration tests fix.

nsoranzo (approved):

PR #8106, opened 2019-06-06
[19.05] Restored google analytics inits in mako templates

Nerdinacan:
Looks like I commented out a call to the analytics and sentry inits during my reorganization of all those init scripts. Restored that, added some minor parameter checking to make sure the google analytics code is actually configured.

### DIFF
--- a/templates/base.mako
+++ b/templates/base.mako
@@ -65,7 +65,9 @@
 
     ${ galaxy_client.load( app=self.js_app ) }
     ${ galaxy_client.config_sentry( app=self.js_app ) }
-    ${ galaxy_client.config_google_analytics( app=self.js_app ) }
+    %if self.js_app and self.js_app.config and self.js_app.config.ga_code:
+        ${ galaxy_client.config_google_analytics(self.js_app.config.ga_code) }
+    %endif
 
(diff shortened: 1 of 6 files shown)


martenson:
@Nerdinacan Thank you. Please target 19.05 as we want to have this on Main and we broke it in the release.

edit: is there a reasonable way how we can test calling to sentry and ga? Maybe load the page and check for the scripts network call presence?

Nerdinacan:
There's a console log message for sentry initialization already and I've added a new one there that explicitly mentions that it's trying to initialize analytics and complains if there's no google identifier in the configs.

As for an automated test: javascript unit-tests are out since this all happens on the mako templates, prior to javascript being loaded. It is possible in principle to write selenium tests that check to see if a script has been loaded on page, though I'm reluctant to write a pile of integration tests that essentially test the horrible maze of mako templates we're actively working very hard to destroy.

Nerdinacan:
I just force-pushed a version that was explicitly built off the 19.05 tag, though I'm pretty sure that's what the original was now that I look at the log... Oh well. Force-push mania.

dannon (approved):

PR #8107, opened 2019-06-06
Use official "login with" buttons for OIDC

hexylena:
What do people think of this? Swapping to login buttons:

image

I am not a Vue person, so, please comment mainly on the idea, and make whatever changes you like directly to this PR to make it more like the vue code that everyone else is writing.

### DIFF
--- a/client/galaxy/scripts/components/login/Login.vue
+++ b/client/galaxy/scripts/components/login/Login.vue
@@ -30,10 +30,17 @@
                         </b-card-footer>
                     </b-card>
                 </b-form>
-                <b-button v-for="idp in oidc_idps" :key="idp" class="d-block mt-3" @click="submitOIDCLogin(idp)">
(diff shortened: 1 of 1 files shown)


martenson:
I am fine with it, wouldn't be too terrible if we converted them to some more fitting/neutral colors (like B&W?).

dannon:
And same size/ratio at least?  Otherwise it's going to look like a hodgepodge of stickers.

hexylena:
At least for Elixir, I'm not sure they'd be happy with us modifying their button like that. The whole point is to make it easily recognisable for users of their auth service that you can "use your favourite academic login here", they have style guidelines and everything. I assume the same applied for google, that there is some pressure to retain the original logos. I'd imagine it's more OK to resize/crop as needed.

addition: I recognise that the google one is a bit hard to see but I picked first "Signin with google" button I found on their page, perhaps there are others.

Issue #8099, opened 2019-06-05
Show welcome after OIDC login


hexylena:
I'm testing release_19.05 currently, and noticed that after logging in with OIDC, we get a blank main panel (normally would be white, I have browser css to make it black.) It would be helpful if this showed the normal welcome.html instead

![image](https://user-images.githubusercontent.com/458683/58960797-2cf4b180-87a8-11e9-97cf-ba47615939f6.png)

hexylena:
@VJalili would you be able to take a look at this? We're getting ready to deploy elixir OIDC with 19.05 and I worry about the user experience here.

hexylena:
Another note, I accidentally hit `refresh` today after doing a test, the URL was still set to the elixir callback and it attempted to resubmit that generating an error. If we can redirect to / after the callback this would be optimal I think

VJalili:
> noticed that after logging in with OIDC, we get a blank main panel ... It would be helpful if this showed the normal welcome.html instead

@dannon any thoughts? 

https://github.com/galaxyproject/galaxy/blob/47fd60fccec9e52c78e62ae00e41a8725385eff7/lib/galaxy/web/base/controller.py#L270-L272

> Another note, I accidentally hit refresh today after doing a test, the URL was still set to the elixir callback and it attempted to resubmit that generating an error.

I wonder if this is related to the previous issue, because after callback is handled that URL should not be available.

martenson:
resolved per https://github.com/galaxyproject/galaxy/pull/8110#issuecomment-814894595

PR #8110, opened 2019-06-06
Redirect to welcome page after successful handling of oidc callback.

VJalili:
Fix xref: https://github.com/galaxyproject/galaxy/issues/8099

This PR implements: 
- redirect to welcome page after oidc callback is successfully handled; 
- remove callback url, hence remove the ability of resending the callback url by refreshing the page. 

One issue is that it shows welcome page without the side panels. Any thoughts?

ping @dannon @erasche

### DIFF
--- a/lib/galaxy/web/base/controller.py
+++ b/lib/galaxy/web/base/controller.py
@@ -280,7 +280,10 @@ class JSAppLauncher(BaseUIController):
         Should not be used with url_for -- see
         (https://github.com/galaxyproject/galaxy/issues/1878) for why.
         """
-        self._check_require_login(trans)
+        if kwd.get("successful_oidc_callback", False):
+            trans.response.send_redirect(web.url_for(controller="root", action="welcome"))
+        else:
+            self._check_require_login(trans)
         return self._bootstrapped_client(trans, **kwd)
 
     def _bootstrapped_client(self, trans, app_name='analysis', **kwd):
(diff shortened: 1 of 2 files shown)


lib/galaxy/web/base/controller.py
```diff
@@ -280,7 +280,10 @@ def client(self, trans, **kwd):
         Should not be used with url_for -- see
         (https://github.com/galaxyproject/galaxy/issues/1878) for why.
         """
-        self._check_require_login(trans)
+        if kwd.get("successful_oidc_callback", False):
```
dannon, line 283: I haven't had a chance to set this up and tinker with it locally, but it doesn't seem right to push this logic into the default client endpoint.  I can try a bit later to see what I can come up with, if you'd like.

VJalili: We can push it to `authnz.callback`, but I remember we had issues with that previously. Regardless, please feel free to modify if you see a more appropriate solution.

VJalili:
ping @hexylena

hexylena:
@bgruening you might deploy on test?

bgruening:
@VJalili @hexylena thanks, we will note that as potential backport.

mvdbeek:
Thanks @VJalili, I vaguely recall we've had something that addresses this in the 21.01 cycle. Feel free to re-open if that's still needed.

hexylena:
Tested, seems to work fine now on EU

PR #8118, opened 2019-06-07
Allow declaring depedency resolvers right in Galaxy config, serialization.

jmchilton:


lib/galaxy/util/plugin_config.py
```diff
@@ -30,12 +30,12 @@ def plugins_dict(module, plugin_type_identifier):
     return plugin_dict
 
 
-def load_plugins(plugins_dict, plugin_source, extra_kwds={}):
+def load_plugins(plugins_dict, plugin_source, extra_kwds={}, plugin_type_keys=['type']):
```
mvdbeek:
```suggestion
def load_plugins(plugins_dict, plugin_source, extra_kwds=None, plugin_type_keys=('type',)):
    if extra_kwds is None:
        extra_kwds = {}
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -2364,6 +2364,12 @@ mapping:
       job_config:
         !include job_config_schema.yml
 
+      dependency_resolvers:
+        type: seq
+        desc: |
+          Rather than specifying a dependency_resolvers_config_file, the definition of the
+          resolvers to enable can be embedded into Galaxy's config with this option.
```
mvdbeek, line 2371:
```suggestion
          resolvers to enable can be embedded into Galaxy's config with this option.
          This has no effect if a dependency_resolvers_config_file is used.
```
It's implied by the `rather than`, but it might still be helpful ?

mvdbeek (approved):
I like it!

PR #8120, opened 2019-06-08
Pin SQLAlchemy to ``<1.3`` and update all dependencies

VJalili:
This PR does: 

- pins social-core and sqlalchemy versions;
- updates dependencies using `make update-dependencies`

ping @nsoranzo

### DIFF
--- a/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
+++ b/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
@@ -165,7 +165,7 @@ schema-salad==2.7.20181126142424
 shellescape==3.4.1
 simplejson==3.16.0
 six==1.11.0
-social-auth-core[openidconnect]==3.1.0
+social-auth-core[openidconnect]==3.1.0+gx0
 sqlalchemy-migrate==0.12.0
 sqlalchemy-utils==0.33.11
 sqlalchemy==1.2.18
--- a/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
+++ b/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
@@ -38,7 +38,7 @@ cffi==1.12.3
 chardet==3.0.4
 cheetah3==3.2.2
 cliff==2.14.1
-cloudauthz==0.2.0
+cloudauthz==0.3.0
 cloudbridge==2.0.0
 cmd2==0.8.9
 contextlib2==0.5.5 ; python_version < '3.5'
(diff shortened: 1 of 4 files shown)


lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
```diff
@@ -144,32 +146,32 @@ python-novaclient==11.0.0
...
-s3transfer==0.2.0
+ruamel.yaml==0.15.97
+s3transfer==0.2.1
 scandir==1.10.0 ; python_version < '3.5'
 schema-salad==2.7.20181126142424
 shellescape==3.4.1
 simplejson==3.16.0
 six==1.11.0
-social-auth-core[openidconnect]==3.1.0+gx0
+social-auth-core[openidconnect]==3.1.0
```
nsoranzo: This was changed directly in this file in commit 5723e15cce8bc493a782b5aa18b894d230da6261 , so `make update-dependencies` is incorrectly reverting it. Can you apply:
```diff
diff --git a/lib/galaxy/dependencies/pipfiles/default/Pipfile b/lib/galaxy/dependencies/pipfiles/default/Pipfile
index 6741c92561..02777c6b6b 100644
--- a/lib/galaxy/dependencies/pipfiles/default/Pipfile
+++ b/lib/galaxy/dependencies/pipfiles/default/Pipfile
@@ -75,7 +75,7 @@ pyparsing = "*"
 "Fabric3" = "*"
 paramiko = "*"
 python-genomespaceclient = "<2.0"
-social_auth_core = {version = "==3.1.0", extras = ['openidconnect']}
+social_auth_core = {version = "==3.1.0+gx0", extras = ['openidconnect']}
 cloudauthz = "<=0.2.0"
 gxformat2 = "*"
 
```
to `lib/galaxy/dependencies/pipfiles/default/Pipfile` ? Or even pin it to the new 3.2.0 , if that works.

Then re-run `make update-dependencies`.

VJalili: @nsoranzo reverted the change of social-auth-core; however, running `make update-dependencies` keeps dropping `+gx0` (and also updates `sqlalchemy` to its current latest version that causes some unrelated issues), so I did revert without running `make update-dependencies`. 

Similarly, running `make update-dependencies` reverts `social-auth-core[openidconnect]==3.2.0` to `3.1.0`. So, I left that version change to be addressed in a different PR.

nsoranzo: Did you update the correct file `lib/galaxy/dependencies/pipfiles/default/Pipfile` before re-running `make update-dependencies` as in my comment above?
It did the right thing for me.
Also, it would be good to pin `sqlalchemy` to the current version in the same file (again, before re-running `make update-dependencies`).

nsoranzo: I've pushed a commit implementing these to your branch.

VJalili: Yes, my mistake, I've been changing a different file. Fixed it following your comment, but I guess we pushed at the same time :)  Thanks for fixing it.

nsoranzo (approved):

PR #8122, opened 2019-06-09
Add support for cleaning up k8s jobs after a specified number of seconds since finish

nuwang:
This adds support for cleaning up finished jobs after a specified number of seconds using the k8s ttlSecondsAfterFinished feature. However, this is still an alpha feature introduced in k8s v1.12 which must be enabled through featureGates, and therefore, I'll leave this PR as a draft till such time as the future of that feature becomes clear.

### DIFF
--- a/lib/galaxy/jobs/runners/util/pykube_util.py
+++ b/lib/galaxy/jobs/runners/util/pykube_util.py
@@ -88,7 +88,7 @@ def _find_object_by_name(clazz, pykube_api, object_name, namespace=None):
     return obj
 
 
-def stop_job(job, cleanup="always"):
+def stop_job(job, cleanup="always", use_job_ttl=None):
     job_failed = (job.obj['status']['failed'] > 0
                   if 'failed' in job.obj['status'] else False)
     # Scale down the job just in case even if cleanup is never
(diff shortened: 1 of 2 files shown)


mvdbeek:
I've bumped it to 20.05, also looks like something went wrong merging in dev.

nuwang:
@mvdbeek I've rebased on dev. Feature is still in alpha.

nuwang:
Since the feature has been in k8s for a while now, and because it is optional with no negative functional impact, I think it can be safely merged in. The build failures appear to be unrelated.

afgane (approved):

lib/galaxy/jobs/runners/util/pykube_util.py
```diff
@@ -97,13 +97,17 @@ def stop_job(job, cleanup="always"):
...
-        job.api.raise_for_status(r)
+        if use_job_ttl is None:
+            delete_options = {
+                "apiVersion": "v1",
+                "kind": "DeleteOptions",
+                "propagationPolicy": "Background"
+            }
+            r = job.api.delete(json=delete_options, **job.api_kwargs())
+            job.api.raise_for_status(r)
+        else:
```
mvdbeek, line 108: What happens when the user requests cancellation of the job (via the API or by deleting all outputs) ?
Will this only happen when the ttl is reached ?

nuwang: I think you're right, this would cause the job to not be cleaned up till it runs to completion, even if the user requests cancellation. 

The original reason for doing this was to be able to inspect recently completed containers, without retaining them indefinitely, but I now feel that this no longer serves a purpose.

I guess we can just revert the stop_job function to what it was, and just send the `ttlSecondsAfterFinished` as part of the job spec anyway. This would provide an additional layer of confidence that clean up will occur, but cleanup will usually be more or less be immediate (depending on the policy).

I'll issue a new PR for this.

nuwang: Fixed in: https://github.com/galaxyproject/galaxy/pull/9784. Thanks for flagging this @mvdbeek

nuwang: Just noting that specifying `cleanup="never"`along with `ttlSecondsAfterFinished` should restore similar behaviour with the added caveat that the job working directory may need to be manually cleaned up.

PR #7531, opened 2019-03-14
[WIP] Rev to sqlalchemy 1.3.

jmchilton:
We should do this or pin 1.2 in our Pipfile.

PR #8126, opened 2019-06-10
Update Python dependencies, unpin SQLAlchemy

nsoranzo:
Alternative to https://github.com/galaxyproject/galaxy/pull/7531 , let's see what is still breaking.


jmchilton:
Awesome - thanks so much for the fixes!

PR #8137, opened 2019-06-12
Fix loading job_conf from YAML files.

jmchilton:
I think previous testing of this dependency handling stuff was just with embedded job conf.

### DIFF
--- a/lib/galaxy/dependencies/__init__.py
+++ b/lib/galaxy/dependencies/__init__.py
@@ -69,7 +69,8 @@ class ConditionalDependencies(object):
                     pass
             else:
                 try:
-                    job_conf_dict = yaml.safe_load(job_conf_path)
+                    with open("job_conf_path", "r") as f:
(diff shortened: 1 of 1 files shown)


lib/galaxy/dependencies/__init__.py
```diff
@@ -69,7 +69,8 @@ def load_job_config_dict(job_conf_dict):
                     pass
             else:
                 try:
-                    job_conf_dict = yaml.safe_load(job_conf_path)
+                    with open("job_conf_path", "r") as f:
```
mvdbeek, line 72: Maybe we can also tighten the
`if '.xml' in job_conf_path:` with `if job_conf_path.endswith('.xml')` ?

mvdbeek (approved):

lib/galaxy/dependencies/__init__.py
jmchilton: ``.xml.sample`` though

PR #8138, opened 2019-06-12
Use Python 3 for package/ unit testing.

jmchilton:
Is this all I need to do... I guess we will see.


jmchilton:
Ugh - I hit the jstree unit tests not working again @martenson - this time just in the context of Python 3 because of the relative imports. I'm going to fix the doctests since you relented to changing the file previously - I hope that is okay.

nsoranzo (approved):

jmchilton:
Thanks for the fixes @nsoranzo!

PR #8143, opened 2019-06-12
[WIP] Fix env var name GALAXY_TOOL_DEPENDENCY_DIR

jdavcs:
This is an apparent bug. (unless I'm missing something obvious - in which case, sorry!)
incorrect: GALAXY_TOOL_DEPENDENCY_DIR
correct: GALAXY_TEST_TOOL_DEPENDENCY_DIR

The present code overrides the value without reading it.

See this line: https://github.com/galaxyproject/galaxy/blob/72310185d9a74b20ab1a5de7a25e6bf5cd4efd62/test/base/driver_util.py#L244

Also this: 
https://github.com/galaxyproject/galaxy/blob/72310185d9a74b20ab1a5de7a25e6bf5cd4efd62/run_tests.sh#L215

### DIFF
--- a/test/base/driver_util.py
+++ b/test/base/driver_util.py
@@ -162,7 +162,7 @@ def setup_galaxy_config(
         library_import_dir = None
     job_config_file = os.environ.get('GALAXY_TEST_JOB_CONFIG_FILE', default_job_config_file)
     tool_path = os.environ.get('GALAXY_TEST_TOOL_PATH', 'tools')
-    tool_dependency_dir = os.environ.get('GALAXY_TOOL_DEPENDENCY_DIR', None)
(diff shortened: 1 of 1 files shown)


jdavcs:
OK, so it appears that as a result of this change tool_dependency_dir is no longer none, and, I think, this [line](https://github.com/galaxyproject/galaxy/blob/72310185d9a74b20ab1a5de7a25e6bf5cd4efd62/test/base/driver_util.py#L161) is no longer executed, so temp dir is never created. Which breaks all those tests.

Was I mistaken and is this not a bug? (the only place this env var is read is [here](https://github.com/galaxyproject/galaxy/blob/72310185d9a74b20ab1a5de7a25e6bf5cd4efd62/test/shed_functional/base/twilltestcase.py#L73)) If so, I'll close the PR.

jdavcs:
I'll get to this later, for now this is too low of a priority.

PR #8152, opened 2019-06-15
Re-try dependency manager serialization and embedded config.

jmchilton:
PR #8118 - adding to_dict to the dependency manager and an embedded loading of dependency resolution options from galaxy.yml - was really quite poorly thought through on my part.

Certain options could be embedded - but only the resolvers and the actual dictified representation of the DependencyManager couldn't be actually embedded properly in the app config - but this is exactly what I wanted to do on the Pulsar side and it would make no sense to have the exact same options in both apps and configured differently.

So now the actual dictified representation of the whole DependencyManager can be directly embedded in galaxy.yml as the ``dependency_resolution:`` parameter. All the parameters can still be configured at the top level of the YAML file but with 'dependency' in the config option name (some prefixes, some suffixes, etc... a hodgepodge that has grown organically). Give the messiness of the existing options might be better to encourage using ``dependency_resolution:`` in the future.

I think this further cleaning hopefully improves the interface to that module. The simplifications to scripts/manage_tool_dependencies.py  and config.py seem like good ones.

~Note to @jmchilton - add tests for "none" handling in tool_dependency_dir option.~ (Done)

### DIFF
--- a/lib/galaxy/tool_util/deps/__init__.py
+++ b/lib/galaxy/tool_util/deps/__init__.py
@@ -10,7 +10,8 @@ from collections import OrderedDict
 
 from galaxy.util import (
     hash_util,
-    plugin_config
+    plugin_config,
+    string_as_bool,
 )
 from galaxy.util.oset import OrderedSet
 from .container_resolvers import ContainerResolver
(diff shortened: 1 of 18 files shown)


lib/galaxy/tool_util/deps/__init__.py
```diff
@@ -31,27 +32,59 @@
...
-        "cache": getattr(config, "use_cached_dependency_manager", False),
-        "app_config": config,
-        "dependency_resolvers": getattr(config, "dependency_resolvers", None),
-    }
-    return build_dependency_manager_from_dict(config_dict)
+def build_dependency_manager(app_config_dict=None, resolution_config_dict=None, conf_file=None, default_tool_dependency_dir=None):
+    """Build a DependencyManager object from app and/or resolution config.
 
+    If app_config_dict is specified, it should be application configuration information
+    and configuration options are generally named to identify the context is dependency
```
mvdbeek:
```suggestion
    and configuration options are generally named to identify the context of dependency
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -2371,6 +2375,31 @@ mapping:
...
+            desc: See description of 'precache_dependencies'.
+          cache_dir:
+            type: str
+            default: '<tool_dependency_cache_dir>/_cache'
+            desc: See description of 'tool_dependency_cache_dir'.
+          default_base_path:
+            default: database/dependencies
+            desc: See description of 'tool_dependency_dir'.
+          resolvers:
+            desc: See description of 'dependency_resolvers'.
```
mvdbeek, line 2402:
```suggestion
            desc: See description of 'dependency_resolvers'.

```

mvdbeek (approved):
Awesome, works great, I prefer this a lot over the separate config file.

PR #8154, opened 2019-06-16
Kubernetes / CLI runner integration test fixes

mvdbeek:
These are required to run the kubernetes tests. The job config files now have to contain `.xml` (https://github.com/galaxyproject/galaxy/pull/7854). `stream_to_file` may receive StringIO objects, so chunk may be unicode. In that case we need to call `encode`.

### DIFF
--- a/lib/galaxy/datatypes/sniff.py
+++ b/lib/galaxy/datatypes/sniff.py
@@ -76,6 +76,8 @@ def stream_to_open_named_file(stream, fd, filename, source_encoding=None, source
         else:
             # Compressed files must be encoded after they are uncompressed in the upload utility,
             # while binary files should not be encoded at all.
+            if isinstance(chunk, text_type):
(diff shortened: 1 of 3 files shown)


jmchilton:
Sorry about breaking that :(

jmchilton (approved):

mvdbeek:
No worries, was easy enough to figure out. Also I reviewed the PR, so could have thought about that myself. On top of that I should have added a suffix in the first place. And then I should probably move this to yml, since it'll be easier to adjust dynamically ... :). And https://github.com/galaxyproject/galaxy/pull/8050 broke some part of the upload by url functionality ...

nsoranzo:
I think the "job config file name should contain `.xml` if it's in XML format" has been introduced in https://github.com/galaxyproject/galaxy/pull/7854 , not #8118 , or am I missing something?

mvdbeek:
Yes, you're right of course!

PR #8162, opened 2019-06-17
fix for test data in sub directories

bernt-matthias:
in order to get the data set names for test data `basename` should be applied. otherwise tests are failing 
for tools that
- organize test data is in subdirectories and
- assume that the id of the data set will be identical to the file name. 

For an example see https://github.com/galaxyproject/tools-iuc/pull/2439.

@nsoranzo suggested that this was introduced here: https://github.com/galaxyproject/galaxy/commit/6791f3b99905ac0db0a8e2c23f38a4bce8b1d3c8 (galaxy 19.01). 

I just played around a bit and this seems to work. But its not unlikely that I overlooked something. Also a test would be nice ensure this in the future.

### DIFF
--- a/lib/galaxy/tools/verify/interactor.py
+++ b/lib/galaxy/tools/verify/interactor.py
@@ -341,7 +341,7 @@ class GalaxyInteractorApi(object):
                 tool_input.update({
                     "files_%d|type" % i: "upload_dataset",
                 })
-            name = os.path.basename(test_data['name'])
+            name = test_data['name']
         else:
             name = os.path.basename(fname)
             tool_input.update({
--- a/lib/galaxy/tools/verify/interactor.py
+++ b/lib/galaxy/tools/verify/interactor.py
@@ -341,9 +341,9 @@ class GalaxyInteractorApi(object):
                 tool_input.update({
                     "files_%d|type" % i: "upload_dataset",
                 })
-            name = test_data['name']
+            name = os.path.basename(test_data['name'])
         else:
-            name = fname
+            name = os.path.basename(fname)
             tool_input.update({
                 "files_0|NAME": name,
                 "files_0|type": "upload_dataset",


mvdbeek (approved):

lib/galaxy/tools/verify/interactor.py
```diff
@@ -341,9 +341,9 @@ def stage_data_async(self, test_data, history_id, tool_id, force_path_paste=Fals
                 tool_input.update({
                     "files_%d|type" % i: "upload_dataset",
                 })
-            name = test_data['name']
+            name = os.path.basename(test_data['name'])
```
nsoranzo: The change on line 346 should be enough for https://github.com/galaxyproject/tools-iuc/pull/2439 , are you that's actually a path?

bernt-matthias: Not sure about anything .. as usual. My coding in Galaxy is still often based on intuition :/ 

The change on 364 applies `basename` to `fname` which should be equal to `test_data['fname']` (https://github.com/galaxyproject/galaxy/blob/da91d724bcd4329098af01d5807c25bcfa364d85/lib/galaxy/tools/verify/interactor.py#L318). So I intuitively applied basename in both branches of the if statement. Initially I thought about applying directly on line 318. But fname is used at multiple other places, e.g. https://github.com/galaxyproject/galaxy/blob/da91d724bcd4329098af01d5807c25bcfa364d85/lib/galaxy/tools/verify/interactor.py#L353

nsoranzo: I agree with the change on line 346, but I think the one here may be not needed, since that's on `test_data['name']`, not `test_data['fname']` .

nsoranzo (approved):

nsoranzo:
Do we need to backport to 19.01 ?

martenson:
@nsoranzo I would like to do that. At least for 19.05

bernt-matthias:
yep, would be great to have this usable in planemo for the various tool repos.

martenson:
I cherry picked it to 19.05

PR #8179, opened 2019-06-18
Fix job wrapper fail for queued jobs

mvdbeek:
whose input datasets are purged. The error only occurs if `outputs_to_working_directory` is set.

The new test then also revealed that we didn't properly pause jobs with inputs in error state (they'd just be ignored), which is fixed by https://github.com/galaxyproject/galaxy/pull/8179/commits/920925da162b71f63cbd6ad5482a24c7079b3139

### DIFF
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1163,10 +1163,9 @@ class JobWrapper(HasResourceParameters):
         return tool_evaluator
 
     def _fix_output_permissions(self):
-        if self.working_directory_exists():
-            for path in [dp.real_path for dp in self.get_mutable_output_fnames()]:
-                if os.path.exists(path):
-                    util.umask_fix_perms(path, self.app.config.umask, 0o666, self.app.config.gid)
+        for path in [dp.real_path for dp in self.get_mutable_output_fnames()]:
+            if os.path.exists(path):
+                util.umask_fix_perms(path, self.app.config.umask, 0o666, self.app.config.gid)
 
     def fail(self, message, exception=False, tool_stdout="", tool_stderr="", exit_code=None, job_stdout=None, job_stderr=None):
         """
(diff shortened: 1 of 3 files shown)


lib/galaxy/jobs/__init__.py
```diff
@@ -1090,6 +1090,12 @@ def working_directory(self):
                 job, base_dir='job_work', dir_only=True, obj_dir=True)
         return self.__working_directory
 
+    def working_directory_exists(self):
+        job = self.get_job()
```
nsoranzo, line 1102: You can move this line after the `if`, to save a database query when returning `False`.

mvdbeek: Good point!

mvdbeek: Actually that doesn't work, we need it for `job.object_store_id is None:`

nsoranzo: Ouch, sorry!

mvdbeek: but the entire check is wrong, this should just be `return self.app.object_store.exists(job, base_dir='job_work', dir_only=True, obj_dir=True)`, otherwise we return `False` when we shouldn't. Not sure what I was thinking here.

nsoranzo: That I was trying to understand :)

mvdbeek: Rebased with the fix, thanks for thinking this through!

lib/galaxy/jobs/__init__.py
```diff
@@ -1151,9 +1155,10 @@ def _get_tool_evaluator(self, job):
         return tool_evaluator
 
     def _fix_output_permissions(self):
-        for path in [dp.real_path for dp in self.get_mutable_output_fnames()]:
-            if os.path.exists(path):
-                util.umask_fix_perms(path, self.app.config.umask, 0o666, self.app.config.gid)
+        if self.working_directory_exists():
```
nsoranzo: Nitpicking question: would it make sense to move this check to `fail()`, before calling this method? I *think* it's not needed when `_fix_output_permissions()` is called inside `finish()`.

mvdbeek: true, we can do that.

nsoranzo (approved):

Issue #8135, opened 2019-06-12
setting metadata errors when jobs are aborted by cluster


bernt-matthias:
A job setting the metadata (of a huge maf file) is aborted by our cluster after the run time limit. This leads to the following errors (furthermore job remains in yellow state .. should be resubmitted):

```
Traceback (most recent call last):
  File "/work/songalax/galaxy/database/jobs_directory/031/31006/tmpY5cyF_", line 1, in <module>
    from galaxy_ext.metadata.set_metadata import set_metadata; set_metadata()
  File "/gpfs1/data/galaxy_server/galaxy/lib/galaxy_ext/metadata/set_metadata.py", line 120, in set_metadata
    set_meta_kwds = stringify_dictionary_keys(json.load(open(filename_kwds)))  # load kwds; need to ensure our keywords are not unicode
IOError: [Errno 2] No such file or directory: '/work/songalax/galaxy/database/jobs_directory/031/31006/metadata_kwds_HistoryDatasetAssociation_128057_m8CFLA'
]
```

```
Traceback (most recent call last):
  File "lib/galaxy/jobs/runners/__init__.py", line 717, in finish_job
    self._finish_or_resubmit_job(job_state, stdout, stderr, exit_code)
  File "lib/galaxy/jobs/runners/__init__.py", line 459, in _finish_or_resubmit_job
    job_state.job_wrapper.finish(stdout, stderr, exit_code, check_output_detected_state=check_output_detected_state)
  File "lib/galaxy/jobs/__init__.py", line 1475, in finish
    self.tool.exec_after_process(self.app, inp_data, out_data, param_dict, job=job)
  File "lib/galaxy/tools/__init__.py", line 2256, in exec_after_process
    if external_metadata.external_metadata_set_successfully(dataset, app.model.context):
  File "lib/galaxy/model/metadata.py", line 799, in external_metadata_set_successfully
    rval, rstring = json.load(open(metadata_files.filename_results_code))
IOError: [Errno 2] No such file or directory: u'/work/songalax/galaxy/database/jobs_directory/031/31006/metadata_results_HistoryDatasetAssociation_128057_eGMpe1'
```

```
Traceback (most recent call last):
  File "lib/galaxy/jobs/runners/__init__.py", line 130, in run_next
    method(arg)
  File "lib/galaxy/jobs/runners/__init__.py", line 720, in finish_job
    job_state.job_wrapper.fail("Unable to finish job", exception=True)
[...]

PR #8182, opened 2019-06-18
[19.05] Fix external_metadata_set_successfully when metadata file doesn't exist

mvdbeek:
Fixes https://github.com/galaxyproject/galaxy/issues/8135


jmchilton:
Thank you so much!

PR #7734, opened 2019-04-12
hyphy-vision

rdvelazquez:
- Adds a visualization for results from [hyphy-tools](https://github.com/galaxyproject/tools-iuc/tree/master/tools/hyphy).
- Currently live at https://galaxy.hyphy.org
- I think the idea is to eventually add the hyphy-tools to usegalaxy.org; getting this viz in is an important step; there are also some minor revision still to be done to the tools themselves (that will be tracked at http://github.com/galaxyproject/tools-iuc) 

hyphyVision_galaxy

Thank you @stephenshank, @stevenweaver and @davebx

PR #8186, opened 2019-06-18
[19.05] Add hyphy_results.json datatype

rdvelazquez:
Pulling out the datatype addition from https://github.com/galaxyproject/galaxy/pull/7734 as recommended by @mvdbeek.

### DIFF
--- a/config/datatypes_conf.xml.sample
+++ b/config/datatypes_conf.xml.sample
@@ -160,7 +160,7 @@
     <datatype extension="loom" type="galaxy.datatypes.binary:Loom" description="An HDF5-based Loom File" mimetype="application/octet-stream" display_in_upload="true"/>
     <datatype extension="h5ad" type="galaxy.datatypes.binary:Anndata" description="An HDF5-based anndata File" mimetype="application/octet-stream" display_in_upload="true"/>
     <datatype extension="mz5" type="galaxy.datatypes.binary:H5" subclass="true" mimetype="application/octet-stream" display_in_upload="true"/>
-    <datatype extension="hyphy_json_results" type="galaxy.datatypes.text:Json" mimetype="application/json" subclass="true" display_in_upload="false"/>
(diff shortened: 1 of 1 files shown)


config/datatypes_conf.xml.sample
```diff
@@ -160,6 +160,7 @@
     <datatype extension="loom" type="galaxy.datatypes.binary:Loom" description="An HDF5-based Loom File" mimetype="application/octet-stream" display_in_upload="true"/>
     <datatype extension="h5ad" type="galaxy.datatypes.binary:Anndata" description="An HDF5-based anndata File" mimetype="application/octet-stream" display_in_upload="true"/>
     <datatype extension="mz5" type="galaxy.datatypes.binary:H5" subclass="true" mimetype="application/octet-stream" display_in_upload="true"/>
+    <datatype extension="hyphy_json_results" type="galaxy.datatypes.text:Json" mimetype="application/json" subclass="true" display_in_upload="false"/>
```
mvdbeek: what do you think about this:
```suggestion
    <datatype extension="hyphy_results.json" type="galaxy.datatypes.text:Json" mimetype="application/json" subclass="true" display_in_upload="false"/>
```
When users download the file it will then be called `dataset.hyphy_results.json` instead of `dataset.hyphy_json_results `, which might be a little nicer

rdvelazquez: That sounds good!

mvdbeek:
@galaxybot test this

rdvelazquez:
@mvdbeek do you know what's going on with the failing api test?

mvdbeek:
Yes, the test runner didn't shut down cleanly, we fixed that on dev I think. Nothing to worry about.

config/datatypes_conf.xml.sample
mvdbeek: Can you accept the commit then ?

mvdbeek (approved):

davebx (approved):

PR #8187, opened 2019-06-18
Added trigger to dataset table to update history.update_time

Nerdinacan:
This is a database trigger that updates the history table (updated_time) whenever a dataset changes. Currently, modifications to a dataset belonging to a history do not necessarily propagate upwards to the history or history_dataset_association table.

Having an accurate update_time on the history object will allow the client to more intelligently poll for updates. This should save a lot of querying from the client.

Alternative ideas I threw out include:

1. Having a view, or equivalently an ORM relationship object which does a subquery to generate the current update_time via max(update_time) on a 2-table subquery join. This seems like a performance sink since we'll be spamming this update_date every couple of seconds looking for updates until such time as we have a push-tech like a websocket.

2. Sifting through the server-side code to each place we update a dataset and modify it to do the equivalent work of the trigger. This could still work, but on the whole triggers will be more performant and maintaining timestamps is probably the best use of a trigger. (Except maybe audit trails.)

PostgreSQL has a proprietary trigger format that requires the installation of a database function which is referenced by the trigger, so I have split the migration according to postgres/non-postres.

I tested the installation on the default SQLite installation that worked fine. I also attempted to test against MySQL only to find that our migration scripts fail against MySQL way before they even get to this update, so I guess we're not super-concerned about that SQL variant.


lib/galaxy/model/migrate/versions/0155_add_content_update_time.py
```diff
@@ -0,0 +1,105 @@
...
+    build_drop_trigger("UPDATE").execute(bind=migrate_engine)
+    build_drop_trigger("DELETE").execute(bind=migrate_engine)
+
+
+def build_trigger(op):
+    create_trigger_template = """
+        CREATE TRIGGER BEFORE_{operation}_DATASET
+            BEFORE {operation} ON dataset
+            BEGIN
+                update history
```
nsoranzo: Can you use uppercase for the SQL keywords also here and below?

lib/galaxy/model/migrate/versions/0155_add_content_update_time.py
```diff
@@ -0,0 +1,105 @@
...
+    pg_drop_trigger.execute(bind=migrate_engine)
+    build_drop_trigger("INSERT").execute(bind=migrate_engine)
+    build_drop_trigger("UPDATE").execute(bind=migrate_engine)
+    build_drop_trigger("DELETE").execute(bind=migrate_engine)
+
+
+def build_trigger(op):
+    create_trigger_template = """
+        CREATE TRIGGER BEFORE_{operation}_DATASET
+            BEFORE {operation} ON dataset
```
nsoranzo: Why not
```suggestion
            AFTER {operation} ON dataset
```
? It makes more sense to me to update the history timestamp after the change to its content.

lib/galaxy/model/migrate/versions/0155_add_content_update_time.py
```diff
@@ -0,0 +1,105 @@
...
+            update history h
+            set update_time = current_timestamp
+            from history_dataset_association hda
+            where h.id = hda.history_id
+            and hda.dataset_id = NEW.id;
+            return NEW;
+        end;
+        $BODY$;
+        CREATE TRIGGER update_history_update_time
+            BEFORE INSERT OR DELETE OR UPDATE
```
nsoranzo:
```suggestion
            AFTER INSERT OR DELETE OR UPDATE
```
See below for explanation.

lib/galaxy/model/migrate/versions/0155_add_content_update_time.py
```diff
@@ -0,0 +1,105 @@
...
+    when contents have changed. Installs a function and a trigger
+    for postgres, other sql variants only require the trigger def
+    """
+
+    pg_create_trigger = DDL("""
+        CREATE FUNCTION update_history_content_update_time()
+            RETURNS trigger
+            LANGUAGE 'plpgsql'
+        AS $BODY$
+        begin
```
nsoranzo: Can you use uppercase for SQL keywords here and below?

lib/galaxy/model/migrate/versions/0155_add_content_update_time.py
```diff
@@ -0,0 +1,105 @@
...
+            and hda.dataset_id = NEW.id;
+            return NEW;
+        end;
+        $BODY$;
+        CREATE TRIGGER update_history_update_time
+            BEFORE INSERT OR DELETE OR UPDATE
+            ON dataset
+            FOR EACH ROW
+            EXECUTE PROCEDURE update_history_content_update_time();
+    """).execute_if(dialect='postgresql')
```
nsoranzo: Wouldn't be easier (and more consistent with other migrations) to use:
```python
        if migrate_engine.name in ['postgres', 'postgresql']:
            DDL(...).execute(bind=migrate_engine)
        else:
            ...
```
instead of the various `execute_if()`s ?

lib/galaxy/model/migrate/versions/0155_add_content_update_time.py
Nerdinacan: Sure, I'm not picky.

lib/galaxy/model/migrate/versions/0155_add_content_update_time.py
Nerdinacan: Sure, I guess an error could get thrown or something. After might be more appropriate.

lib/galaxy/model/migrate/versions/0155_add_content_update_time.py
Nerdinacan: Ahhh sure, was just using the API I found in the docs, but if you have a preferred way of doing it, I'm easy.

lib/galaxy/model/migrate/versions/0155_add_content_update_time.py
Nerdinacan: Yep.

lib/galaxy/model/migrate/versions/0155_add_content_update_time.py
```diff
@@ -0,0 +1,102 @@
...
+    else:
+        build_drop_trigger('INSERT').execute(bind=migrate_engine)
+        build_drop_trigger('UPDATE').execute(bind=migrate_engine)
+        build_drop_trigger('DELETE').execute(bind=migrate_engine)
+
+
+def build_trigger(op):
+    create_trigger_template = """
+        CREATE TRIGGER AFTER_{operation}_DATASET
+            AFTER {operation} ON dataset
```
nsoranzo:
```suggestion
            AFTER {operation}
            ON dataset
            FOR EACH ROW
```

`FOR EACH ROW` is required by MySQL and optional on SQLite. According to [SQLite docs](https://www.sqlite.org/lang_createtrigger.html), "at this time SQLite supports only FOR EACH ROW triggers, not FOR EACH STATEMENT triggers", so it may be added in the future and being explicit seems safer.

lib/galaxy/model/migrate/versions/0155_add_content_update_time.py
```diff
@@ -0,0 +1,102 @@
...
+            BEGIN
+                UPDATE history h
+                SET update_time = current_timestamp
+                FROM history_dataset_association hda
+                WHERE h.id = hda.history_id
+                AND hda.dataset_id = NEW.id;
+                RETURN NEW;
+            END;
+            $BODY$;
+            CREATE TRIGGER update_history_update_time
```
nsoranzo: I was thinking we may want to establish a trigger naming convention, what about `trigger_TABLE_OPERATION`? For this trigger that would be `trigger_dataset_aiudr`, where `aiudr` stands for "after insert, delete of update for each row".

lib/galaxy/model/migrate/versions/0155_add_content_update_time.py
```diff
@@ -0,0 +1,102 @@
...
+        pg_drop_trigger.execute(bind=migrate_engine)
+    else:
+        build_drop_trigger('INSERT').execute(bind=migrate_engine)
+        build_drop_trigger('UPDATE').execute(bind=migrate_engine)
+        build_drop_trigger('DELETE').execute(bind=migrate_engine)
+
+
+def build_trigger(op):
+    create_trigger_template = """
+        CREATE TRIGGER AFTER_{operation}_DATASET
```
nsoranzo: Here the trigger name could be `trigger_dataset_a{op_initial}r`, where you set `op_initial` to `op.lower()[0]`

nsoranzo (commented):
Thanks for the changes, a couple more suggestions!

nsoranzo:
Ah, I also have a question: what about changes to the `history_dataset_association` table, do we need another trigger to update `history` for those too?

Nerdinacan:
@nsoranzo  I don't think I need a trigger on that many-many table. I'm already hearing some changes, likely through the ORM. It's the changes directly to the referenced dataset which are not propagating upwards.

I'm not against adding more update_time propagation triggers if it turns out that I DO need it in the future though. But that would not solve any immediate problems for me at this time.

lib/galaxy/model/migrate/versions/0155_add_content_update_time.py
Nerdinacan: Sure. I've renamed all the triggers to that format.

lib/galaxy/model/migrate/versions/0155_add_content_update_time.py
Nerdinacan: Okie dokie.

nsoranzo (approved):

Nerdinacan:
@nsoranzo  I may have to eat my words. 

I did, in fact, just run into a situation that might necessitate another trigger on the HDA table (or an equivalent ORM update).

Is there a way to put this PR on "hold" so nobody merges it until I get a chance to look more carefully? Or should I just close it?

Nerdinacan:
Thanks!

nsoranzo:
Maybe the trigger on dataset could update history_dataset_association and another trigger on history_dataset_association could update history?

Let us know when it's ready for review again.

Nerdinacan:
Yeah we could do a full chained propagation like that. @mvdbeek also suggested hooking into sql-alchemy's session flush event as another means of accomplishing this without resorting to triggers. That has the advantage of being in the python code instead of the DB which is probably easier to maintain.

Since we may have to do this in more tables, it might be worth another look. I think either approach is reasonable, but it's probably easier to maintain a few lines of python than a bunch of triggers. I'll fiddle around with it.

Thanks for looking at what I got this far. Mainly I wanted to present the idea and see if anybody had strong objections to this strategy. I really don't care how we update the timestamp so long as it's updated by the time I spam it. :)

nsoranzo:
@Nerdinacan The `downgrade()` in a migration should still undo the changes made in the `upgrade()` (if feasible), but you also need to update `lib/galaxy/model/mapping.py` with the changes from `upgrade()` since for new Galaxy installations the database is not generated via migrations any more, but directly from the definitions in `mapping.py` .
Please let me know if you have doubts.

Nerdinacan:
@nsoranzo I actually am a little confused now. Do you not run the migrations at all on new installations as well? Shouldn't the migrations return the db to the state it was in before the upgrade() ran? If the timestamps were in mappings originally then the state before the upgrade() ran already had timestamps. Isn't the conditional check the correct move? 

And, if the migrations DO NOT run on new installations does that mean the triggers I implemented need to somehow be run at the same time mappings does in a new installation?

martenson:
@Nerdinacan we do not run migration on clean new db since https://github.com/galaxyproject/galaxy/pull/8300

Nerdinacan:
@martenson Ok cool. So I DO want those triggers installed at the same time mappings goes in then. Because otherwise new installations will not receive them.

martenson:
@Nerdinacan That is my understanding also.

Nerdinacan:
@martenson @nsoranzo Ok I'll try to figure out where best to execute the installation.

Nerdinacan:
@nsoranzo @martenson  I think this is actually ready now. Finally!

It's been through a couple of iterations: mostly trying to understand the proper return values for the various types of triggers and how that impacted writes.

I moved the trigger generation code inside the model directory and install it when the rest of the database is created. Hopefully i did that properly. I also added some structure around the creation of the triggers so people can reuse the SQL generation as desired.

It's a pretty messy history for 6 file changes. Let me know if you'd like me to give you a cleaner commit to look at.

lib/galaxy/model/migrate/versions/0162_add_content_update_time.py
```diff
@@ -0,0 +1,47 @@
+"""
+Adds timestamps to hdac table. Adds triggers to dataset, hda, hdac tables
```
mvdbeek:
```suggestion
Adds timestamps to hdca table. Adds triggers to dataset, hda, hdca tables.
```

mvdbeek (approved):
Looks good to me, the current version number right now would be 164. If you think this is good to go can you update the migration and remove the [WIP] in the PR title ?

lib/galaxy/model/migrate/versions/0155_add_content_update_time.py
jdavcs: @Nerdinacan @nsoranzo aren't all these supposed to be `AFTER`? Current code uses `BEFORE`. Unless there's been more discussion and it's been agreed to use `BEFORE`, in which case please ignore this comment.

nsoranzo: @ic4f It was change back to `BEFORE` by @Nerdinacan in commit 335499fc5449dfd7ff38104ac84f5e522220425c , not sure why.

jdavcs: Oh, OK, thanks for checking @nsoranzo

PR #8195, opened 2019-06-19
Generalized Kubernetes Staging (Dependency Resolution or Containers)

jmchilton:
Renamed the Pulsar runner derivative for Kubernetes. It now works using just one container and dependency resolution if docker isn't enabled for that destination. Dependency resolution in the container can be configured using nested configuration in the job conf yaml file (``puslar_app_config: {dependency_resolution: {}}``) but a default is set by the Pulsar client library that configures a Conda environment that is configured inside the container. If a container is configured for the job destination, the two-container co-execution strategy from #7835 is still used (but this PR makes that something of an uninteresting implementation detail).

Original Description:

~Builds on #8152 (un-merged) and #7835 (merged) to implement a Kubernetes runner with Pulsar staging but that works with dependency resolution inside of the Pulsar container instead of a second co-executing container.~

~This requires a Pulsar release and a Pulsar Docker container publication (against this branch of Pulsar https://github.com/galaxyproject/pulsar/pull/195)... but it essentially works currently I think with those changes.~

~The example dependency resolution shown here uses a Conda install setup inside the Pulsar container but the goal longer term I guess would be to setup separate Conda resolvers for that and for CVMFS - so it would use CVMFS if available and would auto install locally inside the container if not.~

~The other longer term goal would be to merge the actual runner implementation between this and #7835 - so that you wouldn't need to care if one or two containers would be used. That detail would just be handled by the Pulsar client based on whether a container is configured or not.~

### DIFF
--- a/test/integration/test_kubernetes_staging.py
+++ b/test/integration/test_kubernetes_staging.py
@@ -31,7 +31,7 @@ runners:
     load: galaxy.jobs.runners.local:LocalJobRunner
     workers: 1
   pulsar_k8s:
-    load: galaxy.jobs.runners.pulsar:PulsarKubernetesCoexecutionJobRunner
+    load: galaxy.jobs.runners.pulsar:PulsarKubernetesJobRunner
     galaxy_url: ${infrastructure_url}
     amqp_url: ${amqp_url}
 
@@ -62,7 +62,7 @@ runners:
     load: galaxy.jobs.runners.local:LocalJobRunner
     workers: 1
   pulsar_k8s:
-    load: galaxy.jobs.runners.pulsar:PulsarKubernetesDependencyResolvingJobRunner
+    load: galaxy.jobs.runners.pulsar:PulsarKubernetesJobRunner
     galaxy_url: ${infrastructure_url}
     amqp_url: ${amqp_url}
 
@@ -74,16 +74,6 @@ execution:
       runner: pulsar_k8s
       pulsar_app_config:
         message_queue_url: '${container_amqp_url}'
-        dependency_resolution:
-          cache: false
-          use: true
-          default_base_path: /pulsar_dependencies
-          cache_dir: /pulsar_dependencies/_cache
-          resolvers:
-          - type: conda
-            auto_init: true
-            auto_install: true
-            prefix: '/pulsar_dependencies/conda'
       env:
         - name: SOME_ENV_VAR
           value: '42'
--- a/test/integration/test_kubernetes_staging.py
+++ b/test/integration/test_kubernetes_staging.py
@@ -17,7 +17,7 @@ import tempfile
 
 from base import integration_util  # noqa: I100,I202
 from base.populators import skip_without_tool
-from .test_containerized_jobs import MulledJobTestCases  # noqa: I201
+from .test_containerized_jobs import EXTENDED_TIMEOUT, MulledJobTestCases  # noqa: I201
 from .test_job_environments import BaseJobEnvironmentIntegrationTestCase  # noqa: I201
 
 TOOL_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir, os.pardir, 'tools'))
@@ -25,8 +25,7 @@ AMQP_URL = os.environ.get("GALAXY_TEST_AMQP_URL", "amqp://guest:guest@localhost:
 GALAXY_TEST_KUBERNETES_INFRASTRUCTURE_HOST = os.environ.get("GALAXY_TEST_KUBERNETES_INFRASTRUCTURE_HOST", "host.docker.internal")
 
 
-def job_config(jobs_directory):
-    job_conf_template = string.Template("""
+CONTAINERIZED_TEMPLATE = """
 runners:
   local:
     load: galaxy.jobs.runners.local:LocalJobRunner
@@ -54,7 +53,50 @@ execution:
 tools:
   - id: upload1
     environment: local_environment
-""")
+"""
+
+
+DEPENDENCY_RESOLUTION_TEMPLATE = """
+runners:
+  local:
+    load: galaxy.jobs.runners.local:LocalJobRunner
+    workers: 1
+  pulsar_k8s:
+    load: galaxy.jobs.runners.pulsar:PulsarKubernetesDependencyResolvingJobRunner
+    galaxy_url: ${infrastructure_url}
+    amqp_url: ${amqp_url}
+
+execution:
+  default: pulsar_k8s_environment
+  environments:
+    pulsar_k8s_environment:
+      k8s_config_path: ${k8s_config_path}
+      runner: pulsar_k8s
+      pulsar_app_config:
+        message_queue_url: '${container_amqp_url}'
+        dependency_resolution:
+          cache: false
+          use: true
+          default_base_path: /pulsar_dependencies
+          cache_dir: /pulsar_dependencies/_cache
+          resolvers:
+          - type: conda
+            auto_init: true
+            auto_install: true
+            prefix: '/pulsar_dependencies/conda'
+      env:
+        - name: SOME_ENV_VAR
+          value: '42'
+    local_environment:
+      runner: local
+tools:
+  - id: upload1
+    environment: local_environment
+"""
+
+
+def job_config(template_str, jobs_directory):
+    job_conf_template = string.Template(template_str)
     port = os.environ.get('GALAXY_TEST_PORT')
     assert port
     infrastructure_url = "http://%s:%s" % (GALAXY_TEST_KUBERNETES_INFRASTRUCTURE_HOST, port)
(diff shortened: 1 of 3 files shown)


nuwang:
Wow, you're on a roll! :-) Trying to keep up with the details of this, but is my understanding of the overall sequence of events still current?
```
1. staging pod fetches data (init container)
2. job pod/mulled pod runs tool (also an init container)
3. staging pod flushes data back (main container)
```
Also, am I correct in assuming that if the staging pod were to use galaxylib to fetch data (provided object_store_conf.xml and other galaxy configs are mapped in), the data can be fetched directly from s3/gcp etc? Or is this what's already happening?

jmchilton:
@nuwang Not quite right.

https://github.com/galaxyproject/galaxy/pull/7835 - used two container pods - neither one was an init-container. One container was Pulsar+a-subset-of-Galaxy and the other the tool container (presumably a biocontainer/mulled container typically). The Pulsar container could use whatever method it wants to stage (see https://github.com/galaxyproject/galaxy/pull/7835#issuecomment-503124764) - the default is to just talk to Galaxy like regular MQ Pulsar would. Getting the direct object store loading of data working is sort of independent of this - I've gotten it working locally but I'm not sure I have a demo ready and I haven't yet gotten it working with Kubernetes. The WIP on that is on the Galaxy side for directly loading data from Pulsar is in (https://github.com/galaxyproject/galaxy/pull/7924) and the Pulsar side has already been merged (https://github.com/galaxyproject/pulsar/pull/191/commits/d130b11b095f8354e9b0e81d488e63355d58d57a).

> am I correct in assuming that if the staging pod were to use galaxylib to fetch data (provided object_store_conf.xml and other galaxy configs are mapped in), the data can be fetched directly from s3/gcp etc? Or is this what's already happening?

In theory the pieces are all here once these branches are stitched together properly. Let me simplify the configuration, get the branches merged, do some Pulsar releases - and then you could folks can try 😄 - one hiccup might be if the bucket object store serialized format doesn't have everything in it that it needs for instance.

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -317,6 +317,10 @@ mapping:
           used if you experience long waiting times before a job is actually submitted
           to your cluster.
 
+          This only affects tools where some requirements can be resolved but not others,
+          most modern best practice tools can use prebuilt environments in the Conda
+          directory.
```
mvdbeek: 💯 -- and we should probably announce deprecation. This was a hack when we didn't have the all-in-one resolution yet. In the context of conda this probably makes things worse if it is actually used.

jmchilton:
@mvdbeek you had mentioned some of these tests are broken? I've been testing the Docker and Singularity versions of these and I think they are broken also?

mvdbeek:
The tests weren't broken when I checked out this PR (except for the new resolution), but they are now indeed. Somewhere a wrong path to involucro (`database/none/involucro`) creeps in on current dev (including https://github.com/galaxyproject/galaxy/pull/8355), I'm looking into it.

mvdbeek:
@galaxybot test this

martenson:
Tests seem to pass, but now this is in dependency-conflict. I assume if that is resolved we can merge.

mvdbeek:
@martenson did you test this locally ? Kubernetes tests don't run on Jenkins. I'll have a look later if you haven't done this already.

martenson:
@mvdbeek I did not test this.

jmchilton:
It is possible the tests don't work perfectly - but I think this should still be merged if it isn't breaking CI. This is an important upgrade to that Kubernetes functionality and we don't want the cloud team targeting the old way of configuring this stuff

mvdbeek:
But it does break previously working tests for me, on dev I get:
```
=============================== 3 passed, 83 warnings in 50.30 seconds ================================
```
On this PR (rebased against dev):
```
__________________ KubernetesStagingContainerIntegrationTestCase.test_mulled_simple ___________________

self = <integration.test_kubernetes_staging.KubernetesStagingContainerIntegrationTestCase testMethod=test_mulled_simple>

    def test_mulled_simple(self):
        self.dataset_populator.run_tool("mulled_example_simple", {}, self.history_id)
        self.dataset_populator.wait_for_history(self.history_id, assert_ok=True)
        output = self.dataset_populator.get_history_dataset_content(self.history_id, timeout=EXTENDED_TIMEOUT)
>       assert "0.7.15-r1140" in output
E       AssertionError: assert '0.7.15-r1140' in '/pulsar_staging/3/command.sh: line 92: bwa: command not found\n'

test/integration/test_containerized_jobs.py:31: AssertionError
___________________ KubernetesDependencyResolutionIntegrationTestCase.test_explicit ___________________

self = <integration.test_kubernetes_staging.KubernetesDependencyResolutionIntegrationTestCase testMethod=test_explicit>

    def test_explicit(self):
        self.dataset_populator.run_tool("mulled_example_explicit", {}, self.history_id)
        self.dataset_populator.wait_for_history(self.history_id, assert_ok=True)
        output = self.dataset_populator.get_history_dataset_content(self.history_id, timeout=EXTENDED_TIMEOUT)
>       assert "0.7.15-r1140" in output
E       AssertionError: assert '0.7.15-r1140' in '/pulsar_staging/1/command.sh: line 92: bwa: command not found\n'

test/integration/test_containerized_jobs.py:25: AssertionError
________________ KubernetesDependencyResolutionIntegrationTestCase.test_mulled_simple _________________

self = <integration.test_kubernetes_staging.KubernetesDependencyResolutionIntegrationTestCase testMethod=test_mulled_simple>

    def test_mulled_simple(self):
        self.dataset_populator.run_tool("mulled_example_simple", {}, self.history_id)
        self.dataset_populator.wait_for_history(self.history_id, assert_ok=True)
        output = self.dataset_populator.get_history_dataset_content(self.history_id, timeout=EXTENDED_TIMEOUT)
>       assert "0.7.15-r1140" in output
E       AssertionError: assert '0.7.15-r1140' in '/pulsar_staging/2/command.sh: line 92: bwa: command not found\n'

test/integration/test_kubernetes_staging.py:164: AssertionError
```

mvdbeek:
OK, with pulsar from master it looks better and it's just the mulled simple test that fails:
```
============================================== FAILURES ===============================================
__________________ KubernetesStagingContainerIntegrationTestCase.test_mulled_simple ___________________

self = <integration.test_kubernetes_staging.KubernetesStagingContainerIntegrationTestCase testMethod=test_mulled_simple>

    def test_mulled_simple(self):
        self.dataset_populator.run_tool("mulled_example_simple", {}, self.history_id)
>       self.dataset_populator.wait_for_history(self.history_id, assert_ok=True)

test/integration/test_containerized_jobs.py:29:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
test/base/populators.py:208: in wait_for_history
    return wait_on_state(lambda: self._get("histories/%s" % history_id), desc="history state", assert_ok=assert_ok, timeout=timeout)
test/base/populators.py:1429: in wait_on_state
    return wait_on(get_state, desc=desc, timeout=timeout)
test/base/populators.py:1518: in wait_on
    value = function()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

    def get_state():
        response = state_func()
        assert response.status_code == 200, "Failed to fetch state update while waiting. [%s]" % response.content
        state = response.json()["state"]
        if state in skip_states:
            return None
        else:
            if assert_ok:
>               assert state == "ok", "Final state - %s - not okay." % state
E               AssertionError: Final state - error - not okay.

test/base/populators.py:1426: AssertionError
```

test/integration/test_kubernetes_staging.py
```diff
@@ -73,35 +105,65 @@ def job_config(jobs_directory):
...
         config["jobs_directory"] = cls.jobs_directory
         config["file_path"] = cls.jobs_directory
-        config["job_config_file"] = job_config(cls.jobs_directory)
-
+        config["job_config_file"] = job_config(CONTAINERIZED_TEMPLATE, cls.jobs_directory)
         config["default_job_shell"] = '/bin/sh'
-        # Disable tool dependency resolution.
+        # Disable local tool dependency resolution.
         config["tool_dependency_dir"] = "none"
-        config["enable_beta_mulled_containers"] = "true"
```
mvdbeek: Ah, that's why the coexecution test fails, this is still needed.

mvdbeek (approved):
Looks good and works fine with `enable_beta_mulled_containers` in `KubernetesStagingContainerIntegrationTestCase `.

test/integration/test_kubernetes_staging.py
jmchilton: I'm not wrong... Galaxy is wrong https://github.com/galaxyproject/galaxy/pull/8569 😆. I don't know why I dropped this probably was testing some specific test and wanted to isolate the problem. I have been meaning to do #8569 for some time though.

mvdbeek: Haha, OK, can you rebase this against dev or #8569? I'll merge this then.

jmchilton: Once the tests pass for #8569 I'll merge that and then rebase this.

jmchilton: Merge now? Rightly Martin doesn't like it when I merge a PR I've force pushed since the last approve 😅.

test/integration/test_kubernetes_staging.py
```diff
@@ -101,6 +132,38 @@ def test_job_environment(self):
...
+    @classmethod
+    def handle_galaxy_config_kwds(cls, config):
+        config["jobs_directory"] = cls.jobs_directory
+        config["file_path"] = cls.jobs_directory
+        config["job_config_file"] = job_config(DEPENDENCY_RESOLUTION_TEMPLATE, cls.jobs_directory)
+
+        config["default_job_shell"] = '/bin/sh'
+        # Disable tool dependency resolution.
+        config["tool_dependency_dir"] = "none"
+        config["enable_beta_mulled_containers"] = "true"
```
mvdbeek, line 158: That isn't necessary anymore, right ?

mvdbeek (approved):
Still good :)

test/integration/test_kubernetes_staging.py
jmchilton: Will fix it in a new PR 😆 - thanks @mvdbeek

Issue #8196, opened 2019-06-19
Trim whitespace in URL upload box?


afgane:
If there is a leading whitespace in the data upload form, the files won't upload. From the user perspective, seems like a whitespace trim would be an acceptable automatic operation to perform by the UI?

![Screen Shot 2019-06-19 at 1 41 25 PM](https://user-images.githubusercontent.com/338642/59790929-03a34d80-929f-11e9-8e58-1fd659790d69.png)

martenson:
This is definitely a bug, I think we fixed it in the past but maybe it resurfaced.

jdavcs:
The blank line above the edited if-statement is unintentional. If this solution works, I'll remove it.

afgane:
This bug has resurfaced:
![image](https://user-images.githubusercontent.com/338642/72851176-ee4d3200-3c78-11ea-88fd-7c7b6e5109e0.png)

This is on 20.01.

martenson:
@ic4f are you still up for solving this? I am unclear about why #8197 was closed.

ic4f:
@martenson absolutely! I believe there was a security issue involved that was taken care of separately.  I think @nsoranzo brought it up (sorry if I'm mistaken - it was almost a year ago, so I may be forgetting stuff). Since the whitespace is not resolved, yes, sure, I'll take care of it. However, I'm off until next Thursday, so I'll get to it week after next.

PR #8197, opened 2019-06-19
Trim leading whitespace when uploading from url

jdavcs:
Solves this: https://github.com/galaxyproject/galaxy/issues/8196

### DIFF
--- a/lib/galaxy/tools/actions/upload_common.py
+++ b/lib/galaxy/tools/actions/upload_common.py
@@ -121,18 +121,23 @@ def persist_uploads(params, trans):
             elif type(f) == dict and 'local_filename' not in f:
                 raise Exception('Uploaded file was encoded in a way not understood by Galaxy.')
 
-            if 'url_paste' in upload_dataset and upload_dataset['url_paste']:
-                upload_dataset['url_paste'] = upload_dataset['url_paste'].strip()
-                if upload_dataset['url_paste']:
-                    upload_dataset['url_paste'] = datatypes.sniff.stream_to_file(
-                        StringIO(validate_url(upload_dataset['url_paste'], trans.app.config.fetch_url_whitelist_ips)),
-                        prefix="strio_url_paste_"
-                    )
-            else:
-                upload_dataset['url_paste'] = None
+            upload_dataset['url_paste'] = _get_value_or_none(upload_dataset, 'url_paste')
+            if upload_dataset['url_paste']:
+                upload_dataset['url_paste'] = datatypes.sniff.stream_to_file(
+                    StringIO(validate_url(upload_dataset['url_paste'], trans.app.config.fetch_url_whitelist_ips)),
+                    prefix="strio_url_paste_"
+                )
+
             new_files.append(upload_dataset)
         params['files'] = new_files
     return params
+
+
+def _get_value_or_none(d, key):
+    val = None
+    if key in d and d[key]:
+        val = d[key].strip()
+    return val or None
 
 
 def handle_library_params(trans, params, folder_id, replace_dataset=None):
--- a/lib/galaxy/tools/actions/upload_common.py
+++ b/lib/galaxy/tools/actions/upload_common.py
@@ -26,6 +26,8 @@ def validate_url(url, ip_whitelist):
     if not (url.lstrip().startswith('http://') or url.lstrip().startswith('https://')):
         return url
 
+    # Strip leading whitespace
+    url = url.lstrip()
     # Extract hostname component
     parsed_url = urlparse(url).netloc
     # If credentials are in this URL, we need to strip those.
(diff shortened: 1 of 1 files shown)


lib/galaxy/tools/actions/upload_common.py
```diff
@@ -120,11 +120,14 @@ def persist_uploads(params, trans):
                                                    local_filename=local_filename)
             elif type(f) == dict and 'local_filename' not in f:
                 raise Exception('Uploaded file was encoded in a way not understood by Galaxy.')
-            if 'url_paste' in upload_dataset and upload_dataset['url_paste'] and upload_dataset['url_paste'].strip() != '':
-                upload_dataset['url_paste'] = datatypes.sniff.stream_to_file(
-                    StringIO(validate_url(upload_dataset['url_paste'], trans.app.config.fetch_url_whitelist_ips)),
-                    prefix="strio_url_paste_"
-                )
+
+            if 'url_paste' in upload_dataset and upload_dataset['url_paste']:
```
martenson:
```suggestion
            if 'url_paste' in upload_dataset:
```

jdavcs: I think we need the second condition: 'url_paste' in dict tests the presence of the key; the value still may be None - in which case calling strip() on the value will raise an attributeerror, no?

lib/galaxy/tools/actions/upload_common.py
```diff
@@ -120,11 +120,14 @@ def persist_uploads(params, trans):
                                                    local_filename=local_filename)
             elif type(f) == dict and 'local_filename' not in f:
                 raise Exception('Uploaded file was encoded in a way not understood by Galaxy.')
-            if 'url_paste' in upload_dataset and upload_dataset['url_paste'] and upload_dataset['url_paste'].strip() != '':
-                upload_dataset['url_paste'] = datatypes.sniff.stream_to_file(
-                    StringIO(validate_url(upload_dataset['url_paste'], trans.app.config.fetch_url_whitelist_ips)),
```
martenson: Maybe just adding a `strip()` to this line instead could work. So the `else` branch is triggered the same as before

lib/galaxy/tools/actions/upload_common.py
martenson: True, could be the case of None.

lib/galaxy/tools/actions/upload_common.py
jdavcs: We can, of course. Very minor issue though, IMHO: we're calling strip() twice. It doesn't matter for performance or readability, but if down the road we need to do something else to that value, we'll have to remember to call strip (a third time). If the pasted url should have no whitespace, I'd trim it before doing any kind of processing. But that's a very minor consideration; besides, there's the added line as a downside.. Up to you.

martenson: In your current implementation if you get `"    "` you return `""`, in the previous one you'll return `None`. Not sure how important it is in this case but upload can to be pretty finicky.

jdavcs: sounds very important! checking....

jdavcs: @martenson what do you think of this approach? I split it in two: (a) getting the value if it exists, and (b) processing that value. I think this is slightly more readable, but I'm torn. If you think it's too much, let's go with what you suggested - just calling strip() again.
But if this does work, i'll add a couple unit tests.

martenson: Are you still sure you'd choose it over this?

```diff
-                    StringIO(validate_url(upload_dataset['url_paste'], trans.app.config.fetch_url_whitelist_ips)),
+                    StringIO(validate_url(upload_dataset['url_paste'].strip(), trans.app.config.fetch_url_whitelist_ips)),
```

martenson: > if down the road we need to do something else to that value

That url value disappears in either case, it is either None, or a file handle.

jdavcs: It does disappear.. And there's really no place we would reuse that function in this file. So yes, I think you're right, let's keep it simple. Will submit in a min.

jdavcs: Done.

nsoranzo:
2 API tests are failing.

nsoranzo:
I fear that's not the correct approach, what if the user wants to upload some text (not a list of URLs) which starts on purpose with some spacing?

jdavcs:
@nsoranzo - yes. But then, I'm wondering, wouldn't it it be better to treat data (to upload) and URLs (to fetch data from) differently? Those are very different processes, and I think the system should distinguish between the two as early as possible. How to process the CR character should be a consideration for the pasted or fetched data, but, I think, it should have no bearing on how we treat a list of URLs. In other words, first get the data (pasted input or fetch from pasted URLs), and only then process it. 
So, what if we decouple fetching data from processing data? As one of the benefits, trimming surrounding whitespace or doing any other URL-related processing would become so easy.

martenson:
Could be enough to move the lstrip to the `validate_url` method.

jdavcs:
@martenson - yep, will do right away - should solve the immediate problem, but I'm curious what you think about my other suggestion.

lib/galaxy/tools/actions/upload_common.py
```diff
@@ -26,6 +26,8 @@ def validate_url(url, ip_whitelist):
     if not (url.lstrip().startswith('http://') or url.lstrip().startswith('https://')):
         return url
 
+    # Strip leading whitespace
+    url = url.lstrip()
```
nuwang, line 30: should this be just `url = url.strip()`, since whitespace on either side is invalid?

jdavcs:
@nuwang I try to minimize the footprint here. Stripping on both ends in a previous iteration led to 2 api tests failing (when testing for POSIX standard on/off): the carriage return character ended up being stripped off, while the test expected it to be preserved. This may or may not be relevant for this iteration, since we modify a local variable.. But I didn't want any surprises, so I changed the minimum.

PR #8211, opened 2019-06-21
new regular expression filter, and efficient sort filter

bernt-matthias:
Main part of the PR: 

- New filter to filter by regular expressions (including test)

I had the following use case. The mothur aligndb data table mixes together aligned and unaligned sequences (some of the tools can use both and some accept only aligned sequences). The file names in the data table can be categorized by the file extension. So I could add new data tables or add a column to the table, but both options are not backward compatible and require the admins to update data tables. This fix would allow to stick with the old data tables.

While checking the sources I also added: 

- use likely more efficient python `sorted` implementation for the sort filter (before it was insertion sort which is O(n^2); O(n^3) if the insertions take linear time).
- more detailed documentation of the filter types
- attribute_value_splitter has been marked deprecated (could not find a single tool using this, originally added here https://github.com/galaxyproject/galaxy/commit/f72c32dcf7a105cb81aacee49189bd6503f79243 for gff_filtering, but this tool seems not to use this anymore, also the used attribute `name_val_separator` is neither documented nor supported by the xsd)

### DIFF
--- a/lib/galaxy/tool_util/xsd/galaxy.xsd
+++ b/lib/galaxy/tool_util/xsd/galaxy.xsd
@@ -3622,8 +3622,9 @@ These values are defined in the module
 [/lib/galaxy/tools/parameters/dynamic_options.py](https://github.com/galaxyproject/galaxy/blob/master/lib/galaxy/tools/parameters/dynamic_options.py)
 in the ``filter_types`` dictionary. 
 
+Deprecated filter types:
 
-* ``attribute_value_splitter``: 
+* ``attribute_value_splitter`` 
         ]]></xs:documentation>
       </xs:annotation>
     </xs:attribute>
--- a/lib/galaxy/tool_util/xsd/galaxy.xsd
+++ b/lib/galaxy/tool_util/xsd/galaxy.xsd
@@ -3599,12 +3599,24 @@ demonstrates splitting up strings into multiple values.
     <xs:attribute name="type" type="FilterType" use="required">
       <xs:annotation>
         <xs:documentation xml:lang="en"><![CDATA[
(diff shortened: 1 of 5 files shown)


bernt-matthias:
Will change this to a separate filter, since it's actually not backward compatible (eg if the value is a regexp) and also because "static" suggests something different. Thanks to @jmchilton for the comment.

bernt-matthias:
Finished from my side. 

But I do not completely understand why `expect_failure` works here:

https://github.com/galaxyproject/galaxy/blob/82acc37806a642b62ab311ed64efca1d29d4bb13/test/functional/tools/filter_static_regexp.xml#L62

Discussed previously with @mvdbeek that this should only indicate job execution failure:

https://github.com/galaxyproject/galaxy/pull/7730#issuecomment-516428878

At least that's how I understood it.

bernt-matthias:
Fancy ideas for an extension that I find really appealing: 

- allow search and replace by regular expression: add a new attribute to the filter (eg replace). For instance `<filter value="ID=(.*)" replace="\1">`

- if we don't want the extra attribute then we could just check in the search if there are any groups in the match object and concatenate them with the value of the separator attribute

- one could also imagine that the regexp.split function might be useful as a more powerful replacement of the multipe_value_splitter.

Would like to have some comments before starting to implement anything.

bernt-matthias:
Some of the other filters allow multiple columns, might be also an option here, but this might as well be implemented in the tool using multiple filters.

Either 

`<filter column="x,y,z">`

or 

```
<filter column="x">
<filter column="y">
<filter column="z">
```

guess this is equivalent

PR #8216, opened 2019-06-24
[19.05] Prefer prebuilt node when creating envs with nodeenv

mvdbeek:
From the manual:
```
--prebuilt
Install node.js from prebuilt package
```

That seems appropriate and helpfully prevents some travis timeouts
on tools-iuc / tools-devteam repos.


mvdbeek:
Meh, there seem to be some connectivity issues, this probably doesn't help.

nsoranzo:
@mvdbeek ~Maybe this could help? https://docs.travis-ci.com/user/caching/#yarn-cache~

Edit: Sorry, this doesn't help for tool testing, where we only install nodeenv but not the NPM packages.

PR #8220, opened 2019-06-25
Run make config-rebuild to update config/docs

jdavcs:
This is to update galaxy.yml.sample and galaxy_options.rst to be consistent with the latest update to config_schema.yml (PRs https://github.com/galaxyproject/galaxy/pull/7897 and https://github.com/galaxyproject/galaxy/pull/8118)

(generated with `make config-rebuild`)

NOTE: I assume we run this each time we update config_schema.yml? I saw this while rebuilding with my own updates, and I didn't want to add this older update to an unrelated PR.

### DIFF
--- a/config/galaxy.yml.sample
+++ b/config/galaxy.yml.sample
@@ -172,7 +172,7 @@ galaxy:
 
   # Time to sleep between attempts if database_wait is enabled (in
   # seconds).
-  #database_wait_sleep: 1
+  #database_wait_sleep: 1.0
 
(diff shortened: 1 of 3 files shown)


hexylena (approved):
> NOTE: I assume we run this each time we update config_schema.yml? I

Correct, not sure where the documentation is for that, perhaps we should add a comment to config_schema.yaml

jdavcs:
Thanks @erasche . And yes, added a comment to config_schema.yml.

PR #8224, opened 2019-06-25
Add nodejs dependency to parse_values_from_file

mvdbeek:
This is working nicely as we preserve the Galaxy environment for non-standard tool classes anyway. That said we could also hard code a nodejs dependency in the ExpressionTool class (might be nicer for expression tools defined within workflows ?).

### DIFF
--- a/tools/expression_tools/parse_values_from_file.xml
+++ b/tools/expression_tools/parse_values_from_file.xml
@@ -1,5 +1,8 @@
 <tool name="Parse parameter value" id="param_value_from_file" version="0.1.0" tool_type="expression">
     <description>from dataset</description>
+    <requirements>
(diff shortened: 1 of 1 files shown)


hexylena:
xref discussion https://github.com/galaxyproject/training-material/pull/1393#issuecomment-505310395

this looks like an improvement for us at least for sure

jmchilton:
This isn't right - the tool is declaring its runtime - it shouldn't redeclare requirements like that explicitly. If you want to use tool requirements - update the parser to supply them for expression tools?

mvdbeek:
> This isn't right - the tool is declaring its runtime - it shouldn't redeclare requirements like that explicitly.

Agreed, that's probably the better way.

PR #8226, opened 2019-06-25
Update examples for pytest test selector syntax

mvdbeek:


mvdbeek:
The old syntax does still work, as pointed out by @nsoranzo. We do link to the pytest selects in the text, so it should be fine as is.

PR #8227, opened 2019-06-25
hyphy-vision

stevenweaver:
* Adds a visualization for results from hyphy-tools.
* Currently live at https://galaxy.hyphy.org

I think the idea is to eventually add the hyphy-tools to usegalaxy.org; getting this viz in is an important step; there are also some minor revision still to be done to the tools themselves (that will be tracked at http://github.com/galaxyproject/tools-iuc)

56064370-dfc40880-5d3f-11e9-8aaa-871e305d53dd


stevenweaver:
I'm going to guess that the reservations with accepting this PR has to do with including hyphy-vision source files into the galaxy repo as opposed to including hyphy-vision as a dependency in a file such as `./client/package.json`. 

Would the galaxy team feel more comfortable with this PR if we placed the dependency in `./client/package.json` or a new package.json file?

The @veg team followed what I had done for the hivtrace visualization. The [hivtrace](https://github.com/galaxyproject/galaxy/tree/dev/config/plugins/visualizations/hivtrace) visualization was written at a time when the package.json file and accompanying yarn build steps were still not quite ready to go. 

I have a meeting that includes @nekrut and @davebx on Tuesday, and it is important to have a status update regarding our outstanding PRs before then.

Best,
Steven

dannon:
@stevenweaver You're right that this isn't how we will want to ship visualization dependencies in the future, but I wouldn't hold up merge based on that if it's important to get this in right now.  Down the road, `./client/package.json` won't be the destination for dependencies (we're intentionally separating pluggable components like this from the primary galaxy client's dependencies), but it's on the right path. What we're moving towards is adding a dependencies definition to the viz config (`hyphyvision.xml` here) that'll specify a yarn/npm-compatible module string (this could be published on npm, or it could just be a github repo) that Galaxy can install and build automatically.  So what you'd actually commit upstream or share with others would simply be that single visualization xml file and, in this case, the additional datatype specification.

We didn't want to upend viz things right before GCC so some of that got put on hold for a bit, but I'll talk to @guerler and figure out a concrete plan for our next steps -- in addition to what I've summarized there are a bunch of new ideas are floating around right now for visualizations and I'd like to get them organized and make a single pass updating our viz spec that we can move forward with soon.  I'll also make an actual test pass over this tomorrow and verify it's good to go should we want to just merge it mostly as-is and follow up later.

davebx (approved):

bgruening:
I think this needs again a rebase :(

davebx:
@bgruening agreed. I'll take care of it.

davebx:
@bgruening now _this_ looks a lot nicer, doesn't it?

config/datatypes_conf.xml.sample
```diff
@@ -1 +0,0 @@
-../lib/galaxy/config/sample/datatypes_conf.xml.sample
```
bgruening: this looks wrong ... the link is the new way ... the real file should be removed.

config/plugins/visualizations/hyphyvision/src/.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
bgruening: I guess this can be part of the main gitignore?

config/datatypes_conf.xml.sample
davebx: Ah, my mistake.

bgruening (commented):
I'm not sure if the static content needs to be part of this PR, it is strange since you have the package.json.

@guerler can probably help here.

config/plugins/visualizations/hyphyvision/src/.gitignore
davebx: Interesting, looks like it already is.

```
{base}dave@gaius:~/github-repos/hyphy-vision-rebase $ grep node_modules .gitignore 
# JS, Local node_modules, and bower_components directories
node_modules
```

davebx:
Thanks for looking at this, Björn. The rebase must have messed up a few things.

stevenweaver:
@bgruening - Please see @dannon's comments earlier in this PR regarding static content.

config/plugins/visualizations/hyphyvision/config/hyphyvision.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE visualization SYSTEM "../visualization.dtd">
+<visualization name="HyPhy result viewer">
+  <data_sources>
+    <data_source>
+      <model_class>HistoryDatasetAssociation</model_class>
+      <test test_attr="ext" result_type="datatype">hyphy_json_results</test>
```
martenson: this needs to match the datatype which is `hyphy_results.json`

martenson (commented):
This seems to work well. I'd like to see the css minified and the resulting commit of this PR squashed into one. Let me know if you want me to do this.

We can get this to 19.09 if we address the above without delay.

Thank you for the contribution!

cc @stevenweaver

martenson:
The css minification actually does not save us much, roughly 20%, so not crucial.

stevenweaver:
@martenson looks great and thank you. Let me know if you need anything more from me.

stevenweaver:
Thank you @martenson and @davebx!

PR #8228, opened 2019-06-25
Fix index out of bounds exception when a dataset is empty or missing columns

innovate-invent:
If columns are referenced but not present, logic already exists to exclude them.

### DIFF
--- a/tools/stats/filtering.py
+++ b/tools/stats/filtering.py
@@ -186,7 +186,7 @@ if not check_expression(cond_text):
 # Work out which columns are used in the filter (save using 1 based counting)
 used_cols = sorted(set(int(match.group()[1:])
                    for match in re.finditer(r'c(\d)+', cond_text)))
-largest_col_index = max(used_cols)
+largest_col_index = min(in_columns, max(used_cols))
 
 # Prepare the column variable names and wrappers for column data types. Only
 # cast columns used in the filter.


mvdbeek (dismissed):
Thanks @innovate-invent!

tools/stats/filtering.py
```diff
@@ -186,7 +186,7 @@ def stop_err(msg):
 # Work out which columns are used in the filter (save using 1 based counting)
 used_cols = sorted(set(int(match.group()[1:])
                    for match in re.finditer(r'c(\d)+', cond_text)))
-largest_col_index = max(used_cols)
+largest_col_index = min(in_columns, max(used_cols))
```
mvdbeek, line 189: This will lead to everything being invalid lines (but a green dataset) since the filter condition can never be fulfilled. I'm not sure that's always the correct thing to do. It seems like we should error out by default when this happens (in most cases I assume this is not intended), with a tool option to output an empty file if that's a legitimate outcome. Does that seem reasonable ?

innovate-invent: The configuration options should be something like:
Yes/No: Allow empty datasets
Select:
- All expected columns must be present
- Filter records with fewer than expected columns
- Output records with fewer than expected columns

mvdbeek: Yep, that sounds good to me.

PR #8229, opened 2019-06-25
Kallistobus

astrovsky01:
Add in datatypes for kallisto bus and bustools

### DIFF
--- a/config/datatypes_conf.xml.sample
+++ b/config/datatypes_conf.xml.sample
@@ -378,8 +378,8 @@
       <display file="igv/vcf.xml"/>
       <converter file="vcf_bgzip_to_tabix_converter.xml" target_datatype="tabix"/>
     </datatype>
-    <datatype extension="ec" type="galaxy.datatypes.binary:Binary" subclass="true" display_in_upload="true" />
-    <datatype extension="idx" type="galaxy.datatypes.binary:Binary" subclass="true" display_in_upload="true" />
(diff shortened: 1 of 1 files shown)


config/datatypes_conf.xml.sample
```diff
@@ -367,6 +367,8 @@
       <display file="igv/vcf.xml"/>
       <converter file="vcf_bgzip_to_tabix_converter.xml" target_datatype="tabix"/>
     </datatype>
+    <datatype extension="ec" type="galaxy.datatypes.binary:Binary" subclass="true" display_in_upload="true" />
```
mvdbeek: This and idx are very short extensions -- is there maybe an alternative to it ?
`idx` also seems very general

mvdbeek (commented):
Could you maybe drop the merge commits from this branch ?

config/datatypes_conf.xml.sample
astrovsky01: idx is the output filetype of part of kallisto tools and input for another. I could add hidden_in_upload?

astrovsky01:
@mvdbeek how do I do that?

config/datatypes_conf.xml.sample
mvdbeek: That's not my concern, the extensions need to be unique across Galaxy and users should be able to tell what file it is based on the extension. How about kallisto.ec and kallisto.idx ?

astrovsky01: Ah. I can do that.

mvdbeek (approved):
Thanks a lot, looks good to me.

Issue #8072, opened 2019-05-31
Dynamic Watching and Reloading of a subset of config.py


jmchilton:
Break ``config.py`` into a subset of variables that can safely be reloaded after app start -``message_box_content``, ``admin_users``, ``cleanup_job``, and ``welcome_url``. A monitor task should be added to monitor the config file for changes and reloading these.

Potentially relevant xrefs:
- https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/config_watchers.py
- https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config.py
- https://github.com/galaxyproject/galaxy/pull/4617/files
- https://github.com/galaxyproject/galaxy/pull/6215

For the variables we that are updated to do this - we should make sure they aren't cached on objects (``self.admin_users = app.config.admin_users`` - they always need to be accessed through the ``config`` object - or some cool proxy thereof).

---

Discussion:

@natefoo Any chance I could convince you to add to the "main tasks" making a dynamically changeable masthead banner?

@jmchilton
Next week of course and certainly!
How do you want it to work?

@natefoo
Either watching (via fs events) a file on disk or something in the UI
(open to suggestions)

@jmchilton
All the plumbing for a masthead is already there right? We just need to change it? We could do that

@natefoo
Yeah, currently it's done via the config file though, so changing it requires a restart

@jmchilton
``message_box_content``
Alright. I'll sketch something out and assign it, perfect

@natefoo
Awesome sauce

martenson:
could we please add tool panel search related configs?
```
  #tool_name_boost: 9.0
  #tool_section_boost: 3.0
  #tool_description_boost: 2.0
  #tool_label_boost: 1.0
  #tool_stub_boost: 5.0
  #tool_help_boost: 0.5
  #tool_search_limit: 20
  #tool_enable_ngram_search: false
  #tool_ngram_minsize: 3
  #tool_ngram_maxsize: 4
```

jmchilton:
@martenson That requires rebuilding the search index though I think, I like the idea but it requires more plumbing to issue a callback on changed values or something - so I'd prefer to defer that to an iteration 2.

martenson:
@jmchilton boosts are attributes of the search instance, not the index creation

jmchilton:
Sounds easy enough then, thanks @martenson !

jdavcs:
[...]

PR #921, opened 2015-10-15
Make Galaxy (almost) packageable/installable

natefoo:
# 2019 Update

This won't be entirely ready for `pip install galaxy` but does what's needed to quickly finish it up. The plan is for the [galaxy-web-apps package](https://github.com/galaxyproject/galaxy/tree/dev/packages/web_apps/) package essentially a full install of Galaxy with its web stack, and the [galaxy metapackage](https://github.com/galaxyproject/galaxy/tree/dev/packages/meta/) will be pinned to stable releases. Because the packagification of Galaxy began just before updating this PR happened, the method originally used here (one big package) is no longer valid, and the metapackage isn't working yet. That will happen in a subsequent PR.

This PR includes some important useful pieces:
- `GalaxyConfiguration.config_dir` is `dirname(config_file)` (or `.` if there is no config file)
- `GalaxyConfiguration.data_dir` is `./data`
- No assumption that `cwd` is `galaxy_root`
- No assumption that configs are in `./config/`
- Sample configs are no longer required
- Galaxy now finds templates, static, and tools via symlinks in the galaxy lib dir (but these are not packageable yet)
- Sample configs moved to `lib/galaxy/config/sample`

The exception to the changes above: if running from `cwd = galaxy_root`, then `config_dir` and `data_dir` are `<cwd>/config` and `<cwd>/database` as before.

## Updated TODO list for the next PR
- [ ] Finish the metapackage
- [ ] Update `galaxy-config` for YAML configs
- [ ] Add `galaxy-uwsgi` or `galaxy -m uwsgi`
- [ ] Package `static/`
- [ ] Other directories: `external_services/`, `tool-data/`, ...
- [ ] Create a `galaxy-set-metadata` command to replace the tempfile method currently used
- [ ] Move `galaxy-manage-db` to galaxy-app after #8293 I think
- [ ] Fix tool migrations
- [ ] Other things I'm not thinking of

# Original PR

Building on #428, this attempts to make Galaxy installable via the standard Pythonic mechanism. To use:
1. Make a virtualenv and activate it
2. `PYTHONPATH= pip install --upgrade pip wheel`
3. `make wheels`
4. `PYTHONPATH= pip install -i https://wheels.galaxyproject.org/simple/ 'dist/galaxy-16.7.dev0-py2-none-any.whl[postgresql]'`
5. Create a directory somewhere, copy `config/galaxy.ini.sample` to it as `galaxy.ini`
6. `galaxy-paster serve galaxy.ini`

Alternatively:
[...]

PR #8230, opened 2019-06-26
[WIP] Add dynamic watching and reloading of a subset of config.py

jdavcs:
This addresses this: https://github.com/galaxyproject/galaxy/issues/8072

Requesting a (brief) preliminary review.

Summary of approach:
1. Add 2 options to galaxy.yml: one turns reloading on/off; the other lists the options to be reloaded
2. We don't want users to be able to select arbitrary options for reloading. The list of options that are safe to reload is located in config.py: users may choose a subset of these options (any others will be ignored)
3. I add a watcher to config_watchers.py for galaxy.yml (leaning heavily on @natefoo 's implementation)
4. The reloading happens in config.py.

Ping @jmchilton . Is this a reasonable approach? Also, can we assume the core config is a yaml file? (if not, I'll add support for ini format). And is this a good way to get its path? https://github.com/ic4f/galaxy/blob/dev_dynamic_config/lib/galaxy/config.py#L946

If this approach is reasonable, I'll add tests before submitting final version. 
Thanks!

### DIFF
--- a/lib/galaxy/config.py
+++ b/lib/galaxy/config.py
@@ -943,7 +943,7 @@ class Configuration(object):
 
 
 def get_core_config_file():
-    return os.environ['GALAXY_CONFIG_FILE'] or 'config/galaxy.yml'
+    return os.environ.get('GALAXY_CONFIG_FILE') or 'config/galaxy.yml'
 
 
 def reload_config_options(current_config):
--- a/lib/galaxy/config.py
+++ b/lib/galaxy/config.py
@@ -126,6 +126,25 @@ LOGGING_CONFIG_DEFAULT = {
 }
 """Default value for logging configuration, passed to :func:`logging.config.dictConfig`"""
 
+RELOADABLE_CONFIG_OPTIONS = set([
+    # These are core config options (specified in galaxy.yml) that can be safely reloaded without a restart.
+    # To enable dynamic reloading, set watch_config and watch_config_options in galaxy.yml.
+    'message_box_content',
+    'welcome_url',
+    'tool_name_boost',
+    'tool_section_boost',
+    'tool_description_boost',
+    'tool_label_boost',
+    'tool_stub_boost',
+    'tool_help_boost',
+    'tool_search_limit',
+    'tool_enable_ngram_search',
+    'tool_ngram_minsize',
+    'tool_ngram_maxsize',
+    'admin_users',
+    'cleanup_job'
+])
+
 
 def resolve_path(path, root):
     """If 'path' is relative make absolute by prepending 'root'"""
@@ -921,6 +940,30 @@ class Configuration(object):
             return string
 
         return [parse(v) for v in allowed_origin_hostnames if v]
+
+
+def get_core_config_file():
+    return os.environ['GALAXY_CONFIG_FILE'] or 'config/galaxy.yml'
+
+
+def reload_config_options(current_config):
+
+    def _load_galaxy_core_config():
+        # Assume yaml format
+        config_file = get_core_config_file()
+        with open(os.path.abspath(config_file), 'r') as f:
+            configs = yaml.safe_load(f)
+            return configs['galaxy']
+
+    # load modified config file
+    new_config = _load_galaxy_core_config()
+    # check each option marked for dynamic reloading
+    for option in new_config['watch_config_options']:
+        # reload if option can be reloaded and its value is set in the modified config
+        if option in RELOADABLE_CONFIG_OPTIONS and option in new_config:
+            if getattr(current_config, option) != new_config.get(option):
+                setattr(current_config, option, new_config[option])
+                log.info('Reloading %s' % option)
 
 
 def parse_dependency_options(kwargs, root, dependency_resolvers_config_file):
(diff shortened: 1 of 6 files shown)


jmchilton:
Umm... why is this optional? Is other reloading optional? I feel like we should just always reload the reloadable options - maybe I'm wrong though...

> Also, can we assume the core config is a yaml file? 

Yes.

lib/galaxy/config.py
```diff
@@ -921,6 +942,30 @@ def parse(string):
         return [parse(v) for v in allowed_origin_hostnames if v]
 
 
+def get_core_config_file():
+    return os.environ.get('GALAXY_CONFIG_FILE') or 'config/galaxy.yml'
+
+
+def reload_config_options(current_config):
+
```
jmchilton, line 950: Can you use ``galaxy.util.load_app_properties`` to load the properties. I think that will be more correct - take into account environment variables.

jdavcs:
> Umm... why is this optional? Is other reloading optional? I feel like we should just always reload the reloadable options - maybe I'm wrong though...

I wasn't sure of this use case when galaxy is used in the wild.. If an admin would never need to turn this off, then it doesn't have to be optional. I gave the admin the option to (a) select what should be reloaded and (b) whether anything should be reloaded at all. If we don't need this, I'll remove them; will make it simpler.

EDIT: watch_tools and watch_job_rules are disabled by default. So I'll leave this as is (disabled by default) - i.e., adhere to the principle of least surprise. That is, unless anyone feels it should be enabled by default (with all reloadable options preselected).

lib/galaxy/config.py
```diff
@@ -921,6 +942,30 @@ def parse(string):
         return [parse(v) for v in allowed_origin_hostnames if v]
 
 
+def get_core_config_file():
+    return os.environ.get('GALAXY_CONFIG_FILE') or 'config/galaxy.yml'
```
jmchilton, line 946: I would not assume a default here - if GALAXY_CONFIG_FILE isn't set maybe just not do the reloading. Is there a way to tell what file uwsgi loaded? If that is injected somehow I'd trust that also.

lib/galaxy/config.py
jdavcs: Yes, I think so. I was looking for this in the codebase.. didn't find, so wrote my own. So yes, will do.

lib/galaxy/config.py
jdavcs: So, if this env var is not set, then check what uwsgi loaded (I would think it's possible - will check). Otherwise don't reload. 
I assume uwsgi could load one of these: https://github.com/galaxyproject/galaxy/blob/dev/scripts/galaxy-main#L63

jmchilton: I'm sure somehow a config file is found and used - but I want to track that and not try to refind it from inside the app - if that makes sense. Re-finding with slightly different logic would error prone - it'd work 85% of the time and we'd be reloading unreferenced configs the other 15% depending on how the app is loaded. With the galaxy-main script in particular - we may want to somehow update the config dict or environment to record what config file we used.

jdavcs: Yes, I agree on refinding. 
So galaxy-main is where (and when) the config file is determined? (i.e., that's the source of truth about where the configuration values come from?)

jmchilton: I think this is how to find the config file properly when coming from a wsgi server:
https://github.com/galaxyproject/galaxy/pull/921/files#diff-46f86ea0d82b84642e4218279ef61442R59

```
if global_conf and "__file__" in global_conf:
```

galaxy-main is a different app entry point that isn't used by webservers. Not sure if it sets this or not - but it probably should.

natefoo:
@ic4f This approach looks good and I agree with @jmchilton's comments so far. I think the all-or-nothing on/off option makes sense, but doubt anyone would need fine-grained control over the options to watch. That said, since it's implemented you could leave it.

This ties in to the question (for another issue, not here) of whether we need to perhaps add a "basic sample" config that only documents the options that admins are likely to set.

jdavcs:
@natefoo thanks for looking into this! I should probably wrap this up.. 
If it's unlikely anyone would want to reload individual options, let's keep it simple and not implement it (and reduce the chance of errors occurring due to interactions between rarely used options.) I'll modify the code to make it all-or-nothing on/off.

> This ties in to the question (for another issue, not here) of whether we need to perhaps add a "basic sample" config that only documents the options that admins are likely to set.

Yes! ..Can't resist: as per Henry Spencer (the guy who wrote regex), "the flip side of configurability is an urgent need for good defaults and an easy way to set everything to the default." So, yes!

lib/galaxy/config.py
jdavcs: Found it:  config file is found in a util/ function find_config_file() (surprise...) Furthermore, that method is called from BaseAppConfiguration - so the config file is already available as app.config.config_file. Thus, there's no need to re-find it - so there's no need for this additional function. PR forthcoming. Marking as resolved.

lib/galaxy/config.py
jdavcs: On second thought, I believe we don't need to take into account env vars here. We are watching the config file: when it changes we update the current (reloadable) config values. Those config values, initially, may have been loaded from the config file or env vars or command args - but that's irrelevant at this point, as they are being overridden by a runtime modification to the config file). I think "reloadable" implies "reload value when I manually change the config file, regardless of how that value was initially set".

jdavcs: On the other hand, `load_app_properties` contains config reading logic that is more elaborate than just `yaml.safe_load`. But then there's also the overriding part from os.environ. So, that function reads from a file AND overrides values if env vars are set. I think I'll modify load_app_properties so we have the option to use it for reading only.

jdavcs:
Replaced by #8475

PR #8232, opened 2019-06-26
[WIP] Implement optional and default_value for workflow input parameters.

jmchilton:
Initial commit here from CWL branch (https://github.com/common-workflow-language/galaxy/pull/47) where it does more or less work for executing CWL workflows - but that branch depends on a ``FieldTypeToolParameter`` for capturing the structured default value and probably has no functional GUI for configuring these things.

TODO:
 - [x] Implement a gxformat2 syntax for specifying an input as optional.
 - [x] Implement a gxformat2 syntax for specifying a default value.
 - [x] Write tests for optional in vanilla Galaxy workflows.
 - [x] Write tests for default_value in vanilla Galaxy workflows.
 - [x] Establish Galaxy workflow semantics for specifying an input as optional with no default value.
 - [x] Rework this backend to work without ``FieldTypeToolParameter ``  (use of ``TextToolParameter `` in initial commit is a bootstrap).
 - [ ] Implement a GUI works for marking an input as optional.
 - [ ] Implement a GUI for specifying a default value.

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -4655,6 +4655,15 @@ class WorkflowStep(RepresentById):
     def tool_uuid(self):
         return self.dynamic_tool and self.dynamic_tool.uuid
 
+    @property
+    def input_default_value(self):
+        tool_inputs = self.tool_inputs
+        tool_state = tool_inputs
+        default_value = tool_state.get("default")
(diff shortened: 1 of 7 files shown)


jmchilton:
This branch didn't have the GUI work I did at the GCC it seems. I was confused.

PR #8235, opened 2019-06-26
Refactor to cleanup dependencies around galaxy.web.stack.

jmchilton:
Move galaxy.web.stack.database_heartbeat into galaxy.model - since it was the only thing in galaxy.web.stack that used galaxy.model at all and it didn't import anything else from galaxy.web.stack.

Once that was done, I refactored this whole package (galaxy.web.stack) into galaxy.web_stack. It doesn't use the framework setup or exported by the base galaxy.web package so it wouldn't seem to belong there. It also has very different dependencies than the rest of that. ~And finally and most importantly, this eliminates almost all dependency from galaxy.jobs and galaxy.tools and galaxy.workflows to anything in galaxy.web.~ (Update: I went through a did a more thorough check after more refactoring and those modules do depend on ``url_for`` a lot still - but they really shouldn't right? I think this is still a good refactoring into more modular chunks but I could be convinced I'm wrong I think.)


mvdbeek (approved):
Sorry about the arbitrary placement of DatabaseHeartbeat, I think it made more sense while I was still developing it.

PR #8239, opened 2019-06-27
unzip/zip collection: set format equal to input format

bernt-matthias:
currently `unzip collection` produces data sets of the format data. 

- in the workflow editor these can't be connected to downstream tools, except if the output data type is explicitly set which limits applicability of workflows
- in interactive processing the user needs to wait until the unzip collection tool finishes 

My fix is (again) guesswork, but seems to work -- maybe there is a more elegant solution. 

Also I'm not sure if the fix for `zip collection` is the best way to do it.


mvdbeek (approved):
This looks fine to me, a nice improvement!

PR #8241, opened 2019-06-27
Implement galaxy-web-apps package outline.

jmchilton:
Builds on #8237.


mvdbeek (approved):
The readmes are not all pointing to the right pypi urls, but we can fix this later!
Thanks so much @jmchilton !

jxtx:
I don't understand the switch s/web.base/webapps.base/

Previously the web package was for the framework and webapps only contained specific apps. Can you elaborate a bit on why this is better?

jmchilton:
@jxtx: 

tl;dr - so we can create an abstract galaxy-web-framework package on PyPI.

Longer version:

From the commit message:

> The rest of galaxy.web does not import galaxy.tools, galaxy.workflows, galaxy.managers, etc.. so this makes galaxy.web feel more like a generic galaxy-agnostic framework module - really only the grids helper imports anything from the galaxy-data package at all and it is just some small helpers.

I'm creating a package ``galaxy-web-framework`` from ``galaxy.web`` that is pretty generic and not tied to specifics of the Galaxy application logic. ``galaxy.web.base.controller`` contains obviously a bunch of Galaxy business logic - but it is the only thing in ``galaxy.web`` that does. Additionally that base controller is only used from the ``galaxy-webapps`` package / ``galaxy.webapps`` - so it feels appropriate to place the logic closer to those files. Additionally, separating these into separate packages helps ensures ``galaxy.webapps`` depends on ``galaxy.web`` but not vice versa (DAGs are good).

So if you insist we don't refactor this way - my next step would be to start by just not creating a ``galaxy-web-framework`` package (I think there is benefit but it might be speculative complexity) and just package the controllers and abstract framework code together into ``galaxy-webapps`` (i.e. both ``galaxy.web`` and ``galaxy.webapps`` together). If at some point we decide we want to extract the non-Galaxy application logic framework pieces into an abstract package - I guess we could refactor all of ``galaxy.web`` besides the base controller out into a ``galaxy.web_framework`` module/package. But a ``galaxy.web`` package that just contains the base controller feels silly.

Either way - we mostly want to reduce that base controller over time (at least the mixins and such that I'm most familiar with) as logic gets refactored into managers - so it doesn't belong in something that is attempting to be a generic framework.

Another compromise might be to refactor to ``galaxy.webapps._base`` or something that makes it clear this isn't something we are thinking we'd export from webapps - it is just utilities used by the package.

jxtx:
> So if you insist we don't refactor this way 

It was just a question! I didn't understand the commit message, this helped clarify. Thanks.

Are you think of breaking up galaxy-webapps itself in the future and then this would end up in something like galaxy-webapps-base? It would be good to not keep the webapps tied together.

jmchilton:
@jxtx I think it could definitely make sense to break them down more. Glad my comments helped clarify.

natefoo (approved):

Issue #8223, opened 2019-06-25
Data Managers - Read/Update/Delete operations on Data Tables


selten:
Data managers currently can append data tables (as far as I know). I'd like to suggest that read (this is already possible through the API but it would be nice if a data manager could be provided the current status of the data table), update and delete operations are added as well.
This would allow extra flexibility with data tables or data sources that have lots of fluctuations, for instance databases where sometimes data is deleted from. 

I believe this would also fit in well with https://github.com/galaxyproject/galaxy/issues/6819, as this would already create the update operation required for such a text editor.

selten:
After discussing on Gitter, reproducibility could be at risk with an update operation.

A delete operation might be useful nonetheless though. Also, being able to get the contents of the data table into the data manager could be useful for offloading deduplication to data manager.

PR #8250, opened 2019-06-28
Data managers - allow removal of entries

selten:
Implements the request from https://github.com/galaxyproject/galaxy/issues/8223

This adds an easy optional change to a data manager to remove data as well. It is fully backwards compatible with the data manager format that was used.
It uses features that the API also uses (https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/api/tool_data.py#L52) to delete entries in a data table. 

It also migrates data managers documentation from https://galaxyproject.org/admin/tools/data-managers/ to https://docs.galaxyproject.org/en/master/.

This change adds a bit more flexibility to data managers where they get the ability to e.g. remove data from a data table that is invalid or might have expired.


jmchilton:
We already have an API endpoint that does this - so I see no issue with this. I also like having the documentation moved here. @blankenberg I'll give you a couple days to comment, but if I haven't heard from you then I'm going to merge.

jmchilton:
@galaxybot test this

selten:
Could anyone maybe help me clarify why the `test.api.test_datasets.DatasetsApiTestCase.test_show` from `api test python 3` and the `test.integration.test_upload_configuration_options.AdminsCanPasteFilePathsTestCase.test_admin_path_paste_libraries` from `integration test` would fail?   
    
The odd part is that this only failed `api test python 3` and not `api test` as well as `integration test` and not for `integration test python 3`.

jmchilton:
Probably transient failures - I wouldn't worry about these.

jmchilton:
Awesome - thanks so much!

PR #8252, opened 2019-06-30
ease the usage of galaxy.yml in ansible

lecorguille:
Using anchor, you just need to modify the galaxy.yml and the ansible-galaxy role will do the almost rest. Thus, it's easier to sync the galaxy.yml after an update to the instance!?

```yaml
- name: galaxy server
  hosts: galaxy
  vars_files:
    - "{{ inventory_dir }}/group_vars/galaxy/main.yml"
    - "{{ inventory_dir }}/group_vars/galaxy/galaxy.yml"
  handlers:
    - name: Restart Galaxy
      supervisorctl:
        name: galaxy
        state: restarted
  pre_tasks:
    - name: Install Dependencies
      package:
        name: ['python-psycopg2', 'git', 'python-virtualenv', 'make']
  roles:
     - role: geerlingguy.pip
     - role: usegalaxy-eu.supervisor
     - role: galaxyproject.galaxy
```

### DIFF
--- a/config/galaxy.yml.sample
+++ b/config/galaxy.yml.sample
@@ -16,7 +16,7 @@
 # 
 # Config hackers are encouraged to check there before asking for help.
 # 
-uwsgi:
+uwsgi: &UWSGI
 
   # The address and port on which to listen.  By default, only listen to
   # localhost (galaxy will not be accessible over the network).  Use
@@ -96,7 +96,7 @@ uwsgi:
   # Ensure application threads will run if `threads` is unset.
   enable-threads: true
 
-galaxy:
+galaxy: &GALAXY
 
   # By default, Galaxy uses a SQLite database at
   # 'database/universe.sqlite'.  You may use a SQLAlchemy connection
@@ -1772,3 +1772,8 @@ galaxy:
   # reason to disable it.
   #use_pbkdf2: true
 
+galaxy_config:
+  "uwsgi":
+    <<: *UWSGI
+  "galaxy":
+    <<: *GALAXY


martenson:
the sample file is not directly modified but rather generated via https://github.com/galaxyproject/galaxy/blob/dev/Makefile#L94 from https://github.com/galaxyproject/galaxy/blob/dev/doc/source/admin/galaxy_options.rst

lecorguille:
Damn! 😝 
Ok, but before trying, imagine that I succeed, is it something that can be acceptable?

jmchilton:
uwsgi has very particular YAML processing - I don't think this would work - would it?

lecorguille:
@jmchilton It seems so!
It's what we done here: [galaxy.yml](https://gitlab.com/ifb-elixirfr/usegalaxy-fr/infrastructure/blob/preprod/production/group_vars/galaxy/galaxy.yml#L19) and [playbook_galaxy.yml](https://gitlab.com/ifb-elixirfr/usegalaxy-fr/infrastructure/blob/preprod/playbook_galaxy.yml#L12)

BTW, I didn't find the time yet to implement that using your procedure.

mvdbeek:
@nuwang I thought you might have thoughts on this or advise for @lecorguille

nuwang:
@mvdbeek Thanks for bringing this up. My thoughts on this now are that it would be strongly desirable to move the uwsgi config to a separate file as you found was possible last time, because it really isn't yaml and makes the galaxy config file unmodifiable programatically, which is a bad situation to be in. Since it's relatively less likely that someone has modified the defaults in uwsgi.yaml, I think that this could be done with comparatively less disruption. It works quite well in the k8s setup.

@lecorguille The issue that will cause this to fail is the difference in how yaml arrays are defined here:
[link](https://gitlab.com/ifb-elixirfr/usegalaxy-fr/infrastructure/blob/preprod/production/group_vars/galaxy/galaxy.yml#L47)
And here:
https://github.com/galaxyproject/galaxy/blob/43af97847d3a45f056a602c5cbab255bcc3ce433/lib/galaxy/config/sample/galaxy.yml.sample#L45

The repeating static-map in the latter case is invalid yaml, and will cause only the last entry to be included when you use a yaml anchor.

The reason this works with ansible-galaxy is because it has special logic to convert the valid yaml to invalid yaml that uwsgi understands:
https://github.com/galaxyproject/ansible-galaxy/blob/f450af8a6b8ad5c7c400603f4eed135810c8d81f/filter_plugins/uwsgi_yaml.py#L55

And even the libyaml version of uwsgi still doesn't use valid yaml: https://github.com/unbit/uwsgi/issues/2097
so this is unlikely to be resolved in the short run.

Instead, what if you modify this [line](https://gitlab.com/ifb-elixirfr/usegalaxy-fr/infrastructure/blob/preprod/production/group_vars/galaxy/galaxy.yml#L1783) to use a [jinja2 combine](http://toroid.org/ansible-combine) instead of a yaml anchor? That way, you might still be able to do what you're suggesting without needing to modify the sample config at all.

nuwang (changes requested):
@mvdbeek Thanks for bringing this up. My thoughts on this now are that it would be strongly desirable to move the uwsgi config to a separate file as you found was possible last time, because it really isn't yaml and makes the galaxy config file unmodifiable programatically, which is a bad situation to be in. Since it's relatively less likely that someone has modified the defaults in uwsgi.yaml, I think that this could be done with comparatively less disruption. It works quite well in the k8s setup.

@lecorguille The issue that will cause this to fail is the difference in how yaml arrays are defined here:
[link](https://gitlab.com/ifb-elixirfr/usegalaxy-fr/infrastructure/blob/preprod/production/group_vars/galaxy/galaxy.yml#L47)
And here:
https://github.com/galaxyproject/galaxy/blob/43af97847d3a45f056a602c5cbab255bcc3ce433/lib/galaxy/config/sample/galaxy.yml.sample#L45

The repeating static-map in the latter case is invalid yaml, and will cause only the last entry to be included when you use a yaml anchor.

The reason this works with ansible-galaxy is because it has special logic to convert the valid yaml to invalid yaml that uwsgi understands:
https://github.com/galaxyproject/ansible-galaxy/blob/f450af8a6b8ad5c7c400603f4eed135810c8d81f/filter_plugins/uwsgi_yaml.py#L55

And even the libyaml version of uwsgi still doesn't use valid yaml: https://github.com/unbit/uwsgi/issues/2097
so this is unlikely to be resolved in the short run.

Instead, what if you modify this [line](https://gitlab.com/ifb-elixirfr/usegalaxy-fr/infrastructure/blob/preprod/production/group_vars/galaxy/galaxy.yml#L1783) to use a [jinja2 combine](http://toroid.org/ansible-combine) instead of a yaml anchor? That way, you might still be able to do what you're proposing (which is very useful) without needing to modify the sample config at all.

mvdbeek:
Given @nuwang's suggestion I'm going to close this one, but thanks for this nice idea @lecorguille!

PR #8253, opened 2019-06-30
[19.05] Expression tool fix: Include workdir path when reading env var from file

mvdbeek:
`env_setup_commands` runs before `cd $working_directory` in the job
file template, so just using the relative path fails with
```
  File "_evaluate_expression_.py", line 1, in <module>
    from galaxy_ext.expressions.handle_job import run; run()
  File "/opt/galaxy/server/lib/galaxy_ext/expressions/handle_job.py", line 31, in run
    with open(environment_path, "r") as f:
IOError: [Errno 2] No such file or directory: ''

cat: tmp6CL_1e: No such file or directory
```


mvdbeek:
Came up in https://github.com/galaxyproject/training-material/pull/1393#issuecomment-505277093

mvdbeek:
4th time's the charm, now avoiding to pass an absolute path.

mvdbeek:
Thanks a lot @martenson!

PR #8254, opened 2019-07-01
[19.05] Fix regex rule count

mvdbeek:
Otherwise the rule validation will complain with `Problem applying regular expression, wrong number of groups found.`.
Needed for https://galaxyproject.github.io/training-material/topics/galaxy-data-manipulation/tutorials/upload-rules/tutorial.html#apply-rules-to-existing-collections

### DIFF
--- a/client/galaxy/scripts/mvc/rules/rule-definitions.js
+++ b/client/galaxy/scripts/mvc/rules/rule-definitions.js
@@ -253,7 +253,7 @@ const RULES = {
                 component.addColumnRegexTarget = rule.target_column;
                 component.addColumnRegexExpression = rule.expression;
                 component.addColumnRegexReplacement = rule.replacement;
-                component.addColumnRegexGroupCount = rule.group_count;
+                component.addColumnRegexGroupCount = parseInt(rule.group_count);
             }
             let addColumnRegexType = "global";
             if (component.addColumnRegexGroupCount) {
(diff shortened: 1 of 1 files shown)


hexylena (approved):

client/galaxy/scripts/mvc/rules/rule-definitions.js
```diff
@@ -253,7 +253,7 @@ const RULES = {
                 component.addColumnRegexTarget = rule.target_column;
                 component.addColumnRegexExpression = rule.expression;
                 component.addColumnRegexReplacement = rule.replacement;
-                component.addColumnRegexGroupCount = rule.group_count;
+                component.addColumnRegexGroupCount = parseInt(rule.group_count);
```
dannon, line 256: rule.group_count will never be a string starting with 0, right?  parseInt can have unexpected behavior when radix is not specified.  It's probably safest to be explicit here, like:

```suggestion
                component.addColumnRegexGroupCount = parseInt(rule.group_count, 10);
```

mvdbeek: Thanks, excellent point. I've updated all other occurences of parseInt in the rule builder as well, because users can enter a leading 0.

dannon:
@mvdbeek Thanks!

PR #8264, opened 2019-07-05
Implement 'select' Input Parameter in Workflow Editor

jraysajulga:
_GCC2019 CoFest (@mvdbeek, @rplanel)_

**Create backend and frontend for select parameters in workflows**

This project would introduce a new syntax and user interface that lets users create select lists (as in tools) in the workflow editor that will be connectable to select lists in tools.
Screen Shot 2019-07-05 at 6 12 51 AM

- [x] "**Front end:**" Add "Select List" option to _Input Parameter_ in the Workflow Editor.
    - [x] Display those options in the executing-workflow pane.
- [x] **Back end:** Construct list of available options from the select-type inputs from tools connect to that Input Parameter.
    - [x] Handle multiple outputs
        - [x] Multiple labels for one value
        - [ ] Multiple values for one label? (Ex: **Yes** is --clip and -q for SFF converter and Trim, respectively; would this be useful to enable?)
   - [x] Intersection
       - [ ] Case: intersection provides a set of one or zero options → error message? 
- [x] Handle select with "multiple" and/or "checkbox" enabled
- [x] Handle select parameters nested within a conditional
- [x] Handle select parameters with radio buttons


mvdbeek:
@galaxybot test this

jraysajulga:
This is my first shot at adding internal Galaxy code so please modify as necessary. Thanks!

mvdbeek:
Thanks a lot @jraysajulga, this will be a great addition, I'll review and make the necessary changes in the incoming days!

bgruening:
@jraysajulga this is super-cool!

jraysajulga:
> Thanks a lot @jraysajulga, this will be a great addition, I'll review and make the necessary changes in the incoming days!

@mvdbeek Great. Thanks a lot for guiding me through this.

> @jraysajulga this is super-cool!

@bgruening Thanks! You are super cool, too!

jj-umn:
Can this work for a select based on data tables, e.g. HISAT2 and StringTie?   Would be nice to select a genome: hg19, hg38, or mm10 and link to HISAT2 (HISAT2 index) and StringTie (2bit reference).   This would only if the respective data tables used the same name for its index value.

mvdbeek:
That would work with this implementation, yes. I'll need to find some time to review and add tests (if no one else wants to jump at it ...).

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,34 @@ def get_runtime_inputs(self, **kwds):
...
+                # Select options within conditionals have a deeper hierarchy: cases
+                if connection.input_name not in tool_inputs and "|" in connection.input_name:
+                    param_name, case_name = connection.input_name.split("|")
+                    param = tool_inputs.data[param_name]
+                    if param.type == "conditional":
+                        sel_case = [case for case in param.cases if case_name in list(case.inputs.data.keys())][0]
+                        input_data = sel_case.inputs.data[case_name]
+                else:
+                    input_data = tool_inputs.data[connection.input_name]
+                static_options.append(input_data.static_options)  # Aggregation input select options from several connections
```
jj-umn: In order to pick up dynamic options, I think this might need to be changed to:
static_options.append(input_data.get_options(self.trans,None))  # Aggregation input select options from several connections

jj-umn (changes requested):

lib/galaxy/workflow/modules.py
jraysajulga: Thanks @jj-umn!

jraysajulga:
@mvdbeek Any idea why this is failing on only "py35_first_startup"?

```
sudo apt-get install -y libpython3.5-dev00:00
Exit code: 100
#!/bin/bash -eo pipefail
sudo apt-get install -y libpython3.5-dev
Reading package lists... Done

Building dependency tree       

Reading state information... Done

E: Unable to locate package libpython3.5-dev
E: Couldn't find any package by glob 'libpython3.5-dev'
E: Couldn't find any package by regex 'libpython3.5-dev'
Exited with code 100
```

mvdbeek:
We fixed that in https://github.com/galaxyproject/galaxy/pull/8337, if you rebase against dev that error should go away

jj-umn:
@mvdbeek @bgruening @jmchilton Any chance we could get this included in the 19.09 release?  
This would be useful for simplifying inputs for complex workflows.
I've checked it out on my local dev install and it worked with hisat2 and bwa-mem getting intersection of cached genome indexes for the tools.

mvdbeek:
Thanks for giving this a shot @jj-umn, this is definitely a priority, there are some things to think through and some tests to be added, it does look good.

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,34 @@ def get_runtime_inputs(self, **kwds):
...
         input = parameter_class(None, element)
+
+        # Retrieve possible runtime options for 'select' type inputs
+        if parameter_type == 'select' and connections:
+            static_options = []
+            for connection in connections:
+                tool_inputs = connection.input_step.module.tool.inputs
+                # Account for select options within conditionals
+                if connection.input_name not in tool_inputs and "|" in connection.input_name:
+                    param_name, case_name = connection.input_name.split("|")
```
mvdbeek: Does this work for nested conditionals ? I don't remember in what form they appear here, but it is possible you'd need to do `split("|", 1)` to only split off the first `|` since the rest may be flat in `tool_inputs.data`, or otherwise you'd need to recurse here.

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,34 @@ def get_runtime_inputs(self, **kwds):
...
+                    param = tool_inputs.data[param_name]
+                    if param.type == "conditional":
+                        sel_case = [case for case in param.cases if case_name in list(case.inputs.data.keys())][0]
```
mvdbeek: in python you can do
```
>>> d = {'key': 'value'}
>>> 'key' in d
True
```
so this should work and is more lightweight
```suggestion
                        sel_case = [case for case in param.cases if case_name in case.inputs.data][0]
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,34 @@ def get_runtime_inputs(self, **kwds):
...
+                        input_data = sel_case.inputs.data[case_name]
+                else:
+                    input_data = tool_inputs.data[connection.input_name]
+                static_options.append(input_data.get_options(self.trans, None))  # Aggregation input select options from several connections
```
mvdbeek: `other_values` is treated as a dictionary in various tool parameters, so that would be more correct.
```suggestion
                static_options.append(input_data.get_options(self.trans, {}))  # Aggregation input select options from several connections
```

I assume some parameter types are probably not going to work because of that shortcut, for instance because `other_value` may contain references to selected datasets in the tool / workflow form, but I think that's OK for now.

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,34 @@ def get_runtime_inputs(self, **kwds):
...
+            if static_options != []:
```
mvdbeek:
```suggestion
            if static_options:
```
is what we usually do to check if lists aren't empty

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,34 @@ def get_runtime_inputs(self, **kwds):
...
+                # Intersection based on values.
+                intxn_vals = set.intersection(*[set([option[1] for option in options]) for options in static_options])
```
mvdbeek: The next suggestion uses a generator and directly uses a set comprehension instead of casting the list to a set
```
>>> d = {'key': 'value'}
>>> g = ({k} for k in d)
>>> g
<generator object <genexpr> at 0x1017cb5d0>
>>> set.intersection(*g)
{'key'}
```
```suggestion
                intxn_vals = set.intersection(*({option[1] for option in options}) for options in static_options))
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,34 @@ def get_runtime_inputs(self, **kwds):
...
+                intxn_opts = [option for options in static_options for option in options if option[1] in intxn_vals]
```
mvdbeek: Since you're using this as a set you can use
```suggestion
                intxn_opts = {option for options in static_options for option in options if option[1] in intxn_vals}
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,34 @@ def get_runtime_inputs(self, **kwds):
...
+                d = defaultdict(list)  # Collapse labels with same values
```
mvdbeek:
```suggestion
                d = defaultdict(set)  # Collapse labels with same values
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,34 @@ def get_runtime_inputs(self, **kwds):
...
+                for label, value, selected in list(set(intxn_opts)):
```
mvdbeek:
```suggestion
                for label, value, selected in intxn_opts:
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,34 @@ def get_runtime_inputs(self, **kwds):
...
+                    d[value].append(label)
```
mvdbeek:
```suggestion
                    d[value].add(label)
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,34 @@ def get_runtime_inputs(self, **kwds):
...
+                intxn_input = copy(input_data)
+                intxn_input.static_options = [(', '.join(set(label)), value, False) for value, label in list(d.items())]
```
mvdbeek:
```suggestion
                intxn_input.static_options = [(', '.join(label), value, False) for value, label in d.items()]
```

mvdbeek (commented):
This looks pretty good, I added some suggestions and questions in line. I'd also be absolutely great to have unit tests for the various SelectParameter types we have (https://github.com/galaxyproject/galaxy/blob/04f196439dbc4e41f2a49603bb89c0e0a5b62178/test/unit/tools/test_select_parameters.py#L7 might be a start to see how one could approach writing these). That's totally not necessary though to merge the PR, we can add them later, just let me know.

In the same vein we could also add API tests once https://github.com/jmchilton/gxformat2/pull/11 is merged, but that's also more work.

jj-umn:
@mvdbeek @jraysajulga  I wrote a tool just to more throughly test this, including nested conditionals.  Looks like a little more work to do.

lib/galaxy/workflow/modules.py
jj-umn: @mvdbeek Good question regarding nested conditionals and it also applies to repeats.  
I couldn't figure out how to add a suggested replacement for multiple lines, so added where the first line was changed.
To work for nesting of conditionals or repeats:
```suggestion
                    param_name, case_name = connection.input_name.split("|",1)
                    if param_name in tool_inputs.data:
                        param = tool_inputs.data[param_name]
                    elif re.match('.+_\d+',param_name):  # check for repeat label
                        param_name,param_idx = param_name.rsplit('_',1)
                        param = tool_inputs.data[param_name]
                    while param.type in ("conditional", "repeat"):
                        if "|" in case_name:
                            param_name, case_name = case_name.split("|",1) 
                            if param.type == "conditional":
                                sel_cases = [case for case in param.cases if param_name in case.inputs.data]
                                if sel_cases:
                                    sel_case = sel_cases[0] 
                                elif re.match('.+_\d+',param_name):  # check for repeat label
                                    param_name,param_idx = param_name.rsplit('_',1)
                                    sel_case = [case for case in param.cases if param_name in case.inputs.data][0]
                                param = sel_case.inputs.data[param_name]
                            elif param.type == "repeat":
                                if param_name in param.inputs:
                                    param = param.inputs[param_name]
                                elif re.match('.+_\d+',param_name):  # check for repeat label
                                    param_name,param_idx = param_name.rsplit('_',1)
                                    param = param.inputs[param_name]
                        else:
                            if param.type == "conditional":
                                sel_case = [case for case in param.cases if case_name in case.inputs.data][0]
                                input_data = sel_case.inputs.data[case_name]
                            elif param.type == "repeat":
                                input_data = param.inputs[case_name]
                            break
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,63 @@ def get_runtime_inputs(self, **kwds):
...
+                    param = tool_inputs.data[param_name]
+                    if param.type == "conditional":
+                        sel_case = [case for case in param.cases if case_name in case.inputs.data][0]
+                        input_data = sel_case.inputs.data[case_name]
+                else:
+                    input_data = tool_inputs.data[connection.input_name]
+                static_options.append(input_data.get_options(self.trans, {}))  # Aggregation input select options from several connections
+            if static_options:
+                # Intersection based on values.
+                intxn_vals = set.intersection(*({option[1] for option in options}) for options in static_options))
```
mvdbeek: You've got a syntax error here, but it's hard to spot.
```suggestion
                intxn_vals = set.intersection(*({option[1] for option in options} for options in static_options))
```

lib/galaxy/workflow/modules.py
```diff
@@ -1,9 +1,11 @@
 """
 Modules used in building workflows
 """
+import copy
```
jj-umn: Actually need "from copy import copy"
```suggestion
```

lib/galaxy/workflow/modules.py
```diff
@@ -1,9 +1,11 @@
...
 import json
 import logging
 import re
+from collections import defaultdict
 from collections import OrderedDict
```
jj-umn, line 8:
```suggestion
from collections import OrderedDict
from copy import copy
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,63 @@ def get_runtime_inputs(self, **kwds):
...
+
+        # Retrieve possible runtime options for 'select' type inputs
+        if parameter_type == 'select' and connections:
+            static_options = []
+            for connection in connections:
+                tool_inputs = connection.input_step.module.tool.inputs
+                # Account for select options within conditionals
+                if connection.input_name not in tool_inputs and "|" in connection.input_name:
+                    param_name, case_name = connection.input_name.split("|", 1)
+                    if param_name in tool_inputs.data:
```
jj-umn: With OrderedDict  ".data"  no longer applies.
```suggestion
                    if param_name in tool_inputs:
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,63 @@ def get_runtime_inputs(self, **kwds):
...
+                        param = tool_inputs.data[param_name]
```
jj-umn:
```suggestion
                        param = tool_inputs[param_name]
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,63 @@ def get_runtime_inputs(self, **kwds):
...
+                    elif re.match('.+_\\d+', param_name):  # check for repeat label
```
jj-umn, line 714: Using '\d+' to find repeat index as in 'rep_name_2'
```suggestion
                    elif re.match('.+_\d+', param_name):  # check for repeat label
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,63 @@ def get_runtime_inputs(self, **kwds):
...
+                        param_name, param_idx = param_name.rsplit('_', 1)
+                        param = tool_inputs.data[param_name]
```
jj-umn:
```suggestion
                        param = tool_inputs[param_name]
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,63 @@ def get_runtime_inputs(self, **kwds):
...
+                    while param.type in ("conditional", "repeat"):
+                        if "|" in case_name:
+                            param_name, case_name = case_name.split("|", 1)
+                            if param.type == "conditional":
+                                sel_cases = [case for case in param.cases if param_name in case.inputs.data]
```
jj-umn:
```suggestion
                                sel_cases = [case for case in param.cases if param_name in case.inputs]
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,63 @@ def get_runtime_inputs(self, **kwds):
...
+                                if sel_cases:
+                                    sel_case = sel_cases[0]
+                                elif re.match('.+_\\d+', param_name):  # check for repeat label
```
jj-umn:
```suggestion
                                elif re.match('.+_\d+', param_name):  # check for repeat label
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,63 @@ def get_runtime_inputs(self, **kwds):
...
+                                    param_name, param_idx = param_name.rsplit('_', 1)
+                                    sel_case = [case for case in param.cases if param_name in case.inputs.data][0]
```
jj-umn:
```suggestion
                                    sel_case = [case for case in param.cases if param_name in case.inputs][0]
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,63 @@ def get_runtime_inputs(self, **kwds):
...
+                                param = sel_case.inputs.data[param_name]
```
jj-umn:
```suggestion
                                param = sel_case.inputs[param_name]
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,63 @@ def get_runtime_inputs(self, **kwds):
...
+                            elif param.type == "repeat":
+                                if param_name in param.inputs:
+                                    param = param.inputs[param_name]
+                                elif re.match('.+_\\d+', param_name):  # check for repeat label
+                                    param_name, param_idx = param_name.rsplit('_', 1)
+                                    param = param.inputs[param_name]
+                        else:
+                            if param.type == "conditional":
+                                sel_case = [case for case in param.cases if case_name in case.inputs.data][0]
```
jj-umn:
```suggestion
                                sel_case = [case for case in param.cases if case_name in case.inputs][0]
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,63 @@ def get_runtime_inputs(self, **kwds):
...
+                                input_data = sel_case.inputs.data[case_name]
```
jj-umn:
```suggestion
                                input_data = sel_case.inputs[case_name]
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,63 @@ def get_runtime_inputs(self, **kwds):
...
+                            elif param.type == "repeat":
+                                input_data = param.inputs[case_name]
+                            break
+                    param = tool_inputs.data[param_name]
```
jj-umn: Remove.   Taken care of within while loop. 
```suggestion
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,63 @@ def get_runtime_inputs(self, **kwds):
...
+                    if param.type == "conditional":
```
jj-umn: Remove. Taken care of within while loop.
```suggestion
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,63 @@ def get_runtime_inputs(self, **kwds):
...
+                        sel_case = [case for case in param.cases if case_name in case.inputs.data][0]
```
jj-umn: Remove. Taken care of within while loop.
```suggestion
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,63 @@ def get_runtime_inputs(self, **kwds):
...
+                        input_data = sel_case.inputs.data[case_name]
```
jj-umn: Remove. Taken care of within while loop.
```suggestion
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,63 @@ def get_runtime_inputs(self, **kwds):
...
+                else:
+                    input_data = tool_inputs.data[connection.input_name]
```
jj-umn:
```suggestion
                    input_data = tool_inputs[connection.input_name]
```

jj-umn:
@jraysajulga   @mvdbeek and I suggested some code changes.  ( I remembered to run flake8 this time. )   
Tested this with workflow with nested repeats and conditionals and mixture of static and dynamic selects on galaxy dev:
commit 053510678f5078939f1737384775d770112b04ab (HEAD -> dev, origin/dev, origin/HEAD)
Merge: b179c72412 63ddb90c06
Author: Dannon <dannon.baker@gmail.com>
Date:   Tue Aug 20 11:36:38 2019 -0400

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,59 @@ def get_runtime_inputs(self, **kwds):
...
+                        param_name, param_idx = param_name.rsplit('_', 1)
+                        param = tool_inputs[param_name]
+                    while param.type in ("conditional", "repeat"):
+                        if "|" in case_name:
+                            param_name, case_name = case_name.split("|", 1)
+                            if param.type == "conditional":
+                                sel_cases = [case for case in param.cases if param_name in case.inputs]
+                                if sel_cases:
+                                    sel_case = sel_cases[0]
+                                elif re.match('.+_\\d+', param_name):  # check for repeat label
```
jj-umn: The regex pattern should be declared as a **raw** string:  **r**'.+_\d+'  
The repeat_param will be the repeat name which can be at least one character: '.+'
followed by the underscore character: '_'
followed by an integer repeat index: '\d+'
```suggestion
                                elif re.match(r'.+_\d+', param_name):  # check for repeat label
```

lib/galaxy/workflow/modules.py
```diff
@@ -697,6 +700,59 @@ def get_runtime_inputs(self, **kwds):
...
+                                    param_name, param_idx = param_name.rsplit('_', 1)
+                                    sel_case = [case for case in param.cases if param_name in case.inputs][0]
+                                param = sel_case.inputs[param_name]
+                            elif param.type == "repeat":
+                                if param_name in param.inputs:
+                                    param = param.inputs[param_name]
+                                elif re.match('.+_\\d+', param_name):  # check for repeat label
```
jj-umn:
```suggestion
                                elif re.match(r'.+_\d+', param_name):  # check for repeat label
```

jj-umn:
@jraysajulga   I added code suggestions to correct the regex patterns.   The regex pattern should be declared as a raw string using the 'r' stringprefix.  That should prevent the issue with a backslash in the pattern.   
'.+_\\d+' would match:   'repeat_name_\dddd'    literal backslash and 1 or more 'd' characters
r'.+_\d+' should match 'repeat_name_10'    one or more digit characters, i.e 0-9

jmchilton:
Amazing contribution and super useful. Thanks all - great work. I've rebased this with significant changes into https://github.com/galaxyproject/galaxy/pull/9086. This approach here seems to work fine and would work fine going forward I think but #9086 is just more aligned with directions I'd like to move the workflow system in though (expression tools, CWL).

mvdbeek:
https://github.com/galaxyproject/galaxy/pull/9086 is merged now and will be part of 20.01. Thanks everyone, this was a really cool project! I hope you'll be building cool stuff with this :).

PR #8270, opened 2019-07-05
Update Sphinx RST files for lib/

jdavcs:
Running sphinx 1.8.5 updates all files

Ping @nsoranzo

Issue #8240, opened 2019-06-27
Small-to-Medium Tasks to Improve Galaxy Packages.


jmchilton:
As mentioned in https://github.com/galaxyproject/galaxy/issues/8140 improving Galaxy ``packages/`` provides a lot of small-to-medium size issues that could be tackled by newcomers with different skillsets and help a wide variety of different parts of the Galaxy code base.

We also now have some rudimentary CI for these packages now. Hence, each new package that has tests of any complexity ensures that pytest imports all the files and finds any dependency issues during CI.  This ensures over the long term we're dealing with dependencies between packages more intelligently, tracking external dependencies in a more structured way, etc.. This will have a large, automatically enforced impact on the quality of the code we are writing. This has the makings of some of the really great hackathon projects of GCCs past.

- [x] Merge @natefoo's installable branch galaxyproject/galaxy#921
- [ ] Add Sphinx docs to the projects and shared build system - focused on docstrings and improved READMEs at first.
   - [ ] Could think about how to split out some of the docs for things like the mulled utilities (there are docs in galaxy-lib but not in Galaxy, that needs to be fixed).
- [ ] Review all non-.py files in all the ``packages/`` and make sure they are being added to the ``PACKAGE_DATA`` in the corresponding ``setup.py`` file.
- [x] Fix galaxy-data importing a couple ``tool_shed/`` utilities. Possibly refactor ``tool_shed/util/common_util``, ``tool_shed/util/encoding_util``, and ``tool_shed/util/common_util`` out into ``galaxy-util``. Possibly refactor ``galaxy.model.tool_shed_install`` out of ``galaxy-data``. Possibly refactor the model methods that require the ``common_util`` import into ``galaxy-app``.
- [x] Make sure all unit tests find a home. Pick a unit test not symbolically linked in, find a home, make sure the dependencies are correct and it works on CI.
- [x] Refactor ``galaxy.quota`` to not consume ``trans`` so it belongs more neatly in ``galaxy-data``. The methods consume ``trans`` also take in a user and and a history - so the idea would be anywhere they are called with ``agent.f(trans=trans)``  refactor by adding a method to ``trans`` so it would be ``def trans.f(): agent(user=self.user, history=self.history)`` for better encapsulation. 
- [ ] Create a ``Pipfile`` for each package - hard dependencies should be tracked as close as possible to the code, write some infrastructure for merging them and building our main ``Pipfile`` from these.
[...]

Issue #8073, opened 2019-05-31
Modules to be moved to `galaxy-tool-util` package


nsoranzo:
These modules are needed by Planemo and are currently provided by `galaxy-lib`, but are not yet in `galaxy-tool-util`:

- [x] `galaxy.tools.lint` #8276
- [x] `galaxy.tools.linters` #8276
- [x] `galaxy.tools.loader_directory`
- [ ] `galaxy.tools.verify` #8284

Plus it would be nice to have the `galaxy.xsd` file presently found in:

- [x] `lib/galaxy/tools/xsd/` https://github.com/galaxyproject/galaxy/pull/8098

PR #8276, opened 2019-07-06
Move linters from tools into tool_util

jdavcs:
Also, update setup.py for tool_util package.

Ref: https://github.com/galaxyproject/galaxy/issues/8240

Note: `make updaterst` has not been run; waiting on
https://github.com/galaxyproject/galaxy/pull/8270

Ping @jmchilton


nsoranzo:
The same needs to be done for `galaxy.tools.verify` (see https://github.com/galaxyproject/galaxy/issues/8073), would you like to do that too as part of this PR?
It's quite self-contained, so having a separate one is absolutely fine if you prefer/don't have time.

nsoranzo (approved):

jdavcs:
@nsoranzo Yes, can do. Working on it, will add to this PR. (correction: "verify" >> separate PR)

Issue #8262, opened 2019-07-04
Tools Panel - Vue Component


inkuzmin:
We've discussed this task with @bgruening during the ELIXIR All Hands meeting in Lisbon and with @martenson during the third day of GCC2019 in Freiburg.

The idea is twofold:

1. To create a vue component for the tool panel.
2. To provide a better search results view.
2.1. To expand the panel with results over the middle panel.
2.2. To sort results by score.

Right now I am going to prototype it. Any comments or suggestions will be welcomed!

dannon:
Maybe pursue this as a hackathon project and loop in a few more folks to help?  This sounds like an ideal project for that.

It'd be great if this could unify the toolbox components between the standard interface and the workflow editor in the process.

PR #8283, opened 2019-07-06
Some work on the tool menu styles.

jxtx:
Highlighting the background of items was added at some point, but did
not extend to fill the entire width and was not applied to the section
elements. This attempts to make things a bit more visually appealing
and consistent.

Additionally there are some debatable changes:

- Removed the indent for items, I think it is clear from the size and
  weight changes
- Remove underlines, the background highlight is an equally good
  indicator that these are clickable
- Top level labels are now large like the sections, second level
  returns to the default font size

tool_menu_styles

### DIFF
--- a/client/galaxy/docs/bootstrap.md
+++ b/client/galaxy/docs/bootstrap.md
@@ -89,7 +89,6 @@ Disabled
        </div>
      </div>
    </div>
- </div>
 ```
 
 ```vue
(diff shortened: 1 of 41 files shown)


jxtx:
Problems in the workflow editor. Working on it.

martenson:
I like the changes, one concern I have is the reduction of real estate we have, especially for showing search results which is already pretty bad. could be remedied by implementation of https://github.com/galaxyproject/galaxy/issues/8262

Galaxy_png___100___Galaxy__RGB_8___

jxtx:
@martenson Can certainly drop the y padding by some. I don't feel too strongly. This seems more consistent with the overall padding elsewhere, but not perfect.

I think I will work on #8262 tomorrow.

PR #8285, opened 2019-07-07
Some work on tool menu styles

jxtx:
Highlighting the background of items was added at some point, but did
not extend to fill the entire width and was not applied to the section
elements. This attempts to make things a bit more visually appealing
and consistent.

Additionally there are some debatable changes:

Removed the indent for items, I think it is clear from the size and
weight changes
Remove underlines, the background highlight is an equally good
indicator that these are clickable
Top level labels are now large like the sections, second level
returns to the default font size

(This PR previously murdered by a force push)

tool_menu_styles

### DIFF
--- a/client/galaxy/scripts/entry/panels/tool-panel.js
+++ b/client/galaxy/scripts/entry/panels/tool-panel.js
@@ -141,9 +141,9 @@ const ToolPanel = Backbone.View.extend({
                     <div class="toolSectionPad"/>
                     <div class="toolSectionPad"/>
                     <div class="toolSectionTitle" id="title_XXinternalXXworkflow">
-                        <span>
+                        <a>
                             ${_l("Workflows")}
-                        </span>
+                        </a>
                     </div>
                         <div id="internal-workflows" class="toolSectionBody">
                             <div class="toolSectionBg"/>
--- a/client/galaxy/style/scss/base.scss
+++ b/client/galaxy/style/scss/base.scss
@@ -1483,10 +1483,6 @@ div.permissionContainer {
     min-height: 100%;
 }
 
-div.toolSectionWrapper {
-    @extend .mb-1;
-}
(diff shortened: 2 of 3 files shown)


jxtx:
@martenson I tightened up the spacing a bit without needing to reduce the click area. How does this look to you?

(Sorry, I had to move this to a branch and clobber the old PR).

martenson:
Thanks for adjusting that, the height difference seems negligible now.

At the moment we have 4 different font combinations (screenshot), do you think it would be helpful to drop the 500 font-weight from the tool name to the standard 400. We could underline on hover if highlighting the name is important.

Galaxy

jxtx:
Really trying to avoid underline since we already have the background changing. Certainly could drop the weight to 400. I suppose the distinction between title and description doesn't really matter to the user...

guerler:
I like the consistent highlighting but I wonder if removing the indention for tools within sections is a good idea. It seemed easier to see which tools belong to which section with the indention. Maybe we can get an additional opinion on that @martenson. Other than that I noticed that tools which are added to the tool panel without sections appear without left-margin.

Screen Shot 2019-07-08 at 12 45 40 PM

martenson:
@guerler I don't have a strong opinion on it, I really like how our history menu ended up if that helps. 
Galaxy___Test

Good point about the tools without a section.

nsoranzo:
I'd prefer to keep the indentation.

jxtx:
Indentation is back (for now), items outside sections are fixed.

guerler (approved):

nsoranzo:
> I suppose the distinction between title and description doesn't really matter to the user...

I've missed this, actually I think it would be good to keep some visual differentiation between the tool name and description (e.g. tool name in bold), given that for some tools the description is a separate sentence.

PR #8293, opened 2019-07-07
package-ify top-level Python files in lib/galaxy.

jmchilton:


lib/galaxy/app.py
```diff
@@ -11,6 +11,7 @@
 import galaxy.quota
 import galaxy.security
 from galaxy import config, job_metrics, jobs
+from galaxy.config_watchers import ConfigWatchers
```
natefoo, line 14: Should this maybe be `galaxy.config.watchers` after #921?

natefoo (approved):

lib/galaxy/app.py
natefoo: Note to future self: "yes" -@jmchilton

PR #8294, opened 2019-07-07
Library date propagate

bwlang:
This propagates leaf entry update times  up  the tree to top level libraries.
With this addition users will be able to sort by date in top level library folders to quickly reach the most recently updated folders

### DIFF
--- a/lib/galaxy/managers/library_datasets.py
+++ b/lib/galaxy/managers/library_datasets.py
@@ -101,7 +101,7 @@ class LibraryDatasetsManager(datasets.DatasetAssociationManager):
             ldda.dbkey = new_genome_build
             changed = True
         if changed:
-            ldda.updateParentFolderUpdateTimes(trans)
+            ldda.updateParentFolderUpdateTimes()
             trans.sa_session.add(ldda)
             trans.sa_session.flush()
         return changed
--- a/lib/galaxy/managers/library_datasets.py
+++ b/lib/galaxy/managers/library_datasets.py
@@ -101,7 +101,7 @@ class LibraryDatasetsManager(datasets.DatasetAssociationManager):
             ldda.dbkey = new_genome_build
             changed = True
         if changed:
-            ldda.updateParentFolderUpdateTimes()
+            ldda.update_parent_folder_update_times()
             trans.sa_session.add(ldda)
             trans.sa_session.flush()
         return changed
--- a/lib/galaxy/managers/library_datasets.py
+++ b/lib/galaxy/managers/library_datasets.py
@@ -101,7 +101,7 @@ class LibraryDatasetsManager(datasets.DatasetAssociationManager):
             ldda.dbkey = new_genome_build
             changed = True
         if changed:
-            ldda.updateParentFolderUpdateTimes()
+            ldda.updateParentFolderUpdateTimes(trans)
             trans.sa_session.add(ldda)
             trans.sa_session.flush()
         return changed
--- a/lib/galaxy/managers/library_datasets.py
+++ b/lib/galaxy/managers/library_datasets.py
@@ -101,6 +101,7 @@ class LibraryDatasetsManager(datasets.DatasetAssociationManager):
             ldda.dbkey = new_genome_build
             changed = True
         if changed:
+            ldda.updateParentFolderUpdateTimes()
             trans.sa_session.add(ldda)
             trans.sa_session.flush()
         return changed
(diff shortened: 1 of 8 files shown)


bwlang:
@martenson : any ideas about why time updates are not sticking?

lib/galaxy/model/__init__.py
```diff
@@ -3592,6 +3594,29 @@ def to_dict(self, view='collection'):
             rval['metadata_' + name] = val
         return rval
 
+    def updateParentFolderUpdateTimes(self):
+        # sets the update_time for all continaing folders up the tree
+        ldda = self
+        db_session = object_session(ldda)
```
martenson: try adding
```
db_session.expire_all()
db_session.flush()
```

here to make sure that something in memory is not overriding the objects. Idea of @jmchilton

martenson: no dice, says Brad

martenson:
Just to capture what we've learned for the future generations: Some text statements do not look like sql to alchemy and hence it does not autocommit them hence https://github.com/galaxyproject/galaxy/pull/8294/commits/6cde398e6287207a7f3652fee43fcdf6d1c63bd7 is needed.

>Note that SQLAlchemy’s usual “autocommit” behavior applies to text() constructs implicitly - that is, statements which begin with a phrase such as INSERT, UPDATE, DELETE, or a variety of other phrases specific to certain backends, will be eligible for autocommit if no transaction is in progress.

https://docs.sqlalchemy.org/en/13/core/sqlelement.html#sqlalchemy.sql.expression.text
https://docs.sqlalchemy.org/en/13/core/connections.html#sqlalchemy.engine.Connection.execution_options.params.autocommit

bwlang:
@martenson 
not sure what's going on with the test here - it's failing with an error saying that the SQL is invalid. but when i try via python-i scripts/db_shell.py like this:
```
>>> res = sa_session.execute('''WITH RECURSIVE parent_folders_of(folder_id) AS
...                 (SELECT folder_id
...                 FROM library_dataset
...                 WHERE id = 1
...                 UNION ALL SELECT library_folder.parent_id
...                 FROM library_folder,
...                         parent_folders_of
...                 WHERE library_folder.id = parent_folders_of.folder_id )
...                 UPDATE library_folder
...                 SET update_time =
...                 (SELECT update_time
...                 FROM library_dataset_dataset_association
...                 WHERE id = 1)
...                 FROM parent_folders_of
...                 WHERE library_folder.id = parent_folders_of.folder_id''')
```
I get:
```
res.rowcount
1
```
Any ideas?

martenson:
@bwlang it might not be sqlite compatible :/

bwlang:
these top 2 failures look like jenkins issues to me... please let me know if there is something I should fix.

lib/galaxy/model/__init__.py
```diff
@@ -3592,6 +3594,32 @@ def to_dict(self, view='collection'):
             rval['metadata_' + name] = val
         return rval
 
+    def updateParentFolderUpdateTimes(self):
```
nsoranzo:
```suggestion
    def update_parent_folder_update_times(self):
```

lib/galaxy/webapps/base/controller.py
```diff
@@ -546,6 +546,7 @@ def _copy_hda_to_library_folder(self, trans, hda_manager, from_hda_id, folder_id
         # I don't see a reason why hdas copied into libraries should not be visible.
         # If there is, refactor `ldda.visible = True` to do this only when adding HDCAs.
         ldda.visible = True
+        ldda.updateParentFolderUpdateTimes()
```
nsoranzo:
```suggestion
        ldda.update_parent_folder_update_times()
```

lib/galaxy/managers/library_datasets.py
```diff
@@ -101,6 +101,7 @@ def _set_from_dict(self, trans, ldda, new_data):
             ldda.dbkey = new_genome_build
             changed = True
         if changed:
+            ldda.updateParentFolderUpdateTimes()
```
nsoranzo:
```suggestion
            ldda.update_parent_folder_update_times()
```

jmchilton (approved):

martenson:
A very nice PR, thank you @bwlang !

bwlang:
@martenson : Thanks for your help with getting this working and the other lib features.  I think we decreased library entropy a bit last week :)

PR #8300, opened 2019-07-08
Skip migrations on entirely clean new databases

mvdbeek:
This should be a very nice speedup for ansible tutorials, tests, planemo etc. ... and there'd be less reasons to even include a prebuilt database with docker-galaxy-stables, kubernetes etc, allowing even smaller images.

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -26,6 +26,8 @@ from six.moves import configparser
 
 from galaxy.containers import parse_containers_config
 from galaxy.exceptions import ConfigurationError
+from galaxy.model import mapping
+from galaxy.model.tool_shed_install.migrate.check import create_or_verify_database as tsi_create_or_verify_database
 from galaxy.tool_util.deps.container_resolvers.mulled import DEFAULT_CHANNELS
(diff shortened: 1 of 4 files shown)


mvdbeek:
And of course downgrades/upgrades continue to work as usual, even if the initial database was created without going through each and every single migration.

jmchilton:
I suppose we should rebuild the Docker image without a pre-migrated database now?

mvdbeek:
Yep, I guess all projects that ship a Galaxy database should update. I already started modernizing the testing image in https://github.com/mvdbeek/galaxy/commit/2a0b24cdd9484c52f0a02e858f01cfcb3bd2da82
with the stretch goal of kicking all the unneeded stuff out and to have a two-stage build with an isolated virtualenv container (basically just installing requirements.txt + requirements-dev.txt), which should avoid having to chown the virtualenv folder on startup. With shorter startup times we could split the tests into smaller pieces and get faster turnaround (hopefully ...).

natefoo:
How did I miss this?!? :heart_eyes:

martenson:
@natefoo It's marvelous right? I feel dumb for not doing it earlier.

Issue #930, opened 2015-10-16
purge user to permanently delete all associated objects


scholtalbers:
incl. visualizations.
There is a button to purge users after delete, but it does not permanently delete histories, nor does it touch the visualizations. 
What I'm looking for is a convenient way to 'reset' users, like those accounts used for training and I could do that if the purge_user is fully removing everything. After successful cleanup, I would undelete the user again when required.

PR #8299, opened 2019-07-08
Implement undelete user function with api tests

juleengraham:

Issue #3843, opened 2017-03-29
API delete group


bernt-matthias:
I would like to delete a group via the API (actually via bioblend) but I was told (https://github.com/galaxyproject/bioblend/issues/216) that this is currently not supported by the API.

martenson:
The title states `delete user` is it user deletion, group deletion, or group modification you are interested in?

bernt-matthias:
Sorry for the confusion. I mean group deletion. User deletion and group modification are already supported by bioblend.

PR #8309, opened 2019-07-08
Implement undelete and purge API for users

cganote:
Refactored the controller vs. manager for the deletion and purge of users. Purge now wipes out the username (regardless of gdrp setting) and takes out histories. Undelete should now work.

### DIFF
--- a/lib/galaxy/managers/users.py
+++ b/lib/galaxy/managers/users.py
@@ -114,19 +114,16 @@ class UserManager(base.ModelManager, deletable.PurgableManagerMixin):
         if not self.app.config.allow_user_deletion:
             raise exceptions.ConfigDoesNotAllowException('The configuration of this Galaxy instance does not allow admins to delete/undelete users.')
         # Check first if it is already deleted. If so, say so and do nothing.
-        # Unsure if we should add username to the log.debug or if that breaks the gdrp support
+        # Unsure if we should add username to the log.debug or if that breaks the GDPR support
         if user.deleted:
             log.debug("User already deleted, but setting delete flag just in case.")
         if user.purged:
             log.debug("Cannot delete purged user, but setting delete flag just in case.")
-
         user.deleted = True
-        log.debug("User %s deleted" % user.username)
         self.session().add(user)
         self.session().flush()
 
     def undelete(self, user):
-        log.debug("Requested undelete of user %s" % user.username)
         if not self.app.config.allow_user_deletion:
             raise exceptions.ConfigDoesNotAllowException('The configuration of this Galaxy instance does not allow admins to delete/undelete users.')
         if user.purged:
@@ -138,13 +135,10 @@ class UserManager(base.ModelManager, deletable.PurgableManagerMixin):
         self.session().flush()
 
     def purge(self, user):
-        log.debug("Requested purge of user %s" % user)
         if not self.app.config.allow_user_deletion:
             raise exceptions.ConfigDoesNotAllowException('The configuration of this Galaxy instance does not allow admins to delete/undelete users.')
         if not user.deleted:
             raise exceptions.MessageException('User \'%s\' has not been deleted, so it cannot be purged.' % user.email, 'error')
-        else:
-            log.debug("Purging user %s" % user)
         private_role = self.app.security_agent.get_private_user_role(user)
         # Delete History
         for active_history in user.active_histories:
@@ -196,7 +190,6 @@ class UserManager(base.ModelManager, deletable.PurgableManagerMixin):
                 role.description = role.description.replace(user.email, email_hash)
             user.email = email_hash
             user.username = uname_hash
-        log.debug("User credentials changed. email is now %s and name is %s" % (user.email, user.username))
         # Redact user addresses as well
         if self.app.config.redact_user_address_during_deletion:
             user_addresses = self.session().query(self.app.model.UserAddress) \
--- a/lib/galaxy/managers/users.py
+++ b/lib/galaxy/managers/users.py
@@ -4,7 +4,9 @@ Manager and Serializer for Users.
 import logging
 import random
 import socket
+import time
 from datetime import datetime
+from galaxy.util.hash_util import new_secure_hash
 
 from markupsafe import escape
 from sqlalchemy import and_, desc, exc, func, true
@@ -36,7 +38,6 @@ you may want to notify an administrator.
 If you're having trouble using the link when clicking it from email client, you
 can also copy and paste it into your browser.
 """
-
 
 class UserManager(base.ModelManager, deletable.PurgableManagerMixin):
     foreign_key_name = 'user'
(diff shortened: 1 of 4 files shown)


martenson:
@galaxybot test this

martenson:
@galaxybot test this

hexylena:
> Purge now wipes out the username...and takes out histories.

:tada::heart::tada::heart::tada::heart:

cganote (commented):
Great, thank you!

martenson:
just now noticed https://github.com/galaxyproject/galaxy/pull/8299 :/

martenson:
I cherry picked the test commits from https://github.com/galaxyproject/galaxy/pull/8299 and fixed them. This PR now supersedes the other one.

martenson:
@galaxybot test this

hexylena:
Fixes https://github.com/galaxyproject/galaxy/issues/930

lib/galaxy/managers/users.py
```diff
@@ -109,10 +111,108 @@ def create(self, email=None, username=None, password=None, **kwargs):
...
+            raise exceptions.ConfigDoesNotAllowException('The configuration of this Galaxy instance does not allow admins to delete users.')
+        # Check first if it is already deleted. If so, say so and do nothing.
+        # Unsure if we should add username to the log.debug or if that breaks the GDPR support
+        if user.deleted:
+            log.debug("User already deleted, but setting delete flag just in case.")
+        if user.purged:
+            log.debug("Cannot delete purged user, but setting delete flag just in case.")
         user.deleted = True
         self.session().add(user)
         self.session().flush()
```
nsoranzo: I'd just replace lines 117-125 with:
```python
        super(UserManager, self).delete(user)
```

lib/galaxy/managers/users.py
```diff
@@ -109,10 +111,108 @@ def create(self, email=None, username=None, password=None, **kwargs):
...
+        """Remove the deleted flag for the given user."""
+        if not self.app.config.allow_user_deletion:
+            raise exceptions.ConfigDoesNotAllowException('The configuration of this Galaxy instance does not allow admins to undelete users.')
+        if user.purged:
+            raise exceptions.ItemDeletionException('Purged user cannot be undeleted.')
+        if not user.deleted:
+            log.debug('User \'%s\' has not been deleted, so it cannot be undeleted.' % user.email)
+        user.deleted = False
+        self.session().add(user)
+        self.session().flush()
```
nsoranzo: I'd replace lines 133-137 with:
```python
        super(UserManager, self).undelete(user)
```

lib/galaxy/managers/users.py
```diff
@@ -109,10 +111,108 @@ def create(self, email=None, username=None, password=None, **kwargs):
...
+
+    def purge(self, user):
+        """Purge the given user. They must have the deleted flag already."""
+        if not self.app.config.allow_user_deletion:
+            raise exceptions.ConfigDoesNotAllowException('The configuration of this Galaxy instance does not allow admins to delete or purge users.')
+        if not user.deleted:
+            raise exceptions.MessageException('User \'%s\' has not been deleted, so they cannot be purged.' % user.email)
```
nsoranzo, line 132: I'd remove lines 143-144, purging without deleting first is fine by me.

lib/galaxy/managers/users.py
```diff
@@ -109,10 +111,108 @@ def create(self, email=None, username=None, password=None, **kwargs):
...
+                addr.address = new_secure_hash(addr.address + pseudorandom_value)
+                addr.city = new_secure_hash(addr.city + pseudorandom_value)
+                addr.state = new_secure_hash(addr.state + pseudorandom_value)
+                addr.postal_code = new_secure_hash(addr.postal_code + pseudorandom_value)
+                addr.country = new_secure_hash(addr.country + pseudorandom_value)
+                addr.phone = new_secure_hash(addr.phone + pseudorandom_value)
+                self.session().add(addr)
+
+        self.session().add(user)
+        self.session().flush()
```
nsoranzo: I'd replace lines 166 and 213-214 with:
```python
        super(UserManager, self).purge(user)
```

lib/galaxy/managers/users.py
martenson: I think it is a useful extra safety layer before a destructive function is called. @cganote was it meant like this?

cganote: Yep I was hoping to have a little bit of a safety net

nsoranzo: I don't think this check belongs to the manager.

bernt-matthias:
Any chance to cover this old one by me https://github.com/galaxyproject/galaxy/issues/3843

martenson:
@bernt-matthias I wouldn't tie it to this PR though

bernt-matthias:
> I wouldn't tie it to this PR though

sure

lib/galaxy/managers/users.py
martenson: I think it is better here than in the API controller though. Where would you put it?

martenson:
@galaxybot test this

martenson:
@galaxybot test this

test/api/test_users.py
```diff
@@ -74,6 +77,36 @@ def test_admin_update(self):
...
+        user = self._setup_user(TEST_USER_EMAIL_DELETE)
+        self._delete("users/%s" % user["id"], admin=True)
+        updated_user = self._get("users/deleted/%s" % user['id'], admin=True).json()
+        assert updated_user['deleted'] is True, updated_user
+
+    def test_purge_user(self):
+        """Delete user and then purge them."""
+        user = self._setup_user(TEST_USER_EMAIL_PURGE)
+        response = self._delete("users/%s" % user["id"], admin=True)
+        self._assert_status_code_is(response, 200)
```
nsoranzo:
```suggestion
        self._assert_status_code_is_ok(response)
```

test/api/test_users.py
```diff
@@ -74,6 +77,36 @@ def test_admin_update(self):
...
+        data = dict(purge="True")
+        response = self._delete("users/%s" % user["id"], data=data, admin=True)
+        self._assert_status_code_is(response, 200)
```
nsoranzo:
```suggestion
        self._assert_status_code_is_ok(response)
```

lib/galaxy/managers/users.py
nsoranzo: In the UI (`lib/galaxy/webapps/galaxy/controllers/admin.py` ?), I don't think it's the API business to ask users for confirmation or similar.

lib/galaxy/managers/users.py
```diff
@@ -109,7 +111,91 @@ def create(self, email=None, username=None, password=None, **kwargs):
...
+
+    def undelete(self, user):
+        """Remove the deleted flag for the given user."""
+        if not self.app.config.allow_user_deletion:
+            raise exceptions.ConfigDoesNotAllowException('The configuration of this Galaxy instance does not allow admins to undelete users.')
+        if user.purged:
+            raise exceptions.ItemDeletionException('Purged user cannot be undeleted.')
+        super(UserManager, self).undelete(user)
+
+    def purge(self, user):
```
nsoranzo:
```suggestion
    def purge(self, user, flush=True):
```

lib/galaxy/managers/users.py
```diff
@@ -109,7 +111,91 @@ def create(self, email=None, username=None, password=None, **kwargs):
         return user
 
     def delete(self, user):
```
nsoranzo:
```suggestion
    def delete(self, user, flush=True):
```

lib/galaxy/managers/users.py
```diff
@@ -109,7 +111,91 @@ def create(self, email=None, username=None, password=None, **kwargs):
...
-        user.deleted = True
+        """Mark the given user deleted."""
+        if not self.app.config.allow_user_deletion:
+            raise exceptions.ConfigDoesNotAllowException('The configuration of this Galaxy instance does not allow admins to delete users.')
+        super(UserManager, self).delete(user)
+
+    def undelete(self, user):
```
nsoranzo:
```suggestion
    def undelete(self, user, flush=True):
```

lib/galaxy/managers/users.py
```diff
@@ -109,7 +111,91 @@ def create(self, email=None, username=None, password=None, **kwargs):
         return user
 
     def delete(self, user):
-        user.deleted = True
+        """Mark the given user deleted."""
+        if not self.app.config.allow_user_deletion:
+            raise exceptions.ConfigDoesNotAllowException('The configuration of this Galaxy instance does not allow admins to delete users.')
+        super(UserManager, self).delete(user)
```
nsoranzo:
```suggestion
        super(UserManager, self).delete(user, flush=flush)
```

lib/galaxy/managers/users.py
```diff
@@ -109,7 +111,91 @@ def create(self, email=None, username=None, password=None, **kwargs):
...
+
+    def undelete(self, user):
+        """Remove the deleted flag for the given user."""
+        if not self.app.config.allow_user_deletion:
+            raise exceptions.ConfigDoesNotAllowException('The configuration of this Galaxy instance does not allow admins to undelete users.')
+        if user.purged:
+            raise exceptions.ItemDeletionException('Purged user cannot be undeleted.')
+        super(UserManager, self).undelete(user)
```
nsoranzo:
```suggestion
        super(UserManager, self).undelete(user, flush=flush)
```

lib/galaxy/managers/users.py
```diff
@@ -109,7 +111,91 @@ def create(self, email=None, username=None, password=None, **kwargs):
...
+                addr.institution = new_secure_hash(addr.institution + pseudorandom_value)
+                addr.address = new_secure_hash(addr.address + pseudorandom_value)
+                addr.city = new_secure_hash(addr.city + pseudorandom_value)
+                addr.state = new_secure_hash(addr.state + pseudorandom_value)
+                addr.postal_code = new_secure_hash(addr.postal_code + pseudorandom_value)
+                addr.country = new_secure_hash(addr.country + pseudorandom_value)
+                addr.phone = new_secure_hash(addr.phone + pseudorandom_value)
+                self.session().add(addr)
+        # Purge the user
+        user.purged = True
```
nsoranzo:
```suggestion
        super(UserManager, self).purge(user, flush=flush)
```

lib/galaxy/managers/users.py
```diff
@@ -109,7 +111,91 @@ def create(self, email=None, username=None, password=None, **kwargs):
...
         self.session().add(user)
```
nsoranzo:
```suggestion
```

lib/galaxy/managers/users.py
```diff
@@ -109,7 +111,91 @@ def create(self, email=None, username=None, password=None, **kwargs):
...
         self.session().flush()
```
nsoranzo:
```suggestion
```

lib/galaxy/managers/users.py
martenson: This is not a confirmation, this is a built-in safety measure before a destructive operation happens. I think it is much aligned with how carefully Galaxy handles similar situations.

martenson:
@nsoranzo I think I'm not a fan of going from api to manager then to mixin, then same manager again, then same mixin again for one method. It feels confusing so I'd like to avoid that.

lib/galaxy/managers/users.py
nsoranzo: History purge doesn't have this check in either the manager or the API.

martenson: Purge history is a user-centric operation that has an extra ui confirmation. Purge user is an admin operation, usually way more destructive, and exposed through API. I do not think the extra confirmation belongs just to the UI. I am not opposed to adding a UI check, but this PR does a good job at trying to minimize impact of potential admin errors.

nsoranzo:
> @nsoranzo I think I'm not a fan of going from api to manager then to mixin, then same manager again, then same mixin again for one method. It feels confusing so I'd like to avoid that.

I think these are good abstractions. For reference:
- the API takes care of the communication with the client, checks the inputs and permissions, then invokes the methods of the right manager
- the manager inherits and specialises some methods from superclasses and mixins, ensuring a consistent interface
- the `PurgableManagerMixin` mixin implements a generic `purge()` method, which just calls the `delete()` method and sets the `purged` attribute of the item
- for `UserManager`, we need to specialise both the `delete()` and `purge()` method, inherited from `PurgableManagerMixin`

martenson:
I don't think the path manager>mixin>manager>mixin>done  is better than manager>done but at this point it seems like a matter of opinion and I will defer to yours.

nsoranzo (approved):

hexylena (approved):
Fine for me :)

PR #8312, opened 2019-07-08
[WIP] nglview support other molecular datatypes

chrisbarnettster:
Easy as pie to support these other datatypes, that's already in the commit

Still to do:
* [ ] Gro support. Gro is set to be a tabular file...
* [ ] fix builtin selections which don't always update for PDB and for small molecules. 
* [ ] extend for pv
* [ ] allow for selections - there is an elegant DSL and this function could be brought to the UI
* [ ] allow cumulative representations


bgruening:
@chrisbarnettster thanks a lot for working on this.
Looping in @anuprulez ....

PR #8314, opened 2019-07-08
Fix bug in a unit test

jdavcs:
There was a bug in a context manager inside a unit test. This caused
nondeterministic behavior of some unit tests after a refactoring.

Bug fixed; added 3 unit tests to test the context manager.

Ping @jmchilton @natefoo


natefoo (approved):
Good catch!

test/unit/tools/test_tool_deps.py
```diff
@@ -719,25 +719,69 @@ def _first_conda_resolver_options(dm):
 
 
 @contextmanager
-def __environ(values, remove=[]):
+def __environ(values, keys_to_remove=[]):
```
mvdbeek: I distribute bonus points for eliminating mutable default args, see https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments. The way you use it here it's fine,
but better to just say `keys_to_remove=None` and then 
```
if keys_to_remove is None:
    keys_to_remove = []
```
in the function body.

test/unit/tools/test_tool_deps.py
```diff
@@ -719,25 +719,69 @@ def _first_conda_resolver_options(dm):
...
     """
     Modify the environment for a test, adding/updating values in dict `values` and
     removing any environment variables mentioned in list `remove`.
```
nsoranzo:
```suggestion
    removing any environment variables mentioned in list `keys_to_remove`.
```

test/unit/tools/test_tool_deps.py
```diff
@@ -719,25 +719,69 @@ def _first_conda_resolver_options(dm):
...
+        os.environ must be preserved across calls to __environ
+        when updating existing or adding new key:value pairs
+    """
+    key, val = 'foo_test_678363', '42'
+    os.environ[key] = val
+    new_key, new_val1 = 'foo_test_new', '53'  # add new key + value
+    new_val2 = '43'  # update old key's value
+    to_update = {new_key: new_val1, key: new_val2}
+    with __environ(values=to_update):
+        pass
```
nsoranzo: What about testing that `os.environ` has been updated here?

test/unit/tools/test_tool_deps.py
```diff
@@ -719,25 +719,69 @@ def _first_conda_resolver_options(dm):
...
+    """
+        os.environ must be preserved across calls to __environ
+        when removing key:value pairs
+    """
+    key, val = 'foo_test_678363', '42'
+    os.environ[key] = val
+    new_key, new_val = 'foo_test_new', '53'  # add new key + value
+    to_update = {new_key: new_val}
+    with __environ(values=to_update, keys_to_remove=[key, new_key]):
+        pass
```
nsoranzo: Same here.

test/unit/tools/test_tool_deps.py
jdavcs: @mvdbeek Yes! (and there's no need to create [] in the body: if it's none, we never use it - so a simple check will be sufficient). Thx!

test/unit/tools/test_tool_deps.py
jdavcs: @nsoranzo Of course!

martenson:
switching to wip until the comments are addressed

test/unit/tools/test_tool_deps.py
jdavcs: Done!

test/unit/tools/test_tool_deps.py
jdavcs: Done!

test/unit/tools/test_tool_deps.py
jdavcs: Done.

test/unit/tools/test_tool_deps.py
jdavcs: Done.

jdavcs:
I have made several modifications to the context manager + tests (also, see notes in latest commit). As far as I can tell, the main change to the context manager (no more new_keys) shouldn't cause any issues. Please take a look. @nsoranzo @mvdbeek

jdavcs:
E: Unable to locate package libpython3.5-dev??
Can someone rerun the tests, please?

jmchilton:
Looks good to me, that bug with Python 3.5 is fixed upstream so I wouldn't worry about it. Take out of WIP?

jmchilton (approved):

jdavcs:
@jmchilton thanks! Yes, all good to go then.

jdavcs:
This one is new: OperationalError: (psycopg2.OperationalError) FATAL:  remaining connection slots are reserved for non-replication superuser connections. It's unrelated to this PR (as far as I can judge). Will rerun tests.

jdavcs:
Ready for merging!

nsoranzo (approved):

PR #8319, opened 2019-07-09
Use lowercase true/false in config_schema.yml

nsoranzo:
since these are the only allowed values for booleans in YAML 1.2: https://yaml.org/spec/1.2/spec.html#id2803629

make config-rebuild


mvdbeek (approved):

mvdbeek:
Thanks @nsoranzo !

PR #8321, opened 2019-07-10
Fix kubernetes runner on py3

mvdbeek:
instance_id is a string, so this can't be compared to an integer
on python 3. Broken in ac1b25ed4c069abf1c8d810ba26d5be43717766d,
prior to that the comparison was `len(self._galaxy_instance_id) > 0:`


nsoranzo (approved):

nsoranzo:
I've been pushing for running the py3 version of Jenkins tests by default instead of the py2 ones for a while, I suppose that would have caught this bug...

mvdbeek:
unfortunately not, we would need to target a kubernetes setup in the tests (which might be a smart move anyway, we could have a persistent k8s cluster somewhere).

PR #8322, opened 2019-07-10
Sniff fastqsanger more aggressively

wm75:
For the first time, I've encountered a Phred+33 fastq file with quality scores over 44, which is the current limit accepted by the fastqsanger sniffer.
The resulting failure to detect fastqsanger could have been avoided by checking the lower bound of quality scores like in this PR.
I decided to leave the 44/M limit unchanged in order to avoid possible false detection of low-quality +64
reads.


lib/galaxy/datatypes/sequence.py
```diff
@@ -714,6 +714,9 @@ def sniff_prefix(self, file_prefix):
         True
         >>> FastqCSSanger().sniff(fname)
         True
+        >>> fname = get_test_fname('2.fastqsanger')
```
nsoranzo: Can you avoid filenames that were used for files moved to https://github.com/galaxyproject/galaxy-test-data/ ?

```suggestion
        >>> fname = get_test_fname('4.fastqsanger')
```

lib/galaxy/datatypes/sequence.py
```diff
@@ -818,9 +821,24 @@ class FastqSanger(Fastq):
     @staticmethod
     def quality_check(lines):
         """Presuming lines are lines from a fastq file, return True if the qualities are compatible with sanger encoding"""
+        is_ambiguous = True
+        within_upper_bound = True
         for line in islice(lines, 3, None, 4):
-            if not all(_ >= '!' and _ <= 'M' for _ in line[0]) or ' ' in line:
+            if ' ' in line:
```
nsoranzo:
```suggestion
            line0 = line[0]
            if ' ' in line0:
```

lib/galaxy/datatypes/sequence.py
```diff
@@ -818,9 +821,24 @@ class FastqSanger(Fastq):
...
                 return False
+            if is_ambiguous:
+                for q in line[0]:
```
nsoranzo:
```suggestion
                for q in line0:
```

lib/galaxy/datatypes/sequence.py
```diff
@@ -818,9 +821,24 @@ class FastqSanger(Fastq):
...
+                    if q < '!' or q > '~':
+                        return False
+                    if q < ';':
+                        is_ambiguous = False
+                    elif q > 'M':
+                        within_upper_bound = False
+            else:
+                if any(_ < '!' or _ > '~' for _ in line[0]):
```
nsoranzo:
```suggestion
                if any(_ < '!' or _ > '~' for _ in line0):
```

lib/galaxy/datatypes/sequence.py
```diff
@@ -818,9 +821,24 @@ class FastqSanger(Fastq):
...
         for line in islice(lines, 3, None, 4):
-            if not all(_ >= '!' and _ <= 'M' for _ in line[0]) or ' ' in line:
+            if ' ' in line:
                 return False
+            if is_ambiguous:
+                for q in line[0]:
+                    if q < '!' or q > '~':
+                        return False
+                    if q < ';':
+                        is_ambiguous = False
```
nsoranzo:
```suggestion
                        # encoding is Phred+33
                        is_ambiguous = False
```

lib/galaxy/datatypes/sequence.py
```diff
@@ -818,9 +821,24 @@ class FastqSanger(Fastq):
     @staticmethod
     def quality_check(lines):
         """Presuming lines are lines from a fastq file, return True if the qualities are compatible with sanger encoding"""
+        is_ambiguous = True
+        within_upper_bound = True
```
nsoranzo:
```suggestion
        within_typical_upper_bound = True
```

lib/galaxy/datatypes/sequence.py
```diff
@@ -818,9 +821,24 @@ class FastqSanger(Fastq):
...
+            if ' ' in line:
                 return False
+            if is_ambiguous:
+                for q in line[0]:
+                    if q < '!' or q > '~':
+                        return False
+                    if q < ';':
+                        is_ambiguous = False
+                    elif q > 'M':
+                        within_upper_bound = False
```
nsoranzo:
```suggestion
                        within_typical_upper_bound = False
```

lib/galaxy/datatypes/sequence.py
```diff
@@ -818,9 +821,24 @@ class FastqSanger(Fastq):
...
+            else:
+                if any(_ < '!' or _ > '~' for _ in line[0]):
+                    return False
+        if is_ambiguous and not within_upper_bound:
```
nsoranzo:
```suggestion
        if is_ambiguous and not within_typical_upper_bound:
```

lib/galaxy/datatypes/sequence.py
wm75: Ah, that is an interesting point! I have to say I had no clue why @mvdbeek added the `or ' ' in line` in commit https://github.com/galaxyproject/galaxy/commit/a151b63a622424c1f5f21f97169121b4c7bcc5d5 so I thought I better just stick with it. I don't think this is supposed to check for presence of a space in the quality string since `' ' < '!'` anyway.

nsoranzo: Well, `lines` is an array of arrays that comes from `iter_headers(file_prefix, sep='\n', count=1000)`, so each line should contain a single element.

wm75: Yeah, I figured this out after a while, but why would you want to check that this single element is the string `' '`?

mvdbeek:
> I decided to leave the 44/M limit unchanged in order to avoid possible false detection of low-quality +64

That seems like a better fix to me though, we shouldn't care about deprecated quality formats IMO and would make the PR a lot simpler. I think we want to allow more false positives to the most common quality format and keep the check simple

wm75:
Honestly, I don't have a strong opinion on this and, personally, have no interest in legacy fastq formats either. OTOH, the more complex solution avoids false-positives and is likely to be the more robust approach: increasing the upper score limit will likely have to be repeated every once in a while (at least until we finally declare fastq as equal to fastqsanger), but I have yet to see fastqsanger data where no base quality is below 59-33=26.
So, by changing to the more complex code we could get correct detecttion of fastqsanger almost all of the time. At least that's my thought - @dpryan79, do you maybe want to have a look whether idea and implementation make sense?

dpryan79:
I'm :+1: on just declaring all fastq files fastq sanger. It's true for everything produced these days, so if people import some crappy >8 year old data it's not unreasonable to demand they manually specify the format.

wm75:
If such a consensus could be reached anytime soon that would in fact be a great outcome of this PR :smile:
Right now, however, I'm facing the immediate issue that the fastq input for our rewritten exome-seq tutorial doesn't get recognized correctly and confuses people with a warning about them declaring fastqsanger when the data doesn't appear to be of that type. Not exactly ideal for new users :(

lib/galaxy/datatypes/sequence.py
```diff
@@ -818,9 +821,25 @@ class FastqSanger(Fastq):
     @staticmethod
     def quality_check(lines):
         """Presuming lines are lines from a fastq file, return True if the qualities are compatible with sanger encoding"""
+        is_ambiguous = True
+        within_typical_upper_bound = True
         for line in islice(lines, 3, None, 4):
-            if not all(_ >= '!' and _ <= 'M' for _ in line[0]) or ' ' in line:
```
mvdbeek, line 822: So what was the highest value ? wikipedia says 45 for moleculo reads. I guess we could reasonably go until 50 (`S`, right?) and still wouldn't falsely sniff solexa and illumina 1.3-1.5 as fastqsanger (because that should correspond to 24, and the majority of reads should have higher quality bases, falling outside of the range).

wm75: The highest in our training dataset is 46 (`O`). I couldn't find any information about currently used upper limits on standard Illumina platforms (all official Illumina material I found is very old, and I have no idea how up-to-date the Wikipedia article is).
`S` would actually correspond only to 19; Solexa scores start at 59 (`;`), but that would be a score of -5 (according to Wikipedia).
Anyway my idea was to leave the limit at `M` since then there would not be any change in Galaxy's sniffing outcome - except for cases that were clearly wrong detections before (those that must be fastqsanger because they contain ASCII values < 59).
If you think we should raise it to `S` or whatever other limit we can surely do that, too.

mvdbeek: I think we should only raise it to `S`, yes (and thanks for the `-5`, forgot that the scale actually starts there). This is far more lightweight and easier to understand, plus the current logic would fail for reads with high quality bases (i.e `within_typical_upper_bound ` is `False`) that don't also have a low quality base (`is_ambiguous ` is `True`). You'd assume that as sequencing technology improves we'd also have fewer low quality bases, so I don't think this is a good heuristic.

wm75: Ok, so just one last defense of the heuristics before I reduce this to the minimal change:
> plus the current logic would fail for reads with high quality bases (i.e `within_typical_upper_bound ` is `False`) that don't also have a low quality base (`is_ambiguous ` is `True`).

in which case it would behave exactly like the *status quo*! As the title says the proposal would have made detection of fastqsanger *more* robust, not less. It would have used proof for fastqsanger if possible and have guessed in ambiguous situations.

Anyway, lets just move on and make the training material work.

mvdbeek (approved):

lib/galaxy/datatypes/sequence.py
nsoranzo: I think we haven't resolved the doubt about the `' ' in line` check yet: should it be left as it is (I think it's wrong), changed to `' ' in line[0]` or dropped altogether?

mvdbeek: Looks like it probably didn't do what it was supposed to do either way and no one complained, so let's drop it.

nsoranzo (approved):

nsoranzo:
Thanks @wm75 !

bgruening:
@nsoranzo is this something that should be backported?

nsoranzo:
Up to @wm75 , I can help if needed.

wm75:
Thanks @nsoranzo for offering help! We will probably just fix this on EU separately for now (main does not yet have all the necessary updated tools for following the training installed anyway).

Issue #8325, opened 2019-07-11
Data libraries: incorrect permission check for modifying folder name


scholtalbers:
Roles listed under `Roles that can modify this folder` are not able to modify the folder name and description although they should.

This is because the wrong permission check is done, will open PR for the fix.

PR #8326, opened 2019-07-11
check correct permission for allowing a user to modify lib folder

scholtalbers:
Fixes the issue described in #8325

### DIFF
--- a/lib/galaxy/managers/folders.py
+++ b/lib/galaxy/managers/folders.py
@@ -78,6 +78,23 @@ class FolderManager(object):
             folder = self.check_accessible(trans, folder)
         return folder
 
+    def check_modifyable(self, trans, folder):
+        """
+        Check whether the user can modify the folder (name and description).
+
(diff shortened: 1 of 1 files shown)


lib/galaxy/managers/folders.py
```diff
@@ -171,7 +188,7 @@ def update(self, trans, folder, name=None, description=None):
         """
         changed = False
         if not trans.user_is_admin:
-            if not self.check_manageable(trans, folder):
+            if not self.check_modifyable(trans, folder):
```
mvdbeek: The raise below is unreachable, so you could drop this and just do
`self.check_modifyable(trans, folder)`

mvdbeek (approved):

mvdbeek:
@galaxybot test this

lib/galaxy/managers/folders.py
scholtalbers: done

PR #8334, opened 2019-07-14
Update anchor tags and enhance navigation cues for screen readers 

WilliamHolden:
This pull request does two things. The first is that it changes any non link anchor tags to use href="javascript:void(0)" instead of href="#".

Both of these href attributes are trying to achieve the same thing so we should only use one for consistency reasons. There are a few reasons to prefer href="javascript:void(0)" over href="#".

1. href="#" changes the actual URL leading to unwanted effects like the page scrolling to the top 
2. href="#" causes backbone to call an internal checkURL() function which leads to unwanted effects like the issue below:
https://github.com/galaxyproject/galaxy/issues/7893
3. href="#" doesn't work well with screen readers (this can be mitigated by adding role="button")

There are other reasons as well. See the top answer on https://stackoverflow.com/questions/134845/which-href-value-should-i-use-for-javascript-links-or-javascriptvoid0

Ideally, a lint rule will be added to enforce this change in the future but I could not find a good  eslint plugin. The closest one I found was "jsx-a11y/href-no-hash" but I think this rule was created for jsx specifically

The next thing this pull request does is make some enhancements to the navigational cues for screen readers through the use of aria-labels, tabindex, and role attributes. There is still some work to be done here and I will make another pull request with more navigational cue updates.


client/galaxy/scripts/mvc/library/library-dataset-view.js
```diff
@@ -652,7 +652,7 @@ var LibraryDatasetView = Backbone.View.extend({
                 <!-- BREADCRUMBS -->
                 <ol class="breadcrumb">
                     <li class="breadcrumb-item">
-                        <a title="Return to the list of libraries" href="#">Libraries</a>
+                        <a title="Return to the list of libraries" href="javascript:void(0)">Libraries</a>
```
martenson: In some cases, like this one, you are actually removing a link to the 'base' of hashed url, breaking the current functionality (return to the list of libraries).

WilliamHolden: Good catch. I did not realize these were actual links and not buttons. I will fix this. For a scenario like this would you prefer to keep using ```href="#"``` or instead use ```href=""```. Both are valid.

I also now just noticed that there are some places where we are currently using ```href=""``` for buttons and then returning false in the click handler to prevent the href from reloading the page. For consistency reasons, I plan on changing these to use ```javascript:void(0)```.

martenson: I have no preference as long as the functionality is consistent. :)

WilliamHolden: Ok, I don't have a preference either. I kept the ```href="#"``` just in case since we know it currently works. I have updated the PR to fix the issue. I have tested most of the links and they are working. I will still do more testing just to be sure. Let me know if you find any problems  with this updated PR. Thanks!

WilliamHolden:
I have another WIP branch which depends on some of the changes in this PR. When this PR passes review and gets merged I will make the PR for the WIP branch into galaxy dev (when it's not WIP anymore of course). For now I have set up a temporary PR in my own repo to keep track of the changes: https://github.com/WilliamHolden/galaxy/pull/1.

martenson (approved):

martenson:
Thanks @WilliamHolden !

Issue #8330, opened 2019-07-12
[Feature] Split Large Collections with Integer


DarianHole:
Hi all,

I was asked by @Takadonet in regards to him being asked by multiple people from our lab about implementing a new collections tool that breaks apart large collections. This tool will take a collection and an integer and then split the collection into multiple datasets based on the integer.

Ex. A collection of 10 items and an input integer of 5 would produce 2 collections of 5 datasets

I have a baseline code that currently does this on my dev branch and I was wondering what others thoughts were on this situation before I create any PRs and go further with it.

Thanks for your time in advance

bernt-matthias:
What's the rationale of such a splitting operation?

DarianHole:
Hi @bernt-matthias, 

Sorry for the slow response but I believe that some of our users wanted it in cases where they do an analysis on a larger dataset of say 1000 samples and then want to take subsets of that collection to run different tests on and/or compare the subsets instead of downloading the data again.

That or I was asked for it for some phylogenetics but I am not sure in that case what they wanted it for.

I can also create it in the toolshed if you feel that that is more appropriate.

Thanks

bernt-matthias:
Depending on the use case it might be an option to create a nested list. I guess (not sure) this would be useful if splitting is used for separate processing. 

I can't answer if TS is better. 

Anyway, please go ahead and submit a PR here or at the IUC...

DarianHole:
Done pending comments/critiques

PR #8336, opened 2019-07-15
Split Large Collections into Smaller Ones

DarianHole:
From #8330 

## What it Does ##

A small tool that splits large collections into multiple smaller ones based on a maximum integer input

ex. A large collection of 1000 split by 200 would give 5 smaller collections of 200 samples.

## Needs Finished / Needs to be Looked At ##
- Unit Tests --> Fixed/maybe add more??
- ~~Comments on whats happening~~
- ~~Name + Description?~~
- ~~Output name clarity?~~

I'm very new to development so please leave me any feedback on this PR  regarding:
- The overall code
- Any recommendations to add or change (especially regarding names)
- Whether you feel this is better off in the TS (I found collection modification better in the base galaxy code than the TS as I've been looking into both)

Thanks for your time and input!


bernt-matthias:
Maybe add the possibility to specify a fraction. E. G values from 0 to 1.

DarianHole:
That's interesting, an example being that when you specify 0.25 it splits it into 4 collections?

I'll look into it

DarianHole:
I could also switch it from having the input be the maximum number of elements in each collection to the number of collections wanted back if that seems more reasonable/useful

bernt-matthias:
Also thought about this. But the other solution seems more powerful and requires the user only to do a simple calculation to achieve the same, since `1/#collections` is equal to the fraction that the user can enter. BUT: How to get 3 collections, the user can not specify 1/3 as float exactly. 

Maybe have a boolean switch that controls if the entered value is the number of collections or the number of elements.

Alternatively one could also have a conditional having the three options: 1. number of elements. 2. number of collections 3. fraction, but this just blows up the tool form in my opinion.

DarianHole:
I'm working at adding the switch in the xml wrapper to determine if the user wants to get an output based on the number of collections out or the maximum number of elements. I feel that doing it this way is the easiest to understand from a user perspective and allows the flexibility wanted.

Kindof like how the filter from file tool has on to specify if you want to filter based on absence or presence 

Thanks for the comments! It helps a lot!

mvdbeek:
@galaxybot test this

mvdbeek:
@galaxybot test py3

test/api/test_tools.py
```diff
@@ -1867,6 +1867,23 @@ def test_map_over_collection_structured_like(self):
...
+    def test_split_collection_by_number(self):
+        with self.dataset_populator.test_history() as history_id:
+            hdca_id = self.dataset_collection_populator.create_list_in_history(history_id, contents=[("A", "A"), ("B", "B")]).json()['id']
+            self.dataset_populator.wait_for_history(history_id, assert_ok=True)
+            inputs = {
+                "input": {'values': [dict(src="hdca", id=hdca_id)]},
+                "how|how_split": "number_of_collections",
+                "how|num_collecton": 2
+            }
+            generated_collections = self._run("__FILTER_FROM_FILE__", history_id, inputs, assert_ok=True)['generated_collections']
```
mvdbeek: This doesn't look right to me, and that's where the API test is failing in https://jenkins.galaxyproject.org/job/docker-api-py3/246/testReport/junit/test.api.test_tools/ToolsTestCase/test_split_collection_by_number/

lib/galaxy/tools/split_collection_by_number.xml
```diff
@@ -0,0 +1,101 @@
...
+        <test>
+            <param name="input">
+                <collection type="list">
+                    <element name="in1" value="simple_line.txt" />
+                    <element name="in2" value="simple_line_alternative.txt" />
+                </collection>
+            </param>
+            <param name="how_split" value="maximum_elements" />
+            <param name="num_collection" value="1" />
+            <output_collection name="output" type="list">
```
mvdbeek: This is where the tool test is failing (https://jenkins.galaxyproject.org/job/docker-framework-py3/157/testReport/junit/test.functional/test_toolbox_pytest/test_tool___SPLIT_COLLECTION_BY_NUMBER___test_1_/)
I think you may need to use `output_0`, `output_1` etc as parameter name here.

You can run the tool tests in isolation with `run_tests.sh -framework -id __SPLIT_COLLECTION_BY_NUMBER__ `

test/api/test_tools.py
DarianHole: Currently, I cannot pass the input integer to the tool API test:

Trying this should pass the integer in:
```
inputs = {
                "input": {'values': [dict(src="hdca", id=hdca_id)]},
                "how|how_split": "number_of_collections",
                "how|num_collecton": {'values': 2}
            }
            self._run("__SPLIT_COLLECTION_BY_NUMBER__", history_id, inputs, assert_ok=True)
```

Although I get an AssertionError: `Request status code (500) was not expected value 200. Body was {u'err_msg': u'Uncaught exception in exposed API method:', u'err_code': 0` due to:
```
  File "/home/CSCScience.ca/dhole/repos/galaxy/lib/galaxy/webapps/galaxy/api/tools.py", line 576, in _patch_library_inputs
    for index, value in enumerate(v['values']):
TypeError: 'int' object is not iterable
```

If I try to make it a list, it goes back to telling me that:
```
def _report_status_code_error(response, expected_status_code):
        try:
            body = response.json()
        except Exception:
            body = "INVALID JSON RESPONSE <%s>" % response.content
        assertion_message = ASSERT_FAIL_STATUS_CODE % (response.status_code, expected_status_code, body)
>       raise AssertionError(assertion_message)
E       AssertionError: Request status code (400) was not expected value 200. Body was {u'err_data': {u'how|num_collection': u"parameter 'num_collection': the attribute 'value' must be set for non optional parameters"}, u'err_msg': u"parameter 'num_collection': the attribute 'value' must be set for non optional parameters", u'err_code': 0}
```

Any suggestions? Or am I ok with just the wrapper tests that did work with the below changes

mvdbeek: Sorry, I should have mentioned that `'generated_collections'` is not a valid dictionary key returned by `self._run`, your input parameters seem fine.

DarianHole: Oh I'll try it the way that the filter from file tool has it. Although the way I've been currently testing has only `self.run`  and it's been giving me those errors

Edit:
With the `implicit_collections` it still does not like having the input as 2

test/api/test_tools.py
```diff
@@ -1867,6 +1867,23 @@ def test_map_over_collection_structured_like(self):
             assert second_collection_level['collection_type'] == 'list'
             assert second_collection_level['elements'][0]['element_type'] == 'hda'
 
+    @skip_without_tool("__SPLIT_COLLECTION_BY_NUMBER__")
+    def test_split_collection_by_number(self):
+        with self.dataset_populator.test_history() as history_id:
+            hdca_id = self.dataset_collection_populator.create_list_in_history(history_id, contents=[("A", "A"), ("B", "B")]).json()['id']
+            self.dataset_populator.wait_for_history(history_id, assert_ok=True)
+            inputs = {
+                "input": {'values': [dict(src="hdca", id=hdca_id)]},
```
mvdbeek: providing this as a list of dicts is meant for when you run multiple inputs in batch, you can use
```suggestion
                "input": {'values': {'src': 'hdca', 'id': hdca_id}},
```
here.

test/api/test_tools.py
```diff
@@ -1867,6 +1867,23 @@ def test_map_over_collection_structured_like(self):
...
+                "how|how_split": "number_of_collections",
+                "how|num_collecton": 2
+            }
+            generated_collections = self._run("__FILTER_FROM_FILE__", history_id, inputs, assert_ok=True)['generated_collections']
+            output_collection_1, output_collection_2 = generated_collections
+            self.dataset_populator.wait_for_history(history_id, assert_ok=True)
+            assert output_collection_1['collection_type'] == output_collection_2['collection_type'] == 'list:list', output_collection_1
```
mvdbeek: You're splitting a list into multiple lists. Since your source list is of type `list` your output list type will also be `list`.

test/api/test_tools.py
mvdbeek: It really should work though.
Try `generated_collections = self._run("__FILTER_FROM_FILE__", history_id, inputs, assert_ok=True)['output_collections']`, which is probably what you want.

mvdbeek: There's no implicit collections involved at all, these are only created if the input is "mapped over", i.e if you specify the input collection type to `list` but you give a `list:list` you'll get implicit collections.

test/api/test_tools.py
```diff
@@ -1867,6 +1867,23 @@ def test_map_over_collection_structured_like(self):
...
+            hdca_id = self.dataset_collection_populator.create_list_in_history(history_id, contents=[("A", "A"), ("B", "B")]).json()['id']
+            self.dataset_populator.wait_for_history(history_id, assert_ok=True)
+            inputs = {
+                "input": {'values': [dict(src="hdca", id=hdca_id)]},
+                "how|how_split": "number_of_collections",
+                "how|num_collecton": 2
+            }
+            generated_collections = self._run("__FILTER_FROM_FILE__", history_id, inputs, assert_ok=True)['generated_collections']
+            output_collection_1, output_collection_2 = generated_collections
+            self.dataset_populator.wait_for_history(history_id, assert_ok=True)
```
mvdbeek: This wait doesn't do anything, that's implied by `assert_ok=True`
```suggestion
```

DarianHole: I think its something to do with the test framework as all I have is this:
```
def test_split_collection_by_number(self):
        with self.dataset_populator.test_history() as history_id:
            hdca_id = self.dataset_collection_populator.create_list_in_history(history_id, contents=["A", "A", "B", "B"]).json()['id']
            self.dataset_populator.wait_for_history(history_id, assert_ok=True)
            inputs = {
                "input": {'values': [dict(src="hdca", id=hdca_id)]},
                "how|how_split": "number_of_collections",
                "how|num_collecton": {'values': 2}
            }
            generated_collections = self._run("__SPLIT_COLLECTION_BY_NUMBER__", history_id, inputs, assert_ok=True)['output_collections']
```
And the error is coming from _patch_library_inputs as:
```
File "/home/CSCScience.ca/dhole/repos/galaxy/lib/galaxy/webapps/galaxy/api/tools.py", line 576, in _patch_library_inputs
    for index, value in enumerate(v['values']):
TypeError: 'int' object is not iterable
```
The value of 2 is getting input correctly, it just won't let it go through it seems. Making it a list doesn't work in any capacity either from all the ways I've tried it.

Thanks for all the suggestions though, I'll keep playing around with it.

mvdbeek: I can also do a PR against your branch, the tests are passing for me with the suggested changes.

mvdbeek: Have a look at https://github.com/phac-nml/galaxy/pull/4 which fixes the API test

DarianHole: I've realized that I've had num_collecton instead of num_collection this whole time. 

Thanks for the help at it

lib/galaxy/tools/__init__.py
```diff
@@ -3038,12 +3038,89 @@ def produce_outputs(self, trans, out_data, output_collections, incoming, history
...
+
+                    if getattr(dce_object, "history_content_type", None) == "dataset":
+                        copied_value = dce_object.copy(force_flush=False)
+                    else:
+                        copied_value = dce_object.copy()
+                    current_collection_elements[element_identifier] = copied_value
+
+                self._add_datasets_to_history(history, itervalues(current_collection_elements))
+                output_collections.create_collection(
+                    self.outputs["output"], "output_{}".format(i), elements=current_collection_elements
```
mvdbeek, line 3114: The problem here is that you're dynamically creating outputs (`output_0`, `output_1`) etc.
This makes it hard to predict how many outputs will be generated ahead of time, and these dynamically generated outputs won't be connectable in the workflow editor. That doesn't necessarily mean we don't want this, but I just wanted to point this out.

mvdbeek:
@galaxybot test this

DarianHole:
Need to merge in the other tool changes as well

mvdbeek:
@galaxybot test this

DarianHole:
@mvdbeek Thanks again for the help!

lib/galaxy/tools/__init__.py
DarianHole: Yes that is hopefully not a problem, the main goal/factor of this was to allow someone to do an initial analysis on say 200 samples and then split that collection into 4 to run a more computationally demanding analysis on 50 of them without needing to reupload the samples

bernt-matthias:
Just thought about an alternative:

1. Get id's with `collection_element_identifiers`
2. Split the ID file with `split_file_to_collection` into a collection
3. `Filter List from contents of a file` using the original collection and the collection of splitted id's.

Wonder if this does the same.

Even if this is the case, I still consider this tool useful.

DarianHole:
Thanks for the working alternative, I'll pass that along to my labs that have requested this

lib/galaxy/tools/__init__.py
jmchilton: Yeah - I never intended for this to work this way. It is concerning, I'm not sure it is wrong though.

mvdbeek:
@DarianHole how is the alternative working out ? Do you still want to get this merged ?

DarianHole:
@mvdbeek I didn't get too far on an alternative unfortunately with the start of my school semester in September taking over my life. 

If it still works correctly and fits into what the plans are for Galaxy then merge it in. I can also get back to figuring another solution out next week when I'm done class and can look at all the new changes that would potentially make it more feasible.

bernt-matthias:
I have tested this tool a bit: 

- all output collections have the same name, maybe we can come up with something better. Also the input collection is not mentioned: Maybe "Split collection on Collection X: Elements Y-Z"
- there seems to be a bit of code duplication (the for loop creating the current_collection_elements OrderedDict). It should be possible to have the same functionality and implement only one case, e.g. implement only "number of collections wanted" and compute this number for the other case.

Most importantly we should think about the problem how this tool could be used in workflows @jmchilton @mvdbeek could nested collections help here: i.e. just create one collection that contains the splitted collections.

@DarianHole  I have rebased the branch and added a little fix (odict does not exist anymore). If you give me write access I can push it directly to your branch.

DarianHole:
@bernt-matthias, I'm sorry but I don't think I have administrator privileges as I cannot access any of the security settings. I can pull into it though

bernt-matthias:
In your fork you should be able to give me push permission, otherwise I would only prepare a PR for my singe commit and let you do the rebasing.

DarianHole:
The fork that this was done on is the organization that I worked for in the summer. I do not have access to its security to change any permissions from what I can gather from looking at all of the help pages

bernt-matthias:
OK. Got it. https://github.com/phac-nml/galaxy/pull/5

bernt-matthias:
Btw. my previous comment on the usability also applies to my suggested alternative workflow.

PR #8337, opened 2019-07-15
test removing uwsgi weird python package from circleci

martenson:


martenson:
circleci does not need libpython-dev because it compiles its own python for the image which includes it

ready for review

nsoranzo (approved):

PR #8290, opened 2019-07-07
Don't show the pip progress bar when installing under CI

natefoo:
The CircleCI output is currently full of pip progress bars, this should greatly reduce the output.

PR #8343, opened 2019-07-16
Fix use of ``PIP_PROGRESS_BAR`` env variable

nsoranzo:
`PIP_PROGRESS_BAR` environment variable is used by `pip` to set the value for the `--progress-bar` option. It seems that specifying it also only the command line does not override it in some configurations, e.g. on TravisCI Linux builds of BioBlend, see e.g. https://travis-ci.org/galaxyproject/bioblend/jobs/558857557

Follow-up on https://github.com/galaxyproject/galaxy/pull/8290 , fix issue reported in https://github.com/galaxyproject/galaxy/pull/8290#issuecomment-509860833 .

### DIFF
--- a/packages/test.sh
+++ b/packages/test.sh
@@ -3,8 +3,7 @@
 set -e
 
 # Don't display the pip progress bar when running under CI
-PIP_PROGRESS_BAR=
-[ "$CI" = "true" ] && PIP_PROGRESS_BAR='--progress-bar off'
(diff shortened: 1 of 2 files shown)


natefoo (approved):
I didn't even know [this](https://pip.pypa.io/en/stable/user_guide/#environment-variables) was a thing when I picked `$PIP_PROGRESS_BAR`, heh. Nice improvement, thanks!

PR #8344, opened 2019-07-16
[WIP] sniffing improvements for galaxy-data package

martenson:
the sniffer order is required for its functionality

### DIFF
--- a/packages/data/config/datatypes_conf.xml.sample
+++ b/packages/data/config/datatypes_conf.xml.sample
@@ -1,0 +1,1 @@
+../../../lib/galaxy/config/sample/datatypes_conf.xml.sample
--- a/lib/galaxy/model/store/discover.py
+++ b/lib/galaxy/model/store/discover.py
@@ -14,6 +14,7 @@ import six
 
 import galaxy.model
 from galaxy import util
+from galaxy.datatypes import registry, sniff
 from galaxy.exceptions import (
     RequestParameterInvalidException
 )
@@ -512,6 +513,12 @@ def persist_hdas(elements, model_persistence_context):
                 fields_match = discovered_file.match
                 designation = fields_match.designation
                 ext = fields_match.ext
+                if not ext or ext == 'auto':
+                    r = registry.example_datatype_registry_for_sample()
+                    try:
+                        ext = sniff.guess_ext(discovered_file[0], r.sniff_order)
+                    except Exception:
+                        log.warning("Failed to sniff extension for dataset [%s]", discovered_file[0])
                 dbkey = fields_match.dbkey
                 info = element.get("info", None)
                 link_data = discovered_file.match.link_data
--- a/packages/data/MANIFEST.in
+++ b/packages/data/MANIFEST.in
@@ -1,3 +1,4 @@
 include *.rst LICENSE
 include galaxy/datatypes/test/*
+include galaxy/config/datatypes_conf.xml.sample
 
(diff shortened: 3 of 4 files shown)


packages/data/config/datatypes_conf.xml.sample
```diff
@@ -0,0 +1 @@
+../../../lib/galaxy/config/sample/datatypes_conf.xml.sample
```
jmchilton, line 1: You're going to need to put this in either the MANIFEST or package data for setup.py also I think.

martenson: where do you think it makes more sense?

jmchilton: setup.py I think

martenson: So from I've read it seems to be needed in both places. Please correct me if I'm wrong.

nsoranzo: > So from I've read it seems to be needed in both places. Please correct me if I'm wrong.

Both is the safest, see https://setuptools.readthedocs.io/en/latest/setuptools.html#including-data-files

nsoranzo:
> the sniffer order is required for its functionality

Where exactly? For the only place where I found it mentioned (`packages/data/galaxy/datatypes/registry.py`), I don't think this will work as it is.

martenson:
@nsoranzo Without the canonical source of sniffing order the results will be different. I think the registry needs to be created using the sample config in order to match sniffing in Galaxy.

packages/data/config/datatypes_conf.xml.sample
nsoranzo: I think this file should be `packages/data/galaxy/config/sample/datatypes_conf.xml.sample`

nsoranzo:
I think we need to symlink the unit tests from `test/unit/datatypes/` to see if this works.

martenson:
@nsoranzo if what works?

jmchilton (approved):

nsoranzo:
The position of the symlink file inside packages/data/ .

martenson:
> The position of the symlink file inside packages/data/ .

I chose that arbitrarily. I doubt it will work with the `example_datatype_registry_for_sample` method if that is what you meant.

nsoranzo:
> > The position of the symlink file inside packages/data/ .
> 
> I chose that arbitrarily.

In https://github.com/galaxyproject/galaxy/pull/8344#discussion_r305075157 I proposed `packages/data/galaxy/config/sample/datatypes_conf.xml.sample` which is the same place where the `galaxy-app` package would place it, see https://github.com/galaxyproject/galaxy/pull/8357 . At this point I'm not sure which package(s) should provide this and the other sample files.

> I doubt it will work with the `example_datatype_registry_for_sample` method if that is what you meant.

Yes, `example_datatype_registry_for_sample` needs to be fixed and other tests will fail when properly tested, see https://github.com/galaxyproject/galaxy/pull/8358 .

martenson:
`example_datatype_registry_for_sample` works when moved in the proper place but also needs to be modified to have the correct path to the folder with converters (defaults to `converters_path="lib/galaxy/datatypes/converters"`)

martenson:
I switched this to WIP since I'd like to expand it to include sniffing functionality when building Galaxy objects externally.

The configuration of samples is still troublesome though, how do we differentiate between paths in Galaxy proper and paths when this is a package? Galaxy proper currently has `sample_conf = os.path.join(galaxy_dir, "lib", "galaxy", "config", "sample", "datatypes_conf.xml.sample")`

mvdbeek:
I guess a proper config package is needed anyway, this would then be the canonical source for package installed and source installed galaxy

PR #8279, opened 2019-07-06
[WIP] Restructure unit tests for tool_util package

jdavcs:
Ref: https://github.com/galaxyproject/galaxy/issues/8240

PR #8345, opened 2019-07-17
Restructure unit tests for tool_util package

jdavcs:
This replaces a previous (now closed) PR:
https://github.com/galaxyproject/galaxy/pull/8279

Ref: https://github.com/galaxyproject/galaxy/issues/8240

A few changes compared to prev. PR:
- tests for code in galaxy/tool_util moved under test/unit/tool_util
- common parts (used in both tools and tool_util) factored out into
  unittest_utils/sample_data.py
- symlink to sample_data.py in test/unit/tool_util/ is needed to make
  the tool_util package tests run. This is not an ideal solution, but it
  avoids duplication of the sample data used by tests in tools/ and
  tool_util/.


jmchilton:
Awesome - thanks so much!

PR #8351, opened 2019-07-18
Add links to admin entry page

jdavcs:
This is a result of a whim. I had to look for a link on the admin main page and clicked on what I expected to be a link. It wasn't. So here's a quick fix.

This is a temporary solution. Ideally, this should be generated using the same logic that is used for generating the links on the left menu panel (that's currently done w/Backbone). And no page reloads, of course.

I'm undecided on whether this is an improvement or not: the links, I think, improve the user experience; BUT clicking them reloads the page, whereas clicking the same links in the left menu doesn't. Which is putting emphasis on inconsistency. But then we have the user preferences page, which is a similar kind of page, and it includes such links.  

Ping @martenson @bgruening


martenson:
Good quick improvement, thanks!

PR #8359, opened 2019-07-19
catch exceptions occuring in expression validators

bernt-matthias:
rationale: in any case the user should not be bothered with python exceptions, but they are still of interest for admins and tool developers.

Hence I suggest in case of any exception to raise a value Error with the validator's message (hence the user will see the message intended by the validator) and to write an error to the log for admins and tool
developers.

Still does not help in this use case where two validators check if the input is a list of integers within a range: 

```
<validator type="regex" message="a comma separated list of integer values is required"><![CDATA[[+-]?[0-9]+(,[+-]?[0-9]+)*]]></validator>
      <validator type="expression" message="a comma separated list of integer values in the range 0:100 is required"><![CDATA[len(value.split(',')) == len([_ for _ in value.split(',') if 0 <= int(_) <= 100])
]]></validator>
```

Maybe we should log to `debug` and maybe include the StackTrace in the log, but this might require an additional import. Opinions?

### DIFF
--- a/lib/galaxy/tools/parameters/validation.py
+++ b/lib/galaxy/tools/parameters/validation.py
@@ -113,10 +113,15 @@ class ExpressionValidator(Validator):
         self.expression = compile(expression, '<string>', 'eval')
 
     def validate(self, value, trans=None):
-        if not(eval(self.expression, dict(value=value))):
-            message = self.message
-            if self.substitute_value_in_message:
-                message = message % value
+        message = self.message
+        if self.substitute_value_in_message:
+            message = message % value
+        try:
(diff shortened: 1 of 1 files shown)


jmchilton:
Can you change to debug and then I will merge.

natefoo:
If you want to include the stack trace with `debug` logging, you could set the `exc_info` kwarg as described [in the docs](https://docs.python.org/2.7/library/logging.html#logging.Logger.debug).

bernt-matthias:
@natefoo thanks for the tip

@jmchilton changed as requested. indeed debug is better.

Issue #8341, opened 2019-07-16
xlsx format is not recognize


FredericBGA:
on galaxy 19_05 we can't upload an xlsx file and let galaxy sniff the datatype.
it works when you set the datatype yourself.

the errors on galaxy.eu and galaxy.org are different...
I have the same error as galaxy.org on my instance.

galaxy.eu
> Fatal error: Exit code 1 ()
> [Errno 2] No such file or directory: SafeStringWrapper(unicode:__lt__class __sq__galaxy.tools.wrappers.ToolParameterValueWrapper__sq____gt__,__lt__class __sq__galaxy.util.object_wrapper.SafeStringWrapper__sq____gt__,__lt__class __sq__numbers.Number__sq___

galaxy.org:
> Traceback (most recent call last):
>   File "/cvmfs/main.galaxyproject.org/galaxy/tools/data_source/upload.py", line 326, in <module>
>     __main__()
>   File "/cvmfs/main.galaxyproject.org/galaxy/tools/data_source/upload.py", line 319, in __main__
>     metadata.append(add_file(dataset, registry, output_path))
>   File "/cvmfs/main.galaxyproject.org/galaxy/tools/data_source/upload.py", line 128, in add_file
>     convert_spaces_to_tabs=dataset.space_to_tab,
>   File "/cvmfs/main.galaxyproject.org/galaxy/lib/galaxy/datatypes/upload_util.py", line 54, in handle_upload
>     convert_spaces_to_tabs=convert_spaces_to_tabs,
>   File "/cvmfs/main.galaxyproject.org/galaxy/lib/galaxy/datatypes/sniff.py", line 724, in handle_uploaded_dataset_file_internal
>     auto_decompress=auto_decompress,
>   File "/cvmfs/main.galaxyproject.org/galaxy/lib/galaxy/datatypes/sniff.py", line 663, in handle_compressed_file
>     sniffed_ext = run_sniffers_raw(filename, sniff_datatypes)
>   File "/cvmfs/main.galaxyproject.org/galaxy/lib/galaxy/datatypes/sniff.py", line 482, in run_sniffers_raw
>     file_prefix = FilePrefix(filename_or_file_prefix)
>   File "/cvmfs/main.galaxyproject.org/galaxy/lib/galaxy/datatypes/sniff.py", line 541, in __init__
>     compressed_format, f = compression_utils.get_fileobj_raw(filename, "rb")
[...]

PR #8362, opened 2019-07-20
Fix autodetection of zip file format (including xlsx)

wm75:
addresses #8341 and adds integration and unit tests


lib/galaxy/util/compression_utils.py
```diff
@@ -47,8 +47,10 @@ def get_fileobj_raw(filename, mode="r", compressed_formats=None):
         compressed_format = 'bz2'
     elif 'zip' in compressed_formats and zipfile.is_zipfile(filename):
         # Return fileobj for the first file in a zip file.
-        with zipfile.ZipFile(filename, mode) as zh:
-            fh = zh.open(zh.namelist()[0], mode)
+        # 'b' is not allowed in the ZipFile mode argument
```
mvdbeek, line 50:
```suggestion
        # 'b' is not allowed in the ZipFile mode argument
        # since files will always be opened in binary mode.
        # If text mode is required fh will be wrapped in
        # TextIOWrapper.
```

test/unit/test_compression_util.py
```diff
@@ -2,20 +2,33 @@
 import tempfile
 import unittest
 
-from galaxy.util.compression_utils import CompressedFile
+from galaxy.util.compression_utils import CompressedFile, get_fileobj_raw
```
mvdbeek:
```suggestion
import six

from galaxy.util.compression_utils import (
    CompressedFile, 
    get_fileobj_raw,
)
```

This is just our import convention, and six is needed to assert we get the correct type from  `get_fileobj_raw `

test/unit/test_compression_util.py
```diff
@@ -2,20 +2,33 @@
...
 
 
 class CompressionUtilTestCase(unittest.TestCase):
 
     def test_compression_safety(self):
-        self.assert_safety("test-data/unsafe.tar", False)
```
mvdbeek: Why did you remove this test ?

test/unit/test_compression_util.py
```diff
@@ -26,3 +39,13 @@ def assert_safety(self, path, expected_to_be_safe):
                     CompressedFile(path).extract(temp_dir)
         finally:
             shutil.rmtree(temp_dir, ignore_errors=True)
+
+    def assert_format_detected(self, path, expected_fmt, allowed_fmts=None):
+        fmt1 = get_fileobj_raw(path, "r", allowed_fmts)[0]
```
mvdbeek: I'd do this in a for loop and also verify that we get the correct type:
```
for mode, expected_type in [('r', six.text_type), ('rb', six.binary_type), ('rt', six.text_type), ('U', six.text_type)]:
    fmt, fh = get_fileobj_raw(path, mode, allowed_fmts)
    assert fmt == expected_fmt
    assert isinstance(fh.read(1), expected_type)
```
I also like to avoid unittest's assertEqual (or unittest functionality in general), pytest has nice reporting for assertion errors.

mvdbeek (commented):
Great, that looks correct to me, I'd just make the testing a bit more explicit.
Did you drop that one assert_safety test on purpose ?

test/unit/test_compression_util.py
wm75: Oops, removed accidentally. Thanks for catching!

test/unit/test_compression_util.py
wm75: Thanks for the nice idea of using the six types!
The only complication I had to deal with on top is that when the file format does not get autodetected, then, in text mode, the stream might not be decodable using utf-8. I decided to just skip the type check in this situation.

wm75: Ha, found a more elegant approach. It is possible to read 0 bytes/characters from the stream to learn whether it's binary or text. This avoids the encoding issue.

mvdbeek (approved):

lib/galaxy/util/compression_utils.py
```diff
@@ -47,8 +47,13 @@ def get_fileobj_raw(filename, mode="r", compressed_formats=None):
...
+        # since it always opens files in binary mode.
+        # For emulating text mode, we will be returning the binary fh in a
+        # TextIOWrapper.
+        zf_mode = mode.replace('b', '')
+        with zipfile.ZipFile(filename, zf_mode) as zh:
+            fh = zh.open(zh.namelist()[0], zf_mode)
```
nsoranzo, line 56: @wm75 Will this work with a zipped binary file? I'd have thought to use `mode` instead of `zf_mode`.

mvdbeek: Yes, the mode is always binary, no matter what you select here. The unit tests assert this.
(indirectly though, by checking that that reading from the file handle returns `six.binary_type`)

wm75: @nsoranzo surprised me too, but the `zh.open` does not accept 'b' in mode either and always returns a binary stream.

wm75: @mvdbeek glad you suggested checking the stream type in the tests!

martenson:
Thanks @wm75, this helps a lot!

PR #8367, opened 2019-07-22
Pin python version for fiona package

mvdbeek:
This currently causes our framework tests to fail.
This happened at the same time as the switch to
python 3.7, so I'm hoping this will fix the tests.


mvdbeek:
Will be fixed upstream: https://github.com/conda-forge/fiona-feedstock/pull/132#issuecomment-513773754

PR #8368, opened 2019-07-22
Porting Tools Panel to Vue.js (resolves #8262)

### LINKED ISSUE #8262: Tools Panel - Vue Component

inkuzmin:
This is a work in progress on porting the tool panel to Vue.js we started during the CoFest at GCC2019 in Freiburg. Please note, that this PR is not ready to be merged yet.

In this PR the tools' list is reimplemented in Vue.js. The implementation was focused on feature parity. The mark-up was preserved.

Next steps:
- [x] Display workflows
- [x] Add the search bar
- [x] Make the search bar work with the tools' list
- [x] Remove original tools panel
- [x] Search bar should be fixed at top
- [x] Add SidePanel Vue component
- [x] Decide how it should be integrated
- [x] Add (un)collapse/resize functionality to the SidePanel
- [x] Add buttons to the ToolBox
- [x] Move SidePanel instantiation to `tool-panel.js`
- [x] Reuse `mountVueComponent.js`
- [x] Check SidePanel to be able to handle the right panel

Future work and ideas:
- Reuse the new tools panel for the Workflows Editor
- Improve search as proposed in #8262
- Make it possible to mark tool as favourite from the tools panel itself


bgruening:
Thanks @inkuzmin - this looks awesome already!

inkuzmin:
@bgruening thank you! One more session and it will be ready for review 🙂

inkuzmin:
The integration of the component is something we need to discuss actually.

If I understand correctly, presently, a side panel abstracted out in the threefold way: a header with buttons (`unified-panel-header`), a controls bar (`unified-panel-controls`) and a main view (`unified-panel-body`). During the initialisation of the specific panel the data for these layers are supplied separately: 
* Buttons: https://github.com/galaxyproject/galaxy/blob/fcaeb857b8bbf41d11c40fb7c27e58827d7038c8/client/galaxy/scripts/layout/panel.js#L31-L33

* The main view (for example, the list of tools):
https://github.com/galaxyproject/galaxy/blob/fcaeb857b8bbf41d11c40fb7c27e58827d7038c8/client/galaxy/scripts/layout/panel.js#L36

* Controls (the search bar):
https://github.com/galaxyproject/galaxy/blob/fcaeb857b8bbf41d11c40fb7c27e58827d7038c8/client/galaxy/scripts/mvc/tool/tools.js#L703-L706

I see the logic behind this generalisation pattern and why it was useful, but it doesn't feel very natural in case of the Vue components usage. Although in principle it probably can be accomplished with Vue slots, it comes with its costs (as any generalisation would) and it's not obvious that in the case of Vue implementation the benefits outweigh them. To name some, it hinders the communication between components and implies certain restrictions on the implementation (which is not necessarily an undesirable thing, but Vue implementation implies some restrictions on its own and it is not very clear that they would play together smoothly).

A SidePanel component which is just a column with collapse/resize buttons at the bottom that can be initialised as "left" or "right", and would be agnostic to what would happen to be placed inside is the solution I would opt for, but I don't know if it fits the current architecture well. Do you know what can be rendered at the left side panel apart from the tools panel?

What do you think about it and what is your vision of the integration?

inkuzmin:
This commit shows the architectural solution for the integration I described above.

guerler:
@inkuzmin thank you for working on this. I think having an agnostic sidepanel component (only handling the collapsing) is a great idea. To simplify we could consider making "getVueComponent" return the SidePanel component directly instead of just the Toolbox. In another step we might be able to skip calling panel.js and instead directly load the vue component for the toolpanel in page.js. Overall I think you are definitely on the right track. I also tested this PR and it already works well. We should merge this after some cleanup. Thanks again for your help on this project.

Nerdinacan:
I agree about the SidePanel issue @inkuzmin. 

In the history rebuild I had to write a somewhat brutal adapter to move some of that restrictive structure out of the way. I understand what was trying to be accomplished, but you're absolutely right that the "grouping" is all wrong for the solution.

I ended up resolving it with a temporary backbone adapter that essentially removed all the overly restrictive structure of the panel and mounted a Vue component in the resulting hole. This is obviously a temporary measure and we should definitely come up with some kind of "resizable panel" wrapper component to put our stuff inside.

Here's my temporary adapter in case it's of any help. I would love to get to a point where this kind of nonsense is not needed.
https://github.com/Nerdinacan/galaxy/blob/history_model_4/client/galaxy/scripts/components/History/HistoryPanelProxy.js

(P.S. Sorry for the wait! I'll pull this down and do a full review tomorrow, very close to getting a big chunk off my plate.)

guerler:
@Nerdinacan `We should definitely come up with some kind of "resizable panel" wrapper component to put our stuff inside.` That is what @inkuzmin started to build with `SidePanel.vue`. I think we should use it instead of the backbone wrapper.

Nerdinacan:
@guerler Yeah I'll check that component out.

My adapter's main purpose was not preserving the functionality of the side panel but finding an appropriate place to chop out old backbone structure and mount the new component without disrupting the old parts of the app. But I'm happy to implement a common SidePanel wrapper that handles resizing, etc.

inkuzmin:
@guerler 
@Nerdinacan 
Thank you for sharing your thoughts, it will definitely help!

@Nerdinacan, very neat implementation of the Vue components, a lot of interesting solutions to learn from!

Regarding the SidePanel, I'll check it to be ready to handle the right panel, may be it could be reused somehow indeed.

inkuzmin:
We discussed the possibility of including this work in the upcoming Galaxy release with Björn. I am up to it. I will be happy to fix the code according to the reviews and comments next week.

jmchilton:
This conflicts with https://github.com/galaxyproject/galaxy/pull/8331/files I think - any chance you can merge this and update the PR?

inkuzmin:
@jmchilton sure! I synchronised it with the `upstream/dev` and added missing functionality to the new implementation of the favorites button.

jmchilton:
@inkuzmin Thanks so much, awesome. Can you also fix the ES lint issues caught by the CircleCI tests?

https://circleci.com/gh/galaxyproject/galaxy/33621?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

```
$ eslint -c .eslintrc.js galaxy/scripts --ext .js,.vue

/home/circleci/repo/client/galaxy/scripts/components/Tool.vue
  12:21  error  Elements in iteration expect to have 'v-bind:key' directives  vue/require-v-for-key

/home/circleci/repo/client/galaxy/scripts/components/ToolBox.vue
   18:21  error  Custom elements in iteration require 'v-bind:key' directives  vue/valid-v-for
   33:21  error  Elements in iteration expect to have 'v-bind:key' directives  vue/require-v-for-key
   94:24  error  '_' is not defined                                            no-undef
   95:21  error  '_' is not defined                                            no-undef
   99:40  error  '_' is not defined                                            no-undef
  107:33  error  '_' is not defined                                            no-undef

/home/circleci/repo/client/galaxy/scripts/components/ToolSection.vue
  14:25  error  Custom elements in iteration require 'v-bind:key' directives  vue/valid-v-for
  15:25  error  Elements in iteration expect to have 'v-bind:key' directives  vue/require-v-for-key

✖ 9 problems (9 errors, 0 warnings)
```

inkuzmin:
@jmchilton absolutely! I added unique keys for list items and explicitly imported libraries that are used.

jmchilton:
Re-running the Selenium tests to see if the failures are related.

inkuzmin:
> to see if the failures are related

I think they are. I tried to keep the mark-up as close to the original one as possible as long as it doesn't interfere with the potential benefits of porting. In this case the problem is that I didn't preserve `id` attributes. The reason is that using `id`'s reduces component's reusability. For example, rendering two components with the same `id` at the same HTML document would make its validation fail.

There are two ways to solve it: to loosen restrictions and add `id`'s back or to fix the tests to use different selectors instead. Which one would you recommend to proceed with?

martenson:
+1 to change test to use better selector

jmchilton:
I was confused why there wasn't more red in this PR and so I found some unused code after these changes and pushed a commit with that code being deleted.

jmchilton:
I'll take a look at the test while I'm in here... I think this is actually coming from a tour not a test but should be an easy fix either way.

jmchilton:
Thanks a bunch, this is awesome stuff!

martenson:
Good job @inkuzmin 🥇

inkuzmin:
Yay!
@jmchilton huge thanks for wrapping this up!
@martenson, @guerler, @Nerdinacan thank you for your insightful comments!

PR #8369, opened 2019-07-22
Make filter work with file containing header

ccedmendoza:
This filter doesn't work when the GFT file contains a '#' header

### DIFF
--- a/tools/filters/gff/gtf_filter_by_attribute_values_list.py
+++ b/tools/filters/gff/gtf_filter_by_attribute_values_list.py
@@ -7,60 +7,29 @@
 from __future__ import print_function
 
 import sys
-
-
-def parse_gff_attributes(attr_str):
-    """
-    Parses a GFF/GTF attribute string and returns a dictionary of name-value
-    pairs. The general format for a GFF3 attributes string is
-        name1=value1;name2=value2
-    The general format for a GTF attribute string is
-        name1 "value1" ; name2 "value2"
-    The general format for a GFF attribute string is a single string that
-    denotes the interval's group; in this case, method returns a dictionary
-    with a single key-value pair, and key name is 'group'
-    """
-    attributes_list = attr_str.split(";")
-    attributes = {}
-    for name_value_pair in attributes_list:
-        # Try splitting by space and, if necessary, by '=' sign.
-        pair = name_value_pair.strip().split(" ")
-        if len(pair) == 1:
-            pair = name_value_pair.strip().split("=")
-        if len(pair) == 1:
-            # Could not split for some reason -- raise exception?
-            continue
-        if pair == '':
-            continue
-        name = pair[0].strip()
-        if name == '':
-            continue
-        # Need to strip double quote from values
-        value = pair[1].strip(" \"")
-        attributes[name] = value
-
-    if len(attributes) == 0:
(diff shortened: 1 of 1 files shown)


bernt-matthias:
seems a good point. the other tools in this directory use a similar construct, but also check for empty lines which should also be done here. 

https://github.com/galaxyproject/galaxy/blob/f7d60083c8919cbc09691d0db85f661aa16b6a16/tools/filters/gff/gff_filter_by_attribute.py#L265

I would also suggest to insert a comment line and an empty line in the example gtf files in order to show that it works and (more importantly) to make sure that it does not break again.

ccedmendoza:
Hi @bernt-matthias ,
I've been looking for the test files of this tool, to add the comment and empty lines, but I can't find them.
Thank you for your time.

bernt-matthias:
Seems that they are in a separate repo: https://github.com/galaxyproject/galaxy-test-data

ccedmendoza:
@bernt-matthias ,
Thank you for your help. This should resolve the empty lines and commented lines handling.
I also made a PR to change the test data.

tools/filters/gff/gtf_filter_by_attribute_values_list.py
```diff
@@ -6,60 +6,31 @@
...
-    output.close()
+    with open(output_file, 'w') as output:
+        with open(gff_file) as gff:
+            for line in gff:
+                # Keep commented lines
+                if line.startswith('#'):
+                    output.write(line)
+                else:
+                    # Create pattern using attribute to filter
+                    prog = re.compile(r".*" + re.escape(attribute_name) + r" \"(.+?)\"\;")
```
bernt-matthias, line 29: Not sure if the re works correctly. I guess the last part should be more like this `r" \"[^\"]+\";".  Otherwise greedy regular expressions (I remember that this is the default) will match multiple attributes at once. Also not sure if the `?` is needed (I think not). 

I would also apply the regular expression only on the attributes column of the gtf.

tools/filters/gff/gtf_filter_by_attribute_values_list.py
```diff
@@ -6,60 +6,31 @@
...
 
     # Filter GFF file using ids.
-    output = open(output_file, 'w')
-    for line in open(gff_file):
-        fields = line.split('\t')
-        attributes = parse_gff_attributes(fields[8])
-        if (attribute_name in attributes) and (attributes[attribute_name] in ids_dict):
-            output.write(line)
-    output.close()
+    with open(output_file, 'w') as output:
```
bernt-matthias, line 21: You can put multiple with statements on one line `with A() as a, B() as b:` .. up to you. I prefer less nesting.

mvdbeek:
I've bumped the milestone to 20.05, since I think we may need to restructure the test-data repository to follow the galaxy releases. Independently of this the tool version should be bumped, a python requirement should be added to the tool itself, and the regex.compile step should be moved out of the for loop. Finally I'm not sure a smaller change that just takes care of empty lines and comments wouldn't be more appropriate for tools shipped with Galaxy, given that these really are maintenance only.

bernt-matthias:
superseded by https://github.com/galaxyproject/galaxy/pull/11365

PR #8371, opened 2019-07-22
Use bootstrap classes to fix form css issue

### LINKED ISSUE #8288: tool form data selector renders off pane
this is on dev with the genome exporter tool

Galaxy

WilliamHolden:
fixes #8288

### DIFF
--- a/client/galaxy/scripts/mvc/ui/ui-select-content.js
+++ b/client/galaxy/scripts/mvc/ui/ui-select-content.js
@@ -335,7 +335,7 @@ const View = Backbone.View.extend({
         this.button_type = new Ui.RadioButton.View({
             value: this.model.get("current"),
             data: this.button_data,
-            cls: "mr-2 mt-2",
+            cls: "pr-lg-2",
             onchange: function(value) {
                 self.model.set("current", value);
                 self.trigger("change");
@@ -346,7 +346,6 @@ const View = Backbone.View.extend({
         this.button_dialog = new Ui.Button({
             icon: "fa-folder-open-o",
             tooltip: "Browse Datasets",
-            cls: "mt-2",
             onclick: () => {
                 const current = this.model.get("current");
                 const cnf = this.config[current];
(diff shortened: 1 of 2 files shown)


guerler:
@WilliamHolden thanks for working on this. The flow works well but in some cases it seems to fold too early and the spacing inbetween the buttons and the select field goes missing:

Screen Shot 2019-07-23 at 1 13 29 AM

WilliamHolden:
Thanks for reviewing @guerler. 

As for the two issues you mentioned:

1. The reason the form field wraps early sometimes is because of large file names. It prefers to wrap the select field and keep the text rather than not wrapping and cutting off the text like it does currently. 

2. The vertical spacing between the buttons goes missing because there is no vertical padding on the elements. I can add the padding but I would probably have to use this solution which I'm not a huge fan of: https://stackoverflow.com/questions/30887071/margin-top-only-when-the-flex-item-is-wrapped.

If I can think of a better solution, I will try to implement. I think both are fixable and number 2 should definitely be fixed but I can't decide if the first issue is a feature or a bug. Any other input is always appreciated before I start implementing a fix.


Here is how it looks side by side.

**Current on dev:**
old


**With updates from current PR**
new

WilliamHolden:
Updated the pull request. 


newestgif


> The flow works well but in some cases it seems to fold too early
> 

I'm still unsure what's the best way to handle this.

WilliamHolden:
The last thing I need to add for this PR is how to handle large file names in the select field. Do we want to wrap the field when the filename is large or cut off the text like we do currently. Anyone have opinions on this? @guerler @dannon @martenson .

This is the current functionality on dev:

currentdevwrap


This is the current functionality with the PR:

currentprwrap

guerler:
@WilliamHolden I think the first variation is better. In my opinion the second option creates to much motion.

guerler:
Maybe we can fix the cut off. In the first example it cuts the string off with just displaying the history id "29:" although more space is left.

WilliamHolden:
@guerler I like your suggestions. I have updated the PR to only wrap vertically on medium size screens(small laptops, tablets etc...) and smaller. On larger screens, the three elements will always be horizontal. I also updated the PR so that the string inside the select field makes use of all the available whitespace. Let me know what you think.

latestpr

guerler:
This looks great to me and is elegant. Thank you for fixing this.

martenson:
I really appreciate the thought you put into this, thanks @WilliamHolden.

PR #8373, opened 2019-07-23
Hide references card if array is empty

guerler:


mvdbeek (approved):
Nice improvement, thanks @guerler!

PR #8376, opened 2019-07-24
Identify visualization static path from absolute path

guerler:
Follow-up for #8346. Adjusting static path for visualizations.

### DIFF
--- a/lib/galaxy/visualization/plugins/plugin.py
+++ b/lib/galaxy/visualization/plugins/plugin.py
@@ -5,6 +5,7 @@ from a query string and render a webpage based on those data.
 import copy
 import logging
 import os
+import re
 
 import mako.lookup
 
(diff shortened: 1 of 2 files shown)


lib/galaxy/visualization/plugins/plugin.py
```diff
@@ -127,6 +128,12 @@ def _get_url(self):
             return url_for(controller='visualization', action=self.name)
         return url_for('visualization_plugin', visualization_name=self.name)
 
+    def _get_static_path(self, path):
+        match = re.match(r'.*/config/(.*)', path)
```
mvdbeek: How about
```
if '/config/' in path:
    return path.rsplit('/config/')[0]
```
That'd avoid the regex and should give the same result.
I'm just wondering what happens if `None` is returned,
will this raise a proper exception somewhere or will it
leave admins/viz developers wondering why visualizations don't work ?
Should we maybe also log something at https://github.com/galaxyproject/galaxy/blob/9ec952c0a7a6db510f01ea3b3e9459b2d2de11be/lib/galaxy/visualization/plugins/registry.py#L189 if there's no `config` ?

jmchilton (approved):

PR #3399, opened 2017-01-10
Markdown datatype support.

jmchilton:
With test tools demonstrating HTML and Markdown outputs from tools.

Implements #3396 as part of the Qiime hackathon.

PR #8388, opened 2019-07-30
Improve color relationships between text and background

guerler:
This PR follows-up on our style consolidation efforts. It consolidates the control on foreground colors (for regular-text, muted-text and borders) by deriving them consistently from `$brand-dark` and then contrasting them with background colors derived from `$brand-light`. The default color theme remains unchanged, except minor contrast improvements for `$text-muted` elements and the substitution of custom gray tones. 

Screen Shot 2019-07-30 at 3 56 30 PM

**For Example:** 
Previously background and text colors were mostly derived from one channel i.e. by darkening brand-light. This required brand-light to be an actual light color. @jxtx thanks for pointing this out. With this PR we separate text and background colors into two contrasting channels. Swapping brand-light and brand-dark, and replacing primary/secondary brand colors with orange now results in a legible theme (Example 1, below). A remaining challenge is that primary/secondary brand colors have to contrast brand-light and brand-dark. This might be something to be addressed in another follow-up if necessary.

1. Setting `$brand-dark=#f8f9fa`, `$brand-light=$2c3143` and `$brand-primary/secondary=orange` results in:
image

2. Setting `$brand-primary: red; $brand-secondary: green; $brand-light: black; $brand-dark: orange;` results in:

Screen Shot 2019-07-31 at 1 22 40 AM

3. Setting `$brand-dark:#003399` results in:

Screen Shot 2019-07-31 at 12 13 32 AM

4. Setting `$brand-dark:#003399` and increasing the darkness of `$text-color` by 20% results in:

Screen Shot 2019-07-31 at 12 23 24 AM

### DIFF
--- a/client/galaxy/scripts/components/PluginList.vue
+++ b/client/galaxy/scripts/components/PluginList.vue
@@ -20,15 +20,15 @@
                             <div v-else class="ui-thumbnails-icon fa fa-eye" />
                         </td>
                         <td>
-                            <div class="ui-thumbnails-title font-weight-bold text-dark">{{ plugin.html }}</div>
-                            <div class="ui-thumbnails-text text-dark">{{ plugin.description }}</div>
+                            <div class="ui-thumbnails-title font-weight-bold">{{ plugin.html }}</div>
+                            <div class="ui-thumbnails-text">{{ plugin.description }}</div>
                         </td>
                     </tr>
                     <tr v-if="!fixed">
                         <td />
                         <td v-if="plugin.name == name">
                             <div v-if="hdas && hdas.length > 0">
-                                <div class="font-weight-bold text-dark">Select a dataset to visualize:</div>
+                                <div class="font-weight-bold">Select a dataset to visualize:</div>
                                 <div class="ui-select">
                                     <select class="select" v-model="selected">
                                         <option v-for="file in hdas" :key="file.id" :value="file.id">{{
--- a/client/galaxy/scripts/layout/page.js
+++ b/client/galaxy/scripts/layout/page.js
@@ -172,8 +172,8 @@ const View = Backbone.View.extend({
             `<div id="everything">
                 <div id="background"/>
                 <div id="masthead"/>
-                <div id="messagebox"/>
-                <div id="inactivebox" class="panel-warning-message" />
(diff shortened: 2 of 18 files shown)


guerler:
@galaxybot test this

hexylena:
love this! It makes it so much easier. A few things for discussion maybe:

1. Should the select2 be styled as well?
   
    image

2. and the buttons here, I guess I should set brand secondary here, it can't easily be a derived colour or?

   image

3. I could see an argument for making the background white here? Since that will skew people's colour perception

   image

4. With `$brand-light: white; $brand-dark: red; $brand-secondary: green;`, the borders seem to disappear? looks like the following could be overridden

   ```
   .ui-select .select2-container .select2-choice { border: white } 
   ```

   image

5. Some job states are a bit unreadable, but this feels like a really hard problem to solve generically? If I set `-light` to really a light colour it looks OK, if it's a dark and `-dark` is a light colour, then it's an issue. Is there any chance to ensure a contrasting colour for the text, white if it is dark / black if it is light?

   image

martenson (approved):

guerler:
@erasche thank you so much for the quick review and helpful comments.

**1.** I added the styling to select2 popups.

**4.** This can be fixed by changing the factor of the border-color brightening. In e236758f50617534d2c6acc6251537727030253c I moved this option directly below the brand colors in the theme file and named it `$border-color` for easier access.

**3.** If you are ok with it I would rather leave the background of the color picker as is for now. I want to avoid placing fixed colors into elements as much as possible.

**2. & 5.** are related and probably the main limitation of the theming right now. It requires that primary/secondary brand colors contrast brand light and brand dark. Resolving this limitation would make it much easier to pick viable themes. Both should be fixable in a follow-up. They require revisiting how bootstrap theme colors are derived from brand colors to ensure that colors with proper contrast are picked. The bootstrap theme colors are defined at the bottom of our theme file. I would rather leave this enhancement to a follow-up PR.

Thanks again.

hexylena (approved):
Sounds great!

guerler:
Thank you for the review!

hexylena:
Thanks for making it so much easier to theme things!

PR #8393, opened 2019-07-31
patch gid handling on the file system

aroudgar:
Galaxy tries to change the group name of files using chown command to be identical to the user name. In our hpc environment this creates lots of problem for users.

### DIFF
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1416,7 +1416,6 @@ class JobWrapper(HasResourceParameters):
                 try:
                     # Attempt to short circuit NFS attribute caching
                     os.stat(dataset.dataset.file_name)
-                    os.chown(dataset.dataset.file_name, os.getuid(), -1)
                     trynum = self.app.config.retry_job_output_collection
                 except (OSError, ObjectNotFound) as e:
                     trynum += 1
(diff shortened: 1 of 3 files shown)


bernt-matthias:
Could you describe the problems a bit more?

aroudgar:
In our hpc environment (computecanada.ca) the GID of the user is not necessarily the same as the user's UID.  Every user has disk usage quota based on his UID and GID and if the user try to change GID they will get out of quota. Every user has its his own galaxy instance and they can run it on their own account. They are responsible to run and maintain their own galaxy. Admin will give them a web interface for every and each galaxy server. The problem start when users try to install toolsheds.  During the installation the GID of new files that are installed changed and become the same as UID. So, users receive an error message of disk quota error during the installation.

lib/galaxy/jobs/__init__.py
```diff
@@ -1416,7 +1416,6 @@ def _finish_dataset(self, output_name, dataset, job, context, final_job_state, r
                 try:
                     # Attempt to short circuit NFS attribute caching
                     os.stat(dataset.dataset.file_name)
-                    os.chown(dataset.dataset.file_name, os.getuid(), -1)
```
bernt-matthias, line 1419: The -1 should force chown not to touch the gid. So this should not be the source of your problem. Note that also the owner is not changed (other wise root would be needed). The intention of this code block is to deal with problems regarding NFS file systems as indicated by the comment above.

aroudgar: yes, you are right. I double checked and indeed I can drop the last patch. Only the first two patches are in the place.

bernt-matthias:
Maybe you could also consider https://github.com/galaxyproject/galaxy/blob/736c1e9b6d10ca379b3cdbe3115d1e86be5606f5/lib/galaxy/config/sample/galaxy.yml.sample#L1654

PR #8394, opened 2019-08-01
make cleanup script use logging instead of print

bernt-matthias:


martenson:
thanks @bernt-matthias !

Issue #8401, opened 2019-08-01
Unable to view bigwig and bigbed files in IGB uploaded through galaxy instance.


nzahara1:
I have hosted galaxy on Amazon EC2 instance.
Trying to view bigwig and bigbed files in IGB uploaded via galaxy and getting the below error message in IGB : Error reading R+ tree header: bad magic = -2003829722 while reading the file.

Galaxy console error : 
"GET /display_application/df7a1f0c02a5b08e/igb_bigwig/View/383651628846c1ff/data/BigWig_HomoSapien.bigWig HTTP/1.1" 200 8841495 "-" "IGB/9.1.0; Mac OS X/10.14.5 (x86_64); en_US; Java/1.8.0_221"
[uwsgi-http key: 54.208.249.92:8080 client_addr: 209.239.128.14 client_port: 38882] hr_write(): Broken pipe [plugins/http/http.c line 565]

martenson:
some other info from chat:

> it is using ngnix; It is just happening with bigbed and bigwig files; I am able to visualize the data for other file formats

I was able to reproduce this locally with a test bigwig file

nzahara1:
The issue is happening for bed files as well... 

Error message : http://54.208.249.92:8080/display_application/1e8ab44153008be8/igb_bed/View/4dabc5eb4dd2db33/data/Bed_HomoSapien.bed is not reachable, and is not cached!

Galaxy console message :

 GET /display_application/1e8ab44153008be8/igb_bed/View/4dabc5eb4dd2db33/data/Bed_HomoSapien.bed => generated 7405568 bytes in 590 msecs (HTTP/1.1 200) 3 headers in 100 bytes (116 switches on core 3)
[uwsgi-http key: 54.208.249.92:8080 client_addr: 209.239.128.14 client_port: 1534] hr_write(): Broken pipe [plugins/http/http.c line 565]

mvdbeek:
You need a proxy that supports byte-range requests for data types that support this, which includes bigwig and bam files. You mention that you're using nginx, but `[uwsgi-http key: 54.208.249.92:8080 client_addr: 209.239.128.14 client_port: 1534] hr_write(): Broken pipe [plugins/http/http.c line 565]` indicates that uwsgi is responding directly.

Make sure to use nginx's uwsgi protocol, not http. An extensive documentation is at https://docs.galaxyproject.org/en/release_19.05/admin/nginx.html.

To verify byte-range requests work you can use curl:

[...]

PR #8404, opened 2019-08-02
mention IGB in the docs

martenson:
xref https://github.com/galaxyproject/galaxy/issues/8401


mvdbeek (approved):
In principle uwsgi should also be able to do that, but it appears it needs to be started with `--honor-range`
https://uwsgi-docs.readthedocs.io/en/latest/Changelog-1.9.6.html#preliminary-support-for-the-http-range-header
haven't tried that.

jmchilton (approved):

PR #8409, opened 2019-08-03
argument tag: automatically replace dashes by underscores

bernt-matthias:
Long options often contain dashes, e.g. `--long-option`. Since dashes are not allowed in variable names tool developers always need to specify `<param name="long_option" argument="--long-option">`. Now its only `<param argument="--long-option">`. 

Possible extension: 
- test somewhere here or in planemo if the name in a valid variable name
- simply autogenerate a valid variable name by replacing everything but letters, numbers and underscores by underscores

I guess the former option would be preferable, because its more transparent and I have never seen such variable names.


lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -1788,7 +1788,7 @@ statement. A good example tool that demonstrates many conditional parameters is
     </param>
     <when value="tsv">
         <param name="input_table" type="data" format="tabular" label="Tabular File" argument="--input-fp"/>
-        <param name="process_obs_metadata" type="select" label="Process metadata associated with observations when converting" argument="--process-obs-metadata">
+        <param type="select" label="Process metadata associated with observations when converting" argument="--process-obs-metadata">
```
```suggestion
        <param argument="--process-obs-metadata" type="select" label="Process metadata associated with observations when converting">
```
Since this replaces the `name` argument I think it should come first.
Also we should update https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#parameter-name-argument-and-help once merged.

mvdbeek:
@galaxybot test py3

mvdbeek (approved):
nice, I like this a lot!

lib/galaxy/tool_util/xsd/galaxy.xsd
bernt-matthias: update to help started: https://github.com/galaxy-iuc/standards/pull/55

bernt-matthias:
What do you think about linting variable names? Would this help? I guess that invalid names derived from argument occur really seldomly.

mvdbeek:
> What do you think about linting variable names? Would this help? I guess that invalid names derived from argument occur really seldomly.

So that planemo lint returns an error if you have an invalid param name, like `<param name="hi-there"/>` ? I think that'd be useful, certainly if you configure your editor to run `planemo lint` on Galaxy tools.

bernt-matthias:
OK. Would I need to do this here: lib/galaxy/tool_util/linters/inputs.py and lib/galaxy/tool_util/linters/outputs.py?

bernt-matthias:
I guess we should add a unit test like you did here: https://github.com/galaxyproject/galaxy/pull/8409?

jmchilton (approved):

lib/galaxy/tool_util/linters/outputs.py
```diff
@@ -22,6 +23,9 @@ def lint_output(tool_xml, lint_ctx):
         output_attrib = output.attrib
         if "name" not in output_attrib:
             lint_ctx.warn("Tool output doesn't define a name - this is likely a problem.")
+        else:
+            if not is_valid_cheetah_placeholder(output_attrib["name"]):
+                lint_ctx.warn("Tool output name %s is not a valid cheetah placeholder.", output_attrib["name"])
```
nsoranzo:
```suggestion
                lint_ctx.warn("Tool output name [%s] is not a valid Cheetah placeholder.", output_attrib["name"])
```

lib/galaxy/tool_util/linters/inputs.py
```diff
@@ -24,6 +24,10 @@ def lint_inputs(tool_xml, lint_ctx):
 
         param_type = param_attrib["type"]
         param_name = param_attrib.get("name", param_attrib.get("argument"))
+
+        if not is_valid_cheetah_placeholder(param_name):
+            lint_ctx.warn("Param input [%s] is not a valid cheetah placeholder.", param_name)
```
nsoranzo:
```suggestion
            lint_ctx.warn("Param input [%s] is not a valid Cheetah placeholder.", param_name)
```

lib/galaxy/tool_util/linters/_util.py
```diff
@@ -1,5 +1,11 @@
+import re
 
 
 def is_datasource(tool_xml):
     """Returns true if the tool is a datasource tool"""
     return tool_xml.getroot().attrib.get('tool_type', '') == 'data_source'
+
+
+def is_valid_cheetah_placeholder(name):
+    """Returns true if name is a valid cheetah placeholder"""
```
nsoranzo:
```suggestion
    """Returns true if name is a valid Cheetah placeholder"""
```

lib/galaxy/tools/parameters/basic.py
```diff
@@ -84,6 +84,13 @@ class ToolParameter(Dictifiable):
     Describes a parameter accepted by a tool. This is just a simple stub at the
     moment but in the future should encapsulate more complex parameters (lists
     of valid choices, validation logic, ...)
+
+    >>> from galaxy.util.bunch import Bunch
+    >>> trans = Bunch(app=None)
+    >>> p = TextToolParameter(None, XML('<param argument="--parameter-name" type="text" value="default" />'))
+    >>> print(p.name)
+    parameter_name
```
nsoranzo:
```python
    >>> assert p.name == 'parameter_name'
```

lib/galaxy/tools/parameters/basic.py
```diff
@@ -84,6 +84,13 @@ class ToolParameter(Dictifiable):
     Describes a parameter accepted by a tool. This is just a simple stub at the
     moment but in the future should encapsulate more complex parameters (lists
     of valid choices, validation logic, ...)
+
+    >>> from galaxy.util.bunch import Bunch
+    >>> trans = Bunch(app=None)
+    >>> p = TextToolParameter(None, XML('<param argument="--parameter-name" type="text" value="default" />'))
```
nsoranzo: Given this is a doctest for `ToolParameter`, I'd use:
```suggestion
    >>> p = ToolParameter(None, XML('<param argument="--parameter-name" type="text" value="default" />'))
```
and update the assert below as needed.

lib/galaxy/tool_util/linters/inputs.py
```diff
@@ -24,6 +24,10 @@ def lint_inputs(tool_xml, lint_ctx):
 
         param_type = param_attrib["type"]
         param_name = param_attrib.get("name", param_attrib.get("argument"))
```
nsoranzo: You need implement the same as in https://github.com/galaxyproject/galaxy/pull/8409/files#diff-9baf995401cfeb779edf8731ebaf0d2dR256-R260 , otherwise validating a param with no `name` and `argument` that contains `-` will fail.

bernt-matthias: True. I guess the best would be to use the actual function https://github.com/galaxyproject/galaxy/blob/370776bb60a929389667984436698e8a8e8c2ec3/lib/galaxy/tools/parameters/basic.py#L255 instead of reimplementing it here.

nsoranzo: Yep, maybe extracted from the class.

lib/galaxy/tools/parameters/basic.py
bernt-matthias: done

lib/galaxy/tools/parameters/basic.py
```diff
@@ -84,6 +84,12 @@ class ToolParameter(Dictifiable):
     Describes a parameter accepted by a tool. This is just a simple stub at the
     moment but in the future should encapsulate more complex parameters (lists
     of valid choices, validation logic, ...)
+
+    >>> from galaxy.util.bunch import Bunch
+    >>> trans = Bunch(app=None)
+    >>> p = ToolParameter(None, XML('<param argument="--parameter-name" type="text" value="default" />'))
+    >>> assert p.name == 'parameter_name'
+    >>> assert sorted(p.to_dict(trans).items()) == [('area', False), ('argument', '--parameter-name'), ('datalist', []), ('help', ''), ('hidden', False), ('is_dynamic', False), ('label', ''), ('model_class', 'ToolParameter'), ('name', 'parameter_name'), ('optional', False), ('refresh_on_change', False), ('type', 'text'), ('value', 'default')]
```
nsoranzo:
```suggestion
    >>> assert sorted(p.to_dict(trans).items()) == [('argument', '--parameter-name'), ('help', ''), ('hidden', False), ('is_dynamic', False), ('label', ''), ('model_class', 'ToolParameter'), ('name', 'parameter_name'), ('optional', False), ('refresh_on_change', False), ('type', 'text'), ('value', None)]
```

bernt-matthias:
any idea why `from galaxy.tools.parameters.basic import ToolParameter` does not work in the linter py file?

lib/galaxy/tool_util/linters/inputs.py
```diff
@@ -1,6 +1,7 @@
 """This module contains a linting functions for tool inputs."""
+from galaxy.tools.parameters.basic import ToolParameter
```
nsoranzo: Importing from `galaxy.tools` inside `galaxy.tool_util` would break the splitting in packages. I'd move the `parse_name()` method from `galaxy.tools.parameters.basic.ToolParameter` to `galaxy.tool_util.parser.xml.XmlInputSource` . Ping @jmchilton for double checking the idea.

lib/galaxy/tool_util/linters/inputs.py
```diff
@@ -23,7 +24,11 @@ def lint_inputs(tool_xml, lint_ctx):
             continue
 
         param_type = param_attrib["type"]
-        param_name = param_attrib.get("name", param_attrib.get("argument"))
+        param_name = ToolParameter.parse_name(param)
```
mvdbeek:
```suggestion
        param_name = param_attrib.get('name')
        if param_name is None:
            param_name = param_attrib['argument'].lstrip('-').replace("-", "_")
```
should work fine and requires less internals of Galaxy, which is a good thing for the linter.

nsoranzo: @mvdbeek `, None)` is not needed, it's the default for `dict.get()` .

Also your version is missing the `.lstrip('-')` .

I think that my suggestion above would still be a better solution (less code duplication), given both files are part of the `tool_util` package.

bernt-matthias: I would also agree more with @nsoranzo

mvdbeek: I don't think this belongs in `XmlInputSource`, this is tool parameter logic that is independent of the type of input source. At the minimum you'd also have to add it to `YamlInputSource` or `InputSource`.

bernt-matthias: OK. How about this: We move the actual functionality to galaxy/tool_util/parser/util.py adding a new function there, e.g. `argument_to_name`. 

This could then be used in 
* tool_util/linters/inputs.py 
* tools/parameters/basic.py

nsoranzo: I think moving it to `InputSource`, as suggested by Marius, is better.

bernt-matthias: Okidoki. Lets see if I got this right :)

lib/galaxy/tool_util/parser/interface.py
```diff
@@ -290,6 +290,25 @@ def get_bool(self, key, default):
...
+    def parse_name(self):
+        """Return name of an input source
+        returns the name or if absent the argument property
+        in the later case leading dashes are stripped and
+        internal dashes are replaced by underscore
+        """
+        return self._parse_name(self.get('name'), self.get('argument'))
+
+    @staticmethod
+    def _parse_name( name, argument ):
```
nsoranzo:
```suggestion
    def _parse_name(name, argument):
```

lib/galaxy/tool_util/parser/interface.py
```diff
@@ -290,6 +290,25 @@ def get_bool(self, key, default):
...
+        return self._parse_name(self.get('name'), self.get('argument'))
+
+    @staticmethod
+    def _parse_name(name, argument):
+        """Return name of an input source, see parse_name.
+        This is a helper allowing to use the functionality independent of
+        InputSource instances.
+        """
+        if name is None:
+            assert argument is not None
```
nsoranzo:
```suggestion
            if not argument:
                raise ValueError("parameter must specify a 'name' or 'argument'.")
```

lib/galaxy/tool_util/parser/interface.py
```diff
@@ -290,6 +290,25 @@ def get_bool(self, key, default):
...
+    def parse_name(self):
+        """Return name of an input source
+        returns the name or if absent the argument property
+        in the later case leading dashes are stripped and
+        internal dashes are replaced by underscore
+        """
+        return self._parse_name(self.get('name'), self.get('argument'))
+
+    @staticmethod
+    def _parse_name(name, argument):
```
nsoranzo: @bernt-matthias  Now that I see it implemented, I agree with you that a function in `lib/galaxy/tool_util/parser/util.py` would look better (when imported in the linter) than this static method, but I still like having the `parse_name()` method in this class. Would you agree to move this and keep the one above?

Sorry for going back and forth on this.

bernt-matthias: No worries. Coding is an iterative process :) 

Just trying to wrap my head around the failing tests ... I assumed that this `input_source` actually is an `InputSource`

https://github.com/galaxyproject/galaxy/blob/0f2ab9a08da99133541ea0b5f4b51467ef390bd9/lib/galaxy/tools/parameters/basic.py#L255

seems not to be the case.

bernt-matthias: Riddle after riddle. Any idea why these imports are not working?

lib/galaxy/tool_util/parser/util.py
```diff
@@ -69,3 +69,16 @@ def _set_oom(obj):
...
+    """Determine name of an input source from name and argument
+    returns the name or if absent the argument property
+    in the later case leading dashes are stripped and
+    internal dashes are replaced by underscore
+    """
+    if name is None:
+        if not argument:
+            raise ValueError("parameter must specify a 'name' or 'argument'.")
+        name = argument.lstrip('-').replace("-", "_")
+    return name
```
nsoranzo:
```suggestion
    return name
```

lib/galaxy/tool_util/parser/interface.py
nsoranzo: > Any idea why these imports are not working?

There are now circular imports between lib/galaxy/tool_util/parser/util.py and lib/galaxy/tool_util/parser/interface.py

I don't think `util` should be importing from `interface`, but not sure what's the best way to resolve this.

jmchilton: Good question, I'm not sure I care between the bad choices. Refactor the ToolStdio classes into util or revert to using a staticmethod without the import, or move the utility into the interface class. They are all more or less fine and not particularly great.

bernt-matthias:
@jmchilton: I hope that refactoring everything stdio related into a separate file is a badder (äh better) solution.

I cleaned the commit history and force pushed.

lib/galaxy/tool_util/parser/interface.py
nsoranzo: I think you forgot to remove the static method from here.

bernt-matthias: indeed. fixed

lib/galaxy/tool_util/linters/inputs.py
```diff
@@ -23,7 +24,11 @@ def lint_inputs(tool_xml, lint_ctx):
             continue
 
         param_type = param_attrib["type"]
-        param_name = param_attrib.get("name", param_attrib.get("argument"))
+        param_name = _parse_name(param_attrib.get("name"), param_attrib.get("name"))
```
nsoranzo:
```suggestion
        param_name = _parse_name(param_attrib.get("name"), param_attrib.get("argument"))
```

lib/galaxy/tool_util/parser/interface.py
```diff
@@ -290,6 +290,14 @@ def get_bool(self, key, default):
     def parse_label(self):
         return self.get("label")
 
+    def parse_name(self):
+        """Return name of an input source
+        returns the name or if absent the argument property
+        in the later case leading dashes are stripped and
```
nsoranzo:
```suggestion
        In the latter case, leading dashes are stripped and
```

lib/galaxy/tool_util/parser/interface.py
```diff
@@ -290,6 +290,14 @@ def get_bool(self, key, default):
...
+        internal dashes are replaced by underscore
```
nsoranzo:
```suggestion
        all remaining dashes are replaced by underscores.
```

lib/galaxy/tool_util/parser/util.py
```diff
@@ -1,71 +1,18 @@
...
-def _set_oom(obj):
-    obj.error_level = StdioErrorLevel.FATAL_OOM
-
-
-def _set_fatal(obj):
-    obj.error_level = StdioErrorLevel.FATAL
+def _parse_name(name, argument):
+    """Determine name of an input source from name and argument
+    returns the name or if absent the argument property
+    in the later case leading dashes are stripped and
```
nsoranzo:
```suggestion
    In the latter case, leading dashes are stripped and
```

lib/galaxy/tool_util/parser/util.py
```diff
@@ -1,71 +1,18 @@
...
+    internal dashes are replaced by underscore
```
nsoranzo:
```suggestion
    all remaining dashes are replaced by underscores.
```

nsoranzo (commented):
Nice refactoring!

nsoranzo (approved):

mvdbeek:
@bernt-matthias can you rebase the PR against current dev ?

bernt-matthias:
Rebased. Lets see if the tests still work :)

jmchilton:
Awesome - thanks so much!

PR #8414, opened 2019-08-05
Record Galaxy version job was created with.

jmchilton:
We have so much job code dependent on what version of Galaxy created a job. This should help keep track of that without having to hit the job directory to guess how Galaxy behaved at job creation. Also good to track in general for traceability/reproducibility and such.


mvdbeek (approved):
Cool, thanks @jmchilton!

PR #8418, opened 2019-08-06
[19.05] Try to fix too many workflow step output associations.

jmchilton:
Certain entries on main have hundreds of thousands of such records because they're being created at each iteration of workflow scheduling. This leads to serious memory and runtime issues over time.

Cleanup extra rows with:

```sql
DELETE FROM workflow_invocation_step_output_dataset_association
WHERE id NOT IN (

SELECT max(w.id) FROM
workflow_invocation_step_output_dataset_association as w GROUP BY workflow_invocation_step_id, dataset_id, output_name

)
```

### DIFF
--- a/lib/galaxy/workflow/modules.py
+++ b/lib/galaxy/workflow/modules.py
@@ -594,7 +594,7 @@ class InputModule(WorkflowModule):
         progress.set_outputs_for_input(invocation_step, step_outputs)
 
     def recover_mapping(self, invocation_step, progress):
-        progress.set_outputs_for_input(invocation_step)
+        progress.set_outputs_for_input(invocation_step, already_persisted=True)
 
 
 class InputDataModule(InputModule):
@@ -1304,17 +1304,6 @@ class ToolModule(WorkflowModule):
             raise Exception(message)
 
         return complete
-
-    def recover_mapping(self, invocation_step, progress):
-        outputs = {}
-
-        for output_dataset_assoc in invocation_step.output_datasets:
-            outputs[output_dataset_assoc.output_name] = output_dataset_assoc.dataset
-
-        for output_dataset_collection_assoc in invocation_step.output_dataset_collections:
-            outputs[output_dataset_collection_assoc.output_name] = output_dataset_collection_assoc.dataset_collection
-
-        progress.set_step_outputs(invocation_step, outputs)
 
     def _effective_post_job_actions(self, step):
         effective_post_job_actions = step.post_job_actions[:]
--- a/lib/galaxy/workflow/run.py
+++ b/lib/galaxy/workflow/run.py
@@ -405,7 +405,7 @@ class WorkflowProgress(object):
         else:
             return step_outputs[output_name]
 
-    def set_outputs_for_input(self, invocation_step, outputs=None):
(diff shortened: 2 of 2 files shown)


natefoo (approved):

bgruening:
This seems very promising :). Thanks @jmchilton! Do you want to communicate this in the release notes? Or over the admin, public server list? I guess Admins need to run this query, isn't it?

hexylena:
I could not get that query to complete in a reasonable amount of time on EU (~30+ minutes, I think it could run for a more hours. So I started reading and learnt some things I did not know about large `not in` queries!

Some references which I found helpful

- https://www.hagander.net/blog/speeding-up-not-in-66/
- https://stackoverflow.com/questions/17037508/sql-when-it-comes-to-not-in-and-not-equal-to-which-is-more-efficient-and-why/17038097#17038097
- https://stackoverflow.com/questions/17813492/postgres-not-in-performance

So the following rewrite might be useful for you @natefoo (and anyone else with lotsa rows)

```sql
 WITH exclude_list AS (
        SELECT max(w.id) as id
        FROM workflow_invocation_step_output_dataset_association as w
        GROUP BY workflow_invocation_step_id, dataset_id, output_name
)

DELETE
FROM workflow_invocation_step_output_dataset_association wisoda
WHERE NOT EXISTS (SELECT 1 FROM exclude_list WHERE wisoda.id = exclude_list.id);
```

Which completes in a <1-2 min on eu.

I ran this a couple times with `SELECT count(*) ... [NOT] EXISTS` and it seems to return the correct number of records.

I *believe* this is a valid translation of the original `not in` query, but this is based on the following:

- the exclude list is identical to the original not in subquery
- selecting the IDs of the rows which will be retained (`in` / `exists`) returns a list of the correct size
- For the above query, running that query `except all` the original subquery returns 0 rows, so the rows that will be retained are identical to the subquery.

Extraneous words here because I wanted to be sure I'm doing the right thing before I delete some `23782722` rows from our DB.

I've updated gxadmin with the new query https://github.com/usegalaxy-eu/gxadmin/commit/d9d5fc287d0040a9f53943a77e87ced72b39bbb1

jmchilton:
Amazing! Your's seems correct - I'd just ignore mine in gxadmin and use yours - clearly mine isn't performant at all which I was really worried about. Can you do the same thing for workflow_invocation_step_output_dataset_collection_association in that gxadmin query? I'm sure it has the same problem but probably just fewer rows I'd imagine given fewer tools that output collections? Thanks a million @erasche.

natefoo:
@jmchilton's was running for ~2.5 days before I interrupted it, @erasche yours ran in ~10 minutes. :smile: 

To be fair, John told me it would be slow.

hexylena:
Oh wow 2,5 days! You're way more patient than me, it took 30 minutes and I went "that can't be right, they're such simple queries, what the heck"

> Can you do the same thing for workflow_invocation_step_output_dataset_collection_association

Of course! I'll update the query and we'll run it tomorrow.

natefoo:
> they're such simple queries

37M rows in my case though.

hexylena:
available if y'all need it.

https://usegalaxy-eu.github.io/gxadmin/#/README.mutate?id=mutate-drop-extraneous-workflow-step-output-associations

PR #8424, opened 2019-08-07
Fix authentication provider exception not logged

BioinformaticsPlatformWIV-ISP:
This a copy of the PR  #7139 closed after a failing rebase tentative.
During an external provider authentication, log and reraise the exception properly instead of ignoring it when the provider raise one.

### DIFF
--- a/lib/galaxy/auth/__init__.py
+++ b/lib/galaxy/auth/__init__.py
@@ -64,8 +64,9 @@ class AuthManager(object):
                 if auth_results[0] is True:
                     try:
                         auth_return = parse_auth_results(trans, auth_results, options)
-                    except Conflict:
-                        break
+                    except Conflict as conflict:
+                        log.exception(conflict)
+                        raise
                     return auth_return
                 elif auth_results[0] is None:
                     auto_email = str(auth_results[1]).lower()
--- a/lib/galaxy/webapps/galaxy/controllers/user.py
+++ b/lib/galaxy/webapps/galaxy/controllers/user.py
@@ -14,6 +14,7 @@ from galaxy import (
     util,
     web
 )
+from galaxy.exceptions import Conflict
 from galaxy.managers import users
 from galaxy.queue_worker import send_local_control_task
 from galaxy.security.validate_user_input import (
@@ -82,7 +83,10 @@ class User(BaseUIController, UsesFormDefinitionsMixin, CreatesApiKeysMixin):
         """
         Does the autoregistration if enabled. Returns a message
         """
-        autoreg = trans.app.auth_manager.check_auto_registration(trans, login, password)
+        try:
+            autoreg = trans.app.auth_manager.check_auto_registration(trans, login, password)
+        except Conflict as conflict:
+            return "Auto-registration failed, {}".format(conflict), None
         user = None
         if autoreg["auto_reg"]:
             email = autoreg["email"]


nsoranzo:
@BioinformaticsPlatformWIV-ISP I've rebased your PR.

nsoranzo:
@galaxybot test this

BioinformaticsPlatformWIV-ISP:
@nsoranzo  Thank you for your help,  doing the rebase ourselves was a bad experience.

lib/galaxy/auth/__init__.py
```diff
@@ -64,8 +64,9 @@ def check_auto_registration(self, trans, login, password, no_password_check=Fals
                 if auth_results[0] is True:
                     try:
                         auth_return = parse_auth_results(trans, auth_results, options)
-                    except Conflict:
-                        break
+                    except Conflict as conflict:
+                        log.exception(conflict)
+                        raise
```
nsoranzo, line 69: If the `raise` here is meant to propagate the error to the user, you need to catch this exception at https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/controllers/user.py#L85

BioinformaticsPlatformWIV-ISP: The exception is now propagated to the user, i.e. converted to a message return by the autoregistration action inside of the user controller.

lib/galaxy/webapps/galaxy/controllers/user.py
```diff
@@ -82,33 +82,35 @@ def __autoregistration(self, trans, login, password):
...
-                    trans.handle_user_login(user)
-                    trans.log_event("User (auto) created a new account")
-                    trans.log_event("User logged in")
-                if "attributes" in autoreg and "roles" in autoreg["attributes"]:
-                    self.__handle_role_and_group_auto_creation(
-                        trans, user, autoreg["attributes"]["roles"],
-                        auto_create_groups=autoreg["auto_create_groups"],
-                        auto_create_roles=autoreg["auto_create_roles"],
+        if autoreg :
+            if autoreg["auto_reg"]:
```
nsoranzo: You can combine the 2 ifs and save one indentation level:
```python
        if autoreg and autoreg["auto_reg"]:
```

jvanbraekel: I thought about it . However, its imply that the next else have to be updated to an  `elif autoreg: `  (to be able to discriminate the case when autoreg exist and autoreg["auto_reg"] not).  That make this piece of code less intuitive because at the first reading,  its look like the 'same' check is done twice.
That being said, I can go for the 'elif cascade' if you think that indentation level save its wroth the change.

nsoranzo: I don't see an `else` at level of `if autoreg:` (and I don't think one is needed!), so I'm not sure I understand your comment.

jvanbraekel: My comment concern the `else: message = "No such user or invalid password."` If I go for a ` if autoreg and autoreg["auto_reg"]:` like you suggest this one cannot  be place at the same level as the 'if-and' without changes, otherwise `message` would be overridden in case of an conflict exception. It neither can be placed underneath the 'if-and' because it will never fire-up.

lib/galaxy/webapps/galaxy/controllers/user.py
```diff
@@ -82,33 +82,35 @@ def __autoregistration(self, trans, login, password):
         """
         Does the autoregistration if enabled. Returns a message
         """
-        autoreg = trans.app.auth_manager.check_auto_registration(trans, login, password)
+        try :
+            autoreg = trans.app.auth_manager.check_auto_registration(trans, login, password)
+        except Conflict as conflict:
+            message = "Auto-registration failed, {}".format(conflict)
+            autoreg = None
```
nsoranzo:
```suggestion
            return message, None
```

lib/galaxy/webapps/galaxy/controllers/user.py
nsoranzo: I see, thanks for explaining. I'd suggest an early return then, see suggestion above.

nsoranzo:
@galaxybot test this

nsoranzo (approved):

mvdbeek:
Thanks @BioinformaticsPlatformWIV-ISP!

Issue #8399, opened 2019-08-01
Browser tab job updates


astrovsky01:
As with certain websites such as gmail, galaxy should have some way of alerting users who are using multiple tabs of job completions (or failures) in the tab box at the top of the screen.

Credit @martenson for the image.

![Pull_Requests_·_galaxyproject_galaxy_and_Inbox_-_emulatorer_gmail_com_-_Gmail](https://user-images.githubusercontent.com/40246333/62307243-1aa99380-b438-11e9-85c8-9d4e9d770fcb.png)

martenson:
Would indicate the number of jobs finished since you last time you checked the tab. Idea is to show this when the tab is in the background, once user switches to it, the changed title should go away.

astrovsky01:
Alternative/possible addition onto this: true notifications from the browser on job completion/failure

<img width="1680" alt="Screen Shot 2019-08-06 at 4 41 34 PM" src="https://user-images.githubusercontent.com/40246333/62575622-1ed71600-b869-11e9-8f56-3ed75988fbcc.png">


*Update: notifications enabled
https://github.com/galaxyproject/galaxy/pull/8427

PR #5439, opened 2018-02-01
Dynamic post-processing actions

anatskiy:
The PR adds a new feature by executing a custom Python function after a workflow run. In this example, Galaxy sends a simple desktop notification:

screenshot from 2018-02-01 20-41-13

screenshot from 2018-02-01 20-42-45

All credits go to @bgruening, I just adapted the 3-year-old feature to the latest dev.

PR #8427, opened 2019-08-07
Notifications on job completion

astrovsky01:
This PR adds push notifications on job completion. The toggle for the setting is in user preferences.

Next steps:
* User select which kinds of jobs (on data upload or not?)
* User preferences based on workflows (every tool, or only on completion of final tool?)


astrovsky01:
Currently will notify on any job completion, including cancellation.

astrovsky01:
Updates:
* Altered message on job pass/fail
* No message on job cancellation
* Red icon for job failure
** Red icon will also be used for browser tab updates as referenced in https://github.com/galaxyproject/galaxy/issues/8399

astrovsky01:
Next task: return the tab to original favicon and remove job completion number on page view.

astrovsky01:
Tab now resets on viewing the page, also linked the tab icon's ability to change with the user's notification preference.

ezgif com-video-to-gif (1)

bernt-matthias:
Like the general idea, but I could imagine that popup notifications might become annoying if you run a workflow (which have their own email notification system). In particular if many jobs are processed in parallel.

Maybe it's an idea to disable it for jobs in workflows? Or replace the pop-up with a more decent signal, eg make the tab greenish/reddish if there are new finished/failed jobs. 

In general I like the idea of a notification system. Eg a checkbox next to the execute button to get mail notification could be really handy for long running tools.

astrovsky01:
@bernt-matthias the next idea is to make it so push notifications are a job-specific thing, where a user would have a check box next to the execute button to request a notification on update. In that case, the notification would remain on screen until the user specifically dismisses it. The tab's update will be for all jobs running, so a number completed will show in the tab's text if completed while tabbed away. The color change of the favicon will also only be on notification request.

bernt-matthias:
Agreed. But I'm thinking more of jobs that run for days or weeks. Then I would find a mail handy. Otherwise the user would need to repeatedly connect to galaxy to see the job status. 


Instead of a checkbox I could also imagine a user specific option specifying a time span for jobs to send an email upon completion.

astrovsky01:
I was thinking of using SLURM functionalitu for the email, and I like the time span idea.

astrovsky01:
I really like that idea (maybe a slider on the preferences?) I think SLURM has that functionality, but not all HPCs we run it all do, so I'll check out alternatives!

dannon:
@astrovsky01 @bernt-matthias Check out the workflow editor's "email notification" action, which sends an email when a particular step in the workflow completes.  It's implemented as a post job action and the same functionality could be used for individual jobs without much extra work.

bernt-matthias:
@dannon great idea

astrovsky01:
@dannon I was looking at the account verification email, but that would probably work better. Thanks!

dannon:
@astrovsky01 Yep, it should.  There's already a hook to execute PJAs on job completion, so you'll just need to add a way to *create* new PJAs on individual job submission.

bgruening:
@astrovsky01 this might be also interesting and related. Would love to get this in as well: https://github.com/galaxyproject/galaxy/pull/5439

dannon:
@galaxybot test this

PR #8429, opened 2019-08-07
[WIP] Use cElementTree for xml macro loading.

jmchilton:
Speeds up Python 2 macro expansion a good bit for unit tests.

### DIFF
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -10,7 +10,10 @@ import tempfile
 import threading
 from collections import OrderedDict
 from datetime import datetime
-from xml.etree import ElementTree
+try:
+    from xml.etree import cElementTree as ElementTree
+except ImportError:
+    from xml.etree import ElementTree
 
 import packaging.version
 import webob.exc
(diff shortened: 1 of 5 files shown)


lib/galaxy/tools/__init__.py
```diff
@@ -10,7 +10,10 @@
 import threading
 from collections import OrderedDict
 from datetime import datetime
-from xml.etree import ElementTree
+try:
+    from xml.etree import ElementInclude, cElementTree as ElementTree
```
nsoranzo:
```suggestion
    from xml.etree import cElementTree as ElementTree
```

lib/galaxy/tools/__init__.py
```diff
@@ -10,7 +10,10 @@
...
+except ImportError:
+    from xml.etree import ElementInclude, ElementTree
```
nsoranzo:
```suggestion
    from xml.etree import ElementTree
```

nsoranzo:
The integration tests fail because of mixing the 2 types of trees, see https://stackoverflow.com/questions/16467640/python-etree-parser-isnt-appending-an-element

PR #8430, opened 2019-08-07
optimize unrestricted library dataset permission checks

martenson:
this should optimize permission checking for unrestricted datasets

cc @pvanheus


pvanheus:
+1 - in my little bit of testing, this shaves 0.01 seconds / item off the loading of library contents.

PR #5920, opened 2018-04-19
Downgrade some annoying warnings

hexylena:

PR #8439, opened 2019-08-09
 Change 'a when tag has not been defined for' level to debug

mvdbeek:
We've discussed this for a while going back and forth over this.
Ultimately it's probably not that useful for admins and it clutters
the warnings. There is a way to configure logging in planemo
based on the python module (galaxyproject/planemo@1eaff26)
and planemo lint will catch this anyway.

xref https://github.com/galaxyproject/tools-iuc/pull/2523 / https://github.com/galaxyproject/galaxy/pull/5920/

replaces https://github.com/galaxyproject/galaxy/pull/5920/

### DIFF
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -1131,7 +1131,7 @@ class Tool(Dictifiable):
                     test_param_input_source = input_source.parse_test_input_source()
                     group.test_param = self.parse_param_elem(test_param_input_source, enctypes, context)
                     if group.test_param.optional:
-                        log.warning("Tool with id %s declares a conditional test parameter as optional, this is invalid and will be ignored." % self.id)
+                        log.debug("Tool with id %s declares a conditional test parameter as optional, this is invalid and will be ignored." % self.id)
                         group.test_param.optional = False
                     possible_cases = list(group.test_param.legal_values)  # store possible cases, undefined whens will have no inputs
                     # Must refresh when test_param changes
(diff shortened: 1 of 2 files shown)


hexylena (approved):

mvdbeek:
Alright, https://github.com/galaxyproject/galaxy/pull/8439/commits/c6c92dbc33d2cc153867e842b6e8168022c17d95 worked at least once. I'll trigger a few more tests. We still shut down all the test app objects a second time because of `atexit.register(app.shutdown)`, but `atexit.unregister` doesn't exist on python 2, and I don't think this is the root of the problem, unless this somehow keeps the app objects around and that would lead to accumulating sqlalchemy connections somehow.

mvdbeek:
@galaxybot test py3

mvdbeek:
@galaxybot test this

mvdbeek:
@galaxybot test py3

mvdbeek:
@jmchilton what do you think about https://github.com/galaxyproject/galaxy/pull/8439/commits/c6c92dbc33d2cc153867e842b6e8168022c17d95 ? It seems using a separate db for connections solved the issue as well, but this one in isolation also worked a couple of times now.

jmchilton:
I feel like I like my solution better? I don't feel strongly either way though - sorry I didn't see this before working on a different solution :(

mvdbeek:
Alright, dropped the commit and the verbose logging, this should be ready.

PR #8440, opened 2019-08-09
add assert for n_lines in tool output verifications 

bgruening:
This can be handy in tool and workflow testing.

### DIFF
--- a/lib/galaxy/tool_util/verify/asserts/text.py
+++ b/lib/galaxy/tool_util/verify/asserts/text.py
@@ -23,7 +23,6 @@ def assert_has_line(output, line):
 def assert_has_n_lines(output, n):
     """Asserts the specified output contains ``n`` lines."""
     n_lines_found = len(output.splitlines())
-        pass
     assert n_lines_found == int(n), "Expected %s lines in output, found %s lines" % (n, n_lines_found)
 
 
--- a/lib/galaxy/tool_util/verify/asserts/text.py
+++ b/lib/galaxy/tool_util/verify/asserts/text.py
@@ -26,6 +26,7 @@ def assert_has_n_lines(output, n):
         pass
     assert counter == int(n), "Expected number of lines (%s) do not match with obtained number ('%s')" % (n, counter)
 
+
 def assert_has_text_matching(output, expression):
     """ Asserts the specified output contains text matching the
     regular expression specified by the argument expression."""
--- a/lib/galaxy/tool_util/verify/asserts/text.py
+++ b/lib/galaxy/tool_util/verify/asserts/text.py
@@ -20,6 +20,12 @@ def assert_has_line(output, line):
     assert match is not None, "No line of output file was '%s' (output was '%s') " % (line, output)
 
 
+def assert_has_n_lines(output, n):
+    """ Asserts the specified output if the number of lines match the given number."""
+    for counter, line in enumerate(output.split('\n')):
+        pass
+    assert counter == int(n), "Expected number of lines (%s) do not match with obtained number ('%s')" % (n, counter)
+
 def assert_has_text_matching(output, expression):
     """ Asserts the specified output contains text matching the
     regular expression specified by the argument expression."""
--- a/lib/galaxy/tool_util/verify/asserts/text.py
+++ b/lib/galaxy/tool_util/verify/asserts/text.py
@@ -21,7 +21,7 @@ def assert_has_line(output, line):
 
 
 def assert_has_n_lines(output, n):
-    """ Asserts the specified output if the number of lines match the given number."""
+    """Asserts the specified output contains ``n`` lines."""
     for counter, line in enumerate(output.split('\n')):
         pass
     assert counter == int(n), "Expected number of lines (%s) do not match with obtained number ('%s')" % (n, counter)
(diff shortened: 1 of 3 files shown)


test/functional/tools/validation_default.xml
```diff
@@ -25,6 +25,7 @@
       <output name="out_file1">
          <assert_contents>
             <has_line line="__dq__ X echo __dq__moo" />
+            <has_n_lines n="1" />
```
bernt-matthias, line 28: Should the has line not be:

```
__dq__ ; echo __dq__moo
```

bernt-matthias:
Like it. But would be even more useful if a range could be given. E. G.
- min max
- n +- x
- specifying a list of allowed numbers, eg explicitly or start:stop[:increment]

Like the last option most...

bgruening:
What should the increment do? A fuzzy match like the `delta` from sim_size might make sense, although I can not come up with an immediate use-case. But increment?

bernt-matthias:
Maybe the number of lines must be even (fasta) or a multiple of 4 (fastq).

bernt-matthias:
All this might as well be solved with a regexp on the whole file `([^\n]+\n){min,max}`. But I guess this might be to imperformant.

.. 

Just collecting ideas.

lib/galaxy/tool_util/verify/asserts/text.py
```diff
@@ -20,6 +20,13 @@ def assert_has_line(output, line):
     assert match is not None, "No line of output file was '%s' (output was '%s') " % (line, output)
 
 
+def assert_has_n_lines(output, n):
+    """ Asserts the specified output if the number of lines match the given number."""
```
mvdbeek:
```suggestion
    """Asserts the specified output contains ``n`` lines."""
```

lib/galaxy/tool_util/verify/asserts/text.py
```diff
@@ -20,6 +20,13 @@ def assert_has_line(output, line):
...
+    for counter, line in enumerate(output.split('\n')):
```
mvdbeek:
```suggestion
    n_lines_found = len(output.splitlines())
```
will also work for other types of newlines, if we don't want that we can also use `output.count('\n')`.

lib/galaxy/tool_util/verify/asserts/text.py
```diff
@@ -20,6 +20,13 @@ def assert_has_line(output, line):
...
+        pass
+    assert counter == int(n), "Expected number of lines (%s) do not match with obtained number ('%s')" % (n, counter)
```
mvdbeek:
```suggestion
    assert n_lines_found == int(n), "Expected %s lines in output, found %s lines" % (n, n_lines_found)
```

test/functional/tools/validation_default.xml
mvdbeek: There's a basic sanitizer that is active on all ToolParameter instances, that's mentioned in https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-inputs-param-sanitizer.
So to get unsanitized values you need to add a Sanitizer and allow certain values.
This happens via the InputValueWrapper when accessing the variable in Cheeath at https://github.com/galaxyproject/galaxy/blob/7830c54a60369e5faf47ee1c55f2de3604e8251e/lib/galaxy/tools/parameters/basic.py#L216

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -1602,6 +1603,12 @@ module.
           </xs:documentation>
         </xs:annotation>
       </xs:element>
+      <xs:element name="has_n_lines" type="xs:anyType">
+        <xs:annotation>
+          <xs:documentation><![CDATA[Asserts a file matching the specified n (``lines``) in the output (e.g. ``<has_n_lines n="3" />``).]]>
```
```suggestion
          <xs:documentation><![CDATA[Asserts that an output contains ``n`` lines, e.g. ``<has_n_lines n="3" />``.]]>
```

mvdbeek (commented):
Very useful, thanks @bgruening.

bernt-matthias:
@mvdbeek OK I understand the effect. But I don't get the rationale of replacing X by default.

lib/galaxy/tool_util/verify/asserts/text.py
bgruening: I was just concerned about memory usage, but we assume that test data stays small?

bgruening:
@bernt-matthias max and min might be useful in a few cases but this should be a different assert type?

bernt-matthias:
I always like to generalize :) 

My rationale is that with min=max you would have the behavior of this assert. Then this one would be superfluous. 

Still might be convenient for tool developers to have both, but then my choice for implementing it would be to have one for testing min max and one for testing for an exact value (but the latter would just call the code of the former).

mvdbeek:
@galaxybot test this

jmchilton:
I'm with @bgruening - I think we should merge as is and push min/max into a new assertion.

mvdbeek:
> @mvdbeek OK I understand the effect. But I don't get the rationale of replacing X by default.

It's the other way round, non-safe things get replaced by X. So this includes `;`

lib/galaxy/tool_util/verify/asserts/text.py
```diff
@@ -20,6 +20,13 @@ def assert_has_line(output, line):
     assert match is not None, "No line of output file was '%s' (output was '%s') " % (line, output)
 
 
+def assert_has_n_lines(output, n):
+    """Asserts the specified output contains ``n`` lines."""
+    n_lines_found = len(output.splitlines())
+        pass
```
mvdbeek:
```suggestion
```

mvdbeek (approved):
Nice, thanks @bgruening!

bgruening:
Thanks for the merge!

Issue #6306, opened 2018-06-08
Collection Models and State Revision


jmchilton:
A 2018 revision of collection models and job state are probably in order along the lines of https://github.com/galaxyproject/galaxy/pull/5013. 

In particular I think we need to:
 - Stop associating every input collection dataset with the jobs it is being used to create - we are tracking collection input associations and that should be enough in theory hopefully. Just associate the collection. What we are doing right now just doesn't scale well. This won't be easy because job polling depends on these datasets being present.
 - Fix on_string for collection jobs at the same time. ``on_string`` is ridiculous for tools with collection inputs because these datasets are picked up as inputs and so random collection contents HIDs are used to give names to collections.
 
This also would bring us a little closer to allowing dataset collection contents be something that isn't an HDA (https://github.com/galaxyproject/galaxy/issues/1810).

jmchilton:
Note to self - here is optimization WIP that didn't help because of this problem https://github.com/jmchilton/galaxy/tree/1805_tools_ready_fix.

PR #5091, opened 2017-11-29
Render arbitrarily large collections in the UI.

jmchilton:
This builds on #5038 (render arbitrarily nested lists in the UI) and fixes #3795 (which is both a usability problem with large collections and serious backend memory issue). Only around 1000 elements of a collection will be rendered now and these will be fetched much more efficiently. It'd be nice if we could paginate or progressively fetch more elements beyond that - but that should probably wait for a major rewrite and be tied with search and such to make navigation of such collection possible.

This problem is tackled along three prongs. 

1. Reduces the transfer size, number of database fetches, and number of database joins to pull out the history collection structure for rendering a collection in the UI. a475f9c
2. Introduces the concept of "fuzzy count" limit when fetching large collections - to limit the order-of-magnitude-per-level-of-depth of the collection fetched via the API while still fetching the collection at every level of depth for nested collections. 62c3e9c
3. Updates the UI for more consistent collection subtitles, display of element counts, and warns the user if only a subset of elements are being display.

This screenshot generated on Jenkins from this PR is of a collection that is rendering a more limited number of elements (the test case overrides the client to reduce the fuzzy count limit from 1000 to 2 to verify the limiting works).

Issue #3787, opened 2017-03-21
Ensure that similarly used objects have unique ids client-wide.


martenson:
A bug unearthed in https://github.com/galaxyproject/galaxy/pull/3559 points to some issues we might have with id handling. There a library was populated with datasets from a collection that had the same ID as one of the datasets inside. Given how BBone collection works (enforces uniqueness on it) it treats them as one object and discards one.

There were some hints that history already uses some solution to this, maybe based on `type_id` property of hd(c)a.

The solution to this needs to be system-wide as imports/exports will expect same dogma.

tagging people that were in the gitter debate: @jmchilton @dannon @mvdbeek

hexylena:
Is this still an issue? Will it be fixed in the history rewrite or is it already better with Vue stuff?

martenson:
I believe this is still an issue, however I did not review the rewrite, it should probably fix it.

hexylena:
Cool :+1:

xref #8441

martenson:
I think we should verify though.

mvdbeek:
It's not a problem anymore with vue, and I think we can use type hints on the particular entity ids to avoid accidentally using the wrong id.

PR #8441, opened 2019-08-09
History component refactor

Nerdinacan:
This is a complete rebuild of the right hand history panel. Since it's such a big code change, I built it as a beta feature that is accessed by first logging in, then going to the giant Gear-Menu-Of-Doom&trade; and picking "Beta History Panel" at the bottom of the list.

Definitely still a WIP. There's a lot to look at here, and I'm still debugging some of it. Unit tests are coming in the next commit.

**Goals:**
Cache and manage API requests to reduce server/client traffic.
Modernize the UI code
Make a testable UI
Pave the way for more extensive history feature additions in the future (history graph, real-time updates)
Implement a reactive data model.

### Notable changes

**Database Timestamp Triggers**

One way to reduce server traffic is to compare dates and only ask for a smaller set of results. To that end, I have implemented triggers to keep the history table's update_time updated as changes occur to its contents. There are still more refinements to be done here.

Instead of creating some kind of complex ORM association object in the server side code, I thought it might be more performant to simply implement a couple timestamp triggers on the dataset, HDA, HDCA tables that update history when they change. I think this is one of the few uses for database triggers that really makes sense.

**RxJS**

RxJS is the primary mechanism behind the polling and manual content retrieval processes. Although RxJS can be a little more sophisticated to parse than something as simple as a promise, observables are vastly more powerful than promise based query implementations and pave the way for more advanced client-side data options such as real-time web-socket or push-server updates, collaborative editing, etc.

The most involved parts are currently in the model/ContentLoader objects, and I will be providing extensive documentation about how these objects work. These observables control when we request data from the api either through human interaction with the UI or through automatic server-updates, currently implemented through polling (but since incoming changes are represented as observables, it will be trivial to transition to any other update mechanism such as a websocket.)

**RxDB / IndexedDB**

HDA/HDAC/Dataset and Collection data are stored in [IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API) using an NPM package [RxDB](https://github.com/pubkey/rxdb) which allows our components to generate and subscribe to live queries that point at the local cached contents. These live query observables will emit updates when the cache does.

Since IndexDB is local and shared across browser tabs, this means UI data updates are instant across tabs. You won't be able to have stale data in one tab that doesn't jive with another tab any more. Nor is it necessary to ever "refresh" your history. 

**Forthcoming:**
Unit tests
Selenium tests
Documentation for the ContentLoader, polling cycle, caching mechanisms
Possibly implement shared layout components with tool panel redesign PR


bgruening:
Awesome! Thanks @Nerdinacan, now I have something to play with over the weekend :)
Which kind of feedback are you interested in the first round?

bgruening:
Hi @Nerdinacan,

I started to poke around with it and just wrote down what I found. Hopefully, this is useful.


- the migration does not work on SQLite (still the same problem as in your comments, a warning about multiple statements in SQLite)
- the grey color for a dataset is identical with the history bg color
- reload button is missing
- dataset upload does not work, the window is not opening and I do not see any error in the JS console :(
    - it works in the legacy view
- if I edit attributes like title/annotation in the dataset, the size of the dataset is changing a lot and everything is jumping around a lot
- 'operations on multiple datasets' the checkbox button looks confusing to me as it always has the arrow in it when everything is unselected it still looks like everything is selected
- I find the 3 gear symbols more confusing than one
- if you collapse the dataset, it fades out the preview at first and then collapses it, these are two steps and it jumps too much for my taste
- the dataset preview should go over the entire size of the dataset even if the previewed data is really small (e.g. one column)
- on SQLite the color state update does not seem to work, if I schedule a workflow 2 steps, first stays yellow second stays grey ... clicking on the grey turns it green, clicking afterward on the (still) yellow one it turns also green
- saved histories view does not change the history
- scheduling a workflow with 'Send results to a new history' does not update this awesome quick-history-select
- talking about this quick-history-select, this is great!!! I saw that if you have a tool open or a run-workflow page and you change the history the tool/workflow does not
  update the dataset-picker
- multiple history view button is missing -> Oh found it in one of the menus.
- if I go into a collection view, it is empty ... classical history view shows them
  - "TypeError: e is null" in JS Console
- deleting a collection will open it in my case
- with Chrome (66) I see not boxes at all :( (no error in the console)

There are so many nice ideas included that makes a real difference in usage. Thanks Mason!

hexylena:
We should verify that #3787 is fully addressed here

Nerdinacan:
Thanks for peeking @bgruening! Yeah there's lots of issues to iron out still. Thanks for the quick QA run!

Mainly I wanted you guys to see the rough architecture and give me your thoughts on the new tech I threw into the mix, specifically RxJS and RxDB/IndexedDB. I think they're the right tools for the job, but would love your thoughts on the matter.

Nerdinacan:
> I find the 3 gear symbols more confusing than one

I agree that 3 gears is not ideal. But mainly I wanted to break that monster menu into 3 specific contexts: One set of things that operates on histories, one set of things that operates on the current history, and one set that operates on the datasets.

I think the fact that they all look like gears is probably the worst part, but I'm happy to address cosmetics like this once the bugs are ironed out of the central mechanism.

Nerdinacan:
@erasche 

> We should verify that #3787 is fully addressed here

I will check it out. I'm not using the traditional BBone model objects, so it is unlikely that the new code will have the same issues, however, not all of the old Bbone models have been removed from the client as of yet. I'll see if it's something that intersects with this PR.

martenson:
@Nerdinacan fwiw on `dev` we have slimmed the menu down substantially, and e.g. the operations on all histories are all gone.
<details><summary>screenshot</summary>

Galaxy___Test
</details>

Nerdinacan:
> @Nerdinacan fwiw on `dev` we have slimmed the menu down substantially, and e.g. the operations on all histories are all gone.

Thanks @martenson I'll trim the fat off my menus and see what's left. I still think there's value to associating types of operations with specific blocks of the interface, but it'll be nice to not have to juggle so many options.

martenson:
>I still think there's value to associating types of operations with specific blocks of the interface

I like what you did with the separation, I think we will collectively iron it out and it will end up being much better than the menu of death.

guerler:
Thank you for working on this. We added a small feature to disable the visualizations button in the history panel if admin decides to set a configuration flag. #8450.

Nerdinacan:
@erasche About the polling, I actually poll for a lot less data now because I filter pretty extensively, but I did suggest to Jeremy earlier in this process that an exponential die-off with user inactivity (and/or no result changes), might be a cool thing to do for the polling. I think it's a decent idea. 

I also thought it might be a good idea to have a round-robin setup for situation in which 5 history panels are up, and presumably all would be submitting poll requests. That seems like it wouldn't be hard to add in either since it's just a matter of combining observables now.

But the better answer to all these problems is a websocket. :)

> We added a small feature to disable the visualizations button in the history panel if admin decides to set a configuration flag. #8450.

@guerler  I will be sure to add that toggle in as I work through the bugs. (I decided not to include visualization data with the standard queries and only get it on demand.) Thanks!

Nerdinacan:
In the interest of getting more of this functionality into the light of day, I'm temporarily removing the complicated virtual scroller I've been banging my head against.  I think this feature is going to be necessary for windowing through the larger histories and collections, but since it's a complicated beast, I'll be addressing that in my next PR once the rest of the bugs are ironed out. I just want to get the rest of this code tested and out there.

In order to do that virtual scrolling, we'll also need to address the collection API since there's currently no way for us to granularly page through dataset collections in order to even implement a virtual scroller.

Nerdinacan:
> Thank you for working on this. We added a small feature to disable the visualizations button in the history panel if admin decides to set a configuration flag. #8450.

I created a config Vuex store module recently (#9019) and am now reading that config option to show/hide the visualizations button.

mvdbeek:
I've removed the WIP label in the title, otherwise @galaxybot puts the label back, hope that is ok.

martenson:
notes from a first pass on this version:

* the history does not seem to respect 'current history', when I switch current at multi-history view and then go back to analyze data change is not reflected
* history icons, hda icons, and other elements are missing BS tooltips, I think especially for tiny icon elements with no labels and in a redesign like this we should provide the tooltips often and not just using the 'html' title that has considerable delay
* inline editing of hda names, annotations, and tags feels clunky, examples:
    * If I type and click elsewhere it is not saved
    * Sometimes if I click elsewhere it will hide the textbox, other times it will not
* Green/red/blue background with grey text is hard to read (Tag field, hid), maybe more contrast in these cases would help
* Link to "tool help" does not work (actually removing the icon would be best imo)
    * got `http://127.0.0.1/api/tools/upload1/build?history_id=null 400 (Bad Request)`
* "edit history tags" when inside of collection feels unnecessary
* "Back one" to get out of collection could be more prominent, this is operation that everybody eventually clicks on when going back to top history
* Collapsed hdas could take less horizontal space to fit more, it feels like there is plenty of empty canvas
* The 'expand' and 'select' icons look almost identical while next to each other at the same time.
Galaxy
* Some file previews have green background (escaping?), same with dataset errors - we could present them formatted
* hda details (1 line, 1 comments; format tabular; database?) all look alike with unusual spacing; we could use font weight as in old history interface to help navigation
* I did not find a way how to show hidden or deleted datasets
* I did not find a 'refresh' button, I understand we hope it will never be needed, but experience says we were never there yet, are we now?
* If I 'select all' including a collection, I cannot unselect it
    * `[Vue warn]: Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value. Prop being mutated: "selected"`
* the 'fast switch' history select changes the height of the masthead when clicked, so everything "jumps"; also scrolling the dropdown scrolls the whole page, not just the history select; limiting max height of this element could be useful
Galaxy

p.s. maybe it is time to switch from the 'gear' menu to the established 'hamburger' menu?

bgruening:
Cool! Thanks @Nerdinacan! And thanks @martenson very detailed.
A few points from me:

* the database can not be set, previously you could click the `?`
* the preview could expand to use 100% of space. With small data it looks like that:
grafik
* in general like Martin mentioned there is a lot of white space
* there is also a lot white space between the dataset number and the first icons. If the name should not be there (like in the old one) we could display all icons always?

In general, this works pretty well in my hands and I only found small things. I would be in favor to roll this out in the next release and gather advanced feeback from large instances and power users.

martenson:
@bgruening are you suggesting rolling out as default or rolling out as opt-in beta? I can see the latter, but can hardly imagine the former.

bgruening:
The latter, I think @Nerdinacan did a great job to make this optional and this gives us some time to collect a lot of feedback during 1 or even 2 cyclers.

Nerdinacan:
@martenson @bgruening 

Awesome, thanks for all the analysis! I'll do a quick design pass and address some of these things. A couple notes though:

> the history does not seem to respect 'current history', when I switch current at multi-history view and then go back to analyze data change is not reflected

I actually did that by design. I am trying to imagine a good reason that the server should be in charge of what the current history is. Perhaps there is a subtlety to the server-side internals that i am not aware of? In general, I think server persistence of current history is a barrier in the drive to make the API less state dependent.

Can you explain to me why it is currently necessary for the server to track which history the user is on? It wouldn't be hard to restore that feature.

> Collapsed hdas could take less horizontal space to fit more, it feels like there is plenty of empty canvas

The goal here was to put the entire menu into the same place every time, which led to an admittedly taller HDA/HDAC box, but less weirdness with operations being found in multiple places. I can skootch the design around, but was trying to make it less "jumpy" as it expands and contracts by re-using as many elements as possible.

> I did not find a way how to show hidden or deleted datasets

If you HAVE hidden datasets, a dataset filter button is available when you open the filters. Same for deleted. Maybe they should be there but disabled if you have no deleted and/or hidden sets. 

image


> I did not find a 'refresh' button, I understand we hope it will never be needed, but experience says we were never there yet, are we now?

If you need a refresh button in the UI, the client (as designed) is broken. :) If the client is broken and you need a refresh button, your browser comes with one. :) Just kidding, I can put something in there that means "clear the client cache and reload fresh history" if you want one. Might be good for debugging.

> the 'fast switch' history select changes the height of the masthead when clicked, so everything "jumps"; also scrolling the dropdown scrolls the whole page, not just the history select; limiting max height of this element could be useful

That's a new one for me, what browser were you using, and also what browsers would you like me to target? Up to now it's been all about just getting the parts moving.


I'm happy to spend a couple more days looking at these issues before rolling it out, even in beta form. What's a couple more?

martenson:
> That's a new one for me, what browser were you using, and also what browsers would you like me to target? Up to now it's been all about just getting the parts moving.

I use chrome. I suspect you can reproduce everywhere but you have to have a lot of histories, so they fill all horizontal space of the screen in the list.

I think our targets are Chrome and Firefox and Edge. We don't stay on the edge but also don't do long term support of old crap.

martenson:
> Can you explain to me why it is currently necessary for the server to track which history the user is on?

I am not sure I have a good answer. Maybe things like data_source tools when backend is receiving an async call from remote server, I doubt the remote party has a history id so they may need some target.

I understand client-only current history, but then we will adjust the multi-history, and bunch of other places to respect that one too. Probably not needed for beta opt-in though.

jxtx:
> I actually did that by design. I am trying to imagine a good reason that the server should be in charge of what the current history is. Perhaps there is a subtlety to the server-side internals that i am not aware of? ...

> Can you explain to me why it is currently necessary for the server to track which history the user is on? It wouldn't be hard to restore that feature.

The only reason is that this was originally written in 2006 and everything was done on the server side. It was the era of "progressive enhancement" where nothing important was done in Javascript and everything had to work without it.

But why do we need to keep it? Because *many* things depend on knowing the current history. Things like the run workflow form, the tool form, charts, ... 

Until everything is rewritten this probably needs to stay in sync.

martenson:
> Because many things depend on knowing the current history

I am not sure how many things depend on it in the backend.

Nerdinacan:
> need to keep it? Because _many_ things depend on knowing the current history. Things like the run workflow form, the tool fo

@martenson @jxtx 

Well, I can keep it synced in Vuex for now. The specific parts of the legacy client Martin mentioned, I can hunt down and point at Vuex, that'd be a start I guess. Since it's only a beta panel, how about I do that in a separate PR later? It's going to touch the legacy code in a bunch of places in an already complicated PR.

martenson:
@Nerdinacan As I said above I am fine with beta just ignoring this. However note that the data_source example I mentioned is a backend use of current history. I believe @jxtx examples were all client-side.

jxtx:
I'm not sure all my examples are client. For example, tool forms: `trans.history` is still used on the server side to populate `GenomeBuildParameter`, `BaseDataToolParameter`, ... to populate the available options.

Nerdinacan:
> Link to "tool help" does not work (actually removing the icon would be best imo)
got http://127.0.0.1/api/tools/upload1/build?history_id=null 400 (Bad Request)

This is what's supposed to happen with the tool help. Which tool were you looking for help on, specifically? It's dependent on data that's available, or maybe I'm caching it incorrectly.

image

martenson:
I think these are just freshly uploaded datasets. First click after startup (and any subsequent).
Galaxy
Galaxy

martenson:
With the latest version I have trouble loading any toolform

```
vue.runtime.esm.js?2b0e:619 [Vue warn]: Error in v-on handler: "TypeError: obj.on is not a function"

found in

---> <Tool> at galaxy/scripts/components/Tool.vue
       <ToolSection> at galaxy/scripts/components/ToolSection.vue
         <ToolBox> at galaxy/scripts/components/ToolBox.vue
           <Root>
```

```
vue.runtime.esm.js?2b0e:1888 TypeError: obj.on is not a function
    at tryCatchOn (backbone.js?ab5c:178)
    at child.Events.listenTo (backbone.js?ab5c:152)
    at child.initialize (tool-form-base.js?87e6:31)
    at child.Backbone.View (backbone.js?ab5c:1300)
    at child [as constructor] (backbone.js?ab5c:2060)
    at new child (backbone.js?ab5c:2060)
    at child.initialize (tool-form.js?c58c:21)
    at child.Backbone.View (backbone.js?ab5c:1300)
    at new child (backbone.js?ab5c:2060)
    at child._loadToolForm (AnalysisRouter.js?ccc8:400)
    at child.home (AnalysisRouter.js?ccc8:375)
    at child.execute (router.js?8f4f:47)
    at Object.eval [as callback] (backbone.js?ab5c:1678)
    at eval (backbone.js?ab5c:1961)
    at Function._.some._.any (underscore.js?17fb:286)
    at Backbone.History.loadUrl (backbone.js?ab5c:1959)
    at Backbone.History.navigate (backbone.js?ab5c:2024)
    at child.navigate (backbone.js?ab5c:1695)
    at child.push (router.js?8f4f:35)
    at VueComponent.open (Tool.vue?2993:52)
    at click (Tool.vue?96bf:22)
    at invokeWithErrorHandling (vue.runtime.esm.js?2b0e:1854)
    at HTMLAnchorElement.invoker (vue.runtime.esm.js?2b0e:2179)
    at HTMLAnchorElement.original._wrapper (vue.runtime.esm.js?2b0e:6911)
```

martenson:
@Nerdinacan At the committers meeting we've decided to push this to 20.05, mainly because the inconsistencies with 'current history' concept were considered very important by some. Namely users operating on the tool form would be exposed to a lot of confusion between what they see in the history and what options they are getting in the tool form. 

I'd like to give you some examples but cannot open tool form on this branch at the moment (see above).

client/galaxy/scripts/components/History/model/DatasetCollection/DatasetCollection.js
```diff
@@ -0,0 +1,110 @@
+/**
+ * Our API is very inconsistent which makes re-using components difficult.
```
jmchilton: Can you create an issue for what you'd like to see different - or maybe just a comment on https://github.com/galaxyproject/galaxy/issues/6306 or https://github.com/galaxyproject/galaxy/issues/1810 - which ever seems more appropriate?

Is this a formatting issue or a fundamental issue (attribute names don't match up) with the models (frustrating that the history panel uses the element identifier in lieu of the dataset name when element is in a collection)?

client/galaxy/scripts/components/History/model/caching/schema/datasetCollection.schema.js
```diff
@@ -0,0 +1,81 @@
...
+        treenode: {
+            type: "object",
+            properties: {
+                element_index: { type: "integer" },
+                element_identifier: { type: "string" },
+                element_type: { type: "string" },
+                model_class: { type: "string" },
+                id: { type: "string" },
+
+                // find and kill the person who named this 'object'
```
jmchilton: It is impressive code but you need to not threaten to kill me in it!

client/galaxy/scripts/components/History/model/DatasetCollection/DatasetCollection.js
Nerdinacan: @jmchilton 

I will do that. There are a couple problems, I'll gather up all my whining into one place for you.

I think the biggest issue is that a dataset that is returned in a collection is not the same format as a dataset that exists on a tree, and it is not possible to configure the fields that are returned in a consistent way.

The tree format of the dataset collection return is particularly clumsy in my opinion.

And if I recall one of the things that's making this thing very hard to scale for large collections is the inability to page (limit/offset) through the contents of a dataset collection.

I'll do a more thorough summary in an issue as you have suggested though, thanks!

Nerdinacan:
> @Nerdinacan At the committers meeting we've decided to push this to 20.05, mainly because the inconsistencies with 'current history' concept were considered very important by some. Namely users operating on the tool form would be exposed to a lot of confusion between what they see in the history and what options they are getting in the tool form.
> 
> I'd like to give you some examples but cannot open tool form on this branch at the moment (see above).

Nerdinacan:
@martenson 

Makes sense to me Martin. I found a big problem in the caching layer hunting down one of your feature requests and it led me on a week-long digression into wrapping the caching operators. I'll be working on restoring the client sync for "current history" today though I do still believe that to be a design flaw and not a design feature. :)

I'll look at the tool problems you've outlined too.

client/galaxy/scripts/components/History/model/caching/schema/datasetCollection.schema.js
Nerdinacan: Ooooh, so you confess!

Yeah I think I had a dyslexic stroke when I was wading through a tree structure with names like "element", and "object", haha.

Nerdinacan:
@martenson 

Well, it wasn't easy addressing that tool form issue you found.

Essentially the issue is that all the tool forms are closely-coupled with the existing current history panel model objects from Backbone, which is no big surprise, since that's symptomatic of the reasons we want to replace the Backbone nonsense in the first place.

That being said, what I ended up doing was creating a new Backbone view extended from the existing current history panel view, and overrode the render method to dump the new Vue components in place. Then I monitor Vuex to see when the current history changes, and alert the legacy code via the existing history change methods.

This preserves the existing logic and will keep legacy code working, but it also means that everything the legacy current history panel does, including Ajax calls, will occur even when it is not being viewed because the beta panel is open.

It may be that I can devise a more sophisticated emulation for the existing backbone logic, but I'm not sure it's worth the effort since the REAL goal is to replace all the code that is using that redundant logic in the first place.

**Long story short:** You will see extra ajax calls when the beta panel is open until it is no longer "beta", these are being made by the legacy code.

The issue of actually alerting the server when the user changes the current history was trivial. That was done in about 20 minutes.

jmchilton:
There seems to be some huge regressions here in terms of collections. I'm not saying everything needs to work by the time this is merged but I'd like to at least have a plan in place to address everything.  

1) The job info isn't used to render the collection state - so collections with failed jobs render as green. 

Beta interface:

Screen Shot 2020-01-30 at 12 09 36 PM

dev:

Screen Shot 2020-01-30 at 12 09 20 PM

2) Creating new collections seems to break the GUI:

Screen Shot 2020-01-30 at 11 47 00 AM

3) I haven't seen any evidence of the cool progress bar for collections working (maybe it does though and these other problems are masking it - the job info endpoint does seem to be hit). xref https://github.com/galaxyproject/galaxy/pull/5013/commits/5a9f0dd11eb0ba0abfed29ae9da1d0a5a5a01360 part of https://github.com/galaxyproject/galaxy/pull/5013.

4)

And finally I'm nervous you're hitting the full collection to get the description you're caching. I don't have hard proof you're doing this but I think you are. We shouldn't ever be hitting ``root/api/histories/<historyid>/contents/dataset_collections/<hdcaid>`` and we shouldn't be caching the complete result. That API endpoint can take down the server and the endpoint is not performant even when the server survives. We should assume collections can be unbounded in size and never fetch the whole thing. 

xref https://github.com/galaxyproject/galaxy/pull/5091

The current UI only renders portions of the large collections and warns the user that this is happening. Obviously one could do better - but I think this is a fine start. Progressively pulling more as requested is probably even more challenging now that caching is in place?

Nerdinacan:
@jmchilton 

First of all, THANK YOU. 

This is the kind of feedback I've been waiting for. My goal is to give you guys something you can easily modify, not to make a perfect UI the first time. So I'm definitely less concerned about minor display/css/layout issues and more concerned about the actual functioning of the thing. My goal is to make a system whereby any of you can easily nickle-and-dime the layout to your hearts' content once it's delivering the right functionality.

> Creating new collections seems to break the GUI:

This may be straight bug in the new caching schema. Schema mismatches between updated versions of the client is a thing it has to deal with. It should be handled invisibly, but apparently is not, I'll investigate and try to create some reliable test cases.

It's either that, or the schema of the type of returned collection you created is inconsistent with the schema I have previously defined. I'll see if that's the case too, and I fear that might actually be the problem.

> The job info isn't used to render the collection state - so collections with failed jobs render as green.

As I currently have it, collection "state" comes from "populated_state" from a response from the following endpoint. The idea was to have the HDA/HDCA update_time be accurate so that I can query smaller result sets. This is why I put those triggers into the DB.

```
/api/histories/9752b387803d3e1e/contents?context=contentpoll&v=dev&view=summary&keys=accessible&order=hid-dsc&offset=0&limit=200&q=deleted&qv=False&q=purged&qv=False&q=visible&qv=True&q=update_time-gt&qv=2020-01-31T19:07:55Z
```

```json
[
{
"history_content_type": "dataset_collection",
"update_time": "2020-01-31T19:08:33.553669",
"populated_state_message": null,
"name": "foo",
"type_id": "dataset_collection-33b43b4e7093c91f",
"deleted": false,
"type": "collection",
"history_id": "9752b387803d3e1e",
"tags": [],
"visible": true,
"job_source_id": null,
"job_source_type": null,
"create_time": "2020-01-31T19:08:33.553661",
"hid": 10,
"url": "/api/histories/9752b387803d3e1e/contents/dataset_collections/33b43b4e7093c91f",
"collection_type": "list",
"element_count": 2,
"id": "33b43b4e7093c91f",
**"populated_state": "ok"**
}
]
```

Is this not the right way to evaluate job state? (Clearly it must not be.)

> And finally I'm nervous you're hitting the full collection to get the description you're caching.

I am not only nervous about that, I complained semi-loudly. When I drill into a collection, I do indeed issue a request to the following endpoint from api/history_contents.py because it seemed like the right place to go from reading the api comments.

```
/api/histories/{history_id}/contents/{type}/{id}
Ex: /api/histories/9752b387803d3e1e/contents/dataset_collections/33b43b4e7093c91f
```

I do not think there are any useful filtering parameters that I can issue to this endpoint other than fuzzy_count which seems problematic. (I will carefully read your xref.) 

In a perfect world, we would have an endpoint to query against that had a limit/offset and update_time-gt filter options, maybe even completely without tree data. I think it would be genuinely easier to just have a flat list of collection content and to let the client assemble it into an appropriate tree.

> Progressively pulling more as requested is probably even more challenging now that caching is in place?

That's actually one of the things that took up a lot of time this last year. The need for this kind of highly dynamic scrolling mechanism was one of my reasons for wanting to move to Rxjs for data retrieval management.

I believe I have a solution for history contents, which I was going to extend to history collection contents, but Jeremy and I got to thinking that might be a good thing to add on in a subsequent PR after a basic beta display was generated.

I can revisit that as well. Most of the work is laying around in a separate branch and it involves a dynamic scrolling mechanism that was indeed pretty tricky to put together (as you have surmised). On the up-side, since it's just a bunch of RxJS operators, it's tricky, but it's only about 50 lines of code.

Maybe it's as simple as just generating a new endpoint that only delivers a flat list of datasets belonging to a given parent (history or other collection)? Is that do-able? Given one of those with limit/offset and update_time criteria I could finish it up.

jmchilton:
> Is this not the right way to evaluate job state? (Clearly it must not be.)

``populated_state`` refers only to whether the backbone of the collection is defined. All the elements of the collection may correspond to jobs and those jobs have their own state and the the datasets have their own state as well. The GUI currently uses an aggregation of the job states to guess the collection state. There is no field in the database corresponding to the state of the whole collection - because I worried that trying to coordinate 10,000 different jobs managing a single variable would cause contention and break everything. I could be wrong or there might be some postgres view magic that could save us. I'd want to see evidence that a new solution is performant because the current solution seems to be working for us (even if the GUI logic is a bit convoluted as a result).

> I do not think there are any useful filtering parameters that I can issue to this endpoint other than fuzzy_count which seems problematic. 

I think we need an API to get the collection's full contents but one level at a time - but parameterized to allow pagination within a level. That doesn't exist but shouldn't be too hard to make. fuzzy_count is a hack because we didn't have this API and more importantly we didn't have the GUI framework to display the contents of a collection in that fashion. Clearly fuzzy_count sucks but the GUI is operating much better than before it was added and loading a collection could crash Galaxy.

Nerdinacan:
> The GUI currently uses an aggregation of the job states to guess the collection state. There is no field in the database corresponding to the state of the whole collection - because I worried that trying to coordinate 10,000 different jobs managing a single variable would cause

I'm trying to build an api endpoint that provides job state given a history_id and an update_time right now because there's a big TODO:... placeholder in the location in the api where that functionality belongs. The existing "have to know what you're looking for before you ask for it" plan is not ideal. 

I'm attempting to wade through the aggregate state calculation convolutions to see if I can promptly determine that aggregate value, hopefully made speedier by the assumption that there will be an minimum update_time as a required input filtering parameter so that the required subqueries come back quicker.

You are right, a direct SQL statement or proc would be vastly more efficient. As it almost always is. I'm not there yet. Do we have opinions on SQL views?


> I think we need an API to get the collection's full contents but one level at a time - but parameterized to allow pagination within a level.

That is indeed what I plan to do in the follow-up PR which will address dynamic scrolling through large content lists, including dataset collection contents. The UI doesn't display a whole tree at a time so it's nonsense for the api to deliver it as one.

Nerdinacan:
@jmchilton 

I believe I have addressed your collection job-status updates by adding one more trigger on the job table. When job state changes, the history update_time is updated and then it shows up in the normal polling. A subsequent content query results and it picks up the update. That's actually the way I originally intended it to work.

Do you see any flaws in that strategy?

EDIT: Ahh wait, I just saw the problem. We're still not actually retrieving the "state" for the collection because that data is not available in the content query. I'll look again. Dammit.

martenson:
@Nerdinacan on the 20.05 committers meeting there was a lot of hope that we could get this in soon after the 20.05 release and deploy it on Test. Could you please find some time on de-conflicting and polishing this PR? I haven't tested it in quite some time (months probably) and would be happy to if you call it ready.

Nerdinacan:
> @Nerdinacan on the 20.05 committers meeting there was a lot of hope that we could get this in soon after the 20.05 release and deploy it on Test. Could you please find some time on de-conflicting and polishing this PR? I haven't tested it in quite some time (months probably) and would be happy to if you call it ready.

I'm doing that right now. It took a long time to get the required dependencies merged and one still isn't:

https://github.com/galaxyproject/galaxy/pull/9722

But this should be ready soon.

Nerdinacan:
Performance tuning is getting close. I think it's down to cosmetics and testing.

**Notable changes:** 

- Abandoned Rxdb. I just got tired of its bugs and it was semi-straightforward if not simply to use PouchDB directly. Now live queries are handled by pouchdb-live-find. This also eliminated the need for us to maintain schemas in what should have been a schemaless data store, because that was a (seemingly needless) RxDB requirement, not a PouchDB requirement. Updates to the cached data are much simpler now since it's just a json block with a few indexed fields.
- Abandoned separation between "summary" and "detailed" history content.
- Added a worker to handle caching. Found a great library that exposes worker functions as either promises or observables.The client components can consume these functions almost as if they weren't in the worker at all (with a couple minor caveats). The worker now handles all caching, cache-watching, polling and outgoing ajax calls used to update the cache. There's potential here to expand this feature into other processor-intensive parts of the client. (probably the graph generation that comes next.)
- Switched virtual scrollers. The new scroller's api is not as intuitive but its performance is much much better.

bgruening:
@Nerdinacan is this ready to be tested again? Is it more or less feature frozen? I can spend a few hours and test it if you like.
The test failures look relevant.

Can't wait to see this in action on our test server :)

mvdbeek:
I'm gonna push this to 20.09, let's try to get this in soon after the freeze, either as a whole or as smaller pieces!

Nerdinacan:
I think.... I'm done.

Beer me.

bgruening:
> Beer me.

:100: Relly just beer, send me your address and I will send you some good Whiskey -  we all know you like it!

bgruening:
@Nerdinacan thanks for your continuous work on this. Here is my small feedback after using it on a few hundreds of elements.

* scrolling works nicely in general
  * it seems to be CPU heady under FF, but I guess this is not avoidable
  * if I scroll to fast I end up in a grey area, maybe we can stop scrolling into endless grey area and indicate that data is fetching?
  * at the bottom of the history is a grey area that stays forever
* download button is closing the box
* boxes are not draggable into tools

* select all button does not work

grafik

* and I can not build a collection atm, I get a `TypeError: can't convert null to object`
* Collapse Expanded Elements returns a `TypeError: e.emit is not a function`
* tool reload refreshed the entire page

Thanks @Nerdinacan!

Nerdinacan:
Awesome @bgruening, 

I will look at your finding right now. it's been pretty difficult finding a system that makes scrolling smooth and continuous when the loading is not but I think it's finally there. I'll look at the flaws you found right now.

Also, it may not be relevant for your issues but is there a way you and I can share some test histories so that we're all looking at the same thing?

bgruening:
> Also, it may not be relevant for your issues but is there a way you and I can share some test histories so that we're all looking at the same thing?

Not sure how :( I just created 1000 text files, nearly empty, and uploaded them, copied them a few times from one history into anoter and so on :)
What we can do is share a screen (zoom session?) and I can show you what I see.

Nerdinacan:
Check Gitter! I'm around.

bgruening:
When I normally scroll, not jump with the scrollbar, loading of new elements does not work for me
Sometimes one single element is added
jumping with the scrollbar works, at least in FF
things to reproduce:

* clear cache, not with your button, but with a browser plugin. Your clear cache button does not work or has a different outcome
* scroll dow, not super fast - ~60 elements I see and end and it does not load
* jump with the scrollbar down to 3/4 --> content will load
* jump up to the beginning again

the problem seems to be that the API returns [].

Nerdinacan:
> Returns something. It the last element in my history
> 
> >

Lemme look again, that looks like something I just spent 2 days trying to squash.

bgruening:
* the scrolling is now much smoother and works well for most of the times
* clicking download-button closes the green-dataset-box
* tool-reload button is refreshing the entire page

* When I change a filetype, the rendering of the history seems off, see the following image:

grafik

There are many small things that need to be improved but overall I think it's ready for larger testing.

Nerdinacan:
@bgruening Thanks man. I'm in the same boat.

I could keep whittling at thus thing forever. Right now I'm trying to understand how it's failing existing integration tests when the base UI only changes when you explicitly request the beta side-panel. I suspect that a lot of tests are just broken in dev.

mvdbeek:
> I suspect that a lot of tests are just broken in dev.

They do pass in dev (most of the time). You can download the artifacts of the failing tests from https://github.com/galaxyproject/galaxy/actions/runs/409474827. There's a lot of 404's, maybe something in the routing isn't right ?

Nerdinacan:
> > I suspect that a lot of tests are just broken in dev.
> 
> They do pass in dev (most of the time). You can download the artifacts of the failing tests from https://github.com/galaxyproject/galaxy/actions/runs/409474827. There's a lot of 404's, maybe something in the routing isn't right ?

I just did a careful diff of my branch against div. There's definitely something funky about this branch. It's been around so long, I think I may have hosed a prior merge or something. I'm going to hand-build it tomorrow with only the new relevant changes and remove some of the wierd stuff I found.

Nerdinacan:
Pretty sure this is merge-able at this point. If so, I'd like to put it in dev in beta format and start addressing bugs in smaller PRs if everybody's cool with that plan.

dannon:
Yep, that's the plan.  I'll take another look later today.

dannon:
Alright, I played with this a lot more and read through it all again.  This is a big step forward towards a more performant, reactive history panel and application in general, and it really nails a lot of the important stuff.  There are definitely some issues, rough edges, and general polish needed before we'll bring it out of beta but this is in a great spot to merge now.

As mentioned, it's behind a toggle so naive users won't be impacted by these changes at all until we've had more time to refine things, but getting it in now will get more eyes on it and will allow us to start iterating.

dannon (approved):

bgruening:
🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉

This such a huge step! Thanks a bunch, @Nerdinacan!

PR #8449, opened 2019-08-12
Use outputs/ and configs/ directories for Galaxy jobs.

jmchilton:
More structured job directories aligned with Pulsar with real ramifications in terms of job isolation & security. The tool container shouldn't be expected to mount the job directory read-write - creating an outputs directory like used by Pulsar is a way around this - only the outputs and working directory will need to be mounted rw. This will also allow Pulsar to read these files when doing extended metadata collection remotely - since Pulsar won't default to mounting this directory read-write.


jmchilton:
Merged as part of #8625

PR #3734, opened 2017-03-09
Quota recalculation logic consolidation

dannon:
Refactors quota recalculation, tested logout and admin reset functionality w/ both sqlite and postgres, seems to work well -- having a hard time locating a bug corresponding to reports that it isn't working for users when they log out.  We can keep an eye on that, though.

PR #8452, opened 2019-08-13
Replace in-python quota calc with a SQL query

hexylena:
This replaces N database operations (datasets / 1000), (histories / 1000) with a single SQL query that returns just the number we're interested in.

It's one of the operations that constantly generates logging messages about slow queries, and we force recalculate queries regularly in order to cut down on support emails, so it'd be nice if it was faster.

Does anyone like writing SQLAlchemy? Maybe it's just that my job has become "write plain SQL all day every day" but 90% of the time of this PR was figuring out how to translate the query into SA.


lib/galaxy/model/__init__.py
```diff
@@ -508,19 +508,24 @@ def calculate_disk_usage(self):
...
+        #   joining that on Histories which aren't purged (and owned by us)
+        #   joining on LDDAs where the current dataset is NOT in the LDDA table
+        #   as a CTE
+        cte = db_session \
+            .query(Dataset.id, func.coalesce(Dataset.total_size, func.coalesce(Dataset.file_size, 0)).label("file_size")) \
+            .distinct(Dataset.id) \
+            .join(HistoryDatasetAssociation).filter_by(purged=False) \
+            .join(History).filter_by(user_id=self.id, purged=False) \
+            .outerjoin(LibraryDatasetDatasetAssociation, Dataset.id == LibraryDatasetDatasetAssociation.dataset_id) \
+            .filter(LibraryDatasetDatasetAssociation.id==None) \
```
mvdbeek: To avoid the linter complaining you can either use
```suggestion
            .filter(LibraryDatasetDatasetAssociation.id.is_(None)) \
```
or add `  # noqa: E711 ` and add the missing whitsepace around the operator.

lib/galaxy/model/__init__.py
```diff
@@ -508,19 +508,24 @@ def calculate_disk_usage(self):
...
+            .cte("per_user_datasets")
+
+        q = db_session.query(func.sum(cte.c.file_size)).select_from(cte)
+        return q.one()[0]
```
mvdbeek:
```suggestion
        return q.first()
```
should be fine, right ?

mvdbeek (commented):
Nice

lib/galaxy/model/__init__.py
mvdbeek: Or null() as you did, that's all fine.

hexylena:
Needs a refactoring so I've moved to WIP, will rewrite quickly.

lib/galaxy/model/__init__.py
hexylena: > one()
> Return exactly one result or raise an exception.

here we only expect one, anything else would be something wrong. Unfortunately will still need [0] in both cases to select just the first column

mvdbeek: If you mean you'd be relying on one() to raise an error I don't think that'll work with func.sum, this actually returns `(None)` if you use for instance a non-existent user id.

To stay in line with the previous implementation you could use .first() or 0.

hexylena:
Ok so here's the deal: Using SQLalchemy with a couple of CTEs, SA decides to "optimise" that into a subquery. The performance of this is 10-20x worse for me than CTEs (10s for "our favourite" instead of 0.5s).  It works it's just slow :(

If anyone can coerce SA to generate **this exact query** with the ORM, then, awesome, please fix it :)

Otherwise some numbers that might convince you to just merge my plain sql ;)

User | Datasets | Previous | New
--- | --- | --- | ---
Our Favourite | 2940799 | 29s | 538.730 ms
Bjoern | 48316 | 12s | 180.005 ms


my last python version:

```python
per_user_histories = s.query(History.id) \
    .filter_by(user_id=1, purged=False) \
    .cte("per_user_histories")

per_hist_hdas = s.query(HistoryDatasetAssociation.dataset_id) \
    .distinct(HistoryDatasetAssociation.dataset_id) \
    .filter(HistoryDatasetAssociation.purged == False, HistoryDatasetAssociation.history_id.in_(per_user_histories))\
    .cte("per_hist_hdas")

per_user_datasets = s.query(
        Dataset.id, 
        func.coalesce(Dataset.total_size, func.coalesce(Dataset.file_size)).label("file_size")
    )\
    .outerjoin(LibraryDatasetDatasetAssociation, Dataset.id == LibraryDatasetDatasetAssociation.dataset_id) \
    .filter(LibraryDatasetDatasetAssociation.id == None, Dataset.id.in_(per_hist_hdas))\
    .cte("per_user_datasets")

q = s.query(func.sum(per_user_datasets.c.file_size)).select_from(per_user_datasets)
print(q.all())
print(q.statement)
```

mvdbeek:
Just checking this one out I noticed that `calculate_disk_usage` AFAICT will only be called if not on postgresql because of https://github.com/erasche/galaxy/blob/7830c54a60369e5faf47ee1c55f2de3604e8251e/lib/galaxy/model/__init__.py#L545

On postgres one would instead call `pgcalc` which does a slightly different query:
https://github.com/erasche/galaxy/blob/7830c54a60369e5faf47ee1c55f2de3604e8251e/lib/galaxy/model/util.py#L6

Could those be united ? I also noticed 100% CPU usage and a hang when attempting to create a new user on a clean new sqlite database, but I haven't checked if that's due to the PR or something being broken on dev.

mvdbeek:
> but I haven't checked if that's due to the PR or something being broken on dev.

broken on dev too, trying to figure out what's going on there now.

hexylena:
Ok, sounds good. Thanks for catching that, hadn't seen the others yet because I'm blind.

mvdbeek:
OK, actually the hang is specific to the PR, it appears on sqlite, probably because of
```
galaxy.queue_worker ERROR 2019-08-14 07:20:41,387 [p:39748,w:4,m:0] [Thread-1] Error running control task type: recalculate_user_disk_usage
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1244, in _execute_context
    cursor, statement, parameters, context
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.7/site-packages/sqlalchemy/engine/default.py", line 550, in do_execute
    cursor.execute(statement, parameters)
sqlite3.OperationalError: wrong number of arguments to function coalesce()

  File "lib/galaxy/queue_worker.py", line 349, in process_task
    result = f(self.app, **body['kwargs'])
  File "lib/galaxy/queue_worker.py", line 220, in recalculate_user_disk_usage
    user.calculate_and_set_disk_usage()
  File "lib/galaxy/model/__init__.py", line 548, in calculate_and_set_disk_usage
    new = self.calculate_disk_usage()
  File "lib/galaxy/model/__init__.py", line 533, in calculate_disk_usage
    """, {'user_id': self.id})
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.7/site-packages/sqlalchemy/orm/session.py", line 1269, in execute
    clause, params or {}
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 988, in execute
    return meth(self, multiparams, params)
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.7/site-packages/sqlalchemy/sql/elements.py", line 287, in _execute_on_connection
    return connection._execute_clauseelement(self, multiparams, params)
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1107, in _execute_clauseelement
    distilled_params,
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1248, in _execute_context
    e, statement, parameters, cursor, context
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1466, in _handle_dbapi_exception
    util.raise_from_cause(sqlalchemy_exception, exc_info)
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.7/site-packages/sqlalchemy/util/compat.py", line 399, in raise_from_cause
    reraise(type(exception), exception, tb=exc_tb, cause=cause)
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.7/site-packages/sqlalchemy/util/compat.py", line 153, in reraise
    raise value.with_traceback(tb)
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1244, in _execute_context
    cursor, statement, parameters, context
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.7/site-packages/sqlalchemy/engine/default.py", line 550, in do_execute
    cursor.execute(statement, parameters)
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) wrong number of arguments to function coalesce()
[SQL:
            WITH per_user_histories AS
            (
                SELECT history.id as id
                FROM history
                WHERE history.user_id = ?
                    AND history.purged = false
            ),
            per_hist_hdas AS (
                SELECT DISTINCT history_dataset_association.dataset_id as id
                FROM history_dataset_association
                WHERE history_dataset_association.purged = false
                    AND history_dataset_association.history_id in (SELECT id from per_user_histories)
            )

            SELECT sum(coalesce(dataset.total_size, coalesce(dataset.file_size)))
            FROM dataset
            LEFT OUTER JOIN library_dataset_dataset_association ON dataset.id = library_dataset_dataset_association.dataset_id
            WHERE dataset.id in (SELECT id from per_hist_hdas)
                AND library_dataset_dataset_association.id IS NULL;
        ]
[parameters: (4,)]
(Background on this error at: http://sqlalche.me/e/e3q8)
```

mvdbeek:
```
diff --git a/lib/galaxy/model/__init__.py b/lib/galaxy/model/__init__.py
index 7193e32ca3..6dfc6a3d11 100644
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -525,7 +525,7 @@ class User(Dictifiable, RepresentById):
                     AND history_dataset_association.history_id in (SELECT id from per_user_histories)
             )
 
-            SELECT sum(coalesce(dataset.total_size, coalesce(dataset.file_size)))
+            SELECT sum(coalesce(dataset.total_size, dataset.file_size))
             FROM dataset
             LEFT OUTER JOIN library_dataset_dataset_association ON dataset.id = library_dataset_dataset_association.dataset_id
             WHERE dataset.id in (SELECT id from per_hist_hdas)

```
fixes the exception and the hang and works on sqlite

hexylena:
That's a bug on my end. Second coalesce needs to be 0. Thanks for the fix,
will apply in a bit when I do more testing.

Op wo 14 aug. 2019 13:29 schreef Marius van den Beek <
notifications@github.com>:

hexylena:
looking at https://github.com/galaxyproject/galaxy/pull/3734 it was changed to be in a while loop, do we really want that operation to be unbounded in time? Could we make it do it once and be done? While loops make me nervous

mvdbeek:
Ah, yeah, that's not great, I'd just try a couple of times and give up with a warning message ?
But after your optimization it should complete so quick that I think we can just do it once. It's the current state, I don't think we need to wait for the value to stabilize.

hexylena:
Ok, I've made my changes a bit more aggressive but it results in more code removed :)

- everything moved into `__init__.py`, utils removed.
- the function signature changes, now an update does not return a value. `RETURNING` is not supported in sqlite, and if we remove that, we can use an identical query across DBs (win!)
- we remove the while loops for changing disk usage. Unbounded while makes me nervous + now the query is super fast, so seems less important?

lib/galaxy/model/__init__.py
```diff
@@ -509,43 +508,57 @@ def calculate_disk_usage(self):
...
+        if dryrun:
+            r = sa_session.execute(ctes + sql_calc, {'id': self.id})
+            return r.fetchone()[0]
         else:
-            new = pgcalc(db_session, self.id)
-        if new not in (current, None):
-            self.set_disk_usage(new)
-            db_session.add(self)
-            db_session.flush()
+            r = sa_session.execute(ctes + sql_update, {'id': self.id})
```
mvdbeek, line 558: Do you know if the mapper is smart enough to immediately reflect the new `disk_usage` value ?
I suspect it is not, so we may need to refresh the object with `sa_session.refresh(self)`.

hexylena: I don't, added your suggestion.

mvdbeek:
still wip @erasche  ?

hexylena:
Nope :) test away!

mvdbeek (approved):
Works great with existing and new sqlite and postgres databases. Thanks a lot @erasche!

jmchilton:
Awesome, thanks!

Issue #8183, opened 2019-06-18
GitLab error reporting additions


selten:
I'd like to suggest the following:

*   Change to Session portion of the code to allow the environment variable no_proxy to be set as well
*   Allow assigning a new issue on GitLab immediately to the last committer if it is not the default project (toggleable). Quick note on code to use:
    ```
    project.commits.list()[0].attributes['author_name']

    gitlab.users.list(username='USER')[0].get_id()
    
    project.issues.create({
        'title': 'test issue', 
        'description': 'test issue', 
        'assignee_ids': [1]
    })
    ```

I'll try to implement these myself

PR #8455, opened 2019-08-13
Implement #8183 for further GitLab integration. 

selten:
Allow toggling of the proxy functionality, add automatic assignment to the last person that commited to the repository.

### DIFF
--- a/lib/galaxy/config/sample/error_report.yml.sample
+++ b/lib/galaxy/config/sample/error_report.yml.sample
@@ -81,3 +81,4 @@
 #   gitlab_default_repo_owner: galaxyproject
 #   gitlab_default_repo_name: galaxy
 #   gitlab_default_repo_only: true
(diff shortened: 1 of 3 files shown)


hexylena (commented):
looks good! haven't tested personally, but I guess you're the primary user of this reporter, the code all looks valid

mvdbeek (approved):

mvdbeek:
@galaxybot test this

mvdbeek:
Thanks @selten!

hexylena:
Oh, I guess my 'review' comments never were recorded/submitted, only thing was "why make proxy optional if it can be autodetected/configured by presence of those env vars" but this is a small detail.

selten:
@erasche Because the no_proxy variable isn't being respected seemingly. So therefore there are use cases where turning this off completely is preferential.

PR #8456, opened 2019-08-13
Add file extensions when downloading from data libraries

selten:
Currently if the file was imported from the history and the dataset name didn't specify a file extension in there, it will download a file without a file extension.
In order to make it more convenient for the users, this adds the extension if possible.

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/library_datasets.py
+++ b/lib/galaxy/webapps/galaxy/api/library_datasets.py
@@ -623,6 +623,9 @@ class LibraryDatasetsController(BaseAPIController, UsesVisualizationMixin, Libra
                 zpath = os.path.split(path)[-1]  # comes as base_name/fname
                 outfname, zpathext = os.path.splitext(zpath)
 
+                if zpathext == '':
+                    path = '.'.join([path, ldda.extension]) # Add extension if we don't have one
+
                 if is_composite:
                     # need to add all the components from the extra_files_path to the zip
                     if zpathext == '':
@@ -647,6 +650,8 @@ class LibraryDatasetsController(BaseAPIController, UsesVisualizationMixin, Libra
                         efp, fname = os.path.split(fpath)
                         if fname > '':
                             fname = fname.translate(trantab)
+                        if zpathext == '':
+                            fname = '.'.join([fname, ldda.extension])  # Add extension if we don't have one
                         try:
                             if format == 'zip':
                                 archive.add(fpath, fname)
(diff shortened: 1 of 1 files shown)


lib/galaxy/webapps/galaxy/api/library_datasets.py
```diff
@@ -623,6 +623,9 @@ def traverse(folder):
                 zpath = os.path.split(path)[-1]  # comes as base_name/fname
                 outfname, zpathext = os.path.splitext(zpath)
 
+                if zpathext == '':
+                    path = '.'.join([path, ldda.extension])  # Add extension if we don't have one
```
mvdbeek: This doesn't seem to have any effect, path isn't consumed after this.

lib/galaxy/webapps/galaxy/api/library_datasets.py
```diff
@@ -647,6 +650,8 @@ def traverse(folder):
                         efp, fname = os.path.split(fpath)
                         if fname > '':
                             fname = fname.translate(trantab)
+                        if zpathext == '':
```
mvdbeek: I'd do this in all cases, otherwise any library item with a `.` in the name will be without the proper extension. I'd also simplify this to
`fname = "{fname}.{extension}".format(fname=fname, extension=ldda.extension)`
My inclination would be to also change that on line 632. This had been present since the initial commit, but I don't really see the reason for that. Do you have any objections to that @martenson ?

lib/galaxy/webapps/galaxy/api/library_datasets.py
```diff
@@ -702,7 +707,11 @@ def traverse(folder):
                 fStat = os.stat(dataset.file_name)
                 trans.response.set_content_type(ldda.get_mime())
                 trans.response.headers['Content-Length'] = int(fStat.st_size)
-                fname = ldda.name
+                zname, zext = os.path.splitext(ldda.name)
```
mvdbeek: I'd replace lines 710-714 with `fname = "{fname}.{extension}".format(fname=ldda.name, extension=ldda.extension)`

mvdbeek (commented):
I like it, I'd just be more stringent and do it all the time.

lib/galaxy/webapps/galaxy/api/library_datasets.py
selten: What about lines 671-674?

mvdbeek: Ah right, but then as in my previous comments I'd contract this to `path = "{path}.{extension}".format(path=path, extension=ldda.extension)` without the if statement.

selten:
I updated it with your suggestions.

jmchilton (approved):

martenson:
@galaxybot test this

martenson:
@selten I invited you to a galaxyproject organization so @galaxybot tests your PRs and listens to you, please make sure your password is strong and consider enabling 2FA for github if you accept

bgruening:
Thanks @selten!

PR #8462, opened 2019-08-14
Refactor reading config properties from a file

jdavcs:
Required for https://github.com/galaxyproject/galaxy/pull/8230 (closing; will do a new PR)
Addresses https://github.com/galaxyproject/galaxy/issues/8072

Factor out reading a config file from `load_app_properties()` into a separate function. The original function read and combined properties from kwargs, a config file, and env vars. Reading config properties from a file *without* overriding them with env vars was impossible. As a result of this refactoring, the logic for reading a config file still exists in one place only, but now it can be used by other callers. One such use case is #8072 where config options are reloaded at runtime when the config file is modified (and, thus, needs to be read).   

Notable changes:
1. Previous logic: init properties with kwds, then update from config file, then update env vars. Issues: values from kwds were overridden with env vars for ini files and discarded for yaml files.
New logic: init properties, then update from config file, then env vars, then kwds - for both ini and yaml files (see my comment on gitter 8/14/19 11:46)
2. Change kwds default arg from {} to None (because mutable default arg, and because @mvdbeek is handing out bonus points https://github.com/galaxyproject/galaxy/pull/8314#discussion_r301501595 )
3. Simplified logic, removed redundant code, added error handling, added unit tests.

Ping @jmchilton and @natefoo (who previously edited this code).


lib/galaxy/util/properties.py
```diff
@@ -105,9 +92,40 @@ def load_app_properties(
...
 
+    # update from kwds
+    if kwds:
+        properties.update(kwds)
+
     return properties
 
 
+def read_properties_from_file(config_file, config_section=None):
+    properties = {}
```
jmchilton, line 103: Add ``__file__`` to the properties?

jmchilton (approved):
Awesome - thanks!

lib/galaxy/util/properties.py
jdavcs: It is already added: if file is ini, it will be added here:
https://github.com/galaxyproject/galaxy/blob/ec74687d12f0ab6ddba6e6da66df2e4ab3547d3c/lib/galaxy/util/properties.py#L130
if it's yaml, then here:
https://github.com/galaxyproject/galaxy/blob/ec74687d12f0ab6ddba6e6da66df2e4ab3547d3c/lib/galaxy/util/properties.py#L107
And if there's no file, then here: 
https://github.com/galaxyproject/galaxy/blob/ec74687d12f0ab6ddba6e6da66df2e4ab3547d3c/lib/galaxy/util/properties.py#L82

jdavcs: links in comment were wrong, corrected now.

jdavcs: it's also tested in the unit tests: all default properties are added in both cases (ini and yaml)

jmchilton: Ahh - I see it now. Cool!

PR #8466, opened 2019-08-15
Fix absolutely positioned elements and children from displaying off-screen.

dannon:
I came across this in dev, seems to have been broken in the style overhaul (specifically 308b376a77a67667ff86a456cc05daff0e72e5f2), but I'm wondering if this isn't something unique to my situation given nobody else noticed it?

Without this, my page layout looks like:

image

Note masthead off screen and `#everything` shifted up.


jxtx:
That's a weird one (never seen it) but this is cleaner anyway.

jxtx (approved):

PR #8467, opened 2019-08-15
Improve cleanup of local job runner.

jmchilton:
https://jenkins.galaxyproject.org/job/docker-api/16801/testReport/junit/test.api.test_datasets/DatasetsApiTestCase/test_show/

### DIFF
--- a/lib/galaxy/jobs/runners/local.py
+++ b/lib/galaxy/jobs/runners/local.py
@@ -214,10 +214,16 @@ class LocalJobRunner(BaseJobRunner):
             return False
 
     def _terminate(self, proc):
-        os.killpg(proc.pid, 15)
-        sleep(1)
-        if proc.poll() is None:
-            os.killpg(proc.pid, 9)
+        process_exists = True
+        try:
+            os.killpg(proc.pid, 15)
+        except OSError as e:
+            if e.errno == errno.ESRCH:
+                process_exists = False
+        if process_exists:
+            sleep(1)
+            if proc.poll() is None:
+                os.killpg(proc.pid, 9)
         return proc.wait()  # reap
 
     def __poll_if_needed(self, proc, job_wrapper, job_id):


lib/galaxy/jobs/runners/local.py
```diff
@@ -214,10 +214,16 @@ def _check_pid(self, pid):
             return False
 
     def _terminate(self, proc):
-        os.killpg(proc.pid, 15)
-        sleep(1)
-        if proc.poll() is None:
-            os.killpg(proc.pid, 9)
+        process_exists = True
+        try:
+            os.killpg(proc.pid, 15)
```
nsoranzo, line 219:
```suggestion
            proc.terminate()
```

lib/galaxy/jobs/runners/local.py
```diff
@@ -214,10 +214,16 @@ def _check_pid(self, pid):
...
+        except OSError as e:
+            if e.errno == errno.ESRCH:
+                process_exists = False
+        if process_exists:
+            sleep(1)
+            if proc.poll() is None:
+                os.killpg(proc.pid, 9)
```
nsoranzo:
```suggestion
                proc.kill()
```

nsoranzo (commented):
There is similar code inside the `stop_job()` method, maybe that can also call `_terminate()` ?

Also, there's a `kill_pid` function defined in `lib/galaxy/jobs/runners/util/kill.py` , which is used in `lib/galaxy/jobs/runners/util/cli/shell/local.py` .

Finally `Popen` used in `lib/galaxy/jobs/runners/util/cli/shell/local.py` should use a `preexec_fn`, I think.

lib/galaxy/jobs/runners/local.py
natefoo: `os.killpg()` != `Popen.terminate()` right? From [what I can tell](https://docs.python.org/2/library/subprocess.html#subprocess.Popen.terminate), `Popen.terminate()` is not going to signal all pids in the process group?

nsoranzo: But this is invoking `os.killpg()` on a process id ( https://docs.python.org/2/library/subprocess.html#subprocess.Popen.pid ), not a process group id, right?

natefoo: The call to `subprocess.Popen()` uses `os.setpgrp()` to set the process's pgid to its pid (i.e. ensures that it is the root of its own process group instead of Galaxy's process group), so the pgid to kill is the pid.

nsoranzo: Ah, TIL! I'll add a comment to document that.

nsoranzo: @natefoo I did some testing on this today and I noticed that `os.killpg()` was not killing orphan processes left by `proc`. This can be tested with `'python  forking_orphan_processes.py'`  as `command_line` , where ` forking_orphan_processes.py` is https://gist.github.com/nsoranzo/d68793e9fb6269f997eb37d2747c262e

This seems to be fixed by using `preexec_fn=os.setsid` (instead of `preexec_fn=os.setpgrp`) when calling `subprocess.Popen()` . What do you think?

lib/galaxy/jobs/runners/local.py
```diff
@@ -214,10 +214,16 @@ def _check_pid(self, pid):
...
-            os.killpg(proc.pid, 9)
+        process_exists = True
+        try:
+            os.killpg(proc.pid, 15)
+        except OSError as e:
+            if e.errno == errno.ESRCH:
+                process_exists = False
+        if process_exists:
+            sleep(1)
+            if proc.poll() is None:
```
nsoranzo, line 225: Given the discussion above, this is not consistent with working on a process group, given that `proc.poll()` only checks if the "main" child process has terminated.

lib/galaxy/jobs/runners/local.py
natefoo: @nsoranzo It's working for me if I add a sleep to the script so I can kill the job (via Galaxy by deleting the output). I think what's missing here is that we should `os.killpg()` the process group after a normal exit to attempt to clean up orphans?

lib/galaxy/jobs/runners/local.py
natefoo: This cold use `os.killpg(proc.pid, 0)` I believe?

nsoranzo: > This cold use `os.killpg(proc.pid, 0)` I believe?

How? It doesn't seem to have any effect. I would just do something like:
```python
            try:
                os.killpg(proc.pid, signal.SIGTERM)
            except OSError as e:
                if e.errno != errno.ESRCH:
                    raise
```

lib/galaxy/jobs/runners/local.py
nsoranzo: Yeah, I can't reproduce it myself, I probably mixed things up with my unsuccessful attempt to use `os.waitpid(-proc.pid, 0)` to wait for the orphan processes. I like the idea of the `os.killpg()` call after `proc.wait()`, I'll work on a patch.

lib/galaxy/jobs/runners/local.py
natefoo: Will it raise `ESRCH` if the original process is dead but other processes in the group are still running?

nsoranzo: > Will it raise `ESRCH` if the original process is dead but other processes in the group are still running?

No, it doesn't.

PR #8470, opened 2019-08-16
Fix #975

### LINKED ISSUE #975: Confusion - 60 vs 10 day default in cleanup scripts
Not sure this is really an issue, but it caught me out (without harmful consequence) once.

In `cleanup_datasets.py` the default is `-d 60` for things over 60 days old. 60 days is used repeatedly in examples on the wiki:

https://wiki.galaxyproject.org/Admin/Config/Performance/Purge%20Histories%20and%20Datasets

However... the `purge_xxxxxx.sh` scripts are all using `-d 10`. After seeing 60 days everywhere, that's not what I was expecting. I should know to check the script first, but it feels a little odd.

hexylena:
switch to the safer one, xref #975, I can imagine this being confusing if you don't read the script before running it.

### DIFF
--- a/scripts/cleanup_datasets/purge_datasets.sh
+++ b/scripts/cleanup_datasets/purge_datasets.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 
 cd "$(dirname "$0")"/../..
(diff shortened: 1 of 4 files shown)


jmchilton:
Awesome, thanks so much!

PR #8473, opened 2019-08-16
Fix MySQL database creation and migrations

### LINKED ISSUE #6401: Migrations broken on mysql
This is with a fresh mysql 5.7 installation and database:

https://gist.github.com/mvdbeek/ef980138db903a9a0ad617f2ac95b025

Some of them are harmless and referencing keys that we eliminated later on, but a couple newer ones seem problematic, like

~~https://gist.github.com/mvdbeek/ef980138db903a9a0ad617f2ac95b025#file-mysql_migration-txt-L536~~ sequencing request table, eliminated long ago

~~https://gist.github.com/mvdbeek/ef980138db903a9a0ad617f2ac95b025#file-mysql_migration-txt-L872~~ eliminated in migration 91

https://gist.github.com/mvdbeek/ef980138db903a9a0ad617f2ac95b025#file-mysql_migration-txt-L1241 -- probably breaks data manager job association table on mysql, solution is to change TEXT type to `TrimmedString(255)` for `data_manager_id` in DataManagerJobAssociation. Which we also do for the tool_id, so this feels like the right thing to do

https://gist.github.com/mvdbeek/ef980138db903a9a0ad617f2ac95b025#file-mysql_migration-txt-L1391 -- needs to be 64 chars long max, so `workflow_invocation_step_output_dataset_collection_association_ibfk_1` is too long.

nsoranzo:
Fix https://github.com/galaxyproject/galaxy/issues/6401 and https://github.com/galaxyproject/galaxy/issues/8400 .

Also:
- fix `create_db.sh`
- Enable on-screen logging for `create_db.py` and `manage_db.py`
- Another huge round of migration refactoring
- Handle better an SQLAlchemy Migrate issue when adding a column with both a `ForeignKey` and an index in SQLite.
- Remove `index=True` from some `Column()` definitions in `lib/galaxy/model/mapping.py` when the index was not part of any migration.

In total, this removes over 500 lines of code.

To test:
- migrations on SQLite:
  `$ rm -f database/universe.sqlite && sqlite3 database/universe.sqlite 'VACUUM;' && ./create_db.sh`
- migrations on PostgreSQL: create an empty 'galaxy' database, then:
  `$ GALAXY_CONFIG_DATABASE_CONNECTION=postgresql://@/galaxy?host=/var/run/postgresql ./create_db.sh`
- migrations on MySQL: create an empty 'galaxy' database, then:
  `$ GALAXY_CONFIG_DATABASE_CONNECTION='mysql://DB_USER:DB_PASS@localhost/galaxy' ./create_db.sh`
- database creation from scratch on SQLite:
  `$ rm -f database/universe.sqlite && ./create_db.sh`
- database creation from scratch on PostgreSQL (the PostgreSQL role needs to be able to create new databases): remove the 'galaxy' database, then:
  `$ GALAXY_CONFIG_DATABASE_CONNECTION=postgresql://@/galaxy?host=/var/run/postgresql ./create_db.sh`
- migrations on MySQL: delete the 'galaxy' database, then:
  `$ GALAXY_CONFIG_DATABASE_CONNECTION='mysql://DB_USER:DB_PASS@localhost/galaxy' ./create_db.sh`

### DIFF
--- a/lib/galaxy/model/mapping.py
+++ b/lib/galaxy/model/mapping.py
@@ -1161,8 +1161,7 @@ model.FormDefinition.table = Table(
     Column("update_time", DateTime, default=now, onupdate=now),
     Column("name", TrimmedString(255), nullable=False),
     Column("desc", TEXT),
-    Column("form_definition_current_id", Integer,
-        ForeignKey("form_definition_current.id", name='for_def_form_def_current_id_fk', use_alter=True), index=True),
+    Column("form_definition_current_id", Integer, ForeignKey("form_definition_current.id", use_alter=True), index=True, nullable=False),
     Column("fields", JSONType()),
     Column("type", TrimmedString(255), index=True),
     Column("layout", JSONType()))
--- a/lib/galaxy/model/mapping.py
+++ b/lib/galaxy/model/mapping.py
@@ -15,6 +15,7 @@ from sqlalchemy import (
     false,
     ForeignKey,
     func,
+    Index,
     Integer,
     MetaData,
     not_,
@@ -198,8 +199,10 @@ model.History.table = Table(
     Column("importing", Boolean, index=True, default=False),
     Column("genome_build", TrimmedString(40)),
     Column("importable", Boolean, default=False),
-    Column("slug", TEXT, index=True),
-    Column("published", Boolean, index=True, default=False))
+    Column("slug", TEXT),
+    Column("published", Boolean, index=True, default=False),
+    Index('ix_history_slug', 'slug', mysql_length=200),
+)
 
 model.HistoryUserShareAssociation.table = Table(
     "history_user_share_association", metadata,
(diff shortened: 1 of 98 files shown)


jmchilton:
Wow, amazing! Thanks so much @nsoranzo !

jmchilton (approved):

martenson:
Does this PR mean we should un-deprecate mysql?

mvdbeek:
We should also have a on-demand test setup for mysql (`galaxybot test mysql`) if we undeprecate IMO. We can run those prior to release or when we suspect issues. That the migrations work is no guarantee that tests or regular operation works, which we should probably aim for if we list mysql as supported.

jmchilton:
We could run the Python 3 API tests in MySQL and the Python 2 tests in Postgres and just add one new test to our regular lineup. I don't know though.. we can fix this and say it is awesome and still say MySQL is much less tested and supported

martenson:
We have the metal to run mysql tests on every PR. At this time it would not slow things down I believe.

mvdbeek:
Are we opposed to just triggering mysql tests as needed ? I'm not sure we really have the resources to quadruple our tests (* 2 for python versions, * 2 for databases), plus 4 times the amount of infrequent but frequent enough to be annoying false positive test failures.

martenson:
@mvdbeek Not opposed, but then it gets tested only when a very small subset of people, that tend to remember, remember. We do not test py3 without a trigger atm and I am ok with not running mysql_py3 on PR in adition to postgres_py3. So we are either adding 100% more load or 50%.

mvdbeek:
But we break py3 far more often, that should be priority if we want to add more permanent tests.
I'm basically arguing for running these tests as part of the release process. Also isn't this 100% more load ?

martenson:
@mvdbeek Yes, let's add postgres_py3 to permanent suite. And yeah, I made a correction to my %.

nsoranzo:
> Does this PR mean we should un-deprecate mysql?

I would swap `deprecated` to `unsupported` in doc/source/admin/production.md , if that sounds reasonable.

I think if we add API tests on Python3+MySQL for each PR that would be already a good coverage (for an unsupported feature).

PR #8474, opened 2019-08-17
Node is called nodejs in conda land.

bgruening:
Our official conda version should be available in conda-forge otherwise the --no-create-venv feature does not work.


nsoranzo (approved):
Are you running `./run.sh` or `./scripts/common_startup.sh` with `--skip-venv` ? Otherwise this shouldn't be needed, since Galaxy by default creates a virtualenv also for Conda installations, and installs `node` in the venv with `nodeenv`.

bgruening:
I was running `./run.sh --no-create-venv`. Thanks for looking into it.

dannon (approved):
New node version works fine; using `nodejs` instead of node in this specific case seems fine.

PR #8475, opened 2019-08-17
Enable dynamic reloading of select config options

jdavcs:
This replaces #8230 and addresses #8072 . 

Summary:
1. Add a watch_config option to galaxy.yml. Turning it on enables dynamic reloading of select config options at runtime
2. The list of options that are safe to reload is located in `galaxy/config/__init__.py`
3. Watching config file for changes implemented in `galaxy/config_watchers.py` (also see `galaxy/util/watcher.py`)
4. Reloading config options implemented in `galaxy/config/__init__.py` (made possible by #8462)

Implementation notes:
1. Addressed issues discussed in #8230:   
  - Remove option to select subset of reloadable properties for reload (it's all or nothing)
  - Option loading happens in one place (config file is not re-found/re-read)
2. Why each reloadable property has its own `_set_[foo]` method.
Previously, each of the now-reloadable properties were set in a simple statement of the form: `self.foo = kwargs.get('foo', default)`. Reloading a property requires different logic: `self.foo = [value from modified file]`. Furthermore, at least one property requires additional setting logic (admin_users >> admin_users_list); such logic may expand for these or future reloadable properties. Thus, setting the value of a reloadable property in two different places would lead to code duplication.  (My solution may not be optimal, so, of course, I'm open to suggestions)

   Other considerations: I don't think I can easily group all properties in one `(re)load_all` method: properties cannot be reloaded in batch (only those that were modified should be reloaded). Also, that would require placing different reloading logic (e.g., welcome_url vs. admin_users) into one method.

   See comment above RELOADABLE_CONFIG_OPTIONS in `galaxy/config/__init__.py` on how to add new reloadable properties.

3. If a reloadable property is deleted from the config file at runtime, it won't be modified. We don't know whether it was initially loaded from this config file (and so has been indeed deleted), or was set to a default, loaded from an env var, etc. - so we don't know the correct value to reset it to (and whether it needs resetting). Therefore, properties will not be deleted on reload (although they can be set to None, of course).

4. Unit tests are included. 
I've moved `test/unit/config/` under `test/unit/webapps/config_manage/`, and created a new `test/unit/config/` for all config-related tests. I believe structurally this makes more sense.

5. Minor thing: `tool_stub_boost` property has been added to `config/__init__.py` (it seems to have been left out as an oversight).


lib/galaxy/config/__init__.py
```diff
@@ -800,6 +823,64 @@ def __init__(self, **kwargs):
                 'filters': ['stack']
             }
 
+    def _set_reloadable_properties(self, kwargs):
+        for key, default_value in RELOADABLE_CONFIG_OPTIONS.items():
+            method = self._get_property_set_method(key)
+            value = kwargs.get(key) or default_value
+            method(value)
```
mvdbeek: You could also do `setattr(self, key, value)` and skip the boilerplate below.
The little bit of processing that `_set_admin_users ` does could be replaced by a property + setter.

lib/galaxy/util/watcher.py
```diff
@@ -154,6 +154,18 @@ def _handle(self, event):
...
 
 
+class SimpleFileModifiedEventHandler(FileSystemEventHandler):
+    """ Minimalist event handler; when all we need is callback() on file modified """
+    def __init__(self, watcher):
+        self.watcher = watcher
+
+    def on_modified(self, event):
+        path = os.path.abspath(event.src_path)
+        callback = self.watcher.file_callbacks.get(path, None)
```
mvdbeek:
```suggestion
        callback = self.watcher.file_callbacks.get(path)
```
since None is the default. That's just for consistency.

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -370,6 +370,16 @@ mapping:
           found, rules are automatically reloaded. Takes the same values as the
           'watch_tools' option.
 
+      watch_config:
```
mvdbeek: Can we find something more specific ? `watch_core_config` maybe ?

doc/source/admin/galaxy_options.rst
```diff
@@ -477,6 +477,20 @@
...
 
+~~~~~~~~~~~~~~~~
+``watch_config``
+~~~~~~~~~~~~~~~~
+
+:Description:
+    Set to True to enable monitoring of a subset of options in the
+    core configuration file (See RELOADABLE_CONFIG_OPTIONS in
+    lib/galaxy/config/__init__.py).  If changes are found, modified
+    options are automatically reloaded. Takes the same  values as the
```
mvdbeek:
```suggestion
    options are automatically reloaded. Takes the same values as the
```

doc/source/admin/galaxy_options.rst
```diff
@@ -477,6 +477,20 @@
...
 
 
+~~~~~~~~~~~~~~~~
+``watch_config``
+~~~~~~~~~~~~~~~~
+
+:Description:
+    Set to True to enable monitoring of a subset of options in the
+    core configuration file (See RELOADABLE_CONFIG_OPTIONS in
+    lib/galaxy/config/__init__.py).  If changes are found, modified
```
mvdbeek:
```suggestion
    lib/galaxy/config/__init__.py). If changes are found, modified
```

doc/source/admin/galaxy_options.rst
```diff
@@ -477,6 +477,20 @@
 :Type: str
 
 
+~~~~~~~~~~~~~~~~
+``watch_config``
+~~~~~~~~~~~~~~~~
+
+:Description:
+    Set to True to enable monitoring of a subset of options in the
```
mvdbeek: `Monitor a subset of options in the` ?
I think it's better to avoid uppercase `True` given that we expect lowercase `true` as value. Phrasing it this way is also more in line with other boolean options in this file.

mvdbeek (commented):
Very nice! When you have some time, could also take a look at https://github.com/galaxyproject/galaxy/pull/8078 ?

doc/source/admin/galaxy_options.rst
mvdbeek:
```suggestion
    lib/galaxy/config/__init__.py). If changes are found, modified
```

lib/galaxy/config/__init__.py
jdavcs: I didn't think of it! It does get rid of a lot of boilerplate. You mean something like this?
```python
def _set_reloadable_properties(self, kwargs):
    for key, default_value in RELOADABLE_CONFIG_OPTIONS.items():
        value = kwargs.get(key) or default_value
        setattr(self, key, value)

def update_reloadable_property(self, key, value):
    if key in RELOADABLE_CONFIG_OPTIONS:  # check again because this is critical
        setattr(self, key, value)

@property
def admin_users(self):
    return self._admin_users

@admin_users.setter
def admin_users(self, value):
    self.admin_users = value
    self.admin_users_list = [u.strip() for u in self.admin_users.split(',') if u]
```

jdavcs:
> Very nice! When you have some time, could also take a look at #8078 ?

Of course! Will do tomorrow.

lib/galaxy/util/watcher.py
jdavcs: Yes. Also here: https://github.com/galaxyproject/galaxy/blob/d02d627aa7e1e8e66923febb15db9d18579bb32c/lib/galaxy/util/watcher.py#L130

lib/galaxy/config/__init__.py
mvdbeek: Yep, though the setter line should be
`self._admin_users = value`

doc/source/admin/galaxy_options.rst
jdavcs: Fixed in next commit.

doc/source/admin/galaxy_options.rst
jdavcs: Fixed in next commit.

doc/source/admin/galaxy_options.rst
jdavcs: I agree. Besides, that suggest a Python value, which is not the case here. Fixed in next commit.

doc/source/admin/galaxy_options.rst
jdavcs: Duplicate - see comment above.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Changed to `watch_core_config` .

lib/galaxy/config/__init__.py
jdavcs: Of course! whoops!
Will push commit in a few minutes. Thanks for the thorough review!

mvdbeek:
> 1. Why each reloadable property has its own `_set_[foo]` method.
>    Previously, each of the now-reloadable properties were set in a simple statement of the form: `self.foo = kwargs.get('foo', default)`. Reloading a property requires different logic: `self.foo = [value from modified file]`. Furthermore, at least one property requires additional setting logic (admin_users >> admin_users_list); such logic may expand for these or future reloadable properties. Thus, setting the value of a reloadable property in two different places would lead to code duplication.  (My solution may not be optimal, so, of course, I'm open to suggestions)
>    Other considerations: I don't think I can easily group all properties in one `(re)load_all` method: properties cannot be reloaded in batch (only those that were modified should be reloaded). Also, that would require placing different reloading logic (e.g., welcome_url vs. admin_users) into one method.
>    See comment above RELOADABLE_CONFIG_OPTIONS in `galaxy/config/__init__.py` on how to add new reloadable properties.

I guess that's addressed by using property getter/setters only for the values that need some processing (`admin_users` is the only one currently).

doc/source/admin/galaxy_options.rst
```diff
@@ -477,6 +477,20 @@
...
+~~~~~~~~~~~~~~~~~~~~~
+
+:Description:
+    Set to true to enable monitoring of a subset of options in the
+    core configuration file (See RELOADABLE_CONFIG_OPTIONS in
+    lib/galaxy/config/__init__.py). If changes are found, modified
+    options are automatically reloaded. Takes the same values as the
+    'watch_tools' option.
+:Default: ``false``
+:Type: str
```
mvdbeek, line 515:
```suggestion
:Type: bool
```

doc/source/admin/galaxy_options.rst
```diff
@@ -477,6 +477,20 @@
 :Type: str
 
 
+~~~~~~~~~~~~~~~~~~~~~
+``watch_core_config``
+~~~~~~~~~~~~~~~~~~~~~
+
+:Description:
+    Set to true to enable monitoring of a subset of options in the
```
mvdbeek: Would you be opposed to `Monitor a subset of options in the ...` ?

lib/galaxy/util/watcher.py
```diff
@@ -154,6 +154,18 @@ def _handle(self, event):
                 callback(path=path)
 
 
+class SimpleFileModifiedEventHandler(FileSystemEventHandler):
```
mvdbeek: This handler doesn't work for moved files, like when a vim buffer is moved to the destination path. Is there any reason the default handler doesn't work for you ?
It works just fine for me.

lib/galaxy/config_watchers.py
```diff
@@ -37,6 +39,12 @@ def __init__(self, app, start_thread=True):
             self.job_rule_watcher = get_watcher(app.config, 'watch_job_rules', monitor_what_str='job rules')
         else:
             self.job_rule_watcher = get_watcher(app.config, '__invalid__')
+        self.core_config_watcher = get_watcher(
+            app.config,
+            'watch_core_config',
+            monitor_what_str='core config file',
+            event_handler_class=SimpleFileModifiedEventHandler
```
mvdbeek:
```suggestion
```
As mentioned below, the handler doesn't work for moving files into place, but the default EventHandler works fine.
Plus it makes sure that there really are changes that require a reload by checking the md5 of the file.

lib/galaxy/config_watchers.py
```diff
@@ -14,6 +15,7 @@
 )
 from galaxy.util.watcher import (
     get_watcher,
+    SimpleFileModifiedEventHandler,
```
mvdbeek:
```suggestion
```

jdavcs:
> > 1. Why each reloadable property has its own `_set_[foo]` method...
> >    ... See comment above RELOADABLE_CONFIG_OPTIONS in `galaxy/config/__init__.py` on how to add new reloadable properties.
> 
> I guess that's addressed by using property getter/setters only for the values that need some processing (`admin_users` is the only one currently).

Yes - so all the '_set...' methods are gone. The comment in `__init__.py` on how to add new options has been edited too.

doc/source/admin/galaxy_options.rst
jdavcs: Oh, I see now. 'watch_tools' accepts more than true/false, and we use the same range of values.

doc/source/admin/galaxy_options.rst
jdavcs: Not at all!

jdavcs: Also changing the other 4 occurrences of the same wording (all in config_schema.yml)

jmchilton:
@ic4f Will you cherry-pick this and clean it up - the unit tests are passing but I'm not sure everything still works?

https://github.com/galaxyproject/galaxy/commit/c5356b298c76c0c65cb288a84dc7caa23a9815bd

I've switch the reloadable properties constant into something read from the schema - that way we don't have to duplicate all the defaults again and we can use the new "reloadable" property when rendering RST in the future. This is more declarative also I think. I promised Dannon years ago when the schemas were merged that we'd work toward de-duplicating default values - this is a step in that direction.

I think since config/ is in the app package, config_manage.py needs to be re-factored out of webapps and into ``galaxy/`` top-level or the new ``galaxy/config`` to get the imports working properly between packages.

jdavcs:
@jmchilton - of course! Working on it..

lib/galaxy/util/watcher.py
jdavcs: Changed to default handler. There was no big reason for using the other one - seemed like that was reducing complexity at the time. You are right about moved files and md5 hash.

lib/galaxy/config_watchers.py
jdavcs: Fixed.

lib/galaxy/config_watchers.py
jdavcs: Fixed.

lib/galaxy/config_watchers.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, app, start_thread=True):
             app.config,
             'watch_core_config',
             monitor_what_str='core config file',
-            event_handler_class=SimpleFileModifiedEventHandler
+            event_handler_class=EventHandler
```
mvdbeek: this is the default, the import and the specification is not needed.

jdavcs: Done.

mvdbeek:
You'll also need to rebase against dev since https://github.com/galaxyproject/galaxy/pull/8078 got merged today

jdavcs:
@mvdbeek I know - I'm merging config_watchers at the moment! Sorry I didn't get to look over your PR before it was merged!

lib/galaxy/util/watcher.py
```diff
@@ -143,7 +143,7 @@ def _handle(self, event):
         # look at dest if it exists else use src.
         path = getattr(event, 'dest_path', None) or event.src_path
         path = os.path.abspath(path)
-        callback = self.watcher.file_callbacks.get(path)
+        callback = self.watcher.file_callbacks.get(path, None)
```
jdavcs: Well, that was stupid. Will change back with next commit.

jdavcs:
This seems to be ready (unless api test fails). This all should be improved as part of #8493, of course (@jmchilton : I'm thinking the same for moving config_manage under config/)
..and it fails, for no apparent reason AFAICS. Restarting... And it passed!
(Hmmm.. Previous build (16891) for api test succeeded. I didn't restart it, so I don't know why there's a failing 16892)

jmchilton:
We worked so hard to get the galaxy.webapps imports out of galaxy-app. I'll merge as is - but I really want that fixed. After I merge can you open a clean PR that just moves that and ping me - I'll try to see what is causing the failure.

jdavcs:
@jmchilton - doing it now. Will ping.

PR #8477, opened 2019-08-19
Fix startup checks when using a Conda env

nsoranzo:
Unless `--skip-venv` is specified, Galaxy creates/activates a virtualenv also when using a Conda environment.


mvdbeek (approved):
Thanks @nsoranzo!

PR #8358, opened 2019-07-19
[WIP] Build and install wheel for packages when testing them

nsoranzo:
With `pip install -e` the installed packages were just links to the corresponding source code.

This will highlight a number of issues with some packages (help in fixing them welcome!).

PR #8480, opened 2019-08-19
Properly test packages, add galaxy-test-tools and galaxy-model-tools package

mvdbeek:
Builds on https://github.com/galaxyproject/galaxy/pull/8358


packages/app/setup.py
```diff
@@ -55,7 +55,7 @@ def get_var(var_name):
     'galaxy.tools.error_reports',
     'galaxy.tools.expressions',
     'galaxy.tools.filters',
-    'galaxy.tools.imp_exp',
+    'galaxy.model_tools.imp_exp',
```
nsoranzo:
```suggestion

```

and add `galaxy-model-tools` to `packages/app/requirements.txt`

jmchilton:
Fascinating... wasn't a direction I was expecting but looks good.

jxtx:
Cool, but why "model_tools". The name is not immediately intuitive to me.

mvdbeek:
No good reason, we've been calling them that way in the codebase, they're also known as Database Operation tools. We could also call them "special tools", "framework tools" or something else altogether. Do you have a suggestion ?

jmchilton:
This will break import/export jobs that occur during an upgrade I think. We broke those last cycle and no one noticed. The little we tried to do to save them just turned out to be unneeded complexity and wasted effort IMO - so I'd just like to do that one more time and keep this as is without worrying about these potential jobs.

mvdbeek:
I think we may be able to rescue those jobs if we provide a fallback import at the old location -- that should be possible, right ? `In tools/__init__.py`
```
from galaxy.model_tools import imp_exp  # noqa: whatever the right number is  # fallback for old imp/exp jobs
```

jmchilton:
I mean - we could definitely fix them. I don't think it is worth adding the lines of code to do that though. It will likely not really save a job across all Galaxies everywhere IMO.

jxtx:
Framework tools makes more sense to me.
-- 
jt (mobile)

jmchilton:
😿 this was cool stuff - hopefully we find some time to keep it going at some point.

PR #8481, opened 2019-08-19
add tests for filters

bernt-matthias:
Just had the feeling that filters (with all their options) deserve some tests. 

Will be a a side project .. lets see how far I get.


bernt-matthias:
Hey this is fun. Really start to love filters https://github.com/galaxyproject/galaxy/blob/e57e52b846f22b314584ea86877afe38816e23e7/test/functional/tools/filter_multiple_splitter.xml#L8

PR #8483, opened 2019-08-20
[19.05] Fix error reporting when input dataset is NoneType

### LINKED ISSUE #8448: Uncaught exception in exposed API method: get_total_size
Uncaught exception in exposed API method:

```
  File "lib/galaxy/web/framework/decorators.py", line 282, in decorator
  File "lib/galaxy/webapps/galaxy/api/jobs.py", line 184, in common_problems
    if input_dataset_instance.get_total_size() == 0:
AttributeError: 'NoneType' object has no attribute 'get_total_size'
```

mvdbeek:
The traceback is
```
  File "lib/galaxy/web/framework/decorators.py", line 282, in decorator
  File "lib/galaxy/webapps/galaxy/api/jobs.py", line 184, in common_problems
    if input_dataset_instance.get_total_size() == 0:
AttributeError: 'NoneType' object has no attribute 'get_total_size'
```

... now it would be great to know when this happens, cause it looks
like it shouldn't happen at all.

Fixes #8448

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/jobs.py
+++ b/lib/galaxy/webapps/galaxy/api/jobs.py
@@ -181,7 +181,7 @@ class JobController(BaseAPIController, UsesLibraryMixinItems):
         has_duplicate_inputs = False
         for job_input_assoc in job.input_datasets:
             input_dataset_instance = job_input_assoc.dataset
-            if input_dataset_instance is None:
+            if not input_dataset_instance:
                 continue
             if input_dataset_instance.get_total_size() == 0:
                 has_empty_inputs = True


lib/galaxy/webapps/galaxy/api/jobs.py
```diff
@@ -181,7 +181,7 @@ def common_problems(self, trans, id, **kwd):
         has_duplicate_inputs = False
         for job_input_assoc in job.input_datasets:
             input_dataset_instance = job_input_assoc.dataset
-            if input_dataset_instance is None:
+            if not input_dataset_instance:
```
nsoranzo, line 184: Would that make a difference for the reported traceback? It says `'NoneType' object has no attribute 'get_total_size'` , so `input_dataset_instance` must be `None`.

mvdbeek: Thanks for pointing out https://github.com/galaxyproject/galaxy/pull/8194, I had concluded because the check is already `is None` that we'd have something like `''`. So I guess this is probably already fixed.

mvdbeek: ah, and of course that doesn't make sense, because `''` is not NoneType.

PR #8489, opened 2019-08-21
Update requirements and use pyuwsgi

mvdbeek:
Discussing this on gitter today @nsoranzo and me found out that the pywusgi wheels are not built with libyaml, so use the same embedded parser that our custom wheels use. Our wheels appears to be not built with ssl support, and neither are the pyuwsgi wheels. So we should be able to use this as a drop-in replacement.

### DIFF
--- a/lib/galaxy/dependencies/pipfiles/default/Pipfile
+++ b/lib/galaxy/dependencies/pipfiles/default/Pipfile
@@ -38,7 +38,7 @@ SQLAlchemy-Utils = "*"
 Mercurial = {version = "<=3.7.3", markers = "python_version < '3'"}
 nodeenv = "*"
 pycryptodome = "*"
-uWSGI = "*"
+pyuwsgi = "*"
 pysam = "==0.15.2"
 bdbag = "==1.4.1"  # 1.5.0 requires Python >=2.7.9
 bleach = "*"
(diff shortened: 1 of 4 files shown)


nsoranzo:
I think you should update the following 3 imports to `pyuwsgi`:
```
lib/galaxy/util/__init__.py
63:    import uwsgi

lib/galaxy/web_stack/__init__.py
13:    import uwsgi

lib/galaxy/web_stack/transport.py
11:    import uwsgi
```

mvdbeek:
Does that also work for uwsgi built from the source tarball ? If not, are we ok with that ?

mvdbeek:
I see, it doesn't actually work with pyuwsgi (if used in the REPL, but see next answer). ~~Should we do try / except ImportError and fallback or just go all the way ?~~

mvdbeek:
Wait, it's more complicated than that, the uwsgi module is injected by uwgi (https://uwsgi-docs.readthedocs.io/en/latest/PythonModule.html#the-uwsgi-python-module). I think that probably hasn't changed ? The `pyuwsgi` module just seems to offer some additional ways to launch uwsgi.

nsoranzo:
If that's the case and everything works, I'm happy with that!

nsoranzo (approved):

natefoo (approved):
Would there be any benefit/drawback from explicitly calling `pyuwsgi` rather than `uwsgi`?

mvdbeek:
I guess on the off chance that someone would want to use another uwsgi not calling pyuwsgi might be considered an advantage ? On the other hand that'd be a problem if we use some pyuwsgi-specific things, like the possibility to run uwsgi via `pyuwsgi.run()`. Since we don't do that I'd be in favor of keeping it as is

natefoo:
Good enough, we can always adjust later if needed. Thanks @mvdbeek!

PR #8490, opened 2019-08-21
Fix some data_dir stuff in galaxy.config.

jmchilton:


lib/galaxy/config/__init__.py
```diff
@@ -587,7 +586,7 @@ def __init__(self, **kwargs):
             self.nginx_upload_store = os.path.abspath(self.nginx_upload_store)
         self.object_store = kwargs.get('object_store', 'disk')
         self.object_store_check_old_style = string_as_bool(kwargs.get('object_store_check_old_style', False))
-        self.object_store_cache_path = resolve_path(kwargs.get("object_store_cache_path", "database/object_store_cache"), self.data_dir)
+        self.object_store_cache_path = resolve_path(kwargs.get("object_store_cache_path", "object_store_cache"), self.data_dir)
```
```suggestion
        self.object_store_cache_path = self.resolve_path(kwargs.get("object_store_cache_path", os.path.join(self.data_dir, "object_store_cache")))
```

lib/galaxy/config/__init__.py
```diff
@@ -371,7 +370,7 @@ def __init__(self, **kwargs):
         # should probably drop the backward compatiblity to save the path check.
         self.container_image_cache_path = resolve_path(kwargs.get("container_image_cache_path", "container_images"), self.data_dir)
         if not os.path.exists(self.container_image_cache_path):
-            self.container_image_cache_path = self.resolve_path(kwargs.get("container_image_cache_path", "database/container_cache"))
+            self.container_image_cache_path = resolve_path(kwargs.get("container_image_cache_path", "container_cache"), self.data_dir)
```
nsoranzo: Wouldn't this break setups where in `galaxy.yml` this property is configured to a path relative to the Galaxy root?
```suggestion
            self.container_image_cache_path = self.resolve_path(kwargs.get("container_image_cache_path", os.path.join(self.data_dir, "container_cache")))
```

lib/galaxy/config/__init__.py
```diff
@@ -353,7 +353,6 @@ def __init__(self, **kwargs):
         self.template_path = resolve_path(kwargs.get("template_path", "templates"), self.root)
         self.template_cache = resolve_path(kwargs.get("template_cache_path", "compiled_templates"), self.data_dir)
         self.job_queue_cleanup_interval = int(kwargs.get("job_queue_cleanup_interval", "5"))
-        self.cluster_files_directory = os.path.abspath(kwargs.get("cluster_files_directory", "database/pbs"))
         self.cluster_files_directory = os.path.abspath(resolve_path(kwargs.get("cluster_files_directory", "pbs"), self.data_dir))
```
```suggestion
        self.cluster_files_directory = self.resolve_path(kwargs.get("cluster_files_directory", os.path.join(self.data_dir, "pbs")))
```

lib/galaxy/config/__init__.py
jmchilton: I assumed data_dir was absolute - but I guess not. I based this on some other code - I wonder if they were handled differently somewhere else though.

nsoranzo: `self.data_dir` should be absolute, but `container_image_cache_path` could be set to a relative path, which should be still be interpreted w.r.t. `self.root`, not `self.data_dir` .

nsoranzo (approved):

PR #8492, opened 2019-08-21
[19.05] remove gops from venv list

bgruening:
I'm not sure why this tool was in the list but it is broken on usegalaxy.eu and usegalaxy.org and we get bad GTN evaluations.

@natefoo you probably want to patch this.

### DIFF
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -174,11 +174,12 @@ GALAXY_LIB_TOOLS_UNVERSIONED = [
 # Tools that needed galaxy on the PATH in the past but no longer do along
 # with the version at which they were fixed.
 GALAXY_LIB_TOOLS_VERSIONED = {
-    "sam_to_bam": packaging.version.parse("1.1.3"),
-    "PEsortedSAM2readprofile": packaging.version.parse("1.1.1"),
-    "fetchflank": packaging.version.parse("1.0.1"),
     "Extract genomic DNA 1": packaging.version.parse("3.0.0"),
+    "fetchflank": packaging.version.parse("1.0.1"),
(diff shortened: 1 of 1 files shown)


bgruening:
@wm75 if you ever wonder why the intersect tool ... :) Found it! Now I deserve cold milk.

mvdbeek:
So the situation is that the tool doesn't use its actual, annotated requirement, but it works if removed from the this list ? In that case you need to move the tool to `GALAXY_LIB_TOOLS_VERSIONED ` with the current version that works.

bgruening:
Thanks, @mvdbeek added it!

mvdbeek:
Hmm, this isn't right though, the tool imports this:
```
from galaxy.tools.util.galaxyops import fail, parse_cols_arg, skipped
```
so it does need an importable Galaxy.
I don't quite understand how this can work if the tool is not on that list ? Does the job handler environment define a PYTHONPATH or similar that points to ~~Galaxy's virtualenv~~ Galaxy's lib folder ?

The tutorials could also use bedtools, I don't think this tools does anything bedtools can't do (better) ?

bgruening:
@mvdbeek this is part of the package:

```
galaxy@sn04:~/$ ls -l /usr/local/tools/_conda/envs/__galaxy-ops\@1.1.0/lib/python2.7/site-packages/galaxy/tools/util/
total 3
drwxrwxr-x. 2 galaxy galaxy   4 Aug 25  2017 galaxyops
-rw-rw-r--. 2 galaxy galaxy   0 Jun  6  2017 __init__.py
-rw-rw-r--. 2 galaxy galaxy 171 Jun  6  2017 __init__.pyc
```

mvdbeek:
That clears that up!

mvdbeek (approved):

PR #8495, opened 2019-08-22
Fix two bugs introduced in #8078

mvdbeek:


mvdbeek:
The failing tests will be fixed by the other open PRs

Issue #8496, opened 2019-08-22
Issue with external metadata setting on dev


mvdbeek:
Currently fails with
```
galaxy.jobs.runners.local DEBUG 2019-08-22 11:02:53,758 [p:36783,w:0,m:2] [LocalRunner.work_thread-2] execution finished: /Users/mvandenb/src/galaxy/database/jobs_directory/000/4/galaxy_4.sh
galaxy.metadata DEBUG 2019-08-22 11:02:53,806 [p:36783,w:0,m:2] [LocalRunner.work_thread-2] setting metadata externally failed for HistoryDatasetAssociation 6: 'MutationList' object has no attribute '_key'
galaxy.jobs DEBUG 2019-08-22 11:02:53,945 [p:36783,w:0,m:2] [LocalRunner.work_thread-2] job 4 ended (finish() executed in (171.134 ms))
galaxy.jobs.runners.local DEBUG 2019-08-22 11:02:53,973 [p:36783,w:0,m:2] [LocalRunner.work_thread-3] execution finished: /Users/mvandenb/src/galaxy/database/jobs_directory/000/5/galaxy_5.sh
galaxy.metadata DEBUG 2019-08-22 11:02:54,015 [p:36783,w:0,m:2] [LocalRunner.work_thread-3] setting metadata externally failed for HistoryDatasetAssociation 7: 'MutationList' object has no attribute '_key'
```

might be py3 related

PR #8497, opened 2019-08-22
Fix setting metadata externally

mvdbeek:
First step is not allowing internal metadata setting in the tests.

### DIFF
--- a/lib/galaxy/datatypes/binary.py
+++ b/lib/galaxy/datatypes/binary.py
@@ -1024,7 +1024,7 @@ class Biom2(H5):
             elif 'format' in attributes:  # biom 2.0
                 dataset.metadata.format = util.unicodify(attributes['format'])
             dataset.metadata.type = util.unicodify(attributes['type'])
-            dataset.metadata.shape = tuple(attributes['shape'])
+            dataset.metadata.shape = tuple((util.unicodify(_) for _ in attributes['shape']))
             dataset.metadata.generated_by = util.unicodify(attributes['generated-by'])
             dataset.metadata.creation_date = util.unicodify(attributes['creation-date'])
             dataset.metadata.nnz = int(attributes['nnz'])
--- a/lib/galaxy/datatypes/binary.py
+++ b/lib/galaxy/datatypes/binary.py
@@ -1017,20 +1017,20 @@ class Biom2(H5):
             f = h5py.File(dataset.file_name)
             attributes = dict(f.attrs.items())
 
-            dataset.metadata.id = attributes['id']
-            dataset.metadata.format_url = attributes['format-url']
+            dataset.metadata.id = util.unicodify(attributes['id'])
+            dataset.metadata.format_url = util.unicodify(attributes['format-url'])
             if 'format-version' in attributes:  # biom 2.1
-                dataset.metadata.format_version = '.'.join(map(str, list(attributes['format-version'])))
+                dataset.metadata.format_version = '.'.join((util.unicodify(_) for _ in list(attributes['format-version'])))
             elif 'format' in attributes:  # biom 2.0
-                dataset.metadata.format = attributes['format']
-            dataset.metadata.type = attributes['type']
+                dataset.metadata.format = util.unicodify(attributes['format'])
+            dataset.metadata.type = util.unicodify(attributes['type'])
             dataset.metadata.shape = tuple(attributes['shape'])
-            dataset.metadata.generated_by = attributes['generated-by']
-            dataset.metadata.creation_date = attributes['creation-date']
+            dataset.metadata.generated_by = util.unicodify(attributes['generated-by'])
+            dataset.metadata.creation_date = util.unicodify(attributes['creation-date'])
             dataset.metadata.nnz = int(attributes['nnz'])
 
         except Exception as e:
-            log.warning('%s, set_meta Exception: %s', self, e)
+            log.warning('%s, set_meta Exception: %s', self, util.unicodify(e))
 
     def set_peek(self, dataset, is_multi_byte=False):
         if not dataset.dataset.purged:
@@ -1038,9 +1038,9 @@ class Biom2(H5):
             try:
                 f = h5py.File(dataset.file_name)
                 for k, v in dict(f.attrs).items():
-                    lines.append('%s:  %s' % (k, v))
+                    lines.append('%s:  %s' % (k, util.unicodify(v)))
             except Exception as e:
-                log.warning('%s, set_peek Exception: %s', self, e)
+                log.warning('%s, set_peek Exception: %s', self, util.unicodify(e))
             dataset.peek = '\n'.join(lines)
             dataset.blurb = nice_size(dataset.get_size())
         else:
--- a/lib/galaxy/datatypes/binary.py
+++ b/lib/galaxy/datatypes/binary.py
@@ -811,7 +811,7 @@ class Loom(H5):
     def sniff(self, filename):
         if super(Loom, self).sniff(filename):
             try:
-                with h5py.File(filename) as loom_file:
+                with h5py.File(filename, 'r') as loom_file:
                     return bool(loom_file.attrs.get('LOOM_SPEC_VERSION', False))
             except Exception:
                 return False
@@ -834,13 +834,13 @@ class Loom(H5):
     def set_meta(self, dataset, overwrite=True, **kwd):
         super(Loom, self).set_meta(dataset, overwrite=overwrite, **kwd)
         try:
-            with h5py.File(dataset.file_name) as loom_file:
-                dataset.metadata.title = util.unicodify(loom_file.attrs.get('title', None))
-                dataset.metadata.description = util.unicodify(loom_file.attrs.get('description', None))
-                dataset.metadata.url = util.unicodify(loom_file.attrs.get('url', None))
-                dataset.metadata.doi = util.unicodify(loom_file.attrs.get('doi', None))
-                dataset.metadata.loom_spec_version = util.unicodify(loom_file.attrs.get('LOOM_SPEC_VERSION', None))
-                dataset.creation_date = util.unicodify(loom_file.attrs.get('creation_date', None))
+            with h5py.File(dataset.file_name, 'r') as loom_file:
+                dataset.metadata.title = util.unicodify(loom_file.attrs.get('title'))
+                dataset.metadata.description = util.unicodify(loom_file.attrs.get('description'))
+                dataset.metadata.url = util.unicodify(loom_file.attrs.get('url'))
+                dataset.metadata.doi = util.unicodify(loom_file.attrs.get('doi'))
+                dataset.metadata.loom_spec_version = util.unicodify(loom_file.attrs.get('LOOM_SPEC_VERSION'))
+                dataset.creation_date = util.unicodify(loom_file.attrs.get('creation_date'))
                 dataset.metadata.shape = tuple(loom_file['matrix'].shape)
 
                 tmp = list(loom_file['layers'].keys())
@@ -880,7 +880,7 @@ class Anndata(H5):
     def sniff(self, filename):
         if super(Anndata, self).sniff(filename):
             try:
-                with h5py.File(filename) as f:
+                with h5py.File(filename, 'r') as f:
                     return all(attr in f for attr in ['X', 'obs', 'var'])
             except Exception:
                 return False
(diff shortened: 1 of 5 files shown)


mvdbeek:
Hmm, looks like the tests don't show the same issue as in https://github.com/galaxyproject/galaxy/issues/8496. I'll check some other python versions, but fwiw this fixes it for me:
```
diff --git a/lib/galaxy/metadata/set_metadata.py b/lib/galaxy/metadata/set_metadata.py
index 5248696637..4bf0c31dc5 100644
--- a/lib/galaxy/metadata/set_metadata.py
+++ b/lib/galaxy/metadata/set_metadata.py
@@ -18,6 +18,8 @@ import sys
 from six.moves import cPickle
 from sqlalchemy.orm import clear_mappers
 
+import galaxy.model.custom_types
+galaxy.model.custom_types.MutationList._key = None  # when run externally sqlalchemy doesn't populate _key
 import galaxy.model.mapping  # need to load this before we unpickle, in order to setup properties assigned by the mappers
 from galaxy.model.custom_types import total_size
 from galaxy.tool_util.provided_metadata import parse_tool_provided_metadata
@@ -29,8 +31,6 @@ from galaxy.util import (
 logging.basicConfig()
 
-galaxy.model.Job()  # this looks REAL stupid, but it is REQUIRED in order for SA to insert parameters into the classes defined by the mappers --> it appears that instantiating ANY mapper'ed class would suffice here
-
 
 def set_meta_with_tool_provided(dataset_instance, file_dict, set_meta_kwds, datatypes_registry, max_metadata_value_size):
     # This method is somewhat odd, in that we set the metadata attributes from tool,
```

nsoranzo:
Where does the `_key` attribute come from?

mvdbeek:
I suppose the sqlalchemy mapper, since the internal fallback works ... I'm looking into that now.

mvdbeek:
On the bright side this did catch some unexpected issues: https://jenkins.galaxyproject.org/job/docker-integration-py3/198/

mvdbeek:
OK, I got the full traceback which is more helpful:
```
galaxy.metadata DEBUG 2019-08-22 14:27:06,483 [p:4375,w:0,m:1] [LocalRunner.work_thread-1] setting metadata externally failed for HistoryDatasetAssociation 51: Traceback (most recent call last):
  File "/Users/mvandenb/src/galaxy/lib/galaxy/metadata/set_metadata.py", line 108, in set_metadata_portable
    dataset = cPickle.load(open(filename_in, 'rb'))  # load DatasetInstance
  File "/Users/mvandenb/src/galaxy/lib/galaxy/model/custom_types.py", line 245, in extend
    list.extend(self, (MutationObj.coerce(self._key, v) for v in values))
  File "/Users/mvandenb/src/galaxy/lib/galaxy/model/custom_types.py", line 245, in <genexpr>
    list.extend(self, (MutationObj.coerce(self._key, v) for v in values))
AttributeError: 'MutationList' object has no attribute '_key'
```

I got this with
```diff
diff --git a/lib/galaxy/metadata/set_metadata.py b/lib/galaxy/metadata/set_metadata.py
index 5248696637..562e4a0ff2 100644
--- a/lib/galaxy/metadata/set_metadata.py
+++ b/lib/galaxy/metadata/set_metadata.py
@@ -14,6 +14,7 @@ import json
 import logging
 import os
 import sys
+import traceback

 from six.moves import cPickle
 from sqlalchemy.orm import clear_mappers
@@ -123,7 +124,7 @@ def set_metadata_portable():
             dataset.metadata.to_JSON_dict(filename_out)  # write out results of set_meta
             json.dump((True, 'Metadata has been set successfully'), open(filename_results_code, 'wt+'))  # setting metadata has succeeded
         except Exception as e:
-            json.dump((False, unicodify(e)), open(filename_results_code, 'wt+'))  # setting metadata has failed somehow
+            json.dump((False, traceback.format_exc()), open(filename_results_code, 'wt+'))  # setting metadata has failed somehow

     write_job_metadata(tool_job_working_directory, job_metadata, set_meta, tool_provided_metadata)
```
which is maybe not a bad way to log this stuff ?

nsoranzo:
> which is maybe not a bad way to log this stuff ?

For the `set_metadata` special case probably yes, normally `log.exception()` takes care of that.

mvdbeek:
This is ready, the failing tests should be transient.

lib/galaxy/datatypes/binary.py
```diff
@@ -1017,30 +1017,30 @@ def set_meta(self, dataset, overwrite=True, **kwd):
...
+                dataset.metadata.format_version = '.'.join((util.unicodify(_) for _ in list(attributes['format-version'])))
             elif 'format' in attributes:  # biom 2.0
-                dataset.metadata.format = attributes['format']
-            dataset.metadata.type = attributes['type']
-            dataset.metadata.shape = tuple(attributes['shape'])
-            dataset.metadata.generated_by = attributes['generated-by']
-            dataset.metadata.creation_date = attributes['creation-date']
+                dataset.metadata.format = util.unicodify(attributes['format'])
+            dataset.metadata.type = util.unicodify(attributes['type'])
+            dataset.metadata.shape = tuple((util.unicodify(_) for _ in attributes['shape']))
```
nsoranzo: These are numbers in a `numpy.ndarray`, so the change is not necessary.
```suggestion
            dataset.metadata.shape = tuple(attributes['shape'])
```

lib/galaxy/datatypes/binary.py
```diff
@@ -1017,30 +1017,30 @@ def set_meta(self, dataset, overwrite=True, **kwd):
             f = h5py.File(dataset.file_name)
             attributes = dict(f.attrs.items())
 
-            dataset.metadata.id = attributes['id']
-            dataset.metadata.format_url = attributes['format-url']
+            dataset.metadata.id = util.unicodify(attributes['id'])
+            dataset.metadata.format_url = util.unicodify(attributes['format-url'])
             if 'format-version' in attributes:  # biom 2.1
-                dataset.metadata.format_version = '.'.join(map(str, list(attributes['format-version'])))
+                dataset.metadata.format_version = '.'.join((util.unicodify(_) for _ in list(attributes['format-version'])))
```
nsoranzo: These are numbers in a numpy.ndarray, so `str()` should be enough.
```suggestion
                dataset.metadata.format_version = '.'.join(str(_) for _ in attributes['format-version'])
```

lib/galaxy/datatypes/binary.py
```diff
@@ -1017,30 +1017,30 @@ def set_meta(self, dataset, overwrite=True, **kwd):
             f = h5py.File(dataset.file_name)
```
```suggestion
            f = h5py.File(dataset.file_name, 'r')
```

Also use a `with` statement?

lib/galaxy/datatypes/binary.py
```diff
@@ -835,12 +835,12 @@ def set_meta(self, dataset, overwrite=True, **kwd):
         super(Loom, self).set_meta(dataset, overwrite=overwrite, **kwd)
         try:
             with h5py.File(dataset.file_name) as loom_file:
-                dataset.metadata.title = loom_file.attrs.get('title', None)
-                dataset.metadata.description = loom_file.attrs.get('description', None)
-                dataset.metadata.url = loom_file.attrs.get('url', None)
-                dataset.metadata.doi = loom_file.attrs.get('doi', None)
-                dataset.metadata.loom_spec_version = loom_file.attrs.get('LOOM_SPEC_VERSION', None)
-                dataset.creation_date = loom_file.attrs.get('creation_date', None)
+                dataset.metadata.title = util.unicodify(loom_file.attrs.get('title', None))
```
```suggestion
                dataset.metadata.title = util.unicodify(loom_file.attrs.get('title'))
```

lib/galaxy/datatypes/binary.py
```diff
@@ -835,12 +835,12 @@ def set_meta(self, dataset, overwrite=True, **kwd):
...
+                dataset.metadata.description = util.unicodify(loom_file.attrs.get('description', None))
```
```suggestion
                dataset.metadata.description = util.unicodify(loom_file.attrs.get('description'))
```

lib/galaxy/datatypes/binary.py
```diff
@@ -835,12 +835,12 @@ def set_meta(self, dataset, overwrite=True, **kwd):
...
+                dataset.metadata.url = util.unicodify(loom_file.attrs.get('url', None))
```
```suggestion
                dataset.metadata.url = util.unicodify(loom_file.attrs.get('url'))
```

lib/galaxy/datatypes/binary.py
```diff
@@ -835,12 +835,12 @@ def set_meta(self, dataset, overwrite=True, **kwd):
...
+                dataset.metadata.doi = util.unicodify(loom_file.attrs.get('doi', None))
```
```suggestion
                dataset.metadata.doi = util.unicodify(loom_file.attrs.get('doi'))
```

lib/galaxy/datatypes/binary.py
```diff
@@ -835,12 +835,12 @@ def set_meta(self, dataset, overwrite=True, **kwd):
...
+                dataset.metadata.loom_spec_version = util.unicodify(loom_file.attrs.get('LOOM_SPEC_VERSION', None))
```
```suggestion
                dataset.metadata.loom_spec_version = util.unicodify(loom_file.attrs.get('LOOM_SPEC_VERSION'))
```

lib/galaxy/datatypes/binary.py
```diff
@@ -835,12 +835,12 @@ def set_meta(self, dataset, overwrite=True, **kwd):
...
+                dataset.creation_date = util.unicodify(loom_file.attrs.get('creation_date', None))
```
```suggestion
                dataset.creation_date = util.unicodify(loom_file.attrs.get('creation_date'))
```

lib/galaxy/datatypes/binary.py
```diff
@@ -835,12 +835,12 @@ def set_meta(self, dataset, overwrite=True, **kwd):
         super(Loom, self).set_meta(dataset, overwrite=overwrite, **kwd)
         try:
             with h5py.File(dataset.file_name) as loom_file:
```
```suggestion
            with h5py.File(dataset.file_name, 'r') as loom_file:
```

otherwise the file is locked.

The same should be fixed at lines 814, 883 and 1006.

lib/galaxy/datatypes/binary.py
```diff
@@ -1017,30 +1017,30 @@ def set_meta(self, dataset, overwrite=True, **kwd):
...
 
         except Exception as e:
-            log.warning('%s, set_meta Exception: %s', self, e)
+            log.warning('%s, set_meta Exception: %s', self, util.unicodify(e))
 
     def set_peek(self, dataset, is_multi_byte=False):
         if not dataset.dataset.purged:
             lines = ['Biom2 (HDF5) file']
             try:
                 f = h5py.File(dataset.file_name)
```
```suggestion
                f = h5py.File(dataset.file_name, 'r')
```

Also this file is not explicitly closed, use `with` statement?

lib/galaxy/datatypes/binary.py
mvdbeek: I thought so too, but https://jenkins.galaxyproject.org/job/docker-integration-py3/202/testReport/test.integration/test_datatype_upload/test_upload_datatype_auto_biom2_sparse_otu_table_hdf5_biom2_/ led me to add the https://github.com/galaxyproject/galaxy/pull/8497/commits/3dd500478df5e0db96de502fe14569cf92462d75

mvdbeek: ah, I guess they are some of the numpy numbers, like np.int8 or something like that, I can cast to int instead (if these are always int).

lib/galaxy/datatypes/binary.py
```diff
@@ -1002,45 +1002,38 @@ def sniff(self, filename):
...
+        with h5py.File(dataset.file_name, 'r') as f:
             attributes = dict(f.attrs.items())
 
-            dataset.metadata.id = attributes['id']
-            dataset.metadata.format_url = attributes['format-url']
+            dataset.metadata.id = util.unicodify(attributes['id'])
+            dataset.metadata.format_url = util.unicodify(attributes['format-url'])
             if 'format-version' in attributes:  # biom 2.1
-                dataset.metadata.format_version = '.'.join(map(str, list(attributes['format-version'])))
+                dataset.metadata.format_version = '.'.join((str(_) for _ in list(attributes['format-version'])))
```
```suggestion
                dataset.metadata.format_version = '.'.join(str(_) for _ in attributes['format-version'])
```

lib/galaxy/datatypes/binary.py
```diff
@@ -1002,45 +1002,38 @@ def sniff(self, filename):
...
-            dataset.metadata.creation_date = attributes['creation-date']
+                dataset.metadata.format = util.unicodify(attributes['format'])
+            dataset.metadata.type = util.unicodify(attributes['type'])
+            dataset.metadata.shape = tuple((int(_) for _ in attributes['shape']))
+            dataset.metadata.generated_by = util.unicodify(attributes['generated-by'])
+            dataset.metadata.creation_date = util.unicodify(attributes['creation-date'])
             dataset.metadata.nnz = int(attributes['nnz'])
 
-        except Exception as e:
-            log.warning('%s, set_meta Exception: %s', self, e)
```
nsoranzo, line 1033: We are a bit inconsistent in how we deal with exceptions in `set_meta` methods, should we just log them or let them bubble up?
See e.g. https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/datatypes/binary.py#L308

mvdbeek: I guess that depends case by case whether or not the dataset works correctly. I think I didn't mean to remove this one here though.

nsoranzo: OK, so if you restore it I think it's ready to merge.

nsoranzo:
@mvdbeek Did you mean to include commit 75295b5eae3a2935b7f11204e54236fc8f497315 here ?

mvdbeek:
Nope, one sec

lib/galaxy/datatypes/binary.py
```diff
@@ -1002,45 +1002,41 @@ def sniff(self, filename):
         False
         """
         if super(Biom2, self).sniff(filename):
-            try:
-                f = h5py.File(filename)
+            with h5py.File(filename, 'r') as f:
                 attributes = list(dict(f.attrs.items()))
                 required_fields = ['id', 'format-url', 'type', 'generated-by', 'creation-date', 'nnz', 'shape']
```
```suggestion
                required_fields = {'id', 'format-url', 'type', 'generated-by', 'creation-date', 'nnz', 'shape'}
```

lib/galaxy/datatypes/binary.py
```diff
@@ -1002,45 +1002,41 @@ def sniff(self, filename):
...
                 return set(required_fields).issubset(attributes)
```
```suggestion
                return required_fields.issubset(f.attrs.keys())
```

lib/galaxy/datatypes/binary.py
```diff
@@ -1002,45 +1002,41 @@ def sniff(self, filename):
         False
         """
         if super(Biom2, self).sniff(filename):
-            try:
-                f = h5py.File(filename)
+            with h5py.File(filename, 'r') as f:
                 attributes = list(dict(f.attrs.items()))
```
```suggestion
```

lib/galaxy/datatypes/binary.py
```diff
@@ -1002,45 +1002,41 @@ def sniff(self, filename):
...
-            elif 'format' in attributes:  # biom 2.0
-                dataset.metadata.format = attributes['format']
-            dataset.metadata.type = attributes['type']
-            dataset.metadata.shape = tuple(attributes['shape'])
-            dataset.metadata.generated_by = attributes['generated-by']
-            dataset.metadata.creation_date = attributes['creation-date']
-            dataset.metadata.nnz = int(attributes['nnz'])
-
+            with h5py.File(dataset.file_name, 'r') as f:
+                attributes = dict(f.attrs.items())
```
```suggestion
                attributes = f.attrs
```

lib/galaxy/datatypes/binary.py
```diff
@@ -1002,45 +1002,41 @@ def sniff(self, filename):
...
 
     def set_peek(self, dataset, is_multi_byte=False):
         if not dataset.dataset.purged:
             lines = ['Biom2 (HDF5) file']
             try:
-                f = h5py.File(dataset.file_name)
-                for k, v in dict(f.attrs).items():
-                    lines.append('%s:  %s' % (k, v))
+                with h5py.File(dataset.file_name) as f:
+                    for k, v in dict(f.attrs).items():
```
```suggestion
                    for k, v in f.attrs.items():
```

nsoranzo (approved):

PR #8498, opened 2019-08-22
Fix transiently failing test_cancel_job_with_admin_message test

mvdbeek:
This was originally fixed in 9b815ce7a6b7c774340fb27b91919126048bd8fb
which somehow got lost.

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/jobs.py
+++ b/lib/galaxy/webapps/galaxy/api/jobs.py
@@ -472,12 +472,14 @@ class JobController(BaseAPIController, UsesVisualizationMixin):
                 decoded_job_id = self.decode_id(job_id)
             except Exception:
                 raise exceptions.MalformedId()
-            return self.job_manager.get_accessible_job(trans, decoded_job_id)
+            job = self.job_manager.get_accessible_job(trans, decoded_job_id)
         else:
             hda_ldda = kwd.get("hda_ldda", "hda")
             # Following checks dataset accessible
             dataset_instance = self.get_hda_or_ldda(trans, hda_ldda=hda_ldda, dataset_id=dataset_id)
-            return dataset_instance.creating_job
+            job = dataset_instance.creating_job
+        trans.sa_session.refresh(job)
+        return job
 
     @expose_api
     def create(self, trans, payload, **kwd):
--- a/lib/galaxy/webapps/galaxy/api/jobs.py
+++ b/lib/galaxy/webapps/galaxy/api/jobs.py
@@ -472,14 +472,12 @@ class JobController(BaseAPIController, UsesVisualizationMixin):
                 decoded_job_id = self.decode_id(job_id)
             except Exception:
                 raise exceptions.MalformedId()
-            job = self.job_manager.get_accessible_job(trans, decoded_job_id)
+            return self.job_manager.get_accessible_job(trans, decoded_job_id)
         else:
             hda_ldda = kwd.get("hda_ldda", "hda")
             # Following checks dataset accessible
             dataset_instance = self.get_hda_or_ldda(trans, hda_ldda=hda_ldda, dataset_id=dataset_id)
-            job = dataset_instance.creating_job
-        trans.sa_session.refresh(job)
-        return job
+            return dataset_instance.creating_job
 
     @expose_api
     def create(self, trans, payload, **kwd):
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -3115,8 +3115,10 @@ class HistoryDatasetAssociation(DatasetInstance, HasTags, Dictifiable, UsesAnnot
         # Since this class is a proxy to rather complex attributes we want to
         # display in other objects, we can't use the simpler method used by
         # other model classes.
+        log.debug("Getting HDA summary - start")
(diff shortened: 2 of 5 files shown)


lib/galaxy/webapps/galaxy/api/jobs.py
```diff
@@ -472,12 +472,14 @@ def __get_job(self, trans, job_id=None, dataset_id=None, **kwd):
...
-            return self.job_manager.get_accessible_job(trans, decoded_job_id)
+            job = self.job_manager.get_accessible_job(trans, decoded_job_id)
         else:
             hda_ldda = kwd.get("hda_ldda", "hda")
             # Following checks dataset accessible
             dataset_instance = self.get_hda_or_ldda(trans, hda_ldda=hda_ldda, dataset_id=dataset_id)
-            return dataset_instance.creating_job
+            job = dataset_instance.creating_job
+        trans.sa_session.refresh(job)
+        return job
```
jmchilton: This doesn't seem like the right fix - does it work for sure?

Update: I'll look at it more though.

jmchilton:
example... https://jenkins.galaxyproject.org/job/docker-integration/10282/testReport/junit/test.integration.test_local_job_cancellation/LocalJobCancellationTestCase/test_cancel_job_with_admin_message/

... expunge_all() call which should mitigate the need for this: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/web/framework/webapp.py#L189

... maybe this API endpoint doesn't create a ``trans`` - that doesn't sound right though.

mvdbeek:
hmm, should that possibly be expire_all() instead of expunge_all() (https://docs.sqlalchemy.org/en/13/orm/session_state_management.html#when-to-expire-or-refresh)?

jmchilton:
My guess based on the one job I've seen is that the problem is the same as we are using this hack (https://github.com/galaxyproject/galaxy/blob/dev/test/api/test_jobs.py#L346) for. When I looked at this before, I think the problem is related to the job completing too quickly - and the polling thread blocking on something... I never really figured out what was happening though. I'm not sure though - I guess this isn't consistent with refresh fixing the problem.

mvdbeek:
> My guess based on the one job I've seen is that the problem is the same as we are using this hack (https://github.com/galaxyproject/galaxy/blob/dev/test/api/test_jobs.py#L346) for.

It might be the same issue indeed, but the job runs for 240 seconds, I can't see how that can enter `ok` too fast, unless there is another bug with the job states. I'll keep digging to confirm or disregard my hypothesis.

mvdbeek:
You might be right actually, those timings are weird:
```
galaxy.jobs.handler INFO 2019-08-21 02:13:28,554 (2) Job dispatched
galaxy.jobs.command_factory INFO 2019-08-21 02:13:28,673 Built script [/tmp/tmpzrL2Yt/tmpD9aWUp/tmplrqCon/database/job_working_directory_8NKtng/000/2/tool_script.sh] for tool command [cat /tmp/tmpzrL2Yt/tmpD9aWUp/tmplrqCon/database/files/7/f/8/dataset_7f846d48-247d-4587-9675-570d7848e2e9.dat  > /tmp/tmpzrL2Yt/tmpD9aWUp/tmplrqCon/database/files/7/e/b/dataset_7eb50a3f-f35e-4d81-8cda-73957b36fa5a.dat; sleep '240']
galaxy.jobs.runners DEBUG 2019-08-21 02:13:28,696 (2) command is: rm -rf working; mkdir -p working; cd working; /bin/bash /tmp/tmpzrL2Yt/tmpD9aWUp/tmplrqCon/database/job_working_directory_8NKtng/000/2/tool_script.sh > ../tool_stdout 2> ../tool_stderr; return_code=$?; cd '/tmp/tmpzrL2Yt/tmpD9aWUp/tmplrqCon/database/job_working_directory_8NKtng/000/2'; 
python "metadata/set.py"; sh -c "exit $return_code"
galaxy.jobs.runners.local DEBUG 2019-08-21 02:13:28,757 (2) executing job script: /tmp/tmpzrL2Yt/tmpD9aWUp/tmplrqCon/database/job_working_directory_8NKtng/000/2/galaxy_2.sh
galaxy.jobs DEBUG 2019-08-21 02:13:28,823 (2) Persisting job destination (destination id: local)
galaxy.jobs.runners.local DEBUG 2019-08-21 02:17:32,065 execution finished: /tmp/tmpzrL2Yt/tmpD9aWUp/tmplrqCon/database/job_working_directory_8NKtng/000/2/galaxy_2.sh
galaxy.model.metadata DEBUG 2019-08-21 02:17:32,090 loading metadata from file for: HistoryDatasetAssociation 2
galaxy.jobs DEBUG 2019-08-21 02:17:32,144 job 2 ended (finish() executed in (70.253 ms))
galaxy.tool_util.verify.interactor INFO 2019-08-21 02:19:08,287 Wait for job to start running exceeded walltime [total 102.2, max 60], terminating.
```

It looks like the job submission API is not returning before the job has finished. That's weird 🤔

jmchilton:
Some sort of DB lock maybe? Maybe we should slow request logging on for the tests. Maybe we should tail the postgres logs in the container and capture that somehow with the test data. Thanks for looking at this @mvdbeek - clearly I tried before and gave up.

PR #8501, opened 2019-08-22
Update node-sass, old version not compatible with node 12

innovate-invent:
node 12 is not compatible with sass 4.11
https://github.com/sass/node-sass/issues/2648

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -104,7 +104,7 @@
     "karma-webpack": "^4.0.0-rc.2",
     "mini-css-extract-plugin": "^0.5.0",
     "mocha": "^5.2.0",
-    "node-sass": "^4.11.0",
+    "node-sass": "^4.12.0",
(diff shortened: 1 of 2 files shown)


martenson:
@galaxybot test this

martenson:
I don't think we want to update already released code with build dependency changes. I will open a PR against dev. Thanks for bringing this up.

innovate-invent:
@martenson 19.05 fails to build though.

martenson:
Only with newer node though.

dannon:
@innovate-invent Are you able to use the node version shipped with Galaxy?  This exact problem is why we started pinning to, autoinstalling, and using particular node versions.

nsoranzo:
The node version supported in 19.05 is [10.13.0](https://github.com/galaxyproject/galaxy/blob/release_19.05/client/.node_version) .

Issue #8493, opened 2019-08-21
Refactor configuration schema loading


jdavcs:
Currently, default values for configuration properties are specified in `config_schema.yml` files + Python modules (under `config/` for galaxy and `webpapps/` for reports and TS). The values from the `config_schema.yml` files are read in by scripts called from Makefiles (in root + `doc/` directories). The values hardcoded into Python modules are loaded into Configuration objects at startup. Needless to say, this duplication can be eliminated.

Furthermore, in a [recent modification](https://github.com/galaxyproject/galaxy/pull/8475) a property attribute ('reloadable') has been added to the schema. The value of this attribute determines how the property is processed at runtime. Properties marked as 'reloadable', along with their default values are now being read from `webapps/galaxy/schema_config.yml`, whereas all other properties are still assigned default values hardcoded in `config/__init__.py`.

Based on the discussion [here](https://github.com/galaxyproject/galaxy/commit/c5356b298c76c0c65cb288a84dc7caa23a9815bd) (and, to some extent, [here](https://github.com/galaxyproject/galaxy/pull/8475) ), I propose to refactor this, so that `config_schema.yml` is the single source of truth for configuration schema information stored there (default values and more).

If there are no objections, I'd like to take this on.

jdavcs:
The bulk of this work is done: config options (+ their attributes) are loaded from schema. The rest is ongoing refactoring, enhancement, cleanup, etc. 
Keeping this open because there's still much redundancy between the schema and the config processing code.

ic4f:
This can be closed now (see #10172).

PR #8509, opened 2019-08-23
Factor out yaml read/write from config_manage.py

jdavcs:
This is a step towards refactoring config as described in #8493.

Summary of changes:
- Move yaml processing into new utility module under galaxy/util (is this a good location?)
- Factor out Schema classes out of config_manage into schema module
- Add unit test for new schema module (to verify that the schema is loaded correctly)
- Move OrderedLoader out of function scope to accommodate unit test 

Rationale:
The primary responsibility of `config_manage.py` is servicing calls from Makefiles (linting, building sample yaml, building rst, converting from ini, etc.). This included reading a schema definition from a yaml file and loading it into a Schema (or AppSchema) object. 

Due to the need for reading the schema at startup (as per #8493), `config_manage` is now imported into `config/__init__`. 

Given that (1) `config_manage.py` was intended to be used as a script rather than a library module, and (2) loading yaml is only a small fraction of what it does, pulling out this functionality into a utility module and then using it for both use cases seems a reasonable approach.

### DIFF
--- a/lib/galaxy/config/config_manage.py
+++ b/lib/galaxy/config/config_manage.py
@@ -25,8 +25,15 @@ except ImportError:
 if __name__ == '__main__':
     sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir, os.pardir)))
 
+from galaxy.config.schema import AppSchema, Schema
 from galaxy.util import safe_makedirs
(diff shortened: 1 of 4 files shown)


jmchilton:
Solid refactoring, thanks so much @ic4f !

PR #8510, opened 2019-08-24
Fix default job metrics in dev/.

jmchilton:
Broken with the config re-write as part of #921.

The argument for setting based on the sample here and not other places is because the in-Galaxy default is different than the galaxy-job-metrics as a library default (no metrics) I suppose. In that sense, the sample file does contain some logic that doesn't necessarily belong in ``galaxy.job_metrics``.


nsoranzo (approved):
#921 also removed support for searching some config files in the root galaxy directory (e.g. `data_manager_conf.xml`), that we had for back compatibility. Should we mention this in the release notes?

bgruening:
Thanks John!

PR #8511, opened 2019-08-24
Implement workflow invocation reports (alternate syntax)

jmchilton:
This PR is the same as #8030 but with one (substantial) commit to switch the syntax from using markdown-it-container blocks (https://github.com/markdown-it/markdown-it-container) to using decorated code blocks (similar to RMarkdown https://www.rstudio.com/wp-content/uploads/2015/02/rmarkdown-cheatsheet.pdf) to inject Galaxy specific Markdown instructions.

As a very simple example, the old syntax was:

<pre>
::: job_parameters step=qc_step
::: 
</pre>

The new syntax is:

<pre>
```{galaxy_job_parameters step=qc_step}
```
</pre>

On balance the previous syntax was a bit more concise but I prefer the newer RMarkdown-y syntax I think. It seems better specified to me because there are matching braces and we don't have to worry about embedding fences in blocks or blocks in fences, etc... The documents are also closer to real Markdown documents (but I guess I could see that being argued either way). The implementation is clearly cleaner in this RMarkdown-y branch but if we opt for the other syntax we could probably still refactor some of that in this direction anyway.

The components render the same so I'll not duplicate all the screenshots but I will embed the new example report with the new syntax.

```yaml
class: GalaxyWorkflow
name: My Cool Workflow
inputs:
  input_1: data
  image_input: data
  input_list: collection
outputs:
  output_1:
    outputSource: first_cat/out_file1
  output_image:
    outputSource: image_cat/out_file1
steps:
  first_cat:
    tool_id: cat
    in:
      input1: input_1
  image_cat:
    tool_id: cat
    in:
      input1: image_input
  qc_step:
    tool_id: qc_stdout
    state:
      quality: 9
    in:
      input: input_1
report:
  markdown: |
    ## About This Report

    This report is generated from markdown content in the workflow YAML/JSON.
    Workflow invocation reports provide a customizable way for workflow authors
    to summarize the results of a workflow execution. There is a default markdown
    template if a workflow does not define one (right now it just shows the inputs,
    outputs, and workflow - but this will evolve).

    This report is written in "Galaxy Workflow Flavored Markdown". This workflow
    variant of "Galaxy Flavored Markdown" contains workflow-relative references.
    Dataset, collections, and jobs are referenced by step labels, input labels, and output
    labels - rather than by object IDs. The Galaxy invocation report generator plugin
    translates this to "Galaxy Flavored Markdown" at the time the report is viewed and
    that format references actual object IDs (by database ID internally, and encoded ID
    when exported to the client via the API).

    An upshot of translating the workflow markdown to this second neutral format that
    has no concept of the workflow invocation is that client side rendering (and much
    of the backend processing) is completely general and not tied to workflows or
    invocations. The same markdown components could potentially be used to render pages,
    history annotations, describe library folders, etc..

    The next two sections demonstrate the auto generated inputs and outputs sections
    in the default workflow invocation report template.

        ## Workflow Inputs
        ```{galaxy_invocation_inputs}
        ```

        ## Workflow Outputs
        ```{galaxy_invocation_outputs}
        ```

    ## Workflow Inputs
    ```{galaxy_invocation_inputs}
    ```

    ## Workflow Outputs
    ```{galaxy_invocation_outputs}
    ```

    The auto-generated sections could be hand-crafted from workflow markdown also,
    listing out each input and output explicitly. The auto generated sections are merely
    a convenience to avoid needing to that by hand and for supplying a default report
    for workflows that don't define a custom one report.

    ## More Custom Content

    The rest of this report demonstrates more the directives allowed in the report
    markdown. ``invocation_outputs`` and ``invocation_inputs`` are not allow in
    "Galaxy Flavored Markdown" (the non-workflow invocation specific format), but
    the remainder of these are allowed - they would just need to reference object
    IDs instead of inputs, outputs, and steps.

    Once can reference an output and embed a display of it as follows:

        ```{galaxy_history_dataset_display output=output_1}
        ```

    ```{galaxy_history_dataset_display output=output_1}
    ```

    Inputs can be referenced and displayed the same way:

        ```{galaxy_history_dataset_display input=input_1}
        ```

    ```{galaxy_history_dataset_display input=input_1}
    ```

    ---

    Images can be embedded directly into the report as follows:

        ```{galaxy_history_dataset_as_image output=output_image}
        ```

    ```{galaxy_history_dataset_as_image output=output_image}
    ```

    ---

    Dataset peek content can be displayed to quickly provided an embedded
    summary of an input or output:

        ```{galaxy_history_dataset_peek output=output_1}
        ```

    ```{galaxy_history_dataset_peek output=output_1}
    ```

    ---

    Dataset "info" content can be displayed as well:

        ```{galaxy_history_dataset_info input=input_1}
        ```

    ```{galaxy_history_dataset_info input=input_1}
    ```

    ---

    Collections can be displayed:

        ```{galaxy_history_dataset_collection_display input=input_list}
        ```

    ```{galaxy_history_dataset_collection_display input=input_list}
    ```

    ---

    The whole workflow can be embedded to provide some context and display
    annotations and steps.

        ```{galaxy_workflow_display}
        ```

    ```{galaxy_workflow_display}
    ```

    ---

    Job parameters can be summarized:

        ```{galaxy_job_parameters step=qc_step}
        ```

    ```{galaxy_job_parameters step=qc_step}
    ```

    ---

    Job metrics can be summarized as well:

        ```{galaxy_job_metrics step=image_cat}
        ```

    ```{galaxy_job_metrics step=image_cat}
    ```

    ---

    Tool standard out and error are also available for steps.

        ```{galaxy_tool_stdout step=qc_step}
        ```

    ```{galaxy_tool_stdout step=qc_step}
    ```

        ```{galaxy_tool_stderr step=qc_step}
        ```

    ```{galaxy_tool_stderr step=qc_step}
    ```

    ---

    *fin*
```


jmchilton:
No one is making public comments about the syntax for embedding Galaxy objects in Markdown - I think because everyone wants this and wants this soon and is nervous about the conversation devolving. But people I really respect have weighed in privately and have good ideas that should be aired before this is merged/finalized. So I'm going to summarize 6 options here and do a poll. Please vote with the comment type on this issue. The voting is non-binding (unless you are a committer and will vote with your ``-1``). I want to get a final PR out soon so we can branch, so there will be about 18 hours to vote before I make a choice.

----

1. 😕 

<pre>
::: job_metrics step=image_cat
:::
</pre>

Pro: Done, #8030. 
Cons: Is unlike all other Markdown dialects commonly used. Used markdown-it-container to implement "quickly" - but then later decided it is much cleaner in implementation not to use this plugin.

---

2. 😄 


<pre>
```{galaxy_job_metrics step=image_cat}
```
</pre>

Pro: Done, here. Looks vaguely like RMarkdown if you squint.
Con: Maybe doesn't follow the spirit of RMarkdown because the 'code' inside isn't rendered/used.

---

3. 👎 

<pre>
```{g job_metrics step=image_step}
```
</pre>

Pro: Looks vaguely like RMarkdown if you squint - a bit more than the above example since ``g`` proceeds all commands.
Con: Maybe doesn't follow the spirit of RMarkdown because the 'code' inside isn't rendered/used. John doesn't like how the syntax feels more ad-hoc and less like a function call.

---

4. 🚀 

<pre>
```galaxy
{job_metrics step=image_cat}
```
</pre>

Pro: Looks vaguely like the above two examples - maybe doesn't follow the spirit of Markdown because the code isn't displayed in Galaxy though -  however if rendered outside of Galaxy as literal blocks though, what is going to happen here is clear which the above doesn't.

---

5. 🎉 

<pre>
```galaxy
job_metrics(step=image_cat)
```
</pre>

Pro: Looks vaguely like the above example, but if you think about these as code blocks - this looks more like a real language that would be vaguely C/R/Python-ish.

---

6. 👀 

<pre>
 ```galaxy
(job_metrics :step image_cat)
```
</pre>

Pro: Looks vaguely like the above two examples, but if you think about these as code blocks - this looks more like a real language that John would want to implement 😆.

dannon:
@jmchilton Thanks for putting all the thought into this and for laying out these various syntax options.

jmchilton:
My dreams of implementing a lisp dialect somewhere inside of Galaxy will have to be delayed. I'll see if I can implement 🎉-flavored Markdown - thanks all.

martenson:
I believe superseded by https://github.com/galaxyproject/galaxy/pull/8543

PR #8512, opened 2019-08-24
Markdown-based Galaxy Pages.

jmchilton:
## Overview

Utilize the architecture for different variants of Galaxy Markdown (internal vs. external IDs, Vue components, etc..) setup in #8543 to implement Markdown based pages.

- Page revisions now have a content type - can be HTML (old style) or newer Markdown. New API tests to test validation, creation, etc..
- New button to view a page directly from the editor (for both versions) - warns before navigates away.
- Initial shot at a Markdown editor toolbar that allows embedding Galaxy directives for selected objects - workflows, datasets, collections.
   - Extracted a ``SelectionDialog`` Vue component out of ``DataDialog`` and used to create workflow and collection selection variants. With new Vue tests.
- Add help button to Markdown editor toolbar for pages.
- Swap page saving for both editors to happen via the API - which is much more well tested and allows us to delete a few lines of older controller code.

Builds heavily on https://github.com/galaxyproject/galaxy/pull/8047 (Toward a more modern frontend and backend for Pages). 

## Gallery

Creation page with new format option, warning about HTML editor having known bugs and becoming read-only in the future. (So we can axe the Ace editor).

Screen Shot 2019-12-08 at 10 09 17 PM

Validation still could use a lot of work, but it does include feedback to the end user with specific problems (including line numbers)

Screen Shot 2019-12-08 at 9 58 00 PM

Help dialog (mirroring that for the invocation report editor) available from the Markdown toolbar

Screen Shot 2019-12-08 at 9 58 24 PM

New workflow selection dialog:

Screen Shot 2019-12-08 at 9 58 36 PM

New collection selection dialog:

Screen Shot 2019-12-08 at 9 58 47 PM


jmchilton:
Took this out of WIP because really the roughest part I think is the actual components used in rendering the Markdown - but these are just as rough in invocation reports which are in 19.09. More eyeballs and more use of these components I think could lead faster improvements.

This also introduces the collection selection dialog which I think is important first step for the enhanced collection selection laid out in the roadmap for #9030 / http://bit.ly/simplified-workflow-ui.

client/galaxy/scripts/utils/data.js
```diff
@@ -41,15 +43,44 @@ export function dialog(callback, options = {}) {
...
+ */
+export function workflowDialog(callback, options = {}) {
+    Object.assign(options, {
+        callback: callback
+    });
+    _mountSelectionDialog(WorkflowDialog, options);
+}
+
+/**
+ * Opens a modal dialog for workflow selection
```
mvdbeek:
```suggestion
 * Opens a modal dialog for dataset collection selection
```

mvdbeek (commented):
There's a couple of things I noticed, I don't think we need to address all before merging:

- In the markdown editor there are no tooltips for the icons (dataset, collection, workflow). I think that would help first time users and should be part of the PR.

- Embedding a collection looks a bit weird compared to embedding Datasets or Workflows:
Screenshot 2019-12-16 at 12 50 10
Not sure how hard it'd be to address this, and given it's new functionality I think it'd be OK to merge as is.

- All selections are for the currently active history, which I think is fine for now.
- We should be able to include entire histories, which again could be for a followup PR.

- Old-style pages lose the masthead.
- The old-style editor doesn't let you embed Histories with dataset collections. This seems to be a problem with `getGalaxyInstance`. If I remove the Galaxy.debug statements from controlled-fetch-collections as in https://github.com/mvdbeek/galaxy/commit/153ba0ce5151aaf120af9367316798ab4896bf46 the history embeds fine. I haven't checked if this was a pre-existing bug.

- A preview in the editor context would be cool. but again I don't think it's necessary for merging.

- The migration needs to be 164 after merging #8972

jmchilton:
Thanks for the detailed review and for the fixes. I've moved the things I think can be addressed in the future to #9118. My per-bullet responses below:

> In the markdown editor there are no tooltips for the icons (dataset, collection, workflow). I think that would help first time users and should be part of the PR.

Fixed - here is a screenshot. Also brought in the "Embed Dataset as Image" button from #8893 since I was working on the tool bar to fix this.

Screen Shot 2019-12-16 at 6 48 54 AM

> Embedding a collection looks a bit weird compared to embedding Datasets or Workflows:

"A bit" is an understatement. This is an existing issue with Invocation Reports though and is hard to deal with until we have a clear picture what widgets we should be done (which probably requires #8441 also?). I've added your screenshot to #9112.

> All selections are for the currently active history, which I think is fine for now.

I agree but we don't have Vue components to do this yet - I've added this point to the checklist in #9118. Since we are re-using components - this is already tracked and would be addressed by #9110 / #9114.

> We should be able to include entire histories, which again could be for a followup PR.

I could potentially be convinced but I don't agree initially - we have plenty of empirical evidence that no one used pages and so we aren't beholden to replicating all of their features.

If you really think it should be re-implemented - feel free to add it to #9118 / #9119.

> Old-style pages lose the masthead.

Not in my hands? How do I replicate this? Just create an old page and click on it in the grid? I've tried in dev and on my branch - for published and unpublished pages.

> The old-style editor doesn't let you embed Histories with dataset collections. This seems to be a problem with getGalaxyInstance. If I remove the Galaxy.debug statements from controlled-fetch-collections as in mvdbeek@153ba0c the history embeds fine. I haven't checked if this was a pre-existing bug.

I've tested this against dev and it was an issue there also - pages aren't used so I'm actually not sure this ever worked. If it broke recently though - perhaps in #7117 / #7005?

I've merged your commit into the branch though, thanks!

> The migration needs to be 164 after merging #8972

Rebased with this fix.

mvdbeek:
> Not in my hands? How do I replicate this? Just create an old page and click on it in the grid? I've tried in dev and on my branch - for published and unpublished pages.

Checked your code out again just now and it was fine now, I'm sure something went wrong on my side.



> Also brought in the "Embed Dataset as Image" button from #8893 since I was working on the tool bar to fix this.

Can't get this to work unfortunately, tried with a couple of datasets and always get this:
Screenshot 2019-12-17 at 20 39 24

mvdbeek:
Oh, got it, it needs to be an image. That is awesome, I can really see myself doing some write up in this!
Not sure if we we could do something to make it more clear that this should be an image dataset.

It is in the help though, so I think this is fine.

mvdbeek (approved):

PR #8521, opened 2019-08-27
Fix killing containers error-ing out jobs.

jmchilton:
We started properly killed docker jobs when jobs are deleted in #7494, but we aren't suppressing error messages that can cause normal docker jobs to fail. This hack fixes that.

### DIFF
--- a/lib/galaxy/tool_util/deps/container_classes.py
+++ b/lib/galaxy/tool_util/deps/container_classes.py
@@ -282,9 +282,15 @@ class DockerContainer(Container, HasDockerLikeVolumes):
             **docker_host_props
         )
         kill_command = docker_util.build_docker_simple_command("kill", container_name=self.container_name, **docker_host_props)
+        # Suppress standard error below in the kill command because it can cause jobs that otherwise would work
+        # to fail. Likely, in these cases the container has been stopped normally and so cannot be stopped again.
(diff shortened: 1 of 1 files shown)


lib/galaxy/tool_util/deps/container_classes.py
```diff
@@ -282,9 +282,14 @@ def containerize_command(self, command):
             **docker_host_props
         )
         kill_command = docker_util.build_docker_simple_command("kill", container_name=self.container_name, **docker_host_props)
+        # Suppress standard error below in the kill command because it can cause jobs that work other to fail,
+        # because the container stopped normally I believe. A less hacky approach might be to check if the container
```
martenson: > it can cause jobs that work other to fail, because the container stopped normally I believe

Please rephrase this, it is not clear enough, at least to me.

jmchilton: Meant otherwise not other... but I've redone it all to be a bit more clear. Let me know if it still doesn't make sense.

martenson (approved):

lib/galaxy/tool_util/deps/container_classes.py
martenson: Thanks, much clearer.

PR #8526, opened 2019-08-28
Break config/ dependency on config_manage script

jdavcs:
This is another step towards refactoring config as described in #8493.

Summary of changes:
- Change `__init__` method signature in `schema.py` to break its dependency on object defined in `config_manage.py`
- Adjust relevant unit test + `config_manage.py`
- Reverse dependency: `config_manage.py` now imports the schema file path from `config/__init__.py`

Rationale:
An AppSchema object requires (1) the path to the schema file, and (2) the app name. Both were passed to its constructor as members of an object defined in `config_manage.py`. This made the AppSchema object (and, consequently, any modules importing it) dependent on `config_manage.py`, which is problematic (see #8509 for rationale). Also, schema_path and app_name were the only properties of the passed object used by AppSchema; all other properties were specific to `config_manage.py`. This commit breaks this dependency, and will allow to read in a schema file at startup without using `config_manage.py`.

Concerns:
1. I am not sure that a constant in `__init__.py` is the best place for specifying the path to the schema file. 
2. Now `config_manage.py` imports the schema path from `__init__.py`. I think a better solution would be for both modules to access this value from another location, but creating a module just for that seems overkill.
In both cases, I think the suggested approach is good enough for now, and can be edited as needed.

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -24,6 +24,7 @@ import yaml
 from six import string_types
 from six.moves import configparser
 
+from galaxy.config.schema import AppSchema
 from galaxy.containers import parse_containers_config
 from galaxy.exceptions import ConfigurationError
 from galaxy.model import mapping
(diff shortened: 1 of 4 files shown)


jmchilton:
Awesome - thanks!

PR #4262, opened 2017-06-28
Library tags

bwlang:
This adds support for #tags in library datasets, auto-tagging on input (based on filenames) and copying tags to history datasets on import.

PR #8530, opened 2019-08-28
Add API to tag library datasets

abretaud:
Since #4262, when importing some files to a data library, filename can be added as a tag automatically.
With this PR, I extended the api to allow adding/updating any tags to library datasets.
I also modified the data lib templates to display tags in the UI. Though they're not displayed with pretty colors, and I didn't make the changes to the UI to add/remove tags like history datasets. I'm mostly interested in the API in fact.
I'm also preparing a PR for bioblend for this
Any comments welcome!

### DIFF
--- a/lib/galaxy/managers/library_datasets.py
+++ b/lib/galaxy/managers/library_datasets.py
@@ -238,7 +238,7 @@ class LibraryDatasetsManager(datasets.DatasetAssociationManager):
         rval['update_time'] = ldda.update_time.strftime("%Y-%m-%d %I:%M %p")
         rval['can_user_modify'] = trans.user_is_admin or trans.app.security_agent.can_modify_library_item(current_user_roles, ld)
         rval['is_unrestricted'] = trans.app.security_agent.dataset_is_public(ldda.dataset)
-        rval['tags'] = self.tag_handler.get_tags_str(ldda.tags).split(', ')
+        rval['tags'] = self.tag_handler.get_tags_str(ldda.tags)
 
         #  Manage dataset permission is always attached to the dataset itself, not the the ld or ldda to maintain consistency
         rval['can_user_manage'] = trans.user_is_admin or trans.app.security_agent.can_manage_dataset(current_user_roles, ldda.dataset)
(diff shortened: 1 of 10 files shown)


abretaud:
OMG now with colors!

tag_1

tag_2

lib/galaxy/managers/library_datasets.py
```diff
@@ -238,7 +238,7 @@ def serialize(self, trans, ld):
         rval['update_time'] = ldda.update_time.strftime("%Y-%m-%d %I:%M %p")
         rval['can_user_modify'] = trans.user_is_admin or trans.app.security_agent.can_modify_library_item(current_user_roles, ld)
         rval['is_unrestricted'] = trans.app.security_agent.dataset_is_public(ldda.dataset)
-        rval['tags'] = self.tag_handler.get_tags_str(ldda.tags)
+        rval['tags'] = self.tag_handler.get_tags_str(ldda.tags).split(', ')
```
martenson: This is effectively changing the API response format, I am not sure how used this endpoint is, but we try not to break it this easily.

abretaud: Thanks, fixed :)
Do you think I should return a str too in https://github.com/galaxyproject/galaxy/pull/8530/files#diff-4a1789d4affda47540d7b94cac91340eR117? a list looks more natural to me, but I can change if we want the two response format to be coherent

martenson: I think so, consistency is more important here I believe. Thanks for doing that!

martenson:
```
Uncaught TypeError: this.model.get(...).split is not a function
library-folderrow-view.js?7537:82 
```

abretaud:
strange, I didn't have this error, but added a check to be sure

abretaud:
@martenson in fact I think the check in a3922cf are needed otherwise it tries to split already splitted list when sorting on another column or modifying a dataset = when rerendering
I have another commit almost ready for little problems I found while writing bioblend unit tests

martenson:
@abretaud perfect, that makes sense, I was not able to reproduce the error I saw before so I thought it is superfluous, but sorting on other columns was exactly the trigger

abretaud:
Ok, I think it's ready now, there are some tests in https://github.com/galaxyproject/bioblend/pull/296 that should pass once this one is merged
Selenium test python 3 is still red, but I don't think it comes from this pr
I'm taking a break for the next 3 weeks, feel free to make any changes!

jmchilton (approved):

martenson:
Thank you @abretaud !

PR #8531, opened 2019-08-28
Add data types to docs.galaxyproject.org

selten:


doc/source/dev/data_types.md
```diff
@@ -0,0 +1,311 @@
+
+# Data Types
+
+## Adding a New Data Type
+
+**Note:** This is the 'old' method for doing this. It is strongly suggested to use the Galaxy ToolShed method for adding data types.
```
nsoranzo: That's not the case, we strongly encourage to add datatypes to the galaxy repo, not the Tool Shed.

martenson:
Very nice docs! Thanks @selten !

doc/source/dev/data_types.md
selten: Sorry for this, I meant 'custom' datatypes that are not used outside of an organisation with this comment. Of course for datatypes that are used by certain tools, these should definitely be added to the Galaxy repo.

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,311 @@
...
+We'll add the new data type to the `<registration>` tag section of the `datatypes_conf.xml` file. Sample `<datatype>` tag attributes in this section are:
+
+```xml
+<datatype extension="ab1" type="galaxy.datatypes.images:Ab1" mimetype="application/octet-stream" display_in_upload="true"/>
+```
+
+where
+
+-  ``extension`` - the data type's Dataset file extension (e.g., `ab1`, `bed`, `gff`, `qual`, etc.) `type` - the path to the class for that data type.
+-  `mimetype` - if present (it's optional), the data type's mime type `display_in_upload` - if present ( it's optional and defaults to False ), the associated file extension will be displayed in the "File Format" select list in the "Upload File from your computer" tool in the "Get Data" tool section of the tool panel.
```
nsoranzo: Missing a line break here.

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,311 @@
...
+from galaxy.datatypes.sniff import *
+from cgi import escape
+import urllib
+from bx.intervals.io import *
+from galaxy.datatypes import metadata
+from galaxy.datatypes.metadata import MetadataElement
+from galaxy.datatypes.tabular import Tabular
+
+
+class Foobar( Tabular ):
```
nsoranzo:
```suggestion
class Foobar(Tabular):
```

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,311 @@
...
+    """Tab delimited data in foo format"""
+    file_ext = "foo"
+
+    MetadataElement( name="columns", default=3, desc="Number of columns", readonly=True )
```
nsoranzo:
```suggestion
    MetadataElement(name="columns", default=3, desc="Number of columns", readonly=True)
```

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,311 @@
...
...xy.datatypes.foobar.py`, that will contain the `Foobar` class (in this example we could simply add the `Foobar` class to the `galaxy.datatypes.tabular.py` code file). Keep in mind that your new data type class should be placed in a file that is appropriate (based on it's superclass), and that the file will need to be imported by `lib/galaxy/datatypes/registry.py`. You will need to include a `file_ext` attribute to your class and create any necessary functions to override the functions in your new data type's superclass (in our example, the `galaxy.datatypes.tabular.Tabular` class). In our example below, we have set our class's `file_ext` attribute to "foo" and we have overridden the `__init__()`, `init_meta()` and `sniff()` functions. It is important to override functions (especially the meta data and sniff functions) if the attributes of your new class differ from those of it's superclass. Note: sniff functions are not required to be included in new data type classes, but if the sniff function is missing, Galaxy will end up associating the default data type and file extension (`Text` and `txt` in our example) with the file. For binary files, the default would be `Data` and `data`.
```
nsoranzo: In the `datatypes_conf.xml` example above you use `galaxy.datatypes.tabular:Foobar`, so you need to also add `Foobar` to `lib/galaxy/datatypes/tabular.py`, not a new file.

```suggestion
We'll now add the `Foobar` class to `lib/galaxy/datatypes/tabular.py`. Keep in mind that your new data type class should be placed in a file that is appropriate (based on its superclass), and that the file will need to be imported by `lib/galaxy/datatypes/registry.py`. You will need to include a `file_ext` attribute to your class and create any necessary functions to override the functions in your new data type's superclass (in our example, the `galaxy.datatypes.tabular.Tabular` class).
In our example below, we have set our class's `file_ext` attribute to "foo" and we have overridden the `__init__()`, `init_meta()` and `sniff()` functions. It is important to override functions (especially the meta data and sniff functions) if the attributes of your new class differ from those of its superclass. Note: sniff functions are not required to be included in new data type classes, but if the sniff function is missing, Galaxy will call the superclass method.
```

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,311 @@
...
+import logging, os, sys, time, sets, tempfile, shutil
+import data
+from galaxy import util
+from galaxy.datatypes.sniff import *
+from cgi import escape
+import urllib
+from bx.intervals.io import *
+from galaxy.datatypes import metadata
+from galaxy.datatypes.metadata import MetadataElement
+from galaxy.datatypes.tabular import Tabular
```
nsoranzo, line 125: I'd remove all imports for the reason above.

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,311 @@
...
+    MetadataElement( name="columns", default=3, desc="Number of columns", readonly=True )
+
+
+    def __init__ (self, **kwd):
+        """Initialize foobar datatype"""
+        Tabular. __init__ (self, **kwd)
+        self.do_something_else()
+
+
+    def init_meta( self, dataset, copy_from=None ):
```
nsoranzo:
```suggestion
    def init_meta(self, dataset, copy_from=None):
```

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,311 @@
...
+        Tabular.init_meta( self, dataset, copy_from=copy_from )
```
nsoranzo:
```suggestion
        Tabular.init_meta(self, dataset, copy_from=copy_from)
```

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,311 @@
...
+        if elems_len == 8:
+            try:
+                map( int, [hdr[6], hdr[7]] )
```
nsoranzo:
```suggestion
                map(int, [hdr[6], hdr[7]])
```

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,311 @@
...
+                proceed = True
+            except:
+                pass
+
+
+    def sniff( self, filename ):
```
nsoranzo:
```suggestion
    def sniff(self, filename):
```

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,311 @@
...
+        headers = get_headers( filename, '\t' )
```
nsoranzo:
```suggestion
        headers = get_headers(filename, '\t')   
```

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,311 @@
...
+                pass
+
+
+    def sniff( self, filename ):
+        headers = get_headers( filename, '\t' )   
+        try:
+            if len(headers) < 2:
+                return False
+            for hdr in headers:
+                if len( hdr ) > 1 and hdr[0] and not hdr[0].startswith( '#' ):
```
nsoranzo:
```suggestion
                if len(hdr) > 1 and hdr[0] and not hdr[0].startswith('#'):
```

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,311 @@
...
+                    if len(hdr) != 8:
+                        return False
+                    try:
+                        map( int, [hdr[6], hdr[7]] )
```
nsoranzo:
```suggestion
                        map(int, [hdr[6], hdr[7]])
```

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,311 @@
...
+### Defining Basic Composite Datatypes
+
+The example below defines a basic composite datatype which is composed of 2 component files along with a tool-written or user-uploaded primary file. In this example the primary file is a report summarizing the results. The two component files (`results.txt`, `results.dat`) contain the results; `results.txt` is a text file and is handled as such during upload whereas `results.dat` is flagged as binary, allowing a binary file to be uploaded for that component. During upload, 3 sets of file upload/paste interfaces are provided to the user. A file must be provided for the primary (index) file as well as `results.txt`; `results.dat` is flagged as optional so may be left blank.
+```python
+class BasicComposite(Text):      
+
+    composite_type = 'basic'
+
+
+    def __init__ ( self, **kwd ):
```
nsoranzo:
```suggestion
    def __init__ (self, **kwd):
```

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,311 @@
...
+        Text. __init__ ( self, **kwd )
```
nsoranzo:
```suggestion
        Text. __init__ (self, **kwd)
```

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,311 @@
...
+        self.add_composite_file( 'results.txt' )
```
nsoranzo:
```suggestion
        self.add_composite_file('results.txt')
```

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,311 @@
...
+        self.add_composite_file( 'results.dat', is_binary = True, optional = True )
```
nsoranzo:
```suggestion
        self.add_composite_file('results.dat', is_binary = True, optional = True)
```

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,311 @@
...
+```
+
+These files can be specified on the command line in the following fashion:
+
+```xml
+<command>someTool.sh $input1 ${os.path.join( input1.extra_files_path, 'results.txt' )}
```
nsoranzo:
```suggestion
<command>someTool.sh '$input1' '${os.path.join(input1.extra_files_path, 'results.txt')}'
```

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,311 @@
...
+        Text. __init__ ( self, **kwd )
+        self.add_composite_file( 'results.txt' )
+        self.add_composite_file( 'results.dat', is_binary = True, optional = True )
+```
+
+These files can be specified on the command line in the following fashion:
+
+```xml
+<command>someTool.sh $input1 ${os.path.join( input1.extra_files_path, 'results.txt' )} 
+${os.path.join( input1.extra_files_path, 'results.dat' )} $output1</command>
```
nsoranzo:
```suggestion
'${os.path.join(input1.extra_files_path, 'results.dat')}' '$output1'</command>
```

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,311 @@
...
+        return "\n".join( rval )
+```
+
+The file specified as `%s.ped` is found at `os.path.join( input1.extra_files_path, '%s.ped' % input1.metadata.base_name )`.
+
+It should be noted that changing the datatype of datasets which use this substitution method will cause an error if the metadata parameter 'base_name' does not exist in a datatype that the dataset is set to. This is because the value within 'base_name' will be lost -- if the datatype is set back to the original datatype, the default metadata value will be used and the filenames might not match the basename.
+
+To prevent this from occurring, set `allow_datatype_change` to `False`. The dataset's datatype will not be able to be changed.
+
+### Future Directions
```
nsoranzo: I'd remove this subsection.

nsoranzo:
I see this is derived from:
- https://galaxyproject.org/admin/datatypes/adding-datatypes/
- https://galaxyproject.org/admin/datatypes/composite-datatypes/
- https://galaxyproject.org/toolshed/datatypes-features/

which should become redirects when this is merged.

@selten Can you check if there's anything from https://galaxyproject.org/admin/datatypes/adding-complete-datatypes/ that could be useful here?

doc/source/dev/data_types.md
jmchilton: I don't think these datatypes should be added to a custom tool shed at all either. The tool shed is never the right answer for data types - it has no way of handling versioning at all. If you don't want to or can't use galaxyproject/galaxy I think forking Galaxy is a better method than using the tool shed.

martenson: Or just maintaining custom datatype commits and rebase them on new releases. Shouldn't conflict often.

selten: I'm removing it from this.

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,312 @@
...
+
+This approach was implemented to be backwards compatible with existing Rgenetics datasets/tools (implemented before composite datatypes existed). Different approaches should be considered. Suggestions and comments are welcome.
+
+## Galaxy Tool Shed - Data Types
```
nsoranzo, line 496: Given the discussion above, should all the following just be removed? It can stay on the Hub.

selten: Given that it is a feature, I'd opt to keep it in, I'm adding the 'completely new data type' part as well.

jmchilton: +1 to just removing it - I cannot stress how much I discourage using these. If we want a section - I'd say something like just "These are deprecated and shouldn't be used." and not include a technical discussion of how to write/use them.

selten: I added that section that they are deprecated and shouldn't be used.

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,607 @@
...
+
+import pkg_resources
+pkg_resources.require( "bx-python" )
+
+import logging, os, sys, time, sets, tempfile, shutil
+import data
+from galaxy import util
+from galaxy.datatypes.sniff import *
+from cgi import escape
+import urllib
```
nsoranzo: Can you replace all these imports with:

```python
from galaxy.datatypes.sniff import get_headers
from galaxy.datatypes.tabular import Tabular
```

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,607 @@
...
+and start up your server. Were you watching the logs 
+carefully? No? Wondering why your module isn't showing 
+up in the upload tool? Well, if you dig through your 
+logs you'll see this message:
+
+```
+galaxy.datatypes.registry ERROR 2014-07-17 12:43:23,939 Error importing datatype module galaxy.datatypes.csequence: 'module' object has no attribute 'csequence'
+Traceback (most recent call last):
+  File "/home/hxr/work/galaxy-central/lib/galaxy/datatypes/registry.py", line 208, in load_datatypes
+    module = getattr( module, mod )
```
nsoranzo:
```suggestion
    module = getattr(module, mod)
```

doc/source/dev/data_types.md
```diff
@@ -0,0 +1,607 @@
...
+import csequence
+```
+
+Once you've done this, your server will start up and 
+the datatype will be available. Please note that this 
+problem can be avoided by using the Tool Shed to store 
+your datatypes. There, this issue will be avoided as 
+Galaxy handles imports from Tool Shed installed 
+datatypes differently than from locally installed 
+datatypes.
```
nsoranzo: Is this still the case? I think all that's needed any more since commit ce54045b20996584f1b83a1966d4f526f3f4fdf2 (from 2014).

PR #8533, opened 2019-08-28
[19.05] Hack to fix Pulsar for containerized jobs.

jmchilton:
This isn't great, I have a better fix but it requires an update to Pulsar and pulsar-galaxy-lib. This fix should work with older versions of Pulsar and the client library I think though.


bgruening:
This seems to work nicely at usegalaxy.eu now. The tool_script is transferred and executed. This part of the puzzle is solved, thanks a lot @jmchilton!

bgruening (approved):

PR #8538, opened 2019-08-28
Improve and refactor process group handling in the local job runners

nsoranzo:
Fix traceback seen in

https://jenkins.galaxyproject.org/job/docker-api/16801/testReport/junit/test.api.test_datasets/DatasetsApiTestCase/test_show/

Continues work started by @jmchilton in https://github.com/galaxyproject/galaxy/pull/8467 .

### DIFF
--- a/lib/galaxy/jobs/runners/local.py
+++ b/lib/galaxy/jobs/runners/local.py
@@ -2,10 +2,8 @@
 Job runner plugin for executing jobs on the local system via the command line.
 """
 import datetime
-import errno
 import logging
 import os
-import signal
 import subprocess
 import tempfile
 import threading
@@ -16,9 +14,13 @@ from galaxy.job_execution.output_collect import default_exit_code_file
 from galaxy.util import (
     asbool,
 )
-from ..runners import (
+from . import (
     BaseJobRunner,
     JobState
+)
+from .util.process_groups import (
+    check_pg,
+    kill_pg
 )
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


lib/galaxy/jobs/runners/local.py
```diff
@@ -116,11 +118,9 @@ def queue_job(self, job_wrapper):
                 self._handle_container(job_wrapper, proc)
 
                 terminated = self.__poll_if_needed(proc, job_wrapper, job_id)
+                proc.wait()  # reap
```
natefoo, line 121: Good catch, we were leaving behind a lot of unreaped processes before, I guess?

lib/galaxy/jobs/runners/local.py
```diff
@@ -116,11 +118,9 @@ def queue_job(self, job_wrapper):
...
                 if terminated:
                     return
```
natefoo, line 123: Should we else-kill-PG here?

natefoo (approved):
Very nice, thanks @nsoranzo!

lib/galaxy/jobs/runners/local.py
nsoranzo: Actually no, before `proc.wait()` was called inside `self._terminate(proc)` by `self.__poll_if_needed()` before returning `True`.
Now `kill_pg()` receives `pgid` instead of `proc`, so reap needs to be done here.

natefoo (approved):

PR #8542, opened 2019-08-29
update repo issue handling 

martenson:
* make status/review the default state of a PR
* update the roadmap section to reflect reality

closes https://github.com/galaxyproject/galaxy/issues/5713

PR #8545, opened 2019-08-29
Remove duplicate logic for 'is_admin' checks

jdavcs:
This removes duplicate logic for checking if a user is an admin user. Splitting a string of user emails and checking if a given email is in that list should be located in one place only. (same goes for related code, like "`user and user.email in [a list of emails]`")

- Also some minor syntax cleanup/simplification
- And mocking the mock

Done as part of #8493.

### DIFF
--- a/lib/galaxy/managers/users.py
+++ b/lib/galaxy/managers/users.py
@@ -234,25 +234,18 @@ class UserManager(base.ModelManager, deletable.PurgableManagerMixin):
         Do not pass trans to simply check if an existing user object is an admin user,
         pass trans when checking permissions.
         """
-        admin_emails = self._admin_emails()
         if user is None:
             # Anonymous session or master_api_key used, if master_api_key is detected
             # return True.
-            rval = bool(trans and trans.user_is_admin)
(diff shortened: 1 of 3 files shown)


lib/galaxy/managers/users.py
```diff
@@ -234,25 +234,18 @@ def is_admin(self, user, trans=None):
         Do not pass trans to simply check if an existing user object is an admin user,
         pass trans when checking permissions.
         """
-        admin_emails = self._admin_emails()
         if user is None:
```
jdavcs, line 237: better: `if not user:`

jdavcs:
get_code_and_test are not among the checks. No idea why. They run (all passed) on my circleci account.

jdavcs:
@martenson I thought status/review is what we use when a pr is ready for review?

martenson:
@ic4f yeah, but that was yesterday 🤣 https://github.com/galaxyproject/galaxy/pull/8542

jdavcs:
@martenson got it! missed it. Thanks!

Issue #6493, opened 2018-07-12
Restore a deleted workflow


bernt-matthias:
potentially stupid question: is there a way to restore a deleted workflow. In Galaxy 18.05 and dev I could not find a way that does not involve editing the data base.

mvdbeek:
They are just being marked as deleted. Undeleting is definitely something I would like to expose in the UI and API.

natefoo:
A bit more info for admins: the table in question is `stored_workflow`. Once you find the correct workflow (e.g. with the help of the `user_id` column), update `deleted = false`.

This would be great to have in the UI/API.

assuntad23:
Closed with #14794 & #14816

PR #8551, opened 2019-08-30
Implementation of the workflow list as vue component

guerler:
This PR reimplements the current backbone-based workflow list as vue component. It also extends the workflows api by adding `annotations` to its response which allows us to display the workflow description in the list. The `annotations` are currently added by default, however we should consider to make their availability optional if there are performance concerns.

Screen Shot 2019-09-17 at 11 44 02 AM


hexylena:
0. gosh this looks amazing.
1. Since nametags don't exist in workflow context, is there any interest to use the coloured tag component there for all tags? Could make scanning those easier.
2. Any desire to address https://github.com/galaxyproject/galaxy/issues/6493 in this PR? (but no pressure)

guerler:
@galaxybot test

guerler:
@galaxybot test this

jmchilton:
Ready for review? Remove the WIP tag?

guerler:
Yes, it is ready for review. Thank you.

client/galaxy/scripts/components/Workflow/services.js
```diff
@@ -0,0 +1,114 @@
+import axios from "axios";
+import { getGalaxyInstance } from "app";
+
+/** Workflow data request helper **/
+export class Services {
```
dannon, line 5: Most of this should probably be pushed into the central store w/ vuex?

client/galaxy/scripts/components/Workflow/services.js
```diff
@@ -0,0 +1,114 @@
...
+                    resolve(response.data);
+                })
+                .catch(e => {
+                    reject(this._errorMessage(e));
+                });
+        });
+    }
+
+    deleteWorkflow(id) {
+        return new Promise((resolve, reject) => {
```
dannon: Axios returns a promise that we shouldn't have to wrap again, I think.  (might be cleaner yet to use async/await if we're going to have this separate services model here)

test/selenium_tests/test_workflow_run.py
```diff
@@ -199,7 +199,7 @@ def workflow_run_specify_inputs(self, inputs):
     def workflow_run_with_name(self, name):
         self.workflow_index_open()
         self.workflow_index_search_for(name)
-        self.workflow_index_click_option("Run")
+        self.workflow_click_option(".fa-play")
```
dannon: I'd recommend against tying tests to particular fontawesome classes; if this is a significant element it should have another identifier we can use.

client/galaxy/scripts/components/Workflow/WorkflowDropdown.vue
```diff
@@ -0,0 +1,161 @@
...
+    </div>
+</template>
+<script>
+import { getAppRoot } from "onload/loadConfig";
+import { Services } from "./services.js";
+export default {
+    props: ["workflow"],
+    data() {
+        return {
+            ownerOperations: [
```
dannon: It seems like it might be cleaner/clearer to just define these operations in the component above.  Then event bindings wouldn't have to go through the string/switch statement below, and logic would be a lot simpler.

test/selenium_tests/test_workflow_management.py
```diff
@@ -83,27 +83,6 @@ def test_index_search(self):
         self.workflow_index_search_for("searchforthis")
         self._assert_showing_n_workflows(1)
 
-    @selenium_test
```
jmchilton, line 86: 😢

dannon:
I'm trying to get the headers for these bootstrap tables displaying a little better, and I'm going to add the published column back here (w/ test) -- though with a simple checkmark instead of "Yes/No", etc.  That's going to be in a separate followup PR though.

Thanks Sam

PR #8552, opened 2019-08-31
make calculate_or_set_disk_usage support sqlite

qiagu:
It seems sqlite uses 0 instead of false.

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -523,7 +523,6 @@ class User(Dictifiable, RepresentById):
         Utility to calculate (returning a value) or just set the disk usage
         (returning None / applying immediately)
         """
-        sql_false = '0' if 'sqlite' in sa_session.bind.dialect.name else 'false'
 
         ctes = """
             WITH per_user_histories AS
@@ -531,15 +530,15 @@ class User(Dictifiable, RepresentById):
                 SELECT history.id as id
                 FROM history
                 WHERE history.user_id = :id
-                    AND history.purged = %s
+                    AND history.purged = '0'
             ),
             per_hist_hdas AS (
                 SELECT DISTINCT history_dataset_association.dataset_id as id
                 FROM history_dataset_association
-                WHERE history_dataset_association.purged = %s
+                WHERE history_dataset_association.purged = '0'
                     AND history_dataset_association.history_id in (SELECT id from per_user_histories)
             )
-        """ % (sql_false, sql_false)
+        """
 
         sql_calc = """
             SELECT sum(coalesce(dataset.total_size, coalesce(dataset.file_size, 0)))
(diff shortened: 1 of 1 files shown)


qiagu:
@erasche Could you please have a look at this simple PR at your convenience? Any other reviewer is also welcome.

martenson:
So what happens if this method is run on sqlite before this PR?

qiagu:
@martenson Report errors actually. Message like `missing: false`.

martenson:
Can we switch this to be `0` without having special cases for some platforms? It seems postgres supports 0. https://www.postgresql.org/docs/8.1/datatype-boolean.html

qiagu:
Good call! We can try.

qiagu:
'0' works for both sqlite and postgresql in my test. Cool.

hexylena (approved):
looks good to me.

mvdbeek:
This leads to 100% CPU load when logging in a user a logged out user on sqlite. @qiagu what error did you see initially ?

mvdbeek:
and weirdly enough also seems to hang on postgresql, though I'm not 100% sure it's this PR that causes it.

qiagu:
@mvdbeek I have a test Galaxy built on sqlite. Failure occurred when calling the the function. I notice longer time is needed in user registration when running selenium test on my other PR. Maybe it's related to the high CPU.

hexylena:
Ugh, ok, will look at this first thing tomorrow.

hexylena:
Ok, I've tried to reproduce under sqlite + postgres and cannot reproduce.

I'm on 8ac4dc65ab316763d01975bd1d1b10ca90ee1481 but both logging in/out, and executing the update script both execute nearly instantly.

mvdbeek:
Thanks for checking @erasche, I hope it's just some weird config on my end. It's also weird that this is the logging in, not the logging out which triggers this.

hexylena:
yeah that does feel odd!

here are my deltas from standard config, in case that helps your debugging.

```yml
galaxy:
  container_image_cache_path: /home/hxr/arbeit/galaxy/containers
  enable_beta_mulled_containers: true
  database_connection: 'postgresql://postgres:postgres@localhost/postgres?sslmode=disable'
  conda_auto_install: true
  conda_auto_init: true
  use_cached_dependency_manager: true
  watch_tools: true
  watch_job_rules: true
  enable_beta_mulled_containers: true
  containers_resolvers_config_file: /home/hxr/arbeit/galaxy/config/container_resolvers_conf.xml
  tool_data_table_config_path: /cvmfs/data.galaxyproject.org/byhand/location/tool_data_table_conf.xml,/cvmfs/data.galaxyproject.org/managed/location/tool_data_table_conf.xml
  interactive_environment_plugins_directory: config/plugins/interactive_environments
  user_preferences_extra_conf_path: config/conf.yml
  galaxy_infrastructure_url: 'http://localhost'
  dynamic_proxy_manage: false
  dynamic_proxy_external_proxy: true
  sanitize_all_html: false
  allow_user_impersonation: true
  expose_dataset_path: true
```

martenson:
@erasche `enable_beta_mulled_containers` is duplicated

hexylena:
Looks like I really wanted it on. Thanks for the catch.

martenson:
@erasche yeah, but it would make you sad when you wanted it off :)

PR #8559, opened 2019-09-02
create dynamic options from_dataset for multiple inputs

bernt-matthias:
If `from_dataset` refers to a input data set with `multiple="true"` no options were generated so far. 
With this changes options are generated from all inputs selected in the inputs. 

Though .. it does not work for collections yet .. suggestions welcome. 

- One would need to iterate over either a HDA, a list of HDAs, a HDCA (or maybe a list of HDCAs and I would like to do this without using 4 if branches. Any ideas?
- Also wondering how to make it work for nested list

Even more wondering if this needs to be limited, since collections might get huge. but maybe we can leave this as tool developer responsibility, it just needs proper documentation. 

Idea came while working on tests for the filters: https://github.com/galaxyproject/galaxy/pull/8481

### DIFF
--- a/lib/galaxy/tools/parameters/dynamic_options.py
+++ b/lib/galaxy/tools/parameters/dynamic_options.py
@@ -650,20 +650,24 @@ class DynamicOptions(object):
 
     def get_fields(self, trans, other_values):
         if self.dataset_ref_name:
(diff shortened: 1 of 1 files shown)


bernt-matthias:
superseeded by https://github.com/galaxyproject/galaxy/pull/8599

PR #8560, opened 2019-09-02
[19.05] Allow missing H5 groups in Loom files

nsoranzo:
Fix the following warning when uploading to Galaxy a Loom file produced by
Bioconductor LoomExperiment:

```
Warning: The file 'Type' was set to 'loom' but the file does not appear to be of that type
WARNING:galaxy.datatypes.binary:<galaxy.datatypes.binary.Loom object at 0x2b1245ada410>, set_meta Exception: "Unable to open object (object 'layers' doesn't exist)
```

Also extend the Loom sniffer for when the optional `LOOM_SPEC_VERSION` attribute is missing.

### DIFF
--- a/lib/galaxy/datatypes/binary.py
+++ b/lib/galaxy/datatypes/binary.py
@@ -843,7 +843,7 @@ class Loom(H5):
                 dataset.creation_date = loom_file.attrs.get('creation_date', None)
                 dataset.metadata.shape = tuple(loom_file['matrix'].shape)
 
-                tmp = list(loom_file['layers'].keys())
+                tmp = list(loom_file.get('layers', {}).keys())
                 dataset.metadata.layers_count = len(tmp)
                 dataset.metadata.layers_names = tmp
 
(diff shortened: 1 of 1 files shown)


lib/galaxy/datatypes/binary.py
```diff
@@ -855,11 +860,15 @@ def set_meta(self, dataset, overwrite=True, **kwd):
                 dataset.metadata.col_attrs_count = len(tmp)
                 dataset.metadata.col_attrs_names = tmp
 
-                tmp = list(loom_file['col_graphs'].keys())
+                # According to the Loom file format specification, col_graphs
+                # and row_graphs are mandatory groups, but files created by
+                # Bioconductor LoomExperiment do not always have them:
+                # https://github.com/Bioconductor/LoomExperiment/issues/7
+                tmp = list(loom_file.get('col_graphs', {}).keys())
```
bgruening, line 867: But we should stick to the specification here and hack around this in the tools?

nsoranzo: This PR prevents an unnecessary warning in the dataset info, these metadata elements would be empty any way.
We don't have control on how Loom dataset are created using Bioconductor LoomExperiment (could be outside Galaxy or in any number of Galaxy tools), the correct solution is to fix https://github.com/Bioconductor/LoomExperiment/issues/7 .

PR #8567, opened 2019-09-04
fix history export for jobs running as real user

bernt-matthias:
the temporary directory used for the history export permissions 700 and the owner is the galaxy system user, i.e. it can't be accessed by the real user.

the easiest fix seems to be to chmod the temporary directory to 755.

### DIFF
--- a/lib/galaxy/tools/imp_exp/__init__.py
+++ b/lib/galaxy/tools/imp_exp/__init__.py
@@ -4,6 +4,7 @@ import shutil
 import tempfile
 
 from galaxy import model
+from galaxy import util
 from galaxy.model import store
 from galaxy.version import VERSION_MAJOR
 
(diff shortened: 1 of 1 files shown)


bernt-matthias:
I guess I need to add a mock `GalaxyAppConfiguration` to the failing tests. Has this been done already somewhere else where I could get some inspiration?

nsoranzo:
Do you have an error or traceback? I don't feel very comfortable with read permissions for everyone on tmp files.

bernt-matthias:
The stderr of the export job was: 

```
Error creating history archive: [Errno 13] Permission denied: '/work/songalax/galaxy/database/tmp/tmp5Qutxr'
```

Which is as far as I can see it the missing read permission for the directory which is about to be zipped.

An alternative would be to call the external chown script on the folder.... would be easy to implement - I guess :)

jmchilton:
This is not great - can you write this to the job directory? That would be a bit of a hack still since I think this is happening in a web threat and not in the job handler - but all this I/O is already happening in the web thread so that wouldn't be worse?

bernt-matthias:
@jmchilton your idea sounds great since the job directory is chowned to the real user anyway. 

Problem is that it seems that we don't have access to the job dir from here. As far as I can see only a `Job` object is initialised which seemingly has not directory assigned. ... Lots of guessing here from my side.

bernt-matthias:
superseeded by https://github.com/galaxyproject/galaxy/pull/9001

PR #8569, opened 2019-09-04
Enable mulled containers by default for 19.09.

jmchilton:
It has been in beta for too long and it is has really seen some serious use over the last year. I think it is time to enable it by default?


mvdbeek (approved):

martenson (approved):

martenson:
Exciting and overdue!

PR #8571, opened 2019-09-04
Remove collection ID character restriction from 'Relabel collection from file' tool

### LINKED ISSUE #8564: Relabel List Identifiers more restrictive than tool to build collection
https://github.com/galaxyproject/galaxy/blob/e9d8c900cb2a6e794b90a99e351c281b71d9123e/lib/galaxy/tools/__init__.py#L2966

There is no real specification on what is and isn't permitted in a collection element identifier.
This restricts it to nothing more than alpha-numeric, dash, and underscore. That is more restrictive than the upload tool to create collections.

innovate-invent:
Resolves: https://github.com/galaxyproject/galaxy/issues/8564

### DIFF
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -2962,9 +2962,6 @@ class RelabelFromFileTool(DatabaseOperationTool):
             for i, dce in enumerate(hdca.collection.elements):
                 dce_object = dce.element_object
                 add_copied_value_to_new_elements(new_labels[i], dce_object)
-        for key in new_elements.keys():
-            if not re.match(r"^[\w\-_]+$", key):
-                raise Exception("Invalid new colleciton identifier [%s]" % key)
         self._add_datasets_to_history(history, itervalues(new_elements))
         output_collections.create_collection(
             next(iter(self.outputs.values())), "output", elements=new_elements


jmchilton:
This restriction matches what I encourage in the Planemo docs (https://planemo.readthedocs.io/en/latest/writing_advanced.html#processing-identifiers) - I'd prefer to not lift the restriction here. I get really nervous about shell characters in collection identifiers exactly because they may become file names and I don't want information loss/imprecise mapping between collections and supplied inputs to tools.

Update: I'm not unwilling to compromise here - maybe make this an option? Or expand the regex to include more characters as long as they are shell safe? Or maybe @mvdbeek disagrees with me that any of this is needed at all.

mvdbeek:
I think we should allow ` `, `.`, `_` and `-` at least, these are very common in sample names, and often you'd want to use those separators to split the name into more structured metadata (I'm thinking of group tags, though that's not really structured either). I agree with @jmchilton (and probably everyone) that imprecise mapping is bad, but in a sense that regex in the planemo docs already does that, and in practice I think it works ok.

Of course that means everywhere we consume element identifiers we need to quote, but we should do that anyway, and we can have these characters in dataset names already.

So let's keep the regex but add ` \.,`, and we can skip `_` since it is included in `\w`

mvdbeek:
I opened a PR that adds back a more lenient character restriction: https://github.com/galaxyproject/galaxy/pull/8757

PR #8579, opened 2019-09-05
Modify tool panel section text color, size and margin to improve readability

### LINKED ISSUE #8577: tool panel detailing
During the last dev cycle we made a bunch of changes to toolpanel design and I think I found some inconsistencies in the current visuals, especially regarding readability.

At least 3 things have changed: color of all fonts, margins, and font size. One of the bad side effects is that the labels don't fit one line.

I think it would be better if we get back the gray for section labels and some margins. Something like this: 

But we also might want to leave the blue font in the section names and tools themselves, since blue (`#2c3143`) seems to be the new Galaxy standard

Another thing I have noticed, the tool id part of the tool entry in the panel should probably still be higher font weight or something, given we removed the underline.

guerler:
Fixes #8577

### DIFF
--- a/client/galaxy/style/scss/base.scss
+++ b/client/galaxy/style/scss/base.scss
@@ -1357,6 +1357,9 @@ div.permissionContainer {
     }
     background: $panel-bg-color;
     min-height: 100%;
+    .tool-old-link {
+        font-weight: bold;
+    }
 }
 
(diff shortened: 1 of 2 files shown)


jmchilton (approved):

martenson:
thanks @guerler

PR #8581, opened 2019-09-06
set uwsgi umask default

bernt-matthias:
seems that uWSGI sometimes has a default umask of 000:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=866058

I suggest to chose a somewhat safer default in the uWSGI
section of the galaxy.yml

### DIFF
--- a/doc/source/admin/cluster.md
+++ b/doc/source/admin/cluster.md
@@ -330,7 +330,7 @@ Since this is a complex problem, the current solution does have some caveats:
 
 ### Configuration
 
-You'll need to ensure that all datasets are stored on the filesystem such that they are readable by all users that will use Galaxy: either made readable by a group, or world-readable. If using a group, set your `umask(1)` to `027` or for world-readable, use `022` Setting the umask assumes your underlying filesystem uses POSIX permissions, so if this is not the case, your environment changes may be different.
+You'll need to ensure that all datasets are stored on the filesystem such that they are readable by all users that will use Galaxy: either made readable by a group, or world-readable. If using a group, set your `umask(1)` to `027` or for world-readable, use `022` Setting the umask assumes your underlying filesystem uses POSIX permissions, so if this is not the case, your environment changes may be different. For uWSGI setups (which are the default since release 19.01) the default umask is set in galaxy.yml (with a default of `027`).
 
 The directory specified in `new_file_path` in the Galaxy config should be world-writable, cluster-accessible (via the same absolute path) and have its sticky bit (+t) set. This directory should also be cleaned regularly using a script or program as is appropriate for your site, since temporary files created here may not always be cleaned up under certain conditions.
 
--- a/doc/source/admin/cluster.md
+++ b/doc/source/admin/cluster.md
@@ -330,7 +330,7 @@ Since this is a complex problem, the current solution does have some caveats:
 
 ### Configuration
 
-You'll need to ensure that all datasets are stored on the filesystem such that they are readable by all users that will use Galaxy: either made readable by a group, or world-readable. If using a group, set your `umask(1)` to `027` or for world-readable, use `022` Setting the umask assumes your underlying filesystem uses POSIX permissions, so if this is not the case, your environment changes may be different. For uWSGI setups (which are the default since release 19.01) the default umask is set in galaxy.yml (with a default of `027`).
+You'll need to ensure that all datasets are stored on the filesystem such that they are readable by all users that will use Galaxy: either made readable by a group, or world-readable. If using a group, set your `umask(1)` to `027` or for world-readable, use `022` Setting the umask assumes your underlying filesystem uses POSIX permissions, so if this is not the case, your environment changes may be different. For uWSGI setups (which are the default since release 19.01) the default umask is set in the `uwsgi:` section of `galaxy.yml` .
 
 The directory specified in `new_file_path` in the Galaxy config should be world-writable, cluster-accessible (via the same absolute path) and have its sticky bit (+t) set. This directory should also be cleaned regularly using a script or program as is appropriate for your site, since temporary files created here may not always be cleaned up under certain conditions.
 
--- a/lib/galaxy/config/sample/galaxy.yml.sample
+++ b/lib/galaxy/config/sample/galaxy.yml.sample
@@ -97,8 +97,11 @@ uwsgi:
   enable-threads: true
 
   # On some systems uWSGI has a default umask of 000, for Galaxy a 
-  # somewhat safer default is chosen
-  umask: 022
+  # somewhat safer default is chosen. If Galaxy submits jobs as real
+  # user then all users needs to be able to read the files, i.e.
+  # the the umask needs to be 0o022 or the Galaxy users need to be in the
+  # same group as the Galaxy system user
+  umask: 0o027
 galaxy:
 
   # The directory that will be prepended to relative paths in options
--- a/lib/galaxy/config/sample/galaxy.yml.sample
+++ b/lib/galaxy/config/sample/galaxy.yml.sample
@@ -96,6 +96,7 @@ uwsgi:
   # Ensure application threads will run if `threads` is unset.
   enable-threads: true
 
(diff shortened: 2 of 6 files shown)


nsoranzo:
Does this manifest in the creation of files with unsafe permissions?

bernt-matthias:
On our system job working directories, data sets, and conda environments were created with 666 or 777, respectively if this option is not set.

bernt-matthias:
Some of these could be fixed in the Galaxy sources (all except conda) by adding a call to https://github.com/galaxyproject/galaxy/blob/5c78159d610fe18713c77522062283a60c5eac30/lib/galaxy/util/__init__.py#L1289 after all file/directory creation using a safe permission pattern (in addition to considering the umask .. which might be 0).

bernt-matthias:
but setting the umask properly seems better to me

natefoo:
I'd suggest `027` to ensure that e.g. datasets are not world readable:

```console
% rm foo && umask 022 && touch foo && ls -l foo
-rw-r--r-- 1 nate nate 0 Sep  6 11:09 foo
% rm foo && umask 027 && touch foo && ls -l foo
-rw-r----- 1 nate nate 0 Sep  6 11:09 foo
```

Could you also add this option to `scripts/get_uwsgi_args.py` so that it'll be set by `run.sh` if there is no `galaxy.yml` or the option is not set in `galaxy.yml`?

bernt-matthias:
Can do. I will add to the docs that 022 is needed if jobs run as real user. Or if desired for other reasons.

bernt-matthias:
OK. Changed it to 027 

- I guess octal representation is better, but I'm not sure if this is properly parsed from the yml file -- I could test on my test system
- added it also to the script (but this is based a bit on guessing .. so please check)
- also added to docs

bernt-matthias:
Not sure if the failing tests are related..?

nsoranzo:
> Not sure if the failing tests are related..?

Looks like they are, since they are permission-related. In particular there seems to be an issue with tempfiles.

lib/galaxy/config/sample/galaxy.yml.sample
```diff
@@ -96,6 +96,12 @@ uwsgi:
   # Ensure application threads will run if `threads` is unset.
   enable-threads: true
 
+  # On some systems uWSGI has a default umask of 000, for Galaxy a 
+  # somewhat safer default is chosen. If Galaxy submits jobs as real
+  # user then all users needs to be able to read the files, i.e.
+  # the the umask needs to be 0o022 or the Galaxy users need to be in the
```
nsoranzo:
```suggestion
  # the the umask needs to be '022' or the Galaxy users need to be in the
```

lib/galaxy/config/sample/galaxy.yml.sample
```diff
@@ -96,6 +96,12 @@ uwsgi:
...
+  # same group as the Galaxy system user
+  umask: 027
```
nsoranzo, line 104:
```suggestion
  umask: '027'
```

doc/source/admin/cluster.md
```diff
@@ -330,7 +330,7 @@ Since this is a complex problem, the current solution does have some caveats:
 
 ### Configuration
 
-You'll need to ensure that all datasets are stored on the filesystem such that they are readable by all users that will use Galaxy: either made readable by a group, or world-readable. If using a group, set your `umask(1)` to `027` or for world-readable, use `022` Setting the umask assumes your underlying filesystem uses POSIX permissions, so if this is not the case, your environment changes may be different.
+You'll need to ensure that all datasets are stored on the filesystem such that they are readable by all users that will use Galaxy: either made readable by a group, or world-readable. If using a group, set your `umask(1)` to `027` or for world-readable, use `022` Setting the umask assumes your underlying filesystem uses POSIX permissions, so if this is not the case, your environment changes may be different. For uWSGI setups (which are the default since release 19.01) the default umask is set in galaxy.yml (with a default of `027`).
```
nsoranzo:
```suggestion
You'll need to ensure that all datasets are stored on the filesystem such that they are readable by all users that will use Galaxy: either made readable by a group, or world-readable. If using a group, set your `umask(1)` to `027` or for world-readable, use `022` Setting the umask assumes your underlying filesystem uses POSIX permissions, so if this is not the case, your environment changes may be different. For uWSGI setups (which are the default since release 19.01) the default umask is set in the `uwsgi:` section of `galaxy.yml` .
```

nsoranzo (commented):
I think you need to quote the umask in `galaxy.yml` to be kept as a string.

nsoranzo (approved):

bernt-matthias:
Question from yesterday was: which uwsgi version is used on my system (one that is installed by galaxy or another one). 

How can I find out?

bernt-matthias:
`ps aux | grep uwsgi`: 

gives

```
GALAXY_ROOT/.venv/bin/python2.7 .venv/bin/uwsgi --yaml GALAXY_ROOT/config/galaxy.yml --static-safe GALAXY_ROOT/client/galaxy/images --master --daemonize2 /var/log/galaxy/galaxy.log --pidfile2 galaxy.pid
```

So I guess its the correct one 

`.venv/bin/uwsgi --version` gives 2.0.18

lib/galaxy/config/sample/galaxy.yml.sample
```diff
@@ -96,6 +96,12 @@ uwsgi:
   # Ensure application threads will run if `threads` is unset.
   enable-threads: true
 
+  # On some systems uWSGI has a default umask of 000, for Galaxy a
```
mvdbeek: I think this also needs to be appended to https://github.com/mvdbeek/galaxy/blob/153ba0ce5151aaf120af9367316798ab4896bf46/lib/galaxy/config/config_manage.py#L165

bernt-matthias: Sure.

nsoranzo: @bernt-matthias You really need to copy the whole config option documentation to the value of the `'desc'` key in `lib/galaxy/config/config_manage.py`, then execute `make config-rebuild` to update the sample files.

bernt-matthias: OK, but if I do this then there are also changes in other docs. Should I just commit those as well?

mvdbeek: Yes, that'd be good.

mvdbeek:
Sorry, can you rebase out the last commit, rebase against dev (including the conflicting https://github.com/galaxyproject/galaxy/pull/9125) and then run make config-rebuild again ?

lib/galaxy/config/sample/galaxy.yml.sample
```diff
@@ -901,8 +910,8 @@ galaxy:
   # prevent a class of attack called clickjacking
   # (https://www.owasp.org/index.php/Clickjacking).  If you configure a
   # proxy in front of Galaxy - please ensure this header remains intact
-  # to protect your users.  Uncomment and leave empty to not set the `X
-  # -Frame-Options` header.
+  # to protect your users.  Uncomment and leave empty to not set the
```
nsoranzo: Interestingly this and other differences are due to using [TextWrapper](https://docs.python.org/3/library/textwrap.html#textwrap.TextWrapper) from Python 3. I've improved this for both 2 and 3 in https://github.com/galaxyproject/galaxy/pull/9128 .

bernt-matthias:
Seems that I got the rebase correct now :)

nsoranzo (approved):
Maybe worth to rebase (again!) on top of https://github.com/galaxyproject/galaxy/pull/9128 when that's merged to have a clean diff, but no big deal.

mvdbeek:
Just merged https://github.com/galaxyproject/galaxy/pull/9128, can you try rebasing again (sorry!) ?

mvdbeek (approved):

nsoranzo (approved):

PR #8583, opened 2019-09-06
Adding name option to tests

astrovsky01:
Adds optional names for tests in tool xml files.

### DIFF
--- a/lib/galaxy/tool_util/xsd/galaxy.xsd
+++ b/lib/galaxy/tool_util/xsd/galaxy.xsd
@@ -749,7 +749,7 @@ be present in ``test`` definition.</xs:documentation>
         <xs:documentation xml:lang="en">Maximum amount of time to let test run.</xs:documentation>
       </xs:annotation>
     </xs:attribute>
-    <xs:attribute name="name" type="xs:string" use="optional">
+    <xs:attribute name="title" type="xs:string" use="optional">
         <xs:annotation>
             <xs:documentation xml:lang="en">Title given to a test to be displayed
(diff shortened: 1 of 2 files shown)


lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -749,6 +749,14 @@ be present in ``test`` definition.</xs:documentation>
         <xs:documentation xml:lang="en">Maximum amount of time to let test run.</xs:documentation>
       </xs:annotation>
     </xs:attribute>
+    <xs:attribute name="name" type="xs:string" use="optional">
```
mvdbeek: what do you think about `title`, that mirrors what we do for the tool tag, and would leave space for `description` should we decide to use that one day

astrovsky01: After talking with @bgruening about the tours that are being worked on, I was saving "title" to be used for the tour step names, while "name" would be for internal planemo testing, similar how <param name> vs <param label> work. If you disagree, I can change it.

mvdbeek: Why should this be different in tours ?

astrovsky01: The name label would be used to show what exactly is being changed/tested, while the tour title would be in order to explain the purpose that test. I was thinking one would be more from a dev-perspective, while the other is more user-oriented.

astrovsky01: I suppose that using ```label``` would do just as well. I'll change to ```title```.

PR #8589, opened 2019-09-07
Move interactive tools

bgruening:
This PR moves the interactive tools into a more appropriate place. The proposed plan it to leave them there for the 19.09 release cycle and stress test the proposed interface/API. If the community is happy we want to remove them from Galaxy and treat them like Galaxy tools.

Does `./tools/interactive`  make to much promises about the future? Anyone prefers `./config/plugins/interactive`?


mvdbeek:
Makes more sense to have them in `./tools/interactive` to me. But shouldn't they be regular tool shed tools ?

martenson:
@mvdbeek I assume that is what @bgruening meant by 

> If the community is happy we want to remove them from Galaxy and treat them like Galaxy tools.

bgruening:
@mvdbeek we are not yet sure if the interface is stable enough. 3 more month to test things, more IEs, Shiny-stuff, writing docs and so on and then reevaluate everything in December. If we think the interface is stable let's move this into tools-iuc.
Bonus points if we can test those tools before we move them to tools-iuc

jmchilton (approved):

PR #8593, opened 2019-09-09
[19.09] WIP: Revert 8377, can lead to hang on login on wrong password

mvdbeek:
Though I don't understand how yet


martenson:
@mvdbeek what should we do with this? It seems like a thing that could block a release?

mvdbeek:
Yeah, but I couldn't figure out what was wrong. It seems to work fine for everyone else, so I think I'll close it and keep that in mind if anyone reports something similar.

PR #8597, opened 2019-09-09
Fix Pulsar Container Handling (less hacky)

jmchilton:
Less hacky version of #8533 for dev/20.01 and requiring a newer Pulsar.


jmchilton:
Merged as part of #8625

PR #8599, opened 2019-09-09
Dynamic options and metadata filter fixes

bernt-matthias:
Dynamic options:

- allow creation dynamic options from data parameters with `multiple="true"` and collections see: 
https://github.com/galaxyproject/galaxy/blob/3cb9d3cc5a048fe12f498f1d5a9a02a62e69d86b/test/functional/tools/dbkey_filter_collection.xml

Metadata Filter: 

- Fixes the filter to the documented behavior: Previously the filter removed all options if no meta data was defined in the referred file (which apparently is also not good for using tools in workflows https://github.com/galaxyproject/tools-iuc/pull/2498). This was because the filter assumed a meta data value of `None` if the meta data is not set. Therefore this https://github.com/galaxyproject/galaxy/blob/8ac4dc65ab316763d01975bd1d1b10ca90ee1481/lib/galaxy/tools/parameters/dynamic_options.py\#L201 evaluated to True only for non-existent meta data keys (note that the old statement in this if block never worked). 
The check should be be against the value given in the no_value attribute of the filter
specification. 
	- For the implementation I used the element_is_set function which also just checked for the existence of the meta data. Now it does what the name suggests.
	- For lists and collections the filter depended on the multiple attribute of the filter if all elements were considered. Since this was not documented I changed it to always use all elements
- If the refrered datasets/collections have multipe meta data values set then all of these are allowed, i.e. the options must match at least one of these values. making `meta_value` a list also seems in line with the fact that `compare_meta_value` can process lists of meta data values.

For discussion: 

- I started implementing filtering on more complex metadata types (lists, dicts, files): https://github.com/galaxyproject/galaxy/blob/eca469a53a51bd8717e98636eab62173b02518a4/lib/galaxy/tools/parameters/dynamic_options.py#L166 Not sure if this is really useful as it is now-- in particular for files. I plan to implement `from_dataset_metadata` for selects .. then it might be useful. Maybe for now remove File metadata filtering
- It might be a good idea to deprecate or remove the metadata filtering implemented in RemoveValueFilter https://github.com/galaxyproject/galaxy/blob/eca469a53a51bd8717e98636eab62173b02518a4/lib/galaxy/tools/parameters/dynamic_options.py#L474

More ideas:

- I would suggest to remove/deprecate the `multiple` (and `separator`) attributes [have they ever been used]. Reason: One _should_ be able to achieve the same in combination with the `multiple_splitter` filter.

### DIFF
--- a/lib/galaxy/tools/parameters/dynamic_options.py
+++ b/lib/galaxy/tools/parameters/dynamic_options.py
@@ -2,8 +2,8 @@
 Support for generating the options for a SelectToolParameter dynamically (based
 on the values of other parameters or other aspects of the current state)
 """
-from collections import OrderedDict
 import copy
+from collections import OrderedDict
 import logging
 import os
 import re
--- a/lib/galaxy/tools/parameters/dynamic_options.py
+++ b/lib/galaxy/tools/parameters/dynamic_options.py
@@ -7,7 +7,6 @@ import logging
 import os
 import re
 from io import StringIO
-from collections import OrderedDict
 
 from galaxy.model import (
     HistoryDatasetAssociation,
@@ -206,7 +205,7 @@ class DataMetaFilter(Filter):
             if not r.metadata.element_is_set(self.key):
                 continue
             _add_meta(meta_value, r.metadata.get(self.key))
-        log.error("meta_value %s" % str(meta_value))
+
         # if no meta data value could be determined just return a copy
         # of the original options
         if len(meta_value) == 0:
--- a/lib/galaxy/tools/parameters/dynamic_options.py
+++ b/lib/galaxy/tools/parameters/dynamic_options.py
@@ -158,6 +158,11 @@ class DataMetaFilter(Filter):
         return self.ref_name
 
     def filter_options(self, options, trans, other_values):
+        def _add_meta( meta_value, m ):
+            if isinstance(m, list):
+                meta_value |= set(m)
+            else:
+                meta_value.add(m)
         def compare_meta_value(file_value, dataset_value):
             if isinstance(dataset_value, list):
                 if self.multiple:
@@ -185,18 +190,18 @@ class DataMetaFilter(Filter):
         # - for data sets: the meta data value
         # in both cases only meta data that is set (i.e. differs from the no_value)
         # is considered
-        meta_value = None
+        meta_value = set()
         if is_data_list:
-            meta_value_set = set([_.metadata.get(self.key, None) for _ in ref if _.metadata.element_is_set(self.key)])
-            meta_value_set.discard(None)
-            if len(meta_value_set) > 0:
-                meta_value = list(meta_value_set)
+            for r in ref:
+                if not r.metadata.element_is_set(self.key):
+                    continue
+                _add_meta(meta_value, r.metadata.get(self.key))
         else:
             if ref.metadata.element_is_set(self.key):
-                meta_value = [ref.metadata.get(self.key, None)]
+                _add_meta(meta_value, ref.metadata.get(self.key))
         # if no meta data value could be determined just return a copy
         # of the original options
-        if meta_value is None:
+        if len(meta_value) == 0:
             return copy.deepcopy(options)
 
         if self.column is not None:
@@ -213,7 +218,7 @@ class DataMetaFilter(Filter):
                     "selected" : 2
                 }
                 self.dynamic_option.largest_index = 2
-            for value in reduce(lambda x, y: x + y, meta_value):
+            for value in meta_value:
                 options.append((value, value, False))
             return options
 
--- a/lib/galaxy/tools/parameters/dynamic_options.py
+++ b/lib/galaxy/tools/parameters/dynamic_options.py
@@ -2,6 +2,7 @@
 Support for generating the options for a SelectToolParameter dynamically (based
 on the values of other parameters or other aspects of the current state)
 """
+import copy
 import logging
 import os
 import re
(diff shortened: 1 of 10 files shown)


jmchilton (approved):

bernt-matthias:
I dont understand the failing unit test https://circleci.com/gh/galaxyproject/galaxy/36869?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link 

Can someone help me understand?

bernt-matthias:
Still dont understand the error. One could simply change the expected error to `parameter 'my_name': requires a value, but no legal values defined`

But I don't know if this makes sense.

bernt-matthias:
Currently only metadata given in the form of text or lists is used properly. Anything else, like dict, orderedDict, or even files is currently ignored (or better: I don't know what would happen). For the files my guess would be that the filenames are listed.

Metadata seems to be int, string, list, Ordered dict (which are MetadataParameter) and file names (which are metadata.FileParameter ) .. so one can differentiate them.

Question is if there is an easy way to address the data of interest

How about the following:

- [x] for int and string metadata: just add them to the options
- [x] for metadata given as lists: we extend the options list by the contents of the list
- [ ] for metadata given as mapping types: add the key value pairs as comma separated strings to the options list
  - [x] implement 
  - [ ] add test
- [ ] for files: add the lines of the file to the options list (would only work for text files)
  - [ ] implement
  - [ ] test

For the latter case a combination of the regexp and multiple_splitter filters the the relevant pieces of the info can be extracted.

One thing that one needs to keep in mind is that usually only a part of large files is considered.

One alternative would be to implement something analogous to from_dataset, e.g. from_dataset + key ...

lib/galaxy/tools/parameters/dynamic_options.py
```diff
@@ -158,8 +163,20 @@ def get_dependency_name(self):
         return self.ref_name
 
     def filter_options(self, options, trans, other_values):
+        def _add_meta(meta_value, m):
+            if isinstance(m, list):
+                meta_value |= set(m)
+            elif isinstance(m, dict) or isinstance(m, OrderedDict):
```
mvdbeek:
```suggestion
            elif isinstance(m, dict):
```
OrderedDict is an instance of dict.

lib/galaxy/tools/parameters/dynamic_options.py
```diff
@@ -158,8 +163,20 @@ def get_dependency_name(self):
...
+                meta_value |= set(["%s,%s" % (k, v) for k, v in m.iteritems()])
```
mvdbeek:
```suggestion
                meta_value |= set(["%s,%s" % (k, v) for k, v in m.items()])
```

mvdbeek (commented):
I think the API test errors are probably relevant ?

bernt-matthias:
> I think the API test errors are probably relevant ?

Likely.

I stalled the work on this PR because I was asking myself how far we can/want to push this. In particular getting information from metadata files might be dangerous (because it might take long) and limiting this (e.g. to reading the first MB only) might be limiting the usability.

mvdbeek:
Yup, I agree those are the issues, but I think there are some good fixes in there.

bernt-matthias:
OK. I removed the parsing of metadata files. Same question is of course relevant for collections. Maybe we should parse at most 1MB of the complete collection? Or just take the first file. 

Thinking if I should split out the treatment of the `no_value` since this is a bug that considers backporting (https://github.com/galaxyproject/tools-iuc/pull/3264). Or would you consider the complete PR a bugfix?

davebx (approved):

bernt-matthias:
I hope that the failing API test is unrelated :) .. even if getting options from implicitly converted data sets might be an interesting test case.

mvdbeek:
The test fails locally for me as well, very likely this is related.

bernt-matthias:
https://github.com/galaxyproject/galaxy/pull/8599/commits/1882aa641cdde1c51c077fd983f2241971308377 should fix the failing api tests. when I introduced the change to `to_hda_representative` I  liked the idea that the type of the returned value does not depend on a parameter. I still do .. but it would require more changes. 

I also thought a bit more about the tests that failed (test_implicit_conversion_and_reduce). FYI the test has been introduced here 
https://github.com/galaxyproject/galaxy/pull/8932. Seems that the use case is the following: 

The test runs column_multi_param (https://github.com/galaxyproject/galaxy/blob/dev/test/functional/tools/column_multi_param.xml) using a fasta in a list and a data_column that is set to one.

Actually the input needs tabular. So an implicit conversion would be needed (and I guess this is the aim of the test). 

The odd thing is that the data_column parameter that refers to the data input would not work anyway, since the conversion would need to run before the tool started, i.e. upon loading the tool form. So the tested case does not work in reality anyway.

However the test sets the data_column to one explicitly .. maybe this is OK for the test.

mvdbeek:
> However the test sets the data_column to one explicitly .. maybe this is OK for the test.

That's fine, and that's how workflows work too.

bernt-matthias:
> I liked the idea that the type of the returned value does not depend on a parameter. I still do .. but it would require more changes.

Actually I liked the idea that the return type does not depend on the parameter **and the number of returned HDAs** - because this requires additional tests in the calling code. If the tests work I may do this in another PR. If not I tend to revert the last commit and do it right here.

bernt-matthias:
Seems that all tests pass.

bernt-matthias:
@davebx @mvdbeek do we need/want a backport?

mvdbeek:
Maybe, can you run the IUC tests against dev ?

bernt-matthias:
Can do .. even if @davebx just merged https://github.com/galaxyproject/tools-iuc/pull/3264 which would have been a nice test case.

bernt-matthias:
One more problem revealed by the planemo tests https://github.com/galaxyproject/galaxy/pull/10691 ..

bernt-matthias:
With https://github.com/galaxyproject/galaxy/pull/10691 I was able to successfully test all IUC tools: https://github.com/galaxyproject/tools-iuc/actions/runs/357789135

Note, I skipped all tools that fail on 20.09.

test/functional/tools/samples_tool_conf.xml
```diff
@@ -44,8 +44,11 @@
   <tool file="inputs_as_json_profile.xml" />
   <tool file="inputs_as_json_with_paths.xml" />
   <tool file="filter_static_regexp.xml" />
+  <tool file="select_from_dataset.xml" />
   <tool file="dbkey_filter_input.xml" />
   <tool file="dbkey_filter_multi_input.xml" />
+  <tool file="dbkey_filter_collection.xml" />
+  <tool file="dbkey_filter_paired_collection.xml" />
```
nsoranzo, line 51: @bernt-matthias It was a while ago, but did you mean to add a `dbkey_filter_paired_collection.xml` test tool? Or should we remove this entry?

bernt-matthias: @nsoranzo I guess I wanted to add it, but forgot to add the actual test tool. I guess for now its best to remove the entry from samples_tool_conf.xml .

nsoranzo: OK, thanks! Done in https://github.com/galaxyproject/galaxy/pull/13035

PR #8601, opened 2019-09-10
Add Tool Manager Component for Admin Panel

guerler:
This PR follows-up on the toolshed separation task by adding a Tool Manager component to the admin panel. The view is accessible under the `Install and Uninstall` option in the admin panel, followed by checking the `Installed only` radio button. This PR does not remove any legacy code. The removal of legacy code should be done in a separate PR. This Tool Manager component is an addition the Toolshed component introduced in #8090. This PR also contains a style fix for admin panel labels.

image

Limitation: The `long_description` is missing in the detail view of installed repositories. Currently, this information is not provided by the toolshed api endpoint when requesting repository details by `name` and `owner`. However this is barely noticable and not worth a separate request to the same api using the repository `id`. All other attributes required to render the repository details view are provided. These attributes are `id`, `name`, `owner` and `description`. If the `long_description` is added to the toolshed response in the future, it will appear in the repository details view.


jmchilton:
Does toolshed really need a column in this display? That is a lot of real estate and I think we should strongly discourage custom tool sheds. Feel free to discount my opinion though because I don't understand why we have a tool shed installed data manager list at all.

guerler:
@jmchilton thanks for the comment. I removed the `toolshed` column from the installed repository view. Although I wonder if this might lead to confusion in cases in which the same tool has been installed from two different toolshed servers. The `data manager` installation is just a test case to validate that the interface is able to install data manager tools if needed. If we want to prevent this, the toolshed server should not provide `data manager` tools in its search results.

mvdbeek:
I think we do need this column, but it can go to the repo details, I think.

guerler:
I think it would be good to remove the toolshed url as @jmchilton suggested. It is probably enough that users can access the toolshed details by clicking on `Show additional details and dependencies.` in the repository details view. Thanks for the comments.

jmchilton:
I'd like to have a consistent syntax, styling, classes, etc.. for these loading messages. Any interest in cherry-picking ~https://github.com/jmchilton/galaxy/commit/92e7b1064b9a7b808a88194e2764ccef1f743d05~ https://github.com/jmchilton/galaxy/commit/333ba7faf0  into this branch?

Update: New commit without some debug stuff left in by accident.

jmchilton (approved):

jmchilton:
Thanks for adding that, really great stuff overall @guerler - I learned a lot reading your code

I'd like in general to see this migrate toward being more clearly about managing "tool shed repositories" and not tools - and having a separate component for managing tools independent of the tool shed (link to the repository if installed, refactor Jobs.vue to see recent executions, summarize job metrics/users/etc.., run tests from the GUI, manage dependencies/containers #8741). I was a bit worried at first because this PR was so much effort that seemed counter to that, but now that I see it all together I think having the components decomposed like this actually is great and helps make my vision possible. Thanks a bunch @guerler !

martenson:
I think this is a great step towards unification of the repo (un)installation interface that we planned in Luray.

I am all down for hiding/deprioritizing non-critical information in order to make the UI more convenient and easy to use. However one thing that I think disappeared  in this PR (and is present in dev) is the `Monitor installation` menu item. At the moment there is not way admin can see what repositories are in non-final installation state, which is information I deem critical.

Until we have a better solution for it I think we should re-add the menu item.

client/galaxy/scripts/entry/panels/admin-panel.js
```diff
@@ -94,46 +94,35 @@ const AdminPanel = Backbone.View.extend({
...
                 items: [
                     {
-                        title: _l("Install or Uninstall"),
+                        title: _l("Install and Uninstall"),
                         url: "admin/toolshed",
                         target: "__use_router__",
                         enabled: this.settings.is_tool_shed_installed
                     },
                     {
-                        title: _l("Monitor installation"),
```
martenson: this is where it was removed

martenson:
Marvelous. 🚀

guerler:
Thanks a lot for the comments and help with this project!

PR #8607, opened 2019-09-10
[19.09] Add lower-case condition to user email validator

### LINKED ISSUE #8602: Login fails if e-mail in different case than stored in the database
The galaxy server in question uses authentication via LDAP exclusively, with auto-registration enabled. The e-mail address returned from the LDAP server for a specific user is in the form `User.Name@example.org` (note the upper case letters), but stored in the galaxy database completely in lower case. If the user tries to log in using the e-mail address, the login fails with the following message: `Auto-registration failed, contact your local Galaxy administrator. User with that email already exists.`

guerler:
Fixes #8602.

### DIFF
--- a/lib/galaxy/security/validate_user_input.py
+++ b/lib/galaxy/security/validate_user_input.py
@@ -33,6 +33,8 @@ def validate_email(trans, email, user=None, check_dup=True, allow_empty=False):
         return message
     if not(VALID_EMAIL_RE.match(email)):
         message = "The format of the email address is not correct."
+    elif not email.islower():
+        message = "Please use lower-case characters for your email."
     elif len(email) > EMAIL_MAX_LEN:
         message = "Email address cannot be more than %d characters in length." % EMAIL_MAX_LEN
     elif check_dup and trans.sa_session.query(trans.app.model.User).filter_by(email=email).first():


lib/galaxy/security/validate_user_input.py
```diff
@@ -33,6 +33,8 @@ def validate_email(trans, email, user=None, check_dup=True, allow_empty=False):
         return message
     if not(VALID_EMAIL_RE.match(email)):
         message = "The format of the email address is not correct."
+    elif not email.islower():
+        message = "Please use lower-case characters for your email."
```
nsoranzo, line 37: This function is used in several places and I don't think we can just assume that Galaxy databases don't contain mixed-case email addresses. I have an alternative approach in the works though, should be finished soon.

guerler: This function will only prevent that new upper-case emails are entered into the database. It does not affect existing upper-case emails already stored. I think it is the correct solution for this problem but I am looking forward to your suggestion. I believe the existing upper-case emails should be approached as duplicate accounts and resolved.

nsoranzo: @guerler Here's my alternative solution: https://github.com/galaxyproject/galaxy/pull/8631

PR #8609, opened 2019-09-10
[19.09] Fix default for docker_sudo in new jobs configs.

jmchilton:
We implemented Docker support in a time when a lot of examples required sudo I think or maybe I was just very confused... but I've come to regret the default ``docker_sudo`` being on (True). All example job configs for Docker for Galaxy now include setting that to false as boilerplate.

YAML based job configuration files give us the opportunity to change this for new configuration files I think. This attempts to change the default to False (which also matches Singularity) for new YAML based configuration files.

### DIFF
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -168,6 +168,10 @@ def job_config_xml_to_dict(config, root):
         environment["metrics"] = metrics_to_dict
 
         params = JobConfiguration.get_params(config, destination)
+        # Handle legacy XML enabling sudo when using docker by default.
+        if "docker_sudo" not in params:
+            params["docker_sudo"] = "true"
+
         # TODO: handle enabled/disabled in configure_from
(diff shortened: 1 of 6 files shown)


bgruening (approved):

bgruening:
> #x1B[91mERROR  : Home directory is not owned by calling user: /tmp/tmpyHPjuj/000/1/home

Interesting this happens during the Singularity tests.

jmchilton:
Indeed - it is because the base class for Docker and Singularity containers descriptions doesn't dispatch on the container type here (https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tool_util/deps/container_classes.py#L75). So singularity will use singularity defaults but will use ``docker_sudo``, ``docker_sudo_cmd``, etc... I guess I need to fix that bug first.

jmchilton:
Merged latest release_19.05 to get Singularity fix for failing tests from #8614.

PR #8610, opened 2019-09-10
[19.09] Fix autocreation of shed_*_conf.xml

### LINKED ISSUE #8484: Cannot install from Toolshed in fresh checkout
On a fresh checkout of `dev`, shed_tool_conf.xml.sample no longer exists, so shed_tool_conf.xml will not end up existing.

UI does not raise any error, but log shows:
```
galaxy.webapps.galaxy.api.tool_shed_repositories DEBUG 2019-08-20 12:18:52,926 [p:93571,w:1,m:0] [uWSGIWorker1Core0] The tool_config_file setting in galaxy.ini must include at least one shed tool configuration file name with a <toolbox> tag that includes a tool_path attribute value which is a directory relative to the Galaxy installation directory in order to automatically install tools from a tool shed into Galaxy (e.g., the file name shed_tool_conf.xml whose <toolbox> tag is <toolbox tool_path="database/shed_tools">).  For details, see the "Installation of Galaxy tool shed repository tools into a local Galaxy instance" section of the Galaxy tool shed wiki at https://galaxyproject.org/installing-repositories-to-galaxy/
```

Can manually restore a shed_tool_conf.xml from a previous .sample as workaround for now...

Most likely occurred around https://github.com/galaxyproject/galaxy/pull/921

Someone should confirm that all needed sample config files are available or being handle properly from a _fresh_ checkout.

natefoo:
Also stop checking `cwd` for `galaxy.yml`.

Fixes #8484 

Opinions welcome on `shed_tool_config_file`. I think this was something @jmchilton and I discussed many years ago, and it was part of #921, but I sorta forgot about it. Rationale is in the release notes text below.

----

For the release notes:

# Tool configuration file handling changes
  
## Shed tool configuration file

There have been a few changes to the way that Galaxy loads tool panel configuration files (e.g. `tool_conf.xml` and `shed_tool_conf.xml`) that deployers should be aware of before upgrading.

A new configuration option named `shed_tool_config_file` has been added, with a default value of `config/shed_tool_conf.xml`, and that path has been removed from the default value of the `tool_config_file` option.  This new option makes it possible for Galaxy to create the shed tool configuration file (e.g.  `shed_tool_conf.xml`) file on demand and moves us one step closer to a Galaxy installation that does not need to be run from the root of its source directory.

If you make no changes to your configuration, you should not encounter any problems. However, if you are using a non-default path to your primary shed tool config file, you should modify these options to remove the path of your primary `shed_tool_conf.xml` from `tool_config_file` and set it in `shed_tool_config_file`. Although Galaxy will continue to operate properly if you do not make this change, failure to do so will result in a second shed tool panel config file being created and loaded at startup time.

Although an unusual configuration, if you have multiple shed tool config files, you can still load them in `tool_config_file`, just be sure that one of them is set in `shed_tool_config_file`.

Note that Galaxy will fail to start if any of the files specified in `tool_conf_file` cannot be read at startup, whereas it will only fail to start if the file specified in `shed_tool_config_file` cannot be created (if it does not already exist).

## Migrated tools configuration file

Galaxy contains a (now unused) system for automatically installing tools from the Tool Shed that had been migrated out of the Galaxy source code after the creation of the Tool Shed. Older deployments that used this system may have tools in the migrated tools config file, `config/migrated_tools_conf.xml.sample` (by default). As of this release, an empty `migrated_tools_conf.xml` file is not created by default for new installations, but will still be read if it is found at the default path. The path can be changed with the `migrated_tools_config` option, and you can force Galaxy to fail startup if the file is missing by adding it to the `tool_config_file` option.

# Configuration file path defaults

## Deprecated paths will no longer be searched

The default location for Galaxy's configuration files has been in the `config/` directory for many years, prior to which the default was the root of the Galaxy directory. As of this release, Galaxy will no longer look in the current directory for configuration files, so you will need to specify the path to `galaxy.yml` e.g. with `uwsgi --yaml /path/to/galaxy.yml`, `GALAXY_CONFIG_FILE=/path/to/galaxy.yml run.sh`, or similar means, if it is not in the default location.

## Default paths to additional configuration files

As of this release, all other configuration files (e.g. `datatypes_conf.xml`, `job_conf.xml`) are searched for in the same directory as `galaxy.yml`. Prior to this release, the default was in the `config/` directory of the current working directory at startup (which has always been, and for the time being continues to be, the root of the Galaxy source).  This means that you do not need to individually configure the paths to these files in `galaxy.yml` if you are storing them in non-default locations.

### DIFF
--- a/lib/galaxy/config/sample/galaxy.yml.sample
+++ b/lib/galaxy/config/sample/galaxy.yml.sample
@@ -197,8 +197,21 @@ galaxy:
   # Tool config files, defines what tools are available in Galaxy. Tools
   # can be locally developed or installed from Galaxy tool sheds.
   # (config/tool_conf.xml.sample will be used if left unset and
-  # config/tool_conf.xml does not exist).
-  #tool_config_file: config/tool_conf.xml,config/shed_tool_conf.xml
+  # config/tool_conf.xml does not exist). Can be a single file, a list
+  # of files, or (for backwards compatibility) a comma-separated list of
+  # files.
+  #tool_config_file: config/tool_conf.xml
+
+  # Tool config file for tools installed from the Galaxy Tool Shed. Must
+  # be writable by Galaxy and generally should not be edited by hand. In
+  # older Galaxy releases, this file was part of the tool_config_file
+  # option. It is still possible to specify this file (and other shed-
+  # enabled tool config files) in tool_config_file, but in the standard
+  # case of a single shed-enabled tool config, this option is
+  # preferable. This file will be created automatically upon tool
+  # installation, whereas Galaxy will fail to start if any files in
+  # tool_config_file cannot be read.
+  #shed_tool_config_file: config/shed_tool_conf.xml
 
   # Enable / disable checking if any tools defined in the above non-shed
   # tool_config_files (i.e., tool_conf.xml) have been migrated from the
--- a/lib/galaxy/webapps/galaxy/config_schema.yml
+++ b/lib/galaxy/webapps/galaxy/config_schema.yml
@@ -186,14 +186,30 @@ mapping:
           nodes that will run Galaxy jobs, unless using Pulsar.
 
       tool_config_file:
-        type: str
-        default: config/tool_conf.xml,config/shed_tool_conf.xml
+        type: list or str
+        default: config/tool_conf.xml
         required: false
         desc: |
           Tool config files, defines what tools are available in Galaxy.
           Tools can be locally developed or installed from Galaxy tool sheds.
           (config/tool_conf.xml.sample will be used if left unset and
-          config/tool_conf.xml does not exist).
+          config/tool_conf.xml does not exist). Can be a single file, a list of
+          files, or (for backwards compatibility) a comma-separated list of files.
+
+      shed_tool_config_file:
+        type: str
+        default: config/shed_tool_conf.xml
+        required: false
+        desc: |
+          Tool config file for tools installed from the Galaxy Tool Shed. Must
+          be writable by Galaxy and generally should not be edited by hand. In
+          older Galaxy releases, this file was part of the tool_config_file
+          option. It is still possible to specify this file (and other
+          shed-enabled tool config files) in tool_config_file, but in the
+          standard case of a single shed-enabled tool config, this option is
+          preferable. This file will be created automatically upon tool
+          installation, whereas Galaxy will fail to start if any files in
+          tool_config_file cannot be read.
 
       check_migrate_tools:
         type: bool
--- a/lib/galaxy/webapps/galaxy/config_schema.yml
+++ b/lib/galaxy/webapps/galaxy/config_schema.yml
@@ -186,7 +186,7 @@ mapping:
           nodes that will run Galaxy jobs, unless using Pulsar.
 
       tool_config_file:
-        type: list or str
+        type: any
         default: config/tool_conf.xml
         required: false
         desc: |
--- a/doc/source/admin/galaxy_options.rst
+++ b/doc/source/admin/galaxy_options.rst
@@ -231,8 +231,28 @@
     Tool config files, defines what tools are available in Galaxy.
     Tools can be locally developed or installed from Galaxy tool
     sheds. (config/tool_conf.xml.sample will be used if left unset and
-    config/tool_conf.xml does not exist).
-:Default: ``config/tool_conf.xml,config/shed_tool_conf.xml``
+    config/tool_conf.xml does not exist). Can be a single file, a list
+    of files, or (for backwards compatibility) a comma-separated list
+    of files.
+:Default: ``config/tool_conf.xml``
+:Type: list or str
+
+
+~~~~~~~~~~~~~~~~~~~~~~~~~
+``shed_tool_config_file``
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+:Description:
+    Tool config file for tools installed from the Galaxy Tool Shed.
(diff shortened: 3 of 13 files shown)


martenson:
>  The path can be changed with the migrated_tools_conf.xml option

I assume this is `migrated_tools_config` option?

lib/galaxy/config/sample/galaxy.yml.sample
```diff
@@ -197,8 +197,21 @@ galaxy:
   # Tool config files, defines what tools are available in Galaxy. Tools
   # can be locally developed or installed from Galaxy tool sheds.
   # (config/tool_conf.xml.sample will be used if left unset and
-  # config/tool_conf.xml does not exist).
-  #tool_config_file: config/tool_conf.xml,config/shed_tool_conf.xml
+  # config/tool_conf.xml does not exist). Can be a single file, a list
+  # of files, or (for backwards compatibility) a comma-separated list of
+  # files.
+  #tool_config_file: config/tool_conf.xml
```
jdavcs, line 203: What if 'config_dir' is set by the admin to something other than 'config'? It would affect this, right? If so, maybe set the default to 'tool_conf.xml', and add to the desc something like 'Default value will be resolved to 'config/tool_conf.xml' where 'config' is the default value of the 'config_dir' option'? We don't *use* this value in `config/__init__.py` (where more 'config/' is hardcoded w/filenames), but if we did, then it would be a issue, i suspect. (I haven't fully grokked the processing of these config options, so it's a question, not a suggestion.)

natefoo: Yeah, I think maybe up at the top of the `galaxy` section in the sample we should put a header that explains how relative paths in the file work, like "all paths are relative to the directory in which galaxy.yml lives unless otherwise stated." I actually need to double check that everything works as intended - for example, some paths like the path to `tool-data/` are relative to `galaxy_root` aka `cwd`.

jdavcs: I was actually changing some defaults and adding those sort of comments where applicable, but paused when I saw your PR (if it's merged tomorrow, I'd rather wait it out rather than resolving 90 changed lines: I have at least as many for the same file :) 
I think it will be easier to do it per option: some are relative to config/, some to database/, and some, yes, to cwd(). More importantly, some of those defaults *will be used*: since every single  config option 'foo' is now auto-loaded and assigned to 'self.foo', I want to replace all `kwargs.get('foo', bar)` with `self.foo`. So, stuff like `x = resolve_path(kwargs.get('foo', bar), some-path)` would become `x = resolve_path(self.foo, some-path)`. Of course, this depends on what is 'bar', and how different it is from the default specified in `config_schema.yml`. Which is why if a default is specified as `config/foo`, doing the above would result in `config/config/foo`. And bad things would happen.

natefoo: > I think it will be easier to do it per option: some are relative to config/, some to database/, and some, yes, to cwd(). 

Yeah, ok, this makes sense. Sorry, I had a quick look over your config standardization PR last night but haven't had a chance to grok it.

We might even want to make a standard for this, like a new option in the schema (`path_relative_to`?) with values being other config vars (e.g. `data_dir`) or `cwd`, and then those can be automatically added to the comment just above the var in the sample?

Also, anything relative to `cwd` was supposed to be eliminated in #921, but not all of that was complete when it was merged.

jdavcs: `path_relative_to` - I think this is a very good idea! I'll play around with it..
`cwd`: I won't touch those for now; once the rest mostly done we can address these on a case by case basis.

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -186,14 +186,30 @@ mapping:
           nodes that will run Galaxy jobs, unless using Pulsar.
 
       tool_config_file:
-        type: str
-        default: config/tool_conf.xml,config/shed_tool_conf.xml
+        type: list or str
```
jmchilton: Just think you can makeup syntax as you go huh 😆?  ``KeyError: 'list or str'`` - this needs to be ``any`` I think. Config validation doesn't work at all on 19.09 though so I guess I can see why and it doesn't seem to have tests.

natefoo: I make it up, you implement it. It's a great relationship.

lib/galaxy/config/sample/galaxy.yml.sample
natefoo: It looks like I missed the mark in #921, if you look at [_parse_config_file_options](https://github.com/galaxyproject/galaxy/blob/be5aafd34b1f400704ef863684c9b742e1ca7427/lib/galaxy/config/__init__.py#L165) it resolves any configured values relative to `self.root` (aka `cwd`). Only the defaults if the option is unset are relative to other directories.

I'll probably address that in a separate PR though.

jdavcs: And those I'm not touching yet because they are in the BaseAppConfiguration, which does not load the schema (although the child can pass the loaded defaults via _set_config_base of course).
Also self.tool_data_path, i think - also resolves to cwd. https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L284

martenson:
thanks @natefoo

jdavcs:
Thank you @martenson! (been waiting on this :)

Issue #8586, opened 2019-09-06
prevent in-place building of TS search index


martenson:
@guerler and myself identified that every full hour for a couple of minutes MTS search is not responsive and our guess is that the in-place index regeneration might have to do with it.

PR #8611, opened 2019-09-10
[19.09] regenerate ts search indexes in tmp directory

martenson:
should fix https://github.com/galaxyproject/galaxy/issues/8586


guerler:
@martenson thank you!

PR #8618, opened 2019-09-11
python-ldap python3 update and documentation change

### LINKED ISSUE #8615: python-ldap on python3
When setting up 19.09 on python3 (probably this doesn't just affect 19.09) and using LDAP as authentication mechanism, it tries to use python-ldap version 2.4.44 however this gives the following error:
```python
  Downloading https://files.pythonhosted.org/packages/bb/6c/ad05e4d337bd0079d830604ec0cbcafa2b7ddd4376030cc14dde0e15766e/python-ldap-2.4.44.tar.gz (298kB)
    100% |████████████████████████████████| 307kB 26.6MB/s
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-ofdbvfak/python-ldap/setup.py", line 53
        print name + ': ' + cfg.get('_ldap', name)
                 ^
    SyntaxError: invalid syntax	
```

I switched the dependency to 3.2.0 in https://github.com/galaxyproject/galaxy/blob/release_19.09/lib/galaxy/dependencies/conditional-requirements.txt. I will comment on this once I confirm that this works.

selten:
Fix Active Directory example and bump the version of `python-ldap` to 3.2.0 as this works on Python 3. Fixes #8615.

### DIFF
--- a/lib/galaxy/config/sample/auth_conf.xml.sample
+++ b/lib/galaxy/config/sample/auth_conf.xml.sample
@@ -102,8 +102,8 @@
             <!-- For Active Directory: -->
 <!--        <bind-user>{sAMAccountName}@dc1.example.com</bind-user>
             <bind-password>{password}</bind-password>
-            <auto-register-username>{sAMAccountName}</auto-register-username>
-            <auto-register-email>{mail}</auto-register-email>
+            <auto-register-username>{username}</auto-register-username>
+            <auto-register-email>{email}</auto-register-email>
             <auto-register-roles>{gidNumber}</auto-register-roles>
 -->
             <!-- For OpenLDAP: -->
--- a/lib/galaxy/dependencies/conditional-requirements.txt
+++ b/lib/galaxy/dependencies/conditional-requirements.txt
@@ -10,7 +10,7 @@ statsd
 docker
 azure-storage==0.32.0
 # PyRods not in PyPI
-python-ldap==2.4.44
+python-ldap==3.2.0
 python-pam
 galaxycloudrunner
 


lib/galaxy/config/sample/auth_conf.xml.sample
```diff
@@ -102,8 +102,8 @@
             <!-- For Active Directory: -->
 <!--        <bind-user>{sAMAccountName}@dc1.example.com</bind-user>
             <bind-password>{password}</bind-password>
-            <auto-register-username>{sAMAccountName}</auto-register-username>
-            <auto-register-email>{mail}</auto-register-email>
+            <auto-register-username>{username}</auto-register-username>
+            <auto-register-email>{email}</auto-register-email>
```
nsoranzo, line 106: Are those changes due to the upgrade to v3.2.0 ? I've been using these values with our Active Directory server for 5 years.

selten: I believe so. I was also using those values for a while, never had any issue. Then I upgraded to 3.2.0 and there were no errors showing up after making these changes. The logging that I did also showed in the dictionary, it only had 'email' and 'username' left in it.

nsoranzo: I've tested on our production Active Directory server with `python-ldap` 3.2.0 on Python 2.7 and 3.6, it works fine except for the fact that on Python3 the returned values are bytes instead of strings.
I've opened a https://github.com/galaxyproject/galaxy/pull/8823 to revert these config changes and fix the Python3 issue.

PR #8621, opened 2019-09-11
[19.09] allow the TS search API to bypass wildcard search

martenson:
and increase default page size on client
and adjust indexing and searching of special keywords to be case-insensitive


martenson:
This is now ready for review.

PR #8627, opened 2019-09-12
Refactor collection state description into a Vue component.

jmchilton:


jmchilton:

PR #8628, opened 2019-09-12
LDAP search member of filter + Username/email generation

BioinformaticsPlatformWIV-ISP:
Hello galaxy team,

This is the new version, generated form the latest dev branch of the pull request #7343 .
It contains only the LDAP related changes,  without the ones belonging to #8424 (already merged) .


lib/galaxy/auth/providers/ldap_ad.py
```diff
@@ -151,6 +156,12 @@ def ldap_search(self, email, username, options):
                 # setup search
                 attributes = [_.strip().format(**params)
                               for _ in options['search-fields'].split(',')]
+                if string_as_bool(options.get('ldap-autogenerate', False)):
+                    if str(username).lower() == 'none':
+		        attributes.append('displayName')
```
nsoranzo, line 161: Please don't use tabs.

lib/galaxy/auth/providers/ldap_ad.py
```diff
@@ -177,7 +188,7 @@ def ldap_search(self, email, username, options):
                     raise ConfigurationError("Missing or mismatching LDAP parameters for %s. Make sure the %s is "
                                              "included in the 'search-fields'." %
                                              (self.role_search_option, self.role_search_attribute))
-                log.critical(params)
+                log.debug(params)
```
nsoranzo, line 191: This can be removed, I was doing that in one of my unfinished branches.

lib/galaxy/auth/providers/ldap_ad.py
```diff
@@ -151,6 +156,12 @@ def ldap_search(self, email, username, options):
                 # setup search
                 attributes = [_.strip().format(**params)
                               for _ in options['search-fields'].split(',')]
+                if string_as_bool(options.get('ldap-autogenerate', False)):
+                    if str(username).lower() == 'none':
```
nsoranzo, line 160:
```suggestion
                    if not username:
```

lib/galaxy/auth/providers/ldap_ad.py
```diff
@@ -208,9 +219,23 @@ def authenticate(self, email, username, password, options):
         attributes = {}
         if self.auto_create_roles_or_groups:
             attributes['roles'] = params[self.role_search_option]
+
+        if string_as_bool(options.get('ldap-autogenerate', False)):
+            if str(username) == 'None':
```
nsoranzo, line 224:
```suggestion
            if not username:
```

lib/galaxy/auth/providers/ldap_ad.py
```diff
@@ -208,9 +219,23 @@ def authenticate(self, email, username, password, options):
...
+                register_username = params['displayName']
+                register_email = _get_subs(options, 'auto-register-email', params)
+            else:
+                register_username = _get_subs(options, 'auto-register-username', params)
+                register_email = params['mail']
+        else:
+	    register_email = _get_subs(options, 'auto-register-email', params)
```
nsoranzo, line 231: Tab

nsoranzo (commented):
It seems that the same can be achieved by correctly configuring `<search-fields>`, `<auto-register-username>` and `<auto-register-email>`, or am I missing something?

BioinformaticsPlatformWIV-ISP:
With the merging and rebasing, a few commits from February this year seem the have been lost. The issue with the auto-registration was already solved then as the code was indeed not necessary. I have reapplied the same changes now.

lib/galaxy/auth/providers/ldap_ad.py
```diff
@@ -255,3 +275,4 @@ class ActiveDirectory(LDAP):
 
 
 __all__ = ('LDAP', 'ActiveDirectory')
+
```
nsoranzo: Can you please remove the blank line for the linter?

nsoranzo (changes requested):
Can you please document `search-memberof-filter` in `config/auth_conf.xml.sample` ?

lib/galaxy/auth/providers/ldap_ad.py
```diff
@@ -169,15 +178,14 @@ def ldap_search(self, email, username, options):
                             # keep role names as list
                             params[self.role_search_option] = attrs[attr]
                         elif attr in attrs:
-                            params[attr] = str(attrs[attr][0])
+                            params[attr] = ' '.join(attrs[attr])
```
nsoranzo: Wouldn't it be better to keep `params['memberOf']` as an array? Spaces can be present in the entries.
That could be implemented by restoring this line and changing line 177 to:
```python
                        if attr == 'memberOf' or (self.role_search_attribute and attr == self.role_search_attribute[1:-1]):
```

BioinformaticsPlatformWIV-ISP: As such, it crash (KeyError exception) when trying to access params['memberOf'] at line 219 : `if search_filter not in params['memberOf']:` . Likely because `self.role_search_option` does'nt have any value at that point. We have dedicate one new 'switch' case to this. In addition since its now a list and we are doing a partial search on it we have to use list comprehension to handle it whereas it was simple 'in' string check in a string before.

nsoranzo: Why do you need a partial search, by the way?

nsoranzo:
@BioinformaticsPlatformWIV-ISP I think the wrong commits were force-pushed today.

nsoranzo:
@BioinformaticsPlatformWIV-ISP The PR is still messed-up. I think you want to go back to commit 82d5b8036f88a036c0ef0dbb873c9015722de5b9 . Let me know if you want me to do it.

lib/galaxy/auth/providers/ldap_ad.py
BioinformaticsPlatformWIV-ISP: In ours case, we are not responsible of the AD node creation, so we can't choose their names. The internal AD naming rules leads to names like '6452_GALAXY_TEST' when we typically want to use matching term such as 'GALAXY_TEST' or eventually 'GALAXY'.  We also use full names (exact match thus), for other galaxy instances.

nsoranzo: So `6452_GALAXY_TEST` is the group name (not username)? Weird!

BioinformaticsPlatformWIV-ISP:
> 
> 
> @BioinformaticsPlatformWIV-ISP The PR is still messed-up. I think you want to go back to commit [82d5b80](https://github.com/galaxyproject/galaxy/commit/82d5b8036f88a036c0ef0dbb873c9015722de5b9) . Let me know if you want me to do it.

Indeed, there was mergin comflicts too, to solve this I have recreate a new branch directly from a clone of the galaxy repo for this PR( #8697) . I will remove the old branch.

PR #8629, opened 2019-09-12
[19.09] Only schedule file watching if an observer exists

mvdbeek:
The observer may not exist if watching is turned off or if the current
thread is not the active watcher thread. This fixes
https://github.com/galaxyproject/planemo/issues/955

### DIFF
--- a/lib/galaxy/tools/toolbox/watcher.py
+++ b/lib/galaxy/tools/toolbox/watcher.py
@@ -162,10 +162,6 @@ class ToolWatcher(BaseWatcher):
         self.tool_dir_callbacks = {}
         self.monitored_dirs = {}
 
-    def monitor(self, dir):
-        if self.observer is not None:
-            self.observer.schedule(self.event_handler, dir, recursive=False)
-
     def watch_file(self, tool_file, tool_id):
         tool_file = os.path.abspath(tool_file)
         self.tool_file_ids[tool_file] = tool_id
--- a/lib/galaxy/tools/toolbox/watcher.py
+++ b/lib/galaxy/tools/toolbox/watcher.py
@@ -163,7 +163,8 @@ class ToolWatcher(BaseWatcher):
         self.monitored_dirs = {}
 
     def monitor(self, dir):
-        self.observer.schedule(self.event_handler, dir, recursive=False)
+        if self.observer is not None:
+            self.observer.schedule(self.event_handler, dir, recursive=False)
 
     def watch_file(self, tool_file, tool_id):
         tool_file = os.path.abspath(tool_file)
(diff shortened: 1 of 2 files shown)


nsoranzo (commented):
Could it be a problem that directories are anyway added to `self.monitored_dirs` if the thread later becomes the one in charge of monitoring?

Also `self.monitored_dirs` should be a set instead of a dict.

mvdbeek:
> Could it be a problem that directories are anyway added to `self.monitored_dirs` if the thread later becomes the one in charge of monitoring?

I think that's a good thing, but we should use those paths when starting a previously inactive watcher to get it into the same state.

nsoranzo:
> > Could it be a problem that directories are anyway added to `self.monitored_dirs` if the thread later becomes the one in charge of monitoring?
> 
> I think that's a good thing, but we should use those paths when starting a previously inactive watcher to get it into the same state.

I'd probably move `self.monitored_dirs[dir_path] = dir_path` inside the `monitor()` method.

lib/galaxy/util/watcher.py
```diff
@@ -71,11 +71,22 @@ def __init__(self, observer_class, even_handler_class, **kwargs):
...
         self.event_handler = even_handler_class(self)
+        self.monitored_dirs = {}
 
     def start(self):
         if self.observer is None:
             self.observer = self.observer_class()
             self.observer.start()
+            self.resume_watching()
+
+    def monitor(self, dir, recursive=False):
```
nsoranzo: `dir` is a standard library built-in function.

```suggestion
    def monitor(self, dir_path, recursive=False):
```

lib/galaxy/util/watcher.py
```diff
@@ -71,11 +71,22 @@ def __init__(self, observer_class, even_handler_class, **kwargs):
...
+        self.monitored_dirs[dir] = recursive
```
nsoranzo:
```suggestion
        self.monitored_dirs[dir_path] = recursive
```

lib/galaxy/util/watcher.py
```diff
@@ -71,11 +71,22 @@ def __init__(self, observer_class, even_handler_class, **kwargs):
...
+        if self.observer is not None:
+            self.observer.schedule(self.event_handler, dir, recursive=recursive)
```
nsoranzo:
```suggestion
            self.observer.schedule(self.event_handler, dir_path, recursive=recursive)
```

lib/galaxy/tools/toolbox/watcher.py
```diff
@@ -162,9 +162,6 @@ def __init__(self, observer_class, even_handler_class, toolbox):
         self.tool_dir_callbacks = {}
         self.monitored_dirs = {}
```
nsoranzo:
```suggestion
```

nsoranzo (commented):
Thanks!

nsoranzo (approved):

PR #8632, opened 2019-09-12
Fix `test_validate_simple_config` unit test

nsoranzo:
Also rebuild config file and its docs.


jdavcs (approved):
Thanks!
(I guess I/we should regenerate these every time the schema is modified.)

PR #8633, opened 2019-09-12
[WIP] Implement jobs_summary API for workflow invocations.

jmchilton:
API backbone used to generate the progress bar for collections but applied to a workflow invocation. Should be helpful in quickly summarizing all the job states in a workflow invocation.

Much more well thought through alternative to the broken #7317 that now offers the ability to summarize across the invocation or summarize each step aggregating across all jobs in a step.

Future steps would be to extend this to work properly with subworkflows and to incorporate WorkflowInvocationStep ``state`` into ``populated_state`` info - since it can encode similar information about whether a step has been expanded yet or not mirroring what ``populated_state`` describes for collections.

### DIFF
--- a/lib/galaxy/managers/jobs.py
+++ b/lib/galaxy/managers/jobs.py
@@ -294,28 +294,101 @@ class JobSearch(object):
         return None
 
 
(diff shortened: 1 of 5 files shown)


jmchilton:

PR #8636, opened 2019-09-12
Alias /api/invocations to /api/workflows/<unused_id>/invocations.

jmchilton:
We never used the workflow_id being sent.


jmchilton:

PR #8637, opened 2019-09-12
Hack around int default for float type in Galaxy schema.

jmchilton:
Causing dev branch CI tests to fail.


jmchilton:
also fixed in https://github.com/galaxyproject/galaxy/pull/8632 - which I didn't see. Let me verify the change in that PR makes sense and I'll close this.

PR #8639, opened 2019-09-12
Modify schema defaults and their loading

jdavcs:
- Modify defaults in the schema so that correct values are loaded (add
  notes to description where applicable).
- When the default in schema was inconcsistent with the default loaded
  in `__init__.py`, use the latter.
- Replace `kwargs.get(foo, default)` with `self.foo` if default is set
  correctly in the schema.
- Remove redundant assignments and variables (there are still more
  left).
- The defalt for `citation_cache_lock_dir` should be `citations/locks`.
- In a few cases where defaults are NOT loaded in `__init__.py`, and the
  values are clearly meant to be suggested values, move those defaults
  into description.

Ref #8493

Most likely will need to be resolved against #8610

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -258,21 +258,20 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
             self.thread_local_log = threading.local()
 
         # Install database related configuration (if different).
-        self.install_database_connection = kwargs.get("install_database_connection", None)
         self.install_database_engine_options = get_database_engine_options(kwargs, model_prefix="install_")
 
         # Where dataset files are stored
-        self.file_path = resolve_path(kwargs.get("file_path", "files"), self.data_dir)
+        self.file_path = resolve_path(self.file_path, self.data_dir)
         # new_file_path and legacy_home_dir can be overridden per destination in job_conf.
-        self.new_file_path = resolve_path(kwargs.get("new_file_path", "tmp"), self.data_dir)
+        self.new_file_path = resolve_path(self.new_file_path, self.data_dir)
         override_tempdir = string_as_bool(kwargs.get("override_tempdir", "True"))
         if override_tempdir:
             tempfile.tempdir = self.new_file_path
         self.shared_home_dir = kwargs.get("shared_home_dir", None)
-        self.openid_consumer_cache_path = resolve_path(kwargs.get("openid_consumer_cache_path", "openid_consumer_cache"), self.data_dir)
+        self.openid_consumer_cache_path = resolve_path(self.openid_consumer_cache_path, self.data_dir)
         self.cookie_path = kwargs.get("cookie_path", None)
-        self.tool_path = resolve_path(kwargs.get("tool_path", "tools"), self.root)
-        self.tool_data_path = resolve_path(kwargs.get("tool_data_path", "tool-data"), os.getcwd())
+        self.tool_path = resolve_path(self.tool_path, self.root)
+        self.tool_data_path = resolve_path(self.tool_data_path, os.getcwd())
         if not running_from_source and kwargs.get("tool_data_path", None) is None:
             self.tool_data_path = resolve_path("tool-data", self.data_dir)
         self.builds_file_path = resolve_path(kwargs.get("builds_file_path", os.path.join(self.tool_data_path, 'shared', 'ucsc', 'builds.txt')), self.root)
@@ -287,7 +286,7 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
             self.integrated_tool_panel_tracking_directory = resolve_path(integrated_tool_panel_tracking_directory, self.root)
         else:
             self.integrated_tool_panel_tracking_directory = None
-        self.toolbox_filter_base_modules = listify(kwargs.get("toolbox_filter_base_modules", "galaxy.tools.filters,galaxy.tools.toolbox.filters"))
+        self.toolbox_filter_base_modules = listify(self.toolbox_filter_base_modules)
         self.tool_filters = listify(kwargs.get("tool_filters", []), do_strip=True)
         self.tool_label_filters = listify(kwargs.get("tool_label_filters", []), do_strip=True)
         self.tool_section_filters = listify(kwargs.get("tool_section_filters", []), do_strip=True)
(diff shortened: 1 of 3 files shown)


lib/galaxy/config/__init__.py
```diff
@@ -262,21 +262,20 @@ def _process_config(self, kwargs):
...
         # Install database related configuration (if different).
-        self.install_database_connection = kwargs.get("install_database_connection", None)
         self.install_database_engine_options = get_database_engine_options(kwargs, model_prefix="install_")
 
         # Where dataset files are stored
-        self.file_path = resolve_path(kwargs.get("file_path", "files"), self.data_dir)
+        self.file_path = resolve_path(self.file_path, self.data_dir)
         # new_file_path and legacy_home_dir can be overridden per destination in job_conf.
-        self.new_file_path = resolve_path(kwargs.get("new_file_path", "tmp"), self.data_dir)
+        self.new_file_path = resolve_path(self.new_file_path, self.data_dir)
```
nsoranzo, line 266: I think there is a serious problem, introduced in #921 (and so present also in the `release_19.09` branch), with resolving relative paths with respect to `self.data_dir` or anything that's not `self.root` .
The issue, which I mentioned already in https://github.com/galaxyproject/galaxy/pull/8490#discussion_r316313454 , is that #921 changed how these settings are interpreted if the admin used a relative path, from being relative to `self.root` to being relative to e.g. `self.data_dir`.

Something like https://github.com/galaxyproject/galaxy/pull/8490/files#diff-46f86ea0d82b84642e4218279ef61442R582 can be used to fix the issue for 19.09 , not sure how to integrate it with the changes in `dev` though.

Ping @natefoo @ic4f @jmchilton

jdavcs: The referenced code doesn't change the way `self.new_file_path` is resolved. The 2 lines are equivalent: 
1. 'tmp' is assigned to `self.new_file_path `as the default:
https://github.com/galaxyproject/galaxy/pull/8639/files#diff-a73e218444bd4a6820a525616a23f40dR186
https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L198
2. It is overwritten by the value in kwargs:
https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L199
3. And all is resolved to `self.data_dir` (which, indeed *may be relative* if a value is provided here:
https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L129 - but that's a separate issue, I think).

Unless I'm mistaken, that's essentially the same fix as in [the edit you pointed to](https://github.com/galaxyproject/galaxy/pull/8490/files#diff-46f86ea0d82b84642e4218279ef61442R582)

The purpose of this refactoring, along with many other similar edits is to simplify all that code before making more significant modifications.

I think we should open a separate issue for `data_dir` resolution (and related stuff). @natefoo and I have been discussing this (path resolution in config) over the past several days. I'm working on this at the moment (and happy to discuss, of course!)

natefoo: We're going to try to address that with something like what I've done [here](https://github.com/natefoo/galaxy/commit/3a3279b497f790b113d6cf2721b64b148d0087fb).

nsoranzo: @ic4f As said above, the issue originates in #921, not here. But this change may make it trickier to "restore" the original defaults.

> Unless I'm mistaken, that's essentially the same fix as in the edit you pointed to

That's not the case, the fix I linked makes the default value resolve w.r.t. `self.data_dir`, while a user-defined value would be resolved w.r.t. to `self.root`, without breaking the old behaviour.

@natefoo I've just had a look at your proposed solution, I think I have ambivalent feelings:
- is it over-complicated?
- is it better to warn admins to update their config file, and have to update our documentation/training, or would it better to keep the old behaviour for user-specified values?

jdavcs: @nsoranzo I see what you mean. However, the 2 lines you refer to in this commit are identical in terms of what they do. I don't think this edit makes it tricker to fix the data_dir resolution issue; quite the contrary, it removes redundant code: kwargs have already been read, and "tmp" has been already used as the default. Without this edit, we are doing those things twice. (and not just new_file_path). 

Resolving a path w.r.t one absolute dir or another can be automated, based on, for example, the value of a `path_resolves_to` attribute specified in the schema. That, and some logic similar to what is done in BaseAppConfiguration (in `_set_config_base`). 

But again, this particular commit, I think, doesn't preclude any of this.

jdavcs:
Rebased.

lib/galaxy/config/__init__.py
jdavcs: @nsoranzo @natefoo : may I mark this as resolved? There are several commits depending on this one. Besides, there's one coming up on path resolution (which is where we can discuss how to account for old configs that may have "database/" and "config/" prepended to user-defined values, if that commit does not offer a satisfactory solution).
If not, that's perfectly fine - I'm happy to discuss and revise/resubmit, etc.

jdavcs: Marking as resolved. I guess.

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -45,14 +45,16 @@ mapping:
 
       database_connection:
         type: str
-        default: sqlite:///./database/universe.sqlite?isolation_level=IMMEDIATE
```
natefoo, line 48: Is the default intentionally removed here and elsewhere?

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -274,13 +290,14 @@ mapping:
 
       conda_prefix:
         type: str
-        default: <tool_dependency_dir>/_conda
         required: false
         desc: |
-          conda_prefix is the location on the filesystem where Conda packages and environments are installed
-          IMPORTANT: Due to a current limitation in conda, the total length of the
+          conda_prefix is the location on the filesystem where Conda packages and environments are
+          installed IMPORTANT: Due to a current limitation in conda, the total length of the
```
natefoo: Pretty sure this limitation hasn't been true since everything was rebuilt with conda-build 2.0 a couple years ago, @bgruening? Not that you added this but while you're modifying it, might as well correct it.

natefoo (commented):
Just a couple minor things and unfortunately it needs another rebase. :cry:

lib/galaxy/webapps/galaxy/config_schema.yml
bgruening: Yes I think we can remove it. if there are still old packages around with this limitation we should rebuild them on conda.

natefoo: Thanks for the confirmation!

jdavcs: Done. Removed starting with "IMPORTANT..."

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Yes. It is reassigning in `config/__init__.py` to the same value but without assuming "database" as the default for "data_dir". If we are no set the default here (which we should, I think), I'd do it after we have the functionality of #8689 (or smth similar) in place. Because there are other similar cases.

natefoo: Got it, I'll merge after the conflict is resolved. Thanks!

jdavcs: Thank you!!!
Resolving (this one and the following ones).

jdavcs:
Rebased.

natefoo (approved):

Issue #8641, opened 2019-09-13
Data Managers in Tool Shed issues


selten:
I'm using Galaxy 19.09 on Python 3.7.

I have a tool shed repository with multiple data managers.

## `data_manager_conf.xml`
These were all added to `data_manager_conf.xml` in the following way:
```xml
<?xml version="1.0"?>
<data_managers>

    <!-- Comment 1 -->
    <data_manager tool_file="data_manager/DM1.xml" id="dm1" version="VERSION">
        <data_table name="DT1">
            <output>
               columns here
            </output>
        </data_table>
    </data_manager>

    <!-- Comment 2 -->
    <data_manager tool_file="data_manager/DM2.xml" id="dm2" version="VERSION">
        <data_table name="DT2">
            <output>
                columns here
            </output>
        </data_table>
    </data_manager>
</data_managers>
```

In the parser, where the exception is thrown, I added a traceback that's shown for debugging purposes, this gave me the following error:
```
tool_shed.galaxy_install.tools.data_manager WARNING 2019-09-12 19:47:44,394 [p:54106,w:1,m:0] [uWSGIWorker1Core1] Encountered unexpected element '<function Comment at 0x7f26eb7d7488>':
<!-- Comment 1 -->
    
galaxy.tools.data_manager.manager DEBUG 2019-09-12 19:47:44,456 [p:54106,w:1,m:0] [uWSGIWorker1Core1] Loaded Data Manager: TOOLSHED/repos/USER/REPO/data_manager/DM2/VERSION
tool_shed.galaxy_install.tools.data_manager WARNING 2019-09-12 19:47:44,456 [p:54106,w:1,m:0] [uWSGIWorker1Core1] Encountered unexpected element '<function Comment at 0x7f26eb7d7488>':
<!-- Comment 2-->
```

I changed them to:
```xml
<?xml version="1.0"?>
<data_managers>
    <data_manager tool_file="data_manager/DM1.xml" id="dm1" version="VERSION">
        <data_table name="DT1">
            <output>
               columns here
            </output>
        </data_table>
    </data_manager>

    <data_manager tool_file="data_manager/DM2.xml" id="dm2" version="VERSION">
        <data_table name="DT2">
            <output>
                columns here
            </output>
        </data_table>
    </data_manager>
</data_managers>
```

## `tool_data_table_conf.xml.sample`
I had a comment on the first line and the rest similar to:
```xml
[...]

PR #8643, opened 2019-09-13
[19.09] Fix issue with xml comment handling on python 3

mvdbeek:
Fixes https://github.com/galaxyproject/galaxy/issues/8641

### DIFF
--- a/lib/galaxy/util/tool_shed/xml_util.py
+++ b/lib/galaxy/util/tool_shed/xml_util.py
@@ -11,16 +11,25 @@ from galaxy.util import (
 log = logging.getLogger(__name__)
 
 
-class Py27CommentedTreeBuilder(XmlET.TreeBuilder):
+class CommentedTreeBuilder(XmlET.TreeBuilder):
+
+    non_comment_seen = False
 
     def doctype(*args):
         # handle deprecation warning for XMLParsing a file with DOCTYPE
         pass
 
+    def start(self, tag, attrib):
+        if tag is not XmlET.Comment:
+            self.non_comment_seen = True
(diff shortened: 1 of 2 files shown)


lib/galaxy/util/tool_shed/xml_util.py
```diff
@@ -11,16 +11,25 @@
...
     def doctype(*args):
         # handle deprecation warning for XMLParsing a file with DOCTYPE
         pass
 
+    def start(self, tag, attrib):
+        if tag is not XmlET.Comment:
+            self.non_comment_seen = True
+        super(CommentedTreeBuilder, self).start(tag, attrib)
+
     def comment(self, data):
```
nsoranzo, line 27: When is this method called? It looks like is has been introduced for Python 3.8?
https://docs.python.org/3.8/library/xml.etree.elementtree.html#xml.etree.ElementTree.TreeBuilder.comment

mvdbeek: I think it's always been called if it existed ? It may just not have been defined previously. If you don't define this you're dropping the comments altogether, and keeping the comments is the point of CommentedTreeBuilder. That's speculation and trying though, I'm not sure how this is implemented.

mvdbeek: fwiw, the implementation is at https://github.com/python/cpython/blob/master/Lib/xml/etree/ElementTree.py, and we may just use the `insert_comments=True` parameter of the TreeBuilder class on python 3

mvdbeek: except that then we couldn't use the workaorund of not adding top level comments

PR #8645, opened 2019-09-13
[19.09]Openlayers viz fixes

dannon:
xref #7888.

shpfile conversion to geojson has been fixed along with several other minor issues, still need to address the refresh (which is happening in a followup PR)


jmchilton:
@dannon would you like this to block the release?

dannon:
@jmchilton Nope.  I'm still working on a little more followup but I don't imagine it'll impact many users, so let's truncate this PR here and get it into the release.  I'll follow up with a separate PR for the other stuff.

martenson:
Using this visualization yields js console errors while it request for non-existing libraries. The same situation is on dev though.

Galaxy

PR #8648, opened 2019-09-13
Simplify config code (minor edits)

jdavcs:
Minor changes, details in commit messages.

(I will rebuild rst and sample yaml files when all these are merged, because, it is likely, some conflict resolution may need to happen first)

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -267,8 +267,10 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
         self.tool_data_path = os.path.join(os.getcwd(), self.tool_data_path)
         if not running_from_source and kwargs.get("tool_data_path", None) is None:
             self.tool_data_path = os.path.join(self.data_dir, "tool-data")
-        self.builds_file_path = os.path.join(self.root, kwargs.get("builds_file_path", os.path.join(self.tool_data_path, 'shared', 'ucsc', 'builds.txt')))
-        self.len_file_path = os.path.join(self.root, kwargs.get("len_file_path", os.path.join(self.tool_data_path, 'shared', 'ucsc', 'chrom')))
+        if self.builds_file_path is None:
+            self.builds_file_path = os.path.join(self.root, self.tool_data_path, 'shared', 'ucsc', 'builds.txt')
+        if self.len_file_path is None:
+            self.len_file_path = os.path.join(self.root, self.tool_data_path, 'shared', 'ucsc', 'chrom')
         # Galaxy OIDC settings.
         self.oidc_config = kwargs.get("oidc_config_file", self.oidc_config_file)
         self.oidc_backends_config = kwargs.get("oidc_backends_config_file", self.oidc_backends_config_file)
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -571,7 +571,6 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
             self.visualization_plugins_directory += ",%s" % ie_dirs
 
         self.proxy_session_map = resolve_path(kwargs.get("dynamic_proxy_session_map", "session_map.sqlite"), self.data_dir)
-
         self.manage_dynamic_proxy = string_as_bool(kwargs.get("dynamic_proxy_manage", "True"))  # Set to false if being launched externally
 
         # InteractiveTools propagator mapping file
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -258,21 +258,20 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
             self.thread_local_log = threading.local()
 
         # Install database related configuration (if different).
-        self.install_database_connection = kwargs.get("install_database_connection", None)
         self.install_database_engine_options = get_database_engine_options(kwargs, model_prefix="install_")
 
         # Where dataset files are stored
-        self.file_path = resolve_path(kwargs.get("file_path", "files"), self.data_dir)
+        self.file_path = resolve_path(self.file_path, self.data_dir)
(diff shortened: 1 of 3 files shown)


jdavcs:
@jmchilton please take a look at involucro/tool_dependency_dir assignment (I removed the check for 'none', but that could be a special case, maybe?)

nsoranzo:
@galaxybot test this

jmchilton:
"tool_dependency_dir" can be "None" - especially from ini. It was the old default and still the easiest way to disable tool dependencies? 

- https://github.com/galaxyproject/galaxy/commit/9f7d921d5d387f0704db66f9e190688122be9170
- https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tool_util/deps/__init__.py#L51

Maybe we should handle this being a proper ``null`` value when coming from the YAML and we should document the option being null better in the config schema, but this does seem to be needed to me at first glance. If ``tool_dependency_dir`` is null, we may want to still use containers (in fact we probably do want to) so we can't use ``tool_dependency_dir``.

Maybe I'm wrong though.

jdavcs:
I see 2 main issues here:
1. Currently we do not have a way to set a config value to None (proper null) if there's a default specified. We should, and this example is a perfect use case: 'dependencies' probably should remain as the default (later resolved to <data_dir>/dependencies), but it should be possible to override it with null.
2. There could be older galaxies with 'none' (as a string) specified for that option (possibly, other options too). Our current setup will break these settings, which shouldn't happen

Here's a possible solution:

1. Allow user to override option defaults with none (case-insensitive, quotes or no quotes). We could add a string_as_none function to lib/util. We *could* specify a 'nullable' schema attribute (and use it along with default/type/desc...) and enable it by default for all options and explicitly disable it where it makes no sense (booleans and numeric types). Or we could say all is nullable, and expect that people won't do stupid things. Or we could enable it by default for all string types (which makes less sense, because we have 'any' type).

2. The above would take care of 'none' values in older configurations. However, if we have more potential cases like this one (not involving 'none'), we could convert such old values before processing them: just group stuff like `if foo=='old special value': foo='new way of specifying that special value` into a method called somewhere here https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L223
That way we'd "clean" all older configs *before* processing them (and the cleaning logic would be all grouped together).

What do you think? @jmchilton

jdavcs:
Rebased.

lib/galaxy/config/__init__.py
```diff
@@ -432,6 +432,10 @@ def _process_config(self, kwargs):
             containers_resolvers_config_file = os.path.join(self.root, containers_resolvers_config_file)
         self.containers_resolvers_config_file = containers_resolvers_config_file
 
+        # tool_dependency_dir can be "none" (in old configs). If so, set it to schema default
+        if self.tool_dependency_dir and self.tool_dependency_dir.lower() == 'none':
```
jdavcs, line 430: @jmchilton This should take care or "none" values from old commits. I'll come up with a better/more generalized solution in a later PR (very soon). This, I hope, is a good enough solution for now. 
(`self.tool_dependency_dir` will be overwritten by whatever value is passed in kwargs, including a string "none", so this will catch it)

jdavcs: Marking as resolved.

lib/galaxy/config/__init__.py
```diff
@@ -435,20 +423,19 @@ def _process_config(self, kwargs):
...
-            if target_dir == "none":
-                target_dir = "dependencies"
-            target_dir = resolve_path(target_dir, self.data_dir)
-            involucro_path = os.path.join(target_dir, "involucro")
-        self.involucro_path = resolve_path(involucro_path, self.root)
-        mulled_channels = kwargs.get('mulled_channels')
-        if mulled_channels:
-            self.mulled_channels = [c.strip() for c in mulled_channels.split(',')]
+        # tool_dependency_dir can be "none" (in old configs). If so, set it to schema default
+        if self.tool_dependency_dir and self.tool_dependency_dir.lower() == 'none':
```
natefoo, line 430: This could be `if string_as_bool_or_none(self.tool_dependency_dir) is None:` (`string_as_bool_or_none` would need to be imported from `galaxy.util` although I'm surprised we don't use it in here already).

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -817,7 +854,10 @@ mapping:
 
       registration_warning_message:
         type: str
-        default: Please register only one account - we provide this service free of charge and have limited computational resources. Multi-accounts are tracked and will be subjected to account termination and data deletion.
+        default: |
+          Please register only one account - we provide this service free of charge and have limited
+          computational resources. Multi-accounts are tracked and will be subjected to account
+          termination and data deletion.
```
natefoo:
```suggestion
          termination and data deletion.
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -845,7 +885,10 @@ mapping:
 
       inactivity_box_content:
         type: str
-        default: Your account has not been activated yet.  Feel free to browse around and see what's available, but you won't be able to upload data or run jobs until you have verified your email address.
+        default: |
+          Your account has not been activated yet.  Feel free to browse around and see what's
+          available, but you won't be able to upload data or run jobs until you have verified your
+          email address.
```
natefoo:
```suggestion
          email address.
```

natefoo (approved):

lib/galaxy/config/__init__.py
nsoranzo: @natefoo Seems a bit misleading to me to use `string_as_bool_or_none()` given that `self.tool_dependency_dir` is normally a path, not a boolean.

natefoo: That's a good point.

PR #8651, opened 2019-09-16
Refactor/cleanup config loading (misc edits)

jdavcs:
Misc. edits. Details/rationale in commit messages.

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -426,11 +426,10 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
         # tool_dependency_dir can be "none" (in old configs). If so, set it to schema default
         if self.tool_dependency_dir and self.tool_dependency_dir.lower() == 'none':
             self.tool_dependency_dir = None
-
         if self.involucro_path is None:
-            self.involucro_path = os.path.join(self.data_dir, self.tool_dependency_dir, "involucro")
+            target_dir = self.tool_dependency_dir or self.appschema['tool_dependency_dir'].get('default')
+            self.involucro_path = os.path.join(self.data_dir, target_dir, "involucro")
         self.involucro_path = os.path.join(self.root, self.involucro_path)
-
         if self.mulled_channels:
             self.mulled_channels = [c.strip() for c in self.mulled_channels.split(',')]
 
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -266,7 +266,7 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
         self.tool_data_path = os.path.join(self.root, self.tool_data_path)
         if not running_from_source and kwargs.get("tool_data_path", None) is None:
             self.tool_data_path = os.path.join(self.data_dir, "tool-data")
-        if self.builds_file_path is None:
+        self.builds_file_path = os.path.join(self.tool_data_path, self.builds_file_path)
             self.builds_file_path = os.path.join(self.root, self.tool_data_path, 'shared', 'ucsc', 'builds.txt')
         if self.len_file_path is None:
             self.len_file_path = os.path.join(self.root, self.tool_data_path, 'shared', 'ucsc', 'chrom')
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -354,16 +354,16 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
         activation_email = kwargs.get('activation_email', None)
         self.email_from = kwargs.get('email_from', activation_email)
         self.myexperiment_target_url = kwargs.get('my_experiment_target_url', 'www.myexperiment.org')
+
         #  Get the disposable email domains blacklist file and its contents
-        self.blacklist_location = kwargs.get('blacklist_file', None)
         self.blacklist_content = None
-        if self.blacklist_location is not None:
-            self.blacklist_file = os.path.join(self.root, kwargs.get('blacklist_file', None))
-            try:
-                with open(self.blacklist_file) as blacklist:
-                    self.blacklist_content = [line.rstrip() for line in blacklist.readlines()]
-            except IOError:
-                log.error("CONFIGURATION ERROR: Can't open supplied blacklist file from path: " + str(self.blacklist_file))
+        self.blacklist_file = os.path.join(self.root, self.blacklist_file)
+        try:
+            with open(self.blacklist_file) as f:
+                self.blacklist_content = [line.rstrip() for line in f]
+        except IOError:
+            log.error("CONFIGURATION ERROR: Can't open supplied blacklist file from path: " + str(self.blacklist_file))
+
         self.persistent_communication_rooms = listify(kwargs.get("persistent_communication_rooms", []), do_strip=True)
         # The transfer manager and deferred job queue
         self.enable_beta_job_managers = string_as_bool(kwargs.get('enable_beta_job_managers', 'False'))
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -29,7 +29,6 @@ from galaxy.containers import parse_containers_config
 from galaxy.exceptions import ConfigurationError
 from galaxy.model import mapping
 from galaxy.model.tool_shed_install.migrate.check import create_or_verify_database as tsi_create_or_verify_database
-from galaxy.tool_util.deps.container_resolvers.mulled import DEFAULT_CHANNELS
 from galaxy.util import (
     ExecutionTimer,
     listify,
@@ -436,8 +435,6 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
 
         if self.mulled_channels:
             self.mulled_channels = [c.strip() for c in self.mulled_channels.split(',')]
-        else:
-            self.mulled_channels = DEFAULT_CHANNELS
 
         default_job_resubmission_condition = kwargs.get('default_job_resubmission_condition', '')
         if not default_job_resubmission_condition.strip():
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -336,7 +336,7 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
         self.template_path = os.path.join(self.root, kwargs.get("template_path", "templates"))
         self.template_cache = os.path.join(self.data_dir, self.template_cache_path)
         self.job_queue_cleanup_interval = int(kwargs.get("job_queue_cleanup_interval", "5"))
-        self.cluster_files_directory = self.resolve_path(kwargs.get("cluster_files_directory", os.path.join(self.data_dir, "pbs")))
+        self.cluster_files_directory = self.resolve_path(kwargs.get("cluster_files_directory"), os.path.join(self.data_dir, "pbs"))
 
         # Fall back to legacy job_working_directory config variable if set.
         self.jobs_directory = os.path.join(self.data_dir, kwargs.get("jobs_directory", self.job_working_directory))
@@ -348,7 +348,7 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
         # should probably drop the backward compatiblity to save the path check.
         self.container_image_cache_path = os.path.join(self.data_dir, kwargs.get("container_image_cache_path", "container_images"))
         if not os.path.exists(self.container_image_cache_path):
-            self.container_image_cache_path = self.resolve_path(kwargs.get("container_image_cache_path", os.path.join(self.data_dir, "container_cache")))
+            self.container_image_cache_path = self.resolve_path(kwargs.get("container_image_cache_path"), os.path.join(self.data_dir, "container_cache"))
         self.output_size_limit = int(kwargs.get('output_size_limit', 0))
         # activation_email was used until release_15.03
         activation_email = kwargs.get('activation_email', None)
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -296,11 +296,11 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
 
         self.password_expiration_period = timedelta(days=int(kwargs.get("password_expiration_period", 0)))
 
-        self.shed_tool_data_path = kwargs.get("shed_tool_data_path", None)
         if self.shed_tool_data_path:
             self.shed_tool_data_path = os.path.join(self.root, self.shed_tool_data_path)
         else:
             self.shed_tool_data_path = self.tool_data_path
+
         self.running_functional_tests = string_as_bool(kwargs.get('running_functional_tests', False))
         self.enable_tool_shed_check = string_as_bool(kwargs.get('enable_tool_shed_check', False))
         if isinstance(self.hours_between_check, string_types):
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -557,11 +557,9 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
         # Statistics and profiling with statsd
         self.statsd_host = kwargs.get('statsd_host', '')
 
-        ie_dirs = kwargs.get('interactive_environment_plugins_directory', None)
+        ie_dirs = self.interactive_environment_plugins_directory
         self.gie_dirs = [d.strip() for d in (ie_dirs.split(",") if ie_dirs else [])]
-        if ie_dirs and not self.visualization_plugins_directory:
-            self.visualization_plugins_directory = ie_dirs
-        elif ie_dirs:
+        if ie_dirs:
             self.visualization_plugins_directory += ",%s" % ie_dirs
 
         self.proxy_session_map = os.path.join(self.data_dir, self.dynamic_proxy_session_map)
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -291,8 +291,8 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
         self.user_tool_section_filters = listify(kwargs.get("user_tool_section_filters", []), do_strip=True)
         self.has_user_tool_filters = bool(self.user_tool_filters or self.user_tool_label_filters or self.user_tool_section_filters)
 
-        self.tour_config_dir = os.path.join(self.root, kwargs.get("tour_config_dir", "config/plugins/tours"))
-        self.webhooks_dirs = os.path.join(self.root, kwargs.get("webhooks_dir", "config/plugins/webhooks"))
+        self.tour_config_dir = os.path.join(self.root, self.tour_config_dir)
+        self.webhooks_dirs = os.path.join(self.root, self.webhooks_dir)
 
         self.password_expiration_period = timedelta(days=int(kwargs.get("password_expiration_period", 0)))
 
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -267,7 +267,7 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
         if not running_from_source and kwargs.get("tool_data_path", None) is None:
             self.tool_data_path = os.path.join(self.data_dir, "tool-data")
         self.builds_file_path = os.path.join(self.tool_data_path, self.builds_file_path)
-        if self.len_file_path is None:
+        self.len_file_path = os.path.join(self.tool_data_path, self.len_file_path)
             self.len_file_path = os.path.join(self.root, self.tool_data_path, 'shared', 'ucsc', 'chrom')
         # Galaxy OIDC settings.
         self.oidc_config = kwargs.get("oidc_config_file", self.oidc_config_file)
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -268,7 +268,6 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
             self.tool_data_path = os.path.join(self.data_dir, "tool-data")
         self.builds_file_path = os.path.join(self.tool_data_path, self.builds_file_path)
         self.len_file_path = os.path.join(self.tool_data_path, self.len_file_path)
-            self.len_file_path = os.path.join(self.root, self.tool_data_path, 'shared', 'ucsc', 'chrom')
         # Galaxy OIDC settings.
         self.oidc_config = kwargs.get("oidc_config_file", self.oidc_config_file)
         self.oidc_backends_config = kwargs.get("oidc_backends_config_file", self.oidc_backends_config_file)
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -267,7 +267,6 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
         if not running_from_source and kwargs.get("tool_data_path", None) is None:
             self.tool_data_path = os.path.join(self.data_dir, "tool-data")
         self.builds_file_path = os.path.join(self.tool_data_path, self.builds_file_path)
-            self.builds_file_path = os.path.join(self.root, self.tool_data_path, 'shared', 'ucsc', 'builds.txt')
         if self.len_file_path is None:
             self.len_file_path = os.path.join(self.root, self.tool_data_path, 'shared', 'ucsc', 'chrom')
         # Galaxy OIDC settings.
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -364,7 +364,6 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
                     self.blacklist_content = [line.rstrip() for line in blacklist.readlines()]
             except IOError:
                 log.error("CONFIGURATION ERROR: Can't open supplied blacklist file from path: " + str(self.blacklist_file))
-        self.smtp_ssl = kwargs.get('smtp_ssl', None)
         self.persistent_communication_rooms = listify(kwargs.get("persistent_communication_rooms", []), do_strip=True)
         # The transfer manager and deferred job queue
         self.enable_beta_job_managers = string_as_bool(kwargs.get('enable_beta_job_managers', 'False'))
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -357,12 +357,13 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
 
         #  Get the disposable email domains blacklist file and its contents
         self.blacklist_content = None
-        self.blacklist_file = os.path.join(self.root, self.blacklist_file)
-        try:
-            with open(self.blacklist_file) as f:
-                self.blacklist_content = [line.rstrip() for line in f]
-        except IOError:
-            log.error("CONFIGURATION ERROR: Can't open supplied blacklist file from path: " + str(self.blacklist_file))
+        if self.blacklist_file:
+            self.blacklist_file = os.path.join(self.root, self.blacklist_file)
+            try:
+                with open(self.blacklist_file) as f:
+                    self.blacklist_content = [line.rstrip() for line in f]
+            except IOError:
+                log.error("CONFIGURATION ERROR: Can't open supplied blacklist file from path: " + str(self.blacklist_file))
 
         self.persistent_communication_rooms = listify(kwargs.get("persistent_communication_rooms", []), do_strip=True)
         # The transfer manager and deferred job queue
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -425,7 +425,7 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
 
         # tool_dependency_dir can be "none" (in old configs). If so, set it to schema default
         if self.tool_dependency_dir and self.tool_dependency_dir.lower() == 'none':
-            self.tool_dependency_dir = self.appschema['tool_dependency_dir'].get('default')
+            self.tool_dependency_dir = None
 
         if self.involucro_path is None:
             self.involucro_path = os.path.join(self.data_dir, self.tool_dependency_dir, "involucro")
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -336,7 +336,7 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
         self.template_path = os.path.join(self.root, kwargs.get("template_path", "templates"))
         self.template_cache = os.path.join(self.data_dir, self.template_cache_path)
         self.job_queue_cleanup_interval = int(kwargs.get("job_queue_cleanup_interval", "5"))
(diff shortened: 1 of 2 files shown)


ic4f:
@galaxybot test this

jdavcs:
Rebased.

lib/galaxy/config/__init__.py
```diff
@@ -339,40 +333,37 @@ def _process_config(self, kwargs):
...
             self.container_image_cache_path = self.resolve_path(kwargs.get("container_image_cache_path", os.path.join(self.data_dir, "container_cache")))
         self.output_size_limit = int(kwargs.get('output_size_limit', 0))
         # activation_email was used until release_15.03
         activation_email = kwargs.get('activation_email', None)
         self.email_from = kwargs.get('email_from', activation_email)
         self.myexperiment_target_url = kwargs.get('my_experiment_target_url', 'www.myexperiment.org')
-        self.instance_resource_url = kwargs.get('instance_resource_url', None)
+
         #  Get the disposable email domains blacklist file and its contents
-        self.blacklist_location = kwargs.get('blacklist_file', None)
```
nsoranzo, line 358: The `config/disposable_email_blacklist.conf.sample` sample file is not usable as it is, so the default for `blacklist_file` is actually `None`.
With these changes, `ERROR:galaxy.config:CONFIGURATION ERROR: Can't open supplied blacklist file from path: /home/circleci/repo/config/disposable_email_blacklist.conf` will always be logged at startup, see e.g. https://circleci.com/gh/galaxyproject/galaxy/38709

I see 2 possible solutions:
- change the default in `lib/galaxy/webapps/galaxy/config_schema.yml` to null
- comment out the present content of `config/disposable_email_blacklist.conf.sample` and update the code below to skip lines starting with `#`

jdavcs:
Rebased to accommodate new commits to #8648 (addressing reviewer comment)

lib/galaxy/config/__init__.py
jdavcs: I'll remove the default from config_schema.yaml - that will resolve to None automatically. I think it's simple and clean. (+ an edit in `config/__init__.py` to prevent `os.path.join` on None)
Thanks!

ic4f:
@galaxybot test this

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -823,15 +860,15 @@ mapping:
 
       instance_resource_url:
         type: str
-        default: https://galaxyproject.org/
         required: false
         desc: |
           URL of the support resource for the galaxy instance.  Used in activation
           emails.
 
+          Sample default 'https://galaxyproject.org/'
```
```suggestion
          Example value 'https://galaxyproject.org/'
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -823,15 +860,15 @@ mapping:
...
+
       blacklist_file:
         type: str
-        default: config/disposable_email_blacklist.conf
```
nsoranzo, line 848: I'd re-add this as "Example value" below.

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -454,6 +471,9 @@ mapping:
           container resolver isn't enabled, and will install on demand unless
           involucro_auto_init is set to false.
 
+          Sample default 'database/dependencies/involucro' where 'database/dependencies' is the
```
```suggestion
          Sample default '<tool_dependency_dir>/involucro'
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -454,6 +471,9 @@ mapping:
...
+          default value of the 'tool_dependency_dir' option.
```
```suggestion
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -551,18 +571,22 @@ mapping:
 
       builds_file_path:
         type: str
-        default: tool-data/shared/ucsc/builds.txt
         required: false
         desc: |
-          File containing old-style genome builds
+          File containing old-style genome builds.
+
+          Sample default 'tool-data/shared/ucsc/builds.txt' where 'tool-data' is the default value
```
```suggestion
          Sample default '<tool_data_path>/shared/ucsc/builds.txt'
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -551,18 +571,22 @@ mapping:
...
+          of the 'tool_data_path' option.
```
```suggestion
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -551,18 +571,22 @@ mapping:
...
 
       len_file_path:
         type: str
-        default: tool-data/shared/ucsc/chrom
         required: false
         desc: |
           Directory where chrom len files are kept, currently mainly used by trackster
 
+          Sample default 'tool-data/shared/ucsc/chrom' where 'tool-data' is the default value of the
```
```suggestion
          Sample default '<tool_data_path>/shared/ucsc/chrom'
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -551,18 +571,22 @@ mapping:
...
+          'tool_data_path' option.
```
```suggestion
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -610,8 +632,7 @@ mapping:
           config/plugins/interactive_environments to load Galaxy's stock plugins.
           These will require Docker to be configured and have security considerations,
           so proceed with caution. The path is relative to the Galaxy root dir.  To use
-          an absolute path begin the path with '/'.  This is a comma
-          separated list.
+          an absolute path begin the path with '/'.  This is a comma separated list.
```
```suggestion
          an absolute path begin the path with '/'.  This is a comma-separated list.
```

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Sounds good, and I assume you meant this for all instances of 'Sample default' (correct me if you meant just this one)

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Done

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Done.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Done. (as per previous comment/edit)

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Done.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Done.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Done.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Done.

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -464,11 +484,12 @@ mapping:
 
       mulled_channels:
         type: str
-        default: 'conda-forge,bioconda'
```
nsoranzo: I think this is actually the default value:

https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L453

If you restore this, you can simplify `lib/galaxy/config/__init__.py` I think.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Yes. Compound adjective before a noun. Done.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: It's one of the edits I wasn't sure about. If we do this (I agree, it makes perfect sense), we should also remove this: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tool_util/deps/mulled/mulled_build.py#L42
and 8 occurrences in 3 other files. If you agree, I'll do this as a separate PR.

jdavcs: and instead of those occurrences, simply use the value loaded at startup (i.e., the schema default unless overwritten) that is stored in app.config

nsoranzo: I don't think you can remove it from `lib/galaxy/tool_util/deps/mulled/mulled_build.py`, that's part of galaxy-tool-util package which is used outside of Galaxy (and so doesn't use Galaxy config files).

jdavcs: Oh, I see. So those channels would be specified in 2 places.. On the other hand that gets rid of an import; and that value, I suppose, would very rarely need to be modified, if ever. Yes, I guess that's a reasonable tradeoff. Changing.

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -274,13 +290,14 @@ mapping:
 
       conda_prefix:
         type: str
-        default: <tool_dependency_dir>/_conda
```
nsoranzo: I was thinking, should this instead be set to `default: _conda`, to be interpreted w.r.t. to `self.tool_dependency_dir` ?

Same goes for `tool_dependency_cache_dir` and `involucro_path`

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -551,18 +569,20 @@ mapping:
 
       builds_file_path:
         type: str
-        default: tool-data/shared/ucsc/builds.txt
```
nsoranzo: Similarly, this could be `default: shared/ucsc/builds.txt`, to be interpreted w.r.t. `self.tool_data_path`

Same goes for `len_file_path`

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Yes. And I believe there are more like that. I haven't touched them for the sake of keeping this PR small enough. I'd rather make such additional changes in a follow-up. I'll make this edit, but I'd rather wait with similar edits until this one gets merged.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Same as above.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Correction. I do prefer to leave this until these are merged. `conda_prefix` is not used in `config/__Init__.py`, but it is used (and resolved) in `tool_util/deps/resolvers/conda.py` and `tool_util/deps/conda_util.py`. If/when we set this default, we should also revise how that value is resolved in those two files. I believe this should be left for a different PR. I'm adding it to my TODO list.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: These two done.

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -551,17 +569,18 @@ mapping:
...
+        default: shared/ucsc/builds.txt
         required: false
         desc: |
-          File containing old-style genome builds
+          File containing old-style genome builds. Value will be resolved to
+          '<tool_data_path>/shared/ucsc/builds.txt'.
 
       len_file_path:
         type: str
-        default: tool-data/shared/ucsc/chrom
```
nsoranzo: Add:
```yaml
        default: shared/ucsc/chrom
```

jdavcs: well, that was stupid. Thanks. Done, pushing.

lib/galaxy/config/__init__.py
```diff
@@ -258,36 +250,37 @@ def _process_config(self, kwargs):
...
         if override_tempdir:
             tempfile.tempdir = self.new_file_path
         self.shared_home_dir = kwargs.get("shared_home_dir", None)
-        self.openid_consumer_cache_path = resolve_path(kwargs.get("openid_consumer_cache_path", "openid_consumer_cache"), self.data_dir)
+        self.openid_consumer_cache_path = os.path.join(self.data_dir, self.openid_consumer_cache_path)
         self.cookie_path = kwargs.get("cookie_path", None)
-        self.tool_path = resolve_path(kwargs.get("tool_path", "tools"), self.root)
-        self.tool_data_path = resolve_path(kwargs.get("tool_data_path", "tool-data"), os.getcwd())
+        self.tool_path = os.path.join(self.root, self.tool_path)
+        self.tool_data_path = os.path.join(os.getcwd(), self.tool_data_path)
```
```suggestion
        self.tool_data_path = os.path.join(self.root, self.tool_data_path)
```

jdavcs: Done.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Marking as resolved for now, to cleanup this PR. Will make this change in a follow-up. (if you'd rather leave this as not resolved for now, that's fine - please reopen.)

lib/galaxy/config/__init__.py
```diff
@@ -258,36 +250,37 @@ def _process_config(self, kwargs):
...
-        self.tool_path = resolve_path(kwargs.get("tool_path", "tools"), self.root)
-        self.tool_data_path = resolve_path(kwargs.get("tool_data_path", "tool-data"), os.getcwd())
+        self.tool_path = os.path.join(self.root, self.tool_path)
+        self.tool_data_path = os.path.join(self.root, self.tool_data_path)
         if not running_from_source and kwargs.get("tool_data_path", None) is None:
-            self.tool_data_path = resolve_path("tool-data", self.data_dir)
-        self.builds_file_path = resolve_path(kwargs.get("builds_file_path", os.path.join(self.tool_data_path, 'shared', 'ucsc', 'builds.txt')), self.root)
-        self.len_file_path = resolve_path(kwargs.get("len_file_path", os.path.join(self.tool_data_path, 'shared', 'ucsc', 'chrom')), self.root)
+            self.tool_data_path = os.path.join(self.data_dir, "tool-data")
+        if self.builds_file_path is None:
```
```suggestion
        self.builds_file_path = os.path.join(self.tool_data_path, self.builds_file_path)
```

lib/galaxy/config/__init__.py
```diff
@@ -258,36 +250,37 @@ def _process_config(self, kwargs):
...
+            self.builds_file_path = os.path.join(self.root, self.tool_data_path, 'shared', 'ucsc', 'builds.txt')
```
```suggestion
```

lib/galaxy/config/__init__.py
```diff
@@ -258,36 +250,37 @@ def _process_config(self, kwargs):
...
+        if self.len_file_path is None:
```
```suggestion
        self.len_file_path = os.path.join(self.tool_data_path, self.len_file_path)
```

lib/galaxy/config/__init__.py
```diff
@@ -258,36 +250,37 @@ def _process_config(self, kwargs):
...
+            self.len_file_path = os.path.join(self.root, self.tool_data_path, 'shared', 'ucsc', 'chrom')
```
```suggestion
```

lib/galaxy/config/__init__.py
```diff
@@ -435,22 +422,19 @@ def _process_config(self, kwargs):
...
-            involucro_path = os.path.join(target_dir, "involucro")
-        self.involucro_path = resolve_path(involucro_path, self.root)
-        mulled_channels = kwargs.get('mulled_channels')
-        if mulled_channels:
-            self.mulled_channels = [c.strip() for c in mulled_channels.split(',')]
-        else:
-            self.mulled_channels = DEFAULT_CHANNELS
+        # tool_dependency_dir can be "none" (in old configs). If so, set it to schema default
+        if self.tool_dependency_dir and self.tool_dependency_dir.lower() == 'none':
+            self.tool_dependency_dir = self.appschema['tool_dependency_dir'].get('default')
```
nsoranzo: According to `tool_dependency_dir` documentation in `lib/galaxy/webapps/galaxy/config_schema.yml` , it seems `None` and `none` can be used to disable tool dependency resolution, see also https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tool_util/deps/__init__.py#L51-L52

Ping @jmchilton 

```suggestion
            self.tool_dependency_dir = None
```

nsoranzo:
It's getting difficult for me as reviewer (and I suppose also for you due to the rebases) to keep track of the stacked PRs.
Would it make sense to close the old ones and concentrate on one?

natefoo:
I merged the other PRs, so it should now be possible to rebase this on dev and only retain the commits that are unique.

jdavcs:
> I merged the other PRs, so it should now be possible to rebase this on dev and only retain the commits that are unique.

Thank you! And I'm addressing the comments for this PR at the moment.
Rebasing now....

lib/galaxy/config/__init__.py
jdavcs: I'll commit this suggestion, but here's my reasoning for testing for None.
Currently, this cannot be None. However, if it is, `os.path.join` will, of course, complain. And here's how it could become not None: although a default value is currently assigned, we should be able to override that with a null (see https://github.com/galaxyproject/galaxy/pull/8648#issuecomment-532300644 + preseeding comment by @jmchilton).  I will implement this in a follow-up PR. Also, once we iron out #8651, most paths requiring resolving will be handled differently. The old version had the default hardcoded in this file, so it was safe. I'm OK not checking for None (for now), as this will be rewritten very soon. (Please correct me if this reasoning is incorrect.)

lib/galaxy/config/__init__.py
jdavcs: Same as previous edit.

lib/galaxy/config/__init__.py
jdavcs: Yes, this makes sense. Not sure what I was thinking.

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -1637,7 +1634,7 @@ mapping:
           Heartbeat log filename. Can accept the template variables {server_name} and
           {pid}
 
-          Sample default 'heartbeat_{server_name}.log'
+          Example value 'heartbeat_{server_name}.log'
```
```suggestion
          Sample default 'heartbeat_{server_name}.log'
```

lib/galaxy/config/__init__.py
jdavcs: It looks like this change broke some integration tests: in the old version tool_dependency_dir was never None, but now it can be - and so `os.path.join` got a NoneType as a result. I'll fix this today.

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -52,7 +52,7 @@ mapping:
           instead.  This string takes many options which are explained in detail in the
           config file documentation.
 
-          Sample default 'sqlite:///./database/universe.sqlite?isolation_level=IMMEDIATE'
+          Example value 'sqlite:///./database/universe.sqlite?isolation_level=IMMEDIATE'
```
```suggestion
          Sample default 'sqlite:///<data_dir>/universe.sqlite?isolation_level=IMMEDIATE'
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -52,7 +52,7 @@ mapping:
...
           where './database' is the resolved default value of the 'data_dir' option).
```
```suggestion
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -137,7 +137,7 @@ mapping:
           separate the tool shed install database (all other options listed above
           but prefixed with install_ are also available).
 
-          Sample default 'sqlite:///./database/universe.sqlite?isolation_level=IMMEDIATE'
+          Example value 'sqlite:///./database/universe.sqlite?isolation_level=IMMEDIATE'
```
```suggestion
          Defaults to the value of the 'database_connection' option.
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -295,7 +295,7 @@ mapping:
           conda_prefix is the location on the filesystem where Conda packages and environments are
           installed.
 
-          Sample default '<tool_dependency_dir>/_conda'
+          Example value '<tool_dependency_dir>/_conda'
```
```suggestion
          Sample default '<tool_dependency_dir>/_conda'
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -376,7 +376,7 @@ mapping:
           By default the tool_dependency_cache_dir is the _cache directory
           of the tool dependency directory.
 
-          Sample default '<tool_dependency_dir>/_cache'
+          Example value '<tool_dependency_dir>/_cache'
```
```suggestion
          Sample default '<tool_dependency_dir>/_cache'
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -470,8 +470,7 @@ mapping:
           container resolver isn't enabled, and will install on demand unless
           involucro_auto_init is set to false.
 
-          Sample default 'database/dependencies/involucro' where 'database/dependencies' is the
-          default value of the 'tool_dependency_dir' option.
+          Example value '<tool_dependency_dir>/involucro'
```
```suggestion
          Sample default '<tool_dependency_dir>/involucro'
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -571,21 +568,19 @@ mapping:
...
         type: str
+        default: shared/ucsc/builds.txt
         required: false
         desc: |
-          File containing old-style genome builds.
-
-          Sample default 'tool-data/shared/ucsc/builds.txt' where 'tool-data' is the default value
-          of the 'tool_data_path' option.
+          File containing old-style genome builds. Value will be resolved to
+          '<tool_data_path>/shared/ucsc/builds.txt'.
```
```suggestion
          with respect to <tool_data_path>.
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -571,21 +568,19 @@ mapping:
...
         type: str
+        default: shared/ucsc/chrom
         required: false
         desc: |
-          Directory where chrom len files are kept, currently mainly used by trackster
-
-          Sample default 'tool-data/shared/ucsc/chrom' where 'tool-data' is the default value of the
-          'tool_data_path' option.
+          Directory where chrom len files are kept, currently mainly used by trackster. Value will
+          be resolved to '<tool_data_path>/shared/ucsc/chrom'.
```
```suggestion
          be resolved with respect to <tool_data_path>.
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -672,12 +665,15 @@ mapping:
...
       cluster_files_directory:
         type: str
-        default: database/pbs
+        default: pbs
         required: false
         desc: |
           If using a cluster, Galaxy will write job scripts and stdout/stderr to this
           directory.
 
+          Default value will be resolved to 'database/pbs' where 'database' is the
```
```suggestion
          Value will be resolved with respect to <data_dir>.
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -672,12 +665,15 @@ mapping:
...
+          default value of the 'data_dir' option).
```
```suggestion
```

nsoranzo (commented):
I think that "Sample default" is better for values that cannot be set in `config_schema.yml` because they are not static, but there is actually a default value.
On the other hard, "Example value" is better when there is no default value (static or calculated).

lib/galaxy/config/__init__.py
nsoranzo: Also, after this line, should `self.tool_dependency_dir` be resolved w.r.t. to `self.data_dir` if it is not None?
At the moment it's only used for the involucro subdir, but maybe better to have it fixed if it is more used in the future?

lib/galaxy/config/__init__.py
nsoranzo: I don't get what's the use case for allowing these to be `None`, to be honest. I'd not go down that route if it's not necessary.

lib/galaxy/config/__init__.py
jdavcs: Very possibly. But I'd rather not do it here. I don't understand its use well enough, and it is accessed + used for path resolving in multiple files. It was not resolved here prior to these changes, so I'd rather leave it as is for now: I am simply not sure I wouldn't be breaking something. Also, I'd rather not introduce new changes in the same PR, if possible.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: I see, that makes sense. I believe some of them may be handled with `path_resolves_to` once that is ready, but for now I'll change them all.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Yes, that's much better. Done.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Done.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Done.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Done.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Done.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Done.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Done.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Done. There may be other instances of "resolved to" versus "resolved with respect to". I agree the latter form is more correct; so that would be another minor todo item for later.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Done.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Done.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: and done.

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -53,7 +53,6 @@ mapping:
           config file documentation.
 
           Sample default 'sqlite:///./database/universe.sqlite?isolation_level=IMMEDIATE'
```
nsoranzo: I think you missed this one:

```suggestion
          Sample default 'sqlite:///<data_dir>/universe.sqlite?isolation_level=IMMEDIATE'
```

jdavcs: Fixing it now.

lib/galaxy/config/__init__.py
```diff
@@ -291,16 +288,16 @@ def _process_config(self, kwargs):
         self.user_tool_section_filters = listify(kwargs.get("user_tool_section_filters", []), do_strip=True)
         self.has_user_tool_filters = bool(self.user_tool_filters or self.user_tool_label_filters or self.user_tool_section_filters)
 
-        self.tour_config_dir = os.path.join(self.root, kwargs.get("tour_config_dir", "config/plugins/tours"))
-        self.webhooks_dirs = os.path.join(self.root, kwargs.get("webhooks_dir", "config/plugins/webhooks"))
+        self.tour_config_dir = os.path.join(self.root, self.tour_config_dir)
+        self.webhooks_dirs = os.path.join(self.root, self.webhooks_dir)
```
nsoranzo, line 292: I just noticed that:
- `webhooks_dir` can be a comma-separated list of directories, so resolving a list w.r.t `self.root` would affect only the first element of the list.
- the assigned attribute `self.webhooks_dirs` has a final `s`, while the config property doesn't

Given that each element of the list is resolved w.r.t. `self.root` any way in https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/util/__init__.py#L1584 , I'd suggest to remove this line:

```suggestion
```

and apply this one-line patch:
```patch
diff --git a/lib/galaxy/app.py b/lib/galaxy/app.py
index bd1de55caa..726c3380f3 100644
--- a/lib/galaxy/app.py
+++ b/lib/galaxy/app.py
@@ -155,7 +155,7 @@ class UniverseApplication(config.ConfiguresGalaxyMixin):
         # Tours registry
         self.tour_registry = ToursRegistry(self.config.tour_config_dir)
         # Webhooks registry
-        self.webhooks_registry = WebhooksRegistry(self.config.webhooks_dirs)
+        self.webhooks_registry = WebhooksRegistry(self.config.webhooks_dir)
         # Load security policy.
         self.security_agent = self.model.security_agent
         self.host_security_agent = galaxy.model.security.HostAgent(
```

lib/galaxy/config/__init__.py
```diff
@@ -354,17 +351,17 @@ def _process_config(self, kwargs):
...
-            self.blacklist_file = os.path.join(self.root, kwargs.get('blacklist_file', None))
+        if self.blacklist_file:
+            self.blacklist_file = os.path.join(self.root, self.blacklist_file)
             try:
-                with open(self.blacklist_file) as blacklist:
-                    self.blacklist_content = [line.rstrip() for line in blacklist.readlines()]
+                with open(self.blacklist_file) as f:
+                    self.blacklist_content = [line.rstrip() for line in f]
             except IOError:
                 log.error("CONFIGURATION ERROR: Can't open supplied blacklist file from path: " + str(self.blacklist_file))
```
nsoranzo, line 363:
```suggestion
                log.error("CONFIGURATION ERROR: Can't open supplied blacklist file from path: %s", self.blacklist_file)
```

lib/galaxy/config/__init__.py
jdavcs: Would you mind if I do this as a separate PR? I can do it right now - will take a couple of minutes.. 
The reason I'm asking is (1) it's not fixing something this PR modified (so the potential list of such edits is very long); and (2) it's ~20 minutes away from passing the last test after which it can be merged - then I can finally rebase 8689. If I make this change, it'll launch all the tests all over again - and that'll take hours.

nsoranzo: Sure, you can address all 3 remaining comments in #8689 or a separate PR.

nsoranzo (approved):

lib/galaxy/config/__init__.py
jdavcs: Thank you! Will do.

jdavcs: I believe the same applies to `tour_registry` (except the trailing 's'). Same thing: comma-separated list, resolved in the same way. I'm changing both.

PR #8657, opened 2019-09-17
Datatypes for MassSpec DIA dlib,elib

jj-umn:
Add datatypes for Data Independent Acquisition (DIA) Mass Spectrometry
analysis.   Dlib and Elib are both Sqlite formats.

Since most SQLite derived galaxy datatypes use table names for
sniffing, I’ve refactored that to use a common method in the base
SQlite class.


jmchilton:
Thanks for the new data types and thanks for the code cleanup of the existing code!

PR #8659, opened 2019-09-17
Add further data manager tests

selten:
Update data manager test IDs and add tests for data table XML files with comments in them.


selten:
Seems like data manager tests aren't running. Also not when triggered using `./run_tests.sh -d`

mvdbeek:
Thanks for the effort, I appreciate wanting to test this. Unfortunately there are some things that I don't think are quite correct.

- run_test.sh -d is meant for installed data managers, and that works correctly.
- if we want to make sure comments are parsed correctly, I don't think we should add another data manager
- the test failures do look related, that's probably the change in the test config table

I think it would be more useful to make sure tables and managers are loaded correctly in a unit test.
If you do want to test the functionality as a whole I'd suggest adding another integration test as in https://github.com/galaxyproject/galaxy/blob/1eb1d947d9a897b6335f32bc42b557de1e59e74e/test/integration/test_data_manager.py#L84

selten:
I'll close this PR then.

PR #8660, opened 2019-09-17
Fix bug in build_dependency_manager()

### LINKED ISSUE #8656: shed requires options to boot it shouldn't
https://github.com/galaxyproject/usegalaxy-playbook/commit/294f464d67bc30a898ce223aad37cd6aa77ff1d5

stacktrace of failed boot: https://gist.github.com/martenson/374e8e24e25034228d085953b9e87a09#file-gistfile1-txt

TS boot should also never go this path I think: `galaxy.tool_util.deps.installable WARNING 2019-09-10 11:57:13,602 [p:31286,w:0,m:0] [MainThread] Conda not installed and auto-installation disabled.`

jdavcs:
This fixes #8656 , so https://github.com/galaxyproject/usegalaxy-playbook/commit/294f464d67bc30a898ce223aad37cd6aa77ff1d5 should be no longer necessary.

The argument default_tool_dependency_dir should not be assigned a value
by default, because sometimes (at least for TS), it should be `None`. If set by default, it will cause this to be always `True`: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tool_util/deps/__init__.py#L77
and, as a result, TS will try to create <data_dir>/dependencies and load conda.

Also, add default value `None` to `get()` methods of each of the webapps
configuration objects (that way those methods behave more like a dictionary's `get()`. 

Note: the default_tool_dependency_dir argument is probably redundant; also, the get() methods can be moved to the base class, etc.; but those belong in separate commits, I think.

Requesting John's review as this modifies 450ae76ddb297312b226d8bf8a2f8ef58afd993f

EDIT: apparently I can't request reviewers using the reviewers feature. So, ping @jmchilton @martenson

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -760,7 +760,7 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
             if explicit:
                 log.warning("Sanitize log file explicitly specified as '%s' but does not exist, continuing with no tools whitelisted.", self.sanitize_whitelist_file)
 
-    def get(self, key, default):
+    def get(self, key, default=None):
         return self.config_dict.get(key, default)
 
     def get_bool(self, key, default):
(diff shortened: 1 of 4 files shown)


jmchilton (approved):

nsoranzo (approved):

martenson (approved):
nice find @ic4f, thanks!

martenson:
when I think about it we should probably backport this

PR #8661, opened 2019-09-17
Extend the list of attributes to be reported.

VJalili:
In addition to attributes reported, report `create_time` for `collection` view, and `create_time`, `ratings`, `reviews`,  and `reviewers` for `element` view.


nsoranzo:
> In addition to attributes reported, report `create_time` for `collection` view, and `create_time`, `ratings`, `reviews`, and `reviewers` for `element` view.

Just a warning: I wouldn't expect many ratings/reviews, that's an almost unused feature of the Tool Shed.

VJalili:
True, though that view is supposed to return as much info as possible for a given entry. If a value for those fields are set, we can use them, otherwise, we can just ignore them.

dannon:
Unrelated test failures, going to go ahead and merge this as it seems reasonable to expose these attributes via the API.

VJalili:
Thanks for merging this.

PR #8662, opened 2019-09-18
Have CI ensure client PRs are formatted.

jmchilton:
Tired of having to ignore and revert unrelated changes when formatting my client changes, this should make sure dev stays clean. People may disagree on whether we should be using prettier - but hopefully we can all agree if we're going to use it we should enforce it for each PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -188,6 +188,16 @@ jobs:
       - run: cd client && yarn install --frozen-lockfile
       - *save_yarn_cache
       - run: cd client && yarn run eslint
+  js_formatted:
+    docker:
+      - image: circleci/node:10-browsers
+    <<: *set_workdir
+    steps:
+      - *restore_yarn_cache
+      - run: cd client && yarn install --frozen-lockfile
+      - *save_yarn_cache
+      - run: cd client && yarn run prettier
+      - run: git diff --exit-code -- client
 workflows:
   version: 2
   get_code_and_test:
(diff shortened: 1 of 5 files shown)


nsoranzo:
Do it make sense to merge it with the `js_lint` job?

jmchilton:
That sounds like an excellent idea.

dannon:
Good idea, thanks!  I'm -0 on the additional Makefile targets, though; I'd rather we encourage usage of the client related actions through the standard package scripts.

martenson:
I see makefile as an educative shortcut that can also teach you how to do it through yarn.

jmchilton:
I use ``make client-watch`` and ``make client-format`` daily currently, obviously I do know how to dip into yarn and cd into client but I prefer not to.

Update: speling fiix

dannon:
@martenson Sure, I guess it cuts both ways -- I think it probably has just as much of an isolation effect keeping people out of the client/package.json guts, but this isn't one I care enough to push for, hence the -0 on that particular bit.  ( to be clear, either way, this is a good and useful change overall)

.circleci/config.yml
```diff
@@ -188,6 +188,13 @@ jobs:
       - run: cd client && yarn install --frozen-lockfile
       - *save_yarn_cache
       - run: cd client && yarn run eslint
+      - run: cd client && yarn run prettier-check
+    steps:
```
martenson: this key is duplicated above

dannon:
The lint will pass once https://github.com/galaxyproject/galaxy/pull/8674 is in.  I PR'd the merge forward since there was a conflict in all the configuration changes recently -- wanted to make sure that got a second pair of eyes.

nsoranzo (approved):

martenson:
@dannon thanks for merging

PR #8664, opened 2019-09-18
[19.09] Bug fix - add missing invocation state tracking, report link.

jmchilton:
This is imperfect and rough but provides a bare minimum implementation of a bunch of important functionality that the absence of represents serious enough usability problems that I think it belongs in 19.09.

For admins, there is no way to track workflow invocations being scheduled currently and we routinely ask people to set the state on the workflow invocation column with SQL directly. This is a serious issue and this PR provides a new list in the admin interface of active invocations with state tracking (for steps scheduled and jobs executed) with a cancel button for each.

The same component that gets embedded in the grid for state tracking is also now added to the landing page after workflows are executed. Each invocation gets its own tracking widget - we don't name workflow invocations so the creation time is used to distinguish between invocations (we could probably iterate on this and do something with inputs?). Even if ephemeral, this provides users a way of tracking workflow invocation progress and notifying them if scheduling fails. Previously we had no such notifications. It also provides them a way to cancel workflow invocations (less of a bug then these other points, but still important).

Once the workflow has finished executing, an invocation report link will appear - again the lack of a link inside of the app to the invocation report added in 19.09. Previously there was no link in the UI for this report.

## Admin Grid

Screen Shot 2019-09-18 at 10 55 50 AM

## Admin Grid with Expanded Details

Screen Shot 2019-09-18 at 10 55 56 AM

## Workflow Ran, Invocation Scheduling

Screen Shot 2019-09-18 at 10 55 38 AM

## Workflow Ran, Invocation Scheduled

Screen Shot 2019-09-18 at 10 56 23 AM

## Workflow Ran, Invocation Cancelled 

Screen Shot 2019-09-18 at 11 05 36 AM

### DIFF
--- a/client/galaxy/scripts/entry/panels/admin-panel.js
+++ b/client/galaxy/scripts/entry/panels/admin-panel.js
@@ -42,6 +42,12 @@ const AdminPanel = Backbone.View.extend({
                         id: "admin-link-jobs"
                     },
                     {
+                        title: _l("Manage workflow invocations"),
+                        url: "admin/invocations",
+                        target: "__use_router__",
+                        id: "admin-link-invocations"
+                    },
+                    {
                         title: _l("Local data"),
                         url: "admin/data_manager",
                         target: "__use_router__",
--- a/client/galaxy/scripts/components/JobStates/CollectionJobStates.vue
+++ b/client/galaxy/scripts/components/JobStates/CollectionJobStates.vue
@@ -1,0 +1,92 @@
+<template>
+    <div v-if="!jobSourceType || jobSourceType == 'Job' || isTerminal">
+        {{ simpleDescription }}
+    </div>
+    <div v-else-if="!jobStatesSummary || !jobStatesSummary.hasDetails()">
+        <div class="progress state-progress">
+            <span class="note"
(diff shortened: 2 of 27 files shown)


client/galaxy/scripts/entry/panels/admin-panel.js
```diff
@@ -41,6 +41,12 @@ const AdminPanel = Backbone.View.extend({
                         target: "__use_router__",
                         id: "admin-link-jobs"
                     },
+                    {
+                        title: _l("Manage workflow invocations"),
```
martenson, line 45:
```suggestion
                        title: _l("Workflow invocations"),
```

martenson (approved):
Works fine in my testing. Thanks @jmchilton !

bgruening:
Uh, this is super cool. I think this will be super useful for @wm75 during the next release cycle!

Thanks John, it was really hard to get to this page before.

bgruening:
All green now and it works for me. This will make testing and improving those reports in the next 3 month way simpler. Thanks a lot!

client/galaxy/scripts/entry/panels/admin-panel.js
jmchilton: Crap - sorry I missed this - I'll fix this in a follow up.

PR #8667, opened 2019-09-18
Get rid of fix timeout in admin app Selenium test.

jmchilton:
Try to avoid these because dockerized, CI tests run slower typically. This test I think is unstable because of this fixed, small timeout. Jobs definitely can take longer than 5 seconds in the tests.


jmchilton:
Those failures are unrelated - can this be merged?

PR #8672, opened 2019-09-18
add .gitattributes file and normalize line endings to use LF instead instead of CRLF.

WilliamHolden:
This pull request adds a git attributes file so that all developers can have shared attributes regardless of their individual development environments. The only line I have added so far is 
``` * text=auto ```
This line ensures that text files that any contributor introduces to the repository have their line endings normalized. 

This pull request also removes any existing windows style CRLF line endings and replaces them with Unix style LF line endings. 

The only change in all of these file is the line endings. (except the .gitattributes file)

Please review and let me know if you see any issues with the change.


jmchilton:
Thanks for the contribution!

With regard to - lib/galaxy/datatypes/test/dosimzml - I'm pretty sure we want that to DOS line endings for the test cases using it?

I'm skeptical of converting the JS libraries from external sources also. I feel like they should be left alone?

Can we have this only affect new files or exclude these files somehow?

dannon:
I agree it'd be nice to just leave those included libraries alone.  Many haven't been touched in a long time and the next step for them is removal from the repo (dynatree in particular, as the components are replaced).

WilliamHolden:
> With regard to - lib/galaxy/datatypes/test/dosimzml - I'm pretty sure we want that to DOS line endings for the test cases using it?

This seems a little strange to me to require CRLF line endings but I'm not too familiar with how this file works. If this file does require CRLF, then it's probably a good idea to specify that in the gitattriutes file too.

> I'm skeptical of converting the JS libraries from external sources also. I feel like they should be left alone?

I think the JS libraries should have been using LF line endings in the beginning. When they were committed initially, they were probably altered to use CRLF from the original. I was just converting them back. It's not that important though since these files don't get touched anyway so I can definitely revert if it makes this feel like a safer change.

> Can we have this only affect new files or exclude these files somehow?

Yes both of these options are possible. Based on your comment above, I will revert the js libraries and test/dosimzml to use CRLF. I'll also edit the gitattributes file to force test/dosimzml to use CRLF. This way if someone edits and commits this specific file later while on linux, it will not autoconvert the file to use LF. 

Let me know if you think that is a good idea. And thanks for reviewing!

WilliamHolden:
> I agree it'd be nice to just leave those included libraries alone. Many haven't been touched in a long time and the next step for them is removal from the repo (dynatree in particular, as the components are replaced).

gotcha. I understand where you guys are coming from. See the comment above and let me know if you think that's a good idea.

Also thanks for reviewing!

WilliamHolden:
closed in favor of https://github.com/galaxyproject/galaxy/pull/8677

PR #8676, opened 2019-09-18
More robust login/logout with Selenium.

jmchilton:
- There was a UI bug with masthead not reflecting being logged in that we needed to work around previously but I'm hoping that is now fixed and this proves to be more robust than that previous sleep command.
- Capture more test context when problems occur during setup (i.e. if ensure_registered is enabled).
- Better error reporting/exception message for login failures.
- Better error message when a Selenium logout failure occurs.

### DIFF
--- a/lib/galaxy/selenium/navigates_galaxy.py
+++ b/lib/galaxy/selenium/navigates_galaxy.py
@@ -427,19 +427,14 @@ class NavigatesGalaxy(HasDriver):
             confirm=confirm
         ))
         self.wait_for_and_click(self.navigation.registration.selectors.submit)
-        # Give the browser a bit of time to submit the request.
-        # It would be good to eliminate this sleep, but it can't be because Galaxy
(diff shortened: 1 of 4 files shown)


jmchilton:
Second run of the selenium tests in this branch and twice without any tests failing even before retrying. Typically even when Selenium has been green over the last year a test or two has needed to be retried internal to the framework - I wonder if more of those were login problems than I thought.

That or I've broken the test dumping 😅.

jmchilton:
@galaxybot test this

jmchilton:
Spoke too soon, some legitimate Selenium failures here. Look unrelated to login at first glance but I'm not sure.

https://jenkins.galaxyproject.org/job/docker-selenium-py3/837/

jmchilton:
Same two tests again...

https://jenkins.galaxyproject.org/job/docker-selenium-py3/842/

martenson (approved):
nice improvements, thanks!

PR #8677, opened 2019-09-18
add .gitattributes file and normalize line endings to use LF in…

WilliamHolden:
Previous PR here: https://github.com/galaxyproject/galaxy/pull/8672


tool-data/shared/ucsc/manual_builds.txt.sample
```diff
@@ -1,708 +1,708 @@
-phiX	phiX174	phiX=5386
```
jmchilton: This is an externally downloaded file as well - I don't think we should modify it.

WilliamHolden: Hmm I would have thought this was a safe change, but I'll trust your intuition and remove it just in case. It just means we will have a few files committed to our repo with CRLF line endings which isn't a huge deal as long as the files aren't altered often. Since this file hasn't been touched in years, I won't add a gitattribute rule for this file.

WilliamHolden: PR has been updated and this file was reverted back to the original on dev.

nsoranzo:
Thanks @WilliamHolden !

PR #8678, opened 2019-09-19
Improve CI reliability by annotating external resources.

jmchilton:
And skipping if github, usegalaxy.org, or the tool shed is down for tests that depend on these.


martenson (approved):
nice idea

PR #8684, opened 2019-09-20
Fixes for BootstrapVue Tables

dannon:
WIP, going through and fixing up some invalid table field/slot definitions.


guerler:
We should fix this also for `Display Applications`, another issue I noticed is that the modal for the `Data dialog` has no fixed height anymore.

dannon:
@guerler I'll check out `Display Applications` now.  Not sure if the modal change is intentional and appropriate, or something that is no longer applied.

edit: Ahh, I see I fixed the fields but didn't notice the missing slot/button there, thanks!

guerler:
Thank you!

PR #8685, opened 2019-09-20
Restrict `/api/histories/shared_with_me` to logged-in users

nsoranzo:
It was returning an empty list anyway for non-authenticated users.

Also differentiate error message for sessionless API request when anonymous access is allowed (i.e. `expose_api_anonymous`).


nsoranzo:
Selenium test failures are unrelated.

jmchilton:
Thanks for this solid clean up @nsoranzo - really appreciated

PR #8646, opened 2019-09-13
Replace resolve_path() with os.path.join()

jdavcs:
This is waiting on https://github.com/galaxyproject/galaxy/pull/8639 (first commit: 2762a9d)

Rationale for modification in commit message.

PR #8687, opened 2019-09-20
[19.09] Restore resolution of relative paths w.r.t. Galaxy root

nsoranzo:
for config files.

Broken in https://github.com/galaxyproject/galaxy/pull/921/ .

For context, see https://github.com/galaxyproject/galaxy/pull/8639#discussion_r325621897

Includes also elimination of `resolve_path` function as in https://github.com/galaxyproject/galaxy/pull/8646

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -94,13 +94,6 @@ LOGGING_CONFIG_DEFAULT = {
 """Default value for logging configuration, passed to :func:`logging.config.dictConfig`"""
 
 
-def resolve_path(path, root):
-    """If 'path' is relative make absolute by prepending 'root'"""
-    if not os.path.isabs(path):
-        path = os.path.join(root, path)
-    return path
-
-
 def find_root(kwargs):
     root = os.path.abspath(kwargs.get('root_dir', '.'))
     return root
@@ -164,13 +157,13 @@ class BaseAppConfiguration(object):
                 setattr(self, var + '_set', True)
             else:
                 for value in values:
-                    if os.path.exists(resolve_path(value, self.root)):
+                    if os.path.exists(os.path.join(self.root, value)):
                         path = value
                         break
                 else:
                     path = values[-1]
                 setattr(self, var + '_set', False)
-            setattr(self, var, resolve_path(path, self.root))
+            setattr(self, var, os.path.join(self.root, path))
 
         for var, values in listify_defaults.items():
             paths = []
@@ -180,7 +173,7 @@ class BaseAppConfiguration(object):
             else:
                 for value in values:
                     for path in listify(value):
-                        if not os.path.exists(resolve_path(path, self.root)):
+                        if not os.path.exists(os.path.join(self.root, path)):
                             break
                     else:
                         paths = listify(value)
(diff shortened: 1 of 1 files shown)


natefoo:
This undoes the work of not assuming everything is relative to the Galaxy root. I have a solution for this [here](https://github.com/natefoo/galaxy/commit/3a3279b497f790b113d6cf2721b64b148d0087fb) that I believe @ic4f is integrating with his recent work to pull the relative directory from the schema.

nsoranzo:
To give a bit more context for these changes:
- #921 would break working installation of Galaxy on update to 19.09 if an admin were using a relative path for e.g. `file_path`
- I think we are too late in the 19.09 release cycle to address this issue with complex changes as the solution proposed by @natefoo . If we decide to keep the new behaviour, I'd prefer to have it ironed out on `dev`
- I see a clear use case for an admin using a path relative to the Galaxy root, e.g. `file_path = ../files` to keep some stuff outside of the git tree. But I fail to see a use case for specifying a path relative to e.g. `data_dir` .

natefoo:
> * I think we are too late in the 19.09 release cycle to address this issue with complex changes as the solution proposed by @natefoo . If we decide to keep the new behaviour, I'd prefer to have it ironed out on `dev`

I can get on board with that.

> * I see a clear use case for an admin using a path relative to the Galaxy root, e.g. `file_path = ../files` to keep some stuff outside of the git tree. But I fail to see a use case for specifying a path relative to e.g. `data_dir` .

Other configs relative to the `config_dir` is the more common case here. But the bigger point is that when Galaxy is installed, the root dir doesn't exist anymore, so nothing should be thought of as relative to it anymore.

jdavcs:
> * I think we are too late in the 19.09 release cycle to address this issue with complex changes as the solution proposed by @natefoo . If we decide to keep the new behaviour, I'd prefer to have it ironed out on `dev`
Same as @natefoo, I agree. Can we not merge this into dev (i.e., only 19.09)? It will be a pain to reverse it all over again.

With regard to the added complexity of the proposed solution: I'm opening a PR in a few minutes; it uses that approach, only generalized it a bit. So some complexity is removed, while some is added. Whether we go with that solution or develop something else, I think we need to automate processing of configuration options to get rid of all  that redundancy.

jdavcs (approved):
I'm OK with this, pending @natefoo approval; also I very much prefer these changes to be limited to the 19.09 branch (nor dev).

nsoranzo:
> Other configs relative to the `config_dir` is the more common case here.

@natefoo Makes sense, but user-specified values for all options with a default path in `config_dir` are presently still resolved w.r.t. to `self.root` !

> But the bigger point is that when Galaxy is installed, the root dir doesn't exist anymore, so nothing should be thought of as relative to it anymore.

I think we can resolve this for `dev` in https://github.com/galaxyproject/galaxy/pull/8689 ?

> also I very much prefer these changes to be limited to the 19.09 branch (not dev).

@ic4f This will need to be carefully merged forward.

natefoo (approved):
Thanks @nsoranzo, looks like we have the path forward worked out.

jdavcs:
> Thanks @nsoranzo, looks like we have the path forward worked out.

> @ic4f This will need to be carefully merged forward.

OK. Is it possible to merge my PRs before this merge? There are 4 that, I think, are ready (and there's another that I'm editing based on @nsoranzo's suggestions). Rebasing one after another is getting difficult and error-prone. So, if possible, I'd like to get them out of the way, especially before we merge these changes. #8639, #8646, #8648 : all tests passed except selenium (unrelated). #8651 : tests running.

nsoranzo:
> OK. Is it possible to merge my PRs before this merge?

Only the forward-merge need to wait, I think this is needed for the 19.09 release.

jdavcs:
@nsoranzo Oh yes, that's what I meant - the forward-merge into dev, not this merge into 19.09.

PR #8689, opened 2019-09-22
Automate path resolution for config options

jdavcs:
See details in commit message.

This allows to:
- add 'path_resolves_to' attributes to properties in the schema (text like "relative path foo will be resolved with respect to bar" can be auto-generated for sample config files and docs)
- for all config options that have 'path_resolves_to' specified, get rid of all `os.path.join/path_resolve` stuff in `config/__init__.py`
- current galaxy installations that have relative paths specified as `config/foo` or `database/foo` will not be broken (warnings will be issued, and path will be resolved automatically - as per @natefoo 's solution)
- path resolution can be specified to arbitrary depth (A > B > ... > root)

Note: path resolution (and schema validation) is handled recursively. However, this doesn't have any effect on performance: both are called once at startup; the graph is trivial: a few vertices (there are only so many paths we need to specify), each with degree=1 or 0 (path_resolves_to can only point to one parent at most), and each vertex is visited once. Of course, this can be done iteratively, but that would make the code less intuitive with no visible performance gain (because the cost of allocating a stack frame for each call is negligible for this n).

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -190,12 +190,17 @@ class BaseAppConfiguration(object):
 class GalaxyAppConfiguration(BaseAppConfiguration):
     deprecated_options = ('database_file', 'track_jobs_in_database')
     default_config_file_name = 'galaxy.yml'
+    # {key: config option, value: deprecated directory name}
+    # If value == first dir in a user path that resolves to key, it will be stripped from the path
+    deprecated_dirs = {'config_dir': 'config', 'data_dir': 'database'}
 
     def __init__(self, **kwargs):
         self._load_schema()  # Load schema from schema definition file
         self._load_config_from_schema()  # Load default propery values from schema
+        self._validate_schema_paths()  # check that paths can be resolved
         self._update_raw_config_from_kwargs(kwargs)  # Overwrite default values passed as kwargs
         self._create_attributes_from_raw_config()  # Create attributes for LOADED properties
+        self._resolve_paths(kwargs)  # Overwrite attributes (not _raw_config) w/resolved paths
         self._process_config(kwargs)  # Finish processing configuration
 
     def _load_schema(self):
@@ -203,24 +208,80 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
         self.appschema = self.schema.app_schema
 
     def _load_config_from_schema(self):
-        self._raw_config = {}  # keeps track of values provided to the app
-        self.reloadable_options = set()
+        self._raw_config = {}  # keeps track of startup values (kwargs or schema default)
+        self.reloadable_options = set()  # config options we can reload at runtime
+        self._paths_to_resolve = {}  # {config option: referenced config option}
         for key, data in self.appschema.items():
             self._raw_config[key] = data.get('default')
             if data.get('reloadable'):
                 self.reloadable_options.add(key)
+            if data.get('path_resolves_to'):
+                self._paths_to_resolve[key] = data.get('path_resolves_to')
+
+    def _validate_schema_paths(self):
+
+        def check_exists(option, key):
+            if not option:
+                message = "Invalid schema: property '{}' listed as path resolution target " \
+                    "for '{}' does not exist".format(resolves_to, key)
+                raise_error(message)
+
+        def check_type_is_str(option, key):
+            if option.get('type') != 'str':
+                message = "Invalid schema: property '{}' should have type 'str'".format(key)
+                raise_error(message)
+
+        def check_is_dag():
+            def walk(key):
+                if key:
+                    if key in visited:
+                        raise_error('Invalid schema: cycle detected')
+                    visited.add(key)
+                    next = self.appschema[key].get('path_resolves_to')
+                    walk(next)
+
+            for key in self._paths_to_resolve:
+                visited = set()
+                walk(key)
+
+        def raise_error(message):
+            log.error(message)
+            raise ConfigurationError(message)
+
+        for key, resolves_to in self._paths_to_resolve.items():
+            parent = self.appschema.get(resolves_to)
+            check_exists(parent, key)
+            check_type_is_str(parent, key)
+            check_type_is_str(self.appschema[key], key)
+            check_is_dag()  # must be called last: walks entire graph
 
     def _update_raw_config_from_kwargs(self, kwargs):
-        type_converters = {
-            'bool': string_as_bool,
-            'int': int,
-            'float': float,
-        }
+
+        def convert_datatype(key, value):
+            datatype = self.appschema[key].get('type')
+            if datatype in type_converters:
+                return type_converters[datatype](value)
+            return value
+
+        def strip_deprecated_dir(key, value):
+            resolves_to = self.appschema[key].get('path_resolves_to')
+            if resolves_to:  # value is a path that will be resolved
+                dir = value.split(os.sep)[0]  # get first directory component
+                if dir == self.deprecated_dirs[resolves_to]:  # dir is deprecated for this option
+                    ignore = dir + os.sep
+                    log.warning(
+                        "Paths for the '%s' option are now relative to '%s', remove the leading '%s' "
+                        "to suppress this warning: %s", key, resolves_to, ignore, value
+                    )
+                    return value[len(ignore):]
+            return value
+
+        type_converters = {'bool': string_as_bool, 'int': int, 'float': float}
+
         for key, value in kwargs.items():
             if key in self.appschema:
-                datatype = self.appschema[key].get('type')
-                if datatype in type_converters:
-                    value = type_converters[datatype](value)
+                value = convert_datatype(key, value)
+                value = strip_deprecated_dir(key, value)
                 self._raw_config[key] = value
 
     def _create_attributes_from_raw_config(self):
@@ -228,6 +289,30 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
             if hasattr(self, key):
                 raise ConfigurationError("Attempting to override existing attribute '%s'" % key)
             setattr(self, key, value)
+
+    def _resolve_paths(self, kwargs):
+
+        def resolve(key):
+            if key in _cache:  # resolve each path only once
+                return _cache[key]
+
+            path = getattr(self, key)  # path prior to being resolved
+            parent = self.appschema[key].get('path_resolves_to')
+            if not parent:  # base case: nothing else needs resolving
+                return path
+            parent_path = resolve(parent)  # recursively resolve parent path
+            if path is not None:
+                path = os.path.join(parent_path, path)  # resolve path
+            else:
+                path = parent_path  # or use parent path
+
+            setattr(self, key, path)  # update property
+            _cache[key] = path  # cache it!
+            return path
+
+        _cache = {}
+        for key in self._paths_to_resolve:
+            resolve(key)
 
     def _process_config(self, kwargs):
         self.config_dict = kwargs
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -200,6 +200,11 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
         self._validate_schema_paths()  # check that paths can be resolved
         self._update_raw_config_from_kwargs(kwargs)  # Overwrite default values passed as kwargs
         self._create_attributes_from_raw_config()  # Create attributes for LOADED properties
+
+        self.config_dict = kwargs
+        self.root = find_root(kwargs)
+        self._set_config_base(kwargs)  # must be called prior to _resolve_paths()
+
         self._resolve_paths(kwargs)  # Overwrite attributes (not _raw_config) w/resolved paths
         self._process_config(kwargs)  # Finish processing configuration
 
@@ -312,10 +317,6 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
             resolve(key)
 
     def _process_config(self, kwargs):
-        self.config_dict = kwargs
-        self.root = find_root(kwargs)
-        self._set_config_base(kwargs)
-
         # Resolve paths of other config files
         self.parse_config_file_options(kwargs)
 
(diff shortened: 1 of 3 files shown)


jdavcs:
Ping @nsoranzo: (I can't request reviewers beyond those suggested by the system). Provided this passes the tests, would like to hear your opinion on this. As I mentioned, I'm happy to rewrite all of it or take a different approach, as long as we address standardizing config processing and getting rid of all redundant code, etc.
Ping @jmchilton : this includes one edit you suggested; it does not include overriding schema defaults with null: I'll do it separately not t0 lump all together.

lib/galaxy/config/__init__.py
```diff
@@ -180,57 +173,142 @@ def _parse_config_file_options(self, defaults, listify_defaults, config_kwargs):
...
+
+        type_converters = {'bool': string_as_bool, 'int': int, 'float': float}
+
         for key, value in kwargs.items():
             if key in self.appschema:
-                datatype = self.appschema[key].get('type')
-                if datatype in type_converters:
-                    value = type_converters[datatype](value)
+                value = convert_datatype(key, value)
+                value = strip_deprecated_dir(key, value)
```
nsoranzo: What about:
```suggestion
                if not running_from_source:
                    value = strip_deprecated_dir(key, value)
```

lib/galaxy/config/__init__.py
```diff
@@ -180,57 +173,142 @@ def _parse_config_file_options(self, defaults, listify_defaults, config_kwargs):
...
+        def convert_datatype(key, value):
+            datatype = self.appschema[key].get('type')
+            if datatype in type_converters:
+                return type_converters[datatype](value)
+            return value
+
+        def strip_deprecated_dir(key, value):
+            resolves_to = self.appschema[key].get('path_resolves_to')
+            if resolves_to:  # value is a path that will be resolved
+                dir = value.split(os.sep)[0]  # get first directory component
```
nsoranzo: `dir` is a Python standard library function, I'd use a different name for the variable.

lib/galaxy/config/__init__.py
```diff
@@ -180,57 +173,142 @@ def _parse_config_file_options(self, defaults, listify_defaults, config_kwargs):
...
+                raise_error(message)
+
+        def check_type_is_str(option, key):
+            if option.get('type') != 'str':
+                message = "Invalid schema: property '{}' should have type 'str'".format(key)
+                raise_error(message)
+
+        def check_is_dag():
+            def walk(key):
+                if key:
```
nsoranzo: Is that check needed?

lib/galaxy/config/__init__.py
```diff
@@ -180,57 +173,142 @@ def _parse_config_file_options(self, defaults, listify_defaults, config_kwargs):
...
+                    if key in visited:
+                        raise_error('Invalid schema: cycle detected')
+                    visited.add(key)
+                    next = self.appschema[key].get('path_resolves_to')
```
nsoranzo: `next` is also a Python standard library function, but I think you can just get rid of this line.
```suggestion
```

lib/galaxy/config/__init__.py
```diff
@@ -180,57 +173,142 @@ def _parse_config_file_options(self, defaults, listify_defaults, config_kwargs):
...
+        def raise_error(message):
+            log.error(message)
+            raise ConfigurationError(message)
+
+        for key, resolves_to in self._paths_to_resolve.items():
+            parent = self.appschema.get(resolves_to)
+            check_exists(parent, key)
+            check_type_is_str(parent, key)
+            check_type_is_str(self.appschema[key], key)
+            check_is_dag()  # must be called last: walks entire graph
```
nsoranzo: I think this line should be outside of the for loop.
```suggestion
        check_is_dag()  # must be called last: walks entire graph
```

lib/galaxy/config/__init__.py
```diff
@@ -180,57 +173,142 @@ def _parse_config_file_options(self, defaults, listify_defaults, config_kwargs):
...
+                raise_error(message)
+
+        def check_is_dag():
+            def walk(key):
+                if key:
+                    if key in visited:
+                        raise_error('Invalid schema: cycle detected')
+                    visited.add(key)
+                    next = self.appschema[key].get('path_resolves_to')
+                    walk(next)
```
```suggestion
                    if parent:
                        walk(parent)
```

lib/galaxy/config/__init__.py
```diff
@@ -180,57 +173,142 @@ def _parse_config_file_options(self, defaults, listify_defaults, config_kwargs):
...
+
+            for key in self._paths_to_resolve:
```
nsoranzo, line 241:
```suggestion
            visited = set()
            for key in self._paths_to_resolve:
```

lib/galaxy/config/__init__.py
```diff
@@ -180,57 +173,142 @@ def _parse_config_file_options(self, defaults, listify_defaults, config_kwargs):
...
+                visited = set()
```
```suggestion
                if key not in visited:
```

lib/galaxy/config/__init__.py
```diff
@@ -180,57 +173,142 @@ def _parse_config_file_options(self, defaults, listify_defaults, config_kwargs):
...
+                walk(key)
```
```suggestion
                    walk(key)
```

lib/galaxy/config/__init__.py
jdavcs: Absolutely! Will change.

lib/galaxy/config/__init__.py
jdavcs: I'll change it, but I can't get rid of it: it's used on the next line

lib/galaxy/config/__init__.py
jdavcs: Yes. If key is None, we've reached the base case: None will be returned and recursion terminates.

lib/galaxy/config/__init__.py
nsoranzo: See comment on the next line!

lib/galaxy/config/__init__.py
nsoranzo: Ah, OK, then I'd still remove it from here and do the check before recursing (presently line 238).

lib/galaxy/config/__init__.py
```suggestion
                    parent = self._paths_to_resolve[key]
```

nsoranzo (commented):
I've updated some of the comments/suggestions.

lib/galaxy/config/__init__.py
jdavcs: Yes. Thank you for catching this! (it would've been O(n^2))

lib/galaxy/config/__init__.py
jdavcs: Like this?
```python
 def walk(key):
     if key in visited:                                                            
         raise_error('Invalid schema: cycle detected')                             
     visited.add(key)
     next_key = self.appschema[key].get('path_resolves_to')                        
     if next_key:
         walk(next_key)  
```
I don't mind. To me, testing for the base case at the top of the function makes it more intuitive (that's how my reasoning would work), but I don't mind doing it this way.

nsoranzo: Yes, see the code suggestions below!

test/unit/config/test_path_resolves_to.py
```diff
@@ -0,0 +1,141 @@
...
+from galaxy.config import GalaxyAppConfiguration
+
+
+MOCK_DEPRECATED_DIRS = {
+    'my_config_dir': 'old-config',
+    'my_data_dir': 'old-database',
+}
+
+# Mock properties loaded from schema (all options represent paths)
+MOCK_SCHEMA = {
```
jmchilton, line 12: You could have just loaded the actual Galaxy schema and it would have been a better test IMO - this much mocking is increasing the test isolation but making the test more brittle, verbose and as a result harder to maintain. I'd encourage you to mock less and broaden what you think about as a unit - mocking out expensive interactions and not every interaction. I've read many articles, books that argue both ways and experimented with both kinds of tests in this code base and I've just developed an intuition that this is too much mocking. I think the most valuable tests in "test/unit" either look more like quasi-integration tests or are testing really complex "units".

Given that this is just an opinion and likely people on the team even would disagree with me - maybe I'd just encourage writing a bit more of integration test with your next feature and see if it feels like it would catch more errors/be easier to maintain - and do this mock heavy testing again if it doesn't feel right. This testing is certainly way better than no testing.

For this code base and functionality like this - I'd rather see just one integration test where a couple paths are correctly rewritten - and then break down into unit tests like this when there are more than a couple edge cases, complex error handling, etc..

jdavcs: I trust your intuition, as my experience with large code bases is limited. And I do think this should be covered with integration tests as well.

I try to restrain myself from doing clever things in test code. I'll be more conservative going forward.

I still feel that in this particular case (and similar cases) we need simple tests that isolate specific functionality. So, my motivation is not so much to keep IO out, but rather to be able to set a very specific context (which is not always present in the real app) - and test that that context is handled as expected.

Maintainability. These tests use generic property names (I used `my_config_dir` instead of `config_dir` on purpose), so they test not specific properties, but the interaction of properties - which is exactly what I want to test. Also, using this staged context prevents the test from breaking when a concrete property name is changed.

All that said, I do recognize and agree with all the downsides you mention. Going forward, I'll keep this in mind and will try to evaluate carefully the tradeoffs between test isolation and the complexity mocking adds w.r.t. maintaining tests (and ensuring they continue testing what they are expected to test over time). Also will do less verbose, limited to what's critical, etc. And I'll try to add integration tests where appropriate.

lib/galaxy/config/__init__.py
jdavcs: So, if I got it right, here's the proposed code:
```python
def check_is_dag():
    def walk(key):
        if key in visited:
            raise_error('Invalid schema: cycle detected')
        visited.add(key)
        parent = self._paths_to_resolve[key]
        if parent:
            walk(parent)

    visited = set()
    for key in self._paths_to_resolve:
        if key not in visited:
            walk(key)
```
This won't work. 
Consider this graph: A->B, C->B. Here we have {A, C} in `paths_to_resolve`. When we get to C, we'll then get its parent, which is B, and of course B is in the visited set, so it will look like it found a cycle. (this graph is 2 directories resolved w.r.t. the same parent directory). So we do need to create an empty set for each search path.
Also, this: `parent = self._paths_to_resolve[key]` won't get us B, because B does not need to be resolved (it doesn't have the `path_resolves_to` attribute, so it's not in that collection (there'll be always at least one sink node, otherwise there has to be a cycle).

I'll change `next` to `parent` and will switch the key checks as we discussed.

nsoranzo: You're right on both counts, thanks for the explanation!
I think the following should work though:

```python
def check_is_dag():
    def visit(key):
        visited.add(key)
        stack.add(key)
        parent = self.appschema[key].get('path_resolves_to')
        if parent:
            if parent not in visited:
                visit(parent)
            elif parent in stack:
                raise_error('Invalid schema: cycle detected')
        stack.remove(key)

    visited = set()
    stack = set()
    for key in self._paths_to_resolve:
        if key not in visited:
            visit(key)
```

This never visits a node twice and has the least recursion calls.

xref.: https://www.geeksforgeeks.org/detect-cycle-in-a-graph/

jdavcs: I must thank you for this discussion, because through this I think we have an optimal solution.

First, I do see a few issues in the approach above. I wouldn't use a set to represent a stack simply because it's an unordered collection (so it cannot be a LIFO data structure). Also, if we need to do a check like `if foo in bar`, bar should not be a stack (because conceptually, we can only peek or pop from the top of a stack). 

Next, the described cycle detection algorithm is applicable to broader problems. We have a very specific graph: all vertices have at most 1 outgoing edge (because a path cannot resolve to more than one parent), so we never need to backtrack, and so a classic DFS cycle detection algorithm (that looks for back edges) is not needed.

Most importantly, if we use a stack, we don't need recursion at all. The tradeoff is that an iterative algorithm will be usually less intuitive than a recursive algorithm (because we have to maintain our own stack). And I think readability (i.e., developer time) is more important here, given our tiny n (I estimate, unless we are resolving at least 10,000 directories, this won't matter; and in any case we'd profile first.

With that in mind, I was about to make a stack-based iterative solution... Until I realized that for this problem we don't even need a stack (because since a vertex can only have at most 1 outgoing edge, the size of the stack is 0 or 1). Which brings us to a trivial, dirt-simple solution! Here's the  pseudocode:

```
for v in G:
    visited = {}
    while v:
        visited.add(v)
        v = v.next
        if v and v in visited: error (cycle)
```
That's all!

Concrete implementation in next commit. Thank you, again - it was a fun derivation!

jdavcs:
Rebased.

lib/galaxy/config/__init__.py
jdavcs: Is this resolved?

jdavcs:
Rebased.

jdavcs:
One unit test failure unrelated (nondeterministic test).

dannon:
This is working for me pretty well in some additional local testing (some with very, very old configurations, in particular).  Thanks, @ic4f.

jdavcs:
Thank you, @dannon ! And thank you @jmchilton and @nsoranzo for the detailed review/suggestions, and @natefoo for the really cool idea :)

lib/galaxy/config/__init__.py
```diff
@@ -229,11 +292,31 @@ def _create_attributes_from_raw_config(self):
...
+
+            path = getattr(self, key)  # path prior to being resolved
+            parent = self.appschema[key].get('path_resolves_to')
+            if not parent:  # base case: nothing else needs resolving
+                return path
+            parent_path = resolve(parent)  # recursively resolve parent path
+            if path is not None:
+                path = os.path.join(parent_path, path)  # resolve path
+            else:
+                path = parent_path  # or use parent path
```
nsoranzo, line 309: @ic4f Why override a `None` with the `parent_path` ?

jdavcs: Because a path for an option can be not specified, while the `path_resolves_to` attribute *will* be.  In which case, we want to fall back to the parent path of that option. For example:

optionA: has a value 'myfile', resolves to: 'optionB'
optionB: does not have a value (no default or overridden with None - whatever), but resolve to: 'optionC'
optionC: has a value 'filesdir'

In this case, we want 'myfile' to be in 'filesdir'.

Here's the unit test for this usecase https://github.com/galaxyproject/galaxy/pull/8689/files#diff-6882d9765012ff6356c14d0e336c78c1R164

PR #8691, opened 2019-09-23
Interactive modification of invocation report.

jmchilton:
Press control-e while viewing an invocation report and allow the Markdown to be edited in realtime. Press escape to go back to the viewing mode and see the changed markdown rendered.

Results aren't persisted anywhere but copy and paste can be used to save template back into the workflow and this becomes like an interactive preview functionality.

### DIFF
--- a/client/galaxy/scripts/components/WorkflowInvocationReport.vue
+++ b/client/galaxy/scripts/components/WorkflowInvocationReport.vue
@@ -1,15 +1,22 @@
 <template>
-    <markdown :markdown-config="markdownConfig"> </markdown>
+    <markdown :markdown-config="markdownConfig" v-if="!edit"></markdown>
+    <markdown-editor
+        :initialMarkdown="markdownConfig.invocation_markdown"
+        :onupdate="onupdate"
+        v-else
+    ></markdown-editor>
 </template>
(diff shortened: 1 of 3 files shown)


client/galaxy/scripts/components/WorkflowInvocationReport.vue
```diff
@@ -19,18 +26,47 @@ export default {
...
+        this.loadMarkdownConfig();
+        window.addEventListener("keyup", this.keyListener);
+    },
+    destroyed: function() {
+        window.removeEventListener("keyup", this.keyListener);
     },
     methods: {
-        ajaxCall: function(url) {
+        keyListener: function(event) {
+            if (event.keyCode == 27) {
```
martenson: Can we bind this (more or less hidden feature) to something else please? I don't think this fits the standard expectations for hitting an escape key. `e` , `ctrl`+`e`?

jmchilton:
WIP until @martenson's comments are addressed

jmchilton:
Done, thanks for the suggestion @martenson

martenson (approved):

martenson:
Thank you John.

PR #8668, opened 2019-09-18
[WIP] [19.09] backport: Fix bug in build_dependency_manager()

martenson:
backport of https://github.com/galaxyproject/galaxy/pull/8660

PR #8695, opened 2019-09-23
Fix bug introduced in #8660

jdavcs:
Also ref #8668

### DIFF
--- a/lib/galaxy/tools/toolbox/base.py
+++ b/lib/galaxy/tools/toolbox/base.py
@@ -1161,6 +1161,8 @@ class BaseGalaxyToolBox(AbstractToolBox):
         app_config_dict = self.app.config.config_dict
         conf_file = app_config_dict.get("dependency_resolvers_config_file")
         default_tool_dependency_dir = self.app.config.get('tool_dependency_dir')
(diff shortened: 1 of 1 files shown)


guerler:
@ic4f I still see a forbidden (403) error when attempting to install toolshed repositories through the api using the client interface in the admin panel.

jdavcs:
@guerler from what I can tell, I don't think this is related to #8660. However, I am not sure: I'm not very familiar with how TS works. 
Ping @martenson : if we revert those changes ( https://github.com/galaxyproject/usegalaxy-playbook/commit/294f464d67bc30a898ce223aad37cd6aa77ff1d5 ) and that fixes the problem, I'll look again. Otherwise, I don't think I know TS well enough (not yet) to fix this quickly.

guerler:
@ic4f I think it is.

jdavcs:
@guerler try to set the 2 config options from the usegalaxy-playbook i referred to in my comment. If it works, that means 8660 did not resolve that issue correctly. If it doesn't work, that would mean than  something else is  happening (other than 8660). [edited]

nsoranzo:
This will be resolved when 19.09 is merged forward.

PR #8696, opened 2019-09-24
[19.09] Build dependency manager only for galaxy app

### LINKED ISSUE #8656: shed requires options to boot it shouldn't

nsoranzo:
Fix https://github.com/galaxyproject/galaxy/issues/8656

Broken in https://github.com/galaxyproject/galaxy/pull/8152

Add test for bug introduced in https://github.com/galaxyproject/galaxy/pull/8660 when trying to fix #8656 on the dev branch.

### DIFF
--- a/lib/galaxy/tools/toolbox/base.py
+++ b/lib/galaxy/tools/toolbox/base.py
@@ -1161,7 +1161,8 @@ class BaseGalaxyToolBox(AbstractToolBox):
         return looks_like_a_tool(path, enable_beta_formats=getattr(self.app.config, "enable_beta_tool_formats", False))
 
     def _init_dependency_manager(self):
-        if self.app.name != 'galaxy':
+        use_tool_dependency_resolution = getattr(self.app, "use_tool_dependency_resolution", True)
+        if not use_tool_dependency_resolution:
             self.dependency_manager = NullDependencyManager()
             return
         app_config_dict = self.app.config.config_dict
--- a/lib/galaxy/tools/toolbox/base.py
+++ b/lib/galaxy/tools/toolbox/base.py
@@ -15,7 +15,10 @@ from six import iteritems
 from six.moves.urllib.parse import urlparse
 
 from galaxy.exceptions import MessageException, ObjectNotFound
-from galaxy.tool_util.deps import build_dependency_manager
+from galaxy.tool_util.deps import (
+    build_dependency_manager,
+    NullDependencyManager
+)
 from galaxy.tool_util.loader_directory import looks_like_a_tool
 from galaxy.util import (
     ExecutionTimer,
@@ -1158,6 +1161,9 @@ class BaseGalaxyToolBox(AbstractToolBox):
         return looks_like_a_tool(path, enable_beta_formats=getattr(self.app.config, "enable_beta_tool_formats", False))
 
     def _init_dependency_manager(self):
+        if self.app.name != 'galaxy':
+            self.dependency_manager = NullDependencyManager()
+            return
         app_config_dict = self.app.config.config_dict
         conf_file = app_config_dict.get("dependency_resolvers_config_file")
         default_tool_dependency_dir = os.path.join(self.app.config.data_dir, "dependencies")
--- a/lib/galaxy/webapps/tool_shed/app.py
+++ b/lib/galaxy/webapps/tool_shed/app.py
@@ -64,6 +64,7 @@ class UniverseApplication(object):
         from galaxy.managers.citations import CitationsManager
         self.citations_manager = CitationsManager(self)
         # The Tool Shed makes no use of a Galaxy toolbox, but this attribute is still required.
+        self.use_tool_dependency_resolution = False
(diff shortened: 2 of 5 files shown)


lib/galaxy/tools/toolbox/base.py
```diff
@@ -1158,6 +1161,9 @@ def _looks_like_a_tool(self, path):
         return looks_like_a_tool(path, enable_beta_formats=getattr(self.app.config, "enable_beta_tool_formats", False))
 
     def _init_dependency_manager(self):
+        if self.app.name != 'galaxy':
```
jmchilton: This seems very backwards to me - this class clearly states it handles dependency resolution in the description - it is called ``BaseGalaxyToolBox`` after all. This class shouldn't need to be called with an app and with an app named a particular thing in order to function this way.

Please cherry-pick in https://github.com/galaxyproject/galaxy/commit/902aabcba7925c424bf381ca1b49f4484cfc16e5 or something like that?

nsoranzo: @jmchilton I've included a slightly modified version of your commit, thanks!

lib/galaxy/webapps/tool_shed/app.py
```diff
@@ -64,6 +64,7 @@ def __init__(self, **kwd):
         from galaxy.managers.citations import CitationsManager
         self.citations_manager = CitationsManager(self)
         # The Tool Shed makes no use of a Galaxy toolbox, but this attribute is still required.
+        self.config.use_tool_dependencies = False
```
jdavcs: This is OK for this edit. However, in the long run, I think we should try to keep any modifications to the app.config object to the modules that load it at startup (i.e., files `config/__init__.py` for galaxy and `config.py` for TS and reports).

test/base/driver_util.py
```diff
@@ -292,11 +288,12 @@ def setup_galaxy_config(
...
         config["enable_tool_shed_check"] = enable_tool_shed_check
         config["hours_between_check"] = 0.001
+    tool_dependency_dir = os.environ.get('GALAXY_TOOL_DEPENDENCY_DIR')
     if tool_dependency_dir:
         config["tool_dependency_dir"] = tool_dependency_dir
-        # Used by shed's twill dependency stuff - todo read from
-        # Galaxy's config API.
-        os.environ["GALAXY_TEST_TOOL_DEPENDENCY_DIR"] = tool_dependency_dir
+    # Used by shed's twill dependency stuff
+    # TODO: read from Galaxy's config API.
```
jdavcs, line 295: completely agree!

lib/galaxy/tools/toolbox/base.py
```diff
@@ -1158,6 +1161,9 @@ def _looks_like_a_tool(self, path):
         return looks_like_a_tool(path, enable_beta_formats=getattr(self.app.config, "enable_beta_tool_formats", False))
 
     def _init_dependency_manager(self):
+        if not getattr(self.app.config, "use_tool_dependencies", True):
```
jdavcs: yes,  I think this is a better approach than checking the app name.

jdavcs (approved):
I don't know this part of the codebase in sufficient depth, so I've only added a couple of general comments. Most importantly, I agree with the change from checking the app name to checking an (explicitly named) config property. Better design + intuitive read. Thanks!

lib/galaxy/webapps/tool_shed/app.py
nsoranzo: In the original suggestion from @jmchilton , this was `self.app.use_tool_dependencies`, I can just back to that if you prefer, no strong opinion from me.

jdavcs: No preference. Now that I think of it, it could belong in either one, and config seems reasonable, I think. (I think if it were to go into the app object, that would imply that it is not a configurable property in the first place - and that's a discussion that, I think, belongs outside of this PR)

jmchilton (approved):
I slightly preferred the check on ``app`` for different reasons than @ic4f's critique but that is also a valid point. I don't feel strongly about it though so approving as is - feel free to merge. Thanks for the fix @nsoranzo - sorry about breaking this.

lib/galaxy/webapps/tool_shed/app.py
nsoranzo: I've restored John's version, I think it is suppose to mean that the Tool Shed app does not support tool dependency resolution, so it's not configurable. On Galaxy instead, it is indeed supported and configurable.

jdavcs: OK, that makes sense.

PR #8697, opened 2019-09-24
LDAP search member of filter 

BioinformaticsPlatformWIV-ISP:
Hello,

Latest version of the pr #8628 (previously #7543) generated from a clone of the origin galaxy repo dev branch. This should avoid merge conflics and is up to date with the dev branch.

### DIFF
--- a/lib/galaxy/auth/providers/ldap_ad.py
+++ b/lib/galaxy/auth/providers/ldap_ad.py
@@ -220,7 +220,7 @@ class LDAP(AuthProvider):
             attributes['roles'] = params[self.role_search_option]
         return (True,
                 _get_subs(options, 'auto-register-email', params),
-                transform_publicname(_get_subs(options, 'auto-register-username', params), check_dup=False),
+                transform_publicname(_get_subs(options, 'auto-register-username', params)),
                 attributes)
 
     def _authenticate(self, params, options):
--- a/lib/galaxy/auth/providers/ldap_ad.py
+++ b/lib/galaxy/auth/providers/ldap_ad.py
@@ -5,6 +5,7 @@ Created on 15/07/2014
 """
 
 import logging
+import re
 
 from galaxy.exceptions import ConfigurationError
 from galaxy.util import string_as_bool
@@ -23,6 +24,10 @@ def _get_subs(d, k, params):
     if k not in d or not d[k]:
         raise ConfigurationError("Missing '%s' parameter in LDAP options" % k)
     return str(d[k]).format(**params)
+
+
+def _clean_register_username(displayname):
+    return re.sub('[^a-z0-9]+', '_', displayname.lower())
 
 
 def _parse_ldap_options(options_unparsed):
@@ -151,6 +156,8 @@ class LDAP(AuthProvider):
                 # setup search
                 attributes = [_.strip().format(**params)
                               for _ in options['search-fields'].split(',')]
+                if 'search-memberof-filter' in options:
+                    attributes.append('memberOf')
                 suser = l.search_ext_s(_get_subs(options, 'search-base', params),
                     ldap.SCOPE_SUBTREE,
                     _get_subs(options, 'search-filter', params), attributes,
@@ -168,6 +175,8 @@ class LDAP(AuthProvider):
                         if self.role_search_attribute and attr == self.role_search_attribute[1:-1]:  # strip brackets
                             # keep role names as list
                             params[self.role_search_option] = attrs[attr]
+                        elif attr == 'memberOf':
+                            params[attr] = attrs[attr]
                         elif attr in attrs:
                             params[attr] = str(attrs[attr][0])
                         else:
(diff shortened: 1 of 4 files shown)


nsoranzo:
@galaxybot test this

nsoranzo:
@BioinformaticsPlatformWIV-ISP Can you please check the commit I've pushed to this PR? The rest looks good to me!

nsoranzo:
@galaxybot test this

lib/galaxy/security/validate_user_input.py
```diff
@@ -65,24 +67,27 @@ def validate_publicname(trans, publicname, user=None):
     return ''
 
 
-def transform_publicname(trans, publicname, user=None):
-    # User names must be at least four characters in length and contain only lower-case
-    # letters, numbers, and the '-' character.
+def transform_publicname(trans, publicname, user=None, check_dup=True):
```
BioinformaticsPlatformWIV-ISP: `transform_publicname` has two mandatory argument `trans` and `publicname`.  Only the first one is receive when called from `authenticate`. Lead to crash (exception raise) in this case. In addition, `trans` is only used when duplication check is enabled.

lib/galaxy/security/validate_user_input.py
```diff
@@ -65,24 +67,27 @@ def validate_publicname(trans, publicname, user=None):
...
+    FILL_CHAR is used to extend or replace characters.
+    """
     # TODO: Enhance to allow generation of semi-random publicnnames e.g., when valid but taken
     if user and user.username == publicname:
         return publicname
-    elif publicname not in ['None', None, '']:
+    if publicname not in ['None', None, '']:
         publicname = publicname.lower()
         publicname = re.sub(VALID_PUBLICNAME_SUB, FILL_CHAR, publicname)
         publicname = publicname.ljust(PUBLICNAME_MIN_LEN + 1, FILL_CHAR)[:PUBLICNAME_MAX_LEN]
```
BioinformaticsPlatformWIV-ISP, line 80: `FILL_CHAR` is set to '-' (minus) whereas our old clean function was using '_ ' (underscore) instead. This is valid but a little annoying for us. We will have two kind of pattern for galaxy public username. One with '_' and the other one with '-' depending of when the username has been generated (before or after galaxy 20.1) .  I think that as admin we can deal with it but for ours user it will be awkward.

lib/galaxy/security/validate_user_input.py
```diff
@@ -65,24 +67,27 @@ def validate_publicname(trans, publicname, user=None):
...
-    # User names must be at least four characters in length and contain only lower-case
-    # letters, numbers, and the '-' character.
+def transform_publicname(trans, publicname, user=None, check_dup=True):
+    """
+    Transform publicname to respect the minimum and maximum string length, and
+    the allowed characters.
+    FILL_CHAR is used to extend or replace characters.
+    """
     # TODO: Enhance to allow generation of semi-random publicnnames e.g., when valid but taken
     if user and user.username == publicname:
```
BioinformaticsPlatformWIV-ISP: Do you have any use case in mind where `transform_publicname` is called with an existing user and a matching publicname ?

nsoranzo: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/tool_shed/controllers/user.py#L157

lib/galaxy/auth/providers/ldap_ad.py
```diff
@@ -204,12 +209,18 @@ def authenticate(self, email, username, password, options):
...
+            if not any(search_filter in ad_node_name for ad_node_name in params['memberOf']):
+                return failure_mode, '', ''
+
         attributes = {}
         if self.auto_create_roles_or_groups:
             attributes['roles'] = params[self.role_search_option]
         return (True,
                 _get_subs(options, 'auto-register-email', params),
-                _get_subs(options, 'auto-register-username', params),
+                transform_publicname(_get_subs(options, 'auto-register-username', params), check_dup=False),
```
BioinformaticsPlatformWIV-ISP: See the other comment. Crash when called this way, without the `trans` parameter.  If there is no access to `trans ` in this context, maybe separate the public name transformation and the duplication check in two functions and only call the transformation specific one here ?

nsoranzo: Yes, sorry for the trouble, that's a good plan, I'll work on a patch.

lib/galaxy/security/validate_user_input.py
nsoranzo: I guess you can use a database query to update the existing Galaxy usernames to use `-`?

nsoranzo:
@galaxybot test this

lib/galaxy/auth/providers/ldap_ad.py
nsoranzo: It should be fixed now.

BioinformaticsPlatformWIV-ISP:
Your latest changes have fix the 'transform_publicname' crash, authentication process is now back to normal. Thank you for your help.

nsoranzo (approved):

PR #8543, opened 2019-08-29
Implement workflow invocation reports (🎉 syntax)

jmchilton:
Builds on https://github.com/galaxyproject/galaxy/pull/8511 and #8030 which had the workflow invocation reports feature but with other syntaxes.

The community took a vote (https://github.com/galaxyproject/galaxy/pull/8511#issuecomment-525886525) and the overwhelming favorite was this programming style declaration of a vaguely C-ish language embedded in quote literal fences/blocks.

    ```galaxy
    job_metrics(step=image_cat)
    ```

or

    ```galaxy
    invocation_inputs()
    ```

The advantage over the previous syntaxes is that the literal block itself contains information on what to render so alternative Markdown rendering will at least make it clear what should be there on some level. The "language" identifier at the start of the block feels less arbitrary than the second syntax (#8511) and while this looks less superficially like RMarkdown at the entry-point - the breakdown of what is considered the language declaration and what is being rendered is perhaps closer in spirit. 

Other than the Markdown syntax this PR is the same as the predecessor in terms of APIs, tests, features.

Sample workflow is:

```yaml
class: GalaxyWorkflow
name: My Cool Workflow
inputs:
  input_1: data
  image_input: data
  input_list: collection
outputs:
  output_1:
    outputSource: first_cat/out_file1
  output_image:
    outputSource: image_cat/out_file1
steps:
  first_cat:
    tool_id: cat
    in:
      input1: input_1
  image_cat:
    tool_id: cat
    in:
      input1: image_input
  qc_step:
    tool_id: qc_stdout
    state:
      quality: 9
    in:
      input: input_1
report:
  markdown: |
    ## About This Report

    This report is generated from markdown content in the workflow YAML/JSON.
    Workflow invocation reports provide a customizable way for workflow authors
    to summarize the results of a workflow execution. There is a default markdown
    template if a workflow does not define one (right now it just shows the inputs,
    outputs, and workflow - but this will evolve).

    This report is written in "Galaxy Workflow Flavored Markdown". This workflow
    variant of "Galaxy Flavored Markdown" contains workflow-relative references.
    Dataset, collections, and jobs are referenced by step labels, input labels, and output
    labels - rather than by object IDs. The Galaxy invocation report generator plugin
    translates this to "Galaxy Flavored Markdown" at the time the report is viewed and
    that format references actual object IDs (by database ID internally, and encoded ID
    when exported to the client via the API).

    An upshot of translating the workflow markdown to this second neutral format that
[...]

PR #8698, opened 2019-09-24
[19.09] Bug fix in loading stored workflows from workflow ids.

jmchilton:
Broken when introduced in #8543 I believe.

### DIFF
--- a/lib/galaxy/managers/workflows.py
+++ b/lib/galaxy/managers/workflows.py
@@ -67,7 +67,7 @@ class WorkflowsManager(object):
             # see if they have passed in the UUID for a workflow that is attached to a stored workflow
             workflow_uuid = uuid.UUID(workflow_id)
             workflow_query = trans.sa_session.query(trans.app.model.StoredWorkflow).filter(and_(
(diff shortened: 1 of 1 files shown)


lib/galaxy/managers/workflows.py
```diff
@@ -77,7 +77,7 @@ def get_stored_workflow(self, trans, workflow_id, by_stored_id=True):
         else:
             workflow_id = decode_id(self.app, workflow_id)
             workflow_query = trans.sa_session.query(trans.app.model.StoredWorkflow).filter(and_(
-                trans.app.model.StoredWorkflow.latest_workflow_id == trans.app.model.Workflow.id,
+                trans.app.model.StoredWorkflow.id == trans.app.model.Workflow.stored_workflow_id,
```
nsoranzo, line 80: Shouldn't be the same also on line 70?

jmchilton: Ahhh - I was wondering what in the world I was thinking when I added that. Makes sense where I got it from now, thanks @nsoranzo !

nsoranzo (approved):

PR #8699, opened 2019-09-24
User-facing Workflow Invocation Grid

jmchilton:
- Refactor generic grid out of admin-centric invocation grid introduced in #8664 
- Add contextual menus for target workflow and history based on WorkflowDropdown added in #8551.
- Add workflow and history vuex stores for caching storedworkflow and history data fetched from the server (based on vuex-cache pattern introduced in #8044
- Update all workflow APIs to allow fetching stored workflows from a stored workflow ID or a workflow ID using the new query parameter (``instance=false`` - override default of ``false`` to ``true`` to fetch by a ``model.Workflow`` ID instead of a ``model.StoredWorkflow`` ID).

I had a request from another developer that the basic version of this grid (5d916e3) without the contextual menus should be backported to 19.09 - I think that is reasonable but I also recognize it was already a big ask to get #8664 in.

Screen Shot 2019-09-24 at 10 19 59 AM

Screen Shot 2019-09-24 at 10 20 44 AM

Screen Shot 2019-09-24 at 10 20 52 AM

### DIFF
--- a/client/galaxy/scripts/components/History/HistoryDropdown.vue
+++ b/client/galaxy/scripts/components/History/HistoryDropdown.vue
@@ -11,9 +11,7 @@
         </b-link>
         <div class="dropdown-menu" aria-labelledby="history-dropdown">
             <a class="dropdown-item" :href="urlView">View</a>
-            <!--
             <a class="dropdown-item" href="#" @click="onSwitch">Switch To</a>
-            -->
             <a class="dropdown-item" :href="urlShowStructure">Structure</a>
             <a class="dropdown-item" :href="urlSharing">Sharing</a>
         </div>
@@ -21,7 +19,7 @@
 </template>
 <script>
 import { getAppRoot } from "onload/loadConfig";
-import jQuery from "jquery";
+import { getGalaxyInstance } from "app";
 
 export default {
     props: ["history"],
@@ -37,9 +35,8 @@ export default {
     },
     methods: {
         onSwitch: function() {
-            // this switches the history, but doesn't trigger a reload yet...
-            // extract as service from history-model?
-            jQuery.getJSON(`${getAppRoot()}history/set_as_current?id=${this.history.id}`);
+            const Galaxy = getGalaxyInstance();
+            Galaxy.currHistoryPanel.switchToHistory(this.history.id);
         }
     }
 };
--- a/client/galaxy/scripts/components/HistoryDropdown.vue
+++ b/client/galaxy/scripts/components/HistoryDropdown.vue
@@ -1,0 +1,46 @@
+<template>
+    <div>
+        <b-link
+            id="history-dropdown"
+            class="history-dropdown font-weight-bold"
+            data-toggle="dropdown"
+            aria-haspopup="true"
+            aria-expanded="false"
+        >
+            {{ history.name }}
+        </b-link>
+        <div class="dropdown-menu" aria-labelledby="history-dropdown">
+            <a class="dropdown-item" :href="urlView">View</a>
+            <!--
+            <a class="dropdown-item" href="#" @click="onSwitch">Switch To</a>
+            -->
+            <a class="dropdown-item" :href="urlShowStructure">Structure</a>
+            <a class="dropdown-item" :href="urlSharing">Sharing</a>
+        </div>
+    </div>
+</template>
+<script>
+import { getAppRoot } from "onload/loadConfig";
+import jQuery from "jquery";
+
+export default {
+    props: ["history"],
+    data() {
+        return {
+            urlView: `${getAppRoot()}histories/view?id=${this.history.id}`,
+            urlShowStructure: `${getAppRoot()}histories/show_structure?id=${this.history.id}`,
+            urlSharing: `${getAppRoot()}histories/sharing?id=${this.history.id}`
+        };
+    },
+    created() {
+        this.root = getAppRoot();
+    },
+    methods: {
+        onSwitch: function() {
+            // this switches the history, but doesn't trigger a reload yet...
+            // extract as service from history-model?
+            jQuery.getJSON(`${getAppRoot()}history/set_as_current?id=${this.history.id}`);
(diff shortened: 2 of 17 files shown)


client/galaxy/scripts/components/admin/Invocations.vue
```diff
@@ -3,18 +3,16 @@
         <h2 class="mb-3">
```
jmchilton: File should be moved out of ``components/admin``.

client/galaxy/scripts/components/HistoryDropdown.vue
```diff
@@ -0,0 +1,46 @@
...
+        };
+    },
+    created() {
+        this.root = getAppRoot();
+    },
+    methods: {
+        onSwitch: function() {
+            // this switches the history, but doesn't trigger a reload yet...
+            // extract as service from history-model?
+            jQuery.getJSON(`${getAppRoot()}history/set_as_current?id=${this.history.id}`);
```
guerler: We should request the data without using jquery.

client/galaxy/scripts/components/User/RecentInvocations.vue
```diff
@@ -0,0 +1,41 @@
...
+
+export default {
+    components: {
+        Invocations
+    },
+    data() {
+        return {
+            invocationItems: [],
+            loading: true,
+            busy: true,
```
guerler: Wondering if and how we use the `busy` attribute here and if it is equivalent to `loading`.

jmchilton:
Rabased and Committed: Tweaks to invocations grid based on comments.

- Eliminate unneeded busy variable.
- Refactor setting current history to share with history-model, enable in grid - eliminate the use of jquery.
- Move Invocations.vue (the base grid) out of components/admin and into components/Workflow/

client/galaxy/scripts/components/User/RecentInvocations.vue
jmchilton: Yeah - eliminated that - good point!

client/galaxy/scripts/components/Workflow/Invocations.vue
```diff
@@ -0,0 +1,138 @@
...
+                        <workflow-invocation-state :invocationId="row.item.id" :provideContext="false" />
+                    </b-card>
+                </template>
+                <template v-slot:cell(workflow_id)="data">
+                    <div v-if="!ownerGrid || !getWorkflowByInstanceId(data.item.workflow_id)">
+                        {{ data.item.workflow_id }}
+                    </div>
+                    <div v-else>
+                        <workflow-dropdown :workflow="getWorkflowByInstanceId(data.item.workflow_id)" />
+                        <!--
```
mvdbeek: We can drop this, right ?

client/galaxy/scripts/components/History/HistoryDropdown.vue
```diff
@@ -0,0 +1,44 @@
...
+            urlSharing: `${getAppRoot()}histories/sharing?id=${this.history.id}`
+        };
+    },
+    created() {
+        this.root = getAppRoot();
+    },
+    methods: {
+        onSwitch: function() {
+            // this switches the history, but doesn't trigger a reload yet...
+            // extract as service from history-model?
```
mvdbeek: You've done that, right ?

client/galaxy/scripts/components/History/HistoryDropdown.vue
```diff
@@ -0,0 +1,44 @@
...
+            urlShowStructure: `${getAppRoot()}histories/show_structure?id=${this.history.id}`,
+            urlSharing: `${getAppRoot()}histories/sharing?id=${this.history.id}`
+        };
+    },
+    created() {
+        this.root = getAppRoot();
+    },
+    methods: {
+        onSwitch: function() {
+            // this switches the history, but doesn't trigger a reload yet...
```
mvdbeek: I'm not sure if this is correct and at which level to place this, but `Galaxy.currHistoryPanel.switchToHistory(this.history.id);` seems to work in my testing.

mvdbeek (commented):
This is so cool! But click on the row (but not on workfow or history) to expand the invocations feels a bit awkward. Maybe a `+` sign on the left would help ?

jmchilton:
The row expanding details was copied over from the jobs list interface in the admin UI - but I agree it isn't super clear. I had assumed the repository grid and details worked the same but it doesn't. I'll take another shot at this.

client/galaxy/scripts/components/History/HistoryDropdown.vue
jmchilton: Yeah - I think I'm going to revert the service extraction and de-duplication and just use this line. I'm worried about conflicts with other history panel work on going - this seems to be a very clean solution - thanks!

client/galaxy/scripts/components/Workflow/Invocations.vue
jmchilton: Yup, I've cleaned this up in a rebase. Thanks!

client/galaxy/scripts/components/History/HistoryDropdown.vue
jmchilton: I've cleaned this up in a rebase. Thanks!

mvdbeek (approved):
There's a glitch with the drop down appearing higher up than you'd expect it, but having the overview is very nice.
Screenshot 2019-11-19 at 15 51 23

martenson:
switching to WIP since it has conflicts and outstanding comments

jmchilton:
Seems the drop down menu was not compatible with the table being marked 'responsive'. Fixed and rebased to resolve conflicts.

martenson:
I pushed a tiny icon change. Rest LGTM. Thanks!

mvdbeek:
Adding a screenshot for the release notes:
Screenshot 2020-02-04 at 11 33 00

PR #8616, opened 2019-09-11
[19.09] Polish YAML based job configuration for 19.09 release.

jmchilton:
Translate much more of the XML advanced configuration docs to job_conf.sample_advanced.yml and fill in small functionality holes missing from along the way.

See individual commits for details - builds on #8614 #8609.

See https://github.com/galaxyproject/galaxy/pull/7854 for the original PR adding YAML based job configuration to 19.09.

PR #8702, opened 2019-09-24
[19.09] Specify a default tool error reports plugin config rather than reading it from the sample

natefoo:


lib/galaxy/util/plugin_config.py
```diff
@@ -8,6 +9,9 @@
 from galaxy.util.submodules import import_submodules
 
 
+PluginConfigSource = collections.namedtuple('PluginConfigSource', ['type', 'source'])
```
jmchilton, line 12: 🎩

jmchilton (approved):

lib/galaxy/util/plugin_config.py
natefoo: I strongly dislike accessing non-local tuples by index.

jmchilton: It was a good tophat 🤣

nsoranzo: Shouldn't also `load_plugins()` been modified at line 40 to use the keys?

natefoo: @nsoranzo There are other callers still passing a tuple apparently.

nsoranzo: Are there? I couldn't find any in `lib/galaxy/job_metrics/__init__.py` , `lib/galaxy/tool_util/deps/__init__.py` , `lib/galaxy/tool_util/deps/containers.py` or `lib/galaxy/tools/error_reports/__init__.py` .

natefoo: [Here's one](https://github.com/galaxyproject/galaxy/blob/5c1380fff60a73261d37ce3a9cb53e5ede56416c/lib/galaxy/tool_util/deps/__init__.py#L126).

jmchilton: This is going to conflict https://github.com/galaxyproject/galaxy/pull/8616/files isn't it, I shouldn't have taken so long to respond to the review of it.

jmchilton: https://github.com/galaxyproject/galaxy/pull/8616/commits/28b423b78b8050f9f26d4591a22365736f4ec61f

PR #8709, opened 2019-09-25
Merge release_19.09 into dev

nsoranzo:
Change `log_events` and `log_actions` default value to `false` in `lib/galaxy/webapps/galaxy/config_schema.yml` to reflect what was in `lib/galaxy/config/__init__.py` .

### DIFF
--- a/lib/galaxy/webapps/galaxy/config_schema.yml
+++ b/lib/galaxy/webapps/galaxy/config_schema.yml
@@ -259,7 +259,7 @@ mapping:
           to this path, such as the default conda prefix, default Galaxy packages path, legacy
           tool shed dependencies path, and the dependency cache directory.
 
-          Set the string to None to explicitly disable tool dependency handling.
+          Set the string to null to explicitly disable tool dependency handling.
           If this option is set to none or an invalid path, installing tools with dependencies
           from the Tool Shed or in Conda will fail.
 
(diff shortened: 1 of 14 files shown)


lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -267,7 +267,7 @@ mapping:
           to this path, such as the default conda prefix, default Galaxy packages path, legacy
           tool shed dependencies path, and the dependency cache directory.
 
-          Set the string to None to explicitly disable tool dependency handling.
+          Set the string to null to explicitly disable tool dependency handling.
```
jdavcs, line 270: So, right now this will evaluate to the string "null", which will be then resolved w.r.t `data_dir`, because the code only checks for the string "none" (which is a hack anyway). As we've discussed earlier, a user should be able to overwrite a schema default with a proper null. This has not been implemented yet, so I'll go ahead and implement it. I'm not suggesting that you undo this for now because (a) "null" is the more appropriate wording, and (b) that functionality needs to be added anyway.

jdavcs (approved):

lib/galaxy/webapps/galaxy/config_schema.yml
nsoranzo: I think you are mistaken, specifying `tool_dependency_dir: null` in `config/galaxy.yml` works as expected, setting `self.tool_dependency_dir` to `None` (type `NoneType`) in `lib/galaxy/config/__init__.py` .

jdavcs: You are right! I didn't know yaml understood null.

PR #8712, opened 2019-09-26
GRT: Ensure correct encoding, no partial lines

hexylena:
Younger, less wise @erasche made some strange and questionable decisions. Among them, a bazillion write calls because oh no building a string buffer. the memory usage! the horror. muh performance. (/s) 

So fixing that so we don't write partial lines if we bail out, and additionally writing utf8 because I found a non-ascii character in a tool output name which was fun to track down. But such things should be permitted, so, let's support it?

it could be backported to 19.09, but, no one is really running this other than us anyway sooooo...

### DIFF
--- a/scripts/grt/export.py
+++ b/scripts/grt/export.py
@@ -21,6 +21,7 @@ import galaxy.config
 from galaxy.objectstore import build_object_store_from_config
 from galaxy.util import hash_util
 from galaxy.util.script import app_properties_from_args, populate_config_args
+from galaxy.util import unicodify
 
 sample_config = os.path.abspath(os.path.join(os.path.dirname(__file__), 'grt.yml.sample'))
 default_config = os.path.abspath(os.path.join(os.path.dirname(__file__), 'grt.yml'))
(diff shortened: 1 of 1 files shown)


scripts/grt/export.py
```diff
@@ -155,16 +156,15 @@ def annotate(label, human_label=None):
...
-                handle_job.write('\n')
+                line = [
+                    str(job[0]), # id
+                    job[2], # tool_id
+                    job[3], # tool_version
+                    job[4], # state
+                    str(job[5]) # create_time
+                ]
+                cline = unicodify('\t'.join(line) + '\n')
+                handle_job.write(cline.encode('utf-8'))
```
nsoranzo: The `.encode('utf-8')` is not needed if you open `handle_job` with:
```python
import io
...
    handle_job = io.open(REPORT_BASE + '.jobs.tsv', 'w', encoding='utf-8')
```

hexylena: Huh, neat. Can do!

nsoranzo (approved):

scripts/grt/export.py
```diff
@@ -155,16 +157,15 @@ def annotate(label, human_label=None):
...
-                handle_job.write(job[2])  # tool_id
-                handle_job.write('\t')
-                handle_job.write(job[3])  # tool_version
-                handle_job.write('\t')
-                handle_job.write(job[4])  # state
-                handle_job.write('\t')
-                handle_job.write(str(job[5]))  # create_time
-                handle_job.write('\n')
+                line = [
+                    str(job[0]), # id
```
nsoranzo: flake8 requires 2 spaces before inline comments.

scripts/grt/export.py
```diff
@@ -21,6 +22,7 @@
 from galaxy.objectstore import build_object_store_from_config
 from galaxy.util import hash_util
```
nsoranzo:
```suggestion
from galaxy.util import (
    hash_util,
    unicodify
)
```

scripts/grt/export.py
```diff
@@ -21,6 +22,7 @@
...
 from galaxy.util.script import app_properties_from_args, populate_config_args
+from galaxy.util import unicodify
```
nsoranzo:
```suggestion
```

hexylena:
Thanks @nsoranzo!

PR #8714, opened 2019-09-26
[WIP] Tool recommendation in Galaxy using deep learning

anuprulez:
This PR adds the tool recommendation feature:
- Show the following tools after each tool execution and for tool blocks on workflow editor. It can be done in Galaxy by enabling two config variables in `config/galaxy.yml`:
    - `enable_tool_recommendations = true`
    - `tool_recommendation_model_path = `https://github.com/anuprulez/download_store/raw/tool_recommendation_model/tool_recommendation_model/tool_recommendation_model.hdf5`
- Model to predict tools is learned using deep learning and is downloaded from a remote location (`https://github.com/anuprulez/download_store/raw/tool_recommendation_model/tool_recommendation_model/tool_recommendation_model.hdf5`) to the path `galaxy/database`. The model is only read from this path and is not part of the Galaxy codebase.
- The input files needed to create this model can be collected using the following scripts from any Galaxy instance (thanks to @erasche):
    - Workflows (https://github.com/usegalaxy-eu/gxadmin/blob/dev/gxadmin#L1380)
    - Tool usage (https://github.com/usegalaxy-eu/gxadmin/blob/dev/gxadmin#L1345)
- A tool (`Create a model to recommend tools`) available on the European Galaxy server can be used to create this model. The two `tabular` files above are the inputs to this tool. There are multiple parameters for this tool which are explained in the tool's `help` section. This model should be created regularly (say every month) to keep it updated with new workflows and tools.
- A Galaxy API (in `WorkflowsAPIController`) reads the model and compute predictions.
- Each predicted tool has a score that is learned by the neural network. The scores of all predicted tools are normalized by dividing them with the maximum predicted score.
- The prediction score is boosted by getting a dot product of the predicted scores and the weights of tools (learned using their usage overtime on Galaxy).
- The predictions are filtered using datatype constraints and tool connections in workflows. The predictions with incompatible datatypes and non-existent tool connections are removed.
- In addition to the predicted tools, there is a provision to add and remove tools by the Galaxy admins by setting `admin_tool_recommendations_path` config variable. This information of tools can be encoded in `config/tool_recommendations_overwrite.yml.sample` file. Format for specifying these sections in `config/tool_recommendations_overwrite.yml.sample`. The tools specified as `deprecated` do not appear in the list of recommendations for any tool and a warning message is shown when a request for recommendations is made for these tools in the workflow editor and when they are executed. Adding additional tools in the recommendations can be useful to promote newly added tools.

To remove (deprecated) tools:
For example - the tools `tool _id_deprecated_1` is deprecated
```
<<tool _id_deprecated_1>>:
   - is_deprecate: True
   - text_message: <<message to show once they are recommended by the deep learning model>>
<<tool_id_deprecated_2>>:
   - is_deprecate: True
   - text_message: <<message to show once they are recommended by the deep learning model>>
<<tool_id_deprecated_3>>:
   - is_deprecate: True
   - text_message: <<message to show once they are recommended by the deep learning model>>
```

To add additional predictions:
For example: the tools `tool_id1_1` and `tool_id1_2` are recommended for `tool_id1`.
```
 <<tool_id1>>:
   - tool_id: <<tool_id1_1>>
     name: <<display text: name and description of the tool>>
     i_extensions:
       - <<input file extension>>
       - <<input file extension>>
   - tool_id: <<tool_id1_2>>
     name: <<display text: name and description of the tool>>
     i_extensions:
       - <<input file extension of this tool>>
       - <<input file extension of this tool>>
 <<tool_id2>>:
   - tool_id: <<tool_id2_1>>
     name: <<display text: name and description of the tool>>
     i_extensions:
       - <<input file extension>>
       - <<input file extension>>
   - tool_id: <<tool_id2_2>>
     name: <<display text: name and description of the tool>>
     i_extensions:
       - <<input file extension of this tool>>
       - <<input file extension of this tool>>
```
- It is an optional feature and can be enabled/disabled by setting `enable_tool_recommendations` and `tool_recommendation_model_path` config variables.
- There is an API test in `test/api/test_workflows.py` to check the validity of the model.
- It is functional on the European Galaxy server.

A few screenshots:

tool

workflow

Your suggestions are welcome!


lib/galaxy/config/sample/galaxy.yml.sample
```diff
@@ -1796,6 +1796,22 @@ galaxy:
   # persistent_communication_rooms is a comma-separated list of rooms
   # that should be always available.
   #persistent_communication_rooms: null
+
```
jdavcs: I believe these new config properties should be added to `webapps/galaxy/config_schema.yml`, then  run `make config rebuild` to rebuild all sample yaml and rst files.

lib/galaxy/config/__init__.py
```diff
@@ -363,6 +363,10 @@ def _process_config(self, kwargs):
                 log.error("CONFIGURATION ERROR: Can't open supplied blacklist file from path: " + str(self.blacklist_file))
 
         self.persistent_communication_rooms = listify(kwargs.get("persistent_communication_rooms", []), do_strip=True)
+        self.enable_tool_recommendations = string_as_bool(kwargs.get('enable_tool_recommendations', 'False'))
```
jdavcs: Once these config options are added to `config_schema.yml` (see other comment), corresponding properties will be generated automatically (and converted to correct datatype). See this: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L191 - these 5 lines summarize what the new config does (it's a very recent modification).

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -603,6 +609,168 @@ def build_module(self, trans, payload={}):
...
+            try:
+                d_key = "weight_" + str(weight_ctr)
+                weights = self.trained_model.get(d_key).value
+                model_weights.append(weights)
+                weight_ctr += 1
+            except Exception as exception:
+                break
+        # set the model weights
+        self.loaded_model.set_weights(model_weights)
+        tool_sequence = tool_sequence.split(",")
```
jdavcs: This would give you a list with an empty string if `tool_sequence` is '' - which, I guess, is not what you want here. You could use `[]` as the default in line 620. Then instead of lines 620, 661 and 662, you could have this:
```python
tool_sequence = payload.get('tool_sequence', [])
if tool_sequence:
    tool_sequence = tool_sequence.split(',')[::-1]
```
This will give you a reversed list or an empty list.
These are all just minor suggestions.

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -603,6 +609,168 @@ def build_module(self, trans, payload={}):
...
+            self.compatible_tools = json.loads(self.trained_model.get('compatible_tools').value)
+            self.tool_weights = json.loads(self.trained_model.get('class_weights').value)
+            self.tool_weights_sorted = dict()
+            tool_pos_sorted = [int(key) for key in self.tool_weights.keys()]
+            for k in tool_pos_sorted:
+                self.tool_weights_sorted[k] = self.tool_weights[str(k)]
+            self.reverse_dictionary = dict((v, k) for k, v in self.model_data_dictionary.items())
+        model_weights = list()
+        weight_ctr = 0
+        while True:
```
jdavcs: This is an infinite loop that is terminated by an exception. Conceptually, this translates to "go on forever until it breaks".  Instead, I would suggest using an explicit check (e.g., `if foo == bar: break`) that breaks out of the loop (I can't tell from the code what that check would be). This would (a) make the code easier to understand, and (b) wouldn't use exceptions for control flow.

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -603,6 +609,168 @@ def build_module(self, trans, payload={}):
...
+                        prediction_data["children"].extend(item["recommendations"])
+                        break
+
+            # get the root name for displaying after tool run
+            for t_id in self.all_tools:
+                if t_id == last_tool_name:
+                    prediction_data["name"] = self.all_tools[t_id][1]
+                    break
+            return prediction_data
+        except Exception as exp:
```
jdavcs: I think this `try/except` is too long: it's hard to tell what `exp` is supposed to be caught here (it's a catch-all).

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -603,6 +609,168 @@ def build_module(self, trans, payload={}):
...
+            return prediction_data
```
jdavcs: ..and it is not handled. So, essentially, it anything bad happens in this large block of code, it will fail silently and simply return the value. Instead, I think we should narrow down the `try/except`, log the *specific* error and either proceed or fail, depending on what that error is (and be explicit about it).

jdavcs:
Overall, looks like cool stuff! I hope my comments were not too annoying, sorry if they are! 
I didn't dive deep into the code, so here's just a general observation. If you break up very long methods into smaller chunks, it would make the code (a) much easier to read and understand; and (b) more testable. I realize performance may be critical in such code, but unless something is called 10000x of times, I don't believe it will affect performance. So my vote would be for readability.
Also, tests. While we can't test the *output* of a recommender system, we can test its critical functionality. Also, a few unit tests would not only ensure that code does what we expect it to do, they may also help others understand the code better. 
Again, these are only my suggestions.
Cool stuff!

anuprulez:
@ic4f thank a lot for your comments

martenson:
re commit `New model trained with 19_09` - could the model be fetched on demand, and not shipped with Galaxy? It seems prone to get obsolete plus shipping deps like this is not very future facing.

anuprulez:
@martenson I am getting the model from remote now (using a config variable) and not with the repository anymore.

martenson:
@anuprulez thank you!

mvdbeek:
I've bumped this to 20.05. There are some config conflicts here, bit sure the WIP label was intentional ?

bgruening:
@anuprulez can you please take care of this one and remove the WIP label when you are ready.

anuprulez:
Transported to new PR: https://github.com/galaxyproject/galaxy/pull/9440

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -2947,6 +2947,47 @@ mapping:
...
+      topk_recommendations:
+        type: int
+        default: 20
+        required: false
+        desc: |
+          Set the number of predictions/recommendations to be made by the model
+
+      admin_tool_recommendations_path:
+        type: str
+        default: 'config/tool_recommendations_overwrite.yml'
```
jdavcs, line 2975: better:
```yaml
default: 'tool_recommendations_overwrite.yml'
path_resolves_to: config_dir
```
Otherwise it will be inside **a** config directory, not **the** config directory if `config` is overwritten.

All the cases in the schema that use `config/` in the path either mean a specific config relative to root (like `tour_config_dir`) or are handled as special cases here https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L751 .

Sorry, config can be rather confusing - I should add a short guide for devs soon.

PR #8720, opened 2019-09-26
Revise toolshed config dependency check error message in repository install manager

### LINKED ISSUE #8715: Galaxy fails to install tools with dependencies
On dev when I do not set `tool_dependency_dir` in galaxy.yml it should default to `database/dependencies` but instead installation fails.

cc @ic4f @guerler @natefoo 

```
  # Various dependency resolver configuration parameters will have
  # defaults set relative to this path, such as the default conda
  # prefix, default Galaxy packages path, legacy tool shed dependencies
  # path, and the dependency cache directory.  Set the string to None to
  # explicitly disable tool dependency handling. If this option is set
  # to none or an invalid path, installing tools with dependencies from
  # the Tool Shed or in Conda will fail.
  #tool_dependency_dir: database/dependencies
```

`err_msg: "Tool dependencies can be automatically installed only if you set the value of your 'tool_dependency_dir' setting in your Galaxy configuration file (galaxy.ini) and restart your Galaxy server.  "`

guerler:
Follow-up to #8660. Fixes #8715.

### DIFF
--- a/lib/tool_shed/galaxy_install/install_manager.py
+++ b/lib/tool_shed/galaxy_install/install_manager.py
@@ -722,8 +722,6 @@ class InstallRepositoryManager(object):
         install_repository_dependencies = install_options.get('install_repository_dependencies', False)
         install_resolver_dependencies = install_options.get('install_resolver_dependencies', False)
         install_tool_dependencies = install_options.get('install_tool_dependencies', False)
-        if install_tool_dependencies:
-            self.__assert_can_install_dependencies()
         new_tool_panel_section_label = install_options.get('new_tool_panel_section_label', '')
         tool_panel_section_mapping = install_options.get('tool_panel_section_mapping', {})
         shed_tool_conf = install_options.get('shed_tool_conf', None)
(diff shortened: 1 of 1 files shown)


jmchilton (approved):
It feels like the check should be modernized and fixed but eliminated entirely. I'm fine with this though - we don't have any testing to verify the check works clearly so if we decide to do this better later we can add that with tests.

jdavcs (approved):
`app.tool_dependency_dir` *can* be None. I'm afraid I can't estimate whether this check is needed here (I don't know this part of the codebase well enough), so I'll rely on @jmchilton 's opinion.

nsoranzo:
I cannot reproduce #8715, is this still needed after #8696 was merged? If yes, how to reproduce?

nsoranzo:
`app.config.tool_dependency_dir` can be `None`, while I don't think `app.tool_dependency_dir` should be `None` just because `tool_dependency_dir` is not set in `galaxy.yml` .

See https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L1102 and https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tool_util/deps/__init__.py#L83

guerler:
This underlying issue seems to be fixed now by a recent change in dev. Thank you @nsoranzo! Should we still remove the check here or at least revise the message?

nsoranzo:
> Should we still remove the check here or at least revise the message?

I think we can revise the message to something like "Tool dependencies installation is disabled in your configuration files."

nsoranzo (approved):

PR #8728, opened 2019-09-30
Do not overwrite existing attributes

jdavcs:
This is a safeguard against introducing a bug. This edit prevents the
config loading code from overwriting the value of an existing attribute.

The config loading procedure automatically creates attributes of the
GalaxyAppConfiguration object for each property in the schema. If the
schema contains a property with the same name as an existing attribute
(e.g., `reloadable_options`, `schema`, etc.), that attribute will be
overwritten. Worse, if the call to `_set_config_base()` is moved before
the call to `_create_attributes_from_raw_config()`, the values of
`config_dir` and `data_dir` will be set to `None`, because those are
present in the schema. This fix prevents this from happening.

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -225,10 +225,9 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
 
     def _create_attributes_from_raw_config(self):
         for key, value in self._raw_config.items():
-            if not hasattr(self, key):
-                setattr(self, key, value)
-            else:
-                log.debug("Can't override existing attribute %s ", key)
+            if hasattr(self, key):
+                raise ConfigurationError("Attempting to override existing attribute %s ", key)
+            setattr(self, key, value)
 
     def _process_config(self, kwargs):
         self.config_dict = kwargs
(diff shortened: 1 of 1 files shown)


lib/galaxy/config/__init__.py
```diff
@@ -225,7 +225,10 @@ def _update_raw_config_from_kwargs(self, kwargs):
 
     def _create_attributes_from_raw_config(self):
         for key, value in self._raw_config.items():
-            setattr(self, key, value)
+            if not hasattr(self, key):
+                setattr(self, key, value)
+            else:
+                log.debug("Can't override existing attribute %s ", key)
```
```suggestion
                log.debug("Can't override existing attribute '%s'", key)
```
or should that be a full-blown exception instead?

jdavcs: Yeah, probably. I want to think a bit and see if I can come up with a use case for when we want just a warning..  If I can't think of one, I'll change this shortly.

nsoranzo: It may be safer to raise an Exception now and change the code if/when an use case emerges.

jdavcs: Yes, you are right. Changing.

jdavcs: Does this look right?
`raise ConfigurationError("Attempting to override existing attribute %s ", key)`

nsoranzo: I'd write it as:
```python
            if hasattr(self, key):
                raise ConfigurationError("Attempting to override existing attribute '%s'" % key)
            setattr(self, key, value)
```

jdavcs: Yes, of course.

jdavcs:
Error in circleci is unrelated (nondeterministic test)

nsoranzo (approved):

PR #8735, opened 2019-09-30
Fix client formatting

nsoranzo:
Broken when I merged forward 19.09.

List `client-lint` instead of `client-eslint` when running `make` with no target.


dannon:
@nsoranzo Thanks, looks good.

PR #8738, opened 2019-10-01
[19.09] Release notes for 19.09.

jmchilton:


martenson:
I think the notes are in a really good shape, thanks @jmchilton!

martenson:
@natefoo  I've pushed notes regarding the tool configuration changes, please review. I am thinking about shortening it, maybe extracting the full notes into separate document and including only summary?

hexylena:
Will get the user notes wrapped up today.

hexylena:
Added some gifs/etc. If someone could review the user-facing ones, that would be amazing. (I'm running really low on time today and am out for next 3 days.)

PR #8742, opened 2019-10-02
Client build process refinement

dannon:
Tweaks gulp build targets to be more sensible and clear.  This came about looking at the client build process deconstruction issue referenced below; more clear separation of concerns in the gulp tasks.

Drops targets from primary client package that should not be used anymore (we've moved the building of these to the individual packages).

### DIFF
--- a/client/gulpfile.js
+++ b/client/gulpfile.js
@@ -90,9 +90,10 @@ function cleanPlugins() {
     return del(["../static/plugins/{visualizations,interactive_environments}/*"], { force: true });
 }
 
-module.exports.fonts = fonts;
-module.exports.stageLibs = stageLibs;
-module.exports.cleanPlugins = cleanPlugins;
-module.exports.buildPlugins = buildPlugins;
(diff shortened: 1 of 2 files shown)


dannon:
@natefoo Do we have a good way to coordinate these targets being implemented when the next test/whatever rollout happens, relative the older 3 targeted in https://github.com/galaxyproject/ansible-galaxy/pull/85?

natefoo:
Implemented in Ansible you mean? I added version detection there a bit ago for the node version stuff so we should be able to add conditionals for this.

dannon:
@natefoo Yeah, exactly.  When we go to deploy the next rev that contains this code the items (https://github.com/galaxyproject/ansible-galaxy/pull/85/files#diff-7e1ee3914098357abbc81ad7a8d7eacaR15) will just be 'plugins' and 'client'.

natefoo:
Thanks, I created an issue for it in galaxyproject/ansible-galaxy#88.

PR #8746, opened 2019-10-03
Multi subdomains cookie support

julozi:
When attaching several subdomain names to the same Galaxy instance, by default each subdomain gets its own session cookie. This mean that when you are switching from one subdomain to the other, you have to login in back again. It can be quite confusing for the users and some of them believe they need to register again when using a new subdomain.

This PR adds an option to Galaxy configuration to make it possible to share the same session cookie over multiple subdomains.


lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -2799,3 +2799,22 @@ mapping:
           Allow disabling pbkdf2 hashing of passwords for legacy situations.
           This should normally be left enabled unless there is a specific
           reason to disable it.
+
+      use_multidomains:
```
dannon: Maybe consolidate these two options and have the logic be 'if multidomains_root is set, use it'?  Then it's all handled by the single configuration param.

julozi: Indeed, way more simple !

dannon:
You'll want to do a `make config-rebuild` as well (from the Galaxy root directory) and commit the diff of that.  When editing these configs, now we just work on the schema and then run that, which builds the documentation and adds things to the sample correctly.

lib/galaxy/config/sample/galaxy.yml.sample
```diff
@@ -1845,3 +1845,10 @@ galaxy:
   # reason to disable it.
   #use_pbkdf2: true
 
+  # Tell Galaxy that multiple domains sharing the same root are associated
+  # to this instance and wants to share the same session cookie.
+  # This allow a user to stay logged in when passing from one subdomain
+  # to the other.
+  # This root domain will be written in the unique session cookie shared
+  # by all subdomains.
+  #multidomains_root: null
```
hexylena: maybe name it something like `cookie_domain_root`? or something to make it more clear related to the cookies?

julozi: Or just `cookie_domain`, which is also the name of the property in the cookie ?

hexylena: +1!

dannon: Yeah, that's perfect.  Once you swap that and rebuild the config this is good to go IMO; nice improvement here.

dannon:
@galaxybot test this

dannon:
I pushed a minor fix that makes the configuration change compatible with the reports and toolshed webapps.

dannon:
@galaxybot test this

dannon:
@julozi Thanks again for the contribution!

martenson:
It would be nice for this to make it to the admin docs at http://docs.galaxyproject.org

dannon:
@martenson I was thinking the autogenerated information from the schema was enough for a relatively niche option like this, but if you want to add something where you think it makes more sense I'm happy to merge it.

PR #8749, opened 2019-10-03
axios-mock-adapter for cleaner, more expansive testing

jmchilton:
Now testing the URL generated properly and we're skipping the rewrite of internals of the job metrics store - both of which I think are nice upshots. For my money the whole thing is cleaner also. I think this test is better now, but I'll admit there is always a cost to introducing new dependencies.


dannon:
Definitely cleaner than rewire -- good find!

dannon:
Unrelated jenkins failure.

Issue #8751, opened 2019-10-03
19.09 Reports Fixes


jmchilton:
More things to fix in 19.09.

- Hide run button when editing report.
- Add small syntax guide somewhere in the GUI.
- Fix syntax for spaces in input/output names.

All reported by @erasche.

PR #8753, opened 2019-10-03
[19.09] More invocation report fixes.

jmchilton:
Fixes https://github.com/galaxyproject/galaxy/issues/8751.

- Hide run/settings button when editing report - since saving won't work anyway. Less jarring and won't result in broken workflows.
- Add small syntax guide to the invocation report editor.
- Fix invocation syntax for spaces in input/output names.


bgruening (approved):
:) Cool. Restarted test and now everything is green.

Thanks John!

PR #8757, opened 2019-10-04
Allow space, dot and comma in relabel identifier tool

mvdbeek:
These characters are allow in dataset names, so I don't think it makes sense to restrict them at this level. Alternative to #8571.


innovate-invent:
Keep in mind that currently there is no restriction on collection IDs from the upload tool.
If you don't include `[` and `]` a user can't download datasets, then reupload them as a collection without renaming all of the files.

mvdbeek:
Those aren't really reasonable characters to **set** as collection identifiers IMO, and uploading and building a list doesn't involve the RelabelFromFileTool.

innovate-invent:
There should be uniform handling of collection identifiers across the code-base. As it stands you can't use the relabel tool to modify existing identifiers adding suffixes.

jmchilton:
Thanks!

PR #8758, opened 2019-10-04
[WIP][19.09] Fix interactive tools on Kubernetes.

jmchilton:
I think this approach has some chance of working if Galaxy is running in the same Kubernetes environment as the jobs - certainly not going to work for a Galaxy external to the Kubernetes job environment.

Testing and debugging is going to be... a thing...

### DIFF
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1123,6 +1123,14 @@ class JobWrapper(HasResourceParameters):
                             job.id)
 
     @property
+    def guest_ports(self):
+        if self.interactivetools:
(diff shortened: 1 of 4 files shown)


martenson:
I think this shouldn't be release blocking. ITs are an unstable feature and expected to change a lot in the next months with *maybe* being in Beta in 20.01

afgane:
@jmchilton could we please get an update on the plans for this functionality? 

Also, in terms of the current functionality, if we cherry picked this change, should we expect the ITs to be functional for a demo at least?

jmchilton:
Nope - I wouldn't expect this to work at all. It hasn't been tested - I'm not really sure how to test it. It is at best a very broad outline of one way to potentially try to implement this.

PR #8764, opened 2019-10-07
Fix visibility of tool panel footer

### LINKED ISSUE #8761: Tool panel cannot be hidden?
Teaching a training today we noticed that the tool panel seems to be lacking controls for resizing or hiding?

We often hid this to show e.g. fastqc reports. We either would hide both panels, or just the tool panel. Hiding the history panel alone was uninteresting, as then we lost the context of "which report we're looking at" that was being communicated to the students. Hiding the tool panel was appealing because it was a lot of visual noise unrelated to what we wanted to communicate?

Was this done intentionally? Is there some reason we should not be able to hide that?

scrathat:
Fixes #8761

### DIFF
--- a/client/galaxy/scripts/components/ToolBox.vue
+++ b/client/galaxy/scripts/components/ToolBox.vue
@@ -142,5 +142,6 @@ export default {
     display: flex;
     flex-flow: column;
     height: 100%;
+    overflow: auto;
 }
 </style>


dannon:
Thanks for the fix.  This does resolve the issue for now and is good fix for 19.09.  Down the road we should refactor unified panel handling so that they don't require a local override for styling and the extra wrapper, when used like they currently are in this file.

dannon:
(opening backport...)

PR #8773, opened 2019-10-08
[19.09] Fix (and simplify) dataset calculation.

dannon:
This fixes (re)calculating and setting user disk_usage.  I don't know exactly *why* the other version of this code wasn't correctly persisting the newly computed values, but this one does, and it's a little simpler.

Additionally it fixes the method to where it always returns the newly computed value instead of having to return None due to the lack of a RETURNING clause in sqlite.

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -520,8 +520,8 @@ class User(Dictifiable, RepresentById):
 
     def _calculate_or_set_disk_usage(self, sa_session, dryrun=True):
         """
-        Utility to calculate (returning a value) or just set the disk usage
-        (returning None / applying immediately)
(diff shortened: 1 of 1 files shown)


hexylena:
Untested but looks like a nice simplification!

lib/galaxy/model/__init__.py
```diff
@@ -524,7 +524,7 @@ def _calculate_or_set_disk_usage(self, sa_session, dryrun=True):
         (returning None / applying immediately)
```
nsoranzo: Need to update docstring here?

dannon: Yep.

nsoranzo (approved):

natefoo (approved):
Thanks @dannon!

PR #8776, opened 2019-10-09
Cleanup SQL query in ``_calculate_or_set_disk_usage()``

nsoranzo:
In particular use `NOT purged` instead of `purged = '0'`.


martenson:
I am not sure but didn't we establish that does not work on sqlite?

nsoranzo:
> I am not sure but didn't we establish that does not work on sqlite?

No, what doesn't work on SQLite is `purged = false`, see https://github.com/galaxyproject/galaxy/pull/8552/files

I tested on SQLite and `NOT purged` does work as expected.

dannon:
@nsoranzo Thanks, looks good.

PR #8778, opened 2019-10-09
[19.09] Force a flush after user disk usage recalculation.

dannon:
xref #8773 -- this seems to work fine locally with or without the flush, but on main we were seeing it not apply after the transaction was finished.

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -547,6 +547,7 @@ class User(Dictifiable, RepresentById):
         usage = sa_session.scalar(sql_calc, {'id': self.id})
         if not dryrun:
             self.set_disk_usage(usage)
+            sa_session.flush()
         return usage
 
     @staticmethod


qiagu:
I believe this will finally solve the problem on Main. Great work @dannon

lib/galaxy/model/__init__.py
```diff
@@ -547,6 +547,7 @@ def _calculate_or_set_disk_usage(self, sa_session, dryrun=True):
         usage = sa_session.scalar(sql_calc, {'id': self.id})
         if not dryrun:
             self.set_disk_usage(usage)
+            sa_session.flush()
```
nsoranzo, line 550: Should this flush be inside the `set_disk_usage()` method instead?
It doesn't make a practical difference, given that `set_disk_usage` seems to be called only from here at the moment.

nsoranzo (approved):

PR #8779, opened 2019-10-09
[19.09] Fix to 'portable' metadata for output metadata files.

jmchilton:
Isolating the changes produced by set-metadata to metadata/ fixes Pulsar remote metadata handling. The changes were always meant to keep metadata isolated to that directory anyway - so this is a good conceptual fix as well.


natefoo (approved):
Works for me and it looks like those test errors are unrelated?

Issue #8769, opened 2019-10-08
19.09 error: lazy load operation of attribute 'tool_dependencies' cannot proceed


bgruening:
This is happening unregularly with 19.09. We have seen this a few times during various workshops this week.

```
Traceback (most recent call last):
  File "/opt/galaxy/server/lib/galaxy/jobs/runners/__init__.py", line 237, in prepare_job
    stderr_file=stderr_file,
  File "/opt/galaxy/server/lib/galaxy/jobs/runners/__init__.py", line 278, in build_command_line
    stderr_file=stderr_file,
  File "/opt/galaxy/server/lib/galaxy/jobs/command_factory.py", line 71, in build_command
    __handle_dependency_resolution(commands_builder, job_wrapper, remote_command_params)
  File "/opt/galaxy/server/lib/galaxy/jobs/command_factory.py", line 182, in __handle_dependency_resolution
    if local_dependency_resolution and job_wrapper.dependency_shell_commands:
  File "/opt/galaxy/server/lib/galaxy/jobs/__init__.py", line 920, in dependency_shell_commands
    job_directory=self.working_directory
  File "/opt/galaxy/server/lib/galaxy/tools/__init__.py", line 1646, in build_dependency_shell_commands
    installed_tool_dependencies=self.installed_tool_dependencies,
  File "/opt/galaxy/server/lib/galaxy/tools/__init__.py", line 1657, in installed_tool_dependencies
    installed_tool_dependencies = self.tool_shed_repository.tool_dependencies_installed_or_in_error
  File "/opt/galaxy/server/lib/galaxy/model/tool_shed_install/__init__.py", line 418, in tool_dependencies_installed_or_in_error
    for tool_dependency in self.tool_dependencies:
  File "/opt/galaxy/venv/lib/python2.7/site-packages/sqlalchemy/orm/attributes.py", line 282, in __get__
    return self.impl.get(instance_state(instance), dict_)
  File "/opt/galaxy/venv/lib/python2.7/site-packages/sqlalchemy/orm/attributes.py", line 710, in get
    value = self.callable_(state, passive)
  File "/opt/galaxy/venv/lib/python2.7/site-packages/sqlalchemy/orm/strategies.py", line 688, in _load_for_state
    % (orm_util.state_str(state), self.key)
[...]

PR #8780, opened 2019-10-10
[19.09] Fix detached parent instance error in 8769

mvdbeek:
It's sort of expected that using the `ToolShedRepositoryCache` we can run into the situation where the original session from which the `ToolShedRepository` objects are loaded can be closed.

We normally work around that case in https://github.com/galaxyproject/galaxy/blob/7c5e72a7443d85ef50990f44fec2e173433cbb4f/lib/galaxy/tools/cache.py#L185

https://github.com/galaxyproject/galaxy/issues/8769 occurs when accessing attributes of `ToolShedRepository` objects, so I assume that's why the `except DetachedInstanceError` try/except logic doesn't work.

The workaround is to eagerload all `ToolShedRepository` attributes up front. I think medium term I'd like to redesign the whole install db system.


nsoranzo:
Can we get rid of the `rebuild()` call then?

mvdbeek:
Yeah, we can probably get rid of it. We still need to rebuild the cache occasionally, but I'd assume we can't run into the exception anymore.

lib/galaxy/tools/cache.py
```diff
@@ -173,24 +166,7 @@ def rebuild(self):
...
-            self.rebuild()
-            return self._get_installed_repository(tool_shed=tool_shed,
-                                                  name=name,
-                                                  owner=owner,
-                                                  installed_changeset_revision=installed_changeset_revision,
-                                                  changeset_revision=changeset_revision,
-                                                  repository_id=repository_id)
-
-    def _get_installed_repository(self, tool_shed=None, name=None, owner=None, installed_changeset_revision=None, changeset_revision=None, repository_id=None):
+     def get_installed_repository(self, tool_shed=None, name=None, owner=None, installed_changeset_revision=None, changeset_revision=None, repository_id=None):
```
nsoranzo: Indentation is 5 spaces here.

mvdbeek: Oops, thanks!

nsoranzo (approved):

martenson:
I am not a fan of the second commit going to release. Could we move that one to dev please?

mvdbeek:
sure, can you cherry-pick and rebase this? I think eu needs this asap and I'm about to change trains.

martenson:
sure thing

bgruening:
I think we need to revert this one. It increases the memory usage of handler 4 fold.

grafik

handler3 is running the reverted Galaxy codebase. Handler 0,1,2 are running this patch. As you can see the memory consumption is 4 times higher :(

Sorry!

Issue #8781, opened 2019-10-10
Data libraries list is not scrollable when overflowing the screen


hexylena:
Due to new CSS it seems that if the data lib list is longer than the page, you can't scroll within that page to reach the bottom and must search/zoom out.

martenson:
I found the culprit I think. 
![DevTools_-_127_0_0_1_library_list](https://user-images.githubusercontent.com/1814954/66579614-1641db00-eb4b-11e9-977e-de69b4cf4c47.png)

martenson:
Oh, and it is present on 19.09 so we need to fix for release I think.

martenson:
is this possible? https://github.com/galaxyproject/galaxy/blob/dev/client/galaxy/style/scss/base.scss#L158 https://github.com/galaxyproject/galaxy/blob/dev/config/plugins/visualizations/hyphyvision/static/main.css#L7756

martenson:
This is the fix, but I don't know what I am breaking.

```diff
diff --git a/client/galaxy/style/scss/base.scss b/client/galaxy/style/scss/base.scss
index e605f95e06..529bbfb3fb 100644
--- a/client/galaxy/style/scss/base.scss
+++ b/client/galaxy/style/scss/base.scss
@@ -134,7 +134,7 @@ body {
 #right {
     @extend .position-absolute;
     @extend .h-100;
-    @extend .overflow-hidden;
+    // @extend .overflow-hidden;
 }
 #left {
     @extend .unified-panel;
@@ -154,7 +154,7 @@ body {
     }
 }
 #center {
-    @extend .overflow-hidden;
+    // @extend .overflow-hidden;
     left: $panel-width;
     right: $panel-width;
     .center-container {
```

PR #8783, opened 2019-10-10
Data Libraries scrollbar fix

inkuzmin:
Fixes the bug we've discovered during the Galaxy training, specifically, not working scrolling at the data libraries page.

Tested in Google Chrome v77.0.3865.90 and in Firefox 69.0.

### DIFF
--- a/client/galaxy/scripts/mvc/library/library-dataset-view.js
+++ b/client/galaxy/scripts/mvc/library/library-dataset-view.js
@@ -107,6 +107,8 @@ var LibraryDatasetView = Backbone.View.extend({
         $(".tooltip").remove();
         var template = this.templateDataset();
         this.$el.html(template({ item: this.model }));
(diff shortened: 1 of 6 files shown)


martenson:
I think we should fix it up in the cascade instead. Hiding scrollbars cannot be the default. https://github.com/galaxyproject/galaxy/issues/8781#issuecomment-540622453

inkuzmin:
@martenson it would be great indeed! I haven't seen the bug report, my assumption was just that it's been set to "hidden" with the "important" declaration for a good (although unknown to me) reason.

martenson:
@inkuzmin that was a hack, but we cannot add this hack to every block element and call it a day, I hope 😭

martenson:
I opened a more general bugfix linked above, thanks for opening this @inkuzmin !

inkuzmin:
@martenson thanks!

PR #8785, opened 2019-10-10
Drop DetachedInstanceError workaround

martenson:
This shouldn't be needed anymore now that we eagerload all relationship attributes.

xref https://github.com/galaxyproject/galaxy/pull/8780


nsoranzo:
I've pushed https://github.com/galaxyproject/galaxy/pull/8780 forward and rebased this on top.

martenson:
@nsoranzo thanks!

PR #8786, opened 2019-10-10
Specify height for library container, fix scrolling

guerler:


martenson:
this adds scrollbar to the inner element (i.e. the scrollbar is not on the right side of the browser window). I really think we need to root out the thing that forces hiding scrollbars on #center with a `!important` clause.

PR #8787, opened 2019-10-10
[19.09] remove inherited #center overflow

martenson:
this broke libraries, and generally shouldn't be used
with such scope I believe

It is unclear what is purpose of these rules + there are other like this in the `base.scss`.

fixes https://github.com/galaxyproject/galaxy/issues/8781
supersedes https://github.com/galaxyproject/galaxy/pull/8783

### DIFF
--- a/client/galaxy/style/scss/base.scss
+++ b/client/galaxy/style/scss/base.scss
@@ -134,7 +134,6 @@ body {
 #right {
     @extend .position-absolute;
     @extend .h-100;
-    @extend .overflow-hidden;
 }
 #left {
     @extend .unified-panel;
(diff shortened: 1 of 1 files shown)


dannon:
@guerler Do you remember why this was added?

guerler:
Some legacy attributes we kept till now but previously used without the important tag. https://github.com/galaxyproject/galaxy/blame/a1ebbe4528b779ca394af3d56847456142b8bab7/static/style/src/less/base.less#L122

martenson:
Big switch was from `hidden` to `hidden !important`, but overall base.scss needs to burn.

guerler (approved):

guerler:
Thanks for this fix @martenson.

PR #8789, opened 2019-10-10
[19.09] improve tool panel searching

martenson:
Fixes and improvements I've found while looking in too panel search pains.

- use logical OR to join parsed search terms
instead of default AND; however use scaling factor of 0.9
to make sure multi-term-hits are prioritized
- cast every query to lowercase, under some circumstances
the results would differ before
- do not prepend a leading `*` wildcard for the individual terms, this should make things much faster on a big index, it will also yields less, but better results

### DIFF
--- a/lib/galaxy/tools/search/__init__.py
+++ b/lib/galaxy/tools/search/__init__.py
@@ -19,6 +19,7 @@ from whoosh.filedb.filestore import (
     RamStorage
 )
 from whoosh.qparser import MultifieldParser
+from whoosh.qparser import OrGroup
 from whoosh.scoring import BM25F
 
 from galaxy.util import ExecutionTimer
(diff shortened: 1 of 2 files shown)


lib/galaxy/tools/search/__init__.py
```diff
@@ -130,15 +133,17 @@ def search(self, q, tool_name_boost, tool_section_boost, tool_description_boost,
             )
         )
         # Set query to search name, description, section, help, and labels.
-        parser = MultifieldParser(['name', 'description', 'section', 'help', 'labels', 'stub'], schema=self.schema)
+        og = OrGroup.factory(0.9)
```
bgruening, line 141: @martenson can we add a comment here and explain what the factory and the 0.9 are doing.
"The Or query will score documents higher that contain multiple of the query term/token. The 0.9 is a scaling factor that indicates how much better a document is rated. "

martenson:
It is explained in the whoosh docs that anybody touching this needs to
read, but I will add a link here.

lib/galaxy/tools/search/__init__.py
martenson: docs added

jmchilton (approved):

dannon:
Errors unrelated.  Thanks!

PR #8737, opened 2019-10-01
Start using path_resolves_to; add integration tests

jdavcs:
Summary:
This is the first attempt to use the `path_resolves_to` schema attribute to automate path resolution: 
- a statement of the form `self.foo = os.path.join(self.data_dir, self.foo)` is removed from `config/__init__.py`
- attribute `path_resolves_to: data_dir` is added to foo's definition in config_schema.yml.
As a result, path resolution logic is located in one place only, and is stated declaratively.

A few issues have been taken care of (#8728, https://github.com/galaxyproject/galaxy/pull/8689/commits/491a8bf3e76c3d7632fb89e433309cbb8b75da06 ).

Other modifications:
- Added integration tests to ensure that paths are resolved correctly for config options that use the `path_resolves_to` attribute.
- Added integration tests for a few config options to ensure schema defaults are assigned.
@jmchilton is this a reasonable approach to test this? If so, I'll add the rest.

Also ping @natefoo, @nsoranzo .

PR #8793, opened 2019-10-12
One test case to rule them all

jdavcs:
This is based on #8737 (one new commit).

This adds integration tests for all configuration properties specified in the schema. Out of these, 226 pass locally (tested on 2 machines, so I hope the results will be the same on Jenkins); another 56 are commented out because they require edits to code outside of this file (I will address them in a follow-up). Added time: 18 sec (when running locally). [updated: 229->226/53->56: see last commit]

This commit does not change anything in the code other than adding this file. Its purpose is to establish a ground truth prior to any further refactoring of the configuration system (and, thus, catch any unintended side effects during subsequent edits). Together with the recently added unit tests, this should make modifying configuration-related code a little safer, I hope.

WHAT IS TESTED
This tests (1) automatic creation of configuration properties; and (2) assignment of default values that are specified in the schema and, in some cases, are also processed at load time (paths resolved, csv strings converted to lists, etc). 
For example:
- for property `foo` defined in the schema with a default value of `42`, the test will ensure that `app.config.foo == 42`
- if `foo`'s default value is further processed at load time (e.g., listified), the test would ensure that
`app.config.foo == [42]`, (or `== [42, 43, 44]` if the default was a csv str `42,43,44`
- if `foo`'s default was not set, the test ensures its value is null
- if `foo`'s default was not set, but a default value was assigned at load time, the test will test that value. And so forth. [for more details, see comment at the top of the new file]

I based this on top of my previous PR (as opposed to `dev`) because some of the follow-up commits will use the new `path_resolve_to` feature from #8689. And even if they didn't, basing this on `dev` would lead to 2 separate branches, each with many edits to the same set of files.

### DIFF
--- a/test/integration/conftest.py
+++ b/test/integration/conftest.py
@@ -1,0 +1,8 @@
+def pytest_unconfigure(config):
+    try:
+        # This needs to be run if no test were run.
+        from .test_config_defaults import DRIVER
+        DRIVER.tear_down()
+        print("Galaxy test driver shutdown succesfull")
+    except Exception:
+        pass
--- a/test/integration/test_config_defaults.py
+++ b/test/integration/test_config_defaults.py
@@ -1,13 +1,86 @@
+import os
+from datetime import timedelta
+
 from base import integration_util
 
 
 class ConfigDefaultsTestCase(integration_util.IntegrationTestCase):
     """
-    Test automatic creation of configuration properties and assignment of
-    default values specified in the schema.
+    This tests: (1) automatic creation of configuration properties; and
+    (2) assignment of default values that are specified in the schema and, in
+    some cases, are also processed at load time (paths resolved, csv strings
+    converted to lists, etc).
+
+    This test case should have tests for ALL properties listed in the main
+    schema file. Whenever a property's default value is changed (edits to
+    schema or configuration loading procedures), this test code must be
+    modified to reflect that change.
+
+    Test assumptions for a default configuration:
+    - If a default is set and not modified at load time, expect schema default.
+    - If a default is not set, expect null.
+    - If a default is set and modified at load time, the test should reflect that
+      (if a default is specified in the schema, it is expected that it will be used
+      in some form at load time; otherwise it should not be listed as a default).
+
+    Configuration options NOT tested:
+    - config_dir (value overridden for testing)
+    - data_dir (value overridden for testing)
+    - new_file_path (value overridden for testing)
+    - logging (mapping loaded in config/; TODO)
+    - dependency_resolution (nested properties; TODO)
+    - job_config (no obvious testable defaults)
     """
+
+    def __init__(self, *args, **kwargs):
+        super(ConfigDefaultsTestCase, self).__init__(*args, **kwargs)
+        # use lazy loading for attributes below
+        self._root_dir = None
+        self._config_dir = None
+        self._data_dir = None
+        self._tool_data_path = None
+
     def get_default(self, key):
-        return self._app.config.appschema[key]['default']
+        # Returns default value or None (because if default value is not
+        # specified in the schema, we expect None  by default)
+        return self._app.config.appschema[key].get('default', None)
+
+    def get_csv_default_as_list(self, key):
+        # Use when schema default is a comma-separated values list
+        # which is converted to a list
+        return self._app.config.appschema[key]['default'].split(',')
+
+    def get_default_in_root_dir(self, key):
+        # Use when schema default is resolved with respect to the installation's root directory
+        self._root_dir = self._root_dir or self._app.config.root
+        return self._resolve(self._root_dir, self.get_default(key))
+
+    def get_default_in_config_dir(self, key):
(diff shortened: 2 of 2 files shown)


jdavcs:
Rebased to include bd9f188 which fell through the cracks in a series of rebases.

jdavcs:
Rebased to exclude 3 tests that test values that are overridden for testing on non-sqlite dbs. Locally they pass of course; but not on Jenkins.

test/integration/test_config_defaults.py
```diff
@@ -0,0 +1,1107 @@
...
+    # def test_default_database_connection(self):
+    # TODO: untestable; refactor config/__init__ to test
+
+    # def test_default_database_engine_option_pool_size(self):
+    # TODO: default value overridden for tests runnign on non-sqlite databases
+
+    # def test_default_database_engine_option_max_overflow(self):
+    # TODO: default value overridden for tests runnign on non-sqlite databases
+
+    def test_default_database_engine_option_pool_recycle(self):
```
mvdbeek: Could you parametrize all the `test_.*` methods using @pytest.mark.parametrize ?
I'd build a list of config strings up front (or even discover them from the schema) and then parametrize a generic function, i.e
```python
CONFIG_PROPERTIES = [
    'database_engine_option_server_side_cursors',
    'database_engine_option_pool_recycle',
]


@pytest.mark.parametrize('config_property', CONFIG_PROPERTIES)
def test_default_config_property(config_property):
    expect = self.get_default(config_property)
    assert expect == getattr(self._app.config, config_property)
```

If you can auto-discover these from the schema: even better :)

jdavcs: Probably yes. Marking this as WIP for now, while I'm consolidating tests.

jdavcs: @mvdbeek how's this? I can barely stop myself from adding 'one test to find them' to the commit message :)

jdavcs (commented):
The new commit gets rid of 915 lines of code. Thank you for the suggestion, @mvdbeek! 
The approach is the same as here: https://github.com/galaxyproject/galaxy/commit/9e07d9cc062bf8a21bfab72c2505191e1d3c86e4#diff-cb4f71ec7457ece91d85d328b489f3da

I've tried to make this as least hackish as possible. The benefits, I think, outweigh the relative ugliness: 227 tests got reduced to just one simple parametrized test. Also, if an option (or a group of options) require a special way to calculate their expected value, that's done by simply defining a function, and adding a new item to the `CUSTOM` dictionary. Of course, a regular test can be added too: as long as it's added after the generic test, it can use `DRIVER` to access the config object.

jdavcs:
I just realized this overrides a revert in my previous merged PR https://github.com/galaxyproject/galaxy/commit/b962807c8214ee07e82b2b4e6e2c4219081f26e4
Fixed and rebased.

nsoranzo:
> The new commit gets rid of 915 lines of code. Thank you for the suggestion, @mvdbeek!

Maybe it could make sense to squash (some of) the commits.

jdavcs:
@nsoranzo Yes, will do. Apparently, I got 4 extra commits in during a rebase (not my commits). 
And done.

test/integration/conftest.py
```diff
@@ -0,0 +1,8 @@
+def pytest_unconfigure(config):
+    try:
+        # This needs to be run if no test were run.
```
mvdbeek, line 3: That doesn't look quite right but is probably necessary because these tests are a mix of unittest and pytest. Normally all fixtures should cleanup after themselves.

test/integration/test_config_defaults.py
```diff
@@ -1,22 +1,191 @@
...
+    'user_library_import_symlink_whitelist': listify,
+    'tool_filters': listify,
+    'tool_label_filters': listify,
+    'tool_section_filters': listify,
+    'persistent_communication_rooms': listify,
+}
+
+# TODO: split into (1) do not test; and (2) todo: fix and test
+DO_NOT_TEST = [
+    'config_dir',  # value overridden for testing
```
mvdbeek, line 67: So now that we have a nice getter/setter system we could probably also record when something is not the default value.
Not for this PR of course.

mvdbeek (approved):
Very nice! Ready to merge @ic4f ?

test/integration/conftest.py
jdavcs: We do need it, I think (and this one should be pure pytest): if you keep that test, but don't pass any data, then the fixture will be called, but no tests will be generated, and the unconfigure hook will clean up. Remove the hook, and the fixture will keep running until you kill it manually.

test/integration/test_config_defaults.py
jdavcs: Yes, I'm sure we can! And I'm certainly not done with this -- but yes, that would be for a subsequent PR. (as well as cleaning up/normalizing the getter setter system)

jdavcs:
> Very nice! Ready to merge @ic4f ?

Yes!

PR #8795, opened 2019-10-12
Fix some integration tests for config values

jdavcs:
Intended as [WIP].
This is a catch-all for configuration-related refactorings and test fixes, while subsequent commits logically fit within one PR. I'll keep this open as a WIP, cherry-picking individual commits into separate PRs. 

Initially, I planned to accumulate similar modifications and submit as one PR, but now I think it's much easier to deal with (review, merge, etc.) one commit at a time. I'll still keep this open as a temporary "testing ground".

See commit messages for details on modifications.


jdavcs:
Ping @dannon , @VJalili : I assume the config files for oidc should be `xml`, not `yml`, right? 
The schema says xml, the sample files are xml, `authnz/managers.py` and `authnz\__init__.py` both say xml, but `config/__init__.py` tries to load yml (https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L677 ). 
I've traced the change to this commit:
https://github.com/galaxyproject/galaxy/commit/5c5197e84120551b551164f6b73eea2df260a157

I've changed this back to xml, is this correct?

jdavcs:
@jmchilton : I assume the `config/` prefix doesn't belong in this assignment:
https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L686
Unless the intention was to have it inside `config/config/`? (because
`_in_config_dir()` places that value inside the `config_dir`). I'll add it back if it should be there.

VJalili:
@ic4f  correct, oidc configs are all xml.

jdavcs:
Closing this one as all edits have been implemented and merged as separate PRs. One last commit (63ff8a0) will be part of a follow-up that will include similar edits.

Issue #8796, opened 2019-10-13
k8s Job falling in unknown state potentially resolved!


pcm32:
One of the long standing, rarest and very difficult to reproduce bugs in the k8s runner is that a job rarely could enter a state that couldn't be resolved by the runner. We used to have a condition 

https://github.com/galaxyproject/galaxy/commit/6fc3d09b60d98e2daf8285e7ff8864edb07d764c#diff-286f7cbfde32d70afa51e89060189342L448

signalling this in an older version of the runner, but during the past changes since 19.05 it was removed (happily my containers are still at that version and I managed to finally catch the error live to understand it).

What happens here 

https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/jobs/runners/kubernetes.py#L423

is that to fill the numbers of `succeeded`, `active` or `failed` we rely on the `status` hash being filled. However, if the runner asks for the status too quickly after submission, what seems to happen is that the API server doesn't have time to fill in the the status values and returns an empty `status` dictionary:

```
galaxy.jobs.runners.kubernetes INFO 2019-10-13 00:31:10,546 [p:425,w:2,m:0] [KubernetesRunner.monitor_thread] Kubernetes job 'galaxy-my-instance-952' not classified as succ., active or failed. Full Job object:
[...]

PR #8799, opened 2019-10-14
Handles k8s job recently created with no status set yet

pcm32:
Addresses #8796 by keeping the non-started job state when the object has no status set.

### DIFF
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -426,7 +426,7 @@ class KubernetesJobRunner(AsynchronousJobRunner):
             # had time to fill in the object status since the
             # job was created only too recently.
             if len(job.obj['status']) == 0:
-                return job_state            
+                return job_state
             if 'succeeded' in job.obj['status']:
                 succeeded = job.obj['status']['succeeded']
             if 'active' in job.obj['status']:
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -402,14 +402,6 @@ class KubernetesJobRunner(AsynchronousJobRunner):
                                                     namespace=self.runner_params['k8s_namespace'])
         if len(jobs.response['items']) == 1:
             job = Job(self._pykube_api, jobs.response['items'][0])
-            # Check if job.obj['status'] is empty,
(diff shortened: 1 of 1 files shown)


lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -402,6 +402,14 @@ def check_watched_item(self, job_state):
                                                     namespace=self.runner_params['k8s_namespace'])
         if len(jobs.response['items']) == 1:
             job = Job(self._pykube_api, jobs.response['items'][0])
+            # Check if job.obj['status'] is empty,
```
mvdbeek: Can you move this to line 431 ?
It seems like that would be more coherent.
Is this response documented anywhere ?
Is this an empty list or an empty string ?

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -402,6 +402,14 @@ def check_watched_item(self, job_state):
...
+            # return job_state unchanged if this is the case
+            # as probably this means that the k8s API server hasn't
+            # have time to fill in the object status since the
```
```suggestion
            # had time to fill in the object status since the
```

lib/galaxy/jobs/runners/kubernetes.py
nuwang: I too was wondering whether this needs an upstream bug report in k8s or something.

pcm32: Well, I don't think that this is really a failure, I have seen k8s filling API objects progressively as information becomes available by the API server in other cases.

pcm32: Sure @mvdbeek, I can move to 431, but none of what happens in the middle is really needed if the list is empty, are you happy with that?

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -420,6 +420,13 @@ def check_watched_item(self, job_state):
             else:
                 max_pod_retries = 1
 
+            # Check if job.obj['status'] is empty,
+            # return job_state unchanged if this is the case
+            # as probably this means that the k8s API server hasn't
+            # had time to fill in the object status since the
+            # job was created only too recently.
+            if len(job.obj['status']) == 0:
+                return job_state
```
```suggestion
                return job_state
```

mvdbeek (approved):

PR #8800, opened 2019-10-14
Allow recording more performance statistics to statsd.

jmchilton:
- If ``database_log_query_count`` is set, record `sql.<route>` as time spent in database queries across the API request and a statsd counter for `sqlqueries.<route>` as the number of sqlalchemy queries executed. These statistics were super useful in optimizing tool execution code for all the PRs that came out of the work on #6563, I recovered them from logs in the past though.
- Migrate a bunch of ExecutionTimers I used for profiling this code into new StructuredExecutionTimers that now record their timings into statsd. Prefix the random timings with the `internal.<python_module>` to come up with something of a namespace to identify these timings.
- Add new timings for job running, job handler, and workflow scheduling activities to try to start getting a grasp on the throughput of that code in production.

### DIFF
--- a/lib/galaxy/jobs/runners/__init__.py
+++ b/lib/galaxy/jobs/runners/__init__.py
@@ -2,7 +2,6 @@
 Base classes for job runner plugins.
 """
 import datetime
-import logging
 import os
 import string
 import subprocess
@@ -39,10 +38,11 @@ from galaxy.util import (
     unicodify,
 )
 from galaxy.util.bunch import Bunch
+from galaxy.util.logging import get_logger
 from galaxy.util.monitors import Monitors
 from .state_handler_factory import build_state_handlers
 
-log = logging.getLogger(__name__)
+log = get_logger(__name__)
 
 STOP_SIGNAL = object()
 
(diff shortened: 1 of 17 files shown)


mvdbeek (approved):
Very nice! Should we add this option to the config schema or leave it undocumented ?

mvdbeek:
@jmchilton there's a conflict now, if you can resolve it I"m happy to merge

jmchilton:
Good point - added information to the config schema. I suspect we will run it in production on usegalaxy* and find it to not be a problem, if we do we should re-visit this writeup.

VJalili:
- Can you elaborate how this computes performance delta? 
- How does it differentiate between performance degradation due to execution environment load and introduced code performance penalty? 
- Does wait times of job scheduling affect performance monitor?

lib/galaxy/jobs/runners/__init__.py
```diff
@@ -129,7 +129,13 @@ def run_next(self):
             except Exception:
                 name = 'unknown'
             try:
+                action_str = 'galaxy.jobs.runners.%s.%s' % (self.__class__.__name__.lower(), name)
+                action_timer = self.app.execution_timer_factory.get_timer(
+                    'internals.%s' % action_str,
+                    'job runner action %s for job ${job_id} executed' % (action_str)
+                )
                 method(arg)
+                log.trace(action_timer.to_str(job_id=job_id))
```
VJalili, line 138: so if one wants to see if there are any performance degradation introduced by a PR, then they would need to scan/parse the logs?

jmchilton:
> Can you elaborate how this computes performance delta?

This doesn't. These are just statistics I previously parsed out of logs manually to see differences when optimizing that now are captured in a more structured way.

PR #8802, opened 2019-10-15
Stop trying to guess a web port

jdavcs:
This is a call to a nonexistent method.

It was removed here:
https://github.com/galaxyproject/galaxy/commit/d187b289518288eb595ccc777bf67bce716defd6

### DIFF
--- a/lib/galaxy/visualization/plugins/interactive_environments.py
+++ b/lib/galaxy/visualization/plugins/interactive_environments.py
@@ -199,7 +199,7 @@ class InteractiveEnvironmentRequest(object):
         }
 
         web_port = self.attr.galaxy_config.galaxy_infrastructure_web_port
-        conf_file['galaxy_web_port'] = web_port or self.attr.galaxy_config.guess_galaxy_port()
+        conf_file['galaxy_web_port'] = web_port
 
         if self.attr.viz_config.has_option("docker", "galaxy_url"):
             conf_file['galaxy_url'] = self.attr.viz_config.get("docker", "galaxy_url")


lib/galaxy/visualization/plugins/interactive_environments.py
```diff
@@ -199,7 +199,7 @@ def get_conf_dict(self):
         }
 
         web_port = self.attr.galaxy_config.galaxy_infrastructure_web_port
-        conf_file['galaxy_web_port'] = web_port or self.attr.galaxy_config.guess_galaxy_port()
+        conf_file['galaxy_web_port'] = web_port
```
nsoranzo, line 202: It seems `self.attr.galaxy_config.galaxy_infrastructure_web_port` can be `None` (see https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L583-L584 ), but also the (now removed) `guess_galaxy_port()` could return `None`, so I guess it is not a problem here?

jmchilton:
This should target release_19.09 right?

jmchilton (approved):

lib/galaxy/visualization/plugins/interactive_environments.py
jdavcs: `galaxy_config.galaxy_infrastructure_web_port`  won't be `None` unless explicitly set to `None` after this is approved: https://github.com/galaxyproject/galaxy/pull/8795/commits/e50807a7c6ce050cedab623546f43d1b2b75ac52 

@jmchilton (or @natefoo) can you please take a look at that commit (url + web_port edit): I've described it in the commit message, and I'm not absolutely sure it won't cause any issues (it shouldn't, I think)

jdavcs:
@jmchilton I didn't consider 19.09: I discovered it while working on config stuff, which targets 20.01. But yes, I suppose 19.09 makes sense.

lib/galaxy/visualization/plugins/interactive_environments.py
jdavcs: @nsoranzo I've changed the milestone to 19.09, so yes, I think regardless of the edits I referred to in the other PR, this shouldn't be a problem. It's `None` or the explicitly set `web_port` (or, later, the default `web_port` after that other PR).

nsoranzo:
Changing the milestone is useful only to help us manage PRs and issues, but what you need to do to target release_19.09 is:
- checkout release_19.09 from upstream in a new local branch release_19.09_sg_guess_port:
  ```
  git checkout --no-track -b release_19.09_sg_guess_port upstream/release_19.09
  ```
- Cherry-pick your commit from this PR:
  ```
  git cherry-pick da91150ddd14e9a0e61581631d03cd7ce938a946
  ```
- Push your branch release_19.09_sg_guess_port to your fork on GitHub:
  ```
  git push --set-upstream origin release_19.09_sg_guess_port
  ```
- Open a new pull request from the new branch against galaxyproject/release_19.09

jdavcs:
@nsoranzo  - thank you for the detailed steps - very helpful! Will do, of course (in a meeting right now, so a little later today)

jdavcs:
should i close this one in favor of #8809?

jdavcs:
closed in favor or #8809.

PR #8807, opened 2019-10-15
Implement optional and default_value for workflow input parameters.

jmchilton:
#8232 + transforming workflow state for the tool form client.

### DIFF
--- a/lib/galaxy/workflow/modules.py
+++ b/lib/galaxy/workflow/modules.py
@@ -658,31 +658,41 @@ class InputParameterModule(WorkflowModule):
     name = "Input parameter"
     default_parameter_type = "text"
     default_optional = False
+    default_default_value = ''
     parameter_type = default_parameter_type
     optional = default_optional
+    default_value = default_default_value
 
     def get_inputs(self):
         # TODO: Use an external xml or yaml file to load the parameter definition
         parameter_type = self.state.inputs.get("parameter_type", self.default_parameter_type)
         optional = self.state.inputs.get("optional", self.default_optional)
+        default_value = self.state.inputs.get("default", self.default_value) or ''
         input_parameter_type = SelectToolParameter(None, XML(
             '''
-            <param name="parameter_type" label="Parameter type" type="select">
+            <param name="parameter_type" label="Parameter type" type="select" value="%s">
                 <option value="text">Text</option>
                 <option value="integer">Integer</option>
                 <option value="float">Float</option>
                 <option value="boolean">Boolean (True or False)</option>
                 <option value="color">Color</option>
             </param>
-            '''))
+            ''' % parameter_type))
         for i, option in enumerate(input_parameter_type.static_options):
             option = list(option)
             if option[1] == parameter_type:
                 # item 0 is option description, item 1 is value, item 2 is "selected"
                 option[2] = True
                 input_parameter_type.static_options[i] = tuple(option)
+        input_default_value = TextToolParameter(None, XML(
+            '''
+            <param name="default" label="Default Value" value="%s">
+            </param>
+            '''
+            % default_value))
         return OrderedDict([("parameter_type", input_parameter_type),
-                            ("optional", BooleanToolParameter(None, Element("param", name="optional", label="Optional", type="boolean", value=optional)))])
+                            ("optional", BooleanToolParameter(None, Element("param", name="optional", label="Optional", type="boolean", value=optional))),
+                            ("default", input_default_value)])
 
     def get_runtime_inputs(self, **kwds):
         parameter_type = self.state.inputs.get("parameter_type", self.default_parameter_type)
@@ -691,8 +701,14 @@ class InputParameterModule(WorkflowModule):
             raise ValueError("Invalid parameter type for workflow parameters encountered.")
         parameter_class = parameter_types[parameter_type]
         parameter_kwds = {}
-        if parameter_type in ["integer", "float"]:
-            parameter_kwds["value"] = str(0)
+        if "default" in self.state.inputs:
+            optional = True
+        default_value = self.state.inputs.get("default", self.default_default_value)
+        if default_value:
+            parameter_kwds["value"] = str(default_value)
+        else:
+            if parameter_type in ["integer", "float"]:
+                parameter_kwds["value"] = str(0)
 
         # TODO: Use a dict-based description from YAML tool source
         element = Element("param", name="input", label=self.label, type=parameter_type, optional=str(optional), **parameter_kwds)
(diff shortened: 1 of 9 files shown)


lib/galaxy/workflow/modules.py
```diff
@@ -658,44 +669,121 @@ class InputParameterModule(WorkflowModule):
...
                 input_parameter_type.static_options[i] = tuple(option)
-        return OrderedDict([("parameter_type", input_parameter_type),
-                            ("optional", BooleanToolParameter(None, Element("param", name="optional", label="Optional", type="boolean", value=optional)))])
+
+        parameter_type_cond = Conditional()
+        parameter_type_cond.name = "parameter_definition"
+        parameter_type_cond.test_param = input_parameter_type
+        cases = []
+
+        log.info("optional value is %s " % optional)
```
mvdbeek: Do you want to keep this in ?

lib/galaxy/workflow/modules.py
```diff
@@ -658,44 +669,121 @@ class InputParameterModule(WorkflowModule):
...
+            optional_cond = Conditional()
+            optional_cond.name = "optional"
+            optional_cond.test_param = optional_value
+
+            when_text = ConditionalWhen()
+            when_text.value = param_type
+            when_text.inputs = OrderedDict()
+            when_text.inputs["optional"] = optional_cond
+
+            log.info("parameter_def is %s" % parameter_def)
```
mvdbeek: Same here ?

test/api/test_workflows.py
```diff
@@ -2135,6 +2138,54 @@ def test_run_with_validated_parameter_connection_valid(self):
...
+
+            failed = False
+            try:
+                self._run_jobs(WORKFLOW_PARAMETER_INPUT_INTEGER_OPTIONAL, test_data="""
+data_input:
+  value: 1.bed
+  type: File
+""", history_id=history_id, wait=True, assert_ok=True)
+            except AssertionError as e:
+                print(e)
```
mvdbeek: couldn't we just run the workflow outside of the try/except here ?

test/api/test_workflows.py
```diff
@@ -2135,6 +2138,54 @@ def test_run_with_validated_parameter_connection_valid(self):
...
+    def test_run_with_validated_parameter_connection_default_values(self):
+        with self.dataset_populator.test_history() as history_id:
+            self._run_jobs(WORKFLOW_PARAMETER_INPUT_INTEGER_DEFAULT, test_data="""
+data_input:
+  value: 1.bed
+  type: File
+""", history_id=history_id, wait=True, assert_ok=True)
+            self.dataset_populator.wait_for_history(history_id, assert_ok=True)
+            content = self.dataset_populator.get_history_dataset_content(history_id)
+            assert len(content.split("\n")) == 4, content
```
mvdbeek:
```suggestion
            assert len(content.splitlines()) == 3, content
```
that feels more intuitive

test/api/test_workflows.py
```diff
@@ -2135,6 +2138,54 @@ def test_run_with_validated_parameter_connection_valid(self):
...
+data_input:
+  value: 1.bed
+  type: File
+int_input:
+  value: 1
+  type: raw
+""", history_id=history_id, wait=True, assert_ok=True)
+            self.dataset_populator.wait_for_history(history_id, assert_ok=True)
+            content = self.dataset_populator.get_history_dataset_content(history_id)
+            assert len(content.split("\n")) == 2, content
```
mvdbeek:
```suggestion
            assert len(content.splitlines()) == 1, content
```

test/api/test_workflows.py
```diff
@@ -2135,6 +2138,54 @@ def test_run_with_validated_parameter_connection_valid(self):
             jobs = self._history_jobs(history_id)
             assert len(jobs) == 1
 
+    def test_run_with_int_parameter(self):
+        with self.dataset_populator.test_history() as history_id:
+            failed = False
+            try:
```
mvdbeek: The try/except could also be replaced with
```
with pytest.raises(AssertionError, match='(int_input) has no input'):
```
but it's totally fine the way is now.

lib/galaxy/workflow/modules.py
```diff
@@ -658,44 +669,121 @@ class InputParameterModule(WorkflowModule):
...
+        parameter_type = parameter_def["parameter_type"]
+        optional = parameter_def["optional"]
         if parameter_type not in ["text", "boolean", "integer", "float", "color"]:
             raise ValueError("Invalid parameter type for workflow parameters encountered.")
         parameter_class = parameter_types[parameter_type]
         parameter_kwds = {}
-        if parameter_type in ["integer", "float"]:
+
+        if optional:
+            default_value = str(parameter_def.get("default", self.default_default_value))
```
mvdbeek: We probably shouldn't cast to string here ?

mvdbeek (approved):
I like this a lot, but we should fix the default values that are selected without user intervention (i.e define a required integer parameter in the workflow editor, it shouldn't default to '0' in the workflow run form imo).

I started working on this in https://github.com/mvdbeek/galaxy/commit/04c555d2ef484b8c6454b661d5e47b6ff3a118f7 but I still need to wrap my head around how we forward invalid parameter selections in the regular tool form and why you can only create a non-optional `IntegerToolParameter` with a valid value.

If you can drop the debug statements (assuming they are just for debugging) I'm happy to merge and build on this.

mvdbeek:
> why you can only create a non-optional `IntegerToolParameter` with a valid value.

I guess that's just what we say in the schema: https://docs.galaxyproject.org/en/latest/dev/schema.html#id24
> If false, parameter must have a value. Defaults to "false".

I do feel like that's a weird choice for a mandatory workflow integer parameter though, but maybe I'm wrong.

jmchilton:
I should have taken a pass at cleaning up the debug (prints, unneeded try/excpt, etc..) - sorry @mvdbeek. I'll rebase with a cleaning up of all of this - I'll eliminate that extra ``str`` cast as well and see if that was important.

jmchilton:
I'm well on the record that the tool state interactions are absolutely the most terrifying part of Galaxy in terms of long term technical viability. I'm in no rush to push something through that complicates things and leaves them in a partially implemented state. This can wait for us to implement (or leverage) the concept of "a required value without a default" so we can look at things all together and make judgements.

Frankly - I'm uncomfortable with merging this without implementing optional data inputs at the same time 🤷‍♂.

mvdbeek:
> This can wait for us to implement (or leverage) the concept of "a required value without a default"

We should line out what that would involve and look like. Would this be another attribute on parameters such as `no_default_value="true"`, or would we just do what we do for optional parameters, but raise an exception when attempting to execute the tool if no value has been chosen (and have the tool form highlight that parameter in some way) ?
What you're worried about is having a tool state (let's say an integer parameter) in the database that doesn't have a value, and it'd be difficult to know if that's legitimately without value ?

jmchilton:
> Would this be another attribute on parameters such as no_default_value="true", or would we just do what we do for optional parameters, but raise an exception when attempting to execute the tool if no value has been chosen (and have the tool form highlight that parameter in some way) ?

I don't know - either seems fine to me. Whatever is cleanest?

> What you're worried about is having a tool state (let's say an integer parameter) in the database that doesn't have a value, and it'd be difficult to know if that's legitimately without value ?

I'm just worried about more crazy paths through basic.py frankly. This state thing though seems to be a legitimate concern - especially for executed tools - but hopefully we'd catch this problem before getting that far.

martenson:
this is now in conflict, please resolve it

jmchilton:
Closing in favor of #8807. Applying the workflow module interface changes to a second application clears up I think what I stumbled into here...

PR #8810, opened 2019-10-15
Add installation monitor popup to toolshed client

guerler:
This PR adds a component to the installation client which displays a list of currently installing repositories. The list becomes visible when clicking the toggle link. The goal of this PR is to replace the currently existing "Monitor Installation" mako-based view in the admin tool panel.

Screen Shot 2019-10-22 at 1 35 07 PM


guerler:
@galaxybot test this

martenson:
I think it'd be beneficial to hide this interface behind a toggle button (maybe bottom right under the search box) since it can take a lot of prime space for result browsing and you don't always care about the state of installations. Other than that I like the implementation. I wasn't able to test uninstallation because every attempt yields https://gist.github.com/martenson/67945138bbdb131baab9a181a1ccce11

mvdbeek:
@martenson that's fixed in https://github.com/galaxyproject/galaxy/pull/8835

martenson:
Thanks @guerler, nice implementation.

PR #8812, opened 2019-10-16
[19.09] Fix workflow scheduler conf default path.

jmchilton:
Pointed out by @ic4f


nsoranzo (approved):

nsoranzo:
Broken in #921, I missed it in https://github.com/galaxyproject/galaxy/pull/8494

PR #8813, opened 2019-10-16
client-side current user store module

Nerdinacan:
Adds a store module for the current user. It contains a temporary fallback to the galaxy instance, an ajax load as the store is initialized, and a public facing User class which we can expand on.

My expectation is that this will grow as we slowly cut out usage of the getGalaxyInstance() singleton and replace it with standard SPA token-based authentication architecture.


dannon:
@Nerdinacan Thanks for pulling this bit out!  I'll check it out.

Nerdinacan:
My pleasure. There's not much to it yet, but I use it in one place in History. Figured you could build on it.

dannon (approved):

Issue #8243, opened 2019-06-27
Repos that were updated without a version bump are not installable to the new revision


hexylena:
Opening this to start some discussion maybe over performance during installation of many tools (especially many which are already installed.)

Currently usegalaxy.eu's jenkins job which (weekly) ensures that everything in https://github.com/usegalaxy-eu/usegalaxy-eu-tools is installed to our server, takes around [3-5 hours](https://build.galaxyproject.eu/job/usegalaxy-eu/job/install-tools/buildTimeTrend) on average (less if it crashes due to galaxyproject/ephemeris#133, more if it finishes completely.)

What can we do to accelerate this?

- Could we fetch from galaxy a list of everything already installed + revisions? Surely that endpoint could be relatively quick since it'd be:
  ```
  select tool_shed, name, owner, changeset_revision from tool_shed_repository;
  ```

  And diff that locally before installing?
- Does anyone have any nice suggestions for diffing YAML files? Maybe I could find some ugly hack to find out which was the last built commit, and then diff the yamls and find tools + IDs, and just process those?

cc @bgruening

hexylena:
@mvdbeek please change the title as you see fit. 

Good to know it's another bug due to the fun that is tool updates.

rhpvorderman:
> I don't think it's ephemeris job to figure out the installable vs updatable revisions (you could do that by checking for gaps in the installable revisions from the toolshed, then check whether any installed revision falls in the gap preceeding the requested revision). But that would really fix the wrong end of the problem IMO.

But if the galaxy-api is fixed it takes some time to get in a stable version. And this problem is annoying for the EU galaxy time NOW. 
So a hack in ephemeris is not that bad. Ephemeris can release a new version tomorrow so to speak. We can add a `# FIXME` comment so we can the delete the hackish code later if it gets in the way.

I do agree that it is not the proper way to fix this in ephemeris, but these EU galaxy people are quite essential to the galaxy effort, so it is nice not to let them wait on another Galaxy release and let them spent their time on more useful and fun stuff.

EDIT: TLDR; I am volunteering. Starting pycharm now...

rhpvorderman:
Oh, and before I disappoint I can only work on this for a few moments today. But I will try to fix this on Monday.

rhpvorderman:
[...]

PR #8815, opened 2019-10-16
[19.09] Fix non-version (minor) tool shed repository updates when installing revisions through the API

### LINKED ISSUE #6698: problems with installing tool updates
With ephemeris (and ansible-galaxy-tools which is using ephemeris) we have noticed problems when updating tools: https://github.com/galaxyproject/ephemeris/issues/111

examples are tools with
- new revision but same tool version number
- tool suites (e.g. openms_suite)

I checked the later via the Galaxy UI  on a "recent" dev instance and got `Repository has already been installed` for installing the openms_suite when outdated versions of the suite are installed already.

natefoo:
~~Incomplete but I wanted to run tests to see what breaks.~~

You still can't install a specific revision if that revision is not the newest "installable" revision (e.g. the ones marked "Repository metadata is associated with this change set." [here](https://toolshed.g2.bx.psu.edu/repository/view_changelog?id=9ff2d127cd7ed6bc)) for a particular tool version, but at least you can do an update like you can from the UI.

There's code that is supposed to update any repo dependencies if new repo deps are added in an update (which I personally think you should not do, but it is currently supported/allowed), this is refactored from the controller method but ~~is incomplete and completely untested~~ didn't work before anyway?? There is also code that is supposed to update any tool dependencies if new tool deps are added in an update, I made no attempt to support these via the API, and do not think doing so is worth the effort.

This could also be targeted against dev as I am a little nervous about all the potential install/upgrade scenarios this could break. Practically every permutation of repo state/deps/contents/etc. has its own code path.

Fixes #6698 #8243 galaxyproject/ephemeris#111 galaxyproject/ephemeris#140

### DIFF
--- a/lib/galaxy/webapps/galaxy/controllers/admin_toolshed.py
+++ b/lib/galaxy/webapps/galaxy/controllers/admin_toolshed.py
@@ -18,10 +18,7 @@ from tool_shed.galaxy_install.grids import admin_toolshed_grids
 from tool_shed.galaxy_install.installed_repository_manager import InstalledRepositoryManager
 from tool_shed.galaxy_install.metadata.installed_repository_metadata_manager import InstalledRepositoryMetadataManager
 from tool_shed.galaxy_install.repository_dependencies import repository_dependency_manager
-from tool_shed.galaxy_install.tools import (
-    data_manager,
-    tool_panel_manager
-)
+from tool_shed.galaxy_install.tools import tool_panel_manager
 from tool_shed.util import (
     common_util,
     encoding_util,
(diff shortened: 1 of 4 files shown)


mvdbeek:
Looks promising, the first 2 errors probably occurred because an unaccounted for update happened ?

natefoo:
So I tried manually testing whether the repository-dependency-update-in-a-non-version-TSR-update stuff works, and it didn't... but I also tried it in plain 19.09 and it doesn't appear to work there either. I would imagine this is how most suite updates would occur? I never use suites so I'm not sure.

mvdbeek:
Yeah, I don't think suite updates have ever worked ? Maybe if you updated one and then went to repair it ? Not a big deal IMO.

mvdbeek (approved):
Fantastic, thanks for working this out @natefoo!

natefoo:
Thanks @mvdbeek!

martenson:
Marvelous @natefoo 🤩

bgruening:
Is that working for anyone? Running this with our automatic ephemeris based installation it creates __a lot__ of `new` and `error` repositories. It seems it turns old repos into a new state.

grafik


@natefoo are you running this against .org?

natefoo:
Yes, I used it to (I think) update tools on Main. Was this addressed in #8959?

PR #8816, opened 2019-10-17
add option select_type_workflow_threshold

### LINKED ISSUE #8200: Select lists are less convenient when we run a workflow
There is a difference in the way the lists are displayed when we use a tool or we we use the same tool in a workflow.

Example with BlastN:
<a href="https://imgbb.com/">2019-06-20-Screen-Hunter-04</a>

I can start to enter `UNI` in order to shrink the list

But if BlastN is used in a workflow, when I run the worflow I can't do the same:
<a href="https://imgbb.com/">2019-06-20-Screen-Hunter-05</a>

And we can't select all, or have a nice boxe when a genome is selected:
<a href="https://imgbb.com/">2019-06-20-Screen-Hunter-06</a>

Is it possible to display lists the same way ? Is it simple to do that (javascript / bootstrap issue) ?

Because I'm back from two days of training of our users, and this behavior will be an issue for me.

Our reference genome list is quite long (and will be possibly longer...). And our users complain about that.
It is somewhat related to #8145 
I tried to add a filter, but this is not a great solution (in my admin point of view). I don't want to modify all XML that deal with a .loc file!

I'm still using 19_01, I will try to upgrade to 19_05 next week.

FredericBGA:
It should resolve #8200 
thank you @dannon a,d @ic4f  for the help provided!

### DIFF
--- a/client/galaxy/scripts/mvc/form/form-parameters.js
+++ b/client/galaxy/scripts/mvc/form/form-parameters.js
@@ -89,6 +89,21 @@ export default Backbone.Model.extend({
             radiobutton: Ui.RadioButton
         };
         var SelectClass = classes[input_def.display] || Ui.Select;
+        // use Select2 fields or regular select fields in workflow launch form?
+        // check select_type_workflow_threshold option
+        const Galaxy = getGalaxyInstance();
+        var searchable = true;
+        if ( input_def.flavor == "workflow") {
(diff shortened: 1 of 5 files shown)


dannon:
@galaxybot test this

jmchilton:
This was conflicted and touching unrelated files - I'm going to close it in lieu of #9105

PR #8792, opened 2019-10-11
[19.09] Fix 'NoneType' object has no attribute 'get_tool_relative_path' with data_manager_manual

abretaud:
While running biomaj2galaxy tests with Galaxy 19.05 I get this error when trying to populate a data table with the [manual data manager](https://github.com/galaxyproject/tools-iuc/tree/master/data_managers/data_manager_manual):

```
Traceback (most recent call last):
  File "lib/galaxy/jobs/runners/__init__.py", line 224, in prepare_job
    job_wrapper.prepare()
  File "lib/galaxy/jobs/__init__.py", line 863, in prepare
    tool_evaluator.set_compute_environment(compute_environment, get_special=get_special)
  File "lib/galaxy/tools/evaluation.py", line 118, in set_compute_environment
    out_data=out_data, tool=self.tool, param_dict=incoming)
  File "lib/galaxy/tools/__init__.py", line 1719, in call_hook
    return code(*args, **kwargs)
  File "../shed_tools/toolshed.g2.bx.psu.edu/repos/iuc/data_manager_manual/6524e573d9c2/data_manager_manual/data_manager/data_manager_manual.py", line 57, in exec_before_job
    target_dir, tool_path, relative_target_dir = tdtm.get_target_install_dir( tool_shed_repository )
  File "lib/tool_shed/tools/data_table_manager.py", line 111, in get_target_install_dir
    tool_path, relative_target_dir = tool_shed_repository.get_tool_relative_path(self.app)
AttributeError: Error in 'Manual Data Manager' hook 'exec_before_job', original message: 'NoneType' object has no attribute 'get_tool_relative_path'
```

I guess the bug was introduced in #7316, a tool_shed_repository object was no longer accessible from the data manager code.

I tested the patch with 19.05, but it should work the same with 19.09 (too late to test this week :) )

It would be cool to have it backported to 19.05 too (I have a branch ready for that if needed https://github.com/galaxyproject/galaxy/compare/release_19.05...abretaud:dmfix1905?expand=1)

(There's another PR in IUC to fix another problem in the data manager https://github.com/galaxyproject/tools-iuc/pull/2634)

PR #8817, opened 2019-10-17
[19.09] Fix data_manager_manual with new data tables + cleanup and fixes

mvdbeek:
Builds on #8792 by @abretaud 

- Fixes and unifies DM loading
- More useful traceback when loading DM failed
- Implement get_tool_relative_path, needed for running data_manager_manual without install db
- Test data_manager_manual in an integration test

### DIFF
--- a/lib/galaxy/model/tool_shed_install/__init__.py
+++ b/lib/galaxy/model/tool_shed_install/__init__.py
@@ -42,7 +42,7 @@ class ToolShedRepository(object):
         self.installed_changeset_revision = installed_changeset_revision
         self.changeset_revision = changeset_revision
         self.ctx_rev = ctx_rev
-        self.metadata = metadata
+        self.metadata = metadata or {}
         self.includes_datatypes = includes_datatypes
         self.tool_shed_status = tool_shed_status
         self.deleted = deleted
@@ -81,10 +81,7 @@ class ToolShedRepository(object):
 
     @property
     def shed_config_filename(self):
-        shed_config_filename = None
-        if self.metadata:
-            shed_config_filename = self.metadata.get('shed_config_filename', shed_config_filename)
-        return shed_config_filename
+        return self.metadata.get('shed_config_filename', None)
 
     @shed_config_filename.setter
     def shed_config_filename(self, value):
@@ -122,7 +119,7 @@ class ToolShedRepository(object):
 
     def guess_shed_config(self, app, default=None):
         tool_ids = []
-        metadata = self.metadata or {}
+        metadata = self.metadata
         for tool in metadata.get('tools', []):
             tool_ids.append(tool.get('guid'))
         for shed_tool_conf_dict in app.toolbox.dynamic_confs(include_migrated_tool_conf=True):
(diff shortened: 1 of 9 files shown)


mvdbeek:
I think this is ready for review now.

lib/galaxy/model/tool_shed_install/__init__.py
```diff
@@ -79,12 +79,17 @@ def can_reinstall_or_activate(self):
...
+    @property
+    def shed_config_filename(self):
         shed_config_filename = None
         if self.metadata:
             shed_config_filename = self.metadata.get('shed_config_filename', shed_config_filename)
         return shed_config_filename
 
+    @shed_config_filename.setter
+    def shed_config_filename(self, value):
+        self.metadata['shed_config_filename'] = value
```
nsoranzo, line 88: Should we make sure that `self.metadata` is not `None` by using `self.metadata = metadata or {}` in the class `__init__()` method (line 45)?

Then so many methods can be simplified.

lib/galaxy/tools/data_manager/manager.py
```diff
@@ -75,8 +71,8 @@ def load_manager_from_elem(self, data_manager_elem, tool_path=None, add_manager=
             if e.errno == errno.ENOENT:
                 # File does not exist
                 return None
-        except Exception as e:
-            log.error("Error loading data_manager '%s':\n%s" % (e, util.xml_to_string(data_manager_elem)))
+        except Exception:
+            log.error("Error loading data_manager", exc_info=True)
```
nsoranzo:
```suggestion
            log.exception("Error loading data_manager")
```

lib/galaxy/tools/data_manager/manager.py
```diff
@@ -143,65 +139,31 @@ def load_from_element(self, elem, tool_path):
...
         self.version = elem.get('version', self.version)
-        tool_shed_repository_id = None
+        tool_shed_repository = None
         tool_guid = None
 
         if path is None:
             tool_elem = elem.find('tool')
             assert tool_elem is not None, "Error loading tool for data manager. Make sure that a tool_file attribute or a tool tag set has been defined:\n%s" % (util.xml_to_string(elem))
-            path = tool_elem.get("file", None)
+            path = tool_elem.get("file", path)
```
nsoranzo:
```suggestion
            path = tool_elem.get("file")
```

lib/galaxy/model/tool_shed_install/__init__.py
mvdbeek: I think that should work, I'll give it a try

lib/galaxy/tools/data_manager/manager.py
```diff
@@ -143,65 +139,31 @@ def load_from_element(self, elem, tool_path):
...
-                    shed_conf = self.data_managers.app.toolbox.get_shed_config_dict_by_filename(shed_conf_file, None)
-                    if shed_conf:
-                        tool_path = shed_conf.get("tool_path", tool_path)
+            tool_shed_repository = self.data_managers.app.toolbox.get_tool_repository_from_xml_item(tool_elem, path)
+            self.tool_shed_repository_info_dict = dict(tool_shed=tool_shed_repository.tool_shed,
+                                                       name=tool_shed_repository.name,
+                                                       owner=tool_shed_repository.owner,
+                                                       installed_changeset_revision=tool_shed_repository.installed_changeset_revision)
+            # use shed_conf_file to determine tool_path
+            shed_conf_file = elem.get("shed_conf_file", None)
```
nsoranzo:
```suggestion
            shed_conf_file = elem.get("shed_conf_file")
```

bgruening:
@mvdbeek here are now some conflicts.

jmchilton (approved):

lib/galaxy/tools/__init__.py
```diff
@@ -520,7 +520,6 @@ def tool_shed_repository(self):
                                                             name=self.repository_name,
                                                             owner=self.repository_owner,
                                                             installed_changeset_revision=self.installed_changeset_revision,
-                                                            repository_id=self.repository_id,
```
nsoranzo, line 523: Not sure I understand this change.

mvdbeek: If you specify a repository_id we'll not be able to use ToolConfRepository items, which don't have an id. We should only provide the repository_id when working in the galaxy_install module.

PR #8791, opened 2019-10-11
Revert "[19.09] Fix detached parent instance error in 8769"

bgruening:
I think we need to revert this one. It increases the memory usage of handler 4 fold.

grafik

handler3 is running the reverted Galaxy codebase. Handler 0,1,2 are running this patch. As you can see the memory consumption is 4 times higher :(

Sorry!

PR #8818, opened 2019-10-17
[19.09] Don't use a thread local ToolShedRepositoryCache, this may have been …

mvdbeek:
…a memory leak

Every thread would need to build the cache again when first accessing it. Now that we eagerload the relationships (https://github.com/galaxyproject/galaxy/pull/8791 only went into dev) we shouldn't need to keep the cache thread local. This should be faster and more memory efficient, and actually it looks like the thread-local cache didn't work as intended before (writing tests now to verify it does work now).

### DIFF
--- a/lib/galaxy/tools/cache.py
+++ b/lib/galaxy/tools/cache.py
@@ -137,6 +137,7 @@ class ToolShedRepositoryCache(object):
         # Repositories loaded from database
         self.repositories = []
         self.repos_by_tuple = defaultdict(list)
+        self.rebuild()
 
     def add_local_repository(self, repository):
         self.local_repositories.append(repository)
--- a/lib/galaxy/tools/cache.py
+++ b/lib/galaxy/tools/cache.py
@@ -1,13 +1,7 @@
 import logging
 import os
 from collections import defaultdict
-from threading import (
-    local,
-    Lock,
-)
-
-from sqlalchemy import inspect
-from sqlalchemy.orm.exc import DetachedInstanceError
+from threading import Lock
 
 from galaxy.util import unicodify
 from galaxy.util.hash_util import md5_hash_file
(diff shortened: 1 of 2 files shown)


lib/galaxy/tools/cache.py
```diff
@@ -138,59 +132,30 @@ class ToolShedRepositoryCache(object):
...
-            repositories = self.cache.repositories
-        tool_shed_repositories = [repo for repo in repositories if isinstance(repo, self.app.install_model.ToolShedRepository)]
-        if tool_shed_repositories and inspect(tool_shed_repositories[0]).detached:
-            self.rebuild()
-            repositories = self.cache.repositories
-        return repositories
+        return self.repositories
 
     @property
     def tool_shed_repos_by_tuple(self):
```
nsoranzo: Since this property is only used once, I'd just drop it and replace it with `self.repos_by_tuple` inside `get_installed_repository()` below.

mvdbeek: Thanks! Added that in the rebase

lib/galaxy/tools/cache.py
```diff
@@ -138,66 +132,29 @@ class ToolShedRepositoryCache(object):
 
     def __init__(self, app):
         self.app = app
-        self.cache = local()
+        self.local_repositories = []
```
nsoranzo, line 136: Can you add a comment about `self.local_repositories` clarifying it stores tools that are not in the install database?

mvdbeek: Yup, done.

test/unit/tools/test_tool_shed_repository_cache.py
```diff
@@ -0,0 +1,117 @@
...
+    ('github.com', 'example', 'galaxyproject', '2', '1', None, True),
+    ('github.com', 'example', 'galaxyproject', '500', '1', None, False),
+    ('github.com', 'example', 'galaxyproject', '1', '500', None, False),
+    ('github.com', 'example', 'galaxyproject', '2', '1', 1, True),
+    ('github.com', 'example', 'galaxyproject', '2', '1', 500, False),
+    ('github.com', 'example', 'galaxyproject', '19', '18', None, True),
+])
+def test_get_installed_repository(tool_shed_repository_cache, repos, tool_shed, tool_conf_repos, name, owner, changeset_revision, installed_changeset_revision, repository_id, repo_exists):
+    tool_shed_repository_cache.rebuild()
+    print(tool_shed_repository_cache.local_repositories)
```
nsoranzo:
```suggestion
```

test/unit/tools/test_tool_shed_repository_cache.py
```diff
@@ -0,0 +1,117 @@
...
+    ('github.com', 'example', 'galaxyproject', '1', None, None, True),
+    ('github.com', 'example', 'galaxyproject', None, '1', None, True),
+    ('github.com', 'example', 'galaxyproject', '2', '1', None, True),
+    ('github.com', 'example', 'galaxyproject', '500', '1', None, False),
+    ('github.com', 'example', 'galaxyproject', '1', '500', None, False),
+    ('github.com', 'example', 'galaxyproject', '2', '1', 1, True),
+    ('github.com', 'example', 'galaxyproject', '2', '1', 500, False),
+    ('github.com', 'example', 'galaxyproject', '19', '18', None, True),
+])
+def test_get_installed_repository(tool_shed_repository_cache, repos, tool_shed, tool_conf_repos, name, owner, changeset_revision, installed_changeset_revision, repository_id, repo_exists):
```
nsoranzo:
```suggestion
def test_get_installed_repository(tool_shed_repository_cache, repos, tool_conf_repos, tool_shed, name, owner, changeset_revision, installed_changeset_revision, repository_id, repo_exists):
```

nsoranzo (approved):

nsoranzo:
Thanks @mvdbeek ! Merged forward.

martenson:
This is looking good so far on both .eu and .org. Thank you @mvdbeek for working on this.

guerler:
Yep, thanks a lot for working on this @mvdbeek!

PR #8823, opened 2019-10-18
LDAP fixes

nsoranzo:
- Partial revert of #8618 , see https://github.com/galaxyproject/galaxy/pull/8618#discussion_r323488794
- Python3 fixes


mvdbeek (approved):
Should we open a bug report in python-ldap, or is there a legitimate reason for returning byte strings here ?

PR #8827, opened 2019-10-18
Added client-side application configs for each target environment

Nerdinacan:
This creates 3 files on the client that allow us to configure the application based on the target environment. I'm going to need this shortly for the indexDB driver setup, but it can be used for any application-specific configs, so I figured I'd PR this separately.

Webpack allows you to configure with a function instead of a flat object. Its signature is

```js
module.exports = (env, argv) => {
   ...build your webpack
   return { big object }
}
```

This conveniently lets us set up aliases based on the target environment, which is traditionally set in webpack by passing a -p or -d param on the CLI build command.

The end result is now you can do this:

```js
import config from "config"
```
And it will import values from this folder:
Screen Shot 2019-10-18 at 12 47 35 PM

...from the file whose name matches the environment.

Currently those configs are empty except for a name. I imported it in one file (scripts/onload/index.js) just to prove that it works. I'll be using those objects to store configs for a PouchDB driver (indexDB for development/production, in-memory for testing).


dannon:
@Nerdinacan Thanks for incorporating those changes and resolving the conflict; I'll check this out shortly.

dannon (approved):

PR #8829, opened 2019-10-18
Adds hdca timestamps and update trigger (for real this time)

Nerdinacan:
I made this a while ago but now I'm ready to actually use them. This adds standard timestamps to the hdca table and update triggers to propagate changes from datasets and collections up to the history.

I think I let this thing rattle along for so long that the original PR got deleted. Sorry about that!

### DIFF
--- a/lib/galaxy/model/migrate/versions/0160_add_hdac_update_time.py
+++ b/lib/galaxy/model/migrate/versions/0160_add_hdac_update_time.py
@@ -1,0 +1,36 @@
+"""
+Adds create_time and update_time to hdac.
+"""
+import datetime
+import logging
+
+from sqlalchemy import Column, DateTime, MetaData, Table
+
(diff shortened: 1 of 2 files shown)


nsoranzo:
Thanks @Nerdinacan! 3 comments:
- You need also to update `lib/galaxy/model/mapping.py` accordingly since in latest Galaxy releases the database is initially created from that file instead of going through each migration.
- I'd merge the 2 migrations into 1
- The original PR was not deleted, it's still open at https://github.com/galaxyproject/galaxy/pull/8187

nsoranzo:
Ah, forgot to mention that the API test failures are relevant, the triggers seem to create deadlocks on PostgreSQL.

Nerdinacan:
Weird! I looked for the original PR before I did this and couldn't find it. I'll close this one, implement your changes and update the original. Thanks @nsoranzo

PR #7402, opened 2019-02-22
Activate api-based Toolshed interface in Admin panel

guerler:
This PR activates the new api-based Toolshed installer in the Galaxy's admin panel. This is the first step to properly branch the Toolshed and Galaxy repositories, so that they can maintained individually.

PR #8830, opened 2019-10-18
Remove Toolshed Server from Galaxy codebase

guerler:
This PR separates Galaxy and the Toolshed Server code base into individual git repositories. The Toolshed Server will be maintained at https://github.com/galaxyproject/toolshed.


jmchilton:
I would feel much more comfortable with this if we weren't essentially creating two sources of truths for the ``galaxy`` namespace and for the ``toolshed`` namespace in Python. If one project could import the other from PyPI I'd feel much better - and this doesn't look like it is there yet at all.

What is the pain point you're trying to address here - nothing in ``lib/`` right? Could you just clone the ``templates/`` directory in repo to get the same effective outcome?

martenson:
> Could you just clone the templates/ directory in repo to get the same effective outcome?

@jmchilton what do you mean by this?

jmchilton:
@martenson ...

Things are presumably being separated because they need to be changed and it is painful to do when both apps are pointing at the same code, they need to evolve independently I suppose. Is just forking the templates directory enough or do we really need to fork the actual code as well?

I'm not opposed to forking the codebases long term - I just wish one could import the other so we don't have two different implementations of ``galaxy`` and ``toolshed`` packages in two different places.

jmchilton:
Like if we could refactor this subset of ``toolshed`` (https://github.com/guerler/galaxy/tree/remove_admin_toolshed_controller/lib/tool_shed) - the set Galaxy needs, into ``galaxy.toolshed`` or something like that and include it in the ``galaxy-app`` package (https://github.com/galaxyproject/galaxy/tree/dev/packages/app) and then have toolshed import it from PyPI and not include ``galaxy`` in its codebase... we'd have just one source of truth for every package.

jmchilton:
We don't even have to aggressively refactor... we could maybe just do...

```python
import * from galaxy.toolshed.x
```

for each ``toolshed.x`` module in that tree. Then all the tool shed references in templates and tool shed controller code wouldn't need to be updated.

guerler:
@jmchilton not sure if I understand "Then all the tool shed references in templates and tool shed controller code wouldn't need to be updated.". This PR removes all toolshed templates and controllers from the Galaxy repository.

jmchilton:
I'm responding to what is in both repositories simultaneously - I'm not ready to just duplicate all this code in this fashion. So over in the other repo those controllers are still existent and still doing imports.

guerler:
With other repo you mean the `galaxy/toolshed` repo? The current code overlap is in the `metadata_generator.py` and a subset of utilities. It would be fine to import the utilities and share them with the toolshed server repository. Also the same process we do here for Galaxy has to be done for the Toolshed repository in the future.

jmchilton:
Yes.

jmchilton:
I mean like this... https://github.com/galaxyproject/galaxy/pull/8905

The goal being cleanly removing the entirety of ``lib/tool_shed`` from the Galaxy repo since it would no longer have any dependency on that stuff. And then importing ``lib/galaxy`` via PyPI from the toolshed repo and removing ``lib/galaxy`` entirely and cleanly from there.

There would be one source of truth for all Python files then.

dannon:
Do you think there are significant concerns about just 'freezing' the toolshed as-is and leaving it alone?  Other than some recent-ish python3 fixes, I am not recalling many instances where we'd actually need to propagate changes to the multiple sources of truth.

Since it works as-is right now, I thought the plan was just to completely freeze the current state of stuff in a parallel branch and just leave it alone.

(to be clear -- I don't think that refactoring as proposed is a bad idea, I just don't see it as a necessary prerequisite when we're essentially freezing/abandoning the toolshed as-is)

guerler:
Freezing the state of the Toolshed server as is makes sense.

jmchilton:
> Do you think there are significant concerns about just 'freezing' the toolshed as-is and leaving it alone? 

If I was sure we were done with it... I'm not. It is a still a critical piece of infrastructure and we still update it.

martenson:
>  completely freeze the current state of stuff

I don't understand how we can freeze a critical piece of our infrastructure without alternative. If anything we should make it as easy to maintain as possible which John's comments would help with.

dannon:
> I don't understand how we can freeze a critical piece of our infrastructure without alternative.

I'm not sure what you mean by this -- 'without alternative'?  We're not deleting the toolshed, there's no need for an alternative -- it'd serve everyone best by continuing exactly as it is, without having to worry about changes from the core Galaxy.

I thought we decided years ago that it was maintenance mode, and my two cents say that since then the frustration in keeping the client functional within the common codebase have probably outweighed the benefits we get from having a single source of truth.  

That is, updates affecting the toolshed codebase are few and far between, and we're the only people that (I think) we want running one anyway, these days.  We're not talking about broad impact in terms of moving the toolshed to a separate frozen branch.

martenson:
to rephrase: Every Galaxy needs to install tools from a toolshed to be functional. There is no alternative. That makes toolshed a critical piece to maintain. 

Maybe I just don't understand what you mean by freezing.

guerler:
@martenson previous galaxy instances use toolshed makos to install and manage toolshed repositories. Although not visible in the interface of 20.01, the makos and controllers still exist. So far we had to keep those galaxy controllers in our repository because they are used by the toolshed and by legacy galaxy instances. The state of these controllers will be frozen in the toolshed server repository since our current galaxy dev does not use these controllers anymore.

dannon:
@martenson Ok, sure, I don't think anyone disagrees with that -- the toolshed needs to keep working exactly as it is right now.  I think we're just trying to figure out the balance of the benefit of getting all the toolshed code out of the core repo ASAP compared to the concerns that (fairly rare, at this point) framework changes will need to propagate to both repositories.

guerler:
We spent a lot of time to untangle the Toolshed server code from Galaxy e.g. in the past Galaxy's login mechanism was imported by Toolshed controllers which made it difficult to replace the authentication makos, same goes for grids and panel makos. There have been quite a few similar instances of time consuming code conflicts in the past. With this step here the remaining overlap for Galaxy is reduced to a subset of utility functions (from lib/tool_shed) and it will be reduced further. However I am open to suggestions to make this transition easier/better.

guerler:
@jmchilton are you saying that we should move `lib/toolshed` to `lib/galaxy/toolshed` to avoid a namespace overlap?

jdavcs:
(if the following is not relevant, please ignore)

With regard to untangling. `galaxy/webapps/tool_shed/config.py` depends on `galaxy/config/` in multiple ways. `ToolShedAppConfiguration` extends `BaseAppConfiguration` and uses much of its functionality. 

If we are freezing TS code, I suppose, this dependency needs to be broken up as well. Since TS, as I understand, should be maintenance only, one way to do this would be to duplicate `BaseAppConfiguration` in its current state and place it in `tool_shed/config.py`. That would also make any further modifications to Galaxy's configuration more straightforward and much safer. (just like it has been done for `get_database_engine_options()`.

guerler:
@ic4f this PR removes the Toolshed server from the Galaxy repository, not the other way around. The untouched Toolshed server code (including Galaxy modules used by the Toolshed server) will be maintained separately at`galaxyproject/toolshed`. The goal of this PR is to allow us to develop and maintain Galaxy without having to maintain the Toolshed server, legacy controllers and makos at the same time.

jdavcs:
@guerler so once this is merged, the files under webapps/tool_shed (and others, of course) will be removed from the galaxy codebase?

guerler:
Yes, and maintained at `galaxyproject/toolshed`.

jdavcs:
then it's a non-issue in this case. Thanks.

dannon:
@ic4f Yep, you nailed it right here:  (though applying to a lot more than just the configuration)
> That would also make any further modifications to Galaxy's configuration more straightforward and much safer.

jmchilton (changes requested):
I think it would be wiser to do this incrementally and preserve the tests throughout and to not fork overlapping code. I think the proper first step is to separate the templates - this seems to be the major pain point in recent months - a branch that does that is here https://github.com/jmchilton/galaxy/tree/sep_templates.

guerler:
We have done this incrementally (#8734, #8601, #7402, #7047, #8327), the code overlap for Galaxy is minimal and restricted to `lib/toolshed`, also all Toolshed Server tests are preserved in the forked Toolshed Server repository. Are you referring to a loss of coverage for actual Galaxy modules in the Galxy repository which were previously tested by the Toolshed Server code?

guerler:
Superseeded by #8905.

PR #8835, opened 2019-10-21
[19.09] Only use cached repositories where needed & safe

mvdbeek:
We only need them when loading tools and data managers, for repository operations such as install/uninstall/modify we should get repos directly from the database.

Fixes ` InvalidRequestError: Object '<ToolShedRepository at 0x12c086ed0>' is already attached to session '60' (this is '36')` when uninstalling repositories. discovered by @guerler


guerler:
Thank you for the quick fix!

PR #8837, opened 2019-10-22
Fix rare bug in config loading + add unit test

jdavcs:
The value of a path segment may be None if the config option is set to None in kwargs.
This may be a rare scenario, but it may happen, and it does happen when
running integration tests: job_config_file is overridden with `None`.

PR includes fix + new unit test that exposes the bug.

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -283,7 +283,8 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
         for key, value in kwargs.items():
             if key in self.appschema:
                 value = convert_datatype(key, value)
-                value = strip_deprecated_dir(key, value)
+                if value:
+                    value = strip_deprecated_dir(key, value)
                 self._raw_config[key] = value
 
     def _create_attributes_from_raw_config(self):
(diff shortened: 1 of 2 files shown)


lib/galaxy/config/__init__.py
```diff
@@ -283,7 +283,8 @@ def strip_deprecated_dir(key, value):
         for key, value in kwargs.items():
             if key in self.appschema:
                 value = convert_datatype(key, value)
-                value = strip_deprecated_dir(key, value)
+                if value:
```
mvdbeek, line 286:
```suggestion
                if value is not None:
```
Based on the description that seems more correct to me ?

jdavcs: It should be `if value` (or `if value is not None`) . Sorry - my description does sound confusing: I meant that it *may* be None, in which case the code breaks, so this is a fix (because the code will try to split a None value). 
~~That said, while looking into this I found a case that would break things: value may be not a string (if set by mistake).~~ [EDIT: checked, not possible, so all good.] .  I'll test for more edge cases and will submit a fix later today.

jdavcs: I checked. In this case it cannot be a wrong datatype: so no additional bug. 
May I mark as resolved?

PR #8839, opened 2019-10-22
First update defaults with explicit kwds, then update with env overrides

mvdbeek:
Fixes https://github.com/peterjc/galaxy_blast/issues/120 and https://github.com/galaxyproject/planemo/issues/969

PR #8841, opened 2019-10-22
[19.09] First update defaults with explicit kwds, then update with env overrides

mvdbeek:
Backport of #8839 -- was targeting dev when I wanted 19.09

Fixes https://github.com/peterjc/galaxy_blast/issues/120 and https://github.com/galaxyproject/planemo/issues/969


jdavcs (commented):
Yep, agreed.

PR #8843, opened 2019-10-22
General server dependency updates (was/including catalina fix)

dannon:
This started as an update to cryptography (the first commit here, which does fix startup on Catalina), but I decided that this early in the cycle would be a good time to do a full-pass update as well.  The first commit has been backported to 19.05 in #8845


lib/galaxy/dependencies/pipfiles/default/Pipfile
```diff
@@ -70,7 +70,7 @@ bioblend = "*"
 boto = "*"
 kombu = "*"
 psutil = "*"
-pulsar-galaxy-lib = ">=0.14.0.dev1"
+pulsar-galaxy-lib = "==0.14.0.dev1"
```
nsoranzo, line 71: This is necessary because of a pipenv issue: https://github.com/pypa/pipenv/issues/3928

nsoranzo:
I just noticed that https://github.com/galaxyproject/galaxy/pull/8543 added 2 new dependencies to the Pipfile without running `make update-dependencies`: https://github.com/galaxyproject/galaxy/pull/8543/files#r338021231

This is now adding 9 new dependencies which I think may be not used: markdown, weasyprint, CairoSVG, Pyphen, cairocffi, cssselect2, html5lib, pdfrw and tinycss2.

nsoranzo (approved):

PR #8845, opened 2019-10-22
[19.05] Backported cryptography update;

dannon:
This turns out to be the culprit causing python startups on catalina to fail (though I don't know exactly why -- has to do with openssl ).

xref #8843

### DIFF
--- a/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
+++ b/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
@@ -42,7 +42,7 @@ cloudauthz==0.2.0
 cloudbridge==2.0.0
 cmd2==0.8.9
 contextlib2==0.5.5 ; python_version < '3.5'
-cryptography==2.6.1
+cryptography==2.8
 cwltool==1.0.20180721142728
 debtcollector==1.21.0
 decorator==4.4.0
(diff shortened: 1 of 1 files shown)


martenson:
Should we go further back?

dannon:
@martenson Sure, I guess.  19.05, you think?  I'll retarget this PR.

(My guess is that most folks running on a catalina-based osx machine aren't doing so on their 'stable' deployments, so two versions back is probably safe enough.)

Edit:
So, looking at 19.05, that's using cryptography==2.6.1, which might not have the same problem as 2.7 does.  Checking before force-pushing back over this to retarget.

martenson:
I agree, 19.05 is good target.

lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
```diff
@@ -42,7 +42,7 @@ cloudauthz==0.2.0
 cloudbridge==2.0.0
 cmd2==0.8.9
 contextlib2==0.5.5 ; python_version < '3.5'
-cryptography==2.6.1
+cryptography==2.8
```
nsoranzo, line 44: I think you can also remove `asn1crypto` from this file, which is no more a dependency of `cryptography`: https://github.com/pyca/cryptography/blob/master/CHANGELOG.rst#28---2019-10-16

dannon: Sure.

nsoranzo (approved):

PR #8850, opened 2019-10-22
[19.09] add caret to toolsheds dropdown in admin menu

martenson:
so it is clear this is a menu. cc @natefoo 

before
Galaxy

after
Galaxy

### DIFF
--- a/client/galaxy/scripts/components/Toolshed/Index.vue
+++ b/client/galaxy/scripts/components/Toolshed/Index.vue
@@ -8,8 +8,8 @@
                 @change="setQuery"
                 @keydown.esc="setQuery()"
             />
-            <b-input-group-append>
-                <b-btn :disabled="!queryInput" @click="setQuery()">
+            <b-input-group-append v-b-tooltip.hover title="clear search (esc)">
+                <b-btn @click="setQuery()">
                     <i class="fa fa-times" />
                 </b-btn>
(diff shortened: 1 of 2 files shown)


mvdbeek:
Could we make this more button-ish maybe ? I feel like that's still easy to miss (but definitely better!)

martenson:
@mvdbeek Button feels too dominant/heavyhanded for such a niche function. I think it is enough to have the caret.


Galaxy

mvdbeek:
yeah, 👍

dannon:
Thanks for this.  I had initially missed it was a dropdown for selecting the toolshed as well.

guerler:
Thank you for fixing this. We could also re-use what we have for the installation monitor in #8810. Its a `fa fa-angle-double-down` in front of the link. Otherwise we can revisit this another time. Thank you.

martenson:
@guerler Oh, I did not notice that. I thought of carets as our 'standard' given we use them in masthead and all select2 dropdowns including the toolform. 

This feels more like a menu than the collapsible status you implemented for the monitor. So I am happy with the caret here and the double angle there I think.

guerler:
Cool, that works. Thank you.

guerler (approved):

natefoo (approved):
Thanks @martenson!

Issue #8838, opened 2019-10-22
Option to compare test output size to a value rather than a file


simonbray:
If I want to check the file size of an output when writing a test for a tool, I have to use `compare="sim_size"`. When a test output is large (let's say 1MB) it seems a bit ridiculous to upload a file just to do this comparison. It would be nice to also have an option like `check_size="1000000"`.

Is there a good reason not to implement such a feature? If not, I'd like to give it a try.

simonbray:
Or maybe something like this already exists, but I couldn't find anything in the docs. :)

martenson:
@simonbray I don't think it exists and it sounds like a good idea.

PR #8852, opened 2019-10-23
Add has_size option

simonbray:
Allows checking file size in tests directly without having to use `sim_size` and a test file. See https://github.com/galaxyproject/galaxy/issues/8838


simonbray:
Some questions:
- I guess I should write a test, where should this go?
- `galaxy.xsd` would also have to be updated in the planemo repo, should I also submit a PR for that?
- how can I edit the docs? The source for https://docs.galaxyproject.org/en/latest/dev/schema.html doesn't seem to be on github.

lib/galaxy/tool_util/verify/asserts/size.py
```diff
@@ -0,0 +1,9 @@
+import io
+import sys
+
+
+def assert_has_size(output_bytes, value, delta=0):
+    """Asserts the specified output has a size of the specified value"""
+    output_temp = io.BytesIO(output_bytes)
```
mvdbeek: `output_bytes` is supposed to be bytes already, is that not the case ? iow can you skip io.BytesIO ?

mvdbeek: also technically `sys.getsizeof` returns the memory used by a variable, that's not the same as the file size

mvdbeek: I think `len(output_bytes)` should be accurate here.

mvdbeek:
> * I guess I should write a test, where should this go?

As far as I can tell there are no unit tests for test **assertions**. If you want to write these that would be awesome, it should be relatively straightforward. I think it's also OK if you just add the assertion to one of the test tools in `test/functional/tools`.

> * `galaxy.xsd` would also have to be updated in the planemo repo, should I also submit a PR for that?

We're going (or maybe we already did?) to refactor planemo to get this file from the tool_util package, so no.

> * how can I edit the docs? The source for https://docs.galaxyproject.org/en/latest/dev/schema.html doesn't seem to be on github.

It'll be generated from the schema

nsoranzo:
> We're going (or maybe we already did?) to refactor planemo to get this file from the tool_util package, so no.

I've got a WIP patch locally, but first we need to address https://github.com/common-workflow-language/cwltool/issues/1155

jmchilton (approved):

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -1742,6 +1742,12 @@ module.
       </xs:element>
       <xs:element name="has_archive_member" type="AssertHasArchiveMember">
       </xs:element>
+      <xs:element name="has_size" type="xs:anyType">
+        <xs:annotation>
+          <xs:documentation><![CDATA[Asserts the specified output has a size of the specified value (e.g. ``<has_size value="10000">``).]]>
```
martenson:
```suggestion
          <xs:documentation><![CDATA[Asserts the specified output has a size of the specified value in bytes (e.g. ``<has_size value="10000">``).]]>
```

martenson:
thanks @simonbray !

PR #8854, opened 2019-10-23
[19.09] Fix toolshed selection display glitch

dannon:
Noticed the following bug in firefox when testing #8850 (this bug is unrelated to that code change, though).  When you click the dropdown, the whole thing jumps out into a separate left aligned block.

image

### DIFF
--- a/client/galaxy/scripts/components/Toolshed/ServerSelection.vue
+++ b/client/galaxy/scripts/components/Toolshed/ServerSelection.vue
@@ -40,3 +40,8 @@ export default {
     }
 };
 </script>
+<style scoped>
+span.dropdown {
+    display: inline-block;
+}
+</style>


martenson:
It does not jump for me in FF or FF dev edition. 🤔

dannon:
@martenson Interesting.  Maybe it has to do with my resolution?  I dunno.  In any event, it does jump around like that for me, and swapping from block to inline-block fixes it.

Yeah, I tried multiple resolutions, that's not it.  This is on FF developer 71.0b3 (64-bit):  (gif taken on a branch not including the carat, but it's the same)

busted

martenson:
Can't see it even after update to 71.0b3, I just hope this is not some kind of a shitbug that will kick us elsewhere too.

dannon:
@martenson I just tried chrome and it's doing it there, too.  Both OSX and Windows.  I dunno, man, heh.  In any event, I'll keep my eyes peeled, but this was the only spot I saw it saw far.  Thanks for the merge.

PR #8856, opened 2019-10-23
Refactor default_locale config option; test

jdavcs:
Also, edit mako template to use the config's default

NOTE: This is cherry-picked from #8795


jdavcs:
@dannon that's also the config edit we discussed a week ago. I'll address the rest of the hard-coded values in that template in a follow-up commit.
https://github.com/galaxyproject/galaxy/pull/8856/files#diff-0d9dfb852dbb35fea35a91d2c9e42592R16

PR #8859, opened 2019-10-23
Fix misc. integration tests for config values

jdavcs:
1. Most of the fixes involve removing the `config` prefix from the default
value specified in the schema, and enabling the test for that value.

2. The default values for `oidc_config_file` and `oidc_backends_config_file`
are `xml` files: that's what is specified as defaults in the schema, and
that's what the sample files are in `config/sample/`.  Aslo,
`authnz/managers.py` and `authnz/__init__.py` both refer to `xml` files.

However, the loaded values are `yml` files; that's done here:
https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L677
I've traced it to this commit:
https://github.com/galaxyproject/galaxy/commit/5c5197e84120551b551164f6b73eea2df260a157

These should be `xml` (confirmed: https://github.com/galaxyproject/galaxy/pull/8795#issuecomment-541750971 )

3. Also, for consistency across the file, I've slightly edited the wording
of the note about values to be resolved with respect to another option
(with or without using the `path_resolves_to` attribute).

NOTE: This is cherry-picked from #8795 (partially)


jdavcs:
Unit test failure unrelated; non-deterministic test. So, ready for review.

jmchilton:
Can you backport the yml -> xml changes to 19.05?

PR #8863, opened 2019-10-23
Make message_box style and visible options reloadable

jdavcs:
As per discussion on slack with @natefoo 

@dannon , @guerler - will this cause any issues on the client?


dannon:
@ic4f Should work fine.  The only (fairly minor -- and not new) consideration would be that if you toggle these and they autoreload clients won't see the update until a hard page refresh.

jdavcs:
@dannon yes, of course. And I think someone who goes into the trouble of editing a yml file can live with having to refresh the browser ;)

PR #8865, opened 2019-10-24
Do not specify None as default in dict.get()

jdavcs:
Because it is the default.

Reason for edit: reduce visual clutter before further refactoring.

(This is a borderline pointless edit, but reducing visual clutter in that file, I think, is worth it.)


bgruening (approved):

bgruening:
Thanks @ic4f!

PR #8867, opened 2019-10-24
[WIP] Add is_set(property) method to config; test

jdavcs:
`app.config.is_set(property)` will return `True` if property was set by the user (including setting to `None`), and `False` otherwise. If property is not defined in the schema, an exception is raised.

Purpose: will soon replace the `[property]_set` attribute created for config file options here:
https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L156

Source: discussion with @natefoo .

Limitations: this covers config options defined in the schema only. So, if there's a `foo` attribute defined on `app.config`, which may be set by the user, calling `is_set('foo')` will raise an exception. But, I think, it's best to cross that bridge when we get there.

UPDATE: references to all `_set` attributes replaced with calls to `is_set()` (+attributes themselves no longer created during config loading).

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -173,7 +173,6 @@ class BaseAppConfiguration(object):
         for var, values in defaults.items():
             if config_kwargs.get(var) is not None:
                 path = config_kwargs.get(var)
-                setattr(self, var + '_set', True)
             else:
                 for value in values:
                     if os.path.exists(os.path.join(self.root, value)):
@@ -181,14 +180,12 @@ class BaseAppConfiguration(object):
                         break
                 else:
                     path = values[-1]
-                setattr(self, var + '_set', False)
             setattr(self, var, os.path.join(self.root, path))
 
         for var, values in listify_defaults.items():
             paths = []
             if config_kwargs.get(var) is not None:
                 paths = listify(config_kwargs.get(var))
-                setattr(self, var + '_set', True)
             else:
                 for value in values:
                     for path in listify(value):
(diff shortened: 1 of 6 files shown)


lib/galaxy/config/__init__.py
```diff
@@ -716,6 +717,11 @@ def sentry_dsn_public(self):
         else:
             return None
 
+    def is_set(self, property):
+        if property not in self._raw_config:
+            raise Exception("Property does not exist: '%s'" % property)
```
mvdbeek, line 727: There are some debugging things that are not in the schema ... personally I think logging a warning would be enough here, but not opposed to documenting everything either.

jdavcs: Here's an example: it is used somewhere in code to check if some config file is set, and if no, fall back to the corresponding sample config. And the config file property name is misspelled in that place - so, as a result, it will always fall back to the sample, logging a warning (which will be lost in the output). Also, since this is used by developers and not by users or admins, this may happen only during development. So, I think, failing fast and loudly would be a safer option here? (i.e., "stop now! you are checking a non-existent config...")

As for the debugging use cases, I think they will simply use getattr(), as they do now. In any case, we certainly won't just remove all `foo_set` attributes without ensuring that any `foo` is either in the schema, or is recognized by the new `is_set(foo)` method (with some edit), or has another way to check if it was set.

jdavcs: @mvdbeek may i mark this as resolved? Also, if you think a warning is still a better choice, I'm perfectly fine with that.

jdavcs:
...back to the drawing board on this one.
EDIT: my assumptions for creating the new `is_set()` method were incorrect. `is_set()` will return `True` if an option has been set by the user (i.e., the schema default has been overwritten), which includes overwriting with `None`. However, if an option is overwritten with `None`, the old `[option]_set` will evaluate to `False`. @natefoo this is just an FYI: I haven't abandoned all those config file loading refactorings we've discussed, but this may take more time than I initially thought.

PR #8872, opened 2019-10-24
Update check_py3_compatibility script for new ack versions.

dannon:
ACK_OPTIONS is no longer an option; just include them directly in command invocation.

### DIFF
--- a/.ci/check_py3_compatibility.sh
+++ b/.ci/check_py3_compatibility.sh
@@ -6,17 +6,6 @@ else
     ACK=ack
 fi
 
-export ACK_OPTIONS=" --type python \
---ignore-dir=.git \
---ignore-dir=.tox \
---ignore-dir=.venv \
---ignore-dir=.venv3 \
---ignore-dir=client/node_modules \
---ignore-dir=database \
---ignore-dir=doc/build \
---ignore-dir=eggs \
---ignore-dir=static/maps \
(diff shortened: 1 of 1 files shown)


.ci/check_py3_compatibility.sh
```diff
@@ -28,7 +17,18 @@ xmlrpclib"
 
 ret=0
 for mod in $PYTHON2_ONLY_MODULES; do
-    $ACK "^import $mod(\n|\.)|^from $mod import "
+    $ACK " --type python \
```
nsoranzo:
```suggestion
    $ACK --type python \
```

.ci/check_py3_compatibility.sh
```diff
@@ -28,7 +17,18 @@ xmlrpclib"
...
+    --ignore-dir=.tox \
+    --ignore-dir=.venv \
+    --ignore-dir=.venv3 \
+    --ignore-dir=client/node_modules \
+    --ignore-dir=database \
+    --ignore-dir=doc/build \
+    --ignore-dir=eggs \
+    --ignore-dir=static/maps \
+    --ignore-dir=static/scripts \
+    ^import $mod(\n|\.)|^from $mod import "
```
nsoranzo:
```suggestion
    "^import $mod(\n|\.)|^from $mod import "
```

PR #8875, opened 2019-10-24
Fix and test behavior of non-existent from_work_dir outputs.

### LINKED ISSUE #8743: Jobs with optionally written outputs fail when job destination has outputs_to_working_directory=True
Have a tool that creates a bunch of output files, on an optional basis, using `from_work_dir` -- sometimes the dataset isn't created. This works fine when `outputs_to_working_directory` is False. But the job fails when `outputs_to_working_directory=True` due to:
https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/jobs/__init__.py#L1593
```
        outputs_to_working_directory = util.asbool(self.get_destination_configuration("outputs_to_working_directory", False))
        if outputs_to_working_directory and not self.__link_file_check():
            for dataset_path in self.get_output_fnames():
                try:
                    shutil.move(dataset_path.false_path, dataset_path.real_path)
                    log.debug("finish(): Moved %s to %s" % (dataset_path.false_path, dataset_path.real_path))
                except (IOError, OSError):
                    # this can happen if Galaxy is restarted during the job's
                    # finish method - the false_path file has already moved,
                    # and when the job is recovered, it won't be found.
                    if os.path.exists(dataset_path.real_path) and os.stat(dataset_path.real_path).st_size > 0:
                        log.warning("finish(): %s not found, but %s is not empty, so it will be used instead"
                                    % (dataset_path.false_path, dataset_path.real_path))
                    else:
                        # Prior to fail we need to set job.state
                        job.set_state(final_job_state)
                        return self.fail("Job %s's output dataset(s) could not be read" % job.id)
```

Error message when `outputs_to_working_directory=True` is `Job 1824's output dataset(s) could not be read`.

Job should complete as green, just the same as when `outputs_to_working_directory=False`.

jmchilton:
fixes https://github.com/galaxyproject/galaxy/issues/8743 (w/tests)

### DIFF
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1240,7 +1240,7 @@ class JobWrapper(HasResourceParameters):
                         shutil.move(dataset_path.false_path, dataset_path.real_path)
                         log.debug("fail(): Moved %s to %s", dataset_path.false_path, dataset_path.real_path)
                     except (IOError, OSError) as e:
-                        log.error("fail(): Missing output file in working directory: %s", unicodify(e))
+                        log.debug("fail(): Missing output file in working directory: %s", unicodify(e))
             for dataset_assoc in job.output_datasets + job.output_library_datasets:
                 dataset = dataset_assoc.dataset
                 self.sa_session.refresh(dataset)
@@ -1591,7 +1591,15 @@ class JobWrapper(HasResourceParameters):
 
         outputs_to_working_directory = util.asbool(self.get_destination_configuration("outputs_to_working_directory", False))
         if outputs_to_working_directory and not self.__link_file_check():
-            for dataset_path in self.get_output_fnames():
+            if self.output_paths is None:
+                self.compute_outputs()
+            for (name, (_, dataset_path)) in self.output_hdas_and_paths.items():
+                if dataset_path.false_path and not os.path.exists(dataset_path.false_path):
+                    output = self.tool.find_output_def(name)
+                    if output and getattr(output, 'from_work_dir') is not None:
+                        # This doesn't actually have to be present, unfortunately we cannot distinguish between
+                        # NFS issues and missing outputs in this case.
+                        continue
                 try:
                     shutil.move(dataset_path.false_path, dataset_path.real_path)
                     log.debug("finish(): Moved %s to %s" % (dataset_path.false_path, dataset_path.real_path))
--- a/test/functional/tools/collection_creates_pair_from_work_dir.xml
+++ b/test/functional/tools/collection_creates_pair_from_work_dir.xml
@@ -1,0 +1,30 @@
+<tool id="collection_creates_pair_from_work_dir" name="collection_creates_pair_from_work_dir">
+  <command>
+    echo "Hello World ${input1}" > forward
+  </command>
+  <inputs>
+    <param name="input1" type="integer" value="42">
+    </param>
+  </inputs>
+  <outputs>
+    <collection name="paired_output" type="paired" label="Output Pair">
+      <data name="forward" format="txt" from_work_dir="forward" />
+      <data name="reverse" format_source="txt" from_work_dir="reverse" />
+    </collection>
+  </outputs>
+  <tests>
+    <test>
+      <param name="input1" value="34" />
+      <output_collection name="paired_output" type="paired">
+        <element name="forward">
+          <assert_contents>
+            <has_text_matching expression="^Hello World 34\n$" />
+          </assert_contents>
+        </element>
+        <element name="reverse" checksum="sha1$da39a3ee5e6b4b0d3255bfef95601890afd80709" />
+      </output_collection>
+    </test>
+  </tests>
+  <help>
+  </help>
+</tool>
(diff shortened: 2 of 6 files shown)


test/functional/tools/collection_creates_pair_from_work_dir.xml
```diff
@@ -0,0 +1,30 @@
...
+    echo "Hello World ${input1}" > forward
+  </command>
+  <inputs>
+    <param name="input1" type="integer" value="42">
+    </param>
+  </inputs>
+  <outputs>
+    <collection name="paired_output" type="paired" label="Output Pair">
+      <data name="forward" format="txt" from_work_dir="forward" />
+      <data name="reverse" format_source="txt" from_work_dir="reverse" />
```
bgruening: format_source?

jmchilton: Good catch, rebased with a fix.

lib/galaxy/jobs/__init__.py
```diff
@@ -1591,7 +1591,15 @@ def fail():
...
         if outputs_to_working_directory and not self.__link_file_check():
-            for dataset_path in self.get_output_fnames():
+            if self.output_paths is None:
+                self.compute_outputs()
+            for (name, (_, dataset_path)) in self.output_hdas_and_paths.items():
+                if dataset_path.false_path and not os.path.exists(dataset_path.false_path):
+                    output = self.tool.find_output_def(name)
+                    if output and getattr(output, 'from_work_dir') is not None:
+                        # This doesn't actually have to be present, unfortunately we cannot distinguish between
+                        # NFS issues and missing outputs in this case.
```
mvdbeek, line 1601: This fails with external metadata setting for collection_creates_pair_from_work_dir, but not output_empty_work_dir, probably because output_empty_work_dir doesn't define a datatype:
```
galaxy.metadata DEBUG 2019-11-04 16:03:05,017 setting metadata externally failed for HistoryDatasetAssociation 114: Traceback (most recent call last):
  File "/Users/mvandenb/src/galaxy/lib/galaxy/metadata/set_metadata.py", line 137, in set_metadata_portable
    set_meta(dataset, file_dict)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/metadata/set_metadata.py", line 107, in set_meta
    set_meta_with_tool_provided(new_dataset_instance, file_dict, set_meta_kwds, datatypes_registry, max_metadata_value_size)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/metadata/set_metadata.py", line 68, in set_meta_with_tool_provided
    dataset_instance.datatype.set_meta(dataset_instance, **set_meta_kwds)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/datatypes/data.py", line 816, in set_meta
    dataset.metadata.data_lines = self.count_data_lines(dataset)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/datatypes/data.py", line 835, in count_data_lines
    with compression_utils.get_fileobj(dataset.file_name) as in_file:
  File "/Users/mvandenb/src/galaxy/lib/galaxy/util/compression_utils.py", line 30, in get_fileobj
    return get_fileobj_raw(filename, mode, compressed_formats)[1]
  File "/Users/mvandenb/src/galaxy/lib/galaxy/util/compression_utils.py", line 61, in get_fileobj_raw
    return compressed_format, io.open(filename, mode, encoding='utf-8')
FileNotFoundError: [Errno 2] No such file or directory: '/Users/mvandenb/src/galaxy/database/jobs_directory/000/109/outputs/galaxy_dataset_114.dat'
```

I'm not sure this is a good idea in general, it feels like failing is the proper option here given that nothing indicates that `from_work_dir` will work even if the file doesn't exist. Even if we'd fall back and just `touch` the output file setting the metadata for some datatypes is going to fail (can't `pysam.index()` an empty file for instance) -- and that's also going to fail now if `outputs_to_working_directory` is disabled.

We could introduce a new flag for optional outputs though. It'd still have the metadata problems, but at least it's explicitly allowed to not be present by the tool author. And you could also just discover the datasets if uncertain, or `touch` all `from_work_dir` outputs in the wrapper.

blankenberg: I think from a backwards compatibility perspective, the proposed flag might be backwards. Since `outputs_to_working_directory` defaults to `False`, assuming the behavior of `from_work_dir` had been Green/OK (haven't actually checked older checkouts), existing behavior should be maintained.

Regardless, things of this nature at a tool-design level need to be handled consistently regardless of the job configurations.

mvdbeek:
I think we fixed this somewhere along the way, since the tests are passing now.
An alternative fix is was thinking about is
```
diff --git a/lib/galaxy/jobs/command_factory.py b/lib/galaxy/jobs/command_factory.py
index 0fd189a703..c72d9c2c78 100644
--- a/lib/galaxy/jobs/command_factory.py
+++ b/lib/galaxy/jobs/command_factory.py
@@ -244,7 +244,7 @@ def __handle_metadata(commands_builder, job_wrapper, runner, remote_command_para

 def __copy_if_exists_command(work_dir_output):
     source_file, destination = work_dir_output
-    return f"if [ -f {source_file} ] ; then cp {source_file} {destination} ; fi"
+    return f"if [ -f {source_file} ] ; then cp {source_file} {destination} ; else touch {destination} ; fi"


 class CommandsBuilder:
```
but again, if we don't have to add it let's not do it.

ajkarloss:
I have similar issue with MLST tool in galaxy 21.09. MLST creates one or two output files depends upon the input file. When the second file is not ceated by the tool since there are no results, galaxy sets the state to error. It may take a while to update the galaxy. So, I tried the solution above but the problem still persists. Do I need to restart the galaxy?

Issue #8642, opened 2019-09-13
weberror library + python3


selten:
I believe this issue was caused by my compiled mako templates being outdated (python 2 instead of python 3 templates) as discussed here: https://github.com/galaxyproject/galaxy/issues/8617.

However, somehow in the whole list of dependencies for Galaxy the 'weberror' library seems to be included. This library is, at the time of writing this issue, not python 3 compatible. I have created a pull request to make it Python 3 compatible though (https://github.com/Pylons/weberror/pull/21).

It throws the following errors (examples):
```
Traceback (most recent call last):
  File "lib/galaxy/webapps/galaxy/buildapp.py", line 190, in uwsgi_app
    return galaxy.web.framework.webapp.build_native_uwsgi_app(app_factory, "galaxy")
  File "lib/galaxy/web/framework/webapp.py", line 988, in build_native_uwsgi_app
    uwsgi_app = paste_factory({}, load_app_kwds=app_kwds)
  File "lib/galaxy/webapps/galaxy/buildapp.py", line 163, in app_factory
    webapp = wrap_in_middleware(webapp, global_conf, app.application_stack, **kwargs)
  File "lib/galaxy/webapps/galaxy/buildapp.py", line 1185, in wrap_in_middleware
    from weberror import evalexception
  File "GLXFOLDER/.venv/lib/python3.7/site-packages/weberror/evalexception.py", line 127
    except ValueError, ve:
                     ^
SyntaxError: invalid syntax
```

```
Traceback (most recent call last):
  File "lib/galaxy/webapps/galaxy/buildapp.py", line 190, in uwsgi_app
    return galaxy.web.framework.webapp.build_native_uwsgi_app(app_factory, "galaxy")
  File "lib/galaxy/web/framework/webapp.py", line 988, in build_native_uwsgi_app
    uwsgi_app = paste_factory({}, load_app_kwds=app_kwds)
  File "lib/galaxy/webapps/galaxy/buildapp.py", line 163, in app_factory
    webapp = wrap_in_middleware(webapp, global_conf, app.application_stack, **kwargs)
  File "lib/galaxy/webapps/galaxy/buildapp.py", line 1185, in wrap_in_middleware
    from weberror import evalexception
  File "GLXFOLDER/.venv/lib/python3.7/site-packages/weberror/evalexception.py", line 25, in <module>
    import httplib
ModuleNotFoundError: No module named 'httplib'
```

[...]

PR #8877, opened 2019-10-24
Eliminate interactive debugging middleware

dannon:
xref galaxyproject/galaxy#8642 and https://github.com/galaxyproject/galaxy/issues/1715
(check weberror off 1715 when this is merged)

At one point I prototyped a https://palletsprojects.com/p/werkzeug/ implementation of this which could be an alternative to weberror, but that was a long time ago and I 1) don't remember what was broken, but something wasn't quite ready to go and 2) can't find it anyway.  It just doesn't seem like we need to worry about this, these days, though, so it's probably not worth the effort to reimplement and maintain.

(skipping deprecation notice since it's a developer-only feature, and not one many people know about and/or use anymore (per informal survey) anyway)

### DIFF
--- a/doc/source/admin/galaxy_options.rst
+++ b/doc/source/admin/galaxy_options.rst
@@ -29,10 +29,8 @@
 
 :Description:
     By default, Galaxy uses a SQLite database at
-    'database/universe.sqlite'.  You may use a SQLAlchemy connection
-    string to specify an external database instead.  This string takes
-    many options which are explained in detail in the config file
-    documentation.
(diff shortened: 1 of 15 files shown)


nsoranzo:
@dannon I've added a commit to fix the error middleware loading, please double check.

dannon:
@nsoranzo Good catch, looks fine to me.

nsoranzo (approved):

natefoo (approved):

PR #8878, opened 2019-10-24
[19.09] Galaxy InteractiveTools cluster support fixes from EU

natefoo:
Fixes from @erasche and @bgruening who were the real victims - I am simply upstreaming their work. And basic DRMAA support.

The container resolver sets up a signal 0 trap to kill the container, but even setting up a trap on `SIGKILL`, `SIGINT`, `ERR`, and  `EXIT`, I was unable to cause the container to terminate when running under Slurm. So for the moment in my testbed I'm using a Slurm Epilog script like so:

```bash
#!/bin/bash

[ -n "$SLURM_JOB_ID" ] || { echo "No job id!"; exit 1; }

(
workdir=$(scontrol show job "$SLURM_JOB_ID" | sed 's/.*\( \|^\)WorkDir=\([^ ]*\)\( \|$\).*/\2/p;d')
container_config="${workdir}/container_config.json"

if [ -d "$workdir" -a -f "$container_config" ]; then
    container_id=$(jq -r '.container_name'  "$container_config")
    docker kill "$container_id"
fi
) 2>&1 >/dev/null #>/tmp/epilog."$SLURM_JOB_ID"

exit 0
```

I'm not running this on Test or Main, and when I do get GITs on them, I plan to be using Singularity  containers, which can't escape Slurm.

### DIFF
--- a/lib/galaxy/jobs/runners/drmaa.py
+++ b/lib/galaxy/jobs/runners/drmaa.py
@@ -335,6 +335,9 @@ class DRMAAJobRunner(AsynchronousJobRunner):
                     state = ajs.old_state
                 else:
                     continue
+            if ajs.running:
+                # TODO: stop checking at some point
+                ajs.job_wrapper.check_for_entry_points()
             if ajs.check_limits():
                 self.work_queue.put((self.fail_job, ajs))
                 continue
--- a/lib/galaxy_ext/container_monitor/monitor.py
+++ b/lib/galaxy_ext/container_monitor/monitor.py
@@ -1,15 +1,16 @@
 import json
-import time
 import os
-import subprocess
 import socket
+import subprocess
 import sys
 import tempfile
+import time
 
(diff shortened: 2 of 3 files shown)


bgruening:
Please feel free to take the other IT changes as well. Otherwise, I will do this next week, when I'm back. Do you prefer dev or 19.09?

natefoo:
19.09 as they are bug fixes that will be needed for most/all production deployments.

This was all I needed to get it working (with the [ie2 branch](https://github.com/usegalaxy-eu/gie-nodejs-proxy/tree/ie2) of the Node.js proxy) but I'll look back through the rest of your patches to see if there's more to upstream.

jmchilton (approved):

jmchilton:
(I had posted a comment at a specific line - but Github made it just a general comment in my UI so I deleted it.)

dannon:
@jmchilton For what it's worth, it showed up directly at the line for me.

PR #8879, opened 2019-10-25
[19.09] Install manager unit test and python 3 hg util exception handling fixes

mvdbeek:
On the install manager side these are only cosmetic fixes in normal operation, but I think this has the potential to be buggy when errors occur on python 3. Same for the hg_util fixes.
The unit test is more of a poor excuse, but at least it exercises installation and update (with the actual cloning and TS interaction mocked out and pretending there is no tool to load).

### DIFF
--- a/lib/tool_shed/galaxy_install/install_manager.py
+++ b/lib/tool_shed/galaxy_install/install_manager.py
@@ -461,7 +461,7 @@ class InstallRepositoryManager(object):
             raw_text = util.url_get(tool_shed_url, password_mgr=self.app.tool_shed_registry.url_auth(tool_shed_url), pathspec=pathspec, params=params)
         except Exception:
             message = "Error attempting to retrieve installation information from tool shed "
-            message += "%s for revision %s of repository %s owned by %s:" % \
+            message += "%s for revision %s of repository %s owned by %s" % \
                 (tool_shed_url, changeset_revision, name, owner)
             log.exception(message)
             raise exceptions.InternalServerError(message)
--- a/lib/tool_shed/galaxy_install/install_manager.py
+++ b/lib/tool_shed/galaxy_install/install_manager.py
@@ -120,10 +120,10 @@ class InstallToolDependencyManager(object):
             # There is currently only one fabric method.
             tool_dependency = self.install_and_build_package(install_environment, tool_dependency, actions_dict)
         except Exception as e:
-            log.exception('Error installing tool dependency %s version %s.', str(tool_dependency.name), str(tool_dependency.version))
+            log.exception('Error installing tool dependency %s version %s.', tool_dependency.name, tool_dependency.version)
             # Since there was an installation error, update the tool dependency status to Error. The remove_installation_path option must
             # be left False here.
-            error_message = '%s\n%s' % (self.format_traceback(), str(e))
+            error_message = '%s\n%s' % (self.format_traceback(), util.unicodify(e))
             tool_dependency = tool_dependency_util.set_tool_dependency_attributes(self.app,
                                                                                   tool_dependency=tool_dependency,
                                                                                   status=self.app.install_model.ToolDependency.installation_status.ERROR,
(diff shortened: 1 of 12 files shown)


lib/tool_shed/galaxy_install/repository_dependencies/repository_dependency_manager.py
```diff
@@ -309,8 +309,8 @@ def get_repository_dependencies_for_installed_tool_shed_repository(self, app, re
         pathspec = ['repository', 'get_repository_dependencies']
         try:
             raw_text = url_get(tool_shed_url, password_mgr=app.tool_shed_registry.url_auth(tool_shed_url), pathspec=pathspec, params=params)
-        except Exception as e:
-            log.error("The URL\n%s\nraised the exception:\n%s\n", build_url(tool_shed_url, pathspec=pathspec, params=params), str(e))
+        except Exception:
+            log.exception("The URL\n%s\nraised the exception:\n%s\n", build_url(tool_shed_url, pathspec=pathspec, params=params))
```
nsoranzo:
```suggestion
            log.exception("Error while trying to get URL: %s", build_url(tool_shed_url, pathspec=pathspec, params=params))
```

lib/tool_shed/galaxy_install/install_manager.py
```diff
@@ -453,17 +453,17 @@ def get_repository_components_for_installation(self, encoded_tsr_id, encoded_tsr
...
         pathspec = ['api', 'repositories', 'get_repository_revision_install_info']
         try:
             raw_text = util.url_get(tool_shed_url, password_mgr=self.app.tool_shed_registry.url_auth(tool_shed_url), pathspec=pathspec, params=params)
-        except Exception as e:
+        except Exception:
             message = "Error attempting to retrieve installation information from tool shed "
-            message += "%s for revision %s of repository %s owned by %s: %s" % \
-                (str(tool_shed_url), str(changeset_revision), str(name), str(owner), str(e))
-            log.warning(message)
+            message += "%s for revision %s of repository %s owned by %s:" % \
```
nsoranzo:
```suggestion
            message += "%s for revision %s of repository %s owned by %s" % \
```

nsoranzo (approved):

jmchilton (approved):

lib/tool_shed/util/hg_util.py
```diff
@@ -395,21 +396,21 @@ def init_repository(repo_path):
...
         if isinstance(e, subprocess.CalledProcessError):
-            error_message += "\nOutput was:\n%s" % e.output
+            error_message += "\nOutput was:\n%s" % unicodify(e.output)
         raise Exception(error_message)
 
 
 def changeset2rev(repo_path, changeset_revision):
     """
-    Return the revision number corresponding to a specified changeset revision.
+    Return the numeric revision number (as a string) corresponding to a specified changeset revision.
```
nsoranzo: Numeric number doesn't read well
```suggestion
    Return the revision number (as a string) corresponding to a specified changeset revision.
```

lib/tool_shed/util/repository_util.py
```diff
@@ -89,6 +89,7 @@ def create_or_update_tool_shed_repository(app, name, description, installed_chan
     If a record defined by the received tool shed, repository name and owner does not exist, create
     a new record with the received information.
     """
+    metadata_dict = metadata_dict if metadata_dict is not None else {}
```
nsoranzo:
```suggestion
    metadata_dict = metadata_dict or {}
```

lib/tool_shed/util/hg_util.py
```diff
@@ -395,21 +396,21 @@ def init_repository(repo_path):
...
         rev = subprocess.check_output(['hg', 'id', '-r', changeset_revision, '-n'], stderr=subprocess.STDOUT, cwd=repo_path)
     except Exception as e:
-        error_message = "Error looking for changeset '%s': %s" % (changeset_revision, e)
+        error_message = "Error looking for changeset '%s': %s" % (changeset_revision, unicodify(e))
         if isinstance(e, subprocess.CalledProcessError):
-            error_message += "\nOutput was:\n%s" % e.output
+            error_message += "\nOutput was:\n%s" % unicodify(e.output)
         raise Exception(error_message)
-    return int(rev.strip())
+    return rev.strip()
```
nsoranzo: What was the issue with this? I think I prefer the way it was, with casting to `str` by the caller when needed, especially since:
- `subprocess.check_output()` return a bytestring
- casting to int here finds any issue as early as possible.

mvdbeek: There's only one place in the code where we need `ctx_rev` to be numeric (RepositoryMetadata.numeric_revions), while everywhere else it should be a string.
I agree though that it's nice to check this is an integer,
so how about `return str(int(rev.strip())))` ?

nsoranzo: `changeset2rev` is called only in 3 places, so it's not a big problem to cast in the caller if needed. Anyway, I think I've fixed the toolshed tests by casting to `str` where needed in an updated commit, let me know if you dislike it!

mvdbeek: I would prefer it if ctx_rev was always a string, so I did that in my last commit. Does that seem like a reasonable compromise ?

nsoranzo (approved):

PR #8882, opened 2019-10-25
Normalize myexperiment_target_url / myexperiment_url

jdavcs:
Remove duplication; rely on schema to provide default.

I assume these 2 are the same thing (I briefly checked with @dannon ). I selected the `...target_url` (not to change this: `client/galaxy/scripts/components/Workflow/WorkflowImport.vue`). 

The only change in values is here: now the assigned default will contain port 80:
https://github.com/galaxyproject/galaxy/compare/dev...ic4f:sg_config34?expand=1#diff-46f86ea0d82b84642e4218279ef61442L436

@jmchilton , @guerler : is this correct?


jdavcs:
Rebased!

PR #8883, opened 2019-10-25
Unpin bdbag.

jmchilton:
Python 2.7.9 problem fixed at some point...

https://github.com/fair-research/bdbag/commit/6175ec342f92c1ecb06189be3c6ec61b89bfc52a#diff-2eeaed663bd0d25b7e608891384b7298

### DIFF
--- a/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
+++ b/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
@@ -19,9 +19,9 @@ azure-storage-blob==1.3.1
 azure-storage-common==1.4.2
 azure-storage-nspkg==3.1.0
 babel==2.7.0
-bagit==1.6.4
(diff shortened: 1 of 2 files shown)


lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
```diff
@@ -21,7 +21,7 @@ azure-storage-nspkg==3.1.0
 babel==2.7.0
 bagit==1.6.4
```
nsoranzo: This needs to be updated too, I think. I suppose that's why it's WIP.

nsoranzo:
The new bdbag also requires `setuptools-scm==3.3.3` .

jmchilton:
Thanks for the heads up @nsoranzo. Fixed in a rebase.

nsoranzo (approved):

PR #8884, opened 2019-10-25
Do not override schema defaults if they exist

jdavcs:
Instead of `_defaults_to('some val')`, use schema default.

The following default values have been changed to use schema
defaults (I assume that was the intent):

key: old default / schema default

genomespace_ui_url: none / https://gsui.genomespace.org/jsui/
communication_server_port: none / 7070
communication_server_host: none / http://localhost
persistent_communication_rooms: none / []
require_login: none / false
inactivity_box_content: none / schema text
registration_warning_message: none  / schema text
welcome_url: none / /static/welcome.html

I did not modify default values if the schema default was the opposite (true vs
false) or if it was overwritten in the config loading code.

### DIFF
--- a/lib/galaxy/managers/configuration.py
+++ b/lib/galaxy/managers/configuration.py
@@ -47,7 +47,7 @@ class ConfigSerializer(base.ModelSerializer):
             'myexperiment_target_url'           : _required_attribute,
             'wiki_url'                          : _required_attribute,
             'search_url'                        : _required_attribute,
-            'mailing_lists'                     : _required_attribute,
+            'mailing_lists'                     : _defaults_to(self.app.config.mailing_lists_url),
             'screencasts_url'                   : _required_attribute,
             'genomespace_ui_url'                : _required_attribute,
             'citation_url'                      : _required_attribute,
--- a/lib/galaxy/managers/configuration.py
+++ b/lib/galaxy/managers/configuration.py
@@ -29,67 +29,67 @@ class ConfigSerializer(base.ModelSerializer):
 
     def add_serializers(self):
         def _defaults_to(default):
-            return lambda i, k, **c: getattr(i, k, default)
+            return lambda item, key, **context: getattr(item, key, default)
 
         self.serializers = {
             # TODO: this is available from user data, remove
             'is_admin_user'                     : lambda *a, **c: False,
 
-            'brand'                             : _defaults_to(''),
+            'brand'                             : _defaults_to(self.app.config.brand or ''),
             # TODO: this doesn't seem right
-            'logo_url'                          : lambda i, k, **c: self.url_for(i.get(k, '/')),
-            'logo_src'                          : lambda i, k, **c: self.url_for('/static/favicon.png'),
-            'terms_url'                         : _defaults_to(''),
-            'myexperiment_target_url'           : _defaults_to("www.myexperiment.org"),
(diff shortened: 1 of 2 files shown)


lib/galaxy/managers/configuration.py
```diff
@@ -29,67 +29,67 @@ def default_serializer(self, config, key):
...
-            'message_box_visible'               : _defaults_to(False),
-            'message_box_class'                 : _defaults_to('info'),
-            'server_startttime'                 : lambda i, k, **c: server_starttime,
-            'mailing_join_addr'                 : _defaults_to('galaxy-announce-join@bx.psu.edu'),
-            'server_mail_configured'            : lambda i, k, **c: True if getattr(i, 'smtp_server', False) else False,
-            'registration_warning_message'      : _defaults_to(None),
-            'welcome_url'                       : _defaults_to(None),
-            'show_welcome_with_login'           : _defaults_to(True),
-            'cookie_domain'                     : _defaults_to(None),
+            'require_login'                     : _defaults_to(self.app.config.require_login),
```
jmchilton: This looks really confused right - since ``_defaults_to`` is what is used if ``config`` doesn't have the attribute but the default your filling in is an unprotected access of that exact value from ``config``. This would all read a lot better I think if ``_defaults_to(self.app.config.require_login)`` was just ``_required_attribute``

with 

```python
            def _required_attribute(i, k, **c):
                assert hasattr(i, k)
                return getattr(i, k)
```

jdavcs: @jmchilton  I don't think access to such attributes should be protected: it it's defined in the schema, the config object is guaranteed to have that attribute. Consider these 2 lines: 
1) https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L226 - this is where the value for the future attribute is loaded: it's either the schema default, or `None` (for `require_login`, `data` is this dict: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/config_schema.yml#L1976 ), and ` None` is the default used in `get()`, of course.
2) https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L301 - that's where the attribute is created.

This was one of my goals with the config edits: if there is a `foo` in the schema, it should be always accessible as `config.foo`, regardless of whether a default is specified, or whether it has been overwritten with `None`, etc.

If it's *not* in the schema, then it is either (a) appropriate to override it with a default based on context; or (b) if it's the same across the codebase, it might belong in the schema.

Is this reasonable?

jmchilton: I’m not suggesting you should use a protected access pattern here.

jmchilton:
```python
'require_login': _defaults_to(self.app.config.require_login),
```

causes _defaults_to to be applied to config right? So you're effectively saying:

```python
getattr(self.app.config, 'require_login', self.app.config.require_login)
```

better just to annotate that it has a required value in some fashion. My idea was:

```python
'require_login': _required_attribute,
```

jdavcs: Ah, of course! Yes, this makes sense. I'll make the change. Thanks!

ic4f:
unit test failure unrelated.

jdavcs:
unit test failure unrelated.

jmchilton:
Thanks for the cleanup.

PR #8885, opened 2019-10-26
Use schema defaults in galaxy masthead

jdavcs:
Same edit as has been done to `default_locale` in this file in https://github.com/galaxyproject/galaxy/pull/8856

I've used `or ''` to ensure an empty string instead of null, as there's no default value in the schema.

EDIT: this also fixes a bug: the masthead template would've never received the value of `mailing_lists_url` if that were set by the user (as per documentation) because of this:
`app.config.get("mailing_lists", "https://galaxyproject.org/mailing-lists"` (i.e., `mailing_lists_url` vs `mailing_lists`)


jdavcs:
~~well, that's not good at all...~~ Fixed.

jmchilton:
Looks awesome - thanks!

PR #8886, opened 2019-10-26
[19.09] Discard messages created before QueueWorker was started

mvdbeek:
And don't delete the local control queue. Deleting the local
control queue may throw off other consumers bound to that control
queue. This shouldn't happen, but apparently it can if Galaxy
runs in zerg mode ?


bgruening:
In conjunction with https://github.com/galaxyproject/galaxy/pull/8887 it seems to work on EU. Alone it increases the memory consuption ...

jmchilton (approved):

PR #8887, opened 2019-10-26
[19.09] Fix excessive memory consumption by not loading repository.metadata column

mvdbeek:
This reduces the memory consumed by the ToolShedRepositoryCache for a full rebuild of main's install database from 684 MB to 14 MB. The largest chunk of savings comes from `defer(self.app.install_model.ToolShedRepository.metadata)`. Not eagerloading the unused (except in the galaxy_install module, where we don't load repos from the cache) the metadata makes up the biggest chunk, reducing the memory footprint to 58MB.

Also fixes the metadata generator. I used metadata resetting for all repos as a "benchmark". Total memory consumption doesn't go higher than 782M now.


bgruening:
Thanks so much for working on this and optimizing it! @mvdbeek I think you should add a big comment blog in the code explaining the lessens learnt ...

bgruening:
This seems to work on EU!

PR #8890, opened 2019-10-27
Cleanup config loading (minor fixes)

jdavcs:
These are minor fixes - a few refactorings in config loading + a few edits to clarify/correct schema.
Grouped into one PR because all are similar and trivial.

See details in commit messages.


jdavcs:
py35 unit test failure unrelated (flaky test, passed previously)

PR #8891, opened 2019-10-27
[19.05] Fix unicodify on exceptions with non-utf bytestring args

mvdbeek:


lib/galaxy/util/__init__.py
```diff
@@ -1018,10 +1018,15 @@ def unicodify(value, encoding=DEFAULT_ENCODING, error='replace', strip_null=Fals
...
     >>> s = u'lâtín strìñg'; assert unicodify(s.encode('latin-1')) == u'l\ufffdt\ufffdn str\ufffd\ufffdg'
     >>> s = u'lâtín strìñg'; assert unicodify(s.encode('latin-1'), error='ignore') == u'ltn strg'
+    >>> assert unicodify(Exception(u'¼ cup of flour'.encode('latin-1')), error='ignore') == ' cup of flour'
     """
     if value is None:
         return value
     try:
+        if isinstance(value, Exception):
+            # This is for backwards compatibility for python 2, but works on python 3 as well
+            if value.args:
```
nsoranzo:
```suggestion
            if len(value.args) == 1:
```

mvdbeek: In what cases would `if value.args` not work ? Is the idea to not assume that the message is in the first arg ? As far as I understand this would for instance skip casting IOError exceptions (https://docs.python.org/2/library/exceptions.html#exceptions.BaseException.args)

mvdbeek: ok, just trying it I see the point:
```
>> try:
...     open('bla123')
... except Exception as e:
...     x = e
...
>>> e
IOError(2, 'No such file or directory')
>>> e.args
(2, 'No such file or directory')
```

mvdbeek: I guess we could cast all bytestring elements in args and replace the args tuple, but depending on what the exception does we may break more things than we fix, so let's just go with your suggestion.

nsoranzo (approved):

lib/galaxy/util/__init__.py
nsoranzo: Yeah, `e.args` can have more than one element and `str(e)` (and `unicode(e)`) behave differently if there is exactly one element.

PR #8893, opened 2019-10-28
PDF Export of Galaxy Markdown

jmchilton:
Builds on other Markdown Pages #8512 and extra Markdown validation #9099.

What is here:
- Add an export link to the page when rendering Markdown for workflow exports and Markdown pages.
- Implement API and backend code to generate Markdown from these documents:
  - Generate "internal Galaxy markdown" for the target document - IDs are integers - not workflow labels used by report definition or encoded IDs used by the front-end to render. 
   - Run this through ``to_basic_markdown`` - that generates a Markdown document without Galaxy extensions used by the client Vue component. Try to replace dynamic content with static variants at this step. 
    - Use the [Python-Markdown](https://github.com/Python-Markdown/markdown) library to generate HTML from that. Haven't found any good reason to use this versus [markdown2](https://github.com/trentm/python-markdown2). 
    - Bleach the resulting HTML using galaxy.util.santize_html - added a test case for this given security concerns.
    - Convert it to PDF using [WeasyPrint](https://weasyprint.org/) (had to pin an older version since it doesn't support Python 2 anymore).
- Implement an option to attach a custom stylesheet to the export from HTML to PDF to brand PDFs generated by a Galaxy instance.

## Alternatives

I think most of the other ways to generated PDF from Markdown in Python - aren't pure Python solutions (e.g. https://stackoverflow.com/a/31117621 or pandoc). WeasyPrint might not scale and we may need to introduce pandoc as a dependency - but probably best to start here and see how far it will take us? Even if we do implement pandoc so we can do like Word Docs also or something - at least this could provide the fallback implementation for when Pandoc is unavailable.


mvdbeek:
Hmm, seems that there are some problems with the weasyprint wheels?:
```
  File "/Users/mvandenb/.vscode/extensions/ms-python.python-2019.11.50794/pythonFiles/ptvsd_launcher.py", line 43, in <module>
    main(ptvsdArgs)
  File "/Users/mvandenb/.vscode/extensions/ms-python.python-2019.11.50794/pythonFiles/lib/python/old_ptvsd/ptvsd/__main__.py", line 432, in main
    run()
  File "/Users/mvandenb/.vscode/extensions/ms-python.python-2019.11.50794/pythonFiles/lib/python/old_ptvsd/ptvsd/__main__.py", line 316, in run_file
    runpy.run_path(target, run_name='__main__')
  File "/Users/mvandenb/.pyenv/versions/3.6.9/lib/python3.6/runpy.py", line 263, in run_path
    pkg_name=pkg_name, script_name=fname)
  File "/Users/mvandenb/.pyenv/versions/3.6.9/lib/python3.6/runpy.py", line 96, in _run_module_code
    mod_name, mod_spec, pkg_name, script_name)
  File "/Users/mvandenb/.pyenv/versions/3.6.9/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/mvandenb/src/galaxy/scripts/paster.py", line 23, in <module>
    serve.run()
  File "/Users/mvandenb/src/galaxy/lib/galaxy/util/pastescript/serve.py", line 1055, in run
    invoke(command, command_name, options, args[1:])
  File "/Users/mvandenb/src/galaxy/lib/galaxy/util/pastescript/serve.py", line 1061, in invoke
    exit_code = runner.run(args)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/util/pastescript/serve.py", line 226, in run
    result = self.command()
  File "/Users/mvandenb/src/galaxy/lib/galaxy/util/pastescript/serve.py", line 642, in command
    app = loadapp(app_spec, name=app_name, relative_to=base, global_conf=vars)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/util/pastescript/loadwsgi.py", line 294, in loadapp
    return loadobj(APP, uri, name=name, **kw)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/util/pastescript/loadwsgi.py", line 320, in loadobj
    return context.create()
  File "/Users/mvandenb/src/galaxy/lib/galaxy/util/pastescript/loadwsgi.py", line 762, in create
    return self.object_type.invoke(self)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/util/pastescript/loadwsgi.py", line 276, in invoke
    filtered = context.next_context.create()
  File "/Users/mvandenb/src/galaxy/lib/galaxy/util/pastescript/loadwsgi.py", line 762, in create
    return self.object_type.invoke(self)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/util/pastescript/loadwsgi.py", line 188, in invoke
    return fix_call(context.object, context.global_conf, **context.local_conf)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/util/pastescript/loadwsgi.py", line 91, in fix_call
    val = callable(*args, **kw)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/webapps/galaxy/buildapp.py", line 62, in app_factory
    webapp.add_ui_controllers('galaxy.webapps.galaxy.controllers', app)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/web/framework/webapp.py", line 131, in add_ui_controllers
    module = import_module(module_name)
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/Users/mvandenb/src/galaxy/lib/galaxy/webapps/galaxy/controllers/page.py", line 22, in <module>
    from galaxy.managers.pages import (
  File "/Users/mvandenb/src/galaxy/lib/galaxy/managers/pages.py", line 17, in <module>
    from galaxy.managers.markdown_util import (
  File "/Users/mvandenb/src/galaxy/lib/galaxy/managers/markdown_util.py", line 25, in <module>
    from weasyprint import CSS, HTML
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.6/site-packages/weasyprint/__init__.py", line 375, in <module>
    from .css import preprocess_stylesheet  # noqa
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.6/site-packages/weasyprint/css/__init__.py", line 29, in <module>
    from . import computed_values
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.6/site-packages/weasyprint/css/computed_values.py", line 16, in <module>
    from .. import text
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.6/site-packages/weasyprint/text.py", line 18, in <module>
    import cairocffi as cairo
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.6/site-packages/cairocffi/__init__.py", line 50, in <module>
    ('libcairo.so', 'libcairo.2.dylib', 'libcairo-2.dll'))
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.6/site-packages/cairocffi/__init__.py", line 45, in dlopen
    raise OSError(error_message)  # pragma: no cover
OSError: no library called "cairo" was found
no library called "libcairo-2" was found
cannot load library 'libcairo.so': dlopen(libcairo.so, 2): image not found
cannot load library 'libcairo.2.dylib': dlopen(libcairo.2.dylib, 2): image not found
cannot load library 'libcairo-2.dll': dlopen(libcairo-2.dll, 2): image not found
```

dannon:
@mvdbeek I had to install pango and cairo libraries to get it to work earlier.  (on a mac `brew install pango cairo` worked.

jmchilton:
Sad day - I'll hide this behind a flag. If we need optional external things ... we should probably do pandoc from conda-forge maybe? I don't know...

lib/galaxy/managers/markdown_util.py
```diff
@@ -66,71 +237,178 @@ def ready_galaxy_markdown_for_export(trans, internal_galaxy_markdown):
...
+def to_pdf(trans, basic_markdown, css_paths=[]):
+    as_html = to_html(basic_markdown)
+    directory = tempfile.mkdtemp('gxmarkdown')
+    index = os.path.join(directory, "index.html")
+    try:
+        output_file = codecs.open(index, "w", encoding="utf-8", errors="xmlcharrefreplace")
+        output_file.write(as_html)
+        output_file.close()
+        html = HTML(filename=index)
+        stylesheets = [CSS(string=pkg_resources.resource_string(__name__, 'markdown_export_base.css'))]
```
dannon: This file doesn't exist for me, preventing generation of the pdf.

jmchilton: Missing git add. I'll push it in the next rebase - thanks for the catch.

bgruening:
@jmchilton have you considered to use a Galaxy tool for this, or will this take too long?
I assume those reports will become very popular and we might want to schedule those conversions. Then the pandoc dependency would also not be a problem.

jmchilton:
@bgruening I feel like this would lead to a "history export"-like user experience and set of hacks through the backend. We will probably need to do that at some point but I think it will require like a month of re-engineering this kind of job - and it is not clear to me right now what that should look like. Presumably the outputs should not be tracked in histories and this requires internals we don't want to expose to tools - seems like it might be better to using a MQ w/worker pool and the dependency manager instead of shoving everything into "tools".

mvdbeek:
I had a similar suggestion as @bgruening but I agree that a MQ would be nicer

bgruening:
@jmchilton I don't think it is required in this PR. But I feel like we need to outsource those jobs to the cluster and not compute it on the head node. Aysam is already using a Vis tool to execute large plottings on the cluster. History exports are similar I agree, as they also tend to be large. Another example is anykind of compression ... like library exports. Ideally, this does not trash the head node disc.

For the MQ, maybe that is better, but the jobs really need to go to the cluster imho.

jmchilton:
> this does not trash the head node disc.

The reports grab only a bounded amount of data per dataset - so it is bounded in a way that history exports and collection downloads are not. I agree this could be a serious problem though, but lets wait until we have evidence that it is before abandoning simpler APIs and interfaces.

jmchilton:
Given that the direct PDF download might not always be available without doing the PUT dance that history exports do... I should mark this API as beta before we finally release 20.0X. Doesn't need to to hold up branching though.

Sorry for the delay @mvdbeek and thanks for the merge @dannon !

lib/galaxy/config/__init__.py
```diff
@@ -757,6 +757,9 @@ def parse_config_file_options(self, kwargs):
             user_preferences_extra_conf_path=[self._in_config_dir('user_preferences_extra_conf.yml')],
             workflow_resource_params_file=[self._in_config_dir('workflow_resource_params_conf.xml')],
             workflow_schedulers_config_file=[self._in_config_dir('workflow_schedulers_conf.xml')],
+            markdown_export_css=[self._in_config_dir('markdown_export.css')],
+            markdown_export_css_pages=[self._in_config_dir('markdown_export_pages.css')],
+            markdown_export_css_invocation_reports=[self._in_config_dir('markdown_export_invocation_reports.css')],
```
mvdbeek, line 762: @jmchilton Did you mean to ship those css files with Galaxy / forgot to git add ? Or is that meant as an optional thing ?
I think the .css can specify some things we should do by default, like re-size images to fit on page.

jmchilton: These are for optional local instance theming - there is a file that applies globally though that was committed as part of this PR https://github.com/galaxyproject/galaxy/pull/8893/files#diff-b7164468c47f82e12c9ce4cc7ab669e6.

mvdbeek: Awesome, thanks!

PR #8894, opened 2019-10-28
If config_file is None, use cwd, not its parent

jdavcs:
Addresses this comment:
https://github.com/galaxyproject/galaxy/pull/921#discussion_r304351292

@natefoo , @nsoranzo - is this correct?

(the added empty lines improve readability, I hope)

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -111,6 +111,7 @@ class BaseAppConfiguration(object):
         self.global_conf_parser = configparser.ConfigParser()
         if not self.config_file and self.global_conf and "__file__" in self.global_conf:
             self.config_file = os.path.join(self.root, self.global_conf['__file__'])
+
         if self.config_file is None:
             log.warning("No Galaxy config file found, running from current working directory: %s", os.getcwd())
         else:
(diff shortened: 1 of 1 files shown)


nsoranzo (approved):

nsoranzo:
I'm still not sure if overwriting `self.config_dir` at line 139 is correct, maybe we can refactor that too?

jmchilton:
This needs to be back ported to 19.09 I think.

jmchilton (approved):

jdavcs:
@nsoranzo about line 139: I don't know. As I understand, if the config file is not specified, cannot be found, etc., and galaxy is running from source tree, just use the `config/` dir in the source tree. Should it default to some other location?

jdavcs:
backporting in a min..

nsoranzo:
@ic4f One issue is that if a value is specified for `'config_dir'` in `config_kwargs`, it is overwritten by `os.path.join(self.root, 'config')` .

jdavcs:
@nsoranzo you're right! (`config_file` looked too much like `config_dir` - so i didn't notice this!). I'll suggest a fix later today.

PR #8901, opened 2019-10-29
[19.09] Revert hacking IPs in container monitor script.

jmchilton:
It breaks the Mac setup and seems just wrong, docker might not be on localhost. This needs to be made more specific in some way and should handle those socker commands not working on OS X.

### DIFF
--- a/lib/galaxy/tool_util/deps/docker_util.py
+++ b/lib/galaxy/tool_util/deps/docker_util.py
@@ -225,6 +225,12 @@ def _docker_prefix(
 
 
 def parse_port_text(port_text):
+    """
+
+    >>> slurm_ports = parse_port_text("8888/tcp -> 0.0.0.0:32769")
+    >>> slurm_ports[8888]['host']
+    '0.0.0.0'
+    """
     ports = None
     if port_text is not None:
         ports = {}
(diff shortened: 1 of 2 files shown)


natefoo:
Yeah, I didn't take too good of a look at this, sorry about that.

bgruening:
I understand this, but merging this revert will break our and Nates setup afaik. I'm afraid we are sacrificing real-world deployment for OSX users and hypothetical deployments when we merge the PR as it is :(

nsoranzo:
There's a flake8 error to fix.

jmchilton:
It is marked WIP - I'm not suggesting we merge as is, but someone should fix it. I don't have a slurm or condor setup so I'd hope someone else offer but if no one will I can give it a try. Can someone at least provide a dump of that ``ports`` variable on your respective setup so I can think more about this?

natefoo:
Here's the raw output:

```console
[root@clone0n0 ~]# docker port d2b7340119564d63bcb0277141c67339
8888/tcp -> 0.0.0.0:32769
```

jmchilton:
Does this feel like an okay compromise?

bgruening (approved):

PR #8280, opened 2019-07-06
get tests on data package running

dglaetzer:

PR #8905, opened 2019-10-29
Refactor galaxy_install and required dependencies into Galaxy's namespace.

jmchilton:
The goal in my mind would be for Galaxy to depend ultimately on nothing in the ``tool_shed`` package so they could be cleanly separated into different repositories (or source tree however they are laid out) with non-overlapping code and no duplication.


mvdbeek (approved):

packages/app/setup.py
```diff
@@ -63,6 +63,22 @@ def get_var(var_name):
     'galaxy.tools.toolbox.lineages',
     'galaxy.tools.util',
     'galaxy.tools.util.galaxyops',
+    'galaxy.tool_shed',
+    # TODO:
+    # 'galaxy.tool_shed.galaxy_install',
```
guerler: Can you resolve these todos to complete the transition? If I understand this correctly`galaxy_install` is used only by Galaxy but is located in `lib/tool_shed` and imports toolshed utilities which have not been moved to `lib/galaxy/tool_shed` yet.

jmchilton: I'm happy to complete the transition - I was hoping to get this piece merged to ensure the conflict with #8830 wouldn't prevent this from being merged and the approach would not be -1'd by anyone before continuing but I am committed to getting this done. A smaller PR is inline with say https://github.com/galaxyproject/galaxy/pull/8280 - not resulting in the perfect package structure but an incremental step in the correct direction.

If I have your assurance this approach is acceptable though I will continue to work on this. If your okay with the approach is there a reason to not merge this piece as is?

jmchilton:
Moving back to WIP - messed up hg_util transformation here. I wonder why toolshed tests didn't catch this.

packages/app/setup.py
jmchilton: That is all of galaxy_install that is needed by Galaxy after the controller changes in your PR I suppose. No need to transfer grids or utility containers I suppose?

guerler: Right, and I think it would be great to have this transition within this PR i.e. move `galaxy_install` to `lib/galaxy/tool_shed`. #8830 can wait for that.

lib/tool_shed/metadata/metadata_generator.py
```diff
@@ -1,1111 +1,3 @@
-import json
-import logging
-import os
-import tempfile
+from galaxy.tool_shed.metadata.metadata_generator import MetadataGenerator
```
mvdbeek, line 1: ❤️

packages/app/setup.py
```diff
@@ -63,6 +63,22 @@ def get_var(var_name):
     'galaxy.tools.toolbox.lineages',
     'galaxy.tools.util',
     'galaxy.tools.util.galaxyops',
+    'galaxy.tool_shed',
+    'galaxy.tool_shed.galaxy_install',
+    'galaxy.tool_shed.galaxy_install.datatypes',
+    # TODO:
+    # 'galaxy.tool_shed.galaxy_install.grids',
```
mvdbeek, line 70: This can be removed from Galaxy after the separation, right ?

packages/app/setup.py
```diff
@@ -63,6 +63,22 @@ def get_var(var_name):
...
+    'galaxy.tool_shed.galaxy_install.metadata',
+    'galaxy.tool_shed.galaxy_install.migrate',
+    'galaxy.tool_shed.galaxy_install.migrate.versions',
+    'galaxy.tool_shed.galaxy_install.repository_dependencies',
+    'galaxy.tool_shed.galaxy_install.tool_dependencies',
+    'galaxy.tool_shed.galaxy_install.tool_dependencies.recipe',
+    'galaxy.tool_shed.galaxy_install.tools',
+    # 'galaxy.tool_shed.galaxy_install.utility_containers',
```
mvdbeek, line 78: And this too ?

mvdbeek (approved):
Awesome, thanks so much @jmchilton!

packages/app/setup.py
jmchilton: These are still used by the web controllers in this branch but are deleted from #8830. So I think when we clean out the web controllers they can just disappear and I didn't bother to migrate them. I don't understand this to be tied to the tool shed separation since the tool shed doesn't depend on them - it is more about the tests being replaced or removed.

mvdbeek: OK, I assumed that the toolshed uses them (because I also thought Galaxy doesn't use those controllers anymore). Fine to delete them in #8830.

PR #3572, opened 2017-02-07
Set 'deleted' to True when purging a history dataset association

nsoranzo:
Fix https://github.com/galaxyproject/galaxy/issues/3548 , bug introduced in commit 58f677ad4f4473d0549dd417f187cc1f7a3a9a4e.

Add API and unit test cases.
Add migration to fix misaligned databases.

PR #8908, opened 2019-10-30
make pgcleanup also delete HDAs

bernt-matthias:
follow up to https://github.com/galaxyproject/galaxy/pull/3572

### DIFF
--- a/scripts/cleanup_datasets/pgcleanup.py
+++ b/scripts/cleanup_datasets/pgcleanup.py
@@ -304,7 +304,7 @@ class PurgesHDAs(object):
                           implicitly_converted_dataset_association.id AS id),
              deleted_icda_purged_child_hda_ids
           AS (     UPDATE history_dataset_association
-                      SET purged = true{update_time_sql}
+                      SET purged = true, deleted = true{update_time_sql}
                      FROM deleted_icda_ids
                     WHERE deleted_icda_ids.hda_id = history_dataset_association.id),
              metadata_file_events
@@ -437,7 +437,7 @@ class UpdateHDAPurgedFlag(Action):
     _action_sql = """
         WITH purged_hda_ids
           AS (     UPDATE history_dataset_association
-                      SET purged = true
+                      SET purged = true, deleted = true
                      FROM dataset
                         WHERE history_dataset_association.dataset_id = dataset.id
                           AND dataset.purged
@@ -575,7 +575,7 @@ class PurgeDeletedUsers(PurgesHDAs, RemovesMetadataFiles, Action):
                      FROM purged_history_ids),
              purged_hda_ids
           AS (     UPDATE history_dataset_association
-                      SET purged = true{update_time_sql}
+                      SET purged = true, deleted = true{update_time_sql}
                      FROM purged_history_ids
                     WHERE purged_history_ids.id = history_dataset_association.history_id
                           AND NOT history_dataset_association.purged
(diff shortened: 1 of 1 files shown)


bernt-matthias:
Not sure if the change in https://github.com/galaxyproject/galaxy/pull/3572 is actually as intended (as far as I got it). It seems that Galaxy still ends up with HDAs with deleted = False and purged = True. 

Also not sure if the old cleanup script is doing deleting purged histories.

natefoo (approved):
This makes it consistent with purging histories from the UI/API, although I still disagree with the premise.

bernt-matthias:
> I still disagree with the premise.

My main concern is that pgcleanup does not delete data sets (at least @nsoranzo told me so :) I have only little clue) if one of their data sets is purged but not deleted. One could also change this. 

But I guess consistency is important.

natefoo:
It currently [shouldn't care](https://github.com/galaxyproject/galaxy/blob/de36ef10d6d46481c1a741b2dd7b2a31fbae2db2/scripts/cleanup_datasets/pgcleanup.py#L916) about the HDA deleted flag when deciding to delete a dataset. Or am I misunderstanding the issue?

bernt-matthias:
My original problem (and I'm not sure yet if this PR solves it.. ) is that pgcleanup does not **purge** many of my deleted data sets. The suggestion of @nsoranzo was that HDAs with deleted=f (and purged=t) could be the problem.

nsoranzo:
> My original problem (and I'm not sure yet if this PR solves it.. ) is that pgcleanup does not **purge** many of my deleted data sets. The suggestion of @nsoranzo was that HDAs with deleted=f (and purged=t) could be the problem.

This surely is a problem for `scripts/cleanup_datasets/cleanup_datasets.py` : https://github.com/galaxyproject/galaxy/blob/dev/scripts/cleanup_datasets/cleanup_datasets.py#L360-L364

bernt-matthias:
As @nsoranzo noticed another place where HDAs are purged but deleted is here: 

https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/controllers/history.py#L344

he might come up with a fix soon. Maybe we can add this here?

nsoranzo:
@bernt-matthias I think a separate PR for that would be better. Given @natefoo approved, I'm going to merge this.

bernt-matthias:
@nsoranzo ok

Was just wondering how to trigger purging of data sets (using pgcleanup) which already have an HDA with deleted=false and purged=true. Should one trigger again: 

https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/model/migrate/versions/0134_hda_set_deleted_if_purged.py

nsoranzo:
I can re-add the migration as migration 0160 as part of my upcoming PR. In the mean time, you can obtain the same effect by executing this SQL statement: `UPDATE history_dataset_association SET deleted=TRUE WHERE purged AND NOT deleted;`

bernt-matthias:
Thanks @nsoranzo . I tried this on my dev instance. Lets see if the pgcleanup cron job cleans everything tomorrow...

nsoranzo:
@bernt-matthias Both the issue in https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/controllers/history.py and the migration postponing are addressed in https://github.com/galaxyproject/galaxy/pull/8940 .

PR #8914, opened 2019-10-30
Simplify logo replacement, fix height and alignment

guerler:


dannon:
This does take a good step towards unifying the logo, but it should also remove the rest of the references to `favicon.ico` and set a header link on the core templates specifying the `rel="icon"`, indicating the new logo.  Otherwise browsers will continue to try to fetch two potentially different icons.

guerler:
Makes sense, I'll follow-up on it. Thank you!

PR #8917, opened 2019-10-31
[WIP] If skipping client build, skip node install.

jmchilton:
No reason for doc building tests to fail on install node.

https://jenkins.galaxyproject.org/job/doc-building-test/1382

In general there may be times you want node installed despite not needing the client built, but I don't think we setup these install commands (and version pinning) for those use cases we did it for client building.

### DIFF
--- a/scripts/common_startup.sh
+++ b/scripts/common_startup.sh
@@ -219,21 +219,21 @@ else
     echo "The Galaxy client build is being skipped due to the SKIP_CLIENT_BUILD environment variable."
 fi
 
-# Install node if not installed
-if [ -n "$VIRTUAL_ENV" ]; then
-    if ! in_venv "$(command -v node)" || [ "$(node --version)" != "v${NODE_VERSION}" ]; then
-        echo "Installing node into $VIRTUAL_ENV with nodeenv."
-        nodeenv -n "$NODE_VERSION" -p
-    fi
-elif [ -n "$CONDA_DEFAULT_ENV" ] && [ -n "$CONDA_EXE" ]; then
-    if ! in_conda_env "$(command -v node)"; then
-        echo "Installing node into '$CONDA_DEFAULT_ENV' Conda environment with conda."
-        $CONDA_EXE install --yes --override-channels --channel conda-forge --channel defaults --name "$CONDA_DEFAULT_ENV" nodejs="$NODE_VERSION"
-    fi
-fi
-
 # Build client if necessary.
 if [ $SKIP_CLIENT_BUILD -eq 0 ]; then
+    # Install node if not installed
+    if [ -n "$VIRTUAL_ENV" ]; then
+        if ! in_venv "$(command -v node)" || [ "$(node --version)" != "v${NODE_VERSION}" ]; then
+            echo "Installing node into $VIRTUAL_ENV with nodeenv."
+            nodeenv -n "$NODE_VERSION" -p
+        fi
+    elif [ -n "$CONDA_DEFAULT_ENV" ] && [ -n "$CONDA_EXE" ]; then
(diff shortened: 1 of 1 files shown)


nsoranzo:
I think that's not possible any more because of expression tools, see https://github.com/galaxyproject/galaxy/pull/7556

jmchilton:
I mean that is what I was referring to with...

> In general there may be times you want node installed despite not needing the client built, but I don't think we setup these install commands (and version pinning) for those use cases we did it for client building.

... would you be okay with at least not pinning the version of node for expression tools? I think that might avoid this install everytime as well?

jmchilton:
https://github.com/galaxyproject/galaxy/pull/7556/files#r267546530

> As for this not being hidden behind a flag, I suspect at some point someone will ask for a flag (given 6e865a3 it will probably be me) but I'd wait until someone asks to add that logic because we could be wrong and hopefully are.

😆

jmchilton:
I guess this is wrong because of the unit tests...

nsoranzo:
> would you be okay with at least not pinning the version of node for expression tools? I think that might avoid this install everytime as well?

As suggested by @dannon on Gitter, I think it would be better to try to cache the wheels for the test builds.

martenson:
> but I don't think we setup these install commands (and version pinning) for those use cases

Maybe we could define the other use cases that need node and give them their own path of obtaining it. Then everything else can be excluded.

nsoranzo:
> As suggested by @dannon on Gitter, I think it would be better to try to cache the wheels for the test builds.

The following patch implements pip caching on Docker, but not sure if it would work on Jenkins nodes:

```diff
$ git diff -- run_tests.sh
diff --git a/run_tests.sh b/run_tests.sh
index b95451d8da..dbaea00cdc 100755
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -313,6 +313,9 @@ then
     docker --version
     echo "Launching docker container for testing with extra args ${DOCKER_RUN_EXTRA_ARGS}..."
     name=$(python -c 'import re; import uuid; print re.sub("-","",str(uuid.uuid4()))')
+    # Create a cache dir for pip, so it has the right owner
+    DOCKER_PIP_CACHE_DIR="$HOME"/.cache/docker_galaxy_pip
+    mkdir -p "$DOCKER_PIP_CACHE_DIR"
     _on_exit() {
         docker kill $name
     }
@@ -321,8 +324,10 @@ then
         -e "BUILD_NUMBER=$BUILD_NUMBER" \
         -e "GALAXY_TEST_DATABASE_TYPE=$db_type" \
         -e "LC_ALL=C" \
+        -e "PIP_CACHE_DIR=/pip_cache_dir" \
         --rm \
         --name=$name \
+        -v "$DOCKER_PIP_CACHE_DIR":/pip_cache_dir \
         -v "$(pwd)":/galaxy \
         -v "$(pwd)"/test/docker/base/run_test_wrapper.sh:/usr/local/bin/run_test_wrapper.sh "$DOCKER_IMAGE" "$@"
     exit $?
```

mvdbeek:
That patch looks nice, want to PR it @nsoranzo ?

nsoranzo:
> That patch looks nice, want to PR it @nsoranzo ?

Sure, https://github.com/galaxyproject/galaxy/pull/8929 .

I'm just not sure how much it would be useful on Jenkins, i.e. if the home directory on Jenkins' nodes is preserved across builds.

mvdbeek:
I think it is, but if it is not we can probably override it in the build with `DOCKER_PIP_CACHE_DIR=${DOCKER_PIP_CACHE_DIR:-"$HOME"/.cache/docker_galaxy_pip}` ?

PR #8921, opened 2019-11-01
Simplify Python deps installation in `testing-base` Dockerfile

nsoranzo:
Also use `become` instead of `sudo` in Ansible playbook (`sudo` is deprecated since Ansible 1.9).


mvdbeek (approved):
Great, thanks @nsoranzo!

PR #8924, opened 2019-11-01
Config: drop redundant `template_cache` attribute

nsoranzo:
Just use `template_cache_path` everywhere.


jdavcs (approved):
Yes! One special case less.

Also, *not* for this PR, but as a note to self/to all, the following might be similar cases:
`proxy_session_map` vs `dynamic_proxy_session_map`
`manage_dynamic_proxy` vs `dynamic_proxy_manage`
`jobs_directory` vs `jobs_working_directory`
`mailing_lists` vs `mailing_lists_url` (this one not in `config/__init__`, but, likely, same thing)

bgruening (approved):

PR #8926, opened 2019-11-04
Enable secondary authorization

VJalili:
adds option to authz users against external authz providers, e.g., gcp.


jmchilton:
Thanks!

PR #8929, opened 2019-11-04
Set up a cache dir for pip when running tests with Docker

nsoranzo:
This may need a Jenkins plugin like https://plugins.jenkins.io/jobcacher to be useful for our CI jobs.


nsoranzo:
It seems that (at least for some Jenkins jobs) the pip cache implemented by this PR is being used without any change on Jenkins side.

mvdbeek (approved):

Issue #8801, opened 2019-10-14
Automatic Format Conversion happening, but wrong format used in next workflow step


hepcat72:
I used the CLIP-Seq workflow from this tutorial:

https://galaxyproject.github.io/training-material/topics/transcriptomics/tutorials/clipseq/tutorial.html

At one point, a step produces a bedgraph, but the next step takes a bigwig and fails with an error indicative that it received a bedgraph. The history however shows that a bigwig was automatically created and if you download and re-upload that file, the next step will succeed if you use it. Also, if I add an explicit conversion step in the workflow, it works.

So it seems there’s a problem with automatic format conversion.

mvdbeek:
There is no step that uses bigwig inputs in the workflow or the tutorial. Could you tell me what you're doing for this failure to appear ?

hepcat72:
Oh geez. I apologize. I linked the wrong tutorial/workflow.  I was working on both a CLIP-Seq and Methylation Seq project at the same time.  It's the methylation-seq tutorial/workflow that had this issue:

https://galaxyproject.github.io/training-material/topics/epigenetics/tutorials/methylation-seq/tutorial.html

There is a text reformatting step on a bedgraph file that is connected to a computeMatrix step that takes bigwig.

hepcat72:
Here are some details...

In the json of the workflow in the tutorial, step 20 (computeMatrix) takes the output of step 19 (a bedgraph file from the "awk/text reformatting" tool) as bigwig input.  In this screenshot, you can see that it gets automatically converted to bigwig:

<img width="274" alt="bedgraphnotbigwig" src="https://user-images.githubusercontent.com/2300532/67430119-9a667a80-f5af-11e9-9b94-e16cd2b49418.png">

Yet, computeMatrix fails with the error:

```
Fatal error: Exit code 1 ()
[bwHdrRead] There was an error while reading in the header!
Traceback (most recent call last):
  File "/galaxy/galaxy-app/tool-dependencies/_conda/envs/__deeptools@2.5.1/bin/computeMatrix", line 11, in <module>
    main(args)
  File "/galaxy/galaxy-app/tool-dependencies/_conda/envs/__deeptools@2.5.1/lib/python3.6/site-packages/deeptools/computeMatrix.py", line 422, in main
    hm.computeMatrix(scores_file_list, args.regionsFileName, parameters, blackListFileName=args.blackListFileName, verbose=args.verbose, allArgs=args)
[...]

PR #8932, opened 2019-11-05
[19.05] Fix implicitly converted datasets unused when mapping over

mvdbeek:
Fixes https://github.com/galaxyproject/galaxy/issues/8801
Also sets `textable` when dataset needs to be converted, so users can enter columns up front.


bernt-matthias:
I successfully tested trinity on this this branch. Hooray. Thanks a lot. 

Out of interest, why is this change targeting 19.05 and not dev?

jmchilton (approved):

mvdbeek:
Damn, we don't have access to conda dependencies in the API tests. I'll see if I can get another conversion to work.

mvdbeek:
> Out of interest, why is this change targeting 19.05 and not dev?

It fixes a bug ... one that never worked, but it was supposed to work. It's a relatively small change.

~~Turns out though that selecting columns from implicitly converted datasets also doesn't work.
For that I'll target against dev because it seems that more changes are necessary.~~

jmchilton:
This is really elegant and awesome!

bernt-matthias:
Thanks a lot. Again out of interest: how will this find its way to 19.09 and dev?

mvdbeek:
We'll merge this forward ... this can either happen via PR (if you want to see the tests passing) or direct commit. Forward merges don't need approval (though I don't think we explicitly state this in https://github.com/galaxyproject/galaxy/blob/dev/doc/source/project/organization.rst#handling-pull-requests).

nsoranzo:
@mvdbeek It's in https://github.com/galaxyproject/galaxy/blob/dev/doc/source/project/organization.rst#direct-commit-access : "A committer may only commit directly to the Galaxy repository... cherry-picking and/or merging of existing approved commits to other branches."

bernt-matthias:
Superb. Will this happen automatically? If there is a chance to forget it, then I would volunteer to backport it.

mvdbeek:
It's not automatic, but @nsoranzo probably did this, as there are no commits in 19.05 that are not in 19.09 and dev (https://github.com/galaxyproject/galaxy/compare/release_19.09...release_19.05?expand=1)

bernt-matthias:
Thanks for the clarification, and for reminding me that this tool exists on the github page.

PR #8935, opened 2019-11-06
Fix unbounded variable error in `GET /api/datasets/{encoded_dataset_id}` exception handling

mvdbeek:
Fixes https://sentry.galaxyproject.org/sentry/main/issues/302429/:
```
UnboundLocalError: local variable 'rval' referenced before assignment
  File "galaxy/web/framework/decorators.py", line 157, in decorator
  File "galaxy/webapps/galaxy/api/datasets.py", line 159, in show
    return rval
```


nsoranzo:
Wouldn't it be easier to just remove the try/except altogether?

nsoranzo (approved):

PR #8936, opened 2019-11-06
Add link for interactive tools

anuprulez:


bgruening:
Thanks @anuprulez!

jmchilton:
This confused Anton also, I think this is a usability bug that can be backported to 19.09 even. Thanks so much.

jmchilton (approved):

client/galaxy/scripts/components/ToolEntryPoints/ToolEntryPoints.vue
```diff
@@ -25,12 +25,15 @@
             </ul>
         </div>
 
-        You may also access all active InteractiveTools from the User menu.
+        You may also access all active InteractiveTools from the
+        <a :href="appRoot() + 'interactivetool_entry_points/list'">User menu</a>.
```
dannon: This should probably be a computed property.

dannon:
TS test is unrelated to this.  Thanks for the fix, and the tweaks @anuprulez

PR #8938, opened 2019-11-06
[19.09] Broaden library description/info/message

martenson:
This is a bit of a hack but given the lack of clarity of what
field does what on what object (info/message/description on ldda/ld/folder)
it should help with the data presentation.

Will allow things like this to work well: https://github.com/usegalaxy-eu/shared-data/pull/5

cc @bgruening


bgruening:
I think this is a reasonable choice, especially if it is currently very hard to get informations to the end-user in an automatic way. Thanks Martin!

mvdbeek (approved):

PR #8939, opened 2019-11-06
[19.09] check_for_entry_points() more agressively in the condor job runner

bgruening:
Using the condor job runner, this change is needed as otherwise jobs are missed and a URL is not generated until handler restart.

### DIFF
--- a/lib/galaxy/jobs/runners/condor.py
+++ b/lib/galaxy/jobs/runners/condor.py
@@ -176,10 +176,7 @@ class CondorJobRunner(AsynchronousJobRunner):
             job_id = cjs.job_id
             galaxy_id_tag = cjs.job_wrapper.get_id_tag()
             try:
-                if os.stat(cjs.user_log).st_size == cjs.user_log_size:
-                    if cjs.job_wrapper.tool.tool_type == 'interactive':
-                        # If running, check for entry points...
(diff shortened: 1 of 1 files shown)


lib/galaxy/jobs/runners/condor.py
```diff
@@ -177,6 +177,9 @@ def check_watched_items(self):
             galaxy_id_tag = cjs.job_wrapper.get_id_tag()
             try:
                 if os.stat(cjs.user_log).st_size == cjs.user_log_size:
+                    if cjs.job_wrapper.tool.tool_type == 'interactive':
+                        # If running, check for entry points...
+                        cjs.job_wrapper.check_for_entry_points()
```
mvdbeek: Would it work if you instead said `if cjs.job_wrapper.tool.tool_type != 'interactive' and os.stat(cjs.user_log).st_size == cjs.user_log_size` and then leave it the way it is now ? That way you'd call `check_for_entry_points ` only if condor reports the job as running (= `s1 and not (s4 or s7)`)

mvdbeek (approved):

bgruening:
I hope this will work, I hadn't time to test this in production - sorry. This entire runner needs to be replaced.

PR #4574, opened 2017-09-07
Fix delete option in history menu

dannon:
Updates history options delete and purge to use modern models/refresh.  This fixes #4444

Issue #7735, opened 2019-04-12
deleting history should send you to a recent/new history


martenson:
xref https://github.com/galaxyproject/galaxy/issues/5472

bernt-matthias:
Seems to be solved in https://github.com/galaxyproject/galaxy/pull/8940?

PR #8940, opened 2019-11-07
Set `deleted` attribute of history datasets when purging histories from grid

nsoranzo:
Use `history_manager.purge()` which implements that correctly.
Also:
- Remove unused `delete()` method of `HistoryController`, last use removed in PR #4574 .
- Remove unused code. These grid operations were dropped in commit f3b08e6 .

### DIFF
--- a/lib/galaxy/webapps/galaxy/controllers/history.py
+++ b/lib/galaxy/webapps/galaxy/controllers/history.py
@@ -293,23 +293,6 @@ class HistoryController(BaseUIController, SharableMixin, UsesAnnotations, UsesIt
                         kwargs['refresh_frames'] = ['history']
                 elif operation == "undelete":
                     status, message = self._list_undelete(trans, histories)
-                elif operation == "unshare":
-                    for history in histories:
-                        for husa in trans.sa_session.query(trans.app.model.HistoryUserShareAssociation) \
-                                                    .filter_by(history=history):
-                            trans.sa_session.delete(husa)
-                elif operation == "enable import via link":
-                    for history in histories:
-                        if not history.importable:
-                            self._make_item_importable(trans.sa_session, history)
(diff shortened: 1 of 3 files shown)


jmchilton (approved):

lib/galaxy/webapps/galaxy/controllers/history.py
```diff
@@ -283,33 +302,13 @@ def list(self, trans, **kwargs):
                     else:
                         kwargs['refresh_frames'] = ['history']
                 elif operation in ("delete", "delete permanently"):
-                    if operation == "delete permanently":
-                        status, message = self._list_delete(trans, histories, purge=True)
-                    else:
-                        status, message = self._list_delete(trans, histories)
+                    status, message = self._list_delete(trans, histories, purge=(operation == "delete permanently"))
```
bernt-matthias, line 305: I like this kind of code.

lib/galaxy/webapps/galaxy/controllers/history.py
```diff
@@ -373,8 +351,16 @@ def _list_delete(self, trans, histories, purge=False):
                 part += " but the datasets were not removed from disk because that feature is not enabled in this Galaxy instance"
             message_parts.append("%s.  " % part)
         if deleted_current:
+            # if this history is the current history for this session,
```
bernt-matthias, line 354: Cool. Seems to solve https://github.com/galaxyproject/galaxy/issues/7735

bernt-matthias (approved):
Seems to be fine for me. But note I'm still lacking a lot of knowledge and overview on the project.

bernt-matthias:
Should this go to 19.09?

nsoranzo:
Given there is migration that will fix this retroactively when an admin updates to 20.01, I think it is not critical to backport it.

PR #8943, opened 2019-11-07
Move Jobs updated time display logic to client.

jmchilton:
Displays actual time now on mouseover in addition to elapsed time now.

I want this for two reasons - first to eliminate the the admin/jobs web controller, we really should be using the jobs API for this component. This brings us closer to having these two controllers in sync (still quite a bit work to do I'm sure). Second, I want to have a date display component for the invocations grid after merge of #8699 - the time display in those grids is very rough currently.

### DIFF
--- a/client/galaxy/scripts/components/UtcDate.vue
+++ b/client/galaxy/scripts/components/UtcDate.vue
@@ -1,0 +1,51 @@
+<template>
+    <span class="utc-time" v-if="mode == 'date'" :title="elapsedTime">
+        {{ date }}
+    </span>
+    <span class="utc-time utc-time-elapsed" v-else :title="date">
+        {{ elapsedTime }}
+    </span>
+</template>
+
+<script>
+export default {
+    props: {
+        date: {
+            type: String,
+            required: true
+        },
+        mode: {
+            type: String,
+            default: "date" // or elapsed
+        }
+    },
+    computed: {
+        elapsedTime: function() {
+            var utcUpdate = new Date(this.date);
+            var now = new Date();
+            var utcNow = new Date(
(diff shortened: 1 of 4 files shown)


client/galaxy/scripts/components/UtcDate.vue
```diff
@@ -0,0 +1,51 @@
...
+            type: String,
+            required: true
+        },
+        mode: {
+            type: String,
+            default: "date" // or elapsed
+        }
+    },
+    computed: {
+        elapsedTime: function() {
```
dannon, line 25: We have moment as dependency right now (due to handsontable, so if we're changing that out this is moot), and if you wanted to use that it'd simplify the bits below quite a bit.

(not a required or even requested change, just throwing this out there for thought)

Nerdinacan:
Would this be better implemented as a directive or a filter perhaps?

```html
<span>{{ someRawDate | utcdate }}</span>
<span v-utcdate>{{ rawDate }}</span>
```

Or maybe as just a model property? I think I had a property of my client-side model container that did this on the history model at one point or another.

```javascript
class Job {

    get utcDate() {
        return moment.utc(this.update_time);
    }
}
```

client/galaxy/scripts/components/UtcDate.vue
martenson: :+1: for momentjs, pure JS is horrendous with dates

martenson (approved):

martenson:
@Nerdinacan the current implementation feels like a pretty clean solution to me, especially since this will hopefully be reused often. Do you still think it'd be better implemented differently?

Issue #8851, opened 2019-10-23
modifying ldda through UI fails with tag error


martenson:
This happens when you try modify the dataset at e.g. `http://127.0.0.1/library/list#folders/<folder_id>/datasets/<ldda_id>`, anything you want to change will fail with tag error unless a tag is already present (and you are unable to change tags).

`{"err_msg": "tags must be a list of strings: <type 'unicode'>", "err_code": 400008}`

PR #8944, opened 2019-11-08
[19.09] Validate tag strings as well as list of tags

mvdbeek:
I think it'd be better if we consistently used lists for tags,
but it seems too late in 19.09 to make possibly bigger changes
here. Fixes https://github.com/galaxyproject/galaxy/issues/8851
broken in https://github.com/galaxyproject/galaxy/pull/8530.

### DIFF
--- a/lib/galaxy/managers/library_datasets.py
+++ b/lib/galaxy/managers/library_datasets.py
@@ -139,10 +139,7 @@ class LibraryDatasetsManager(datasets.DatasetAssociationManager):
                 val = validation.validate_and_sanitize_basestring(key, val)
                 validated_payload[key] = val
             if key in ('tags'):
-                if isinstance(val, list):
-                    val = validation.validate_and_sanitize_basestring_list(key, val)
-                else:
-                    val = validation.validate_and_sanitize_basestring(key, val)
+                val = validation.validate_and_sanitize_basestring_list(key, util.listify(val))
                 validated_payload[key] = val
         return validated_payload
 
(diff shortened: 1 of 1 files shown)


lib/galaxy/managers/library_datasets.py
```diff
@@ -139,7 +139,10 @@ def _validate_and_parse_update_payload(self, payload):
                 val = validation.validate_and_sanitize_basestring(key, val)
                 validated_payload[key] = val
             if key in ('tags'):
-                val = validation.validate_and_sanitize_basestring_list(key, val)
+                if isinstance(val, list):
+                    val = validation.validate_and_sanitize_basestring_list(key, val)
+                else:
+                    val = validation.validate_and_sanitize_basestring(key, val)
```
nsoranzo: What about:

```python
                val = listify(val)
                val = validation.validate_and_sanitize_basestring_list(key, val)
```

I think that at line 108 it is assumed that this must be a list.

mvdbeek: Yes, but I don't want to change the incoming payload here. `tag1,tag2` would then become `['tag1,tag2']`, that seems worse.

nsoranzo: No, `galaxy.util.listify()` splits a string on commas.

nsoranzo:
There's an unrelated commit from #8945, can you remove it?

mvdbeek:
Sure, done.

nsoranzo (approved):

martenson:
I glanced at it yesterday and there are more issues, please don't merge yet.

martenson:
I'll open a non-blocking followup.

PR #8946, opened 2019-11-08
[19.09] Fix package requirements

mvdbeek:
Seems needed for running the `galaxy-config` script:
```
  File "/Users/mvandenb/src/galaxy-packaging/.venv/bin/galaxy-config", line 11, in <module>
    load_entry_point('galaxy-app', 'console_scripts', 'galaxy-config')()
  File "/Users/mvandenb/src/galaxy-packaging/.venv/lib/python3.7/site-packages/pkg_resources/__init__.py", line 489, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/Users/mvandenb/src/galaxy-packaging/.venv/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2852, in load_entry_point
    return ep.load()
  File "/Users/mvandenb/src/galaxy-packaging/.venv/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2443, in load
    return self.resolve()
  File "/Users/mvandenb/src/galaxy-packaging/.venv/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2449, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/Users/mvandenb/src/galaxy/packages/app/galaxy/config/__init__.py", line 46, in <module>
    from galaxy.web.formatting import expand_pretty_datetime_format
  File "/Users/mvandenb/src/galaxy/packages/web_framework/galaxy/web/__init__.py", line 5, in <module>
    from .framework import url_for
  File "/Users/mvandenb/src/galaxy/packages/web_framework/galaxy/web/framework/__init__.py", line 5, in <module>
    from . import base
  File "/Users/mvandenb/src/galaxy/packages/web_framework/galaxy/web/framework/base.py", line 19, in <module>
    from paste.request import get_cookies
ModuleNotFoundError: No module named 'paste'
```

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -48,12 +48,12 @@ from galaxy.web_stack import (
     get_stack_facts,
     register_postfork_function
 )
+from galaxy.webapps.galaxy import GALAXY_CONFIG_SCHEMA_PATH
 from ..version import VERSION_MAJOR
 
 log = logging.getLogger(__name__)
 
 GALAXY_APP_NAME = 'galaxy'
-GALAXY_CONFIG_SCHEMA_PATH = 'lib/galaxy/webapps/galaxy/config_schema.yml'
 LOGGING_CONFIG_DEFAULT = {
     'disable_existing_loggers': False,
     'version': 1,
--- a/packages/job_execution/requirements.txt
+++ b/packages/job_execution/requirements.txt
@@ -1,1 +1,2 @@
 galaxy-data
+pulsar-galaxy-lib
--- a/lib/galaxy/config/config_manage.py
+++ b/lib/galaxy/config/config_manage.py
@@ -36,6 +36,7 @@ from galaxy.util.yaml_util import (
     ordered_dump,
     ordered_load,
 )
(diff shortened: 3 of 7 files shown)


packages/job_execution/requirements.txt
```diff
@@ -1 +1,2 @@
 galaxy-data
+pulsar-galaxy-lib
```
jmchilton, line 2: Pulsar imports this package... so this can't fly.

lib/galaxy/config/__init__.py
```diff
@@ -48,12 +48,12 @@
     get_stack_facts,
     register_postfork_function
 )
+from galaxy.webapps.galaxy import GALAXY_CONFIG_SCHEMA_PATH
```
jmchilton, line 51: galaxy-webapps imports galaxy-app - so I think we need to move the schema not do this.

jdavcs: The schema file is only referenced in `config/__init__.py` and `config/config_manage.py`; and, I think, those are the only two places it should be referenced from. So, considering a schema file is integral to `galaxy`, +1 for moving it (into `galaxy`, I assume). (@martenson - related to what we discussed..)

PR #8948, opened 2019-11-08
Email notifications on job completion

dannon:
@astrovsky01 this wraps up what we'd talked about, let me know how it looks to you.


jmchilton (approved):
Need to fix the migration number - I'll do that before merging. Thanks Dannon!

jmchilton:
Thanks @dannon - really nice idea!

martenson:
I am a bit concerned about the volume of emails we might be generating through this and potentially pushing us towards spam filters. @natefoo do you have an opinion?

martenson:
I am a bit concerned about the volume of emails we might be generating through this and potentially pushing us towards spam filters. @natefoo do you have an opinion?

astrovsky01:
@martenson even though it's an opt-in selection per job?

martenson:
Well it is on every job, we run hundreds of thousands a month just on Main, if people don't use it I am not afraid, but if 10% of jobs will use it we will suddenly be sending a lot of emails.

dannon:
My hunch is that 10% of jobs is an incredibly high expectation considering that workflows and any jobs submitted via the API won't see this (but have had the functionality for years).  The other bit of this is that these emails are explicitly requested by the user and, as such, I don't expect them to be reported as spam.

In any event, it's a reasonable consideration and it's worth keeping an eye on, maybe set a reminder to see usage after this has gone live?

If it does appear to be a problem, which I doubt, we can always add a configuration flag to disable notification emails, and then we can set that exclusively for main so other servers can continue to use the functionality.

martenson:
> maybe set a reminder to see usage after this has gone live

Great idea, querying the job params in like March and checking the volume would quantify this.

mvdbeek:
For the release notes, this is what the button looks like:
Screenshot 2020-02-04 at 11 52 06

PR #8949, opened 2019-11-08
Refactor Toolshed webapp into tool_shed package.

jmchilton:
Refactor ``galaxy.webapps.tool_shed`` into ``tool_shed.webapp``, continuing the goal of separating the Tool Shed from Galaxy and eliminating any imports of ``tool_shed`` namespace from the ``galaxy`` namespace.

Builds on #8905.


mvdbeek (approved):
Nice!

PR #8950, opened 2019-11-08
[19.09] Fix the running icon wobblies

dannon:
This fixes running icon 'wobblies'.

I started an upgrade to fa5 which will allow much more flexible usage (svg and js-based glyphs), but this is a good incremental step which makes a big improvement right now without touching much else, so I'm more comfortable with targeting 19.09 with it.


mvdbeek (approved):
Nice, thanks @dannon!

PR #8185, opened 2019-06-18
[19.05] Fix missing terms and conditions on login/register pages.

dannon:
Includes minor restructuring of component containers for reusability and easier formatting of the login landing page, and a client format.

PR #8951, opened 2019-11-09
[19.09] Several registration fixes.

dannon:
This fixes the registration warning blurb ("please don't create duplicate accounts bit" on main), and the missing mailing list join checkbox.  Additionally it embeds the terms in an iframe to the right of registration, similarly to how we do for the welcome on login.


dannon:
~Ugh, just looking through the history I had also done https://github.com/galaxyproject/galaxy/pull/8185/files, which never got merged.  Marking this WIP to consolidate some changes from that, I guess.~

The substantive fixes are in this branch; I'll follow up with the refactoring from #8185 separately so we can just get this in.

mvdbeek (approved):
Nice! 
Screenshot 2019-11-13 at 16 19 51

PR #8953, opened 2019-11-10
grouping: strip newlines

### LINKED ISSUE #8383: Tool Grouping1 (Group data) adds extra lines for 1 column input
The tool "Group data by a column and perform aggregate operation on other columns. (Galaxy Version 2.1.3)" generates 1 extra line with the result of the aggregate function such as "count" for each line in the "group by" column. Expected to have no extra lines, so that each line in "group by" column as the result of the aggregate function on the same line. Occurs with 1-column tabular input, but not with 2 column input.

I think usegalaxy has the current version: id="Grouping1" name="Group" version="2.1.3".

To reproduce, see:
history = dat_190726_1.test_join_tool
This history is currently accessible via link and published:
https://usegalaxy.org:/u/tshtatland/h/dat1907261testjointool

Input 1-column tabular file:

```
a
b
c
b
c
c
```

Incorrect output (6 lines, expected 3 lines:
```
a
	1
b
	2
c
	3
```

**Temporary workaround:**

As seen in the history above. Add a dummy second column to your 1-column tabular file. For example, paste the file with itself. Use tool "Paste two files side by side (Galaxy Version 1.0.0)" with your 1-column file as input 1 and input 2:

```
a	a
b	b
c	c
b	b
c	c
c	c
```

With this input, the Group tool works as expected.

bernt-matthias:
fixes https://github.com/galaxyproject/galaxy/issues/8383

more generally: if grouping is done on the last column then the newline was retained and contained in the output.

Alternatively one could call something like `line.rstrip("\r\n")` if stripping of arbitrary white spaces from the grouping column might be to intrusive.

### DIFF
--- a/tools/stats/grouping.py
+++ b/tools/stats/grouping.py
@@ -124,7 +124,9 @@ def main():
             item = line.split("\t")[group_col]
         except IndexError:
             stop_err("The following line didn't have %s columns: %s" % (group_col + 1, line))
-
+        # in case of grouping by the last column the newline character needs to be removed
+        # removal of other white spaces seems also useful
+        item = item.strip()
         if ignorecase == 1:
             return item.lower()
         return item
--- a/tools/stats/grouping.py
+++ b/tools/stats/grouping.py
@@ -121,12 +121,9 @@ def main():
 
     def is_new_item(line):
         try:
-            item = line.split("\t")[group_col]
+            item = line.rstrip("\r\n").split("\t")[group_col]
         except IndexError:
             stop_err("The following line didn't have %s columns: %s" % (group_col + 1, line))
-        # in case of grouping by the last column the newline character needs to be removed
-        # removal of other white spaces seems also useful
-        item = item.strip()
         if ignorecase == 1:
             return item.lower()
         return item
--- a/tools/stats/grouping.xml
+++ b/tools/stats/grouping.xml
@@ -1,4 +1,4 @@
-<tool id="Grouping1" name="Group" version="2.1.3">
+<tool id="Grouping1" name="Group" version="2.1.4">
   <description>data by a column and perform aggregate operation on other columns.</description>
   <requirements>
     <requirement type="package" version="1.15.4">numpy</requirement>


bernt-matthias:
Guess this bug was introduced here https://github.com/galaxyproject/galaxy/commit/b18aa199aa57e87ef62013f5c4542f7491ec9b19#diff-f89ca29e3a9027fdd2a0bc64e5f51912 where the `line.strip()` was removed. 

Maybe my suggestion to use `line.strip("\r\n")` could be a good idea.

bernt-matthias:
Also guess a backport to 19.05 would be necessary.

tools/stats/grouping.py
```diff
@@ -121,10 +121,9 @@ def main():
 
     def is_new_item(line):
         try:
-            item = line.split("\t")[group_col]
+            item = line.rstrip("\r\n").split("\t")[group_col]
```
mvdbeek, line 124: `rstrip("\r\n")` will chop off any number of trailing newlines ... but I guess this is probably ok.

mvdbeek (approved):

tools/stats/grouping.py
mvdbeek: but I think we should bump the version number.

nsoranzo: This is already operating on a single line, I don't think there can be more than one "\r\n".

nsoranzo:
I think we should add `-f` to the `sort` command line on line 111 when `ignorecase == 1`, otherwise lines with the same lowercased value but different letter case can be grouped separately.

bernt-matthias:
> I think we should add -f to the sort command line on line 111 when ignorecase == 1, otherwise lines with the same lowercased value but different letter case can be grouped separately.

Seems that this is already implemented: https://github.com/galaxyproject/galaxy/blob/2ea05e81f8549eb63bfd4db9f6b2971e1d0848e3/tools/stats/grouping.py#L112

nsoranzo:
> > I think we should add -f to the sort command line on line 111 when ignorecase == 1, otherwise lines with the same lowercased value but different letter case can be grouped separately.
> 
> Seems that this is already implemented:
> 
> https://github.com/galaxyproject/galaxy/blob/2ea05e81f8549eb63bfd4db9f6b2971e1d0848e3/tools/stats/grouping.py#L112

Ah, not sure how I missed it, I guessed I expected to see it before the inputfile, not appended at the end, but if it works no problem.

nsoranzo (approved):

mvdbeek:
I agree we should backport this, want to do that @bernt-matthias ?

bernt-matthias:
19.09 cherry picking done. 

stuggling with 19.05:

- `git checkout release_19.05`
- `git checkout -b topic/19.05-grouping-strip-newline`
- `git cherry-pick 3a0d9325a6df4c2fa6e8ce8cc39e6c067890a7fe 2ea05e81f8549eb63bfd4db9f6b2971e1d0848e3`

gives me

```
[topic/19.05-grouping-strip-newline 7c312147c6] grouping: strip newlines
 Date: Sun Nov 10 22:16:00 2019 +0100
 1 file changed, 3 insertions(+), 1 deletion(-)
error: Your local changes to the following files would be overwritten by merge:
	packages/data/tests/unittest_utils
Please commit your changes or stash them before you merge.
Aborting
fatal: cherry-pick failed
```

the file seems unrelated.

Any ideas.

mvdbeek:
you've still had some unstaged changes. you can do `git cherry-pick --abort` and look at your current state. If you decide you don't need those changes anymore you can `git reset --hard`, or if you want to store those changes for later you can `git stash` to put them in a buffer that you can `git apply` to another branch.

bernt-matthias:
Strange thing is that: 

`git status`

```
On branch topic/19.05-grouping-strip-newline
Your branch is up to date with 'release_19.05'.

nothing to commit, working tree clean
```

mvdbeek:
Haven't seen that before, but got this as well with `git cherry-pick 3a0d9325a6df4c2fa6e8ce8cc39e6c067890a7fe 2ea05e81f8549eb63bfd4db9f6b2971e1d0848e3`. Running
```
git cherry-pick 3a0d9325a6df4c2fa6e8ce8cc39e6c067890a7fe
git cherry-pick 2ea05e81f8549eb63bfd4db9f6b2971e1d0848e3
```
worked fine though.

PR #8956, opened 2019-11-11
Remove tool shed API examples.

jmchilton:
These aren't tested, bioblend likely has better documentation (https://bioblend.readthedocs.io/en/latest/api_docs/toolshed/all.html), and we should be encouraging TS interactions to be done through Planemo and not raw API calls.


mvdbeek (approved):
Some of these operations aren't available in bioblend (capsule import/export), but I think it's ok to drop this. Are you sure `reset_metadata_on_repositories.py` isn't useful anymore ?
I never had to use it, but we also don't have a replacement afaict.

martenson:
@mvdbeek we occasionally used metadata reset in the MTS admin interface, but it has been a year or more since I had to do it I think. I am in favor of removing it, it shouldn't be necessary imo, and if it is, we should fix it properly instead of using external API call that provides no extra arguments. (this is an argument for removing the whole API that does this)

Issue #8952, opened 2019-11-10
19.09 toolshed installation errors


bgruening:
1: All repositories with a minor update, like Alevin (908a8d400fa2)

![grafik](https://user-images.githubusercontent.com/469983/68528469-600d1500-02f3-11ea-9f87-98796ae22205.png)

will be turned into the `new` state during the update with ephemeris

2: Going from `new` state to `install`

```
Nov 09 13:23:50 sn04.bi.uni-freiburg.de uwsgi[4107748]: tool_shed.util.hg_util ERROR 2019-11-09 13:23:50,455 [p:4124529,w:4,m:0] [uWSGIWorker4Core0] Error cloning repository: Command '['hg', 'clone', '-r', '6', u'https://toolshed.g2.bx.psu.edu/repos/iuc/bcftools_call', u'/opt/galaxy/shed_tools/toolshed.g2.bx.psu.edu/repos/iuc/bcftools_call/3296c3680d96/bcftools_call']' returned non-zero exit status 255
Nov 09 13:23:50 sn04.bi.uni-freiburg.de uwsgi[4107748]: Output was:
Nov 09 13:23:50 sn04.bi.uni-freiburg.de uwsgi[4107748]: abort: destination '/opt/galaxy/shed_tools/toolshed.g2.bx.psu.edu/repos/iuc/bcftools_call/3296c3680d96/bcftools_call' is not empty
```

Here it would help now to also install the repo if it is either in `new` state or in `error` state I think.

```
Nov 09 13:22:10 sn04.bi.uni-freiburg.de uwsgi[4107748]: [pid: 4124485|app: 0|req: 9911/42155] 88.66.136.83 () {84 vars in 1575 bytes} [Sat Nov  9 13:22:10 2019] POST /admin_toolshed/repository_installation_status_updates => generated 2 bytes in 23 msecs (HTTP/1.1 200) 3 headers in 96 bytes (1 switches on core 0)
Nov 09 13:22:11 sn04.bi.uni-freiburg.de uwsgi[3191341]: galaxy.datatypes.registry DEBUG 2019-11-09 13:22:11,317 [p:119216,w:4,m:0] [Thread-1] Loaded external metadata tool: __SET_METADATA__
[...]

PR #8959, opened 2019-11-11
[19.09] Make repository updates more robust

mvdbeek:
Fixes the tracebacks in #8952 and skips repository installation if requesting an uninstallable revision whose next installable revision is already installed. Should also be more resilient towards on disk repositories being on disk when installing a repository (use fetch+pull instead of clone) and repositories being absent when updating (use clone instead of fetch). Also prefers installing repos into shed_tool_conf.xml files. Includes integration tests for install/uninstall/update

### DIFF
--- a/lib/galaxy/model/tool_shed_install/__init__.py
+++ b/lib/galaxy/model/tool_shed_install/__init__.py
@@ -92,6 +92,8 @@ class ToolShedRepository(object):
         Return the in-memory version of the shed_tool_conf file, which is stored in the config_elems entry
         in the shed_tool_conf_dict.
         """
+        if default is None:
+            default = {}
 
         def _is_valid_shed_config_filename(filename):
             for shed_tool_conf_dict in app.toolbox.dynamic_confs(include_migrated_tool_conf=True):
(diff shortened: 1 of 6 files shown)


lib/galaxy/model/tool_shed_install/__init__.py
```diff
@@ -92,6 +92,8 @@ def get_shed_config_dict(self, app, default=None):
         Return the in-memory version of the shed_tool_conf file, which is stored in the config_elems entry
         in the shed_tool_conf_dict.
         """
+        if default is None:
+            default = {}
```
nsoranzo: I'd just remove the `default` argument for both this method and `guess_shed_config()`, it's never used in their calls except in `lib/tool_shed/metadata/metadata_generator.py` , where it's set to `{}` anyway.

Also, I think on line 105:
```python
            self.guess_shed_config(app, default=default)
```

should be:

```python
            return self.guess_shed_config(app, default=default)
```

or, if you agree with the suggestion above:

```python
            return self.guess_shed_config(app)
```

mvdbeek: Yep, makes sense to me.

lib/galaxy/model/tool_shed_install/__init__.py
```diff
@@ -137,16 +136,16 @@ def guess_shed_config(self, app, default=None):
                             if tool_id in tool_ids:
                                 self.shed_config_filename = name
                                 return shed_tool_conf_dict
-        if self.includes_datatypes or self.includes_data_managers:
+        if self.includes_datatypes or self.includes_data_managers or not default:
```
nsoranzo:
```suggestion
        if self.includes_datatypes or self.includes_data_managers:
```

lib/galaxy/model/tool_shed_install/__init__.py
```diff
@@ -136,7 +136,7 @@ def guess_shed_config(self, app):
                             if tool_id in tool_ids:
                                 self.shed_config_filename = name
                                 return shed_tool_conf_dict
-        if self.includes_datatypes or self.includes_data_managers or not default:
+        if self.includes_datatypes or self.includes_data_managers:
```
mvdbeek: damn, that wasn't right,  we should do this unconditionally now

mvdbeek:
I think this is ready for review. I've tested this manually and I think this works ok.

mvdbeek:
OK, wrote some integration tests and discovered that we're now installing into using tool_conf.xml by default, which is not as intended. Though I think that was pre-existing.

mvdbeek:
Alright, now with an integration test case that does install/uninstall and update.

mvdbeek:
The failing integration test is unrelated, the testtoolshed has a problem serving https://testtoolshed.g2.bx.psu.edu/api/repositories/get_repository_revision_install_info?name=data_manager_sam_fasta_index_builder&owner=devteam&changeset_revision=406896e00d0e

PR #8962, opened 2019-11-11
Create packages for Galaxy tests and infrastructure.

jmchilton:
Move the tool shed tests into ``lib/tool_shed`` so they will be live with that code and are isolated from Galaxy tests.

Create new packages:
- ``galaxy-test-base``: Galaxy testing utilities that don't depend on an in-memory Galaxy application or ``galaxy-app``  (as described in https://github.com/galaxyproject/galaxy/pull/8960). Suitable for building test cases against remote Galaxy instances.
- ``galaxy-test-driver``: Galaxy testing utilities that depend on Galaxy internals and that construct Galaxy objects (as described in https://github.com/galaxyproject/galaxy/pull/8960).
- ``galaxy-test-api``: The Galaxy API tests - relocated into ``lib/galaxy_test/api``.
- ``galaxy-test-selenium``: The Galaxy API tests - relocated into ``lib/galaxy_test/selenium``.

The unit and integration tests aren't packaged because I don't know there would be any value in doing this. They depend on Galaxy's internals and can't really be run on external instances - I'm not sure why anyone would ever pip install them. Might be worth moving them into ``lib/galaxy_test`` for consistency and if we decide they belong in a package - I think they would be fine ``galaxy-test-driver``.

Once this and #8949 are merged, we can migrate the remaining bits of ``galaxy-test-driver`` that depend on the tool shed out and into the tool shed code. Eliminating even more ``import tool_shed`` code from Galaxy code files. A future tool shed package can just have a dev dependency on the new ``galaxy-test-driver`` package to leverage most the framework that powers the tool shed tests.


lib/galaxy_test/api/_framework.py
```diff
@@ -0,0 +1,5 @@
+# TODO: implement a remote-only version of ApiTestCase if galaxy_test.driver isn't
+# available. There is not reason a vast majority of these tests depend on Galaxy internals.
```
mvdbeek:
```suggestion
# available. Most of these tests do not depend on Galaxy internals.
```

mvdbeek (approved):
Awesome, very cool!

PR #8964, opened 2019-11-12
Add new GROMACS datatypes: ndx, xvg, edr

simonbray:


simonbray:
ping @tsenapathi, in case you want to use these in your tools

lib/galaxy/datatypes/binary.py
```diff
@@ -991,7 +991,24 @@ class Xtc(GmxBinary):
     """
 
     file_ext = "xtc"
-    magic_number = 1995  # reference: https://github.com/gromacs/gromacs/blob/cec211b2c835ba6e8ea849fb1bf67d7fc19693a4/src/gromacs/fileio/xtcio.cpp
+    magic_number = 1995  # reference: https://github.com/gromacs/gromacs/blob/master/src/gromacs/fileio/xtcio.cpp
```
mvdbeek: I'd prefer if the references remain tied to a specific commit, that should allow us to go back even if the file location changes in the future.

mvdbeek (approved):

lib/galaxy/datatypes/binary.py
simonbray: Sure, makes sense

PR #8970, opened 2019-11-12
[19.05] Backport #8953 grouping strip newline

bernt-matthias:


mvdbeek (approved):

tools/stats/grouping.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="Grouping1" name="Group" version="2.1.3">
+<tool id="Grouping1" name="Group" version="2.1.4">
```
jmchilton, line 1: I dislike breaking workflows on an existing widely used version - can we not just target 19.09? I'm fine if you insist it is important - I'm just nervous.

bernt-matthias: I don't insist (I can cherry-pick it anyway on my local 19.05 instance). Just feels odd to leave the buggy tool (it will fail always if grouping on the last column).

The problem with the built-in tools seems that having multiple versions of the same tool seems not possible even if they are versioned, or am I wrong?

mvdbeek: Yep, that's why we migrated out a large portion of tools that used to ship with Galaxy. If it's larger changes it is usually better to write a new tools and upload it to the toolshed. In this case I don't think that workflows will really "break", they'll just show a warning sign ... which I think is better than potentially getting a wrong result.

jmchilton (approved):

PR #8972, opened 2019-11-12
Add process ID to worker_process table

selten:


lib/galaxy/model/database_heartbeat.py
```diff
@@ -79,7 +80,7 @@ def worker_process(self):
     def update_watcher_designation(self):
         worker_process = self.worker_process
         if not worker_process:
-            worker_process = WorkerProcess(server_name=self.server_name, hostname=self.hostname)
+            worker_process = WorkerProcess(server_name=self.server_name, hostname=self.hostname, pid=os.getpid())
```
mvdbeek: We also need some logic that updates worker pids, this will only do it for new workers (meaning a new server_name/hostname combination). Maybe add a pid attribute to the DatabaseHeartbeat. Since this needs to happen after forking I'd set this to None in `__init__` and only set the pid in `start `. Then you can drop the new argument to `WorkerProcess` on this line and just do `worker_process.pid = self.pid`. I think that should work.

lib/galaxy/model/migrate/versions/0161_worker_process_pid.py
```diff
@@ -0,0 +1,38 @@
...
+from __future__ import print_function
+
+import logging
+
+from sqlalchemy import (
+    Column,
+    MetaData
+)
+
+from galaxy.model.custom_types import UUIDType
```
mvdbeek: This isn't used, but you need to import the sqlachemy Integer type and use that instead of the python int.
```suggestion
```

lib/galaxy/model/migrate/versions/0161_worker_process_pid.py
```diff
@@ -0,0 +1,38 @@
...
+Adds `pid` column to worker_process table.
+"""
+
+from __future__ import print_function
+
+import logging
+
+from sqlalchemy import (
+    Column,
+    MetaData
```
mvdbeek:
```suggestion
    Integer
    MetaData
```

lib/galaxy/model/migrate/versions/0161_worker_process_pid.py
```diff
@@ -0,0 +1,38 @@
...
+log = logging.getLogger(__name__)
+
+
+def upgrade(migrate_engine):
+    print(__doc__)
+    metadata = MetaData()
+    metadata.bind = migrate_engine
+    metadata.reflect()
+
+    pid_column = Column('pid', int)
```
mvdbeek:
```suggestion
    pid_column = Column('pid', Integer)
```

lib/galaxy/model/database_heartbeat.py
```diff
@@ -1,3 +1,4 @@
+import os
```
mvdbeek: The linter complains about the order here, this should be on line 4

mvdbeek (approved):

mvdbeek:
Thanks @selten!

PR #8978, opened 2019-11-13
Remove workflow support from tool shed.

jmchilton:
Was removed from the Galaxy client already.


lib/tool_shed/webapp/templates/webapps/tool_shed/repository/common.mako
```diff
@@ -299,12 +299,6 @@
                     else:
                         folder_label = "%s<i> - repository tools require handling of these dependencies</i>" % folder_label
                     col_span_str = 'colspan="4"'
-                elif folder.workflows:
```
mvdbeek, line 302: This is still exposed and functional when browsing for instance https://toolshed.g2.bx.psu.edu/repository/view_workflow?render_repository_actions_for=tool_shed&workflow_name=facfac2e74737e720e1402394d9542cdb37a905a%3A64656570546f6f6c732042414d205143&repository_metadata_id=cc20e5d619d52d98

That said I'm OK to deprecate this.

mvdbeek (approved):

PR #8980, opened 2019-11-13
[19.05] Fix test repository self reference URL broken in dev.

jmchilton:
Targeting a branch without the change instead of dev should be more stable anyway.


mvdbeek (approved):
Thanks!

PR #8988, opened 2019-11-13
Remove admin tool shed controller code (Part 1)

jmchilton:
Clean up some low hanging fruit here - removing completely unused code and removing code unused after replacing tool shed test functions with API calls.


martenson (approved):
solid step, thanks!

PR #8990, opened 2019-11-13
Assorted fixes for extended metadata handling.

jmchilton:
Fixes for esoteric tool feature edge cases I promised in #8930. Re-enable the 3 failing tests.

PR #8991, opened 2019-11-14
More fixes to extended_metadata when using outputs_to_working_directory

mvdbeek:
Builds on https://github.com/galaxyproject/galaxy/pull/8990. One noteworthy improvement is that we now always set `false_path` (the path that datasets get rewritten to/from in the job working directory) to `galaxy_dataset_<uuid>.dat` instead of `galaxy_dataset_<dataset_id>.dat`


test/integration/test_extended_metadata_outputs_to_working_directory.py
```diff
@@ -0,0 +1,23 @@
+"""Integration tests for the Pulsar embedded runner."""
+
+from galaxy_test.driver import integration_util
+from .test_extended_metadata import (
+    ExtendedMetadataIntegrationInstance,
+    TEST_TOOL_IDS
+)
+
+
+class ExtendedMetadataOutputsToWorkingDirIntegrationInstance(ExtendedMetadataIntegrationInstance):
```
jmchilton, line 10: ❤️

martenson: hot damn

PR #8995, opened 2019-11-15
Make mol2 visible

bgruening:
as requested by a user


nsoranzo (approved):

mvdbeek:
@bgruening you pushed to the galaxyproject namespace. Since that happened to me too I set `git config remote.origin.pushurl "you really didn't want to do that"` which will not let you do that.

PR #8998, opened 2019-11-15
[WIP][19.09] New outputs/ directory needs to be mounted rw.

jmchilton:
Added in ad573b215802aaaa9d09e1d2e3e2974e79295dfe.

### DIFF
--- a/lib/galaxy/tool_util/deps/container_classes.py
+++ b/lib/galaxy/tool_util/deps/container_classes.py
@@ -178,7 +178,7 @@ class HasDockerLikeVolumes(object):
             if self.job_info.tool_directory:
                 defaults += ",$tool_directory:default_ro"
             if self.job_info.job_directory:
-                defaults += ",$job_directory:default_ro"
+                defaults += ",$job_directory:default_ro,$job_directory/outputs:rw"
             if self.job_info.tmp_directory is not None:
                 defaults += ",$tmp_directory:rw"
             if self.app_info.outputs_to_working_directory:


mvdbeek (dismissed):

lib/galaxy/tool_util/deps/container_classes.py
```diff
@@ -178,7 +178,7 @@ def add_var(name, value):
             if self.job_info.tool_directory:
                 defaults += ",$tool_directory:default_ro"
             if self.job_info.job_directory:
-                defaults += ",$job_directory:default_ro"
+                defaults += ",$job_directory:default_ro,$job_directory/outputs:rw"
```
mvdbeek, line 181: So do singularity and docker handle that correctly ? I think there's a change they may not :/

jmchilton: The working directory uses the same trick though right? Everything is broken if this doesn't work 😿.

mvdbeek: I have to admit that I cheated when I deployed singularity and just mounted everything rw, but I can check

jmchilton: I think Singularity is a known quantity and we work around it?

> Galaxy allows the mount try "default_ro" which translates to ro for Docker and ro for Singularity iff no subdirectories are rw (Singularity does not allow ro parent directories with rw subdirectories).

mvdbeek: ok, I think that's fine then.

jmchilton:
We didn't add this feature in 19.09 :( , thanks for hunting that down @mvdbeek.

PR #8999, opened 2019-11-15
[19.09] Fix unspecified shed_tool_config_file taking precedence over old shed_tool_conf.xml

mvdbeek:
This fixes https://github.com/galaxyproject/galaxy/issues/8952#issuecomment-554346736, but this was very hard to diagnose and I also fixed some inconsistencies along the way.

With https://github.com/galaxyproject/galaxy/commit/be5aafd34b1f400704ef863684c9b742e1ca7427 Galaxy now creates/loads a default shed_tool_config_file if it is not part of the tool_config setting. This led to automated installations going to the wrong shed_tool_conf .xml file, one that was never referenced in any config.

The way this was done was also slightly wrong, the actual app.config.tool_config file was never updated, and so any upon toolbox reload would lose this file (xref https://github.com/galaxyproject/galaxy/blob/406e77e05c3d35239b0ecc13f980354187bc09fa/lib/galaxy/queue_worker.py#L186) ... but it would remain in the integrated_tool_panel_conf.xml file

So this PR only loads/creates the default, unspecified shed_tool_config_file if no shed-enabled tool config file has been loaded, and also fixes the way this is done (needs to be added to self.config.tool_configs for reloads).

This PR also unifies the methods for getting a shed_tool_config_dict, fixes or enhances some logging messages and raises appropriate exceptions if no shed_enabled config is loaded and repository installations are attempted and removes now unreachable code.

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -1109,7 +1109,7 @@ class ConfiguresGalaxyMixin(object):
                 log.warning(
                     "The default shed tool config file (%s) has been added to the tool_config_file option, if this is "
                     "not the desired behavior, please set shed_tool_config_file to your primary shed-enabled tool "
-                    "config file"
+                    "config file", self.config.shed_tool_config_file
                 )
             tool_configs.append(self.config.shed_tool_config_file)
         # The value of migrated_tools_config is the file reserved for containing only those tools that have been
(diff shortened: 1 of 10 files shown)


mvdbeek:
The failed test is unrelated, the TTS has some problems atm.

natefoo (approved):
Thank you for cleaning up my mess, @mvdbeek!

mvdbeek:
Alright, with the last commit the tests are passing as well!

bgruening:
:) ok, I will run everything again as soon as I have time. Gracias!

PR #9001, opened 2019-11-16
fix history import and export for real user setups

bernt-matthias:
Adds necessary calls to the `external_chown_script` for files and directories that need to be readable / writable by the user running the job. 

Essentially ready, but I guess some beautification could be nice (e.g., maybe marge the _chown function somehow with the chown of the job class) -- but I prefer to wait first for reviews.

@mvdbeek suggested that an integration test would be nice, but this seems out of reach for me.

Below in one of the comments I collected further questions / todos.

Note that I tested on import/export on usegalaxy.eu .. and it seemed to fail as well.

### DIFF
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -57,6 +57,7 @@ from galaxy.tool_util.output_checker import (
 )
 from galaxy.util import (
     parse_xml_string,
+    RWXRWXRWX,
     safe_makedirs,
     unicodify
 )
@@ -2207,9 +2208,11 @@ class JobWrapper(HasResourceParameters):
     def change_ownership_for_run(self):
         job = self.get_job()
         external_chown_script = self.get_destination_configuration("external_chown_script", None)
-        if job.user is not None:
-            external_chown(self.working_directory, self.user_system_pwent,
+        if job.user is not None and external_chown_script is not None:
+            ret = external_chown(self.working_directory, self.user_system_pwent,
                            external_chown_script, description="working directory")
+            if not ret:
+                os.chmod(self.working_directory, RWXRWXRWX)
 
     def reclaim_ownership(self):
         job = self.get_job()
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1930,9 +1930,6 @@ class JobWrapper(HasResourceParameters):
         for (hda, dataset_path) in self.output_hdas_and_paths.values():
             if hda == dataset:
                 return dataset_path
-#         if getattr(dataset, "fake_dataset_association", False):
-#             log.error("GET_OUTPUT_PATH FAKE %s" %(dataset.file_name))
-#             return dataset.file_name
         raise KeyError("Couldn't find job output for [{}] in [{}]".format(dataset, self.output_hdas_and_paths.values()))
 
     def get_mutable_output_fnames(self):
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -8,7 +8,6 @@ import json
 import logging
 import os
 import pwd
-import shlex
 import shutil
 import string
 import sys
@@ -65,6 +64,7 @@ from galaxy.util import (
 )
 from galaxy.util.bunch import Bunch
 from galaxy.util.expressions import ExpressionContext
+from galaxy.util.path import external_chown
 from galaxy.util.xml_macros import load
 from galaxy.web_stack.handlers import ConfiguresHandlers
 
@@ -2207,33 +2207,19 @@ class JobWrapper(HasResourceParameters):
             # The tool is unavailable, we try to move the outputs.
             return False
 
-    def _change_ownership(self, username, gid):
-        job = self.get_job()
-        external_chown_script = self.get_destination_configuration("external_chown_script", None)
-        if external_chown_script is not None:
-            cmd = shlex.split(external_chown_script)
-            cmd.extend([self.working_directory, username, str(gid)])
-            log.debug('({}) Changing ownership of working directory with: {}'.format(job.id, ' '.join(cmd)))
-            try:
-                commands.execute(cmd)
-            except commands.CommandLineException as e:
-                log.error('external script failed: %s' % unicodify(e))
-
     def change_ownership_for_run(self):
         job = self.get_job()
         external_chown_script = self.get_destination_configuration("external_chown_script", None)
-        if external_chown_script and job.user is not None:
-            try:
-                self._change_ownership(self.user_system_pwent[0], str(self.user_system_pwent[3]))
-            except Exception:
-                log.exception('({}) Failed to change ownership of {}, making world-writable instead'.format(job.id, self.working_directory))
-                os.chmod(self.working_directory, RWXRWXRWX)
+        if job.user is not None:
+            external_chown(self.working_directory, self.user_system_pwent,
+                           external_chown_script, description="working directory")
 
     def reclaim_ownership(self):
         job = self.get_job()
         external_chown_script = self.get_destination_configuration("external_chown_script", None)
-        if external_chown_script and job.user is not None:
-            self._change_ownership(self.galaxy_system_pwent[0], str(self.galaxy_system_pwent[3]))
+        if job.user is not None:
+            external_chown(self.working_directory, self.galaxy_system_pwent,
+                           external_chown_script, description="working directory")
 
     @property
     def user_system_pwent(self):
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1819,7 +1819,6 @@ class JobWrapper(HasResourceParameters):
                         if e.errno != errno.ENOENT:
                             raise
                 self.external_output_metadata.cleanup_external_metadata(self.sa_session)
-            galaxy.tools.imp_exp.JobExportHistoryArchiveWrapper(self.app, self.job_id).cleanup_after_job()
             galaxy.tools.imp_exp.JobImportHistoryArchiveWrapper(self.app, self.job_id).cleanup_after_job()
             if delete_files:
                 self.object_store.delete(self.get_job(), base_dir='job_work', entire_dir=True, dir_only=True, obj_dir=True)
@@ -1965,7 +1964,7 @@ class JobWrapper(HasResourceParameters):
         self.output_paths = [t[2] for t in results]
         self.output_hdas_and_paths = {t[0]: t[1:] for t in results}
         if special:
-            false_path = dataset_path_rewriter.rewrite_dataset_path(special.dataset, 'output')
+            false_path = dataset_path_rewriter.rewrite_dataset_path(special, 'output')
             dsp = DatasetPath(special.dataset.id, special.dataset.file_name, false_path)
             self.output_paths.append(dsp)
             self.output_hdas_and_paths["output_file"] = [special.fda, dsp]
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1922,26 +1922,32 @@ class JobWrapper(HasResourceParameters):
         return [os.path.basename(str(fname)) for fname in self.get_output_fnames()]
 
     def get_output_fnames(self):
+        log.error("GET_OUTPUT_FNAMES")
         if self.output_paths is None:
             self.compute_outputs()
         return self.output_paths
 
     def get_output_path(self, dataset):
+        log.error("GET_OUTPUT_PATH")
         if self.output_paths is None:
             self.compute_outputs()
         for (hda, dataset_path) in self.output_hdas_and_paths.values():
+            log.error("GET_OUTPUT_PATH %s %s" %(hda,dataset))
             if hda == dataset:
                 return dataset_path
-        if getattr(dataset, "fake_dataset_association", False):
-            return dataset.file_name
+#         if getattr(dataset, "fake_dataset_association", False):
+#             log.error("GET_OUTPUT_PATH FAKE %s" %(dataset.file_name))
+#             return dataset.file_name
         raise KeyError("Couldn't find job output for [{}] in [{}]".format(dataset, self.output_hdas_and_paths.values()))
 
     def get_mutable_output_fnames(self):
+        log.error("GET_MUTABLE_OUTPUT_FNAME")
         if self.output_paths is None:
             self.compute_outputs()
         return [dsp for dsp in self.output_paths if dsp.mutable]
 
     def get_output_hdas_and_fnames(self):
+        log.error("GET_OUTPUT_HDAS_AND_FNAMES")
         if self.output_hdas_and_paths is None:
             self.compute_outputs()
         return self.output_hdas_and_paths
@@ -1950,6 +1956,8 @@ class JobWrapper(HasResourceParameters):
         dataset_path_rewriter = self.dataset_path_rewriter
 
         job = self.get_job()
+        log.error("JOB %s " %(job))
+        log.error("JOB DIR %s " %(dir(job)))
         # Job output datasets are combination of history, library, and jeha datasets.
         special = self.sa_session.query(model.JobExportHistoryArchive).filter_by(job=job).first()
         false_path = None
@@ -1957,6 +1965,8 @@ class JobWrapper(HasResourceParameters):
         results = []
         for da in job.output_datasets + job.output_library_datasets:
             da_false_path = dataset_path_rewriter.rewrite_dataset_path(da.dataset, 'output')
+            log.error("NORMAL da: %s" %(da))
+            log.error("NORMAL da.dataset %s -> da_false_path %s" %(da.dataset, da_false_path))
             mutable = da.dataset.dataset.external_filename is None
             dataset_path = DatasetPath(da.dataset.dataset.id, da.dataset.file_name, false_path=da_false_path, mutable=mutable)
             results.append((da.name, da.dataset, dataset_path))
@@ -1964,9 +1974,18 @@ class JobWrapper(HasResourceParameters):
         self.output_paths = [t[2] for t in results]
         self.output_hdas_and_paths = {t[0]: t[1:] for t in results}
         if special:
-            false_path = dataset_path_rewriter.rewrite_dataset_path(special, 'output')
+            false_path = dataset_path_rewriter.rewrite_dataset_path(special.dataset, 'output')
+            log.error("SPECIAL special %s" %(special))
+            log.error("SPECIAL special.dataset %s -> false_path %s" %(special.dataset, false_path))
             dsp = DatasetPath(special.dataset.id, special.dataset.file_name, false_path)
             self.output_paths.append(dsp)
+            log.error("SPECIAL DIR %s"%str(dir(special)))
+            log.error("SPECIAL JOB %s " %(special.job))
+            log.error("SPECIAL NAME %s"%str(special.export_name))
+            log.error("SPECIAL ... %s"%str(special.__dict__))
+
+            self.output_hdas_and_paths[ "output_file" ] = [special.fda, dsp]
+        log.error( "self.output_hdas_and_paths %s"% str(self.output_hdas_and_paths) )
         return self.output_paths
 
     def get_output_file_id(self, file):
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -56,9 +56,7 @@ from galaxy.tool_util.output_checker import (
     DETECTED_JOB_STATE,
 )
 from galaxy.util import (
-    commands,
     parse_xml_string,
-    RWXRWXRWX,
     safe_makedirs,
     unicodify,
 )
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -58,7 +58,7 @@ from galaxy.tool_util.output_checker import (
 from galaxy.util import (
     parse_xml_string,
     safe_makedirs,
-    unicodify,
+    unicodify
 )
 from galaxy.util.bunch import Bunch
 from galaxy.util.expressions import ExpressionContext
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1922,17 +1922,14 @@ class JobWrapper(HasResourceParameters):
         return [os.path.basename(str(fname)) for fname in self.get_output_fnames()]
 
     def get_output_fnames(self):
-        log.error("GET_OUTPUT_FNAMES")
         if self.output_paths is None:
             self.compute_outputs()
         return self.output_paths
 
     def get_output_path(self, dataset):
-        log.error("GET_OUTPUT_PATH")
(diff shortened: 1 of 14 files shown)


bernt-matthias:
Currently stuck at an apparently trivial problem: 

I modified the JobExportHistoryArchive to contain the FakeDatasetAssociation:

https://github.com/galaxyproject/galaxy/blob/b6d4f70298053fb09417bac78f359453d1d7ced3/lib/galaxy/model/__init__.py#L1442

which is initialised here

https://github.com/galaxyproject/galaxy/blob/b6d4f70298053fb09417bac78f359453d1d7ced3/lib/galaxy/tools/actions/history_imp_exp.py#L119

the log.error on the foillowing line outputs something like: 

```
JobExportHistoryArchive <galaxy.model.JobExportHistoryArchive(None) at 0x2b6a0ec8e850> created with fda <galaxy.model.FakeDatasetAssociation object at 0x2b6a0eb4e590>
```

Note the `None` 

If I try to use this object 

https://github.com/galaxyproject/galaxy/blob/b6d4f70298053fb09417bac78f359453d1d7ced3/lib/galaxy/tools/evaluation.py#L89

https://github.com/galaxyproject/galaxy/blob/b6d4f70298053fb09417bac78f359453d1d7ced3/lib/galaxy/jobs/__init__.py#L1954

I always get: `AttributeError: 'JobExportHistoryArchive' object has no attribute 'fda'`

Strange is also that if I print the object I get special `<galaxy.model.JobExportHistoryArchive(55) at 0x2b6a4d59ac50>`  I.e. no `None` and a seemingly different object.

I wiped out all `pyc` files and restarted Galaxy...

bernt-matthias:
TODOs: 

- [ ] target 19.05, 19.09 or dev 
- [ ] Check if `model.GenomeIndexToolData` also needs special care everywhere where `JobExportHistoryArchive` does  (e.g. add `FakeDataAssociation`). See here:
https://github.com/galaxyproject/galaxy/blob/b6d4f70298053fb09417bac78f359453d1d7ced3/lib/galaxy/jobs/__init__.py#L1067
- [x] Since there is now a HDA changes to the cleanup script might be necessary.
- [ ] Do both of these lines need to be called? https://github.com/galaxyproject/galaxy/blob/c2580727060583462abc76acec00577090c16da0/lib/galaxy/jobs/__init__.py#L1789
- [ ] hardcoding of "output_file" seems quite dirty (export case)
- [ ] currently neither hidden nor deleted data sets are included, do we want to change this?

bernt-matthias:
@mvdbeek your commit seems to solve the problem ... which brings me back to this https://github.com/galaxyproject/galaxy/pull/8567 .. LOL

My latest commit (https://github.com/galaxyproject/galaxy/pull/9001/commits/a69c608e51df146a1d64d1499a3fb059698136af) is an unsuccessful test to get the jobs working dir which results in an error: 

```
sqlalchemy.exc.ArgumentError: Object <galaxy.model.Job(9504) at 0x2ba689378510> is not legal as a SQL literal value
```

I guess there is no job yet.

bernt-matthias:
Last commit is a better alternative for https://github.com/galaxyproject/galaxy/pull/8567.

I guess some refactoring of the call to the external chown function could be nice. I will try to implement something in util on which everything else could depend.

Which no brings me to an error during the import:

```
galaxy.jobs ERROR 2019-11-17 17:01:38,647 [p:189723,w:0,m:1] [UnivaJobRunner.work_thread-3] Unable to cleanup job 9524
  File "lib/galaxy/jobs/__init__.py", line 1790, in cleanup
    galaxy.tools.imp_exp.JobImportHistoryArchiveWrapper(self.app, self.job_id).cleanup_after_job()
  File "lib/galaxy/tools/imp_exp/__init__.py", line 53, in cleanup_after_job
    model_store.perform_import(new_history, job=job, new_history=True)
  File "lib/galaxy/model/store/__init__.py", line 168, in perform_import
    datasets_attrs = self.datasets_properties()
  File "lib/galaxy/model/store/__init__.py", line 772, in datasets_properties
    datasets_attrs = load(open(datasets_attrs_file_name))
IOError: [Errno 2] No such file or directory: u'/gpfs1/work/songalax/galaxy-dev/database/tmp/tmpoHvvkM/datasets_attrs.txt'
```

LOL. But I guess this can also be solved with chown. But I need to investigate this.

mvdbeek:
> * target 19.05, 19.09 or dev

dev first, we can then see what parts (or perhaps all) we want to backport.

> * Check if `model.GenomeIndexToolData` also needs special care everywhere where `JobExportHistoryArchive` does  (e.g. add `FakeDataAssociation`). See here:

Keep it simple, that can be done in another PR

> * Since there is now a HDA changes to the cleanup script might be necessary.

I don't see a new HDA here ?

> * Do both of these lines need to be called? 
>     
>       
>         [galaxy/lib/galaxy/jobs/__init__.py](https://github.com/galaxyproject/galaxy/blob/c2580727060583462abc76acec00577090c16da0/lib/galaxy/jobs/__init__.py#L1789)

Seems unrelated to the PR ? Smaller, isolated changes are often better.

> * hardcoding of "output_file" seems quite dirty (export case)

You can discover the output from the tool that is associated to the job. But that might be unnecessary complexity so long as the export tools is the only tool that falls into the `special` case. Just mark it as a constant (uppercase variable) is enough IMO.  

> * [ ]  currently neither hidden nor deleted data sets are included, do we want to change this?

Perhaps in another PR.

lib/galaxy/tools/imp_exp/__init__.py
```diff
@@ -1,17 +1,34 @@
...
-from galaxy.model import store
+from galaxy.util import unicodify
 from galaxy.version import VERSION_MAJOR
 
 log = logging.getLogger(__name__)
 
 ATTRS_FILENAME_HISTORY = 'history_attrs.txt'
 
 
+def _chown(path, jeha, app, user):
```
mvdbeek: don't we have a function like that already somewhere in the codebase ? could we re-use it ?

lib/galaxy/model/__init__.py
```diff
@@ -1422,6 +1422,22 @@ def dataset(self):
         return None
 
 
+# Set up output dataset association for export history jobs. Because job
+# uses a Dataset rather than an HDA or LDA, it's necessary to set up a
+# fake dataset association that provides the needed attributes for
+# preparing a job.
+class FakeDatasetAssociation (object):
```
mvdbeek, line 1567: You can use this one in https://github.com/galaxyproject/galaxy/blob/972d641b1abec0117de812f0200eb3bad81f1822/lib/galaxy/tools/evaluation.py#L89 now to avoid duplication

lib/galaxy/tools/imp_exp/__init__.py
bernt-matthias: Yeah, at least here: https://github.com/galaxyproject/galaxy/blob/b91a205f89c906ce062abc6ae5bb7b5ac73ebfa8/lib/galaxy/tools/actions/upload_common.py#L320 and here https://github.com/galaxyproject/galaxy/blob/9662164ad68b39adf5a5606a7aa8e388f6a79f1e/lib/galaxy/jobs/__init__.py#L2169 these are currently bound to a class which we can't access from here as far as I see.

I was thinking about introducing a new function (doing popen, and error handling, and maybe pwent stuff) in util that could be reused. Would this be a good idea?

lib/galaxy/model/__init__.py
bernt-matthias: The one in galaxy/lib/galaxy/tools/evaluation.py should be already gone, isn't it?

mvdbeek: yup, I'm just blind ... :)

lib/galaxy/tools/imp_exp/__init__.py
mvdbeek: Yes, I think that would be great.

bernt-matthias: OK. Function added. Need to test on my setup. 

Will introduce this function at the other places later in a separate PR.

Currently the code is missing to call the chown also for archive_source in the import case. Would be nice, but I do not have an idea how to access the path here: https://github.com/galaxyproject/galaxy/blob/c2580727060583462abc76acec00577090c16da0/lib/galaxy/jobs/__init__.py#L1789

lib/galaxy/tools/imp_exp/__init__.py
```diff
@@ -1,10 +1,13 @@
+import getpass
 import logging
 import os
+import shlex
 import shutil
+import subprocess
 import tempfile
 
 from galaxy import model
-from galaxy.model import store
```
mvdbeek, line 7: I think this was done on purpose to ease refactoring, maybe keep it ?

lib/galaxy/tools/imp_exp/__init__.py
```diff
@@ -40,15 +49,15 @@ def cleanup_after_job(self):
         new_history = None
         try:
             archive_dir = jiha.archive_dir
-            model_store = store.get_import_model_store_for_directory(archive_dir, app=self.app, user=user)
+            external_chown(archive_dir, jiha.job.user.system_user_pwent(str(getpass.getuser())),
```
mvdbeek:
```suggestion
            external_chown(archive_dir, jiha.job.user.system_user_pwent(getpass.getuser()),
```
this is already a string, on python2 and python3.

lib/galaxy/tools/imp_exp/__init__.py
```diff
@@ -108,9 +119,12 @@ def cleanup_after_job(self):
...
         jeha = self.sa_session.query(model.JobExportHistoryArchive).filter_by(job_id=self.job_id).first()
-        if jeha:
-            temp_dir = jeha.temp_directory
-            try:
-                shutil.rmtree(temp_dir)
-            except Exception as e:
-                log.debug('Error deleting directory containing attribute files (%s): %s' % (temp_dir, e))
+        if not jeha:
+            return
+        external_chown(jeha.temp_directory, jeha.job.user.system_user_pwent(str(getpass.getuser())),
```
mvdbeek:
```suggestion
        external_chown(jeha.temp_directory, jeha.job.user.system_user_pwent(getpass.getuser()),
```

lib/galaxy/tools/imp_exp/__init__.py
mvdbeek: Not sure about the import archive_source, that would probably be handled somewhere during upload ?

Could you also use the new function in this PR as well ? This avoids having slightly divergent functions scattered in the code, and is also a good way to figure out if anything is missing in the new function (which looks good!).

lib/galaxy/tools/imp_exp/__init__.py
bernt-matthias: done

bernt-matthias:
Now the new function is used everywhere. Lets see what the tests say: for which parts (jobs, upload,...) do we have tests using a real user setup? 

Note that I removed the chmod 777 in case of a failure in chown of the job dir (I think this should only happen in case of misconfigurations)

bernt-matthias:
Regarding archive_source. I think the last trace is here: https://github.com/galaxyproject/galaxy/blob/27c6c71214af3c1ff9e6c7c093bbc1cb3096eb4f/lib/galaxy/tools/actions/history_imp_exp.py#L54

in the following lines the incoming dict is somehow added to the job. Wondering if it can retrieved from within JobImportHistoryArchiveWrapper....

bernt-matthias:
Not sure how to fix the unit tests. 

Seems that in the tests for the  `jiha` which is set here:

https://github.com/galaxyproject/galaxy/blob/c6062abe5768a5460e6cc57e477121a8d127bae5/lib/galaxy/tools/imp_exp/__init__.py#L43

`jiha.job` has a None user attribute.

Also just adding a `model.User(...)`  to the job here https://github.com/galaxyproject/galaxy/blob/adaf0fe31c2afd4f8405a950c72fdea8db1e34b0/test/unit/tools/test_history_imp_exp.py#L32

does not help (odd anyway that model.Job has no user attribute).

bernt-matthias:
integration and selenium failures seem unrelated.

mvdbeek:
@bernt-matthias bumping this to 20.05, but let me know if I can do something to help. Given this is a bug I'm happy to re-target after branching. I do think this fixes history import already for `outputs_to_working_directory`, which seems like a pretty nice fix on its own.

bernt-matthias:
@mvdbeek agreed. Missing is a test for these cases, e.g. for  `outputs_to_working_directory` and real user setups.

mvdbeek:
I can create an integration test for outputs_to_working_directory, the real user setup seems a little trickier.

bernt-matthias:
Hey @mvdbeek seems that this is it :) api and integration tests pass. The failing tests are unrelated, or?

bernt-matthias:
Would like to squash a few more commits. Would make it easier for me to cherry pick the commits to my release branch.

lib/galaxy/jobs/__init__.py
```diff
@@ -2199,37 +2196,19 @@ def __link_file_check(self):
...
         external_chown_script = self.get_destination_configuration("external_chown_script", None)
-        if external_chown_script and job.user is not None:
-            try:
-                self._change_ownership(self.user_system_pwent[0], str(self.user_system_pwent[3]))
-            except Exception:
-                log.exception('(%s) Failed to change ownership of %s, making world-writable instead' % (job.id, self.working_directory))
-                os.chmod(self.working_directory, RWXRWXRWX)
+        if job.user is not None:
+            external_chown(self.working_directory, self.user_system_pwent,
+                           external_chown_script, description="working directory")
```
bernt-matthias, line 2210: Important question: Is it OK not to `os.chmod(self.working_directory, RWXRWXRWX)` in case of error? Or should this be integrated in `external_chown` or just here?

bernt-matthias:
Seems that this PR is more ready than ever. Tests are finally passing. Final review would be welcome. If wanted I can also test on our installation.

lib/galaxy/jobs/__init__.py
mvdbeek: Is this change critical for fixing hisitory import/export ? If not I would keep the current behavior and change it in another PR.

lib/galaxy/tools/imp_exp/export_history.py
```diff
@@ -26,6 +26,9 @@ def create_archive(export_directory, out_file, gzip=False):
     except Exception as e:
         print('Error creating history archive: %s' % unicodify(e), file=sys.stderr)
         return 1
+    finally:
+        if os.path.exists(export_directory):
+            shutil.rmtree(export_directory, ignore_errors=True)
```
mvdbeek: You don't need that with `ignore_errors=True`
```suggestion
        if os.path.exists(export_directory):
```
```suggestion
        shutil.rmtree(export_directory, ignore_errors=True)
```

lib/galaxy/util/path/__init__.py
```diff
@@ -343,6 +345,26 @@ def canonicalize(self, ext):
...
+def external_chown(path, pwent, external_chown_script, description="file"):
+    """
+    call the external chown script (if not None) to change
+    the user and group of the given path, and additional description
+    of the file/path for the log message can be given
+    """
+    if external_chown_script is None:
+        return
+
+    try:
```
mvdbeek, line 358: You could use commands.execute here, right? (https://github.com/galaxyproject/galaxy/pull/9544)

mvdbeek (commented):
Nice, almost there!

mvdbeek:
Some of the selenium errors look legitimate:
```
galaxy.tools: ERROR: Exception caught while attempting to execute tool with id 'upload1':
  File "/galaxy/lib/galaxy/tools/__init__.py", line 1601, in handle_single_execution
    collection_info=collection_info,
  File "/galaxy/lib/galaxy/tools/__init__.py", line 1685, in execute
    return self.tool_action.execute(self, trans, incoming=incoming, set_output_hid=set_output_hid, history=history, **kwargs)
  File "/galaxy/lib/galaxy/tools/actions/upload.py", line 29, in execute
    rval = self._setup_job(tool, trans, incoming, dataset_upload_inputs, history)
  File "/galaxy/lib/galaxy/tools/actions/upload.py", line 52, in _setup_job
    json_file_path = upload_common.create_paramfile(trans, uploaded_datasets)
  File "/galaxy/lib/galaxy/tools/actions/upload_common.py", line 384, in create_paramfile
    trans.user.system_user_pwent(trans.app.config.real_system_username),
AttributeError: 'NoneType' object has no attribute 'system_user_pwent'
galaxy.tools.execute: WARNING: There was a failure executing a job for tool [upload1] - Error executing tool with id 'upload1': 'NoneType' object has no attribute 'system_user_pwent'
galaxy.tools.execute: DEBUG: Executed 1 job(s) for tool upload1 request (63.236 ms)
```

lib/galaxy/tools/actions/upload_common.py
```diff
@@ -396,8 +377,10 @@ def _chown(path):
             # TODO: This will have to change when we start bundling inputs.
             # Also, in_place above causes the file to be left behind since the
             # user cannot remove it unless the parent directory is writable.
-            if link_data_only == 'copy_files' and trans.app.config.external_chown_script:
-                _chown(uploaded_dataset.path)
+            if link_data_only == 'copy_files':
```
mvdbeek:
```suggestion
            if link_data_only == 'copy_files' and trans.user:
```
should fix the selenium tests.

lib/galaxy/util/path/__init__.py
bernt-matthias: > Is this change critical for fixing hisitory import/export ? If not I would keep the current behavior and change it in another PR.

Could be a good idea. But then `external_chown` needs a proper return value: maybe `True` for success and `False` for failed executions or `external_chown_script is None`

bernt-matthias:
Thanks @mvdbeek .. last failing test seems unrelated to me. Will go for a test on our instance.

mvdbeek (approved):
Great, thanks a lot @bernt-matthias!
Let me know if this works and I'll merge it.

bernt-matthias:
Export to a file and import from a file seems successful on our instance. 

But import from an URL raises some problems (And I could need a bit of your intuition where to start digging): 

Galaxy tries to chown the URL: `Changing ownership of https://galaxy.intranet.ufz.de:8443/history/export_archive?id=f204fff4adbaeeeb with: ... ` which somehow translates to the file `GALAXY_ROOT/https:/galaxy.intranet.ufz.de:8443/history/export_archive?id=f204fff4adbaeeeb` which is not allowed on out system.

Then there are some tracebacks which could be consequence of the first one

```
job failed, detected state generic_error, standard error is - [Traceback (most recent call last):
  File "/gpfs1/data/galaxy_server/galaxy-dev/lib/galaxy/tools/imp_exp/unpack_tar_gz_archive.py", line 94, in <module>
    main(options, args)
  File "/gpfs1/data/galaxy_server/galaxy-dev/lib/galaxy/tools/imp_exp/unpack_tar_gz_archive.py", line 83, in main
    check_archive(archive_file, dest_dir)
  File "/gpfs1/data/galaxy_server/galaxy-dev/lib/galaxy/tools/imp_exp/unpack_tar_gz_archive.py", line 47, in check_archive
    with tarfile.open(archive_file, mode='r') as archive_fp:
  File "/gpfs1/data/galaxy_server/galaxy-dev/.venv/lib/python3.6/tarfile.py", line 1574, in open
    raise ReadError("file could not be opened successfully")
tarfile.ReadError: file could not be opened successfully
```

```
Unable to cleanup job 9879
  File "lib/galaxy/jobs/__init__.py", line 1814, in cleanup
    galaxy.tools.imp_exp.JobImportHistoryArchiveWrapper(self.app, self.job_id).cleanup_after_job()
  File "lib/galaxy/tools/imp_exp/__init__.py", line 58, in cleanup_after_job
    model_store.perform_import(new_history, job=job, new_history=True)
  File "lib/galaxy/model/store/__init__.py", line 169, in perform_import
    datasets_attrs = self.datasets_properties()
  File "lib/galaxy/model/store/__init__.py", line 776, in datasets_properties
    datasets_attrs = load(open(datasets_attrs_file_name))
FileNotFoundError: [Errno 2] No such file or directory: '/gpfs1/work/songalax/galaxy-dev/database/tmp/tmpuw32c2rh/datasets_attrs.txt'
```

mvdbeek:
Is the history you're trying to import via URL public and shared ? That's a prerequisite for import via URL to work and that has hit me previously as well.

mvdbeek:
There's also a possibility that import via URL from the same server is prohibited (in that case you'd probably want to just share the history,  but maybe it'd work if you add your server's IP to `fetch_url_whitelist`). Does this also fail if you import from a public server ?

bernt-matthias:
> Is the history you're trying to import via URL public and shared ?
> There's also a possibility that import via URL from the same server is prohibited 

Tried with a public history (https://usegalaxy.eu/u/maze/h/unnamed-history) to import on our system: same problem. Seems like the URL is interpreted as filename at some point.

bernt-matthias:
With the last commit import from URL works (tested for usegalaxy.eu -> our local instance). 

I guess the import from URL case is untested in Galaxy's testing framework?

bernt-matthias:
Played around with the download from the local URL. Seems that there is a redirect to the login page... Maybe its because we have `require_login: true`

bernt-matthias:
@mvdbeek do you think the failing selenium test is related. otherwise: ready?

lib/galaxy_test/api/_framework.py
```diff
@@ -1,3 +1,5 @@
+from unittest import TestCase
```
mvdbeek, line 1: I think that might be not quite right ? Did this maybe go wrong with the rebase ?

bernt-matthias: Actually this got lost during the rebase. If I just use `class ApiTestCase(FunctionalTestCase, UsesApiTestCaseMixin):` instead of class `ApiTestCase(FunctionalTestCase, UsesApiTestCaseMixin, TestCase):` I get 


```
____________________ ERROR collecting test/integration/test_history_import_export.py ____________________
test/integration/test_history_import_export.py:5: in <module>
    class ImportExportHistoryOutputsToWorkingDirTestCase(ImportExportHistoryTestCase, IntegrationTestCase):
E   TypeError: Cannot create a consistent method resolution
E   order (MRO) for bases TestCase, UsesApiTestCaseMixin
```

But maybe there is another solution for this Problem -- but I was unable to find one for days.

mvdbeek: OK, no, I think that's fine then

lib/galaxy/jobs/__init__.py
```diff
@@ -1932,8 +1930,9 @@ def get_output_path(self, dataset):
         for (hda, dataset_path) in self.output_hdas_and_paths.values():
             if hda == dataset:
                 return dataset_path
-        if getattr(dataset, "fake_dataset_association", False):
-            return dataset.file_name
+#         if getattr(dataset, "fake_dataset_association", False):
+#             log.error("GET_OUTPUT_PATH FAKE %s" %(dataset.file_name))
+#             return dataset.file_name
```
mvdbeek:
```suggestion
```

mvdbeek (approved):
If you can drop the commented out code this is good to go!

mvdbeek:
Thanks so much for working this out @bernt-matthias!

PR #8994, opened 2019-11-15
Merge 19.09 into dev

nsoranzo:

PR #9006, opened 2019-11-19
Fix tool image loading (was broken due to failing repo dir resolution)

dannon:
Popped up in https://github.com/galaxyproject/galaxy/commit/5f4df5eb932dbb91f66e010624925928c10fac69.  I think this more correctly keeps the previous behavior, while also not throwing it into an infinite loop.

@bgruening reported this, can be tested with a tool like https://usegalaxy.eu/shed_tool_static/toolshed.g2.bx.psu.edu/bgruening/deeptools_plot_enrichment/deeptools_plot_enrichment/3.0.2.0/plotEnrichment_output.png

### DIFF
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -937,8 +937,8 @@ class Tool(Dictifiable):
             if tool_dir.parts[-1] == self.repository_name:
                 return str(tool_dir)
             for parent in tool_dir.parents:
-                if parent.parts[-1] == self.repository_name:
-                    return str(parent)
+                if parent.stem == self.repository_name:
+                    return str(tool_dir)
             else:
                 log.error("Problem finding repository dir for tool [%s]" % self.id)
 
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -1,6 +1,7 @@
 """
 Classes encapsulating galaxy tools and tool configuration.
 """
+import itertools
 import json
 import logging
 import os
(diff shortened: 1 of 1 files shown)


mvdbeek:
This should also be fixed by forward merging https://github.com/galaxyproject/galaxy/pull/8994/commits/2f67794c78b66bd1c49c8464dc86da0804fb7adc, but using `.stem` is clearly better than `parts[-1]`, do you want to update this or should I do that ?

dannon:
@mvdbeek Argh, I didn't see there was another fix.  Yeah, I'll update this.

dannon:
@mvdbeek Done; you happy enough with the other version for 19.09, or do you think we should backport this, too?

mvdbeek (approved):
Yeah, I think the 19.09 one isn't broken, let's just keep this in dev

lib/galaxy/tools/__init__.py
```diff
@@ -934,11 +935,9 @@ def _repository_dir(self):
 
         if getattr(self, 'tool_shed', None):
             tool_dir = Path(self.tool_dir)
-            if tool_dir.parts[-1] == self.repository_name:
-                return str(tool_dir)
-            for parent in tool_dir.parents:
-                if parent.parts[-1] == self.repository_name:
-                    return str(parent)
+            for repo_dir in itertools.chain([tool_dir], tool_dir.parents):
+                if repo_dir.stem == self.repository_name:
```
nsoranzo, line 939: @dannon @mvdbeek What about changing `repo_dir.stem` to `repo_dir.name` ? That should be correct even in the unlikely case that `self.repository_name` contains a `.`

PR #9007, opened 2019-11-19
fix path to ts uwsgi app

martenson:
Otherwise run_tool_shed.sh fails (but works with uwsgi directly).

Possibly broken in https://github.com/galaxyproject/galaxy/pull/8958/commits/559b664b07fe73b0382cba5569b5ee6b7feb6416


martenson:
I think [this](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/config_manage.py#L336) is also wrong, but am not sure if it breaks anything.

martenson:
ping @jmchilton

jmchilton:
Awesome thanks!

PR #9010, opened 2019-11-19
Expose python version in api/config

hexylena:
So everyone can prove they're running py3 :wink: (more for stats/etc.)


lib/galaxy/managers/configuration.py
```diff
@@ -89,6 +90,7 @@ def _defaults_to(default):
             'registration_warning_message'      : _defaults_to(None),
             'welcome_url'                       : _defaults_to(None),
             'show_welcome_with_login'           : _defaults_to(True),
+            'python'                            : _defaults_to(list(sys.version_info)),
```
jmchilton: Can we just expose ``sys.version_info.major``? Point versions may make it clear what security exploits to target right?

hexylena: [Some IRL discussion]

That's fair! (*Counterpoint*: we already expose version of galaxy and there are known bugs for everything outside of 1 year old).

At least e.g. `3.7` would give us something actionable, which python features we can consider using maybe, I'm not interested in point releases either.

martenson (approved):

PR #9012, opened 2019-11-19
Merge release_19.09 into dev.

jmchilton:

PR #9013, opened 2019-11-19
Merge 19.09 into dev

nsoranzo:
More ToolShed-related conflicts.


nsoranzo:
Ha, I didn't notice #9012, this rearranges the conflincting imports slightly differently:

```python
from galaxy.util.script import (
    app_properties_from_args,
    populate_config_args
)
from tool_shed.webapp import (
    config as ts_config,
    model as ts_model
)
```

instead of:

```python
from galaxy.util.script import app_properties_from_args, populate_config_args
from tool_shed.webapp import config as ts_config
from tool_shed.webapp import model as ts_model
```

mvdbeek (approved):
I prefer this style of imports!

PR #9015, opened 2019-11-20
Fix API decorator order

nsoranzo:
in order to get a simple error message instead of an HTML error page when accessing API methods protected by the `@require_admin` decorator.


jmchilton:
Merci beaucoup!

PR #9018, opened 2019-11-20
Implement datasets grid as vue component

guerler:


jmchilton (approved):

lib/galaxy/webapps/galaxy/api/tags.py
```diff
@@ -0,0 +1,43 @@
...
+        """
+        item_id = payload.get("item_id")
+        item_class = payload.get("item_class")
+        item_tags = payload.get("item_tags")
+        if item_id is None:
+            raise MessageException("Please provide the item id (item_id).")
+        if item_class is None:
+            raise MessageException("Please provide the item class (item_class).")
+        if item_tags and len(item_tags) > 0:
+            item_tags = ",".join(item_tags).encode('utf-8')
```
mvdbeek: You'll get bytestrings here which seems wrong. Was there any reason for the encode call ?
```suggestion
            item_tags = ",".join(item_tags)
```

lib/galaxy/webapps/galaxy/api/tags.py
```diff
@@ -0,0 +1,43 @@
...
+        self.get_tag_handler(trans).delete_item_tags(user, item)
+        self.get_tag_handler(trans).apply_item_tags(user, item, item_tags)
+        trans.sa_session.flush()
+
+    def _get_item(self, trans, item_class_name, id):
+        """
+        Get an item based on type and id.
+        """
+        item_class = self.get_tag_handler(trans).item_tag_assoc_info[item_class_name].item_class
+        item = trans.sa_session.query(item_class).filter(item_class.id == id)[0]
```
mvdbeek: Does this work ?  `Query.filter` returns a Query object that you can't index, I think this should be:
```suggestion
        item = trans.sa_session.query(item_class).filter(item_class.id == id).first()
```

mvdbeek (dismissed):
I don't think in the current state the tags endpoint works. I think my suggestions should work, but it'd be even better if you could add an API test for the new endpoint.

lib/galaxy/webapps/galaxy/api/tags.py
guerler: This is how tags are filtered in the corresponding controller endpoint: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/controllers/tag.py#L206

mvdbeek: Can you write a test for this? I tried it in the db_shell and it doesn't work.

guerler: I added an api test case and applied the commit suggestion.

lib/galaxy/webapps/galaxy/api/tags.py
guerler: Got it from: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/controllers/tag.py#L77

mvdbeek: We want to deal with bytestrings as little as possible, and I don't see a reason to do that here. Can you update both places please ?

lib/galaxy/webapps/galaxy/api/tags.py
mvdbeek: Thanks!

guerler:
I have not changed anything but added a test case. It also works using `first()`. I removed the unicode encoding from the previous controller and the new api endpoint. Thank you for your review.

martenson:
couple of notes:

* the search is strictly case sensitive ("notebook" does not find "Notebook")
* name of the dataset is a dropdown menu, but UI does not indicate that
* I'm unclear about what this icon means:
Galaxy___martenson
* Somehow it shows a dataset without a name for me, I see nothing suspicious in the history itself
Galaxy___martenson

jmchilton:
I'm going to get pushback for this comment but I am me so I'll make the point and then drop it for a year. People don't use this endpoint and the old one sucked. The detailed reviews are very nice but they have downsides. Please be careful about encouraging your colleagues to polish things that will never be used - I would argue this isn't how we should allocate our time as a project. We should merge this with rough searching and questionable icons - and then fix it once we get complaints and determine it is a priority. I'd of course love to pile on and say my History menu drop down from #8699 is pretty cool and should be used here... but we need to move on.

guerler:
> couple of notes:
> 
> * the search is strictly case sensitive ("notebook" does not find "Notebook")

The datasets apis search endpoint has to be revisited, I did not want to include this within this PR.

> * name of the dataset is a dropdown menu, but UI does not indicate that

This will be solved soon, we are currently discussing how to handle it here and in the workflow list.

> * I'm unclear about what this icon means:
>   Galaxy___martenson

Its indicates that the dataset is in error. I added a tooltip to indicate that.

> * Somehow it shows a dataset without a name for me, I see nothing suspicious in the history itself

What is the dataset name? Can you check if the datasets api returns the correct json for it.

>   Galaxy___martenson

Thank you for the review.

martenson:
@guerler It actually seems that dataset has a legit empty name. I thought that is not possible, but seems we allow it. Thanks for the change.

martenson (approved):

PR #9019, opened 2019-11-20
Added application config store to Vuex

Nerdinacan:
Added a store module to access the galaxy configs from /api/configuration for use in configuring Vue components.

### DIFF
--- a/client/galaxy/scripts/store/configStore/queries.js
+++ b/client/galaxy/scripts/store/configStore/queries.js
@@ -2,7 +2,7 @@ import axios from "axios";
 import { prependPath } from "utils/redirect";
 
 export async function getConfig() {
-    const url = prependPath('/api/configuration');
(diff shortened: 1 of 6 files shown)


dannon:
Thanks, works for me, definitely good to keep migrating this kind of shared state to vuex

client/galaxy/scripts/store/configStore/queries.js
```diff
@@ -0,0 +1,11 @@
+import axios from "axios";
+import { prependPath } from "utils/redirect";
+
+export async function getConfig() {
+    const url = prependPath("/api/configuration");
```
martenson, line 5: I suspect this code is part of the javascript run on toolshed, which does not have this endpoint, hence broken -https://github.com/galaxyproject/galaxy/issues/9061

PR #9021, opened 2019-11-21
[19.09] Don't crash on containers without build number

mvdbeek:
Fixes `planemo lint --biocontainers tools/fasta_nucleotide_color_plot/fasta_nucleotide_color_plot.xml` for tools whose container has no build number (e.g https://biocontainers.pro/#/tools/openjdk)


jmchilton:
Fixed by updating Planemo's usage instead.

PR #9023, opened 2019-11-21
Default format 2 workflows to be downloaded as YAML.

jmchilton:
Exporting them as JSON or YAML wrapped in json (existing options) proved to be repeatedly confusing at the Elixir Biohackathon 2019. Next step toward making these the default export I think.

Issue #3088, opened 2016-10-25
Export Histories into Research Objects


bgruening:
@dannon and I talked yesterday about better and more structured way of exporting and reusing Galaxy histories. Exporting this into a http://www.researchobject.org might be a good solution.

lparsons:
Many of my users would also love a way to export a history _without_ the data (sort of like a workflow). In a way, "Create workflow from history" kind of does this, but it's often a bit broken and doesn't save the names/etc. Basically, having a way to retain history items, but remove the underlying dataset, would be very useful for archival purposes. One could retain the input data and some final output datasets, but remove the unnecessary intermediate files without losing the history of what was done.

nsoranzo:
Export of histories as BagIt bags was implemented in https://github.com/galaxyproject/galaxy/pull/7367 , but it was just a first step as explained in https://github.com/galaxyproject/galaxy/issues/4345#issuecomment-480811571 .

HadleyKing:
This is something we are looking at with [BioCompute](https://github.com/biocompute-objects) using the galaxy history AND galaxy workflow JSON outputs. I have been searching but have not been able to find specific documentation about how the the key:value pairs are generated and what they mean, for either of those objects. Does this exist?

stain:
Now it would make sense as a BagIt to just add a `ro-crate-metadata.json` file according to [RO-Crate](https://w3id.org/ro/crate/1.0/) - this could at least describe the workflow JSON and the derivation. 

Adding BioCompute IEEE 2791 would also make sense as I don't think it has a packaging at the moment.

Further work could look at the history provenance and model it according to [CWLProv](https://w3id.org/cwl/prov) with timestamps etc - but that would be more detailed and kept in a separate [PROV](https://www.w3.org/TR/prov-overview/) document - we would have to decide which flavour, e.g. PROV-XML vs PROV-JSON vs JSON-LD vs Turtle.

stain:
See also #9077. We suggested this as topic for the BCC [CoFest](https://www.open-bio.org/events/bosc/collaborationfest/) - hoping @HadleyKing et al will join!

HadleyKing:
> Now it would make sense as a BagIt to just add a `ro-crate-metadata.json` file according to [RO-Crate](https://w3id.org/ro/crate/1.0/) - this could at least describe the workflow JSON and the derivation.
> 
> Adding BioCompute IEEE 2791 would also make sense as I don't think it has a packaging at the moment.
> 
[...]

Issue #2491, opened 2016-06-13
Support wfprov and Research Objects Export of Workflow Invocations


jmchilton:
The discussion of workflow provenance exports came up on the CWL gitter channel and I received this advice from @stain:

"I would be biased, but I would of course recommend you use wfprov and put it in say JSON-LD format together with a wfdesc description of the (possibly implied) workflow and the values as individual files, then package it all as a ResearchObject bundle ZIP file - that way you can have multiple options for how to format things, e.g. you can include also a Galaxy-native representations and have a semi-structured file layout that users might understand (e.g. reflecting their workspace)
what is fun with Galaxy is that users build workflows "by accident" - so you can choose to either just model that as individual Process invocations without any overall workflow, or to imply the workflow from the values.
but the second might mean you have to keep historical values that since disappeared from the workspace
(or at least its provenance)
https://github.com/common-workflow-language/common-workflow-language/issues/84#issuecomment-195718238 shows a good representation approach by @ntijanic - however for interoperability I would do it in pure JSON-LD rather than in Schema Salad."

hexylena:
We have this now, right?

mvdbeek:
yeah :)

PR #9024, opened 2019-11-21
[WIP] Model Store Workflow Invocation Import/Export

jmchilton:
Early days here still... builds a bit on #9023 (mostly just for a clean diff).

Adds an API endpoint for a version that doesn't include files - this should be quick enough to handle in a web thread. A version that would include files like history export should probably mirror that and I'm not quite sure what the future of history import/export APIs look like. @mvdbeek thinks the jobs should just output the results to a history to simplify a lot of the plumbing and the UI.

Will land up supporting a Galaxy native format that sort of mirrors history exports, as well as bagit packaging.

It is probably a bug that history import/export doesn't include invocation information - this will hopefully fix that issue. Really meant a stepping stone for https://github.com/galaxyproject/galaxy/issues/3088 / https://github.com/galaxyproject/galaxy/issues/2491.

### DIFF
--- a/lib/galaxy/managers/workflows.py
+++ b/lib/galaxy/managers/workflows.py
@@ -321,6 +321,22 @@ class WorkflowContentsManager(UsesAnnotations):
             dict_or_raw_description = RawWorkflowDescription(dict_or_raw_description)
         return dict_or_raw_description
 
+    def read_workflow_from_path(self, app, user, path):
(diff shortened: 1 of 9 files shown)


mvdbeek:
@jmchilton is there anything in here we haven't already merged as part of other PRs ?

jmchilton:
I don't think so?

PR #9029, opened 2019-11-21
Hot module reloading using webpack-dev-server for client dev

dannon:
This isn't currently perfect -- it can occasionally go a little wonky due to old routing and reloads (which, for what it's worth, is going away soon anyway) -- but if that happens it just requires a page refresh to get back in sync.  It's also WIP while I get CSS updates working correctly (need to use style-loader for hmr instead of the extracted css file)

Paste is not supported as this relies on uWSGI routing.

The new `GALAXY_CLIENT_DEV_SERVER` env var modifies uwsgi routing to point to the hot reloading server, which can be run with `yarn run webpack-dev-server` (or the `client-dev-server` make target).

So, to test, do something like:
```
GALAXY_CLIENT_DEV_SERVER=1 sh run.sh
make client-dev-server
```

(and then open Galaxy in a browser -- separately edit something like ToolBox.vue and watch the automatic updates)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,11 +159,11 @@ client-production-maps: node-deps ## Rebuild client-side artifacts for a product
 client-format: node-deps ## Reformat client code
 	cd client && yarn run prettier
 
-client-watch: node-deps ## A useful target for parallel development building.
+client-watch: node-deps ## A useful target for parallel development building.  See also client-dev-server.
 	cd client && yarn run watch
 
-client-hmr-server: node-deps ## Starts a webpack dev server for client development (HMR enabled)
-	cd client && yarn run hmr-server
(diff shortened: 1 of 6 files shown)


jmchilton:
Seems like a huge time saver and awesome functionality. Can't the CSS stuff just come later?

dannon:
@jmchilton It can, but I thought I almost had it!  :)  (and then sunk more hours than I care to recall into webpack hell...)

~If I don't figure it out in the next 30 minutes I'll take it out of WIP and will follow up with the css bits separately.~  Yeah, it's not happening this morning; got sidelined.  I'll follow up separately.

Nerdinacan:
Thank you so much.

PR #9030, opened 2019-11-22
[WIP] Prototype of Minimal Workflow-Centric UI

jmchilton:
Experimental prototype developed based on discussions at the ELIXIR Biohackathon 2019 around convergence of [Galaksio](https://github.com/SGBC/galaksio) and other projects aimed at providing simpler workflow interfaces. 

A full report of the discussion and action items for the Galaxy framework are contained in the following document.

http://bit.ly/simplified-workflow-ui

### DIFF
--- a/client/galaxy/scripts/components/WorkflowHome.vue
+++ b/client/galaxy/scripts/components/WorkflowHome.vue
@@ -1,0 +1,23 @@
+<template>
+    <div>
+        <b>Workflows</b>
+        <workflow-list :simplified="true" />
+        <b>Invocations</b>
+        <recent-invocations :showTitle="false" :simplified="true" />
+    </div>
+</template>
+
+<script>
+import WorkflowList from "components/Workflow/WorkflowList";
+import RecentInvocations from "components/User/RecentInvocations";
+
+export default {
+    components: {
+        WorkflowList,
+        RecentInvocations,
+    },
+    data() {
+        return {};
+    },
+};
+</script>
--- a/client/src/components/Masthead/Masthead.vue
+++ b/client/src/components/Masthead/Masthead.vue
@@ -65,6 +65,9 @@ export default {
             type: Object,
             default: null,
         },
+        refreshMenu() {
+            this.baseTabs = fetchMenu(this.menuOptions);
+        },
     },
     data() {
         return {
--- a/client/src/components/Workflow/Invocations.vue
+++ b/client/src/components/Workflow/Invocations.vue
@@ -1,9 +1,9 @@
 <template>
     <div class="invocations-list">
-        <h2 class="mb-3">
+        <h2 class="mb-3" v-if="showTitle">
             <span id="invocations-title">{{ title }}</span>
         </h2>
-        <b-alert v-if="headerMessage" variant="info" show>
+        <b-alert v-if="headerMessage && showTitle" variant="info" show>
             {{ headerMessage }}
         </b-alert>
         <b-alert class="index-grid-message" :variant="messageVariant" :show="showMessage">{{ message }}</b-alert>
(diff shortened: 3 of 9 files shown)


nsoranzo:
Thanks @TKlingstrom for proposing the Biohackathon 2019 project and for supporting the merge of Galaksio concepts upstream!

client/galaxy/scripts/components/Workflow/WorkflowList.vue
```diff
@@ -85,31 +85,27 @@ import { Services } from "./services.js";
...
-                },
-                {
-                    key: "bookmark"
-                },
-                {
-                    key: "execute",
-                    label: ""
-                }
-            ],
+            fields: this.simplified ?  [NAME_FIELD, TAGS_FIELD, EXECUTE_FIELD] : [NAME_FIELD, TAGS_FIELD, BOOKMARK_FIELD, EXECUTE_FIELD],
```
guerler: This might be easier to read and work with if we just create two dictionaries like `fieldsSimple` and `fieldsComplex` and then add a computed parameter name `fields` which returns one of them depending on the `simplified` flag. Same goes for the handling of field sets in Invocations.vue.

client/galaxy/scripts/layout/page.js
```diff
@@ -22,7 +22,12 @@ const View = Backbone.View.extend({
             show_inactivity_warning: false,
             inactivity_box_content: "",
             hide_panels: false,
-            hide_masthead: false
+            hide_masthead: false,
+            masthead_show_analysis: true,
+            masthead_show_workflows: true,
+            masthead_show_workflows: true,
+            masthead_show_activities: false,
+            masthead_show_scratchbook: true,
```
guerler: If we would introduce `hide_masthead_analysis` instead the default would be false per definition and most of these values would not have to be initialized/specified when building the masthead. Also the naming would be more consistent with the parameters above.

client/galaxy/scripts/components/Workflow/WorkflowList.vue
jmchilton: Good idea - sounds a lot cleaner! I will do this. Thanks.

client/galaxy/scripts/layout/page.js
jmchilton: The problem with a meta property like that is that we've discussed per role display of features in the UI. Might be better to just break everything down now? Then we can do ``view_default && role_default`` for each option? But maybe that could be pushed down a layer? All this code will be obsoleted by #9071 - I'll revisit this and try both after that is integrated and I redo this.

afgane:
I just came across this PR and wanted to add some thoughts about the layout of the workflow list. The layout may not necessarily be in scope of this PR but maybe a good to open some discussion about the ideas. Here are some mockup drawings that have been annotated and are hopefully self sufficient. 

image

image

bgruening:
@afgane that looks cool. Is the right-hand side is only shown when something is selected? What is in this place when nothing is selected? Is the workflow-list moving to the left when you select something? Are multiple-selects allowed?

afgane:
> Is the right-hand side is only shown when something is selected? What is in this place when nothing is selected? 
Either way. We could have an _info_ panel similar to how Google Docs has it that is toggled with a button and is empty (or shows some instructional text, advertises a feature, etc.) until a workflow is selected at which point it populates with details specific to the workflow. 

Alternatively, it could be like Box where the panel is always shown and has two tabs for sharing and details.
image


> Is the workflow-list moving to the left when you select something? 

It just narrows the view. If you look at John's screenshot above, that page is way too sparse as is.

> Are multiple-selects allowed?

What is a scenario where you would select multiple workflows? Regardless, sure, they panel can just say "Multiple workflows selected".

PR #9034, opened 2019-11-25
escape double hyphen in IT sample config, making it valid XML

martenson:


config/job_conf.xml.interactivetools
```diff
@@ -18,13 +18,15 @@
...
                If you develop interactive tools on your 'localhost' and don't have a proper domain name
-               you need to tell all Docker containers a hostname where Galaxy is running. 
-               This can be done via the --add-host parameter during the `docker run` command.
+               you need to tell all Docker containers a hostname where Galaxy is running.
+               This can be done via the add-host parameter during the `docker run` command.
                'my-hostname' here is an arbritrary hostname that matches the IP address of your
                Galaxy host. Make sure this hostname ('my-hostname') is also set in your galaxy.yml file, e.g.
-               `galaxy_infrastructure_url: http://my-hostname:8080`. 
+               `galaxy_infrastructure_url: http://my-hostname:8080`.
+              Note: XML does not allow double hyphen hence the line below has it escaped as &#45;
```
nsoranzo:
```suggestion
              Note: XML does not allow double hyphens in comments, hence the line below has them escaped as &#45;
```

PR #9047, opened 2019-11-29
[19.05] Fix input parameter comparisons

mvdbeek:
Fixes https://github.com/galaxyproject/tools-iuc/pull/2702#issuecomment-559718768
Broken in https://github.com/galaxyproject/galaxy/pull/7826

The implementation of `__gt__` and `__eq__` is simpler now. One possible breaking change is that a TextToolParameter with value "5" is not equal to a literal 5 anymore. I doubt anyone relied on this behavior.

### DIFF
--- a/lib/galaxy/tools/wrappers.py
+++ b/lib/galaxy/tools/wrappers.py
@@ -79,10 +79,7 @@ class InputValueWrapper(ToolParameterValueWrapper):
         self.value = value
         self._other_values = other_values
 
-    def __eq__(self, other):
-        return self.get_cast_value() == other
-
-    def get_cast_value(self):
+    def _get_cast_value(self):
         cast = {
             'text': str,
             'integer': int,
@@ -90,6 +87,9 @@ class InputValueWrapper(ToolParameterValueWrapper):
             'boolean': bool,
         }
         return cast.get(self.input.type, str)(self)
+
+    def __eq__(self, other):
+        return self._get_cast_value() == other
 
     def __ne__(self, other):
         return not self == other
@@ -112,7 +112,7 @@ class InputValueWrapper(ToolParameterValueWrapper):
         return getattr(self.value, key)
 
     def __gt__(self, other):
-        return self.get_cast_value() > other
+        return self._get_cast_value() > other
 
     def __int__(self):
         return int(float(self))
--- a/lib/galaxy/tools/wrappers.py
+++ b/lib/galaxy/tools/wrappers.py
@@ -80,14 +80,12 @@ class InputValueWrapper(ToolParameterValueWrapper):
         self._other_values = other_values
 
     def __eq__(self, other):
-        if isinstance(other, string_types):
+        if isinstance(other, string_types) or (isinstance(other, InputValueWrapper) and other.input.type == 'text'):
             return str(self) == other
-        elif isinstance(other, int):
-            return int(self) == other
-        elif isinstance(other, float):
+        elif isinstance(other, (int, float)) or (isinstance(other, InputValueWrapper) and other.input.type in ('float', 'integer')):
             return float(self) == other
         else:
-            return super(InputValueWrapper, self) == other
+            return self.value == other
 
     def __ne__(self, other):
         return not self == other
@@ -110,14 +108,12 @@ class InputValueWrapper(ToolParameterValueWrapper):
         return getattr(self.value, key)
 
     def __gt__(self, other):
-        if isinstance(other, string_types):
+        if isinstance(other, string_types) or (isinstance(other, InputValueWrapper) and other.input.type == 'text'):
             return str(self) > other
-        elif isinstance(other, int):
-            return int(self) > other
-        elif isinstance(other, float):
+        elif isinstance(other, (int, float)) or (isinstance(other, InputValueWrapper) and other.input.type in ('float', 'integer')):
             return float(self) > other
         else:
-            super(InputValueWrapper, self).__gt__(other)
+            return self.value > other
 
     def __int__(self):
         return int(float(self))
--- a/lib/galaxy/tools/wrappers.py
+++ b/lib/galaxy/tools/wrappers.py
@@ -80,12 +80,16 @@ class InputValueWrapper(ToolParameterValueWrapper):
         self._other_values = other_values
 
     def __eq__(self, other):
-        if isinstance(other, string_types) or (isinstance(other, InputValueWrapper) and other.input.type == 'text'):
-            return str(self) == other
-        elif isinstance(other, (int, float)) or (isinstance(other, InputValueWrapper) and other.input.type in ('float', 'integer')):
-            return float(self) == other
-        else:
-            return self.value == other
+        return self.get_cast_value() == other
+
+    def get_cast_value(self):
+        cast = {
+            'text': str,
+            'integer': int,
+            'float': float,
+            'boolean': bool,
+        }
+        return cast.get(self.input.type, str)(self)
 
     def __ne__(self, other):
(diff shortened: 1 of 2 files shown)


lib/galaxy/tools/wrappers.py
```diff
@@ -80,14 +80,12 @@ def __init__(self, input, value, other_values={}):
         self._other_values = other_values
 
     def __eq__(self, other):
-        if isinstance(other, string_types):
+        if isinstance(other, string_types) or (isinstance(other, InputValueWrapper) and other.input.type == 'text'):
             return str(self) == other
-        elif isinstance(other, int):
-            return int(self) == other
-        elif isinstance(other, float):
+        elif isinstance(other, (int, float)) or (isinstance(other, InputValueWrapper) and other.input.type in ('float', 'integer')):
```
nsoranzo: I'd keep int and float separate, just to be sure.

lib/galaxy/tools/wrappers.py
```diff
@@ -80,14 +80,12 @@ def __init__(self, input, value, other_values={}):
...
             return float(self) == other
         else:
-            return super(InputValueWrapper, self) == other
+            return self.value == other
```
nsoranzo: Not sure I'd go that far.

lib/galaxy/tools/wrappers.py
```diff
@@ -110,14 +108,12 @@ def __getattr__(self, key):
...
+        if isinstance(other, string_types) or (isinstance(other, InputValueWrapper) and other.input.type == 'text'):
             return str(self) > other
-        elif isinstance(other, int):
-            return int(self) > other
-        elif isinstance(other, float):
+        elif isinstance(other, (int, float)) or (isinstance(other, InputValueWrapper) and other.input.type in ('float', 'integer')):
             return float(self) > other
         else:
-            super(InputValueWrapper, self).__gt__(other)
+            return self.value > other
```
nsoranzo: Here too.

lib/galaxy/tools/wrappers.py
mvdbeek: I think we need to treat them the same, otherwise the last test I added (5 < 5.1) would fail, or did I miss something?

nsoranzo: Wouldn't the following work?

```suggestion
        elif isinstance(other, int) or (isinstance(other, InputValueWrapper) and other.input.type == 'integer'):
            return int(self) == other
        elif isinstance(other, float) or (isinstance(other, InputValueWrapper) and other.input.type == 'float'):
```

lib/galaxy/tools/wrappers.py
mvdbeek: We could instead raise an exception here, but this should work for comparing color or boolean parameters for instance.

lib/galaxy/tools/wrappers.py
mvdbeek: If self wraps a float this would lead to wrong results, e.g 5.1 == 5. It would be ok if we made this 
```
elif self.input.type == 'integer' and  isinstance(other, int) or (isinstance(other, InputValueWrapper) and other.input.type == 'integer'):
```
But then we also need to take care of the case where `self.input.type == 'integer'` and other is float-ish ... .
On the other hand I think casting to float should work fine in most/all cases.

mvdbeek: Or ... we turn this around and cast self to the correct type. That seems like less guesswork.

lib/galaxy/tools/wrappers.py
nsoranzo: I'd prefer to cross that bridge when there is a reported issue/use case, especially since this PR is targeting 19.05.

mvdbeek: I agree with the sentiment, but what's your suggestion ? This doesn't work at all right now (none of the parent classes define `__gt__` or `__eq__`). I'd be in favor of casting self to string.

test/unit/tools/test_wrappers.py
```diff
@@ -83,17 +85,40 @@ def test_raw_object_wrapper():
...
     assert wrapper < 10
+    assert wrapper <= 5.1
+
+
+@with_mock_tool
+def test_input_value_wrapper_input_value_wrapper_comparison(tool):
+    wrapper = valuewrapper(tool, 5, "integer")
+    assert str(wrapper) == valuewrapper(tool, "5", "text")
+    assert int(wrapper) == valuewrapper(tool, "5", "integer")
+    assert wrapper == valuewrapper(tool, "5", "text")
```
nsoranzo: I think this should not be allowed, too much magic, even in normal Python `"5" == 5` is False.

mvdbeek: But that has previously worked, see the test above that I didn't modify.

mvdbeek: that said this naturally falls out of casting InputValueWrapper to InputValueWrapper.input.type (https://github.com/galaxyproject/galaxy/pull/9047/commits/155660083d66d00423c2ed5ef6ee9028badd4883), which maybe means we should do this ?

test/unit/tools/test_wrappers.py
```diff
@@ -83,17 +85,40 @@ def test_raw_object_wrapper():
...
+    wrapper = valuewrapper(tool, 5, "integer")
     assert str(wrapper) == "5"
     assert int(wrapper) == 5
-    assert wrapper == "5"
+    assert wrapper != "5"
     assert wrapper == 5
     assert wrapper == 5.0
     assert wrapper > 2
     assert wrapper < 10
+    assert wrapper <= 5.1
```
nsoranzo:
```suggestion
    assert wrapper < 5.1
```

test/unit/tools/test_wrappers.py
```diff
@@ -83,17 +85,40 @@ def test_raw_object_wrapper():
...
+def test_input_value_wrapper_input_value_wrapper_comparison(tool):
+    wrapper = valuewrapper(tool, 5, "integer")
+    assert str(wrapper) == valuewrapper(tool, "5", "text")
+    assert int(wrapper) == valuewrapper(tool, "5", "integer")
+    assert wrapper != valuewrapper(tool, "5", "text")
+    assert wrapper == valuewrapper(tool, "5", "integer")
+    assert wrapper == valuewrapper(tool, "5", "float")
+    assert wrapper > valuewrapper(tool, "2", "integer")
+    assert wrapper < valuewrapper(tool, "10", "integer")
+    assert wrapper <= valuewrapper(tool, "5.1", "float")
```
nsoranzo:
```suggestion
    assert wrapper < valuewrapper(tool, "5.1", "float")
```

lib/galaxy/tools/wrappers.py
```diff
@@ -80,14 +80,16 @@ def __init__(self, input, value, other_values={}):
...
-            return str(self) == other
-        elif isinstance(other, int):
-            return int(self) == other
-        elif isinstance(other, float):
-            return float(self) == other
-        else:
-            return super(InputValueWrapper, self) == other
+        return self.get_cast_value() == other
+
+    def get_cast_value(self):
```
nsoranzo:
```suggestion
    def _get_cast_value(self):
```
unless you intended this to be exposed.

Also I'd move it before `__eq__()`.

nsoranzo (commented):
I like this much better, thanks!

mvdbeek:
> I like this much better, thanks!

Yep, thanks for the help and discussions.

nsoranzo (approved):

PR #9059, opened 2019-12-02
[19.05] Small refactor of ``InputValueWrapper`` comparison methods

nsoranzo:
Add more tests.

Follow-up on #9054.


mvdbeek (approved):
Nice, thanks for the cleanup!

PR #9062, opened 2019-12-04
Add a config option to allow customizing the brand title

guerler:


lib/galaxy/managers/configuration.py
```diff
@@ -40,6 +40,7 @@ def _required_attribute(item, key, **context):
             # TODO: this is available from user data, remove
             'is_admin_user'                     : lambda *a, **c: False,
             'brand'                             : _required_attribute,
+            'brand_enabled'                     : _defaults_to(True),
```
jdavcs: you can use `_required_attribute` - that way it will default to the default value specified in the schema. The only reason for using `_defaults_to` in that file is to address special cases - when the schema's default is different or when the option  is not defined in the schema. (maybe `_required_attribute` was not the best name choice, it's definitely not intuitive...)

guerler: Cool, thank you!

bernt-matthias:
Can't one just set the brand to `"" `?

martenson:
I am confused by this too. Why is the presence of `brand` config option not treated as a flag?

jdavcs:
There is no default specified in the schema, so just *not* setting the `brand` option will result in `app.config.brand` being `None`.  +1 for treating it as a flag.

guerler:
This configuration option determines wheter the word "Galaxy" is shown in the masthead brand title or not. Setting the brand to "" or " " does not hide it. This option can be helpful if admins want to use custom logos which e.g. already contain the word "Galaxy". An alternative to this PR would be to create a new default logo for Galaxy which contains the term "Galaxy" in it.

martenson:
What about defaulting brand to "Galaxy", and deleting the hardcoded Galaxy? (Sorry for the bikeshedding 😖)

guerler:
@martenson then usegalaxy.eu and other servers would just show the Galaxy icon and e.g. "Europe". No worries and thanks for the review. Nothing wrong with spending a few minutes to make sure that adding a new config option is the way to go or not.

martenson:
@guerler They'll change their configs to what they want.

guerler:
So we would have to tell in the release notes that all 100+ servers would have to change their config to fix the name otherwise it just changes with the update. I would like to avoid that.

martenson:
I'd probably be fine with that, but I understand it is not smooth. In this case I am also fine with adding a new option, either as this PR does or something like galaxy_brand that defaults to `Galaxy`.

martenson (approved):
Could we please rename the option to something like `show_galaxy_in_masthead` or similar, so the relationship to `brand` is more clear?

martenson (approved):

martenson:
thanks @guerler !

PR #9070, opened 2019-12-06
[WIP] Remove templates/webapps/tool_shed.

jmchilton:
Should be moved into lib/tool_shed/webapp/templates right - I suspect I missed these files for a reason though. Just checking the tests.


nsoranzo:
Several ToolShed tests fail with `TemplateLookupException: Cant locate template for uri u'/webapps/tool_shed/repository/common.mako'`

martenson:
I suspect it includes galaxy admin controllers looking for shed's makos (like `admin/tool_shed_repository/select_tool_panel_section.mako`)

PR #9072, opened 2019-12-06
Disable the admin_toolshed controller (except for tests).

jmchilton:
Screen Shot 2019-12-06 at 12 29 27 PM

Everyone wants the controller gone - the admin entry point has been replaced with API + Vue so disable. But we are keeping the controller around just until we can unwind all the tests from the old Galaxy admin GUI (some of them already have been redirected to the use the API - at least in my cleanup branch e.g. https://github.com/galaxyproject/galaxy/pull/8981/commits/9b50d0631a0468f272fc544c89afcd0fa8183413) so enable it only if tests are running.

### DIFF
--- a/lib/galaxy/webapps/galaxy/controllers/admin_toolshed.py
+++ b/lib/galaxy/webapps/galaxy/controllers/admin_toolshed.py
@@ -1,12 +1,14 @@
 import json
 import logging
 import os
+from functools import wraps
 
 from six import string_types
 from sqlalchemy import or_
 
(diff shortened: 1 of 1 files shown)


guerler:
Thank you!

jmchilton:
Out of WIP - sorry for the delay.

mvdbeek:
@jmchilton can you resolve the conflict ?

mvdbeek (approved):

PR #9078, opened 2019-12-07
[19.09] Fix mulled version comparison

mvdbeek:
We noticed that often times version_sorted wouldn't return the newest build for a given version.
To fix this we need to compare the versions excluding the build suffix (`--py3_0`), since the suffixes don't follow a particular versioning scheme. Instead we first sort by modification date, then by version (stripped of the build suffix if present).

### DIFF
--- a/lib/galaxy/tool_util/deps/mulled/util.py
+++ b/lib/galaxy/tool_util/deps/mulled/util.py
@@ -81,7 +81,7 @@ def mulled_tags_for(namespace, image, tag_prefix=None):
 
 
 def split_tag(tag):
-    """Split mulled image name into conda version and conda build."""
+    """Split mulled image tag into conda version and conda build."""
     return tag.rsplit('--', 1)
 
 
--- a/lib/galaxy/tool_util/deps/mulled/util.py
+++ b/lib/galaxy/tool_util/deps/mulled/util.py
@@ -3,6 +3,7 @@ from __future__ import print_function
 
 import collections
 import hashlib
+import re
 import sys
 import threading
 import time
@@ -13,7 +14,9 @@ try:
 except ImportError:
     requests = None
 
+BUILD_NUMBER_REGEX = re.compile(r'\d+$')
 MULLED_TAG_CACHE = collections.defaultdict(dict)
+PARSED_TAG = collections.namedtuple('ParsedTag', 'tag version build_string build_number')
 
 
 def create_repository(namespace, repo_name, oauth_token):
@@ -38,7 +41,7 @@ def quay_versions(namespace, pkg_name):
     if 'tags' not in data:
         raise Exception("Unexpected response from quay.io - not tags description found [%s]" % data)
 
-    return [tag for tag in data['tags'] if tag != 'latest']
+    return [tag for tag in data['tags'].keys() if tag != 'latest']
 
 
 def quay_repository(namespace, pkg_name):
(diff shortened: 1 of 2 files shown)


bgruening:
@mvdbeek modification date seems to be brittle, especially if we take Singularity on CVMFS into play, with caching and so on.

If the problem is the build string, can we first sort the version without the build-string and then sort the build-number -- which is part of the build-string?

mvdbeek:
No, there's no scheme in the build numbers. This isn't ideal, but they're
not going to change the API response.

bgruening:
Are you referring to `--py3_0`? This is the build string. I'm referring to `_0` in this case as the build number. The higher the build number the more recent the container should be.

mvdbeek:
Yes, but that doesn't matter if the py3 changes to py4. and this happens a lot. If you want to check this you can cherry-pick this commit into dev and go to the container resolver menu, you'll see that this works as well as we can given the information we get from the tag alone. There isn't actually any issue with sorting the _0, this works fine because these are always bumped in chronological order.

mvdbeek:
In fact you can go to any quay io biocontainer repo that has a reasonable number of available tags, and you'll see that you cannot sort by version plus _n alone

mvdbeek:
OK, I think I see the order in bwa even though the first part of the build suffix changes. I guess that could work ?

bgruening:
The build-suffix can change, yes, but the build-number, the thing after the last `_` should always increase. I would simply ignore the suffix and only take the `build-number`.

lib/galaxy/tool_util/deps/mulled/util.py
```diff
@@ -79,14 +82,32 @@ def mulled_tags_for(namespace, image, tag_prefix=None):
...
-    return version, build
+    return tag.split('--', 1)
+
+
+def parse_tag(tag):
+    """Split mulled image name into version and build number."""
+    version = tag
+    build_string = "-1"
+    if '--' in tag:
+        version, build_string = tag.rsplit('--', 1)
```
bgruening, line 93: isn't that what split_tag() should do?

mvdbeek:
OK, now with tests and new sort order:
- build string
- build number
- version

lib/galaxy/tool_util/deps/mulled/util.py
```diff
@@ -79,14 +82,32 @@ def mulled_tags_for(namespace, image, tag_prefix=None):
...
+    elif '-' in tag:
+        # Should be mulled multi-container image tag
+        version, build_string = tag.rsplit('-', 1)
+    build_number = int(BUILD_NUMBER_REGEX.search(tag).group(0))
```
bgruening, line 97: It would be safer to search on the build_string here, isn't it?

lib/galaxy/tool_util/deps/mulled/util.py
```diff
@@ -79,14 +82,32 @@ def mulled_tags_for(namespace, image, tag_prefix=None):
...
 def split_tag(tag):
     """Split mulled image name into conda version and conda build."""
-    version = tag.split('--', 1)[0]
-    build = tag.split('--', 1)[1]
-    return version, build
+    return tag.split('--', 1)
+
+
+def parse_tag(tag):
+    """Split mulled image name into version and build number."""
```
bgruening: Do we have here the `image name` or the tag?

lib/galaxy/tool_util/deps/mulled/util.py
```diff
@@ -79,14 +82,32 @@ def mulled_tags_for(namespace, image, tag_prefix=None):
 
 def split_tag(tag):
     """Split mulled image name into conda version and conda build."""
-    version = tag.split('--', 1)[0]
-    build = tag.split('--', 1)[1]
-    return version, build
+    return tag.split('--', 1)
```
bgruening: rsplit?

mvdbeek: don't want to change this, this is used in other places

lib/galaxy/tool_util/deps/mulled/util.py
mvdbeek: tag

lib/galaxy/tool_util/deps/mulled/util.py
mvdbeek: `split_tag` is used on image names, so this wouldn't return the correct elements.

lib/galaxy/tool_util/deps/mulled/util.py
mvdbeek: that wouldn't work for mulled suffixes

mvdbeek: actually it might work, but I'd have to change the fallback and the regex, else -1 becomes 1. It just seems safer to use it on the unmodified tag

lib/galaxy/tool_util/deps/mulled/util.py
mvdbeek: although that's probably safer, will do it

lib/galaxy/tool_util/deps/mulled/util.py
```diff
@@ -79,14 +82,32 @@ def mulled_tags_for(namespace, image, tag_prefix=None):
 
 def split_tag(tag):
     """Split mulled image name into conda version and conda build."""
```
bgruening: This is also not the image name but the image tag. Can you please change this older message.

mvdbeek: Done, does it look ok now @bgruening ?

bgruening (approved):

lib/galaxy/tool_util/deps/mulled/util.py
bgruening: Yes :) Thanks!

PR #9079, opened 2019-12-07
[WIP] xml parser: input elements get attributes also from sub tags

bernt-matthias:
As previously discussed on gitter it is sometimes helpful to specify attributes as subtags. In particular because this allows to use CDATA. This already worked for the attributes `label` and `help`. 

This adds `value` to this list. 

The xml parser can now do this for each attribute, but the xsd limits the functionality to the attributes mentioned above.


bernt-matthias:
Anyone an idea why the generated command lines contain True / False instead of the respective true/falsevalue?

bernt-matthias:
Would be nice to extend this to allattributes where one often needs special characters, e.g. `pattern="(?P&lt;designation&gt;.+)\.tsv"`

PR #9086, opened 2019-12-10
Even Richer Workflow Inputs

jmchilton:
Builds on #9073 (Richer Workflow Inputs).

Adds:

 - Ability to specify formats for data and collection inputs. Currently collections have no way to restrict formats - and data parameters will be restricted automatically but workflow authors may be able to provide more precise, correct data types than Galaxy can (or does) infer. Formats may be specified as a single value or a list of values in the corresponding workflow format. 
 - Ability to add "restrictions" on text data inputs - so they appear as static select boxes from supplied key/values, values determined dynamically (this commit includes #8264), or a text box with suggestions. These changes include changing the workflow editor logic so that text input can connect to ``select`` inputs.

I had wanted to use Repeats for these, but they don't work with the workflow module framework right now and it wasn't obvious how to adjust things so they would. Instead users supply these as common separated values in a text box. Since we're not persisting ideas related to the client plumbing into the database - there should be nothing to prevent us from re-working the interface to use different client form stuff in the future as long as we're on board with supporting the follow Format 2 syntax elements:

```
inputs:
  fasta_input:
     type: data
     format: fasta
```

```
inputs:
  sequences:
     type: data
     format:
      - fasta
      - fastqsanger
```

```
inputs:
  sequencer:
     type: text
     restrictions:
      - iseq
      - miniseq
      - miseq
      - nextseq
      - nanopore
 ```

```
inputs:
  sequencer:
     type: text
     restrictions:
      - value: iseq
        label: iSeq
      - value: miniseq
        label: MiniSeq
      - value: miseq
        label: MiSeq
      - value: nextseq
        label: NextSeq
      - value: nanopore
        label: Nanopore
 ```

```
inputs:
  sequencer:
     type: text
     restrictOnConnections: true
```

```
inputs:
  sequencer:
     type: text
     suggestions:
      - iseq
      - miniseq
```


jmchilton:
Sleeping on this I think my new plan is to wrap the ``restrictions`` parameter in a conditional with four options:

- Do not specify restrictions (default).
- Attempt restriction based on connections.
- Provide list of restricted values.
- Provide list of suggested values.

For restrict based on select connections - I've rebased  #8264 into a single commit with 3 authors https://github.com/galaxyproject/galaxy/commit/98c9b50a944fd1954442b26ca3f09810310439e1  that I can apply to this and rework.

Plan for select merging would be I think:
 - Rework parameter finding to visit_input_values right (would eliminate a lot of the code).
 - Remove the select option - just make it part of the text option.

The corresponding Format 2 syntaxes for four options would be something like:

```
inputs:
  sequencer:
     type: text
```


```
inputs:
  sequencer:
     type: text
     restrictions:
      - iseq
      - miniseq
```

```
inputs:
  sequencer:
     type: text
     restrictOnConnections: true
```

```
inputs:
  sequencer:
     type: text
     suggestions:
      - iseq
      - miniseq
```

jmchilton:
Well I messed up the cherry-pick because I started refactoring while I should have been merging the conflicts and committing first - so it is hard to see the lineage - but the authors of #8264 are on the commit and a good amount of the code is still there. Using visit_input_values really did reduce the amount of code needed for that.

I think otherwise the idea worked out:

Screen Shot 2019-12-10 at 11 05 03 AM

I think a lot could go wrong when generating those values so the option in the GUI is called "Attempt restriction based on connections" and this code fallbacks to a plain text input if option generation or intersection fails.

A couple more ideas:
 - If just one connection is made, just use the options - because intersection breaks the ordering.
 - Allow restriction and suggestion lists to define key value pairs. This could be rich in the Format2 source and something like : ``key1:A description of key2, key2:A description of key2`` in the client form.

mvdbeek:
This looks amazing, and yeah, visit_input_values is the right thing.

jmchilton:
Having these constructs for building and configuring inputs and having the abstractions to separate that representation from whats gets persisted... combined with having expression tooling framework merged... we could finally build my arbitrary expression module with user described inputs and outputs...

That combined this with the conditionals proposal for CWL (https://github.com/common-workflow-language/cwl-v1.2/pull/4) which is so clean, well thought through, and  I think would map relatively cleanly on top of all of these new Galaxy constructs... 

It isn't on the roadmap... but it is hard not to keep going 😆...

lib/galaxy/workflow/modules.py
```diff
@@ -626,21 +719,32 @@ class InputDataCollectionModule(InputModule):
...
-                <option value="paired">Dataset Pair</option>
-                <option value="list:paired">List of Dataset Pairs</option>
-            </param>
-            ''' % collection_type))
-        return dict(collection_type=input_collection_type)
+        parameter_def = self._parse_state_into_dict()
+        collection_type = parameter_def["collection_type"]
+        optional = parameter_def["optional"]
+        collection_type_source = dict(name="collection_type", label="Collection type", type="text", value=collection_type)
+        collection_type_source["options"] = [
```
mvdbeek, line 730: ❤️

lib/galaxy/tool_util/parser/factory.py
```diff
@@ -82,7 +82,10 @@ def get_input_source(content):
     consume using the tool input source interface.
     """
     if not isinstance(content, InputSource):
-        content = XmlInputSource(content)
+        if isinstance(content, dict):
```
mvdbeek, line 85: Can you update the the docstring for this ?

lib/galaxy/workflow/modules.py
```diff
@@ -652,51 +756,281 @@ def get_all_outputs(self, data_only=False):
...
+        parameter_type_cond.test_param = input_parameter_type
+        cases = []
+
+        for param_type in ["text", "integer", "float"]:
+            if param_type == "text":
+                if parameter_type == "text":
+                    default = parameter_def.get("default") or ""
+                else:
+                    default = ""
+                input_default_value = TextToolParameter(None, XML(
```
mvdbeek: Can we use the new dict source for this now ?

mvdbeek (approved):
This is awesome! There are still some rough edges (as expected), but we can fix these later IMO.
One thing we should do is disconnect optional parameter output nodes from inputs that are required, and we should have some more validation prior to launching the workflow.

lib/galaxy/tool_util/parser/factory.py
jmchilton: Done locally - will push in a second.

lib/galaxy/workflow/modules.py
jmchilton: Done locally - will push in a second. Eliminates XML and Element imports 😄.

jmchilton:
> One thing we should do is disconnect optional parameter output nodes from inputs that are required

Yeah... I forgot about the editor - the declared formats are also not used. I'll give a shot at these two and see what I come up with.

> we should have some more validation prior to launching the workflow

Do you have a specific thing that wasn't caught?

mvdbeek:
Yeah, making a dataset optional, then not providing it. Same with an optional integer parameter connected to a required value and not provided.

mvdbeek:
But I'm happy to merge without this if this looks to be a bigger effort (I could also look into that).

jmchilton:
First cut at the workflow editor. Handles optionals in data and collections, incorporates the format information. Handles updates to some degree. There are some corner cases I think - I'm not sure if they need to be addressed pre-merge:

- Optional parameter types aren't validated at the connection level.
- Switching format types or going from non-optional to optional on inputs could cause existing connections to become invalid. These aren't destroyed automatically.

client/galaxy/scripts/mvc/workflow/workflow-terminals.js
```diff
@@ -370,6 +377,12 @@ var BaseInputTerminal = Terminal.extend({
             `Effective output data type(s) [${cat_outputs}] do not appear to match input type(s) [${this.datatypes}].`
         );
     },
+    _producesAcceptableDatatypeAndOptionalness(other) {
+        if (!this.optional && !this.multiple && other.optional) {
```
mvdbeek, line 381: I guess that will work if a tool author properly handles the iteration, but in my tests I used the cat tool in https://toolshed.g2.bx.psu.edu/repository?repository_id=2593fd36ae8011aa which does
```
#echo ' '.join(['"%s"' % $file for $file in $inputs])#
```
which evaluates to `"None"`. I guess the proper way is `#echo ' '.join(['"%s"' % $file for $file in $inputs if $file])#` or make sure that an empty multi-data parameter behaves like `[]`.
The tool form doesn't let you do this and insists on passing a dataset ... thought I can see this being useful.
So I guess this is fine for now and we should see if it is possible to fix this style of iteration on the Galaxy side ?

mvdbeek:
Awesome, this is a big step!

client/galaxy/scripts/mvc/workflow/workflow-terminals.js
jmchilton: The tool form insist on a dataset even if it is marked as optional? I thought I hacked around this years ago for Dan... ugh

mvdbeek: it's the other way round, this code allows connecting optional datasets to required multi-data parameters. I think there is some merit in not allowing that, as the tool form does.

jmchilton: Oh - that is 100% intentional on my part. You should be able to connect two optional data parameters to a required multi-parameter - in case one of them is set, right? The tool form lets you do this - run two tools and pick the output that is produced manually and send it to the next tool 😃. This should be caught at runtime IMO so you can do conditionally things

lib/galaxy/workflow/modules.py
```diff
@@ -626,77 +723,316 @@ class InputDataCollectionModule(InputModule):
...
+
+                when_restrict_static_restrictions = ConditionalWhen()
+                when_restrict_static_restrictions.value = "staticRestrictions"
+                when_restrict_static_restrictions.inputs = OrderedDict()
+
+                when_restrict_static_suggestions = ConditionalWhen()
+                when_restrict_static_suggestions.value = "staticSuggestions"
+                when_restrict_static_suggestions.inputs = OrderedDict()
+
+                # Repeats don't work - so use common separated list for now.
```
guerler, line 921: Why do repeats not work?

jmchilton: They don't - I don't know why exactly. I spent a few hours trying to walk through what was happening with the parameter expansion logic and trying a variety of tweaks and was pretty convinced it wasn't a trivial amount of code away from working.

We could do some cool stuff around allowing users to declare expressions with multiple inputs and outputs if we could get that to work.

PR #9087, opened 2019-12-10
Add a datatype subclass for CAT (Contig Annotation Tool)

jj-umn:
This datatype will be assigned when the cat_prepare tool creates a "database"  in a user history.  The other CAT tools may select that database form the history.


bgruening (approved):
The failing test seems unrelated.

PR #9092, opened 2019-12-11
Make callbacks for tour and config watchers a control task

jdavcs:
Re:
https://github.com/mvdbeek/galaxy/commit/e3f3af91830c1c998134fcf0b9ceb1ad09739a2c#r36356830

@mvdbeek is this correct? Should this be backported to 19.09?

### DIFF
--- a/lib/galaxy/config_watchers.py
+++ b/lib/galaxy/config_watchers.py
@@ -1,7 +1,5 @@
-from functools import partial
 from os.path import dirname
 
-from galaxy.config import reload_config_options
 from galaxy.queue_worker import job_rule_modules
 from galaxy.tools.toolbox.watcher import (
     get_tool_conf_watcher,
@@ -81,11 +79,11 @@ class ConfigWatchers(object):
         if self.app.config.config_file:
             self.core_config_watcher.watch_file(
                 self.app.config.config_file,
-                callback=partial(reload_config_options, self.app.config)
+                callback=lambda path: self.app.queue_worker.send_control_task('reload_core_config')
             )
             self.tour_watcher.watch_directory(
                 self.app.config.tour_config_dir,
-                callback=lambda path: self.app.tour_registry.reload_tour(path)
+                callback=lambda path: self.app.queue_worker.send_control_task('reload_tour', kwargs={'path': path})
             )
         self.active = True
 
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -868,7 +868,7 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
 Configuration = GalaxyAppConfiguration
 
 
-def reload_config_options(current_config, path=None):
+def reload_config_options(current_config):
     """ Reload modified reloadable config options """
     modified_config = read_properties_from_file(current_config.config_file)
     for option in current_config.reloadable_options:
--- a/lib/galaxy/queue_worker.py
+++ b/lib/galaxy/queue_worker.py
@@ -29,6 +29,7 @@ from six.moves import reload_module
 
 import galaxy.queues
 from galaxy import util
+from galaxy.config import reload_config_options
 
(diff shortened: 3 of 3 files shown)


jdavcs:
Unit test failure unrelated.

lib/galaxy/config_watchers.py
```diff
@@ -81,11 +79,11 @@ def start(self):
         if self.app.config.config_file:
             self.core_config_watcher.watch_file(
                 self.app.config.config_file,
-                callback=partial(reload_config_options, self.app.config)
+                callback=lambda path: self.app.queue_worker.send_control_task('reload_core_config')
```
mvdbeek, line 82:
```suggestion
                callback=lambda: self.app.queue_worker.send_control_task('reload_core_config')
```

mvdbeek (commented):
Yep, that looks good and work correctly.

lib/galaxy/config_watchers.py
ic4f: This will raise an error: watcher invokes the callback with an argument: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/util/watcher.py#L175

mvdbeek: Why, you're not using `path` ?

jdavcs: hmm. My comment just disappeared (was editing it after posting). Here it is, just in case:
This will raise an error: watcher invokes the callback with an argument: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/util/watcher.py#L175
(i agree, doesn't look good, but we'd need to refactor watcher most likely).
And no, we don't need path for reloading config options.

mvdbeek: so either I don't know my python, but `lambda path: self.app.queue_worker.send_control_task('reload_core_config')`

is the same as
```
def lambda(path):
    return self.app.queue_worker.send_control_task('reload_core_config')
```
So I don't understand why you want to include `path` in the lambda ?

mvdbeek: The line you quote doesn't look relevant, the path variable there comes from a watchdog event: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/util/watcher.py#L149

jdavcs: Well, you *do* know your python, so I suspect it's just me not looking deep enough. But here's what I get when I test it (enable, then edit galaxy.yml):

```
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/home/sergey/0dev/galaxy/galaxy/current/.venv/local/lib/python2.7/site-packages/watchdog/observers/api.py", line 199, in run
    self.dispatch_events(self.event_queue, self.timeout)
  File "/home/sergey/0dev/galaxy/galaxy/current/.venv/local/lib/python2.7/site-packages/watchdog/observers/api.py", line 368, in dispatch_events
    handler.dispatch(event)
  File "/home/sergey/0dev/galaxy/galaxy/current/.venv/local/lib/python2.7/site-packages/watchdog/events.py", line 322, in dispatch
    self.on_any_event(event)
  File "lib/galaxy/util/watcher.py", line 137, in on_any_event
    self._handle(event)
  File "lib/galaxy/util/watcher.py", line 175, in _handle
    callback(path=path)
TypeError: <lambda>() takes no arguments (1 given)
```
I'm heading to the office, will look into  it once I get there (back online in ~50min)

mvdbeek: Ah, I see. `path` isn't used, but the the event handler passes it. My bad.

mvdbeek (approved):

mvdbeek:
I'll backport this together with https://github.com/galaxyproject/galaxy/pull/9094 and https://github.com/galaxyproject/galaxy/pull/9082

Issue #9085, opened 2019-12-10
Enabling watchers in multiprocess mode causes a segfault on a Mac


jdavcs:
A segmentation fault occurs consistently under the following conditions:
- enable any watcher: tours, config, tools, job_rules
- db: postgres (similar issue w/sqlite but, possibly, another cause)
- OS: macOS Mojave 10.14.5
- tested under Python: 2.7, 3.5, 3.6
- uwsgi processes: > 1 (if 1, then no issue)
- uwsgi master: true (otherwise runtime error)
- galaxy branch: dev

This *does not* happen on Linux.
The issue was first discovered by @dannon  .

The segfault is triggered on calling `observer.start()`, which leads here:
https://github.com/gorakhargosh/watchdog/blob/master/src/watchdog/observers/fsevents.py#L137
which leads here:
https://github.com/gorakhargosh/watchdog/blob/master/src/watchdog_fsevents.c#L337

Here's the relevant traceback:

```
galaxy.web_stack INFO 2019-12-09 21:44:22,018 [p:4189,w:1,m:0] [MainThread] Galaxy server instance 'main.web.1' is running
Starting server in PID 4189.
serving on http://127.0.0.1:8080
galaxy.web_stack INFO 2019-12-09 21:44:22,028 [p:4196,w:2,m:0] [MainThread] Galaxy server instance 'main.web.2' is running
galaxy.model.database_heartbeat DEBUG 2019-12-09 21:44:22,056 [p:4189,w:1,m:0] [database_heartbeart_main.web.1.thread] main.web.1 is config watcher
galaxy.util.watcher DEBUG 2019-12-09 21:44:22,060 [p:4189,w:1,m:0] [database_heartbeart_main.web.1.thread] Watching for changes in directory: /Users/sergey/0dev/galaxy/py3.6.9_dev/config/plugins/tours
galaxy.model.database_heartbeat DEBUG 2019-12-09 21:44:22,063 [p:4196,w:2,m:0] [database_heartbeart_main.web.2.thread] main.web.2 is config watcher
Fatal Python error: Segmentation fault

Current thread 0x000070000dcdf000 (most recent call first):
  File "/Users/sergey/0dev/galaxy/py3.6.9_dev/.venv/lib/python3.6/site-packages/watchdog/observers/fsevents.py", line 138 in run
  File "/Users/sergey/.pyenv/versions/3.6.9/lib/python3.6/threading.py", line 916 in _bootstrap_inner
[...]

PR #9094, opened 2019-12-11
[19.09] Don't watch tool files through watchdog

mvdbeek:
These will be watched more efficiently through the ToolConfigWatcher.
Watching all tool files (which watches their directories) can lead to enormous CPU usage when many tools are installed. Explicitly indicated tool directories (with `<tool_dir dir="dump_dir"/>`) will still be watched through watchdog.

If I activate `watch_tools` (this should only be done in dev anyway) I couldn't get any watchdog events to appear. Forcing  watchdog to use the PollingObserver then led to 100% CPU load. Not sure this is related to https://github.com/galaxyproject/galaxy/issues/9085, but it haven't seen [this traceback](https://github.com/galaxyproject/galaxy/issues/9085#issuecomment-564070010) anymore

### DIFF
--- a/lib/galaxy/tools/toolbox/base.py
+++ b/lib/galaxy/tools/toolbox/base.py
@@ -867,8 +867,6 @@ class AbstractToolBox(Dictifiable, ManagesIntegratedToolPanelMixin):
         if not tool.id.startswith("__"):
             # do not monitor special tools written to tmp directory - no reason
             # to monitor such a large directory.
-            if self._tool_watcher:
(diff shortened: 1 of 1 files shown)


mvdbeek:
Actually this might fix the `test_tool_reload_for_broken_tool ` test ... if instead of the callback we want (https://github.com/mvdbeek/galaxy/blob/a0d373ed43215cfd7358d0ffc98d37936d47aae5/test/unit/tools/test_toolbox.py#L578) we trigger the ToolWatcher callback

jdavcs (approved):

PR #9103, opened 2019-12-12
Merge release_19.09 into dev

mvdbeek:
Not sure I resolved the admin job invocations correctly, can you have a look @dannon or @jmchilton ?


client/galaxy/scripts/components/admin/Jobs.vue
```diff
@@ -129,8 +129,8 @@
                         :value="data.item['job_info']['id']"
                     ></b-form-checkbox>
                 </template>
-                <template v-slot:cell(job_info)="data">
```
nsoranzo, line 132: I didn't seen you also tried to merge 19.09 forward and opened https://github.com/galaxyproject/galaxy/pull/9104 . I resolved the conflict here differently, see https://github.com/galaxyproject/galaxy/pull/9104/files#diff-8fd44636765df0b938c8ff566954e688

lib/galaxy/tool_util/deps/mulled/util.py
```diff
@@ -13,6 +14,9 @@
 
 log = logging.getLogger(__name__)
 
+BUILD_NUMBER_REGEX = re.compile(r'\d+$')
+MULLED_TAG_CACHE = collections.defaultdict(dict)
```
nsoranzo: Another difference, I removed this line since `MULLED_TAG_CACHE` is now unused.

mvdbeek: Ah yeah, I deleted that one now.

client/galaxy/scripts/components/admin/Jobs.vue
mvdbeek: Ah, yeah, comparing that with dannon's PR, I think your variant is correct.

PR #9105, opened 2019-12-12
add option select_type_workflow_threshold (rebase)

jmchilton:
Rebase of #8816, cleaner diff in Github now.

I'm a bit uneasy about this - it doesn't feel like something that should be configurable - but that said I understand why we don't and why we want to.

The simplified workflow run interface sketched out in http://bit.ly/simplified-workflow-ui - would cause dozens of fewer select boxes to render per large workflow - so maybe that would mitigate this problem somewhat.


dannon:
Yeah, down the road when we have either (or both) lighter weight select widgets or simply fewer to display on the run-workflow page we can change the default to always use the 'rich' widget.  This option can be deprecated at that point, but until then this looks good to me.

Thanks @FredericBGA!

jmchilton:
Thanks for the merge @dannon!

PR #9106, opened 2019-12-12
Make failing to load tools from a tool_dir non-fatal.

blankenberg:
Previously, having e.g. a missing directory prevented the remainder of the `tool_conf.xml` file from being read.

```
<tool_dir dir="/path/no/exist/tools" />
```

### DIFF
--- a/lib/galaxy/tools/toolbox/base.py
+++ b/lib/galaxy/tools/toolbox/base.py
@@ -828,19 +828,19 @@ class AbstractToolBox(Dictifiable, ManagesIntegratedToolPanelMixin):
                 return None
 
         tool_loaded = False
-        try:
-            for name in os.listdir(directory):
-                if name.startswith(('.', '_')):
-                    # Very unlikely that we want to load tools from a hidden or private folder
-                    continue
-                child_path = os.path.join(directory, name)
-                if os.path.isdir(child_path) and recursive:
-                    self.__watch_directory(child_path, elems, integrated_elems, load_panel_dict, recursive)
-                elif self._looks_like_a_tool(child_path):
-                    quick_load(child_path, async_load=False)
-                    tool_loaded = True
-        except Exception:
-            log.exception("Failed to read tool directory %s.", directory)
+        if not os.path.isdir(directory):
+            log.error("Failed to read tool directory %s.", directory)
+            return
+        for name in os.listdir(directory):
+            if name.startswith(('.', '_')):
+                # Very unlikely that we want to load tools from a hidden or private folder
+                continue
+            child_path = os.path.join(directory, name)
+            if os.path.isdir(child_path) and recursive:
+                self.__watch_directory(child_path, elems, integrated_elems, load_panel_dict, recursive)
+            elif self._looks_like_a_tool(child_path):
+                quick_load(child_path, async_load=False)
+                tool_loaded = True
         if (tool_loaded or force_watch) and self._tool_watcher:
             self._tool_watcher.watch_directory(directory, quick_load)
 
--- a/lib/galaxy/tools/toolbox/base.py
+++ b/lib/galaxy/tools/toolbox/base.py
@@ -828,16 +828,19 @@ class AbstractToolBox(Dictifiable, ManagesIntegratedToolPanelMixin):
                 return None
 
         tool_loaded = False
-        for name in os.listdir(directory):
-            if name.startswith('.' or '_'):
-                # Very unlikely that we want to load tools from a hidden or private folder
-                continue
-            child_path = os.path.join(directory, name)
-            if os.path.isdir(child_path) and recursive:
-                self.__watch_directory(child_path, elems, integrated_elems, load_panel_dict, recursive)
-            elif self._looks_like_a_tool(child_path):
-                quick_load(child_path, async_load=False)
-                tool_loaded = True
+        try:
+            for name in os.listdir(directory):
+                if name.startswith('.' or '_'):
+                    # Very unlikely that we want to load tools from a hidden or private folder
+                    continue
+                child_path = os.path.join(directory, name)
+                if os.path.isdir(child_path) and recursive:
+                    self.__watch_directory(child_path, elems, integrated_elems, load_panel_dict, recursive)
+                elif self._looks_like_a_tool(child_path):
+                    quick_load(child_path, async_load=False)
+                    tool_loaded = True
+        except Exception:
+            log.exception("Failed to read tool directory %s.", directory)
         if (tool_loaded or force_watch) and self._tool_watcher:
             self._tool_watcher.watch_directory(directory, quick_load)
 
--- a/lib/galaxy/tools/toolbox/base.py
+++ b/lib/galaxy/tools/toolbox/base.py
@@ -830,7 +830,7 @@ class AbstractToolBox(Dictifiable, ManagesIntegratedToolPanelMixin):
         tool_loaded = False
         try:
             for name in os.listdir(directory):
-                if name.startswith('.' or '_'):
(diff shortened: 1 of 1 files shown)


lib/galaxy/tools/toolbox/base.py
```diff
@@ -828,16 +828,19 @@ def quick_load(tool_file, async_load=True):
...
-                continue
-            child_path = os.path.join(directory, name)
-            if os.path.isdir(child_path) and recursive:
-                self.__watch_directory(child_path, elems, integrated_elems, load_panel_dict, recursive)
-            elif self._looks_like_a_tool(child_path):
-                quick_load(child_path, async_load=False)
-                tool_loaded = True
+        try:
+            for name in os.listdir(directory):
+                if name.startswith('.' or '_'):
```
jdavcs: `if name.startswith(('.', '_')):` (tuples work here as of Python 2.5)
Because `'.' or '_'` will evaluate to `'.'`

blankenberg: Great catch :)

lib/galaxy/tools/toolbox/base.py
```diff
@@ -828,16 +828,19 @@ def quick_load(tool_file, async_load=True):
...
+                if name.startswith(('.', '_')):
+                    # Very unlikely that we want to load tools from a hidden or private folder
+                    continue
+                child_path = os.path.join(directory, name)
+                if os.path.isdir(child_path) and recursive:
+                    self.__watch_directory(child_path, elems, integrated_elems, load_panel_dict, recursive)
+                elif self._looks_like_a_tool(child_path):
+                    quick_load(child_path, async_load=False)
+                    tool_loaded = True
+        except Exception:
```
jdavcs: I'm wondering... This Exception will catch *everything* in this block, including the recursive call. It fails silently - so we'll never know if something bad happens elsewhere. If `directory` doesn't exist, it'll fail silently - which, I think, is the intended behavior. However, if something happens here `self._tool_watcher.watch_directory(directory, quick_load)` on a recursive call, we'll never know - it'll be caught by `Exception`. In other words:
```python
def foo(n):
    if n == -1: 
        return 
    try:
        foo(n-1)
    except Exception:
        print('do nothing')
    bar(n)
def bar(n):
    1/n
foo(1)
```
The above will never catch division by zero in `bar`.

If the goal is to not die on missing directory, maybe we could use something like this on line 831?
```python
if not os.path.exists(directory): 
    log.exception(...)
    return
``` 
(I assume that's the desired outcome?)

blankenberg: The `log.exception()` call will provide the traceback:

```
galaxy.tools.toolbox.base ERROR 2019-12-12 17:01:31,348 [p:17931,w:0,m:0] [MainThread] Failed to read tool directory /path/no/exist/tools.
  File "lib/galaxy/tools/toolbox/base.py", line 832, in __watch_directory
    for name in os.listdir(directory):
OSError: [Errno 2] No such file or directory: '/path/no/exist/tools'
```

The goal is to prevent any error here from keeping the rest of the processes to continue.

jdavcs: Oh, OK, then that's fine.

jdavcs (approved):

jmchilton:
What is the use case here prompting this? I'm ``-0`` without further justification - I assume by default most admins would want Galaxy to not startup if a configuration directory isn't present. My intuition is this is typically going to be a typo or misconfiguration that needs to be addressed immediately right?

blankenberg:
> What is the use case here prompting this? I'm `-0` without further justification - I assume by default most admins would want Galaxy to not startup if a configuration directory isn't present. My intuition is this is typically going to be a typo or misconfiguration that needs to be addressed immediately right?

I develop a large amount of Galaxy tools. These tools are spread across multiple repositories and multiple branches. By far the easiest way to do realistic/actual testing/development on these tools is to use a real/permanent Galaxy instance with a tool_conf.xml that points to these under-development tool directories. Sometimes one or more of the referenced tool directories don't exist, because a branch has been switched.

The current behavior is that Galaxy stops loading the tool_conf.xml at the first missing directory, but Galaxy continues to load 'just fine'. Except it isn't just fine, only tools found before the missing entry are loaded, any found after are not. This really does not make any sense, lacks robustness, and is inconsistent with the behavior seen when trying to load a single tool configuration file that doesn't exist. 

```
  <tool file="/path/no/exist/tools/tool.xml" />
  <tool_dir dir="/path/no/exist/tools" />
```

jdavcs:
I think there may be two distinct use cases here: dev, where it's reasonable to ignore missing directories, and production, where failing immediately is the expected mode. So, just an idea: maybe have a config option? Something like `ignore_missing_tool_directories`, set to `false` by default?

blankenberg:
> I think there may be two distinct use cases here: dev, where it's reasonable to ignore missing directories, and production, where failing immediately is the expected mode. So, just an idea: maybe have a config option? Something like `ignore_missing_tool_directories`, set to `false` by default?

I honestly can't think of an actual valid use case for the current behavior, where the failed loading of one set of tools completely prevents separate and independent sets of tools from loading.

Galaxy is still starting up 'fine', but potentially with an abstractly nonsensical set of missing tools.

lib/galaxy/tools/toolbox/base.py
```diff
@@ -828,16 +828,19 @@ def quick_load(tool_file, async_load=True):
...
+            for name in os.listdir(directory):
+                if name.startswith(('.', '_')):
+                    # Very unlikely that we want to load tools from a hidden or private folder
+                    continue
+                child_path = os.path.join(directory, name)
+                if os.path.isdir(child_path) and recursive:
+                    self.__watch_directory(child_path, elems, integrated_elems, load_panel_dict, recursive)
+                elif self._looks_like_a_tool(child_path):
+                    quick_load(child_path, async_load=False)
+                    tool_loaded = True
```
nsoranzo: Shouldn't `tool_loaded` check the return value of `quick_load()` ?

lib/galaxy/tools/toolbox/base.py
nsoranzo: Maybe instead of the try/except, we can use:
```python
        if not os.path.isdir(directory):
            log.error("Failed to read tool directory %s.", directory)
            return
```
before the for loop.

blankenberg:
Came to bump milestone, glad it was already done. Some great discussion here, would be super awesome if one of yinz took over this issue.

jdavcs:
I'm happy to take over, only I'm not sure how.. Do I push directly to @blankenberg 's branch? 
@nsoranzo : I agree with your latest comment. But do you think we should have `log.exception(...)` before `return`?

nsoranzo:
> I'm happy to take over, only I'm not sure how.. Do I push directly to @blankenberg 's branch?

Yep, you should be able to.

> @nsoranzo : I agree with your latest comment. But do you think we should have `log.exception(...)` before `return`?

Good point, I've updated the code in my comment.

jdavcs:
Done.

nsoranzo (approved):

nsoranzo:
@ic4f Would you like to take a look at https://github.com/galaxyproject/galaxy/pull/9106#discussion_r358028297 (not necessarily as part of this PR)?

jdavcs:
Can I merge this after the tests pass? 
~~EDIT: Wait, just saw your comment. Looking.~~

jdavcs:
> @ic4f Would you like to take a look at [#9106 (comment)](https://github.com/galaxyproject/galaxy/pull/9106#discussion_r358028297) (not necessarily as part of this PR)?

Of course! Will handle it as a separate PR.

PR #9108, opened 2019-12-13
Add a test for a tool with `from_working_dir`, `outputs_to_working_directory` and `pulsar`

bgruening:
The combination of:
  * from_work_dir in tools
  * outputs_to_working_directory
  * pulsar

seems to cause problem in our setup. Let's see if the test fails here.


jmchilton:
The docker container uses a shared file system across the Galaxy process and the embedded Pulsar component - so these tests miss a lot of potential kinds of errors and are often green when they shouldn't be.

bgruening:
@jmchilton should I close this one?

xref: https://github.com/galaxyproject/pulsar/issues/193

PR #9115, opened 2019-12-14
Automatically redirect to OIDC provider when only auth option

machristie:
When Galaxy is configured to only use a single OIDC provider for authentication, it is desirable for /login to automatically redirect to that provider's login page instead of requiring the user to select that provider on the Galaxy login page.

This PR changes /login to automatically do this redirection when:

1. enable_oidc is true
2. there is only one OIDC provider in oidc_backends_config.xml
3. there are no other authenticators as defined in auth_config.xml

With this automatic redirection, a problem can occur on logout if the user is still logged into the identity provider. Since the Galaxy logout page redirects to the login page, if the user tries to logout of Galaxy but is still logged into the identity provider the user will end up logged into Galaxy again. The steps are:
* user logs out at /user/logout (menu.js)
* /user/logout -> /login
* /login -> idp login page
* user is still logged in, so idp login page -> Galaxy home page and user is logged in

Hence, the logout logic has been changed to redirect to the home page instead of the login page.

### DIFF
--- a/lib/galaxy/webapps/galaxy/controllers/root.py
+++ b/lib/galaxy/webapps/galaxy/controllers/root.py
@@ -17,6 +17,7 @@ from galaxy import (
     managers,
     web
 )
+from galaxy.auth.util import get_authenticators
 from galaxy.model.item_attrs import UsesAnnotations
 from galaxy.util import (
     FILENAME_VALID_CHARS,
@@ -83,6 +84,17 @@ class RootController(controller.JSAppLauncher, UsesAnnotations):
         """
         User login path for client-side.
         """
+        # directly redirect to oidc provider if 1) enable_oidc is True, 2)
+        # there is only one oidc provider, 3) auth_conf.xml has no authenticators
+        if (trans.app.config.enable_oidc and
+                len(trans.app.config.oidc) == 1 and
+                len(trans.app.auth_manager.authenticators) == 0):
+
+            provider = trans.app.config.oidc[0]
+            success, message, redirect_uri = trans.app.authnz_manager.authenticate(provider, trans)
+            if success:
+                return trans.response.send_redirect(redirect_uri)
+
         return self.template(trans, 'login',
                              redirect=redirect,
                              # an installation may have it's own welcome_url - show it here if they've set that
--- a/lib/galaxy/webapps/galaxy/controllers/root.py
+++ b/lib/galaxy/webapps/galaxy/controllers/root.py
@@ -17,7 +17,6 @@ from galaxy import (
     managers,
     web
 )
-from galaxy.auth.util import get_authenticators
 from galaxy.model.item_attrs import UsesAnnotations
 from galaxy.util import (
     FILENAME_VALID_CHARS,
--- a/client/galaxy/scripts/layout/menu.js
+++ b/client/galaxy/scripts/layout/menu.js
@@ -18,7 +18,7 @@ function logoutClick() {
         if (galaxy.user) {
             galaxy.user.clearSessionStorage();
         }
-        window.top.location.href = `${galaxy.root}login`;
+        window.top.location.href = galaxy.root;
     });
 }
 


lib/galaxy/webapps/galaxy/controllers/root.py
```diff
@@ -83,6 +84,17 @@ def login(self, trans, redirect=None, **kwd):
...
+        # directly redirect to oidc provider if 1) enable_oidc is True, 2)
+        # there is only one oidc provider, 3) auth_conf.xml has no authenticators
+        if (trans.app.config.enable_oidc and
+                len(trans.app.config.oidc) == 1 and
+                len(trans.app.auth_manager.authenticators) == 0):
+
+            provider = trans.app.config.oidc[0]
+            success, message, redirect_uri = trans.app.authnz_manager.authenticate(provider, trans)
+            if success:
+                return trans.response.send_redirect(redirect_uri)
```
VJalili, line 95: this section has overlap with the controller, so I guess it should leverage it to avoid duplication (and possibly diverging logic).

VJalili (changes requested):

lib/galaxy/webapps/galaxy/controllers/root.py
machristie: Hi @VJalili, thanks for reviewing. When you say "this section has overlap with the controller", which controller do you mean? Are you saying this logic should be in the authnz_manager?

VJalili: Sorry I was not clear, [this is the endpoint in the AuthNZ controller](https://github.com/galaxyproject/galaxy/blob/850d483bf260cba969ddb21de2f5404c15e09d1c/lib/galaxy/webapps/galaxy/controllers/authnz.py#L47-L49) that I was referring to. I wonder how we can leverage that controller hence to avoid re-implementing a similar logic in different places. (Having same logic implemented in different place can be problematic when we modify one and possibly forget the other one, therefore, I would recommend we leverage the aforelinked controller for calling the authnz manager.)

nuwang: I think you're right - this logic should probably be moved to the authnz manager, and this login method can call the authnz manager's login instead? It looks to me like this login method should ideally be retired soon anyway and is something that'll have to happen as part of retiring mako and shifting to Vue? However, it looks like the authnz manager doesn't support non-oidc login. https://github.com/galaxyproject/galaxy/blob/850d483bf260cba969ddb21de2f5404c15e09d1c/lib/galaxy/webapps/galaxy/controllers/authnz.py#L43
I was going to suggest removing that line in the authnz manager and if `enable_oidc` is false, it could redirect to the standard login page instead. Then, this method in root.py could call the authnz manager's login method instead. Would that work?
ping @dannon?

dannon: Not sure I follow.  The route @VJalili links above requires a specific provider and handles the handshake, yes, but it doesn't present any of the interface.

We still want a first class route to /login that serves the (already Vue) login entrypoint.

nuwang: Sorry, I hadn't realized that the endpoint Vahid pointed to was actually provider specific, so no, what I said doesn't apply.

dannon: @VJalili it's a worthy consideration and if you'd like to refactor this down the road, we can certainly do that.  Did you mean something like this?
```
        if (trans.app.config.enable_oidc and
                len(trans.app.config.oidc) == 1 and
                len(trans.app.auth_manager.authenticators) == 0):

            return trans.webapp.controllers['authnz'].login(trans, trans.app.config.oidc[0])
```
It doesn't feel to me like that really cleans up a lot, and without trying it it *looks* like you would end up in a weird state potentially without a page to look at, since it returns json?

Anyway, I do think this works for now and I'd like to go ahead and get it in, certainly feel free to follow up down the road though.

VJalili: I guess the snippet @dannon has mentioned could be helpful leveraging the controller instead of re-implementing its logic here, which we may easily forget updating it when making any changes to the controller. That brings me to a second point, would it be useful to implement a unit test for this functionality?

VJalili:
I don't think the conversation was resolved.

PR #9116, opened 2019-12-15
Automatically use extended-base-image if necessary

mvdbeek:
Any explicitly set destination image will still be honored. I think this is the minimal change we can do without replacing the involucro based container building.
Also fixes version comparison in `best_search_result` and adds a github workflow that only tests the mulled utilities.

This could be complimentary to https://github.com/galaxyproject/galaxy/pull/9088 and allow us to containerize more tools. This doesn't fix missing environment variables normally set by conda activate.

### DIFF
--- a/.github/workflows/mulled.yaml
+++ b/.github/workflows/mulled.yaml
@@ -1,0 +1,27 @@
+name: Mulled Unit Tests
+on: [push, pull_request]
+jobs:
+
+  test:
+    name: Test
+    runs-on: ubuntu-18.04
+    strategy:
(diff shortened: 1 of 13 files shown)


mvdbeek:
A passing github workflow run is here: https://github.com/mvdbeek/galaxy/commit/fa1c8db202c2d16d2c4ec73d7005121d5a7247fb/checks?check_suite_id=359125972

jmchilton (approved):

bgruening (approved):
Nice!

PR #9124, opened 2019-12-17
Add admin functionality to histories API

thomcuddihy:
Added the ability for Admins to query all histories using the API.

This was implemented as part of a new tool for managing users for Galaxy Australia, and generating stats/graphs for reporting measures.

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/histories.py
+++ b/lib/galaxy/webapps/galaxy/api/histories.py
@@ -146,11 +146,16 @@ class HistoriesController(BaseAPIController, ExportsHistoryMixin, ImportsHistory
         filters = []
         # support the old default of not-returning/filtering-out deleted histories
         filters += self._get_deleted_filter(deleted, filter_params)
-        # check if user is admin and optional parameter 'all' is True
+        # check if user is admin and get optional parameter 'all' 
         admin = trans.user_is_admin
         all_histories = util.string_as_bool(kwd.get('all', False))
-        if not (admin and all_histories):
-            # users are limited to requesting only their own histories (here)
+        # if parameter 'all' is true, throw exception is not admin
+        # else add current user filter to query (default behaviour)
+        if all_histories:
+            if not admin:
+                message = "Only admins can query all histories"
+                raise exceptions.AdminRequiredException(message)
+        else:
             filters += [self.app.model.History.user == current_user]
         # and any sent in from the query string
         filters += self.filters.parse_filters(filter_params)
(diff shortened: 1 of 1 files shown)


lib/galaxy/webapps/galaxy/api/histories.py
```diff
@@ -144,8 +146,12 @@ def index(self, trans, deleted='False', **kwd):
         filters = []
         # support the old default of not-returning/filtering-out deleted histories
         filters += self._get_deleted_filter(deleted, filter_params)
-        # users are limited to requesting only their own histories (here)
-        filters += [self.app.model.History.user == current_user]
+        # check if user is admin and optional parameter 'all' is True
+        admin = trans.user_is_admin
+        all_histories = util.string_as_bool(kwd.get('all', False))
```
jmchilton, line 150: Can you throw ``galaxy.exceptions.AdminRequiredException`` if all_histories is true but admin is not? I think the new if would read a bit better too if you did that.

lib/galaxy/webapps/galaxy/api/histories.py
```diff
@@ -144,8 +146,17 @@ def index(self, trans, deleted='False', **kwd):
         filters = []
         # support the old default of not-returning/filtering-out deleted histories
         filters += self._get_deleted_filter(deleted, filter_params)
-        # users are limited to requesting only their own histories (here)
-        filters += [self.app.model.History.user == current_user]
+        # check if user is admin and get optional parameter 'all' 
+        admin = trans.user_is_admin
```
jdavcs: Minor suggestion. You only need to check the user's admin status if `all_history` is `True`. So you could move this line inside the `if` stmt; but you could also get rid of it completely and do `if not trans.user_is_admin` instead - which, I think, will read even better.

lib/galaxy/webapps/galaxy/api/histories.py
thomcuddihy: Added. Thanks for the exception tip.

lib/galaxy/webapps/galaxy/api/histories.py
thomcuddihy: Added. Thanks mate

jdavcs: looks great!

martenson:
@galaxybot test this

martenson:
Thank you @thomcuddihy for the contribution!

PR #9127, opened 2019-12-17
assign discovered extension, dbkey and designation for assign_primary=true

bernt-matthias:
I would find this really useful but I guess that this would also render the primary output useless for workflows, or?

- [ ] some if statements could be necessary to check if ext, designation, and dbkey were covered by pattern

### DIFF
--- a/lib/galaxy/job_execution/output_collect.py
+++ b/lib/galaxy/job_execution/output_collect.py
@@ -379,8 +379,17 @@ def collect_primary_datasets(job_context, output, input_ext):
                 # Before I guess pop() would just have thrown an IndexError
                 raise Exception("Problem parsing metadata fields for file %s" % filename)
             designation = fields_match.designation
+            ext = fields_match.ext
+            if ext == "input":
+                ext = input_ext
+            dbkey = fields_match.dbkey
+            if dbkey == INPUT_DBKEY_TOKEN:
+                dbkey = job_context.input_dbkey
             if filename_index == 0 and extra_file_collector.assign_primary_output and output_index == 0:
                 new_outdata_name = fields_match.name or "{} ({})".format(outdata.name, designation)
+                outdata.change_datatype(ext)
+                outdata.dbkey = dbkey
+                outdata.designation = designation
                 outdata.dataset.external_filename = None  # resets filename_override
                 # Move data from temp location to dataset location
                 job_context.object_store.update_from_file(outdata.dataset, file_name=filename, create=True)
@@ -389,12 +398,6 @@ def collect_primary_datasets(job_context, output, input_ext):
             if name not in primary_datasets:
                 primary_datasets[name] = OrderedDict()
             visible = fields_match.visible
-            ext = fields_match.ext
-            if ext == "input":
-                ext = input_ext
-            dbkey = fields_match.dbkey
-            if dbkey == INPUT_DBKEY_TOKEN:
-                dbkey = job_context.input_dbkey
             # Create new primary dataset
             new_primary_name = fields_match.name or "{} ({})".format(outdata.name, designation)
             info = outdata.info
--- a/test/functional/tools/multi_output_assign_primary_ext_dbkey.xml
+++ b/test/functional/tools/multi_output_assign_primary_ext_dbkey.xml
@@ -1,0 +1,31 @@
+<tool id="multi_output_assign_primary" name="multi_output_assign_primary" version="0.1.0">
+  <command>
+    echo "1" > sample1.hg38.tsv;
+    echo "2" > sample2.hg19.csv;
+  </command>
+  <inputs>
+    <param name="num_param" type="integer" value="7" />
+    <param name="input" type="data" />
+  </inputs>
+  <outputs>
+	<data name="simple_dbkey_ext">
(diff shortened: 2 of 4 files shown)


bernt-matthias:
Scenario from here https://github.com/galaxyproject/tools-iuc/pull/2752:

In this scenario there seem to be (at least) 2 options to determine the output's datatype:

1. evaluate the content of the file at runtime and and set the extension of the file accordingly which is then discovered OR
2. on the result of a rather complex (e.g. regular) expression on a text input content. 
3. auto_format="true"  .. question is if this is usable in workflows

I have no idea how to express the 2nd option.

mvdbeek:
> I would find this really useful but I guess that this would also render the primary output useless for workflows, or?

I think that would be fine, you'd have to use a change datatype PJA in any case

> 2\. on the result of a rather complex (e.g. regular) expression on a text input content.

That is an interesting option, can you elaborate how that would look like ?

> 3\. auto_format="true"  .. question is if this is usable in workflows

I think option 1 is preferable if you can determine the correct datatype extension at runtime. The sniffers are not and don't guarantee to be perfect.

The code now looks pretty good to me, can you add a test for this ?

bernt-matthias:
What is a PJA?

> That is an interesting option, can you elaborate how that would look like ?

In the case of that tool the output is xml in most cases except if certain strings occur in the expression input: https://github.com/galaxyproject/tools-iuc/blob/e6ab9fcd2bdb1f10f682fb463006b23034eff60a/tools/xpath/xpath.xml#L21 e.g. `string()` or `text()`. In the latter case the data type would be txt. Based on this one could set a cheetah variable to "xml" / "txt". Then one would need to be able to use the value of the variable in `<data>` tag. Something similar to `format_source`.

I will add a test or reformulate an existing one.

mvdbeek:
Sorry, PJA is PostJobAction, those things you can add to outputs in the workflow editor (change output format, tags, names, etc.).

mvdbeek:
> e.g. `string()` or `text()`. In the latter case the data type would be txt. Based on this one could set a cheetah variable to "xml" / "txt". Then one would need to be able to use the value of the variable in `<data>` tag. Something similar to `format_source`.

I think that would be really nice, but also a lot more complicated. You'd probably need to delay this until the cheetah command section is fully evaluated, and even then I could see potential inconsistency issues if you reference `$output.ext` elsewhere in the command section. You could perhaps limit this to (hidden or not hidden) text parameters that could be selected using sth. like `format_from_text_param`. 
In the future those text parameters could ***maybe*** be javascript expressions.

bernt-matthias:
Here is a simple test note that `<discover_datasets format="txt">`, i.e. the format (txt) is overwritten by the discovered data format

- would be nice to have something like this for metadata, as far as I can recall my experiments the discovered dbkey is not regarded if the input already has a dbkey
- I also got into troubles with multiple data outputs that had discover_datasets

test/functional/tools/samples_tool_conf.xml
```diff
@@ -19,6 +19,7 @@
   <tool file="multi_output.xml" />
   <tool file="multi_output_configured.xml" />
   <tool file="multi_output_assign_primary.xml" />
+  <tool file="multi_output_assign_primary_ext_dbkey.xml" />
```
jmchilton, line 28: Can you add this to the list of tools (``TEST_TOOL_IDS``) tested in test/integration/test_extended_metadata.py.

test/functional/tools/multi_output_assign_primary_ext_dbkey.xml
```diff
@@ -0,0 +1,31 @@
...
+  <command>
+    echo "1" > sample1.hg38.tsv;
+    echo "2" > sample2.hg19.csv;
+  </command>
+  <inputs>
+    <param name="num_param" type="integer" value="7" />
+    <param name="input" type="data" />
+  </inputs>
+  <outputs>
+	<data name="simple_dbkey_ext">
```
nsoranzo: There a few tabs, can you please replace them with spaces?

lib/galaxy/job_execution/output_collect.py
```diff
@@ -378,8 +378,17 @@ def collect_primary_datasets(job_context, output, input_ext):
...
             designation = fields_match.designation
+            ext = fields_match.ext
+            if ext == "input":
+                ext = input_ext
+            dbkey = fields_match.dbkey
+            if dbkey == INPUT_DBKEY_TOKEN:
+                dbkey = job_context.input_dbkey
             if filename_index == 0 and extra_file_collector.assign_primary_output and output_index == 0:
                 new_outdata_name = fields_match.name or "%s (%s)" % (outdata.name, designation)
+                outdata.change_datatype(ext)
```
mvdbeek, line 390: change_datatype does a couple of things we don't need to do when the dataset is new (and that would be skipped after checking with the object store if the file exists).
```suggestion
                outdata.extension = ext
```
Setting up the metadata attributes should be handled on line 443.

mvdbeek (approved):
Very nice, just a minor comment.

bernt-matthias:
@mvdbeek rebasing done. failing tests seem unrelated to me.

mvdbeek:
Aweomse, thank you @bernt-matthias!

jmchilton:
Thanks for this fix @bernt-matthias - looks awesome.

PR #9129, opened 2019-12-17
[WIP] Standing Galaxy Markdown Improvements

jmchilton:
Building on #8893 which I took out of WIP.

So far this:
 - Implements ``job_metrics`` and ``job_parameters`` directives when exporting to PDF.
 - Implement a new ``history_dataset_embedded`` Galaxy Markdown tag (both GUI and PDF export) - it places the dataset in the context of the report/page without visual separation or metadata. I suspect sometimes you want to show the dataset as its own interesting thing, with its own metadata - and sometimes you want to not break the flow of the report.


jmchilton:
These we merged into the final version of #8893.

PR #9136, opened 2019-12-18
Parse brand label option in legacy mako

guerler:
Follow-up: #9062


martenson:
There are also some quirks with old makos which prepend `/` to the brand, so if there is no `display_galaxy_brand` the whole masthead will be just `/ Europe`. 

edit: nevermind, those were ts and reports templates that I saw

PR #9138, opened 2019-12-18
Refactor Workflow Run Form Toward Vue (part 1)

jmchilton:
Move the workflow run landing, header, alert messages, and success handling into Vue from AnalysisRouter.js and tool-form-composite.js.

Somethings are cleaner:

- Invocation status Vue components are embedded directly as intended and not need external mounting from scanning tags in the DOM.
- The whole success page looks a lot cleaner in Vue than embedded in Backbone templates with pieced together conditional log and stuff via string concatenation. 
- Analysis router mounting looks a lot more like other parts of the app.

Some things are a bit messy now:

- Had to introduce a new component with a progress bar in a button to match what we had for backbone. 
- The interaction between this button in Vue and the step rendering logic in Backbone is messy (passing through a callback that takes three parameters). 

The messiness is unfortunate - but there will always be some messiness when bridging Vue and non-Vue code and I think on the whole this is a step in the right direction.

### DIFF
--- a/client/galaxy/scripts/components/Workflow/Run/WorkflowRun.vue
+++ b/client/galaxy/scripts/components/Workflow/Run/WorkflowRun.vue
@@ -6,6 +6,7 @@
         <b-alert v-if="loading" variant="info" show>
             <loading-span message="Loading workflow run data" />
         </b-alert>
+        <workflow-run-success v-else-if="invocations != null" :invocations="invocations" :workflowName="workflowName" />
         <div v-else>
             <div ref="run" class="ui-form-composite">
                 <div class="ui-form-composite-messages mb-4" ref="messages">
@@ -48,6 +49,7 @@ import axios from "axios";
 
 import WaitButton from "components/WaitButton";
 import LoadingSpan from "components/LoadingSpan";
+import WorkflowRunSuccess from "./WorkflowRunSuccess";
 import { getAppRoot } from "onload";
 import ToolFormComposite from "mvc/tool/tool-form-composite";
 import { errorMessageAsString } from "utils/simple-error";
@@ -55,7 +57,8 @@ import { errorMessageAsString } from "utils/simple-error";
 export default {
     components: {
         LoadingSpan,
-        WaitButton
+        WaitButton,
+        WorkflowRunSuccess
     },
     props: {
         workflowId: { type: String }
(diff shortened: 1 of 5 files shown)


client/galaxy/scripts/components/Workflow/Run/WorkflowRun.vue
```diff
@@ -0,0 +1,118 @@
...
+                this.workflowName = runData.name;
+                this.loading = false;
+                this.$nextTick(() => {
+                    const el = this.$refs["form"];
+                    const formProps = {
+                        el,
+                        setRunButtonStatus: this.setRunButtonStatus,
+                        handleInvocations: this.handleInvocations
+                    };
+                    this.runForm = new ToolFormComposite.View(_.extend(runData, formProps));
```
guerler: I suggest to use Object.assign() here to avoid using underscore.

client/galaxy/scripts/components/Workflow/Run/WorkflowRunSuccess.vue
```diff
@@ -0,0 +1,110 @@
...
+                <a :href="historyTarget">Switch to that history now</a>.
+            </p>
+            <p v-else>
+                You can check the status of queued jobs and view the resulting data by refreshing the History pane, if
+                this has not already happened automatically.
+            </p>
+        </div>
+        <workflow-invocation-state
+            v-for="invocation in invocations"
+            v-bind:key="invocation.id"
```
guerler: Is equal to :key, moving forward we should decide if we want to omit v-bind or not.

dannon: I'd vote for standardizing to the shorthand syntax everywhere.  `:key` here, `@click` for click actions, and so on.  (that said, this is super easy to enforce and swap with linting and autofix, and I don't think it has to happen with this PR -- I'll look into that now)

guerler:
I noticed that the header section (with the run button) scrolls out of the screen, when scrolling through the steps. This is probably related to some mismatch in the flex description. Also the button check icon class name is missing/undefined in the dom markup. Other than that this looks good to me. Thank you so much for working on this.

jmchilton:
I believe I've addressed all the review comments here. Thanks for the catches - a lot of important little things I wouldn't have noticed!

PR #9139, opened 2019-12-18
make more tool parsing errors contain id

bernt-matthias:
When working on a set of tools its good to know which tool creates the trouble.

### DIFF
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -922,7 +922,7 @@ class Tool(Dictifiable):
                     self.__tests = parse_tests(self, tests_source)
                 except Exception:
                     self.__tests = None
-                    log.exception("Failed to parse tool tests")
+                    log.exception("Failed to parse tool tests for tool '%s'", self.id)
             else:
                 self.__tests = None
             self.__tests_populated = True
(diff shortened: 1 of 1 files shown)


lib/galaxy/tools/__init__.py
```diff
@@ -1129,7 +1129,7 @@ def parse_input_elem(self, page_source, enctypes, context=None):
                 # Use float instead of int so that 'inf' can be used for no max
                 group.max = float(input_source.get("max", "inf"))
                 assert group.min <= group.max, \
-                    ValueError("Min repeat count must be less-than-or-equal to the max.")
+                    ValueError("Tool with id '%s': min repeat count must be less-than-or-equal to the max.")
```
bgruening: you need to include here: self.id

lib/galaxy/tools/__init__.py
```diff
@@ -1174,10 +1174,10 @@ def parse_input_elem(self, page_source, enctypes, context=None):
                         try:
                             possible_cases.remove(case.value)
                         except Exception:
-                            log.debug("Tool %s: a when tag has been defined for '%s (%s) --> %s', but does not appear to be selectable." %
+                            log.debug("Tool with id '%s': a when tag has been defined for '%s (%s) --> %s', but does not appear to be selectable." %
```
bgruening: here is also self.id needed

bgruening (approved):

bgruening:
Nice thanks.

Issue #3208, opened 2016-11-22
Enhance admin UI to allow manual user activation.


martenson:
To circumnavigate issues with unreliable email providers that eat the activation emails.

ping @jennaj

jmchilton:
Hey @ic4f - how does this sound?

jdavcs:
I'm on it.

martenson:
for bonus points, xref: https://github.com/galaxyproject/galaxy/issues/6981

jmchilton:
Amazing and awesome, thanks so much @ic4f !

dannon:
Added in #9143

PR #9143, opened 2019-12-18
Enable manual user activation

jdavcs:
Re #3208
Assumptions: confirmation not needed; single user activation only.
Refactoring: move user activation into user manager; update user controller.
TODO: selenium test (ideally, I think).

### DIFF
--- a/lib/galaxy/managers/users.py
+++ b/lib/galaxy/managers/users.py
@@ -522,6 +522,11 @@ class UserManager(base.ModelManager, deletable.PurgableManagerMixin):
                 log.exception('Subscribing to the mailing list has failed.')
                 return "Subscribing to the mailing list has failed."
 
+    def activate(self, trans, user):
+        user.active = True
+        trans.sa_session.add(user)
+        trans.sa_session.flush()
+
 
 class UserSerializer(base.ModelSerializer, deletable.PurgableSerializerMixin):
     model_manager_class = UserManager
--- a/lib/galaxy/managers/users.py
+++ b/lib/galaxy/managers/users.py
@@ -522,10 +522,10 @@ class UserManager(base.ModelManager, deletable.PurgableManagerMixin):
                 log.exception('Subscribing to the mailing list has failed.')
                 return "Subscribing to the mailing list has failed."
 
-    def activate(self, trans, user):
+    def activate(self, user):
         user.active = True
-        trans.sa_session.add(user)
-        trans.sa_session.flush()
+        self.session().add(user)
+        self.session().flush()
 
 
 class UserSerializer(base.ModelSerializer, deletable.PurgableSerializerMixin):
--- a/lib/galaxy/webapps/galaxy/controllers/admin.py
+++ b/lib/galaxy/webapps/galaxy/controllers/admin.py
@@ -153,6 +153,10 @@ class UserListGrid(grids.Grid):
                             allow_multiple=False),
         grids.GridOperation("Generate New API Key",
                             allow_multiple=False,
+                            async_compatible=True),
+        grids.GridOperation("Activate User",
+                            condition=(lambda item: not item.active),
+                            allow_multiple=False,
                             async_compatible=True)
     ]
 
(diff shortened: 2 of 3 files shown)


nsoranzo (approved):

jdavcs:
Afterthoughts. 
1. The activation button is displayed regardless of the value of `app.config.user_activation_on`. Rationale: option might have been on, then turned off, with account(s) still needing activation.
2. I did not add a "deactivate user" option because I think it's not necessary. If, however, it could be helpful, pls let me know. Same for activating multiple users.

martenson:
> Rationale: option might have been on, then turned off, with account(s) still needing activation.

If the activation is turned off, the flag is never checked -> I think the button should be hidden.


> I did not add a "deactivate user"

+1

martenson:
I don't see the second commit in the list or in the files diff. 😕

jdavcs:
> I don't see the second commit in the list or in the files diff. 😕

What second commit?

martenson:
Oh that was just a reference, nevermind. 🤣

jdavcs:
> If the activation is turned off, the flag is never checked -> I think the button should be hidden.

I was about to agree.. But then I thought about it; I don't think we should disable that button if the flag is not set. (1) If the feature is disabled, there will be no new accounts that are not activated - so the button will not be displayed for those. It will only be displayed for non-activate accounts, and non-active accounts can only be possible if the feature (a) is enabled, or (b) has been enabled previously. And in the latter, I think it's convenient to have a button to cleanup after unsetting the flag. And (2): the flag does not disable user *activation status* - it disables the *policy* that requires that users must confirm their email address, after which the system activates their account. User status exists regardless of that policy. The button is an admin override for that status. What do you think? (also, I'm not a galaxy admin, so I'm only guessing here)

martenson:
I followed your rationale but suspect you might be overthinking it. I don't see a benefit in displaying control elements for features that are explicitly turned off.

jdavcs:
> I followed your rationale but suspect you might be overthinking it. I don't see a benefit in displaying control elements for features that are explicitly turned off.

I might be. OK, I'll figure out how to make that change.

jdavcs:
I can't see a better way to do this. In general, passing the app object may be useful in other scenarios down the road: the app holds the config, and grid elements may need to behave differently based on config settings. @martenson ? @nsoranzo : this has changed considerably since your approval.
EDIT: ping @dannon too - just noticed your reaction)

lib/galaxy/managers/users.py
```diff
@@ -522,6 +522,11 @@ def send_subscription_email(self, email):
                 log.exception('Subscribing to the mailing list has failed.')
                 return "Subscribing to the mailing list has failed."
 
+    def activate(self, trans, user):
```
dannon: We don't need trans in this function if you use `self.session()`

dannon (changes requested):
I would not do this like this.

lib/galaxy/webapps/galaxy/controllers/admin.py
```diff
@@ -136,22 +136,26 @@ def get_value(self, trans, grid, user):
...
                             target="top"),
         grids.GridOperation("Recalculate Disk Usage",
-                            condition=(lambda item: not item.deleted),
+                            condition=(lambda item, app: not item.deleted),
                             allow_multiple=False),
         grids.GridOperation("Generate New API Key",
+                            allow_multiple=False,
+                            async_compatible=True),
+        grids.GridOperation("Activate User",
+                            condition=(lambda item, app: app.config.user_activation_on and not item.active),
```
dannon: I would suggest not adding the option in this list, swapping the lambda back to a single argument without `app`, and then just conditionally add the GridOperation when the grid is initialized, based on the config at hand.

dannon:
@ic4f Sorry, github wouldn't accept the review comment submission, and now it's attached all wonky.  Let me know if the suggestion isn't clear and I can get an example together

jdavcs:
@dannon - no worries! I don't see the wonky :) thing, and your comments make sense. I wasn't at all sure adding the arg was the best/only possible way. Will change it. Thanks for reviewing!

dannon:
@ic4f Ok, cool, just let me know if I can help :)  And, thanks for working on the suggestions!

jdavcs:
@dannon Updated. Thanks - now I see it's done that way for other operations too.

dannon:
@ic4f Much better, thank you.

martenson:
Thanks @ic4f !

PR #9144, opened 2019-12-19
[19.05] Restore returning "None" when casting unselected multiselect param to str

nsoranzo:
Broken in commit 7d3b9c53bbabd4f780f8d051717d8d9c0008267d .

Before the above-mentioned commit, when a `<param name="blah" type="select" multiple="true"/>` was left unselected, its value was `None`, casted to string as `"None"`.
Tool authors would check if the user has selected a value with something like:

```
#if str($blah) != 'None':
```

After the above-mentioned commit, the value of the unselected param is `[]`, casted to string as `""`, which breaks tools using the check above.
This restores the previous behaviour.

Seen in pull request:

https://github.com/galaxyproject/tools-iuc/pull/2691

### DIFF
--- a/lib/galaxy/tools/parameters/basic.py
+++ b/lib/galaxy/tools/parameters/basic.py
@@ -903,7 +903,7 @@ class SelectToolParameter(ToolParameter):
             return value
 
     def to_param_dict_string(self, value, other_values={}):
-        if value is None:
+        if value in (None, []):
             return "None"
         if isinstance(value, list):
             if not self.multiple:
--- a/test/unit/tools/test_wrappers.py
+++ b/test/unit/tools/test_wrappers.py
@@ -42,9 +42,29 @@ def test_select_wrapper_simple_options(tool):
     </param>''')
     parameter = SelectToolParameter(tool, xml)
     wrapper = SelectToolParameterWrapper(parameter, "x")
-    assert str(wrapper) == "x"
     assert wrapper.name == "blah"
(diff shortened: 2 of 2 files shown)


qiagu:
Cool.
I also notice tools were broken due to the change. Thanks for changing back. @nsoranzo

mvdbeek (approved):
☹️ I also noticed another case for optional integers, which fails with
```
2019-12-18 22:30:00,408 ERROR [galaxy.jobs.runners] (74) Failure preparing job
  File "/tmp/tmpm_q8jx_d/galaxy-dev/lib/galaxy/util/__init__.py", line 1039, in unicodify
    value = text_type(value)
  File "/home/runner/.planemo/gx_venv_3.7_release_20.01/lib/python3.7/site-packages/Cheetah/Template.py", line 1053, in __unicode__
    return getattr(self, mainMethName)()
  File "cheetah_DynamicallyCompiledCheetahTemplate_1576708200_402827_81563.py", line 166, in respond
ValueError: invalid literal for int() with base 10: ''
```

mvdbeek (approved):
Nevermind, this fixes that case too, it was guarded by a `!= 'None'`

bernt-matthias (approved):
Looks good to me

wm75:
Confirmed locally: this is making the samtools_view tests  pass again :fireworks:
Thanks a lot @nsoranzo and @bernt-matthias for tracking this down and fixing it!

Issue #9111, opened 2019-12-14
Simplified Workflow Run Component


jmchilton:
This is the second of two pages we outlined in http://bit.ly/simplified-workflow-ui - it would be a simplified workflow run component.

For the purposes of this issue - lets call a best practice workflow one with "no disconnected runtime inputs" - i.e. all inputs are actually annotated input modules (data_input, data_collection_input, or parameter_input modules). These have been improving rapidly and #9086 will address a number of remaining issues that might have previously forced a workflow to use runtime inputs instead of inputs (inputs will now be connectable to select parameters, input files can be optional, etc..). 

By requiring best practice workflows (we can validate this server side before rendering the form), we believe we can and should focus on providing a very simplified interface by default that just exposes workflow inputs, and not individual steps. The document above has a longer discussion of this.

If only inputs are being sent to the GUI and rendered - the workflow run should render a lot faster and the API call to run the workflow form can just send the inputs (no need to send the full tool state for every step) which should be more robust long term and is how we should encourage users to use the workflow API.

An “advanced” button to switch out of this simplified mode and to the full workflow run form with all steps could be exposed for typical Galaxy instances. Clinical applications, or any server where the maintainers decide additional workflow tweaking should be prohibited, would have the option to remove the “advanced mode” (could also be role-based).

hexylena:
Is this complete with #9151?

mvdbeek:
I'm gonna say yes ✅   ... I don't think we're missing anything from the description. Thanks so much @jmchilton!

PR #9147, opened 2019-12-19
Refactor Workflow Run Form Toward Vue (part 2)

jmchilton:
Builds on #9138. New commits include:

- Breaking out the actual form (below the run landing header) into its own component so we can substitute an alternative component for #9111. 
- Breaks a bunch of workflow run request parsing logic out of the backbone view an into a plain old JS object that I'm calling a model for a lack of a better term. This should allow the logic to be shared more easily between this form and the simplified one for #9111. 
- Introduces a test case that mocks out the run request logic with a real request from usegalaxy.org and ensures Vue delegates the request properly and the "model" parses it correctly.
- Implement service layer for workflow runs - per advice of @guerler.

### DIFF
--- a/client/galaxy/scripts/components/Workflow/Run/WorkflowRun.test.js
+++ b/client/galaxy/scripts/components/Workflow/Run/WorkflowRun.test.js
@@ -1,0 +1,52 @@
+import MockAdapter from "axios-mock-adapter";
+import axios from "axios";
+import { __RewireAPI__ as rewire } from "./WorkflowRun";
+import WorkflowRun from "./WorkflowRun.vue";
+import { mount, createLocalVue } from "@vue/test-utils";
+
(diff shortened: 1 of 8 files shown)


jmchilton:
@galaxybot test this

jmchilton:
Rebased on top of the prettier changes. 

For future @jmchilton's reference - repeat cycle for each conflict:

```
git checkout --theirs -- galaxy/scripts ; yarn prettier ; git add galaxy/scripts ; git commit
```

Backup of last pre-rebase branch at https://github.com/jmchilton/galaxy/tree/wfrun2vue_prerebase.

dannon:
Works well in my testing, refactoring looks good.  I think we're probably want to continue to move most of the service layer stuff (not a whole lot there for this particular PR, obv) into Vuex, but that won't hold up anything in terms of merging this.

Issue #9148, opened 2019-12-19
Tool parameter casting issues


mvdbeek:
This includes https://github.com/galaxyproject/galaxy/pull/9144, seen in test 3 of bowtie2:
```
Traceback (most recent call last):
  File "/Users/mvandenb/src/galaxy/lib/galaxy/util/__init__.py", line 1039, in unicodify
    value = text_type(value)
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.6/site-packages/Cheetah/Template.py", line 1053, in __unicode__
    return getattr(self, mainMethName)()
  File "cheetah_DynamicallyCompiledCheetahTemplate_1576771185_4247239_70901.py", line 680, in respond
  File "/Users/mvandenb/src/galaxy/lib/galaxy/tools/wrappers.py", line 94, in __ne__
    return not self == other
  File "/Users/mvandenb/src/galaxy/lib/galaxy/tools/wrappers.py", line 91, in __eq__
    return self._get_cast_value(other) == other
  File "/Users/mvandenb/src/galaxy/lib/galaxy/tools/wrappers.py", line 88, in _get_cast_value
    return cast.get(self.input.type, str)(self)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/tools/wrappers.py", line 117, in __int__
    return int(float(self))
  File "/Users/mvandenb/src/galaxy/lib/galaxy/tools/wrappers.py", line 120, in __float__
    return float(str(self))
ValueError: could not convert string to float:
```

mvdbeek:
```
diff --git a/lib/galaxy/tools/wrappers.py b/lib/galaxy/tools/wrappers.py
index 3f1b3eabf8..75f81390ba 100644
--- a/lib/galaxy/tools/wrappers.py
+++ b/lib/galaxy/tools/wrappers.py
@@ -77,7 +77,7 @@ class InputValueWrapper(ToolParameterValueWrapper):
         if self.input.type == 'boolean' and isinstance(other, string_types):
             return str(self)
         # For backward compatibility, allow `$wrapper != ""` for optional non-text param
-        if self.input.optional and self.value is None and isinstance(other, string_types):
+        if self.input.optional and self.value is None and self.input_type != 'boolean':
             return str(self)
         cast = {
             'text': str,
```
would work ... since neither int or float could be cast if the value is None anyway.

PR #9149, opened 2019-12-19
[19.05] Fix optional unset integer and float params

mvdbeek:
Neither of them can be cast to int/float respectively.
Fixes test 3 in bowtie2, which errors with:
```
2019-12-19 17:11:34,022 ERROR [galaxy.jobs.runners] (12) Failure preparing job
  File "/Users/mvandenb/src/galaxy/lib/galaxy/util/__init__.py", line 1039, in unicodify
    value = text_type(value)
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.6/site-packages/Cheetah/Template.py", line 1053, in __unicode__
    return getattr(self, mainMethName)()
  File "cheetah_DynamicallyCompiledCheetahTemplate_1576771894_0091798_17143.py", line 680, in respond
  File "/Users/mvandenb/src/galaxy/lib/galaxy/tools/wrappers.py", line 94, in __ne__
    return not self == other
  File "/Users/mvandenb/src/galaxy/lib/galaxy/tools/wrappers.py", line 91, in __eq__
    return self._get_cast_value(other) == other
  File "/Users/mvandenb/src/galaxy/lib/galaxy/tools/wrappers.py", line 88, in _get_cast_value
    return cast.get(self.input.type, str)(self)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/tools/wrappers.py", line 117, in __int__
    return int(float(self))
  File "/Users/mvandenb/src/galaxy/lib/galaxy/tools/wrappers.py", line 120, in __float__
    return float(str(self))
ValueError: could not convert string to float:
```

Fixes https://github.com/galaxyproject/galaxy/issues/9148

### DIFF
--- a/lib/galaxy/tools/wrappers.py
+++ b/lib/galaxy/tools/wrappers.py
@@ -83,7 +83,7 @@ class InputValueWrapper(ToolParameterValueWrapper):
         if self.input.type == 'boolean' and isinstance(other, string_types):
             return str(self)
         # For backward compatibility, allow `$wrapper != ""` for optional non-text param
-        if self.input.optional and self.value is None and isinstance(other, string_types):
+        if self.input.optional and self.value is None and self.input.type != 'boolean':
             return str(self)
         cast = {
             'text': str,
--- a/test/unit/tools/test_wrappers.py
+++ b/test/unit/tools/test_wrappers.py
@@ -108,15 +108,19 @@ def test_raw_object_wrapper():
     assert not false_wrapper
 
 
-def valuewrapper(tool, value, paramtype):
+def valuewrapper(tool, value, paramtype, optional=False):
     if paramtype == "integer":
-        parameter = IntegerToolParameter(tool, XML('<param name="blah" type="integer" value="10" min="0" />'))
+        optional = 'optional="true"' if optional else 'value="10"'
+        parameter = IntegerToolParameter(tool, XML('<param name="blah" type="integer" %s min="0" />' % optional))
     elif paramtype == "text":
-        parameter = TextToolParameter(tool, XML('<param name="blah" type="text" value="foo"/>'))
+        optional = 'optional="true"' if optional else 'value="foo"'
+        parameter = TextToolParameter(tool, XML('<param name="blah" type="text" %s/>' % optional))
     elif paramtype == "float":
-        parameter = FloatToolParameter(tool, XML('<param name="blah" type="float" value="10.0"/>'))
+        optional = 'optional="true"' if optional else 'value="10.0"'
+        parameter = FloatToolParameter(tool, XML('<param name="blah" type="float" %s/>' % optional))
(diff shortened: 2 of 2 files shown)


jmchilton:
Re-running that selenium test - seems potentially related.

mvdbeek:
Seems to have passed now 🎉

mvdbeek:
but we can hold off until https://github.com/mvdbeek/tools-iuc/runs/356811399 is done, that should give us an idea about any new regressions

mvdbeek:
No regressions and 6 more tests fixed, I think this should be good to go.

jmchilton:
Amazing!!! Thanks so much.

lib/galaxy/tools/wrappers.py
```diff
@@ -83,7 +83,7 @@ def _get_cast_value(self, other):
         if self.input.type == 'boolean' and isinstance(other, string_types):
             return str(self)
         # For backward compatibility, allow `$wrapper != ""` for optional non-text param
-        if self.input.optional and self.value is None and isinstance(other, string_types):
+        if self.input.optional and self.value is None and self.input.type != 'boolean':
             return str(self)
```
nsoranzo, line 87: What about?
```python
        if self.input.optional and self.value is None:
            if isinstance(other, string_types):
                return str(self)
            else:
                return None
```

wm75: Seems reasonable. At least, `None` seems to be expected by the bowtie2 and bwa-mem wrappers. The current fix only works sort of by chance because those wrappers do this with an optional integer param:

```
#if $rg_param("PI") != None
    #set $rg_pi = str($rg_param("PI"))
#else 
    #set $rg_pi = ''
#end if
```

Seems that today that whole logic is superfluous and could be replaced with just:
`#set $rg_pi = str($rg_param("PI"))`
but these wrappers seem to document that, historically, the expectation was that the comparison should be made against `None` not `''`.

mvdbeek: I'll give it a try.

nsoranzo: @wm75 Exactly, in 19.01 and earlier the bowtie2 code above would never reach the `else` branch because `$rg_param("PI")` is evaluated as an object.
With the recent changes in 19.05, the optional param is instead casted as `int(float(str(value)))`, which fails because `str(value)` evaluates to `""` and `float("")` raises the exception in the PR description.
I feel that evaluating an empty optional param to `None` makes more sense than `""`, except when comparing against a string, and can even simplify some convoluted syntax we have to use at the moment.

mvdbeek: I agree, and also ran the tests against tools-iuc, it seems we've now eliminated all cheetah errors (with or without the @nsoranzo's patch). I'll open a PR with the patch.

PR #9151, opened 2019-12-19
Simplified Workflow Run Form

jmchilton:
Builds on refactoring the workflow run form landing and orchestration into Vue (#9147).

Implements #9111 - part of https://github.com/orgs/galaxyproject/projects/47 and outlined in http://bit.ly/simplified-workflow-ui.

## What is in the PR:

- Add new Galaxy configuration option - ``simplified_workflow_run_ui`` set to ``off`` by default.
- If instead the admin has set this parameter to ``prefer``, on the workflow run page scan the workflow run request and if three conditions are met show a simplified tool form. These conditions are:
  - No workflow "resource parameters" - pretty niche, not documented, I don't think anyone outside of Canada is using them and even them I'm not sure if they ever got to production.
  - No "open" tools (i.e. tools with disconnected runtime inputs).
  - No "replacement parameters" defined outside PJAs. I'm calling #{} inside PJA and ${} inside tool steps both "replacement parameters" because the user populates them the same way in the GUI. The API only handles them inside the context of the PJA - it seems the GUI is responsible for replacing them at runtime in the traditional form.
- The simplified workflow form:
   - Drops tool and subworkflow steps from rendering all together and instead just renders the inputs. These are not rendered as separate "steps" or forms - but as just different inputs the same clean, simple form (more the like tool GUI). Labels (or step indexes as a fallback) are used at the input name, step annotations are used as the input help text.
   - Simplify the workflow request made by the client - send only replacement parameters and inputs - do not serialize tool state for each module. I think this makes what we are tracking more structured and should be more performant as well. Prevents storing HDA IDs in unstructured JSON blobs in the database as well.
   - Drops history target option to simplify the UI - ``simplified_workflow_run_ui_target_history`` can be set to either 'current' (default) or 'new' to adjust this parameter Galaxy-wide for the simplified form.
   - Drops job caching option to simplify the UI - ``simplified_workflow_run_ui_job_cache`` can be set to 'off' (default) or 'on' to change this parameter Galaxy-wide for the simplified form.
   - Drops resource parameters - we've already verified there are none for simplified workflows.

What is in the PR:

## Screenshots

Original Form:

Screen Shot 2019-12-19 at 4 11 11 PM

Simplified Form:

Screen Shot 2019-12-19 at 4 10 59 PM

Clicking to open advanced form from simplified one:

inaction

### DIFF
--- a/client/src/components/Workflow/Run/WorkflowRunFormSimple.vue
+++ b/client/src/components/Workflow/Run/WorkflowRunFormSimple.vue
@@ -1,0 +1,114 @@
+<template>
+    <div ref="form"></div>
+</template>
+
+<script>
+import Form from "mvc/form/form-view";
+import { invokeWorkflow } from "./services";
+
+export default {
+    props: {
+        model: {
+            type: Object,
+            required: true,
+        },
+        setRunButtonStatus: {
+            type: Function,
+            required: true,
+        },
+        targetHistory: {
+            type: String,
+            default: "current",
+        },
+        useJobCache: {
+            type: Boolean,
+            default: false,
+        },
+    },
+    data() {
+        return {
+            form: null,
+            inputTypes: {},
+        };
+    },
+    computed: {},
+    created() {
+        this.form = new Form(this.formDefinition());
+        this.$nextTick(() => {
+            const el = this.$refs["form"];
+            el.appendChild(this.form.$el[0]);
+        });
+    },
+    methods: {
+        formDefinition() {
+            const inputs = [];
+
+            // Add workflow parameters.
+            Object.values(this.model.wpInputs).forEach((input) => {
+                const inputCopy = Object.assign({}, input);
+                // do we want to keep the color if we're not showing steps?
+                inputCopy.color = undefined;
+                inputs.push(inputCopy);
+                this.inputTypes[inputCopy.name] = "replacement_parameter";
+            });
+
+            // Add actual input modules.
+            this.model.steps.forEach((step, i) => {
+                const is_input =
+                    ["data_input", "data_collection_input", "parameter_input"].indexOf(step.step_type) != -1;
+                if (!is_input) {
+                    return;
+                }
+
+                const stepName = new String(step.step_index);
+                const stepLabel = step.step_label || new String(step.step_index + 1);
+                const help = step.annotation;
+                const longFormInput = step.inputs[0];
+                const stepAsInput = Object.assign({}, longFormInput, { name: stepName, help: help, label: stepLabel });
+                // disable collection mapping...
+                stepAsInput.flavor = "module";
+                inputs.push(stepAsInput);
+                this.inputTypes[stepName] = step.step_type;
+            });
+
+            const def = {
+                inputs: inputs,
+            };
+            return def;
+        },
+        execute() {
+            const replacementParams = {};
+            const inputs = {};
+            const formData = this.form.data.create();
+            for (const inputName in formData) {
+                const value = formData[inputName];
+                const inputType = this.inputTypes[inputName];
+                if (inputType == "replacement_parameter") {
+                    replacementParams[inputName] = value;
+                } else if (inputType == "data_input" || inputType == "data_collection_input") {
+                    inputs[inputName] = value;
+                }
+            }
+            const data = {
+                replacement_dict: replacementParams,
+                inputs: inputs,
+                inputs_by: "step_index",
+                batch: true,
+                use_cached_job: this.useJobCache,
+            };
+            if (this.targetHistory == "current") {
+                data.history_id = this.model.historyId;
+            } else {
+                data.new_history_name = this.model.name;
+            }
+            invokeWorkflow(this.model.workflowId, data)
+                .then((invocations) => {
+                    this.$emit("submissionSuccess", invocations);
+                })
+                .catch((error) => {
+                    console.log(error);
+                });
+        },
+    },
+};
+</script>
--- a/lib/galaxy/webapps/galaxy/config_schema.yml
+++ b/lib/galaxy/webapps/galaxy/config_schema.yml
@@ -2436,6 +2436,38 @@ mapping:
           When false, the most recently added compatible item in the history will
           be used for each "Set at Runtime" input, independent of others in the workflow.
 
+      simplified_workflow_run_ui:
+        type: str
+        default: 'prefer'
+        enum: ['off', 'prefer']
+        required: false
+        desc: |
+          If set to 'off' by default, always use the traditional workflow form that renders
+          all steps in the GUI and serializes the tool state of all steps during
+          invocation. Set to 'prefer' to default to a simplified workflow UI that
+          only renders the inputs if possible (the workflow must have no disconnected
+          runtime inputs and not replacement parameters within tool steps). In the
+          future 'force' may be added an option for Galaskio-style servers that should
+          only render simplified workflows.
+
+      simplified_workflow_run_ui_target_history:
+        type: str
+        default: 'current'
+        enum: ['current', 'new']
+        required: false
+        desc: |
+          When the simplified workflow run form is rendered, should the invocation outputs
+          be sent to the 'current' history or a 'new' history.
+
+      simplified_workflow_run_ui_job_cache:
+        type: str
+        default: 'off'
+        enum: ['on', 'off']
+        required: false
+        desc: |
+          When the simplified workflow run form is rendered, should the invocation use job
+          caching. This isn't a boolean so an option for 'show-selection' can be added later.
+
       myexperiment_target_url:
         type: str
         default: www.myexperiment.org:80
--- a/client/src/components/Workflow/Run/WorkflowRun.vue
+++ b/client/src/components/Workflow/Run/WorkflowRun.vue
@@ -44,9 +44,22 @@
                 <workflow-run-form
                     ref="runform"
                     :model="model"
+                    v-if="!simpleForm"
                     :set-run-button-status="setRunButtonStatus"
                     @submissionSuccess="handleInvocations"
                 />
+                <div v-else>
+                    <workflow-run-form-simple
+                        ref="runform"
+                        :model="model"
+                        :set-run-button-status="setRunButtonStatus"
+                        :target-history="simpleFormTargetHistory"
+                        :use-job-cache="simpleFormUseJobCache"
+                        @submissionSuccess="handleInvocations"
+                    />
+                    <!-- Options to default one way or the other, disable if admins want, etc.. -->
+                    <a href="#" @click="showAdvanced">Expand to full workflow form.</a>
+                </div>
             </div>
         </span>
     </span>
@@ -58,6 +71,7 @@ import WaitButton from "components/WaitButton";
 import LoadingSpan from "components/LoadingSpan";
 import WorkflowRunSuccess from "./WorkflowRunSuccess";
 import WorkflowRunForm from "./WorkflowRunForm";
+import WorkflowRunFormSimple from "./WorkflowRunFormSimple";
 import { WorkflowRunModel } from "./model.js";
 import { errorMessageAsString } from "utils/simple-error";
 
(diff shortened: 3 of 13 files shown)


nsoranzo:
The cleanliness and simplicity of the UI is amazing, thanks @jmchilton !
@TKlingstrom will love it!

client/galaxy/scripts/components/WaitButton.vue
```diff
@@ -0,0 +1,68 @@
+/** Button variant with waiting and progress tracking. */
+<template>
+    <button
+        class="ui-button-default"
+        v-bind:class="[waiting ? 'disabled' : '', variantClass]"
```
guerler: I wonder if we should avoid this one-line if statements in the template and consistently use computed values instead. I personally lean towards computed values going forward.

martenson: I also think computed vs ternary has a clear winner

client/galaxy/scripts/components/Workflow/Run/WorkflowRun.vue
```diff
@@ -0,0 +1,162 @@
...
+                this.loading = false;
+            })
+            .catch(response => {
+                console.log(response);
+                this.error = errorMessageAsString(response);
+            });
+    },
+    methods: {
+        execute() {
+            this.$refs["run-form"].execute();
```
guerler: It would be nice if we could develop a cleaner way for the interaction between the vue-component and backbone sub views. Maybe a reusable wrapper in order to avoid the over use of `$refs`.

jmchilton: Thanks for the review @guerler!

From https://vuejs.org/v2/guide/components-edge-cases.html

> $refs are only populated after the component has been rendered, and they are not reactive. It is only meant as an escape hatch for direct child manipulation - you should avoid accessing $refs from within templates or computed properties.
 
This usage is neither within a template or a computed property. I'm not sure I see value in avoiding ``$refs`` here - this seems like a totally fine instance of this. Is there a reason you disagree? I'm trying to understand the concern.

Since ``run-form`` can be one of two forms - I think any property handling thing would lead to duplication in the template stuff but I did look around for ideas nonetheless...

I found this:

https://www.reddit.com/r/vuejs/comments/79stdr/how_to_avoid_refs_here/

The idea I guess there would be that we addd a property "submittingForm" to the parent I suppose (WorkflowRun in this case).  And pass submittingForm to both the child elements. I think that would be kind of yucky - especially because I think we would need to add a ``watch`` to both of the child components?

```
watch: 
  formSubmitting: function(value) {
     if(value) {
        execute();
    }
  }
```

Which seems objectively more code and duplication and subjectively less intuitive to me. The top comment in that reddit thread (only 3 points I'll admit):

> As you demonstrate, the code becomes less clear, and harder to reason about in this instance.
> This seems like a good candidate for calling a specific method on your child component this.$refs.canvas.snapshot(), and having that child component fire an event which the parent captures to get the canvas data.
> The only "reaching in" you're doing is calling an available method on your child.

Maybe a wrapper like you suggest would somehow get around this sort of imitating function calls with properties and events... but I don't see it?

client/galaxy/scripts/components/WaitButton.vue
jmchilton: Will do, thanks!

jmchilton: Rebased and included in https://github.com/galaxyproject/galaxy/pull/9138 the oldest/smallest PR that contained this. Will rebase these other PRs on top of that - but they are all going to lose their selenium testing :(.

client/galaxy/scripts/components/Workflow/Run/WorkflowRun.vue
guerler: Thanks for the detailed explanation. It makes sense to me and is a reusable pattern. A tiny improvement would be to use a name without dash to avoid quotation marks.

jmchilton: Fair - happy to make that change.

jmchilton: Done - and rebased into #9147.

bgruening:
Would it be possible to show the simplified form even when `simplified_workflow_run_ui = off` but some GET parameter is sent? I think this would be nice for integration from Dockstore or workflowhub.eu.

dannon:
What's the argument for putting this behind a default-off configuration option?

jmchilton:
@dannon 

Just me being conservative (and before my work today this not all the options on the screen worked - batch parameters were not working). At worst giving people a release to test it out and form opinions. I'm happy to make this the default I think though. Arguments against it might be more clicks to get to get to certain options and a jarring user experience when best practice workflows render the new way and other workflows render the old way.

@bgruening 

It can be overridden now with ``simplified_workflow_run_ui=prefer`` in the query string.

dannon:
@jmchilton Well, it would definitely be best if it worked, but yeah -- I think this becoming the default with a simple toggle (could even set a user preference for this very easily, to allow individuals to prefer simple or advanced) to show the whole form is totally fine.

guerler:
This is great, thank you so much for simplifying the execution form. While looking into it I noticed that the lazy loading for steps seems to be impeded, maybe it is something we can address here too. ref #10034

client/galaxy/scripts/components/Workflow/Run/WorkflowRun.vue
```diff
@@ -83,20 +110,49 @@ export default {
...
+                    if (model.hasReplacementParametersInToolForm) {
+                        console.log("cannot render simple workflow form - has ${} values in tool steps");
+                        simpleForm = false;
+                    }
+                    // If there are required parameters in a tool form (a disconnected runtime
+                    // input), we have to render the tool form steps and cannot use the
+                    // simplified tool form.
+                    if (model.hasOpenToolSteps) {
+                        console.log(
+                            "cannot render simple workflow form - one or more tools have disconnected runtime inputs"
```
guerler: These probably should be shown as error messages on the screen.

jmchilton: These aren't errors though (?) - if any of these conditions are met the simple the full form is just rendered. I was thinking about adding a "lint for best practices" button in the workflow editor - that might be a better spot for user messages around these?

Although - I am noticing I don't have that simple_form=false here in this branch like the others.

bgruening (approved):
Awesome! Thanks John!

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -2436,6 +2436,38 @@ mapping:
...
+        desc: |
+          If set to 'off' by default, always use the traditional workflow form that renders
+          all steps in the GUI and serializes the tool state of all steps during
+          invocation. Set to 'prefer' to default to a simplified workflow UI that
+          only renders the inputs if possible (the workflow must have no disconnected
+          runtime inputs and not replacement parameters within tool steps). In the
+          future 'force' may be added an option for Galaskio-style servers that should
+          only render simplified workflows.
+
+      simplified_workflow_run_ui_target_history:
```
dannon, line 2453: This and the setting below really feel like they should be user preferences, not galaxy-wide configuration settings for the sake of simplifying a page.  For more context, this notion is probably stemming from my work separating the client from the server, but this *more* tightly couples the displayed interface with the server and its configuration.

jmchilton: This project stemmed from a desire to present a completely different, simplified UI around workflows - including one that doesn't even need to present the concept of a history to users. If you wanted to deploy such an app you really wouldn't want to make this a user setting. 

I'm all for decoupling - the corresponding attributes on Vue components do have defaults. They should work stand-alone in other scenarios. So they are decoupled properly in my mind. Just because we don't want to force the developer using the components to define these doesn't mean the admins shouldn't be allowed to pick defaults. Likewise, just because it would likely be good to allow user preferences in the abstract for more typical uses doesn't mean we shouldn't allow admins to be able to control how the app is configured as a default.

Having one part of the UI depend on Galaxy's concept of a user preference tab and the backend for storing and fetching those - is in fact more coupling to my mind than this. Again, the components would work without those so it seems like a fine sort of optional coupling. We have independent ways to configure these things and the component internals don't depend on how those are being configured - that would seem to be the goal to me.

Anyway, while I agree with some of this, adding the user preferences seems like outside the scope of the initial implementation. If you insist on my adding setting I will consider it to get this done, but even in that scenario I still think the defaults should be configurable by the deployer. Though I'm totally open to there being a different way for the deployer to configure the client than Galaxy's YAML.

dannon: >  If you insist 

@jmchilton I really wasn't insisting on anything, sorry if you thought as much.

I think this component should be seen as a simple, togglable level of detail view rather than a different, server-set preference.  In my mind the second pass, the actual workflow-centric interface addition, that's what would add these extra properties (however they end up) and pass them to this view.  This is nice progress as a straightforward level of detail toggle on the existing workflow interface.  That's what I was trying to get at above, with the configuration defaults -- it doesn't serve anyone to force one view over the other with just this piece, I don't think, and I suppose a user could just flip the interface one way or another if they wanted anyway by swapping a bit of client-side data.

Anyway, for now, it's fair enough that until we have a way to configure more (and default) client(s), that this can live in the core galaxy configuration.  If you want this configurable at this point, then it's fine.

client/galaxy/scripts/components/Workflow/Run/WorkflowRun.vue
dannon: Agreed it doesn't make sense to bubble this to the user; there's nothing for them to act on here and it's not something we can/should pause any interaction for.

client/src/components/Workflow/Run/WorkflowRunFormSimple.vue
```diff
@@ -0,0 +1,117 @@
...
+            const el = this.$refs["form"];
+            $(el).append(this.form.$el);
+        });
+    },
+    methods: {
+        formDefinition() {
+            const inputs = [];
+
+            // Add workflow parameters.
+            _.each(this.model.wpInputs, (input, i) => {
```
dannon: We should be avoiding underscore when possible at this point.

client/src/components/Workflow/Run/WorkflowRunFormSimple.vue
```diff
@@ -0,0 +1,117 @@
...
+                const inputCopy = Object.assign({}, input);
+                // do we want to keep the color if we're not showing steps?
+                inputCopy["color"] = undefined;
```
dannon: We generally prefer dot notation.

client/src/components/Workflow/Run/WorkflowRunFormSimple.vue
```diff
@@ -0,0 +1,117 @@
...
+            form: null,
+            inputTypes: {},
+        };
+    },
+    computed: {},
+    created() {
+        this.form = new Form(this.formDefinition());
+        this.$nextTick(() => {
+            const el = this.$refs["form"];
+            $(el).append(this.form.$el);
```
dannon: We should avoid jquery in vue components.

jmchilton:
This is ancient code at this point 😂 - I’ll clean up the jquery and underscore today - thanks for the review.

bgruening:
Awesome! Yeah!!!!

TKlingstrom:
Wow thanks Jon and other people commenting, I really look forward to looking at this next week. 

(Also hugely sorry for disappearing and not doing anything for half a year, I do not control time in my my work life nor my private life as much as I would like but it's moving in the right direction at least).

afgane:
This is indeed excellent. Nonetheless, 3 comments:
- The config values are not included in `galaxy.yml.sample`
- If there is a link to switch to the full workflow form, it feels it should be a toggle so you can switch back to the simplified view
- It would be great to include the workflow annotation text at the top of the run form

bgruening:
@afgane can you create a new issue with those ideas, please. I think we can iterate here with many other small things. We also discussed briefly to include maybe a small image of the workflow etc.

mvdbeek:
Looks like the input step parameters are not actually sent in the simplified form:
```json
{"replacement_dict":{},"inputs":{"0":{"values":[{"hid":1,"id":"db210f2338fbfc8b","keep":false,"name":"Pasted Entry","src":"hda","tags":[]}],"batch":false}},"inputs_by":"step_index","batch":true,"use_cached_job":false,"history_id":"6169422af4f6057a"}
```

Screenshot 2020-08-20 at 14 22 58

@jmchilton do you know of the top of your head what's going on here ?

mvdbeek:
Got it, https://github.com/galaxyproject/galaxy/blob/dev/client/src/components/Workflow/Run/WorkflowRunFormSimple.vue#L88 doesn't include `parameter_input`, I'll have a fix in a minute

Issue #6981, opened 2018-11-06
Add admin API for resending user activation email


hexylena:
Currently if the user mis-types their email address, I have to correct this + impersonate them + click "resend activation email". It would be better if I could have an API to just reset based on encoded user ID.

jdavcs:
Closing as resolved by #9154 (and as per discussion in that PR). If an API endpoint becomes necessary, please reopen.

PR #9154, opened 2019-12-20
Enable admin resend activation email

jdavcs:
Re #6981
@erasche I misread the issue and implemented a button for the admin UI instead of an API endpoint. Which was careless of me.. But since it's done, will this solve the issue, and is it useful at all? Or is an API endpoint preferable? (I'll implement the endpoint if need be) 
@dannon I used same approach.

### DIFF
--- a/lib/galaxy/webapps/galaxy/controllers/admin.py
+++ b/lib/galaxy/webapps/galaxy/controllers/admin.py
@@ -532,6 +532,7 @@ class AdminGalaxy(controller.JSAppLauncher, AdminActions, UsesQuotaMixin, QuotaP
         allow_multiple=False
     )
     activate_operation = grids.GridOperation("Activate User", condition=(lambda item: not item.active), allow_multiple=False)
+    resend_activation_email = grids.GridOperation("Resend Activation Email", condition=(lambda item: not item.active), allow_multiple=False)
 
     @web.expose
     @web.require_admin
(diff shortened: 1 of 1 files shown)


martenson:
With your previous PR you closed https://github.com/galaxyproject/galaxy/issues/3208

jdavcs:
> With your previous PR you closed #3208

But that's a different issue? It's mentioned in 3208's discussion.

dannon:
I'm guessing having the admin be able to manually activate a user is sufficient here, and that this might not be necessary?

jdavcs:
Close it then? I don't mind - was minimal effort anyway.

hexylena:
> is it useful at all

yes!!!! This is great! APIs are great for some admins, but in general admins should be able to do everything they need without resorting to the API.

hexylena:
> I'm guessing having the admin be able to manually activate

In EU's case, sometimes emails didn't get sent for one reason or another. We occasionally had to manually send the activation link to the users.

I don't want to activate users for them, because their accessing of the activation link confers consent to the ToS/our privacy policy.

jdavcs:
@erasche thanks for the comment - this certainly clarifies the use case for this!

dannon:
Cool, happy to defer to @erasche's admin experiences here -- I hadn't considered the ToS agreement angle, makes sense to me.

jdavcs:
@galaxybot test this

bgruening (approved):
@ic4f cool beans!

PR #9157, opened 2019-12-21
"Normalize" more custom functions in admin jobs index.

jmchilton:
Continuing thread of #8943,  #9093... trying to standardize the jobs index Vue component for re-use.

- Eliminate info_url from API web controller response - wasn't used and isn't available in regular jobs API calls that I'd like to get to.
- Eliminate showing the command-line as a row of the table - it is redundant with the expanded row details which are already there, was ugly when embedded in the row, and simplifies the form nicely to get rid of.
- Eliminate "inputs" from the web controller response - it isn't available in the regular API and I replaced the inputs handling with reuse of the JobParameters Vue component. We get this for free from the job id, it renders a lot richer information about the parameters, input names, etc... and things are hyperlinked so inputs can be navigated to.
- Add JobMetrics - since we have a component to do that for free that looks a lot like the JobParameters component. 

Before:

Screen Shot 2019-12-21 at 12 19 28 PM

After:

Screen Shot 2019-12-21 at 12 17 11 PM

I think the couple more clicks to certain pieces of information is worth it to get us closer to a re-usable component built on the standard APIs especially given the richer information we are now able to present (full tool parameters with links and handling non-dataset inputs as well as job metrics). If people feel strongly about fast index-level access to input states or something though - I can try to rework the traditional APIs to expose this efficiently but my initial thought is that it is not worth the effort.

### DIFF
--- a/client/galaxy/scripts/components/admin/JobDetails.vue
+++ b/client/galaxy/scripts/components/admin/JobDetails.vue
@@ -1,0 +1,39 @@
+<template>
+    <b-card>
+        <h5>Command Line</h5>
+        <pre class="text-white bg-dark"><code class="break-word">{{ commandLine }}</code></pre>
+        <h5>Job Parameters</h5>
+        <job-parameters :jobId="jobId" :includeTitle="false" />
+        <h5>Job Metrics</h5>
+        <job-metrics :jobId="jobId" :includeTitle="false" />
+    </b-card>
+</template>
+
+<script>
+import { JobMetrics } from "components/JobMetrics";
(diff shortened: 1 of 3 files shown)


martenson:
I like it.

jdavcs:
+1

bgruening (approved):

nsoranzo:
@jmchilton This was merged while still labelled WIP, hope that's not a problem.

jmchilton:
Not a problem - thanks all!

PR #9162, opened 2019-12-23
Add support for logging out of OIDC Identity Provider

machristie:
This PR adds support for logging out of an OIDC Identity Provider when the user logs out of Galaxy. This makes sense for some IDPs that exist primarily to provide authentication for Galaxy such as a private Keycloak server.  Logging out of the IDP is controlled by a new configuration option in oidc_backends_config.xml called `<enable_idp_logout>`.

The implementation works by creating a cookie called `oidc-provider` when the user logs in through an OIDC IDP to record which provider was used. Then, on logout (menu.js), /authnz/logout is called to determine if IDP logout is supported and enabled for the IDP and if so a redirect URI is returned. The user's browser is redirect to this redirect URI which terminates the user's session in the IDP. The IDP then redirects back to the Galaxy home page.

Support for IDP logout is only implemented for Custos/Keycloak since the PSA based backends don't support logout ([PSA only supports disconnecting](https://python-social-auth.readthedocs.io/en/latest/logging_out.html), but [logout support was added for SAML](https://github.com/python-social-auth/social-core/issues/199) recently so it may someday be available for other backends).


lib/galaxy/authnz/managers.py
```diff
@@ -126,7 +127,8 @@ def _parse_custos_config(self, config_xml):
             'client_id': config_xml.find('client_id').text,
             'client_secret': config_xml.find('client_secret').text,
             'redirect_uri': config_xml.find('redirect_uri').text,
-            'realm': config_xml.find('realm').text}
+            'realm': config_xml.find('realm').text,
+            'enable_idp_logout': config_xml.findtext('enable_idp_logout', 'false').lower() == "true"}
```
nuwang: The `galaxy.util.asbool` function is a potential replacement for this: https://github.com/galaxyproject/galaxy/blob/0204c9b65e1dac407046112b7db7901a36eed971/lib/galaxy/util/__init__.py#L925

lib/galaxy/authnz/managers.py
```diff
@@ -260,6 +262,23 @@ def callback(self, provider, state_token, authz_code, trans, login_redirect_url)
             log.exception(msg)
             return False, msg, (None, None)
 
+    def logout(self, provider, trans, post_logout_redirect_url=None):
+        try:
+            # check if logout is enabled for this idp and return false if not
+            unified_provider_name = self._unify_provider_name(provider)
+            if self.oidc_backends_config[unified_provider_name]['enable_idp_logout'] is False:
+                return False, "IDP logout is not enabled for {}".format(provider), None
```
nuwang, line 284: Are the return values for this documented somewhere?

lib/galaxy/webapps/galaxy/buildapp.py
```diff
@@ -72,6 +72,7 @@ def app_factory(global_conf, load_app_kwds={}, **kwargs):
     webapp.add_route('/authnz/{provider}/login', controller='authnz', action='login', provider=None)
     webapp.add_route('/authnz/{provider}/callback', controller='authnz', action='callback', provider=None)
     webapp.add_route('/authnz/{provider}/disconnect', controller='authnz', action='disconnect', provider=None)
+    webapp.add_route('/authnz/logout', controller='authnz', action='logout')
```
nuwang: Should this be a per provider logout as well?

lib/galaxy/authnz/managers.py
machristie: Good suggestion, will do.

lib/galaxy/authnz/managers.py
machristie: Good catch, I'll document.

lib/galaxy/webapps/galaxy/buildapp.py
machristie: I think having per provider logout could work too, but this seemed more natural to me from the perspective of the UI. For login, the user has picked a specific provider to log into. However, on logout the user has only indicated that they want to logout. /authnz/logout figures out what provider was used to authenticate (using the `identity-provider` cookie) and logs the user out of that provider.

nuwang: The UI perspective makes sense. What do you think about the API perspective? I feel as if having a per provider logout allows programmatic logout from specific providers, although I'm not sure there's a clear use case. Perhaps a programmatic global logout from a specific provider like keycloak only through the API perhaps?  Is the only necessary change to read the cookie from the client side instead of the server?  I also feel like `authnz/logout` should log the user out from all providers, not just the last logged in one? Just some thoughts, but no strong views on which may be better.

machristie: What your saying makes a lot of sense. I do like that per provider logout is consistent with the other provider methods.

> Is the only necessary change to read the cookie from the client side instead of the server?

Yeah I think that's basically it. However, I think I would change it so that server side the cookie is used to determine the logout API URL and this is provided to the client. I'll work on updating the PR.

machristie: Hi @nuwang , I looked into passing a value to the UI, but it seemed simpler and more in keeping with the Galaxy client design to add an API method for getting the provider specific logout url, so I did that. The frontend now calls `/authnz/logout` which reads the oidc-provider cookie and if it exists, redirects to `/authnz/{provider}/logout`.

nuwang (approved):
@machristie Thanks for the changes, looks great to me!

dannon:
I'll get keycloak configured again for testing this morning and will merge this today when it works.

Resolved conflicts with upstream, pushed the merge here.

dannon:
This is working well for me so I'm going to go ahead and merge it when the tests finish this time.  I do think we might want to address the potential refactoring @nuwang mentions, and we might want to drop `get_logout_url` since, with teh cookie, we actually have all the information we need to call the correct method from the client side from the start.

So, you could check for and read `oidc-provider`, you'll know it's `custos`, and can then call authnz/custos/logout, right?

Anyway, thank you again for the enhancement.

dannon (approved):

machristie:
Thanks @dannon 

> So, you could check for and read oidc-provider, you'll know it's custos, and can then call authnz/custos/logout, right?

Yes that's right. I went with the approach I took so that only the backend needs to know about cookie instead of the front end and back end needing to know about the cookie. That makes it easier to change the code around setting the cookie, if necessary, going forward. But the approach of the client reading the cookie I think would be fine too, especially if we don't anticipate change in the logic around setting the cookie. I hope that makes sense.

I can put together a PR to do that refactor, maybe this weekend.

nuwang:
@machristie That makes sense. In any case, in future perhaps an enhancement may be to have the logout url log you out of all providers, and the provider specific logout url log you out of that provider only. For now, this seems like a pretty good way for things to work, and we can implement that pretty easily if we ever need it, so I think we may as well leave this as is if @dannon agrees.

PR #9163, opened 2019-12-24
[20.01] mulled: use threading.Event in PrintProgress to avoid unnecessary wai…

mvdbeek:
…ting

cherry-picked from https://github.com/galaxyproject/galaxy-lib/pull/155


mvdbeek:
@galaxybot test this

Issue #9114, opened 2019-12-14
Implement Collection Selection Dialog


jmchilton:
This is the collection issue mirroring the data issue outlined in https://github.com/galaxyproject/galaxy/issues/9110. The idea outlined there (and in http://bit.ly/simplified-workflow-ui) is that we can simplify the UX for running workflows by eliminating the tool panel and upload options if we can provide context-aware alternatives directly from the workflow run form. #8512 introduces a new collection selection dialog component, it needs to be enhanced significantly as follows:

- [ ] Add a button to the tool form for this widget to match the button for dataset.
- [ ] Delegate to the upload component to allow upload/creation of collections directly from the selection component (or a button next to it on the tool form).
- [ ] Allow restricting the upload and create options based on the collection type annotated in the tool form.
- [ ] Allow restricting the upload and create options based on the new format for the collection that will be annotated after #9086.

Stretch Goals:

 - [ ] Allow selecting collections from other histories.
 - [ ] Allow selecting library folders and creating collections from them.

Issue #9110, opened 2019-12-14
Improved Data Selection Dialog for Simplified Workflow Run


jmchilton:
As part of https://github.com/galaxyproject/galaxy/projects/15 (Simplified Workflow-centric User Experience) we would like to eliminate the history and tool panel as a landing point for certain Galaxy configurations or certain classes of users. To enable this we would like to enhance the data selection dialog to allow uploading data directly from it (or a companion widget beside it in the workflow run). This combined with creating a history per workflow run - would allow us to eliminate the history panel and tool panel in this view of Galaxy. Additionally, we could leverage the input types, collection types, etc.. to provide context-aware uploads to further simplify the user experience. The upload widget would be less daunting if you need to the corresponding upload needed to be a fastqsanger - we could eliminate type selection, composite and collection tabs, do extra validation by default, eliminate some advanced options, etc.. 

I think some of these issues could be tool form issues instead of the selection widget but I'm tracking them all together here for now. Happy to break them up into more specific issues and just track everything on https://github.com/galaxyproject/galaxy/projects/15

Roughly in order of importance:

- [x] Add upload option to the tool form or selection widget.
- [x] Revise the upload widget to allow more specific upload experience if the target format(s) are known ahead of time.
- [x] Allow restricting upload widget to focus a specific format or set of formats.
- [ ] Make data selection parameter a drag/drop target and allow dropping data into this form
- [ ] The data dialog can already expose library data and could be extended to allow fetching data from other histories.

As @bgruening pointed out in http://bit.ly/simplified-workflow-ui:

> This could also be used to create single-tool interfaces. We do have a lot of tools that are just used once. Visualization tools (like circos) or Annotation tools like antismash would profit from its own view (history, toolbox hidden). This new data selection would make this possible as well.

PR #9167, opened 2019-12-27
Vue-ify Upload Dialog w/small fixes

jmchilton:

- Removed upload-button.js - mostly unused after #8368.
- Simplifications and ES6 stuff to existing upload Backbone components.
- Swap upload-view to UploadModal and use ``b-modal`` instead of Galaxy's modal stuff - use ``no-enforce-focus`` to ensure Select2 works within the modal.
- Migrate usage of upload Tabs to ``b-tabs`` - (initial shim for Backbone views in early commits replaced with Vue components for each tab type in subsequent commits).
- Replace each ``mvn/upload/<type>/<type>-view.js`` Backbone View with a Vue component.
   - Migrate composite-view.js to a new Composite.vue - eliminates Backbone view for everything but the row. Still relies of Backbone models.
   - Migrate collection-view.js to a new Collection.vue - eliminates Backbone view for everything but the row. Still relies of Backbone models.
   - Migrate rules-input-view.js to a new RulesInput.vue - eliminates Backbone view, never used row views or models.
   - Migrate default-view.js to a new Default.vue - eliminates Backbone view,  eliminates Backbone view for everything but the row rending. Still relies on Backbone models.
    - More reuse than Backbone code given mixins and components - this PR vastly reduces the duplication between the collection and default uploads.
- Parameterized lazy limited widget handling between Default.vue and Collection.vue - either can have a lazy limited set via a Vue property. Really reduces the divergence between default upload and collection upload nicely. I added some tests for this and fixed an off by one error inside of lazy-limited.js.
- Largely unrelated to Vue, but I took a first crack at eliminating a lot of the copy-paste duplication between the Backbone views in default-row.js and collection-row.js. Just cleaning up past messes I made ahead of potentially Vue-ifying these as well.
- Eliminates some qunit tests, adds more modern vuejs-utils tests. Not exactly one-to-one because some jquery plugins behave different in karma vs. qunit but definitely considerably more tests now.
- Generalize special case handling of FTP files that uploads them all in a single request for default uploads to work with collection uploads also - removing divergence between two upload modalities.


jmchilton:
~~Continued the vue-ification throwing in the tabs and the actual composite uploader view... select2 still not working. I think the way the tabs work - this is all now creating the select2 components later... I had hope that would fix things.~~ (Fixed see comment below)

jmchilton:
It was the tabindex on b-modal... https://github.com/select2/select2-bootstrap-theme/issues/41

https://bootstrap-vue.js.org/docs/components/modal/ no-enforce-focus on the modal seems to fix it - but it is strongly discouraged. I guess it is no worse than our current modals?

jmchilton:
@galaxybot test this

guerler:
Noticed that the select2 dropdown width is fixed to the parent dom element width:

sofi finance offer 1st

guerler:
Noticed that the select2 dropdown width seems to be fixed to the actual parent dom width:

widthselect2

client/galaxy/scripts/components/Upload/Composite.vue
```diff
@@ -0,0 +1,211 @@
+<template>
+    <upload-wrapper wrapperClass="upload-view-composite">
+        <div class="upload-helper" v-show="showHelper">Select extension to show files required to upload</div>
```
guerler: Maybe "Select Composite Type to show required files." instead since the select field is labeled: "Composite Type".

jmchilton: Good catch - made the change.

jmchilton:
I've fixed the Select2 width issue with https://github.com/galaxyproject/galaxy/pull/9167/commits/224a92aaa290efb2ccf4061bc7db48ae5bd2e0d2 I believe. Thanks for the catch @guerler

guerler:
I think we need to vue-ify the backbone rows too, other than that this looks great.

jmchilton:
> I think we need to vue-ify the backbone rows too

Eventually that would make sense, but there is no reason that would have to happen as part of this PR I assume. I did this to enable #9110 and #9114 not to just to Vue for the sake of Vue-ing.

dannon:
This seems fine as an incremental step.  As we eliminate the rest of the backbone (especially the models..) it'll all be a lot cleaner.

Thanks!

PR #9170, opened 2019-12-28
[20.01] Set Conda environment variables  when running biocontainers

mvdbeek:
This is needed for containers built from Conda packages that populate
various environment variables in their activate.d scripts.
One example for which this is needed is augustus.

It would probably be better to set and/or record these variables when
building the container, but I think this should cover what is needed
in the IUC (augustus_training only, feelnc, sina).
This should bring biocontainers closer to how one would work with
Conda packages and should benefit other tool maintainers that
haven't started using these workarounds.

### DIFF
--- a/lib/galaxy/tool_util/deps/container_classes.py
+++ b/lib/galaxy/tool_util/deps/container_classes.py
@@ -56,6 +56,20 @@ if not found:
     subprocess.check_call(cmd, shell=True)
 EOF
 '''
+SOURCE_CONDA_ACTIVATE = """
+# Check if container was created by installing conda packages,
+# and if so, source scripts to populate environment variables
+# that would be set by activating the conda environment.
+if [ -d /usr/local/etc/conda/activate.d ]; then
+  export CONDA_PREFIX=/usr/local
+  for f in /usr/local/etc/conda/activate.d/*.sh; do
+    case "$f" in
+      "/usr/local/etc/conda/activate.d/activate-"*) :;;
+      *) . "$f" ;;
+    esac;
+  done
+fi
+"""
 
 
 @six.add_metaclass(ABCMeta)
@@ -85,20 +99,9 @@ class Container(object):
 
     @property
     def source_environment(self):
-        return """
-# Check if container was created by installing conda packages,
-# and if so, source scripts to populate environment variables
-# that would be set by activating the conda environment.
-if [ -d /usr/local/etc/conda/activate.d ]; then
-  export CONDA_PREFIX=/usr/local
-  for f in /usr/local/etc/conda/activate.d/*.sh; do
-    case "$f" in
-      "/usr/local/etc/conda/activate.d/activate-"*) :;;
-      *) . "$f" ;;
-    esac;
-  done
-fi
-"""
+        if self.container_description and not self.container_description.explicit:
+            return SOURCE_CONDA_ACTIVATE
+        return ""
 
     @abstractmethod
     def containerize_command(self, command):
--- a/lib/galaxy/tool_util/deps/container_classes.py
+++ b/lib/galaxy/tool_util/deps/container_classes.py
@@ -85,7 +85,11 @@ class Container(object):
 
     @property
     def source_environment(self):
-        return """if [ -d /usr/local/etc/conda/activate.d ]; then
+        return """
+# Check if container was created by installing conda packages,
+# and if so, source scripts to populate environment variables
+# that would be set by activating the conda environment.
+if [ -d /usr/local/etc/conda/activate.d ]; then
   export CONDA_PREFIX=/usr/local
   for f in /usr/local/etc/conda/activate.d/*.sh; do
     case "$f" in
--- a/lib/galaxy/tool_util/deps/container_classes.py
+++ b/lib/galaxy/tool_util/deps/container_classes.py
@@ -83,6 +83,19 @@ class Container(object):
     def shell(self):
         return DEFAULT_CONTAINER_SHELL if not self.container_description else self.container_description.shell
 
+    @property
+    def source_environment(self):
+        return """if [ -d /usr/local/etc/conda/activate.d ]; then
+  export CONDA_PREFIX=/usr/local
+  for f in /usr/local/etc/conda/activate.d/*.sh; do
+    case "$f" in
+      "/usr/local/etc/conda/activate.d/activate-"*) :;;
+      *) . "$f" ;;
+    esac;
+  done
+fi
+"""
+
     @abstractmethod
     def containerize_command(self, command):
         """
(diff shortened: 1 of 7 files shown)


test/integration/test_scripts.py
```diff
@@ -185,6 +184,9 @@ def write_config_file(self):
         config_dir = self.config_dir
         path = os.path.join(config_dir, "galaxy.yml")
         self._test_driver.temp_directories.extend([config_dir])
+        config = self._raw_config
+        # Update config dict with database_connection, which might be set through env variables
+        config['database_connection'] = self._app.config.database_connection
         with open(path, "w") as f:
             yaml.dump({"galaxy": self._raw_config}, f)
```
nsoranzo: Should this be:
```suggestion
            yaml.dump({"galaxy": config}, f)
```
?

mvdbeek: It should, thanks!

.github/workflows/jenkins.yaml
```diff
@@ -0,0 +1,48 @@
...
+      run: eval ${{ steps.minikube.outputs.launcher }}
+    - name: Check pods
+      if: matrix.test-suite == 'integration'
+      run: |
+        kubectl get pods
+    - name: Cache venv dir
+      uses: actions/cache@v1
+      id: cache-venv-galaxy
+      with:
+        path: .venv
```
nsoranzo: Should we cache both `~/.cache/pip` and `.venv` ? When this will be used for multiple galaxy branches, different dependencies will be installed in the virtualenv.

mvdbeek: But then you would still have a cache miss in~/.cache/pip, I'd expect this wouldn't reduce test setup time. We could possibly key on the target branch, but I don't think there would be any benefit to caching both locations

nsoranzo: > But then you would still have a cache miss in `~/.cache/pip`

Why? If you try `pip install -v somepackage==someversion` with alternate package versions, you'll see that pip will use cached data.

mvdbeek: How is that possible ? You only have one version in the cache, right ?

mvdbeek: I mean the way the github cache works it isn't possible. I understand it would make sense if the cache was uploaded after a successful run, but if the cache was hit it won't be uploaded.

nsoranzo: Ah, I didn't know that the `actions/cache` for GitHub workflows works differently than the Travis cache (where the cache is updated at every successful build).

Anyway, it seems there are separate caches for different branches https://help.github.com/en/actions/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache , so caching the venv should be fine.

Maybe the key should include the hash of requirements.txt though? As in https://github.com/actions/cache/blob/master/examples.md#simple-example

mvdbeek: I've rebased that into https://github.com/galaxyproject/galaxy/pull/9170/commits/c76c2322f470b294fcb48a661ff0e9bfda3d07b1

lib/galaxy/tool_util/deps/container_classes.py
```diff
@@ -83,6 +83,19 @@ def resolve_dependencies(self):
     def shell(self):
         return DEFAULT_CONTAINER_SHELL if not self.container_description else self.container_description.shell
 
+    @property
+    def source_environment(self):
+        return """if [ -d /usr/local/etc/conda/activate.d ]; then
```
bgruening: Do we want to add a comment into the script, that this is for automatically build containers?

bgruening (commented):
@mvdbeek should we skip this for explicitly defined containers?

lib/galaxy/tool_util/deps/container_classes.py
mvdbeek: Sure, I can do that

mvdbeek:
> @mvdbeek should we skip this for explicitly defined containers?

We don't know what are explicitly defined containers at this point (unless I'm missing something). The check itself is probably not more (or less) efficient than figuring this out on the python level, so I'm not concerned about performance.

test/functional/tools/job_environment_default.xml
```diff
@@ -1,6 +1,6 @@
 <tool id="job_environment_default" name="job_environment_default" version="0.1.0" profile="18.01">
     <requirements>
-      <container type="docker">busybox:ubuntu-14.04</container>
+        <container type="docker">mvdbeek/source-activate-minimal:0.1</container>
```
bgruening: I would skip those cases.

bgruening (commented):
@mvdbeek with explicitly defined containers I mean, tools that have set a `<container type="docker">` requirement. Also tools of type="interactive" should be skiped isn't it?.

mvdbeek:
> @mvdbeek with explicitly defined containers I mean, tools that have set a `<container type="docker">` requirement

Why should explicitly defined containers not be biocontainers or be built through our ecosystem ?
That seems perfectly valid ?

bgruening:
> > @mvdbeek with explicitly defined containers I mean, tools that have set a `<container type="docker">` requirement
> 
> Why should explicitly defined containers not be biocontainers or be built through our ecosystem ?
> That seems perfectly valid ?

Imho, they should be defined as a package, if there are biocontainers and a conda package exists.

mvdbeek:
> Imho, they should be defined as a package, if there are biocontainers and a conda package exists.

Sure, but how do we know everyone adheres to this ? I don't understand what's the harm with checking if the activate.d directory exists. I think this makes less assumptions about how people use the dependencies.

mvdbeek:
@bgruening I've excluded the sourcing mechanism for explicit container annotations in https://github.com/galaxyproject/galaxy/pull/9170/commits/5eef8d163cbb0d784a30e19e3e7dc529ac9c6eca, how do you feel about this now ?

bgruening (approved):
@mvdbeek sorry, took me a while!

LGTM.

jmchilton:
I also think it is is a good idea to only enable this for biocontainers. It is a gross hack (not the Galaxy implementation, but that it is needed at all) and so it should be minimized. Also our jobs have sooo much overhead already and we just keep adding more - short circuiting the check when we can is good.

mvdbeek:
> It is a gross hack (not the Galaxy implementation, but that it is needed at all)

I know, that's why I think this should be set when building the image, not when running it. 
https://github.com/galaxyproject/galaxy/pull/9088 was my first try on this, but it looks like that's not going to be accepted (looking further I see that the manually built biocontainers also set random ENV vars ... https://github.com/BioContainers/containers/search?p=2&q=ENV&unscoped_q=ENV so I'm not sure why we shouldn't do it). I hope we can revisit this soon.

mvdbeek:
Thanks for the feedback everyone, I do appreciate it.

Issue #9080, opened 2019-12-08
Admin UI issues


afgane:
1. Links on the Admin Dashboard do not respect prefix (see the URL & the link pointed to by _Manage jobs_):
<img width="453" alt="Screen_Shot_2019-12-08_at_10_45_28_AM" src="https://user-images.githubusercontent.com/338642/70391928-33a6f000-19a8-11ea-861a-70dba25de902.png">

2. Admin management panel shows up twice
![Pasted_Image_12_8_19__10_53_AM](https://user-images.githubusercontent.com/338642/70392012-30f8ca80-19a9-11ea-8dc5-3af8bd47426c.png)

3. Links to icons for certain installed tool states are broken
![image](https://user-images.githubusercontent.com/338642/70392076-d90e9380-19a9-11ea-9d06-b2f423e23e23.png)

martenson:
@afgane Galaxy version?

ha, now it can be seen in the menu itself, so 19.09, nice

dannon:
It's 19.09; I'm going through the admin interface now working up fixes.  Should be ready shortly.  (went down a bit of a rabbit hole trying to change more of this than I should have for the stable release; cutting back a bit)

dannon:
@afgane I'll need to catch up with you and look more into problem 3 above -- it has to do with the uwsgi static configuration being employed there (not following the symlinks correctly).

afgane:
OK, let's plan on Monday (at AnVIL meeting for the rest of this week).

PR #9179, opened 2020-01-06
[20.01] Add static-safe mapping to config management

dannon:
Was previously only defined in uwsgi default args which not everything uses.

This closes the loop on the last point of https://github.com/galaxyproject/galaxy/issues/9080

xref: https://github.com/galaxyproject/galaxy-helm/pull/87


nsoranzo:
@galaxybot test this

PR #9189, opened 2020-01-08
Attempt to configure new Galaxies to storing files by UUID by default.

jmchilton:
Extended metadata has been merged #8930. Time to circle back and make it more robust. One of the hardening goals is simplified configuration - so ideally I think Galaxy should store objects by UUID by default so it can then in turn use extended metadata by default.

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -171,6 +171,9 @@ class BaseAppConfiguration(object):
     def _in_sample_dir(self, path):
         return os.path.join(self.sample_config_dir, path)
 
+    def _in_data_dir(self, path):
+        return os.path.join(self.data_dir, path)
+
     def _parse_config_file_options(self, defaults, listify_defaults, config_kwargs):
         for var, values in defaults.items():
             if config_kwargs.get(var) is not None:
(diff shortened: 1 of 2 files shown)


lib/galaxy/config/__init__.py
```diff
@@ -760,6 +773,7 @@ def parse_config_file_options(self, kwargs):
             markdown_export_css=[self._in_config_dir('markdown_export.css')],
             markdown_export_css_pages=[self._in_config_dir('markdown_export_pages.css')],
             markdown_export_css_invocation_reports=[self._in_config_dir('markdown_export_invocation_reports.css')],
+            file_path=[self._in_data_dir('files'), self._in_data_dir('objects')],
```
jdavcs: at this point, the value of `self.file_path` *is* `self._in_data_dir('objects')` (because the default from the schema is resolved w.r.t `data_dir` before `parse_config_file_options()` is called. So, maybe something like this:
`file_path=[self._in_data_dir('files'), self.file_path)` ?
That way we also don't throw away the default value `objects`.

jmchilton: That would look less readable/transparent in my opinion? Maybe best to think of the objects in the config schema as more of a documentation thing in this case? If you insist though I can make the change - I see what you're saying.

jdavcs: I agree on readability, but I do think making use of the default value for all config options makes the system more predictable and easier to understand (i.e., all configs processed in a uniform way), which, I feel, is a good tradeoff. 
If we throw away the default here, that would create a special case; and I think limiting those when possible would be a good thing. (Ideally, after some refactoring, I’d handle such simple cases in the schema: the default would be `objects, files`, both resolving to `data_dir` and then listified).
Maybe a simple comment in the code would be sufficient to address readability?

jmchilton: I've made the requested change, I think it is not great but it is not a big thing either way.

jmchilton:
Merged as part of #9197 but the Github GUI seems confused.

lib/galaxy/config/__init__.py
jdavcs: Thank you!

Issue #7975, opened 2019-05-15
Job Search/cache  not finding equivalent BWA, bowtie and other mappers jobs


Takadonet:
Been playing around with the job caching ability that @mvdbeek has implemented a few releases ago and works very well for most of our tools.

Only exception is that all mapper tools will never find their equivalent job. To my knowledge is looks like the issue is that the job_to_input_dataset table is not matching up with the name column for datasets.

Below is JSON values of parameters being checked. What is  NOT matching in the table is **fastq_input1** and **ref_file** values. So either we have to add _input_type_ and _reference_source_ into the job_to_input_dataset table or remove it from JSON when searching. 

`{**'name_2': 'input_type|fastq_input1'**

 'name_3': 'analysis_type'

 **'name_1': 'reference_source|ref_file'**

 'name_6': 'input_type'

 'tool_version_1': '0.7.17.4'

 'name_4': 'reference_source'

 'name_5': 'rg'

 'state_1': 'new'

 'state_3': 'waiting'
 'state_2': 'queued'
 'state_5': 'ok'
 'state_4': 'running'
 'value_4': '{"\__current_case__": 2
 "adv_se_options": {"\__current_case__": 1
 "adv_se_options_selector": "do_not_set"}
 "fastq_input1": {"values": [{"id": %
 "src": "hda"}]}
 "input_type_selector": "single"}'
 'value_1': '{"\__current_case__": 0
 "analysis_type_selector": "illumina"}'
 'value_3': '{"\__current_case__": 3
 "rg_selector": "do_not_set"}'
 'value_2': '{"\__current_case__": 1
 "index_a": "auto"
 "ref_file": {"values": [{"id": %
 "src": "hda"}]}
 "reference_source_selector": "history"}'
 'tool_id_1': 'toolshed.g2.bx.psu.edu/repos/devteam/bwa/bwa/0.7.17.4'
 'dataset_version_1': 0
 'dataset_version_2': 0
 'param_1': 1
 'id_2': 9
 'id_1': 4}
`  
 

---------------------------------------------------------------

 id  | job_id | dataset_id |     **name**     | dataset_version 
-----+--------+------------+--------------+-----------------
 269 |    125 |          4 | **ref_file**       |              90
 270 |    125 |          9 | **fastq_input1** |              13

Takadonet:
[...]

PR #9190, opened 2020-01-08
#7975 Fix name variable used in job caching lookup table

mgopez:
#7975: When trying to run a previously executed job, the caching step fails to find the previous job as the input parameters are different than the database entries.

- In jobs.py the `path_key` variable is used to hold the name of the input to search in the SQL query.
     - For certain tools this variable was being assigned a value that contains the full path of the input.
     - e.g. `library|input_2`
          -  Whereas within the database the input name is stored as `input_2`

* Fix in this commit makes it so that the `path_key` variable is set to be the input name that is consistent with the database entry by extracting the string that follows the last pipe character ' | '
* Tested using bwa, bowtie2, and shovill. Confirmed that the equivalent job was found and re-used.

### DIFF
--- a/lib/galaxy/managers/jobs.py
+++ b/lib/galaxy/managers/jobs.py
@@ -84,7 +84,7 @@ class JobSearch(object):
         def populate_input_data_input_id(path, key, value):
             """Traverses expanded incoming using remap and collects input_ids and input_data."""
             if key == 'id':
-                path_key = get_path_key(path[:-2])
+                path_key = get_path_key(path[:-2]).split('|')[-1]
                 current_case = param_dump
                 for p in path:
                     current_case = current_case[p]


mvdbeek:
@galaxybot test this

lib/galaxy/managers/jobs.py
```diff
@@ -84,7 +84,7 @@ def by_tool_input(self, trans, tool_id, tool_version, param=None, param_dump=Non
         def populate_input_data_input_id(path, key, value):
             """Traverses expanded incoming using remap and collects input_ids and input_data."""
             if key == 'id':
-                path_key = get_path_key(path[:-2])
+                path_key = get_path_key(path[:-2]).split('|')[-1]
```
mvdbeek, line 87: That's some serious detective work here! Interesting that this fixes things for you, that is probably only correct for a specific parameter type or combination.
I'm worried about matching the wrong input, especially since `get_path_key` goes to great lengths trying to maintain the `|` separated form.
Could you start by adding a test case that fails without this change ?  There is one test in https://github.com/galaxyproject/galaxy/blob/release_20.01/lib/galaxy_test/api/test_tools.py#L630, and now the challenge would be to figure out which dataset parameter is stored without the prefix. Once you know that you can either use an existing framework tool (in test/functional/tools) that fails in the same way or construct a new one for this purpose.

mgopez: Thank you! Let me have a go at creating the test and seeing if I can find which parameter we're looking for.

mvdbeek:
@hellothisisMatt thanks again for your work. I've included your changes in a followup!

PR #9192, opened 2020-01-08
[20.01] Add require_login bypass for entire authnz controller.

dannon:
Without this, when require_login is set you can't use authnz-based login.


jmchilton (approved):
Makes good sense to me.

PR #9193, opened 2020-01-09
[20.01] Run integration tests on github, fix various bugs (interactive tools, kubernetes, test framework)

mvdbeek:
We didn't run any of the tests that need docker on Jenkins (since the tests already run in docker).
This fixes:
  - Interactive tools on python 3
  - ssh/paramiko based tests and host key policy (they only worked on OSX)
  - Updates mulled_build API tests for new API
  - Fixes any_target_requires_extended_base logic
  - Many testing framework fixes

The only integration tests we still skip is test_kubernetes_staging.py ... for this we would need a means to spin up a local checkout of Galaxy in k8s. I'm hoping we can do this with galaxy-k8s in the future.

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -1,0 +1,48 @@
+name: Integration
+on: [push, pull_request]
+env:
+  GALAXY_TEST_DBURI: 'postgres://postgres:postgres@localhost:5432/galaxy?client_encoding=utf8'
+jobs:
+
+  test:
+    name: Test
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        python-version: [3.7, 2.7]
+        subset: ['upload_datatype', 'extended_metadata', 'kubernetes', 'not (upload_datatype or extended_metadata or kubernetes)']
+    services:
+      postgres:
+        image: postgres:11
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: postgres
+        ports:
+          - 5432:5432
+    steps:
+    - name: Setup Minikube
+      if: matrix.subset == 'kubernetes'
+      id: minikube
+      uses: CodingNagger/minikube-setup-action@v1.0.2
+    - name: Launch Minikube
+      if: matrix.subset == 'kubernetes'
+      run: eval ${{ steps.minikube.outputs.launcher }}
+    - name: Check pods
+      if: matrix.subset == 'kubernetes'
+      run: |
+        kubectl get pods
+    - uses: actions/checkout@v1
+      with:
+        fetch-depth: 1
+    - uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Cache venv dir
+      uses: actions/cache@v1
+      id: cache-venv-galaxy
+      with:
+        path: .venv
+        key: cache-venv-galaxy-${{ matrix.python-version }}-${{ hashFiles('requirements.txt') }}
+    - name: Run tests
+      run: './run_tests.sh -integration test/integration -- -k "${{ matrix.subset }}"'
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        python-version: [3.7, 2.7]
+        python-version: [3.7]
         subset: ['upload_datatype', 'extended_metadata', 'kubernetes', 'not (upload_datatype or extended_metadata or kubernetes)']
     services:
       postgres:
--- a/scripts/grt/export.py
+++ b/scripts/grt/export.py
@@ -146,150 +146,146 @@ def main(argv):
     blacklisted_tools = config['sanitization']['tools']
 
     annotate('export_jobs_start', 'Exporting Jobs')
-    handle_job = io.open(REPORT_BASE + '.jobs.tsv', 'w', encoding='utf-8')
-    handle_job.write('\t'.join(('id', 'tool_id', 'tool_version', 'state', 'create_time')) + '\n')
-    for offset_start in range(last_job_sent, end_job_id, args.batch_size):
-        logging.debug("Processing %s:%s", offset_start, min(end_job_id, offset_start + args.batch_size))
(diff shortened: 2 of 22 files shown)


.github/workflows/integration.yaml
```diff
@@ -0,0 +1,48 @@
...
+env: 
+  GALAXY_TEST_DBURI: 'postgres://postgres:postgres@localhost:5432/galaxy?client_encoding=utf8'
+jobs:
+
+  test:
+    name: Test
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        python-version: [3.7, 2.7]
```
nsoranzo: Drop 2.7? It's 2020 :wink: 

```suggestion
        python-version: [3.7]
```

.github/workflows/integration.yaml
```diff
@@ -0,0 +1,48 @@
...
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Cache venv dir
+      uses: actions/cache@v1
+      id: cache-venv-galaxy
+      with:
+        path: .venv
+        key: cache-venv-galaxy-${{ matrix.python-version }}-${{ hashFiles('requirements.txt') }}
+    - name: Run tests
+      run: './run_tests.sh -integration test/integration -- -k "${{ matrix.subset }}"'
```
nsoranzo, line 48:
```suggestion
      run: './run_tests.sh -integration -- -k "${{ matrix.subset }}"'
```

mvdbeek: That doesn't work:
```
./run_tests.sh -integration -- -k bla
invalid option: -k
```

.github/workflows/integration.yaml
mvdbeek: More than happy to, but I do want to see them green once at least.

.github/workflows/integration.yaml
```diff
@@ -0,0 +1,52 @@
...
+  GALAXY_TEST_DBURI: 'postgres://postgres:postgres@localhost:5432/galaxy?client_encoding=utf8'
+jobs:
+
+  test:
+    name: Test
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        python-version: [2.7]
+        subset: ['datatype']
```
nsoranzo:
```suggestion
        subset: ['upload_datatype']
```

mvdbeek: Thanks, I've rebased this into the initial commit.

jmchilton (approved):

scripts/grt/export.py
```diff
@@ -146,7 +146,7 @@ def annotate(label, human_label=None):
     blacklisted_tools = config['sanitization']['tools']
 
     annotate('export_jobs_start', 'Exporting Jobs')
-    handle_job = io.open(REPORT_BASE + '.jobs.tsv', 'w', encoding='utf-8')
+    handle_job = codecs.open(REPORT_BASE + '.jobs.tsv', 'w', encoding='utf-8')
```
nsoranzo: Not sure I understand the reason for this change, we are writing only unicodified content to this file handle.

mvdbeek: On python 3, on python 2 with bytestring literals https://github.com/mvdbeek/galaxy/blob/05b253124ac32eca1521d11a40ad2bc6ff865755/scripts/grt/export.py#L150 and https://github.com/mvdbeek/galaxy/blob/05b253124ac32eca1521d11a40ad2bc6ff865755/scripts/grt/export.py#L184 and more instances of this cause problems

nsoranzo: OMG, how this ever worked? Can I change this back to use `io` and change those lines to write Unicode?

mvdbeek: Of course!

.github/workflows/integration.yaml
```diff
@@ -0,0 +1,48 @@
...
+env:
+  GALAXY_TEST_DBURI: 'postgres://postgres:postgres@localhost:5432/galaxy?client_encoding=utf8'
+jobs:
+
+  test:
+    name: Test
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        python-version: [3.7, 2.7]
```
mvdbeek:
```suggestion
        python-version: [3.7]
```
I guess something went wrong somewhere, I though I rebased it into the first commit ...

nsoranzo: I temporarily dropped it to run the test on 2.7 too.

mvdbeek: Oops, I was a bit too fast then

.github/workflows/integration.yaml
```diff
@@ -0,0 +1,48 @@
...
+env:
+  GALAXY_TEST_DBURI: 'postgres://postgres:postgres@localhost:5432/galaxy?client_encoding=utf8'
+jobs:
+
+  test:
+    name: Test
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        python-version: [3.7]
```
mvdbeek, line 12:
```suggestion
        python-version: [2.7, 3.7]
```

nsoranzo: Test are still running, probably not needed.

nsoranzo (approved):

nsoranzo:
Thanks a lot @mvdbeek for all the work on this!

Should we disable integration tests on Jenkins? For >=20.01 at least.

Also, do you plan to move other tests to GH workflows?

martenson:
@nsoranzo Unless we have other tests that we can't run on jenkins (or can run better on GHci) I think we are at a good spot without the need to move things around. The spread also makes our testing framework more robust I'd speculate.

mvdbeek:
Yes to disabling Jenkins integration tests on >= 20.01, otherwise I agree with @martenson.
Let's see how this service evolves, maybe it will make sense in the future to shift some nodes (especially the cloud workers from eu/au) to the [self-hosted runners](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/hosting-your-own-runners) to extend the capacity. The current capacity is limited to 20/40 concurrent jobs (20 for all, 40 for pro accounts), which I think we would exhaust too quickly if we moved all our testing jobs.

nsoranzo:
I've whitelisted the `release_19.05` and `release_19.09` target branches in the configuration of `docker-integration-py3` jobs on Jenkins: https://jenkins.galaxyproject.org/job/docker-integration-py3/jobConfigHistory/showDiffFiles?timestamp1=2019-11-19_05-54-42&timestamp2=2020-01-09_17-06-56

Let's see if this does the right thing!

PR #9196, opened 2020-01-09
Add a tool test for dynamic select list populated by referring

gregvonkuster:
to a reference dataset contained within nested repeats.  The <filter> tag here...
```
<filter type="data_meta" key="dbkey" ref="signal" column="1"/>
```
...will not work because the ref attribute refers to a param buried within ```<repeat>``` blocks:
```
<repeat name="condition_repeat"...
    <repeat name="signal_repeat"...
        <param name="signal"...
```
The tool will still function correctly, but the dynamically generated select list for the reference genome will not be filtered to the genome associated with the selected inputs for signal.  This is a limitation of the parameter framework.

### DIFF
--- a/test/functional/tools/ref_data_within_repeats.xml
+++ b/test/functional/tools/ref_data_within_repeats.xml
@@ -1,0 +1,43 @@
+<tool id="ref_data_within_repeats" name="ref_data_within_repeats" version="1.0.0">
+    <description>Filter reference genome to associated dataset within nested repeat block</description>
+    <command>
+        echo $reference_genome > $output
+    </command>
+    <inputs>
+        <repeat name="condition_repeat" title="Condition" min="1">
+            <param name="condition_name" type="text" value="cond" label="Condition name">
+                <validator type="empty_field" />
+            </param>
+            <repeat name="signal_repeat" title="Signal" min="1">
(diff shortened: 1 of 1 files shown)


bgruening:
xref: https://github.com/galaxyproject/tools-iuc/pull/2754

This PR is meant as an issue that already implements a test case to demonstrate the issue :)

Thanks @gregvonkuster

bernt-matthias:
As an alternative for the repeat you might use a collection where the element_identifiers specify the condition names. I would even prefer this to a repeat.

BUT. This also does not work yet, but we are close with this https://github.com/galaxyproject/galaxy/pull/8599 ..

gregvonkuster:
> As an alternative for the repeat you might use a collection where the element_identifiers specify the condition names. I would even prefer this to a repeat.
> 
> BUT. This also does not work yet, but we are close with this #8599 ..

I agree.  This would be a nice change to the tool when the collection works.  The repeats were used because this tool was written about the time that collections were first being introduced into Galaxy.

test/functional/tools/ref_data_within_repeats.xml
```diff
@@ -0,0 +1,43 @@
+<tool id="ref_data_within_repeats" name="ref_data_within_repeats" version="1.0.0">
```
jmchilton, line 1: I think you need to add this to test/samples_tool_conf.xml to ensure that test runs.

PR #9197, opened 2020-01-09
Allow mixed store_by objectstore configurations.

jmchilton:
Allow a single Galaxy instance to store data both ways (both by UUID and by ID) - this should allow evolving toward remote extended metadata for existing big instances (main, EU, etc..). 

xref 
- https://github.com/galaxyproject/galaxy/pull/7154 (original PR adding storing by UUID)
- https://github.com/galaxyproject/galaxy/pull/7650 (followup PR fixing a bunch of edgier stuff)

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -171,6 +171,9 @@ class BaseAppConfiguration(object):
     def _in_sample_dir(self, path):
         return os.path.join(self.sample_config_dir, path)
 
+    def _in_data_dir(self, path):
+        return os.path.join(self.data_dir, path)
+
     def _parse_config_file_options(self, defaults, listify_defaults, config_kwargs):
         for var, values in defaults.items():
             if config_kwargs.get(var) is not None:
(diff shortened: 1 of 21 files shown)


mvdbeek (approved):
Very nice! I suppose the next step is to automatically use extended metadata collection if all outputs are created in a store by uuid object store ?

jmchilton:
I think the next step should probably actually be getting main or EU to use the a UUID object store for new files and then enable extended metadata manually? Then maybe consider that?

At least down this branch of the effort. I think more end-to-end testing with Pulsar, containers, etc... is needed. Gotta figure out what that looks like but certainly feeling emboldened by Github actions to do some cool things.

bgruening:
@jmchilton could you add some documentation about a potential upgrade strategy? Is it possible to bind certain job-destinations to certain object-stores? If so EU could create such an a new object store and move gradually tools over. This would make me more confident in trying it early.

ic4f:
@nsoranzo thank you - that's the steps I'm following. 
I think I misunderstood the `self._in_data_dir('files')` part when I was making that change - now I see it was *not* duplication with a hardcoded default. I'll fix this today.

ic4f:
The `file_path` issue discussed above has been resolved in #10617.

PR #9204, opened 2020-01-13
Fix sending email error on python-3.7

shenweiyan:
Fix sending email error( `ValueError: server_hostname cannot be an empty string or start with a leading dot.`) in python-3.7.

### DIFF
--- a/lib/galaxy/util/__init__.py
+++ b/lib/galaxy/util/__init__.py
@@ -1484,7 +1484,6 @@ def send_mail(frm, to, subject, body, config, html=None):
         s = smtplib.SMTP_SSL(config.smtp_server)
     else:
         s = smtplib.SMTP(config.smtp_server)
-    s.connect(config.smtp_server)
     if not smtp_ssl:
         try:
             s.starttls()
--- a/lib/galaxy/util/__init__.py
+++ b/lib/galaxy/util/__init__.py
@@ -1481,9 +1481,9 @@ def send_mail(frm, to, subject, body, config, html=None):
 
     smtp_ssl = asbool(getattr(config, 'smtp_ssl', False))
     if smtp_ssl:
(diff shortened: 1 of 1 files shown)


bernt-matthias:
What is the value of `config.smtp_server` in the case where it fails. 

I would have expected that this is caught here:

https://github.com/galaxyproject/galaxy/blob/a94abfefa779a42d555350a93ca8129c5fd19f2c/lib/galaxy/util/__init__.py#L1471

shenweiyan:
> What is the value of `config.smtp_server` in the case where it fails.
> 
> I would have expected that this is caught here:
> 
> https://github.com/galaxyproject/galaxy/blob/a94abfefa779a42d555350a93ca8129c5fd19f2c/lib/galaxy/util/__init__.py#L1471

Here is my test:
ValueError for python 3.7

lib/galaxy/util/__init__.py
```diff
@@ -1481,9 +1481,9 @@ def send_mail(frm, to, subject, body, config, html=None):
 
     smtp_ssl = asbool(getattr(config, 'smtp_ssl', False))
     if smtp_ssl:
-        s = smtplib.SMTP_SSL()
+        s = smtplib.SMTP_SSL(config.smtp_server)
     else:
-        s = smtplib.SMTP()
+        s = smtplib.SMTP(config.smtp_server)
     s.connect(config.smtp_server)
```
nsoranzo: The `connect()` is called by `SMTP()`/`SMTP_SSL()` if the `host` parameter is passed, so this line shouldn't be needed any more.

```suggestion
```

nsoranzo:
@galaxybot test this

nsoranzo:
@galaxybot test this

nsoranzo:
Thanks @shenweiyan for the bugfix!

PR #8570, opened 2019-09-04
Add api for async

FredericBGA:
Async mode is now available thought API

It seems to work, I can access to `/api/async/fasta_sequences/<>/<>?STATUS=OK&URL=<>&key=<>`.

There is still a lot of work to do or review:
- [ ] Functionality should not be copied and pasted between controllers
- [ ] API additions are more permanent changes to Galaxy than many other potential changes and so a second opinion on API changes should be sought.
- [ ] New API functionality should include functional tests.
- [ ] Changes to reflect modifications to the API should be pushed upstream to the BioBlend project if possible.

PR #9206, opened 2020-01-13
Api key in xml

FredericBGA:
Thi PR adds an new reserved variable `$__user_key__` in Galaxy Tool XML file (https://docs.galaxyproject.org/en/latest/dev/schema.html#reserved-variables)

It's very convenient for us, regarding #8570

I will understand if this PR is not be merged if it is considered as a security concern.


FredericBGA:
hum, sorry, I missed something, the whole API async commits are here...

PR #8897, opened 2019-10-28
[19.09] Structured access to Galaxy internals for (Interactive) Tools

jmchilton:

PR #9207, opened 2020-01-13
add user.api_key in XML

FredericBGA:
This PR adds an new reserved variable $__user_key__ in Galaxy Tool XML file (https://docs.galaxyproject.org/en/latest/dev/schema.html#reserved-variables)

It's very convenient for us, regarding #8570

I will understand if this PR is not be merged if it is considered as a security concern.

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -571,16 +571,21 @@ class User(Dictifiable, RepresentById):
             user_id = '%d' % user.id
             user_email = str(user.email)
             user_name = str(user.username)
+            api_keys = [one_key.key for one_key in user.api_keys]
+            # get the more up to date (is the tab sorted by id?)
+            user_key = api_keys[0]
         else:
             user = None
             user_id = 'Anonymous'
             user_email = 'Anonymous'
             user_name = 'Anonymous'
+            user_key = 'Anonymous'
         environment = {}
         environment['__user__'] = user
         environment['__user_id__'] = environment['userId'] = user_id
         environment['__user_email__'] = environment['userEmail'] = user_email
         environment['__user_name__'] = user_name
+        environment['__user_key__'] = user_key
         return environment
 
     @staticmethod


jmchilton:
In 19.09 we implemented an environment variable based method for getting this that seems mildly more secure. https://github.com/galaxyproject/galaxy/pull/8897/files#diff-b35b658d6793dc4534f7e4082aa38641

```xml
<environment_variables>
    <environment_variable name="GALAXY_API_KEY" inject="api_key" />
</environment_variables>
```

Is it possible to use that for your application? The PR says it was added for interactive tools - but I think the syntax should work for any tool?

FredericBGA:
ok, I will try this way. I'm still using 19.05. I will test as soon as I install the 20.01
thank you.

mvdbeek:
@FredericBGA were you satisfied with @jmchilton's suggestion ?

FredericBGA:
> 
> 
> @FredericBGA were you satisfied with @jmchilton's suggestion ?

I don't know. I didn't have time to test yet.

In a XML for an async tool: `tool_type="data_source_async"` I had created this PR in order to be able to do this kind of thing: `<param name="API_KEY" type="hidden" value="$__user_key__" />`
So the URL has an API_KEY parameter.

I will need to test if I can do something as simple as `<param name="API_KEY" type="hidden" value="$GALAXY_API_KEY" />`

FredericBGA:
@mvdbeek I've tried, I don't see how to use the injection of api_key with an async tool.

mvdbeek:
It's an environment variable, so you can reference this variable in your `command` section. Is that what you did ?

FredericBGA:
> 
> 
> It's an environment variable, so you can reference this variable in your `command` section. Is that what you did ?

sorry  if I misunderstand, but I'm working with an async tool (`tool_type="data_source_async"`), I need to build an URL with params. And these parameters are set using `<param>` in the `input` section of the XML. I used this PR to set the api key in such a parameter.
By the way I've tried to set up a "regular" tool that uses api injection, and it works. But api injection seems to be quite specific to regular tools.

bernt-matthias:
So you want that a param that defaults to the user's API key and allow that the user enters a different API key? 

Then you could use `""` as default and use the injected api key in this case in the command section, or?

Alternatively you may use parsec or bioblend to interact with galaxy, then you don't need to deal with URLs?

FredericBGA:
@bernt-matthias There is no Bioblend interaction right now for the asynchronous controller.
that's why I've submitted also PR: https://github.com/galaxyproject/galaxy/pull/8570. 
And as my galaxy instance is authenticated, I think that I need a key in order to access to this url `/api/async/fasta_sequences/`.
I can not use the API injection in the command line as I don't have any command line. The asynchronous tool is defined like this:
```
<?xml version="1.0"?>
    <tool name="test_async" id="test_async" tool_type="data_source_async" version="2.1">
    <description>test server</description>
    <command interpreter="python">data_source.py $output $__app__.config.output_size_limit</command>
    <inputs action="https://bioinformatics/test.cgi" check_values="false" method="get" target="_to
p">
        <display>go test server $GALAXY_URL</display>
        <param name="GALAXY_URL" type="baseurl" value="/api/async/test_async" />
        <param name="API_KEY" type="hidden" value="$__user_key__" />
    </inputs>
    <uihints minwidth="800"/>
    <outputs>
        <data name="output" format="fasta" />
    </outputs>
    <options sanitize="False" refresh="True"/>
</tool>
```
I followed the examples created by @hexylena (many thanks!): https://github.com/hexylena/galaxy-data_source-examples
When my data are ready on the test server, I need to get in touch with Galaxy (and send the URL of the file which Galaxy can feed the dataset). Without the API key I can't (I face the login page otherwise). That's why the user's API key is provided in the URL.

I hope that https://github.com/galaxyproject/galaxy/pull/8570 will be merged one day.

Or maybe I can do in another way?

jmchilton:
> I can not use the API injection in the command line as I don't have any command line.

You just want to modify that command with data_source.py right? I think you do have a command line.

FredericBGA:
> 
> 
> > I can not use the API injection in the command line as I don't have any command line.
> 
> You just want to modify that command with data_source.py right? I think you do have a command line.

thank you @jmchilton. So I need to:
- use the `<environment_variable name="GALAXY_API_KEY" inject="api_key" />` in the xml
- modify `tools/data_source/data_source.py` in order to use GALAXY_API_KEY (and set this as a CGI parameter) if it's defined.

I've tried to work on data_source.py without so much success, but I will retry.

FredericBGA:
Hello

I tried to avoid this modification of the model in this PR: https://github.com/galaxyproject/galaxy/pull/10323

This is the only way I found to get access (for data_source_async tools as the tool runner will create an URL, not a command line),  to the env variables that are defined in XML (`<environment_variables>`).

I would understand if this is to tricky and I can close these PR.

FredericBGA:
closed because https://github.com/galaxyproject/galaxy/pull/10323 works.

PR #9210, opened 2020-01-13
[20.01] Backport 9175: Update galaxy.tool_util.cwl for latest changes in CWL branch.

mvdbeek:


lib/galaxy/tool_util/cwl/parser.py
```diff
@@ -56,29 +64,42 @@
...
-def tool_proxy(tool_path=None, tool_object=None, strict_cwl_validation=True):
+
+def tool_proxy(tool_path=None, tool_object=None, strict_cwl_validation=True, tool_directory=None, uuid=None):
     """ Provide a proxy object to cwltool data structures to just
     grab relevant data.
     """
     ensure_cwltool_available()
-    tool = to_cwl_tool_object(
+    # if uuid is None:
+    #    raise Exception("tool_proxy must be called with non-None uuid")
```
nsoranzo, line 86: Is that something that should be removed or uncommented later?

lib/galaxy/tool_util/cwl/parser.py
```diff
@@ -56,29 +64,42 @@
...
     )
     return tool
 
 
-def tool_proxy_from_persistent_representation(persisted_tool, strict_cwl_validation=True):
+def tool_proxy_from_persistent_representation(persisted_tool, strict_cwl_validation=True, tool_directory=None):
+    """Load a ToolProxy from a previously persisted representation."""
     ensure_cwltool_available()
-    tool = to_cwl_tool_object(persisted_tool=persisted_tool, strict_cwl_validation=strict_cwl_validation)
+    if PERSISTED_REPRESENTATION == "cwl_tool_object":
```
nsoranzo, line 100: Isn't this always True?

lib/galaxy/tool_util/cwl/parser.py
```diff
@@ -94,28 +115,40 @@ def load_job_proxy(job_directory, strict_cwl_validation=True):
...
         from schema_salad import sourceline
         from schema_salad.ref_resolver import file_uri
-        uri = file_uri(os.getcwd()) + "/"
+        path = tool_directory
+        if path is None:
+            path = os.getcwd()
+        uri = file_uri(path) + "/"
         sourceline.add_lc_filename(tool_object, uri)
-        tool_object, _ = schema_loader.raw_document_loader.resolve_all(tool_object, uri)
+        # tool_object, _ = schema_loader.raw_document_loader.resolve_all(tool_object, uri, checklinks=False)
```
nsoranzo:
```suggestion
```

lib/galaxy/tool_util/cwl/schema.py
```diff
@@ -2,94 +2,88 @@
...
-            )
+        tool = load_tool.make_tool(
+            process_definition.uri,
+            process_definition.loading_context,
+        )
         return tool
 
 
 schema_loader = SchemaLoader()
 non_strict_schema_loader = SchemaLoader(strict=False)
```
nsoranzo:
```suggestion
```

lib/galaxy/tool_util/cwl/parser.py
```diff
@@ -1180,6 +1297,38 @@ def __init__(self, name, output_data_type, output_type, path=None, fields=None):
         self.fields = fields
 
 
+def get_outputs(path):
```
nsoranzo: This seems unused and introduce a (hidden) circular import, can we drop it?

jmchilton (approved):

lib/galaxy/tool_util/cwl/parser.py
jmchilton: It is an semi-open question in the CWL branch what the persisted representation should be. I think the current setting is right and once the plumbing is all merged and these things work - it will probably be this way - but I'm hesitant to commit to a path until the whole thing is ready to go and working.

lib/galaxy/tool_util/cwl/parser.py
jmchilton: This is used downstream - https://github.com/common-workflow-language/galaxy/blob/cwl-1.0/scripts/run_cwl.py#L104. I'd rather keep it and fix the circular import if possible?

lib/galaxy/tool_util/cwl/parser.py
jmchilton: This is a last minute change to get the usage from Planemo to pass - I'm happy to remove it I think. Probably should somehow make this required when loading tools from a toolbox and not requires from library behavior but I can just create an issue for that.

jmchilton: https://github.com/common-workflow-language/galaxy/issues/113

lib/galaxy/tool_util/cwl/parser.py
nsoranzo: @jmchilton Sure, can you give it a try?

jmchilton: I misunderstood the cycle at first - yeah I'll move them into a new file in the CWL branch and we can just remove them here.

nsoranzo: Cool, I've remove this method and the 2 following classes for now then. Thanks!

jmchilton: https://github.com/galaxyproject/galaxy/commit/100e4401dac734b7c2f30df05d5de0e683b02d96 - fixed downstream. Thanks for cleaning this up @nsoranzo

jmchilton:
Thanks for all the work on this. It may seem like a big mess - but it is a huge help to the CWL effort to have this synchronized every so often. I'll try to publish new 20.1 dev packages now

jmchilton:
galaxy-tool-util-20.1.0.dev0 and galaxy-util-20.1.0.dev0 have been published. I'm not sure I see a reason to publish a galaxy-containers-20.1.0.dev0 since it is unchanged. I've created a minimum dependency of galaxy-util-20.1.0.dev0 on the package galaxy-tool-util-20.1.0.dev0.

PR #9213, opened 2020-01-14
Provide clues for advanced history search

assuntad23:
**Why?**
To increase awareness and ease of access to Advanced Searching in Histories. 

**How?**
Added a button/link to Advanced Searching documentation to the input-query component, as an optional (default is null) value.

**How to Test:** 
1. Pull my branch
2. sh run.sh
3. Click on the ? icon in the Histories panel search box. Note how it takes you to the documentation.
4. Type input in and note how the text does not go behind the icons.
5. Notice how the search box in the tool panel does not have an icon, despite using a common component.  
6. Navigate to "view all histories". Note: All of the search input boxes here are are also equipped with Advanced Search button/link.  
7. Ask me if you have any questions. I appreciate feedback.


martenson:
I like the take and especially the small touches like not hiding the query text behind the icons. 

From what I understand of the original issue it would be nice to have something like a popover element that would highlight the most useful parts of the syntax without leaving the page but also contain the link to the full documentation, as it has now.

martenson:
@assuntad23 I'll be happy to review this. Could you please rebase to resolve the conflicts? Let me know if you want me to walk you through how.

martenson:
Hi @assuntad23. I have resolved the conflict locally however I think you need to make some adjustments to make this work given the Galaxy's styles in these parts have changed since you implemented this.

I also think that the popover that shows you the information right away, instead of sending you to an entirely different context (galaxy hub site), would be of real value here. I think we already include the boostrap-vue library which [implements it](https://bootstrap-vue.org/docs/components/popover). Please let me know if you want some more guidance.

Thanks for working on this.

assuntad23:
@martenson, I've created a new Pull Request for this, because of some merge issues. 

https://github.com/galaxyproject/galaxy/pull/9696

I've changed it to a popover, with the link to full documentation embedded. However, I was not easily able to prevent text from going behind the icon, as I did here, because when I attempted to (by changing the width to 85%), it messed up the clear icon in the tools panel.

nsoranzo:
Closing in favour of #9696.

PR #9214, opened 2020-01-14
[20.01] Throw error if authnz attempts to create new user with existing email

guerler:
It seems that the authnz process does create users with already existing emails. This PR adds check flags to prevent this from happening. This PR requires a follow-up in order to understand why the authnz module does attempt to create a new user for an already existing user.


guerler:
Superseded by #9233.

PR #9215, opened 2020-01-14
[20.01] Fix extended base for singularity and add base_image_for_targets

mvdbeek:
This is easier to use in planemo (we don't need to figure out which
image to use based on the return value of
any_target_requires_extended_base) and leaves open the possibility for
custom base image annotation in conda recipes or other places.


mvdbeek:
Thanks for the merge @jmchilton! Could we get another galaxy-tool-util pre-release with these changes ?

jmchilton:
On it...

Edited: Done!

Issue #9205, opened 2020-01-13
Two separate Trackster issues (with fixes)


kaukrise:
Hi Galaxy developers,

first of all, I love what you have achieved with this project, your work is much appreciated.
In the process of testing a new Galaxy installation on our local cluster, I encountered several issues with the Trackster visualisation, two of which also seem to be present on usegalaxy.org:

1. **The genome navigation bar is not visible**. To reproduce, simply open any Trackster visualisation. I saw that the `unified-panel-body` `div` has an `overflow: visible` style attribute, which is supposed to make the navigation bar visible outside of its `div`. This is applied [here](https://github.com/galaxyproject/galaxy/blob/d6de316b5392d5d62a599b08a08c25fbd93d0dac/client/galaxy/scripts/viz/trackster/tracks.js#L1005). However, the default CSS stylesheet overrides this with an `!important` flag.

2. **Dataset history is not loading when adding a new track**. To reproduce, click the "add track" button in a Trackster visualisation - you will see the loading wheel spin but never fully load. The logs reveal that this is caused by an `sqlalchemy` bug which requires [this statement](https://github.com/galaxyproject/galaxy/blob/d6de316b5392d5d62a599b08a08c25fbd93d0dac/lib/galaxy/webapps/galaxy/controllers/visualization.py#L47) to be wrapped in `text()`.

The fixes are fairly straightforward in both cases, and I would be happy to submit pull requests for both. I read the contribution guidelines, so would it be acceptable if I created two separate pull requests, each based on the `20.01` branch? Or would you prefer them to be based on `dev`?

Thank you!

Kai

nsoranzo:
@kaukrise Thanks for the bug reports and the offer to contribute, much appreciated!
You surely can open pull requests against the (yet unreleased) release_20.01 branch, or even release_19.05 or release_19.09, which are the 2 presently supported releases (if the issue is present there too).

dannon:
Thanks for both of these fixes @kaukrise.  They've both been merged and merged forward through 20.01 to dev as well.

kaukrise:
Thank you guys, that was incredibly quick!

PR #9216, opened 2020-01-14
Wrap SQL statement in a text() function as required by sqlalchemy

kaukrise:
Fixes a history loading bug in Trackster. The issue this fixes is #9205

### DIFF
--- a/lib/galaxy/webapps/galaxy/controllers/visualization.py
+++ b/lib/galaxy/webapps/galaxy/controllers/visualization.py
@@ -11,7 +11,14 @@ from paste.httpexceptions import (
     HTTPNotFound
 )
 from six import string_types
-from sqlalchemy import and_, desc, false, or_, true, text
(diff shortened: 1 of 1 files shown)


lib/galaxy/webapps/galaxy/controllers/visualization.py
```diff
@@ -11,7 +11,7 @@
     HTTPNotFound
 )
 from six import string_types
-from sqlalchemy import and_, desc, false, or_, true
+from sqlalchemy import and_, desc, false, or_, true, text
```
nsoranzo:
```suggestion
from sqlalchemy import (
    and_,
    desc,
    false,
    or_,
    text,
    true,
)
```

nsoranzo:
@galaxybot test this

nsoranzo (approved):

dannon:
@kaukrise Thanks!

PR #9219, opened 2020-01-14
[20.01] Fix: Added support for specifying k8s job pod priority class

nuwang:
So that jobs don't evict the Galaxy pod when resources are scarce.


mvdbeek (approved):

mvdbeek:
thanks @nuwang !

afgane (approved):

PR #9221, opened 2020-01-15
Added HiGlass interactive environment

msauria:
This includes adding the pairix and BedSQlite datatypes, a converter for interval to pairix, and explicitly creating a chromosome length datatable.


lib/galaxy/datatypes/converters/interval_to_pairix_converter.py
```diff
@@ -0,0 +1,69 @@
...
+            end2Col = i + 1
+    col_line = False
+    for line in open(input_name, 'r'):
+        if not line.startswith('#'):
+            break
+        if line.startswith('#columns:'):
+            col_line = True
+            break
+    if not col_line and min(chrom1Col, start1Col, end1Col, chrom2Col, start2Col, end2Col) == 0:
+        print >> sys.stderr, ("Not all of the required column information is present")
```
jdavcs: this will break under python3. The following should solve it:
```from __future__ import print_function``` (add to the import stms at the top)
```print("Not all of the required column information is present", file=sys.stderr)```

File: lib/galaxy/tools/data/__init__.py [hunk 2]
```diff
@@ -78,6 +78,7 @@ def __init__(self, tool_data_path, config_filename=None, tool_data_table_config_
             if not single_config_filename:
                 continue
             self.load_from_config_file(single_config_filename, self.tool_data_path, from_shed_config=False)
+            self.load_length_table(len_file_path, self.tool_data_path, single_config_filename)
```
jdavcs: Very minor suggesion. Maybe call this only if `len_file_path` is present? It's not needed otherwise, and you'll be able to get rid of lines 241-242. Seems more straightforward, I think
```python
if len_file_path:
    self.load_length_table(len_file_path, self.tool_data_path, single_config_filename)
```

jdavcs: thank you!

jdavcs (approved):

jdavcs:
@msauria did I label these correctly?

lib/galaxy/datatypes/interval.py
```diff
@@ -1447,6 +1451,116 @@ def sniff(self, filename):
...
+                dataset.metadata.end2Col = i + 1
+            elif name == 'start2':
+                dataset.metadata.start2Col = i + 1
+            elif name == 'end2':
+                dataset.metadata.end2Col = i + 1
+        index_file = dataset.metadata.pairix_index
+        if not index_file:
+            index_file = dataset.metadata.spec['pairix_index'].param.new_file(dataset=dataset)
+        try:
+            pypairix.build_index(dataset.file_name, sc=dataset.metadata.chromCol, bc=dataset.metadata.startCol,
```
mvdbeek: Is it absolutely necessary to add pypairix ? This seems like quite a heavy-weight dependency, it would be great if we could avoid this. If the indexing isn't very heavy this could be done in the tool itself for instance.

lib/galaxy/datatypes/interval.py
```diff
@@ -1447,6 +1451,116 @@ def sniff(self, filename):
...
+                                 ec=dataset.metadata.endCol, sc2=dataset.metadata.chrom2Col,
+                                 bc2=dataset.metadata.start2Col, ec2=dataset.metadata.end2Col,
+                                 delimiter=dataset.metadata.delimiter, force=1)
+            subprocess.Popen(['mv', '%s.px2' % dataset.file_name, index_file.file_name])
```
mvdbeek:
```suggestion
            shutil.move( '%s.px2' % dataset.file_name, index_file.file_name)
```

mvdbeek: (and a `shutil` import at the top)

mvdbeek (commented):
Was it intentional to add `test-data/matrix.mcool` ? It seems unused, but if it is it should go to https://github.com/galaxyproject/galaxy-test-data/

lib/galaxy/datatypes/interval.py
jdavcs: You mean add the pypairix dependency *to the tool* instead of galaxy?

mvdbeek: Yes, and move the index generation to the tool as well.

jdavcs: Yes, makes sense. @msauria I'll be happy to help if needed.

msauria:
It seems reasonable to move the indexing to the tool, given the specialized datatype.

lib/galaxy/datatypes/binary.py
```diff
@@ -206,12 +206,14 @@ def matches_any(self, target_datatypes):
...
         uncompressed_target_datatypes = []
-
+        log.info("%s" % (str(target_datatypes)))
         for target_datatype in target_datatypes:
             if hasattr(target_datatype, "uncompressed_datatype_instance") and target_datatype.compressed_format == self.compressed_format:
                 uncompressed_target_datatypes.append(target_datatype.uncompressed_datatype_instance)
             else:
                 compressed_target_datatypes.append(target_datatype)
+        log.info("%s" % (str(uncompressed_target_datatypes)))
+        log.info("%s" % (str(compressed_target_datatypes)))
```
mvdbeek, line 216:
```suggestion
```

lib/galaxy/config/sample/datatypes_conf.xml.sample
```diff
@@ -379,6 +384,8 @@
     </datatype>
     <datatype extension="interval_index" type="galaxy.datatypes.binary:Binary" subclass="true"/>
     <datatype extension="tabix" type="galaxy.datatypes.binary:Binary" subclass="true"/>
+    <datatype extension="gzip" type="galaxy.datatypes.binary:Binary" subclass="true"/>
```
mvdbeek, line 387: why was this and bz2 added ?

lib/galaxy/datatypes/binary.py
```diff
@@ -206,12 +206,14 @@ def matches_any(self, target_datatypes):
         """
         compressed_target_datatypes = []
         uncompressed_target_datatypes = []
-
+        log.info("%s" % (str(target_datatypes)))
```
mvdbeek, line 209:
```suggestion
```

lib/galaxy/datatypes/converters/interval_to_pairix_converter.xml
```diff
@@ -0,0 +1,17 @@
+<tool id="CONVERTER_interval_to_pairix" name="Convert Interval to Pairix" version="1.0.0">
+    <!-- <description>__NOT_USED_CURRENTLY_FOR_CONVERTERS__</description> -->
+    <!-- Used on the metadata edit page. -->
```
mvdbeek, line 3:
```suggestion
```

lib/galaxy/datatypes/converters/interval_to_pairix_converter.xml
```diff
@@ -0,0 +1,17 @@
...
+    <command><![CDATA[
+python '$__tool_directory__/interval_to_pairix_converter.py'
+'$input' '$output' $input.metadata.column_names
```
mvdbeek, line 6:
```suggestion
'$input' '$output' '$input.metadata.column_names'
```

lib/galaxy/datatypes/converters/interval_to_pairix_converter.xml
```diff
@@ -0,0 +1,17 @@
...
+    
+]]></command>
+    <inputs>
+        <param name="input" type="data" format="interval" label="Interval file"/>
+    </inputs>
+    <outputs>
+        <data name="output" format="pairix"/>
+    </outputs>
```
mvdbeek, line 14: Can you add a test here please ?

lib/galaxy/datatypes/converters/interval_to_pairix_converter.xml
```diff
@@ -0,0 +1,17 @@
...
+    <help>
+    </help>
```
mvdbeek, line 16:
```suggestion
```

lib/galaxy/datatypes/interval.py
```diff
@@ -1447,6 +1449,104 @@ def sniff(self, filename):
...
+        return (('chr' in columns or 'chr1' in columns)
+                & ('pos' in columns or 'pos1' in columns or 'start' in columns or 'start1' in columns)
+                & ('chr2' in columns)
+                & ('pos2' in columns or 'start2' in columns))
+
+    def sniff_prefix(self, file_prefix):
+        return self.sniff(file_prefix.filename)
+
+    def set_meta(self, dataset, **kwd):
+        """ Creates the index for the pairix file. """
```
mvdbeek, line 1501:
```suggestion
```

lib/galaxy/datatypes/interval.py
```diff
@@ -1447,6 +1449,104 @@ def sniff(self, filename):
...
+        # the start coordinate, and if a stop coordinate is present.
+        # In order to produce a pairx index, we need to look at the
+        # required header line "#columns:". The sniffer should already
+        # have ensured that this line is present.
+        Tabular.set_meta(self, dataset, skip=0)
+        file_iterator = util.compression_utils.get_fileobj(dataset.file_name)
+        sep_character = None
+        comment_lines = 0
+        for line in file_iterator:
+            log.info(line)
```
mvdbeek, line 1512:
```suggestion
```

lib/galaxy/tools/data/__init__.py
```diff
@@ -785,6 +794,42 @@ def to_dict(self):
         return rval
 
 
+class LengthToolDataTable(TabularToolDataTable, Dictifiable):
```
mvdbeek, line 797: What is this needed for ?

lib/galaxy/datatypes/interval.py
```diff
@@ -1447,6 +1449,104 @@ def sniff(self, filename):
...
+            if sep_character is not None:
+                break
+            elif len(line) > 0 and line[0] != '#':
+                fields = line.split()
+                if len(fields) > 1:
+                    sep_character = line.split(fields[1])[0].split(fields[0])[-1]
```
mvdbeek, line 1518: I don't think I understand what's going on here. Shouldn't sep_character always be tab ?

lib/galaxy/datatypes/interval.py
```diff
@@ -1447,6 +1449,104 @@ def sniff(self, filename):
...
+                comment_lines += 1
+                line = line.rstrip('\n\r')
+                if line.startswith('#columns: '):
+                    line = line.lower().replace('chrom', 'chr').replace('chromosome', 'chr').replace('stop', 'end')
+                    columns = line.split()[1:]
+        dataset.metadata.delimiter = sep_character
+        dataset.metadata.comment_lines = comment_lines
+        dataset.metadata.column_names = columns
+        dataset.metadata.columns = len(columns)
+        for i, name in enumerate(columns):
```
mvdbeek, line 1529:
```suggestion
        for i, name in enumerate(columns, 1):
```
and then drop the `+ 1` below. 

From the specs (https://github.com/4dn-dcic/pairix/blob/master/pairs_format_specification.md#standard-format) I understand the first 7 columns are required, reserved and in a fixed order, so can we assert that ?

lib/galaxy/datatypes/converters/interval_to_pairix_converter.xml
```diff
@@ -0,0 +1,17 @@
+<tool id="CONVERTER_interval_to_pairix" name="Convert Interval to Pairix" version="1.0.0">
+    <!-- <description>__NOT_USED_CURRENTLY_FOR_CONVERTERS__</description> -->
```
mvdbeek, line 2: Can you add the dependencies on bgzip (or is that samtools/htslib ?) and coreutils here ?

mvdbeek (commented):
I suppose we still want the datatypes and the converter, but I believe we're going to get the Interactive tool version of higlass.

mvdbeek:
@astrovsky01 do you want to take over this PR ?

astrovsky01:
I thought that we did the IT instead of the IE in the end?

mvdbeek:
Don't we still need the datatypes ?

astrovsky01:
They're not used in the current version of the IT. If they are added, we can put it in then, but doing so now would just add extra datatypes.

mvdbeek:
Sounds good.

PR #9222, opened 2020-01-15
Properly quote paths in upload tool

bernt-matthias:
We try to make sure that paths are properly quoted for all tools in IUC, etc but we missed upload... 

So while playing around with https://github.com/galaxyproject/tools-iuc/issues/577 the first thing that failed was upload.

### DIFF
--- a/tools/data_source/upload.xml
+++ b/tools/data_source/upload.xml
@@ -3,7 +3,7 @@
   <description>from your computer</description>
   <action module="galaxy.tools.actions.upload" class="UploadToolAction"/>
   <command>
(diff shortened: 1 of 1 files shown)


bernt-matthias:
In the end we know now for sure, that nobody ever had a path with spaces that is relevant for galaxy job execution (galaxy root, file path, job working dir,... ). Maybe it would be a good idea to check this prior to executing galaxy.. Just to make sure that is stays like this.

nsoranzo:
Or we can modify the checkout step in `.github/workflows/integration.yaml` by setting a `path:` containing a space to test this scenario.

bernt-matthias:
> Or we can modify the checkout step in .github/workflows/integration.yaml by setting a path: containing a space to test this scenario.

definitely worth a try.. Just to see what else fails.

PR #9223, opened 2020-01-16
Use a Galaxy root path with a space for integration tests

nsoranzo:
and fix uncovered bugs.

### DIFF
--- a/lib/galaxy/tool_util/deps/commands.py
+++ b/lib/galaxy/tool_util/deps/commands.py
@@ -118,16 +118,12 @@ def _wait(cmds, **popen_kwds):
     return stdout
 
 
-def download_command(url, to=STDOUT_INDICATOR, quote_url=False):
+def download_command(url, to=STDOUT_INDICATOR):
     """Build a command line to download a URL.
 
     By default the URL will be downloaded to standard output but a specific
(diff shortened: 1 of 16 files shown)


lib/galaxy/tool_util/deps/commands.py
```diff
@@ -118,16 +118,12 @@ def _wait(cmds, **popen_kwds):
     return stdout
 
 
-def download_command(url, to=STDOUT_INDICATOR, quote_url=False):
+def download_command(url, to=STDOUT_INDICATOR):
```
jmchilton, line 121: quote_url is still used in Planemo right? Why break backward compatibility on this?

nsoranzo: I forgot about that! It's quite unsecure though; if I fix it in Planemo, would it be OK? Or would you like to keep it any way?

nsoranzo:
Green and ready for review!

lib/galaxy/tool_util/deps/commands.py
nsoranzo: Done in https://github.com/galaxyproject/planemo/pull/984

mvdbeek (approved):

PR #9224, opened 2020-01-16
[20.01] Extended base annotation fixes

mvdbeek:
Sort best package by upload time (lowest priority) in addition to version and build number.

This is necessary for getting the extended-base annotation for qiime 1.9.1,
which does not contain the annotation in
linux-64/qiime-1.9.1-np110py27_1.tar.bz2 while in the newer
linux-64/qiime-1.9.1-np112py27_1.tar.bz2 it is present. This depends of course
on `conda search` returning packages sorted by upload time and is a little
brittle. It does seem better than always picking the oldest upload time as we
do without this change.  Also looks in meta.yaml for extended base annotation,
since sometimes this isn't present in info/about.json.

### DIFF
--- a/lib/galaxy/tool_util/deps/conda_util.py
+++ b/lib/galaxy/tool_util/deps/conda_util.py
@@ -526,7 +526,11 @@ def best_search_result(conda_target, conda_context, channels_override=None, offl
     try:
         res = commands.execute(search_cmd)
         res = unicodify(res)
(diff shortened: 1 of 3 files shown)


jmchilton:
I'll mint a new release!

mvdbeek:
Thanks!

jmchilton:
Done - thanks @mvdbeek !

PR #9225, opened 2020-01-16
[20.01] Fix: Added support for specifying k8s job affinity and tolerations

nuwang:
This patch allows job placement to be controlled on k8s, as without it, jobs can sometimes overwhelm the head node.

### DIFF
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -8,6 +8,8 @@ import os
 import re
 from time import sleep
 
+import yaml
+
 from galaxy import model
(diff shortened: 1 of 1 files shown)


mvdbeek (approved):
Thanks @nuwang!

nuwang:
Linting was failing due to a missing newline, so I added that in.

nuwang:
Tests that passed earlier seem to be failing now due to an unrelated error.

mvdbeek:
Yeah, I think something broke the cached venv in the integration test, https://github.com/galaxyproject/galaxy/pull/9228/commits/cc36560c64f92cf72ebf7c991dd84821fadc5f23 should be one way to fix it.

PR #9228, opened 2020-01-17
[20.01] More mulled-build-files fixes

mvdbeek:


lib/galaxy/tool_util/deps/mulled/mulled_build.py
```diff
@@ -69,6 +69,7 @@
     cp -r /data/dist/* /tmp/conda/
 
 %%post
+    rm -R /usr/local || true
```
bgruening, line 72: The PR looks good to me, I'm just not sure about this one and if there is anything in this dir by default.

Also I'm surprised that this code path is still used by default, as we switched to autoconvert Docker-to-Singularity a few month ago.

Thanks @mvdbeek

mvdbeek: > The PR looks good to me, I'm just not sure about this one and if there is anything in this dir by default.

Just the folder structure and no files:
```
root@5aa74404c62d:/# ls -l -r /usr/local
total 32
drwxrwsr-x 2 root staff 4096 Dec 29 15:04 src
drwxrwsr-x 3 root staff 4096 Dec 29 15:04 share
drwxrwsr-x 2 root staff 4096 Dec 29 15:04 sbin
lrwxrwxrwx 1 root staff    9 Dec 29 15:04 man -> share/man
drwxrwsr-x 2 root staff 4096 Dec 29 15:04 lib
drwxrwsr-x 2 root staff 4096 Dec 29 15:04 include
drwxrwsr-x 2 root staff 4096 Dec 29 15:04 games
drwxrwsr-x 2 root staff 4096 Dec 29 15:04 etc
drwxrwsr-x 2 root staff 4096 Dec 29 15:04 bin
```

jmchilton:
New package released, thanks!

PR #9233, opened 2020-01-17
[20.01] External auth login refinement

dannon:
Explains to the user that (currently -- though there is WIP to address this where you'll get a screen like "Hey, this user already exists -- do you want to link these accounts?  If so, enter your credentials for...") you must first be logged in as the user you'd like to associate with, and prevents duplicate creation in the process.
Fixes a corner case in login loop, logout clearer and more robust.
Cleaner URL (doesn't leave /callback cruft there for the user) by using redirect.


guerler (approved):
Awesome, this also fixes the landing page issue upon login. Works well for me and looks good. Thank you.

dannon:
@guerler Thanks for testing it!

PR #9234, opened 2020-01-18
[20.01] Mount job_working_directory/configs rw

mvdbeek:
Many tools have this pattern where they move the config file somewhere,
which doesn't work if it is mounted read-only.

### DIFF
--- a/lib/galaxy/tool_util/deps/container_classes.py
+++ b/lib/galaxy/tool_util/deps/container_classes.py
@@ -179,7 +179,9 @@ class HasDockerLikeVolumes(object):
             if self.job_info.tool_directory:
                 defaults += ",$tool_directory:default_ro"
             if self.job_info.job_directory:
-                defaults += ",$job_directory:default_ro,$job_directory/outputs:rw,$job_directory/configs"
+                defaults += ",$job_directory:default_ro,$job_directory/outputs:rw"
+                if self.tool_info.profile <= 19.09:
+                    defaults += ",$job_directory/configs:rw"
             if self.job_info.tmp_directory is not None:
                 defaults += ",$tmp_directory:rw"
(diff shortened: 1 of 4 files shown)


bernt-matthias:
Is this related: https://github.com/galaxyproject/galaxy/pull/9236?

mvdbeek:
I would say they're separate.

lib/galaxy/tool_util/deps/container_classes.py
```diff
@@ -179,7 +179,7 @@ def add_var(name, value):
             if self.job_info.tool_directory:
                 defaults += ",$tool_directory:default_ro"
             if self.job_info.job_directory:
-                defaults += ",$job_directory:default_ro,$job_directory/outputs:rw"
+                defaults += ",$job_directory:default_ro,$job_directory/outputs:rw,$job_directory/configs"
```
jmchilton: Should this end in ``:rw``?

jmchilton:
Many tools? That is gross - it prevents the tools from being re-submittable. Any interest in limiting the profile version of this behavior?

lib/galaxy/tool_util/deps/container_classes.py
mvdbeek: It should, I wonder why this seems to work regardless ?

mvdbeek:
> Many tools? That is gross - it prevents the tools from being re-submittable. Any interest in limiting the profile version of this behavior?

Not that many (maybe 10 in all of the iuc), and it's usually the last step in the tool. We could also include the configs dir in https://github.com/galaxyproject/galaxy/pull/9245, that'd take care of that as well.

mvdbeek:
Alright I made the configs:rw conditional on the tool profile and made this safe for resubmission.

PR #1405, opened 2015-12-29
Implement special config file to dump parameters as json.

jmchilton:
```
    <configfiles>
        <inputs name="inputs" />
    </configfiles>
```

Version is specified as 1, because the Galaxy tool plumbing doesn't really distinguish between parameter is '', parameter was explicitly supplied as None, parameter was absent, etc... - everything is a just a string in some ways. This distinction may be more important with JSON and for instance the CWL will require more precision, so we may want to revise Galaxy's tool handling in a backward compatible way someday.

Both @erasche  and @bgruening are working on tools that seem they would benefit from this.

PR #9236, opened 2020-01-18
inputs configfile: test for filename and data_style attributes

bernt-matthias:
Shows a potential problem: with `filename` the file is created in the JOBDIR not JOBDIR/working.

Also I was expecting (or hoping) that also outputs are added to the json.

### DIFF
--- a/test/functional/tools/inputs_as_json.xml
+++ b/test/functional/tools/inputs_as_json.xml
@@ -1,9 +1,9 @@
 <tool id="inputs_as_json" name="inputs_as_json" version="1.0.0">
     <command detect_errors="exit_code">
-        python $check_inputs $inputs $test_case
+        python $check_inputs inputs.json $test_case
     </command>
     <configfiles>
-        <inputs name="inputs" />
+        <inputs name="inputs" filename="inputs.json" />
         <!-- Can specify with fixed path in working directory instead:
         <inputs name="inputs" filename="input.json" />
         -->
@@ -21,6 +21,7 @@ def assert_equals(x, y):
 
 if test_case == "1":
     assert_equals(as_dict["test_case"], 1)
+    assert "file_test" not in as_dict
     assert_equals(as_dict["text_test"], "foo")
     assert_equals(as_dict["booltest"], True)
     assert_equals(as_dict["booltest2"], True)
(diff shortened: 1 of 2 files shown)


mvdbeek:
are you sure they're created in the jobdir ? https://github.com/mvdbeek/galaxy/blob/f93e4d6121496d5d90e4550d410d02bcc305dbcf/lib/galaxy/tools/evaluation.py#L528 should place them in the `configs/` dir.

I think this might fix it?
```
diff --git a/lib/galaxy/tools/evaluation.py b/lib/galaxy/tools/evaluation.py
index c21be02f8b..1c178d8b28 100644
--- a/lib/galaxy/tools/evaluation.py
+++ b/lib/galaxy/tools/evaluation.py
@@ -525,6 +525,7 @@ class ToolEvaluator(object):
             if not os.path.exists(directory):
                 os.makedirs(directory)
             if filename is not None:
+                directory = os.path.join(self.local_working_directory, "working")
                 config_filename = os.path.join(directory, filename)
             else:
                 fd, config_filename = tempfile.mkstemp(dir=directory)
```

bernt-matthias:
> should place them in the configs/ dir. 

Correct. Got this wrong somehow. 

> I think this might fix it?

Indeed the file is then created in `working`. But the problem is that this dir is deleted by the job script https://github.com/galaxyproject/galaxy/blob/c53fb623d8cf3af99732a3fe9d1f513f9b538c2b/lib/galaxy/jobs/command_factory.py#L104. 

Thinking about this I get the impression that we should drop the `filename` attribute of inputs. I don't see a usecase of generating the script with a specific name in `config`. The tool developer can still copy it to `working` using whatever name she desires.

Seems that `inputs` were implemented here https://github.com/galaxyproject/galaxy/pull/1405 already with the filename attribute, but since the docs were not correct so far might be safe to drop... 

@jmchilton since you implemented this originally: what is your take on this?

mvdbeek:
> Indeed the file is then created in `working`. But the problem is that this dir is deleted by the job script

This is a problem also for data managers that precreate an output file in ~~that directory~~ the `outputs` directory that also get's deleted on resubmission. I think we should do something about this, but I'm not sure what.

mvdbeek:
I've cherry-picked your commits into https://github.com/galaxyproject/galaxy/pull/9245, they're actually a nice test for this.

mvdbeek:
I'll close this since https://github.com/galaxyproject/galaxy/pull/9245 got merged. Thanks @bernt-matthias!

Issue #9241, opened 2020-01-20
Exception during tool execution


mvdbeek:
Seen in
https://github.com/mvdbeek/tools-iuc/runs/398680061#step:10:633:
```
2020-01-20T10:34:14.9786332Z 2020-01-20 10:34:14,974 ERROR [galaxy.tools] Exception caught while attempting tool execution:
2020-01-20T10:34:14.9787128Z Traceback (most recent call last):
2020-01-20T10:34:14.9789375Z   File "/home/runner/.planemo/gx_venv_3.7_release_20.01/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1246, in _execute_context
2020-01-20T10:34:14.9790073Z     cursor, statement, parameters, context
2020-01-20T10:34:14.9790998Z   File "/home/runner/.planemo/gx_venv_3.7_release_20.01/lib/python3.7/site-packages/sqlalchemy/engine/default.py", line 581, in do_execute
2020-01-20T10:34:14.9791567Z     cursor.execute(statement, parameters)
2020-01-20T10:34:14.9792168Z psycopg2.errors.UndefinedFunction: operator does not exist: integer = boolean
2020-01-20T10:34:14.9792461Z LINE 3: WHERE history_dataset_association.id = true
2020-01-20T10:34:14.9792893Z                                              ^
2020-01-20T10:34:14.9793157Z HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
2020-01-20T10:34:14.9793379Z 
2020-01-20T10:34:14.9793559Z 
2020-01-20T10:34:14.9793800Z The above exception was the direct cause of the following exception:
2020-01-20T10:34:14.9793992Z 
2020-01-20T10:34:14.9794231Z Traceback (most recent call last):
2020-01-20T10:34:14.9794979Z   File "/tmp/tmpvahl86j6/galaxy-dev/lib/galaxy/tools/__init__.py", line 1517, in handle_single_execution
2020-01-20T10:34:14.9803112Z     collection_info=collection_info,
2020-01-20T10:34:14.9804371Z   File "/tmp/tmpvahl86j6/galaxy-dev/lib/galaxy/tools/__init__.py", line 1599, in execute
2020-01-20T10:34:14.9845469Z     return self.tool_action.execute(self, trans, incoming=incoming, set_output_hid=set_output_hid, history=history, **kwargs)
[...]

PR #9242, opened 2020-01-20
[19.09] Limit async hack to async tools

mvdbeek:
Fixes disco tests and https://github.com/galaxyproject/galaxy/issues/9241


jmchilton:
Thanks!

Issue #3289, opened 2016-12-08
Ensure clean job working directory if job was re-queued


Takadonet:
Want to ensure that the job working directory is always empty when a job starts up.

In our cluster environment, jobs can be requeue by slurm because a higher priority job has bump it off or a node has failed. When the job is re-queue using slurm, it will still have the same job ID and working directory but its <job_directory>/working could have some partial datasets already because it already ran but not to completion.  As such most tools expect to have an empty working/ .

My solution would be to add _rm -rf working/_ at 
https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/jobs/command_factory.py#L89
i.e
commands_builder.prepend_command("**rm -rf working/ ;** mkdir -p working; cd working")

jmchilton:
Simple solution - that should probably work. I like it. I'll open a PR.

Takadonet:
Sweet.

PR #9245, opened 2020-01-21
[20.01] Fix deletion of outputs and working before job

mvdbeek:
Both directories may contain important files for a job, so deleting these directories may be inappropriate.
We originally implemented this for https://github.com/galaxyproject/galaxy/issues/3289

I noticed this with data managers (and potentially other OutputParameterJSONTool tools) that populate datasets in the `outputs/` directory if using `outputs_to_working_directory`. These used to be pre-populated in the job working directory, but with https://github.com/galaxyproject/galaxy/commit/ad573b215802aaaa9d09e1d2e3e2974e79295dfe they are moved to `outputs`/ (which is good), but which gets deleted during job setup.

The approach here is to copy the `working` and `outputs` directories to `_working` and `_outputs` before running the actual jobs, and if re-running the job deleting `working` and `outputs` and copying the backups back into place.
This seems to be the most robust way to make sure that resubmission by the DRM will start with proper configfiles and pre-populated outputs. The cost should be minimal as most of the time these are empty
directories.
This would also help with #9236 (comment)

### DIFF
--- a/lib/galaxy/jobs/command_factory.py
+++ b/lib/galaxy/jobs/command_factory.py
@@ -99,15 +99,13 @@ def build_command(
         # usually working will already exist, but it will not for task
         # split jobs.
 
-        # Remove the working directory incase this is for instance a SLURM re-submission.
+        # Remove all non-readonly files in the workfing directory incase this is a re-submission.
         # xref https://github.com/galaxyproject/galaxy/issues/3289
-        commands_builder.prepend_command("""# Delete working/ and outputs/ only if job was resubmitted
+        commands_builder.prepend_command("""
 if [ -f .job_started ]; then
-    rm -rf working outputs; mkdir -p working outputs
+    find working outputs -perm /0222  -delete 2> /dev/null || true
 fi
-touch .job_started
-cd working
-""", sep='')
+mkdir -p working outputs; touch .job_started; cd working""")
 
     container_monitor_command = job_wrapper.container_monitor_command(container)
     if container_monitor_command:
--- a/lib/galaxy/jobs/command_factory.py
+++ b/lib/galaxy/jobs/command_factory.py
@@ -19,6 +19,13 @@ YIELD_CAPTURED_CODE = 'sh -c "exit $return_code"'
 SETUP_GALAXY_FOR_METADATA = """
 [ "$GALAXY_VIRTUAL_ENV" = "None" ] && GALAXY_VIRTUAL_ENV="$_GALAXY_VIRTUAL_ENV"; _galaxy_setup_environment True
 """
+PREPARE_DIRS = """mkdir -p working outputs
+if [ -f .job_started ]; then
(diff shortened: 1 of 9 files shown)


lib/galaxy/jobs/command_factory.py
```diff
@@ -19,6 +19,13 @@
 SETUP_GALAXY_FOR_METADATA = """
 [ "$GALAXY_VIRTUAL_ENV" = "None" ] && GALAXY_VIRTUAL_ENV="$_GALAXY_VIRTUAL_ENV"; _galaxy_setup_environment True
 """
+PREPARE_DIRS = """mkdir -p working outputs
+if [ -f .job_started ]; then
+    rm -rf working/ outputs/; cp -R _working working; cp -R _outputs outputs
+else
+    cp -R working _working; cp -R outputs _outputs
```
jmchilton, line 26: Add a random suffix to prevent collision?

natefoo: Something time-derived like we do for the tmp (and home?) dirs would be preferable than random.

mvdbeek: what would be the scenario for a collision ?

jmchilton: Retried twice? I guess... those are cps and mvs though - maybe I didn't look at this close enough before responding...

mvdbeek: The idea is that we'd always remove `working/` and `outputs/` and then copy their "backups" when the job previously created `.job_started`, so I think it should be safe ?

natefoo:
While this will usually be nothing, it has the potential to be very slow, especially for interrupted tools that have written GB- to TB- of intermediary/scratch files.

mvdbeek:
> While this will usually be nothing, it has the potential to be very slow, especially for interrupted tools that have written GB- to TB- of intermediary/scratch files.

Why is that ? It'll be an `rm -rf` just like it was before, or am I missing something ?

natefoo:
I see, it's like a backup of the dirs prior to any execution occurring.

mvdbeek:
> I see, it's like a backup of the dirs prior to any execution occurring.

yup, that's the idea!

lib/galaxy/jobs/command_factory.py
```diff
@@ -19,6 +19,13 @@
 SETUP_GALAXY_FOR_METADATA = """
 [ "$GALAXY_VIRTUAL_ENV" = "None" ] && GALAXY_VIRTUAL_ENV="$_GALAXY_VIRTUAL_ENV"; _galaxy_setup_environment True
 """
+PREPARE_DIRS = """mkdir -p working outputs
+if [ -f .job_started ]; then
```
mvdbeek: Actually this is redundant, I could just do
```suggestion
if [ -d _working]; then
```

and skip the `touch` below.

PR #9246, opened 2020-01-21
[19.09] python3 fixes for setups using drmaa and realuser setups

bernt-matthias:
- Popen returns bytes in python3
- strings are needed

analoguous to what is happening in the slurm runner I use unicodify here

... seeing all the Popen code duplication in the Galaxy sources it might be a good idea to have a util function.

### DIFF
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -2092,6 +2092,7 @@ class JobWrapper(HasResourceParameters):
             log.debug('(%s) Changing ownership of working directory with: %s' % (job.id, ' '.join(cmd)))
             p = subprocess.Popen(cmd, shell=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
             stdout, stderr = p.communicate()
+            stdout, stderr = unicodify(stdout).strip(), unicodify(stdout).strip()
             if p.returncode != 0:
                 log.error('external script failed.')
                 log.error('stdout was: %s' % stdout)
--- a/lib/galaxy/jobs/runners/drmaa.py
+++ b/lib/galaxy/jobs/runners/drmaa.py
@@ -18,7 +18,7 @@ from galaxy.jobs.runners import (
     AsynchronousJobRunner,
     AsynchronousJobState
 )
-from galaxy.util import asbool
+from galaxy.util import asbool, unicodify
 
 drmaa = None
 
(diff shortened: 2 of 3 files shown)


lib/galaxy/jobs/runners/drmaa.py
```diff
@@ -18,7 +18,7 @@
     AsynchronousJobRunner,
     AsynchronousJobState
 )
-from galaxy.util import asbool
+from galaxy.util import asbool, unicodify
```
mvdbeek:
```suggestion
from galaxy.util import (
    asbool,
    unicodify,
)
```

mvdbeek (commented):
Yeah, unifying this pattern would be good (in a PR to dev!). There's already https://github.com/galaxyproject/galaxy/blob/9e76dcd9d381545ef627b63ec406dccf34bc9ae9/lib/galaxy/tool_util/deps/commands.py#L97 which is nice and can probably be used in many places (though `_wait` needs to be unicodified as well).

mvdbeek (approved):

lib/galaxy/jobs/__init__.py
```diff
@@ -2092,6 +2092,7 @@ def _change_ownership(self, username, gid):
             log.debug('(%s) Changing ownership of working directory with: %s' % (job.id, ' '.join(cmd)))
             p = subprocess.Popen(cmd, shell=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
             stdout, stderr = p.communicate()
+            stdout, stderr = unicodify(stdout).strip(), unicodify(stdout).strip()
```
nsoranzo, line 2095: Sorry for being late on this, but this should have been:

```python
            stdout, stderr = unicodify(stdout).strip(), unicodify(stderr).strip()
```

lib/galaxy/jobs/runners/univa.py
```diff
@@ -30,8 +30,8 @@
 import subprocess
 import time
 
-from galaxy import util
 from galaxy.jobs.runners.drmaa import DRMAAJobRunner
+from galaxy.util import size_to_bytes, unicodify
```
nsoranzo, line 34:
```python
from galaxy.util import (
    size_to_bytes,
    unicodify,
)
```

lib/galaxy/jobs/__init__.py
mvdbeek: That's also a textbook example of why we shouldn't copy-paste our code around 😆

bernt-matthias: Indeed :) Will open a PR with the necessay changes including the other suggestions of @nsoranzo

PR #9248, opened 2020-01-21
[WIP] Update cellxgene to enable setting options and saving created user annotations.

blankenberg:
Todo
- Should ensure a good way to terminate the running cellxgene server from inside
- Requires an updated docker container to be made available


blankenberg:
potential solution for shutdown of cellxgene: https://github.com/chanzuckerberg/cellxgene/pull/1127

PR #9249, opened 2020-01-22
Replace right panel in Workflow Editor with Vue component

guerler:
This PR replaces the right panel in the workflow editor with a Vue component and redirects update request through the workflow api. It removes legacy code mako and controller code. Additionally the PR removes the redundant WorkflowTags component in the WorkflowList and replace it with the shared common Tags helper component. The Tag component has been refactored to avoid prop changes within the subcomponent.


guerler:
@galaxybot test this

PR #9253, opened 2020-01-22
[19.09] fix bug introduced in https://github.com/galaxyproject/galaxy/pull/9246

bernt-matthias:
sorry for the copy paste error


lib/galaxy/jobs/runners/univa.py
```diff
@@ -31,7 +31,11 @@
 import time
 
 from galaxy.jobs.runners.drmaa import DRMAAJobRunner
-from galaxy.util import size_to_bytes, unicodify
+from galaxy.util import (
+    size_to_bytes,
```
mvdbeek:
```suggestion
    size_to_bytes,
```

PR #9258, opened 2020-01-23
Make `str.join()`s more efficient

nsoranzo:
by using generator expressions and tuples instead of lists.

### DIFF
--- a/lib/galaxy/visualization/plugins/plugin.py
+++ b/lib/galaxy/visualization/plugins/plugin.py
@@ -71,10 +71,10 @@ class VisualizationPlugin(ServesTemplatesPluginMixin):
         self.name = name
         self.config = config
         base_url = context.get('base_url', '')
-        self.base_url = '/'.join([base_url, self.name]) if base_url else self.name
+        self.base_url = '/'.join((base_url, self.name)) if base_url else self.name
         self.static_path = self._get_static_path(self.path)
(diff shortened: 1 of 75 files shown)


lib/galaxy/visualization/plugins/plugin.py
```diff
@@ -71,10 +71,10 @@ def __init__(self, app, path, name, config, context=None, **kwargs):
         self.name = name
         self.config = config
         base_url = context.get('base_url', '')
-        self.base_url = '/'.join([base_url, self.name]) if base_url else self.name
+        self.base_url = '/'.join((base_url, self.name)) if base_url else self.name
```
bgruening, line 74: Maybe also remove the join in favor of a pure string concat, as you did below?

bgruening (approved):
Thanks, @nsoranzo! Just one small comment.

lib/galaxy/visualization/plugins/plugin.py
nsoranzo: In general I think the join is clearer when there are multiple variables to join using a separator, like in this case.
In the case below there was only one variable really, so it seemed better to just append to it using `+`.
Hopefully that makes sense?

bgruening: :+1:

PR #9260, opened 2020-01-23
[20.01] Fix unselected select parameter comparison to empty string

mvdbeek:
`$wrapper == ""` and `$wrapper == "None"` had been valid ways
to check for unselected / optional select parameters.
This restores the `$wrapper == ""` form as used by gffcompare
https://github.com/galaxyproject/tools-iuc/blob/master/tools/gffcompare/gffcompare.xml#L41

### DIFF
--- a/lib/galaxy/tools/wrappers.py
+++ b/lib/galaxy/tools/wrappers.py
@@ -79,6 +79,8 @@ class InputValueWrapper(ToolParameterValueWrapper):
         # For backward compatibility, allow `$wrapper != ""` for optional non-text param
         if self.input.optional and self.value is None:
             if isinstance(other, string_types):
+                if other == '':
+                    return ''
                 return str(self)
             else:
                 return None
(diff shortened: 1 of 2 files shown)


bgruening (approved):

lib/galaxy/tools/wrappers.py
```diff
@@ -79,6 +79,8 @@ def _get_cast_value(self, other):
         # For backward compatibility, allow `$wrapper != ""` for optional non-text param
         if self.input.optional and self.value is None:
             if isinstance(other, string_types):
+                if other == '':
+                    return ''
```
nsoranzo, line 83: This change doesn't seem necessary for the mentioned optional `select` use case, or am I missing something?

nsoranzo (commented):
Thanks! I'll backport this to 19.09.

lib/galaxy/tools/wrappers.py
mvdbeek: I think this is more correct, otherwise $wrapper == "" is not going to work (the inverse of the backwards compatibility example on line 80)

nsoranzo: That also works because `str(self)` returns `""` when `self.value` is None, due to `to_param_dict_string()` implementation.

Issue #9275, opened 2020-01-29
Email notification upon job completion is sent once per mapped over job


mvdbeek:
So for a collection with a 1000 items you'll receive 1000 emails.

Other points:
 - there's only one email address those mails come from, so for usegalaxy.org you'll receive an email from activate@galaxyprojecy.org
 - the subject line is `Galaxy workflow step notification 'Test'`

davelopez:
This one is still biting and causing account activation emails to be lost because of external rate limiting when a user does this map over with completion emails. We should revisit it.

mvdbeek:
💯. We could remove that feature for workflows, we have per-invocation completion now.

PR #9277, opened 2020-01-29
[20.01] Run email post job action only once per implicit collection jobs

mvdbeek:
Fixes the remainder of https://github.com/galaxyproject/galaxy/issues/9275

### DIFF
--- a/lib/galaxy/jobs/actions/post.py
+++ b/lib/galaxy/jobs/actions/post.py
@@ -520,7 +520,7 @@ class ActionBox(object):
                 implicit_collection_jobs_association = job.implicit_collection_jobs_association
                 if implicit_collection_jobs_association:
                     summary = summarize_jobs_to_dict(sa_session, implicit_collection_jobs_association.implicit_collection_jobs)
-                    # PJA get's executed prior to changing the job state,
+                    # PJA gets executed prior to changing the job state,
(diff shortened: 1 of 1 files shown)


lib/galaxy/jobs/actions/post.py
```diff
@@ -511,5 +514,14 @@ def execute_on_mapped_over(cls, trans, sa_session, pja, step_inputs, step_output
 
     @classmethod
     def execute(cls, app, sa_session, pja, job, replacement_dict=None):
+        execute_action = True
         if pja.action_type in ActionBox.actions:
+            if pja.action_type in ActionBox.run_once:
+                implicit_collection_jobs_association = job.implicit_collection_jobs_association
+                if implicit_collection_jobs_association:
+                    summary = summarize_jobs_to_dict(sa_session, implicit_collection_jobs_association.implicit_collection_jobs)
+                    # PJA get's executed prior to changing the job state,
```
dannon:
```suggestion
                    # PJA gets executed prior to changing the job state,
```

mvdbeek:
@jmchilton mentioned that this could have a pretty bad performance penalty (each job checking each jobs status for implicit collections). So maybe we shouldn't do this for now / until we have a more performant way to do this.

PR #9278, opened 2020-01-29
[20.01] Backwards compatible build number parsing

mvdbeek:
- Restore parsing build numbers separated with `--`. I changed this in #9268
but the alternative `--` is also in use at https://github.com/BioContainers/multi-package-containers/blob/master/combinations/hash.tsv#L14

- Don't include build number 0 for single images and have conda build take precedence over image_build
for single target images. If overriding the build suffix is really necessary one can use the `name_override` column.

With those 2 commits I can get https://github.com/BioContainers/multi-package-containers/pull/966/files to pass without building any new images


jmchilton:
https://pypi.org/project/galaxy-tool-util/20.1.0.dev5/

PR #9279, opened 2020-01-29
extend config file inputs tests

bernt-matthias:
Shows that optional selects get `"None"` in the json file. I guess this needs to be changed to `None` -- i.e. its a bug. 

Any hint where the code is that generates this?

For files the result is analogous to what we deal with in cheetah. But I would `[]` still more intuitive than `[None]` for optional multiple selects. May be we can change this? But I'm fine with the state of the art .. since its at least consistent.


bernt-matthias:
Sorry .. I overwrote some recent changes because it mis-`git`ed things again :( will change this tomorrow.

lib/galaxy/tools/parameters/wrapped_json.py
```diff
@@ -79,20 +79,25 @@ def _data_input_to_path(v):
...
+        json_value = _cast_if_not_none(value_wrapper, float, empty_to_none=True)
     elif input_type == "integer":
-        json_value = _cast_if_not_none(value, int, empty_to_none=True)
+        json_value = _cast_if_not_none(value_wrapper, int, empty_to_none=True)
     elif input_type == "boolean":
-        json_value = _cast_if_not_none(value, bool)
+        json_value = _cast_if_not_none(value_wrapper, bool)
+    elif input_type == "select":
+        if input.multiple:
+            json_value = [_ for _ in _cast_if_not_none(value_wrapper.value, list)]
```
jmchilton, line 92: If you can get the tests working, this change definitely makes sense to me!

bernt-matthias: Guess this is possible. 

I just would like to use the opportunity to remark that this PR (in addition to fix the "None" problem for selects) also changes the following:

- multiple selects are now lists (before comma separated string)

Also I'm not sure if data_column should be cast to int, because of `<param type="data_column" 
use_header_names="true".../>` (but I could not get this attribute working as expected anyway).

mvdbeek (approved):
Looks good to me, are you happy with this @bernt-matthias ?

bernt-matthias:
Yes.

PR #9281, opened 2020-01-29
[20.01] Tag display fixes for ui-select.

### LINKED ISSUE #9191: Dataset tags: "name:tag" displayed in workflow+tool form input select lists. "#tag" would be better/more consistent
Might be ticketed or adjusted for an upcoming release already .. if so, please close this out as dup with link to other. It definitely is not new (has been discussed before) but can't find a ticket or PR for it. ping @guerler @dannon 

__Punch-line__: Could we display tags in the workflow/tool select menus the same way are displayed in other views? For UX consistency? eg: `#tag` & `tag`. 

**Good-1:** A dataset in a history panel - collapsed. Example dataset has two tags, `#original` and `original`. The display is as expected (hash-prefixed tags shown + no-hash-prefixed tags hidden).

dataset-with-two-tag-types-collapsed

**Good-2:** That same dataset in a history panel - expanded with "edit tags" toggled to show. This display is informative & easily interpreted by end-users. Tags are displayed exactly as entered:

dataset-with-two-tag-types-expanded

**Good-3:** That same dataset in the "Saved Datasets" view (User > Datasets). This display is also informative & easily interpreted.

dataset-with-two-tag-types-in-saved-datasets-view

**Odd-1:** That same dataset in a workflow's dataset select list. The display is informative _yet inconsistent with other views_. Notice the difference between how `#original` versus `original` is shown. Each "hash-prefixed" tag has extra `name:` added in (cryptic/exposed code artifact?). Can see in screenshot how cluttered this type of tag display formatting can become if 2+ hash-prefixed tags are assigned to datasets:

dataset-with-two-tag-types-in-wf-input-select-list

**Odd-2:** That same dataset in a tool form's dataset select list. Same display as workflow select, eg: `name:tag` instead of `#tag`.

dataset-with-two-tag-types-in-tool-input-select-list

dannon:
Fix tag display coloring calculation in `ui-select` component to match the history.  This code was previously basing the color key on the truncated value, not the full string as elsewhere does.
Also formats nametags/hashtags as appropriate, instead of leaving `name:` in the string.
minor: refactored to drop underscore use here.
fixes #9191

### DIFF
--- a/client/galaxy/scripts/mvc/ui/ui-select-default.js
+++ b/client/galaxy/scripts/mvc/ui/ui-select-default.js
@@ -211,18 +211,17 @@ var View = Backbone.View.extend({
                     return `
                     ${_.escape(result.text)}
                     <div>
-                        ${_.reduce(
-                            filteredTags.slice(0, 5),
-                            (memo, tag) => {
-                                const tagColors = keyedColorScheme(tag.slice(5));
-                                return `${memo}&nbsp;<div style="background-color: ${tagColors.primary}; color: ${
-                                    tagColors.contrasting
-                                }; border: 1px solid ${
-                                    tagColors.darker
-                                }" class="badge badge-primary badge-tags">${_.escape(tag)}</div>`;
-                            },
-                            ""
-                        )}
+                        ${filteredTags.slice(0, 5).reduce((memo, tag) => {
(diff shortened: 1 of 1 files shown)


mvdbeek (approved):

mvdbeek:
Thanks @dannon!

PR #9289, opened 2020-01-31
MS Edge Fixes

### LINKED ISSUE #9250: Running 'select first' lines with internet explorer overwrites sent job parameters
We experienced this in a training, users ran `Show beginning1` and `bgruening/text_processing/tp_head_tool/1.1.0`, and in all cases wrote `5` in the number of lines they wanted, and upon running found 10 lines, both in the information and the result file.

It seems to be internet explorer related 🙃🙃🙃

dannon:
Fixes issues with:
- bootstrap-vue needs the transpilation pass to work in edge
- edge can't handle onchange and oninput (if input manipulates the value) at the same time and just silently never fires change if you try.

fixes https://github.com/galaxyproject/galaxy/issues/9250

### DIFF
--- a/client/galaxy/scripts/mvc/ui/ui-slider.js
+++ b/client/galaxy/scripts/mvc/ui/ui-slider.js
@@ -24,7 +24,7 @@ const View = Backbone.View.extend({
         this.$text = this.$(".ui-form-slider-text");
         this.$slider = this.$(".ui-form-slider-element");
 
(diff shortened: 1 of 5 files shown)


dannon:
Looks like this doesn't play nicely with rewire for the unit tests (but it does work in real usage...) so I've got to fix that still.

dannon:
Ok, this should hopefully pass the tests now, and will fix the IE11/Edge problems folks reported.

PR #9291, opened 2020-01-31
Do not strip tab characters from data_column reference and clarify documentation for use_header_names

bernt-matthias:
- add a test (also as example)
- clarify in the docs
- bugfix empty columns must not be stripped

### DIFF
--- a/lib/galaxy/tool_util/xsd/galaxy.xsd
+++ b/lib/galaxy/tool_util/xsd/galaxy.xsd
@@ -2563,8 +2563,10 @@ Set to ``false`` to not force user to select an option in the list.</xs:document
             <xs:documentation xml:lang="en">Used only if the ``type`` attribute
 value is ``data_column``, if ``true`` Galaxy assumes first row of ``data_ref``
 is a header and builds the select list with these values rather than the more
-generic ``c1`` ... ``cN``.
-</xs:documentation>
+generic ``c1`` ... ``cN`` (i.e. it will be ``c1: head1`` ... ``cN: headN``).
+Note that the content of the Cheetah variable is still
+the column index.
+            </xs:documentation>
           </xs:annotation>
         </xs:attribute>
         <xs:attribute name="display" type="DisplayType">
@@ -4830,7 +4832,7 @@ write out a JSON representation of the tool parameters.
 
 *Example*
 
-The following will create a cheetah variable that can be evaluated as ``$inputs`` that
+The following will create a Cheetah variable that can be evaluated as ``$inputs`` that
 will contain the tool parameter inputs.
 
 ```xml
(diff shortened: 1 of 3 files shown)


lib/galaxy/tools/parameters/basic.py
```diff
@@ -1271,7 +1271,7 @@ def get_options(self, trans, other_values):
             try:
                 with open(dataset.get_file_name(), 'r') as f:
                     head = f.readline()
-                cnames = head.rstrip().split('\t')
+                cnames = head.rstrip("\n\r ").split('\t')
```
bgruening, line 1274: I would assume this as the default. The documentation says "whitespaces", is \r not included?

bernt-matthias: Default is that `\n\r\t `. The problem is that this removes the first or last columns if they are empty.

test/functional/tools/column_param.xml
```diff
@@ -1,10 +1,13 @@
 <tool id="column_param" name="Column Param">
-  <command>
-    cut -f '$col' '$input1' > '$output1'
-  </command>
+  <command><![CDATA[
+    echo "col $col" >> '$output1' &&
```
jmchilton: This is odd to add to that file right - can you echo these to a new second output?

bernt-matthias: Good idea.

jmchilton (approved):

jmchilton:
Looks awesome - can you merge this once the framework tests pass? Thanks so much for the enhancements.

bernt-matthias:
Hmm. framework tests timed out :(

test/functional/tools/column_param.xml
```diff
@@ -1,23 +1,34 @@
 <tool id="column_param" name="Column Param">
-  <command>
-    cut -f '$col' '$input1' > '$output1'
-  </command>
+  <command><![CDATA[
+    cut -f '$col' '$input1' >> '$output1'
+    echo "col $col" >> '$output2' &&
+    echo "col_names $col_names" >> '$output2' &&
```
nsoranzo:
```suggestion
cut -f '$col' '$input1' > '$output1' &&
echo "col $col" > '$output2' &&
echo "col_names $col_names" >> '$output2'
```

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -2563,8 +2563,10 @@ Set to ``false`` to not force user to select an option in the list.</xs:document
             <xs:documentation xml:lang="en">Used only if the ``type`` attribute
 value is ``data_column``, if ``true`` Galaxy assumes first row of ``data_ref``
```
```suggestion
value is ``data_column``. If ``true``, Galaxy assumes the first row of ``data_ref``
```

test/functional/tools/column_param.xml
```diff
@@ -1,23 +1,34 @@
...
+echo "col_names $col_names" >> '$output2'
+  ]]></command>
   <inputs>
     <param type="data" format="tabular" name="input1" label="Input 1" />
     <param name="col" type="data_column" data_ref="input1" label="Column to Use" />
+    <param name="col_names" type="data_column" data_ref="input1" use_header_names="true" label="Column to Use" />
   </inputs>
   <outputs>
     <data name="output1" format="tabular" />
+    <data name="output2" format="tabular" />
```
nsoranzo:
```suggestion
    <data name="output2" format="txt" />
```

nsoranzo (approved):

PR #9294, opened 2020-01-31
[20.01] Prevent caching of bootstrapped client data

dannon:
Attempting to address the last bits of https://github.com/galaxyproject/galaxy/issues/7249

The javascript itself can/will still be cached, and that's the brunt of the data transmitted when filling out the app.  The majority of this particular response is the bootstrapped data related primarily to current user that we don't really want to cache.

### DIFF
--- a/lib/galaxy/webapps/base/controller.py
+++ b/lib/galaxy/webapps/base/controller.py
@@ -276,6 +276,10 @@ class JSAppLauncher(BaseUIController):
         return self._bootstrapped_client(trans, **kwd)
 
     def _bootstrapped_client(self, trans, app_name='analysis', **kwd):
+        # This includes contextualized user options in the bootstrapped data; we don't want to cache it.
+        trans.response.headers['Cache-Control'] = ['no-cache', 'no-store', 'must-revalidate']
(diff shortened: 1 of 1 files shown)


lib/galaxy/webapps/base/controller.py
```diff
@@ -276,6 +276,10 @@ def client(self, trans, **kwd):
         return self._bootstrapped_client(trans, **kwd)
 
     def _bootstrapped_client(self, trans, app_name='analysis', **kwd):
+        # This includes contextualized user options in the bootstrapped data; we don't want to cache it.
+        trans.response.headers['Cache-Control'] = ['no-cache', 'no-store', 'must-revalidate']
```
mvdbeek, line 280: Might make sense to consolidate this in a decorator (in dev) ? We do this in a couple of places.

mvdbeek (approved):
Looks good to me, thanks @dannon !

lib/galaxy/webapps/base/controller.py
dannon: Yep, I actually started doing exactly that yesterday, but for a separate PR for dev

nsoranzo:
Merged forward.

PR #9297, opened 2020-02-03
[19.09] Add build_sites_config_file option to config_schema

mvdbeek:
We need this for the default value. Without the defaults
many external display applications won't load.

### DIFF
--- a/doc/source/admin/galaxy_options.rst
+++ b/doc/source/admin/galaxy_options.rst
@@ -732,6 +732,17 @@
 :Type: str
 
 
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+``build_sites_config_file``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+:Description:
+    File that defines the builds (dbkeys) available at sites used by
+    display applications and the URL to those sites.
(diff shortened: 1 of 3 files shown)


jmchilton (approved):

dannon (approved):

jdavcs:
I would suggest using `path_resolves_to: config_dir` instead of hardcoding `config` into the path. We have that mechanism and it works for `data_dir`. Otherwise, changing the value of `config_dir` would invalidate this path. All the other `config` instances in the schema (except those where `config/` is relative to root) should be adjusted accordingly (pending a bit of refactoring in `config/__init__.py`, which I really should get to ..yesterday).

mvdbeek:
> I would suggest using `path_resolves_to: config_dir` instead of hardcoding `config` into the path

That's not in 19.09, is it ?

jdavcs:
Oh, this is for 19.09? Should've checked the target branch name! Sorry, then yes, of course this is fine.

jdavcs (approved):

nsoranzo:
Maybe merge forward via PR to address @ic4f's point in `release_20.01` and following?

mvdbeek:
Yep, on it.

PR #9305, opened 2020-02-04
Fixes reports app under py3 

dannon:
…..by swapping to the more standard (modern) log configuration function (which the toolshed already uses)


nsoranzo:
Thanks @dannon ! Should this be backported to 20.01 ?

dannon:
@nsoranzo Probably, yeah -- will do.

PR #9306, opened 2020-02-04
Clean up old 'blue' style path, update documentation and samples

dannon:
to refer to plain old 'static' now.

Includes a config rebuild (necessary for the changes here) that picked up a few extra bits that look like they got left out of a previous PR?

### DIFF
--- a/doc/source/admin/nginx.md
+++ b/doc/source/admin/nginx.md
@@ -117,7 +117,7 @@ http {
         server_name _;
 
         # use a variable for convenience
-		set $galaxy_root /srv/galaxy/server;
+        set $galaxy_root /srv/galaxy/server;
 
         # Enable HSTS
         add_header Strict-Transport-Security "max-age=15552000; includeSubdomains";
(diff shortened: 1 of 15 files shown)


doc/source/admin/nginx.md
```diff
@@ -130,10 +130,6 @@ http {
         }
 
 		# serve framework static content
-		location /static/style {
-			alias $galaxy_root/static/style/blue;
-			expires 24h;
-		}
 		location /static {
```
nsoranzo: Not connected to this PR, but I noticed that the code here is mixing spaces and tabs: convert all tabs to spaces?

dannon: @nsoranzo Sure, I'll take a look.

dannon: (done)

doc/source/admin/nginx.md
```diff
@@ -203,8 +199,8 @@ previous section:
             }
 
             # serve framework static content
-            location /galaxy/static/style {
-                alias $galaxy_root/static/style/blue;
+            location /static {
```
nsoranzo: Shouldn't this be
```suggestion
            location /galaxy/static {
```

dannon: Yes it should, good catch.

nsoranzo:
Last one: should also https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy_test/driver/driver_util.py#L462 be updated?

dannon:
@nsoranzo Looks like it; that join foiled my grepping!  I'll make another quick check and fix that up.

nsoranzo (approved):

natefoo (approved):
Thanks @dannon!

PR #9307, opened 2020-02-04
Separate automated schema loading from config options

jdavcs:
Update: this is ready for review.

Delegate loading+processing configuration from schema to separate module. 

`config/__init__.py` is doing too much. This is an attempt to separate 2 different responsibilities: (1) automatically loading and processing all config properties from the schema; and (2) setting individual config properties and handling special cases. The former is delegated to a `ConfigurationLoader` module. The idea is to use this loader module for any further automated schema processing, whereas `GalaxyAppConfiguration` is to be used for processing individual config properties (as it did before).

Last year's modifications to the config system, while simplifying configuration loading, have done the opposite to the configuration module's code. This modification is an attempt to address this problem.

(Edit: more detailed rationale in commit messages)

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -26,7 +26,7 @@ from beaker.util import parse_cache_config_options
 from six import string_types
 from six.moves import configparser
 
-from galaxy.config.schema import AppSchema
+from galaxy.config.loader import ConfigurationLoader
 from galaxy.containers import parse_containers_config
 from galaxy.exceptions import ConfigurationError
 from galaxy.model import mapping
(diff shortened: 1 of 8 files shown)


bgruening (approved):
Great refactoring @ic4f!

jdavcs:
Thank you for reviewing, @bgruening!

test/unit/config/test_load_config.py
```diff
@@ -29,71 +29,71 @@ def mock_process_config(self, kwargs):
...
-    assert type(config._raw_config['property3']) is float
-    assert type(config._raw_config['property4']) is bool
+    assert type(config.raw_config['property1']) is str
+    assert type(config.raw_config['property2']) is int
+    assert type(config.raw_config['property3']) is float
+    assert type(config.raw_config['property4']) is bool
 
 
-def test_update_raw_config_from_kwargs(mock_init):
+def test_updateraw_config_from_kwargs(mock_init):
```
nsoranzo: I'd revert this change.

test/unit/config/test_load_config.py
```diff
@@ -29,71 +29,71 @@ def mock_process_config(self, kwargs):
...
-    assert type(config._raw_config['property3']) is float
-    assert type(config._raw_config['property4']) is bool
+    assert type(config.raw_config['property1']) is str
+    assert type(config.raw_config['property2']) is int
+    assert type(config.raw_config['property3']) is float
+    assert type(config.raw_config['property4']) is bool
 
 
-def test_update_raw_config_from_string_kwargs(mock_init):
+def test_updateraw_config_from_string_kwargs(mock_init):
```
nsoranzo: Revert also this one.

test/unit/config/test_load_config.py
```diff
@@ -29,71 +29,71 @@ def mock_process_config(self, kwargs):
...
-    assert type(config._raw_config['property3']) is float
-    assert type(config._raw_config['property4']) is bool
+    assert type(config.raw_config['property1']) is str
+    assert type(config.raw_config['property2']) is int
+    assert type(config.raw_config['property3']) is float
+    assert type(config.raw_config['property4']) is bool
 
 
-def test_update_raw_config_from_kwargs_with_none(mock_init):
+def test_updateraw_config_from_kwargs_with_none(mock_init):
```
nsoranzo: Revert.

test/unit/config/test_load_config.py
```diff
@@ -29,71 +29,71 @@ def mock_process_config(self, kwargs):
...
+    assert config.raw_config['property1'] is None
+    assert config.raw_config['property2'] is None
+    assert config.raw_config['property3'] is None
+    assert config.raw_config['property4'] is None
+    assert config.raw_config['property5'] is None
+    assert config.raw_config['property6'] is None
 
 
-def test_update_raw_config_from_kwargs_falsy_not_none(mock_init):
+def test_updateraw_config_from_kwargs_falsy_not_none(mock_init):
```
nsoranzo: Revert.

test/unit/config/test_load_config.py
jdavcs: Of course! Thank you for catching this. Stupid sed (all of these are the result of `:%s/_raw_config/raw_config/g`. Such a careless mistake... Yikes.

test/unit/config/test_load_config.py
jdavcs: The 4 comments above: done. Marking as resolved.

lib/galaxy/config/loader.py
```diff
@@ -0,0 +1,128 @@
+import logging
+import os
+
+from galaxy.exceptions import ConfigurationError
+from galaxy.util import string_as_bool
+
+GALAXY_APP_NAME = 'galaxy'
+GALAXY_CONFIG_SCHEMA_PATH = 'lib/galaxy/webapps/galaxy/config_schema.yml'
```
nsoranzo: The 2 constants above seems not used in this file, remove?

lib/galaxy/config/loader.py
```diff
@@ -0,0 +1,128 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class ConfigurationLoader():
```
nsoranzo:
```suggestion
class ConfigurationLoader(object):
```

lib/galaxy/config/loader.py
jdavcs: Oh yes! (they were used in loader in the first iteration)

lib/galaxy/config/__init__.py
```diff
@@ -209,135 +210,29 @@ def _parse_config_file_options(self, defaults, listify_defaults, config_kwargs):
...
-            setattr(self, key, path)  # update property
-            _cache[key] = path  # cache it!
-            return path
-
-        _cache = {}
-        for key in self._paths_to_resolve:
-            resolve(key)
+    def _load_config(self, kwargs):
+        loader = ConfigurationLoader()
+        loader.load(self, kwargs)
```
nsoranzo, line 233: Here `loader.load()` has the side effect of modifying values in `self`, which makes me wonder if `ConfigurationLoader` is the right abstraction here.
I don't have an alternative solution though.

jdavcs: That side effect is by design: creating schema-loaded attributes on the config object is one of the loader's primary responsibilities. You have a point, of course: by doing so the loader manipulates the internal structure of the config object. Of course, these two end up being very tightly coupled.

I'm not sure whether this is a good enough solution; it might be. Or maybe, alternatively, the loader could prepare a dict and supply it to the config, and then the config would set its own attributes based on that data (and that data would be the current `raw_config`). That might be a better approach. I'll give this some more thought tomorrow.

nsoranzo: If they are so tightly coupled there is no point in separating some methods in a different class, IMO. This kind of side effects need to be very well documented, and will still surprise other developers.

The alternative approach could work, I guess. Maybe @jmchilton has suggestions? I tend to refer to him as expert in class/module separation :bowing_man:

jdavcs: @nsoranzo you are right - it is not the right abstraction. We can do better. I'll close this in favor of a follow-up. Thank you for the detailed review!

jdavcs:
I'm closing this in favor of #9371 (currently WIP)

PR #9310, opened 2020-02-04
Allow setting container path for docker_volumes

AndreasSko:
I'm working on getting InteractiveTools to run on my docker-compose setup on Mac. The problem I faced was that I had to mount GALAXY_ROOT from a different path at the host, while keeping it consistent inside the container. Trying `<param id="docker_volumes">/Users/../export/galaxy:/galaxy:rw</param>` resulted in the error: `Unparsable volumes string in configuration`, which to me seemed not quite right. 
The function `preprocess_volumes` only allowed to set the host path (HOST:MODE) and therefore forced to use it inside the container, which in some cases might be problematic. With this change it is also possible to set the container path (HOST:CONTAINER:MODE), as described in https://docs.docker.com/storage/bind-mounts/

### DIFF
--- a/lib/galaxy/tool_util/deps/container_classes.py
+++ b/lib/galaxy/tool_util/deps/container_classes.py
@@ -103,6 +103,8 @@ def preprocess_volumes(volumes_raw_str, container_type):
     ['/a/b:rw']
     >>> preprocess_volumes("/a/b:ro,/a/b/c:rw", DOCKER_CONTAINER_TYPE)
     ['/a/b:ro', '/a/b/c:rw']
+    >>> preprocess_volumes("/a/b:/a:ro,/a/b/c:/a/b:rw", DOCKER_CONTAINER_TYPE)
+    ['/a/b:/a:ro', '/a/b/c:/a/b:rw']
     >>> preprocess_volumes("/a/b:default_ro,/a/b/c:rw", DOCKER_CONTAINER_TYPE)
     ['/a/b:ro', '/a/b/c:rw']
     >>> preprocess_volumes("/a/b:default_ro,/a/b/c:rw", SINGULARITY_CONTAINER_TYPE)
@@ -115,8 +117,12 @@ def preprocess_volumes(volumes_raw_str, container_type):
 
     for volume_raw_str in volumes_raw_strs:
         volume_parts = volume_raw_str.split(":")
-        if len(volume_parts) > 2:
+        if len(volume_parts) > 3:
             raise Exception("Unparsable volumes string in configuration [%s]" % volumes_raw_str)
+        if len(volume_parts) == 3:
+            volume_parts = ["%s:%s" % (volume_parts[0], volume_parts[1]), volume_parts[2]]
+        if len(volume_parts) == 2 and volume_parts[1] not in ("rw", "ro", "default_ro"):
+            volume_parts = ["%s:%s" % (volume_parts[0], volume_parts[1]), "rw"]
         if len(volume_parts) == 1:
             volume_parts.append("rw")
         volumes.append(volume_parts)


jmchilton:
Galaxy repeatedly makes assumptions that the job path on the Galaxy host matches that of the container it is run inside. I strongly suggest using Pulsar for jobs if this can't be managed for your setup - but I guess that won't work with ITs right now. 

Are ITs working end-to-end with this change?

I know I have at times needed to add I think it was ``/private`` to Docker for Mac options in order to get things to work with consistent paths across host and container - is that possible for you? (I'm not using compose so my setup could be very different admittedly.)

Screen Shot 2020-02-04 at 9 48 44 AM

AndreasSko:
The goal of my setup (https://github.com/AndreasSko/docker-galaxy-stable/tree/interactive-tools/compose-v2) is to do as few initial steps as possible, so I'm mounting all the Galaxy files at the working-dir (EXPORT_DIR=./export), as this should work everywhere. For the containers though the path to Galaxy is always `/galaxy`, so no problem with that. 
Now the InteractiveTools are started inside the Galaxy container, but to my understanding Docker always mounts volumes based on the host (so mounting /galaxy from inside the Galaxy container won't work, as there is no host-side /galaxy). So to keep the assumption of the path being the same everywhere, I would need the option to specify the correct container path.

But as specifying this is just optional, this shouldn't change the behavior of existing setups, right?

bgruening:
> Are ITs working end-to-end with this change?

@AndreasSko it is working with those changes right?

AndreasSko:
Yeah, the IT containers are able to run with this change now. Still fighting a bit with the networking and routing part, but I guess thats another story 😄

bgruening:
@jmchilton do you still think this causes more harm? It seems it makes OSX testing and developing easier. Thanks.

mvdbeek:
The problem is that this will probably only work if you mount the entire galaxy root as one big volume. It would be better if deployers don't do that, and if we need to break this in the future, downstream projects and deployments will need to adapt to this ... if it is going to be used it is technical debt IMO.

Isn't the right way (apart from using pulsar) to do this to use a docker volume for the persistent data ?

AndreasSko:
@mvdbeek I don't see why that change would only allow to mount galaxy-root inside the container, as the only difference that would bring is to allow to set a different path (so from host path `/User/../export/galaxy` to container path `/galaxy`, for example). You can still define "sub-mounts" (/galaxy:ro, /galaxy/database:rw etc.), which worked at least in my testing. So my base problem is pretty much described here: https://stackoverflow.com/a/31381323

Docker volumes are not that easily reachable from the host (in that you first need to find out where it actually is stored) compared to a simple ./export folder. Also I'm not even sure if volumes would actually solve this problem.

mvdbeek:
> You can still define "sub-mounts" (/galaxy:ro, /galaxy/database:rw etc.), which worked at least in my testing.

Right, but if you start moving these around to different places than what Galaxy assumes it is going to break. That path translation should be handled by pulsar IMO.
I have no problem with someone merging this, but I also think you don't want to use this.

> Docker volumes are not that easily reachable from the host (in that you first need to find out where it actually is stored) compared to a simple ./export folder. Also I'm not even sure if volumes would actually solve this problem.

It would take some more effort for Galaxy to understand volumes. That they are not easily reachable from the host is true, but the point of docker volumes is precisely that they are portable and independent of the host.

AndreasSko:
> Right, but if you start moving these around to different places than what Galaxy assumes it is going to break. That path translation should be handled by pulsar IMO.

I'm running Galaxy inside a container and in there the paths are all standard `/galaxy..`, so the path-assumption for Galaxy always hold. This volume part is actually the only time were Galaxy "sees" the underlying host-path (my config currently looks something like this `<param id="docker_volumes">{{ HOST_EXPORT_DIR }}/$galaxy_root:$galaxy_root:ro,...</param>`). As Galaxy normally doesn't need to be aware that it is running inside a container (except for this `docker run` case), I think Pulsar path translation would be a bit overkill. This change actually just brings the volume-syntax of Galaxy in-sync with the Docker one.

bgruening:
I also think the changes are actually allowing more flexibility and bring the Docker syntax closer to the documented upstream one. The path to Galaxy code is not changed during the deployment, relative to Galaxy - so I can't really see how this PR can be harmful :(

AndreasSko:
Is there any news if this PR is mergeable? :)

mvdbeek:
Thanks @AndreasSko, please use responsibly :)

martenson:
@AndreasSko it is, and also merged. Thanks for the contribution!

Issue #9308, opened 2020-02-04
Workflow save-as broken on 20.01


mvdbeek:
If you use the workflow editor save-as function you'll be greeted with

```
Error - <class 'TypeError'>: save_workflow_as() got an unexpected keyword argument 'from_tool_form'
URL: http://127.0.0.1:8080/workflow/save_workflow_as
File '/Users/mvandenb/src/galaxy/lib/galaxy/web/framework/middleware/error.py', line 154 in __call__
  app_iter = self.application(environ, sr_checker)
File '/Users/mvandenb/src/galaxy/.venv/lib/python3.6/site-packages/paste/recursive.py', line 85 in __call__
  return self.application(environ, start_response)
File '/Users/mvandenb/src/galaxy/.venv/lib/python3.6/site-packages/paste/httpexceptions.py', line 640 in __call__
  return self.application(environ, start_response)
File '/Users/mvandenb/src/galaxy/lib/galaxy/web/framework/base.py', line 143 in __call__
  return self.handle_request(environ, start_response)
File '/Users/mvandenb/src/galaxy/lib/galaxy/web/framework/base.py', line 222 in handle_request
  body = method(trans, **kwargs)
File '/Users/mvandenb/src/galaxy/lib/galaxy/web/framework/decorators.py', line 57 in call_and_format
  rval = func(self, trans, *args, **kwargs)
TypeError: save_workflow_as() got an unexpected keyword argument 'from_tool_form'
```

PR #9315, opened 2020-02-04
[20.01] Fix workflow editor save as functionality

mvdbeek:
Fixes https://github.com/galaxyproject/galaxy/issues/9308

### DIFF
--- a/lib/galaxy/webapps/galaxy/controllers/workflow.py
+++ b/lib/galaxy/webapps/galaxy/controllers/workflow.py
@@ -573,7 +573,7 @@ class WorkflowController(BaseUIController, SharableMixin, UsesStoredWorkflowMixi
             return {'id': trans.security.encode_id(stored_workflow.id), 'message': 'Workflow %s has been created.' % workflow_name}
 
     @web.json
-    def save_workflow_as(self, trans, workflow_name, workflow_data, workflow_annotation=""):
+    def save_workflow_as(self, trans, workflow_name, workflow_data, workflow_annotation="", from_tool_form=False):
         """
             Creates a new workflow based on Save As command. It is a new workflow, but
             is created with workflow_data already present.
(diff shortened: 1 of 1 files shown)


lib/galaxy/webapps/galaxy/controllers/workflow.py
```diff
@@ -603,6 +603,7 @@ def save_workflow_as(self, trans, workflow_name, workflow_data, workflow_annotat
                     trans,
                     stored_workflow,
                     workflow_data,
+                    from_tool_form=from_tool_form,
```
guerler, line 606: We should move the default for this attribute further down unless it makes sense to be able to set this flag at controller level. Thanks a lot for the fix!

mvdbeek: This shouldn't be a controller endpoint at all, but I don't think we should change this right now.

guerler: Right it will be an api endpoint in the future. Still, we can define the default for `from_tool_form` further downstream in the call chain i.e. in `update_workflow_from_raw_description` or even further downstream.

mvdbeek: It is, it's just that this function is being passed an unused keyword. This bug can also be fixed by adding `**kwargs` and dropping `from_tool_form` or by changing the caller signature from https://github.com/jmchilton/galaxy/blob/c48be70805990de2d234cb618acdf032a537acda/client/galaxy/scripts/mvc/workflow/workflow-view.js#L692

jmchilton: This fix looks good to me for what it is worth. Arguably the web controller should define the default as ``True`` - but since the API controller shouldn't IMO - I think setting it in the client and passing it through this way is the correct thing to do. (Until we actually put the form rendering logic in the client and then all of this could go away, but if it is on the backend we should be specifying we're building special tool states for the form stuff like this.)

jmchilton (approved):

PR #9316, opened 2020-02-04
[20.01] Various fixes and enhancements to reports

mvdbeek:
With those fixes in place I was able to produce this report PDF:
Screenshot 2020-02-04 at 19 39 30


mvdbeek:
I guess https://github.com/galaxyproject/galaxy/pull/9316/commits/2804f5e17fa809f41a91533f1a364ec13754d7b3 isn't necessary with quotes, but I only figured this out when the unit test failed. If this needs spaces the examples should use quotes as well ... so should we allow this without quotes or not ?

jmchilton (approved):
Amazing - thanks so much for the awesome example!

Issue #9313, opened 2020-02-04
Reports app broken in Python 3


lparsons:
There are a few areas where the Reports app is not compatible with Python 3, this ticket serves to document those.

* **User disk usage**
```
Error - <class 'TypeError'>: '<' not supported between instances of 'decimal.Decimal' and 'NoneType'
URL: http://galaxy-reports-dev.princeton.edu/users/user_disk_usage?sort_id=default&order=default
File 'lib/galaxy/web/framework/middleware/error.py', line 154 in __call__
  app_iter = self.application(environ, sr_checker)
File '/data/galaxy/galaxy-app/.venv/lib/python3.6/site-packages/paste/recursive.py', line 85 in __call__
  return self.application(environ, start_response)
File '/data/galaxy/galaxy-app/.venv/lib/python3.6/site-packages/paste/httpexceptions.py', line 640 in __call__
  return self.application(environ, start_response)
File 'lib/galaxy/web/framework/base.py', line 143 in __call__
  return self.handle_request(environ, start_response)
File 'lib/galaxy/web/framework/base.py', line 222 in handle_request
  body = method(trans, **kwargs)
File 'lib/galaxy/webapps/reports/controllers/users.py', line 181 in user_disk_usage
  users = sorted(trans.sa_session.query(galaxy.model.User).all(), key=operator.attrgetter(str(sort_id)), reverse=_order)
TypeError: '<' not supported between instances of 'decimal.Decimal' and 'NoneType'
```

* **Disk space maintenance** - **FIXED in #9319**
```
File 'lib/galaxy/web/framework/middleware/error.py', line 154 in __call__
  app_iter = self.application(environ, sr_checker)
File '/data/galaxy/galaxy-app/.venv/lib/python3.6/site-packages/paste/recursive.py', line 85 in __call__
  return self.application(environ, start_response)
File '/data/galaxy/galaxy-app/.venv/lib/python3.6/site-packages/paste/httpexceptions.py', line 640 in __call__
  return self.application(environ, start_response)
File 'lib/galaxy/web/framework/base.py', line 143 in __call__
  return self.handle_request(environ, start_response)
File 'lib/galaxy/web/framework/base.py', line 222 in handle_request
  body = method(trans, **kwargs)
File 'lib/galaxy/webapps/reports/controllers/system.py', line 33 in index
  file_path, disk_usage, datasets, file_size_str = self.disk_usage(trans, **kwd)
[...]

PR #9319, opened 2020-02-05
[20.01] Python3: unicodify the output of ``subprocess.check_output()``

nsoranzo:
By default it's a bytestring.

Also:
- don't use `shell=True`
- add a `splitlines()` forgotten in commit 670897ec5ca2284c7fad453c3d595defe3cca280
- raise exception instead of returning `error_info` in `docker_to_singularity()`
- use a random temp dir in `singularity_container_test()`

Fix second bug reported in https://github.com/galaxyproject/galaxy/issues/9313 .


nsoranzo:
@simonbray You may want to check the changes to `lib/galaxy/tool_util/deps/mulled/mulled_update_singularity_containers.py` .

simonbray:
@nsoranzo, thanks, looks good to me.

mvdbeek (approved):

mvdbeek:
Thanks @nsoranzo, very nice!

PR #9320, opened 2020-02-05
[19.09] Move sam_pileup to GALAXY_LIB_TOOLS_VERSIONED

mvdbeek:
Will be fixed in https://github.com/galaxyproject/tools-devteam/pull/545

### DIFF
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -163,7 +163,6 @@ GALAXY_LIB_TOOLS_UNVERSIONED = [
     "lastz_paired_reads_wrapper",
     "subRate1",
     "substitutions1",
-    "sam_pileup",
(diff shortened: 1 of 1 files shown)


jmchilton (approved):

lib/galaxy/tools/__init__.py
```diff
@@ -188,6 +187,7 @@
     "lastz_wrapper_2": packaging.version.parse("1.3"),
     "PEsortedSAM2readprofile": packaging.version.parse("1.1.1"),
     "sam_to_bam": packaging.version.parse("1.1.3"),
+    "sam_pileup": packaging.version.parse("1.1.2"),
```
nsoranzo, line 190: @mvdbeek Shouldn't the version be `1.1.3` here?

PR #9323, opened 2020-02-05
Remove unused and deprecated parameters from k8s runner

nuwang:


bernt-matthias:
Because I just stumbled over it: could you check if this https://github.com/galaxyproject/galaxy/blob/470031e1564deaed7e67d041b8563f824a52fb3b/lib/galaxy/jobs/runners/kubernetes.py#L272 is still done?

mvdbeek:
@nuwang do you want to get this into 20.05 ?

nuwang:
@bernt-matthias Thanks for pointing that out, that's no longer used either. I'll update the PR.
@mvdbeek Somehow, I missed both the comments on this thread. Sorry about that. Will update the PR for the next release.

mvdbeek:
@nuwang I've rebased this against dev, I'm gonna say let's merge this, unless you think this is not ready ?

nuwang:
@mvdbeek Thanks for rebasing. I think it's good to go!

PR #9327, opened 2020-02-06
[20.01] Fix string module use

mvdbeek:


lib/galaxy/visualization/plugins/interactive_environments.py
```diff
@@ -357,7 +356,7 @@ def _get_command_inject_env(self):
                 envsets.append(item[2:])
             elif item.startswith('--env'):
                 envsets.append(item[5:])
-        return dict(map(lambda s: string.split(s, '=', 1), envsets))
+        return dict(map(lambda s: s.split(s, '=', 1), envsets))
```
nsoranzo:
```suggestion
        return dict(map(lambda s: s.split('=', 1), envsets))
```

tools/plotting/bar_chart.py
```diff
@@ -43,7 +42,7 @@ def main(tmpFileName):
     try:
         in_file = open(sys.argv[1], 'r')
         xtic = int(sys.argv[2])
-        col_list = string.split(sys.argv[3], ",")
+        col_list = sys.argv[3].split(",")
```
jdavcs:
```suggestion
        col_list = sys.argv[3].split(',')
```
for the sake of consistency?

jdavcs (approved):

lib/galaxy/visualization/plugins/interactive_environments.py
```diff
@@ -357,7 +356,7 @@ def _get_command_inject_env(self):
                 envsets.append(item[2:])
             elif item.startswith('--env'):
                 envsets.append(item[5:])
-        return dict(map(lambda s: string.split(s, '=', 1), envsets))
+        return dict(map(lambda s: s.split('=', 1), envsets))
```
nsoranzo: If you prefer a dict comprehension:

```suggestion
        return dict(_.split('=', 1) for _ in envsets)
```

nsoranzo (approved):

PR #9330, opened 2020-02-06
[20.01] Fix use of next() as a meethod instead of a function

natefoo:


mvdbeek (approved):

dannon (approved):

dannon:
next should be fine with python 2.6+

PR #9332, opened 2020-02-06
Replacing Workflow Editor Node with Vue-component

guerler:
This PR replaces the static node template building operations in the `prebuildNode` function with a `Node` component. In the future this shall enable us to substantially overhaul the workflow nodes appearance and behavior. The `Node` component is wrapped in a parent `div` whose CSS properties and drag-drop features are managed by the workflow editor client code. In follow-ups these CSS operations will be reduced to modifications of node size and position only. Currently the node title and additional content related changes are still handled by the workflow client code. This PR also starts stripping a majority of the `Backbone` dependency from the workflow client code. Instead of `Backbone` views and models the workflow editor will strictly use ES6 classes and jQuery to modify the node css and add drag-drop features.

### DIFF
--- a/client/galaxy/scripts/components/Workflow/Editor/Index.vue
+++ b/client/galaxy/scripts/components/Workflow/Editor/Index.vue
@@ -44,7 +44,7 @@
 </template>
 
 <script>
-import WorkflowView from "mvc/workflow/workflow-view";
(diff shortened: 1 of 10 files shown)


guerler:
Thank you for the review!

Issue #8808, opened 2019-10-15
Some content of the job working directories is deleted despite cleanup_job = onsuccess


bernt-matthias:
On 19.05 I observed that the `galaxy_JOBID.[sh,e,o]` files are deleted for failed jobs are deleted despite `cleanup_job = onsuccess`. The job dir itself and other files like `tool_script.sh` are not deleted. 

If someone could tell me where to start digging I might fix it.

innovate-invent:
This issue is also in 19.09.
I just attempted setting `cleanup_job: never` and it still deletes the galaxy_***.sh file.

bernt-matthias:
This function does the deletion

https://github.com/galaxyproject/galaxy/blob/b2a61c09c605222cb935512fb873e759400c467f/lib/galaxy/jobs/runners/__init__.py#L593

don't know yet, from where the function is called.

PR #9335, opened 2020-02-07
disable activation of galaxy venv for jobs

bernt-matthias:
So far only a test .. I don't understand what this is good for. So lets disable this and check what fails.

### DIFF
--- a/lib/galaxy/jobs/runners/util/job_script/DEFAULT_JOB_FILE_TEMPLATE.sh
+++ b/lib/galaxy/jobs/runners/util/job_script/DEFAULT_JOB_FILE_TEMPLATE.sh
@@ -19,10 +19,10 @@ _galaxy_setup_environment() {
     # These don't get cleaned on a re-run but may in the future.
     [ -z "$_GALAXY_JOB_TMP_DIR" -a ! -f "$_GALAXY_JOB_TMP_DIR" ] || mkdir -p "$_GALAXY_JOB_TMP_DIR"
     [ -z "$_GALAXY_JOB_HOME_DIR" -a ! -f "$_GALAXY_JOB_HOME_DIR" ] || mkdir -p "$_GALAXY_JOB_HOME_DIR"
-    if [ "$GALAXY_VIRTUAL_ENV" != "None" -a -f "$GALAXY_VIRTUAL_ENV/bin/activate" \
-         -a "`command -v python`" != "$GALAXY_VIRTUAL_ENV/bin/python" ]; then
-        . "$GALAXY_VIRTUAL_ENV/bin/activate"
-    fi
+#     if [ "$GALAXY_VIRTUAL_ENV" != "None" -a -f "$GALAXY_VIRTUAL_ENV/bin/activate" \
+#          -a "`command -v python`" != "$GALAXY_VIRTUAL_ENV/bin/python" ]; then
+#         . "$GALAXY_VIRTUAL_ENV/bin/activate"
+#     fi
 }
 
 $integrity_injection


mvdbeek:
https://github.com/mvdbeek/galaxy/commit/5994311d8274d8b946f51d28efcca17d31a05bbc
is the original implementation. I don't think this should be ~~touched~~ removed. You might want to open an issue with the problem you're seeing ~~instead~~.

bernt-matthias:
Maybe .. But one has to acknowledge that surprisingly many (all .. except one that is likely unrelated) tests are passing. 

I have not opened an issue yet, because I don't understand what's going on. What I observe is the following: The tool gops_join_1 fails despite the apparently successful activation of its conda environment with a python `ImportError`. Since I have observed previously that the combination of a python venv and conda environments may be problematic (https://github.com/galaxyproject/planemo/issues/965) I wanted to check if this may also be the case here.

And then I was wondering if one needs to load Galaxy's venv at all for the tool_script.

Lets look at (the relevant parts of) a job script 

1. `_galaxy_setup_environment` is called
   - sets `PYTHONPATH` (I guess only if preserve_python_environment is set or for special tools as you mentioned on gitter https://github.com/mvdbeek/galaxy/blob/c7078e5ad273d4497c86f8eea352e9b14c75e051/lib/galaxy/tools/__init__.py#L133) 
   - activates Galaxy's venv as defined by `GALAXY_VIRTUAL_ENV`
2. `tool_script.sh is called`
3. activate Galaxy's venv as defined by `_GALAXY_VIRTUAL_ENV` (if not already activated)
4. set metadata

```
#!/bin/bash

_galaxy_setup_environment() {
    local _use_framework_galaxy="$1"
...
    if [ "$GALAXY_LIB" != "None" -a "$_use_framework_galaxy" = "True" ]; then
        if [ -n "$PYTHONPATH" ]; then
            PYTHONPATH="$GALAXY_LIB:$PYTHONPATH"
        else
            PYTHONPATH="$GALAXY_LIB"
        fi
        export PYTHONPATH
    fi
...
    if [ "$GALAXY_VIRTUAL_ENV" != "None" -a -f "$GALAXY_VIRTUAL_ENV/bin/activate" \
         -a "`command -v python`" != "$GALAXY_VIRTUAL_ENV/bin/python" ]; then
        . "$GALAXY_VIRTUAL_ENV/bin/activate"
    fi
}


...
GALAXY_VIRTUAL_ENV="/gpfs1/data/galaxy_server/galaxy/.venv"
_GALAXY_VIRTUAL_ENV="/gpfs1/data/galaxy_server/galaxy/.venv"
PRESERVE_GALAXY_ENVIRONMENT="False"
GALAXY_LIB="/gpfs1/data/galaxy_server/galaxy/lib"

...

rm -rf working; mkdir -p working; cd working; /bin/bash /work/songalax/galaxy/database/jobs_directory/059/59338/tool_script.sh > ../tool_stdout 2> ../tool_stderr; return_code=$?; cd '/work/songalax/galaxy/database/jobs_directory/059/59338'; 
export PATH=$PATH:'/gpfs1/data/galaxy_server/galaxy/database/dependencies/_conda/envs/__bcftools@1.5/bin' ; python "metadata/set.py"; sh -c "exit $return_code"
```

I just can't see a reason to make the venv available to the tool_script (but I still know only a few parts of the universe). 

- For the docker runner there seems to be an option to disable the loading: https://github.com/mvdbeek/galaxy/blob/f5205df698425eedeb514feb27886c3146a2e281/lib/galaxy/jobs/runners/godocker.py#L377
- For kubernetes there is also a link indicating that this is done (but it seems that this is not the case anymore) https://github.com/mvdbeek/galaxy/blob/f5205df698425eedeb514feb27886c3146a2e281/lib/galaxy/jobs/runners/kubernetes.py#L277

For the the metadata setting there is a separate `_galaxy_setup_environment True` call which would activate the venv. But to make this work correctly I would expect that `GALAXY_VIRTUAL_ENV` should be initialized with `"None"` .. so maybe the tests don't catch problems with absent venv anyway.

mvdbeek:
> Maybe .. But one has to acknowledge that surprisingly many (all .. except one that is likely unrelated) tests are passing.

Yes, as mentioned in the commit I linked to this is for overwriting the GALAXY_VIRTUAL_ENV on a per destination basis. That is not something we test.

> I have not opened an issue yet, because I don't understand what's going on.

That's fine, that is really not a prerequisite to open an issue. A plain description of what doesn't work is more than enough, you don't have to fix every bug yourself. For instance in your case it's probably sufficient if you included the logs for one tool  run.

mvdbeek:
Can you post the import error ? A quick test with planemo doesn't reveal any issue for gops_join_1

bernt-matthias:
Sorry for not opening an issue so far. Here is the import error:

```
  File "/gpfs1/data/galaxy_server/galaxy/shed_tools/toolshed.g2.bx.psu.edu/repos/devteam/intersect/33b3f3688db4/intersect/gops_intersect.py", line 23, in <module>
    from galaxy.tools.util.galaxyops import fail, parse_cols_arg, skipped
ModuleNotFoundError: No module named 'galaxy'
```

mvdbeek:
Are the dependencies installed ? This is provided by https://github.com/galaxyproject/tools-devteam/blob/master/tool_collections/gops/macros.xml#L6. Can you also post the dependency resolution log for that tool run ?

bernt-matthias:
> dependency resolution log

what do you mean with this? 

`conda_activate.log`? this is empty. 

In the logs I see the following .. which looks really odd

```
Built script [/work/songalax/galaxy/database/jobs_directory/058/58676/tool_script.sh] for tool command [[ "$(basename "$CONDA_DEFAULT_ENV")" = "$(basename '/gpfs1/data/galaxy_server/galaxy/database/dependencies/_conda/envs/mulled-v1-a4cb0f5781d6f9de5e288400a2ac9fa34202068e5723b723249629522600dc67')" ] ||
:done ; python '/gpfs1/data/galaxy_server/galaxy/shed_tools/toolshed.g2.bx.psu.edu/repos/devteam/intersect/33b3f3688db4/intersect/gops_intersect.py' '/gpfs1/data/galaxy_server/galaxy/database/files/000/139/dataset_139948.dat' '/gpfs1/data/galaxy_server/galaxy/database/files/000/139/dataset_139988.dat' '/work/songalax/galaxy/database/jobs_directory/058/58676/galaxy_dataset_221636.dat'  -1 1,2,3,6  -2 1,2,3,6  -m 1]
```

Unfortunately the `galaxy_JOBID.sh` is missing for that job (I always thought because of this https://github.com/galaxyproject/galaxy/issues/8808)

mvdbeek:
Before that you should have the resolution logs. Where Galaxy tries to find the dependencies.

bernt-matthias:
Just restarted it. Now I got the complete thing:

```
galaxy.tool_util.deps DEBUG 2020-02-20 10:58:20,981 [p:10245,w:0,m:1] [UnivaJobRunner.work_thread-1] Using dependency bx-python version 0.7.1 of type conda
galaxy.tool_util.deps DEBUG 2020-02-20 10:58:20,981 [p:10245,w:0,m:1] [UnivaJobRunner.work_thread-1] Using dependency galaxy-ops version 1.0.0 of type conda
galaxy.jobs.command_factory INFO 2020-02-20 10:58:21,041 [p:10245,w:0,m:1] [UnivaJobRunner.work_thread-1] Built script [/work/songalax/galaxy/database/jobs_directory/060/60354/tool_script.sh] for tool command [[
 "$(basename "$CONDA_DEFAULT_ENV")" = "$(basename '/gpfs1/data/galaxy_server/galaxy/database/dependencies/_conda/envs/mulled-v1-a4cb0f5781d6f9de5e288400a2ac9fa34202068e5723b723249629522600dc67')" ] ||
MAX_TRIES=3
COUNT=0
while [ $COUNT -lt $MAX_TRIES ]; do
    . '/gpfs1/data/galaxy_server/galaxy/database/dependencies/_conda/bin/activate' '/gpfs1/data/galaxy_server/galaxy/database/dependencies/_conda/envs/mulled-v1-a4cb0f5781d6f9de5e288400a2ac9fa34202068e5723b72324
9629522600dc67' > conda_activate.log 2>&1
    if [ $? -eq 0 ];then
        break
    else
        let COUNT=COUNT+1
        if [ $COUNT -eq $MAX_TRIES ];then
            echo "Failed to activate conda environment! Error was:"
            cat conda_activate.log
            exit 1
        fi
        sleep 10s
    fi
done ; [ "$(basename "$CONDA_DEFAULT_ENV")" = "$(basename '/gpfs1/data/galaxy_server/galaxy/database/dependencies/_conda/envs/mulled-v1-a4cb0f5781d6f9de5e288400a2ac9fa34202068e5723b723249629522600dc67')" ] ||
MAX_TRIES=3
COUNT=0
while [ $COUNT -lt $MAX_TRIES ]; do
    . '/gpfs1/data/galaxy_server/galaxy/database/dependencies/_conda/bin/activate' '/gpfs1/data/galaxy_server/galaxy/database/dependencies/_conda/envs/mulled-v1-a4cb0f5781d6f9de5e288400a2ac9fa34202068e5723b723249629522600dc67' > conda_activate.log 2>&1
    if [ $? -eq 0 ];then
        break
    else
        let COUNT=COUNT+1
        if [ $COUNT -eq $MAX_TRIES ];then
            echo "Failed to activate conda environment! Error was:"
            cat conda_activate.log
            exit 1
        fi
        sleep 10s
    fi
done ; python '/gpfs1/data/galaxy_server/galaxy/shed_tools/toolshed.g2.bx.psu.edu/repos/devteam/intersect/33b3f3688db4/intersect/gops_intersect.py' '/gpfs1/data/galaxy_server/galaxy/database/files/000/139/dataset_139948.dat' '/gpfs1/data/galaxy_server/galaxy/database/files/000/139/dataset_139988.dat' '/work/songalax/galaxy/database/jobs_directory/060/60354/galaxy_dataset_240541.dat'  -1 1,2,3,6  -2 1,2,3,6  -m 1]
```

mvdbeek:
What's the output of
`. '/gpfs1/data/galaxy_server/galaxy/database/dependencies/_conda/bin/activate' '/gpfs1/data/galaxy_server/galaxy/database/dependencies/_conda/envs/mulled-v1-a4cb0f5781d6f9de5e288400a2ac9fa34202068e5723b723249629522600dc67' && python -c "from galaxy.tools.util.galaxyops import fail, parse_cols_arg, skipped "` ?

If that fails you can remove the conda environment and rebuild it.

bernt-matthias:
Really odd. I can call the tool script from the comman line without problems. But if I run the Galaxy tool it fails. 

Running even as the same user.

mvdbeek:
You're not setting PYTHONPATH or PYTHONLIB anywhere, are you ? maybe drop a `env` statement in the tool to check your environment.

bernt-matthias:
Puh... indeed PYTHONPATH is set by the python provided on our system. There is crazy stuff going on (modification of sys.path) in the sitecustomize.py script that is contained in there. Seems to be an easybuild python.

Could you explain me why setting PYTHONPATH could be bad here? I could need some arguments for the discussions with our admins.

mvdbeek:
https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPATH for what it does. I'm not sure what exactly is going on, but could you try this?:
```
diff --git a/lib/galaxy/tool_util/deps/resolvers/conda.py b/lib/galaxy/tool_util/deps/resolvers/conda.py
index 83b17e85e6..b7435cd883 100644
--- a/lib/galaxy/tool_util/deps/resolvers/conda.py
+++ b/lib/galaxy/tool_util/deps/resolvers/conda.py
@@ -46,6 +46,7 @@ CONDA_SOURCE_CMD = """[ "$(basename "$CONDA_DEFAULT_ENV")" = "$(basename '{envir
 MAX_TRIES=3
 COUNT=0
 while [ $COUNT -lt $MAX_TRIES ]; do
+    unset PYTHONPATH PYTHONHOME
     . '{activate_path}' '{environment_path}' > conda_activate.log 2>&1
     if [ $? -eq 0 ];then
         break

```

bernt-matthias:
I can do this next week... Since I currently can not restart my Galaxy instances.

bernt-matthias:
@mvdbeek Just stumbled over this again and tested your suggestion. Unfortunately it does not work. 

I did some additional research: In the original setup I checked the content of `$PATH` on tool execution is:

- 1. `GALAXYHOME/.venv/bin`
- 2. `/software/easybuild-broadwell/software/MPI/GCC/7.3.0-2.30/OpenMPI/3.1.1/Python/3.6.6/bin` this is due to our systems software system (an easybuild python provided via lmod). 
- 3. `GALAXYHOME/database/dependencies/_conda/envs/__biopython@1.70/bin`

I do not understand how this order is generated. In the job script the conda env is activated last .. so the entry should be the first, isn't it? This is similar to a problem that some had with planemo (where also venv's and conda envs are used in combination) https://github.com/galaxyproject/planemo/issues/965. 

The fix suggested over there also helps here: `conda config --set auto_activate_base false` or remove the `conda init` generated part from `.bashrc`. 

Looks a bit like this one https://github.com/conda/conda/issues/9392 which is the only issue that I could find where it is reported that the conda env path is inserted in the middle of PATH.

mvdbeek:
> I do not understand how this order is generated. In

Is that for `gops_intersect` or `gops_join`? I hadn't noticed this previously since you wrote that you're using `gops_join`, but `gops_join` does not use `gops_intersect.py`.
If that is gops_join then
https://github.com/mvdbeek/galaxy/blob/f566c87b3cebfae314a79488ceb0b6f3e4d2924e/lib/galaxy/tools/__init__.py#L176 would trigger https://github.com/mvdbeek/galaxy/blob/f566c87b3cebfae314a79488ceb0b6f3e4d2924e/lib/galaxy/tool_util/deps/resolvers/conda.py#L407 and the Conda environment's PATH would be appended. Now none of these use bioconda, so I have no idea where this is coming from.

bernt-matthias:
My last message considered https://github.com/galaxyproject/tools-iuc/tree/master/tools/ncbi_entrez_eutils ..

nsoranzo:
> The fix suggested over there also helps here: conda config --set auto_activate_base false or remove the conda init generated part from .bashrc.

This issue and the "remove conda init from .bashrc" workaround were confirmed by @NatPRoach on https://gitter.im/galaxyproject/dev

bernt-matthias:
> "remove conda init from .bashrc" workaround

which unfortunately is nearly impossible to implement in a Galaxy that submits jobs as real user... but I guess an upstream fix is needed.

nsoranzo:
> > "remove conda init from .bashrc" workaround
> 
> which unfortunately is nearly impossible to implement in a Galaxy that submits jobs as real user... but I guess an upstream fix is needed.

Can we just deactivate any active conda env before activating the one for the tool requirements?

bernt-matthias:
Sounds like a good idea. `conda deactivate` seems not to fail if no environment is active.

 I could do this also locally in my environment_setup_file .. or might this be useful in general?

nsoranzo:
> Sounds like a good idea. `conda deactivate` seems not to fail if no environment is active.
> 
> I could do this also locally in my environment_setup_file .. or might this be useful in general?

I mean in general, as part of the job script.

bernt-matthias:
Thanks for the help @mvdbeek .. guess there is no need to keep this open

PR #9336, opened 2020-02-07
[20.01] Add requirements, fix legacy tools and move them out of GALAXY_LIB_TOOLS_UNVERSIONED

mvdbeek:
~~Needs a new galaxy-sequence-utils release~~

### DIFF
--- a/tools/maf/vcf_to_maf_customtrack.py
+++ b/tools/maf/vcf_to_maf_customtrack.py
@@ -13,51 +13,43 @@ UNKNOWN_NUCLEOTIDE = '*'
 
 class PopulationVCFParser(Iterator):
     def __init__(self, reader, name):
-        self.reader = reader
+        self.reader = iter(reader)
         self.name = name
         self.counter = 0
 
-    def __next__(self):
-        rval = []
-        vc = next(self.reader)
-        for i, allele in enumerate(vc.alt):
-            rval.append(('%s_%i.%i' % (self.name, i + 1, self.counter + 1), allele))
-        self.counter += 1
-        return (vc, rval)
-
     def __iter__(self):
-        while True:
-            yield next(self)
+        for vc in self.reader:
+            rval = []
+            for i, allele in enumerate(vc.alt):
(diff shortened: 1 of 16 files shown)


tools/maf/vcf_to_maf_customtrack.xml
```diff
@@ -1,5 +1,9 @@
-<tool id="vcf_to_maf_customtrack1" name="VCF to MAF Custom Track" version="1.0.0">
+<tool id="vcf_to_maf_customtrack1" name="VCF to MAF Custom Track" version="1.0.1">
   <description>for display at UCSC</description>
+  <requirements>
+    <requirement type="package" version="1.1.3">galaxy_sequence_utils</requirement>
```
jdavcs: 1.1.4?

tools/maf/vcf_to_maf_customtrack.py
```diff
@@ -13,51 +13,43 @@
 
 class PopulationVCFParser(Iterator):
     def __init__(self, reader, name):
-        self.reader = reader
+        self.reader = iter(reader)
```
jdavcs, line 16: A tangent. I'm wondering... We have to do this because the reader's `__iter__` method does not return `self` (i.e., it's not an iterator without that), right? Would it makes sense to refactor that code so that it doesn't require this step (wrapping in `iter()` and, thus, is more straightforward to use?

mvdbeek: Sure, why not, but I don't think that would be a good use of time. These tools are legacy and I doubt anyone would really build new stuff based on galaxy-sequence-utils.

jdavcs: Oh OK, makes sense. Never mind then, there's only so much time.

bgruening (approved):

mvdbeek:
I think that's all we can do for Converters and tools shipped with galaxy. It would be good to move maf_utilties and gff_utils to a separate package (or galaxy-sequence-utils), with that in place I think only `upload1`, `__DATA_FETCH__` and `send_to_cloud` would need to remain on the list (and maybe we could even make them depend on a fatter set of galaxy dependencies).

PR #9337, opened 2020-02-07
[20.01] Various python 3 fixes from sentry

mvdbeek:
Fixes a couple of issue from sentry that are clearly python 3 related. Also fixes a bit of client code for starting circster.

### DIFF
--- a/lib/galaxy/visualization/data_providers/genome.py
+++ b/lib/galaxy/visualization/data_providers/genome.py
@@ -1223,11 +1223,11 @@ class BBIDataProvider(GenomeDataProvider):
 
             # Start with N samples.
             num_points = num_samples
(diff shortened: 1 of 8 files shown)


lib/galaxy/visualization/data_providers/genome.py
```diff
@@ -1223,11 +1223,11 @@ def summarize_region(bbi, chrom, start, end, num_points):
 
             # Start with N samples.
             num_points = num_samples
-            step_size = (end - start) / num_points
+            step_size = int((end - start) / num_points)
```
jdavcs: maybe just use floor division operator `//`?

jmchilton (approved):

bgruening:
Thanks @mvdbeek!

PR #9342, opened 2020-02-07
Backward compatibility with py2 for TSV<->CSV conversion

### LINKED ISSUE #9341: TSV<->CSV conversion failing
```
  File "/home/dave/github-repos/galaxy/lib/galaxy/datatypes/converters/tabular_csv.py", line 44, in <module>
    main()
  File "/home/dave/github-repos/galaxy/lib/galaxy/datatypes/converters/tabular_csv.py", line 22, in main
    convert_to_csv(input_fname, output_fname)
  File "/home/dave/github-repos/galaxy/lib/galaxy/datatypes/converters/tabular_csv.py", line 40, in convert_to_csv
    writer.writerow(line.strip().split('\t'))
TypeError: a bytes-like object is required, not 'str'
```

davebx:
Resolves #9341


lib/galaxy/datatypes/converters/tabular_csv.py
```diff
@@ -7,7 +7,17 @@
...
 
+from contextlib import contextmanager
+
+@contextmanager
+def compat_open(file, mode):
+    if sys.version_info[0] > 2:
+        fh = open(file, mode, newline='')
+    else:
+        fh = open(file, mode + 'b')
+    yield fh
```
jdavcs:
```suggestion
    try:
        yield fh
    finally:
        fh.close()
```
The file needs to be closed upon exit from the context manager; I also think the yield statement needs to be wrapped in a try/finally: if an exception is raised in the `with` block, it will be re-raised in the `yield` statement. Without error handling whatever follows the yield statement (`fh.close()`) won't be executed.

I'm not sure whether we should worry about not wrapping the code that actually opens the file. Wrapping it in the same try block seems wrong, but multiple try blocks seem like overkill here. Depends on what we want the system to do if the file doesn't open here; and I don't know.

jdavcs (changes requested):
Nice solution! (see inline comment).
Other than that, just 3 minor lint errors.

nsoranzo:
This can be simplified by using only the Py3 syntax if https://github.com/galaxyproject/galaxy/pull/9345/files#diff-810d43c5eb8bd1186917f323691dd578 gets merged.

lib/galaxy/datatypes/converters/tabular_csv.py
jdavcs: Correction: see comment by @nsoranzo

jdavcs:
1. The unit test failure is unrelated.
2. I'm ambivalent about merging as is vs removing py2 compatibility altogether, pending #9345 as per @nsoranzo . On the one hand, the current version fixes a problem that currently exists in a reasonable way, without breaking the converter under Py2. On the other hand, once 9345 is merged, this will have introduced more "cruft" into the code base. 
I'll go ahead and approve this. Unless others feel it's better to wait for 9345, my suggestion is to merge this as is.

jdavcs (approved):
See previous comment about merging as is vs. waiting for the referenced PR.

jdavcs:
#9345 has been merged, so we don't need to make this Python2-compatible anymore.

mvdbeek:
Thanks @davebx, I'm gonna close this since we merged https://github.com/galaxyproject/galaxy/pull/9345

PR #9343, opened 2020-02-08
[19.09] Add bam_native alias of unsorted.bam

mvdbeek:
Fixes https://sentry.galaxyproject.org/sentry/main/issues/307648/events/?cursor=1579818609000%3A2%3A0:
```
Datatype class not found for extension 'bam_native'
```
Probably an old version of bowtie2 that uses this datatype.

### DIFF
--- a/lib/galaxy/config/sample/datatypes_conf.xml.sample
+++ b/lib/galaxy/config/sample/datatypes_conf.xml.sample
@@ -41,6 +41,7 @@
       <converter file="to_coordinate_sorted_bam.xml" target_datatype="bam"/>
       <converter file="to_qname_sorted_bam.xml" target_datatype="qname_sorted.bam"/>
     </datatype>
+    <datatype extension="bam_native" type="galaxy.datatypes.binary:BamNative" mimetype="application/octet-stream" subclass="true"/>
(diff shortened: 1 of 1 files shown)


jmchilton:
I don't think we want to add this to core just for an old version of bowtie2? Can we just disable the tool or fix it on disk?

mvdbeek:
Fair enough, I think this can be solved with a minor update: https://github.com/galaxyproject/tools-iuc/commit/27c8f335b461d8bb1918c8ec9cf7a272714bd7b1

nsoranzo:
@mvdbeek I don't think devteam's bowtie2 was ever released on TS or TTS with `bam_native` datatype.

mvdbeek:
I think it was, and it's installed on usegalaxy.org

mvdbeek:
OK, I'm wrong, https://toolshed.g2.bx.psu.edu/repository/view_changeset?ctx_str=c3dd1aeb7d07&id=126c0918b5459666 means we didn't upload https://github.com/galaxyproject/tools-iuc/commit/27c8f335b461d8bb1918c8ec9cf7a272714bd7b1

... so how do we get bam_native usage on usegalaxy.org ?

nsoranzo:
Does Sentry store proper tracebacks? I can't see any from the link in the description.

mvdbeek:
It's just a warning unfortunately.

PR #9344, opened 2020-02-09
Create Python3 virtualenv by default

nsoranzo:
- Allow specifying Python to use to create virtualenv using `GALAXY_PYTHON` environment variable.
- Integrate Python support docs from https://galaxyproject.org/admin/python/
- Update virtualenv installation to use v16.7.9

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,24 +97,20 @@ jobs:
       - *restore_repo_cache
       - *install_tox
       - run: tox -e py27-unit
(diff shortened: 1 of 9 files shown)


nsoranzo:
I've fixed the `doc-building-test` on Jenkins with https://github.com/galaxyproject/starforge-recipes/pull/62 , a new commit here and updating the build commands.

Now green and ready for review!

mvdbeek (approved):

PR #9345, opened 2020-02-09
[20.01] Add requirements to (almost) all converters, profile and minor fixes

mvdbeek:
Skipped those converters that are in https://github.com/galaxyproject/galaxy/pull/9336 and those that currently still import stuff from Galaxy that we can't replace yet with a proper conda package.


mvdbeek:
The output data for the new tests is missing ... it turns out planemo will run all converter tests, even if testing an unrelated tool :(.

mvdbeek:
Hmm, so this is because planemo runs `run_tests.sh functional.test_toolbox`, which tests all loaded tools. Since Galaxy instances managed by planemo don't load any tools, testing the toolbox will only run the desired tool tests. Except now the converters have tests, so they will be run as well.

Given that this behavior seems like a shortcut it might be best to parse out tool ids and test cases before starting the test instance and then to select those tests.

KevinJ123 (changes requested):
Pull request

mvdbeek:
damn, times out without revealing the errors :/

mvdbeek:
Bumped the timeout up to 120 minutes. I guess we don't need to run the converter tests with every PR, but I think it's good to run them until they pass at least once.

mvdbeek:
Alright, green (mostly)! This is ready for review.

.github/workflows/converter_tests.yaml
```diff
@@ -0,0 +1,29 @@
...
+      id: pip-cache
+      with:
+        path: ~/.cache/pip
+        key: pip-cache-${{ matrix.python-version }}-${{ hashFiles('requirements.txt') }}
+    - name: fetch test data
+      run: git clone https://github.com/galaxyproject/galaxy-test-data && cp -R galaxy-test-data/* test-data
+    - name: Install planemo
+      run: pip install planemo
+    - name: Run tests
+      run: 'planemo test --galaxy_python_version ${{ matrix.python-version }} --galaxy_root .  lib/galaxy/datatypes/converters/*xml'
```
jmchilton, line 29: This is a really interesting new test, this is going to catch a lot of potential problems outside just converters!

jmchilton:
Amazing!

PR #9346, opened 2020-02-09
Remove Backbone and Globals dependency from Workflow client code

guerler:
This PR removes `Backbone` dependencies and `workflow_globals` from the `mvc/workflow` client code. It converts all classes to ES6. The `workflow-manager` class is used as interface for the Vue-components. It emits events and is instantiated in `Workflow/Index.vue`.

The `workflow-view` file has been removed. It contained a variety of functions which have been distributed as follows:
1) `API` calls were converted from `ajax` to `axios` and moved to `Editor/services.js`. They are consumed by `Workflow/Index.vue`. Loading and saving workflows is controlled in Vue now.
2) The version select building has been moved to the `WorkflowPanel` component which contains the corresponding select field.
3) Remaining helpers were moved to `Workflow/utilities` and stripped of async-requests. These helpers create visible content and modals. They can be converted to Vue components in the future if necessary.
4) The workflow parameters were integrated into Vue and are now shown together with the other workflow `Attributes`.


jmchilton:
Awesome sauce - this all looks fantastic! Thanks so much @guerler.

guerler:
Thank you!

PR #9348, opened 2020-02-10
[20.01] Make linters more robust on python 3

mvdbeek:
Fixes
```
  File "..\planemo\venv\lib\site-packages\galaxy\tool_util\loader_directory.py", line 174, in looks_like_xml
    start_contents = f.read(5 * 1024)
  File "..\planemo\venv\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 1713: character maps to <undefined>
```
reported by @VJalili 

Still weird that this is cp1252 though.

### DIFF
--- a/lib/galaxy/tool_util/loader_directory.py
+++ b/lib/galaxy/tool_util/loader_directory.py
@@ -171,7 +171,10 @@ def looks_like_xml(path, regex=TOOL_REGEX):
         return False
 
     with open(path, "r") as f:
-        start_contents = f.read(5 * 1024)
+        try:
+            start_contents = f.read(5 * 1024)
+        except UnicodeDecodeError:
+            return False
         if regex.search(start_contents):
             return True
 
--- a/lib/galaxy/tool_util/loader_directory.py
+++ b/lib/galaxy/tool_util/loader_directory.py
@@ -1,6 +1,7 @@
 """Utilities for loading and reasoning about unparsed tools in directories."""
 import fnmatch
 import glob
+import io
 import logging
 import os
 import re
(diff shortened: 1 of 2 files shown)


lib/galaxy/tool_util/loader_directory.py
```diff
@@ -171,7 +171,10 @@ def looks_like_xml(path, regex=TOOL_REGEX):
         return False
 
     with open(path, "r") as f:
```
nsoranzo:
```suggestion
    with io.open(path, "r", encoding='utf-8') as f:
```

with adding `import io` at the top.

lib/galaxy/tool_util/loader_directory.py
```diff
@@ -171,7 +171,10 @@ def looks_like_xml(path, regex=TOOL_REGEX):
...
-        start_contents = f.read(5 * 1024)
+        try:
```
mvdbeek, line 175: we could also use a bytestring regex instead, but I doubt Galaxy will be able to read the file ...

lib/galaxy/tool_util/loader_directory.py
mvdbeek: I don't think we want to do this here, it prevents people from choosing a different locale.

nsoranzo: With are already using this in many other places, including `lib/galaxy/tool_util/verify/__init__.py` . I think it solves more issues (e.g. strange users' locales) than it creates.

mvdbeek:
Can you push to my branch?

nsoranzo (approved):

PR #9349, opened 2020-02-10
[20.01] Allow remote jobs to configure Interactive Tool entry points.

jmchilton:
Allow remote container port monitoring for ITs.

Tested with Pulsar but not sure there is anything Pulsar specific about the implementation - simply set container_monitor_result on the job destination to 'callback' (instead of implicit default of 'file'). This causes the monitor.py script to hit a new job_ports API (that mirrors job_files API) and uses the same security mechanism (per-job keys that are only active while the job is active).

PR contains fixes for embedded Pulsar configuration handling, embedded Pulsar under uwsgi, and the introduction of new feature that allows dynamic expansion of ``$UWSGI_PORT`` inside of galaxy_infrastructure_url.

Builds on open PR https://github.com/galaxyproject/galaxy/pull/8846.  Uses infrastructure URL abstraction for tools setup in https://github.com/galaxyproject/galaxy/pull/8897. 

xref original interactive tools PR (https://github.com/galaxyproject/galaxy/pull/7494).

### DIFF
--- a/lib/galaxy/job_execution/ports/view.py
+++ b/lib/galaxy/job_execution/ports/view.py
@@ -36,7 +36,7 @@ class JobPortsView(object):
         # Verify job is active. Don't update the contents of complete jobs.
         sa_session = self._app.model.context.current
         job = sa_session.query(model.Job).get(job_id)
-        if job.finished:
+        if not job.running:
             error_message = "Attempting to read or modify the files of a job that has already completed."
             raise exceptions.ItemAccessibilityException(error_message)
         return job
--- a/lib/galaxy/job_execution/ports/view.py
+++ b/lib/galaxy/job_execution/ports/view.py
@@ -1,0 +1,46 @@
+import logging
+
+from galaxy import (
+    exceptions,
+    model,
+    util
+)
+
+log = logging.getLogger(__name__)
+
+
+class JobPortsView(object):
+
+    def __init__(self, app):
+        self._app = app
+
+    def register_container_information(self, job_id, **kwd):
+        job = self.__authorize_job_access(job_id, **kwd)
+        container_runtime = kwd.get("container_runtime")
+        log.info(kwd)
+        self._app.interactivetool_manager.configure_entry_points(job, container_runtime)
+        return {"message": "ok"}
+
+    # Copy/paste from JobFilesView - TODO: de-duplicate.
+    def __authorize_job_access(self, encoded_job_id, **kwargs):
+        key = "job_key"
+        if key not in kwargs:
+            error_message = "Job files action requires a valid '%s'." % key
+            raise exceptions.ObjectAttributeMissingException(error_message)
+
+        job_id = self._security.decode_id(encoded_job_id)
+        job_key = self._security.encode_id(job_id, kind="jobs_files")
+        if not util.safe_str_cmp(kwargs["job_key"], job_key):
+            raise exceptions.ItemAccessibilityException("Invalid job_key supplied.")
+
+        # Verify job is active. Don't update the contents of complete jobs.
+        sa_session = self._app.model.context.current
+        job = sa_session.query(model.Job).get(job_id)
+        if job.finished:
+            error_message = "Attempting to read or modify the files of a job that has already completed."
+            raise exceptions.ItemAccessibilityException(error_message)
+        return job
+
+    @property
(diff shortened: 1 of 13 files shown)


jmchilton:
Progress - the monitor config is running in Pulsar's container and I filled in details related to the key handling. I don't think the initial ``container_config.json`` is available "remotely" yet though - that will probably require some Pulsar changes.

jmchilton:
It works! I should probably pull 74337fc9783fe6a9e05551287e6bb6c50dd2e894 out and open a different PR against dev since this is a PR targeting 20.01.

mvdbeek:
I'm OK with getting this into 20.01, but looks like there is a timeout in the new test?:
```
test/integration/test_interactivetools_api.py:99: 
test/integration/test_interactivetools_api.py:68: in wait_on_entry_points_active
    return wait_on(active_entry_points, "entry points to become active")

function = <function BaseInteractiveToolsIntegrationTestCase.wait_on_entry_points_active.<locals>.active_entry_points at 0x7fb25eaeb050>
desc = 'entry points to become active', timeout = 60

    def wait_on(function, desc, timeout=DEFAULT_TIMEOUT):
        delta = .25
        iteration = 0
        while True:
            total_wait = delta * iteration
            if total_wait > timeout:
                timeout_message = "Timed out after %s seconds waiting on %s." % (
                    total_wait, desc
                )
>               raise TimeoutAssertionError(timeout_message)
E               galaxy_test.base.populators.TimeoutAssertionError: Timed out after 60.25 seconds waiting on entry points to become active.

lib/galaxy_test/base/populators.py:1563: TimeoutAssertionError
```

jmchilton:
Well that is going to be hard to debug... hmm...

jmchilton:
```
test/integration/test_interactivetools_api.py::InteractiveToolsPulsarIntegrationTestCase::test_multi_server_realtime_tool PASSED [ 71%]
test/integration/test_interactivetools_api.py::InteractiveToolsPulsarIntegrationTestCase::test_simple_execution PASSED [ 71%]
```

Oh... actually some regression with the older local host tests it seems. The Pulsar tests are passing. That is helpful.

lib/galaxy/job_execution/ports/view.py
```diff
@@ -0,0 +1,46 @@
...
+
+        job_id = self._security.decode_id(encoded_job_id)
+        job_key = self._security.encode_id(job_id, kind="jobs_files")
+        if not util.safe_str_cmp(kwargs["job_key"], job_key):
+            raise exceptions.ItemAccessibilityException("Invalid job_key supplied.")
+
+        # Verify job is active. Don't update the contents of complete jobs.
+        sa_session = self._app.model.context.current
+        job = sa_session.query(model.Job).get(job_id)
+        if job.finished:
```
mvdbeek: Does it make sense to return the job if it's state is paused / queued ... anything that is not running ?

lib/galaxy/jobs/__init__.py
```diff
@@ -2099,22 +2100,33 @@ def container_monitor_command(self, container, **kwds):
...
-        container_config = os.path.join(work_dir, "container_config.json")
-
-        # TODO: implement callback via URL callback for configuring ports instead
-        #       of fs polling.
-        # if self.app.config.galaxy_infrastructure_url_set:
-        #    # TODO: settable from job destination...
-        #    infrastructure_url = self.app.config.galaxy_infrastructure_url
-        # else:
-        #    infrastructure_url = "host.docker.internal"
+        configs_dir = os.path.join(work_dir, "configs")
```
mvdbeek: We also have `__build_config_files` in the ToolEvaluator class, maybe this logic could be split between the ToolEvaluator and the Interactive tool class ? 

Those imports be dropped btw, we already `import os` at the top.

lib/galaxy/job_execution/ports/view.py
jmchilton: This code was copied from job files - where that made sense. You're right here we could probably restrict it to running. I'll do that!

mvdbeek:
There are 2 linting errors in https://app.circleci.com/jobs/github/galaxyproject/galaxy/70799/parallel-runs/0/steps/0-103

mvdbeek (approved):
Awesome 🎉

PR #9353, opened 2020-02-11
add a new function to get the local ip address

gmauro:
In our setup, the worker nodes do not have a f.q.d.n. and the previous routine was returning a not so useful '127.0.0.1'.
This function doesn't rely on hostname and doesn't need net access
Inspired by https://stackoverflow.com/questions/166506/finding-local-ip-addresses-using-pythons-stdlib


nsoranzo:
I see that `socket.gethostbyname(socket.gethostname())` is also used in `lib/galaxy/config/__init__.py` , maybe move `get_ip()` definition to `lib/galaxy/util/sockets.py` ?

gmauro:
@nsoranzo make sense.

nsoranzo:
@galaxybot test this

hexylena:
@gmauro I fixed a merge conflict, can you confirm this still looks OK to you as well?

gmauro:
Fine for me. Thank you @hexylena

PR #9354, opened 2020-02-11
[19.09] Expunge tool shed repository objects, prevents ParentInstanceDetachedError

mvdbeek:
We use SQLAlchemy's `expire_on_commit` option for sessions, which means that attributes are fetched from the database after a commit occurs.
The session used for establishing the cache however may have been closed, leading to `ParentInstanceDetached` errors.
Expunging the repo prevents the refresh and should be safe since the cache will be rebuilt upon repository installations and uninstallations.

~~I still need to check how this plays together with the tool dependency relationship.~~
There is a test that confirms this works fine for the old-style tool dependency relationships as well.

### DIFF
--- a/lib/galaxy/tools/cache.py
+++ b/lib/galaxy/tools/cache.py
@@ -149,12 +149,22 @@ class ToolShedRepositoryCache(object):
         self.repos_by_tuple[(repository.tool_shed, repository.owner, repository.name)].append(repository)
 
     def rebuild(self):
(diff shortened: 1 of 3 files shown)


mvdbeek:
Why this only affects some instances and not others: Possibly this depends on whether a separate install database is used. If it isn't, any flush could potentially lead to expired attributes.

mvdbeek:
Test failures are unrelated.

Issue #3241, opened 2016-12-01
Expand Selenium Test Coverage


jmchilton:
https://github.com/galaxyproject/galaxy/pull/3239 introduces Selenium testing of the Galaxy GUI. It should be expanded to cover as much of the GUI as possible - starting with newer JavaScript stuff since it is less likely to radically change as we get rid of mako.

These could be implemented as Python tests or even as tours that walk through various aspects of this because tour testing is available. I'm starting with a flat list - but anyone should feel free to go in an expand things - add more details, add more things to test. The closer these are to filled out little stories (what the user does, what the user expects to see, etc...) the easier it will be to implement these.

 - [x] Various upload properties.
 - [ ] History operations over many datasets
 - [ ] Multi-history view
   - [x] Collection rendering.
   - [ ] Dataset rendering.
   - [x] Creating new history from this view (https://github.com/galaxyproject/galaxy/pull/9505)
   - [x] Switching/Copying (https://github.com/galaxyproject/galaxy/pull/9505) (https://github.com/galaxyproject/galaxy/pull/9527)
   - [x] Deleting/purging (https://github.com/galaxyproject/galaxy/pull/9505)
   - [ ] Retest a subset of history panel operations (multiple dataset operations, etc..)
 - [x] History pagination (https://github.com/jmchilton/galaxy/tree/paging)
 - [x] Show structure of history.
 - [x] Export history.
 - [ ] Export history citations.
 - [x] Creating Collections
 - [x] Navigating Collections
 - [ ] Tool favorites.
 - [ ] Tool Form:
     - [x] Simple run.
     - [x] Simple re-run.
     - [x] citations
     - [ ] Tool help rendering.
     - [ ] Rerun and replace values in collection.
     - [x] Rule Parameter
     - [ ] Color parameter
     - [ ] Drill down parameter
     - [ ] Integer and float parameters w/validation.
     - [ ] Using Collections in the tool form.
     - [ ] Multi-mapping in the tool form.
     - [ ] Boolean tool parameters. 
 - [ ] Connecting nodes in the workflow editor.
   - [x] Simple.
   - [ ] Test collections connections.
   - [ ] Test collections mapping.
   - [ ] Test collections reduction.
   - [ ] Test invalid datatype rejected.
   - [ ] Test invalid collection connection.
   - [ ] Test simple input parameters.
   - [ ] Labelling outputs.
   - [ ] Labelling outputs must be unique.
   - [ ] Post job actions - rename.
[...]

PR #9355, opened 2020-02-11
implement selenium test User->Preferences->Manage information

OlegZharkov:
Expanding Selenium tests coverage according to https://github.com/galaxyproject/galaxy/issues/3241

Test capabilities under ```{http://{galaxy}/user/information```
- test changing user email
- test public name
- test user address (a tricky one, will appreciate a comment on it, see ```test_personal_information.py:get_address_input_field()``` )

### DIFF
--- a/client/galaxy/scripts/components/User/UserPreferences.vue
+++ b/client/galaxy/scripts/components/User/UserPreferences.vue
@@ -5,16 +5,16 @@
             {{ message }}
         </b-alert>
         <p>
-            You are logged in as <strong>{{ email }}</strong
+            You are logged in as <strong id="user-preferences-current-email">{{ email }}</strong
             >.
         </p>
         <b-row class="ml-3 mb-1" v-for="(link, index) in activeLinks" :key="index">
             <i :class="['pref-icon pt-1 fa fa-lg', link.icon]" />
             <div class="pref-content pr-1">
-                <a v-if="link.onclick" @click="link.onclick" href="javascript:void(0)"
+                <a :id="link.id" v-if="link.onclick" @click="link.onclick" href="javascript:void(0)"
                     ><b>{{ link.title }}</b></a
                 >
-                <a v-else :href="`${baseUrl}/${link.action}`"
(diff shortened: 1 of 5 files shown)


mvdbeek:
@galaxybot test this

bgruening:
@OlegZharkov this needs a rebase :(

OlegZharkov:
> @OlegZharkov this needs a rebase :(

@bgruening  Oh, thanks for an update! I see that GenoSpace was removed from user preferences and it led to this conflict. I will deal with it

jmchilton:
@galaxybot test this

davebx:
@OlegZharkov On the test_user_address() test, if we add the label to the div like so, in `client/galaxy/scripts/mvc/form/form-input.js:178-181`:

```
            this.$collapsible.hide();
        }
        this.$field.attr("data-label", this.model.get("label"));
    },
```
You could get away with eliminating the `get_address_input_field()` method entirely and doing
```
           input_field = get_address_form() \
                .find_element_by_xpath('.//div[@data-label="' + input_field_label + '"]') \
                .find_element_by_tag_name('input')
```

It doesn't feel like _exactly_ the right solution, but it would make it possible to at least get to the desired parent directly.

OlegZharkov:
Thanks a lot, @davebx! I just tried your idea with ```"data-label"```. This is exactly what I was looking for. That would be way cleaner, than searching for a parent element.

OlegZharkov:
@galaxybot test this

nsoranzo:
@galaxybot test this

davebx:
I'm rerunning the API test on jenkins, looks like it failed due to unrelated reasons.

OlegZharkov:
thanks @davebx, it looks like it. Why do we actually run selenium on jenkins? Why not regular github actions pipeline?

davebx (approved):

davebx:
@OlegZharkov I truly can't remember, it's been so long. My first guess would be to have these tests running on an environment we control, but that is only a guess.

OlegZharkov:
@davebx @mvdbeek thanks for reviewing!

PR #9358, opened 2020-02-11
Also expunge repos

mvdbeek:
Just opening this to get the toolshed test results.


mvdbeek:
Cool, this works, so I guess that's proof that the session is completely independent and that https://github.com/galaxyproject/galaxy/pull/9354 works

PR #9359, opened 2020-02-11
[20.01] Use RenamedTemporaryFile for safer integrated_tool_panel_file writing

mvdbeek:
If not using the largely undocumented and debug-y `integrated_tool_panel_tracking_directory`
we should use RenamedTemporaryFile which creates a temp file in the
directory where `integrated_tool_panel_file` lives, and upon succesful
write rename the file to `integrated_tool_panel_file`.
We use this in other places in the code and it seems to work well
when multiple threads are potentially modifying the target file.

### DIFF
--- a/lib/galaxy/tools/toolbox/integrated_panel.py
+++ b/lib/galaxy/tools/toolbox/integrated_panel.py
@@ -1,11 +1,11 @@
 import os
 import shutil
 import string
-import tempfile
 import time
 import traceback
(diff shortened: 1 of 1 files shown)


natefoo (approved):

jmchilton (approved):

nsoranzo (approved):

mvdbeek:
I'm gonna take those 3 approvals and merge :).

PR #9367, opened 2020-02-12
[20.01] Styleguide fixes

dannon:
This gets it functional again, still needs work.  All of the components need stubbed out data, etc., to render and be interactively tinkerable but this can be a nice list of components with their props/etc, currently.  That extra work can happen in `dev`, though.  This makes it work for the release.


mvdbeek:
Thanks @dannon !

PR #9368, opened 2020-02-12
Fix success message in user preferences, add help text to toolbox filter page

guerler:
This PR fixes the display of status messages in the user preference panel. This can be tested by changing and saving a user property in the `Manage Information` section. Currently dev does not show a success message.
Additionally this PR adds a help label to the Toolbox filter page. Currently visiting the Toolbox filter page can be confusing since by default filters are not functional. Instead the example filters initially display error messages until properly configured.
Finally this PR fixes an issue with the error handling on quota editing. If an invalid string is entered into the quota editing field in the admin panel, a `ValueError` is created but not parsed properly to the client. Admin actions commonly throw errors of type `ActionInputError`, which are then caught and handled by consuming classes like e.g. the admin controller.


dannon (commented):
@guerler Thanks for resolving that conflict.  This all looks good to me.

dannon (approved):

dannon:
(failing test was the virtualenv bit that's failing across the board right now; unrelated)

PR #9371, opened 2020-02-13
Move schema loading/validating code out of config

jdavcs:
This replaces #9307. 

Motivation: `config/__init__.py` is doing too much. Last year's modifications to the config system, while simplifying configuration loading, have done the opposite to the configuration module's code, and especially to the `GalaxyAppConfiguration` class. This is an attempt to address this problem by separating 2 different responsibilities of the `GalaxyAppConfiguration` class: (1) automatically loading and processing all config properties from the schema; and (2) setting individual config properties and handling special cases. My first attempt (#9307) was to delegate the former to a ConfigurationLoader module and use it for any further schema processing, while using GalaxyAppConfiguration for setting individual config properties (as it did before).

However, concerns were raised during the code review: (a) the schema loader object was manipulating the internal structure of the config object; and (b) such tight coupling of 2 objects suggested the separation may have been not necessary (https://github.com/galaxyproject/galaxy/pull/9307#pullrequestreview-356385467). 

This, I think, is a better approach. Instead of delegating everything to a new "builder" object, I've split up the stuff that IMO does not belong in `GalaxyAppConfiguration` into 2 parts: (a) schema-related; and (b) configuration-related. All schema-related functionality has been moved to the schema object: loading collections of schema data + validating that data are now the schema object's responsibilities. All the configuration-related stuff that deals with all attributes + their processing, and does *not* include setting individual, galaxy-specific properties, has been moved up to the base config class.

I've updated the tests; I've also simplified the mocking in the unit tests in `test/unit/config`. 

I expect this will simplify the config code and make modifications easier. Also, as a side benefit, we can now eliminate much redundant code from TS's config (e.g., #9379, if needed).

Ping @nsoranzo, @bgruening.  

(EDIT: more details in config messages.)

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -104,6 +104,9 @@ def find_root(kwargs):
 
 
 class BaseAppConfiguration(object):
+    # Override in subclasses (optional): {KEY: config option, VALUE: deprecated directory name}
+    # If VALUE == first directory in a user-supplied path that resolves to KEY, it will be stripped from that path
+    deprecated_dirs = None
 
     def __init__(self, **kwargs):
         self.config_dict = kwargs
@@ -113,7 +116,7 @@ class BaseAppConfiguration(object):
         self._raw_config = self.schema.defaults.copy()  # Save schema defaults as initial config values (raw_config)
         self._update_raw_config_from_kwargs(kwargs)  # Overwrite raw_config with values passed in kwargs
         self._create_attributes_from_raw_config()  # Create attributes based on raw_config
-        self._resolve_paths(self.schema.paths_to_resolve)  # Overwrite attribute values with resolved paths
+        self._resolve_paths()  # Overwrite attribute values with resolved paths
 
     def _set_config_base(self, config_kwargs):
 
@@ -177,12 +180,6 @@ class BaseAppConfiguration(object):
         # Override in subclasses
         raise Exception('Not implemented')
 
-    def _deprecated_dirs(self):
-        # Override in subclasses (optional)
-        # {KEY: config option, VALUE: deprecated directory name}
-        # If VALUE == first directory in a user path that resolves to KEY, it will be stripped from the path
-        return None
-
     def _update_raw_config_from_kwargs(self, kwargs):
 
         def convert_datatype(key, value):
@@ -196,7 +193,7 @@ class BaseAppConfiguration(object):
             resolves_to = self.schema.paths_to_resolve.get(key)
             if resolves_to:  # value is a path that will be resolved
                 first_dir = value.split(os.sep)[0]  # get first directory component
-                if first_dir == self._deprecated_dirs().get(resolves_to):  # first_dir is deprecated for this option
+                if first_dir == self.deprecated_dirs.get(resolves_to):  # first_dir is deprecated for this option
                     ignore = first_dir + os.sep
                     log.warning(
                         "Paths for the '%s' option are now relative to '%s', remove the leading '%s' "
@@ -210,7 +207,7 @@ class BaseAppConfiguration(object):
         for key, value in kwargs.items():
             if key in self.schema.app_schema:
                 value = convert_datatype(key, value)
-                if value and self._deprecated_dirs():
+                if value and self.deprecated_dirs:
                     value = strip_deprecated_dir(key, value)
                 self._raw_config[key] = value
 
@@ -221,14 +218,14 @@ class BaseAppConfiguration(object):
             else:
                 log.debug("Attribute '%s' is set and cannot be overwritten with value '%s'" % (key, value))
 
-    def _resolve_paths(self, paths_to_resolve):
+    def _resolve_paths(self):
 
         def resolve(key):
             if key in _cache:  # resolve each path only once
                 return _cache[key]
 
             path = getattr(self, key)  # path prior to being resolved
-            parent = paths_to_resolve.get(key)
+            parent = self.schema.paths_to_resolve.get(key)
             if not parent:  # base case: nothing else needs resolving
                 return path
             parent_path = resolve(parent)  # recursively resolve parent path
@@ -242,7 +239,7 @@ class BaseAppConfiguration(object):
             return path
 
         _cache = {}
-        for key in paths_to_resolve:
+        for key in self.schema.paths_to_resolve:
             resolve(key)
 
     def _in_managed_config_dir(self, path):
@@ -294,6 +291,7 @@ class BaseAppConfiguration(object):
 class GalaxyAppConfiguration(BaseAppConfiguration):
     deprecated_options = ('database_file', 'track_jobs_in_database')
     default_config_file_name = 'galaxy.yml'
+    deprecated_dirs = {'config_dir': 'config', 'data_dir': 'database'}
 
     def __init__(self, **kwargs):
         super(GalaxyAppConfiguration, self).__init__(**kwargs)
@@ -301,9 +299,6 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
 
     def _load_schema(self):
         return AppSchema(GALAXY_CONFIG_SCHEMA_PATH, GALAXY_APP_NAME)
-
-    def _deprecated_dirs(self):
-        return {'config_dir': 'config', 'data_dir': 'database'}
 
     def _process_config(self, kwargs):
         # Resolve paths of other config files
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -118,7 +118,14 @@ class BaseAppConfiguration(object):
         self._resolve_paths(self.schema.paths_to_resolve)  # Overwrite attributes (not _raw_config) w/resolved paths
 
     def _load_schema(self):
+        # Override in subclasses
         raise Exception('Not implemented')
+
+    def _deprecated_dirs():
+        # Override in subclasses (optional)
+        # {KEY: config option, VALUE: deprecated directory name}
+        # If VALUE == first directory in a user path that resolves to KEY, it will be stripped from the path
+        return None
 
     def _update_raw_config_from_kwargs(self, kwargs):
 
@@ -133,7 +140,7 @@ class BaseAppConfiguration(object):
             resolves_to = self.schema.paths_to_resolve.get(key)
             if resolves_to:  # value is a path that will be resolved
                 first_dir = value.split(os.sep)[0]  # get first directory component
-                if first_dir == self.deprecated_dirs[resolves_to]:  # first_dir is deprecated for this option
+                if first_dir == self._deprecated_dirs().get(resolves_to):  # first_dir is deprecated for this option
                     ignore = first_dir + os.sep
                     log.warning(
                         "Paths for the '%s' option are now relative to '%s', remove the leading '%s' "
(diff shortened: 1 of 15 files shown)


lib/galaxy/config/schema.py
```diff
@@ -33,8 +37,68 @@ def __init__(self, schema_path, app_name):
         self.raw_schema = self._read_schema(schema_path)
         self.description = self.raw_schema.get("desc", None)
         app_schema = self.raw_schema['mapping'][app_name]['mapping']
-        super(AppSchema, self).__init__(app_schema)
+        self._preprocess(app_schema)
+        super().__init__(app_schema)
```
nsoranzo: This syntax is Python3-only, I don't think we are at that stage yet.

jdavcs: I wasn't sure how Python3-only our code could be. I'll keep it Python2-compatible.

lib/galaxy/config/__init__.py
```diff
@@ -162,6 +172,78 @@ def _set_config_directories():
...
 
+    def _load_schema(self):
+        # Override in subclasses
+        raise Exception('Not implemented')
+
+    def _deprecated_dirs(self):
+        # Override in subclasses (optional)
+        # {KEY: config option, VALUE: deprecated directory name}
+        # If VALUE == first directory in a user path that resolves to KEY, it will be stripped from the path
+        return None
```
nsoranzo: I'd keep `deprecated_dirs` as a dict, which here in `BaseAppConfiguration` would just be `{}`.

lib/galaxy/config/__init__.py
```diff
@@ -100,11 +100,21 @@
...
+
+    def __init__(self, **kwargs):
+        self.config_dict = kwargs
+        self.root = find_root(kwargs)
+        self._set_config_base(kwargs)
+        self.schema = self._load_schema()  # Load schema from schema definition file
+        self._raw_config = self.schema.defaults.copy()  # Save schema defaults as initial config values (raw_config)
+        self._update_raw_config_from_kwargs(kwargs)  # Overwrite raw_config with values passed in kwargs
+        self._create_attributes_from_raw_config()  # Create attributes based on raw_config
+        self._resolve_paths(self.schema.paths_to_resolve)  # Overwrite attribute values with resolved paths
```
nsoranzo: No point in passing `self.schema.paths_to_resolve` as a parameter since it's an internal method.

lib/galaxy/config/__init__.py
```diff
@@ -162,6 +172,78 @@ def _set_config_directories():
...
+                if value and self._deprecated_dirs():
+                    value = strip_deprecated_dir(key, value)
+                self._raw_config[key] = value
+
+    def _create_attributes_from_raw_config(self):
+        for key, value in self._raw_config.items():
+            if not hasattr(self, key):
+                setattr(self, key, value)
+            else:
+                log.debug("Attribute '%s' is set and cannot be overwritten with value '%s'" % (key, value))
```
nsoranzo: Is it safe to just `log.debug()` in this case? It was raising a `ConfigurationError` exception before.

lib/galaxy/config/__init__.py
jdavcs: Yes, that was an oversight. Fixing.

lib/galaxy/config/__init__.py
jdavcs: Done.

lib/galaxy/config/__init__.py
jdavcs: In my opinion, it is not only safe but also necessary. Here're my thoughts: (commit message): https://github.com/galaxyproject/galaxy/pull/9371/commits/777379729198a893519ad58f167be761a4ea867d 
And here's our original discussion (you and I made it an exception in the first place): https://github.com/galaxyproject/galaxy/pull/8728#pullrequestreview-295263870
(I unresolved that discussion so it's not hidden)

lib/galaxy/config/__init__.py
```diff
@@ -100,11 +100,24 @@
...
 def find_root(kwargs):
-    root = os.path.abspath(kwargs.get('root_dir', '.'))
-    return root
+    return os.path.abspath(kwargs.get('root_dir', '.'))
 
 
 class BaseAppConfiguration(object):
+    # Override in subclasses (optional): {KEY: config option, VALUE: deprecated directory name}
+    # If VALUE == first directory in a user-supplied path that resolves to KEY, it will be stripped from that path
+    deprecated_dirs = None
```
nsoranzo, line 109:
```suggestion
    deprecated_dirs = {}
```

natefoo: Isn't this still relevant? If the subclass doesn't override, the `.get()`s on this below will raise an exception.

lib/galaxy/config/__init__.py
natefoo: Seems this should at least be `log.warning()` then? How would the attrs be created prior to calling this method?

lib/galaxy/tools/toolbox/base.py
```diff
@@ -1217,9 +1217,10 @@ def _init_dependency_manager(self):
             return
         app_config_dict = self.app.config.config_dict
         conf_file = app_config_dict.get("dependency_resolvers_config_file")
-        default_tool_dependency_dir = os.path.join(self.app.config.data_dir, self.app.config.appschema['tool_dependency_dir'].get('default'))
-        self.dependency_manager = build_dependency_manager(app_config_dict=app_config_dict, conf_file=conf_file,
-                                                           default_tool_dependency_dir=default_tool_dependency_dir)
+        default_tool_dependency_dir = os.path.join(
+            self.app.config.data_dir, self.app.config.schema.app_schema['tool_dependency_dir'].get('default'))
```
natefoo: What is the distinction between `config.schema.defaults['foo']` and `config.schema.app_schema['foo'].get('default')`?

lib/tool_shed/webapp/config.py
```diff
@@ -34,6 +38,18 @@ class ConfigurationError(Exception):
...
+
+    def _update_raw_config_from_kwargs(self, kwargs):
+        pass  # Intentionally does nothing. To enable, remove + call base __init__().
+
+    def _create_attributes_from_raw_config(self):
+        pass  # Intentionally does nothing. To enable, remove + call base __init__().
+
+    def _resolve_paths(self, paths_to_resolve):
+        pass  # Intentionally does nothing. To enable, remove + call base __init__().
+
```
natefoo, line 52: Why doesn't the TS use these?

natefoo (approved):
Now I get what you were telling me about using DAGs, that part is pretty neat. I think overall the structure of this looks pretty good.

lib/tool_shed/webapp/config.py
jdavcs: Oh, but it's about to :-) https://github.com/galaxyproject/galaxy/pull/9379

lib/galaxy/tools/toolbox/base.py
jdavcs: @natefoo None at all! It makes sense to use it, of course; but the mock schema has no property `defaults` - so toolbox unit tests will fail. It's a trade-off between adding some minor technical dept here, and modifying galaxy mock (`MockAppConfig`, which has a definition for `MockSchema`, which is a `Bunch` and doesn't have a `defaults` property). etc.

natefoo:
I unresolved a couple resolved comments that I added comments to

lib/galaxy/tools/toolbox/base.py
natefoo: Ah ok. It'd be fairly difficult for folks to know whether it's safe to use `config.schema.defaults` in a given scenario then, though?

lib/galaxy/config/__init__.py
jdavcs: It's only called here: https://github.com/galaxyproject/galaxy/blob/163ba59bb189cbb3bc42c7f7404ee075a6da8aeb/lib/galaxy/config/__init__.py#L210 - and the code checks if the variable has been set. I don't foresee any other use cases for it, so I think `deprecated_dirs = None` is a more straightforward way to say "you ain't gonna need it, unless you know you need it". Whereas an empty dict implies the possibility of loading stuff into it. 
That said, if you think such a safeguard is a reasonable precaution, I'll make it a dict as per Nicola's suggesion.

lib/galaxy/tools/toolbox/base.py
jdavcs: Yes, indeed.. It's supposed to simplify, not to add a special case. I'll look into modifying the schema mock object to accommodate this. Leaving this open/unresolved for now.

lib/galaxy/config/__init__.py
natefoo: Here as well: https://github.com/galaxyproject/galaxy/pull/9371/files#diff-46f86ea0d82b84642e4218279ef61442R196 but I see that's covered by the code you linked.

lib/galaxy/config/__init__.py
jdavcs: The following are legit cases when this would happen: `config_dir`, `data_dir`, `managed_data_dir` - i.e., the properties that are set in the base class' `_set_config_base()` method **and** defined in the schema. In the previous setup they had to be defined *after* this method was called but *before* any path resolution, as they were used in paths. So there was a rigid sequence of method calls: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L224 (see lines 224-230). Such tight coupling, of course, prevents refactoring (and is bad for other reasons).

I think the setting of base config files is a special case and has nothing to do with automatically reading schema and values from kwargs ( here's my thoughts: https://github.com/galaxyproject/galaxy/commit/777379729198a893519ad58f167be761a4ea867d ). And so considering these legit cases when this can happen, I think we can treat it as normal behavior: create attribute unless it exists. 

Besides, I think, the only time that can bite someone, is when you add a new instance variable to the config object and use a name for it that's in the schema (or vise-versa). But that's a remote possibility; besides, it's easily debuggable, especially considering the `create_attr...` method is called so early: https://github.com/galaxyproject/galaxy/blob/163ba59bb189cbb3bc42c7f7404ee075a6da8aeb/lib/galaxy/config/__init__.py#L118 - anything conflicting with this would have to be created *earlier* than that call. In summary, I think treating it as expected behavior is reasonable.

lib/galaxy/config/__init__.py
jdavcs: Yep, that's the one I was referring to.

lib/galaxy/config/__init__.py
jdavcs: Scratch all the above. See latest commit - just check the key against 3 base config attrs and raise exception for anything else.

lib/galaxy/tools/toolbox/base.py
jdavcs: This, too, proved to be annoyingly simple. See last commit.

lib/galaxy/config/__init__.py
natefoo: Does it need an `else: log.something()` rather than silently passing now?

jdavcs: I don't think so: it's expected behavior and happens on every load:
```
if attribute doesn't exist:
    create attribute - as expected 
else:
    if exists and is NOT a base config:
        raise error
    else: (is a base config)
        do not create attribute - as expected
```

natefoo: Ok, thanks for the clarification

natefoo (approved):

jdavcs:
Thank you for the detailed review @natefoo!

bgruening:
Ah, sorry, missed that one. This is cool! Thanks @ic4f!

jdavcs:
@bgruening No worries, and thank you!

Issue #9369, opened 2020-02-12
Special characters in a citation


stephanflemming:
Special characters are not displayed properly. 

See https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/iuc/sina/sina/ for example.

`Jörg Peplies, Frank Oliver Glöckner`

is displayed as 

`Peplies, JÃ¶rg and GlÃ¶ckner, Frank Oliver`

ping @OlegZharkov

OlegZharkov:
Looks like backend problem https://usegalaxy.eu/api/tools/toolshed.g2.bx.psu.edu/repos/iuc/sina/sina/1.5.0+galaxy1/citations returns  ```'\\u00c3\\u00b6'``` which is UTF-8 encoded version of ```ö``` as hex representation (0xc3b6) .  I guess, can be possibly fixed on frontend with something like:
```
str = "\\u00c3\\u00b6".replaceAll("\\u00","")
b = {Library}.fromHexToBytearray(str)
b.decode('UTF-8')
```
But I guess, that would be quite hackish. We need to fix backend API

PR #9373, opened 2020-02-13
fix special characters in citation field

OlegZharkov:
Fix for https://github.com/galaxyproject/galaxy/issues/9369
A lot of python dark magic in here. To make it work, please delete cache in ```{galaxy}/database/citations/*```

### DIFF
--- a/lib/galaxy/managers/citations.py
+++ b/lib/galaxy/managers/citations.py
@@ -47,7 +47,7 @@ class DoiCache(object):
         doi_url = "https://doi.org/" + doi
         headers = {'Accept': 'text/bibliography; style=bibtex, application/x-bibtex'}
         req = requests.get(doi_url, headers=headers)
-        return req.text
+        return req.content.decode("utf-8")
 
     def get_bibtex(self, doi):
         createfunc = functools.partial(self._raw_get_bibtex, doi)
--- a/lib/galaxy/managers/citations.py
+++ b/lib/galaxy/managers/citations.py
@@ -47,7 +47,8 @@ class DoiCache(object):
         doi_url = "https://doi.org/" + doi
         headers = {'Accept': 'text/bibliography; style=bibtex, application/x-bibtex'}
         req = requests.get(doi_url, headers=headers)
-        return req.content.decode("utf-8")
+        req.encoding = req.apparent_encoding
+        return req.text
 
     def get_bibtex(self, doi):
         createfunc = functools.partial(self._raw_get_bibtex, doi)
--- a/lib/galaxy/managers/citations.py
+++ b/lib/galaxy/managers/citations.py
@@ -45,7 +45,7 @@ class DoiCache(object):
 
     def _raw_get_bibtex(self, doi):
         doi_url = "https://doi.org/" + doi
-        headers = {'Accept': 'text/bibliography; style=bibtex, application/x-bibtex'}
+        headers = {'Accept': 'application/x-bibtex'}
         req = requests.get(doi_url, headers=headers)
         req.encoding = req.apparent_encoding
         return req.text


lib/galaxy/managers/citations.py
```diff
@@ -47,7 +47,7 @@ def _raw_get_bibtex(self, doi):
         doi_url = "https://doi.org/" + doi
         headers = {'Accept': 'text/bibliography; style=bibtex, application/x-bibtex'}
         req = requests.get(doi_url, headers=headers)
-        return req.text
+        return req.content.decode("utf-8")
```
mvdbeek: that should be exactly the same as `req.text`:
```

In [1]: url = 'https://usegalaxy.org/api/version'

In [2]: import requests

In [3]: r = requests.get(url)

In [4]: r.text
Out[4]: '{"version_major": "20.01", "extra": {}}'

In [5]: r.content
Out[5]: b'{"version_major": "20.01", "extra": {}}'

In [6]: r.content.decode('utf-8')
Out[6]: '{"version_major": "20.01", "extra": {}}'
```

OlegZharkov: Try 
```
import requests
doi_url = "https://doi.org/10.1093/bioinformatics/bts252"
headers = {'Accept': 'text/bibliography; style=bibtex, application/x-bibtex'}
req = requests.get(doi_url, headers=headers)
print(req.content.decode("utf-8"))
```
and 
```
import requests
doi_url = "https://doi.org/10.1093/bioinformatics/bts252"
headers = {'Accept': 'text/bibliography; style=bibtex, application/x-bibtex'}
req = requests.get(doi_url, headers=headers)
print(req.text)
```

nsoranzo: It seems the issue is that `req.encoding` is 'ISO-8859-1', while it actually is 'utf-8'.

mvdbeek:
```suggestion
        req.encoding = req.apparent_encoding
        return req.text
```
might be a better approach here (https://2.python-requests.org/en/master/api/#requests.Response.apparent_encoding).

OlegZharkov: @nsoranzo 
Ok, should I change it to 
```
req = requests.get(doi_url, headers=headers)
req.encoding = req.apparent_encoding
```

OlegZharkov: Yeah, just found the same [here](https://stackoverflow.com/a/52615216) . ```apparent_encoding``` looks better :thumbsup:

lib/galaxy/managers/citations.py
```diff
@@ -47,6 +47,7 @@ def _raw_get_bibtex(self, doi):
         doi_url = "https://doi.org/" + doi
         headers = {'Accept': 'text/bibliography; style=bibtex, application/x-bibtex'}
```
nsoranzo:
```suggestion
        headers = {'Accept': 'application/x-bibtex'}
```
seems to work out of the box with `req.text`. Only difference is that the output is formatted with newlines and tabs (not sure that matters).

OlegZharkov: Looks like it is just fine this way, works out of the box. I got properly rendered umlauts on galaxy page. ```apparent_encoding``` solution kind of looks better to me. If in some unrealistic case,  they decide to change default style, we are prepared :satisfied: . But I don't really mind, I can force push the stuff with headers, so which one should I use?

nsoranzo: I'd probably do both. For reference, this is documented at https://citation.crosscite.org/docs.html

OlegZharkov:
@nsoranzo @mvdbeek Thanks for reviewing it!

nsoranzo (approved):

nsoranzo:
@galaxybot test this

mvdbeek (approved):

dannon:
Thank you!

OlegZharkov:
@dannon @mvdbeek @nsoranzo Can you close https://github.com/galaxyproject/galaxy/issues/9369 please? thanks!

mvdbeek:
I will backport this to 19.09, thanks for the fix!

PR #9374, opened 2020-02-14
Add dataset deserializer for 'datatype'

### LINKED ISSUE #4363: Implement API for datatype changes
xref https://github.com/galaxyproject/bioblend/issues/243

simonbray:
The purpose of this change (suggested by @nsoranzo) is to allow datatypes to be changed via the API/bioblend.

I see that being able to update a datatype for a running job might be an issue, so please let me know if another approach would be better.

Fixes #4363.


nsoranzo:
It would be good to have an API test for that, I think.

simonbray:
Added a test.

nsoranzo (approved):

lib/galaxy_test/api/test_datasets.py
```diff
@@ -136,3 +136,16 @@ def test_tag_change(self):
...
+        original_hda = self._get(
+            "histories/{history_id}/contents/{hda_id}".format(history_id=self.history_id, hda_id=hda_id)).json()
+        assert original_hda['extension'] == 'txt'
+        update_payload = {
+            'extension': 'dat'
+        }
+        updated_hda = self._put(
+            "histories/{history_id}/contents/{hda_id}".format(history_id=self.history_id, hda_id=hda_id),
+            update_payload).json()
+        assert updated_hda['extension'] == 'dat'
```
mvdbeek: I think this shows it does not work, it doesn't update the metadata (probably just sets the extension in the database), and dat is not a valid extension, so this must fail.
For an implementation you might want to look at https://github.com/mvdbeek/galaxy/blob/cd7a2c3418d6540ab3bb3177b0b0216c54ebff56/lib/galaxy/webapps/galaxy/controllers/dataset.py#L365.

A test should then also demonstrate that datatype specific metadata has been set.

mvdbeek (changes requested):

lib/galaxy_test/api/test_datasets.py
nsoranzo: Ops, that looked too easy! @simonbray You can probably move the implementation that Marius linked to a new `DatasetAssociationDeserializer.deserialize_extension()` .

simonbray:
@mvdbeek @nsoranzo, not sure if you were notified by my reopening the PR, but I had another go at this. Another review would be much appreciated.

simonbray:
Sorry, just added one more thing - updating a dataset which is busy should result in an error.

mvdbeek:
Nice, that looks better!

lib/galaxy/managers/datasets.py
```diff
@@ -684,6 +686,17 @@ def deserialize_metadatum(self, dataset_assoc, key, val, **context):
         # ...?
         return unwrapped_val
 
+    def deserialize_datatype(self, item, key, val, **context):
+        if item.datatype.allow_datatype_change and context['trans'].app.datatypes_registry.get_datatype_by_extension(val).allow_datatype_change:
+            if managers.hdas.HDAManager(self.app).ok_to_edit_metadata(item.dataset_id):
```
nsoranzo:
```suggestion
            if DatasetAssociationManager(self.app).ok_to_edit_metadata(item.dataset_id):
```

Or `ok_to_edit_metadata()` should actually be a method of the `DatasetInstance` class in `lib/galaxy/model/__init__.py`, I think.

lib/galaxy/managers/datasets.py
```diff
@@ -684,6 +686,17 @@ def deserialize_metadatum(self, dataset_assoc, key, val, **context):
         # ...?
         return unwrapped_val
 
+    def deserialize_datatype(self, item, key, val, **context):
+        if item.datatype.allow_datatype_change and context['trans'].app.datatypes_registry.get_datatype_by_extension(val).allow_datatype_change:
```
nsoranzo:
```suggestion
        if item.datatype.allow_datatype_change and self.app.datatypes_registry.get_datatype_by_extension(val).allow_datatype_change:
```

lib/galaxy/managers/datasets.py
```diff
@@ -684,6 +686,17 @@ def deserialize_metadatum(self, dataset_assoc, key, val, **context):
...
+            if managers.hdas.HDAManager(self.app).ok_to_edit_metadata(item.dataset_id):
+                context['trans'].app.datatypes_registry.change_datatype(item, val)
```
nsoranzo:
```suggestion
                item.change_datatype(val)
```

lib/galaxy/managers/datasets.py
```diff
@@ -684,6 +686,17 @@ def deserialize_metadatum(self, dataset_assoc, key, val, **context):
...
+                context['trans'].sa_session.flush()
+                context['trans'].app.datatypes_registry.set_external_metadata_tool.tool_action.execute(context['trans'].app.datatypes_registry.set_external_metadata_tool, context['trans'], incoming={'input1': item}, overwrite=False)  # overwrite is False as per existing behavior
+                return item.datatype
+        context['trans'].response.status = 500
+        context['trans'].error_message = 'Datatype could not be updated'
+        return False
```
nsoranzo: Should raise an exception instead.

lib/galaxy/managers/datasets.py
```diff
@@ -684,6 +685,17 @@ def deserialize_metadatum(self, dataset_assoc, key, val, **context):
...
 
+    def deserialize_datatype(self, item, key, val, **context):
+        if item.datatype.allow_datatype_change and self.app.datatypes_registry.get_datatype_by_extension(val).allow_datatype_change:
+            if DatasetAssociationManager(self.app).ok_to_edit_metadata(item.dataset_id):
+                item.change_datatype(val)
+                context['trans'].sa_session.flush()
+                self.app.datatypes_registry.set_external_metadata_tool.tool_action.execute(self.app.datatypes_registry.set_external_metadata_tool, context['trans'], incoming={'input1': item}, overwrite=False)  # overwrite is False as per existing behavior
+                return item.datatype
+        context['trans'].response.status = 500
+        context['trans'].error_message = 'Datatype could not be updated.'
```
nsoranzo:
```suggestion
```

lib/galaxy/managers/datasets.py
```diff
@@ -684,6 +685,17 @@ def deserialize_metadatum(self, dataset_assoc, key, val, **context):
         # ...?
         return unwrapped_val
 
+    def deserialize_datatype(self, item, key, val, **context):
+        if item.datatype.allow_datatype_change and self.app.datatypes_registry.get_datatype_by_extension(val).allow_datatype_change:
+            if DatasetAssociationManager(self.app).ok_to_edit_metadata(item.dataset_id):
+                item.change_datatype(val)
```
nsoranzo: To return more specific error messages, I would restructure this as:
```python
        if not item.datatype.allow_datatype_change:
        	raise SomeException("message")
        if not self.app.datatypes_registry.get_datatype_by_extension(val).allow_datatype_change:
        	raise SomeException("message")
        if not DatasetAssociationManager(self.app).ok_to_edit_metadata(item.dataset_id):
        	raise AnotherException("message")
        item.change_datatype(val)
        ...
```

lib/galaxy/managers/datasets.py
```diff
@@ -686,15 +686,16 @@ def deserialize_metadatum(self, dataset_assoc, key, val, **context):
...
-                context['trans'].sa_session.flush()
-                self.app.datatypes_registry.set_external_metadata_tool.tool_action.execute(self.app.datatypes_registry.set_external_metadata_tool, context['trans'], incoming={'input1': item}, overwrite=False)  # overwrite is False as per existing behavior
-                return item.datatype
-        context['trans'].response.status = 500
-        context['trans'].error_message = 'Datatype could not be updated.'
-        raise Exception('Datatype could not be updated.')
+        if not item.datatype.allow_datatype_change:
+            raise Exception("The current datatype does not allow datatype changes.")
+        if not self.app.datatypes_registry.get_datatype_by_extension(val).allow_datatype_change:
+            raise Exception("Either the target datatype does not exist, or it does not allow datatype changes.")
```
nsoranzo:
```suggestion
		if not self.app.datatypes_registry.get_datatype_by_extension(val):
			raise RequestParameterInvalidException("The target datatype does not exist.")
        if not self.app.datatypes_registry.get_datatype_by_extension(val).allow_datatype_change:
            raise RequestParameterInvalidException("The target datatype does not allow datatype changes.")
```

lib/galaxy/managers/datasets.py
```diff
@@ -686,15 +686,16 @@ def deserialize_metadatum(self, dataset_assoc, key, val, **context):
...
-            if DatasetAssociationManager(self.app).ok_to_edit_metadata(item.dataset_id):
-                item.change_datatype(val)
-                context['trans'].sa_session.flush()
-                self.app.datatypes_registry.set_external_metadata_tool.tool_action.execute(self.app.datatypes_registry.set_external_metadata_tool, context['trans'], incoming={'input1': item}, overwrite=False)  # overwrite is False as per existing behavior
-                return item.datatype
-        context['trans'].response.status = 500
-        context['trans'].error_message = 'Datatype could not be updated.'
-        raise Exception('Datatype could not be updated.')
+        if not item.datatype.allow_datatype_change:
+            raise Exception("The current datatype does not allow datatype changes.")
```
nsoranzo: You can use a more specific exception (need a `from galaxy.exceptions import RequestParameterInvalidException` )

```suggestion
            raise RequestParameterInvalidException("The current datatype does not allow datatype changes.")
```

lib/galaxy/managers/datasets.py
```diff
@@ -684,6 +685,20 @@ def deserialize_metadatum(self, dataset_assoc, key, val, **context):
...
 
+    def deserialize_datatype(self, item, key, val, **context):
+        if not item.datatype.allow_datatype_change:
+            raise exceptions.RequestParameterInvalidException("The current datatype does not allow datatype changes.")
+        if not self.app.datatypes_registry.get_datatype_by_extension(val):
+            raise exceptions.RequestParameterInvalidException("The target datatype does not exist.")
+        if not self.app.datatypes_registry.get_datatype_by_extension(val).allow_datatype_change:
+            raise exceptions.RequestParameterInvalidException("The target datatype does not allow datatype changes.")
+        if not DatasetAssociationManager(self.app).ok_to_edit_metadata(item.dataset_id):
+            raise exceptions.RequestParameterInvalidException("Dataset metadata could not be updated, probably because it is not in an 'ok' state.")
```
nsoranzo: That's not an invalid parameter in the request, so I would raise a normal exception (I don't think we have any specific Galaxy exception), which should also return a 500 HTTP error code (this is in fact more a server than user error).

```suggestion
            raise Exception("Dataset metadata could not be updated, probably because it is not in an 'ok' state.")
```

lib/galaxy_test/api/test_datasets.py
```diff
@@ -134,3 +135,30 @@ def test_tag_change(self):
...
+            "histories/{history_id}/contents/{hda_id}".format(history_id=self.history_id, hda_id=hda_id),
+            {'datatype': 'tabular'}).json()
+        assert successful_updated_hda['extension'] == 'tabular'
+        assert successful_updated_hda['data_type'] == 'galaxy.datatypes.tabular.Tabular'
+        assert len(successful_updated_hda['visualizations']) > 20
+
+        invalidly_updated_hda = self._put(  # try updating with invalid datatype
+            "histories/{history_id}/contents/{hda_id}".format(history_id=self.history_id, hda_id=hda_id),
+            {'datatype': 'invalid'})
+        self._assert_status_code_is(invalidly_updated_hda, 400)
```
nsoranzo:
```suggestion
        invalidly_updated_hda_response = self._put(  # try updating with invalid datatype
            "histories/{history_id}/contents/{hda_id}".format(history_id=self.history_id, hda_id=hda_id),
            {'datatype': 'invalid'})
        self._assert_status_code_is(invalidly_updated_hda_response, 400)
```

lib/galaxy_test/api/test_datasets.py
```diff
@@ -134,3 +135,30 @@ def test_tag_change(self):
...
+        assert original_hda['extension'] == 'txt'
+        assert original_hda['data_type'] == 'galaxy.datatypes.data.Text'
+        assert len(original_hda['visualizations']) < 10
+
+        update_while_incomplete = self._put(  # try updating datatype before upload is complete
+            "histories/{history_id}/contents/{hda_id}".format(history_id=self.history_id, hda_id=hda_id),
+            {'datatype': 'tabular'}).status_code
+        assert update_while_incomplete == 400
+
+        time.sleep(10)  # now wait for upload to complete
```
nsoranzo:
```suggestion
        self.dataset_populator.wait_for_history(self.history_id)  # now wait for upload to complete
```

lib/galaxy_test/api/test_datasets.py
```diff
@@ -1,4 +1,5 @@
 import textwrap
+import time
```
nsoranzo:
```suggestion
```

lib/galaxy_test/api/test_datasets.py
```diff
@@ -134,3 +135,30 @@ def test_tag_change(self):
...
+        original_hda = self._get(
+            "histories/{history_id}/contents/{hda_id}".format(history_id=self.history_id, hda_id=hda_id)).json()
+        assert original_hda['extension'] == 'txt'
+        assert original_hda['data_type'] == 'galaxy.datatypes.data.Text'
+        assert len(original_hda['visualizations']) < 10
+
+        update_while_incomplete = self._put(  # try updating datatype before upload is complete
+            "histories/{history_id}/contents/{hda_id}".format(history_id=self.history_id, hda_id=hda_id),
+            {'datatype': 'tabular'}).status_code
+        assert update_while_incomplete == 400
```
nsoranzo:
```suggestion
        update_while_incomplete_response = self._put(  # try updating datatype before upload is complete
            "histories/{history_id}/contents/{hda_id}".format(history_id=self.history_id, hda_id=hda_id),
            {'datatype': 'tabular'})
        self._assert_status_code_is(update_while_incomplete_response, 500)
```

lib/galaxy_test/api/test_datasets.py
```diff
@@ -134,3 +135,30 @@ def test_tag_change(self):
...
         assert 'name:tag_d' in updated_hda['tags']
         assert 'name:tag_e' in updated_hda['tags']
+
+    def test_update_datatype(self):
+        hda_id = self.dataset_populator.new_dataset(self.history_id)['id']
+        original_hda = self._get(
+            "histories/{history_id}/contents/{hda_id}".format(history_id=self.history_id, hda_id=hda_id)).json()
+        assert original_hda['extension'] == 'txt'
+        assert original_hda['data_type'] == 'galaxy.datatypes.data.Text'
+        assert len(original_hda['visualizations']) < 10
```
nsoranzo: Maybe checking that a tabular-specific visualization is not present in `original_hda['visualizations']` is more future-proof? Same (reversed) below.

lib/galaxy_test/api/test_datasets.py
```diff
@@ -142,23 +141,23 @@ def test_update_datatype(self):
...
-        successful_updated_hda = self._put(
+        successful_updated_hda_response = self._put(
             "histories/{history_id}/contents/{hda_id}".format(history_id=self.history_id, hda_id=hda_id),
             {'datatype': 'tabular'}).json()
-        assert successful_updated_hda['extension'] == 'tabular'
-        assert successful_updated_hda['data_type'] == 'galaxy.datatypes.tabular.Tabular'
-        assert len(successful_updated_hda['visualizations']) > 20
+        assert successful_updated_hda_response['extension'] == 'tabular'
+        assert successful_updated_hda_response['data_type'] == 'galaxy.datatypes.tabular.Tabular'
+        'scatterplot' not in successful_updated_hda_response['visualizations']
```
nsoranzo: Shouldn't this be:
```suggestion
        'scatterplot' in successful_updated_hda_response['visualizations']
```

simonbray: Yes, also it's not actually an assertion... just fixed it, thanks.

lib/galaxy/managers/datasets.py
```diff
@@ -684,6 +683,20 @@ def deserialize_metadatum(self, dataset_assoc, key, val, **context):
...
 
+    def deserialize_datatype(self, item, key, val, **context):
+        if not item.datatype.allow_datatype_change:
+            raise exceptions.RequestParameterInvalidException("The current datatype does not allow datatype changes.")
+        if not self.app.datatypes_registry.get_datatype_by_extension(val):
+            raise exceptions.RequestParameterInvalidException("The target datatype does not exist.")
+        if not self.app.datatypes_registry.get_datatype_by_extension(val).allow_datatype_change:
+            raise exceptions.RequestParameterInvalidException("The target datatype does not allow datatype changes.")
+        if not DatasetAssociationManager(self.app).ok_to_edit_metadata(item.dataset_id):
+            raise Exception("Dataset metadata could not be updated, probably because it is not in an 'ok' state.")
```
mvdbeek:
```suggestion
            raise RequestParameterInvalidException("Dataset metadata could not be updated because it is used as input or output of a running job.")
```
I would make this a `RequestParameterInvalidException`, this shouldn't be a 500.

mvdbeek (approved):
Awesome, that looks great to me!

lib/galaxy/managers/datasets.py
simonbray: `used as input or output` is not clear for me - I only encountered this error so far when a dataset is uploading (like in the test) or a previous datatype change is not complete

mvdbeek: I just edited the message, `used as input or output of a running job.` which is what you're describing, right ?

lib/galaxy_test/api/test_datasets.py
```diff
@@ -134,3 +134,30 @@ def test_tag_change(self):
...
+        original_hda = self._get(
+            "histories/{history_id}/contents/{hda_id}".format(history_id=self.history_id, hda_id=hda_id)).json()
+        assert original_hda['extension'] == 'txt'
+        assert original_hda['data_type'] == 'galaxy.datatypes.data.Text'
+        assert 'scatterplot' not in [viz['name'] for viz in original_hda['visualizations']]
+
+        update_while_incomplete_response = self._put(  # try updating datatype before upload is complete
+            "histories/{history_id}/contents/{hda_id}".format(history_id=self.history_id, hda_id=hda_id),
+            {'datatype': 'tabular'})
+        self._assert_status_code_is(update_while_incomplete_response, 500)
```
mvdbeek, line 149: I think that could lead to unstable API tests as performance increases. We've got a a tool that can sleep before running, I'd use that for this test:
```
        inputs = {
            'input1': {'src': 'hda', 'id': hda_id},
            'sleep_time': 10,
        }
        run_response = self.dataset_populator.run_tool(
            "cat_data_and_sleep",
            inputs,
            history_id,
            assert_ok=False,
        )
        queued_id = run_response["outputs"][0]["id"]
```

lib/galaxy/managers/datasets.py
```diff
@@ -684,6 +683,20 @@ def deserialize_metadatum(self, dataset_assoc, key, val, **context):
...
 
+    def deserialize_datatype(self, item, key, val, **context):
+        if not item.datatype.allow_datatype_change:
+            raise exceptions.RequestParameterInvalidException("The current datatype does not allow datatype changes.")
+        if not self.app.datatypes_registry.get_datatype_by_extension(val):
+            raise exceptions.RequestParameterInvalidException("The target datatype does not exist.")
+        if not self.app.datatypes_registry.get_datatype_by_extension(val).allow_datatype_change:
+            raise exceptions.RequestParameterInvalidException("The target datatype does not allow datatype changes.")
+        if not DatasetAssociationManager(self.app).ok_to_edit_metadata(item.dataset_id):
+            raise Exception("Dataset metadata could not be updated because it is used as input or output of a running job.")
```
mvdbeek, line 694:
```suggestion
            raise RequestParameterInvalidException("Dataset metadata could not be updated because it is used as input or output of a running job.")
```
This isn't a server error, I think this falls into the invalid request parameter domain

mvdbeek:
I'm gonna merge this so we can branch, but I think we should re-work the test and change the exception to RequestParameterInvalidException

simonbray:
Okay, I'll make a new PR for that.

Thanks a lot for the help with this, this feature will be really useful.

lib/galaxy/managers/datasets.py
nsoranzo: I previously suggested the inverse :laughing: . I suppose it's border-line, Galaxy could put the request in a queue instead of returning an error. Also, the dataset may be in use by another user with whom it is shared.

mvdbeek: I agree about the queue, but I think this is nothing admins should see as errors in the logs, which would happen with bare Exceptions, right ?

PR #9375, opened 2020-02-14
[20.01] Backport edge fixes for 20.01

dannon:
Flattened, and dropped the dependency updates
xref #9289


dannon:
Huh.  Looks like `gh pr` will always push the branch in question to the destination fork instead of the configured origin (in this case, my fork).  Good to know, and I'll delete it after.

mvdbeek:
Thanks @dannon, I've deleted the branch.

PR #8674, opened 2019-09-18
Merge remote-tracking branch 'upstream/release_19.09' into dev

dannon:
Resolved conflict at https://github.com/galaxyproject/galaxy/pull/8674/files#diff-46f86ea0d82b84642e4218279ef61442L237, correct resolution was to drop those as was done in 19.09, right @ic4f, @natefoo?

PR #9376, opened 2020-02-14
migrated_tools_conf should be in mutable_config_dir

jdavcs:
This used to be in `config_dir` by default, but should be in `mutable_confiig_dir`.

1. Keep it consistent with ansible-galaxy. (as per discussion with @natefoo)
2. Also, correct description of integrated_tool_panel_config (unrelated minor fix)

Context: 
* current ansible-galaxy: https://github.com/galaxyproject/ansible-galaxy/blob/master/defaults/main.yml#L192
* galaxy's `config/__init__.py` as it used to be prior to #8610: https://github.com/galaxyproject/galaxy/pull/8674/commits/be5aafd34b1f400704ef863684c9b742e1ca7427#diff-46f86ea0d82b84642e4218279ef61442L209

### DIFF
--- a/doc/source/admin/galaxy_options.rst
+++ b/doc/source/admin/galaxy_options.rst
@@ -303,7 +303,7 @@
     tool shed upon a new release, they will be added to this tool
     config file.
     The value of this option will be resolved with respect to
-    <config_dir>.
+    <mutable_config_dir>.
 :Default: ``migrated_tools_conf.xml``
 :Type: str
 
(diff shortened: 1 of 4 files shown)


jdavcs:
~~Failed unit test is unrelated.~~ All pass.

nsoranzo:
- Given #8610 targeted 19.09, shouldn't this also target that release?
- You need to also execute `make config-rebuild` and commit the changes.

jdavcs:
@nsoranzo do you mean target *only* that release, or target dev + 19.09 (and then 20.01 too?)

mvdbeek:
Just 19.09 is fine, we can merge forward then.

PR #9379, opened 2020-02-17
Remove more duplication from TS, galaxy config

jdavcs:
The TS part of this is based on #9371 where part of config processing was moved from GalaxyAppConfiguration into BaseAppConfiguration. Since that base class is the parent of ToolShedAppConfiguration, much of TS config code becomes redundant. In particular, TS config now uses the same schema defaults/kwargs loading process as galaxy's config.
The galaxy config changes are minor cleanup.

Related to issue #8493.

EDIT: the following description is from a follow-up PR, but now that a few commits have been moved around, it really belongs in this PR:
Main theme: factor out common configuration settings into a mixin class. Much of TS's config code is duplication of code in Galaxy's main config object (there's even a TODO in the code: https://github.com/galaxyproject/galaxy/blob/dev/lib/tool_shed/webapp/config.py#L191 )

Argument for not moving common code into BaseAppConfiguration:

* `BaseAppConfiguration` contains code that is used by the configuration code of both apps, Galaxy and TS - i.e., base config paths, schema loading, etc.
* `CommonConfigurationMixin` will contain code that happens to be present in both
Galaxy and TS, but deals with individual configuration properties.

I didn't move all common code into the mixin: there are many settings that are the same for both Galaxy and TS; however, factoring out individual lines, that are identical, would make the configuration of both apps much less readable. I think it's more important to have the individual config steps in one place (i.e., one method) for each app, even if some properties happen to have the same values.

The rest is mostly minor cleanup.

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -320,6 +320,15 @@ class CommonConfigurationMixin(object):
         """Determine if the provided user is listed in `admin_users`."""
         return user is not None and user.email in self.admin_users_list
 
+    @property
+    def sentry_dsn_public(self):
+        """
+        Sentry URL with private key removed for use in client side scripts,
+        sentry server will need to be configured to accept events
+        """
+        if self.sentry_dsn:
+            return re.sub(r"^([^:/?#]+:)?//(\w+):(\w+)", r"\1//\2", self.sentry_dsn)
+
 
 class GalaxyAppConfiguration(BaseAppConfiguration, CommonConfigurationMixin):
     deprecated_options = ('database_file', 'track_jobs_in_database')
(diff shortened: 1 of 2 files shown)


dannon:
I just tried this in a fresh, unconfigured clone and I see:
```
No Galaxy config file found, running from current working directory: /Users/yoplait/work/galaxy-clean
  File "lib/tool_shed/webapp/buildapp.py", line 64, in app_factory
    app = UniverseApplication(global_conf=global_conf, **kwargs)
  File "lib/tool_shed/webapp/app.py", line 30, in __init__
    self.config = config.Configuration(**kwd)
  File "lib/tool_shed/webapp/config.py", line 83, in __init__
    self.admin_users_list = [u.strip() for u in self.admin_users.split(',') if u]
AttributeError: 'NoneType' object has no attribute 'split'
```

ic4f:
Wow, that's surprising. Thanks!

dannon:
@ic4f (I'm sure it's the swap from `''` as the default to None, but it makes me wonder what else was changed and maybe hasn't been exercised yet)

ic4f:
@dannon Yep. And it's actually taken care of in the subsequent PR: https://github.com/galaxyproject/galaxy/blob/6d6427296c9c8da9305d4e1d279b2dc66c5e76c0/lib/galaxy/config/__init__.py#L311
`admin_users` is a property, and has a setter that checks if that value has been set, and initialized it to `[]` if not. In the subsequent PR that bit (as many others) has been moved into a mixin - which is used by the TS config. I'll reorganize those changes so this doesn't happen in this PR.

dannon:
@ic4f Awesome, sounds good, thanks!

ic4f:
WIP. Will rebase when #9528 is merged.

ic4f:
Thank you @dannon !

PR #9383, opened 2020-02-17
Fix workflow run missing tool error message

guerler:


jmchilton:
There is an open PR - https://github.com/galaxyproject/galaxy/pull/9147 - it seems to have conflicted now but I'll rebase today.

guerler:
@jmchilton an alternative would be to close this PR if #9147 has already fixes for the issues mentioned here. Thanks for the review.

jmchilton:
One of these changes was already in https://github.com/galaxyproject/galaxy/pull/9147 - I've ported the other one over and rebased against latest workflow client changes.

PR #9384, opened 2020-02-17
[20.01] Hyphy-vision fix and updates.

dannon:
This upgrades hyphy-vision from a version `2.5 alpha` to the current official release `2.6.3`.  Includes various fixes required to get it all building again.

The previous version of this plugin expected several libraries to be available, exposed on the window, which are no longer provided by galaxy internals (and can't ever be as stable versions provided by the framework).  This new version of the plugin explicitly includes and provides those libraries to the plugin itself as a part of the build process.

I also dropped build artifacts from the repo, and added it to the autobuild plugin list.


mvdbeek:
Looks like there is an interactive prompt stopping the first startup tests:
```
$ webpack -p
One CLI for webpack must be installed. These are recommended choices, delivered as separate packages:
 - webpack-cli (https://github.com/webpack/webpack-cli)
   The original webpack full-featured CLI.
We will use "yarn" to install the CLI via "yarn add -D".
Do you want to install 'webpack-cli' (yes/no): 
Too long with no output (exceeded 10m0s): context deadline exceeded
```

dannon:
@mvdbeek Weird, let me see what's up here.  `webpack-cli` should be installed and available there

mvdbeek (approved):

PR #9385, opened 2020-02-18
[19.09] Prevent tool execution getting stuck in history when a chromInfo data…

mvdbeek:
…set is in paused state

If a chromInfo dataset is paused (or in other non-terminal states)
execution will not start until the dataset is ready, which may be never.
So instead we prefer datasets in OK state and ignore those that
will never be OK. In theory that may fix some odd issues where
jobs never start running.


mvdbeek:
We should also put this behavior under a profile version and have an explicit means to request a chromInfo file ... which might be through the dbkeys tool data table ?

PR #9386, opened 2020-02-18
CSS changes for a modern look and feel 

bedroesb:
I took away multiple borders to have a more modern feeling including buttons, tables and info boxes:

### Before:

Screenshot from 2020-02-18 11-43-48

Screenshot from 2020-02-18 11-44-54

### After:

Screenshot from 2020-02-18 11-44-00

Screenshot from 2020-02-18 11-45-09

### DIFF
--- a/client/galaxy/style/scss/base.scss
+++ b/client/galaxy/style/scss/base.scss
@@ -147,13 +147,11 @@ body {
     @extend .unified-panel;
     left: 0px;
     width: $panel-width;
-    border-right: $layout-border;
 }
 #right {
     @extend .unified-panel;
     right: 0px;
     width: $panel-width;
-    border-left: $layout-border;
 }
 #right > .unified-panel-footer {
     .drag {
@@ -250,7 +248,6 @@ body {
     height: $panel_footer_height;
     line-height: $panel_footer_height;
     width: 100%;
-    border-top: $layout-border;
     background: $panel-footer-bg-color;
     color: $panel-header-text-color;
     a {
@@ -845,7 +842,6 @@ table.colored tr.header {
     background: $table-heading-bg;
     background-repeat: repeat-x;
     background-position: top;
-    border-bottom: solid $brand-secondary 1px;
     font-weight: bold;
 }
 
--- a/client/galaxy/style/scss/base.scss
+++ b/client/galaxy/style/scss/base.scss
@@ -402,7 +402,7 @@ body {
     .navbar-brand {
         @extend .position-absolute;
         @extend .font-weight-bold;
-        font-family: verdana;
+        font-family: sans-serif;
         font-size: 1.25rem;
         left: 0.8rem;
         line-height: 0rem;
@@ -845,7 +845,7 @@ table.colored tr.header {
     background: $table-heading-bg;
     background-repeat: repeat-x;
     background-position: top;
-    border-bottom: solid $table-border-color 1px;
+    border-bottom: solid $brand-secondary 1px;
     font-weight: bold;
 }
 
--- a/client/galaxy/style/scss/base.scss
+++ b/client/galaxy/style/scss/base.scss
@@ -402,7 +402,7 @@ body {
     .navbar-brand {
         @extend .position-absolute;
         @extend .font-weight-bold;
-        font-family: sans-serif;
+        font-family: Verdana, sans-serif;
         font-size: 1.25rem;
         left: 0.8rem;
         line-height: 0rem;
--- a/client/galaxy/style/scss/overrides.scss
+++ b/client/galaxy/style/scss/overrides.scss
@@ -1,5 +1,5 @@
 .navbar-fixed-top .navbar-brand {
-    font-family: Verdana;
+    font-family: sans-serif;
     font-weight: bold;
     font-size: 160%;
 }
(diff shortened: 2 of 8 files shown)


dannon:
@bedroesb Hey, thanks for the PR.  Sam and I are chatting about this one, but I just wanted to let you know we saw it and you weren't throwing code into the void :)

client/galaxy/style/scss/base.scss
```diff
@@ -402,7 +402,7 @@ body {
     .navbar-brand {
         @extend .position-absolute;
         @extend .font-weight-bold;
-        font-family: verdana;
+        font-family: sans-serif;
```
dannon:
```suggestion
        font-family: Verdana, sans-serif;
```

client/galaxy/style/scss/overrides.scss
```diff
@@ -1,5 +1,5 @@
 .navbar-fixed-top .navbar-brand {
-    font-family: Verdana;
+    font-family: sans-serif;
```
dannon:
```suggestion
    font-family: Verdana, sans-serif;
```

dannon (commented):
Sam's going to add a bit more commentary on the rest, but I wanted to mention the font tweak.  Instead of just dropping Verdana for the default sans-serif font, we should just make it fail through to sans-serif on systems where Verdana isn't available.

bedroesb:
@dannon Thank you for your reply! Reason why I took away Verdana is because the font that is now given on our desktops here is not really fitting with the other fonts being used:
Screenshot from 2020-02-20 10-03-13

dannon:
@bedroesb Yep, no worries -- I noticed that in your screenshot and figured that's what you were trying to adjust.  The line I suggested as a modification should handle your concern while still using Verdana proper when it's available.

client/galaxy/style/scss/theme/blue.scss
```diff
@@ -19,7 +19,7 @@ $brand-masthead-text-hover: $brand-toggle;
 $text-color: $brand-dark;
 $text-muted: lighten($text-color, 10%);
 $text-light: lighten($text-color, 20%);
-$border-color: lighten($text-color, 60%);
+$border-color: $brand-secondary;
```
guerler: Thank you for working on this. I think removing the borders from the tables is a great idea but I would keep the borders for buttons. I think it improves their contrast. Either way remaining borders should be derived from text colors such that secondary `brand-colors` can be picked at random without affecting the contrast and legibility of texts and their borders. There is more information on this available at #8388.

client/galaxy/style/scss/theme/blue.scss
```diff
@@ -145,7 +145,7 @@ $form-border: darken($form-heading-bg, 20%);
 // Table
 $table-heading-bg: $panel-bg-color;
 $table-bg-accent: darken($panel-bg-color, 5%);
-$table-border-color: $border-color;
+$table-border-color: rgba(0, 0, 0, 0);
```
guerler: See comment above. Border colors should be either transparent or derived from current border-colors which are derived from text-colors.

client/galaxy/style/scss/ui.scss
```diff
@@ -191,6 +191,7 @@ $ui-margin-horizontal-large: $margin-v * 2;
         @extend .card-header;
         @extend .rounded;
         background: $portlet-bg-color;
+        border-bottom: 1px solid rgba(0, 0, 0, 0);
```
guerler: See comment above.

client/galaxy/style/scss/ui/search-input.scss
```diff
@@ -29,10 +29,10 @@
         font-size: $font-size-base;
         line-height: $line-height-base;
         color: $text-color;
-        border: $border-default;
+        border: none;
         @include border-radius($border-radius-extralarge);
         max-width: auto;
-        background: $white;
+        background: $brand-secondary;
```
guerler: This background is too dark, it makes the search field look like a button, also all other text fields have bright colors given the current blue.css theme configuration. For consistency I would restore the previous background. Thanks again for working on this and thank you for your contribution.

client/galaxy/style/scss/theme/blue.scss
bedroesb: In my humble opinion taking away the borders gives it more an app/software feeling instead of a bootstrap website feeling, and more modern. But this is off course a personal opinion ;)

client/galaxy/style/scss/ui/search-input.scss
bedroesb: This background was chosen so it did not need a border, and to stay within the same color palette that is now being used. But I agree, it is now more like a button. I had a feeling there was on purpose a distinction made between the two search fields, the ones that are very rounded and the more rectangular ones.

client/galaxy/style/scss/theme/blue.scss
bedroesb: Thank you, I will change this.

bedroesb:
I would like to compliment the latest changes that were done on webdesign side, I think there are some huge improvements made lately.

guerler (approved):

dannon:
This looks good to me now, thanks!

PR #9391, opened 2020-02-18
[20.01] Fix for qp method used in GEDAs.

blankenberg:
We will always force casting object to string before doing quote_plus.


mvdbeek (approved):

mvdbeek:
Thanks @blankenberg !

PR #9400, opened 2020-02-19
[20.01] Prevent anonymous users from being able to request email notification on job completion

dannon:
Still not exactly sure how this happened other than maybe someone anonymously submitting a job via the API, but there was one occurrence of this on main that subsequently caused problems with the job finish method (repeatedly).

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/tools.py
+++ b/lib/galaxy/webapps/galaxy/api/tools.py
@@ -551,9 +551,14 @@ class ToolsController(BaseAPIController, UsesVisualizationMixin):
         for job in vars.get('jobs', []):
             rval['jobs'].append(self.encode_all_ids(trans, job.to_dict(view='collection'), recursive=True))
             if inputs.get('send_email_notification', False):
-                job_email_action = trans.model.PostJobAction('EmailAction')
-                job.add_post_job_action(job_email_action)
-                new_pja_flush = True
+                # Unless an anonymous user is invoking this via the API it
+                # should never be an option, but check and enforce that here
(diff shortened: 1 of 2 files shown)


lib/galaxy/webapps/galaxy/api/tools.py
```diff
@@ -551,9 +551,14 @@ def _create(self, trans, payload, **kwd):
         for job in vars.get('jobs', []):
             rval['jobs'].append(self.encode_all_ids(trans, job.to_dict(view='collection'), recursive=True))
             if inputs.get('send_email_notification', False):
-                job_email_action = trans.model.PostJobAction('EmailAction')
-                job.add_post_job_action(job_email_action)
-                new_pja_flush = True
+                # Unless an anonymous user is invoking this via the API it
+                # should never be an option, but check and enforce that here
+                if trans.user is None:
+                    raise exceptions.ToolExecutionError("Anonymously run jobs cannot send an email notification.")
```
mvdbeek, line 557: My gut feeling is that silently ignoring this in the action might be better. If I had to choose between having to resubmit and not getting an email I'd choo
se the latter.

dannon: I was torn between this and malformed request.  The only way this should happen is via api submission anonymously, and I think it's probably most appropriate to inform instead of silently no-op.  (In the current client, it shouldn't be possible for someone to get here by using the provided interface, so I don't feel like this should have this 'slack', I guess?)

mvdbeek (approved):

PR #8109, opened 2019-06-06
new data types for OpenMS

bernt-matthias:
Needed here https://github.com/galaxyproteomics/tools-galaxyp/issues/266

- [ ] analysisXML
- [ ] cachedMzML
  - consists of multple files. 
- [x] `dta`, `dta2d`, `edta` 
  - [x] `dta`: 2 column space separated float/int?  https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Documentation/nightly/html/a09160.html
  - [x] `edta` in the examples I found tabular data with inconsistent headers `RT	mz	Int	charge	Meta2`, `RT	MZ	INT	CHARGE	rt1	mz1	int1	charge1	Rt2	Mz2	Int2	Charge2	rT3	mZ3	inT3	chargE3`
    - https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Documentation/nightly/html/a09164.html#ac29c69c6f4a4441f625da4eecfe033f9
  - [x] `dta2d` 3 column float/int https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Documentation/nightly/html/a09156.html
- [ ] fid seems to be a binary file + some text file
  - test data would be here https://github.com/OpenMS/OpenMS/tree/develop/src/tests/class_tests/openms/data/XMassFile_test (fid + acqus)
- [x] kroenik: https://abibuilder.informatik.uni-tuebingen.de/archive/openms/Documentation/nightly/html/a09372.html#details
- [x] `mascotxml`
- [x] ms (Internal format of sirius https://bio.informatik.uni-jena.de/repository/dist-release-local/de/unijena/bioinf/ms/sirius/4.0.1/sirius-4.0.1-manual.pdf)  -> add sirius.ms as subclass
- [x] mrm -> subclass https://github.com/galaxyproject/galaxy/pull/8109/commits/f756b718e294768faa0d1149a28097e4416a4c1b
- [x] `mztab`
  - [x] `mzTab` (v1.x designed for proteomics) https://github.com/HUPO-PSI/mzTab/blob/master/specification_document-releases/1_0-Proteomics-Release/mzTab_format_specification.pdf
  - [x] `mzTab-M` (= v2.x mzTab) https://github.com/HUPO-PSI/mzTab/blob/master/specification_document-releases/2_0-Metabolomics-Release/mzTab_format_specification_2_0-M_release.adoc#relationship-to-mztab-1.0
- [x] novor -> just map to txt
[...]

PR #9405, opened 2020-02-20
sqlite sniffer: check table names for equality

bernt-matthias:
Not sure if the derived classes always give the full list of tables. If so a check for equality (instead for subset) this would be better. 

Unfortunately not all have tests.

Otherwise I would add a parameter that can control if the check should be done for equality

### DIFF
--- a/lib/galaxy/datatypes/binary.py
+++ b/lib/galaxy/datatypes/binary.py
@@ -1403,10 +1403,7 @@ class SQlite(Binary):
             tables_query = "SELECT name FROM sqlite_master WHERE type='table' ORDER BY name"
             result = c.execute(tables_query).fetchall()
             result = [_[0] for _ in result]
-            for table_name in table_names:
-                if table_name not in result:
-                    return False
-            return True
+            return set(result) == set(table_names)
         except Exception as e:
             log.warning('%s, sniff Exception: %s', self, e)
         return False
--- a/lib/galaxy/datatypes/binary.py
+++ b/lib/galaxy/datatypes/binary.py
@@ -1395,15 +1395,23 @@ class SQlite(Binary):
         except Exception:
             return False
 
-    def sniff_table_names(self, filename, table_names):
+    def sniff_table_names(self, filename, table_names, equality=False):
+        """
+        test if the tables defined in the SQLite DB (filename) are
+        a subset of given list of tables (tablenames). if equality
+        is true the test is for equality
+        """
         # All table names should be in the schema
         try:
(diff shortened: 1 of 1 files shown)


lib/galaxy/datatypes/binary.py
```diff
@@ -1403,10 +1403,7 @@ def sniff_table_names(self, filename, table_names):
             tables_query = "SELECT name FROM sqlite_master WHERE type='table' ORDER BY name"
             result = c.execute(tables_query).fetchall()
             result = [_[0] for _ in result]
-            for table_name in table_names:
-                if table_name not in result:
-                    return False
-            return True
+            return set(result) == set(table_names)
```
mvdbeek: You can  also do `result = (_[0] for _ in result)`. And here I'd use `set(result).issubset(set(table_names))` to maintain the logic.

nsoranzo: Also `set(result).issubset(table_names)` works.

bernt-matthias:
I restored the subset check as default behavior and added the possibility to check for equality.

This will be needed here https://github.com/galaxyproject/galaxy/pull/8109 for the OSW and PQP datatypes since the tables of OSW are a superset of the tables of PQP.

bernt-matthias:
I solved it differently ... should we keep the PR .. maybe removing the equality check again but use the issubset function?

lib/galaxy/datatypes/binary.py
```diff
@@ -1395,18 +1395,23 @@ def sniff(self, filename):
...
             conn = sqlite.connect(filename)
             c = conn.cursor()
             tables_query = "SELECT name FROM sqlite_master WHERE type='table' ORDER BY name"
             result = c.execute(tables_query).fetchall()
-            result = [_[0] for _ in result]
-            for table_name in table_names:
-                if table_name not in result:
-                    return False
-            return True
+            result = (_[0] for _ in result)
```
```suggestion
            result = set(_[0] for _ in result)
```

lib/galaxy/datatypes/binary.py
```diff
@@ -1395,18 +1395,23 @@ def sniff(self, filename):
...
+            if equality:
+                return set(table_names) == set(result)
```
```suggestion
                return set(table_names) == result
```

lib/galaxy/datatypes/binary.py
```diff
@@ -1395,18 +1395,23 @@ def sniff(self, filename):
...
+            else:
+                return set(table_names).issubset(set(result))
```
```suggestion
                return set(table_names).issubset(result)
```

bernt-matthias:
I guess we can drop this. Solved it differently.

PR #9411, opened 2020-02-20
[20.01] Fix shed image resolution

dannon:
...to actually test base-path images as intended, (and then fail through to static/images).

### DIFF
--- a/lib/galaxy/webapps/galaxy/controllers/shed_tool_static.py
+++ b/lib/galaxy/webapps/galaxy/controllers/shed_tool_static.py
@@ -53,4 +53,4 @@ class ShedToolStatic(BaseUIController):
                     trans.response.set_content_type(mime)
             return open(asset_path, 'rb')
         else:
-            return None
+            raise RequestParameterInvalidException()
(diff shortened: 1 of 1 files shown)


lib/galaxy/webapps/galaxy/controllers/shed_tool_static.py
```diff
@@ -26,17 +32,25 @@ def index(self, trans, shed, owner, repo, tool, version, image_file):
...
         tool = trans.app.toolbox.get_tool(guid)
-        repo_path = tool._repository_dir
-        if 'static/images' not in image_file:
-            path = join(repo_path, 'static', 'images', image_file)
+        repo_path = os.path.abspath(tool._repository_dir)
+        asset_path = os.path.abspath(join(repo_path, image_file))
+
+        # test specified image_file path exactly first, then fail through to
+        # other locations.
+        # Might want to swap this around and check for static/images first if
```
mvdbeek, line 40: Yeah, large majority of images are in static/images, we should check that first.

PR #9413, opened 2020-02-21
[20.01] Fixes for IGV GEDA handling of genomes

blankenberg:
also adds genbank


mvdbeek:
Thanks @blankenberg !

PR #9415, opened 2020-02-21
[20.01] Fixes for get_estimated_display_viewport

mvdbeek:
Map objects cannot be unpacked, decrease log level, use context managers
for opening files.

Silences https://sentry.galaxyproject.org/sentry/main/issues/517474/:
```
line (##sequence-regionSL4.0ch07	1	67883646
) uses an unsupported ##sequence-region definition.
```
and similar.

### DIFF
--- a/lib/galaxy/datatypes/interval.py
+++ b/lib/galaxy/datatypes/interval.py
@@ -167,40 +167,36 @@ class Interval(Tabular):
             start = sys.maxsize
             end = 0
             max_col = max(chrom_col, start_col, end_col)
-            fh = open(dataset.file_name)
-            while True:
-                line = fh.readline(VIEWPORT_READLINE_BUFFER_SIZE)
-                # Stop if at end of file
-                if not line:
-                    break
-                # Skip comment lines
-                if not line.startswith('#'):
-                    try:
-                        fields = line.rstrip().split('\t')
-                        if len(fields) > max_col:
-                            if chrom is None or chrom == fields[chrom_col]:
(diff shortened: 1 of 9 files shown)


lib/galaxy/datatypes/interval.py
```diff
@@ -167,40 +167,39 @@ def get_estimated_display_viewport(self, dataset, chrom_col=None, start_col=None
...
-                        pass
-                # Make sure we are at the next new line
-                readline_count = VIEWPORT_MAX_READS_PER_LINE
-                while line.rstrip('\n\r') == line:
-                    assert readline_count > 0, Exception('Viewport readline count exceeded for dataset %s.' % dataset.id)
-                    line = fh.readline(VIEWPORT_READLINE_BUFFER_SIZE)
+            with open(dataset.file_name) as fh:
+                for line in util.iter_start_of_line(fh, VIEWPORT_READLINE_BUFFER_SIZE):
+                    # Stop if at end of file
                     if not line:
```
nsoranzo: Is this check still needed, given you are using a for loop?

mvdbeek: it's not, I dropped it in the other function~s~ and forgot this one, thanks!

mvdbeek:
@nsoranzo I restored the map unpacking in the last commit here.

nsoranzo:
> @nsoranzo I restored the map unpacking in the last commit here.

Thanks, I've pushed 2 other small follow-ups to #9416.

nsoranzo (approved):

PR #9416, opened 2020-02-21
[20.01] Fix wrong map use in many places

mvdbeek:
Map objects can't be serialized to JSON, so they can't be sent out as API responses, they can only be iterated over once, they can't be unpacked like lists or tuples, and just calling `map()` doesn't evaluate whatever function is mapped.

Some of these look pretty serious, and overall I think we should use them less often because often it is not immediately clear if using map is safe. List comprehensions or generators are often a better choice.

### DIFF
--- a/lib/galaxy/jobs/dynamic_tool_destination.py
+++ b/lib/galaxy/jobs/dynamic_tool_destination.py
@@ -66,11 +66,12 @@ def get_keys_from_dict(dl, keys_list):
     """
     This function builds a list using the keys from nest dictionaries
     """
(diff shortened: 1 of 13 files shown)


nsoranzo:
@mvdbeek I don't disagree with preferring comprehensions, but for unpacking `a, b = map(bool, [0, 1])` works fine for me on Python 3.6. Am I missing something?

mvdbeek:
... of course not, I need to check more things and assume less. I'll restore the map unpacking

Issue #9417, opened 2020-02-23
Calling ByteSize twice fails with floating point numbers


pcm32:
Currently the k8s runner is calling twice the ByteSize convenience class producing an error when provided initially a floating point number with units:

```
from galaxy.util.bytesize import ByteSize

# The equivalent of what happens within https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/jobs/runners/kubernetes.py#L294
mem_trans1 = ByteSize('0.9Gi').value
print mem_trans1
# 966367641.6

# Then second call to bytesize in https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/jobs/runners/kubernetes.py#L306
# produces the equivalent to:
mem_trans2 = ByteSize(mem_trans1).to_unit('M', as_string=False)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "galaxy/util/bytesize.py", line 32, in __init__
    self.value = parse_bytesize(value)
  File "galaxy/util/bytesize.py", line 51, in parse_bytesize
    value = value.upper()
AttributeError: 'float' object has no attribute 'upper'
```

This is really an issue with ByteSize, which is producing a float on the first call but cannot handle a float when receiving it later.

PR #9418, opened 2020-02-23
Handles floats correctly in ByteSize

pcm32:
This PR deals with issue #9417


pcm32:
Would it be possible to apply this to 20.01 branch as well? thanks!

pcm32:
Might be good to add a test case for ByteSize... but not sure where that is.

jmchilton:
I pushed this to 20.01 also, thanks for the fix.

PR #9421, opened 2020-02-24
[20.01] Conda startup fixes

nsoranzo:
- Don't check if python is from conda when `CONDA_EXE` env var is set.
  Since at least version 3.7.4 of the `python` package from the `defaults` conda channel, `python -V` doesn't reveal its origin any more.
- Some fixes suggested by `shellcheck`.
-  Source the full path of conda `activate` script.
  Otherwise, if a virtualenv is active on top of conda base env (e.g. when running `py37-first_startup` test via tox in TravisCI), the wrong `activate` script is sourced.
- Fix `CONDA_PREFIX` environment variable definition when using a shell not supported by conda.
- Upgrade Checkout GitHub Action to v2

### DIFF
--- a/.github/workflows/converter_tests.yaml
+++ b/.github/workflows/converter_tests.yaml
@@ -9,7 +9,7 @@ jobs:
       matrix:
         python-version: [3.7]
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
       with:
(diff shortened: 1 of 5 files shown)


mvdbeek:
Very nice, thanks so much @nsoranzo!

PR #9423, opened 2020-02-24
Fix first_startup Travis tests

nsoranzo:
- Fix py27-first_startup Travis job to actually use Python 2
  `scripts/common_startup.sh` now prefers python3 when creating Galaxy's virtual environment, so it needs to be run when inside the tox virtualenv, where `GALAXY_PYTHON` is defined.
- Let Galaxy create the `_galaxy_` conda env


nsoranzo:
Rebased and green!

Issue #8514, opened 2019-08-26
smarter Galaxy tool test for image comparision 


bgruening:
It would be nice if tests in Galaxy tools could be a little bit smarter in comparing images. Currently what most people do is using `sim_size` which is ok'ish.

We could add a comparison based on code from https://github.com/jterrace/pyssim maybe. Bonus points if we can just use pillow and dependencies that are already in Galaxy and can avoid the usage of skimage and co.

Pointed out by @erasche: https://github.com/JohannesBuchner/imagehash

mtekman:
https://github.com/holman/spaceman-diff

I've seen mentioned somewhere

and  imagemagick's compare function, which has an actual metric for number of changed pixels + fuzz option

`compare image1 image2 diff -fuzz N% -metric AE`

leobiscassi:
Hey @bgruening, I'm looking for ways to start contributing with galaxy code. I think this is a nice opportunity to get more comfortable with galaxy code base, can I go ahead with this issue?

martenson:
@leobiscassi please feel free to! If you keep us in the loop using this issue that'd be great.

leobiscassi:
@martenson great, I'll do that! thanks in advance

bgruening:
Awesome @leobiscassi! Let us know if you have questions here or on [gitter](https://gitter.im/galaxyproject/Lobby).

leobiscassi:
Hey guys, after some exploration to galaxy code, I think this task it's add a new comparison method like suggested by @bgruening after `sim_size` [1] condition in the _verify_ module [2], right!?

I'd like to know whether this addition shouldn't have major implications on xml.py [3] and test_parsing.py [4] like I see or I'm missing something.

[1] https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tool_util/verify/__init__.py#L126
[2] https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tool_util/verify
[3] https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tool_util/parser/xml.py
[4] https://github.com/galaxyproject/galaxy/blob/dev/test/unit/tool_util/test_parsing.py

bgruening:
[...]

PR #9425, opened 2020-02-24
Add delta_frac for sim_size to specify relative file size differences

bernt-matthias:
use case: I develop a a lare tool collection where some files differ much (ususally the large files) and some only a bit (the small files). by applying a large delta to all files meaningful differences in small files are not detected.

I guess this might also help here a bit https://github.com/galaxyproject/galaxy/issues/8514

### DIFF
--- a/lib/galaxy/tool_util/parser/util.py
+++ b/lib/galaxy/tool_util/parser/util.py
@@ -1,6 +1,7 @@
 from collections import OrderedDict
 
-DEFAULT_DELTA = "10000"
+DEFAULT_DELTA = 10000
+DEFAULT_DELTA_FRAC = None
 
 
 def is_dict(item):
--- a/lib/galaxy/tool_util/parser/xml.py
+++ b/lib/galaxy/tool_util/parser/xml.py
@@ -9,7 +9,10 @@ from math import isinf
 import packaging.version
 
 from galaxy.tool_util.deps import requirements
-from galaxy.tool_util.parser.util import DEFAULT_DELTA
+from galaxy.tool_util.parser.util import (
+    DEFAULT_DELTA,
+    DEFAULT_DELTA_FRAC
+)
 from galaxy.util import string_as_bool, xml_text, xml_to_string
 from .interface import (
     InputSource,
@@ -614,6 +617,7 @@ def __parse_test_attributes(output_elem, attrib, parse_elements=False, parse_dis
     attributes['lines_diff'] = int(attrib.pop('lines_diff', '0'))
     # Allow a file size to vary if sim_size compare
     attributes['delta'] = int(attrib.pop('delta', DEFAULT_DELTA))
+    attributes['delta_frac'] = float(attrib['delta_frac']) if 'delta_frac' in attrib else DEFAULT_DELTA_FRAC
     attributes['sort'] = string_as_bool(attrib.pop('sort', False))
     attributes['decompress'] = string_as_bool(attrib.pop('decompress', False))
     extra_files = []
--- a/lib/galaxy/tool_util/parser/yaml.py
+++ b/lib/galaxy/tool_util/parser/yaml.py
@@ -3,7 +3,10 @@ from collections import OrderedDict
 import packaging.version
 
 from galaxy.tool_util.deps import requirements
-from galaxy.tool_util.parser.util import DEFAULT_DELTA
+from galaxy.tool_util.parser.util import (
+    DEFAULT_DELTA,
+    DEFAULT_DELTA_FRAC
+)
 from .interface import (
     InputSource,
     PageSource,
@@ -219,7 +222,8 @@ def _parse_test(i, test_dict):
         defaults = {
             'compare': 'diff',
             'lines_diff': 0,
-            'delta': int(DEFAULT_DELTA),
+            'delta': DEFAULT_DELTA,
+            'delta_frac': DEFAULT_DELTA_FRAC,
             'sort': False,
         }
         # TODO
--- a/lib/galaxy/tool_util/verify/__init__.py
+++ b/lib/galaxy/tool_util/verify/__init__.py
@@ -16,7 +16,10 @@ try:
 except ImportError:
     pysam = None
 
-from galaxy.tool_util.parser.util import DEFAULT_DELTA
+from galaxy.tool_util.parser.util import (
+    DEFAULT_DELTA,
+    DEFAULT_DELTA_FRAC
+)
 from galaxy.util import unicodify
 from galaxy.util.compression_utils import get_fileobj
 from .asserts import verify_assertions
@@ -132,11 +135,7 @@ def verify(
             elif compare == 're_match_multiline':
                 files_re_match_multiline(local_name, temp_name, attributes=attributes)
             elif compare == 'sim_size':
-                delta = int(attributes.get('delta', DEFAULT_DELTA))
-                s1 = len(output_content)
-                s2 = os.path.getsize(local_name)
-                if abs(s1 - s2) > int(delta):
-                    raise AssertionError('Files %s=%db but %s=%db - compare by size (delta=%s) failed' % (temp_name, s1, local_name, s2, delta))
+                files_delta(local_name, temp_name, attributes=attributes)
             elif compare == "contains":
                 files_contains(local_name, temp_name, attributes=attributes)
             else:
@@ -192,6 +191,20 @@ def _verify_checksum(data, checksum_type, expected_checksum_value):
         template = "Output checksum [%s] does not match expected [%s] (using hash algorithm %s)."
         message = template % (actual_checksum_value, expected_checksum_value, checksum_type)
         raise AssertionError(message)
+
+
+def files_delta(file1, file2, attributes=None):
+    """Check the contents of 2 files for size differences."""
+    if attributes is None:
+        attributes = {}
+    delta = attributes.get('delta', DEFAULT_DELTA)
+    delta_frac = attributes.get('delta_frac', DEFAULT_DELTA_FRAC)
+    s1 = os.path.getsize(file1)
+    s2 = os.path.getsize(file2)
+    if abs(s1 - s2) > delta:
+        raise AssertionError('Files %s=%db but %s=%db - compare by size (delta=%s) failed' % (file1, s1, file2, s2, delta))
+    if delta_frac is not None and not (s1 - (s1 * delta_frac) <= s2 <= s1 + (s1 * delta_frac)):
+        raise AssertionError('Files %s=%db but %s=%db - compare by size (delta_frac=%s) failed' % (file1, s1, file2, s2, delta_frac))
 
 
 def files_diff(file1, file2, attributes=None):
--- a/lib/galaxy/tool_util/xsd/galaxy.xsd
+++ b/lib/galaxy/tool_util/xsd/galaxy.xsd
@@ -1312,9 +1312,14 @@ When this attribute is true and ``compare`` is set to ``diff``, try to decompres
 ]]></xs:documentation>
       </xs:annotation>
     </xs:attribute>
-    <xs:attribute name="delta" type="xs:integer">
+    <xs:attribute name="delta" type="xs:integer" default="10000">
       <xs:annotation>
-        <xs:documentation xml:lang="en">If ``compare`` is set to ``sim_size``, this is the number of bytes different allowed (default 10000).</xs:documentation>
+        <xs:documentation xml:lang="en">If ``compare`` is set to ``sim_size``, this is the maximum allowed absolute size difference (in bytes) between the data set that is generated in the test and the file in ``test-data/`` that is referenced by the ``file`` attribute. Can be combined with ``delta_frac``.</xs:documentation>
+      </xs:annotation>
+    </xs:attribute>
+    <xs:attribute name="delta_frac" type="xs:float">
+      <xs:annotation>
+        <xs:documentation xml:lang="en">If ``compare`` is set to ``sim_size``, this is the maximum allowed relative size difference between the data set that is generated in the test and the file in ``test-data/`` that is referenced by the ``file`` attribute. A value of 0.1 means that the file that is generated in the test can differ by at most 10% of the file in ``test-data``. The default is not to check for  relative size difference. Can be combined with ``delta``.</xs:documentation>
       </xs:annotation>
     </xs:attribute>
   </xs:complexType>
--- a/test/unit/test_verify.py
+++ b/test/unit/test_verify.py
@@ -6,6 +6,7 @@ import pytest
 
 from galaxy.tool_util.verify import (
     files_contains,
+    files_delta,
     files_diff,
     files_re_match,
     files_re_match_multiline,
@@ -20,14 +21,18 @@ MULTILINE_MATCH = b".*"
 TestFile = collections.namedtuple('TestFile', 'value path')
 
 
-def generate_tests(multiline=False):
+def test_file_list():
     files = []
     for b, ext in [(F1, '.txt'), (F2, '.txt'), (F3, '.pdf'), (F4, '.txt'), (MULTILINE_MATCH, '.txt')]:
         fd, path = tempfile.mkstemp(suffix=ext)
         with os.fdopen(fd, 'wb') as out:
             out.write(b)
         files.append(TestFile(b, path))
-    f1, f2, f3, f4, multiline_match = files
+    return files
+
+
+def generate_tests(multiline=False):
+    f1, f2, f3, f4, multiline_match = test_file_list()
     if multiline:
         tests = [(multiline_match, f1, {'lines_diff': 0, 'sort': True}, None)]
     else:
@@ -40,6 +45,23 @@ def generate_tests(multiline=False):
     return tests
 
 
+def generate_tests_sim_size():
+    f1, f2, f3, f4, multiline_match = test_file_list()
+    # tests for equal files
+    tests = [(f1, f1, None, None),  # pass default values
+             (f1, f1, {'delta': 0}, None),  # pass for values that should always pass
+             (f1, f1, {'delta_frac': 0.0}, None)]
+    # tests for two different file (diff is 422 lines, 0.011709602)
+    tests += [(f1, f2, None, None),  # pass default values
+             (f1, f2, {'delta': 422}, None),  # test around the actual difference of two different
+             (f1, f2, {'delta_frac': 84.4}, None),
+             (f1, f2, {'delta': 421}, AssertionError),
+             (f1, f2, {'delta_frac': 84.3}, AssertionError),
+             (f1, f2, {'delta': 422, 'delta_frac': 84.3}, AssertionError),  # test with combination where at least one leads to an assertion
+             (f1, f2, {'delta': 421, 'delta_frac': 84.4}, AssertionError)]
+    return tests
+
+
 @pytest.mark.parametrize('file1,file2,attributes,expect', generate_tests())
 def test_files_contains(file1, file2, attributes, expect):
     if expect is not None:
@@ -47,6 +69,15 @@ def test_files_contains(file1, file2, attributes, expect):
             files_contains(file1.path, file2.path, attributes)
     else:
         files_contains(file2.path, file2.path, attributes)
+
+
+@pytest.mark.parametrize('file1,file2,attributes,expect', generate_tests_sim_size())
+def test_files_delta(file1, file2, attributes, expect):
+    if expect is not None:
+        with pytest.raises(expect):
+            files_delta(file1.path, file2.path, attributes)
+    else:
+        files_delta(file1.path, file2.path, attributes)
 
 
 @pytest.mark.parametrize('file1,file2,attributes,expect', generate_tests())
--- a/test/functional/tools/samples_tool_conf.xml
+++ b/test/functional/tools/samples_tool_conf.xml
@@ -55,6 +55,7 @@
   <tool file="qc_stdout.xml" />
   <tool file="md5sum.xml" />
   <tool file="checksum.xml" />
+  <tool file="sim_size_delta.xml" />
   <tool file="composite_shapefile.xml" />
   <tool file="is_valid_xml.xml" />
   <tool file="param_value_from_file.xml"/>
--- a/test/functional/tools/sim_size_delta.xml
+++ b/test/functional/tools/sim_size_delta.xml
@@ -1,0 +1,22 @@
+<tool id="sim_size_delta" name="sim_size_delta" version="0.1.0">
+  <command><![CDATA[
+    head -c 80 '$in' > '$out_smaller' &&
+    cp '$in' '$out_larger' &&
+    head -c 20 '$in' >> '$out_larger'
+  ]]></command>
+  <inputs>
+    <param name="in" type="data" format="data"/>
+  </inputs>
+  <outputs>
+    <data name="out_smaller" format="data" label="smaller"/>
+    <data name="out_larger" format="data" label="larger"/>
+  </outputs>
+  <tests>
+    <!-- take a test file of 100b size and add/remove 20b -->
+    <test expect_num_outputs="2">
+      <param name="in" value="1.sff" ftype="txt" />
+      <output name="out_smaller" value="1.sff" compare="sim_size" delta="20" delta_frac="0.2"/>
+      <output name="out_larger" value="1.sff" compare="sim_size" delta="20" delta_frac="0.2"/>
+    </test>
+  </tests>
+</tool>


nsoranzo:
I'm not sure overloading `delta` is a good idea, maybe add a separate `frac_delta` attribute?

lib/galaxy/tool_util/verify/__init__.py
```diff
@@ -131,10 +131,12 @@ def get_filename(filename):
             elif compare == 're_match_multiline':
                 files_re_match_multiline(local_name, temp_name, attributes=attributes)
             elif compare == 'sim_size':
-                delta = attributes.get('delta', '100')
+                delta = float(attributes.get('delta', '10000'))
```
mvdbeek: We shouldn't increase the default delta, that'll alter existing tools' test results

bernt-matthias: True, but the default used here was just inconsistent with:  https://github.com/galaxyproject/galaxy/blob/5214ba0b5b1af85ec32973656e9b73fe0e9af7f4/lib/galaxy/tool_util/parser/xml.py#L615

mvdbeek: We should have a "constant" for that.

bernt-matthias:
> I'm not sure overloading delta is a good idea, maybe add a separate frac_delta attribute?

Also thought about this, but this would be to the cost of adding quite a bit of logic to the linting o the tool xml.

lib/galaxy/tool_util/verify/__init__.py
bernt-matthias: Is there already a place where such constants used for parsing the xml are stored?

mvdbeek: I don't think so, but maybe `lib/galaxy/tool_util/parser/util.py` would be a good start ?

lib/galaxy/tool_util/parser/xml.py
```diff
@@ -612,7 +613,7 @@ def __parse_test_attributes(output_elem, attrib, parse_elements=False, parse_dis
     # Number of lines to allow to vary in logs (for dates, etc)
     attributes['lines_diff'] = int(attrib.pop('lines_diff', '0'))
     # Allow a file size to vary if sim_size compare
-    attributes['delta'] = float(attrib.pop('delta', '10000'))
+    attributes['delta'] = float(attrib.pop('delta', DEFAULT_DELTA))
```
mvdbeek: Can you also replace the equivalent part in yaml.py ?

bernt-matthias: Excellent idea, in particular since yet another default was used there.

jmchilton:
> I'm not sure overloading delta is a good idea, maybe add a separate frac_delta attribute?

+100 - we shouldn't have a parameter that does one thing from 0-1 and a very different thing > 1 IMO. It is a lot more work and that sucks, but I think it is so much clearer.

bernt-matthias:
Then I will follow the majority.. lets see when I find time.

bernt-matthias:
Btw. there seems no test for `sim_size` and `delta`. Where and how would this go?

I will rebase this one after https://github.com/galaxyproject/galaxy/pull/9439 is merged.

nsoranzo:
> Btw, there seems no test for `sim_size` and `delta`. Where and how would this go?

Great, I wanted to suggest that! I think they can be added to `test/unit/test_verify.py`

bernt-matthias:
Did some preparations for the unit tests. But I can't get my head arround how to set this up for delta in the current setup. 

Seems that all tests setups run through all test functions and are expected to behave the same. Is this correct?

test/unit/test_verify.py
```diff
@@ -49,6 +54,15 @@ def test_files_contains(file1, file2, attributes, expect):
         files_contains(file2.path, file2.path, attributes)
 
 
+@pytest.mark.parametrize('file1,file2,attributes,expect', generate_tests())
```
mvdbeek: Use a separate function (or a flat list if you don't have to do anything fancy for generating the test case) for the second argument here (which should be an iterable that can be unpacked to as many values as you have arguments)

bernt-matthias:
Does someone know if `os.path.getsize` works with symlinks?

nsoranzo:
> Does someone know if `os.path.getsize` works with symlinks?

It seems to return the size of the linked file, after resolving, on Linux.

bernt-matthias:
Then I'm satisfied with the current state.

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -1314,7 +1314,12 @@ When this attribute is true and ``compare`` is set to ``diff``, try to decompres
     </xs:attribute>
     <xs:attribute name="delta" type="xs:integer">
       <xs:annotation>
-        <xs:documentation xml:lang="en">If ``compare`` is set to ``sim_size``, this is the number of bytes different allowed (default 10000).</xs:documentation>
+        <xs:documentation xml:lang="en">If ``compare`` is set to ``sim_size``, this is the number of bytes different allowed. Can be combined with ``delta_frac``</xs:documentation>
+      </xs:annotation>
+    </xs:attribute>
+    <xs:attribute name="delta_frac" type="xs:float">
+      <xs:annotation>
+        <xs:documentation xml:lang="en">If ``compare`` is set to ``sim_size``, the size of the smaller file in bytes has to be at least this fraction times the size of the larger file in bytes. Can be combined with ``delta``. (default 1.0).</xs:documentation>
```
mvdbeek: I find the text hard to understand. I guess the point is to allow a margin ? Since this seems independent of the delta I wouldn't mention it and rename this to `size_margin`. Which file is bigger doesn't matter for a margin, so I think that would be easier to understand.


```suggestion
    <xs:attribute name="size_margin" type="xs:float" default="0">
      <xs:annotation>
        <xs:documentation xml:lang="en">If ``compare`` is set to ``sim_size`` the test will fail if the difference in output size exceeds the test data output by ``size_margin``. 0 means no margin, 0.1 means 10% margin, 2 means the output may be 2 times larger or smaller than the provided test output dataset. Note that ``sim_size`` is applied independently. </xs:documentation>
```

bernt-matthias: I like the idea of giving examples and I agree that the original formulation has problems.

But there are some words in your formulation that are wrong or might be misleading:

- actually the value of `delta_frac` has to be between 0 and 1 (larger values work, but are meaningless since 1 already allows any difference)
- In your example a value of 0.1 would allow the size differ by a factor of 10, and a 10% size difference should correspond to a delta_frac=0.9
- "difference" indicates (to me) that the error margin is additive 
- "exceeds the test data output" actually its the lager of the test file and the test output
- Also note that the default is actually 1.

> Note that ``sim_size`` is applied independently

actually one needs to set ``sim_size``, its `delta` that is applied independently.

We can still think about the semantics that are easiest to understand. Currently the test is `min/max > delta_frac`, i.e. `min > delta_frac * max`. 

But we could as well invert this `max/min > delta_frac`. Maybe even `log(min/max)>delta_frac` (which makes expressing orders of magnitude easy).

Few comments from you would be great. Then I will work on the formulation.

mvdbeek: I think I've changed my mind, `margin` isn't well-defined, `delta_frac` is better.
How about
```suggestion
    <xs:attribute name="delta_frac" type="xs:float">
      <xs:annotation>
        <xs:documentation xml:lang="en">If ``compare`` is set to ``sim_size``, the relative difference in size allowed between the generated output and the provided test dataset. 0 means no difference in size allowed, 0.1 means 10% of difference allows, 2 means that the difference may be twice the size of the provided test dataset. Note that ``delta`` is applied independently. (default 0).</xs:documentation>
```

The implementation then could be (assuming s1 is the provided test dataset, not sure on that)
```python
actual_delta = abs(s1 - s2)
actual_delta_frac = actual_delta / s1
if actual_delta_frac > delta_frac:
    raise AssertionError
```

bernt-matthias: I would prefer the current implementation over your suggestion, because it appears to be (more) analogous to the implementation for `delta` (`abs(s1 - s2)`), i.e. it just tests for the absolute (`delta`) / relative (`delta_frac`) size difference and it does not matter which file is the provided test dataset and which the test output.

mvdbeek: If you can rewrite the help to be more understandable and calculatable by a user I can get on board with that. Which file is the provided file should be easy to figure out.

bernt-matthias: Inverted the meaning of the value: `min/max > delta` got `1-min/max>delta` (which is equivalent to `abs(s1/s2)/max>delta`).

This is nearly what you meant previously (`actual_delta_frac = actual_delta / s1`) .. I now see the advantage that both values allow to make a statement on the same thing (the abs difference) .. just in different ways.

mvdbeek: It does read better now, but  I think it's inconvenient that the meaningful values max out at 1.

mvdbeek (dismissed):

mvdbeek:
There's a test failure now. If you fix this I think we can merge (I think allowing > 100% size difference probably doesn't make a lot of sense).

bernt-matthias:
> I think allowing > 100% size difference probably doesn't make a lot of sense

Idea was to check for file existence only...

bernt-matthias:
Thanks for the reminder. Your question 

> How do I restrict it to 110% ?

made me think about this again. I guess the current implementation is much more intuitive. That is, everything is wrt the file size of the file in test-data. So 0.1 means +- 10% of the file size in test-data.

Since I always confused which of s1 and s2 is which file I have also a tool ready that tests this that I could easily add... might be worth it since currently there is not tool testing for delta/delta_frac

bernt-matthias:
Narf. Using the default of `float('inf')` for delta_frac does not work. Since again (de)-jsonifying  does not work. 

Since https://github.com/galaxyproject/galaxy/commit/ad96541ff85057665eb749e3b0f67b7123e53ecf `safe_dumps` transforms `Inf` to `'__Infinity__'`. 

As mentioned in the commit

> We should only be shipping valid JSON now, but it still needs to be parsed and handled on API consumers, of course.

it seems that the Galaxy API consumer can't cope with itself. hrm...

mvdbeek:
> Since [ad96541](https://github.com/galaxyproject/galaxy/commit/ad96541ff85057665eb749e3b0f67b7123e53ecf) `safe_dumps` transforms `Inf` to `'__Infinity__'`.

Well that's not right and this dates from 2014, consequently I don't think we can merge this with https://github.com/galaxyproject/galaxy/pull/9425/commits/8f72dfe8ee1a3e4394976ef272057961030891c0.
`json.dumps` encodes inf just fine, and NaN is not valid JSON.

 Is that really necessary for this PR ?

bernt-matthias:
In the tests for https://github.com/galaxyproject/galaxy/pull/9425/commits/47ae6b2816b6ee73aa76cbe5bb8e90505ec2c267 I have seen `'__Infinity__'` and a good explanation seemed to be that the test description is encoded in json at some point which introduces this string. 

For instance in the converter tests https://github.com/galaxyproject/galaxy/runs/643979914:

```
======================================================================
FAIL: ( CONVERTER_interval_to_bigwig_0 ) > Test-1
----------------------------------------------------------------------
  File "/home/runner/work/galaxy/galaxy/galaxy root/test/functional/test_toolbox.py", line 99, in test_tool
    self.do_it(tool_version=tool_version, test_index=test_index)
  File "/home/runner/work/galaxy/galaxy/galaxy root/test/functional/test_toolbox.py", line 36, in do_it
    verify_tool(tool_id, self.galaxy_interactor, resource_parameters=resource_parameters, test_index=test_index, tool_version=tool_version, register_job_data=register_job_data)
  File "/home/runner/work/galaxy/galaxy/galaxy root/lib/galaxy/tool_util/verify/interactor.py", line 852, in verify_tool
    raise e
  File "/home/runner/work/galaxy/galaxy/galaxy root/lib/galaxy/tool_util/verify/interactor.py", line 848, in verify_tool
    job_stdio = _verify_outputs(testdef, test_history, jobs, tool_id, data_list, data_collection_list, galaxy_interactor, quiet=quiet)
  File "/home/runner/work/galaxy/galaxy/galaxy root/lib/galaxy/tool_util/verify/interactor.py", line 1023, in _verify_outputs
    raise JobOutputsError(found_exceptions, job_stdio)
galaxy.tool_util.verify.interactor.JobOutputsError: could not convert string to float: '__Infinity__'
```

I did reset to this commit for demonstration purposes (for by own reference the sha of the other commit: 8f72dfe8ee).

> Is that really necessary for this PR ?

No, we just need a good default value for `delta_frac`. `Inf` would have been perfect in my opinion since it would accept any difference -- which is what is needed. Maybe `None` or a negative value which we would need to test for to skip the `delta_frac` test ...

mvdbeek:
Yes, that makes sense. What I meant is that encoding `inf` as `__Infinity__` is not necessary (anymore?). `json.loads(json.dumps(float('inf')))` works just fine, so I'd remove https://github.com/jmchilton/galaxy/blob/53f5a3097f0e6736132530617d8eb2ad5937443b/lib/galaxy/util/json.py#L38 instead of adapting to this non-standard encoding.

bernt-matthias:
Jep, this was also my first idea. But then I wondered what happens if `'Infinity'` appears somewhere in a tool, eg a param value. If this is treated by the json en/decoder `'__Infinity__'` seems a wee bit safer. 

I could change it and add a test.

bernt-matthias:
But on an even closer thought the problem allready would exist, since `json.loads` already decodes `'Infinity'`...

mvdbeek:
That is all very unfortunate, I was thinking that if a json.dumps / json.loads roundtrip preserves `inf` we don't need to do this special handling ... but that isn't actually JSON spec compliant.
I think https://github.com/galaxyproject/galaxy/commit/8f72dfe8ee1a3e4394976ef272057961030891c0 is good after all, sorry for the noise.

mvdbeek:
Yeah, ... this is all tricky. This is from the json.dumps docstrings
```
If ``allow_nan`` is false, then it will be a ``ValueError`` to
serialize out of range ``float`` values (``nan``, ``inf``, ``-inf``) in
strict compliance of the JSON specification, instead of using the
JavaScript equivalents (``NaN``, ``Infinity``, ``-Infinity``).
```
I'd argue our consumers are mostly JavaScript and python, which deal fine with this.
Maybe let's try dropping the encoding and see what the tests will say ?

bernt-matthias:
Lets still try it .. :) 

To my surprise `planemo` happily accepts: 

```
<tool id="infinity" name="infinity" version="0.1.0">
  <command><![CDATA[
    echo "text $intext" > '$NaN' &&
    echo "float $infloat" >> '$NaN' &&
    echo "select $inselect" >> '$NaN' &&
    echo "param $Infinity" >> '$NaN'
  ]]></command>
  <inputs>
    <param name="intext" type="text" value=""/>
    <param name="infloat" type="float" value=""/>
    <param name="inselect" type="select" >
        <option value="Infinity">Infinity</option>
        <option value="NaN">NaN</option>
    </param>
    <param name="Infinity" type="text" value=""/>
  </inputs>
  <outputs>
    <data name="NaN" format="txt" label="out"/>
  </outputs>
  <tests>
    <test expect_num_outputs="1">
      <param name="intext" value="Infinity" />
      <param name="infloat" value="Inf" />
      <param name="inselect" value="Infinity" />
      <param name="Infinity" value="VAL" />
      <output name="NaN">
        <assert_contents>
          <has_line line="text Infinity"/>
          <has_line line="float inf"/>
          <has_line line="select Infinity"/>
          <has_line line="param VAL"/>
          <has_text_matching expression="Infinity"/>
        </assert_contents>
      </output>
    </test>
  </tests>
</tool>
```

also with python's default handling of Inf,... (Note that `Inf` in the float parameter is really taken as `float('Inf')` ). 

With the old non-standard encoding function adding `delta_frac="Infinity"` fails with the same Exception as above.

So your suggestion is now to revert to https://github.com/galaxyproject/galaxy/commit/8f72dfe8ee1a3e4394976ef272057961030891c0 and to have the additional `__Infinity__` -> `inf` conversion in the decoder?

bernt-matthias:
> And just doing JSON.parse('{"a": ["1", 3, Infinity]}') fails in the browser console.

How about `JSON.parse('{"a": ["1", 3, "Infinity"]}')` ... ?

mvdbeek:
> JSON.parse('{"a": ["1", 3, "Infinity"]}')

that's the string Infinity, no parser of json would have problems with this.

> So your suggestion is now to revert to [8f72dfe](https://github.com/galaxyproject/galaxy/commit/8f72dfe8ee1a3e4394976ef272057961030891c0) and to have the additional `__Infinity__` -> `inf` conversion in the decoder?

This seems like the most correct thing to do, but I don't like it at all. This is potentially an expensive conversion.

> To my surprise `planemo` happily accepts:

I don't think we need to worry about the backend, this is unlikely to be an issue. The problem are API consumers that adhere to the JSON standard, such as the Galaxy interface.

I don't like addressing this in this PR, I think we need to have a separate discussion about this.

Can you remind me why infinity is a good default ? Do we need a default value ? Can the default value be `None` if we need one ?

bernt-matthias:
> Can you remind me why infinity is a good default ? Do we need a default value ? Can the default value be None if we need one ?

We need to set a default such that any difference is allowed (otherwise we would change the testing behavior). 

- `delta_frac=Inf` would naturally yield true for `(s1 - (s1 * delta_frac) <= s2 <= s1 + (s1 * delta_frac))`
- choosing some large float value might be an option. In theory having a large test file (s1) might overflow `sys.float_info.max` (`1.7976931348623157e+308`), but since test files are required to be small this is only a theoretical concern. 
- `None` will cause problems, since we use `float(delta_frac)` at several places.
- A negative number or `0` as default seems to be an easy workaround. Then the test would be `delta_frac >= 0 and not (s1 - (s1 * delta_frac) <= s2 <= s1 + (s1 * delta_frac))` to raise the Exception

The last option seems to be the easiest solution, ...

mvdbeek:
> We need to set a default such that any difference is allowed (otherwise we would change the testing behavior).

None ? I mean, noone has used this before, the default value we choose should only apply if `delta_frac` is used.

mvdbeek:
> * `None` will cause problems, since we use `float(delta_frac)` at several places.

If `None` skip the entire calculation.

lib/galaxy/tool_util/verify/__init__.py
```diff
@@ -194,6 +193,20 @@ def _verify_checksum(data, checksum_type, expected_checksum_value):
...
+    """Check the contents of 2 files for size differences."""
+    if attributes is None:
+        attributes = {}
+    delta = int(attributes.get('delta', DEFAULT_DELTA))
+    delta_frac = float(attributes.get('delta_frac', DEFAULT_DELTA_FRAC))
+    s1 = os.path.getsize(file1)
+    s2 = os.path.getsize(file2)
+    if abs(s1 - s2) > delta:
+        raise AssertionError('Files %s=%db but %s=%db - compare by size (delta=%s) failed' % (file1, s1, file2, s2, delta))
+    if not (s1 - (s1 * delta_frac) <= s2 <= s1 + (s1 * delta_frac)):
```
mvdbeek:
```suggestion
    if delta_frac is not None and not (s1 - (s1 * delta_frac) <= s2 <= s1 + (s1 * delta_frac)):
```

bernt-matthias:
`None` would be good, but https://github.com/galaxyproject/galaxy/blob/f98ce201c2a8ed537df70e3106197868616c245b/lib/galaxy/tool_util/parser/xml.py#L620 would also need special treatment. So I guess a default value `<=0` (which makes no sense anyway) and the suggested change of the if statement would lead to the desired behavior:

> the default value we choose should only apply if delta_frac is used.

Fine for you? Then I would go for this option.

lib/galaxy/tool_util/parser/util.py
```diff
@@ -1,6 +1,7 @@
 from collections import OrderedDict
 
 DEFAULT_DELTA = "10000"
+DEFAULT_DELTA_FRAC = float('inf')
```
mvdbeek:
```suggestion
DEFAULT_DELTA_FRAC = None
```

lib/galaxy/tool_util/parser/yaml.py
```diff
@@ -220,6 +223,7 @@ def _parse_test(i, test_dict):
             'compare': 'diff',
             'lines_diff': 0,
             'delta': int(DEFAULT_DELTA),
+            'delta_frac': float(DEFAULT_DELTA_FRAC),
```
mvdbeek:
```suggestion
            'delta_frac': DEFAULT_DELTA_FRAC,
```

lib/galaxy/tool_util/parser/util.py
```diff
@@ -1,6 +1,7 @@
 from collections import OrderedDict
 
 DEFAULT_DELTA = "10000"
```
mvdbeek:
```suggestion
DEFAULT_DELTA = 10000
```

lib/galaxy/tool_util/parser/yaml.py
```diff
@@ -220,6 +223,7 @@ def _parse_test(i, test_dict):
             'compare': 'diff',
             'lines_diff': 0,
             'delta': int(DEFAULT_DELTA),
```
mvdbeek:
```suggestion
            'delta': DEFAULT_DELTA,
```

lib/galaxy/tool_util/verify/__init__.py
```diff
@@ -194,6 +193,20 @@ def _verify_checksum(data, checksum_type, expected_checksum_value):
         raise AssertionError(message)
 
 
+def files_delta(file1, file2, attributes=None):
+    """Check the contents of 2 files for size differences."""
+    if attributes is None:
+        attributes = {}
+    delta = int(attributes.get('delta', DEFAULT_DELTA))
+    delta_frac = float(attributes.get('delta_frac', DEFAULT_DELTA_FRAC))
```
mvdbeek:
```suggestion
    delta = attributes.get('delta', DEFAULT_DELTA)
    delta_frac = attributes.get('delta_frac', DEFAULT_DELTA_FRAC)
```

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -1312,9 +1312,14 @@ When this attribute is true and ``compare`` is set to ``diff``, try to decompres
...
       </xs:annotation>
     </xs:attribute>
-    <xs:attribute name="delta" type="xs:integer">
+    <xs:attribute name="delta" type="xs:integer" default="10000">
       <xs:annotation>
-        <xs:documentation xml:lang="en">If ``compare`` is set to ``sim_size``, this is the number of bytes different allowed (default 10000).</xs:documentation>
+        <xs:documentation xml:lang="en">If ``compare`` is set to ``sim_size``, this is the maximum allowed absolute size difference (in bytes) between the data set that is generated in the test and the file in ``test-data/`` that is referenced by the ``file`` attribute. Can be combined with ``delta_frac``.</xs:documentation>
+      </xs:annotation>
+    </xs:attribute>
+    <xs:attribute name="delta_frac" type="xs:float" default="INF">
```
```suggestion
    <xs:attribute name="delta_frac" type="xs:float">
```

lib/galaxy/util/json.py
```diff
@@ -66,11 +66,7 @@ def safe_dumps(*args, **kwargs):
     json.dumps to blow up if it encounters Infinity/NaN, and we 'fix' it before
     re-encoding.
     """
-    try:
-        dumped = json.dumps(*args, allow_nan=False, **kwargs)
-    except ValueError:
-        obj = swap_inf_nan(copy.deepcopy(args[0]))
-        dumped = json.dumps(obj, allow_nan=False, **kwargs)
+    dumped = json.dumps(*args, **kwargs)
```
mvdbeek:
```suggestion
    try:
        dumped = json.dumps(*args, allow_nan=False, **kwargs)	
    except ValueError:	
        obj = swap_inf_nan(copy.deepcopy(args[0]))	
        dumped = json.dumps(obj, allow_nan=False, **kwargs)
```

mvdbeek (commented):
I think this should work fine.

lib/galaxy/tool_util/verify/__init__.py
bernt-matthias: I see. I will go for it. Can you edit your suggestion `frace` -> `frac`.

lib/galaxy/tool_util/parser/xml.py
```diff
@@ -614,6 +617,7 @@ def __parse_test_attributes(output_elem, attrib, parse_elements=False, parse_dis
     attributes['lines_diff'] = int(attrib.pop('lines_diff', '0'))
     # Allow a file size to vary if sim_size compare
     attributes['delta'] = int(attrib.pop('delta', DEFAULT_DELTA))
+    attributes['delta_frac'] = float(attrib.pop('delta_frac', DEFAULT_DELTA_FRAC))
```
mvdbeek:
```suggestion
    attributes['delta_frac'] = float(attrib['delta_frac']) if 'delta_frac' in attrib else DEFAULT_DELTA_FRAC
```

lib/galaxy/tool_util/verify/__init__.py
mvdbeek: Sure, fixed this one and the one above.

bernt-matthias:
Any need for a tool test using delta / delta_frac? .. I would have this ready.

bernt-matthias:
Should I squash the commits? Since it's quite messy in here :) .. or is this done while merging?

mvdbeek:
I think that'd be nice, but the unit tests are also OK for me. I can squash the commits on merge, unless you want to edit the commit messages.

mvdbeek (approved):
Wow, that was more complicated than I thought. Thanks a lot @bernt-matthias!

bernt-matthias:
Test has been added .. unfortunately we can't test the failure case (delta / delta_frac smaller then needed).

PR #9429, opened 2020-02-25
Use a Galaxy root path with a space for converters and mulled tests

nsoranzo:
Also:
- Fix integration tests for checkout@v2 action: the working directory is not updated any more when `path` is specified
- Use checkout action to clone the galaxy-test-data repo and rsync to move them in place
- Install virtualenv<20 when running tests via tox in a directory containing spaces, xref. https://github.com/pypa/virtualenv/pull/1660


mvdbeek:
Awesome, thanks again @nsoranzo !

PR #9432, opened 2020-02-25
[20.01] Fix login to accounts using different email capitalization

mvdbeek:
https://github.com/galaxyproject/galaxy/pull/8631 added case-insensitive matches to user emails. If two accounts with different email capitalizations exist one of them will be picked at random for login.
To fix this we first check for exact match and only if there is no direct match we lowercase the comparison.

### DIFF
--- a/lib/galaxy/managers/users.py
+++ b/lib/galaxy/managers/users.py
@@ -489,7 +489,9 @@ class UserManager(base.ModelManager, deletable.PurgableManagerMixin):
                 return "Failed to produce password reset token. User not found."
 
     def get_reset_token(self, trans, email):
-        reset_user = trans.sa_session.query(self.app.model.User).filter(func.lower(self.app.model.User.table.c.email) == email.lower()).first()
+        reset_user = trans.sa_session.query(self.app.model.User).filter(self.app.model.User.table.c.email == email.lower()).first()
+        if not reset_user and email != email.lower():
+            reset_user = trans.sa_session.query(self.app.model.User).filter(func.lower(self.app.model.User.table.c.email) == email.lower()).first()
         if reset_user:
             prt = self.app.model.PasswordResetToken(reset_user)
             trans.sa_session.add(prt)
--- a/lib/galaxy/webapps/galaxy/controllers/user.py
+++ b/lib/galaxy/webapps/galaxy/controllers/user.py
@@ -135,9 +135,14 @@ class User(BaseUIController, UsesFormDefinitionsMixin, CreatesApiKeysMixin):
         if not login or not password:
             return self.message_exception(trans, "Please specify a username and password.")
         user = trans.sa_session.query(trans.app.model.User).filter(or_(
-            func.lower(trans.app.model.User.table.c.email) == login.lower(),
+            trans.app.model.User.table.c.email == login,
             trans.app.model.User.table.c.username == login
         )).first()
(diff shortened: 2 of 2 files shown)


lib/galaxy/webapps/galaxy/controllers/user.py
```diff
@@ -135,9 +135,14 @@ def __validate_login(self, trans, payload={}, **kwd):
         if not login or not password:
             return self.message_exception(trans, "Please specify a username and password.")
         user = trans.sa_session.query(trans.app.model.User).filter(or_(
-            func.lower(trans.app.model.User.table.c.email) == login.lower(),
+            trans.app.model.User.table.c.email == login,
             trans.app.model.User.table.c.username == login
         )).first()
+        if not user and login.lower() != login:
+            user = trans.sa_session.query(trans.app.model.User).filter(or_(
```
nsoranzo:
```suggestion
            user = trans.sa_session.query(trans.app.model.User).filter(
```

lib/galaxy/webapps/galaxy/controllers/user.py
```diff
@@ -135,9 +135,14 @@ def __validate_login(self, trans, payload={}, **kwd):
...
+                func.lower(trans.app.model.User.table.c.email) == login.lower(),
```
nsoranzo:
```suggestion
                func.lower(trans.app.model.User.table.c.email) == login.lower()
```

lib/galaxy/webapps/galaxy/controllers/user.py
```diff
@@ -135,9 +135,14 @@ def __validate_login(self, trans, payload={}, **kwd):
...
+                trans.app.model.User.table.c.username == login
```
nsoranzo: No need to try again to match as a username.

```suggestion
```

lib/galaxy/webapps/galaxy/controllers/user.py
```diff
@@ -135,9 +135,14 @@ def __validate_login(self, trans, payload={}, **kwd):
...
+            )).first()
```
nsoranzo:
```suggestion
            ).first()
```

dannon (approved):

nsoranzo (approved):

lib/galaxy/managers/users.py
```diff
@@ -489,7 +489,9 @@ def send_reset_email(self, trans, payload={}, **kwd):
                 return "Failed to produce password reset token. User not found."
 
     def get_reset_token(self, trans, email):
-        reset_user = trans.sa_session.query(self.app.model.User).filter(func.lower(self.app.model.User.table.c.email) == email.lower()).first()
+        reset_user = trans.sa_session.query(self.app.model.User).filter(self.app.model.User.table.c.email == email.lower()).first()
```
nsoranzo, line 492: @mvdbeek Just noticed while merging forward: should this have been:

```python
        reset_user = trans.sa_session.query(self.app.model.User).filter(self.app.model.User.table.c.email == email).first()
```

(i.e. without the `.lower()` on `email`) ?

mvdbeek: Yes

nsoranzo: https://github.com/galaxyproject/galaxy/pull/9446

Issue #9251, opened 2020-01-22
Unify Popen calls


bernt-matthias:
Discussion started in https://github.com/galaxyproject/galaxy/pull/9246. 

There are many 

```
p = subprocess.Popen(...)
stdout, stderr = p.communicate(...)
```

calls which should be unified with an already existing function from  https://github.com/galaxyproject/galaxy/blob/9e76dcd9d381545ef627b63ec406dccf34bc9ae9/lib/galaxy/tool_util/deps/commands.py#L97

`_wait` needs to be unicodified since otherwise bytes are returned in py3 instead of str.

bernt-matthias:
- [ ] close_fds
  - lib/galaxy/containers/__init__.py
  - lib/galaxy/visualization/plugins/interactive_environments.py
  - lib/galaxy/tools/expressions/evaluation.py
  - lib/galaxy/tool_shed/galaxy_install/tool_dependencies/recipe/install_environment.py + bufsize + cwd
- [ ] use stdin
  - lib/galaxy/datatypes/converters/bgzip.py: https://github.com/galaxyproject/galaxy/issues/9672
  - 
- [ ] use returned process 
  - lib/galaxy/datatypes/dataproviders/external.py
  - lib/galaxy/util/lazy_process.py
  - lib/galaxy/jobs/runners/util/sudo.py
  - lib/galaxy_test/driver/driver_util.py + preexec_fn + cwd
- [ ] use env -> maybe shell_process
  - lib/galaxy/tool_util/deps/brew_exts.py
  - lib/galaxy/tool_util/deps/commands.py + shell
  - lib/galaxy/tool_util/deps/resolvers/modules.py
  - lib/galaxy/tool_util/deps/resolvers/lmod.py
  - lib/galaxy/util/pastescript/serve.py
  - lib/galaxy/jobs/runners/__init__.py + shell clse_fds, preexec_fn
  - lib/galaxy/jobs/runners/condor.py   + -"-
  - lib/galaxy/jobs/runners/local.py    + -"-
  - lib/galaxy/jobs/runners/util/cli/shell/local.py + -"-
- [ ] use shell=True
  - lib/galaxy/tool_util/deps/container_classes.py
  - lib/galaxy/tool_util/verify/test_data.py (because it uses strings and command chaining with &&)
  - lib/galaxy/job_metrics/collectl/cli.py
- [ ] read ony begin of stdout
  - lib/galaxy/tool_shed/galaxy_install/migrate/check.py
[...]

PR #9437, opened 2020-02-25
[19.09] Fix `SearchGuiArchive.set_meta()`

bernt-matthias:
probably with py3 reading a file from a zip file became binary. hence `startswith` and `split` fail. 

Q is if this is the best possible fix? 

- Maybe use binary-strings for `split` and `startswith`?
- Or is there something is galaxy.utils that could help?

I target 19.09 because this is the version I'm currently running on. Could be changed to 20.01 and I cherry pick for my instance.

### DIFF
--- a/lib/galaxy/datatypes/binary.py
+++ b/lib/galaxy/datatypes/binary.py
@@ -2111,8 +2111,7 @@ class SearchGuiArchive(CompressedArchive):
                 with zipfile.ZipFile(dataset.file_name) as tempzip:
                     if 'searchgui.properties' in tempzip.namelist():
                         with tempzip.open('searchgui.properties') as fh:
-                            for line in fh:
-                                line = line.decode()
+                            for line in io.TextIOWrapper(fh):
                                 if line.startswith('searchgui.version'):
                                     version = line.split('=')[1].strip()
                                     dataset.metadata.searchgui_version = version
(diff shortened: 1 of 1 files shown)


jdavcs:
I think this is the simplest solution, and it works for both Py2 and Py3. Given the subsequent string manipulation, text makes more sense than bytes. +1. Needs to be merged forward, I believe.

jdavcs (approved):
unit test failure is unrelated.

lib/galaxy/datatypes/binary.py
```diff
@@ -2112,6 +2112,7 @@ def set_meta(self, dataset, overwrite=True, **kwd):
                     if 'searchgui.properties' in tempzip.namelist():
                         with tempzip.open('searchgui.properties') as fh:
                             for line in fh:
```
```suggestion
                            for line in io.TextIOWrapper(fh):
```

lib/galaxy/datatypes/binary.py
```diff
@@ -2112,6 +2112,7 @@ def set_meta(self, dataset, overwrite=True, **kwd):
...
+                                line = line.decode()
```
```suggestion
```

bernt-matthias:
Thanks @nsoranzo for the suggestion. Do you think that this also might be the preferred solution for https://github.com/galaxyproject/galaxy/issues/9251 .. instead of `unicodify`?

nsoranzo:
> Thanks @nsoranzo for the suggestion. Do you think that this also might be the preferred solution for #9251 .. instead of `unicodify`?

I think `unicodify()` is fine for that, since stdout and stderr are plain strings.

nsoranzo (approved):

nsoranzo (changes requested):
Ah, you need to add an `import io` at the top.

nsoranzo (approved):

PR #9438, opened 2020-02-25
iRODS object store rewrite

kxk302:
Renamed rods.py to irods.py. Used the new client code (python_irodsclient) instead of deprecated PyRods code. There is an integration tests that tests the client code API against a local dockerized iRODS instance. test_datatype_upload integration test, which loads file into Galaxy passes.


test/integration/objectstore/test_irods.py
```diff
@@ -0,0 +1,177 @@
...
+    @classmethod
+    def tearDownClass(cls): 
+        print('Starting tearDownClass')
+        super(IrodsObjectStoreIntegrationTestCase, cls).tearDownClass()
+        IrodsObjectStoreIntegrationTestCase.stop_irods_server(container_name=IrodsObjectStoreIntegrationTestCase.container_name)       
+        print('Finishing tearDownClass')
+
+    @classmethod
+    def start_irods_server(cls, container_name, postgres_port=5432, irods_port=1247, docker_image='kxk302/irods-server:0.1'):
+        start_server = ['docker', 'run', '-h', 'localhost', '-p', '{port}:5432'.format(port=postgres_port), '-p', '{port}:1247'.format(port=irods_port), '-d', '--name', container_name, docker_image]
```
mvdbeek: Why do you need to expose postgres to the host ? This is going to collide with https://github.com/galaxyproject/galaxy/blob/dev/.github/workflows/integration.yaml#L22

nsoranzo:
Are you sure that the license for `test/integration/objectstore/irods_beginner_training_2016.pdf` allows redistribution? Maybe use one of the existing files in test-data/

lib/galaxy/objectstore/irods.py
```diff
@@ -0,0 +1,710 @@
...
+        self.session.cleanup()
+
+    def _initialize(self):
+        if irods is None:
+            raise Exception(IRODS_IMPORT_MESSAGE)
+
+        self.home =  "/" + self.zone + "/home/" + self.username
+
+        self.session = self._configure_connection(host=self.host, port=self.port, user=self.username, password=self.password, zone=self.zone)
+        # Clean cache only if value is set in galaxy.ini
```
mvdbeek:
```suggestion
        # Clean cache only if value is set in galaxy config file
```

lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
```diff
@@ -43,6 +43,7 @@ pytest-metadata==1.8.0
 pytest-postgresql==1.4.1
 pytest-pythonpath==0.7.3
 pytest==4.6.9
+python-irodsclient==0.8.2
```
nsoranzo, line 47: I think this should be added to lib/galaxy/dependencies/conditional-requirements.txt (there's a note about PyRods already).

lib/galaxy/objectstore/irods.py
```diff
@@ -0,0 +1,710 @@
...
+        if self.cache_size != -1:
+            # Convert GBs to bytes for comparison
+            self.cache_size = self.cache_size * 1073741824
+            # Helper for interruptable sleep
+            self.sleeper = Sleeper()
```
mvdbeek: Is it a problem if multiple threads and processes execute this independently ?

lib/galaxy/objectstore/irods.py
```diff
@@ -0,0 +1,710 @@
...
+    def parse_xml(cls, config_xml):
+        return parse_config_xml(config_xml)
+
+    def to_dict(self):
+        as_dict = super(IRODSObjectStore, self).to_dict()
+        as_dict.update(self._config_to_dict())
+        return as_dict
+
+    def __cache_monitor(self):
+        time.sleep(2)  # Wait for things to load before starting the monitor
```
mvdbeek: What things are this ? Can we have a smarter check here ?

lib/galaxy/objectstore/irods.py
```diff
@@ -0,0 +1,710 @@
...
+            return False
+
+        log.debug("Pulled data object '%s' into cache to %s", rel_path, self._get_cache_path(rel_path))
+        
+        with data_obj.open('rb') as data_obj_fp, open(self._get_cache_path(rel_path), "wb") as cache_fp:
+            content = data_obj_fp.read()
+            cache_fp.write(content)
+        return True
+
+    def _push_to_os(self, rel_path, source_file=None, from_string=None):
```
mvdbeek: Since this is pushing to irods maybe `push_to_irods` is a better name ?

lib/galaxy/objectstore/irods.py
```diff
@@ -0,0 +1,710 @@
...
+        if irods is None:
+            raise Exception(IRODS_IMPORT_MESSAGE)
+
+        self.home =  "/" + self.zone + "/home/" + self.username
+
+        self.session = self._configure_connection(host=self.host, port=self.port, user=self.username, password=self.password, zone=self.zone)
+        # Clean cache only if value is set in galaxy.ini
+        if self.cache_size != -1:
+            # Convert GBs to bytes for comparison
+            self.cache_size = self.cache_size * 1073741824
```
mvdbeek: We have some helper utilities in https://github.com/mvdbeek/galaxy/blob/a1f9ae28ec64540056aa29eba329b84f714fb8b3/lib/galaxy/util/bytesize.py#L47 that might be a good fit for dealing with byte units.

lib/galaxy/objectstore/irods.py
```diff
@@ -0,0 +1,710 @@
...
+                try:
+                    if source_file != cache_file:
+                        # FIXME? Should this be a `move`?
+                        shutil.copy2(source_file, cache_file)
+                    self._fix_permissions(cache_file)
+                except OSError:
+                    log.exception("Trouble copying source file '%s' to cache '%s'", source_file, cache_file)
+            else:
+                source_file = self._get_cache_path(rel_path)
+            # Update the file on S3
```
mvdbeek:
```suggestion
            # Update the file in iRODS
```

lib/galaxy/objectstore/irods.py
```diff
@@ -0,0 +1,710 @@
...
+                return True
+            # for JOB_WORK directory
+            elif base_dir:
+                if not os.path.exists(rel_path):
+                    os.makedirs(rel_path)
+                return True
+            else:
+                return False
+
+        # TODO: Sync should probably not be done here. Add this to an async upload stack?
```
mvdbeek: Right, exists should be fast since it is tried often and many operations depend on it being reasonably fast.
Is it really necessary to do this here, or does this mostly concern files that will be transferred later in `update_from_file`

lib/galaxy/objectstore/irods.py
mvdbeek:
```suggestion
            # Update the file in iRODS
```

lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
kxk302: Thx, will move it there.

kxk302: The file has this comment "These dependencies are only required when certain config options are set".  On a second thought, should irods be moved there or stay?

lib/galaxy/objectstore/irods.py
kxk302: Done

lib/galaxy/objectstore/irods.py
kxk302: Great idea. Code re-use and one less magic number :)

lib/galaxy/objectstore/irods.py
kxk302: I am not sure. Per my chat with @jmchilton, I used s3.py as a template for irods.py. So, some of the code (including this cache cleaner) comes directly from there.

lib/galaxy/objectstore/irods.py
kxk302: Same thing here (I borrowed the caching code directly from s3.py. It works).

lib/galaxy/objectstore/irods.py
kxk302: Sure

lib/galaxy/objectstore/irods.py
kxk302: Done

kxk302: Done again :)

test/integration/objectstore/test_irods.py
kxk302: I will remove that port mapping then. Did that so I could connect to iRODS Postgres from host via psql, mostly to test/confirm that the dockerized iRODS instance is working fine.

lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
mvdbeek: If we have tests that utilize irods (yes) I think we should have it in both places.
`pinned-dev-requirements.txt` dependencies will be installed when running tests, the stuff in `conditional-requirements.txt` will be installed based on config values (in normal operation). The conditional dependency check doesn't work for dynamically constructed test config settings.

lib/galaxy/objectstore/irods.py
mvdbeek: It's a reasonable start, but I wouldn't assume that s3.py is perfect or even actively used.

lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
nsoranzo: For conditional requirements have a look at the pythos object store and its kamaki package dependency at https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/dependencies/__init__.py#L175

mvdbeek: Indeed, that is how you do it, but we still need it in lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt

lib/galaxy/objectstore/irods.py
kxk302: I agree that exists() should be light weight and fast. I'm trying to understand how something can be in cache and not in iRODS. Can remove this (or comment it out with a comment for the time being).

mvdbeek: Yeah, I think we can revisit that when it turns out to be necessary.

lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
kxk302: Ok, added it to conditional-requirements.txt  as well.

lib/galaxy/objectstore/irods.py
kxk302: Right. The caching mechanism seems to work, but I am not sure about cache cleanup. Will see if I can write a unit test for this.

lib/galaxy/objectstore/irods.py
kxk302: Ok. I'll add a comment to that effect.

kxk302: Never mind, there already is a comment there!

lib/galaxy/objectstore/irods.py
dannon: We did talk a while back about moving all of the object store caching into a unified module with a global cache.  If you can, along the way here, identify how that refactoring should happen and maybe even do it that'd be a nice addition over duplicated code.

(this would be a parent abstraction `CachingObjectStore` or the like)

.gitignore
```diff
@@ -79,7 +79,6 @@ shed_tool_data_table_conf.xml
 job_conf.xml
 data_manager_conf.xml
 shed_data_manager_conf.xml
-object_store_conf.xml
```
nsoranzo: Why this change?

lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
nsoranzo: @kxk302 Can you also add `python-irodsclient = "*"` to the `[dev-packages]` section of `lib/galaxy/dependencies/pipfiles/default/Pipfile` and run `make update-dependencies` ?

This should add the required `xmlrunner` to `lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt` (though there may be unrelated changes to the `*requirements.txt` files you can ignore.

.gitignore
kxk302: In order to test irods, I need to override the objectstore to be irods.  I need this file checked in (which has irods details).I guess I could re-name the file, if this must remain in .gitignore.

lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
kxk302: @nsoranzo Added the line and ran the command. It changed many versions in the following files:

lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
lib/galaxy/dependencies/pipfiles/flake8/pinned-requirements.txt

Not sure if these changes are valid (or have side effects)

.gitignore
nsoranzo: You can use `git add --force PATH` to check in a file that would be normally ignored. After that, git will take into account modifications to it in `git status` and so on.

kxk302: Correct. I think I will undo the change to .gitignore file and rename the config file to irods_object_store_conf.xml instead.

lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
nsoranzo: You can use `git add -u -p lib/galaxy/dependencies/pipfiles/` to select only the relevant changes (xmlrunner line added to `lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt` , I think), then `git checkout -- lib/galaxy/dependencies/pipfiles/` to get rid of the unwanted changes.

kxk302: Thanks. I got rid of un-wanted changes.

kxk302:
> Are you sure that the license for `test/integration/objectstore/irods_beginner_training_2016.pdf` allows redistribution? Maybe use one of the existing files in test-data/

Turns out that the test is not needed. Removed the PDF and the test.

lib/galaxy/objectstore/irods.py
```diff
@@ -0,0 +1,674 @@
...
+            return False 
+
+        if self.cache_size > 0 and data_obj.__sizeof__() > self.cache_size:
+            log.critical("File %s is larger (%s) than the cache size (%s). Cannot download.",
+                         rel_path, data_obj.__sizeof__(), self.cache_size)
+            return False
+
+        log.debug("Pulled data object '%s' into cache to %s", rel_path, self._get_cache_path(rel_path))
+        
+        with data_obj.open('r') as data_obj_fp, open(self._get_cache_path(rel_path), "wb") as cache_fp:
```
mvdbeek, line 353: Do you read bytestrings or unicode with `data_obj.open('r').read()` ?  reading with `r` and writing with `wb` looks a little funny

test/integration/test_datatype_upload_irods.py
```diff
@@ -0,0 +1,40 @@
+import os
+import tempfile
+
+import pytest
+
+from galaxy.datatypes.registry import Registry
```
mvdbeek:
```suggestion
```

test/integration/test_datatype_upload_irods.py
```diff
@@ -0,0 +1,40 @@
+import os
+import tempfile
```
mvdbeek:
```suggestion
```

test/integration/test_datatype_upload_irods.py
```diff
@@ -0,0 +1,40 @@
...
+
+import pytest
+
+from galaxy.datatypes.registry import Registry
+from galaxy_test.driver import integration_util
+from .test_upload_configuration_options import BaseUploadContentConfigurationInstance
+from .test_datatype_upload import TEST_CASES
+from .test_datatype_upload import upload_datatype_helper
```
mvdbeek:
```suggestion
from .test_datatype_upload import (
    temp_file,
    TEST_CASES,
    upload_datatype_helper,
    UploadTestDatatypeDataTestCase,
```

test/integration/test_datatype_upload_irods.py
```diff
@@ -0,0 +1,40 @@
+import os
+import tempfile
+
+import pytest
+
+from galaxy.datatypes.registry import Registry
+from galaxy_test.driver import integration_util
+from .test_upload_configuration_options import BaseUploadContentConfigurationInstance
```
mvdbeek:
```suggestion
```

test/integration/test_datatype_upload_irods.py
```diff
@@ -0,0 +1,40 @@
+import os
```
mvdbeek, line 2:
```suggestion
```

test/integration/test_datatype_upload_irods.py
```diff
@@ -0,0 +1,40 @@
...
+from galaxy.datatypes.registry import Registry
+from galaxy_test.driver import integration_util
+from .test_upload_configuration_options import BaseUploadContentConfigurationInstance
+from .test_datatype_upload import TEST_CASES
+from .test_datatype_upload import upload_datatype_helper
+
+SCRIPT_DIRECTORY = os.path.abspath(os.path.dirname(__file__))
+GALAXY_ROOT = os.path.abspath('%s/../../' % SCRIPT_DIRECTORY)
+DATATYPES_CONFIG = os.path.join(GALAXY_ROOT, 'lib/galaxy/config/sample/datatypes_conf.xml.sample')
+PARENT_SNIFFER_MAP = {'fastqsolexa': 'fastq'}
```
mvdbeek:
```suggestion
```

test/integration/test_datatype_upload_irods.py
```diff
@@ -0,0 +1,40 @@
...
+OBJECT_STORE_CONFIG_FILE = os.path.join(SCRIPT_DIRECTORY, "irods_object_store_conf.xml")
+# Run test for only the first 10 test files
+IRODS_TEST_CASES = dict(list(TEST_CASES.items())[0:10])
+
+class UploadTestDatatypeDataTestCase(BaseUploadContentConfigurationInstance):
```
mvdbeek:
```suggestion
class UploadTestDatatypeDataTestCase(UploadTestDatatypeDataTestCase):
```

test/integration/test_datatype_upload_irods.py
```diff
@@ -0,0 +1,40 @@
...
+instance = integration_util.integration_module_instance(UploadTestDatatypeDataTestCase)
+
+@pytest.fixture
+def temp_file():
+    with tempfile.NamedTemporaryFile(delete=True, mode='wb') as fh:
+        yield fh
+
+registry = Registry()
+registry.load_datatypes(root_dir=GALAXY_ROOT, config=DATATYPES_CONFIG)
+
```
mvdbeek:
```suggestion
```

test/integration/test_datatype_upload_irods.py
```diff
@@ -0,0 +1,40 @@
...
+GALAXY_ROOT = os.path.abspath('%s/../../' % SCRIPT_DIRECTORY)
+DATATYPES_CONFIG = os.path.join(GALAXY_ROOT, 'lib/galaxy/config/sample/datatypes_conf.xml.sample')
+PARENT_SNIFFER_MAP = {'fastqsolexa': 'fastq'}
+OBJECT_STORE_CONFIG_FILE = os.path.join(SCRIPT_DIRECTORY, "irods_object_store_conf.xml")
+# Run test for only the first 10 test files
+IRODS_TEST_CASES = dict(list(TEST_CASES.items())[0:10])
+
+class UploadTestDatatypeDataTestCase(BaseUploadContentConfigurationInstance):
+    framework_tool_and_types = False
+    datatypes_conf_override = DATATYPES_CONFIG
```
mvdbeek:
```suggestion
```

lib/galaxy/objectstore/irods.py
```diff
@@ -0,0 +1,674 @@
...
+        #             exists = True
+        #         else:
+        #             # print("***2.2 %s exists but differs in size from cache (in_cache=False)" % cache_path)
+        #             exists = False
+        #     else:
+        #         # Although not perfect decision making, this most likely means
+        #         # that the file is currently being uploaded
+        #         # print("***3 %s found in cache but not in S3 (in_cache=True)" % cache_path)
+        #         exists = True
+        # else:
```
mvdbeek:
```suggestion
```

lib/galaxy/objectstore/irods.py
mvdbeek: I think that was a misunderstanding, what I meant is that this we can't push to the object store here. If additional syncing is necessary we can revisit this later if necessary,

test/integration/test_datatype_upload_irods.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+@pytest.fixture
+def temp_file():
+    with tempfile.NamedTemporaryFile(delete=True, mode='wb') as fh:
+        yield fh
+
+
+registry = Registry()
+registry.load_datatypes(root_dir=GALAXY_ROOT, config=DATATYPES_CONFIG)
```
mvdbeek:
```suggestion
```

test/integration/test_datatype_upload_irods.py
```diff
@@ -0,0 +1,46 @@
+import os
+import tempfile
+
+import pytest
+
+from galaxy.datatypes.registry import Registry
+from galaxy_test.driver import integration_util
+
+from .test_datatype_upload import TEST_CASES, upload_datatype_helper
```
mvdbeek:
```suggestion
from .test_datatype_upload import (
    temp_file,
    TEST_CASES,
    upload_datatype_helper,
    UploadTestDatatypeDataTestCase
)
```

test/integration/test_datatype_upload_irods.py
```diff
@@ -0,0 +1,46 @@
...
+from .test_upload_configuration_options import \
+    BaseUploadContentConfigurationInstance
```
mvdbeek:
```suggestion
```

lib/galaxy/objectstore/irods.py
mvdbeek: Looks good now, but can you drop the TODO statement in line 466 as well ?

test/integration/test_datatype_upload_irods.py
```diff
@@ -0,0 +1,31 @@
+
```
mvdbeek, line 1:
```suggestion
import os
```
Got a bit carried away here :)

kxk302: :)

.github/workflows/integration.yaml
```diff
@@ -48,4 +48,4 @@ jobs:
         key: pip-cache-${{ matrix.python-version }}-${{ hashFiles('galaxy root/requirements.txt') }}
     - name: Run tests
       run: './run_tests.sh -integration test/integration -- -k "${{ matrix.subset }}"'
-      working-directory: 'galaxy root'
+      working-directory: 'galaxy-root'
```
mvdbeek: You can just drop this line and line 39 to avoid the space. But we do need to have this fixed before merging.

lib/galaxy/objectstore/irods.py
```diff
@@ -0,0 +1,668 @@
...
+
+        self.home = "/" + self.zone + "/home/" + self.username
+
+        self.session = self._configure_connection(host=self.host, port=self.port, user=self.username, password=self.password, zone=self.zone)
+
+    def _configure_connection(self, host='localhost', port='1247', user='rods', password='rods', zone='tempZone'):
+        with iRODSSession(host=host, port=port, user=user, password=password, zone=zone) as session:
+            # Throws NetworkException if connection fails
+            try:
+                session.pool.get_connection()
```
mvdbeek, line 220: Is there a timeout option here ? This seems to hang forever if IRODS isn't quite up yet.

.github/workflows/integration.yaml
kxk302: I re-named them to galaxy in the subsequent push. When we have a fix for the bug (space in dir name), I can change it back to verify.

test/integration/test_datatype_upload_irods.py
```diff
@@ -0,0 +1,107 @@
...
+instance = integration_util.integration_module_instance(UploadTestDatatypeDataTestCase)
+
+
+@pytest.fixture
+def temp_file():
+    with tempfile.NamedTemporaryFile(delete=True, mode='wb') as fh:
+        yield fh
+
+
```
mvdbeek:
```suggestion
```

test/integration/test_datatype_upload_irods.py
```diff
@@ -0,0 +1,107 @@
...
+import os
+import string
+import subprocess
+import tempfile
+import time
+
+import pytest
+
+from galaxy_test.driver import integration_util
+from .test_datatype_upload import (TEST_CASES, upload_datatype_helper, UploadTestDatatypeDataTestCase)
```
mvdbeek:
```suggestion
from .test_datatype_upload import (
    temp_file
    TEST_CASES,
    upload_datatype_helper, 
    UploadTestDatatypeDataTestCase
)
```

but please also check if you can move the temp_file fixture to conftest.py file in `test/integration`. If you do that they will be injected as need and no import is necessary.

lib/galaxy/objectstore/irods.py
mvdbeek: Can we remove this ?

lib/galaxy/objectstore/irods.py
mvdbeek: I think we need some clarification here. If the tests pass with `rb` i think we should use that.

.github/workflows/converter_tests.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
       with:
-        path: 'galaxy root'
+        path: 'galaxy'
```
mvdbeek: Can you revert the changes to this file ?

.github/workflows/integration.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         python-version: [3.7]
-        subset: ['upload_datatype', 'extended_metadata', 'kubernetes', 'not (upload_datatype or extended_metadata or kubernetes)']
+        subset: ['upload_datatype_irods']
```
mvdbeek: Please revert the changes to this file

.github/workflows/mulled.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
       with:
-        path: 'galaxy root'
+        path: 'galaxy'
```
mvdbeek: Please revert the changes to this file

CONTRIBUTORS.md
```diff
@@ -1,6 +1,6 @@
 # Contributors
 
-The following individuals have contributed code to Galaxy:
+The following individuals have contributed code to Galaxy:
```
mvdbeek:
```suggestion
The following individuals have contributed code to Galaxy:
```

lib/galaxy/objectstore/irods.py
```diff
@@ -0,0 +1,676 @@
...
+                log.debug("Pushed data from string '%s' to collection '%s'", from_string, data_object_path)
+            else:
+                start_time = datetime.now()
+                log.debug("Pushing cache file '%s' of size %s bytes to collection '%s'", source_file, os.path.getsize(source_file), rel_path)
+
+                # Create sub-collection first
+                self.session.collections.create(collection_path, recurse=True)
+
+                file_content = None
+                with open(source_file, 'rb') as content_file:
```
jmchilton: This shouldn't all be in memory right? This will routinely read multi-gigabyte files in production I suspect. Can we set a buffer size and merge the loops into one?

mvdbeek: That applies to https://github.com/galaxyproject/galaxy/pull/9438/files#diff-b64aafa89fa13b4e704cc68ff34b2da1R373 as well.

.github/workflows/converter_tests.yaml
kxk302: Sure.

.github/workflows/integration.yaml
kxk302: Sure

.github/workflows/mulled.yaml
kxk302: Sure

lib/galaxy/objectstore/irods.py
kxk302: This code was deleted per our discussion.

lib/galaxy/objectstore/irods.py
kxk302: Confirmed that iRODS data objects are binary objects and 'r' flag is really a binary read.

test/integration/test_datatype_upload_irods.py
kxk302: Applied change.

test/integration/test_datatype_upload_irods.py
kxk302: Applied change.

test/integration/test_datatype_upload_irods.py
kxk302: Applied change.

lib/galaxy/objectstore/irods.py
kxk302: Removed locally. Part of my next push.

.github/workflows/integration.yaml
kxk302: Removed line 39 and 51 locally.

lib/galaxy/objectstore/irods.py
kxk302: @jmchilton Sure. Will do. Do you have an example buffered read in Galaxy that I can template this on?

kxk302: @mvdbeek your link brings me back to this PR! Could you please clarify what you are referring to? Thx

mvdbeek: Sorry, line 374 which should also use chunks.
To read/write in chunks do
```python
from functools import partial
...
CHUNK_SIZE = 2**20 # 1MB

...
with open(source_file, 'rb') as content_file, data_obj.open('w') as data_obj_fp:
    for chunk in iter(partial(content_file.read, CHUNK_SIZE), b''):
        data_obj_fp.write(chunk)
```

You can limit the number of bytes to read with an integer argument to `read`. You know you've reached the end of the file when read() returns nothing, so a more old-school implementation would be:
```python
chunk = content_file.read(CHUNK_SIZE)
while chunk:
    data_obj_fp.write(chunk)
    chunk = content_file.read(CHUNK_SIZE)

Comment (kxk302) [reply lib/galaxy/objectstore/irods.py]: Thanks!

File: lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt [hunk 35]
```diff
@@ -63,3 +64,4 @@ watchdog==0.10.2
 wcwidth==0.1.8
 wrapt==1.12.0
 zipp==1.2.0
+
```
mvdbeek:
```suggestion
```

lib/galaxy/objectstore/irods.py
```diff
@@ -0,0 +1,653 @@
+"""
+Object Store plugin for the Integrated Rule-Oriented Data Store (iRODS)
+"""
+from functool import partial
```
mvdbeek:
```suggestion
```

lib/galaxy/objectstore/irods.py
```diff
@@ -0,0 +1,653 @@
...
+import logging
+import os
+import shutil
+from datetime import datetime
+try:
```
mvdbeek, line 9:
```suggestion
from functools import partial
try:
```

lib/galaxy/objectstore/irods.py
kxk302: Added timeout to config.

.github/workflows/integration.yaml
mvdbeek: Can you restore the tests please ?

test/integration/test_datatype_upload_irods.py
mvdbeek: Can you push the temp_file fixture to a conftest.py file please ? https://docs.pytest.org/en/latest/fixture.html#conftest-py-sharing-fixture-functions

.github/workflows/integration.yaml
kxk302: Hmm, somehow I missed this. Pushing it next.

.github/workflows/integration.yaml
```diff
@@ -35,8 +35,6 @@ jobs:
       run: |
         kubectl get pods
     - uses: actions/checkout@v2
-      with:
```
mvdbeek: Please restore this file from the current dev branch, e.g with `git checkout origin/dev -- .github/workflows/integration.yaml`.

kxk302: will do.

mvdbeek: In that case I'd suggest not marking this as resolved until you've pushed the change, that makes it easier to keep an overview for reviewer and author of what is left to do.

kxk302: Sure

mvdbeek:
@galaxybot test this

mvdbeek (commented):
There's a new empty file called `continue`, ca you `git rm` that ?
Can you also add this patch?
```
diff --git a/lib/galaxy/dependencies/__init__.py b/lib/galaxy/dependencies/__init__.py
index 328e553cb7..0d51436ebf 100644
--- a/lib/galaxy/dependencies/__init__.py
+++ b/lib/galaxy/dependencies/__init__.py
@@ -175,6 +175,9 @@ class ConditionalDependencies(object):
     def check_kamaki(self):
         return 'pithos' in self.object_stores
 
+    def check_python_irodsclient(self):
+        return "irods" in self.object_stores
+
     def check_watchdog(self):
         install_set = {'auto', 'True', 'true', 'polling'}
         return (self.config['watch_tools'] in install_set or

```
That will make sure python-irodsclient will be installed if an irods objectstore is configured. (On stable releases and outside of a testing context we don't install dependencies from `pinned-dev-requirements.txt`)

After that it is ready to go IMO.

test/integration/test_datatype_upload_irods.py
kxk302: Done.

mvdbeek:
Awesome, thanks so much @kxk302, that is a great start!

lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
nsoranzo: prettytable and xmlrunner should also have been added here, will be fixed in https://github.com/galaxyproject/galaxy/pull/9529

PR #9439, opened 2020-02-25
Unify default values used for delta (sim_size)

bernt-matthias:
cherry picked from https://github.com/galaxyproject/galaxy/pull/9425

TODOs:

- [x] verify that 10000 is really the default currently used 
- [x] apply `int()` also to yaml parsing

### DIFF
--- a/lib/galaxy/tool_util/parser/xml.py
+++ b/lib/galaxy/tool_util/parser/xml.py
@@ -9,7 +9,7 @@ from math import isinf
 import packaging.version
 
 from galaxy.tool_util.deps import requirements
-from galaxy.tool_util.parser import DEFAULT_DELTA
+from galaxy.tool_util.parser.util import DEFAULT_DELTA
 from galaxy.util import string_as_bool, xml_text, xml_to_string
 from .interface import (
     InputSource,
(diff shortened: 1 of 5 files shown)


lib/galaxy/tool_util/parser/xml.py
```diff
@@ -613,6 +614,7 @@ def __parse_test_attributes(output_elem, attrib, parse_elements=False, parse_dis
     attributes['lines_diff'] = int(attrib.pop('lines_diff', '0'))
     # Allow a file size to vary if sim_size compare
     attributes['delta'] = int(attrib.pop('delta', '10000'))
```
```suggestion
```

lib/galaxy/tool_util/parser/yaml.py
```diff
@@ -1,8 +1,8 @@
 from collections import OrderedDict
 
 import packaging.version
-
```
nsoranzo, line 4: The empty line need to stay to separate third-party from "internal" imports

nsoranzo:
Can you also add (copy/paste is fine) the default value to lib/galaxy/tool_util/xsd/galaxy.xsd

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -1314,7 +1314,7 @@ When this attribute is true and ``compare`` is set to ``diff``, try to decompres
     </xs:attribute>
     <xs:attribute name="delta" type="xs:integer">
       <xs:annotation>
-        <xs:documentation xml:lang="en">If ``compare`` is set to ``sim_size``, this is the number of bytes different allowed.</xs:documentation>
+        <xs:documentation xml:lang="en">If ``compare`` is set to ``sim_size``, this is the number of bytes different allowed (default 10.000).</xs:documentation>
```
nsoranzo: 10.000 is the same as 10 in English :D
```suggestion
        <xs:documentation xml:lang="en">If ``compare`` is set to ``sim_size``, this is the number of bytes different allowed (default 10000).</xs:documentation>
```

PR #9442, opened 2020-02-25
[20.01] Fix data tables vue display

dannon:
This restores the data tables display grid in the admin panel.  Previously no entries would render.

xref https://gitter.im/galaxyproject/dev?at=5e55953b8f8af6553a03a863


natefoo:
Thanks @dannon !

PR #9377, opened 2020-02-14
Make mutable_config_dir configurable + rename

jdavcs:
As per recent discussions on gitter. Ref #9362 
I haven't added any new tests or safeguards - it seems to be a straightforward change after all. Unless I'm missing something. 
Ping @natefoo @mvdbeek @pvanheus 

EDIT: I chose `managed_config_dir`, but I'm open to `galaxy_managed_config_dir` (or other suggestions).

Issue #9362, opened 2020-02-11
Make mutable_config_dir configurable


jdavcs:
There's been a discussion on https://gitter.im/galaxyproject/admins on this (thank you, @pvanheus, for bringing this up!); the general consensus has been to make the change. This should include adding `mutable_config_dir` as an option to the schema, setting the `path_resolves_to` property for the options that are resolved w.r.t `mutable_config_dir`, updating the `GalaxyAppConfiguration` class accordingly, and adding/modifying relevant tests. Most likely, as a result of this modification, ansible galaxy will need to be updated as well.

jdavcs:
TODO: after 9377 is merged, ansible-galaxy needs to be updated.

jdavcs:
Closing this as completed. Feel free to reopen if I missed anything.

PR #9444, opened 2020-02-26
Make mutable_config_dir configurable + rename

jdavcs:
This replaces #9377 
As per recent discussions on gitter. Ref #9362
I haven't added any new tests or safeguards - it seems to be a straightforward change after all. Unless I'm missing something. I chose managed_config_dir, but I'm open to galaxy_managed_config_dir (or other suggestions).

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -147,7 +147,7 @@ class BaseAppConfiguration(object):
                 if self.config_dir is None:
                     self.config_dir = os.getcwd()
                 if self.data_dir is None:
-                    self.data_dir = os.path.join(self.config_dir, 'data')
+                    self.data_dir = self._in_config_dir('data')
                 if self.managed_config_dir is None:
                     self.managed_config_dir = self._in_data_dir('config')
 
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -149,7 +149,7 @@ class BaseAppConfiguration(object):
                 if self.data_dir is None:
                     self.data_dir = os.path.join(self.config_dir, 'data')
                 if self.managed_config_dir is None:
-                    self.managed_config_dir = os.path.join(self.data_dir, 'config')
+                    self.managed_config_dir = self._in_data_dir('config')
 
             # TODO: do we still need to support ../shed_tools when running_from_source?
             self.shed_tools_dir = self._in_data_dir('shed_tools')
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -152,7 +152,7 @@ class BaseAppConfiguration(object):
                     self.managed_config_dir = os.path.join(self.data_dir, 'config')
 
             # TODO: do we still need to support ../shed_tools when running_from_source?
-            self.shed_tools_dir = os.path.join(self.data_dir, 'shed_tools')
+            self.shed_tools_dir = self._in_data_dir('shed_tools')
 
             log.debug("Configuration directory is %s", self.config_dir)
             log.debug("Data directory is %s", self.data_dir)
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -379,7 +379,7 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
         self.oidc_config = kwargs.get("oidc_config_file", self.oidc_config_file)
         self.oidc_backends_config = kwargs.get("oidc_backends_config_file", self.oidc_backends_config_file)
         self.oidc = []
-        self.integrated_tool_panel_config = os.path.join(self.managed_config_dir, self.integrated_tool_panel_config)
+        self.integrated_tool_panel_config = self._in_managed_config_dir(self.integrated_tool_panel_config)
         integrated_tool_panel_tracking_directory = kwargs.get('integrated_tool_panel_tracking_directory')
         if integrated_tool_panel_tracking_directory:
             self.integrated_tool_panel_tracking_directory = os.path.join(self.root, integrated_tool_panel_tracking_directory)
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -135,6 +135,8 @@ class BaseAppConfiguration(object):
             self.data_dir = config_kwargs.get('data_dir')
             self.sample_config_dir = os.path.join(os.path.dirname(__file__), 'sample')
             self.managed_config_dir = config_kwargs.get('managed_config_dir')
+            if self.managed_config_dir:
+                self.managed_config_dir = os.path.abspath(self.managed_config_dir)
 
             if running_from_source:
                 if not self.config_dir:
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -133,37 +133,36 @@ class BaseAppConfiguration(object):
                 self.config_dir = os.path.abspath(self.config_dir)
 
             self.data_dir = config_kwargs.get('data_dir')
-            # mutable_config_dir is intentionally not configurable. You can
-            # override individual mutable configs with config options, but they
-            # should be considered Galaxy-controlled data files and will by default
-            # just live in the data dir
             self.sample_config_dir = os.path.join(os.path.dirname(__file__), 'sample')
+            self.managed_config_dir = config_kwargs.get('managed_config_dir')
 
             if running_from_source:
                 if self.config_dir is None:
                     self.config_dir = os.path.join(self.root, 'config')
                 if self.data_dir is None:
                     self.data_dir = os.path.join(self.root, 'database')
-                self.mutable_config_dir = self.config_dir
+                if self.managed_config_dir is None:
+                    self.managed_config_dir = self.config_dir
             else:
                 if self.config_dir is None:
                     self.config_dir = os.getcwd()
                 if self.data_dir is None:
                     self.data_dir = os.path.join(self.config_dir, 'data')
-                self.mutable_config_dir = os.path.join(self.data_dir, 'config')
+                if self.managed_config_dir is None:
+                    self.managed_config_dir = os.path.join(self.data_dir, 'config')
 
             # TODO: do we still need to support ../shed_tools when running_from_source?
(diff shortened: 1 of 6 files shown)


lib/galaxy/config/__init__.py
```diff
@@ -133,37 +133,36 @@ def _set_config_directories():
                 self.config_dir = os.path.abspath(self.config_dir)
 
             self.data_dir = config_kwargs.get('data_dir')
-            # mutable_config_dir is intentionally not configurable. You can
-            # override individual mutable configs with config options, but they
-            # should be considered Galaxy-controlled data files and will by default
-            # just live in the data dir
             self.sample_config_dir = os.path.join(os.path.dirname(__file__), 'sample')
+            self.managed_config_dir = config_kwargs.get('managed_config_dir')
```
nsoranzo, line 137:
```suggestion
            self.managed_config_dir = config_kwargs.get('managed_config_dir')
            if self.managed_config_dir:
                self.managed_config_dir = os.path.abspath(self.managed_config_dir)
```

lib/galaxy/config/__init__.py
```diff
@@ -133,37 +133,36 @@ def _set_config_directories():
...
                 if self.config_dir is None:
                     self.config_dir = os.getcwd()
                 if self.data_dir is None:
                     self.data_dir = os.path.join(self.config_dir, 'data')
-                self.mutable_config_dir = os.path.join(self.data_dir, 'config')
+                if self.managed_config_dir is None:
+                    self.managed_config_dir = os.path.join(self.data_dir, 'config')
 
             # TODO: do we still need to support ../shed_tools when running_from_source?
             self.shed_tools_dir = os.path.join(self.data_dir, 'shed_tools')
```
nsoranzo: Not sure if you like the helper methods or not, just trying to be consistent, feel free to disregard.

```suggestion
            self.shed_tools_dir = self._in_data_dir('shed_tools')
```

lib/galaxy/config/__init__.py
```diff
@@ -133,37 +133,36 @@ def _set_config_directories():
...
+                if self.managed_config_dir is None:
+                    self.managed_config_dir = self.config_dir
             else:
                 if self.config_dir is None:
                     self.config_dir = os.getcwd()
                 if self.data_dir is None:
                     self.data_dir = os.path.join(self.config_dir, 'data')
-                self.mutable_config_dir = os.path.join(self.data_dir, 'config')
+                if self.managed_config_dir is None:
+                    self.managed_config_dir = os.path.join(self.data_dir, 'config')
```
```suggestion
                    self.managed_config_dir = self._in_data_dir('config')
```

lib/galaxy/config/__init__.py
```diff
@@ -133,37 +133,36 @@ def _set_config_directories():
...
                 if self.data_dir is None:
                     self.data_dir = os.path.join(self.root, 'database')
-                self.mutable_config_dir = self.config_dir
+                if self.managed_config_dir is None:
+                    self.managed_config_dir = self.config_dir
             else:
                 if self.config_dir is None:
                     self.config_dir = os.getcwd()
                 if self.data_dir is None:
                     self.data_dir = os.path.join(self.config_dir, 'data')
```
```suggestion
                    self.data_dir = self._in_config_dir('data')
```

lib/galaxy/config/__init__.py
```diff
@@ -380,7 +379,7 @@ def _process_config(self, kwargs):
         self.oidc_config = kwargs.get("oidc_config_file", self.oidc_config_file)
         self.oidc_backends_config = kwargs.get("oidc_backends_config_file", self.oidc_backends_config_file)
         self.oidc = []
-        self.integrated_tool_panel_config = os.path.join(self.mutable_config_dir, self.integrated_tool_panel_config)
+        self.integrated_tool_panel_config = os.path.join(self.managed_config_dir, self.integrated_tool_panel_config)
```
```suggestion
        self.integrated_tool_panel_config = self._in_managed_config_dir(self.integrated_tool_panel_config)
```

lib/galaxy/config/__init__.py
```diff
@@ -133,37 +133,36 @@ def _set_config_directories():
...
             self.sample_config_dir = os.path.join(os.path.dirname(__file__), 'sample')
+            self.managed_config_dir = config_kwargs.get('managed_config_dir')
 
             if running_from_source:
                 if self.config_dir is None:
                     self.config_dir = os.path.join(self.root, 'config')
                 if self.data_dir is None:
                     self.data_dir = os.path.join(self.root, 'database')
-                self.mutable_config_dir = self.config_dir
+                if self.managed_config_dir is None:
```
nsoranzo: Question: if the admin sets `managed_config_dir` to the empty string `''`, it is neither resolved to an absolute path (even with my suggestion above, which follows what is done for `config_dir`) nor overridden here.

It's not clear what an admin would intend, but I think we should decide how to interpret that.

The same applies for `config_dir`, `data_dir`.

lib/galaxy/config/__init__.py
jdavcs: Yep, consistent and better.

lib/galaxy/config/__init__.py
jdavcs: I will apply this for now, but I plan to get rid of such cases by setting `path_resolves_to` in the schema - there are quite a few of those, and more..

I don't want to move on with those until #9371 is in - otherwise I'd have to add more changes to it, which are not directly related to its main theme, and I want to avoid that and keep it more or less atomic. It's not an urgent PR, but it's holding up a bunch of modifications I have in mind.. Sorry for being such a pest, but could you take a look at it, please?

lib/galaxy/config/__init__.py
jdavcs: I think setting a config option to `''` is a valid use case. That's how an admin would override the schema default if they wanted no value to be set. It will indeed break things if the `config_dir` path is set to `''`, but so would other unreasonable settings. Besides, in this case galaxy would fail fast and loudly: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L839

So maybe just leave as is?

Otherwise, one way to handle it, I think, would be to change these lines: https://github.com/galaxyproject/galaxy/blob/823526a1fc6c14c868f473f067e68c37ecd5197a/lib/galaxy/config/__init__.py#L140 (6 total) from 
`if self.foo is None:` 
to 
`if not self.foo:`

So, we treat `''` same as "not set", and set it to defaults.
What do you think?

lib/galaxy/config/__init__.py
jdavcs: If those options are NOT set by the admin (`managed_config_dir`, `data_dir`, etc.), they are eventually resolved w.r.t. an absolute path. If they ARE set by an admin, I think I would leave it up to the admin. There are many config options that are paths, and if some are made absolute by the system, while others are not, that may be unexpected behavior to an admin, I think? (all paths are equal, but some are more equal than the rest..)

nsoranzo: I think you are making an argument for making all paths absolute...

lib/galaxy/config/__init__.py
nsoranzo: I have to work on training material for the Galaxy Admin Training next week, feel free to have it reviewed by another committer!

lib/galaxy/config/__init__.py
natefoo: It's ok if the admin sets it relative but is there harm in making it absolute internally?

lib/galaxy/config/__init__.py
nsoranzo: Changing to `if not` would also be my preferred solution.

jdavcs: OK, can do.

lib/galaxy/config/__init__.py
jdavcs: Thanks! Will do.

lib/galaxy/config/__init__.py
jdavcs: OK, no quick answer. I'm thinking of (a) ways it may break something and (b) at what point it would be safe to do so. For example, we want to call `strip_deprecated_dir` BEFORE we make user-set path absolute. There may be other issues like this one. Thinking...

nsoranzo: Most user-specified paths are made absolute when we call `_in_data_dir()` or sister methods on them, given that `self.data_dir` &c. are already absolute.

jdavcs: Exactly. If the user sets `foopath='foo'`, and then it's passed to `_in_data_dir()`, you'll get something like `[abs data_dir path]/foo`. But if we *first* make that an abs path, then `_in_data_dir()` won't affect it and it'll remain `[abs foo path]`. So there will be no way for a user to set a dir to 'foo' and have galaxy place it inside its `data_dir`. In other words, I'm not sure we are not trying to assume too much about user-set paths here.

nsoranzo: I'm not advocating for any other changes apart for this one, where the value comes directly from the admin.

natefoo: @ic4f does `managed_config_dir` get passed through `_in_data_dir()` later? If so then I suppose it renders @nsoranzo's suggestion moot, unless I misunderstand?

jdavcs: it doesn't - I'm currently testing different config setups. So far it appears that if the admin sets it to some value that is a relative str, it stays relative.

jdavcs: Double checking, but I think we, indeed, may want to make it absolute - most likely along with `data_dir` and `config_dir`. I'm still checking this..

jdavcs: OK, so I'm making the change based on this premise: *if a config option is set by the user to a RELATIVE path, it must be ALWAYS resolved w.r.t some absolute directory by galaxy*.
I'll make the change for `managed_config_dir` here and will address the rest in a separate PR - not to convolute this one. I'll also document this (along with other config-related dev guidelines) - as soon as the rest of pending stuff in config is mostly done.

nsoranzo (approved):
Fantastic, thanks for bearing with me!

jdavcs:
@nsoranzo well, you were right :-)
Thank you for approving!

jdavcs:
I'm merging this: test failure (py27_unit: `jobs/test_expression_run.py`) is unrelated.

PR #9446, opened 2020-02-26
[20.01] Fix `UserManager.get_reset_token()` after #9432

nsoranzo:
Remove `lower()` also from the right side.


mvdbeek (approved):
Thanks a lot!

PR #9455, opened 2020-02-27
Move sanitize_whitelist into mutable_config_dir

jdavcs:
As per recent discussion on gitter.
Rationale: keep consistent with galaxy helm chart setup (see https://gitter.im/galaxyproject/dev?at=5e5804e25c32162f5ba51485 ) 
Ping @almahmoud @dannon

### DIFF
--- a/doc/source/admin/galaxy_options.rst
+++ b/doc/source/admin/galaxy_options.rst
@@ -2150,7 +2150,9 @@
     this file are trusted and will not have their HTML sanitized on
     display.  This can be manually edited or manipulated through the
     Admin control panel -- see "Manage Display Whitelist"
-:Default: ``config/sanitize_whitelist.txt``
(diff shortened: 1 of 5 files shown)


dannon:
@ic4f Is there a migration path for this or will existing `santize_whitelist.txt` files (whether in the old default `/config` or elsewhere) just keep on working?

jdavcs:
@dannon I haven't thought of this; thanks for pointing out! The previous default should work, of course (with a deprecation warning).  I'll add this in the next commit.

jdavcs:
Checking for an old (deprecated) value of `sanitize_whitelist_file` (and using it + outputting a warning message) should be factored out into a helper method in `BaseAppConfiguration`. However, I don't want to make such a change until we need it for at least one more analogous case. Premature optimization and all that.

PR #9456, opened 2020-02-27
[20.01] Fix newlines in final release notes

mvdbeek:


dannon:
@mvdbeek Did we remove these in previous release notes, too, or is something wrong with the rendering where we should be generating subheadings for them?

mvdbeek:
I don't see them in https://github.com/galaxyproject/galaxy/blob/dev/doc/source/releases/19.09.rst so I guess we did remove them before

PR #9458, opened 2020-02-28
Fix `test_run_simple()` test broken on Python 2 by virtualenv 20.0.5

nsoranzo:
xref. pypa/virtualenv#1670

Example of failing py27-unit build on CircleCI: https://app.circleci.com/jobs/github/galaxyproject/galaxy/73395

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,7 +95,9 @@ jobs:
     <<: *set_workdir
     steps:
       - *restore_repo_cache
-      - *install_tox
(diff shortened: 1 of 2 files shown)


jdavcs:
Just to confirm: it works locally for me too.

nsoranzo:
That's another virtualenv issue: https://github.com/pypa/virtualenv/issues/1670

nsoranzo:
Ready to merge!
Or should we just drop all py27* test jobs on `dev`?

mvdbeek:
IMO yes, but let's have that discussion in another PR ? I think that effectively means we stop supporting Python 2.

Issue #9459, opened 2020-02-28
i18n not work in v20.01


shenweiyan:
It seems i18n is failed in v20.01, I can't see any translation available for Simplified Chinese on https://usegalaxy.org/ !

dannon:
Should be fixed in the latest 20.01.

PR #9460, opened 2020-02-28
[20.01] Use browser locale over global default

mvdbeek:
Not 100% sure this is correct, but it seems logical to me.
Should fix https://github.com/galaxyproject/galaxy/issues/9459.

### DIFF
--- a/client/galaxy/scripts/app/galaxy.js
+++ b/client/galaxy/scripts/app/galaxy.js
@@ -209,7 +209,7 @@ GalaxyApp.prototype._initUserLocale = function _initUserLocale(options) {
             ? "__root"
             : (navigator.language || navigator.userLanguage || "__root").toLowerCase();
 
-    const locale = user_locale || global_locale || nav_locale;
+    const locale = user_locale || nav_locale || global_locale;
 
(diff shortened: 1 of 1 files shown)


dannon:
@mvdbeek The change looks reasonable to me, but this code hasn't changed in a long time, and i18n is working fine for me.  When I change browser locale to `zh`, for example, it works fine.  Are you seeing it fail for you, too?

mvdbeek:
Yes, if I set firefox to french and go in anonymous mode to usegalaxy.org I don't see french. Same with a local Galaxy instance.

dannon:
@mvdbeek I'm perplexed.  
image


(doing a little more debugging to see what's up)

mvdbeek:
Hmm, can you try this locally and check if you don't have anything in `user_locale`, which would take precedence ?

dannon:
@mvdbeek I think there still *might* be a race condition, otherwise I wouldn't be seeing what I did in the screenshot above (anonymous user, incognito browser, langauge set to `fr`).  But this does adjust priorities correctly IMO, and it fixes the one bug related to the swap from `null` to `'auto'` as the default value.

PR #9464, opened 2020-02-29
Merge 20.01 forward and rerun make update-dependencies

dannon:
Had to resolve a conflict in the merge forward, and then reran dependency update to make sure it was all right.


dannon:
Huh, I'll have to look into the test failures here when I have a sec.

nsoranzo:
I've restored the pin for gxformat2 to 0.10.1, it seems the new 0.11.1 version breaks Galaxy, ping @jmchilton . Any way, I'd say solving this is for another PR.

nsoranzo (approved):

dannon:
Thanks for tracking down the problematic dependency @nsoranzo!

Issue #9467, opened 2020-03-02
20.01 bug: edit on filtered workflow list opens wrong workflow


scholtalbers:
When the workflow list page is filtered, the wrong workflow (first of the unfiltered list) will be opened in edit mode.

dannon:
Fixed in #9468

PR #9468, opened 2020-03-02
[20.01] Fix workflow list filtering dropdown

mvdbeek:
Not sure if this is the right fix but `computed` sounds like it would be less static and it does solve https://github.com/galaxyproject/galaxy/issues/9467 for me.


dannon:
Yeah, this is right.  The sort isn't actually reordering the dom, it just changes the table-level data, which propagates through the property bindings.  `data` isn't reactive like that, so it wouldn't have changed.  (and `computed` *is* reactive, hence the updates to the urls)

dannon (approved):

dannon:
(following up with an update to the history dropdown in a minute)

guerler:
Thanks for fixing this! @mvdbeek

PR #9471, opened 2020-03-02
Render workflow node connections using d3

guerler:
This PR replaces the `canvas` rendering method for workflow node connectors with `d3`. This yields a better resolution. With this PR the connectors render as follows:

Screen Shot 2020-03-02 at 1 47 41 PM

<b>Now:</b>

Screen Shot 2020-03-02 at 1 51 33 PM

<b>Before:</b>

Screen Shot 2020-03-02 at 1 49 16 PM


guerler:
@galaxybot test

client/galaxy/scripts/components/Workflow/Editor/Node.vue
```diff
@@ -1,9 +1,10 @@
 <template>
     <div>
-        <div class="toolFormTitle unselectable clearfix">
-            <i :class="iconClass" />
-            <span class="node-title">{{ title }}</span>
-            <span class="sr-only">&nbsp;Node</span>
+        <div class="node-header unselectable clearfix">
+            <div class="header float-left">
```
dannon, line 4: Without additional styles to fix the display, this can cause tool titles to wrap where they previously would not.  Merging anyway since there's the more style-focused followup PR we can address this in.

client/galaxy/scripts/mvc/workflow/workflow-connector.js
```diff
@@ -1,16 +1,17 @@
 import $ from "jquery";
-
+import * as d3 from "d3";
```
dannon, line 2: In general we want to get away from this and use specific imports `import {select} from ...", but our version of d3 is out of date and not worth fussing with right now since this is how we use it elsewhere; we can fix them all in a single pass.

PR #9473, opened 2020-03-02
Backporting PR #9162 into 20.01

almahmoud:
This is needed for the logout with Keycloak in GVL setup to log the user out of Keycloak as well, not just Galaxy. Without this, user will be logged out of Galaxy, but re-clicking login will get them back in without asking for credentials.


dannon:
(test failure is unrelated, thanks for opening the backport @almahmoud)

PR #9474, opened 2020-03-03
Remove mapping related abstraction classes from workflow client, increase node header contrast

guerler:
Requires #9471. This PR removes the folder icon and instead uses an enlarged connector icon to indicate output and input mapping types. It also replaces the clickable-asterisk indicating workflow outputs with a checkbox. The PR increases the header variant from secondary to primary style and the transparency of the grid background from 60% to 65% to further highlight the workflow nodes. Additionally this PR consolidates and removes mapping related abstraction classes.

<b>Now</b>

Screen Shot 2020-03-11 at 9 41 26 AM

<b>Before</b>

Screen Shot 2020-03-02 at 1 47 41 PM

### DIFF
--- a/client/galaxy/scripts/components/Workflow/Editor/Node.test.js
+++ b/client/galaxy/scripts/components/Workflow/Editor/Node.test.js
@@ -11,15 +11,15 @@ describe("Node", () => {
                 type: "tool"
             }
         });
-        const icon = wrapper.find("i");
-        expect(icon.classes()).to.contain("fa-wrench");
-        const toolLinks = wrapper.findAll("a");
-        expect(toolLinks.length).to.equal(2);
+        const icon = wrapper.findAll("i");
+        expect(icon.at(2).classes()).to.contain("fa-wrench");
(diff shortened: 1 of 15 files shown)


jmchilton (approved):
Looks good to me, lots of cleanup to abstractions here. Thanks so much!

guerler:
@jmchilton, thanks a lot for the review. I talked to a few more people about consolidating the three node header icons into a single dropdown and received mixed responses. Although visually more appealing, a valid critic of the dropdown is that it requires two instead of one click to remove a node. As a consequence, I have restored the previous three icon view in the node header to avoid controversial changes within this PR. I rebased the rest. Thanks again for the review!

guerler:
Thanks @dannon and @bgruening for the comments. I have restored the previous 3-icon header to avoid 2-clicks for now and will revisit this issue in a separate PR.

bgruening:
Not sure what's up with the tests, but :+1: from my site!

PR #9480, opened 2020-03-03
[20.01] Fix data manager table headers

dannon:
This restores the header with the data table name and reload button.  I'm going to add linting in dev for these invalid slot specifications so this stops happening.


mvdbeek:
Thanks @dannon !

PR #9481, opened 2020-03-03
Add eslint rule to check for invalid slot specifications

dannon:
xref #9480 (this will currently fail to lint, until that is merged and merged forward, demonstrating that it would have caught the error)

Additionally drops two eslint plugins we were not using (and won't really benefit from).


guerler:
Thanks @dannon.

PR #9482, opened 2020-03-03
Another undocumented option: job metrics path

hexylena:


lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -2417,6 +2417,13 @@ mapping:
           Details" option in the history. This option also exposes the command line to
           non-administrative users. Administrators can always see dataset paths.
 
+      job_metrics_config_file:
+        type: str
+        default: config/job_metrics_conf.xml
+        required: false
+        desc: |
+          XML config file that contains the job metric collection configuration.
```
jdavcs, line 2425:
```suggestion
          XML config file that contains the job metric collection configuration.
          
          The value of this option will be resolved with respect to <config_dir>.
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -2417,6 +2417,13 @@ mapping:
           Details" option in the history. This option also exposes the command line to
           non-administrative users. Administrators can always see dataset paths.
 
+      job_metrics_config_file:
+        type: str
+        default: config/job_metrics_conf.xml
```
jdavcs:
```suggestion
        default: job_metrics_conf.xml
```

jdavcs: It will be resolved w.r.t. `config_dir` here (regardless of what `config_dir` is set to): https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L765

jdavcs (approved):
The test failure was caused by incorrect handling of this type of config option (one that contains 'sample' config file). Similar config options are also marked 'do not test'. I've added this option to the `DO_NOT_TEST` list, so the test will pass.

test/integration/test_config_defaults.py
```diff
@@ -178,6 +178,7 @@ def expected_default_config_dir(value):
     'id_secret',  # broken: default overridden
     'job_config',  # no obvious testable defaults
     'job_config_file',  # broken: remove 'config/' prefix from schema
+    'job_metrics_config_file',
```
nsoranzo, line 181: Need a comment for why this doesn't work yet?

jdavcs: I was considering that. That whole file deserves a TODO comment - I'm working on it this week. Let's leave as is - I'll have an update for that file within a few days.

jdavcs:
Thank you @hexylena! (Here's to a little more structure in galaxy's config!)

PR #9484, opened 2020-03-04
[20.01] Don't safe_loads multiple times on the same values

dannon:
when decoding tool state.

WIP, there's a bit of history in the commit graph on this handling; seeing if this is viable or if all the tests will break.

(yeah, they all broke)


mvdbeek:
I'll remove the milestone, can always get it in a backport it as a bug

dannon:
@mvdbeek Yep, totally fine with me.  Sam's going to take a pass at the followup here

PR #9489, opened 2020-03-05
Print error message if no action is selected in pgcleanup script

nsoranzo:
Also, clearer exception message in `read_properties_from_file()`.


jmchilton:
Good enhancement - thanks so much!

PR #9491, opened 2020-03-05
Skip migration if uninitialized database exists

mvdbeek:
As we do when Galaxy created the database. This is useful if the database was pre-created because the Galaxy user doesn't have permissions to create new databases.


nsoranzo:
The API test failure seems related: https://jenkins.galaxyproject.org/job/docker-api-py3/3018/console

mvdbeek:
I think it's temporary and happens occasionally. I'll re-trigger

mvdbeek:
It's also passing for #9492 which also has this commit: https://jenkins.galaxyproject.org/job/docker-api-py3/3019/console

nsoranzo (approved):

Issue #9486, opened 2020-03-04
parameter name collision confuses inputs ready check


mvdbeek:
https://help.galaxyproject.org/t/job-execution-error-tool-executes-too-early/3042/10

PR #9493, opened 2020-03-05
Make access to parameters in sections also require full path

bernt-matthias:
So far the access to values of parameter in sections did not require to specify the full access path.

For

```
<section name="sec">
   <param name="par">
</section>
```

in format_source just `par` was used. With the change the access is format_source `sec|par`.

In cheetah `$sec.par` is already used.

The problem was discovered for an automatic tool conversion where two parameters with the same name existed where one was in a section and the other wasn't. Then one did overwrite the other.

### DIFF
--- a/lib/galaxy/managers/workflows.py
+++ b/lib/galaxy/managers/workflows.py
@@ -837,7 +837,7 @@ class WorkflowContentsManager(UsesAnnotations):
                             input_connections_type[input.name] = "dataset"
                         if isinstance(input, DataCollectionToolParameter):
                             input_connections_type[input.name] = "dataset_collection"
-                visit_input_values(module.tool.inputs, module.state.inputs, callback)
+                visit_input_values(module.tool.inputs, module.state.inputs, callback, profile=module.tool.profile)
                 # post_job_actions
                 pja_dict = {}
                 for pja in step.post_job_actions:
@@ -1142,7 +1142,7 @@ class WorkflowContentsManager(UsesAnnotations):
                 if module.tool:
                     # If the tool is installed we attempt to verify input values
                     # and connections, otherwise the last known state will be dumped without modifications.
-                    visit_input_values(module.tool.inputs, module.state.inputs, callback)
+                    visit_input_values(module.tool.inputs, module.state.inputs, callback, profile=module.tool.profile)
 
             # Encode input connections as dictionary
             input_conn_dict = {}
--- a/lib/galaxy/tool_util/xsd/galaxy.xsd
+++ b/lib/galaxy/tool_util/xsd/galaxy.xsd
@@ -186,21 +186,21 @@ automatically if profile is set to any version newer than ``16.01``, such as
 
 16.04
 
-- Disable implicit extra file collection. All dynamic extra file collection requires a `discover_datasets` tag.
-- Disable `format="input"` and require explicit metadata targets (`metadata_source`, `format_source`).
-- Disable `interpreter` use `$__tool_directory__`.
-- Disable `$param_file` use `configfile`
-- Disable default tool version of 1.0.0.
+- Disable implicit extra file collection. All dynamic extra file collection requires a ``discover_datasets`` tag.
+- Disable ``format="input"`` and require explicit metadata targets (``metadata_source``, ``format_source``).
+- Disable ``interpreter`` use ``$__tool_directory__``.
+- Disable ``$param_file`` use ``configfile``.
+- Disable default tool version of ``1.0.0``.
 - Use non zero exit code as default stdio error condition (before non-empty stderr).
 
 18.01
 
 - Use a separate home directory for each job.
-- Introduce `provided_metadata_style` with default `"default"` before `"legacy"`.
+- Introduce ``provided_metadata_style`` with default ``"default"`` before ``"legacy"``.
 
 18.09 
 
-- References to other inputs need to be fully qualified by using `|`.
+- References to other inputs need to be fully qualified by using ``|``.
 - Do not allow provided but illegal default values.
 - Do not use Galaxy python environment for `manage_data` tools.
 
@@ -211,13 +211,20 @@ automatically if profile is set to any version newer than ``16.01``, such as
 20.05
 
 - json config files: 
-   - unselected optional `select` and `data_column` parameters get `None` instead of `"None"`
-   - multiple `select` and `data_column` parameters are lists (before comma separated string)
+   - unselected optional ``select`` and ``data_column`` parameters get ``None`` instead of ``"None"``
+   - multiple ``select`` and ``data_column`` parameters are lists (before comma separated string)
 
 20.09
 
-- Exit immediately if a command exits with a non-zero status (`set -e`)
+- Exit immediately if a command exits with a non-zero status (``set -e``)
 - Assume sort order for collection elements
+
+21.01
+
+- all references to input parameters from output and output collections
+(e.g. with ``structured_like`` or ``format_source``) need to be qualified using ``|``.
+
+
   </xs:documentation>
         </xs:annotation>
       </xs:attribute>
@@ -4110,7 +4117,7 @@ The default is ``galaxy.json``.
   <xs:attributeGroup name="OutputCommon">
     <xs:attribute name="format_source" type="xs:string">
       <xs:annotation>
-        <xs:documentation xml:lang="en">This sets the data type of the output dataset(s) to be the same format as that of the specified tool input.</xs:documentation>
+        <xs:documentation xml:lang="en">This sets the data type of the output dataset(s) to be the same format as that of the specified tool input. If the input is contained in a section or conditional the name of the input must be fully qualified if profile >= 21.01.</xs:documentation>
       </xs:annotation>
     </xs:attribute>
     <xs:attribute name="label" type="xs:string">
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -1530,7 +1530,7 @@ class Tool(Dictifiable):
         """
         # HACK: Yet another hack around check_values -- WHY HERE?
         if self.check_values:
-            visit_input_values(self.inputs, values, callback)
+            visit_input_values(self.inputs, values, callback, profile=self.profile)
 
     def expand_incoming(self, trans, incoming, request_context, input_format='legacy'):
         rerun_remap_job_id = None
@@ -1784,7 +1784,7 @@ class Tool(Dictifiable):
                 else:
                     messages[prefixed_name] = error
 
-        visit_input_values(self.inputs, values, validate_inputs)
+        visit_input_values(self.inputs, values, validate_inputs, profile=self.profile)
         return messages
 
     def build_dependency_cache(self, **kwds):
@@ -2304,7 +2304,7 @@ class Tool(Dictifiable):
                     return map_to_history(value)
             elif isinstance(input, DataCollectionToolParameter):
                 return map_to_history(value)
-        visit_input_values(tool_inputs, params, mapping_callback)
+        visit_input_values(tool_inputs, params, mapping_callback, profile=self.profile)
 
     def _compare_tool_version(self, job):
         """
--- a/lib/galaxy/tools/evaluation.py
+++ b/lib/galaxy/tools/evaluation.py
@@ -78,7 +78,7 @@ class ToolEvaluator:
         def validate_inputs(input, value, context, **kwargs):
             value = input.from_json(value, request_context, context)
             input.validate(value, request_context)
-        visit_input_values(self.tool.inputs, incoming, validate_inputs)
+        visit_input_values(self.tool.inputs, incoming, validate_inputs, profile=self.tool.profile)
 
         # Restore input / output data lists
         inp_data, out_data, out_collections = job.io_dicts()
--- a/lib/galaxy/tools/parameters/__init__.py
+++ b/lib/galaxy/tools/parameters/__init__.py
@@ -5,6 +5,7 @@ Classes encapsulating Galaxy tool parameters.
 from json import dumps
 
 from boltons.iterutils import remap
+import packaging
 
 from galaxy.util import unicodify
 from galaxy.util.expressions import ExpressionContext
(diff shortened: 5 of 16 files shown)


mvdbeek:
That's a nice test, fixing this probably also fixes https://github.com/galaxyproject/galaxy/issues/9486

bernt-matthias:
Cool. Do you know if this worked in the past and just broke or if this never worked?

mvdbeek:
You're not supposed to duplicate parameter names (if they're not the same thing, as in conditional switches), so there is a chance it never worked, but some years ago we also reworked the param dict lookup for unqualified parameter names that could have affected this.

bernt-matthias:
I understand and would never do this when crating wrappers manually.

But, I stumbled over this when autogenerating wrappers for OpenMS .. so I don't have any influence on the parameter names .. or need quite a bit of additional code which I would like to avoid.

mvdbeek:
sure, and that's why I like your test. clearly we don't want to mask `input_data_1 ` with `test_section .input_data_1 `

bernt-matthias:
Good thing is that the test fails: `Dataset metadata verification for [file_ext] failed, expected [fastqsanger] but found [fastqsolexa]`

But from my test that brought be to this bug I would have expected `Dataset metadata verification for [file_ext] failed, expected [fastqsanger] but found [data]`

bernt-matthias:
So here is a possible fix .. at least this test is working now as expected. Maybe this will have side effects, since inputs in sections are now referenced as `section|input`.

Not sure if this is actually a bug in the visitor:

- somehow the `parent_prefix` is empty when it probably shouldn't so I return `name_prefix` https://github.com/galaxyproject/galaxy/blob/972fd28601dbb8815d51e6fd4118c948ee2ae314/lib/galaxy/tools/parameters/__init__.py#L128
- alternatively the calling functions could use `prefixed_name`

bernt-matthias:
Hey @mvdbeek .. this gets a bit tricky and I could need a bit of feedback. 

Seemingly `visit_input_values` did not add prefixes for sections and conditionals so far. That's why the input in the new test was masked. My changes would change this and require that the full name with prefixes is necessary for all attributes referring to inputs (so far I'm aware of `format_source` and `structured_like`. 

I guess this would break a few tools in its current state (quick check at IUC confirms this). 

Also there are some side effects which I do not understand yet (e.g. the py3 api tests).

I'm thinking about a good way to fix this. 

- one could add both the qualified and the unqualified name here https://github.com/galaxyproject/galaxy/blob/9ba1f4550cfa9d25074dd0764fa7189aed10eb36/lib/galaxy/tools/actions/__init__.py#L64
 
- I also remember that in tests with sections / conditionals it is possible address the inputs with and without the section or conditional name (i.e. `section|id` or just `id`). But I'm not sure if I'm right. If so then maybe we can add the same flexibility here in the same way?

What's your take on this?

bernt-matthias:
Reformulated title and descriptions and opened a branch on IUC https://github.com/bernt-matthias/tools-iuc/tree/topic/test-format_source for testing. Tests are running here https://github.com/bernt-matthias/tools-iuc/actions/runs/83804958

bernt-matthias:
The change seems to have some side effect on workflows (corresponding API tests should be fixed) and conditionals (totally forgot that conditionals are also affected)

There was a comment (likely) from @jmchilton and it seems that the behaviour is now as desired

https://github.com/galaxyproject/galaxy/blob/802ddafd561175f4ad957206ee057454b5257715/lib/galaxy_test/api/test_workflows.py#L2957

jmchilton:
I thought it was intentional that sections don't require the prefix addition? Won't this break things?

bernt-matthias:
With respect to tools this seems not to be the case. I just tested the complete IUC repo against this branch (https://github.com/bernt-matthias/tools-iuc/actions/runs/83804958).. All tests passed. I checked some tools and found that in cheetah the access is already `$section.parameter`.
With this respect the change improves consistency. 

Main question seems to be where `visit_input_values` is called. Seems to be only 5 places in the code.

mvdbeek:
> I thought it was intentional that sections don't require the prefix addition? Won't this break things?

It will and we shouldn't change the test tools.

> in format_source just `par` was used. With the change the access is format_source `sec|par`.

Can we not make both work ?

bernt-matthias:
> I thought it was intentional that sections don't require the prefix addition?

Do you recall the idea behind this intention? On a first glance it seems inconsistent to require prefix in cheetah but not in other places. 

> It will 

Maybe :). So far the good news is that all tests here pass .. as well as all IUC tools -- even if I do not understand why in parts: e.g. https://github.com/galaxyproject/tools-iuc/blob/7f0dbcf650d20acd80d7082bdae9759d51a1480c/tools/sickle/sickle.xml .. does not use prefix addition for conditionals (admittedly as documented: https://docs.galaxyproject.org/en/master/dev/schema.html#tool-outputs-data) .. and should fail for this branch .. hrm.

> and we shouldn't change the test tools.

For `identifier_in_conditional.xml ` I thinks the cheetah was just wrong, since it accessed a parameter in a conditional without using the conditional, i.e. it was `$input1` instead of  `$outer_cond.input1`. I was surprised that this actually worked. 

> Can we not make both work ?

Good question. The first thing that came to my mind is to tweak callback helper but for this one would need to define and implement an failure case (exception) for the callback functions. E.g. try the new behavior and if it fails spit out a warning and try the old behavior. 

But I guess this idea is much more complicated to implement than it sounds.

mvdbeek:
> Do you recall the idea behind this intention?

It's the old default. Just because this isn't widely used in modern tool doesn't mean we can break this. Of course fully-prefixed is better.

> Maybe :)

Case in point are the test tools.

> I thinks the cheetah was just wrong, since it accessed a parameter in a conditional without using the conditional, i.e. it was `$input1` instead of `$outer_cond.input1`. I was surprised that this actually worked.

Yes, that used to work and I insist that this must continue working. If you think there is no way to have both working we can make the new behavior conditional on a new `profile` version.

bernt-matthias:
I agree completely. The `profile` option seems to be the best solution. You already increased the milestone to 20.09 which seems realistic.

mvdbeek:
Happy to get this in if you get this working ... it is a bug after all.

bernt-matthias:
Will think about if its possible to change the behavior without breaking stuff.

bernt-matthias:
Just thought about implementing a profile version. Seems that the function is called in quite a few places:

- lib/galaxy/managers/workflows.py
- lib/galaxy/workflow/modules.py
- lib/galaxy/tools/__init__.py
- lib/galaxy/tools/evaluation.py
- lib/galaxy/tools/parameters/meta.py

Is there a better way then introducing a new `profile` parameter (in the function itself and probably quite a lot of the calling functions, and in the functions calling them, etc).

bernt-matthias:
Does someone has an idea for my last question?

> Is there a better way then introducing a new profile parameter (in the function itself and probably quite a lot of the calling functions, and in the functions calling them, etc).

Btw. xrefs between input parameters seem to work in yet another way and xref to parameters in sections seem not to work at all https://github.com/galaxyproject/galaxy/pull/11026

bernt-matthias:
Hey @mvdbeek . I'm making good progress with the profile version. The only problem seems to be two calls in https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/workflow/modules.py where `tool` is a `Bunch` and I do not know what to do there yet. 

I also extended the unit tests for the function to cover both profile versions. I'm still not entirely sure if this is "kind of a bug"

See the lines 66 ff: https://github.com/galaxyproject/galaxy/blob/c4947ff3dcc95711bc60537e64928bff0ff41d17/lib/galaxy/tools/parameters/__init__.py#L66

Parameters e, g, and h are in the section d, but this is inconsistently represented in the prefix (i.e. d is not in the prefix of e, but it is for g and h). More precisely the old implementation skipped the parent level which makes `prefix+name !=  prefixed_name`. 

I guess this would become apparent for sections or condiationals with 2 or more nesting levels and output parameter referring to them .. quite unlikely ..

Left are to undo the changes to the test tools and replace them by new tests. I guess we need a profile version for each of them, or?

mvdbeek:
> The only problem seems to be two calls in https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/workflow/modules.py where `tool` is a `Bunch` and I do not know what to do there yet.

Where exactly are you seeing this ?

bernt-matthias:
> > The only problem seems to be two calls in https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/workflow/modules.py where `tool` is a `Bunch` and I do not know what to do there yet.
> 
> Where exactly are you seeing this ?

In the failing unit tests: 

https://app.circleci.com/pipelines/github/galaxyproject/galaxy/12501/workflows/215d048b-9fd6-42f2-856f-d70a58b47d27/jobs/123787

mvdbeek:
You can add a profile here: https://github.com/jmchilton/galaxy/blob/b1fc8430bb6b743954b2221d2141f016d0d1e4c9/test/unit/workflows/test_modules.py#L443

bernt-matthias:
I guess we should put the change to identifier_in_conditional.xml to a separate PR (if at all .. do not understand why this worked before).

> You can add a profile here: https://github.com/jmchilton/galaxy/blob/b1fc8430bb6b743954b2221d2141f016d0d1e4c9/test/unit/workflows/test_modules.py#L443

Should we add tests here for the new profile? Currently I just added the default profile.

> I thought it was intentional that sections don't require the prefix addition? Won't this break things?

It seems that `structured_like` already was supposed to work like this from 18.09 https://github.com/galaxyproject/galaxy/commit/6cfe5880e13e0f704782afb72ebafd7fac3d45d4 but I'm not sure if it really did because this was not covered by tests (as far as I can see).

mvdbeek:
Yeah, let's not change test/functional/tools/identifier_in_conditional.xml. The rest looks good to me, I don't think we need extra profile tests given that you have the high-level framework tests.

mvdbeek (dismissed):

bernt-matthias:
The remaining two workflow api tests should pass now again after I undid the changes. 

Considering the source comment of @jmchilton `# Wish it was qualified for conditionals but it doesn't seem to be. -John` I'm wondering if we should test also the profile version? If yes what do you think is the easiest way to do this over there?

Do I need to create duplicates of the tests using a mapper2 tool xml for the 21.01 profile?

bernt-matthias:
But in the end it seems a odd/wrong to me that access to variables in workflows might be inconsistent from tool to tool .. depending on their corresponding profile.

mvdbeek:
I think that (the PJA rename) should be straightforward to fix if we always pass in the fully-prefixed versions. I've experimented a bit with this, and I think we can get away with something like
```
diff --git a/lib/galaxy/tools/evaluation.py b/lib/galaxy/tools/evaluation.py
index 6ca6482b09..696d593b63 100644
--- a/lib/galaxy/tools/evaluation.py
+++ b/lib/galaxy/tools/evaluation.py
@@ -432,7 +432,18 @@ class ToolEvaluator:
                     # Remove key so that new wrapped object will occupy key slot
                     del param_dict[key]
                     # And replace with new wrapped key
-                    param_dict[wrap_with_safe_string(key, no_wrap_classes=ToolParameterValueWrapper)] = wrap_with_safe_string(value, no_wrap_classes=ToolParameterValueWrapper)
+                    wrapped_value = wrap_with_safe_string(value, no_wrap_classes=ToolParameterValueWrapper)
+                    # add fallback for tools < 21.01
+                    if self.tool.profile < 21.01:
+                        key_fragments = key.split('|')
+                        for i in range(1, len(key_fragments)):
+                            key_fragment = "|".join(key_fragments[(i):])
+                            wrapped_key = wrap_with_safe_string(key_fragment, no_wrap_classes=ToolParameterValueWrapper)
+                            if wrapped_key not in param_dict:
+                                param_dict[wrapped_key] = wrapped_value
+                            else:
+                                param_dict[wrapped_key] = wrapped_value
+                    param_dict[wrap_with_safe_string(key, no_wrap_classes=ToolParameterValueWrapper)] = wrapped_value

```
which means we don't have to pass the profile version around, instead we always generate the fully prefixed version internally and generate the fallback for the param_dict.
However, something doesn't quite work there yet, I'll spend a bit more time on this.

mvdbeek:
I spent the better part of today looking at this, and I suspect this isn't quite working as it should. It breaks for profile >= 21.01 and multiple="true" iteration in the command block, and it only works in other cases because the fallback that should only be used in rare cases (https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/evaluation.py#L288) saves the day.

The code itself looks good and results in fully prefixed job parameters and JobToInputDatasetAssociations, which I think is a big plus if we can use that for all tool executions. I think we can handle unprefixed parameters in the params_dict. Given this fixes a bug, I will branch without this, but we should come back to it.

mvdbeek:
fwiw that is where I stopped today: https://github.com/galaxyproject/galaxy/commit/d1445c6c0e60c463eab6f59eef38b1feeb1e166b

martenson:
I am not sure how it happened but it seems this feature has gotten to the [xsd](https://github.com/galaxyproject/galaxy/blob/release_21.05/lib/galaxy/tool_util/xsd/galaxy.xsd#L2444) before it was merged.

bernt-matthias:
> I am not sure how it happened but it seems this feature has gotten to the [xsd](https://github.com/galaxyproject/galaxy/blob/release_21.05/lib/galaxy/tool_util/xsd/galaxy.xsd#L2444) before it was merged.
> 
> edit: xsd modified [here](https://github.com/galaxyproject/galaxy/pull/11026/files#diff-7609c5c4f82006f4dafa20d26bfce4b2c77680442044ff1fb0ce4e28cf7a2879R2411)

True. I can't recall either. We should make sure that we either finish this PR or revert the change  for 21.09

bernt-matthias:
Hey @martenson .. just checked: The change is not in dev, but I can find it in the 21.05 branch. I'm confused how this happened. Should I revert it only on the 21.05 branch?

bernt-matthias:
> fwiw that is where I stopped today: [d1445c6](https://github.com/galaxyproject/galaxy/commit/d1445c6c0e60c463eab6f59eef38b1feeb1e166b)

Just thought to rebase this to keep the PR up to date. Should I just include your commit before doing this?

martenson:
Well that is confusing, one can hope that this was the only thing that is stuck there in the limbo. :/

nsoranzo:
> Hey @martenson .. just checked: The change is not in dev, but I can find it in the 21.05 branch. I'm confused how this happened. Should I revert it only on the 21.05 branch?

I think it's also in dev?

https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tool_util/xsd/galaxy.xsd#L2524-L2530

mvdbeek:
I think I fixed the backward compatibility issue in https://github.com/galaxyproject/galaxy/pull/15978 ... if you want to pursue this further you could disallow the fallback lookup with a profile version. I didn't do this because I feel like this could easily result in broken tools if you update the profile version but the tests don't cover all parameter references.

PR #9494, opened 2020-03-05
unicodify message

ieguinoa:
Not sure if unicodify() is the way to go but I was getting an error calling parse_external_id() using Python 3.6.8 when submitting jobs to condor as the  output from communicate() was a byte sequence and not a string.

### DIFF
--- a/lib/galaxy/jobs/runners/util/condor/__init__.py
+++ b/lib/galaxy/jobs/runners/util/condor/__init__.py
@@ -78,6 +78,7 @@ def condor_submit(submit_file):
     try:
         submit = Popen(('condor_submit', submit_file), stdout=PIPE, stderr=STDOUT)
         message, _ = submit.communicate()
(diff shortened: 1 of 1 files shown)


lib/galaxy/jobs/runners/util/condor/__init__.py
```diff
@@ -78,6 +78,7 @@ def condor_submit(submit_file):
     try:
         submit = Popen(('condor_submit', submit_file), stdout=PIPE, stderr=STDOUT)
         message, _ = submit.communicate()
+        message=unicodify(message)
```
nsoranzo:
```suggestion
        message = unicodify(message)
```

ieguinoa:
Thanks @nsoranzo

nsoranzo:
@galaxybot test this

nsoranzo (approved):

PR #9496, opened 2020-03-05
Drop custom json encoding for tool test API endpoint

nvk747:


lib/galaxy/webapps/galaxy/api/tools.py
```diff
@@ -205,7 +206,7 @@ def test_data(self, trans, id, **kwd):
         def json_encodeify(obj):
             if isinstance(obj, OrderedDict):
                 return dict(obj)
```
nsoranzo: This doesn't seem to be needed after d6cf262ffe9c7c0a4eac4e47aa750340c8f0010e

lib/galaxy/webapps/galaxy/api/tools.py
```diff
@@ -205,7 +206,7 @@ def test_data(self, trans, id, **kwd):
...
-            elif isinstance(obj, map):
+            elif inspect.isclass(map) and isinstance(obj, map):
                 return list(obj)
```
nsoranzo: @mvdbeek This also shouldn't be needed any more after https://github.com/galaxyproject/galaxy/pull/9416 ?

I think we may just drop the method.

mvdbeek: Yep, done that. Targeting dev for more extended test coverage, we can backport after merge

mvdbeek:
@galaxybot test this

PR #9501, opened 2020-03-06
Added whitelisting option for user registration

MiguelJulia:
I added the option to avoid user registration if email domain does not match with the allowed ones for the server, specially usefull for private galaxy instances running in small institutions or companies.

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -468,7 +468,6 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
         #  Create whitelist file to accept only certain email domains
         self.whitelist_content = None
         if self.whitelist_file:
-            self.whitelist_file = os.path.join(self.root, self.whitelist_file)
             try:
                 with open(self.whitelist_file) as f:
                     self.whitelist_content = [line.rstrip() for line in f]
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -465,6 +465,16 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
             except IOError:
                 log.error("CONFIGURATION ERROR: Can't open supplied blacklist file from path: %s", self.blacklist_file)
 
+        #  Create whitelist file to accept only certain email domains
+        self.whitelist_content = None
+        if self.whitelist_file:
+            self.whitelist_file = os.path.join(self.root, self.whitelist_file)
+            try:
+                with open(self.whitelist_file) as f:
+                    self.whitelist_content = [line.rstrip() for line in f]
+            except IOError:
+                log.error("CONFIGURATION ERROR: Can't open supplied whitelist file from path: %s", self.whitelist_file)
+
         self.persistent_communication_rooms = listify(self.persistent_communication_rooms, do_strip=True)
         # The transfer manager and deferred job queue
         self.enable_beta_job_managers = string_as_bool(kwargs.get('enable_beta_job_managers', 'False'))
--- a/lib/galaxy/security/validate_user_input.py
+++ b/lib/galaxy/security/validate_user_input.py
@@ -46,6 +46,13 @@ def validate_email(trans, email, user=None, check_dup=True, allow_empty=False):
             domain = ('.').join(domain.split('.')[-2:])
         if domain in trans.app.config.blacklist_content:
             message = "Please enter your permanent email address."
+    #  If the whitelist is not empty filter out any domain not in the list.
+    elif trans.app.config.whitelist_content is not None:
+        domain = email.split('@')[1]
+        if len(domain.split('.')) > 2:
+            domain = ('.').join(domain.split('.')[-2:])
+        if domain not in trans.app.config.whitelist_content:
+            message = "Please enter an allowed domain email address for this server."
     return message
 
 
--- a/lib/galaxy/security/validate_user_input.py
+++ b/lib/galaxy/security/validate_user_input.py
@@ -39,6 +39,13 @@ def validate_email(trans, email, user=None, check_dup=True, allow_empty=False):
         message = "Email address cannot be more than %d characters in length." % EMAIL_MAX_LEN
     elif check_dup and trans.sa_session.query(trans.app.model.User).filter(func.lower(trans.app.model.User.table.c.email) == email.lower()).first():
         message = "User with email '%s' already exists." % email
+    #  If the whitelist is not empty filter out any domain not in the list and ignore blacklist.
+    elif trans.app.config.whitelist_content is not None:
+        domain = email.split('@')[1]
+        if len(domain.split('.')) > 2:
+            domain = ('.').join(domain.split('.')[-2:])
+        if domain not in trans.app.config.whitelist_content:
+            message = "Please enter an allowed domain email address for this server."
     #  If the blacklist is not empty filter out the disposable domains.
     elif trans.app.config.blacklist_content is not None:
         domain = email.split('@')[1]
@@ -46,13 +53,6 @@ def validate_email(trans, email, user=None, check_dup=True, allow_empty=False):
             domain = ('.').join(domain.split('.')[-2:])
         if domain in trans.app.config.blacklist_content:
             message = "Please enter your permanent email address."
-    #  If the whitelist is not empty filter out any domain not in the list.
-    elif trans.app.config.whitelist_content is not None:
-        domain = email.split('@')[1]
-        if len(domain.split('.')) > 2:
-            domain = ('.').join(domain.split('.')[-2:])
-        if domain not in trans.app.config.whitelist_content:
-            message = "Please enter an allowed domain email address for this server."
     return message
 
 
--- a/lib/galaxy/webapps/galaxy/config_schema.yml
+++ b/lib/galaxy/webapps/galaxy/config_schema.yml
@@ -904,7 +904,8 @@ mapping:
 
       whitelist_file:
         type: str
-        default: config/disposable_email_whitelist.conf
+        default: disposable_email_whitelist.conf
+        path_resolves_to: config_dir
         required: false
         desc: |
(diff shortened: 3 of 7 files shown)


lib/galaxy/security/validate_user_input.py
```diff
@@ -46,6 +46,13 @@ def validate_email(trans, email, user=None, check_dup=True, allow_empty=False):
             domain = ('.').join(domain.split('.')[-2:])
         if domain in trans.app.config.blacklist_content:
             message = "Please enter your permanent email address."
+    #  If the whitelist is not empty filter out any domain not in the list.
+    elif trans.app.config.whitelist_content is not None:
```
hexylena: So with `elif` here, these will be mutually exclusive options, if blacklist file is set, whitelist will be ignored. This sounds reasonable but should be documented.

MiguelJulia: Thanks for the reminder! I am going to add to the documentation thatif blacklist is defined then whitelist is ignored.

lib/tool_shed/webapp/config_schema.yml
```diff
@@ -482,6 +482,16 @@ mapping:
           disposable email address during the registration.  If their address domain
           matches any domain in the blacklist, they are refused the registration.
 
+      whitelist_file:
+        type: str
+        default: config/disposable_email_whitelist.conf
```
jdavcs:
```suggestion
        default: disposable_email_whitelist.conf
        path_resolves_to: config_dir
```

jdavcs: We can have it in the `config_dir` without hardcoding `config` into the default (using `path_resolves_to`)

lib/tool_shed/webapp/config_schema.yml
```diff
@@ -482,6 +482,16 @@ mapping:
...
+        required: false
+        desc: |
+          E-mail domains whitelist is used for filtering out users that are using
+          email address not belonging to the instition/company domains during the
```
jdavcs:
```suggestion
          email address not belonging to the institution/company domains during the
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -902,6 +902,18 @@ mapping:
 
           Example value 'config/disposable_email_blacklist.conf'
 
+      whitelist_file:
+        type: str
+        default: config/disposable_email_whitelist.conf
+        required: false
+        desc: |
+          E-mail domains whitelist is used for filtering out users that are using
```
jdavcs: I found the wording rather confusing (I know it's based on the description of blacklist file). Also, I think the logic is incorrect (unless I'm misunderstanding..): if a user's email address domain is *not* on the list, they should be denied registration, right?
Here's a suggested wording, feel free to use any/all or adapt, etc. My only real concern is the logic.

E-mail domains whitelist is used to specify allowed email address domains. If the list is non-empty and a user attempts registration using an email address belonging to a domain that is not on the list, registration will be denied. This is a more restrictive option than <blacklist_file>, and therefore, in case <blacklist_file> is defined, <whitelist_file> will be ignored.

Also, now that I think of it, I think "more restrictive" implies that if `a` is more restrictive than `b`, then `a` takes precedence if both are defined, no? Should it say *less restrictive* instead? Not sure here..

lib/galaxy/config/__init__.py
```diff
@@ -465,6 +465,16 @@ def _process_config(self, kwargs):
             except IOError:
                 log.error("CONFIGURATION ERROR: Can't open supplied blacklist file from path: %s", self.blacklist_file)
 
+        #  Create whitelist file to accept only certain email domains
+        self.whitelist_content = None
+        if self.whitelist_file:
+            self.whitelist_file = os.path.join(self.root, self.whitelist_file)
```
jdavcs:
```suggestion
```
It will be in the `config_dir` - it's placed there automatically by the config system. (unless we want it to be in galaxy root, that is).

jdavcs (changes requested):

lib/tool_shed/webapp/config_schema.yml
MiguelJulia: Ok, I will change it. But this is based on the black list feature, which hardcodes config in the same files. Do you want me to change it for blacklist too?

jdavcs: I wouldn't worry about blacklist: it's one of the many things we need to change in the config system (unless there is a good reason for keeping that file in galaxy root); also changing it would require gracefully handling existing setups, so it's best done in a separate PR. Also, a `config` prefix would be stripped away here: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L291

lib/galaxy/webapps/galaxy/config_schema.yml
MiguelJulia: Thanks for the re-wording of the documentation, it is clearer now. I will add it.

Not sure about the definition of restrictive and its priority either. This rule is more restrictive in the sense that it allows less domains (only the ones specified) than the blacklist option (which allows all but the ones specified). 

If the admin chooses to whitelist certain domains it is removing a ton of domains that are not excluded with blacklist, with the aim of "closing" the registration of new users to only the owners of an email address with the allowed domains. Then, why doing the efford of blacklisting an extensive list of domains when they are not going to be allowed any way?

I sincerelly don't know. Both priority options could be argued, so the one you want to choose is fine.

lib/galaxy/config/__init__.py
MiguelJulia: They are specified in the block of blacklist, do you want me to remove them from there too?

EDIT: I autoreplay myself. No, blacklist needs it, leaving it as it is and removing that line only for whitelist.

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Yep, you are right about more restrictive. However, I think the logic in the `validate_user_input.py` file may be incorrect: we don't check whitelist if blacklist is defined. So, let's say we have `whitelist` set to "foo.com" and `blacklist` set to "spam.com". Then when someone registers with "john@not-allowed-domain.com", they will be able to do that because the blacklist will not filter it out and whitelist will be ignored.

So , if both lists are defined (which makes little sense, but still..), you'd expect whitelist to take precedence and filter out everything not on the list, and blacklist to be ignored. Is this correct? If so, we just need to switch the `elif` statements.

lib/galaxy/security/validate_user_input.py
jdavcs: @hexylena they *are* mutually exclusive, but shouldn't it be the other way round - if whitelist is set, ignore the blacklist? (see discussion further down the page)

lib/galaxy/webapps/galaxy/config_schema.yml
MiguelJulia: Yep, let's give priority to whitelist then.

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -902,6 +902,18 @@ mapping:
...
+        type: str
+        default: disposable_email_whitelist.conf
+        path_resolves_to: config_dir
+        required: false
+        desc: |
+          E-mail domains whitelist is used to specify allowed email address domains.
+          If the list is non-empty and a user attempts registration using an email
+          address belonging to a domain that is not on the list, registration will be
+          enied. This is a more restrictive option than <blacklist_file>, and therefore,
+          in case <blacklist_file> is defined, <whitelist_file> will be ignored.
```
jdavcs:
```suggestion
          in case <whitelist_file> is defined, <blacklist_file> will be ignored.
```
Now that the login in `validate_user_input.py` has been updated.

lib/tool_shed/webapp/config_schema.yml
```diff
@@ -482,6 +482,18 @@ mapping:
...
+        type: str
+        default: disposable_email_whitelist.conf
+        path_resolves_to: config_dir
+        required: false
+        desc: |
+          E-mail domains whitelist is used to specify allowed email address domains.
+          If the list is non-empty and a user attempts registration using an email
+          address belonging to a domain that is not on the list, registration will be
+          enied. This is a more restrictive option than <blacklist_file>, and therefore,
+          in case <blacklist_file> is defined, <whitelist_file> will be ignored.
```
jdavcs: same as above (galaxy's main config_schema)

jdavcs (approved):
Looks great!

martenson:
The disposable domain blacklist config option in Galaxy is long overdue a refactor to a more inclusive language. The upstream project has made this change a while back (https://github.com/martenson/disposable-email-domains/pull/177).

I'd appreciate if naming in this PR could be refactored to something like 'allowlist'. I'll be happy to implement this if you are OK with it @MiguelJulia.

disclaimer: I am also a maintainer of the upstream project

MiguelJulia:
@martenson Please, tell me what variable names have to be changed and what would be a better naming choice and I will change it during the day.

jdavcs:
@martenson @MiguelJulia My suggestion would be to keep these refactorings separate. The purpose of this PR is to introduce a whitelist - I think it's a well-defined change to the system. IMHO, if there is any technical dept that it introduces, that would be the extraction of the domain name from an email address - which, I think, is best handled as a follow-up PR. If the blacklist needs to be refactored, let's do it together with the whitelist - at which point we can apply a different naming convention to both, etc, etc.

martenson:
Thanks @MiguelJulia, this is a great first PR! @ic4f is right, my concerns will fit better as a separate work.

jdavcs:
Thank you, all!

MiguelJulia:
Thanks for your help!!

mvdbeek:
Thanks @MiguelJulia, really cool that this came out of the admin training!

PR #9505, opened 2020-03-09
Selenium Multi-History View

OlegZharkov:
Another round of expanding test coverage from https://github.com/galaxyproject/galaxy/issues/3241
What is implemented:
- Multi-history view
  - [x]  Creating new history from this view.
  - [x]  Switching.
  - [x]  Copying.
  - [x]  Deleting and Purging

I want to finish Multi-history View, but some thing from https://github.com/galaxyproject/galaxy/issues/3241 are not really clear to me. @jmchilton what did you mean by ```Dataset rendering``` and ```Retest a subset of history panel operations (multiple dataset operations, etc..)```. Don't we already test subsets  in  ```test_list_list_list_display``` and ```test_list_list_display```? 

Otherwise, everyone who's interested, feel free to review available tests

### DIFF
--- a/client/galaxy/scripts/mvc/history/multi-panel.js
+++ b/client/galaxy/scripts/mvc/history/multi-panel.js
@@ -212,10 +212,14 @@ var HistoryViewColumn = Backbone.View.extend(baseMVC.LoggableMixin).extend({
         `<div class="text-right col-4">
             <% if( !data.history.purged ){ %>
                 <div class="panel-menu btn-group">
-                    <button type="button" class="btn btn-secondary dropdown-toggle" data-toggle="dropdown">
+                    <button 
+                    history-dropdown-btn="<%= data.history.id %>" 
+                    type="button"
+                    class="btn btn-secondary dropdown-toggle"
+                    data-toggle="dropdown">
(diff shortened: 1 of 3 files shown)


davebx:
@OlegZharkov #9514 might be relevant to these tests, as it changes `self.current_history()` to return the most recently switched-to (or created) history rather than just the most recently created.

jmchilton:
@galaxybot test this

OlegZharkov:
@davebx 
I see. Thanks for update, it is a good observation. I will rebase and try again

jmchilton:
> Dataset rendering

We do expand a couple datasets in test_list_list_list_display - but we don't check any of the properties, etc... Like clicking info from the multi-view etc... opens a new tab but it opens in the main panel in the normal history view. I meant testing stuff like that.

> Retest a subset of history panel operations (multiple dataset operations, etc..)

I don't think we test the "Operations on Multiple Datasets" button in this view of the histories either - I think that is what I meant by that. The referenced tests I think create the collections via the API - if just one created a collection from the multi-history view that would be a good step forward?

OlegZharkov:
thank you @jmchilton !

PR #9508, opened 2020-03-09
Adjust dataset error reporting page, remove email option, remove long info text

### LINKED ISSUE #9296: Bug reporting page asks for email even when logged in
There are two issues here. One is that it doesn't seem necessary to ask for the email since we already know what it is. The second is that the user can enter an arbitrary email, which means you could potentially spam someone with it. The behaviour seems particularly desirable when `require_login=True`, and/or `user_activation_on=True`, since the user email on record is guaranteed to be correct.

guerler:
Fixes #9296. This PR simplifies the error reporting page, the email information is redundant and should be changed in the User Preference view, if necessary. Also this PR removes the statements referring to the "usual" behavior of a Galaxy Administrator. Error reporting can be very helpful and, in my opinion should be made as easy as possible, even if administrators might not be able to respond adequately to all reports in due time. This PR also makes minor title and style adjustments to the error reporting page.

### DIFF
--- a/client/galaxy/scripts/mvc/dataset/dataset-error.js
+++ b/client/galaxy/scripts/mvc/dataset/dataset-error.js
@@ -23,7 +23,6 @@ var View = Backbone.View.extend({
     // Fetch data for the selected dataset and
     render: function() {
         var data_url = `${getAppRoot()}api/datasets/${this.model.get("dataset_id")}`;
-
         Utils.get({
             url: data_url,
             success: dataset => {
@@ -34,7 +33,7 @@ var View = Backbone.View.extend({
                         this.render_error_page(dataset, job);
                         this.find_common_problems(job);
                     },
-                    error: response => {
+                    error: () => {
                         var error_response = {
                             status: "error",
                             message: "Error occurred while loading the job.",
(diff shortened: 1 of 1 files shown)


nsoranzo:
I think this will prevent anonymous users from reporting issues?

client/galaxy/scripts/mvc/dataset/dataset-error.js
```diff
@@ -173,50 +170,26 @@ var View = Backbone.View.extend({
...
-        */
-
-        var form = new Form({
-            title: _l("Error Report"),
-            inputs: inputs,
+        const form = new Form({
+            inputs: [
+                {
+                    type: "text",
+                    hidden: true,
```
nsoranzo: Can you make this hidden only if the user is not anonymous? Otherwise they won't be able to insert their email address to receive a reply.

I'd leave it not-hidden for all users, so one could choose a different email address to receive this kind of emails.

guerler: I enabled the email for anonymous users although I am not sure how useful dataset error reporting is for an anon user, since the session will be cleared.

guerler:
@nsoranzo thanks for the comments, I have applied the changes.

nsoranzo (approved):

PR #9123, opened 2019-12-16
Client build work (HMR, build upgrades, refactoring asset placement)

dannon:
Fixes HMR build to be more reliable, and to work with styles -- you can now edit base.scss (or whatever) and it'll automatically reload an open browser with the new style applied.  

While getting this working I also upgraded a lot of the core devDependencies involved in the process, tweaked our webpack config a bit, and did some general refactoring in terms of build & static artifact placement trying to work towards a single standard 'dist' with less exposed.  I eliminated `june_2007_style` and adjusted the few things touching this.

Style sourcemap resolution should be fixed.

The general production build is marginally faster now.  (but you're really going to want to use the HMR build for developing now, anyway -- it's very fast comparatively).

Addressed comments on a standard for vue v-bind and v-on by enabling vue/strongly-recommended (with prettier compromises asserted) and standardizing the codebase on the recommended practice.

PR #9511, opened 2020-03-10
Update charts entrypoint inclusion of dist artifacts

dannon:
Fixes a bug that @OlegZharkov ran into when he was following the charts tutorial; it's a spot I missed updating in #9123


bgruening:
Thanks @dannon!

PR #9513, opened 2020-03-10
fix python3 error in the condor job runner

bgruening:


nsoranzo:
This is equivalent to https://github.com/galaxyproject/galaxy/pull/9494 , which I just merged. Thanks anyway!

Issue #8704, opened 2019-09-25
Optimize changing Galaxy job/dataset states.


jmchilton:
Two big issues - handling large collections in analyses effectively and providing more feedback to users - would be greatly aided by being able to more efficiently change a job's state.

Two related issues:

- https://github.com/galaxyproject/galaxy/issues/4959
- https://github.com/galaxyproject/galaxy/pull/7791

The second issue in particular lays out a short term and long term plan to mitigate issues around this.

The short term is to migrate the guts of ``job_wrapper.change_state`` to a single SQL query to be executed atomically by the database. The longer term goal would be eliminate the process of tracking job states on dataset instances - for instance when the job goes into the 'running' state, we update all the HDAs as well. There will need to be dataset states still - e.g. metadata failure but we could handle that logic in Python ("state is job state if dataset state is None else if dataset state").

I'm convinced in addition to allowing us to make progress on the size of collections we can handle effectively and provide more feedback to users - this will also mitigate if not outright eliminate certain classes of state related bugs when dealing with Postgres (https://github.com/galaxyproject/galaxy/pull/8498) and sqlite (https://github.com/galaxyproject/galaxy/issues/1158). change_state with its refresh and such is implicitly trying to lock things that only the database can lock - this is inefficient and error prone. Another benefit of this would hopefully then be faster and less buggy test suites.

ic4f:
Single SQL query for `job_wrapper.change_state` (referenced above) implemented in #9722.

PR #9515, opened 2020-03-12
Add trigger to optimize job change_state()

jdavcs:
This is the first step in addressing #8704 (also, relevant for #7791) 

This db trigger replaces the loop inside JobWrapper's `change_state()` method: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/jobs/__init__.py#L1367
On job state update, for each `HDA` and `LDDA` associated with the job via `JobToOutputDataset`/`JobToOutputLibraryDataset`:
  - `HDA`/`LDDA`: `info` and `update_time` are updated
  - associated `Dataset`: `state` and `update_time` are updated
All values are taken from the updated `Job` table row (including `update_time`).

NOTES: 
1. The dataset and HDA/LDDA `update_time` is updated *regardless* of whether the dataset's state has been changed. However, the trigger is fired only if the job's `info` or `state` column are mentioned in the update command (I assume SA's flush() would include them in the command only if those attributes had been modified, but I don't know for sure).
2. Suggested trigger naming convention based in part on @nsoranzo suggestion + https://severalnines.com/database-blog/postgresql-triggers-and-stored-function-basics
I think we *do* need to add an "action" suffix because there may be >1 triggers with the same table/when/operation specs. Also, readability.
3. I did not generate the trigger name automatically: IMHO flexibility+readability justify some minor redundancy.
4. I'm using one name for postgres trigger and trigger function. I can't think of a reason not to.
5. I explicitly drop the trigger *and* the function. I'm aware of the `cascade` option, but I think dropping objects explicitly makes sense here, especially for the migration downgrade use case (I'd prefer the code to list the objects being dropped).
6. This does not support MySQL. Should it?
7. The `update_time` value is obtained from the updated `Job`: that way we avoid having the Job's `update_time` set by the client and its associated datasets - by the database.
8. This does not include any tests yet. It should.
9. @Nerdinacan thank you for laying out all that groundwork for adding triggers. The code + discussion and code review in that PR were a great help!

This is ready for (initial) review, although I'm not sure what this may break..

Ping @jmchilton @natefoo @nsoranzo .

### DIFF
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1364,22 +1364,7 @@ class JobWrapper(HasResourceParameters):
             log.warning("(%s) Ignoring state change from '%s' to '%s' for job "
                         "that is already terminal", job.id, job.state, state)
             return
-        for dataset_assoc in job.output_datasets + job.output_library_datasets:
-            dataset = dataset_assoc.dataset
-            if not job_supplied:
-                self.sa_session.refresh(dataset)
-            state_changed = dataset.raw_set_dataset_state(state)
-            if state_changed:
-                # Arguably a hack to get state changes to appear in the history panel because
-                # the history panel polls on hda.update_time and ignores hda.dataset.update_time.
-                # For those less pragmatic needing a more theoretically sound reason for the update,
-                # perhaps arguments can be made that the entity that is the HDA
-                # really should be described as updated since its effective state did change and its
-                # RESTful representation in the API does change as a result of the above dataset update.
-                dataset.update()
-            if info:
-                dataset.info = info
-            self.sa_session.add(dataset)
+
         if info:
             job.info = info
         job.set_state(state)
--- a/lib/galaxy/model/mapping.py
+++ b/lib/galaxy/model/mapping.py
@@ -42,7 +42,7 @@ from galaxy.model.custom_types import JSONType, MetadataType, TrimmedString, UUI
 from galaxy.model.orm.engine_factory import build_engine
 from galaxy.model.orm.now import now
 from galaxy.model.security import GalaxyRBACAgent
-from galaxy.model.triggers import install_timestamp_triggers
+from galaxy.model.triggers import create_triggers
 
 log = logging.getLogger(__name__)
 
@@ -2872,11 +2872,10 @@ def init(file_path, url, engine_options=None, create_tables=False, map_install_m
 
     result = ModelMapping(model_modules, engine=engine)
 
-    # Create tables if needed
+    # Create tables + other database objects if needed
     if create_tables:
         metadata.create_all()
-        install_timestamp_triggers(engine)
-        # metadata.engine.commit()
+        create_triggers(engine)
 
     result.create_tables = create_tables
     # load local galaxy security policy
--- a/lib/galaxy/model/migrate/versions/0166_add_job_state_update_trigger.py
+++ b/lib/galaxy/model/migrate/versions/0166_add_job_state_update_trigger.py
@@ -1,0 +1,30 @@
+"""
+Migration script to add a trigger to Job table that updates associated datasets
+on job update.
+"""
+
+from __future__ import print_function
+
+import logging
+
+from sqlalchemy import MetaData
+
+from galaxy.model.triggers import JobStateTrigger
+
+log = logging.getLogger(__name__)
+metadata = MetaData()
+
+
+def upgrade(migrate_engine):
+    print(__doc__)
+    metadata.bind = migrate_engine
+    metadata.reflect()
+    trigger = JobStateTrigger(migrate_engine)
+    trigger.create()
+
+
+def downgrade(migrate_engine):
+    metadata.bind = migrate_engine
+    metadata.reflect()
+    trigger = JobStateTrigger(migrate_engine)
+    trigger.drop()
--- a/lib/galaxy/model/triggers.py
+++ b/lib/galaxy/model/triggers.py
@@ -1,8 +1,181 @@
 """
 Database trigger installation and removal
+
+Recommended trigger naming convention: tr_{table name}_{when}_{operation}_{action details}, where:
+    table_name: table on which the trigger is created
+    when: b|a|i for before, after, or instead of
+    operation: i|u|d for insert, update, or delete
+    action details: specify what the trigger does
 """
 
 from sqlalchemy import DDL
+
+
+def create_triggers(engine):
+    JobStateTrigger(engine).create()
+    install_timestamp_triggers(engine)
+
+
+class Trigger(object):
+    """
+    Base class for triggers. A concrete trigger class should extend this class and override
+    the `get_sql_create_postgres_trigger()` and `get_sql_create_sqlite_trigger()` methods.
+    """
+    POSTGRES = 0
+    SQLITE = 1
+
+    def __init__(self, engine, table_name, trigger_name):
+        self.engine = engine
+        self.table = table_name
+        self.trigger = trigger_name  # for postgres, used for both trigger and function name
+        self._set_db_type()
+
+    def create(self):
+        if self.db == Trigger.POSTGRES:
+            sql = self.get_sql_create_postgres_trigger()
+        else:
+            sql = self.get_sql_create_sqlite_trigger()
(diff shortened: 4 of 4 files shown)


jdavcs:
Jenkins is complaining about a semicolon in sql. Will look at it tomorrow.

mvdbeek:
Very interesting! I'm a bit nervous about moving this to a trigger, it might well be the right thing to do. 

Did you check the performance of updating this with a update query ? We know when and where we are changing the job state (which is different from the myriad of places where we update a history as in https://github.com/galaxyproject/galaxy/pull/8187, which is also more of a computed value), so that seems like it could be addressed by shifting
```
for dataset_assoc in job.output_datasets + job.output_library_datasets:
```
to a query ... which passes through the sqlalchemy session. One advantage being that sqlalchemy will refresh all attributes of loaded objects. This seems like it needs to be an additional step if you use triggers ? It might happen even without passing through the session, but if it doesn't we may see some odd state-related bugs.

These are testable things though, so looking forward to see where this is going.

jmchilton:
>  It might happen even without passing through the session, but if it doesn't we may see some odd state-related bugs.

Some great points @mvdbeek - thanks! My intuition is that none of the threads that change the job state would do anything with these dataset states (other than maybe update them again - e.g. in error threads). If that is the case, presumably the trigger is a bit better? Not doing the sync of the objects back after the update is to my best guess a good thing? That said if we performance data for both approaches and this wasn't any better than the stored query, I'd definitely prefer to avoid the triggers.

But this is a big, difficult choice for sure. I don't have a strong sense about the right thing to do or any hard data - other than that this is on critical paths for processing many datasets at a time.

mvdbeek:
Yeah, we may want to try both approaches ... and then there also your suggestion (hope I'm paraphrasing that correctly) to have the dataset state simply be the job state, unless a metadata error occurred (https://github.com/galaxyproject/galaxy/pull/7791#issuecomment-528369549).

jdavcs:
I will implement the sqlalchemy update query approach as an alternative then (in a separate PR). Then we can compare their performance. Marking this WIP for now.

I could use some advice on testing performance. What would be a realistic scenario in terms of load? Is it 10s/100s/1000s of datasets per job, or multiple (10s/100s/1000s?) jobs in a batch, or both? And what's a reasonable way to do that programmatically (write an api test? or print/log statements w/timing info?)

Thank you for all the feedback!

mvdbeek:
> I could use some advice on testing performance. What would be a realistic scenario in terms of load?

I don't know if the performance test needs to be a real test, I think if the query is 10 - 100 faster than the loop we should be good enough and we'd most likely be hitting other bottlenecks at that point. A simple script with 1k job outputs will do IMO.

jdavcs:
@mvdbeek thanks! But how do I launch a job programmatically? (only way I know is in the context of a test, which is why I brought it up - and I wouldn't use that for testing performance if I can just write a script)

mvdbeek:
You don't need to launch the job, you just prepare the objects in code

mvdbeek:
Answered on gitter, but to preserve it for others:
You need to setup Job and HistoryDatasetAssociation objects and call `add_output_dataset_association` with those HDAs.

An easy way to do that is with scripts/db_shell.py, to be called liked this (activate your venv first):
```
ipython -i scripts/db_shell.py -- -c config/galaxy.yml
```
This is the gist of what I had in mind:
```
j = Job()
h = HistoryDatasetAssociation()
j.add_output_dataset('name1',  h)
sa_session.add(j)
sa_session.flush()
```

jdavcs:
@mvdbeek I believe we also need a Dataset object, so:
```
j = Job()
ds = Dataset(state=Dataset.states.NEW)
h = HistoryDatasetAssociation()
h.dataset = ds
j.add_output_dataset('name1',  h)
sa_session.add(j)
sa_session.flush()
```
A very preliminary result on trigger vs python loop: my quick and dirty test shows a 10% (only!) improvement on 1000K HDAs (with associated datasets, so both hdas + datasets are updated), run locally on postgresql. This is not final and most likely I'm skipping something that happens/updates in a real scenario. I'll post my code as a gist after I check these numbers.

mvdbeek:
yeah, make sure to run `expire_all` on your session.

jdavcs:
Here's how I'm testing: https://gist.github.com/ic4f/ec83dfc75a3936cef093995dd00985a4
Results for running locally (10 jobs x 100 datasets = 1K):

postgres:
- loop: 7.25
- trigger: 4.23

sqlite:
- loop: 7.82
- trigger: 3.20 

@mvdbeek I assume you meant adding `expire_all` to the trigger-based `change_state()` code?

Two questions:
- Am I testing this correctly? I expected very different results, so I must be doing it wrong, no?
- If this indeed is the most speedup we can get (assuming the trigger code is correct), is it worth it? If it is, I'll implement the sqlalchemy one update query alternative (I assume it'll be faster than the loop but slower than the trigger.)

mvdbeek:
> @mvdbeek I assume you meant adding `expire_all` to the trigger-based `change_state()` code?

yeah, you need to add that before each benchmark.

> * Am I testing this correctly? I expected very different results, so I must be doing it wrong, no?

Looks good, but `expire_all` should have an effect on the loop. I'd like to also see that compared with the query approach.

jdavcs:
I did try to add it to each benchmark - it didn't change the result significantly. I'll update the code and numbers.
So this kind of speed up for this method (58% for the trigger approach) will indeed make a difference under heavy load?

mvdbeek:
I'm not sure ... I was expecting larger speedups. I'd still like to see the query approach as well, seeing as that'd be less intrusive.

jdavcs:
I agree about the query approach being better; it's just that I expected `change_state` to be a more significant bottleneck. I'll implement it and rerun the tests with `expire_all`.

ic4f:
Well, a raw sql query executed through the SA session gives comparable performance to the trigger:

* loop: 7.49
* query  4.39 (trigger was 4.23)

Same script: https://gist.github.com/ic4f/ec83dfc75a3936cef093995dd00985a4

I haven't tested on sqlite, but I think it is safe to assume that SA query is the better choice.

Nerdinacan:
i really like your trigger ORM extension classes! I was thinking about adding implementing something like that after I put in the view classes in my last PR.

ic4f:
@Nerdinacan awesome! Thanks :) I was initially going to ping you if/when we moved forward on the trigger option for the job state, but that's not a certainty yet. In any case, if these can be reused, partially or not, that's great. Together we can make them better, I'm sure. 
And yes, I'm following the discussion on your other PR - I'm sure we'll all share a doc and/or will zoom about it soon.

ic4f:
@Nerdinacan Re: merging update time with my trigger - i'm sure we can do that (provided that trigger will be the option we select for job state - it may or may not)

Nerdinacan:
> @Nerdinacan Re: merging update time with my trigger - i'm sure we can do that (provided that trigger will be the option we select for job state - it may or may not)

Totally. I just need a valid HDCA.update_time, and have no preference as to how we obtain it.

jmchilton:
This question is a bit of design aside but it might help me understand the client side requirements driving these change and understand what possible configurations might work on the backend. Do we need to update update_time on the HDCAs or would a new field in the API response for HDA and HDCAs (maybe called "job_source_update_time") that grabbed an update_time from the job_source work (either a Job or ImplicitCollectJobs)? I sort of made the case here (https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/jobs/__init__.py#L1380) that updating the HDA update_time when the dataset changes is a questionable design change, I would think that would apply even more so in the collection context.

I think since when we copy HDAs and HDCAs we don't copy the jobs or ImplicitCollectionJobs - it would prevent some race conditions at the edges when an HDCA or HDA was being copied while the job source update time was being updated. That increased correctness makes me think it might be a better design choice.

If we did that we could then drop that hack for HDAs also after the new client is the only client.

I guess this would though reflect my clear bias (still without data) toward reducing write contention and improving write performance at the cost of read performance.

jmchilton:
That comment was on the wrong PR - sorry all.

ic4f:
Alternative approach (SA query) implemented in #9722. Closing. 
(Note: see this PR for sqlite/postgres trigger implementation.)

PR #9516, opened 2020-03-12
Fix toolshed tests

mvdbeek:
We should really rename those options as well!


lib/tool_shed/webapp/config.py
```diff
@@ -112,6 +112,8 @@ def __init__(self, **kwargs):
         self.terms_url = kwargs.get('terms_url', None)
         self.blacklist_location = kwargs.get('blacklist_file', None)
         self.blacklist_content = None
+        self.whitelist_location = kwargs.get('blacklist_file', None)
```
nsoranzo:
```suggestion
        self.whitelist_location = kwargs.get('whitelist_file', None)
```

nsoranzo (approved):

PR #9518, opened 2020-03-12
Drop remaining Python 2.7 CircleCI tests

nsoranzo:


mvdbeek (approved):

jmchilton:
Why are we dropping the unit tests?

nsoranzo:
Because we are slowly dropping Python2 support, as announced in https://docs.galaxyproject.org/en/master/releases/19.09_announce.html#python

jmchilton:
These tests should be the last thing to go - they cost nothing and help ensure we're not breaking things without understanding what we are breaking. My understanding is that we were dropping support in terms of what we want people to deploy, we didn't all agree that Python 2 should just start breaking. Am I crazy? These tests test library functions that are used by projects that still support Python 2. I'm -0.9 on this PR.

nsoranzo:
I think that those project should also drop Python2 support or pin their dependencies to <=20.01 galaxy-* packages.
The 19.09 release notes (which were discussed and agreed) say "Galaxy 19.09 is the last release that officially supports Python 2", so we have already extended the support to 20.01.

mvdbeek:
I see this as another signal that we don't support Python 2 anymore. We have removed more important Python 2 tests already, and if downstream projects want to support Python 2 they should pin 19.09 as the last release that supports Python 2.

jmchilton:
I am completely unpersuaded by this. Same comment as https://github.com/galaxyproject/bioblend/pull/312#issuecomment-576746271 applies. There are PRs in 20.05 that need to be applied to Planemo - the Python 3 effort I supported was about making the platform more robust, portable, accessible - not about cudgeling friends of the project.

mvdbeek:
How would you like to see the transition occur ? I don't want to alienate any of our friends either, but I also think the cost of us supporting Python 2 would be significant (I concede that I don't know how many tests are broken since we disabled Python 2 testing, but in the same way we also haven't heard anything from our friends regarding the Python 3 transition).

jmchilton:
My desired transition would be that we keep Python 2 on life support and accept PRs and such that fix things, keep linting and testing, etc... until such a time as there is a significant feature that is important to the project that would cause us to break it. You're more productive than me and you tend to fix a lot more bugs than me - so I'll take you at your word that Python 2 support costs us - but for my development it doesn't/hasn't. That disparity in our productivity is why this isn't a -1 - if you feel held back then I'll let you break it. For my own tastes and my own development - I'd much rather know what we are breaking when we break it and I'd rather abandon Python 3 for a tangible, valuable feature addition that requires it. Supporting Python 3 was a good goal and a good feature, not supporting Python 2 is neither a feature nor a good goal.

jmchilton:
I've come around on this, I was wrong - I'm +0 on this now. We don't have enough bandwidth to continue to support Python 2.

nsoranzo:
@jmchilton Thanks for reconsidering! If that helps, we can restrict py27-unit tests to certain `packages/` that are needed by downstream projects?

nsoranzo:
Rebased.

PR #9520, opened 2020-03-13
Add HyPhy composite file datatype

guerler:
This PR adds a composite file datatype for the HyPhy viewer which wraps three individual files into one for easier selection in visualizations. ping @stephenshank.


mvdbeek:
Is that still WIP ?

guerler:
Yes, this is a draft of the datatype, we might add more files to the composition.

guerler:
ping @dannon, you did merge this into a branch already right?

Issue #3146, opened 2016-11-10
Feature request: Supporting SAML as External Authenticator 


abdulrahmanazab:
Hi Galaxy,

You are now supporting openID as external authenticator, which is great :) 

Now, this feature is requested by ELIXIR-EU and NeIC (Nordic e-Infrastructure Collaboration). The main benefit is to be able to support eduGAIN and Kalmar2 (both are based on SAML). Not having SAML supported is stopping us from having more public servers. We have one in Oslo 'Lifeportal' where we had to hack Galaxy to support Kalmar2 and eduGAIN, which is not the best practice.

nsoranzo:
https://en.wikipedia.org/wiki/Security_Assertion_Markup_Language

jxtx:
Thanks @nsoranzo. Why is this not best handled at the webserver proxy level by an authentication module? Is more support needed from Galaxy than just authentication here?

abdulrahmanazab:
Thanks for the response! 
@jxtx Technically the answer is 'yes'. But the issue is not only technical, but political (especially for NeIC). It is 'for some reason' a requirement that that portal application 'itself' supports SAML for it to be deployed in certain sites. What is needed from Galaxy (in this issue) is to provide support for SAML based External authenticators (similar to Galaxy's support for openID).

bgruening:
So @abdulrahmanazab are you planning to work on this :)

hexylena:
Most people who need this do so via apache/nginx, and probably solved for everyone in this thread with AAI :)

PR #9524, opened 2020-03-15
[WIP] Add SAML authentication

ksuderman:
This is a WIP to add SAML authentication to Galaxy as suggested in galaxyproject/galaxy#3146

This PR enables users to authenticate via a SAML Identify Provider (IdP) and the login process has been tested with the IdPs at samltest.id and onelogin.com.

## Discussion

***Social-Auth vs python3-saml***
The python3-saml library is used directly but it may be desirable to use the [Social Auth SAML Backend](https://python-social-auth-docs.readthedocs.io/en/latest/backends/saml.html) to be consistent with the OIDC authentication classes.  However, the Social Auth backend is just a wrapper around the python3-saml library and doesn't provide any obvious benefits other than consistency with the other authentication classes.  I would even say the Social Auth backend is more complicated to use than using the python3-saml library directly.

***URLs***
The saml.py controller uses URLs in the `/authnz/saml` url-space.  This should not be a problem unless someone plans to implement an OIDC provider named `saml`.  Is there another base URL that should be used for SAML authentication?

## TODO
- [ ] Add tests
- [x] Changes to support `enable_saml` and `saml_config_dir` fields in the galaxy.yml config.
- [ ] Test and validate the above assertion is correct
- [ ] Documentation:
  - [ ] Configuration: settings.json, generating certificates, etc.
  - [ ] Update doc/source/admin/authentication.md et al.
- [ ] Logout process
  - [ ] Maintain state information
  - [ ] User initiated logout
  - [ ] IdP initiated logout


VJalili:
It would be ideal to use Python Social Auth (PSA) for authentication (and authorization) when possible. It seems they have implemented a [SAML backend](https://github.com/python-social-auth/social-core/blob/master/social_core/backends/saml.py).

eancelet:
Hi Keith,
I'm interested in SAML authentication with Galaxy.
I see that your PR wasn't merge. 
Is it because you didn't have time to move from python3-saml library to Python Social Auth?
Or did you find another way to achieve your goal?
Thanks

ksuderman:
Amazing timing. I am just starting to update this PR again to bring it in line with the latest Galaxy code base and to fix the conflicts and failing tests.

The current Python Social Auth code in Galaxy assumes an OIDC backend so adding SAML to the mix would require at least changes to the `galaxy.authnz` module and `AuthnzManager` and `PSAAuthnz` classes.  The changes aren't difficult, but are much more invasive than this simple PR. I will coordinate with the Galaxy team to determine the best way forward.

eancelet:
Good news :)  
I'll follow your work then.
Thanks

PR #9525, opened 2020-03-16
Use mixin for config settings code shared between Galaxy and TS

jdavcs:
EDIT: this one is ready. However, as it is based on #9379 (first new commit is c2afc0a: "use mixin for common config settings"), it would be easiest to first get that one merged (pending review), after which I'll rebase this one.

Note to reviewer: as in 9379, the commits (11 total) are atomic, each focused on one type of edit. I've added commit messages where those edits are not self-explanatory. I think, looking at the  changes per commit, sequentially, would make more sense than all changes combined. 

Main theme: factor out common configuration settings into a mixin class. Much of TS's config code is duplication of code in Galaxy's main config object (there's even a TODO in the code: https://github.com/galaxyproject/galaxy/blob/dev/lib/tool_shed/webapp/config.py#L191 )

Argument for *not* moving common code into `BaseAppConfiguration`:
- `BaseAppConfiguration` contains code that is *used by the configuration code* of both apps, Galaxy and TS - i.e., base config paths, schema loading, etc.
- `CommonConfigurationMixin` will contain code that *happens to be present* in both
  Galaxy and TS, but deals with individual configuration properties.

I didn't move all common code into the mixin: there are many settings that are the same for both Galaxy and TS; however, factoring out individual lines, that are identical, would make the configuration of both apps much less readable. I think it's more important to have the individual config steps in one place (i.e., one method) for each app, even if some properties happen to have the same values.

The rest is mostly minor cleanup.

For comparison: 
- TS config before: https://github.com/galaxyproject/galaxy/blob/dev/lib/tool_shed/webapp/config.py
- TS config after: https://github.com/galaxyproject/galaxy/blob/6d6427296c9c8da9305d4e1d279b2dc66c5e76c0/lib/tool_shed/webapp/config.py

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -317,6 +317,15 @@ class CommonConfigurationMixin(object):
         """Determine if the provided user is listed in `admin_users`."""
         return user is not None and user.email in self.admin_users_list
 
+    @property
+    def sentry_dsn_public(self):
+        """
+        Sentry URL with private key removed for use in client side scripts,
+        sentry server will need to be configured to accept events
+        """
+        if self.sentry_dsn:
+            return re.sub(r"^([^:/?#]+:)?//(\w+):(\w+)", r"\1//\2", self.sentry_dsn)
+
 
 class GalaxyAppConfiguration(BaseAppConfiguration, CommonConfigurationMixin):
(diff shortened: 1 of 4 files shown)


ic4f:
WIP for now: applying some of these commits to #9379

ic4f:
Closing in favor of #9582

PR #9526, opened 2020-03-16
Rename and modify current_history_contents method

davebx:
I think the new functionality will be useful in the future, when we want to retrieve or compare history contents or history datasets from sources other than the current active history.


jmchilton:
Lovely - thanks so much!

PR #9527, opened 2020-03-16
Combine selenium tests for creating and switching histories

davebx:
Since new histories are automatically current, I think it's sufficient to test switching _back_ to the original history after creating a new one.

This also adds a way to select the individual history columns in the multi-view and the switch-to button for each history by history ID.


lib/galaxy_test/selenium/test_history_multi_view.py
```diff
@@ -95,40 +95,34 @@ def test_purge_history(self):
...
+        3. Switch back to the original history. A button should appear on the old,
+           previously created history that allows switching back to that one, and
+           the history ID should now match the ID of the history with which we
+           started.
+        '''
+        self.home()
+        original_history_id = self.current_history_id()
+        # Load the multi-view
+        self.components.history_panel.multi_view_button.wait_for_and_click()
+        # Creating a new history should automatically switch to it
```
OlegZharkov, line 113: Nice idea, although it might be still useful to keep one ```self.assert_history(original_history_id, histories_number=1)``` here, just to assert that there is only one history, before clicking on create button. Thx for test!

lib/galaxy/selenium/navigation.yml
```diff
@@ -213,6 +213,11 @@ multi_history_view:
         selector: '//button[contains(text(), "Copy")]'
 
 
+multiple_histories:
```
jmchilton: I know this isn't practiced perfectly - but I feel like the components here should be visual components in the UI - not descriptions of test things? Can you fold these under ``multi_history_view`` higher in the file maybe? The idea is to reuse this file in tours, screenshot generation, tutorial generation, etc.. so the stuff in this file should describe the UI not the tests I think.

lib/galaxy_test/selenium/test_history_multi_view.py
```diff
@@ -95,40 +95,34 @@ def test_purge_history(self):
...
+    def test_switch_history(self):
+        '''
+        1. Load the multi history view. There should be a selector for the button
+           to create a new history.
+        2. Create a new history. This *should* automatically switch to the newly
+           created history.
+        3. Switch back to the original history. A button should appear on the old,
+           previously created history that allows switching back to that one, and
+           the history ID should now match the ID of the history with which we
+           started.
```
jmchilton, line 105: ❤️

lib/galaxy/selenium/navigation.yml
davebx: Good point, it looks cleaner now.

lib/galaxy_test/selenium/test_history_multi_view.py
davebx: Oh, good point, let me add that.

jmchilton (approved):

jmchilton:
Looks great - can you merge this once the selenium tests pass on Jenkins! Thanks so much!

PR #9528, opened 2020-03-16
Make Toolshed Python 3 compatible & tests

### LINKED ISSUE #1715: Python 3 support.

mvdbeek:
Hardest part was to fix twill. My fork now passes all tests (https://github.com/mvdbeek/twill3), and I think they had been broken for many years.
Most of the Galaxy / Toolshed changes are standard bytestring / unicode issues. In particular mercurial functions expect bytestring input and return bytestrings as output.

Also moves the toolshed tests to github CI.

Close #1715 .


lib/galaxy/dependencies/pipfiles/default/pinned-dev-requirements.txt
```diff
@@ -56,7 +56,7 @@ sphinx-rtd-theme==0.4.3
 sphinx==1.8.5
 sphinxcontrib-websupport==1.1.2
 testfixtures==6.14.0
-twill==0.9.1 ; python_version < '3'
+-e git://github.com/mvdbeek/twill3.git@725889b778a783816a242de3f60f636f7f2dbabf#egg=twill
```
nsoranzo:
```suggestion
-e git+https://github.com/Cito/twill.git@427f5e647a4279dd5ac5665601ce9e78175c712f#egg=twill
```

Can you also update the Pipfile with:

```diff
diff --git a/lib/galaxy/dependencies/pipfiles/default/Pipfile b/lib/galaxy/dependencies/pipfiles/default/Pipfile
index 5cf77f6156..a11b144c28 100644
--- a/lib/galaxy/dependencies/pipfiles/default/Pipfile
+++ b/lib/galaxy/dependencies/pipfiles/default/Pipfile
@@ -25,7 +25,7 @@ Sphinx = "*"
 sphinx_markdown_tables = "*"
 sphinx_rtd_theme = "*"
 testfixtures = "*"
-twill = {version = "==0.9.1", markers = "python_version < '3'"}
+twill = {git = 'https://github.com/Cito/twill.git', ref = '2.0', editable=true}
 watchdog = "*"
 
 [packages]
```

mvdbeek: Still needs a minor update for RagioGroup handling in my fork.

mvdbeek:
I think the related tests all passed, but
```
___________ HistoriesApiTestCase.test_import_export_include_deleted ____________

self = <galaxy_test.api.test_histories.HistoriesApiTestCase testMethod=test_import_export_include_deleted>

    def test_import_export_include_deleted(self):
        history_name = "for_export_include_deleted"
        history_id = self.dataset_populator.new_history(name=history_name)
        self.dataset_populator.new_dataset(history_id, content="1 2 3")
        deleted_hda = self.dataset_populator.new_dataset(history_id, content="1 2 3")
        self.dataset_populator.delete_dataset(history_id, deleted_hda["id"])
    
        imported_history_id = self._reimport_history(history_id, history_name, wait_on_history_length=2, export_kwds={"include_deleted": "True"})
        self._assert_history_length(imported_history_id, 2)
    
        def upload_job_check(job):
            assert job["tool_id"] == "upload1"
    
        def check_ok(hda):
            assert hda["state"] == "ok", hda
            assert hda["deleted"] is True, hda
    
        self._check_imported_dataset(history_id=imported_history_id, hid=1, job_checker=upload_job_check)
>       self._check_imported_dataset(history_id=imported_history_id, hid=2, hda_checker=check_ok, job_checker=upload_job_check)

lib/galaxy_test/api/test_histories.py:230: 
lib/galaxy_test/api/test_histories.py:338: in _check_imported_dataset
    hda_checker(imported_dataset_metadata)

hda = {'accessible': True, 'annotation': None, 'api_type': 'file', 'create_time': '2020-03-17T17:10:53.421738', ...}

    def check_ok(hda):
>       assert hda["state"] == "ok", hda
E       AssertionError: {'accessible': True, 'annotation': None, 'api_type': 'file', 'create_time': '2020-03-17T17:10:53.421738', ...}
E       assert 'discarded' == 'ok'
E         - discarded
E         + ok

lib/galaxy_test/api/test_histories.py:226: AssertionError
```
Is quite odd ... I assume we should have been checking for discarded state here.

mvdbeek:
Alright, after https://github.com/galaxyproject/galaxy/pull/9528/commits/94da0f313f5a24082728af1fff226280926af584, which fixes a silly bug I introduced the tests are passing with upstream twill. So I guess this is ready to be reviewed.

nsoranzo:
Unfortunately this needs a rebase now.

lib/galaxy/tool_shed/util/hg_util.py
```diff
@@ -113,10 +113,8 @@ def get_repo_for_repository(app, repository=None, repo_path=None):
         hg,
         ui
     )
-    if repository is not None:
-        return hg.repository(ui.ui(), repository.repo_path(app))
-    if repo_path is not None:
-        return hg.repository(ui.ui(), repo_path)
+    repo_path = repo_path if repo_path else repository.repo_path(app)
```
jdavcs: In the previous version if `repository` was set, the `repo_path` argument was not checked. The new version changes this: `repository` is not checked if `repo_path` is set. And although this version sounds more logical (and elegant), there's this: if `repository` is `None` (which is possible here, right?), then this will raise an error (`repository.repo_path(app)`)

lib/tool_shed/test/base/twilltestcase.py
```diff
@@ -248,137 +163,36 @@ def login(self, email='test@bx.psu.edu', password='testuser', username='admin-us
...
-                    continue
-                controls[control.name] = control
-        # No refresh_on_change attribute found in current form, so process as usual
+        if form is not None:
+            f = form
+        else:
+            for i, f in enumerate(self.showforms()):
+                if i == form_no:
+                    break
+        controls = {c.name: c for c in f.inputs}
```
jdavcs: Is it possible that the return value of `self.showforms()` is empty? If so, `i` and `f` will not be initialized and (if `form` is `None`) line 184 will raise an error.

lib/tool_shed/test/base/twilltestcase.py
```diff
@@ -1253,21 +1059,34 @@ def install_repository(self, name, owner, category_name, install_resolver_depend
...
-            kwd = self.set_form_value(form, kwd, 'new_tool_panel_section_label', new_tool_panel_section_label)
-        submit_button_control = form.find_control(type='submit')
-        assert submit_button_control is not None, 'No submit button found for form %s.' % form.attrs.get('id')
-        self.submit_form(form.attrs.get('id'), str(submit_button_control.name), **kwd)
+            kwds['new_tool_panel_section_label'] = new_tool_panel_section_label
+        submit_button = [inp.name for inp in form.inputs if getattr(inp, 'type', None) == 'submit']
+        if len(submit_button) == 0:
+            # TODO: refactor, use regular TS install API
+            submit_kwargs = {inp.name: inp.value for inp in tc.browser.forms[0].inputs if getattr(inp, 'type', None) == 'submit'}
+            payload = {_: form.inputs[_].value for _ in form.fields.keys()}
```
jdavcs, line 1080: Very minor, completely subjective, feel free to ignore.  To me, `_` signals that the variable carries no meaning of its own and its sole purpose, if it's a counter variable, is facilitating iteration. But here it is a dictionary key. Maybe give it a generic name, just to emphasize it's not a throwaway? Fine by me either way.

jdavcs (changes requested):

lib/galaxy/tool_shed/util/hg_util.py
mvdbeek: True, but you need to pass either repo_path or repository, so this still can't happen (because if `repo_path` is passed in you wouldn't evaluate the right side). This method's gone after https://github.com/galaxyproject/galaxy/pull/9580, so let me rebase that

lib/tool_shed/test/base/twilltestcase.py
mvdbeek: Right, so you would only want to call `submit_form` if there is a form. That's the same problem you'd see in the existing code. I can raise an explicit exception to make it clearer.

lib/tool_shed/test/base/twilltestcase.py
mvdbeek: I have no preference between `k` and `_`, but I think we use `_ for _ in` more frequently than `k for k in`, so I'd tend go with that form.

nsoranzo: I am probably responsible for a good chunk of the `_`s in comprehensions, but I now realise that the generally accepted meaning of it is for a throwaway value.
So I will personally not use it in this context for new code.
Thanks for raising it @ic4f !

lib/tool_shed/test/base/twilltestcase.py
jdavcs: since it's a reasonable expectation (called only if there is a form), then I think it's fine as is.

lib/tool_shed/test/base/twilltestcase.py
jdavcs: Sounds good. And if at some point we decide to change/prettify any of such code, it would belong in a different PR, of course.

jdavcs (approved):

lib/galaxy/dependencies/pipfiles/default/Pipfile
```diff
@@ -25,7 +25,7 @@ Sphinx = "*"
 sphinx_markdown_tables = "*"
 sphinx_rtd_theme = "*"
 testfixtures = "*"
-twill = {version = "==0.9.1", markers = "python_version < '3'"}
+twill = {git = 'https://github.com/mvdbeek/twill.git', ref = 'get_galaxy_tests_working', editable=true}
```
nsoranzo:
```suggestion
twill = "*"
```

.github/workflows/toolshed.yaml
```diff
@@ -0,0 +1,25 @@
+name: Toolshed
+on: [push, pull_request]
+jobs:
+  test:
+    name: Test
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        python-version: [3.7]
```
nsoranzo, line 12: Should this use a postgres service like the previous Jenkins tests?

nsoranzo:
I've rebased the PR, condensing the changes to `lib/galaxy/dependencies/pipfiles/default/` in a single commit.
@mvdbeek Can you check if the commits I've added are OK with you?

lib/galaxy/tool_shed/galaxy_install/tool_dependencies/recipe/install_environment.py
```diff
@@ -246,18 +248,18 @@ def handle_complex_command(self, command, job_name=""):
 
     def log_results(self, command, fabric_AttributeString, file_path):
         """Write attributes of fabric.operations._AttributeString to a specified log file."""
-        mode = 'ab' if os.path.exists(file_path) else 'wb'
+        mode = 'a' if os.path.exists(file_path) else 'w'
```
mvdbeek, line 251: My thought here was that it is better to keep this as bytes, given that this is used when compiling packages, which may emit non-unicode errors. At least we'd have the unaltered log file on disk then. On the other hand we don't really use this anymore, so ... 🤷

.github/workflows/toolshed.yaml
mvdbeek: Good point, that also revealed that the install model database (if created separately) and the tool shed databases were not created automatically. I've added this in https://github.com/galaxyproject/galaxy/pull/9528/commits/dd63ce043d48083bc40a2a96ebfa4ef74e8963af.

nsoranzo (approved):

nsoranzo:
Thanks @mvdbeek, finally closing #1715 is a nice Easter present!

Issue #9352, opened 2020-02-11
Text editor tool


simonbray:
It would be useful to have a simple text editor tool in Galaxy. For example, I'm working on developing molecular dynamics tools on usegalaxy.eu; these tools use config files which often need to be modified slightly between runs. It would be really nice to be able to do this in Galaxy rather than downloading and reuploading.

The tool should work as follows: the user selects a dataset to edit, the dataset is loaded into a very simple text editor interface and the user makes some changes. Then the results are saved to a new dataset. Optionally a diff file could be saved as well.

As always I'm willing to implement this myself but would appreciate some tips where to start.

(I'm aware this is also possible using the interactive tools on usegalaxy.eu but this seems like overkill for such a simple functionality.)

simonbray:
ping @anuprulez, @OlegZharkov - maybe one of you is interested in helping? I would probably need some help with JS stuff. Björn suggested this could be integrated as a visualization.

simonbray:
(Though this poses the same problem that Pavan has currently, i.e. how to return an output from the visualization to the history as a dataset.)

bernt-matthias:
Do you have an example of such a file. Wondering if there is an alternative...

simonbray:
We've been discussing a number of alternatives. One would be to create tools based on packages like parmed or GromacsWrapper (https://github.com/ParmEd/ParmEd; https://github.com/Becksteinlab/GromacsWrapper), but I don't think this provides the required flexibility.

Another idea I had was a generic tool for manipulating text files (PR: https://github.com/bgruening/galaxytools/pull/954) but I think this is not very intuitive for the user.

OlegZharkov:
@simonbray and @pavanvidem. If you just want to export (modified) visualization to history, here is how you do it. 
https://github.com/galaxyproject/galaxy/blob/1a60ac03e477602290ab11cf0015c18efc917c73/config/plugins/visualizations/editor/templates/editor.mako#L34-L49
Maybe I will  need to make it as a separate library

simonbray:
ping @pavanvidem

OlegZharkov:
@simonbray @pavanvidem

pavanvidem:
sorry, somehow I missed this thread regardless of multiple pings. This is exactly what we were also looking for our visualization. @anuprulez is the developer of this visualization. He should understand better than me how it works. From the comments I understood that there is a way which is great!

PR #9531, opened 2020-03-17
Editor as visualization 

OlegZharkov:
Implementation of https://github.com/galaxyproject/galaxy/issues/9352
Current look:
image
Or if you hide tools and history:
image

Works as expected, you can edit your dataset and export its modified version to history (with special characters/tabs/spaces).
You're always welcome to propose changes (maybe background color, should I change it to canonical white?) 
Tags are not copied from original file. Does anyone know how to upload file already with tags? At least not an option in [upload.xml](https://github.com/galaxyproject/galaxy/blob/dev/tools/data_source/upload.xml)
Otherwise, feel free to try or just take a look at ```editor.mako```


anuprulez:
Do you really have `413` files for this visualisation?

dannon:
@anuprulez Looks like it's the build dependencies for the ace editor here; we should be able to repackage that like the other visualizations have been done recently, and rebase out the static files.  I can help follow up on that.  (not a big deal, though -- and this is very cool @OlegZharkov!)

simonbray:
This looks awesome Oleg, thanks a lot!

I am more in favour of the white background, I think it fits better with the rest of the Galaxy colour scheme.

OlegZharkov:
@anuprulez yeah, that was ace-editor. It took me a little while, but everything is squeezed in single ```script.js``` file with webpack. File structure should look way better now. 

@simonbray so I changed backgroud color to white 
image
Export button is semi-transparent, but it will be highlighted on hover

@dannon do you know if it's possible to preserve tag? Do we actually need this feature?

If someone wants to try it locally, you would need to run this now:
```
cd galaxy/config/plugins/visualizations/editor
npm install
npm run build
```

and then standart ```make client``` at galaxy root dir

Thanks a lot for testing!

simonbray:
@OlegZharkov, thanks, looks better than the dark theme IMO. Is it possible to make the export button a bit more visible against the white background?

OlegZharkov:
@simonbray sure, it's possible. But the idea was to make it semi-transparent, so it will not cover the text behind it. Example:
image

To avoid confusion, it's highlighted if you hover the mouse

image

OlegZharkov:
@galaxybot test this

OlegZharkov:
I guess, we are good to go here. A huge thanks to @dannon and @mvdbeek!
@simonbray I will create another PR to EU after it gets merged

dannon (approved):
@OlegZharkov Thanks for the work on this, it looks great, and it's working well for me!

OlegZharkov:
@dannon 
Thanks a lot !!

simonbray:
Thanks everyone!

OlegZharkov:
@simonbray I guess, you can close https://github.com/galaxyproject/galaxy/issues/9352

PR #9532, opened 2020-03-17
Eslint one-var spec / airbnbstyle

dannon:
Minor, but another small step towards airbnb-base.  This was almost always consistently followed already, so it seemed like a good idea to make it fully consistent and apply the rule.


dannon:
The failing/cancelled github-driven integration tests are unrelated to this PR.

guerler (approved):

PR #9533, opened 2020-03-18
[20.01] Free space before running tests

mvdbeek:
Otherwise kubernetes "taints" nodes with `node.kubernetes.io/disk-pressure` and cleans up old images ... it does take 5 minutes for that refresh to be signalled back, which is why the first couple of tests time out and then they start working again.

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -21,6 +21,8 @@ jobs:
         ports:
           - 5432:5432
     steps:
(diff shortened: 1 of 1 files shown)


nsoranzo (approved):

dannon:
Thanks for tracking this down; I wondered what was up with these.

jmchilton:
❤️ - amazing detective work, so impressive.

PR #9535, opened 2020-03-18
[20.01] History import/export fixes

mvdbeek:
Prior to this you could export histories with failed datasets, but on import the dataset state would be set to OK (despite the job state being correctly set to error).

The fix is to also export the HDA state. Additionally Galaxy will fallback to using the job state if the HDA state has not been serialized. Includes unit tests and API tests. The last commit should make the `test_import_export_include_deleted` test more stable.

### DIFF
--- a/lib/galaxy/model/store/__init__.py
+++ b/lib/galaxy/model/store/__init__.py
@@ -828,6 +828,7 @@ class DirectoryImportModelStore1901(BaseDirectoryImportModelStore):
         for output_key in job_attrs['output_datasets']:
             output_hda = _find_hda(output_key)
             if output_hda:
+                output_hda.state = imported_job.state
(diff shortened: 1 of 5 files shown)


lib/galaxy/model/store/__init__.py
```diff
@@ -828,6 +832,9 @@ def _connect_job_io(self, imported_job, job_attrs, _find_hda, _find_hdca):
         for output_key in job_attrs['output_datasets']:
             output_hda = _find_hda(output_key)
             if output_hda:
+                if not self.dataset_state_serialized:
+                    # dataset state has not been serialized, get state from job
+                    output_hda.state = imported_job.state
```
jmchilton, line 837: Nifty!

jmchilton (approved):

PR #9537, opened 2020-03-19
[20.01] Retry reading tool conf on IOError

mvdbeek:
I think that might stabilize the test_tool_reload* unit tests.

### DIFF
--- a/lib/galaxy/tools/toolbox/watcher.py
+++ b/lib/galaxy/tools/toolbox/watcher.py
@@ -71,6 +71,8 @@ class ToolConfWatcher(object):
 
     def __init__(self, reload_callback, tool_cache=None):
         self.paths = {}
+        self.drop_on_next_loop = set()
+        self.drop_now = set()
         self.cache = tool_cache
         self._active = False
         self._lock = threading.Lock()
@@ -131,11 +133,16 @@ class ToolConfWatcher(object):
                     # in rare cases `path` may be deleted between `os.path.exists` calls
                     # and reading the file from the filesystem. We do not want the watcher
                     # thread to die in these cases.
-                    try:
-                        del hashes[path]
-                        del paths[path]
-                    except KeyError:
-                        pass
+                    if path in self.drop_now:
+                        log.warning("'%s' could not be read, removing from watched files")
+                        try:
+                            del hashes[path]
+                            del paths[path]
+                        except KeyError:
+                            pass
+                    else:
+                        log.debug("'%s could not be read")
+                        self.drop_on_next_loop.add(path)
                     if self.cache:
                         self.cache.cleanup()
                     do_reload = True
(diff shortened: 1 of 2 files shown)


lib/galaxy/tools/toolbox/watcher.py
```diff
@@ -131,11 +133,16 @@ def check(self):
                     # in rare cases `path` may be deleted between `os.path.exists` calls
                     # and reading the file from the filesystem. We do not want the watcher
                     # thread to die in these cases.
-                    try:
-                        del hashes[path]
-                        del paths[path]
-                    except KeyError:
-                        pass
+                    if path in self.drop_now:
+                        log.warning("'%s' could not be read, removing from watched files")
```
nsoranzo:
```suggestion
                        log.warning("'%s' could not be read, removing from watched files", path)
```

lib/galaxy/tools/toolbox/watcher.py
```diff
@@ -71,6 +71,8 @@ class ToolConfWatcher(object):
 
     def __init__(self, reload_callback, tool_cache=None):
         self.paths = {}
+        self.drop_on_next_loop = set()
+        self.drop_now = set()
```
nsoranzo: Shouldn't `drop_on_next_loop` and `drop_now` be local variables inside the `check()` method (like `hashes`)?

mvdbeek: Yes, good thinking!

lib/galaxy/tools/toolbox/watcher.py
```diff
@@ -131,11 +133,16 @@ def check(self):
...
-                        del paths[path]
-                    except KeyError:
-                        pass
+                    if path in drop_now:
+                        log.warning("'%s' could not be read, removing from watched files", path)
+                        try:
+                            del hashes[path]
+                            del paths[path]
+                        except KeyError:
+                            pass
```
nsoranzo: If `path` is not in `hashes`, `path` would never be removed from `paths`.

```suggestion
                        del paths[path]
                        if path in hashes:
                            del hashes[path]
```

lib/galaxy/tools/toolbox/watcher.py
```diff
@@ -131,11 +133,14 @@ def check(self):
...
-                        del hashes[path]
+                    if path in drop_now:
+                        log.warning("'%s' could not be read, removing from watched files", path)
                         del paths[path]
-                    except KeyError:
-                        pass
+                        if path in hashes:
+                            del hashes[path]
+                    else:
+                        log.debug("'%s could not be read")
```
nsoranzo:
```suggestion
                        log.debug("'%s could not be read", path)
```

nsoranzo (approved):

nsoranzo:
@mvdbeek I've retargeted this to 20.01 (I think that was your original intention, given the branch history?) and fixed the `test_run_simple` test in py27-unit also for 20.01. Hope that's OK.

mvdbeek:
sure, thanks for that. let's hope we see a bit more green soon!

PR #9538, opened 2020-03-19
Fix citation ids and vue unit test in ToolsView

OlegZharkov:
Thanks to failed test [here](https://circleci.com/gh/galaxyproject/galaxy/77652?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link), we realized that we have spaces in ids...

 Neither HTML4 nor HTML5 allows space characters in ID attribute values. So this PR replaces all spaces with underscores. Thus we also fix failed test (it's quite sad to see unit tests failing). Should be pretty straight forward and easy to merge.


bgruening:
Awesome, thanks @OlegZharkov

PR #9541, opened 2020-03-20
[20.01] Incremental indexing for tool shed whoosh indexes

mvdbeek:
The new TS client makes intensive use of the repo and tool whoosh indexes that are built outside of the repo update cycle. For deployers to find new repos in the admin interface we need to constantly rebuild the index (or couple repo updates with index updates, since they now are effectively tied together).
Doing this puts heavy load on the tool shed (it is almost unreachable during the last couple of days ...)

This PR implements incremental indexing by ordering repos by update time. Once we reach a repo that is present in the whoosh index and that has the same update time as the repo we're trying to add we break out of the update loop. This means in most cases we just need to index 0 or 1 new repos.

This includes unit tests for the core changes. I have uploaded a small fraction of repos form the IUC to a local tool shed. Without these changes indexing took 3 seconds, with the changes it's around 300 ms,
where most time is spent initializing the toolshed model.

### DIFF
--- a/lib/tool_shed/util/shed_index.py
+++ b/lib/tool_shed/util/shed_index.py
@@ -26,6 +26,7 @@ def get_or_create_index(whoosh_index_dir):
     tool_index_dir = os.path.join(whoosh_index_dir, 'tools')
     if not os.path.exists(whoosh_index_dir):
         os.makedirs(whoosh_index_dir)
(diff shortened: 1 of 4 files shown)


gregvonkuster:
Awesome!  Thanks @mvdbeek :)

PR #9544, opened 2020-03-20
unify usage of Popen

bernt-matthias:
work toward https://github.com/galaxyproject/galaxy/issues/9251 

most of the "easy" cases are done. here is a list of the remaining grouped by the reason why its not an easy case:

- [ ] close_fds
  - lib/galaxy/containers/__init__.py
  - lib/galaxy/visualization/plugins/interactive_environments.py
  - lib/galaxy/tools/expressions/evaluation.py
  - lib/galaxy/tool_shed/galaxy_install/tool_dependencies/recipe/install_environment.py + bufsize + cwd
- [ ] use stdin
  - lib/galaxy/datatypes/converters/bgzip.py
- [ ] use returned process 
  - lib/galaxy/datatypes/dataproviders/external.py
  - lib/galaxy/util/lazy_process.py
  - lib/galaxy/jobs/runners/util/sudo.py
  - lib/galaxy_test/driver/driver_util.py + preexec_fn + cwd
- [ ] use env -> maybe shell_process
  - lib/galaxy/tool_util/deps/brew_exts.py
  - lib/galaxy/tool_util/deps/commands.py + shell
  - lib/galaxy/tool_util/deps/resolvers/modules.py
  - lib/galaxy/tool_util/deps/resolvers/lmod.py
  - lib/galaxy/util/pastescript/serve.py
  - lib/galaxy/jobs/runners/__init__.py + shell clse_fds, preexec_fn
  - lib/galaxy/jobs/runners/condor.py   + -"-
  - lib/galaxy/jobs/runners/local.py    + -"-
  - lib/galaxy/jobs/runners/util/cli/shell/local.py + -"-
- [ ] use shell=True
  - lib/galaxy/tool_util/deps/container_classes.py
  - lib/galaxy/tool_util/verify/test_data.py (because it uses strings and command chaining with &&)
  - lib/galaxy/job_metrics/collectl/cli.py
- [ ] read ony begin of stdout
  - lib/galaxy/tool_shed/galaxy_install/migrate/check.py
  - lib/galaxy/jobs/transfer_manager.py

### DIFF
--- a/lib/galaxy/datatypes/molecules.py
+++ b/lib/galaxy/datatypes/molecules.py
@@ -18,8 +18,10 @@ from galaxy.datatypes.sniff import (
 from galaxy.datatypes.tabular import Tabular
 from galaxy.datatypes.util.generic_util import count_special_lines
 from galaxy.datatypes.xml import GenericXml
-from galaxy.tool_util.deps import commands
-from galaxy.util import unicodify
+from galaxy.util import (
+    commands,
+    unicodify
+)
 
 log = logging.getLogger(__name__)
 
--- a/lib/galaxy/datatypes/molecules.py
+++ b/lib/galaxy/datatypes/molecules.py
@@ -2,7 +2,6 @@
 import logging
 import os
 import re
-import subprocess
 
 from galaxy.datatypes import (
     data,
@@ -19,6 +18,7 @@ from galaxy.datatypes.sniff import (
 from galaxy.datatypes.tabular import Tabular
 from galaxy.datatypes.util.generic_util import count_special_lines
 from galaxy.datatypes.xml import GenericXml
+from galaxy.tool_util.deps import commands
 from galaxy.util import unicodify
 
 log = logging.getLogger(__name__)
@@ -26,17 +26,18 @@ log = logging.getLogger(__name__)
 
 def count_lines(filename, non_empty=False):
     """
-        counting the number of lines from the 'filename' file
+    counting the number of lines from the 'filename' file
     """
+    if non_empty:
+        cmd = ['grep', '-cve', r'^\s*$', filename]
+    else:
+        cmd = ['wc', '-l', filename]
     try:
-        if non_empty:
-            out = subprocess.Popen(['grep', '-cve', r'^\s*$', filename], stdout=subprocess.PIPE)
-        else:
-            out = subprocess.Popen(['wc', '-l', filename], stdout=subprocess.PIPE)
-        return int(out.communicate()[0].split()[0])
-    except Exception:
-        pass
-    return 0
+        out = commands.execute(cmd)
+    except commands.CommandLineException as e:
+        log.error(str(e))
+        return 0
+    return int(out.split()[0])
 
 
 class GenericMolFile(data.Text):
--- a/lib/galaxy/datatypes/molecules.py
+++ b/lib/galaxy/datatypes/molecules.py
@@ -37,7 +37,7 @@ def count_lines(filename, non_empty=False):
     try:
         out = commands.execute(cmd)
     except commands.CommandLineException as e:
-        log.error(str(e))
+        log.error(unicodify(e))
         return 0
     return int(out.split()[0])
 
--- a/lib/galaxy/datatypes/util/generic_util.py
+++ b/lib/galaxy/datatypes/util/generic_util.py
@@ -1,4 +1,4 @@
-from galaxy.tool_util.deps import commands
+from galaxy.util import commands
 
 
 def count_special_lines(word, filename, invert=False):
--- a/lib/galaxy/datatypes/util/generic_util.py
+++ b/lib/galaxy/datatypes/util/generic_util.py
@@ -1,19 +1,18 @@
-import subprocess
+from galaxy.tool_util.deps import commands
 
 
 def count_special_lines(word, filename, invert=False):
     """
-        searching for special 'words' using the grep tool
-        grep is used to speed up the searching and counting
-        The number of hits is returned.
+    searching for special 'words' using the grep tool
+    grep is used to speed up the searching and counting
+    The number of hits is returned.
     """
+    cmd = ["grep", "-c", "-E"]
+    if invert:
+        cmd.append('-v')
+    cmd.extend([word, filename])
     try:
-        cmd = ["grep", "-c", "-E"]
-        if invert:
-            cmd.append('-v')
-        cmd.extend([word, filename])
-        out = subprocess.Popen(cmd, stdout=subprocess.PIPE)
-        return int(out.communicate()[0].split()[0])
-    except Exception:
-        pass
-    return 0
+        out = commands.execute(cmd)
+    except commands.CommandLineException:
+        return 0
+    return int(out)
--- a/lib/galaxy/jobs/runners/univa.py
+++ b/lib/galaxy/jobs/runners/univa.py
@@ -27,13 +27,12 @@ known bugs/problems:
 import logging
 import re
 import signal
-import subprocess
 import time
 
 from galaxy.jobs.runners.drmaa import DRMAAJobRunner
+from galaxy.tool_util.deps import commands
 from galaxy.util import (
-    size_to_bytes,
-    unicodify
+    size_to_bytes
 )
 
 log = logging.getLogger(__name__)
@@ -157,11 +156,10 @@ class UnivaJobRunner(DRMAAJobRunner):
         # even if this could be disambiguated by the stderr message the `qstat -u "*"`
         # way seems more generic
         cmd = ['qstat', '-u', '"*"']
-        p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-        stdout, stderr = p.communicate()
-        stdout, stderr = unicodify(stdout).strip(), unicodify(stderr).strip()
-        if p.returncode != 0 or stderr != "":
-            log.exception('`%s` returned %d, stderr: %s' % (' '.join(cmd), p.returncode, stderr))
+        try:
+            stdout = commands.execute(cmd)
+        except commands.CommandLineException as e:
+            log.error(str(e))
             raise self.drmaa.InternalException()
         state = self.drmaa.JobState.UNDETERMINED
         for line in stdout.split('\n'):
@@ -201,17 +199,15 @@ class UnivaJobRunner(DRMAAJobRunner):
         # available immediately a simple retry mechanism is implemented ..
         # max wait is approx 1min)
         while True:
-            p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-            stdout, stderr = p.communicate()
-            stdout, stderr = unicodify(stdout).strip(), unicodify(stderr).strip()
-            if p.returncode != 0:
-                if slp <= 32 and "job id {jobid} not found".format(jobid=job_id) in stderr:
-                    # log.debug('`%s` returned %s, stderr: %s => retry after %ds' % (' '.join(cmd), p.returncode, stderr, slp))
+            try:
+                stdout = commands.execute(cmd)
+            except commands.CommandLineException as e:
+                if slp <= 32 and "job id {jobid} not found".format(jobid=job_id) in e.stderr:
                     time.sleep(slp)
                     slp *= 2
                     continue
                 else:
-                    log.exception('`%s` returned %s, stderr: %s' % (' '.join(cmd), p.returncode, stderr))
+                    log.exception(str(e))
                     return self.drmaa.JobState.UNDETERMINED
             else:
                 break
--- a/lib/galaxy/jobs/runners/univa.py
+++ b/lib/galaxy/jobs/runners/univa.py
@@ -30,8 +30,8 @@ import signal
 import time
 
 from galaxy.jobs.runners.drmaa import DRMAAJobRunner
-from galaxy.tool_util.deps import commands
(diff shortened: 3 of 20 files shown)


lib/galaxy/datatypes/converters/bgzip.py
```diff
@@ -6,11 +6,12 @@
 usage: %prog in_file out_file
 """
 import optparse
-import subprocess
 import tempfile
 
 import pysam
 
+from galaxy.util import commands
```
mvdbeek: I don't think we'll want to add another dependency here, this file is run by external tools.

lib/galaxy/datatypes/util/generic_util.py
```diff
@@ -1,19 +1,19 @@
...
-        out = subprocess.Popen(cmd, stdout=subprocess.PIPE)
-        return int(out.communicate()[0].split()[0])
-    except Exception:
-        pass
-    return 0
+        out = commands.execute(cmd)
+    except commands.CommandLineException:
+        return 0
+    # TODO is split necessary here? seems that grep -c just returns the number
+    return int(out.split()[0])
```
mvdbeek: int works fine even if there's whitespaces in the string, let's drop this

lib/galaxy/jobs/runners/univa.py
```diff
@@ -157,11 +156,10 @@ def _get_drmaa_state_qstat(self, job_id, extinfo):
...
         cmd = ['qstat', '-u', '"*"']
-        p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-        stdout, stderr = p.communicate()
-        stdout, stderr = unicodify(stdout).strip(), unicodify(stderr).strip()
-        if p.returncode != 0 or stderr != "":
-            log.exception('`%s` returned %d, stderr: %s' % (' '.join(cmd), p.returncode, stderr))
+        try:
+            stdout = commands.execute(cmd)
+        except commands.CommandLineException as e:
+            log.exception(str(e))
```
mvdbeek: That's going to log the traceback and the nicely formatted `CommandLineException `, which is redundant. Maybe use `log.error` or just `log.exception(e)`?

lib/galaxy/util/commands.py
```diff
@@ -110,9 +111,11 @@ def argv_to_str(command_argv, quote=True):
     return " ".join(map_func(c) for c in command_argv if c is not None)
 
 
-def _wait(cmds, **popen_kwds):
+def _wait(cmds, input=None, **popen_kwds):
     p = subprocess.Popen(cmds, **popen_kwds)
-    stdout, stderr = p.communicate()
+    stdout, stderr = p.communicate(input)
+    if not six.PY2:
+        stdout, stderr = unicodify(stdout).strip(), unicodify(stderr).strip()
```
mvdbeek:
```suggestion
    stdout, stderr = unicodify(stdout).strip(), unicodify(stderr).strip()
```
We want unicode everywhere. This is fine for python 2 as well.

mvdbeek:
This is a nice improvement and I think it's pretty much ready in its current state and should fix some borderline cases on python 3. Do you think you can take this out of WIP so we can get it into 20.05 ?

lib/galaxy/jobs/runners/univa.py
mvdbeek: Actually this is redundant either way:
```python
In [15]: try:
    ...:     raise commands.CommandLineException('echo', 'bla', 'blurb', 1)
    ...: except Exception as e:
    ...:     log.exception(e)
    ...:
    ...:
Failed to execute command-line echo, stderr was:
-------->>begin stderr<<--------
blurb
-------->>end stderr<<--------
-------->>begin stdout<<--------
bla
-------->>end stdout<<--------
  File "<ipython-input-15-8f4a354ec6e3>", line 2, in <module>
    raise commands.CommandLineException('echo', 'bla', 'blurb', 1)
galaxy.tool_util.deps.commands.CommandLineException: Failed to execute command-line echo, stderr was:
-------->>begin stderr<<--------
blurb
-------->>end stderr<<--------
-------->>begin stdout<<--------
bla
-------->>end stdout<<--------
```

Given the raise in the line below I'd maybe go with log.error(str(e)),
which would given you something like
```python
In [16]: try:
    ...:     raise commands.CommandLineException('echo', 'bla', 'blurb', 1)
    ...: except Exception as e:
    ...:     log.error(str(e))
    ...:
    ...:
Failed to execute command-line echo, stderr was:
-------->>begin stderr<<--------
blurb
-------->>end stderr<<--------
-------->>begin stdout<<--------
bla
-------->>end stdout<<--------
```

bernt-matthias:
> Do you think you can take this out of WIP so we can get it into 20.05 ?

Sure. I will work on your suggestions today.

lib/galaxy/util/commands.py
bernt-matthias: Actually I took this from here: https://github.com/galaxyproject/galaxy/blob/25f33300ab737358fc033d6e5aed00cd700a7216/lib/galaxy/util/commands.py#L42 

should I also change this?

lib/galaxy/datatypes/converters/bgzip.py
bernt-matthias: Should I: 

- just drop the change? or
- add unicodify here... but I guess this also comes with another dependency or
- add a python2 requirement to the tools using this script (I guess they could need it anyway)

lib/galaxy/jobs/runners/univa.py
bernt-matthias: done. did the same for drmaa.py

lib/galaxy/util/commands.py
mvdbeek: IMO yes, but I do not understand why we did this in the first place. Both python 2 and python 3 seem to support writing unicode to stdout/stderr. @nsoranzo any advice here ?

lib/galaxy/datatypes/converters/bgzip.py
mvdbeek: I think the current code works fine on python 3 (we run converter tests now).
I wouldn't touch this, or if you do I would replace this file and just put the command in cheetah.

nsoranzo: > I wouldn't touch this, or if you do I would replace this file and just put the command in cheetah.

It should be possible by replacing `pysam.tabix_compress()` with `bgzip` from the `htslib` package.

bernt-matthias: Then I will undo these changes. replacing pysam sounds like a separate PR -- what would be the advantage?

bernt-matthias: > or if you do I would replace this file and just put the command in cheetah

since 4-5 tools use this python script it seems that this would result in quite a bit of code duplication.

mvdbeek: Yes, I think that's for another PR. I guess getting rid of the additional wrapper is the main advantage.

mvdbeek: > since 4-5 tools use this python script it seems that this would result in quite a bit of code duplication.

These don't need to be separate the way they are now. You can use one converter for many inputs.

lib/galaxy/datatypes/molecules.py
```diff
@@ -19,24 +18,28 @@
...
-            out = subprocess.Popen(['grep', '-cve', r'^\s*$', filename], stdout=subprocess.PIPE)
-        else:
-            out = subprocess.Popen(['wc', '-l', filename], stdout=subprocess.PIPE)
-        return int(out.communicate()[0].split()[0])
-    except Exception:
-        pass
-    return 0
+        out = commands.execute(cmd)
+    except commands.CommandLineException as e:
+        log.error(str(e))
```
nsoranzo:
```suggestion
        log.error(unicodify(e))
```

mvdbeek: Is that necessary ? `CommandLineException` already ensures that user-input is unicode

lib/galaxy/util/commands.py
```diff
@@ -110,9 +111,10 @@ def argv_to_str(command_argv, quote=True):
     return " ".join(map_func(c) for c in command_argv if c is not None)
 
 
-def _wait(cmds, **popen_kwds):
+def _wait(cmds, input=None, **popen_kwds):
     p = subprocess.Popen(cmds, **popen_kwds)
-    stdout, stderr = p.communicate()
+    stdout, stderr = p.communicate(input)
+    stdout, stderr = unicodify(stdout).strip(), unicodify(stderr).strip()
```
nsoranzo: I'm not sure we should `strip()` here, maybe sometime spaces are significant?

lib/galaxy/datatypes/molecules.py
nsoranzo: Mainly for consistency, since we did a pass recently adding `unicodify()` around all exceptions in `log.method("Message %s" % e)` calls. Also I don't think we make sure that the command part of the `CommandLineException` message is Unicode.

lib/galaxy/util/commands.py
nsoranzo: I think this should be kept as it is now in the PR, i.e. unicodify only on Python3 inside `redirect_aware_commmunicate()`, unicodify always inside `_wait()`.
The redirect code was written mainly for planemo and it's safer to keep as it is.

lib/galaxy/datatypes/converters/bgzip.py
bernt-matthias: Made a note here https://github.com/galaxyproject/galaxy/issues/9672

lib/galaxy/util/commands.py
bernt-matthias: ok

lib/galaxy/datatypes/molecules.py
bernt-matthias: Done, here and in all other places changed by this PR.

lib/galaxy/util/commands.py
bernt-matthias: Done and strip restored at original places.

mvdbeek:
Ah, that test error is unrelated, I think we need to wait until the tools search index is done before returning the response.

mvdbeek (approved):

lib/galaxy/jobs/runners/univa.py
```diff
@@ -157,11 +157,10 @@ def _get_drmaa_state_qstat(self, job_id, extinfo):
         # even if this could be disambiguated by the stderr message the `qstat -u "*"`
         # way seems more generic
         cmd = ['qstat', '-u', '"*"']
-        p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-        stdout, stderr = p.communicate()
-        stdout, stderr = unicodify(stdout).strip(), unicodify(stderr).strip()
-        if p.returncode != 0 or stderr != "":
-            log.exception('`%s` returned %d, stderr: %s' % (' '.join(cmd), p.returncode, stderr))
+        try:
+            stdout = commands.execute(cmd).split()
```
nsoranzo:
```suggestion
            stdout = commands.execute(cmd).strip()
```

lib/galaxy/jobs/runners/univa.py
```diff
@@ -201,17 +200,15 @@ def _get_drmaa_state_qacct(self, job_id, extinfo):
...
+            try:
+                stdout = commands.execute(cmd).strip()
+            except commands.CommandLineException as e:
+                if slp <= 32 and "job id {jobid} not found".format(jobid=job_id) in e.stderr:
                     time.sleep(slp)
                     slp *= 2
                     continue
                 else:
-                    log.exception('`%s` returned %s, stderr: %s' % (' '.join(cmd), p.returncode, stderr))
+                    log.exception(unicodify(e))
```
nsoranzo:
```suggestion
                    log.error(unicodify(e))
```

nsoranzo (approved):

nsoranzo:
Thanks @bernt-matthias , really nice refactoring!

nsoranzo:
@jmchilton This is going to break planemo when we publish the next galaxy-tool-util packages on PyPI.

mvdbeek:
Does it make sense to provide a `command` module in tool_util that just imports from galaxy-util ?
We could emit a deprecation warning in this fallback module

nsoranzo:
> Does it make sense to provide a `command` module in tool_util that just imports from galaxy-util ?
> We could emit a deprecation warning in this fallback module

Alternatively in planemo and pulsar we can either pin the package release or try import as in https://github.com/galaxyproject/pulsar/blob/master/pulsar/managers/queued_external_drmaa.py#L9-L12

jmchilton:
I think we need to be better about providing a stable API - at least where it is pretty reasonable to do so. I'd definitely be in favor of the re-import idea - it will work for older Planemo and Pulsars also.

bernt-matthias:
Does planemo and pulsar functionality rely on `execute`, `_wait`, and `CommandLineException`? One could also have a module in util containing these three functions and keep everything else in the original place. As far as I see the three functions are independent of the rest.

mvdbeek:
> Does planemo and pulsar functionality rely on `execute`, `_wait`, and `CommandLineException`?

Yes

> One could also have a module in util containing these three functions and keep everything else in the original place.

That would still mean imports in planemo would have to change, which is what we try to avoid for packages that are using galaxy-util, galaxy-tool-util etc. I think this is what John meant by 

> we need to be better about providing a stable API

PR #9546, opened 2020-03-20
Implementation of Selenium test for signout from user preferences

Realjuan:
Implementation of Selenium test for to signout from user preferences page which is on the todo list at #3241
Registers a user, navigates to the user preferences, clicks on signout and cancels, makes sure the email is still the same, click on signout and confirms and checks to see if user has been signout.


lib/galaxy/selenium/navigation.yml
```diff
@@ -70,6 +70,7 @@ masthead:
 
 preferences:
   selectors:
+    sign_out: "#edit-preferences-custom-builds"
```
bgruening: This ID looks wrong. custom-builds should not be the ID for sign-out.

Realjuan: I will look for the proper way to change the id and fix it. I will also check the other ids on that page to see if its the only one.

Realjuan: Id has been changed, and I will ask @OlegZharkov to double check that there is no side effect.

OlegZharkov:
@jmchilton @davebx 
We really need to find out a way to bundle user preferences in a separate directory

lib/galaxy_test/selenium/test_sign_out.py
```diff
@@ -0,0 +1,22 @@
...
+
+class SignOutTestCase(SeleniumTestCase):
+    @selenium_test
+    def test_sign_out(self):
+        email = self._get_random_email()
+        self.register(email)
+        self.click_masthead_user()
+        self.components.masthead.preferences.wait_for_and_click()
+        self.components.preferences.sign_out.wait_for_and_click()
+        cancel_button = self.driver.find_element_by_xpath("//button[@id='button-0']")
```
OlegZharkov: Very nice test, especially ```assert not self.is_logged_in()``` very cool. I have just a minor notice. Can you make something similar to ```libraries.folder.import_datasets_ok_button``` in ```navigation.yaml``` instead of xpath here, please?

lib/galaxy/selenium/navigation.yml
```diff
@@ -88,6 +89,11 @@ change_user_address:
       type: xpath
       selector: '//span[contains(text(), "Insert Address")]'
 
+sign_out:
+  selectors:
+    cancel_button: '.modal-footer .buttons #button-0'
+    sign_out_button: '.modal-footer .buttons #button-0'
```
OlegZharkov: Hey, I don't think button id is correct here

OlegZharkov:
@galaxybot test this

OlegZharkov (approved):
Thanks a bunch for your fixes @Realjuan. That's an awesome test.

jmchilton (approved):

jmchilton:
Amazing - thanks so much!

OlegZharkov:
thanks @Realjuan :+1:

Issue #8375, opened 2019-07-24
"null" string input in workflow ui does not persist


crutching:
I was trying to use the string "null" for one of my "text" tool params in the workflow UI.  Upon saving the workflow, then reentering the workflow editor, the input box is blanked out.  In the resulting workflow json, entry appears as empty quotes.  Was able to reproduce for multiple text parameters.

martenson:
I suspect javascript `null` value masking this at some point.

crutching:
One other note, I am currently using v19.05.  In a much older instance running off of v15.05, this issue did not exist.

mvdbeek:
Looking at this I see that the payload coming in via `PUT /api/workflows/{id}` already contains `""` as the parameter value when you enter the string `null`, so probably something is going wrong on the client side.

dannon:
xref #9484, which was old WIP exploring what's going on here.

PR #9549, opened 2020-03-22
[20.01] Fix setting annotation that contains only integers

mvdbeek:
This fixes https://sentry.galaxyproject.org/sentry/main/issues/572478/:
```
TypeError: argument cannot be of 'int' type, must be of text type
  File "galaxy/web/framework/decorators.py", line 157, in decorator
  File "galaxy/webapps/galaxy/controllers/dataset.py", line 353, in set_edit
    annotation = sanitize_html(payload.get('annotation'))
  File "galaxy/util/sanitize_html.py", line 48, in sanitize_html
    return bleach.clean(htmlSource, **kwd)
  File "bleach/__init__.py", line 84, in clean
    return cleaner.clean(text)
  File "bleach/sanitizer.py", line 162, in clean
    raise TypeError(message)
```

That said it looks like the client switches types based on what the user enters in what should be a text field ... that seems bad. This looks similar to what happens when you enter a literal null (xref https://github.com/galaxyproject/galaxy/issues/8375).

### DIFF
--- a/lib/galaxy/util/sanitize_html.py
+++ b/lib/galaxy/util/sanitize_html.py
@@ -3,6 +3,9 @@ HTML Sanitizer (lists of acceptable_* ripped from feedparser)
 """
 import bleach
 
+from galaxy.util import unicodify
+
+
 _acceptable_elements = ['a', 'abbr', 'acronym', 'address', 'area', 'article',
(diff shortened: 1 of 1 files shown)


mvdbeek:
Thanks @dannon !

PR #9551, opened 2020-03-22
Disable workflow parameter switch for data inputs, augment form handling

guerler:
This PR disables the workflow parameter double-arrow switch for data inputs, since those are already attached to the workflow nodes. Additionally it fixes switching between workflow tool forms and the reports help in the editor panel. The customization flow for workflow tool forms has been revised such that it does not require additional callbacks. The PR also removes `jquery` from the `workflow-forms` class.

### DIFF
--- a/client/galaxy/scripts/components/Workflow/Editor/EditorPanel.vue
+++ b/client/galaxy/scripts/components/Workflow/Editor/EditorPanel.vue
@@ -6,11 +6,11 @@
             </div>
         </div>
         <div class="unified-panel-body workflow-right">
-            <div v-if="canvas" class="m-1">
+            <div class="m-1" v-show="canvas">
                 <slot name="attributes" />
                 <div id="right-content" class="right-content" />
             </div>
-            <ReportHelp v-else />
+            <ReportHelp v-show="!canvas" />
         </div>
     </div>
 </template>
--- a/client/galaxy/scripts/mvc/workflow/workflow-forms.js
+++ b/client/galaxy/scripts/mvc/workflow/workflow-forms.js
@@ -1,4 +1,4 @@
-import $ from "jquery";
+import axios from "axios";
 import { getAppRoot } from "onload/loadConfig";
 import { getGalaxyInstance } from "app";
 import _l from "utils/localization";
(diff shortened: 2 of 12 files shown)


client/galaxy/scripts/mvc/workflow/workflow-forms.js
```diff
@@ -11,95 +11,85 @@ export class DefaultForm {
...
+        });
     }
-    _customize(form) {
-        var options = form.model.attributes;
-        Utils.deepeach(options.inputs, input => {
+    _customize(node) {
+        const inputs = node.config_form.inputs;
+        Utils.deepeach(inputs, input => {
             if (input.type) {
-                input.connectable = true;
```
mvdbeek, line 97: Did you check if that is maybe required for connecting subworkflows ?

guerler: Thank you for reviewing. Is this related to subworkflows? I think this only affects individual input elements in the workflow tool forms.

Now:

Screen Shot 2020-03-22 at 3 31 26 PM

Before:

Screen Shot 2020-03-22 at 3 31 26 PM

mvdbeek: Alright, makes sense!

client/galaxy/scripts/components/Workflow/Editor/EditorPanel.vue
```diff
@@ -6,11 +6,11 @@
             </div>
         </div>
         <div class="unified-panel-body workflow-right">
-            <div v-if="canvas" class="m-1">
+            <div class="m-1" v-show="canvas">
```
dannon, line 9: Since canvas is a prop that won't be frequently changing dynamically, v-if makes more sense here unless I'm missing something.  v-show still renders, it just sets display:hidden.  We don't even want to render if canvas is falsy

guerler: In principle we would prefer `v-if` but in this case, we want to make sure that the tool forms are preserved when switching to the workflow reports view. At least as long as they are being populated by the current utility function.

dannon: Ah-ha!  The forms, okay, yeah -- those are pretty heavyweight and we should preserve the elements.  Makes sense, thanks!

guerler: Thank you for the review.

guerler:
Thanks for the review!

PR #9552, opened 2020-03-22
Prettier 2.0

dannon:
Upgrades to the new major release of prettier.  Full notes here (https://prettier.io/blog/2020/03/21/2.0.0.html) but lots of nice tweaks like standardized dangling commas (always on -- makes for clearer diffs and fewer merge conflicts when adjusting a comma separate dlist) and constant parens around fat arrow function args.

Still doesn't handle vue template html well, which is what I was most hopeful for, but they're still working on it.


jmchilton:
Love trailing commas - huge +1 from me! Thanks @dannon

OlegZharkov:
That is nice. Is it possible to configure prettier to process also Python files?

dannon:
@OlegZharkov Not with Prettier, but there are other formatters available for python like https://black.readthedocs.io/en/stable/.  I had a PR doing this at one point, but we were much more set in our ways for the python codebase and there was enough pushback that we decided to revisit it later.

Issue #9539, opened 2020-03-19
Paramiko shell plugin doesn't load system host keys


eancelet:
Hi Paramiko Shell plugin users,
As the default missing host key policy changed since 20.01 (from AutoAddPolicy to RejectPolicy), I realized that my system host key file isn't used by Galaxy.  
If I add self.ssh.load_system_host_keys() in ParamikoShell.connect(), it works.   
Is it a lack in Galaxy's code or did I miss something?
Thanks

mvdbeek:
Hi @eancelet,

I changed this intentionally (use of `AutoAddPolicy` is a bit of a security issue), but forgot to update the documentation: 
`<param id="paramiko_strict_host_key_checking">False</param>` would switch this to `paramiko.WarningPolicy()` which should work fine, but ideally you'd manually accept the host keys once for the Galaxy user.

eancelet:
Hi Marius,  
Sorry for the response time.  
Hi understand the policy change for security reason.  But do you make Galaxy aware how the hosts registered on the Galaxy user /home?
I had to implicitely  
https://github.com/eancelet/galaxy/commit/521861bbe2e9c754533620256a42d6fd33ef6b4c

mvdbeek:
That looks good, can you open a PR with this commit ?

PR #9553, opened 2020-03-23
[20.01] paramiko: load system host keys

eancelet:
Related to issue #9539


mvdbeek:
Awesome, thank you so much @eancelet!

PR #9554, opened 2020-03-24
Abstract ObjectStore interface implementation 

VJalili:
This PR is a follow-up of a [discussion](https://github.com/galaxyproject/galaxy/pull/4840#discussion_r316298166) on the [user-based ObjectStore PR](https://github.com/galaxyproject/galaxy/pull/4840). 

Design Objective
---
The goal is to abstract ObjectStore implementation so that an _instance_ of it can multiplex method invocations depending on input arguments. For instance, store data belonging to `user_1` on `local_path_1`, otherwise use `local_path_2`.  For this example, an instance of ObjectStore is initialized to use `local_path_2` and [is scoped to the lifetime of the Galaxy instance](https://github.com/galaxyproject/galaxy/blob/f51a12ba7f02f1d8b7aa9a414c00d6e9bdfebc71/lib/galaxy/config/__init__.py#L1101-L1103), while a new instance of ObjectStore should be initialized to persist data belonging to `user_1` on `local_path_1` with the lifetime of request execution. The objective is to provide such functionality with the least possible footprint on the higher-up in call hierarchy, borrowing logics from the _facade design pattern_.

Implementation Note
---

The [`ObjectStore`](https://github.com/galaxyproject/galaxy/blob/f51a12ba7f02f1d8b7aa9a414c00d6e9bdfebc71/lib/galaxy/objectstore/__init__.py#L41) type acts as an interface and a base type for all derived ObjectStore types (e.g., [`ConcreteObjectStore`](https://github.com/galaxyproject/galaxy/blob/f51a12ba7f02f1d8b7aa9a414c00d6e9bdfebc71/lib/galaxy/objectstore/__init__.py#L255) type).  This PR breaks the `ObjectStore` type to `IObjectStore` and `BaseObjectStore`, respectively defining the interface and a base implementation of the interface.  All the derived types are the modified to inherit from `BaseObjectStore` instead of `ObjectStore`. 

The interface implementation in `BaseObjectStore` uses a method name [`_invoke`](https://github.com/VJalili/galaxy/blob/01548f714444ed66ccc843d83c78b635fbb681b8/lib/galaxy/objectstore/__init__.py#L281-L282
), defined as it follows.

```python
def _invoke(self, delegate, obj, **kwargs):
    return self.__getattribute__("_" + delegate)(obj, **kwargs)
```

This method can then be extended (e.g., in #4840) to implement runtime multiplexing; for instance (pseudo implementation): 

```python 
class BaseObjectStore(IObjectStore):
    def _invoke(self, delegate, obj, **kwargs):
        if self.belongs_to("user_1", obj):
            # create a new instance of objectstore for user 1, 
            # and call its create method.
            instance = self.get_object_store_instance_for("user_1")
            return instance.__getattribute__("_" + delegate)(obj, **kwargs) 
        else:
            # call the create method on the backend defined for 
            # the instance-wide objectstore.
            return self.__getattribute__("_" + delegate)(obj, **kwargs)

    def create(self, obj, **kwargs):
        return self._invoke('create', obj, **kwargs)

# somewhere higher call in the call hierarchy, e.g., in job execution
self.object_store.create(dataset)
```

The following changes are made to ensure the `_invoke` method of `BaseObjectStore` is called for all the methods of `IObjectStore` interface:

- `BaseObjectStore` provides an implementation of all the methods of `IObjectStore`, all the a common implementation as: 

    ```python
    return self._invoke('method_name', obj, **kwargs)
    ```
- the interface implementation in `BaseObjectStore` is _enforced to be "final"_, i.e., the method implementations cannot be overridden in the derived types. 

- interface method names in the derived types are prefixed by an underscore (`_`), so to differentiate between base and derived implementation. For instance: 

    ```python
    class IObjectStore(object):
        @abc.abstractmethod
        def create(self, obj, ...):
            raise NotImplementedError()

    class BaseObjectStore(IObjectStore):
        def _invoke(self, delegate, obj, **kwargs):
            return self.__getattribute__("_" + delegate)(obj, **kwargs)

        def create(self, obj, **kwargs):
            return self._invoke('create', obj, **kwargs)

    class ConcreteObjectStore(BaseObjectStore):
        pass

    class DiskObjectStore(ConcreteObjectStore):
        def _create(self, obj, **kwargs):
            # method implementation.
    ```
 
ping @jmchilton @blankenberg


lib/galaxy/objectstore/__init__.py
```diff
@@ -38,9 +39,10 @@
 log = logging.getLogger(__name__)
 
 
-class ObjectStore(object):
+class IObjectStore(object):
```
jmchilton: I'd just call it ObjectStore still - we haven't prefixed other interfaces in Galaxy with I.

jmchilton:
This looks like a fine refactoring to me, thanks! I can see this as being the basis of something more generic and clean than the Distributed object store stuff potentially. Can you rebase with the recent changes to irods.

jmchilton (approved):

lib/galaxy/objectstore/__init__.py
VJalili: I refactored this back to `ObjectStore`.

VJalili:
@jmchilton Thank you. I implemented the changes that you suggested.

PR #9559, opened 2020-03-24
It migrate from EU

astrovsky01:


jmchilton:
It all looks good to me - using the modern, supported environment variable injection stuff. I don't see any security issues - I think this is a totally fine thing to ship with Galaxy until it has a real home. Hopefully this helps reduce the divergence for the EU.

PR #9566, opened 2020-03-26
Do not open hgweb_config in binary mode

jdavcs:
Fixes TS error on clean startup under Python3: 

Same as: #5639
Under Python3, strings can not be written to byte mode file handlers.
writing resulted in:
TypeError: a bytes-like object is required, not 'str'


mvdbeek:
Let's merge https://github.com/galaxyproject/galaxy/pull/9528 instead, please ?

ic4f:
@mvdbeek absolutely! Was about to comment that there seems to be more.. Closing in favor of #9528

PR #9567, opened 2020-03-26
Move toolshed makos to toolshed template directory

guerler:
Test PR to further separate Galaxy from Toolshed makos.


jmchilton:
This seems backward - I believe we are trying to separate them in such a way that Tool Shed depends on Galaxy not Galaxy on the Tool Shed (https://github.com/galaxyproject/galaxy/pull/8949). Maybe it is just the naming of "webapps/tool_shed" that is uncomfortable here. That component isn't depending on a tool shed webapp - the tool shed webapp makos were all moved in https://github.com/galaxyproject/galaxy/pull/8958 - all the stuff left in here is Galaxy's copy of Galaxy stuff.

guerler:
These templates are not used by Galaxy, they are artifacts of the Toolshed integration into Galaxy and are used only in Toolshed tests. Additionally these files were simply duplicated and now exist in Galaxy directories and Toolshed directories at the same time. This is confusing, at least we should move these Toolshed injections to a separate directory. It will also important to understand what the remaining conflicts and dependencies are.

jmchilton:
Galaxy isn't just a client JS application - my understanding is these makos are used by the controller in Galaxy that is used to run the tool shed tests from the Galaxy side. These tests test both Galaxy and the Tool Shed. 

> Additionally these files were simply duplicated and now exist in Galaxy directories and Toolshed directories at the same time. This is confusing

The ones used by the Galaxy application in Galaxy are in templates and the tool shed web application has its own templates else where. The ones in Galaxy's templates directory are only needed until the tests are all ported to use Galaxy API or to use the new web client.
 
> at least we should move these Toolshed injections to a separate directory to indicate their special nature.

That would be fine with me - any refactoring to get the word webapp out the name would also clarify things a bit I think.

guerler:
That sounds good, thanks for looking into it. I'll revisit this PR and then move these files into a subdirectory. It has not a very high priority right now.

PR #9568, opened 2020-03-26
Check if path is set before joining

jdavcs:
If a config property is not set (is null), then resolving it w.r.t a parent using any of the `_in_foo_dir()` helpers will raise an error as all of them use `os.path.join` that expects a string. To address this, one can set any such path to `''`, like this: https://github.com/galaxyproject/galaxy/blob/dev/lib/tool_shed/webapp/config.py#L153
EDIT: Or check every such path before calling the helper.

Checking this centrally (as done in this PR) solves not only the issue of duplication, ~~but makes the code conceptually clean: if `foo` belongs in directory `bar` we shouldn't have to set it to anything, least of all an empty string, just to have it resolve correctly.~~ EDIT: as per code review: we won't assume this if (a) the path is not set by the user; and (b) a default is not specified in the schema.

I encountered this issue while running a clean install after #9379 . That PR adds `in_root_dir`; I'll  rebase it after this is merged.

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -246,16 +246,19 @@ class BaseAppConfiguration(object):
             resolve(key)
 
     def _in_managed_config_dir(self, path):
-        return os.path.join(self.managed_config_dir, path)
+        return self._in_dir(self.managed_config_dir, path)
 
     def _in_config_dir(self, path):
-        return os.path.join(self.config_dir, path)
+        return self._in_dir(self.config_dir, path)
 
     def _in_sample_dir(self, path):
-        return os.path.join(self.sample_config_dir, path)
+        return self._in_dir(self.sample_config_dir, path)
 
     def _in_data_dir(self, path):
-        return os.path.join(self.data_dir, path)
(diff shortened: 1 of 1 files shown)


nsoranzo:
What is the config that had this issue in #9379 ? Doesn't it have a default value? I think that most if not all path should have a default, maybe there is a reason to set it to `None`? In that case, `_in_dir()` could return `None` instead of `_dir`.

ic4f:
@nsoranzo it was `hgweb_config_dir` - which used to be set to `''` by default for this very reason (the line I linked to in the description). Also, we recently discussed that all paths should be made absolute paths (@natefoo was on that discussion, I think). I'm afraid it's all too easy to forget that a path has no default in the schema (and so is set to null) which would cause the same error when we try to make such a path absolute.
That said, maybe a path should indeed remain `None` if it's not set.  So, the logic would be "`foo` resolves to `bar` if set OR if a default is specified;  otherwise it remains `None`".  Does this sound right? In which case let's do what you suggest.

nsoranzo (approved):

ic4f:
Thank you for reviewing, @nsoranzo !

PR #9569, opened 2020-03-26
Tool recommendations in workflow editor

anuprulez:
This PR:
- extends the integration of tool recommendation feature by adding it in the workflow editor:
wf-recomm

- refactors tool recommendation D3 tree to resize automatically with window's size

77862691-660d7f00-721d-11ea-930d-1e3bdf28da98

77862690-6574e880-721d-11ea-8529-d859703391d6

77862689-6443bb80-721d-11ea-9e73-371350951e94

- clips long tool names and show their full names in the respective tooltip
- maintains the same color as used in the workflow editor

ping @guerler


bgruening:
@guerler said we should ping him for a review :)

client/galaxy/scripts/components/Workflow/Editor/WorkflowRecommendationsVue.vue
```diff
@@ -0,0 +1,125 @@
...
+            showLoading: true
+        };
+    },
+    created() {
+        this.loadRecommendations();
+    },
+    computed: {},
+    methods: {
+        async loadRecommendations() {
+            const response = await axios.post(`${getAppRoot()}api/workflows/get_tool_predictions`, {
```
guerler: Axios calls should be placed into the service.js file.

client/galaxy/scripts/components/Workflow/Editor/utilities.js
```diff
@@ -7,6 +7,7 @@ import WorkflowIcons from "components/Workflow/icons";
 import { DefaultForm, ToolForm } from "mvc/workflow/workflow-forms";
 import { loadWorkflow } from "./services";
 import { hide_modal, show_message, show_modal } from "layout/modal";
+import Modal from "mvc/ui/ui-modal";
```
guerler: Where is this used?

client/galaxy/scripts/components/Workflow/Editor/workflowRecommendations.js
```diff
@@ -0,0 +1,74 @@
+import $ from "jquery";
```
guerler: Everything is already in Vue, I am not sure why the mounting below is necessary. Also jquery should not be used in new code unless absolutely necessary.

client/galaxy/scripts/components/Workflow/Editor/utilities.js
anuprulez: It was added by mistake. Sorry!

client/galaxy/scripts/components/Workflow/Editor/WorkflowRecommendationsVue.vue
anuprulez: ok, will do it

client/galaxy/scripts/components/Workflow/Editor/workflowRecommendations.js
anuprulez: I needed to communicate `Node.vue` and `WorkflowRecommendationsVue`. I will look out for a way to avoid it and jquery too.

anuprulez: I have removed the helper file, but not sure how to avoid the mounting. The new template should be appended to the body tag.

anuprulez: Also, I have found one small issue with the height of the D3 tree. When the number of items becomes > 15, it gets distorted. Working on it.

client/galaxy/scripts/components/Workflow/Editor/Node.vue
```diff
@@ -70,6 +82,14 @@ export default {
         canClone() {
             return this.type != "subworkflow";
         },
+        isEnabled() {
+            const Galaxy = getGalaxyInstance();
+            const isRecommendationEnabled = Galaxy.config.enable_tool_recommendations;
+            if (isRecommendationEnabled === true || isRecommendationEnabled === "true") {
```
guerler: Is there a case in which this is a string?

guerler:
For some tools without prediction results the predictor triggers the following error:

`Traceback (most recent call last):
  File "lib/galaxy/web/framework/decorators.py", line 295, in decorator
  File "lib/galaxy/webapps/galaxy/api/workflows.py", line 661, in get_tool_predictions
    recommended_tools = self.tool_recommendations.compute_tool_prediction(trans, tool_sequence)
  File "lib/galaxy/tools/recommendations.py", line 199, in compute_tool_prediction
    sample[idx] = int(self.model_data_dictionary[tool_name])`

client/galaxy/scripts/components/Workflow/Editor/workflowRecommendations.js
anuprulez: z203
z202
z201

client/galaxy/scripts/components/Workflow/Editor/Node.vue
anuprulez: yes, "true" was not necessary.

anuprulez:
> For some tools without prediction results the predictor triggers the following error:
> 
> `Traceback (most recent call last): File "lib/galaxy/web/framework/decorators.py", line 295, in decorator rval = func(self, trans, *args, **kwargs) File "lib/galaxy/webapps/galaxy/api/workflows.py", line 661, in get_tool_predictions recommended_tools = self.tool_recommendations.compute_tool_prediction(trans, tool_sequence) File "lib/galaxy/tools/recommendations.py", line 199, in compute_tool_prediction sample[idx] = int(self.model_data_dictionary[tool_name])`

Do you have some specific tools which break this feature?

anuprulez:
@galaxybot test this

anuprulez:
@galaxybot test this

anuprulez:
would someone like to merge it or are there more comments?
Thank you!

anuprulez:
All tests passed!

OlegZharkov:
Um, I think, you could use ```git rebase -i HEAD~6```

anuprulez:
Why?

OlegZharkov:
This will squash commits. Most probably maintainers won't be happy to have six ```merge dev of``` commits in their history.

anuprulez:
Thanks, good idea!

client/galaxy/scripts/components/Workflow/Editor/Node.vue
```diff
@@ -69,9 +91,20 @@ export default {
             }
             return null;
         },
+        popoverId() {
+            return `popover-${this.id}-${this._uid}`;
```
guerler: Can you explain why both ids are needed?

client/galaxy/scripts/components/Workflow/Editor/Node.vue
```diff
@@ -69,9 +91,20 @@ export default {
...
+        },
         canClone() {
             return this.type != "subworkflow";
         },
+        isEnabled() {
+            const Galaxy = getGalaxyInstance();
+            const isRecommendationEnabled = Galaxy.config.enable_tool_recommendations;
+            if (isRecommendationEnabled === true) {
```
guerler: Should just return `isRecommendationEnabled` right? Or is there a special case where this explicit if is required?

client/galaxy/scripts/components/Workflow/Editor/Node.vue
```diff
@@ -80,6 +113,18 @@ export default {
         onClone() {
             this.node.clone();
         },
+        onCreate(toolId, event) {
```
guerler, line 115: This is duplication of `onInsertTool` in `Index.vue`. Please emit an event `this.$emit("onInsertTool", ...)` instead and handle the request in `Index.vue` for now.

guerler:
This looks good to me, please check my last two comments.

client/galaxy/scripts/components/Workflow/Editor/Node.vue
anuprulez: On workflow editor, a toolbox can appear multiple times. If we use the same popover id for multiple (and same) tools, then it is shown only for one box and other tools with same ids do not have popover. Therefore to make ids unique for each toolbox, I am appending `this._uid` id

client/galaxy/scripts/components/Workflow/Editor/Node.vue
anuprulez: They are similar but in `onCreate(toolId, event) ` call, we only have `tool_id` and do not have `tool_name`. We get it by the service call `getModule` and then using the tool's name and id, we create tool's node (toolbox). We can emit an event from Node.vue and then still need to have a separate method in Index.vue

client/galaxy/scripts/components/Workflow/Editor/Node.vue
anuprulez: Fixed it, thanks!

client/galaxy/scripts/components/Workflow/Editor/Node.vue
anuprulez: Additionally, we hide the popover in Node.vue `this.popoverShow = false` after creating the toolbox

guerler: Then we can parse a custom event and handle it in the parent. The `popoverShow` can be set to `false` before emitting the event.

client/galaxy/scripts/components/Workflow/Editor/Node.vue
guerler: The `id` should be unique. I wonder if it has been set properly.

anuprulez: The `id` is unique but its the tool's id. I looked at it further and found out that in method `mountWorkflowNode`  (https://github.com/galaxyproject/galaxy/blob/577443181c9d814d14280b5e3c30259ba2756860/client/galaxy/scripts/mvc/workflow/workflow-manager.js#L140) 
the `id` is actually the `content_id` (tool's id) and not the id of the parent node (which is unique and takes form as `wf-node-step ...`). Once this call is over, only then these ids are assigned in `add_node` method (https://github.com/galaxyproject/galaxy/blob/577443181c9d814d14280b5e3c30259ba2756860/client/galaxy/scripts/mvc/workflow/workflow-manager.js#L114)

Therefore the unique ids of each node is not available in `Node.vue` . To fix that I created one counter in `workflow-manager.js` where all the nodes are created and `Node.vue` is invoked. I use this counter in `Node.vue` to assign unique ids to each popover.

anuprulez:
Thank you @guerler

bgruening:
Cool, nice Easter present. Thanks @anuprulez and @guerler !

guerler:
Thank you @anuprulez for this neat feature.

juleengraham:
Hi @anuprulez! I'm trying to incorporate this tool recommendation feature in another that I'm implementing. However, I'm not getting any recommended tools for any of the tools I pass.

In my code, I've imported your getToolPredictions function from services.js
`import { getToolPredictions } from "components/Workflow/Editor/services";`

and used it in another function as shown below.
`onRecommend(toolId) {
            const requestData = {
                tool_sequence: toolId,
            };
            
            getToolPredictions(requestData)
                .then((responsePred) => {
                    const predictedData = responsePred.predicted_data.children;
                    if (predictedData.length > 0) {
                        this.hasRelatedTools = true;
                    }
                    return predictedData;
                })
                .catch((e) => {
                    console.error(e);
                });
        },`

but `responsePred.predicted_data.children` is an empty array. Am I missing something?

Thanks!

PR #9570, opened 2020-03-27
Index bam files with``-c`` flag if contig size exceeds ``2 ** 29``

### LINKED ISSUE #6956: BAM indexes for very large contigs
from https://biostar.usegalaxy.org/p/29884/

Dear community,

is there a way to modify the default way Galaxy creates the index *.BAI files? We work with a large salamander genome, therefore, quite some coordinates in the BAM files are beyond 2^30. Therefore, the index files have to be created using either biobambam2:bamSort or samtools index -c -m 14. However, I cannot really figure out if there is a simple way to "teach" Galaxy how to do that. We have our own institute's Galaxy instance, therefore, I can change pretty much anything there...

I am sorry if a similar question was posted before and would be thankful for any advice.

Best regards, Sergej

FredericBGA:
This is 98% based on @mvdbeek  work (https://github.com/mvdbeek/galaxy/commit/e386b8dbd002bb5ada6bae06829acb0909eb7575). Thank you!

Should close #6956

The limit seems to be 2\*\*29 -1 (I still have issues with 2**30): 
[biostar](https://www.biostars.org/p/111984/)


FredericBGA:
sorry, I will send another commit, the binary.py is not merged! This is the file based on my 19.05 version.

lib/galaxy/datatypes/binary.py
```diff
@@ -458,20 +458,40 @@ class Bam(BamNative):
     track_type = "ReadTrack"
     data_sources = {"data": "bai", "index": "bigwig"}
 
-    MetadataElement(name="bam_index", desc="BAM Index File", param=metadata.FileParameter, file_ext="bai", readonly=True, no_value=None, visible=False, optional=True)
+    MetadataElement(name="bam_index", desc="BAM Index File", param=metadata.FileParameter, file_ext="bam.bai", readonly=True, no_value=None, visible=False, optional=True)
```
nsoranzo: Is this safe to change?

mvdbeek: I don't think it is (even though it definitely is handy, but we can overwrite this elsewhere). 
```suggestion
    MetadataElement(name="bam_index", desc="BAM Index File", param=metadata.FileParameter, file_ext="bai", readonly=True, no_value=None, visible=False, optional=True)
```

mvdbeek:
@galaxybot test this

mvdbeek:
Thank you @FredericBGA!

dannon:
It looks like this has added `bam_csi_index` download links to all (old, too) bam files, even if they don't have the metadata (seen on usegalaxy.org).  This shouldn't be available, right?

bgruening:
@dannon it seems that we also can not display BAM files anymore.

dannon:
@bgruening Yeah, I'm seeing the same thing.  I'll spend some time on this tomorrow to try to sort out what's going on with the display.

PR #9574, opened 2020-03-29
Add Vue components for Workflow Inputs and Outputs, reduce Workflow loading time

guerler:
This PR is the next step in our effort to vueify the Workflow editor.

Here are some of the changes contained within:
* Removes `workflow-node.js`, `workflow-node-view.js`, `workflow-data-view.js`
* Adds node input and output components and accesses them from the node component
* Removes parsing of jQuery element ($el) for workflow terminals and nodes
* Rewrites the node state initialization/update routines for Vue
* Fixes workflow node labeling
* Fixes workflow output selection behavior
* Fixes displaying new workflow output label in node with proper fallback labels
* Avoids insertion of duplicate labels by highlighting the problem in the submission form
* Removes legacy node width resizing mechanism
* Removes jQuery from node screenreader
* Now uses the screenreader list to create and remove connections
* Collects workflow output related code in a model file without dependencies
* Moves all related files to the `components/Workflow/Editor` directory
* Restores previously failing minor test cases
* Reduces the loading time for workflows (see figure below)

Requires: #9551.


client/galaxy/scripts/mvc/workflow/workflow-view-node.js
```diff
@@ -40,54 +38,53 @@ export class NodeView {
...
         }
-        var terminalView = this.terminalViews[input.name];
-        var terminalViewClass = TerminalViews.InputTerminalView;
+        var terminalClass = Terminals.InputTerminal;
         if (input.input_type == "dataset_collection") {
-            terminalViewClass = TerminalViews.InputCollectionTerminalView;
+            terminalClass = Terminals.InputParameterTerminal;
         } else if (input.input_type == "parameter") {
-            terminalViewClass = TerminalViews.InputParameterTerminalView;
+            terminalClass = Terminals.InputCollectionTerminal;
```
jmchilton: this looks to have been swapped?

guerler: Thanks!

mvdbeek:
Moved this to 20.09 since it's a draft. This is looking very nice!

guerler:
@mvdbeek thanks for catching the post job action handling issue! 🥇

jmchilton:
My first pass at visual inspection of the code looks really good - very nice refactoring and nice modernization. Thanks for fixing the annotation functionality and Selenium tests and thank you preserving the unit and functional testing coverage through big changes - that is hugely appreciated. I think I encoded all my worries about edges cases in there over time.

Out of curiosity - have you investigated working with large workflows? The old code could handle maybe a couple hundreds nodes with a bit of slow down right? I don't have any reason to believe there are regressions - I'm always just nervous about VueJS falling over under everything it manages but it constantly impresses me.

guerler:
Thank you for reviewing this. Those `qunit` tests were extremely helpful, they explain quite a bit about the intended behavior. Thanks a lot for having added them. You raise a good point, regarding the performance. So far I just compared it subjectively by adding 40-50 nodes but I think more systematic performance tests are due and they might hint towards opportunities to improve the performane for very large workflows. I will look into it.

guerler:
I wrote a short script to create workflows using the `cat1` tool and measured the time it takes to load each workflow into the workflow editor. Locally I programmatically measured the delay times, for usegalaxy however I used a stop watch. I repeated each run a few times and averaged the results. This analysis lead to some insights and allowed me significantly improve the results with dd8bc6c.

Picture1

The workflow files are available at: https://gist.github.com/guerler/b419fed4f135b39a67829cf9e87a6c3d

jmchilton:
Amazing - that was way more than I needed and it looks awesome. Super impressive top to bottom!

Jenkins generated screenshots are here - https://jenkins.galaxyproject.org/job/docker-selenium-py3/4298/artifact/4298-test-screenshots/ - they all look good.

Issue #8772, opened 2019-10-08
More canvas for tool search results


nekrut:
Galaxy contains many tools (which is fantastic). This however makes it difficult to find tools and the current tool search is not that helpful (also see #3276). 

# The problem

Here is an example, suppose we are searching for tools that deal with `fastq` format. So one would type `fastq` in the search box on top of the tool panel and will get this:

![fastq-search](https://user-images.githubusercontent.com/4291636/66417993-2e8ceb00-e9cf-11e9-99c0-9dc8775c20cd.png)

As one can see the results are listed according to tool categories and **not** by ranks given by the search engine:

`GET https://usegalaxy.org/api/tools?q=fastq`

```
0: "toolshed.g2.bx.psu.edu/repos/devteam/fastqc/fastqc/0.52"
1: "toolshed.g2.bx.psu.edu/repos/devteam/fastq_combiner/fastq_combiner/1.1.1"
2: "toolshed.g2.bx.psu.edu/repos/devteam/fastq_combiner/fastq_combiner/1.0.1"
3: "toolshed.g2.bx.psu.edu/repos/devteam/fastq_groomer/fastq_groomer/1.1.1"
4: "toolshed.g2.bx.psu.edu/repos/devteam/fastq_groomer/fastq_groomer/1.0.4"
5: "toolshed.g2.bx.psu.edu/repos/lparsons/cutadapt/cutadapt/1.16.3"
6: "toolshed.g2.bx.psu.edu/repos/lparsons/cutadapt/cutadapt/1.16.5"
7: "toolshed.g2.bx.psu.edu/repos/iuc/mothur_make_fastq/mothur_make_fastq/1.39.5.0"
8: "toolshed.g2.bx.psu.edu/repos/iuc/mothur_make_fastq/mothur_make_fastq/1.36.1.0"
9: "toolshed.g2.bx.psu.edu/repos/iuc/mothur_fastq_info/mothur_fastq_info/1.39.5.0"
10: "toolshed.g2.bx.psu.edu/repos/iuc/mothur_fastq_info/mothur_fastq_info/1.36.1.0"
11: "toolshed.g2.bx.psu.edu/repos/jjohnson/mothur_toolsuite/mothur_make_fastq/1.20.0"
12: "toolshed.g2.bx.psu.edu/repos/jjohnson/mothur_toolsuite/mothur_fastq_info/1.25.0"
[...]

Issue #3276, opened 2016-12-06
display ranking of tool search results


joachimwolff:
- The ranking of search results is not as good as it could be. With the default ranking settings, a search for the tool "BAM-to-SAM" is having an exact hit but is only ranked at position three of three. See attached image.

- It looks like k-mers are not computed. A little typo like "PileOMth" instead of "PileOMeth" generates no result instead of one. It would be better for the users if little typos are allowed. 
![screenshot from 2016-12-06 17-16-59](https://cloud.githubusercontent.com/assets/13435136/20933464/662e2e20-bbd8-11e6-9c66-f06ed9d9e1fd.png)

martenson:
@joachimjacob 
You can configure the instance's search here: https://github.com/galaxyproject/galaxy/blob/dev/config/galaxy.ini.sample#L891

The results shown are not ranked top to bottom - it respects their order in the tool panel instead.

Agreed that fuzzy search would be an improvement and worth an issue.

joachimwolff:
@martenson 
I understand why it is done this way right now and i know how to tune the parameters. With not that many tools it is a good way to present the search results. But with an increasing number of tools it is getting confusing where the tool is - and if it was found. For example "cufflinks" is an exact hit and it is at position number nine. 
![screenshot from 2016-12-06 17-31-35](https://cloud.githubusercontent.com/assets/13435136/20934184/8acf811e-bbda-11e6-99d2-4ccbef577e56.png)

martenson:
@joachimjacob I think one of the reasons why the search results are not re-ordered dynamically for each call is to prevent confusion. 

Positions in the toolpanel after search are not ordered, so there is no 'position nine' for cufflinks. Also what you call an 'exact hit' is not that clear in many cases since we search on more attributes than the displayed name.

Maybe you could try experimenting with `tool_search_limit` to provide less results to filter this?

I understand what you are asking for though and it makes sense to me too.

nsoranzo:
Since fuzzy search was implemented in https://github.com/galaxyproject/galaxy/pull/3356 and I don't think we are going to support a ranking of results, I'm going to close this. Feel free to re-open if I've missed something.

joachimwolff:
[...]

PR #9579, opened 2020-03-31
Sort Tool Search Result by ranking

juleengraham:


dannon:
@juleengraham Thanks for opening the PR for this.  I'll go ahead and check it out, but the commit history here looks like something went wrong.  Can you clean that up?  Let me know if you need help.

juleengraham:
Address issue #3276 and part of  #8772

juleengraham:
@galaxybot test this

juleengraham:
@galaxybot test this

martenson:
Hi @juleengraham. This PR does not work for me. For any search string I am getting

```
vue.runtime.esm.js?2b0e:1888 TypeError: Cannot read property 'forEach' of undefined
    at eval (utilities.js?b271:51)
    at Array.map (<anonymous>)
    at filterTools (utilities.js?b271:50)
    at VueComponent.categories (ToolBox.vue?02cd:92)
```

This is because when you run `category.elems.forEach(...`and one of the "categories" is a tool section label, it has no `elems`


Other observations: 

- the filtering should only trigger (and the buttons display) when there is at least one result available.
- it would be nice to have a test for this



As a side note it would be nice to unify the vocabulary used here and consistently use tool panel "section" instead of "category". I believe that is what Galaxy uses in configuration and elsewhere.

Also using active forms of verbs in docs is better than the passive forms. e.g.:

>//Filters out to only display categories that have tools that were in results

`Filter out categories without tools in results`

martenson:
@juleengraham thank you for the quick turnaround. The filtering now works well.

I suggest to make not showing section names the default (and still offer a button to show them). 

But more importantly I believe that in case where we do not show sections we should default to pure ordering by the ranking. This is what was [suggested](https://github.com/galaxyproject/galaxy/issues/8772#issuecomment-542269850) in the issue and I expect that it will make the results more consistent.

Currently when the #2 search hit is in the last section of the tool panel, it will be displayed in the very end of the results (no matter if with sections or without). A better way would be to display it second when no sections are visible, and in the section when they are. I am not sure whether sections should be reordered to respect the rankings or not. Maybe we can leave that part as is and decide later.

juleengraham:
@martenson Sure, I can make sorting without sections the default.
Re the sorting, it is currently set to have tools listed in the order that they rank. Is that not happening on your end? Do you have an example?

Screen Shot 2020-04-28 at 7 42 43 PM
Screen Shot 2020-04-28 at 7 42 52 PM

martenson:
@juleengraham sure, this is from my local instance, with fairly normal toolpanel

Galaxy
Galaxy

martenson:
I found one more thing. When clearing the results with an X there is a following console error:

```
TypeError: Cannot read property 'length' of undefined
    at VueComponent.onQuery (ToolBox.vue?02cd:131)
    at invokeWithErrorHandling (vue.runtime.esm.js?2b0e:1854)
    at VueComponent.invoker (vue.runtime.esm.js?2b0e:2179)
    at invokeWithErrorHandling (vue.runtime.esm.js?2b0e:1854)
    at VueComponent.Vue.$emit (vue.runtime.esm.js?2b0e:3888)
    at VueComponent.checkQuery (ToolSearch.vue?1716:39)
```

juleengraham:
@martenson 
`[Vue warn]: Invalid prop: type check failed for prop "category". Expected Object, got Number with value 0.`
This error might have been fixed with the latest commit but I wasn't able to reproduce it so let me know if it's still appearing and if the button still appears when there are no results.

client/galaxy/scripts/components/Panels/ToolBox.vue
```diff
@@ -128,7 +115,7 @@ export default {
     methods: {
         onQuery(query) {
             this.query = query;
-            this.queryFilter = (query.length >= 3) ? query : null
+            this.queryFilter = (query && query.length) >= 3 ? query : null;
```
martenson: This probably won't work since you're comparing a bool to an int.

martenson:
@juleengraham I can still see it (two different ones actually)
Galaxy

client/galaxy/scripts/components/Panels/ToolBox.vue
```diff
@@ -106,6 +115,7 @@ export default {
     methods: {
         onQuery(query) {
             this.query = query;
+            this.queryFilter = (query && (query.length >= 3)) ? query : null;
```
guerler: Do we need this? The `ToolSearch` component has `minQueryLength` already set to `3`.

client/galaxy/scripts/components/Panels/ToolBox.vue
```diff
@@ -79,15 +88,19 @@ export default {
...
+                return filterTools(this.toolbox, this.results);
+            }
         },
         isUser() {
             const Galaxy = getGalaxyInstance();
             return !!(Galaxy.user && Galaxy.user.id);
         },
         workflows() {
-            return [
+            this.workflow = [
```
dannon: Computed properties should never have side effects (manipulating state)

martenson:
@juleengraham thanks for all the fixes, it looks very nice, great job!

p.s. please ping me next time, just a rebase does not create notification so I'd have to check periodically which does not work for me atm.

PR #9580, opened 2020-04-01
TS performance fixes

mvdbeek:


lib/tool_shed/tools/tool_version_manager.py
```diff
@@ -38,7 +38,7 @@ def get_version_lineage_for_tool(self, repository_id, repository_metadata, guid)
         is called only from the Tool Shed.
         """
         repository = repository_util.get_repository_by_id(self.app, repository_id)
-        repo = hg_util.get_repo_for_repository(self.app, repository=repository)
+        repo = hg_util.repository.hg_repo
```
nsoranzo:
```suggestion
        repo = repository.hg_repo
```

lib/tool_shed/util/hg_util.py
```diff
@@ -121,6 +125,13 @@ def get_readable_ctx_date(ctx):
     return ctx_date
 
 
+def get_repo_for_repository(app, repository=None, repo_path=None):
```
nsoranzo: Maybe just merge this function into `Repository.hg_repo()` ?

lib/tool_shed/webapp/model/__init__.py
```diff
@@ -191,6 +194,12 @@ def __init__(self, id=None, name=None, type=None, remote_repository_url=None, ho
         self.deprecated = deprecated
         self.create_time = create_time
 
+    @property
+    def hg_repo(self):
+        if WEAK_HG_REPO_CACHE.get(self) is None:
```
nsoranzo:
```suggestion
        if not WEAK_HG_REPO_CACHE.get(self):
```

lib/tool_shed/webapp/model/__init__.py
```diff
@@ -191,6 +194,12 @@ def __init__(self, id=None, name=None, type=None, remote_repository_url=None, ho
...
+            WEAK_HG_REPO_CACH[self] = hg_util.get_repo_for_repository(self)
```
nsoranzo:
```suggestion
            WEAK_HG_REPO_CACHE[self] = hg_util.get_repo_for_repository(self)
```

lib/tool_shed/webapp/model/__init__.py
```diff
@@ -191,6 +194,12 @@ def __init__(self, id=None, name=None, type=None, remote_repository_url=None, ho
...
+        return WEAK_HG_REPO_CACH[self]
```
nsoranzo:
```suggestion
        return WEAK_HG_REPO_CACHE[self]
```

nsoranzo:
If any one else is wondering way commit 0c43bb08aefac1ff0b97977b748c9d14965bc5ff was needed, see https://docs.sqlalchemy.org/en/13/orm/constructors.html

lib/tool_shed/webapp/model/__init__.py
```diff
@@ -260,21 +275,21 @@ def get_tool_dependencies(self, app, changeset_revision):
...
-        repo = hg_util.get_repo_for_repository(app, repository=self)
-        tip_rev = repo.changelog.tiprev()
+    def is_new(self):
+        tip_rev = self.hg_repo.changelog.tiprev()
         return tip_rev < 0
 
-    def repo_path(self, app):
-        return app.hgweb_config_manager.get_entry(os.path.join("repos", self.user.username, self.name))
+    def repo_path(self, app=None):
+        # Keep app argument for compaibility with tool_shed_install Repository model
```
nsoranzo:
```suggestion
        # Keep app argument for compatibility with tool_shed_install Repository model
```

lib/tool_shed/webapp/model/__init__.py
```diff
@@ -201,14 +216,14 @@ def admin_role(self):
         raise Exception('Repository %s owned by %s is not associated with a required administrative role.' %
                         (str(self.name), str(self.user.username)))
 
-    def allow_push(self, app):
-        repo = hg_util.get_repo_for_repository(app, repository=self)
-        return repo.ui.config('web', 'allow_push')
+    @property
```
nsoranzo: I wouldn't use a property here since there is a `set_allow_push()` method which is not possible to decorate as a setter.

mvdbeek:
Turns out caching hg_repo is breaking all the TS tests. I guess we'd want to only cache when we know we're not going to modify repo contents, but that needs some more thought.

mvdbeek:
Alright, I think all the relevant tests are passing. The kubernetes failures are unrelated, working on them in a separate PR.

nsoranzo (approved):

PR #9584, opened 2020-04-02
[20.01] Pulsar Kubernetes Enhancements and Fixes

jmchilton:
- Rev Pulsar library and container in order to implement cancelling Kubernetes jobs when the Galaxy job is stopped.
- Implement a k8s_galaxy_instance parameter that namespaces jobs (and increases test robustness as a result).
- Eliminate the need to fix a port to run integration tests - using #9349

All this is important in its own right but also setup to implementing ITs on remote Kubernetes clusters - this work allows reproducibly recovering Job objects needed to poll for podIP address that will be needed in subsequent work.

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -27,10 +27,15 @@ jobs:
     steps:
     - name: Prune unused docker image, volumes and containers
       run: docker system prune -a -f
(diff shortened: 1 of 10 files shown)


jmchilton:
I think other versions of this seem to have made it in at other times.

PR #9587, opened 2020-04-02
[20.05] Assign random password in user manager if none specified

nuwang:
I ran into this error when trying to integrate an OpenID connect provider.

```
112.134.192.179 - - [01/Apr/2020:21:09:52 +1000] "GET /authnz/custos/callback?code=itp%3Aac%3A476282cd-194b-4d96-baf2-0cc465ee0760&state=UfFhjyMS8oPcJ4Z9rRliKRdgHJevDz HTTP/1.1" 500 - "https://galaxy-aust-staging.genome.edu.au/root/login?is_logout_redirect=true" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.149 Safari/537.36"
  File "lib/galaxy/authnz/managers.py", line 260, in callback
    return True, message, backend.callback(state_token, authz_code, trans, login_redirect_url)
  File "lib/galaxy/authnz/custos_authnz.py", line 110, in callback
    user = trans.app.user_manager.create(email=email, username=username)
  File "lib/galaxy/managers/users.py", line 96, in create
    user.set_password_cleartext(password)
  File "lib/galaxy/model/__init__.py", line 393, in set_password_cleartext
    self.password = new_secure_hash(text_type=cleartext)
  File "lib/galaxy/util/hash_util.py", line 72, in new_secure_hash
galaxy.model.database_heartbeat DEBUG 2020-04-01 21:09:53,146 [p:10875,w:1,m:0] [database_heartbeart_main.web.1.thread] main.web.1 is not config watcher
    assert text_type is not None
AssertionError
```

It turns out that the user_manager does not handle empty passwords correctly. The issue appears to be straightforward, but was wondering how it was not breaking authnz completely, and whether this was caused by a recent code change perhaps?

### DIFF
--- a/lib/galaxy/authnz/custos_authnz.py
+++ b/lib/galaxy/authnz/custos_authnz.py
@@ -108,7 +108,6 @@ class CustosAuthnz(IdentityProvider):
                         raise Exception("There already exists a user with email %s.  To associate this external login, you must first be logged in as that existing account." % email)
                 else:
                     user = trans.app.user_manager.create(email=email, username=username)
-                    user.set_random_password()
                     trans.sa_session.add(user)
                     trans.sa_session.flush()
             custos_authnz_token = CustosAuthnzToken(user=user,
--- a/lib/galaxy/security/validate_user_input.py
+++ b/lib/galaxy/security/validate_user_input.py
@@ -26,17 +26,44 @@ FILL_CHAR = '-'
 PASSWORD_MIN_LEN = 6
 
 
-def validate_email(trans, email, user=None, check_dup=True, allow_empty=False):
-    """
-    Validates the email format, also checks whether the domain is blacklisted in the disposable domains configuration.
-    """
+def validate_email_str(email):
+    """Validates a string containing an email address."""
     message = ''
-    if (user and user.email == email) or (email == "" and allow_empty):
-        return message
     if not(VALID_EMAIL_RE.match(email)):
         message = "The format of the email address is not correct."
     elif len(email) > EMAIL_MAX_LEN:
         message = "Email address cannot be more than %d characters in length." % EMAIL_MAX_LEN
+    return message
+
+
+def validate_password_str(password, confirm):
+    if password != confirm:
+        return "Passwords do not match."
+    if not password or len(password) < PASSWORD_MIN_LEN:
+        return "Use a password of at least %d characters." % PASSWORD_MIN_LEN
+    return ""
+
+
+def validate_publicname_str(publicname):
+    """Validates a string containing a public username."""
+    if len(publicname) < PUBLICNAME_MIN_LEN:
+        return "Public name must be at least %d characters in length." % (PUBLICNAME_MIN_LEN)
+    if len(publicname) > PUBLICNAME_MAX_LEN:
+        return "Public name cannot be more than %d characters in length." % (PUBLICNAME_MAX_LEN)
+    if not(VALID_PUBLICNAME_RE.match(publicname)):
+        return "Public name must contain only lower-case letters, numbers, '.', '_' and '-'."
+    return ''
+
+
+def validate_email(trans, email, user=None, check_dup=True, allow_empty=False):
+    """
+    Validates the email format, also checks whether the domain is blacklisted in the disposable domains configuration.
+    """
+    if (user and user.email == email) or (email == "" and allow_empty):
+        return ''
+    message = validate_email_str(email)
+    if message:
+        pass
     elif check_dup and trans.sa_session.query(trans.app.model.User).filter(func.lower(trans.app.model.User.table.c.email) == email.lower()).first():
         message = "User with email '%s' already exists." % email
     #  If the blacklist is not empty filter out the disposable domains.
@@ -56,12 +83,9 @@ def validate_publicname(trans, publicname, user=None):
     """
     if user and user.username == publicname:
         return ''
-    if len(publicname) < PUBLICNAME_MIN_LEN:
-        return "Public name must be at least %d characters in length." % (PUBLICNAME_MIN_LEN)
-    if len(publicname) > PUBLICNAME_MAX_LEN:
-        return "Public name cannot be more than %d characters in length." % (PUBLICNAME_MAX_LEN)
-    if not(VALID_PUBLICNAME_RE.match(publicname)):
-        return "Public name must contain only lower-case letters, numbers, '.', '_' and '-'."
+    message = validate_publicname_str(publicname)
+    if message:
+        return message
     if trans.sa_session.query(trans.app.model.User).filter_by(username=publicname).first():
         return "Public name is taken; please choose another."
     return ''
@@ -82,8 +106,4 @@ def transform_publicname(publicname):
 
 
 def validate_password(trans, password, confirm):
-    if len(password) < PASSWORD_MIN_LEN:
-        return "Use a password of at least %d characters." % PASSWORD_MIN_LEN
-    if password != confirm:
-        return "Passwords do not match."
-    return ""
+    return validate_password_str(password, confirm)
--- a/lib/tool_shed/util/admin_util.py
+++ b/lib/tool_shed/util/admin_util.py
@@ -4,11 +4,11 @@ import time
 from sqlalchemy import false, func
 
 from galaxy import util, web
+from galaxy.security.validate_user_input import validate_password
 from galaxy.util import inflector
 from galaxy.util.hash_util import new_secure_hash
 from galaxy.web.form_builder import CheckboxField
 from tool_shed.util.web_util import escape
-
 
 log = logging.getLogger(__name__)
 compliance_log = logging.getLogger('COMPLIANCE')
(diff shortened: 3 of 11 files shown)


jmchilton (approved):

lib/galaxy/authnz/custos_authnz.py
```diff
@@ -106,7 +106,6 @@ def callback(self, state_token, authz_code, trans, login_redirect_url):
                         raise Exception("There already exists a user with email %s.  To associate this external login, you must first be logged in as that existing account." % email)
                 else:
                     user = trans.app.user_manager.create(email=email, username=username)
-                    user.set_random_password()
```
VJalili, line 111: @nuwang Does it require updating the following method as well?
https://github.com/galaxyproject/galaxy/blob/7855103d2c3846772c20299a06b43e94db8ed201/lib/galaxy/model/__init__.py#L5706-L5719

VJalili (changes requested):

VJalili:
@nuwang  My apologies for the late response. I guess you're using a provider via the [CustosAuthnz](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/authnz/custos_authnz.py); have you tried authentication flow against a [PSA-based](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/authnz/psa_authnz.py) provider as well? (e.g., Google or Elixir)

lib/galaxy/authnz/custos_authnz.py
nuwang: @VJalili Thanks for pointing that out. I think that can remain as is, because if kwargs happens to contain a password, that will be ignored and always overwritten with a random password explicitly. So no change in semantics.

nuwang:
> @nuwang My apologies for the late response. I guess you're using a provider via the [CustosAuthnz](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/authnz/custos_authnz.py); have you tried authentication flow against a [PSA-based](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/authnz/psa_authnz.py) provider as well? (e.g., Google or Elixir)

I only tried with Custos because only Custos exposed configuration through a well-known endpoint.

nuwang:
@VJalili Just a ping. Is this looking ok to go?

test/unit/managers/test_UserManager.py
```diff
@@ -185,6 +185,14 @@ def test_login(self):
...
         self.assertEqual(response["message"], "Success.")
 
+    def test_empty_password(self):
+        self.log("should be able to create a user with no password")
+        user = self.user_manager.create(email='user@nopassword.com', username='nopassword')
+        self.assertIsNotNone(user.id)
+        self.assertIsNotNone(user.password)
+        # should not be able to login with a null or empty password
+        self.assertFalse(check_password("", user.password))
+        self.assertFalse(check_password(None, user.password))
```
afgane, line 195: There's a linting error here, requiring one extra empty line.

dannon:
@nuwang This looks fine, but I am curious about "but was wondering how it was not breaking authnz completely, and whether this was caused by a recent code change perhaps?" as well.  Did you manage to look into this more?

VJalili:
@nuwang can you please test this against a PSA-based backend and make sure it does not break them? Testing against Google should be sufficient. 

Unfortunately I guess with Custos we diverged a bit, so we should make sure the PSA-based backend is equally developed/maintained.

nuwang:
@dannon I believe this refactoring commit caused it: https://github.com/galaxyproject/galaxy/commit/2a18857bba9a728084963f37f57a1801ec40f660#diff-74566421da44253f075336826614cf9c
`trans.app.user_manager.create(email=email, username=username)` did not support empty passwords, while the earlier code created a User object directly, hence the empty password was not an issue.

@VJalili I agree that it would be nice to have PSA tested better, but that needs a separate integration test I think. I don't think this commit is related, and right now, I can't commit the time to test it unfortunately.

dannon:
@nuwang Thanks for the context; it looks like there's a bit more to do here if that's the code that broke it.  Let me test it out.

dannon:
@nuwang I just figured out how we never saw this in any of the interim work -- do you have `use_pbkdf2` set to false?  We migrated that to the recommended default ~7 years ago, and that's the only way this bug would get hit that I can see.

nuwang:
@dannon Yes, that seems to be the case. I hit this on the GA staging server, and presumably, that was carrying old defaults forward. Might be an issue with older cloudman versions even. How does one set about changing this? I imagine just flipping the setting won't do?

However, I don't quite follow why only setting pbkdf2 triggers this. Won't hitting this line do so? https://github.com/galaxyproject/galaxy/pull/9587/files#diff-74566421da44253f075336826614cf9cL108

dannon:
@nuwang It's all going to go through https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/model/__init__.py#L392, which only uses new_secure_hash to hash a password when `use_pbkdf2` is disabled.  Otherwise we use galaxy.security.passwords.hash_password, which already generates a random password without blowing up when you pass it None.
```
>  galaxy.security.passwords.hash_password_PBKDF2(None)
'PBKDF2$sha256$100000$1W6gGBjfjwP50id5$b6B40Ckxb1f4m/r3fceRy9rCrfaBdY08'
```

nuwang:
Right, that makes sense. So this fix should perhaps be made in `new_secure_hash` instead to handle empty clear text?

nsoranzo:
> Right, that makes sense. So this fix should perhaps be made in `new_secure_hash` instead to handle empty clear text?

I was thinking that instead we should also raise an Exception in `galaxy.security.passwords.hash_password()` if `password` is None.
Presently `None` is a valid password:

```python
>>> from galaxy.security.passwords import check_password, hash_password
>>> hashed = hash_password(None)                                                                                                                                                                              
>>> check_password(None, hashed)                                                                                                                                                                              
True
```

dannon:
@nsoranzo Good catch; I shouldn't have said it generates a random password -- it generates a valid hash of the null password!  :)  That should probably throw an exception, agreed.

Galaxy still needs refactoring to make user creation all go through the manager, but that probably shouldn't happen here.  To handle all the cases where someone's trying to set a null password, should we just update `set_password_cleartext` to set a random password when invoked with `cleartext==None`?

nsoranzo:
Seems a good plan! To summarise:
1) add `galaxy.security.passwords.hash_password()`:
```python
    if password is None:
        raise Exception("password cannot be None")
```
2) Move Nuwan's changes from `UserManager.create()` to `User.set_password_cleartext()` and I'd also add a `log.info()` when the password is `None` and we create a random one.

nuwang:
@dannon @nsoranzo I did a bit of refactoring on the password handling. See what you think.
Instead of validating only an empty password in `set_password_cleartext`, I thought it might be better to validate all properties of a password, including length. The decision of whether or not to set a random password will remain with the caller, instead of `set_password_cleartext`. I think this would be more in line with caller expectations.

While doing this, also noticed a bit of repetition between the admin controller, toolshed and galaxy in terms of validation, so tried to factor that out too. One minor difference is that the toolshed public name regex is a bit different from the Galaxy one, and I opted to use the Galaxy public name regex. If this is a problem, I can restore the original regex for the toolshed.

mvdbeek:
This looks good to me, what do you think @nsoranzo @dannon ?

mvdbeek:
I've change the base to 20.05, since this does seem to have the potential to affect user-facing behavior. If something goes wrong it seems easier to isolate this if we roll it out with 20.05. Feel free to backport it further if this is absolutely required in 20.01.

mvdbeek (approved):

lib/galaxy/security/validate_user_input.py
```diff
@@ -26,17 +26,44 @@
...
     if not(VALID_EMAIL_RE.match(email)):
         message = "The format of the email address is not correct."
     elif len(email) > EMAIL_MAX_LEN:
         message = "Email address cannot be more than %d characters in length." % EMAIL_MAX_LEN
+    return message
+
+
+def validate_password_str(password, confirm):
+    if password != confirm:
+        return "Passwords do not match."
```
nsoranzo, line 41: @nuwang Given your (nice!) refactoring, I'd have left this check to `validate_password()`, so this method would only have `password` as parameter. What do you think?

lib/tool_shed/util/admin_util.py
```diff
@@ -685,13 +685,8 @@ def reset_user_password(self, trans, **kwd):
...
                 confirm = kwd.get('confirm', None)
-                if len(password) < 6:
-                    message = "Use a password of at least 6 characters."
-                    status = 'error'
-                    break
-                elif password != confirm:
-                    message = "Passwords do not match."
-                    status = 'error'
+                message = validate_password(trans, password, confirm)
+                if message:
```
nsoranzo, line 689: I think we are now missing:
```python
                    status = 'error'
```
here.

lib/galaxy/security/validate_user_input.py
nuwang: Sure, that sounds better.

lib/tool_shed/util/admin_util.py
nuwang: Thanks for this catch! Just didn't notice that at all. Will fix.

nsoranzo: https://github.com/galaxyproject/galaxy/pull/9851

PR #9591, opened 2020-04-07
Add job_states_summary option and supporting view architecture to HDCA content results 

Nerdinacan:
Update: the previous contents query optimizations were separated into a separate PR (#9680) and have since been merged, so I have rebased this PR and removed all those optimizations.

### Consolidation of the collection job states into the primary contents query

Added a serialization option to the HDCA result set that gives an overview of the job states working inside that collection by doing an additional eager subquery against the implicit jobs tables.

I realize that a couple additional eager joins in the HDCA component query adds load, but the existence of these values in this query means the client can completely avoid a second set of polling that would otherwise be required.

UPDATE: My initial attempt simply joined all the implicit collection job states and listed them in an object keyed by job-id, but I have since been informed that the number of these job states might be in the thousands, and all that's needed here is a count of all the states, so I've reformatted the output to be a list of job status counts instead of individual job state listings.

To that end, the summary now uses a SQL view to do the job state counts. My hope is that this will be a more performant way to process these results.

### Job state change trigger updates corresponding hdca.update_time

This change also relies on the validity of the hdca.update_time field, so an additional trigger was added to the job table to alert the HDCA when one of its jobs has changed state.

### New Module: galaxy.model.view

This PR includes a new mechanism to register and use/join sql views as ORM entities. 

For each view, a Table is defined for the row-schema of the view, and associated with a selectable which can be either an ORM query or a straight SQL string, representing the contents of the view select itself.

The gotcha is any foreign key declarations (for use in declaring ORM relationships) have to be explicitly defined in the "main" model relationship mapping declarations because the Table in the view definition is not registered on the same MetaData instance (because the view row schema is not really a table), and thus standard ForeignKey() Column parameters aren't going to recognize other tables in the db.


jmchilton:
This is both exciting and makes me very anxious.

I'm planning to do a deep dive on big collection performance next week - I'll try to analyze this PR and @ic4f's job state PRs in the context of thousands of jobs and large collections. I need to figure out how to do this testing in a reproducible way that others can run - honestly I've tried many times to document and do that but never quite succeeded to make it turn key.

My anxiety is around the trigger that updates the update time. That kind of update contention on those rows across tens of thousands of jobs seems... problematic. I don't have data and I certainly could easily be wrong so I want to figure that out. My approach obviously requires reading from the database more - but tried to avoid update rows that were more likely to lock, etc.. - seemed like a safer trade off in my mind. Certainly I'm hoping I'm wrong because I understand this could drastically simplify the client and API.

I do appreciate this being isolated out like this - should make testing cleaner, easier - than having to do it within the context of the huge history re-write.

Maybe I'll host a Zoom call or everyone who cares can collaborate on a document or something to try to capture the testing large collections operations more reproducibly and transparently.

Nerdinacan:
@jmchilton Please do include me on that zoom meeting. 

One of my big obstacles is generating an appropriate test set to run these things against. 

I'm certainly not against managing that update_time through the ORM code if that is better than firing a trigger 1000 times. 

Additionally, I think there's another type of trigger which fires not on every row change, but once per DB transaction, but I'm not sure whether that type is supported in SQLITE. I'm also not sure whether that would be helpful to you with respect to the frequent job updates you're talking about. Are all those job updates individual transactions, or are they being sent in as a bulk transaction?

jmchilton:
> I'm certainly not against managing that update_time through the ORM code if that is better than firing a trigger 1000 times.

I think the trigger would be the most appropriate way if you want the collection update to affect the HDCA. It just didn't do this in the past and so there was not row contention during updates for collection jobs. Certainly if we're 100% committed to synchronized state in this fashion - the database is going to do be a better job than sqlalchemy and something we'd hack in there - unless I'm missing something. In other words the trigger is not causing me anxiety, the synchronized writing of that state is. 

Also lets not say 1000 - 1,000 works today, we need a path that lets us grow to 10,000 and 100,000.

> Are all those job updates individual transactions, or are they being sent in as a bulk transaction?

The job updates are all currently happening in other threads and one at a time mostly. For certain classes of jobs that might change in the future - but that would be a corner case I suspect. I think we should plan for 50,000 jobs finishing individually over the course of several minutes in many other threads and processes at scale and how that would perform.

> Are all those job updates individual transactions, or are they being sent in as a bulk transaction?

I'll try to find a way to loop you in, I appreciate you taking my concerns seriously and wanting to be involved. Hopefully we can work together to find a solution that makes everyone happy.

Nerdinacan:
@jmchilton 

Well given that, then there's still the idea of materialized views, which is basically just like a frozen temp table that reflects the results of the sql definition of a view. Essentially you manually tell the view to update itself periodically.

Maybe we could write an HDCA materialized view where the update_time was a calculated value. Then we refresh the materialized view on some appropriate schedule? So then the contents query becomes something like:

```sql
select manycontentfields
from (
   select hda_fields 
   from history_dataset_association 
   where history_id = xyz
   and name = 'foo'
   and deleted = false
   and purged = false
   and update_time > 5minsago

   union all

   select hdca_fields 
   from HDCA_MATERIALIZED_VIEW
   where history_id = xyz
   and name = 'foo'
   and deleted = false
   and purged = false
   and update_time > 5minsago
)
order by hid desc
limit 100 offset 0
```

I had considered replacing our weird serialization nonsense with a 2 views (for HDC and HDCA) anyway, purely for reasons of simplicity.

lib/galaxy/managers/history_contents.py
```diff
@@ -397,9 +405,10 @@ def _subcontainer_id_map(self, id_list):
         component_class = self.subcontainer_class
         query = (self._session().query(component_class)
             .filter(component_class.id.in_(id_list))
-            .options(eagerload('collection'))
-            .options(eagerload('tags'))
-            .options(eagerload('annotations')))
+            .options(joinedload('collection'))
+            .options(joinedload('tags'))
+            .options(joinedload('annotations'))
+            .options(joinedload('job_state_summary')))
```
jmchilton: Will this always happen now? Are we doing a huge join across that table for things we will only need if 'job_state_summary' is requested from the user - which the current client won't do? Or is this deferred in some way?

jmchilton:
I'm still thinking through how to test this but two quick comments:

> To that end, the summary now uses a SQL view to do the job state counts. My hope is that this will be a more performant way to process these results.

This makes a lot sense to me.

> Changing "union" to "union all" gives a performance bump too because "union" also sorts through all those table results looking for duplicates that will not exist. Union all just stacks the results

This seems awesome - really keen insight and great catch. Any chance you're willing to breakout just what enhances the performance of the current query? Not strictly needed but it would cleanup the bigger PR a bit and seems like a great fix.

lib/galaxy/managers/history_contents.py
Nerdinacan: It will happen on the secondary query which retrieves the collections which match the filtered content query.

There are 3 queries in the contents api endpoint, the union I described at the top of the PR on which all the filtering runs, then two additional ones which return the details for HDA and HDCAs which match the filtering criteria which has already run

Thus the last 2 queries are a "select fields from hdca where id in 1,3,4,5" scenario where the IDS are collected from the first unioned query. It is this smaller result set that will have the additional eager join.

SQLAlchemy offers a couple other ways to eager join an additional model, but I felt a join to a view might be one of the more attractive ones. 

Another promising option is a sql-in join which joins the entire joined model using a select .... in ID1, ID2, ID3 scenario then marries the results under the hood.

https://docs.sqlalchemy.org/en/13/orm/loading_relationships.html#sqlalchemy.orm.selectinload

jmchilton:
This question is a bit of design aside but it might help me understand the client side requirements driving these change and understand what possible configurations might work on the backend. Do we need to update update_time on the HDCAs or would a new field in the API response for HDA and HDCAs (maybe called "job_source_update_time") that grabbed an update_time from the job_source work (either a Job or ImplicitCollectJobs)? I sort of made the case here (https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/jobs/__init__.py#L1380) that updating the HDA update_time when the dataset changes is a questionable design change, I would think that would apply even more so in the collection context.

I think since when we copy HDAs and HDCAs we don't copy the jobs or ImplicitCollectionJobs - it would prevent some race conditions at the edges when an HDCA or HDA was being copied while the job source update time was being updated. That increased correctness makes me think it might be a better design choice.

If we did that we could then drop that hack for HDAs also after the new client is the only client.

I guess this would though reflect my clear bias (still without data) toward reducing write contention and improving write performance at the cost of read performance.

Nerdinacan:
The client design goal here was to elminate potentially unnecessary extra polling mechanisms for simple updates. Having a filterable, reliable change timestamp on the HDCA made that possible.

If you see value in introducing a second job-specific timestamp into the model for whatever reason, all I would probably end up doing is generating that update_time via another view as follows:

```
create hda_summary_view as
      select hda.hid, hda.id as hda_id, null as hdca_id, max(hda.update_time, hda.job_update_time_dealie) as update_time
      from history_dataset_association hda

create hdca_summary_view as
    select hdca.hid, null as hda_id, hdca.id as hdca_id, max(hdca.update_time, hdca.job_update_time_dealie) as update_time
    from history_dataset_collection_association hdca
```

Then the main content query would replace the existing HDA tables with the views

And the rest of my plan works just fine.

Sound good?

Nerdinacan:
> I'm still thinking through how to test this but two quick comments:
> 
> > To that end, the summary now uses a SQL view to do the job state counts. My hope is that this will be a more performant way to process these results.
> 
> This makes a lot sense to me.
> 
> > Changing "union" to "union all" gives a performance bump too because "union" also sorts through all those table results looking for duplicates that will not exist. Union all just stacks the results
> 
> This seems awesome - really keen insight and great catch. Any chance you're willing to breakout just what enhances the performance of the current query? Not strictly needed but it would cleanup the bigger PR a bit and seems like a great fix.

Sure, I can do just the changes to that content query and put them in another pr.

Nerdinacan:
> I guess this would though reflect my clear bias (still without data) toward reducing write contention and improving write performance at the cost of read performance.

Another thing we could do is modify the triggers to only fire when the state changes. On an update, a trigger gets a NEW and OLD rowset. We can not bother updating hdca unless the state actually changes, yea?

lib/galaxy/managers/history_contents.py
Nerdinacan: Currently the existing classes do not pass in the user-requested serialization parameters into the content query generation class. Theoretically, we could add that as a new parameter, which would allow the content query to conditionally decide whether or not to eager join stuff like this, but when the new history client is live, it would probably ask for it every time.

If your goal is to make the new serialization option available to the new history client without unduly impacting the existing client's query, I could add a few extra noodles to that spaghetti code.

Nerdinacan: @jmchilton I've added the additional parameter to the content query assembly to make sure that the eager load against the view only occurs if the user explicitly selects it by specifying it in the keys= parameter to the contents endpoint.

The weakness of the serialization class scheme is evident here as the logic to evaluate whether or not to do the eager load is brittle now. If somebody decides to move the job_states_summary field into a pre-defined view then the logic will fail.

I think refactoring the serializers is in order but it can be done later.

jmchilton:
Screen Shot 2020-04-29 at 2 20 10 PM

I am making progress on the performance stuff - I was excited about this cool chart but I went to start summarizing it in this comment and it and realized in many ways it isn't measuring what I wanted... at all. Oh well, I'll keep trying. @ic4f is in the middle of a move this week so the coordination isn't happening anyway. Lets plan to revisit that next week - hopefully I have made some more concrete progress on how to test these things.

Taking notes in : https://docs.google.com/document/d/1cLC9cOlb6pJ_rPFP7-6jfoFVsB0tp9VwcWjwJOGZTKk/edit?usp=sharing but they aren't quite ready for public consumption yet.

Nerdinacan:
I think I'm going to split this PR again to extract the new View model stuff because I really want to use views for other things, and it looks like your primary concerns revolve around trigger performance.

Nerdinacan:
Closing because all these features have moved to other PRs.

Issue #9530, opened 2020-03-17
zip are unzipped


lecorguille:
I'm aware that zip isn't very welcome within Galaxy :)

But we still need them for example for NMR Data. The structure of one sample is rather complex.
<img width="963" alt="Capture d’écran 2020-03-17 à 10 33 08" src="https://user-images.githubusercontent.com/11072022/76842557-bdcde280-683a-11ea-96ec-a0abca616ab6.png">

So that's why I create this nasty but practical 
 no_unzip datatype: https://toolshed.g2.bx.psu.edu/view/lecorguille/no_unzip_datatype/7800ba9a4c1e

Since the 19.01 : 
 - the auto-detection still trigger the unzip and finally keep only one file
 - if we force no_unzip: 
```
Traceback (most recent call last):
  File "/shared/mfs/data/galaxy/server/tools/data_source/upload.py", line 326, in <module>
    __main__()
  File "/shared/mfs/data/galaxy/server/tools/data_source/upload.py", line 319, in __main__
    metadata.append(add_file(dataset, registry, output_path))
  File "/shared/mfs/data/galaxy/server/tools/data_source/upload.py", line 128, in add_file
    convert_spaces_to_tabs=dataset.space_to_tab,
  File "/shared/mfs/data/galaxy/server/lib/galaxy/datatypes/upload_util.py", line 70, in handle_upload
    if check_content and hasattr(datatype, 'sniff') and not datatype.sniff(path):
  File "/shared/mfs/data/galaxy/mutable-data/shed_tools/toolshed.g2.bx.psu.edu/repos/lecorguille/no_unzip_datatype/7800ba9a4c1e/no_unzip_datatype/no_unzip_datatypes.py", line 27, in sniff
    zf = zipfile.ZipFile(filename)
  File "/shared/mfs/data/software/miniconda/envs/_galaxy_/lib/python2.7/zipfile.py", line 770, in __init__
    self._RealGetContents()
  File "/shared/mfs/data/software/miniconda/envs/_galaxy_/lib/python2.7/zipfile.py", line 811, in _RealGetContents
    raise BadZipfile, "File is not a zip file"
zipfile.BadZipfile: File is not a zip file
```

I solved that with a quick and dirty patch: 
https://github.com/galaxyproject/galaxy/blob/dev/tools/data_source/upload.py#L97
```
auto_decompress = dataset.get('auto_decompress', True)
```
In:
```
auto_decompress = False
```

[...]

PR #9595, opened 2020-04-09
Sniffer for multi-member zip archives

lecorguille:
As explained in this issue: https://github.com/galaxyproject/galaxy/issues/9530

The PR won't change anything if they is only one single file within the zip file.
But automatically assign the zip datatype if there are multiple files nested.

I know that it's a very old debate but still the 19.01 I can't use my no_unzip datatype anymore.

🙏

### DIFF
--- a/lib/galaxy/datatypes/binary.py
+++ b/lib/galaxy/datatypes/binary.py
@@ -249,6 +249,16 @@ class CompressedZipArchive(CompressedArchive):
         except Exception:
             return "Compressed zip file (%s)" % (nice_size(dataset.get_size()))
 
+    def sniff(self, filename):
+        with zipfile.ZipFile(filename) as zf:
+            zf_files = zf.infolist()
+            count = 0
+            for f in zf_files:
+                if f.file_size > 0 and not f.filename.startswith('__MACOSX/') and not f.filename.endswith('.DS_Store'):
+                    count += 1
+                if count > 1:
+                    return True
+
 
 class GenericAsn1Binary(Binary):
     """Class for generic ASN.1 binary format"""
--- a/lib/galaxy/config/sample/datatypes_conf.xml.sample
+++ b/lib/galaxy/config/sample/datatypes_conf.xml.sample
@@ -843,6 +843,8 @@
     <sniffer type="galaxy.datatypes.binary:Cpt"/>
     <sniffer type="galaxy.datatypes.binary:Edr"/>
     <sniffer type="galaxy.datatypes.binary:Vel"/>
+    <sniffer type="galaxy.datatypes.binary:Xlsx"/>
+    <sniffer type="galaxy.datatypes.binary:CompressedZipArchive"/>
     <sniffer type="galaxy.datatypes.annotation:Augustus"/>
     <sniffer type="galaxy.datatypes.triples:Rdf"/>
     <sniffer type="galaxy.datatypes.blast:BlastXml"/>


lib/galaxy/datatypes/binary.py
```diff
@@ -198,6 +198,18 @@ def display_peek(self, dataset):
...
+        # If the zip file contains multiple files then return true
+        zf = zipfile.ZipFile(filename)
+        zf_files = zf.infolist()
+        # Filter the directories within the file list
+        zf_files = [x.filename for x in zf_files if x.file_size > 0]
+        # Filter hidden file within the file list
+        hidden_file_patterns = ["__MACOSX",".DS_Store","^\."]
+        regex = re.compile('|'.join(hidden_file_patterns))
+        zf_files = [x for x in zf_files if x not in list(filter(regex.search, zf_files))]
+        return len(zf_files) > 1
```
```suggestion
        UNDESIRED_FILES = {'__MACOSX', '.DS_Store'}
        with zipfile.ZipFile(filename) as zf:
            zf_files = zf.infolist()
            count = 0
            for f in zf_files:
                if x.file_size > 0 and not f.filename in UNDESIRED_FILES and not f.filename.startswith('.'):
                    count += 1
                if count > 1:
                    return True
```
That should have the same effect, but close the ZipFile after we're done and avoids the regex and breaks out of the loop once we know we've got a multi-member archive.

lecorguille: Hum ...
regexp were efficiant for this:
```
dossier1/.DS_Store
__MACOSX/dossier1/._.DS_Store
```

My new proposal:
```python
    def sniff2(filename):
        UNDESIRED_FILES = {'.*__MACOSX.*', '.*\.DS_Store.*'}
        regex = re.compile('|'.join(UNDESIRED_FILES))
        with zipfile.ZipFile(filename) as zf:
            zf_files = zf.infolist()
            count = 0
            for f in zf_files:
                if f.file_size > 0 and not regex.match(f.filename) and not f.filename.startswith('.'):
                    count += 1
                if count > 1:
                    return True
```

mvdbeek: Why do you want to use a regex without a regular expression? Simple set lookup is going to be a lot more efficient.

lecorguille: `        UNDESIRED_FILES = {'.*__MACOSX.*', '.*\.DS_Store.*'}`
It was a misstake, now there are some real regexp 😛

lib/galaxy/datatypes/binary.py
```diff
@@ -200,16 +201,16 @@ def display_peek(self, dataset):
...
-        zf = zipfile.ZipFile(filename)
-        zf_files = zf.infolist()
-        # Filter the directories within the file list
-        zf_files = [x.filename for x in zf_files if x.file_size > 0]
-        # Filter hidden file within the file list
-        hidden_file_patterns = ["__MACOSX",".DS_Store","^\."]
-        regex = re.compile('|'.join(hidden_file_patterns))
-        zf_files = [x for x in zf_files if x not in list(filter(regex.search, zf_files))]
-        return len(zf_files) > 1
+        UNDESIRED_FILES = {'.*__MACOSX.*', '.*\.DS_Store.*'}
```
mvdbeek: Do you have examples of these ? At least the `.DS_Store` files never have any suffix or prefix.

lecorguille: You're right.
```
dossier1/.DS_Store
__MACOSX/dossier1/._.DS_Store
```
I can simplify with: 
```
UNDESIRED_FILES = {'.*__MACOSX.*', '.*\.DS_Store'}
```

lib/galaxy/datatypes/binary.py
```diff
@@ -198,6 +199,18 @@ def display_peek(self, dataset):
             return dataset.peek
         except Exception:
             return "Compressed binary file (%s)" % (nice_size(dataset.get_size()))
+    
+    def sniff(self, filename):
+        UNDESIRED_FILES = {'.*__MACOSX.*', '.*\.DS_Store.*'}
+        regex = re.compile('|'.join(UNDESIRED_FILES))
```
```suggestion
```

lib/galaxy/datatypes/binary.py
```diff
@@ -198,6 +199,18 @@ def display_peek(self, dataset):
...
+        with zipfile.ZipFile(filename) as zf:
+            zf_files = zf.infolist()
+            count = 0
+            for f in zf_files:
+                if f.file_size > 0 and not regex.match(f.filename) and not f.filename.startswith('.'):
```
```suggestion
                if f.file_size > 0 and not f.filename.startswith('__MACOSX/') and not f.filename.endswith('.DS_store'):
```

mvdbeek (commented):
Please drop the regex, it isn't necessary.
Fwiw here is a the contents of a multi-subdirectory zip file:
```
In [4]: z = zipfile.ZipFile('test.zip')

In [5]: z.filelist
Out[5]:
[<ZipInfo filename='test/' filemode='drwxr-xr-x'>,
 <ZipInfo filename='test/.DS_Store' compress_type=deflate filemode='-rw-r--r--' file_size=8196 compress_size=1182>,
 <ZipInfo filename='__MACOSX/test/._.DS_Store' compress_type=deflate filemode='-rw-r--r--' file_size=120 compress_size=53>,
 <ZipInfo filename='test/A/' filemode='drwxr-xr-x'>,
 <ZipInfo filename='test/A/.DS_Store' compress_type=deflate filemode='-rw-r--r--' file_size=6148 compress_size=178>,
 <ZipInfo filename='__MACOSX/test/A/._.DS_Store' compress_type=deflate filemode='-rw-r--r--' file_size=120 compress_size=53>,
 <ZipInfo filename='test/A/Galaxy-Workflow-Parameters_from_file.ga' compress_type=deflate filemode='-rw-r--r--' file_size=13229 compress_size=2235>,
 <ZipInfo filename='__MACOSX/test/A/._Galaxy-Workflow-Parameters_from_file.ga' compress_type=deflate filemode='-rw-r--r--' file_size=530 compress_size=327>]
```
The `__MACOSX` folder is always at the root.

lib/galaxy/datatypes/binary.py
```diff
@@ -6,6 +6,7 @@
 import io
 import logging
 import os
+import re
```
```suggestion
```

mvdbeek:
OK, this failure is legitimate
```
            # Some things can't be uploaded, if attempting to upload these datasets we mention why
            assert 'invalid' in dataset['misc_info'] or 'unsupported' in dataset['misc_info']
            return
        elif dataset['state'] == 'error' and 'empty' in dataset['misc_info']:
            return
        else:
            # state should be OK
            assert dataset['state'] == 'ok'
        # Check that correct datatype has been detected
        file_ext = dataset['file_ext']
        if 'false' in test_data.path:
            # datasets with false in their name are not of a specific datatype
            assert file_ext != PARENT_SNIFFER_MAP.get(expected_file_ext, expected_file_ext)
        else:
>           assert file_ext == PARENT_SNIFFER_MAP.get(expected_file_ext, expected_file_ext)
E           AssertionError: assert 'zip' == 'xlsx'
E             - zip
E             + xlsx
```

xlsx are zip compressed, so we need to move the zip sniffer lower / the xlsx sniffer higher.

lecorguille:
My bad, I was in the wrong class `CompressedArchive` instead of `CompressedZipArchive` (I shouldn't edit directly on github).
Let see what will happening before changing the sniff order.

mvdbeek (approved):
Awesome, thanks a lot @lecorguille!

lecorguille:
> Awesome, thanks a lot @lecorguille!

Are you kiding me. You did 90% of the work!

PR #9597, opened 2020-04-09
[WIP] Working towards testing all config properties. Next draft.

jdavcs:


ic4f:
Standalone workflow for `test_config_defaults` works fine on my fork. Will try to enable all those (50+) commented out tests/properties now.

ic4f:
Closed in favor of #9631. 
(The minimal test driver works in the context of a standalone workflow with other integration tests disabled. However, since the driver does *not* launch the Galaxy app, we don't need an integration test in the first place: unit tests are all that is needed here.)

PR #9600, opened 2020-04-10
Allow Tool Data Tables to be backed by refgenie.

blankenberg:
Can enable by setting `refgenie_config_file` in galaxy.yml to an existing refgenie configuration file.

refgenie: http://refgenie.databio.org/


lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -589,6 +589,13 @@ mapping:
           will use a less efficient monitoring scheme that may work in wider range of scenarios than
           the watchdog default.
 
+      refgenie_config_file:
+        type: str
+        default: ''
```
jdavcs: Does it *need* to be an empty string if not set? If a default is not specified in the schema, `app.config.refgenie_config_file` will have the value `None`, which would be the expected behavior, right?

blankenberg: No, it should be `null` as shown before 0c86a4fc3dfba22f29141f96741386d8afb767ef, but unfortunately this does not pass tests when set.

It is set to an empty string to pass the config syntax / config tests. This is very unfortunate. The inconsistency in behavior between a "missing" `default` value and having `default: null` is really, really bad. Explicit is better than implicit, and they should definitely not act differently. Because of this behavior, while `null` makes more logical sense (and was the original default), explicitly setting the default to `''` is, as a whole, less incorrect.

Better fix is to correct the behavior of the config testing / verification to enable the default to be explicitly setable to `null`, but that is outside the scope of this PR. It could potentially be done with the addition of a `nullable` attribute, that allows values to be set as null or not.

blankenberg: I've convinced myself to pretend that I didn't notice the really, really bad inconsistency (should still be addressed), and am back to `null` in 17ba3d07752bd8cfa3d39b68a73c96626599830d thanks @ic4f !

lib/galaxy/tools/data/__init__.py
```diff
@@ -786,11 +796,114 @@ def to_dict(self):
...
+            for tagged_asset in asset_list:
+                asset, tag = tagged_asset.rsplit(':', 1)
+                if asset != self.rg_asset:
+                    continue
+                digest = rgc.id(genome, asset, tag=tag)
+                uuid = 'refgenie:%s/%s:%s@%s' % (genome, self.rg_asset, tag, digest)
+                display_name = description or '%s/%s' % (genome, tagged_asset)
+
+                def _seek_key(key):
+                    return rgc.get_asset(genome, asset, tag_name=tag, seek_key=key)
```
stolarczyk: If I may chime in:

Is there any particular reason you used `RefGenConf.get_asset` rather than `RefGenConf.seek` ?

`RefGenConf.get_asset` method is deprecated and will be removed in the next `refgenconf` release, see: https://github.com/databio/refgenconf/blob/0d9908decd2f880ca4928bfecad12b920ddf904e/refgenconf/refgenconf.py#L282-L313

We recommend using `RefGenConf.seek` instead. It both matches the CLI and does not assert asset existence _by default_. The latter is imperative when dealing with non-existing asset paths, like `hg38/bowtie2_index`, which returns a ready-to-use input for Bowtie2 software.

nsheff: he probably wrote that before we made that API change... but yes, should switch to `seek` now!

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: @blankenberg Is the inconsistency due to galaxy's config handling? The config system is not without fault, but a simple matter of creating an attribute based on the schema should work correctly. The attribute is created (regardless of whether it's set or not) in this loop: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L222
(So it's `null` by default.) The config integration test should work automatically too. If that didn't happen as expected, please let me know and I'll look into it - that would need to be fixed.

lib/galaxy/tools/data/__init__.py
```diff
@@ -362,6 +366,11 @@ def configure_and_load(self, config_element, tool_data_path, from_shed_config=Fa
                     log.debug('Loading Data Table URL "%s" as filename "%s".', filename, tmp_file.name)
                     filename = tmp_file.name
                     tmp_file.flush()
+                else:
+                    # Pull the filename from a global config
+                    filename = file_element.get('from_config', None) or None
+                    if filename:
+                        filename = self.other_config_dict.get(filename, None)
```
jdavcs, line 373:
```suggestion
                        filename = self.other_config_dict.get(filename)
```
`None` is returned by default

lib/galaxy/tools/data/__init__.py
```diff
@@ -362,6 +366,11 @@ def configure_and_load(self, config_element, tool_data_path, from_shed_config=Fa
                     log.debug('Loading Data Table URL "%s" as filename "%s".', filename, tmp_file.name)
                     filename = tmp_file.name
                     tmp_file.flush()
+                else:
+                    # Pull the filename from a global config
+                    filename = file_element.get('from_config', None) or None
```
jdavcs, line 371:
```suggestion
                    filename = file_element.get('from_config', None)
```
It will be `None` regardless. Also, if `file_element` is a dictionary object, the second argument is not needed (same as comment below)

lib/galaxy/tools/data/__init__.py
blankenberg: explicit is better than implicit

lib/galaxy/tools/data/__init__.py
```diff
@@ -517,7 +526,8 @@ def parse_file_fields(self, reader, errors=None, here="__HERE__"):
         TODO: Allow named access to fields using the column names.
         """
         separator_char = "<TAB>" if self.separator == "\t" else self.separator
-
+        if isinstance(reader, str):
```
jdavcs: I completely agree with **not** doing `open(filename`) on an argument in a method call (your edit on line 518). However, now that it's a filename as it should be, I think the argument name should be changed accordingly. Also, I suggest treating it as a string (i.e., not doing "if the reader is a string, make it into a reader, otherwise assume it's a reader"). If there's other code that calls this method passing a file handle instead, my suggestion is to change that other code, or refactor this file to make the 2 options (filename vs. file handle) more explicit (maybe separate keyword args, etc...)

lib/galaxy/tools/data/__init__.py
blankenberg: the `or None` will make sure that `filename` remains `None` (as it was before it entered this logical scope) even if someone set the `from_config` to an empty string (an explicit empty value for the table config).

lib/galaxy/tools/data/__init__.py
blankenberg: No real reason, just missed swapping this one over during the update, I got the actual 'get_asset' one below it, will fix this one too. Nice catch, thanks.

lib/galaxy/tools/data/__init__.py
jdavcs: Oh, OK. It'll still eval to `False` on the next line though ('' is falsy), but the extra safeguard doesn't hurt.

lib/galaxy/tools/data/__init__.py
blankenberg: Definitely agree that passing a string to a variable named `reader` is not super great, definitely worth a polish. Thanks for these thorough reviews @ic4f !

lib/galaxy/tools/data/__init__.py
```diff
@@ -506,18 +515,18 @@ def parse_column_spec(self, config_element):
...
-    def parse_file_fields(self, reader, errors=None, here="__HERE__"):
+    def parse_file_fields(self, filename, errors=None, here="__HERE__"):
         """
         Parse separated lines from file and return a list of tuples.
 
         TODO: Allow named access to fields using the column names.
         """
         separator_char = "<TAB>" if self.separator == "\t" else self.separator
-
+        reader = open(filename)
```
jdavcs: Thank you for making the change! One last comment (the rest looks fine to me). The `reader` is never closed here. So, maybe `with open(filename) as reader:`?

blankenberg: This like doubles the size of the PR. I have killed this `reader`, long live `fh`. 

Thanks @ic4f

jdavcs (approved):
Thank you for making the changes!

lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
```diff
@@ -152,6 +152,7 @@ pyuwsgi==2.0.18.post0
 pyyaml==5.3
 rdflib-jsonld==0.4.0
 rdflib==4.2.2
+refgenconf==0.7.0
```
nsoranzo, line 157: Can you please add `refgenconf` to  lib/galaxy/dependencies/pipfiles/default/Pipfile

blankenberg: Yes, nice catch. Thanks.

lib/galaxy/dependencies/pipfiles/default/Pipfile
```diff
@@ -83,6 +83,7 @@ cloudbridge = "*"
 social_auth_core = {version = "==3.1.0+gx0", extras = ['openidconnect']}
 cloudauthz = "==0.6.0"
 gxformat2 = "*"
+refgenconf = "==0.7.0"
```
nsoranzo: Is the strict pin necessary?

blankenberg: Depends on the stability of the refgenconf python API. The last update from `0.6.2` to `0.7.0` was not completely backwards compatible, and would have broken somethings if the version auto-bumped.

Safest to pin, but perhaps @nsheff can weigh in?

stolarczyk: I'd suggest making it `refgenconf = "==0.7.0"` or `refgenconf = ">=0.7.0"`, if possible. You're right, you can't go any lower with the version number to use the functionality needed here, but future versions should be backwards compatible with respect to that.

nsheff: Is the only question regarding backwards compatibility related to the asset paths returned for hisat2_index and bwa_index? 

If so, if we update these, then it should be good. We'll try to keep everything backwards-compatible, but we consider these "bug fixes" on the recipes since the seek keys should return what is actually used by the tool. 

will it be possible to change the pin the future? So,  the pin might be prudent at this point, but could be unpinned after we're sure things aren't changing anymore. I *think* we're there after this change, but I may be wrong...

nsheff: also -- probably want to pin to `0.9.1`, since that will include the new seek keys you requested.

blankenberg: @nsheff yes, we can change the pin later. But concern here would be keeping the definitions of the data tables, which a user could have deployed as-is here, with an update to a seek_key would cause a mismatch for that user during upgrade.

`0.9.1` has the changes to hisat2_index and bwa_index?

nsoranzo: To clarify our pinning process, normally we leave most Python dependencies unpinned in this `Pipfile`, and every couple of months generate from it a `requirements.txt` file with the actual pinnings (including the dependencies' dependencies) using `pipenv`.

nsheff: wait -- I was confused, sorry -- `0.9.1` is for `refgenie`, not `refgenconf`. you are not relying on refgenie here, only on refgenconf. So, ignore that comment. `0.7.0` is correct, that's the latest version.

Refgenconf does not change with the update to the seek keys -- those changes happen in refgenie. So, what matters will actually be what version of refgenie you used to create the assets. Right now, you haven't got any way to *build* assets in Galaxy yet, right? So, users are only going to be using `pull` for assets. In that case, fixing these seek keys is not actually a change to any software, it's just a matter of fixing the actual assets on the server.

I've just made a change to the building component ([here](https://github.com/databio/refgenie/pull/172), and we'll be able to deploy these changes to the server shortly.

Anyway -- what I'm trying to say is the mismatches will not be caused by refgenconf changes -- it's all going to be remote anyway. So the pin doesn't help you with that.

blankenberg: For Galaxy driven creation, the building will likely come from a tool(s) added to the toolshed that uses refgenie, which can be added outside of the core Galaxy release cycle. 

I imagine users may want to create their own assets on the command line and load them into their Galaxy instance. 

But it sounds like we can consider the previous asset pull or creation in refgenie to be slightly bugged, with fixes applied or pending. So we should be safe to change our two data table definitions to match what is expected by the underlying executable. Cool, I'll try to find some time to remove, pull, and retest.

mvdbeek (approved):

PR #9602, opened 2020-04-10
gff2bed: show that mixed 6/12 col bed is produced

bernt-matthias:
seems that the converter produces mixed 6 column / 12 column bed format which makes downstream tools struggle.

test data temporarily stored in test-data .. only for demonstration .. I guess this does not work in the test.

Should we just fill with 6 empty columns in the 6col case?

Main question : is it really a bug / feature?


nsoranzo:
I agree that this seems to be problematic, I read on http://genome.ucsc.edu/FAQ/FAQformat#format1 that "The number of fields per line must be consistent throughout any single set of data in an annotation track."

Do you have an example of a downstream tool that is affected?

bernt-matthias:
Jep. rseqc read distribution for instance. It expects 12 column bed (another issue with this is here 
https://github.com/galaxyproject/galaxy/issues/9604). 

I discovered this when using gencode annotations (gtf) and fed it to the rseqc tool. The problem seems to be that there are transcript annotations which generate 12cols and exon annotations which create 6 columns.

My idea was to change the converter such that always 12 columns are generated, potentially the last 6 empty. If you agree I can prepare this.

nsoranzo:
Thanks. I've spent some time looking at other GFF/GTF-to-BED conversion tools, and there seems to be 2 types (see https://www.biostars.org/p/321562/#426456):
- line-by-line (gtf2bed and gff2bed from [bedops](https://bedops.readthedocs.io/)), which simply convert every feature without grouping exons;
- only transcripts (UCSC gtf2ToGenePred, gff3ToGenePred and genePredToBed), which group the exons of a transcript together in one line, discarding the other lines. This is used in https://github.com/galaxyproject/tools-iuc/blob/master/tools/gtfToBed12/gtfToBed12.xml

There is also another Galaxy GFF-to-BED converter in `lib/galaxy/datatypes/converters/gff_to_bed_converter.xml` which is simpler than `tools/filters/gff2bed.xml`, and always produces BED6 files.

A possible solution would be to:
- modify `lib/galaxy/datatypes/converters/gff_to_bed_converter.xml` to set `bed6` as output format
- move ``tools/filters/gff2bed.xml`` to `lib/galaxy/datatypes/converters/` and modify it to become as GFF-to-BED12 converter.

I can help with that if it sounds a good plan.

bernt-matthias:
Agreed. Not sure about "move tools/filters/gff2bed.xml".

nsoranzo:
> Agreed. Not sure about "move tools/filters/gff2bed.xml".

OK, I've added a commit that for now just merges the 2 `gff_to_bed_converter.py` scripts, and left a comment in `gff2bed.xml`.

Planemo tests pass for both converters now. If my changes look good to you, I think the only thing left is to move the new test data to https://github.com/galaxyproject/galaxy-test-data/

bernt-matthias:
I will check asap. Reminds me of https://github.com/galaxyproject/galaxy/pull/8369 which also needs test data over there and is slightly related.

test-data/gencode.v29.annotation.bed
```diff
@@ -1,35 +1,35 @@
-chr1	11868	14409	gene	0	+
+chr1	11868	14409	gene	0	+	11868	14409	0	1	2541	0
```
bernt-matthias: Wondering if the gene name should be in the 4th column?

lib/galaxy/config/sample/datatypes_conf.xml.sample
```diff
@@ -140,7 +140,7 @@
       <!-- <display file="genetrack.xml" /> -->
     </datatype>
     <datatype extension="gff" type="galaxy.datatypes.interval:Gff" display_in_upload="true" description="GFF lines have nine required fields that must be tab-separated." description_url="https://wiki.galaxyproject.org/Learn/Datatypes#GFF">
-      <converter file="gff_to_bed_converter.xml" target_datatype="bed"/>
+      <converter file="gff_to_bed_converter.xml" target_datatype="bed6"/>
```
bernt-matthias, line 159: Would it be an option to add a gff -> bed12 converter?

nsoranzo: That would be the updated `tools/filters/gff2bed.xml` , but would only work if we move/copy/symlink it in `lib/galaxy/datatypes/converters/`
Ideally we could move it, but that would make the tool disappear from the panel (https://github.com/galaxyproject/galaxy/issues/9649)  and may break existing workflows (I haven't actually checked the latter).

lib/galaxy/datatypes/converters/gff_to_bed_converter.py
```diff
@@ -1,33 +1,118 @@
...
+        cur_transcript_chrome = None
+        cur_transcript_id = None
+        cur_transcript_strand = None
+        cur_transcripts_blocks = []  # (start, end) for each block
+    with open(output_name, 'w') as out, open(input_name) as in_fh:
+        for i, line in enumerate(in_fh):
             line = line.rstrip('\r\n')
             if line and not line.startswith('#'):
                 try:
+                    # GFF format: chrom source, name, chromStart, chromEnd, score, strand, attributes
```
nsoranzo:
```suggestion
                    # GFF format: chrom, source, feature, chromStart, chromEnd, score, strand, frame, attributes
```

test-data/gencode.v29.annotation.bed
nsoranzo: > Wondering if the gene name should be in the 4th column?

Yes, that's indeed a bug in the original implementation, see my other comment.

lib/galaxy/datatypes/converters/gff_to_bed_converter.py
```diff
@@ -1,33 +1,118 @@
...
-                    out.write("%s\t%s\t%s\t%s\t0\t%s\n" % (elems[0], start, elems[4], name, strand))
+
+                    if bedcols == 6:
+                        # BED6 format: chrom, chromStart, chromEnd, name, score, strand
+                        out.write("%s\t%s\t%s\t%s\t0\t%s\n" % (elems[0], start, elems[4], name, strand))
+                        continue
+
+                    coords = (int(start), int(elems[4]))
+                    attributes = parse_gff_attributes(elems[8])
+                    t_id = attributes.get("transcript_id", None)
```
nsoranzo:
```suggestion
                    t_id = attributes.get("transcript_id")
```

lib/galaxy/datatypes/converters/gff_to_bed_converter.py
```diff
@@ -1,33 +1,118 @@
...
                     start = str(int(elems[3]) - 1)
                     strand = elems[6]
                     if strand not in ['+', '-']:
                         strand = '+'
-                    # GFF format: chrom source, name, chromStart, chromEnd, score, strand
-                    # Bed format: chrom, chromStart, chromEnd, name, score, strand
-                    #
-                    # Replace any spaces in the name with underscores so UCSC will not complain
+                    # Replace any spaces in the name with underscores so UCSC will not complain.
                     name = elems[2].replace(" ", "_")
```
nsoranzo, line 71: This should be replaced with something like: `name = t_id or attributes.get('gene_id') or attributes.get('ID') or elems[2].replace(" ", "_")` , but we would need to also move lines 79 and 80 before this one and `parse_gff_attributes` would have to be imported also for the gff-to-bed6 converter.
The import is presently problematic because there is no conda package for [galaxy-data](https://pypi.org/project/galaxy-data/) yet, and also the PyPI version is quite old (19.9.0.dev3) and still a pre-release.

mvdbeek:
I've pushed this to 20.09 due to Nicola's comments, but as a bug we can always backport this.

mvdbeek:
Happy to get this in as a bug, I think we're almost there ?

nsoranzo:
Rebased to fix a conflict.

lib/galaxy/datatypes/converters/gff_to_bed_converter.py
bernt-matthias: So it sound to me like adding a recipe for galaxy-data is a good next step: https://github.com/bioconda/bioconda-recipes/pull/26365

PR #9603, opened 2020-04-10
Trust certificate

FredericBGA:
The pending PR for planemo is coming

### DIFF
--- a/lib/galaxy/tool_util/verify/interactor.py
+++ b/lib/galaxy/tool_util/verify/interactor.py
@@ -106,6 +106,15 @@ class GalaxyInteractorApi(object):
         if kwds.get('user_api_key_is_admin_key', False):
             self.master_api_key = self.api_key
         self.keep_outputs_dir = kwds["keep_outputs_dir"]
+        self.verify = True
+        if kwds.get('trust', True):
+            # disable the following warning:
+            # Unverified HTTPS request is being made.
+            # Adding certificate verification is strongly advised.
+            # See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
+            import urllib3
+            urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+            self.verify = False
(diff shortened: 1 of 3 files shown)


FredericBGA:
https://github.com/galaxyproject/planemo/pull/1010

lib/galaxy/tool_util/verify/interactor.py
```diff
@@ -106,6 +106,15 @@ def __init__(self, **kwds):
         if kwds.get('user_api_key_is_admin_key', False):
             self.master_api_key = self.api_key
         self.keep_outputs_dir = kwds["keep_outputs_dir"]
+        self.verify = True
+        if kwds.get('trust', True):
```
jmchilton, line 110: Should the default here be False?

FredericBGA: Yes!
I will do that. I'm going to resend a PR without the commit related to my previous branch...

FredericBGA:
I'm going to resend a PR without the commit related to my previous branch...

lib/galaxy/tool_util/verify/interactor.py
FredericBGA: > 
> 
> Should the default here be False?

No. The default behavior of Request is to check the certificate (so this is the default of the Planemo test command), so `verify=True`. When you decide to trust the certificate with `--trust` then Request will use `verify=False`

FredericBGA: I've documented the option for planemo like this:

> @click.option(
>     "--trust",
>     is_flag=True,
>     help="When set, Requests will trust SSL certificates for HTTPS requests. "
>          "It sets verify=False in requests call",
>     default=False,
> )

nsoranzo: I'd suggest to replace the above 2 lines with:

```python
        self.verify = not kwds.get('trust', False)
        if not self.verify:
```

and remove line 117.

PR #9606, opened 2020-04-10
Fix removal of workflow outputs from workflow nodes in editor 

### LINKED ISSUE #9605: Conditional outputs are not removed in workflow editor on dev
This should be a relatively recent change, it still works on 20.01.

switching_bug

guerler:
Fixes #9605.


mvdbeek (approved):
Awesome, thanks for the quick fix. Also I love the work you put into the editor re-write, this feels super snappy!

guerler:
Thank you for catching this.

PR #9609, opened 2020-04-13
Enabled various object store types to store data objects by uuid.

kxk302:
... not by (unique) id generated by Galaxy.

An observation: Ran integration test the uploads file to Galaxy while object store is set to irods. Integration test uploads 10 files. Previously, when object id was being used, those 10 objects would have ids 0 to 9 and all would go to /tempZone/home/rods/000 folder. Now, that uuid is used, the 10 files are put in 10 different folders (see below):

 C- /tempZone/home/rods/0  
  C- /tempZone/home/rods/1  
  C- /tempZone/home/rods/2  
  C- /tempZone/home/rods/5  
  C- /tempZone/home/rods/7  
  C- /tempZone/home/rods/9  
  C- /tempZone/home/rods/c  
  C- /tempZone/home/rods/d  
  C- /tempZone/home/rods/f


kxk302:
@jmchilton I have changed Distributed Object Store to be able to support any backends (not just Disk Object Stores). The change is in lib/galaxy/objectstore/__init__.py. I added an integration test (test/integration/test_datatype_upload_dos.py) that uploads files to 1) a Distributed Object Store with 2 disk object stores as backends, and 2) a Distributed Object Store with 1 disk object store and 1 irods object store as backends

@mvdbeek I've updated all object stores so they could store objects by uuid. The changes are in lib/galaxy/objectstore/ folder. The files changed are azure_blob.py, cloud.py, irods.py, and pithos.py. I have also modified lib/tool_shed/webapp/model/__init__.py to fix the issue with loading tests in VS code (moved mercurial imports to inside the function that uses them, per our discussion). Finally, I renamed a class name in test/integration/test_datatype_upload_irods.py, so its name is more specific.,

jmchilton (approved):
Looks amazing to me!

test/integration/test_datatype_upload_dos.py
```diff
@@ -0,0 +1,161 @@
...
+            <cache path="${temp_directory}/object_store_cache" size="1000"/>
+            <extra_dir type="job_work" path="${temp_directory}/job_working_directory_irods"/>
+            <extra_dir type="temp" path="${temp_directory}/tmp_irods"/>
+        </backend>
+    </backends>
+</object_store>
+""")
+
+
+def start_irods(container_name):
```
mvdbeek: This and `stop_irods` looks like it's a duplicate, can you double-check if anything else here can be removed ?

test/integration/test_datatype_upload_dos.py
```diff
@@ -0,0 +1,161 @@
...
+                )
+            )
+        config["object_store_config_file"] = config_path
+
+
+class UploadTestDosIrodsAndDiskTestCase(UploadTestDatatypeDataTestCase):
+
+    @classmethod
+    def handle_galaxy_config_kwds(cls, config):
+        temp_directory = cls._test_driver.mkdtemp()
```
mvdbeek: I noticed that in test_datatype_upload_irods.py there's a class definition like this:
```
class UploadTestDatatypeDataTestCase(UploadTestDatatypeDataTestCase):
```
Can you change this to `IrodsUploadTestDatatypeDataTestCase` and then inherit from this class for UploadTestDosIrodsAndDiskTestCase ? I think with that you can drop `setUpClass ` and `tearDownClass` from `UploadTestDosIrodsAndDiskTestCase`

I also think you can skip the duplicated logic in handle_galaxy_config_kwds and do something like 

```
    def handle_galaxy_config_kwds(cls, config):
        UploadTestDatatypeDataTestCase.handle_galaxy_config_kwds(config)
        temp_directory = cls.object_stores_parent
        with open(config_path, "w") as f:
            f.write(
                IRODS__OBJECT_STORE_CONFIG.safe_substitute(
                    {
                        "temp_directory": temp_directory,
                        "host": IRODS_OBJECT_STORE_HOST,
                        "port": IRODS_OBJECT_STORE_PORT,
                        "timeout": IRODS_OBJECT_STORE_TIMEOUT,
                        "username": IRODS_OBJECT_STORE_USERNAME,
                        "password": IRODS_OBJECT_STORE_PASSWORD,
                        "resource": IRODS_OBJECT_STORE_RESOURCE,
                        "zone": IRODS_OBJECT_STORE_ZONE
                    }
                )
            )
        config["object_store_config_file"] = config_path
```

lib/galaxy/objectstore/__init__.py
```diff
@@ -712,54 +708,24 @@ def parse_xml(clazz, config_xml, legacy=False):
...
-                    'weight': weight,
-                    'max_percent_full': maxpctfull,
-                    'files_dir': path,
-                    'extra_dirs': extra_dirs,
-                    'type': elem_type,
-                }
-                if store_by is not None:
-                    backend_dict['store_by'] = store_by
-                backends.append(backend_dict)
+        for b in config_xml.find('backends'):
```
mvdbeek: Is that correct also if legacy is true ? Should this be `config_xml.findall(backends)` or `backends_root.find('backends') ` maybe ? In your PR it looks like `backends_root ` is now unused, in which case you can remove this.

mvdbeek (commented):
Nice, well done! Just some minor cleanup needed and please see my comment on the `parse_xml` method.

test/integration/test_datatype_upload_dos.py
kxk302: Sure, will import from _irods test and delete these.

kxk302: Actually, if I have UploadTestDosIrodsAndDiskTestCase extend IrodsUploadTestDatatypeDataTestCase (per your next comment), I don't need to import the methods at all.

test/integration/test_datatype_upload_dos.py
kxk302: I extended the class as you suggested (good idea). However, I don't think I an re-use the code in handle_galaxy_config_kwds as they are operating on 2 different configs (OBJECT_STORE_CONFIG in irods test file, and IRODS_OBJECT_STORE_CONFIG is dos test).

lib/galaxy/objectstore/__init__.py
kxk302: backends_root is used I setting a field of config_dict. I'll double check the legacy true/false logic again to see if its correct.

kxk302: Ok, I looked into this a bit more. I had removed the from_xml(), but upon further inspection, I think that may not be right. In from_xml(), if config_xml is None, we read off a config file specified by distributed_object_store_file field of config (and we set the legacy fag to True). For backward compatibility, I added the method back (Unless you guys are sure its a stale code and can be removed with no side effect). 

I also reverted the for loop's first line (the one that iterates over backends). 

This should make my changes fully backward compatible.

test/integration/test_datatype_upload_dos.py
mvdbeek: You could make it work. The example I posted above should take care of using `IRODS__OBJECT_STORE_CONFIG`, it was just missing `config_path = os.path.join(temp_directory, "object_store_conf.xml")`. It's not super-important, but I think it would help maintenance of these tests.

test/integration/test_datatype_upload_dos.py
```diff
@@ -0,0 +1,131 @@
+
```
mvdbeek:
```suggestion
```

lib/galaxy/objectstore/__init__.py
mvdbeek: Awesome, yes, looks good.

test/integration/test_datatype_upload_dos.py
mvdbeek: A nicer approach could be creating a `object_store_config` class attribute that hold the expanded object store config, so the base class can just write this to disk and set the appropriate config value.

mvdbeek:
(if you're ready please don't forget to mark this as ready for review)

bgruening:
@kxk302 that looks awesome.

@natefoo and I just discussed yesterday that we would love to activate UUIDs on ORG and EU with 20.05 :)

test/integration/test_datatype_upload_dos.py
kxk302: Ok. I added two class variables to the super class to store the config and config path. Will push the code now for review.

mvdbeek (approved):
Thanks, that's an important feature. If you're ready can you mark the PR as ready for review ?

PR #9612, opened 2020-04-14
add trust option

FredericBGA:
This allows the "planemo test" command working with a self signed certificate

Option is documented in Planemo like this:
> @click.option(
>     "--trust",
>     is_flag=True,
>     help="When set, Requests will trust SSL certificates for HTTPS requests. "
>          "It sets verify=False in requests call",
>     default=False,
> )

`verify=True` is the default for Requests.

### DIFF
--- a/lib/galaxy/tool_util/verify/interactor.py
+++ b/lib/galaxy/tool_util/verify/interactor.py
@@ -102,21 +102,19 @@ class GalaxyInteractorApi(object):
     def __init__(self, **kwds):
         self.api_url = "%s/api" % kwds["galaxy_url"].rstrip("/")
         self.master_api_key = kwds["master_api_key"]
-        self.api_key = self.__get_user_key(kwds.get("api_key"), kwds.get("master_api_key"), test_user=kwds.get("test_user"))
-        if kwds.get('user_api_key_is_admin_key', False):
-            self.master_api_key = self.api_key
-        self.keep_outputs_dir = kwds["keep_outputs_dir"]
-        self.verify = True
-        if kwds.get('trust', True):
+        self.verify = not kwds.get('noverify', False)
+        if not self.verify:
             # disable the following warning:
             # Unverified HTTPS request is being made.
             # Adding certificate verification is strongly advised.
             # See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
             import urllib3
             urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
-            self.verify = False
+        self.api_key = self.__get_user_key(kwds.get("api_key"), kwds.get("master_api_key"), test_user=kwds.get("test_user"))
+        if kwds.get('user_api_key_is_admin_key', False):
+            self.master_api_key = self.api_key
+        self.keep_outputs_dir = kwds["keep_outputs_dir"]
         self._target_galaxy_version = None
-
         self.uploads = {}
 
     @property
--- a/lib/galaxy/tool_util/verify/interactor.py
+++ b/lib/galaxy/tool_util/verify/interactor.py
@@ -102,7 +102,7 @@ class GalaxyInteractorApi(object):
     def __init__(self, **kwds):
         self.api_url = "%s/api" % kwds["galaxy_url"].rstrip("/")
         self.master_api_key = kwds["master_api_key"]
-        self.verify = not kwds.get('noverify', False)
+        self.verify = kwds.get('verify', False)
         if not self.verify:
             # disable the following warning:
             # Unverified HTTPS request is being made.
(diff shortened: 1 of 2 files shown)


lib/galaxy/tool_util/verify/interactor.py
```diff
@@ -106,6 +106,15 @@ def __init__(self, **kwds):
         if kwds.get('user_api_key_is_admin_key', False):
             self.master_api_key = self.api_key
         self.keep_outputs_dir = kwds["keep_outputs_dir"]
+        self.verify = True
+        if kwds.get('trust', True):
```
nsoranzo:
```suggestion
        self.verify = not kwds.get('trust', False)
        if not self.verify:
```

lib/galaxy/tool_util/verify/interactor.py
```diff
@@ -106,6 +106,15 @@ def __init__(self, **kwds):
...
+            # disable the following warning:
+            # Unverified HTTPS request is being made.
+            # Adding certificate verification is strongly advised.
+            # See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
+            import urllib3
+            urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+            self.verify = False
```
nsoranzo:
```suggestion
```

lib/galaxy_test/driver/api.py
```diff
@@ -126,6 +126,7 @@ class ApiTestInteractor(BaseInteractor):
     def __init__(self, test_case):
         admin = getattr(test_case, "require_admin_user", False)
         test_user = TEST_USER if not admin else ADMIN_TEST_USER
+        self.verify = True
```
nsoranzo: I don't think that's needed.
```suggestion
```

lib/galaxy/tool_util/verify/interactor.py
nsoranzo: Also this could be cleaner if the option was named `verify` instead of `trust` (with default `True` in that case).

FredericBGA: Thank you for the review.
I didn't want to change the command line for all those don't need the option.
`--verify` is virtually in the command line as Requests `verify=True` is the default.
So maybe to be clearer, it should be named `--noverify`? So only the people like me (with self signed certificate) will have to add an option to the command line.
What do you think?

lib/galaxy_test/driver/api.py
FredericBGA: I was trying to deal with the test that failed. 
 `AttributeError: 'GalaxyInteractorApi' object has no attribute 'verify'`
(https://github.com/galaxyproject/galaxy/pull/9612/checks?check_run_id=585645835#step:11:597)
Could you help me?

lib/galaxy/tool_util/verify/interactor.py
nsoranzo: You don't need to change the Planemo option, just the Galaxy side, but `--noverify` actually looks nice to me.

lib/galaxy_test/driver/api.py
nsoranzo: You need to move the new lines 109-117 before the call to `__get_user_key()` method on line 105 (which in turn call `_get()`, which looks for `self.verify`)

lib/galaxy/tool_util/verify/interactor.py
```diff
@@ -102,12 +102,19 @@ class GalaxyInteractorApi(object):
...
         self.api_url = "%s/api" % kwds["galaxy_url"].rstrip("/")
         self.master_api_key = kwds["master_api_key"]
+        self.verify = not kwds.get('noverify', False)
+        if not self.verify:
+            # disable the following warning:
+            # Unverified HTTPS request is being made.
+            # Adding certificate verification is strongly advised.
+            # See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
+            import urllib3
+            urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
```
nsoranzo, line 112: A thought: does this disable the warnings for the whole Galaxy app? Maybe we can live with the warnings?

FredericBGA: I don't know. I think it will disable only the given warning but I have no idea about the scope of this.
Tha said, the warning is quite ugly on command line:

> /softs/bioinfo/galaxy-dev/.venv/lib/python2.7/site-packages/urllib3/connectionpool.py:847: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
>   InsecureRequestWarning)

FredericBGA:
Hello

Thanks to @nsoranzo it works if you add the self signed certificate in the root trusted bundle and then export the same variable before launch the interface:

    export REQUESTS_CA_BUNDLE=/etc/pki/tls/certs/ca-bundle.trust.crt
    ./run.sh

galaxyproject/planemo#981 (comment)

nsoranzo:
Cool, thank you for all the work on this issue any way!

PR #9614, opened 2020-04-14
[20.01] Fix ITs to work in remote Kubernetes

jmchilton:
Notes on running the test case in the test file. It works... sometimes. Some sort of timing issue in there but I think the over arching feature "works" more or less.

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -635,8 +635,8 @@ class GalaxyAppConfiguration(BaseAppConfiguration):
 
         # InteractiveTools propagator mapping file
         self.interactivetools_map = self.resolve_path(kwargs.get("interactivetools_map", os.path.join(self.data_dir, "interactivetools_map.sqlite")))
-        self.interactivetool_prefix = kwargs.get("interactivetools_prefix", "interactivetool")
-        self.interactivetool_proxy_host = kwargs.get("interactivetools_proxy_host", None)
+        self.interactivetools_prefix = kwargs.get("interactivetools_prefix", "interactivetool")
+        self.interactivetools_proxy_host = kwargs.get("interactivetools_proxy_host", None)
 
         self.containers_conf = parse_containers_config(self.containers_config_file)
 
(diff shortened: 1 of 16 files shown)


lib/galaxy/config/__init__.py
```diff
@@ -636,7 +636,7 @@ def _process_config(self, kwargs):
         # InteractiveTools propagator mapping file
         self.interactivetools_map = self.resolve_path(kwargs.get("interactivetools_map", os.path.join(self.data_dir, "interactivetools_map.sqlite")))
         self.interactivetool_prefix = kwargs.get("interactivetools_prefix", "interactivetool")
-        self.interactivetool_proxy_host = kwargs.get("interactivetool_proxy_host", None)
+        self.interactivetool_proxy_host = kwargs.get("interactivetools_proxy_host", None)
```
nsoranzo: Should the attribute be renamed too? (here and wherever is used)

jmchilton: This matches the pattern of prefix - but it looks like prefix is just used in one place so I can change both of them.

nsoranzo: Yep, I think that would help @ic4f's config refactorings too.

jdavcs: Yes, it would! If we can keep those identifies in sync with the schema, those 3 lines can be removed (separate PR - after integration testing for those and other config properties works correctly) Thank you!

mvdbeek:
Can you cherry-pick https://github.com/galaxyproject/galaxy/commit/3100009fc65159368105d21dddf654e6c3bd7154 into this branch ? It might reveal what's going on (or maybe target dev directly ?)

mvdbeek:
Ah, you already did that ... weird that I can't find the node state in the logs

test/integration/test_kubernetes_staging.py
```diff
@@ -87,64 +96,95 @@
...
-@integration_util.skip_unless_fixed_port()
-class KubernetesStagingContainerIntegrationTestCase(BaseJobEnvironmentIntegrationTestCase, MulledJobTestCases):
+@integration_util.skip_unless_amqp()
+class BaseKubernetesStagingTest(BaseJobEnvironmentIntegrationTestCase, MulledJobTestCases):
+    # Test leverages $UWSGI_PORT in job code, need to set this up.
+    require_uwsgi = True
 
     def setUp(self):
-        super(KubernetesStagingContainerIntegrationTestCase, self).setUp()
+        super(BaseKubernetesStagingTest, self).setUp()
```
mvdbeek, line 123: Can you overwrite the dataset populator with `KubernetesDatasetPopulator` here ? That should at least print out the node state.

jmchilton:
Still broken - I guess one theory I had glancing over the new output is that the IT test isn't the problem but it creates a job that is preventing other jobs from running. Let me make sure that IT job is being killed.

mvdbeek (approved):

PR #9618, opened 2020-04-15
Add tests for admin interface jobs, dependencies, and containers sections

davebx:


jmchilton:
There is a ``skip_if_jenkins`` decorator in ``lib/galaxy_test/driver/integration_util.py`` that can be used to skip tests when they are running on jenkins.

jmchilton:
xref https://github.com/galaxyproject/galaxy/pull/9652 - that PR should be something you can build on to get these tests running with a usable Docker environment.

davebx:
@jmchilton You're hecking right, it was trivial to extract the container part of #9652 into #9728, thanks for that.

PR #9620, opened 2020-04-16
Fix ELIXIR AAI

lecorguille:
As suggested by @hexylena on Gitter https://gitter.im/galaxyproject/admins?at=5e95af245b98016d6a35f178:

___

Gildas Le Corguillé @lecorguille avr. 14 14:39
>Bonjour,
>Have there been any changes in the Elixir AAI I missed?
```
  File "/shared/mfs/data/galaxy/server/lib/galaxy/authnz/managers.py", line 156, in _get_authnz_backend
    return True, "", identity_provider_class(unified_provider_name, self.oidc_config, self.oidc_backends_config[unified_provider_name])
  File "/shared/mfs/data/galaxy/server/lib/galaxy/authnz/psa_authnz.py", line 117, in __init__
    del self.config["SOCIAL_AUTH_SECONDARY_AUTH_PROVIDER"]
KeyError: 'SOCIAL_AUTH_SECONDARY_AUTH_PROVIDER'
The user is getting : Login failed for an unknown reason.
```
Helena Rasche @hexylena avr. 14 14:40
>there have been changes, I got an email some days ago about them changing which attributes are sent but that shouldn't affect that.
> that looks like an easy fix at least, just wrap in an if > 'SOCIAL_AUTH_SECONDARY_AUTH_PROVIDER' in self.config:

Gildas Le Corguillé @lecorguille avr. 14 17:04
```python
        # Secondary AuthZ with Google identities is currently supported
        if provider != "google":
            if 'SOCIAL_AUTH_SECONDARY_AUTH_PROVIDER' in self.config:
                del self.config["SOCIAL_AUTH_SECONDARY_AUTH_PROVIDER"]
                del self.config["SOCIAL_AUTH_SECONDARY_AUTH_ENDPOINT"]
```
___

I don't really know if it's armless or not but it works.

### DIFF
--- a/lib/galaxy/authnz/psa_authnz.py
+++ b/lib/galaxy/authnz/psa_authnz.py
@@ -114,8 +114,9 @@ class PSAAuthnz(IdentityProvider):
 
         # Secondary AuthZ with Google identities is currently supported
         if provider != "google":
-            del self.config["SOCIAL_AUTH_SECONDARY_AUTH_PROVIDER"]
-            del self.config["SOCIAL_AUTH_SECONDARY_AUTH_ENDPOINT"]
+            if 'SOCIAL_AUTH_SECONDARY_AUTH_PROVIDER' in self.config:
+                del self.config["SOCIAL_AUTH_SECONDARY_AUTH_PROVIDER"]
+                del self.config["SOCIAL_AUTH_SECONDARY_AUTH_ENDPOINT"]
 
     def _setup_idp(self, oidc_backend_config):
         self.config[setting_name('AUTH_EXTRA_ARGUMENTS')] = {'access_type': 'offline'}
--- a/lib/galaxy/authnz/psa_authnz.py
+++ b/lib/galaxy/authnz/psa_authnz.py
@@ -116,6 +116,7 @@ class PSAAuthnz(IdentityProvider):
         if provider != "google":
             if 'SOCIAL_AUTH_SECONDARY_AUTH_PROVIDER' in self.config:
                 del self.config["SOCIAL_AUTH_SECONDARY_AUTH_PROVIDER"]
+            if 'SOCIAL_AUTH_SECONDARY_AUTH_ENDPOINT' in self.config:
                 del self.config["SOCIAL_AUTH_SECONDARY_AUTH_ENDPOINT"]
 
     def _setup_idp(self, oidc_backend_config):


bgruening:
@lecorguille usegalaxy.eu is still working for me. Can you try as well please? Is that something that happens with newer Galaxy versions? We are still running 19.09.

lib/galaxy/authnz/psa_authnz.py
```diff
@@ -114,8 +114,9 @@ def __init__(self, provider, oidc_config, oidc_backend_config):
 
         # Secondary AuthZ with Google identities is currently supported
         if provider != "google":
-            del self.config["SOCIAL_AUTH_SECONDARY_AUTH_PROVIDER"]
-            del self.config["SOCIAL_AUTH_SECONDARY_AUTH_ENDPOINT"]
+            if 'SOCIAL_AUTH_SECONDARY_AUTH_PROVIDER' in self.config:
+                del self.config["SOCIAL_AUTH_SECONDARY_AUTH_PROVIDER"]
+                del self.config["SOCIAL_AUTH_SECONDARY_AUTH_ENDPOINT"]
```
hexylena, line 120:
```suggestion
            if 'SOCIAL_AUTH_SECONDARY_AUTH_ENDPOINT' in self.config:
                del self.config["SOCIAL_AUTH_SECONDARY_AUTH_ENDPOINT"]
```

hexylena:
@bgruening the deletion of those keys was added in 20.01: https://github.com/galaxyproject/galaxy/pull/8926/files#diff-f7cddc18d60b7d8e80e8c1336ef1dfb8R115

So EU doesn't hit this bug yet, but they will.

hexylena:
From the email I've now received 4 times for some reason:


> You are contacted, because you are the administrator of one of the services connected to ELIXIR AAI.
>
> In the period 13.4.2020 - 17.4.2020 (we will specify the exact date soon), there will be a change in the ‘eduPersonEntitlement’ (SAML2 protocol) and ‘eduperson_entitlement’ (OpenID Connect) attribute values. 
>
> The changes are as follows:
> 
> OpenID Connect protocol
>
> Currently, OpenID Connect services receive a complete list of all user groups.
>
> Newly, it will be necessary to specify groups the service is interested in. If no group will be specified, the claim value will be empty.
>
> You can request list of assigned groups by sending a list of groups together with information about the client to email: aai-contact@elixir-europe.org.
>
>
> For OIDC protocol connected services only:
>
> !At the same time the scopes 'groupNames', 'eduPersonEntitlement' and 'edu_person_entitlements' will be removed. If you use some of these attributes, you will need to reconfigure your service to use the 'eduperson_entitlement' scope, which contains the same values. Your client has the 'eduperson_entitlement' scope already enabled. The value of the scopes described in points 1 and 2 in the previous paragraph!

So, shouldn't affect us, but maybe they've made other changes too. I'm not reading those mails as closely as I used to.

lecorguille:
I contacted them and got this reply:


>We don't have experiences with Galaxy internals and this change is
inside the Galaxy lib, so it is not related to Social Python ELIXIR AAI
plugin. Sorry for not being helpful here.

hexylena (approved):

hexylena:
If those attributes need to be removed, let's remove them like this, safely. I guess AAI doesn't send those attributes, so, maybe this was always and bug and we've just caught it now that the first european server upgrades, but the fix looks good to me.

jmchilton:
Thanks all.

PR #9621, opened 2020-04-16
Port and Vuefy AWS estimate from EU

OlegZharkov:
Refactoring ```AWS estimate``` functionality from EU server, initially implemented [here]( https://github.com/usegalaxy-eu/galaxy/blob/release_19.09_europe/templates/show_params.mako#L121). 

Of course configurable and requires metrics to be enabled. Example:
image

Should I change text or add something?

### DIFF
--- a/client/galaxy/scripts/components/JobMetrics/ec2.json
+++ b/client/galaxy/scripts/components/JobMetrics/ec2.json
@@ -1,0 +1,306 @@
+[
+  {
+    "name": "t2.nano",
+    "mem": 0.5,
+    "price": 0.0067,
+    "priceunit": "Hrs",
+    "vcpus": 1,
+    "cpu": "Intel Xeon Family"
+  },
+  {
+    "name": "t3.nano",
+    "mem": 0.5,
+    "price": 0.006,
+    "priceunit": "Hrs",
+    "vcpus": 2,
+    "cpu": "Intel Skylake E5 2686 v5 (2.5 GHz)"
+  },
+  {
+    "name": "t2.micro",
+    "mem": 1,
+    "price": 0.0134,
+    "priceunit": "Hrs",
+    "vcpus": 1,
+    "cpu": "Intel Xeon Family"
+  },
+  {
+    "name": "t3.micro",
+    "mem": 1,
+    "price": 0.012,
+    "priceunit": "Hrs",
+    "vcpus": 2,
+    "cpu": "Intel Skylake E5 2686 v5 (2.5 GHz)"
+  },
+  {
+    "name": "t2.small",
+    "mem": 2,
+    "price": 0.0268,
+    "priceunit": "Hrs",
+    "vcpus": 1,
+    "cpu": "Intel Xeon Family"
+  },
+  {
+    "name": "t3.small",
+    "mem": 2,
+    "price": 0.024,
+    "priceunit": "Hrs",
+    "vcpus": 2,
+    "cpu": "Intel Skylake E5 2686 v5 (2.5 GHz)"
+  },
+  {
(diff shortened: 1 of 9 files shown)


doc/source/admin/galaxy_options.rst
```diff
@@ -1376,6 +1376,16 @@
 :Type: bool
 
 
+~~~~~~~~~~~~~~~~
+``aws_estimate``
+~~~~~~~~~~~~~~~~
+
+:Description:
+    Enable AWS estimate.
```
bgruening: a little bit more description would be nice. Something along these lines.
This flag enables an AWS cost estimate for every job based on their runtime matrices. For this, the CPU and memory usage and the runtime is takes and mapped against AWS pricing table (path ...). Please note, that those numbers
are only estimates.

OlegZharkov: @bgruening Thanks, it is a good idea, I will do.

jmchilton (approved):

client/galaxy/scripts/components/JobMetrics/ec2.json
```diff
@@ -0,0 +1,306 @@
+[
```
dannon, line 1: Is there a way to fetch this programmatically?  My concern is that with instances and their pricing changing this is going to be a maintenance burden.

bgruening: https://github.com/hexylena/awstimator

Possible, we can update this once in a while. EU will take care.

dannon: Sounds good.  Ideally the client just fetches this from an external source on demand, but I'm fine with including it for now -- especially since it's an 'estimate' anyway, and an optional feature.

dannon (approved):

OlegZharkov:
Thanks!

client/galaxy/scripts/components/JobMetrics/ec2.json
hexylena: The upstream datasource is gigantic (~1.2Gb), so probably not possible to fetch on demand.

We can update that repo with jenkins or so. Yes, emphasize estimate ;)

OlegZharkov: We can include processing of that huge file in CI and then fetch it from CD as an artifact. (not sure, if CD is possible on circle/github_actions, but I can do it very fast on gitlab CI)

hexylena: Sure, definitely an option!

My feeling is Return on (time) Investment of automating the update of this file is very low, I'm not sure anyone will complain even if the numbers are 1-2 years out of date, since they are not making any decisions based on this estimate, and few people see it.

dannon: Yeah, to clarify I definitely don't mean fetching gigs of data on demand -- Just the 300 line .json.

OlegZharkov: @dannon Yeah. Probably, next time we need to update it, we can include https://github.com/hexylena/awstimator in github CI (maybe compliment it a little with sorting and deduplication). Then we fetch output using this https://developer.github.com/v3/actions/artifacts/ (hopefully we don't end up with CORS)

P.S. ofc it's low priority. As @hexylena mentioned, might be not worth the time investment.

dannon: @OlegZharkov Sounds good to me.  Galaxy has a built-in CORS handling mechanism, wouldn't be a problem hopefully.

PR #9622, opened 2020-04-16
Fix tensorflow bug

anuprulez:
Adds the same TensorFlow graph for loading the model and predicting

ping @guerler 

thanks!


guerler:
Thanks!

PR #9626, opened 2020-04-17
Change session timeout redirect URL to login entrypoint.

dannon:
More refactoring to do for other login redirect cases that I'll follow up on in dev, but this particular one hit us a while back.


afgane (approved):

dannon:
Test failures are unrelated.

Issue #6100, opened 2018-05-09
rule based uploader does not scroll well/at all


martenson:
Note that columns go all the way to H but I cannot scroll horizontally. However there seems to be a dedicated space below the table for a scrollbar?

I can scroll vertically but there is also no scrollbar visible. No visible space for it this time though.

I am on Chrome MacOS with Apple touchpad.

ping @jmchilton 

![screenshot 2018-05-09 10 34 30](https://user-images.githubusercontent.com/1814954/39820778-97694986-5374-11e8-951a-7d2362aba1a3.png)

jmchilton:
It scrolls for a chosen few, that is the will of the rule builder. You have not been chosen - this saddens me but that is the will of the builder.

martenson:
@jmchilton reported by @nekrut 🙃

martenson:
Moving this to 18.09 since it is a non-blocking corner case.

jmchilton:
I just retried this and works for me under Chrome and Firefox on my Macbook Pro. I'm sure there is a problem - one too many divs maybe with the scrolls on the wrong div or something - I just cannot exercise the problem. Maybe we can work through it Tuesday? I'm pretty stumped.

martenson:
@jmchilton sure, we can explore whenever you are around.

jmchilton:
Spent some time on this today but didn't make a lot of progress:

- @mvdbeek verified this was a problem, but it sometimes works with the trackpad but not his mouse. It will scroll to the right but not the left with his trackpad.
- I've started to try to replicate the bug inside a fiddle - if anyone wants to play with it http://jsfiddle.net/Lauwkuus/5/ or http://jsfiddle.net/Lauwkuus/3/.. It was broken on my laptop for a while, but then I restarted the computer and all my problems went away.
- I've also been playing around with handsontable 3.0.0 in a local branch - but since I cannot consistently replicate the problem yet I can't tell if it is improving anything. The PR that is interesting over there is https://github.com/handsontable/handsontable/pull/4994.

mvdbeek:
I actually have a retina display and see some flickering in their example
fiddle that you linked on gitter, even though the scrolling worked fine. No
clue if that's handsontable 3 or not. I'll have a look at the fiddle
tomorrow.

mvdbeek:
@assuntad23 saved the day (year ? decade ?), this should be fixed by https://github.com/galaxyproject/galaxy/pull/9627

PR #9627, opened 2020-04-17
Changed CSS overflow property from scroll to hidden to make table scrollable

assuntad23:
Why you made the changes (e.g. references to GitHub issues being fixed).
Potential solution to ticket #6100

    A description of the implementation of the changes.
Modified the CSS overflow property from scroll to hidden to make the table scrollable.

    How to test the changes, if you haven't included specific tests already.
Open the Rule Builder and supply a large amount of data (in terms of column-width/count and row count) to make sure you are able to scroll to see all of it.

### DIFF
--- a/client/galaxy/scripts/components/RuleCollectionBuilder.vue
+++ b/client/galaxy/scripts/components/RuleCollectionBuilder.vue
@@ -446,7 +446,7 @@
                         :data="hotData.data"
                         :col-headers="colHeadersDisplay"
                         :read-only="true"
-                        :stretch-h="all"
+                        stretch-h="all"
                     >
                     </hot-table>
                 </div>
--- a/client/galaxy/scripts/components/RuleCollectionBuilder.vue
+++ b/client/galaxy/scripts/components/RuleCollectionBuilder.vue
@@ -446,7 +446,7 @@
                         :data="hotData.data"
                         :col-headers="colHeadersDisplay"
                         :read-only="true"
-                        stretch-h="all"
+                        :stretch-h="all"
                     >
                     </hot-table>
                 </div>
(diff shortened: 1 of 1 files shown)


assuntad23:
@mvdbeek Marius, can you see if this resolves the issue for you?  As I understand it, it only happens for some environments.

mvdbeek:
Awesome, you rock!
I'd create a gif if licecap didn't stop working for some inexplicable reason.
Both scroll directions work perfectly, this closes https://github.com/galaxyproject/galaxy/issues/6100!

Screenshot 2020-04-20 at 15 29 02

mvdbeek:
@galaxybot test this

mvdbeek (approved):

client/galaxy/scripts/components/RuleCollectionBuilder.vue
```diff
@@ -446,7 +446,7 @@
                         :data="hotData.data"
                         :col-headers="colHeadersDisplay"
                         :read-only="true"
-                        stretch-h="all"
+                        :stretch-h="all"
```
dannon: I'm not sure I understand this one; was this not applying correctly before?

assuntad23: @dannon I noticed it as an inconsistency and thought it could be part of the problem, so I decided to fix it while I was there. 

I'm not sure that it made any difference with regards to this, but consistency is important.

dannon: Consistency is important, but this isn't the same thing as binding, for example `:col-headers="colHeadersdisplay"`.  The colon is shorthand for setting up v-bind, which is used to dynamically bind attributes to an expression.  This is just a static attribute we want to set.

dannon:
The integration test failure is unrelated.  Thanks, @assuntad23!

dannon (approved):

PR #9631, opened 2020-04-20
Add unit tests for all schema-defined config properties

jdavcs:
This enables testing for all schema-defined config properties. These are unit tests; they test the values in the `AppConfiguration` object without launching the Galaxy application. Thus, we test the configuration option values that are supplied to Galaxy by default (currently: 312 tested + 6 pending) 

*Note on integration testing.* The `test_config_defaults` integration test does not test all of these properties (57 properties are excluded, mostly due to the test driver overwriting their default values). Still, that test is needed, as it guards against any of these properties (minus excluded) being modified by the Galaxy app on startup.

Replaces #9597; ref #8493 (ongoing).

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -348,10 +348,15 @@ class GalaxyAppConfiguration(BaseAppConfiguration, CommonConfigurationMixin):
 
     def __init__(self, **kwargs):
         super(GalaxyAppConfiguration, self).__init__(**kwargs)
+        self._override_tempdir(kwargs)
         self._process_config(kwargs)
 
     def _load_schema(self):
         return AppSchema(GALAXY_CONFIG_SCHEMA_PATH, GALAXY_APP_NAME)
+
+    def _override_tempdir(self, kwargs):
+        if string_as_bool(kwargs.get("override_tempdir", "True")):
+            tempfile.tempdir = self.new_file_path
 
     def _process_config(self, kwargs):
         # Resolve paths of other config files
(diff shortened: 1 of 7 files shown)


ic4f:
no idea what's causing all those circleci failures..

mvdbeek:
Can you change https://github.com/galaxyproject/galaxy/blob/dev/test/unit/unittest_utils/galaxy_mock.py#L184 to `raise AttributeError(name)` ? It may or may not fix the unit tests, but that should be the correct way to implement `__getattr__`

ic4f:
Found it. This line - https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L383 - sets `tempfile.tempdir`, which is a global variable that defines the value of the default `dir` argument to the functions in Python's `tempfile` module. Which breaks half of the tests, of course. Fixed, broke a few others. Minor stuff + cleanup left.

mvdbeek:
> Found it. This line - https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L383 - sets `tempfile.tempdir`, which is a global variable that defines the value of the default `dir` argument to the functions in Python's `tempfile` module.

Excellent detective work, and ... should Galaxy should be doing that ? It is nice and central, but also a little surprising.

mvdbeek (approved):

ic4f:
@mvdbeek oh yes, I agree, it's surprising to say the least. *Not surprisingly*, Python docs explicitly discourage this. I'll get to it soon (as part of refactoring the rest of config - which is so much easier/safer with these tests in place). Thank you for reviewing and merging!

PR #9634, opened 2020-04-20
Skip visualization build, if no changes were registered

OlegZharkov:
After some discussion with @dannon, we realized that some mechanism for skipping visualizations build is required. This PR ensures that we build them just once (when ```{viz_name}_plugin_build_hash.txt``` is absent) and skip further viz builds if no changes were done. The check is done similarly to [current client build](https://github.com/galaxyproject/galaxy/blob/dev/scripts/common_startup.sh#L222]).

 Advantages:
- we don't build them every time we change something in client.
- we build only visualizations, which were actually changed (not all of them, whole the time)
- adding more visualizations makes every build time-consuming. This check makes it just faster

Skipping unchanged builds:

image

@dannon Should I also skip ```cleanPlugins``` and ```stagePlugins``` in [gulpfile](https://github.com/galaxyproject/galaxy/blob/dev/client/gulpfile.js#L95)?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -178,5 +178,4 @@ packages/*/*.egg-info
 config/plugins/visualizations/**/static/script.js
 # TODO: Really need to follow up on this and make it standard.
 config/plugins/visualizations/**/static/main.css
-
-config/plugins/visualizations/**/static/*_plugin_build_hash.txt
+config/plugins/visualizations/**/static/plugin_build_hash.txt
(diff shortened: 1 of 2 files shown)


dannon:
@OlegZharkov Thanks for working on this!  You're on the right track, but this needs to check for a plugin-specific "plugin_build_hash.txt" or the like, that gets saved to each plugin directory and compared against -- instead of the full client build hash.

The clean/link steps are currently comparatively very fast, so not a huge priority, but I think it'd be optimal to ultimately refactor each plugin to a series of build steps (clean, build, link -- all behind the git rebuild logic).  This, instead of the three clean, build, link steps that each process *all* plugins.

dannon:
Thanks @OlegZharkov!

OlegZharkov:
@dannon thanks for wrapping it up ;)

PR #9639, opened 2020-04-21
[20.01] CSS change to fix scrolling on Rule Collection Builder

assuntad23:


client/galaxy/scripts/components/RuleCollectionBuilder.vue
```diff
@@ -1987,19 +1987,19 @@ export default {
 <style>
 .table-column {
     width: 100%;
-    /* overflow: scroll; */
+    overflow: hidden;
```
dannon, line 1990: Now that I look at this a second time -- are we sure that this fairly generic classname doesn't leak and affect anything else?  This block isn't scoped.

PR #9643, opened 2020-04-21
clean up code to use helper function instead

lawrence14701-zz:
Hi, 

This is my first commit to the galaxy project. I would like to start committing to the client code. This is a simple change to reuse the helper function so the code looks cleaner. I tested it in the localhost and the page behaves the same with this change.


bgruening:
@lawrence14701 cool! Welcome to Galaxy. Let us know if you have any questions and join us at https://gitter.im/galaxyproject/dev if you like.

OlegZharkov (approved):
Thanks. That's a good observation.

PR #9644, opened 2020-04-22
Remove DO_NOT_TEST from config unit tests

jdavcs:
100% of schema-derived config properties are covered by unit tests. 
Detailed comments in commit messages.

### DIFF
--- a/doc/source/admin/galaxy_options.rst
+++ b/doc/source/admin/galaxy_options.rst
@@ -3845,11 +3845,11 @@
     from workflow_resource_params_file). If this this is a function
     reference it will be passed various inputs (workflow model object
     and user) and it should produce a list of input IDs. If it is a
-    path it is expected to an XML or YAML file describing how to map
-    group names to parameter descriptions (additional types of
+    path it is expected to be an XML or YAML file describing how to
+    map group names to parameter descriptions (additional types of
     mappings via these files could be implemented but haven't yet -
     for instance using workflow tags to do the mapping).
-:Default: ``config/workflow_resource_mapper_conf.yml``
+:Default: ``None``
 :Type: str
 
 
--- a/doc/source/admin/galaxy_options.rst
+++ b/doc/source/admin/galaxy_options.rst
@@ -1381,9 +1381,7 @@
 ~~~~~~~~~~~~~~~~
 
 :Description:
-    This flag enables an AWS cost estimate for every job based on their runtime matrices.
-    CPU, RAM and runtime usage is mapped against AWS pricing table.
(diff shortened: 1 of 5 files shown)


doc/source/admin/galaxy_options.rst
```diff
@@ -1381,9 +1381,7 @@
 ~~~~~~~~~~~~~~~~
 
 :Description:
-    This flag enables an AWS cost estimate for every job based on their runtime matrices.
-    CPU, RAM and runtime usage is mapped against AWS pricing table.
-    Please note, that those numbers are only estimates.
+    Enable AWS estimate.
```
nsoranzo: I think the proper fix here would be to copy the long option description to `lib/galaxy/webapps/galaxy/config_schema.yml`

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -2841,14 +2839,13 @@ mapping:
...
         required: false
         desc: |
           This parameter describes how to map users and workflows to a set of workflow
           resource parameter to present (typically input IDs from workflow_resource_params_file).
           If this this is a function reference it will be passed various inputs (workflow model
           object and user) and it should produce a list of input IDs. If it is a path
-          it is expected to an XML or YAML file describing how to map group names to parameter
+          it is expected to be an XML or YAML file describing how to map group names to parameter
           descriptions (additional types of mappings via these files could be implemented but
           haven't yet - for instance using workflow tags to do the mapping).
```
nsoranzo, line 2852: I think this help should mention the that `config/workflow_resource_mapper_conf.yml.sample` is an example of such a config file.

doc/source/admin/galaxy_options.rst
jdavcs: Done. It appears that text was only in `galaxy_options.rst`, so it got overwritten on rebuild. I assumed that was an updated/shortened description. Thanks!

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Done. Added as "sample default path" to the description.

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -2841,17 +2841,18 @@ mapping:
...
 
+          Sample default path 'config/workflow_resource_mapper_conf.yml'
```
```suggestion
          Sample default path 'config/workflow_resource_mapper_conf.yml.sample'
```

ic4f:
Thank you!

PR #9647, opened 2020-04-23
Tools new canvas

juleengraham:
Add more space for tools during tool search


client/galaxy/scripts/components/Panels/ToolBox.vue
```diff
@@ -11,6 +11,11 @@
         </div>
         <div class="unified-panel-controls">
             <tool-search :query="query" placeholder="search tools" @onQuery="onQuery" @onResults="onResults" />
+
+            <div class="float-none" v-if="results">
+                <button class="btn btn-secondary btn-sm" v-if="!show" @click="onToggle">Show Categories</button>
+                <button class="btn btn-secondary btn-sm" v-if="show" @click="onToggle">Hide Categories</button>
```
guerler: We should use a computed variable for the button label, then a single `<button/>` component would be enough.

client/galaxy/scripts/components/Panels/ToolBox.vue
```diff
@@ -44,11 +57,14 @@ import ToolSection from "./Common/ToolSection";
...
 import UploadButton from "./Buttons/UploadButton";
 import FavoritesButton from "./Buttons/FavoritesButton";
-import { filterToolSections } from "./utilities";
+import { filterToolsinCats, filterTools, resizePanel } from "./utilities";
 import { getGalaxyInstance } from "app";
 import { getAppRoot } from "onload";
 import _l from "utils/localization";
 
+const EXPANDED_WIDTH = 400;
+const DEFAULT_WIDTH = 288;
```
guerler: Can we move these into `css` and use percentages if possible?

juleengraham: Not using the hardcoded numbers. Changed to calculated values based on the left and centre panels

Issue #9646, opened 2020-04-22
Boolean "Simple input used for workflow logic" do not attach in workflows (Galaxy 20.01)


Tomnl:
When using the boolean option for "Simple input used for workflow logic" a connection cannot be made.

See below for example using Galaxy tool msPurity.dimsPredictPuritySingle

![image](https://user-images.githubusercontent.com/5541632/80025478-a2f52a80-84d8-11ea-9013-df9c9fc96dc9.png)

innovate-invent:
can you post a link to the tool xml?

Tomnl:
Sure. 

https://github.com/computational-metabolomics/mspurity-galaxy/blob/master/tools/msPurity/dimsPredictPuritySingle.xml#L93

I have tested for other tools though and get the same problem
e.g. XCMS fillpeaks (https://github.com/workflow4metabolomics/tools-metabolomics/tree/master/tools/xcms_fillpeaks) - (https://github.com/workflow4metabolomics/tools-metabolomics/blob/master/tools/xcms_macro/macros_xcms.xml#L93)

Marius on gitter mentioned that there is a problem with the editor for Galaxy @mvdbeek

mvdbeek:
Thanks for the report @Tomnl, this should be fixed by https://github.com/galaxyproject/galaxy/pull/9648

PR #9648, opened 2020-04-23
[20.01] Fix workfow parameter connections and default values

mvdbeek:
in workflow editor and run form.

The main problem was that
`default_source = dict(name="default", label="Default Value",
type=parameter_type)`, where `parameter_type` is the top-level
parameter that is always `text`. This has changed with the
Conditional introduced in 20.01.

In addition BooleanToolParameters use `check` instead of `value`.
I also added ColorToolParameter.

One thing for dev is that I think we should pull the default
values out of the optional conditional so we can set defaults
for required parameters.

Fixes https://github.com/galaxyproject/galaxy/issues/9646

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -27,10 +27,15 @@ jobs:
     steps:
     - name: Prune unused docker image, volumes and containers
       run: docker system prune -a -f
+    - name: Clean dotnet folder for space
+      if: matrix.subset == 'kubernetes'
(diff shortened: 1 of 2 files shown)


jmchilton (approved):

jmchilton:
Pushed a merge with release_20.01 to try to trigger the Github tests and that seemed to work. Previous green dot didn't include them (https://github.com/galaxyproject/galaxy/actions/runs/85811642).

mvdbeek:
Awesome, thanks!

PR #9651, opened 2020-04-23
weasyprint in requirements

OlegZharkov:
```weasyprint``` needed to be in ```requirements.txt``` in order to produce PDFs from workflow reports.
I guess, there was a reason not to put it there, so this PR is mostly just to start discussion about its relevance. Otherwise export button, just doesn't work.

[code reference](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/managers/markdown_util.py#L437)

There is also a possibility to export PDF using just frontend https://github.com/MrRio/jsPDF and I can try it out if needed. But since there is already implementation for PDF export (which works), it would be just easier to use it.

### DIFF
--- a/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
+++ b/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
@@ -194,3 +194,4 @@ webob==1.8.6
 whoosh==2.7.4
 wrapt==1.12.1
 zipp==1.2.0
+weasyprint==51


mvdbeek:
> I guess, there was a reason not to put it there

indeed. https://github.com/galaxyproject/galaxy/pull/8893#issuecomment-567169967

dannon:
Ahh, yeah, I remember now.  Sorry I didn't recall that this was the problem @OlegZharkov.

OlegZharkov:
@dannon Yeah, that might be a good idea. Also, what's your opinion on https://github.com/MrRio/jsPDF? They claim to export PDF just from a div

nsoranzo:
I haven't looked closely, but I think we were looking to convert from Markdown, so maybe something like https://www.npmjs.com/package/markdown-pdf , which has few extra dependencies?

dannon:
js-based pdf generation might not be ideal since we want to support the report API endpoint for programmatic access.  I guess we could run nodejs via python serverside as well or something, if we really find the perfect js library, but looking at candidate python options first seems like the right next move here.

jmchilton:
Had a discussion in the committers meeting - I think we want to piggy back off the dependency resolution stuff and bring in pandoc from Conda. I hope that is okay @OlegZharkov

OlegZharkov:
> pandoc

sure, I guess I will close it for now

simonbray:
Maybe you could consider removing nonfunctional features (i.e. remove the API endpoint and the link to the PDF).

dannon:
It's functional if you install the dependencies.  In better news, weasyprint has new maintainers, is active again, and removing the cairo dependency is the next thing on the roadmap.  So we can *probably* just give it a little time and it'll be a lot easier to have as a regular dependency that 'just works'.

PR #9652, opened 2020-04-24
Framework for Browser Testing of Custom Configured Galaxy Instances

jmchilton:
During CI, all existing Selenium tests currently run against a single, static Galaxy instance with a fixed configuration (same as framework, API, etc..). The existing integration test framework runs each test in a customized Galaxy instance that is configured for the test - allowing testing of various non-standard options.

This PR introduces a new testing framework that allow Selenium testing of Galaxy instances configured using the integration test infrastructure - allowing browser testing of customized Galaxy instances. A test case testing the upload tool's FTP popover is included to exercise this infrastructure.

This PR includes infrastructure for running these tests using Github Actions. Github Actions are very convenient for integration tests because it is easy to configure custom services via Docker containers. So this PR should enable testing of functionality that requires Docker such as https://github.com/galaxyproject/galaxy/pull/9618 for instance. Our existing Docker tests run this way.

I don't think Jenkins should be replaced for Github Actions in general for Selenium testing since Jenkins publishes screenshots and detailed debug information with multiple useful files for each failure, but this does set us down a path toward being able to do that if we find an alternate way to publish or store that data.

### DIFF
--- a/.github/workflows/integration_selenium.yaml
+++ b/.github/workflows/integration_selenium.yaml
@@ -1,0 +1,45 @@
+name: Integration Selenium
+on: [push, pull_request]
+env:
+  GALAXY_TEST_DBURI: 'postgres://postgres:postgres@localhost:5432/galaxy?client_encoding=utf8'
+  GALAXY_TEST_SELENIUM_REMOTE: '1'
+  GALAXY_TEST_SELENIUM_REMOTE_PORT: "4444"
+  GALAXY_SKIP_CLIENT_BUILD: '0'
+jobs:
+  test:
+    name: Test
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        python-version: [3.7]
+    services:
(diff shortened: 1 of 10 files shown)


bernt-matthias:
Out of interest: What kind of functionality is tested with the selenium tests? Would it be possible to test real user setups or specific job submission setups (SGE/torque/...)

bgruening:
@bernt-matthias selenium is exercising the UI. So you can simulate mouse-clicks. SGE/torque etc is tested inside the Galaxy Docker project.

bgruening (approved):
Cool - Thanks John!

mvdbeek:
Thanks @jmchilton !

@bernt-matthias We already test job submission against slurm and torque that we spin up during integration testing (no selenium needed, as Bjoern said). If you can't feasibly dockerize this setup you could create a github workflow and set the environment up there.

PR #9653, opened 2020-04-24
[20.01] Fix vcf_bgzip mimetype, allow cors, tbi from metadata element

mvdbeek:
The mimetype would be text/plain otherwise, which is not compatible with byte-range requests. `allow_cors` doesn't quite seem to work in my tests, but we'll need this for jbrowse.

Also skips an additional conversion by using the tabix index, which again is helpful when loading datasets from a shared history.


mvdbeek:
Thanks @dannon !

> doesn't quite seem to work in my
> test

To clarify, this is likely jbrowse's fault, it doesn't send a `Origin` header, works fine otherwise

Issue #7998, opened 2019-05-17
Allow tool tests to assert what .loc data they require


jmchilton:
usegalaxy.org bug hunting project to improve automated tool testing against main.

- A PR that modifies the testing script wrapper  #7356 
- A PR that adds syntax to the tool XML #7905
- The tool data table https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/api/tool_data.py
- A tool with example tool tests data (https://github.com/galaxyproject/tools-iuc/tree/master/tools/hisat2).

Presumably the test:

```xml
        <test expect_num_outputs="1">
            <param name="type" value="paired" />
            <param name="source" value="indexed" />
            <param name="input_1" ftype="fastqsanger" dbkey="phiX" value="hisat_input_1_forward.fastq" />
            <param name="input_2" ftype="fastqsanger" dbkey="phiX" value="hisat_input_1_reverse.fastq" />
            <output name="output_alignments" file="hisat_output_1.bam" ftype="bam" lines_diff="2" />
        </test>
```

should be able to assert something like:

```xml
        <test expect_num_outputs="1">
            <requires_tool_data table="hisat2_indexes" dbkey="phiX" />
            <param name="type" value="paired" />
            <param name="source" value="indexed" />
            <param name="input_1" ftype="fastqsanger" dbkey="phiX" value="hisat_input_1_forward.fastq" />
            <param name="input_2" ftype="fastqsanger" dbkey="phiX" value="hisat_input_1_reverse.fastq" />
            <output name="output_alignments" file="hisat_output_1.bam" ftype="bam" lines_diff="2" />
        </test>
```

And tool tests should 'skip' if that data is unavailable.

mvdbeek:
This is totally worth having, but just for the sake of recording it:

When parsing a tool we can record whether a parameter populates it's options using `from_data_table` or `from_file` and include this in the error message if a job can't be run.
So the current error message looks sth. like:
```
RunToolException: Error creating a job for these tool inputs - An invalid option was selected for index, u'equCab2chrM', please verify
``` 

If this was instead something like:
```
ToolDataTableEntryMissingException: Error creating a job for these tool inputs - invalid option u'equCab2chrM was selected for  parameter 'index', which populates its options from tool data table 'all_fasta' , please verify
```
the test client could decide to mark the test as `skipped`. That could be complementary to explicitly stating that a tool data table is required as @jmchilton has suggested.

PR #9655, opened 2020-04-24
[20.05] More structured errors for tool validation problems.

jmchilton:
The goal is improved error handling during tool testing - trying to see if we can get to a point where we can skip tests if the target server simply doesn't have the needed .loc files. Alternative to https://github.com/galaxyproject/galaxy/issues/7998.

### DIFF
--- a/lib/galaxy/tool_util/verify/interactor.py
+++ b/lib/galaxy/tool_util/verify/interactor.py
@@ -375,9 +375,7 @@ class GalaxyInteractorApi(object):
                     "files_0|file_data": file_content
                 }
         submit_response_object = self.__submit_tool(history_id, "upload1", tool_input, extra_data={"type": "upload_dataset"}, files=files)
-        if submit_response_object.status_code != 200:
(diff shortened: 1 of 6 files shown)


mvdbeek:
> can skip tests if the target server simply doesn't have the needed .loc files.

Mechanism-wise this should just flat-out never attempt it. The loc files may be there, but are most likely containing the full-sized reference dataset instead of a small subset. What do you think about baking some logic into the Tool class for filtering tests that reference incompatible parameter types ?

jmchilton:
> The loc files may be there, but are most likely containing the full-sized reference dataset instead of a small subset.

Ohhhh... that is gross. I thought they had modified dbkeys for the subsets. I've removed the skip option from the CLI since it sounds like you don't think it will be effective but I've retained the more structured error handling - I think that stuff is all good plumbing changes.

jmchilton:
The CLI changes are in a commit at the head of this branch https://github.com/jmchilton/galaxy/tree/structured_errors_1 if the need arises for them.

jmchilton:
I'll reopen this against dev.

PR #9656, opened 2020-04-25
Allow submission of tools with unset hidden parameters

bernt-matthias:
A hidden parameter without value leads to a tool that can't be executed.

```
<tool id="test" name="test" version="1.0">
    <description>blah</description>
    <command detect_errors="exit_code">
        echo 1 > $out
    </command>
    <inputs>
        <param name="test" argument="-test" type="hidden"/>
    </inputs>
    <outputs>
        <data name="out" format="txt"/>
    </outputs>
    <tests>
        <test expect_num_outputs="1"/>
    </tests>
</tool>
```

Hard to find this, because even `planemo test` will be successful.

- Reformulated one error mesage where the param name was missing.
  So the error is now easier to find in the tool.
- Allow for a test that has only expect_num_outputs

### DIFF
--- a/client/src/mvc/tool/tool-form.js
+++ b/client/src/mvc/tool/tool-form.js
@@ -337,7 +337,7 @@ const View = Backbone.View.extend({
                 Galaxy.emit.debug("tool-form::validate()", "Retrieving input objects failed.");
                 continue;
             }
-            if (!input_def.optional && input_value == null) {
+            if (!input_def.optional && input_value == null && input_def.type != "hidden") {
                 this.form.highlight(input_id);
                 return false;
             }
--- a/lib/galaxy/tool_util/linters/inputs.py
+++ b/lib/galaxy/tool_util/linters/inputs.py
@@ -32,19 +32,21 @@ def lint_inputs(tool_xml, lint_ctx):
         if param_type == "data":
             if "format" not in param_attrib:
                 lint_ctx.warn("Param input [%s] with no format specified - 'data' format will be assumed.", param_name)
-
-        if param_type == "select":
+        elif param_type == "select":
             dynamic_options = param.get("dynamic_options", None)
             if dynamic_options is None:
                 dynamic_options = param.find("options")
 
             select_options = _find_with_attribute(param, 'option', 'value')
             if any(['value' not in option.attrib for option in select_options]):
-                lint_ctx.error("Option without value")
+                lint_ctx.error("Select [%s] has option without value", param_name)
 
             if dynamic_options is None and len(select_options) == 0:
                 message = "No options defined for select [%s]" % param_name
                 lint_ctx.warn(message)
+        elif param_type in ["hidden", "hidden_data"]:
+            if "value" not in param_attrib:
+                lint_ctx.error("Hidden param [%s] has no value", param_name)
 
         # TODO: Validate type, much more...
 
--- a/lib/galaxy/tool_util/linters/inputs.py
+++ b/lib/galaxy/tool_util/linters/inputs.py
@@ -45,8 +45,8 @@ def lint_inputs(tool_xml, lint_ctx):
                 message = "No options defined for select [%s]" % param_name
                 lint_ctx.warn(message)
         elif param_type in ["hidden", "hidden_data"]:
-            if "value" not in param_attrib:
-                lint_ctx.error("Hidden param [%s] has no value", param_name)
+            if "value" not in param_attrib and ('optional' not in option.attrib or option.attrib['optional'] == 'false'):
+                lint_ctx.error("Hidden non-optional param [%s] has no value (make it optional or provide a value)", param_name)
 
         # TODO: Validate type, much more...
 
--- a/lib/galaxy/tool_util/linters/inputs.py
+++ b/lib/galaxy/tool_util/linters/inputs.py
@@ -44,10 +44,6 @@ def lint_inputs(tool_xml, lint_ctx):
             if dynamic_options is None and len(select_options) == 0:
                 message = "No options defined for select [%s]" % param_name
                 lint_ctx.warn(message)
-        elif param_type in ["hidden", "hidden_data"]:
-            if "value" not in param_attrib and ('optional' not in option.attrib or option.attrib['optional'] == 'false'):
-                lint_ctx.error("Hidden non-optional param [%s] has no value (make it optional or provide a value)", param_name)
-
         # TODO: Validate type, much more...
 
     conditional_selects = tool_xml.findall("./inputs//conditional")
--- a/lib/galaxy/tool_util/linters/inputs.py
+++ b/lib/galaxy/tool_util/linters/inputs.py
@@ -37,7 +37,7 @@ def lint_inputs(tool_xml, lint_ctx):
             if dynamic_options is None:
                 dynamic_options = param.find("options")
 
-            select_options = _find_with_attribute(param, 'option', 'value')
+            select_options = param.findall('./option')
             if any(['value' not in option.attrib for option in select_options]):
                 lint_ctx.error("Select [%s] has option without value", param_name)
 
--- a/lib/galaxy/tool_util/linters/tests.py
+++ b/lib/galaxy/tool_util/linters/tests.py
@@ -15,14 +15,16 @@ def lint_tsts(tool_xml, lint_ctx):
     num_valid_tests = 0
     for test in tests:
         has_test = False
-        test_expect = {"expect_failure", "expect_exit_code", "expect_num_outputs"}
+        test_expect = ("expect_failure", "expect_exit_code", "expect_num_outputs")
         for te in test_expect:
             if te in test.attrib:
                 has_test = True
-        test_assert = {"assert_stdout", "assert_stderr", "assert_command"}
+                break
+        test_assert = ("assert_stdout", "assert_stderr", "assert_command")
         for ta in test_assert:
             if len(test.findall(ta)) > 0:
                 has_test = True
+                break
 
         output_data_names, output_collection_names = _collect_output_names(tool_xml)
         found_output_test = False
--- a/lib/galaxy/tool_util/linters/tests.py
+++ b/lib/galaxy/tool_util/linters/tests.py
@@ -15,14 +15,14 @@ def lint_tsts(tool_xml, lint_ctx):
     num_valid_tests = 0
     for test in tests:
         has_test = False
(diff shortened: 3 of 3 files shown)


lib/galaxy/tool_util/linters/inputs.py
```diff
@@ -29,22 +29,31 @@ def lint_inputs(tool_xml, lint_ctx):
         if not is_valid_cheetah_placeholder(param_name):
             lint_ctx.warn("Param input [%s] is not a valid Cheetah placeholder.", param_name)
 
+        if param_type not in ["text", "integer", "float", "color", "boolean",
```
mvdbeek: This is already being verified through the schema. Try adding a nonsense parameter type and lint it, this will fail with
```
Applying linter tool_xsd... FAIL
.. ERROR: Invalid XML found in file: CreateInterval.xml. Errors [/var/folders/df/6xqpqpcd7h73b6jpx9t6cwhw0000gn/T/tmpwo5mb19x:7:0:ERROR:SCHEMASV:SCHEMAV_CVC_ENUMERATION_VALID: Element 'param', attribute 'type': [facet 'enumeration'] The value 'notype' is not an element of the set {'text', 'integer', 'float', 'color', 'boolean', 'genomebuild', 'library_data', 'select', 'data_column', 'hidden', 'hidden_data', 'baseurl', 'file', 'data', 'drill_down', 'group_tag', 'data_collection'}.]
```

bernt-matthias: cool. undone this change. will fixup the commit.

lib/galaxy/tool_util/linters/inputs.py
```diff
@@ -29,22 +29,31 @@ def lint_inputs(tool_xml, lint_ctx):
...
 
             select_options = _find_with_attribute(param, 'option', 'value')
             if any(['value' not in option.attrib for option in select_options]):
-                lint_ctx.error("Option without value")
+                lint_ctx.error("Select [%s] has option without value", param_name)
 
             if dynamic_options is None and len(select_options) == 0:
                 message = "No options defined for select [%s]" % param_name
                 lint_ctx.warn(message)
+        elif param_type in ["hidden", "hidden_data"]:
```
mvdbeek: I don't remember what those were good for, are you 100% certain they always require a value ?
Both parameters seem to be initialized with `None` as their value.

bernt-matthias: Not at all for `hidden_data`. 

For `hidden` the tool can't be executed (the execute button just does nothing). But you are right, this might be caused by a bug elsewhere.

lib/galaxy/tool_util/linters/tests.py
```diff
@@ -15,7 +15,9 @@ def lint_tsts(tool_xml, lint_ctx):
     num_valid_tests = 0
     for test in tests:
         has_test = False
-        if "expect_failure" in test.attrib or "expect_exit_code" in test.attrib:
+        if "expect_failure" in test.attrib \
```
mvdbeek: That doesn't look like an improvement. If you want to change this how about creating a set outside of the loop `test_expect = {'expect_failure', 'expect_exit_code', 'expenct_num_outputs'}`
and then`if test_expect.intersection(set(test.attrib)):` here ?

lib/galaxy/tool_util/linters/inputs.py
mvdbeek: If you figure this out we should add some minimal documentation. Just searching the codebase I'm still not certain how they work exactly.

bernt-matthias:
Now I'm confused:

```
<param name="test" argument="-test" type="text" hidden="true"/>
```

works, but linting fails saying that hidden is not an attribute.

mvdbeek (approved):

lib/galaxy/tool_util/linters/inputs.py
```diff
@@ -32,19 +32,21 @@ def lint_inputs(tool_xml, lint_ctx):
...
             if "format" not in param_attrib:
                 lint_ctx.warn("Param input [%s] with no format specified - 'data' format will be assumed.", param_name)
-
-        if param_type == "select":
+        elif param_type == "select":
             dynamic_options = param.get("dynamic_options", None)
             if dynamic_options is None:
                 dynamic_options = param.find("options")
 
             select_options = _find_with_attribute(param, 'option', 'value')
```
nsoranzo: There was a copy-paste error here since this code was introduced in commit 630ac3658f3c64bd9a9abe5757cce0ed01cad229 , this will never find select options without a value.

```suggestion
            select_options = param.findall('./option')
```

lib/galaxy/tool_util/linters/inputs.py
nsoranzo: Counter-examples:
- for `hidden`:
  - https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/imp_exp/exp_history_to_archive.xml (this is a special tool though)
  - https://github.com/bgruening/galaxytools/blob/master/chemicaltoolbox/pgchem/pgchem/db_similarity_search.xml#L41
- for `hidden_data`: https://github.com/galaxyproject/tools-iuc/blob/master/tool_collections/cufflinks/cufflinks/cufflinks_wrapper.xml#L135

lib/galaxy/tool_util/linters/tests.py
```diff
@@ -15,14 +15,14 @@ def lint_tsts(tool_xml, lint_ctx):
...
         has_test = False
-        if "expect_failure" in test.attrib or "expect_exit_code" in test.attrib:
-            has_test = True
-        if len(test.findall("assert_stdout")) > 0:
-            has_test = True
-        if len(test.findall("assert_stderr")) > 0:
-            has_test = True
-        if len(test.findall("assert_command")) > 0:
-            has_test = True
+        test_expect = {"expect_failure", "expect_exit_code", "expect_num_outputs"}
```
nsoranzo:
```suggestion
        test_expect = ("expect_failure", "expect_exit_code", "expect_num_outputs")
```

lib/galaxy/tool_util/linters/tests.py
```diff
@@ -15,14 +15,14 @@ def lint_tsts(tool_xml, lint_ctx):
...
+        for te in test_expect:
+            if te in test.attrib:
+                has_test = True
+        test_assert = {"assert_stdout", "assert_stderr", "assert_command"}
```
nsoranzo:
```suggestion
        test_assert = ("assert_stdout", "assert_stderr", "assert_command")
```

lib/galaxy/tool_util/linters/tests.py
```diff
@@ -15,14 +15,14 @@ def lint_tsts(tool_xml, lint_ctx):
...
-        if len(test.findall("assert_stdout")) > 0:
-            has_test = True
-        if len(test.findall("assert_stderr")) > 0:
-            has_test = True
-        if len(test.findall("assert_command")) > 0:
-            has_test = True
+        test_expect = {"expect_failure", "expect_exit_code", "expect_num_outputs"}
+        for te in test_expect:
+            if te in test.attrib:
+                has_test = True
```
nsoranzo, line 21:
```suggestion
                has_test = True
                break
```

lib/galaxy/tool_util/linters/tests.py
```diff
@@ -15,14 +15,14 @@ def lint_tsts(tool_xml, lint_ctx):
...
+        test_assert = {"assert_stdout", "assert_stderr", "assert_command"}
+        for ta in test_assert:
+            if len(test.findall(ta)) > 0:
+                has_test = True
```
nsoranzo, line 26:
```suggestion
                has_test = True
                break
```

lib/galaxy/tool_util/linters/inputs.py
bernt-matthias: Thanks for looking this up @nsoranzo . Lets see if those tools can be executed. I re-tested my example from above. If one adds `optional="true"` the tool can be executed.  

The same holds for the cufflinks example (which has `hidden_data`), i.e. removing the `optional="true"` leads to a tool that can not be executed. 

For the pgchem there are two hidden inputs in two when branches and therefor can not be executed at all (it also has a typo which make it invalid ` </outputs>>`). 

I also had the impression that the tool form jumped a bit when I hit the execute button ... maybe it tries to scroll to the hidden input and show an error message? 

So, the question is if we want to extend the linting in the suggested way, or if we decide to call it a bug that such tools can not be executed. 

Since running the test tool via planemo works in any case (independent of a value or optional input) and also the history export tool works I tend to the latter and drop (most of) this PR... but I have no idea where to look for the bug. I guess its somewhere in the JS.

nsoranzo: > but I have no idea where to look for the bug. I guess its somewhere in the JS.

If some older version of Galaxy can run them, you can try to `git bisect` to find when this broke.

bernt-matthias: Tested 16.04, 17.01, 18.01, and 19.01 (using planemo serve): the test tool can not be executed in any of them. 

So I add a check for optional and then it should be fine.

mvdbeek: So the client prevents submission here (and that also explains why `optional="true"` works): https://github.com/galaxyproject/galaxy/blob/01ed5b92a54e22578893256b7ded0a86e236a5b3/client/src/mvc/tool/tool-form.js#L340

Since you can submit the tool via the API (that's why planemo test works) I would suggest removing that check for hidden parameter as a quick fix. Even better would be not sending out hidden parameters at all, I don't think the client needs to see these at all ... and in fact that may leak "secret"-y values. Not that that would be the right mechanism for secrets anyway.

bernt-matthias: OK. Seems that my guess on the JS change was wrong or not stylish enough :) suggestions? 

Can edit the commits later on if desired... removing the reverted ones...

client/src/mvc/tool/tool-form.js
```diff
@@ -337,7 +337,7 @@ const View = Backbone.View.extend({
                 Galaxy.emit.debug("tool-form::validate()", "Retrieving input objects failed.");
                 continue;
             }
-            if (!input_def.optional && input_value == null) {
+            if (!input_def.optional && input_value == null && input_def.type != 'hidden') {
```
mvdbeek: You can run prettier which auto-formats client code with
`make client-format`

Issue #9572, opened 2020-03-28
Rate-limit for tool shed's /api/repositories/updates 


mvdbeek:
We see episodic requests at a high rate coming in to /api/repositories/updates in the tool shed. [This route](https://github.com/galaxyproject/galaxy/blob/2feb9f9a6a594a694be1b352f462879a01d7f40d/lib/tool_shed/webapp/api/repositories.py#L693) calls out to external mercurial commands (with considerable python startup overhead) and overall it seems difficult to optimize the efficiency of this route. This route is being used in batch by Galaxy servers polling for updates. This can be very heavy, especially when multiple Galaxy processes (meaning multiple `UpdateRepositoryManager` all requesting updates) are started at the same time.

I think we should start by making sure that `url_get` https://github.com/galaxyproject/galaxy/blob/2feb9f9a6a594a694be1b352f462879a01d7f40d/lib/galaxy/util/__init__.py#L1679
understands HTTP code 429 and the `Retry-After` header. That means that in the future we can rate-limit this route (or a variant thereof) without affecting consumers. I suspect that for the existing route it would probably be ok to just add `time.sleep(0.3)` to reduce the CPU load given that the update requests are synchronous and not issued in parallel (meaning we wouldn't build up a huge backlog).

mvdbeek:
Solved this with a uwsgi hammer and `reload-on-rss`: https://github.com/galaxyproject/usegalaxy-playbook/commit/f2427d69caf792bd8a6092cfc8893d7fb68cbb7d

PR #9659, opened 2020-04-27
Respect retry-after header, raise on unexpected status codes

mvdbeek:
This is a pre-requisite for eventually rate-limiting certain tool shed routes: https://github.com/galaxyproject/galaxy/issues/9572


packages/util/requirements.txt
```diff
@@ -5,5 +5,6 @@ docutils
 markupsafe
 packaging
 pyyaml
+requests
```
martenson, line 8: should this have `responses` too?

mvdbeek: No, `responses` is for unit tests only.

martenson:
Thanks @mvdbeek, this is useful. Rate limiting and overall more restrictive/controlled approach to handling API calls is something we should move towards.

PR #9667, opened 2020-04-27
Checkbox bugfix

### LINKED ISSUE #9663: Checkboxes not checkable on dev
I could have sworn that I already opened an issue, but there it is:
On current dev checkboxes are displayed, but are not clickable.

dannon:
Fixes #9663


mvdbeek:
That fixed it partially, checkboxes now work, and to get radio buttons to work I added https://github.com/mvdbeek/galaxy/commit/a7b3b814a0c5161e724862578400719043a2b22d in the same way you did here, and now it works but it is a little too close to to text maybe ?
Screenshot 2020-04-28 at 12 32 03
I removed our modifiers in ac7bf233d0 and that looks nice:
Screenshot 2020-04-28 at 12 43 36


There's one more instance of `font-family: FontAwesome;` in https://github.com/dannon/galaxy/blob/a7b3b814a0c5161e724862578400719043a2b22d/client/galaxy/style/scss/ui/paired-collection-creator.scss#L267, that should probably also be changed ?

dannon:
Yeah, I'm fixing up the pair creator now, then we can get this in.

mvdbeek:
Thank you @dannon !

PR #9674, opened 2020-04-28
add doc for discover_datasets sort_by

bernt-matthias:
I guess this is also needed for `data > discover_datasets`? 

In this context I would really like to understand the meaning of `name`, `filename` and `designation` .. I never understood this.

Also wondering if this deserves a test .. never have seen this in action.


jmchilton:
test/unit/tools/test_collect_primary_datasets.py has the unit tests I wrote for this I think - but yeah functional tests with actual tools would be awesome.

bernt-matthias:
I'm on it. Any comments on my question regarding the meaning of `name`, `filename` and `designation`

PR #9677, opened 2020-04-28
Fixed an issue in object store integration tests

kxk302:
As I was writing the documentation for Galaxy debugging, I wanted to test my document, so I tried to debug object store integration tests. I put a breakpoint in irods.py and the break point was never hit! I debugged the issue and it seems in handle_galaxy_config_kwds(), object_store_conf_file was not assigned, causing the default object store to be used. The tests would pass but they were not doing what was expected of them. Added a line handle_galaxy_config_kwds() to assign object_store_conf_file and the tests now pass AND work as expected.


kxk302:
Thx @jmchilton!

PR #9678, opened 2020-04-28
Enhance toolshed selenium test.

davebx:
This adds a check for the presence of the upgrade notification when an older version of a tool is installed.


dannon (approved):
👍 watched it a few times locally, looks good.

PR #9680, opened 2020-04-28
Optimize history_content union and filtering query

Nerdinacan:
This is a subset of #9591. It was requested that this chunk be separated from the other changes in that PR to ease testing complexity. Additionally, these optimizations do not add any new functionality and should be able to work on the existing code as-is.

### Optimize the primary history contents join

Attempting to create a more efficient history contents query. The api/history/#/contents endpoint is one that the client needs to hit frequently for updates and I found that the primary mechanism that unions history datasets and history collections into one query was running inefficiently.

I think the basic strategy of the original code is reasonable, but it looks like nobody was paying attention to the SQL that was emitted by sqlalchemy.

Essentially what i did was to turn the primary history contents join from this:

```sql
select manycontentfields
from (
   select hda_fields 
   from history_dataset_association 
   where history_id = xyz
   union
   select hdca_fields 
   from history_dataset_collection_association 
   where history_id = xyz
)
where name = 'foo'
and deleted = false
and purged = false
and update_time > 5minsago
order by hid desc
limit 100 offset 0
```

Into this:

```sql
select manycontentfields
from (
   select hda_fields 
   from history_dataset_association 
   where history_id = xyz
   and name = 'foo'
   and deleted = false
   and purged = false
   and update_time > 5minsago

   union all

   select hdca_fields 
   from history_dataset_collection_association 
   where history_id = xyz
   and name = 'foo'
   and deleted = false
   and purged = false
   and update_time > 5minsago
)
order by hid desc
limit 100 offset 0
```

It's not a big change, but the current code materializes (loads in entirety) 2 whole tables, stacks them on top of each other, and sorts through them looking for duplicates that will never exist. I would imagine this gets pretty ugly for big histories. The new union will only result in a few rows.

Changing "union" to "union all" gives a performance bump too because "union" also sorts through all those table results looking for duplicates that will not exist. Union all just stacks the results

### DIFF
--- a/lib/galaxy/managers/history_contents.py
+++ b/lib/galaxy/managers/history_contents.py
@@ -294,12 +294,11 @@ class HistoryContentsManager(containers.ContainerManagerMixin):
             if orm_filter.filter_type == "orm_function":
                 contained_query = contained_query.filter(orm_filter.filter(self.contained_class))
                 subcontainer_query = subcontainer_query.filter(orm_filter.filter(self.subcontainer_class))
-        contents_query = contained_query.union(subcontainer_query)
+            elif orm_filter.filter_type == "orm":
+                contained_query = self._apply_orm_filter(contained_query, orm_filter.filter)
+                subcontainer_query = self._apply_orm_filter(subcontainer_query, orm_filter.filter)
 
-        for orm_filter in filters:
-            if orm_filter.filter_type == "orm":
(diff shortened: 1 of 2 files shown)


jmchilton (approved):
Amazing - thanks for responding so quickly!

mvdbeek:
Awesome, thank you @Nerdinacan !

PR #9642, opened 2020-04-21
Python37 lock fix

dannon:
Fixes #9641 

One concern is that `pbkdf2_hmac` was added in python 2.7.8.  I'm not *that* worried about it because that was available July 2, 2014, and we're not supporting 2.7 anymore anyway, but...  It's worth thinking about.

PR #9681, opened 2020-04-29
[20.01] Backport auth process lock fix

dannon:
flattened backport of, and with additions to (for compatibility), #9642 

I managed to see this in 3.6.10, which prompted me to want to finish this backport if it's going to affect more python versions than 3.7+

When we merge this forward we can drop the compatibility logic if we really want to cut off older pythons, but an informal survey of admins did find a few at 2.7.5, which wouldn't work without the failthrough.

### DIFF
--- a/lib/galaxy/security/passwords.py
+++ b/lib/galaxy/security/passwords.py
@@ -89,9 +89,7 @@ def pbkdf2_bin(data, salt, iterations=COST_FACTOR, keylen=KEY_LENGTH, hashfunc=H
             h = mac.copy()
             h.update(x)
             digest = h.digest()
-            if six.PY2:
-                return digest, [ord(_) for _ in digest]
-            return digest, digest
+            return digest, [ord(_) for _ in digest]
 
         buf = []
         salt = smart_str(salt)
--- a/lib/galaxy/security/passwords.py
+++ b/lib/galaxy/security/passwords.py
@@ -6,8 +6,6 @@ from operator import xor
 from os import urandom
 from struct import Struct
 
-import six
-
 from galaxy.util import (
     safe_str_cmp,
     smart_str,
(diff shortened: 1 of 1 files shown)


lib/galaxy/security/passwords.py
```diff
@@ -67,29 +67,38 @@ def check_password_PBKDF2(guess, hashed):
...
+        if isinstance(hashfunc, six.string_types):
+            hashfunc = getattr(hashlib, hashfunc, None)
+        hashfunc = hashfunc or hashlib.sha1
+        mac = hmac.new(smart_str(data), None, hashfunc)
+
+        def _pseudorandom(x, mac=mac):
+            h = mac.copy()
+            h.update(x)
+            digest = h.digest()
+            if six.PY2:
```
mvdbeek: I guess we don't need the guard anymore

mvdbeek (approved):

lib/galaxy/security/passwords.py
dannon: Ahh, good point -- I left the whole block intact for the backport, but this can be simplified.  I guess technically python 3.3 doesn't have hashlib.pbkdf2_hmac, either, but surely nobody made the leap to 3 and is running something that old that we don't support anyway.  I'll drop this.

lib/galaxy/security/passwords.py
```diff
@@ -67,29 +67,36 @@ def check_password_PBKDF2(guess, hashed):
...
+    if hasattr(hashlib, 'pbkdf2_hmac'):
+        # Use hashlib.pbkdf2_hmac, new in python 2.7.8
+        data = smart_str(data)
+        salt = smart_str(salt)
+        return hashlib.pbkdf2_hmac(hashfunc, data, salt, iterations, keylen)
+    else:
+        # Use the old implementation of hashing as a fallback
+        if isinstance(hashfunc, six.string_types):
+            hashfunc = getattr(hashlib, hashfunc, None)
+        hashfunc = hashfunc or hashlib.sha1
```
nsoranzo: Given that all callers of this function are in this file, I think this can be simplified.

```suggestion
        hashfunc = getattr(hashlib, hashfunc, None)
```

PR #9682, opened 2020-04-29
Collection pagination api resource route

Nerdinacan:
More changes to the api to support the new history client.

### Pagination parameters for collection contents
In order to manage large collections on the client, it'll be necessary to have a way to paginate through collections. Currently collections are delivered as a complete composed tree which is unnecessary because the client only views one leaf of that tree at a time. This is also processing intensive for large collections and makes pagination an exercise in black magic.

The new /contents endpoint displays the direct children of the indicated parent collection_id. It has limit and offset parameters but no filtering features yet as those don't exist in the client.

If we do later implement filtering features, I'd like to also renovate the existing filter syntax for the other api endpoints (q= & qv=) as the current methodology offers absolutely no utility beyond the standard url parameter format (field=value). But, I think this is probably worth its own PR.

The new endpoint is located at
```
/api/dataset_collections/{parent_id}/contents?limit=100&offset=20
```

...where the parent_id is the dataset_collection.id (not the hdca.id) whose contents are to be displayed.

### Contents url property in HDCA and dataset collection element results
In addition, a contents_url property has been added to HDCA summary serializations as well as the  collection element items on the new endpoint itself to make drill-down requests easier.

#### /api/dataset_collections/{parent_id}/contents?limit=100&offset=20
Screen Shot 2020-04-28 at 10 55 26 PM

#### /api/histories/f597429621d6eb2b/contents?order=hid&v=dev
Screen Shot 2020-04-28 at 10 56 59 PM


mvdbeek:
Nice! I guess my only request would be a simple API test, let me know if you need any help with that.

lib/galaxy/webapps/galaxy/api/dataset_collections.py
```diff
@@ -81,6 +86,38 @@ def show(self, trans, instance_type, id, **kwds):
...
+        GET /api/dataset_collection/{id}/contents
+
+        Optional pagination parameters
+            limit:  integer
+            offset: integer
+        """
+
+        limit = kwds.get('limit', None)
+        offset = kwds.get('offset', None)
+        decoded_id = trans.app.security.decode_id(id)
```
jmchilton: You're sure the way you're using this you have a DatasetCollection ID and not a HistoryDatasetCollectionAssocation ID? It is easy to make these two overlap for all IDs in simple tests - but they definitely won't in production. The other API methods in this class take in an association ID and do security checks on the associated object. They also do security checks on the associated object - I'm nervous about that not being here... I don't know if it is enough to just check the datasets.

jmchilton: No - I guess looking at your example - the collection ID is different than the HDCA type ID. So this seems fine. Nice work. I'm still nervous about security but let me think about that.

jmchilton: ```dictify_element_reference``` doesn't do any checks - so I the other methods that use it are counting on the check of the HDCA for security. I think that means your API route gives access to dataset contents without any checks unless I'm missing something.

Nerdinacan: @jmchilton I wondered about security as well. As I see it, only the top level of the collections have an HDCA and therefore a direct connection to the history and therefore the user/security layer, unless I misunderstand. 

I could maybe write a secondary security query by using a recursive check up the parents of the selected layer to find a matching user history if you think that would be appropriate. It'll give me an excuse to exploit recursive CTEs in a query.

The only reason it's not here now is because nested collections don't necessarily have an HCDA, right? They are all considered to be part of the single collection object. Did I understand that correctly?

Nerdinacan:
@mvdbeek 
> Nice! I guess my only request would be a simple API test, let me know if you need any help with that.

Agreed, I still need to write a few tests for it.

lib/galaxy/webapps/galaxy/api/dataset_collections.py
Nerdinacan: @jmchilton I've included a standard security check from the existing code-base as well as a lookup that makes sure the indicated collection is actually a member of the referenced hdca.

Nerdinacan:
Ok, format of the url has changed slightly. Now, to get the contents of a collection you also pass in the HDCA id so that we can properly perform authorization. The new Format is:

```
/api/dataset_collections/[encoded hdca id]/contents/[encoded parent collection id]
```
This will run a check using the HDCA to ensure the user can access the HDCA then run a second check to make sure the indicated collection is a member of that HDCA using the following sql:

```sql

WITH RECURSIVE element_parents(dataset_collection_id) AS 
(
    SELECT dataset_collection_element.dataset_collection_id AS dataset_collection_id 
    FROM dataset_collection_element 
    WHERE dataset_collection_element.child_collection_id = 7 
    OR dataset_collection_element.dataset_collection_id = 7
    UNION 
    SELECT dce.dataset_collection_id AS dataset_collection_id 
    FROM dataset_collection_element AS dce, element_parents AS ep 
    WHERE dce.child_collection_id = ep.dataset_collection_id
)
    SELECT hdca.id AS hdca_id 
    FROM element_parents JOIN history_dataset_collection_association AS hdca 
    ON hdca.collection_id = element_parents.dataset_collection_id 
    WHERE hdca.id = 3
```

.... which was much more irritating to write in SQLAlchemy than it would have been otherwise. The query is a quick recursive CTE expression which shuffles up the collection parents until we find one that matches the relevant HDCA.

Nerdinacan:
> Nice! I guess my only request would be a simple API test, let me know if you need any help with that.

Done!

jmchilton:
Beautiful - excellent tests, cool query, clean API - I love it all. Thanks a bunch @Nerdinacan

jmchilton (approved):

jmchilton:
Opps - it was conflicted and I tried to push a merge and it had a syntax error. Sorry

PR #9684, opened 2020-04-29
xsd, docs, and test for sort_by (discover_datasets)

bernt-matthias:
Since `planemo lint` failed so far for this attribute we can be quite sure that this was never used (in a tool)

- order of sort_comp and sort_by was wrong for sort_by in collections
- sort_by was missing for data output
- adds a test (though we can't check the generated order .. I manually tested with `planemo serve`)

follow up on https://github.com/galaxyproject/galaxy/pull/9674

### DIFF
--- a/lib/galaxy/tool_util/parser/xml.py
+++ b/lib/galaxy/tool_util/parser/xml.py
@@ -605,12 +605,13 @@ def __parse_output_collection_elem(output_collection_elem):
 
 def __parse_element_tests(parent_element):
     element_tests = {}
-    for element in parent_element.findall("element"):
+    for idx, element in enumerate(parent_element.findall("element")):
         element_attrib = dict(element.attrib)
         identifier = element_attrib.pop('name', None)
         if identifier is None:
             raise Exception("Test primary dataset does not have a 'identifier'")
         element_tests[identifier] = __parse_test_attributes(element, element_attrib, parse_elements=True)
+        element_tests[identifier][1]["element_index"] = idx
     return element_tests
 
 
--- a/lib/galaxy/tool_util/verify/interactor.py
+++ b/lib/galaxy/tool_util/verify/interactor.py
@@ -708,12 +708,6 @@ def verify_hid(filename, hda_id, attributes, test_data_downloader, hid="", datas
 def verify_collection(output_collection_def, data_collection, verify_dataset):
     name = output_collection_def.name
 
-    def get_element(elements, id):
-        for element in elements:
-            if element["element_identifier"] == id:
-                return element
-        return False
-
     expected_collection_type = output_collection_def.collection_type
     if expected_collection_type:
         collection_type = data_collection["collection_type"]
@@ -731,15 +725,23 @@ def verify_collection(output_collection_def, data_collection, verify_dataset):
             raise AssertionError(message)
 
     def verify_elements(element_objects, element_tests):
+        i = 0
         for element_identifier, element_test in element_tests.items():
             if isinstance(element_test, dict):
                 element_outfile, element_attrib = None, element_test
             else:
                 element_outfile, element_attrib = element_test
 
-            element = get_element(element_objects, element_identifier)
-            if not element:
-                template = "Failed to find identifier [%s] for testing, tool generated collection elements [%s]"
+            element = None
+            while i < len(element_objects):
+                if element_objects[i]["element_identifier"] == element_identifier:
+                    element = element_objects[i]
+                    i += 1
+                    break
+                i += 1
+
+            if element is None:
+                template = "Failed to find identifier [%s] for testing or elements in wrong order, tool generated collection elements [%s]"
                 message = template % (element_identifier, element_objects)
                 raise AssertionError(message)
 
--- a/lib/galaxy/tool_util/verify/interactor.py
+++ b/lib/galaxy/tool_util/verify/interactor.py
@@ -725,8 +725,17 @@ def verify_collection(output_collection_def, data_collection, verify_dataset):
             raise AssertionError(message)
 
     def verify_elements(element_objects, element_tests):
-        i = 0
+        sorted_test_ids = [None] * len(element_tests)
         for element_identifier, element_test in element_tests.items():
+            if isinstance(element_test, dict):
+                element_outfile, element_attrib = None, element_test
+            else:
+                element_outfile, element_attrib = element_test
+            sorted_test_ids[element_attrib["element_index"]] = element_identifier
+
+        i = 0
+        for element_identifier in sorted_test_ids:
+            element_test = element_tests[element_identifier]
             if isinstance(element_test, dict):
                 element_outfile, element_attrib = None, element_test
             else:
@@ -741,8 +750,10 @@ def verify_collection(output_collection_def, data_collection, verify_dataset):
                 i += 1
 
             if element is None:
-                template = "Failed to find identifier [%s] for testing or elements in wrong order, tool generated collection elements [%s]"
-                message = template % (element_identifier, element_objects)
+                template = "Failed to find identifier '%s' of test collection %s in the tool generated collection elements %s (at the correct position)"
+                eo_ids = [_["element_identifier"] for _ in element_objects]
+                message = template % (element_identifier, sorted_test_ids,
+                                      eo_ids)
                 raise AssertionError(message)
 
             element_type = element["element_type"]
--- a/lib/galaxy/web/framework/decorators.py
+++ b/lib/galaxy/web/framework/decorators.py
@@ -325,7 +325,7 @@ def format_return_as_json(rval, jsonp_callback=None, pretty=False):
 
     Use `pretty=True` to return pretty printed json.
     """
-    dumps_kwargs = dict(indent=4) if pretty else {}
+    dumps_kwargs = dict(indent=4, sort_keys=True) if pretty else {}
     json = safe_dumps(rval, **dumps_kwargs)
     if jsonp_callback:
         json = "{}({});".format(jsonp_callback, json)
--- a/lib/galaxy/web/framework/decorators.py
+++ b/lib/galaxy/web/framework/decorators.py
@@ -325,7 +325,7 @@ def format_return_as_json(rval, jsonp_callback=None, pretty=False):
 
     Use `pretty=True` to return pretty printed json.
     """
-    dumps_kwargs = dict(indent=4, sort_keys=True) if pretty else {}
+    dumps_kwargs = dict(indent=4) if pretty else {}
     json = safe_dumps(rval, **dumps_kwargs)
     if jsonp_callback:
         json = "{}({});".format(jsonp_callback, json)
--- a/test/functional/tools/discover_sort_by.xml
+++ b/test/functional/tools/discover_sort_by.xml
@@ -1,9 +1,10 @@
 <tool id="discover_sort_by" name="discover_sort_by" version="0.1.0">
     <command><![CDATA[
-    echo "1" > 1.txt &&
-    echo "2" > 2.txt &&
-    echo "10" > 10.txt
-  ]]></command>
+for i in \$(seq 1 10);
+do
+    echo "\$i" > \$i.txt;
+done
+]]></command>
   <inputs/>
   <outputs>
     <collection name="collection_numeric_name" type="list" label="num">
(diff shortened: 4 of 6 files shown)


test/functional/tools/discover_sort_by.xml
```diff
@@ -0,0 +1,26 @@
+<tool id="collection_split_on_column" name="collection_split_on_column" version="0.1.0">
```
mvdbeek: You have to add the tool to test/functional/tools/samples_tool_conf.xml to run the tests

bernt-matthias: Hah. Just forgot to commit it.

bernt-matthias:
seems ready

test/functional/tools/discover_sort_by.xml
```diff
@@ -0,0 +1,26 @@
...
+      <discover_datasets pattern="__name_and_ext__" sort_by="reverse_numeric_name" directory="outputs" />
+    </collection>
+    <data name="data_numeric_name">
+      <discover_datasets pattern="__name_and_ext__" format="tabular" assign_primary_output="true" sort_by="reverse_numeric_name" directory="outputs" visible="true"/>
+    </data>
+  </outputs>
+  <tests>
+    <test>
+      <param name="input1" value="tinywga.fam" />
+      <output_collection name="collection_numeric_name" type="list" count="11"/>
```
mvdbeek: Don't you want to ascertain the new sort order ? You can list `<element/>` items here

bernt-matthias: That's true, but I did not expect the order of the elements is relevant. Lets try.

bernt-matthias: Test currently do not check for the order, but only if the elements contained in the test definition are produced by the test: https://github.com/galaxyproject/galaxy/blob/94390c2af8fb1ce8f54c6665810991da36e6331a/lib/galaxy/tool_util/verify/interactor.py#L717

This would be (quite) easy to change, but 

- the test definition is a dict and dictionary order can only be used for python >= 3.7 
- dictionary order would always be lexicographic as defined for strings
- it seems that the dictionary order is lost in the `get_tool_tests` function called here https://github.com/galaxyproject/galaxy/blob/94390c2af8fb1ce8f54c6665810991da36e6331a/lib/galaxy/tool_util/verify/interactor.py#L803

Seems difficult. 

Spontaneous ideas

- sort the test elements in verify_elements as given by `sort_by`
- leave it as is and consider the test just checking for bugs

mvdbeek: > * the test definition is a dict and dictionary order can only be used for python >= 3.7

3.6, it was made official in 3.7



> * dictionary order would always be lexicographic as defined for strings

this should be by insertion order on newer python's, right ?

Anyway, I agree this is not as straightforward. Maybe add a  in the test, and then we can reconsider once sort order is guaranteed or we decide to write an API test, which could also assert this.

bernt-matthias: > 3.6, it was made official in 3.7

true

> this should be by insertion order on newer python's, right ?

also true, my mistake

> Anyway, I agree this is not as straightforward. Maybe add a in the test, and then we can reconsider once sort order is guaranteed or we decide to write an API test, which could also assert this.

Still I have the urge to understand this :) Somewehere here:

https://github.com/galaxyproject/galaxy/blob/94390c2af8fb1ce8f54c6665810991da36e6331a/lib/galaxy/tool_util/verify/interactor.py#L141

the order seems to get lost. Any suggestions where I could dig?

bernt-matthias:
Got it. 

Problem seems to be that `format_return_as_json` is called with `pretty=True` which sorts keys in the generated json (https://github.com/galaxyproject/galaxy/blob/7543b91001890a4969aef9575a455387a4f843ed/lib/galaxy/web/framework/decorators.py#L323) 

`pretty=True` because `debug` is `True` `https://github.com/galaxyproject/galaxy/blob/7543b91001890a4969aef9575a455387a4f843ed/lib/galaxy/web/framework/decorators.py#L297` over here. 

Curious if this causes tests to fail... 

One could add a tests with wrong element order ... ?

We should add docs here if this gets accepted https://docs.galaxyproject.org/en/master/dev/schema.html#tool-tests-test-output-collection

lib/galaxy/web/framework/decorators.py
```diff
@@ -326,7 +326,7 @@ def format_return_as_json(rval, jsonp_callback=None, pretty=False):
 
     Use `pretty=True` to return pretty printed json.
     """
-    dumps_kwargs = dict(indent=4, sort_keys=True) if pretty else {}
+    dumps_kwargs = dict(indent=4) if pretty else {}
```
mvdbeek: I don't think we should do this until we sunset python 3.5. You could add the index of the element [here](https://github.com/mvdbeek/galaxy/blob/dev/lib/galaxy/tool_util/parser/xml.py#L601) and then sort the test elements where needed. Do you maybe want to just add the xsd changes in for 20.05 ?

bernt-matthias: Your are probably right -- even if it seems safe in most places where `format_return_as_json` is used (pretty is either set to `False` or set to `True` only in debug mode). 

The only place where its `True` is here: 

https://github.com/galaxyproject/galaxy/blob/7eb2fcb708e7b63e17800c87613ddfa5497c0654/lib/galaxy/webapps/galaxy/api/workflows.py#L485

An alternative might be to add another parameter (e.g. `sort_keys=False`) to `format_return_as_json`. Tell me your preference and I will go ahead :)

mvdbeek: I'd go with the xsd changes only, but if you want to dig deeper I think that adding the index of the element to the test definition seems more solid to me.

bernt-matthias:
Done :) 

Also changed the text of the Exception to something like:  

```
Failed to find identifier '2' of test collection ['1', '10', '2'] in the tool generated collection elements ['1', '2', '10'] (at the correct position)
```

before the complete dict for the tool generated collection elements was dumped .. For me there was never any information of interest except for the IDs. Do you think this is OK.

bernt-matthias:
Still, for `<output><discovered_dataset>` order is not checked. Only the first element needs to be at the correct position as far as I see.

test/functional/tools/discover_sort_by.xml
```diff
@@ -0,0 +1,77 @@
...
+  </outputs>
+  <tests>
+    <!-- note: this test is only checking for bugs in the sorting functions
+         but not for the actual sorting, since the.
+         even if the test lists the discovered datasets in the correct order
+         the order is currently not tested
+         - test is only if the elements listed in the test
+           are in the list of generated elements
+         - furthermore the order of the elements listed in the test
+           is not preserved in the test functions -->
```
mvdbeek: I'm not quite sure I understand the comment, but also, does the comment still apply ?

mvdbeek (approved):

lib/galaxy/tool_util/verify/interactor.py
```diff
@@ -715,16 +709,35 @@ def get_element(elements, id):
...
+                    i += 1
+                    break
+                i += 1
 
-            element = get_element(element_objects, element_identifier)
-            if not element:
-                template = "Failed to find identifier [%s] for testing, tool generated collection elements [%s]"
-                message = template % (element_identifier, element_objects)
+            if element is None:
+                template = "Failed to find identifier '%s' of test collection %s in the tool generated collection elements %s (at the correct position)"
```
mvdbeek, line 753: ~~Isn't `the corrent position` `i` ?~~
Is the correct position the index of the iteration over sorted_test_ids ?

mvdbeek: and wouldn't this mean you can simplify the logic to

```python
for element_index, expected_element_identifier in enumerate(sorted_test_ids):
    assert element_objects[element_index]["element_identifier"] == expected_element_identifier
```
?

(You could add some length handling for missing elements)

test/functional/tools/discover_sort_by.xml
bernt-matthias: you are right. I removed it.

lib/galaxy/tool_util/verify/interactor.py
bernt-matthias: No, the test is just checking if the elements listed in the test occur in the same order in the generated elements. 

Thereby the tool author does not need to list all elements (which might be very many).

mvdbeek: Alright, then how about this:
```python
for element_index, expected_element_identifier in enumerate(sorted_test_ids):
    if expected_element_identifier is None:
        break
    assert element_objects[element_index]["element_identifier"] == expected_element_identifier
```
That way it's clear at which position the test data differs, and I think the logic is a bit easier to follow ?

bernt-matthias: I don't see why this should work :( .. 

The `None` entries from here: https://github.com/galaxyproject/galaxy/blob/a5b898c35bee08ada119bd0f7c3912242b1bae1c/lib/galaxy/tool_util/verify/interactor.py#L728

should all be overwritten by the actual identifiers. I only use `[None] * len(element_tests)` to initialize a list of the needed length.

I extended the test in order to clarify what I meant. The tool new generates 10 elements, but the test only uses three of them to verify the sorting order. The number of generated elements is tested with the `count` attribute. 

Formally spoken we test if the list of elements in the test is a subsequence of the generated elements.

mvdbeek: what is element_tests ? Is that not the list of specified tests?

bernt-matthias: Its a `dict`, looking like this

```
{'1': [None, {'assert_list': [{'attributes': {'expression': '^.*$'}, 'children': [], 'tag': 'has_text_matching'}], 'checksum': None, 'compare': 'diff', 'decompress': False, 'delta': 10000, 'delta_frac': None, 'element_index': 0, 'elements': {}, 'extra_files': [], 'lines_diff': 0, 'md5': None, 'metadata': {}, 'primary_datasets': {}, 'sort': False}], '10': [None, {'assert_list': [{'attributes': {'expression': '^.*$'}, 'children': [], 'tag': 'has_text_matching'}], 'checksum': None, 'compare': 'diff', 'decompress': False, 'delta': 10000, 'delta_frac': None, 'element_index': 2, 'elements': {}, 'extra_files': [], 'lines_diff': 0, 'md5': None, 'metadata': {}, 'primary_datasets': {}, 'sort': False}], '2': [None, {'assert_list': [{'attributes': {'expression': '^.*$'}, 'children': [], 'tag': 'has_text_matching'}], 'checksum': None, 'compare': 'diff', 'decompress': False, 'delta': 10000, 'delta_frac': None, 'element_index': 1, 'elements': {}, 'extra_files': [], 'lines_diff': 0, 'md5': None, 'metadata': {}, 'primary_datasets': {}, 'sort': False}]}
```

mvdbeek:
That's definitely a nice improvement, thanks a lot @bernt-matthias!

lib/galaxy/tool_util/verify/interactor.py
jmchilton: Maybe calling it element_index is a bad idea if it isn't expected to match the actual element index. Something like expected_order_index maybe would be better?

I don't really have the throughput to deal with this, but maybe this change requiring the order to match should be hidden behind a profile version tag for the tool? If there wasn't a ton of tool breakage maybe it is not worth the effort, I will just update Planemo regardless.

lib/galaxy/tool_util/parser/xml.py
```diff
@@ -605,12 +605,13 @@ def __parse_output_collection_elem(output_collection_elem):
...
 def __parse_element_tests(parent_element):
     element_tests = {}
-    for element in parent_element.findall("element"):
+    for idx, element in enumerate(parent_element.findall("element")):
         element_attrib = dict(element.attrib)
         identifier = element_attrib.pop('name', None)
         if identifier is None:
             raise Exception("Test primary dataset does not have a 'identifier'")
         element_tests[identifier] = __parse_test_attributes(element, element_attrib, parse_elements=True)
+        element_tests[identifier][1]["element_index"] = idx
```
jmchilton, line 614: This breaks YAML parsing because the general interface assumes this is here but only the XML parser adds it. This would break CWL tools, YAML tools, and Galaxy workflows (this last one breaks Planemo CI after this was published). So I'm going to refactor this addition into TestCollectionOutputDef - it will just make sure this is populated there. I hope that makes sense.

jmchilton:
I think my issues are addressed in https://github.com/galaxyproject/galaxy/pull/10434/files. What do y'all think about that approach? Sorry for not catching this sooner.

PR #9685, opened 2020-04-29
[20.01] Don't treat ConnectedValues as runtime input

mvdbeek:
This is needed for connected workflow parameters in subworkflows.
Otherwise these values need to be filled in in the workflow run form (which isn't correct and also doesn't work).

### DIFF
--- a/lib/galaxy/workflow/modules.py
+++ b/lib/galaxy/workflow/modules.py
@@ -540,7 +540,7 @@ class SubWorkflowModule(WorkflowModule):
                     if input.type in ['data', 'data_collection']:
                         return
 
-                    if is_runtime_value(value):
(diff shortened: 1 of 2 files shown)


lib/galaxy/workflow/modules.py
```diff
@@ -540,7 +540,7 @@ def callback(input, prefixed_name, prefixed_label, value=None, **kwds):
                     if input.type in ['data', 'data_collection']:
                         return
 
-                    if is_runtime_value(value):
+                    if is_runtime_value(value) and runtime_to_json(value)["__class__"] != "ConnectedValue":
```
guerler, line 543: Recommending to define a `is_connected_value` function here.

jmchilton (approved):

mvdbeek:
Thank you @guerler. I'll follow up with this in dev!

PR #9690, opened 2020-04-30
Add CSS class to messages in error report pages to make errors more visible

hexylena:
From

image

To

image


dannon:
Good idea, thanks!

PR #9699, opened 2020-05-01
[20.01] Cannot find migrate directory for tools database

Takadonet:
Upgrading from Galaxy 19.05 to 20.01 getting nice stacktrace indicating that could not from migrate directory for [tools section](https://github.com/galaxyproject/galaxy/blob/release_20.01/lib/galaxy/model/orm/scripts.py#L32).

Looks like it moved from Release 19.05 to release 20.01

### DIFF
--- a/lib/galaxy/model/orm/scripts.py
+++ b/lib/galaxy/model/orm/scripts.py
@@ -29,7 +29,7 @@ DATABASE = {
         },
     "tools":
         {
-            'repo': 'tool_shed/galaxy_install/migrate',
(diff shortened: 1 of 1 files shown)


mvdbeek (approved):
Thanks @Takadonet ! It's safe to turn off `check_migrate_tools`, which skips this check. I don't think we'll be migrating things to the toolshed anymore (and if we do it'll be a big announcement).

Takadonet:
Awesome to know thanks @mvdbeek

PR #9700, opened 2020-05-02
Refactor helper functions

lawrence14701-zz:
This pull request condenses the size of the analysis router file by reusing the _display_vue_helper function. Some legacy code also was removed from the show_custom_builds function.

### DIFF
--- a/client/galaxy/scripts/entry/analysis/AnalysisRouter.js
+++ b/client/galaxy/scripts/entry/analysis/AnalysisRouter.js
@@ -160,16 +160,11 @@ export const getAnalysisRouter = (Galaxy) =>
         },
 
         show_visualizations_sharing: function () {
-            const sharingInstance = Vue.extend(Sharing);
-            const vm = document.createElement("div");
-            this.page.display(vm);
-            new sharingInstance({
-                propsData: {
-                    id: QueryStringParsing.get("id"),
-                    plural_name: "Visualizations",
-                    model_class: "Visualization",
-                },
-            }).$mount(vm);
+            this._display_vue_helper(Sharing, {
+                id: QueryStringParsing.get("id"),
+                plural_name: "Visualizations",
+                model_class: "Visualization",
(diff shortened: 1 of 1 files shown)


jmchilton:
@galaxybot test this

client/galaxy/scripts/entry/analysis/AnalysisRouter.js
```diff
@@ -339,24 +312,11 @@ export const getAnalysisRouter = (Galaxy) =>
...
-            const vm = document.createElement("div");
-            this.page.display(vm);
-            new workflowImportInstance().$mount(vm);
+            this._display_vue_helper(WorkflowImport);
         },
 
         show_custom_builds: function () {
-            const historyPanel = this.page.historyPanel.historyView;
-            if (!historyPanel || !historyPanel.model || !historyPanel.model.id) {
-                window.setTimeout(() => {
```
dannon: Worth noting that this does actually change the behavior here, though I don't recall why we had to have a delay in.

dannon: Ahh.  You can see the problem this was there to address by navigating directly to `http://galaxy_app/custom_builds`.  Central panel loads before the history, causing an issue.

The workaround that was in place isn't great -- the component itself should probably recheck.  Would you like to address that here so there's not a (admittedly minor) regression?

lawrence14701-zz: Hey Dannon, I removed that change, for now, I'll address it in a separate PR. Thanks.

dannon:
@lawrence14701 Thanks!

client/galaxy/scripts/entry/analysis/AnalysisRouter.js
dannon: @lawrence14701 Thanks!

Issue #9252, opened 2020-01-22
Make it more clear when a history search is active


hexylena:
I keep getting confused users saying "where is my data?" and they've accidentally searched for something.

Maybe a warning box (á la empty history) just saying "Your history has been filtered due to a search" or something would help these users?

PR #9706, opened 2020-05-03
Display msg history no items

lawrence14701-zz:
This addresses the following problem:
https://github.com/galaxyproject/galaxy/issues/9252

Previously when the user searched for histories and no results were shown the message displayed was 'no items'. Many users who accidentally searched got confused by this and thought their histories were gone. This pull request addresses the issue by displaying a message that communicates the user has searched and no results were found.


dannon:
I don't think this addresses the concern in the issue -- that is related to the history panel (on the right) itself, when you search it.  This is a useful refinement to the grid search, though, thank you!

PR #9707, opened 2020-05-04
restore tool_util.deps.commands

bernt-matthias:
In https://github.com/galaxyproject/galaxy/pull/9544 `tool_util.deps.commands` was moved to `utils.commands`

since this probaly breaks planemo we restore it such that it just imports from the new location logs a deprecation warning.

Not sure if this is correct, but I did not want to forget about it. 

Is there some mechanism to deal with deprecation, e.g. should be fix a release when this is going to be removed?

### DIFF
--- a/lib/galaxy/tool_util/deps/commands.py
+++ b/lib/galaxy/tool_util/deps/commands.py
@@ -1,0 +1,7 @@
+import logging
+
+import galaxy.util.commands as commands
+
+log = logging.getLogger(__name__)
+
+log.warning("importing galaxy.tool_util.deps.commands is deprecated use galaxy.util.commands instead")
--- a/lib/galaxy/tool_util/deps/commands.py
+++ b/lib/galaxy/tool_util/deps/commands.py
@@ -1,7 +1,6 @@
-import logging
+import warnings
 
 import galaxy.util.commands as commands
 
-log = logging.getLogger(__name__)
 
-log.warning("importing galaxy.tool_util.deps.commands is deprecated use galaxy.util.commands instead")
+warnings.warn("Importing galaxy.tool_util.deps.commands is deprecated, use galaxy.util.commands instead", DeprecationWarning)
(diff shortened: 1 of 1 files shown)


lib/galaxy/tool_util/deps/commands.py
```diff
@@ -0,0 +1,7 @@
+import logging
```
mvdbeek:
```suggestion
import warnings
```

lib/galaxy/tool_util/deps/commands.py
```diff
@@ -0,0 +1,7 @@
...
+
+import galaxy.util.commands as commands
+
+log = logging.getLogger(__name__)
```
mvdbeek:
```suggestion
```

lib/galaxy/tool_util/deps/commands.py
```diff
@@ -0,0 +1,7 @@
...
+
+log.warning("importing galaxy.tool_util.deps.commands is deprecated use galaxy.util.commands instead")
```
mvdbeek:
```suggestion
warnings.warn("Importing galaxy.tool_util.deps.commands is deprecated, use galaxy.util.commands instead", DeprecationWarning)
```

bernt-matthias:
Is the import correct this way? I did not check...

lib/galaxy/tool_util/deps/commands.py
```diff
@@ -0,0 +1,6 @@
+import warnings
+
+import galaxy.util.commands as commands
```
mvdbeek:
```suggestion
from galaxy.util.commands import (  # noqa: F401
    argv_to_str,
    CommandLineException,
    download_command,
    execute,
    redirect_aware_commmunicate,
    redirecting_io,
    shell,
    shell_process,
    which,
)
```

mvdbeek (commented):
Good point, you need to import the functions for equivalence.

mvdbeek (approved):

nsoranzo (approved):

PR #9709, opened 2020-05-04
[20.01] Fix collection mapping in extended metadata mode.

jmchilton:
~I guess this needs tests 😆, lets see what CI says as is though.~ Done.


mvdbeek (approved):
Now that the test is there, is it still WIP @jmchilton ?

PR #9713, opened 2020-05-04
View support for galaxy model, hdca.job_state_summary option

Nerdinacan:
I split this view support out of the main history API PR (#9591) since we are exploring performance issues in that PR that are not related to the view support, and I would like to use database views for other features.

### New Module: galaxy.model.view

This PR includes a new mechanism to register and use/join sql views as ORM entities. 

For each view, a Table is defined for the row-schema of the view, and associated with a selectable which can be either an ORM query or a straight SQL string, representing the contents of the view select itself.

The gotcha is any foreign key declarations (for use in declaring ORM relationships) have to be explicitly defined in the "main" model relationship mapping declarations because the Table in the view definition is not registered on the same MetaData instance (because the view row schema is not really a table), and thus standard ForeignKey() Column parameters aren't going to recognize other tables in the db.

### Added collection job states summary as serialization option in the primary contents query

Added a serialization option to the HDCA result set that gives an overview of the job states working inside that collection by doing an additional eager subquery against the implicit jobs tables.

This extra field is optional and only appears in the query if the 'job_state_summary' key is present in the history contents request URL.

### DIFF
--- a/lib/galaxy/managers/hdcas.py
+++ b/lib/galaxy/managers/hdcas.py
@@ -254,4 +254,11 @@ class HDCASerializer(
                                                      history_id=self.app.security.encode_id(i.history_id),
                                                      id=self.app.security.encode_id(i.id),
                                                      type=self.hdca_manager.model_class.content_type),
+            'job_state_summary'         : self.serialize_job_state_summary
         })
+
+    def serialize_job_state_summary(self, hdca, key, **context):
+        states = hdca.job_state_summary.__dict__.copy()
+        del states['_sa_instance_state']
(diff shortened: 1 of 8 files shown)


jmchilton (approved):

jmchilton:
The unit test failures seem legit? Under get_code_and_test

Nerdinacan:
> The unit test failures seem legit? Under get_code_and_test

It does. I have to tear it apart. I think create/drop under sqlite has some kind of problem.

jmchilton:
Thanks a bunch!

Issue #8238, opened 2019-06-27
Data Libraries - importing datasets as collection - potential UI bug


AjitPS:
Hi, I had a Galaxy UI question about a potential bug `re: collections from data libraries -> imported to history`. 

When I import datasets from a `data library` folder, I can import multiple datasets `as a collection` and tick the option to `hide original elements`. However, my history shows **both** the created collection and all the datasets separately listed too (though I was expecting just a collection in the history interface).

Also, then if I delete the separate datasets from history but keep the collection, my collection `job` immediately gets paused (goes green, but inside the output collection most datasets are `grey` & asking to "fix input dataset"), despite the collection still being in my history?

I guess the datasets are needed as the collection is just a `wrapper` for them but even so, is there a way to hide datasets `automatically` and only have collection visible in history when importing from data library as a collection? 

I saw this on my own Galaxy instance and tested it on **usegalaxy.org** and **GalaxyEU** and found the same issue in all 3 cases. Is this a known Galaxy UI bug? thanks..

AjitPS:
would be good to know if this can be fixed soon in Galaxy...

nsoranzo:
Fixed in #9715.

PR #9715, opened 2020-05-05
[20.01] Make collection import from library import into correct history

mvdbeek:
While Galaxy respected the history to create the collection in, the
dataset would be imported into trans.current_history, ignoring the
requested history

Fixes https://github.com/galaxyproject/galaxy/issues/8238


jmchilton:
Oppps - thanks for the fix!

PR #9718, opened 2020-05-05
Feature/persist rule builder collection in web storage

assuntad23:
Why you made the changes (e.g. references to GitHub issues being fixed).

I added this component so that rule sets can be reused over multiple datasets or collections, without being re-built from scratch.

A description of the implementation of the changes.
- Added a SavedRuleSelector component that saves the rules when the user uploads the dataset/collection
- Added a drop-down button that sorts through local storage and displays the last 10 rule sets that were used
- When a user selects one, it applies that rule set to the current dataset/collection

How to test the changes, if you haven't included specific tests already.

Using the Rule Builder Collection, upload a dataset/collection with a few rules.  After clicking upload, go back into the Rule Builder Collection and drop down the Saved Rules button, you should be able to select the date/time stamp from the collection you just ran.  When you click the rules should appear and be applied to your new dataset/collection.


jmchilton:
Failing selenium test looks unrelated. Thanks @assuntad23! 

Screenshot from Selenium tests on Jenkins:



Probably that + in the icon doesn't belong there - you're not adding a rule you're replacing the whole set right? Also the word rules in the button feels like it doesn't belong given the other buttons are operations on rules but they don't have the word rules. My first instinct would be to make that wrench button at the top - one of the buttons that you can either click or use as a drop down and if it is a drop down use it for restoring old rules?

Also can you look at other Vue components that have been added recently and try to write a VueJS test for this? Sam writes good tests for most of his new components - might want to take a look at something like https://github.com/galaxyproject/galaxy/pull/9574/files for reference.

client/galaxy/scripts/components/RuleBuilder/SavedRulesSelector.vue
```diff
@@ -0,0 +1,74 @@
...
+        };
+    },
+    props: {
+        builder: {
+            required: true,
+        },
+    },
+    computed: {
+        numOfSavedRules: function () {
+            return this.getRules().length;
```
jmchilton: I don't think you should call methods from computed properties.

https://forum.vuejs.org/t/calling-method-in-computed-property/56448

Vue tracks these computed properties and what internal state they depend on. If a method is being called loading an external, non-reactive data source I don't think it can do that. Let me tinker with this and see if I can offer some advice.

jmchilton:
Amazing - thanks so much!

PR #9721, opened 2020-05-05
Various spelling and log message error fixes.

jmchilton:
All stemming from performance monitoring stuff.

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -283,7 +283,7 @@ class BaseAppConfiguration(object):
                 setattr(self, var + '_set', True)
             else:
                 for value in values:
-                    if os.path.exists(root_join(value)):
+                    if os.path.exists(value):
                         path = value
                         break
                 else:
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -275,7 +275,7 @@ class BaseAppConfiguration(object):
             except TypeError:
                 template = "Failed to set configuration variable %s, value %s of wrong type %s"
                 message = template % (var, val, type(val))
-                raise Exception(message)
+                raise ConfigurationError(message)
 
         for var, values in defaults.items():
             if config_kwargs.get(var) is not None:
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -268,19 +268,28 @@ class BaseAppConfiguration(object):
         return os.path.join(_dir, path) if path else None
 
     def _parse_config_file_options(self, defaults, listify_defaults, config_kwargs):
+
+        def root_join(val):
+            try:
+                return os.path.join(self.root, val)
+            except TypeError:
+                template = "Failed to set configuration variable %s, value %s of wrong type %s"
+                message = template % (var, val, type(val))
+                raise Exception(message)
+
         for var, values in defaults.items():
             if config_kwargs.get(var) is not None:
                 path = config_kwargs.get(var)
                 setattr(self, var + '_set', True)
             else:
                 for value in values:
-                    if os.path.exists(os.path.join(self.root, value)):
+                    if os.path.exists(root_join(value)):
                         path = value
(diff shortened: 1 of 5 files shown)


lib/galaxy/config/__init__.py
```diff
@@ -269,18 +269,27 @@ def _in_dir(self, _dir, path):
...
+                    message = template % (var, val, type(val))
+                    raise Exception(message)
+
             if config_kwargs.get(var) is not None:
                 path = config_kwargs.get(var)
                 setattr(self, var + '_set', True)
             else:
                 for value in values:
-                    if os.path.exists(os.path.join(self.root, value)):
+                    if os.path.exists(root_join(value)):
```
nsoranzo: I don't think that's needed here since we are dealing with a default value.

lib/galaxy/config/__init__.py
```diff
@@ -290,15 +299,15 @@ def _parse_config_file_options(self, defaults, listify_defaults, config_kwargs):
             else:
                 for value in values:
                     for path in listify(value):
-                        if not os.path.exists(os.path.join(self.root, path)):
+                        if not os.path.exists(root_join(path)):
```
nsoranzo: Here we are also dealing with default values.

lib/galaxy/config/__init__.py
```diff
@@ -269,18 +269,27 @@ def _in_dir(self, _dir, path):
 
     def _parse_config_file_options(self, defaults, listify_defaults, config_kwargs):
         for var, values in defaults.items():
+
+            def root_join(val):
```
nsoranzo: Isn't defining the function inside the for loop bad for performance?

jmchilton: Yeah - I didn't notice that. Just rebased.

lib/galaxy/config/__init__.py
jmchilton: It is still more readable and consistent right? I prefer it like it is but happy to switch if you insist.

jdavcs: I think we don't need to join here at all: all default values are now absolute paths (see https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L752 ); furthermore, those are different directories, not just root.

lib/galaxy/config/__init__.py
jdavcs: same comment as above

lib/galaxy/config/__init__.py
```diff
@@ -268,19 +268,28 @@ def _in_dir(self, _dir, path):
...
 
     def _parse_config_file_options(self, defaults, listify_defaults, config_kwargs):
+
+        def root_join(val):
+            try:
+                return os.path.join(self.root, val)
+            except TypeError:
+                template = "Failed to set configuration variable %s, value %s of wrong type %s"
+                message = template % (var, val, type(val))
+                raise Exception(message)
```
jdavcs: if we were to raise an exception, I would suggest a `ConfigurationError` like here https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L834

nsoranzo:
```suggestion
                raise ConfigurationError(message)
```

lib/galaxy/config/__init__.py
nsoranzo:
```suggestion
                    if os.path.exists(value):
```

lib/galaxy/config/__init__.py
nsoranzo:
```suggestion
                        if not os.path.exists(path):
```

lib/galaxy/config/__init__.py
jdavcs: +1 for @nsoranzo 's suggestion

nsoranzo (approved):

dannon:
Failing client test is unrelated.

PR #9691, opened 2020-04-30
Optimize setting external ID for jobs.

jmchilton:
Local and DRMAA runner update job_destination params but I think this has been done in enqueue and we don't need rewrite for no reason. I checked out the runners and they don't update the parameters. Resubmission might(?) but I think the resubmit integration tests will suss out whether we need to re-save the job destination parameters in here. Lets hope not.

Last graph in following chart is average times in ms for this code block before (first peaks) and after (third peaks) for 100 jobs with on output each. Second peak is a hybrid of optimizations bundled with an attempt to update the HDCA update_time.

Screen Shot 2020-04-30 at 9 08 21 AM

PR #9571, opened 2020-03-27
WIP: Update dataset states w/one query: initial pass

jdavcs:
~~This doesn't work yet~~. (update: appears to work). The code works if executed from `db_shell`, so the SQL is, most likely, correct. I'm ~~debugging~~  figuring out a way to test this. 
Re #9515.

PR #9722, opened 2020-05-05
Implement HDCA update_time

jmchilton:
This brings in the state update optimization from #9571 - I prefer this to the trigger approach in #9515 in part because the trigger as implemented was broken (there was a syntax error and if it did work it would always update the info fields for instance - though they should not be copied over at the end of the job for instance when they've been set to other values elsewhere). I extended that approach to also be used when ``job.set_final_state`` is called and replaced our call that WorkflowInvocationStep update times with this lower-level SQL approach. 

All of that was largely just to be able to also bring in a change that updates HDCA update_time at the same time these other operations are being executed for #9591 and the new frontend history component.

I can't really show any speed up by doing all of this (#9691 was a much bigger effect - but relatively narrow and doesn't help line up things to alleviate my concerns around #7791). I think there is a small slowdown if anything as a result of all this in parallel. I hit some slow query updates and large spikes in times when testing that has me nervous but nothing consistent or blocking. So I'm not super comfortable with this... but I don't really feel comfortable slowing down anything based on my concerns without evidence and I did remember that we are essentially doing the same problematic synchronized writes around ``WorkflowInvocationStep`` just like HDCAs. So we're not introducing a new order of magnitude problem here. We are just doing it twice.

Perhaps problems around implementing that step update - including this hack - https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/jobs/__init__.py#L1744 are part of my concern with doing this for HDCAs also. Clearly we needed to do something hacky to get that working in the distance past and avoid difficult to debug/track lockups and I didn't understand why 😅.

There are a number of things we can do from this point to get back small amounts of the lost time and optimize things toward more efficient job state changes (for #7791). We can update the update time in the job source objects (for HDCAs and datasets for HDAs) since we join against those in the queries anyway - so why not speed up the writes and eliminate the extra join at that time. I outlined this a bit in https://github.com/galaxyproject/galaxy/pull/9591#issuecomment-620805706. Likewise, we should probably normalize dataset state and info so the jobs info is just used instead of copied around the database. Another important thing we can do is make the update of WorkflowInvocationStep and HDCA update time async with a MQ - get it out of the main job update loop - and throttle the updates for a given object, but that depends on infrastructure not yet ready.

### DIFF
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1372,28 +1372,13 @@ class JobWrapper(HasResourceParameters):
             log.warning("(%s) Ignoring state change from '%s' to '%s' for job "
                         "that is already terminal", job.id, job.state, state)
             return
-        for dataset_assoc in job.output_datasets + job.output_library_datasets:
-            dataset = dataset_assoc.dataset
-            if not job_supplied:
-                self.sa_session.refresh(dataset)
-            state_changed = dataset.raw_set_dataset_state(state)
-            if state_changed:
-                # Arguably a hack to get state changes to appear in the history panel because
-                # the history panel polls on hda.update_time and ignores hda.dataset.update_time.
-                # For those less pragmatic needing a more theoretically sound reason for the update,
-                # perhaps arguments can be made that the entity that is the HDA
-                # really should be described as updated since its effective state did change and its
-                # RESTful representation in the API does change as a result of the above dataset update.
-                dataset.update()
-            if info:
-                dataset.info = info
-            self.sa_session.add(dataset)
         if info:
             job.info = info
         job.set_state(state)
         self.sa_session.add(job)
         if flush:
             self.sa_session.flush()
+        job.update_output_dataset_states()
 
     def get_state(self):
         job = self.get_job()
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1376,9 +1376,9 @@ class JobWrapper(HasResourceParameters):
             job.info = info
         job.set_state(state)
         self.sa_session.add(job)
+        job.update_output_states()
         if flush:
             self.sa_session.flush()
-        job.update_output_dataset_states()
 
     def get_state(self):
         job = self.get_job()
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1714,14 +1714,13 @@ class JobWrapper(HasResourceParameters):
         self.tool.call_hook('exec_after_process', self.app, inp_data=inp_data,
                             out_data=out_data, param_dict=param_dict,
                             tool=self.tool, stdout=job.stdout, stderr=job.stderr)
-        job.command_line = unicodify(self.command_line)
 
         collected_bytes = 0
         # Once datasets are collected, set the total dataset size (includes extra files)
         for dataset_assoc in job.output_datasets:
             if not dataset_assoc.dataset.dataset.purged:
-                dataset_assoc.dataset.dataset.set_total_size()
-                collected_bytes += dataset_assoc.dataset.dataset.get_total_size()
+                # don't call get_total_size - forces a flush we don't want in here.
+                collected_bytes += dataset_assoc.dataset.dataset.set_total_size()
 
         if job.user:
             job.user.adjust_total_disk_usage(collected_bytes)
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1719,8 +1719,7 @@ class JobWrapper(HasResourceParameters):
         # Once datasets are collected, set the total dataset size (includes extra files)
         for dataset_assoc in job.output_datasets:
             if not dataset_assoc.dataset.dataset.purged:
-                # don't call get_total_size - forces a flush we don't want in here.
-                collected_bytes += dataset_assoc.dataset.dataset.set_total_size()
+                collected_bytes += dataset_assoc.dataset.set_total_size()
 
         if job.user:
             job.user.adjust_total_disk_usage(collected_bytes)
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -2400,6 +2400,7 @@ class Dataset(StorableObject, RepresentById):
             if self.object_store.exists(self, extra_dir=rel_path, dir_only=True):
                 for root, dirs, files in os.walk(self.extra_files_path):
                     self.total_size += sum([os.path.getsize(os.path.join(root, file)) for file in files if os.path.exists(os.path.join(root, file))])
+        return self.total_size
 
     def has_data(self):
         """Detects whether there is any data"""
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -1150,8 +1150,24 @@ class Job(JobLike, UsesCreateAndUpdateTime, Dictifiable, RepresentById):
         for dataset_assoc in self.output_datasets:
             return dataset_assoc.dataset.tool_version
 
-    def update_output_dataset_states(self):
+    def update_output_states(self):
         sql = '''
+            UPDATE history_dataset_collection_association
+            SET update_time = :update_time
+            WHERE id in (
+                SELECT hdca.id
+                FROM history_dataset_collection_association hdca
+                INNER JOIN implicit_collection_jobs icj
+                    on icj.id = hdca.implicit_collection_jobs_id
+                INNER JOIN implicit_collection_jobs_job_association icjja
+                    on icj.id = icjja.implicit_collection_jobs_id
+                WHERE icjja.job_id = :job_id
+                UNION
+                SELECT hdca2.id
+                FROM history_dataset_collection_association hdca2
+                WHERE hdca2.job_id = :job_id
+            );
+
             UPDATE dataset
             SET
                 state = :state,
@@ -1200,7 +1216,7 @@ class Job(JobLike, UsesCreateAndUpdateTime, Dictifiable, RepresentById):
             'update_time': galaxy.model.orm.now.now()
         }
         sa_session.execute(sql, params)
-        sa_session.flush()
+        # sa_session.flush()
 
 
 class Task(JobLike, RepresentById):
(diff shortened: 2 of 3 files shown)


Nerdinacan:
Sold!

As long as the date gets updated I'm happy. Want me to close out my trigger-based PR? (#9591) Since the rest of that PR has migrated into separate other PRs?

jmchilton:
Sounds good to me @Nerdinacan - thanks for patience on this issue.

lib/galaxy/model/__init__.py
```diff
@@ -1142,6 +1172,75 @@ def command_version(self):
...
+        params = {
+            'job_id': self.id,
+            'state': self.state,
+            'info': self.info,
+            'update_time': galaxy.model.orm.now.now()
+        }
+        for statement in statements:
+            sa_session.execute(statement, params)
+        # sa_session.flush()
+
```
mvdbeek:
```suggestion

```

mvdbeek (approved):

lib/galaxy/model/__init__.py
```diff
@@ -1119,8 +1119,38 @@ def to_dict(self, view='collection', system_details=False):
 
     def set_final_state(self, final_state):
         self.set_state(final_state)
-        if self.workflow_invocation_step:
-            self.workflow_invocation_step.update()
```
mvdbeek, line 1131: `WorkflowInvocationStep.update` and `WorkflowInvocation.update` is now unused, so we should drop that, right ?

lib/galaxy/jobs/__init__.py
```diff
@@ -1721,14 +1706,13 @@ def fail():
...
-        job.command_line = unicodify(self.command_line)
 
         collected_bytes = 0
         # Once datasets are collected, set the total dataset size (includes extra files)
         for dataset_assoc in job.output_datasets:
             if not dataset_assoc.dataset.dataset.purged:
-                dataset_assoc.dataset.dataset.set_total_size()
-                collected_bytes += dataset_assoc.dataset.dataset.get_total_size()
+                # don't call get_total_size - forces a flush we don't want in here.
+                collected_bytes += dataset_assoc.dataset.dataset.set_total_size()
```
nsoranzo:
```suggestion
                collected_bytes += dataset_assoc.dataset.set_total_size()
```

lib/galaxy/jobs/__init__.py
```diff
@@ -1721,14 +1706,13 @@ def fail():
...
                             tool=self.tool, stdout=job.stdout, stderr=job.stderr)
-        job.command_line = unicodify(self.command_line)
 
         collected_bytes = 0
         # Once datasets are collected, set the total dataset size (includes extra files)
         for dataset_assoc in job.output_datasets:
             if not dataset_assoc.dataset.dataset.purged:
-                dataset_assoc.dataset.dataset.set_total_size()
-                collected_bytes += dataset_assoc.dataset.dataset.get_total_size()
+                # don't call get_total_size - forces a flush we don't want in here.
```
nsoranzo: It's quite confusing that `set_total_size` doesn't save to the database but `get_total_size` does...
Maybe add an explicit `, flush=True` parameter to both?

lib/galaxy/jobs/__init__.py
mvdbeek: Both versions are correct. job.output_datasets are `JobToOutputDatasetAssociation` objects. So `dataset_assoc.dataset` is the HDA and `dataset_assoc.dataset.dataset` is a `Dataset`. I guess it saves one function call to use `dataset_assoc.dataset.dataset.set_total_size()` ?

nsoranzo: My reasoning is that the HDA is correct level of abstraction here, unless there is an important performance gain in skipping a method call. This may become a bug if one day the HDA method start doing more than just call the `Dataset` method.

lib/galaxy/model/__init__.py
```diff
@@ -1142,6 +1172,75 @@ def command_version(self):
...
+                state = :state,
+                update_time = :update_time
+            WHERE id IN (
+                SELECT ldda.dataset_id FROM library_dataset_dataset_association ldda
+                INNER JOIN job_to_output_library_dataset jtold
+                ON jtold.ldda_id = ldda.id AND jtold.job_id = :job_id
+            );
+        ''', '''
+            UPDATE history_dataset_association
+            SET
```
Nerdinacan, line 1224: If performance is your primary concern you might consider updating from joins rather than with where-in subqueries. Traditionally that has been a more performant way to do this in many databases, though I do find it harder to understand visually, and I think the way you've written it is easier to understand for future coders.

If you choose to go that route, because of the variety of sql syntaxes for an update-from, this is a reasonably good place to leverage the orm's query builder funcs.

lib/galaxy/jobs/__init__.py
jmchilton: That is fair - I mean this is performant critical code to my mind but the extra python function call is the least of our problems so I'll use the better abstraction.

lib/galaxy/jobs/__init__.py
jmchilton: Can I just drop the comment then? It isn't confusing that set_size returns the size right - the confusion is that get size initializes the data but that isn't relevant if the comment weren't here?

lib/galaxy/jobs/__init__.py
mvdbeek: But the size refers to the dataset, not the history dataset relation ? In my mind we'll have to separate out extra files at one point in the future and relate that back to the dataset, so that we can have HDAs of different datatypes sharing a dataset. I think that's a shortcoming of our current model and pretending that HDAs have a size seems like a step away from that (Admittedly that was not my initial argument :D)

jmchilton: I don't think I agree with that - at least at first glance. The dataset is physical collection of files. I think the right way to handle different composite files between HDAs would be to duplicate the dataset with different contents and find some other mechanism to de-duplicate actual files? A concept below Dataset maybe?

jmchilton: I guess put another way Metadata files and extra files made very different choices and should be converged in some fashion - but toward datasets would make more sense to me and my view of things. I could be wrong though.

mvdbeek: But even if we had a concept below Dataset the file size wouldn't belong on the HDA but the Dataset? It seems counterintuitive to me for the relation to have a size, especially since the composite file content is also part of the Dataset. Effectively we get and set the size of the `Dataset` here, so I'm fine with merging it as is, but I feel like the current form is clearer both to what actually happens and how we may want to go forward ?

mvdbeek: I didn't see your second comment before posting my last one. I guess we can revisit this when someone starts working on it, let's get this in. MetadataFile being part of the HDA seems to make more sense to me, I didn't think about that. In fact making extra files a first class object that relates to HDAs would probably also neatly close that gap I am worried about in our model.

lib/galaxy/jobs/__init__.py
jdavcs: +1 on @nsoranzo's suggesion. I think, it *is* unexpected that `get_...` may update the database. (just a lazy load with calling the db if value not set would be fine, but to me this feels different).
Maybe not `flush=True`, but `set_if_none=True` or `calculate_if_none`? That way the argument explicitly says *what* will be done (vs. `flush`: what updates will be flushed?)
~~(BTW, `set_size` does *not* return anything, unless I'm looking in the wrong place).~~ I see this now.

lib/galaxy/model/__init__.py
```diff
@@ -5307,9 +5405,6 @@ class WorkflowInvocationStep(Dictifiable, RepresentById):
         # FAILED='failed',  TODO: implement and expose
     )
 
-    def update(self):
```
jdavcs, line 5310: This breaks this unit test: https://github.com/galaxyproject/galaxy/blob/dev/test/unit/test_galaxy_mapping.py#L550
I assume this is in response to this comment: https://github.com/galaxyproject/galaxy/pull/9722#discussion_r428079456 ? Is the test outdated? Or do we need `update()` after all? @mvdbeek ?

mvdbeek: Yeah, that unit test looks like it should be updated. ~~Probably just need to drop that line you referenced.~~

lib/galaxy/model/__init__.py
jdavcs: `update from` is not supported in sqlite, and the code needs to support both postgres and sqlite. Checking the current db engine and executing the appropriate query is easy (here's one example https://github.com/galaxyproject/galaxy/pull/9515/files#diff-5da75d9743ecb1b3e59a6ba6527ccb16R63). My concern is that this module is already 6000+ lines long. It's bad enough it contains raw SQL.. Unless we *know* that using a join for postgres updates does indeed address a bottleneck, then sure, let's have 2 versions and execute them conditionally. Other than that, I'd keep it simple.

lib/galaxy/model/__init__.py
jdavcs: It seems to me this could be moved into a separate PR.  Updating the `WorkflowInvocationStep` class (and related tests) is a non-essential tangent here. I fear it may delay things. My vote is to skip it, get the PR in, and then address this. @jmchilton?

lib/galaxy/jobs/__init__.py
nsoranzo: @jmchilton I had nothing against that comment, it was useful. What I have always found confusing is the implementation of both `get_total_size` and `set_total_size` methods. What I am proposing is:

```python
    def get_total_size(self, flush=True):
        if self.total_size is not None:
            return self.total_size
        return self.set_total_size(flush=flush)

    def set_total_size(self, flush=True):
        if self.file_size is None:
            self.set_size()
        self.total_size = self.file_size or 0
        rel_path = self._extra_files_rel_path
        if rel_path is not None:
            if self.object_store.exists(self, extra_dir=rel_path, dir_only=True):
                for root, dirs, files in os.walk(self.extra_files_path):
                    self.total_size += sum([os.path.getsize(os.path.join(root, file)) for file in files if os.path.exists(os.path.join(root, file))])
        if flush:
            db_session = object_session(self)
            db_session.flush()
        return self.total_size
```

That's very probably material for another PR, but what do you think of this?

@ic4f Regarding the parameter name, `flush` is used in all Galaxy performance-sensitive code to mean if a modified object should be flushed to the database or the caller is responsible for that, so I think its meaning would be clear in this context.

jdavcs: @nsoranzo thanks for clarifying, this makes sense. +1

jdavcs (approved):

ic4f:
I believe this is ready for merging?

PR #9725, opened 2020-05-06
Upgrade to the newer nodejs lts (v10 -> v12; 12.16.3)

dannon:
Wanted to do this early in the release cycle, along with client dependency updates.  Trying this first, separately, since I have a hunch we may need to manually clean out the nodeenv install.

Seeing if this breaks anything obvious, first, and then it'd be good to hear if folks are able to run this on their machines.


dannon:
Actually seems pretty okay?  It'd be good if other people on other OSes could verify this is fairly smooth and that we don't have to do any manual cleaning of old versions, etc.

I'm going to build on this with some other changes, including dropping node-sass and all its wonky platform bindings.

jmchilton (approved):

jmchilton:
Client watching has stopped working for me:

```
/Users/john/workspace/galaxy/client/node_modules/chokidar/lib/fsevents-handler.js:28
  return (new fsevents(path)).on('fsevent', callback).start();
          ^

TypeError: fsevents is not a constructor
```

Per https://stackoverflow.com/questions/56057325/how-to-fix-typeerror-fsevents-is-not-a-constructor-react-error I tried reinstalling node modules and noticed:

```
warning Error running install script for optional dependency: "/Users/john/workspace/galaxy/client/node_modules/ava/node_modules/fsevents, /Users/john/workspace/galaxy/client/node_modules/chokidar/node_modules/fsevents: Command failed.
```

Is this just me? Maybe I need an OS update.

dannon:
@jmchilton It's not just you; I just noticed client watching was busted for me this morning, though I know it was working before :/  I've been fighting with it for a bit now trying to figure out a fix.

PR #9726, opened 2020-05-06
[20.05] Fix custos credential fetching in python3

dannon:
b64encode needs bytes here, and other fixes

### DIFF
--- a/lib/galaxy/authnz/custos_authnz.py
+++ b/lib/galaxy/authnz/custos_authnz.py
@@ -230,7 +230,7 @@ class CustosAuthnz(IdentityProvider):
         # Set custos endpoints
         clientIdAndSec = self.config['client_id'] + ":" + self.config['client_secret']
         eps = requests.get(self.config['url'],
-                           headers={"Authorization": "Basic %s" % base64.b64encode(clientIdAndSec)},
+                           headers={"Authorization": "Basic %s" % base64.b64encode(util.smart_str(clientIdAndSec))},
                            verify=False, params={'client_id': self.config['client_id']})
         endpoints = eps.json()
         self.config['authorization_endpoint'] = endpoints['authorization_endpoint']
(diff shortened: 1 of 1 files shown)


lib/galaxy/authnz/custos_authnz.py
```diff
@@ -230,7 +230,7 @@ def _load_config_for_custos(self):
         # Set custos endpoints
         clientIdAndSec = self.config['client_id'] + ":" + self.config['client_secret']
         eps = requests.get(self.config['url'],
-                           headers={"Authorization": "Basic %s" % base64.b64encode(clientIdAndSec)},
+                           headers={"Authorization": "Basic %s" % base64.b64encode(util.smart_str(clientIdAndSec))},
```
nsoranzo:
```suggestion
                           headers={"Authorization": "Basic %s" % util.unicodify(base64.b64encode(util.smart_str(clientIdAndSec)))},
```

dannon: @nsoranzo (nevermind, I think I might see what's up)

nsoranzo: On Python3 always, otherwise this would result in:
`"Basic b'bWRma2Jua2RmYmRm'"` instead of `'Basic bWRma2Jua2RmYmRm'`

dannon: ^ yep, I saw it once I flipped python versions and tested.  Thanks!  (and there are several other instances that need to be fixed, too)

nsoranzo (approved):

nsoranzo:
Failed py35-unit test seems legit?

dannon:
@nsoranzo Yes, it's legit.  It's popping up because I actually fixed the configuration loading with https://github.com/galaxyproject/galaxy/pull/9726/commits/a57959b80987694b6922650df3d89852fe8aefc2, I believe.  I'll update the tests.

nsoranzo (approved):

PR #9728, opened 2020-05-06
Move containers test from selenium to selenium integration.

davebx:
This properly tests the containers section from #9618 instead of skipping the test entirely if running on jenkins. Also properly completes the Admin Interface/Container grid checkbox on #3241


jmchilton:
Awesome - I'm so glad this worked out! Thanks so much @davebx - really cool stuff.

PR #9729, opened 2020-05-06
Force container image names to lower case to satisfy Docker

nuwang:
This is to prevent:
```
InvalidImageName: Failed to apply default image tag "quay.io/biocontainers/Mimodd:0.1.9--py36_0": couldn't parse image reference "quay.io/biocontainers/MiModD:0.1.9--py36_0": invalid reference format: repository name must be lowercase
```
as Docker does not allow mixed case image names.


mvdbeek (approved):
Nice, good find!

nuwang:
@mvdbeek Closing in favour of 20.01 patch: https://github.com/galaxyproject/galaxy/pull/9733

PR #9730, opened 2020-05-06
[20.05] oidc-based authnz provider overhaul; breaking changes for keycloak

dannon:
Split out of #9726, since it's related to the specification of authnz providers and not simply the python3 custos fixes, and because it presents a larger issue that isn't solved yet.  WIP.

(this is expected to fail tests currently)

### DIFF
--- a/lib/galaxy/authnz/custos_authnz.py
+++ b/lib/galaxy/authnz/custos_authnz.py
@@ -219,8 +219,7 @@ class CustosAuthnz(IdentityProvider):
         self.config['userinfo_endpoint'] = "https://cilogon.org/oauth2/userinfo"
 
     def _load_config_for_custos(self):
-        well_known_uri = self._get_well_known_uri_from_url(self.config['provider'])
-        self.config['well_known_oidc_config_uri'] = well_known_uri
+        self.config['well_known_oidc_config_uri'] = self._get_well_known_uri_from_url(self.config['provider'])
         # Set custos endpoints
         clientIdAndSec = self.config['client_id'] + ":" + self.config['client_secret']
         eps = requests.get(self.config['url'],
--- a/lib/galaxy/authnz/custos_authnz.py
+++ b/lib/galaxy/authnz/custos_authnz.py
@@ -19,6 +19,7 @@ from ..authnz import IdentityProvider
 log = logging.getLogger(__name__)
 STATE_COOKIE_NAME = 'custos-state'
 NONCE_COOKIE_NAME = 'custos-nonce'
+KEYCLOAK_BACKENDS = {'custos', 'cilogon', 'keycloak'}
 
 
 class CustosAuthnz(IdentityProvider):
--- a/lib/galaxy/authnz/custos_authnz.py
+++ b/lib/galaxy/authnz/custos_authnz.py
@@ -32,28 +32,28 @@ class CustosAuthnz(IdentityProvider):
         if oidc_backend_config.get('credential_url'):
             # Keycloak client secret used to get token for custos
             self.config['credential_url'] = oidc_backend_config['credential_url']
-            self._get_custos_credentials()
+            if provider == 'custos':
+                self._get_custos_credentials()
         self.config['redirect_uri'] = oidc_backend_config['redirect_uri']
         self.config['ca_bundle'] = oidc_backend_config.get('ca_bundle', None)
         self.config['extra_params'] = {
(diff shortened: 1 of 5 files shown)


mvdbeek:
@dannon given the draft and bug labels I'll push this to 20.09

dannon:
I'm out this week but this should probably be resolved with 20.05

almahmoud:
I rebased against latest `release_20.05`, added keycloak as a separate idp name, and tested that it still works with GVL. This is a breaking change (discussed with @afgane, @dannon, and @juleengraham) in terms of the `oidc_backends_config` for Keycloak and Custos. For Keycloak, `<url>` will also include the realm, and `<realm>` is no more. Example working Keycloak config:
```
<?xml version="1.0"?>
<OIDC>
    <provider name="keycloak">
        <url>https://demoiraclass.cloudve.org/auth/realms/master/</url>
        <client_id>local-am-gxy</client_id>
        <client_secret>2079a5ac-0e5b-4735-85b1-2fdc76e6dba5</client_secret>
        <redirect_uri>http://localhost:8080/authnz/keycloak/callback</redirect_uri>
        <enable_idp_logout>true</enable_idp_logout>
    </provider>
</OIDC>
```
For Custos, the `<url>` should not be the full well-known as was before, but should stop after the version and will construct the well-known from it, same as keycloak.
CILogon doesn't change afaik.

lib/galaxy/authnz/custos_authnz.py
```diff
@@ -23,36 +23,37 @@
...
-            self.config['end_session_endpoint'] = well_known_oidc_config['end_session_endpoint']
-        elif provider == 'cilogon':
+        # if 'well_known_oidc_config_uri' in oidc_backend_config:
+        #     self.config['well_known_oidc_config_uri'] = oidc_backend_config['well_known_oidc_config_uri']
+        #     well_known_oidc_config = self._load_well_known_oidc_config(
+        #         self.config['well_known_oidc_config_uri'])
+        #     self.config['authorization_endpoint'] = well_known_oidc_config['authorization_endpoint']
+        #     self.config['token_endpoint'] = well_known_oidc_config['token_endpoint']
+        #     self.config['userinfo_endpoint'] = well_known_oidc_config['userinfo_endpoint']
+        #     self.config['end_session_endpoint'] = well_known_oidc_config['end_session_endpoint']
```
afgane: Why not remove these commented lines?

lib/galaxy/authnz/custos_authnz.py
```diff
@@ -226,16 +218,26 @@ def _load_config_for_cilogon(self):
...
+        self.config['userinfo_endpoint'] = well_known_oidc_config['userinfo_endpoint']
+        self.config['end_session_endpoint'] = well_known_oidc_config['end_session_endpoint']
+
+    def _load_config_for_keycloak(self):
+        self.config['well_known_oidc_config_uri'] = self._get_well_known_uri_from_url(self.config['provider'])
+        well_known_oidc_config = self._load_well_known_oidc_config(self.config['well_known_oidc_config_uri'])
+        self.config['authorization_endpoint'] = well_known_oidc_config['authorization_endpoint']
+        self.config['token_endpoint'] = well_known_oidc_config['token_endpoint']
+        self.config['userinfo_endpoint'] = well_known_oidc_config['userinfo_endpoint']
+        self.config['end_session_endpoint'] = well_known_oidc_config['end_session_endpoint']
```
mvdbeek: Since those line are identical to those in `_load_config_for_custos ` can you push them into something like `_setup_endpoint_config` ?

lib/galaxy/authnz/managers.py
```diff
@@ -113,7 +113,7 @@ def _parse_oidc_backends_config(self, config_file):
                     self.oidc_backends_config[idp] = self._parse_idp_config(child)
                     self.oidc_backends_implementation[idp] = 'psa'
                     self.app.config.oidc[idp] = {'icon': self._get_idp_icon(idp)}
-                elif idp == 'custos' or idp == 'cilogon':
+                elif idp == 'custos' or idp == 'cilogon' or idp == 'keycloak':
```
mvdbeek: Can you add something like `CUSTOS_BACKENDS = {'custos', 'cilogin', 'keycloak'}` and then use this here and on line 177 and line 273

mvdbeek (approved):
Thanks @dannon!

dannon:
Lacking integration tests (which we should have for this eventually), we've manually tested both keycloak and custos and it's all working happily now.

PR #9731, opened 2020-05-06
[20.01] Fix gff3.gz and gff3.bz2 datatypes

mvdbeek:
We just added the auto_compressed attributes, but we missed actually making sure the code is ok in https://github.com/galaxyproject/galaxy/pull/8967.


jmchilton:
Thanks!

PR #9734, opened 2020-05-06
[20.01] Use right extension in bz2_to_uncompressed.xml

mvdbeek:


nsoranzo (approved):

jmchilton:
Opps - that looks like classic John 😅 but I didn't do a blame. Anyway, thanks for the fix!

PR #9735, opened 2020-05-06
Update to IT comment

astrovsky01:


dannon:
To elaborate on what happens here; using `host.docker.internal` will override /etc/hosts within the container, preventing the actual host loop from working at all.  It was interesting to find.

PR #9736, opened 2020-05-07
[20.01] Allow anonymous downloading of public workflows

mvdbeek:
This allows downloading public workflows without API key.
We already allow this in the workflow controller, and permissions are checked in the `get_stored_accessible_workflow` method in the WorkflowsManager class.

### DIFF
--- a/lib/galaxy_test/api/test_workflows.py
+++ b/lib/galaxy_test/api/test_workflows.py
@@ -273,7 +273,7 @@ class WorkflowsApiTestCase(BaseWorkflowsApiTestCase):
         self._assert_status_code_is(show_response, 400)
 
     def test_cannot_show_private_workflow(self):
-        workflow_id = self.workflow_populator.simple_workflow("test_not_importportable")
+        workflow_id = self.workflow_populator.simple_workflow("test_not_importable")
         with self._different_user():
             show_response = self._get("workflows/%s" % workflow_id)
             self._assert_status_code_is(show_response, 403)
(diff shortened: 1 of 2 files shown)


hexylena:
@bebatut probably interesting for you w/r/t ptdk

nsoranzo:
Should we add an API test checking that downloading a public workflow works?
Also, do we already have a test checking that a non-public workflow cannot be downloaded by an unauthorised user?

mvdbeek:
Excellent advise as usual @nsoranzo, of course I picked the wrong API decorator, this needed to be `expose_api_raw_anonymous_and_sessionless` and not `expose_api_raw_anonymous`.

lib/galaxy_test/api/test_workflows.py
```diff
@@ -281,6 +282,22 @@ def test_cannot_show_private_workflow(self):
             workflows_url = self._api_url("workflows/%s" % workflow_id)
             assert get(workflows_url).status_code == 403
 
+    def test_cannot_download_private_workflow(self):
+        workflow_id = self.workflow_populator.simple_workflow("test_not_importportable")
```
nsoranzo:
```suggestion
        workflow_id = self.workflow_populator.simple_workflow("test_not_importable")
```

lib/galaxy_test/api/test_workflows.py
```diff
@@ -281,6 +282,22 @@ def test_cannot_show_private_workflow(self):
...
+        with self._different_user():
+            with pytest.raises(AssertionError) as excinfo:
+                self._download_workflow(workflow_id)
+            assert '403' in str(excinfo.value)
+        workflows_url = self._api_url("workflows/%s/download" % workflow_id)
+        assert get(workflows_url).status_code == 403
+
+    def test_anon_can_download_public_workflow(self):
+        workflow_id = self.workflow_populator.simple_workflow("test_not_importportable", publish=True)
```
nsoranzo:
```suggestion
        workflow_id = self.workflow_populator.simple_workflow("test_published", publish=True)
```

nsoranzo (commented):
Thanks for adding the tests!

nsoranzo (approved):

PR #9737, opened 2020-05-07
Swap out node-sass for the new standard sass compiler

dannon:
This swaps the older node-sass (ruby/libsass) for dart-sass (compiles to js) for compatibility across node versions, platforms, etc. (node-gyp issues...).  dart-sass is the reference implementation now, and it's the way to go for sass compilation.

The main reason for this, though, is to avoid all the various issues you run into with node-sass's approach binding to libsass requiring constant rebuilds and manual intervention when switching node versions, updating anything, etc.


jmchilton:
Awesome - thanks @dannon!

PR #9283, opened 2020-01-30
Swap from bootstrap-tour to tourist

dannon:
This should also fix https://github.com/galaxyproject/galaxy/issues/6944

I still need to fix up a couple style issues, so, WIP.

PR #9739, opened 2020-05-07
Early 20.09 full client dependency update.

dannon:
~Builds on #9737 after a rebase on #9725, just want to get tests running here.  (and I'll rebase it again once that's in...)~  It's a clean branch against dev now.

Everything has been updated except the following major version bumps:

- `handsontable` and `@handsontable/vue` This one should be done by someone with more familiarity with the rule builder.  A problem here is that handsontable has changed their licensing from MIT, and we may need to look for alternatives in the future. 
- `jquery` and `jquery-migrate` This will be done separately in a jquery-specific update.
- `d3` maybe @guerler can look at this one?
- `bootstrap-tour` I have a separate branch replacing this with bootstrap-tourist https://github.com/galaxyproject/galaxy/pull/9283
- `latex-to-unicode-converter` This disappears in #9757


dannon:
`FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory`  is why the circleci tests are failing; working on a fix.

client/galaxy/scripts/app/singleton.js
```diff
@@ -6,8 +6,10 @@ import addLogging from "utils/add-logging";
 import { GalaxyApp } from "./galaxy";
 import { serverPath } from "utils/serverPath";
 
-export function setGalaxyInstance(factory) {
-    console.warn("setGalaxyInstance", serverPath());
+export function setGalaxyInstance(factory, testing = false) {
```
Nerdinacan: Maybe use the client side config system to detect whether testing or not?


```javascript
import config from "config";

export function setGalaxyInstance(factory) {
   let testing = config.someTestingFlagYouMake;
   ...
}

// client/galaxy/config/testing.js
export {
    name: "unit testing configs",
    debug: false,
    yourTestingFlag: true
    ...
}
```

dannon: @Nerdinacan Feels bad to admit I forgot we added that; good idea.

Nerdinacan: I probably would have forgotten too except I'm staring right at it for the rxdb configs.

PR #9740, opened 2020-05-07
[20.05] Sort Workflow list by last updated time instead of created time

guerler:
The workflow list contains a column which for a given workflow, displays the time passed since the workflow was created. This PR replaces this value and displays the time passed since a workflow was updated instead.


dannon (approved):
Looks good, updated is likely more useful.  It looks like we actually already originally sorted by 'updated time', but there was just no display column, and, upon sorting, there was no way to get back to the original sort order short of a full page reload.

PR #9742, opened 2020-05-08
doc: dynamic destinations better set the id in JobDestination

bernt-matthias:
otherwise it looks odd in the logs (None) and confuses people like me ...


jmchilton:
Thanks!

Issue #9741, opened 2020-05-08
Tool Search improvements


juleengraham:
- [ ] Tool Search doesn't return tool when it should be an exact match e.g. "cat1" returns no results but the Concatenate Datasets tool_id is "cat1"
- [x] Tools not showing up for partial strings e.g. "ilter" doesn't return any filter tools
- [x] Fuzzy tool search for potential typos and misspellings

martenson:
For fuzzy search check https://github.com/galaxyproject/galaxy/pull/3356

martenson:
xref https://github.com/galaxyproject/galaxy/issues/2272

wm75:
Not sure whether to report 20.05 search issues here, or in the old thread.
Here's an old one that still does not work correctly in 20.05:
https://github.com/galaxyproject/galaxy/issues/2272#issuecomment-573080510

Type any search term ending in a trailing whitespace and the search will return what appears to me like arbitrary results.
Solution to this might be as easy as stripping trailing whitespace from the query before searching?

bgruening:
I think what @wm75 is seeing is that the tool search or the index-build step is not recognizing `hidden="true" ` tools.

EU has toolshed.g2.bx.psu.edu/repos/iuc/snpeff/snpEff/4.3r.1 installed, which is hidden. toolshed.g2.bx.psu.edu/repos/iuc/snpeff/snpEff/4.3+T.galaxy1 is not hidden. The search returns `4.3r.1` which is hidden in the tool panel and will not be shown.

wm75:
> I think what @wm75 is seeing is that the tool search or the index-build step is not recognizing `hidden="true" ` tools.

Yes, that's another observation, but it's independent from the whitespace issue above.

mvdbeek:
Since I've been playing with this today:

> * Tools not showing up for partial strings e.g. "ilter" doesn't return any filter tools
> * [ ]  Fuzzy tool search for potential typos and misspellings

that works when using `tool_enable_ngram_search`, but not without. Maybe we should make this the default ?

martenson:
I think there are parts here that still need addressing.

mvdbeek:
Ah, it shouldn't have closed the whole issue, I just referenced https://github.com/galaxyproject/galaxy/issues/9741#issuecomment-667956458. That said, I'm gonna tick off the last 2 boxes, since that seems to work.

PR #9744, opened 2020-05-08
Allow tool search on tool id and wildcard before search string

juleengraham:


juleengraham:
Address parts of isuue #9741

lib/galaxy/tools/search/__init__.py
```diff
@@ -162,7 +163,7 @@ def search(self, q, tool_name_boost, tool_section_boost,
             return rval
         else:
             # Use asterisk Whoosh wildcard so e.g. 'bow' easily matches 'bowtie'
-            parsed_query = self.parser.parse(cleaned_query + '*')
+            parsed_query = self.parser.parse('*' + cleaned_query + '*')
```
jmchilton, line 191: Exciting - I always search by this and this never works. Thanks so much.

Maybe use the short ID (it is unfortunately called ``tool.old_id``, not a great name) so you don't have to modify this? I would think you'd generally want to search on the short id anyway. 

Update: I see based on the title you were trying to do both things, never mind this comment about the prefix then. I still think you might want to search by old_id instead of id though.

martenson: I don't remember exactly why I removed this asterisk, but I think it was a combination of performance, false hits, and ngram support. See https://github.com/galaxyproject/galaxy/commit/d0eb6aba558e99f8461ae0677d6a54b545c0dc92 and https://github.com/galaxyproject/galaxy/pull/8789 for the change

> do not prepend a leading * wildcard for the individual terms, this should make things much faster on a big index, it will also yields less, but better results

lib/galaxy/webapps/galaxy/api/tools.py
```diff
@@ -360,6 +360,7 @@ def _search(self, q):
         :return type: dict
         """
         tool_name_boost = self.app.config.get('tool_name_boost', 9)
+        tool_id_boost = self.app.config.get('tool_id_boost', 9)
```
martenson, line 363: We should add the config option with its default value to the sample file.

mvdbeek: @juleengraham could you do this please ? I think this is ready to be merged otherwise.

lib/galaxy/tools/search/__init__.py
```diff
@@ -170,17 +172,23 @@ def search(self, q, tool_name_boost, tool_section_boost,
...
         cleaned_query = q.lower()
         if tool_enable_ngram_search is True:
             rval = self._search_ngrams(cleaned_query, tool_ngram_minsize, tool_ngram_maxsize, tool_search_limit)
             return rval
         else:
             cleaned_query = ' '.join(token.text for token in self.rex(cleaned_query))
             # Use asterisk Whoosh wildcard so e.g. 'bow' easily matches 'bowtie'
-            parsed_query = self.parser.parse(cleaned_query + '*')
+            #parsed_query = self.parser.parse('*' + cleaned_query + '*' + '~2')
+            parsed_query = self.parser.parse('*' + cleaned_query + '*' + ' ' + cleaned_query + '~2')
```
mvdbeek: Why are there both queries in there ? Seems like you'd only want the commented one above ?
I'd really prefer not to switch to `FuzzyTermPlugin`. We've switched the default to `tool_enable_ngram_search`, which handles fuzzy searches as well.
If that's ok with you I'd suggest keeping that as a follow-up if we discover problems with the ngram search ? Or, if you want to put in the extra work to make it a config option to use the `FuzzyTermPlugin`

mvdbeek:
Moving the milestone, I don't think I reviewed the FuzzyTerm changes here.

juleengraham:
@mvdbeek should I just remove the fuzzy search changes? i think the only fuzzy search-related bit is the "~2" (and the related imports)

mvdbeek:
That would be great, there's also a stray print statement.

PR #9746, opened 2020-05-08
linkify Data Libraries description and bugfix

OlegZharkov:
This PR will linkify elements in Library description. It will create an ```<a>``` element, that will open link a new tab on click
image

Additionally, don't run ```$('#center [data-toggle="tooltip"]').tooltip({trigger: "hover"}```, if title is there. I guess, we don't really need an additional title on hover. I assume it was the reason, that caused tooltip glitch.

### DIFF
--- a/client/galaxy/scripts/mvc/library/library-librarylist-view.js
+++ b/client/galaxy/scripts/mvc/library/library-librarylist-view.js
@@ -5,6 +5,9 @@ import { getGalaxyInstance } from "app";
 import { Toast } from "ui/toast";
 import mod_library_model from "mvc/library/library-model";
 import mod_library_libraryrow_view from "mvc/library/library-libraryrow-view";
+import linkifyJq from 'linkifyjs/jquery';
+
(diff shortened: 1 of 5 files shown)


bgruening:
Thanks @OlegZharkov

dannon:
I'm always hesitant to strap anything else onto jquery; can we do this on render without the jquery linkage?

OlegZharkov:
> I'm always hesitant to strap anything else onto jquery (especially given we want to remove it completely); can we do this on render without the jquery linkage?

Yeah, it's totally understandable. jQuery is definitely a legacy crap and you're right, probably adding a new plugin is not the best idea. I hope, it is better now. (I wanted to say that babel will take care, but to my surprise `querySelectorAll` is supported even by Internet Explorer 0_o wow)

OlegZharkov:
@dannon thank you

dannon:
@OlegZharkov Thank you!

PR #9753, opened 2020-05-11
[20.01] Check dataset is accessible and use payload in job error API

nsoranzo:
Without the `payload` parameter, it cannot be correctly used by BioBlend.

Also:
- `self.decode_id()` already raise a `MalformedId` exception
- Use `RequestParameterMissingException` instead of `ObjectAttributeMissingException`  when no object is to be created
- Prevent a `KeyError` exception if `dataset_id` is missing from the payload
- Check that the job is in error and the dataset is an output of the job

### DIFF
--- a/lib/galaxy_test/api/test_jobs.py
+++ b/lib/galaxy_test/api/test_jobs.py
@@ -53,7 +53,7 @@ class JobsApiTestCase(ApiTestCase):
 
         # Verify number of ok jobs is actually greater.
         count_increased = False
-        for i in range(10):
+        for _ in range(10):
             new_count = len(self.__uploads_with_state("ok"))
             if original_count < new_count:
                 count_increased = True
@@ -228,10 +228,11 @@ class JobsApiTestCase(ApiTestCase):
             )
             run_response = self._post("tools", data=payload).json()
             job_id = run_response['jobs'][0]["id"]
+            self.dataset_populator.wait_for_job(job_id)
             dataset_id = run_response['outputs'][0]['id']
             response = self._post('jobs/%s/error' % job_id,
                                   data={'dataset_id': dataset_id})
-            assert response.status_code == 200
+            assert response.status_code == 200, response.json()
 
     @skip_without_tool('detect_errors_aggressive')
     def test_report_error_anon(self):
@@ -243,9 +244,9 @@ class JobsApiTestCase(ApiTestCase):
         job_id = run_response['jobs'][0]["id"]
         dataset_id = run_response['outputs'][0]['id']
         response = requests.post('%s/jobs/%s/error' % (self.galaxy_interactor.api_url, job_id),
-                                 params={'email': 'someone@domain.com', 'dataset_id': dataset_id},
+                                 data={'email': 'someone@domain.com', 'dataset_id': dataset_id},
                                  cookies=cookies)
-        assert response.status_code == 200
+        assert response.status_code == 200, response.json()
 
     @uses_test_history(require_new=True)
     def test_deleting_output_keep_running_until_all_deleted(self, history_id):
@@ -597,7 +598,7 @@ class JobsApiTestCase(ApiTestCase):
     def _search(self, payload, expected_search_count=1):
         # in case job and history aren't updated at exactly the same
         # time give time to wait
-        for i in range(5):
+        for _ in range(5):
             search_count = self._search_count(payload)
             if search_count == expected_search_count:
                 break
--- a/lib/galaxy_test/api/test_jobs.py
+++ b/lib/galaxy_test/api/test_jobs.py
@@ -232,7 +232,7 @@ class JobsApiTestCase(ApiTestCase):
             dataset_id = run_response['outputs'][0]['id']
             response = self._post('jobs/%s/error' % job_id,
                                   data={'dataset_id': dataset_id})
-            assert response.status_code == 200, response.json()
+            assert response.status_code == 200, response.text
 
     @skip_without_tool('detect_errors_aggressive')
     def test_report_error_anon(self):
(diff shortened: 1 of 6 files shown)


lib/galaxy_test/api/test_jobs.py
```diff
@@ -243,9 +244,9 @@ def test_report_error_anon(self):
         job_id = run_response['jobs'][0]["id"]
```
nsoranzo, line 244: Since I've introduced a check in the `JobController.error()` API to verify that the job is in error, I need to add `self.dataset_populator.wait_for_job(job_id)` here, but that fails with:
```
urllib3.connectionpool DEBUG 2020-05-12 12:26:41,891 Starting new HTTP connection (1): 127.0.0.1:8506
galaxy.web.framework.decorators ERROR 2020-05-12 12:26:41,925 Uncaught exception in exposed API method:
  File "/usr/users/ga002/soranzon/software/nsoranzo_galaxy/lib/galaxy/web/framework/decorators.py", line 282, in decorator
  File "/usr/users/ga002/soranzon/software/nsoranzo_galaxy/lib/galaxy/webapps/galaxy/api/jobs.py", line 144, in show
    job = self.__get_job(trans, id)
  File "/usr/users/ga002/soranzon/software/nsoranzo_galaxy/lib/galaxy/webapps/galaxy/api/jobs.py", line 366, in __get_job
    return self.job_manager.get_accessible_job(trans, decoded_job_id)
  File "/usr/users/ga002/soranzon/software/nsoranzo_galaxy/lib/galaxy/managers/jobs.py", line 58, in get_accessible_job
    belongs_to_user = (job.user == trans.user) if job.user else (job.session_id == trans.get_galaxy_session().id)
AttributeError: 'NoneType' object has no attribute 'id'
urllib3.connectionpool DEBUG 2020-05-12 12:26:41,926 http://127.0.0.1:8506 "GET /api/jobs/0beccb6ec965c2fd?full=False&key=46459b20d08abf80d32d647860e67069 HTTP/1.1" 500 None
```
because it uses the test user API key not the anonymous user cookie.

Should I:
- drop this API test? (we don't have many other API tests for anonymous users any way)
- or drop the check that the job is in error?

Extending the test populators to support anonymous session seems to be a much larger project. Suggestions?

jmchilton: I'd comment out the check that the job is in error until we can provide some support in populators for doing more with anonymous testing. We don't have much anonymous user testing but that is sort of a reason to not drop the tests we have in my mind.

If instead of you decide to just drop the test though I won't be -1 or anythinig... just in my mind the test is more valuable than the job in error check.

mvdbeek: We shouldn't remove this test, I added it because anonymous error reporting was broken. I think we may want to allow submitting reports for all job states (say someone wants to report an empty dataset, or a job that hangs in the queue or ...). Not that this is exposed in the UI ... but I don't think the state check is necessary.

nsoranzo: Thanks both for the feedback, I've removed the check that the job is in error, but kept the `self.dataset_populator.wait_for_job(job_id)` call in `test_report_error()` above.

nsoranzo:
Green and ready to review!

lib/galaxy_test/api/test_jobs.py
```diff
@@ -243,9 +244,9 @@ def test_report_error_anon(self):
...
         dataset_id = run_response['outputs'][0]['id']
         response = requests.post('%s/jobs/%s/error' % (self.galaxy_interactor.api_url, job_id),
-                                 params={'email': 'someone@domain.com', 'dataset_id': dataset_id},
+                                 data={'email': 'someone@domain.com', 'dataset_id': dataset_id},
                                  cookies=cookies)
-        assert response.status_code == 200
+        assert response.status_code == 200, response.json()
```
mvdbeek:
```suggestion
        assert response.status_code == 200, response.text
```
It's very likely that non-200 status codes won't be valid JSON.

mvdbeek (approved):

lib/galaxy_test/api/test_jobs.py
nsoranzo: This worked fine in my testing:
```
self = <galaxy_test.api.test_jobs.JobsApiTestCase testMethod=test_report_error_anon>

    @skip_without_tool('detect_errors_aggressive')
    def test_report_error_anon(self):
        # Need to get a cookie and use that for anonymous tool runs
        cookies = requests.get(self.url).cookies
        payload = json.dumps({"tool_id": "detect_errors_aggressive",
                              "inputs": {"error_bool": "true"}})
        run_response = requests.post("%s/tools" % self.galaxy_interactor.api_url, data=payload, cookies=cookies).json()
        job_id = run_response['jobs'][0]["id"]
        dataset_id = run_response['outputs'][0]['id']
        response = requests.post('%s/jobs/%s/error' % (self.galaxy_interactor.api_url, job_id),
                                 data={'email': 'someone@domain.com', 'dataset_id': dataset_id},
                                 cookies=cookies)
>       assert response.status_code == 200, response.text
E       AssertionError: {"err_msg": "Job is not in error", "err_code": 400008}
E       assert 400 == 200
E         -400
E         +200

lib/galaxy_test/api/test_jobs.py:248: AssertionError
```

and is used already in `lib/galaxy_test/api/test_search.py` and `lib/galaxy_test/api/test_tools.py`

mvdbeek: Make it a 500 and it will probably not work. I don't think it's a good idea to assume any response will be in JSON if you don't see the expect status code.

mvdbeek: And we should change those other instances too. It's a minor thing, but it isn't quite right.

mvdbeek (approved):

PR #9754, opened 2020-05-11
Add build artifact destination for plugins

dannon:
`<plugin>/static/` is stuff that is included and available as static content.  This adds a `dist` for plugin build artifacts that is .gitignored.

Also updates a few other things I noticed that need to be in .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ database/job_working_directory
 database/jobs_directory
 database/mulled
 database/object_store_cache
+database/objects
 database/openid_consumer_cache
 database/pbs
 database/shed_tools
--- a/.gitignore
+++ b/.gitignore
@@ -179,8 +179,8 @@ packages/*/*.egg-info
 *.rej
 *~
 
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -21,20 +21,25 @@ lib/galaxy/web/framework/static/scripts
...
+database/openid_consumer_cache
 database/pbs
 database/shed_tools
 database/test_errors
 database/tmp
-database/*.sqlite
-database/openid_consumer_cache
-database/dependencies
+database/tool_cache
+database/tool_search_index
```
OlegZharkov, line 42: Thank you @dannon!
```suggestion
database/tool_search_index
database/objects
```
Can we also add `database/objects`?

dannon: Sure, what goes in there?

OlegZharkov: I guess, it stores tool output blobs. LIke... Fastqc files are stored there (at least on my system)

bgruening:
Thanks @dannon!

OlegZharkov:
Thank you

PR #9755, opened 2020-05-11
Client watch fix -- force everything to use modern chokidar

dannon:
xref https://github.com/galaxyproject/galaxy/pull/9725#issuecomment-626889032

This is kind of a nuclear option, but after trying to reconcile all the fsevents bs for a couple hours I gave in and it seems to work just fine in my testing.


jmchilton:
This also fixes the mocha tests for me so I'm merging. Thanks for the fast turn around @dannon

PR #9757, opened 2020-05-11
Citations overhaul

dannon:
Swaps from a mix of deprecated and abandoned libraries to citation-js.  I tried this a while back but there were a few concerns.  Hopefully those are addressed in the newer version of citation-js, because the libraries we were using to do this before are defunct.  (https://yarnpkg.com/package/latex-to-unicode-converter, for one)

Refactored and fixed up the styling to look more coherent (removed bibtex toggle button, gave an appearance of tabs showing the two (for now) forms of display).

Includes a few minor style fixes cleaning up invalid style definitions I came across along the way.

From this:
image

To this:
image


dannon:
The failing selenium test passes for me locally; I probably need to tinker with the timing/waits.

dannon:
The remaining failing selenium test (test_history_multi_view.HistoryMultiViewTestCase.test_create_new_old_slides_next) is unrelated.

client/galaxy/scripts/components/Workflow/Run/WorkflowRun.vue
```diff
@@ -10,7 +10,7 @@
             <workflow-run-success
                 v-else-if="invocations != null"
                 :invocations="invocations"
-                :workflowName="workflowName"
+                :workflow-name="workflowName"
```
mvdbeek, line 13: OK, obvious noob question: are `:workflowName` and `:workflow-name` equivalent ?

mvdbeek (commented):
Can you resolve the conflict @dannon ?

client/galaxy/scripts/components/Workflow/Run/WorkflowRun.vue
dannon: Yep; html attribute names are case insensitive so kebab-case is preferred here; Vue automagically converts this.  https://vuejs.org/v2/style-guide/#Prop-name-casing-strongly-recommended

dannon:
@mvdbeek On it, thanks for taking a look.

dannon:
@mvdbeek Looks like it's good to go now, all green.

mvdbeek:
Very nice, but I found some minor issues. These are citations from bwa-mem (https://usegalaxy.org/root?tool_id=toolshed.g2.bx.psu.edu/repos/devteam/bwa/bwa_mem/0.7.17.1):
Before:
Screenshot 2020-05-24 at 11 29 04
After:
Screenshot 2020-05-24 at 11 29 14

The url is not clickable anymore, and the 3rd citation (https://github.com/galaxyproject/tools-iuc/blob/master/tools/bwa/bwa-mem.xml#L342), which is bibtex only is a bit too short to be useful. I'm not sure if we can always make something useful out of the bibtex formatting, but if we can get the url clickable I think that would be helpful.

dannon:
@mvdbeek I added the link back, good catch -- I wanted the primary citation output in specific (APA) format for copying without the extra markup but meant to tag that link back on the end.  I think the other bibtex entry is just busted, investigating.

dannon:
@mvdbeek The problem for that bwa-mem tool is capitalized bibtex field names.  The standard says these should be fine, per http://www.bibtex.org/Format/, but it doesn't seem to be the modern practice.   I can add a layer tweaking this, I guess?

mvdbeek:
We do recommend doi anyway, I am happy to merge without the tweak.

dannon:
@mvdbeek Yeah; if it seems like a big problem I can add in a layer to massage bibtex field names to a standard case, but it'd be great to merge this in and keep moving.

mvdbeek:
Thanks a lot @dannon, the new citations look very nice!

PR #9758, opened 2020-05-12
Fix the cleanup condition when job state is OK and cleanup_job is set as "onsuccess'

gmauro:
The cleanup condition is not verified when the job state is OK and cleanup_job is set as "onsuccess'.
This PR invert a double negation and fixes the condition.
In our environment, we have accumulated millions of files due to this.


nsoranzo (approved):
Introduced in commit 8f3bb497e7d4453104546743eccb18c3b24c1a53

bgruening:
@gmauro I guess its save to push this to 20.01 or even 19.09.

nsoranzo:
@galaxybot test this

jmchilton:
Crap - I'm sorry :(.

gmauro:
> @gmauro I guess its save to push this to 20.01 or even 19.09.
OK. I created a new PR targetting release_19.09  #9761

PR #9610, opened 2020-04-13
Startup speedups (XML doc cache, delay non-essential tool param parsing & indexing)

mvdbeek:
The new watermark for **full app startup** with main's full toolbox from CVMFS in 9 - 11 seconds (down from 55 - 57 on the dev branch). 

There are three ideas in here: 
- First we store a string with the expanded version of the tool xml document in a dogpile cache (and we override the default pickling since pickling is dead-slow).
- Then we parse tool inputs and outputs only when needed (i.e building to tool form / module for the first time)
- Push toolbox indexing to the background

things left to do:
- ~~move the cache decorator to own module~~
- ~~the modification time needs to be included in the cache key~~
- ~~separate the cache regions by shed tool config~~

If we store the dogpile cache (a single ~ 130 MB file) on CVMFS I think we can really speedup first time bootstrapping of instances (I suppose CVMFS works better for large files).

xref https://github.com/galaxyproject/galaxy/issues/7752

PR #9759, opened 2020-05-12
[20.05] Switch to lxml for xml parsing

mvdbeek:
In https://github.com/galaxyproject/galaxy/pull/9610 I started parsing tool documents using lxml for better speed. Unfortunately you can't add `xml.etree.ElementTree.Element`s to documents parsed using lxml, so when we add some dynamic parameters like the JobResourceParameters we get this:
```
galaxy.tools.toolbox.base ERROR 2020-05-11 12:41:10,107 Error reading tool from path: toolshed.g2.bx.psu.edu/repos/devteam/bowtie2/749c918495f7/bowtie2/bowtie2_wrapper.xml
  File "/cvmfs/test.galaxyproject.org/galaxy/lib/galaxy/tools/toolbox/base.py", line 657, in _load_tool_tag_set
    tool = self.load_tool(concrete_path, guid=guid, tool_shed_repository=tool_shed_repository, use_cached=False, tool_cache_data_dir=tool_cache_data_dir)
  File "/cvmfs/test.galaxyproject.org/galaxy/lib/galaxy/tools/toolbox/base.py", line 851, in load_tool
    tool = self.create_tool(config_file=config_file, tool_shed_repository=tool_shed_repository, guid=guid, tool_cache_data_dir=tool_cache_data_dir, **kwds)
  File "/cvmfs/test.galaxyproject.org/galaxy/lib/galaxy/tools/__init__.py", line 298, in create_tool
    tool = self._create_tool_from_source(tool_source, config_file=config_file, **kwds)
  File "/cvmfs/test.galaxyproject.org/galaxy/lib/galaxy/tools/__init__.py", line 316, in _create_tool_from_source
    return create_tool_from_source(self.app, tool_source, **kwds)
  File "/cvmfs/test.galaxyproject.org/galaxy/lib/galaxy/tools/__init__.py", line 237, in create_tool_from_source
    tool = ToolClass(config_file, tool_source, app, **kwds)
  File "/cvmfs/test.galaxyproject.org/galaxy/lib/galaxy/tools/__init__.py", line 508, in __init__
    self.populate_resource_parameters(tool_source)
  File "/cvmfs/test.galaxyproject.org/galaxy/lib/galaxy/tools/__init__.py", line 1318, in populate_resource_parameters
    inputs.append(resource_xml)
TypeError: Argument 'element' has incorrect type (expected lxml.etree._Element, got xml.etree.ElementTree.Element)
``` 

Here I switched all of our ElementTree usage to lxml (if available) and I also tried unifying the names with which we do the imports. The central place to import etree from should be `galaxy.util` now, which falls back to stdlib if lxml is not available (as could happen when using `galaxy-util`, `galaxy-tool-util`, etc).

### DIFF
--- a/lib/galaxy/authnz/__init__.py
+++ b/lib/galaxy/authnz/__init__.py
@@ -24,13 +24,13 @@ class IdentityProvider(object):
         :type provider: string
         :param provider: is the name of the identity provider (e.g., Google).
 
-        :type config: xml.etree.ElementTree.Element
+        :type config: lxml.etree.ElementTree._Element
         :param config: Is the configuration element of the provider
             from the configuration file (e.g., oidc_config.xml).
             This element contains the all the provider-specific
             configuration elements.
 
-        :type backend_config: xml.etree.ElementTree.Element
+        :type backend_config: lxml.etree.ElementTree._Element
         :param backend_config: Is the configuration element of the backend of
             the provider from the configuration file (e.g.,
             oidc_backends_config.xml). This element contains all the
--- a/lib/galaxy/jobs/runners/util/cli/job/torque.py
+++ b/lib/galaxy/jobs/runners/util/cli/job/torque.py
@@ -80,7 +80,7 @@ class Torque(BaseJobExec):
         rval = {}
         for line in status.strip().splitlines():
             try:
-                tree = parse_xml_string(line.strip)
+                tree = parse_xml_string(line.strip())
                 assert tree.tag == 'Data'
                 break
             except Exception:
--- a/lib/galaxy/jobs/runners/util/cli/job/torque.py
+++ b/lib/galaxy/jobs/runners/util/cli/job/torque.py
@@ -1,8 +1,4 @@
 from logging import getLogger
-try:
-    import xml.etree.cElementTree as et
-except ImportError:
-    import xml.etree.ElementTree as et
 
 try:
     from galaxy.model import Job
@@ -12,6 +8,7 @@ except ImportError:
     from pulsar.util import enum
     job_states = enum(RUNNING='running', OK='complete', QUEUED='queued')
 
+from galaxy.util import parse_xml_string
 from ..job import BaseJobExec
 
 log = getLogger(__name__)
@@ -83,7 +80,7 @@ class Torque(BaseJobExec):
         rval = {}
         for line in status.strip().splitlines():
             try:
-                tree = et.fromstring(line.strip())
+                tree = parse_xml_string(line.strip)
                 assert tree.tag == 'Data'
                 break
             except Exception:
--- a/lib/galaxy/tool_shed/galaxy_install/tools/tool_panel_manager.py
+++ b/lib/galaxy/tool_shed/galaxy_install/tools/tool_panel_manager.py
@@ -1,12 +1,16 @@
 import errno
 import logging
-from xml.etree import ElementTree as XmlET
+
+import lxml.etree as XmlET
 
 from galaxy.exceptions import RequestParameterInvalidException
 from galaxy.tool_shed.util.basic_util import strip_path
 from galaxy.tool_shed.util.repository_util import get_repository_owner
 from galaxy.tool_shed.util.shed_util_common import get_tool_panel_config_tool_path_install_dir
-from galaxy.util import xml_to_string
+from galaxy.util import (
+    parse_xml_string,
+    xml_to_string,
+)
 from galaxy.util.renamed_temporary_file import RenamedTemporaryFile
 from galaxy.util.tool_shed.common_util import remove_protocol_and_user_from_clone_url
 from galaxy.util.tool_shed.xml_util import parse_xml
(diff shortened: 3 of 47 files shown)


lib/galaxy/jobs/runners/util/cli/job/torque.py
```diff
@@ -83,7 +80,7 @@ def parse_status(self, status, job_ids):
         rval = {}
         for line in status.strip().splitlines():
             try:
-                tree = et.fromstring(line.strip())
+                tree = parse_xml_string(line.strip)
```
natefoo: Did you mean to remove the parens here?

lib/galaxy/authnz/__init__.py
```diff
@@ -24,13 +24,13 @@ def __init__(self, provider, config, backend_config):
         :type provider: string
         :param provider: is the name of the identity provider (e.g., Google).
 
-        :type config: xml.etree.ElementTree.Element
+        :type config: lxml.etree.ElementTree._Element
```
natefoo, line 27: Just a doc so I'm probably bikeshedding here but does this make sense when the parser might not be lxml?

lib/galaxy/tool_shed/galaxy_install/tools/tool_panel_manager.py
```diff
@@ -109,24 +112,24 @@ def generate_tool_elem(self, tool_shed, repository_name, changeset_revision, own
                            tool, tool_section):
         """Create and return an ElementTree tool Element."""
         if tool_section is not None:
-            tool_elem = XmlET.SubElement(tool_section, 'tool')
+            tool_elem = etree.SubElement(tool_section, 'tool')
         else:
-            tool_elem = XmlET.Element('tool')
+            tool_elem = etree.Element('tool')
```
natefoo, line 117: Need to get `Element` first with `Element = getattr(etree, '_Element', etree.Element)` as you did elsewhere?

lib/galaxy/tool_shed/tools/data_table_manager.py
```diff
@@ -18,21 +18,21 @@ def generate_repository_info_elem(self, tool_shed, repository_name, changeset_re
                                       parent_elem=None, **kwd):
         """Create and return an ElementTree repository info Element."""
         if parent_elem is None:
-            elem = XmlET.Element('tool_shed_repository')
+            elem = etree.Element('tool_shed_repository')
```
natefoo, line 21: `Element = getattr(etree, '_Element', etree.Element)`?

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -273,7 +273,7 @@ def populate_state(request_context, inputs, incoming, state, errors={}, prefix='
     """
     Populates nested state dict from incoming parameter values.
     >>> from collections import OrderedDict
-    >>> from xml.etree.ElementTree import XML
+    >>> from lxml.etree import XML
```
natefoo:
```suggestion
    >>> from galaxy.util import XML
```
?

lib/galaxy/util/__init__.py
```diff
@@ -236,34 +242,39 @@ def unique_id(KEY_SIZE=128):
...
+        parser = etree.XMLParser(remove_comments=remove_comments)
     try:
-        root = tree.parse(fname, parser=ElementTree.XMLParser(target=DoctypeSafeCallbackTarget()))
-        for elem in root.iter('*'):
-            if elem.text is not None:
-                elem.text = elem.text.strip()
-            if elem.tail is not None:
-                elem.tail = elem.tail.strip()
-    except ParseError:
+        # Restore ENOENT that would otherwise be an OSError in lxml
```
natefoo: Weird filesystem semantics make it so it's usually safer to try open, catch, and raise if opening fails (usually faster than checking for existence first too) - is the `OSError`'s `errno` attribute set incorrectly in the case of lxml?

Plus IOErrors are OSErrors in Python 3.

lib/galaxy/util/tool_shed/xml_util.py
```diff
@@ -47,7 +28,7 @@ def create_element(tag, attributes=None, sub_elements=None):
     key / value pairs in the received attributes and sub_elements.
     """
     if tag:
-        elem = XmlET.Element(tag)
+        elem = etree.Element(tag)
```
natefoo, line 31: `Element = getattr(etree, '_Element', etree.Element)` ?

natefoo (approved):

lib/galaxy/util/tool_shed/xml_util.py
mvdbeek: This is only necessary for the `isinstance` check, since the actual `Element` class is `_Element` and `Element` is the constructor.

lib/galaxy/jobs/runners/util/cli/job/torque.py
mvdbeek: Definitely not, thanks!
```suggestion
                tree = parse_xml_string(line.strip())
```

lib/galaxy/util/__init__.py
mvdbeek: > is the `OSError`'s `errno` attribute set incorrectly in the case of lxml?

indeed, and that leads to this fun undefined behavior where we catch `IOError`s but then don't do anything if errno is not `ENOENT` - https://github.com/mvdbeek/galaxy/blob/a253acc31bf8aaa903e8e5fdbe44b31b204a0393/lib/galaxy/tools/data_manager/manager.py#L71.

I'll add another `except IOError` and change the errno there if appropriate.

mvdbeek: `e.errno` is `None` for missing files. I've added the missing errno in https://github.com/galaxyproject/galaxy/pull/9759/commits/0a8634b2f70be23c3fae85b88d067844acc2b998, I think this should be good enough as a workaround.

PR #9760, opened 2020-05-12
[20.01] Set UTF-8 for parsing tool xml

AndreasSko:
When trying to install a tool (with Python3) that contained non-ascii characters (like toolshed.g2.bx.psu.edu/repos/rnateam/viennarna_rnafold/bdb786715d28/viennarna_rnafold/rnafold.xml), I encountered the error:

> UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 5074: ordinal not in range(128)

This change will set the encoding to UTF-8, which should not hurt when normally expecting ASCII encoding, as they are compatible to each other. Tested both on 20.01 and 20.05 - this fixed the problem.

### DIFF
--- a/lib/galaxy/util/tool_shed/xml_util.py
+++ b/lib/galaxy/util/tool_shed/xml_util.py
@@ -91,11 +91,10 @@ def parse_xml(file_name, check_exists=True):
     if check_exists and not os.path.exists(file_name):
         return None, "File does not exist %s" % str(file_name)
 
-    with open(file_name, 'r') as fobj:
-        try:
-            tree = XmlET.parse(fobj, parser=XmlET.XMLParser(target=CommentedTreeBuilder()))
-        except Exception as e:
-            error_message = "Exception attempting to parse %s: %s" % (str(file_name), str(e))
-            log.exception(error_message)
-            return None, error_message
+    try:
+        tree = XmlET.parse(file_name, parser=XmlET.XMLParser(target=CommentedTreeBuilder()))
+    except Exception as e:
+        error_message = "Exception attempting to parse %s: %s" % (str(file_name), str(e))
+        log.exception(error_message)
+        return None, error_message
     return tree, error_message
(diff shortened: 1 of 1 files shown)


AndreasSko:
CC @bgruening

lib/galaxy/util/tool_shed/xml_util.py
```diff
@@ -91,7 +91,7 @@ def parse_xml(file_name, check_exists=True):
     if check_exists and not os.path.exists(file_name):
         return None, "File does not exist %s" % str(file_name)
 
-    with open(file_name, 'r') as fobj:
+    with open(file_name, 'r', encoding='utf-8') as fobj:
```
nsoranzo: For people still stuck with Python2:
```suggestion
    with io.open(file_name, 'r', encoding='utf-8') as fobj:
```

bgruening:
@AndreasSko can you apply this to release_20.05, please

AndreasSko:
Thanks for the quick review @nsoranzo @bgruening ! :)
Rebased to release_20.05

nsoranzo (approved):

nsoranzo:
@galaxybot test this

lib/galaxy/util/tool_shed/xml_util.py
```diff
@@ -91,7 +91,7 @@ def parse_xml(file_name, check_exists=True):
     if check_exists and not os.path.exists(file_name):
         return None, "File does not exist %s" % str(file_name)
 
-    with open(file_name, 'r') as fobj:
+    with io.open(file_name, 'r', encoding='utf-8') as fobj:
```
mvdbeek: ElementTree parses bytes internally, so the better fix is to not open `file_name` but pass it straight into `XmlET.parse(file_name, parser=XmlET.XMLParser(target=CommentedTreeBuilder()))`. This should handles non-utf8 documents better.

mvdbeek:
https://github.com/galaxyproject/galaxy/pull/9759 should also solve this for 20.05, but we may still want to target 20.01 with this fix or my suggestion.

bernt-matthias:
Isn't the Citation the root of the problem?

nsoranzo:
@AndreasSko Can you please resolve the merge conflict?

bgruening:
Is this still needed after @mvdbeek PR was merged?

mvdbeek:
Not against 20.05, but it may make sense to patch 20.01 that way.

mvdbeek:
@galaxybot test this

mvdbeek (approved):

PR #9761, opened 2020-05-12
[19.09] Fix the cleanup condition when job state is OK and cleanup_job is set as "onsuccess'

gmauro:
As suggested into #9758, I am pushing it back to release_19.09


nsoranzo (approved):

nsoranzo:
@galaxybot test this

nsoranzo:
I think the issue affects only resubmitted jobs, for the other successful jobs the call to [job_wrapper.finish()](https://github.com/gmauro/galaxy/blob/release_19.09-cleanup_job/lib/galaxy/jobs/runners/__init__.py#L521) should have deleted the job files.

Also, I have a cron job to remove old unsuccessful jobs' files after 30 days which would remove also these spurious directories:
```
tmpwatch --ctime 30d database/jobs_directory/
```

nsoranzo:
Thanks @gmauro!

PR #9762, opened 2020-05-12
Add certificate and token verification

juleengraham:


lib/galaxy/authnz/custos_authnz.py
```diff
@@ -220,6 +230,28 @@ def _validate_nonce(self, trans, nonce_hash):
...
+        response = requests.get(self.config['certificate_endpoint'],
+                           headers={"Authorization": "Basic %s" % util.unicodify(base64.b64encode(util.smart_str(clientIdAndSec)))},
+                           verify=False, params={'metadata.owner_type': "CUSTOS",
+                                                'metadata.resource_type': "SERVER_CERTIFICATE"})
+        self.config['ca_bundle'] = response.json().get("value")
+
+        res = ssl.get_default_verify_paths()
+        f = open(res[0], "a")
+        f.write(self.config['ca_bundle'])
+        f.close()
```
jdavcs:
```suggestion
        with open(res[0], "a") as f:
            f.write(self.config['ca_bundle'])
```
This way `f` is always closed, even if an exception was raised.

juleengraham: thanks!

PR #9768, opened 2020-05-14
Added the missing return value. Added try/catch for another irods call.

kxk302:
After my fix yesterday, I have only seen 5 irods errors on Sentry

Yesterday, I added a generic catch clause to data_object_exists(), but forgot to include the return statement. Its included now. There are 3 entries in Sentry where a NetworkException is thrown from data_object_exists(), even though there is generic catch clause surrounding the code. I don't know how to explain this (except maybe a timing issue, where the code had the generic clause, but the old software was running?)  

There are issues Sentry, where  a NetworkException is thrown in push_to_irods().  The call to exists() could throw a NetworkException. Wrapped that call with a try/catch as well.


mvdbeek (approved):
Thank you @kxk302!

Issue #9689, opened 2020-04-30
Workflow reports - enable embedding of extra files produced by tools


wm75:
In a nutshell, what I would like to be able to do is something along the lines of:

`invocation_outputs(primary_dataset.files_path['very-informative-extra-file'])`

Think of a tool like FastQC for example. The html it produces is a report on its own already and embedding that whole thing in an outer workflow report often feels a bit awkward. What I would typically like to do instead is just embed one or two particular images from the tool report into the workflow report. Those images are actually available as separate files referenced by the primary dataset's files_path attribute. If those extra files could be made accessible through syntax similar to the one above, it'd be great!

wm75:
@OlegZharkov If you would like to work on this, I can explain things in more detail offline if you want.

jmchilton:
Sounds like a great idea - @OlegZharkov want to prototype this and then ping me on it? Thanks so much.

OlegZharkov:
@jmchilton 
Well, it will take a while to do. Can you assign me on this issue, please?

jmchilton:
This has been done by @OlegZharkov - thanks for the request and for the implementation. Really nice stuff!

PR #9769, opened 2020-05-14
Workflow reports - composite datasets

OlegZharkov:
Implementation of https://github.com/galaxyproject/galaxy/issues/9689
This prototype allows to embed files from individual composite datasets to workflow reports in the multiple ways.

General use: 

 \```galaxy
 history_dataset_composite(output="{output_label}", path="{Filepath}", label="{link_custom_text}", image={true})
 \```
- **path** a path to a preferred file. If it's wrong, it will render `file {path} not found`. If absent,  it will give embed a link to the whole dataset. 
- **label** Custom text on `<a>` element (link text). If abesent, it will render `DATASET: {dataset_name} FILEPATH: {path}`.
- If **path** is a `Directory`. It will render the content of specified directory.
- If **path** and **link** are absent, user will get a link to whole dataset with _dataset_id_ as a label
- If **image** equals _true_ or _"true"_, image will be embedded in report.

Example:

this [report](https://gist.github.com/OlegZharkov/535c7a9880108b7398c2a30d4b306463) will get you:

image

image

Of course, you're always welcome to propose new features. Should I change/improve something? Maybe Directory content table? I made it in the center and I am still not sure if it was a good idea. Should I align it back to the left part of the div (like a regular markdown element)?

For someone, who wants to reproduce it or just play around, feel free to try this [workflow](https://files.gitter.im/OlegZharkov/b2Gv/Galaxy-Workflow-Unnamed_workflow.ga) (Note: workflow requires `Fastqc` tool, that can be easily downloaded directly from toolshed in admin interface. I ran it with this [sequence](https://zenodo.org/record/61771/files/GSM461178_untreat_paired_subset_1.fastq
)) 

Thanks a lot fot @wm75 and @jmchilton !

P.S.  Additionally, it fixes bug, where 3rd argument of a function call was not registered in workflow reports (minor issue and we didn't have function calls with 3 arguments till now)
Cheers!


jmchilton:
The functionality seems amazing and way broader than I was expecting. This is very impressive - way to implement this so fast and so completely.

I love the functionality and the framework fixes - I hope you're okay if I nitpick the syntax a bit. I think the syntax I would like will be ... harder to implement but I don't know how much harder. My big concern is we had favored lots of different functions with really simple inputs that all produced one kind of presentational element and this is a complex function that produces different things - it doesn't feel like it gels with the existing functionality.

For instance, previously we had three functions ``history_dataset_display``, ``history_dataset_embedded``, and ``history_dataset_as_image``. If we had had one function with parameters to declare it embedded vs normal vs as an image - it would match I think what you implemented here.

Another small nitpick is you can have extra files on non-composite datatypes (it isn't common yet, but it might be more common as we incorporate more CWL features and such) so I'd like to avoid the word composite if possible. My preferred syntax might be something like:

---

```
history_dataset_index(output, [path])
```

To produce the index of a directory in a composite dataset.

---

Update:

```
history_dataset_as_image
```

and

```
history_dataset_embedded
```

to take in a path optionally and then work on an extra file.

---

And then implement another new tag for generating links to files:

```
history_dataset_link(output, [path], [label])
```

---

You've already done so much and I'm asking a lot so I'm more than happy to help - just let me know I will find time. This is really cool stuff.

OlegZharkov:
Hi @jmchilton. Thanks a lot for the detailed review. I reworked it, so it will satisfy listed requirements. Before I continue further implementation (mostly testing), I just wanted to ask you if you're fine with the current functionality. 

#### `history_dataset_link`
it didn't change much from the initial version:

 \```galaxy
 history_dataset_link(output="{output_label}", path="{Filepath}", label="{link_custom_text}"
 \```

- **path** a path to a preferred file. If it's wrong, it will render `file {path} not found`. If absent,  it will render a link to the whole dataset. 
- **label** Custom text on `<a>` element (link text). If abesent, it will render `DATASET: {dataset_name} FILEPATH: {path}`.
- If **path** is a `Directory` (subdir). It will show ``Path: {this.path} is a directory!``
- If **path** and **label** are absent, user will get a link to whole dataset with _dataset_id_ as a label

this
https://gist.github.com/OlegZharkov/032df0999a660279602f5ed44143ce59
will render:

image


####  `history_dataset_index(output="{output_label}", path="{Filepath}")`
a new component that will show directory index. Reworked from scratch with Vue and bootrap table (does not depend on server rendering anymore)

**path** to a subdir. If absent, it will show dataset's rootDirectory tree. If dataset is not composite,  provided path leads to a file or if the directory doesn't exist, it will say corresponding message. 

example: 
https://gist.github.com/OlegZharkov/ce845e8a8217e33f023886b016ea1e41

will get you: 

image


image


and finally
#### history_dataset_as_image(output="{output_label}", path="{Filepath}")

If used, it will show a picture inside workflow report. If the path is specified, but not valid, it will get you ```Image is not found``` message.

Example:
https://gist.github.com/OlegZharkov/2859abbb6630cb7e462273119a1caff3
will create:

image


If you want to try locally, feel free to use this workflow. Report is already included
https://gist.github.com/OlegZharkov/9bf073b5d4ed49555bc9853372a60064


P.S.
Most things are (hopefully) cached on the client side now, so it shouldn't create too many server requests. 

P.S.S.
If I need to implement full-stack integration test, should I use cypress instead of selenium now?

client/galaxy/scripts/components/Markdown/Markdown.vue
```diff
@@ -193,6 +214,8 @@ export default {
                 render_embedded_items();
                 mountJobMetrics({ includeTitle: false });
                 mountJobParameters({ includeTitle: false });
+                mountDatasetComponents();
```
jmchilton, line 217: This is so awesome!

client/galaxy/scripts/components/Dataset/DatasetLink/DatasetLink.vue
```diff
@@ -0,0 +1,67 @@
...
+
+            if (!this.pathDestination.fileLink) return `Path: ${this.path} was not found!`;
+
+            if (this.label !== undefined && this.label !== "undefined") {
+                return this.label;
+            } else {
+                return `${
+                    this.pathDestination.datasetRootDir && this.path
+                        ? `DATASET: ${this.pathDestination.datasetRootDir} FILEPATH: ${this.path}`
+                        : `${this.history_dataset_id}`
```
jmchilton, line 58: Can you use the Dataset name here instead? Is datasetRootDir somehow the dataset name in the line above? It seemed like it in the example image but the name here is a bit confusing.

jmchilton:
This is so cool, I'm really into the macro names and argument names. The report Markdown facing part of this is perfect to my mind.

jmchilton:
I pushed 96b0646 that does a lot more validation of these new multi-argument macros - both as unit tests of the parser and as richer linting.

jmchilton:
And I already see a problem with that - probably shouldn't have pushed that right into your branch - sorry.

OlegZharkov:
> And I already see a problem with that - probably shouldn't have pushed that right into your branch - sorry.

no worries, what is the problem?

client/galaxy/scripts/components/Dataset/DatasetLink/DatasetLink.vue
OlegZharkov: Yes, _dataset name_ and _dataset root dir name_  are actually the same

bgruening:
Merged as part of #9938

jmchilton:
Thanks so very much - this was really awesome work @OlegZharkov ! The other PR was merged with a rebased version of that commit I added that was cleaned up a bit.

PR #9776, opened 2020-05-16
[20.05] inputs config file generation depend on profile

### LINKED ISSUE #9773: Empty select-input treated as [] instead of "None" in 20.05
I'm trying to run [this](https://github.com/usegalaxy-eu/workflow-testing/blob/master/sklearn/ard/ard.ga) workflow on 20.05 (using Planemo test), which worked in 20.01 with Python 3. When running the sklearn_searchcv-tool the error occurs:

> Traceback (most recent call last):
>   File "/galaxy/database/jobs_directory/000/4/configs/tmp_1z1p9jo", line 72, in <module>
>     options['scoring'] = get_scoring(options['scoring'])
>   File "<string>", line 295, in get_scoring
> AttributeError: 'list' object has no attribute 'split'
> ]

In the end I tracked the error down to Galaxy "interpreting" an empty `select`-input differently between 20.01 and 20.05:

**20.01:**
```
"options": {
      "scoring": {
        "secondary_scoring": "None",
        "primary_scoring": "r2"
      },
```
**20.05:**
```
"scoring": {
        "secondary_scoring": [],
        "primary_scoring": "r2"
      },
```

However, I'm not sure if this change is actually desired or considered a bug..

The full inputs for the tool:
[20.01_py3.txt](https://github.com/galaxyproject/galaxy/files/4635700/20.01_py3.txt)
[20.05_py3.txt](https://github.com/galaxyproject/galaxy/files/4635701/20.05_py3.txt)

bernt-matthias:
fixes https://github.com/galaxyproject/galaxy/issues/9773

follow up on https://github.com/galaxyproject/galaxy/pull/9279

do we need a test for the old profile?

### DIFF
--- a/lib/galaxy/tool_util/xsd/galaxy.xsd
+++ b/lib/galaxy/tool_util/xsd/galaxy.xsd
@@ -4865,6 +4865,8 @@ A contrived example of a tool that uses this is the test tool
 
 By default this file will not contain paths for data or collection inputs. To include simple
 paths for data parameters set the ``data_style`` attribute to ``paths`` (see [inputs_as_json_with_paths.xml](https://github.com/galaxyproject/galaxy/blob/dev/test/functional/tools/inputs_as_json_with_paths.xml) for an example).
+
+For tools with profile >=20.05 a select with ``multiple="true"`` is rendered as a list which is empty if nothing is selected. For smaller profiles lists are rendered as comma separated string which and ``"None"`` in case nothing is selectecd.
 ]]></xs:documentation>
     </xs:annotation>
     <xs:simpleContent>
--- a/lib/galaxy/tool_util/xsd/galaxy.xsd
+++ b/lib/galaxy/tool_util/xsd/galaxy.xsd
@@ -4866,7 +4866,7 @@ A contrived example of a tool that uses this is the test tool
 By default this file will not contain paths for data or collection inputs. To include simple
 paths for data parameters set the ``data_style`` attribute to ``paths`` (see [inputs_as_json_with_paths.xml](https://github.com/galaxyproject/galaxy/blob/dev/test/functional/tools/inputs_as_json_with_paths.xml) for an example).
 
-For tools with profile >=20.05 a select with ``multiple="true"`` is rendered as a list which is empty if nothing is selected. For smaller profiles lists are rendered as comma separated string which and ``"None"`` in case nothing is selectecd.
+For tools with profile >= 20.05 a select with ``multiple="true"`` is rendered as an array which is empty if nothing is selected. For older profile versions select lists are rendered as comma separated strings or a literal ``null`` in case nothing is selected.
 ]]></xs:documentation>
     </xs:annotation>
     <xs:simpleContent>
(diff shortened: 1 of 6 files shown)


lib/galaxy/tools/parameters/wrapped_json.py
```diff
@@ -88,7 +92,7 @@ def _data_input_to_path(v):
     elif input_type == "boolean":
         json_value = _cast_if_not_none(value_wrapper, bool)
     elif input_type == "select":
-        if input.multiple:
+        if input.multiple and profile >= '20.05':
```
mvdbeek: `profile` is a float, so you'd have to use `if input.multiple and profile >= 20.05:` 
But we usually use `packaging.version.parse(str(profile)) >= packaging.version.parse('20.05')`

mvdbeek:
No harm in having in a test :).

AndreasSko:
Thank you for your quick response with this PR, @bernt-matthias ! :) I tested it with the workflow, but unfortunately this does not completely solve to problem (yet). The input is now: `"secondary_scoring": "[]"`, but "should" be: `"secondary_scoring": "None"`

lib/galaxy/tools/parameters/wrapped_json.py
```diff
@@ -88,7 +92,7 @@ def _data_input_to_path(v):
...
             json_value = [_ for _ in _cast_if_not_none(value_wrapper.value, list)]
         else:
             json_value = _cast_if_not_none(value_wrapper.value, str)
```
AndreasSko:
```suggestion
            json_value = _cast_if_not_none(value_wrapper, str)
```
This fixed https://github.com/galaxyproject/galaxy/issues/9773  for me, though not sure if that has some different side-effects

bernt-matthias: Thanks @AndreasSko your suggestion is correct, casting `value_wrapper` was the implementation before https://github.com/galaxyproject/galaxy/pull/9279

lib/galaxy/tools/parameters/wrapped_json.py
bernt-matthias: Thanks for the suggestion.

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -4865,6 +4865,8 @@ A contrived example of a tool that uses this is the test tool
 
 By default this file will not contain paths for data or collection inputs. To include simple
 paths for data parameters set the ``data_style`` attribute to ``paths`` (see [inputs_as_json_with_paths.xml](https://github.com/galaxyproject/galaxy/blob/dev/test/functional/tools/inputs_as_json_with_paths.xml) for an example).
+
+For tools with profile >=20.05 a select with ``multiple="true"`` is rendered as a list which is empty if nothing is selected. For smaller profiles lists are rendered as comma separated string which and ``"None"`` in case nothing is selectecd.
```
```suggestion
For tools with profile >= 20.05 a select with ``multiple="true"`` is rendered as an array which is empty if nothing is selected. For older profile versions select lists are rendered as comma separated strings or a literal ``null"`` in case nothing is selected.
```

bernt-matthias:
Should the target branch be changed?

mvdbeek:
Up to you, I tried it but since it was based on dev I couldn't do it without rebasing :).

bernt-matthias:
OK. Seems that rebasing worked.

mvdbeek (approved):
Thank you @bernt-matthias and @AndreasSko! These are important fixes.

Issue #9777, opened 2020-05-16
shed_tool_conf.xml not automatically created in 20.05


AndreasSko:
When installing new tools on a clean Galaxy setup, the `shed_tool_conf.xml` file is not created anymore. Rather the error occurs:

> galaxy.util.tool_shed.xml_util ERROR 2020-05-16 14:32:31,473 [p:229,w:1,m:0] [uWSGIWorker1Core2] Exception attempting to parse /galaxy/config/shed_tool_conf.xml: Error reading file '/galaxy/config/shed_tool_conf.xml': failed to load external entity "/galaxy/config/shed_tool_conf.xml"
> Traceback (most recent call last):
>   File "lib/galaxy/util/tool_shed/xml_util.py", line 75, in parse_xml
>     tree = galaxy_parse_xml(file_name, remove_comments=False, strip_whitespace=False)
>   File "lib/galaxy/util/__init__.py", line 237, in parse_xml
>     tree = etree.parse(fname, parser=parser)
>   File "src/lxml/etree.pyx", line 3519, in lxml.etree.parse
>   File "src/lxml/parser.pxi", line 1839, in lxml.etree._parseDocument
>   File "src/lxml/parser.pxi", line 1865, in lxml.etree._parseDocumentFromURL
>   File "src/lxml/parser.pxi", line 1769, in lxml.etree._parseDocFromFile
>   File "src/lxml/parser.pxi", line 1163, in lxml.etree._BaseParser._parseDocFromFile
>   File "src/lxml/parser.pxi", line 601, in lxml.etree._ParserContext._handleParseResultDoc
>   File "src/lxml/parser.pxi", line 711, in lxml.etree._handleParseResult
>   File "src/lxml/parser.pxi", line 638, in lxml.etree._raiseParseError
> OSError: Error reading file '/galaxy/config/shed_tool_conf.xml': failed to load external entity "/galaxy/config/shed_tool_conf.xml"
> galaxy.tool_shed.galaxy_install.tools.tool_panel_manager ERROR 2020-05-16 14:32:31,475 [p:229,w:1,m:0] [uWSGIWorker1Core2] Exception attempting to parse /galaxy/config/shed_tool_conf.xml: Error reading file '/galaxy/config/shed_tool_conf.xml': failed to load external entity "/galaxy/config/shed_tool_conf.xml"

[...]

PR #9779, opened 2020-05-17
Fix bug;  simplify creation of default shed conf files

jdavcs:
1\. Fix bug #9777 

2\. Explicitly check for file existence (before creating default file) instead of relying on exception handling. 

The logic prior to #9759 (try parsing file; if exception occurs, create default file) worked because `parse_xml` tried to open the file using a `with` statement, without handling the exception that occurred if the file was not found. ([see prev. version](https://github.com/galaxyproject/galaxy/blob/c998d709d85363344a1a92ef9514cdeca86815f8/lib/galaxy/util/tool_shed/xml_util.py#L94)) That exception was handled by the caller, `ToolPanelManager`, with the default file created inside the `except` clause.

The current version of `parse_xml` handles all exceptions (good!), so the caller's `except` clause responsible for creating the default file is never reached. Although we could check the exception type and re-raise it (if `errno.ENOENT`), I think, this version is simpler and more straightforward. Also, we're not using exception handling to implement control flow.

2\. We do not need to wrap the call to `parse_xml` in `try/except` because any exceptions are handled.

3\. The same applies to `data_manager.py`.

4\. We no longer need the `check_exists` parameter in `parse_xml`.

5\. I've removed [this line](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tool_shed/galaxy_install/tools/tool_panel_manager.py#L50): if the file is not loaded a (more specific) exception will be logged anyway.

Ideally, this needs dedicated automated tests, but that may delay fixing the referenced bug, which I think is more important here.

Ping @natefoo, @mvdbeek

### DIFF
--- a/lib/galaxy/tool_shed/galaxy_install/tools/data_manager.py
+++ b/lib/galaxy/tool_shed/galaxy_install/tools/data_manager.py
@@ -62,8 +62,8 @@ class DataManagerHandler(object):
 
             # If conf file does not exist, create it.
             if not os.path.exists(shed_data_manager_conf_filename):
+                log.info('Creating shed data manager config with default contents: %s', shed_data_manager_conf_filename)
                 with open(shed_data_manager_conf_filename, 'w') as fh:
-                    log.info('Creating shed data manager config with default contents: %s', shed_data_manager_conf_filename)
                     fh.write(SHED_DATA_MANAGER_CONF_XML)
 
             # Load existing data managers.
--- a/lib/galaxy/tool_shed/galaxy_install/tools/data_manager.py
+++ b/lib/galaxy/tool_shed/galaxy_install/tools/data_manager.py
@@ -1,4 +1,3 @@
-import errno
 import logging
 import os
 import time
@@ -60,16 +59,15 @@ class DataManagerHandler(object):
             repository_tools_by_guid = {}
             for tool_tup in repository_tools_tups:
                 repository_tools_by_guid[tool_tup[1]] = dict(tool_config_filename=tool_tup[0], tool=tool_tup[2])
+
+            # If conf file does not exist, create it.
+            if not os.path.exists(shed_data_manager_conf_filename):
+                with open(shed_data_manager_conf_filename, 'w') as fh:
+                    log.info('Creating shed data manager config with default contents: %s', shed_data_manager_conf_filename)
+                    fh.write(SHED_DATA_MANAGER_CONF_XML)
+
             # Load existing data managers.
-            try:
-                tree, error_message = parse_xml(shed_data_manager_conf_filename, check_exists=False)
-            except (OSError, IOError) as exc:
-                if exc.errno == errno.ENOENT:
-                    with open(shed_data_manager_conf_filename, 'w') as fh:
-                        fh.write(SHED_DATA_MANAGER_CONF_XML)
-                    tree, error_message = parse_xml(shed_data_manager_conf_filename)
-                else:
-                    raise
+            tree, error_message = parse_xml(shed_data_manager_conf_filename)
             if tree is None:
                 return rval
             config_elems = [elem for elem in tree.getroot()]
--- a/lib/galaxy/tool_shed/galaxy_install/tools/tool_panel_manager.py
+++ b/lib/galaxy/tool_shed/galaxy_install/tools/tool_panel_manager.py
@@ -41,8 +41,8 @@ class ToolPanelManager(object):
 
         # If shed_tool_conf file does not exist, create it.
         if not os.path.exists(shed_tool_conf_filename):
+            log.info('Creating shed tool config with default contents: %s', shed_tool_conf_filename)
             with open(shed_tool_conf_filename, 'w') as fh:
-                log.info('Creating shed tool config with default contents: %s', shed_tool_conf_filename)
                 fh.write(shed_tool_conf_dict['create'])
 
         tree, error_message = parse_xml(shed_tool_conf_filename)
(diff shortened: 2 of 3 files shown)


lib/galaxy/tool_shed/galaxy_install/tools/data_manager.py
```diff
@@ -60,16 +59,15 @@ def install_data_managers(self, shed_data_manager_conf_filename, metadata_dict,
             repository_tools_by_guid = {}
             for tool_tup in repository_tools_tups:
                 repository_tools_by_guid[tool_tup[1]] = dict(tool_config_filename=tool_tup[0], tool=tool_tup[2])
+
+            # If conf file does not exist, create it.
+            if not os.path.exists(shed_data_manager_conf_filename):
+                with open(shed_data_manager_conf_filename, 'w') as fh:
+                    log.info('Creating shed data manager config with default contents: %s', shed_data_manager_conf_filename)
```
nsoranzo:
```suggestion
                log.info('Creating shed data manager config with default contents: %s', shed_data_manager_conf_filename)
                with open(shed_data_manager_conf_filename, 'w') as fh:
```

mvdbeek (commented):
Does it actually fix https://github.com/galaxyproject/galaxy/pull/9779/ ?

I would think that you have to change https://github.com/galaxyproject/galaxy/blob/b8da698b239a11d93a3f78d382666fda3ec58c51/lib/galaxy/tools/toolbox/base.py#L184 as well.

> The current version of parse_xml handles all exceptions (good!),

That not what `parse_xml` is supposed to do, it should raise an execption. It's justy that we catch it (which  is a bug) https://github.com/galaxyproject/galaxy/blob/b8da698b239a11d93a3f78d382666fda3ec58c51/lib/galaxy/util/__init__.py#L249. 
~I think it'd be better if you could check why that doesn't work.
We should be doing the same in galaxy.util.tool_shed.xml_util.parse_xml, which restores the original behavior.I would say the proper fix is that galaxy.util.tool_shed.xml_util.parse_xml uses `galaxy.util.parse_xml` internally with an option to keep comments intact.~


For file related handling it's better to use try/except instead of os.path.exists (in python).

lib/galaxy/tool_shed/galaxy_install/tools/tool_panel_manager.py
```diff
@@ -24,31 +24,28 @@ def __init__(self, app):
...
-                    fh.write(shed_tool_conf_dict['create'])
-                tree, error_message = parse_xml(shed_tool_conf)
-            else:
-                log.error('Unable to load shed tool config: %s', shed_tool_conf)
-                raise
+
+        # If shed_tool_conf file does not exist, create it.
+        if not os.path.exists(shed_tool_conf_filename):
+            with open(shed_tool_conf_filename, 'w') as fh:
+                log.info('Creating shed tool config with default contents: %s', shed_tool_conf_filename)
```
nsoranzo:
```suggestion
            log.info('Creating shed tool config with default contents: %s', shed_tool_conf_filename)
            with open(shed_tool_conf_filename, 'w') as fh:
```

lib/galaxy/tool_shed/galaxy_install/tools/data_manager.py
```diff
@@ -60,16 +59,15 @@ def install_data_managers(self, shed_data_manager_conf_filename, metadata_dict,
...
-            try:
-                tree, error_message = parse_xml(shed_data_manager_conf_filename, check_exists=False)
-            except (OSError, IOError) as exc:
-                if exc.errno == errno.ENOENT:
-                    with open(shed_data_manager_conf_filename, 'w') as fh:
-                        fh.write(SHED_DATA_MANAGER_CONF_XML)
-                    tree, error_message = parse_xml(shed_data_manager_conf_filename)
-                else:
-                    raise
+            tree, error_message = parse_xml(shed_data_manager_conf_filename)
```
nsoranzo, line 70: My suggestion (for both this file and `lib/galaxy/tool_shed/galaxy_install/tools/tool_panel_manager.py`) would be to replace the use of `parse_xml()` from `galaxy.util.tool_shed.xml_util` to the one from `galaxy.util` (with `remove_comments=False, strip_whitespace=False` as needed).

Reasons:
- `error_message` is not checked and basically unused
- it would be better to let the exception bubble up if there is a syntax error in the XML file.

jdavcs: OK, trying this.

ic4f:
@mvdbeek 
> Does it actually fix #9779 ?

It does - and changing `toolbox/base.py` isn't needed. [This `except` clause](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/toolbox/base.py#L181) is never reached, regardless of how we fix/don't fix this. (I don't know why, haven't traced it that far)

> That not what `parse_xml` is supposed to do, it should raise an exception. It's just that we catch it (which is a bug) ...
> For file related handling it's better to use try/except instead of os.path.exists (in python).

Yes, you're right; `try/except` will prevent a race condition here. Still, I'm not sure that re-raising an error is the optimal solution here (as per #9781). Even though it's a simple fix, it doesn't look straightforward considering the logic it implements: if no file, create default. Maybe it is optimal; but I want to try to find an alternative first.

mvdbeek:
I agree that there are better ways to implement this, but that's refactoring that doesn't need to target 20.05 IMO.

ic4f:
Oh, good point. Let's use your fix in #9781 to fix the bug, and I'll work on this one for 20.09.

mvdbeek:
I think this is not needed anymore ? Can we close this @ic4f ?

ic4f:
> I think this is not needed anymore ? Can we close this/
Agreed, closing.

PR #9782, opened 2020-05-18
Add selenium test for user toolbox filters.

davebx:
This also adds a few filters to `examples.py.sample` and copies the file to `examples.py` before running tests.


jmchilton:
I love the test and coverage - I just worry that running run_tests.sh shouldn't change your production configuration and it does the way you've setup.

Can you place the example rules you want in a subdirectory of ``test/integration_selenium`` - maybe called ``filter_example`` and then setup ``config['toolbox_filter_base_modules'] = "integration_selenium/fitler_example"`` during the galaxy configuration section of an integration test version of this? I think that would be a better behaved test.

jmchilton:
Progress but I was hoping for a bit more isolation. Are you okay with b802b0d - I think this is what I was getting at?

PR #9790, opened 2020-05-19
Fix refactoring in linkifing libraries description

OlegZharkov:
fix refactoring typo in https://github.com/galaxyproject/galaxy/pull/9746/files#diff-b83467811fbbeac22dcb8303e80384fbR139

Just a little typo in https://github.com/galaxyproject/galaxy/pull/9746. Should be easy to merge ;)

PR #9791, opened 2020-05-19
[20.05] Backport linkifying library description

OlegZharkov:
Backporting commits from  https://github.com/galaxyproject/galaxy/pull/9746 and https://github.com/galaxyproject/galaxy/pull/9790 to 20.05

### DIFF
--- a/client/galaxy/scripts/mvc/library/library-librarylist-view.js
+++ b/client/galaxy/scripts/mvc/library/library-librarylist-view.js
@@ -136,7 +136,7 @@ var LibraryListView = Backbone.View.extend({
             $("#center").css("overflow", "auto");
         }
 
-        linkifyHtmlElements("[description=true]");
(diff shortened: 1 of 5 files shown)


dannon:
While I'm pretty sure it's 'safe' to backport; I'm not sure we should, since it's not a bugfix and it's primarily cosmetic?

OlegZharkov:
@dannon it's also a bug fix here https://github.com/galaxyproject/galaxy/pull/9791/files#diff-b83467811fbbeac22dcb8303e80384fbR134

before 
bug

now:
fix


It's not a major issue, but it looks a quite weird, when it blinks

mvdbeek:
I guess there's no harm getting this in. Thanks @OlegZharkov!

OlegZharkov:
Thank you  @dannon and @mvdbeek !

PR #9793, opened 2020-05-20
Modify loading of email white/blacklists

### LINKED ISSUE #9780: pgcleanup complains about disposable_email_whitelist
Error is:

```
ERROR _process_config(): CONFIGURATION ERROR: Can't open supplied whitelist file from path: /gpfs1/data/galaxy_server/galaxy-dev/config/disposable_email_whitelist.conf
```

jdavcs:
Fix #9780 + some refactoring + change of behavior:

Currently, an error is logged when galaxy tries to open a file based on the value of `whitelist_file` config option. This happens on each startup because the option has a default value specified in the schema, so it is always set. I think logging an error is misleading if the user did not set the option, in which case skipping loading the file is expected behavior.

I've considered several options, such as warning if the file is missing or raising an error only if the file is missing and the value is *not* the schema default. The behavior in this current version, I think, will be least surprising for the user: do not provide a default value (so the option will be set only if the user explicitly sets it), and raise (not log) an error if the specified file is not found. 

Specific changes:
- Refactor loading of whitelist and blacklist options: remove duplication
- Expect both files to be in <config_dir> (unless absolute path set). This was the case from the start, when `config/` [was hardcoded into the default value of `blacklist_file`](https://github.com/galaxyproject/galaxy/commit/3b5d8e58376d56a2c210aa043b385f5f1809cd5f#diff-aca5b6c2096078bc08a5a5af7b9b1bdaR270)
- Raise error if option is set but file is not found
- use `strip` instead of `rstrip` for each line (because why not)
- Remove default value from whitelist (least surprise)
- Minor edits to schema wording
- Make example file names consistent (remove "disposable" prefix: makes no sense in whitelist, and unnecessary in blacklist)
- Adjust unit test (value still tested, but not for a default path)

Other considerations:
- Should we provide [a graceful way for handling old settings](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L514)? 
- This is a minor bug with multiple changes to behavior. I don't think we should backport it to 20.05.

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -717,12 +717,8 @@ class GalaxyAppConfiguration(BaseAppConfiguration, CommonConfigurationMixin):
             }
 
     def _load_list_from_file(filepath):
-        try:
-            with open(filepath) as f:
-                return [line.strip() for line in f]
-        except IOError:
-            log.error("CONFIGURATION ERROR: Can't open supplied file: %s", filepath)
-            raise
+        with open(filepath) as f:
+            return [line.strip() for line in f]
 
     def _set_galaxy_infrastructure_url(self, kwargs):
         # indicate if this was not set explicitly, so dependending on the context a better default
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -475,24 +475,8 @@ class GalaxyAppConfiguration(BaseAppConfiguration, CommonConfigurationMixin):
         activation_email = kwargs.get('activation_email')
         self.email_from = self.email_from or activation_email
 
-        #  Get the disposable email domains blacklist file and its contents
-        self.blacklist_content = None
-        if self.blacklist_file:
-            self.blacklist_file = self._in_root_dir(self.blacklist_file)
-            try:
-                with open(self.blacklist_file) as f:
-                    self.blacklist_content = [line.rstrip() for line in f]
-            except IOError:
-                log.error("CONFIGURATION ERROR: Can't open supplied blacklist file from path: %s", self.blacklist_file)
-
-        #  Create whitelist file to accept only certain email domains
-        self.whitelist_content = None
-        if self.whitelist_file:
-            try:
-                with open(self.whitelist_file) as f:
-                    self.whitelist_content = [line.rstrip() for line in f]
-            except IOError:
-                log.error("CONFIGURATION ERROR: Can't open supplied whitelist file from path: %s", self.whitelist_file)
+        self.blacklist_content = self._load_list_from_file(self._in_config_dir(self.blacklist_file)) if self.blacklist_file else None
+        self.whitelist_content = self._load_list_from_file(self._in_config_dir(self.whitelist_file)) if self.whitelist_file else None
 
         self.persistent_communication_rooms = listify(self.persistent_communication_rooms, do_strip=True)
         # The transfer manager and deferred job queue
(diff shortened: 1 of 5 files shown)


lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -930,22 +930,32 @@ mapping:
...
         desc: |
           E-mail domains blacklist is used for filtering out users that are using
-          disposable email address during the registration.  If their address domain
-          matches any domain in the blacklist, they are refused the registration.
+          disposable email addresses at registration.  If their address domain matches any
+          domain on the list, they are refused registration.
 
-          Example value 'config/disposable_email_blacklist.conf'
+          If this option is set, galaxy will expect the file to be at the specified
+          location and will raise an error if the file is not found.
```
nsoranzo: Not sure this clarification is needed.

lib/galaxy/config/__init__.py
```diff
@@ -732,6 +716,14 @@ def _process_config(self, kwargs):
                 'filters': ['stack']
             }
 
+    def _load_list_from_file(filepath):
+        try:
+            with open(filepath) as f:
+                return [line.strip() for line in f]
+        except IOError:
+            log.error("CONFIGURATION ERROR: Can't open supplied file: %s", filepath)
+            raise
```
nsoranzo: I would simplify to:
```suggestion
        with open(filepath) as f:
            return [line.strip() for line in f]
```

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: I wasn't sure either. I'll remove it.

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -930,22 +930,26 @@ mapping:
...
 
-          Example value 'config/disposable_email_blacklist.conf'
+          Example value 'email_blacklist.conf'
+
+          The value of this option will be resolved with respect to <config_dir>.
 
       whitelist_file:
         type: str
-        default: disposable_email_whitelist.conf
-        path_resolves_to: config_dir
```
nsoranzo, line 941: `path_resolves_to` is not fully used yet, but why remove it?

jdavcs: `path_resolves_to` requires a default value. If we specify a default value, the attribute will be set, and an error will be raised (or logged in the current version) when galaxy tries to open that file on startup (that was the bug in #9780). I think the problem with having a default value set for a path is that there is no way to know whether a file is missing because the user did not intend to use it, or because they did, chose to stick with the default, and then forgot to provide the file. In the first case we don't want an error, in the second case we do.

nsoranzo: Maybe `path_resolves_to` shouldn't require a default value? I think the only place where a `None` value would be a problem is https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L241-L242 . Maybe these 2 lines could be removed?

nsoranzo (approved):

lib/galaxy/webapps/galaxy/config_schema.yml
jdavcs: Maybe; I'm not sure. There might be corner cases where a default would be needed. I'll check/test this. (I'm mostly off today, so I might not get to it until tomorrow.)

ic4f:
@nsoranzo I will follow-up on decoupling `default` value from `path_resolves_to` in a separate PR (unless I find any issues, in which case I'll ping you).
@dannon thanks for merging!

PR #9795, opened 2020-05-20
Babel peer dependency fix,

dannon:
Adds a seemingly undocumented babel peer dependency required for our particular version; to be dropped as an explicit dependency in https://github.com/galaxyproject/galaxy/pull/9739 (which fixes this in a core babel version upgrade)


dannon:
The one failing selenium test here is unrelated.

PR #9798, opened 2020-05-20
Improve mocha test output for rule DSL specification tests.

jmchilton:
Still a run of conformance tests without docs but this should hopefully make it slightly more clear what is happening.


mvdbeek (approved):
I think this needs a run of prettier, otherwise LGTM!

PR #9799, opened 2020-05-20
Reduce default dogpile logging for testing...

jmchilton:
... currently is a bit overwhelming and makes parsing test output more tricky.


mvdbeek:
Thanks @jmchilton!

PR #9800, opened 2020-05-20
InteractiveTool unit test fix

dannon:
(split off #9795, since it's totally unrelated to the changes there, I just happened to try to fix it in that branch)

This test broke recently with the release of chrome 83 (as I eventually figured out).  Previously, you could trigger clicks on the hidden checkbox just fine with the `.trigger("click")` approach, but that no longer works.  Using `setChecked` from vue-test-utils sets the value and propagates model updates, making this all work again.

### DIFF
--- a/client/galaxy/scripts/components/InteractiveTools/InteractiveTools.test.js
+++ b/client/galaxy/scripts/components/InteractiveTools/InteractiveTools.test.js
@@ -1,5 +1,6 @@
 import InteractiveTools from "./InteractiveTools";
 import { mount, createLocalVue } from "@vue/test-utils";
+import flushPromises from "flush-promises";
 import _l from "utils/localization";
 import Vue from "vue";
 import testInteractiveToolsResponse from "./testData/testInteractiveToolsResponse";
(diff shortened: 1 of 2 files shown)


dannon:
I have tracked this down, I think, to a change in chromium itself when upgrading from chrome 81 to chrome 83.  81 does not exhibit the failure, but 83 does.  Still digging.

client/galaxy/scripts/components/InteractiveTools/InteractiveTools.test.js
```diff
@@ -23,51 +24,40 @@ describe("ToolsView/ToolsView.vue", () => {
                 },
             },
         });
-        emitted = wrapper.emitted();
         axiosMock.onGet("/api/entry_points?running=true").reply(200, testInteractiveToolsResponse);
         axiosMock.onPost("/interactivetool/list").reply(200, { status: "ok", message: "ok" });
-        await Vue.nextTick();
-        await Vue.nextTick();
-        await Vue.nextTick();
+        await flushPromises();
```
jmchilton, line 29: This is exciting - I use this repeated nextTick hack all over. Thanks for cleaning this up.

dannon: @jmchilton Yeah, I was going to make a pass swapping them across the board separately.  nextTick is fine for waiting on internal component rendering, but for triggered events and axios, etc., we should just use flushPromises.

PR #9801, opened 2020-05-21
Make AuthN attribute optional for Azure and GCP.

VJalili:
Authentication information (e.g., Google OIDC tokens) are used to identify a user to a cloud service provider and get temporary authorization tokens. Hence making an OIDC-based authentication of a user a prerequisite for cloud authorization. 

However, the authentication tokens are used for AWS only, and Azure and Google Cloud Platform (GCP) do not yet support/utilize them. Therefore, this PR makes these tokens optional for Azure and GCP. Doing so, those Galaxy instances authenticating a user via username and password, would not need to setup OIDC to use Azure and GCP. 

Additionally, this PR added integration tests asserting if a cloudauthz config can be POSTed to the `api/cloud/authz` API for `azure` or `gcp` without specifying `authn` information. This test is a following up from a chat on gitter; for a quick ref: 

> @VJalili : I am getting TypeError: unhashable type: 'CloudAuthz' when trying to add a new cloudauthz via POST to /api/cloud/authz running on latest dev. 
I tracked it all the way to [the config variable](https://github.com/galaxyproject/galaxy/blob/b37629b56b80672b8c88b38ae844d429935f9930/lib/galaxy/model/__init__.py#L5813). Renaming this variable fixes it.

> @mvdbeek: @VJalili the config attribute is backed by a JSONType column (https://github.com/mvdbeek/galaxy/blob/dev/lib/galaxy/model/mapping.py#L161). This means the config attribute should be something that can be serialzed as JSON. It looks like the config is actually a CloudAuthz instance in your case. The code on the dev branch should make sure the config is a dict (https://github.com/mvdbeek/galaxy/blob/dev/lib/galaxy/webapps/galaxy/api/cloudauthz.py#L119), so I’m not sure what’s going on in your case, but as a first step I would suggest writing an integration test for this.

The test is sending a POST request with the following object as payload where `config` is a dictionary ([xref](https://github.com/VJalili/galaxy/blob/809ca2c737fc6fa6a45d96a179e8c149a912e581/test/integration/cloudauthz/test_cloudauthz.py#L37-L44)):

```json
payload = {
    "provider": "azure",
     "config": {
        "tenant_id": "tenant_id",
        "client_id": "client_id",
        "client_secret": "client_secret"
    }
}
```

However, the API receives the following payload where `config` is a list: 

```json
{
    "provider": "azure", 
    "config": ["tenant_id", "client_id", "client_secret"], 
    "key": "..."
}
```

ping @mvdbeek

### DIFF
--- a/lib/galaxy/managers/cloudauthzs.py
+++ b/lib/galaxy/managers/cloudauthzs.py
@@ -61,7 +61,7 @@ class CloudAuthzsSerializer(base.ModelSerializer):
             'user_id'      : lambda i, k, **c: self.app.security.encode_id(i.user_id),
             'provider'     : lambda i, k, **c: str(i.provider),
             'config'       : lambda i, k, **c: i.config,
-            'authn_id'     : lambda i, k, **c: self.app.security.encode_id(i.authn_id),
+            'authn_id'     : lambda i, k, **c: self.app.security.encode_id(i.authn_id) if i.authn_id else None,
             'last_update'  : lambda i, k, **c: str(i.last_update),
             'last_activity': lambda i, k, **c: str(i.last_activity),
             'create_time'  : lambda i, k, **c: str(i.create_time),
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -5963,7 +5963,7 @@ class CustosAuthnzToken(RepresentById):
         self.refresh_expiration_time = refresh_expiration_time
 
 
-class CloudAuthz(RepresentById):
+class CloudAuthz(object):
     def __init__(self, user_id, provider, config, authn_id, description=""):
         self.id = None
         self.user_id = user_id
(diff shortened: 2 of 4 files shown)


jmchilton:
New test fails:

```
test/integration/cloudauthz/test_cloudauthz.py::DefineCloudAuthzTestCase::test_post_cloudauthz_without_authn FAILED [ 95%]
```

VJalili:
Yes, I wrote this test to show there is something broken, yet I don't have good clues what/where. 

The `config` variable is expected to be a dictionary, but the API receives a list instead. That is the cause of the error, but not sure why the dict is converted to a list (which is the list of dictionary keys).

mvdbeek (approved):
Thanks for the test @VJalili. The trick is that you either also need `__hash__` if you add `__eq__`, or you drop `__eq__` which I've chosen for here. I think this is ok now, but please have a look. I'm dropping the milestone for now, since this is clearly a bug we can backport that to any release you like.

VJalili:
@mvdbeek Thank you, I guess the [`equals`](https://github.com/galaxyproject/galaxy/pull/9801/files#diff-5805bf58afdc31ee4a9c4e788301b11dL5986) method would suffice for now: 

https://github.com/galaxyproject/galaxy/blob/29aad8212365c56ddec947adb46eac6d9a89762b/lib/galaxy/webapps/galaxy/api/cloudauthz.py#L136

I'm not sure when this was broken, but I guess targeting the 20.01 release would be reasonable.

mvdbeek:
Sounds good, will open the PR later.

Issue #8540, opened 2019-08-29
Failed job when using collections in a workflow makes the following output collection unusable


Delphine-L:
When I run a workflow on a collection, and one of the steps had failed jobs, the next output collections have paused job. 
These collections then cannot be used in collection manipulation tools such as filter failed jobs or flatten collections... It means I need to filter out the failed jobs at the point of failure and rerun all the following steps. 
Would it be possible to be able to filter out paused jobs too?

mvdbeek:
If you re-run the failed job you also have the option to replace the original item (assuming a re-run would fix the problem) and continue from there. It's simple enough to implement this for the filter failed tool, I'll open a PR later.

cgirardot:
@mvdbeek I am also in the case where 3 out 5 jobs failed in my collection. Input is a paired-end dataset collection and I can't see/find the option you are mentioning after clicking on `re-run`. 

Here is what I see

<img width="1725" alt="Screenshot 2019-09-11 at 16 44 03" src="https://user-images.githubusercontent.com/10990795/64707525-6ded1800-d4b3-11e9-8b7f-78742b2afb47.png">

if I hit `Execute`, I get : 

<img width="686" alt="Screenshot 2019-09-11 at 16 44 09" src="https://user-images.githubusercontent.com/10990795/64707575-7c3b3400-d4b3-11e9-8060-6a586b5e9c0d.png">

Any idea why ? Note this is on our local server, running 19.05

mvdbeek:
Resume dependencies from this job has the same effect, it's just named differently when there are jobs paused that depend on this dataset. Make sure you don't change the input dataset selection, "cannot create multiple jobs when remapping existing jobs" means that you are mapping over the tool (which creates multiple jobs) instead of just running the same job hat failed.

cgirardot:
I did not change anything in the form but the "Resume dependencies". Also I am pretty sure I saw the option you mentioned (replace the original item) in other context. 
By default, the input form looks like : 
<img width="1937" alt="Screenshot 2019-09-11 at 17 13 09" src="https://user-images.githubusercontent.com/10990795/64710014-89f2b880-d4b7-11e9-8712-5cee3262f136.png">

which looks wrong (the 469 collection is not the one used as input of this job), so I changed the collection to the right one but got the same error. 

Should I change the input type to select the datasets directly like this ? 
[...]

Issue #4860, opened 2017-10-25
rerun/tool info: oddity with data set collections


bernt-matthias:
I run a tool (eg trim galore or hisat2) on a (paired) data set collection. When I open the tool form via the rerun button of a data set the input is not set correctly to the collection that was actually used as input, but the topmost fastq collection in my history is chosen. 

Also the info page ("View Details") does not show which data set was used as input, but only two rows: 

Single-end or paired-end reads? paired_collection 
Paired Collection No dataset collection.

So currently the only way to retrace the input seems to be that name of the data set collection that is set by the tool.

mvdbeek:
I think the empty rows in "View Details" should be fixed by https://github.com/galaxyproject/galaxy/pull/4665/commits/19818cb380889e584741caf4a78462e0502f30c1, which is only in dev for the moment, the rerun behavior of the tool form I haven't seen yet. In general we should implement the rerun button for dataset collections ... .

bernt-matthias:
I'm running on dev which I updated a day ago (https://github.com/galaxyproject/galaxy/commit/19818cb380889e584741caf4a78462e0502f30c1 is included).

mvdbeek:
And the job has been run on that version as well ? Could you add a screenshot to be sure we're talking of the same problem ?

bernt-matthias:
Exactly. But the input collections have been created before the update.

mvdbeek:
That shouldn't be a problem, I think.

blankenberg:
Seeing similar behavior here.
 - Create paired dataset collection of fastq files
 - Run e.g. Trimmomatic (Creates 2 output collections: paired and unpaired)
 - Run bwa-mem on Paired output of Trimmomatic

![image](https://user-images.githubusercontent.com/1231190/37713881-90d01d18-2cee-11e8-847f-40866682f275.png)

info on one of the bwa outputs in the collection -- 'No dataset collection':
![image](https://user-images.githubusercontent.com/1231190/37714066-f4760cf6-2cee-11e8-9453-aeb20a0ca13b.png)

rerun on same bwa output -- 'unpaired' selected:
![image](https://user-images.githubusercontent.com/1231190/37714099-101482a8-2cef-11e8-9f1d-a93bdf5ffa97.png)


Since it has no value for input paired collection, it is selecting the 'newest' paired collection. 

[...]

Issue #8977, opened 2019-11-12
rerun and replace dataset in collection fail to coordinate multiple outputs


qiagu:
When running a tool on a collection of inputs, outputs will be a collection. If each job produces multiple outputs, the results will be multiple collections. In the latter scenario, rerun and replace a failed a dataset fail to make other outputs replaced in their collection. I'm not sure my statement is generalized right, since the other output failed me was a collection from `discover_datasets`.

bgruening:
@qiagu can this be closed? Please see: https://github.com/galaxyproject/galaxy/pull/9802

Issue #8403, opened 2019-08-02
What limits usability of collections?


mvdbeek:
I would like to see wider adoption of collections among Galaxy users, and I'm trying to understand how we can improve collections (or the documentation surrounding collections) to reach that goal. So if you've tried them recently and were stuck at any point and felt that going back to regular datasets was the better option please let us know!

ThomasWollmann:
- Visual feedback of tool outputs into collections is unsatisfactory. You do not know if just collection content is fetching or results still have to be computed. Also job logs, rerun and these features are not available.
- handling of collections in the tool wrappers is very cumbersome, since discover_datasets and datatypes are messy.
- https://github.com/bgruening/docker-jupyter-notebook/issues/11
- Support for complex dataset filtering would be nice
- Support for sending subsets of nested collections to a workflow would be awesome.
- https://toolshed.g2.bx.psu.edu/repository/view_repository?sort=name&operation=view_or_manage_repository&id=c30c030673c90378 does not support creation of nested collections from zipped files. I do not know how to fix it, since I can not get the "recurse" option not working.

mmiladi:
* collection is not well scalable. My browser has difficulties with opening and searching collections of large size.
* takes too much (twice? unnecessary?) hidden entries. histories in the "show hidden" mode are super lengthy & unmanageable. A "show hidden but not the collection single entries" button would be great! :)
* restart an entry from an output collection: There seems to be no way to automatically switch between restarting single entry to restarting the whole collection (at least idk)
* Different tags for different entries of a collection are not shown.
* A sustainable and reliable way to make collection out of a zipped input would be great!

hexylena:
- failed collections which contain no datasets do not offer a way to view or report errors

ThomasWollmann:
> * A sustainable and reliable way to make collection out of a zipped input would be great!

Checkout: https://toolshed.g2.bx.psu.edu/repository/view_repository?sort=name&operation=view_or_manage_repository&id=c30c030673c90378

bernt-matthias:
For me its often the automatic naming of data sets, which mainly comes from the fact that with `multiple="true"` collections are not treated as collections but as a bag of data sets. Take for instance the following history that I created in the (short) mothur tutorial from the GTN. 

![Screenshot from 2019-08-06 11-30-37](https://user-images.githubusercontent.com/25689525/62532559-f2a99e00-b844-11e9-932a-2fff9b5f2d40.png)

[...]

PR #9802, opened 2020-05-21
Allow tool submission with DatasetCollectionElements / make build_for_rerun consumable by API

mvdbeek:
This is necessary for re-running failed mapped-over jobs.
The most obvious example is rerunning a job that consumes a `paired` collection where the initial job was mapped over from a `list:paired` collection, but applies to all situations where we map nested collections over collection input.

This also makes the `state_inputs` part of `jobs/<id>/build_for_rerun` a valid input for submission via the tools api endpoint.

### DIFF
--- a/client/galaxy/scripts/mvc/ui/ui-select-content.js
+++ b/client/galaxy/scripts/mvc/ui/ui-select-content.js
@@ -451,14 +451,14 @@ const View = Backbone.View.extend({
     /** Library datasets are displayed and selected together with history datasets,
         Dataset collection elements are displayed together with history dataset collections **/
     _patchValue: function (v) {
-        const patch_to = { ldda: "hda", dce: "hdca" };
+        const patchTo = { ldda: "hda", dce: "hdca" };
         if (v.values) {
             _.each(v.values, (v) => {
                 this._patchValue(v);
             });
-        } else if (patch_to[v.src]) {
+        } else if (patchTo[v.src]) {
             v.origin = v.src;
-            v.src = patch_to[v.src];
+            v.src = patchTo[v.src];
             v.id = `${v.origin}${v.id}`;
         }
     },
--- a/client/galaxy/scripts/mvc/ui/ui-select-content.js
+++ b/client/galaxy/scripts/mvc/ui/ui-select-content.js
@@ -122,7 +122,7 @@ const Configurations = {
     ],
 };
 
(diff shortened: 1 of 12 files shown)


mvdbeek:
If I understand https://github.com/galaxyproject/galaxy/issues/8977 correctly, this PR should fix that, it should fix https://github.com/galaxyproject/galaxy/issues/8540#issuecomment-530414495, should allow https://github.com/galaxyproject/galaxy/issues/8403#issuecomment-539086605, and fix https://github.com/galaxyproject/galaxy/issues/4860#issue-268414248.

@nsoranzo @jmchilton do you think this is the right way to go? Should I add another endpoint (`/api/jobs/<id>/rerun`) that only reruns jobs (so no parameter changes possible)? I guess this endpoint wouldn't really be consumable by the UI, but that endpoint wouldn't need all the other stuff that is currently consumed by the client (and the client doesn't actually consume `state_inputs`, so we could remove `state_inputs` from `Tool.to_dict`) and it would make sense in the bioblend context, I think.

I think allowing DCE input would also be nice for @guerler's Browse Datasets button, since we could maintain the element identifier at that step (https://github.com/galaxyproject/galaxy/issues/8149#issuecomment-502017864).

mvdbeek:
Alright, I think this is ready for review. I spun off some improvements that follow from tracking job input DCEs in https://github.com/mvdbeek/galaxy/tree/job_cache_improvements.

client/galaxy/scripts/mvc/ui/ui-select-content.js
```diff
@@ -444,11 +448,17 @@ const View = Backbone.View.extend({
...
     },
 
-    /** Library datasets are displayed and selected together with history datasets **/
+    /** Library datasets are displayed and selected together with history datasets,
+        Dataset collection elements are displayed together with history dataset collections **/
     _patchValue: function (v) {
-        if (v.src == "ldda") {
-            v.src = "hda";
-            v.origin = "ldda";
+        const patch_to = { ldda: "hda", dce: "hdca" };
```
dannon: Minor convention note; camelCase in js, not snake_case like python.

mvdbeek: Thanks, I've updated the commit!

assuntad23 (approved):

jmchilton:
Sorry for the delay in merging this.

PR #9803, opened 2020-05-21
[WIP] Implement nested collection inputs for workflow tests

jmchilton:
This ties together seemingly random threads to get this done and tested within Galaxy.

There is this function in ``galaxy.tool_util.cwl.util`` called ``galactic_job_json`` that parses a CWL-style "job" (the inputs for a tool or a workflow execution as a JSON object) and parses it into a representation for Galaxy - uploading inputs to Galaxy as needed. The interaction with the API was originally abstracted out as two functions so that the Galaxy test code (in the CWL branch) and the Planemo driver code could interact with the API in different ways. Planemo used bioblend (loosely) and the Galaxy code used the test framework's "populators" abstractions.

This PR creates a new a layer above ``galactic_job_json`` that has some extension points that the Planemo/bioblend layer can take advantage of but by default will just use the galaxy_interactor stuff already being used by tool tests using galaxy-tool-util. So this new layer doesn't depend on test "populators" or bioblend and should allow important code to be shared between the Galaxy test framework, the CWL branch of Galaxy, and Planemo. Having to update the code in both places was a significant problem with the CWL branch.

Here is a commit demonstrating how it simplifies the CWL branch (jmchilton@d9e774f). I don't have a matching commit for Planemo yet but the code is located here https://github.com/galaxyproject/planemo/blob/master/planemo/galaxy/activity.py#L181 that will be simplified after this commit is utilized.

This main big commit here establishes that layer, refactors a couple of tool tests to use it, and then extends it to implement nested collections.

Another positive upshot of this commit is there is now some in core test coverage of ``galactic_job_json`` - an important library function.

There are some other smaller commits here to setup some stuff in galaxy-tool-util to be shared between Galaxy testing populators and Planemo.

### DIFF
--- a/lib/galaxy/tool_util/client/staging.py
+++ b/lib/galaxy/tool_util/client/staging.py
@@ -1,0 +1,196 @@
+"""Client for staging inputs for Galaxy Tools and Workflows.
+
+Implement as a connector to serve a bridge between galactic_job_json
+utility and a Galaxy API library.
+"""
+import abc
+import json
+import logging
+import os
+
+import six
+
+from galaxy.tool_util.cwl.util import (
+    DirectoryUploadTarget,
+    FileLiteralTarget,
+    FileUploadTarget,
+    galactic_job_json,
+    path_or_uri_to_uri,
+)
+
+log = logging.getLogger(__name__)
+
+UPLOAD_TOOL_ID = "upload1"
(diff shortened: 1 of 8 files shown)


jmchilton:
Broken because it doesn't implement tags - that was next on my todo list anyway in terms of testing syntax so I will close this and reopen once tags are implemented.

PR #9805, opened 2020-05-21
Higlass InteractiveTool update

astrovsky01:
This version of higlass allows all current standard datatypes to be uploaded on launch. The only modification remaining is the names of the files, which each show up with extra extensions appended onto the filenames


dannon:
Going to go ahead and merge this in so we can continue to iterate on it within this release.  Thanks @astrovsky01

PR #9806, opened 2020-05-22
switch branch of IUC CI earlier in release process

bernt-matthias:
I guess switching devteam tools to github actions might be a good idea.


scripts/bootstrap_history.py
```diff
@@ -266,7 +267,7 @@
 
 - [ ] **Ensure Tool Tests use Latest Release**
 
-    - [ ]  Update GALAXY_RELEASE in https://github.com/galaxyproject/tools-iuc/blob/master/.travis.yml#L6
+    - [ ]  Update GALAXY_RELEASE in https://github.com/galaxyproject/tools-iuc/blob/master/.github/workflows/pr.yaml
```
mvdbeek: Maybe move the `**Ensure Tool Tests use Latest Release**` section under `**Deploy and Test Release**` ?

bernt-matthias: I just merged it. I think updating both workflows is the simplest solution. But one could argue the the PR workflow should be updated after the release in order to spare tool developers from potential bugs in pre-release Galaxy.

bernt-matthias:
Thanks for merging @jmchilton last piece missing is then this one: https://github.com/galaxyproject/tools-devteam/pull/558

PR #9807, opened 2020-05-22
Add FlowCytometry and MetaCyto datatypes

pcm32:
This PR adds datatypes for flow cytometry tools from the Immport Galaxy setup.


bernt-matthias:
Tests for the sniffers would be nice. Like here https://github.com/pcm32/galaxy/blob/a871cf36f2578c0dc8f9c7a8c35a89c1fe56b791/lib/galaxy/datatypes/binary.py#L116

lib/galaxy/datatypes/flow.py
```diff
@@ -0,0 +1,435 @@
...
+        except Exception:
+            return "Binary RDS flowFrame (%s)" % (data.nice_size(dataset.get_size()))
+
+    def sniff(self, filename):
+        """
+        Checking if the file is a flowFrame R object.
+
+        For this to work, need to have install checkFlowframe.R via bioconda
+        conda install ig-checkflowtypes
+        """
```
bernt-matthias: Is this also possible via pip? Then this could be added to requirements?

lib/galaxy/datatypes/flow.py
```diff
@@ -0,0 +1,435 @@
...
+        try:
+            return dataset.peek
+        except Exception:
+            return "Text Flow file (%s)" % (data.nice_size(dataset.get_size()))
+
+    def sniff(self, filename):
+        """Quick test on file formatting and values"""
+        with open(filename, "r") as f:
+            f.readline()
+            values = f.readline().strip().split("\t")
```
bernt-matthias: This might be to generic to be sniffable.

lib/galaxy/datatypes/flow.py
```diff
@@ -0,0 +1,435 @@
...
+                return False
+        except Exception:
+            False
+
+    def get_mime(self):
+        """Returns the mime type of the datatype"""
+        return 'application/octet-stream'
+
+
+Binary.register_sniffable_binary_format('flowset', 'flowset', FlowSet)
```
bernt-matthias: Shouldn't this be done via the sniffers section in datatype.conf?

lib/galaxy/datatypes/flow.py
```diff
@@ -0,0 +1,435 @@
...
+        """Quick test on file formatting and values"""
+        with open(filename, "r") as f:
+            population = f.readline().strip().split("\t")[0]
+            if population != "Population":
+                return False
+            values = f.readline().strip().split("\t")
+            for vals in values:
+                if not is_number(vals):
+                    return False
+            return True
```
bernt-matthias: Same as flowclustered?

lib/galaxy/datatypes/metacyto.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def sniff(self, filename):
+        """Quick test on file headings"""
+        with open(filename, "r") as f:
+            # last one == fraction
+            headers = f.readline().strip().split("\t")
+            if headers[-1] != "fraction":
+                return False
+            hdrs = ["fcs_files", "cluster_id", "label", "fcs_names"]
+            for h in hdrs:
```
bernt-matthias: Maybe use Set intersection?

lib/galaxy/datatypes/flow.py
```diff
@@ -0,0 +1,435 @@
...
+            last_col = f.readline().strip().split("\t")[-1]
+            if last_col != "Percentage_stdev":
+                return False
+            values = f.readline().strip().split("\t")
+            for vals in values:
+                if not is_number(vals):
+                    return False
+            return True
+
+    def get_mime(self):
```
bernt-matthias: Is this a valid mime type? Why overwrite the Foo from tabular?

pcm32:
Disclaimer: I only inherited this, I didn't write the code, so I'm not an expert on it.

lib/galaxy/datatypes/flow.py
pcm32: I have now removed the need for this, to avoid adding dependencies. Thanks!

lib/galaxy/datatypes/flow.py
pcm32: what do you mean by too generic here? could you elaborate please? thanks!

lib/galaxy/datatypes/flow.py
pcm32: yes, this is done as well in datatype conf.xml, is this redundant then? This is the first time I mess so closely with datatypes, so I'm a bit ignorant on this.

lib/galaxy/datatypes/flow.py
pcm32: Not sure why the person who wrote this did it, I agree it is unnecessary and I have removed it, thanks for spotting it!

lib/galaxy/datatypes/flow.py
pcm32: I removed all RData ones and just left them as subclass of Rdata in the datatype.conf.xml file, so this one in particular is gone, but there are others in the file (like the fcs one).

lib/galaxy/datatypes/metacyto.py
pcm32: Thanks, I have changed all of these for set intersections as suggested.

lib/galaxy/datatypes/flow.py
bernt-matthias: I think the chance of false positive hits of the sniffer is to high. Essentially it returns `true` for all tsv files with only numbers in it.

lib/galaxy/datatypes/flow.py
bernt-matthias: I'm not sure

lib/galaxy/datatypes/flow.py
pcm32: well, apparently a flowtext file seems to be just that. I could in turn ask pandas to read the table and ask for summaries on all columns, but it will be just that, numbers on every row besides the header. I'm guessing is not worth it for an sniffer.

lib/galaxy/datatypes/flow.py
pcm32: This one also request in the previous lines to have `Population_ID` as first token of the header.

pcm32:
I have added test to sniffers for every datatypes I could find examples for (as I mentioned before, I sort of inherited this, so I don't have access to a wealth of files for these datatypes).

pcm32:
Apparently sniffer tests fail due to assertions like https://github.com/galaxyproject/galaxy/pull/9807/checks?check_run_id=1858944753#step:12:960, where the assertion expect the format to be tabular, but then it is a value overridden (file_ext) in the datatype class. Could someone advice on what I am supposed to do here? Not override that value? Or change it somewhere else as well? Thanks!

martenson:
> Apparently sniffer tests fail due to assertions like https://github.com/galaxyproject/galaxy/pull/9807/checks?check_run_id=1858944753#step:12:960, where the assertion expect the format to be tabular, but then it is a value overridden (file_ext) in the datatype class. Could someone advice on what I am supposed to do here? Not override that value? Or change it somewhere else as well? Thanks!

This might be caused by the sniffing order, where the file is sniffed as tabular because your new more specific tabular datatype has undefined sniffing order, so the sniffer never gets there.

lib/galaxy/datatypes/flow.py
```diff
@@ -0,0 +1,312 @@
+# -*- coding: utf-8 -*-
+######################################################################
+#                  Copyright (c) 2016 Northrop Grumman.
+#                          All rights reserved.
+######################################################################
+
```
mvdbeek:
```suggestion
```

lib/galaxy/datatypes/flow.py
```diff
@@ -0,0 +1,312 @@
...
+            return False
+
+        return True
+
+    def get_mime(self):
+        """Returns the mime type of the datatype"""
+        return 'application/octet-stream'
+
+
+Binary.register_sniffable_binary_format("fcs", "fcs", FCS)
```
mvdbeek:
```suggestion
```

lib/galaxy/datatypes/flow.py
mvdbeek: Yeah, if this can't be made more specific we need to remove the sniffing

lib/galaxy/datatypes/flow.py
```diff
@@ -0,0 +1,312 @@
...
+        """
+        with open(filename, "r") as f:
+            population = f.readline().strip().split("\t")[-1]
+            if population != "Population":
+                return False
+            values = f.readline().strip().split("\t")
+            for vals in values:
+                if not is_number(vals):
+                    return False
+            return True
```
mvdbeek: Same here, that's not specific enough

lib/galaxy/datatypes/metacyto.py
```diff
@@ -0,0 +1,99 @@
+# -*- coding: utf-8 -*-
+######################################################################
+#                  Copyright (c) 2016 Northrop Grumman.
+#                          All rights reserved.
+######################################################################
+
```
mvdbeek:
```suggestion
```

mvdbeek (commented):
```
__________________ [doctest] galaxy.datatypes.flow.FCS.sniff ___________________
049         Checking if the file is in FCS format. Should read FCS2.0, FCS3.0
050         and FCS3.1
051 
052         Based on flowcore:
053         https://github.com/RGLab/flowCore/blob/27141b792ad65ae8bd0aeeef26e757c39cdaefe7/R/IO.R#L667
054 
055         Try to guess if the file is a fcs file.
056         >>> from galaxy.datatypes.sniff import get_test_fname
057         >>> fname = get_test_fname('Accuri_C6_A01_H2O.fcs')
058         >>> FCS().sniff(fname)
Expected:
    True
Got:
    False
```
I don't think the sniffer works as expected. No need to mess with the sniff order, base classes are run last.

lib/galaxy/datatypes/flow.py
pcm32: But this is asking for a specific form in the heading. See line 153.

lib/galaxy/datatypes/flow.py
pcm32: So should I just remove the sniffer method or the whole class and just add it as a sub datatype in the conf?

lib/galaxy/datatypes/flow.py
pcm32: But is true that it could recognize other files not meant to be this.

lib/galaxy/datatypes/flow.py
pcm32: Please let me know if I should remove this as well then @mvdbeek. Thanks

pcm32:
The binary sniffer is fixed, but it is still failing in this sort of assertions:
```
        # Check that correct datatype has been detected
        file_ext = dataset['file_ext']
        if 'false' in test_data.path:
            # datasets with false in their name are not of a specific datatype
            assert file_ext != PARENT_SNIFFER_MAP.get(expected_file_ext, expected_file_ext)
        else:
>           assert file_ext == PARENT_SNIFFER_MAP.get(expected_file_ext, expected_file_ext)
E           AssertionError: assert 'tabular' == 'flowclr'
E             - flowclr
E             + tabular
```

this is the case for all the tabular inheriting cases. How do you deal with that? Thanks!

pcm32:
@mvdbeek I have removed most of the specific classes for all the tabular formats on flow and switched those to tabular on the sample datatype config. Are the ones in metacyto specific enough or should I remove them too? They are asking for specific headers to be available, but not sure if that is considered specific enough. I don't mind, happy to keep or set to tabular, just let me know. I suspect that the failures are now due to the sniffer order, as I have fixed the binary sniffer test that you pointed out. But not sure how to fix the error in my previous comment (which I guess will come up with the metacyto ones). Thanks!

pcm32:
Currently failing with things like https://github.com/galaxyproject/galaxy/pull/9807/checks?check_run_id=1913938914#step:12:823

lib/galaxy/datatypes/flow.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def is_number(s):
+    try:
+        float(s)
+        return True
+    except ValueError:
+        return False
+
+
```
nsoranzo: This function seems unused.
```suggestion
```

lib/galaxy/datatypes/metacyto.py
```diff
@@ -0,0 +1,79 @@
...
+            return "MetaCyto Cluster Stats"
+
+    def sniff_prefix(self, file_prefix):
+        """Quick test on file headings"""
+        if file_prefix.startswith("fcs_files\tcluster_id\tlabel\tfcs_names"):
+            header_line = file_prefix.string_io().readline()
+            if header_line.strip().split("\t")[-1] == 'fraction':
+                return True
+            elif file_prefix.truncated and file_prefix.string_io().read() == header_line:
+                return True
```
nsoranzo, line 23:
```suggestion
                return True
        return False
```

lib/galaxy/datatypes/flow.py
```diff
@@ -0,0 +1,80 @@
...
+                # we only need to check ioffs 2 to 5
+                for _i in range(4):
+                    coffs.append(f.read(8).decode("utf8", errors="ignore"))
+
+            ioffs = [float(version)] + [int(x) for x in coffs]
+            for ioff in ioffs:
+                if ioff is None or math.isnan(ioff):
+                    return False
+        except Exception:
+            return False
```
nsoranzo:
```suggestion
        with open(filename, 'rb') as f:
            # assume 1 byte = 1 char
            version = f.read(6).decode("utf8", errors="ignore")
            if version not in ["FCS2.0", "FCS3.0", "FCS3.1"]:
                return False
            version = version.replace("FCS", "")
            tmp = f.read(4).decode("utf8", errors="ignore")
            if tmp != "    ":
                return False
            coffs = []
            # we only need to check ioffs 2 to 5
            for _i in range(4):
                coffs.append(f.read(8).decode("utf8", errors="ignore"))

        ioffs = [float(version)] + [int(x) for x in coffs]
        for ioff in ioffs:
            if ioff is None or math.isnan(ioff):
                return False
```

lib/galaxy/datatypes/flow.py
```diff
@@ -0,0 +1,51 @@
...
+        Checking if the file is in FCS format. Should read FCS2.0, FCS3.0
+        and FCS3.1
+
+        Based on flowcore:
+        https://github.com/RGLab/flowCore/blob/27141b792ad65ae8bd0aeeef26e757c39cdaefe7/R/IO.R#L667
+        """
+        content = file_prefix.contents_header_bytes[:42].decode()
+        version = content[:6]
+        assert version in ["FCS2.0", "FCS3.0", "FCS3.1"]
+        assert content[6:10] == '    '
```
nsoranzo: It's better to return `False` instead of using `assert`s, which can be disabled via `python -O`.

mvdbeek: Wow, I did not know that and this seems fully insane 😆

lib/galaxy/datatypes/metacyto.py
mvdbeek: Seems like more code without any benefit ? I don't think we consistently do this ?

nsoranzo: We have over 200 sniffers, many returning `True` unless some conditions are met, and many others returning `False` unless some conditions are met. I just think explicit is better than implicit when skimming through the code.

pcm32:
Thanks @mvdbeek and @nsoranzo for the extra help. Let me know if I need to remove any other sniffer that is not specific enough.

pcm32:
Could this be merged or would you like me to remove / change something else (maybe there are still certain unspecific sniffing?). Thanks for your help.

lib/galaxy/datatypes/flow.py
mvdbeek: @pcm32 can you address this

pcm32: Ok, done now! Hopefully I didn't introduce any linting issues.

lib/galaxy/datatypes/flow.py
```diff
@@ -0,0 +1,53 @@
...
+            return False
+        if content[6:10] != '    ':
+            return False
+        # we only need to check ioffs 2 to 5
+        int(content[10:42].replace(' ', ''))
+        return True
+
+    def get_mime(self):
+        """Returns the mime type of the datatype"""
+        return 'application/octet-stream'
```
nsoranzo:
```suggestion
```

lib/galaxy/datatypes/metacyto.py
```diff
@@ -0,0 +1,80 @@
...
+            dataset.blurb = nice_size(dataset.get_size())
+        else:
+            dataset.peek = 'file does not exist'
+            dataset.blurb = 'file purged from disk'
+
+    def display_peek(self, dataset):
+        try:
+            return dataset.peek
+        except Exception:
+            return "MetaCyto Cluster Stats"
```
nsoranzo: Isn't it better to just use the `set_peek()`/`display_peek()` methods of the `Tabular` parent class?

lib/galaxy/datatypes/metacyto.py
```diff
@@ -0,0 +1,80 @@
...
+            elif file_prefix.truncated and file_prefix.string_io().read() == header_line:
+                return True
+        return False
+
+
+class mClrList(Tabular):
+    """Class describing a list of cluster definitions used for MetaCyto analysis"""
+    file_ext = "metacyto_clr.txt"
+
+    def set_peek(self, dataset, is_multi_byte=False):
```
nsoranzo: Same as above. If you can drop `set_peek()` and `display_peek()`, then this can just be defined with `type="galaxy.datatypes.tabular:Tabular" subclass="true"` in `lib/galaxy/config/sample/datatypes_conf.xml.sample` .

lib/galaxy/datatypes/metacyto.py
```diff
@@ -0,0 +1,80 @@
...
+            return dataset.peek
+        except Exception:
+            return "MetaCyto Cluster Definitions List"
+
+
+class mSummary(Tabular):
+    """Class describing the summary table output by MetaCyto after FCS preprocessing"""
+    file_ext = "metacyto_summary.txt"
+
+    def set_peek(self, dataset, is_multi_byte=False):
```
nsoranzo: Same as above (but this have a sniffer, so keep the class!)

lib/galaxy/config/sample/datatypes_conf.xml.sample
```diff
@@ -313,6 +313,23 @@
...
+    <datatype extension="flowmfi" type="galaxy.datatypes.tabular:Tabular" mimetype="text/tab-separated-values" display_in_upload="true" description="A Flow MFI file with a '.flowmfi' file extension." />
+    <datatype extension="flowstat1" type="galaxy.datatypes.tabular:Tabular" mimetype="text/tab-separated-values" display_in_upload="true" description="A Flow Stats file with a '.flowstat1' file extension." />
+    <datatype extension="flowstat2" type="galaxy.datatypes.tabular:Tabular" mimetype="text/tab-separated-values" display_in_upload="true" description="A Flow Stats file with a '.flowstat2' file extension." />
+    <datatype extension="flowstat3" type="galaxy.datatypes.tabular:Tabular" mimetype="text/tab-separated-values" display_in_upload="true" description="A Flow Stats file with a '.flowstat3' file extension." />
+    <datatype extension="flowscore" type="galaxy.datatypes.tabular:Tabular" mimetype="text/tab-separated-values" display_in_upload="true" description="A Flow Score file with a '.flowscore' file extension." />
```
nsoranzo: These are all missing `subclass="true"`.

lib/galaxy/datatypes/metacyto.py
pcm32: Sure, I can remove these.

lib/galaxy/config/sample/datatypes_conf.xml.sample
pcm32: Thanks, done!

lib/galaxy/datatypes/metacyto.py
pcm32: Thanks, done!

lib/galaxy/datatypes/metacyto.py
pcm32: Ok, done!

pcm32:
Thanks again for the valuable reviews guys, hopefully I didn't introduce any linting issues.

nsoranzo (approved):

PR #9808, opened 2020-05-22
Fix bug: add self arg to method

jdavcs:


dannon:
xref https://github.com/galaxyproject/galaxy/pull/9793

dannon:
Failing test is unrelated.

PR #9809, opened 2020-05-22
Allow uploading directly from the tool form

jmchilton:
Implementation of https://github.com/galaxyproject/galaxy/issues/9110.

uploadform480

Video demonstrates two options for how to reach the upload dialog - embedding this right in the data selection row and placing it as another option in the data dialog. I suspect for the general Galaxy user experience, yet another button in the data selector row is too much so the second variant here might be a better default for stock Galaxy experiences. However, keeping the option for an embedded button for stand-alone tools-as-webapps and simplified workflows UX that want to hide the concepts of histories and libraries entirely seems like a good idea - it reduces the number of clicks and exposure to Galaxy concepts required to get going with a workflow run or tool execution. So I've hidden the extra button behind a new config option ``enable_upload_from_form_button``.

In addition to providing a shorthand for getting data quickly into tool form inputs, another goal is to simplify the upload experience based on the context provided by the form input parameters. For instance in the demo above, this is a single input - so collection and rules uploads are disabled and when a single file is chosen the options for adding more are not available. Not in the demo, but the browser file picker also does not allow multiple files to be chosen for instance.

In addition to handling "multiple vs single" logic, the form also now can be supplied with a set of "formats" (or short extensions) and the display will adapt to them. If the input can't be a composite file for instance that tab no longer available. The file types in the drop down are also restricted to "auto" and subtypes of the specified formats. This functionality uses the recently merged #9955 that refactored the client datatype comparison logic from the workflow editor for reuse in exactly this use case.

### DIFF
--- a/lib/galaxy/webapps/galaxy/config_schema.yml
+++ b/lib/galaxy/webapps/galaxy/config_schema.yml
@@ -2158,6 +2158,15 @@ mapping:
           When using LDAP for authentication, allow administrators to pre-populate users
           using an additional form on 'Create new user'
 
+      enable_upload_from_form_button:
+        type: bool
+        default: false
+        required: false
+        desc: |
+          If enabled, add another button to tool form data inputs that allow uploading
+          data from the tool form in fewer clicks (at the expense of making the form more
+          complicated).
+
       allow_user_dataset_purge:
         type: bool
         default: true
--- a/lib/galaxy/webapps/galaxy/config_schema.yml
+++ b/lib/galaxy/webapps/galaxy/config_schema.yml
@@ -2158,15 +2158,18 @@ mapping:
           When using LDAP for authentication, allow administrators to pre-populate users
           using an additional form on 'Create new user'
 
-      enable_upload_from_form_button:
-        type: bool
-        default: false
+      upload_from_form_button:
+        type: str
+        default: 'always-off'
         required: false
+        enum: ['always-on', 'always-off']
         desc: |
-          If enabled, add another button to tool form data inputs that allow uploading
+          If 'always-on', add another button to tool form data inputs that allow uploading
           data from the tool form in fewer clicks (at the expense of making the form more
-          complicated).
+          complicated). This applies to workflows as well.
 
+          Avoiding making this a boolean because we may add options such as 'in-single-form-view'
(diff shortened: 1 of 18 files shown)


shenweiyan:
Awesome！

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -2149,6 +2149,15 @@ mapping:
           When using LDAP for authentication, allow administrators to pre-populate users
           using an additional form on 'Create new user'
 
+      enable_upload_from_form_button:
```
bgruening: I think this option is useful to overwrite in the final single-app view. So even when this is false, it should be visible in the single-form view. This said. Should this setting be a string "normal,single-app,none" or something similar?

bgruening (commented):
This works well for me! @jmchilton can you rebase this please? One comment included.

lib/galaxy/webapps/galaxy/config_schema.yml
jmchilton: Is b42af61850ea0f5b4798c7e4f567e4c2a447083f good enough for now? We don't really have a single form view ready to go yet right? Once we finalize a lot of these other open PRs it'll be worth coming back and revisiting?

jmchilton:
Thanks so much for the review @bgruening !

lib/galaxy/webapps/galaxy/config_schema.yml
bgruening: Yes, this is great John!

bgruening:
The client needs to be reformatted I think.

bgruening:
:heart:  Thanks John!

jmchilton:
Thanks for the merge - this is awesome!

bgruening:
Oh yes it is! Thanks for all your work and for implementing all of this. I am pretty sure this will be used a lot in diverse communities.

shenweiyan:
@jmchilton Can this function be used normally in release-20.09?

bgruening:
@shenweiyan yes. It is actiavted on usegalaxy.eu if you wanna try it.

shenweiyan:
> @shenweiyan yes. It is actiavted on usegalaxy.eu if you wanna try it.

Thank you for this greate job. It can be activated with `upload_from_form_button: always-on`.

PR #9811, opened 2020-05-24
Don't use Backbone to define masthead elements.

jmchilton:


client/galaxy/scripts/layout/menu.js
```diff
@@ -34,315 +42,301 @@ export function logoutClick() {
...
+}
+
+function loadWebhookMenuItems() {
+    //
+    // Webhooks
+    //
+    Webhooks.load({
+        type: "masthead",
+        callback: function (webhooks) {
+            $(document).ready(() => {
```
OlegZharkov, line 320: not a big deal and I guess wouldn't worth time, if it doesn't just work out of the box... But with ``` document.addEventListener('DOMContentLoaded', ()=>{} );``` we might be able to get rid of ```import $ from "jquery";``` in this file

jmchilton: I will make this change, thanks!

jmchilton: Change included in #9822 https://github.com/galaxyproject/galaxy/pull/9822/commits/7aa10bca2c14c2f59c40dc2f04424bc087a18e73.

OlegZharkov: Great!

PR #9820, opened 2020-05-27
Add selenium test for user API keys

davebx:
Also adds get_api_key() method to navigates_galaxy.py to compare with
what is rendered.


davebx:
@galaxybot test this

lib/galaxy/selenium/navigation.yml
```diff
@@ -71,7 +71,10 @@ preferences:
     sign_out: "#edit-preferences-sign-out"
     change_password: "#edit-preferences-password"
     manage_information: '#edit-preferences-information'
+    manage_api_key: '#edit-preferences-api-key'
     current_email: "#user-preferences-current-email"
+    api_key: "[data-label='Current API key:'] > input"
```
jmchilton: It looks like you defined this but didn't use it in the code? Can you do that? Otherwise this looks awesome - thanks @davebx !

davebx: It wasn't usable in the code, but I left it there with the reasoning that it might become usable one day. Then again, at that point it might be a different selector entirely, so it does make sense to remove it.

bgruening (approved):

bgruening:
Thanks @davebx!

PR #9071, opened 2019-12-06
[WIP] Porting Masthead to Vue.js

inkuzmin:
This is a work in progress on porting the masthead to Vue.js.

- [x] Add the new masthead below the original one
- [x] Achieve markup parity using https://bootstrap-vue.js.org/docs/components/navbar/ (thanks @guerler for the hint)
- [x] Add tooltips
- [x] "Please login or register to use this feature."
- [x] Add actions (links/routes), skip the scratch-book for a moment
- [x] Figure out what is the "nav-note" and port it
- [x] Figure out what the "noscratchbook" property is responsible for
- [x] Add the scratch-book
- [x] Highlight an active item correctly
- [x] Use dot notation for properties accessors
- [x] Refactor NavItem into a separate component
- [x] Handle dividers at the dropdown menu
- [x] Refactor NavItem classes as computed properties
- [x] Remove the original NavBar
- [ ] Check if any functionality is still missing and if that's the case port it
- [ ] Remove covered code
- [ ] Prettify code, make tests pass

PR #9822, opened 2020-05-27
More reactive, tested Masthead.

jmchilton:
... also less broken since there are some bug fixes in here.

- Eliminate another layer of Backbone.View (the one previously in layout/masthead.js) and put the Masthead component directly in the page.
- Fix onbefeoreunload handling broken in #9071 I think.
- Remove more underscore and jQuery usage.
- Make config handling around brand config reactive and implement a unit test.
- Eliminate hackery around loading webhooks in different ways in different context, just load them when component is created. Place them in a data entry so the underlying menu structure can change reactively and these will remain.
- Add unit test for loading masthead webhooks.
- Make actual entries in the masthead more reactive - previously it would just be generated once in masthead.js 

Builds on #9811 (which converted layout/menu.js to use a POJO instead of Backbone and #9810 which contains other masthead fixes.

### DIFF
--- a/client/galaxy/scripts/components/Masthead/Masthead.test.js
+++ b/client/galaxy/scripts/components/Masthead/Masthead.test.js
@@ -1,4 +1,4 @@
-import Masthead from "./Masthead.vue";
+import { default as Masthead, __RewireAPI__ as rewire } from "./Masthead.vue";
 import { mount, createLocalVue } from "@vue/test-utils";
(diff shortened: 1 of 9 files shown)


dannon:
Working well for me, thanks.

jmchilton:
Thanks for the quick merge of a big change @dannon, much appreciated.

PR #9824, opened 2020-05-28
[20.01] Tolerate exception in container resolution

mvdbeek:
We're potentially contacting external services here, so all sorts of unexpected things could happen there. Noticed this because quay.io is down atm.

### DIFF
--- a/lib/galaxy/tool_util/deps/containers.py
+++ b/lib/galaxy/tool_util/deps/containers.py
@@ -236,7 +236,11 @@ class ContainerRegistry(object):
 
     def find_best_container_description(self, enabled_container_types, tool_info, **kwds):
         """Yield best container description of supplied types matching tool info."""
-        resolved_container_description = self.resolve(enabled_container_types, tool_info, **kwds)
+        try:
+            resolved_container_description = self.resolve(enabled_container_types, tool_info, **kwds)
+        except Exception:
+            log.exception("Could not get container description for tool '%s'", tool_info.tool_id)
+            return None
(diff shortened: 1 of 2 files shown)


lib/galaxy/tool_util/deps/containers.py
```diff
@@ -236,7 +236,11 @@ def __resolvers_dict(self):
 
     def find_best_container_description(self, enabled_container_types, tool_info, **kwds):
         """Yield best container description of supplied types matching tool info."""
-        resolved_container_description = self.resolve(enabled_container_types, tool_info, **kwds)
+        try:
+            resolved_container_description = self.resolve(enabled_container_types, tool_info, **kwds)
+        except Exception:
+            log.exception("Could not get container description")
```
nsoranzo: Should we also add `tool_info` in the log message?

mvdbeek: That's a good idea. The exception did include the quay.io URL, but I guess we may be catching other exception as well.

mvdbeek:
And since quay.io is now just not responding we wait forever for the response, so I added a 10 second timeout.

mvdbeek:
You can see the new log message in https://github.com/galaxyproject/galaxy/pull/9824/checks?check_run_id=717004690#step:11:3432 :)

nsoranzo:
> And since quay.io is now just not responding we wait forever for the response, so I added a 10 second timeout.

Cool, there is also a `requests.post()` call to quay.io at line 33 which may need a timeout.

nsoranzo (approved):

PR #9825, opened 2020-05-28
Cleanup big masthead refactoring.

jmchilton:
Opps - sorry for the mess.


client/galaxy/scripts/components/Masthead/MastheadItem.vue
```diff
@@ -103,7 +103,6 @@ export default {
             };
         },
         iconClasses() {
-            console.log("in iconClasses...");
```
dannon, line 106: I even saw this in my first read and meant to mention it, but forgot to get back to it after testing.  Thanks.

PR #9827, opened 2020-05-28
Cleanup and refactoring client code around upload/data dialog

jmchilton:
This is a number of atomic cleanup and refactoring commits that enable #9809 to be just a clean single commit adding the functionality to embed uploads into the tool form and/or data selection dialog box. See individual commits for details - but it is all small stuff - removing underscore, fixing initialization of the upload form, eliminating older style ``self``, etc..

### DIFF
--- a/client/galaxy/scripts/components/Upload/config.js
+++ b/client/galaxy/scripts/components/Upload/config.js
@@ -1,0 +1,14 @@
+import { getGalaxyInstance } from "app";
+import { getAppRoot } from "onload";
+
+export function initializeUploadDefaults(propsData = {}) {
+    const Galaxy = getGalaxyInstance();
+    const appRoot = getAppRoot();
+    return Object.assign(propsData, {
+        uploadPath: Galaxy.config.nginx_upload_path || `${appRoot}api/tools`,
+        chunkUploadSize: Galaxy.config.chunk_upload_size,
+        ftpUploadSite: Galaxy.config.ftp_upload_site,
+        defaultGenome: Galaxy.config.default_genome,
+        defaultExtension: Galaxy.config.default_extension,
+    });
+}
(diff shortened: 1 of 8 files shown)


client/galaxy/scripts/components/Upload/utils.js
```diff
@@ -0,0 +1,7 @@
+/*
+ * Utilities for working with upload data structures.
+ */
+
+export function findExtension(extensions, id) {
+    return extensions.find((extension) => extension.id == id);
```
dannon, line 6: Minor, but I think it's clearer to leave this straightforward 1-liner in the original context, rather than abstracted into a new module.

client/galaxy/scripts/components/Upload/config.js
```diff
@@ -0,0 +1,14 @@
+import { getGalaxyInstance } from "app";
+import { getAppRoot } from "onload";
+
+export function initializeUploadDefaults(propsData = {}) {
+    const Galaxy = getGalaxyInstance();
+    const appRoot = getAppRoot();
+    return Object.assign(propsData, {
+        uploadPath: Galaxy.config.nginx_upload_path || `${appRoot}api/tools`,
```
dannon, line 8: I haven't tried it yet, but can this be functional default props in the component, simplifying how this is all spread out?

dannon:
Test failure unrelated.

client/galaxy/scripts/components/Upload/utils.js
jmchilton: Fair. I would agree too if it were just used in once place, but I reuse it downstream in another place so I established an abstraction to reuse.

dannon: Not worth tweaking now (as I already merged it anyway, of course), but even with multiple uses the single line is clear, standard javascript and I think abstracting it makes it more difficult to see what's happening at a glance.

PR #9828, opened 2020-05-28
Show tool with the latest version

anuprulez:
Show tool with the latest version as recommendations. There are some tools with Ids such as Filter1, Cut1 with no-tool version information embedded in these. But there are tools such as Bowtie2, Trimmomatic which contain their versions embedded in their ids. While showing recommendations, it is useful to fetch the latest version of these tools (and users can change the versions if they like).

### DIFF
--- a/lib/galaxy/tools/recommendations.py
+++ b/lib/galaxy/tools/recommendations.py
@@ -2,6 +2,7 @@
 
 import json
 import logging
+import operator
 import os
 
 import h5py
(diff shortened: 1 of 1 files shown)


lib/galaxy/tools/recommendations.py
```diff
@@ -100,13 +100,32 @@ def __set_model(self, trans, remote_model_url):
...
         return True
 
+    def __get_latest_tool(self, a_tools):
+        """
+        Sort tool versions and take the latest one
+        """
+        all_tools = dict()
+        for t in a_tools:
+            t_versions = list()
+            for t_v in a_tools[t]:
```
bernt-matthias: Would be better to get the index without sorting. 

Something along the lines of:
```
max_index, max_value = max(enumerate(values), key=operator.itemgetter(1))
```

anuprulez: thanks, will test it and update

lib/galaxy/tools/recommendations.py
```diff
@@ -100,13 +101,29 @@ def __set_model(self, trans, remote_model_url):
             for k in tool_pos_sorted:
                 self.tool_weights_sorted[k] = tool_weights[str(k)]
             # collect ids and names of all the installed tools
+            all_t = dict()
             for tool_id, tool in trans.app.toolbox.tools():
                 t_id_renamed = tool_id
```
mvdbeek, line 106: You can use `if tool.is_latest_version` and then drop the other changes

mvdbeek (changes requested):
You can simplify this a bit

PR #9830, opened 2020-05-29
i18n for tool page about email notification

shenweiyan:
Update tool-form.js, i18n for tool page about email notification.


jmchilton:
Awesome - thanks!

PR #9831, opened 2020-05-29
Add option to text asserts to specify number of occurrences

simonbray:
For example, this would be handy to check the number of chemical structures in an SD file; this uses `$$$$` as a separator, which can be counted to determine the number of molecules in the file.

After writing this code, I realized the same thing is actually already possible using regex: e.g. `<has_text_matching expression="^([^\$]+?\$\$\$\$){33}?$"/>` for a file containing 33 molecules. However, I think this would still be a convenient alternative: `<has_text text="$$$$" n="33"/>` is way nicer both to read and to write.

### DIFF
--- a/lib/galaxy/tool_util/verify/asserts/text.py
+++ b/lib/galaxy/tool_util/verify/asserts/text.py
@@ -1,10 +1,14 @@
 import re
 
 
-def assert_has_text(output, text):
+def assert_has_text(output, text, n=None):
     """ Asserts specified output contains the substring specified by
     the argument text."""
-    assert output.find(text) >= 0, "Output file did not contain expected text '%s' (output '%s')" % (text, output)
+    if not n:
+        assert output.find(text) >= 0, "Output file did not contain expected text '%s' (output '%s')" % (text, output)
+    else:
+        matches = re.findall(re.escape(text), output)
+        assert len(matches) == int(n), "Expected %s matches for '%s' in output file (output '%s'); found %s" % (n, text, output, len(matches))
 
 
 def assert_not_has_text(output, text):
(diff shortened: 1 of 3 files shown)


lib/galaxy/tool_util/verify/asserts/text.py
```diff
@@ -1,10 +1,14 @@
 import re
 
 
-def assert_has_text(output, text):
+def assert_has_text(output, text, n=None):
     """ Asserts specified output contains the substring specified by
     the argument text."""
-    assert output.find(text) >= 0, "Output file did not contain expected text '%s' (output '%s')" % (text, output)
+    if not n:
```
jdavcs: In the unlikely case that the function is called with `n=0`, this clause will execute, so we assert that `text` exists, although the argument implies 0 occurrences (`n=0`), which is not the expected behavior, I think. Maybe we should add a comment that if n is not set or n=0, the assertion will be made for 1 or more occurrences? Or, if *there is* a use case for checking for 0 occurrences (I don't know if there is), we could handle it here (if n is explicitly set to 0).

lib/galaxy/tool_util/verify/asserts/text.py
```diff
@@ -13,11 +17,15 @@ def assert_not_has_text(output, text):
     assert output.find(text) < 0, "Output file contains unexpected text '%s'" % text
 
 
-def assert_has_line(output, line):
+def assert_has_line(output, line, n=None):
     """ Asserts the specified output contains the line specified the
     argument line."""
-    match = re.search("^%s$" % re.escape(line), output, flags=re.MULTILINE)
-    assert match is not None, "No line of output file was '%s' (output was '%s') " % (line, output)
+    if not n:
```
jdavcs: same as above

lib/galaxy/tool_util/verify/asserts/text.py
simonbray: Oh yes, thanks for the catch.

Maybe the simplest option is `if n != None`? Not that I think there's a usecase for `n=0` (that's what `not_has_text` is for), but to avoid unexpected behavior, as you say.

Another option would be to have a default of `n=1` and then `if n != 1`.

nsoranzo: ~I'd vote for `n=1` as default.~

jdavcs: I think `if n is not None` (`is not`, not `!=`) may be a better option (as per @simonbray 's suggestion). `n=1` implies a count of **1**. So we would be changing assumptions for existing code (code that calls this function without the `n` argument and assumes 1 *or more* occurrences of `text` - we'd be changing that to exactly 1. The most obvious default, I think, would be "1 or more", with "not set" being the next best thing.

nsoranzo: You're right, I didn't think that through.

jdavcs (approved):

ic4f:
Looks good to me. Thank you!

mvdbeek:
Very cool, thanks a lot @simonbray!

PR #9833, opened 2020-05-29
[20.05] Migrate condor runner to jwd for job-files

bgruening:
Condor is still writing to `cluster_files_directory` this commit migrates to locations of job files into the `job_working_dir` as other runners do.

It is also including a small hack to migrate jobs that have been starting in Galaxy version that still use `cluster_files_directory`.

Thanks to @natefoo for the tip.

I'm happy to retarget `dev` if this is not considered a bug.

### DIFF
--- a/lib/galaxy/jobs/runners/condor.py
+++ b/lib/galaxy/jobs/runners/condor.py
@@ -57,6 +57,19 @@ class CondorJobRunner(AsynchronousJobRunner):
         self._init_monitor_thread()
         self._init_worker_threads()
 
+    def __old_state_paths(self, cjs):
+        """For recovery of jobs started prior to standardizing the naming of
+        files in the AsychronousJobState object
+        """
+        if cjs.job_wrapper is not None:
+            job_file = "%s/galaxy_%s.sh" % (self.app.config.cluster_files_directory, cjs.job_wrapper.job_id)
+            if not os.path.exists(cjs.job_file) and os.path.exists(job_file):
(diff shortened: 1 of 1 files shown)


lib/galaxy/jobs/runners/condor.py
```diff
@@ -57,6 +57,19 @@ def __init__(self, app, nworkers):
         self._init_monitor_thread()
         self._init_worker_threads()
 
+    def __old_state_paths(self, cjs):
+        """For recovery of jobs started prior to standardizing the naming of
+        files in the AsychronousJobState object
+        """
+        if cjs.job_wrapper is not None:
+            job_file = "%s/galaxy_%s.sh" % (self.app.config.cluster_files_directory, cjs.job_wrapper.job_id)
```
mvdbeek, line 66: Maybe do `cluster_files_dir_and_id = (self.app.config.cluster_files_directory, cjs.job_wrapper.get_id_tag())` and then use that in the string formatting below?

mvdbeek (approved):
I think it makes sense to get this in, I would call it a bug :).

lib/galaxy/jobs/runners/condor.py
```diff
@@ -57,6 +57,22 @@ def __init__(self, app, nworkers):
         self._init_monitor_thread()
         self._init_worker_threads()
 
+    def __old_state_paths(self, cjs):
+        """For recovery of jobs started prior to standardizing the naming of
+        files in the AsychronousJobState object
+
```
mvdbeek:
```suggestion

```

natefoo (approved):
Sorry we missed the Condor runner back when this was done for the others!

bgruening:
Thanks a lot!

PR #9846, opened 2020-06-03
[Selenium] History Panel Tags&Annotation

OlegZharkov:
Selenium test for changing Tags and Annotation of history panel.


jmchilton:
Amazing - thanks so much!

OlegZharkov:
Thank you @jmchilton

PR #9850, opened 2020-06-04
[20.05] Move some tools that no longer need galaxy on the PATH

bernt-matthias:
https://github.com/galaxyproject/tools-devteam/pull/560
https://github.com/galaxyproject/tools-devteam/pull/559

Maybe more tools can be removed. Any idea how to identify such tools automatically?

I'm targeting 20.05 because this would be handy for the PR above.


jdavcs (approved):
Very neat refactoring, @nsoranzo! I would've never thought that module had so many instances of not closed file handles! (also, +1 on `_` for throwaway var names)

nsoranzo (approved):

PR #9851, opened 2020-06-04
[20.05] Follow-up on #9587

nsoranzo:
Follow-up on #9587.

- Small refactor to ``validate_password()``.
- Add back deleted line.
- Don't use mutables as argument defaults
- Clean up unused variables.

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -388,7 +388,7 @@ class User(Dictifiable, RepresentById):
         """
         Set user password to the digest of `cleartext`.
         """
-        message = validate_password_str(cleartext, cleartext)
+        message = validate_password_str(cleartext)
         if message:
             raise Exception("Invalid password: %s" % message)
         if User.use_pbkdf2:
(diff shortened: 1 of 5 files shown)


lib/galaxy/model/__init__.py
```diff
@@ -3226,7 +3226,7 @@ def to_dict(self, view='collection', expose_dataset_path=False):
         if hda.extended_metadata is not None:
             rval['extended_metadata'] = hda.extended_metadata.data
 
-        for name, spec in hda.metadata.spec.items():
+        for name in hda.metadata.spec.keys():
```
jdavcs, line 3229: Minor suggestion only:`for name in hda.metadata.spec:` (`keys()` not needed here, unless you want it to be more explicit?). Same for the other 2 cases below.

nsoranzo: Following the Zen of Python "Explicit is better than implicit", I think it makes it clearer we are dealing with a dict.

jdavcs (approved):

nuwang:
Thanks for fixing that @nsoranzo. Lgtm!

PR #9853, opened 2020-06-05
Added support for dynamically refreshing drilldowns

nuwang:
While the drilldown component allows hierarchical structures to be displayed in a tool form, currently, the entire tree must be preloaded. This is often not feasible when the tree structure is complex.

This PR adds support for fetching additional child nodes as the tree structure is explored, by making a server request on each node expansion. The change is largely backward compatible, except that when `refresh_on_change` is True, the client component includes data on which node was expanded along with the currently selected value(s).

This allows server side code to retrieve the expanded node's value and fetch its child nodes only.

For example, with the following definition:
```
<param name="target_folder" refresh_on_change="true" type="drill_down" display="radio" hierarchy="exact"
           multiple="true" label="Choose Target Directory"
           dynamic_options="get_owncloud_folders(trans=__trans__, target_folder=target_folder, list_dirs_only=True)"
           help="Select target folder to save files in.">
    </param>
```

The following data is sent over the wire when a node is expanded:

image

With the following result:

image

There is a working version of a tool that uses this here:
https://galaxy-aust-staging.genome.edu.au/

Select Get Data -> Import from OwnCloud

Tool source: https://github.com/usegalaxy-au/Galaxy-Owncloud-Integration/blob/master/Galaxy_OwncloudImportExport/send_to_owncloud.xml

### DIFF
--- a/lib/galaxy/tools/parameters/basic.py
+++ b/lib/galaxy/tools/parameters/basic.py
@@ -1412,6 +1412,8 @@ class DrillDownSelectToolParameter(SelectToolParameter):
         return self.options
 
     def get_legal_values(self, trans, other_values):
+        if self.is_dynamic:
+            return None
         def recurse_options(legal_values, options):
             for option in options:
                 legal_values.append(option['value'])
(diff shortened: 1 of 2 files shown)


lib/galaxy/tools/parameters/basic.py
```diff
@@ -1433,18 +1435,23 @@ def from_json(self, value, trans, other_values={}):
             if self.optional:
                 return None
             raise ValueError("parameter '%s': an invalid option (%r) was selected" % (self.name, value))
-        elif not legal_values:
+        elif not legal_values and not self.is_dynamic:
             raise ValueError("parameter '%s': requires a value, but no legal values defined" % (self.name))
+        if isinstance(value, dict):
+            clicked_node = value.get('last_clicked_node')
```
jmchilton, line 1441: I'm really nervous about this complexity, in this file in particular and without test cases. @nuwang and I discussed adding WebDav to the upload component and rule builder via a plugin framework as an alternative to this. I will try to prototype this... I hope...

guerler (commented):
I agree with @jmchilton and adding it it as plugin for the uploader would be great.

jmchilton:
I think we want to just close this @nuwang - hope that is okay?

nuwang:
Yes np. @jmchilton

PR #9855, opened 2020-06-05
Add support for relative paths in dependency resolver mappings

rupertnash:
I have been using the dependency resolver feature of tool_util from cwltool, however there is no support for dependency mapping files to specified by a relative path (I really want this for maintaining a set of such files in a way that can be shared between multiple users of an HPC system)

In this patch I implement the feature I want. Please let me know if this needs any other work and I do this on Wednesday.

Thanks for the useful library!

### DIFF
--- a/lib/galaxy/tool_util/deps/resolvers/__init__.py
+++ b/lib/galaxy/tool_util/deps/resolvers/__init__.py
@@ -5,7 +5,7 @@ from abc import (
     abstractmethod,
     abstractproperty,
 )
-
+import os.path
 import six
 import yaml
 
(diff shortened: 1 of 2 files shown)


lib/galaxy/tool_util/deps/resolvers/__init__.py
```diff
@@ -5,7 +5,7 @@
     abstractmethod,
     abstractproperty,
 )
-
+import os.path
```
nsoranzo: We sort the imports in 3 groups: standard library, 3rd party and internal. Can you move this right after `import errno` please?

rupertnash: Done! Apologies for the broken unit tests - tox was failing to set up the env correctly so I gave up!

mr-c:
@nsoranzo friendly ping :-)

jmchilton:
Should be good - thanks for the fix and sorry for the delay.

PR #9858, opened 2020-06-08
[Selenium] History structure

OlegZharkov:
Since, we are going to rework history structure in the following couple of weeks, this PR brings just basic test. 
Flow:
- upload test data 
- click on menu - history structure
- Assure dataset details are not visible 
- Expand dataset structure
- Expand dataset details
- Shrink dataset structure
- Assure details and dataset are not visible

Update:
forgot to mention. It also fixes JS logging. If we setup debug level on python, it produces cyclic object and ```JSON.stringify()``` breaks badly, then structure is not rendered properly. 
[This check](https://github.com/galaxyproject/galaxy/pull/9858/files#diff-d34fe858167e556f4301d76b6372ec8cR295) fixes the problem

### DIFF
--- a/client/galaxy/scripts/utils/metrics-logger.js
+++ b/client/galaxy/scripts/utils/metrics-logger.js
@@ -296,7 +296,7 @@ function usefulToString(arg) {
         } catch (e) {
             // If arg has cyclic reference, return String
             // otherwise rendering stops and we have incomplete page
-            console.error(e)
+            console.error(e);
             return String(arg);
         }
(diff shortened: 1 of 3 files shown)


OlegZharkov:
@galaxybot test this

nsoranzo:
@galaxybot test this

jmchilton:
So awesome - thanks so much!

OlegZharkov:
thank you @jmchilton

PR #9860, opened 2020-06-08
Allow composite and extra file uploads via data fetch API.

jmchilton:
This PR is doing a few interconnected things simultaneously - at a high level it is adding "composite file"/"extra files" functionality to the newer data fetch (or upload 2.0) API and it is enabling nested collection inputs and tags in workflow tests (continuing thread from https://github.com/galaxyproject/galaxy/pull/9803). Doing the former turned out to be needed to do the latter well.

# Extra Files API

This PR adds the ability to upload composite datatype files as well as free form directories of extra files to uploads done through the data fetch or upload 2.0 API. Both of these options have been in upload1 for a while. The API is much more natural to express composite and extra files in - and works much more naturally with collections and tags so it made a lot of sense to accomplish the second big task by implementing this first.

# Workflow Tests - Nested Collections and Tags 

There is this function in ``galaxy.tool_util.cwl.util`` called ``galactic_job_json`` that parses a CWL-style "job" (the inputs for a tool or a workflow execution as a JSON object) and parses it into a representation for Galaxy - uploading inputs to Galaxy as needed. The interaction with the API was originally abstracted out as two functions so that the Galaxy test code (in the CWL branch) and the Planemo driver code could interact with the API in different ways. Planemo used bioblend (loosely) and the Galaxy code used the test framework's "populators" abstractions.

This PR creates a new layer above `galactic_job_json` that has some extension points that the Planemo/bioblend layer can take advantage of but by default will just use the galaxy_interactor stuff already being used by tool tests using galaxy-tool-util. So this new layer doesn't depend on test "populators" or bioblend and should allow important code to be shared between the Galaxy test framework, the CWL branch of Galaxy, and Planemo. Having to update the code in both places was a significant problem with the CWL branch.

Here is a commit demonstrating how it simplifies the CWL branch (jmchilton/galaxy@d9e774f). I don't have a matching commit for Planemo yet but the code is located here https://github.com/galaxyproject/planemo/blob/master/planemo/galaxy/activity.py#L181 that will be simplified after this commit is utilized.

Another positive upshot of this commit is there is now some in core test coverage of galactic_job_json - an important library function.

There are some other smaller commits here to setup some stuff in galaxy-tool-util to be shared between Galaxy testing populators and Planemo and refactoring of test case stuff to get testing done well.

### DIFF
--- a/lib/galaxy/model/store/discover.py
+++ b/lib/galaxy/model/store/discover.py
@@ -47,6 +47,7 @@ class ModelPersistenceContext(object):
         dbkey,
         name,
         filename=None,
+        extra_files=None,
         metadata_source_name=None,
         info=None,
         library_folder=None,
@@ -147,8 +148,12 @@ class ModelPersistenceContext(object):
                 self.object_store.update_from_file(primary_data.dataset, file_name=filename, create=True)
             else:
                 primary_data.link_to(filename)
-            # We are sure there are no extra files, so optimize things that follow by settting total size also.
-            primary_data.set_size(no_extra_files=True)
+            if extra_files:
+                persist_extra_files(self.object_store, extra_files, primary_data)
+                primary_data.set_size()
+            else:
+                # We are sure there are no extra files, so optimize things that follow by settting total size also.
+                primary_data.set_size(no_extra_files=True)
 
         # If match specified a name use otherwise generate one from
         # designation.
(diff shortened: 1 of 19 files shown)


shenweiyan:
Awesome job.

lib/galaxy/tools/data_fetch.py
```diff
@@ -152,23 +232,52 @@ def _resolve_src(item):
...
+                        name = item.get("name")
+                        if not prefix:
+                            item_prefix = name
+                        else:
+                            item_prefix = prefix + "/" + name
+                        walk_extra_files(item.get("elements"), prefix=item_prefix)
+                    else:
+                        name, src_path = _has_src_to_path(upload_config, item)
+                        if prefix:
+                            rel_path = prefix + "/" + name
```
mvdbeek: We should probably use os.path.sep here (or os.path.join)

lib/galaxy/model/store/discover.py
```diff
@@ -147,8 +148,12 @@ def create_dataset(
                 self.object_store.update_from_file(primary_data.dataset, file_name=filename, create=True)
             else:
                 primary_data.link_to(filename)
-            # We are sure there are no extra files, so optimize things that follow by settting total size also.
-            primary_data.set_size(no_extra_files=True)
+            if extra_files:
```
mvdbeek, line 151: We should add that handling to populate_collection_elements as well

lib/galaxy/tools/data_fetch.py
```diff
@@ -152,23 +232,52 @@ def _resolve_src(item):
...
+        staged_extra_files = None
+        if extra_files:
+            # TODO: optimize to just copy the whole directory to extra files instead.
+            assert not upload_config.link_data_only, "linking composite dataset files not yet implemented"
+            extra_files_path = path + "_extra"
+            staged_extra_files = extra_files_path
+            os.mkdir(extra_files_path)
+
+            def walk_extra_files(items, prefix=""):
+                print(items)
```
mvdbeek:
```suggestion
```

lib/galaxy/tools/data_fetch.py
```diff
@@ -152,23 +232,52 @@ def _resolve_src(item):
...
+                for item in items:
+                    print(item)
```
mvdbeek:
```suggestion
```

mvdbeek (commented):
I've opened a PR with the things I found at https://github.com/galaxyproject/galaxy/pull/10022 for the tests.

mvdbeek:
@jmchilton can you pull in the last 2 commits in https://github.com/galaxyproject/galaxy/pull/10022 ? Happy to merge then.

jmchilton:
Amazing - thanks for the detailed review and the fixes. You are awesome.

mvdbeek (approved):

mvdbeek:
No problem, this is great!

PR #9862, opened 2020-06-10
Add allowed_cilogon list functionality (idphint short list)

juleengraham:


lib/galaxy/webapps/galaxy/controllers/authnz.py
```diff
@@ -136,3 +136,10 @@ def get_cilogon_idps(self, trans, **kwargs):
             return json.loads(url_get('https://cilogon.org/idplist/', params=dict(kwargs)))
         except Exception as e:
             raise Exception("Invalid server response. %s." % str(e))
+
+    @web.expose
+    @web.json
+    def get_allowed_cilogon_idps(self, trans, **kwargs):
```
dannon: Instead of the additional endpoint (and request to fetch it), let's just apply the filtering in `get_cilogon_idps` above.

lib/galaxy/authnz/managers.py
```diff
@@ -157,12 +157,19 @@ def _parse_custos_config(self, config_xml):
             rtv['well_known_oidc_config_uri'] = config_xml.find('well_known_oidc_config_uri').text
         if config_xml.find('idphint') is not None:
             rtv['idphint'] = config_xml.find('idphint').text
+        # can prob get rid of idphint ^
+        if config_xml.find('allowed_idps') is not None:
+            self.allowed_idps = config_xml.find('allowed_idps').text
```
dannon: Can you add an example to the sample for this?

juleengraham: @dannon do you mean adding an example in the docs or adding a comment here of what it would look like?

juleengraham:
@dannon, invalid entityIds shouldn't work since it filters the cilogon list that's obtained from cilogon, so only valid entityIds would match. this filtering has been moved to the server-side but same logic should still apply

dannon:
Rebased to resolve conflicts cleanly; waiting on tests and then I'll merge this.

dannon:
Test failures unrelated to this.

PR #9864, opened 2020-06-10
Change usage of whitelist/blacklist terms to allowlist/blocklist

natefoo:
Although not explicitly derived from race, the terms "whitelist" and "blacklist" connote, even if unintentional, that "white = good" and "black = bad." The terms "allow" and "deny" are neutral.

Changes should be backwards compatible. Marked WIP as a test for the `renamed_options` stuff would probably be a good idea.


ic4f:
+1 on the renaming.

Two of the failed unit tests fail because those options need to be renamed in the unit test as well: 
https://github.com/galaxyproject/galaxy/blob/dev/test/unit/config/test_config_values.py#L44
https://github.com/galaxyproject/galaxy/blob/dev/test/unit/config/test_config_values.py#L90
Otherwise, one property is not listified before the test, and the other doesn't have its expected path set (see [this comment](https://github.com/galaxyproject/galaxy/blob/dev/test/unit/config/test_config_values.py#L108) for a description of expected paths for that unit test)

lib/galaxy/config/__init__.py
```diff
@@ -507,15 +523,19 @@ def _process_config(self, kwargs):
...
+        if not os.path.isfile(_sanitize_allowlist_path):  # then check old default location
+            for deprecated in (
+                    self._in_managed_config_dir('sanitize_whitelist.txt'),
+                    self._in_root_dir('config/sanitize_whitelist.txt')
+                ):
+                if os.path.isfile(deprecated):
+                    log.warning("The path '%s' for the 'sanitize_allowlist_file' config option is "
+                        "deprecated and will be no longer checked in a future release. Please consult "
+                        "the latest version of the sample configuration file." % deprecated)
+                    _sanitize_allowlist_path = deprecated
```
jdavcs, line 539: If both paths point to a file (i.e., the one in root may have been left over by mistake), then the second path will override the first (i.e., we have the same file in `[managed_config_dir]/` and in `[root]/config` - so the one in `root`  is used, although the one in `managed_config_dir` is, likely, the more recent version. So maybe swap the paths in lines 529-30?

lib/galaxy/config/__init__.py
```diff
@@ -313,6 +314,13 @@ def root_join(val):
                 setattr(self, var + '_set', False)
             setattr(self, var, [root_join(x) for x in paths])
 
+        # update kwargs to set unset renamed options
```
jdavcs: The problem is that we use the `kwargs` arguments passed to `GalaxyAppConfiguration` very early: [here](https://github.com/galaxyproject/galaxy/blob/5b9b1d26373040c54b14b7b49855a319f08a9f9c/lib/galaxy/config/__init__.py#L121) and even [here](https://github.com/galaxyproject/galaxy/blob/5b9b1d26373040c54b14b7b49855a319f08a9f9c/lib/galaxy/config/__init__.py#L116). There is (or there may be) code that depends on the assumption that these arguments will not play any further role in loading (or [reloading](https://github.com/galaxyproject/galaxy/blob/5b9b1d26373040c54b14b7b49855a319f08a9f9c/lib/galaxy/config/__init__.py#L874)) the config, once they have been processed in `_update_raw_config_from_kwargs()`.

I agree that seamless backward compatibility is a requirement here. So, if we have to modify these arguments, let's do it before any subsequent processing, and, maybe, let's be very explicit about this:
```python
def __init__(self, **kwargs):
    self._append_renamed_options(kwargs)  # or any other method name
    self.config_dict = kwargs   # and the rest stays as is
```
What do you think?

lib/galaxy/config/__init__.py
natefoo: Doesn't the `break` on 537 take care of that?

jdavcs: @natefoo  yeah, of course it does! I simply chose not to notice it...

lib/galaxy/config/__init__.py
natefoo: Makes sense, done in the latest commit, thanks!

jdavcs (approved):

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -1639,15 +1640,14 @@ mapping:
...
         type: str
-        default: sanitize_whitelist.txt
+        default: sanitize_allowlist.txt
         required: false
         desc: |
-          Whitelist sanitization file.
           Datasets created by tools listed in this file are trusted and will not have
           their HTML sanitized on display.  This can be manually edited or manipulated
-          through the Admin control panel -- see "Manage Display Whitelist"
+          through the Admin control panel -- see "Manage Display Allowlist"
```
```suggestion
          through the Admin control panel -- see "Manage Allowlist"
```

lib/galaxy/tool_util/deps/mulled/mulled_list.py
```diff
@@ -67,25 +67,25 @@ def get_conda_envs(filepath):
...
-    Return list of quay containers that do not exist as singularity containers. Files stored in a blacklist will be ignored
+    Return list of quay containers that do not exist as singularity containers. Files stored in a blocklist will be ignored
     """
-    blacklist = []
-    if blacklist_file:
-        blacklist = open(blacklist_file).read().split('\n')
-    return [n for n in quay_list if n not in singularity_list and n not in blacklist]
+    blocklist = []
+    if blocklist_file:
+        blocklist = open(blocklist_file).read().split('\n')
```
nsoranzo:
```suggestion
        with open(blocklist_file) as fh:
            blocklist = fh.read().split('\n')
```

lib/galaxy/tool_util/deps/mulled/mulled_list.py
```diff
@@ -67,25 +67,25 @@ def get_conda_envs(filepath):
...
+def get_missing_envs(quay_list, conda_list, blocklist_file=None):
     r"""
     Compares list of conda envs and docker containers and returns missing conda envs
     """
-    blacklist = []
-    if blacklist_file:
-        blacklist = open(blacklist_file).read().split('\n')
+    blocklist = []
+    if blocklist_file:
+        blocklist = open(blocklist_file).read().split('\n')
```
nsoranzo:
```suggestion
        with open(blocklist_file) as fh:
            blocklist = fh.read().split('\n')
```

lib/galaxy/config/__init__.py
```diff
@@ -122,6 +124,16 @@ def __init__(self, **kwargs):
...
         self._resolve_paths()  # Overwrite attribute values with resolved paths
 
+    def _process_renamed_options(self, kwargs):
+        """Update kwargs to set any unset renamed options to values of old-named options, if set.
+
+        Does not remove the old options from kwargs so that deprecated option usage can be logged.
+        """
+        if self.renamed_options is not None:
+            for old, new in self.renamed_options.items():
+                if new not in kwargs and old in kwargs:
```
```suggestion
                if old in kwargs and new not in kwargs:
```

lib/tool_shed/webapp/config.py
```diff
@@ -101,10 +101,10 @@ def _process_config(self, kwargs):
         # Galaxy flavor Docker Image
         self.user_activation_on = None
         self.registration_warning_message = kwargs.get('registration_warning_message')
-        self.blacklist_location = kwargs.get('blacklist_file')
-        self.blacklist_content = None
-        self.whitelist_location = kwargs.get('whitelist_file')
-        self.whitelist_content = None
+        self.email_domain_blocklist_location = kwargs.get('email_domain_blocklist_file')
```
nsoranzo:
```suggestion
```
Or maybe this line can just be dropped, @ic4f ?

lib/tool_shed/webapp/config.py
```diff
@@ -101,10 +101,10 @@ def _process_config(self, kwargs):
...
+        self.email_domain_blocklist_content = None
+        self.email_domain_allowlist_location = kwargs.get('email_domain_allowlist_file')
```
nsoranzo:
```suggestion
```

lib/tool_shed/webapp/config.py
jdavcs: If we rename the property to match the name in the schema, then no, we don't need to fetch it again (we shouldn't, really).

I'm not sure why those 4 lines (104-107) exist at all. I assume the TS app does NOT access `galaxy/security/validate_user_input`? And if so, those files are never used in TS.

If these lines *are*  needed, we can move [this](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L491) into [this](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L317).

Separate PR maybe?

nsoranzo: Thanks, I've updated my suggested changes to now remove the lines defining the 2 `_location` attributes, which seems to not be used anywhere. Not sure about the `_content` attributes.

jdavcs: I'm +1 on your suggested edit.
I think both `*_content` attributes are not needed: they were introduced in 3340d6c when TS was part of `lib/galaxy`. However, I think it may be easier/more fitting to make that change separately, with some testing (along with other TS config cleanup - I suspect there's more like this).

client/galaxy/scripts/nls/fr/locale.js
```diff
@@ -220,7 +220,7 @@ define({
     "Reload a tool's configuration": "Recharger une configuration d'outils",
     "Review tool migration stages": "Vérifier les étapes de migration d'outils",
     "View Tool Error Logs": "Voir les logs d'erreur d'outils",
-    "Manage Display Whitelist": "Gérer la liste blanche d'outils graphiques",
+    "Manage Allowlist": "Gérer la liste blanche d'outils graphiques",
```
mvdbeek:
```suggestion
    "Manage Allowlist": "Gérer la liste autorisée d'outils graphiques",
```

lib/tool_shed/webapp/config.py
natefoo: The TS tests were failing before these changes due to the reference in `validate_user_input`. Are you sure this is unnecessary?

client/galaxy/scripts/nls/fr/locale.js
natefoo: merci!

lib/tool_shed/webapp/config.py
jdavcs: Not sure at all, now that you mention this! Let's leave as is; I'll look at this now and, once this PR is merged, I'll have this cleaned up.

nsoranzo: @natefoo `validate_email()` in the `validate_user_input` module uses the `_content` attributes, so the `_location` attributes can be removed already.

nsoranzo (approved):

nsoranzo:
Thanks @natefoo !

PR #9866, opened 2020-06-11
[20.05] Invalidate tool cache when tool modified outside of Galaxy

### LINKED ISSUE #9863: Caching issue when rebuilding
When updating a tool xml file and rerunning the app, it seems we are not invalidating the old cache correctly, as updates are not applied to the running instance.

mvdbeek:
We store modtime and if the modtime changes we regenerate the cache entry. Fixes https://github.com/galaxyproject/galaxy/issues/9863

### DIFF
--- a/lib/galaxy/tool_util/parser/xml.py
+++ b/lib/galaxy/tool_util/parser/xml.py
@@ -1,5 +1,4 @@
 import logging
-import os
 import re
 import sys
 import traceback
@@ -500,14 +499,13 @@ class XmlToolSource(ToolSource):
         help_elem = self.root.find('help')
         return help_elem.text if help_elem is not None else None
 
+    @property
     def macro_paths(self):
         return self._macro_paths
 
-    def paths_and_modtimes(self):
-        paths_and_modtimes = {p: os.path.getmtime(p) for p in self.macro_paths()}
-        if self._source_path:
-            paths_and_modtimes[self._source_path] = os.path.getmtime(self._source_path)
-        return paths_and_modtimes
+    @property
+    def source_path(self):
+        return self._source_path
 
     def parse_tests_to_dict(self):
         tests_elem = self.root.find("tests")
--- a/lib/galaxy/tool_util/parser/xml.py
+++ b/lib/galaxy/tool_util/parser/xml.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import re
 import sys
 import traceback
@@ -51,6 +52,7 @@ class XmlToolSource(ToolSource):
         self.root = xml_tree.getroot()
         self._source_path = source_path
         self._macro_paths = macro_paths or []
+        self._paths_and_modtimes = {}
         self.legacy_defaults = self.parse_profile() == "16.01"
 
     def to_string(self):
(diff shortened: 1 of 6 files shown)


lib/galaxy/tool_util/parser/xml.py
```diff
@@ -502,6 +504,13 @@ def parse_help(self):
     def macro_paths(self):
         return self._macro_paths
 
+    def paths_and_modtimes(self):
+        if not self._paths_and_modtimes:
+            self._paths_and_modtimes = {p: os.path.getmtime(p) for p in self._macro_paths}
```
```suggestion
            self._paths_and_modtimes = {p: os.path.getmtime(p) for p in self.macro_paths()}
```

lib/galaxy/tool_util/parser/xml.py
```diff
@@ -502,6 +504,13 @@ def parse_help(self):
...
+            if self._source_path:
+                self._paths_and_modtimes[self._source_path] = os.path.getmtime(self._source_path)
```
```suggestion
            source_path = getattr(self, "_source_path", None)
            if source_path:
                self._paths_and_modtimes[source_path] = os.path.getmtime(source_path)
```

nsoranzo (commented):
I think with my proposed changes you can move the whole implementation to `lib/galaxy/tool_util/parser/interface.py`

lib/galaxy/tool_util/parser/xml.py
```diff
@@ -502,6 +504,13 @@ def parse_help(self):
     def macro_paths(self):
         return self._macro_paths
 
+    def paths_and_modtimes(self):
+        if not self._paths_and_modtimes:
```
nsoranzo: Is it safe to store this in the `XmlToolSource` object or is it better to regenerate the return value at each call?

mvdbeek:
> I think with my proposed changes you can move the whole implementation to `lib/galaxy/tool_util/parser/interface.py`

`if getattr(instance, "attribute", default)` always feels quite hacky to me ... and I think it makes sense to separate how you get `paths_and_modtimes` for the different ToolSource classes, since I don't know what macro mechanism we're going to use in YamlToolSource and the other variants.

lib/galaxy/tool_util/parser/xml.py
mvdbeek: We're not re-using it right now, so yes, we can do that.

nsoranzo:
> `if hasattr(instance, "attribute")` always feels quite hacky to me ... and I think it makes sense to separate how you get `paths_and_modtimes` for the different ToolSource classes, since I don't know what macro mechanism we're going to use in YamlToolSource and the other variants.

We can simple define a `source_path` property like we have `macro_paths` (which also could be a property BTW). All the 3 present subclasses of `ToolSource` have `self._source_path` defined in their `__init__()` method, so we don't need 3 different implementations.

lib/galaxy/tool_util/parser/xml.py
```diff
@@ -499,9 +499,14 @@ def parse_help(self):
         help_elem = self.root.find('help')
         return help_elem.text if help_elem is not None else None
 
+    @property
     def macro_paths(self):
         return self._macro_paths
 
+    @property
+    def source_path(self):
+        return self._source_path
```
nsoranzo, line 508: Can this be the default implementation in `ToolSource()` ? Or copy this to `CwlToolSource` and `YamlToolSource`.

nsoranzo (approved):

assuntad23 (approved):
This fixed it! Thanks!

PR #9869, opened 2020-06-11
Improved UploadModal state handling during initialization.

jmchilton:
Track whether datatypes and genomes have been set and use the LoadingSpan component to display information about this wait if the component is not yet ready.

This was part of #9827 which was merged into dev, but I think the state initialization things here make the dialog more robust and might fix problems @jennaj reported internally today.

### DIFF
--- a/client/galaxy/scripts/components/Upload/UploadModal.vue
+++ b/client/galaxy/scripts/components/Upload/UploadModal.vue
@@ -12,7 +12,7 @@
         <template v-slot:modal-header>
             <h4 class="title" tabindex="0">{{ title }}</h4>
         </template>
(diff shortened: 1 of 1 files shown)


mvdbeek:
Probably also fixes #9877. dev still runs a commit prior to https://github.com/galaxyproject/galaxy/pull/9827 and shows the same issue as #9877. Updating test.galaxyproject.org now to see if that fixes it.

mvdbeek:
https://github.com/galaxyproject/galaxy/pull/9827 seems to have fixed it on test.galaxyproject.org.

mvdbeek:
argh, this targeted the wrong branch. Opened #9878 which targets 20.05

PR #9875, opened 2020-06-12
Add database data table

selten:
Current data tables do not support locking of the underlying files when additions or removals are made. Due to the multi-process nature of Galaxy deployments, this could cause corruptions to occur within the data table. 
This PR aims to resolve that issue by storing the data commonly found in data tables inside the database of the Galaxy server.


selten:
@blankenberg Do you think you could review this at some point?

jmchilton:
We'd like to have a broader conversation around this - it would be a big change - but one thing that came up during the committers meeting is that it would be better if this used a separate sqlite database. There is an example of something like that in https://github.com/galaxyproject/galaxy/tree/dev/lib/galaxy/model/tool_shed_install.

selten:
> We'd like to have a broader conversation around this - it would be a big change - but one thing that came up during the committers meeting is that it would be better if this used a separate sqlite database. There is an example of something like that in https://github.com/galaxyproject/galaxy/tree/dev/lib/galaxy/model/tool_shed_install.

I think putting this onto a SQLite database might actually negate some of the benefits that are gained by using Galaxy's included database (especially when normally using PostgreSQL).
I think going for SQLite would mean that each data table would require its own SQLite file to maintain concurrency of multiple data managers running often and there may still be problems due to the concurrency within Galaxy with opening the SQLite database and writing to it simultaneously with multiple processes. Especially when Galaxy is installed on e.g. NFS (https://www.sqlite.org/howtocorrupt.html#_filesystems_with_broken_or_missing_lock_implementations). Therefore I believe moving to SQLite could result in the same issues as I'm currently experiencing with the files.

By having it inside PostgreSQL (which would be the default for me), I avoid the locking problem and thereby the corruption. Would using a separate schema in PostgreSQL be another option?

nsoranzo:
> By having it inside PostgreSQL (which would be the default for me), I avoid the locking problem and thereby the corruption. Would using a separate schema in PostgreSQL be another option?

An optional separate database connection (so any DBMS) is actually how the mentioned tool_shed_install database is configured, see `install_database_connection` in `config/galaxy.yml.sample` .

lib/galaxy/config/__init__.py
```diff
@@ -549,6 +549,12 @@ def _process_config(self, kwargs):
         self.database_engine_options = get_database_engine_options(kwargs)
         self.database_create_tables = string_as_bool(kwargs.get('database_create_tables', 'True'))
         self.database_encoding = kwargs.get('database_encoding')  # Create new databases with this encoding
+
+        # Database data tables connection
+        if not self.data_table_connection:  # Provide default if not supplied by user
+            db_path = self._in_data_dir('data_tables.sqlite')
+            self.data_table_connection = 'sqlite:///%s?isolation_level=IMMEDIATE' % db_path
```
innovate-invent: Can you make this a template?
I can see a use case where data tables need to be stored in a different db per table (sqlite databases)
template values should include table id and tool repo id

lib/galaxy/model/data_tables/mapping.py
```diff
@@ -0,0 +1,117 @@
...
+
+data_tables_model.DataTableRowAssociation.table = Table(
+    'data_table_row_association',
+    metadata,
+    Column("id", Integer, primary_key=True),
+    Column("data_table_id", Integer, ForeignKey("data_table.id"), index=True),
+    Column("data_table_row_id", Integer, ForeignKey("data_table_row.id"), index=True),
+)
+
+data_tables_model.DataTableField.table = Table(
```
innovate-invent, line 73: I don't think I agree with the idea of a generic table. This means there are #col * # rows rows in this table and doesn't allow leveraging row level optimizations when querying. How hard would it be to dynamically generate the table schema based on the description in the tool_data_table_conf.xml?

lib/galaxy/config/__init__.py
selten: SQLite is the default, same for the regular Galaxy database. However, for a productive deployment I'd see PostgreSQL (or MySQL) as a better suited system. By making this a template, I'd effectively cause a lot more checks to use it everywhere, thereby greatly increasing the complexity. Whereas, if people want to use it in the most efficient way, they'd probably rely on PostgreSQL or MySQL anyway.   
   
SQLite for me is the default, but not the recommended productive setup. Just like with the regular database connection.

lib/galaxy/model/data_tables/mapping.py
selten: SQL databases are optimized to work in rows. Your comment on not leveraging the row level optimization is actually incorrect. I leverage those. This dynamically scalable approach can be used with some more advanced SQL queries to create a single table with all the values. The following function creates a temporary table to handle these types of queries. https://github.com/galaxyproject/galaxy/blob/af7697c9ca7839fa1e6b4a5d05c7c5ebcc3a91ba/lib/galaxy/tools/data/__init__.py#L1183-L1214    
   
In the end this is more a question of where do you add most difficulty, in the queries or in the initial setup. I chose to get more difficult queries as that's easier to handle than a more difficult initial setup. The current setup doesn't harm scalability nor does it hamper possible queries that you may want to execute. It merely adds a few "WITH" queries in order to view the table in an easier format.

lib/galaxy/config/__init__.py
innovate-invent: The issue of forcing all tables into a single location is it is incompatible with deployments that mount in loc files rather than having a data manager generate them. If we are replacing loc files, the ability to keep the database file with the tool needs to be retained.

lib/galaxy/model/data_tables/mapping.py
innovate-invent: How does your generic table handle scanning a data table with a 1,000,000 rows?

selten: At ~200000 entries within SQLite it takes an average of 0.19 seconds to retrieve a row based on all entries within the same row. I ran 1000 random tests.

lib/galaxy/config/__init__.py
selten: This system runs alongside the loc files. It doesn't directly replace them.   
Technically what you want could be done by using the file path system.   
For instance a system like:
```xml
<tables>
    <!-- Locations of indexes in the Bowtie mapper format -->
    <table name="bowtie_indexes" comment_char="#">
        <columns>value, dbkey, name, path</columns>
        <file path="tool-data/bowtie_indices.sqlite" />
    </table>
</tables>
```
is what you mean if I am correct. Which indeed isn't possible with this pull request. However most of the code could technically be reused to accomplish this.

innovate-invent: ah, yes. That would be a reasonable alternative. It would need to support the full sqlalchemy connector url scheme though. In the event that a non-sqlite database is being used separately.

selten: The problem with that approach is, that especially when working with data managers and tools (and their updates), that Galaxy will have to merge all the SQLite files on run-time to get one coherent data table. As each data manager and each tool using a data table have a `data_table_conf.xml.sample` (or similar) file in their repository or the Tool Shed blocks it (I believe that's the case). So for each data manager or tool update it'd add one SQLite file for the same data table. So for querying you actually may lose some of the benefits of SQL.   
This is what currently happens for the Tabular Tool Data Table (loc files).

innovate-invent: Yes, I was wondering how you would handle that. Would it be better to remove the ability to split the table across multiple tools and just have a column that tracks the last tool to update the row?
This way when a tool is uninstalled the data for that tool can optionally be deleted from the table?

selten: I think I misunderstood your view on the data table system initially. I read replacing the loc files with a larger and replicable database. I didn't consider continuing to keep it on the filesystem. Which brings with it a slew of other problems that I've so far avoided... It may be as simple as adding support for the `<file path="tool-data/bowtie_indices.sqlite" />` elements though in its easiest form, whereby you may not have multiple files that need merging...   
I think that may require an extra PR, because that's a significant change to the data managers if the ability to merge files is lost in this scenario.

innovate-invent: The issue is it tries to cram all data into one database. I need the ability to choose different database solutions depending on the access patterns and size of data for each table.

selten: In that case you'd effectively need a special configuration file just for all your data tables that use databases all with connection strings. While this PR doesn't account for that and provides a general solution, it may also be an option to just store it all in a PostgreSQL database, whether it's big or small. PostgreSQL should be able to handle it fine regardless.

lib/galaxy/config/__init__.py
```diff
@@ -549,6 +549,12 @@ def _process_config(self, kwargs):
         self.database_engine_options = get_database_engine_options(kwargs)
         self.database_create_tables = string_as_bool(kwargs.get('database_create_tables', 'True'))
         self.database_encoding = kwargs.get('database_encoding')  # Create new databases with this encoding
+
+        # Database data tables connection
+        if not self.data_table_connection:  # Provide default if not supplied by user
```
jmchilton: Can you reuse ``get_database_engine_options`` with like a data_tables prefix? This would allow the admin to configure a lot more of this.

doc/source/admin/galaxy_options.rst
```diff
@@ -4491,4 +4491,19 @@
 :Type: str
 
 
+~~~~~~~~~~~~~~~~~~~~~~~~~
+``data_table_connection``
```
jmchilton: Can we do ``data_table_database_connection`` for consistency?

lib/galaxy/tools/data/__init__.py
```diff
@@ -896,11 +899,338 @@ def _remove_entry(self, values):
...
+    """
+    dict_collection_visible_keys = ['name']
+    type_key = 'database'
+
+    datatable_name = None
+    datatable_allow_duplicate_entries = True
+    sa_session = None
+    datatable_id = None
+
+    def __init__(self, config_element, tool_data_path, from_shed_config=False, filename=None, tool_data_path_files=None, other_config_dict=None, app=None):
```
jmchilton: Can you get by without passing ``app`` all the way through here. Maybe just pass ``app.data_tables_model`` (if available)? I think that is all you land up using (that and the tool_shed name link - which you can just assume you cannot load these if you don't have the model).

jmchilton:
Is there a way to write some integration tests for this - like subclass the ``test_data_manager.py`` tests to populate these? Or will these not work with data managers?

doc/source/admin/galaxy_options.rst
selten: I've renamed it, thanks for the suggestion!

selten: I've renamed it, thanks for the suggestion!

lib/galaxy/tools/data/__init__.py
selten: Thanks for the suggestion, I've now limited it to just the data_tables_model!

selten:
> Is there a way to write some integration tests for this - like subclass the `test_data_manager.py` tests to populate these? Or will these not work with data managers?

I added some tests now, these are fully compatible with data managers.

test/functional/tool-data/sample_tool_data_tables.xml
```diff
@@ -11,4 +11,7 @@
     <columns>value, dbkey, name, path</columns>
     <file path="${__HERE__}/fasta_indexes.loc" />
   </table>
+  <table name="testdatabase" comment_char="#" type="database">
```
jmchilton: Should this not have a comment_char?

selten: I do think that's not needed, I've now removed it.

PR #9876, opened 2020-06-13
Encrypt extra user preferences

nuwang:
This PR encrypts extra user preferences, allowing extra data such as ~passwords~API keys to be stored in user preferences.

While not all preferences need to be encrypted, with this PR, extra_user_preferences can be used to store sensitive information, such as the owncloud username/API key example shown below.
image

### DIFF
--- a/lib/galaxy/managers/users.py
+++ b/lib/galaxy/managers/users.py
@@ -1,6 +1,7 @@
 """
 Manager and Serializer for Users.
 """
+import json
 import logging
 import random
 import socket
@@ -315,7 +316,9 @@ class UserManager(base.ModelManager, deletable.PurgableManagerMixin):
 
     # ---- preferences
     def preferences(self, user):
-        return dict((key, value) for key, value in user.preferences.items())
+        return dict((key, value) if key != 'extra_user_preferences' else
+                    (key, json.dumps(user.get_extra_preferences(self.app.security)))
+                    for key, value in user.preferences.items())
 
     # ---- roles and permissions
     def private_role(self, user):
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -372,16 +372,25 @@ class User(Dictifiable, RepresentById):
         self.credentials = []
         # ? self.roles = []
 
-    @property
-    def extra_preferences(self):
+    def get_extra_preferences(self, security):
         data = {}
         extra_user_preferences = self.preferences.get('extra_user_preferences')
         if extra_user_preferences:
             try:
-                data = json.loads(extra_user_preferences)
+                decoded_user_prefs = security.decode_str(
+                    extra_user_preferences, kind="prefs")
+                data = json.loads(decoded_user_prefs)
             except Exception:
-                pass
+                try:
+                    # backward compatibility with unencrypted preferences
+                    data = json.loads(extra_user_preferences)
+                except Exception:
+                    pass
         return data
+
+    def set_extra_preferences(self, security, value):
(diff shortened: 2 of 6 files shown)


dannon:
Only read through this quickly, but my read is that this encrypts all user preferences using the instance secret key?  We probably (?) don't want to do this, because it prevents ever changing that key for the instance, in addition to providing little extra security.  

Additionally, if the machine the database is on is compromised, I would expect Galaxy's configfiles are in the open too, and with that this encryption is broken.

nuwang:
@dannon This encrypts only extra_preferences, not preferences themselves. I used id_secret assuming it would be stable for the lifetime of a Galaxy installation. If not, a different setting could be introduced.

Agree that this is not zero risk, but the idea was that the database being compromised alone is not sufficient to leak the data - both the machine running the database, and the machine containing the Galaxy config would need to be compromised. Secondly, the information store here needs to be decryptable, so that the system can take action on the user's behalf. Any information stored here would need to ideally be revokable API keys. This is a compromise between convenience and security, and I think the decision to use it or not can be left to individual installations?

dannon:
For secure credential storage we really need to use an external vault, which I thought was the longer term plan.

Yes, the id secret can change.  We actually had to rotate it on main one time, and realizing it broke all the pages (which was a bug -- they should not have been using encoded ids as links), I had to write and manually run migration code to fix it all it all.

I understand wanting to have data encrypted at rest; before we store real user passwords in preferences we should have that capability.  I just don't think this is the right way to go about it.  I don't think it's particularly likely that someone gains control of a reasonably secured (listening only to local connections for Galaxy, etc) database without having the config file that they could then use to decrypt the data.

nuwang:
I agree an external Vault would be better. This was a relatively low-cost way to do this in the interim which is roughly proportional to the sensitivity of the data we expect would be stored in it, at least for the use cases we had in mind, and the motivation that an actor would have to compromise said data. e.g. An attacker would probably stand to gain more from compromising the smtp password, oidc client secret or object store access key from the Galaxy configs, than they would from compromising this encryption key. As such, the idea was to put in a disclaimer to not use it for highly sensitive use-cases, and to use it based on the individual installation's ability to secure their system. There would definitely not be any compulsion to use it other than by choice.

For using a Vault itself, can you outline the plan for what you had in mind? If it's a matter of using an externalized vault when available, and falling back to the database when not, it may be a relatively simple change.

dannon:
I guess it's the 'roughly proportional' bit that I'm worried about here in terms of security.  Specifically, I worry that this provides the appearance of security of user preferences and saved secrets without actually adding much more than (to my estimation) preventing unintentional database administrator eyes on the data or the like.  Which, seeing the example above with the owncloud password, might still be a reasonable thing to do.

We don't have a concrete implementation plan for the vault other than it builds on the custos authnz work and can certainly design it with insecure database storage as a fallback -- that's a really good idea.  I don't have a timeline for it, but not imminently.

To be clear, I'm definitely not throwing a -1 on this, just trying to point out concerns I have with the approach, and give a little info on future plans.  I think we can probably make this work, with an eye towards the future development, but I do worry a little about the extra migration/maintenance burden and the problems admins might run into.

If we do want to do this, maybe use a separate secret like `user_preference_encryption_secret`?  This way it isn't entangled with the same secret we use for sessions, ids, etc., and they can be changed (or not) independently.  We'd also still need a mechanism like in the script I showed for migrating from one secret to another (or to/from unencrypted forms).

afgane:
To avoid needing custom migration scripts in the future for secrets such as these, I believe Fernet keys could be used and they serve exactly that purpose: https://cryptography.io/en/latest/fernet/

PR #9878, opened 2020-06-15
[20.05] Improved UploadModal state handling during initialization.

mvdbeek:
Backport of https://github.com/galaxyproject/galaxy/pull/9869.

Track whether datatypes and genomes have been set and use the LoadingSpan component to display information about this wait if the component is not yet ready.


jmchilton:
Sorry for targeting the wrong branch there :(

PR #9882, opened 2020-06-16
Adding user id to key so savedRules are specific for users

assuntad23:
Why: So that Saved Rules are specific to a user
What:  Added User ID to the key saved in Local Storage
How: Create a rule as one user, and save it; then try to access it as another user or as someone who is not logged in.

### DIFF
--- a/client/src/components/RuleBuilder/SavedRulesSelector.test.js
+++ b/client/src/components/RuleBuilder/SavedRulesSelector.test.js
@@ -30,7 +30,7 @@ describe("SavedRulesSelector", () => {
     });
 
     it("should emit a click event when a session is clicked", async () => {
-        wrapper.setProps({user: "test_user"})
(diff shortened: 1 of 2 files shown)


dannon:
@assuntad23 You may want to look into using sessionStorage here instead:  https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage

assuntad23:
@dannon Thanks for the suggestion! I had considered using that in the beginning, but I thought a user may want their rules to stick around a little bit longer, so I went with Local Storage instead. 

I'm not tied to it, and I don't actually know whether this feature will be used

dannon:
@assuntad23 Sounds good!

dannon:
Failing selenium tests are unrelated and fixed in https://github.com/galaxyproject/galaxy/pull/10216

PR #9883, opened 2020-06-17
Mount workflow nodes in v-for loop

guerler:
Currently, workflow nodes are inserted into the editor by manually mounting individual components. In other words, the `Workflow` component creates `Node` components by making calls to the `buildNode` function of the `manager` module. This PR replaces this process with a regular `v-for` loop. `Node` components are now linked to the `steps` property and their lifecycle is managed by `Vue`. Additionally this PR removes the `manager` module, and `jQuery` helpers from node and canvas position handling routines. This PR is another step towards a better workflow editor client architecture.


dannon:
As an incremental refactoring step this looks good; we'll want to revisit the model handling down the road.

PR #9884, opened 2020-06-17
[20.01] Input dataset module format filtering fix; 

dannon:
This would previously turn `fasta` into `f,,,,a,,,,s,,,...` and so on when you actually tried to run a workflow using this functionality. It was possible to increasingly proliferate commas on runs to the point where it crashed a webserver.

From 281961976a0

### DIFF
--- a/lib/galaxy/workflow/modules.py
+++ b/lib/galaxy/workflow/modules.py
@@ -661,7 +661,7 @@ class InputModule(WorkflowModule):
         if "format" in state:
             formats = state["format"]
             if formats:
-                formats = ",".join(formats)
+                formats = ",".join(listify(formats))
                 state["format"] = formats
         state = json.dumps(state)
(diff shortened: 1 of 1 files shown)


mvdbeek (approved):
Excellent! That's nice fix.

nsoranzo:
The issue was introduced in commit 281961976a03689c824325d00891998dab19003b , can you target release_20.01 instead?

dannon:
@nsoranzo Yeah, I saw that and linked it above earlier, noticing the all-numerical short hash, heh.  Anyway, I can change the base, sure.

nsoranzo (approved):
Thanks! Commit id was not in the original email from GitHub, sorry I didn't see you added it!

jmchilton:
Thanks for the fix, sorry about that.

PR #9891, opened 2020-06-18
Added missing dev flag to hot module reload script, sourcemaps work now

Nerdinacan:
Webpack hot module reload script was missing the "-d" flag which causes source maps to be generated.

Developing in chrome without source maps is no fun.


dannon (approved):

OlegZharkov (approved):
@Nerdinacan I just tried it. Thanks a lot, that's very useful

dannon:
Thanks; ran into this before myself and never got around to getting the fix back upstream.

PR #9892, opened 2020-06-18
[20.05] Output SRA manifest as tabular dataset

mvdbeek:
This adds a new datatype that is mostly the same as Tabular, but that skips rendering of the first line, which contains the column names (that depend on user selection on the SRA run selector).
(This should be useful for other datatypes that have or set `column_name` as well).

Screenshot 2020-06-18 at 18 54 59

This also makes the sra_source tool output a tsv file instead of a csv file, which should make operations on this file more streamlined in Galaxy (i.e use in the rule builder, relabel, tagging, etc) and that makes it easier to feed the manifest into the fasterq-dump tool.

It would be great of we can get this into 20.05 for the SRA webinar (https://galaxyproject.org/news/2020-06-galaxy-update/#galaxy-and-the-ncbi-sequence-read-archive-sra) next week. Calling this a bug is a bit of a stretch ... so let me know if I should better target dev.

### DIFF
--- a/client/galaxy/scripts/mvc/dataset/data.js
+++ b/client/galaxy/scripts/mvc/dataset/data.js
@@ -316,7 +316,7 @@ var TabularDatasetChunkedView = Backbone.View.extend({
         _.each(
             chunk.ck_data.split("\n"),
             function (line, index) {
-                if (line !== "") {
+                if (line !== "" && index >= chunk.data_line_offset) {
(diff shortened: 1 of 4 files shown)


dannon (approved):

dannon:
Relevant tests were passing prior to the minor last tweak allowing datatype upload, and I tested locally as well.

PR #9898, opened 2020-06-20
[20.05] Fix webpack import proxy prefix

dannon:
This fixes webpack dynamic imports when served with a prefix (like `<server>/mygalaxy/`)


mvdbeek:
The js-lint are probably real ?

dannon:
@mvdbeek Real, but we want to ignore it; this is a special case.  Let me add an exception.

mvdbeek (approved):

PR #9899, opened 2020-06-21
Fix DeprecationWarning: Flags not at the start of the expression

nsoranzo:
Fix:

```
lib/galaxy/util/inflection.py:188: DeprecationWarning: Flags not at the start of the expression '^(?i)(ox)$'
    match = re.search(rules[rule][0], word, re.IGNORECASE)
```

Also modernise the code a bit.

### DIFF
--- a/lib/galaxy/util/inflection.py
+++ b/lib/galaxy/util/inflection.py
@@ -138,13 +138,22 @@ class English(Base):
 
     This is the default Inflector for the Inflector obj
     """
+    uncountable_words = ('equipment', 'information', 'rice', 'money', 'species', 'series', 'fish', 'sheep', 'sms')
+
+    irregular_words = {
+        'person': 'people',
+        'man': 'men',
+        'child': 'children',
+        'sex': 'sexes',
+        'move': 'moves'
+    }
 
     def pluralize(self, word):
         '''Pluralizes English nouns.'''
 
         rules = [
             ['(?i)(quiz)$', '\\1zes'],
-            ['^(?i)(ox)$', '\\1en'],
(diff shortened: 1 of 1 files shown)


lib/galaxy/util/inflection.py
```diff
@@ -138,13 +138,22 @@ class English(Base):
 
     This is the default Inflector for the Inflector obj
     """
+    uncountable_words = ('equipment', 'information', 'rice', 'money', 'species', 'series', 'fish', 'sheep', 'sms')
```
jdavcs, line 141: I think species, fish (unless meaning food) and sheep are countable nouns

jdavcs: oh, I see, it's an existing inflection script. And these values refer to how the plural form should be constructed - in which case the logic is correct, I think. So, never mind. (still, poor choice of variable name by the script's original author, I think.)

nsoranzo: Yep, in https://github.com/bermi/Python-Inflector/pull/2/files (from which I took some ideas) for Spanish they are called `non_changing_words`. I can change it here too if you want.

jdavcs: I don't mind leaving it as is. It's a very minor thing, it's good enough; besides touching the linguistic side of it might lead us down the garden path - I'm sure there are many more nouns that are pluralized according to the same rule.

jdavcs (approved):
I was going to make a few more suggestions, but I think they are more appropriate as a standalone PR. I'll ping you when I submit it.

ic4f:
Cool refactoring (pythonization!), @nsoranzo!

PR #9902, opened 2020-06-23
Trim the inflector

jdavcs:
This is a follow-up to #9899. One thing led to another. Overall:
1. I've removed the code that was not used and not tested, and added tests for the rest.
2. I've made a few edits to the inflection rules (This is not the best inflection algorithm, but it's good enough for our purposes. I've tried to fix the most obvious problems. Because English.)
3. I've simplified the rest. We don't need all those abstractions.

See commit messages for details on edits.

Ping @nsoranzo


lib/galaxy/util/inflection.py
```diff
@@ -4,347 +4,116 @@
...
-        '''Transform a string to its unaccented and underscored
-        version ready to be inserted in friendly URLs'''
-        return self.Inflector.urlize(text)
+    def _transform(self, rules, word, pluralize=True):
+        return self._handle_nonchanging(word) or self._handle_irregular(word, pluralize=pluralize) or self._apply_rule(rules, word) or word
 
-    def demodulize(self, module_name):
-        return self.Inflector.demodulize(module_name)
+    def _handle_nonchanging(self, word):
+        if word.lower() in self.NONCHANGING_WORDS:
```
nsoranzo: I think your modified version will pluralize `misinformation`, `subspecies` and so on, right?

lib/galaxy/util/inflection.py
```diff
@@ -4,347 +4,116 @@
...
+        if number_of_records != 1:
+            return self.pluralize(word)
+        return word
 
-    def urlize(self, text):
-        '''Transform a string to its unaccented and underscored
-        version ready to be inserted in friendly URLs'''
-        return self.Inflector.urlize(text)
+    def _transform(self, rules, word, pluralize=True):
+        return self._handle_nonchanging(word) or self._handle_irregular(word, pluralize=pluralize) or self._apply_rule(rules, word) or word
```
nsoranzo:
```suggestion
        return self._handle_nonchanging(word) or self._handle_irregular(word, pluralize=pluralize) or self._apply_rules(rules, word) or word
```

lib/galaxy/util/inflection.py
```diff
@@ -4,347 +4,116 @@
...
+                form_a, form_b = form_b, form_a
+            match = re.search('(' + form_a + ')$', word, re.IGNORECASE)
+            if match:
+                return re.sub('(?i)' + form_a + '$', match.expand('\\1')[0] + form_b[1:], word)
 
-    def foreignKey(self, class_name, separate_class_name_and_id_with_underscore=1):
-        ''' Returns class_name in underscored form, with "_id" tacked on at the end.
-        This is for use in dealing with the database.'''
-        return self.Inflector.foreignKey(class_name, separate_class_name_and_id_with_underscore)
+    def _apply_rule(self, rules, word):
```
nsoranzo:
```suggestion
    def _apply_rules(self, rules, word):
```

lib/galaxy/util/inflection.py
jdavcs: Oh yes, it will.. I can change that part back. The point it, computational linguistics is hairy and complex, and rules usually don't work very well. "heep" (not heap) was just one example I came up with that broke the suffix replacement approach. Your examples are definitely more relevant to our domain than "heep", so I'm, of course, fine with using the suffix approach. But, again, I think the overall algorithm is good enough in our case.

nsoranzo: I don't think the previous code was broken actually:
```python
        for uncountable_word in self.uncountable_words:
            if lower_cased_word[-1 * len(uncountable_word):] == uncountable_word:
                return word
```
When `uncountable_word` is `sheep` and `lower_cased_word` is `heep`, `lower_cased_word[-1 * len(uncountable_word):]` is `heep`, so the test would be correctly `False`.

A better version of the above test would be `lower_cased_word.endswith(uncountable_word)`

nsoranzo:
I kind-of liked the idea that we may support other languages, but maybe we introduce a better interface when/if we cross that bridge.

lib/galaxy/util/inflection.py
```diff
@@ -4,347 +4,119 @@
...
+    def _transform(self, rules, word, pluralize=True):
+        return self._handle_nonchanging(word) or self._handle_irregular(word, pluralize=pluralize) or self._apply_rule(rules, word) or word
 
-    def demodulize(self, module_name):
-        return self.Inflector.demodulize(module_name)
+    def _handle_nonchanging(self, word):
+        lower_cased_word = word.lower()
+        # Check if word is an item or the suffix of any item in NONCHANGING_WORDS
+        for nonchanging_word in self.NONCHANGING_WORDS:
+            if lower_cased_word[-len(nonchanging_word):] == nonchanging_word:
```
nsoranzo:
```suggestion
            if lower_cased_word.endswith(uncountable_word):
```

lib/galaxy/util/inflection.py
jdavcs: Yes, you're right. I got it backwards. And yes, of course `endswith` is better.

ic4f:
> I kind-of liked the idea that we may support other languages, but maybe we introduce a better interface when/if we cross that bridge.

Yes, I agree.

nsoranzo (approved):

ic4f:
Thank you for reviewing, @nsoranzo!

Issue #9903, opened 2020-06-23
WorkflowStepExecutionTracker.record_success can be extremely slow


mvdbeek:
This is the calling stack for one of usegalaxy.org's workflow handlers.
The critical part in our code is probably https://github.com/galaxyproject/galaxy/blob/278a0333724e84209f317bc200c86ad5165f575a/lib/galaxy/tools/execute.py#L409

sqlalchemy emits a backref that fetches all implicit collection jobs here, and I think this happens for each job within the collection ... so the runtime grows exponentially here with every collection member. Just looking at the logs one might even think that the handler is dead since nothing seems to happen.

```
    __buffer_rows (sqlalchemy/engine/result.py:1434)
    _init_metadata (sqlalchemy/engine/result.py:1412)
    __init__ (sqlalchemy/engine/result.py:775)
    get_result_proxy (sqlalchemy/dialects/postgresql/psycopg2.py:583)
    _execute_context (sqlalchemy/engine/base.py:1267)
    _execute_clauseelement (sqlalchemy/engine/base.py:1103)
    _execute_on_connection (sqlalchemy/sql/elements.py:293)
    execute (sqlalchemy/engine/base.py:984)
    _execute_and_instances (sqlalchemy/orm/query.py:3430)
    __iter__ (sqlalchemy/ext/baked.py:444)
    all (sqlalchemy/ext/baked.py:544)
    _emit_lazyload (sqlalchemy/orm/strategies.py:892)
    <lambda> (<string>:1)
    _load_for_state (sqlalchemy/orm/strategies.py:750)
    get (sqlalchemy/orm/attributes.py:722)
    set (sqlalchemy/orm/attributes.py:994)
    append (sqlalchemy/orm/attributes.py:751)
    emit_backref_from_scalar_set_event (sqlalchemy/orm/attributes.py:1478)
    fire_replace_event (sqlalchemy/orm/attributes.py:1041)
    set (sqlalchemy/orm/attributes.py:1018)
    __set__ (sqlalchemy/orm/attributes.py:272)
    record_success (galaxy/tools/execute.py:409)
    execute_single_job (galaxy/tools/execute.py:74)
    execute (galaxy/tools/execute.py:102)
    execute (galaxy/workflow/modules.py:1757)
    _invoke_step (galaxy/workflow/run.py:266)
    invoke (galaxy/workflow/run.py:190)
    __invoke (galaxy/workflow/run.py:83)
[...]

PR #9904, opened 2020-06-23
Set implicit collection jobs only once per collection of mapped over jobs

mvdbeek:
This can get very expensive for large collections.

Needs some performance checks, and i'm curious if that will pass on its own. xref #9903


mvdbeek:
Merged as part of https://github.com/galaxyproject/galaxy/pull/10019

PR #9905, opened 2020-06-23
Vue rewrite of the data-library-folder

OlegZharkov:
This PR is a preliminary step for library-folder lazy loading pagination. Instead of implementing pagination on top of legacy backbone code, I assume it would be just better to Vuefy Library folder functionality.

The state so far:

- **The table:**
	 - [x] Dataset/Folder icon
	 - [x] Unrestricted/Private/Restricted dataset icons
	 - [x] Subfolder/Datasets-description on `name` click
	 - [x] Empty Folder View
	 - [x] Data Type
	 - [x] Proper size rendering
	 - [x] Date Updated in "ago" format
	 - [x] Manage button
	 - [x] Sorting
	 - [x] Selecting (UPDATE: this giff is old already, focus outline is fixed and not shown on table headers anymore):
another

	 - [x] Pagination bottom bar with total pages (no lazy loading at the moment, probably next PR)
	 - [x] Edit mode:
	 - [x] 'Parent folder' btn

	 - [x] Linkify description
	 - [x] Undelete dataset/folder

	 - [x] Shrinking long description: (UPDATE: unlike in giff, "..." added and "text on hover"-area extended):
        
Peek 2020-06-25 20-24

- **The top bar**:
	 - [x] Search
	 - [x] Home button
	 - [x] Add new folder button 
	 - [x] Add new datasets
		  - [x] from history
		  - [x] from path
		  - [x] from userdir
		  - [x] from importdir
	 - [x] Export to history
		  - [x] as Datasets
		  - [x] as Collections
	 - [x] Download selected (zip, tar, tar.gz)
	 - [x] Delete selected
	 - [x] Details (partially recycled legacy code)
	 - [x] Include Deleted
	 - [x] Path Bar

If someone is aware of how to use backbone code in Vue component, please let me know, that knowledge will make my life a ton easier. Re-implementing all "include datasets" modals at once, probably won't be something pleasant.

P.S.
This PR also ships advanced pagination on backend, but not on frontend yet (Probably next PR)


OlegZharkov:
I remove [WIP] for now and would highly appreciate some comment, especially, if someone can try it and play around with this interface.

What is covered by Selenium:
 - new folder creation, edit desc, ensure description shrinks 
 - upper folder
 - import dataset from path, add dataset from file, add dataset from history
 - download 
 - delete
 - show details
 - path-bar

I am still kind of afraid to delete old Backbone implementation. I didn't touch it ... yet, I just moved everything on the new path `/library/folders/{id}`.
 Alternatively (on request), I can preserve old Backbone implementation, which can be turned on in the config. Like in galaxy.yaml

Thanks and if you notice some bugs (that most probably will occur eventually), please let me know! I would be happy to fix and polish, whatever you think is necessary.

dannon:
@OlegZharkov Thanks for all the work on this, I'll try to take it for a spin later.  Re: backbone implementation -- if this has feature parity, we should just delete it.

OlegZharkov:
> @OlegZharkov Thanks for all the work on this, I'll try to take it for a spin later. Re: backbone implementation -- if this has feature parity, we should just delete it.

it is a feature parity. But I am kinda afraid to lose that. I am still thinking to preserve it and hide legacy implementation in config ... just in case.

dannon:
@OlegZharkov Yeah, no worries!  :)   It's in the git history forever, we can always refer back to it or restore things if need be.

client/src/components/LibraryFolder/LibraryFolder.vue
```diff
@@ -0,0 +1,514 @@
...
+                @updateSearch="updateSearchValue($event)"
+                @refreshTable="refreshTable"
+                @fetchFolderContents="fetchFolderContents($event)"
+                @deleteFromTable="deleteFromTable"
+                :folderContents="folderContents"
+                :include_deleted="include_deleted"
+                :folder_id="folder_id"
+                :selected="selected"
+                :metadata="folder_metadata"
+            ></FolderTopBar>
```
guerler: Very minor, but these should be self closing tags.

client/src/components/LibraryFolder/LibraryFolder.vue
```diff
@@ -0,0 +1,514 @@
...
+                :per-page="perPage"
+                :current-page="currentPage"
+                selectable
+                :select-mode="selectMode"
+                @row-selected="onRowSelected"
+                ref="folder_content_table"
+                @filtered="onFiltered"
+                show-empty
+            >
+                <template v-slot:empty="">
```
guerler: Is the `=""` needed. I assume it can be removed.

client/src/components/LibraryFolder/LibraryFolder.vue
```diff
@@ -0,0 +1,514 @@
...
+                            :value="row.item.name"
+                            rows="3"
+                        ></textarea>
+                    </div>
+                    <div v-else-if="!row.item.deleted">
+                        <a :href="createContentLink(row.item)">{{ row.item.name }}</a>
+                    </div>
+                    <!-- Deleted Item-->
+                    <div v-else>
+                        <div style="color: grey;">{{ row.item.name }}</div>
```
guerler: Static style should be removed.

client/src/components/LibraryFolder/LibraryFolder.vue
```diff
@@ -0,0 +1,514 @@
...
+                            v-else
+                            class="form-control input_folder_description"
+                            :ref="'description' + row.item.id"
+                            :value="row.item.description"
+                            rows="3"
+                        ></textarea>
+                    </div>
+                    <div v-else>
+                        <div class="description-field" v-if="getMessage(row.item)">
+                            <div v-if="getMessage(row.item).length > 40 && !expandedMessage.includes(row.item.id)">
```
guerler: The numeric value should be part of the `data` function.

client/src/components/LibraryFolder/LibraryFolder.vue
```diff
@@ -0,0 +1,514 @@
...
+                <!-- Description -->
+                <template v-slot:cell(message)="row">
+                    <div v-if="row.item.editMode">
+                        <textarea
+                            v-if="row.item.isNewFolder"
+                            class="form-control input_folder_description"
+                            :ref="'description' + row.item.id"
+                            v-model="row.item.description"
+                            rows="3"
+                        ></textarea>
```
guerler, line 100: These tags should also be self closing.

client/src/components/LibraryFolder/TopToolbar/FolderTopBar.vue
```diff
@@ -0,0 +1,382 @@
+<template>
+    <div>
+        <div id="path-bar" />
+
+        <div class="form-inline d-flex align-items-center mb-2">
+            <a class="mr-1 btn btn-secondary" href="/library/list" data-toggle="tooltip" title="Go to first page">
```
guerler: The href needs to be derived using `getAppRoot()` otherwise it'll fail in cases where the Galaxy root is not at `/`.

client/src/components/LibraryFolder/TopToolbar/FolderTopBar.vue
```diff
@@ -0,0 +1,382 @@
...
+                <form class="form-inline">
+                    <b-input-group size="sm">
+                        <b-form-input
+                            class="mr-1"
+                            v-on:input="updateSearch($event)"
+                            type="search"
+                            id="filterInput"
+                            placeholder="Search"
+                        >
+                            >
```
guerler: Is this `>` intentional?

client/src/components/LibraryFolder/TopToolbar/FolderTopBar.vue
```diff
@@ -0,0 +1,382 @@
...
+                        <button @click="showDetails" class="primary-button toolbtn-show-locinfo" type="button">
+                            <font-awesome-icon icon="info-circle" />
+                            Details
+                        </button>
+                    </span>
+                    <div class="form-check logged-dataset-manipulation mr-1" v-if="logged_dataset_manipulation">
+                        <b-form-checkbox
+                            id="checkbox-1"
+                            :checked="include_deleted"
+                            v-on:input="toggle_include_deleted($event)"
```
guerler: Recommending to use the `@` shorthand here and above.

client/src/components/LibraryFolder/TopToolbar/FolderTopBar.vue
```diff
@@ -0,0 +1,382 @@
...
+            library_import_dir: false,
+            allow_library_path_paste: false,
+            list_genomes: [],
+            list_extensions: [],
+            // datatype placeholder for extension auto-detection
+            auto: {
+                id: "auto",
+                text: "Auto-detect",
+                description:
+                    "This system will try to detect the file type automatically." +
```
guerler: Recommending to use es6 multline string here instead of concatenation.

client/src/components/LibraryFolder/TopToolbar/FolderTopBar.vue
```diff
@@ -0,0 +1,382 @@
...
+        this.library_import_dir = Galaxy.config.library_import_dir;
+        this.allow_library_path_paste = Galaxy.config.allow_library_path_paste;
+        if (
+            this.user_library_import_dir !== null ||
+            this.allow_library_path_paste !== false ||
+            this.library_import_dir !== null
+        ) {
+            this.multiple_add_dataset_options = true;
+        }
+        const contains_file_or_folder = this.folderContents.find((el) => el.type === "folder" || el.type === "file");
```
guerler, line 235: `===` is usually only required when comparing against `0` in order to avoid implicit conversion. In these cases `==` is sufficient.

client/src/components/LibraryFolder/TopToolbar/FolderTopBar.vue
```diff
@@ -0,0 +1,382 @@
...
+                isNewFolder: true,
+                type: "folder",
+                name: "",
+                description: "",
+            });
+            this.$emit("refreshTable");
+        },
+        downloadData(format) {
+            const { datasets, folders } = this.findCheckedItems();
+            if (this.selected.length === 0) {
```
guerler, line 286: Since there is no risk of implicit conversion here `==` is sufficient.

client/src/mvc/library/library-dataset-view.js
```diff
@@ -671,7 +671,7 @@ var LibraryDatasetView = Backbone.View.extend({
                     <% _.each(item.get("full_path"), function(path_item) { %>
                         <% if (path_item[0] != item.id) { %>
                             <li class="breadcrumb-item">
-                                <a title="Return to this folder" href="#/folders/<%- path_item[0] %>">
+                                <a title="Return to this folder" href="/library/folders/<%- path_item[0] %>">
```
guerler: Please use `getAppRoot()` here too.

guerler:
Thank you so much for working on this. This is a very nice step forward. I noticed a few minor style issues other than that it looks good to me and I realize that this is a transitional PR. The complete conversion of the Library Client code to Vue will take additional steps. Besides the issues above I noticed a few UI glitches which possibly can be solved within this PR or in follow-ups. 
1) The Edit/Manage buttons should have some margin in-between.
2) When selecting a Library, one can see the tool panel appear and then slide out of the screen (this is probably not an issue caused by this PR)
3) The sort buttons are not aligned with the text.
4) The check icon (select/unselect all) should be an actual checkbox icon. This would more clearly indicate that it is clickable.

Screen Shot 2020-08-28 at 12 34 44 PM

OlegZharkov:
> Thank you so much for working on this. This is a very nice step forward. I noticed a few minor style issues other than that it looks good to me and I realize that this is a transitional PR. The complete conversion of the Library Client code to Vue will probably take additional steps. Besides the issues I mentioned above in the code, I noticed a few UI glitches which possibly can be solved within this PR or in follow-ups.
> 


Thanks a lot!!!
that's a very great review! I will fix these things in the following days. I am very very glad, that I am not the only one who didn't find any major bugs :sweat_smile:  (like some functionality issues)
My intention was to completely Vuefy library folder, not really a transitional, but a final one. I wanted to remove the old folder code and I probably should do it in a few commits, shouldn't I?

> 1. The Edit/Manage buttons should have some margin in-between.

great catch :+1: 
> 2. When selecting a Library, one can see the tool panel appear and then slide out of the screen (this is probably not an issue caused by this PR)

I genuinely thought, it's a feature. It was a part of API. If it's not intended, I will dig into it and remove this animation.
> 3. The sort buttons are not aligned with the table header text.

might be tricky... but you're right, it should be aligned 
> 4. The check icon (select/unselect all) should be an actual checkbox icon. This would more clearly indicate that it is clickable.

yeah, that is kinda controversial. I wanted to make it a bit different from original library, but after a while I was also thinking to change it back

client/src/components/LibraryFolder/TopToolbar/FolderTopBar.vue
OlegZharkov: yeah, this is inherited from legacy code. I will change :+1:

client/src/components/LibraryFolder/LibraryFolder.vue
OlegZharkov: do you want this check as a separate function?

client/src/components/LibraryFolder/TopToolbar/FolderTopBar.vue
OlegZharkov: thanks :sweat_smile: 
that's a typo

martenson:
> I wanted to make it a bit different from original library, but after a while I was also thinking to change it back

We've all been there @OlegZharkov 😂

client/src/components/LibraryFolder/TopToolbar/FolderTopBar.vue
OlegZharkov: Actually... why do we care whether there is a risk of implicit conversion or not here? Why do we have to use `==`, if there is no risk?

client/src/components/LibraryFolder/TopToolbar/FolderTopBar.vue
dannon: `==` will work, but I can't think of a good reason not to be explicit and require type be a string?

client/src/components/LibraryFolder/TopToolbar/FolderTopBar.vue
dannon: We should lean *towards* using explicit equality `===` where we can, not relying on whether or not `==` is sufficient.  `===` is the right thing to use here.

OlegZharkov: I agree with @dannon, `===` should be preferable

OlegZharkov:
Thanks a lot for the review @guerler!!!
I applied suggestions and deleted the legacy code. I kept `library-folder-view.js`, because we still need it on `folder_permissions` path... we can vuefy that in another PR and remove `library-folder-view.js` as well 

Does it look better now?

OlegZharkov:
Can we attach 20.09 milestone there, please? It's ready and passes everything...

if so, someone powerful should do it

mvdbeek:
Is this good for another review / merge @OlegZharkov ?

dannon:
I started a review but need to finish; there are some things I would like to address prior to merge.

OlegZharkov:
> Is this good for another review / merge @OlegZharkov ?

yes, it's good. I did rebase, because I wanted to fix selenium

mvdbeek:
Never a bad idea to rebase, but we test merging of the tip of the branches. re-running just the failed tests should be enough to see the new results

OlegZharkov:
> Never a bad idea to rebase, but we test merging of the tip of the branches. re-running just the failed tests should be enough to see the new results

Well, I cannot re-run failed tests. Also I wanted to have this https://github.com/galaxyproject/galaxy/pull/10216

dannon:
I have a few tweaks for this that I'd like to get in during the freeze, but I'm going to go ahead and merge this as-is so we can keep iterating on it.

OlegZharkov:
Thanks a lot!! @dannon @guerler

dannon:
@OlegZharkov Thank you, this is a great start towards rewriting libraries.

martenson:
@OlegZharkov 👏

PR #9908, opened 2020-06-24
Typo in the galaxy.yml.sample

gmauro:
typo in the galaxy.yml.sample


bgruening:
I guess you need to fix it here: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/config_schema.yml

ic4f:
@gmauro Thank you - good catch! We don't edit this file directly - it is generated by running `make config-rebuild` from the installation directory. The error is here: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/config_manage.py

gmauro:
OK @ic4f , I close here and edit there.

PR #9910, opened 2020-06-24
Rev pulsar-galaxy-lib for latest fixes.

jmchilton:


nsoranzo:
The corresponding change to `Pipfile` is included in https://github.com/galaxyproject/galaxy/pull/9912

PR #9911, opened 2020-06-24
[WIP] Allow occasionally polling of potentially lost Pulsar MQ jobs

jmchilton:
Matching merged and published Pulsar PR @ https://github.com/galaxyproject/pulsar/pull/228

Rudely clobbered from @natefoo commit history at https://github.com/natefoo/galaxy/commits/pulsar-amqp-status-poll and rebased on dev - which had some significant changes to how polling is determined and such (for ITs on Pulsar).

### DIFF
--- a/lib/galaxy/jobs/runners/pulsar.py
+++ b/lib/galaxy/jobs/runners/pulsar.py
@@ -3,6 +3,7 @@
 More information on Pulsar can be found at https://pulsar.readthedocs.io/ .
 """
 
+import copy
 import errno
 import logging
 import os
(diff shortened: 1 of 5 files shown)


lib/galaxy/jobs/runners/pulsar.py
```diff
@@ -856,7 +882,13 @@ def __async_update(self, full_status):
                 galaxy_job_id = remote_job_id
             job, job_wrapper = self.app.job_manager.job_handler.job_queue.job_pair_for_id(galaxy_job_id)
             job_state = self._job_state(job, job_wrapper)
-            self._update_job_state_for_status(job_state, full_status["status"], full_status=full_status)
+            old_state = job_state.old_state
+            _job_state = self._update_job_state_for_status(job_state, full_status["status"], full_status=full_status)
+            if (self.runner_params.params['poll_jobs_older_than'] and _job_state is not None
+                    and old_state == _job_state.old_state):
+                # Received an update, no state change (or the state is terminal) so reset update_time
+                job_wrapper.touch(job)
```
jmchilton, line 889: @natefoo I'm nervous about repeatedly polling via the MQ until this is reached. If there is some transient, short term issue with the MQ that might result in a lot of messages. That in turn might cause more problems - overloading the queue, the first status telling Galaxy that everything is fine and subsequent ones issuing lost statuses, stuff like that - resulting in failing jobs that completed fine or jobs failing in very misleading ways?

I wonder if just calling the touch call here should be done when issuing the update instead? Seems vaguely less problematic - but I have no idea, that is just intuition.

natefoo: Queue overload is most likely not a problem (we churn through tens of thousands of messages that can accumulate when Pulsar stops consuming). Even message order wouldn't be an issue, as long as a job can't be changed from terminal to non-terminal? The bigger issue would be if Pulsar has forgotten about a successfully completed job and returns an error state when there's an ok state message sitting somewhere in the queue. Can we have Pulsar hang on to the final state indefinitely?

PR #9914, opened 2020-06-24
Fix name of container resolver method name.

jmchilton:
These can be YAML files no problem.

### DIFF
--- a/lib/galaxy/tool_util/deps/containers.py
+++ b/lib/galaxy/tool_util/deps/containers.py
@@ -204,9 +204,9 @@ class ContainerRegistry(object):
             log.debug("Unable to find config file '%s'", conf_file)
             return self.__default_containers_resolvers()
         plugin_source = plugin_config.plugin_source_from_path(conf_file)
-        return self.__parse_resolver_conf_xml(plugin_source)
+        return self.__parse_resolver_conf(plugin_source)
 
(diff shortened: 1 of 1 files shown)


lib/galaxy/tool_util/deps/containers.py
```diff
@@ -204,9 +204,9 @@ def __build_container_resolvers(self, app_info):
             log.debug("Unable to find config file '%s'", conf_file)
             return self.__default_containers_resolvers()
         plugin_source = plugin_config.plugin_source_from_path(conf_file)
-        return self.__parse_resolver_conf_xml(plugin_source)
+        return self.__parse_resolver_conf(plugin_source)
```
nsoranzo: I don't think we need name mangling.

```suggestion
        return self._parse_resolver_conf(plugin_source)
```

lib/galaxy/tool_util/deps/containers.py
```diff
@@ -204,9 +204,9 @@ def __build_container_resolvers(self, app_info):
...
 
-    def __parse_resolver_conf_xml(self, plugin_source):
+    def __parse_resolver_conf(self, plugin_source):
```
nsoranzo:
```suggestion
    def _parse_resolver_conf(self, plugin_source):
```

nsoranzo (approved):

PR #9917, opened 2020-06-25
Do not automatically overwrite the output labels for subworkflows

guerler:
This PR improves enforcing unique subworkflow output labels by avoiding automated labeling of workflow outputs since that can easily lead to duplicated labels. This change only effects newly inserted subworkflows and the label/name visible to the user remains unchanged.

Example shown in release 20.01:

Screen Shot 2020-06-26 at 3 17 03 PM

### DIFF
--- a/lib/galaxy/workflow/modules.py
+++ b/lib/galaxy/workflow/modules.py
@@ -488,7 +488,6 @@ class SubWorkflowModule(WorkflowModule):
                     for data_output in data_outputs:
                         data_output_uuid = data_output.get('uuid') or object()
                         if data_output['name'] == workflow_output['output_name'] or data_output_uuid == workflow_output_uuid:
-                            data_output['label'] = label
                             data_output['name'] = label
                             # That's the right data_output
                             break
--- a/test/unit/workflows/test_modules.py
+++ b/test/unit/workflows/test_modules.py
@@ -233,10 +233,8 @@ def test_subworkflow_new_outputs():
     assert len(outputs) == 2, len(outputs)
     output1, output2 = outputs
     assert output1["name"] == "out1"
-    assert output1["label"] == "out1"
     assert output1["extensions"] == ["input"]
     assert output2["name"] == "4:out_file1", output2["name"]
-    assert output2["label"] == "4:out_file1", output2["label"]
 
 
 def _construct_steps_for_map_over():


lib/galaxy/workflow/modules.py
```diff
@@ -481,9 +481,8 @@ def get_all_outputs(self, data_only=False):
                 step = subworkflow_dict['steps'][order_index]
                 data_outputs = step['outputs']
                 for workflow_output in step['workflow_outputs']:
-                    label = workflow_output['label']
-                    if not label:
-                        label = "%s:%s" % (order_index, workflow_output['output_name'])
+                    label = workflow_output['label'] or workflow_output['output_name']
+                    label = "%s:%s" % (order_index, label)
```
mvdbeek: I think this is intentional, the order_index is not necessary stable, and the label is important to identify connections.
As is this would probably break existing workflows. What is the usecase you're trying to enable ?

mvdbeek: I should add I that added `label = "%s:%s" % (order_index, workflow_output['output_name'])` as a fallback to make it more obvious how to use outputs from a subworkflow, since the requirement to label them wasn't (and still isn't) exposed in the UI. I am still not 100% sure that this was the right choice.

guerler: Thanks for reviewing this. The bug I observe is reproducible when a 3-fold nested subworkflow is created. I created a 1st workflow containing a simple tool like `Add Column`, then imported it as a subworkflow into a 2nd workflow and finally imported the 2nd workflow as subworkflow into a 3rd workflow. When importing the 1st into the 2nd workflow labels are added to the workflow outputs and when the 2nd workflow is imported into the 3rd workflow those labels remain unchanged i.e. additional indices are not added.

The fixed labels for subworkflows are added at: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/workflow/modules.py#L491.
I am working on a fix, maybe the condition validating the uuids for workflow outputs is the issue.

guerler: I dug a little deeper and noticed that the underlying reason for this issue is that the new client accepts and preserves subworkflow output labels if they are specified in the backend. As a consequence, the workflow output label is not further augmented by adding the step number if the workflow is nested again since the corresponding output label will be explicitly available (https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/workflow/modules.py#L485).
In contrast, the previous workflow displayed the subworkflow output label but did not parse it back when the workflow was saved. I adjusted this PR to fix this issue and restore the equivalent behavior by not parsing the label to the client. The client will fallback on the name which contains the same value without having to explicitly store a label which would interfere with consecutive workflow nesting iterations.

mvdbeek (approved):
I think that is correct now.

bgruening (approved):

PR #9925, opened 2020-06-26
[20.05] Enable regex validator to validate lists of values.

blankenberg:
Previously would just always fail due to inability to regex on a list.


bernt-matthias:
Seems like a good idea. Do you see a possibility to add test(s)? Maybe testing for the list case and the single value case.

Also an extension of the corresponding documentation would be super nice. Maybe just referring to the test...

bernt-matthias:
Thanks for the test. I was just wondering in which cases a list is given to the validator? Is it `<select/param multiple="true"...>`?

Also wondering if the other validators may receive lists...?

mvdbeek (approved):

PR #9930, opened 2020-06-29
[20.05] explicitly cast oidc providers to a list

almahmoud:
Without the explicit casting, that line returns `TypeError: 'dict_keys' object does not support indexing` in py3.
`.keys()` returns an indexable list in py3 but a non-idexable `dict_keys` in py3.
https://stackoverflow.com/a/17322707

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/users.py
+++ b/lib/galaxy/webapps/galaxy/api/users.py
@@ -1,6 +1,7 @@
 """
 API operations on User objects.
 """
+import copy
 import json
 import logging
 import re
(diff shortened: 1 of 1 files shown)


lib/galaxy/webapps/galaxy/controllers/root.py
```diff
@@ -89,7 +89,7 @@ def login(self, trans, redirect=None, is_logout_redirect=False, **kwd):
                 len(trans.app.config.oidc) == 1 and
                 len(trans.app.auth_manager.authenticators) == 0
                 and is_logout_redirect is False):
-            provider = trans.app.config.oidc.keys()[0]
+            provider = list(trans.app.config.oidc.keys())[0]
```
nsoranzo: Not a performance issue here, but you can skip the (potentially expensive) list creation with:
```suggestion
            provider = next(iter(trans.app.config.oidc.keys()))
```

nsoranzo:
Did you mean to target the 20.05 branch? It's targeting `dev` at the moment.

almahmoud:
I did mean to target 20.05, that was my bad. I'll open a PR targetting the right branch.

almahmoud:
Opened https://github.com/galaxyproject/galaxy/pull/9931 (didn't realize force pushing would close this PR)

PR #9928, opened 2020-06-29
Created unit test to upload/download/delete file into Galaxy

kxk302:
This test is used to load test Galaxy via Locust.

PR #9933, opened 2020-06-29
irods methods now create/destroy irods connections.

kxk302:
Reading irods forum, it seems the best practice is for apps to not maintain a connection to irods, but rather create/destroy them as needed. 

We are seeing some irods Sentry errors. One theory is that irods server drops stale connections (without client knowing). After this, the first call would times out and a new connections is created and all is good until the same issue repeats. 

Creating/dropping connections, should solve this problem.  

@mvdbeek I think this PR should be reviewed before https://github.com/galaxyproject/galaxy/pull/9928

### DIFF
--- a/lib/galaxy/objectstore/irods.py
+++ b/lib/galaxy/objectstore/irods.py
@@ -205,7 +205,10 @@ class IRODSObjectStore(DiskObjectStore, CloudConfigMixin):
             _config_dict_error('extra_dirs')
         self.extra_dirs.update(extra_dirs)
 
-        self._initialize()
+        if irods is None:
+            raise Exception(IRODS_IMPORT_MESSAGE)
+
+        self.home = "/" + self.zone + "/home/" + self.username
         log.debug("irods __init__ %s", reload_timer)
 
     def shutdown(self):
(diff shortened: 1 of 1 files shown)


mvdbeek:
This looks odd. You probably don't want to run shutdown on every method and instead make the connection a property if this is really the right thing to do, but this seems to run counter to the existence of the session pool in the python-irodsclient. Are these irods forums publically available ? Shouldn't python-irodsclient handle such dropped connections ? Did you manage to get a handle on this on the irods server side ? I don't think the Galaxy object store layer is the right place to manage this.

kxk302:
> This looks odd. You probably don't want to run shutdown on every method and instead make the connection a property if this is really the right thing to do, but this seems to run counter to the existence of the session pool in the python-irodsclient. Are these irods forums publically available ? Shouldn't python-irodsclient handle such dropped connections ? Did you manage to get a handle on this on the irods server side ? I don't think the Galaxy object store layer is the right place to manage this.

I agree that this looks odd. The irods forum is public, I believe. See if you can access it (https://groups.google.com/g/irod-chat). There is an old thread that discusses this connection issue. I asked for an update on this topic. 

I need to dig into irods-client code further.

Anyways, the Sentry errors are still around, and when Nate and I looked into it, it seems the un-used connections are dropped. Creating/dropping a connection for every irods call should circumvent this issue.

mvdbeek:
Might be worth checking with the python-irodsclient folks, I suppose this must be a common problem then. Did you manage to reproduce this locally ?

kxk302:
Created a new PR for this (https://github.com/galaxyproject/galaxy/pull/9941)

Issue #9727, opened 2020-05-06
Add job_id column to Dataset table.


jmchilton:
HDCAs already have one but HDAs (and LDDAs) need something comparable. I think it belongs on the dataset instead of the HDA but I might be wrong? The target job_id should essentially be the result of creating_job.id - linked below:

https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/model/__init__.py#L2830

I think the hardest part here is the database migration - which will take a long time on main and I'm not sure how to translate all that logic to SQL but if you can that would be great. Another challenge will be to ensure that all new datasets have this set - so make sure new tool created outputs have this set, collection elements, copied datasets between histories, datasets inside of copied collections, imported datasets, uploaded datasets, discovered datasets, etc...

ic4f:
I assume the line reference to `creating_job` should be https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/model/__init__.py#L2953

jdavcs:
Fixed in #10222

PR #9934, opened 2020-06-30
[WIP] Add job_id column to dataset table; use trigger to update

jdavcs:
Re: #9727 

This is a prototype. It includes the new `job_id` column on the `dataset` table, and a trigger that handles setting its value on new records for HDAs. It works locally.

Thoughts:
1. Regarding the choice for the `job_id` column between dataset and dataset instance, I think dataset makes more sense. My (potentially faulty) reasoning: all datasets are the result of running a job, but not all dataset instances are created by a job (copying a history is one such example). Thus, a "created by" relationship *always* exists between a job and a dataset, but *not* always between a job and a dataset instance.
2. So far, I haven't been able to come up with a reasonable way to set the `job_id` value on new dataset records in the code using the sqlalchemy session. The job is flushed to the database *after* the dataset, so when the dataset is flushed, the `job id` that should be stored with it simply doesn't exist (this is the case at least in the context of the local job runner). But I'm sure there is a way to do that.
3. Using the `job_to_output_dataset` table to tie a dataset to its "creating job" seems a logical choice. But I may be wrong.
4. I may be wrong about all of this.

To do (provided this is not incorrect):
1. Do the same for `job_to_output_library_dataset`
2. Replace trigger with a sqlalchemy session implementation (assuming it's a better way to it)
3. Handle existing records. After looking at the data on main (thank you @natefoo!) I suspect there may be orphaned datasets (dataset records without corresponding jobs). We could, maybe, update what's possible and have a programmatic fallback for cases when the `job_id` column is null. 
4. Tests.

Any feedback, tips, etc. would be most appreciated, as always.


ic4f:
Jenkins doesn't like the trigger SQL. It runs for me locally (on postgresql). I'll debug if the trigger approach has merit (pending feedback).

ic4f:
Restarting this in new branch; implementing in SA (not db trigger), so no need for this version.

PR #9938, opened 2020-06-30
Implement more Galaxy Markdown components.

jmchilton:
- invocation-time - The iso time of the workflow invocation time.
- generate-time - The iso time of the time of generation of the Markdown.
- generate-galaxy-version - The Galaxy MAJOR_VERSION at the time of generation of the Markdown.
- dataset-type - The ext/format/datatype of the target dataset.
- dataset-name - The name of the target dataset.

Builds on the amazing work from @OlegZharkov in #9769


OlegZharkov:
If it depends on #9769, you can push directly in my branch. I don't mind. After BCC submission deadline (this week), I am only planning to implement vue-tests in #9769, but no major changes, just tests and small refactoring.

lib/galaxy/managers/markdown_util.py
```diff
@@ -122,6 +123,14 @@ def _remap(container, line):
...
                 hda = hda_manager.get_accessible(object_id, trans.user)
                 rval = self.handle_dataset_info(line, hda)
+            elif container == "history_dataset_type":
+                assert object_id is not None
+                hda = hda_manager.get_accessible(object_id, trans.user)
+                rval = self.handle_dataset_type(line, hda)
+            elif container == "history_dataset_name":
+                assert object_id is not None
+                hda = hda_manager.get_accessible(object_id, trans.user)
+                rval = self.handle_dataset_name(line, hda)
```
OlegZharkov, line 133: we really should put these things in a separate function and run it against array of keywords. Because we have pretty much 100 identical lines.

OlegZharkov (approved):
thanks John

lib/galaxy/managers/markdown_util.py
jmchilton: It is is sort of common in visitor patterns but you're probably right - this could be made more precise for sure.

PR #9941, opened 2020-06-30
session is a property now

kxk302:
We create/destroy session for each irods call.


kxk302:
@mvdbeek This is the code change for having session as a property, and creating/destroying it for each irods call.

PR #9944, opened 2020-06-30
Using context manager for irods session object

kxk302:


lib/galaxy/objectstore/irods.py
```diff
@@ -1,6 +1,7 @@
 """
 Object Store plugin for the Integrated Rule-Oriented Data Store (iRODS)
 """
+from contextlib import contextmanager
```
mvdbeek: This should go to line 8.

mvdbeek (approved):

lib/galaxy/objectstore/irods.py
kxk302: Moved the import.

PR #9947, opened 2020-07-01
Strip leading whitespace before calling urlparse

### LINKED ISSUE #8196: Trim whitespace in URL upload box?

jdavcs:
Fixes #8196 
Also replace six with urllib

What's happening: with a leading whitespace, `urlparse` will return the following value:
```python
ParseResult(scheme='', netloc='', path=' https://docs.python.org/3/library/urllib.parse.html', params='', query='', fragment='') 
```
Trailing whitespace is irrelevant.

ping @martenson

### DIFF
--- a/lib/galaxy/tools/actions/upload_common.py
+++ b/lib/galaxy/tools/actions/upload_common.py
@@ -6,9 +6,9 @@ import socket
 import tempfile
 from collections import OrderedDict
 from json import dump, dumps
+from urllib.parse import urlparse
 
 from six import StringIO
-from six.moves.urllib.parse import urlparse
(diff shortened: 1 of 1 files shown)


ic4f:
selenium test failures unrelated.

ic4f:
On second thought, added `lstrip()` to initial check (see last commit). If user submits a URL w/leading white space, let's assume it's intended as a URL and trim it.

PR #9949, opened 2020-07-02
Modify email address domain extraction

jdavcs:
This modifies the handling of `email_domain_allowlist_file` and `email_domain_blocklist_file`. The former matches the full domain of the email address, while the latter matches the base domain only. Rationale: we want to block disposable email addresses, which may contain irrelevant subdomains (i.e., `@foo.spam.com`). However, we do not want to match addresses with ccTLDs (e.g., `@*.org.au`) against their base only (`org.au`) on the `allowlist_file`. (thanks, @natefoo)

What's included:
- refactor domain extraction
- add tests
- edit description in config schema, rebuild


jmchilton:
Looks solid!

PR #9950, opened 2020-07-02
[20.05] Fixes to package dependency declarations.

jmchilton:


packages/test_driver/requirements.txt
```diff
@@ -1,4 +1,5 @@
 galaxy-data
+galaxy-web-apps
 nose
 pytest
```
nsoranzo:
```suggestion
pyyaml
sqlalchemy-utils
six
```

packages/test_driver/requirements.txt
```diff
@@ -1,4 +1,5 @@
 galaxy-data
+galaxy-web-apps
```
nsoranzo:
```suggestion
galaxy-app
galaxy-data
galaxy-test-base
galaxy-tool-util
galaxy-util
galaxy-web-apps
galaxy-web-framework
```

jmchilton: Shouldn't we just rely on galaxy-web-apps to recursively bring those in?

nsoranzo: I think the correct approach is to add all packages from which this package directly imports a module, in the same way we list both bioblend and requests in `lib/galaxy/dependencies/pipfiles/default/Pipfile` even if the former depends on the latter.

jmchilton: I'm skeptical it is appropriate here... but I do see the value generally. I've merged this change - thanks so much and thanks for the explanation.

nsoranzo (approved):

jmchilton:
Thanks for the merge @nsoranzo !

PR #9951, opened 2020-07-02
Refactor toward real cohesion toward testing packages. (1/n)

jmchilton:
For publishing useful tests to PyPI, there needs to be a base TestCase class for API, Framework, Selenium, etc... tests that doesn't depend on the Galaxy app - and instead just knows and utilizes the API.

The relevant TODO from the code is in ``lib/galaxy_test/api/_framework.py``.

```
# TODO: implement a remote-only version of ApiTestCase if galaxy_test.driver isn't
# available. Most of these tests do not depend on Galaxy internals.
from galaxy_test.driver.api import ApiTestCase

__all__ = ('ApiTestCase', )
```

Practically speaking, this means better separation of utilities and environment parsing out of ``lib/galaxy_test/driver`` and into ``lib/galaxy_test/base``. This is because the former depends on ``galaxy-web-apps`` and the latter doesn't. It also means making sure the API/Selenium/Framework tests do not depend on internals - things that need a fixed Galaxy configuration should be pushed into integration test suite and the API tests should be runnable against any Galaxy instance.

The first commit here is a step toward better separation described above and the second commit loosens the requirements of the API tests so we no longer need a security object when running the tests (further reducing the dependencies of the galaxy-test-base package also).

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/dataset_collections.py
+++ b/lib/galaxy/webapps/galaxy/api/dataset_collections.py
@@ -4,6 +4,7 @@ import routes
 
 from galaxy import managers
 from galaxy.exceptions import ObjectNotFound
+from galaxy.managers.base import decode_id
 from galaxy.managers.collections_util import (
     api_payload_to_create_params,
     dictify_dataset_collection_instance,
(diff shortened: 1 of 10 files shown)


nsoranzo:
@jmchilton Should the `status/WIP` label be removed?

jmchilton:
@nsoranzo Yup, thanks for the catch!

bgruening:
Thanks John!

PR #9952, opened 2020-07-03
[20.05] Various small fixes for galaxy packages. 

jmchilton:

### DIFF
--- a/packages/containers/tests/containers
+++ b/packages/containers/tests/containers
@@ -1,0 +1,1 @@
+../../../test/unit/containers
--- a/packages/containers/tests/test_containers.py
+++ b/packages/containers/tests/test_containers.py
@@ -1,1 +1,0 @@
-../../../test/unit/test_containers.py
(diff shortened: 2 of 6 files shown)


nsoranzo:
Wouldn't it be simpler to move `test/unit/test_containers.py` to `test/unit/containers/test_containers.py` ? With the symlink these tests are run twice by pytest.

jmchilton:
Opps - looks like I duplicated the test. I meant to create a directory called containers and just symlink that so we don't have to manage individual files. Ideally each package could correspond to one or a couple unit test directories (i.e. https://github.com/galaxyproject/galaxy/blob/dev/packages/data/tests/datatypes) and the unit tests wouldn't have to be managed file by file. I clearly messed up this commit though.

jmchilton:
Sorry about that confusion - does this look any better?

nsoranzo (approved):

bgruening:
How do we trigger uploads to pypi? https://pypi.org/project/galaxy-objectstore

nsoranzo:
It's still a manual process AFAIK.

PR #9955, opened 2020-07-04
Rework client datatype handling logic for reuse

jmchilton:
For #9809 I need to be able to determine if one short extension is a subtype of another short extension. This logic was duplicated a couple places in the workflow components so I wanted to extract it for reuse and eliminate this duplication. While I was in there I also implemented a new API that lets the client fetch the needed datatype data for the workflow editor in one call instead of two, I also implemented a test case for the extract logic and the new API endpoint (both in the client and in the API tests).


bgruening (approved):
Cool, restarted a test, let's see if it completes this time.

PR #9956, opened 2020-07-04
[20.05] Allow setting metadata_strategy per destination

mvdbeek:
This allows instances that mix classic job runners with pulsar to use
metadata_strategy: extended on destinations that are not served by
pulsar. Calling this a bug since it prevents instances that use pulsar
and another job runner from using extended metadata.

### DIFF
--- a/lib/galaxy/config/sample/galaxy.yml.sample
+++ b/lib/galaxy/config/sample/galaxy.yml.sample
@@ -1741,6 +1741,15 @@ galaxy:
   # database.
   #enable_job_recovery: true
 
+  # Determines how metadata will be set. Valid values are `directory`,
+  # `extended` and `legacy`. In extended mode jobs will decide if a tool
+  # run failed, the object stores  configuration is serialized and made
+  # available to the job and is used for writing output datasets to the
+  # object store as part of the job and dynamic output discovery (e.g.
+  # discovered datasets <discover_datasets>, unpopulated collections,
+  # etc) happens as part of the job.
+  #metadata_strategy: directory
+
   # Although it is fairly reliable, setting metadata can occasionally
   # fail.  In these instances, you can choose to retry setting it
   # internally or leave it in a failed state (since retrying internally
--- a/lib/galaxy/config/sample/galaxy.yml.sample
+++ b/lib/galaxy/config/sample/galaxy.yml.sample
@@ -1743,7 +1743,7 @@ galaxy:
 
   # Determines how metadata will be set. Valid values are `directory`,
   # `extended` and `legacy`. In extended mode jobs will decide if a tool
-  # run failed, the object stores  configuration is serialized and made
+  # run failed, the object stores configuration is serialized and made
   # available to the job and is used for writing output datasets to the
   # object store as part of the job and dynamic output discovery (e.g.
   # discovered datasets <discover_datasets>, unpopulated collections,
--- a/lib/galaxy/config/sample/job_conf.xml.sample_advanced
+++ b/lib/galaxy/config/sample/job_conf.xml.sample_advanced
@@ -401,6 +401,17 @@
                try it early - it will slightly speed up local jobs by
                embedding metadata calculation in job script itself.
           -->
+          <param id="metadata_strategy">directory</param>
+          <!-- If set overwrites metadata_strategy (configurable in galaxy.yml).
+               Valid values are `directory` (default), `extended` and `legacy`.
+               Setting metadata_strategy to extended requires that all object stores
+               that can be written to store datasets by uuid. In extended mode jobs
+               will decide if a tool run failed, the object stores configuration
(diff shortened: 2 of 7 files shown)


lib/galaxy/config/sample/galaxy.yml.sample
```diff
@@ -1741,6 +1741,15 @@ galaxy:
   # database.
   #enable_job_recovery: true
 
+  # Determines how metadata will be set. Valid values are `directory`,
+  # `extended` and `legacy`. In extended mode jobs will decide if a tool
+  # run failed, the object stores  configuration is serialized and made
```
bgruening:
```suggestion
  # run failed, the object stores configuration is serialized and made
```

lib/galaxy/config/sample/job_conf.xml.sample_advanced
```diff
@@ -401,6 +401,17 @@
                try it early - it will slightly speed up local jobs by
                embedding metadata calculation in job script itself.
           -->
+          <param id="metadata_strategy">directory</param>
+          <!-- If set overwrites metadata_strategy (configurable in galaxy.yml).
```
bgruening:
```suggestion
          <!-- If set, overwrites `metadata_strategy` (configurable in galaxy.yml).
```

lib/galaxy/config/sample/job_conf.xml.sample_advanced
```diff
@@ -401,6 +401,17 @@
...
+               Valid values are `directory` (default), `extended` and `legacy`.
+               Setting metadata_strategy to extended requires that all object stores
+               that can be written to store datasets by uuid. In extended mode jobs
+               will decide if a tool run failed, the object stores configuration
+               is serialized and made available to the job and is used for writing output
+               datasets to the object store as part of the job and dynamic output discovery
+               (e.g. discovered datasets <discover_datasets>, unpopulated collections, etc)
+               happens as part of the job.
```
bgruening:
```suggestion
               Valid values are `directory` (default), `extended` and `legacy`.
               Setting `metadata_strategy` to `extended` requires that all object stores
               that can be written to store datasets by uuid. In `extended` mode jobs
               will decide if a tool run failed, the object stores configuration
               is serialized and made available to the job and is used for writing output
               datasets to the object store as part of the job and dynamic output discovery
               (e.g. discovered datasets <discover_datasets>, unpopulated collections, etc)
               happens as part of the job.
```

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -2638,6 +2638,18 @@ mapping:
...
 
+      metadata_strategy:
+        type: str
+        required: false
+        default: directory
+        desc: |
+          Determines how metadata will be set. Valid values are `directory`, `extended` and `legacy`.
+          In extended mode jobs will decide if a tool run failed, the object stores
+          configuration is serialized and made available to the job and is used for
+          writing output datasets to the  object store as part of the job and dynamic
```
bgruening:
```suggestion
          writing output datasets to the object store as part of the job and dynamic
```

test/integration/test_pulsar_embedded_metadata.py
```diff
@@ -16,6 +16,10 @@ class EmbeddedMetadataPulsarIntegrationInstance(integration_util.IntegrationInst
     @classmethod
     def handle_galaxy_config_kwds(cls, config):
         config["job_config_file"] = EMBEDDED_PULSAR_JOB_CONFIG_FILE
+        config['object_store_store_by'] = 'uuid'
+        # metadata_strategy is being set to directory in embedded_pulsar_metadata_job_conf.yml,
+        # since extended_metadata doesn't work yet on pulsar
+        config['metadata_strategy'] = 'extended'
```
bgruening, line 22: What are you testing here? Shouldn't this fail because pulsar does not work yet?

bgruening (commented):
Cool thanks.

test/integration/test_pulsar_embedded_metadata.py
mvdbeek: I tried to explain that in the comment above, this tests that the `metadata_strategy` can be overridden in the destination.

bgruening: Yeah, but once this is fixed the test will not test this anymore, correct? Or in other words, once pulsar is capable of extended-metadata, we do not test the overwriting anymore. Should we worry about this?

mvdbeek: I don't think we should at this point. I can move this to a separate test if you like.

bgruening:
Looks like the failing tests are relevant :(

jmchilton:
Awesome - thanks for much for working in this direction. I really dropped the ball here and I appreciate y'all working on this so much!

PR #9957, opened 2020-07-04
[20.05] Update pysam to 0.16.0.1

innovate-invent:
Pysam 0.15.4 fails to install


mvdbeek:
Thanks @innovate-invent!

PR #9959, opened 2020-07-05
Progress toward proper packaging of galaxy-web-framework package.

jmchilton:
-  ``galaxy.web.framework.webapp`` used something from ``galaxy-app`` so we can't have the DAG we're aiming for (galaxy-webapps depends on galaxy-app depends on galaxy-framework). So I've refactored this file to ``galaxy.webapps.base.webapp`` (which I think is not imported directly anywhere in galaxy-app - so this is better).
- Added a stub of unit test to get this package under CI.


packages/web_framework/requirements.txt
```diff
@@ -1 +1,3 @@
 galaxy-data
+galaxy-web-stack
+paste
```
bgruening, line 8: The test fails because we are missing the python babel library. Can it be that we need to add it here?

packages/web_framework/requirements.txt
```diff
@@ -1 +1,4 @@
+babel
 galaxy-data
+galaxy-web-stack
+paste
```
nsoranzo:
```suggestion
babel
Cheetah3
galaxy-data
galaxy-util
galaxy-web-stack
Mako
MarkupSafe
paste
requests
Routes
six
SQLAlchemy
WebOb
```

test/unit/web/framework/test_framework_base.py
```diff
@@ -0,0 +1,18 @@
+from galaxy.web.framework.base import WebApplication
```
nsoranzo: Should this file path be `test/unit/web_framework/test_framework_base.py` instead?

packages/web_framework/tests/web
```diff
@@ -0,0 +1 @@
+../../../test/unit/web
```
nsoranzo: I think we should have `packages/web_framework/tests/web_framework` pointing to `../../../test/unit/web_framework` instead.

nsoranzo (commented):
Should the `web_apps` package called `webapps` ? That's how we spell it everywhere else and it's not packaging a `web.apps` module but `webapps`.

nsoranzo:
>     * `galaxy.web.framework.webapp` used something from `galaxy-app` so we can't have the DAG we're aiming for

I think it would be nice to have the DAG added to a `package/README.md` or somewhere for reference.

packages/web_framework/requirements.txt
jmchilton: I'll commit this - but I remain nervous given that we have like very specific requirements for SQLAlchemy for instance right (we pin it often)? We're duplicating that dependency over multiple applications instead of inheriting the single requirement from galaxy-data - which I sort of conceptually think of as bringing in our SQL Alchemy dependence.

Maybe we should find a longer term solution toward synchronizing Galaxy's Pipfile with the individual packages - maybe that would make me feel less yucky about this.

Do you have any ideas? I guess my two ideas would be to create Pipfiles for each individual project and then write a tool to combine them or to extend the Pipfile metadata with metadata about which Galaxy packages each dependency relies on. I had been leaning toward the former, but if you're wanting us to list each of these dependencies explicitly like this I'm guessing the latter would be a lot less duplication.

jmchilton:
> I think it would be nice to have the DAG added to a package/README.md or somewhere for reference.

In https://github.com/galaxyproject/galaxy/pull/7313 the architecture docs were moved to training-material, I think there will be a nice diagram there after https://github.com/galaxyproject/training-material/pull/1907 is merged. The diagram is: https://github.com/galaxyproject/training-material/blob/ed530ead3bdd10c21752fe504923b0baa6f86c7f/topics/dev/images/core_packages.plantuml.svg. Happy to copy the SVG into the packages directory or something if you'd like.

packages/web_framework/requirements.txt
nsoranzo: My question would be: how will we actually be going to install Galaxy in the future?
1) as a collection of packages, in which case the pinning to exact versions will be done with pipenv (or poetry) on the top-level package
2) as a monolithic app (i.e. like now), in which case we would need to merge the individual package requirements into a Pipfile, and then pin to exact versions as usual. For the merge step, I suppose we could that with just `cat` and `sort -uf`

nsoranzo:
> > I think it would be nice to have the DAG added to a package/README.md or somewhere for reference.
> 
> In #7313 the architecture docs were moved to training-material, I think there will be a nice diagram there after [galaxyproject/training-material#1907](https://github.com/galaxyproject/training-material/pull/1907) is merged. The diagram is: https://github.com/galaxyproject/training-material/blob/ed530ead3bdd10c21752fe504923b0baa6f86c7f/topics/dev/images/core_packages.plantuml.svg. Happy to copy the SVG into the packages directory or something if you'd like.

Oh nice, that's absolutely fine by me, we could probably point to the slides from `CONTRIBUTING.md`

> Should the `web_apps` package called `webapps` ? That's how we spell it everywhere else and it's not packaging a `web.apps` module but `webapps`.

What do you think about this comment of mine?

packages/web_framework/requirements.txt
jmchilton: That is a big question and I don't know that we can predict how everything will work in practice. I think the plan was that the ``galaxy`` package on PyPI would be published to by ``packages/meta/`` - which would depend on the individual packages (or really just galaxy-web-apps).

As like a starting point, I would imagine we could pin very specific packages for ``galaxy`` (i.e. our pinned requirements now) and keep everything else unpinned (except where they are needed). A second approach to that might be ``pip install galaxy[pinned]`` which would give you pinned versions of everything. I don't know if you can hide pinned versions behind a qualifier like that - but that would be nice IMO.

I'm not sure how we organize our dependencies for individual packages should depend a lot on those details though? I'm probably missing something though.

jmchilton:
> Should the web_apps package called webapps ? That's how we spell it everywhere else and it's not packaging a web.apps module but webapps.

I think I just thought it looked funny given the other packages were ``galaxy-web-stack`` and ``galaxy-framework`` - dashes between every word. I don't have a strong opinion though and I'm happy to change it.

packages/web_framework/requirements.txt
nsoranzo: > I'm not sure how we organize our dependencies for individual packages should depend a lot on those details though?

I agree it shouldn't (since we agree exact pins are appropriate only for the full application, not the packages). I asked the question simply because in case of option 1) we don't need to merge the various package requirements into a single Pipfile.

nsoranzo:
> > Should the web_apps package called webapps ? That's how we spell it everywhere else and it's not packaging a web.apps module but webapps.
> 
> I think I just thought it looked funny given the other packages were `galaxy-web-stack` and `galaxy-framework` - dashes between every word. I don't have a strong opinion though and I'm happy to change it.

True, but I think I prefer to keep it consistent with our current spelling and module name (could be another PR if you prefer, of course).

jmchilton:
> True, but I think I prefer to keep it consistent with our current spelling and module name (could be another PR if you prefer, of course).

Given this one is finally green - I'll take you up on that and I'll do it in my next packaging PR. Thanks for your input and for thinking through this packing stuff - you insight is really appreciated.

packages/web_framework/requirements.txt
jmchilton: > 1) we don't need to merge the various package requirements into a single Pipfile.

I see - I was assuming the git based install would be common ... for a long time after had packages available. And still available for a long time after that - I suspect then we'd still need to keep our scripts and such in place for managing dependencies.

jmchilton:
> Oh nice, that's absolutely fine by me, we could probably point to the slides from CONTRIBUTING.md

👍 to this also! I'll see if I can work something into my next PR.

dannon:
This appears to have broken the reports webapp, I think?

PR #9960, opened 2020-07-06
Decouple API and Selenium tests from galaxy-test-driver (and the whole app!)

jmchilton:
This is in some ways the ultimate separation of launching and configuring Galaxy instances for tests (galaxy-test-driver) from being able to interact with a running Galaxy for tests (galaxy-test-base). The latter does not require the whole app - just galaxy-tool-util (the old galaxy-lib). Being able to pip install the API and Selenium tests without bringing in the whole app is important to make them usable outside the context of the app - for testing remote, running servers for instance.

If galaxy-test-driver is available the test cases will still spawn a Galaxy instance if needed, but this is not longer required for import or use.

### DIFF
--- a/lib/galaxy/util/__init__.py
+++ b/lib/galaxy/util/__init__.py
@@ -1707,6 +1707,15 @@ def download_to_file(url, dest_file_path, timeout=30, chunk_size=2 ** 20):
                 f.write(chunk)
 
 
+class classproperty(object):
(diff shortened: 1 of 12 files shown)


bgruening (approved):

bgruening:
Thanks John! This will be a great talk!

PR #9962, opened 2020-07-06
actually `set -e` for non-legacy tools

bernt-matthias:
before the proper default strict mode was only used if the command block was absent

### DIFF
--- a/lib/galaxy/tool_util/parser/xml.py
+++ b/lib/galaxy/tool_util/parser/xml.py
@@ -488,12 +488,14 @@ class XmlToolSource(ToolSource):
 
     def parse_strict_shell(self):
         command_el = self._command_el
-        if command_el is not None:
-            return string_as_bool(command_el.get("strict", "False"))
-        elif self.legacy_defaults:
-            return False
+        if self.legacy_defaults:
+            default = "False"
         else:
-            return True
+            default = "True"
+        if command_el is not None:
+            return string_as_bool(command_el.get("strict", default))
+        else:
+            return string_as_bool(default)
 
     def parse_help(self):
         help_elem = self.root.find('help')
--- a/lib/galaxy/tool_util/parser/xml.py
+++ b/lib/galaxy/tool_util/parser/xml.py
@@ -488,7 +488,7 @@ class XmlToolSource(ToolSource):
 
     def parse_strict_shell(self):
         command_el = self._command_el
-        if self.legacy_defaults:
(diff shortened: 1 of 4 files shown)


mvdbeek:
That's a good one ... now the question is should we hide the actually working implementation behind a newer profile version ? We may potentially be breaking tools here that might have just ignored a non-zero exit code. Can we run this against the iuc and see how many tools would break there ?

bernt-matthias:
> now the question is should we hide the actually working implementation behind a newer profile version ?

Guess this would be the least intrusive option.

> Can we run this against the iuc and see how many tools would break there ?

https://github.com/bernt-matthias/tools-iuc/runs/842267887

bernt-matthias:
First run did not work out .. this should https://github.com/bernt-matthias/tools-iuc/runs/842309398

bernt-matthias:
Finally https://github.com/bernt-matthias/tools-iuc/actions/runs/160322986

- now its 44 test(s) out of 3032 (with enasearch disabled)
- before 24 test(s) out of 2998 (last completed CI workflow) 

Difficult to compare. Here are some tools that now fail: fastq_dump, fasterq_dump, sam_dump, scikit_bio_diversity_beta_diversity, ont_fast5_api_single_to_multi_fast5, fasttree, ncbi_eutils_*,

mvdbeek:
So the sratools fail because there is an error when getting the version ... so as is in this PR we might break otherwise working tools. Maybe it would make sense to add a `|| true` where we template the version_command. I checked if something 
 like 0d933c0065a161e6e1820d36448baccb55e9658e works and it does. What do you think about that ?

bernt-matthias:
> What do you think about that ?

Seems like a workaround, but problems with version commands would also not be detected automatically anymore.

mvdbeek:
Right, we could hide that under a 20.09 flag if you like. But given that these tools are widely used I don't think we should merge this as is now. As I see it the options are now to have `set -e` active by default if there is a command block only on >= 20.09, or we add that workaround and make the workaround conditional on < 20.09.

mvdbeek (approved):
I guess my preference would have gone for just making `|| true` conditional on the < 20.09 profile, but there's no 100% correct choice here. This will definitely not break tools with an upgrade to 20.09. Thanks a lot for digging this up and fixing it @bernt-matthias!

mvdbeek:
Thanks a lot @bernt-matthias!

PR #9963, opened 2020-07-06
Added doc explaining what happens when you upload a file to Galaxy.

kxk302:
This documents the process to enqueue a job that does the file upload. There will be future documentation on how the job is dequeued/processed.


kxk302:
@jmchilton This documentation is somewhat inline with the video series that you just published. Would you mind taking a look at this PR? Thanks!

PR #9964, opened 2020-07-06
More packaging touches.

jmchilton:
Builds on https://github.com/galaxyproject/galaxy/pull/9959 and addresses comments about renaming web-apps -> webapps and README link.

Also starts the process of categorizing all the unit tests into their proper packages and improving the testing done as part of the package testing CI.


packages/util/setup.py
```diff
@@ -61,7 +61,7 @@ def get_var(var_name):
 
 
 test_requirements = [
-    # TODO: put package test requirements here
+    "respones",
```
nsoranzo:
```suggestion
    "responses",
```

nsoranzo:
Can you rebase now that https://github.com/galaxyproject/galaxy/pull/9959 is merged?

nsoranzo:
Some relevant test failures in https://app.circleci.com/pipelines/github/galaxyproject/galaxy/8745/workflows/aa457609-e122-4a17-b8ef-2701980390a3/jobs/95867/steps

jmchilton:
I keep bumping into the problem of how to share unit test helpers between packages. I think I'm going to refactor some helper methods into like a test_helpers Python file for each package and just place it in ``lib/galaxy``. If we land up hating that - I think it is plausible to place all the unit tests below ``lib/galaxy_test/unit`` in a namespace package and publish parts of this with each package to ship the unit tests with the package. We could also ship the unit test helpers in their own packages after we do that refactoring (e.g. ``galaxy-objectstore-tests``) but having everything structured better ahead of that refactoring would be good even if we have to slightly pollute the non-test packages.

nsoranzo (approved):

nsoranzo:
I've removed the `status/WIP` label, I hope that's OK. Feel free to merge when it's green.

jmchilton:
It is ready to go - the last failing test is unrelated.

PR #9953, opened 2020-07-03
Small tweaks to packaging of unit test utilities.

jmchilton:
`clean_multiline_string` was used in multiple places across modules in different packages so I simply refactored it into `galaxy.util` and added a test case (used in multiple places and has tests a lower bar than some other code in `galaxy.util`).

The `tempfilecache` abstraction was only used in one place, so I've refactored it right into that file.

Together these should simplify the package tests for `galaxy-data` (and hopefully other packages also).

PR #9966, opened 2020-07-07
Refactor more unit tests into the galaxy-data package.

jmchilton:
If that lib/galaxy/objectstore/test_helpers.py file looks off - see my comment here (https://github.com/galaxyproject/galaxy/pull/9964#issuecomment-654515780) for more context and alternative ideas.

I tried to refactor more model store testing into here but ultimately failed on a test case that depends on metadata tool stuff in the rest of Galaxy (https://github.com/jmchilton/galaxy/commit/dd14f939d93dbf0cdd0ff7ad2dfac4cb5e3c1c89) - future iterations of this should work on decomposing these things with more cohesion and better separation of those model tests that remain.


nsoranzo:
Naive question: can the stuff in `lib/galaxy/objectstore/test_helpers.py` be merged into `test/unit/unittest_utils/galaxy_mock.py` ?

nsoranzo:
Also maybe move `test/unit/data/test_id_encode_decode.py` to `test/unit/data/security/test_id_encode_decode.py` ?

jmchilton:
> Naive question: can the stuff in lib/galaxy/objectstore/test_helpers.py be merged into test/unit/unittest_utils/galaxy_mock.py ?

If you look at the imports in galaxy_mock.py - it needs stuff from all over the app - it needs to live at one of the highest packages of the package DAG. It shouldn't therefore be used to test stuff in galaxy-data and galaxy-objectstore. Part of the reason why I started ripping apart unittest_utils (https://github.com/galaxyproject/galaxy/pull/9953/files).

As an aside to an aside, I think galaxy_mock is too expansive. Anything that uses it would be better called an integration test than a unit test. It is hard to get away from global "god objects" in testing when your app depends on the same bad patterns - but there are things we could do better than galaxy_mock. For instance have the tests decorate different views of app it needs to make it clearer what interfaces the component under test is consuming.

jmchilton:
> Also maybe move test/unit/data/test_id_encode_decode.py to test/unit/data/security/test_id_encode_decode.py ?

Yeah, I meant to do that! Thanks.

nsoranzo:
> > Naive question: can the stuff in lib/galaxy/objectstore/test_helpers.py be merged into test/unit/unittest_utils/galaxy_mock.py ?
> 
> If you look at the imports in galaxy_mock.py - it needs stuff from all over the app - it needs to live at one of the highest packages of the package DAG. It shouldn't therefore be used to test stuff in galaxy-data and galaxy-objectstore. Part of the reason why I started ripping apart unittest_utils (https://github.com/galaxyproject/galaxy/pull/9953/files).

Thanks for the reply, it was indeed a naive question! I guess I'm not a fan of having the objectstore test helpers under `lib/galaxy/objectstore/` . Maybe this file and the various pieces of `unittest_utils` could become `test/unit/unittest_utils/objectstore_helpers.py` , `test/unit/unittest_utils/data_helpers.py` and so on, mirroring the packages they are needed for?

jmchilton:
> Maybe this file and the various pieces of unittest_utils could become test/unit/unittest_utils/objectstore_helpers.py , test/unit/unittest_utils/data_helpers.py and so on, mirroring the packages they are needed for?

Then we have to link in all the helpers together for all packages that use them - even if the dependencies are not meet (so the data package could access app's unit test helpers even though the dependencies aren't meant to be configured that way)? I really dislike that. I'd rather put in the extra work and properly package the unit tests at that point.

nsoranzo:
> Then we have to link in all the helpers together for all packages that use them - even if the dependencies are not meet (so the data package could access app's unit test helpers even though the dependencies aren't meant to be configured that way)? I really dislike that. I'd rather put in the extra work and properly package the unit tests at that point.

We don't have to symlink the whole `test/unit/unittest_utils/` directory into all packages though, right? For example for galaxy-objectstore we could symlink only `test/unit/unittest_utils/objectstore_helpers.py`, while for galaxy-data we could symlink `test/unit/unittest_utils/objectstore_helpers.py` and a future `test/unit/unittest_utils/data_helpers.py` and so on as needed.
Not saying it's great, but also not terrible I think.

jmchilton:
That is ... okay I guess. I dislike it but it would be a step forward from where we are now so I'll refactor this to do that. Once these things are all lined up and in place it will maybe be easier to articulate my concerns also.

nsoranzo (approved):

PR #9971, opened 2020-07-08
Fix workflow run default icon

guerler:


jmchilton:
Sorry about breaking that - thanks for the fix!

Issue #9968, opened 2020-07-08
issue with extended metadata_strategy and pulsar


gmauro:
I am testing the pr https://github.com/galaxyproject/galaxy/pull/9956 and my set up is:
* a pulsar destination with `<param id="metadata_strategy">directory</param>`enabled
* a backend with `store_by=uuid` enabled
* Galaxy release_20.05

I have uploaded a bam file two times, the first time with `metadata_strategy=extended' and then without.
Then I have executed the "samtools stats" tool on them against the pulsar destination.

The job using the dataset uploaded without the `metadata_strategy=extended' runs correctly, the other one fails with this error:
```
Jul 08 11:25:23 test.internal.usegalaxy.eu.novalocal python[873067]: galaxy.jobs.handler DEBUG 2020-07-08 11:25:23,014 Grabbed job(s): 787
Jul 08 11:25:23 test.internal.usegalaxy.eu.novalocal python[873067]: galaxy.jobs.mapper DEBUG 2020-07-08 11:25:23,065 (787) Mapped job to destination id: static_remote_cluster_mq_it01
Jul 08 11:25:23 test.internal.usegalaxy.eu.novalocal python[873067]: galaxy.jobs.handler DEBUG 2020-07-08 11:25:23,097 (787) Dispatching to pulsar_eu_it01 runner
Jul 08 11:25:23 test.internal.usegalaxy.eu.novalocal python[873067]: galaxy.jobs DEBUG 2020-07-08 11:25:23,109 (787) Persisting job destination (destination id: static_remote_cluster_mq_it01)
Jul 08 11:25:23 test.internal.usegalaxy.eu.novalocal python[873067]: galaxy.objectstore DEBUG 2020-07-08 11:25:23,125 Selected backend 'dnb03' for creation of Dataset 877
Jul 08 11:25:23 test.internal.usegalaxy.eu.novalocal python[873067]: galaxy.objectstore DEBUG 2020-07-08 11:25:23,160 Using preferred backend 'dnb03' for creation of Job 787
Jul 08 11:25:23 test.internal.usegalaxy.eu.novalocal python[873067]: galaxy.jobs DEBUG 2020-07-08 11:25:23,172 (787) Working directory for job is: /data/dnb03/galaxy_db/test/job_working_directory/000/787
[...]

PR #9976, opened 2020-07-09
add new test for pulsar, external-metadata and outputs_to_working_directory

bgruening:
Let's see if this fails.

xref: https://github.com/galaxyproject/galaxy/issues/9968


bgruening:
@mvdbeek did I do this wrong? Should this be in the other file? I guess we don't want to run another GHA job?

mvdbeek:
I don't this add additional test coverage. FWIW the upload tool always runs local, so I don't think that would uncover anything related to https://github.com/galaxyproject/galaxy/issues/9968. I think https://github.com/galaxyproject/galaxy/issues/9968#issuecomment-656149111 is a good lead.

mvdbeek:
I'm gonna close this for now, but I'll pick this and https://github.com/galaxyproject/galaxy/issues/9968#issuecomment-656149111 up after BCC.

PR #9986, opened 2020-07-12
Simplify workflow invocations grid, remove drop down, add workflow execution option

guerler:
This PR revises the workflow invocations grid. The grid currently has a series of options and dropdowns which might be misleading and overcrowd the interface. Some operation should not be made available like e.g. the workflow `deletion` and `renaming` options. These options do not return any confirmation of the action taken. Instead a user might assume that `deletion` deletes the invocation not the entire workflow.

Screen Shot 2020-07-24 at 1 23 23 PM

### DIFF
--- a/client/galaxy/scripts/components/Workflow/Invocations.vue
+++ b/client/galaxy/scripts/components/Workflow/Invocations.vue
@@ -6,11 +6,9 @@
         <b-alert variant="info" show v-if="headerMessage">
             {{ headerMessage }}
         </b-alert>
-        <b-alert v-if="loading" variant="info" show>
-            <loading-span message="Loading workflow invocation job data" />
-        </b-alert>
+        <loading-span v-if="loading" message="Loading workflow invocations" />
         <div v-else>
-            <b-alert v-if="!invocationItemsComputed.length" variant="secondary" show>
+            <b-alert v-if="!invocationItemsComputed.length" variant="info" show>
                 {{ noInvocationsMessage }}
             </b-alert>
             <b-table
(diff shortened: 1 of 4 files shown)


client/galaxy/scripts/components/Workflow/Invocations.vue
```diff
@@ -23,85 +21,63 @@
...
+                    </b-link>
                 </template>
                 <template v-slot:cell(history_id)="data">
-                    <div v-if="!ownerGrid || !getHistoryById(data.item.history_id)">
-                        {{ data.item.history_id }}
-                    </div>
-                    <div v-else>
-                        <history-dropdown :history="getHistoryById(data.item.history_id)" />
-                    </div>
+                    {{ getHistoryNameById(data.item.history_id) }}
```
jmchilton: At least have a way to navigate to the history?

guerler: I added links for the history column to allow switching between histories.

mvdbeek (commented):
This looks nicer, but can you restore the update time (I find that very useful for long-running workflows) and the workflow invocation id (this could maybe go to the expanded view with the jobs and steps)? I also liked the view and edit options, but I could see maybe not wanting it in this menu for simplicity.

client/galaxy/scripts/components/Workflow/Invocations.vue
```diff
@@ -23,85 +21,71 @@
...
+                        <b>{{ getHistoryNameById(data.item.history_id) }}</b>
+                    </b-link>
+                </template>
+                <template v-slot:cell(create_time)="data">
+                    <UtcDate :date="data.value" mode="elapsed" />
+                </template>
+                <template v-slot:cell(execute)="data">
+                    <b-button
+                        v-b-tooltip.hover.bottom
+                        title="Rerun Workflow"
```
mvdbeek:
```suggestion
                        title="Run Workflow"
```
We will want a re-run button in the future, but that will need some backend work to figure out the parameters.

guerler: Thank you for your suggestions. I relabeled the button to "Run", added the update time and the workflow invocation id.

mvdbeek:
Beautiful, I really like the position of the workflow invocation id!

mvdbeek (approved):

PR #9921, opened 2020-06-26
[20.01] Make cookie handling more robust

dannon:
Fixes #9916

PR #9991, opened 2020-07-13
[20.01] Cookie names to work with PR #9921 and py3 Exception printing

almahmoud:
`Nonce mismatch` error was occurring when logging in with OIDC (Keycloak). Noticed that the `nonce` cookie was returning `None`, and then Dannon informed me that `galaxy-` prefix is needed for cookie names after https://github.com/galaxyproject/galaxy/pull/9921.
Also, when the error was occurring, a subsequent error was happening when trying to print `e.message` because of changes to `Exception` in py3, so changed that as well.

This should also be ported to 20.05


dannon (approved):
Looks good, this should work with the new cookie handling.  Thanks!

PR #9995, opened 2020-07-14
Package rules DSL specification.

jmchilton:
It is used in many different places.


nsoranzo:
Would `test/unit/data/` be a good place for `rules_dsl_spec.yml` ? Trying to understand if it's useful outside of unit and Mocha testing.

jmchilton:
> Would test/unit/data/ be a good place for rules_dsl_spec.yml ?

I don't see that as an improvement over where it is now - but I'm happy to move it if that gets this done.

> Trying to understand if it's useful outside of unit and Mocha testing.

It is also used in the Python unit tests it looks like.

I assumed it was used by the Selenium tests also but that doesn't appear to be the case. If you'd just prefer me not package it and set it beside the unit test I suppose that could work also.

nsoranzo (approved):

nsoranzo:
Is the failed js_unit test legit?

jmchilton:
I've seen that one fail intermittently across PRs - I don't think it is related. I've restarted it and I'm confident it will turn green.

nsoranzo:
> I've seen that one fail intermittently across PRs - I don't think it is related. I've restarted it and I'm confident it will turn green.

OK, thanks, I did restart it once before asking and it failed again, so preferred to double check.

PR #9996, opened 2020-07-14
Add toolset

juleengraham:
Integrating toolset feature, similar to EU. Integrate the tools/view.

### DIFF
--- a/lib/galaxy/tools/toolbox/toolset.py
+++ b/lib/galaxy/tools/toolbox/toolset.py
@@ -6,7 +6,7 @@ import yaml
 
 class Toolset():
 
-    def __init__(self, config_filenames, tool_root_dir, app):
+    def __init__(self, app):
         self._toolsets = {}
         self._toolset_name = ""
         self._tools_by_id = {}
--- a/lib/galaxy/tools/toolbox/toolset.py
+++ b/lib/galaxy/tools/toolbox/toolset.py
@@ -1,0 +1,49 @@
+
+import os
+import string
+import yaml
+
+
+class Toolset():
+
+    def __init__(self, config_filenames, tool_root_dir, app):
+        self._toolsets = {}
+        self._toolset_name = ""
+        self._tools_by_id = {}
+
+    def toolset_list(self):
+        self.toolsets.keys()
+
+    def get_toolset(self, toolset_path):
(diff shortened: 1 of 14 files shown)


lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -3084,3 +3084,17 @@ mapping:
...
+
+      enable_toolset:
+        type: bool
+        default: false
+        required: false
+        desc: |
+          Allow the use of pre-determined toolsets in the toolbox and for the toolsview integration.
+          If it is enabled and set to true, please enable 'toolset_path' as well
+
+      toolset_path:
```
jmchilton, line 3096: Probably better to just enable this if toolset_path is set - no need for two parameters I assume?

lib/galaxy/tools/toolbox/toolset.py
```diff
@@ -0,0 +1,49 @@
...
+
+                tool_ids = self.get_toolset_ids(self._toolset_name)
+                print("TOOL_IDS: ", tool_ids)
+            else:
+                continue
+        print("NUMBER TOOLSETS: ", len(self._toolsets))
+
+    def get_toolset_ids(self, toolset_name):
+        toolset = self._toolsets.get(toolset_name)
+        return [tool['name'] for tool in toolset]
```
jmchilton, line 49: Is this name here somehow IDs?

jmchilton:
Amazing start - very exciting!

lib/galaxy/webapps/galaxy/config_schema.yml
juleengraham: good point!

lib/galaxy/tools/toolbox/toolset.py
juleengraham: I was using the format of EU's yaml files, and they have been using 'name' instead of id. Either is fine by me but we would have to choose one and stick with it so that the correct one is used in the toolset files and in the documentation.

https://github.com/usegalaxy-eu/usegalaxy-eu-tools

guerler:
Can you provide a screenshot?

mvdbeek:
I've pushed this to 21.01. I think this is going to be very cool!

jmchilton:
I think these ideas made it into #12327, going to close this for now. Let me know if I'm wrong though.

PR #9998, opened 2020-07-14
Optimize iRODS ``_exists()`` method

kxk302:
To improve the performance of irods calls

### DIFF
--- a/lib/galaxy/objectstore/irods.py
+++ b/lib/galaxy/objectstore/irods.py
@@ -454,33 +454,21 @@ class IRODSObjectStore(DiskObjectStore, CloudConfigMixin):
         return False
 
     def _exists(self, obj, **kwargs):
-        in_irods = False
         rel_path = self._construct_path(obj, **kwargs)
 
-        # Check cache
-        if self._in_cache(rel_path):
+        # Check cache and irods
+        if self._in_cache(rel_path) or self._data_object_exists(rel_path):
             return True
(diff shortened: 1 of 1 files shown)


nsoranzo (commented):
If that's correct, then you need to remove the `in_cache` variable from this method.

kxk302:
> If that's correct, then you need to remove the `in_cache` variable from this method.

That is correct. Please see the revised code I just pushed. Thx

lib/galaxy/objectstore/irods.py
```diff
@@ -454,20 +454,20 @@ def file_ready(self, obj, **kwargs):
         return False
 
     def _exists(self, obj, **kwargs):
-        in_cache = in_irods = False
+        in_irods = False
```
mvdbeek: You can simplify this method further. The github interface doesn't let me create a nice multi-line suggestion here, but this is what it boils down to:
```python
if self._in_cache(rel_path) or self._data_object_exists(rel_path):
    return True
```
At this point the only way we can still return `True` is if `dir_only ` and `base_dir` are both not "False-y".
So you can do
```python
if dir_only and base_dir:
    if not os.path.exists(rel_path):
        os.makedirs(rel_path)
    return True
return False
```

kxk302: Good idea. Revised code accordingly.

mvdbeek (approved):

PR #9999, opened 2020-07-15
API + UI for importing workflows from a GA4GH TRS server.

jmchilton:
Defaults to allowing import from Dockstore, which added Galaxy support with the latest release and has followed up with several bug fixes for us.

The UI and API are both roughly functional but there isn't a link to it yet in the analysis app. Once search is done that will make sense, but this is still something I think we want merged and ready to go for 20.09 because it provides a callback endpoint they can send us - (e.g. they should be able to link to https://usegalaxy.org/workflows/trs_import?trs_server=dockstore.org&trs_id=%23workflow%2Fgithub.com%2Fjmchilton%2Fgalaxy-workflow-dockstore-example-1%2Fmycoolworkflow and that seems to work in local testing). 

Check out the demo or an earlier version of this:

https://www.youtube.com/watch?v=D59ftjdme3o

### DIFF
--- a/client/src/components/Workflow/TrsServerSelection.vue
+++ b/client/src/components/Workflow/TrsServerSelection.vue
@@ -19,7 +19,7 @@
                     class="dropdown-item"
                     href="javascript:void(0)"
                     role="button"
-                    @click.prevent="onUrlSelection(selection)"
+                    @click.prevent="onTrsSelection(selection)"
                     >{{ selection.label }}</a
                 >
             </div>
--- a/client/src/components/Workflow/TrsServerSelection.vue
+++ b/client/src/components/Workflow/TrsServerSelection.vue
@@ -1,0 +1,68 @@
+<!-- Modelled after Sam's ToolShed ServerSelection.vue -->
+<template>
+    <span v-if="!loading" class="m-1 text-muted">
+        <span v-if="showDropdown" class="dropdown">
+            <b-link
+                id="dropdownTrsServer"
+                data-toggle="dropdown"
+                aria-haspopup="true"
+                aria-expanded="false"
+                class="font-weight-bold"
+            >
+                {{ selection.label }}
+                <span class="fa fa-caret-down" />
+            </b-link>
+            <div class="dropdown-menu" aria-labelledby="dropdownTrsServer">
+                <a
+                    v-for="selection in trsServers"
+                    :key="selection.id"
+                    class="dropdown-item"
+                    href="javascript:void(0)"
+                    role="button"
+                    @click.prevent="onUrlSelection(selection)"
+                    >{{ selection.label }}</a
+                >
+            </div>
+        </span>
+        <span v-else>
+            <b-link :href="selection.link_url" target="_blank" class="font-weight-bold" :title="selection.doc">
+                {{ selection.label }}
+            </b-link>
+        </span>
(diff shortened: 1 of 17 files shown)


lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -1155,6 +1155,19 @@ mapping:
...
             standard).
 
+      trs_servers_config_file:
+        type: str
+        default: 'trs_servers_conf.yml'
+        required: false
+        desc: |
+          Allow import of workflows from the TRS servers configured in
+          the specified YAML or JSON file. The file should be a list with
+          and 'id', 'label', and 'api_url' for each entry. Optionally,
```
bgruening:
```suggestion
          'id', 'label', and 'api_url' for each entry. Optionally,
```

bgruening (commented):
@jmchilton should we add here a trs_servers_conf.yml.sample file, maybe with Dockstore as documentation, as we do for other files as well.

jmchilton:
@bgruening done, thanks for the review!

client/src/components/Workflow/TrsServerSelection.vue
```diff
@@ -0,0 +1,68 @@
...
+            >
+                {{ selection.label }}
+                <span class="fa fa-caret-down" />
+            </b-link>
+            <div class="dropdown-menu" aria-labelledby="dropdownTrsServer">
+                <a
+                    v-for="selection in trsServers"
+                    :key="selection.id"
+                    class="dropdown-item"
+                    href="javascript:void(0)"
```
bgruening, line 20: I get that this was needed in old days, but is this still needed in Vue days?

bgruening (commented):
@jmchilton thanks! It works for me nicely. 

One question that was popping up is, should we check if the imported workflow is actually able to run on the imported instance? Warn the user and abort the importing optionally? This is not needed in this PR ... but what is the plan here? Also with respect to the search? Should the search show all workflows even when they are not useful on a given instance.

client/src/components/Workflow/TrsServerSelection.vue
jmchilton: I don't know for sure - I got it by copy and pasting code that was last modified with this commit (https://github.com/galaxyproject/galaxy/pull/8334) which seems fairly recent.

jmchilton:
> One question that was popping up is, should we check if the imported workflow is actually able to run on the imported instance?

I mean... yes of course we should. 

> Also with respect to the search? Should the search show all workflows even when they are not useful on a given instance.

Well this is the problem with standards right - we will probably never be able to get this information from the GA4GH API. But the rest of it - warning users before importing, providing a form for requesting tool installs, etc... should be done. Hopefully in a followup PR and hopefully in a way that benefits other forms of workflow import also. I noticed some other bugs with the existing functionality related to landing on the workflow list after importing a workflow - this needs to be more informative and more robust.

bgruening:
> warning users before importing, providing a form for requesting tool installs, etc... should be done.

Thats what I was talking about. If we can filter/warn on the Galaxy side.

Thanks John.

jmchilton:
Thanks for the detailed review and the merge @bgruening - I really appreciate it. I'll try to make some progress on your longer term concerns.

PR #10008, opened 2020-07-16
Don't flush after each dataset creation in DistributedObjectStore

mvdbeek:
I wasn't sure if this is safe in all instances (=followed up by a flush elsewhere), so I created dc3ca6d2698545e755eab66e2c95888019444794 ... and I think I have examined every place that calls *store.update_from_file and *store.create. The only danger I can see here is if for some reason we don't flush (an uncaught exception somewhere) after calling create / update_from_file. In that case the next access to such a dataset would trigger `__get_store_id_for `, which would be slower but also should be able to recover the object_store_id.

### DIFF
--- a/lib/galaxy/objectstore/__init__.py
+++ b/lib/galaxy/objectstore/__init__.py
@@ -15,10 +15,6 @@ import time
 from collections import OrderedDict
 
 import yaml
-try:
-    from sqlalchemy.orm import object_session
-except ImportError:
-    object_session = None
 
 from galaxy.exceptions import ObjectInvalid, ObjectNotFound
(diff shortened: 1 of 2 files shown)


lib/galaxy/objectstore/__init__.py
```diff
@@ -15,10 +15,6 @@
 from collections import OrderedDict
 
 import yaml
-try:
-    from sqlalchemy.orm import object_session
-except ImportError:
-    object_session = None
```
jmchilton, line 21: ❤️ what a beautiful commit from a package dependency and cohesion standpoint!

PR #10009, opened 2020-07-17
Properly update active version in workflow attribute form

guerler:
Thank you for catching this @mvdbeek


mvdbeek (approved):

mvdbeek:
Thanks for the quick fix @guerler!

PR #10010, opened 2020-07-17
Log when an exception is encountered while indexing a MAF.

davebx:
This is part of debugging and fixing a data table issue on test.galaxyproject


lib/galaxy/datatypes/util/maf_utilities.py
```diff
@@ -296,7 +296,8 @@ def maf_index_by_uid(maf_uid, index_location_file):
                     return bx.align.maf.MultiIndexed(maf_files, keep_open=True, parse_e_rows=False)
                 except Exception as e:
                     raise Exception('MAF UID (%s) found, but configuration appears to be malformed: %s' % (maf_uid, e))
-        except Exception:
+        except Exception as e:
+            log.exception(e)
```
natefoo, line 300: You don't need to pass the exception object to `log.exception()`, it's logged automatically.

Issue #6887, opened 2018-10-17
Things to do to make the job cache more useful


mvdbeek:
- [ ] Datasets imported from a library are not recognized as being the same
- [x] The input and output datasets need to be owned by the user -- accessible should be enough (though that might be  a concern with our accessible by default policy)
- [ ] #6781 having the HID as part of the name really requires one to have the exact same history structure
- [ ] A tool property that says "ignore the element_identifier / name when caching" would be helpful
- [ ] A tool property to indicate the tool outputs are not reproducible (e.g. tools that produce random results on purpose)
- [ ] An additional setting that would allow the user to specify that an exact metadata match isn't necessary
- [ ] Enable cache override via UI element in the tool form.
- [ ] A preview
- [ ] Make datasets read-only in object store https://github.com/galaxyproject/galaxy/issues/8263
- [ ] If we have a hash for dataset contents we can "resume" cached execution when intermediate files get deleted or when the dataset metadata for an input file didn't match but the metadata mismatch didn't result in a different output
- [x] Add something to the UI to indicate that a job was recovered from cached data
- [ ] https://github.com/galaxyproject/galaxy/issues/20050

hexylena:
ping @wm75 who had similar complaints when he tried it yesterday.

wm75:
Agree on all of the above. Plus, I'd add:
A more informative status message of what's happening. In my hands whenever the server was using a cached result, after clicking Execute the page appeared to be frozen sending the job.
It was only after several minutes that it switched to job submission confirmation. At the same time, the new datasets suddenly appeared in my history.
I can only speculate, but maybe that's because of the large size of Galaxy EU's database?

mvdbeek:
I don't think database size should have a major effect. How many datasets were you generating there ?

wm75:
per job? Between one and four.
Or are you asking how many datasets I own on the server (not sure about that, but maybe 2000)?

mvdbeek:
No, the number of items a tool execution would produce. So yeah, that is less than ideal ...

lecorguille:
If you shared a history for training purpose, perhaps, this feature can be useful, isn't it?
-> Not from library, same history structure

But maybe we hit this wall?
> The input and output datasets need to be owned by the user

since the datasets and the cached jobs are owned by the instructor?

bwlang:
Would be good to add something to the UI to indicate that a job was recovered from cached data.  Also, would be cool to enable cache override via UI element in tool form

[...]

PR #10020, opened 2020-07-21
Job cache improvements

### LINKED ISSUE #7975: Job Search/cache  not finding equivalent BWA, bowtie and other mappers jobs

mvdbeek:
Just putting this out now in case we'd want to hack on this.
With these changes I can run the entire SARS-CoV-2 variation workflow and have existing jobs be picked up while new jobs are being executed. Could use some tests (manual and code).

Fixes https://github.com/galaxyproject/galaxy/issues/7975.

If we wanna hack on this we could add support for dataset hashes for instance.
There's a bunch of other ideas here: https://github.com/galaxyproject/galaxy/issues/6887

### DIFF
--- a/lib/galaxy/job_execution/output_collect.py
+++ b/lib/galaxy/job_execution/output_collect.py
@@ -160,7 +160,6 @@ def collect_dynamic_outputs(
             log.exception("Problem gathering output collection.")
             collection.handle_population_failed("Problem building datasets for collection.")
 
-        has_collection.finalize(collection_builder._collection_type_description)
         job_context.add_dataset_collection(has_collection)
 
 
--- a/lib/galaxy/job_execution/output_collect.py
+++ b/lib/galaxy/job_execution/output_collect.py
@@ -160,6 +160,7 @@ def collect_dynamic_outputs(
             log.exception("Problem gathering output collection.")
             collection.handle_population_failed("Problem building datasets for collection.")
 
+        has_collection.finalize(collection_builder._collection_type_description)
         job_context.add_dataset_collection(has_collection)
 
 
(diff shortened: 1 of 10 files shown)


mvdbeek:
Taking this out of WIP. This is definitely an improvement over the current code and fixes some missing query filters that made the query results way too large, so would be good to get into 20.09.

PR #10024, opened 2020-07-22
GTN in Galaxy

hexylena:
I've implemented a webhook to overlay the GTN materials in Galaxy, and provide integration with Galaxy itself.

out

Materials will be accessible from the masthead

image

And upon activation, will be rendered in the main panel.

Selection_011

The blue buttons? Those don't show up in the GTN, they're hidden and only render when the GTN is accessed within Galaxy. Upon click they will close the popup and activate the tool (or upload interface) within the UI.

The magic buttons require the additional setup of a proxy route for the GTN as the iframe needs to be on the same domain as the Galaxy server to enable this to work:

```
location /training-material/ {
    proxy_pass https://training.galaxyproject.org/training-material/;
}
```

A small price for this feature I'd argue. Otherwise, it renders the GTN without buttons.

Whenever you want to visit the GTN again, clicking the button will show the popup at the same point you were in the tutorial when you left.


config/plugins/webhooks/gtn/script.js
```diff
@@ -0,0 +1,106 @@
+$(document).ready(function() {
+	var GTNView = Backbone.View.extend({
```
dannon: I need to refresh my memory as to what my actual plan was for this, but assuming a global `Backbone` (or even `jQuery`/`$`) won't be robust long term.  Nothing to fix here, right now, because this will work -- but we'll want to revisit it.

Ideally we have common build infrastructure with visualization bundling here that can convert these into fully encapsulated, runnable scripts.

hexylena: Easy for me to remove JQuery, and backbone is totally unnecessary here, happy to remove both and just write plain JS now if that's preferable?

dannon: Plain old JS is definitely preferable, and is more future-proof, thanks!

hexylena: Updated: Removed backbone, jquery is a bit more difficult, so maybe leaving that until there's a long term vue fix?

dannon: That's fine with me; as much as I want to get rid of it, we're still stuck with jquery for a while :)  (Backbone will be gone sooner, much easier to replace the related components)

bgruening:
WTF, this is awesome! Magic @hexylena!

Slugger70:
Bloody oath! This is an awesome idea. Grouse.

hexylena:
Just for the record: I think there are potentially better long term solutions where Galaxy embraces training development/creation/learning within Galaxy by merging training with workflows (á la workflow reporting) but that's a long term goal. In the short term however, this makes a real improvement for users and helps users get in touch with us + training much more easily.

I'm going to make an quick addition to test the url for the iframe: if it's loading properly /training-materials/, go with it. If not: then update the iframe for the real training materials url. They won't get the fancy "click button to run", but they'll still get the GTN directly in their Galaxy.

@shiltemann and I discussed adding a feature to run WFs from this view as well, but their usage *within* the tutorials is not very high (basically metagenomics/metatranscriptomics only) so we are explicitly not doing that currently.

hexylena:
Added a test, if /training-material/ is present, it will use that. If it isn't, then it falls back to an iframe for the public site, without the fancy click-to-run buttons, and a notice so admins can set that up if they want:

image

which links to documentation (wow, I actually bothered to write some.) Explicitly tested with /galaxy/ proxy prefix and it works OK.

I think this is potentially good to go?

dannon:
Without extra route configuration (but with the webhook enabled) I'm getting an undefined error when trying to activate it; debugging now.

dannon:
@hexylena Thanks!  This is great stuff.

hexylena:
Thanks so much @dannon. Looking forward to using this in my next training.

PR #10025, opened 2020-07-22
[20.05] Mulled packaging dependency fixes

mvdbeek:
So `pip install galaxy-tool-util[mulled]` will give you functional mulled commands. xref https://github.com/galaxyproject/galaxy-lib/issues/156

### DIFF
--- a/packages/tool_util/setup.py
+++ b/packages/tool_util/setup.py
@@ -97,7 +97,7 @@ setup(
     include_package_data=True,
     install_requires=requirements,
     extras_require={
-        'condatesting': ['jinja2'],
+        'mulled': ['conda', 'cytoolz', 'jinja2', 'Whoosh'],
     },
     license="AFL",
     zip_safe=False,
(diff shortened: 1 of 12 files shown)


bgruening:
It also uses the `conda` package, the API at the moment. I needed to add this as well. But we can fix this on the conda side.

nsoranzo:
@mvdbeek Can you please try to remove all new and old symlinks to individual test files in `packages/tool-util/tests/` (but keep `__init__.py`) and add a single symlink: `packages/tool_util/tests/tool_util` -> `../../../test/unit/tool_util/`

packages/tool_util/setup.py
```diff
@@ -97,7 +97,7 @@ def get_var(var_name):
     include_package_data=True,
     install_requires=requirements,
     extras_require={
-        'condatesting': ['jinja2'],
+        'mulled': ['conda', 'cytoolz', 'jinja2', 'Whoosh'],
```
nsoranzo: What's cytoolz for? I couldn't find a reference in the repo (on the phone).

nsoranzo (approved):

packages/tool_util/setup.py
mvdbeek: cytoolz is an undeclared dependency of the conda package on pypi.

Issue #9046, opened 2019-11-28
Feature request: AWS cost estimate


scholtalbers:
On usegalaxy.eu (https://github.com/usegalaxy-eu/galaxy/blob/release_19.09_europe/templates/show_params.mako#L227) there is a cost estimate for running the same job on AWS.
This might be a nice feature for any Galaxy instance to give the user(and admin) an idea of what their computation costs. 

Wishlist:
- total cost per job (instance, data transfer)
- total cost of a history (sum of all jobs + data storage)
- cost of storage usage

The instance estimate is what .eu is already providing, but it doesn't yet include any datastorage/transfer costs. See also https://gitter.im/usegalaxy-eu/Lobby?at=5ddf80f446397c721c4b1b85

nsoranzo:
Ping @bgruening

bgruening:
I know I talked to Jelle and then he created this issue :)

HadleyKing:
@nsoranzo, 
This is one of the things that @kee007ney and I are trying to work on with our project. We would be happy to try and help out with developing/testing this.

bgruening:
@HadleyKing let me know if you need more informations. @scholtalbers already linked some code above.

HadleyKing:
I think Step 1: enable this script (or a copy of it) to load when you select an invocation under the User -> [Workflow Invocations ](http://localhost:8080/workflows/invocations#)
Once we were able to see an example that could aggregate the requisite info, we could tweak it and test it for accuracy.

HadleyKing:
Or maybe it could be part of the [View](http://localhost:8080/workflow/display_by_id?id=f2db41e1fa331b3e) screen as well. Maybe under `About this Workflow`?

bgruening:
I will talk to @OlegZharkov ... maybe he has time to look into it. I guess some portable Vue components with a flexible config file to provide the pricing-table could be useful.

HadleyKing:
I would be happy to help... but I am a total noob. 
@jpat1546 and I want to implement this functionality into a [BCO extension](https://github.com/biocompute-objects/BCO_Specification/blob/1.4.0/docs/bco-domains.md#23-extension-domain-extension_domain).

HadleyKing:
[...]

PR #10029, opened 2020-07-24
Added functions to extract dictionary for AWS cost estimate

HadleyKing:
RE https://github.com/biocompute-objects/galaxy/issues/28 and https://github.com/galaxyproject/galaxy/issues/9046 and https://github.com/galaxyproject/galaxy/pull/9621

    GET /api/invocations/{invocations_id}/aws_estimate
    Return a dictionary with an AWS estimate metrics for a workflow invocation.
Changes to be committed:
	modified:   lib/galaxy/webapps/galaxy/api/workflows.py
	modified:   lib/galaxy/webapps/galaxy/buildapp.py


lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -1005,6 +1007,46 @@ def invocation_step(self, trans, invocation_id, step_id, **kwd):
...
+                        if hasattr(job_input.dataset, 'dataset_id'):
+                            ds_metrics[job_input.dataset.id].update({'filename': job_input.dataset.name})
+                            metrics[i].append(ds_metrics[job_input.dataset.id])
+                    for job_output in job.output_datasets:
+                        if hasattr(job_output.dataset, 'dataset_id'):
+                            ds_metrics[job_output.dataset.id].update({'filename': job_output.dataset.name})
+                            metrics[i].append(ds_metrics[job_output.dataset.id])
+
+        aws_dict = metrics
+        return aws_dict
```
bgruening: you can return here metrics and remove the aws_dict entirely

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -1005,6 +1007,46 @@ def invocation_step(self, trans, invocation_id, step_id, **kwd):
...
+                'uuid': str(h.dataset.uuid)
+            }
+        for i, step in enumerate(workflow_invocation.steps):
+            if step.workflow_step.type == 'tool':
+                for job in step.jobs:
+                    metrics[i] = summarize_job_metrics(trans, job)
+                    for job_input in job.input_datasets:
+                        if hasattr(job_input.dataset, 'dataset_id'):
+                            ds_metrics[job_input.dataset.id].update({'filename': job_input.dataset.name})
+                            metrics[i].append(ds_metrics[job_input.dataset.id])
```
bgruening: Can you not simply append `{'filename': job_input.dataset.name}` and remove ds_metrics?

bgruening (commented):
Maybe I'm not seeing it, but the mapping to AWS does not happen here, isn't it? This API is returning matrices about an entire workflow. So maybe the endpoints should be renamed?

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -1005,6 +1007,46 @@ def invocation_step(self, trans, invocation_id, step_id, **kwd):
...
+    def _generate_aws_estimate(self, trans, invocation_id, **kwd):
+        """
+        build AWS cloud estimate
+        """
+        decoded_workflow_invocation_id = self.decode_id(invocation_id)
+        workflow_invocation = self.workflow_manager.get_invocation(trans, decoded_workflow_invocation_id)
+        history = workflow_invocation.history
+
+        aws_dict, metrics, ds_metrics = {}, {}, {}
+        h_contents = self.history_contents_manager.contained(history)
```
mvdbeek: I would suggest not iterating over the history contents here, this may contain many more datasets that are not part of the workflow invocation. You can get size and other hda attributes from the job inputs and outputs

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -1005,6 +1007,46 @@ def invocation_step(self, trans, invocation_id, step_id, **kwd):
...
+        for h in h_contents:
+            ds_metrics[h.dataset.id] = {
+                'file_size': int(h.dataset.file_size),
+                'total_size': int(h.dataset.total_size),
```
mvdbeek: For these purposes I think only `total_size` matters.

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -1005,6 +1007,46 @@ def invocation_step(self, trans, invocation_id, step_id, **kwd):
...
+                'uuid': str(h.dataset.uuid)
+            }
+        for i, step in enumerate(workflow_invocation.steps):
+            if step.workflow_step.type == 'tool':
+                for job in step.jobs:
+                    metrics[i] = summarize_job_metrics(trans, job)
+                    for job_input in job.input_datasets:
+                        if hasattr(job_input.dataset, 'dataset_id'):
+                            ds_metrics[job_input.dataset.id].update({'filename': job_input.dataset.name})
```
mvdbeek: If using collections in workflows the dataset name is usually not informative. I'd suggest instead populating this with `'job_input_name': job_input.name`, which is the output parameter of the tool that corresponds to this dataset

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -1005,6 +1007,46 @@ def invocation_step(self, trans, invocation_id, step_id, **kwd):
...
+                            metrics[i].append(ds_metrics[job_input.dataset.id])
+                    for job_output in job.output_datasets:
+                        if hasattr(job_output.dataset, 'dataset_id'):
+                            ds_metrics[job_output.dataset.id].update({'filename': job_output.dataset.name})
```
mvdbeek:
```suggestion
                            ds_metrics[job_output.dataset.id].update({'job_output_name': job_output.name})
```

lib/galaxy/webapps/galaxy/api/workflows.py
mvdbeek: If you build this dictionary in the for loop below you'd also address @bgruening's comment

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -1005,6 +1007,46 @@ def invocation_step(self, trans, invocation_id, step_id, **kwd):
...
+        for h in h_contents:
+            ds_metrics[h.dataset.id] = {
+                'file_size': int(h.dataset.file_size),
+                'total_size': int(h.dataset.total_size),
+                'uuid': str(h.dataset.uuid)
+            }
+        for i, step in enumerate(workflow_invocation.steps):
+            if step.workflow_step.type == 'tool':
+                for job in step.jobs:
+                    metrics[i] = summarize_job_metrics(trans, job)
```
mvdbeek: This way you'll only collect the last jobs' metrics or a workflow step since you overwrite `metrics[i]` for each job.

I'd suggest making metrics a list of step metrics, where each metric is a dictionary describing the step and the metric. The structure could look something like
```python
[{'step_id': step.id,
  'step_label': step.label,
  jobs: [{'job_id': job.id, 'job_metrics': metrics, 'job_inputs': [], 'job_outputs': []]
}]
```

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -1005,6 +1007,46 @@ def invocation_step(self, trans, invocation_id, step_id, **kwd):
...
+                    for job_output in job.output_datasets:
+                        if hasattr(job_output.dataset, 'dataset_id'):
+                            ds_metrics[job_output.dataset.id].update({'filename': job_output.dataset.name})
+                            metrics[i].append(ds_metrics[job_output.dataset.id])
+
+        aws_dict = metrics
+        return aws_dict
+
+    @expose_api
+    def export_aws_estimate(self, trans, invocation_id, **kwd):
```
mvdbeek: Can you rename this to `invocation_metrics` or similar ? The output may not actually contain aws estimates if that is not configured.

mvdbeek (commented):
That's a nice idea, hope my suggestions are helpful.

HadleyKing:
> Maybe I'm not seeing it, but the mapping to AWS does not happen here, isn't it? This API is returning matrices about an entire workflow. So maybe the endpoints should be renamed?

No it does not. I was trying to keep it more generalized for use in other unforeseen circumstances but wanted to still keep the "provenance" of what inspired the functions. I think it makes the most sense for me to utilize these functions in the front end or another aspect but have them committed for general use. Thanks for the suggestions. I have generalized the functions and endpoints as you suggested.

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -1005,6 +1007,62 @@ def invocation_step(self, trans, invocation_id, step_id, **kwd):
...
+                        metrics[i]['runtime_seconds'] = value['raw_value']
+                    for job_input in job.input_datasets:
+                        if hasattr(job_input.dataset, 'dataset_id'):
+                            for h in [x for x in h_contents if x.dataset_id == job_input.dataset.id]:
+                                h_metrics = {
+                                    'total_size': int(h.dataset.total_size),
+                                    'uuid': str(h.dataset.uuid),
+                                    'job_input_name': job_input.name
+                                }
+                                metrics[i]['job_inputs'].append(h_metrics)
```
mvdbeek: You don't need the history contents at all here:
```suggestion
                        if job_input.dataset.dataset:
                            dataset = job_input.dataset
                            dataset_metrics = {
                                'total_size': int(dataset.total_size),
                                'id': trans.security.encode_id(dataset.id),
                                'model_class': dataset.__class__.__name__,
                                'job_input_name': job_input.name
                            }
                            metrics[i]['job_inputs'].append(dataset_metrics)
```
I'd use the item id instead of the uuid. The id is the primary identifier we use in Galaxy, with this id you can get more info about this dataset for instance ... we don't support querying by uuid for all api endpoints. The input might also be a LIbraryDatasetAssociation, so that's why I'd also dump the model_class into the metrics.

You can use the same technique for job outputs.

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -1005,6 +1007,62 @@ def invocation_step(self, trans, invocation_id, step_id, **kwd):
...
+        h_contents = self.history_contents_manager.contained(history)
+
+        for i, step in enumerate(workflow_invocation.steps):
+            if step.workflow_step.type == 'tool':
+                tool_id = step.workflow_step.tool_id.split('/')[-2:]
+                for job in step.jobs:
+                    job_metrics = summarize_job_metrics(trans, job)
+                    metrics[i] = {
+                        'tool_name': tool_id[0],
+                        'tool_version': tool_id[1],
```
mvdbeek:
```suggestion
                        'tool_version': job.tool_version,
```

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -1005,6 +1007,62 @@ def invocation_step(self, trans, invocation_id, step_id, **kwd):
...
+        metrics, h_metrics = {}, {}
+        h_contents = self.history_contents_manager.contained(history)
+
+        for i, step in enumerate(workflow_invocation.steps):
+            if step.workflow_step.type == 'tool':
+                tool_id = step.workflow_step.tool_id.split('/')[-2:]
+                for job in step.jobs:
+                    job_metrics = summarize_job_metrics(trans, job)
+                    metrics[i] = {
+                        'tool_name': tool_id[0],
```
mvdbeek:
```suggestion
                        'tool_id': job.tool_id,
```

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -1005,6 +1007,62 @@ def invocation_step(self, trans, invocation_id, step_id, **kwd):
...
+        decoded_workflow_invocation_id = self.decode_id(invocation_id)
+        workflow_invocation = self.workflow_manager.get_invocation(trans, decoded_workflow_invocation_id)
+        history = workflow_invocation.history
+
+        metrics, h_metrics = {}, {}
+        h_contents = self.history_contents_manager.contained(history)
+
+        for i, step in enumerate(workflow_invocation.steps):
+            if step.workflow_step.type == 'tool':
+                tool_id = step.workflow_step.tool_id.split('/')[-2:]
```
mvdbeek:
```suggestion
```

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -1005,6 +1007,62 @@ def invocation_step(self, trans, invocation_id, step_id, **kwd):
...
+        build a dictionary of invocation metrics
+        """
+        decoded_workflow_invocation_id = self.decode_id(invocation_id)
+        workflow_invocation = self.workflow_manager.get_invocation(trans, decoded_workflow_invocation_id)
+        history = workflow_invocation.history
+
+        metrics, h_metrics = {}, {}
+        h_contents = self.history_contents_manager.contained(history)
+
+        for i, step in enumerate(workflow_invocation.steps):
```
mvdbeek, line 1315: Would it make sense to make sure all steps are scheduled and have finished executing ?

HadleyKing: I guess it would. I have always thought about these things in a 'historical' sense so that did not occur to me. How would you do that? Set an error if the invocation has not finished? Or set it to wait? Not sure how to do that best.

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -1005,6 +1007,62 @@ def invocation_step(self, trans, invocation_id, step_id, **kwd):
...
+    def _generate_invocation_metrics(self, trans, invocation_id, **kwd):
+        """
+        build a dictionary of invocation metrics
+        """
+        decoded_workflow_invocation_id = self.decode_id(invocation_id)
+        workflow_invocation = self.workflow_manager.get_invocation(trans, decoded_workflow_invocation_id)
+        history = workflow_invocation.history
+
+        metrics, h_metrics = {}, {}
+        h_contents = self.history_contents_manager.contained(history)
```
mvdbeek:
```suggestion
```

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -1005,6 +1007,62 @@ def invocation_step(self, trans, invocation_id, step_id, **kwd):
...
+                            }
+                            metrics[i]['job_inputs'].append(dataset_metrics)
+                    for job_output in job.output_datasets:
+                        if hasattr(job_output.dataset, 'dataset_id'):
+                            for h in [x for x in h_contents if x.dataset_id == job_output.dataset.id]:
+                                h_metrics = {
+                                    'total_size': int(h.dataset.total_size),
+                                    'uuid': str(h.dataset.uuid),
+                                    'job_output_name': job_output.name
+                                }
```
mvdbeek:
```suggestion
                        if job_output.dataset.dataset:
                            dataset = job_output.dataset
                            h_metrics = {
                                'total_size': int(dataset.total_size),
                                'id': trans.security.encode_id(dataset.id),
                                'model_class': dataset.__class__.__name__,
                                'job_output_name': job_output.name
                            }
```

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -1005,6 +1007,62 @@ def invocation_step(self, trans, invocation_id, step_id, **kwd):
         )
         return self.__encode_invocation_step(trans, invocation_step)
 
+    def _generate_invocation_metrics(self, trans, invocation_id, **kwd):
+        """
+        build a dictionary of invocation metrics
+        """
+        decoded_workflow_invocation_id = self.decode_id(invocation_id)
+        workflow_invocation = self.workflow_manager.get_invocation(trans, decoded_workflow_invocation_id)
+        history = workflow_invocation.history
```
mvdbeek:
```suggestion
```

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -57,6 +58,7 @@ class WorkflowsAPIController(BaseAPIController, UsesStoredWorkflowMixin, UsesAnn
     def __init__(self, app):
         super(WorkflowsAPIController, self).__init__(app)
         self.history_manager = histories.HistoryManager(app)
+        self.history_contents_manager = history_contents.HistoryContentsManager(app)
```
mvdbeek:
```suggestion
```

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -21,9 +21,10 @@
 )
 from galaxy.managers import (
     histories,
+    history_contents,
```
mvdbeek:
```suggestion
```

mvdbeek (commented):
I think this is almost OK now, but are you sure you wouldn't want to walk along the invocation on the client ? I am not opposed to merging this (maybe labeling it beta), but I think there are some details here that are not captured in the metrics. subworkflow invocations are not taken into account currently, counting of initial workflow step input datasets (they are not technically produced by the invocation), discovered output collections (they don't have a JobToOutputDatasetAssociation, so this endpoint wouldn't count them).

We could implement this on the backend like you're doing here, but I think we have almost everything in place to let the client collect these things.

HadleyKing:
I am much less comfortable with client side coding, so I did it on the backend because that is what I knew how to use. I am happy to accomplish this any way that is best all around.

PR #10037, opened 2020-07-27
Fix reports webapp template resolution

dannon:
This is a quickfix to make reports functional again.  The other option is to resolve the galaxy root from this file?  Either way, this package is reaching out into an external 'templates' directory though.  Maybe shove all that into this webapp?

This broke with the refactoring in https://github.com/galaxyproject/galaxy/pull/9959


nsoranzo:
I guess the `static` and `templates` symlinks in `lib/galaxy/web/framework/` can be removed?

jmchilton (approved):

dannon:
@nsoranzo I'll try it out, but yes, I'd guess those can likely go.  I hope nothing is using the /static mapping anymore, but I'll grep around and click through the apps.

lib/galaxy/config/sample/reports.yml.sample
```diff
@@ -52,7 +52,7 @@ uwsgi:
 
   # Mount the web application (e.g. Galaxy, Reports, etc.) at the given
   # URL prefix. Cannot be used together with 'module:' above.
-  #mount: /galaxy=galaxy.webapps.galaxy.buildapp:uwsgi_app()
+  #mount: /galaxy=galaxy.webapps.reports.buildapp:uwsgi_app()
```
dannon, line 55: Ran into this one when setting up reports at a prefix; the example here (unlike the regular module line) is incorrect for reports. I just realized, though, that it will get overwritten by a make config-rebuild, but I don't immediately see where I can contextualize this reports default differently from the galaxy config -- anyone know off the top of their head?

mvdbeek: Took the liberty to just push that right to your branch in https://github.com/galaxyproject/galaxy/pull/10037/commits/9487f5f541a5255dbe2075b787de4db5936845ea, hope that's ok.

mvdbeek (approved):

lib/galaxy/config/sample/reports.yml.sample
dannon: @mvdbeek Always totally fine, thanks!  I forgot to come back to this one.

PR #10042, opened 2020-07-29
[20.05] fix OSError in fastq_to_fqtoc converter

bgruening:
This fixes an `OSError: telling position disabled by next() call` error in our fastq_to_fqtoc converter.


dannon:
For more context, we did some extra debugging and it turns out you'd only see this error under python3 when the input fastq was more than 4000000 lines long.  Swapping from the `for line in` iterator allows tell() to work correctly.

PR #10044, opened 2020-07-30
Add GALAXY_ALLOW_FUTURE_DATABASE option in migration check

dannon:
Not set by default, simply a convenience for developers.  Also intentionally not adding this as a formal config option because it should probably only ever be used by developers frequently switching branches.


jmchilton:
Great!

dannon:
I don't suppose we want to call this a bug and apply it to 20.05, do we?  (having just bumped into it yet again... :))

PR #10045, opened 2020-07-30
Improve workflow layout algorithm

hexylena:
I wrote [a rubbish workflow plotter](https://github.com/hexylena/galaxy-workflow-plotter) forever ago and someone asked about it today. I was annoyed I never added a better layout algorithm and so I went looking today for a better algorithm while my blast runs.

I found this: https://github.com/kieler/elkjs 

> [ELK] is a layer-based layout algorithm that is particularly suited for node-link diagrams with an inherent direction and ports (explicit attachment points on a node's border). It is based on the ideas originally introduced by Sugiyama et al.

So I tried my hand at replacing Galaxy's layout with it. What do y'all think? Personally I like it a lot better.

**Default**
image

**ELK**
image

**Default**
image

detail image (:musical_note: Where did you come from, where did you go? :notes:)

image

**ELK**

image

Some admittedly imperfect things there, but, I'd argue an improvement. No longer noodle soup :ramen:!

For reference, the manual layout i'd been using, which only looks good on that version of galaxy due to the very different noodle bezier parameters.

image

- ~~Current bug: noodles bug out until you manually touch nodes. Dunno how to fix that sorry.~~ JK, old variable. Thanks CircleCI for catching it.
- This (somewhat intentionally) throws out the retaining of any user control over layout that the original considered.

### DIFF
--- a/client/galaxy/scripts/components/Workflow/Editor/Index.vue
+++ b/client/galaxy/scripts/components/Workflow/Editor/Index.vue
@@ -273,11 +273,9 @@ export default {
         onLayout() {
             this.canvasManager.drawOverview();
             this.canvasManager.scrollToNodes();
-            return import(/* webpackChunkName: "workflowLayout" */ "components/Workflow/Editor/modules/layout.js").then(
-                (layout) => {
-                    layout.default(this);
-                }
-            );
+            return import(/* webpackChunkName: "workflowLayout" */ "./modules/layout.js").then((layout) => {
+                layout.autoLayout(this);
+            });
         },
         onAttributes() {
             showAttributes();
--- a/client/galaxy/scripts/components/Workflow/Editor/Index.vue
+++ b/client/galaxy/scripts/components/Workflow/Editor/Index.vue
@@ -106,7 +106,6 @@ import {
     showForm,
     saveAs,
 } from "./modules/utilities";
-import { autoLayout } from "./modules/layout";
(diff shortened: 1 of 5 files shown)


dannon:
Ahh.  My one concern is the size of the library.  Let me see what I can do here to make it load async?

hexylena:
Sure, that makes sense.

dannon:
Should be a deferred load on demand now, thanks @hexylena.

hexylena:
The layout can probably be improved a bit by knowledge of where the noodle attaches, for what it's worth. [Example from their docs](https://rtsys.informatik.uni-kiel.de/elklive/examples.html?e=ports%2FportConstraints). Not sure how to translate that to JS yet,

Added two other options I found in one of their demos and that helps a bit too:

Before | After 6974ede8d6
--- | --- 
image | image

hexylena:
After 3d9e0555d9, I've added connections via 'ports' which have some specific positioning on the node (east/west/etc). This can probably be further refined but honestly it looks good enough to me now, using the most important features of the library for layout.

Before | After 
--- | ---
image | image

I'm quite happy with this, so, ready from my side :shipit:

dannon:
@hexylena This is great; the layout is much, jmuch better.  Thanks again.

hexylena:
Cheers @dannon! Appreciate the review once again. I look forward to using it!

guerler:
So much better! Thank you @hexylena

PR #10048, opened 2020-07-30
[20.01] Fix running workflows with parameters restricted on connections...

jmchilton:
... when they are copied from a single tool source.

This is a runtime problem so these workflows were probably very broken when they tried to execute but nothing in the database or representation should need to change - making the fix somewhat clean at least.

### DIFF
--- a/lib/galaxy/workflow/modules.py
+++ b/lib/galaxy/workflow/modules.py
@@ -992,7 +992,7 @@ class InputParameterModule(WorkflowModule):
                 options = None
                 if static_options and len(static_options) == 1:
                     # If we are connected to a single option, just use it as is so order is preserved cleanly and such.
-                    options = [{"value": o[0], "label": o[1]} for o in static_options[0]]
+                    options = [{"value": o[1], "label": o[0]} for o in static_options[0]]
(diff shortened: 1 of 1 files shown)


lib/galaxy/workflow/modules.py
```diff
@@ -992,7 +992,7 @@ def callback(input, prefixed_name, context, **kwargs):
                 options = None
                 if static_options and len(static_options) == 1:
                     # If we are connected to a single option, just use it as is so order is preserved cleanly and such.
-                    options = [{"value": o[0], "label": o[1]} for o in static_options[0]]
+                    options = [{"value": o[1], "label": o[0]} for o in static_options[0]]
```
dannon:
```suggestion
                    options = [{"label": o[0], "value": o[1]} for o in static_options[0]]
```

dannon: (just for my sanity and consistency in order of indices, also, seems like maybe the reason this was bugged in the first place ;) )

jmchilton: sounds good, thanks

PR #10056, opened 2020-08-03
Fix copy/paste in workflow editor

dannon:
`window.getSelection().toString() === ""` is not reliably the text selection contents across all browsers; using an active element heuristic instead seems to be more robust.

(if this seems fine and works for folks, I'll backport it flattened to 20.05)


mvdbeek:
Nice, that does look cleaner!

Issue #10030, opened 2020-07-24
duplicate entries in tool search result set


martenson:
This is today on .org, tried a couple of terms, every result hit seems duplicated 5 or more times

![Galaxy](https://user-images.githubusercontent.com/1814954/88412061-4526f600-cdd9-11ea-8fec-6e4a0985daab.png)

martenson:
This substantially limits number of results and overall usefulness of the tool search.

bgruening:
On EU we see 4 requests. I guess this correlates with the number of web workers?

martenson:
@bgruening seems unclear since in the example above different hits have different number of duplicates

bgruening:
Can you scroll down. In my case it is not sorted.

martenson:
```
[
    "toolshed.g2.bx.psu.edu/repos/bgruening/sklearn_sample_generator/sklearn_sample_generator/1.0.8.2",
    "toolshed.g2.bx.psu.edu/repos/bgruening/sklearn_sample_generator/sklearn_sample_generator/1.0.8.2",
    "toolshed.g2.bx.psu.edu/repos/bgruening/sklearn_sample_generator/sklearn_sample_generator/1.0.8.2",
    "toolshed.g2.bx.psu.edu/repos/bgruening/sklearn_sample_generator/sklearn_sample_generator/1.0.8.2",
    "toolshed.g2.bx.psu.edu/repos/bgruening/sklearn_sample_generator/sklearn_sample_generator/1.0.8.2",
    "toolshed.g2.bx.psu.edu/repos/iuc/bedtools/bedtools_randombed/2.26.0.0",
    "toolshed.g2.bx.psu.edu/repos/iuc/bedtools/bedtools_randombed/2.26.0.0",
    "toolshed.g2.bx.psu.edu/repos/iuc/bedtools/bedtools_randombed/2.26.0.0",
    "toolshed.g2.bx.psu.edu/repos/iuc/bedtools/bedtools_randombed/2.26.0.0",
    "toolshed.g2.bx.psu.edu/repos/iuc/bedtools/bedtools_randombed/2.26.0.0",
    "toolshed.g2.bx.psu.edu/repos/iuc/bedtools/bedtools_randombed/2.29.0",
    "toolshed.g2.bx.psu.edu/repos/iuc/bedtools/bedtools_randombed/2.29.0",
    "toolshed.g2.bx.psu.edu/repos/iuc/bedtools/bedtools_randombed/2.29.0",
[...]

PR #10061, opened 2020-08-04
[20.05] Fix non-unique entries in tool search index

mvdbeek:
Fixes https://github.com/galaxyproject/galaxy/issues/10030, broken in 19c4ff42823638d51792a98486cb28e9b92a6923, where I made the assumption that we always start with a fully-populated index.

This resulted in somewhere between 1 and N (for N web handlers) entries in the search index.
Fixed this by using `update_document` and making ID unique in the Whoosh schema.

`update_document` behaves like `add_document` if the document id is not in the index:
https://whoosh.readthedocs.io/en/latest/indexing.html?highlight=unique#updating-documents

### DIFF
--- a/lib/galaxy/tools/search/__init__.py
+++ b/lib/galaxy/tools/search/__init__.py
@@ -50,7 +50,7 @@ class ToolBoxSearch(object):
     """
 
     def __init__(self, toolbox, index_dir=None, index_help=True):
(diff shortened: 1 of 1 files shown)


bgruening:
Uh, nice find!

bgruening (approved):

dannon:
Yep, this looks good to me, thanks @mvdbeek

PR #10068, opened 2020-08-05
[20.05] Don't display download link for optional, non-existant metadata files

mvdbeek:
This became an issue when we added optional MetadaFile elements (the csi index) to Bam files in 20.05.

### DIFF
--- a/lib/galaxy/managers/datasets.py
+++ b/lib/galaxy/managers/datasets.py
@@ -523,12 +523,13 @@ class _UnflattenedMetadataDatasetAssociationSerializer(base.ModelSerializer,
         """
         meta_files = []
         for meta_type in dataset_assoc.metadata_file_types:
-            meta_files.append(
-                dict(file_type=meta_type,
-                     download_url=self.url_for('history_contents_metadata_file',
-                                               history_id=self.app.security.encode_id(dataset_assoc.history_id),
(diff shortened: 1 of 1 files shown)


mvdbeek:
@galaxybot test this

dannon:
This does fix the download link.  I think we might still have a problem with https://github.com/galaxyproject/galaxy/pull/9570 not generating .bais when downstream tools are expecting it, though

mvdbeek:
That's true, but that only happens for bam files for which we couldn't possibly generate .bai files. Not sure if there is a good solution to it ?

dannon:
@mvdbeek What would happen before csi index was available?  Would the metadata setting simply fail, failing the job?  @wm75 hit this on eu this week and it was not obvious what the issue was, other than downstream jobs would fail.

dannon (approved):

mvdbeek:
> What would happen before csi index was available?

Yeah, I think the metadata job would fail.

> @wm75 hit this on eu this week and it was not obvious what the issue was, other than downstream jobs would fail.

I think that might have been fixed by https://github.com/galaxyproject/galaxy/pull/10059, which fixes `metadata_source` in output data tags. @wm75 was that the bedtools intersect issue ?

PR #10070, opened 2020-08-05
Allow numbers as filenames while adding datasets to library folder

OlegZharkov:
If the filename of dataset is a number, it will raise an exception. This PR allows filenames to be numbers

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/library_datasets.py
+++ b/lib/galaxy/webapps/galaxy/api/library_datasets.py
@@ -396,9 +396,10 @@ class LibraryDatasetsController(BaseAPIController, UsesVisualizationMixin, Libra
         path = kwd.get('path', None)
         if path is None:
             raise exceptions.RequestParameterMissingException('The required attribute path is missing.')
-        folder = self.folder_manager.get(trans, folder_id)
         if isinstance(path, int):
-            path = str(path)
+            raise exceptions.RequestParameterInvalidException('The required attribute path is Integer.')
+
+        folder = self.folder_manager.get(trans, folder_id)
 
         source = kwd.get('source', None)
         if source not in ['userdir_file', 'userdir_folder', 'importdir_file', 'importdir_folder', 'admin_path']:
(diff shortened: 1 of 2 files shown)


lib/galaxy/webapps/galaxy/api/library_datasets.py
```diff
@@ -402,6 +402,8 @@ def load(self, trans, payload=None, **kwd):
         if path is None:
             raise exceptions.RequestParameterMissingException('The required attribute path is missing.')
         folder = self.folder_manager.get(trans, folder_id)
+        if isinstance(path, int):
+            path = str(path)
```
mvdbeek: I don't know, I think this should be a string when passed in from the client, and I wonder if this doesn't hide something else going wrong there.

lib/galaxy/webapps/galaxy/api/library_datasets.py
```diff
@@ -401,8 +401,8 @@ def load(self, trans, payload=None, **kwd):
         path = kwd.get('path', None)
         if path is None:
             raise exceptions.RequestParameterMissingException('The required attribute path is missing.')
-        if isinstance(path, int):
-            raise exceptions.RequestParameterInvalidException('The required attribute path is Integer.')
+        if not isinstance(path, str):
+            raise exceptions.RequestParameterInvalidException('The required attribute path is not String.')
```
mvdbeek, line 400:
```suggestion
            raise exceptions.RequestParameterInvalidException('The required attribute path is not a string.')
```

OlegZharkov:
@mvdbeek 
Does it look better now? It will be always String now, Python won't cast it as Integer.
I just added a check on backend: "raise an exception if path is not a String"

mvdbeek (approved):

mvdbeek:
Yep, thanks for digging!

OlegZharkov:
> Yep, thanks for digging!

a sure thing, thanks a lot for reviewing it!

lib/galaxy/webapps/galaxy/api/library_datasets.py
dannon: @OlegZharkov Looks like the article got dropped after the rebase?

bgruening:
Thanks @OlegZharkov!

PR #10072, opened 2020-08-06
Fix library zip download

dannon:
This zipfile creation code all needs refactoring, WIP

This fixes zip archive downloads in the library @OlegZharkov

### DIFF
--- a/lib/galaxy/util/streamball.py
+++ b/lib/galaxy/util/streamball.py
@@ -48,12 +48,13 @@ class ZipBall(object):
     def __init__(self, tmpf, tmpd):
         self._tmpf = tmpf
         self._tmpd = tmpd
+        self.wsgi_status = None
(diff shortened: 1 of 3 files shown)


bgruening:
@dannon we should either get this under testing or maybe drop a few of the archive options. Our users usually do not know the difference between all of them and maybe we should make a decision and only offer one? maybe two options with more recent compression formats that we also test.

dannon:
@bgruening Yep, I agree; right now it's handled differently in a couple places and could use centralization and testing in a refactoring pass.

OlegZharkov:
Thanks a lot @dannon! that was extremely fast response.

I probably should write a selenium test for this, so it doesn't suddenly break again.

bgruening (approved):

bgruening:
@OlegZharkov I take your word for some tests :)

Issue #10075, opened 2020-08-06
Custos Issues


juleengraham:
- [x] **Remember this IdP’ checkbox:** Add an option on the Galaxy login page to remember which IdP the user has selected
- [ ] **Existing account error handling:** if a user tries to link an external identity to an account that already exists in galaxy, the server currently just throws an exception. Need to handle this suitably. To recreate, create a username/pwd account with gmail, then logout, and try to log in using Custos with Google as the IdP provider. (https://sentry.galaxyproject.org/sentry/test/issues/844498/ -- `There already exists a user this email.  To associate this external login, you must first be logged in as that existing account.`)
- [x] **User lookup on login:** Show confirmation page before actually creating the user. Ask the user if they already have an account and instead need to link it, or create a new account and agree to the service terms. (when a user logs in using an external IdP, Galaxy should check if that username already exists and offer the user to link their existing account with the external identity. COmanage Match may be something to consider leveraging here.)
- [x] Make sure **email verification** is triggered if a new user is created via Custos
- [x] **Link multiple external accounts:** a user should be able to link multiple external IpD accounts in Galaxy on their profile
- [x] Accessing http://localhost:8080/user/external_ids access denied
- [x] After typing username and pwd, hitting Enter should trigger button press
- [ ] On logout, need to clear custos-state (ending up in wrong galaxy account via most recently associated custos id)

dannon:
@ahmedhamidawan Linking you in on this; it might be worth taking a look at the issues outlined here for context.

PR #10078, opened 2020-08-07
Allow option to remember the previously used IDP for login

juleengraham:
- if user checks the "remember instutition" checkbox on login with custos/cilogon, that idp is pre-selected the next time they log in (until the cookie expires)


juleengraham:
#10075

PR #10083, opened 2020-08-10
Progressively render conditional as they are being displayed

### LINKED ISSUE #9615: Circos tool UI performance
Re-running Circos jobs is incredibly slow (~30 seconds, locally, client-production, from clicking re-run to the tool being shown) 

I'm responding to reviewers currently, and one asked for lessons learned and limitations. There are a number, but one of the real user experience problems is how slow it is to load the interface. The jobs actually runs more quickly than pressing re-run.

## The Tool

Here is the xml of the version I'm testing with all macros expanded. (What is in IUC currently, but maybe this is easier for people)

[circos.xml](https://github.com/galaxyproject/galaxy/files/4480120/circos.xml.txt)

The parameter set for this tool is quite large, which is unavoidable:

image

So

## Some profile screenshots 
image

or in chrome

image

## Profiles

Below I've attached the profiles of clicking the re-run button for a single job. Is there any development time in the near future that will be allocated for such performance issues with large tools? Should we add a spinner in the main panel in the short term?

There are no network requests between build for rerun, and then images from help loading

image

Profiles:

[firefox-circos-rerun-profile.json.txt](https://github.com/galaxyproject/galaxy/files/4480040/firefox-circos-rerun-profile.json.txt)
[chrome rerun profile.json](https://usegalaxy.eu/static/vgcn/chrome-rerun-profile.json.txt)
[chrome rerun profile empty history.json](https://usegalaxy.eu/static/vgcn/chrome-profile-empty-history.json.txt)

## Reproducing

You can mostly reproduce this issue with an older version of the tool than the one I've included above with [this history on EU](https://usegalaxy.eu/u/helena-rasche/h/circos-microbe-tutorial) That takes ~10-15 seconds to load, not as bad.

And [here is a history export](https://usegalaxy.eu/static/vgcn/Galaxy-History-Circos-testing-for-export.tar.gz) from my current history, with the latest "Circos Plot" being the one in question. 

## Solutions?

Short term maybe a spinner in the main area? I know I've hit re-run again when I thought it wasn't doing anything.

mvdbeek:
This reduces the time to render https://github.com/hepcat72/tools-iuc/tree/xml2json/tools/ncbi_entrez_eutils from ~ 90 seconds to immediate. I think this approach should also work for other elements that hide things, like repeats and sections. Might also fix https://github.com/galaxyproject/galaxy/issues/9615 ?

### DIFF
--- a/client/galaxy/scripts/mvc/form/form-section.js
+++ b/client/galaxy/scripts/mvc/form/form-section.js
@@ -15,7 +15,11 @@ var View = Backbone.View.extend({
         this.inputs = options.inputs;
         this.parameters = new Parameters();
         this.setElement($("<div/>"));
-        this.render();
+        this.rendered = false;
+        this.elements = {};
+        if (options.skip_render !== true) {
+            this.renderOnce();
+        }
     },
 
     /** Render section view */
@@ -25,6 +29,14 @@ var View = Backbone.View.extend({
         _.each(this.inputs, (input) => {
             self.add(input);
         });
+    },
+
+    /** Only call render function if view has not been rendered before */
+    renderOnce: function () {
+        if (this.rendered === false) {
+            this.render();
+            this.rendered = true;
+        }
     },
 
     /** Add a new input element */
@@ -60,7 +72,9 @@ var View = Backbone.View.extend({
             for (var i in input_def.cases) {
                 var sub_section = new View(this.app, {
                     inputs: input_def.cases[i].inputs,
+                    skip_render: true,
                 });
+                this.elements[input_def.id + "_" + i] = sub_section;
                 this._append(sub_section.$el.addClass("ui-portlet-section pl-2"), `${input_def.id}-section-${i}`);
             }
             field.model.set("onchange", (value) => {
@@ -76,6 +90,8 @@ var View = Backbone.View.extend({
                         }
                     }
                     if (i == selectedCase && nonhidden) {
+                        const selectedView = self.elements[input_def.id + "_" + i];
+                        selectedView.renderOnce();
                         section_row.fadeIn("fast");
                     } else {
                         section_row.hide();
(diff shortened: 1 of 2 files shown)


mvdbeek:
https://github.com/galaxyproject/galaxy/pull/10083/commits/b582642dd6e7f78c147aadbaeccb21a28ce00e5d reduces the tim e to render to about 200ms.

client/galaxy/scripts/mvc/form/form-section.js
```diff
@@ -27,6 +31,14 @@ var View = Backbone.View.extend({
         });
     },
 
+    /** Only call render function if view has not been rendered before */
```
dannon, line 34: Do we ever need to rerender based on changes in the model?

mvdbeek: Maybe, but then the model changes should just trigger `render` ... which is my unverified thought / intuition. Do you have an idea how to check this ?

bgruening:
Impressive! I think what should be tested are more dynamic tools using data_column and still code tags (ping @hrhotz).

mvdbeek:
I can put this on test.galaxyproject.org when the tests have passed (unless @natefoo says no 😆 )

client/galaxy/scripts/mvc/form/form-section.js
dannon: @mvdbeek Not off the top of my head; I haven't looked at the tool form logic here in a long time.  This was just the only thing I thought of that might be a concern.  @guerler would probably know if it's something to worry about; hopefully it's not.

mvdbeek:
Hmm, that one selenium test with workflow integers failed (https://jenkins.galaxyproject.org/job/docker-selenium-py3/5156/artifact/5156-test-errors/) -- but I think it works locally. I'll try again.

mvdbeek:
Ah, there's just the `n` in the step `annotation` missing ... I don't think that test failure is related.
last
```
  File "/galaxy/lib/galaxy_test/selenium/framework.py", line 153, in func_wrapper
    return f(self, *args, **kwds)
  File "/galaxy/lib/galaxy_test/selenium/test_workflow_editor.py", line 129, in test_integer_input
    assert annotation == "my cool annotation"
AssertionError
```

hrhotz:
I will try to find time to look into this...

client/galaxy/scripts/mvc/form/form-section.js
dannon: @mvdbeek I spent some time testing this, and I'm not worried about it.  Nice improvement here.

dannon (approved):

PR #10084, opened 2020-08-10
Selenium: Library Dataset Download Zip

OlegZharkov:
This PR introduces new Selenium feature. It defines default download directory in gecko and chrome driver before we run selenium test.
 It allows us to check downloading dataset from Library-Folder (zip format). This has been recently fixed in https://github.com/galaxyproject/galaxy/pull/10072 and hopefully implemented test prevents us to break this functionality again. 

The idea is quite straightforward, we download selected dataset and verify its presence in download dir (`/tmp` by default)

This Test is already compatible with new major library folder re-write in https://github.com/galaxyproject/galaxy/pull/9905 and probably should be merged before 9905.

### DIFF
--- a/client/src/mvc/library/library-folderrow-view.js
+++ b/client/src/mvc/library/library-folderrow-view.js
@@ -267,13 +267,13 @@ var FolderRowView = Backbone.View.extend({
                         <% if( content_item.get("description").length > 40 ) { %>
                             <td data-toggle="tooltip" data-placement="auto"
                                 title='<%= _.escape(content_item.get("description")) %>'>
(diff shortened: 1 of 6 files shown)


OlegZharkov:
As we seen in https://github.com/galaxyproject/galaxy/pull/10090 Jenkins test, ".desciption-field" is a bad selector for selecting dataset and it might fail. Changed directly to checkbox. Should pass everything now

bgruening (approved):
Cool! More tests!!!!! :)

OlegZharkov:
@bgruening thanks a lot!

PR #10087, opened 2020-08-10
Fix change_datatype PJA when dataset is not yet in object store

### LINKED ISSUE #10085: Failed Invocation of workflow
Hello,

I am trying to invoke a workflow on usegalaxyEU server and it reports that Invocation scheduling failed. Half of the tools involved in the workflow runs and the rest doesn't queue.

Screen Shot 2020-08-10 at 12 50 02 PM

here is the workflow :https://usegalaxy.eu/u/subina/w/metatranscriptomics-full-workflow---gtn-imported-from-uploaded-file-2
History: https://usegalaxy.eu/u/subina/h/imported-t3ametatranscriptomicsworkflow .

It will be great if we can troubleshoot this issue!

mvdbeek:
Fixes a fallout from the recent flush optimizations. Should fix #10085 (@bgruening any chance you could try this ?).
- [x] Maybe we should push the is_new check into has_data?

### DIFF
--- a/lib/galaxy/datatypes/registry.py
+++ b/lib/galaxy/datatypes/registry.py
@@ -587,7 +587,7 @@ class Registry:
         # initialization.
         if data.has_data():
             data.set_size()
-            data.init_meta(copy_from=data)
+        data.init_meta(copy_from=data)
         return data
 
     def load_datatype_converters(self, toolbox, installed_repository_dict=None, deactivate=False, use_cached=False):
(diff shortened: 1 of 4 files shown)


bgruening:
We can deploy this yes, can you give me a go when you think it is ready?

mvdbeek:
I'm just moving one of the API tests over to integration tests -- I think we have a test that would have failed if the testing framework was configured to use `store_by: id`.

mvdbeek:
The test fails with the same stacktrace (without 79e735e38f9b0808a612ffdcd124631ac2c4800b), so I don't think we need to test this separately @bgruening.

bgruening:
Cool, thanks a lot!

PR #10089, opened 2020-08-11
Count all terminal jobs when rendering job complete string

mvdbeek:
This previously only counted OK states, which was very confusing if you're waiting for your jobs to run, but they've actually errored or were deleted. Maybe paused should also count as a terminal state, or maybe the progress bar should be colored based on the number of OK / RUNNING / PAUSED / ERROR / DELETED ?

I am planning to make this a dropdown with jobs per state in the future (to mirror what I've done in https://github.com/galaxyproject/planemo/pull/1048), but that's a nice first step that I think we should backport eventually.


dannon (approved):

mvdbeek:
I'll probably rework this quite a bit in the next release cycle, but this should be fine.

PR #10090, opened 2020-08-11
Selenium Delete Dataset from library folder

OlegZharkov:
This PR implements deleting dataset from library folder. Quite straight forward.

It is another step before https://github.com/galaxyproject/galaxy/pull/9905 and of course compatible already with both  implementations


bgruening (approved):

dannon (approved):

OlegZharkov:
@dannon @bgruening thanks a lot!

PR #10094, opened 2020-08-12
Remove duplicates from tool recommendations

anuprulez:
Remove duplicates from tool recommendations in the workflow editor popover

### DIFF
--- a/client/src/components/ToolRecommendation.vue
+++ b/client/src/components/ToolRecommendation.vue
@@ -19,6 +19,7 @@ import * as d3 from "d3";
 import { getAppRoot } from "onload/loadConfig";
 import { getDatatypesMapper } from "components/Datatypes";
 import { getToolPredictions } from "components/Workflow/Editor/modules/services";
+import { getCompatibleRecommendations } from "components/Workflow/Editor/modules/utilities";
 
 export default {
     props: {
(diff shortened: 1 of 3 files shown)


dannon:
This looks fine for filtering on the client side, but is there a reason for sending the duplicates in the API response in the first place?  Can we just update the API to send a unique set?

anuprulez:
The API sends unique ones. The recommendations get duplicated while checking for compatibility using similar logic as used for showing orange and green colors in the workflow editor when pipes are connected between two tools.

https://github.com/galaxyproject/galaxy/blob/f74bc93211b0186949bad823c55f4993d65bbb2b/client/src/components/Workflow/Editor/Recommendations.vue#L129

These two `for loops` induce duplicates (but not always)

dannon:
@anuprulez Ahh, good!  What's a good tool or minimal workflow to test this with?

anuprulez:
@dannon `Bowtie2` can be a good example. Even though the API returns unique ones, a few recommendations get duplicated due to the compatibility checks (between all datatypes of the outgoing and incoming tools). I am working to include duplicate removal in the already existing `for loops` instead of writing new ones.

Please refer to the attached example of `Bowtie2` in the workflow editor:

Screenshot from 2020-08-12 18-41-24

dannon:
@anuprulez Thanks!  I just opened #10096, too, which I ran into testing.

anuprulez:
@dannon I should check these version upgrades. Maybe the API needs an adjustment. `Tensorflow 2.0` has brought a lot of changes and the `Keras` scripts should change accordingly. We used `Tensorflow 1.x` because it was compatible with the model (which has been trained using `Tensorflow 1.x`). There could be a problem while if the version of `Tensorflow` and/or `Keras` used in creating the model and inferencing from it differ. I will verify it soon. Thanks for your inputs.

dannon:
@anuprulez This works in my hands now.  It looks like there might be enough duplication between the two spots you've updated that it would warrant refactoring -- is that something you'd want to do now?

Additionally, in the future if you want you can just run `make client-format` or, from the client directory, `yarn run prettier` and it'll reformat all the files at once instead of having to track the issues down individually.

anuprulez:
@dannon Yes, it needs refactoring - there is some common script at two places where recommendations are collected. Working on it. Thanks!

anuprulez:
@dannon I moved the duplicate scripts to a method in a `utilities.js` file which serves both the checkpoints. In addition, I have found a minor issue (the toolboxes of the recommendations in the workflow editor do not take the names and tool names just appear as "name"). I have found the solution to this bug and will be fixing once this PR is merged. 

Thanks for your suggestions!

dannon:
@anuprulez Oh funny, I just opened #10101 noticing the same thing.  I should have read my email first :)

I saw the refactoring earlier and it looked good to me.  Feel free to open a separate PR addressing that, since this one is now passing relevant tests.

PR #10096, opened 2020-08-12
Tensorflow update to support python 3.8

dannon:
```
Collecting keras==2.2.4
  Using cached Keras-2.2.4-py2.py3-none-any.whl (312 kB)
ERROR: Could not find a version that satisfies the requirement tensorflow==1.15.2 (from -r /dev/stdin (line 5)) (from versions: 2.2.0rc3, 2.2.0rc4, 2.2.0, 2.3.0rc0, 2.3.0rc1, 2.3.0rc2, 2.3.0)
ERROR: No matching distribution found for tensorflow==1.15.2 (from -r /dev/stdin (line 5))
```
Ran into the above trying to test tool recommendations against python 3.8.5.  This should make it work and will be compatible with previous versions  as well, I hope -- opening to run tests.

### DIFF
--- a/lib/galaxy/tools/recommendations.py
+++ b/lib/galaxy/tools/recommendations.py
@@ -57,15 +57,15 @@ class ToolRecommendations():
             # keras is not downloaded because of conditional requirement and Galaxy does not build
             try:
                 from keras.models import model_from_json
-                import tensorflow.compat.v1 as tf
-                tf.disable_v2_behavior()
+                import tensorflow as tf
+                tf.compat.v1.disable_v2_behavior()
             except Exception:
                 trans.response.status = 400
                 return False
             # set graph and session only once
             if self.graph is None:
                 self.graph = tf.Graph()
-                self.session = tf.Session(graph=self.graph)
+                self.session = tf.compat.v1.Session(graph=self.graph)
             model_weights = list()
             counter_layer_weights = 0
             self.tool_recommendation_model_path = self.__download_model(remote_model_url)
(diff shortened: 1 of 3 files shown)


dannon:
Continuing discussion from #10094;

> @dannon I should check these version upgrades. Maybe the API needs an adjustment. Tensorflow 2.0 has brought a lot of changes and the Keras scripts should change accordingly. We used Tensorflow 1.x because it was compatible with the model (which has been trained using Tensorflow 1.x). There could be a problem while if the version of Tensorflow and/or Keras used in creating the model and inferencing from it differ. I will verify it soon. Thanks for your inputs.

Thanks!  Yeah, I used the tensorflow v1 compatibility layer which seems to be working in my hands with these commits.  I'm testing a followup using native TF2 interface (including built-in keras), and can post how it goes.

mvdbeek (approved):

bgruening:
ping @anuprulez

lib/galaxy/tools/recommendations.py
```diff
@@ -56,15 +56,15 @@ def __set_model(self, trans, remote_model_url):
             # import moves from the top of file: in case the tool recommendation feature is disabled,
             # keras is not downloaded because of conditional requirement and Galaxy does not build
             try:
-                from keras.models import model_from_json
                 import tensorflow as tf
+                tf.compat.v1.disable_v2_behavior()
```
anuprulez, line 60: @dannon I excluded this line and it worked fine. Can you confirm this? Thanks.

anuprulez:
I added one comment and all other changes except `tf.compat.v1.disable_v2_behavior()` are ok. :+1:

dannon:
@anuprulez Sorry, I missed the followup here.  Is there a downside to disabling v1 behavior?  It was the recommended upgrade path in the documentation I was reading, and I figured it would expose any potential issues.

PR #10098, opened 2020-08-12
fix masthead dropdown

### LINKED ISSUE #10092: Masthead menus do not disappear unless menu title clicked
It seems the masthead menu items became a toggle where when one clicks on a menu title, the menu opens but to hide the menu, you have to click on the menu title again as opposed to a random place on the page. This does not seem like feature to me?

image

OlegZharkov:
This PR fixes https://github.com/galaxyproject/galaxy/issues/10092 . 

Peek 2020-08-12 16-58

It took me a bit to understand why did it stop working. It was a race between Vue component and `galaxy_main` iframe. 

We added additional stuff to our home page and it slowed down the rendering, thus `document.getElementById("galaxy_main");` was returning null. Also we don't need this listener, if we don't have dropdown  `MastheadItem` component (added a check for that)

Thanks a lot @afgane for reporting!

### DIFF
--- a/client/src/components/Masthead/MastheadItem.vue
+++ b/client/src/components/Masthead/MastheadItem.vue
@@ -118,13 +118,13 @@ export default {
             return document.getElementById("galaxy_main");
         },
     },
-    mounted() {
-        if (this.galaxyIframe) {
+    updated() {
+        if (this.$refs.dropdown && this.galaxyIframe) {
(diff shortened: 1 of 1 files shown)


dannon:
Thanks, @OlegZharkov!

OlegZharkov:
thank you @dannon!

PR #10102, opened 2020-08-13
Test import dataset from path

OlegZharkov:
This test was commented out in `test_library_contents.py` and I really need it in https://github.com/galaxyproject/galaxy/pull/9905

This PR is created in order to test it on jenkins. It successfully passed in https://github.com/galaxyproject/galaxy/pull/9905 already on Jenkins. So I want to investigate and find out what is going on


OlegZharkov:
Similarly to #9905 this test passed with old library as well. I guess, we can uncomment it.

bgruening:
Cool thanks @OlegZharkov!

OlegZharkov:
thanks a lot @bgruening

PR #10104, opened 2020-08-13
Lib Folder pagination (backend)

OlegZharkov:
Some code from https://github.com/galaxyproject/galaxy/pull/9905
Backend stuff probably should not be there, so I rebased it in this PR
It allows client to ask for `limit` and `offset`

example:
http://localhost:8080/api/folders/{fodler_id}/contents?limit={int}&offset={int}

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/folder_contents.py
+++ b/lib/galaxy/webapps/galaxy/api/folder_contents.py
@@ -41,6 +41,10 @@ class FolderContentsController(BaseAPIController, UsesLibraryMixin, UsesLibraryM
         response as a separate object providing data for
         breadcrumb path building.
 
+        ..example:
+            limit and offset can be combined. Skip the first two and return five:
+                '?limit=3&offset=5'
+
         :param  folder_id: encoded ID of the folder which
             contents should be library_dataset_dict
         :type   folder_id: encoded string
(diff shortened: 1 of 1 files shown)


lib/galaxy/webapps/galaxy/api/folder_contents.py
```diff
@@ -232,7 +239,19 @@ def _load_folder_contents(self, trans, folder, include_deleted):
                 #         subfolder.api_type = 'folder'
                 #         content_items.append( subfolder )
 
-        for dataset in folder.datasets:
+        datasets_size = int(len(folder.datasets))
```
bgruening: len() should already return you an int()

bgruening (commented):
You can add an example as seen here: https://github.com/galaxyproject/galaxy/blob/2777db0e46f1185e47c41049abdb61f99ac1e2e7/lib/galaxy/webapps/galaxy/api/datasets.py#L88

lib/galaxy/webapps/galaxy/api/folder_contents.py
OlegZharkov: thanks, I will check

OlegZharkov:
> You can add an example as seen here:
> 
> https://github.com/galaxyproject/galaxy/blob/2777db0e46f1185e47c41049abdb61f99ac1e2e7/lib/galaxy/webapps/galaxy/api/datasets.py#L88

I can do that of course. But here we already have some sort of example https://github.com/galaxyproject/galaxy/pull/10104/files#diff-4a1789d4affda47540d7b94cac91340eR35 should I add an additional one?

bgruening (approved):

OlegZharkov:
thanks a lot @bgruening @bgruening @mvdbeek !!!

PR #10107, opened 2020-08-13
User self-delete

### LINKED ISSUE #6081: [GDPR] Implement manual control over Right to be Forgotten
Just need to let users self-delete. This is a post-may-25 goal.

astrovsky01:
Allow users to delete their own accounts without needing to go through an admin.

Requires confirmation by entering the account's associated email

fixes https://github.com/galaxyproject/galaxy/issues/6081

### DIFF
--- a/client/src/components/User/UserPreferences.vue
+++ b/client/src/components/User/UserPreferences.vue
@@ -37,32 +37,34 @@
             <div class="pref-content pr-1">
                 <a href="javascript:void(0)"><b v-b-modal.modal-prevent-closing>Delete Account</b></a>
                 <div class="form-text text-muted">
-                    Delete your user account. This will not purge your account immediately.
+                    Delete your account on this Galaxy server.
                 </div>
                 <b-modal
                     id="modal-prevent-closing"
+                    centered
                     ref="modal"
-                    title="Type your user email for this account as confirmation."
+                    title="Account Deletion"
+                    title-tag="h2"
                     @show="resetModal"
                     @hidden="resetModal"
                     @ok="handleOk"
                 >
                     <p>
-                        <b
-                            >This action cannot be undone. Your account will be permanently deleted, along with the data
-                            contained in it.</b
-                        >
+                        <b>
+                            This action cannot be undone. Your account will be permanently deleted, along with the data
+                            contained in it.
+                        </b>
                     </p>
-                    <form ref="form" @submit.stop.prevent="handleSubmit">
+                    <b-form ref="form" @submit.prevent="handleSubmit">
                         <b-form-group
                             :state="nameState"
-                            label="User email"
+                            label="Enter your user email for this account as confirmation."
                             label-for="Email"
                             invalid-feedback="Incorrect email"
                         >
                             <b-form-input id="name-input" v-model="name" :state="nameState" required></b-form-input>
                         </b-form-group>
-                    </form>
+                    </b-form>
                 </b-modal>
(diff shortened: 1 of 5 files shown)


astrovsky01:
@dannon we should also talk security, since this is no longer protected by admin status

lib/galaxy/webapps/galaxy/api/users.py
```diff
@@ -229,7 +229,6 @@ def update(self, trans, id, payload, **kwd):
         self.user_deserializer.deserialize(user_to_update, payload, user=current_user, trans=trans)
         return self.user_serializer.serialize_to_view(user_to_update, view='detailed')
 
-    @web.require_admin
```
dannon, line 232: > @dannon we should also talk security, since this is no longer protected by admin status

@astrovsky01 Yes, definitely.  You'll need to check inside the method if the individual making the request is an admin or if the target matches the authenticated transaction user.  As it is, just dropping @admin would allow anyone to delete arbitrary users.

dannon (changes requested):

lib/galaxy/webapps/galaxy/api/users.py
astrovsky01: I have the secondary request for inputting of the user's email and check it against who is logged in, but I feel as though password input should also be necessary.

dannon: I did see that, but that's not enough as it's purely clientside.  Never trust the client :)

One could skip the interface and just do:

`curl -X "DELETE" http://localhost:8080/api/users/NOTMYUSER`

And it'd work.  This must have serverside validation -- you'll want to compare `(trans.user_is_admin or trans.user == user_to_delete)`

astrovsky01: I'd figured, yeah.

client/src/components/User/UserPreferences.vue
```diff
@@ -178,10 +220,40 @@ export default {
...
+            // Prevent modal from closing
+            bvModalEvt.preventDefault();
+            // Trigger submit handler
+            this.handleSubmit();
+        },
+        async handleSubmit() {
+            if (!this.checkFormValidity()) {
+                return false;
+            }
+            var value = await axios.get(`${getAppRoot()}api/users/${this.userId}`);
```
dannon: We can drop this extra request since it's just a client side check since it's not really doing anything.   Simply validate that the email entered matches the clientside Galaxy user email, and attempt to post.  If the user is invalid, it'll get rejected by the API anyway.

client/src/components/User/UserPreferences.vue
```diff
@@ -178,10 +220,40 @@ export default {
...
+            var email = JSON.parse(JSON.stringify(value.data)).email;
+            var userId = JSON.parse(JSON.stringify(value.data)).id;
+            if (email === this.name) {
+                this.nameState = true;
+                await fetch(`${getAppRoot()}api/users/${userId}`, { method: "DELETE" });
```
dannon: Can you use `axios` for this request too?  And, after, on success, log out of Galaxy.

dannon (changes requested):

mvdbeek:
I'll push this to 21.01 due to the outstanding comments, but this would be very nice to get in @astrovsky01. If you can address the comments before the freeze on Friday we can still get this into 20.09.

PR #10108, opened 2020-08-14
[21.01] Properly log out of oidc on galaxy logout

juleengraham:
#10075

### DIFF
--- a/client/src/layout/menu.js
+++ b/client/src/layout/menu.js
@@ -14,12 +14,11 @@ export function userLogout(logoutAll = false) {
     const galaxy = getGalaxyInstance();
     const session_csrf_token = galaxy.session_csrf_token;
     const url = `${galaxy.root}user/logout?session_csrf_token=${session_csrf_token}&logout_all=${logoutAll}`;
-    
+
     var identities;
-    getIdentityProviders()
-        .then((results) => {
-            identities = results;
-        });
+    getIdentityProviders().then((results) => {
+        identities = results;
+    });
     axios
         .get(url)
         .then(() => {
--- a/client/src/layout/menu.js
+++ b/client/src/layout/menu.js
@@ -1,7 +1,6 @@
 import axios from "axios";
 import { getGalaxyInstance } from "app";
 import _l from "utils/localization";
-import { CommunicationServerView } from "layout/communication-server-view";
 import { getIdentityProviders } from "components/User/ExternalIdentities/service";
 
 const POST_LOGOUT_URL = "root/login?is_logout_redirect=true";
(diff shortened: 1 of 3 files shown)


lib/galaxy/authnz/managers.py
```diff
@@ -152,7 +152,7 @@ def _parse_custos_config(self, config_xml):
             'client_id': config_xml.find('client_id').text,
             'client_secret': config_xml.find('client_secret').text,
             'redirect_uri': config_xml.find('redirect_uri').text,
-            'enable_idp_logout': asbool(config_xml.findtext('enable_idp_logout', 'false'))}
+            'enable_idp_logout': asbool(config_xml.findtext('enable_idp_logout', 'true'))}
```
dannon: If I have two services, let's say two galaxy instances, both authenticated via custos, will this log me out of both, or just one?

client/src/layout/menu.js
```diff
@@ -15,7 +23,15 @@ export function logoutClick() {
             }
             // Check if we need to logout of OIDC IDP
             if (galaxy.config.enable_oidc) {
-                return axios.get(`${galaxy.root}authnz/logout`);
```
dannon, line 23: This doesn't look quite right.  What was authnz/logout returning?  The appropriate logout should have been passed through and handled as the redirect_uri.

dannon:
Please see failures in ci, as well.

client/src/layout/menu.js
juleengraham: The logout function wasn't running at all originally because the provider wasn't being passed (the endpoint is defined to receive provider in buildapp). I checked and the redirect is being returned

lib/galaxy/authnz/managers.py
juleengraham: no it does not. tested locally with two instances. I believe this is because each instance has its own database with a different token so they are operating independently

juleengraham: that is the preferred behaviour, right?

mvdbeek:
@dannon @juleengraham is this still needed ? If yes, can you resolve the conflict ?

afgane:
This is still needed. I can't figure out how to rerun the failing test though because the failure may have been an intermittent one... if someone could find the button please?

dannon:
@afgane This is still WIP; juleen is working on a different way to handle this without redundant requests.

dannon:
To update this again -- the email comparison method here doesn't work with multiple identities and we're working on alternatives.  But it does need to get into 21.01.

mvdbeek:
As a bug we can get this in anytime, but in order to manage the process I'll drop the target and instead put it in the release process for now: https://github.com/galaxyproject/galaxy/issues/10740#issuecomment-755705561

PR #10111, opened 2020-08-14
Don't fail to complete the job fail method if job outputs cannot be updated

natefoo:
They're failed outputs anyway. Otherwise this exception bubbles up and the handler will attempt to call `fail()` repeatedly forever until it works.

### DIFF
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1290,7 +1290,7 @@ class JobWrapper(HasResourceParameters):
                     dataset.extension = 'data'
                 try:
                     self.__update_output(job, dataset)
-                except:
+                except Exception:
                     # Failure to update the output of a failed job should not prevent completion of the failure method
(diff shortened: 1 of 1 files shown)


lib/galaxy/jobs/__init__.py
```diff
@@ -1288,7 +1288,12 @@ def fail(self, message, exception=False, tool_stdout="", tool_stderr="", exit_co
                 dataset.mark_unhidden()
                 if dataset.ext == 'auto':
                     dataset.extension = 'data'
-                self.__update_output(job, dataset)
+                try:
+                    self.__update_output(job, dataset)
+                except:
```
dannon: In this case I guess a bare exception probably makes sense?

mvdbeek: I think this should always be
```suggestion
                except Exception:
```
Otherwise we might catch `KeyboardInterrupt` or other things we shouldn't be catching.

PR #10112, opened 2020-08-15
Fix history sharing link (ref. #10040)

### LINKED ISSUE #10040: History sharing link in export confirmation message causes error
When exporting a history, the confirmation message includes a link to `history/sharing`:
`You will need to make this history 'accessible' in order to import this to another galaxy sever.`
However, the history ID is not passed (e.g., like when clicking the `share or publish` option in the history actions menu), so we get an internal server error page with this traceback:

```
  File "lib/galaxy/web/framework/decorators.py", line 170, in decorator
TypeError: sharing() missing 1 required positional argument: 'id'
```

jdavcs:
Fixes #10040.
I'm not sure whether this is the correct approach here. (ping @dannon)

**Tangent/discovered issue:** both `api.histories.HistoriesController` and `controllers.history.HistoryController` share `SharableMixin`. The mixin's `sharing()` method references `self.manager` and `self.serializer`, which exist only on `HistoriesConstoller`; `HistoryController` has `history_manager` and `history_serializer`. I haven't looked further into this (rabbit hole), but if we need both controllers, it will need to be addressed.

### DIFF
--- a/lib/galaxy/webapps/galaxy/controllers/history.py
+++ b/lib/galaxy/webapps/galaxy/controllers/history.py
@@ -1132,7 +1132,7 @@ class HistoryController(BaseUIController, SharableMixin, UsesAnnotations, UsesIt
         url = url_for(controller='history', action="export_archive", id=id, qualified=True)
         return trans.show_message("Exporting History '%(n)s'. You will need to <a href='%(share)s' target='_top'>make this history 'accessible'</a> in order to import this to another galaxy sever. <br/>"
                                   "Use this link to download the archive or import it to another Galaxy server: "
-                                  "<a href='%(u)s'>%(u)s</a>" % ({'share': url_for(controller='', action='histories/sharing', id=id), 'n': history.name, 'u': url}))
+                                  "<a href='%(u)s'>%(u)s</a>" % ({'share': url_for('/histories/sharing', id=id), 'n': history.name, 'u': url}))
         # TODO: used in this file and index.mako
 
     @web.expose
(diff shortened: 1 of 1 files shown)


lib/galaxy/webapps/galaxy/controllers/history.py
```diff
@@ -1130,9 +1130,9 @@ def export_archive(self, trans, id=None, gzip=True, include_hidden=False, includ
...
         self.queue_history_export(trans, history, gzip=gzip, include_hidden=include_hidden, include_deleted=include_deleted)
         url = url_for(controller='history', action="export_archive", id=id, qualified=True)
-        return trans.show_message("Exporting History '%(n)s'. You will need to <a href='%(share)s'>make this history 'accessible'</a> in order to import this to another galaxy sever. <br/>"
+        return trans.show_message("Exporting History '%(n)s'. You will need to <a href='%(share)s' target='_top'>make this history 'accessible'</a> in order to import this to another galaxy sever. <br/>"
                                   "Use this link to download the archive or import it to another Galaxy server: "
-                                  "<a href='%(u)s'>%(u)s</a>" % ({'share': url_for(controller='history', action='sharing'), 'n': history.name, 'u': url}))
+                                  "<a href='%(u)s'>%(u)s</a>" % ({'share': url_for(controller='', action='histories/sharing', id=id), 'n': history.name, 'u': url}))
```
dannon:
```suggestion
                                  "<a href='%(u)s'>%(u)s</a>" % ({'share': url_for(action='histories/sharing', id=id), 'n': history.name, 'u': url}))
```

PR #10117, opened 2020-08-17
Performance testing infrastructure.

jmchilton:


.github/workflows/perf_tests.yaml
```diff
@@ -0,0 +1,93 @@
...
+      id: pip-cache
+      with:
+        path: ~/.cache/pip
+        key: pip-cache-${{ matrix.python-version }}-${{ hashFiles('galaxy root/requirements.txt') }}
+    - name: Move test data
+      run: rsync -av --remove-source-files --exclude .git galaxy-test-data/ 'galaxy root/test-data/'
+    - name: Move test data base
+      run: rsync -av --remove-source-files --exclude .git galaxy-test-data/ 'galaxy root base/test-data/'
+    - name: Run tests on PR branch
+      run: 'GALAXY_TEST_DBURI=$GALAXY_TEST_DBURI_PR ./run_tests.sh --ci_test_metrics --structured_data_report_file "pr.json" -api
```
jdavcs: @jmchilton Do we need this to be quoted? There seem to be no special characters that require quoting in YAML. If we *do* need quotes here, then the closing quote is missing, right?
EDIT: yep, the quote is missing - it got dropped with the path to a specific api test.

.github/workflows/perf_tests.yaml
```diff
@@ -0,0 +1,93 @@
...
+      working-directory: 'galaxy root'
+    - name: Run tests on target branch
+      run: 'GALAXY_TEST_DBURI=$GALAXY_TEST_DBURI_TARGET ./run_tests.sh --ci_test_metrics --structured_data_report_file "target.json" -api
```
jdavcs: same as above

hexylena:
> I like the idea of using statsd but it seems excessive and I'm worried the telegraf scraping is at too wide for test cases and it seems like unnecessary infrastructure in the context of narrow API

telegraf/statsd seem a bit excessive maybe, have you considered something like https://github.com/airspeed-velocity/asv/ ? I tested it out with gxadmin (not a normal use case) and it was pretty nice for per-commit/per-infra setup performance numbers.

mvdbeek:
That does look interesting, but the docs are a bit basic. Did you manage to do timings around log statements ? The benchmarks all seem to revolve around total runtime when asv runs the actual code, which doesn't seem like it would be applicable / detailed enough ?

hexylena:
Hey @mvdbeek. I didn't try and do timings around log statements, only around select function calls. I think it grew out of [benchmarking individual numpy function calls](https://pv.github.io/numpy-bench/) which felt like a good fit for galaxy, but I didn't look closely at the PR and if you're doing it based off of the diff between two log statements, maybe it isn't such an easy solution.

That said, if you want to use their viz infrastructure, since you're already writing out JSON, you could just write something like [their JSON format](https://github.com/usegalaxy-eu/gxadmin/blob/master/.asv/results/mk/e79151c6-virtualenv-py2.7.json) (ugly I know) and get the nice auto-generated website for free?

.github/workflows/perf_tests.yaml
```diff
@@ -57,7 +57,7 @@ jobs:
       run: 'GALAXY_TEST_DBURI=$GALAXY_TEST_DBURI_PR ./run_tests.sh --ci_test_metrics --structured_data_report_file "pr.json" -api
```
jdavcs: @jmchilton also line 57 ;)

jmchilton:
I spent some time looking at allure for another context, it wasn't exactly what I'd want for this but it might be able to get there with the right plugin and some bootstrap. The numpy benchmarking tool looks awesome though.

What this PR currently produces is comparisons of the PR branch versus the target that look like:

Screen Shot 2020-12-15 at 12 09 13 PM

Screen Shot 2020-12-15 at 12 08 59 PM

You can download these artifacts from the Github action right now. There is another version that is just the timings without the comparison that I wanted to include on all the tests - but it really slowed down the tests immensely and to the point where they didn't get close to finishing it looks like so I reverted that change in  the latest commit and just did the PR vs target comparison for the performance tests instead of for all the API tests.

jmchilton:
Older version of this PR backed up (https://github.com/jmchilton/galaxy/pull/new/performance_tests_backup). Did a very stripped down rebase into a single commit just now.

ic4f:
Can I run`test_workflow_framework_performance.py` as a standalone test? I'm getting a logging error: `ValueError: I/O operation on closed file` - is this because I missed an option I should've set or because it's not supposed to be run separately?

jmchilton:
It works for me to just run it directly with pytest (https://gist.github.com/jmchilton/e6007becb21db94db4c392b861094f6d). Is there any more context around that error?

ic4f:
Thank you for the gist - that's exactly the output I got. I was referring to this error (which, I assume, is not relevant) - https://gist.github.com/jmchilton/e6007becb21db94db4c392b861094f6d#file-gistfile1-txt-L36  
I'll review this tonight, or tomorrow at the latest.

scripts/tests_markdown.tpl
```diff
@@ -0,0 +1,138 @@
+{% macro print_statistics_table(has_metrics) -%}
+
+        | | *Value* |
+        |-|-|
+        | **Sum** *(ms)* | ``{{ has_metrics.sum }}`` |
+        | **Median** *(ms)* | ``{{ has_metrics.median }}`` |
+        | **Mean** *(ms)* | ``{{ has_metrics.mean }}`` |
+        | **Standard Deviation** | ``{{ has_metrics.stdev }}`` |
```
jdavcs, line 8:
```suggestion
        | **Sum** *(ms)* | ``{{ has_metrics.sum|sum }}`` |
        | **Median** *(ms)* | ``{{ "%.2f" % has_metrics.median }}`` |
        | **Mean** *(ms)* | ``{{ "%.2f" % has_metrics.mean }}`` |
        | **Standard Deviation** | ``{{ "%.4f" % has_metrics.stdev if has_metrics.stdev is not none else "n/a" }}`` |
```
A very minor tweak for purely aesthetic reasons? I don't think the number of significant digits matters as long as it's consistent, and sum is always an int. Standard deviation can be the same or +2 digits, etc, as long as it's consistent.

scripts/tests_markdown.py
```diff
@@ -0,0 +1,306 @@
...
+
+
+def __inject_statistics(from_dict):
+    raw_values = from_dict['raw']
+    from_dict['sum'] = sum(raw_values)
+    from_dict['median'] = statistics.median(raw_values)
+    if len(raw_values) > 1:
+        from_dict['stdev'] = statistics.stdev(raw_values)
+    else:
+        from_dict['stdev'] = "n/a"
```
jdavcs, line 268:
```suggestion
        from_dict['stdev'] = None
```
To accommodate float formatting of statistics in the templates (see suggesion below). Also, that way we leave the choice of *how* to display stuff to the template, which, I think, is the appropriate place for it anyway.

scripts/tests_markdown_compare.tpl
```diff
@@ -0,0 +1,62 @@
+{% macro print_statistics_table(has_metrics) -%}
+
+        | | *{{ has_metrics.keys() | join("* | *") }}* |
+        |-|-{{ has_metrics.values() | map(attribute="absent", default="-") | join("-|-") }}-|
+        | **Sum** *(ms)* | ``{{ has_metrics.values() | map(attribute="sum") | join("`` | ``") }}`` |
+        | **Median** *(ms)* | ``{{ has_metrics.values() | map(attribute="median") | join("`` | ``") }}`` |
+        | **Mean** *(ms)* | ``{{ has_metrics.values() | map(attribute="mean") | join("`` | ``") }}`` |
+        | **Standard Deviation** | ``{{ has_metrics.values() | map(attribute="stdev") | join("`` | ``") }}`` |
```
jdavcs, line 8: same formatting suggestion as in `tests_markdown.tpl`

.github/workflows/perf_comp_api_tests.yaml
```diff
@@ -0,0 +1,93 @@
...
+        name: 'Performance Metrics PR'
+        path: 'galaxy root/pr.html'
+    - name: Generate combined test HTML
+      run: '. .venv/bin/activate; python scripts/tests_markdown.py --output_path pr_vs_target.html pr.json "../galaxy root base/target.json"'
+      working-directory: 'galaxy root'
+    - uses: actions/upload-artifact@v2.0.1
+      with:
+        name: 'Performance Metrics (PR vs Target)'
+        path: 'galaxy root/pr_vs_target.html'
+    #- name: Sleep 5
```
jdavcs: I'm fine with this block (and the block above) being commented out; but just checking - is it supposed to be left in the final version? And if so, would you mind adding a brief comment explaining its purpose? (I think it would be helpful, unless it's completely obvious, in which case please ignore my comment)

scripts/tests_markdown.py
```diff
@@ -0,0 +1,306 @@
...
+        raw_data_dicts.append(_prepare_raw_data(arg))
+    environment = _merge_summarizes(raw_data_dicts)
+    environment.update(common_env)
+    template = template_env.get_template(TEMPLATE_COMPARE_FILE)
+    output = template.render(environment)
+    return output
+
+
+def _empty_statistics():
+    return {"count": 0, "sum": "n/a", "median": "n/a", "stdev": "n/a", "mean": "n/a"}
```
jdavcs, line 60:
```suggestion
    return {"count": 0, "sum": 0, "median": None, "stdev": None, "mean": None}
```
This is relevant *only* if you make the formatting change in the template (see suggestion below), this needs to be reset to the proper values. I think 0 is appropriate for `sum`, but maybe `None` makes more sense?

scripts/tests_markdown.py
```diff
@@ -0,0 +1,306 @@
...
+
+        test_endpoints = test['internals_metrics']
+        for api_endpoint, endpoint_metrics in test_endpoints.items():
+            _ensure_has_dict_at_key(internals, api_endpoint,
+                total_time={'raw': []},
+            )
+            internals[api_endpoint]['label'] = endpoint_metrics['label']
+            internals[api_endpoint]['total_time']['raw'].extend(endpoint_metrics['total_time']['raw'])
+
+    for api_endpoint, endpoint_metrics in api_endpoints.items():
```
jdavcs:
```suggestion
    for endpoint_metrics in api_endpoints.values():
```
keys are not used

scripts/tests_markdown.py
```diff
@@ -0,0 +1,306 @@
...
+        __inject_statistics(endpoint_metrics['total_time'])
+        __inject_statistics(endpoint_metrics['sql_time'])
+        __inject_statistics(endpoint_metrics['sql_queries'])
+
+    for _, endpoint_metrics in internals.items():
```
jdavcs:
```suggestion
    for endpoint_metrics in internals.values():
```
same as above?

jdavcs (approved):
This is so very cool! (that is, once I figured out how to run the tests and generate the reports!) I'm very sorry it took me so long to review this - now it has to be rebased. All my comments are minor suggestions only.

.github/workflows/perf_comp_api_tests.yaml
jmchilton: I think I meant to drop this whole file when cleaning it up - I wanted to find a way to just do this on request. I'll rebase without this.

scripts/tests_markdown.py
jmchilton: I kind of wish I wouldn't have - there is not a clean way to do the map and it seems like they prefer formatting before you hit jinja2. https://github.com/pallets/jinja/pull/906#issuecomment-548393215.

I'm going to try to rework this to do the formatting for the fix precision but not do the formatting in the template.

jdavcs: Oh, I see. Sorry for suggesting that route! And yes, it's not like we use that data in the template in different ways, so it might as well come formatted as a string.

jmchilton:
Exploiting that approve to merge without a final check because I need to show some progress at PI meeting Monday, but I just rebased with the above comments. Thanks for the detailed review @ic4f!

ic4f:
Thank you for putting all that work into addressing those minor comments! Again, sorry for the delayed review @jmchilton !

PR #10118, opened 2020-08-17
Render workflow invocation details

mvdbeek:
Isn't that useful yet, but I think we can build something cool for this.

Screenshot 2020-08-17 at 15 52 14

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -5297,27 +5297,41 @@ class WorkflowInvocation(UsesCreateAndUpdateTime, Dictifiable, RepresentById):
             rval['steps'] = steps
 
             inputs = {}
-            for step in self.steps:
-                if step.workflow_step.type == 'tool':
-                    for job in step.jobs:
-                        if job is None:
-                            continue
-                        for step_input in step.workflow_step.input_connections:
-                            output_step_type = step_input.output_step.type
-                            if output_step_type in ['data_input', 'data_collection_input']:
-                                src = "hda" if output_step_type == 'data_input' else 'hdca'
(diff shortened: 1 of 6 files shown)


lib/galaxy/model/__init__.py
```diff
@@ -5297,34 +5297,49 @@ def to_dict(self, view='collection', value_mapper=None, step_details=False, lega
...
-                                src = "hda" if output_step_type == 'data_input' else 'hdca'
-                                for job_input in job.input_datasets:
-                                    if not job_input.dataset:
-                                        continue
-                                    if job_input.name == step_input.input_name:
-                                        inputs[str(step_input.output_step.order_index)] = {
-                                            "id": job_input.dataset_id, "src": src,
-                                            "uuid" : str(job_input.dataset.dataset.uuid) if job_input.dataset.dataset.uuid is not None else None
-                                        }
+            for input_item_association in self.input_datasets + self.input_dataset_collections:
```
jmchilton, line 5300: ❤️ 

Last time I changed something like this I did break some people's code I think (see ``legacy_job_state`` in the code base), but what you're doing here is a much better thing to do IMO. I love this.

guerler:
This could be useful, I suggest to either just display a few attributes and properly format them or to display the entire json using prettier and courier font to indicate that it is a raw json. Maybe in a scrollable div, similar to how job submission errors are displayed.

mvdbeek:
Thanks @guerler, that was just my first steps with vue. There's much more progress in https://github.com/galaxyproject/galaxy/compare/dev...mvdbeek:work_off_of_masons_branch ... I'd even call it feature-complete, but we first need to get the history PR in.

PR #10121, opened 2020-08-18
Revision of Workflow Reports

guerler:
This PR revises the appearance and layout of Workflow Reports / Pages building on the work of @jmchilton. It also adds Vue components for all embeddable elements, legacy mako templates and jquery have been removed. The overall layout of reports is now similar to an online publication. A short fixed header (see blue bar below) has been added, it contains the reports creation time and additional basic metadata above it.
The figure shows an example report with random placeholder content, as always its all open for suggestions:

image

Note that the image of the Workflow shown above is a regular screenshot, although it would be nice to produce it automatically in the future.


client/src/components/Markdown/Markdown.vue
```diff
@@ -30,170 +91,25 @@ const md = MarkdownIt();
...
-    $("." + tag).each((i, el) => {
-        const objectId = $(el).attr(idAttr);
-        const meta = objects[objectId][metadataKey];
-        $(el).find("code").text(meta);
-    });
-}
-
 export default {
+    store: store,
+    components: {
```
OlegZharkov, line 97: wow, this is super cool! way way cleaner than calling mount for each component

guerler: Cool (: Thanks for looking through this. I noticed that https://github.com/galaxyproject/galaxy/blob/e64712e1fbb840f6b22d1283895e6189e0b26027/client/src/components/Dataset/DatasetAsImage/DatasetAsImage.vue uses the store while the wrapper I added just uses the `link` see: https://github.com/galaxyproject/galaxy/blob/e64712e1fbb840f6b22d1283895e6189e0b26027/client/src/components/Markdown/Elements/HistoryDatasetAsImage.vue. If necessary, we can adjust or consolidate those two back into one.

bgruening:
Oh my goodness ... what ... how ... :heart:  :heart:  :heart:  :heart:  :heart:  :heart:

lib/galaxy/managers/workflows.py
```diff
@@ -583,6 +585,106 @@ def _workflow_to_dict_run(self, trans, stored, workflow):
             'workflow_resource_parameters': self._workflow_resource_parameters(trans, stored, workflow),
         }
 
+    def _workflow_to_dict_preview(self, trans, workflow):
+        """
+        Builds workflow dictionary containing input labels and values.
```
mvdbeek, line 590: If this is to be embedded into a workflow report would it make more sense to process a workflow invocation ?

guerler: I think in many cases it might make sense to display an `Invocation` instead or in addition to a `Workflow` but they don't exclude each other. The content of the dictionary generated above is similar to the content which was previously generated by the Workflow `display.mako`. I tried to use one of the available `Workflow` export formats like yml, ga, etc but none seemed to produce the data needed to render this content.

bgruening:
Hi @guerler, I was able to create a crash with this simple workflow 
[Galaxy-Workflow-Unnamed_workflow.txt](https://github.com/galaxyproject/galaxy/files/5098285/Galaxy-Workflow-Unnamed_workflow.txt)

```
Diese Seite verwendet die nicht standardisierte Eigenschaft "zoom". Stattdessen sollte calc() in den entsprechenden Eigenschaftswerten oder "transform" zusammen mit "transform-origin: 0 0" verwendet werden. report
Polyfills are running polyfills.js:28:12
JQMIGRATE: Migrate is installed with logging active, version 1.4.1 jquery-migrate.js:26:16
Configs: production configs 
Object { name: "production configs" }
index.js:21:8
accessing (get) window.Galaxy /workflows/invocations/report monitor.js:18:20
Synchrone XMLHttpRequests am Haupt-Thread sollte nicht mehr verwendet werden, weil es nachteilige Effekte für das Erlebnis der Endbenutzer hat. Für weitere Hilfe siehe http://xhr.spec.whatwg.org/ jquery.js:9064:8
Initializing javascript application: app report:44:17
setGalaxyInstance /workflows/invocations/report singleton.js:12:16
runInitializations analysis /workflows/invocations/report standardInit.js:48:16
XHRGEThttp://127.0.0.1:8080/api/invocations/33b43b4e7093c91f/report
[HTTP/1.1 500 Internal Server Error 109ms]

Error: Request failed with status code 500
    exports createError.js:16
    exports settle.js:17
    onreadystatechange xhr.js:61
    exports xhr.js:36
    exports xhr.js:12
    exports dispatchRequest.js:52
    promise callback*u.prototype.request Axios.js:61
    t Axios.js:76
    exports bind.js:9
    loadMarkdownConfig WorkflowInvocationReport.vue:71
    created WorkflowInvocationReport.vue:35
    VueJS 4
    _display_vue_helper AnalysisRouter.js:107
    show_workflow_invocation_report AnalysisRouter.js:194
    execute router.js:45
    Backbone 2
    F Underscore
    Backbone 2
    initialize page.js:93
    Backbone 2
    <anonymous> index.js:20
    R standardInit.js:49
    R standardInit.js:49
    RxJS 5
WorkflowInvocationReport.vue:76

```

bgruening:
As a general question, do you think we can return some dummy rendering of every component (maybe even useful as a test)? That way the user could render such a report inside the workflow editor and gets a feeling of how such a report would look like before the workflow is executed.

guerler:
@bgruening thanks a lot for looking into this. a) The crash is not related to this PR. It is caused by the backend due to the false markdown entry: `history_dataset_info("foo")` which should be `history_dataset_info(output="foo")`. Regardless this shows that we need to revisit the backend to improve its resilience. Instead of crashing, a proper error should be displayed to the user. 
b) As of right now, placeholders/dummies are not rendered in the current Markdown editor, but we can consider implementing them when working on the Editor itself. Alternatively we could split the screen in the Editor, and show the raw markdown on the left and the rendered preview on the right.

bgruening:
Ok, I can follow the argument that this is a more general problem :) I tried to play around with all kinds of things and thought the UI should give me some indication. But happy to take this to another PR.

:+1: The rests works very well!

jmchilton (approved):
Amazing - a million thanks!

client/src/components/JobParameters/JobParameters.vue
```diff
@@ -56,16 +56,19 @@ export default {
...
             type: String,
+            default: null,
         },
         datasetType: {
             type: String,
             default: "hda",
         },
         param: {
             type: String,
+            default: undefined,
```
OlegZharkov, line 71: why some are null and others are undefined?

client/src/components/Markdown/Markdown.vue
OlegZharkov: > Cool (: Thanks for looking through this. I noticed that https://github.com/galaxyproject/galaxy/blob/e64712e1fbb840f6b22d1283895e6189e0b26027/client/src/components/Dataset/DatasetAsImage/DatasetAsImage.vue uses the store while the wrapper I added just uses the `link` see: https://github.com/galaxyproject/galaxy/blob/e64712e1fbb840f6b22d1283895e6189e0b26027/client/src/components/Markdown/Elements/HistoryDatasetAsImage.vue. If necessary, we can adjust or consolidate those two back into one.

Consolidation would be a good idea. Otherwise, we should replace ``  with `<b-img>` in https://github.com/galaxyproject/galaxy/blob/e64712e1fbb840f6b22d1283895e6189e0b26027/client/src/components/Dataset/DatasetAsImage/DatasetAsImage.vue

client/src/components/JobParameters/JobParameters.vue
guerler: I prefer all of them to be always `null` as default (or an actual value), however this component which is reused in this PR, relies on `param` to be `undefined` and I did not want to change it yet. See below: https://github.com/galaxyproject/galaxy/blob/241ca75bb62588c6814dc44bc01d484cabb6911d/client/src/components/JobParameters/JobParameters.vue#L96
Apparently there is also an unnecessary condition duplication in this line, once the default is set to `null`, the condition should be: `!!this.param`. As I said however I just fixed the linting error complaing about a missing `default` value and avoided modifying the existing code in this component. I can do that in the follow-up.

OlegZharkov (approved):
Thanks a lot @guerler . That's an amazing PR :+1:

bgruening:
Thanks for the review @OlegZharkov! Thanks @guerler - super cool!

guerler:
Thanks so much for the reviews and comments! I will keep them in mind for the follow-up.

PR #10122, opened 2020-08-18
expose dataset 'in use state' through api

scholtalbers:
This allows one to check if a dataset is currently actively in use as in- or output.


OlegZharkov:
@galaxybot test this

PR #10123, opened 2020-08-18
Jest-based client testing.

dannon:
This replaces all of our karma+mocha+chai tests with jest-based unit testing.  Jest is faster, has more features, a better API, better output, better mock and watch functionality.  It's the standard we'll want to work against moving forward for clientside unit testing.  It supports coverage and snapshot testing which I haven't enabled yet, but look forward to experimenting with.

Just try it out, the developer experience is significantly better than the old stuff.  Watching files while you're working with filters works really, really well:

`yarn run jest-watch savedrules`
`yarn run jest-watch tags`

Or you can run without watching like:
`yarn run jest`
`yarn run jest workflow/editor`
`yarn run jest store`


Nerdinacan:
Love it. 

I would also love it if we could include suggestions on how to place a breakpoint and debug your unit tests. That is sometimes super arcane information.

dannon:
@Nerdinacan I'll include some debugging info, I still need to go through the docs with updates.

This also doesn't yet replace all the Simon/rewire/etc with jest based mocks, which I want to do.

client/README.md
```diff
@@ -49,197 +48,172 @@ for inspection of live javascript to facilitate debugging, use:
...
-processes here.  The first command below starts a special webpack dev server
+to iterate when developing the client. You'll need to start two separate
+processes here. The first command below starts a special webpack dev server
 after a client build, and the second starts a Galaxy server like usual, but
 with extra mappings that redirect client artifact requests to the mentioned
 webpack dev server.
 
-    make client-dev-server
-    GALAXY_CLIENT_DEV_SERVER=1 sh run.sh
+    make client-dev-server GALAXY_CLIENT_DEV_SERVER=1 sh run.sh
```
nsoranzo: I think these should stay on separate lines?

client/README.md
```diff
@@ -49,197 +48,172 @@ for inspection of live javascript to facilitate debugging, use:
...
 
 Note that this only works under uWSGI due to the extra internal routing rules
-employed.  If you're using the older Paste-based galaxy webserver you'll need
-to swap it over to take advantage of this functionality.
-
+employed. If you're using the older Paste-based galaxy webserver you'll need to
```
nsoranzo, line 86:
```suggestion
employed. If you're using the older Paste-based Galaxy webserver you'll need to
```

client/README.md
```diff
@@ -49,197 +48,172 @@ for inspection of live javascript to facilitate debugging, use:
...
 
-## Testing Technologies
+### Testing Technologies
 
-[Galaxy uses Mocha](https://mochajs.org/) for its client-side unit testing
-framework with [Chai as its assertion library](https://www.chaijs.com/) and
-[karma as a test runner](https://karma-runner.github.io/latest/index.html).
-The karma configs are contained in the client/karma folder. Unit tests are run
-using Chrome headless by default.
+[Galaxy uses Jest](https://jestjs.io/) for its client-side unit testing
```
nsoranzo, line 112:
```suggestion
Galaxy uses [Jest](https://jestjs.io/) for its client-side unit testing
```

client/README.md
dannon: Yeah; vim wrap fail.  Thanks!

client/README.md
```diff
@@ -249,18 +223,22 @@ Separate your concerns and identify the developer-facing methods and functions
 you expect them to use. Test THOSE. Everything else should probably be
 considered an implementation detail.
 
-Assume nobody cares *how* your code works, we just need to know that the public
-API you designed *does* work. If performance problems or new tech necessitate a
-re-write, these tests become a guide for the next implementation.
+The other side of the same coin is to test *only* the unit in question.  If your
+component has a model that uses a service that touches Vuex, which then uses
+Axios to fetch some data -- don't test all that at once.  Break things apart and
+mock functionality to isolate testing to units.  End to end testing is a
```
nsoranzo, line 230:
```suggestion
mock functionality to isolate testing to units.  End-to-end testing is a
```

dannon:
(selenium test failure is legit -- I need to fix it)

bgruening:
Oo, this has a merge conflict now...

dannon:
@bgruening Thanks for the ping, I'll fix it!

client/src/components/Dataset/DatasetName.test.js
```diff
@@ -1,35 +1,45 @@
 import { mount } from "@vue/test-utils";
 import DatasetName from "./DatasetName";
+import { getNewAttachNode } from "../../../tests/jest/helpers";
```
OlegZharkov: import { getNewAttachNode } from "jest/helpers";
```suggestion
import { getNewAttachNode } from "jest/helpers";
```

client/src/components/RuleBuilder/SavedRulesSelector.test.js
```diff
@@ -1,15 +1,19 @@
 import Vue from "vue";
 import { mount } from "@vue/test-utils";
 import SavedRulesSelector from "components/RuleBuilder/SavedRulesSelector";
+import { getNewAttachNode } from "../../../tests/jest/helpers";
```
OlegZharkov: import { getNewAttachNode } from "jest/helpers";
```suggestion
import { getNewAttachNode } from "jest/helpers";
```

client/src/components/Tags/StatelessTags.test.js
```diff
@@ -7,8 +7,9 @@ describe("Tags/StatelessTags.vue", () => {
     const localVue = createLocalVue();
     localVue.filter("localize", (value) => _l(value));
 
-    let testTags = ["abc", "def", "ghi"];
-    let wrapper, emitted;
+    const testTags = ["abc", "def", "ghi"];
+    let wrapper;
+    let emitted;
```
OlegZharkov, line 12: I just saw it in couple of places. Is there a reason why separate lines?

client/src/components/JobDestinationParams/JobDestinationParams.test.js
```diff
@@ -15,53 +12,50 @@ describe("JobDestinationParams/JobDestinationParams.vue", () => {
...
-        expect(params.length).to.equals(responseKeys.length);
+        expect(params.length).toBe(responseKeys.length);
 
         for (let counter = 0; counter < responseKeys.length - 1; counter++) {
             const parameter = params.at(counter).findAll("td");
             const parameterTitle = parameter.at(0).text();
             const parameterValue = parameter.at(1).text();
 
-            assert(responseKeys.includes(parameterTitle), "rendered parameter should exist in test data!");
+            expect(responseKeys.includes(parameterTitle)).toBeTruthy();
```
OlegZharkov, line 54: loosing all error messages isn't that great....
If you want, I can try https://github.com/mattphillips/jest-expect-message on this PR.
Otherwise, we might want to consider to define smth like this globally. 
https://stackoverflow.com/a/45375363/4870846

client/src/components/Dataset/DatasetName.test.js
dannon: Good catch; I think I did this when moving stuff around and forgot to clean it up.

client/src/components/Tags/StatelessTags.test.js
dannon: There's still a bit of a mix in the codebase, but the vast majority adheres to https://eslint.org/docs/rules/one-var.  I probably need to update the rule to enforce let/const as well.

client/src/components/JobDestinationParams/JobDestinationParams.test.js
OlegZharkov: > loosing all error messages isn't that great....
> If you want, I can try https://github.com/mattphillips/jest-expect-message on this PR.
> Otherwise, we might want to consider to define smth like this globally.
> https://stackoverflow.com/a/45375363/4870846

sorry, wrote this yesterday and did not submit. It's very weird thing on github. I doesn't write reviews right away, you kinda have to confirm afterwards 0_o . We all should move to gitlab :+1:

client/src/components/Tags/StatelessTags.test.js
OlegZharkov: but still, why not together?

client/src/components/JobDestinationParams/JobDestinationParams.test.js
dannon: We don't need the extra messages with well written tests.  Part of the point of the BDD descriptors is that when this actually fails Jest gives a very nice message that says exactly what's wrong.

client/src/components/Tags/StatelessTags.test.js
dannon: Here's a good explanation I mostly agree with.  https://github.com/airbnb/javascript#variables.

PR #10124, opened 2020-08-18
Copy to clipboard, instead of pop-up

OlegZharkov:
It replaces "copy pop-up" on tools page and copies link automatically

image

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -29,6 +29,7 @@
     "bootstrap-tour": "https://github.com/sorich87/bootstrap-tour.git#65a49742e131d19f41e3f5bf63588995f7b8a9e0",
     "bootstrap-vue": "^2.15.0",
     "citation-js": "^0.5.0-alpha.5",
+    "copy-to-clipboard": "^3.3.1",
     "d3": "3",
     "decode-uri-component": "^0.2.0",
     "elkjs": "^0.6.2",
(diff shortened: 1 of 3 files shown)


dannon:
@OlegZharkov Thanks!  When this was added, browser support for writing to the clipboard wasn't what it is now.  These days we may be able to just drop the extra dependency and use clipboard-write?

OlegZharkov:
@dannon

Oh, that's actually a very good idea. But that won't work on IE. Do we drop it?

dannon:
Looks like IE is .5% of the traffic on usegalaxy.org, and in testing it doesn't even load for me on IE anyway.  I think we can just drop the compatibility shim here.  I do need to add a warning saying "Hey, if you're using IE, please swap to firefox/chrome/edge/any-modern-thing to use Galaxy".

OlegZharkov:
> Looks like IE is .5% of the traffic on usegalaxy.org, and in testing it doesn't even load for me on IE anyway. I think we can just drop the compatibility shim here and use the clipboard API directly. I do need to add a warning saying "Hey, if you're using IE, please swap to firefox/chrome/edge/any-modern-thing to use Galaxy", though, because right now it's just a blank page.

.5% such a number :)

I kinda thought that we still support it. I changed it to clipboard-write. Thanks a lot for advise! :+1:

dannon:
@OlegZharkov I thought a little more about it and I added a fallback to the old functionality in the event the standard API isn't working, just in case.  I think this is great though, thank you for working on it!

OlegZharkov:
> @OlegZharkov I thought a little more about it and I added a fallback to the old functionality in the event the standard API isn't working, just in case. I think this is great though, thank you for working on it!

thanks!

PR #10125, opened 2020-08-18
Irods data migration

kxk302:
This is not fully cooked yet, but I thought I create the PR to get some feedback. The Python script does the following:

1. It is given a dataset id as input
2. Reads the row in the dataset table to get the UUID and object store ID
3. Given the object store ID, it gets the object store path (passed in as a hashmap input to the Python script)  
4. Creates the sub-folder name for source file (on disk)
5. Creates the sub-folder name for the destination file (on irods)
6. Constructs full path for source and destination files
7. Creates the irods subdirectory
8. iputs the source file as destination file

The idea is that this Python script is called in Bash script in a loop for a range of object IDs, to gradually migrate disk data to irods, initially on Test and then on Main.


kxk302:
@mvdbeek @natefoo I am trying to understand whether the above logic works for files generated by, say, fastqc, which have extra files in subdirectories.

scripts/objectstore/copy_files_to_irods.py
```diff
@@ -0,0 +1,94 @@
+import os
+import subprocess
+import sys
+import time
+sys.path.insert(0, '../../packages/objectstore/')
```
kxk302: Note to self. This path update is not needed. Must install all Galaxy packages in a venv instead (plus psycopg2)

scripts/objectstore/copy_files_to_irods.py
```diff
@@ -0,0 +1,94 @@
...
+
+from psycopg2 import connect, sql
+
+from galaxy.util import directory_hash_id
+
+
+def copy_file_to_irods(dataset_id, objectstore_id="", galaxy_dir="/Users/kxk302/workspace/galaxy", object_store_info={"files1" : "database/files1", "files2" : "database/files2"}, irods_home="/tempZone/home/rods"):
```
kxk302: Can remove objectstore_id from list of input parameters and down in sql_statement.

lib/galaxy/config/sample/object_store_conf.xml.sample
```diff
@@ -40,6 +40,20 @@
              <extra_dir type="temp" path="database/tmp_s3"/>
         </object_store>
         -->
+        
+        <!-- Sample iRods Object Store
```
kxk302, line 44: Added sample irods configuration to comments section.

scripts/objectstore/copy_files_to_irods.py
```diff
@@ -0,0 +1,112 @@
...
+        print("cursor.fetchall() completed")
+        for num, row in enumerate(table_data):
+            print("row: ", row)
+            print("\n")
+            uuid = row[12]
+            objectid = row[0]
+            object_store_id = row[11]
+            print("UUID: ", uuid)
+            print("\n")
+            uuid_with_dash = uuid[0:8] + "-" + uuid[8:12] + "-" + uuid[12:16] + "-" + uuid[16:20] + "-" + uuid[20:]
```
natefoo: You can do this in a more standard way with [the uuid lib](https://docs.python.org/3/library/uuid.html) and `uuid_with_dash = str(uuid.UUID(uuid_without_dash))`

scripts/objectstore/copy_files_to_irods.py
```diff
@@ -0,0 +1,112 @@
...
+            print("\n")
+            print("galaxy_dir:", galaxy_dir)
+            print("\n")
+
+            disk_sub_folder = os.path.join(*directory_hash_id(objectid))
+            print("disk_sub_folder: ", disk_sub_folder)
+            irods_sub_folder = os.path.join(*directory_hash_id(uuid_with_dash))
+            print("irods_sub_folder: ", irods_sub_folder)
+
+            disk_file_path = galaxy_dir + "/" + object_store_path + "/" + disk_sub_folder + "/" + "dataset_" + str(objectid) + ".dat"
```
natefoo: In production instances (including Test and Main) these paths are not necessarily under `galaxy_dir`. It should be ok to just pass the full path in `object_store_info`. [On Test](https://github.com/galaxyproject/usegalaxy-playbook/blob/master/env/test/templates/galaxy/config/object_store_conf.xml.j2) this will be:

```python
object_store_info = {'corral': '/galaxy-repl/test/files'}
```

And [on Main](https://github.com/galaxyproject/usegalaxy-playbook/blob/master/env/main/files/galaxy/config/object_store_conf.xml) it'll be:

```python
object_store_info = {
    'corral-repl': '/galaxy-repl/test/files',
    'corral-repl-psu': '/galaxy-repl/main/psufiles'
}

```

You can use `os.path.join()` here as well instead of string concatenation.

scripts/objectstore/copy_files_to_irods.py
```diff
@@ -0,0 +1,112 @@
...
+    objectid = None
+    object_store_id = None
+    disk_sub_folder = None
+    disk_file_path = None
+    irods_sub_folder = None
+    irods_file_path = None
+    icommand_mkdir = None
+    icommand_put = None
+
+    sql_statement = "SELECT * FROM dataset WHERE state = \'ok\' AND deleted = False AND purged = False AND id = {};".format(dataset_id)
```
natefoo: I think it'd be quicker to do one big select outside the loop and then pass the results in to the loop. Something like:

```sql
SELECT id, object_store_id, uuid
FROM dataset
WHERE state = 'ok'
    AND NOT deleted
    AND NOT purged
    AND id >= {start_dataset_id}
    AND id <= {end_dataset_id}
    AND object_store_id in {object_store_info.keys()}
```

scripts/objectstore/copy_files_to_irods.py
```diff
@@ -0,0 +1,112 @@
...
+        )
+        cursor.execute(sql_object)
+        print("cursor.execute() completed")
+
+        table_data = cursor.fetchall()
+        print("cursor.fetchall() completed")
+        for num, row in enumerate(table_data):
+            print("row: ", row)
+            print("\n")
+            uuid = row[12]
```
natefoo: Column ordering isn't fixed on instances that ran schema migrations over time, [psycopg2.extras.namedtuple](https://www.psycopg.org/docs/extras.html#namedtuple-cursor) is a nice way to deal with this, but I would also explicitly list the columns you need in the `SELECT` rather than `SELECT *`.

scripts/objectstore/copy_files_to_irods.py
```diff
@@ -0,0 +1,112 @@
...
+            print("irods_file_path: ", irods_file_path)
+
+            icommand_mkdir = "imkdir -p " + irods_sub_folder
+            print("icommand_mkdir: ", icommand_mkdir)
+            icommand_put = "iput " + disk_file_path + " " + irods_file_path
+            print("icommand_put: ", icommand_put)
+
+            subprocess.run([icommand_mkdir])
+            time.sleep(1)
+            subprocess.run([icommand_put])
```
natefoo: Two forks per dataset is probably going to be pretty expensive. This seems like a good case to use python-irodsclient. In addition, since all these operations will happen in rapid succession we can probably get away with creating a single iRODS session, which will avoid TCP connection overhead (and give us some insight on whether that's feasible for batch operations in Galaxy).

natefoo (commented):
Good start, thanks for putting this up for feedback!

I left a number of comments inline, but have a few additional ones:

1. You'll also need to check for the existence of a `dataset_<objectid>_files/`directory next to each `dataset_<objectid>.dat` and upload it and its contents to iRODS if it exists. Only some datasets will have this and it should not be created in iRODS if it doesn't exist on disk.
2. I can't recall - does iRODS checksum objects when uploaded or does it do it asynchronously? Although it will be pretty expensive we should maybe checksum them locally prior to upload to iRODS and then compare that against iRODS' checksum after uploading. We could also do this as a second "verification" pass after the data are uploaded to iRODS.
3. Once the dataset in iRODS is verified, we'll need to update the `object_store_id` and delete the on-disk copy. This probably makes sense as a second pass (with the verification in step 2)?

kxk302:
> Good start, thanks for putting this up for feedback!
> 
> I left a number of comments inline, but have a few additional ones:
> 
> 1. You'll also need to check for the existence of a `dataset_<objectid>_files/`directory next to each `dataset_<objectid>.dat` and upload it and its contents to iRODS if it exists. Only some datasets will have this and it should not be created in iRODS if it doesn't exist on disk.
> 2. I can't recall - does iRODS checksum objects when uploaded or does it do it asynchronously? Although it will be pretty expensive we should maybe checksum them locally prior to upload to iRODS and then compare that against iRODS' checksum after uploading. We could also do this as a second "verification" pass after the data are uploaded to iRODS.
> 3. Once the dataset in iRODS is verified, we'll need to update the `object_store_id` and delete the on-disk copy. This probably makes sense as a second pass (with the verification in step 2)?

1. Yes. I figured the need for this when running fastqc.
2. Have to look into this.
3. Agree

scripts/objectstore/copy_files_to_irods.py
kxk302: > You can do this in a more standard way with [the uuid lib](https://docs.python.org/3/library/uuid.html) and `uuid_with_dash = str(uuid.UUID(uuid_without_dash))`

Great. Will do.

scripts/objectstore/copy_files_to_irods.py
kxk302: > In production instances (including Test and Main) these paths are not necessarily under `galaxy_dir`. It should be ok to just pass the full path in `object_store_info`. [On Test](https://github.com/galaxyproject/usegalaxy-playbook/blob/master/env/test/templates/galaxy/config/object_store_conf.xml.j2) this will be:
> 
> ```python
> object_store_info = {'corral': '/galaxy-repl/test/files'}
> ```
> 
> And [on Main](https://github.com/galaxyproject/usegalaxy-playbook/blob/master/env/main/files/galaxy/config/object_store_conf.xml) it'll be:
> 
> ```python
> object_store_info = {
>     'corral-repl': '/galaxy-repl/test/files',
>     'corral-repl-psu': '/galaxy-repl/main/psufiles'
> }
> ```
> 
> You can use `os.path.join()` here as well instead of string concatenation.

Ok. Will revise accordingly.

scripts/objectstore/copy_files_to_irods.py
kxk302: Right, that is one SQL statement instead of many and should be faster.

scripts/objectstore/copy_files_to_irods.py
kxk302: Sure

scripts/objectstore/copy_files_to_irods.py
kxk302: Sure, that was its all Python code, which is even better.

scripts/objectstore/copy_files_to_irods.py
kxk302: I think if i specify the columns that I need in the SELECT statement, I should be good here.

scripts/objectstore/copy_files_to_irods.py
```diff
@@ -0,0 +1,112 @@
...
+            password=connection_info["password"]
+        )
+        conn.cursor()
+
+    except Exception as e:
+        print(e)
+        return
+
+    osi_keys = list(object_store_info.keys())
+    osi_keys_quoted = ["'" + key + "'" for key in osi_keys]
```
natefoo: You actually don't need to worry about quoting these yourself, you can pass lists as arguments to a psycopg2 cursor's `execute()` method and it will properly quote for you. e.g.:

```python
sql = """SELECT id, object_store_id, uuid
                       FROM dataset
                       WHERE state = %s
                       AND NOT deleted
                       AND NOT purged
                       AND id >= %s
                       AND id <= %s
                       AND object_store_id IN %s
"""
args = ('ok', start_dataset_id, end_dataset_id, osi_keys)
print(cursor.mogrify(sql, args))
cursor.execute(sql, args)
```

kxk302: Revised this accordingly.

kxk302:
I revised the script so now 1) irods calculates the checksum for each file added, 2) checksum of local file is calculated as well, and 3) after a file is added to irods, we get the file (and its checksum) and compare it to local file's checksum.

scripts/objectstore/copy_files_to_irods.py
```diff
@@ -140,9 +143,31 @@ def copy_files_to_irods(start_dataset_id, end_dataset_id, object_store_info_file
...
 
             # Add disk file to collection
-            options = {kw.DEST_RESC_NAME_KW: 'demoResc'}
+            options = {kw.DEST_RESC_NAME_KW: 'demoResc', kw.REG_CHKSUM_KW : ''}
             session.data_objects.put(disk_file_path, irods_file_path, **options)
 
+            # Calculate disk file checksum before uploading it to irods
+            # After disk file is uploaded to irods, we get the file checksum from irods and compare it with the calculated disk file checksum
+            # Note that disk file checksum is ASCII, whereas irods file checksum is Unicode
+            disk_file_checksum = get_file_checksum(disk_file_path)
```
bgruening: @kxk302 we once discussed to calculate for all files a checksum in Galaxy and store this as part of every dataset metadata in Galaxy. This will enable many interesting use-cases maybe its really time to get this done, so you do not need to calculate it here.

kxk302: @bgruening the checksum calculation is automatically done by irods (You just need to set a configuration flag). what I'm doing here is calculating the checksum outside of irods, so I can compare it to irods checksum, as test for upload to irods being valid. The upload process in this script is a one time thing, just to move existing disk files into irods, an won't be used afterwards.

kxk302:
Ok, I merged the scripts. copy_files_to_irods.py now can copy files to irods or verify checksum/delete disk file depending on a new input argument. I'll delete verify_irods_checksum.py later.

The revised script calculates checksum when adding files to irods for all the files in the extra directory (e.g., generated by tools like fastqc). The verification also verifies the checksum of all files in the extra directory (it recursively traverses the directory and verifies the checksum for all files).

@Nate I think we can test this tomorrow on Test.

kxk302:
Ok, I cleaned up the script (removed debug print statements), added documentation to help option, and tested the script locally. I have 3 files on disk (one of them is the output of fastqc and comes with an extra directory). Ran the script in copy mode. It copied all files (and the extra folder) to my irods instance running locally (checksums are calculated and kept in irods). I then ran the script in checksum mode. It calculated the disk file checksums and compared them to checksum value in irods. Since they matched, it deleted the file on disk and updated object store id (in dataset table) to irods object store id (if there is a checksum discrepancy, file delete and object store id update are not done). @natefoo, I think we can run this on Test for a small dataset Id range to test it. Are u available later today for this? Thanks.

kxk302:
I modified the script so that the irods resource can be selected based on the last time a dataset was used as input/output in a job. That is compared to a cuttoff date, and if the last access time is prior to the cutoff date it goes to a tape resource, otherwise to a regular (non-tape) resource.

kxk302:
Removed the draft status.

scripts/objectstore/copy_files_to_irods.py
```diff
@@ -0,0 +1,366 @@
...
+    disk_file_checksum = subprocess.check_output(checksum_cmd, shell=True)
+    # remove '\n' from the end of disk_file_checksum
+    disk_file_checksum_len = len(disk_file_checksum)
+    disk_file_checksum_trimmed = disk_file_checksum[0:(disk_file_checksum_len - 1)]
+    # Return Unicode string
+    return disk_file_checksum_trimmed.decode("utf-8")
+
+
+def print_help_msg():
+    print("\nThis script copies datasets from Disk to Irods. It operates on a range of dataset IDs (start_dataset_id and end_dtatset_id, both inclusive).")
```
mvdbeek: Could you make that a multi-line string please ?
something like
```
help = """
Line 1 and 
Line 2 and so
on
"""
```

scripts/objectstore/copy_files_to_irods.py
```diff
@@ -0,0 +1,366 @@
...
+if __name__ == '__main__':
+    start_dataset_id = None
+    end_dataset_id = None
+    object_store_info_file = None
+    irods_info_file = None
+    db_connection_info_file = None
+    copy_or_checksum = None
+
+    try:
+        opts, args = getopt.getopt(sys.argv[1:], "hs:e:o:i:d:c:", ["start_dataset_id=", "end_dataset_id=", "object_store_info_file=", "irods_info_file=", "db_connection_info_file=", "copy_or_checksum="])
```
mvdbeek: Can you use argparse here for a friendlier interface ?

mvdbeek:
I've cleared the milestone here, since it's an adjacent script there's no need to have this in a particular Galaxy version IMO.

scripts/objectstore/copy_files_to_irods.py
kxk302: Sure.

scripts/objectstore/copy_files_to_irods.py
kxk302: Sure.

kxk302:
@mvdbeek Do you think this PR can be merged?

kxk302:
BTW, I just pushed a fix for lint issue (had to re-order a couple of imports).

kxk302:
Hmm, the import issues still persist. Need to look into it.

kxk302:
Ok, I think I finally fixed the import order issue. Seems VS Code was saving the imports in the wrong order. For the time being, I manually fixed the order, until I figure out why VSCode is acting up.

kxk302:
@mvdbeek The import errors are fixed and there is only a Selenium test issue. Would u be able to merge this? Thx

PR #10129, opened 2020-08-19
[20.05] Open pages API a bit to allow anonymous users to view published pages.

dannon:
@tnabtaf Ran into this while trying to view a published page in an incognito window.  Accessibility checks are enforced on on page access for logged in users (or not), so there's no change in actual content access here.


mvdbeek (approved):

mvdbeek:
We also have tests confirming we still get a 403, in case anyone is wondering https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy_test/api/test_pages.py#L143

PR #10134, opened 2020-08-20
Markdown Editor Revision

guerler:
This PR revises the Markdown Editor by adding a toolbar which allows users to select embeddable items from a dropdown menu. The toolbar is consistently available in the workflow editor, page editor and invocation reports editor. Following the selection of an item in the workflow editor, users are presented with a list of either output labels or step label, depending on the item inserted. Additionally users may select item ids using the data dialog view. The data dialog view has been generalized to be re-usable for jobs and invocations.

image

Example of Label selector:
image


bgruening:
@guerler this is a very cool enhancement!

I have a few comments collected when I tried it:

* indicate that the user is in the workflow-editor (worfkflow name :: Workflow editor) in the title?
* I would still offer the help sidebar (optional); question mark in the panel on the right side?
* if you pick a dataset-option for example
    * "select dataset from history" should not be the default imho; I never considered this as a real-use case? Why should I include a dataset that is not part of my workflow? Those datasets can go away before the workflow is finished etc.
    * "Select Output label" --> can we add here a big hint, how this list is generated and how a user needs to add labels (maybe with a link back to the workflow)
    * "Select Output label" could be some fancy workflow-dataset picker, but this is a different feature
    * it would be nice if the clickable area to select a radio-box would be larger maybe the entire column
    * if you open that dialog fresh, the radio-buttons are not keyboard-navigatable, maybe the focus should be on the radio boxes by default
* workflow_display --> should this one not offer a default of the current executed workflow, display another workflow is probably a very niche use-case
* is the light-blue background intended?

@guerler, do you think its useful to make the output label, editable directly from the workflow box, or is this just confusing?

jmchilton:
> workflow_display --> should this one not offer a default of the current executed workflow, display another workflow is probably a very niche use-case

I haven't tried it - but this really shouldn't even work 😦 - the workflow report markdown be design should only be able to reference objects from the workflow. Otherwise nothing is portable, etc...

guerler:
> > workflow_display --> should this one not offer a default of the current executed workflow, display another workflow is probably a very niche use-case
> 
> I haven't tried it - but this really shouldn't even work 😦 - the workflow report markdown be design should only be able to reference objects from the workflow. Otherwise nothing is portable, etc...

Yes, I agree. Otherwise exporting workflows would lead to issues, this has been resolved.

guerler:
Followed-up on @enis great suggestion to use the side panel instead of using a separate toolbar. Thank you. Currently only functional in Workflow editor. Making adjustments now to fix the views for Pages.

bgruening:
* inserting an element in the middle for example puts the focus to the end of the document, which is confusing with long documents
* `from_history` as the label is not needed I think
* maybe have the categories expanded by default?
* bring the old help-panel back?
* closing the report page yields into a broken workflow (noodles removed)

guerler:
Thank you for so much for your detailed comments @bgruening. I am still working through some of the issues and testing the existing changeset but this is what I have so far:

1) inserting an element in the middle for example puts the focus to the end of the document, which is confusing with long documents

Has been fixed.

2) from_history as the label is not needed I think

Has been removed.

3) maybe have the categories expanded by default?

Categories are now expanded.

4) bring the old help-panel back?

I added the help button back to the toolbar.

5) closing the report page yields into a broken workflow (noodles removed)

Has been fixed.

6) indicate that the user is in the workflow-editor (worfkflow name :: Workflow editor) in the title?

Has been fixed.

7) "Select Output label" --> can we add here a big hint, how this list is generated and how a user needs to add labels (maybe with a link back to the workflow)

I added a more descriptive description to the selection modal.

8) "Select Output label" could be some fancy workflow-dataset picker, but this is a different feature

True, we could consider to select all datasets of a workflow and allow users to add/edit labels directly in the reports view. I think this should be a separate PR, since it probably requires a more significant amount of commits to achieve that.

9) it would be nice if the clickable area to select a radio-box would be larger maybe the entire column

I noticed this and I would like to revisit this generally so the changes would also apply to radiobuttons used elsewhere e.g. in the tool form.

10) if you open that dialog fresh, the radio-buttons are not keyboard-navigatable, maybe the focus should be on the radio boxes by default

The focus is on the radio boxes but keyboard navigation still seems to be not functional. I am not entirely sure yet if this is a general issue or if it is related to the fact that the radiobuttons are displayed in the modal. This should be investigated together with 9 and can hopefully be applied such that other locations using radio buttons can benefit from it.

11) workflow_display --> should this one not offer a default of the current executed workflow, display another workflow is probably a very niche use-case

This has been fixed. In the workflow editor only the current workflow is inserted.

12) is the light-blue background intended?

Yes, do you think a different color would be better?

bgruening:
Thanks for working all of this. This is getting super smooth :+1: 

> Yes, do you think a different color would be better?

It is confusing and feels "dirty". Why not just white?

client/src/components/WorkflowInvocationState/WorkflowInvocationState.vue
```diff
@@ -125,7 +125,7 @@ export default {
             return this.invocationSchedulingTerminal && this.jobStatesTerminal;
         },
         invocationLink: function () {
-            return getUrl(`workflows/invocations/report?id=${this.invocationId}`);
+            return getUrl(`page/edit_report?id=${this.invocationId}`);
```
mvdbeek, line 128: ~~Is that correct ?~~
I guess that's a bit nitpicky, but I prefer `workflows/invocations/report?id=${this.invocationId}` over `page/edit_report?id=${this.invocationId}`, the former I read as give me the invocation report for this invocation id, while I read the latter as `let me edit the page report for this invocation id`, which feels odd ?

mvdbeek: I also don't seem to be able to edit or create a page from an invocation, that would be pretty cool.

mvdbeek:
OK, the error is not specific to the invocation report :). I think we can merge this and iterate on it.

jmchilton:
I didn't find time to review - so I reserve the right to complain in the future 😆. I'll try to test it out in the next couple days here. But the screenshots look amazing! Thanks so much @guerler

guerler:
@mvdbeek there was some confusion about the ability of editing invocation reports which has been clarified now. I will remove the edit reports mako in a follow-up before the freeze. Thanks for merging this PR in a timely fashion.

PR #10135, opened 2020-08-20
Fix parameter_input dropped in new workflow form

mvdbeek:
Fixes https://github.com/galaxyproject/galaxy/pull/9151#issuecomment-677635591

### DIFF
--- a/client/src/components/Workflow/Run/WorkflowRunFormSimple.vue
+++ b/client/src/components/Workflow/Run/WorkflowRunFormSimple.vue
@@ -5,7 +5,7 @@
 <script>
 import Form from "mvc/form/form-view";
 import { invokeWorkflow } from "./services";
-import isWorkflowInput from "components/Workflow/constants";
+import { isWorkflowInput } from "components/Workflow/constants";
 
 export default {
     props: {
--- a/client/src/components/Workflow/Run/WorkflowRunFormSimple.vue
+++ b/client/src/components/Workflow/Run/WorkflowRunFormSimple.vue
@@ -5,6 +5,7 @@
 <script>
 import Form from "mvc/form/form-view";
 import { invokeWorkflow } from "./services";
+import isWorkflowInput from "components/Workflow/constants";
 
 export default {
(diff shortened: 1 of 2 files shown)


jmchilton:
Sorry about this - thanks for the fix.

client/src/components/Workflow/Run/WorkflowRunFormSimple.vue
```diff
@@ -5,6 +5,7 @@
 <script>
 import Form from "mvc/form/form-view";
 import { invokeWorkflow } from "./services";
+import isWorkflowInput from "components/Workflow/constants";
```
dannon:
```suggestion
import { isWorkflowInput } from "components/Workflow/constants";
```

client/src/components/Workflow/constants.js
```diff
@@ -0,0 +1,7 @@
+const WorkflowInputs = ["data_input", "data_collection_input", "parameter_input"];
+
+function isWorkflowInput(stepType) {
+    return WorkflowInputs.includes(stepType);
+}
+
+export { isWorkflowInput as default };
```
dannon:
```suggestion
const WorkflowInputs = ["data_input", "data_collection_input", "parameter_input"];

export function isWorkflowInput(stepType) {
    return WorkflowInputs.includes(stepType);
}
```

client/src/components/Workflow/Run/WorkflowRunFormSimple.vue
dannon: Not sure I grok `constants` either -- what else goes in here?  Might be more `util`?

mvdbeek: Yeah, I initially just had the array in there, but thought a function might be nicer. I can make this util, or, considering we're not re-suing this yet anywhere, I could also add this straight to WorkflowRunFormSimple.vue. Seems like we'd need this for invocation step displaying as well, but I can move this when we actually need it.

PR #10139, opened 2020-08-20
Add warning for non-existing tool version

### LINKED ISSUE #10035: Requesting a tool with a non-existent version does not show an error
E.g. the following URL:

https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/devteam/fastqc/fastqc/0.doesnotexist

Does not show any warning when it switches to the newer version of the tool. Could we popup a message similar to "This job was run with a version of the tool that isn't available, we've switched to ..."?

I think that might cause surprises for users of #10024 if galaxy behaves as if that version of the tool is there when it isn't, and they've been silently shown a different version than requested.

OlegZharkov:
Fixes https://github.com/galaxyproject/galaxy/issues/10035

image

Thanks a lot to @hexylena for reporting!

### DIFF
--- a/client/src/mvc/tool/tool-form.js
+++ b/client/src/mvc/tool/tool-form.js
@@ -53,7 +53,10 @@ const View = Backbone.View.extend({
                                 // we can rely on tool_id, only if it's installed from toolshed
                                 // if no version provided and tool_id is not in toolshed style, don't show the warning
                                 const client_version =
-                                    options.version || (options.tool_id.match(/^(toolshed)\./) && options.tool_id);
+                                    options.version ||
+                                    (options.tool_id && options.tool_id.match(/^(toolshed)\./)
+                                        ? options.tool_id
+                                        : false);
 
                                 if (
                                     data.versions &&
--- a/client/src/mvc/tool/tool-form.js
+++ b/client/src/mvc/tool/tool-form.js
@@ -51,9 +51,13 @@ const View = Backbone.View.extend({
                                     return;
                                 }
 
-                                const user_version = (options.version || options.tool_id).replace(/.*\//, "");
+                                const client_version = options.version || options.tool_id;
 
-                                if (data.versions && !data.versions.includes(user_version)) {
+                                if (
+                                    data.versions &&
+                                    client_version &&
+                                    !data.versions.includes(client_version.replace(/.*\//, ""))
+                                ) {
                                     Toast.warning(
                                         `Specified version is not found! Using ${data.version} instead`,
                                         "Error",
--- a/client/src/mvc/tool/tool-form.js
+++ b/client/src/mvc/tool/tool-form.js
@@ -52,7 +52,7 @@ const View = Backbone.View.extend({
                                 }
 
                                 const user_version = options.id.replace(/.*\//, "");
-                                if (!data.versions.includes(user_version)) {
+                                if (data.versions && !data.versions.includes(user_version)) {
                                     Toast.warning(
                                         `Specified version is not found! Using ${data.version} instead`,
                                         "Error",
--- a/client/src/mvc/tool/tool-form.js
+++ b/client/src/mvc/tool/tool-form.js
@@ -51,7 +51,8 @@ const View = Backbone.View.extend({
                                     return;
                                 }
 
-                                const user_version = options.id.replace(/.*\//, "");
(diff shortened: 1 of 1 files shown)


bgruening:
@OlegZharkov can you reformat the client please?

client/src/mvc/tool/tool-form.js
```diff
@@ -49,6 +50,20 @@ const View = Backbone.View.extend({
                                     window.location = getAppRoot();
                                     return;
                                 }
+
+                                const user_version = options.id.replace(/.*\//, "");
+                                if (!data.versions.includes(user_version)) {
+                                    Toast.error(
```
dannon:
```suggestion
                                    Toast.warning(
```

dannon: Since there was actually an alternative version, let's not throw the error Toast, but a warning?

OlegZharkov: sure

OlegZharkov:
> Test failure looks legit; `data.versions` is undefined?

you're right. Somehow, I cannot reproduce it locally.
When I run ```yarn run test```
I am getting 
```
02 09 2020 17:48:58.259:ERROR [karma-server]: Error: Found 1 load error
    at Server.<anonymous> (/home/oleg/galaxy/client/node_modules/karma/lib/server.js:189:27)
    at Object.onceWrapper (events.js:421:28)
    at Server.emit (events.js:327:22)
    at emitListeningNT (net.js:1347:10)
    at processTicksAndRejections (internal/process/task_queues.js:83:21)
```

Weird, I should probably investigate

OlegZharkov:
> Test failure looks legit; `data.versions` is undefined?

Anyway, this PR passes it now. Thanks a lot for the lead!

hexylena:
> Thanks a lot to @hexylena for reporting!

Thanks for implementing @OlegZharkov this looks great!

mvdbeek:
I think the selenium errors might be relevant here ?
last

OlegZharkov:
> I think the selenium errors might be relevant here ?
you're right. sorry, should be better now

client/src/mvc/tool/tool-form.js
```diff
@@ -49,6 +50,25 @@ const View = Backbone.View.extend({
                                     window.location = getAppRoot();
                                     return;
                                 }
+
+                                const client_version = options.version || options.tool_id;
+
+                                if (
+                                    data.versions &&
+                                    client_version &&
+                                    !data.versions.includes(client_version.replace(/.*\//, ""))
```
mvdbeek, line 64: is this valid syntax ? wouldn't this need to be
```suggestion
                                    !data.versions.includes(client_version.replace("/.*\//", ""))
```

mvdbeek: I also don't think we'd want to display this everywhere that we load a tool ~~model~~ form. This should probably be a warning on the tool form, which is the only place where a user could do anything about this. If multiple tool forms are open this is also going to trigger toasts all over the place. I don't know how exactly we're handling this in backbone, but in Vue you'd emit an event and then bind an error handler to the event that renders a warning in the correct place.

dannon: It's valid, javascript regexes can be defined inline like that.

dannon: Whoops, didn't see the other comment when I responded, only the bit about the syntax.  I agree with @mvdbeek about how this should be displayed; toasts aren't a great fit here.

client/src/mvc/tool/tool-form.js
```diff
@@ -49,6 +50,25 @@ const View = Backbone.View.extend({
...
+                                ) {
+                                    Toast.warning(
```
guerler: You can replace this Toast block with: `data.message = "Requested version unavailable.";` This will then display the warning in the form's `alert` element (after `form.model.set(data);` in the line below has been executed).

image

OlegZharkov: @guerler 
that's a very very good advise. thanks!

OlegZharkov:
thanks to @guerler, Toast is removed. 
image

nsoranzo:
> thanks to @guerler, Toast is removed.

Can you please update the PR title?

OlegZharkov:
Thanks @mvdbeek, @dannon, @guerler

PR #10141, opened 2020-08-21
Test setting of all config values; uniformly process config files

### LINKED ISSUE #9973: The path for config files doesn't be resolved with respect to <config_dir>
I have this configuration
```
# The directory that will be prepended to relative paths in options
# specifying other Galaxy config files (e.g. datatypes_config_file).
# Defaults to the directory in which galaxy.yml is located.
config_dir: /opt/galaxy/config
	
# Optional configuration file similar to `job_config_file` to specify
# which Galaxy processes should schedule workflows.
# The value of this option will be resolved with respect to
# <config_dir>.
workflow_schedulers_config_file: workflow_schedulers_conf.xml

# Location of the configuration file containing extra user
# preferences.
# The value of this option will be resolved with respect to
# <config_dir>.
user_preferences_extra_conf_path: user_preferences_extra_conf.yml
```
and so I expect Galaxy looks for the files into this path: 
`/opt/galaxy/config/` 
but instead, I get this message in the log

```
Jul 08 22:20:29 test.internal.usegalaxy.eu.novalocal python[1367818]: galaxy.workflow.scheduling_manager INFO 2020-07-08 22:20:29,244 
Cannot find workflow schedulers plugin config file '/opt/galaxy/server/workflow_schedulers_conf.xml', using default scheduler.
...
Jul 08 22:50:15 test.internal.usegalaxy.eu.novalocal uwsgi[1383859]: WARNING:galaxy.config:Config 
file (/opt/galaxy/server/user_preferences_extra_conf.yml) could not be found or is malformed.

```

jdavcs:
[This is close to done.]

Fixes #9973 (here's the description of a specific bug: https://github.com/galaxyproject/galaxy/issues/9973#issuecomment-655792443)

**Overall problem.** We had tests (both integration and unit) for testing *default* values of configuration options. We did *not* test overriding those defaults. Which is why we were not catching at least one type of bugs (9973 shows one that surfaced). The issues are due to non-uniform processing of config options: we load them uniformly ([using defaults from the schema, adding some processing steps, etc.](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L114)). But then we reload some of them in `parse_config_file_options` - which overrides not only the schema defaults, but also any processing steps previously executed. User-set relative config file paths being resolved with respect to the root directory instead of their correct location is one example of a bug. Not being able to set any of the options that pass through `parse_config_file_options` to `None` is another unintended consequence (currently, the user *can* explicitly set all other config options to `None`; if we want to disallow this for these options, we can do that explicitly, and not as an implicit side-effect).

**What this PR contains:** 
- Tests for loading config option defaults (326 out of 329) + overriding defaults (302 out of 329). Tests for default values have been partially rewritten (they should be simpler now, much easier to read)
- Changes to config module, config schema + minor edits to a few other modules (those that made use of `app.config.foo_set`: this has been replaced with `app.config.is_set('foo')`)
- Minor edits to the testing infrastructure enabling testing of these options (mostly different handling of loading values into the mock config)
The config schema has been changed to make use of `path_resolves_to` instead of hardcoding prefixes (`config/`, `data/`, etc.).

The PR consists of many commits - that's intentional. Each commit is focused on one change; the commit message contains details about what was done and why. The "Refactor processing of ..." commit messages are not all the same.

This PR will be ready for final review in a few more commits: my goal is to remove `parse_config_file_options` and enable tests for the few remaining config options. Meanwhile, any feedback is, as always, very much appreciated.

When this is done, I'll write a "How-to" entry on configuration edits for the documentation.

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -676,7 +676,7 @@ class GalaxyAppConfiguration(BaseAppConfiguration, CommonConfigurationMixin):
             with open(self.user_preferences_extra_conf_path) as stream:
                 self.user_preferences_extra = yaml.safe_load(stream)
         except Exception:
-            if self.user_preferences_extra_conf_path_set:
+            if self.is_set('user_preferences_extra_conf_path'):
                 log.warning('Config file (%s) could not be found or is malformed.' % self.user_preferences_extra_conf_path)
             self.user_preferences_extra = {'preferences': {}}
 
@@ -797,7 +797,6 @@ class GalaxyAppConfiguration(BaseAppConfiguration, CommonConfigurationMixin):
             shed_tool_config_file=[self._in_managed_config_dir('shed_tool_conf.xml')],
             shed_tool_data_table_config=[self._in_managed_config_dir('shed_tool_data_table_conf.xml')],
             tool_destinations_config_file=[self._in_config_dir('tool_destinations.yml')],
-            user_preferences_extra_conf_path=[self._in_config_dir('user_preferences_extra_conf.yml')],
             workflow_resource_params_file=[self._in_config_dir('workflow_resource_params_conf.xml')],
             workflow_schedulers_config_file=[self._in_config_dir('workflow_schedulers_conf.xml')],
             # self.file_path set to self._in_data_dir('objects') by schema
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -797,8 +797,6 @@ class GalaxyAppConfiguration(BaseAppConfiguration, CommonConfigurationMixin):
             shed_tool_config_file=[self._in_managed_config_dir('shed_tool_conf.xml')],
             shed_tool_data_table_config=[self._in_managed_config_dir('shed_tool_data_table_conf.xml')],
             tool_destinations_config_file=[self._in_config_dir('tool_destinations.yml')],
-            tool_sheds_config_file=[self._in_config_dir('tool_sheds_conf.xml')],
-            trs_servers_config_file=[self._in_config_dir('trs_servers_conf.yml')],
             user_preferences_extra_conf_path=[self._in_config_dir('user_preferences_extra_conf.yml')],
             workflow_resource_params_file=[self._in_config_dir('workflow_resource_params_conf.xml')],
             workflow_schedulers_config_file=[self._in_config_dir('workflow_schedulers_conf.xml')],
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -793,7 +793,6 @@ class GalaxyAppConfiguration(BaseAppConfiguration, CommonConfigurationMixin):
             job_metrics_config_file=[self._in_config_dir('job_metrics_conf.xml'), self._in_sample_dir('job_metrics_conf.xml.sample')],
             local_conda_mapping_file=[self._in_config_dir('local_conda_mapping.yml')],
             modules_mapping_files=[self._in_config_dir('environment_modules_mapping.yml')],
-            shed_tool_config_file=[self._in_managed_config_dir('shed_tool_conf.xml')],
             shed_tool_data_table_config=[self._in_managed_config_dir('shed_tool_data_table_conf.xml')],
             tool_destinations_config_file=[self._in_config_dir('tool_destinations.yml')],
             # self.file_path set to self._in_data_dir('objects') by schema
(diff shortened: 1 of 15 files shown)


test/integration/test_config_defaults.py
```diff
@@ -161,7 +161,7 @@ def expected_default_config_dir(value):
     'conda_auto_init',  # broken: default overridden
     'config_dir',  # value overridden for testing
     'data_dir',  # value overridden for testing
-    'data_manager_config_file',  # broken: remove 'config/' prefix from schema
+    'data_manager_config_file',  # value overidden for testing
```
jmchilton, line 166: Can this whole test go? Are there comparable unit tests at this point?

Having to explicitly place everything we override in here is sort of gross and unexpected - also it clashes with GALAXY_CONFIG_OVERRIDE_ and such for ad hoc testing - which I guess could break all the integration tests, but yeah...

lib/galaxy/config/__init__.py
```diff
@@ -132,6 +133,14 @@ def _process_renamed_options(self, kwargs):
                 if old in kwargs and new not in kwargs:
                     kwargs[new] = kwargs[old]
 
+    def is_set(self, key):
+        """Check if a configuration option has been explicitly set."""
+        # NOTE: This will check all supplied keyword arguments, including those not in the schema.
+        # To check only schema options, change the line below to `if property not in self._raw_config:`
+        if property not in self._kwargs:
```
```suggestion
        if key not in self._kwargs:
```

lib/galaxy/config/__init__.py
```diff
@@ -132,6 +133,14 @@ def _process_renamed_options(self, kwargs):
...
+            log.warning("Configuration option does not exist: '%s'" % key)
```
nsoranzo, line 141: Is this needed or is just for debug?

lib/galaxy/config/__init__.py
jdavcs: @nsoranzo oh my! thank you for catching this!

lib/galaxy/config/__init__.py
jdavcs: Definitely not for debugging. As for needed or not, we had a discussion with @mvdbeek a while ago: 
https://github.com/galaxyproject/galaxy/pull/8867#pullrequestreview-306394120
My preference is to err on the side of caution, and I'd even raise an exception (see discussion^). However, if it safe to log a warning or even ignore this, I don't mind. @natefoo - `config.is_set('foo')`replaces `config.foo_set` - should we not guard/guard/guard aggressively against a case where `foo` is misspelled by a dev?

test/integration/test_config_defaults.py
jdavcs: Yes, it could go. If you think it doesn't serve its purpose or causes more problems than it solves, I don't mind dropping it (I'd still test the schema defaults though). That said, a few arguments:
1. It's for config options defined in the schema *only*. It's not for everything we override (or create) in the config module.
2. We load these options automatically (which is tested), but we do not test overriding them. The logic for overriding options varies, and in some cases we were doing it wrong (`parse_config_file _options`, as per #9973). This test should prevent such bugs.
3. At least in theory, this unit test should not affect any integration or other unit tests.

That said, I do agree it's ugly. I tried generalizing it and setting new values automatically based on data type, but there are  too many special cases - I couldn't come up with an elegant solution. So it's not elegant. If the test also feels unexpected, that's worse (I can't judge that myself - I'm too "used to it").

To sum up, if you still feel it's better dropped, I'll drop it. In that case, I think I'll write a few tests targeting specific individual options (like the various config files).

lib/galaxy/config/__init__.py
nsoranzo: It could have made sense in #8867 to log this message when you were checking if the key is in `self._raw_config`, but now you are checking if it's in `self._kwargs` and it doesn't make sense to log if it's just not set by the admin.

jdavcs: Another whoops. It should be checking `self._raw_config`, of course! Thank you. I've accidentally overwritten your previous fix (key > property), added it w/this fix.

test/integration/test_config_defaults.py
jdavcs: I'll remove the tests generated for overriding schema defaults before opening this for review. Keeping them for now while I ensure all other tests pass (they do pass locally).

ic4f:
Al tests pass locally. Investigating...

ic4f:
Closing. Replaced with #10172

PR #10143, opened 2020-08-22
[20.05] Update toolshed/ie js dist locations

### LINKED ISSUE #10142: Toolshed js ?
Executive summary: Running a local toolshed in "out of the box" mode seems broken in dev at present. Quick and dirty fix:
make static/scripts/bundled and soft link static/dist/* static/scripts/bundled/ works - in case anyone needs it.

Longer version:

Looks like default toolshed base javascript path needs updating.

Main toolshed seems to fill page script src templates like 
```
<script src="/static/scripts/bundled/libs.chunk.js?v=1598068810" type="text/javascript"></script>
<script src="/static/scripts/bundled/base.chunk.js?v=1598068810" type="text/javascript"></script>
```
and it works :)

A recent dev clone produces the same - but those paths do not exist after building the client - after all, galaxy currently uses
```
<script src="/static/dist/libs.chunk.js?v=1597905027" type="text/javascript"></script>
<script src="/static/dist/base.chunk.js?v=1597905027" type="text/javascript"></script>
<script src="/static/dist/analysis.bundled.js?v=1597905027" type="text/javascript"></script>
```

Am I missing something or could this do with some competent adjustment?

dannon:
Should fix #10142

### DIFF
--- a/config/plugins/interactive_environments/common/templates/ie.mako
+++ b/config/plugins/interactive_environments/common/templates/ie.mako
@@ -1,8 +1,8 @@
 <%def name="load_default_js()">
 ${h.css( 'base' ) }
-${h.js('bundled/libs.chunk',
-       'bundled/base.chunk',
-       'bundled/generic.bundled')}
+${h.dist_js('libs.chunk',
+            'base.chunk',
+            'generic.bundled')}
 </%def>
 
 <%def name="default_javascript_variables()">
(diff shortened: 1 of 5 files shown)


nsoranzo:
Should we remove the now unused `js` and `templates` functions from `lib/galaxy/web/framework/helpers/__init__.py` ?

dannon:
@nsoranzo Yeah.  At the time the concern was not dropping the compatibility layer for things in `config`, but those are probably busted without this anyway -- I should probably standardize it all back to just `js` and drop the dist-specific behavior at this point.

Issue #7121, opened 2018-12-17
uploading xml files with HTML content yields "The uploaded file contains invalid HTML content"


fbergmann:
Hello, 

in trying to bring our software (COPASI, http://copasi.org) so we can use it from Galaxy I've encountered the issue, that uploading our supported file formats fails with: 

```The uploaded file contains invalid HTML content```

these are either SBML files (http://sbml.org), or COPASI files. Each files that are pretty much XML, in which elements can be annotated using HTML notes. 

I've tried to use the galaxy sbml type for SBML, for COPASI both the 'xml' and the 'cps' datatypes. I've noticed this does happen not just in my own local instance, but also on the usegalaxy.eu server. 

This seems to happen in `galaxy.datatypes.sniff`: 

```https://docs.galaxyproject.org/en/latest/_modules/galaxy/datatypes/sniff.html```

Any help would be appreciated

nsoranzo:
@fbergmann If you can provide example SBML and COPASI files that would be helpful.

bgruening:
@fbergmann does this mean that if you upload files and select `sbml` that you get this error as well?

It seems we do not have an example file in our repo. Can you provide us with a tiny valid example file, please?

fbergmann:
Hello, of course, i should have attached some in the first place: in our repository under: 

https://github.com/copasi/COPASI/tree/develop/TestSuite/distribution

are test files (brusselator-model.xml SBML, brusselator.cps COPASI) that i would have expected to load without issues. 

@bgruening and yes, i manually selected `cps` and `sbml` and got this error.

nsoranzo:
For both files the issue seems to be the line `<meta name="qrichtext" content="1"/>` which is flagged by https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/util/checkers.py#L36

fbergmann:
for both formats, that tag (as well as the href which would have flagged)  are valid.

nsoranzo:
@fbergmann Sure, at the moment I'm just investigating where this error is coming from and documenting it here.

The original code for `check_html()` is from commit 3f5da01132f5ed507e2f8197a5e761087ec0f6c6 , it may be time for a rethink.

fbergmann:
I've upgraded to the latest dev version, and I am still having this issue.

fbergmann:
[...]

PR #10144, opened 2020-08-22
WIP: SBML upload fix - #7121

bgruening:
xref: https://github.com/galaxyproject/galaxy/issues/7121

This one needs discussion as it does not fix all problems with SBML :(

### DIFF
--- a/lib/galaxy/util/checkers.py
+++ b/lib/galaxy/util/checkers.py
@@ -16,13 +16,17 @@ from galaxy.util.image_util import image_type
 
 HTML_CHECK_LINES = 100
 CHUNK_SIZE = 2 ** 15  # 32Kb
-HTML_REGEXPS = (
-    re.compile(r"<A\s+[^>]*HREF[^>]+>", re.I),
-    re.compile(r"<IFRAME[^>]*>", re.I),
-    re.compile(r"<FRAMESET[^>]*>", re.I),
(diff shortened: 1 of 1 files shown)


lib/galaxy/util/checkers.py
```diff
@@ -16,13 +16,17 @@
...
+HTML_REGEXPS = {
+    "link": re.compile(r"<A\s+[^>]*HREF[^>]+>", re.I),
+    "iframe": re.compile(r"<IFRAME[^>]*>", re.I),
+    "frameset": re.compile(r"<FRAMESET[^>]*>", re.I),
+    "meta": re.compile(r"<META[\W][^>]*>", re.I),
+    "script": re.compile(r"<SCRIPT[^>]*>", re.I),
+}
+
+HTML_ALLOWLIST = {
+    "meta": re.compile(r'<meta name="qrichtext" content="1"\s*/>', re.I),
```
nsoranzo, line 28: If I google for "meta qrichtext" I see mainly stuff related to Qt QTextEdit, so maybe that's a red herring?

bgruening: The underlying problem is really that any HTML can be embedded. From all test files that I had, this at least made it possible to load 90% of all files. I hope someone else has an idea here, how we can solve this, or if you should simply support smbl.gz files only.

Issue #9977, opened 2020-07-09
TypeError: Object of type _Attrib is not JSON serializable


innovate-invent:
While trying to view data types in the admin panel I get an error 500 in the browser console and nothing is listed.
I am not sure if this is a bug or if I have misconfigured Galaxy. If I go to change the datatype of a dataset, it lists all of the expected data types.

Galaxy 20.05 commit e75a792fc631ff216de6664d399f6bbe371086cd

<details><summary>Log</summary>

```
172.25.0.1 - - [09/Jul/2020:21:42:14 +0000] "GET /admin/data_types_list HTTP/1.1" 500 - "http://localhost/admin/data_types" "Mozilla/5.0 (X11; Linux x86_64; rv:78.0) Gecko/20100101 Firefox/78.0"
Traceback (most recent call last):
  File "/srv/galaxy/lib/galaxy/web/framework/middleware/error.py", line 154, in __call__
    app_iter = self.application(environ, sr_checker)
  File "/srv/galaxy/venv/lib/python3.8/site-packages/paste/recursive.py", line 85, in __call__
    return self.application(environ, start_response)
  File "/srv/galaxy/venv/lib/python3.8/site-packages/paste/httpexceptions.py", line 640, in __call__
    return self.application(environ, start_response)
  File "/srv/galaxy/lib/galaxy/web/framework/base.py", line 145, in __call__
    return self.handle_request(environ, start_response)
  File "/srv/galaxy/lib/galaxy/web/framework/base.py", line 224, in handle_request
    body = method(trans, **kwargs)
  File "/srv/galaxy/lib/galaxy/web/framework/decorators.py", line 58, in call_and_format
    return format_return_as_json(rval, jsonp_callback, pretty=(pretty or trans.debug))
  File "/srv/galaxy/lib/galaxy/web/framework/decorators.py", line 330, in format_return_as_json
    json = safe_dumps(rval, **dumps_kwargs)
  File "/srv/galaxy/lib/galaxy/util/json.py", line 70, in safe_dumps
    dumped = json.dumps(*args, allow_nan=False, **kwargs)
  File "/usr/local/lib/python3.8/json/__init__.py", line 234, in dumps
    return cls(
  File "/usr/local/lib/python3.8/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/local/lib/python3.8/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
[...]

PR #10148, opened 2020-08-25
[20.05] Fix admin data_types_list bug

cat-bro:
Issue #9977

The admin data types list is breaking because it tries to json dump an lxml _Attrib object.


dannon (approved):

dannon:
Thanks for the fix, tested locally and it works fine with this.

Issue #2968, opened 2016-09-23
Enhancement: Add more download options for datasets and histories


jennaj:
1. Compress datasets upon download
User request, see this biostars post for usage example: https://biostar.usegalaxy.org/p/19684/

2. Enable download of history archives by FTP (ideally with Filezilla, but curl or wget would also be nice and would parallel dataset download by URL). Why?? Archives >= 1k fail download from http://usegalaxy.org and maybe other Galaxy flavors.
Related to several user posts, including https://biostar.usegalaxy.org/p/22538/

bgruening:
Two things we might want to consider. The compression should not happen on the web-server but as part of the job-submission. The other is that if such a download failed, for some reason we need to clean up properly. People can easily fill up disc space by downloading multiple files that are cached. This does not count into the quota as far as I know.

jennaj:
@bgruening What do you think about an FTP download holding area that is an analog of the current FTP upload holding area? Per-account, outside of quota, datasets expire after N days or after downloaded successfully (data currently expires in 3 days for FTP upload or until moved into a history)? 

Then some way for users to browse that area (ideally with the same FTP clients used to upload - like Filezilla - but also line-command - both using galaxy server URL + account credentials) to drag files back to their computer in clients or find out the URL for the dataset to execute a fetching command (could be FTP, not just wget or curl). 

An idea .. basically the reverse of FTP upload populated by a compress-to-download job, or possibly straight-up batch download staging (originals, compressed or not).

bgruening:
Uh I like this idea!!!! And it should be not so hard to tie this into the existing FTP infrastructure I suppose. We could use the same FTP directory, or do you think this is to confusing? We could have a special folder called `download` that is hidden in the upload dialog?

jennaj:
I'd like to bring this idea up again to be prioritized. We have recently been getting more requests about how to effectively download histories.

Users are asking for known methods, like an FTP client, even if only for convenience (same simple method for getting data in AND out) or want to download line-command because they want to capture status/tracking of the transfer through tools they prefer to work with (including scripts). 

What are the current thoughts?

GuyReeves:
I wonder if this tool might do quite a bit if a tool were a solution (or intermediate step)

https://toolshed.g2.bx.psu.edu/repository?repository_id=d077871367f67b47

a simple tool which handles the compression of multiple datasets though a data collection (handles BAMs and the BAI index correctly).


This idea has come up else where.

[...]

PR #10152, opened 2020-08-26
Infrastructure for writing to pluggable Galaxy file sources.

jmchilton:
Initial plugin implementation added to 20.09 branch with https://github.com/galaxyproject/galaxy/pull/9888

### DIFF
--- a/lib/galaxy/files/sources/galaxy.py
+++ b/lib/galaxy/files/sources/galaxy.py
@@ -12,6 +12,7 @@ class UserFtpFilesSource(PosixFilesSource):
             root=root,
             label=label,
             doc=doc,
+            writable=True,
         )
         posix_kwds.update(kwd)
         if "delete_on_realize" not in posix_kwds:
--- a/lib/galaxy/files/sources/galaxy.py
+++ b/lib/galaxy/files/sources/galaxy.py
@@ -22,7 +22,7 @@ class UserFtpFilesSource(PosixFilesSource):
     def get_prefix(self):
         return None
 
-    def get_schema(self):
+    def get_scheme(self):
         return "gxftp"
 
 
@@ -42,7 +42,7 @@ class LibraryImportFilesSource(PosixFilesSource):
     def get_prefix(self):
         return None
 
-    def get_schema(self):
+    def get_scheme(self):
         return "gximport"
 
 
@@ -62,7 +62,7 @@ class UserLibraryImportFilesSource(PosixFilesSource):
     def get_prefix(self):
         return None
 
-    def get_schema(self):
+    def get_scheme(self):
         return "gxuserimport"
 
 
--- a/lib/galaxy/files/sources/posix.py
+++ b/lib/galaxy/files/sources/posix.py
@@ -65,11 +65,17 @@ class PosixFilesSource(BaseFilesSource):
     def _write_from(self, target_path, native_path, user_context=None):
         effective_root = self._effective_root(user_context)
         target_native_path = self._to_native_path(target_path, user_context=user_context)
-        # TODO: enforce_symlink_security
-        target_native_path = os.path.normpath(target_native_path)
-        assert target_native_path.startswith(os.path.normpath(effective_root))
+        if self.enforce_symlink_security:
+            if not safe_contains(effective_root, target_native_path, allowlist=self._allowlist):
+                raise Exception("Operation not allowed.")
+        else:
+            target_native_path = os.path.normpath(target_native_path)
+            assert target_native_path.startswith(os.path.normpath(effective_root))
 
-        # TODO: ensure directory exists, etc...
+        target_native_path_parent = os.path.dirname(target_native_path)
+        if not os.path.exists(target_native_path_parent):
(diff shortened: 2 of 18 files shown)


lib/galaxy/files/sources/galaxy.py
```diff
@@ -12,6 +12,7 @@ def __init__(self, label="FTP Directory", doc="Galaxy User's FTP Directory", roo
             root=root,
             label=label,
             doc=doc,
+            writable=True,
```
bgruening, line 15: Oh this is super cool! However, currently, our standard FTP server is usually not configured to allow downloads. We would need a second server with a different directory that only allows downloads and no uploads - is that correct @natefoo? If so, is it time to introduce `ftp_download_*` options?

jmchilton: I would think that would be a very natural place to start dumping large files (history exports, collection zips, etc.), so I think swapping the default here to ``True`` makes sense. It can be overridden explicitly with the config file but it probably makes sense to have galaxy.yml options also.

I guess for options we should make ``ftp_upload_dir`` an alias for ``ftp_dir`` and ``user_library_dir`` an alias for ``user_library_import_dir`` and then add ``ftp_dir_writable`` (default True) and ``user_library_dir_writable`` (default False)?

bgruening: > so I think swapping the default here to True makes sense.

Yes for sure. This is IMHO the easiest to set up globally and has the biggest impact.
But then I learned I can not download those datasets from the standard FTP directories :)

> ftp_dir_writable

I'm not sure if this is enough. Would like to have @natefoo opinion here. Is it still recommended to make the upload-ftp upload-only? I don't think having two FTP servers running is a big deal. But maybe all of that is not necessary anymore?

Another aspect is that two different directories for up- and download might make also sense from the admin point of view.
We have cron jobs running cleaning old unused FTP uploads ... we probably want to handle "archives" (downloads) differently.

jmchilton: I'm happy to get @natefoo's insights also. But I will just say none of this prevents you from configuring two FTP directories - the whole thing is configurable now. I just think that is an advanced configuration and shouldn't be the default assumed. 

If we're going to make a push toward documenting more advanced configurations - I don't think we should take user library import dir, ftp dir, and add a new fixed named thing like ftp download dir. We should instead work toward just documenting how to configure any of these things you want in a galaxy files configuration - including other directories with other options, etc..

bgruening:
xref: https://github.com/galaxyproject/galaxy/issues/2968

lib/galaxy/files/sources/galaxy.py
bgruening: > If we're going to make a push toward documenting more advanced configurations - I don't think we should take user library import dir, ftp dir, and add a new fixed named thing like ftp download dir. We should instead work toward just documenting how to configure any of these things you want in a galaxy files configuration - including other directories with other options, etc..

Ok, you convinced me. If we can easily use a second FTP this is not a problem at all. But the default configuration should be secure imho.

Thanks for this PR!

jmchilton:
Taking this out of WIP - I think this is a good atomic improvement. Next steps I think would be:

- Actually using this option to write out to the remote files server for the history export tool.
- Allowing users to create new directories when selecting them.

lib/galaxy/files/sources/posix.py
```diff
@@ -62,6 +62,16 @@ def realize_to(self, source_path, native_path, user_context=None):
...
             shutil.move(source_native_path, native_path)
 
+    def _write_from(self, target_path, native_path, user_context=None):
+        effective_root = self._effective_root(user_context)
+        target_native_path = self._to_native_path(target_path, user_context=user_context)
+        # TODO: enforce_symlink_security
+        target_native_path = os.path.normpath(target_native_path)
+        assert target_native_path.startswith(os.path.normpath(effective_root))
+
+        # TODO: ensure directory exists, etc...
```
bgruening: @jmchilton this looks crucial.

jmchilton: Yeah - I guess I need to write some more tests also. Thanks for catching that - I'll move this back into WIP.

jmchilton: I think this is taken care of now? Plus some tests for link security and such.

bgruening (approved):
Works flawlessly for me!

bgruening:
I restarted the failing test. @mvdbeek do you think we can get this in. It would be nice for the Beacon-COVID-19 project and would make my life a little bit easier.

mvdbeek (approved):
This is truly awesome, thanks a lot @jmchilton!

bgruening:
:1st_place_medal: :heart:

jmchilton:
I thought that this was going to take a lot more work to build out some applications and get people excited - sounds like y'all already have that covered though. That is awesome - I'm excited to see how this is used!

bgruening:
Mine will just be hacks, as usual, but this makes such hakes I think way easier :)

PR #10154, opened 2020-08-26
Allow unary plus and minus in filtering

wm75:
Adding `USub` to the list of allowed ast node types makes it possible to use filter expressions like `c1 >= -5` (which addresses a bug report received at usegalaxy.eu).
Allowing this type (and also `UAdd`) seems benign to me. Does anyone else see a problem with this?


bgruening (approved):
Looks good to me.

PR #10155, opened 2020-08-26
Fix determining 'int', 'float' column type in Tabular

selten:
If a `_` is present in an int or float, python 3 will still validate it as being int or float and remove the `_`.

### DIFF
--- a/lib/galaxy/datatypes/tabular.py
+++ b/lib/galaxy/datatypes/tabular.py
@@ -308,20 +308,20 @@ class Tabular(TabularData):
             raise ValueError("Tried to compare unknown column types: {} and {}".format(column_type1, column_type2))
 
         def is_int(column_text):
+            # Don't allow underscores in numeric literals (PEP 515)
+            if '_' in column_text:
+                return False
             try:
-                # Don't allow underscores in numeric literals (PEP 515)
-                if '_' in column_text:
-                    return False
                 int(column_text)
                 return True
             except ValueError:
                 return False
 
         def is_float(column_text):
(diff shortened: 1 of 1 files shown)


lib/galaxy/datatypes/tabular.py
```diff
@@ -309,13 +309,17 @@ def type_overrules_type(column_type1, column_type2):
 
         def is_int(column_text):
             try:
+                if any(c in ['_'] for c in column_text):
```
nsoranzo:
```suggestion
                # Don't allow underscores in numeric literals (PEP 515)
                if '_' in column_text:
```

lib/galaxy/datatypes/tabular.py
```diff
@@ -309,13 +309,17 @@ def type_overrules_type(column_type1, column_type2):
...
+                    return False
                 int(column_text)
                 return True
             except ValueError:
                 return False
 
         def is_float(column_text):
             try:
+                if any(c in ['_'] for c in column_text):
```
nsoranzo:
```suggestion
                # Don't allow underscores in numeric literals (PEP 515)
                if '_' in column_text:
```

wm75: or slightly cleaner, move the check out of the try/except

wm75:
What about those same methods in BaseCSV?

selten:
@wm75 I just did that, I also added it to BaseCSV. I don't use that myself at all, so I would've forgotten about it.

nsoranzo (approved):

PR #10159, opened 2020-08-27
Fix for the hour value depicted as a float value in CPU and Job Runtime

npinter:
Hours in CPU and Job Runtime are depicted as a float value of the full CPU and Job Runtime, respectively.
grafik

This fix floors the hour value.

### DIFF
--- a/lib/galaxy/job_metrics/formatting.py
+++ b/lib/galaxy/job_metrics/formatting.py
@@ -15,4 +15,4 @@ def seconds_to_str(value):
     elif value < 3600:
         return "%s minutes" % round(value / 60, 2)
     else:
-        return "{} hours and {} minutes".format(int(value / 3600), round((value % 3600) / 60, 2))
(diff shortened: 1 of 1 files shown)


lib/galaxy/job_metrics/formatting.py
```diff
@@ -15,4 +15,4 @@ def seconds_to_str(value):
     elif value < 3600:
         return "%s minutes" % round(value / 60, 2)
     else:
-        return "{} hours and {} minutes".format(round(value / 3600, 2), round((value % 3600) / 60, 2))
+        return "{} hours and {} minutes".format(int(value / 3600), round((value % 3600) / 60, 2))
```
mvdbeek:
```suggestion
        return "{} hours and {} minutes".format(value // 3600, round((value % 3600) / 60, 2))
```

mvdbeek (approved):
Nice find, just a minor stylistic suggestion here.

dannon (approved):

PR #10170, opened 2020-08-31
Toolbox refactoring, styling, and a couple bugfixes

dannon:
- Swaps to explicit panel property configuration instead of inheriting entire global configuration blob.
- Fixes a bug where tools view was actually rendering the toolbox twice, unnecessarily.
- Fixes a 'double scroller' bug where in some situations you'd get two scrollers in the panels.
- Adjusts basis for border color to be aligned with panels instead of text -- this is an issue I ran into when styling for the anvil branch.
- Relocates tool section label styles to be scoped and specific, not global
- Removes a few extraneous divs that served no purpose.

Updates tool section label styling (toplevel and subsection) to be more easily visually parsed.  -- I'm definitely looking for more feedback on the section label styling.  It's a bit of a pickle because it'd almost be nicer to standardize towards a third tier of section (or a drawer with toolSections in it -- something we've talked about specifically in the context of data ingress recently) instead of having a mix of labels and sections, but this is an option.

After:
image

### DIFF
--- a/client/src/components/Panels/Common/ToolSection.test.js
+++ b/client/src/components/Panels/Common/ToolSection.test.js
@@ -40,7 +40,7 @@ describe("ToolSection", () => {
         await Vue.nextTick();
         const $names = wrapper.findAll(".name");
         expect($names.at(1).text()).to.equal("name");
-        const $label = wrapper.find(".label");
+        const $label = wrapper.find(".tool-panel-label");
         expect($label.text()).to.equal("text");
         $sectionName.trigger("click");
         await Vue.nextTick();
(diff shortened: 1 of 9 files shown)


bgruening:
Thanks  @dannon! We prefer the first one.

hexylena:
Maybe add some padding to the top of those labels? They feel very squished together

bgruening:
Great work @dannon! Works nicely in my hands.

dannon:
This might also also now fixes the selenium tests that were broken with the tool search caching refinements a while back.

bgruening:
Thanks @dannon!

PR #10172, opened 2020-08-31
Uniformly process config files

### LINKED ISSUE #9973: The path for config files doesn't be resolved with respect to <config_dir>

### LINKED ISSUE #10187: Resolving config file paths works correctly either for tests or for user-set values, but not both
When the test driver launches Galaxy, it supplies test-specific configuration settings; among them *relative paths* to configuration files. For example, `tool_data_table_config_path`:
```
lib/galaxy/config/sample/tool_data_table_conf.xml.sample
test/functional/tool-data/sample_tool_data_tables.xml
```
These paths are relative to Galaxy's root, so when resolved, they should be:
```
/[galaxy-root]/lib/galaxy/config/sample/tool_data_table_conf.xml.sample
/[galaxy-root]/test/functional/tool-data/sample_tool_data_tables.xml
```
However, the config system treats them as any user-set value and resolves them w.r.t their parent directory, as per schema, which is `config_dir` (which has been resolved w.r.t Galaxy's root). This **is the expected behavior**, but it results in incorrect paths:
```
/[config-dir]/lib/galaxy/config/sample/tool_data_table_conf.xml.sample
/[config-dir]/test/functional/tool-data/sample_tool_data_tables.xml
```

The error does not happen because of `config._parse_config_file_options()` which resolves all config file options w.r.t Galaxy root (https://github.com/galaxyproject/galaxy/blob/release_20.05/lib/galaxy/config/__init__.py#L305). Which is incorrect: it works for tests, but not for user-set values (see bug description in #9973).

The solution for this is not complicated (any day now: #10172), but I think this is critical enough to be recorded as a standalone issue.

Ping @natefoo (FYI)

jdavcs:
This is ready for review. It replaces #10141.
Fixes #9973 (specific bug: https://github.com/galaxyproject/galaxy/issues/9973#issuecomment-655792443), #10187; also closes ongoing #8493.

**Overall:** this streamlines loading of 33 config files (most of them defined in Galaxy and TS schemas), which used to be individually processed [here](https://github.com/galaxyproject/galaxy/blob/release_20.05/lib/galaxy/config/__init__.py#L756), [here](https://github.com/galaxyproject/galaxy/blob/release_20.05/lib/tool_shed/webapp/config.py#L45), and then [here](https://github.com/galaxyproject/galaxy/blob/release_20.05/lib/galaxy/config/__init__.py#L274). Now, we just need to specify a config option's default value and parent directory in the schema. Any additional processing can be specified [here](https://github.com/galaxyproject/galaxy/blob/release_20.05/lib/galaxy/config/__init__.py#L372). A line needs to be also added to a [unit test](https://github.com/galaxyproject/galaxy/blob/release_20.05/test/unit/config/test_config_values.py) and an [integration test](https://github.com/galaxyproject/galaxy/blob/release_20.05/test/integration/test_config_defaults.py). 

**Motivation:** 
The core problem was non-uniform processing of config file options: we loaded them uniformly ([using defaults from the schema, adding some processing steps, etc.](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/__init__.py#L114)). But then we reloaded some of them in `parse_config_file_options` - that overrides not only the schema defaults, but also any processing steps previously executed. User-set relative config file paths being resolved with respect to the root directory instead of their correct location is one example of a bug (see #9973). Not being able to set any of the options that pass through `parse_config_file_options` to `None` is another unintended consequence (currently, prior to this PR, the user *can* explicitly set all other config options to `None`; if we want to disallow this for these options, we can do that explicitly, and not as an implicit side-effect).

**What this contains:** 
1. Remove `parse_config_file_options`; all options are now processed uniformly.
1. Replace references to`config.foo_set` with calls to `config.is_set('foo')`
1. Generalize handling of sample files added to defaults (see preprocessing step before resolving paths: [here](https://github.com/galaxyproject/galaxy/blob/841261ec6a0c4ec3fbe1dc28f23ace8ad221ee99/lib/galaxy/config/__init__.py#L126))
1. Generalize handling cases where we select one or all files from a list of files (see postprocessing step after resolving paths: [here](https://github.com/galaxyproject/galaxy/blob/841261ec6a0c4ec3fbe1dc28f23ace8ad221ee99/lib/galaxy/config/__init__.py#L128)). @natefoo: I've used your code for both cases - this replaces `defaults` and `listify_defaults` as in lines [274-305](https://github.com/galaxyproject/galaxy/blob/release_20.05/lib/galaxy/config/__init__.py#L274)
1. Add a few missing config options to the schema
1. Apply same refactoring to TS config (step was necessary)
1. Resolve paths in `driver_util.py` *before* passing the test config to Galaxy app. This fixes #10187 (see commit bbfddce)
1. Add unit tests for all of the above (+refactor lots of old tests; remove mocks where possible)

In the first version, I generated unit tests for overwriting defaults for each config option defined in the schema ([see old PR](https://github.com/galaxyproject/galaxy/blob/4e5eb31384919c13bc46bf83f7e26827b431cb30/test/unit/config/test_config_values.py)). I dropped them after [feedback](https://github.com/galaxyproject/galaxy/pull/10141#pullrequestreview-473403530) (thanks, @jmchilton!); also because the problematic code is now removed, and because the generalized functionality is covered by tests (see all unit test files in this PR).

I'll write a "How-to" entry on configuration edits for the documentation after all edits.

Sorry for making this so large. I considered splitting this, but decided that it will make more sense to keep it as one logical unit. Each commit is focused on one change; the commit message contains details about what was done and why. The "Refactor processing of ..." commit messages are not all the same. There's only one change that is modified in 2 commits (making `data_manager_config_file` path absolute in `driver_util.py`: the later commit contains a generalized solution).


ic4f:
Note to self: status so far: tests pass up to 8f90fcd; next commit (f8f150b) seems to be failing 1-3 integration tests that are unrelated.

ic4f:
Note to self / status so far: all green as of f922c18.

jmchilton (approved):

PR #10174, opened 2020-09-01
Send activation email upon custos account creation

juleengraham:
if user activation is set in galaxy.yml
address email verification in #10075

### DIFF
--- a/lib/galaxy/authnz/custos_authnz.py
+++ b/lib/galaxy/authnz/custos_authnz.py
@@ -115,8 +115,6 @@ class CustosAuthnz(IdentityProvider):
                     user = trans.app.user_manager.create(email=email, username=username)
                     if trans.app.config.user_activation_on:
                         trans.app.user_manager.send_activation_email(trans, email, username)
-                    trans.sa_session.add(user)
-                    trans.sa_session.flush()
             custos_authnz_token = CustosAuthnzToken(user=user,
                                    external_user_id=user_id,
                                    provider=self.config['provider'],
--- a/lib/galaxy/authnz/custos_authnz.py
+++ b/lib/galaxy/authnz/custos_authnz.py
@@ -113,6 +113,8 @@ class CustosAuthnz(IdentityProvider):
                         raise exceptions.AuthenticationFailed(message)
                 else:
                     user = trans.app.user_manager.create(email=email, username=username)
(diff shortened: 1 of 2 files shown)


lib/galaxy/authnz/custos_authnz.py
```diff
@@ -113,6 +113,8 @@ def callback(self, state_token, authz_code, trans, login_redirect_url):
                         raise exceptions.AuthenticationFailed(message)
                 else:
                     user = trans.app.user_manager.create(email=email, username=username)
+                    if trans.app.config.user_activation_on:
+                        trans.app.user_manager.send_activation_email(trans, email, username)
                     trans.sa_session.add(user)
                     trans.sa_session.flush()
```
dannon:
```suggestion
                    trans.sa_session.add(user)
                    trans.sa_session.flush()
                    if trans.app.config.user_activation_on:
                        trans.app.user_manager.send_activation_email(trans, email, username)
```

dannon: I'd finish the user creation transaction prior to sending the email here; it looks like that's the order in the user manager as well.

dannon: Actually, we need to look deeper here.  Scratch the suggestion; the user create method in the manager invoked intentionally activates the user when created like this, and I think that's the right behavior -- so the user *should* already be activated.  Can you remind me what exactly the situation was that needed this fix?

juleengraham: The user activation isn't being sent. I think the specific situation was that upload wasn't working because the user wasn't activated. Also, when I create a new user on my local instance, I do get the warning saying that the account email verification ahsn't been done

dannon: Ahh!  I was looking at https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/managers/users.py#L101 and read the logic inverted.  So, yes, this looks like it'll do it (though I'd still recommend accepting the suggestion above).

dannon: One other fix we could add here is that I think this is an extra add/flush -- it looks like the user manager handles that and it's not necessary.

dannon (approved):

PR #10175, opened 2020-09-01
Use tool_enable_ngram_search by default, whitespace search fix

mvdbeek:
https://github.com/galaxyproject/galaxy/commit/370fb4aba470ec0bd8c31c1e27ad091e10b04de4 also fixes whitespace search issues in the "legacy search" (https://github.com/galaxyproject/galaxy/issues/9741#issuecomment-666372567). ngram search didn't have this problem to begin with since it uses `StandardAnalyzer` to tokenize the query string.

If this does well I think we should remove the legacy search in 21.01.

### DIFF
--- a/doc/source/admin/galaxy_options.rst
+++ b/doc/source/admin/galaxy_options.rst
@@ -1556,6 +1556,21 @@
 :Type: str
 
 
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+``trs_servers_config_file``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+:Description:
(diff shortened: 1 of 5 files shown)


mvdbeek:
And while I'm at it, https://github.com/galaxyproject/galaxy/pull/10175/commits/bb4e5248187b644bd3e20eed617b32273ff01c9a removes tools that are not the latest version (and whose latest version is not hidden) from the index

jmchilton (approved):

martenson (approved):
We should keep an eye on the `-` "dash" wildcard, I remember it was causing issues even with analysers.

martenson:
Thanks bud!

PR #10180, opened 2020-09-03
Use "Google" instead of "google-openidconnect"

juleengraham:
on external id management page


juleengraham:
Actually, this will cause problems with disconnecting so switching to draft fn

martenson:
I think this is now better controlled using <provider> configuration options by admins

PR #10181, opened 2020-09-03
javascript es6 arrow functions

juleengraham:


dannon:
I do prefer the shorthand here, but, importantly, these are not es6 arrow functions -- it's object literal shorthand.  Arrow functions would actually break if used here, because `this` wouldn't be bound.

Test this with something like (from one of the tweaks above):
```
    methods: {  // original code
        toggleLogin: function () {
            console.debug(this);      // the vue component
...

    methods: {  // this PR (object literal shorthand)
        toggleLogin() {
            console.debug(this);      // the vue component
...

    methods: {  // es6 function
        toggleLogin: () => {
            console.debug(this);      // undefined
```

mvdbeek:
This is still good to go, right @dannon?

mvdbeek (approved):

dannon:
Yes, it's fine.   I fixed the unrelated failing selenium test (had to do with tool cache changes) in another branch.

PR #10189, opened 2020-09-04
Allow CLI runner modules to be loaded from sys.path

natefoo:
This is mainly for Pulsar (we noticed this while attempting to set up a Pulsar Embedded runner), but will eventually be useful for running Galaxy installed.

Pulsar will need the following change:
```patch
diff --git a/pulsar/managers/queued_cli.py b/pulsar/managers/queued_cli.py
index 54be762..c417a61 100644
--- a/pulsar/managers/queued_cli.py
+++ b/pulsar/managers/queued_cli.py
@@ -17,7 +17,7 @@ class CliQueueManager(ExternalBaseManager):

     def __init__(self, name, app, **kwds):
         super(CliQueueManager, self).__init__(name, app, **kwds)
-        self.cli_interface = CliInterface(code_dir='.')
+        self.cli_interface = CliInterface(code_dir=None)
         self.shell_params, self.job_params = split_params(kwds)

     def launch(self, job_id, command_line, submit_params={}, dependencies_description=None, env=[], setup_params=None):
```

And we should probably switch Pulsar to depend on the relevant Galaxy package for this rather than periodically copying this stuff in to the Pulsar source tree.

### DIFF
--- a/lib/galaxy/jobs/runners/util/cli/__init__.py
+++ b/lib/galaxy/jobs/runners/util/cli/__init__.py
@@ -1,6 +1,8 @@
 """
 """
+import importlib
 import json
+import pkgutil
 from glob import glob
 from os import getcwd
 from os.path import (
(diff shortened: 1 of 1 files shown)


mvdbeek:
> And we should probably switch Pulsar to depend on the relevant Galaxy package for this rather than periodically copying this stuff in to the Pulsar source tree.

Shouldn't that be the other way round ? Maintain in pulsar, import in Galaxy ?

natefoo:
The runner util code originates in and is used heavily by Galaxy - unless Pulsar becomes Galaxy's only execution engine I'm not sure why it would make sense for Pulsar to "own" this?

PR #10191, opened 2020-09-05
Switch tool document cache to diskcache

mvdbeek:
This should be more portable across different Python installations.
If the fallback dbm.dumb module was used (instead of dbm.gnu or dbm/ndbm) startup could become very, very slow. diskcache uses SQLite and plain files to store the cache. With this being more portable we should be able to store this on CVMFS and have other Galaxy instances be able to consume the cache.


lib/galaxy/dependencies/pipfiles/default/Pipfile
```diff
@@ -58,6 +58,7 @@ WebOb = "*"
 Mako = "*"
 Babel = "*"
 Beaker = "*"
+diskcache = "*"
```
nsoranzo: Can you also remove `dogpile.cache` from here and `lib/galaxy/dependencies/pipfiles/default/pinned-requirements` (assuming it's not needed by other packages)?

lib/galaxy/tools/cache.py
```diff
@@ -1,22 +1,15 @@
...
 from collections import defaultdict
 from threading import Lock
 
-from dogpile.cache import make_region
-from dogpile.cache.api import (
-    CachedValue,
-    NO_VALUE,
-)
-from dogpile.cache.proxy import ProxyBackend
+from diskcache import Cache, JSONDisk
```
```suggestion
from diskcache import (
    Cache,
    JSONDisk
)
```

lib/galaxy/dependencies/pipfiles/default/Pipfile
mvdbeek: Sure!

lib/galaxy/tools/cache.py
```diff
@@ -25,65 +18,42 @@
 CURRENT_TOOL_CACHE_VERSION = 2
```
nsoranzo: Should we increase (or reset) this value? We now cache `document` instead of `metadata`.

lib/galaxy/dependencies/pipfiles/default/Pipfile
mvdbeek: Actually, I think it might be, we've had dog.pile in our dependencies before I started using it for the document cache (https://github.com/galaxyproject/galaxy/blob/release_20.01/lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt#L55)

lib/galaxy/tools/cache.py
mvdbeek:
```suggestion
CURRENT_TOOL_CACHE_VERSION = 1
```

lib/galaxy/dependencies/pipfiles/default/Pipfile
nsoranzo: OK, but you can still remove it from this file (`Pipfile`), which lists our explicit dependencies.

mvdbeek: I'm a bit confused, it's not in this file, is it ?

nsoranzo: Ah, it should have been there, but you're right it's not, so nothing to do!

nsoranzo:
I've opened https://github.com/mvdbeek/galaxy/pull/14 to move the tool creation logic from ``cache.py`` to ``__init__.py``, which I think it's cleaner than passing a creator method to `get_or_create_cached_tool_source()`. What do you think?

jmchilton (approved):

nsoranzo (approved):

mvdbeek:
I think the failing converter tests are probably real issues, don't merge this yet.

nsoranzo:
From https://app.circleci.com/pipelines/github/galaxyproject/galaxy/9760/workflows/624105c4-10cc-4cf2-b301-4ab789cc000f/jobs/104054 :
 
```
DEBUG:galaxy.tools.cache:Exception while checking tools to remove from cache: 'Index' object has no attribute 'delete'
```

mvdbeek:
Hmm, the converters hang at random places in the tests. When I dump the stack I'm seeing
```
Thread 7509 (idle): "LocalRunner.work_thread-2"
    _execute_child (subprocess.py:1482)
    __init__ (subprocess.py:800)
    queue_job (galaxy/jobs/runners/local.py:108)
    run_next (galaxy/jobs/runners/__init__.py:137)
    run (threading.py:870)
    _bootstrap_inner (threading.py:926)
    _bootstrap (threading.py:890)
```
That's https://github.com/python/cpython/blob/v3.7.9/Lib/subprocess.py#L1482, which we call in https://github.com/mvdbeek/galaxy/blob/85b07b18bdd6a4f0045d1285977ed7d637476862/lib/galaxy/jobs/runners/local.py#L102

I think there's something wrong with spawning the process ... I am not sure if this is related to the PR. On the one hand I haven't seen these timeouts in other PRs, but this happens reproducibly in this PR. @nsoranzo do you have a guess what could be going on here ?

lib/galaxy/tools/__init__.py
```diff
@@ -289,26 +294,40 @@ def tools_by_id(self):
 
     def get_cache_region(self, tool_cache_data_dir):
         if tool_cache_data_dir not in self.cache_regions:
-            self.cache_regions[tool_cache_data_dir] = create_cache_region(tool_cache_data_dir)
+            try:
+                with open(os.path.join(tool_cache_data_dir, 'cache.json')) as cache_file:
+                    self.cache_regions[tool_cache_data_dir] = json.load(cache_file)
```
nsoranzo: Can this JSON file become too large to always have it loaded in memory as part of `app.toolbox` ? I liked having a database-backed cache :pensive:

lib/galaxy/tools/__init__.py
```diff
@@ -566,7 +585,7 @@ def remove_from_cache(self):
         source_path = self.tool_source._source_path
         if source_path:
             for region in self.app.toolbox.cache_regions.values():
-                region.delete(source_path)
+                region.pop(source_path, None)
```
nsoranzo: I think we need to call `persist_cache_region()` here.

lib/galaxy/tools/__init__.py
mvdbeek: Yeah, I'm sorry, I should have put this back into WIP. I have another variant that uses hdf5py that could work, but ... looks like that also has problems with multiple readers and writers. I am now considering writing a single JSON file per tool or debugging the diskcache/sqlite approach ... but I don't have a good lead on how to do that.

lib/galaxy/tools/cache.py
```diff
@@ -1,89 +1,89 @@
...
+
+    def persist(self):
+        if self.writeable_cache_file:
+            self._cache.commit()
+            self.writeable_cache_file.close()
+            self.writeable_cache_file = None
+            self._cache = SqliteDict(self.cache_file, flag='r', encode=encoder, decode=decoder, autocommit=False)
+
+    def set(self, config_file, tool_source):
+        self.make_writable()
```
nsoranzo, line 72: I think it would be better to reopen the cache file inside a `with` statement otherwise other threads/processes would not be able to write to it.

mvdbeek: So the goal of that with statement is to avoid re-assigning `self._cache` here ?

lib/galaxy/tools/__init__.py
mvdbeek: Yep, thanks! I've moved that into the caller so that we just need to persist once per set of removed tools.

mvdbeek:
While this isn't perfect (if starting N processes at the same time there will be N temporary sqlite caches) it avoids file-locking and the dbm module, both of which were problems previously.

lib/galaxy/tools/cache.py
nsoranzo: I realise now that's a bad idea since instantiating `SqliteDict` creates a thread to serialise access to the connection.

Why not have it always open for read and write?

lib/galaxy/tools/cache.py
```diff
@@ -1,89 +1,96 @@
...
+        # Create database if necessary using 'c' flag
+        self._cache = SqliteDict(self.cache_file, flag='c', encode=encoder, decode=decoder, autocommit=False)
+        # Switch SqliteDict back to readonly
+        self._cache.flag = 'r'
+
+    def get(self, config_file):
+        tool_document = self._cache.get(config_file)
+        if not tool_document:
+            return None
+        if tool_document.get('tool_cache_version', 0) != CURRENT_TOOL_CACHE_VERSION:
```
```suggestion
        if tool_document.get('tool_cache_version') != CURRENT_TOOL_CACHE_VERSION:
```

nsoranzo (approved):

lib/galaxy/tools/cache.py
mvdbeek: The idea is that `make_writable` creates a tempfile next to the actual cache where the actual writing happens. When that is done we just the file back into place and re-open in read-only mode (read-only mode is technically not necessary since we call `make_writable` for every action that updates the database). This way multiple processes can write to the database at the same time. If we don't do this we get `database locked` exceptions when you start up multiple separate Galaxy processes at the same time (if the cache database is empty, if it's fully populated we don't ever call any of the update methods) -- we could place a lock around the database file, and serialize access that way, but we haven't had good success with file locking on the Anvil / GCP VMs.

lib/galaxy/dependencies/pipfiles/default/Pipfile
nsoranzo: This is not used any more, should be replaced by `sqlitedict`.

lib/galaxy/dependencies/pipfiles/default/pinned-requirements.txt
```diff
@@ -49,6 +49,7 @@ defusedxml==0.6.0 ; python_version >= '3.0'
 deprecated==1.2.10
 deprecation==2.1.0
 dictobj==0.4
+diskcache==5.0.2
```
nsoranzo: Not used any more.

nsoranzo (approved):

PR #10195, opened 2020-09-07
[20.05] Force pagination on Valid Galaxy Tools -> Tools page

mvdbeek:
If fetching all tools at once this route is both slow and consumes a lot of memory, and we suspect it's one of the reason the TS runs out of memory regularly.

### DIFF
--- a/client/galaxy/scripts/legacy/grid/grid-template.js
+++ b/client/galaxy/scripts/legacy/grid/grid-template.js
@@ -202,6 +202,7 @@ export default {
             var num_page_links = options.num_page_links;
             var cur_page_num = options.cur_page_num;
             var num_pages = options.num_pages;
+            const allow_fetching_all_results = options.allow_fetching_all_results;
 
             // First pass on min page.
             var page_link_range = num_page_links / 2;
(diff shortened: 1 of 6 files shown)


mvdbeek:
Since the TS tests are passing I'm going to push this to the toolshed branch and deploy from there.

mvdbeek:
With this PR applied (last drop in memory below) it seems once a process reaches 1370 MB of resident memory, it sort of stays there without larger bumps.

Screenshot 2020-09-08 at 10 35 12

There are still some minor bumps that might be worth looking at, but if someone wants to merge this I wouldn't object.

mvdbeek:
Looking through the logs, a lot of the remaining slow responses (those don't necessarily equate with requests that consume a lot of memory) are bots accessing mercurial repos changesets and triggering downloads.
We may want to set up something like https://www.mercurial-scm.org/wiki/PublishRepositoriesOnNginx?action=show&redirect=HgServeNginx ... that might also give us better cloning speeds, maybe ?

mvdbeek:
@natefoo I've worked out a config for this @ https://gist.github.com/mvdbeek/81eaf567e832c3e2f971cd04955c7769

Feels snappy, and I think this is probably good to have. That means we can clone without a running toolshed server, and we get a bit of isolation here, since both the tool shed and mercurial can be memory hogs for different reasons. Should be totally safe to just restart uwsgi once a given memory threshold is reached, while that is less so for the tool shed.

PR #10197, opened 2020-09-08
use GalaxyTagHandlerSession for data library uploads

scholtalbers:
Without the `GalaxyTagHandlerSession` I receive integrity errors about tag key 'name' already existing when uploading a dataset with multiple tags.

### DIFF
--- a/lib/galaxy/tools/actions/upload_common.py
+++ b/lib/galaxy/tools/actions/upload_common.py
@@ -232,14 +232,13 @@ def __new_library_upload(trans, cntrller, uploaded_dataset, library_bunch, state
                                                             user=trans.user,
                                                             create_dataset=True,
                                                             sa_session=trans.sa_session)
+    tag_manager = tags.GalaxyTagHandlerSession(trans.sa_session)
     if uploaded_dataset.get('tag_using_filenames', False):
         tag_from_filename = os.path.splitext(os.path.basename(uploaded_dataset.name))[0]
-        tag_manager = tags.GalaxyTagHandlerSession(trans.sa_session)
         tag_manager.apply_item_tag(item=ldda, user=trans.user, name='name', value=tag_from_filename)
 
     tags_list = uploaded_dataset.get('tags', False)
     if tags_list:
-        tag_manager = tags.GalaxyTagHandlerSession(trans.sa_session)
         for tag in tags_list:
             tag_manager.apply_item_tag(item=ldda, user=trans.user, name='name', value=tag)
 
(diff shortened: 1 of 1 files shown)


mvdbeek (commented):
Thanks so much for looking at this. I think that's almost right, but it's better to use the same instance of `GalaxyTagHandlerSession` for all tags to be created.

lib/galaxy/tools/actions/upload_common.py
```diff
@@ -234,12 +234,12 @@ def __new_library_upload(trans, cntrller, uploaded_dataset, library_bunch, state
...
     if uploaded_dataset.get('tag_using_filenames', False):
         tag_from_filename = os.path.splitext(os.path.basename(uploaded_dataset.name))[0]
-        tag_manager = tags.GalaxyTagHandler(trans.sa_session)
+        tag_manager = tags.GalaxyTagHandlerSession(trans.sa_session)
         tag_manager.apply_item_tag(item=ldda, user=trans.user, name='name', value=tag_from_filename)
 
     tags_list = uploaded_dataset.get('tags', False)
     if tags_list:
-        tag_manager = tags.GalaxyTagHandler(trans.sa_session)
+        tag_manager = tags.GalaxyTagHandlerSession(trans.sa_session)
```
mvdbeek: I think you can set up one instance of `GalaxyTagHandlerSession ` on line 314, and then pass this into `new_upload` and `__new_library_upload`. Otherwise adding the same tag prefix twice would still lead to errors.

scholtalbers: thanks for the comment @mvdbeek. I added the instance creation just a few line up. Going up to `new_upload` didn't seem so fit to me as there is also a `trans.app.tag_handler` which actually confuses me a bit. Overall there seem to be many instances of tag_handler being instantiated in the code base.

mvdbeek: There's a global one (`trans.app.tag_handler`) that requires a flush after each added tag (otherwise you're seeing the issue that your trying to fix with this PR). If you don't want to flush after setting each tag you need to use `GalaxyTagHandlerSession`. The reason the PR works for you is because the GalaxyTagHandlerSession scope now includes the for loop on line 242. But, there are other places where the global tag handler is used in this file and it shouldn't so the idea is to use just a single `GalaxyTagHandlerSession` to avoid all of these issues.

mvdbeek (approved):

lib/galaxy/tools/actions/upload_common.py
scholtalbers: got it. I placed the instance creation in `new_upload` as that gets also called in e.g. `upload._precreate_fetched_hdas`

PR #10201, opened 2020-09-09
If a config file is not found, try resolving w.r.t. root

jdavcs:
Addresses the case when an existing Galaxy configuration expects config files to be resolved w.r.t. root (as was the case prior to #10172).

Approach: 
- keep set of options to check in `paths_to_check_against_root`
- after a path is resolved, if option is set and path does not exist, try resolving initially set value w.r.t root; update option if path exists
- handle both single values and lists of values. For lists, check and update each path in list

Unit tests included.

See #10200 for additional context.


jmchilton (approved):

ic4f:
Selenium test failure unrelated.

PR #10203, opened 2020-09-09
Fix failing Selenium test.

jmchilton:
Broken yesterday with https://github.com/galaxyproject/galaxy/pull/9618 I think.

### DIFF
--- a/lib/galaxy/selenium/smart_components.py
+++ b/lib/galaxy/selenium/smart_components.py
@@ -98,3 +98,11 @@ class SmartTarget:
 
     def wait_for_and_send_keys(self, text):
         self.wait_for_visible().send_keys(text)
+
+    def hack_js_click(self):
+        """Hack to force click at JavaScript level (bypassing screen interactions).
+
+        This should generally be avoided.
+        """
+        elem = self._has_driver.driver.find_element(*self._target.element_locator)
+        self._has_driver.driver.execute_script("arguments[0].click();", elem)
--- a/lib/galaxy/selenium/smart_components.py
+++ b/lib/galaxy/selenium/smart_components.py
@@ -98,11 +98,3 @@ class SmartTarget:
 
     def wait_for_and_send_keys(self, text):
         self.wait_for_visible().send_keys(text)
-
-    def hack_js_click(self):
-        """Hack to force click at JavaScript level (bypassing screen interactions).
-
-        This should generally be avoided.
-        """
-        elem = self._has_driver.driver.find_element(*self._target.element_locator)
-        self._has_driver.driver.execute_script("arguments[0].click();", elem)
--- a/lib/galaxy_test/selenium/test_admin_app.py
+++ b/lib/galaxy_test/selenium/test_admin_app.py
@@ -105,7 +105,6 @@ class AdminAppTestCase(SeleniumTestCase):
         manage_jobs = admin_component.manage_jobs
         lock_label = manage_jobs.job_lock_label
         original_label = lock_label.wait_for_text()
-        lock = manage_jobs.job_lock
         lock_label.wait_for_and_click()
         self.sleep_for(self.wait_types.UX_TRANSITION)
         # Make sure the job lock has been toggled.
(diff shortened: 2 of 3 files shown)


lib/galaxy/selenium/smart_components.py
```diff
@@ -98,3 +98,11 @@ def has_class(self, class_name):
...
     def wait_for_and_send_keys(self, text):
         self.wait_for_visible().send_keys(text)
+
+    def hack_js_click(self):
+        """Hack to force click at JavaScript level (bypassing screen interactions).
+
+        This should generally be avoided.
+        """
+        elem = self._has_driver.driver.find_element(*self._target.element_locator)
+        self._has_driver.driver.execute_script("arguments[0].click();", elem)
```
OlegZharkov: Why do you need this? If it's a pseudo element, we have a very good way of `move_to_element_with_offset.click().perform()`

jmchilton: I didn't work for me with this widget 😢 - this is why the tests are broken. I don't have an explanation.

OlegZharkov: I tried that now... Why do you want to click exactly on `#prevent-job-dispatching`?
if you want to lock that thing, you can just do it on `job_lock_label`

lib/galaxy_test/selenium/test_admin_app.py
```diff
@@ -102,20 +102,21 @@ def test_admin_jobs_display(self):
...
+        lock.hack_js_click()
         self.sleep_for(self.wait_types.UX_TRANSITION)
         # Make sure the job lock has been toggled.
-        label = self.driver.find_element_by_xpath("//label[@for='prevent-job-dispatching']/strong")
-        self.assertNotEqual(label.text, original_label)
+        new_label = lock_label.wait_for_text()
+        self.assertNotEqual(new_label, original_label)
         self.screenshot("admin_jobs_locked")
-        self.action_chains().move_to_element_with_offset(lock, -20, 5).click().perform()
+        lock.hack_js_click()
```
OlegZharkov:
```suggestion
        manage_jobs.job_lock_label.wait_for_and_click()
```

lib/galaxy_test/selenium/test_admin_app.py
```diff
@@ -102,20 +102,21 @@ def test_admin_jobs_display(self):
...
         # label for the job lock toggle to verify that job locking actually happens
-        label = self.driver.find_element_by_xpath("//label[@for='prevent-job-dispatching']/strong")
-        lock = self.driver.find_element_by_id("prevent-job-dispatching")
-        original_label = label.text
-        self.action_chains().move_to_element_with_offset(lock, -20, 5).click().perform()
+        manage_jobs = admin_component.manage_jobs
+        lock_label = manage_jobs.job_lock_label
+        original_label = lock_label.wait_for_text()
+        lock = manage_jobs.job_lock
+        lock.hack_js_click()
```
OlegZharkov:
```suggestion
        manage_jobs.job_lock_label.wait_for_and_click()
```

lib/galaxy/selenium/smart_components.py
dannon: If this is a literal checkbox there was an issue with a chrome update that changed the way click() was executed; let me see if I can dig up a commit.

Ahh, it was a radio button, but potentially the same underlying problem?  : https://github.com/galaxyproject/galaxy/commit/8d2d8bbc4d2380f4e3745b7d0de734b166a5dc5d  (`click()` doesn't trigger the change, had to use `setChecked()`)

jmchilton:
@OlegZharkov Yeah - that is a much better idea and seems to work better. I saw your comment before the actual suggestions so I committed directly.

OlegZharkov:
> @OlegZharkov Yeah - that is a much better idea and seems to work better. I saw your comment before the actual suggestions so I committed directly.

no worries, glad to remove `js_hack` :sweat_smile:

OlegZharkov (approved):
thanks a lot

PR #10208, opened 2020-09-10
Abstraction for purging the dataset quota usage from a user.

jmchilton:
Deduplicates the if and the arg call - but the amount of de-duplicated code would be increased in https://github.com/galaxyproject/galaxy/pull/4840/files where there is a second if based on hda.dataset.has_active_storage_media.


dannon (approved):

dannon:
Unrelated test failures.

PR #9077, opened 2019-12-06
Support for BioCompute Object Export 

HadleyKing:
Add API to export a BioCompute Object for a workflow invocation. 
Currently this API is based on the [1.4.0](https://github.com/biocompute-objects/BCO_Specification/tree/1.4.0) of the BioCompute specification, which is the next scheduled release.

PR #10217, opened 2020-09-14
Biocompute Object export cleanup.

jmchilton:
See review of https://github.com/galaxyproject/galaxy/pull/9077/files

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/workflows.py
+++ b/lib/galaxy/webapps/galaxy/api/workflows.py
@@ -1005,12 +1005,6 @@ class WorkflowsAPIController(BaseAPIController, UsesStoredWorkflowMixin, UsesAnn
 
         # pull in the user info from those who the history and workflow has been shared with
         contributing_users = [stored_workflow.user]
-        for association in trans.sa_session.query(model.StoredWorkflowUserShareAssociation):
-            if association.user not in contributing_users:
-                contributing_users.append(association.user)
-        for h_association in trans.sa_session.query(model.HistoryUserShareAssociation):
-            if h_association.user not in contributing_users:
-                contributing_users.append(h_association.user)
 
         # may want to extend this to have more reviewers.
         reviewing_users = [stored_workflow.user]
--- a/lib/galaxy/webapps/galaxy/api/workflows.py
+++ b/lib/galaxy/webapps/galaxy/api/workflows.py
@@ -1012,26 +1012,16 @@ class WorkflowsAPIController(BaseAPIController, UsesStoredWorkflowMixin, UsesAnn
 
         spec_version = kwd.get('spec_version', 'https://w3id.org/ieee/ieee-2791-schema/2791object.json')
 
-        # listing the versions of the workflow for 'version' and 'derived_from'
-        versions = []
         for i, w in enumerate(reversed(stored_workflow.workflows)):
-            version = {
-                'version': i,
-                'update_time': str(w.update_time),
-                'steps': w.step_count,
-                'version_id': trans.security.encode_id(w.stored_workflow.id),
-            }
-            versions.append(version)
             if workflow == w:
-                break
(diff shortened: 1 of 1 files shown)


lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -1119,24 +1110,24 @@ def _generate_invocation_bco(self, trans, invocation_id, **kwd):
...
-                current_tool = dict_workflow['steps'][str(i)]
+                current_step = dict_workflow['steps'][str(i)]
                 pipeline_step = {
                     'step_number': i,
-                    'name': current_tool['name'],
-                    'description': current_tool['annotation'],
-                    'version': current_tool['tool_version'],
+                    'name': current_step['name'],
+                    'description': current_step['annotation'],
+                    'version': current_step['tool_version'],
```
mvdbeek, line 1111: I don't think steps have a tool_version, maybe skip this ? That should be what causes https://jenkins.galaxyproject.org/job/docker-api-py3/5470/testReport/junit/lib.galaxy_test.api.test_workflows/WorkflowsApiTestCase/test_run_subworkflow_simple/

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -1005,34 +1003,27 @@ def _generate_invocation_bco(self, trans, invocation_id, **kwd):
...
                 'version': i,
                 'update_time': str(w.update_time),
-                'steps': len(w.steps),
+                'steps': w.step_count,
                 'version_id': trans.security.encode_id(w.stored_workflow.id),
             }
             versions.append(version)
+            if workflow == w:
+                break
         current_version = versions[-1]['version']
```
nsoranzo: I think lines 1016-1027 can be greatly simplified to:
```python
        current_version = None
        for i, w in enumerate(reversed(stored_workflow.workflows)):
            if workflow == w:
                current_version = i
```

nsoranzo:
Ping @HadleyKing

HadleyKing (commented):
RE: https://github.com/galaxyproject/galaxy/pull/10217/commits/6e790ae2048204093b00878fb31a1c9e36f7afc8 
Is there a replacement or are we waiting for a UI to complete these fields? Would that replacement be something that is an addition to the workflow creation/editing?

lib/galaxy/webapps/galaxy/api/workflows.py
HadleyKing: From the little that I have seen the following line: https://github.com/galaxyproject/galaxy/blob/2f2923d4043dc305a664cb80c91d0a74e3dbec66/lib/galaxy/webapps/galaxy/api/workflows.py#L1077
caused this to work in the limited tests I was able to run. Even pulling in a `None` some times, but I guess that was because I was always using tools from the Tool Shed. For reproducibility the version is important. Do you think we could  find some workaround or alternate solution as opposed to skipping it?

HadleyKing:
> RE: [6e790ae](https://github.com/galaxyproject/galaxy/commit/6e790ae2048204093b00878fb31a1c9e36f7afc8)
> Is there a replacement or are we waiting for a UI to complete these fields? Would that replacement be something that is an addition to the workflow creation/editing?

Never mind. I saw this before seeing the other conversation.

lib/galaxy/webapps/galaxy/api/workflows.py
mvdbeek: My bad, steps may have a tool_version in the exported workflow dict (I'd personally would have worked with the workflow, but that's fine for now), but not if the step is not a tool step. The order of the steps was not correct though, so when you pick out the workflow_dict step using the index (i) of the workflow step you may or may not get the step you think you're getting, so if you pick a non tool step this errors out. I've fixed that in the last commit.

mvdbeek (approved):

PR #10219, opened 2020-09-14
[20.05] Backport kombu update -- includes a bugfix for threadsafe queue instantiation

dannon:


nsoranzo (approved):

nsoranzo (approved):

abretaud:
It seems like updating amqp is needed too, I had a report of this error at startup on another server:

```
 kombu 4.6.11 requires amqp<2.7,>=2.6.0, but you'll have amqp 2.5.2 which is incompatible.
```

PR #10220, opened 2020-09-14
Strip deprecated dir from all paths in comma-separated option values

nsoranzo:
Fix the following traceback in BioBlend testing:

```
WARNING:galaxy.config:Paths for the 'tool_config_file' option are now relative to 'config_dir', remove the leading 'config/' to suppress this warning: config/tool_conf.xml.sample,config/shed_tool_conf.xml,test/functional/tools/samples_tool_conf.xml
WARNING:galaxy.config:Paths for the 'tool_config_file' option should be relative to 'config_dir'. To suppress this warning, move 'tool_config_file' into 'config_dir', or set it's value to an absolute path.
  File "/home/travis/galaxy-dev/lib/galaxy/webapps/galaxy/buildapp.py", line 48, in app_factory
    app = galaxy.app.UniverseApplication(global_conf=global_conf, **kwargs)
  File "/home/travis/galaxy-dev/lib/galaxy/app.py", line 77, in __init__
    self.config.check()
  File "/home/travis/galaxy-dev/lib/galaxy/config/__init__.py", line 920, in check
    raise ConfigurationError("Tool config file not found: %s" % path)
galaxy.exceptions.ConfigurationError: Tool config file not found: /home/travis/galaxy-dev/config/config/shed_tool_conf.xml
```

See https://travis-ci.org/github/galaxyproject/bioblend/jobs/726994011#L1249

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -276,14 +276,17 @@ class BaseAppConfiguration:
         def strip_deprecated_dir(key, value):
             resolves_to = self.schema.paths_to_resolve.get(key)
             if resolves_to:  # value is a path that will be resolved
-                first_dir = value.split(os.sep)[0]  # get first directory component
-                if first_dir == self.deprecated_dirs.get(resolves_to):  # first_dir is deprecated for this option
-                    ignore = first_dir + os.sep
-                    log.warning(
-                        "Paths for the '%s' option are now relative to '%s', remove the leading '%s' "
-                        "to suppress this warning: %s", key, resolves_to, ignore, value
-                    )
-                    return value[len(ignore):]
+                paths = value.split(',')
+                for i, path in enumerate(paths):
+                    first_dir = path.split(os.sep)[0]  # get first directory component
+                    if first_dir == self.deprecated_dirs.get(resolves_to):  # first_dir is deprecated for this option
+                        ignore = first_dir + os.sep
+                        log.warning(
+                            "Paths for the '%s' option are now relative to '%s', remove the leading '%s' "
+                            "to suppress this warning: %s", key, resolves_to, ignore, path
+                        )
+                        paths[i] = path[len(ignore):]
+                return ','.join(paths)
             return value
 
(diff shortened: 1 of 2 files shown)


lib/galaxy/config/__init__.py
```diff
@@ -276,14 +276,17 @@ def convert_datatype(key, value):
...
-                first_dir = value.split(os.sep)[0]  # get first directory component
-                if first_dir == self.deprecated_dirs.get(resolves_to):  # first_dir is deprecated for this option
-                    ignore = first_dir + os.sep
-                    log.warning(
-                        "Paths for the '%s' option are now relative to '%s', remove the leading '%s' "
-                        "to suppress this warning: %s", key, resolves_to, ignore, value
-                    )
-                    return value[len(ignore):]
+                paths = value.split(',')
+                for i, path in enumerate(paths):
```
jdavcs, line 280:
```suggestion
                for i, path in enumerate(paths):
                    path = path.strip()
```

jdavcs: We need the strip here, otherwise it'll only work with comma-separated values that have no white space.
(or just add to a list comprehension in `paths = value.split(',')`

Also, would you mind adding a unit test? I found this by adding this to `test/unit/config/test_path_resolves_to.py`:
```python
def test_kwargs_relative_path_old_prefix_csv_value(mock_init):
    config = BaseAppConfiguration(path4='old-config/foo/file1, old-config/foo/file2,  foo/file3')
    assert config.path4 == ['my-config/foo/file1','my-config/foo/file2', 'my-config/foo/file3']
```

jdavcs (changes requested):

lib/galaxy/config/__init__.py
nsoranzo: Both done, thanks @ic4f !

jdavcs (approved):
Thank you, @nsoranzo!
I'll restart the failing tests, although I think in both cases they are unrelated.

ic4f:
Failing tests unrelated.

PR #10222, opened 2020-09-15
Store job_id on Dataset table. 

jdavcs:
Ref: #9934 (that also contains reasoning for storing job id on `Dataset` vs. `DatasetInstance`)
(Previous db trigger version: #9515; also ref: #7791)

EDIT: see issues #8704, #9727.

This is ready for review. It is meant to record the id of the job that creates a dataset in the dataset table. It should handle the creation of all datasets. I've added tests that cover the creation of a single dataset, multiple datasets, and a discovered dataset. Each case is represented by one tool. I did not write a test for uploaded datasets (which needs to be added eventually), but I've verified those manually.

Misc. notes:
- My first version that worked was 087ed92 . It  handled discovered datasets in a [very awkward](https://github.com/galaxyproject/galaxy/blob/087ed922ecc7844a214c3cd16621e8e66a172081/lib/galaxy/model/store/discover.py#L96) way. The current version improves on it.
- I've added a `get_job_id()` method to `JobContext` objects because I couldn't use the `job_id()` method without changing it: 
-- for `JobContext` it returns [`self.job.id`](https://github.com/galaxyproject/galaxy/blob/release_20.09/lib/galaxy/job_execution/output_collect.py#L296); 
-- for `SessionlessJobContext` it returns a string: [`non-session bound job`](https://github.com/galaxyproject/galaxy/blob/release_20.09/lib/galaxy/job_execution/output_collect.py#L334); however [here](https://github.com/galaxyproject/galaxy/blob/release_20.09/lib/galaxy/job_execution/output_collect.py#L355) it retrieves the correct `job id`.
I don't think both methods should live side-by-side, but I don't understand those classes well enough to modify `job_id()`.

Ping @jmchilton, @mvdbeek, @natefoo.

If this is correct, I'll write a script to handle updating of existing dataset records in the database. 

To do:
- [x] handle one dataset per job
- [x] handle multiple datasets per job
- [x] add integration test
- [x] handle uploads
- [x] handle discovered datasets
- [ ] update integration test to include uploads
- [x] update integration test to include discovered datasets
- [x] refactor: this needs a better abstraction (partially improved, so check)


ic4f:
Some tests seemed to fail consistently (remote only). Investigating. Running the tests on the first commit only.

test/integration/test_save_job_id_on_datasets.py
```diff
@@ -0,0 +1,40 @@
...
+    framework_tool_and_types = True
+
+
+@pytest.mark.parametrize('tool_id', TEST_TOOL_IDS)
+def test_tool_datasets(tool_id):
+    driver = GalaxyTestDriver()
+    driver.setup(TestConfig)
+    driver.run_tool_test(tool_id)
+    session = driver.app.model.context.current
+    job = session.query(model.Job).one()
```
ic4f, line 33: can we use `decode_id()` to get the right job from the db? Then we can reuse the driver. And improve this.

test/integration/test_save_job_id_on_datasets.py
```diff
@@ -0,0 +1,40 @@
...
```
ic4f, line 33: The database is not created for this one test, so this won't work: there are multiple jobs - hence the error.

test/integration/test_save_job_id_on_datasets.py
```diff
@@ -0,0 +1,40 @@
...
+    datasets = session.query(model.Dataset).all()
```
ic4f, line 34: this won't work either for same reason.

ic4f:
So, locally this test passes when run not as a standalone.

ic4f:
This works as of 087ed92. Refactoring before marking as ready for review.

lib/galaxy/job_execution/output_collect.py
```diff
@@ -174,6 +174,9 @@ def find_files(self, output_name, collection, dataset_collectors):
             filenames[discovered_file.path] = discovered_file
         return filenames
 
+    def get_job_id(self):
+        return None  # overwritten in subclassews
```
jdavcs:
```suggestion
        return None  # overwritten in subclasses
```

lib/galaxy/tools/execute.py
```diff
@@ -91,6 +96,7 @@ def execute_single_job(execution_slice, completed_job):
     jobs_executed = 0
     has_remaining_jobs = False
     execution_slice = None
+    job_datasets = {}  # job: list of dataset instances created by job
```
jmchilton, line 103: What is the reason for just not setting this above when building the dictionary?

jmchilton:
This definitely feels like it is on the right track relative to the database trigger. Thanks for working on this @ic4f!

lib/galaxy/tools/execute.py
jdavcs: @jmchilton  Do you mean why it's not initialized inside the `execute_single_job` function, after line 73? 
We init it on line 99, after that `execute_single_job` may be called (line 106) where `job_datasets` may be populated (line 76), after which, back in the `execute` method it may be iterated over (line 118). 
Sorry if I misunderstood the question.

ic4f:
> This definitely feels like it is on the right track relative to the database trigger. Thanks for working on this @ic4f!

@jmchilton, @mvdbeek thank you for looking through this!

ic4f:
TODO: This needs an updated migration number.

ic4f:
Need to fix all those tests... WIP for a short while.

ic4f:
Rebased and green! @mvdbeek

lib/galaxy/model/mapping.py
```diff
@@ -272,6 +272,7 @@
 model.Dataset.table = Table(
     "dataset", metadata,
     Column("id", Integer, primary_key=True),
+    Column('job_id', Integer, ForeignKey('job.id'), index=True),
```
mvdbeek:
```suggestion
    Column('job_id', Integer, ForeignKey('job.id'), index=True, nullable=True),
```

mvdbeek: Can you setup the relationship as well, you can just do `dataset.job = job`, which means you don't need to wait for a flush before you can set the job id. I think that should make some of the code below simpler.

test/integration/test_save_job_id_on_datasets.py
```diff
@@ -0,0 +1,42 @@
+"""
```
mvdbeek, line 1: Can you add a basic test somewhere around there as well ? https://github.com/mvdbeek/galaxy/blob/65479c5cd6981ee6d86077a38369af2c0421d4a2/test/unit/data/test_galaxy_mapping.py#L382

mvdbeek (commented):
Looking nice! Can you add some logic to https://github.com/mvdbeek/galaxy/blob/65479c5cd6981ee6d86077a38369af2c0421d4a2/lib/galaxy/model/__init__.py#L3086 as well ? That'd be an immediate win over the kind-of expensive loop we're doing there.

ic4f:
> Looking nice! Can you add some logic to https://github.com/mvdbeek/galaxy/blob/65479c5cd6981ee6d86077a38369af2c0421d4a2/lib/galaxy/model/__init__.py#L3086 as well ? That'd be an immediate win over the kind-of expensive loop we're doing there.

@mvdbeek It looks like for this we would need to store the creating job_id for a dataset **instance**, right? Because here we only save the id of the job that created the Dataset, not HDAs, etc. If that is the case, I need to rethink *all of this* from scratch.

mvdbeek (approved):

lib/galaxy/tools/execute.py
```diff
@@ -110,6 +116,12 @@ def execute_single_job(execution_slice, completed_job):
     else:
         # Make sure collections, implicit jobs etc are flushed even if there are no precreated output datasets
         trans.sa_session.flush()
+
+    if job_datasets:
+        for job, datasets in job_datasets.items():
+            for dataset_instance in datasets:
+                dataset_instance.dataset.job_id = job.id
```
mvdbeek:
```suggestion
    if job_datasets:
        for job, datasets in job_datasets.items():
            for dataset_instance in datasets:
                dataset_instance.dataset.job = job
```
ideally we haven't flushed the job yet at this point, so it's better to just add the job object directly and let the session resolve this during the flush

lib/galaxy/tools/execute.py
```diff
@@ -72,6 +72,11 @@ def execute_single_job(execution_slice, completed_job):
         if job:
             log.debug(job_timer.to_str(tool_id=tool.id, job_id=job.id))
             execution_tracker.record_success(execution_slice, job, result)
+            # associate dataset instances with the job that creates them
+            if result:
+                datasets = [pair[1] for pair in result if type(pair[1]) == model.HistoryDatasetAssociation]
```
natefoo: These could also be LDDAs I think?

natefoo (approved):

mvdbeek (commented):
.... now I understand the confusion, forgot to submit this

lib/galaxy/tools/execute.py
jdavcs: Yes. Truth be told, I think I was so afraid of these edits breaking something, I was moving half a baby step at a time. The LDDAs here would be the second half. Let me complete that one :)

test/integration/test_save_job_id_on_datasets.py
jdavcs: Do you mean a test of the dataset/job relationship? Like
`create dataset, create job, set dataset.job = job, then persist, then load dataset by id, then assert that dataset.job_id == job.id`? I'll commit this in a min - let me know if you wanted me to test something else.

ic4f:
> > Looking nice! Can you add some logic to https://github.com/mvdbeek/galaxy/blob/65479c5cd6981ee6d86077a38369af2c0421d4a2/lib/galaxy/model/__init__.py#L3086 as well ? That'd be an immediate win over the kind-of expensive loop we're doing there.
> ...
> EDIT: ...Unless we could just do:
> ```
> @property
> def creating_job(self):
>     return self.dataset.job_id
> ```
> thus getting rid of that recursion, unless setting that property _is_ the way to load `self.creating_job_associations` (via side effect of course, by recursivelly calling `_source_dataset_chain`) ..I have a strong feeling something is waiting to be untangled here.

@mvdbeek : I did it. So Now it's not triggering the `source_dataset_chain` derivation, but that it triggered anytime it's touched. And it's touched only in [one more place](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/controllers/dataset.py#L911) - which, I think, is used *for the same purpose* - to get the job that created the dataset instance. If this is correct, I suppose we can get rid of `source_dataset_chain` by modifying the dataset controller to just do `job = hda.dataset.job` - does this make sense? (if so, I'll do a separate PR for that).

lib/galaxy/tools/execute.py
jdavcs: Done.

ic4f:
I've reverted the last change: the edits caused 8 unit test failures (and, possibly integration test failures too), but, I think, that was due to the test code (which needs to be revised). I'll reapply the edits in a follow-up PR, together with revised tests.

lib/galaxy/model/__init__.py
```diff
@@ -3088,6 +3090,7 @@ def _source_dataset_chain(dataset, lst):
 
     @property
     def creating_job(self):
+        # TODO this should work with `return self.dataset.job` (revise failing unit tests)
```
mvdbeek, line 3093: Only once we have a migration script that adds this. Before that you'd want to do `if self.dataset.job: return self.dataset.job` or something like that, then run through the rest.

mvdbeek (approved):

lib/galaxy/model/__init__.py
jdavcs: Thanks! I'll try this.

PR #10225, opened 2020-09-15
[20.05] Decrease pysam verbosity

mvdbeek:
Can be reverted once
https://github.com/pysam-developers/pysam/issues/939 is fixed


mvdbeek:
Since this only affects bam or cram indexes I have only applied it where we open these files, and not the tools or converters that deal with tabix or other formats.

natefoo:
Maybe worth putting in a comment so we remember why this is set?

dannon (approved):

PR #10226, opened 2020-09-15
Improve job error logging

mvdbeek:
If available, report the failure reason.
Also deduplicates redundant log messages.

With these changes there will be just one log message reporting why the job has failed.
For an exit code condition that will look like this:
```
galaxy.tool_util.output_checker DEBUG 2020-09-15 17:24:12,236 Job error detected, failing job. Reasons are [{'type': 'exit_code', 'desc': 'Fatal error: Exit code 127 (Failing exit code.)', 'exit_code': 127, 'code_desc': 'Failing exit code.', 'error_level': 3}]
```
For a match on stdout this is
```
galaxy.tool_util.output_checker DEBUG 2020-09-15 17:43:14,133 Job error detected, failing job. Reasons are [{'type': 'regex', 'stream': 'stdout', 'desc': 'Fatal error: ', 'code_desc': '', 'match': 'ERROR:', 'error_level': 3}]
```

### DIFF
--- a/lib/galaxy/tool_util/output_checker.py
+++ b/lib/galaxy/tool_util/output_checker.py
@@ -96,7 +96,6 @@ def check_output(stdio_regexes, stdio_exit_codes, stdout, stderr, tool_exit_code
                             'code_desc': code_desc,
                             'error_level': stdio_exit_code.error_level,
                         }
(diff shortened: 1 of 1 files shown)


natefoo:
Too late, but this should probably be at least an `INFO` level log.

PR #10231, opened 2020-09-16
Pass user to the ObjectStorePopulator

VJalili:
This code is taken from https://github.com/galaxyproject/galaxy/pull/4840 in order to simplify that branch for the review process. 

@jmchilton's message: 
> Not used yet in Galaxy core yet, but useful for applications where you want object store selection to be based on user in some way. This code was taken from #4840. Part of this is trying to reduce the number of files that branch touches to make review easier - but I'm confident this extension point is good regardless. Also it makes it clear we need to keep the user object in the picture when assigning the object store ID in the future.


jmchilton:
Awesome - thanks!

PR #10233, opened 2020-09-16
User-facing objectstore metadata.

jmchilton:
#6552 implemented the ability for admins to assign job outputs to different object stores at runtime and #10221 implements the ability to assign quotas and perform tracking for different object stores and groups of object stores. So for instance, user preferences or injected tool resource parameters can be used to create files in different object stores and disk usage and quota limits will be tracked and can be used in that calculation.

Still the user has no way of knowing what data went where or what that means for their analysis. Beyond this retrospective use case, as we look forward to user-based selections - we are not currently tracking metadata that might help a user make selections about which object store to use. This PR adds the ability for admins to annotate objectstores with more metadata - enough to solve this first problem and provide some infrastructure for solving that selection problem in the future.

Concrete ObjectStores (defined in either XML or YAML) can now be named and can have a rich Markdown description. A dataset API endpoint has been added that exposes this metadata for a particular existing dataset and this information will now be rendered as part of the dataset info page.

Here are two examples - illustrating how admins might highlight long term, backed up storage and short term scratch storage. They also demonstrate the power of allowing admins to define these annotations using Markdown.

Long Term Storage:

Screen Shot 2020-09-16 at 2 05 25 PM

Short Term Storage:

Screen Shot 2020-09-16 at 2 15 11 PM

### DIFF
--- a/client/src/components/Dataset/DatasetStorage/DatasetStorage.vue
+++ b/client/src/components/Dataset/DatasetStorage/DatasetStorage.vue
@@ -7,9 +7,12 @@
                 This data is stored in a Galaxy ObjectStore named <b>{{ storageInfo.name }}</b
                 >.
             </p>
-            <p v-if="storageInfo.object_store_id">
+            <p v-else-if="storageInfo.object_store_id">
                 This data is stored in a Galaxy ObjectStore with id <b>{{ storageInfo.object_store_id }}</b
                 >.
+            </p>
+            <p v-else>
+                This data is stored in the default configured Galaxy ObjectStore.
             </p>
             <div v-html="descriptionRendered"></div>
         </div>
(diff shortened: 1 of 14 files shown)


client/src/components/Dataset/DatasetStorage/DatasetStorage.vue
```diff
@@ -0,0 +1,69 @@
+<template>
+    <div>
+        <h3 v-if="includeTitle">Dataset Storage</h3>
+        <loading-span v-if="storageInfo == null"> </loading-span>
+        <div v-else>
+            <p v-if="storageInfo.name">
+                This data is stored in a Galaxy ObjectStore named <b>{{ storageInfo.name }}</b
+                >.
+            </p>
+            <p v-if="storageInfo.object_store_id">
```
mvdbeek: If both name and id are defined this seems to take up more screen space than needed, maybe you could combine that into one sentence (or a small table?)?

mvdbeek (approved):
Very cool!

client/src/components/Dataset/DatasetStorage/DatasetStorage.vue
jmchilton: This was mean t to be an else-if I think. Thanks for the catch!

Issue #9879, opened 2020-06-15
Convert Collection Builders to VueJS


jmchilton:
I think these are all in scripts/galaxy/mvc/collection

- base-creator.js
- list-collection-creator.js
- list-of-pairs-collection-creator.js
- pair-collction-creator.js

assuntad23:
Closed with #10240

Issue #9589, opened 2020-04-06
Create a collection of paired datasets: put DEFAULT_FILTERS into Galaxy config, add: dot12s: [".1.fastq", ".2.fastq"]


tshtatland:
Our users have *.1.fastq and *.2.fastq paired end reads. They typically use collections to analyze these data. We need to enable easier filter selection for creating collections.

The typical use case is a follows. The user goes to History tab, selects fastq files, goes to:
```
For all selected | Build list of dataset pairs
Choose filters
Choose from the following filters to change which unpaired reads are shown in the display:
Forward: _1, Reverse: _2
Forward: _R1, Reverse: _R2
Change to:
Forward: .1.fastq, Reverse: .2.fastq
(note the leading dot)
```
I suggest to control the defaults (`DEFAULT_FILTERS: "illumina"`) using a Galaxy config parameter. This would allow the admin to select the most commonly used extensions as defaults.  Also add to the list commonFilters this very commonly used pattern: `dot_fastqs: [".1.fastq", ".2.fastq"],`, see [`client/galaxy/scripts/mvc/collection/list-of-pairs-collection-creator.js`](https://github.com/galaxyproject/galaxy/blob/d7921ae4dda5388dfda1f346a6a3b5ca7436a9ee/client/galaxy/scripts/mvc/collection/list-of-pairs-collection-creator.js):
```
        /** map of common filter pairs by name */
        commonFilters: {
            illumina: ["_1", "_2"],
            Rs: ["_R1", "_R2"],
        },
        /** which commonFilter to use by default */
        DEFAULT_FILTERS: "illumina",

```

bernt-matthias:
@tshtatland would you like to open a pull request, since you already found the place in the code?

jmchilton:
https://github.com/galaxyproject/galaxy/pull/10212/commits/b628309c80d7d7188bd9d4dc40a0ebcae1d59908 is a good example I just did of how to get a simple config option from Galaxy's YAML to the frontend and work with it. That might make this easier.

assuntad23:
completed as part of PR #10240

lldelisle:
I think this is still interesting to have `COMMON_FILTERS` and potentially `commonFilters` as galaxy config parameter.
https://github.com/galaxyproject/galaxy/blob/d7921ae4dda5388dfda1f346a6a3b5ca7436a9ee/client/galaxy/scripts/mvc/collection/list-of-pairs-collection-creator.js#L288-L294

[...]

PR #10240, opened 2020-09-17
Converting Collections to Vue

assuntad23:


client/src/mvc/collection/ListCollectionCreatorModal.vue
```diff
@@ -0,0 +1,52 @@
...
+import _l from "utils/localization";
+import Vue from "vue";
+export default {
+    mixins: [CollectionCreatorModalMixin],
+    methods: {
+        listCollectionCreatorModal: function (elements, options) {
+            options = options || {};
+            options.title = _l("Create a collection from a list of datasets");
+            const { deferred, creatorOptions, showEl } = this.collectionCreatorModalSetup(options); // eslint-disable-line no-unused-vars
+            return import(/* webpackChunkName: "ListCollectionCreator" */ "./ListCollectionCreator.vue").then(
```
jmchilton: Import at the top file like any other component. Then just do a normal Vue.extend.

dannon: Is this actually breaking?  This isn't unreasonable if the ListCollectionCreator component is medium to large (including dependencies) and/or not used very frequently.

jmchilton: I just assumed since it didn't have external dependencies relative to the rest of the app this wasn't needed. Rule builder made sense.

jmchilton:
Screen Shot 2020-12-10 at 9 47 49 AM

It breaks down if you create a duplicated identifier. Any chance you can fix the error handling around that?

assuntad23:
This PR is to resolve issues:  #9879 and #9589

client/src/components/Collections/RuleBasedCollectionCreatorModal.js
```diff
@@ -0,0 +1,118 @@
...
+            return selection.createHDCA(elements, collectionType, name, hideSourceItems, copyElements);
+        },
+    });
+    return promise;
+}
+
+export default {
+    ruleBasedCollectionCreatorModal: ruleBasedCollectionCreatorModal,
+    createCollectionViaRules: createCollectionViaRules,
+};
```
jmchilton, line 91: This makes a lot more sense as functions than the previous Vue components, thanks for adjusting this.

lib/galaxy_test/selenium/test_collection_builders.py
```diff
@@ -69,10 +68,12 @@ def test_build_paired_list_simple(self):
         self.collection_builder_set_name("my awesome paired list")
         self.screenshot("collection_builder_paired_list")
         self.collection_builder_create()
-        self.history_panel_wait_for_hid_ok(5)
+        self.history_panel_wait_for_hid_ok(3)
+        self.history_panel_wait_for_hid_hidden(1)
+        self.history_panel_wait_for_hid_hidden(2)
```
jmchilton, line 73: Awesome - thanks for adjusting this. Makes good sense to me!

lib/galaxy_test/selenium/test_collection_builders.py
```diff
@@ -92,16 +93,15 @@ def test_build_paired_list_hide_original(self):
         second_dataset_reverse = reverse_column.find_elements_by_css_selector("li")[1]
         second_dataset_reverse.click()
 
+        # this toggles the checkbox to not hide originals
         self.collection_builder_hide_originals()
         self.collection_builder_set_name("my awesome paired list")
 
         self.collection_builder_create()
-        self.history_panel_wait_for_hid_ok(3)
```
jmchilton, line 99: We should add some more assertions here - they seem to work in my testing:

https://gist.github.com/jmchilton/276b8a1deb9763d9acb4fc29c667f229

But we can wait until we merge this and do that in a new PR.

lib/galaxy_test/selenium/test_uploads.py
```diff
@@ -90,7 +90,7 @@ def test_upload_pair(self):
     @selenium_test
     def test_upload_pair_specify_extension(self):
         self.upload_list([self.get_filename("1.tabular"), self.get_filename("2.tabular")], name="Test Pair", ext="txt", hide_source_items=False)
```
jmchilton, line 92: If hide_source_items is False, this needs to be 3 right?

Is hide_source_items on ``upload_list`` broken maybe?

jmchilton: No - I guess you're right. If you don't hide the source items you want to copy them... so it should be 5. You see - this is why you can't trust me to answer this question 😆

assuntad23: :smile: it's confusing and counter-intuitive, even though it makes sense when you think about it.

lib/galaxy_test/selenium/test_collection_builders.py
assuntad23: fixed it with the prettier re-run :)

jmchilton:
We could spend forever iterating on this, but lets ship it. Very nice work Assunta!

bgruening:
@assuntad23 that is awesome. Thanks a lot!

PR #10241, opened 2020-09-18
Create Page from Invocation Report

### LINKED ISSUE #10238: Publish Invocation Report to a Page
@guerler and I are on the same page that it would be awesome if you could publish a report to a page. I wanted to create an issue just to outline the backend because that is important to me.

I think it should maybe be:

```
/api/invocations/<id>/report/publish
```

I'd like it to go through the backend instead of through the client so we can leverage the functionality outside the webui (i.e. from future post workflow actions for instance or CLI pipelines). The backend also has functionality that will turn the invocation referenced objects into objects referenced by database ID which is what the pages need - so I think that will work better than translating them to encoded IDs and rewriting the page from those (although that should also work fine I suppose).

guerler:
Invocation reports are not mutable but we can use their markdown content to create a page which can then be edited. This PR redirects the edit button in an Invocation Report to the Page creation form. The Page creation endpoint has been modified such that parsing an Invocation ID will populate the form with details from the corresponding Invocation Report. This PR also removes a temporary mako introduced in a recent changeset.

Resolves #10238.


lib/galaxy/webapps/galaxy/controllers/page.py
```diff
@@ -324,15 +326,32 @@ def create(self, trans, payload=None, **kwd):
         Create a new page.
         """
         if trans.request.method == 'GET':
+            form_title = "Create new Page"
+            title = ""
+            slug = ""
+            content = ""
+            content_format_hide = False
+            content_hide = True
+            if "invocation_id" in kwd:
```
jmchilton, line 335: All this is great but this part makes me a little uneasy, it is on some level what I was trying to avoid since there isn't an API endpoint for generating the page from the invocation but I suppose it is more or less all staying on the backend minus a quick trip in a hidden textarea. 

I'll approve this PR but I would be more excited if you just stuck the ``invocation_id`` in the form here and then passed it to the page creation API as a parameter - maybe like this ``"content_from": {"src": "workflow_invocation", "id": invocation_id}``. And then did this generation in this if block in the page manager on the backend during page creation.

I think the resulting API would be more usable use non-client uses. Since all the API consumer would need would be the invocation ID and they wouldn't need to know how to use the API that generates the report and they wouldn't need to download it and re-submit it in the request.

guerler: I understand and I see your point. Thanks for the detailed explanation. I think we might need both i.e. an augmented API endpoint (which takes the `Invocation ID` and populates the Markdown) and this API-like controller endpoint, since fully automated generation of a page from the `Invocation ID` can fail if the report Markdown contains errors. Those errors cannot be caught by the Workflow editor but become apparent after completing the invocation. I chose this strategy since it displays a potential error with an alert in the page creation form and allows users to modify the imported Markdown content before creating the page.

guerler: Adding this feature to the page manager such that it becomes available in the regular page api endpoint would probably look like this: https://github.com/galaxyproject/galaxy/commit/89f90d3255d73eba2adda404fe1c6da4c7236535

jmchilton: Excellent point about catching error! I like 89f90d3 also.

guerler: Great, I added it to this PR and adding a test case for the API now. Thanks for your comments.

jmchilton:
Awesome - thanks so much!

Issue #10235, opened 2020-09-17
(Cached?) Singularity resolver does not pick newest builds


mvdbeek:
```
Checking with container resolver [CachedMulledSingularityContainerResolver[cache_directory=/cvmfs/singularity.galaxyproject.org/all/]] found description [ContainerDescription[identifier=/cvmfs/singularity.galaxyproject.org/all/fastqc:0.11.8--0,type=singularity]]
```
even though there is `/cvmfs/singularity.galaxyproject.org/all/fastqc:0.11.8--2`

PR #10242, opened 2020-09-18
[20.05] Make cached container resolver pick newest build

torfinnnome:
Traversing the cached container list in reverse should pick up the latest version. Resolves https://github.com/galaxyproject/galaxy/issues/10235.

### DIFF
--- a/lib/galaxy/tool_util/deps/container_resolvers/mulled.py
+++ b/lib/galaxy/tool_util/deps/container_resolvers/mulled.py
@@ -154,7 +154,7 @@ def find_best_matching_cached_image(targets, cached_images, hash_func):
     image = None
     if len(targets) == 1:
         target = targets[0]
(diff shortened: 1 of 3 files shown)


mvdbeek:
This probably just relies on whatever sorting the local filesystem implements, and that sorting may or may not match the versioning scheme that is being used for images. You'd probably need to sort the incoming `cached_images` list first, and I think you can do that with https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tool_util/deps/mulled/util.py#L134

mvdbeek:
Turns out this was a little trickier than I thought to get right. If https://github.com/torfinnnome/galaxy/pull/1 works for you can you merge it ? I think that should resolve the issue.

mvdbeek (approved):

PR #10248, opened 2020-09-19
[20.09] Logout fix and refactoring

dannon:
Followup to #10107, fixes logout (in user delete functionality *and* userpreferences), with a little accompanying refactoring.  This shouldn't ultimately live in layout/menu, but I'm going to leave it there for 20.09.

Also fixes another bug where we said we were logging out all sessions, but were in fact only doing the live one.


mvdbeek:
Looks like the selenium test needs a small adjustment ?

mvdbeek (approved):

dannon:
@mvdbeek Thanks, didn't see that one before and it should be fixed.  The test case just needed to wait since it's not an immediate hard redirect now.

PR #10252, opened 2020-09-21
Fix ObjectStorePopulator use

mvdbeek:
The user argument is now mandatory.
Broken in https://github.com/galaxyproject/galaxy/pull/10231.


mvdbeek:
That was vscodes fault ...

PR #10253, opened 2020-09-21
[20.05] allow validation of ITs

blankenberg:


lib/galaxy/tools/repositories.py
```diff
@@ -29,6 +29,7 @@ def __init__(self, app_name,
         self.temporary_path = tempfile.mkdtemp(prefix='tool_validation_')
         self.config.tool_data_table_config = os.path.join(self.temporary_path, 'tool_data_table_conf.xml')
         self.config.shed_tool_data_table_config = os.path.join(self.temporary_path, 'shed_tool_data_table_conf.xml')
+        self.config.interactivetools_enable = True
```
mvdbeek, line 32: That enabled TS upload, are we sure that the interface is stable now ?

blankenberg: I think it would be reasonable to state so. There has been pretty decent uptake of it currently. I don't think there been altering changes in over a year. Would be good to hear other input though.

Are there any open issues or PRs that would not be compatible?

mvdbeek:
I guess it's fine by me, but can you target the 20.09 branch please ?

blankenberg:
@mvdbeek thanks for looking! I've replaced this with https://github.com/galaxyproject/galaxy/pull/10341 targeting 20.09.

PR #9923, opened 2020-06-26
Python3 cleanup

dannon:
Now that we're officially py3-only, we can clean up the cruft and standardize a few things.

I used some upgrade tooling to do stuff here like swap to a single class style, drop `six` compatibility shims in some places where it was straightforward, standardize print formatting and drop __future__ imports, etc.  There's a good bit more to do, some tweaky cases that we'll want to handle manually, but I wanted to draft this and get tests running.

PR #10258, opened 2020-09-22
Add script to diff two release config files.

hexylena:
Per some discussion with @mvdbeek in gitter. Not integrated or used yet, but, it's a starting point and opening this for discussion.

```
$ .venv/bin/python scripts/release-diff.py v20.05 > doc/source/releases/20.09-config.rst
```

produces [this lovely diff](https://github.com/hexylena/galaxy/blob/release-diff/doc/source/releases/20.09-config.rst)

fyis:

- intentionally doesn't load `!includes` since those can't be followed to the old git revision.

some open questions

- which files should we diff and show? I've done all yml.sample in config.
- Doesn't parse newly added options which aren't added with default values to the config files. We'd have to parse the schemas for that. I didn't have a good feeling for if that would be significantly more useful?

### DIFF
--- a/doc/source/releases/20.09-config.rst
+++ b/doc/source/releases/20.09-config.rst
@@ -1,0 +1,30 @@
+Configuration Changes
+=====================
+
+Changed
+-------
+
+The following configuration options have been changed
+
+config/tool_shed.yml.sample
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+-  uwsgi.static-safe has changed from ``client/galaxy/images`` to ``client/src/assets``
+
+config/galaxy.yml.sample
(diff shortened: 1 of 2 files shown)


mvdbeek:
(But I'm definitely going to use this for the notes)

hexylena:
Sounds good to me! Looking forward to using f-strings, I started a while back and can't live without them.

dannon:
@hexylena Me too.  I even initially converted them all in #9923 and then had to go back and fix it when I realized we still supported 3.5 :/

dannon:
We can use f-strings now!

dannon:
The failing jest test is completely unrelated.  I should probably try to make that more robust though.

PR #10259, opened 2020-09-22
Merge 20.05 into dev

mvdbeek:
Opening a PR because I had to adapt #10237 to the workflow editor changes. I think https://github.com/galaxyproject/galaxy/commit/3f0726daee2ce4c559326fed38c8c10e3d5f0569 is much nicer now actually.


dannon:
Looks like the test failure is legit here

mvdbeek:
Yup, working on the fix

PR #10266, opened 2020-09-22
Filtering.py failing all expressions

almahmoud:
I am not entirely sure that this is the right fix, but with the newest version, all expressions were failing (including the ones in the examples, and the default one). I put print statements around and figured out that they were always failing because of hitting here: https://github.com/almahmoud/galaxy/blob/da860b09c829b804f5efef509882c7b42caa6f2a/tools/stats/filtering.py#L112 with type 'Constant' which was not in the list. This makes it work, but I am not entirely sure of any possible implications of adding 'Constant' to the whitelist if there are any.


nsoranzo (approved):
`ast.Constant` was added in Python 3.6, which is probably why it wasn't noticed before.

dannon:
@almahmoud How about we target 20.09 here?

mvdbeek:
We need to update the Jenkins node to more recent Python versions, that would have at least caught the issue. I think we should probably backport until 20.01 as the oldest supported Galaxy version.

almahmoud:
Is someone already taking care of the backporting or should I make PR's against the old release branches?

dannon:
@almahmoud If you want to open a PR against 20.01 I'll merge it forward through the rest.

PR #10270, opened 2020-09-23
initial commit for gfa2

bernt-matthias:
Wondering if we should enable sniffers for gfa and gfa2?

And we need a test file..


bernt-matthias:
Maybe we can take the test data from https://github.com/ggonnella/gfapy/tree/master/tests/testdata

mvdbeek (approved):

mvdbeek (approved):

Issue #10271, opened 2020-09-23
ToolShed: we can't move hgweb.config file 


FredericBGA:
I tried to move `hgweb.config` from the root directory of Galaxy to the `config` directory.
I tried to change the ToolShed config in `config/tool_shed.yml`:
```
  # Where the hgweb.config file is stored. The default is the Galaxy
  # installation directory.
  hgweb_config_dir: config
```
But it breaks the build of the whoosh index (and I was then unable to install a tool from our ToolShed as they were not found anymore by the api `toolshed/api/repositories`).

All the paths are fine:
```
python scripts/tool_shed/build_ts_whoosh_index.py -c /softs/bioinfo/galaxy-dev/config/tool_shed.yml --config-section tool_shed -d
Full options:
config_file: /softs/bioinfo/galaxy-dev/config/tool_shed.yml
file_path: /softs/bioinfo/galaxy-dev/database/local_toolshed
hgweb_config_dir: /softs/bioinfo/galaxy-dev/config
debug: True
dburi: postgresql://localtools........
config_section: tool_shed
whoosh_index_dir: database/toolshed_whoosh_indexes
```
But the script will try to look for files in a wrong path (there is an extra `config` in it, as `lib/tool_shed/util/shed_index.py` use `hgweb_config_dir` in the join ):
`mercurial.error.RepoError: b'repository /softs/bioinfo/galaxy-dev/config/database/local_toolshed/000/repo_4 not found'`
https://github.com/galaxyproject/galaxy/blob/53217ea908ea6d2410270287a0abb2bea1416edc/lib/tool_shed/util/shed_index.py#L119

mvdbeek:
This is the correct behavior, the entries in hgweb.config have to either be absolute or relative to hgweb_config_dir.
The tool shed should be writing absolute paths in hgweb.config, at least since 20.05. if these are still relative because you haven't added any repositories on 20.05 you'll have to set absolute paths in hgweb.config

PR #10273, opened 2020-09-23
[20.05] Fix hg_repo path handling in shed_index.py

mvdbeek:
Entries are absolute paths anyway (at least since 20.05),
and all paths on the main toolshed and test toolshed are absolute.
This only worked for us because os.path.join does this:
```
Signature: os.path.join(a, *p)
Docstring:
Join two or more pathname components, inserting '/' as needed.
If any component is an absolute path, all previous path components
will be discarded.  An empty last part will result in a path that
ends with a separator.
```

Fixes https://github.com/galaxyproject/galaxy/issues/10271.

### DIFF
--- a/lib/tool_shed/util/shed_index.py
+++ b/lib/tool_shed/util/shed_index.py
@@ -116,7 +116,7 @@ def get_repos(sa_session, file_path, hgweb_config_dir, **kwargs):
         full_last_updated = repo.update_time.strftime("%Y-%m-%d %I:%M %p")
 
         # Load all changesets of the repo for lineage.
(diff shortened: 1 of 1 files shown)


jmchilton:
Unit test failure it legitimate?

mvdbeek:
Yes, though the test is a bit artificial in that I just zipped up hgweb from an existing instance. I'll have a look.

mvdbeek:
So, yeah, the test actually nicely highlights that we can't do this, we've been putting paths relative to hgweb_config_dir into hgweb.config. So if you need to use absolute paths in hgweb.config if you need to place hgweb_config_dir outside of galaxy_root

PR #10279, opened 2020-09-24
Datatypes refactoring

nsoranzo:
- Don't allow datatype change by default if it's composite
- Re-use the HDA deserializer to change datatype in the dataset controller
- Make ``ok_to_edit_metadata()`` a method of ``DatasetInstance``
- Use `super()` to invoke parent class methods in all datatypes
- Remove unnecessarily overridden methods
- Fix issues reported by flake8-bugbear


mvdbeek:
Very nice, thanks a lot @nsoranzo!

PR #10283, opened 2020-09-24
fix lib path quickfix

OlegZharkov:
the path is wrong, if  'enter the folder -> press ".." btn -> enter same folder again'.

Please merge before anyone notices ;)


dannon:
@OlegZharkov Should this target `release_20.09` and not `dev`?

PR #10284, opened 2020-09-24
[20.09] fix folder path

OlegZharkov:
the path is wrong, if 'enter the folder -> press ".." btn -> enter same folder again'.

Please merge before anyone notices ;)

copy of https://github.com/galaxyproject/galaxy/pull/10283 to 20.09 branch

### DIFF
--- a/client/src/mvc/library/library-libraryrow-view.js
+++ b/client/src/mvc/library/library-libraryrow-view.js
@@ -243,7 +243,7 @@ var LibraryRowView = Backbone.View.extend({
                         <td style="color:grey;"><%- library.get("name") %></td>
                     <% } else { %>
                         <td>
-                            <a href="folders/<%- library.get("root_folder_id") %>"><%- library.get("name") %></a>
+                            <a href="/library/folders/<%- library.get("root_folder_id") %>"><%- library.get("name") %></a>
                         </td>
                     <% } %>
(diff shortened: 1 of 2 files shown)


client/src/mvc/library/library-libraryrow-view.js
```diff
@@ -243,7 +243,7 @@ var LibraryRowView = Backbone.View.extend({
                         <td style="color:grey;"><%- library.get("name") %></td>
                     <% } else { %>
                         <td>
-                            <a href="folders/<%- library.get("root_folder_id") %>"><%- library.get("name") %></a>
+                            <a href="/library/folders/<%- library.get("root_folder_id") %>"><%- library.get("name") %></a>
```
dannon: This should actually probably be a `getAppRoot()`-based URL?  I have not tried it, but does this break when serving at a proxy?  e.g. `http://myinfrastructureurl.org/galaxy/library/folders/`?

OlegZharkov: > This should actually probably be a `getAppRoot()`-based URL? I have not tried it, but does this break when serving at a proxy? e.g. `http://myinfrastructureurl.org/galaxy/library/folders/`?

yes, that's right. Should be better now

OlegZharkov:
thanks! @mvdbeek @dannon

Issue #10285, opened 2020-09-24
Histories shared with you by others: No matching entries found for tags:undefined


mvdbeek:
<img width="714" alt="Screenshot 2020-09-24 at 17 48 35" src="https://user-images.githubusercontent.com/6804901/94168785-4aaec400-fe8e-11ea-85b6-a3d35c749c47.png">

I guess this is all ok, but looks like something minor went wrong generating the message.
Probably related to https://github.com/galaxyproject/galaxy/pull/9706/files

mvdbeek:
Fixed in https://github.com/galaxyproject/galaxy/pull/10286

PR #10286, opened 2020-09-24
[20.09] Fix histories shared with me message

mvdbeek:
Fixes https://github.com/galaxyproject/galaxy/issues/10285


nsoranzo (approved):

dannon:
Good catch; yeah, this list doesn't have the advanced search input for tags at all, so it's just undefined.

dannon (approved):

PR #10288, opened 2020-09-24
Add Visualizations to Pages

guerler:
This PR introduces adding visualizations into pages. Currently no parameters other than the dataset can be specified within the page markdown but users can interactively modify the visualization within the page view. This PR does not fix issues with individual visualizations.

Screen Shot 2020-10-05 at 3 44 14 PM


lib/galaxy/managers/markdown_util.py
```diff
@@ -92,46 +92,35 @@ def _remap(container, line):
...
             elif container == "history_dataset_info":
-                assert object_id is not None
                 hda = hda_manager.get_accessible(object_id, trans.user)
                 rval = self.handle_dataset_info(line, hda)
             elif container == "history_dataset_type":
-                assert object_id is not None
                 hda = hda_manager.get_accessible(object_id, trans.user)
                 rval = self.handle_dataset_type(line, hda)
             elif container == "history_dataset_name":
-                assert object_id is not None
```
jmchilton, line 129: I liked these assertions - was this needed?

guerler: Thanks for looking into this. I intend to raise errors instead if that's ok. Have not pushed the commits to this PR yet tho.

guerler: Errors are now raised and parsed in the response dictionary instead of the asserts. Additional validation could be added in follow-ups to the `validate_galaxy_markdown` in `markdown_parse.py`.

OlegZharkov:
Thanks, cool stuff.
Some help about this feature or an example in `client/src/components/Markdown/markdownHelp.js` might be quite useful.

OlegZharkov:
So I am trying to check it out. What am I doing wrong? 
## Workflow
```galaxy
visualization(history_dataset_id="test", id="nvd3_bar_stacked")
```
where "test" is the label of workflow step

lib/galaxy/managers/markdown_util.py
jmchilton: Awesome - thanks!

lib/galaxy/managers/markdown_parse.py
```diff
@@ -32,6 +32,7 @@
     "generate_galaxy_version": [],
     "generate_time": [],
     "invocation_time": ["invocation_id"],
+    "visualization": ["id", "history_dataset_id"],
```
jmchilton, line 35: visualization_id I think.

OlegZharkov: I don't think it's the case. it fetches `${getAppRoot()}api/plugins/${this.args.id}` at [this line](https://github.com/galaxyproject/galaxy/pull/10288/files#diff-e3dc9501b8af517e5d7201a3da7e64daR43) and http://localhost:8080/api/plugins/nvd3_bar_stacked request on network tab works as expected

guerler:
> So I am trying to check it out. What am I doing wrong?
> 
> ```
> visualization(history_dataset_id="test", id="nvd3_bar_stacked")
> ```
> 
> where "test" is the label of workflow step

If "test" is the output label, you need to call it as: `visualization(id="nvd3_bar_stacked", output="test").
However in a future PR, will modify the variable name and use "visualization_id" instead of "id" for consistency.

OlegZharkov:
> output

that's a good catch, however I still couldn't make it work :cry:. I am getting
```
  File "lib/galaxy/managers/markdown_util.py", line 736, in _remap_galaxy_markdown_calls
    return _remap_galaxy_markdown_containers(_remap_container, markdown)
  File "lib/galaxy/managers/markdown_util.py", line 704, in _remap_galaxy_markdown_containers
    (replacement, whole_block) = func(replace)
  File "lib/galaxy/managers/markdown_util.py", line 732, in _remap_container
    return func(match.group(1), matching_line + "\n")
  File "lib/galaxy/managers/markdown_util.py", line 665, in _remap
    ref_object = invocation.get_output_object(name)
  File "lib/galaxy/model/__init__.py", line 5298, in get_output_object
    raise Exception("Failed to find workflow output named [{}], workflow doesn't define output by that name - valid names are {}.".format(label, self.workflow.workflow_output_labels))
```

My report:

```galaxy
visualization(output="test", id="nvd3_bar_stacked")
```
[My workflow ](https://gist.github.com/OlegZharkov/6b414afd3f11ca14e6c91e458a3debb1)

guerler:
> > output
> 
> that's a good catch, however I still couldn't make it work 😢. I am getting
> 
> ```
>   File "lib/galaxy/managers/markdown_util.py", line 736, in _remap_galaxy_markdown_calls
>     return _remap_galaxy_markdown_containers(_remap_container, markdown)
>   File "lib/galaxy/managers/markdown_util.py", line 704, in _remap_galaxy_markdown_containers
>     (replacement, whole_block) = func(replace)
>   File "lib/galaxy/managers/markdown_util.py", line 732, in _remap_container
>     return func(match.group(1), matching_line + "\n")
>   File "lib/galaxy/managers/markdown_util.py", line 665, in _remap
>     ref_object = invocation.get_output_object(name)
>   File "lib/galaxy/model/__init__.py", line 5298, in get_output_object
>     raise Exception("Failed to find workflow output named [{}], workflow doesn't define output by that name - valid names are {}.".format(label, self.workflow.workflow_output_labels))
> ```
> 
> My report:
> 
> ```
> visualization(output="test", id="nvd3_bar_stacked")
> ```
> 
> [My workflow](https://gist.github.com/OlegZharkov/6b414afd3f11ca14e6c91e458a3debb1)

This error message is correct. It looks like you specified a `step` label instead of an `output` label. In order to specify an `output` label, you need to select a workflow step and scroll down to the steps output section at the bottom of the form. Click on `Configure output` and specify a label.

OlegZharkov (approved):
After short offline discussion and help from @guerler, everything worked as expected.
Thanks a lot @guerler, that's pretty cool feature! Looking forward for the 2nd part :)
image

bgruening:
Omg, I want 21.01 already!

Thanks @guerler and @OlegZharkov for the review!

PR #10294, opened 2020-09-25
[20.05] Fix webless handler --attach-to-pool

mvdbeek:
We need to create a connection before reading
dialect.server_version_info. See https://docs.sqlalchemy.org/en/13/core/internals.html#sqlalchemy.engine.interfaces.Dialect
Fixes https://github.com/galaxyproject/galaxy-helm/issues/160


dannon:
I know I ran into this exact same issue recently and we addressed it slightly differently, but now I can't find it.  Familiar to anyone else?

dannon:
I know I fixed exactly this same thing last week with someone, but I don't see another commit here, and (obviously) looks like I forgot to add it to 20.05 then.  Anyone recall a conversation on gitter about it?  I mainly want to follow up on the other end and see if I can drop a patch from somewhere, this looks fine.

PR #10301, opened 2020-09-25
[20.09] Revise quota meter style - backport.

dannon:
This should have targeted 20.09.


jmchilton:
Thanks for the backport @dannon and thanks for the fix @guerler

PR #10304, opened 2020-09-26
Reimplement collection order test, add wrong order, add `expect_test_failure` attribute

bernt-matthias:
- Reimplemented the collection order test, i.e. simpler (aka shorter) code
- Added a 2nd test for wrong order
- For this a new attribute `expect_test_failure` has been added to `<test>` that allows to express that at least one test assumption was not met. This might be handy in general to test failing assumptions in tool tests.

## How to test the changes?
- [x] I've included appropriate [automated tests](https://docs.galaxyproject.org/en/latest/dev/writing_tests.html).
- [ ] This is a refactoring of components with existing test coverage.
- [ ] Instructions for manual testing are as follows:

## License
- [x] I agree to license these contributions under [Galaxy's current license](https://github.com/galaxyproject/galaxy/blob/dev/LICENSE.txt).
- [x] I agree to allow the Galaxy committers to license these and all my past contributions to the core galaxy codebase under the [MIT license](https://opensource.org/licenses/MIT). If this condition is an issue, uncheck and just let us know why with an e-mail to galaxy-committers@lists.galaxyproject.org.

all collection have a wrong order and the test should fail

this seems not to be the case (see discussion in https://github.com/galaxyproject/tools-iuc/issues/3203)

### DIFF
--- a/lib/galaxy/tool_util/verify/interactor.py
+++ b/lib/galaxy/tool_util/verify/interactor.py
@@ -1188,9 +1188,9 @@ def _verify_outputs(testdef, history, jobs, data_list, data_collection_list, gal
                     print(_format_stream(job_stdio[stream], stream=stream, format=True), file=sys.stderr)
         found_exceptions.append(e)
 
-    # if testdef.expect_failure:
-    #     if testdef.outputs:
-    #         raise Exception("Cannot specify outputs in a test expecting failure.")
+    if testdef.expect_failure:
+        if testdef.outputs:
+            raise Exception("Cannot specify outputs in a test expecting failure.")
 
     # Wait for the job to complete and register expections if the final
     # status was not what test was expecting.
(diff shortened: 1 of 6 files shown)


bernt-matthias:
Hi @jmchilton .. Seems that your fix works. Question is how we can test this?

bernt-matthias:
OK. This works now. For some reason the old code also did not work for the working test. I guess it is because we relied on the order of a dict. 

But I needed to change the  semantics of `expect_failure`:

- this currently covers only tool failures (e.g. validators, exit code)
- but we want to check if the test fails. maybe we need another attribute for this or do we want to keep this?

lib/galaxy/tool_util/verify/interactor.py
```diff
@@ -1297,7 +1293,7 @@ def register_exception(e):
         except Exception as e:
             register_exception(e)
 
-    if found_exceptions:
+    if found_exceptions and not testdef.expect_test_failure:
```
jmchilton, line 1296: We never take the next step of asserting the test did in fact fail though right? This is only useful for internal stuff I think - so I'm happy to merge as is.

jmchilton (approved):

PR #10307, opened 2020-09-27
[20.01] Ensure that k8s job ids are unique

nuwang:
Currently, the k8s runner generates an external id for each job by appending the job id to a string of format:
galaxy-<instance_id>-<job_id>

However, when metadata jobs or datatype conversion jobs run, they are spawned with the same job id, which can cause conflicts when jobs are executing in quick succession. (e.g. the first job has not yet been cleaned up before the metadata job is spawned). This PR appends an additional unique id to each job so that the name is always unique.

### DIFF
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -226,7 +226,7 @@ class KubernetesJobRunner(AsynchronousJobRunner):
         (see pod selector) and an appropriate restart policy."""
         k8s_spec_template = {
             "metadata": {
-                "labels": {"app": self.__produce_unique_k8s_job_name(ajs.job_wrapper.get_id_tag())}
+                "labels": {"app": self.__produce_unique_k8s_job_name(ajs.job_wrapper.get_id_tag())[:-5]}
             },
             "spec": {
                 "volumes": self.runner_params['k8s_mountable_volumes'],
@@ -526,7 +526,7 @@ class KubernetesJobRunner(AsynchronousJobRunner):
         """Attempts to delete a dispatched job to the k8s cluster"""
         job = job_wrapper.get_job()
         try:
-            name = self.__produce_unique_k8s_job_name(job.get_id_tag())
+            name = job.job_runner_external_id
             namespace = self.runner_params['k8s_namespace']
             job_to_delete = find_job_object_by_name(self._pykube_api, name, namespace)
             if job_to_delete:
--- a/lib/galaxy/jobs/runners/util/pykube_util.py
+++ b/lib/galaxy/jobs/runners/util/pykube_util.py
@@ -1,9 +1,7 @@
 """Interface layer for pykube library shared between Galaxy and Pulsar."""
 import logging
 import os
-import random
 import re
-import string
 import uuid
 
 try:
@@ -59,9 +57,7 @@ def produce_unique_k8s_job_name(app_prefix=None, instance_id=None, job_id=None):
     if instance_id and len(instance_id) > 0:
         job_name += "%s-" % instance_id
 
-    unique_id = ''.join(random.choices(string.ascii_lowercase, k=4))
-
-    return job_name + job_id + "-" + unique_id
+    return "{}{}-{}".format(job_name, job_id, uuid.uuid4())
 
 
 def pull_policy(params):
--- a/lib/galaxy/jobs/runners/util/pykube_util.py
+++ b/lib/galaxy/jobs/runners/util/pykube_util.py
@@ -1,7 +1,9 @@
 """Interface layer for pykube library shared between Galaxy and Pulsar."""
 import logging
 import os
+import random
 import re
+import string
 import uuid
 
 try:
@@ -57,7 +59,9 @@ def produce_unique_k8s_job_name(app_prefix=None, instance_id=None, job_id=None):
     if instance_id and len(instance_id) > 0:
         job_name += "%s-" % instance_id
 
-    return job_name + job_id
+    unique_id = ''.join(random.choices(string.ascii_lowercase, k=4))
+
+    return job_name + job_id + "-" + unique_id
 
 
 def pull_policy(params):


lib/galaxy/jobs/runners/util/pykube_util.py
```diff
@@ -57,7 +59,9 @@ def produce_unique_k8s_job_name(app_prefix=None, instance_id=None, job_id=None):
     if instance_id and len(instance_id) > 0:
         job_name += "%s-" % instance_id
 
-    return job_name + job_id
+    unique_id = ''.join(random.choices(string.ascii_lowercase, k=4))
+
+    return job_name + job_id + "-" + unique_id
```
mvdbeek:
```suggestion
    return "{}{}-{}".format(job_name, job_id, uuid.uuid4())
```

lib/galaxy/jobs/runners/util/pykube_util.py
```diff
@@ -1,7 +1,9 @@
 """Interface layer for pykube library shared between Galaxy and Pulsar."""
 import logging
 import os
+import random
```
mvdbeek:
```suggestion
```

lib/galaxy/jobs/runners/util/pykube_util.py
```diff
@@ -1,7 +1,9 @@
...
 import re
+import string
```
mvdbeek:
```suggestion
```

mvdbeek:
That is interesting, I wonder if that might somehow play into https://github.com/galaxyproject/galaxy/issues/10245

lib/galaxy/jobs/runners/util/pykube_util.py
nuwang: I thought of avoiding uuid for two reasons - one is that if the name is too long, it runs into the k8s name length limit. If you just take a substring of the uuid though, we are running a more expensive operation for no real benefit, since the job_id is already unique, so we just need a few extra unique characters in the less frequent event of another job with the same id being spawned. I'm not immune to the charms of fewer imports though, so if you think it's better, could perhaps switch to `uuid.uuid4()[:4]` instead?

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -227,7 +227,7 @@ def __get_k8s_job_spec_template(self, ajs):
         (see pod selector) and an appropriate restart policy."""
         k8s_spec_template = {
             "metadata": {
-                "labels": {"app": self.__produce_unique_k8s_job_name(ajs.job_wrapper.get_id_tag())}
+                "labels": {"app": self.__produce_unique_k8s_job_name(ajs.job_wrapper.get_id_tag())[:-5]}
```
nuwang, line 229: Haven't elevated this from draft pending a cleaner solution here - stripping the last 5 characters for the app label works but is hard to understand.

lib/galaxy/jobs/runners/util/pykube_util.py
mvdbeek: 4 seems too short if you're dealing with potentially millions of jobs. But no other external ids include the job id, so how about completely dropping the job_id handling on line 54 and have the external id look like `"{}{}-{}".format(job_name, uuid.uuid4())`

nuwang:
> That is interesting, I wonder if that might somehow play into #10245

Seems possible. It's quite surprising that this flew under the radar for as long as it did. I guess it came up before, but I suppose we kept attributing other reasons to it.

lib/galaxy/jobs/runners/util/pykube_util.py
mvdbeek: (which you can truncate to the k8s limit)

nuwang: I thought the original reason for doing this was perhaps to make it easier to debug/match to the job id, but then again, it looks like we can match it up with the "Runner Job ID" if needed so yes, makes sense. Will update.

innovate-invent:
I believe k8s generates a unique id across the cluster for jobs. It can be accessed via job.metadata['labels']['controller-uid'].

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids

https://serverfault.com/a/977476

nuwang:
It seems better to continue the review with the new PR since that additionally eliminates a chunk of job duplication detection code. I'll close this in favour of continuing the review there.

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -123,44 +123,21 @@ def queue_job(self, job_wrapper):
...
-            job.create()
+        job.create()
+        job_id = job.metadata['name']
 
         # define job attributes in the AsyncronousJobState for follow-up
-        ajs.job_id = k8s_job_name
+        ajs.job_id = job_id
         # store runner information for tracking if Galaxy restarts
-        job_wrapper.set_job_destination(job_wrapper.job_destination, k8s_job_name)
+        job_wrapper.set_job_destination(job_wrapper.job_destination, job_id)
```
nuwang: This should be
```
job_wrapper.set_external_id(job_id)
```
in galaxy 20.05 and higher

nuwang:
@mvdbeek This PR has been updated to combine the work in PR https://github.com/galaxyproject/galaxy/pull/10315 with the following specific modifications: https://github.com/galaxyproject/galaxy/pull/10315#discussion_r500811284

Would be great if you could review.

innovate-invent:
The generateName field does not generate unique names across time. You would need to retain the block of code that checks if it trusts the job.

nuwang:
The untrusted code existed to handle cases where the identical job name would be generated by a newly started Galaxy instance, and the new instance would mistakenly take over the old job. This won't be a problem with generateName because it will always generate a new unique name.

innovate-invent:
I am still not sure I understand why you don't select by uid. It eliminates any possible collision in time.
I have added more commits to my PR you might want to pull regardless.

nuwang:
The rationale in summary is:
a. The "primary key" of a k8s object is its name whereas the uid is not a valid field-selector at all, and selecting by uid is reliant on implicit behaviour
b. You could break the controller-uid label propagation if you add a custom label as you've mentioned here: https://github.com/galaxyproject/galaxy/pull/10315/files#diff-9deffce866ba96965f0de486716f0c4dR105 whereas in contrast, the job-name label is always propagated.
c. It also doesn't seem likely that uid will become a field selector: https://github.com/kubernetes/kubernetes/pull/76870
d. a collision in time in the sense that the original code was accounting for is no longer possible
e. a collision in time could hypothetically happen in rare circumstances in a dev environment but is easily circumvented in a production environment just by giving the galaxy instance a unique name.

So while I do agree that the uid would eliminate any possible collision in time, the fact that the design intent of the uid isn't meant to facilitate selection, outweighs the primary advantage that it provides. Sound ok?

And thanks for those new additions - they look good. Since this PR is targeting 20.01, I'd propose we apply those as an enhancement to dev.

nuwang:
@mvdbeek Have reverted to the original fix as you suggested as there are too many changes for a bug fix, and we can make the new refactored changes targeting dev.

innovate-invent:
This PR is failing to schedule jobs for my deployment:

```
  File "/srv/galaxy/venv/lib/python3.8/site-packages/pykube/http.py", line 99, in raise_for_status
    resp.raise_for_status()
  File "/srv/galaxy/venv/lib/python3.8/site-packages/requests/models.py", line 941, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 422 Client Error: Unprocessable Entity for url: https://172.20.0.1:443/apis/batch/v1/namespaces/islandcompare/jobs
  File "/srv/galaxy/lib/galaxy/jobs/runners/__init__.py", line 137, in run_next
    method(arg)
  File "/srv/galaxy/lib/galaxy/jobs/runners/kubernetes.py", line 155, in queue_job
    job.create()
  File "/srv/galaxy/venv/lib/python3.8/site-packages/pykube/objects.py", line 97, in create
    self.api.raise_for_status(r)
  File "/srv/galaxy/venv/lib/python3.8/site-packages/pykube/http.py", line 106, in raise_for_status
    raise HTTPError(resp.status_code, payload["message"])
pykube.exceptions.HTTPError: Job.batch "galaxy-islandcompare-229218-7c092c24-96a0-4db8-9c67-b3060fd0515d" is invalid: [metadata.labels: Invalid value: "galaxy-islandcompare-229218-7c092c24-96a0-4db8-9c67-b3060fd0515d": must be no more than 63 characters, spec.template.labels: Invalid value: "galaxy-islandcompare-229218-7c092c24-96a0-4db8-9c67-b3060fd0515d": must be no more than 63 characters]
galaxy.jobs.runners.kubernetes DEBUG 2020-10-13 18:04:23,163 (229218/None) Terminated at user's request
```

nuwang:
@innovate-invent Thanks for reporting. This is a quick patch for that issue: https://github.com/galaxyproject/galaxy/pull/10416

PR #9772, opened 2020-05-15
Added test to check that jobs fail when deleted directly via k8s api

nuwang:

PR #10308, opened 2020-09-28
[20.01] Prevent endless cleanup loops when duplicate job or k8s job not found

nuwang:
This fixes an issue where the job is placed back in the watched job queue, although no further progress is possible (e.g. job externally deleted in k8s), resulting in an endless loop.

### DIFF
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -2,7 +2,6 @@
 Offload jobs to a Kubernetes cluster.
 """
 
-import errno
 import logging
 import math
 import os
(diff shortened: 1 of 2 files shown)


lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -468,27 +468,21 @@ def check_watched_item(self, job_state):
...
                 with open(job_state.error_file, 'w') as error_file:
                     error_file.write("No Kubernetes Jobs are available under expected selector app=%s\n" % job_state.job_id)
-            except EnvironmentError as e:
-                # Python 2/3 compatible handling of FileNotFoundError
-                if e.errno == errno.ENOENT:
-                    log.error("Job directory already cleaned up. Assuming already handled for selector app=%s", job_state.job_id)
-                else:
-                    raise
-            return job_state
+            except Exception:
```
mvdbeek: Not sure a blank exception is the right thing here. Looks like we already handle `e.errno == errno.ENOENT:` explicitly.

mvdbeek:
Can you add a testcase for this ? You can base this off of https://github.com/mvdbeek/galaxy/blob/d7ab0120609afddb7caa4c4f83631ba6aa56b215/test/integration/test_kubernetes_runner.py#L220

lib/galaxy/jobs/runners/kubernetes.py
nuwang: The reason I removed the specific exception is because, no matter what the exception is, the job needs to be removed from the watch queue. Reraising it will result in it being placed back in the queue, and we are back to an endless loop? (we've already marked the job as having failed by this point)

mvdbeek: If you have NFS issues and the file isn't there you'd probably need to retry collection in the next few loops, which I suppose is what the FileNotFoundError handling does.

nuwang:
Just realized that the test case was already written. Probably explains why it was failing  :-) https://github.com/galaxyproject/galaxy/pull/9772

lib/galaxy/jobs/runners/kubernetes.py
mvdbeek: That said writing to `job_state.error_file` seems wrong, that is not the purpose of this file, and yeah, the k8s API already doesn't know about the job anymore. What happens if you remove this entirely ?

nuwang: Left it as is because I wasn't sure whether it was meant to be there, but now that you mention it, don't think I've ever seen it work. Will remove.

nuwang:
@mvdbeek The test cases have been included. Can you take a look?

mvdbeek:
Thanks @nuwang!

PR #10313, opened 2020-09-28
possibly reverting #9722 

mvdbeek:
We're seeing high load on the database server, checking now if the tests still pass


mvdbeek:
That's not necessary, https://github.com/galaxyproject/galaxy/pull/10360 fixed the performance issue.

PR #10315, opened 2020-09-28
[20.05] Change k8s runner to use k8s assigned uid to track jobs rather than name

### LINKED ISSUE #10245: Database deadlock
The following exception occurs while running a very large workflow:
```
galaxy.metadata DEBUG 2020-09-18 19:35:46,971 setting metadata externally failed for HistoryDatasetAssociation 54263: Metadata results could not be read from '/data/jobs_directory/024/24438/metadata/metadata_results_outfile'
galaxy.jobs.runners ERROR 2020-09-18 19:35:48,127 (24438/galaxy-islandcompare-24438) Job wrapper finish method failed
  File "/srv/galaxy/venv/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1247, in _execute_context
    self.dialect.do_execute(
  File "/srv/galaxy/venv/lib/python3.8/site-packages/sqlalchemy/engine/default.py", line 590, in do_execute
    cursor.execute(statement, parameters)
psycopg2.errors.DeadlockDetected: deadlock detected
DETAIL:  Process 25865 waits for ShareLock on transaction 646521; blocked by process 4054.
Process 4054 waits for ShareLock on transaction 646527; blocked by process 25743.
Process 25743 waits for ExclusiveLock on tuple (11,36) of relation 17171 of database 16396; blocked by process 25865.
HINT:  See server log for query details.
CONTEXT:  while updating tuple (11,36) in relation "history"
SQL statement "UPDATE history
                    SET update_time = (now() at time zone 'utc')
                    WHERE id = NEW.history_id OR id = OLD.history_id"
PL/pgSQL function update_history_update_time() line 9 at SQL statement


  File "/srv/galaxy/lib/galaxy/jobs/runners/__init__.py", line 540, in _finish_or_resubmit_job
    job_wrapper.finish(tool_stdout, tool_stderr, exit_code, check_output_detected_state=check_output_detected_state, job_stdout=job_stdout, job_stderr=job_stderr)
  File "/srv/galaxy/lib/galaxy/jobs/__init__.py", line 1713, in finish
    self.sa_session.flush()
  File "/srv/galaxy/venv/lib/python3.8/site-packages/sqlalchemy/orm/scoping.py", line 162, in do
    return getattr(self.registry(), name)(*args, **kwargs)
  File "/srv/galaxy/venv/lib/python3.8/site-packages/sqlalchemy/orm/session.py", line 2496, in flush
    self._flush(objects)
  File "/srv/galaxy/venv/lib/python3.8/site-packages/sqlalchemy/orm/session.py", line 2637, in _flush
    transaction.rollback(_capture_exception=True)
  File "/srv/galaxy/venv/lib/python3.8/site-packages/sqlalchemy/util/langhelpers.py", line 68, in __exit__
    compat.raise_(
  File "/srv/galaxy/venv/lib/python3.8/site-packages/sqlalchemy/util/compat.py", line 178, in raise_
    raise exception
  File "/srv/galaxy/venv/lib/python3.8/site-packages/sqlalchemy/orm/session.py", line 2597, in _flush
    flush_context.execute()
  File "/srv/galaxy/venv/lib/python3.8/site-packages/sqlalchemy/orm/unitofwork.py", line 422, in execute
    rec.execute(self)
  File "/srv/galaxy/venv/lib/python3.8/site-packages/sqlalchemy/orm/unitofwork.py", line 586, in execute
    persistence.save_obj(
  File "/srv/galaxy/venv/lib/python3.8/site-packages/sqlalchemy/orm/persistence.py", line 230, in save_obj
    _emit_update_statements(
  File "/srv/galaxy/venv/lib/python3.8/site-packages/sqlalchemy/orm/persistence.py", line 994, in _emit_update_statements
    c = cached_connections[connection].execute(
  File "/srv/galaxy/venv/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 984, in execute
    return meth(self, multiparams, params)
  File "/srv/galaxy/venv/lib/python3.8/site-packages/sqlalchemy/sql/elements.py", line 293, in _execute_on_connection
    return connection._execute_clauseelement(self, multiparams, params)
  File "/srv/galaxy/venv/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1097, in _execute_clauseelement
    ret = self._execute_context(
  File "/srv/galaxy/venv/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1287, in _execute_context
    self._handle_dbapi_exception(
  File "/srv/galaxy/venv/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1481, in _handle_dbapi_exception
    util.raise_(
  File "/srv/galaxy/venv/lib/python3.8/site-packages/sqlalchemy/util/compat.py", line 178, in raise_
    raise exception
  File "/srv/galaxy/venv/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1247, in _execute_context
    self.dialect.do_execute(
  File "/srv/galaxy/venv/lib/python3.8/site-packages/sqlalchemy/engine/default.py", line 590, in do_execute
    cursor.execute(statement, parameters)
sqlalchemy.exc.OperationalError: (psycopg2.errors.DeadlockDetected) deadlock detected
DETAIL:  Process 25865 waits for ShareLock on transaction 646521; blocked by process 4054.
Process 4054 waits for ShareLock on transaction 646527; blocked by process 25743.
Process 25743 waits for ExclusiveLock on tuple (11,36) of relation 17171 of database 16396; blocked by process 25865.
HINT:  See server log for query details.
CONTEXT:  while updating tuple (11,36) in relation "history"
SQL statement "UPDATE history
                    SET update_time = (now() at time zone 'utc')
                    WHERE id = NEW.history_id OR id = OLD.history_id"
PL/pgSQL function update_history_update_time() line 9 at SQL statement

[SQL: UPDATE history_dataset_association SET update_time=%(update_time)s, info=%(info)s, blurb=%(blurb)s, peek=%(_peek)s, tool_version=%(tool_version)s, metadata=%(_metadata)s, version=%(version)s WHERE history_dataset_association.id = %(history_dataset_association_id)s]
[parameters: {'update_time': datetime.datetime(2020, 9, 18, 19, 35, 47, 120934), 'info': '', 'blurb': '2,897 lines, 98 comments', '_peek': '#\n#\n# Core-specific parameters:\n#\n#    Menu parameters:\n', 'tool_version': 'GNU Awk 5.0.1, API: 2.0\n', '_metadata': <psycopg2.extensions.Binary object at 0x7f51075e1ed0>, 'version': 2, 'history_dataset_association_id': 54263}]
(Background on this error at: http://sqlalche.me/e/e3q8)
galaxy.jobs ERROR 2020-09-18 19:35:48,180 fail(): Missing output file in working directory: [Errno 2] No such file or directory: '/data/jobs_directory/024/24438/outputs/galaxy_dataset_02b9521f-28a9-4cb2-9d9c-40f3a8088fd3.dat'
```

The dataset reports `Unable to finish job` as the error.

Galaxy 20.05 (autoscaled uwsgi app, 3 job handlers, 3 workflow schedulers)
```
uwsgi:
      buffer-size: 16384
      processes: 1
      threads: 4
      offload-threads: 2
```

I am not sure how to prevent this.

innovate-invent:
This is a competing PR with https://github.com/galaxyproject/galaxy/pull/10307 to solve the issue of colliding job names.

This uses the uid generated by k8s to track the job rather than any generated name by Galaxy. This guarantees no collision.

Hopefully this will resolve #10245.

### DIFF
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -221,9 +221,9 @@ class KubernetesJobRunner(AsynchronousJobRunner):
                     "app.kubernetes.io/component": "tool",
                     "app.kubernetes.io/part-of": "galaxy",
                     "app.kubernetes.io/managed-by": "galaxy",
-                    "app.galaxy.org/job_id": ajs.job_wrapper.get_id_tag(),
-                    "app.galaxy.org/instance": self._galaxy_instance_id or "",
-                    "app.galaxy.org/handler": self.app.config.server_name,
+                    "app.galaxyproject.org/job_id": ajs.job_wrapper.get_id_tag(),
+                    "app.galaxyproject.org/instance": self._galaxy_instance_id or "",
+                    "app.galaxyproject.org/handler": self.app.config.server_name,
                 }
             },
             "spec": {
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -214,7 +214,14 @@ class KubernetesJobRunner(AsynchronousJobRunner):
         (see pod selector) and an appropriate restart policy."""
         k8s_spec_template = {
             "metadata": {
-                "labels": {"app": self.__produce_unique_k8s_job_name(ajs.job_wrapper.get_id_tag())}
+                "labels": {
+                    "app.kubernetes.io/name": ajs.tool_wrapper.tool.id,
+                    "app.kubernetes.io/instance": "job-"+ajs.job_wrapper.get_id_tag(),
+                    "app.kubernetes.io/version": ajs.tool_wrapper.tool.version,
+                    "app.kubernetes.io/component": "tool",
+                    "app.kubernetes.io/part-of": self._galaxy_instance_id or "galaxy",
+                    "app.kubernetes.io/managed-by": "galaxy",
+                }
             },
             "spec": {
                 "volumes": self.runner_params['k8s_mountable_volumes'],
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -216,11 +216,14 @@ class KubernetesJobRunner(AsynchronousJobRunner):
             "metadata": {
                 "labels": {
                     "app.kubernetes.io/name": ajs.tool_wrapper.tool.id,
-                    "app.kubernetes.io/instance": "job-"+ajs.job_wrapper.get_id_tag(),
+                    "app.kubernetes.io/instance": self.__produce_unique_k8s_job_name(ajs.job_wrapper.get_id_tag()),
                     "app.kubernetes.io/version": ajs.tool_wrapper.tool.version,
                     "app.kubernetes.io/component": "tool",
-                    "app.kubernetes.io/part-of": self._galaxy_instance_id or "galaxy",
+                    "app.kubernetes.io/part-of": "galaxy",
                     "app.kubernetes.io/managed-by": "galaxy",
+                    "app.galaxy.org/job_id": ajs.job_wrapper.get_id_tag(),
+                    "app.galaxy.org/instance": self._galaxy_instance_id or "",
+                    "app.galaxy.org/handler": self.app.config.server_name,
                 }
             },
             "spec": {
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -224,6 +224,7 @@ class KubernetesJobRunner(AsynchronousJobRunner):
                     "app.galaxyproject.org/job_id": ajs.job_wrapper.get_id_tag(),
                     "app.galaxyproject.org/instance": self._galaxy_instance_id or "",
                     "app.galaxyproject.org/handler": self.app.config.server_name,
+                    "app.galaxyproject.org/destination": ajs.job_wrapper.job_destination.id,
                 }
             },
             "spec": {
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -7,7 +7,6 @@ import logging
 import math
 import os
 import re
-from time import sleep
 
 import yaml
 
@@ -20,9 +19,11 @@ from galaxy.jobs.runners import (
 from galaxy.jobs.runners.util.pykube_util import (
     DEFAULT_JOB_API_VERSION,
     ensure_pykube,
-    find_job_object_by_name,
+    find_job_object_by_id,
+    find_pod_object_by_id,
     galaxy_instance_id,
     Job,
+    JOB_ID_LABEL,
     job_object_dict,
     Pod,
     produce_unique_k8s_job_name,
@@ -131,37 +132,23 @@ class KubernetesJobRunner(AsynchronousJobRunner):
             self.__get_k8s_job_spec(ajs)
         )
 
-        # Checks if job exists and is trusted, or if it needs re-creation.
         job = Job(self._pykube_api, k8s_job_obj)
-        job_exists = job.exists()
-        if job_exists and not self._galaxy_instance_id:
-            # if galaxy instance id is not set, then we don't trust matching jobs and we simply delete and
-            # re-create the job
-            log.debug("Matching job exists, but Job is not trusted, so it will be deleted and a new one created.")
-            job.delete()
-            elapsed_seconds = 0
-            while job.exists():
-                sleep(3)
-                elapsed_seconds += 3
-                if elapsed_seconds > self.runner_params['k8s_timeout_seconds_job_deletion']:
-                    log.debug("Timed out before k8s could delete existing untrusted job " + k8s_job_name +
-                              ", not queuing associated Galaxy job.")
-                    return
-                log.debug("Waiting for job to be deleted " + k8s_job_name)
-
-            Job(self._pykube_api, k8s_job_obj).create()
-        elif job_exists and self._galaxy_instance_id:
-            # The job exists and we trust the identifier.
-            log.debug("Matching job exists, but Job is trusted, so we simply use the existing one for " + k8s_job_name)
-            # We simply leave the k8s job to be handled later on by check_watched_item().
-        else:
-            # Creates the Kubernetes Job if it doesn't exist.
-            job.create()
+        job.create()
+        job_id = job.labels.get(JOB_ID_LABEL, False)
+        if not job_id:
+            ## Recover uid label because it wasn't set
+            #job.labels[JOB_ID_LABEL] = job.metadata.uid
+            #job.update()
+            #job_id = job.labels.get(JOB_ID_LABEL, False)
+            #if not job_id:
+            job_wrapper.fail("Unexpected value from job runner", exception=True)
+            log.exception("%s not assigned by k8s to job on invocation: %s" % (JOB_ID_LABEL, job.obj))
+            return
 
         # define job attributes in the AsyncronousJobState for follow-up
-        ajs.job_id = k8s_job_name
+        ajs.job_id = job_id
         # store runner information for tracking if Galaxy restarts
-        job_wrapper.set_external_id(k8s_job_name)
+        job_wrapper.set_external_id(job_id)
         self.monitor_queue.put(ajs)
 
     def __get_pull_policy(self):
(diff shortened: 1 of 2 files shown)


nuwang (commented):
Nice job eliminating a chunk of code. This looks good to me. I think it would be good to squash and fix-up the commit history to a single commit so it's easier to cherry-pick etc.

innovate-invent:
I have tested this with a large workflow and it appears to function correctly. Unfortunately it does not resolve #10245.

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -131,37 +132,23 @@ def queue_job(self, job_wrapper):
             self.__get_k8s_job_spec(ajs)
         )
 
-        # Checks if job exists and is trusted, or if it needs re-creation.
         job = Job(self._pykube_api, k8s_job_obj)
-        job_exists = job.exists()
-        if job_exists and not self._galaxy_instance_id:
```
pcm32, line 137: This piece of code is related to Galaxy instances being spin up again on k8s clusters where they were running in the past or where another Galaxy instance might have run. Could you explain why you are removing this?

pcm32: I see, if you use the internal cluster UUID then this is never really a problem.

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -131,37 +132,23 @@ def queue_job(self, job_wrapper):
...
-            # The job exists and we trust the identifier.
-            log.debug("Matching job exists, but Job is trusted, so we simply use the existing one for " + k8s_job_name)
-            # We simply leave the k8s job to be handled later on by check_watched_item().
-        else:
-            # Creates the Kubernetes Job if it doesn't exist.
-            job.create()
+        job.create()
+        job_id = job.labels.get(JOB_ID_LABEL, False)
+        if not job_id:
+            ## Recover uid label because it wasn't set
```
nuwang: The commented code can be removed? (should also fix linting failure).

nuwang (approved):
LGTM!

lib/galaxy/jobs/runners/kubernetes.py
innovate-invent: I am not sure if we should retain this in the event that k8s stops adding the controller-uid tag. I think the tagging is a side effect of a feature where k8s autopopulates the job labels with the pod labels if they are not specified. The pod is actually the object being tagged by k8s with the controller-uid label.

https://github.com/kubernetes/kubernetes/pull/32110

pcm32: So the k8s job itself is not tracked? That would disallow the possibility of automatic retrials within k8s (ie. a node came down in the middle of the execution or you lost connection to it, which could happen on spot deployments).

pcm32: ahhh, no, you do get job_id there... I'm not sure I understand then.

innovate-invent: When you shedule a Job in k8s it adds a controller-uid label with the Jobs uid to the Pod spec. It then  generates the selector for the Job pods based on this label. Because the Job labels are ommitted k8s then copies all Pod labels including the controller-uid label it just added. I am exploiting this to allow selecting the Job on its uid. If k8s decided they diddnt like this side effect of copying the generated label and removed it then we would have to uncomment the code that recovers it.

nuwang: Hmm... I didn't realize till now that there's no way to actually filter by uid. Without that ability, I'm not sure that using the k8s generated uid buys us anything? We may as well generate a uid ourselves and attach it as a label, which would be filterable. I agree that it's dangerous to rely on default k8s behaviour here - it doesn't seem inconceivable that a sidecar container or similar could affect label propagation, or just changes to k8s behaviour in general, and that could really make a mess of things. I think the original logic of simply using a unique generated label makes more sense in this case.

As for re-attaching to the job @pcm32 - just to make sure I understand - this is to handle a very specific case where the k8s API job is created, but the galaxy handler is shutdown before `job. set_external_id(job_id)` is called? Because my assumption is that once the external job id is saved, then all's well since Galaxy can now track the job. So this would happen in a very narrow time-window, but could nevertheless happen, correct?

innovate-invent: I dont think there will be an issue with using the uid. Sidecars would not affect anything as the uid is still the Jobs uid and not the pods. My only concern with this is that it is not clearly stated functionality in the k8s documentation. Using the k8s uis is important because it is documented to be unique across the cluster for all time. Generating our own ids would require us to guarantee that they are unique across all handlers across the cluster. uids do have collision issues if not generating them in a coordinated fashion.

mvdbeek: What issues are these ?  That's the first I'm hearing of this, and if true would be a gigantic issue for databases using uuids as primary keys.

innovate-invent: Uuids are not guaranteed to be unique, they do collide and you can observe this in high traffic multi datacentre applications. Generally if you need to guarantee uniqueness in the strictest way, you need a central uuid issuer that validates collisions.

The point is that k8s gaurantees uniqueness of its uids.

nuwang: Do we need a k8s generated uid to guarantee a similar level of uniqueness though? The existing job name uses the job_id, and the issue still managed to fly under the radar for quite a while, because it only manifests itself if you run a job that reuses the same job id at the same time, something that happens only for metadata jobs. Appending a uuid to the job_id should make the chances of a collision zero for all practical purposes.

It also seems unlikely that support will be added for fetching objects by uid: https://github.com/kubernetes/kubernetes/pull/76870 The etcd key for the kubernetes object is apparently based on the name, not uid. Therefore, using `metadata.name` seems to be the right way to query the object?

innovate-invent: Because of a limitation in pykube you cant select on name or uid, only labels. K8s autoassigns the job uid to a label allowing selection on uid. The name field is meant to be human readable and isnt really intended to store values important to the machine. I dont know why you would use anything other than the k8s uid to track a k8s object. Generating a uid that is gauranteed to be unique is complicated and anything less is unnecessary and setting someone up for a headache.

dannon: > Do we need a k8s generated uid to guarantee a similar level of uniqueness though? The existing job name uses the job_id, and the issue still managed to fly under the radar for quite a while, because it only manifests itself if you run a job that reuses the same job id at the same time, something that happens only for metadata jobs. Appending a uuid to the job_id should make the chances of a collision zero for all practical purposes.

Completely agree.  If k8s uuids are colliding then job execution likely already has problems in the stack outside of our control (other than, maybe, saying "hey, I see two of these -- kill them both and try again" or something like that.

innovate-invent: I am not sure I understand, k8s uids are the only id in question gauranteed to not collide. Job ids will collide if you have multiple galaxy instances submitting to the same k8s cluster.

nuwang: It's not a limitation in pykube as far as I can see - it's a limitation in the k8s api (see this: https://github.com/kubernetes/kubernetes/pull/76870#issuecomment-485559299)
The k8s resource is located based on the object name, not the uid: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/
As the docs state, the uid is meant to differentiate between historical occurrences of two objects with the same name, not as a resource locator.

Also, jobs from multiple galaxy instances in the same namespace do not collide because of this: https://github.com/galaxyproject/galaxy/blob/c837c04c9a1ce0ef961016c8ccb7f13ba5864894/test/unit/jobs/job_conf.sample_advanced.yml#L140

innovate-invent: Pykube does not forward the interface to select on anything but labels. The interface to select on object fields is supported by k8s: https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/

nuwang: `metadata.uid` is not a valid field selector though: https://stackoverflow.com/questions/59442065/how-to-use-kubernetes-fieldselector-to-query-ownerreferences

innovate-invent: I was trying to point out that metadata.name is not a selector that pykube supports.

nuwang: @innovate-invent `metadata.name` does work as expected in pykube but has a different query syntax: https://github.com/galaxyproject/galaxy/pull/10307/files#diff-9deffce866ba96965f0de486716f0c4dR65
So I've gone ahead and cherry-picked your commits and included them in the original PR here: https://github.com/galaxyproject/galaxy/pull/10307 but with two modifications.
1. The name is used as the selector
2. The name is generated by k8s (using your idea of generateName) but with a custom prefix.
These two changes should give us the best of both worlds I think - lets k8s generate and manage a unique name while we can still use the name as the selector, while avoiding the complications of relying on default label propagation, or not being able to set custom labels on a job. Let me know what you think.

mvdbeek:
We're not going to do such extensive changes to existing releases, but we're trying to get https://github.com/galaxyproject/galaxy/pull/10407 into 20.09

Issue #10306, opened 2020-09-27
pytest 6.1.0 does not collect tests from symlinked directories


nsoranzo:
pytest 6.1.0 (released 2020-09-26) skips directory symlinks when collecting tests, which makes our `test_galaxy_packages` test build fail, see e.g. https://app.circleci.com/pipelines/github/galaxyproject/galaxy/10185/workflows/0ac542f8-fac3-4ed2-93ff-1ae21a54c203/jobs/107303

pytest 6.0.2 works fine.

PR #10320, opened 2020-09-28
Fix pytest breakage in package testing.

jmchilton:


nsoranzo:
Yep: https://github.com/galaxyproject/galaxy/issues/10306

PR #10321, opened 2020-09-29
[20.05] Respect ``use_shared_home="false"`` when running tools in container

nsoranzo:
When a tool doesn't specify a `profile` (or sets a `profile` < 18.01) but use ``use_shared_home="false"`` to ask to set a HOME within the job's directory (normally because the underlying tool writes to the home directory), this works as expected when resolving dependencies with conda.
But when running the same tool in a Docker container (e.g. via `planemo test --biocontainers`), the job script didn't export a temporary HOME into the container (which instead setting `profile="18.01"` correctly does).

Also add several integration tests for `use_shared_home`, of which `DockerizedJobsIntegrationTestCase.test_container_job_environment_explicit_isolated_home` would fail without this fix.

### DIFF
--- a/lib/galaxy/tool_util/parser/xml.py
+++ b/lib/galaxy/tool_util/parser/xml.py
@@ -177,13 +177,6 @@ class XmlToolSource(ToolSource):
         # break or modify any configurations by default.
         return "job_tmp_if_explicit"
 
-    def parse_docker_env_pass_through(self):
-        if self.parse_profile() < "18.01":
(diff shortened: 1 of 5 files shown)


mvdbeek (approved):

mvdbeek:
@jmchilton do you think this ok ?

jmchilton:
Oh - I reviewed this the other day and I didn't merge. Sorry. I think it looks fine.

PR #10323, opened 2020-09-29
Get environment variables for data_source_async tools

FredericBGA:
This PR try to make this one useless: https://github.com/galaxyproject/galaxy/pull/9207

It get environment variables from xml in order to send them as parameters in URL for data_source_async tools.

What I've understood: unlike the regular tools there is no command line that is explicitly created here (so I can not use the variables that are exported in ie: `database/jobs_directory/000/70/galaxy_70.sh`). Instead it creates an URL that is then used as a Request:
https://github.com/galaxyproject/galaxy/blob/4dd932290aacaab106a415b1c48a49be6662f79f/lib/galaxy/webapps/galaxy/controllers/tool_runner.py#L138
That's why I need to find way to get these values when the URL is build.

Let me know if it's ok or if it too specific to my needs. If not I can close these two PR.
Thank you.


lib/galaxy/tools/__init__.py
```diff
@@ -1721,6 +1724,16 @@ def get_static_param_values(self, trans):
                 args[key] = model.User.expand_user_properties(trans.user, param.value)
             else:
                 raise Exception("Unexpected parameter type")
+        if self.tool_type == 'data_source_async':
+            # parse xml in order to get 'environment_variable' values
+            for user_env_var in XmlToolSource(raw_tool_xml_tree(self.config_file)).parse_environment_variables():
```
jmchilton: Re-parsing the XML is not great, ``self.tool.environment_variables`` would be a lot better here.

jmchilton:
> Let me know if it's ok or if it too specific to my needs. If not I can close these two PR.

I think it might be okay - I like that you're restricting it to a specific class of tools. I just wish there was a way to test it.

lib/galaxy/tools/__init__.py
FredericBGA: ok, thank you for pointing this to me, I was not aware of this. I'm trying with a galaxy 20_05. Is this `self.tool.environment_variables` available? Because when I try to print its values, I've got an error `AttributeError: tool`
I will convert this to a draft, and then I will try to find time to work again on that, but based on 21_01!

jmchilton: I guess you're in the tool so it would just be ``self.environment_variables`` - sorry about that.

FredericBGA:
Hi,

tested with an async tool, it works.
I will close this one: https://github.com/galaxyproject/galaxy/pull/9207
Could you please help me to add a test for this?

PR #10326, opened 2020-09-29
Allow files config to use keyed dict with ids as dict key. 

jmchilton:
I've been told this is easier for our cloud infrastructure to reason about.


mvdbeek:
Thanks @jmchilton!

PR #10327, opened 2020-09-29
vuefy dataset information

OlegZharkov:
Another step towards removing `show_params.mako` and transfering that page to Vue.
image


client/src/components/DatasetInformation/mount.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Endpoint for mounting job metrics from non-Vue environment.
+ */
+import $ from "jquery";
```
dannon: Could you do this without jquery please?

OlegZharkov: It is temporary code, this `mount.js` file will be gone as soon as we convert `show_params.mako` to Vue. This is just a way to mount Vue inside `.mako` file.

OlegZharkov: @dannon but you're right, I can do this without jquery

OlegZharkov:
thank you @dannon!

PR #10329, opened 2020-09-29
[20.09] Fix double masthead in some spaces

### LINKED ISSUE #10291: The interactive environment menu bar needs to be fixed.
The appearance of The interactive environment menu bar is not proper.  It's showing up the double menu bar. 

Screen Shot 2020-09-24 at 4 01 16 PM

dannon:
Fixes #10291 

Do we know of any context where Galaxy is legitimately stuffed into a child frame?  If so, I have another commit that checks against the parent having a live 'Galaxy' object and only prevents duplicate masthead load when within another Galaxy, but would allow other circumstances.  But there's more that can go weird there, and I don't want anything else touching a global `Galaxy` if I don't really need to.

### DIFF
--- a/templates/webapps/galaxy/galaxy.masthead.mako
+++ b/templates/webapps/galaxy/galaxy.masthead.mako
@@ -35,11 +35,17 @@
 
     ## load the frame manager
     <script type="text/javascript">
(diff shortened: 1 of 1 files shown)


mvdbeek:
Thanks @dannon!

PR #10334, opened 2020-09-30
Allow 'planemo run' to stage in existing datasets without reuploading 

### LINKED ISSUE #1080: Allow existing dataset / dataset collection ids in jobs for planemo run

simonbray:
Fixes https://github.com/galaxyproject/planemo/issues/1080. This was much easier than I expected.

The idea is that the job or test yaml file can contain `galaxy_id`s for any `File` or `Collection` instead of a local path. Assuming these are valid for a particular galaxy server, the user can then use `planemo run` and execute the workflow using the existing datasets/collections and therefore can skip the upload step.

```
mgf_input_list:
  class: Collection
  galaxy_id: 1e94904b7ba7b9cc
metapeptides:
  class: File
  galaxy_id: 11ac94870d0bb33a71934a4bcff13734
gene_ontology_terms:
  class: File
  galaxy_id: 11ac94870d0bb33aa74376be3811a4fe
```

I tested this with both files and collections and it seems to work fine. Also, specifying `galaxy_id` for some files and `location` for others (i.e. so some files are uploaded and others not) works.

### DIFF
--- a/lib/galaxy/tool_util/cwl/util.py
+++ b/lib/galaxy/tool_util/cwl/util.py
@@ -84,7 +84,7 @@ def path_or_uri_to_uri(path_or_uri):
 
 
 def galactic_job_json(
-    job, test_data_directory, upload_func, copy_func, collection_create_func, tool_or_workflow="workflow"
+    job, test_data_directory, upload_func, collection_create_func, tool_or_workflow="workflow"
 ):
     """Adapt a CWL job object to the Galaxy API.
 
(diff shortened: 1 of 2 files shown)


lib/galaxy/tool_util/cwl/util.py
```diff
@@ -328,9 +332,10 @@ def _ensure_file_exists(file_path):
 
 class FileLiteralTarget:
 
-    def __init__(self, contents, **kwargs):
+    def __init__(self, contents, path=None, **kwargs):
         self.contents = contents
         self.properties = kwargs
+        self.path = path
```
simonbray, line 339: unrelated to this PR, but the `__str__` method of this class requires path to have a value

jmchilton:
I'm glad that was so simple! This seems awesome to me.

PR #10335, opened 2020-09-30
[20.09] Allow 'planemo run' to stage in existing datasets without reuploading

simonbray:
It would be great to get https://github.com/galaxyproject/galaxy/pull/10334 into the 20.09 release as well, so it's available in the next version of galaxy-tool-util. Thanks!

### DIFF
--- a/lib/galaxy/tool_util/cwl/util.py
+++ b/lib/galaxy/tool_util/cwl/util.py
@@ -84,7 +84,7 @@ def path_or_uri_to_uri(path_or_uri):
 
 
 def galactic_job_json(
-    job, test_data_directory, upload_func, copy_func, collection_create_func, tool_or_workflow="workflow"
+    job, test_data_directory, upload_func, collection_create_func, tool_or_workflow="workflow"
 ):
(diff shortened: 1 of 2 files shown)


jmchilton (approved):

lib/galaxy/tool_util/cwl/util.py
```diff
@@ -328,9 +332,10 @@ def _ensure_file_exists(file_path):
...
 class FileLiteralTarget:
 
-    def __init__(self, contents, **kwargs):
+    def __init__(self, contents, path=None, **kwargs):
         self.contents = contents
         self.properties = kwargs
+        self.path = path
 
     def __str__(self):
         return "FileLiteralTarget[path={}] with {}".format(self.path, self.properties)
```
nsoranzo:
```suggestion
    def __init__(self, contents, **kwargs):
        self.contents = contents
        self.properties = kwargs

    def __str__(self):
        return "FileLiteralTarget[contents={}] with {}".format(self.contents, self.properties)
```

mvdbeek:
Very nice, thanks so much @simonbray !

PR #10336, opened 2020-09-30
[20.09] Change k8s runner to use k8s assigned uid to track jobs rather than name

innovate-invent:
Merged into 20.09 branch


mvdbeek:
Thanks @innovate-invent, I think @nuwang has cherry-picked your commits into https://github.com/galaxyproject/galaxy/pull/10407 and we're trying to get this in now.

Issue #10324, opened 2020-09-29
new tool revision from TS can't be installed


bernt-matthias:
For `mothur_align_check` there is revision 6 (6:0399e5aeaac6) available in the TS. According to the Galaxy admin panel I have revision 1 and 5 installed and it tells me that there is a newer revision available but I can't find a way in the admin panel to install it (if I check `Show all` it also shows me only revision 1 and 5). 

It gets more confusing if I check the logs. There I find only traces of the version hashes (not sure if its the correct name) 814a34469480 and 1d7284662c8b which should be revisions 1 and 2. 

Actually I was trying to install the latest revision of `suite_mothur` (using https://github.com/galaxyproject/ansible-galaxy-tools) and in the logs I have seen `Updating repository 'mothur_align_check' to 6:0399e5aeaac6` but nothing happens. 

If I check in the TS the changes in the latest revisions mostly test data changed, but if I'm not wrong the tool version never was bumped. 

(using 20.05)

bernt-matthias:
`hg tip` for the tool installed at 814a34469480 shows `changeset:   6:0399e5aeaac6`

cat-bro:
This might overlap with #10339 if you're seeing the `Newer Version Available` tag for mothur_align_check.

I think the fact that you're seeing revision 5 in the list when revision 6 is installed has to do with how galaxy handles new revisions where the version (hence the tool id) has stayed the same.  The way this is displayed in admin (showing 5 instead of 6) is probably also a bug but wouldn't be fixed by #10339.

PR #10339, opened 2020-10-01
[20.09] Fix 'Newer Version Available' tag in admin tool management section

cat-bro:
The 'Newer Version Available' tag often shows up for tools where the newest version actually is installed.  _groupByNameOwner currently returns one repository for each name-owner but this is not necessarily the newest from the list.  This is one possible solution.

This only works if `parseInt(repository.ctx_rev)` will never throw an exception.  If there is a chance that there would be a repository that has no ctx_rev or a ctx_rev value that cannot be parsed as an integer then I should be putting try/catch around this.

### DIFF
--- a/client/src/components/Toolshed/InstalledList/Index.vue
+++ b/client/src/components/Toolshed/InstalledList/Index.vue
@@ -33,9 +33,7 @@
                     <template v-slot:cell(name)="row">
                         <b-link href="#" role="button" class="font-weight-bold" @click="row.toggleDetails">
                             <div v-if="!isLatest(row.item)">
-                                <b-badge variant="danger" class="mb-2">
-                                    Newer version available!
-                                </b-badge>
+                                <b-badge variant="danger" class="mb-2"> Newer version available! </b-badge>
                             </div>
                             <div class="name">{{ row.item.name }}</div>
                         </b-link>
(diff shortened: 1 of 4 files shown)


bernt-matthias:
Might this fix https://github.com/galaxyproject/galaxy/issues/10324

cat-bro:
For example, the tag currently shows up for abricate and for a few other tools where we do have the latest revision installed.
Screen Shot 2020-10-01 at 4 04 31 pm

mvdbeek (approved):
That looks good to me, thanks a lot!

mvdbeek:
Could you pull in https://github.com/mvdbeek/galaxy/commit/457e6cddb3fa47c33c2639389b58e91bafbbfcc0 ? This also sorts by ctx_rev on the backend. I somehow didn't manage to make a pull request against your branch.

cat-bro:
@mvdbeek thanks for reviewing.  I think that with https://github.com/mvdbeek/galaxy/commit/457e6cddb3fa47c33c2639389b58e91bafbbfcc0 sorting in the front end will not be necessary - that commit will solve the issue I have described and we don't need this PR.

mvdbeek:
I think it's probably good to assert both, I'm not sure the backend change catches all routes

mvdbeek:
Thanks again @cat-bro!

PR #10344, opened 2020-10-01
[20.09] Add missing index migrations

mvdbeek:


lib/galaxy/model/migrate/versions/util.py
```diff
@@ -168,6 +181,7 @@ def drop_index(index, table, column_name=None, metadata=None):
             if not isinstance(table, Table):
                 assert metadata is not None
                 table = Table(table, metadata, autoload=True)
+            index = truncate_index_name(index, table.metadata.bind)
```
jdavcs:
```suggestion
            index_name = truncate_index_name(index, table.metadata.bind)
```

lib/galaxy/model/migrate/versions/util.py
```diff
@@ -168,6 +181,7 @@ def drop_index(index, table, column_name=None, metadata=None):
...
             if index in [ix.name for ix in table.indexes]:
```
jdavcs:
```suggestion
            if index_name in [ix.name for ix in table.indexes]:
```

lib/galaxy/model/migrate/versions/util.py
```diff
@@ -168,6 +181,7 @@ def drop_index(index, table, column_name=None, metadata=None):
...
                 index = Index(index, table.c[column_name])
```
jdavcs:
```suggestion
                index = Index(index_name, table.c[column_name])
```

lib/galaxy/model/migrate/versions/util.py
jdavcs: (a) to keep it consistent with `add_index`; and
(b) to stay sane: one is Index, the other is its name
EDIT: even though `index` here is *not* an instance of `Index` (as per line 180), I still think such explicit naming would make it more readable

jdavcs: Also, see line 188: do we want to log the original name or the truncated name? (i.e., `index` vs `index_name` - if we rename variable; otherwise we have no choice and original name is lost)

lib/galaxy/model/migrate/versions/util.py
```diff
@@ -135,10 +144,14 @@ def add_index(index_name, table, column_name, metadata=None, **kwds):
     :param metadata: Needed only if ``table`` is a table name
     :type metadata: :class:`Metadata`
     """
+    if len(index_name) > 63 and metadata.bind.name in ('postgres', 'postgresql'):
```
jdavcs:
```suggestion
    if len(index_name) > MAX_IDENTIFIER_LEN and metadata.bind.name in ('postgres', 'postgresql'):
```
~~Define at top of file; cite https://www.postgresql.org/docs/12/runtime-config-preset.html#GUC-MAX-IDENTIFIER-LENGTH~~ Better call function, see comment^ (line 53)
So that.. no magic numbers? :-)

lib/galaxy/model/migrate/versions/util.py
```diff
@@ -47,6 +48,14 @@ def localtimestamp(migrate_engine):
         raise Exception('Unable to convert data for unknown database type: %s' % migrate_engine.name)
 
 
+def truncate_index_name(index_name, engine):
+    # does what sqlalchemy does, see https://github.com/sqlalchemy/sqlalchemy/blob/8455a11bcc23e97afe666873cd872b0f204848d8/lib/sqlalchemy/sql/compiler.py#L4696
+    max_index_name_length = engine.dialect.max_index_name_length or engine.dialect.max_identifier_length
```
jdavcs, line 53: we need this number in line 147. Factor out into a function maybe?

lib/galaxy/model/migrate/versions/util.py
mvdbeek: original makes more sense, don't think we need to forward the nitty-gritty of supposedly sqlalchemy internals.

natefoo (approved):

PR #10350, opened 2020-10-02
Script for finding missing db indexes

jdavcs:
To do:
- [x] add github workflow to run on both postgres and sqlite

### DIFF
--- a/lib/galaxy/model/mapping.py
+++ b/lib/galaxy/model/mapping.py
@@ -1125,18 +1125,18 @@ model.WorkflowInvocationOutputDatasetAssociation.table = Table(
     "workflow_invocation_output_dataset_association", metadata,
     Column("id", Integer, primary_key=True),
     Column("workflow_invocation_id", Integer, ForeignKey("workflow_invocation.id"), index=True),
-    Column("workflow_step_id", Integer, ForeignKey("workflow_step.id")),
+    Column("workflow_step_id", Integer, ForeignKey("workflow_step.id"), index=True),
     Column("dataset_id", Integer, ForeignKey("history_dataset_association.id"), index=True),
-    Column("workflow_output_id", Integer, ForeignKey("workflow_output.id")),
+    Column("workflow_output_id", Integer, ForeignKey("workflow_output.id"), index=True),
 )
 
 model.WorkflowInvocationOutputDatasetCollectionAssociation.table = Table(
     "workflow_invocation_output_dataset_collection_association", metadata,
     Column("id", Integer, primary_key=True),
     Column("workflow_invocation_id", Integer, ForeignKey("workflow_invocation.id", name='fk_wiodca_wii'), index=True),
-    Column("workflow_step_id", Integer, ForeignKey("workflow_step.id", name='fk_wiodca_wsi')),
+    Column("workflow_step_id", Integer, ForeignKey("workflow_step.id", name='fk_wiodca_wsi'), index=True),
     Column("dataset_collection_id", Integer, ForeignKey("history_dataset_collection_association.id", name='fk_wiodca_dci'), index=True),
-    Column("workflow_output_id", Integer, ForeignKey("workflow_output.id", name='fk_wiodca_woi')),
+    Column("workflow_output_id", Integer, ForeignKey("workflow_output.id", name='fk_wiodca_woi'), index=True),
 )
 
 model.WorkflowInvocationOutputValue.table = Table(
@@ -1160,7 +1160,7 @@ model.WorkflowInvocationStepOutputDatasetCollectionAssociation.table = Table(
     "workflow_invocation_step_output_dataset_collection_association", metadata,
     Column("id", Integer, primary_key=True),
     Column("workflow_invocation_step_id", Integer, ForeignKey("workflow_invocation_step.id", name='fk_wisodca_wisi'), index=True),
-    Column("workflow_step_id", Integer, ForeignKey("workflow_step.id", name='fk_wisodca_wsi')),
+    Column("workflow_step_id", Integer, ForeignKey("workflow_step.id", name='fk_wisodca_wsi'), index=True),
     Column("dataset_collection_id", Integer, ForeignKey("history_dataset_collection_association.id", name='fk_wisodca_dci'), index=True),
     Column("output_name", String(255), nullable=True),
 )
--- a/lib/galaxy/model/mapping.py
+++ b/lib/galaxy/model/mapping.py
@@ -333,14 +333,6 @@ model.ImplicitlyConvertedDatasetAssociation.table = Table(
     Column("metadata_safe", Boolean, index=True, default=True),
     Column("type", TrimmedString(255)))
 
-model.ValidationError.table = Table(
-    "validation_error", metadata,
-    Column("id", Integer, primary_key=True),
-    Column("dataset_id", Integer, ForeignKey("history_dataset_association.id"), index=True),
-    Column("message", TrimmedString(255)),
-    Column("err_type", TrimmedString(64)),
-    Column("attributes", TEXT))
-
 model.Group.table = Table(
     "galaxy_group", metadata,
     Column("id", Integer, primary_key=True),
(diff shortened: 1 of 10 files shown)


scripts/check_model.py
```diff
@@ -0,0 +1,57 @@
...
+
+    missing_indexes = set(mapping_indexes.keys()) - set(indexes_in_db.keys())
+    if missing_indexes:
+        return [(mapping_indexes[index], index.table, index.column_names) for index in missing_indexes]
+
+
+if __name__ == '__main__':
+    indexes = find_missing_indexes()
+    if indexes:
+        print(json.dumps(indexes, indent=4))
```
nsoranzo:
```suggestion
        print(json.dumps(indexes, indent=4, sort_keys=True))
```

scripts/check_model.py
```diff
@@ -0,0 +1,57 @@
...
+from collections import namedtuple
+
+sys.path.insert(1, os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir, 'lib')))
+
+from sqlalchemy import create_engine, MetaData
+
+from galaxy.model import mapping
+from galaxy.model.orm.scripts import get_config
+
+Index = namedtuple('MissingIndex', 'table column_names')
```
nsoranzo: I think this type can be confused with https://docs.sqlalchemy.org/en/13/core/constraints.html#sqlalchemy.schema.Index . Also, the first argument of `namedtuple()` is usually the name of the new type as a string. So what about:
```suggestion
IndexTuple = namedtuple('IndexTuple', 'table column_names')
```

jdavcs: Yes, definitely! Agreed on both. Will adjust the rest.

ic4f:
This is ready for review.

ic4f:
`ModuleNotFoundError: No module named 'boltons'` - not sure why this not installed for this workflow.

ic4f:
Looks like we still have 4 missing indexes.

scripts/check_model.py
```diff
@@ -0,0 +1,57 @@
...
+                index_tuple = tuple_from_index(index)
+                indexes[index_tuple] = index.name
+        return indexes
+
+    # load metadata from mapping.py
+    metadata = mapping.metadata
+    mapping_indexes = load_indexes(metadata)
+
+    # create EMPTY metadata, then load from database
+    db_url = os.environ.get("GALAXY_CONFIG_OVERRIDE_DATABASE_CONNECTION") or get_config(sys.argv)['db_url']
```
mvdbeek:
```suggestion
    db_url = get_config(sys.argv)['db_url']
```
The override handling is done by get_config.

tox.ini
```diff
@@ -21,6 +21,7 @@ setenv =
     unit: GALAXY_VIRTUAL_ENV={envdir}
     unit: GALAXY_ENABLE_BETA_COMPRESSED_GENBANK_SNIFFING=1
     mulled: GALAXY_TEST_INCLUDE_SLOW=1
+    check_indexes: GALAXY_TEST_FORCE_DATABASE_MIGRATION=1
```
mvdbeek, line 26:
```suggestion
    check_indexes: GALAXY_TEST_FORCE_DATABASE_MIGRATION=1
    check_indexes: GALAXY_SKIP_CLIENT_BUILD=1
```

scripts/check_model.py
mvdbeek: if you want to pass `GALAXY_CONFIG_OVERRIDE_DATABASE_CONNECTION` into tox you need to add this this `passenv` in tox.ini

mvdbeek:
> Looks like we still have 4 missing indexes.

that only seems to happen for sqlite ... I'll see if I can figure out why that is.

mvdbeek:
```
[
    [
        "ix_job_user_id",
        "job",
        "user_id"
    ],
    [
        "ix_history_tag_association_user_id",
        "history_tag_association",
        "user_id"
    ],
    [
        "ix_history_dataset_association_tag_association_user_id",
        "history_dataset_association_tag_association",
        "user_id"
    ],
    [
        "ix_page_tag_association_user_id",
        "page_tag_association",
        "user_id"
    ]
]

```
it's always on user_id and always a foreign key.
They're created in https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/model/migrate/versions/0031_community_and_workflow_tags.py , except for ix_job_user_id which is created in https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/model/migrate/versions/0047_job_table_user_id_column.py

This is because of https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/model/migrate/versions/util.py#L83, but that statement is not true according to https://www.sqlite.org/foreignkeys.html#fk_indexes:

> So, in most real systems, an index should be created on the child key columns of each foreign key constraint.

accordingly `.indexes job` doesn't return the ix_job_user_id

scripts/check_model.py
jdavcs: Indeed it does handle it; not at all obvious though. Thanks for both suggestions.
If this is not in developer docs (and I think it's not), I'll add it.

ic4f:
IN addition to the 4 sqlite indexes, my local testing consistently shows 2 postgres indexes:
```
[
    [
        "ix_job_to_input_dataset_collection_element_job_id",
        "job_to_input_dataset_collection_element",
        "job_id"
    ],
    [
        "ix_job_to_input_dataset_collection_element_dataset_collection_element_id",
        "job_to_input_dataset_collection_element",
        "dataset_collection_element_id"
    ]
]
```
I'm looking into this.

ic4f:
Thanks for the pointers, @mvdbeek! I've come up with something that's hacky, but works. I don't have a better solution for now, but I'll revisit this later: ideally, there should be no special case there.

lib/galaxy/model/mapping.py
```diff
@@ -985,6 +985,7 @@
     Column("from_path", TEXT),
     Column("published", Boolean, index=True, default=False),
     Index('ix_stored_workflow_slug', 'slug', mysql_length=200),
+    Index('ix_stored_workflow_from_path', 'from_path', mysql_length=200),
```
mvdbeek: That doesn't look like a super-useful index, I wouldn't add this.

lib/galaxy/model/migrate/versions/0169_add_missing_indexes.py
```diff
@@ -15,100 +15,130 @@
 metadata = MetaData()
 
 indexes = [
+    [
```
mvdbeek, line 18: You shouldn't touch the 0169 migrations. The 0169 should be merged forward, any additional changes should go to 0170.

lib/galaxy/model/migrate/versions/util.py
```diff
@@ -86,17 +86,22 @@ def add_column(column, table, metadata, **kwds):
...
             # SQLAlchemy Migrate has a bug when adding a column with both a
             # ForeignKey and an index in SQLite. Since SQLite creates an index
             # anyway, we can drop the explicit index creation.
+            # TODO: this is hacky, but it solves this^ problem. Needs better solution.
+            index_to_create = (kwds['index_name'], table, column.name)
             del kwds['index_name']
             column.index = False
         column.create(table, **kwds)
         assert column is table.c[column.name]
+        if index_to_create:
```
mvdbeek, line 103: Did you check if just removing the the if on line 93 is enough ?

jdavcs: Yep, was the first thing I did - many things broke ("no such table" errors, etc.) I didn't look in detail though - just fixed it.

lib/galaxy/model/migrate/versions/0169_add_missing_indexes.py
jdavcs: Oh, OK (I thought we were adding all of these to 20.09 as one migration) Fixing.

lib/galaxy/model/mapping.py
jdavcs: Removing.

PR #10351, opened 2020-10-02
update SQLNamedQuery for proftpd configuration

FredericBGA:
and add a link to Galaxy security passwords documentation


natefoo:
Sorry I missed this. It might be worth noting that some very old Galaxies might still have SHA1 passwords, and a configuration that supports both can be found in [the ansible role template](https://github.com/galaxyproject/ansible-proftpd/blob/master/templates/galaxy_auth.conf.j2) (which also uses split instead of substrings).

That said, we switched to PBKDF2 a loooong time ago and you could tell anyone who fails to login via FTP because they have a SHA1 password to just log in to Galaxy and change their password (which will then be stored in the PBKDF2 format). As a bonus then your user gets a more securely stored password hash.

PR #10357, opened 2020-10-04
Planemo treats discovered collections and manually collections differently

astrovsky01:
If a discovered_dataset collection is created, planemo tests treat it as a single output, no matter how many items are inside. This is as opposed to manually-created collections, which act as num_outputs + 1 for the collection itself (as they do in the history). 

This test will fail right now, because it should be seeing 10 discovered files  + 1 collection + 2 manually created files + 1 collection = 14 items out. Instead, it sees 1 discovered collection out + 2 manually created files + 1 collection = 4 items out.

Either both should act as  a single out, or both should be num_outputs + 1


test/functional/tools/collection_discover_dataset_test.xml
```diff
@@ -0,0 +1,31 @@
+<tool id='discover_test' name='discover_test' version='0.1.0' python_template_version='3.5'>
```
jmchilton: I don't think Galaxy will run tests on this if you don't add it to https://github.com/galaxyproject/galaxy/blob/dev/test/functional/tools/samples_tool_conf.xml

astrovsky01: Thanks, added!

jmchilton:
It did indeed fail.

```
    def _verify_outputs(testdef, history, jobs, tool_id, data_list, data_collection_list, galaxy_interactor, quiet=False):
        assert len(jobs) == 1, "Test framework logic error, somehow tool test resulted in more than one job."
        job = jobs[0]
    
        maxseconds = testdef.maxseconds
        if testdef.num_outputs is not None:
            expected = testdef.num_outputs
            actual = len(data_list) + len(data_collection_list)
            if expected != actual:
                message_template = "Incorrect number of outputs - expected %d, found %s."
                message = message_template % (expected, actual)
>               raise Exception(message)
E               Exception: Incorrect number of outputs - expected 14, found 4.
```

jmchilton:
xref https://github.com/galaxyproject/galaxy/pull/7894

bernt-matthias:
@astrovsky01 do you think that we can close this with the docs added in https://github.com/galaxyproject/galaxy/pull/7894 or is still something missing?

dannon:
Checked with @astrovsky01; we can close this one out now.

PR #10360, opened 2020-10-05
[20.09] Don't use subquery to update workflow invocation step update time

mvdbeek:
Thanks Nate, I don't think the subquery was needed.

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -1158,10 +1158,7 @@ class Job(JobLike, UsesCreateAndUpdateTime, Dictifiable, RepresentById):
         ''', '''
             UPDATE workflow_invocation_step
             SET update_time = :update_time
-            WHERE id in (
-                SELECT workflow_invocation_step.id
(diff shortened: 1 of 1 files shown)


lib/galaxy/model/__init__.py
```diff
@@ -1158,10 +1158,7 @@ def set_final_state(self, final_state):
         ''', '''
             UPDATE workflow_invocation_step
             SET update_time = :update_time
-            WHERE id in (
-                SELECT workflow_invocation_step.id
-                FROM workflow_invocation_step wis
-                WHERE wis.job_id = :job_id
+            WHERE job_id = :job_id
             );
```
dannon:
```suggestion
            WHERE job_id = :job_id;
```

dannon: Or am I missing the other paren somewhere?

dannon (approved):

jmchilton:
Sorry again - thanks so much for the fix and for the debugging.

PR #10364, opened 2020-10-06
convert data sets only if necessary

### LINKED ISSUE #9002: data_ref does not work for tables that are not of type tabular
It seems that `data_ref` only works for data sets that are of type tabular. For other tabular data, e.g. bed files, it does not show columns. 

In the GTN Galaxy 101 tutorial the Group tool is applied on the result of Join (which is bed). The data_column input:

```
<param name="groupcol" label="Group by column" type="data_column" data_ref="input1" />
```

does not show a selection.

bernt-matthias:
fixes: https://github.com/galaxyproject/galaxy/issues/9002

### DIFF
--- a/lib/galaxy/datatypes/registry.py
+++ b/lib/galaxy/datatypes/registry.py
@@ -883,13 +883,19 @@ class Registry:
         return None
 
     def find_conversion_destination_for_dataset_by_extensions(self, dataset_or_ext, accepted_formats, converter_safe=True):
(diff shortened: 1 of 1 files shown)


bernt-matthias:
@mvdbeek Seems that the change has quite some side effects. Maybe its simpler to move the test to the data column parameter class? But is it the better option?

bernt-matthias:
Superseeded by https://github.com/galaxyproject/galaxy/pull/10365  ..

PR #10365, opened 2020-10-06
add test for data_column parameter referring to non-tabular data (bed) and fix bug

bernt-matthias:
To check why https://github.com/galaxyproject/galaxy/issues/9002 is unnoticed by CI.

- [x] with https://github.com/galaxyproject/galaxy/pull/10365/commits/4c5c1350a30fc4af85143d4f4d5904ea2abf0e0b exactly the new test fails.
- [x] starting to fix using the commit(s) of https://github.com/galaxyproject/galaxy/pull/10364 (which will be closed)
- [x] check if https://github.com/galaxyproject/galaxy/blob/2777db0e46f1185e47c41049abdb61f99ac1e2e7/lib/galaxy/tools/parameters/dataset_matcher.py#L222 is necessary
- [ ] use named tuple as return

### DIFF
--- a/lib/galaxy/datatypes/display_applications/parameters.py
+++ b/lib/galaxy/datatypes/display_applications/parameters.py
@@ -89,7 +89,9 @@ class DisplayApplicationDataParameter(DisplayApplicationParameter):
                 rval = data.get_converted_files_by_type(ext)
                 if rval:
                     return rval
-            assert data.find_conversion_destination(self.formats)[0] is not None, "No conversion path found for data param: %s" % self.name
+
+            direct_match, target_ext, converted_dataset = data.find_conversion_destination(self.formats)
+            assert direct_match or target_ext is not None, "No conversion path found for data param: %s" % self.name
             return None
         return data
 
@@ -107,8 +109,12 @@ class DisplayApplicationDataParameter(DisplayApplicationParameter):
             # start conversion
             # FIXME: Much of this is copied (more than once...); should be some abstract method elsewhere called from here
             # find target ext
-            target_ext, converted_dataset = data.find_conversion_destination(self.formats, converter_safe=True)
-            if target_ext and not converted_dataset:
+            direct_match, target_ext, converted_dataset = data.find_conversion_destination(self.formats, converter_safe=True)
+            # TODO: Q: I guess we could skip the first if branch but code might
+            # be more readable
+            if direct_match:
+                pass
+            elif target_ext and not converted_dataset:
                 if isinstance(data, DisplayDataValueWrapper):
                     data = data.value
                 new_data = next(iter(data.datatype.convert_dataset(trans, data, target_ext, return_output=True, visible=False).values()))
--- a/lib/galaxy/datatypes/display_applications/parameters.py
+++ b/lib/galaxy/datatypes/display_applications/parameters.py
@@ -110,22 +110,19 @@ class DisplayApplicationDataParameter(DisplayApplicationParameter):
             # FIXME: Much of this is copied (more than once...); should be some abstract method elsewhere called from here
             # find target ext
             direct_match, target_ext, converted_dataset = data.find_conversion_destination(self.formats, converter_safe=True)
-            # TODO: Q: I guess we could skip the first if branch but code might
-            # be more readable
-            if direct_match:
-                pass
-            elif target_ext and not converted_dataset:
-                if isinstance(data, DisplayDataValueWrapper):
-                    data = data.value
-                new_data = next(iter(data.datatype.convert_dataset(trans, data, target_ext, return_output=True, visible=False).values()))
-                new_data.hid = data.hid
-                new_data.name = data.name
-                trans.sa_session.add(new_data)
-                assoc = trans.app.model.ImplicitlyConvertedDatasetAssociation(parent=data, file_type=target_ext, dataset=new_data, metadata_safe=False)
-                trans.sa_session.add(assoc)
-                trans.sa_session.flush()
-            elif converted_dataset and converted_dataset.state == converted_dataset.states.ERROR:
-                raise Exception("Dataset conversion failed for data parameter: %s" % self.name)
+            if not direct_match:
+                if target_ext and not converted_dataset:
+                    if isinstance(data, DisplayDataValueWrapper):
+                        data = data.value
+                    new_data = next(iter(data.datatype.convert_dataset(trans, data, target_ext, return_output=True, visible=False).values()))
+                    new_data.hid = data.hid
+                    new_data.name = data.name
+                    trans.sa_session.add(new_data)
+                    assoc = trans.app.model.ImplicitlyConvertedDatasetAssociation(parent=data, file_type=target_ext, dataset=new_data, metadata_safe=False)
+                    trans.sa_session.add(assoc)
+                    trans.sa_session.flush()
+                elif converted_dataset and converted_dataset.state == converted_dataset.states.ERROR:
+                    raise Exception("Dataset conversion failed for data parameter: %s" % self.name)
         return self.get_value(other_values, dataset_hash, user_hash, trans)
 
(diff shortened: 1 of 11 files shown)


mvdbeek:
> * is necessary

yes, that's how you discriminate a direct match from no match available. You could eliminate this if statement (and similar constructs in other places that call find_conversion_destination_for_dataset_by_extensions) if you change the return value though to something structured, like a named tuple with `direct_match, converted_ext, converted_dataset`

lib/galaxy/datatypes/display_applications/parameters.py
```diff
@@ -107,8 +109,12 @@ def prepare(self, other_values, dataset_hash, user_hash, trans):
             # start conversion
             # FIXME: Much of this is copied (more than once...); should be some abstract method elsewhere called from here
             # find target ext
-            target_ext, converted_dataset = data.find_conversion_destination(self.formats, converter_safe=True)
-            if target_ext and not converted_dataset:
+            direct_match, target_ext, converted_dataset = data.find_conversion_destination(self.formats, converter_safe=True)
+            # TODO: Q: I guess we could skip the first if branch but code might
+            # be more readable
+            if direct_match:
```
mvdbeek: I'd say `if not direct_match:` and then have the other 2 if statements below seems more readable ?

lib/galaxy/tools/actions/__init__.py
```diff
@@ -88,8 +88,8 @@ def process_dataset(data, formats=None):
                 if not data.datatype.matches_any(formats):
                     # Need to refresh in case this conversion just took place, i.e. input above in tool performed the same conversion
                     trans.sa_session.refresh(data)
-                    target_ext, converted_dataset = data.find_conversion_destination(formats)
-                    if target_ext:
+                    direct_match, target_ext, converted_dataset = data.find_conversion_destination(formats)
```
mvdbeek: You don't need the `if not data.datatype.matches_any(formats)` statement above, you can replace that with 
```
direct_match, target_ext, converted_dataset = data.find_conversion_destination(formats)
if not direct_match:
````

File: lib/galaxy/tools/parameters/dataset_matcher.py [hunk 3]
```diff
@@ -121,8 +121,8 @@ def valid_hda_match(self, hda, check_implicit_conversions=True):
         else:
             if not check_implicit_conversions:
                 return False
-            target_ext, converted_dataset = hda.find_conversion_destination(formats)
-            if target_ext:
+            direct_match, target_ext, converted_dataset = hda.find_conversion_destination(formats)
```
mvdbeek: Same here, you can pull that up to line 119

lib/galaxy/datatypes/display_applications/parameters.py
bernt-matthias: OK.

lib/galaxy/tools/parameters/dataset_matcher.py
bernt-matthias: I think I covered all three comments .. please check my changes.

lib/galaxy/datatypes/display_applications/parameters.py
```diff
@@ -107,19 +109,20 @@ def prepare(self, other_values, dataset_hash, user_hash, trans):
...
-                new_data.hid = data.hid
-                new_data.name = data.name
-                trans.sa_session.add(new_data)
-                assoc = trans.app.model.ImplicitlyConvertedDatasetAssociation(parent=data, file_type=target_ext, dataset=new_data, metadata_safe=False)
-                trans.sa_session.add(assoc)
-                trans.sa_session.flush()
-            elif converted_dataset and converted_dataset.state == converted_dataset.states.ERROR:
-                raise Exception("Dataset conversion failed for data parameter: %s" % self.name)
+            direct_match, target_ext, converted_dataset = data.find_conversion_destination(self.formats, converter_safe=True)
+            if direct_match:
```
mvdbeek:
```suggestion
            if not direct_match:
```

bernt-matthias: You are right. I amended the change to my previous commit

lib/galaxy/datatypes/display_applications/parameters.py
```diff
@@ -107,19 +109,20 @@ def prepare(self, other_values, dataset_hash, user_hash, trans):
...
+                elif target_ext and not converted_dataset:
```
mvdbeek:
```suggestion
                if target_ext and not converted_dataset:
```

bernt-matthias: done

bernt-matthias:
Seems that only using a named tuple as return is left .. if the tests pass .. or?

bernt-matthias:
Is named tuple already used for a return statement somewhere else in Galaxy, seems quite a bit of (programmin) overhead to me...

mvdbeek:
It should just be 1 or 2 extra lines depending on the import, but the way it is now is fine, you don't need to change it.

bernt-matthias:
@mvdbeek I'm a bit puzzled on how to fix the tests (because I don't get all the details). I guess we need a different value for https://github.com/galaxyproject/galaxy/blob/de87b0516a6a2e627d6c5902aabfa4fdee0aab0a/test/unit/tools/test_data_parameters.py#L193 .. but I'm unsure on how to set them in the correct way (if my guess is correct)

mvdbeek:
> .. but I'm unsure on how to set them in the correct way (if my guess is correct)

I think the idea is that you set them to what is appropriate for the current case being tested, I've done that in https://github.com/galaxyproject/galaxy/pull/10365/commits/8ee0348f5e3ddcdc8eb06de628ef2e85847c28b0. `(None, None)` was ambivalent previously, so I think that's a clear improvement.

mvdbeek (approved):

mvdbeek:
Can you run the iuc tests against this branch, just to make sure there are no weird side-effects here. Otherwise I think this is good to go.

bernt-matthias:
https://github.com/galaxyproject/tools-iuc/runs/1258905375?check_suite_focus=true

bernt-matthias:
47 test(s) out of 3057 failed .. to me the number seems higher than it should be .. but I don't see an error that is obviously related ..

mvdbeek:
I agree, I think a new openmpi build is breaking a lot of tests by writing to `/`. Then there's a bunch where the collection order has changed. Only scpipe concerns me a little, I'm not sure what's going on there.

bernt-matthias:
Wild guess: scpipe has a gff3 input and the test gives gff3.gz ..

bernt-matthias:
scpipe test works locally for 20.09 and this branch .. with and without containers .. odd

mvdbeek:
Yeah, weird. One thing I saw in the CI test logs is
```
Error in scan(file = file, what = what, sep = sep, quote = quote, dec = dec,  : 
  line 1 did not have 3 elements
Calls: calculate_QC_metrics -> get_genes_by_GO -> getBM -> read.table -> scan
```
That's not there if I run the scpipe test locally.

mvdbeek:
Actually only the first scpipe test passed locally, there rest failed with
```
sed: can't create temp file '/Users/mvandenb/src/galaxy/database/files/1/8/1/dataset_181db884-d5e2-4d34-bd2a-cf93b855636e.dat9F62Z8': Read-only file system
```

mvdbeek:
https://github.com/galaxyproject/tools-iuc/blob/master/tools/scpipe/scpipe.xml#L77, I'll open a PR. But I don't think any of this is related to this PR.

bernt-matthias:
I guess this deserves backporting?

mvdbeek:
Sure, 20.09 or earlier?

bernt-matthias:
For me 20.09 would be sufficient. I plan to upgrade soon-ish.

PR #10367, opened 2020-10-06
[20.09] Defer metadata strategy resolution in job wrapper

natefoo:
While developing a dynamic rule, I discovered that determining the metadata strategy in the JobWrapper init method causes the resolution of the job destination at wrapper creation time. This can have some adverse performance impacts and also result in less precise destination selection (ideally destination resolution happens as close to dispatching as possible). It also means that destination resolution was happening outside `JobHandler.__verify_job_ready()`, where exceptions in job destination resolution are properly handled.


mvdbeek (approved):
Sorry about that, I was going back and forth over whether to use a property or not when I added this

natefoo:
Nurries!

mvdbeek:
The mulled unit test failure is unrelated, seems bamtools just gained an additional test.

PR #10368, opened 2020-10-07
Drop support for retired Python 3.5

nsoranzo:
Upgrade syntax using `pyupgrade --py36-plus` .

Manually drop several `six` imports.

Also:
- Remove broken pr_cache in scripts/bootstrap_history.py
- Fix broken prefix removal in lib/galaxy/tool_util/deps/mulled/mulled_build.py


mvdbeek:
Probably time to move the remaining tests to github CI ? Otherwise we'd have to update the Docker images to use a more recent Python version.

lib/galaxy/jobs/__init__.py
```diff
@@ -1883,14 +1883,14 @@ def get_session_id(self):
...
-        return '[ -f "{}" ] && . {}'.format(self.app.config.environment_setup_file, self.app.config.environment_setup_file)
+        return f'[ -f "{self.app.config.environment_setup_file}" ] && . {self.app.config.environment_setup_file}'
 
     def get_input_dataset_fnames(self, ds):
         filenames = []
         filenames = [ds.file_name]
         # we will need to stage in metadata file names also
         # TODO: would be better to only stage in metadata files that are actually needed (found in command line, referenced in config files, etc.)
-        for key, value in ds.metadata.items():
+        for value in ds.metadata.values():
```
mvdbeek: ds.metadata is  MetadataCollection instance and doesn't have a values() method. MetadataCollection overwrites `__getattr__` to return None if the attribute is not found, so that explains the framework test failure. To be on the safe side I'd suggest
```suggestion
        for _, value in ds.metadata.items():
```
should there ever be a metadata element called values. Otherwise I guess we could add a values method ?

nsoranzo: Thanks for looking into that, I debugged it this morning and came to the same conclusions. Fixed in a rebase, I'm looking into implement a proper `Mapping`.

nsoranzo:
> Probably time to move the remaining tests to github CI ? Otherwise we'd have to update the Docker images to use a more recent Python version.

Agreed on the way forward, but doesn't need to be part of this already large PR since our Docker image uses Python 3.6:
```shell
$ docker run -it --entrypoint /bin/bash galaxy/testing-base:20.05.00 
root@d69f98c55cd6:/# python3 --version
Python 3.6.9
```

mvdbeek:
And I was wondering why the tests passed!

mvdbeek (approved):

nsoranzo:
By the way, I have not made breaking changes to the Python scripts in `tools/` since most of the corresponding tools don't specify a requirement for a Python version yet.

PR #10371, opened 2020-10-07
[20.09] Fix db-skip-locked invocation grabbing for pool members

mvdbeek:


innovate-invent:
I can confirm this works within my deployment. Thanks!

PR #10372, opened 2020-10-07
Tool recommendation: Add querystrings

anuprulez:
This PR adds query strings to the URLs fetching tool recommendations in order to get statistics of the system's usage over time by searching for specific keywords in the Galaxy logs.


dannon:
Do we really want all this extra info in the logs?  It seems like it might be better to just instrument the tool recommendation endpoint to log to a separate logger if desired.

mvdbeek:
Thanks @anuprulez, I think what @dannon suggested is the right way to do this. Let us know if you need any help with this.

PR #10373, opened 2020-10-07
Limit CPUs used in docker containers

bernt-matthias:
to avoid CPU over utilization of tools


bgruening:
backport to 20.09?

bernt-matthias:
> backport to 20.09?

Unsure if this is a good idea.

bernt-matthias:
Seems that the same is unfortunately not possible for singularity (wo root priviiges) .. I intended to switch "soon" to singularity, but I can't root on our cluster.

PR #10380, opened 2020-10-07
Adding nodeSelector to K8S runner job pod spec

almahmoud:
Similar to `affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms` (which is already supported through `k8s_affinity` and could be a way to accomplish what we need), `nodeSelector` is used to determine which nodes the job can run on.
This is specifically needed in Anvil to restrict jobs run by a user to that user's nodepool(s).

### DIFF
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -64,7 +64,7 @@ class KubernetesJobRunner(AsynchronousJobRunner):
             k8s_run_as_group_id=dict(map=str, valid=lambda s: s == "$gid" or s.isdigit()),
             k8s_fs_group_id=dict(map=int),
             k8s_cleanup_job=dict(map=str, valid=lambda s: s in {"onsuccess", "always", "never"}, default="always"),
-            k8s_pod_retries=dict(map=int, valid=lambda x: int >= 0, default=3),
+            k8s_pod_retries=dict(map=int, valid=lambda x: int(x) >= 0, default=3),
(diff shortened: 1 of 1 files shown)


jmchilton:
Test failures seem conda related, this PR looks good to me.

dannon:
This looks fine, still WIP?

innovate-invent:
While convenient this can also be accomplished by:
```yaml
k8s_affinity: |
              nodeAffinity:
                requiredDuringSchedulingIgnoredDuringExecution:
                  nodeSelectorTerms:
                  - matchExpressions:
                    - key: WorkClass
                      operator: In
                      values:
                      - compute
```

`WorkClass` is the label name and `compute` is its value for nodes

almahmoud:
WIP cause needs to be tested, and John recommended just switching to the job conf YML from XML so that the dictionary is not ugly, and I haven't gotten to do that to test yet. So tldr I think it should just work, but I haven't tested it yet

almahmoud:
@innovate-invent yes, I mentioned in the original comment that it can be accomplished through that, but we're using nodeSelector for everything else so figured i'd add it here as well since it's a small change

almahmoud:
Tested and works as expected.
https://github.com/galaxyproject/galaxy/pull/10380/commits/bc85b9c82a0eead2b9d05bcdb080f6e08f95e114 needs to be backported. I can make a PR with just that commit targeting an older version for the backport.
The nodeSelector feature would be nice to add to 20.09 if still possible. If not, we can cherry-pick for AnVIL and add to 21.01.

PR #10381, opened 2020-10-08
Fix assigning default value to variables in ``run_test_wrapper.sh``

nsoranzo:
`GALAXY_TEST_UID`, `GALAXY_VIRTUAL_ENV` and `GALAXY_TEST_DATABASE_TYPE` were not assigned the default value due to using `:-` instead of `:=` .

Also clean up and fixes suggested by `shellcheck`.


mvdbeek:
Thanks @nsoranzo!

PR #10387, opened 2020-10-09
When compiling dynamic installation confs, ensure location is actually writeable 

dannon:
This makes it so, for example in CVMFS-using deployments, you can't accidentally try to install a tool to a non-writeable shed conf.

### DIFF
--- a/lib/galaxy/tools/toolbox/base.py
+++ b/lib/galaxy/tools/toolbox/base.py
@@ -216,11 +216,13 @@ class AbstractToolBox(Dictifiable, ManagesIntegratedToolPanelMixin):
             )
 
         if parsing_shed_tool_conf:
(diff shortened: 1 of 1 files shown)


lib/galaxy/tools/toolbox/base.py
```diff
@@ -216,11 +216,13 @@ def _init_tools_from_config(self, config_filename):
             )
 
         if parsing_shed_tool_conf:
-            shed_tool_conf_dict = dict(config_filename=config_filename,
-                                       tool_path=tool_path,
-                                       tool_cache_data_dir=tool_cache_data_dir,
-                                       config_elems=config_elems)
-            self._dynamic_tool_confs.append(shed_tool_conf_dict)
+            # if read_only mode, (CVMFS consumer) don't add to dynamic_confs
```
mvdbeek, line 219: This is great!

PR #10394, opened 2020-10-09
add recurse attribute for discover_datasets

bernt-matthias:
for output collections. seems functional but `planemo lint` complains.


bernt-matthias:
Should I backport to 20.09? Otherwise planemo will continue to bug me with a tool that I'm just preparing.

jmchilton:
Yeah -- that would be a good idea! Thanks!

PR #10396, opened 2020-10-09
[20.09] backport #10394 to 20.09 (add recurse attribute for discover_datasets)

bernt-matthias:
https://github.com/galaxyproject/galaxy/pull/10394


bernt-matthias:
Planemo still seems to complain .. https://github.com/galaxyproject/tools-iuc/pull/3232/checks?check_run_id=1250062108 .. do we need this also on another branch (master?) .. or is a change in planemo required?

nsoranzo:
I think we need a release of the `tool-util` package.

PR #10397, opened 2020-10-09
Add k8s recommended labels

innovate-invent:


mvdbeek:
I think that's included https://github.com/galaxyproject/galaxy/pull/10407

PR #10403, opened 2020-10-12
[20.09] Python 3 fix and client build fix for the proxy

bgruening:
Caught by building the containers.

### DIFF
--- a/tools/sr_assembly/velvetg_wrapper.py
+++ b/tools/sr_assembly/velvetg_wrapper.py
@@ -12,11 +12,6 @@ import sys
 assert sys.version_info[:2] >= (2, 6)
 
 
-def stop_err(msg):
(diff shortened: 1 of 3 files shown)


tools/sr_assembly/velvetg_wrapper.py
```diff
@@ -29,7 +29,7 @@ def __main__():
         proc = subprocess.Popen(args=cmdline, shell=True, stderr=subprocess.PIPE)
         returncode = proc.wait()
         # get stderr, allowing for case where it's very large
-        stderr = ''
+        stderr = b''
```
mvdbeek, line 27: Can you also change `stop_err` to just do `sys.exit(msg)` so we get a non-zero exit code ?

mvdbeek (approved):

PR #10404, opened 2020-10-12
Adds Visualization options to Markdown Editor

guerler:
This PR augments the Markdown editor by displaying a form for visualization options before inserting the markdown text. The visualization options are then parsed and populated by the visualization plugin before rendering the visualization. This PR also allows to parameterize all other visualizations by parsing `url` parameters to them corresponding to parameters defined in the visualization xml.


bgruening:
This is awesome! @guerler do you think we can make the parameters more like `foo.bar` instead of `foo|bar`.

jmchilton:
@bgruening Is that a good idea? It would look vaguely more like other programming languages but that isn't how the tool XML works or the workflow formats or the API, etc.. - I'm nervous about having another incompatible way to reference tool parameters. Especially if it is purely aesthetics, that syntax isn't shorter or more structured.

bgruening:
> @bgruening Is that a good idea? It would look vaguely more like other programming languages but that isn't how the tool XML works or the workflow formats or the API, etc.. - I'm nervous about having another incompatible way to reference tool parameters. Especially if it is purely aesthetics, that syntax isn't shorter or more structured.

You are correct. This is simply syntactic suguar. I would argue that we should use what is best for this particular user-base and we expect biologist to write this. Not sure the overlap with tool-writers and API users are that big. But I understand your point its not a big deal.

John are we using `|` in workfow formats and API? In tool XML we can avoid using them afaik by being more explicit.

guerler:
@galaxybot test this

config/plugins/visualizations/media_player/audio_player/config/audio_player.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE visualization SYSTEM "../../visualization.dtd">
-<visualization name="Audio player">
+<visualization name="Audio player" embeddable="false">
```
bgruening: why do you think audio and video should not be playable in a report?

bgruening (approved):
This works great for me and it is a very awesome user feature. Reports will rock hard in 2021!

It would be nice if we could add more tests, maybe a Selenium test as well. @OlegZharkov do you have time to add a Selenium test once this is merged?

config/plugins/visualizations/media_player/audio_player/config/audio_player.xml
guerler: I wanted reports to be printable and exportable e.g. as PDF but maybe that is too restrictive. Also I noticed in a test that the video player might have an issue with playing (some?) `.mp4` files.

guerler: I enabled them for now, maybe we can revisit the audio player and video player in a paper cut issue #10829.

bgruening: > I wanted reports to be printable and exportable e.g. as PDF but maybe that is too restrictive. Also I noticed in a test that the video player might have an issue with playing (some?) .mp4 files.

I think that is a good aim, but we should not restrict more advanced use-cases. We could offer a warning or better fallback to an image taking from a random/first frame.

bgruening (approved):

bgruening:
Thanks @guerler!

bgruening:
Thanks @guerler this is super cool!

PR #10407, opened 2020-10-12
[20.09] K8s selector change

nuwang:
This is a continuation of: https://github.com/galaxyproject/galaxy/pull/10307

It combines the work in PR https://github.com/galaxyproject/galaxy/pull/10315 with the modifications discussed in the original PR.

### DIFF
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -209,8 +209,11 @@ class KubernetesJobRunner(AsynchronousJobRunner):
                     "app.kubernetes.io/instance": self.__produce_k8s_job_prefix(ajs.job_wrapper),
                     "app.kubernetes.io/version": ajs.tool_wrapper.tool.version,
                     "app.kubernetes.io/component": "tool",
-                    "app.kubernetes.io/part-of": self._galaxy_instance_id or "galaxy",
+                    "app.kubernetes.io/part-of": "galaxy",
                     "app.kubernetes.io/managed-by": "galaxy",
+                    "app.galaxy.org/job_id": ajs.job_wrapper.get_id_tag(),
+                    "app.galaxy.org/instance": self._galaxy_instance_id or "",
+                    "app.galaxy.org/handler": self.app.config.server_name,
                 }
             },
             "spec": {
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -214,6 +214,7 @@ class KubernetesJobRunner(AsynchronousJobRunner):
                     "app.galaxyproject.org/job_id": ajs.job_wrapper.get_id_tag(),
                     "app.galaxyproject.org/instance": self._galaxy_instance_id or "",
                     "app.galaxyproject.org/handler": self.app.config.server_name,
+                    "app.galaxyproject.org/destination": ajs.job_wrapper.job_destination.id,
                 }
             },
             "spec": {
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -131,14 +131,14 @@ class KubernetesJobRunner(AsynchronousJobRunner):
         job.create()
         job_id = job.labels.get(JOB_ID_LABEL, False)
         if not job_id:
-            ## Recover uid label because it wasn't set
-            #job.labels[JOB_ID_LABEL] = job.metadata.uid
-            #job.update()
-            #job_id = job.labels.get(JOB_ID_LABEL, False)
-            #if not job_id:
-            job_wrapper.fail("Unexpected value from job runner", exception=True)
-            log.exception("%s not assigned by k8s to job on invocation: %s" % (JOB_ID_LABEL, job.obj))
-            return
+            # Recover uid label because it wasn't set
+            job.labels[JOB_ID_LABEL] = job.metadata.uid
+            job.update()
+            job_id = job.labels.get(JOB_ID_LABEL, False)
+            if not job_id:
+                job_wrapper.fail("Unexpected value from job runner", exception=True)
+                log.exception("%s not assigned by k8s to job on invocation: %s" % (JOB_ID_LABEL, job.obj))
+                return
 
         # define job attributes in the AsyncronousJobState for follow-up
         ajs.job_id = job_id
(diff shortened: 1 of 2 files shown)


nuwang:
ping @innovate-invent @mvdbeek

mvdbeek:
Can you take a look at the failing kubernetes tests ?

innovate-invent:
I tried applying this PR as the current k8s runner is broken given the 20.01 PR.
```
  File "/srv/galaxy/lib/galaxy/jobs/runners/__init__.py", line 137, in run_next
    method(arg)
  File "/srv/galaxy/lib/galaxy/jobs/runners/kubernetes.py", line 129, in queue_job
    self.__get_k8s_job_spec(ajs)
  File "/srv/galaxy/lib/galaxy/jobs/runners/kubernetes.py", line 195, in __get_k8s_job_spec
    k8s_job_spec = {"template": self.__get_k8s_job_spec_template(ajs),
  File "/srv/galaxy/lib/galaxy/jobs/runners/kubernetes.py", line 209, in __get_k8s_job_spec_template
    "app.kubernetes.io/name": ajs.tool_wrapper.tool.id,
AttributeError: 'AsynchronousJobState' object has no attribute 'tool_wrapper'
galaxy.jobs.runners.kubernetes ERROR 2020-10-13 18:43:02,652 (229919/None) User killed running job, but error encountered during termination: job name must not be empty
  File "/srv/galaxy/lib/galaxy/jobs/runners/kubernetes.py", line 512, in stop_job
  File "/srv/galaxy/lib/galaxy/jobs/runners/util/pykube_util.py", line 64, in find_job_object_by_name
    raise ValueError("job name must not be empty")
ValueError: job name must not be empty
```

nuwang:
@innovate-invent I've added you as a collaborator, so feel free to add your label enhancements on top.

innovate-invent:
Does it make sense to keep that as a seperate PR? There are some complications with it.

nuwang:
Sure. That sounds good too. Ideally, we'll get both into 20.09. The label enhancements will be really useful to trace back from the k8s job to the Galaxy job id, since that information is no longer in the name. The job prefix is capped at 23 characters, leaving the rest for a UUID, which is of length 36 chars.

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -226,7 +204,21 @@ def __get_k8s_job_spec_template(self, ajs):
         (see pod selector) and an appropriate restart policy."""
         k8s_spec_template = {
             "metadata": {
-                "labels": {"app": self.__produce_unique_k8s_job_name(ajs.job_wrapper.get_id_tag())[:-5]}
+                "labels": {
+                    "app.kubernetes.io/name": ajs.job_wrapper.tool.tool_id,
+                    "app.kubernetes.io/instance": self.__produce_unique_k8s_job_name(ajs.job_wrapper.get_id_tag()),
+                    "app.kubernetes.io/version": ajs.job_wrapper.tool.version,
+                    "app.kubernetes.io/component": "tool",
```
mvdbeek, line 211: Doesn't job make more sense here ?
```suggestion
                    "app.kubernetes.io/component": "job",
```

nuwang: Actually, since this is a pod spec, tool might be more appropriate?

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -226,7 +204,21 @@ def __get_k8s_job_spec_template(self, ajs):
         (see pod selector) and an appropriate restart policy."""
         k8s_spec_template = {
             "metadata": {
-                "labels": {"app": self.__produce_unique_k8s_job_name(ajs.job_wrapper.get_id_tag())[:-5]}
+                "labels": {
+                    "app.kubernetes.io/name": ajs.job_wrapper.tool.tool_id,
```
```suggestion
                    "app.kubernetes.io/name": ajs.job_wrapper.tool.id,
```

mvdbeek (approved):
Thanks a lot @innovate-invent @nuwang. The relevant tests are passing, so I will merge this later today if there's nothing else coming up.

lib/galaxy/jobs/runners/kubernetes.py
innovate-invent: the component is a tool, the job is the reason for running it

lib/galaxy/jobs/runners/kubernetes.py
innovate-invent: tool.id returns the fully resolved tool id uri. Labels can not contain the special characters that are required for a uri.
I took a guess that tool_id is only the id as specified in the tool xml. This was also why I was leaving this as a separate PR because I am guessing at the correct variables to reference. This is also why I moved tool.id to an annotation.

mvdbeek: Well, tool.tool_id is not a thing. What special characters are not allowed ? Do you maybe want to use `tool.old_id`, which is just the `id` as it appears in the tool xml ?

innovate-invent: ah, thanks! yea, old_id is what I was after.

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -226,7 +204,21 @@ def __get_k8s_job_spec_template(self, ajs):
...
+                    "app.kubernetes.io/component": "tool",
+                    "app.kubernetes.io/part-of": "galaxy",
+                    "app.kubernetes.io/managed-by": "galaxy",
+                    "app.galaxyproject.org/job_id": ajs.job_wrapper.get_id_tag(),
+                    "app.galaxyproject.org/instance": self._galaxy_instance_id or "",
+                    "app.galaxyproject.org/handler": self.app.config.server_name,
+                    "app.galaxyproject.org/destination": ajs.job_wrapper.job_destination.id,
+                },
+                "annotations": {
+                    "app.galaxyproject.org/tool_id": ajs.job_wrapper.tool.id,
```
nuwang, line 220: @mvdbeek I just noticed that this is is still referring to tool.id?

mvdbeek: From https://github.com/galaxyproject/galaxy/pull/10407#discussion_r504816337 I understood this only applies to labels. Is that not correct ?

nuwang: Aah yes, that makes sense. Label selectors need to be dns compliant/url friendly.

Issue #10016, opened 2020-07-20
Augment Workflow Model for Registeries


jmchilton:
I was needing to add workflow authors to the workflow model for Dockstore (the current user is a fine guess I suppose, but it should be a persistent and user editable field IMO).

At the BCC 2020 IWC BoF though people working workflowhub.eu were willing to discuss minimum workflow metadata standards with me and so we went through the work they had done and also isolated "license" (as a URI coming from https://spdx.org/licenses/ presumably) and version information as additional data we would like to capture and make user-specifiable in the UI.

nsoranzo:
During the BCC 2020 Collaboration Fest we worked on a list of possible licenses to use for software and/or content (both can apply to workflows I think): https://docs.google.com/document/d/16vnRtDjrx5eHSl4jXs2vMaDTI6luyyLzU6xMvRHsnbI/edit?usp=sharing

mvdbeek:
I'd say this is done with https://github.com/galaxyproject/galaxy/pull/10408

PR #10408, opened 2020-10-12
Microdata for tools and workflows.

jmchilton:
The aim here is to allow schema.org annotations for creators and SPDX for license on tools and workflows. My primary goal was to enhance the Dockstore plugin to capture more metadata it expects (and allow workflowhub.eu to also take advantage of this metadata).

There is a lot left to do here - but the basic interaction with the workflow editor seems to vaguely work - including linking out to orcid.org and embedded metadata into the page via microdata.

Screen Shot 2020-10-12 at 4 15 41 PM

Screen Shot 2020-10-12 at 4 14 28 PM

Implements #10016


nsoranzo:
> The aim here is to allow schema.org annotations for creators and SPDX for license on tools and workflows. My primary goal was to enhance the Dockstore plugin to capture more metadata it expects (and allow workflowhub.eu to also take advantage of this metadata).

Awesome, thanks for working on this!
 
>     * Prune the default list of licenses we expose to users (we can have metadata on all of them, but the dropdown should be smaller by default. There was some work from Assunta and Nicola on what that list should be, I think I just need to add a visible property or something on the list based on their work).

Here's what we have done during the BCC2020 Collaboration Fest: https://docs.google.com/document/d/16vnRtDjrx5eHSl4jXs2vMaDTI6luyyLzU6xMvRHsnbI/edit?usp=sharing

jmchilton:
I've integrated the recommended list y'all put together @nsoranzo. The workflow can annotate anything from SPDX and it will use the metadata, but the dropdown of options will only show the ones from the recommended list.

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -184,6 +185,17 @@ automatically if profile is set to any version newer than ``16.01``, such as
...
         </xs:annotation>
       </xs:attribute>
+      <xs:attribute name="license" type="xs:string">
+        <xs:annotation>
+          <xs:documentation xml:lang="en">This string specifies any full URI or a
+a short SPDX (https://spdx.org/licenses/) identifier for a license for this tool
+wrapper. The tool wrapper version can be indepedent of the underlying tool. This
+license covers the tool XML and associated scripts shipped with the tool.
+
+This is interpreted as a shcema.org license attribute (see also https://schema.org/license).
```
mr-c:
```suggestion
This is interpreted as a schema.org license attribute (see also https://schema.org/license).
```

nsoranzo:
> I've integrated the recommended list y'all put together @nsoranzo. The workflow can annotate anything from SPDX and it will use the metadata, but the dropdown of options will only show the ones from the recommended list.

Thanks, fantastic!

mvdbeek (approved):

PR #10413, opened 2020-10-13
Update authnz to include Azure as IDP

cginmn:
Updated the psa_authz.py with azure backend to enable Azure as an IDP which is already supported by the used python-social-core components. Updated the oidc_backend sample config to document how to configure Azure as IDP. Very simple change. Only required two lines to pass configuration info to the python-social-auth mechanism galaxy uses.


mvdbeek (approved):
Awesome, thank you very much!

PR #10416, opened 2020-10-13
[20.01] Restrict k8s max job name length to 63 and shorten prefix

nuwang:


innovate-invent:
You should also include a check that instance name is not longer than 63 - len('gxy') - len(uuid).
Truncating the concatenated string will cause collisions as you truncate the uuid removing its uniqueness.
This check should result in an error being logged.

nuwang:
@innovate-invent It's trimming from the right to prevent the uuid from getting truncated. I'll add an assert in the 20.09 PR to ensure that the total prefix length is less than 30 characters

Issue #10328, opened 2020-09-29
In data collections, unless history item is expanded, edit button will lead to incorrect dataset


astrovsky01:
<img width="1383" alt="Screen Shot 2020-09-29 at 12 12 08 PM" src="https://user-images.githubusercontent.com/40246333/94604887-2582d280-024d-11eb-95c8-ff4b780a3f05.png">
<img width="1383" alt="Screen Shot 2020-09-29 at 12 11 58 PM" src="https://user-images.githubusercontent.com/40246333/94604896-29165980-024d-11eb-8baa-b53379279fd2.png">

When attempting to edit information about a dataset inside a collection by clicking the pencil icon, if the history item is not clicked and expanded first, the edit screen that appears will not correspond to the dataset clicked. Once the item is expanded, this goes away, and appears to stay away for that item specifically, but continues for all other datasets until they are expanded in turn.

dannon:
Similar issue to #10311

mvdbeek:
Fixed in https://github.com/galaxyproject/galaxy/pull/10427

PR #10427, opened 2020-10-14
Fix handling of collection element identifiers in history panel

### LINKED ISSUE #10311: scratchbook with dataset collection oddity
Sorry for not knowing a better title for this, but what I found is this:
When you have a collection and you're trying to view one of the datasets contained within it with the scratchbook enabled, then what's getting displayed are the contents of what seems to be a random other dataset (maybe even from other users?).
When you view the same dataset without the scratchbook, or with the scratchbook but the dataset expanded first, then things work as they should.

guerler:
Fixes #10311 and #10328.

### DIFF
--- a/client/src/mvc/dataset/dataset-li-edit.js
+++ b/client/src/mvc/dataset/dataset-li-edit.js
@@ -66,8 +66,9 @@ var DatasetListItemEdit = _super.extend(
                     const Galaxy = getGalaxyInstance();
                     if (Galaxy.router) {
                         ev.preventDefault();
+                        const identifier = self.model.get("element_id") || self.model.get("id");
                         Galaxy.router.push("datasets/edit", {
-                            dataset_id: self.model.attributes.id,
+                            dataset_id: identifier,
                         });
                     }
                 },
--- a/client/src/mvc/dataset/dataset-li.js
+++ b/client/src/mvc/dataset/dataset-li.js
@@ -184,7 +184,8 @@ export var DatasetListItemView = _super.extend(
                     const Galaxy = getGalaxyInstance();
                     if (Galaxy.frame && Galaxy.frame.active) {
                         // Add dataset to frames.
-                        Galaxy.frame.addDataset(self.model.get("id"));
+                        const identifier = self.model.get("element_id") || self.model.get("id");
+                        Galaxy.frame.addDataset(identifier);
                         ev.preventDefault();
                     }
                 };


client/src/mvc/dataset/dataset-li-edit.js
```diff
@@ -66,8 +66,9 @@ var DatasetListItemEdit = _super.extend(
                     const Galaxy = getGalaxyInstance();
                     if (Galaxy.router) {
                         ev.preventDefault();
+                        const identifier = self.model.get("element_id") || self.model.get("id");
                         Galaxy.router.push("datasets/edit", {
-                            dataset_id: self.model.attributes.id,
+                            dataset_id: identifier,
```
mvdbeek, line 71: `element_id` refers to `DatasetCollectionElement` instances, which you can't mix with HDA ids (`self.model.get("id")`). Unless the model change is taken into account elsewhere you're just going to display a random dataset (if it is not private).

guerler: Interesting. Thank you for looking into this. I assumed that all model attributes starting with `element_` are specific to that element e.g. `element_type` is `hda`. `element_name` also seems to be correct. Would that not mean that the `element_id` is inconsistent? How else can we retrieve the actual hda identifier?

mvdbeek: Let me look into this to verify I'm not making this more complicated than necessary. In general `DatasetCollectionElement`  instances can be `hda` or `hdca`, and element id should refer to the id of a `DatasetCollectionElement`. From the dataset collection element you can get the dataset, so I would change https://github.com/guerler/galaxy/blob/4db6cc3c31c52a5ff5f3af43700a4e1fdda269b3/lib/galaxy/webapps/galaxy/controllers/dataset.py#L190 to accept element_id. But let me see if we don't already have access to the hda id at this point.

mvdbeek: OK, this is correct because of https://github.com/guerler/galaxy/blob/4db6cc3c31c52a5ff5f3af43700a4e1fdda269b3/client/src/mvc/collection/collection-model.js#L69 ... which is maybe not a good idea because we actually have something that is the element_id, while in this piece of code element_id is either a HDA id or HDCA id. Pretty confusing and that tripped me up. Still, as a bugfix this is fine.

mvdbeek (approved):
If you pull it out of draft I'm happy to merge!

client/src/mvc/dataset/dataset-li-edit.js
guerler: Cool, thanks for double checking it.

mvdbeek:
We should probably Backport this until 20.01

PR #10432, opened 2020-10-15
Filter duplicate workflow output labels when initializing workflow step outputs

guerler:
This PR introduces an additional condition to prevent the insertion of duplicate output labels when inserting workflow nodes in the workflow editor. To reproduce this, open the workflow editor and insert a tool with explicitly labeled workflow outputs multiple times. You will notice that the label displayed in the `Outputs` subsection of the corresponding workflow step forms are duplicates. Additionally this PR adds a call to properly cleanup output labels when removing workflow steps. Overall these changes improve the workflow label handling.

### DIFF
--- a/lib/galaxy/managers/workflows.py
+++ b/lib/galaxy/managers/workflows.py
@@ -815,7 +815,7 @@ class WorkflowContentsManager(UsesAnnotations):
                     if output_label is not None:
                         if output_label in output_label_index:
                             if output_label not in output_label_duplicate:
-                                output_label_duplicate.append(output_label)
+                                output_label_duplicate.add(output_label)
                         else:
                             output_label_index[output_label] = True
             step_dict['workflow_outputs'] = outputs
--- a/lib/galaxy/managers/workflows.py
+++ b/lib/galaxy/managers/workflows.py
@@ -817,7 +817,7 @@ class WorkflowContentsManager(UsesAnnotations):
                             if output_label not in output_label_duplicate:
                                 output_label_duplicate.add(output_label)
                         else:
-                            output_label_index[output_label] = True
+                            output_label_index.add(output_label)
             step_dict['workflow_outputs'] = outputs
             if len(output_label_duplicate) > 0:
                 output_label_duplicate_string = ", ".join(output_label_duplicate)
--- a/lib/galaxy/managers/workflows.py
+++ b/lib/galaxy/managers/workflows.py
@@ -738,6 +738,7 @@ class WorkflowContentsManager(UsesAnnotations):
         log.info("creator_metadata is %s" % workflow.creator_metadata)
         data['creator'] = workflow.creator_metadata
 
+        output_label_index = {}
         input_step_types = set(workflow.input_step_types)
         # For each step, rebuild the form and encode the state
         for step in workflow.steps:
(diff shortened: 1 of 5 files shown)


mvdbeek:
These are awesome fixes that are difficult to diagnose for users, we should probably include them in 20.09.
Just one question, is the output label guaranteed to be removed from the copy, not the original ?

guerler:
> These are awesome fixes that are difficult to diagnose for users, we should probably include them in 20.09.
> Just one question, is the output label guaranteed to be removed from the copy, not the original ?

Yes, if I understand the question correctly. The output which is initialized first is associated with that label, until that outputs label changes, no other output can be assigned to that label. When that output changes its label to a new unique label, the old label is removed and becomes available again.

mvdbeek:
And the initialization is guaranteed to have happened for the original first ? Could we highlight the conflicting output label, like we do for the step labels now ?

guerler:
> And the initialization is guaranteed to have happened for the original first ? Could we highlight the conflicting output label, like we do for the step labels now ?

Right now, steps are enumerated by the order they have been inserted into the `steps` array i.e. the workflow. Upon loading label duplication is prevented by removing reoccuring labels, i.e. only the first occurance is preserved. If a user enters an label which already exists, the error is highlighted like shown here: 
image

mvdbeek:
> Upon loading label duplication is prevented by removing reoccuring labels, i.e. only the first occurance is preserved.

Will there be any information visible to the user when this happens ? I worry about automatically changing the workflow when it isn't 100% clear and unambiguous which label should be removed. Your example of copying a step is not ambiguous at all, but maybe it would be better to drop the label only when copying steps ?

guerler:
Added an upgrade message to indicate issues with duplicate labels:

image

lib/galaxy/managers/workflows.py
```diff
@@ -737,14 +737,12 @@ def _workflow_to_dict_editor(self, trans, stored, workflow, tooltip=True, is_sub
...
-                data['upgrade_messages'][step.order_index] = upgrade_message
-            if (hasattr(module, "version_changes")) and (module.version_changes):
-                if step.order_index in data['upgrade_messages']:
-                    data['upgrade_messages'][step.order_index][module.tool.name] = "\n".join(module.version_changes)
-                else:
-                    data['upgrade_messages'][step.order_index] = {module.tool.name: "\n".join(module.version_changes)}
+            upgrade_message_dict = {}
+            check_and_update_messages = module.check_and_update_state()
+            if check_and_update_messages:
+                upgrade_message_dict = check_and_update_messages
```
mvdbeek:
```suggestion
            upgrade_message_dict = module.check_and_update_state() or {}
```

lib/galaxy/managers/workflows.py
```diff
@@ -796,6 +791,7 @@ def callback(input, prefixed_name, **kwargs):
 
             # workflow outputs
             outputs = []
+            output_label_duplicate = []
```
mvdbeek:
```suggestion
            output_label_duplicate = set()
```

lib/galaxy/managers/workflows.py
```diff
@@ -804,7 +800,18 @@ def callback(input, prefixed_name, **kwargs):
                     outputs.append({"output_name": output_name,
                                     "uuid": output_uuid,
                                     "label": output_label})
+                    if output_label is not None:
+                        if output_label in output_label_index:
+                            if output_label not in output_label_duplicate:
+                                output_label_duplicate.append(output_label)
```
mvdbeek:
```suggestion
                                output_label_duplicate.add(output_label)
```

lib/galaxy/managers/workflows.py
```diff
@@ -726,7 +726,7 @@ def _workflow_to_dict_editor(self, trans, stored, workflow, tooltip=True, is_sub
         data['steps'] = {}
         data['upgrade_messages'] = {}
         data['report'] = workflow.reports_config or {}
-
+        output_label_index = {}
```
mvdbeek:
```suggestion
        output_label_index = set()
```

lib/galaxy/managers/workflows.py
```diff
@@ -804,7 +800,18 @@ def callback(input, prefixed_name, **kwargs):
                     outputs.append({"output_name": output_name,
                                     "uuid": output_uuid,
                                     "label": output_label})
+                    if output_label is not None:
+                        if output_label in output_label_index:
+                            if output_label not in output_label_duplicate:
+                                output_label_duplicate.append(output_label)
+                        else:
+                            output_label_index[output_label] = True
```
mvdbeek:
```suggestion
                            output_label_index.add(output_label)
```

jmchilton (approved):

guerler:
Rebased it due to minor conflicts.

PR #10434, opened 2020-10-15
[20.09] Rework new collection order testing in 20.09.

jmchilton:
This was added in 20.09 with https://github.com/galaxyproject/galaxy/pull/9684/files.

There were some things I didn't love that I think are corrected here.

- Preserve legacy behavior of not enforcing sort order but make it now contigent on profile being 20.09 or newer.
- Run all the element tests and then do order checking for a more specific error message: ``galaxy.tool_util.verify.interactor.JobOutputsError: Collection identifier '1' found out of order, expected order of ['2', '1', '10'] for the tool generated collection elements ['1', '2', '3', '4', '5', '6', '7', '8', '9', '10']``
- Add test case for older default based on new test case.
- Fix the new attribute that is used to check the order so that it isn't required (this broke Planemo workflow testing) and so the name doesn't clash with an existing element attribute on collection elements in the API/model.

### DIFF
--- a/lib/galaxy/tool_util/parser/xml.py
+++ b/lib/galaxy/tool_util/parser/xml.py
@@ -598,7 +598,7 @@ def __parse_output_collection_elem(output_collection_elem, profile=None):
 
 
 def __parse_element_tests(parent_element, profile=None):
-    element_tests = OrderedDict()
+    element_tests = {}
     for idx, element in enumerate(parent_element.findall("element")):
         element_attrib = dict(element.attrib)
         identifier = element_attrib.pop('name', None)
--- a/lib/galaxy/tool_util/parser/xml.py
+++ b/lib/galaxy/tool_util/parser/xml.py
@@ -513,7 +513,8 @@ class XmlToolSource(ToolSource):
 
         if tests_elem is not None:
             for i, test_elem in enumerate(tests_elem.findall("test")):
-                tests.append(_test_elem_to_dict(test_elem, i))
+                profile = self.parse_profile()
+                tests.append(_test_elem_to_dict(test_elem, i, profile))
 
         return rval
 
@@ -533,10 +534,10 @@ class XmlToolSource(ToolSource):
         return python_template_version
 
 
-def _test_elem_to_dict(test_elem, i):
+def _test_elem_to_dict(test_elem, i, profile=None):
     rval = dict(
         outputs=__parse_output_elems(test_elem),
-        output_collections=__parse_output_collection_elems(test_elem),
+        output_collections=__parse_output_collection_elems(test_elem, profile=profile),
         inputs=__parse_input_elems(test_elem, i),
         expect_num_outputs=test_elem.get("expect_num_outputs"),
         command=__parse_assert_list_from_elem(test_elem.find("assert_command")),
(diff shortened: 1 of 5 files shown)


lib/galaxy/tool_util/parser/xml.py
```diff
@@ -579,36 +580,38 @@ def __parse_command_elem(test_elem):
...
     for idx, element in enumerate(parent_element.findall("element")):
         element_attrib = dict(element.attrib)
         identifier = element_attrib.pop('name', None)
         if identifier is None:
             raise Exception("Test primary dataset does not have a 'identifier'")
-        element_tests[identifier] = __parse_test_attributes(element, element_attrib, parse_elements=True)
-        element_tests[identifier][1]["element_index"] = idx
+        element_tests[identifier] = __parse_test_attributes(element, element_attrib, parse_elements=True, profile=profile)
+        if profile and profile >= "20.09":
+            element_tests[identifier][1]["expected_sort_order"] = idx
```
bernt-matthias, line 609: Do we need `expected_sort_order` at all? Now that we use `OrderedDict`.

bernt-matthias: But then the choice of the dictionary should depend on the profile.

bernt-matthias:
I guess we also need to document somewhere that the check for the order depends on profile >= 20.09

lib/galaxy/tool_util/parser/xml.py
jmchilton: Actually I'm not quite sure the ordered dictionary is doing anything or really should be, this all gets serialized via an API request. So it is better to be explicit.

I will revert that and find a place in the documentation to mention this behavior switch.

bernt-matthias:
Since you just touched this, maybe you have an idea for https://github.com/galaxyproject/galaxy/pull/10304 (which documents that tests involving collections are not working for all but the first test).

jmchilton:
Yeah - I can confirm there is something funky even in this branch. How odd.

jmchilton:
I'm not sure it belongs in this PR because all collection checks on second+ tests are skipped 😟 , but a fix seems to be 

```diff
diff --git a/lib/galaxy/tool_util/verify/interactor.py b/lib/galaxy/tool_util/verify/interactor.py
index 6801b41d4a..f70fd27648 100644
--- a/lib/galaxy/tool_util/verify/interactor.py
+++ b/lib/galaxy/tool_util/verify/interactor.py
@@ -1126,7 +1126,7 @@ class ToolTestDescription:
         self.error = processed_test_dict.get("error", False)
         self.exception = processed_test_dict.get("exception", None)

-        self.output_collections = map(TestCollectionOutputDef.from_dict, processed_test_dict.get("output_collections", []))
+        self.output_collections = list(map(TestCollectionOutputDef.from_dict, processed_test_dict.get("output_collections", [])))
         self.command_line = processed_test_dict.get("command_line", None)
         self.command_version = processed_test_dict.get("command_version", None)
         self.stdout = processed_test_dict.get("stdout", None)
```

mvdbeek (approved):

mvdbeek:
@bernt-matthias do you think this is OK ?

bernt-matthias:
Thanks everybody for looking into this. 

One more question. Was just playing around and was wondering why a tool would generate the following collection order: `'PopA_01.tags', 'PopA_02.tags', 'PopA_01.snps', 'PopA_02.snps', 'PopA_01.alleles', 'PopA_02.alleles'`

PR #10436, opened 2020-10-15
Support storing container cache in Galaxy DB

innovate-invent:
I need a core dev to help with accessing the sqlalchemy connection on line 1117

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -20,8 +20,10 @@ import time
 from datetime import timedelta
 
 import yaml
+from sqlalchemy import MetaData
 from beaker.cache import CacheManager
 from beaker.util import parse_cache_config_options
+from beaker.ext.sqla import make_cache_table
 from six.moves import configparser
 
 from galaxy.config.schema import AppSchema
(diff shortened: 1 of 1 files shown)


mvdbeek:
Can we close this now that https://github.com/galaxyproject/galaxy/pull/10451 is merged ? I don't think the cache should be stored in Galaxy's database.

innovate-invent:
The issue is that the cache is stored on a filesystem. This only works if worker replicas share the same filesystem. I am slowly working towards removing the dependency Galaxy has on a NFS, this is an important step. 
Why shouldn't the cache be stored in the Galaxy database? The cache is currently a database (I think SQLite) on the FS.

innovate-invent:
ping @mvdbeek

mvdbeek:
I still don't think that's right.

innovate-invent:
Can you propose an alternative to storing it on a NFS?

mvdbeek:
You can store it in another database, with a separate connection string.

innovate-invent:
Right! Sorry, I am conflating database with DBMS.
Can you advise on how to plumb in a new connection? Do I simply directly import sqlalchemy or does Galaxy have an abstraction over that?

This will require adding a configuration parameter, is there any preferred way of accessing that?

PR #10437, opened 2020-10-16
Let python-irodsclient session manage connections

kxk302:
With https://github.com/irods/python-irodsclient/pull/218 python-irodsclient will check when a connection was last used and create a new one after a configurable duration, avoiding connection timeouts we saw previously that we addressed with 
https://github.com/galaxyproject/galaxy/pull/9944

This PR reverts back to maintaining/reusing a session in Galaxy. This will avoid the performance overhead of creating/dropping a session for each request.


lib/galaxy/objectstore/irods.py
```diff
@@ -111,31 +110,6 @@ def parse_config_xml(config_xml):
         raise
```
kxk302, line 112: Got rid of context manager for the session.

lib/galaxy/objectstore/irods.py
```diff
@@ -235,8 +209,42 @@ def __init__(self, config, config_dict):
             raise Exception(IRODS_IMPORT_MESSAGE)
 
         self.home = "/" + self.zone + "/home/" + self.username
+        self._initialize()
         log.debug("irods __init__ %s", reload_timer)
 
+
+    def shutdown(self):
```
kxk302, line 221: This does the cleanup.

lib/galaxy/objectstore/irods.py
```diff
@@ -235,8 +209,42 @@ def __init__(self, config, config_dict):
             raise Exception(IRODS_IMPORT_MESSAGE)
 
         self.home = "/" + self.zone + "/home/" + self.username
+        self._initialize()
```
kxk302: This configures a session.

lib/galaxy/objectstore/irods.py
```diff
@@ -235,8 +209,42 @@ def __init__(self, config, config_dict):
...
+
+    def _configure_connection(self, host='localhost', port='1247', user='rods', password='rods', zone='tempZone', timeout=30, poolsize=1):
+        session = iRODSSession(host=host, port=port, user=user, password=password, zone=zone)
+        # Set connection timeout
+        session.connection_timeout = timeout
+        # Throws NetworkException if connection fails
+        try:
+            # We will create as many conections as poolsize.
+            # After creating the connection, release  it
+            # so it goes into the idle connection queue
```
kxk302: Connections are created then released, so they are ready to be used.

lib/galaxy/objectstore/irods.py
```diff
@@ -295,43 +303,41 @@ def _get_cache_path(self, rel_path):
...
-            try:
-                data_obj = session.data_objects.get(data_object_path)
-                return data_obj.__sizeof__()
-            except (DataObjectDoesNotExist, CollectionDoesNotExist):
-                log.warn("Collection or data object (%s) does not exist", data_object_path)
-                return -1
-            except NetworkException as e:
-                log.exception(e)
-                return -1
+       p = Path(rel_path)
```
kxk302: Instead of a managed session, we use the global session (self.session)

kxk302:
@mvdbeek I ran Nate's workflow (which splits a file into multiple files and hashes each) to compare the performance. The split file section was almost 3 times faster. The hash section was slightly faster. 

I also had galaxy use my local python-irodsclient code (which has the refresh logic).

mvdbeek:
Can you pull in an extra-commit that uses your branch of python-irodsclient, so that we can deploy this on test.galaxyproject.org ?

kxk302 (commented):
> Can you pull in an extra-commit that uses your branch of python-irodsclient, so that we can deploy this on test.galaxyproject.org ?

Sure, just pushed the files.

kxk302:
When we deploy Galaxy to Test, if the venv created from scratch? If not, we may have to pip uninstall python_irodsclient v0.8.3 to avoid having 2 version of the code in venv.

mvdbeek:
No, but pip manages this correctly.

lib/galaxy/dependencies/conditional-requirements.txt
```diff
@@ -8,7 +8,8 @@ drmaa
 statsd
 docker
 azure-storage==0.32.0
-python-irodsclient==0.8.3
+# python-irodsclient==0.8.3
+# git+https://github.com/kxk302/python-irodsclient.git
```
mvdbeek:
```suggestion
git+https://github.com/kxk302/python-irodsclient.git
```

lib/galaxy/dependencies/conditional-requirements.txt
```diff
@@ -8,7 +8,8 @@ drmaa
 statsd
 docker
 azure-storage==0.32.0
-python-irodsclient==0.8.3
+# python-irodsclient==0.8.3
+python-irodsclient@https://github.com/kxk302/python-irodsclient/archive/master.zip
```
mvdbeek: mhh, the `'@https://`' part gets ignored, so that doesn't pick up the right dependency

lib/galaxy/objectstore/irods.py
```diff
@@ -235,8 +209,40 @@ def __init__(self, config, config_dict):
...
+            # We will create as many conections as poolsize.
+            # After creating the connection, release  it
+            # so it goes into the idle connection queue
+            for idx in range(poolsize):
+                conn = session.pool.get_connection()
+                session.pool.release_connection(conn)
+        except NetworkException as e:
+            log.error('Could not create iRODS session: ' + str(e))
+            raise
+        return session
```
mvdbeek: You can remove all of this, that's just a remnant of one connection per irods interaction. We really just need to setup one session when setting up the objectstore and close the session when tearing down. Pool management is done by python-irodsclient.

kxk302: Ok, removed the code and simplified it.

lib/galaxy/dependencies/conditional-requirements.txt
kxk302: Added whats in dev-requirements.txt here, in place of the commented out line.

kxk302: BTW, I changed the protocol to http

kxk302:
@mvdbeek I revised the code per your suggestions. The python-irodsclient change is being merged and we will soon have version 0.8.4, which includes the connection refresh logic. I will wait for that to be merged/published, then will update the dependency files to refer to 0.8.4, instead of my Github repo.

PR #10444, opened 2020-10-16
[19.09] Fixing bug in k8s runner's k8s_pod_retrials

almahmoud:
This should forward port without conflicts up until 20.05.
Followed by https://github.com/galaxyproject/galaxy/pull/10443 for 20.09

PR #10445, opened 2020-10-16
[19.01] Fixing bug in k8s runner's k8s_pod_retrials

almahmoud:
Should be able to forward port until 19.05.
Followed by https://github.com/galaxyproject/galaxy/pull/10444 for 19.09->20.05,
Followed by #10443 for 20.09


mvdbeek:
We support the releases going back 1 year, so the last one supported is 19.09.

Issue #10309, opened 2020-09-28
Display invocation scheduling errors with more details


mvdbeek:
```
galaxy.workflow.run ERROR 2020-09-28 04:37:55,594 Failed to schedule Workflow[id=<>,name=Workflow constructed from history 'imported: test2'], problem occurred on WorkflowStep[index=2,type=tool].
Traceback (most recent call last):
  File "/cvmfs/test.galaxyproject.org/galaxy/lib/galaxy/workflow/run.py", line 190, in invoke
    incomplete_or_none = self._invoke_step(workflow_invocation_step)
  File "/cvmfs/test.galaxyproject.org/galaxy/lib/galaxy/workflow/run.py", line 266, in _invoke_step
    use_cached_job=self.workflow_invocation.use_cached_job)
  File "/cvmfs/test.galaxyproject.org/galaxy/lib/galaxy/workflow/modules.py", line 1754, in execute
    workflow_resource_parameters=resource_parameters
  File "/cvmfs/test.galaxyproject.org/galaxy/lib/galaxy/tools/execute.py", line 103, in execute
    execute_single_job(execution_slice, completed_jobs[i])
  File "/cvmfs/test.galaxyproject.org/galaxy/lib/galaxy/tools/execute.py", line 71, in execute_single_job
    job, result = tool.handle_single_execution(trans, rerun_remap_job_id, execution_slice, history, execution_cache, completed_job, collection_info, flush_job=False)
  File "/cvmfs/test.galaxyproject.org/galaxy/lib/galaxy/tools/__init__.py", line 1656, in handle_single_execution
    raise e
  File "/cvmfs/test.galaxyproject.org/galaxy/lib/galaxy/tools/__init__.py", line 1647, in handle_single_execution
    flush_job=flush_job,
  File "/cvmfs/test.galaxyproject.org/galaxy/lib/galaxy/tools/__init__.py", line 1736, in execute
    return self.tool_action.execute(self, trans, incoming=incoming, set_output_hid=set_output_hid, history=history, **kwargs)
  File "/cvmfs/test.galaxyproject.org/galaxy/lib/galaxy/tools/actions/__init__.py", line 292, in execute
    history, inp_data, inp_dataset_collections, preserved_tags, all_permissions = self._collect_inputs(tool, trans, incoming, history, current_user_roles, collection_info)
[...]

PR #10459, opened 2020-10-20
[WIP][20.09] Restore exception handling for item accessibility problems.

jmchilton:
Shouldn't this fix https://github.com/galaxyproject/galaxy/issues/10309? This way the exception falls through the object collecting exceptions and reporting them back instead of short circuiting the workflow scheduling stuff.

This was added in https://github.com/galaxyproject/galaxy/commit/f5c3478ad42102dd5f14879830ba60519e07de98#diff-17974c8048767aa536063088ba97dbc9c4475e7589002c3b2335f9e8490fc920R1600 but it seems unrelated to the rest of that commit?

### DIFF
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -1654,7 +1654,7 @@ class Tool(Dictifiable):
             if len(rval) == 4:
                 execution_slice.datasets_to_persist = rval[2]
                 execution_slice.history = rval[3]
-        except (webob.exc.HTTPFound, exceptions.MessageException) as e:
+        except webob.exc.HTTPFound as e:
             # if it's a webob redirect exception, pass it up the stack
             raise e
         except ToolInputsNotReadyException as e:


mvdbeek:
I think that only changes the logged traceback, but doesn't fix https://github.com/galaxyproject/galaxy/issues/10309.
I do think we should be raising MessageExceptions, so that tool submissions show the proper exception in the UI.

Maybe we can catch MessageExceptions somewhere in the scheduling stack and set some attribute on the invocation object that we can display in the UI ?

jmchilton:
> Maybe we can catch MessageExceptions somewhere in the workflow scheduling stack and set some attribute on the invocation object that we can display in the UI ?

I guess I meant it fixes this part:

> The job polling should also stop at this point, so users don't expect that things might still occur.

You're absolutely right about catching the message and passing it back to the user though. That should be straight-forward and a huge win.

mvdbeek:
> The job polling should also stop at this point, so users don't expect that things might still occur.

Ah, that's true, but I think this prevents passing MessageExceptions back to the user when submitting a tool job (I think).
I should have written a bit more in the commit message.

jmchilton:
Yeah - that makes sense. I wonder if there is a way to propagate the errors better in the tool submission case.

jmchilton:
> Ah, that's true, but I think this prevents passing MessageExceptions back to the user when submitting a tool job (I think).
I should have written a bit more in the commit message.

I spent some time thinking about throwing the exception - but that seems wrong. We're dumping the errors out https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/api/tools.py#L529 -- shouldn't the client just handle them better? Maybe we can make sure the errors are more structured as well.

Issue #10186, opened 2020-09-03
galaxy-tool-util  throws exception


innovate-invent:
Python 3.8.5
pip install galaxy-tool-util
```
> mulled-search --destination docker conda --search vsearch
Traceback (most recent call last):
  File "/home/nolan/Projects/galaxy-container/venv/bin/mulled-search", line 10, in <module>
    sys.exit(main())
  File "/home/nolan/Projects/galaxy-container/venv/lib/python3.8/site-packages/galaxy/tool_util/deps/mulled/mulled_search.py", line 321, in main
    conda_results[item] = conda.get_json(item)
  File "/home/nolan/Projects/galaxy-container/venv/lib/python3.8/site-packages/galaxy/tool_util/deps/mulled/mulled_search.py", line 123, in get_json
    raw_out, err, exit_code = run_command(
TypeError: 'NoneType' object is not callable
```

nsoranzo:
Fixed in https://github.com/galaxyproject/galaxy/pull/10465 .

PR #10465, opened 2020-10-20
[20.09] Fix mulled-search to not assume conda APIs from vanilla environments

jmchilton:
Fixes https://github.com/galaxyproject/galaxy/issues/10186

### DIFF
--- a/lib/galaxy/tool_util/deps/mulled/mulled_search.py
+++ b/lib/galaxy/tool_util/deps/mulled/mulled_search.py
@@ -120,6 +120,8 @@ class CondaSearch():
         Function takes search_string variable and returns results from the bioconda channel in JSON format
 
         """
+        if run_command is None:
+            raise Exception("Search with destination conda can only be run from a Conda environment.")
         raw_out, err, exit_code = run_command(
             'search', '-c',
             self.channel,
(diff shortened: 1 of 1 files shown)


lib/galaxy/tool_util/deps/mulled/mulled_search.py
```diff
@@ -120,6 +120,8 @@ def get_json(self, search_string):
         Function takes search_string variable and returns results from the bioconda channel in JSON format
 
         """
+        if run_command is None:
+            raise Exception("Search with destination conda can only be run from a Conda environment.")
```
mvdbeek: You can also install `galaxy-tool-util[mulled]` (https://github.com/galaxyproject/galaxy/pull/10025)

mvdbeek (approved):

lib/galaxy/tool_util/deps/mulled/mulled_search.py
jmchilton: Oh awesome! I didn't realize pip install conda would work from outside conda - they had pushed back on that years ago. This is cool!

mvdbeek: It's just a small part of the API unfortunately, but it's enough for search to work

jmchilton: Revised the error messages to make this usage more clear.

Issue #10469, opened 2020-10-20
K8S runner labels bug


almahmoud:
```
galaxy.jobs.runners DEBUG 2020-10-20 17:47:16,326 Job [15] queued (28.435 ms)
galaxy.jobs.handler INFO 2020-10-20 17:47:16,332 (15) Job dispatched
galaxy.jobs.runners.kubernetes DEBUG 2020-10-20 17:47:16,339 Starting queue_job for job 15
galaxy.tool_util.deps.containers INFO 2020-10-20 17:47:16,456 Checking with container resolver [ExplicitContainerResolver[]] found description [None]
galaxy.tool_util.deps.containers INFO 2020-10-20 17:47:16,456 Checking with container resolver [MulledDockerContainerResolver[namespace=biocontainers]] found description [None]
galaxy.jobs.command_factory INFO 2020-10-20 17:47:16,503 Built script [/galaxy/server/database/jobs_directory/000/15/tool_script.sh] for tool command [python /galaxy/server/database/tools/filters/convert_characters.py --strip --condense /galaxy/server/database/objects/5/0/6/dataset_506841d8-b4c0-48b0-a004-a9f9e12ecf27.dat s /galaxy/server/database/objects/e/c/9/dataset_ec990ad4-30e1-42da-9b60-1165391a616a.dat]
galaxy.jobs.runners DEBUG 2020-10-20 17:47:16,525 (15) command is: out="${TMPDIR:-/tmp}/out.$$" err="${TMPDIR:-/tmp}/err.$$"
mkfifo "$out" "$err"
trap 'rm "$out" "$err"' EXIT
tee -a stdout.log < "$out" &
tee -a stderr.log < "$err" >&2 & mkdir -p working outputs configs
if [ -d _working ]; then
    rm -rf working/ outputs/ configs/; cp -R _working working; cp -R _outputs outputs; cp -R _configs configs
else
    cp -R working _working; cp -R outputs _outputs; cp -R configs _configs
fi
cd working; /bin/bash /galaxy/server/database/jobs_directory/000/15/tool_script.sh > ../outputs/tool_stdout 2> ../outputs/tool_stderr > '/galaxy/server/database/jobs_directory/000/15/galaxy_15.o' 2> '/galaxy/server/database/jobs_directory/000/15/galaxy_15.e'; return_code=$?; sh -c "exit $return_code"
[...]

PR #10470, opened 2020-10-20
[20.09] Move more non-dns compliant labels to annotations

nuwang:
There are more labels which are not necessarily dns compliant. Have moved more into annotations.

Fixes: https://github.com/galaxyproject/galaxy/issues/10469

### DIFF
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -42,6 +42,8 @@ class KubernetesJobRunner(AsynchronousJobRunner):
     """
     runner_name = "KubernetesRunner"
 
+    LABEL_REGEX = re.compile("[^-A-Za-z0-9_.]")
+
     def __init__(self, app, nworkers, **kwargs):
         # Check if pykube was importable, fail if not
         ensure_pykube()
@@ -205,19 +207,19 @@ class KubernetesJobRunner(AsynchronousJobRunner):
         k8s_spec_template = {
             "metadata": {
                 "labels": {
-                    "app.kubernetes.io/name": ajs.job_wrapper.tool.old_id,
+                    "app.kubernetes.io/name": self.LABEL_REGEX.sub("_", ajs.job_wrapper.tool.old_id),
                     "app.kubernetes.io/instance": self.__produce_k8s_job_prefix(),
-                    "app.kubernetes.io/version": ajs.job_wrapper.tool.version,
+                    "app.kubernetes.io/version": self.LABEL_REGEX.sub("_", str(ajs.job_wrapper.tool.version)),
                     "app.kubernetes.io/component": "tool",
                     "app.kubernetes.io/part-of": "galaxy",
                     "app.kubernetes.io/managed-by": "galaxy",
-                    "app.galaxyproject.org/job_id": ajs.job_wrapper.get_id_tag(),
-                    "app.galaxyproject.org/instance": self._galaxy_instance_id or "",
-                    "app.galaxyproject.org/handler": self.app.config.server_name,
-                    "app.galaxyproject.org/destination": ajs.job_wrapper.job_destination.id,
+                    "app.galaxyproject.org/job_id": self.LABEL_REGEX.sub("_", ajs.job_wrapper.get_id_tag()),
+                    "app.galaxyproject.org/handler": self.LABEL_REGEX.sub("_", self.app.config.server_name),
+                    "app.galaxyproject.org/destination": self.LABEL_REGEX.sub(
+                        "_", str(ajs.job_wrapper.job_destination.id))
                 },
                 "annotations": {
-                    "app.galaxyproject.org/tool_id": ajs.job_wrapper.tool.id,
+                    "app.galaxyproject.org/tool_id": ajs.job_wrapper.tool.id
                 }
             },
             "spec": {


innovate-invent:
app.kubernetes.io/* need to remain labels as they are part of the k8s recommendation. We might need to look at rolling a dns encoding regex to handle weird values.

nuwang:
@innovate-invent Thanks for pointing that out. Have adjusted to match, leaving name and version out for inclusion in a future version.

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -205,19 +205,19 @@ def __get_k8s_job_spec_template(self, ajs):
...
-                    "app.kubernetes.io/name": ajs.job_wrapper.tool.old_id,
-                    "app.kubernetes.io/instance": self.__produce_k8s_job_prefix(),
-                    "app.kubernetes.io/version": ajs.job_wrapper.tool.version,
                     "app.kubernetes.io/component": "tool",
                     "app.kubernetes.io/part-of": "galaxy",
                     "app.kubernetes.io/managed-by": "galaxy",
-                    "app.galaxyproject.org/job_id": ajs.job_wrapper.get_id_tag(),
-                    "app.galaxyproject.org/instance": self._galaxy_instance_id or "",
-                    "app.galaxyproject.org/handler": self.app.config.server_name,
-                    "app.galaxyproject.org/destination": ajs.job_wrapper.job_destination.id,
```
innovate-invent, line 217: job_id, handler, instance, and destination are important as labels so that selectors can be written to reference them

innovate-invent:
What values did you encounter that were not dns compliant? Most of these need to remain labels to allow for selection.

almahmoud:
@innovate-invent `Convert characters1` was the invalid value. See https://github.com/galaxyproject/galaxy/issues/10469. I think it's `ajs.job_wrapper.tool.old_id` but have not verified it

innovate-invent:
Ah, the error has the solution. 
```py
re.sub("[^-A-Za-z0-9_.]", "_", value)
```

almahmoud:
I suppose we could just run all the values through that. I am fairly sure  `ajs.job_wrapper.tool.old_id` is the culprit in this case

almahmoud:
I did version as well for the potential `+`. I'm not sure if `job_id` needs it but did it just in case for the temporary patch and so far so good: https://github.com/almahmoud/galaxy/commit/f4d1a0f041ee2d38b61aaa06bcbfa9ec5cc66b68

lib/galaxy/jobs/runners/kubernetes.py
nuwang: Were there any particular use cases you had in mind?

I've implemented the regex solution for now, just to get past the immediate roadblock, but it's probably worth rationalizing whether each non-standard item should be a label or an annotation.

innovate-invent: The rational is that if you have 20K jobs running and you need to debug one of them, to provide a means of filtering on useful properties.

innovate-invent (approved):

lib/galaxy/jobs/runners/kubernetes.py
nuwang: This info is also queryable from the database. Creating a secondary queryable database in Kubernetes does open up some new possibilities,  but it does have a small cost and I think that motivating it with concrete usecases would be helpful in determining what to include.  E.g. would it be useful to add the user id in there? Do we need to filter by workflow etc.

innovate-invent: This information is important for cluster statistics and compute quota enforcement. De-duplicating this information is a bit of an over optimization.

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -205,19 +207,19 @@ def __get_k8s_job_spec_template(self, ajs):
...
+                    "app.galaxyproject.org/job_id": self.LABEL_REGEX.sub("_", ajs.job_wrapper.get_id_tag()),
+                    "app.galaxyproject.org/handler": self.LABEL_REGEX.sub("_", self.app.config.server_name),
+                    "app.galaxyproject.org/destination": self.LABEL_REGEX.sub(
+                        "_", str(ajs.job_wrapper.job_destination.id))
```
nuwang, line 219: @mvdbeek Something unexpected just happened - the following error came up, presumably because it was necessary to cast job_destination.id to str.
```
galaxy.jobs.runners ERROR 2020-10-22 01:05:46,092 [p:15,w:1,m:0] [KubernetesRunner.work_thread-1] (3) Unhandled exception calling queue_job
File "/galaxy/server/lib/galaxy/jobs/runners/__init__.py", line 137, in run_next
method(arg)
File "/galaxy/server/lib/galaxy/jobs/runners/kubernetes.py", line 131, in queue_job
self.__get_k8s_job_spec(ajs)
File "/galaxy/server/lib/galaxy/jobs/runners/kubernetes.py", line 196, in __get_k8s_job_spec
k8s_job_spec = {"template": self.__get_k8s_job_spec_template(ajs),
File "/galaxy/server/lib/galaxy/jobs/runners/kubernetes.py", line 218, in __get_k8s_job_spec_template
"app.galaxyproject.org/destination": self.LABEL_REGEX.sub("_", ajs.job_wrapper.job_destination.id)
TypeError: expected string or bytes-like object
galaxy.jobs.runners.kubernetes ERROR 2020-10-22 01:05:46,093 [p:15,w:1,m:0] [KubernetesRunner.work_thread-1] (3/None) User killed running job, but error encountered during termination: job name must not be empty
File "/galaxy/server/lib/galaxy/jobs/runners/kubernetes.py", line 515, in stop_job
File "/galaxy/server/lib/galaxy/jobs/runners/util/pykube_util.py", line 63, in find_job_object_by_name
raise ValueError("job name must not be empty")
ValueError: job name must not be empty
```
The unexpected thing is that the tests passed.

innovate-invent: perhaps job_destination.id is not what was intended. I wanted the destination id as specified in the job_conf. That should always be a string.

mvdbeek: Are you using a dynamic destination that maybe doesn't have an id ? If you can reproduce this, can you check if the destination id was `None` or an integer ? I think that would explain why the tests pass. I think you may want to cast tool.version to string. Yaml tools can specify integers or floats as version. That they are strings is an artefact of parsing xml (and if you use a job config yaml file with integer destination ids that might be another explanation why the test didn't fail).

nuwang: Thanks, that makes sense. Will update PR.

mvdbeek:
Thanks everyone!

PR #10471, opened 2020-10-20
Document workflow classes

innovate-invent:
I have been complaining for years for module and class level docs in Galaxy. This is my nudge in the right direction.

Relevant to @mvdbeek and @dannon 

Please expand on the subworkflow functionality and how references are made between the different objects outside of the orm model.


dannon:
Thanks for adding this, it's a good start and we should continue to expand on these docs as mentioned.

Issue #9226, opened 2020-01-16
Subworkflow output starred as parent workflow output does not work


innovate-invent:
This might be related to #7959 but it appears that subworflow outputs are not included in the containing workflow outputs when 'starred' in the parent.
`GET /api/workflows/{workflow_id}/invocations/{invocation_id}` should include the output in its `outputs` property.

jmchilton:
I wrote a test case and it passes, so I think has been fixed. https://github.com/galaxyproject/galaxy/pull/10478 adds the test case to ensure no regressions. Although I couldn't pinpoint a specific commit that I think would have fixed it, I know we've taken several passes at workflow state stuff around outputs and I can imagine it being a lot better than it was with whatever stable release was being used when the issue was open.

innovate-invent:
I think this was on the first release with subworkflows

jmchilton:
Well lets close this then, if it hits again let me know and we can go from there. Thanks for the report.

PR #10478, opened 2020-10-20
Add workflow test for assigning subworkflow outputs as outputs.

jmchilton:
Test case for regression of https://github.com/galaxyproject/galaxy/issues/9226.


jmchilton:
@guerler - here is the requested test I think.

guerler:
@jmchilton looked more into it and it seems that we were talking about two different issues. The editor does register workflow outputs for subworkflows, that they are missing further downstream probably indicates a backend issue. Regarding the output options, the backend does not support pjas for subworkflows yet.

PR #10483, opened 2020-10-21
hide purged datasets from visualizations

### LINKED ISSUE #10278: Purged datasets listed as available for use in visualization drop-down selection
Screen Shot 2020-09-23 at 4 05 36 PM

All datasets that have the appropriate datatype are listed in the available datasets for the visualization drop-down selectors, whether or not they have been hidden or purged

OlegZharkov:
fixes https://github.com/galaxyproject/galaxy/issues/10278

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/plugins.py
+++ b/lib/galaxy/webapps/galaxy/api/plugins.py
@@ -46,7 +46,7 @@ class PluginsController(BaseAPIController):
             result = {}
             result["hdas"] = []
             for hda in history.datasets:
-                if registry.get_visualization(trans, id, hda) and not hda.purged and not hda.deleted:
+                if not hda.deleted and registry.get_visualization(trans, id, hda):
                     result["hdas"].append({
                         "id": trans.security.encode_id(hda.id),
                         "name": hda.name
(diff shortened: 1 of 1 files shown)


lib/galaxy/webapps/galaxy/api/plugins.py
```diff
@@ -46,7 +46,7 @@ def show(self, trans, id, **kwargs):
             result = {}
             result["hdas"] = []
             for hda in history.datasets:
-                if registry.get_visualization(trans, id, hda):
+                if registry.get_visualization(trans, id, hda) and not hda.purged and not hda.deleted:
```
mvdbeek:
```suggestion
                if not hda.deleted and registry.get_visualization(trans, id, hda):
```
Don't think you can have purged but not deleted datasets, but in either case it's cheaper to check if they're deleted first.

mvdbeek: Ah, and we don't want hidden datasets in there either. So maybe iterate over `history.__dataset_contents_iter()` which will exclude deleted and hidden datasets.

OlegZharkov:
thanks @jmchilton @mvdbeek @astrovsky01

Issue #10192, opened 2020-09-06
build-in default view for JSON files


bgruening:
It would be nice if we can offer a pretty JSON view, similar to what we do with tabular files.
Maybe we can use https://vuejsexamples.com/component-for-rendering-a-tree-view-of-json/ or something similar.

OlegZharkov:
> Maybe we can use https://vuejsexamples.com/component-for-rendering-a-tree-view-of-json/ or something similar.

Oh, they even have a dark mode :heart:

hechth:
+1 on this if it isn't already done

PR #10485, opened 2020-10-21
Render previews of notebooks

mvdbeek:
Screenshot 2020-10-21 at 13 21 48

Heavily inspired by https://github.com/galaxyproject/SARS-CoV-2/blob/03a07fd6f519e91806084847eec1b9b159095b8d/.vuepress/components/JupyterNotebook.vue, which we could probably mount directly instead of doing it like this. I think we'd probably also want to defer loading of the additional dependencies. 

If we can figure that out we could also replace the server-side pysam use with https://github.com/GMOD/bam-js (and many more things).


jmchilton:
> On Node.js, notebook.js runs all "text/html"-type outputs through DOMPurify. The browser-based version, however, does not ship with this library; to enable the default behavior, you must <script>-include or require it before initializing notebook.js.

I'm sure it is on your roadmap - just want to make sure it is sanitized before merged.

mvdbeek:
We're hacking around this with `import "!!script-loader!notebookjs";` right now since I couldn't figure out how to do regular imports there, but I think the upside is that the sanitizer should be loaded. Will definitely check though.

jmchilton:
Awesome - amazing and super cool then! What a great approach.

templates/webapps/galaxy/dataset/tabular_chunked.mako
```diff
@@ -9,6 +9,7 @@
     ${parent.javascript_app()}
 
     <script type="text/javascript">
+    %if dataset.datatype.file_ext != 'ipynb':
```
jmchilton, line 12: Totally fine with the hack - if you wanted to build a framework it'd maybe be worth checking subclasses with https://github.com/galaxyproject/galaxy/pull/9955/files and we could probably do json pretty easily then (https://github.com/galaxyproject/galaxy/issues/10192). 

Edit - oh wait, we're in Python here. I guess you wouldn't need that client-side logic. Nevermind, just waisting your time with pointless comments on this PR 😆.

mvdbeek: Well, I do think we should try to avoid mako here and just redirect to a client-side route ... that's my high level view, but right now I have no idea how to do that. @dannon mentioned he might be up for helping with that.

mvdbeek: We could maybe do this entirely on the client-side, skipping the `/datasets/<dataset>/id?preview=true` thing (if we know we've got a datatype that doesn't do anything special on the backend).

dannon:
I'm trying to dig up commentary, but when I proposed packaging https://github.com/galaxyproject/SARS-CoV-2/pull/87 for Galaxy, there was other pushback.

Issue #10486, opened 2020-10-21
Fix maximum_workflow_jobs_per_scheduling_iteration


mvdbeek:
```
Nicola Soranzo @nsoranzo Oct 16 21:11
@mvdbeek Any reason maximum_workflow_jobs_per_scheduling_iteration is not set by default to a big enough number (instead of the present -1) ?
By looking on how it's passed as max_num_jobs to https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/execute.py#L29 , I don't see how it could work: jobs_executed seems to be set to 0 and never updated.
```

nsoranzo:
xref. https://github.com/galaxyproject/galaxy/issues/10442#issuecomment-712459518

mvdbeek:
Done in https://github.com/galaxyproject/galaxy/pull/10490

Issue #10442, opened 2020-10-16
Deleting an history should cancel running workflow invocations


nsoranzo:
This is especially an issue if the inputs for the workflow are from another non-deleted history, meaning that scheduled jobs won't even be paused.

jmchilton:
We do cancel new scheduling iterations because of deleted histories (https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/workflow/run.py#L170). So this must be within a scheduling iteration I assume. I'm nervous about simply rechecking the history between each step, between each job, etc... but clearly scheduling iterations are too long right now if this is a problem. I assume the jobs don't run at least?

nsoranzo:
> We do cancel new scheduling iterations because of deleted histories ([https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/workflow/run.py#L170](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/workflow/run.py?rgh-link-date=2020-10-19T20%3A17%3A37Z#L170)). So this must be within a scheduling iteration I assume.

Yes, the user deleted the history while a very large mapping step was being scheduled.

> I'm nervous about simply rechecking the history between each step, between each job, etc... but clearly scheduling iterations are too long right now if this is a problem. I assume the jobs don't run at least?

I didn't check at the time if the jobs were run (I have separate workflow and job handlers), and now this is too buried in the logs for me to be confident in the answer, unfortunately. But, as you mention, the unnecessary workflow scheduling of such large steps is any way a problem for us.

A possible set of solutions for this could be:
- from `galaxy.managers.histories.HistoryManager.delete()` call `galaxy.managers.workflows.WorkflowsManager.cancel_invocation()` on each invocation running on the history to delete
- change the default of `maximum_workflow_jobs_per_scheduling_iteration` from -1 to a large but positive number.

The use of `maximum_workflow_jobs_per_scheduling_iteration` was suggested by @mvdbeek , but changing its default is my idea (so don't blame him if the idea is stupid :D ).
As mentioned on Gitter, I am not sure the current implementation of `maximum_workflow_jobs_per_scheduling_iteration` is working as expected though. By looking on how it's passed as `max_num_jobs` to https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/execute.py#L29 , `jobs_executed` seems to be set to 0 and never updated (could be partially due to https://github.com/galaxyproject/galaxy/pull/7449 ).

[...]

PR #10490, opened 2020-10-21
[20.09] Fix ``maximum_workflow_jobs_per_scheduling_iteration`` and set 1000 as default value

mvdbeek:
Fixes https://github.com/galaxyproject/galaxy/issues/10486 and should address the bulk of https://github.com/galaxyproject/galaxy/issues/10442 ?

### DIFF
--- a/doc/source/admin/galaxy_options.rst
+++ b/doc/source/admin/galaxy_options.rst
@@ -3181,8 +3181,8 @@
     associated with scheduling workflows at the expense of increased
     total DB traffic because model objects are expunged from the SQL
     alchemy session between workflow invocation scheduling iterations.
-    Set to -1 to disable any such maximum (the default).
-:Default: ``-1``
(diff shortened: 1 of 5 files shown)


jmchilton:
I love the fixes, thank you! The default change this late makes me a little nervous. There are many workflows that would not have been rescheduled in the middle that will be now. It will cause many workflows to be evaluated in very different ways. There is a high probability this will cause cause us to find important bugs. I wouldn't change that default this close to the release but I'm happy to help fix the bugs we find. I'm a +1.

jmchilton (approved):

nsoranzo (approved):

mvdbeek:
I did some manual tests, I think this might not be working correctly. Don't merge it yet.

mvdbeek:
never mind, it does work correctly, but maybe I should have tested this with less than 2000 jobs to create on my laptop 😅

PR #10492, opened 2020-10-21
Add update time to workflow version selector in workflow editor

### LINKED ISSUE #10491: Show workflow date in workflow editor version list

guerler:
image

Additionally this PR displays the workflow version starting with `1` instead of `0`.

Resolves #10491.

### DIFF
--- a/client/src/components/Workflow/Editor/Attributes.vue
+++ b/client/src/components/Workflow/Editor/Attributes.vue
@@ -104,7 +104,7 @@ export default {
             for (let i = 0; i < this.versions.length; i++) {
                 const current_wf = this.versions[i];
                 const update_time = moment.utc(current_wf.update_time).format("MMM Do YYYY");
-                const label = `${current_wf.version}: ${update_time}, ${current_wf.steps} steps`;
+                const label = `${current_wf.version + 1}: ${update_time}, ${current_wf.steps} steps`;
                 versions.push({
                     version: i,
                     label: label,
--- a/client/src/components/Workflow/Editor/Attributes.vue
+++ b/client/src/components/Workflow/Editor/Attributes.vue
@@ -43,6 +43,7 @@
 <script>
 import Vue from "vue";
 import BootstrapVue from "bootstrap-vue";
+import moment from "moment";
 import { Services } from "components/Workflow/services";
 import Tags from "components/Common/Tags";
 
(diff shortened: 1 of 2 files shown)


mvdbeek:
This is great and nicely implies that these versions are internal!

client/src/components/Workflow/Editor/Attributes.vue
```diff
@@ -102,10 +103,8 @@ export default {
             const versions = [];
             for (let i = 0; i < this.versions.length; i++) {
                 const current_wf = this.versions[i];
-                let label = `Version ${current_wf.version}, ${current_wf.steps} steps`;
-                if (i == this.version) {
-                    label = `${label} (active)`;
-                }
+                const update_time = moment.utc(current_wf.update_time).format("MMM Do YYYY");
+                const label = `${current_wf.version + 1}: ${update_time}, ${current_wf.steps} steps`;
```
mvdbeek, line 107: I don't know how I feel about that. How about dropping the integer version altogether?
I don't think we should differ from the internal representation.

guerler: I think the versions should remain enumerated in the select field label. In this case the number in the labels indicate the 1st version, 2nd version and so on, that seems to be more intuitive for users working with their workflows instead of referring to a e.g. version 0. Since this change just affects the label I do not see a problem with it but am open for suggestions.

mvdbeek: My suggestion is to not do that. The version is a query parameter in a few endpoints, how would anyone know that you've added +1 here ?
We also don't add +1 to other versioned things, this feels quite wrong.
 These versions are checkpoints in a sense, it was my mistake to call these versions, that's why I think update time is fine.
If you want to really change this you can change the parsing and serialization on the backend, so that the version numbers are always consistent.

dannon: For a little clarity, this numerical version isn't actually stored anywhere, right, it's just constructed based on the sorted StoredWorkflow->Workflow mapping?  It doesn't seem like this *should* be used in APIs anywhere.

mvdbeek: It is a constructed version. The "right" way fetch specific Workflow objects was added later by adding `style="instance"`. I agree that this is better for addressing specific versions of a workflow. Still, the version is included in API responses, I am doubtful that starting with 1 instead of 0 outweighs the confusion. But hopefully we will have explicit versions in the next release, making this less of an issue ?

PR #10493, opened 2020-10-21
Display minutes in admin jobs cutoff 

### LINKED ISSUE #10474: Change the unit of time when displaying jobs in admin
Currently, under Admin -> Jobs, the cutoff time for displaying current/recent jobs is in seconds, defaulting to 180. Minutes seem like a more applicable unit here given that even the default is set to multiple minutes, not a handful of seconds...

image

OlegZharkov:
fixes https://github.com/galaxyproject/galaxy/issues/10474

### DIFF
--- a/client/src/components/admin/Jobs.vue
+++ b/client/src/components/admin/Jobs.vue
@@ -37,12 +37,11 @@
                             <b-form-group
                                 id="cutoff"
                                 label="Update Jobs"
-                                label-for="cutoff-seconds"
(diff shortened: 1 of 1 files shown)


mvdbeek:
Thanks @OlegZharkov!

OlegZharkov:
Thanks @mvdbeek and @afgane! 
Not sure if we should paginate it, I guess it depends how often admins use that page

Issue #9778, opened 2020-05-17
Improve Galaxy XSD for better IDE autocompletion/help/validation support


davelopez:
Hi!
I think it would be great to have real-time auto-completion, help and validation during XML tool wrappers development (in addition to the good old planemo linter). This can be very useful for new tool developers to ease the learning curve as they can discover the API, read the documentation for best practices and avoid syntax and structural errors as they type.
Many IDEs like IntelliJ, VSCode and others can provide this functionality, but, right now, it seems there is no full support for some valid elements like macros (i.e. the `expand` element as reported in #3029) and many elements are missing documentation or more restrictive type definitions.

I am no expert in XSD, but I would like to contribute to improving it if I can :)

So far, the first difficulty I found is about how to support multiple `expand` elements inside the `tool` element.
Right now, the current `tool` element definition uses the [`<xs:all>`](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tool_util/xsd/galaxy.xsd#L45) indicator to define its children in any given order. Unfortunately, this indicator does not allow more than 1 occurrence of any of its child elements, so the `expand` element will be allowed just once.
After some research, looks like there are some ways to overcome this problem:

1. Replace `<xs:all>` with `<xs:choice maxOccurs="unbounded">`: this will allow having multiple `expand` elements but at the cost of also allowing multiple occurrences for the rest of the elements which I do not think is acceptable.

2. Replace the `<xs:all>` with `<xs:sequence>`: this will allow specifying the `minOccurs` and `maxOccurs` for each child element without the `maxOccurs="1"` limitation at the cost of having to define the child elements in the specific order, which probably would make sense since the [coding style best practices](https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#coding-style) already define this order.

I would probably go for option 2, but since I am fairly new to Galaxy I am unaware of the consequences of replacing this `<xs:all>` indicator and if this may break some other tools relying on the current XSD like planemo, the autogenerated documentation, etc.

Thanks in advance for your inputs!

jmchilton:
[...]

PR #10495, opened 2020-10-21
Add missing documentation for macro elements in XSD

### LINKED ISSUE #3029: Macros are not documented in galaxy.xsd
Macros (`<macros>`, `<import>`, `<token>`, `<xml>`, `<yield/>`, `<expand>`) are not documented in `lib/galaxy/tools/xsd/galaxy.xsd` .

davelopez:
Added some documentation to ``import``, ``token`` and ``xml`` elements defined in the XSD.
The ``yield`` and ``expand`` macro-related elements are not modelled by the schema definition as explained [here](https://github.com/galaxyproject/galaxy/issues/9778#issuecomment-630167658).

Also added a simple type called ``MacroImportType`` to validate the value of ``import`` is an actual XML file.

This PR closes #3029


lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -5848,4 +5905,10 @@ is the only supported options.</xs:documentation>
     </xs:restriction>
   </xs:simpleType>
 
+  <xs:simpleType name="MacroImportType">
+    <xs:restriction base="xs:anyURI">
+      <xs:pattern value="([a-zA-Z0-9_\-/\.])+.xml"/>
```
davelopez: I am not sure if this regex will cover any possible use case or if it will break validation of existing tools.

davelopez: Actually I think this is better...
```suggestion
      <xs:pattern value="[a-zA-Z0-9_\-\.]+.xml"/>
```

davelopez: I am assuming the macro files will be in the same directory as the tool importing them. Is that correct?

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -5848,4 +5905,10 @@ is the only supported options.</xs:documentation>
     </xs:restriction>
   </xs:simpleType>
 
+  <xs:simpleType name="MacroImportType">
+    <xs:restriction base="xs:anyURI">
```
davelopez: This shouldn't be a URI, just a string with the pattern check will be enough.
```suggestion
    <xs:restriction base="xs:string">
```

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -5848,4 +5905,10 @@ is the only supported options.</xs:documentation>
     </xs:restriction>
   </xs:simpleType>
 
+  <xs:simpleType name="MacroImportType">
+    <xs:restriction base="xs:string">
+      <xs:pattern value="[a-zA-Z0-9_\-\.]+.xml"/>
```
mvdbeek, line 5917: I've checked the tools-iuc tools, and this would be fine. I think it wouldn't be a big deal if ultimately it causes a small fraction of tools to fail linting. Best practice should be that the macro file is next to the tool, without space in the path, and for eventual plans of distributing tools differently from what we do today this is a big plus.

mvdbeek (approved):
This is awesome, thanks a lot @davelopez!

lib/galaxy/tool_util/xsd/galaxy.xsd
jmchilton: I'm uneasy with this, it doesn't literally represent the spec which even if we require those files sit next to that file, we don't currently require they have a .xml extension. I guess it isn't a huge deal either way though.

davelopez: Many thanks for the review @mvdbeek and @jmchilton!!

I added the ``.xml`` extension restriction because the [vscode extension for the language server relays on the file extension](https://github.com/davelopez/galaxy-language-server/blob/master/client/package.json#L29) to activate itself. Of course, this can be any other file extension, but it felt natural to use .xml also for the macro files.
Anyway, we can fall back to a simple ``xs:string`` without restrictions or adjust the pattern to better fit the spec.

jmchilton:
Thanks so much!

Issue #10480, opened 2020-10-21
Jobs appear as perpetually running with k8s runner when pods are unschedulable


almahmoud:
When a job is dispatched by the K8S runner but all its pods are unschedulable (eg: no node satisfying mem/cpu requests or satisfying node selector), the job never runs but never errors out and appears as perpetually running in Galaxy.
I think jobs should probably appear as running only after at least one of the pods in the job has started rather than when the job is created, and probably add a timeout for unschedulable jobs to wait for a certain time (eg: in case autoscaler is kicking in) but then error out with a message informing the user that the job requirements could not be satisfied and to contact administrator

almahmoud:
Took this on this morning btw

PR #10496, opened 2020-10-21
Handle unschedulable jobs in k8s runner

almahmoud:
Addresses https://github.com/galaxyproject/galaxy/issues/10480

- [x] Adds `k8s_unschedulable_walltime_limit` which specifies the time limit for a job to wait while pods are unschedulable.
- [x] Makes job not be marked as running in Galaxy if Job is dispatched but job is still unschedulable
- [x] Handles marking the job as failed with a reasonable message and cleaning up the job when unschedulable past the specified limit

Testing everything together:
- [x] Job is not marked as running if the pod is unschedulable
- [x] Specified time limit is respected (tested with 30 seconds, defaulted to 30 mins)
- [x] Job error is properly reported
image
- [x] Job resource is cleaned up in k8s
- [x] Schedulable jobs still run unaffected

### DIFF
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -2,12 +2,11 @@
 Offload jobs to a Kubernetes cluster.
 """
 
+from datetime import datetime
 import logging
 import math
 import os
 import re
-
-from datetime import datetime
 
 import yaml
 
@@ -586,7 +585,7 @@ class KubernetesJobRunner(AsynchronousJobRunner):
             return True
 
         return False
-    
+
     def __job_pending_due_to_unschedulable_pod(self, job_state):
         """
         checks the state of the pod to see if it is unschedulable.
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -597,7 +597,7 @@ class KubernetesJobRunner(AsynchronousJobRunner):
             return False
 
         pod = Pod(self._pykube_api, pods.response['items'][0])
-        is_unschedulable = bool(len([c in pod.obj['status']['conditions'] if c.get("reason") == "Unschedulable"]))
+        is_unschedulable = bool(len([c for c in pod.obj['status']['conditions'] if c.get("reason") == "Unschedulable"]))
         if pod.obj['status']['phase'] == "Pending" and is_unschedulable:
             return True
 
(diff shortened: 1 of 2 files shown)


lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -461,6 +470,20 @@ def check_watched_item(self, job_state):
             # job is no longer viable - remove from watched jobs
             return None
 
+    def _handle_unschedulable_job(self, job, job_state):
```
jmchilton, line 528: Can you get in the habit of putting these things into pykube_util. I know you're not using the Pulsar Kubernetes runner yet, but I think long term it is will be a valuable tool and we should work to keep feature parity - or at least making it easy to have feature parity. I'm going to have to refactor this out of here for reuse in that code and there is not test coverage here so I risk breaking things - if it were developed to be used that way from the start it would be less likely to be broken when I refactor it.

Update: I guess this comment is at the wrong place - I meant it to be around the calculation of is_unschedulable.

almahmoud: Done in https://github.com/galaxyproject/galaxy/pull/10496/commits/dc7d3fdf4226b44b8b929e419a2aa88de5ce25f7#diff-83b685d37c2473c87101dc9de4da0464b79bb2042539132e5c41ee7261e93b98R71

jmchilton (approved):

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -515,6 +525,20 @@ def check_watched_item(self, job_state):
...
+        # Handle unschedulable job that exceeded deadline
+        with open(job_state.error_file, 'a') as error_file:
```
nuwang: I think it would be good to guard against the condition where this file doesn't exist, because I've seen several instances (in other places in the runner) where writing to the error file fails.

mvdbeek: What do we need this file for ? Can we eliminate this ? This was used in older traditional HPC runners that would kill a job and then write something to a specific file, but kubernetes doesn't do that, and we can write to `job_state.fail_message`. I would assume that ends up on the job_stderr column of the job table.

almahmoud: I was just following what is done to handle failures in other places eg https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/jobs/runners/kubernetes.py#L521 . Can we just change this as part of moving to the pulsar runner? Not really worth the time sink now? I can just remove the file and do fail message

mvdbeek:
Thanks @almahmoud!

PR #10497, opened 2020-10-21
Support stopping Interactive Tools without deleting their outputs

natefoo:
This is done by adding new `STOP` and `STOPPED` job states that work similar to (and use the same job stop queue in the job handler) `DELETED_NEW` and `DELETED`. Unlike deleted jobs, stopped jobs are expected to run through the normal job wrapper finish method after the job is killed.

Caveats:
- Only possible from the Active GxIT list currently (idea: for HDAs, can we make the delete "X" a stop sign on GxIT outputs, and once the user clicks that and the job is stopped, it becomes the regular "X"?)
- External metadata will never run (~~no idea what to do about this~~ added an option to retry *just GxIT* set_meta internally, which isn't perfect, but it's better than nothing).
- ~~Local and DRMAA runners only (so far - will do the others before this PR gets out of draft)~~ done except Chronos and probably Pulsar MQ.


lib/galaxy/jobs/handler.py
```diff
@@ -200,7 +200,8 @@ def __check_jobs_at_startup(self):
         jobs_at_startup = []
         if self.track_jobs_in_database:
             in_list = (model.Job.states.QUEUED,
-                       model.Job.states.RUNNING)
+                       model.Job.states.RUNNING,
+                       model.Job.states.STOPPED)
```
nsoranzo, line 204: Need to add 'stopped' to the method docstring?

lib/galaxy/jobs/runners/cli.py
```diff
@@ -168,12 +168,13 @@ def check_watched_items(self):
                 state = job_interface.parse_single_status(cmd_out.stdout, external_job_id)
                 if not state == model.Job.states.OK:
                     log.warning(f'({id_tag}/{external_job_id}) job not found in batch state check, but found in individual state check')
+            job_state = ajs.job_wrapper.get_state()
```
nsoranzo, line 171: You can move this to line 160 and re-use it on line 161.

lib/galaxy/model/__init__.py
```diff
@@ -687,7 +687,9 @@ class Job(JobLike, UsesCreateAndUpdateTime, Dictifiable, RepresentById):
                    FAILED='failed',
                    PAUSED='paused',
                    DELETED='deleted',
-                   DELETED_NEW='deleted_new')
+                   DELETED_NEW='deleted_new',
+                   STOP='stop',
```
nsoranzo: As mentioned on Gitter, STOPPING seems to me more appropriate for a job state, even if it is a request to stop the job.

lib/galaxy/jobs/runners/cli.py
natefoo: This was actually intentional as the single status check in the intervening lines could take a while, so I was worried the status might be stale by here if it was fetched earlier.

lib/galaxy/model/__init__.py
natefoo: Will do. Thanks for the review!

mvdbeek:
Can I convince you to add a test to https://github.com/guerler/galaxy/blob/8d444ff2e5ba139294621482ac2ff5368c7ae905/test/integration/test_interactivetools_api.py#L99 ?

lib/galaxy/jobs/__init__.py
```diff
@@ -1536,6 +1536,8 @@ def _finish_dataset(self, output_name, dataset, job, context, final_job_state, r
             # it would be quicker to just copy the metadata from the originating output dataset,
             # but somewhat trickier (need to recurse up the copied_from tree), for now we'll call set_meta()
             retry_internally = util.asbool(self.get_destination_configuration("retry_metadata_internally", True))
+            if self.tool.tool_type == 'interactive':
```
jmchilton: ``not retry_internally and self.tool.tool_type == 'interactive'``

mvdbeek (approved):
@natefoo can you check the changes I've made ? Especially the logic with ignoring the pulsar returncode.

natefoo:
@mvdbeek seems ok, it's not needed for other runners as well?

mvdbeek:
It's not needed for the local runner, and I don't think we pass the exit code in the same way in other runners (but we don't have tests for those, so 🤷). Seems like the typical DRM would set the state to completed when you stop a job, so that shouldn't count as an error (tested that real quick with torque).

natefoo:
Thanks @mvdbeek !

PR #10310, opened 2020-09-28
Release notes for 20.09

mvdbeek:

PR #10498, opened 2020-10-21
Add guidelines for proper commit messages

jdavcs:
Ref: https://github.com/galaxyproject/galaxy/pull/10310#issuecomment-708407696

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,32 +89,17 @@ https://help.github.com/en/github/getting-started-with-github/git-and-github-lea
    the relevant tests as part of the pull request review process.
 
 6. Write a useful and properly formatted commit message. 
-   Follow [these guidelines and template](https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines):
+   Follow [these guidelines and template](https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines),
+   in particular start your message with a short imperative sentence on a single
+   line, possibly followed by a blank line and a more detailed explanation.
 
-     ```
-     Capitalized, short (50 chars or less) summary (no period at the end)
-     
-     More detailed explanatory text, if necessary.  Wrap it to about 72
-     characters or so.  In some contexts, the first line is treated as the
-     subject of an email and the rest of the text as the body.  The blank
-     line separating the summary from the body is critical (unless you omit
-     the body entirely); tools like rebase will confuse you if you run the
-     two together.
-     
-     Write your commit message in the imperative: "Fix bug" and not "Fixed bug"
-     or "Fixes bug."  This convention matches up with commit messages generated
-     by commands like git merge and git revert.
-     
-     Further paragraphs come after blank lines.
-     
-     - Bullet points are okay, too
-     
-     - Typically a hyphen or asterisk is used for the bullet, followed by a
-       single space, with blank lines in between, but conventions vary here
-     ```
-   (Adapted from https://git-scm.com/book/ch5-2.html#_commit_guidelines. Also, check
-   [Tim Pope's post](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) 
-   for more details).
+   In the detailed explanation it's good to include relevant external references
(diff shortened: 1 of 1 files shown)


martenson (approved):

CONTRIBUTING.md
```diff
@@ -88,10 +88,38 @@ https://help.github.com/en/github/getting-started-with-github/git-and-github-lea
...
+     or "Fixes bug."  This convention matches up with commit messages generated
+     by commands like git merge and git revert.
+     
+     Further paragraphs come after blank lines.
+     
+     - Bullet points are okay, too
+     
+     - Typically a hyphen or asterisk is used for the bullet, followed by a
+       single space, with blank lines in between, but conventions vary here
+     ```
```
nsoranzo: Isn't linking to the above guidelines enough?

jdavcs: Maybe. I excluded the last point about a hanging indent - I think we don't need to require it. However, I wanted to quote the essential: the link points to a text that is *much more detailed*, offers context, explanation, points to further resources. I think providing a summary of the absolute essentials with a link pointing to more may be more useful than just a link. Otherwise, I fear we'll end up restating these essentials over and over in code reviews :-)

nsoranzo: I think it would be more effective to just point to the docs and add instead an example of a real commit message which includes for example a traceback and link to the issue fixed.

`In some contexts, the first line is treated as the
     subject of an email and the rest of the text as the body`

This is irrelevant for us.

> Otherwise, I fear we'll end up restating these essentials over and over in code reviews :-)

I think we mainly review PR titles, which end up in the release notes, but bad commit messages makes it difficult to retrospectively understand the reason for a change, so they are obviously also important.

jdavcs: > I think it would be more effective to just point to the docs

But there are no *docs* per se - it's a common sense format recommended by the authors of the Git Book (and other sources). Which is why I think it's OK for us to have our own, tailored version.

> ... and add instead an example of a real commit message...

I think we could do that. Although crafting a commit message that is generic enough to serve as an example for diverse cases, but still specific enough to be useful might require considerable effort. Maybe I'm wrong.

> `In some contexts, the first line is treated as the subject of an email and the rest of the text as the body`
> 
> This is irrelevant for us.

Agreed! We can drop that line; we quote this as "adapted from...", so we can edit it.

> I think we mainly review PR titles, which end up in the release notes, but bad commit messages makes it difficult to retrospectively understand the reason for a change, so they are obviously also important.

100% agree with that - it's my primary motivation here.

nsoranzo: > > ... and add instead an example of a real commit message...
> 
> I think we could do that. Although crafting a commit message that is generic enough to serve as an example for diverse cases, but still specific enough to be useful might require considerable effort. Maybe I'm wrong.

Something like https://github.com/galaxyproject/galaxy/commit/0429c4d515536f9cca6b70b2abeb019de807c955 ? Not that I wanted one of mine to be used, but it was easier to search in stuff I did.

jdavcs: Yes, I think that's a good example. Let's use it. Do you mind making the change?

Also, I still prefer we keep the essentials as opposed to just having a link. We could trim them to the bare minimum if what we have now is too long (I think we could have that + your example, but I'm OK either way). If we were to trim the text, I'd opt  for leaving the fist line ("Capitalized, short..."), a line from the next paragraph ("More detailed explanatory text, if necessary.  Wrap it to about 72 characters or so."), and I would definitely mention imperative mode. I feel those three items are the most significant points that apply to any commit, which is why I'd like to see them explicitly stated. But I'm OK if you really prefer the link + example with no further text.

jdavcs: @nsoranzo would you like me to make this change? I think it would be good to have this in before tomorrow (paper cuts) - might be helpful to new contributors.

nsoranzo: Sorry, this fell through the cracks! I'll try to do that now.

jdavcs: Great! Thank you - and no worries!

nsoranzo (approved):

ic4f:
@nsoranzo this looks perfect, I think. Thank you! Hopefully this will help improve our commit messages.

Issue #8505, opened 2019-08-23
Rule based uploader doesn't refresh file list on load


hexylena:
![image](https://user-images.githubusercontent.com/458683/63587562-9ebcea00-c5a4-11e9-9adb-32e4ab5982c1.png)

tried closing/re-opening upload dialog, but only page refresh worked.

PR #10505, opened 2020-10-21
Use dataset selector in rules input dialog.

### LINKED ISSUE #8504: Rule based uploader allows selecting deleted + error datasets
image

jmchilton:
Drops buggy custom dataset selection widget.

Fixes #8504 & #8505 - two paper cuts for the price of one.

### DIFF
--- a/client/src/components/Upload/RulesInput.vue
+++ b/client/src/components/Upload/RulesInput.vue
@@ -22,18 +22,16 @@
                 </div>
             </div>
             <div id="upload-rule-dataset-option" class="upload-rule-option" v-if="selectionType == 'dataset'">
-                <div class="upload-rule-option-title">{{ l("Select dataset to load") }}</div>
-                <div class="dataset-selector">
-                    <select2
-                        v-model="selectedDatasetId"
-                        container-class="upload-footer-selection"
-                        placeholder="Select Dataset"
-                    >
-                        <option></option>
-                        <option v-for="dataset of datasets" :key="dataset.id" :value="dataset.id">
-                            {{ dataset.hid }}: {{ dataset.name }}</option
-                        >
-                    </select2>
+                <div class="upload-rule-option-title">
+                    History dataset
+                </div>
+                <div>
+                    <b-link @click="onSelectDataset" v-if="selectedDatasetName == null">
+                        {{ l("Select") }}
+                    </b-link>
+                    <span v-else>
+                        {{ selectedDatasetName }} <font-awesome-icon icon="edit" @click="onSelectDataset" />
+                    </span>
                 </div>
             </div>
         </span>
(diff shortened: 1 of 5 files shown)


OlegZharkov:
Thanks a lot John. I tried it out and it indeed hides deleted datasets, however I still see hidden datasets in the dialog. Could we also fix that within this PR?

client/src/components/Upload/RulesInput.vue
```diff
@@ -22,18 +22,16 @@
...
-                        <option v-for="dataset of datasets" :key="dataset.id" :value="dataset.id">
-                            {{ dataset.hid }}: {{ dataset.name }}</option
-                        >
-                    </select2>
+                <div class="upload-rule-option-title">
+                    History dataset
+                </div>
+                <div>
+                    <b-link @click="onSelectDataset" v-if="selectedDatasetName == null">
+                        {{ l("Select") }}
```
OlegZharkov, line 30: Should we actually always do that? (maybe enforce even, it's the first time I see localization call in galaxy)

client/src/components/Upload/RulesInput.vue
```diff
@@ -22,18 +22,16 @@
...
-                        container-class="upload-footer-selection"
-                        placeholder="Select Dataset"
-                    >
-                        <option></option>
-                        <option v-for="dataset of datasets" :key="dataset.id" :value="dataset.id">
-                            {{ dataset.hid }}: {{ dataset.name }}</option
-                        >
-                    </select2>
+                <div class="upload-rule-option-title">
+                    History dataset
```
OlegZharkov, line 26:
``` suggestion
                    {{ l("History Dataset") }}
```

Sorry, since we already use  `{{ l("History Dataset") }}` in this file...

client/src/components/Upload/RulesInput.vue
jmchilton: We should always do that I think, I've gotten bad at it.

client/src/components/Upload/RulesInput.vue
jmchilton: But the other titles that appear above it with the same style are not all upper case ("Load tabular data from").

Should we make the select list use "Pasted table", "History dataset", etc.. to make it uniform?

OlegZharkov: upper case? other titles also have same `upload-rule-option-title` class and localisation `{{ l("Load tabular data from") }}` [here](https://github.com/galaxyproject/galaxy/blob/e05e0b786e6cc465a6015fa91c7026b90cea2148/client/src/components/Upload/RulesInput.vue#L14)

> Should we make the select list use "Pasted table", "History dataset", etc.. to make it uniform?

can you elaborate that sentence, please?  :sweat_smile:

jmchilton: I think ``l("History Dataset")`` would clash with these other headers:

Screen Shot 2020-10-27 at 10 24 06 AM

where we don't make each word start with a capital.

mvdbeek (approved):

PR #10506, opened 2020-10-21
Do not let uWSGI remap stdin to /dev/null by default

jdavcs:
Set this by default to allow using debugging tools like pdb. (as per suggestion by @nsoranzo)

From https://uwsgi-docs.readthedocs.io/en/latest/ThingsToKnow.html#things-to-know-best-practices-and-issues-read-it:
> By default, stdin is remapped to /dev/null on uWSGI startup. If you need a valid stdin (for debugging, piping and so on) 
add --honour-stdin.

### DIFF
--- a/doc/source/admin/galaxy_options.rst
+++ b/doc/source/admin/galaxy_options.rst
@@ -2636,6 +2636,19 @@
 :Type: float
 
 
+~~~~~~~~~~~~~~~~~
+``tool_id_boost``
+~~~~~~~~~~~~~~~~~
+
+:Description:
+    Boosts are used to customize this instance's toolbox search. The
+    higher the boost, the more importance the scoring algorithm gives
+    to the given field.  Section refers to the tool group in the tool
+    panel.  Rest of the fields are tool's attributes.
+:Default: ``9.0``
+:Type: float
+
+
 ~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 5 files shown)


ic4f:
data_manager integration test error. Locally this passes. Trying again.

dannon:
My question is why isn't this the default for uwsgi in general, if there is no downside?

If there truly isn't one -- weird choice by the uwsgi folks, but okay.  But my gut says that instead of making this special case the default, it'd make more sense to me to roll this into a 'dev mode' for the server.

ic4f:
@dannon I looked before making this change and didn't find any obvious downsides. In the uwsgi source, it's enabled for running a python script or an interactive python shell. I've been running it in that mode for months. But now that you mention it, making it part of a dev or debug mode *does* sound like a more logical choice. @nsoranzo, do you think that makes more sense?

nsoranzo:
We already have a `debug` option in galaxy.yml that enables a small set of other options, but I don't think that can work to enable options for uWSGI.

ic4f:
@nsoranzo I think we can do that. #10577.
I'll close this PR if the alternative is a better approach (I think it is).
Yet another option is to simply add this to the docs, but I prefer the second approach that enables it by default when Galaxy is in debug mode.

ic4f:
Closing, see alternative in #10577.

Issue #10229, opened 2020-09-15
Collection element datasets not hidden when copied for "listpaired". Elements are hidden for "paired".


jennaj:
Example`from history`: https://usegalaxy.org/u/jen-galaxyproject/h/dna-ngs-test-datasets-variant-tuto
Example `to history`: https://usegalaxy.org/u/jen-galaxyproject/h/test-history-copied-data

ping @jmchilton

jmchilton:
@assuntad23 and I are discussing this and #10230 - I just wanted to verify this is a problem for copying whole histories? These histories weren't produced by copying individual datasets right?

jennaj:
The problem seems to be resolved now in 20.09. This was from copying datasets -- not the entire history but I just checked both and are working Ok.

Datasets added to collections still do lose their tags once added to a collection. That is the other issue. And whether that is on purpose or not -- I don't know. Personally I prefer if the tags propagated if start with a hash "#"

jennaj:
Wrong! Still a problem. How to reproduce:

Start here: https://usegalaxy.org/u/jen-galaxyproject/h/dna-ngs-test-datasets-variant-tuto

Try copying dataset 6 (listpaired collection) into a new history -- the elements are not hidden. https://usegalaxy.org/u/jen-galaxyproject/h/test-copy-listpaired

Go back to the original history and try copying dataset 9 or 15 -- the collection elements are hidden

assuntad23:
This is only a problem with list of pairs created before 20.09. Any collections created before that may continue to have this issue.

mvdbeek:
Are you sure about that ? What fixed it ? If I follow @jennaj's instructions I still see that the HDAs are unhidden.

jennaj:
Well, we thought was fixed .. but the "hide original elements" was a factor in one of the tests. Couldn't reproduce after, but also didn't fully start over from scratch. 

My original test didn't hide the original elements. And if that is reproducible by you, then still open!

assuntad23:
fixed in #10508

PR #10508, opened 2020-10-21
modified list-of-pairs to copy datasets as hidden

assuntad23:
Fix for papercut issue #10229 

It only fixes new collections that are created using the updated collection creator. 

Thanks to much help from @jmchilton , @dannon , and @mvdbeek !

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -4593,6 +4593,7 @@ class DatasetCollectionElement(Dictifiable, RepresentById):
                 )
             else:
                 new_element_object = element_object.copy(flush=flush)
+                new_element_object.visible = False
                 if destination is not None and element_object.hidden_beneath_collection_instance:
                     new_element_object.hidden_beneath_collection_instance = destination
                 # Ideally we would not need to give the following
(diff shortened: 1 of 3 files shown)


mvdbeek (approved):
Nice!

lib/galaxy/model/__init__.py
```diff
@@ -4593,6 +4593,7 @@ def copy_to_collection(self, collection, destination=None, element_destination=N
                 )
             else:
                 new_element_object = element_object.copy(flush=flush)
+                new_element_object.visible = False
```
mvdbeek, line 4596: @assuntad23 this is the complementary fix for existing collections. This sets all member HDAs of a newly copied collection to hidden. I can't think of a situation where they should ever remain visible.

mvdbeek:
@galaxybot test this

lib/galaxy/model/__init__.py
assuntad23: Thanks Marius!

assuntad23 (commented):
I think this is good to go, but we may need someone else to approve?

mvdbeek:
This is fine, thanks @assuntad23!

PR #10509, opened 2020-10-21
Truncate excessively long dataset names in the collection creator.

davebx:
This was reported at a meeting some days ago by @nekrut.

Before and after screenshots:

before

after

### DIFF
--- a/client/src/mvc/collection/list-of-pairs-collection-creator.js
+++ b/client/src/mvc/collection/list-of-pairs-collection-creator.js
@@ -37,11 +37,11 @@ var PairView = Backbone.View.extend(baseMVC.LoggableMixin).extend({
 
     template: _.template(
         [
-            '<span class="forward-dataset-name flex-column"><%- pair.forward.name %></span>',
+            '<span class="forward-dataset-name flex-column truncate"><%- pair.forward.name %></span>',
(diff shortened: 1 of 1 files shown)


dannon:
Looks good!  Do you want to address the button in the middle, too?  It looks like "Pair these datasets" is also displaying outside its container.

assuntad23:
Nice job! My initial thought was to make the modal the same size as the other two collection creator modals. This would show more of the name, so that unique names are more visible. It could easily work in conjunction with what you've done.

davebx:
@assuntad23 I like the way you think, that sounds like the most ideal solution

davebx:
@dannon I'm less concerned about the 'Pair these datasets' span, since I shrank the viewport by a lot to get a decent screenshot. In normal sizes it shouldn't be an issue, but if you have concerns I don't see a problem adding the `truncate` class to that span as well.

dannon:
@davebx Up to you; I just noticed the overflow.

dannon:
@davebx Sorry about the overlap here.  Superseded by #10240

davebx:
@dannon thanks, I was just trying to find the PR that made this irrelevant

assuntad23:
Sorry to steal your thunder, @davebx !

PR #10510, opened 2020-10-21
Add enumeration restriction to detect_errors

### LINKED ISSUE #10500: In Galaxy.xsd, attribute detect_errors allows any string as value
The type of the attribute ``detect_errors`` in the ``tool/command`` element ([here](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tool_util/xsd/galaxy.xsd#L2879)) is defined as ``xs:string`` but the documentation states that there are only three possible values.

This seems like a little paper-cut that can be easily fixed defining a ``xs:simpleType`` with enumeration restriction :)

davelopez:
Extended a bit the documentation and added a ``simpleType`` definition called ``DetectErrorType`` to allow only the following values on ``detect_errors`` attribute:
- default
- exit_code
- aggressive

This PR closes #10500

### DIFF
--- a/lib/galaxy/tool_util/xsd/galaxy.xsd
+++ b/lib/galaxy/tool_util/xsd/galaxy.xsd
@@ -2882,8 +2882,8 @@ Prior to Galaxy release 19.01 the stdio block has only been used for non-legacy 
 The ``detect_errors`` attribute of ``command``, if present, loads a preset of error detection checks (for exit codes and content of stdio to indicate fatal tool errors or fatal out of memory errors). It can be one of:
 
 * ``default``: for non-legacy tools with absent stdio block non-zero exit codes are added. For legacy tools or if a stdio block is present nothing is added.
-* ``exit_code``: adds checks for non zero exit codes (The @jmchilton recommendation). The ``oom_exit_code`` parameter can be used to add an additional out of memory indicating exit code.
-* ``aggressive``: adds checks for non zero exit codes, and checks for ``Exception:``, ``Error:`` in the standard error. Additionally checks for messages in the standard error that indicate an out of memory error (``MemoryError``, ``std::bad_alloc``, ``java.lang.OutOfMemoryError``, ``Out of memory``). (The @bgruening recommendation).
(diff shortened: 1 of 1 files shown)


lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -2876,9 +2876,16 @@ Prior to Galaxy release 19.01 the stdio block has only been used for non-legacy
...
+        <xs:attribute name="detect_errors" type="DetectErrorType">
           <xs:annotation>
-            <xs:documentation>One of ``default``, ``exit_code``, ``aggressive``. See "Error detection" above for details.</xs:documentation>
+            <xs:documentation xml:lang="en"><![CDATA[
+The ``detect_errors`` attribute of ``command``, if present, loads a preset of error detection checks (for exit codes and content of stdio to indicate fatal tool errors or fatal out of memory errors). It can be one of:
+
+* ``default``: for non-legacy tools with absent stdio block non-zero exit codes are added. For legacy tools or if a stdio block is present nothing is added.
+* ``exit_code``: adds checks for non zero exit codes (The @jmchilton recommendation). The ``oom_exit_code`` parameter can be used to add an additional out of memory indicating exit code.
+* ``aggressive``: adds checks for non zero exit codes, and checks for ``Exception:``, ``Error:`` in the standard error. Additionally checks for messages in the standard error that indicate an out of memory error (``MemoryError``, ``std::bad_alloc``, ``java.lang.OutOfMemoryError``, ``Out of memory``). (The @bgruening recommendation).
```
```suggestion
* ``exit_code``: adds checks for non zero exit codes. The ``oom_exit_code`` parameter can be used to add an additional out of memory indicating exit code. This is the default when a tool specifies a ``profile`` >= 16.04.
* ``aggressive``: adds checks for non zero exit codes, and checks for ``Exception:``, ``Error:`` in the standard error. Additionally checks for messages in the standard error that indicate an out of memory error (``MemoryError``, ``std::bad_alloc``, ``java.lang.OutOfMemoryError``, ``Out of memory``).
```

I think the planemo documentation is a bit more informal, but no need to trust @bgruening or @jmchilton on this one :).

mvdbeek (approved):

PR #10514, opened 2020-10-21
Drop communication server

### LINKED ISSUE #4217: Galaxy Chat is non functional
I found two issues:

- After the chat server set up I am unable to send messages and getting `raise KeyError('Session is disconnected')` 

- If I disable the chat in galaxy.ini but leave the option ON in the user prefs I will get `GET http://localhost:7070/ net::ERR_CONNECTION_REFUSED`in JS console in every Galaxy page. Additionally the user pref UI does not render this option (since config disabled it) so user has no means to fix this.

dannon:
Fixes #4217 via the method proposed in https://github.com/galaxyproject/galaxy/issues/4217#issuecomment-713105364

I don't think anyone's actively using this anymore, and most communication has moved to (also embeddable) things like Gitter, etc.


bgruening (approved):

doc/source/admin/galaxy_options.rst
```diff
@@ -2636,6 +2636,19 @@
 :Type: float
 
 
+~~~~~~~~~~~~~~~~~
```
dannon, line 2639: This got included with the config rebuild; left out before I guess.

lib/galaxy/webapps/base/static
```diff
@@ -0,0 +1 @@
+../../../../static
```
mvdbeek: Is that meant to be here ?

mvdbeek (approved):

lib/galaxy/webapps/base/static
mvdbeek: (and if yes, do we need to backport this ?)

dannon: Good eye -- unrelated (possible) bugfix for something Kaivan and I were looking at that I didn't notice got lumped in here.  Harmless, but it shouldn't be in this PR.

doc/source/admin/galaxy_options.rst
mvdbeek: if you're going to drop the symlink can you also unstage this ? This is missing in 20.09, I can take care of that (making a list of papercuts to backport now).

OlegZharkov:
Unrelated: we might want to update links in galaxy.ini files

OlegZharkov (approved):

test/unit/config/config_manage/1607_root_filters/config/galaxy.ini
```diff
@@ -1204,12 +1204,6 @@ use_interactive = True
 
 #amqp_internal_connection = sqlalchemy+sqlite:///./database/control.sqlite?isolation_level=IMMEDIATE
 
-# Galaxy real time communication server settings
-#enable_communication_server = False
-#communication_server_host = http://localhost
-#communication_server_port = 7070
-# persistent_communication_rooms is a comma-separated list of rooms that should be always available.
-#persistent_communication_rooms =
```
jmchilton: These files aren't meant to be changed, they're the configuration at older versions of Galaxy used to unit tests things we've changed over time (e.g. for linting). I doesn't hurt anything because they're comments but I think the right thing to do is keep the files the same and mark these options as deprecated and dropped in configuration linting. Here is a commit that does that if I can talk you into rebasing that into this: https://github.com/galaxyproject/galaxy/commit/295628bad657f46fe231355e11fc8708ba38e6ec

dannon: Yeah, I removed it all since it was just commented noise, but I don't have an opinion on this.  I'll incorporate 295628b if you want.

PR #10516, opened 2020-10-21
Fix changing versions in workflow editor

guerler:
This PR fixes a regression in the workflow editor. When switching between version the connections are not restored properly. Thanks for reporting this @afgane.


mvdbeek:
Does that need to go to 20.09, or further back ?

mvdbeek:
Fixes https://github.com/galaxyproject/galaxy/issues/10593.

Issue #10481, opened 2020-10-21
Being able to specify per-tool specs in the k8s runner


almahmoud:
Similar to how containers for specific tools can be mapped to use certain images/requests, it'd be cool to be able to override other specs for the job runner per tool (eg: use node affinity to redirect certain tools to certain kinds of nodepools, give a specific service account only to certain tools, etc...)

mvdbeek:
The way I understand the issue you can do that by mapping tools to destinations in your job conf file. That seems doable both with static or dynamic job destinations mapping.

almahmoud:
It is doable through job conf, but can be annoying to create a separate destination/environment just to change one job spec for a tool. The paper-cut is basically extending https://github.com/galaxyproject/galaxy-helm/blob/master/galaxy/files/rules/k8s_container_mapper.py to change more specs than just requests and container.

PR #10520, opened 2020-10-22
Override some k8s runner params with destination params

almahmoud:
Along with https://github.com/galaxyproject/galaxy-helm/pull/173 accomplishes https://github.com/galaxyproject/galaxy/issues/10481

### DIFF
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -140,6 +140,10 @@ class KubernetesJobRunner(AsynchronousJobRunner):
         job_wrapper.set_external_id(job_id)
         self.monitor_queue.put(ajs)
 
+    def __get_overridable_params(self, job_wrapper, param_key):
+        dest_params = self.__get_destination_params(job_wrapper)
+        return dest_params.get(param_key, self.runner_params[param_key])
+
     def __get_pull_policy(self):
         return pull_policy(self.runner_params)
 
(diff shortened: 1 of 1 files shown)


lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -227,8 +231,10 @@ def __get_k8s_job_spec_template(self, ajs):
                 "containers": self.__get_k8s_containers(ajs),
                 "priorityClassName": self.runner_params['k8s_pod_priority_class'],
                 "tolerations": yaml.safe_load(self.runner_params['k8s_tolerations'] or "[]"),
-                "affinity": yaml.safe_load(self.runner_params['k8s_affinity'] or "{}"),
-                "nodeSelector": yaml.safe_load(self.runner_params['k8s_node_selector'] or "{}")
+                "affinity": yaml.safe_load(self.__get_overridable_params(ajs.job_wrapper,
```
mvdbeek, line 234: Are these things something we could test in https://github.com/mvdbeek/galaxy/blob/4842e0dbb9203209edcccfc72675c9b90e7b8b2d/test/integration/test_kubernetes_runner.py#L93 ? Set to some value that would prevent running the job and then override it in the destination ?

almahmoud: Yeh I could try adding some tests

mvdbeek:
Thanks @almahmoud! If you ever want to get to writing tests I'm happy to help.

Issue #10519, opened 2020-10-21
/api/authenticate/baseauth expects email address instead of username


innovate-invent:
`POST /api/users` and then `GET /api/authenticate/baseauth` returns
```json
{"err_msg": "The user does not exist.", "err_code": 404001}
```
a second `POST /api/users` returns 400008: User with that email already exists.

```
172.17.0.1 - - [21/Oct/2020:23:31:54 +0000] "POST /api/users HTTP/1.1" 200 - "-" "terraform-provider-galaxy/galaxy - blend4go"
[pid: 765|app: 0|req: 1/1] 172.17.0.1 () {40 vars in 535 bytes} [Wed Oct 21 23:31:53 2020] POST /api/users => generated 247 bytes in 94 msecs (HTTP/1.1 200) 3 headers in 139 bytes (1 switches on core 2)
172.17.0.1 - - [21/Oct/2020:23:31:57 +0000] "GET /api/authenticate/baseauth HTTP/1.1" 404 - "-" "terraform-provider-galaxy/galaxy - blend4go"
[pid: 765|app: 0|req: 2/2] 172.17.0.1 () {42 vars in 611 bytes} [Wed Oct 21 23:31:57 2020] GET /api/authenticate/baseauth => generated 59 bytes in 17 msecs (HTTP/1.1 404) 3 headers in 146 bytes (1 switches on core 0)
172.17.0.1 - - [21/Oct/2020:23:33:40 +0000] "POST /api/users HTTP/1.1" 400 - "-" "terraform-provider-galaxy/galaxy - blend4go"
[pid: 765|app: 0|req: 3/3] 172.17.0.1 () {40 vars in 535 bytes} [Wed Oct 21 23:33:40 2020] POST /api/users => generated 119 bytes in 14 msecs (HTTP/1.1 400) 3 headers in 148 bytes (1 switches on core 3)
172.17.0.1 - - [21/Oct/2020:23:33:40 +0000] "GET /api/users?deleted=true&f_name=test HTTP/1.1" 200 - "-" "terraform-provider-galaxy/galaxy - blend4go"
[pid: 765|app: 0|req: 4/4] 172.17.0.1 () {38 vars in 556 bytes} [Wed Oct 21 23:33:40 2020] GET /api/users?deleted=true&f_name=test => generated 2 bytes in 11 msecs (HTTP/1.1 200) 3 headers in 139 bytes (1 switches on core 1)
```

I confirmed that both requests were using the same username/password.

[...]

PR #10521, opened 2020-10-22
[20.09] Allow username or email for authenticate API identity

dannon:
The primary authentication mechanism behaves this way.

Fixes https://github.com/galaxyproject/galaxy/issues/10519

### DIFF
--- a/lib/galaxy/managers/users.py
+++ b/lib/galaxy/managers/users.py
@@ -8,7 +8,7 @@ import time
 from datetime import datetime
 
 from markupsafe import escape
-from sqlalchemy import and_, desc, exc, func, true
+from sqlalchemy import and_, desc, exc, func, or_, true
 
 from galaxy import (
     exceptions,
@@ -276,6 +276,20 @@ class UserManager(base.ModelManager, deletable.PurgableManagerMixin):
         if user is None:
             # TODO: code is correct (401) but should be named AuthenticationRequired (401 and 403 are flipped)
             raise exceptions.AuthenticationFailed(msg, **kwargs)
+        return user
+
+    def get_user_by_identity(self, identity):
+        """Get user by username or email."""
+        session = self.session()
+        user = session.query(self.model_class).filter(or_(
+            self.model_class.table.c.email == identity,
+            self.model_class.table.c.username == identity,
+        )).first()
+        if not user:
+            # Try a case-insensitive match on the email
+            user = session.query(self.model_class).filter(
+                func.lower(self.model_class.table.c.email) == identity.lower()
+            ).first()
         return user
 
     # ---- current
--- a/lib/galaxy/managers/users.py
+++ b/lib/galaxy/managers/users.py
@@ -21,7 +21,7 @@ from galaxy.managers import (
     deletable
 )
 from galaxy.security.validate_user_input import (
-    VALID_PUBLICNAME_RE,
+    VALID_EMAIL_RE,
     validate_email,
     validate_password,
     validate_publicname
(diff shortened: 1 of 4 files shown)


innovate-invent:
Closes #10519

nsoranzo:
Should we consolidate the code here with https://github.com/galaxyproject/galaxy/blob/release_20.09/lib/galaxy/webapps/galaxy/controllers/user.py#L135-L145 , maybe in an additional function in `galaxy.security.validate_user_input` ?

dannon:
@nsoranzo Yeah, it'd be a good idea to followup with refactoring to have a single method for this -- it would have prevented this problem in the first place.

mvdbeek:
@dannon @nsoranzo what do you think about https://github.com/galaxyproject/galaxy/commit/5c7fdad1821d3fb4db032840e107b9eb7942a4ed ? This adds a method to `UserManager` for retrieving users via email or username, using the rules we had in https://github.com/galaxyproject/galaxy/blob/release_20.09/lib/galaxy/webapps/galaxy/controllers/user.py#L135-L145.

nsoranzo:
@mvdbeek Looks good to me, I just added a comment on your commit page.

dannon:
@mvdbeek I can cherrypick that here, but I think using the regex to identify the identification type is probably better than just OR'ing in the query.  I'll update it.  The two regexes match distinct sets and cannot overlap, so you know what kind of identity you're looking for.

mvdbeek:
Thanks @dannon!

lib/galaxy/managers/users.py
```diff
@@ -278,6 +279,22 @@ def error_if_anonymous(self, user, msg="Log-in required", **kwargs):
             raise exceptions.AuthenticationFailed(msg, **kwargs)
         return user
 
+    def get_user_by_identity(self, identity):
+        """Get user by username or email."""
+        user = None
+        if VALID_PUBLICNAME_RE.match(identity):
```
mvdbeek: The unit test fails now, because `VALID_PUBLICNAME_RE ` requires lowercase public names. Has that always been a requirement ? If yes `assert self.user_manager.get_user_by_identity(uppercase_user.username) == uppercase_user` can be dropped from the test, otherwise we could do:
```suggestion
        if VALID_PUBLICNAME_RE.match(identity.lower()):
```

test/unit/managers/test_UserManager.py
```diff
@@ -25,6 +25,8 @@
 user2_data = dict(email='user2@user2.user2', username='user2', password=default_password)
 user3_data = dict(email='user3@user3.user3', username='user3', password=default_password)
 user4_data = dict(email='user4@user4.user4', username='user4', password=default_password)
+uppercase_email_user = dict(email='USER5@USER5.USER5', username='USER5', password=default_password)
+lowercase_email_user = dict(email='user5@user5.user5', username='user5', password=default_password)
```
nsoranzo, line 29:
```suggestion
uppercase_email_user = dict(email='USER5@USER5.USER5', username='user5', password=default_password)
lowercase_email_user = dict(email='user5@user5.user5', username='user5bis', password=default_password)
```

lib/galaxy/managers/users.py
nsoranzo: > The unit test fails now, because VALID_PUBLICNAME_RE requires lowercase public names. Has that always been a requirement ?

Yes, I think so. I've made an alternative suggestion below.

dannon:
I did a quick check and there are actually public usernames with capitalization on main.

mvdbeek (approved):

nsoranzo (approved):

Issue #7721, opened 2019-04-10
No 'get api key' endpoint for remotely authenticated users


innovate-invent:
Currently only baseauth is supported. Need support for anyone who is already authenticated by other means.

innovate-invent:
Closed by #10522

PR #10522, opened 2020-10-22
Simpler API for getting user API key

### LINKED ISSUE #10515: Add API endpoint to allow admins to fetch users API keys
Currently `GET /api/authenticate/baseauth` is the only endpoint that exposes [this function](https://github.com/galaxyproject/galaxy/blob/cf11df1234f5f69edc98a8c7b776b87531fb621b/lib/galaxy/managers/api_keys.py#L16)

The only alternative is `GET /api/users/{id}/api_key/inputs`, parse an awkward response, check if the API key exists and `POST /api/users/{id}/api_key` if not.

Ideally `GET /api/users/{id}/api_key` should be mapped to the linked function that creates the api key if it does not exist.

dannon:
Minor refactoring to drop the deprecated mixin, too.

Fixes https://github.com/galaxyproject/galaxy/issues/10515

### DIFF
--- a/lib/galaxy/webapps/galaxy/buildapp.py
+++ b/lib/galaxy/webapps/galaxy/buildapp.py
@@ -752,6 +752,12 @@ def populate_api_routes(webapp, app):
                           action='api_key',
                           conditions=dict(method=["POST"]))
 
+    webapp.mapper.connect('api_key',
+                          '/api/users/{id}/api_key',
+                          controller='users',
+                          action='get_or_create_api_key',
+                          conditions=dict(method=["GET"]))
+
     webapp.mapper.connect('get_api_key',
                           '/api/users/{id}/api_key/inputs',
                           controller='users',
--- a/lib/galaxy/webapps/galaxy/api/users.py
+++ b/lib/galaxy/webapps/galaxy/api/users.py
@@ -20,7 +20,10 @@ from galaxy import (
     web
 )
 from galaxy.exceptions import ObjectInvalid
-from galaxy.managers import users
+from galaxy.managers import (
+    api_keys,
+    users
+)
 from galaxy.security.validate_user_input import (
     validate_email,
     validate_password,
@@ -55,6 +58,7 @@ class UserAPIController(BaseAPIController, UsesTagsMixin, CreatesApiKeysMixin, B
         self.user_manager = users.UserManager(app)
         self.user_serializer = users.UserSerializer(app)
         self.user_deserializer = users.UserDeserializer(app)
+        self.api_key_manager = api_keys.ApiKeyManager(app)
 
     @expose_api
     def index(self, trans, deleted='False', f_email=None, f_name=None, f_any=None, **kwd):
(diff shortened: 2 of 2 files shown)


lib/galaxy/webapps/galaxy/buildapp.py
```diff
@@ -736,6 +736,12 @@ def connect_invocation_endpoint(endpoint_name, endpoint_suffix, action, conditio
                           action='api_key',
                           conditions=dict(method=["POST"]))
 
+    webapp.mapper.connect('api_key',
+                          '/api/users/{id}/api_key',
+                          controller='users',
+                          action='get_or_create_api_key',
+                          conditions=dict(method=["GET"]))
```
mvdbeek, line 759: We've got separate endpoints for get_api_key (GET to `/api/users/{id}/api_key/inputs`) and set_api_key (`POST to /api/users/{id}/api_key'`), this feels like it's muddying the waters. Maybe we should just re-route /api/users/{id}/api_key/inputs to `GET /api/users/{id}/api_key/` ?

innovate-invent: The response of `/inputs` is a mess. I would much prefer only a json encoded string be returned like `/api/authenticate/baseauth`

jmchilton: Agreed - those endpoints are gross to work with because they're designed for the form API. I like the idea of an API for doing this that is designed to work as an API.

innovate-invent: The other issue is unless you expose the "get or create" function, it creates a race condition where the API key is overwritten by a second competing POST request generating an API key.

innovate-invent:
Related: https://github.com/galaxyproject/galaxy/issues/7721

lib/galaxy/webapps/galaxy/buildapp.py
mvdbeek: I mean, a GET shouldn't modify the underlying resource, and this does that. If you only see GET and `'/api/users/{id}/api_key'` you'd never think it could create an API key. At least make it a POST (or PUT, I don't mind), please ?

dannon: I think GET is actually the most reasonable thing here, actually -- it's still an idempotent operation.  PUT seems weird since you're not actually pushing any data in the payload, but that'd be reasonable too.  Let me think about this.

innovate-invent: POST is the correct operation for this, PUT wouldn't be.

dannon: Care to elaborate?  POST doesn't imply idempotency, which I see as the most relevant attribute of a 'get or create'.  I still think GET is the right action here.

innovate-invent: https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9

I suppose it is a practice vs specification dilemma. You are right, according to the spec a GET operation would be more appropriate. I think the larger issue is that the spec doesn't address whether GET should or shouldn't have side-effects. POST is supposed to create a subordinate at the specified path, which makes it inappropriate in that regard. PUT can't be used because it is supposed to have a request body that is assigned to the resource.

Because we are requesting the resource specified by the URI and not a subordinate, GET should be used.

Edit: this was also debated here: https://stackoverflow.com/questions/21900868/best-http-method-for-get-or-create

I think GET is more convenient as some clients may not be able to specify a request other than GET.

mvdbeek: Alright, GET it is.

PR #10523, opened 2020-10-22
Add S3 plugin for remote files browser

afgane:
Requires `fs-s3fs` package.

and the following in `file_sources_conf.yml` configuration:
```
- type: s3
  id: <Bucket label shown in the UI>
  doc: <Description for the bucket>
  bucket_name: <Bucket name>
  aws_access_key_id: <A_KEY>
  aws_secret_access_key: <S_KEY>
```


jmchilton:
Various test failures all seem related to transient issues with Yarn that I don't imagine could be related to this PR. I pulled it down and verified it didn't break existing webdav or posix stuff so it looks good to me. Thanks @afgane !

PR #10524, opened 2020-10-22
 informative error message on /api/authenticate/baseauth

### LINKED ISSUE #10513: /api/authenticate/baseauth throws exception when auth header missing
`GET /api/authenticate/baseauth` should return an informative response

[example](https://usegalaxy.org/api/authenticate/baseauth)

Currently returns
```json
{"err_msg": "Uncaught exception in exposed API method:", "err_code": 0}
```

OlegZharkov:
resolves https://github.com/galaxyproject/galaxy/issues/10513

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/authenticate.py
+++ b/lib/galaxy/webapps/galaxy/api/authenticate.py
@@ -95,10 +95,8 @@ class AuthenticationController(BaseAPIController):
         try:
             split = encoded_str.strip().split(' ')
         except AttributeError:
-            if not encoded_str:
-                raise exceptions.RequestParameterInvalidException('Authentication is missing')
-            else:
-                raise exceptions.MessageException('Failed to authenticate')
+            raise exceptions.RequestParameterInvalidException('Authentication is missing')
+
 
         # If split is only one element, try to decode the email and password
         # directly.
--- a/lib/galaxy/webapps/galaxy/api/authenticate.py
+++ b/lib/galaxy/webapps/galaxy/api/authenticate.py
@@ -97,7 +97,6 @@ class AuthenticationController(BaseAPIController):
         except AttributeError:
             raise exceptions.RequestParameterInvalidException('Authentication is missing')
 
-
         # If split is only one element, try to decode the email and password
         # directly.
         if len(split) == 1:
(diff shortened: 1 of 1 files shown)


lib/galaxy/webapps/galaxy/api/authenticate.py
```diff
@@ -92,7 +92,15 @@ def _decode_baseauth(self, encoded_str):
...
-        split = encoded_str.strip().split(' ')
+        try:
+            split = encoded_str.strip().split(' ')
+        except AttributeError:
+            if not encoded_str:
+                raise exceptions.MessageException(
+                    'Authentication is missing')
+            else:
+                raise exceptions.MessageException(
+                    'Failed to authenticate')
```
dannon: This is more readable if we don't wrap it so short, right?  We don't strictly adhere to 80 in the Galaxy codebase.

```suggestion
                raise exceptions.MessageException('Authentication is missing')
            else:
                raise exceptions.MessageException('Failed to authenticate')
```

lib/galaxy/webapps/galaxy/api/authenticate.py
```diff
@@ -92,7 +92,13 @@ def _decode_baseauth(self, encoded_str):
 
         :raises: HTTPBadRequest
         """
-        split = encoded_str.strip().split(' ')
+        try:
+            split = encoded_str.strip().split(' ')
+        except AttributeError:
+            if not encoded_str:
+                raise exceptions.MessageException('Authentication is missing')
```
dannon: This should be the more specific RequestParameterInvalidException
```suggestion
                raise exceptions.RequestParameterInvalidException('Authentication is missing')
```

lib/galaxy/webapps/galaxy/api/authenticate.py
```diff
@@ -92,7 +92,13 @@ def _decode_baseauth(self, encoded_str):
...
+            else:
```
dannon: In what other circumstances would this throw an AttributeError?

OlegZharkov: if it's called from another function on backend? I thought it won't be good to always return 'Authentication is missing', but I guess you're right. I will remove it.

dannon: Yeah, what I was getting at was how else would this throw an AttributeError other than encoded_str being missing.  Agreed, the single "RequestParameterInvalidException" is enough.

PR #10526, opened 2020-10-22
More comments for rapid testing optional galaxy.files stuff.

jmchilton:


test/integration/test_webdav.py
```diff
@@ -1,9 +1,7 @@
-# TODO:
-
-# nginx+webdav in Docker.
-# docker run -v /Users/john/workspace/galaxy/test/integration/webdav/data:/media  -e WEBDAV_USERNAME=alice -e WEBDAV_PASSWORD=secret1234 -p 7083:7083 jmchilton/webdavdev
-# Apache Docker host doesn't work because displayname not set in response.
-# docker run -v /Users/john/workspace/galaxy/test/integration/webdav:/var/lib/dav  -e AUTH_TYPE=Basic -e USERNAME=alice -e PASSWORD=secret1234  -e LOCATION=/ -p 7083:80 bytemark/webdav
+# Before running this test, start nginx+webdav in Docker using following command:
+# docker run -v `pwd`/test/integration/webdav/data:/media  -e WEBDAV_USERNAME=alice -e WEBDAV_PASSWORD=secret1234 -p 7083:7083 jmchilton/webdavdev
+# Apache Docker host (shown next) doesn't work because displayname not set in response.
+# docker run -v `pwd`/test/integration/webdav:/var/lib/dav  -e AUTH_TYPE=Basic -e USERNAME=alice -e PASSWORD=secret1234  -e LOCATION=/ -p 7083:80 bytemark/webdav
```
mvdbeek, line 4: In a good amount of other integration tests we start a docker container as part of the test, maybe we could do that here too ? My experience is that tests that don't run in a PR usually break over time.

jmchilton: Agreed it would be better than not, but I don't want perfect to be the enemy of the good. Good to have well documented manual tests even if they aren't in CI.

mvdbeek: Fair enough!

PR #10529, opened 2020-10-22
'copy to clipboard' button on datasets

### LINKED ISSUE #4073: offer 'copy to clipboard' button on datasets
To make clear and easier that people can share the datasets using that URL.

@natefoo 's nice idea

OlegZharkov:
November 3rd is almost there. So why don't we do our own galaxy-elections? :)

if you are in favour of that button, please vote with :+1:
image

This one is  :rocket: 
image

Finally, this option is  :tada:  
image

Thanks, choose wisely

resolves https://github.com/galaxyproject/galaxy/issues/4073

### DIFF
--- a/client/src/mvc/tool/tool-form-base.js
+++ b/client/src/mvc/tool/tool-form-base.js
@@ -183,7 +183,7 @@ export default FormBase.extend({
             tooltip: "View available options",
         });
         menu_button.addMenu({
-            icon: "fa-share",
+            icon: "fa-chain",
             title: _l("Get link"),
             onclick: function () {
                 if (navigator.clipboard) {
--- a/client/src/mvc/tool/tool-form-base.js
+++ b/client/src/mvc/tool/tool-form-base.js
@@ -184,7 +184,7 @@ export default FormBase.extend({
         });
         menu_button.addMenu({
             icon: "fa-chain",
-            title: _l("Get link"),
+            title: _l("Copy link"),
             onclick: function () {
                 if (navigator.clipboard) {
                     navigator.clipboard
--- a/client/src/mvc/dataset/dataset-li.js
+++ b/client/src/mvc/dataset/dataset-li.js
@@ -303,7 +303,7 @@ export var DatasetListItemView = _super.extend(
             var urls = this.model.urls;
             if (!this.isPurged() && urls.download && isSharable)
                 return faIconButton({
-                    faIcon: "fa-share",
+                    faIcon: "fa-chain",
                     title: _l("Copy link"),
                     onclick: function () {
                         navigator.clipboard.writeText(`${window.location.origin}${urls.download}`).then(() => {
--- a/client/src/mvc/dataset/dataset-li.js
+++ b/client/src/mvc/dataset/dataset-li.js
@@ -299,8 +299,9 @@ export var DatasetListItemView = _super.extend(
          *  @returns {jQuery} rendered DOM
          */
         _renderClipboardButton: function () {
+            const isSharable = this.model.attributes.permissions.access.length !== 1;
             var urls = this.model.urls;
-            if (!this.isPurged() && urls && urls.download)
+            if (!this.isPurged() && urls.download && isSharable)
                 return faIconButton({
                     faIcon: "fa-share",
                     title: _l("Copy link"),
--- a/client/src/mvc/dataset/dataset-li.js
+++ b/client/src/mvc/dataset/dataset-li.js
@@ -299,7 +299,8 @@ export var DatasetListItemView = _super.extend(
          *  @returns {jQuery} rendered DOM
          */
         _renderClipboardButton: function () {
-            const isSharable = this.model.attributes.permissions.access.length !== 1;
+            const isSharable =
+                this.model.attributes.permissions && this.model.attributes.permissions.access.length !== 1;
             var urls = this.model.urls;
             if (!this.isPurged() && urls.download && isSharable)
                 return faIconButton({
--- a/client/src/mvc/dataset/dataset-li.js
+++ b/client/src/mvc/dataset/dataset-li.js
@@ -9,7 +9,6 @@ import BASE_MVC from "mvc/base-mvc";
 import _l from "utils/localization";
 import { mountNametags } from "components/Nametags";
 import { Toast } from "ui/toast";
-import { getAppRoot } from "onload/loadConfig";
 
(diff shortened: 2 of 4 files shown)


innovate-invent:
Does this produce a popup that allows manually copying the link? Some browsers block pages from modifying the clipboard.

OlegZharkov:
> Does this produce a popup that allows manually copying the link? Some browsers block pages from modifying the clipboard.

If by 'some browsers' you mean IE we stopped supporting it a while ago. But if there's a real demand... sure, we can add 'manual copy' pop-up

> Also, my vote is #3

then put :tada: emoji :)

innovate-invent:
I use Firefox and have experienced issues with GitHubs 'clone url' copy feature. Perhaps it is just GitHub though.

hexylena:
Since we're bikeshedding: To me, the action is "Share". You're getting a link (chain), sure, but in order to send to someone else. So the Share icon makes the most sense in the context of what the user is trying to do.

OlegZharkov:
> Since we're bikeshedding: To me, the action is "Share". You're getting a link (chain), sure, but in order to send to someone else. So the Share icon makes the most sense in the context of what the user is trying to do.

Well, to advocate for that, I also prefer https://fontawesome.com/icons/share, because we already have it in tools. Since it was there for years, galaxy users might have gotten used to this icon and the button functionality would be intuitive.
image

nsoranzo:
> To me, the action is "Share". You're getting a link (chain), sure, but in order to send to someone else. So the Share icon makes the most sense in the context of what the user is trying to do.

Sure, but nowadays if I see a share icon (e.g. https://material.io/resources/icons/?icon=share&style=baseline ) on my phone, I expect to have a series of sharing option (various social networks, messaging apps, email) and probably also a copy to clipboard, which normally look like options :rocket: or :tada: .
BTW, just noticed that Firefox on desktop also use the chain icon to "Copy Link" from the address bar.

> Well, to advocate for that, I also prefer https://fontawesome.com/icons/share, because we already have it in tools. Since it was there for years, galaxy users might have gotten used to this icon and the button functionality would be intuitive.

I don't think the average user needs to open that menu very often, to be honest.

natefoo:
I didn't know the 👍 option was a share icon since on iOS it's [different](https://developer.apple.com/design/human-interface-guidelines/ios/icons-and-images/system-icons/). Guess it's more like that on Android?

hexylena:
> Sure, but nowadays if I see a share icon ... on my phone

Ah the fabled phone user of Galaxy :P I definitely have that expectation too, on the phone. But on the desktop I don't have that expectation, which I think is probably the bulk of users? I wish we had like, a group of biologists/end users we could poll on 5 minutes notice, sometimes.

> I don't think the average user needs to open that menu very often, to be honest.

ahhh this is probably where I get my expectation within galaxy from, since I use that menu option super often.

nsoranzo:
> I didn't know the +1 option was a share icon since on iOS it's [different](https://developer.apple.com/design/human-interface-guidelines/ios/icons-and-images/system-icons/). Guess it's more like that on Android?

No, on Android it's the one I linked above: https://material.io/resources/icons/?icon=share&style=baseline

martenson:
I am fine with any as long as the popover says `Copy link` 🤣

simonbray:
Copied from gitter:
> just to point out, dataset sharing has to be enabled in user preferences for this to work
> I don't know if there is any good way to inform users of that. It took me a very long time to realize that option existed.

Also, I never clicked on that 'Share tool' icon once, which I suppose goes to show that it's easy to create nice features and hard to persuade users to use them. :)

simonbray:
FWIW an image search for 'copy link to clipboard' returns a lot of screenshots with chain and clipboard icons

bgruening:
>  I wish we had like, a group of biologists/end users we could poll on 5 minutes notice, sometimes.

One of our power users has just voted :)

OlegZharkov:
> Copied from gitter:
> 
> > just to point out, dataset sharing has to be enabled in user preferences for this to work
> > I don't know if there is any good way to inform users of that. It took me a very long time to realize that option existed.

thanks a lot, that's a valid point. If user restricted permissions only to himself, the button will be hidden now.


Also we have a winner, :tada: leads by triple digits! Thank everyone for voting in this poll!

nsoranzo:
@OlegZharkov I think it makes sense to also use the chain icon and "Copy link" text for sharing a tool URL.

OlegZharkov:
> @OlegZharkov I think it makes sense to also use the chain icon and "Copy link" text for sharing a tool URL.

right, will do

client/src/mvc/tool/tool-form-base.js
```diff
@@ -183,7 +183,7 @@ export default FormBase.extend({
             tooltip: "View available options",
         });
         menu_button.addMenu({
-            icon: "fa-share",
+            icon: "fa-chain",
             title: _l("Get link"),
```
nsoranzo:
```suggestion
            title: _l("Copy link"),
```

simonbray:
> > Copied from gitter:
> > > just to point out, dataset sharing has to be enabled in user preferences for this to work
> > > I don't know if there is any good way to inform users of that. It took me a very long time to realize that option existed.
> 
> thanks a lot, that's a valid point. If user restricted permissions only to himself, the button will be hidden.

Very cool, thanks a lot! And it looks like you even are doing it on the basis of the individual dataset's permissions, which is really nice.

hexylena:
> thanks a lot, that's a valid point. If user restricted permissions only to himself, the button will be hidden.

isn't this more confusing for the user? That the button isn't always there for some users and they don't know why? 

I know it would expand scope significantly but I'd really expect that the button exists always, and it guides the user through changing the permissions.

Or at least having it there but disabled with a note "your preferences do not permit sharing links to datasets" so they can know to go edit it.

nsoranzo:
> Or at least having it there but disabled with a note "your preferences do not permit sharing links to datasets" so they can know to go edit it.

:+1:

simonbray:
> But I am not sure, whether it's a good idea to show users a huge list of disabled buttons

I think it's ok, it also provides a nice visual way for the user to quickly see which datasets are shared. :+1:

dannon (approved):

dannon:
Thanks @OlegZharkov

OlegZharkov:
Thanks @dannon and everyone who participated in this poll!

PR #10533, opened 2020-10-22
Disable upload option in default data dialog since history panel is not always available

guerler:
Currently the upload option is shown by default when the data dialog is displayed. However at the moment this option is designed to work only with an existing history panel and for the content selector in the tool form. This PR adds a flag to the data dialog to avoid displaying the upload dialog by default.

### DIFF
--- a/client/src/components/DataDialog/DataDialog.vue
+++ b/client/src/components/DataDialog/DataDialog.vue
@@ -23,7 +23,7 @@
                 <div class="fa fa-caret-left mr-1" />
                 Back
             </b-btn>
-            <b-btn size="sm" class="float-left mr-1" @click="onUpload">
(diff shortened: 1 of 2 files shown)


jmchilton:
This PR adds the same flag https://github.com/galaxyproject/galaxy/pull/10505/files 😅 .

jmchilton (approved):

guerler:
Oh ok, cool. Thanks for the quick review. I made an attempt to modify the uploader such that it works with the history id alone but it would require some additional changes. Apparently the current history collections object is accessed directly at one location. https://github.com/galaxyproject/galaxy/compare/dev...guerler:generalize_uploader?expand=1

jmchilton:
This should probably be backported to 20.09?

PR #10405, opened 2020-10-12
Make Avivator GEDA for OME Tiff

qiagu:

PR #10534, opened 2020-10-22
Make Avivator GEDA and sniffer for ome.tiff datatype

qiagu:
This is a smaller RP split from https://github.com/galaxyproject/galaxy/pull/10405


lib/galaxy/datatypes/images.py
```diff
@@ -2,10 +2,13 @@
 Image classes
 """
 import base64
+import json
 import logging
+import tifffile
 import zipfile
 from urllib.parse import quote_plus
 
+from galaxy.datatypes.metadata import FileParameter, MetadataElement
```
mvdbeek:
```suggestion
from galaxy.datatypes.metadata import (
    FileParameter,
    MetadataElement,
)
```

mvdbeek (commented):
@qiagu can you resolve the conflicts here? It also appears that you can sniff ome.tiff files (as you've added that to `image_type`), so 
- could you add a `sniff` method to the OMETiff class
- add a small test file with the ome.tiff extension to lib/galaxy/datatypes/test
- and remove `subclass="true"` from the datatype config
?

Otherwise this looks fine to me.

mvdbeek:
@qiagu I'll push this to 21.05 due to the outstanding comments, but datatypes can be added before we make the actual release.

qiagu:
@mvdbeek Sorry for taking so long before coming back this PR. 
> could you add a sniff method to the OMETiff class

I don't see any other image class having sniff method and the sniff is mixed with other image types. Do we still need a specific sniff method?

> and remove subclass="true" from the datatype config

It was removed in one of the commits. Could you please double check that?

lib/galaxy_test/api/test_tools_upload.py
```diff
@@ -223,6 +223,13 @@ def test_image_upload_auto(self):
             tiff_metadata = self._upload_and_get_details(fh, file_type="auto")
         self.assertEqual(tiff_metadata["file_ext"], "tiff")
 
+    @skip_without_datatype("ome.tiff")
+    def test_ometiff_upload_sniff(self):
+        tiff_path = TestDataResolver().get_filename("1.ome.tiff")
+        with open(tiff_path, "rb") as fh:
+            tiff_metadata = self._upload_and_get_details(fh, file_type="auto")
+        self.assertEqual(tiff_metadata["file_ext"], "ome.tiff")
+
```
mvdbeek: Please remove this and move  test-data/1.ome.tiff to lib/galaxy/datatypes/test, this will then be automatically tested without having to write a test.

lib/galaxy/dependencies/pinned-requirements.txt
```diff
@@ -175,6 +175,7 @@ stevedore==3.3.0; python_version >= "3.6"
 svgwrite==1.4.1; python_version >= "3.6"
 tempita==0.5.2
 tenacity==6.3.1
+tifffile==2021.2.26
```
mvdbeek, line 190: Can you also add the unpinned dependency to https://github.com/galaxyproject/galaxy/blob/dev/pyproject.toml ?

lib/galaxy_test/api/test_tools_upload.py
qiagu: Done.

lib/galaxy/dependencies/pinned-requirements.txt
qiagu: Done

qiagu:
@mvdbeek  please check again.

lib/galaxy/datatypes/images.py
```diff
@@ -78,6 +84,34 @@ class Tiff(Image):
...
+            offsets_file = dataset.metadata.offsets
+            if not offsets_file:
+                offsets_file = dataset.metadata.spec[spec_key].param.new_file(dataset=dataset)
+            with tifffile.TiffFile(dataset.file_name) as tif:
+                offsets = [page.offset for page in tif.pages]
+            with open(offsets_file.file_name, 'w') as f:
+                json.dump(offsets, f)
+            dataset.metadata.offsets = offsets_file
+        except Exception:
+            pass
```
mvdbeek:
```suggestion
        spec_key = 'offsets'
        offsets_file = dataset.metadata.offsets
        if not offsets_file:
            offsets_file = dataset.metadata.spec[spec_key].param.new_file(dataset=dataset)
        with tifffile.TiffFile(dataset.file_name) as tif:
            offsets = [page.offset for page in tif.pages]
        with open(offsets_file.file_name, 'w') as f:
            json.dump(offsets, f)
        dataset.metadata.offsets = offsets_file
```
I suppose if the metadata setting fails you're not dealing with a valid OMETiff file, and in that case it's better to fail here.

lib/galaxy/datatypes/images.py
```diff
@@ -78,6 +84,34 @@ class Tiff(Image):
...
+
+    def sniff(self, filename):
+        is_tiff = Tiff().sniff(filename)
+        if not is_tiff:
+            return False
```
mvdbeek:
```suggestion
```
that seems like a redundant check

mvdbeek (approved):

lib/galaxy/datatypes/images.py
```diff
@@ -89,27 +89,24 @@ class OMETiff(Tiff):
...
+        with tifffile.TiffFile(dataset.file_name) as tif:
+            offsets = [page.offset for page in tif.pages]
+        with open(offsets_file.file_name, 'w') as f:
+            json.dump(offsets, f)
+        dataset.metadata.offsets = offsets_file
 
     def sniff(self, filename):
-        is_tiff = Tiff().sniff(filename)
-        if not is_tiff:
+        try:
```
mvdbeek: There's no need for the exception handling, an exception is a negative result in the sniffer

mvdbeek:
There's a merge conflict now, can you resolve this @qiagu ?

lib/galaxy/datatypes/images.py
```diff
@@ -89,27 +89,24 @@ class OMETiff(Tiff):
...
+            with tifffile.TiffFile(filename) as tif:
+                if tif.is_ome:
+                    return True
+                return False
+        except tifffile.TiffFileError:
```
qiagu: @mvdbeek  Thanks for your great comments.
The sniff method has been revised. Catch either `tifffile.TiffFileError` error or just catch a general `Exception`, which one do you think is better here?

lib/galaxy/datatypes/images.py
qiagu: Okay. Good to know this.

qiagu: May work in Galaxy without an exception catcher. The sniff method can't be used standalone though.

mvdbeek:
@qiagu can you figure out which version of tiffile is the last version that supports python 3.6 ?

qiagu:
Ah, `2020.9.3` should be the right version. Thanks for pointing that out!
I accidentally delete this branch. A new PR will be initiated. Sorry for the inconvenience caused.

PR #10537, opened 2020-10-23
Generate path resolution note for sample file, docs

### LINKED ISSUE #10531: Add note on path resolving to config options in schema - to be displayed in sample config file, docs
Config option paths are resolved automatically whenever the `path_resolves_to` option is specified in `config_schema.yml`. However, now that the hardcoded prefixes (`config/`, `data/`, etc.) have been removed from schema default values, this behavior is no longer evident.

jdavcs:
Fixes #10531.


jmchilton:
Thanks!

PR #10540, opened 2020-10-23
Allow admins to add an additional custom logo to the masthead

guerler:
This PR adds additional configuration options which enable admins to specify custom logos in addition to the default Galaxy logo.


lib/galaxy/managers/configuration.py
```diff
@@ -43,9 +43,9 @@ def _use_config(config, key, **context):
             'is_admin_user'                     : lambda *a, **c: False,
             'brand'                             : _use_config,
             'display_galaxy_brand'              : _use_config,
-            # TODO: this doesn't seem right
             'logo_url'                          : lambda config, key, **context: self.url_for(config.get(key, '/')),
-            'logo_src'                          : lambda config, key, **context: self.url_for('/static/favicon.png'),
+            'logo_src'                          : lambda config, key, **context: self.url_for(config.get(key, '/static/favicon.png')),
+            'logo_src_secondary'                : lambda config, key, **context: self.url_for(config.get(key)) if config.get(key) else None,
```
jdavcs, line 48:
```suggestion
            'logo_src_secondary'                : lambda config, key, **context: self.url_for(config.get(key)),
```
Since it's in the schema, `None` is already the default here - i.e., `config.get(key)` will always return whatever `logo_src_secondary` is set to, or `None`.

guerler: Yes but you cannot execute `self.url_for` with `None`, that is why a single line if is needed.

jdavcs: Oh, I see. Yes, this makes sense.

PR #10544, opened 2020-10-23
Move Selenium to GitHub Actions

OlegZharkov:
It's pretty much an experimental PR and still a proof of concept. We dynamically create a separate job for each test and run them in parallel (github allows 20 parallel jobs at a time) 
Most probably there is a much better/less hacky way to derive list of  tests than my `test_parser.py`, if so please let me know. 
After it is done, we can harvest screenshots and upload them as artifacts (github stores it for 3 months). 
We will improve the performance dramatically, if we skip client build in every job (will add caching later)

### DIFF
--- a/.ci/run_selenium.sh
+++ b/.ci/run_selenium.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+# Prepare
+python -m venv .venv
+
+source ".venv/bin/activate" # activate virtualenv
+pip install -r lib/galaxy/dependencies/dev-requirements.txt -r requirements.txt  &> /dev/null
+python test/manual/test_parser.py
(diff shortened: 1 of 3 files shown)


mvdbeek:
> run them in parallel (github allows 20 parallel jobs at a time)

It's either 40 or 60 for us, but that is across the entire repo. I don't think we'd want more than 1 or 2 jobs.

OlegZharkov:
> > run them in parallel (github allows 20 parallel jobs at a time)
> 
> It's either 40 or 60 for us, but that is across the entire repo. I don't think we'd want more than 1 or 2 jobs.

40 or 60 per repo? that's quite sad, I thought it's 20 per PR. Should we do the same on jenkins... maybe? Otherwise I can limit it to 2 parallel jobs only... but that might take some time

PR #10548, opened 2020-10-24
[20.09] Fix - allow working around tool tests we expect to fail.

jmchilton:
- Create a YAML file for specifying tool test options.
- Allow skipping tests with a message in that file.
- Allow annotating other metadata about tests (issues/PRs tracked, notes for collaborators, etc..).

Idea would be to check this file into a version control and track the effort of getting tool sets working on a server -or- architecture. I looked at @almahmoud's list of failing AnVIL tests and the first thing I noticed is a set of tests that I knew indicated the tool was working even though one test passed.

If we like the syntax and this gets merged, I'll open a Planemo PR to add display of this metadata to the generated reports.


mvdbeek (approved):
That's great!

PR #10555, opened 2020-10-26
[20.09] AnnData Peek Fix

mtekman:
This should fix the problematic shape parameter detection for AnnData datasets

Once this is merged, this PR (https://github.com/galaxyproject/tools-iuc/pull/3237) should go green.


lib/galaxy/datatypes/binary.py
```diff
@@ -981,12 +981,15 @@ def set_meta(self, dataset, overwrite=True, **kwd):
         super(Anndata, self).set_meta(dataset, overwrite=overwrite, **kwd)
         try:
             with h5py.File(dataset.file_name, 'r') as anndata_file:
+                ## NOTE: It appears that 'dataset.metadata' variables can
+                ##       only be set exactly once, so tread carefully.
```
```suggestion
```
That does not seem right.

lib/galaxy/datatypes/binary.py
```diff
@@ -981,12 +981,15 @@ def set_meta(self, dataset, overwrite=True, **kwd):
...
                 dataset.metadata.title = util.unicodify(anndata_file.attrs.get('title'))
                 dataset.metadata.description = util.unicodify(anndata_file.attrs.get('description'))
                 dataset.metadata.url = util.unicodify(anndata_file.attrs.get('url'))
                 dataset.metadata.doi = util.unicodify(anndata_file.attrs.get('doi'))
                 dataset.creation_date = util.unicodify(anndata_file.attrs.get('creation_date'))
                 # none of the above appear to work in any dataset tested, but could be useful for future
+                # dataset.metadata.shape = util.unicodify(anndata_file.attrs.get('shape'))
```
```suggestion
```

lib/galaxy/datatypes/binary.py
```diff
@@ -1040,15 +1047,20 @@ def _layercountsize(tmp, lennames=0):
...
                     elif hasattr(anndata_file['X'], 'shape'):
                         dataset.metadata.shape = tuple(anndata_file['X'].shape)
-                    else:
-                        dataset.metadata.shape = (int(dataset.metadata.obs_size), int(dataset.metadata.var_size))
+                ##
+                elif n_obs >=0 and n_var >= 0:
+                    dataset.metadata.shape = (int(n_obs), int(n_var))
+                else:
+                    log.warning("No shape data could be determined!")
+                    dataset.metadata.shape = tuple(-1,-1)
```
```suggestion
                else:
                    dataset.metadata.shape = (int(dataset.metadata.obs_size or -1), int(dataset.metadata.var_size or -1))
```
I think that should be the core of the changes you're making. You can remove `n_var` and `n_obs` then.
Or you can do
`dataset.metadata.shape = (int(dataset.metadata.obs_size), int(dataset.metadata.var_size))` like you've had it before and set `obs_size ` and `var` size defaults to -1 using the `MetadataElement `definitions.

mvdbeek (commented):
Can you add the problematic dataset (or a shrunk down variant) in question to lib/galaxy/datatypes/test ?
That will run the upload tests against that file.

mtekman:
@mvdbeek Thanks for the suggestions, they greatly simplified the if-else grave I was digging there.

I will test again my claim that the `dataset.metadata.shape` can only be set once -- maybe the behaviour I was seeing was a result of the poorly defined defaults it had before.

lib/galaxy/datatypes/binary.py
mtekman: Hmm, so maybe my conclusion  was wrong but I am definitely seeing weird behaviour if add 

`dataset.metadata.shape = util.unicodify(anndata_file.attrs.get('shape'))`  on L995

The error message I get is:


```
  File "/home/nope/repos/_galaxy/galaxy-upstream/lib/galaxy/jobs/__init__.py", line 1547, in _finish_dataset
    dataset.set_peek(line_count=line_count)
  File "/home/nope/repos/_galaxy/galaxy-upstream/lib/galaxy/model/__init__.py", line 2779, in set_peek
    return self.datatype.set_peek(self, **kwd)
TypeError: set_peek() got an unexpected keyword argument 'line_count'

  File "/home/nope/repos/_galaxy/galaxy-upstream/lib/galaxy/jobs/runners/__init__.py", line 540, in _finish_or_resubmit_job
    job_wrapper.finish(tool_stdout, tool_stderr, exit_code, check_output_detected_state=check_output_detected_state, job_stdout=job_stdout, job_stderr=job_stderr)
  File "/home/nope/repos/_galaxy/galaxy-upstream/lib/galaxy/jobs/__init__.py", line 1681, in finish
    self._finish_dataset(
  File "/home/nope/repos/_galaxy/galaxy-upstream/lib/galaxy/jobs/__init__.py", line 1550, in _finish_dataset
    dataset.set_peek()
  File "/home/nope/repos/_galaxy/galaxy-upstream/lib/galaxy/model/__init__.py", line 2779, in set_peek
    return self.datatype.set_peek(self, **kwd)
  File "/home/nope/repos/_galaxy/galaxy-upstream/lib/galaxy/datatypes/binary.py", line 1077, in set_peek
    peekstr = "[n_obs x n_vars]\n    %d x %d" % tuple(tmp.shape)
TypeError: 'NoneType' object is not iterable
FAIL
```

So I read this as:

 * `dataset.metadata.shape = util.unicodify(anndata_file.attrs.get('shape'))`  evaluates to `None` 
 * The `if 'X' in dataset.metadata.layers_names:` block lower down is either never reached, or the `dataset.metadata.shape` variable is never overwritten.

Some more tests:

* Evaluating `util.unicodify(anndata_file.attrs.get('shape'))` does not produce any errors, so it's the assignment to `dataset.metadata.shape` that seems to  cause the problem
* Surrounding `dataset.metadata.shape = util.unicodify(anndata_file.attrs.get('shape'))` in a try-catch block does not fix the issue, so I think the if-else block at the bottom *is* being reached.

mvdbeek: Can you remove the try/except catching on line 984 / 1058 ?
From what you describe I think the most likely explanation is that you're raising an exception, and the set_peek code relies on no exception having occurred.

That said I don't see `dataset.metadata.shape = util.unicodify(anndata_file.attrs.get('shape'))` on line 995, has something gone wrong there ? ` File "/home/nope/repos/_galaxy/galaxy-upstream/lib/galaxy/datatypes/binary.py", line 1077, in set_peek` also doesn't match the line numbers here.

mtekman: oh I meant "if you add {code snippet} at this line, this is what you get"

I can push this as a commit if you like, but I'd rather leave it out of the git history

mvdbeek: We can always rebase this out.

mtekman: > Can you remove the try/except catching on line 984 / 1058 ?

We should keep this though, right? I took this template from one of the other h5 classes in the file (Loom)

> From what you describe I think the most likely explanation is that you're raising an exception, and the set_peek code relies on no exception having occurred.

though I can see now that this is main try-except block is what was hiding the errors

mtekman:
@mvdbeek it looks like it worked, (though for reasons I do not understand), where removing the surrounding try-except apparently does not produce the errors associated with the `dataset.metadata.shape = util.unicodify(anndata_file.attrs.get('shape'))` I was seeing before.

lib/galaxy/datatypes/binary.py
```diff
@@ -978,14 +980,14 @@ def sniff(self, filename):
...
-        super(Anndata, self).set_meta(dataset, overwrite=overwrite, **kwd)
-        try:
+            super(Anndata, self).set_meta(dataset, overwrite=overwrite, **kwd)
             with h5py.File(dataset.file_name, 'r') as anndata_file:
                 dataset.metadata.title = util.unicodify(anndata_file.attrs.get('title'))
                 dataset.metadata.description = util.unicodify(anndata_file.attrs.get('description'))
                 dataset.metadata.url = util.unicodify(anndata_file.attrs.get('url'))
                 dataset.metadata.doi = util.unicodify(anndata_file.attrs.get('doi'))
                 dataset.creation_date = util.unicodify(anndata_file.attrs.get('creation_date'))
+                dataset.metadata.shape = util.unicodify(anndata_file.attrs.get('shape'))
```
mvdbeek: `unicodify` in essence turns bytestrings to unicode strings, passing through `None`. `anndata_file.attrs.get('shape')` should be `None` or a tuple, of integers, right ? If you get a tuple from `anndata_file.attrs.get('shape')`  and you pass it into unicodify you'll get a string back out:
```python
In [2]: from galaxy.util import unicodify

In [3]: unicodify((2,2))
Out[3]: '(2, 2)'
```
So that makes me think we have no test yet for a file that specifies the shape.
I would guess that the following is closer to what you want to do:

```suggestion
                dataset.metadata.shape = anndata_file.attrs.get('shape') or dataset.metadata.shape
```
That means you can drop line 1054/1055.

If you decide not to wrap this part in try/except you need to dedent the set_meta function body by 4 spaces.

mtekman: I committed the line and unindented the region, but I kept line 1054/55 since I don't quite see how we're handling cases which have no obvious shape assignment but do have observations and variables.

mvdbeek (approved):

mtekman:
:tada:

PR #10569, opened 2020-10-28
[20.09] Allow execution of workflows for certain tool updates

mvdbeek:
This works around workflows not being executable when versions change.
We usually don't break any parameters or connections, so when we bump a
version of a tool shipped with Galaxy we can add the tool to
`WORKFLOW_SAFE_TOOL_VERSION_UPDATES`. That will let users run the
workflow, while still maintaining a warning that the tool version
doesn't match.

### DIFF
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -1,6 +1,7 @@
 """
 Classes encapsulating galaxy tools and tool configuration.
 """
+import collections
 import itertools
 import json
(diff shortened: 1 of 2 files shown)


mvdbeek:
This doesn't necessarily have to make it into the release, but I think it's a good way to deal with "disappearing" versions of Galaxy-shipped tools until we have a way to also serve old versions.

bgruening (approved):

PR #10572, opened 2020-10-28
[20.09] Stop monitoring terminal Pulsar Kubernetes jobs

natefoo:
@jmchilton also, does it make sense that we query for ports on every loop? It seems like we should probably stop after they've been obtained (unless we expect them to possibly change after first being obtained?).

### DIFF
--- a/lib/galaxy/jobs/runners/pulsar.py
+++ b/lib/galaxy/jobs/runners/pulsar.py
@@ -265,6 +265,10 @@ class PulsarJobRunner(AsynchronousJobRunner):
         if self.use_mq:
             # Might still need to check pod IPs.
             job_wrapper = job_state.job_wrapper
+            persisted_state = job_wrapper.get_state()
+            if persisted_state in model.Job.terminal_states + [model.Job.states.DELETED_NEW]:
+                log.debug("(%s) Watched job in terminal state, will stop monitoring: %s", job_state.job_id, persisted_state)
+                return None
             guest_ports = job_wrapper.guest_ports
             if len(guest_ports) > 0:
                 client = self.get_client_from_state(job_state)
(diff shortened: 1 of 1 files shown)


lib/galaxy/jobs/runners/pulsar.py
```diff
@@ -265,6 +265,10 @@ def check_watched_item(self, job_state):
         if self.use_mq:
             # Might still need to check pod IPs.
             job_wrapper = job_state.job_wrapper
+            persisted_state = job_wrapper.get_state()
+            if persisted_state in model.Job.terminal_states + [model.Job.states.DELETED_NEW]:
+                log.debug("(%s) Watched job in terminal state, will stop monitoring: %s", job_state.job_id, persisted_state)
+                return None
             guest_ports = job_wrapper.guest_ports
```
jmchilton, line 268: Probably should verify there are ports we want to check before fetching new state from the database?

natefoo: If there aren't then there's no need to monitor at all, correct?

lib/galaxy/jobs/runners/pulsar.py
```diff
@@ -265,6 +265,10 @@ def check_watched_item(self, job_state):
         if self.use_mq:
             # Might still need to check pod IPs.
             job_wrapper = job_state.job_wrapper
+            persisted_state = job_wrapper.get_state()
+            if persisted_state in model.Job.terminal_states + [model.Job.states.DELETED_NEW]:
```
jmchilton: Should we bail if the state isn't running I wonder...

lib/galaxy/jobs/runners/pulsar.py
jmchilton: I think so. I guess there will be after we merge your other PR though.

jmchilton (approved):

lib/galaxy/jobs/runners/pulsar.py
natefoo: Well it does need to stay in the monitor queue until it's running, but yeah it doesn't need to start checking ports until it's running.

natefoo:
@jmchilton I think the latest commit takes care of both your comments. It'll now only attempt to set ports once running and if it gets back the job's IP address (and this should always exist by the time the job state is updated to running) then it sets the ports and stops monitoring.

PR #10575, opened 2020-10-28
[20.09] Various fixes for conditional dependencies handling.

jmchilton:
- Fix missing tests.
- Fix reading conditional dependencies from YAML based objectstore config.
- Fix reading conditional dependencies for file sources.

### DIFF
--- a/lib/galaxy/dependencies/__init__.py
+++ b/lib/galaxy/dependencies/__init__.py
@@ -24,7 +24,6 @@ from galaxy.util.properties import (
 class ConditionalDependencies:
     def __init__(self, config_file, config=None):
         self.config_file = config_file
-        self.config = config
         self.job_runners = []
         self.authenticators = []
         self.object_stores = []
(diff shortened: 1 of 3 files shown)


lib/galaxy/dependencies/__init__.py
```diff
@@ -22,21 +22,23 @@
 
 
 class ConditionalDependencies:
-    def __init__(self, config_file):
+    def __init__(self, config_file, config=None):
         self.config_file = config_file
-        self.config = None
+        self.config = config
```
nsoranzo:
```suggestion
```

lib/galaxy/dependencies/__init__.py
```diff
@@ -22,21 +22,23 @@
...
         self.job_runners = []
         self.authenticators = []
         self.object_stores = []
+        self.file_sources = []
         self.conditional_reqs = []
         self.container_interface_types = []
         self.job_rule_modules = []
         self.error_report_modules = []
+        if config is None:
+            self.config = load_app_properties(config_file=self.config_file)
```
nsoranzo, line 36:
```suggestion
            self.config = load_app_properties(config_file=self.config_file)
        else:
            self.config = config
```

mvdbeek (approved):

Issue #10553, opened 2020-10-26
Some UX issues on 20.09


bebatut:
I did not use Galaxy for a while but today I have seen several issues in Galaxy 

On History panel:

- Static text for collection when scrolling

  ![galaxy_history_bug](https://user-images.githubusercontent.com/1842467/97189021-7500e380-17a4-11eb-9a26-18685cb87506.gif)

On Multiple panel

- Strange stuff when scrolling

  ![galaxy_multiple_history_ux_bug](https://user-images.githubusercontent.com/1842467/97189365-cc06b880-17a4-11eb-9c41-a5438021175d.gif)

- Duplication of a dataset (number 18 in the second history)

  ![Screenshot 2020-10-26 at 11 29 03](https://user-images.githubusercontent.com/1842467/97190577-2bb19380-17a6-11eb-8d45-4bd345396ed6.png)

---

Details

**Server**: usegalaxy.eu
**Browser**: Firefox 81.0.2

mvdbeek:
The masthead also seems to be able to scroll, but if you do that in the multi-history view you end up with this:
<img width="2227" alt="Screenshot 2020-10-26 at 17 25 04" src="https://user-images.githubusercontent.com/6804901/97199589-759f7700-17b0-11eb-81b5-7a58fc0f4c3c.png">

wm75:
Same scrolling issues with Chromium.

mvdbeek:
@guerler @OlegZharkov @dannon can I ask one of you to take a look at this ? Would be nice to ship the release without these bugs.

mvdbeek:
> * Duplication of a dataset (number 18 in the second history)

that's not necessarily a bug, if you can share the history we can look at it in more detail. the rest should be fixed.

PR #10579, opened 2020-10-29
[20.09] Fix progress bar floating text

dannon:
This fixes the collection summary display mentioned in #10553 

Bonus bugfix of a random dangling `}` that flashed in default text while status was loading.


mvdbeek:
This has fixed the floating progressbar text ... and it seems to have also fixed https://github.com/galaxyproject/galaxy/issues/10553#issuecomment-716665095 (can see it on the release_20.09 branch, but not here). Does that make sense ?

PR #10556, opened 2020-10-27
[20.09] allow adding dataset to library from history by default

OlegZharkov:
After gitter discussion with @innovate-invent, we discovered that we don't allow importing datasets from history, if `user_library_import_dir`, `allow_library_path_paste` and `library_import_dir` are not defined. This PR fixes this issue.

thanks a lot to @innovate-invent for reporting!

P.S. is 20.09 going to be merged back to dev?

PR #10582, opened 2020-10-29
[20.09] Resolve LDDA <-> Library Dataset circular dependency

mvdbeek:
This eliminates a lot of excessive flushes that were needed to persist new `LibraryDatasetDatasetAssociation` (LDDA) and `LibraryDataset` (LD) objects. These intermittent flushes would make LD objects visible before they are associated with a LDDA, causing the Exception in https://github.com/galaxyproject/galaxy/pull/10556#issuecomment-717175641

```
galaxy.web.framework.decorators ERROR 2020-10-27 17:26:25,817 Uncaught exception in exposed API method:
  File "/Users/mvandenb/src/galaxy/lib/galaxy/web/framework/decorators.py", line 294, in decorator
  File "/Users/mvandenb/src/galaxy/lib/galaxy/webapps/galaxy/api/folder_contents.py", line 111, in index
    dataset = content_item.library_dataset_dataset_association.dataset
AttributeError: 'NoneType' object has no attribute 'dataset'
```

The trick here is to use `post_update=True`, so that ldda and ld rows
are INSERTED, followed by an UPDATE to create the link.
This is explained at https://docs.sqlalchemy.org/en/13/orm/relationship_persistence.html.

This is also the groundwork for speeding up library additions by batching flushes.

### DIFF
--- a/test/unit/data/test_galaxy_mapping.py
+++ b/test/unit/data/test_galaxy_mapping.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import collections
 import unittest
 import uuid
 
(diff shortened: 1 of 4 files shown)


test/unit/data/test_galaxy_mapping.py
```diff
@@ -154,6 +155,36 @@ def assert_display_name_converts_to_unicode(item, name):
...
+        trans = collections.namedtuple('trans', 'user')
+        target_folder = self.model.LibraryFolder(name='library_folder')
+        ldda = hda.to_library_dataset_dataset_association(
+            trans=trans(user=u),
+            target_folder=target_folder,
+        )
+        assert target_folder.item_count == 1
+        assert ldda.id
+        assert ldda.library_dataset.id
+        assert ldda.library_dataset_id
```
nsoranzo, line 171:
```suggestion
```

mvdbeek: I've done this intentionally. One of my earlier attempts was a backref only that wouldn't populate `library_dataset_id `

nsoranzo: My eyes saw 2 duplicate lines, sorry!

mvdbeek: That has to be a first time for 🦅 👀 ! 😆

mvdbeek:
Looks like the selenium test failure is legitimate, looking at this now

PR #10584, opened 2020-10-29
[20.09] Fix missing section values on tool re-run

mvdbeek:
https://github.com/galaxyproject/galaxy/pull/10083 optimized rendering of conditionals and sections so that rendering only happens when conditionals and sections are expanded. On job re-run unexpanded sections may contain non-default values. When submitting the job without expanding the section we'd lose those values and the default value would be submitted.

This reverses the optimization for sections and hidden conditional `</when>` items.

### DIFF
--- a/client/src/mvc/form/form-section.js
+++ b/client/src/mvc/form/form-section.js
@@ -89,9 +89,11 @@ var View = Backbone.View.extend({
                             break;
                         }
                     }
-                    if (i == selectedCase && nonhidden) {
+                    if (i == selectedCase) {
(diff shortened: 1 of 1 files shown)


mvdbeek:
Well, this brings @hexylena's cirocs test case in https://github.com/galaxyproject/galaxy/issues/9615 back to ~ 7 seconds with profiling. Better than the starting situation and better than not using the same parameters, but ... 😢

mvdbeek:
One consolation might be that if we separate model manipulations from rendering, which I think is what we should do if/when we vue-fy tool form parameters, we'll also get a nice performance boost.

PR #10590, opened 2020-10-30
Add API endpoint for current history

simonbray:
Another paper cut - the bioblend method `get_current_history` somewhat misleadingly points to the `most_recently_used` endpoint. It should point at this instead (and optionally a new `get_most_recently_used_history` method could be added as well).


simonbray:
Sorry, correction: `get_most_recently_used_history` exists already and does the same thing as `get_current_history`.

simonbray:
I was informed that the current history is actually stored in a cookie so this PR doesn't make much sense.

Issue #10587, opened 2020-10-29
Replace tool panel sections with EDAM


hexylena:
Since I proposed this in the round table, figured I'd document my vision here a bit.

**Background**

We need a solution for the tool panel sections. They're non-uniform, they're hard to categorise properly, everyone does their own thing, it's a mess. Every admin needs to decide for themselves where the tool should go and they aren't always the best people to make that decision (speaking for myself here, I know nothing about where some rna tools go.)

**Solution**

Burn the tool_conf, shed_tool_conf, and integrated tool panel to the ground. Instead, we rely primarily on the EDAM terms sourced from the tools. For every EDAM term, users will see a "section" in Galaxy. We can natively nest these based on the EDAM hierarchy.

Now, we just have: the list of tools installed (shed database, maybe a list of tools in a tooL_conf like file.) And we have the EDAM terms.

**Possible Issues**

1. Tools aren't annotated with EDAM. Great, we should fix this. Throwing anything without tags in an "UNCATEGORISED" section at the end will be good motivation.
2. Old tools aren't annotated. Let's update them. Tools generally won't change EDAM types over time, so if any of the versions in the lineage are annotated with some tags, then we can take those.
3. Admins want their own sections? @jmchilton proposed that we permit additional ontologies that can be used, and admins can tag these tools (either in the tool, or outside in the database?)

**Benefits**

- No more shed_tool_conf or ITP. @natefoo, @Slugger70 and other admins everywhere rejoice :tada:
- Better EDAM annotation of existing tools
- Hierarchical tool panel
- Users can find tools by well annotated "thing I want to do", and can find them across every applicable section.

hexylena:
Starting places for work:

Review these!

- [ ] galaxyproject/tools-iuc#1905
- [ ] galaxyproject/tools-iuc#1897

Implement this:

- [ ] expose edam in toolbox
- [ ] new toolbox based around edam.

bgruening:
ping @hmenager @matuskalas

hexylena:
ping @inkuzmin

innovate-invent:
I would caution against having tools listed within the edam tree. It could get annoying to find deeply nested tools.
Having a edam tree view adjacent to a flat tool list, which is then filtered by the selection in the tree would be better.
It also provides the opportunity to collapse the tree view and have a simple search box at the top of the tool list.

hexylena:
[...]

PR #10592, opened 2020-10-30
EDAM tool panel

hexylena:
part of #10587. This was honestly pretty easy and it looks great? (Minus the hundreds of unannotated tools.)

image

<details>
  <summary>Original Notes</summary>
  

I just took the existing ITP loading code and replaced it with edam generated topics rather than whatever section they requested.

Sorting based on EDAM would be nice (edit: done!) E.g. the following image:

</details>

Bugs:

- ~we have a copy of the EDAM ontology in the codebase. Surely there's a better way to handle this?~
- Binary server setting, doesn't permit some of the fancy things people wanted in #10587
- Search works but "show sections" in search is now broken
- ~All the tests fail, even though I made this opt-in?~
- ~Sorting based on EDAM ordering would probably be better.~


hexylena:
Went ahead an annotated some of my Circos tools and it's fantastic, I love not having to read through the sections to try and figure out where something should go, it's just automatic.

image

mvdbeek:
> * All the tests fail, even though I made this opt-in?

It's just the unit tests that mock out a config object in various places. Want me to fix that ?

lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -2323,6 +2323,22 @@ mapping:
...
+          sections in lieu of EDAM topics only.
+
+      beta_edam_toolbox_ontology_path:
+        type: str
+        default: EDAM_1.23.tsv
+        path_resolves_to: config_dir
+        required: false
+        desc: |
+          Sets the path to EDAM ontology file.
+
```
mvdbeek:
```suggestion
      beta_edam_toolbox_ontology_path:
        type: str
        default: EDAM_1.23.tsv
        path_resolves_to: data_dir
        required: false
        desc: |
          Sets the path to EDAM ontology file.

      beta_edam_ontology_url:
        type: str
        default: <url_to_wherever we can fetch the ontology>
        required: false
        desc: |
          Sets the url to the EDAM ontology file.
```
And then we can fetch that on startup if it's not present.

hexylena:
> It's just the unit tests that mock out a config object in various places. Want me to fix that ?

Sure, that'd be ace!

lib/galaxy/webapps/galaxy/config_schema.yml
hexylena: Ah, that's a nice suggestion, sure. sounds good.

hexylena: Just to confirm because the diff looks odd: swap path for url, right? remove path completely?

mvdbeek:
> * we have a copy of the EDAM ontology in the codebase. Surely there's a better way to handle this?

Downloading this file now from beta_edam_ontology_url. Can you rebase out config/EDAM_1.23.tsv  ?

I just gave it a go, it looks nice!

lib/galaxy/webapps/galaxy/config_schema.yml
mvdbeek: I think we want to keep both. If the path doesn't exist we download and store the file at the path.

nsoranzo: Why do we want the URL to be configurable? Can't we just download the latest version from http://edamontology.org/EDAM.tsv ?

hexylena: In case EDAM releases a new ontology and people want to update before we update galaxy?

nsoranzo: That URL is supposed to always point to the latest EDAM release.

hexylena: Oh, sorry, misread the URL. Then I guess we need to worry about caching a bit, or? People will have the edam from whenever they first try this setting, and it won't be updated?

mvdbeek: I thought there might be other places / custom ontologies people may want to load. If that's not true we can just make the URL a constant, sure.

nsoranzo: > Oh, sorry, misread the URL. Then I guess we need to worry about caching a bit, or? People will have the edam from whenever they first try this setting, and it won't be updated?

Sure, we have a similar issue with the conda binary.

> I thought there might be other places / custom ontologies people may want to load. If that's not true we can just make the URL a constant, sure.

Given we are using the EDAM annotations in tools, I don't think we need to make that configurable at this stage.

hexylena:
I've added section labels, and ordered all of the contents according to the alphabetically sorted. I thought it would be better, but, not sure how I feel about it in the end. 

image

mvdbeek:
I like that, it is a consistent way to read the panel.

hexylena:
Hid labels without children, ok, I definitely like that too.

image

lib/galaxy/tools/toolbox/base.py
```diff
@@ -823,7 +1013,10 @@ def quick_load(tool_file, async_load=True):
                 integrated_elems[key] = tool
 
                 if async_load:
-                    self._load_tool_panel()
+                    if self.app.config.enable_beta_edam_toolbox:
```
jmchilton, line 1024: Can we instead make this an API parameter and have this more of a dynamic view of the toolbox?

hexylena: That definitely sounds like a better API. Would you have any suggestions for that refactoring? (It's been a while since I've done dev type things) Doing both the edam version and non-edam version parsing, and building two internal toolboxes?

hexylena: Just following up here @jmchilton in case you had any opinions on the implementation :)

dannon:
I haven't tried this yet, but it aligns a lot with what @juleengraham has been working towards to support multiple user/admin/upstream defined toolsets in the tool panel.  Have you followed that work?  (I'm also out of date on the status of that, and haven't tried this yet, but I'm trying to catch up and wanted to comment before I forget!)

hexylena:
> Have you followed that work?

That sounds fantastic, I have not heard about it, or followed it at all. Is there a tracking issue somewhere or so?

juleengraham:
@hexylena Dannon is referring to PR #9996. I actually brought up this EDAM work in our JHU meeting last week and briefly started a discussion about how the two might be related. At the time we weren't really sure what EDAM tags were. I'm sure we're going to talk about it more later today. I know in your original issue, you mention admins wanting to have their own lists and I think that would be a great way to connect these two projects.

Correct me if I'm wrong but EDAM tags seem to be a part of a tool's configuration (so specified by the tool's creator/maintainer?), but more standardised because EDAM comes from a biological background/definition.

hexylena:
@juleengraham ah neat, I hadn't seen that. Admins having their own organisations of the tool panel was definitely discussed in the original, but I think it was potentially a bit orthogonal to the goals of an EDAM view? (i.e. I didn't have bandwidth to tackle that problem.)

> Correct me if I'm wrong but EDAM tags seem to be a part of a tool's configuration (so specified by the tool's creator/maintainer?), but more standardised because EDAM comes from a biological background/definition.

This is exactly correct. Tool developers tag their tools with the functions and topics of the tool. It's not an admin or user controlled feature. But it could definitely fit into the "one of many" views of a tool panel, like it seems you've been tasked with!

juleengraham:
>> This is exactly correct. Tool developers tag their tools with the functions and topics of the tool. It's not an admin or user controlled feature. But it could definitely fit into the "one of many" views of a tool panel, like it seems you've been tasked with!

@hexylena noted! The toolsets weren't necessarily intended to be based on EDAM tags but I could see that being a use case. I started out working based on the EU's pre-curated sets (https://github.com/usegalaxy-eu/usegalaxy-eu-tools) but I've changed the implementation a bit in hopes that it'll be more user-friendly for the admins. I think for now, the lists were meant to be created by the admin of the instance and no individual users

mvdbeek:
As an optional / dev setup this looks fine to me and certainly is a great incentive for more EDAM. Would you be opposed to merging as is (after resolving conflicts) @jmchilton?

mvdbeek (approved):
Can you resolve conflicts @hexylena ?

hexylena:
Yes, absolutely @mvdbeek

hexylena:
Resolved.

jmchilton (approved):
Approve after linting and tests are passed. Sorry for dropping the ball on the review of this. I'd like to take a pass at making it a bit more dynamic but this is a cool step forward. Thanks so much!

hexylena:
No worries. would love for it to be more dynamic too. I look forward to our
multi view toolbox future!

John Chilton <notifications@github.com> schrieb am Do., 7. Jan. 2021, 23:20:

mvdbeek:
@ic4f I'm sure this is an easy one for you, what did we need to do for this unit test failure?
```
____________ test_config_defaults[beta_edam_toolbox_ontology_path] _____________

test_data = TestData(key='beta_edam_toolbox_ontology_path', expected='EDAM.tsv', loaded='/home/circleci/repo/database/EDAM.tsv')

    @pytest.mark.parametrize('test_data', get_config_data(), ids=get_key)
    def test_config_defaults(test_data):
>       assert test_data.expected == test_data.loaded
E       AssertionError: assert 'EDAM.tsv' == '/home/circle...base/EDAM.tsv'
E         - /home/circleci/repo/database/EDAM.tsv
E         + EDAM.tsv

test/unit/config/test_config_values.py:227: AssertionError
```

ic4f:
I'm looking into this - give me a few minutes :-)

ic4f:
@hexylena You need to add the following to this dict: https://github.com/galaxyproject/galaxy/blob/release_20.09/test/unit/config/test_config_values.py#L100
```
'beta_edam_toolbox_ontology_path': self._in_data_dir('EDAM.tsv'), 
```
It tells the test how to prepare the expected value: it places the 'EDAM.tsv' file inside the `data_dir` (which is created at test runtime).

Sorry for the lacking docs (it's all in the comments in that test file).

dannon:
@mvdbeek Yeah, no opposition here.

mvdbeek:
> Sorry for the lacking docs (it's all in the comments in that test file).

The test could maybe tell us that, so if it fails we know what to do ?

ic4f:
@mvdbeek yes (a message can be passed to the test with the rest of the data for cases where such a message makes sense). However, the path generation for the test can be automated: we have the key, we have the schema which contains the `path_resolves_to` value. I think that would be the better way to do it: then we don't have to require this extra step from anyone who's editing a path option in the schema. (writing accompanying tests is good; having to modify a nontrivial testing module for trivial things like a new config option - less so, I think).

PR #10594, opened 2020-10-31
pgcleanup can redirect logging to stdout

gmauro:
this PR allows gxadmin to properly save logging details
https://github.com/usegalaxy-eu/gxadmin/blob/d3e98865bd9e3c509e3144b2f9974b03d41881e2/parts/25-galaxy.sh#L31

### DIFF
--- a/lib/galaxy/util/script.py
+++ b/lib/galaxy/util/script.py
@@ -90,7 +90,7 @@ def _arg_parser():
     return parser
 
 
-def set_log_handler(filename=None, stream=sys.stderr):
+def set_log_handler(filename=None, stream=None):
     if filename:
         handler = logging.FileHandler(filename)
     else:
--- a/lib/galaxy/util/script.py
+++ b/lib/galaxy/util/script.py
@@ -1,6 +1,7 @@
 """Utilities for Galaxy scripts
 """
 import argparse
+import logging
 import os
 import sys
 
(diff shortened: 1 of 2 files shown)


jmchilton:
What do you think about a general log file command-line argument that can be sent "-" to push that to standard output? That seems a bit more standard to me - but I'm not certain.

gmauro:
I am not sure about getting your point. Do you mean something like replacing the log-dir option with a log-file option and if log-file option is not set, the script send logs to stdout?

scripts/cleanup_datasets/pgcleanup.py
```diff
@@ -25,7 +25,7 @@
 import galaxy.config
 from galaxy.exceptions import ObjectNotFound
 from galaxy.objectstore import build_object_store_from_config
-from galaxy.util.script import app_properties_from_args, populate_config_args
+from galaxy.util.script import app_properties_from_args, populate_config_args, set_log_handler
```
nsoranzo:
```suggestion
from galaxy.util.script import (
    app_properties_from_args,
    populate_config_args,
    set_log_handler,
)
```

lib/galaxy/util/script.py
```diff
@@ -87,3 +88,11 @@ def _arg_parser():
     for argument in ARGUMENTS:
         parser.add_argument(*argument[0], **argument[1])
     return parser
+
+
+def set_log_handler(filename=None, stream=sys.stderr):
```
nsoranzo:
```suggestion
def set_log_handler(filename=None, stream=None):
```

gmauro:
thanks @nsoranzo

nsoranzo (approved):

mvdbeek:
Thank you @gmauro!

PR #10602, opened 2020-11-01
[20.09] Mark test cases that use data tables or loc files

mvdbeek:
This does not work if a tool data table entry is selected without
explicitly listing it in the test data setup, but I think that's
a rare case (and that can be fixed now by referencing it explicitly).

xref: https://github.com/galaxyproject/galaxy/issues/7998#issue-445473778

### DIFF
--- a/lib/galaxy/tools/test.py
+++ b/lib/galaxy/tools/test.py
@@ -58,7 +58,7 @@ def description_from_tool_object(tool, test_index, raw_test_dict):
             "expect_failure": raw_test_dict.get("expect_failure", False),
             "required_files": required_files,
             "required_data_tables": required_data_tables,
-            "required_loc_files":  required_loc_files,
+            "required_loc_files": required_loc_files,
             "tool_id": tool.id,
             "tool_version": tool.version,
             "test_index": test_index,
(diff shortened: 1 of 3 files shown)


lib/galaxy/tools/test.py
```diff
@@ -149,26 +153,30 @@ def _process_raw_inputs(tool, tool_inputs, raw_inputs, required_files, parent_co
...
                 else:
-                    processed_value = _process_simple_value(value, param_value)
+                    processed_value = _process_simple_value(value, param_value, required_data_tables, required_loc_files)
                 expanded_inputs[context.for_state()] = processed_value
     return expanded_inputs
 
 
-def _process_simple_value(param, param_value):
-    if isinstance(param, galaxy.tools.parameters.basic.SelectToolParameter) and hasattr(param, 'static_options'):
+def _process_simple_value(param, param_value, required_data_tables, required_loc_files, ):
```
nsoranzo:
```suggestion
def _process_simple_value(param, param_value, required_data_tables, required_loc_files):
```

jmchilton (approved):

lib/galaxy/tools/test.py
```diff
@@ -74,7 +78,7 @@ def description_from_tool_object(tool, test_index, raw_test_dict):
     return ToolTestDescription(processed_test_dict)
 
 
-def _process_raw_inputs(tool, tool_inputs, raw_inputs, required_files, parent_context=None):
+def _process_raw_inputs(tool, tool_inputs, raw_inputs, required_files, required_data_tables, required_loc_files, parent_context=None):
```
jmchilton, line 81: Someday we need to get this talking to the API and compute less on the backend, I've always really hated that this is using the tool object. One should be able to turn just the list of parameters into an API request without needing Galaxy internals... but I think that is a fight for another day. This is important and we need to just get it done and this is really cool.

Issue #10604, opened 2020-11-01
[20.09] psycopg2.errors.NotNullViolation null value in column "dataset_collection_id" violates not-null constraint


bgruening:
Seen in our logs with 20.09.

```
Nov 01 14:58:58 sn04.bi.uni-freiburg.de python[1827028]: galaxy.jobs DEBUG 2020-11-01 14:58:58,991 (12059080) Working directory for job is: /data/dnb03/galaxy_db/job_working_directory/012/059/12059080
Nov 01 14:58:58 sn04.bi.uni-freiburg.de python[1847375]: galaxy.jobs DEBUG 2020-11-01 14:58:58,998 (12059086) Working directory for job is: /data/dnb03/galaxy_db/job_working_directory/012/059/12059086
Nov 01 14:58:59 sn04.bi.uni-freiburg.de python[1847375]: galaxy.objectstore DEBUG 2020-11-01 14:58:59,011 Using preferred backend 'files10' for creation of Dataset 27240484
Nov 01 14:58:59 sn04.bi.uni-freiburg.de python[1819184]: galaxy.job_execution.output_collect ERROR 2020-11-01 14:58:59,036 Problem gathering output collection.
Nov 01 14:58:59 sn04.bi.uni-freiburg.de python[1819184]: Traceback (most recent call last):
Nov 01 14:58:59 sn04.bi.uni-freiburg.de python[1819184]: File "/opt/galaxy/venv/lib64/python3.6/site-packages/sqlalchemy/engine/base.py", line 1284, in _execute_context
Nov 01 14:58:59 sn04.bi.uni-freiburg.de python[1819184]: cursor, statement, parameters, context
Nov 01 14:58:59 sn04.bi.uni-freiburg.de python[1819184]: File "/opt/galaxy/venv/lib64/python3.6/site-packages/sqlalchemy/engine/default.py", line 590, in do_execute
Nov 01 14:58:59 sn04.bi.uni-freiburg.de python[1819184]: cursor.execute(statement, parameters)
Nov 01 14:58:59 sn04.bi.uni-freiburg.de python[1819184]: psycopg2.errors.NotNullViolation: null value in column "dataset_collection_id" violates not-null constraint
Nov 01 14:58:59 sn04.bi.uni-freiburg.de python[1819184]: DETAIL:  Failing row contains (13330967, null, 30637828, null, null, 0, ERR4597396__single).
[...]

PR #10606, opened 2020-11-01
[20.09] Fix IntegrityError when flushing DC/DCE relationship

mvdbeek:
by setting the Foreign Key on left side of one to many DCE/DC relation

Should fix https://github.com/galaxyproject/galaxy/issues/10604


mvdbeek:
I should pretend this was just to figure out how to reproduce https://github.com/galaxyproject/galaxy/issues/10604 🔥

jmchilton:
Awesome, thanks for the fix!

PR #10608, opened 2020-11-02
Remove IUC conda channel from default configuration

bernt-matthias:
I suggest to drop the IUC conda channel. My initial motivation an error caused by the emboss package in this channel:

```
ClobberError: This transaction has incompatible packages due to a shared path.
  packages: iuc::emboss-5.0.0-0, conda-forge::libpng-1.6.37-h21135ba_2
  path: 'share/man/man5/png.5'
```

But I guess its not needed anymore.

### DIFF
--- a/doc/source/admin/galaxy_options.rst
+++ b/doc/source/admin/galaxy_options.rst
@@ -433,7 +433,7 @@
 :Description:
     conda channels to enable by default
     (https://conda.io/docs/user-guide/tasks/manage-channels.html)
-:Default: ``iuc,conda-forge,bioconda,defaults``
+:Default: ``conda-forge,bioconda,defaults``
 :Type: str
 
 
--- a/lib/galaxy/config/sample/galaxy.yml.sample
+++ b/lib/galaxy/config/sample/galaxy.yml.sample
@@ -316,7 +316,7 @@ galaxy:
 
   # conda channels to enable by default
   # (https://conda.io/docs/user-guide/tasks/manage-channels.html)
-  #conda_ensure_channels: iuc,conda-forge,bioconda,defaults
+  #conda_ensure_channels: conda-forge,bioconda,defaults
 
   # Use locally-built conda packages.
   #conda_use_local: false
--- a/lib/galaxy/tool_util/deps/resolvers/conda.py
+++ b/lib/galaxy/tool_util/deps/resolvers/conda.py
@@ -39,9 +39,8 @@ from ..conda_util import (
 DEFAULT_BASE_PATH_DIRECTORY = "_conda"
 DEFAULT_CONDARC_OVERRIDE = "_condarc"
 # Conda channel order from highest to lowest, following the one used in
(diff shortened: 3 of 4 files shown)


nsoranzo:
Can you please run `make config-rebuild` and commit the changes it makes?

nsoranzo (approved):

bernt-matthias:
> Can you please run `make config-rebuild` and commit the changes it makes?

Done. 

Wondering if building (mulled) containers needs to be adapted accordingly?

bgruening:
The reason we added this channel is to have a fallback for anything that is too `ugly` to get into bioconda. Do we still have this fear? If so we should fix/update the EMBOSS package instead of removing the channel.

bernt-matthias:
I understand .. seems reasonable. But with the current channel order its not a fallback, but the default -- isn't it? 

Maybe a bit more background. I was trying to install rnaquast (https://github.com/bioconda/bioconda-recipes/blob/master/recipes/rnaquast/meta.yaml) and for some reason it used emboss 5.0.0. from the IUC channel instead of a newer one from bioconda. So I was thinking that the IUC package causes the problems.

- EMBOSS 5.0.0 its already in bioconda https://github.com/bioconda/bioconda-recipes/tree/master/recipes/emboss/5.0.0 and the recipe seems more or less identical
- ucsc tools seems to be a meta package .. 

- pglite seems not in conda except for IUC channel

So I see that removing the cannel seems not the best idea at the moment. Maybe a different channel priority would be an idea. `channel-priority: disabled` (i.e. --no-channel-priority) would be an option since then IUC packages are only considered in case of ties.

bgruening:
We could remove the EMBOSS package from the IUC channel, without changing anything?
I think our thinking to include iuc at the front was so that we can overwrite packages from other channels if really needed. But again we tried to be prepared for a case that never occured. So if you all want to drop it, I'm ok with it.

bernt-matthias:
> We could remove the EMBOSS package from the IUC channel, without changing anything?

I think this would help for my problem. But to be 100% sure one would need to test he emboss tools with the bioconda package.

> I think our thinking to include iuc at the front was so that we can overwrite packages from other channels if really needed. But again we tried to be prepared for a case that never occured. So if you all want to drop it, I'm ok with it.

Since I'm not sure about the other packages .. the safer change would be to change the channel priority... or?

```
channel_priority (ChannelPriority)
Accepts values of 'strict', 'flexible', and 'disabled'. The default
value is 'flexible'. With strict channel priority, packages in lower
priority channels are not considered if a package with the same name
appears in a higher priority channel. With flexible channel priority,
the solver may reach into lower priority channels to fulfill
dependencies, rather than raising an unsatisfiable error. With channel
priority disabled, package version takes precedence, and the
configured priority of channels is used only to break ties. ...
```

mvdbeek:
It's easy enough to add the channel back with a lower priority, should the need arise.

mvdbeek (approved):

PR #10616, opened 2020-11-02
[20.09] fix  trs_id and trs_server query strings

OlegZharkov:
Fix query string during TRS import
Example link: http://localhost:8080/workflows/trs_import?trs_server=workflowhub.eu&trs_id=2


OlegZharkov:
thanks @jmchilton!

PR #10617, opened 2020-11-02
[20.09] Fix file_path bug; change default assignment of file_path and id/uuid

### LINKED ISSUE #10393: [20.09] File download seems non-functional
I deployed Galaxy 20.09 to my playground environment today, however downloading of files from Galaxy doesn't seem to work.

Hereby the logs that I got:
```
IP - - [09/Oct/2020:16:21:59 +0200] "GET /datasets/DATASETSTRING/display?to_ext=txt HTTP/1.1" 500 - "https://GLXURL/"
  File "lib/galaxy/web/framework/middleware/error.py", line 153, in __call__
    app_iter = self.application(environ, sr_checker)
  File "GLXPATH/.venv/lib/python3.7/site-packages/paste/recursive.py", line 85, in __call__
    return self.application(environ, start_response)
  File "GLXPATH/.venv/lib/python3.7/site-packages/paste/httpexceptions.py", line 640, in __call__
    return self.application(environ, start_response)
  File "lib/galaxy/web/framework/base.py", line 141, in __call__
    return self.handle_request(environ, start_response)
  File "lib/galaxy/web/framework/base.py", line 220, in handle_request
    body = method(trans, **kwargs)
  File "lib/galaxy/webapps/galaxy/controllers/dataset.py", line 186, in display
    return data.datatype.display_data(trans, data, preview, filename, to_ext, offset=offset, ck_size=ck_size, **kwd)
  File "lib/galaxy/datatypes/data.py", line 440, in display_data
    trans.response.headers['Content-Length'] = int(os.stat(data.file_name).st_size)
FileNotFoundError: [Errno 2] No such file or directory: ''

  File "lib/galaxy/web/framework/middleware/batch.py", line 80, in __call__
    return self.application(environ, start_response)
  File "lib/galaxy/web/framework/middleware/request_id.py", line 15, in __call__
    return self.app(environ, start_response)
  File "lib/galaxy/web/framework/middleware/xforwardedhost.py", line 23, in __call__
    return self.app(environ, start_response)
  File "lib/galaxy/web/framework/middleware/translogger.py", line 70, in __call__
    return self.application(environ, replacement_start_response)
  File "lib/galaxy/web/framework/middleware/error.py", line 163, in __call__
    exc_info)
  File "lib/galaxy/web/framework/middleware/translogger.py", line 69, in replacement_start_response
    return start_response(status, headers, exc_info)
SystemError: <built-in function uwsgi_spit> returned a result with an error set

```

jdavcs:
This is ready for review.
1. Change how `file_path` default is set .
3. Change how `uuid`/`id` values are set for `object_store_store_by`.
4. Do not set default value of `object_store_store_by` outside of config module.
5. Add unit test for all of the above.

Fixes #10393, also see discussion starting here: https://github.com/galaxyproject/galaxy/pull/9197#commitcomment-43652671

This may or may not be more than a bug fix. I'm open to splitting it into 20.09 (essential bug fixes) and dev (handling of id/uuid, which I still think is a preventative measure).

Most importantly, is all this logic correct? 
1. Is setting `file_path` default reasonable? (See [this discussion](https://github.com/galaxyproject/galaxy/pull/10617/files#r516241731))
2. Same for setting `object_store_store_by` default; see code block starting on line 879:
https://github.com/galaxyproject/galaxy/blob/526e8e04145cc643a4c6c9568b93719df8adb4cb/lib/galaxy/config/__init__.py#L879
So, instead of relying on a "magic" filename (`files` or `objects`) which may have been set by the user, we simply find a dataset and see whether it's stored by uuid or not - and assign the default accordingly. (the assumption here is that datasets will never be stored by id *and* by uuid under the same `file_path` directory)

Note: a dataset is located via depth-first search (not breadth-first), so there should be no performance issues for a directory structure that has files at the lower level(s), even with millions of directories (millions of *empty* directories on all levels would cause an issue, but such a directory structure is not a reasonable expectation)


ic4f:
@nsoranzo thanks for the title edit - and sorry, I keep forgetting about this format!

lib/galaxy/config/__init__.py
```diff
@@ -260,6 +260,17 @@ def select_one_or_all_paths_from_list():
     def _path_exists(self, path):  # factored out so we can mock it in tests
         return os.path.exists(path)
 
+    def _set_alt_paths(self, option, *alt_paths):
```
nsoranzo, line 263: Passing `alt_paths` as a list seems more natural to me, but I suppose it's personal preference.

lib/galaxy/config/__init__.py
```diff
@@ -260,6 +260,17 @@ def select_one_or_all_paths_from_list():
...
+        # If `option` is not set, check the default. If that path does NOT exist, check *alt_paths.
```
jmchilton: ``objects`` is the default, but I think if ``files`` is present it should be used even if ``objects`` exists. So I think this is backward, but I mean you could make an argument either way I think.

lib/galaxy/config/__init__.py
jdavcs: I considered it; I also considered checking `if type(alt_paths) is list` inside the function, but finally decided on `*alt_paths`. I think, typically, this function will be called with just one argument - so I put the burden of accommodating flexibility on the function, not the caller. But that's the only reasoning I have - I have no real preference here.

lib/galaxy/config/__init__.py
jdavcs: Here's how I thought about this. If `file_path` is not set and we are using the default, there can be only 2 possible values: `objects` and `files` (because anything else would have to be set by the user). Of all the 4 truth table cases (00, 01, 10, 11) only one is interesting: 11 - both exist. In this case, we can safely assume that both have been used (because that directory is created only when a dataset needs to be stored), even though either one could be empty (files deleted by user). If both were used *as defaults*, `files` is the older default, so we'll go with the one that had to be used more recently - `objects`. If either one was used *not* as a default, but set by the user, then it is *no longer set* (by definition), so we treat it as "outdated".

This is after numerous truth tables and head banging.. And I may be wrong. There can be still weird cases (e.g. had previously set to `objects`, then unset and fell back to using the default - `files` - but did not delete the `objects` directory; in this case we'd be pointing to the wrong directory). But for those cases, I think we should simply provide a bit more guidance in the description of `file_path` in `galaxy.yml.sample` and expect the user to have only one directory *if they intend to use it as the default as opposed to setting it*.

Also, I'm purposefully not including `object_store_store_by` in this logic - I think the value of that option may depend on `file_path`, but not the other way round (and even if it does, it should be handled separately).

ic4f:
Linting error is due to f-strings linted on Python 3.5. I can either change the formatting or update to 3.6.

mvdbeek:
@ic4f can you state what the actual bug is that this fixes ? What I gather from https://github.com/galaxyproject/galaxy/issues/10393#issuecomment-717774394 is that this was a problem in 20.05 already. 

> * Change how `uuid`/`id` values are set for `object_store_store_by`.
> * Do not set default value of `object_store_store_by` outside of config module.

These don't seem to be bugs, at least not in relation to https://github.com/galaxyproject/galaxy/issues/10393

mvdbeek:
Also, if we change the way `files_path` is resolved (again), we'll have to explain the upgrade procedure, which means either merging the diverging paths or setting up a distributed/hierarchical object store.

nsoranzo:
The issue is better explained in https://github.com/galaxyproject/galaxy/pull/9197#commitcomment-43669843 and following, if that doesn't help I can explain further, let me know.

nsoranzo:
> Linting error is due to f-strings linted on Python 3.5. I can either change the formatting or update to 3.6.

We still support Python 3.5 on 20.09, so you cannot use f-strings unfortunately.

lib/galaxy/config/__init__.py
```diff
@@ -861,6 +872,22 @@ def _process_config(self, kwargs):
                 'filters': ['stack']
             }
 
+    def _configure_dataset_storage(self):
+        # The default for `file_path` has changed in 20.09; we may need to fall back to the old default
```
```suggestion
        # The default for `file_path` has changed in 20.05; we may need to fall back to the old default
```

lib/galaxy/objectstore/__init__.py
```diff
@@ -322,7 +322,7 @@ def __init__(self, config, config_dict=None, **kwargs):
         if config_dict is None:
             config_dict = {}
         super().__init__(config=config, config_dict=config_dict, **kwargs)
-        self.store_by = config_dict.get("store_by", None) or getattr(config, "object_store_store_by", "id")
+        self.store_by = config_dict.get("store_by") or config.object_store_store_by
```
jmchilton: The objectstore module is meant to be used standalone, it doesn't have direct dependencies on the config module if you look at the package structure. I prefer the getattr for those reasons. This might not even be purely stylistic, I'm not sure Pulsar for instance sets up object_store_by on its config object equivalent.

jmchilton:
I think the logic in 20.05 was totally fine - it needed tests and more comments. But this whole thing is too complicated I think. find_dataset isn't something we want to run on every start and we don't want to put logic that breaks Galaxy abstractions like that in galaxy.util IMO.

Is this complexity because the logic in 20.05 was wrong/bad or is it because it is difficult to implement it with the config module changes?

lib/galaxy/config/__init__.py
```diff
@@ -861,6 +872,22 @@ def _process_config(self, kwargs):
...
+        self._set_alt_paths('file_path', self._in_data_dir('files'))  # this is called BEFORE guessing id/uuid
+        ID, UUID = 'id', 'uuid'
+        if self.is_set('object_store_store_by'):
+            assert self.object_store_store_by in [ID, UUID], f'Invalid value for object_store_store_by [{self.object_store_store_by}]'
```
```suggestion
            assert self.object_store_store_by in [ID, UUID], "Invalid value for object_store_store_by [{}]".format(self.object_store_store_by)
```

mvdbeek:
Can we just revert https://github.com/galaxyproject/galaxy/commit/d14c60aa5d54fcde5a251eb6d961179cd8666a76  ?

ic4f:
> Can we just revert [d14c60a](https://github.com/galaxyproject/galaxy/commit/d14c60aa5d54fcde5a251eb6d961179cd8666a76) ?

We can't: `[option]_set` has been replaced with `is_set([option])`; and the assignment on line 802 is part of the old config file processing that has been fundamentally changed (#10172). The first part of this PR solved this problem in a simple way, I think. (I'll respond to the other comments in a few minutes)

lib/galaxy/objectstore/__init__.py
jdavcs: Thank you for the explanation - I didn't know that. I'll undo this edit.

lib/galaxy/config/__init__.py
mvdbeek: > In this case, we can safely assume that both have been used (

I don't know why, but I have an empty `database/objects` dir. If you flip this around I think it would be OK (as a standalone PR).

lib/galaxy/config/__init__.py
jdavcs: Committing for now, but I may have to revert these more recent commits dealing with uuid/id

lib/galaxy/config/__init__.py
jdavcs: > I don't know why, but I have an empty `database/objects` dir. If you flip this around I think it would be OK (as a standalone PR).

I don't know why that happened either. Flipping around - do you mean checking `files` *before* `objects`? Also, I may have a simple case that breaks the current/old setup  - posting as a separate comment.

lib/galaxy/util/__init__.py
```diff
@@ -173,6 +173,24 @@ def directory_hash_id(id):
...
+    """
+    Perform DFS starting at path. Return the first occurrence of a dataset.
+    `path` should be a directory.
+    Return a DirEntry object or None.
+    """
+    if not os.path.exists(path) or not os.path.isdir(path):
+        return
+    stack = [path]
+    while stack:
+        with os.scandir(stack.pop()) as iter:
```
nsoranzo: `iter` is a standard library function, can you please use a different variable name?

test/unit/util/test_find_dataset.py
```diff
@@ -0,0 +1,24 @@
...
+import tempfile
+
+from galaxy.util import find_dataset
+
+
+def test_find_dataset_no_dir():
+    assert find_dataset('no-such-path') is None
+
+
+def test_find_dataset_no_file():
```
nsoranzo:
```suggestion
def test_find_dataset_no_dat_file():
```

test/unit/util/test_find_dataset.py
```diff
@@ -0,0 +1,24 @@
...
+    td = tempfile.TemporaryDirectory()
+    td1 = tempfile.TemporaryDirectory(dir=td.name)
+    tempfile.mkstemp(dir=td1.name, suffix='.not-dat')
+    assert find_dataset('no-such-path') is None
```
nsoranzo:
```suggestion
    with tempfile.TemporaryDirectory() as td_name:
        td1 = tempfile.TemporaryDirectory(dir=td_name)
        tempfile.mkstemp(suffix='.not-dat', dir=td1.name)
        assert find_dataset(td_name) is None
```

test/unit/util/test_find_dataset.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def test_find_dataset_exists():
+    td = tempfile.TemporaryDirectory()
+    td1 = tempfile.TemporaryDirectory(dir=td.name)
+    f = tempfile.mkstemp(dir=td1.name, suffix='.dat')
+
+    direntry = find_dataset(td.name)
+    assert direntry.name == os.path.basename(f[1])
```
nsoranzo:
```suggestion
    with tempfile.TemporaryDirectory() as td_name:
        td1 = tempfile.TemporaryDirectory(dir=td_name)
        f = tempfile.NamedTemporaryFile(suffix='.dat', dir=td1.name)

        direntry = find_dataset(td_name)
        assert direntry.name == os.path.basename(f.name)
```

ic4f:
> I think the logic in 20.05 was totally fine ...
> Is this complexity because the logic in 20.05 was wrong/bad or is it because it is difficult to implement it with the config module changes?

It's not difficult to implement in the current config module. `file_path` just needs to also check for the alternative `files` location (as per discussion above). The *other* part of the current logic may have an issue which I've tried to address. Both, the current version, and what's proposed here have the same logic up to a point: 
```
if store_by is None: 
  store_by = 'id'
  if file_path not set and endswith 'objects':
    store_by = 'uuid'
  ########### this is what I'm adding ############:
  else if file_path is set:
    find a dataset
    if dataset is uuid: store_by = 'uuid'; else: store_by = 'id'
```
In other words, **if the user has set**  `file_path`, I don't think we can make any conclusions about what that directory name implies about the value of `object_store_store_by` (which is not set by the user). Here's a case where the current logic breaks:
1. Start with a fresh clone of 20.09 and a default galaxy.yml.
2. Start galaxy, download a dataset, stop galaxy.
3. Uncomment `file_path` in galaxy.yml
4. Start galaxy, download another dataset.
At this point you have 2 mutually-exclusive datasets. The first one is accessible only when `file_path` is commented out, the second one - only when `file_path is uncommented. Each one causes a server error when you try to download it when it's not accessible.

The reason for the above error is that we assume that if `file_path` is set, we must be storing by id - which is not always correct. I think the approach I propose solves this problem.

If looking for a dataset is not a good way to do it, I think we still need to address this (using a different approach then). (although finding a dataset should be really cheap here: I estimate it's a very shallow tree, and dfs immediately heads for the bottom level - where the dataset files are). 

..And thanks for reviewing this! (and of course we can drop this approach if it still looks more complicated than it has to be and is not an efficient way to address this issue; of it it's a non-issue after all)

mvdbeek:
> In other words, **if the user has set** `file_path`, I don't think we can make any conclusions about what that directory name implies about the value of `object_store_store_by` (which is not set by the user).

That's OK. There were no issues about this, and this has been in Galaxy since 20.05.  I think we should just document this in the description of `file_path`. I don't think we should do a search as you're doing here on startup.

lib/galaxy/config/__init__.py
mvdbeek: Yes. What John is saying, if `files` exists, use it.

lib/galaxy/config/__init__.py
```diff
@@ -861,6 +872,22 @@ def _process_config(self, kwargs):
...
+            assert self.object_store_store_by in [ID, UUID], "Invalid value for object_store_store_by [{}]".format(self.object_store_store_by)
+        else:
+            self.object_store_store_by = ID
+            if not self.is_set('file_path'):  # choose based on default dir name
+                if os.path.basename(self.file_path) == 'objects':
+                    self.object_store_store_by = UUID
+            else:  # choose based on dataset filename format
+                filename = find_dataset(self.file_path)
+                if not filename or is_uuid(filename):  # dir has no datasets, or contains a uuid dataset
+                    self.object_store_store_by = UUID
```
nsoranzo: Can we resolve

> The reason for the above error is that we assume that if `file_path` is set, we must be storing by id - which is not always correct.

with:

```suggestion
        elif os.path.basename(self.file_path) == 'objects':
            self.object_store_store_by = UUID
        else:
            self.object_store_store_by = ID
```

test/unit/util/test_find_dataset.py
jdavcs: @nsoranzo thanks! (my assert stmt made no sense calling the function with that argument!)

test/unit/util/test_find_dataset.py
jdavcs: Thanks!

lib/galaxy/config/__init__.py
jdavcs: OK. We'll just document the assumptions about `objects` in the `file_path` description, as per @mvdbeek .  I'll update the docs. Committing this now; ~~I may be able to simplify the conditional logic here in a min..~~ (EDIT: it's already simplified!)

@mvdbeek : I'm switching the check for `files` before `objects` in a separate commit.

lib/galaxy/config/__init__.py
jdavcs: ^Done.

mvdbeek (approved):
This looks good to me now, I've tried @nsoranzo's upgrade dance in https://github.com/galaxyproject/galaxy/pull/9197#commitcomment-43707907 and that worked fine.

ic4f:
> This looks good to me now, I've tried @nsoranzo's upgrade dance in [#9197 (comment)](https://github.com/galaxyproject/galaxy/pull/9197#commitcomment-43707907) and that worked fine.

And I did my own comment/upload/uncomment song and dance https://github.com/galaxyproject/galaxy/pull/10617#issuecomment-721235053 - and it works too.

nsoranzo:
@ic4f I'm reworking a bit the unit tests, do you mind if I push a commit to your branch?

ic4f:
> @ic4f I'm reworking a bit the unit tests, do you mind if I push a commit to your branch?

@nsoranzo not at all, please do!

nsoranzo (approved):

test/unit/config/test_config_dataset_storage.py
```diff
@@ -12,31 +14,41 @@ def mock_config_file(monkeypatch):
...
     assert appconfig.object_store_store_by == 'uuid'
 
 
 def test_uuid_2(mock_config_file, monkeypatch):
-    # file_path dir name = `objects`, not set: default value is used
-    monkeypatch.setattr(BaseAppConfiguration, 'is_set', lambda a, b: False)
-    appconfig = config.GalaxyAppConfiguration(file_path='objects')
+    # object_store_store_by not set
+    # file_path not set, `files` dir doesn't exist
+    monkeypatch.setattr(BaseAppConfiguration, '_path_exists', lambda self, path: False)
```
jdavcs, line 34: This is a very good catch! At first I thought we don't need this: `files` doesn't exists on a clean clone - so the test will pass on github. But then if we run it locally and `files` happens to exist, it will fail, of course.

ic4f:
Thank you, everyone, for the thorough review! Merging now.

PR #10620, opened 2020-11-03
Enhancements for galaxy-tool-test script.

jmchilton:
- Try to build more abstractions to allow reuse with ephemeris.
- Enhance to allow parallelization and enhnaced reporting like ephemeris.
- Allow testing a whole tool panel.
- Fix bugs in imported ephemeris logic (skipped failed tests and skipped tests for in counts/reporting for instance).
- Unit tests for abstractions.

### DIFF
--- a/lib/galaxy/tool_util/verify/interactor.py
+++ b/lib/galaxy/tool_util/verify/interactor.py
@@ -466,7 +466,7 @@ class GalaxyInteractorApi:
         return output_data['id']
 
     def delete_history(self, history):
-        return None
+        self._delete(f"histories/{history}")
 
     def __job_ready(self, job_id, history_id=None):
         if job_id is None:
@@ -876,10 +876,12 @@ def verify_tool(tool_id,
                 tool_version=None,
                 quiet=False,
                 test_history=None,
+                no_history_cleanup=False,
                 force_path_paste=False,
                 maxseconds=DEFAULT_TOOL_TEST_WAIT,
                 tool_test_dicts=None,
                 client_test_config=None,
+                skip_with_reference_data=False,
                 skip_on_dynamic_param_errors=False):
     if resource_parameters is None:
         resource_parameters = {}
@@ -900,20 +902,34 @@ def verify_tool(tool_id,
         "test_index": test_index,
     }
     client_config = client_test_config.get_test_config(job_data)
+    skip_message = None
     if client_config is not None:
         job_data.update(client_config)
         skip_message = job_data.get("skip")
-        if skip_message:
-            job_data["status"] = "skip"
-            register_job_data(job_data)
-            return
+
+    if not skip_message and skip_with_reference_data:
+        required_data_tables = tool_test_dict.get("required_data_tables")
+        required_loc_files = tool_test_dict.get("required_loc_files")
+        # TODO: actually hit the API and see if these tables are available.
+        if required_data_tables:
+            skip_message = f"Skipping test because of required data tables ({required_data_tables})"
+        if required_loc_files:
+            skip_message = f"Skipping test because of required loc files ({required_loc_files})"
+
+    if skip_message:
+        job_data["status"] = "skip"
+        register_job_data(job_data)
+        return
 
     tool_test_dict.setdefault('maxseconds', maxseconds)
     testdef = ToolTestDescription(tool_test_dict)
     _handle_def_errors(testdef)
 
+    created_history = False
     if test_history is None:
-        test_history = galaxy_interactor.new_history()
+        created_history = True
+        history_name = f"Tool Test History for {tool_id}/{tool_version}-{test_index}"
+        test_history = galaxy_interactor.new_history(history_name=history_name)
 
     # Upload data to test_history, run the tool and check the outputs - record
     # API input, job info, tool run exception, as well as exceptions related to
@@ -979,7 +995,8 @@ def verify_tool(tool_id,
             job_data["status"] = status
             register_job_data(job_data)
 
-    galaxy_interactor.delete_history(test_history)
+    if created_history and not no_history_cleanup:
+        galaxy_interactor.delete_history(test_history)
 
 
 def _handle_def_errors(testdef):
--- a/lib/galaxy/tool_util/verify/script.py
+++ b/lib/galaxy/tool_util/verify/script.py
@@ -332,7 +332,7 @@ def _arg_parser():
     parser.add_argument('--force_path_paste', default=False, action="store_true", help='This requires Galaxy-side config option "allow_path_paste" enabled. Allows for fetching test data locally. Only for admins.')
     parser.add_argument('-t', '--tool-id', default=ALL_TOOLS, help='Tool ID')
     parser.add_argument('--tool-version', default=None, help='Tool Version (if tool id supplied). Defaults to just latest version, use * to test all versions')
-    parser.add_argument('-i', '--test-index', default=ALL_TESTS, help='Tool Test Index (starting at 0) - by default all tests will run.')
+    parser.add_argument('-i', '--test-index', default=ALL_TESTS, type=int, help='Tool Test Index (starting at 0) - by default all tests will run.')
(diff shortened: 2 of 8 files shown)


mvdbeek:
This is awesome! Can you resolve the conflicts ? Do you think you could add an (optional?) skip for tests the have `required_data_tables` or `requires_loc_files` ?

jmchilton:
> This is awesome! Can you resolve the conflicts ? Do you think you could add an (optional?) skip for tests the have required_data_tables or requires_loc_files ?

Yes, I will do this. @almahmoud also requested a flag to retry failed tests once - which doesn't seem like it should be an issue but I'm going to trust him and implement it.

almahmoud:
@astrovsky01 re-ran a bunch of tests that had failed the first time, and they passed without any changes, on the same instance, re-running them the same way with ephemeris, but in smaller batches not as part of the full list, hence the assumption that the errors were transient, but full disclosure, I have no looked into them ¯\_(ツ)_/¯

mvdbeek:
Is it possible the retry (which would get a new history) passed because the test makes assumption about the input HID (the number of the dataset in the history) ? If you run a new test you'll get a new history, which is closer to what one would test if running managed planemo tests. Still a good option to have, of course.

lib/galaxy/webapps/galaxy/api/tools.py
```diff
@@ -193,13 +193,25 @@ def test_data(self, trans, id, **kwd):
...
-        # TODO: eliminate copy and paste with above code.
         if 'payload' in kwd:
             kwd = kwd.get('payload')
         tool_version = kwd.get('tool_version', None)
-        tool = self._get_tool(id, tool_version=tool_version, user=trans.user)
-        return [t.to_dict() for t in tool.tests]
+        if tool_version == "*":
+            tools = self.app.toolbox.get_tool(id, get_all_versions=True)
+            for tool in tools:
+                if not tool.allow_user_access(user):
```
mvdbeek:
```suggestion
                if not tool.allow_user_access(trans.user):
```

lib/galaxy/tool_util/verify/interactor.py
```diff
@@ -900,20 +902,34 @@ def verify_tool(tool_id,
...
         skip_message = job_data.get("skip")
-        if skip_message:
-            job_data["status"] = "skip"
-            register_job_data(job_data)
-            return
+
+    if not skip_message and skip_with_reference_data:
+        required_data_tables = tool_test_dict.get("required_data_tables")
+        required_loc_files = tool_test_dict.get("required_loc_files")
+        # TODO: actually hit the API and see if these tables are available.
```
mvdbeek, line 913:
```suggestion
```
Even if they are available it's unlikely that they would contain the minimal examples required by tool tests (hg19, sacCer2 etc. are not uncommon unfortunately).

lib/galaxy/tool_util/verify/interactor.py
```diff
@@ -900,20 +902,34 @@ def verify_tool(tool_id,
...
 
     tool_test_dict.setdefault('maxseconds', maxseconds)
     testdef = ToolTestDescription(tool_test_dict)
     _handle_def_errors(testdef)
 
+    created_history = False
     if test_history is None:
-        test_history = galaxy_interactor.new_history()
+        created_history = True
+        history_name = f"Tool Test History for {tool_id}/{tool_version}-{test_index}"
```
mvdbeek, line 931: ❤️

lib/galaxy/tool_util/verify/interactor.py
jmchilton: That is insane that we do that right? How is that acceptable 😅

lib/galaxy/tool_util/verify/script.py
```diff
@@ -46,66 +295,31 @@ def get_option(key):
...
+        client_test_config=tools_client_test_config,
+        force_path_paste=args.force_path_paste,
+        skip_with_reference_data=args.skip_with_reference_data,
+        quiet=not verbose,
+    )
+    test_tools(
+        galaxy_interactor,
+        test_references,
+        results,
+        verify_quiet=not verbose,
```
mvdbeek, line 318:
```suggestion
```
I think that's only needed in verify_kwds

lib/galaxy/tool_util/verify/script.py
```diff
@@ -116,14 +330,19 @@ def _arg_parser():
...
+    parser.add_argument('--tool-version', default=None, help='Tool Version (if tool id supplied). Defaults to just latest version, use * to test all versions')
     parser.add_argument('-i', '--test-index', default=ALL_TESTS, help='Tool Test Index (starting at 0) - by default all tests will run.')
     parser.add_argument('-o', '--output', default=None, help='directory to dump outputs to')
     parser.add_argument('--append', default=False, action="store_true", help="Extend a test record json (created with --output-json) with additional tests.")
     parser.add_argument('-j', '--output-json', default=None, help='output metadata json')
     parser.add_argument('--verbose', default=False, action="store_true", help="Verbose logging.")
     parser.add_argument('-c', '--client-test-config', default=None, help="Test config YAML to help with client testing")
+    parser.add_argument('--suite-name', default=DEFAULT_SUITE_NAME, help="Suite name for tool test output")
+    parser.add_argument('--skip-with-reference-data', default=False, action="store_true", help="Skip tests the Galaxy server believes use data tables or loc files.")
```
mvdbeek: Shouldn't this be the default ?

lib/galaxy/tool_util/verify/script.py
```diff
@@ -13,7 +17,252 @@
...
+    test_id = label_base + "-" + str(test_index)
+
+    def run_test():
+        run_retries = retries
+        job_data = None
+        job_exception = None
+
+        def register(job_data_):
+            nonlocal job_data
+            job_data = job_data_
```
mvdbeek, line 215: Just wanted to point out that this changes the test json structure, and we'd need to adjust for this in planemo. It does make sense to me to cut out the `data` and `has_data` keys, and I think planemo can support both styles.

mvdbeek (commented):
I gave this a go on my laptop serving usegalaxy.org's toolbox, and this is awesome.
https://github.com/jmchilton/galaxy/pull/80 has some changes I had to make for this to work.

lib/galaxy/tool_util/verify/script.py
jmchilton: I definitely didn't mean to change this - I guess why break it would be my argument. I'll change it back.

lib/galaxy/tool_util/verify/script.py
jmchilton: I think it is now, I have mixed feelings about this but realistically I think people will complain less and hit fewer bugs if we make this the default.

mvdbeek: I understand, and I thought I had deleted the comment before sending the review. I can see it both ways, but thanks for making the change!

lib/galaxy/tool_util/verify/script.py
```diff
@@ -339,8 +342,10 @@ def _arg_parser():
     parser.add_argument('--verbose', default=False, action="store_true", help="Verbose logging.")
     parser.add_argument('-c', '--client-test-config', default=None, help="Test config YAML to help with client testing")
     parser.add_argument('--suite-name', default=DEFAULT_SUITE_NAME, help="Suite name for tool test output")
-    parser.add_argument('--skip-with-reference-data', default=False, action="store_true", help="Skip tests the Galaxy server believes use data tables or loc files.")
-    parser.add_argument('--history-per-test-case', default=False, action="store_true", help="Create new history per test case.")
+    parser.add_argument('--with-reference-data', dest="with_reference_data", default=False, action="store_true")
```
mvdbeek, line 345: I think you need to adjust the parameters to use with_reference_data now.

jmchilton: I think I did this now?

mvdbeek (approved):

PR #10625, opened 2020-11-04
Don't specify default bind type for singularity

bernt-matthias:


bernt-matthias:
Successfully tested on my local installation.

mvdbeek (approved):

PR #10631, opened 2020-11-05
add types and folder_id filter to the main data library index API

bgruening:


lib/galaxy/webapps/galaxy/api/library_contents.py
```diff
@@ -104,21 +110,34 @@ def traverse(folder):
...
-                             url=url_for('library_content', library_id=library_id, id=encoded_id, )))
+                             type='folder',
+                             name='/',
+                             url=url_for('library_content', library_id=library_id, id=encoded_id)))
+            root_folder = library.root_folder
+
+        root_folder.api_path = ''
+        # appending all other items in the library recursively
+        for content in traverse(root_folder):
+            if types and content.api_type in types:
```
simonbray:
```suggestion
            if not types or content.api_type in types:
```
Should be correct, tests are passing for me locally with this.

bgruening: :+1:

lib/galaxy/webapps/galaxy/api/library_contents.py
```diff
@@ -54,6 +54,12 @@ def index(self, trans, library_id, **kwd):
         :param  library_id: the encoded id of the library
         :type   library_id: str
 
+        :type   folder_id:  str
+        :param  folder_id:  (optional) a folder_id to restrict the listing to
+
+        :type   types:      str
+        :param  types:      (optional) comma seperated list if contents to index (currently dataset and folder is supportd).
```
mvdbeek, line 61:
```suggestion
        :param  types:      (optional) comma seperated list of contents to index (currently dataset and folder is supported).
```

mvdbeek:
This seems a little odd, wouldn't you want to call into the folder API directly ?

bgruening:
I think this is a valid use case no? You want to filter an entire library. It' also hard to get the folder ID without using the library API.

mvdbeek:
What's the problem with chaining those API calls ? This is going to block for a very long time as it is and I don't think having an API endpoint that returns an unlimited amount of data is a good idea. At a minimum you'd want to add offset and pagination, which the folders endpoint already provides.

bgruening:
That is the point of PR. Currently, the API endpoint takes a large amount of time and returns everything. This PR extends the functionality to filter the result set to only the things that you need. E.g. only the list of folders of a particular library.

mvdbeek:
I'm ok if you only want to return folders in that case. But if you include datasets that's gonna fall apart just as quickly as it does now.

bgruening:
> But if you include datasets that's gonna fall apart just as quickly as it does now.

I'm aware of that. This depends on the folder size in the end. But it's at least an improvement to the current state imho. Pagination on top would be even better. But then again we should recommend the folder API, once we have retrieved the folders.

mvdbeek:
How does the folder id argument fit into this ? If you have the folder id shouldn't you pass via the folder endpoint ?

mvdbeek:
Looks like library things are on the roadmap for Q1, maybe we can revisit this with pagination then ?

mvdbeek:
Do we still need this ?

PR #10637, opened 2020-11-06
Fix saving rule builder rules when uploading in rule editing view

assuntad23:
by moving data of saved rules to a separate component from the template.

Bugfix to fix console log errors when trying to upload collection in mapping view. 
This PR will allow users to upload their rule collection builder regardless of what view they are in.


mvdbeek (approved):
Thanks, that's a nice fix!

Issue #10639, opened 2020-11-06
Metadata setting fails for AnnData 0.7.4 files


pcm32:
When running our tool that converts mtx files into Scanpy AnnData objects in usegalaxy.eu we got an error from the set_meta method in the AnnData class of the binary datatypes code.

Specifically the issues comes from here:
https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/datatypes/binary.py#L1010

Due to the "index" key not being available for "some" AnnData 0.7.4 files.

Our colleague @nh3 (Ni Huang) spotted that Scanpy stores in "index" or "_index" depending on whether the AnnData is a re-write from an older object or from a complete new AnnData creation from other formats. Probably the set_meta method was only tested against that first case.

https://github.com/theislab/anndata/blob/77a633cc825b428d02dabdf75491841c3a175a8d/anndata/_io/h5ad.py#L247

So what we would need is that the set_meta above is more defensive. This history has examples AnnData: https://usegalaxy.eu/u/pmoreno/h/test-older-and-newer-anndata-from-zero although the most useful stuff is at the galaxy logging behind the scenes.

Thanks to Ni Huang, Jon Manning, @mtekman and @bgruening to help spot this.

PR #10640, opened 2020-11-06
[20.09] Fix Anndata index access, use numpy-safe JSON encoder

mtekman:
Re-added try/except block due to failing datasets: https://usegalaxy.eu/u/pmoreno/h/test-older-and-newer-anndata-from-zero

references issue here https://github.com/galaxyproject/galaxy/issues/10639


lib/galaxy/datatypes/binary.py
```diff
@@ -981,78 +981,81 @@ def sniff(self, filename):
...
+                        size = int(tmp.size)
+                    else:
+                        layers = [util.unicodify(x) for x in list(tmp.keys())]
+                        count = len(layers)
+                        size = lennames
+                    return (layers, count, size)
+
+                if 'obs' in dataset.metadata.layers_names:
+                    tmp = anndata_file["obs"]
+                    dataset.metadata.obs_names = [util.unicodify(x) for x in tmp["index"]]
```
pcm32:
```suggestion
                    tmp = anndata_file["obs"]
                    if "index" in tmp:
                        obs_index = "index"
                    else:
                        obs_index = "_index"
                    dataset.metadata.obs_names = [util.unicodify(x) for x in tmp[obs_index]]
```
Maybe we can take the chance to cater for the "_index" 0.7.4 case

bernt-matthias:
Seems preferable to add a test case for the failing case and then really fix the problem. If the guess in the linked issue is correct one would need to access `_index` instead of `index` of the later is unavailable.

pcm32:
Sure, we can leave that fix for later and at least avoid failing now with the exception handling. I have passed @mtekman  some test data, maybe he intended to do it on another PR and I rushed. Anyway is fine for me.

lib/galaxy/datatypes/binary.py
pcm32: feel free to neglect as per below comments.

mtekman:
@bernt This is just a hotfix for now

> Seems preferable to add a test case for the failing case and then really fix the problem. If the guess in the linked issue is correct one would need to access `_index` instead of `index` of the later is unavailable.

this is just a hotfix for now to get the datasets running before next week

Another PR will try address the underlying issue

pcm32:
maybe it would be better then to revert my suggestion and keep this clean for just dealing with the exception/error.

mtekman:
I see no harm in trying to parse both types of datasets, and catching the outliers at the same time

pcm32:
ok, code above is not tested though, but should be fine.

pcm32:
we might want to repeat the operation then on var, where apparently the same applies.

mtekman:
oh good point I hadn't considered that

mvdbeek:
Can we please fix this properly ? I am not a fan of the try/except ... either the information that `set_meta` generates is important, and then we should parse it correctly, or it is not, in which case we should not try to parse it.

If the format is in flux entirely maybe you'd want to raise this with the authors. If I look at https://github.com/theislab/anndata/blame/77a633cc825b428d02dabdf75491841c3a175a8d/anndata/_io/h5ad.py#L251 I would think that the index name is stored in `_index`, always, so what's in here doesn't look correct either.

If we generate wrong metadata that's worse then erroring.

pcm32:
Unfortunately index is not always on "_index", and people could upload AnnData that are older.

mtekman:
@mvdbeek These datasets are likely to further evolve rather than stabilise, and I can't think of a future-proof solution other than try/except

mvdbeek:
> @mvdbeek These datasets are likely to further evolve rather than stabilise, and I can't think of a future-proof solution other than try/except

but then you'll never know that something is not as expected with a new dataset. Isn't there some kind of version indication if this changes so rapidly ?

mtekman:
Likely. I did previously search for version tags in the test data that we have, but was unable to find anything...

I guess we could remove the try/except and test for a few required keys as you say, and then still return a green dataset if none are found

pcm32:
> I guess we could remove the try/except and test for a few required keys as you say, and then still return a green dataset if none are found

Yes please, the current state is just blocking any further tool execution with the red dataset. Maybe if the index key is not found we could have a message saying this, but avoid leaving the dataset in failure state?

pcm32:
> If I look at https://github.com/theislab/anndata/blame/77a633cc825b428d02dabdf75491841c3a175a8d/anndata/_io/h5ad.py#L251 I would think that the index name is stored in `_index`, always, so what's in here doesn't look correct either.

In the associated issue #10640 there is a link to AnnData code showing that even in the newer version there are circumstances where "index" instead of "_index" would be used. This is partly why this flew under the radar.

mvdbeek:
> Maybe if the index key is not found we could have a message saying this, but avoid leaving the dataset in failure state?

Does the PR in its current form, minus the try/except, fail any known datasets ?
If not, it seems much preferable to remove them, so we actually know when there's a weird dataset appearing.

Keep in mind that faulty metadata can lead to wrong analysis results. That is much worse than having to report a bug.

mtekman:
> Does the PR in its current form, minus the try/except, fail any known datasets ?

We need more datasets, and we have some, but they need to be greatly reduced in size before we can add them as test data to galaxy

mvdbeek:
The test can't fail now, can you push the changes ?

mtekman:
> The test can't fail now, can you push the changes ?

Some of the newer datasets are failing locally... give me 30 minutes to debug and then I'll push

mtekman:
I was unable to debug why things were failing locally, so I pushed in the hopes that you guys might....

lib/galaxy/model/metadata.py
```diff
@@ -212,9 +211,9 @@ def to_JSON_dict(self, filename=None):
         if '__validated_state_message__' in dataset_meta_dict:
             meta_dict['__validated_state_message__'] = dataset_meta_dict['__validated_state_message__']
         if filename is None:
```
nsoranzo, line 214:
```suggestion
        encoded_meta_dict = galaxy.model.custom_types.json_encoder.encode(meta_dict)
        if filename is None:
```

lib/galaxy/model/metadata.py
```diff
@@ -212,9 +211,9 @@ def to_JSON_dict(self, filename=None):
...
-            return json.dumps(meta_dict)
+            return galaxy.model.custom_types.json_encoder.encode(meta_dict)
```
nsoranzo:
```suggestion
            return encoded_meta_dict
```

lib/galaxy/model/metadata.py
```diff
@@ -212,9 +211,9 @@ def to_JSON_dict(self, filename=None):
...
         with open(filename, 'wt+') as fh:
-            json.dump(meta_dict, fh)
+            fh.write(galaxy.model.custom_types.json_encoder.encode(meta_dict))
```
nsoranzo:
```suggestion
            fh.write(encoded_meta_dict)
```

lib/galaxy/datatypes/binary.py
```diff
@@ -998,21 +1010,29 @@ def set_meta(self, dataset, overwrite=True, **kwd):
...
                     size = lennames
                 return (layers, count, size)
 
             if 'obs' in dataset.metadata.layers_names:
                 tmp = anndata_file["obs"]
-                dataset.metadata.obs_names = [util.unicodify(x) for x in tmp["index"]]
-                dataset.metadata.obs_layers, \
-                    dataset.metadata.obs_count, \
-                    dataset.metadata.obs_size = _layercountsize(tmp, len(dataset.metadata.obs_names))
+                obs_index = False
```
```suggestion
                obs_index = None
```

lib/galaxy/datatypes/binary.py
```diff
@@ -1022,17 +1042,25 @@ def _layercountsize(tmp, lennames=0):
...
+                dataset.metadata.raw_var_size = z
 
             if 'var' in dataset.metadata.layers_names:
                 tmp = anndata_file["var"]
-                # row names are never used in preview windows
-                # dataset.metadata.var_names = tmp["index"]
-                dataset.metadata.var_layers, \
-                    dataset.metadata.var_count, \
-                    dataset.metadata.var_size = _layercountsize(tmp, len(tmp["index"]))
+                var_index = False
```
```suggestion
                var_index = None
```

mvdbeek (approved):
Thank you @mtekman & @pcm32 & @nsoranzo !

nsoranzo (approved):

Issue #9844, opened 2020-06-03
Top menu disappears at width=991px


jdavcs:
The menu's `ul` element becomes vertical. Happens due to `flex-direction: column` for `.navbar-nav`, I think.
Maybe display a horizontal scrollbar in the middle pane instead?

![1](https://user-images.githubusercontent.com/6132258/83655630-a2968780-a58c-11ea-9d1f-021870628666.png)
![2](https://user-images.githubusercontent.com/6132258/83655648-a6c2a500-a58c-11ea-846d-2c7ff72cf9f4.png)

martenson:
Can reproduce. This renders Galaxy interface completely useless while working with multiple browser windows or on mobile.

dannon:
Fixed in #10641

PR #10641, opened 2020-11-06
[20.09] Fix masthead items disappearing on small screens

mvdbeek:
The fix is to not set toggleable on `<b-navbar>`. I also played with with adding a `<b-navbar-toggle/>` to display a button on the right for small screens, but I didn't manage to get something that looks nice, though it's probably not hard for someone that knows css.

To get a little bit more space in the navbar I'm setting the (non-functional) chat server item to `display: none` and I'm moving the scratchbook to the rightmost position (since it reserves some space just on the right to display active scratchbooks).

With this in place I see the full navbar on my phone.

Fixes https://github.com/galaxyproject/galaxy/issues/9844.

### DIFF
--- a/client/src/components/Masthead/Masthead.vue
+++ b/client/src/components/Masthead/Masthead.vue
@@ -116,7 +116,7 @@ export default {
         },
         tabs() {
             const scratchbookTabs = [this.mastheadState.frame.buttonActive, this.mastheadState.frame.buttonLoad];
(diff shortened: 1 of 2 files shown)


dannon (approved):
This is fine for a 20.09 fix, but we should follow up and fix togglable and use the hamburger properly in dev.

Issue #10638, opened 2020-11-06
'Hide history panel' glitch in the UI


simonbray:
Probably screenshots explain this better than paragraphs of text. Original view:

![image](https://user-images.githubusercontent.com/32272674/98353927-53172480-2020-11eb-976d-04d3b9821f64.png)

After clicking the icon in the bottom right corner to hide the history panel:

![image](https://user-images.githubusercontent.com/32272674/98354030-75a93d80-2020-11eb-84dd-05c3ea07e789.png)

Now I can unexpectedly scroll to the right and see an empty white space with some misplaced icons and text:

![image](https://user-images.githubusercontent.com/32272674/98354149-9d98a100-2020-11eb-987c-223c820dbfad.png)

Seems quite similar to some of the issues in https://github.com/galaxyproject/galaxy/issues/10553.

mvdbeek:
Should be fixed in https://github.com/galaxyproject/galaxy/pull/10644, thanks @simonbray!

PR #10644, opened 2020-11-06
[20.09] Fix hide history panel glitch

mvdbeek:
That fixes https://github.com/galaxyproject/galaxy/issues/10638, but it may break other stuff ?
https://github.com/galaxyproject/galaxy/commit/9eb70d87297778c127d1da3a6acf0363ccb79bff
suggests this was in place to prevent a double scroller bug. I haven't noticed those.

### DIFF
--- a/client/src/style/scss/base.scss
+++ b/client/src/style/scss/base.scss
@@ -159,7 +159,7 @@ body {
     @extend .unified-panel;
     right: 0px;
     width: $panel-width;
-    overflow: unset !important;
+    overflow: unset;
 }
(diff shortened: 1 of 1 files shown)


mvdbeek:
@dannon if you're power is back up, can you take a look at this ? I've tested this with firefox and chrome and haven't seen any double scrolling problems.

dannon:
@mvdbeek This looks fine to me; the left panel was the one exhibiting a double scroller and, of course, they behave differently :)

PR #10650, opened 2020-11-07
Fix ObjectUploadTarget handling in tool util staging code.

jmchilton:
With test cases.

### DIFF
--- a/lib/galaxy/datatypes/text.py
+++ b/lib/galaxy/datatypes/text.py
@@ -117,19 +117,25 @@ class ExpressionJson(Json):
     def set_meta(self, dataset, **kwd):
         """
         """
-        json_type = "null"
-        with open(dataset.file_name) as f:
-            obj = json.load(f)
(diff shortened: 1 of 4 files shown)


lib/galaxy/datatypes/text.py
```diff
@@ -117,19 +117,28 @@ class ExpressionJson(Json):
...
-            elif isinstance(obj, dict):
-                json_type = "object"
-
-        dataset.metadata.json_type = json_type
+        if dataset.has_data():
+            json_type = "null"
+            file_path = dataset.file_name
+            file_size = os.stat(file_path).st_size
+            if file_size == 0:
+                raise Exception("Invalid empty JSON encountered")
```
mvdbeek: `has_data` should already check the size. But should we raise that exception in the else branch ?

jmchilton: Just left over debugging, I'll remove that condition.

PR #10656, opened 2020-11-08
Add a script to start Galaxy under aiohttp

mvdbeek:
This is nice because we can run Galaxy under a pure-python, non-blocking
server, and we properly parse galaxy.yml. That means we can debug in
vscode without needing the old galaxy.ini.

Obviously we'd want to integrate this a bit better so we can decorate our to-be-created async / websocket routes.


mvdbeek:
I think fastAPI is the way to go atm.

PR #10658, opened 2020-11-09
Rev cwltool to 3.0.20201109103151

jmchilton:
I had previously pinned it to a 2019 version for Python 2 compat, but we've now caught up with them and surpassed them in terms of limiting our support with versions of Python.

The next version of cwltool after this (``3.0.20200317142916``)  breaks compatibility with our ``lib/galaxy/tool_util/cwl/parser.py``.

In that version, I think cores is expected to be a real number so deferring to GALAXY_SLOTS at runtime with the following code in that file breaks:

```python
    def _select_resources(self, request, runtime_context=None):
        new_request = request.copy()
        new_request["cores"] = "$GALAXY_SLOTS"
        return new_request
```

(Can quickly see this with the unit test test/unit/tools/test_cwl.py available in the CWL branch of Galaxy.)


mr-c:
The usage of opaque strings here is explicitly legal in CWL v1.0+, so this is a cwltool bug.

>  As noted earlier, an implementation may perform deferred resolution of runtime fields by providing opaque strings for any or all of the following fields; parameter references and expressions may only use the literal string value of the field and must not perform computation on the contents.

_From https://www.commonwl.org/v1.0/CommandLineTool.html#Runtime_environment_

Does anyone have a traceback handy that I can use to narrow down the offending code?

mr-c:
This PR might fix it https://github.com/common-workflow-language/cwltool/pull/1363

mvdbeek:
Can confirm it does!

nsoranzo:
I think there are some missing deps, e.g. argcomplete, will try to update with Pipfile.

nsoranzo:
It could make sense to rebase on top of https://github.com/galaxyproject/galaxy/pull/10660 , if that passes.

jmchilton:
I haven't looked into, but my hunch is cwltool need a pin on a different schema salad?

```
  File "/Users/john/workspace/galaxy/.venv/lib/python3.6/site-packages/cwltool/main.py", line 38, in <module>
    from schema_salad.ref_resolver import (
ImportError: cannot import name 'FetcherCallableType'
```

I'll see what I can do in branch.

jmchilton:
Running the green tests now with a high pin for schema salad also - 15% in and things look like the right shade of green so far.

nsoranzo (approved):

PR #10667, opened 2020-11-09
Add default dbkey when collecting builds

### LINKED ISSUE #10665: Remove dependency on builds.txt
If builds.txt is empty Galaxy will not allow uploads because it is expecting the `?	unspecified (?)` dbkey.
Ideally dbkey should be allowed to be empty or omitted while uploading.

innovate-invent:
Resolves: https://github.com/galaxyproject/galaxy/issues/10665

### DIFF
--- a/lib/galaxy/util/__init__.py
+++ b/lib/galaxy/util/__init__.py
@@ -1177,10 +1177,7 @@ def compare_urls(url1, url2, compare_scheme=True, compare_hostname=True, compare
 
 def read_dbnames(filename):
     """ Read build names from file """
-    class DBNames(list):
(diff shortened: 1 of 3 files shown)


innovate-invent:
This should actually be handled here: https://github.com/galaxyproject/galaxy/blob/3337e249f52492ed8eb5a8f15e967660704e692f/lib/galaxy/util/__init__.py#L1224-L1225

But isn't despite the builds.txt.sample file not existing on my deployment.

mvdbeek:
Thanks, that's a nice fix @innovate-invent!

PR #10668, opened 2020-11-10
wait_for_invocation should take in an assert_ok flag like other wait methods.

jmchilton:
Changes the behavior so we should make sure the API tests pass.


nsoranzo (approved):

mvdbeek:
Hooray, looks good. Want to take it out of WIP @jmchilton ?

Issue #10066, opened 2020-08-05
History dataset separators disappeared from 20.05 user interface


wm75:
Without a line separator between datasets in the history it is really hard to know, especially for hidden/deleted datasets, to tell where information about one dataset begins or ends.

![Screenshot from 2020-08-05 12-49-56](https://user-images.githubusercontent.com/4742081/89404531-71683e00-d71a-11ea-84eb-f3524bdcc6ac.png)

dannon:
Fixed in 10675.

PR #10675, opened 2020-11-10
[20.09] Fix border between list items

dannon:
Dropping borders between list items that otherwise have no spacing has been confusing

Fixes https://github.com/galaxyproject/galaxy/issues/10066

This additionally makes the border inherit the color of the relevant list item when it's a dataset.  Looks better than the grey.

image

### DIFF
--- a/client/src/style/scss/dataset.scss
+++ b/client/src/style/scss/dataset.scss
@@ -189,7 +189,15 @@
     }
 }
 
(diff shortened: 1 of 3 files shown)


mvdbeek (approved):

OlegZharkov (approved):
that's great

PR #10678, opened 2020-11-10
[20.09] Fix UTC Date date change

### LINKED ISSUE #10600: Workflow list sorting issue
Hi, 

just wanted to report a strange behaviour of the workflow list: 

On the EU Server the [workflow list](https://usegalaxy.eu/workflows/list) "updated" column seems not to move accordingly when sorting for other columns: 

Screenshot of original workflow list: 
wf_original_order

Clicking on "Name" sorts workflows according to name and it sorts all other columns accordingly, but the "updated" column stays exactly the same: 
wf_name_sorted

Thanks,
Melanie

OlegZharkov:
Took me a little while to get what is going on here. The Date stayed the same, even though sorting worked fine.

This PR will update date in `UtcDate.vue`  on props change. It will fix UtcDate sorting in Invocations, Datasetlist, InteractiveTools, LibraryFodler and WorkflowList tables. fixes https://github.com/galaxyproject/galaxy/issues/10600

Before:
Peek 2020-11-10 18-38

Now:
my-f-giff

thanks @foellmelanie for reporting

### DIFF
--- a/client/src/components/UtcDate.vue
+++ b/client/src/components/UtcDate.vue
@@ -25,8 +25,12 @@ export default {
         },
     },
     created() {
-        if (this.customFormat) this.processedDate = moment(this.date, this.customFormat).format();
-        else this.processedDate = this.date;
+        this.initialize(this.date);
+    },
+    data: function () {
+        return {
+            processedDate: null,
+        };
     },
     computed: {
         elapsedTime: function () {
@@ -34,6 +38,17 @@ export default {
         },
         fullDate: function () {
             return moment.utc(this.processedDate).format();
+        },
+    },
+    watch: {
+        date: function (newDate) {
+            this.initialize(newDate);
+        },
+    },
+    methods: {
+        initialize: function (date) {
+            if (this.customFormat) this.processedDate = moment(date, this.customFormat).format();
+            else this.processedDate = date;
         },
     },
 };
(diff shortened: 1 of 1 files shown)


dannon:
@OlegZharkov I started opening up a PR to your branch but this was faster.  I've just pushed a different approach to this branch here that I think is a bit simpler -- feel free to dump it if it isn't working for you or if you don't like it.

client/src/components/UtcDate.vue
```diff
@@ -22,33 +22,22 @@ export default {
         },
         customFormat: {
             type: String,
+            default: null,
```
OlegZharkov, line 25: why?

OlegZharkov:
> @OlegZharkov I started opening up a PR to your branch but this was faster. I've just pushed a different approach to this branch here that I think is a bit simpler -- feel free to dump it if it isn't working for you or if you don't like it.

No worries, I tried this change and it works just fine. Thanks :+1:

client/src/components/UtcDate.vue
dannon: Props should always have explicit defaults.

bgruening (approved):
Thanks @OlegZharkov and @dannon!

OlegZharkov:
thanks @bgruening @dannon

PR #10691, opened 2020-11-11
Fix element_is_set for MetadataWrapper

bernt-matthias:
follow up to https://github.com/galaxyproject/galaxy/pull/8599

in the planemo tests in the line https://github.com/bernt-matthias/galaxy/blob/1882aa641cdde1c51c077fd983f2241971308377/lib/galaxy/tools/parameters/dynamic_options.py#L205

`r.metadata` is sometimes a `MetadataWrapper` and sometimes a `MetadataCollection`. Wondering why previous tests did not reveal this .. or should we have done a rebase?

In the former case access to the `element_is_set` function defined in `MetadataCollection` is not possible.

Maybe there is a better solution. Also a check that metadata names should not be equal to attributes of `MetadataCollection` could be nice..?

### DIFF
--- a/lib/galaxy/tools/wrappers.py
+++ b/lib/galaxy/tools/wrappers.py
@@ -214,6 +214,9 @@ class DatasetFilenameWrapper(ToolParameterValueWrapper):
             self.compute_environment = compute_environment
 
         def __getattr__(self, name):
+            if hasattr(self.metadata, name):
+                return getattr(self.metadata, name)
+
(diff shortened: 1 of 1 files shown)


mvdbeek:
Awesome, thanks for the followup.

bernt-matthias:
Then I proceed and prepare a backport to 20.09

mvdbeek:
Sounds good!

PR #10693, opened 2020-11-11
[20.09] Bugfix for tool search filtering for tools outside sections.

### LINKED ISSUE #10325: Tools installed outside of toolbar sections fail to show in search without sections being shown
If, when loaded in, a tool is not placed in a section, when searching for that tool, it does not appear as a result. To make it appear, sections must be selected to be shown.

dannon:
Fixes #10325

Includes a (very) minor unrelated followup to #10678

### DIFF
--- a/client/src/components/UtcDate.vue
+++ b/client/src/components/UtcDate.vue
@@ -22,7 +22,7 @@ export default {
         },
         customFormat: {
             type: String,
-            default: null,
+            default: undefined,
         },
     },
     computed: {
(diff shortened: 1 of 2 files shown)


jmchilton (approved):

client/src/components/UtcDate.vue
```diff
@@ -33,7 +33,7 @@ export default {
             return moment.utc(this.formattedDate).format();
         },
         formattedDate: function () {
-            if (this.customFormat) {
+            if (this.customFormat !== undefined) {
```
OlegZharkov, line 36: ```(this.customFormat)``` is kinda concise and does the same IMHO... sorry

dannon: It's similar, resolving `this.customFormat`'s truthiness, but especially in javascript I lean towards preferring explicit, strict comparisons when you know what the inputs/types are expected to be.  In any situation where this is defined (`!== undefined`), someone is trying to do something with the formatting here and we want it attempted so it, at minimum, blows up if it's not right.

PR #10698, opened 2020-11-11
Add refresh_time to irods configuration, so that connections older than refresh_time are re-created.

kxk302:
This is to avoid NetworkException we were getting when we created a session and maintained it in Galaxy.


lib/galaxy/config/sample/object_store_conf.xml.sample
```diff
@@ -49,6 +49,7 @@
             <resource name="demoResc" />
             <zone name="tempZone" />
             <connection host="localhost" port="1247" timeout="30" poolsize="3"/>
+            <icommand file_size="100000000" num_threads="3"/>
```
kxk302: For files larger than 100 MB, use iput with 3 threads.

lib/galaxy/dependencies/conditional-requirements.txt
```diff
@@ -8,7 +8,7 @@ drmaa
 statsd
 docker
 azure-storage==0.32.0
-python-irodsclient==0.8.4
+python-irodsclient==0.8.5
```
kxk302, line 11: Latest version of client code with connection re-create logic.

lib/galaxy/objectstore/irods.py
```diff
@@ -68,6 +73,13 @@ def parse_config_xml(config_xml):
         port = int(c_xml[0].get('port', 0))
         timeout = int(c_xml[0].get('timeout', 30))
         poolsize = int(c_xml[0].get('poolsize', 3))
+        refresh_time = int(c_xml[0].get('refresh_time', 300))
```
kxk302, line 71: Parse connection_refresh_time

lib/galaxy/objectstore/irods.py
```diff
@@ -146,7 +168,7 @@ class IRODSObjectStore(DiskObjectStore, CloudConfigMixin):
     store_type = 'irods'
 
     def __init__(self, config, config_dict):
-        reload_timer = ExecutionTimer()
+        ipt_timer = ExecutionTimer()
```
kxk302, line 152: Add timers to measure irods calls

lib/galaxy/objectstore/irods.py
```diff
@@ -376,73 +429,77 @@ def _push_to_irods(self, rel_path, source_file=None, from_string=None):
...
         options = {kw.FORCE_FLAG_KW: ''}
 
-        if os.path.exists(source_file):
-            # Check if the data object exists in iRODS
-            collection_path = self.home + "/" + str(subcollection_name)
-            data_object_path = collection_path + "/" + str(data_object_name)
-            exists = False
-            try:
-                exists = self.session.data_objects.exists(data_object_path)
+        if not os.path.exists(source_file):
```
kxk302, line 411: Simplify the logic here. First handle the case where path does not exist.

lib/galaxy/objectstore/irods.py
```diff
@@ -376,73 +429,77 @@ def _push_to_irods(self, rel_path, source_file=None, from_string=None):
...
+                self.session.collections.create(collection_path, recurse=True)
 
-                    end_time = datetime.now()
-                    log.debug("Pushed cache file '%s' to collection '%s' (%s bytes transfered in %s sec)",
-                            source_file, rel_path, os.path.getsize(source_file), end_time - start_time)
-                return True
-            except NetworkException as e:
-                log.exception(e)
-                return False
+                self.put_data_object(source_file, collection_path)
```
kxk302: Instead of directly writing the file, use this helper method that has logic to use iput for large files.

lib/galaxy/objectstore/irods.py
```diff
@@ -547,7 +614,11 @@ def _delete(self, obj, entire_dir=False, **kwargs):
 
             else:
                 # Delete from cache first
-                os.unlink(self._get_cache_path(rel_path))
+                try:
```
kxk302, line 589: Handle case where the file does not exist.

lib/galaxy/objectstore/irods.py
```diff
@@ -663,3 +747,23 @@ def _get_object_url(self, obj, **kwargs):
 
     def _get_store_usage_percent(self):
         return 0.0
+
+    def put_data_object(self, source_file, collection_path):
```
kxk302: Helper metho that uses iput for large files.

kxk302:
I am thinking maybe this can wait. I will look at the multi-threaded python client next week (apparently its been sitting there for a while). If it can be revived soon, it does not make sense for this PR to go through, as it needs to be undone then.

kxk302:
This PR allows configuration of refresh_time (in seconds). An irods connection older than refresh_time will be re-created in the Python irods client code. This is to avoid the NetworkException we were getting on older connection. I'd like to merge this so that I can update Test to take advantage of this feature (prior to this I deployed my branch with these changes on Test).

kxk302:
@mvdbeek @natefoo @jmchilton 
Guys,
When you a chance, could one of you please take a look at this PR?
Thx!

lib/galaxy/objectstore/irods.py
```diff
@@ -213,18 +219,20 @@ def __init__(self, config, config_dict):
         if irods is None:
             raise Exception(IRODS_IMPORT_MESSAGE)
 
-        self.session = iRODSSession(host=self.host, port=self.port, user=self.username, password=self.password, zone=self.zone)
+        self.session = iRODSSession(host=self.host, port=self.port, user=self.username, password=self.password, zone=self.zone, refresh_time=self.refresh_time)
         # Set connection timeout
         self.session.connection_timeout = self.timeout
-        log.debug("irods __init__ %s", reload_timer)
+        log.debug("irods_pt __init__: %s", ipt_timer)
```
mvdbeek, line 225: What is that you're measuring in this timer ? That code runs just once, I don't think it is useful to log here.

lib/galaxy/objectstore/irods.py
```diff
@@ -213,18 +219,20 @@ def __init__(self, config, config_dict):
...
 
     def shutdown(self):
         # This call will cleanup all the connections in the connection pool
         # OSError sometimes happens on GitHub Actions, after the test has successfully completed. Ignore it if it happens.
+        ipt_timer = ExecutionTimer()
         try:
             self.session.cleanup()
         except OSError:
             pass
+        log.debug("irods_pt shutdown: %s", ipt_timer)
```
mvdbeek, line 235: Same, I don't think that is informative

lib/galaxy/objectstore/irods.py
```diff
@@ -273,17 +282,20 @@ def _construct_path(self, obj, base_dir=None, dir_only=None, extra_dir=None, ext
             rel_path = os.path.join(rel_path, str(self._get_object_id(obj)))
         if base_dir:
             base = self.extra_dirs.get(base_dir)
+            log.debug("irods_pt _construct_path: %s", ipt_timer)
             return os.path.join(base, rel_path)
 
         if not dir_only:
             rel_path = os.path.join(rel_path, alt_name if alt_name else "dataset_%s.dat" % self._get_object_id(obj))
+        log.debug("irods_pt _construct_path: %s", ipt_timer)
```
mvdbeek, line 290: Those 3 timings do not involve irods at all, drop it ?

lib/galaxy/objectstore/irods.py
```diff
@@ -354,17 +374,21 @@ def _download(self, rel_path):
         except NetworkException as e:
             log.exception(e)
             return False
+        finally:
+            log.debug("irods_pt _download: %s", ipt_timer)
```
mvdbeek, line 378: That doesn't seem useful, that's just logging how much time `data_objects.get` takes.

mvdbeek (approved):

kxk302:
Thx for review/merge @mvdbeek, 
I had added timer to all methods, including those creating and destroying a session (which are called only once). 
The data_objects.get() timer is supposed to time the method if an exception is thrown. I can move it to inside exception clauses and remove the finally

PR #10700, opened 2020-11-12
Allow expression tools to output copies of input HDAs.

jmchilton:
Important for conditional logic in workflows. Picking between different files in workflow depending on the metadata, file contents, outputs not being skipped, etc..

Also tests null file handling in expression tools and tool test plumbing.

Builds on #10694

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -3181,6 +3181,25 @@ class HistoryDatasetAssociation(DatasetInstance, HasTags, Dictifiable, UsesAnnot
             self.version = self.version + 1 if self.version else 1
             session.add(past_hda)
 
+    def copy_from(self, other_hda):
+        # This deletes the old dataset, so make sure to only call this on new things
+        # in the history (e.g. during job finishing).
+        old_dataset = self.dataset
+        self._metadata = None
+        self.metadata = other_hda.metadata
+        self.info = other_hda.info
+        self.blurb = other_hda.blurb
+        self.peek = other_hda.peek
+        self.extension = other_hda.extension
+        self.designation = other_hda.designation
+        self.deleted = other_hda.deleted
+        self.visible = other_hda.visible
+        self.validated_state = other_hda.validated_state
+        self.validated_state_message = other_hda.validated_state_message
+        self.copy_tags_from(self.history.user, other_hda)
+        self.dataset = other_hda.dataset
+        old_dataset.full_delete()
+
     def copy(self, parent_id=None, copy_tags=None, flush=True, copy_hid=True, new_name=None):
         """
         Create a copy of this HDA.
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -2483,6 +2483,22 @@ class ExpressionTool(Tool):
         with open(expression_inputs_path, "w") as f:
             json.dump(expression_inputs, f)
 
+    def exec_after_process(self, app, inp_data, out_data, param_dict, job=None, **kwds):
+        for key, val in self.outputs.items():
+            if val.output_type == "data":
+                with open(out_data[key].file_name, "r") as f:
+                    src = json.load(f)
+                assert isinstance(src, dict)
+                dataset_id = src["id"]
+                copy_object = None
+                for input_dataset in inp_data.values():
+                    if input_dataset.id == dataset_id:
+                        copy_object = input_dataset
+                        break
+                if copy_object is None:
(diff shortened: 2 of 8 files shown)


lib/galaxy/model/__init__.py
```diff
@@ -3181,6 +3181,25 @@ def __create_version__(self, session):
             self.version = self.version + 1 if self.version else 1
             session.add(past_hda)
 
+    def copy_from(self, other_hda):
+        # This deletes the old dataset, so make sure to only call this on new things
+        # in the history (e.g. during job finishing).
+        old_dataset = self.dataset
+        self._metadata = None
+        self.metadata = other_hda.metadata or dict()
```
mvdbeek: Is that necessary ? I was under the impression that the metadata attribute can't be None.

lib/galaxy/model/__init__.py
```diff
@@ -3181,6 +3181,25 @@ def __create_version__(self, session):
...
+        self.extension = other_hda.extension
+        self.designation = other_hda.designation
+        self.deleted = other_hda.deleted
+        self.visible = other_hda.visible
+        self.validated_state = other_hda.validated_state
+        self.validated_state_message = other_hda.validated_state_message
+        self.copy_tags_from(self.history.user, other_hda)
+        self.dataset = other_hda.dataset
+        old_dataset.full_delete()
+
```
mvdbeek, line 3202: do we need to add basename / can we re-use an existing method ?

nsoranzo:
Can you rebase to make it easier to review?

lib/galaxy/model/__init__.py
```diff
@@ -3181,6 +3181,25 @@ def __create_version__(self, session):
...
+        self.info = other_hda.info
+        self.blurb = other_hda.blurb
+        self.peek = other_hda.peek
+        self.extension = other_hda.extension
+        self.designation = other_hda.designation
+        self.deleted = other_hda.deleted
+        self.visible = other_hda.visible
+        self.validated_state = other_hda.validated_state
+        self.validated_state_message = other_hda.validated_state_message
+        self.copy_tags_from(self.history.user, other_hda)
```
mvdbeek, line 3199: Should we set 
```suggestion
        self.copy_tags_from(self.history.user, other_hda)
        self.copied_from_history_dataset_association = other_hda.copied_from_history_dataset_association
```

?

mvdbeek (commented):
Awesome! I We could even use this to do some of the things DatabaseOperationTools do today.

lib/galaxy/model/__init__.py
mvdbeek: Thought I deleted that comment. Nevermind.

lib/galaxy/model/__init__.py
jmchilton: That is used for tracing back to a job during workflow extraction and this will have its own job though... Is there something else we use this for where we would need this tracking? If you think it is just good to track though I'm happy to add it... I think.

lib/galaxy/model/__init__.py
jmchilton: I copied that code from the copy() method. I can remove it.

lib/galaxy/tools/__init__.py
```diff
@@ -2483,6 +2483,22 @@ def exec_before_job(self, app, inp_data, out_data, param_dict=None):
...
 
+    def exec_after_process(self, app, inp_data, out_data, param_dict, job=None, **kwds):
+        for key, val in self.outputs.items():
+            if val.output_type == "data":
+                with open(out_data[key].file_name, "r") as f:
+                    src = json.load(f)
+                assert isinstance(src, dict)
+                dataset_id = src["id"]
+                copy_object = None
+                for _, input_dataset in inp_data.items():
```
nsoranzo:
```suggestion
                for input_dataset in inp_data.values():
```

lib/galaxy/model/__init__.py
mvdbeek: We only do this in the `__init__` of `HistoryDatasetAssociation` because metadata can be None there.

lib/galaxy/model/__init__.py
mvdbeek: Ah, I thought we used it for cached jobs, but we don't (and we don't need to). Let's not add it then.

mvdbeek (approved):

lib/galaxy/tools/__init__.py
jmchilton: Rebased this change in, thanks!

PR #10703, opened 2020-11-12
Null handling and optional values for integers, floats, and text params.

jmchilton:


dannon:
@jmchilton Will this address the problem I attempted to start poking at in https://github.com/galaxyproject/galaxy/pull/9484?  That'd be awesome if so.

jmchilton:
@dannon I don't think this would fix that, but I think tightening up our testing around types moves us in that direction and the new ``value_json`` inputs and outputs could definitely help try to recreate that issue in the tool form (it might just be workflow thing though). I did rebase to add this test case:

```
+        <test>
+            <param name="text_input" value_json="&quot;null&quot;" />
+            <output name="text_out" value_json="&quot;null&quot;" />
+        </test>
```

... thinking it was going to ruin my day, but it did in fact work - which was utterly shocking.

dannon:
@jmchilton Thanks!  I figured it might be a long shot; I don't think I've ever reproduced that in the regular tool form -- it seemed to be related to double-decoding the payload the workflow endpoint receives.

mvdbeek (approved):

mvdbeek:
That does look related:
```
galaxy.tool_util.verify.interactor.JobOutputsError: Output text_out: Expected object () does not match actual object (__oc____sq__src__sq__: __sq__hda__sq__, __sq__id__sq__: __sq__b114bfef23b8f61b__sq____cc__)
```

test/functional/tools/expression_null_handling_float.xml
```diff
@@ -0,0 +1,22 @@
...
+    </expression>
+    <inputs>
+        <param type="float" label="Float input." name="float_input" optional="true" />
+    </inputs>
+    <outputs>
+        <output type="float" name="float_out" from="float_out" />
+    </outputs>
+    <tests>
+        <test>
+            <param name="float_input" value_json="7.2" />
```
bernt-matthias, line 14: Never have seen this attribute `value_json` .. worth documenting?

mvdbeek: It was documented in https://github.com/galaxyproject/galaxy/pull/10694/

jmchilton:
Those work locally, it must be an issue with how we are using postgres JSON  😢

PR #10704, opened 2020-11-12
[20.01] Bump cryptography dependency to 3.2.1

mvdbeek:


nsoranzo (approved):

mvdbeek:
Fun, a deprecation warning is leaking through to the job stderr.

lib/galaxy/metadata/set_metadata.py
```diff
@@ -15,6 +15,12 @@
 import os
 import sys
 import traceback
+import warnings
+
+# Warning on stderr will cause some jobs to be failed.
+# This silences deprecation warnings in the cryptography module.
+if sys.version_info.major <= 3 and sys.version_info.minor < 6:
```
nsoranzo: Should we make this unconditional, i.e. remove this line?

mvdbeek: We could, but this isn't the right fix, unfortunately.

mvdbeek:
API, Selenium and framework tests passed, but Jenkins didn't update the PR status.

nsoranzo (approved):

PR #10712, opened 2020-11-12
Implement expression tool for picking non-null values for CWL-style merging of conditional branches in native Galaxy workflows

jmchilton:
I have CWL conditional branching working down stream (https://github.com/common-workflow-language/galaxy/pull/123/commits/074175b04b45cb70e556ceb3ff0bd29f5c9131bb) - CWL describes branching by simply attaching an expression to steps of the workflow and replace all the outputs with null values if the expression does not evaluate to ``true``. This will work very naturally in Galaxy workflows and the Galaxy workflow editor.

The matching CWL syntax for merging branches is simply to pick non-null values from various branches to continue the workflow with. The tool contained in this PR is a very natural, native Galaxy way of doing that.

This tool could be used to build some cool conditional logic with collections without even needing the forthcoming 'when' expressions on workflow steps. Imagine for instance, we added a "Replace Failed" (or "Replace Empty") collection operation to match "Filter Failed" (or "Filter Empty") that could replace collection elements with 'null' values or a supplied default. We could then run two different algorithms across a list (e.g. one that works well for small datasets but fails out for large ones and one that works well with large datasets) and then run "Replace Failed" with null values and then map both results over this tool. Using purely data flow constructs you've picked the right tool for each element of a list.

I haven't added this to the default tool panel yet though, might be worth waiting for either those when branching expressions or more collection operation tool support.

### DIFF
--- a/tools/expression_tools/pick_value.xml
+++ b/tools/expression_tools/pick_value.xml
@@ -1,0 +1,372 @@
+<tool name="Pick parameter value" id="pick_value" version="0.1.0" tool_type="expression">
+    <macros>
+        <xml name="booleans">
+            <repeat name="pick_from" title="Pick from">
+                <param name="value" type="boolean" optional="true" label="Value" />
+            </repeat>
+        </xml>
+        <xml name="integers">
+            <repeat name="pick_from" title="Pick from">
+                <param name="value" type="integer" label="Value" optional="true" />
+            </repeat>
+        </xml>
+        <xml name="floats">
+            <repeat name="pick_from" title="Pick from">
+                <param name="value" type="float" label="Value" optional="true" />
+            </repeat>
+        </xml>
+        <xml name="texts">
+            <repeat name="pick_from" title="Pick from">
+                <param name="value" type="text" optional="true" label="Value" />
+            </repeat>
+        </xml>
+        <xml name="datas">
+            <repeat name="pick_from" title="Pick from">
+                <param name="value" type="data" format="data" optional="true" label="Value" />
+            </repeat>
+        </xml>
+        <xml name="param_type">
+            <param name="param_type" type="select" label="Select type of parameter to select from">
+                <option value="data">Dataset</option>
+                <option value="text">Text</option>
+                <option value="integer">Integer</option>
+                <option value="float">Float</option>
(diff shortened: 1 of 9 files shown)


jmchilton:
Ugh... again with the postgres JSON failures. The two tests that pass ``null``  in for a data parameter fail under Postgres but seem to work fine locally in sqlite.

mvdbeek:
https://github.com/galaxyproject/galaxy/pull/10712/commits/31ff7497c4c67c733878a27e3ca7634fe7ee1ada fixed the failing tests for me, let's hope the CI agrees.

tools/expression_tools/pick_value.xml
```diff
@@ -0,0 +1,372 @@
...
+                <option value="float">Float</option>
+                <option value="boolean">Boolean</option>
+            </param>
+        </xml>
+        <xml name="param_pick_style">
+            <param name="pick_style" type="select" label="Picking behavior" help="How should the case when not exactly one value is non-null be handled?">
+                <option value="first">Pick first value</option>
+                <option value="first_or_default">Pick first value (or provide default)</option>
+                <option value="first_or_error">Pick first value (or fail)</option>
+                <option value="only">Pick only value</option>
```
mvdbeek, line 42:
```suggestion
                <option value="first">Pick first non-null value</option>
                <option value="first_or_default">Pick first non-null value (or provide default)</option>
                <option value="first_or_error">Pick first non-null value (or fail)</option>
                <option value="single_or_fail">Require single non-null value (or fail)</option>
```
that seems a little easier to understand (for me, anyway).

mvdbeek:
Hmm
```
galaxy.tool_util.verify.interactor.JobOutputsError: Output text_out: Expected object () does not match actual object (__oc____sq__src__sq__: __sq__hda__sq__, __sq__id__sq__: __sq__b114bfef23b8f61b__sq____cc__)
```
that one passed locally for me, with postgres.

mvdbeek:
I've commented the test again, I think it's not exactly related to this PR anyway.

mvdbeek (approved):

mvdbeek:
@galaxybot test this

PR #10714, opened 2020-11-13
Move k8s job manifest to template

nuwang:
This is just some preliminary work, and doesn't really work, but I thought I'd share it based on the discussion at the round-table.
@innovate-invent I believe you had a more comprehensive set of ideas here, so if you want to create a new PR and do the work there, I'm happy to just close this. Just thought I'd share this in case say the template is of any use.

It also emerged that it may be better to move the templating to Pulsar.


mvdbeek:
I've pushed this to 21.05, but this looks cool, please keep going :).

innovate-invent:
This was halted with the idea that it was moving to Pulsar. I am not sure when I will be available to work on this, but I do intend to.

PR #10715, opened 2020-11-13
add slash command to trigger IUC CI

bernt-matthias:
in conjunction with https://github.com/galaxyproject/tools-iuc/pull/3298

### DIFF
--- a/.github/workflows/slash.yaml
+++ b/.github/workflows/slash.yaml
@@ -1,0 +1,15 @@
+name: Slash Command Dispatch
+on:
+  issue_comment:
+    types: [created]
+jobs:
+  slashCommandDispatch:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


mvdbeek:
I think having just the tools-iuc repo listen is probably enough ? The permissions needed to do this are pretty wide unfortunately.

bernt-matthias:
> I think having just the tools-iuc repo listen is probably enough ?

Guess so.

bernt-matthias:
> I think having just the tools-iuc repo listen is probably enough ? The permissions needed to do this are pretty wide unfortunately.

https://github.com/peter-evans/slash-command-dispatch#token "f you will be dispatching commands to public repositories only then you can use the more limited public_repo scope." .. not sure yet what that means - but sounds helpful

mvdbeek:
We can add this at any point, so I dropped the milestone here.

PR #10716, opened 2020-11-13
Re-sync galaxy.tool_util.cwl with CWL branch of Galaxy.

jmchilton:


lib/galaxy/tool_util/cwl/schema.py
```diff
@@ -35,16 +36,34 @@ def loading_context(self):
...
-            from cwl_utils import etools_to_clt
-            etools_to_clt.main([path, galaxy_path])
-            galaxy_uri = "file://" + galaxy_path
-            uri = galaxy_uri
-        loading_context, process_object, uri = load_tool.fetch_document(uri, loadingContext=loading_context)
+        with tempfile.TemporaryDirectory() as output_dir:
+            suffix = ''
+            if '#' in path:
+                path, suffix = path.split('#')
+            print(f""" -------
```
nsoranzo, line 43: @jmchilton @mvdbeek Should we get rid of these debugging prints or is it too early?

N.B.: that's in the main galaxy repo not the cwl branch.

PR #10720, opened 2020-11-13
Ensure build_for_rerun output is accepted by the tool run API and add remappable method to job model

### LINKED ISSUE #10690: build_for_rerun API does not work with conditional or repeat input parameters
I discovered this here: https://github.com/galaxyproject/bioblend/pull/343#issuecomment-724773705

I can run the `random_lines1` tool correctly by submitting the following dict with the `tool_inputs`:
```
{'num_lines': '1', 'input': {'src': 'hda', 'id': '11ac94870d0bb33a453c98685a11a16c'}, 'seed_source|seed_source_selector': 'set_seed', 'seed_source|seed': 'asdf'}
```

But if I try and get this dict back using `build_for_rerun` the `seed_source` params have changed:

```
{'num_lines': '1', 'input': {'values': [{'id': '11ac94870d0bb33a453c98685a11a16c', 'src': 'hda'}]}, 'seed_source': {'seed_source_selector': 'set_seed', '__current_case__': 1, 'seed': 'asdf'}}
```

and if I try and rerun with the new dict the `seed_source` params do not get set.

image

simonbray:
Fixes https://github.com/galaxyproject/galaxy/issues/10690.

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -1269,8 +1269,8 @@ class Job(JobLike, UsesCreateAndUpdateTime, Dictifiable, RepresentById):
                 elif self.output_dataset_collection_instances:
                     # We'll want to replace this item
                     return 'job_produced_collection_elements'
-            except Exception as exception:
-                log.error(str(exception))
+            except Exception:
+                log.error(f"Error trying to determine if job {{self.id}} is remappable")
         return False
 
 
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -1260,13 +1260,14 @@ class Job(JobLike, UsesCreateAndUpdateTime, Dictifiable, RepresentById):
 
     def remappable(self):
         """
-        Find if job is remappable when rerun
+        Check whether job is remappable when rerun
         """
         if self.state == self.states.ERROR:
             try:
-                if [hda.dependent_jobs for hda in [jtod.dataset for jtod in self.output_datasets] if hda.dependent_jobs]:
-                    return True
-                elif self.output_dataset_collection_instances:
+                for jtod in self.output_datasets:
+                    if jtod.dataset.dependent_jobs:
+                        return True
+                if self.output_dataset_collection_instances:
                     # We'll want to replace this item
                     return 'job_produced_collection_elements'
             except Exception:
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -1270,7 +1270,7 @@ class Job(JobLike, UsesCreateAndUpdateTime, Dictifiable, RepresentById):
                     # We'll want to replace this item
                     return 'job_produced_collection_elements'
             except Exception:
-                log.error(f"Error trying to determine if job {{self.id}} is remappable")
+                log.error(f"Error trying to determine if job {self.id} is remappable")
         return False
 
 
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -1271,7 +1271,7 @@ class Job(JobLike, UsesCreateAndUpdateTime, Dictifiable, RepresentById):
                     # We'll want to replace this item
                     return 'job_produced_collection_elements'
             except Exception:
-                log.error(f"Error trying to determine if job {self.id} is remappable")
+                log.exception(f"Error trying to determine if job {self.id} is remappable")
         return False
 
 
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -1135,6 +1135,8 @@ class Job(JobLike, UsesCreateAndUpdateTime, Dictifiable, RepresentById):
                     }
             rval['outputs'] = output_dict
 
+            rval['remappable'] = self.remappable()
+
         return rval
 
     def set_final_state(self, final_state):
@@ -1257,6 +1259,21 @@ class Job(JobLike, UsesCreateAndUpdateTime, Dictifiable, RepresentById):
         }
         for statement in statements:
             sa_session.execute(statement, params)
+
+    def remappable(self):
+        """
+        Find if job is remappable when rerun
+        """
+        if self.state == self.states.ERROR:
+            try:
+                if [hda.dependent_jobs for hda in [jtod.dataset for jtod in self.output_datasets] if hda.dependent_jobs]:
+                    return True
+                elif self.output_dataset_collection_instances:
+                    # We'll want to replace this item
+                    return 'job_produced_collection_elements'
+            except Exception as exception:
+                log.error(str(exception))
+        return False
 
 
 class Task(JobLike, RepresentById):
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -1135,8 +1135,6 @@ class Job(JobLike, UsesCreateAndUpdateTime, Dictifiable, RepresentById):
                     }
             rval['outputs'] = output_dict
 
-            rval['remappable'] = self.remappable()
-
         return rval
 
     def set_final_state(self, final_state):
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -2252,16 +2252,7 @@ class Tool(Dictifiable):
 
     def _get_job_remap(self, job):
         if job:
-            if job.state == job.states.ERROR:
-                try:
-                    if [hda.dependent_jobs for hda in [jtod.dataset for jtod in job.output_datasets] if hda.dependent_jobs]:
-                        return True
-                    elif job.output_dataset_collection_instances:
-                        # We'll want to replace this item
-                        return 'job_produced_collection_elements'
-                except Exception as exception:
-                    log.error(str(exception))
-        return False
+            return job.remappable()
 
     def _map_source_to_history(self, trans, tool_inputs, params):
         # Need to remap dataset parameters. Job parameters point to original
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -1526,7 +1526,7 @@ class Tool(Dictifiable):
         if self.check_values:
             visit_input_values(self.inputs, values, callback)
 
-    def expand_incoming(self, trans, incoming, request_context):
+    def expand_incoming(self, trans, incoming, request_context, legacy_param_form=True):
         rerun_remap_job_id = None
         if 'rerun_remap_job_id' in incoming:
             try:
@@ -1567,7 +1567,7 @@ class Tool(Dictifiable):
             else:
                 # Update state for all inputs on the current page taking new
                 # values from `incoming`.
-                populate_state(request_context, self.inputs, expanded_incoming, params, errors, simple_errors=False)
+                populate_state(request_context, self.inputs, expanded_incoming, params, errors, simple_errors=False, legacy_param_form=legacy_param_form)
                 # If the tool provides a `validate_input` hook, call it.
                 validate_input = self.get_hook('validate_input')
                 if validate_input:
@@ -1579,7 +1579,7 @@ class Tool(Dictifiable):
         log.info(validation_timer)
         return all_params, all_errors, rerun_remap_job_id, collection_info
 
-    def handle_input(self, trans, incoming, history=None, use_cached_job=False):
+    def handle_input(self, trans, incoming, history=None, use_cached_job=False, legacy_param_form=True):
         """
         Process incoming parameters for this tool from the dict `incoming`,
         update the tool state (or create if none existed), and either return
@@ -1587,7 +1587,7 @@ class Tool(Dictifiable):
         there were no errors).
         """
         request_context = WorkRequestContext(app=trans.app, user=trans.user, history=history or trans.history)
-        all_params, all_errors, rerun_remap_job_id, collection_info = self.expand_incoming(trans=trans, incoming=incoming, request_context=request_context)
+        all_params, all_errors, rerun_remap_job_id, collection_info = self.expand_incoming(trans=trans, incoming=incoming, request_context=request_context, legacy_param_form=legacy_param_form)
         # If there were errors, we stay on the same page and display them
         if any(all_errors):
             # simple param_key -> message string for tool form.
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -1526,7 +1526,7 @@ class Tool(Dictifiable):
         if self.check_values:
             visit_input_values(self.inputs, values, callback)
 
-    def expand_incoming(self, trans, incoming, request_context, legacy_param_form=True):
+    def expand_incoming(self, trans, incoming, request_context, input_format='legacy'):
         rerun_remap_job_id = None
         if 'rerun_remap_job_id' in incoming:
             try:
@@ -1567,7 +1567,7 @@ class Tool(Dictifiable):
             else:
                 # Update state for all inputs on the current page taking new
                 # values from `incoming`.
-                populate_state(request_context, self.inputs, expanded_incoming, params, errors, simple_errors=False, legacy_param_form=legacy_param_form)
+                populate_state(request_context, self.inputs, expanded_incoming, params, errors, simple_errors=False, input_format=input_format)
                 # If the tool provides a `validate_input` hook, call it.
                 validate_input = self.get_hook('validate_input')
                 if validate_input:
@@ -1579,7 +1579,7 @@ class Tool(Dictifiable):
         log.info(validation_timer)
         return all_params, all_errors, rerun_remap_job_id, collection_info
 
-    def handle_input(self, trans, incoming, history=None, use_cached_job=False, legacy_param_form=True):
+    def handle_input(self, trans, incoming, history=None, use_cached_job=False, input_format='legacy'):
         """
         Process incoming parameters for this tool from the dict `incoming`,
         update the tool state (or create if none existed), and either return
@@ -1587,7 +1587,7 @@ class Tool(Dictifiable):
         there were no errors).
         """
         request_context = WorkRequestContext(app=trans.app, user=trans.user, history=history or trans.history)
-        all_params, all_errors, rerun_remap_job_id, collection_info = self.expand_incoming(trans=trans, incoming=incoming, request_context=request_context, legacy_param_form=legacy_param_form)
+        all_params, all_errors, rerun_remap_job_id, collection_info = self.expand_incoming(trans=trans, incoming=incoming, request_context=request_context, input_format=input_format)
         # If there were errors, we stay on the same page and display them
         if any(all_errors):
             # simple param_key -> message string for tool form.
@@ -2194,7 +2194,7 @@ class Tool(Dictifiable):
             'tool_errors'   : self.tool_errors,
             'state_inputs'  : params_to_strings(self.inputs, state_inputs, self.app, use_security=True, nested=True),
             'job_id'        : trans.security.encode_id(job.id) if job else None,
-            'job_remap'     : self._get_job_remap(job),
+            'job_remap'     : job.remappable() if job else None,
             'history_id'    : trans.security.encode_id(history.id) if history else None,
             'display'       : self.display_interface,
             'action'        : action,
@@ -2249,10 +2249,6 @@ class Tool(Dictifiable):
                 group_inputs.append(tool_dict)
             else:
                 group_inputs[input_index] = tool_dict
-
-    def _get_job_remap(self, job):
-        if job:
-            return job.remappable()
 
     def _map_source_to_history(self, trans, tool_inputs, params):
         # Need to remap dataset parameters. Job parameters point to original
--- a/lib/galaxy/tools/parameters/__init__.py
+++ b/lib/galaxy/tools/parameters/__init__.py
@@ -271,7 +271,7 @@ def update_dataset_ids(input_values, translate_values, src):
     return remap(input_values, visit=replace_dataset_ids)
 
 
-def populate_state(request_context, inputs, incoming, state, errors={}, prefix='', context=None, check=True, simple_errors=True, legacy_param_form=True):
+def populate_state(request_context, inputs, incoming, state, errors={}, context=None, check=True, simple_errors=True, input_format='legacy'):
     """
     Populates nested state dict from incoming parameter values.
     >>> from collections import OrderedDict
@@ -312,10 +312,10 @@ def populate_state(request_context, inputs, incoming, state, errors={}, prefix='
     3
     >>> print(state['b'][0]['d'][0]['f']['h'])
     4
-    >>> # now test with legacy_param_form=False
+    >>> # now test with input_format='21.01'
     >>> flat_new = {'a': 1, 'b': [{'c': 2, 'd': [{'e': 3, 'f': {'h': 4, 'g': True}}]}]}
     >>> state_new = OrderedDict()
-    >>> populate_state(trans, inputs, flat_new, state_new, check=False, legacy_param_form=False)
+    >>> populate_state(trans, inputs, flat_new, state_new, check=False, input_format='21.01')
     >>> print(state_new['a'])
     1
     >>> print(state_new['b'][0]['c'])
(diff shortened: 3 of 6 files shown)


nsoranzo:
@galaxybot test this

lib/galaxy/webapps/galaxy/api/jobs.py
```diff
@@ -357,7 +358,13 @@ def build_for_rerun(self, trans, id, **kwd):
...
         if not tool.is_workflow_compatible:
             raise exceptions.ConfigDoesNotAllowException("Tool '%s' cannot be rerun." % (job.tool_id))
-        return tool.to_json(trans, {}, job=job)
+
+        # ensure state_inputs are formatted for rerunning with params_to_incoming
+        rerun_params = tool.to_json(trans, {}, job=job)
+        state_inputs = {}
+        params_to_incoming(state_inputs, tool.inputs, rerun_params['state_inputs'], tool.app)
+        rerun_params['state_inputs'] = state_inputs
+        return rerun_params
```
mvdbeek: Can you do this the other way round, have the payload in https://github.com/galaxyproject/bioblend/pull/343#issuecomment-724773705 be accepted for running tools ?

simonbray: Yep - wasn't sure which way was better

nsoranzo: @mvdbeek To be clear: you mean change the Galaxy tool API to accept also the format returned by `/api/jobs/{id}/build_for_rerun` , right?

lib/galaxy/model/__init__.py
```diff
@@ -1135,6 +1135,8 @@ def to_dict(self, view='collection', system_details=False):
                     }
             rval['outputs'] = output_dict
 
+            rval['remappable'] = self.remappable()
```
simonbray: Not related to the build_for_rerun stuff but this would be handy for the rerun bioblend method

lib/galaxy/webapps/galaxy/api/jobs.py
mvdbeek: Right!

lib/galaxy/model/__init__.py
mvdbeek: This is likely an expensive method that we shouldn't call if it's not needed.
If necessary you could add this either in another view or as a separate endpoint maybe ?

lib/galaxy/webapps/galaxy/api/tools.py
```diff
@@ -511,9 +512,14 @@ def _create(self, trans, payload, **kwd):
...
         # Set up inputs.
-        inputs = payload.get('inputs', {})
-        if not isinstance(inputs, dict):
-            raise exceptions.RequestParameterInvalidException("inputs invalid %s" % inputs)
+        submitted_inputs = payload.get('inputs', {})
+        if not isinstance(submitted_inputs, dict):
+            raise exceptions.RequestParameterInvalidException("inputs invalid %s" % submitted_inputs)
+        try:
+            inputs = {}
+            params_to_incoming(inputs, tool.inputs, submitted_inputs, tool.app)
```
mvdbeek: Can you add an API test for this ? You could modify https://github.com/mvdbeek/galaxy/blob/a2c83ee7224b181c39ed394be7a866aaf88a46c1/lib/galaxy_test/api/test_tools.py#L1550 for instance. I am also thinking that we might want to hide this behind a parameter or a new endpoint. The parameter could be called `legacy_config_form_payload` and default to `True`. Or we add a `api/tools/v2` endpoint that just adds this parameter.

simonbray:
Is it alright to target this to release_20.09?

lib/galaxy/model/__init__.py
```diff
@@ -1258,6 +1258,21 @@ def update_output_states(self):
...
+        """
+        if self.state == self.states.ERROR:
+            try:
+                if [hda.dependent_jobs for hda in [jtod.dataset for jtod in self.output_datasets] if hda.dependent_jobs]:
+                    return True
+                elif self.output_dataset_collection_instances:
+                    # We'll want to replace this item
+                    return 'job_produced_collection_elements'
+            except Exception as exception:
+                log.error(str(exception))
```
nsoranzo:
```suggestion
            except Exception:
                log.exception(f"Error trying to determine if job {{self.id}} is remappable")
```

lib/galaxy_test/api/test_tools.py
```diff
@@ -1589,14 +1590,39 @@ def test_identifier_map_over_multiple_input_in_conditional(self, history_id):
...
         self.assertEqual(output1_content.strip(), "forward\nreverse")
 
+    @skip_without_tool("identifier_in_conditional")
+    @uses_test_history(require_new=False)
+    def test_identifier_map_over_multiple_input_in_conditional_new_payload_form(self, history_id):
+        hdca_id = self._build_pair(history_id, ["123", "456"])
+        inputs = {
+            "outer_cond": {
+                'multi_input': True,
+                '__current_case__': 0,
```
mvdbeek: Is this necessary for the tests to pass ?

simonbray: Yes, but I'm not sure why.

mvdbeek: 😿  I was hoping we'd get a nicer input format out of this. I'll have a look if we can do something about this.

simonbray: To be a bit more helpful: leaving them out causes a KeyError here https://github.com/galaxyproject/galaxy/blob/ca442596c344742c2a7f64d8d0bedac41536c88b/lib/galaxy/tools/parameters/__init__.py#L245

and here

https://github.com/galaxyproject/galaxy/blob/ca442596c344742c2a7f64d8d0bedac41536c88b/lib/galaxy/tools/parameters/__init__.py#L247

mvdbeek: Ideally we wouldn't need to create `incoming`, the dictionary should map cleanly onto parameters, so what I would try is replacing populate_state (https://github.com/mvdbeek/galaxy/blob/dev/lib/galaxy/tools/parameters/__init__.py#L274) with a simpler version that works with a clean dictionary. That's pretty deep in Galaxy's internals though, and I'm not sure it would actually work ... but if it did that would be amazing.
Let me know if you're up for giving this a try, otherwise I'm happy to take a look.

lib/galaxy/webapps/galaxy/api/tools.py
```diff
@@ -511,9 +515,14 @@ def _create(self, trans, payload, **kwd):
             target_history = None
 
         # Set up inputs.
-        inputs = payload.get('inputs', {})
-        if not isinstance(inputs, dict):
-            raise exceptions.RequestParameterInvalidException("inputs invalid %s" % inputs)
+        submitted_inputs = payload.get('inputs', {})
+        if not isinstance(submitted_inputs, dict):
+            raise exceptions.RequestParameterInvalidException("inputs invalid %s" % submitted_inputs)
```
nsoranzo:
```suggestion
            raise exceptions.RequestParameterInvalidException(f"inputs invalid {{submitted_inputs}}")
```

lib/galaxy/webapps/galaxy/api/tools.py
nsoranzo: Cannot we safely guess the format of `submitted_inputs` by checking if any key contains `|` ?

mvdbeek: I'm not sure we can, if there is no `|` in an incoming payload it could be the old format, but there's no conditional, or it could be the new format with nested dictionaries. And I prefer to keep this separate, since we probably want to take a different route through the code in the future.

lib/galaxy_test/api/test_tools.py
simonbray: I don't mind having a try.

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -312,7 +312,82 @@ def populate_state(request_context, inputs, incoming, state, errors={}, prefix='
...
+                for rep_index, rep in enumerate(incoming[input.name]):
+                    new_state = {'__index__' : rep_index}
+                    group_state.append(new_state)
+                    populate_state(request_context, input.inputs, rep, new_state, errors, context=context, check=check, simple_errors=simple_errors, legacy_param_form=False)
+
+        elif input.type == 'conditional':
+            # if input.value_ref and not input.value_ref_in_group:
+            #     test_param_key = prefix + input.test_param.name
+            # else:
+            #     test_param_key = group_prefix + input.test_param.name
```
simonbray: I don't understand what this `value_ref` attribute does. Is it safe to remove these lines?

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -312,7 +312,82 @@ def populate_state(request_context, inputs, incoming, state, errors={}, prefix='
...
+                    populate_state(request_context, input.cases[current_case].inputs, incoming.get(input.name), group_state, errors, context=context, check=check, simple_errors=simple_errors, legacy_param_form=False)
+                    group_state['__current_case__'] = current_case
+                except Exception:
+                    errors[input.test_param.name] = 'The selected case is unavailable/invalid.'
+            group_state[input.test_param.name] = value
+
+        elif input.type == 'section':
+            populate_state(request_context, input.inputs, incoming.get(input.name), group_state, errors, context=context, check=check, simple_errors=simple_errors)
+
+        elif input.type == 'upload_dataset':
```
simonbray: Also not sure how an input can have `upload_dataset` as a type.

simonbray:
Phew, testing looks a bit better after the last commit.

@mvdbeek, maybe you can have another look at this? It's not mergeable yet, but the new `populate_state` seems to be working, at least for tool execution with the API.

lib/galaxy/model/__init__.py
```diff
@@ -1258,6 +1258,21 @@ def update_output_states(self):
         for statement in statements:
             sa_session.execute(statement, params)
 
+    def remappable(self):
+        """
+        Find if job is remappable when rerun
```
nsoranzo:
```suggestion
        Check whether job is remappable when rerun
```

lib/galaxy/model/__init__.py
```diff
@@ -1258,6 +1258,21 @@ def update_output_states(self):
...
+        """
+        if self.state == self.states.ERROR:
+            try:
+                if [hda.dependent_jobs for hda in [jtod.dataset for jtod in self.output_datasets] if hda.dependent_jobs]:
+                    return True
```
nsoranzo: I think this is equivalent and more readable?
```suggestion
                for jtod in self.output_datasets:
                    if jtod.dataset.dependent_jobs:
                        return True
```

lib/galaxy/webapps/galaxy/api/tools.py
```diff
@@ -532,7 +535,10 @@ def _create(self, trans, payload, **kwd):
         # I think it should be a top-level parameter, but because the selector is implemented
         # as a regular tool parameter we accept both.
         use_cached_job = payload.get('use_cached_job', False) or util.string_as_bool(inputs.get('use_cached_job', 'false'))
-        vars = tool.handle_input(trans, incoming, history=target_history, use_cached_job=use_cached_job)
+
+        legacy_param_form = kwd.get('legacy_config_form_payload') not in [False, 'False']
```
nsoranzo: You can use `string_as_bool` from `galaxy.util`

lib/galaxy/webapps/galaxy/api/tools.py
```diff
@@ -465,6 +465,9 @@ def create(self, trans, payload, **kwd):
         """
         POST /api/tools
         Execute tool with a given parameter payload
+
+        :param legacy_config_form_payload: whether the payload is specified in the legacy format, i.e. conditionals and repeats are represented with the '|' symbol (default True).
+        :type legacy_config_form_payload:  bool
```
nsoranzo: Maybe this could be a string to allow for a possible further change of format? So for example `input_format` of type `str` with possible values `legacy` (default) and `21.01` .

lib/galaxy/tools/__init__.py
```diff
@@ -2252,16 +2252,7 @@ def populate_model(self, request_context, inputs, state_inputs, group_inputs, ot
...
-                try:
-                    if [hda.dependent_jobs for hda in [jtod.dataset for jtod in job.output_datasets] if hda.dependent_jobs]:
-                        return True
-                    elif job.output_dataset_collection_instances:
-                        # We'll want to replace this item
-                        return 'job_produced_collection_elements'
-                except Exception as exception:
-                    log.error(str(exception))
-        return False
+            return job.remappable()
```
nsoranzo: I'd drop the method and replace its only call with `job.remappable() if job else None`

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -312,7 +312,82 @@ def populate_state(request_context, inputs, incoming, state, errors={}, prefix='
...
+                populate_state(request_context, input.inputs, incoming.get(rep_prefix, []), rep_state, errors, context=context, check=check, simple_errors=simple_errors)
+        else:
+            param_value = _get_incoming_value(incoming, input.name, state.get(input.name))
+            value, error = check_param(request_context, input, param_value, context, simple_errors=simple_errors) if check else [param_value, None]
+            if error:
+                errors[input.name] = error
+            state[input.name] = value
+
+
+def _populate_state_legacy(request_context, inputs, incoming, state, errors={}, prefix='', context=None, check=True, simple_errors=True):
```
nsoranzo: The recursive calls below should call directly `_populate_state_legacy()` instead of `_populate_state()` .

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -271,7 +271,7 @@ def replace_dataset_ids(path, key, value):
     return remap(input_values, visit=replace_dataset_ids)
 
 
-def populate_state(request_context, inputs, incoming, state, errors={}, prefix='', context=None, check=True, simple_errors=True):
+def populate_state(request_context, inputs, incoming, state, errors={}, prefix='', context=None, check=True, simple_errors=True, legacy_param_form=True):
```
nsoranzo:
```suggestion
def populate_state(request_context, inputs, incoming, state, errors={}, context=None, check=True, simple_errors=True, legacy_param_form=True):
```

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -312,7 +312,82 @@ def populate_state(request_context, inputs, incoming, state, errors={}, prefix='
...
+    >>> print(state_new['b'][0]['c'])
+    2
+    >>> print(state_new['b'][0]['d'][0]['e'])
+    3
+    >>> print(state_new['b'][0]['d'][0]['f']['h'])
+    4
+
     """
+    if legacy_param_form:
+        _populate_state_legacy(request_context, inputs, incoming, state, errors, prefix, context, check, simple_errors)
```
nsoranzo, line 330:
```suggestion
        _populate_state_legacy(request_context, inputs, incoming, state, errors, context, check, simple_errors)
```

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -312,7 +312,82 @@ def populate_state(request_context, inputs, incoming, state, errors={}, prefix='
...
+                    current_case = input.get_current_case(value)
+                    group_state = state[input.name] = {}
+                    populate_state(request_context, input.cases[current_case].inputs, incoming.get(input.name), group_state, errors, context=context, check=check, simple_errors=simple_errors, legacy_param_form=False)
+                    group_state['__current_case__'] = current_case
+                except Exception:
+                    errors[input.test_param.name] = 'The selected case is unavailable/invalid.'
+            group_state[input.test_param.name] = value
+
+        elif input.type == 'section':
+            populate_state(request_context, input.inputs, incoming.get(input.name), group_state, errors, context=context, check=check, simple_errors=simple_errors)
```
nsoranzo:
```suggestion
            populate_state(request_context, input.inputs, incoming.get(input.name), group_state, errors, context=context, check=check, simple_errors=simple_errors, legacy_param_form=False)
```

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -312,7 +312,82 @@ def populate_state(request_context, inputs, incoming, state, errors={}, prefix='
...
+                del group_state[-1]
+            while file_count > len(group_state):
+                new_state = {'__index__' : len(group_state)}
+                for upload_item in input.inputs.values():
+                    new_state[upload_item.name] = upload_item.get_initial_value(request_context, context)
+                group_state.append(new_state)
+            for i, rep_state in enumerate(group_state):
+                rep_index = rep_state['__index__']
+                rep_prefix = '%s_%d|' % (input.name, rep_index)
+                populate_state(request_context, input.inputs, incoming.get(rep_prefix, []), rep_state, errors, context=context, check=check, simple_errors=simple_errors)
```
nsoranzo:
```suggestion
                populate_state(request_context, input.inputs, incoming.get(rep_prefix, []), rep_state, errors, context=context, check=check, simple_errors=simple_errors, legacy_param_form=False)
```

simonbray:
Many thanks for the review @nsoranzo - I added all your suggestions.

Do you have any idea about these two questions:
> - I don't understand what this value_ref attribute does. Is it safe to remove these lines?
> - Also not sure how an input can have upload_dataset as a type.

mvdbeek:
> * I don't understand what this value_ref attribute does. Is it safe to remove these lines?
> * Also not sure how an input can have upload_dataset as a type.

- value_ref can point to parameters outside of a conditional
- I think that would be the case for the history import (hidden) tool

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -312,10 +312,90 @@ def populate_state(request_context, inputs, incoming, state, errors={}, prefix='
...
+        for input in inputs.values():
+            state[input.name] = input.get_initial_value(request_context, context)
+            group_state = state[input.name]
+            if input.type == 'repeat':
+                if len(incoming[input.name]) > input.max or len(incoming[input.name]) < input.min:
+                    errors[input.name] = 'The number of repeat elements is outside the range specified by the tool.'
+                else:
+                    del group_state[:]
+                    for rep_index, rep in enumerate(incoming[input.name]):
+                        new_state = {'__index__' : rep_index}
```
mvdbeek: We should be able to drop `__index__`. This was for backwards compatibility (8 years ago!), according to https://github.com/mvdbeek/galaxy/blob/cdad7c79a9826bbb9273d5776cdcb264da8e0e1e/lib/galaxy/tools/parameters/grouping.py#L111

lib/galaxy/webapps/galaxy/api/tools.py
```diff
@@ -465,6 +465,9 @@ def create(self, trans, payload, **kwd):
         """
         POST /api/tools
         Execute tool with a given parameter payload
+
+        :param input_format: input format for the payload. Options are the default 'legacy' (i.e. conditionals and repeats are represented with the '|' symbol) or '21.01).
```
nsoranzo:
```suggestion
        :param input_format: input format for the payload. Possible values are the default 'legacy' (where inputs nested inside conditionals or repeats are identified with e.g. '<conditional_name>|<input_name>') or '21.01' (where inputs inside conditionals or repeats are nested elements).
```

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -312,10 +312,90 @@ def populate_state(request_context, inputs, incoming, state, errors={}, prefix='
...
+    else:
+        raise Exception(f'Input format {input_format} not recognized; input_format must be either legacy or 21.01.')
+
+
+def _populate_state_legacy(request_context, inputs, incoming, state, errors={}, prefix='', context=None, check=True, simple_errors=True):
     context = ExpressionContext(state, context)
     for input in inputs.values():
         state[input.name] = input.get_initial_value(request_context, context)
+        print(prefix)
+        print(input.name)
```
nsoranzo:
```suggestion
```

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -312,10 +312,90 @@ def populate_state(request_context, inputs, incoming, state, errors={}, prefix='
...
+    >>> print(state_new['b'][0]['d'][0]['e'])
+    3
+    >>> print(state_new['b'][0]['d'][0]['f']['h'])
+    4
+
     """
+    if input_format == 'legacy':
+        prefix = ''
+        _populate_state_legacy(request_context, inputs, incoming, state, errors, prefix, context, check, simple_errors)
+        return
```
nsoranzo:
```suggestion
        return _populate_state_legacy(request_context, inputs, incoming, state, errors=errors, context=context, check=check, simple_errors=simple_errors)
```

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -368,18 +363,8 @@ def populate_state(request_context, inputs, incoming, state, errors={}, context=
...
-                    new_state = {'__index__' : len(group_state)}
-                    for upload_item in input.inputs.values():
-                        new_state[upload_item.name] = upload_item.get_initial_value(request_context, context)
-                    group_state.append(new_state)
-                for i, rep_state in enumerate(group_state):
-                    rep_index = rep_state['__index__']
-                    rep_prefix = '%s_%d|' % (input.name, rep_index)
-                    populate_state(request_context, input.inputs, incoming.get(rep_prefix, []), rep_state, errors, context=context, check=check, simple_errors=simple_errors, input_format='21.01')
+                raise NotImplementedError
+
```
simonbray, line 378: Sorry, I would rather not work on this. The scope of this PR already got way bigger than I expected.

lib/galaxy/model/__init__.py
```diff
@@ -1258,6 +1258,22 @@ def update_output_states(self):
...
+        if self.state == self.states.ERROR:
+            try:
+                for jtod in self.output_datasets:
+                    if jtod.dataset.dependent_jobs:
+                        return True
+                if self.output_dataset_collection_instances:
+                    # We'll want to replace this item
+                    return 'job_produced_collection_elements'
+            except Exception:
+                log.error(f"Error trying to determine if job {self.id} is remappable")
```
nsoranzo:
```suggestion
                log.exception(f"Error trying to determine if job {self.id} is remappable")
```

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -312,7 +312,70 @@ def populate_state(request_context, inputs, incoming, state, errors={}, prefix='
     3
     >>> print(state['b'][0]['d'][0]['f']['h'])
     4
+    >>> # now test with input_format='21.01'
+    >>> flat_new = {'a': 1, 'b': [{'c': 2, 'd': [{'e': 3, 'f': {'h': 4, 'g': True}}]}]}
+    >>> state_new = OrderedDict()
+    >>> populate_state(trans, inputs, flat_new, state_new, check=False, input_format='21.01')
```
nsoranzo:
```suggestion
    >>> nested = {'a': 1, 'b': [{'c': 2, 'd': [{'e': 3, 'f': {'h': 4, 'g': True}}]}]}
    >>> state_new = OrderedDict()
    >>> populate_state(trans, inputs, nested, state_new, check=False, input_format='21.01')
```

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -312,7 +312,70 @@ def populate_state(request_context, inputs, incoming, state, errors={}, prefix='
...
+                    errors[input.name] = 'The number of repeat elements is outside the range specified by the tool.'
+                else:
+                    del group_state[:]
+                    for rep_index, rep in enumerate(incoming[input.name]):
+                        new_state = {}
+                        group_state.append(new_state)
+                        populate_state(request_context, input.inputs, rep, new_state, errors, context=context, check=check, simple_errors=simple_errors, input_format='21.01')
+
+            elif input.type == 'conditional':
+                test_param_value = incoming.get(input.name).get(input.test_param.name)
```
nsoranzo:
```suggestion
                test_param_value = incoming.get(input.name, {}).get(input.test_param.name)
```

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -271,7 +271,7 @@ def replace_dataset_ids(path, key, value):
     return remap(input_values, visit=replace_dataset_ids)
 
 
-def populate_state(request_context, inputs, incoming, state, errors={}, prefix='', context=None, check=True, simple_errors=True):
+def populate_state(request_context, inputs, incoming, state, errors={}, context=None, check=True, simple_errors=True, input_format='legacy'):
```
nsoranzo:
```suggestion
def populate_state(request_context, inputs, incoming, state, errors=None, context=None, check=True, simple_errors=True, input_format='legacy'):
```

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -312,7 +312,70 @@ def populate_state(request_context, inputs, incoming, state, errors={}, prefix='
...
+    1
+    >>> print(state_new['b'][0]['c'])
+    2
+    >>> print(state_new['b'][0]['d'][0]['e'])
+    3
+    >>> print(state_new['b'][0]['d'][0]['f']['h'])
+    4
+
     """
+    if input_format == 'legacy':
```
nsoranzo:
```suggestion
    """
    if errors is None:
        errors = {}
    if input_format == 'legacy':
```

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -312,7 +312,70 @@ def populate_state(request_context, inputs, incoming, state, errors={}, prefix='
...
+                param_value = _get_incoming_value(incoming, input.name, state.get(input.name))
+                value, error = check_param(request_context, input, param_value, context, simple_errors=simple_errors) if check else [param_value, None]
+                if error:
+                    errors[input.name] = error
+                state[input.name] = value
+    else:
+        raise Exception(f'Input format {input_format} not recognized; input_format must be either legacy or 21.01.')
+
+
+def _populate_state_legacy(request_context, inputs, incoming, state, errors={}, prefix='', context=None, check=True, simple_errors=True):
```
nsoranzo:
```suggestion
def _populate_state_legacy(request_context, inputs, incoming, state, errors, prefix='', context=None, check=True, simple_errors=True):
```

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -312,7 +312,70 @@ def populate_state(request_context, inputs, incoming, state, errors={}, prefix='
...
+            group_state = state[input.name]
+            if input.type == 'repeat':
+                if len(incoming[input.name]) > input.max or len(incoming[input.name]) < input.min:
+                    errors[input.name] = 'The number of repeat elements is outside the range specified by the tool.'
+                else:
+                    del group_state[:]
+                    for rep_index, rep in enumerate(incoming[input.name]):
+                        new_state = {}
+                        group_state.append(new_state)
+                        populate_state(request_context, input.inputs, rep, new_state, errors, context=context, check=check, simple_errors=simple_errors, input_format='21.01')
```
nsoranzo:
```suggestion
                        populate_state(request_context, input.inputs, rep, new_state, errors, context=context, check=check, simple_errors=simple_errors, input_format=input_format)
```

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -312,7 +312,70 @@ def populate_state(request_context, inputs, incoming, state, errors={}, prefix='
...
+            elif input.type == 'conditional':
+                test_param_value = incoming.get(input.name).get(input.test_param.name)
+                value, error = check_param(request_context, input.test_param, test_param_value, context, simple_errors=simple_errors) if check else [test_param_value, None]
+                if error:
+                    errors[input.test_param.name] = error
+                else:
+                    try:
+                        current_case = input.get_current_case(value)
+                        group_state = state[input.name] = {}
+                        populate_state(request_context, input.cases[current_case].inputs, incoming.get(input.name), group_state, errors, context=context, check=check, simple_errors=simple_errors, input_format='21.01')
```
nsoranzo:
```suggestion
                        populate_state(request_context, input.cases[current_case].inputs, incoming.get(input.name), group_state, errors, context=context, check=check, simple_errors=simple_errors, input_format=input_format)
```

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -312,7 +312,70 @@ def populate_state(request_context, inputs, incoming, state, errors={}, prefix='
...
+                        group_state['__current_case__'] = current_case
+                    except Exception:
+                        errors[input.test_param.name] = 'The selected case is unavailable/invalid.'
+                group_state[input.test_param.name] = value
+
+            elif input.type == 'section':
+                populate_state(request_context, input.inputs, incoming.get(input.name), group_state, errors, context=context, check=check, simple_errors=simple_errors, input_format='21.01')
```
nsoranzo:
```suggestion
                populate_state(request_context, input.inputs, incoming.get(input.name), group_state, errors, context=context, check=check, simple_errors=simple_errors, input_format=input_format)
```

lib/galaxy/tools/parameters/__init__.py
```diff
@@ -312,7 +312,70 @@ def populate_state(request_context, inputs, incoming, state, errors={}, prefix='
...
+                    for rep_index, rep in enumerate(incoming[input.name]):
+                        new_state = {}
+                        group_state.append(new_state)
+                        populate_state(request_context, input.inputs, rep, new_state, errors, context=context, check=check, simple_errors=simple_errors, input_format='21.01')
+
+            elif input.type == 'conditional':
+                test_param_value = incoming.get(input.name).get(input.test_param.name)
+                value, error = check_param(request_context, input.test_param, test_param_value, context, simple_errors=simple_errors) if check else [test_param_value, None]
+                if error:
+                    errors[input.test_param.name] = error
```
nsoranzo, line 356: Now that I look at this again, maybe it would be good to keep passing around the `prefix` to have more precise keys in the `errors` dict?

lib/galaxy/tools/parameters/__init__.py
simonbray: Sorry, I don't get it @nsoranzo - why is this better?

mvdbeek: This is a mutable default argument. If you add to the dictionary and don't pass in a new `errors` dict in the calling function the same dict will be used over and over, so you may for instance get outdated errors. See https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

lib/galaxy/tools/parameters/__init__.py
mvdbeek: I don't think that's necessary, if I understand this correctly we should be getting back a nested dict with errors here, that is much simpler to translate back in a frontend model for instance, or to relate to your input parameters.

mvdbeek (approved):

lib/galaxy/tools/parameters/__init__.py
simonbray: TIL, thanks!

lib/galaxy/tools/parameters/__init__.py
nsoranzo: @mvdbeek I don't think the `errors` dict is nested in the current implementation.

simonbray: Maybe something like this? https://github.com/galaxyproject/galaxy/pull/10720/commits/5f00cf6ae4246e48c278f88b3e4770b374b2182d

nsoranzo (approved):

nsoranzo:
@mvdbeek Would you like to take another look at this?

lib/galaxy/webapps/galaxy/api/jobs.py
```diff
@@ -155,7 +155,8 @@ def show(self, trans, id, **kwd):
                 job_stderr=job.job_stderr,
                 stderr=job.stderr,
                 stdout=job.stdout,
-                job_messages=job.job_messages
+                job_messages=job.job_messages,
+                job_remap=job.remappable()
```
mvdbeek:
```suggestion
```
remappable used to be something that is a bit of an internal hack and I don't think this should be part of this API endpoint (it only is part of the somewhat internal `build_for_rerun`). It answers 2 different questions (are there dependent jobs ? did the job produce any dataset collection instances) that I think should be dealt with on the client or that should at least be 2 keys in the response.
Looking closer at this I think `if self.output_dataset_collection_instances:` isn't even correct, you'd probably need to ask if this job has a `ImplicitCollectionJobAssociation`.

Maybe we can take this out for now and look at a replacement in a separate PR ?

mvdbeek: The rest is great, makes sense to move the logic to the model.

simonbray: > Maybe we can take this out for now and look at a replacement in a separate PR ?

fine for me

mvdbeek (approved):

bernt-matthias:
Was just debugging a tool with profile 21.01 and noticed that the legacy branch is executed here:

https://github.com/galaxyproject/galaxy/blob/85d5e1269a1df056285aa0b831ba805b11ad2d61/lib/galaxy/tools/parameters/__init__.py#L347

seemed odd to me. Is this expected? Asking here, because this seems to be the last PR that touched this.

mvdbeek:
Yes, that's normal, we haven't touched the client.

PR #10745, opened 2020-11-18
Drop API Key listings from user list 

### LINKED ISSUE #10734: Hide the user API key on Admin by default
The page `Admin->Users` shows all of the users API keys.
image

At the very least, this should be hidden by default (eg, displayed as a bunch of `*` chars), and unless there is a good reason for it, it should just be removed altogether.

dannon:
Admins can get this via impersonate or other means if they really need it, but it's kinda like showing all the passwords and we should not do it by default.

Fixes #10734


afgane (approved):

afgane:
The single failing test seems unrelated.

Issue #10684, opened 2020-11-10
Alter tags of workflow via API


innovate-invent:
Currently it is not possible to alter tags of a workflow after creation via the API.
```
PUT /api/workflows/{id}
{ "tags": [...] }
```
Should be added to that endpoint.

mvdbeek:
You can do that, but I agree the payload is too complicated.

innovate-invent:
ah, tags wasn't included in the docs: https://docs.galaxyproject.org/en/latest/api/api.html#galaxy.webapps.galaxy.api.workflows.WorkflowsAPIController.update

davelopez:
This was addressed in #10748

Issue #10683, opened 2020-11-10
Expose workflow import config via API


innovate-invent:
Currently it is only possible to set a workflow to importable via the API during creation.
```
PUT /api/workflows/{id}
{ "importable": true }
```
should be added to that endpoint

Additionally, `GET /api/workflows/{id}` should return
```json
{ "importable": true, "import_path": "/u/nolan/w/myworkflow" }
```

simonbray:
Hi @innovate-invent, I had a look at this, but it's not really clear to me. How is the setting the importable attribute different to publishing? Is it possible to set in the UI already? Can you clarify this a bit?

innovate-invent:
If you have a look at the workflow share ui there are two options: publish and share via url. I am assuming importable refers to the latter given the logic in the workflow create function that published implies importable.

simonbray:
Okay, makes sense. So importable = available only if you have the URL and published = listed under `workflows/list_published`.

davelopez:
Fixed by #13571
Thank you so much @simonbray!

PR #10748, opened 2020-11-18
Add/document API for publishing/tagging workflows

### LINKED ISSUE #10682: Publish workflow via API
Currently it is not possible to publish a workflow after creation via the API.
```
PUT /api/workflows/{id}
{ "published": true }
```
Should be added to that endpoint.

simonbray:
Fixes https://github.com/galaxyproject/galaxy/issues/10682 and https://github.com/galaxyproject/galaxy/issues/10684.

I also spent some time looking at https://github.com/galaxyproject/galaxy/issues/10683, but I'm not really sure what 'importable' means here - or even how to change it in the UI. Not sure where in the code this needs to be fixed.


mvdbeek (approved):
Awesome, thanks a lot @simonbray!

PR #10751, opened 2020-11-18
Upgrade syntax using ``pyupgrade --py36-plus``

nsoranzo:
Also:
- Fix some flake8 bugbear errors, remove Py2 import
- Do not use unsecure `yaml.load()`
- Put some imports on top

### DIFF
--- a/lib/galaxy/web/framework/base.py
+++ b/lib/galaxy/web/framework/base.py
@@ -20,11 +20,6 @@ from paste.response import HeaderDict
 
 from galaxy.util import smart_str
 
-try:
-    file_types = (file, io.IOBase)
-except NameError:
-    file_types = (io.IOBase, )
-
 log = logging.getLogger(__name__)
 
 #: time of the most recent server startup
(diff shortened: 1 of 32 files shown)


lib/galaxy/web/framework/base.py
```diff
@@ -344,7 +339,7 @@ def _read_lines(self):
     # Adapt `self.__file = None` to Python name mangling of class-private attributes.
     # We need to patch the original FieldStorage class attribute, not the cgi_FieldStorage
     # class.
-    setattr(self, '_FieldStorage__file', None)
+    self._FieldStorage__file = None
```
mvdbeek, line 342: I don't know if that is completely equivalent, seems a bit risky. This needs careful checking in the regular uploader and the history archive uploader. I remember this was a pain when I set this up.

nsoranzo: According to the documentation they are equivalent: https://docs.python.org/3/library/functions.html#setattr
Using `setattr` became unnecessary at commit https://github.com/galaxyproject/galaxy/commit/1dc4cf224108db9756a96b9d50a71f9f099a00f2 , when the attribute name was changed to a fixed string.

mvdbeek: Awesome, thanks for clarifying that!

mvdbeek (approved):

PR #10752, opened 2020-11-18
Add link to view dataset in dataset params

### LINKED ISSUE #10706: Add link to view dataset in dataset params
The view `/datasets/{id}/show_params` should include a `View dataset` link somewhere so that you can click through to inputs and inspect the data without having to track down the dataset in the history.

simonbray:
Fixes https://github.com/galaxyproject/galaxy/issues/10706. I don't really know anything about javascript or vue but this seems to work.

### DIFF
--- a/client/src/components/DatasetInformation/DatasetInformation.vue
+++ b/client/src/components/DatasetInformation/DatasetInformation.vue
@@ -31,7 +31,7 @@
                 </tr>
                 <tr>
                     <td>File contents</td>
-                    <td id="file-contents"><a :href=dataset.download_url>contents</a></td>
+                    <td id="file-contents"><a :href="dataset.download_url">contents</a></td>
                 </tr>
             </tbody>
         </table>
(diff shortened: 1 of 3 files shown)


client/src/components/DatasetInformation/DatasetInformation.vue
```diff
@@ -29,6 +29,10 @@
                     <td>Format</td>
                     <td id="format">{{ dataset.file_ext }}</td>
                 </tr>
+                <tr>
+                    <td>File contents</td>
+                    <td id="file-contents"><a :href=dataset.download_url>contents</a></td>
```
dannon:
```suggestion
                    <td id="file-contents"><a :href="dataset.download_url">contents</a></td>
```

dannon:
@simonbray Thanks for this; looks like the implementation is fine but the test needs a little fix.  If you want, you can run these locally like this for iterative development:

```
cd client
yarn run jest-watch datasetinformation
```

If you'd rather, I can take a poke at it too.

simonbray:
@dannon thanks, I'll have a look at it in a second.

dannon (approved):

PR #10754, opened 2020-11-18
Revision of sharing component, adds url copy to clipboard feature

### LINKED ISSUE #10732: Add a button to copy shared URLs to the clipboard
Copying a shared URL for histories shared via a link or published is cumbersome requiring the user to select the URL by hand. 
image

Would be nice to add a `Copy to clipboard` button for single-click copying of the URL, like so for example:
image

I find it the fact the copy button is on the left hand side of the value is nice for consistency as the values tend to vary in length.

While on this page, it would be nice to also update the pencil edit url icon because it is not from the same UI family as most of the rest of Galaxy:
image

guerler:
This PR revises the sharing component by removing legacy jquery helpers and introducing a slug input component. Additionally it adds a clipboard feature, allowing users to copy the url. It also applies conventional naming and formatting in consensus with newer components and use the font awesome components to render icons. Fixes #10732.

image

### DIFF
--- a/client/src/components/Sharing.vue
+++ b/client/src/components/Sharing.vue
@@ -31,7 +31,7 @@
                             <font-awesome-icon icon="edit" />
                         </b-button>
                         <b-button id="tooltip-clipboard" @click="onCopy" @mouseout="onCopyOut" variant="link" size="sm">
-                            <font-awesome-icon :icon="['far', 'clipboard']" />
+                            <font-awesome-icon icon="link" />
                         </b-button>
                         <b-tooltip target="tooltip-clipboard" triggers="hover">
                             {{ tooltipClipboard }}
@@ -46,7 +46,7 @@
                 </div>
                 <div v-else>
                     Access to this {{ model_class }} is currently restricted so that only you and the users listed below
-                    can access it. Note that sharing a History will also allow access to all of its datasets. 
+                    can access it. Note that sharing a History will also allow access to all of its datasets.
                 </div>
             </div>
             <br />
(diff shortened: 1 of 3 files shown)


nsoranzo:
@guerler We recently had a discussion and vote on the "copy URL to clipboard" icon, and chose the `fa-chain` one, see https://github.com/galaxyproject/galaxy/pull/10529 . That PR could also use a review :)

client/src/components/Sharing.vue
```diff
@@ -82,15 +82,15 @@ import Vue from "vue";
 import BootstrapVue from "bootstrap-vue";
 import { FontAwesomeIcon } from "@fortawesome/vue-fontawesome";
 import { library } from "@fortawesome/fontawesome-svg-core";
-import { faClipboard, faEdit } from "@fortawesome/free-regular-svg-icons";
+import { faLink, faEdit } from "@fortawesome/free-solid-svg-icons";
```
dannon, line 85: is faEdit available in both regular and solid?

guerler: Yep. https://fontawesome.com/icons/edit?style=solid

dannon:
@guerler A nice next step would be to fix up 'share with indivdiual user' so that it doesn't load a new page, but this is a great start and it's working well.  Looks nice.

hexylena:
This seems to remove the need to set 'Also make all objects within the history accessible.', just checking that that's intended, right? If so, that's fantastic, makes sharing SO much less bugprone on EU and other servers that enforce dataset privacy by default within history.

PR #10760, opened 2020-11-18
[20.09] Include dataset and state in valid state exception

mvdbeek:


lib/galaxy/tools/__init__.py
```diff
@@ -2773,7 +2773,7 @@ def check_dataset_instance(input_dataset):
 
             if self.require_dataset_ok:
                 if input_dataset.state != input_dataset.dataset.states.OK:
-                    raise ValueError("Tool requires inputs to be in valid state.")
+                    raise ValueError("Tool requires inputs to be in valid state, but {} is in state '{}".format(input_dataset, input_dataset.state))
```
nsoranzo: Should this be
```suggestion
                    raise ValueError("Tool requires inputs to be in valid state, but dataset {} is in state '{}'".format(input_dataset, input_dataset.state))
```
?

mvdbeek: I was hoping to get the `__repr__` of the ~HDA~ dataset

mvdbeek: (because this may not be an HDA)

nsoranzo: OK, I've updated the suggestion.

nsoranzo (approved):

dannon (approved):

Issue #10765, opened 2020-11-18
Tooltips in RuleBuilder cover buttons


assuntad23:
![Screenshot from 2020-11-18 14-21-38](https://user-images.githubusercontent.com/26912553/99580300-4fcf5180-29ad-11eb-9e12-2cd1cb6440d8.png)

This makes it hard to select the button that you want.

dannon:
Fixed in #10766

PR #10766, opened 2020-11-18
Move tooltips in rule builder below buttons, and put buttons on the same row

assuntad23:
Fix for #10765

- Put all buttons in a horizontal button group
- Tool tips drop to the bottom, options drop up 
- Increased the size of the column by 20px to account for all buttons being beside each other.

Screenshot from 2020-11-18 16-26-31


dannon (approved):
Works well, thanks!

PR #10767, opened 2020-11-18
Allow retrying test output downloads for NFS issues.

jmchilton:


mvdbeek (commented):
That seems like it should be an error. If the dataset is not available for a terminal job we're in trouble, so if tests pass with a retry I think that hides a rather serious problem.

jmchilton:
I need to know if the tools work or they don't - the tools do work - extending this does fix the failing tests. I think workflow testing will indicate if there are issues with this when chained with other jobs but I don't think it is the job of tool testing to solve or even really indicate.

Philosophic differences aside - this doesn't change the default - which is to not wait on this. So this tool would still indicate an error in these conditions unless the runner of the tool explicitly wants to ignore those.

PR #10774, opened 2020-11-19
[20.09] Fixes for singularity execution and resolution

mvdbeek:
Includes #10625

### DIFF
--- a/lib/galaxy/tool_util/deps/container_resolvers/mulled.py
+++ b/lib/galaxy/tool_util/deps/container_resolvers/mulled.py
@@ -447,7 +447,7 @@ class MulledSingularityContainerResolver(MulledDockerContainerResolver):
         self.hash_func = hash_func
         self.auto_install = string_as_bool(auto_install)
 
-    def cached_container_description(self, targets, namespace, hash_func):
+    def cached_container_description(self, targets, namespace, hash_func, resolution_cache):
         return singularity_cached_container_description(targets,
                                                         cache_directory=self.cache_directory,
                                                         hash_func=hash_func)
--- a/lib/galaxy/tool_util/deps/container_classes.py
+++ b/lib/galaxy/tool_util/deps/container_classes.py
@@ -125,8 +125,10 @@ def preprocess_volumes(volumes_raw_str, container_type):
     ['/a/b:/a:ro', '/a/b/c:/a/b:rw']
     >>> preprocess_volumes("/a/b:default_ro,/a/b/c:rw", DOCKER_CONTAINER_TYPE)
     ['/a/b:ro', '/a/b/c:rw']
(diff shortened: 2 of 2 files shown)


lib/galaxy/tool_util/deps/container_resolvers/mulled.py
```diff
@@ -213,7 +213,7 @@ def docker_cached_container_description(targets, namespace, hash_func="v2", shel
     return container
 
 
-def singularity_cached_container_description(targets, cache_directory, hash_func="v2", shell=DEFAULT_CONTAINER_SHELL):
+def singularity_cached_container_description(targets, cache_directory, hash_func="v2", shell=DEFAULT_CONTAINER_SHELL, resolution_cache=None):
```
nsoranzo: Not sure what was the issue exactly, but I think the correct fix would be:

```diff
diff --git a/lib/galaxy/tool_util/deps/container_resolvers/mulled.py b/lib/galaxy/tool_util/deps/container_resolvers/mulled.py
index 5f919ec964..e5ac3af311 100644
--- a/lib/galaxy/tool_util/deps/container_resolvers/mulled.py
+++ b/lib/galaxy/tool_util/deps/container_resolvers/mulled.py
@@ -447,7 +447,7 @@ class MulledSingularityContainerResolver(MulledDockerContainerResolver):
         self.hash_func = hash_func
         self.auto_install = string_as_bool(auto_install)
 
-    def cached_container_description(self, targets, namespace, hash_func):
+    def cached_container_description(self, targets, namespace, hash_func, resolution_cache):
         return singularity_cached_container_description(targets,
                                                         cache_directory=self.cache_directory,
                                                         hash_func=hash_func)
```

(and this can be reverted).

mvdbeek: Yup, that's what I had on the server I was upgrading 🦆 
```
diff --git a/lib/galaxy/tool_util/deps/container_resolvers/mulled.py b/lib/galaxy/tool_util/deps/container_resolvers/mulled.py
index e456c89..9c24451 100644
--- a/lib/galaxy/tool_util/deps/container_resolvers/mulled.py
+++ b/lib/galaxy/tool_util/deps/container_resolvers/mulled.py
@@ -447,7 +447,7 @@ class MulledSingularityContainerResolver(MulledDockerContainerResolver):
         self.hash_func = hash_func
         self.auto_install = string_as_bool(auto_install)

-    def cached_container_description(self, targets, namespace, hash_func):
+    def cached_container_description(self, targets, namespace, hash_func, **kwds):
         return singularity_cached_container_description(targets,
                                                         cache_directory=self.cache_directory,
                                                         hash_func=hash_func)
```

nsoranzo (approved):

PR #10780, opened 2020-11-20
[20.09] fix history_dataset_as_image in workflow invocations

### LINKED ISSUE #10764: Workflow reports history_dataset_as_image with path option broken
In https://github.com/galaxyproject/galaxy/pull/9769 @OlegZharkov implemented additional syntax for workflow reports including the ability to display an extra file (e.g. from a composite datatype) as an image in the report.
This particular functionality relying on an optional path argument to history_dataset_as_image seems to have been broken at some point since then (it seems that the path argument is getting ignored now), maybe as a result of https://github.com/galaxyproject/galaxy/pull/10121 (?), and should be restored.

OlegZharkov:
Just little refactoring issue.
Also in general we are not supposed to use async functions in computed, right?

Anyway, it fixes https://github.com/galaxyproject/galaxy/issues/10764

### DIFF
--- a/client/src/components/Dataset/DatasetAsImage/DatasetAsImage.vue
+++ b/client/src/components/Dataset/DatasetAsImage/DatasetAsImage.vue
@@ -1,6 +1,10 @@
 <template>
     <div>
-        <img v-if="imgUrl" :src="imgUrl" />
+        <div v-if="imageUrl" class="w-50 p-2 float-left">
+            <b-card nobody body-class="p-1">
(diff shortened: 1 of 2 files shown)


bgruening:
Should this target the stable branch?

OlegZharkov:
> Should this target the stable branch?

I could backport it, after it's merged

dannon:
@OlegZharkov Might as well just target 20.09 and merge forward, though?

PR #10781, opened 2020-11-20
Fix reference error with variable declaration

tmcgowan:
An undeclared var was generating a reference error preventing a table from being populated.

From the console log:
"Uncaught ReferenceError: assignment to undeclared variable psmDetailDP"


bgruening:
@tmcgowan you probably want to target release_20.09 with this PR.

tmcgowan:
Redoing the PR with proper branch

PR #10783, opened 2020-11-21
Refactor more generally useful Selenium code into reusable library code.

jmchilton:
And out of the testing specific package.


bgruening:
@OlegZharkov can you have a look, please? Thanks!

OlegZharkov:
Hmm, some general `how to` would also be very very useful.

Issue #10790, opened 2020-11-23
Containers admin/toolbox_dependencies produces uncaught exeception on Galaxy web process


pcm32:
Trying to look at the admin/toolbox_dependencies for containers produces an uncaught exception:

```
galaxy.web.framework.decorators ERROR 2020-11-23 22:03:41,416 [p:1208242,w:2,m:0] [uWSGIWorker2Core0] Uncaught exception in exposed API method:
Traceback (most recent call last):
  File "lib/galaxy/web/framework/decorators.py", line 282, in decorator
    rval = func(self, trans, *args, **kwargs)
  File "lib/galaxy/web/framework/decorators.py", line 101, in decorator
    return func(self, trans, *args, **kwargs)
  File "lib/galaxy/webapps/galaxy/api/container_resolution.py", line 84, in resolve_toolbox
    return self._view.resolve_toolbox(**kwds)
  File "lib/galaxy/tool_util/deps/views.py", line 391, in resolve_toolbox
    rval.append(self.resolve(tool_id=tool_id, **resolve_kwds))
  File "lib/galaxy/tool_util/deps/views.py", line 367, in resolve
    **find_best_kwds
  File "lib/galaxy/tool_util/deps/containers.py", line 58, in resolve
    return self.container_registry.resolve(enabled_container_types, tool_info, **kwds)
  File "lib/galaxy/tool_util/deps/containers.py", line 261, in resolve
    container_description = container_resolver.resolve(enabled_container_types, tool_info, resolution_cache=resolution_cache)
  File "lib/galaxy/tool_util/deps/container_resolvers/mulled.py", line 440, in resolve
    resolution_cache=resolution_cache,
TypeError: cached_container_description() got an unexpected keyword argument 'resolution_cache'
```

I see this in 20.01, but judging by the code on dev, I think that the fix I used on my setup will do here as well.

PR #10791, opened 2020-11-23
Resolve uncaught exception in admin/tool_box_dependencies for containers

pcm32:
This PR deals with #10790, seen originally in 20.01, but judging by the current dev code, this fix is still applicable. This enables the UI to present the admin/tool_box_dependencies for containers.


mvdbeek:
That was already fixed in https://github.com/galaxyproject/galaxy/pull/10774, I'll merge that forward.

PR #10796, opened 2020-11-24
[20.09] Fix list of cached Docker images not getting updated

mvdbeek:
by using the ephemeral cache as it was intended.

In #10451 we started passing in the beaker-backed persistent cache, when the code was expecting to be able to set per-request ephemeral values in the `ResolutionCache` object. This fixes new docker images not being returned in the `CachedDocker` resolver.

That is the simplest fix, but we should probably refactor this in `dev` so that even successful resolution has an expiry date, in case there is an updated image on quay that should be used.


mvdbeek:
Meh, I think we can do better, this should probably go to a short-term cache ...

PR #10801, opened 2020-11-24
Don't cancel other selenium steps, if one step fails

OlegZharkov:
Selenium improvement in Github actions


OlegZharkov:
thanks @nsoranzo!

PR #10805, opened 2020-11-25
show warning on trs-workflow import for anonymous users

OlegZharkov:
Since we want to deploy auto-import [buttons](https://gist.github.com/bgruening/a891880ba98534f11ceb42be73d90646). We need some sort of warning for unregistered users

### DIFF
--- a/client/src/components/Workflow/TrsImport.vue
+++ b/client/src/components/Workflow/TrsImport.vue
@@ -24,7 +24,7 @@
             </div>
         </b-card>
         <b-alert v-else show variant="danger" class="text-center my-2"
-            >Only registered users can import workflows!</b-alert
+            >Anonymous user cannot import workflows, please register or log in</b-alert
         >
     </div>
 </template>
--- a/client/src/components/Workflow/TrsImport.vue
+++ b/client/src/components/Workflow/TrsImport.vue
@@ -77,7 +77,7 @@ export default {
             return this.errorMessage != null;
         },
         isAnonymous() {
-            return !getGalaxyInstance().user.isAnonymous();
+            return getGalaxyInstance().user.isAnonymous();
         },
     },
     watch: {
--- a/client/src/components/Workflow/TrsImport.vue
+++ b/client/src/components/Workflow/TrsImport.vue
@@ -1,6 +1,6 @@
 <template>
     <div>
-        <b-card title="GA4GH Tool Registry Server (TRS) Workflow Import">
+        <b-card v-if="isRegistered" title="GA4GH Tool Registry Server (TRS) Workflow Import">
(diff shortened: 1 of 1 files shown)


client/src/components/Workflow/TrsImport.vue
```diff
@@ -23,6 +23,9 @@
                 <trs-tool :trs-tool="trsTool" v-if="trsTool" @onImport="importVersion(trsTool.id, $event)" />
             </div>
         </b-card>
+        <b-alert v-else show variant="danger" class="text-center my-2"
+            >Only registered users can import workflows!</b-alert
```
mvdbeek: Can you rephrase this to `Anonymous user cannot import workflows, please log in.` or something similar ? That seems like it would be easier to understand for users.

client/src/components/Workflow/TrsImport.vue
```diff
@@ -72,6 +76,9 @@ export default {
         hasErrorMessage() {
             return this.errorMessage != null;
         },
+        isRegistered() {
```
mvdbeek: I think it might be better to have this be `isAnonymous`, since not all users need to be registered to be able to import workflows (e.g. with auto-registration or other fancy/weird setups). I think in practice that won't make a difference, but it more closely reflects what you're checking here.

client/src/components/Workflow/TrsImport.vue
```diff
@@ -72,6 +76,9 @@ export default {
         hasErrorMessage() {
             return this.errorMessage != null;
         },
+        isAnonymous() {
+            return !getGalaxyInstance().user.isAnonymous();
```
mvdbeek:
```suggestion
            return getGalaxyInstance().user.isAnonymous();
```

client/src/components/Workflow/TrsImport.vue
```diff
@@ -1,6 +1,6 @@
 <template>
     <div>
-        <b-card title="GA4GH Tool Registry Server (TRS) Workflow Import">
+        <b-card v-if="isAnonymous" title="GA4GH Tool Registry Server (TRS) Workflow Import">
```
mvdbeek:
```suggestion
        <b-card v-if="!isAnonymous" title="GA4GH Tool Registry Server (TRS) Workflow Import">
```

OlegZharkov: rights, sorry. Lost freshness of vision

jmchilton:
Awesome - thanks! Feel free to backport this 20.09 if that helps with the deployment.

OlegZharkov:
> Awesome - thanks! Feel free to backport this 20.09 if that helps with the deployment.

thanks John, will do

PR #10807, opened 2020-11-25
Fix Library Folder performance

### LINKED ISSUE #10538: data libraries interface speed regression
After the rewrite in 20.09 the DLs interface feels much slower. One of the reasons could be that there is a lot of additional javascript calls in unnecessary places illustrated below.

Galaxy
Galaxy

Galaxy___UMSA___Data_Libraries_and_New_Issue_·_galaxyproject_galaxy

OlegZharkov:
Do not reload entire component, when user moves into another folder. 

fixes https://github.com/galaxyproject/galaxy/issues/10538

### DIFF
--- a/client/src/components/LibraryFolder/LibraryFolder.vue
+++ b/client/src/components/LibraryFolder/LibraryFolder.vue
@@ -340,7 +340,7 @@ export default {
         moveToParentFolder() {
             const path = this.folder_metadata.full_path;
             if (path.length === 1) {
-                return window.location = `${this.root}library/list/`;
+                window.location = `${this.root}library/list/`;
             } else {
                 return this.changeFolderId(path[path.length - 2][0]);
             }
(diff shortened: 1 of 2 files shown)


OlegZharkov:
hmmm, GHA doesn't work for drafts anymore?

mvdbeek:
This looks like a rather big change to 20.09, I would prefer if you would target dev with this.

OlegZharkov:
> This looks like a rather big change to 20.09, I would prefer if you would target dev with this.

Wait, I can remove pagination

mvdbeek (approved):
Very nice!

OlegZharkov:
> Very nice!

@mvdbeek thanks :)

martenson:
Thanks for taking care of this @OlegZharkov !

OlegZharkov:
> Thanks for taking care of this @OlegZharkov !

Thanks for reporting! I should've done this way from the beginning.

PR #10808, opened 2020-11-26
[20.09] Disable tool cache if there is a problem

mvdbeek:
There are a couple of circumstances where the tool document cache is not working correctly, such as if the cache is on NFS or CIFS.

This will disable the cache in case there are any sqlite3.OperationalErrors.
I've verified this works correctly by putting the tool cache on NFS and installing a couple of tools, making sure the tools are loaded both in the installing processes and the various workers.

### DIFF
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -276,10 +276,11 @@ class ToolBox(BaseGalaxyToolBox):
         that re-opens the database after forking.
         """
         for region in self.cache_regions.values():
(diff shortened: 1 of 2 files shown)


nsoranzo (approved):

innovate-invent:
Moving the cache to a hosted database would better handle these situations: https://github.com/galaxyproject/galaxy/pull/10436

mvdbeek:
That is not the same cache, and beaker pickles, which is slower than no cache.

PR #10810, opened 2020-11-26
Implement read-only conda resolver.

jmchilton:


mvdbeek (approved):

mvdbeek:
Can you also add a note to https://github.com/galaxyproject/galaxy/blob/dev/doc/source/admin/dependency_resolvers.rst#conda-dependency-resolver about this ?

lib/galaxy/tool_util/deps/resolvers/__init__.py
```diff
@@ -240,9 +241,16 @@ class InstallableDependencyResolver(metaclass=ABCMeta):
     """ Mix this into a ``DependencyResolver`` and implement to indicate
     the dependency resolver can attempt to install new dependencies.
     """
+    read_only = False
```
nsoranzo, line 245: Given that the presence of the `install_dependency` method or inheriting from `InstallableDependencyResolver` is not going to be a reliable indication of the resolver's ability to install dependencies any more, I suggest to merge this class into `DependencyResolver`.

Also, these following 2 lines need to be updated any way to check for `read_only` instead of using `hasattr`:

https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tool_util/deps/views.py#L147 (`resolver.disabled` should probably also checked here, raising a different exception)

https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tool_util/deps/views.py#L201

mvdbeek:
Pushing this because of the outstanding comment

jmchilton:
I continue to prefer this version, but I have addressed the comment with a separate implementation in https://github.com/galaxyproject/galaxy/pull/10822.

I've rebased both versions and attempted to fix the test error in the other version.

PR #10812, opened 2020-11-26
fix vue-unit test in toolsview.test.js 

OlegZharkov:
Another quite annoying test, that fails only sometimes. If it rendered tools with short description like "new-beginning1",  this line failed 
`expect(infoButton.attributes("aria-expanded") === "true").toBeTruthy();` because there is nothing to expand.

This PR reduces mocked API response to 5 tools only and fixes this unit test


OlegZharkov:
Thanks for merging @mvdbeek

PR #10814, opened 2020-11-27
[20.09] Restore the ability to pull containers from admin menu

mvdbeek:
This used to work in the old dependency menu because it went through
https://github.com/mvdbeek/galaxy/blob/dev/lib/galaxy/tool_util/deps/__init__.py#L263

### DIFF
--- a/lib/galaxy/tool_util/deps/container_resolvers/mulled.py
+++ b/lib/galaxy/tool_util/deps/container_resolvers/mulled.py
@@ -414,13 +414,13 @@ class MulledDockerContainerResolver(ContainerResolver):
                     hash_func=self.hash_func,
                     resolution_cache=resolution_cache,
             ):
-                destination = {}
+                destination_info = {}
                 if destination_for_container_type:
-                    destination = destination_for_container_type(self.container_type)
+                    destination_info = destination_for_container_type(self.container_type)
                 container = CONTAINER_CLASSES[self.container_type](container_description.identifier,
                                                                    self.app_info,
                                                                    tool_info,
-                                                                   destination,
+                                                                   destination_info,
                                                                    {},
                                                                    container_description)
                 self.pull(container)
--- a/lib/galaxy/tool_util/deps/container_resolvers/mulled.py
+++ b/lib/galaxy/tool_util/deps/container_resolvers/mulled.py
@@ -407,7 +407,6 @@ class MulledDockerContainerResolver(ContainerResolver):
                 type=self.container_type,
                 shell=self.shell,
             )
-            destination_for_container_type = kwds.get('destination_for_container_type')
             if install and not self.cached_container_description(
                     targets,
                     namespace=self.namespace,
@@ -415,6 +414,7 @@ class MulledDockerContainerResolver(ContainerResolver):
                     resolution_cache=resolution_cache,
             ):
                 destination_info = {}
+                destination_for_container_type = kwds.get('destination_for_container_type')
                 if destination_for_container_type:
                     destination_info = destination_for_container_type(self.container_type)
                 container = CONTAINER_CLASSES[self.container_type](container_description.identifier,
--- a/lib/galaxy/tool_util/deps/container_resolvers/mulled.py
+++ b/lib/galaxy/tool_util/deps/container_resolvers/mulled.py
@@ -408,16 +408,19 @@ class MulledDockerContainerResolver(ContainerResolver):
                 shell=self.shell,
             )
             destination_for_container_type = kwds.get('destination_for_container_type')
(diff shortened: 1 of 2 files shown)


lib/galaxy/tool_util/deps/container_resolvers/mulled.py
```diff
@@ -408,16 +408,19 @@ def resolve(self, enabled_container_types, tool_info, install=False, session=Non
...
             )
             destination_for_container_type = kwds.get('destination_for_container_type')
-            if install and destination_for_container_type and not self.cached_container_description(
+            if install and not self.cached_container_description(
                     targets,
                     namespace=self.namespace,
                     hash_func=self.hash_func,
                     resolution_cache=resolution_cache,
             ):
+                destination = {}
```
nsoranzo:
```suggestion
                destination_info = {}
```

lib/galaxy/tool_util/deps/container_resolvers/mulled.py
```diff
@@ -408,16 +408,19 @@ def resolve(self, enabled_container_types, tool_info, install=False, session=Non
...
+                if destination_for_container_type:
+                    destination = destination_for_container_type(self.container_type)
```
nsoranzo:
```suggestion
                    destination_info = destination_for_container_type(self.container_type)
```

lib/galaxy/tool_util/deps/container_resolvers/mulled.py
```diff
@@ -408,16 +408,19 @@ def resolve(self, enabled_container_types, tool_info, install=False, session=Non
...
                 container = CONTAINER_CLASSES[self.container_type](container_description.identifier,
                                                                    self.app_info,
                                                                    tool_info,
-                                                                   destination_for_container_type(self.container_type),
+                                                                   destination,
```
nsoranzo:
```suggestion
                                                                   destination_info,
```

nsoranzo (commented):
Also `destination_for_container_type` should probably be renamed to `destination_info_for_container_type` here and at https://github.com/galaxyproject/galaxy/blob/release_20.09/lib/galaxy/tool_util/deps/__init__.py#L263

mvdbeek:
Let's do that in dev ?

lib/galaxy/tool_util/deps/container_resolvers/mulled.py
```diff
@@ -408,16 +408,19 @@ def resolve(self, enabled_container_types, tool_info, install=False, session=Non
                 shell=self.shell,
             )
             destination_for_container_type = kwds.get('destination_for_container_type')
```
nsoranzo:
```suggestion
```

lib/galaxy/tool_util/deps/container_resolvers/mulled.py
```diff
@@ -408,16 +408,19 @@ def resolve(self, enabled_container_types, tool_info, install=False, session=Non
...
-            if install and destination_for_container_type and not self.cached_container_description(
+            if install and not self.cached_container_description(
                     targets,
                     namespace=self.namespace,
                     hash_func=self.hash_func,
                     resolution_cache=resolution_cache,
             ):
+                destination_info = {}
+                if destination_for_container_type:
```
nsoranzo, line 418:
```suggestion
                destination_for_container_type = kwds.get('destination_for_container_type')
                if destination_for_container_type:
```

nsoranzo:
> Let's do that in dev ?

Sure

nsoranzo (approved):

lib/galaxy/tool_util/deps/container_resolvers/mulled.py
```diff
@@ -407,17 +407,20 @@ def resolve(self, enabled_container_types, tool_info, install=False, session=Non
                 type=self.container_type,
                 shell=self.shell,
             )
-            destination_for_container_type = kwds.get('destination_for_container_type')
-            if install and destination_for_container_type and not self.cached_container_description(
```
dannon, line 411: It's this bit right?  It dropped the bypass when a destination is trusted to understand a container type implicitly?

mvdbeek: Are you looking at a bug or weird behavior ? If you wanted to fetch a container previously you had to pass in `destination_for_container_type `, which you don't have available when not preparing a job, for instance when pulling an image from the admin panel. Is this causing excessive pulls, and if so, what does the container resolution config look like ?

mvdbeek: Ah, just saw the issue now.

Issue #10815, opened 2020-11-27
[20.09] trs servers must be set


selten:
I just tried to set no trs servers as I do not wish to use this functionality and I got the following error:
```
Traceback (most recent call last):
  File "lib/galaxy/webapps/galaxy/buildapp.py", line 48, in app_factory
    app = galaxy.app.UniverseApplication(global_conf=global_conf, **kwargs)
  File "lib/galaxy/app.py", line 229, in __init__
    self.trs_proxy = TrsProxy(self.config)
  File "lib/galaxy/workflow/trs_proxy.py", line 58, in __init__
    self._server_dict = {t["id"]: t for t in self._server_list}
TypeError: 'NoneType' object is not iterable
```

based on the following config:
```
# List GA4GH Tool Registry Services to connect Galaxy with.
# 'id' is used to key API interactions and should be unique, the backend uses 'api_url'
# to build requests. The remaining fields 'label', 'link_url', and 'doc' are used to render
# information in the user interface.

# By default only dockstore is configured.
#- id: dockstore
#  api_url: https://dockstore.org/api
#  link_url: https://dockstore.org
#  label: Dockstore
#  doc: |
#    Dockstore is an open platform used by the GA4GH for sharing Docker-based tools and workflows.

# Staging variant of dockstore can be configured.
#- id: staging_dockstore
#  api_url: https://staging.dockstore.org/api
#  link_url: https://staging.dockstore.org
#  label: Staging Dockstore
#  doc: |
#    Staging variant of Dockstore for testing.
```

mvdbeek:
The docs say that `The file should be a list`, so if you would put `[]` as the sole content of the file that would also work around the problem. The PR looks good too.

PR #10817, opened 2020-11-27
Allow TRS servers to not be defined 

selten:
TRS servers are currently required to be set, but if you configure none to be set, Galaxy will error out because it can't find one (#10815).

### DIFF
--- a/lib/galaxy/workflow/trs_proxy.py
+++ b/lib/galaxy/workflow/trs_proxy.py
@@ -55,7 +55,7 @@ class TrsProxy:
         else:
             server_list = DEFAULT_TRS_SERVERS
         self._server_list = server_list if server_list else []
-        self._server_dict = {t["id"]: t for t in self._server_list} if self._server_list else {}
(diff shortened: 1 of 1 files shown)


lib/galaxy/workflow/trs_proxy.py
```diff
@@ -54,8 +54,8 @@ def __init__(self, config=None):
                 server_list = yaml.safe_load(f)
         else:
             server_list = DEFAULT_TRS_SERVERS
-        self._server_list = server_list
-        self._server_dict = {t["id"]: t for t in self._server_list}
+        self._server_list = server_list if server_list else []
+        self._server_dict = {t["id"]: t for t in self._server_list} if self._server_list else {}
```
mvdbeek:
```suggestion
        self._server_dict = {t["id"]: t for t in self._server_list}
```

mvdbeek (approved):

jmchilton:
Thanks for the fix!

Issue #10816, opened 2020-11-27
[20.09] Caching problems


selten:
I am already running 20.09 in my dev system which is running fine. This instance is at commit: https://github.com/galaxyproject/galaxy/commit/fdee49864b1499b55eff6ee664ea8ebbda49df51.   
The update from today put the test system at https://github.com/galaxyproject/galaxy/commit/06b8ab2098fe8987c9b91771de62b2df9e60dd57.

After going to back to the commit from the dev system, the test system works fine.

I believe that https://github.com/galaxyproject/galaxy/pull/10808 causes the following issue:
```
sqlitedict.SqliteMultithread ERROR 2020-11-27 14:03:45,759 [p:36087,w:0,m:5] [Thread-7] Inner exception:
sqlitedict.SqliteMultithread ERROR 2020-11-27 14:03:45,759 [p:36086,w:0,m:4] [Thread-7] Inner exception:
sqlitedict.SqliteMultithread ERROR 2020-11-27 14:03:45,760 [p:36086,w:0,m:4] [Thread-7]   File "[PYTHON]/lib/python3.7/threading.py", line 885, in _bootstrap
    self._bootstrap_inner()
sqlitedict.SqliteMultithread ERROR 2020-11-27 14:03:45,760 [p:36087,w:0,m:5] [Thread-7]   File "[PYTHON]/lib/python3.7/threading.py", line 885, in _bootstrap
    self._bootstrap_inner()


sqlitedict.SqliteMultithread ERROR 2020-11-27 14:03:45,760 [p:36086,w:0,m:4] [Thread-7]   File "[PYTHON]/lib/python3.7/threading.py", line 917, in _bootstrap_inner
    self.run()
sqlitedict.SqliteMultithread ERROR 2020-11-27 14:03:45,760 [p:36087,w:0,m:5] [Thread-7]   File "[PYTHON]/lib/python3.7/threading.py", line 917, in _bootstrap_inner
    self.run()


sqlitedict.SqliteMultithread ERROR 2020-11-27 14:03:45,760 [p:36086,w:0,m:4] [Thread-7]   File "[GALAXY]/.venv/lib/python3.7/site-packages/sqlitedict.py", line 417, in run
    inner_stack = traceback.extract_stack()
[...]

PR #10818, opened 2020-11-27
[20.09] Various bugfixes for reloading the toolbox

mvdbeek:
Fixes https://github.com/galaxyproject/galaxy/issues/10816 and some more things that were problematic.

### DIFF
--- a/lib/galaxy/tools/cache.py
+++ b/lib/galaxy/tools/cache.py
@@ -162,11 +162,11 @@ class ToolCache:
         removed_tool_ids = []
         try:
             with self._lock:
-                persist_callback = None
+                persist_tool_document_cache = False
                 paths_to_cleanup = {(path, tool) for path, tool in self._tools_by_path.items() if self._should_cleanup(path)}
                 for config_filename, tool in paths_to_cleanup:
                     tool.remove_from_cache()
-                    persist_callback = tool.app.toolbox.persist_cache
+                    persist_tool_document_cache = True
                     del self._hash_by_tool_paths[config_filename]
                     if os.path.exists(config_filename):
                         # This tool has probably been broken while editing on disk
@@ -182,8 +182,8 @@ class ToolCache:
                     self._removed_tool_ids.add(tool_id)
                     if tool_id in self._new_tool_ids:
                         self._new_tool_ids.remove(tool_id)
-                if persist_callback:
-                    persist_callback()
+                if persist_tool_document_cache:
+                    tool.app.toolbox.persist_cache()
         except Exception as e:
             log.debug("Exception while checking tools to remove from cache: %s", unicodify(e))
             # If by chance the file is being removed while calculating the hash or modtime
--- a/lib/galaxy/tools/cache.py
+++ b/lib/galaxy/tools/cache.py
@@ -161,11 +161,11 @@ class ToolCache:
         removed_tool_ids = []
         try:
             with self._lock:
-                persist_tool_document_cache = False
+                persist_callback = None
                 paths_to_cleanup = {(path, tool) for path, tool in self._tools_by_path.items() if self._should_cleanup(path)}
                 for config_filename, tool in paths_to_cleanup:
                     tool.remove_from_cache()
-                    persist_tool_document_cache = True
+                    persist_callback = tool.app.toolbox.persist_cache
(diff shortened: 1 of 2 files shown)


mvdbeek:
Test errors are unrelated.

lib/galaxy/tools/cache.py
```diff
@@ -161,11 +161,11 @@ def cleanup(self):
         removed_tool_ids = []
         try:
             with self._lock:
-                persist_tool_document_cache = False
+                persist_callback = None
                 paths_to_cleanup = {(path, tool) for path, tool in self._tools_by_path.items() if self._should_cleanup(path)}
                 for config_filename, tool in paths_to_cleanup:
                     tool.remove_from_cache()
-                    persist_tool_document_cache = True
+                    persist_callback = tool.app.toolbox.persist_cache
```
nsoranzo: I preferred the previous version using a boolean to track if we need to persist, which is also slightly more efficient being in a loop?

lib/galaxy/tools/cache.py
```diff
@@ -53,7 +53,7 @@ def _get_cache(self, flag='r', create_if_necessary=False):
                 if flag == 'r':
                     self._cache.flag = flag
             else:
-                self._cache = SqliteDict(self.cache_file, flag=flag, encode=encoder, decode=decoder, autocommit=False)
+                self._cache = SqliteDict(self.writeable_cache_file and self.writeable_cache_file.name or self.cache_file, flag=flag, encode=encoder, decode=decoder, autocommit=False)
```
nsoranzo: Can you split this in 2 statements for clarity?
Also the `... and ... or ` is a bit "Perlish", `self.writeable_cache_file.name if self.writeable_cache_file else self.cache_file` is more understandable?

lib/galaxy/tools/search/__init__.py
```diff
@@ -90,11 +90,11 @@ def build_index(self, tool_cache, index_help=True):
...
         for indexed_tool_id in indexed_tool_ids:
             indexed_tool = tool_cache.get_tool_by_id(indexed_tool_id)
-            if not indexed_tool:
-                tool_ids_to_remove.add(indexed_tool_id)
-                continue
-            if not indexed_tool.is_latest_version and not indexed_tool.latest_version.hidden:
-                tool_ids_to_remove.add(indexed_tool_id)
+            if indexed_tool:
+                latest_version = indexed_tool.latest_version
+                if latest_version and latest_version.hidden:
```
nsoranzo, line 97: Not sure if it's on purpose, but this seems functionally different from before, where we didn't remove an old version if the latest is hidden, xref. https://github.com/galaxyproject/galaxy/pull/10175#issuecomment-685001855

mvdbeek: Are you sure ? My brain isn't working super-well at this hour, but I think this is equivalent, i've just inverted the continue.

lib/galaxy/tools/search/__init__.py
```diff
@@ -90,11 +90,11 @@ def build_index(self, tool_cache, index_help=True):
         tool_ids_to_remove = (indexed_tool_ids - set(tool_cache._tool_paths_by_id.keys())).union(tool_cache._removed_tool_ids)
         for indexed_tool_id in indexed_tool_ids:
             indexed_tool = tool_cache.get_tool_by_id(indexed_tool_id)
-            if not indexed_tool:
-                tool_ids_to_remove.add(indexed_tool_id)
-                continue
-            if not indexed_tool.is_latest_version and not indexed_tool.latest_version.hidden:
-                tool_ids_to_remove.add(indexed_tool_id)
+            if indexed_tool:
+                latest_version = indexed_tool.latest_version
```
nsoranzo:
```suggestion
            if indexed_tool:
                if indexed_tool.is_latest_version:
                    continue
                latest_version = indexed_tool.latest_version
```

lib/galaxy/tools/search/__init__.py
nsoranzo: The inverted condition (without the fix) would be:
```python
if indexed_tool.is_latest_version or indexed_tool.latest_version.hidden:
    continue
```

See my suggestion above.

jmchilton (approved):

nsoranzo (approved):

PR #10819, opened 2020-11-27
document optional default for multiple in xsd

bernt-matthias:


lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -2827,7 +2827,7 @@ boxes or radio buttons. Defaults to a drop-down menu select list.</xs:documentat
         <xs:attribute name="multiple" type="PermissiveBoolean">
           <xs:annotation>
             <xs:documentation xml:lang="en">Allow multiple valus to be selected.
-Valid with ``data`` and ``select`` parameters.</xs:documentation>
+Valid with ``data`` and ``select`` parameters. Note, multiple selects are optional by default, single selection is the inverse.</xs:documentation>
```
mvdbeek: What do you think about
```suggestion
Valid with ``data`` and ``select`` parameters. ``select`` parameters with ``multiple="true"`` are optional by default.</xs:documentation>
```
?

mvdbeek (approved):

PR #10820, opened 2020-11-27
export _GALAXY_JOB_HOME_DIR and _GALAXY_JOB_TMP_DIR

bernt-matthias:
these two variables are available in containers because of

`docker run ...-e "_GALAXY_JOB_HOME_DIR=$_GALAXY_JOB_HOME_DIR" -e "_GALAXY_JOB_TMP_DIR=$_GALAXY_JOB_TMP_DIR"`

but since they are not exported the same does not hold for non containerised jobs.

Also moved `export GALAXY_SLOTS` to CLUSTER_SLOTS_STATEMENT.sh. Unsure if this export should be changed to something more complex as in MEMORY_STATEMENT.sh.

### DIFF
--- a/lib/galaxy/jobs/runners/util/job_script/DEFAULT_JOB_FILE_TEMPLATE.sh
+++ b/lib/galaxy/jobs/runners/util/job_script/DEFAULT_JOB_FILE_TEMPLATE.sh
@@ -33,7 +33,6 @@ _GALAXY_VIRTUAL_ENV="$galaxy_virtual_env"
 PRESERVE_GALAXY_ENVIRONMENT="$preserve_python_environment"
 GALAXY_LIB="$galaxy_lib"
 _galaxy_setup_environment "$PRESERVE_GALAXY_ENVIRONMENT"
-export _GALAXY_JOB_HOME_DIR
 export _GALAXY_JOB_TMP_DIR
 GALAXY_PYTHON=`command -v python`
 cd $working_directory
--- a/lib/galaxy/jobs/runners/util/job_script/DEFAULT_JOB_FILE_TEMPLATE.sh
+++ b/lib/galaxy/jobs/runners/util/job_script/DEFAULT_JOB_FILE_TEMPLATE.sh
@@ -27,13 +27,14 @@ _galaxy_setup_environment() {
 
 $integrity_injection
 $slots_statement
-export GALAXY_SLOTS
 export PYTHONWARNINGS="ignore"
 GALAXY_VIRTUAL_ENV="$galaxy_virtual_env"
 _GALAXY_VIRTUAL_ENV="$galaxy_virtual_env"
 PRESERVE_GALAXY_ENVIRONMENT="$preserve_python_environment"
 GALAXY_LIB="$galaxy_lib"
 _galaxy_setup_environment "$PRESERVE_GALAXY_ENVIRONMENT"
+export _GALAXY_JOB_HOME_DIR
+export _GALAXY_JOB_TMP_DIR
 GALAXY_PYTHON=`command -v python`
 cd $working_directory
 $memory_statement
--- a/lib/galaxy/tool_util/xsd/galaxy.xsd
+++ b/lib/galaxy/tool_util/xsd/galaxy.xsd
@@ -3094,8 +3094,7 @@ Name | Description
 ``\${GALAXY_SLOTS:-4}`` | Number of cores/threads allocated by the job runner or resource manager to the tool for the given job (here 4 is the default number of threads to use if running via custom runner that does not configure GALAXY_SLOTS or in an older Galaxy runtime).
 ``\$GALAXY_MEMORY_MB`` | Total amount of memory in megabytes (1024^2 bytes) allocated by the administrator (via the resource manager) to the tool for the given job. If unset, tools should not attempt to limit memory usage.
 ``\$GALAXY_MEMORY_MB_PER_SLOT`` | Amount of memory per slot in megabytes (1024^2 bytes) allocated by the administrator (via the resource manager) to the tool for the given job. If unset, tools should not attempt to limit memory usage.
-``\$_GALAXY_JOB_HOME_DIR`` | Path to an empty directory in the job's woring directory that can be used as a home directory.
-``\$_GALAXY_JOB_TMP_DIR`` | Path to an empty directory in the job's woring directory that can be used as a temporary directory.
+``\$_GALAXY_JOB_TMP_DIR`` | Path to an empty directory in the job's working directory that can be used as a temporary directory.
 
 See the [Planemo docs](https://planemo.readthedocs.io/en/latest/writing_advanced.html#cluster-usage)
 on the topic of ``GALAXY_SLOTS`` for more information and examples.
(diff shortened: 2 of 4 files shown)


lib/galaxy/jobs/runners/util/job_script/DEFAULT_JOB_FILE_TEMPLATE.sh
```diff
@@ -27,7 +27,8 @@ _galaxy_setup_environment() {
 
 $integrity_injection
 $slots_statement
-export GALAXY_SLOTS
+export _GALAXY_JOB_HOME_DIR
+export _GALAXY_JOB_TMP_DIR
```
nsoranzo: Shouldn't these 2 go after calling `_galaxy_setup_environment` below?

nsoranzo (commented):
With your change to GALAXY_SLOTS export, you need to add the `export` to https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/jobs/runners/condor.py#L101-L103

bernt-matthias:
> With your change to GALAXY_SLOTS export, you need to add the `export` to https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/jobs/runners/condor.py#L101-L103

True.

lib/galaxy/jobs/runners/util/job_script/DEFAULT_JOB_FILE_TEMPLATE.sh
bernt-matthias: Yes. They should.

bernt-matthias: What do you think should we also document these variables? Maybe here: https://github.com/simonbray/galaxy/blob/dfd0e0b15f70df091099263a4885860b4fe54c36/lib/galaxy/tool_util/xsd/galaxy.xsd#L2830

nsoranzo (approved):

bernt-matthias:
Would you consider this a bug? I feel like to backporting this ...

lib/galaxy/jobs/runners/util/job_script/DEFAULT_JOB_FILE_TEMPLATE.sh
```diff
@@ -27,13 +27,14 @@ _galaxy_setup_environment() {
...
 $integrity_injection
 $slots_statement
-export GALAXY_SLOTS
 export PYTHONWARNINGS="ignore"
 GALAXY_VIRTUAL_ENV="$galaxy_virtual_env"
 _GALAXY_VIRTUAL_ENV="$galaxy_virtual_env"
 PRESERVE_GALAXY_ENVIRONMENT="$preserve_python_environment"
 GALAXY_LIB="$galaxy_lib"
 _galaxy_setup_environment "$PRESERVE_GALAXY_ENVIRONMENT"
+export _GALAXY_JOB_HOME_DIR
```
nsoranzo:
```suggestion
```

`$env_setup_commands` above should set `HOME` to `$_GALAXY_JOB_HOME_DIR`, so this shouldn't be needed.

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -3094,6 +3094,8 @@ Name | Description
 ``\${GALAXY_SLOTS:-4}`` | Number of cores/threads allocated by the job runner or resource manager to the tool for the given job (here 4 is the default number of threads to use if running via custom runner that does not configure GALAXY_SLOTS or in an older Galaxy runtime).
 ``\$GALAXY_MEMORY_MB`` | Total amount of memory in megabytes (1024^2 bytes) allocated by the administrator (via the resource manager) to the tool for the given job. If unset, tools should not attempt to limit memory usage.
 ``\$GALAXY_MEMORY_MB_PER_SLOT`` | Amount of memory per slot in megabytes (1024^2 bytes) allocated by the administrator (via the resource manager) to the tool for the given job. If unset, tools should not attempt to limit memory usage.
+``\$_GALAXY_JOB_HOME_DIR`` | Path to an empty directory in the job's woring directory that can be used as a home directory.
+``\$_GALAXY_JOB_TMP_DIR`` | Path to an empty directory in the job's woring directory that can be used as a temporary directory.
```
```suggestion
``\$_GALAXY_JOB_TMP_DIR`` | Path to an empty directory in the job's working directory that can be used as a temporary directory.
```

Actually I think it's better to not document `$_GALAXY_JOB_HOME_DIR`, it shouldn't be needed in tools.

lib/galaxy/jobs/runners/util/job_script/DEFAULT_JOB_FILE_TEMPLATE.sh
bernt-matthias: Can you give me a link? Can't find it.

lib/galaxy/tool_util/xsd/galaxy.xsd
bernt-matthias: I have seen tools using `$HOME` to write config files. Which is particularly annoying in real user setups. In case `$HOME` is not set (question above) then I would like to leave it.

bernt-matthias:
This is now included in https://github.com/galaxyproject/galaxy/pull/11448

lib/galaxy/jobs/runners/util/job_script/DEFAULT_JOB_FILE_TEMPLATE.sh
nsoranzo: I cannot give you a simple link unfortunately, it's a complex process:

- `JobWrapper.prepare()` (in `lib/galaxy/jobs/__init__.py`) calls `tool_evaluator.build()` to set its `environment_variables` attribute
- `ToolEvaluator.build()` (in `lib/galaxy/tools/evaluation.py`) calls `self.__build_environment_variables()`
- `ToolEvaluator.__build_environment_variables()` (in `lib/galaxy/tools/evaluation.py`) executes:

  ```python
  home_dir = self.compute_environment.home_directory()
  if home_dir:
      environment_variable = dict(name="HOME", value='"%s"' % home_dir, raw=True)
      environment_variables.append(environment_variable)
  ...
  self.environment_variables = environment_variables
  ```
  Note that `self.compute_environment.home_directory()` is equal to `$_GALAXY_JOB_HOME_DIR` for tools with a profile >= 18.01
- Later, `BaseJobRunner.get_job_file()` (in `lib/galaxy/jobs/runners/__init__.py` ) generates the `env_setup_commands` list; in particular, for each environment variable in `job_wrapper.environment_variables`, a shell export statement is appended to `env_setup_commands` (one of this being `HOME="$_GALAXY_JOB_HOME_DIR"; export HOME`)
- `get_job_file()` passes `env_setup_commands` as parameter to `job_script()`
- `job_script()` (in `lib/galaxy/jobs/runners/util/job_script/__init__.py`) replaces the `$env_setup_commands` template placeholder in `_galaxy_setup_environment()` (in `lib/galaxy/jobs/runners/util/job_script/DEFAULT_JOB_FILE_TEMPLATE.sh` ) with the stringified value of its `env_setup_commands` argument

PR #10821, opened 2020-11-27
[20.09] Drop trigger and update time queries

mvdbeek:
I don't think we currently filter on HDA or HDCA update_time in the UI, so this should work around the observed deadlocks, at the cost of having to re-implement a replacement on 21.01 when we'll merge https://github.com/galaxyproject/galaxy/pull/8441.

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -1140,34 +1140,17 @@ class Job(JobLike, UsesCreateAndUpdateTime, Dictifiable, RepresentById):
     def set_final_state(self, final_state):
         self.set_state(final_state)
         # TODO: migrate to where-in subqueries?
-        statements = ['''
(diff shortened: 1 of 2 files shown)


jmchilton (approved):
Thanks for fixing this, I was so worried about this and put so much effort into trying to cause these issues and really just let the project down by failing to produce the deadlocks. I'm really sorry for breaking this and really appreciate you for cleaning it up.

mvdbeek:
I've updated the PR so that we call `drop_timestamp_trigger` outside of a migration and removed everything that isn't necessart ... since I think it's probably a good idea to keep migrations for updates between releases.

nsoranzo:
@mvdbeek Can I merge this forward as it is?

mvdbeek:
Yes, but I do think we should probably make that a migration.

PR #10822, opened 2020-11-28
Implement read-only conda resolver (alternative)

jmchilton:
Move install stuff into base resolver as recommended by @nsoranzo.

I do not prefer this design but I get it. Replacing a hasattr is cleaner to use but I don't think the dependency resolvers that don't have the option to install should have install junk in their class definition. I'll put both of these out there though and let y'all decide.

### DIFF
--- a/lib/galaxy/tool_util/deps/views.py
+++ b/lib/galaxy/tool_util/deps/views.py
@@ -144,8 +144,8 @@ class DependencyResolversView:
         False if not successful
         """
         resolver = self._dependency_resolver(index)
-        if not hasattr(resolver, "install_dependency"):
-            raise exceptions.NotImplemented()
+        if resolver.read_only:
+            raise exceptions.RequestParameterInvalidException("Attempted to install on a read_only dependency resolver.")
 
         name, version, type, extra_kwds = self._parse_dependency_info(payload)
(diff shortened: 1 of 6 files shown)


lib/galaxy/tool_util/deps/views.py
```diff
@@ -144,8 +144,8 @@ def _install_dependency(self, index, **payload):
         False if not successful
         """
         resolver = self._dependency_resolver(index)
-        if not hasattr(resolver, "install_dependency"):
-            raise exceptions.NotImplemented()
+        if resolver.read_only:
+            raise exceptions.RequestParameterInvalidException("Attempted to install on a read_only dependency resolver.")
```
nsoranzo, line 148:
```suggestion
            raise exceptions.RequestParameterInvalidException("Attempted to install on a read_only dependency resolver.")
        if resolver.disabled:
            raise exceptions.RequestParameterInvalidException("Attempted to install on a disabled dependency resolver.")
```

nsoranzo (approved):
Thanks for implementing the alternative, it looks nicer to me but I guess that's expected :D

jmchilton:
Thanks @nsoranzo !

PR #10823, opened 2020-11-30
Add FastAPI framework as next step for API framework

mvdbeek:
[FastAPI](https://fastapi.tiangolo.com/) does a lot of things (websockets, graphql, Oauth2 + JWT tokens, backend tasks, celery integration, sync + async routes, openapi + typing, good performance, file serving, active community and nice documentation) and can be integrated nicely route-by-route without having to toss out everything at once.

This is a quick experiment where I replaced /api/jobs, IFF you start galaxy with `python scripts/fapi.py -c config/galaxy.yml` (need to activate the virtualenv and install uvicorn). The rest of routes continues to work as normal, so Galaxy is fully functional.

Here's the openAPI doc for the new route:

Screenshot 2020-11-30 at 10 20 24

In terms of challenges:
  - I don't think uwsgi can serve ASGI apps
  - We'd have to replace the interactive tool routing
  - We'd have to get buy-in from admins to abandon uwsgi

TODO for this PR:
- [x] [test API with paste *and* uvicorn until we drop (u)wsgi-only support](3ab8a027ecac9b67ac5ba753fb1f0a1fa2fcb8be)
- [x] [fix job lock location](7d3c6db3ce7901e80ced461575d151e199b74cd2)


hexylena:
1. that docs picture is gorgeous, it would make life so much better.
2. No one who deploys ITs in prod uses that routing anyway. And we should really switch to path based proxying (i.e. GIE style) which would address the local routing issues, and then we could toss out this uwsgi routing completely.
3. (as an admin) I've accrued a lot of knowledge and scripting/ansibling to handle uwsgi/deployment. It would be sad to throw out completely, but, sounds like some positive improvements for devs?

mvdbeek:
> 3\. (as an admin) I've accrued a lot of knowledge and scripting/ansibling to handle uwsgi/deployment. It would be sad to throw out completely, but, sounds like some positive improvements for devs?

I'd also be curious about what you think of gunicorn as a process manager, which seems like it's the recommended route for production. https://docs.gunicorn.org/en/latest/signals.html#master-process lists some signals it can handle. Reading this naively it sounds like it can also do zero-downtime restarts, as a sort of replacement for zerglings ?

hexylena:
That potentially looks OK, that there's a master/worker split. I'm not as big of a fan of signals as Nate is, but it looks potentially nice?

scripts/galaxy_main.py
```diff
@@ -0,0 +1 @@
+galaxy-main
```
nsoranzo: Maybe `git mv scripts/galaxy-main scripts/galaxy_main.py` and symlink `galaxy-main` to `galaxy_main.py` instead?

scripts/fapi.py
```diff
@@ -0,0 +1,32 @@
+import uvicorn
```
nsoranzo, line 1: Need to add `uvicorn` to the requirements?

mvdbeek: It's not needed for startup, so as a development-only thing we'd probably move it in dev-requirements ?

Nerdinacan:
+1 for websocket support. I know there's a lot more stuff we'd have to do server side to actually generate a subscription channel with changes that we could consume, but I love that it would be possible.

lib/galaxy/webapps/base/webapp.py
```diff
@@ -23,7 +23,9 @@
 from sqlalchemy.orm.exc import NoResultFound
 
 from galaxy import util
-from galaxy.exceptions import ConfigurationError, MessageException
+from galaxy.exceptions import AuthenticationFailed, ConfigurationError, MessageException
```
nsoranzo:
```suggestion
from galaxy.exceptions import (
    AuthenticationFailed,
    ConfigurationError,
    MessageException,
)
```

lib/galaxy/webapps/galaxy/api/dependencies.py
```diff
@@ -0,0 +1,91 @@
+"""
+This module *does not* contain API routes. It exclusively contains dependencies to be used in FastAPI routes
```
nsoranzo: Maybe this stuff can go in `lib/galaxy/webapps/galaxy/api/__init__.py` ? That would be less confusing, I think.

mvdbeek: Yep, done that for now. I suppose this will grow significantly, but I think that's a good location for now.

lib/galaxy/webapps/base/webapp.py
mvdbeek: Added that in a rebase, thanks!

scripts/galaxy_main.py
mvdbeek: Done that in a rebase!

mvdbeek:
This is by no means complete, but I think it is mergeable. That would make distributing work on various parts easier (xref: https://github.com/galaxyproject/galaxy/projects/20)

mvdbeek:
@nsoranzo do you know why `make update-dependencies` didn't respect pygithub pinning requests to <2.25 ?

nsoranzo:
> @nsoranzo do you know why `make update-dependencies` didn't respect pygithub pinning requests to <2.25 ?

I don't have time to investigate in deep at the moment, you can try to regenerate the Docker image with: `make -C lib/galaxy/dependencies/pipfiles/docker/ clean` and re-run `make update-dependencies`

mvdbeek:
Thanks, no luck there, but I think the manual pin shouldn't be a big deal for now.

jmchilton:
Can you integrate some of the idea from:

- jmchilton/galaxy@0f9279f (add option to integrate testing using uvicorn)
- jmchilton/galaxy@e094706 (clearer transition path for older code,  less duplication, and pushing more logic out of controllers that are tied to particular frameworks)

I know it is a little more work to rework the old APIs as we migrate the new ones - but I think there is value. It pushes more logic out of the controller layer if nothing else, and then it puts us in a better position if we're not able to fully make the transition within one or two releases.

jmchilton:
I didn't mean to insist that all the tests pass under uvicorn, just wanted to have the option to run them so I could play with the branch.

mvdbeek:
Yay, it's green! I'll spin out the zipstream changes for a cleaner PR and rebase, so on the off chance please don't merge it yet.

lib/galaxy/webapps/galaxy/api/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+def get_db(app: UniverseApplication = Depends(get_app)) -> Session:
+    # TODO: return sqlachemy 2.0 style session without autocommit and expire_on_commit!
+    return app.model.session
+
+
+def get_user_manager(app: UniverseApplication = Depends(get_app)):
+    return UserManager(app)
+
+
+def get_session_manager(app=Depends(get_app)) -> GalaxySessionManager:
```
davelopez: Nice to see some type annotations kicking in! :)
```suggestion
def get_session_manager(app: UniverseApplication = Depends(get_app)) -> GalaxySessionManager:
```

lib/galaxy/webapps/galaxy/api/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+def get_app() -> UniverseApplication:
+    return galaxy_app.app
+
+
+def get_db(app: UniverseApplication = Depends(get_app)) -> Session:
+    # TODO: return sqlachemy 2.0 style session without autocommit and expire_on_commit!
+    return app.model.session
+
+
+def get_user_manager(app: UniverseApplication = Depends(get_app)):
```
davelopez:
```suggestion
def get_user_manager(app: UniverseApplication = Depends(get_app)) -> UserManager:
```

lib/galaxy/webapps/galaxy/api/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+        raise HTTPException(status_code=e.status_code, detail=str(e))
+
+
+def get_user(galaxy_session: Optional[model.GalaxySession] = Depends(get_session), api_user: Optional[User] = Depends(get_api_user)) -> Optional[User]:
+    if galaxy_session:
+        return galaxy_session.user
+    return api_user
+
+
+def get_trans(app=Depends(get_app), user: Optional[User] = Depends(get_user),
```
davelopez:
```suggestion
def get_trans(app: UniverseApplication = Depends(get_app), user: Optional[User] = Depends(get_user),
```

lib/galaxy/webapps/galaxy/api/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+              galaxy_session: Optional[model.GalaxySession] = Depends(get_session),
+              ) -> SessionRequestContext:
+    app.model.session.expunge_all()
+    return SessionRequestContext(app=app, user=user, galaxy_session=galaxy_session)
+
+
+def get_admin_user(trans: SessionRequestContext = Depends(get_trans), user_manager: UserManager = Depends(get_user_manager)):
```
davelopez: I'm not 100% sure if the return type can be `User` or if there is some specific `AdminUser` class or something like that.
```suggestion
def get_admin_user(trans: SessionRequestContext = Depends(get_trans), user_manager: UserManager = Depends(get_user_manager)) -> Optional[User]:
```

davelopez (commented):
Love to see type hints spreading, they make my IDE and me happy :)

lib/galaxy/webapps/galaxy/api/__init__.py
mvdbeek: There is a `BooststrapAdminUser` now, so this could be a typing.Union[User, BootstrapAdminUser].
I'm not 100% convinced it makes sense to annotate everything, since in this case I guess we'd have to import the `BooststrapAdminUser` just to create the type annotation. Which I think is fine in this case, but I could see this becoming a burden in other cases.

davelopez: This is really a good point. I haven't thought of this burden because I have used them only in a small project.
Should we come up with some kind of Type Annotation Guidelines for Galaxy? or are there any rules or best practices available for this matter?

mvdbeek: Yes, that is a good idea, but maybe we can freestyle it a bit while we're still getting to know the pros and cons?
My rough thinking is that for everything that we consider public (like API routes) we should annotate all parameters and return types. For things that are really dynamic we might make it optional ? For functions that are meant for re-use, try to include type annotations ?

davelopez: That makes perfect sense. Totally agree!

jmchilton: I think we should probably annotate an alternative admin transaction so we don't need to inject an admin user that isn't used? Or just reuse the old decorator to be honest.

One thing is we're annotating explicit concrete classes for everything and it is causing me some distress - we're creating very explicit dependencies and deep transitive dependencies. If we're going to use types internally for things like app and trans we should be using more precise interfaces that they implement and that can be imported without bringing all of say app's dependencies.

jmchilton:
Hacking a bunch of this I think it is sort of clear to me that we should probably break a bunch of bad habits the old API allowed and we should track the breakage and include it in the release notes. I think a long deprecation process will be difficult with the new type annotations and such that force best practices.

The list so far would include:

- The roles API no longer accept form-data style post parameters.
- The show endpoint for roles now returns a single role instead of a singleton list.

lib/galaxy/webapps/galaxy/api/jobs.py
```diff
@@ -30,10 +43,62 @@
...
+
+
+@cbv(router)
+class FastAPIJobs:
+    job_manager: JobManager = Depends(get_job_manager)
+    job_search: JobSearch = Depends(get_job_search)
+    hda_manager: hdas.HDAManager = Depends(get_hda_manager)
+
+    @router.get('/job_lock')
+    def job_lock_status(self, admin_user=Depends(get_admin_user)) -> JobLock:
```
jmchilton: Is it intentional that this has a different route than the existing controller - because I think the existing route is api/job_lock not api/jobs/job_lock?

mvdbeek: No, just an oversight.

mvdbeek:
> * The roles API no longer accept form-data style post parameters.

This is a good thing IMO, but I just wanted to point out that there is form handling (https://fastapi.tiangolo.com/tutorial/request-forms/) should we have a legitimate need.

jmchilton:
Brilliant work!

PR #10828, opened 2020-11-30
Drop legacy_eager_objectstore_initialization option.

jmchilton:
Reduce branching/complexity around when/where objectstore is getting set for downstream work.


mvdbeek:
Thanks @jmchilton!

PR #10830, opened 2020-12-01
Rely on emails to determine GitLab users

selten:
Rely on emails to determine the GitLab users instead of the username. Usernames are less stable in terms of searching compared to emails.


jmchilton:
Makes sense to me, thanks so much!

PR #10813, opened 2020-11-26
Maintainance Bot with GithubActions

OlegZharkov:
After @nsoranzo suggested to re-implementing p4 bot in gha, I dig a bit and probably it's a good idea to add dynamic labeling and not always label every PR with `triage`. 

This is the first step to do so. If you know another very cool matching label and/or path in [here](https://github.com/galaxyproject/galaxy/compare/dev...OlegZharkov:gha-bot?expand=1#diff-a22c263686553013feaeb0677d26eeb0b8778a756c4311c1fce13384258026aa), please comment or directly contribute right in this PR.

A working example of gha bot creating labels:
https://github.com/OlegZharkov/galaxy/pull/5

Screenshot_20201126_233649

PR #10836, opened 2020-12-01
Fix Label Verifier in Maintenance Bot 

OlegZharkov:
Following discussion in https://github.com/galaxyproject/galaxy/pull/10813#issuecomment-736538688
So what was happening is that github doesn't allow using `GITHUB_TOKEN` in gh-actions on PRs from forked branches ([example](https://github.com/galaxyproject/galaxy/actions/runs/393408692), we got 403 even though John has write access). 

However, if you change  `on: pull_request` to `on: pull_request_target` this particular job will run from from the base of the pull request . Since it's considered as "trusted sources" by github, it will allow operations with `GITHUB_TOKEN`, thus allowing us to create a comment. ([if you want to read more](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/))

So this PR does exactly this. It will fix label-verifier and will create a comment if PR wasn't merge with 'kind/'

examples:
merged without kind https://github.com/OlegZharkov/galaxy/pull/79
merged with kind https://github.com/OlegZharkov/galaxy/pull/80

P.S.
Well, since I am here already, I also updated `labeler.yml`

### DIFF
--- a/.github/workflows/maintainance_bot.yaml
+++ b/.github/workflows/maintainance_bot.yaml
@@ -4,26 +4,23 @@ on:
 
 jobs:
   labeler:
-    name: "Assign Labels"
+    name: Assign labels and milestone
     runs-on: ubuntu-latest
+    env:
+      MILESTONE_NUMBER: 18
     steps:
       - uses: actions/labeler@main
         with:
           repo-token: "${{ secrets.GITHUB_TOKEN }}"
-
-  milestone:
-    name: "Assign Milestone"
-    runs-on: ubuntu-latest
-    steps:
-      - name: Assign Milestone
-        if:  |
-          contains(github.event.pull_request.title, 'WIP') == false &&
-          contains(github.event.pull_request.title, 'merge') == false &&
-          github.event.pull_request.draft == false
-        env:
-          MILESTONE_NUMBER: 18
+      - name: Assign milestone
+        if: |
+          ! github.event.pull_request.milestone &&
+          ! contains(github.event.pull_request.labels.*.name, 'merge') &&
+          ! contains(github.event.pull_request.labels.*.name, 'status/WIP') &&
+          ! contains(github.event.pull_request.title, 'WIP') &&
+          ! github.event.pull_request.draft
         run: |
           curl --request PATCH \
           --url ${{ github.event.pull_request.issue_url }} \
           --header 'authorization: token ${{ secrets.GITHUB_TOKEN }}' \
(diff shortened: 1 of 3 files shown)


nsoranzo:
Thanks, that's indeed the correct approach! Do you mind if I make some changes to `.github/labeler.yml` ?

OlegZharkov:
> Thanks, that's indeed the correct approach! Do you mind if I make some changes to `.github/labeler.yml` ?

yes, please, go ahead

nsoranzo (approved):

OlegZharkov:
Hmm, 13 job failed. Related to https://www.githubstatus.com/incidents/sxthxvdlljk4?

nsoranzo:
@OlegZharkov I've pushed here a commit with changes to `.github/workflows/maintainance_bot.yaml` , hope that's OK. It works fine in my tests but please have a look.

.github/workflows/maintainance_bot.yaml
```diff
@@ -4,26 +4,23 @@ on:
...
-          github.event.pull_request.draft == false
-        env:
-          MILESTONE_NUMBER: 18
+      - name: Assign milestone
+        if: |
+          ! github.event.pull_request.milestone &&
+          ! contains(github.event.pull_request.labels.*.name, 'merge') &&
+          ! contains(github.event.pull_request.labels.*.name, 'status/WIP') &&
+          ! contains(github.event.pull_request.title, 'WIP') &&
+          ! github.event.pull_request.draft
```
OlegZharkov, line 21: do we have label `merge`? Anyway, I also saw negation `!` operator in gha docs, but it didn't work for me. I will try this out now.

OlegZharkov: hm, maybe space was the trick.  `!` works fine, I tried these changes and everything works perfectly. Thanks a bunch @nsoranzo

nsoranzo: > do we have label `merge`? Anyway, I also saw negation `!` operator in gha docs, but it didn't work for me. I will try this out now.

Yes: https://github.com/galaxyproject/galaxy/labels/merge . It's used for backport PRs, so that these are not picked up by the script that generate the release notes. Actually the labeler should also skip them.

nsoranzo (approved):

jmchilton:
Very exciting - really awesome stuff!

OlegZharkov:
thanks

PR #10838, opened 2020-12-01
Add <filter> inline python example for "select" type

tensionhead:


lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -4412,6 +4412,30 @@ would appear as ``$options.vcf_output`` having the values ``'--vcf'`` when true
       </data>
     </outputs>
 ```
+
+Variables that have the type attribute ``select`` need a special python statement of
+the form ``selection and "option" in selection`` inside the ``<filter>`` tags.
```
bernt-matthias:
```suggestion
Note that variables that correspond to optional select parameters are `None` if nothing is selected. Therefore a filter for such a variable looks like the following example.
```

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -4412,6 +4412,30 @@ would appear as ``$options.vcf_output`` having the values ``'--vcf'`` when true
...
 ```
+
+Variables that have the type attribute ``select`` need a special python statement of
+the form ``selection and "option" in selection`` inside the ``<filter>`` tags. 
+
+### Example
+
+```xml
+    <inputs>
+       <param name="outputs" type="select" optional="true">
```
bernt-matthias:
```suggestion
       <param name="output_type" type="select" optional="true">
```

outputs is confusing

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -4412,6 +4412,30 @@ would appear as ``$options.vcf_output`` having the values ``'--vcf'`` when true
...
+            <option value="save_phase">Phase Movie</option>
+            <option value="save_period">Period Movie</option>
+        </param>
+    </inputs>
+
+    <outputs>
+        <data name="phase_out" format="tiff" label="${movie.name[:-4]}_phase">
+            <filter>outputs and "save_phase" in outputs</filter>
+        </data>
```
bernt-matthias:
```suggestion
        <data name="phase_out" format="tiff">
            <filter>output_type and "save_phase" in output_type</filter>
        </data>
```

would suggest not to advocate such a non standard label

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -4412,6 +4412,30 @@ would appear as ``$options.vcf_output`` having the values ``'--vcf'`` when true
...
+        <data name="period_out" format="tiff" label="${movie.name[:-4]}_period">
+            <filter>outputs and "save_period" in outputs</filter>
+        </data>
```
bernt-matthias:
```suggestion
...
```

I think one is sufficient for the example

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -4387,7 +4387,7 @@ appear as ``$options.selection_mode`` in Cheetah. Similarly ``options["vcf_outpu
 would appear as ``$options.vcf_output`` having the values ``'--vcf'`` when true and
 ``''`` when false in Cheetah.
 
-### Example
+### Example
```
bernt-matthias: Can you remove the trailing space

lib/galaxy/tool_util/xsd/galaxy.xsd
bernt-matthias: Maybe also describe what the variable looks like in case of `multiple="true"`

lib/galaxy/tool_util/xsd/galaxy.xsd
tensionhead: agreed..

lib/galaxy/tool_util/xsd/galaxy.xsd
tensionhead: ok, for the exmaple I understand. For my tool getting rid of the `.tif` improved the outputs filenames quite a lot, otherwise you get input.tif.tiff and such things..

lib/galaxy/tool_util/xsd/galaxy.xsd
tensionhead: Your "therefore" is quite strong.. the magic python statement still doesn't make much sense outside of this very special context...

lib/galaxy/tool_util/xsd/galaxy.xsd
tensionhead: I thought 2 makes sense, as for one alone a boolean is enough?

lib/galaxy/tool_util/xsd/galaxy.xsd
bernt-matthias: I think this is pretty standard. If you have a variable that is `None` if it is not set then you have to do this. One could be more explicit and write `outputs is not None and "save_period" in outputs` but this is equivalent. With the first statement you just assert that the outputs is set, i.e. it is a string, and the `in` operator works.

lib/galaxy/tool_util/xsd/galaxy.xsd
bernt-matthias: An alternative would be to link to an existing tool instead of giving the example here. This is also often done in this file. My intention was to keep it short. But I'm also fine with 2.

lib/galaxy/tool_util/xsd/galaxy.xsd
tensionhead: yes you are right!

lib/galaxy/tool_util/xsd/galaxy.xsd
bgruening: @bernt-matthias we will use this documentation more often in editors and interactive hints. I guess there is some advantage in the future to have small examples as part of the documentation.

bgruening: e.g. in https://github.com/galaxyproject/galaxy-language-server

tensionhead:
Is there anything left to resolve?

bgruening (approved):

bgruening:
Thanks @tensionhead, this looks great to me!

bernt-matthias (approved):

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -4412,6 +4412,30 @@ would appear as ``$options.vcf_output`` having the values ``'--vcf'`` when true
...
+            <option value="save_period">Period Movie</option>
+        </param>
+    </inputs>
+
+    <outputs>
+        <data name="phase_out" format="tiff">
+            <filter>output_type and "save_phase" in output_type</filter>
+        </data>
+        <data name="period_out" format="tiff" label="${movie.name[:-4]}_period">
+            <filter>outputs and "save_period" in outputs</filter>
```
```suggestion
            <filter>output_type and "save_period" in output_type</filter>
```

bernt-matthias: This might also be an opportunity to show/explain that filters are applied successively: 

```
<filter>output_type</filter>
<filter>"save_period" in output_type</filter>
```

tensionhead: Not sure why `output_type` should be more clear here

bernt-matthias: You mean the suggestion of @nsoranzo ? The parameter is called `output_type` (not `output`).

If you mean my comment then you might go: 

```
<filter>output_type is not None</filter>
<filter>"save_period" in output_type</filter>
```

mentioning that filters are called successively and then tell the reader that an equivalent short form is: 

```
<filter>output_type and "save_period" in output_type</filter>
```

tensionhead: The name of this parameter is totally free from my understanding

tensionhead:
I guess I am not the right person to finish this

bernt-matthias:
> I guess I am not the right person to finish this

@tensionhead  Clearly you are :) and your contribution is very much appreciated .. I think the only point left to do is to accept the suggestion of @nsoranzo . 

The point is that in the filter for `period_out` you used `outputs`, but the variable is called `output_type` (this is already correct in the filter for `phase_out`).

nsoranzo (approved):

tensionhead:
@bernt-matthias Aah, ya sorry my bad.. got confused with the little code snippets shown during the reviews. Wee if u guys are happy now, I am too :)

nsoranzo:
Thanks @tensionhead !

PR #10844, opened 2020-12-02
Use `github-script` action for maintenance_bot job too

nsoranzo:
Also:
- fix typo in filename
- more explicit binding of argument values

### DIFF
--- a/.github/workflows/labels-verifier.yaml
+++ b/.github/workflows/labels-verifier.yaml
@@ -13,9 +13,13 @@ jobs:
           github.event.pull_request.merged == true &&
           ! contains(join(github.event.pull_request.labels.*.name, ', '), 'kind/') &&
           ! contains(github.event.pull_request.labels.*.name, 'merge')
-        uses: actions/github-script@0.3.0
+        uses: actions/github-script@v3
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
-            const { issue: { number: issue_number }, repo: { owner, repo }  } = context;
-            github.issues.createComment({ issue_number, owner, repo, body: 'This PR was merged without a "kind/" label, please correct.' })
+            github.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              body: 'This PR was merged without a "kind/" label, please correct.',
+            })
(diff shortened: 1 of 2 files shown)


nsoranzo:
Ping @OlegZharkov

.github/workflows/maintenance_bot.yaml
```diff
@@ -19,8 +19,13 @@ jobs:
...
+        uses: actions/github-script@v3
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            github.issues.update({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              milestone: ${{ env.MILESTONE_NUMBER }},
+            });
```
OlegZharkov, line 31: Oh, that's sweet. I wish I knew this earlier

.github/workflows/labels-verifier.yaml
```diff
@@ -13,9 +13,13 @@ jobs:
...
           github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
-            const { issue: { number: issue_number }, repo: { owner, repo }  } = context;
-            github.issues.createComment({ issue_number, owner, repo, body: 'This PR was merged without a "kind/" label, please correct.' })
+            github.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              body: 'This PR was merged without a "kind/" label, please correct.',
+            })
```
OlegZharkov, line 25: to be fair 2 lines solution looked more concise. But we can use this one for consistency :+1:

dannon: Inarguably more concise, but I think this is more readable :)

.github/workflows/maintenance_bot.yaml
nsoranzo: I found out by looking at the [`github-script` action](https://github.com/actions/github-script) you used for the labels verifier, which refers to https://octokit.github.io/rest.js/ . Just replace `octokit.` with `github.` and you have all the power of the GitHub REST API toolkit!

.github/workflows/labels-verifier.yaml
nsoranzo: Indeed I went for readability, especially since the order of arguments here was different from the one on https://octokit.github.io/rest.js/v18#issues-create-comment , so I think specifying the argument names is also safer.

PR #10847, opened 2020-12-03
Always expand tool about section if metadata is available

### LINKED ISSUE #10839: Hide 'About this tool' if content has not been specified
image

guerler:
Now that the tool about section has been integrated, it seems to be better to always show the `About this tool` content without requiring the user to manually expand the area. Also fixes #10839.

### DIFF
--- a/client/src/components/Tool/ToolFooter.vue
+++ b/client/src/components/Tool/ToolFooter.vue
@@ -1,90 +1,71 @@
 <template>
-    <div class="tool-footer">
-        <b-link
-            v-if="hasContent"
-            :aria-expanded="expanded"
-            aria-controls="collapse-about"
-            class="collapse-about"
-            @click="expanded = !expanded"
-        >
-            About this tool
-            <font-awesome-icon v-if="expanded" icon="angle-double-up" />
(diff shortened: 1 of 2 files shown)


jmchilton:
I remember being asked to hide more details by default - but I'm not sure I have an opinion either way to be honest so I'm totally fine with this. Definitely appreciate the bug fix aspect though - we shouldn't be showing an empty metadata section!

jmchilton (approved):

client/src/components/Tool/ToolFooter.vue
```diff
@@ -1,88 +1,71 @@
...
-                    </div>
-                </div>
-                <div v-if="creators && creators.length > 0" class="metadata-section">
-                    <span class="metadata-key">Creators:</span>
-                    <Creators :creators="creators" />
-                </div>
-            </b-card>
-        </b-collapse>
-    </div>
+    <b-card class="tool-footer" v-if="hasContent">
```
mvdbeek, line 2: Can you adjust the selenium test as appropriate then ?
It's looking for `collapse-about`

PR #10852, opened 2020-12-04
[20.09] Close tempfile file descriptor

mvdbeek:
Fixes https://github.com/galaxyproject/ephemeris/issues/166

### DIFF
--- a/lib/galaxy/datatypes/converters/pbed_ldreduced_converter.py
+++ b/lib/galaxy/datatypes/converters/pbed_ldreduced_converter.py
@@ -34,7 +34,8 @@ def timenow():
 def pruneLD(plinktasks=[], cd='./', vclbase=[]):
     """
     """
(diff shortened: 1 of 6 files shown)


mvdbeek:
I systematically looked at the code base and found a couple more instances where we don't close the fd returned by mkstemp in https://github.com/galaxyproject/galaxy/pull/10852/commits/13f597b0ff71372f8a3dc509656ad747156e954f

bernt-matthias:
Wouldn't `with` blocks be even better?

nsoranzo:
> Wouldn't `with` blocks be even better?

I also generally prefer to use `with tempfile.NamedTemporaryFile(..., delete=False) as`

jmchilton (approved):

mvdbeek:
I'd prefer not to refactor on a released branch (there are more mkstemp to fix beyond the cases that are problematic).

Issue #10737, opened 2020-11-17
Update add tag icon


afgane:
Add tag icon appears from a different era of icons compared to most icons in the UI so would be nice to bring it up to date. Two examples where I've come across this icon is when tagging a history
![image](https://user-images.githubusercontent.com/338642/99429991-bafa2480-28d6-11eb-97e1-258e8fa48663.png)

When viewing a list of histories:
![image](https://user-images.githubusercontent.com/338642/99430021-c77e7d00-28d6-11eb-9c76-40f05a439341.png)

Related to the page listing all histories, I don't see a point in the `Status` column which only ever states `Current history`, yet the current history is highlighted in green so I feel that's enough of a highlight and the column could be removed?

OlegZharkov:
can we close this? @astrovsky01

astrovsky01:
Yup, thanks for the ping!

PR #10853, opened 2020-12-04
Switch out tag image

astrovsky01:
Adding in a new image of tags for consistency. This is only a temporary fix while the current package is removed and rewritten. The existing package is too static and can't be modified, so @Nerdinacan and I are going to write a new, more flexible one.


astrovsky01:
Addressing https://github.com/galaxyproject/galaxy/issues/10737

PR #10866, opened 2020-12-07
Don't fetch dev on dev. Map AWS secrets to environment variables

nsoranzo:
Fix the error:

```
fatal: Refusing to fetch into current branch refs/heads/dev of non-bare repository
```


mvdbeek:
Thank you!

PR #10871, opened 2020-12-08
Allow build_list inputs to be optional

mvdbeek:
https://github.com/galaxyproject/galaxy/pull/10863 + fix tests by adding `__BUILD_LIST__ ` to `WORKFLOW_SAFE_TOOL_VERSION_UPDATES `

### DIFF
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -2864,8 +2864,8 @@ class BuildListCollectionTool(DatabaseOperationTool):
         new_elements = OrderedDict()
 
         for i, incoming_repeat in enumerate(incoming["datasets"]):
-            new_dataset = incoming_repeat["input"].copy(copy_tags=tags)
-            new_elements["%d" % i] = new_dataset
+            if incoming_repeat["input"]:
+                new_elements["%d" % i] = incoming_repeat["input"].copy(copy_tags=tags)
 
         self._add_datasets_to_history(history, new_elements.values())
         output_collections.create_collection(
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -192,7 +192,9 @@ GALAXY_LIB_TOOLS_VERSIONED = {
 safe_update = collections.namedtuple("SafeUpdate", "min_version current_version")
 # Tool updates that did not change parameters in a way that requires rebuilding workflows
 WORKFLOW_SAFE_TOOL_VERSION_UPDATES = {
-    'Filter1': safe_update(packaging.version.parse("1.1.0"), packaging.version.parse("1.1.1"))
+    'Filter1': safe_update(packaging.version.parse("1.1.0"), packaging.version.parse("1.1.1")),
+    '__BUILD_LIST__': safe_update(packaging.version.parse("1.0.0"), packaging.version.parse("1.0.1")),
+    '__APPLY_RULES__': safe_update(packaging.version.parse("1.0.0"), packaging.version.parse("1.1.0")),
 }
 
 
--- a/lib/galaxy/tools/build_list.xml
+++ b/lib/galaxy/tools/build_list.xml
@@ -1,6 +1,6 @@
 <tool id="__BUILD_LIST__"
       name="Build List"
-      version="1.0.0"
+      version="1.1.0"
       tool_type="build_list">
   <description>from one or more datasets</description>
   <type class="BuildListCollectionTool" module="galaxy.tools" />
@@ -8,7 +8,7 @@
           class="ModelOperationToolAction"/>
   <inputs>
     <repeat name="datasets" title="Dataset">
-      <param type="data" name="input" label="Input Dataset" />
+      <param type="data" name="input" optional="true" label="Input Dataset" />
     </repeat>
   </inputs>
   <outputs>
@@ -33,6 +33,12 @@ This tool will create a new collection from your history datasets but your quota
               <has_text_matching expression="^This is a line of text.\n$"/>
           </assert_contents>
         </element>
+      </output_collection>
+    </test>
+    <test>
+      <repeat name="datasets">
+      </repeat>
+      <output_collection name="output" type="list" count="0">
       </output_collection>
     </test>
   </tests>


jmchilton (approved):

nsoranzo:
I was working in this alternative, which passes tests:

```diff
diff --git a/lib/galaxy/tools/__init__.py b/lib/galaxy/tools/__init__.py
index 41a7f1f503..720242ba1d 100644
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -2863,8 +2863,8 @@ class BuildListCollectionTool(DatabaseOperationTool):
     def produce_outputs(self, trans, out_data, output_collections, incoming, history, tags=None, **kwds):
         new_elements = OrderedDict()
 
-        for i, incoming_repeat in enumerate(incoming["datasets"]):
-            new_dataset = incoming_repeat["input"].copy(copy_tags=tags)
+        for i, incoming_dataset in enumerate(incoming["inputs"]):
+            new_dataset = incoming_dataset.copy(copy_tags=tags)
             new_elements["%d" % i] = new_dataset
 
         self._add_datasets_to_history(history, new_elements.values())
diff --git a/lib/galaxy/tools/build_list.xml b/lib/galaxy/tools/build_list.xml
index b7bca228ea..5410362fe1 100644
--- a/lib/galaxy/tools/build_list.xml
+++ b/lib/galaxy/tools/build_list.xml
@@ -1,32 +1,26 @@
 <tool id="__BUILD_LIST__"
       name="Build List"
-      version="1.0.0"
+      version="1.2.0"
       tool_type="build_list">
     <description>from one or more datasets</description>
     <type class="BuildListCollectionTool" module="galaxy.tools" />
     <action module="galaxy.tools.actions.model_operations"
             class="ModelOperationToolAction"/>
     <inputs>
-    <repeat name="datasets" title="Dataset">
-      <param type="data" name="input" label="Input Dataset" />
-    </repeat>
+        <param name="inputs" type="data" multiple="true" label="Input Datasets" />
     </inputs>
     <outputs>
         <collection name="output" type="list" label="${on_string} (as list)">
         </collection>
     </outputs>
     <help><![CDATA[
-
 .. class:: infomark
 
 This tool will create a new collection from your history datasets but your quota usage will not increase.
-
     ]]></help>
     <tests>
         <test>
-      <repeat name="datasets">
-        <param name="input" value="simple_line.txt" />
-      </repeat>
+            <param name="inputs" value="simple_line.txt" />
             <output_collection name="output" type="list">
                 <element name="0">
                     <assert_contents>
diff --git a/lib/galaxy_test/api/test_workflows.py b/lib/galaxy_test/api/test_workflows.py
index 13bba6161b..3c0467c00e 100644
--- a/lib/galaxy_test/api/test_workflows.py
+++ b/lib/galaxy_test/api/test_workflows.py
@@ -2841,7 +2841,7 @@ inputs:
 steps:
   - tool_id: __BUILD_LIST__
     in:
-      datasets_0|input: input1
+      inputs: input1
     outputs:
       output:
         hide: true
@@ -2866,7 +2866,7 @@ inputs:
 steps:
   - tool_id: __BUILD_LIST__
     in:
-      datasets_0|input: input1
+      inputs: input1
     outputs:
       output:
         delete_intermediate_datasets: true
@@ -2895,13 +2895,13 @@ inputs:
 steps:
   - tool_id: __BUILD_LIST__
     in:
-      datasets_0|input: input1
+      inputs: input1
     outputs:
       output:
         change_datatype: txt
   - tool_id: __BUILD_LIST__
     in:
-      datasets_0|input: input1
+      inputs: input1
 """, test_data="""
 input1:
   value: 1.fasta
@@ -2929,7 +2929,7 @@ inputs:
 steps:
   - tool_id: __BUILD_LIST__
     in:
-      datasets_0|input: input1
+      inputs: input1
     outputs:
       output:
         rename: "my new name"
diff --git a/lib/galaxy_test/base/data/test_subworkflow_with_integer_input.ga b/lib/galaxy_test/base/data/test_subworkflow_with_integer_input.ga
index 92c05ff1bb..fe874904f3 100644
--- a/lib/galaxy_test/base/data/test_subworkflow_with_integer_input.ga
+++ b/lib/galaxy_test/base/data/test_subworkflow_with_integer_input.ga
@@ -187,7 +187,7 @@
                         "errors": null,
                         "id": 2,
                         "input_connections": {
-                            "datasets_0|input": {
+                            "inputs": {
                                 "id": 0,
                                 "output_name": "output"
                             }
@@ -213,8 +213,8 @@
                         },
                         "post_job_actions": {},
                         "tool_id": "__BUILD_LIST__",
-                        "tool_state": "{\"datasets\": [{\"__index__\": 0, \"input\": {\"__class__\": \"ConnectedValue\"}}], \"__page__\": null, \"__rerun_remap_job_id__\": null}",
-                        "tool_version": "1.0.0",
+                        "tool_state": "{\"inputs\": {\"__class__\": \"ConnectedValue\"}, \"__page__\": null, \"__rerun_remap_job_id__\": null}",
+                        "tool_version": "1.2.0",
                         "type": "tool",
                         "uuid": "a20e88c1-dbe0-4b38-90e0-7d5e02d928c0",
                         "workflow_outputs": [
```

I think that in most cases a multiple="true" data param is more user-friendly than a repeat, although this change would break workflows using the old version of this tool.

@jmchilton I don't think I understand your comment from Gitter: "The repeat was definitely intentional 🧐, the main use case here is mapping over that tool to produce nested lists - I think that works better with normal data inputs. I could be wrong though"
Do you have an example where the multiple param doesn't work?

jmchilton:
This tool isn't meant to be used in the tool form - it is for workflows. For workflows - the repeats are more usable.

The point of this tool is to pass in a fixed number (n) of lists and build partitioned nested lists (list of n-tuples may be way of thinking about it). It is much less clear what is going on in the workflow editor if you use a multiple data parameter. Also the order is super important in that use case - and not even shown in the workflow editor. I'm not sure you can map multiple collections in this fashion over a single parameter at all - and even if you can it is much less clear.

The original PR that added this was about building up tuple-like collections - I really think repeats are better at representing that. I'm a big fan of generally replacing repeats with multiple data parameters - I implemented that and fought for that for a long time. There are scenarios where the repeat makes more sense though and I think this is one.

nsoranzo:
@jmchilton Thanks a lot for the explanation, makes sense now!

PR #10874, opened 2020-12-08
implement a 'duplicate_column_filter'

bernt-matthias:
assume there is a data table with 2 columns: name and value (e.g. mothur data tables and probably many other old ones).

tools using such data tables can not be tested (to the best of my knowledge) since one needs to specify the value .. which is a path that changes with the test environment.

my first idea was to add a filter that duplicates a column, i.e. (name, value) can be transformed into (name, name, value)

this already works (tested for one mothur tool), but only for accessing the columns by index. accessing the columns by name does not work, since the names still refer to the old columns (which might be confusing for the tool developer .. at least it deserves proper documentation)

Example from mothur `seq.error`: 

```
                <param name="template" type="select" label="reference - Select an alignment database" help="">
                    <options from_data_table="mothur_aligndb">
                        <filter type="static_value" value="@FILTER@" column="2"/>
                    </options>
                </param>
```

In the commands block the access was:  `$template.fields.value`. Now we can add `<filter type="duplicate_column" column="0"/>` and access via `$template.fields[2]`.

alternatively
- one might implement this in `DynamicOptions` by adding a new attribute.
- tweak `tool_data_table.conf` to allow to define duplicate columns (btw. where is the doc for `tool_data_table.conf`)
- ... ?

Also adds the possibility to access `SelectToolParameterFieldWrapper` by index. Which was already documented, but did not work since attributes need to be strings.

### DIFF
--- a/lib/galaxy/tool_util/xsd/galaxy.xsd
+++ b/lib/galaxy/tool_util/xsd/galaxy.xsd
@@ -5882,6 +5882,7 @@ and ``bibtex`` are the only supported options.</xs:documentation>
       <xs:enumeration value="static_value"/>
       <xs:enumeration value="regexp"/>
       <xs:enumeration value="unique_value"/>
+      <xs:enumeration value="duplicate_column"/>
       <xs:enumeration value="multiple_splitter"/>
       <xs:enumeration value="add_value"/>
       <xs:enumeration value="remove_value"/>
--- a/lib/galaxy/tools/parameters/dynamic_options.py
+++ b/lib/galaxy/tools/parameters/dynamic_options.py
@@ -339,6 +339,25 @@ class MultipleSplitterFilter(Filter):
         return rval
 
 
+class DuplicateColumnFilter(Filter):
+    """
+    Required Attributes:
+        column: column in options to compare with
+    """
+
+    def __init__(self, d_option, elem):
+        Filter.__init__(self, d_option, elem)
+        column = elem.get("column", None)
+        assert column is not None, "Required 'column' attribute missing from filter"
+        self.column = d_option.column_spec_to_index(column)
+
+    def filter_options(self, options, trans, other_values):
+        rval = []
+        for fields in options:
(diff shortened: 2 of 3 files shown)


mvdbeek:
I'm not sure I understand the scenario where this is helpful and why we need or want to filter duplicate columns ?


> tools using such data tables can not be tested (to the best of my knowledge) since one needs to specify the value .. which is a path that changes with the test environment.

I guess I also don't understand this. Doesn't `{__here__}/some_value` work ?

bernt-matthias:
> I guess I also don't understand this. Doesn't `{__here__}/some_value` work ?

In the value of the test param? Lets try.

mvdbeek:
Ah, OK, I think I understand. In this case you should be able to test with the name column, and if that doesn't work I think that would be the right fix ?

bernt-matthias:
> Ah, OK, I think I understand. In this case you should be able to test with the name column, and if that doesn't work I think that would be the right fix ?

I will try. And I agree that the change seemed awkward from the beginning ..

bernt-matthias:
Both approaches seem not to work. Do you have a suggestion where in the code I could start digging when I want to influence how the value of a test parameter can be influenced.

Just to be sure that I did not made a type here are the params that I used:

```
<param name="type_cond|alignment|template" value="${__HERE__}/HMP_MOCK.v35.align"/>
```

and 

```
<param name="type_cond|alignment|template" value="unaligned-refdata"/>
```

mvdbeek:
I initially didn't understand what you were trying to do, so `<param name="type_cond|alignment|template" value="${__HERE__}/HMP_MOCK.v35.align"/>` is not going to work.

The second thing might work if you augment the legal values of a SelectToolParameter, so that https://github.com/galaxyproject/galaxy/blob/c606c99152d1f057a9b21d613385aadb689c5e94/lib/galaxy/tools/parameters/basic.py#L924 will fall back to the name.

PR #10880, opened 2020-12-09
Dynamic select parameters: also allow to choose legal options by name

bernt-matthias:
this change will allow to choose a legal option by name for dynamic selects (for static selects this is already the case).

use case: data table with 2 columns: name and value (e.g. mothur data tables and probably many other old ones).
tools using such data tables can not be tested since one needs to specify the value .. which is a path that changes with the test environment.

this has been tested for one mothur tool

TODO: 

- [x] should be documented somewhere. but where?

replaces https://github.com/galaxyproject/galaxy/pull/10874


mvdbeek:
I think this looks pretty good, can you add a test? I think it should now also work to have a static select list and select the case by the label, as in  `<option value="value">The value</option>`

bernt-matthias:
> I think this looks pretty good, can you add a test? 

I can try. You mean a functional tool test?

> I think it should now also work to have a static select list and select the case by the label, as in `<option value="value">The value</option>`

Think so.

mvdbeek:
Yep, I think you can just change an existing select test.

bernt-matthias:
It seems that this already worked for static select lists: https://github.com/galaxyproject/galaxy/blob/1a748565f299d1a65ee44005ce38eae296f1c126/test/functional/tools/multi_select.xml#L37

I don't know why ..

mvdbeek:
Maybe because it's the first option that gets selected by default anyway? Odd, but makes sense, I always assumed that worked and it's roundabout proof that it should work that way.

bernt-matthias:
> Maybe because it's the first option that gets selected by default anyway? Odd, but makes sense, I always assumed that worked and it's roundabout proof that it should work that way.

Also works for the 2nd option. Somehow I would like to know where this translation takes place.

bernt-matthias:
Found it https://github.com/galaxyproject/galaxy/blob/1a748565f299d1a65ee44005ce38eae296f1c126/lib/galaxy/tools/test.py#L164 seems to stem from here https://github.com/galaxyproject/galaxy/commit/5b6c092aa0a85455b4daacd318541e92bd08b876 

I will work on a test for dynamically defined options.

mvdbeek:
I think the documentation can go as an example into the schema, like we do for https://github.com/mvdbeek/galaxy/blob/41fd63cdc66d52510546a9d58661f7f3016fcf86/lib/galaxy/tool_util/xsd/galaxy.xsd#L1085 for instance.

bernt-matthias:
Wondering why in the other classes `get_legal_values` does not depend (i.e. simply calls) on `get_options` .. as in mad more clear in https://github.com/galaxyproject/galaxy/pull/10880/commits/908006127c34c9b02d743a854e6e5856228f59b2 (.. but lets see first if this works)

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -1354,6 +1354,38 @@ This tag set defines the tool's input parameters for executing the tool via the
...
+(`optional="true"`) the case that no option is selected can be specified with
+`value=""`.
+
+While in general it is preferred to specify the selected cases by their values it
+is also possible to specify them by their name (i.e. the content of the
+`option` tag that is shown to the user). One use case is a dynamic select that is
+generated from a data table with two columns: name and value where the value is
+a path. Since the path changes with the test environment it can not be used to
+select an option for a test.
+
```
mvdbeek, line 1388: Awesome!

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -1354,6 +1354,38 @@ This tag set defines the tool's input parameters for executing the tool via the
...
+#### ``boolean``
+
+The value of the test parameter should be set to `true` or `false` corresponding
+to the cases that the parameter is checked or not. Alternatively the value
+specified as `truevalue` or `falsevalue` can be given.
+
+#### ``data``
+
+Data input parameters can be given as a file name. The file should exist in the
+`test-data` folder. Multiple files can be specified as comma separated list.
```
mvdbeek, line 1372:
```suggestion
`test-data` folder. Multiple files can be specified as comma separated values.
```

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -1354,6 +1354,38 @@ This tag set defines the tool's input parameters for executing the tool via the
 functional test framework. See [test](#tool-tests-test) documentation for
 some simple examples of parameters.
 
+### Parameter Types
+
+#### ``text``, ``integer``, and ``float``
+
+Values for these parameters are simply given by the desired value.
```
mvdbeek, line 1361:
```suggestion
Values for these parameters are set using the ``value`` attribute.
```

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -1354,6 +1354,38 @@ This tag set defines the tool's input parameters for executing the tool via the
...
+
+#### ``boolean``
+
+The value of the test parameter should be set to `true` or `false` corresponding
```
mvdbeek, line 1365:
```suggestion
The ``value`` attribute of the test parameter should be set to ``true`` or ``false`` corresponding
```

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -1354,6 +1354,38 @@ This tag set defines the tool's input parameters for executing the tool via the
...
+
+#### ``data``
+
+Data input parameters can be given as a file name. The file should exist in the
+`test-data` folder. Multiple files can be specified as comma separated list.
+
+#### ``select``
+
+The value of a select parameter should be specified as the value of one of the
+legal options. If more than one option is selected (`multiple="true"`) they
```
mvdbeek, line 1377:
```suggestion
possible options. If more than one option is selected (`multiple="true"`) they
```

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -1354,6 +1354,38 @@ This tag set defines the tool's input parameters for executing the tool via the
...
+should be given as comma separated list. For optional selects
+(`optional="true"`) the case that no option is selected can be specified with
+`value=""`.
+
+While in general it is preferred to specify the selected cases by their values it
+is also possible to specify them by their name (i.e. the content of the
+`option` tag that is shown to the user). One use case is a dynamic select that is
+generated from a data table with two columns: name and value where the value is
+a path. Since the path changes with the test environment it can not be used to
```
mvdbeek, line 1386:
```suggestion
a path. Since the path changes with the test environment it cannot be used to
```

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -1354,6 +1354,38 @@ This tag set defines the tool's input parameters for executing the tool via the
...
+#### ``data``
+
+Data input parameters can be given as a file name. The file should exist in the
+`test-data` folder. Multiple files can be specified as comma separated list.
+
+#### ``select``
+
+The value of a select parameter should be specified as the value of one of the
+legal options. If more than one option is selected (`multiple="true"`) they
+should be given as comma separated list. For optional selects
```
mvdbeek, line 1378:
```suggestion
should be set as comma separated values. For optional select parameters
```

lib/galaxy/tools/parameters/basic.py
```diff
@@ -921,9 +925,14 @@ def from_json(self, value, trans, other_values={}, require_legal_value=True):
...
-                    raise ParameterValueError("an invalid option ({!r}) was selected (valid options: {})".format(v, ",".join(legal_values)), self.name, v, is_dynamic=self.is_dynamic)
-                rval.append(v)
+                if v in legal_values:
+                    rval.append(v)
+            if len(rval) == 0:
+                for v in value:
+                    if v in fallback_values:
+                        rval.append(fallback_values[v])
+            if len(rval) == 0:
+                raise ParameterValueError("an invalid option ({!r}) was selected (valid options: {})".format(v, ",".join(legal_values)), self.name, v, is_dynamic=self.is_dynamic)
```
mvdbeek:
```suggestion
                if v in legal_values:
                    rval.append(v)
                elif v in fallback_values:
                    rval.append(fallback_values[v])
                else:
                    raise ParameterValueError("an invalid option ({!r}) was selected (valid options: {})".format(v, ",".join(legal_values)), self.name, v, is_dynamic=self.is_dynamic)
```
All values need to be valid, while in your implementation just one value has to be valid.

bernt-matthias: The intention was that all values should be taken from the same column. 

Maybe we can change this to something along the following lines:

```
if set(value).issubset(set(legal_values)):
    return value
elif set(value).issubset(set(fallback_values.keys())):
    return fallback_values.values()
else:
    raise ParameterValueError("invalid options ({!r}) were selected (valid options: {})".format(",".join(set(value)-set(legal_values)), ",".join(legal_values)), self.name, v, is_dynamic=self.is_dynamic)
```

bernt-matthias: Seems that this is the last open point?

lib/galaxy/tools/parameters/basic.py
```diff
@@ -874,23 +874,27 @@ def get_options(self, trans, other_values):
...
     def from_json(self, value, trans, other_values={}, require_legal_value=True):
         try:
             legal_values = self.get_legal_values(trans, other_values)
         except ImplicitConversionRequired:
             return value
+        # if the given value is not found in the set of values of the legal
+        # options we fall back to check if the value is in the set of names of
+        # the legal options. this is done with the fallback_values dict which
+        # allows to determine the corresponding legal values
+        fallback_values = self.get_legal_names(trans, other_values)
```
jmchilton, line 902: Is this something we should add a configuration parameter for - like ``allow_set_by_name`` or something in the parameter definition in the XML?

bernt-matthias: I guess you want to ensure that this does not brake anything?

I would tend to: no. Because

- it already works for statically defined options. so this change just makes this consistent for dynamic options
- a configuration parameter would make this tricky to test with planemo .. there are quite a few tools for which this change allows to add a test for the data tables
- the aim of my implementation (maybe there is still a bug .. see https://github.com/galaxyproject/galaxy/pull/10880#discussion_r546592151) was that values from the name and value columns can't be mixed. I have a hard time finding an example where this could fail. 
- also this applies only to tool tests .. users can not choose from the name column anyway. 

- changing the xml definition might be an option .. but I still tend to no

lib/galaxy/tools/parameters/basic.py
mvdbeek: I have just translated the existing code to also take into account `fallback_values`. I don't follow the column argument. Your alternative suggestion fails for string values:
```
In [1]: set('abcedef')
Out[1]: {'a', 'b', 'c', 'd', 'e', 'f'}
```

Each value in `values` has to be valid, that is how the code was before, and that is not the case with the PR as is, where at least 1 value needs to be valid, so it would pass with invalid values.

bernt-matthias: I think `value` is always a list (or some other iterable) otherwise the for loop `for v in value:` in the original code would have failed, or? So with the small fix to use `.keys()` also for legal_values it should work:

```
if set(value).issubset(set(legal_values.keys())):
    return value
elif set(value).issubset(set(fallback_values.keys())):
    return fallback_values.values()
else:
    raise ParameterValueError("invalid options ({!r}) were selected (valid options: {})".format(",".join(set(value)-set(legal_values)), ",".join(legal_values)), self.name, v, is_dynamic=self.is_dynamic)
```

So this should test if all values are legal values or if all values are legal names. 

We just need to decide if this is the semantics that we want or your suggestion that all values need to be either a legal value or a legal name.

mvdbeek: You're right. I have no preference, but `legal_values` is a list, so should this be
```python
if set(value).issubset(set(legal_values)):
    return value
elif set(value).issubset(set(fallback_values.keys())):
    return fallback_values.values()
else:
    raise ParameterValueError("invalid options ({!r}) were selected (valid options: {})".format(",".join(set(value)-set(legal_values)), ",".join(legal_values)), self.name, v, is_dynamic=self.is_dynamic)
```
?

bernt-matthias: Done. Note that `get_legal_values` returns a `set`. So it should be fine as is.

lib/galaxy/tools/parameters/basic.py
jmchilton: Very good points - the first is particularly convincing.

jmchilton (approved):

test/functional/tools/select_from_dataset.xml
```diff
@@ -57,6 +57,26 @@ echo select_collection '$select_collection' >> '$output'
...
+            <param name="single" value="2.tabular" />
+            <param name="select_single" value="68" />
+            <param name="multiple" value="2.tabular" />
+            <param name="select_multiple" value="71,62" />
+            <param name="collection">
+                <collection type="list">
+                    <element name="2" value="2.tabular" />
+                </collection>
+            </param>
+            <param name="select_collection" value="75,58" />
```
mvdbeek, line 71: This test is failing now:
```
____________________ test_tool[select_from_dataset_test_2] _____________________

testcases = ('select_from_dataset_test_2', <functional.test_toolbox.TestForTool_select_from_dataset testMethod=runTest>, 1)
driver = <functional.test_toolbox_pytest.DefaultGalaxyTestDriver object at 0x7fb17b8cc290>

    @pytest.mark.parametrize("testcases", cases(), ids=idfn)
    def test_tool(testcases, driver):
        test = testcases[1]
>       test.do_it(test_index=testcases[2])

test/functional/test_toolbox_pytest.py:84: 
test/functional/test_toolbox.py:35: in do_it
    verify_tool(tool_id, self.galaxy_interactor, resource_parameters=resource_parameters, test_index=test_index, tool_version=tool_version, register_job_data=register_job_data)
lib/galaxy/tool_util/verify/interactor.py:1041: in verify_tool
    raise e
lib/galaxy/tool_util/verify/interactor.py:1034: in verify_tool
    tool_response = galaxy_interactor.run_tool(testdef, test_history, resource_parameters=resource_parameters)
lib/galaxy/tool_util/verify/interactor.py:456: in run_tool
    submit_response_object = ensure_tool_run_response_okay(submit_response, "execute tool", inputs_tree)

submit_response_object = <Response [400]>, request_desc = 'execute tool'
inputs = {'collection': {'id': '529fd61ab1c6cc36', 'src': 'hdca'}, 'multiple': {'id': '5a78730880510954', 'src': 'hda'}, 'select_collection': ['75', '58'], 'select_multiple': ['71', '62'], ...}

    def ensure_tool_run_response_okay(submit_response_object, request_desc, inputs=None):
        if submit_response_object.status_code != 200:
            message = None
            dynamic_param_error = False
            try:
                err_response = submit_response_object.json()
                if "param_errors" in err_response:
                    param_errors = err_response["param_errors"]
                    if "dbkey" in param_errors:
                        dbkey_err_obj = param_errors["dbkey"]
                        dbkey_val = dbkey_err_obj.get("parameter_value")
                        message = "Invalid dbkey specified [%s]" % dbkey_val
                    for value in param_errors.values():
                        if isinstance(value, dict) and value.get("is_dynamic"):
                            dynamic_param_error = True
                if message is None:
                    message = err_response.get("err_msg") or None
            except Exception:
                # invalid JSON content.
                pass
            if message is None:
                template = "Request to %s failed - invalid JSON content returned from Galaxy server [%s]"
                message = template % (request_desc, submit_response_object.text)
>           raise RunToolException(message, inputs, dynamic_param_error=dynamic_param_error)
E           galaxy.tool_util.verify.interactor.RunToolException: parameter 'select_single': an invalid option ('68') was selected (valid options: 4,1,5,7,8,10,9,2,6,3), parameter 'select_multiple': invalid options ('71,62') were selected (valid options: 4,1,5,7,8,10,9,2,6,3), parameter 'select_collection': invalid options ('58,75') were selected (valid options: 4,1,5,7,8,10,9,2,6,3)
```
Do you need to update `<column name="name" index="0"/>` and `<column name="value" index="0"/>` ?

mvdbeek (approved):

bernt-matthias:
To me the failing test (data_manager_installation_table_reload) looks unrelated - but not sure.

nsoranzo:
> To me the failing test (data_manager_installation_table_reload) looks unrelated - but not sure.

Restarted to double-check.

mvdbeek:
Definitely unrelated.

PR #10886, opened 2020-12-10
Allow relative paths when staging jobs (e.g. with planemo run)

simonbray:
So that a command like `planemo run wf.ga wf-job.yml` with:

```
Dataset 1:
  class: File
  path: "data/dataset1.txt"

```

will work correctly.

Currently, either the job file or the dataset paths need to be specified with an absolute path, otherwise `stage` isn't able to find `job_dir`.


mvdbeek (approved):
Awesome!

mvdbeek:
Thank you @simonbray!

PR #10904, opened 2020-12-11
Vue unit test cleanup, removing jquery, mocking ajax calls

Nerdinacan:
There's a lot of client-side unit tests that die or become erratic due to unnecessary 3rd party library usage or unmocked ajax calls. 

Jest doesn't run in a browser, it runs in node, and code that expects to talk to native browser objects (like document or XHR) either dies or runs without response. This is particularly problematic when jQuery is used. I'll be removing that first, then moving on to all the multitudes of unmocked ajax calls.

**Notable changes:**

1. Removed old babel rewire mocking mechanism in favor of jest mocking tools
2. Removed global Vue instances in unit tests. This is normally pretty innocuous, but as more people hang plugins and mixins off of the Vue instance, this can create problems that extend beyond the scope of a single test. Especially problematic are plugins or mixins that reference legacy code that either includes jQuery or fires instant ajax calls upon include, which includes most of the legacy backbone code, and some of the new Vuex initialization use-cases.
3. Removed usage of a jest utility mounter that referenced a non-existent document object.
4. Replaced mounts with shallowMounts in Vue unit tests when possible. In general it's a good idea to try a shallowMount first and then if you REALLY need to, go ahead and do a full mount. Also ask yourself if what you're writing is really a unit test if you require a full tree mount. Perhaps what you are trying to do would be better accomplished with an integration test.

### DIFF
--- a/client/.babelrc
+++ b/client/.babelrc
@@ -2,7 +2,7 @@
     "presets": ["@babel/preset-env"],
     "env": {
         "test": {
-            "plugins": ["@babel/plugin-transform-runtime", "rewire"]
+            "plugins": ["@babel/plugin-transform-runtime"]
         }
     }
 }
--- a/client/README.md
+++ b/client/README.md
@@ -115,8 +115,7 @@ framework.
 For testing Vue components, we use the [Vue testing
 utils](https://vue-test-utils.vuejs.org/) to mount individual components in a
 test bed and check them for rendered features.  Please use jest-based mocking
-for isolating test functionality.  Some tests exist using Rewire, but those
-should be updated over time.
+for isolating test functionality.
 
 A set of older qUnit tests also exist which will be phased-out as the code they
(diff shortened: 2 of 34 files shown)


bgruening:
Oh yeah, unleash the @Nerdinacan!

Nerdinacan:
Our client-side unit tests still need a lot of attention, but they'll probably have to be addressed one at a time. These global changes are ready now, however.

Nerdinacan:
@dannon
Oooh thanks for getting rid of that browser compatibility notice. The error message tells me what to do to fix it and yet I have stared at it for a year without updating it.

dannon:
@Nerdinacan Hah, yeah, same -- I just never got around to it and now I'm in "silence all the warnings" mode.

Nerdinacan:
I like what you've done with the place.

PR #10907, opened 2020-12-13
Remove trans from stream_dataset_collection()

mvdbeek:
and to_archive(). Also push stream_dataset_collection out of API.

### DIFF
--- a/lib/galaxy/datatypes/data.py
+++ b/lib/galaxy/datatypes/data.py
@@ -360,7 +360,7 @@ class Data(metaclass=DataMeta):
         trans.response.headers["Content-Disposition"] = 'attachment; filename="%s"' % filename
         return open(dataset.file_name, mode='rb')
 
-    def to_archive(self, trans, dataset, name=""):
+    def to_archive(self, dataset, name=""):
         """
         Collect archive paths and file handles that need to be exported when archiving `dataset`.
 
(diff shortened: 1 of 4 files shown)


lib/galaxy/datatypes/data.py
```diff
@@ -360,27 +360,25 @@ def _serve_raw(self, trans, dataset, to_ext, **kwd):
...
         :param name: archive name, in collection context corresponds to collection name(s) and element_identifier,
                      joined by '/', e.g 'fastq_collection/sample1/forward'
         """
-        composite_extensions = trans.app.datatypes_registry.get_composite_extensions()
-        composite_extensions.append('html')  # for archiving composite datatypes
         rel_paths = []
         file_paths = []
-        if dataset.extension in composite_extensions:
-            main_file = "{}.{}".format(name, 'html')
+        if dataset.datatype.composite_type or dataset.extension == 'html':
```
nsoranzo, line 373: This seems much better than using `get_composite_extensions()`, should we do that everywhere that method is used (I count 3 other places). Not necessarily as part of this PR clearly.

nsoranzo (approved):

lib/galaxy/datatypes/data.py
mvdbeek: Yep, will do!

jmchilton (approved):

PR #10908, opened 2020-12-13
More improvements to tool test scripting.

jmchilton:
- Replace --skip-successful with two new mutually exclusive options --skip-previously-successful and --skip-previously-executed.
- Enhance argparse usage for mutual exclusivity of history options also.
- I noticed in some tests of collection operation tools I will sometimes get the ToolInputNotReady problems - I refactored this to be a MessageException with a specific error code that can be checked and retried later by the client.
- Allow the client to specify test-data directories to check against - many of the bundled tools depend on https://github.com/galaxyproject/galaxy-test-data for instance but production instances will not have this available.
- Improve log message when writing outputs to standard output.
- Some older toolshed tools rely on Galaxy's bundled test data - so when generated paths to test-data check this also even if a repository directory is found.
- Refactor the entry point to allow passing in tool filters - I want to develop a downstream script that will test all the tools of a workflow.

### DIFF
--- a/lib/galaxy/tool_util/verify/interactor.py
+++ b/lib/galaxy/tool_util/verify/interactor.py
@@ -103,6 +103,9 @@ class GalaxyInteractorApi:
         self.keep_outputs_dir = kwds["keep_outputs_dir"]
         self.download_attempts = kwds.get("download_attempts", 1)
         self.download_sleep = kwds.get("download_sleep", 1)
+        # Local test data directories.
+        self.test_data_directories = kwds.get("test_data", [])
+
         self._target_galaxy_version = None
 
         self.uploads = {}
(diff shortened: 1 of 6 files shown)


lib/galaxy/tool_util/verify/interactor.py
```diff
@@ -420,7 +438,16 @@ def run_tool(self, testdef, history_id, resource_parameters={}):
             if isinstance(value, list) and len(value) == 1:
                 inputs_tree[key] = value[0]
 
-        submit_response = self.__submit_tool(history_id, tool_id=testdef.tool_id, tool_input=inputs_tree)
+        submit_response = None
+        for _ in range(30):
```
mvdbeek, line 442: That's kind of a low value on a real server. Maybe allow up to maxseconds ?

mvdbeek (approved):

lib/galaxy/tool_util/verify/interactor.py
jmchilton: I've committed that locally and will include it in my next batch of fixes 😆 - I'll merge this while it is green. Thanks for the review!

Issue #10794, opened 2020-11-24
Decorate tests for running against production servers


jmchilton:
Running Galaxies various API + Selenium tests against Galaxy instances would be greatly enhanced if they were decorated in a way to indicate what effect they have on servers (which tests produce new users, require admin keys, etc..) and made these decorations available to pytest for test selection.

Closing https://github.com/galaxyproject/galaxy/issues/3246 in lieu of this - I feel like there was another issue around this but I cannot find it (e.g. the tests producing too many histories or something).

PR #10911, opened 2020-12-14
Add screenshot decorators to selenium tests

OlegZharkov:
Adds `local_screenshotting` to selenium tests, so we can automate screnshot generation for [rule-based upload tutorial](https://training.galaxyproject.org/training-material/topics/galaxy-interface/tutorials/upload-rules/tutorial.html). Uploading to S3 will be done using https://github.com/galaxyproject/ci-testing. GTN will serve images directly directly from our S3 storage.

As we advance automatic screenshotting, we would need to run screenshot generation against external productive server. `local_screenshotting` here means, that these particular screenshots could be produced locally. 

compliments https://github.com/galaxyproject/galaxy/issues/10794


pytest.ini
```diff
@@ -6,3 +6,4 @@ python_paths = lib
 markers =
   data_manager: marks test as a data_manager test
   tool: marks test as a tool test
+  local_screenshotting: marks test as a screenshot producer
```
mvdbeek: How do you feel about `screenshot` ?
If you want to discrimate local and remote you should add a second `local` and/or `remote` marker.
These markers can be combined arbitrarily.

OlegZharkov: Ok, will do

lib/galaxy_test/selenium/test_uploads.py
```diff
@@ -173,6 +178,8 @@ def test_rules_example_2_list(self):
         # self.screenshot("rules_example_2_5_download_complete")
 
     @selenium_test
+    @pytest.mark.screenshot
```
jmchilton: A lot of the tests generate screenshots. Maybe call it training_screenshot or tutorial_screenshot or something?

mvdbeek: I'd probably add another marker for training/tutorial if it makes sense to subset them.

OlegZharkov: > I'd probably add another marker for training/tutorial if it makes sense to subset them.

I genuinely don't think it's something important, it's rather a minor detail. But I really need it for https://github.com/galaxyproject/ci-testing and if it helps to get it merged...

PR #10919, opened 2020-12-14
 Replace StreamBall, replace tarballs with zip

mvdbeek:
The way StreamBall writes to the wsgi response just doesn't work for ASGI frameworks.
The zipstream method seems superior in that it creates a uncompressed zip archive and doesn't need a tempfile.
Additionally I think it makes sense to return zip files, as that is probably the most well known compression format.
This also adds the option to have nginx serve zip archives directly using [mod_zip](https://www.nginx.com/resources/wiki/modules/zip/)


nsoranzo:
I think we could still take into account the value of `app.config.upstream_gzip` and use `ZIP_DEFLATED` instead of `ZIP_STORED` if it is `False` ?
Otherwise we should remove the option.

mvdbeek:
Alright, I've implemented returning a header that is compatible with mod-zip. admins (@natefoo @gmauro @hexylena), do you think https://github.com/galaxyproject/galaxy/blob/d122df8e9b9fb9b6c5a32c54d132487fa44cda30/doc/source/admin/nginx.md#creating-archives-with-mod-zip is reasonable ? An alternative we could implement (but that is more work, and I'm not sure time wouldn't be better spent completely offloading archive creation to celery or jobs) would be to return locations to <api/datasets/<dataset_id>/download/<filename>> which would then be served by x-send-file.

lib/galaxy_test/api/test_tools.py
```diff
@@ -370,9 +370,11 @@ def test_test_data_downloads_security(self):
...
         test_data_response = self._get("tools/%s/test_data_download?filename=velveth_test1" % "composite_output")
         assert test_data_response.status_code == 200
-        with tarfile.open(fileobj=BytesIO(test_data_response.content)) as tar_contents:
-            namelist = tar_contents.getnames()
-            assert len(namelist) == 5
+        with zipfile.ZipFile(BytesIO(test_data_response.content)) as contents:
+            namelist = contents.namelist()
+        assert len(namelist) == 6
+        expected_names = {'velveth_test1/Roadmaps', 'velveth_test1/output.html', 'velveth_test1/Sequences', 'velveth_test1/Log', 'velveth_test1/output/', 'velveth_test1/output/1'}
+        assert len(set(namelist).union(expected_names)) == 6
```
nsoranzo:
```suggestion
        assert set(namelist) == expected_names
```

doc/source/admin/nginx.md
```diff
@@ -353,6 +353,48 @@ galaxy:
 .. _protect-reports:
 ```
 
+### Creating archives with mod-zip
+
+Galaxy creates zip archives when downloading mulitple datasets from a history or a dataset library.
```
nsoranzo:
```suggestion
Galaxy creates zip archives when downloading multiple datasets from a history or a dataset library.
```

doc/source/admin/nginx.md
```diff
@@ -353,6 +353,48 @@ galaxy:
...
+While this works fine for small datasets and few users, nginx can handle the creation of zip archives
+more efficiently using [mod-zip](https://www.nginx.com/resources/wiki/modules/zip/).
+To use this feature install nginx with mod-zip enabled and provide the file locations from which
```
nsoranzo:
```suggestion
To use this feature, install nginx with mod-zip enabled, provide the file locations from which
```

doc/source/admin/nginx.md
```diff
@@ -353,6 +353,48 @@ galaxy:
...
+        location /galaxy_root/database/files/ {
+            internal;
+            alias /galaxy_root/database/files/;
+        }
+}
+```
+
+The `internal;` statement means that the location can only be used for internal nginx requests.
+For external requests, the client error 404 (Not Found) is returned, meaning users cannot
+access arbitrary datasets in /galaxy_root/database/files/.
```
nsoranzo:
```suggestion
access arbitrary datasets in `/galaxy_root/database/files/` .
```

natefoo:
FWIW, although the option specifically mentions `gzip`, the idea was that if your browser and proxy are (as all modern browsers would do) using compression with `Content-Encoding` then you don't want to double compress. For example, usegalaxy.org right now replies with `Content-Encoding: gzip`. But `mod-zip` is certainly a nicer option here.

mvdbeek:
The `upstream_gzip` option was only used together with StreamBall, afaict. I know this worked for tarballs, but if it also works for uncompressed zip archives I can restore the option and set `application/zip` in that case.

doc/source/admin/nginx.md
```diff
@@ -353,6 +353,48 @@ galaxy:
...
+
+        # handle archive create via mod-zip
+        location /galaxy_root/database/files/ {
+            internal;
+            alias /galaxy_root/database/files/;
+        }
+}
+```
+
+The `internal;` statement means that the location can only be used for internal nginx requests.
```
jmchilton, line 391: ❤️

mvdbeek:
Thanks @natefoo, adding `application/zip` to `gzip_types` actually nicely compresses the archive, so that's a good option to have when you can't use mod-zip. Added upstream_gzip back in https://github.com/galaxyproject/galaxy/pull/10919/commits/64b75e074c54556d39e8341b59218f5a55eed88e

jmchilton (approved):

PR #10945, opened 2020-12-17
Allow YAML list format for config options

### LINKED ISSUE #10942: Incorrect processing of config options set to a list of paths
Setting `tool_data_table_config_path` to a list of values causes an error like this one:
```
OSError: Error reading file '/home/sergey/0dev/galaxy/_galaxy/dev/config/['tool_data_table_conf.xml'': failed to load external entity "/home/sergey/0dev/galaxy/_galaxy/dev/config/['tool_data_table_conf.xml'"
```
The cause for the error is incorrect path processing in the config module. This may or may not affect other config options.

jdavcs:
Fixes #10942

This allows specifying multiple values in galaxy.yml as a YAML list in addition to comma-separated string:

```yaml
tool_data_table_config_path:
- path1
- path2

some_other_config: 'path1, path2'
```
EDIT: This is applicable to any options that allow multiple values (not just paths). This includes `admin_users`.

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -275,10 +275,11 @@ class BaseAppConfiguration:
             # check for `not None` explicitly (value can be falsy)
             if value is not None and datatype in type_converters:
                 # convert value or each item in value to type `datatype`
+                f = type_converters[datatype]
                 if isinstance(value, list):
-                    return [type_converters[datatype](item) for item in value]
+                    return [f(item) for item in value]
                 else:
-                    return type_converters[datatype](value)
+                    return f(value)
             return value
 
         def strip_deprecated_dir(key, value):
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -274,13 +274,17 @@ class BaseAppConfiguration:
             datatype = self.schema.app_schema[key].get('type')
             # check for `not None` explicitly (value can be falsy)
             if value is not None and datatype in type_converters:
-                return type_converters[datatype](value)
+                # convert value or each item in value to type `datatype`
+                if type(value) is list:
+                    return [type_converters[datatype](item) for item in value]
+                else:
+                    return type_converters[datatype](value)
             return value
 
         def strip_deprecated_dir(key, value):
             resolves_to = self.schema.paths_to_resolve.get(key)
             if resolves_to:  # value contains paths that will be resolved
-                paths = [path.strip() for path in value.split(',')]
+                paths = listify(value, do_strip=True)
                 for i, path in enumerate(paths):
                     first_dir = path.split(os.sep)[0]  # get first directory component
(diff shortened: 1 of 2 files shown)


lib/galaxy/config/__init__.py
```diff
@@ -274,13 +274,17 @@ def convert_datatype(key, value):
             datatype = self.schema.app_schema[key].get('type')
             # check for `not None` explicitly (value can be falsy)
             if value is not None and datatype in type_converters:
-                return type_converters[datatype](value)
+                # convert value or each item in value to type `datatype`
+                if type(value) is list:
```
```suggestion
                if isinstance(value, list):
```

lib/galaxy/config/__init__.py
```diff
@@ -274,13 +274,17 @@ def convert_datatype(key, value):
...
+                    return [type_converters[datatype](item) for item in value]
```
nsoranzo: Assigning `type_converters[datatype]` to a variable (before the `if`) would make this more readable, I think.

jdavcs: Agreed. Done.

test/unit/config/test_path_resolves_to.py
```diff
@@ -209,6 +209,15 @@ def test_kwargs_listify(mock_init, monkeypatch):
     assert config.path4 == ['my-config/new1', 'my-config/new2']
 
 
+def test_kwargs_as_list_listify(mock_init, monkeypatch):
+    # Expected: use values from kwargs; each value resolved and listified
+    new_path4 = ['new1', 'new2']
+    config = BaseAppConfiguration(path4=new_path4)
+
+    assert config._raw_config['path4'] == 'new1,new2'
```
nsoranzo: This surprised me a bit, I expected this to still be a list, but it is joined into a comma-separated string at https://github.com/ic4f/galaxy/blob/dev_yaml_list_format/lib/galaxy/config/__init__.py#L298 , but only for paths (`admin_users` doesn't have `resolves_to`).

Not sure if this can become a problem at some point, but I thought I should mention it.

nsoranzo (approved):

test/unit/config/test_path_resolves_to.py
jdavcs: I saw it too; and now that you mention this, I think we shouldn't leave it as is: the issue here is that now the `strip_deprecated_dir` function has this unexpected side effect (for list values).  Which, apart from being bad design, can later lead to problems, of course.
I'll fix it. I won't be ideal, but it'll preserve the `value` data type. 
In general, I think the config module is becoming too complicated. I think, if we have a way to mark all path values in the schema (making `path-resolves_to` mandatory for paths is one way), the logic can be simplified considerably. I should get to it soon.

jdavcs: @nsoranzo  Could you take a look at the latest commit?

nsoranzo (approved):

Issue #10947, opened 2020-12-17
Beta history panel issues


mvdbeek:
Super-exciting to have the history PR merged finally, that is a monumental piece of work, congrats @Nerdinacan ! And it looks really good too.
Nevertheless, here are some issues from my first impressions:

- [ ] Opening a tiny history with 2 items takes about 19 seconds with a call to the http://127.0.0.1:8081/api/histories?q=deleted&qv=None&view=betawebclient
  
![Screenshot 2020-12-17 at 10 39 07](https://user-images.githubusercontent.com/6804901/102470382-2eb55b80-4054-11eb-9e02-bdd41228e017.png)

- [x] The blue loading strips occasionally re-appear when hovering over the panel
- [ ] The Collapse Expanded Datasets button fails with 
```
vue.runtime.esm.js:1888 TypeError: e.emit is not a function
    at click (ContentOperations.vue?7cbb:1)
    at zt (vue.runtime.esm.js:1854)
    at n (vue.runtime.esm.js:2179)
    at zt (vue.runtime.esm.js:1854)
    at HTMLButtonElement.n (vue.runtime.esm.js:2175)
    at HTMLButtonElement.o._wrapper (vue.runtime.esm.js:6917)
```
- [ ] Select history dropdown is limited to the history panel div extent, it seems:
![Screenshot 2020-12-17 at 10 43 49](https://user-images.githubusercontent.com/6804901/102470934-dcc10580-4054-11eb-816e-fee3017acef9.png)
- [ ] The history dropdown is not scrollable / scrolls super weirdly, it sorts of stutters
  - [x] Is this why you're doing the call to http://127.0.0.1:8081/api/histories ? In that case, could you just fetch the history names ?
  - [ ] Selecting histories from there is probably not useful, I usually need to see contents, tags and modification times to find the history I want. That might just be my personal preference though
  - [ ] The unhide operation for collection elements is the first icon on the left, where we used to be able to look into the dataset. Maybe you can switch the view button back to that position and move the unhide into the `:` menu ?
 
![Screenshot 2020-12-17 at 10 48 43](https://user-images.githubusercontent.com/6804901/102471505-9029fa00-4055-11eb-920b-47320885fbd8.png)

- [ ] What does the the copy datasets button do / how do I use it ?
![Screenshot 2020-12-17 at 10 56 01](https://user-images.githubusercontent.com/6804901/102472367-881e8a00-4056-11eb-9841-58f1470185c0.png)

[...]

PR #10950, opened 2020-12-17
Replace vue tags component

Nerdinacan:
Replacing my original service dependency injection scheme with a provider-based option. Should be easier to maintain.

Also removing the 3rd party tags library in favor of either the bootstrap tags UI elements or just some native markup.

References usability concerns noted i:


Nerdinacan:
@marius These failed selenium tests are failing for me on dev as well. I looked at them and I don't see how they are related.

This fails on dev (401d061cd093c7af9fe6a1f1abe6e24c247d490d)
lib/galaxy_test/selenium/test_workflow_editor.py

This one passes locally both on this branch and on dev
lib/galaxy_test/selenium/test_history_panel.py

Both look functionally unrelated to the changes.

mvdbeek:
I don't see failing selenium tests here,  but if this isn't WIP anymore can you take it out of draft and rebase against current dev ?

PR #10951, opened 2020-12-17
Redirection after remote user logout

kjetilkl:
We use an external identity provider to authenticate "remote_users" in our Galaxy.  When a user selects "Logout" from the "User" menu, they are supposed to be redirected to the IdP's logout page to invalidate the session. This page was configured with the setting `remote_user_logout_href` in "galaxy.yml". 

It worked fine for many years, but after a recent Galaxy upgrade (don't remember exactly which version), the users were suddenly redirected to Galaxy's own login page instead (which is potentially very confusing).

The submitted hack fixed the problem for us, but I'm not sure if there is supposed to be a better solution.

### DIFF
--- a/client/src/layout/menu.js
+++ b/client/src/layout/menu.js
@@ -28,6 +28,8 @@ export function userLogout(logoutAll = false) {
         .then((response) => {
             if (response.data?.redirect_uri) {
                 window.top.location.href = response.data.redirect_uri;
+            } else if (galaxy.config.use_remote_user && galaxy.config.remote_user_logout_href) {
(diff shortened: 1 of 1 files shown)


dannon:
@kjetilkl Sorry for the delay here -- so it looks like in this situation the server isn't responding with the correct redirect_url, when it should be handling that end of it.  I'll take a look.

mvdbeek:
Looks like https://github.com/galaxyproject/galaxy/pull/11540 is superseeding this. Thanks @kjetilkl, and let us know if you hit any other problems.

PR #10957, opened 2020-12-17
Add tooltip to workflow node checkbox

### LINKED ISSUE #10603: Workflow output checkboes could need a popup on hover
grafik

it would be nice to have a popup with a small explanation what this checkbox does. I think we had those with the previous stars.

guerler:
Fixes #10603

image

### DIFF
--- a/client/src/components/Workflow/Editor/NodeOutput.vue
+++ b/client/src/components/Workflow/Editor/NodeOutput.vue
@@ -5,7 +5,7 @@
             :class="['callout-terminal', output.name]"
             @click="onToggle"
             v-b-tooltip
-            title="Unchecked outputs will be hidden."
+            title="Unchecked outputs will be hidden and are not available as subworkflow outputs."
         >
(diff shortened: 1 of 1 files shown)


client/src/components/Workflow/Editor/NodeOutput.vue
```diff
@@ -1,6 +1,12 @@
 <template>
     <div class="form-row dataRow output-data-row">
-        <div v-if="showCallout" :class="['callout-terminal', output.name]" @click="onToggle">
+        <div
+            v-if="showCallout"
+            :class="['callout-terminal', output.name]"
+            @click="onToggle"
+            v-b-tooltip
+            title="Unchecked outputs will be hidden."
```
mvdbeek:
```suggestion
            title="Unchecked outputs will be hidden and are not available as subworkflow outputs."
```
We really need to separate outputs from hidden. It's reasonable to have hidden workflow outputs.

mvdbeek (approved):

Issue #10937, opened 2020-12-16
Make the current Galaxy version visible to the user


afgane:
As a user it is not clear which (major) version of Galaxy is being used. usegalaxy.org for example, includes the version at the bottom of the custom `welcome.html` page but the framework should make this available by default. Perhaps a new Help -> About page that includes some general info about Galaxy as well as the current version, or even just Help -> Version page that displays the version and links to the Github branch/release.

nsoranzo:
Such page can just expose the information available via the API endpoint `/api/version`

hexylena:
duplicate of https://github.com/galaxyproject/galaxy/issues/3143 (I'll close that one since this has more detail.)

Agreed, it would be nice. Personally I think we need something like every other app's user menu, with your profile name / picture at top right, a drop down menu with information, and then the version displayed there in greyed out text. 

![image](https://user-images.githubusercontent.com/458683/102468201-869e9300-4051-11eb-9c69-62129163404c.png)

astrovsky01:
Any opinion on hybridizing the two? A tab on the help menu that lists version and redirects to the release notes?

afgane:
Starting with just a accessible version would be a great start, but linking it to the release notes would be even better from my perspective.

PR #10960, opened 2020-12-17
Add version tab which links to release notes to help tab

astrovsky01:
Based on https://github.com/galaxyproject/galaxy/issues/10937

### DIFF
--- a/client/src/layout/menu.js
+++ b/client/src/layout/menu.js
@@ -51,6 +51,12 @@ export function userLogoutClient() {
 
 export function fetchMenu(options = {}) {
     const Galaxy = getGalaxyInstance();
+    const { version_minor, version_major } = Galaxy.config;
+    const { release_doc_base_url } = options;
+    const versionUserDocumentationUrl = (version_minor == "dev") 
+        ? "https://docs.galaxyproject.org/en/latest/releases/index.html"
+        : `${release_doc_base_url}${version_major}/releases/${version_major}_announce_user.html`
+
     const menu = [];
     //
     // Analyze data tab.
@@ -193,7 +199,7 @@ export function fetchMenu(options = {}) {
             },
             {
                 title: _l("Galaxy Version: " + Galaxy.config.version_major),
-                url: options.release_doc_base_url  + Galaxy.config.version_major + "_announce_user.html",
+                url:  versionUserDocumentationUrl,
                 target: "_blank",
             },
         ],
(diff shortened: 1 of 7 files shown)


hexylena:
This looks good!

dannon:
So it looks like admins would need to update the 'release url' every time they update, right?  Maybe instead of making this configurable, it's easier to just compose the correct version URL based on the internal version reckoning?

hexylena:
> So it looks like admins would need to update the 'release url' every time they update, right?

I read it differently, that was just the base url for the documentation. It looked like the link was constructed on the fly to include _announce_user.html and the version string?

client/src/layout/menu.js
```diff
@@ -192,6 +191,11 @@ export function fetchMenu(options = {}) {
                 title: _l("Interactive Tours"),
                 url: "tours",
             },
+            {
+                title: _l("Galaxy Version: " + Galaxy.config.version_major),
+                url: options.release_url + Galaxy.config.version_major + "_announce.html",
```
hexylena:
```suggestion
                url: options.release_url + Galaxy.config.version_major + "_announce_user.html",
```

maybe should be the user-facing release notes, more interesting for users.

dannon:
@hexylena Yeah, it's composed -- I read through that too fast.  The naming of the option is a bit misleading, I'd have called it release_doc_base_url or something.

hexylena:
Agreed! base would be better.

client/src/layout/menu.js
astrovsky01: Good call, thanks!

astrovsky01:
So the one thing I noticed when updating the name and just double checking everything was that, though we have a dev announcement page for 21.01, we don't for the user announcements. In the case where someone is running off of dev (for whatever reason), that link doesn't work, and won't until the official release. 
I guess the question is, should we create the same page stand-in for the user next release, or is there a better idea?

dannon:
@astrovsky01 I'd actually go the other way with that and link just to the root release_version page.  And we should put at that root what most people expect to see.

So, instead of:

`https://docs.galaxyproject.org/en/master/releases/20.09_announce.html`

Simply: 
https://docs.galaxyproject.org/en/release_20.09/

(note `master` is going away anyway, to be `main` soon and this will need to be updated)

afgane (approved):
Tested this locally - lgtm.

PR #10962, opened 2020-12-17
Allow submitting tools while beta history is enabled.

jmchilton:


Nerdinacan:
I was discussing this issue with Assunta. Since Galaxy.currentHistoryPanel isn't there when the beta history is open, we could do something like subscribe to the store and provide a skeleton Galaxy.currentHistoryPanel for some of these features that just need the current id.

Like so:

```javascript

// client/src/store/index.js

export function createStore() {


    // ... configures the store


    const store = new Vuex.Store(storeConfig);

    // https://vuex.vuejs.org/api/#subscribe

    store.subscribe((mutation, state) => {
        if (mutation.type == "somethingThatSetsCurrentHistoryId") {
            const fakePanel = {
                whatever: {
                    nestedNonsense: {
                        backboneRequired: {
                            id: state.betaHistory.currentHistoryId
                        }
                    }
                }
            }


            getGalaxyInstance().currHistoryPanel = fakePanel;
        }
    })

    return store;
}

const store = createStore();

export default store;
```

What do you tihnk?

PR #10967, opened 2020-12-18
[20.09] Fix the collection state progress bar

mvdbeek:
Noone missed this enough to open an issue? 😆 
Screenshot 2020-12-18 at 16 32 49


wm75:
Did miss it, but had other things to worry about ;)

jmchilton:
Thanks for the fix!

PR #10969, opened 2020-12-18
Some more small fixes to tool test script.

jmchilton:


lib/galaxy/tool_util/verify/interactor.py
```diff
@@ -326,9 +326,9 @@ def test_data_download(self, tool_id, filename, mode='file', is_output=True):
                 if os.path.exists(local_path):
                     break
 
-        if result is None and os.path.exists(local_path):
+        if result is None and local_path is not None and os.path.exists(local_path):
             if mode == 'file':
-                result = open(local_path, mode='rb')
+                result = open(local_path, mode='rb').read()
```
nsoranzo:
```suggestion
                with open(local_path, mode='rb') as fh:
                    result = fh.read()
```

jmchilton: Opps - yeah! Thanks.

nsoranzo (approved):

Issue #10966, opened 2020-12-18
Workflow invocation problems with large collections


wm75:
Since over a week I'm struggling to get one of my WFs running on .eu. Various versions of the WF (https://usegalaxy.eu/workflow/export_to_file?id=adb3676c3450a227) give invocation failures with tracebacks like https://gist.github.com/bgruening/fc6767ddaf93fd4d9db0d00c45315b19

The exact evaluation problem (and the step at which invocation fails) usually changes when I modify the WF, but so far I haven't found a version that could be invoked successfully on .eu.

The issue only occurs if the WF is run with large enough collections (over 100 samples) and I'm not even sure whether it's a Galaxy core problem or a scheduling configuration issue on .eu, but if somebody could take a look it would be highly  appreciated and I'd be willing to help debugging.

mvdbeek:
Hmm, maybe this is `maximum_workflow_jobs_per_scheduling_iteration`, which defaults to 1000 since 20.09. Any chance you could bump that upat .eu  temporarily ?

wm75:
My vague impression from observing the scheduling dynamics on .eu is actually that scheduling of collections interferes somehow with scheduling of regular jobs with simple dataset inputs (the evaluation problem in the traceback also seems to refer to the output of a step that's not dealing with collections).

Going to rerun with increased `maximum_workflow_jobs_per_scheduling_iteration` now though.

wm75:
So increasing `maximum_workflow_jobs_per_scheduling_iteration` to 2000 gave successful scheduling of the WF.
I'm a bit sceptical still whether that fixed the underlying issue though:
My suspicion is that the **root** cause of the whole issue is that this WF is branched and that the scheduling of simple jobs in one branch interferes with simultaneous scheduling of collections in the other branch. Now what happened with the new setting was that somehow the two branches got scheduled sequentially, i.e. the steps from one branch only got scheduled after all steps from the other branch had been scheduled successfully.

mvdbeek:
Yeah, that makes some sense to me, and that gives me a good angle to look into what's going on there. I don't quite see how the branched progress could influence another branch, but that would explain why I haven't been able to reproduce other reports.

Issue #10738, opened 2020-11-17
[20.09] Filter empty fails to schedule in subworkflow for large collections


innovate-invent:
This is the excerpt of the subworkflow I am running:
![image](https://user-images.githubusercontent.com/1768820/99432780-7b750d80-28c1-11eb-81ea-f904b19879ff.png)

```
GET /api/invocations/{id}
8    
model_class    "WorkflowInvocationStep"
id    "9cf1173826c73159"
update_time    "2020-11-17T11:25:03.658747"
job_id    null
workflow_step_id    "fbed7d1754ce1275"
subworkflow_invocation_id    null
state    "scheduled"
action    null
order_index    8
workflow_step_label    "BLAST_Hit_Filter"
workflow_step_uuid    "67da9292-6f26-445e-b257-9647441df26e"
9    
model_class    "WorkflowInvocationStep"
id    "545812229749f3fe"
update_time    "2020-11-17T11:25:03.663948"
job_id    null
workflow_step_id    "7d14e995f1bd8ec9"
subworkflow_invocation_id    null
state    "new"
action    null
order_index    9
workflow_step_label    null
workflow_step_uuid    "059f43e7-4b59-4adb-95be-e07f8a189e61"
```

```
GET /api/invocations/{id}/steps/545812229749f3fe
model_class | "WorkflowInvocationStep"
id | "545812229749f3fe"
update_time | "2020-11-17T11:25:03.663948"
job_id | null
workflow_step_id | "7d14e995f1bd8ec9"
subworkflow_invocation_id | null
state | "new"
action | null
order_index | 9
workflow_step_label | null
workflow_step_uuid | "059f43e7-4b59-4adb-95be-e07f8a189e61"
outputs | {}
output_collections | {}
jobs | []
```

The output of the BLAST_Hit_Filter step reports success for all jobs. Manually running the Filter Empty tool on the output succeeds, removing 4 datasets. This tool fails to schedule in the workflow for a collection size of 1000 but succeeds for a collection size of 12. No error output is generated in the logs.

Possibly related to https://github.com/galaxyproject/galaxy/issues/8390

innovate-invent:
@mvdbeek setting maximum_workflow_jobs_per_scheduling_iteration = 10 does not cause a collection of 15 datasets to fail

innovate-invent:
I ran a smaller dataset (~700) and found this exception at a different stage of the workflow.
```
[...]

PR #10970, opened 2020-12-18
[20.09] Record invocation_step outputs as appropriate

mvdbeek:
This fixes https://github.com/galaxyproject/galaxy/issues/10966 and probably fixes https://github.com/galaxyproject/galaxy/issues/10738.

The issue occurs when a collection is being mapped over with more than `maximum_workflow_jobs_per_scheduling_iteration` items, and the `WorkflowProgress` instance tries to recover outputs via `recover_mapping`, which relies on the invocation steps' `output_datasets` and `output_dataset_collections` to be persisted.


jmchilton:
Thanks for the fix!

PR #10971, opened 2020-12-19
[21.01] Beta History initial load optimization (a.k.a. Added many data providers)

Nerdinacan:
I retooled the beta history store to do the following:

1. The beta history store no longer requires an $init method. Having that $init function run on instantiation seemed like a good idea at the time, but having global code that runs that doesn't necessarily have anything to do with the currently viewed components..... that's just poor design. Auto, magically instantiated code is harder to debug, and sometimes hard to know how to mock, so I'm removing that concept. I'll also go back and remove it from the other store modules at a later time.

2.  Developed a UserHistories provider object. Most of the internal behavior of this provider simply passes through to Vuex, but it also gives us lifecycle hooks to initialize (see #1) and leaves us room to subscribe to other storage and/or data retrieval layers in the future (caching, or websockets.)

3.  I made the initial history load query faster by requesting fewer fields. This was the original slow-down issue that prompted the PR in the first place. Now, a quick listing query is performed and a deeper history load is performed when the user switches histories.

4. Created providers for item expansion and item selection within the content list. These features were previously baked-in to the History component.. These are simple list management providers, but in the case of the item selection provider, it paves the way for the more elaborate dynamic selection scheme we plan for the future whereby the user might select a range of server-side items based on a query rather than simple item toggling.

5. Worked to further separate the render-based content-item components from their data management features by implementing more straight-forward event emissions. I would like to get this to the point where the dataset UI is completely pluggable. Likewise the most complicated parts of the rxjs have been separated so they can be unit-tested extensively.

6. Corrected a bug in polling which caused it to continue polling after the user had drilled down into a child collection.

7. Completely separated out Rxjs provider code into a separate file so it can be unit-tested independent of Vue.

Addresses:

Implements:
https://github.com/galaxyproject/galaxy/issues/11077


bgruening:
This one works very well in regards to scrolling!!!

Nerdinacan:
Not really sure why that ftp selenium test is failing, I'm having difficulty duplicating the failure locally. Anybody have any ideas?

Nerdinacan:
Ok I'm done! There's more to do but Dannon has convinced me to stop building.

client/src/components/History/ContentItem/ContentItem.js
```diff
@@ -11,22 +11,18 @@
 import Placeholder from "./Placeholder";
 import Dataset from "./Dataset";
 import DatasetCollection from "./DatasetCollection";
+// import SubDataset from "./SubDataset";
```
mvdbeek, line 14:
```suggestion
```

client/src/components/History/ContentItem/ContentItem.js
```diff
@@ -66,24 +63,12 @@ export default {
...
-            get() {
-                return this.isExpanded(this.item);
-            },
-            set(val) {
-                this.setExpanded(this.item, val);
-            },
-        },
         scrolling() {
-            return this.listState.scrolling;
+            return false;
```
mvdbeek: What's the idea here ?

client/src/components/History/ContentItem/DatasetCollection/DscUI.vue
```diff
@@ -6,12 +6,9 @@
...
         :class="{ selected }"
         :data-state="dsc.state"
-        @keydown.arrow-right.self.stop="$emit('select-collection', dsc)"
+        @keydown.arrow-right.self.stop="$emit('selectCollection')"
         @keydown.space.self.stop.prevent="$emit('update:selected', !selected)"
-        @click.stop="
-            $emit('select-collection', dsc);
-            $emit('update:expanded', dsc);
-        "
+        @click.stop="$emit('selectCollection')"
```
mvdbeek: How do we distinguish between expanded and selected ? I'd have assumed `'expandCollection'` would be the event to emit?

mvdbeek: (And whatever we do, we need to sync that with https://github.com/Nerdinacan/galaxy/blob/1ebdb8fb00f47e7a754322b315558614d67b4c49/client/src/components/WorkflowInvocationState/DatasetCollectionUIWrapper.vue#L9)

client/src/components/History/ContentOperations.vue
```diff
@@ -191,6 +193,10 @@ export default {
         },
     },
     methods: {
+        log(...args) {
+            console.log(...args);
+        },
+
```
mvdbeek: I guess that's debugging leftover ?

client/src/components/History/History.vue
```diff
@@ -3,90 +3,113 @@
...
-import { HistoryContentProvider } from "./providers";
+import { HistoryContentProvider, ExpandedItems, SelectedItems } from "./providers";
 import Layout from "./Layout";
 import HistoryMessages from "./HistoryMessages";
 import HistoryDetails from "./HistoryDetails";
 import HistoryEmpty from "./HistoryEmpty";
 import ContentOperations from "./ContentOperations";
 import ToolHelpModal from "./ToolHelpModal";
-import ListMixin from "./ListMixin";
+// import ListMixin from "./ListMixin";
```
mvdbeek:
```suggestion
```

client/src/components/History/History.vue
```diff
@@ -3,90 +3,113 @@
...
 import VirtualScroller from "../VirtualScroller";
 import { HistoryContentItem } from "./ContentItem";
 
 export default {
-    mixins: [ListMixin],
+    // mixins: [ListMixin],
```
mvdbeek:
```suggestion
```

client/src/components/History/HistoryDetails.vue
```diff
@@ -1,129 +1,95 @@
...
                     icon="fas fa-stream"
                     @click="backboneRoute('/histories/citations', { id: history.id })"
                 />
-
                 <PriorityMenuItem
                     key="export-history-to-file"
                     title="Export History to File"
                     icon="fas fa-file-archive"
-                    @click="backboneRoute(`/histories/${history.id}/export`)"
+                    @click="iframeRedirect('/history/export_archive?preview=True')"
```
mvdbeek, line 87:
```suggestion
                    @click="@click="backboneRoute(`/histories/${history.id}/export`)"
```
Maybe a bad merge here ? The `preview` keyword doesn't exist anymore

client/src/components/History/adapters/HistoryPanelProxy.js
```diff
@@ -65,7 +65,7 @@ export const HistoryPanelProxy = Backbone.View.extend({
         // Watch the store, change the fake history model when it changs
         store.subscribe(({ type, payload: newId }) => {
             if (type == "betaHistory/setCurrentHistoryId") {
-                console.log("setCurrentHistoryId", newId);
+                // console.log("setCurrentHistoryId", newId);
```
mvdbeek:
```suggestion
```

client/src/components/History/providers/HistoryContentProvider/watchHistoryContents.js
```diff
@@ -1,11 +1,14 @@
 import { of, combineLatest } from "rxjs";
-import { map, switchMap, debounceTime, scan, catchError, distinctUntilChanged } from "rxjs/operators";
-import { tag } from "rxjs-spy/operators/tag";
+import { map, switchMap, debounceTime, scan, distinctUntilChanged } from "rxjs/operators";
+// import { tag } from "rxjs-spy/operators/tag";
```
mvdbeek:
```suggestion
```

client/src/components/WorkflowInvocationState/WorkflowInvocationDetails.vue
```diff
@@ -56,15 +56,14 @@
 import ParameterStep from "./ParameterStep.vue";
 import WorkflowInvocationDataContents from "./WorkflowInvocationDataContents";
 import WorkflowInvocationStep from "./WorkflowInvocationStep";
-import ListMixin from "components/History/ListMixin";
+// import ListMixin from "components/History/ListMixin";
```
mvdbeek:
```suggestion
```

client/src/components/WorkflowInvocationState/WorkflowInvocationDetails.vue
```diff
@@ -56,15 +56,14 @@
...
 import { monitorHistoryUntilTrue } from "./providers/monitors";
 
 import { mapGetters } from "vuex";
 import { mapCacheActions } from "vuex-cache";
-import { vueRxShortcuts } from "components/plugins";
 
 export default {
-    mixins: [ListMixin, vueRxShortcuts],
+    // mixins: [ListMixin],
```
mvdbeek:
```suggestion
```

client/src/components/Upload/UploadModal.vue
```diff
@@ -259,89 +82,9 @@ const UploadModal = {
...
-                        break;
-                    }
-                }
-                data.payload.inputs = JSON.stringify(inputs);
-            }
-            return data;
-        },
+    },
+    created() {
+        this.eventHub.$on("upload:show", this.show);
```
mvdbeek: We do create multiple upload modals with different props and callbacks, one is in [DataDialog.vue](https://github.com/Nerdinacan/galaxy/blob/1ebdb8fb00f47e7a754322b315558614d67b4c49/client/src/components/DataDialog/DataDialog.vue#L138), and they all get triggered with `upload:show`. If you want to try this, open a tool, click on the button next to a data input, and at the bottom you have the option to upload a dataset.
This doesn't currently show anything since https://github.com/Nerdinacan/galaxy/blob/1ebdb8fb00f47e7a754322b315558614d67b4c49/client/src/components/DataDialog/DataDialog.vue#L136 passes in `modalShow: true` which is not a prop anymore.

I don't know what the solution is there, a single global upload button may work, but then we need to manage props. Maybe custom upload modals are a better solution ?

mvdbeek (commented):
The initial load looks better now. Happy to merge if we can somehow deal with `modalShow` not being a prop anymore.

client/src/components/History/ContentItem/ContentItem.js
Nerdinacan: Ahhh I'm reworking that in another branch. At one point I was showing all "placeholders" while the user was dragging very long lists.

client/src/components/History/ContentOperations.vue
Nerdinacan: yes

client/src/components/History/ContentItem/ContentItem.js
Nerdinacan: I was going to add those changes in here, but I'm trying to reduce the size of these PRs. I have another PR out that is clarifying the content-item hierarchy. I'll remove it for now.

client/src/components/History/ContentItem/DatasetCollection/DscUI.vue
Nerdinacan: What I'm trying to do is separate the UI elements from the data itself so that the behavior of the UI render-only elements can be more-easily re-used for different scenarios.

In the situation you have highlighted I am simply emitting that this collection is selected and when the event is handled at a data-management layer, "selectCollection" is the event that is used to select a sub-collection to drill down into. That event is handled in History.vue.

https://github.com/Nerdinacan/galaxy/blob/1ebdb8fb00f47e7a754322b315558614d67b4c49/client/src/components/History/History.vue#L83

Nerdinacan: Selecting a collection here is different from making a "content selection" for future operations.

The format of the event 'update:variable' was picked specifically to be compatible with standard Vue .sync syntax, i.e.

```html
<Component :var.sync="localVariable" />
```

Is the same as

```html
<Component :var="localVariable" @update:var="localVariable" />
```

But to answer your question, we are distinguishing between these three operations by handling their events in History.vue appropriately. Expansions are handled with a method exposed by the ExpandedItems provider, and selections will be handled by the SelectedItems Provider.

mvdbeek: Sure, that makes sense, but the way this is used in the history you're expanding this collection, which seems distinct from selecting a collection. How would you handle actual selection then (with checkboxes or filter or ...) ... shouldn't that emit `selectCollection` ? I don't mind either way, but this seems a little confusing to me, unless the idea is that you distinguish what the emitted event means in the parent component.

Nerdinacan:
```
<!-- from History.vue --->
<HistoryContentItem
        :item="item"
        :index="index"
        :show-selection="showSelection"
        
        <!-- here the value is being defined and its change is being trapped by the expanded provider -->
        :expanded="isExpanded(item)"
        @update:expanded="setExpanded(item, $event)"
    
        <!-- here is where selection is managed, I wanted this to happen in a separate data provider 
        because we are about to expand on this concept to make it fancier, enabling more than just 
        toggle-selections -->
        :selected="isSelected(item)"
        @update:selected="setSelected(item, $event)"

        <!-- this event drills the history panel down into a sub-collection -->
        @selectCollection="$emit('selectCollection', item)"
    />
```

The advantage of this setup is that most of the behavior of the history is handled in the History.vue comonent and the rest of the components are largely rendering.

Nerdinacan: > (And whatever we do, we need to sync that with https://github.com/Nerdinacan/galaxy/blob/1ebdb8fb00f47e7a754322b315558614d67b4c49/client/src/components/WorkflowInvocationState/DatasetCollectionUIWrapper.vue#L9)

Actually you have highlighted the point of what I'm doing. By Making the event data-agnostic, you are free to manage those events how you please in your own component, as you actually have done. You are able to define your own application-appropriate wrapper for the renderer with it's own handler "toggleExpand" in your case.

client/src/components/History/providers/HistoryContentProvider/watchHistoryContents.js
Nerdinacan: You keep removing useful debugging features we might want to just uncomment later, but ok. Haha.

client/src/components/Upload/UploadModal.vue
Nerdinacan: Hmmm let me look carefully at that. There was only one "instance" of an upload modal as I understood it. It was simply being shown and hidden in a byzantine and unexplained train of props. But, you're telling me the actual goal was to have multiple upload modals open at once?

client/src/components/History/ContentItem/DatasetCollection/DscUI.vue
Nerdinacan: > Sure, that makes sense, but the way this is used in the history you're expanding this collection, which seems distinct from selecting a collection. How would you handle actual selection then (with checkboxes or filter or ...) ... shouldn't that emit `selectCollection` ? I don't mind either way, but this seems a little confusing to me, unless the idea is that you distinguish what the emitted event means in the parent component.

I think I'm having trouble interpreting your concern here. Is your concern "how do we select a collection for use in bulk operations or something, if clicking on it just drills down?" I think that's what you meant. Is it just a question of preferred UX?

Nerdinacan: Yeah I just checked, i don't have any problems selecting a collection either via the space bar, or by clicking the explicit checkbox button in the toolbar. I believe this is similar to the current behavior. Perhaps I'm missing your point.

client/src/components/Upload/UploadModal.vue
mvdbeek: goal is a big word, but the callback and the `modalShow` prop made it possible to upload datasets straight from an open tool form or workflow run form and have the form update its selected dataset. If we can adapt the current singleton or a variant of the upload modal for those purposes that might be fine too.

client/src/components/History/ContentItem/DatasetCollection/DscUI.vue
mvdbeek: I think github has some delays in there today. I'm just worried about the word `selectCollection`, since the way DscUI used in the history it seems like expanding is more appropriate ? But if you want something generic (select) and the parent component decides what this should actually do (expand or focus) then this is fine.

client/src/components/Upload/UploadModal.vue
Nerdinacan: Got it, I'll try to create a good test scenario and a fix while still letting us initialize the Upload in a sane way.

client/src/components/History/ContentItem/DatasetCollection/DscUI.vue
Nerdinacan: Ahhh terminology! I see.

Yes:
1. Expanding opens a dataset to look at its details
2. Selecting adds content to the selection for a later operation
3. "SelectCollection" drills down into a collection

Yes I see your point that "SelectCollection" more closely resembles "expanding" than "selecting". I'll make some changes to make it more coherent. Good catch!

Nerdinacan: I'm just going to call it viewCollection, so it's clearly a 3rd operation.
I'm renaming the panel that holds the drill-down collection contents to "CurrentCollection" as well, and the list of collection breadcrumbs to simply "breadcrumbs"

client/src/components/Upload/UploadModal.vue
Nerdinacan: Alrighty, I found the PR that John originally opened for that functionality.
https://github.com/galaxyproject/galaxy/pull/9809

He's got a nice little video of his goal which I think is an awesome way to describe this stuff. But the way to implement this kind of behavior is with a promise. You don't want to assume everything you need is necessarily there. That's why feature is broken in dev, and that's why the selenium tests are failing (because the inits are not completed yet.)

Instead of this:

```js
 /** Called when user decides to upload new data */
  onUpload() {
      const propsData = {
          multiple: this.multiple,
          format: this.format,
          callback: this.callback,
          modalShow: true,
      };
      mountUploadModal(propsData);
      this.modalShow = false;
  },
```

It should really be something more like:
```js
onUpload() {
  const propsData = {
      multiple: this.multiple,
      format: this.format,
      modalShow: true,
  };
  mountUploadModal(propsData).then(() => {
      this.callback();
  })
}
```

Nerdinacan: I'll rewrite it so that the mount function can return a promise when needed.

Nerdinacan: Avoiding the Function prop value was turning into a needless rabbit-hole. I have a complete jquery-free, select2-free, Galaxy global var-free replacement for Upload almost ready, so I just restored what was there for now.

client/src/components/Upload/mount.js
```diff
@@ -1,11 +1,29 @@
...
+// Upload dialog instance
+let uploadVm = null;
+
+export function mountUploadModal(options = {}) {
     const propsData = initializeUploadDefaults(options);
-    const mounter = mountVueComponent(UploadModal);
-    const container = document.createElement("div");
-    document.body.appendChild(container);
-    mounter(propsData, container);
+
```
mvdbeek, line 11: Don't you need to return a promise in this function for 
```
            mountUploadModal(options).then((result) => {
                if (callback) callback(result);
            });
```
to work ?

Nerdinacan: I was going to do that, but it was turning into a mess. I'm actually almost done rewriting the whole Upload component so I just made the callback work again for now.

mvdbeek: But then you need to drop the promise use in client/src/utils/data.js. Also the select button doesn't appear in the upload modal currently.

Nerdinacan: But to answer your question, yes, you would just make a mount or open function that returns a promise that resolves when an event triggers. The strategy is straightforward and looked something like this:

```js
import { eventHub } from "components/plugins/eventHub";

export function openUploadModal(options) {
    mountUploadModal(options);

    return new Promise((resolve, reject) => {

        eventHub.$once('upload:cancel', (payload) => {
            reject(payload);
        })

        eventHub.$once('upload:close', (payload) => {
            reject(payload);
        })

        eventHub.$once('upload:ok', (payload) => {
            resolve(payload);
        })
    })
}
```

Nerdinacan: But then I have to wade through the component's butchered event handling which led to a testing mess.

Nerdinacan: Eeek, right you are. Forgot to remove all traces of that plan! Sorry it was getting late when I gave up on this thing.

mvdbeek (approved):

dannon:
We're going to rebase/retarget 21.01 and then get this in, it's in good shape now and fixes numerous issues with the new history.

dannon (approved):

PR #10975, opened 2020-12-20
Fix workflow parameter text value restriction when connected to select and text parameter

### LINKED ISSUE #10974: Workflow text parameter with value restriction restricted with free-style text input
A text parameter can be connected to a select parameter and a text parameter, but if you select "Provide list of possible values" the text field overrides all possible values, so you can't select any parameter, and the text field that should be present as a fallback is also not available.

mvdbeek:
Text parameters are freestyle parameters, so shouldn't influence the
restricted values, and text parameters also don't have a get_options
attribute. Fixes #10974.

To be able to test this I added an option allow sending history_id to /api/workflows/{encoded_workflow_id}/download
instead of relying on trans.history for the right history.

### DIFF
--- a/lib/galaxy/workflow/modules.py
+++ b/lib/galaxy/workflow/modules.py
@@ -980,7 +980,7 @@ class InputParameterModule(WorkflowModule):
                     tool_inputs = module.tool.inputs  # may not be set, but we're catching the Exception below.
 
                     def callback(input, prefixed_name, context, **kwargs):
-                        if prefixed_name == connection.input_name:
+                        if prefixed_name == connection.input_name and hasattr(input, 'get_options'):
                             static_options.append(input.get_options(self.trans, {}))
                     visit_input_values(tool_inputs, module.state.inputs, callback)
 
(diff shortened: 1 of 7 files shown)


lib/galaxy/workflow/modules.py
```diff
@@ -980,7 +980,7 @@ def get_runtime_inputs(self, connections=None, **kwds):
                     tool_inputs = module.tool.inputs  # may not be set, but we're catching the Exception below.
 
                     def callback(input, prefixed_name, context, **kwargs):
-                        if prefixed_name == connection.input_name:
+                        if prefixed_name == connection.input_name and not input.type == "text":
```
jmchilton: I think I would be more comfortable with ``hasattr(get_options)`` - seems like it would be less brittle against adding/removing parameter types?

mvdbeek: Just looking through the code, all parameter types that have `get_options` are all subclasses of `SelectToolParameter`. Given this was added specifically for select tool parameter, should we make this
```suggestion
                        if prefixed_name == connection.input_name and isinstance(input, SelectToolParameter):
```
instead ? This seems a little clearer in intent ?
hasattr is fine too, let me know what you prefer.

jmchilton: Either is fine with me - I mildly prefer hasattr as being a little more Pythonic IMO.

mvdbeek: Got it, thanks @jmchilton!

PR #10976, opened 2020-12-20
Add workflow invocation detail component

mvdbeek:
- Shows workflow inputs and outputs (parameters + data)
- Shows all workflow steps (even the steps that haven't been scheduled yet)
- Shows subworkflow invocations inline
- Shows a table of all jobs for an invocation step
- Shows parameters for a job
- Shows dependencies of a job (why hasn't this step started yet ?)
 
Screenshot 2020-12-20 at 19 00 40

TODO:
  - [x] Is the provider pattern the right thing to use ?
  - [x] Should data pass through pouchdb ?
  - [x] History panel should update ?
  - [x] How do I modify/delete objects ?
  - [ ] Indentation can be a bit confusing
  - [ ] Maybe collect datasets / dataset collections as data inputs & data outputs ?


client/src/components/JobInformation/JobInformation.vue
```diff
@@ -88,79 +44,34 @@
...
-        };
+    components: {
+        CodeRow,
+        DecodedId,
     },
     props: {
-        hda_id: {
-            type: String,
-            required: true,
-        },
```
jmchilton: Past John convinced himself there were times the Job API would reject requests for information but that if the an accessible HDA was passed it made sense to render the information. There must be some comments about it server side somewhere? I hope you have thought through all of that. Like maybe this component is sitting on a published Page (it can be used from the Markdown engine) with a public HDA that doesn't belong to the user but that the viewer of the page should be able to see? Maybe this wouldn't work in that case anyway, no promises either way.

mvdbeek: The github UI didn't get updated, but this was merged with https://github.com/galaxyproject/galaxy/pull/10980. I'll see what happens. In case this doesn't work, do you think we can consider a job accessible if any one of its outputs is accessible ?

jmchilton: Who merged that without my review.... ... oh 🤣 .

Umm.. I think effectively it was - if you want to formalize that by just allowing access by job_id I think that is fine. I was clearly being very conservative here - because it is a thorny question and I didn't want to make any decisions, just preserve what was working.

mvdbeek:
Alright, the last 3 commits integrate pouchdb with vuex in the datasetsStore and datasetCollectionsStore. The idea is that we can listen for updates to anything in pouchdb and put that into the store (websocket updates into should integrate nicely with this in the future). Usage is the same in the components as if we didn't use pouchdb. This gives us synchronized updates between shared items in pouchdb and vuex. We can for instance update/delete/hide a dataset or dataset collection that is referenced twice in a invocation, and see that it updates in both places, as well as in the history.

Thanks for giving me a primer on pouchdb @Nerdinacan ... and sorry about shoving that into vuex, that is probably a bad idea 😆.
With this I think I have all the functionality I wanted for a first iteration. There's a ton more we could do (pagination of large collections, make it prettier, more annotations of inputs and outputs, export invocation button, download all outputs etc), but this should be useful at this point.

So from here on I'll write tests tests and more tests and I hope we can get this into 21.01.

Nerdinacan:
Sorry about responding so late. I stayed up too late a few days back and now my sleep schedule is bananas. I'll pull this down and run some tests tonight.

I'm most concerned about the way changes() is failing you. Ideally you should be able to subscribe using monitorContentQuery multiple times with any number of selectors because a single change feed is created for each PouchDB database instance and the filtered for each subscriber.

I'll write some tests to verify that and check out what you've done here.

Nerdinacan:
@mvdbeek 
I think there are some problems with this plan. The monitors you fire up in your $init functions are going to copy everything from the cache into Vuex and it's never going to go away until you close the browser window. There shouldn't be a problem for smaller data sets.

I think the reason you did this is because you're trying to make access to the data look like the Vuex api. But this overhead problem you're seeing is one of the reasons I went for the data provider strategy in the first place.

Is your goal to have easier access to the cached data, or is your goal to explicitly use Vuex?

Is it that you want to be able to have something like a mapGetter to give you method-based access to the cached data changes, and not have to use a provider component? I question why you feel that is necessary, and can show you techniques that will give you that ability, even if you're still using a provider.

But if you are absolutely trying to have the data be in Vuex instead of the cache, I think you're stuck duplicating everything as you do here.

So I guess I could use a statement of your goals here. Do you just want a convenient method-based api? Or do you want Vuex to be the cache?

mvdbeek:
> I'm most concerned about the way changes() is failing you. Ideally you should be able to subscribe using monitorContentQuery multiple times with any number of selectors because a single change feed is created for each PouchDB database instance and then filtered for each individual subscriber's request config.

I tried that, and in the current state that didn't work without this warning: 
```
events.js?faa1:46 MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 _pouch_galaxy-content-development configs listeners added. Use emitter.setMaxListeners() to increase limit
    at _addListener (webpack-internal:///./node_modules/events/events.js:206:15)
    at Changes.addListener (webpack-internal:///./node_modules/events/events.js:222:10)
    at Changes.addListener (webpack-internal:///./node_modules/pouchdb/lib/index-browser.es.js:528:8)
    at changes (webpack-internal:///./node_modules/pouchdb/lib/index-browser.es.js:5579:20)
    at PouchDB.idbChanges (webpack-internal:///./node_modules/pouchdb/lib/index-browser.es.js:6134:12)
    at Changes$1.doChanges (webpack-internal:///./node_modules/pouchdb/lib/index-browser.es.js:1758:28)
    at eval (webpack-internal:///./node_modules/pouchdb/lib/index-browser.es.js:1728:12)
```
I see now how monitorQuery works, and putting log statements around https://github.com/jmchilton/galaxy/blob/f2f12e09660426a2b7f9753612deeac3e5fee7e7/client/src/components/History/caching/db/changes.js#L18 I see that I'm not getting a new feed, so I agree with you that it should work. Does switchMap on the selector in the monitorQuary maybe unsubscribe, triggering a new buildFeed ? Although I think I should have seen that in my debugging ?

I did manage to just subscribe to a generic changes feed and filter myself, but that means doing N filters for everything I'm interested in, I think that's going to scale worse than updating the vuex cache by id.

I also think I got confused between the pouchdb docs, and the actual implementation of the changes filtering in monitorQuery, which uses https://github.com/HitomiTenshi/pouchdb-selector-core/blob/master/src/in-memory-filter.js. It looks like this is also doing fairly naive in-memory filtering. I had assumed this was doing something highly optimized directly on the database. Is that really going to scale for 10-100 thousands of updates, given that each update needs to go through the filter? 

Should we instead be using
```
db.changes({
    ...cfg, selector: {
         someSelector
    }
})
```
? (Which seems like it'd legitimately put us into `MaxListenersExceededWarning` territory 🚀 )

> The monitors you fire up in your $init functions are going to copy everything from the cache into Vuex and it's never going to go away until you close the browser window

I could build a `$or` selector on the ids in `datasetByHDAId`, and maybe some mechanism of reference counting. If there's no getter on a value, drop it from the store ?

> Is your goal to have method-based access to the cached data, or is your goal to explicitly use Vuex?

The goal is to be able to have a global place from which components can get objects by id, and the objects should be always up to date. I am not particularly tied to Vuex, but it does seem to fit the bill. Subscribing to changes in pouchdb is fine as well, but that seems to require filtering every update in every component for each event, or live with thousands of event emitters. With pouchdb updating a Vuex store we'd have to do the filtertin too, but just once per update and we'd only have one event emitter per store.

Nerdinacan:
I'm going to do some homework and see if there's something else closer to the metal for evaluating whether a change event fits a selector, but seeing as the function I use for matching a document to a selector comes right out of the core pouchdb module, I'm not super-hopeful.

During normal operations, for example, that entire history list has 1 subscriber to a single shared internal change feed. 

But if users start making heavily-nested trees of providers each with a separate query (as you may end up doing here in your invocation output tree) we will indeed have to match that change event once for each separate subscriber.

One thing we might try: We could do some separation of the query selectors themselves and share change-feeds among subscribers that share equal selectors. That wouldn't be [too hard to tack-on](https://rxjs-dev.firebaseapp.com/api/operators/groupBy). For instance if you have 10 components on the page that all want updates for dataset id=100, we might be able to assign them the same feed results grouped by that selector, ultimately reducing the number of match comparisons. A more likely scenario is a feed of changes for all content but grouped by history_id, for instance.

I would wager the whole shebang will be quite performant even if we just brute-force it given it's all done in-memory, and inside the worker. Nonetheless I'll check out [PouchDB's sparse, outdated docs](https://pouchdb.com/api.html) to see if there's anything we can leverage in the database layer itself.

Thanks!

mvdbeek:
I'm still writing tests that I should've written from the start, but putting this out of WIP if someone wants to merge it.

jmchilton (approved):

jmchilton:
Amazing!!!!!!!!!!!!!!!

mvdbeek:
Thank you @jmchilton @Nerdinacan!

PR #10980, opened 2020-12-21
Simplify JobInformation and DatasetInformation components

mvdbeek:
JobInformation contained a lot of dataset related infos.
This reworks the JobInformation component to
- only deal with job-related fields
- pushes the off-topic fields into DatasetInformation

This also adds a component that can nicely display rows with
code (stdout/stderr/command line) and that can expand these.
I also refactored out the id decoding logic into a standalone component
for re-use.
Finally I fixed the varying table widths across the different tables by making `table.info_data_table` a global style.

This will be very useful for the invocation PR.
This is how it looks like now:
Screenshot 2020-12-21 at 19 46 04

### DIFF
--- a/client/src/components/JobInformation/CodeRow.vue
+++ b/client/src/components/JobInformation/CodeRow.vue
@@ -2,13 +2,13 @@
     <tr @click="toggleExpanded">
         <td>
             {{ codeLabel }}
-            <span v-if="codeItem && !expanded"><i>(Click to expand)</i></span>
         </td>
         <td v-if="expanded">
             <pre class="code">{{ codeItem }}</pre>
         </td>
         <td v-else-if="codeItem">
             <pre class="code preview">{{ codeItem }}</pre>
+            <b>(Click to expand)</b>
         </td>
         <td v-else><i>empty</i></td>
     </tr>
--- a/client/src/components/JobInformation/CodeRow.vue
+++ b/client/src/components/JobInformation/CodeRow.vue
@@ -1,0 +1,38 @@
+<template>
+    <tr @click="toggleExpanded">
+        <td>
+            {{ codeLabel }}
+            <span v-if="codeItem && !expanded"><i>(Click to expand)</i></span>
+        </td>
+        <td v-if="expanded">
+            <pre class="code">{{ codeItem }}</pre>
+        </td>
+        <td v-else-if="codeItem">
+            <pre class="code preview">{{ codeItem }}</pre>
+        </td>
+        <td v-else><i>empty</i></td>
+    </tr>
+</template>
+<script>
+import Vue from "vue";
+import BootstrapVue from "bootstrap-vue";
+
+Vue.use(BootstrapVue);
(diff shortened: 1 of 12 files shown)


OlegZharkov:
Since, you're here already. Is it possible to Vuefy dependencies in this PR as well?

mvdbeek:
Can we do that in a different PR ?

client/src/components/JobInformation/CodeRow.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+    <tr @click="toggleExpanded">
+        <td>
+            {{ codeLabel }}
+            <span v-if="codeItem && !expanded"><i>(Click to expand)</i></span>
```
OlegZharkov:
```suggestion
```

client/src/components/JobInformation/CodeRow.vue
```diff
@@ -0,0 +1,38 @@
...
+        </td>
+        <td v-if="expanded">
+            <pre class="code">{{ codeItem }}</pre>
+        </td>
+        <td v-else-if="codeItem">
+            <pre class="code preview">{{ codeItem }}</pre>
```
OlegZharkov, line 10:
```suggestion
            <pre class="code preview">{{ codeItem }}</pre>
            <b>(Click to expand)</b>
```

OlegZharkov (commented):
Thanks a lot @mvdbeek, this seem to work perfectly. Especially, I liked `Click to expand` feature.
The only comment I have here, that maybe it would be a good idea, to make `(Click to expand)` more visible. Namely, write this note under the code section, instead of the left (title) column. Just to make more obvious

image

However I don't have a strong opinion about it. If you don't like it, please ignore and merge it, it's a great cleanup, thanks :+1:

mvdbeek:
I really like that, thanks for the suggestion!

client/src/components/JobInformation/CodeRow.vue
mvdbeek: I guess we can remove the whole span, then right ?

OlegZharkov: right :+1:

jmchilton:
Run prettier and I'll merge, looks great to me.

jmchilton (approved):

simonbray:
It would be nice if the expanding could distinguish between clicking and dragging. One of the main things I (and I assume other users) want to do with stderr/stdout errors is to copy the message for an internet search, and it's quite annoying that selecting the text to copy recollapses the box.

OlegZharkov:
> It would be nice if the expanding could distinguish between clicking and dragging. One of the main things I (and I assume other users) want to do with stderr/stdout is to copy the message for an internet search, and it's quite annoying that selecting the text to copy recollapses the box.

we can make it always expand, without recollapsing at all, if that helps. So you can expand and copy your message. 
(Also create a separate issue please)

Issue #9641, opened 2020-04-21
Python 3.7 + uWSGI freeze on login/register.


dannon:
There's a bug with python 3.7 + uWSGI that causes the process to lock up when logging in.

The last lines in the log are:
```
[pid: 40771|app: -1|req: -1/85] 127.0.0.1 () {42 vars in 993 bytes} [Wed Dec 11 08:56:11 2019] GET /galaxy/blah/static/scripts/bundled/assets/af7ae505a9eed503f8b8e6982036873e.woff2 => generated 77160 bytes in 0 msecs via offload() (HTTP/1.0 200) 3 headers in 114 bytes (1 switches on core 2)
galaxy.webapps.galaxy.controllers.user DEBUG 2019-12-11 08:56:12,697 [p:40771,w:1,m:0] [uWSGIWorker1Core2] trans.app.config.auth_config_file: /Users/yoplait/work/galaxy/config/auth_conf.xml
```

From @mvdbeek:
The 3.7 bug was this:
```
Exception ignored in: <function _releaseLock at 0x7fa8eb20b9d8>
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/logging/__init__.py", line 226, in _releaseLock
    _lock.release()
RuntimeError: cannot release un-acquired lock
```
xref:  unbit/uwsgi#1978

Details from conversation on gitter at: https://gitter.im/galaxyproject/dev?at=5df0f8d47d382b523f5bd171

dannon:
After more testing, and debugging with `py-spy`, a fix is in #9642

PR #10983, opened 2020-12-22
Overwrite uwsgi default py-call-osafterfork

jdavcs:
Prevents runtime error in logging module under Python 3.7.1+

Ref:
https://github.com/galaxyproject/galaxy/issues/9641
https://github.com/unbit/uwsgi/issues/1978
https://gitter.im/galaxyproject/wg-backend?at=5fe0f93daa6bb528c3636941 (most recent discussion)

Run `make config-rebuild`.

### DIFF
--- a/lib/galaxy/config/config_manage.py
+++ b/lib/galaxy/config/config_manage.py
@@ -154,7 +154,7 @@ UWSGI_OPTIONS = OrderedDict([
         'type': 'str',
     }),
     ('py-call-osafterfork', {
-        'desc': """Feature necessary for proper mule signal handling. The default is set to false to prevent a runtime error under Python 3.7+ (see https://github.com/unbit/uwsgi/issues/1978).""",
+        'desc': """Feature necessary for proper mule signal handling on Python versions below 3.7.2. The default is set to false to prevent a runtime error under Python 3.7.2 and newer (see https://github.com/unbit/uwsgi/issues/1978).""",
         'default': False,
         'type': 'bool',
     }),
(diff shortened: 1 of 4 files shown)


lib/galaxy/config/config_manage.py
```diff
@@ -691,6 +692,14 @@ def _build_sample_yaml(args, app_desc):
     _write_to_file(args, f, destination)
 
 
+def _overwrite_uwsgi_defaults():
+    # Prevents runtime error in logging module under Python 3.7.1+
```
mvdbeek: Can you add a note to the schema as well ? Admins should know whether or not to use this.

jdavcs: Done. 
Also, apparently, `UWSGI_OPTIONS` is defined in the file (we already overwrite the uwsgi's schema description), so it's all much simpler (I've removed the extra function).

lib/galaxy/config/config_manage.py
```diff
@@ -154,8 +154,8 @@
         'type': 'str',
     }),
     ('py-call-osafterfork', {
-        'desc': """Feature necessary for proper mule signal handling""",
-        'default': True,
+        'desc': """Feature necessary for proper mule signal handling. The default is set to false to prevent a runtime error under Python 3.7+ (see https://github.com/unbit/uwsgi/issues/1978).""",
```
mvdbeek:
```suggestion
        'desc': """Feature necessary for proper mule signal handling on Python versions below 3.7.2. The default is set to false to prevent a runtime error under Python 3.7.2 and newer (see https://github.com/unbit/uwsgi/issues/1978).""",
```

mvdbeek:
Thank you @ic4f!

PR #10984, opened 2020-12-22
Put new toolshed tools in existing tool panel sections if the ID matches

davebx:
This change will place a newly installed toolshed tool, if the target section exists, in that previously defined section in the shed_tool_conf file. Previously the section would simply be placed at the very end of the shed tool config, whether or not that section already existed in the XML.


lib/galaxy/tool_shed/galaxy_install/tools/tool_panel_manager.py
```diff
@@ -56,7 +56,21 @@ def add_to_shed_tool_config(self, shed_tool_conf_dict, elem_list):
...
+                section_found = False
+                if elem_entry.tag == 'section':
+                    # See if there is already a section with a matching ID
+                    for existing_elem in config_elems:
+                        if existing_elem.tag == 'section' and existing_elem.attrib.get('id', None) == elem_entry.attrib.get('id', None):
+                            section_found = True
+                            for child in elem_entry:
+                                existing_elem.append(child)
+                                break
+                            if section_found:
```
bgruening: `section_found` is here always true, left?

davebx: It's set to `False` for each new `<section>` tag in `elem_list`

mvdbeek: At line 64 you're setting this to `True`. I don't see a condition where this could be False after that. The code is not wrong, it's just confusing, you can just break here unconditionally.

davebx: @mvdbeek that's a good point about the unconditional break, I removed the condition and added some more comments to clarify the flow. And then I realized I don't need `section_found` at all, so the code should be even more readable now.

mvdbeek:
This currently leaves an empty section behind in the shed_tool_conf.xml file,
e.g
```
<section id="galaxyp" name="galaxyp" version=""/>
```
Do you have a moment to look at this @davebx ?

lib/galaxy/tool_shed/galaxy_install/tools/tool_panel_manager.py
```diff
@@ -56,7 +56,21 @@ def add_to_shed_tool_config(self, shed_tool_conf_dict, elem_list):
...
+                    for existing_elem in config_elems:
+                        # Compare the section ID for each one to the section ID for the tool being installed.
+                        if existing_elem.tag == 'section' and existing_elem.attrib.get('id', None) == elem_entry.attrib.get('id', None):
+                            # We've found a section, add the incoming tools to it.
+                            for child in elem_entry:
+                                existing_elem.append(child)
+                            # Break out of the config_elems loop back to the elem_list loop.
+                            break
+                    # If we reach this point, no section was found. Create one with contents.
+                    config_elems.append(elem_entry)
```
mvdbeek:
```suggestion
                    else:
                        config_elems.append(elem_entry)
```
This would avoid the empty section getting added.

mvdbeek (approved):

lib/galaxy/tool_shed/galaxy_install/tools/tool_panel_manager.py
```diff
@@ -54,22 +54,33 @@ def add_to_shed_tool_config(self, shed_tool_conf_dict, elem_list):
             root = tree.getroot()
             for elem in root:
                 config_elems.append(elem)
+            # Build a list of tool panel section IDs in memory
```
mvdbeek: Why did you change this ? This looks like a more verbose version of what you had before.

PR #10985, opened 2020-12-23
[20.09] If a workflow defines explicit workflow_outputs, respect those.

jmchilton:
Currently the editor just makes everything an output that isn't hidden.

Without this change, I don't see a way to have non-hidden outputs to tool steps that are not workflow outputs.


mvdbeek (approved):
From a debugging / support perspective I think I'd prefer to target dev with this.

jmchilton:
It was merged then with workflow linting. I'm remaining worried about assigning all these un-explicit workflow outputs. But I should have done a better job catching this sooner and working on the display and usability of these concepts - it is on me.

PR #10987, opened 2020-12-23
Fixes for license and creator saving and updating.

jmchilton:
I was mutating a property, it was possible to save the workflow and lose these, etc.. This new approach is just to delegate to very simple reactive properties in Index.vue and just roll this stuff into the change checking and update framework that it already has. This approach is much cleaner and gives Index.vue the single source of truth (and simply reactive truth) on these properties - which is nice and needed in some downstream work.

I was emulating annotation handling - which is an anti-pattern and should probably be redone to do what I'm doing here. I could have designed a much simpler license selection widget if I would have been just doing this from the get go and not try to emulate annotations.

### DIFF
--- a/client/src/components/Workflow/Editor/Attributes.vue
+++ b/client/src/components/Workflow/Editor/Attributes.vue
@@ -34,11 +34,11 @@
         </div>
         <div id="workflow-license-area" class="mt-2">
             <b>License</b>
-            <LicenseSelector :inputLicense="licenseCurrent" @onLicense="onLicense" />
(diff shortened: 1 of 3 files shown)


dannon:
@jmchilton Good catch here; in a separate branch (client-testing-followup in my repo) I have an updated eslint which checks for these mutated properties, and this was one that got flagged -- I just hadn't gone through and fixed it yet, so thanks!  There are a few other components that do similar things that will need to be addressed, but thanks for taking care of this one.

bgruening:
I count Dannon's comment as approval :)
We need to lock Dannon in so he can PR all his magic branches! :)

Issue #9166, opened 2019-12-26
Diff-based Workflow Updates


jmchilton:
Currently the workflow editor and the update endpoint generally serializes the whole workflow state and rewrites all the steps on each save. This doesn't play well workflows imported from hand crafted files. Comments, styling in the files, extensions not tracked in the database, etc.. will be lost.

This is a very serious limitation for Galaxy syncing workflows (introduced in https://github.com/galaxyproject/galaxy/pull/7019) and is probably a serious limiting factor for people crafting Format 2 workflows as a result. I suspect we need to close the loop and allow Galaxy and users to edit the same files.

If workflow updates didn't update everything all at once but instead produced a sequential set of transformations to apply to the original workflow to produce the target workflow - we could apply the transformations to the underlying workflow representation. We can use ruamel.yaml round-trip dumping (https://yaml.readthedocs.io/en/latest/example.html) to preserve formatting, comments, etc.. This is the technique used by the CWL upgrader for instance (https://github.com/common-workflow-language/cwl-upgrader).

This can probably be done entirely server-side at first - and that would be a great step. Once we have a language for describing the transformations however it would ultimately be good to track workflow editor state changes in terms of these transformations as well.

Going a step further and tracking the transformations in the database as well as extra metadata - we can produce a dynamic "CHANGELOG" for workflow version browser as well.

Implementing this in the workflow editor would make implementing a workflow undo button a lot more feasible - https://github.com/galaxyproject/galaxy/issues/644. We'd have a clear way to generate a description of changes that we can claim to "un-do" and un-doing them could be popping them off the stack of tracked changes.

Issue #8546, opened 2019-08-30
Add a common-problems / check option for workflows


mvdbeek:
Which would for instance flag missing input modules, old-style workflow parameters, tool version changes, disconnected steps, parameter changes, etc ... all sort of severe or not so severe things that could prevent a workflow from running. I think that would be especially useful when Galaxy is used as a workflow engine. 

The missing inputs issue came up in https://help.galaxyproject.org/t/how-to-invoke-workflow-using-galaxys-rest-api/2014/3

mvdbeek:
Done in https://github.com/galaxyproject/galaxy/pull/10988

PR #10988, opened 2020-12-23
UI & API for Supporting Best Practice Workflows

### LINKED ISSUE #6961: Add the ability to enter help text for workflow parameters (and input datasets/collections)
It would be nice to be able to enter a help message for workflow parameters.  I try to make the names of the parameters as descriptive as possible, but you can't exactly describe an equation to use to calculate the value in the parameter/variable name.

I envision it working like this:

1. Click a workflow parameter at the top-right of the workflow editor, which causes a form to show up in the right pane where there is a box to enter help text
2. Enter any help message there (and perhaps a field title?)

In fact, input datasets and collections should have help text as well.  I've tried filling out all those fields, but the only one that shows up in a workflow form is the label.

Use case example: I have a workflow that generates a file of reference amplicon lengths.  The subsequent workflow needs to be run on a set of heavily mutated amplicon data.  It stitches together paired-end reads using FLASH, but in order to get accurate results, the min & max overlap must be calculated somewhat tightly, so I want to give the user the following note & equation to do the calculation:

Set a minimum and maximum overlap of +/- 10 of the expected overlap.  The expected overlap can be calculated by: `overlap = (amplicon_length - (2 * (amplicon_length - read_length)))`.

(Alternatively, if there was a way to have the 1st line of a tool's output used to set a workflow parameter, I might be able to figure out a way to avoid asking the user to enter a value all together.)  I could just ask them for the read length (or use some tool to discern it).

### LINKED ISSUE #9096: Migration path for old-style #{variable} workflow parameters
The new workflow parameters are pretty cool, but usage of `#{variable}` inside text/integer/float fields in existing workflows is widespread (and in some cases easier to use). I think it should be possible to insert instances of appropriate expression tools or workflow parameters on the client-side. When the workflow is then persisted we persist it in its new form. 

I think this would allow eliminating the code on the backend that deals with these parameters (not sure if it is a lot tbh) and makes it easier to get started with the new workflow parameters.

jmchilton:
## Overview

The PR adds a workflow editor panel for displaying information and options aimed at improving workflows and having that conform to best practices.

## Quick Walkthrough

Here is a legacy workflow that uses older-style "workflow parameters" and has a disconnected, non-optional input. Both of these things make this workflow inappropriate for the new simplified workflow UI, usage via the API, and Planemo workflow testing.

Screen Shot 2020-12-23 at 3 52 32 PM

The PR adds an option for a best practices panel.

Screen Shot 2020-12-23 at 3 52 58 PM

This panel explains these issues in detail and lets the user fix them in bulk or one at a time.

Screen Shot 2020-12-23 at 3 53 14 PM

After clicking the fix all button (and I cheated and auto laid out the result), I have a workflow with formal workflow inputs and parameters.

Screen Shot 2020-12-23 at 3 53 48 PM

The parameters still don't have annotations and the input doesn't have an annotation or a label - and so we can see there is more manual work to be done in the "Best Practices" panel.

## Workflow Parameters

Ostensibly, the goal here is to replace older-style "workflow parameters" with steps of type "parameter_input". The latter works with the API, is easier to reason about based on the workflow representation, and can be marked up with type information, unique label, and annotations - all improving usability (fixes #6961).

Part of the difficulty here is that older style "workflow parameters", that I'm going to refer to now as "legacy parameters", work very differently when used to replace tool inputs versus when used as parameters in post job actions. 

"Legacy parameters" when used in post job actions as replacement variables are relatively well behaved. They are added to the database in a structured way and are usable via the API in a very reasonable fashion.

To support this post job action use case with "step parameters", I've modified the workflow runtime to just add these to the workflow replacement dictionary if they aren't explicitly sent via the existing replacement dict submission mechanism. This works pretty transparently. Now post job action variables can reference step parameters. They can now have annotations, etc.. and workflow API consumers just need to understand one uniform way of submitting parameters.

On the other hand, when legacy parameters are used as tool inputs, only the older-style workflow run GUI worked and only worked by replacing the complete state of every step at submission time. We want to get away from that behavior - the UI wasn't performant, the UI was complicated, these workflows couldn't be executed via the API, and for traceability workflows should be submitted only with annotated inputs that are tracked in a structured manner in the database.

The bulk of this work is about providing support to users to replace their legacy parameters with step parameters in the workflow editor. They could have been replaced manually before, but if the parameter was used in multiple places this would have been an error prone process.

To provide this support, this pull request introduces a "Workflow Best Practices" panel capable of performing this complex refactoring, as well as others, via a new fully typed workflow refactoring API.

Fixes #9096

## Workflow Best Practices Panel

The workflow menu now has a "Check for Best Practices" option. This will replace the right pane with a series of collapsible sections labeled pass/warn for various non-best practice behavior. (I use the word linting a lot in the code, but I think I have managed to avoid exposing it to the user.)

The sections each have a significant amount of help text explaining issues as well as the justification for they are not best practice.

The sections include:

- "Parameters" - this ensures no legacy parameters are used and if they are used, they are listed along with an option to convert them to a step input parameter.
- "Disconnected Inputs" - this ensures no workflow steps have disconnected, non-optional inputs. These should all be attached to formal workflow input data parameters - and an option to convert them is presented for each disconnected input.
- "Input Metadata" - this displays any workflow inputs that don't have a label or annotation to the user along with which is missing.
- "Outputs" - this ensures the workflow has at least one explicit workflow output and has no unlabeled workflow outputs. An option is provided to remove all unlabeled outputs as formal workflow outputs.
- "Metadata - License" - this ensures the workflow has a license and provides the user a link to the attributes page to add one if it isn't provided.
- "Metadata - Creator" - this ensures the workflow has creator information annotated and provides the user a link to the attributes page to add it if it isn't provided.

All of the modifications this panel can make automatically are made through the type workflow refactoring API. This API makes it very easy to combine multiple refactoring steps into lists of actions - so this panel also has a link at the top to simply make all the automatic changes it knows how to to fix issues if the user wants to quickly fix up everything.
The refactoring API calls are all mitigated through Index.vue that ensures the workflow is saved and such before executing them. The API also has a dry_run parameter - so the UI attempts the refactoring first in dry run mode before actually making the changes.

Implements #8546

## Typed Workflow Refactoring API

The workflow refactoring API takes in a list of actions to perform (along with a dry_run parameter) and then simply executes them in order to produce the next version of a Workflow for a StoredWorkflow object.
I've implemented a handful of types and tests for each of them. The ability to compose them and execute them in order is powerful as demonstrated by how clean the ability to just execute all refactorings steps in order from Lint panel is.

The actions are all typed using Pydantic, so the API has incredibly intuitive and uniform validation and should be trivial to create documentation from.

This API is the first big piece toward implementing https://github.com/galaxyproject/galaxy/issues/9166. Eventually, I'd like all workflow modifications to come through structured, tested steps that we can just store along with the workflow version in the database. Having such structured edits has numerous benefits including allowing us to generate changelogs, implement undo functionality, and apply simultaneous edits to the database and human written workflow YAML files.

### DIFF
--- a/lib/galaxy/app.py
+++ b/lib/galaxy/app.py
@@ -20,7 +20,10 @@ from galaxy.managers.libraries import LibraryManager
 from galaxy.managers.roles import RoleManager
 from galaxy.managers.tools import DynamicToolManager
 from galaxy.managers.users import UserManager
-from galaxy.managers.workflows import WorkflowsManager
+from galaxy.managers.workflows import (
+    WorkflowContentsManager,
+    WorkflowsManager,
+)
 from galaxy.model.database_heartbeat import DatabaseHeartbeat
 from galaxy.model.tags import GalaxyTagHandler
 from galaxy.queue_worker import (
@@ -105,6 +108,7 @@ class UniverseApplication(config.ConfiguresGalaxyMixin):
         self.history_manager = HistoryManager(self)
         self.hda_manager = HDAManager(self)
         self.workflow_manager = WorkflowsManager(self)
+        self.workflow_contents_manager = WorkflowContentsManager(self)
         self.dependency_resolvers_view = DependencyResolversView(self)
         self.test_data_resolver = test_data.TestDataResolver(file_dirs=self.config.tool_test_data_directories)
         self.library_folder_manager = FolderManager()
--- a/lib/galaxy/workflow/refactor/execute.py
+++ b/lib/galaxy/workflow/refactor/execute.py
@@ -1,0 +1,363 @@
+import logging
+
+from galaxy.exceptions import (
+    RequestParameterInvalidException,
+)
+from galaxy.tools.parameters import (
+    visit_input_values,
+)
+from galaxy.tools.parameters.basic import (
+    ConnectedValue,
+    contains_workflow_parameter,
+    runtime_to_json,
+)
+from .schema import (
+    AddInputAction,
+    AddStepAction,
+    ConnectAction,
+    DisconnectAction,
+    ExtractInputAction,
+    ExtractLegacyParameter,
+    InputReferenceByOrderIndex,
+    OutputReferenceByOrderIndex,
+    RefactorRequest,
+    RemoveUnlabeledWorkflowOutputs,
+    step_reference_union,
+    StepReferenceByLabel,
+    UpdateAnnotationAction,
+    UpdateCreatorAction,
+    UpdateLicenseAction,
+    UpdateNameAction,
+    UpdateReportAction,
+    UpdateStepLabelAction,
+    UpdateStepPositionAction,
+)
+from ..modules import (
+    InputParameterModule,
+    NO_REPLACEMENT,
+)
+
+log = logging.getLogger(__name__)
+
+
+class WorkflowRefactorExecutor:
+
+    def __init__(self, raw_workflow_description, workflow, module_injector):
+        # we mostly use the ga representation, but there may be cases where the
+        # models/modules of existing workflow are more usable.
+        self.raw_workflow_description = raw_workflow_description
+        self.workflow = workflow
+        self.module_injector = module_injector
+
+    def refactor(self, refactor_request: RefactorRequest):
+        for action in refactor_request.actions:
+            action_type = action.action_type
+            refactor_method_name = "_apply_%s" % action_type
+            refactor_method = getattr(self, refactor_method_name, None)
+            if refactor_method is None:
+                raise RequestParameterInvalidException(
+                    f"Unknown workflow editing action encountered [{action_type}]"
+                )
+            refactor_method(action)
+
+    def _apply_update_step_label(self, action: UpdateStepLabelAction):
+        step = self._find_step_for_action(action)
+        step["label"] = action.label
+
+    def _apply_update_step_position(self, action: UpdateStepPositionAction):
+        step = self._find_step_for_action(action)
+        step["position"] = action.position.to_dict()
+
+    def _apply_update_name(self, action: UpdateNameAction):
+        self._as_dict["name"] = action.name
+
+    def _apply_update_annotation(self, action: UpdateAnnotationAction):
+        self._as_dict["annotation"] = action.annotation
+
+    def _apply_update_license(self, action: UpdateLicenseAction):
+        self._as_dict["license"] = action.license
+
+    def _apply_update_creator(self, action: UpdateCreatorAction):
+        self._as_dict["creator"] = action.creator
+
+    def _apply_update_report(self, action: UpdateReportAction):
+        self._as_dict["report"] = {"markdown": action.report.markdown}
+
+    def _apply_add_step(self, action: AddStepAction):
+        steps = self._as_dict["steps"]
+        order_index = len(steps)
+        step_dict = {
+            "order_index": order_index,
+            "id": "new_%d" % order_index,
+            "type": action.type,
+        }
+        if action.tool_state:
+            step_dict["tool_state"] = action.tool_state
+        if action.label:
+            step_dict["label"] = action.label
+        if action.position:
+            step_dict["position"] = action.position.to_dict()
+        steps[order_index] = step_dict
+
+    def _apply_add_input(self, action: AddInputAction):
+        input_type = action.type
+        module_type = None
+
+        tool_state = {}
+        if input_type in ["data", "dataset"]:
+            module_type = "data_input"
+        elif input_type in ["data_collection", "dataset_collection"]:
+            module_type == "data_collection_input"
+            tool_state["collection_type"] = action.collection_type
+        else:
+            if input_type not in InputParameterModule.POSSIBLE_PARAMETER_TYPES:
+                raise RequestParameterInvalidException(f"Invalid input type {input_type} encountered")
+            module_type = "parameter_input"
+            tool_state["parameter_type"] = input_type
+
+        for action_key in ["restrictions", "suggestions", "optional", "default"]:
+            value = getattr(action, action_key, None)
+            if value is not None:
+                tool_state[action_key] = value
+
+        if action.restrict_on_connections is not None:
+            tool_state["restrictOnConnections"] = action.restrict_on_connections
+
+        add_step_kwds = {}
+        if action.label:
+            add_step_kwds["label"] = action.label
+
+        add_step_action = AddStepAction(
+            action_type="add_step",
+            type=module_type,
+            tool_state=tool_state,
+            position=action.position,
+            **add_step_kwds
+        )
+        self._apply_add_step(add_step_action)
+
+    def _apply_disconnect(self, action: DisconnectAction):
+        input_step_dict, input_name, output_step_dict, output_name = self._connection(action)
+        output_order_index = output_step_dict["id"]  # wish this was order_index...
+        # default name is name used for input's output terminal - following
+        # format2 convention of allowing this be absent for clean references
+        # to workflow inputs.
+        all_input_connections = input_step_dict.get("input_connections")
+        self.normalize_input_connections_to_list(all_input_connections, input_name)
+        input_connections = all_input_connections[input_name]
+
+        # multiple outputs attached to this inputs, just detach
+        # that specific one.
+        delete_index = None
+        for connection_index, output in enumerate(input_connections):
+            if output["id"] == output_order_index and output["output_name"] == output_name:
+                delete_index = connection_index
+                break
+        if delete_index is None:
+            raise RequestParameterInvalidException("Failed to locate connection to disconnect")
+        del input_connections[delete_index]
+
+    def _apply_connect(self, action: ConnectAction):
+        input_step_dict, input_name, output_step_dict, output_name = self._connection(action)
+        output_order_index = output_step_dict["id"]  # wish this was order_index...
+        all_input_connections = input_step_dict.get("input_connections")
+        self.normalize_input_connections_to_list(all_input_connections, input_name, add_if_missing=True)
+        input_connections = all_input_connections[input_name]
+        input_connections.append({
+            'id': output_order_index,
+            'output_name': output_name,
+        })
+
+    def _apply_extract_input(self, action: ExtractInputAction):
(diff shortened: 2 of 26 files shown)


lib/galaxy/app.py
```diff
@@ -20,7 +20,7 @@
 from galaxy.managers.roles import RoleManager
 from galaxy.managers.tools import DynamicToolManager
 from galaxy.managers.users import UserManager
-from galaxy.managers.workflows import WorkflowsManager
+from galaxy.managers.workflows import WorkflowContentsManager, WorkflowsManager
```
mvdbeek:
```suggestion
from galaxy.managers.workflows import (
    WorkflowContentsManager,
    WorkflowsManager,
```

lib/galaxy_test/api/test_workflows.py
```diff
@@ -1244,6 +1273,54 @@ def test_run_subworkflow_replacement_parameters(self):
...
+inputs:
+  replaceme: text
+steps:
+  create_2:
+    tool_id: create_2
+    state:
+      sleep_time: 0
+    outputs:
+      out_file1:
+        rename: "${replaceme} name"
```
mvdbeek, line 1312: Does this work if a step label `replaceme` and a step input `replaceme` exists ? What's the order of precedence there and would that be clear to the user ?

Could we make this not use the legacy syntax, which would be confusing (and it is confusing the best-practices panel right now, saying that this input should be fixed)?

lib/galaxy/workflow/refactor/execute.py
```diff
@@ -0,0 +1,363 @@
...
+                if not contains_workflow_parameter(value):
+                    return NO_REPLACEMENT
+
+                if value == target_value:
+                    target_tool_inputs.append((step.order_index, input, prefixed_name))
+                    replace_tool_state = True
+                    return runtime_to_json(ConnectedValue())
+                else:
+                    return NO_REPLACEMENT
+            visit_input_values(tool.inputs, tool_inputs.inputs, callback, no_replacement_value=NO_REPLACEMENT)
```
mvdbeek, line 245: This doesn't visit post job actions, and so every refactoring of PJAs fails afaict, while the refactoring panel offers to do this refactoring.

mvdbeek (commented):
I really love this work, this is awesome and a great foundation to build on!

I do see some issues from my testing though:
For tool inputs the refactoring is works fine (it doesn't handle `{$var1} test` yet), but I'm worried about the PJA replacements. I think it would be better if they didn't continue to use the `${input|basename} extra text` syntax. There's the minor issue that the best practices UI still wants to fix these, but I also think this is confusing to the user.

From the UI/UX viewpoint I think it would be better to use the same toggle we use for connecting workflow parameters, so that there is just one way to define them. Which also has the advantage that you can see in the editor where your parameters are consumed. For PJAs that can reference their own inputs this could be augmented to a select list of available inputs.

I am unsure that we already want to expose the legacy parameter fixing to users ... I can see this being frustrating if you can't actually fix your workflow yet because you're composing your parameter values in a way we can't fix. Or maybe we can offer this, but divide the fixes between `essential` (maybe in red) for disconnected inputs, missing labels, no workflow outputs and `recommended` (in orange/yellow?) for legacy parameters and mention the caveat that this doesn't work for every scenario yet?

lib/galaxy_test/api/test_workflows.py
jmchilton: I thought ``#{}`` was for tool step parameters and ``${}`` was for workflow parameters? Is this wrong?

mvdbeek: ... of course, that explains why this didn't work for me when I tried it in the editor. Which also resolves my comment on execute.py.

jmchilton: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/jobs/actions/post.py#L135 right?

``${}`` would only respect legacy workflow parameters or modern step parameters (of which I've augmented the workflow run form to just render one) and ``#{}`` would only respect tool input names?

Or do you mean the old workflow run form ... I guess I didn't look into that but who is going to have a step parameter with the same name as a legacy parameter in the tool state without having used the tool? I mean... seems an odd corner case but I can fix the legacy tool form to look at step parameters like they are legacy parameters and I think that would solve this.

lib/galaxy/workflow/refactor/execute.py
jmchilton: But that is fine right because with d7e178eca1479c3830526d85adea9c3075f8392e the parameters will work the same way and the tool form will only render the new version.

lib/galaxy_test/api/test_workflows.py
jmchilton: I guess I could also read this as "what if a legacy replacement_dict" is passed to the API and a modern step parameter was passed. My response would be for backward compatibility the ``replacement_dict`` is still used but the UI shouldn't render the options and API consumers are going to much prefer the modern way of specifying these. I think it is fine right?

mvdbeek: No, I think this comment here is resolved with me mixing up tool step and workflow parameters. But I do think referencing new style parameters for PJAs with the `${param}` syntax in the editor is unfortunate. Is there a reason why we can't use the same logic and UI we use for tool step parameters ?

jmchilton: I don't really know what modeling that connection gets us (from modern workflow parameter to PJA).

There is a fixed universe of modern, fixed workflow input parameters - we could easily build a UI to assist users in resolving them when writing PJA. From the backend perspective, I don't know what modeling that connection gets us - we have a dictionary of the inputs, we will probably always need it for legacy reasons, we can validate them, we have a distinct and unambiguous syntax.

This seems good enough and a simpler user and backend experience. We could be more formal in future... but this isn't worse than the legacy parameters right?

Explicit modeling of the tool inputs was really good - because the syntax was ambiguous and you can't reason at all about tool inputs without loading the complex tool - which was highly Galaxy specific. These PJA are no problem to reason about in gxformat2, etc.. well outside the context of Galaxy.

jmchilton:
Would it be useful to break the step parameters reworking and refactoring API commits out into their own PRs?

lib/galaxy_test/api/test_workflows.py
mvdbeek: So the advantages that I see is that you can see where your parameter is used and you can set a PJA parameter based on the output of a expression tool, like `compose_text_param` (haven't tried, maybe that works now too if you reference the output?). Seeing the parameters in use is quite helpful.

I agree that we have other means to potentially build a UI, but if that UI can be the same between tool input parameters and PJA inputs that would be easier for users.

> This seems good enough and a simpler user and backend experience

Not saying we need to implement a UI now, but I don't think `${}` is a good user experience, and i'd rather see less than more use of it in the UI. If we plan to keep this on the backend then I have no objection with the use here.

jmchilton: I guess I could go into why I disagree on some levels but I'll instead just say I think there are two areas where we probably sort of agree in terms of this direction.

- This isn't all PJA, this is just renaming. Those big workflows where the input name is preserved across 25 steps are an anti-pattern and even though tags and collections and step parameters and workflow reports didn't quite kill them they did a good job and I think #10118 will be their death knell 🎆🍾 . In all cases, displaying structure or preserved metadata is better than encoding it in the name. 
- We do need a way to define parameters to a step that don't corresponding 1-to-1 with tool or explicit subworkflow inputs for CWL - and way to define expressions that map those to the underlying inputs - and that should include PJA when mapped to Galaxy. These are fairly fundamental to CWL and are useful I think.

mvdbeek:
I think it makes sense to keep them in one.

Functionality-wise, should refactoring `${param}` in a PJA insert the workflow parameter if it doesn't exist yet ?
If you do define a workflow like that the best practices panel wants to fix it, but will fail with `Failed to find {target_value} in the tool state or any workflow steps`.

mvdbeek:
And do you have a plan for dealing with `${param} and some more text here` ? That could maybe become a `compose_text_param` step ?

jmchilton:
> Functionality-wise, should refactoring ${param} in a PJA insert the workflow parameter if it doesn't exist yet ? Or should refactoring exclude PJAs?

I think it should create the parameter, I don't have any particular objection to legacy parameters just used in PJA - they work fine with the API and and the simplified workflow form renders them okay but I suppose they can't be considered best practice without annotations. They will work with workflow test cases better also I think. So yeah... I will make that change.

jmchilton:
> And do you have a plan for dealing with ${param} and some more text here ? That could maybe become a compose_text_param step ?

I was planning to leave these as a TODO, with that being exactly the right fix. In #10989 I had annotating why they can't be replaced, maybe it is worth the time to handle them though? I wish I knew how many people would be using this.

mvdbeek:
I'd say this is probably quite common, but I could be wrong there. It was the first thing I tried, but that doesn't mean much.

jmchilton:
> Functionality-wise, should refactoring ${param} in a PJA insert the workflow parameter if it doesn't exist yet ? Or should refactoring exclude PJAs?

Okay - I think it does now! I added that and two new test cases.

I toyed with the idea of ``compose_text_param`` but then realized it was in the tool shed and lost interest 😆. It shouldn't be hard per se - it would just introduce a bunch of conditional logic in the UI and I would have to... use the tool shed. I have a bunch of production workflows that make heavy use of these I'm converting - my first scan through them all just had bare parameters - but I didn't pay too close attention. I'll let you know if this dialog proves wildly insufficient.

mvdbeek:
We could move compose_text_param into core ... it does provide some important functionality that maybe one would expect to ship with Galaxy? I put it in the IUC because I thought it'd be cool to maintain expression tools that way.

jmchilton:
I would say it isn't worth the effort at this time - to backport the tool or rewrite the parameters? There is some chance we will want to pull it out someday right? And there is some chance we would want to have like native workflow expressions - maybe on inputs or maybe as a stand-alone module type. If you did open the PR to add it to Galaxy though, I'd be happy to merge it.

mvdbeek:
This is working great, thanks so much!

PR #10994, opened 2020-12-26
[WIP] Migrate htcondor runner to use the python htcondor bindings

bgruening:
Advantages:
* using official bindings and an event logging mechanism if needed
* no hand-crafted parsing of condor job logs
* less io (I hope)
* avoid subprocess usage

Disadvantage:
* potentially many (as many jobs as running) open file handles
* the wheel is 30MB

The general question is if event logging or polling or a mixture is the best approach for us. More information can be found here: https://htcondor.readthedocs.io/en/latest/apis/python-bindings/tutorials/Scalable-Job-Tracking.html 

If the job is running longer than a day, we poll ones, to be save.

I appreciate any comments, as I'm not even sure the python bindings are any better than the current state.

### DIFF
--- a/lib/galaxy/jobs/runners/util/condor/__init__.py
+++ b/lib/galaxy/jobs/runners/util/condor/__init__.py
@@ -1,16 +1,6 @@
 """
 Condor helper utilities.
 """
-from subprocess import (
-    CalledProcessError,
-    check_call
-)
-
-from galaxy.util import (
-    commands,
-    unicodify
-)
-from ..external import parse_external_id
 
 DEFAULT_QUERY_CLASSAD = dict(
     universe='vanilla',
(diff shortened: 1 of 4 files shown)


bgruening:
@natefoo @jmchilton any comments are more than welcome :)

lib/galaxy/jobs/runners/util/condor/__init__.py
```diff
@@ -35,87 +25,24 @@ def submission_params(prefix=SUBMIT_PARAM_PREFIX, **kwds):
     return submission_params
 
 
-def build_submit_description(executable, output, error, user_log, query_params):
+def build_submit_description(submit_params, query_params):
```
jmchilton, line 28: Any chance you can write a second function to do this instead of replacing this one? It is used by Planemo also.

jmchilton:
Wheel size could be addressed a bit by conditionally loading the dependency - like Kubernetes does I believe.

If there are open questions and you're not sure this will be the right approach - why not just write a second runner? It seems like a second condor runner would let us test this more without any chance of breaking things (existing condor installs, pulsar tests, etc..).

jmchilton (approved):

hexylena:
What about mismatch between condor version mismatch between the wheel and the condor service -- it was in the past that you needed a python-htcondor that matched your htcondor version specifically. Is that still an issue?

mvdbeek:
I'll push this to 21.05, but feel free to merge if you're happy. I do like John's suggestion of a new runner, if you're not certain yet.

natefoo:
A conditional dependency for this should be quite simple, see [the one for pbs_python](https://github.com/galaxyproject/galaxy/blob/b8fec3ce1bb4519cb7e018dcc734dfc8b1b831dc/lib/galaxy/dependencies/__init__.py#L182). Also, like this, pbs_python is linked to the specific version of Torque, so for this you can simply let the wheel be built from source on install.

hexylena:
> no hand-crafted parsing of condor job logs

:heart_eyes: 

>  Also, like this, pbs_python is linked to the specific version of Torque, so for this you can simply let the wheel be built from source on install.

Oh, that's fantastic, I'd avoided the python bindings for just parsing the command output for so long just because sometimes it was annoying to have to get a very specific version. This alleviates that concern. 

:+1: if it's a new runner and we can have some time to test both? I'm using Condor in my new job, happy to test it out there.

jmchilton:
Pushing milestone - seems like there is a plan there we just need someone to implement it. If it gets done before the freeze we can just readjust the milestone.

bgruening:
This is now done in https://github.com/galaxyproject/galaxy/pull/21528

PR #10999, opened 2020-12-27
Only enable the workflow save button if it has changes.

jmchilton:
It makes this check internally anyway (https://github.com/galaxyproject/galaxy/blob/dev/client/src/components/Workflow/Editor/modules/services.js#L38) before firing the server request, so don't allow this click if it isn't going to do anything. I think it is helpful to understand if there are outstanding changes. I want to change the annotation stuff to come through editor Index.vue instead of firing its own ad-hoc request (ala #10987) so this button lighting up will make it super clear that an outstanding change has been made.

Also add a unit test for Workflow/Editor/Options.vue and add selectors to allow writing cleaner integration tests/tours.


mvdbeek:
Nice!

PR #11000, opened 2020-12-27
Check Python typing for galaxy-containers and galaxy-tool-util during CI.

jmchilton:
With fixes - including a much cleaner new way to add methods to namedtuples I hadn't known about.

### DIFF
--- a/lib/galaxy/containers/__init__.py
+++ b/lib/galaxy/containers/__init__.py
@@ -15,6 +15,7 @@ from abc import (
     abstractproperty
 )
 from collections import namedtuple
+from typing import Any, Dict, Optional, Type
 
 import yaml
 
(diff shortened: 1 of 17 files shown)


lib/galaxy/containers/__init__.py
```diff
@@ -166,13 +167,13 @@ def map_port(self, port):
 
 class ContainerInterface(metaclass=ABCMeta):
 
-    container_type = None
-    container_class = None
-    volume_class = None
-    conf_defaults = {
+    container_type: Optional[str] = None
+    container_class: Optional[Type[Container]] = None
```
mvdbeek, line 171: That's so helpful, I always struggle with what's expected here!

mvdbeek (approved):
Beautiful!

nsoranzo:
> With fixes - including a much cleaner new way to add methods to namedtuples I hadn't known about.

That's a much nicer syntax, can we replace all `collections.namedtuple`s with `typing.NamedTuple` ?

jmchilton:
> That's a much nicer syntax, can we replace all collections.namedtuples with typing.NamedTuple ?

Seems like a good plan, I'm not volunteering to do it but I'll hit the merge button if someone else does.

Issue #10989, opened 2020-12-24
Polish Workflow Best Practices Panel


jmchilton:
Tracking plans for second iteration of #10988

- [ ] Check for invalid references -  e.g referring to ``#{some_output}``, when some_output is not a tool output?
- [ ] Make sure parameters disappear from attributes after update.
- [ ] Rename "Parameters" in Attributes panel Legacy Parameters and add "fa-magic" button to bring to linting panel.
- [ ] Pre-catch more problems prior to extracting disconnected inputs and legacy parameters that would prevent refactoring.
  - [ ] mixed legacy on tool parameters
  - [ ] Update UI to tell why cannot fix (when canExtract is false)
- [ ] Add another warning to fixAll button - this may be a big change to your workflow. 
- [ ] Add preview of refactoring steps to fixAll button (starting to buildup a workflow diffing components based on refactoring actions).
- [ ] Display creator and license in green dialog as INFO in green panel.
- [ ] Allow embedded editing of creator and license.
- [ ] Display valid inputs and outputs as INFO in green panel.
- [x] Highlight inputs and outputs in the workflow editor on mouseover in linting panel. (https://github.com/galaxyproject/galaxy/pull/11025)
- [x] Button (or automatic) centering of the workflow editor on mouseover in linting panel. (https://github.com/galaxyproject/galaxy/pull/11025)
- [x] Fix annotation editing ala #10987 and uncomment from linting panel.
- [ ] Allow embedded editing of annotation.

jmchilton:
There is a bug that needs to be fixed before the release is finalized where if steps are added or removed before the workflow is saved - the order_index the panel wants to apply is wrong.

mvdbeek:
> There is a bug that needs to be fixed before the release is finalized where if steps are added or removed before the workflow is saved - the order_index the panel wants to apply is wrong.

I think that's fixed, but https://github.com/galaxyproject/galaxy/pull/11527/files#r586484668 still needs to be addressed ?

PR #11001, opened 2020-12-28
Simplify workflow editor metadata handling.

jmchilton:
A few different related changes here.
- Move responsibility for hitting the server to update annotation & name from Attributes.vue to Index.vue. This unifies & simplifies both the server interaction, user experience, and sets up addressing both #9166 & #10989
- Simplify backend related to license and creator in response to simpler frontend in #10987 (easy to simplify the API here since we haven't done a release with this yet).
- Do not create two workflow versions for each new name. We're exposing workflow versions so we should try to constrain the number of new versions we create - without these changes it is easy to create several versions by just renaming the workflow a couple times.
- Add selenium tests for rename workflows and updating annotations.

### DIFF
--- a/client/src/components/Workflow/Editor/Attributes.test.js
+++ b/client/src/components/Workflow/Editor/Attributes.test.js
@@ -14,24 +14,31 @@ Services.mockImplementation(() => {
     };
 });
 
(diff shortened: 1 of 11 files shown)


bgruening (approved):
Cool, thanks for the tests!

PR #11007, opened 2020-12-28
Type checking fixes for galaxy-data (datatypes and models).

jmchilton:

PR #11009, opened 2020-12-29
mypy type checking for more Galaxy packages

jmchilton:
Extends #11007 to include galaxy-auth, galaxy-job-execution, galaxy-web-stack, and galaxy-web-framework. The grids diff is rendering poorly but I simply moved the Grid definition until after GridColumnFilter and GridColumn so I could use those classes in the type annotation for ``standard_filters`` and ``columns``.

### DIFF
--- a/lib/galaxy/web_stack/message.py
+++ b/lib/galaxy/web_stack/message.py
@@ -47,7 +47,7 @@ class ApplicationStackMessageDispatcher:
 
 
 class ApplicationStackMessage(dict):
-    _target: Optional[str] = None
+    target: Optional[str] = None  # type: ignore
     default_handler = None
     _validate_kwargs = ('target',)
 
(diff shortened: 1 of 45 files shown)


lib/galaxy/web_stack/message.py
```diff
@@ -47,7 +47,7 @@ def dispatch(self, msg_str):
 
 
 class ApplicationStackMessage(dict):
-    _target: Optional[str] = None
+    target: Optional[str] = None  # type: ignore
```
bgruening, line 50: What is ignored here? It has a type annotation.

jmchilton: It conflicts with the ``@property`` defined below named ``target``. I tried to rename to the class attribute so it doesn't conflict with the ``@property`` but it broke the json encoding.

PR #11011, opened 2020-12-29
Update locale.js

shenweiyan:
Improve the new version of locale.js Chinese information.

### DIFF
--- a/client/src/nls/zh/locale.js
+++ b/client/src/nls/zh/locale.js
@@ -3,14 +3,14 @@
 define({
     // ----------------------------------------------------------------------------- masthead
     "Analyze Data": "数据分析",
-    "Workflow": "工作流程",
+    Workflow: "工作流程",
     "Shared Data": "数据共享",
     "Data Libraries": "数据库",
     "Published Histories": "历史",
     "Published Workflows": "流程",
     "Published Visualizations": "可视化",
     "Published Pages": "页面",
-    "Visualization": "可视化",
+    Visualization: "可视化",
     "New Track Browser": "创建 Track Browser",
     "Saved Visualizations": "已保存的可视化",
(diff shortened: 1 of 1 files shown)


bgruening:
@shenweiyan thanks a lot!

Do you think those improvements make sense?

65 "Copy": "复制",

67 "Extract Workflow": "提取工作流程",

162 "This dataset is currently uploading": "您的数据集正在上传中",

client/src/nls/zh/locale.js
```diff
@@ -180,54 +187,55 @@ define({
...
+    "Download local tool": false,
     "Tool lineage": false,
-    "Reload a tool's configuration": "重新加载工具的配置",
-    "Review tool migration stages": "查看工具迁移状态",
-    "View Tool Error Logs": "查看工具错误日志",
-    "Manage Allowlist": "管理显示白名单",
-    "Manage Tool Dependencies": "管理工具依赖",
-    Users: "用户",
-    Groups: "组别",
+    "Reload a tool's configuration": false,
```
bgruening, line 229: @shenweiyan is there a reason you removed this translation?

shenweiyan: As an administrators, I think the English interface will be better. And i18n processing is just to make it more convenient for non-administrators to use galaxy.

shenweiyan (commented):
> @shenweiyan thanks a lot!
> 
> Do you think those improvements make sense?
> 
> 65 "Copy": "复制",
> 
> 67 "Extract Workflow": "提取工作流程",
> 
> 162 "This dataset is currently uploading": "您的数据集正在上传中",

Thanks for the reminder, it should be make no sense.

mvdbeek:
Wonderful, this is always appreciated @shenweiyan!

PR #11015, opened 2020-12-30
Cleanup updating stored_workflow attributes in workflows manager.

jmchilton:
The ever-careless @jmchilton did the same thing in two slightly different ways between https://github.com/galaxyproject/galaxy/pull/10988 and https://github.com/galaxyproject/galaxy/pull/11001 - creating his own bizarre race condition.


bgruening:
That happens when McDonalds is closed ;)

PR #11016, opened 2020-12-30
More structured workflow create/update parameter handling

jmchilton:
Builds on #11015. 

See individual commits for details - but the big idea is to reduce duplication between different ways to create workflows and the one way to update one while also using structured, validated options instead of just passing ``**kwds`` around a bunch of different layers.

### DIFF
--- a/lib/galaxy/managers/workflows.py
+++ b/lib/galaxy/managers/workflows.py
@@ -3,6 +3,7 @@ import logging
 import os
 import uuid
 from collections import namedtuple
+from typing import Dict, Optional
 
 from gxformat2 import (
     from_galaxy_native,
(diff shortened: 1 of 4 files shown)


lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -708,7 +712,7 @@ def get_tool_predictions(self, trans, payload, **kwd):
     #
     # -- Helper methods --
     #
-    def __api_import_from_archive(self, trans, archive_data, source=None):
+    def __api_import_from_archive(self, trans, archive_data, source=None, payload={}):
```
mvdbeek:
```suggestion
    def __api_import_from_archive(self, trans, archive_data, source=None, payload=None):
        payload = payload or {}
```

lib/galaxy/managers/workflows.py
```diff
@@ -3,13 +3,15 @@
 import os
 import uuid
 from collections import namedtuple
+from typing import Dict, Optional
```
mvdbeek:
```suggestion
from typing import (
    Dict,
    Optional,
)
```

lib/galaxy/webapps/galaxy/controllers/workflow.py
```diff
@@ -14,7 +14,7 @@
     util,
     web
 )
-from galaxy.managers import workflows
+from galaxy.managers.workflows import MissingToolsException, WorkflowUpdateOptions
```
mvdbeek:
```suggestion
from galaxy.managers.workflows import (
    MissingToolsException,
    WorkflowUpdateOptions,
)
```

mvdbeek (commented):
That's a very nice improvement!

bgruening (approved):
Thanks @jmchilton

PR #11017, opened 2020-12-30
Expand image formats for viewing in OpenSeadragon

gregvonkuster:
@bgruening The OpenSeadragon viewer works with each of these formats, which seem to be the most likely to be used within Galaxy.  We'll eventually need a tool to convert these formats into the dzi (deep zoom image) format to get the real benefits of this viewer.  There are several tool candidates that can be wrapped for Galaxy, so eventually one will be added to the tools-iuc repo.

### DIFF
--- a/config/plugins/visualizations/openseadragon/templates/openseadragon.mako
+++ b/config/plugins/visualizations/openseadragon/templates/openseadragon.mako
@@ -31,7 +31,7 @@
             const xmlns = "http://schemas.microsoft.com/deepzoom/2008";
             const image_types = ["gif", "jpg", "png", "tiff"];
 
-            if (image_types.indexOf(hda_ext) >= 0) {
+            if (image_types.includes(hda_ext)) {
                 var viewer = OpenSeadragon({
                     id: viewer_id,
                     prefixUrl: prefixUrl,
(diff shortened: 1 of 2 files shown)


config/plugins/visualizations/openseadragon/templates/openseadragon.mako
```diff
@@ -29,8 +29,9 @@
             const prefixUrl = "//openseadragon.github.io/openseadragon/images/";
             const simple_image_url = "${h.url_for(controller='/datasets', action='index')}/" + hda_id + "/display";
             const xmlns = "http://schemas.microsoft.com/deepzoom/2008";
+            const image_types = ["gif", "jpg", "png", "tiff"];
 
-            if (hda_ext == 'jpg' || hda_ext == 'png') {
+            if (image_types.indexOf(hda_ext) >= 0) {
```
bgruening: Maybe you can use image_types.includes(hda_ext) ?

bgruening:
@gregvonkuster we need nii, stl, tck, trk and nrrds in Galaxy. Is this something you also need? We have gathered some test data already.

config/plugins/visualizations/openseadragon/templates/openseadragon.mako
gregvonkuster: Done here https://github.com/galaxyproject/galaxy/pull/11017/commits/8e41f40041a53092a90428f40bf840c6fa94243f

gregvonkuster:
> @gregvonkuster we need nii, stl, tck, trk and nrrds in Galaxy. Is this something you also need? We have gathered some test data already.

We're still working this out.  Some of the formats being looked at include ITK MetaImage (.mha/.raw, .mhd) and nii.  I would love to get some nii images from you if you have them.

PR #11020, opened 2020-12-31
Add nifti1 and nifti2 datatype

bgruening:


bgruening:
ping @gregvonkuster

gregvonkuster:
This is awesome!  Thanks so very much @bgruening :)

bgruening:
@mvdbeek or @nsoranzo do you know how to test binary datatypes with `siff_prefix`? I get now: `AttributeError: 'Nifti1' object has no attribute 'sniff'`.

lib/galaxy/datatypes/images.py
```diff
@@ -247,6 +250,100 @@ def sniff(self, filename):
...
+    """
+        Nifti2 format
+        https://brainder.org/2015/04/03/the-nifti-2-file-format/
+    """
+    file_ext = 'nii2'
+
+    def sniff_prefix(self, file_prefix):
+        """Determine if the file is in Nifti2 format.
+        >>> from galaxy.datatypes.sniff import get_test_fname
+        >>> fname = get_test_fname('avg152T1_LR_nifti2.nii.gz')
```
bgruening: Can it be that compressed files can bot be tested?

mvdbeek:
The test data extension must match the extension used in the datatype. `.nii` needs to be either `nii1` or `nii2`.

lib/galaxy/datatypes/images.py
mvdbeek: Compressed files can be tested, but then you need to either add a native datatype or register with auto-compressed types

bgruening:
@mvdbeek the uncompressed data (which is still binary) is too large. Can the extension then be `nii1.gz`? I would like o get those datatypes into 21.01 if possible.

mvdbeek:
The data in this PR ? That's just a couple of bytes if the github UI isn't lying, that's fine ? But yes, then you need to `auto_compressed_types="gz"` to the `<datatype ...>` tag

bgruening:
> The data in this PR ? 

Yes, but I used `head -c` to only extract the first useful bytes.
I also thought its useful to test the .gz extension. Or should be only test the uncompressed version?

> That's just a couple of bytes if the github UI isn't lying, that's fine ? But yes, then you need to auto_compressed_types="gz" to the <datatype ...> tag

It is added, that is what is strange. https://github.com/galaxyproject/galaxy/pull/11020/files#diff-b81a736d556204671b36855bda87e5f71454ff007d1f5d466f2c2dba0b765a1aR309

mvdbeek:
I would only test the uncompressed version, the rest is framework functionality that should be covered by other tests. I guess we're not looking at the auto-compressed extensions. If the sniffer works I think it's fine to just have a small dataset.

bgruening:
Ok, but then everything is as it should be. Maybe it was really just missing the correct test-file names. Let's see.

lib/galaxy/datatypes/proteomics.py
```diff
@@ -998,49 +998,3 @@ def generate_primary_file(self, dataset=None):
                 rval.append(f'<li><a href="{fn}" type="text/plain">{fn}</a>{opt_text}</li>')
         rval.append('</ul></div></html>')
         return "\n".join(rval)
-
-
-class Analyze75(Binary):
```
mvdbeek, line 1003: In theory TS datatypes may have imported from this location ... but I think I'm fine with breaking that.

mvdbeek (approved):

lib/galaxy/datatypes/proteomics.py
bgruening: True ... but ... oh :) I hope no one is ever doing it. I moved Analayze75 here as nii1 is supposed to be compatible with it...

bgruening:
Thanks @mvdbeek!

PR #11026, opened 2021-01-01
Document how parameters in sections are referenced

bernt-matthias:
Quite confusing for tool developers ... 

Maybe parts of the text could be duplicated / moved to other parts of the docs.

Is it a bug that cross referring inputs is only possible on the same or parent levels? From inspecting the code it also appeared to me that the reference needs to be before the referring parameter (in the xml file). The reason for both is in this function https://github.com/galaxyproject/galaxy/blob/91580d1912378dd709abb02ba647ccf7471503eb/lib/galaxy/tools/__init__.py#L1231

### DIFF
--- a/lib/galaxy/tool_util/xsd/galaxy.xsd
+++ b/lib/galaxy/tool_util/xsd/galaxy.xsd
@@ -2383,8 +2383,8 @@ The XML configuration is relatively trivial for sections:
 
 ```xml
 <inputs>
-    <section name="adv" title="Advanced Options" expanded="false">
-        <param name="plot_color" type="color" label="Track color" />
+    <section name="section_name" ... >
+        <param name="parameter_name" ... />
     </section>
 </inputs>
 ```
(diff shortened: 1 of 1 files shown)


lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -2393,9 +2393,25 @@ In your command template, you'll need to include the section name to access the
...
+In output filters sections are represented as dictionary with the same name as the section:
+
+```
+<filter>section_name['parameter_name']</filter>
+```
+
+In order to reference parameters in sections from tags in the `<outputs>` section, e.g. in the `format_source` attribute of `<data>` tags, the syntax is:
+
+```
+section_name|parameter_name
```
bgruening: can you give an example with format_source here?

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -2393,9 +2393,25 @@ In your command template, you'll need to include the section name to access the
...
+```
+
+Until profile 21.01 `parameter_name` was sufficient (https://github.com/galaxyproject/galaxy/pull/9493/files).
+
+Note that references to other parameters in the `<inputs>` section are only possible if the reference is in the same section or its parents, therefor only `parameter_name` is used.
```
bgruening: Can you provide an example here?

In theory we could have a parameter `foo` in the top-level inputs __and__ a parameter `foo` in a section. Do we?

bgruening (commented):
Thanks @bernt-matthias for improving our tool documentation constantly, this is really appreciated!

lib/galaxy/tool_util/xsd/galaxy.xsd
bernt-matthias: sure

lib/galaxy/tool_util/xsd/galaxy.xsd
bernt-matthias: > Can you provide an example here?

Yes. 

> In theory we could have a parameter foo in the top-level inputs and a parameter foo in a section. Do we?

Yes, but this is discouraged ... nevertheless it might occur in automatic tool building (e.g. in OpenMS I had the case, but luckily no cross reference was needed). Wondering if we should file an issue for this?

bernt-matthias:
Since this is the same for repeats and conditionals .. is there a way to have this docs also there? Also it would be good to link this to the docs of `format_source`, `metadata_source`, and `data_ref`, or?

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -2383,8 +2383,8 @@ The XML configuration is relatively trivial for sections:
 
 ```xml
 <inputs>
-    <section name="adv" title="Advanced Options" expanded="false">
-        <param name="plot_color" type="color" label="Track color" />
+    <section name="section_name" ... >
+        <param name="parameter_name" ... />
```
```suggestion
    <section name="section_name" title="Section Title" >
        <param name="parameter_name" type="text" label="A parameter  label" />
```
That is valid, should someone just copy-paste that.

mvdbeek (approved):

PR #11030, opened 2021-01-02
Upload test results as artifact in case of failure

nsoranzo:
Tested on https://github.com/common-workflow-language/galaxy/pull/47


bgruening (approved):
Cool, thanks @nsoranzo

Issue #11005, opened 2020-12-28
Polish Workflow Refactoring API


jmchilton:
Post merge of #10988 which introduced this API.

- [ ] Add new refactoring steps:
  - [x] ``{action: 'upgrade_subworkflow', step <StepReference>}`` https://github.com/galaxyproject/galaxy/pull/11032
  - [x] ``{action: 'upgrade_tool', step <StepReference>}`` https://github.com/galaxyproject/galaxy/pull/11032
  - [x] ``{action: 'fill_step_defaults', step: <StepReference>}`` https://github.com/galaxyproject/galaxy/pull/11032
  - [x] ``{action: 'fill_defaults'}`` https://github.com/galaxyproject/galaxy/pull/11032
  - [x] ``{action: 'update_output_label'}``  https://github.com/galaxyproject/galaxy/pull/11032 (not important, but helped with testing)
  - [ ] ``{action: 'restore_workflow_version'}``
- [x] Rework response to include a structured message for each input action. So things like disconnected inputs/outputs during subworkflow upgrades of tool state upgrades can be presented to the user and tracked. https://github.com/galaxyproject/galaxy/pull/11032
- [x] In ``dry_run=true``, capture the messages and do some extra checking that could predict database problems (specifically, I'm sure we could catch label problems at this step and we probably don't.))  https://github.com/galaxyproject/galaxy/pull/11032
- [x] ``dry_run=true`` should return the effective new state, so the UI can do stuff like ``upgrade_subworkflow`` or ``restore_workflow_version``, render the results and display the messages but still require a "Save" before the new version is committed.  https://github.com/galaxyproject/galaxy/pull/11032
- [ ] Store typed and structured actions and messages in the database along with the new workflow version created.
- [ ] Rework ``hasChanges`` tracking in ``Index.vue`` to track structured list of changes, use the refactoring API when possible.
- [ ] Eventually use refactoring API for all UI interactions that touch the ``Workflow`` object (something like editing the tags might be fine with the old API). (This would essentially implement #9166).

mvdbeek:
Do you think it'd be possible to also extend this to check for invalid references ? E.g referring to #{some_output}, when `some_output` is not a tool output?

jmchilton:
> Do you think it'd be possible to also extend this to check for invalid references ? E.g referring to #{some_output}, when some_output is not a tool output?

[...]

PR #11032, opened 2021-01-02
Implement edit & upgrade subworkflow buttons in workflow editor.

jmchilton:
... building on the enhancements to new workflow refactoring API. This PR is really about the refactoring API and improving its structure but those two buttons are big usability wins for people using subworkflows so I changed the title of this PR. 

Screen Shot 2021-01-04 at 12 35 13 AM

Screen Shot 2021-01-04 at 12 35 19 AM

Screen Shot 2021-01-04 at 12 36 41 AM

Workflow Refactoring API improvements:
- ~Builds on  #11016~ (merged)
- Includes https://github.com/galaxyproject/galaxy/pull/11041
- Fixes bug where the refactor API would create a new subworkflow for each update to an existing workflow.
- Fixes older bug from start of this release cycle where license wouldn't appear in export workflows.
- Implement five new refactoring actions ``fill_defaults``, ``fill_step_defaults``, ``upgrade_subworkflow``, ``upgrade_tool`` (this one is a bit beta but it isn't used in the GUI), ``update_output_label``. 
- All sorts of new tests to clarify test new actions and clarify how the API deals with missing steps, incomplete step state, etc...
- All sorts of new tests to clarify catching and reporting potential issues during subworkflow upgrades.
- Rework the workflow API response to include typed, structured messages in addition to a workflow. This structure is all important to record and store in the database with each version I think (this is outlined in #11005 and important for #9166. Four potential error types are captured:
   - ``tool_version_change``
   - ``tool_state_adjustment``
   - ``connection_drop_forced``
   - ``workflow_output_drop_forced``
- Comprehensive reworking of the ``dry_run`` option to the API. It now works and returns the workflow as it would be if it were modified according to the request. This includes a bunch of new tests that ensure the database isn't changing in response to ``dry_run`` requests.
- Use the new and improved ``dry_run`` option and the typed messages to report potential problems with refactorings in the UI and have the user confirm before up upgrading workflows.

### DIFF
--- a/lib/galaxy/managers/workflows.py
+++ b/lib/galaxy/managers/workflows.py
@@ -970,8 +970,10 @@ class WorkflowContentsManager(UsesAnnotations):
                         step_data_output['collection_type'] = collection_type
         return steps
 
-    def _workflow_to_dict_export(self, trans, stored=None, workflow=None):
+    def _workflow_to_dict_export(self, trans, stored=None, workflow=None, internal=False):
         """ Export the workflow contents to a dictionary ready for JSON-ification and export.
+
+        If internal, use content_ids instead subworkflow definitions.
         """
         annotation_str = ""
         tag_str = ""
@@ -1047,7 +1049,7 @@ class WorkflowContentsManager(UsesAnnotations):
                         action_arguments=pja.action_arguments)
                 step_dict['post_job_actions'] = pja_dict
 
-            if module.type == 'subworkflow':
+            if module.type == 'subworkflow' and not internal:
                 del step_dict['content_id']
                 del step_dict['errors']
                 del step_dict['tool_version']
@@ -1417,7 +1419,7 @@ class WorkflowContentsManager(UsesAnnotations):
         """Apply supplied actions to stored_workflow.latest_workflow to build a new version.
         """
         workflow = stored_workflow.latest_workflow
-        as_dict = self.workflow_to_dict(trans, stored_workflow, style="ga")
+        as_dict = self._workflow_to_dict_export(trans, stored_workflow, workflow=workflow, internal=True)
         raw_workflow_description = self.normalize_workflow_format(trans, as_dict)
         workflow_update_options = WorkflowUpdateOptions(
             fill_defaults=False,
--- a/lib/galaxy/managers/workflows.py
+++ b/lib/galaxy/managers/workflows.py
@@ -45,6 +45,7 @@ from galaxy.workflow.modules import (
     WorkflowModuleInjector
 )
 from galaxy.workflow.refactor.execute import WorkflowRefactorExecutor
+from galaxy.workflow.refactor.schema import RefactorActions
 from galaxy.workflow.reports import generate_report
 from galaxy.workflow.resources import get_resource_mapper_function
 from galaxy.workflow.steps import attach_ordered_steps
@@ -1408,7 +1409,7 @@ class WorkflowContentsManager(UsesAnnotations):
             if default_label and util.unicodify(default_label).lower() not in ['input dataset', 'input dataset collection']:
                 step.label = module.label = default_label
 
-    def refactor(self, trans, stored_workflow, refactor_request):
+    def do_refactor(self, trans, stored_workflow, refactor_request):
         """Apply supplied actions to stored_workflow.latest_workflow to build a new version.
         """
         workflow = stored_workflow.latest_workflow
@@ -1433,6 +1434,15 @@ class WorkflowContentsManager(UsesAnnotations):
                 raw_workflow_description,
                 workflow_update_options,
             )
+
+    def refactor(self, trans, stored_workflow, refactor_request):
+        stored_workflow, errors = self.do_refactor(trans, stored_workflow, refactor_request)
+        # TODO: handle errors...
+        return self.workflow_to_dict(trans, stored_workflow, style=refactor_request.style)
+
+
+class RefactorRequest(RefactorActions):
+    style: str = "export"
 
 
 class WorkflowStateResolutionOptions(BaseModel):
--- a/lib/galaxy/managers/workflows.py
+++ b/lib/galaxy/managers/workflows.py
@@ -411,7 +411,7 @@ class WorkflowContentsManager(UsesAnnotations):
             name=stored_workflow.name,
         )
 
-        if missing_tool_tups:
+        if missing_tool_tups and not workflow_update_options.allow_missing_tools:
             errors = []
             for missing_tool_tup in missing_tool_tups:
                 errors.append("Step %i: Requires tool '%s'." % (int(missing_tool_tup[3]) + 1, missing_tool_tup[0]))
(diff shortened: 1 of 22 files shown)


lib/galaxy/managers/workflows.py
```diff
@@ -403,23 +408,29 @@ def update_workflow_from_raw_description(self, trans, stored_workflow, raw_workf
...
             raise MissingToolsException(workflow, errors)
 
         # Connect up
-        workflow.stored_workflow = stored_workflow
-        stored_workflow.latest_workflow = workflow
+        if not dry_run:
+            workflow.stored_workflow = stored_workflow
+            stored_workflow.latest_workflow = workflow
+        else:
+            stored_workflow = model.StoredWorkflow()  # detached
```
mvdbeek, line 434: I don't think this is a good way to build detached objects (we are using the same trick for transient collections in the CWL branch, and they did actually get persisted) ... they will be attached to a session as soon as you add something coming from a session with a bidirectional relationship. I am also not confident that our session handling prevents a commit occurring elsewhere to affect this. I don't think it's a blocker since I didn't see this in my limited testing, but maybe a better strategy would be using a pydantic model here ?

lib/galaxy/webapps/galaxy/controllers/workflow.py
```diff
@@ -646,13 +646,20 @@ def delete(self, trans, id=None):
...
         an iframe (necessary for scrolling to work properly), which is
         rendered by `editor_canvas`.
         """
         if not id:
+            if workflow_id:
+                workflow = trans.sa_session.query(model.Workflow).get(trans.security.decode_id(workflow_id))
+                stored_workflow = workflow.stored_workflow
+                self.security_check(trans, stored_workflow, True, False)
+                stored_workflow_id = trans.security.encode_id(stored_workflow.id)
+                return trans.response.send_redirect(f'{url_for("/")}workflow/editor?id={stored_workflow_id}')
```
mvdbeek, line 661: This will always load up the latest version of the stored workflow ... is that what you intended ? I tried to fix something similar here: https://github.com/galaxyproject/galaxy/pull/10976/commits/481aecebd4ca7abac1271a848c8e0777e6c64482

lib/galaxy/workflow/refactor/schema.py
```diff
@@ -166,20 +167,53 @@ class UpdateReportAction(BaseAction):
...
+
+
+class FileDefaultsAction(BaseAction):
+    action_type: Literal['fill_defaults']
+
+
+class UpgradeSubworkflowAction(BaseAction):
+    action_type: Literal['upgrade_subworkflow']
+    step: step_reference_union
+    # should be decoded before stuffing it into the database...
```
mvdbeek: Is that a todo ?

lib/galaxy/workflow/refactor/schema.py
```diff
@@ -190,6 +224,47 @@ class UpdateReportAction(BaseAction):
...
+    tool_version_change = 'tool_version_change'
+    tool_state_adjustment = 'tool_state_adjustment'
+    connection_drop_forced = 'connection_drop_forced'
+    workflow_output_drop_forced = 'workflow_output_drop_forced'
+
+
+class RefactorActionExecutionMessage(BaseModel):
+    message: str
+    message_type: RefactorActionExecutionMessageTypeEnum
+    # messages don't have to be bound to a step, but if they are should
```
mvdbeek:
```suggestion
    # messages don't have to be bound to a step, but if they are they should
```

lib/galaxy/workflow/refactor/schema.py
```diff
@@ -190,6 +224,47 @@ class UpdateReportAction(BaseAction):
...
+    # previously connected input.
+    step_label: Optional[str]
+    order_index: Optional[int]
+
+    # messages don't have to be bound to a step with inputs, but if they are
+    # the input name should be specified here. Should be a prefixed name
+    # in the case of tool inputs (e.g. 'cond|repeat_0|input')
+    input_name: Optional[str]
+
+    # messages don't have to be bound to a step with inputs, but if they are
```
mvdbeek: I really like those comments. Can those become descriptions ?

mvdbeek (commented):
Very exciting, this all looks very solid. Unfortunately it failed upgrading my first subworkflow with:
Screenshot 2021-01-06 at 11 13 51

Digging through what might be going on now ...

mvdbeek:
So axios.put in the refactor function is returning a string instead of an object in the data attribute, this is what is failing ... but the response is proper JSON, the header is `application/json`, both in paste and uvicorn.

mvdbeek:
The problem is that `pydantic.json()` produces invalid JSON (integer dict keys). safe_dumps fixes that, so with https://github.com/galaxyproject/galaxy/commit/b3691da7f5d42d84a7f2ddf30683b7eee3a68bdc this is working beatifully.

mvdbeek:
I did first try orjson, which at least threw a TypeError, which is nice ... we may want to consider orjson in the future.

lib/galaxy/managers/workflows.py
jmchilton: I did extensive testing - I found all the places for these actions at least and there were some tricky ones. But in the dry run tests, I flushed the session, performed the actions, flush, and then made sure none of the workflow objects in the database were touched. If we do dry run tests for each new refactoring I think this is fine.

I really think the biggest problem is just the lib/galaxy/workflow/modules.py must injects a bunch of random state onto workflow steps that isn't defined at the model layer. If we had a clean module layer and stateless modules I don't think this would be so bad. Pydanic models that mirror the sqlite ones would be fine also - we probably should have a best practice for doing that.

lib/galaxy/webapps/galaxy/controllers/workflow.py
jmchilton: Yeah - the GUI says "load the subworkflow at the latest version" so that is what is happening here.

Loading an older version in the editor would be bad because you might save over more recent changes. I think having a read-only view of the workflow from the editor though would be awesome - so users can see what is happening in the subworkflow but that wasn't the intention here.

lib/galaxy/workflow/refactor/schema.py
jmchilton: It is a TODO when we start storing these actions in the database along with the new Workflow object they produce - but this PR doesn't dump this stuff into the database so it isn't a TODO we can act on right now. That ongoing goal in tracked in #11005.

jmchilton:
I'm happy to merge your safe_json fix... and I sort of understand it but I don't understand why the tests didn't catch it. Any clue how to exhibit it in the tests?

Reworking the models to use descriptions... is a good idea. I meant to get to it, I'll see if I can get that done tonight.

lib/galaxy/managers/workflows.py
jmchilton: xref https://github.com/galaxyproject/galaxy/pull/11032/files#diff-346d4ccd160634059af0bbdd140037c890f4974b57cf10ee1f88eff0d1576f2fR727

lib/galaxy/workflow/refactor/schema.py
jmchilton: I've revised the comment a bit to clarify this.

lib/galaxy/workflow/refactor/schema.py
jmchilton: Done? It is much harder to read inline now but they will be pretty in an OpenAPI schema someday.

mvdbeek: Thanks!

mvdbeek:
> I'm happy to merge your safe_json fix... and I sort of understand it but I don't understand why the tests didn't catch it. Any clue how to exhibit it in the tests?

You'd need to go from python dump to axios client, so I think only selenium or manual tests would have caught this.

mvdbeek:
Wow, this is such a an important piece of functionality ... writing the new release notes will be a fun exercise!

PR #11038, opened 2021-01-03
Remove migration logic that will never be used now.

jmchilton:
- Remove all logic from 0005_cleanup_datasets_fix.py. Noone is upgrading from a Galaxy running db version 0004 to 21.01.
- Remove transfer job stuff - I would be shocked if that ever did something useful on anything but main/test. At any rate it certainly won't do anything useful at this point.


bgruening:
@ic4f this might be interesting for you. Afaik you wanted to work on sqlalchemy migrations and updates, isn't it?

ic4f:
@bgruening Indeed! Thank you!
@jmchilton I'm working on moving us to alembic. I'll provide a way to auto-migrate if the db is under SQLAlchemy Migrate version control (otherwise it's manual upgrade). However, I think we can (and should) phase out SQLAlchemy Migrate completely by 21.05 (or 21.09?) So, I'm not opposed to this PR, especially if we need to remove that logic now, but if it's mainly for the sake of removing code that is no longer useful, then maybe we can wait and later remove it all? I suspect revision 0005 is not the only one that contains code that's no longer useful (there are also db objects, even on main, that are not used).

jmchilton:
It fails more aggressive type checking and every time we do big static checking passes and formatting passes through the code we have to touch so many useless files. We should remove everything useless as quickly as possible.

ic4f:
@jmchilton thank you for providing the details. Yes, it makes sense, of course, to remove it now.

jdavcs (approved):

jmchilton:
Thanks! And I'm excited to see what you come up with for alembic, let me know when you have some WIP.

PR #11036, opened 2021-01-03
Vue-ify Workflow State Upgrade Message Modal

jmchilton:
- Convert it to Vue and make message handling reactive - with some unit tests for modal state handlings.
- Expand the Selenium test to test out more state in one message.
- Improve the styling - more contrast between step reference and step details.
- Use step label instead of module name if available and workflow icon - to match the rest of workflow styling and provide more visual context.

Before and after based on Selenium screenshots.

Before:

workflow_editor_invalid_state-14

After:

workflow_editor_invalid_state-15

PR #11042, opened 2021-01-03
Vue-ify warnings about embedded subworkflows.

jmchilton:
Reusing workflow response parsing and generalizing Vue component from #11036.

Before:

Screen Shot 2021-01-03 at 11 24 39 AM

After:

Screen Shot 2021-01-03 at 11 28 55 AM

### DIFF
--- a/client/src/components/Workflow/Editor/Index.vue
+++ b/client/src/components/Workflow/Editor/Index.vue
@@ -1,6 +1,11 @@
 <template>
     <div id="columns" class="workflow-client">
         <StateUpgradeModal :stateMessages="stateMessages" />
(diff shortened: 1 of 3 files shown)


bgruening:
@jmchilton should this also include a test, specifially for subworkflows?

jmchilton:
Of course everything should have a test... 😅, there was no existing test coverage so I was hoping to sneak this in without adding some. Are you going to insist on it and if so were you thinking integration testing or unit tests?

bgruening:
;-) maybe its easier to add a test for this in https://github.com/galaxyproject/galaxy/pull/11032?

PR #11043, opened 2021-01-03
Add `negate` attribute to validators .. and tests for all

bernt-matthias:
- Adds a `negate` attribute to all validators. Can be helpful for example
  - for the regexp validator because it is sometimes difficult to implement a regexp, but easy to implement the inverse
  - for the inrange validators to express that a value needs to be outside of a range
- Adds tests to all validators (either unit/doc tests of functional tests -- the later for tools requiring data tables which seemed to be hard to implement as doc test)
- deprecate `MetadataInFileColumnValidator` (data tables should be used nowadays) and the attribute `line_startswith` that was only (properly) implemented in this validator
- deprecate the `*notin*` validators: the `negate` attribute should be used instead
- deprecate `substitute_value_in_message` which was only implemented for `ExpressionValidator`: `%s` in the message is now replaced by all validators.
- more useful default messages
- Remove some code duplication by subclassing
  - `InRangeValidator` as subclass of `ExpressionValidator`
  - `LengthValidator` as subclass of `InRangeValidator`
  - `MetadataInDataTableColumnValidator` as subclass of `ValueInDataTableColumnValidator`

TODOs:

- [x] document how to implement new validators
- [x] `Validator.validate` assert `bool` for `value`
- [x] more useful messages for `negate=True`
- [x] doc `substitute_value_in_message` and implement for all
- [x] `InRangeValidator` as subclass of `ExpressionValidator`
- [x] `LengthValidator` as subclass of `InRangeValidator`
- [x] `MetadataInDataTableColumnValidator` as subclass of `ValueInDataTableColumnValidator`
- [x] test validators for `multiple="true"` data parameters
  - already covered in https://github.com/galaxyproject/galaxy/blob/dev/test/functional/tools/validation_empty_dataset.xml
- [x] deprecate `MetadataInFileColumnValidator`
- [x] deprecate `line_startswith`
- [x] deprecate ``*NotIn*Validators``
- [x] `*InRange*` validators also for strings?
  - maybe later
- [ ] Validators that evaluate a dataset value often include  `hasattr(value, "metadata")` in the validation expression. This is odd/wrong in the negation and may be better here https://github.com/galaxyproject/galaxy/blob/2b95bfceed1177a48ada1423833b91cc601fdeb7/lib/galaxy/tools/parameters/basic.py#L1807 or we require tool developers to add a `MetadataValidator`
- [x] remove debug output
- [ ] adapt linting (deprectaed validators)?
- [x] Bug in functional tests see TODO in `validation_dataset_metadata_in_file`
  - I think this is a feature not a bug. Test files are probably referenced by their name in tests. If the same file is used 2x in the same test this leads to conflicts.

## How to test the changes?
(Select all options that apply)
- [x] I've included appropriate [automated tests](https://docs.galaxyproject.org/en/latest/dev/writing_tests.html).
- [x] This is a refactoring of components with existing test coverage.
- [ ] Instructions for manual testing are as follows:
  1. [add testing steps and prerequisites here if you didn't write automated tests covering all your changes]

## License
- [x] I agree to license these contributions under [Galaxy's current license](https://github.com/galaxyproject/galaxy/blob/dev/LICENSE.txt).
- [x] I agree to allow the Galaxy committers to license these and all my past contributions to the core galaxy codebase under the [MIT license](https://opensource.org/licenses/MIT). If this condition is an issue, uncheck and just let us know why with an e-mail to galaxy-committers@lists.galaxyproject.org.

Sometimes its difficult to implement a regexp, but easy to implement the inverse. 

I guess this might also be of interest for other validators (e.g. InRangeValidator, LengthValidator) and avoid the definition of things like ValueNotInDataTableColumnValidator and ValueInDataTableColumnValidator as I did in https://github.com/galaxyproject/galaxy/pull/7500

Ideas, suggestions?

TODO 

- [x] doc in xsd
- [x] test

### DIFF
--- a/lib/galaxy/tool_util/xsd/galaxy.xsd
+++ b/lib/galaxy/tool_util/xsd/galaxy.xsd
@@ -3643,10 +3643,13 @@ parameters a ``metadata`` validator is added automatically.
 - ``dataset_ok_validator``: Check if the data set is in state OK.
 - ``dataset_metadata_in_range``: Check if a numeric metadata value is within
 a given range.
-- ``dataset_metadata_in_file``: Check if a metadata value is contained in a
-specific column of another data set.
-- ``dataset_metadata_in_data_table`` (``dataset_metadata_not_in_data_table``):
-Check if a metadata value is contained in a column of a data table.
+- ``dataset_metadata_in_data_table``: Check if a metadata value is contained in a column of a data table.
+
+Deprecated data validators:
+
+- ``dataset_metadata_in_file``: Use data tables with ``dataset_metadata_in_data_table``.
+Check if a metadata value is contained in a specific column of another data set.
+- ``dataset_metadata_not_in_data_table``: Use ``dataset_metadata_in_data_table`` with ``negate="true"``.
 
 ### Validators for textual inputs (``text``, ``select``, ...)
 
@@ -3666,8 +3669,12 @@ For ``text`` inputs the following validators are useful:
 
 - ``length``: Check if the length of the value is within a range.
 - ``empty_field``: Check if the string is not empty
-- ``value_in_data_table`` (``value_not_in_data_table``): Check if the value is
+- ``value_in_data_table``: Check if the value is
 contained in a column of a given data table.
+
+Deprecated:
+
+- ``value_not_in_data_table``: Use ``value_in_data_table`` with ``negate="true"``.
 
 ### Validators for numeric inputs (``integer``, ``float``)
 
@@ -3725,15 +3732,16 @@ use in filenames may not contain ``..``.
       <xs:extension base="xs:string">
         <xs:attribute name="type" type="ValidatorType" use="required">
           <xs:annotation>
-            <xs:documentation xml:lang="en"><![CDATA[The list of supported
+            <xs:documentation xml:lang="en"><![CDATA[
+Valid values are: ``expression``, ``regex``, ``in_range``, ``length``,
+``metadata``, ``unspecified_build``, ``no_options``, ``empty_field``,
+``dataset_metadata_in_data_table``, ``value_in_data_table``,
+``dataset_ok_validator``, ``dataset_metadata_in_range``.
+Deprecated validators: ``dataset_metadata_in_file``, ``value_not_in_data_table``, ``dataset_metadata_not_in_data_table``.
+The list of supported
 validators is in the ``validator_types`` dictionary in
 [/lib/galaxy/tools/parameters/validation.py](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/parameters/validation.py).
-Valid values include: ``expression``, ``regex``, ``in_range``, ``length``,
-``metadata``, ``unspecified_build``, ``no_options``, ``empty_field``,
-``dataset_metadata_in_file``,
-``dataset_metadata_in_data_table``, ``dataset_metadata_not_in_data_table``,
-``value_in_data_table``, ``value_not_in_data_table``,
-``dataset_ok_validator``, ``dataset_metadata_in_range``]]></xs:documentation>
+]]></xs:documentation>
           </xs:annotation>
         </xs:attribute>
         <xs:attribute name="message" type="xs:string">
@@ -3785,13 +3793,6 @@ in ``dataset_metadata_in_data_table``, ``dataset_metadata_not_in_data_table``, `
 This can be an integer index to the column or a column name.</xs:documentation>
           </xs:annotation>
         </xs:attribute>
-        <xs:attribute name="line_startswith" type="xs:string">
-          <xs:annotation>
-            <xs:documentation xml:lang="en">Used to indicate lines in the file
-being used for validation start with a this attribute value.
-For use with validators of type ``dataset_metadata_in_file``, ``dataset_metadata_in_data_table``, ``dataset_metadata_not_in_data_table``, ``value_in_data_table``, ``value_not_in_data_tabl``</xs:documentation>
-          </xs:annotation>
-        </xs:attribute>
         <xs:attribute name="min" type="xs:decimal">
           <xs:annotation>
             <xs:documentation xml:lang="en">When the ``type`` attribute value is
@@ -3829,6 +3830,13 @@ not be modified.</xs:documentation>
             <xs:documentation xml:lang="en">Comma-seperated list of metadata
 fields to skip if type is ``metadata``. If not specified, all non-optional
 metadata fields will be checked unless ``check`` attribute is specified.</xs:documentation>
+          </xs:annotation>
+        </xs:attribute>
+        <xs:attribute name="line_startswith" type="xs:string">
+          <xs:annotation>
+            <xs:documentation xml:lang="en">Deprecated. Used to indicate lines in the file
+being used for validation start with a this attribute value.
+For use with validator ``dataset_metadata_in_file``</xs:documentation>
           </xs:annotation>
         </xs:attribute>
       </xs:extension>
--- a/lib/galaxy/tool_util/xsd/galaxy.xsd
+++ b/lib/galaxy/tool_util/xsd/galaxy.xsd
@@ -3644,13 +3644,13 @@ parameters a ``metadata`` validator is added automatically.
 - ``dataset_metadata_in_range``: Check if a numeric metadata value is within
 a given range.
 - ``dataset_metadata_in_data_table``: Check if a metadata value is contained in a column of a data table.
+- ``dataset_metadata_not_in_data_table``: Equivalent to ``dataset_metadata_in_data_table`` with ``negate="true"``.
 
 Deprecated data validators:
 
 - ``dataset_metadata_in_file``: Use data tables with ``dataset_metadata_in_data_table``.
 Check if a metadata value is contained in a specific column of a file in the ``tool_data_path``
 (which is set in Galaxy's config).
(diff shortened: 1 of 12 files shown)


mvdbeek:
Yeah, I like the negation, this is quite simple. If you add it to the xsd I'm happy to merge.

bernt-matthias:
I tried yesterday to implement this for all Validators .. but I struggle with the fact the they return `None` / raise an Exception. With a boolean return this seems much simpler.

bernt-matthias:
@mvdbeek .. do you think https://github.com/galaxyproject/galaxy/pull/11043/commits/6cdac7b395ca0d83bd46a2ad8bce2b4f8165d172 goes in a reasonable direction

- negate for all validators
- doctests for all validators (as far as I can get)

also still quite a few todos...

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -3725,21 +3732,28 @@ use in filenames may not contain ``..``.
...
-``value_in_data_table``, ``value_not_in_data_table``,
-``dataset_ok_validator``, ``dataset_metadata_in_range``]]></xs:documentation>
+]]></xs:documentation>
           </xs:annotation>
         </xs:attribute>
         <xs:attribute name="message" type="xs:string">
           <xs:annotation>
             <xs:documentation xml:lang="en">
-The error message displayed on the tool form if validation fails.</xs:documentation>
+The error message displayed on the tool form if validation fails. A placeholder string ``%s`` will be repaced by the ``value``</xs:documentation>
```
mvdbeek, line 3749: Do we have an example for that ? I'll have to admit I only know what `%s` and `value` are because I saw the docstring tests you added. Does this work for all parameter types, e.g collections and datasets ?

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -3825,6 +3832,18 @@ fields to skip if type is ``metadata``. If not specified, all non-optional
 metadata fields will be checked unless ``check`` attribute is specified.</xs:documentation>
           </xs:annotation>
         </xs:attribute>
+        <xs:attribute name="line_startswith" type="xs:string">
+          <xs:annotation>
+            <xs:documentation xml:lang="en">Deprecated. Used to indicate lines in the file
```
mvdbeek, line 3836: Why is this deprecated ?

lib/galaxy/tools/parameters/validation.py
```diff
@@ -216,176 +291,445 @@ class LengthValidator(Validator):
...
+    >>> # TODO in this light I wonder about the use of this validator at all....
+    >>> notok_hda.set_dataset_state(model.Dataset.states.EMPTY)
+    >>>
+    >>> p = ToolParameter.build(None, XML('''
+    ... <param name="blah" type="data">
+    ...     <validator type="dataset_ok_validator"/>
+    ... </param>
+    ... '''))
+    >>> t = p.validate(ok_hda)
+    >>> t = p.validate(notok_hda)
```
mvdbeek: That's a lot of code for a doctest, I think that would be better as a unit test. I also don't understand the TODO comments, can you eloborate on those ?

bernt-matthias: > That's a lot of code for a doctest, I think that would be better as a unit test.

True. But my experience with unit tests is still to limited to implement it. But, if there would be an example (in the best case for another validator) I might be able to do this.

> I also don't understand the TODO comments, can you eloborate on those ?

If I'm right, then the validators are called from here: https://github.com/galaxyproject/galaxy/blob/04885b4ae227396f49f5eb26a6bf9aca885edd1a/lib/galaxy/tools/parameters/basic.py#L1808, i.e. validators (including the  `dataset_ok_validator`) are only called for datasets that are `OK`. If this is correct, then this validator is useless.

lib/galaxy/tool_util/xsd/galaxy.xsd
bernt-matthias: Because it's only used in the `dataset_metadata_in_file` which is also deprecated in the PR (because data tables should be used).

lib/galaxy/tool_util/xsd/galaxy.xsd
bernt-matthias: > Do we have an example for that ?

Do you mean a functional tool test that we could link?

> I'll have to admit I only know what %s and value are because I saw the docstring tests you added. 

I guess the formulation can be improved.

> Does this work for all parameter types, e.g collections and datasets ?

Probably this would lead to undesired behavior. Maybe we can check the type of `value` is and print the name of value if its a dataset of a collection?

lib/galaxy/tool_util/xsd/galaxy.xsd
mvdbeek: This is not equivalent, this is described as `Check if a metadata value is contained in a
specific column of another data set`. It has its obvious problems, but data tables are not a replacement.

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -3666,9 +3669,13 @@ For ``text`` inputs the following validators are useful:
...
 - ``length``: Check if the length of the value is within a range.
 - ``empty_field``: Check if the string is not empty
-- ``value_in_data_table`` (``value_not_in_data_table``): Check if the value is
+- ``value_in_data_table``: Check if the value is
 contained in a column of a given data table.
 
+Deprecated:
+
+- ``value_not_in_data_table``: Use ``value_in_data_table`` with ``negate="true"``.
+
```
mvdbeek: I don't think one is better than the other, I don't think we should call this deprecated. We can internally replace the implementation if you think that makes the code more maintainable, but we don't want tool authors to update existing tools with such a change (and they may feel compelled to do that if we call it deprecated).

lib/galaxy/tool_util/xsd/galaxy.xsd
bernt-matthias: Then the doc is wrong in the xsd. The correct doc is in the class' doc string: `checks if the value for a dataset's metadata item exists in a file`. The validator gets a file name and then uses: https://github.com/galaxyproject/galaxy/blob/e99057b18ae4afde37c84602665ee563aa0ed555/lib/galaxy/tools/parameters/validation.py#L740 to construct the complete path.

bernt-matthias: But still you might be right that its not equivalent because in `MetadataInDataTableColumnValidator` we can not load arbitrary files, or?

I'm happy to remove marking these as deprecated .. but maybe we can at stress that data tables are better.

lib/galaxy/tool_util/xsd/galaxy.xsd
bernt-matthias: > I don't think one is better than the other, I don't think we should call this deprecated.

OK. Agreed. I will revert this.

bernt-matthias: > We can internally replace the implementation ...

This is already done :)

lib/galaxy/tools/parameters/validation.py
mvdbeek: Are you misreading the `if` condition maybe ? The only time the validators are not run is if the validator sets `requires_dataset_metadata` to true (which is not the case for `DatasetOkValidator`) and the dataset state is not OK.

lib/galaxy/tools/parameters/validation.py
```diff
@@ -216,176 +291,445 @@ class LengthValidator(Validator):
...
+    >>> hist = History()
+    >>> sa_session.add(hist)
+    >>> sa_session.flush()
+    >>> set_datatypes_registry(example_datatype_registry_for_sample())
+    >>> ok_hda = hist.add_dataset(HistoryDatasetAssociation(id=1, extension='interval', create_dataset=True, sa_session=sa_session))
+    >>> ok_hda.set_dataset_state(model.Dataset.states.OK)
+    >>> notok_hda = hist.add_dataset(HistoryDatasetAssociation(id=2, extension='interval', create_dataset=True, sa_session=sa_session))
+    >>> # TODO I do not get 100% why for state!=OK the validator is called
+    >>> # TODO because DataToolParameter.validate.do_validate calls the validator only of state=OK
+    >>> # TODO in this light I wonder about the use of this validator at all....
```
mvdbeek:
```suggestion
```

lib/galaxy/tool_util/xsd/galaxy.xsd
mvdbeek: I tend towards not deprecating them ... I think we encourage data tables elsewhere, so we could just leave it as is, but I don't mind stating it again.

lib/galaxy/tool_util/xsd/galaxy.xsd
mvdbeek: > Do you mean a functional tool test that we could link?

Yeah, that would work.

> Maybe we can check the type of `value` is and print the name of value if its a dataset of a collection?

That sounds good, but I'm a bit nervous about accidentally leaking out internals via the string substitution and binding ourselves so closely to python syntax within tool parameters. To some degree it feels a bit redundant to even substitute in the value ... the frontend knows which value raised the validation error. But I guess this might be useful when running workflows too ? ... @jmchilton do you have any thoughts on this ?

lib/galaxy/tools/parameters/validation.py
bernt-matthias: Got it. You are right.

PR #11045, opened 2021-01-04
Support complex markers in `run_tests.sh`

nsoranzo:
Needed for the WIP CWL conformance tests.

Also, create self-contained HTML reports.


jmchilton:
Awesome sauce - thanks so much!

PR #11046, opened 2021-01-04
Vue-ify Workflow Editor Modals

jmchilton:
While making these things Vue isn't a personal priority, I did some refactoring of message handling in #11032 and I wanted to clarify why I thought centralizing presentation logic in Index.vue was a good idea even though it is already quite massive.


mvdbeek (approved):
I guess this has the same integer key serialization issue that https://github.com/galaxyproject/galaxy/pull/11032/ has, otherwise this looks great.

PR #11047, opened 2021-01-04
add basic Apache Parquet datatype

martenson:
I wonder why there are so many ways to sniff the magic number in our code, but this seems to work in my testing.


jmchilton:
> I wonder why there are so many ways to sniff the magic number in our code, but this seems to work in my testing.

... and all of them keep reopening the file for each new sniff 🙈 - I think https://github.com/galaxyproject/galaxy/pull/11049 introduces better abstractions that should be more performant as well.

martenson:
updated to use new abstraction, thanks @jmchilton

lib/galaxy/datatypes/binary.py
```diff
@@ -2770,6 +2770,28 @@ def sniff(self, dataset):
...
+    Class describing Apache Parquet file (https://parquet.apache.org/)
+    >>> from galaxy.datatypes.sniff import get_test_fname
+    >>> fname = get_test_fname('example.parquet')
+    >>> Parquet().sniff(fname)
+    True
+    >>> fname = get_test_fname('test.mz5')
+    >>> Parquet().sniff(fname)
+    False
+    """
+    file_ext = "parquet"
```
nsoranzo, line 2785: Is there an EDAM datatype for it?

martenson: I had a brief look but it seems generic to me in all scopes. Do you think it fits somewhere?

martenson: afaik Parquet is a csv on steroids, I wouldn't know how to edam-tag csv either

nsoranzo: I couldn't find a specific one either. If you like, you can request it to be added at https://github.com/edamontology/edamontology/issues

nsoranzo (approved):

lib/galaxy/datatypes/binary.py
martenson: If we succeed to enough extent to apply parquet in the mass spectrometry pipelines we'll consider PRing it to EDAM, at the moment I think it is a bit premature.

PR #11049, opened 2021-01-04
Optimized binary datatype sniffing for various binary datatypes.

jmchilton:
Re-opening the file for each new test is bad - the sniff prefix exists and has read in the header already. Just changed all these datatypes to re-use that existing header information to avoid that repeated re-opening of the same file.

### DIFF
--- a/lib/galaxy/datatypes/binary.py
+++ b/lib/galaxy/datatypes/binary.py
@@ -19,7 +19,7 @@ from typing import Optional
 import h5py
 import pysam
 import pysam.bcftools
-from bx.seq.twobit import TWOBIT_MAGIC_NUMBER, TWOBIT_MAGIC_NUMBER_SWAP, TWOBIT_MAGIC_SIZE
+from bx.seq.twobit import TWOBIT_MAGIC_NUMBER, TWOBIT_MAGIC_NUMBER_SWAP
 
 from galaxy import util
 from galaxy.datatypes import metadata
@@ -28,6 +28,7 @@ from galaxy.datatypes.data import (
     get_file_peek,
 )
 from galaxy.datatypes.metadata import DictParameter, ListParameter, MetadataElement, MetadataParameter
+from galaxy.datatypes.sniff import build_sniff_from_prefix
 from galaxy.util import nice_size, sqlite
 from galaxy.util.checkers import is_bz2, is_gzip
 from . import data, dataproviders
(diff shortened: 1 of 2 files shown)


lib/galaxy/datatypes/binary.py
```diff
@@ -2730,12 +2697,12 @@ def __init__(self, **kwd):
         super().__init__(**kwd)
         self._magic = binascii.unhexlify("6be33e6d47530e3c")
 
-    def sniff(self, filename):
+    def sniff_prefix(self, sniff_prefix):
         # The first 8 bytes of any daa file are 0x3c0e53476d3ee36b
```
martenson, line 2701: Not related to the PR but why is this the same as unhexlify input, but reversed?

jmchilton: I don't know - it does seem odd.

lib/galaxy/datatypes/sniff.py
```diff
@@ -597,6 +598,21 @@ def search(self, pattern):
...
+        size = struct.calcsize(pattern)
+        header_bytes = self.contents_header_bytes[0:size]
+        if len(header_bytes) < size:
+            return None
+        return struct.unpack(pattern, header_bytes)
+
+    def magic_header(self, pattern):
+        # unpack header and get first element
+        unpacked = self.unpack_header(pattern)
+        return None if not unpacked else unpacked[0]
```
nsoranzo:
```suggestion
    def magic_header(self, pattern):
        """
        Unpack header and get first element
        """
        size = struct.calcsize(pattern)
        header_bytes = self.contents_header_bytes[:size]
        if len(header_bytes) < size:
            return None
        return struct.unpack(pattern, header_bytes)[0]
```

lib/galaxy/datatypes/sniff.py
```diff
@@ -597,6 +598,21 @@ def search(self, pattern):
...
+
+    def startswith_bytes(self, test_bytes):
+        return self.contents_header_bytes[0:len(test_bytes)] == test_bytes
```
nsoranzo:
```suggestion
        return self.contents_header_bytes.startswith(test_bytes)
```

lib/galaxy/datatypes/sniff.py
martenson: Comment indent looks incorrect in github preview.

lib/galaxy/datatypes/binary.py
hexylena: little/big endian maybe?

lib/galaxy/datatypes/sniff.py
nsoranzo: Thanks @martenson , fixed.

lib/galaxy/datatypes/binary.py
martenson: blame Gulliver

bgruening (approved):

PR #11054, opened 2021-01-05
Revise user experience for history import and export.

jmchilton:
- Revised history import and export, all in Vue and with high-level unit tests for both.
- Works with new and old history panel.
- Enables importing and exporting to remote files (``galaxy.files`` plugins) - hopefully importing and exporting to FTP is more stable, should enable exporting histories on AnVIL.
- Substantially more input, better error handling than previous implementations. Access to job information to debug errors and track progress, better descriptions or links generated, links generated with a JEHA id encoded in them so they are much more robust than the links that required history to not be updated after last export.
- Reduced legacy controller code in lieu of API code.
- Icon to copy the export URL to clipboard.

## Some screenshots

For existing functionality of exporting to link, it now shows out of data exports that can be used as is or allows for generation of another export.

Screen Shot 2021-01-06 at 4 13 08 PM

Slightly different text for updated history exports. This was generated by clicking that link previously and waiting for the job to complete.

Screen Shot 2021-01-06 at 4 43 10 PM

Information about when the export was generated and other job info useful for debugging is reachable from the link.

Screen Shot 2021-01-06 at 4 43 24 PM

Re-using the copy functionality with the the link icon requested by @afgane.

Screen Shot 2021-01-06 at 4 43 32 PM

Can now export to "Remote Files" (galaxy.files plugins - including FTP and user's library directory if they are configured as writable).

Screen Shot 2021-01-06 at 4 43 46 PM

On export, job is monitored.

Screen Shot 2021-01-06 at 4 43 52 PM

Revised the language here to be more optimistic as suggested by @afgane.

Screen Shot 2021-01-06 at 4 44 01 PM

Also re-did history import panel. Three different options can be selected via radio buttons (previously both conflicting options were just shown).

Screen Shot 2021-01-06 at 4 44 47 PM

File selection dialog can now be used, searching works:

Screen Shot 2021-01-06 at 4 45 07 PM

Selected files appear as follows.

Screen Shot 2021-01-06 at 4 45 17 PM

Completed import looks like this:

Screen Shot 2021-01-06 at 4 45 28 PM

Problematic history import provides feedback now (nothing would be shown before - you'd just be sent back to the histories when it started and you'd have to hope eventually your new history would appear). You can see the standard error (initially collapsed but expandable as shown below.

Screen Shot 2021-01-06 at 4 45 58 PM

You can also click to view more information on the failed job if the import fails.

Screen Shot 2021-01-06 at 4 46 04 PM

### DIFF
--- a/client/src/components/HistoryExport/ExportLink.vue
+++ b/client/src/components/HistoryExport/ExportLink.vue
@@ -1,0 +1,61 @@
+<template>
+    <span>
+        <b
+            ><a :href="link">{{ link }}</a></b
+        >
+        <font-awesome-icon
+            v-b-tooltip.hover
+            title="Copy export URL to your clipboard"
+            icon="link"
+            style="cursor: pointer;"
+            @click="copyUrl"
+        />
+        <i
+            title="Information about when the history export was generated is included in the job details. Additionally, if there are issues with export, the job details may help figure out the underlying problem or communicate issues to your Galaxy administrator."
+        >
+            (<a href="#" @click="showDetails">view job details</a>)
+        </i>
+        <b-modal v-model="details" scrollable ok-only>
+            <job-information :job_id="historyExport.job_id" :include-times="true" />
+        </b-modal>
+    </span>
+</template>
+
+<script>
+import { FontAwesomeIcon } from "@fortawesome/vue-fontawesome";
+import { library } from "@fortawesome/fontawesome-svg-core";
+import { faLink } from "@fortawesome/free-solid-svg-icons";
+import { BModal } from "bootstrap-vue";
+import { copy } from "utils/clipboard";
+import JobInformation from "components/JobInformation/JobInformation.vue";
+
+library.add(faLink);
+
+export default {
+    components: { BModal, JobInformation, FontAwesomeIcon },
+    props: {
+        historyExport: {
+            type: Object,
+            required: true,
+        },
+    },
+    data() {
+        return {
+            details: false,
+        };
+    },
+    computed: {
+        link() {
+            return this.historyExport.external_download_permanent_url;
+        },
+    },
+    methods: {
+        showDetails() {
+            this.details = true;
+        },
+        copyUrl() {
+            copy(this.latestExportUrl, "Export URL copied to your clipboard");
+        },
+    },
+};
+</script>
--- a/client/src/components/HistoryExport/ToLink.vue
+++ b/client/src/components/HistoryExport/ToLink.vue
@@ -1,0 +1,166 @@
+<template>
+    <div>
+        <b-alert show variant="danger" v-if="errorMessage" dismissible @dismissed="errorMessage = null">
+            {{ errorMessage }}
+            <JobError
+                v-if="jobError"
+                style="margin-top: 15px;"
+                header="History export job ended in error"
+                :job="jobError"
+            />
+        </b-alert>
+        <div v-if="loadingExports">
+            <loading-span message="Loading history export information from Galaxy server." />
+        </div>
+        <div v-else-if="waitingOnJob">
+            <loading-span message="Galaxy server is preparing history for download, this will likely take a while." />
+        </div>
+        <div v-else-if="latestExportReady">
+            Link for download ready
+            <export-link :historyExport="latestExport" />
+            . Use this link to download the archive or import it on another Galaxy server.
+        </div>
+        <div v-else-if="hasReadyExport">
+            <p>An out of date export is ready <export-link :historyExport="latestReadyExport" />.</p>
+
+            <p>
+                The history has changed since this export was generated,
+                <a class="export-link" href="#" @click="regenerateExport"
+                    >click here to generate a new archive for the current history state</a
+                >.
+            </p>
+        </div>
+        <div v-else>
+            <p>No link for history export ready, {{ whyNoLink }}.</p>
+            <p>
+                <b
+                    ><a class="export-link" href="#" @click="regenerateExport"
+                        >Click here to to generate a new archive for this history.</a
+                    ></b
+                >
+            </p>
+        </div>
+    </div>
+</template>
+
+<script>
+import { getAppRoot } from "onload/loadConfig";
+import axios from "axios";
+import Vue from "vue";
+import BootstrapVue from "bootstrap-vue";
+import { errorMessageAsString } from "utils/simple-error";
+import LoadingSpan from "components/LoadingSpan";
+import ExportLink from "./ExportLink.vue";
+import { waitOnJob } from "components/JobStates/wait";
+import JobError from "components/JobInformation/JobError";
+
+Vue.use(BootstrapVue);
+
+export default {
+    components: { LoadingSpan, ExportLink, JobError },
+    props: {
+        historyId: {
+            type: String,
+            required: true,
+        },
+    },
+    data() {
+        return {
+            errorMessage: null,
+            exports: null,
+            waitingOnJob: false,
+            jobError: null,
+        };
+    },
+    created() {
+        this.loadExports();
+    },
+    computed: {
+        loadingExports() {
+            return this.exports === null;
+        },
+        hasExports() {
+            return this.exports !== null && this.exports.length > 0;
+        },
+        latestExport() {
+            return this.exports && this.exports[0];
+        },
+        readyExports() {
+            return (this.exports || []).filter((e) => e.ready);
+        },
+        hasReadyExport() {
+            return this.latestReadyExport !== null;
+        },
+        latestReadyExport() {
+            return this.readyExports.length > 0 ? this.readyExports[0] : null;
+        },
+        latestExportReady() {
+            return this.latestExport?.ready && this.latestExport?.up_to_date;
+        },
+        latestExportPreparing() {
+            return this.latestExport?.preparing;
+        },
+        whyNoLink() {
+            if (!this.hasExports) {
+                return `no history export ever initiated for this history`;
+            } else {
+                return `previous history export archival likely failed`;
+            }
+        },
+    },
+    methods: {
+        loadExports() {
+            const url = `${getAppRoot()}api/histories/${this.historyId}/exports`;
+            axios
+                .get(url)
+                .then((response) => {
+                    this.exports = response.data;
+                    if (this.latestExportPreparing) {
+                        this.waitOnExportJob(this.latestExport.job_id);
+                    } else {
+                        this.waitingOnJob = false;
+                    }
+                })
+                .catch(this.handleError);
+        },
+        regenerateExport() {
+            this.waitingOnJob = true;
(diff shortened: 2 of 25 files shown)


client/src/components/HistoryExport/ToLink.vue
```diff
@@ -0,0 +1,148 @@
...
+        </div>
+        <div v-else-if="polling">
+            <loading-span message="Galaxy server is preparing history for download, this will likely take a while." />
+        </div>
+        <div v-else>
+            <p>No link for history export ready, {{ whyNoLink }}.</p>
+            <p>
+                <b
+                    ><a class="export-link" href="#" @click="putExportUntilReady"
+                        >Click here to attempt to generate a new archive for this history.</a
```
afgane: Why "_attempt_ to generate"?

client/src/components/HistoryExport/ToLink.vue
```diff
@@ -0,0 +1,148 @@
...
+            return this.latestExport?.external_download_url;
+        },
+        latestExportPreparing() {
+            return this.latestExport?.preparing;
+        },
+        whyNoLink() {
+            if (!this.hasExports) {
+                return `no history export ever initiated for this history`;
+            } else if (!this.latestExport.up_to_date) {
+                return `previous history export not up-to-date (history has changed since export was generated)`;
```
afgane: Is the list of earlier exports available or kept? Listing those here sorted by date & time would be great. If not, explicitly saying that any earlier export if no longer available would be good for clarity.

client/src/components/HistoryExport/ToLink.vue
```diff
@@ -0,0 +1,148 @@
...
+        </div>
+        <div v-else-if="latestExportReady">
+            Link for download ready
+            <b
+                ><a :href="latestExportUrl">{{ latestExportUrl }}</a></b
+            >
+            <font-awesome-icon
+                v-b-tooltip.hover
+                title="Copy export URL to your clipboard"
+                icon="copy"
```
afgane: We should be consistent with our _copy to clipboard_ icon and use `fa-chain`, per https://github.com/galaxyproject/galaxy/pull/10529

(Even though I personally dislike the chain link icon as an indication of copy but I'm late for that discussion.)

client/src/components/HistoryExport/ToRemoteFile.vue
```diff
@@ -0,0 +1,122 @@
...
+        <b-alert show variant="warning" v-if="errorMessage">
+            {{ errorMessage }}
+        </b-alert>
+        <div v-if="waitingOnJob">
+            <loading-span message="Executing history export job, this will likely take a while." />
+        </div>
+        <div v-else-if="jobComplete">
+            <b-alert show variant="success" dismissible @dismissed="reset">
+                <h4>Done!</h4>
+                <p>History appears to have been successfully exported.</p>
```
afgane: _appears_ does not breed with confidence from the user's perspective. As a user, I'd like a _yes/no_ outcome, not a _maybe_ outcome...

afgane (commented):
Will this work with any plugin defined in Galaxy (eg, S3, fa-anvil) or do are changes required to those plugins?

A couple of minor wording comments but this is excellent.

client/src/components/HistoryExport/ToRemoteFile.vue
jmchilton: I mean... have you ever gotten it to work 😆. When does this work? Who is confident in history imports and exports?

I'll change the language though

client/src/components/HistoryExport/ToLink.vue
jmchilton: I mean I copied and pasted this code from ToolFooter.vue - so we're already inconsistent. I'll change it to fa-chain though

client/src/components/HistoryExport/ToLink.vue
jmchilton: Yeah - it is unexposed here - this seems like a step forward but I think we actually can produce links for older exports that have been generated. I was imagining like a "See Older Exports" button that would pop up a table. We could show the exports, job lists, etc...

jmchilton:
> Will this work with any plugin defined in Galaxy (eg, S3, fa-anvil) or do are changes required to those plugins?

I believe so, they seem to implement the _write_from on the interface, it is the same method that I use for the posix stuff.

https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/files/sources/_pyfilesystem2.py#L45

client/src/components/HistoryExport/ToRemoteFile.vue
jmchilton: Changed.

client/src/components/HistoryExport/ToLink.vue
jmchilton: I modified the API response so I can get the information needed from those older exports. It now shows the latest available history export that is ready and I verified the older links seem to work. Info about when it was generated is available through the job information - which I've included a screenshot of. It is an extra click but it is all now information that is available.

As for the even older downloads - are those useful? I don't think anyone would want them right - they are not named or tagged in any interested way. If you want, I'd suggest we put it on the roadmap for Q2 rather then baking it into this PR which is a substantial improvement.

jmchilton:
Given all the questions around my skepticism that these links work consistently - I've taken another pass at this and folded history import in as well. There is now a lot more feedback about what is happening and information about the jobs are now all navigable by the user. This should help debug problems that occur, help convey issues to admins, help them understand when the import or export process is still running and therefore why no results are available. All of these were issues before I think - and substantial ones.

client/src/components/HistoryExport/ExportLink.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+    <span>
+        <b
+            ><a :href="link">{{ link }}</a></b
+        >
```
nuwang, line 5: Minor issue, but the tag formatting seems inconsistent?

jmchilton: I don't do formatting, this is all prettier's doing I think.

client/src/components/JobStates/wait.js
```diff
@@ -0,0 +1,29 @@
...
+                const state = jobResponse.data.state;
+                if (onStateUpdate !== null) {
+                    onStateUpdate(state);
+                }
+                if (JOB_STATES_MODEL.NON_TERMINAL_STATES.indexOf(state) !== -1) {
+                    setTimeout(checkCondition, interval, resolve, reject);
+                } else if (JOB_STATES_MODEL.ERROR_STATES.indexOf(state) !== -1) {
+                    // grab full output to include stderr and
+                    // such when generating error messages.
+                    axios.get(`${jobUrl}?full=true`).then(reject).catch(reject);
```
mvdbeek: I think you could simplify this by always asking for the full job output until the job is terminal.
Those few extra fields shouldn't have a performance impact while the job is non-terminal.

lib/galaxy/webapps/galaxy/api/histories.py
```diff
@@ -496,13 +508,16 @@ def archive_export(self, trans, id, **kwds):
...
-            return dict(download_url=url_for("history_archive_download", id=id, jeha_id=jeha_id))
+            return self.history_export_view.serialize(trans, id, jeha)
         else:
             # Valid request, just resource is not ready yet.
             trans.response.status = "202 Accepted"
-            return ''
+            if jeha:
+                return self.history_export_view.serialize(trans, id, jeha)
+            else:
+                return ''
```
mvdbeek: That's what's being returned on the first click when exporting a history to link, while the UI is waiting for a job id (`this.waitOnExportJob(response.data.job_id)`). That fails, so then there is just a console error and the view never updates.

I think we can return the encoded job id as on line 510, that shouldn't break older code that polls for a download link

client/src/components/HistoryExport/ToRemoteFile.vue
```diff
@@ -0,0 +1,119 @@
...
+                class="mt-3"
+            >
+                <files-input v-model="directory" mode="directory" :requireWritable="true" />
+            </b-form-group>
+            <b-form-group id="name" label-for="name" description="Give the exported file a name." class="mt-3">
+                <b-form-input id="name" v-model="name" placeholder="Name" required></b-form-input>
+            </b-form-group>
+            <b-row align-h="end">
+                <b-col
+                    ><b-button class="export-button" variant="primary" @click="doExport" :disabled="!canExport"
```
mvdbeek, line 35: I think it's a bit confusing that the button is just grey and you can't select any directories if no file sources are configured. I don't mind getting this in as a first version though, but maybe the `FilesInput ` component can make it clear that no sources are configured ? In the future

mvdbeek (commented):
Except for the small hiccup with export to link this is looking great!

lib/galaxy/webapps/galaxy/api/histories.py
jmchilton: This was a last minute change to get the tests working again, I didn't revisit the UI - sorry about that. I'll try to resolve this.

client/src/components/HistoryExport/ToRemoteFile.vue
jmchilton: Revised this again, both import and export now load the available file source plugins. If there are not writable file sources, the tabs disappear and only the link option is available:

Screen Shot 2021-01-08 at 4 32 59 PM

The import option likewise doesn't show that last radio button if there are not available file source plugins.

lib/galaxy/webapps/galaxy/api/histories.py
jmchilton: I've made this change!

client/src/components/JobStates/wait.js
jmchilton: I've made this change!

client/src/components/JobInformation/JobError.vue
```diff
@@ -0,0 +1,70 @@
...
+                <a href="#">See full job details <font-awesome-icon icon="info-circle" /></a>
+            </div>
+            <div class="error-wrapper" v-if="job.stderr" @click="toggleExpanded">
+                Job Standard Error
+                <pre :class="errorClasses">{{ job.stderr }}</pre>
+                <b v-if="!expanded">(Click to expand job standard error)</b>
+            </div>
+            <!-- TODO: modal for reporting error. -->
+        </b-card-text>
+        <b-modal v-model="showInfo" scrollable ok-only>
```
mvdbeek, line 14: b-modal leaves space for a title, we should maybe optionally disable the title in the JobInformation component and set a title here instead. That would probably look better than this:
Screenshot 2021-01-12 at 12 15 59

mvdbeek:
I also noticed that imported histories show up in the multi-history view, but not in the old histories view ... doesn't feel related to this PR, but that is bad UX.

jmchilton:
> imported histories show up in the multi-history view, but not in the old histories view

😱 that seems terrible. I've added that to my bugs to look into during this release stabilization.

PR #11055, opened 2021-01-05
Move framework tests to GitHub workflows + lint test tools

nsoranzo:
Test tool linting (not all of them yet):
- added CDATA to command and help
- single-quote variables for text and dataset params and outputs in Cheetah
- add tool version
- XML element and attribute order per IUC standard
- fixed indentation

Also:
- Fix API tests failing when usegalaxy.org is down

### DIFF
--- a/.github/workflows/framework.yaml
+++ b/.github/workflows/framework.yaml
@@ -1,0 +1,40 @@
+name: Framework tests
+on: [push, pull_request]
+env:
+  GALAXY_TEST_DBURI: 'postgres://postgres:postgres@localhost:5432/galaxy?client_encoding=utf8'
+jobs:
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.7']
+    services:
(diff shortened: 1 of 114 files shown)


jmchilton:
I just reviewed these and they look fine, please just be careful not to fix every bad issue since I believe some of these tools are explicitly testing older, legacy behavior.

nsoranzo:
> I just reviewed these and they look fine, please just be careful not to fix every bad issue since I believe some of these tools are explicitly testing older, legacy behavior.

I'm indeed trying to be careful, and hopefully tests should break when I'm not :) . I'm adding clarifying comments to places where I found such cases.

Thanks for the review!

nsoranzo:
I've disabled the framework tests on Jenkins for the dev and release_21.* branches, which were the last ones we were running on Jenkins for dev. Thanks Jenkins server at PSU, @natefoo and @martenson for the several years of support!

P.S. The Jenkins PSU server is still in use to test Galaxy old releases and supports other important services.

Issue #10889, opened 2020-12-10
Progressively adapt old API routes to fastAPI


mvdbeek:
This is not high-priority (the old routes aren't broken and will continue to work), but it would be great if we can get as many people as possible to get their hands on this task.

An example PR to add fastAPI routes is here: https://github.com/galaxyproject/galaxy/pull/11056
We are aiming at following the [fastAPI documentation](https://fastapi.tiangolo.com/tutorial/) closely.
We will try to keep the routes as simple as possible, pushing complex logic into managers as much as possible.
Schema models may either live next to the manager code, or in lib/galaxy/schema/ for models that are used in
different parts of the app. Schema models for database models should always go to lib/galaxy/schema.

If you have any questions, please ask here or on [gitter](https://gitter.im/galaxyproject/dev) 

List of files in lib/galaxy/webapps/galaxy/api/ : 
*(To prevent duplication of effort, please put down your GitHub username next to the controller you're working on)*

- [ ] annotations.py
- [ ] authenticate.py
- [ ] cloudauthz.py
- [x] cloud.py #16296
- [x] configuration.py #11126
- [ ] container_resolution.py
- [x] dataset_collections.py #12781
- [x] datasets.py #12924
- [x] datatypes.py #11098
- [x] display_applications.py #16156
- [ ] dynamic_tools.py
- [ ] extended_metadata.py
- [x] folder_contents.py #14223
- [x] folders.py  #12131
- [ ] forms.py
- [x] genomes.py #11241
- [x] group_roles.py #12337
- [x] groups.py #17051
- [x] group_users.py #13648
- [x] histories.py #12386
- [x] history_contents.py #12578
- [x] item_tags.py #17064
- [ ] job_files.py
- [x] job_lock.py #10823
- [ ] job_ports.py
- [ ] jobs.py @mvdbeek
- [x] libraries.py #11983
- [x] library_contents.py #18838
- [ ] library_datasets.py
- [x] licenses.py #11056 
- [x] metrics.py #11568
- [ ] page_revisions.py
- [x] pages.py #11262
- [ ] plugins.py
- [ ] provenance.py
- [x] quotas.py #11315
- [x] remote_files.py #11211 
- [x] roles.py #10823
- [ ] sanitize_allow.py
- [ ] search.py
- [x] tags.py  #11224
- [x] tool_data.py #11167
- [ ] tool_dependencies.py @ic4f 
- [ ] tool_entry_points.py
[...]

PR #11056, opened 2021-01-05
Add FastAPI migration of licenses endpoints

davelopez:
Related to #10889
I decided to migrate the easiest API endpoints I found as a FastAPI exercise (and to break the ice) :)

Some comments:
- I decided to add a couple of [type annotated methods](https://github.com/galaxyproject/galaxy/compare/dev...davelopez:fastapi-migration-licenses?expand=1#diff-b0cd1bfe489edaa5d2f79359bf3718b115b1a36ac1cf104328e2021161f22a0aR79-R82) to the `LicensesManager` instead of using the existing ones so I can slightly change their behaviour while maintaining the previous controller intact.
- I assumed that the `/api/licenses/{id}` endpoint will always be invoked using an SPDX identifier as `Path` parameter and not an `url` (the manager supports both). If the URL must be supported the Path parameter should be changed to a Query.

TODO:
- [x] Add functional tests for the licenses API endpoints


lib/galaxy/managers/licenses.py
```diff
@@ -1,10 +1,28 @@
 import json
 import logging
+from typing import List, Optional
```
mvdbeek: We typically do multiple imports from the same module like so:
```suggestion
from typing import (
    List,
    Optional
)
```

lib/galaxy/managers/licenses.py
```diff
@@ -1,10 +1,28 @@
...
 
 from pkg_resources import resource_string
+from pydantic import BaseModel, Field, HttpUrl
```
mvdbeek:
```suggestion
from pydantic import (
    BaseModel,
    Field,
    HttpUrl
)
```

mvdbeek (commented):
That's looking very nice!

lib/galaxy/managers/licenses.py
davelopez: Thanks Marius! Good to know :)

jmchilton:
> I decided to add a couple of type annotated methods to the LicensesManager instead of using the existing ones so I can slightly change their behaviour while maintaining the previous controller intact.

This code hasn't been included in a Galaxy release yet - this is probably the best time to get everything perfect and synchronized. For instance, I would move that ``if license is None: raise ObjectNotFound()`` right into the manager.

davelopez (commented):
Thanks for the quick review!

davelopez:
I tried to fully document the API endpoints to serve as an example.
Also, I took the liberty to add documentation for the API tags [here](https://github.com/galaxyproject/galaxy/pull/11056/commits/d9ff4b1b706b5634c2de987150f6bdc7107f5fff) as explained in the [FastAPI documentation](https://fastapi.tiangolo.com/tutorial/metadata/#metadata-for-tags).

lib/galaxy/webapps/galaxy/fast_app.py
```diff
@@ -5,7 +5,24 @@
...
+        "name": "licenses",
+        "description": "Operations with [SPDX licenses](https://spdx.org/licenses/).",
+    },
+]
+
+
+def add_open_api_tags_metadata(app: FastAPI) -> None:
+    """Adds OpenAPI documentation to the APIRouter tags"""
+    if app:
+        app.openapi_tags = api_tags_metadata
```
mvdbeek: Could we just add this on line 52 as the docs do (`FastAPI(openapi_tags=api_tags_metadata)`)? I'm not sure setting the attribute on the instance is really the public API we should be using?

davelopez: Yes, that was my first approach, but then I noticed that when launching from `/scripts/fapi.py` the app was already instanced in `fapi.py` [line 12](https://github.com/galaxyproject/galaxy/blob/dev/scripts/fapi.py#L12). I am not sure if this script will be just temporal and we assume that the only creation point will be on line 52.

mvdbeek: I think we can do this:
```
diff --git a/scripts/fapi.py b/scripts/fapi.py
index c24d54e9bf..1d94458bce 100644
--- a/scripts/fapi.py
+++ b/scripts/fapi.py
@@ -1,5 +1,4 @@
 import uvicorn
-from fastapi import FastAPI
 from galaxy_main import (
     GalaxyConfigBuilder,
     main,
@@ -9,16 +8,13 @@ from galaxy.util.properties import load_app_properties
 from galaxy.webapps.galaxy.buildapp import app_factory
 from galaxy.webapps.galaxy.fast_app import initialize_fast_app
 
-app = FastAPI()
-
 
 def wsgiloop(args, log):
     config_builder = GalaxyConfigBuilder(args)
     kwds = config_builder.app_kwds()
     kwds = load_app_properties(**kwds)
     gx = app_factory(global_conf=config_builder.global_conf(), **kwds)
-    initialize_fast_app(gx, app=app)
-    uvicorn.run(app)
+    uvicorn.run(initialize_fast_app(gx))
 
 
 if __name__ == '__main__':

```
I only created the app object in the global namespace so we could use gunicorn, but that doesn't work yet with the way we the configuration is read.

mvdbeek: I'll open a separate PR for that so we can keep this PR as a clean example for how to do migrations.

lib/galaxy_test/api/test_licenses.py
```diff
@@ -0,0 +1,23 @@
...
+        self._assert_status_code_is(response, 200)
+
+    def test_get_license(self):
+        licenseId = "Apache-2.0"
+        response = self._get(f"licenses/{licenseId}")
+        self._assert_status_code_is(response, 200)
+        self._assert_matches_license_id(licenseId, response.json())
+
+    def test_404_on_unknown_license(self):
+        licenseId = "uknown"
```
mvdbeek:
```suggestion
        licenseId = "unknown"
```

lib/galaxy_test/api/test_licenses.py
```diff
@@ -0,0 +1,23 @@
...
+        response = self._get(f"licenses/{licenseId}")
+        self._assert_status_code_is(response, 404)
```
mvdbeek, line 19: Haha, so you found a bug in the old API route, which returns a 200 here. Can you use `get_license_by_id` in the old endpoint too ?

lib/galaxy/webapps/galaxy/fast_app.py
davelopez: Cool! I'll remove the `add_open_api_tags_metadata` then :)

lib/galaxy_test/api/test_licenses.py
davelopez: Sure! :)

davelopez:
Did I break the `RolesApiTestCase::test_create_only_admin` test somehow? :flushed:

mvdbeek:
I've seen this fail before in other PRs, just one of those flaky tests we'll try fixing on Monday.

mvdbeek (approved):

lib/galaxy/webapps/galaxy/fast_app.py
```diff
@@ -30,6 +41,7 @@ async def message_exception_middleware(request: Request, exc: MessageException)
         )
 
 
+
```
mvdbeek:
```suggestion
```
sorry about that, I should stop using the github interface to do these things.

davelopez:
Thanks for resolving the conflict! :)

mvdbeek:
Thank you @davelopez!

Issue #10943, opened 2020-12-16
Add /download API endpoint for datasets and HDAs


innovate-invent:
Currently the /display endpoint is being used which does not force the browser to download the dataset and the browser tries to open the file if it is a known mime type.

The endpoint should set `Content-Disposition: attachment; filename=""` with the filename property settable as part of the GET request from the client so that the client can control how the file is presented to the user in the download prompt.

The /download endpoint should be added to both /api/datasets/../download and /api/history/../contents/../download

mvdbeek:
I think you should get the Content-Disposition header if you send raw=true or something like that, but I agree a separate download route would be better

innovate-invent:
raw=true changes `Content-Type	application/json; charset=UTF-8` and causes the browser to try and parse the content as JSON

PR #11058, opened 2021-01-06
[20.09] Allow setting filename_pattern by client

innovate-invent:
Allow passing the filename_pattern in the download request.
exa: `/datasets/f69a3f7a2f8aa212/display?to_ext=txt&filename_pattern=my_custom_file_name.${ext}`

This is a step toward #10943

### DIFF
--- a/lib/galaxy/datatypes/data.py
+++ b/lib/galaxy/datatypes/data.py
@@ -438,7 +438,7 @@ class Data(metaclass=DataMeta):
                 return self._archive_composite_dataset(trans, data, do_action=kwd.get('do_action', 'zip'))
             else:
                 trans.response.headers['Content-Length'] = int(os.stat(data.file_name).st_size)
-                filename = self._download_filename(data, to_ext, hdca=kwd.get("hdca"), element_identifier=kwd.get("element_identifier"))
+                filename = self._download_filename(data, to_ext, hdca=kwd.get("hdca"), element_identifier=kwd.get("element_identifier"), filename_pattern=kwd.get("filename_pattern"))
                 trans.response.set_content_type("application/octet-stream")  # force octet-stream so Safari doesn't append mime extensions to filename
                 trans.response.headers["Content-Disposition"] = 'attachment; filename="%s"' % filename
                 return open(data.file_name, 'rb')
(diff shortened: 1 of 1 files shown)


mvdbeek:
I tend to say that we don't want make this configurable, because clients can choose the filename they like, the header can be ignored and I don't think we want clients to know the available things in `template_values`. Could you maybe elaborate on the use case in https://github.com/galaxyproject/galaxy/issues/10943 ?

innovate-invent:
> because clients can choose the filename they like

How are clients able to choose the filename they like?
This is to allow clients to choose the filename, with the added feature of providing some templated values.

mvdbeek:
> How are clients able to choose the filename they like?

I don't know how to answer this without more details for what you're doing. The client is free to do what it wants, `curl <some url> -o <some_filename>`  will just not use the header. If you're building a web UI, you'd do `<a href="<smn>" download="some_filename">`

innovate-invent:
`<a href="<smn>" download="some_filename">` the download attribute is not generally supported by browsers.

mvdbeek:
That's html5, this should be supported in every browser.

innovate-invent:
Chrome dropped support for cross origin urls, internet explorer simply doesn't support it, Firefox supports it but is problematic, Opera doesn't support it explicitly, Safari does not support it.

mvdbeek:
I'd rather we remove this than add to it, but I guess it doesn't do any harm

innovate-invent:
Thank you!

dannon:
https://caniuse.com/download looks pretty reasonable -- IE is dead, and it should work everywhere else you mentioned?

innovate-invent:
The same-origin restriction alone makes the download attribute unusable for galaxy.

dannon:
That's pretty much the same story?  My opposition isn't a -1 or anything, but I still don't think this is something that belongs on the webserver.  Are there other applications that actually behave like this?

innovate-invent:
How do you mean other applications? and which behavior?
If you mean do other web servers accept a filename to send in the Content-Disposition header? Generally the filename the webserver is suggesting is the one desired, but I don't like having Galaxies branding baked into the file name when using it as a backend. Applications that do want to give control to the client do use this method or use a javascript downloader that loads the file into memory and then "downloads" the file from memory with custom settings. This is complicated when dealing with large files.

dannon:
That's exactly what I was getting at.  I don't love it, but it really is not a big deal.  Surely there's a better way to encapsulate your services to work around the CORS restriction.  Anyway, like I said -- don't love it, but it isn't a big deal.  Not worth spending more time on.

PR #11063, opened 2021-01-06
Add validation for collection parameters

bernt-matthias:
- [x] include min / max?
- [ ] decide if validation of the included data sets is sufficient / if we may want collection validators in the future
- [x] doc in xsd

### DIFF
--- a/lib/galaxy/tools/parameters/basic.py
+++ b/lib/galaxy/tools/parameters/basic.py
@@ -1759,6 +1759,43 @@ class BaseDataToolParameter(ToolParameter):
         else:
             return app.model.context.query(app.model.HistoryDatasetAssociation).get(int(value))
 
+    def validate(self, value, trans=None):
+
+        def do_validate(v):
+            for validator in self.validators:
+                if validator.requires_dataset_metadata and v and hasattr(v, 'dataset') and v.dataset.state != galaxy.model.Dataset.states.OK:
+                    return
+                else:
+                    validator.validate(v, trans)
+
+        dataset_count = 0
+        if value:
+            if self.multiple:
+                if not isinstance(value, list):
+                    value = [value]
+            else:
+                value = [value]
+
+            for v in value:
+                if isinstance(v, galaxy.model.HistoryDatasetCollectionAssociation):
+                    for dataset_instance in v.collection.dataset_instances:
+                        dataset_count += 1
+                        do_validate(dataset_instance)
+                elif isinstance(v, galaxy.model.DatasetCollectionElement):
+                    for dataset_instance in v.child_collection.dataset_instances:
+                        dataset_count += 1
+                        do_validate(dataset_instance)
+                else:
+                    dataset_count += 1
+                    do_validate(v)
+
+        if self.min is not None:
+            if self.min > dataset_count:
+                raise ValueError("At least %d datasets are required for %s" % (self.min, self.name))
+        if self.max is not None:
+            if self.max < dataset_count:
+                raise ValueError("At most %d datasets are required for %s" % (self.max, self.name))
+
 
 class DataToolParameter(BaseDataToolParameter):
     # TODO, Nate: Make sure the following unit tests appropriately test the dataset security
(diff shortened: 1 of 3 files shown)


lib/galaxy/tools/parameters/basic.py
```diff
@@ -2176,6 +2176,44 @@ def to_text(self, value):
...
+            #     if not isinstance(value, list):
+            #         value = [value]
+            # else:
+            #     value = [value]
+
+            for v in value:
+                if isinstance(v, galaxy.model.HistoryDatasetCollectionAssociation):
+                    for dataset_instance in v.collection.dataset_instances:
+                        dataset_count += 1
+                        do_validate(dataset_instance)
```
mvdbeek: Iterating over collections this way is expensive for large collections. 
- Given that we skip the validation if the dataset is not in OK state, can we first check that  all of the collection elements are in OK state with `v.collection.dataset_states_and_extensions_summary` / `v.child_collection.dataset_states_and_extensions_summary` ? The first returned set should contain the states, so it should be of length 1 and only contain `"ok"`.
- Can we apply all validators for each dataset instance, so that we iterate over elements just once, instead of iterating over validators, and then iterating over collections ?

Do you know if the validator is called before or after splitting mapped over inputs ? If this runs before, the dataset_count is the count for all passed in collections. So if the input should be a list collection, but you give it a nested list, it is the sum of the nested list. It might be correct given you're expecting `DatasetCollectionElement ` instances, but if you can double-check this that would be great.

mvdbeek: And with something like
```
diff --git a/lib/galaxy/model/__init__.py b/lib/galaxy/model/__init__.py
index e19d1562f1..2b5f54f56c 100644
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -4209,7 +4209,7 @@ class DatasetCollection(Dictifiable, UsesAnnotations, RepresentById):
                 element.child_collection.finalize(collection_type_description.child_collection_type_description())

     @property
-    def dataset_instances(self):
+    def dataset_instances_query(self):
         db_session = object_session(self)
         if db_session and self.id:
             dc = alias(DatasetCollection.table)
@@ -4229,7 +4229,7 @@ class DatasetCollection(Dictifiable, UsesAnnotations, RepresentById):
                 depth_collection_type = depth_collection_type.split(":", 1)[1]
             select_from = select_from.outerjoin(hda, hda.c.id == de.c.hda_id)
             select_stmt = select([hda]).select_from(select_from).where(dc.c.id == self.id).distinct()
-            return db_session.query(HistoryDatasetAssociation).select_entity_from(select_stmt).filter(HistoryDatasetAssociation.id.isnot(None)).all()
+            return db_session.query(HistoryDatasetAssociation).select_entity_from(select_stmt).filter(HistoryDatasetAssociation.id.isnot(None))
         else:
             # Sessionless context
             instances = []
@@ -4241,6 +4241,14 @@ class DatasetCollection(Dictifiable, UsesAnnotations, RepresentById):
                     instances.append(instance)
             return instances

+    @property
+    def dataset_instances(self):
+        return self.dataset_instances_query.all()
+
+    @property
+    def dataset_instance_count(self):
+        return self.dataset_instances_query().count()
+
     @property
     def dataset_elements(self):
         elements = []
```
we should be able to get the dataset count more efficiently.

mvdbeek: eh, that still fetches all the columns, but if we get to that point we can use func.count

bernt-matthias: > Iterating over collections this way is expensive for large collections.
> 
>     * Given that we skip the validation if the dataset is not in OK state, can we first check that  all of the collection elements are in OK state with `v.collection.dataset_states_and_extensions_summary` / `v.child_collection.dataset_states_and_extensions_summary` ? The first returned set should contain the states, so it should be of length 1 and only contain `"ok"`.
> 
>     * Can we apply all validators for each dataset instance, so that we iterate over elements just once, instead of iterating over validators, and then iterating over collections ?

Good point will. Would this be also useful for the validation in `DataToolParameter` which also may involve collections: https://github.com/galaxyproject/galaxy/blob/ca8435681ae36ebb03717b78c76c7e52089e6b10/lib/galaxy/tools/parameters/basic.py#L1908

I guess it would also be good to check if there is a validator before starting the iteration over all data sets. Otherwise we would iterate even if there is none. 
 
> Do you know if the validator is called before or after splitting mapped over inputs ? If this runs before, the dataset_count is the count for all passed in collections. So if the input should be a list collection, but you give it a nested list, it is the sum of the nested list. It might be correct given you're expecting `DatasetCollectionElement ` instances, but if you can double-check this that would be great.

I guess extending the test is possible by using the `min` and `max` attributes.

bernt-matthias: Just discovered that the count is wrong anyway (also for `DataToolParameter`) if a validator is defined, because the count is accumulated over all validators, i.e. if there are two validators then each data set is counted twice. 

Since a `MetadataValidator` is added per default to each `DataToolParameter` the count is wrong as soon as the tool defines a validator. 

Also odd that `DataCollectionToolParameters` do not add a `MetadataValidator`, or?

mvdbeek: Yeah, so I think the count being wrong you can fix by iterating just once, or by relying on `dataset_instance_count` as in the diff.



> Also odd that `DataCollectionToolParameters` do not add a `MetadataValidator`, or?

That's a grey-area, dataset collections don't have metadata, but I could see that maybe you want to check the metadata of elements.

PR #11068, opened 2021-01-06
Changes to Collection Operation Help sections and parameter names

nekrut:
Collection operations are some of the most powerful tools. Yet their help sections are minimalist at this point. This PR adds some additional clarification to help sections.


jmchilton:
None of these versions should be updated since they don't change the functionality or interface from a Galaxy perspective.

lib/galaxy/tools/zip_collection.xml
```diff
@@ -1,13 +1,14 @@
...
       tool_type="zip_collection">
+  <description>two collections into one</description>
   <type class="ZipCollectionTool" module="galaxy.tools" />
   <action module="galaxy.tools.actions.model_operations"
           class="ModelOperationToolAction"/>
   <inputs>
-    <param type="data" name="input_forward" label="Input Dataset (Forward)" />
-    <param type="data" name="input_reverse" label="Input Dataset (Reverse)" />
+    <param type="data" name="input_forward" label="Input collection 1" help="Example: forward reads"/>
+    <param type="data" name="input_reverse" label="Input collection 2" help="Example: reverse reads"/>
```
jmchilton: This literally takes in datasets - not collections - the type is right there. I think this change in inappropriate. Maybe we could add a new attribute to change the label if a tool is fed a collection. 

Short of that though how about label="Input 1" with help of help="Example: forward read(s)"?

jmchilton:
Everywhere you have "Collection (Paired)" in the diagrams I think this needs to be "List of Pairs" or "Paired List" or "Paired List Collection".

lib/galaxy/tools/zip_collection.xml
nekrut: It takes either datasets or collections

jmchilton: You can map over a collection but that is true of every tool, the literal tool takes in datasets and not collections. I suggest we just drop both words to avoid this argument though?

lib/galaxy/tools/unzip_collection.xml
```diff
@@ -1,29 +1,39 @@
...
 ----
 
-**Example**
+**Example**:
+
+Given a paired collection of forward and reverse reads this tool will "unzip" it into two collections containing forward and reverse reads, respectively:
+
+.. image:: ${static_path}/images/tools/collection_ops/unzip.png
+  :width: 400
+  :alt: Unzipping operation
```
jmchilton, line 36: Thanks for changing the naming in the images - they are really nice!

lib/galaxy/tools/unzip_collection.xml
```diff
@@ -1,29 +1,39 @@
...
+  <description>collection into multiple ones</description>
   <type class="UnzipCollectionTool" module="galaxy.tools" />
   <action module="galaxy.tools.actions.model_operations"
           class="ModelOperationToolAction"/>
     <edam_operations>
         <edam_operation>operation_3359</edam_operation>
     </edam_operations>
   <inputs>
-    <param type="data_collection" collection_type="paired" name="input" label="Input Paired Dataset" />
+    <param type="data_collection" collection_type="paired" name="input" label="Paired List Collection to unzip" />
```
jmchilton: Like in the other tool suggestion - I think this should be more generic since this natively takes in pairs and not list of pairs - I think if we changed it to "Paired input" we can circumvent that argument?

jmchilton:
How are you generating these images - are there SVGs that can be included in the repo?

I wonder if we should have a reshape collection button in the GUI that would show all these options when you click on a collection - or maybe the ones that are available for that collection type and the tools available 🤔.

lib/galaxy/tools/zip_collection.xml
nekrut: It is now simply `Input 1` and `Input 2`

lib/galaxy/tools/unzip_collection.xml
nekrut: Changed to `Paired input to unzip`

nekrut:
> How are you generating these images - are there SVGs that can be included in the repo?
> 
> I wonder if we should have a reshape collection button in the GUI that would show all these options when you click on a collection - or maybe the ones that are available for that collection type and the tools available 🤔.

These are PNG at the moment. Let's get this one merged and I will figure out how to generate SVG. I do not like current PNGs either - they are soft for some reason.

nekrut:
> > If collection contains multiple elements with the same name, all of them will be extracted and a new collection will be created.
> 
> This is not really true. If you map over a nested collection - this operation will be applied to each element in the outer list and extract elements from the inner most collection. The way your phrasing this it sounds like collections can have elements with the same name at the same depth (the cannot) and the output type (collection or dataset) depends on whether it has duplicated names (it doesn't - it depends on the same mapping rules used for every tool and collection in Galaxy).
> 
> Have you seen #12238 I think it will be a better approach to explaining these nuisances to users.

This is a wonderful idea. We will really need SVGs for these

lib/galaxy/tools/build_list.xml
```diff
@@ -1,16 +1,16 @@
 <tool id="__BUILD_LIST__"
-      name="Build List"
+      name="Build collection"
```
mvdbeek: This tool only builds lists, I think that name is misleading ?

lib/galaxy/tools/build_list.xml
```diff
@@ -1,16 +1,16 @@
...
+  <description></description>
   <type class="BuildListCollectionTool" module="galaxy.tools" />
   <action module="galaxy.tools.actions.model_operations"
           class="ModelOperationToolAction"/>
   <edam_operations>
     <edam_operation>operation_2409</edam_operation>
   </edam_operations>
   <inputs>
-    <repeat name="datasets" title="Dataset">
+    <repeat name="datasets" title="Dataset" help="can be dataset or collection. Multiple collection must be of the same length.">
```
mvdbeek: I get what you're saying here, but the input really is a dataset. You can map a collection over the input here, but then the tool is run in batch.
```suggestion
    <repeat name="datasets" title="Dataset" help="If collections are provided they should contain the same number of items.">
```

lib/galaxy/tools/build_list.xml
```diff
@@ -1,16 +1,16 @@
...
       <param type="data" name="input" optional="true" label="Input Dataset" />
```
mvdbeek:
```suggestion
      <param type="data" name="input" optional="true" label="Input Dataset" help="If providing a collection here the tool will be run in batch and one collection per element is created."/>
```

lib/galaxy/tools/build_list.xml
```diff
@@ -20,8 +20,33 @@
   </outputs>
   <help><![CDATA[
 
+========
+Synopsis
+========
+
+Builds a new collection from individual datasets or collections.
```
mvdbeek:
```suggestion
Builds a new list collection from individual datasets or collections.
```

lib/galaxy/tools/build_list.xml
```diff
@@ -20,8 +20,33 @@
...
+
+===========
+Description
+===========
+
+This tool combines individual datasets or collections into a new collection. The simplest scenario is building a new colection from individual datasets (case **A** in the image below). You can merge a collection with individual dataset(s). In this case (see **B** in the image below) the individual dataset(s) will be merged with each element of the input collection to create a nested collection. Finally, two or more collection can be merged together creating a nested collection (case **C** in the image below).
+
+.. class:: warningmark 
+
+**Note**: When merging collections (e.g., case **C** below) the inout collection **must** have equal number of elements.
```
mvdbeek:
```suggestion
**Note**: When merging collections (e.g., case **C** below) the input collection **must** have equal number of elements.
```

lib/galaxy/tools/build_list.xml
```diff
@@ -46,3 +71,6 @@ This tool will create a new collection from your history datasets but your quota
     </test>
   </tests>
 </tool>
+
+
+
```
mvdbeek:
```suggestion
```

lib/galaxy/tools/extract_dataset.xml
```diff
@@ -40,6 +40,22 @@
   </outputs>
   <help><![CDATA[
 
+========
+Synopsis
+========
+
+Extracs datasets from a collection based on either position or name.
```
mvdbeek:
```suggestion
Extracts datasets from a collection based on either position or name.
```

lib/galaxy/tools/extract_dataset.xml
```diff
@@ -40,6 +40,22 @@
...
+
+===========
+Description
+===========
+
+The tool allow extracting datasets based on position (**The first dataset** and **Select by index** options) or name (**Select by element identifier** option). If collection contains multiple elements with the same name, all of them will be extracted and a new collection will be created. Finally
```
mvdbeek: Collection element identifiers are necessarily unique within the collection they belong to. Is this about extracting a nested collection ? e.g extracting element A (or the first element) from
```yaml
outer_collection:
  treatment:
    A
    B
  control:
    A
    B
```
becomes
```
outer_collection:
  treatment
  control
```
wh
?
I don't know that this is the right place to explain how nested collections work, this isn't specific to this tool.
Maybe we need to enhance the collection tutorial to walk through how mapping over works ?

```suggestion
The tool allow extracting datasets based on position (**The first dataset** and **Select by index** options) or name (**Select by element identifier** option). This tool effectively collapses the inner-most collection into a dataset. For nested collections (e.g a list of lists of lists: outer:middle:inner, extracting the inner dataset element) a new list is created where the selected element takes the position of the inner-most collection (so outer:middle, where middle is not a collection but the inner dataset element).
```

lib/galaxy/tools/filter_from_file.xml
```diff
@@ -1,20 +1,20 @@
...
+    <description></description>
     <type class="FilterFromFileTool" module="galaxy.tools" />
     <action module="galaxy.tools.actions.model_operations"
             class="ModelOperationToolAction"/>
     <edam_operations>
         <edam_operation>operation_3695</edam_operation>
     </edam_operations>
     <inputs>
-        <param type="data_collection" name="input" label="Input Collection" help="A list whose elements will be filtered."/>
+        <param type="data_collection" name="input" label="Input Collection" help="collection to filter identifiers in"/>
```
mvdbeek:
```suggestion
        <param type="data_collection" name="input" label="Input Collection" help="A collection whose elements will be filtered."/>
```

lib/galaxy/tools/relabel_from_file.xml
```diff
@@ -1,28 +1,28 @@
...
+                <option value="tabular">Old-new map</option>
             </param>
             <when value="txt">
                 <param type="data" name="labels" format="txt" label="New Identifiers" />
                 <param name="strict" type="boolean" label="Ensure strict mapping" help="If selected, the target file must contain exactly the correct number of lines." truevalue="true" falsevalue="false" />
             </when>
             <when value="tabular">
                 <param type="data" name="labels" format="tabular" label="New Identifiers" />
-                <param name="strict" type="boolean" label="Ensure strict mapping" help="If selected, the target file must contain exactly the correct number of lines and each input identifier must match exactly one element of the input collection." truevalue="true" falsevalue="false" />
+                <param name="strict" type="boolean" label="Ensure strict mapping" help="If YES, the target file must contain exactly the correct number of lines and each input identifier must match exactly one element of the input collection." truevalue="true" falsevalue="false" />
```
mvdbeek:
```suggestion
                <param name="strict" type="boolean" label="Ensure strict mapping" help="If checked, the target file must contain exactly the correct number of lines and each input identifier must match exactly one element of the input collection." truevalue="true" falsevalue="false" />
```

lib/galaxy/tools/relabel_from_file.xml
```diff
@@ -137,23 +137,75 @@
     </tests>
     <help><![CDATA[
 
+========
+Synopsis
+========
+
+This tool changes names of datasets within a collection using identifiers from a supplied file.
```
mvdbeek:
```suggestion
This tool changes identifiers of datasets within a collection using identifiers from a supplied file. 
```
It doesn't change the dataset name. If you unhide the collection you see the dataset names.

lib/galaxy/tools/relabel_from_file.xml
```diff
@@ -137,23 +137,75 @@
...
+
+This tool changes names of datasets within a collection using identifiers from a supplied file. 
+
 .. class:: infomark
 
-        This tool will take an input list and a text file with new identifiers
-        and build a new list with the same datasets but these new identifiers.
-        The order and number of entries in the text file must match the order
-        of the items you want to rename in your dataset collection.
+This document uses terms *name*, *label*, and *identifier* interchangeably: they all mean the same thing - name of individual datasets inside a collection.
```
mvdbeek: Name can be the outer collection's name as you see it in the history, or the dataset name, which is not visible in a collection. The only correct thing here is identifier, and the verb is `to label`.

lib/galaxy/tools/relabel_from_file.xml
```diff
@@ -137,23 +137,75 @@
...
-        and build a new list with the same datasets but these new identifiers.
-        The order and number of entries in the text file must match the order
-        of the items you want to rename in your dataset collection.
+This document uses terms *name*, *label*, and *identifier* interchangeably: they all mean the same thing - name of individual datasets inside a collection. 
+
+===========
+Description
+===========
+
+New names can be supplied as either a simple list or a tab-delimited file mapping old names to the new ones. This is controlled using **How should the new labels be specified?** drop-down:
```
mvdbeek:
```suggestion
New identifiers can be supplied as either a simple list or a tab-delimited file mapping old identifier to the new ones. This is controlled using **How should the new identifiers be specified?** drop-down:
```

lib/galaxy/tools/relabel_from_file.xml
```diff
@@ -137,23 +137,75 @@
...
-        contains the new label. This file may contain less entries than items
-        in the collection. In that case only matching list identifiers will be
-        relabeled.
+**Note** that the order and number of entries in the text file must match the order of the items you want to rename in your dataset collection.
 
-        Valid identifiers must contain only characters (a-z, A-Z), numbers (0-9),
-        dash and underscore (-, _). Punctuation or whitespace are not allowed.
+-------
+
+**Old-new map**
```
mvdbeek:
```suggestion
**Map original identifiers to new ones using a two column table**
```

lib/galaxy/tools/relabel_from_file.xml
```diff
@@ -1,28 +1,28 @@
...
     </edam_operations>
     <inputs>
-        <param type="data_collection" name="input" label="Input Collection" help="A list whose identifiers will be relabelled."/>
+        <param type="data_collection" name="input" label="Input Collection" help="collection to change identifiers in"/>
         <conditional name="how">
             <param type="select" name="how_select" label="How should the new labels be specified?">
-                <option value="txt">Using lines in a simple text file.</option>
-                <option value="tabular">Map original identifiers to new ones using a two column table.</option>
+                <option value="txt">Simple list</option>
+                <option value="tabular">Old-new map</option>
```
mvdbeek:
```suggestion
                <option value="txt">Using lines in a simple text file.</option>
                <option value="tabular">Map original identifiers to new ones using a two column table.</option>
```
I think both option values were more specific and helpful to users.

lib/galaxy/tools/relabel_from_file.xml
```diff
@@ -137,23 +137,75 @@
...
-        of the items you want to rename in your dataset collection.
+This document uses terms *name*, *label*, and *identifier* interchangeably: they all mean the same thing - name of individual datasets inside a collection. 
+
+===========
+Description
+===========
+
+New names can be supplied as either a simple list or a tab-delimited file mapping old names to the new ones. This is controlled using **How should the new labels be specified?** drop-down:
+
+**Simple list**
```
mvdbeek:
```suggestion
**Using lines in a simple text file**
```

lib/galaxy/tools/sort_collection_list.xml
```diff
@@ -1,8 +1,8 @@
 <tool id="__SORTLIST__"
-      name="Sort Collection"
+      name="Sort collection"
       version="1.0.0"
       tool_type="sort_collection">
-    <description>of list of datasets</description>
+    <description></description>
```
mvdbeek, line 5:
```suggestion
    <description>alphabetically, numerically, or using predetermined order from a supplied file</description>
```

lib/galaxy/tools/sort_collection_list.xml
```diff
@@ -142,7 +142,53 @@
...
+
+ Collection: [Horse123]
+             [Mule176] 
+             [Donkey543] 
+
+-------
+
+**Sorting from file**
+
+Alternative, one can supply a single column text file containing elements names in the desired sort order. For example, suppose there a collection::
```
mvdbeek:
```suggestion
Alternative, one can supply a single column text file containing elements identifiers in the desired sort order. For example, suppose there a collection::
```

lib/galaxy/tools/apply_rules.xml
```diff
@@ -1,5 +1,5 @@
  <tool id="__APPLY_RULES__"
-      name="Apply Rule to Collection"
+      name="Apply rules"
       version="1.1.0"
       tool_type="apply_rules_to_collection">
     <type class="ApplyRulesTool" module="galaxy.tools" />
```
mvdbeek, line 5:
```suggestion
    <description>to collection</description>
    <type class="ApplyRulesTool" module="galaxy.tools" />
```

mvdbeek (commented):
That's really comprehensive, but I think there are a few things that need adjustment to be universal or correct.

lib/galaxy/tools/sort_collection_list.xml
nekrut: I really do not want descriptions for collection tools. They all work on collections. If users want to learn more = read help section. It looks cleaner in the tool pane.

lib/galaxy/tools/apply_rules.xml
nekrut: same here

lib/galaxy/tools/extract_dataset.xml
nekrut: I initiated this process in part because collection warrant (1) separate section in GTN and (2) a dedicated manuscript. So, yes we need schematics illustrating these.

lib/galaxy/tools/build_list.xml
nekrut: fixed

lib/galaxy/tools/relabel_from_file.xml
nekrut: done

nekrut:
Returning it back to WIP to convert PNG to SVG as per @jmchilton suggestion. These look much more crisp.

jmchilton:
I think the SVG comment was about tracking the sources so they could be modified and I assumed the sources were readily available - sounds like this is not the case. I think it is fine to use actual PNG for the tools, etc.. I'm not sure this needs to be held up on SVG vs. PNG, though SVG does have some advantages obviously.

lib/galaxy/tools/extract_dataset.xml
```diff
@@ -40,6 +40,22 @@
   </outputs>
   <help><![CDATA[
 
+========
+Synopsis
+========
+
+Extracts datasets from a collection based on either position or name.
```
jmchilton: The name of the dataset is different than the element identifier. This should definitely be element identifier or identifier and not name.

lib/galaxy/tools/tag_collection_from_file.xml
```diff
@@ -1,8 +1,8 @@
 <tool id="__TAG_FROM_FILE__"
-      name="Tag elements from file"
+      name="Tag collection elements"
       version="1.0.0"
       tool_type="tag_from_file">
-    <description>from contents of a file</description>
```
jmchilton, line 5: One could imagine other ways to tag elements - I mean we should have tools that apply simple rules to do this and another that takes an input collection (perhaps from a downstream analysis) and a collection to mirror (perhaps an input) and re-tags the input based on the mirror. For these reasons I think placing file in the description makes a lot of sense - but I won't die on this hill if conciseness is important here I can respect that.

nekrut: yes, let's return to this once we have more tools :)

lib/galaxy/tools/extract_dataset.xml
nekrut: edited

jmchilton:
Thanks for the SVGs. I think on the whole I like most of these changes, there is still a few things in here that I cringe at but I guess it reflects real world usage over my theoretical understanding of how I expected these things to be used and thought of.

PR #11071, opened 2021-01-07
Sync vuex to galaxy

Nerdinacan:
This PR installs a more robust way of syncing legacy global Galaxy properties onto the Vuex store. I had something similar in the past, but it was showing some bugs, especially with the Upload module.

Some of the newer Vue components were clearly authored by importing the global Galaxy variable directly, which is hugely problematic. I feel like the authors didn't have good enough guidance on how to import those variables without making them yet another kind of injected global. It's my hope that this system makes it more clear and will encourage component authors to build within the new ecosystem instead of relying on the untestable global Galaxy objects in the legacy code.

All of these sync functions can eventually be removed as the legacy functionality they bridge is ported into Vue, but until then, this should provide a clean way to get data into Vuex.

This PR doesn't provide a lot of new functionality in and of itself, but I will require it to fix some of the more broken features, the most urgent of which is the Upload module.

The most relevant changes can be seen in:
[client/src/store/index.js](https://github.com/Nerdinacan/galaxy/blob/sync_vuex_to_galaxy/client/src/store/index.js)


client/src/store/syncVuextoGalaxy.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * Watches for key changes in global Galaxy app variable and trnasfers them into Vuex via commits.
```
mvdbeek:
```suggestion
 * Watches for key changes in global Galaxy app variable and transfers them into Vuex via commits.
```

client/src/store/index.js
```diff
@@ -40,34 +46,29 @@ export function createStore() {
...
             informationStore: jobStore,
-            betaHistory: betaHistoryStore,
         },
     };
 
-    // Initialize state
-
+    // Initializes global store elements.
+    // Vuex can't lazy-load, so we much manually insert global data at some point and we currently
+    // don't have a root App component container (which is traditionally used for such operations).
```
mvdbeek, line 57: Can you outline what we need to do for this ? Should this sit one layer above the masthead ?

client/src/components/History/model/syncCurrentHistoryToGalaxy.js
```diff
@@ -1,28 +1,23 @@
...
-    const result$ = currentHistoryPanel$.pipe(
-        switchMap(model => monitorBackboneModel(model, "id").pipe(
-            pluck("id")
-        ))
+export function syncCurrentHistoryToGalaxy(galaxy$, store) {
+    const historyId$ = galaxy$.pipe(
+        pluck("currHistoryPanel", "model"),
+        filter(Boolean),
+        switchMap((model) => monitorBackboneModel(model).pipe(shareReplay(1))),
+        pluck("id"),
```
mvdbeek: This should emit whenever I change history in any of the legacy components, right ?
It doesn't seem to do that, and `console.log("subscription id", id);` only emits on first page load.

mvdbeek: It looks like maybe we're getting a new backbone model with every history switch, so observing the stale backbone model for changes doesn't work (for the history id case).
I am concluding that from how I was able to get it to work with
```
diff --git a/client/src/components/History/model/syncCurrentHistoryToGalaxy.js b/client/src/components/History/model/syncCurrentHistoryToGalaxy.js
index 412461a225..9084e04f47 100644
--- a/client/src/components/History/model/syncCurrentHistoryToGalaxy.js
+++ b/client/src/components/History/model/syncCurrentHistoryToGalaxy.js
@@ -1,13 +1,14 @@
 // Sync Galaxy store to legacy galaxy current history
 
-import { switchMap, pluck, shareReplay, filter, distinctUntilChanged } from "rxjs/operators";
+import { switchMap, pluck, shareReplay, filter, distinctUntilChanged, tap } from "rxjs/operators";
 import { monitorBackboneModel } from "utils/observable/monitorBackboneModel";
 
 export function syncCurrentHistoryToGalaxy(galaxy$, store) {
     const historyId$ = galaxy$.pipe(
         pluck("currHistoryPanel", "model"),
+        tap((model) => console.log('tapped model', model)),
         filter(Boolean),
-        switchMap((model) => monitorBackboneModel(model).pipe(shareReplay(1))),
+        // switchMap((model) => monitorBackboneModel(model).pipe(shareReplay(1))),
         pluck("id"),
         distinctUntilChanged()
     );
diff --git a/client/src/utils/observable/waitForInit.js b/client/src/utils/observable/waitForInit.js
index 702eb965b7..c519a6c893 100644
--- a/client/src/utils/observable/waitForInit.js
+++ b/client/src/utils/observable/waitForInit.js
@@ -10,7 +10,7 @@ import { map, filter, take, takeUntil } from "rxjs/operators";
 // prettier-ignore
 export function waitForInit(selector, cfg = {}) {
     const {
-        spamTime = 100,
+        spamTime = 1000,
         timeout = 10000,
         isValid = (val) => val !== undefined
     } = cfg;
@@ -18,7 +18,7 @@ export function waitForInit(selector, cfg = {}) {
     return interval(spamTime).pipe(
         map(() => selector()),
         filter(isValid),
-        take(1),
-        takeUntil(timer(timeout))
+        // take(1),
+        // takeUntil(timer(timeout))
     );
 }
```
Does that make sense ?

mvdbeek (commented):
I like the approach, but see my comments in line

client/src/store/index.js
Nerdinacan: In a single page app there is usually only one mount or starting point for the app. At that point you would traditionally populate any global data (ajax call to get current user, load application configs etc.), and that would be triggered on some a created() hook in a root <app> component or something.

But, Galaxy is not yet that kind of app and we are mounting in multiple places from within backbone by calling individualized mount functions which will fire up a new Vue instance. Although they will share a global Vuex, the issue becomes "when do we initialize that Vuex instance?"

Vuex is by design a simple synchronous state manager. If you want to get data out,  this means that somebody, somewhere has to fire an action to populate the state, and that is what a provider can give us. because it is a component with lifecycle hooks.

When I am done fixing the Upload modal, or any other Vue component that is mounted independently they would look something like this:

```html
<AppConfig v-slot="config">
   <UploadModal v-if="config" :thing-i-need="config.thing" :other-relevent-setting="config.otherThing" />
</AppConfig>
```
Here AppConfig is a provider that, more than likely is just going to pass out a slot prop peeled directly from the Vuex state, but it is giving us a lifecycle hook which guarantees that we have the opportunity to load the config from the server if it's not yet available. We can check to make sure we don't do it more times than we wanted, but we have the necessary async hook to do so.

The Masthead is the other problematic piece of code that is making testing difficult and I would indeed do something similar there.

Does that make my plans clear?

mvdbeek: Awesome, can you just copy-paste that into a planning issue?

client/src/components/History/model/syncCurrentHistoryToGalaxy.js
Nerdinacan: The intent was that it would fire whenever Galaxy changes its currentHistory value, yes.

Thanks for your experiment! Let me carefully look to see if that's what I wanted. Let me make sure that change in behavior doesn't mess with other places I want to use waitForInit, and if so, I'll refactor stuff appropriately.

client/src/store/index.js
Nerdinacan: https://github.com/galaxyproject/galaxy/issues/11077

mvdbeek: Thanks!

client/src/components/History/model/syncCurrentHistoryToGalaxy.js
Nerdinacan: Good catch! I'm now monitoring all the backbone events on the currHistoryPanel object itself and just peeling the id off the model at that time. 

```javascript
const historyId$ = galaxy$.pipe(
    pluck("currHistoryPanel"),
    switchMap((panel) => {
        // relationship between currHistoryPanel and its model is murky, but I don't care, I'm
        // just going to grab the id fresh after every possible event
        return fromEvent(panel, "all").pipe(
            map(() => panel.model.id)
        );
    }),
    filter(Boolean),
    distinctUntilChanged()
);
```

As for your changes in the diff above, looks like all you did was slow down the initial check loop? You could do that by just passing in different durations like so:

```javascript
waitForInit(getGalaxyInstance, { spamTime: 1000 })
```

But that's a side issue, I don't believe that was the actual problem you found.

client/src/components/History/model/syncCurrentHistoryToGalaxy.js
```diff
@@ -1,28 +1,28 @@
...
-            pluck("id")
-        ))
+export function syncCurrentHistoryToGalaxy(galaxy$, store) {
+    // prettier-ignore
+    const historyId$ = galaxy$.pipe(
+        pluck("currHistoryPanel"),
+        switchMap((panel) => {
+            // relationship between currHistoryPanel and its model is murky, but I don't care, I'm
+            // just going to grab the id fresh after every possible event
+            return fromEvent(panel, "all").pipe(
```
mvdbeek, line 13: Awesome, I didn't know what event to monitor,  `"all"` makes sense. This is working great now.

client/src/components/History/model/syncCurrentHistoryToGalaxy.js
mvdbeek: I just increased this because I didn't find out how to get events out of the backbone model, so I put `tap` operators all over the place 😆

mvdbeek (approved):
Working great now, thanks @Nerdinacan!

Nerdinacan:
@mvdbeek Thanks for peeking. This will be helpful for the next stuff on my list.

PR #11073, opened 2021-01-07
Verify behavior of non-existent from_work_dir outputs.

mvdbeek:
These are just the tests @jmchilton wrote for https://github.com/galaxyproject/galaxy/pull/8875. They pass now, so it seems the fix in that PR is not necessary anymore.

### DIFF
--- a/test/integration/test_job_outputs_to_working_directory.py
+++ b/test/integration/test_job_outputs_to_working_directory.py
@@ -1,0 +1,18 @@
+"""Run various framework tool tests with outputs_to_working_directory."""
+
+from galaxy_test.driver import integration_util
+
+
+class JobOutputsToWorkingDirectoryIntegrationInstance(integration_util.IntegrationInstance):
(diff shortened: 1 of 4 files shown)


test/integration/test_job_outputs_to_working_directory.py
```diff
@@ -0,0 +1,18 @@
+"""Run various framework tool tests with outputs_to_working_directory."""
+
+from galaxy_test.driver import integration_util
+
+
+class JobOutputsToWorkingDirectoryIntegrationInstance(integration_util.IntegrationInstance):
+    """Describe a Galaxy test instance with embedded pulsar configured."""
```
nsoranzo: Update?

mvdbeek:
```suggestion
    """Describe a Galaxy test instance with outputs_to_working_directory enabled."""
```

nsoranzo (approved):

PR #11078, opened 2021-01-07
Allow extra job metadata to be specified for k8s

nuwang:
Allow extra job metadata to be defined as follows:
```
<job_conf>
      <plugin id="k8s" type="runner" load="galaxy.jobs.runners.kubernetes:KubernetesJobRunner">
        ...
        <param id="k8s_job_metadata">
          labels:
              mylabel1: myvalue1
          annotations:
              myannotation1: myvalue1
              myannotation2: myvalue2
        </param>
      </plugin>
</job_conf>
```

### DIFF
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -60,6 +60,7 @@ class KubernetesJobRunner(AsynchronousJobRunner):
             k8s_timeout_seconds_job_deletion=dict(map=int, valid=lambda x: int > 0, default=30),
             k8s_job_api_version=dict(map=str, default=DEFAULT_JOB_API_VERSION),
             k8s_job_ttl_secs_after_finished=dict(map=int, valid=lambda x: x is None or int(x) >= 0, default=None),
+            k8s_job_metadata=dict(map=str, default=None),
             k8s_supplemental_group_id=dict(map=str),
             k8s_pull_policy=dict(map=str, default="Default"),
             k8s_run_as_user_id=dict(map=str, valid=lambda s: s == "$uid" or s.isdigit()),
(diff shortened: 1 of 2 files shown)


mvdbeek:
Can you base the PR on a recent dev branch ? It's still running some very old test config here.

nuwang:
No probs, just pushed it in.

mvdbeek:
Can you add an example and a comment to https://github.com/jmchilton/galaxy/blob/58dd38119ea396e10aa875cfeb54dd9383281062/lib/galaxy/config/sample/job_conf.xml.sample_advanced#L125 ?

nuwang:
@mvdbeek Thanks for catching that. Documented a bunch of missing params.

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -242,6 +243,9 @@ def __get_k8s_job_spec_template(self, ajs):
         # TODO include other relevant elements that people might want to use from
         # TODO http://kubernetes.io/docs/api-reference/v1/definitions/#_v1_podspec
         k8s_spec_template["spec"]["securityContext"] = self.__get_k8s_security_context()
+        extra_metadata = yaml.safe_load(self.runner_params['k8s_job_metadata'] or "{}")
```
jmchilton: This has to be an anti-pattern at this point - we've had YAML job confs for a while and that I implemented explicitly so we aren't loading JSON from XML fragments. Any chance we can at least only do the safe load if it is a string and not a dictionary?

jmchilton: Maybe make map above an identity of some sort?

```suggestion
        extra_metadata = self.runner_params['k8s_job_metadata'] or '{}'
        if isinstance(k8s_job_metadata, str):
            extra_metadata = yaml.safe_load(extra_metadata)
```
Is that what you had in mind @jmchilton ?

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -242,6 +243,11 @@ def __get_k8s_job_spec_template(self, ajs):
         # TODO include other relevant elements that people might want to use from
         # TODO http://kubernetes.io/docs/api-reference/v1/definitions/#_v1_podspec
         k8s_spec_template["spec"]["securityContext"] = self.__get_k8s_security_context()
+        extra_metadata = self.runner_params['k8s_job_metadata'] or '{}'
+        if isinstance(k8s_job_metadata, str):
```
```suggestion
        if isinstance(extra_metadata, str):
```

mvdbeek:
Seems pypi is down, but this works locally. Thanks @nuwang!

PR #11083, opened 2021-01-08
Show OpenAPI documentation for API tags

davelopez:
After merging #11056, I noticed that the `api_tags_metadata` was not passed to the FastAPI constructor.
This should fix it and show the documentation for tags like this:

image

### DIFF
--- a/lib/galaxy/webapps/galaxy/fast_app.py
+++ b/lib/galaxy/webapps/galaxy/fast_app.py
@@ -42,7 +42,7 @@ def add_exception_handler(
 
 
 def initialize_fast_app(gx_app):
-    app = FastAPI()
+    app = FastAPI(openapi_tags=api_tags_metadata)
 
     add_exception_handler(app)
     wsgi_handler = WSGIMiddleware(gx_app)


mvdbeek (approved):

hexylena:
Quick question for the release notes, apologies for not following this stuff more closely, how do I get to this page? Is it something that's accessible from Galaxy, or some separate documentation that gets rendered?

davelopez:
Hi @hexylena, you can access it at `http://127.0.0.1:8000/docs` but it is currently only supported in `dev` if you launch galaxy through `uvicorn` [here](https://github.com/galaxyproject/galaxy/blob/335de72138935c456993abb0e98941ca968ad7c7/lib/galaxy/webapps/galaxy/fast_factory.py#L10).

hexylena:
Thanks for the quick reply. Makes sense! I'll wait until this is visible in prod somewhere to mention it in the user-facing ones then. (and, briefly, any reason for /docs instead of `/api/docs`? or just `/api`? I tried those two before asking here :) )

mvdbeek:
No reason beyond it being the default route. I guess api/docs would be more consistent.

hexylena:
Only asking because I still have a vague hope of oneday seeing the real docs at /docs :)

davelopez:
+1 for the `/api/docs` route :)

PR #11085, opened 2021-01-08
New User Welcome page

astrovsky01:
New user welcome page:

On logging in for the first time, a user will be presented with a configurable set of major topics for Galaxy. Though the basic list would be data, tools, and workflows, this could be easily extended to include admin basics or server-specific information.
Screen Shot 2021-01-08 at 11 48 35 AM

Each of the topics can have several associated presentations, which are listed in a sub-menu:
Screen Shot 2021-01-08 at 11 49 03 AM

Presentations use a bootstrap vue carousel as a slideshow with text commentary below. This will also eventually add gif support on hover.
Screen Shot 2021-01-08 at 11 49 45 AM


guerler:
Thanks for working on this. I think, the icons seem to be a little bit too large in comparison with the text: "Galaxy is a web-based..."

dannon:
To document it here; the next step for getting this integrated and deployed is going to depend on webpack's module federation, which is going to require a (separate PR) version bump to webpack5.

dannon (approved):
I want to follow up with some work to make the inclusion into the app a bit more robust, but it's a functional start that we can iterate on.  And as mentioned, we want to completely untangle the build here and use webpack5's federated modules when we're able to get that working.

client/src/components/NewUserWelcome/NewUserWelcome.vue
```diff
@@ -0,0 +1,177 @@
...
+            this.position.pop();
+        },
+        down(index) {
+            this.position.push(index);
+        },
+    },
+};
+</script>
+
+<style type="text/css">
```
guerler, line 88: These custom styles conflict with theme settings. I think we need to remove them.

PR #11089, opened 2021-01-11
Add FastAPI routes for tours

jdavcs:
Ref #10889 

@mvdbeek, @davelopez: please take a look - here's my first attempt.

Note: A pydantic `TourList` model (as opposed to `List[Tour]`) was necessary for correct serialization in `web/framework/decorators.py`. Even though we can modify [this logic](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/web/framework/decorators.py#L342) to recognize a list of pydantic models, such a list cannot be serialized using the `[model].json()` method.


lib/galaxy/webapps/galaxy/api/tours.py
```diff
@@ -3,16 +3,52 @@
...
 
 log = logging.getLogger(__name__)
 
 
+router = APIRouter(tags=['tours'])
+
+
+@cbv(router)
+class FastAPITours:
+    app: UniverseApplication = Depends(get_app)
```
davelopez: Maybe something more specific here like:
```suggestion
    tour_registry: Registry = Depends(get_tour_registry)
```

where `get_tours_registry` is:

`````python
def get_tour_registry(app: UniverseApplication = Depends(get_app)) -> Registry:
    return app.tour_registry
````

File: lib/galaxy/schema/__init__.py [hunk 2]
```diff
@@ -1,13 +1,44 @@
...
 
     def all_roles(*args) -> list:
         return []
+
+
+class Tour(BaseModel):
+    id: str
+    name: str
+    description: str
+    tags: List[str]
```
davelopez: Will the documentation of the fields (for the OpenAPI docs) be added in a separate PR?

lib/galaxy/schema/__init__.py
```diff
@@ -1,13 +1,44 @@
...
+
+
+class TourList(BaseModel):
+    __root__: List[Tour] = []
+
+
+class TourStep(BaseModel):
+    title: Optional[str] = None
```
mvdbeek: Can you add a description to these fields ?
Something like
```suggestion
    title: Optional[str] = Field(title="Title", description="The title of this tour step", example="Click here")
```
?

lib/galaxy/schema/__init__.py
```diff
@@ -1,13 +1,44 @@
...
-    preferences: typing.Dict[str, str] = {}
+    email: Optional[str] = None
+    preferences: Dict[str, str] = {}
     bootstrap_admin_user = True
 
     def all_roles(*args) -> list:
         return []
+
+
+class Tour(BaseModel):
```
davelopez: Should we put all the API BaseModels for all the routes under `schema/__init__.py` or should they be in their own file like `schema/tours.py`? I am still struggling with Python packages/modules/__init__ best practices 😅

mvdbeek: Since those are probably only used with tours I'd probably put this with `lib/galaxy/tours` ?

davelopez (commented):
Looking great!

lib/galaxy/schema/__init__.py
mvdbeek: They should be with the tours, since we won't re-use the model in other places, so lib/galaxy/tours/schema.py seems like a good idea.

lib/galaxy/tours/__init__.py
```diff
@@ -23,20 +26,24 @@ def tour_loader(contents_dict):
...
+        tours = []
+        for k in self.tours.keys():
+            tourdata = {
+                'id': k,
+                'description': self.tours[k].get('description'),
+                'name': self.tours[k].get('name'),
+                'tags': self.tours[k].get('tags')
+            }
+            tours.append(tourdata)
+        return parse_obj_as(TourList, tours)
```
davelopez: I think you can use here:
```suggestion
        return parse_obj_as(List[Tour], tours)
```
In case you want to use the `List` typing instead of a model for the list. But not 100% sure :)

config/plugins/tours/core.history.yaml
```diff
@@ -1,3 +1,4 @@
+id: galaxy_history
```
jmchilton: Can you explain adding the id to the tours here? Does this get propagated somehow? It looks like the id was calculated in a different way before?

lib/galaxy/schema/__init__.py
jdavcs: That was my first version - until the last PR.
Moving to `tours/schema.py` (as per discussion below)

jdavcs: Done! `tours/schema.py`.

lib/galaxy/tours/__init__.py
jdavcs: Unless I'm mistaken, we can't. That was my first approach, but I ran into the issue I mention in the PR description: this has to be serialized to json, and we cannot use pydantic's `model.json()` if it's a list of pydantic models: we can only serialize each model. To serialize the list you need to use `json.dumps()` (or our `safe_dumps` - same thing) - but `json.dumps()` doesn't know how to handle a model. Thus, using a pydantic model which is a list of models seems to be the most straightforward way to implement this. (It's different for the licenses endpoint where the response model is a list because we don't serialize that list of models, at least for now)

config/plugins/tours/core.history.yaml
jdavcs: The id is calculated based on the filename [here](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tours/__init__.py#L73), so the id from the yaml file is not used at all for tour loading. I checked the client ([client/src/mvc/tours.js](https://github.com/galaxyproject/galaxy/blob/dev/client/src/mvc/tours.js#L43) appears to be the only relevant file) - same there. I considered deleting the id field, but instead decided to add id values to the other 2 tours for consistency, in case there may have been plans for that field. I don't mind either deleting the field or reverting my edit and leaving as is, but having that field optional (given that we formalize it in a pydantic model) doesn't seem right, I think.

lib/galaxy/schema/__init__.py
jdavcs: @davelopez I can add it now. What are the required fields? (I'm looking at OpenAPI specs..) Is it just title, description, example?

davelopez: Yes, those seem to be the most used for documentation, you can also define some restrictions about the fields with the other properties (`ge`, `max_length`, etc.) if it is required.

config/plugins/tours/core.history.yaml
jmchilton: Thanks for looking into that - so it seems it isn't used in a functional capacity anywhere? Want to remove it from core.galaxy_ui.yaml and see if anyone complains? I think we should remove it if it is not being used.

jdavcs: Done! Completely agree on removing dead code.

lib/galaxy/schema/__init__.py
jdavcs: Done.

lib/galaxy/schema/__init__.py
jdavcs: Done.

lib/galaxy/webapps/galaxy/api/tours.py
jdavcs: Done.

ic4f:
The [latest commit](https://github.com/galaxyproject/galaxy/commit/f1274c22caf73ccfe0bdb3cb2d4312862e887033) addresses [this comment in the code review](https://github.com/galaxyproject/galaxy/pull/11089#discussion_r554902286). 

However, this turned out to be a good excuse to get rid of the concrete ToursRegistry class in the api module and replace it with an interface. This follows [this comment](https://github.com/galaxyproject/galaxy/pull/10823#discussion_r545213498); @jmchilton - does this look like the approach you were describing?

ic4f:
Note about naming the abstract class: I struggled with the naming; with `ToursRegistry` as the concrete class, there was nothing that felt right for the abstract class: it's a virtual class, it's not inherited, so calling it `ToursRegistryBase` made no sense. Besides, I don't want to "signal" to the client module that it's an abstract class (even more so for names like `IToursRegistry`). Yes, "abc" shadows the standard library "abc" package, but, still, I think it makes sense to have `foo.Foo` and `foo.abc.Foo` just like we now might have `foo.schema.Foo`. But I'm, of course, by completely open to other suggestions.

jmchilton:
I love the abc approach. In terms of naming stuff my suggestion would be to...

- Refactor``lib/galaxy/tours/__init__.py::ToursRegistry`` -> ``lib/galaxy/tours/_impl.py::ToursRegistryImpl``
- Add a ``def build_tours_registry(config_dir: str) -> ToursRegistry`` factory to ``_impl.py``.
- Refactor ``lib/galaxy/tours/abc/__init__.py::ToursRegistry`` -> ``lib/galaxy/tours/interface.py::ToursRegistry``
- Make ``lib/galaxy/tours/__init__.py`` a file that just imports ``._impl.py::build_tours_registry``, ``.schema.py::(TourStep,Tour,TourDetails,etc..)``, and ``.interface.py::ToursRegistry`` and re-exports the whole module's interface.

At that point I would even consider renaming ``schema.py`` and ``interface.py`` to ``_schema.py`` and ``_interface.py`` to make it clear the interface is the stuff that is exported from ``galaxy.tours`` 's root ``__init__.py``.

lib/galaxy/webapps/galaxy/api/__init__.py
```diff
@@ -85,3 +86,7 @@ def get_admin_user(trans: SessionRequestContext = Depends(get_trans)):
     if not trans.user_is_admin:
         raise AdminRequiredException(require_admin_message(trans.app.config, trans.user))
     return trans.user
+
+
+def get_tours_registry(app: UniverseApplication = Depends(get_app)) -> ToursRegistry:
```
davelopez: Not sure, maybe we should move this function to `lib/galaxy/webapps/galaxy/api/tours.py`?. It seems probable that it will be used only in the `tours` route.

davelopez (commented):
Cool! I like your approach of using an ABC for the ToursRegistry :)

lib/galaxy/webapps/galaxy/api/__init__.py
jdavcs: Done. I didn't include the `UniverseApplication` annotation to avoid importing it into `api/tours.py` (it's already annotated in `get_app`)

ic4f:
> I love the abc approach. In terms of naming stuff my suggestion would be to...

@jmchilton thank you for the detailed suggestions - I think this makes the tours interface much cleaner.

The factory function does not have a return type annotation - that's intentional. That's related to mypy not handling `ABC.register` - I'll post an issue shortly. 

This is (again!) ready for review.

lib/galaxy/tours/__init__.py
```diff
@@ -1,87 +1,7 @@
...
-        except TypeError:
-            log.exception("Tour '%s' could not be loaded, error within file. Possibly spacing related. Please check your yaml syntax.", tour_id)
-        return None
+from ._impl import build_tours_registry  # noqa
+from ._interface import ToursRegistry  # noqa
+from ._schema import Tour  # noqa
+from ._schema import TourCore  # noqa
+from ._schema import TourStep  # noqa
+from ._schema import TourDetails  # noqa
+from ._schema import TourList  # noqa
```
jmchilton: Instead of ``#noqa`` please use ``__all___`` at the end of the file.

```
__all__ = ("build_tours_registry", "Tour", ...)
```

jdavcs: Done

jmchilton: Thanks!

jmchilton (approved):
Amazing! Thanks for the deep dive into this.

PR #11093, opened 2021-01-11
Fix type column in library folder

### LINKED ISSUE #10997: 20.09 regression: Data Type column missing/replaced in data library UI
In 20.05 the data library UI had a `Data Type` column (e.g. https://usegalaxy.fr/library/list#folders/Fb58ecc291a94c03f from usegalaxy.fr which is still at 20.05) which made it possible to sort datasets by format. For folders the value of this column was `folder`.
In 20.09 this was replaced by the `Type` column (e.g. https://usegalaxy.eu/library/folders/Fa21272e5bd712216 ), which is just either `folder` or `file`, basically redundant with the icon on the very left of each line.

Can we have the data type back please?

Ping @OlegZharkov

OlegZharkov:
Fixes https://github.com/galaxyproject/galaxy/issues/10997

### DIFF
--- a/client/src/components/LibraryFolder/LibraryFolder.vue
+++ b/client/src/components/LibraryFolder/LibraryFolder.vue
@@ -153,6 +153,10 @@
                     <font-awesome-icon v-if="row.item.type === 'folder'" :icon="['far', 'folder']" title="Folder" />
                     <font-awesome-icon v-else-if="row.item.type === 'file'" title="Dataset" :icon="['far', 'file']" />
                 </template>
+                <template v-slot:cell(type)="row">
+                    <div v-if="row.item.type === 'folder'">{{ row.item.type }}</div>
+                    <div v-else-if="row.item.type === 'file'">{{ row.item.file_ext }}</div>
+                </template>
                 <template v-slot:cell(raw_size)="row">
                     <div v-if="row.item.type === 'file'" v-html="bytesToString(row.item.raw_size)"></div>
                 </template>


client/src/components/LibraryFolder/LibraryFolder.vue
```diff
@@ -153,6 +153,10 @@
                     <font-awesome-icon v-if="row.item.type === 'folder'" :icon="['far', 'folder']" title="Folder" />
                     <font-awesome-icon v-else-if="row.item.type === 'file'" title="Dataset" :icon="['far', 'file']" />
                 </template>
+                <template v-slot:cell(type)="row">
+                    <div v-if="row.item.type === 'folder'">{{ row.item.type }}</div>
+                    <div v-else-if="row.item.type === 'file'">{{ row.item.file_ext }}</div>
+                </template>
```
nsoranzo, line 159: Should we merged these 2 columns? For a folder it could be the icon, for a dataset its datatype.

OlegZharkov: If we merge them, the folder's type would be empty

OlegZharkov: Currently:
image
if we merge and make both show `file_ext`. 
image

If you want, I can change `row.item.file_ext` to just hardcoded string 'folder', if it makes more sense

nsoranzo: Sorry, I still don't understand Vue much and misunderstood what the code was doing, it's fine as it is then! Thanks for the screenshots.

nsoranzo (approved):

OlegZharkov:
thanks @nsoranzo @jmchilton

nsoranzo:
Thank you for the fix!

PR #11098, opened 2021-01-11
Add FastAPI routes for datatypes

davelopez:
Ref #10889

Another route migrated to FastAPI. This time for `datatypes` operations.

## Some comments and doubts about the migration:
- The pydantic models have been placed in `lib/galaxy/datatypes/schema.py` as commented [here](https://github.com/galaxyproject/galaxy/pull/11089#discussion_r554907211). I am not completely sure if, in this case, this is the best place since that package mainly contains data type definitions (and in the future may be included a `schema` data type that could collide?)
- I am not very happy with my naming skills... 😅 so if you have a better idea for `CompositeFileInfo`, `DatatypesMap`, and the other pydantic models, please make a suggestion :)
- Also I find it a bit hard to document some of the fields and descriptions with meaningful and grammatically correct explanations (English is not my mother-tongue). So any help with that is much appreciated.
- As a first approach and to leave the old API controller intact, I ended up duplicating (adding type annotations) some of the [internal methods](https://github.com/galaxyproject/galaxy/compare/dev...davelopez:fastapi-migration-datatypes?expand=1#diff-7d2152a780aa7c049cf34735fcc435ee1a462dbe4761856524bb87700cedbb17R144) of the legacy controller. I am not proud of it 😓, so maybe I should extract a new `manager` into `lib/galaxy/managers/datatype.py` or just assume that, at some point, the legacy controller will be removed/replaced by the new one?

## Some comments about the current `datatypes` API:
- The `/api/datatypes/mapping` endpoints return in the `class_to_classes` mapping a dictionary that seems to have always the values set to `true`, not sure what this `true` value means. If it is redundant, converting it to a simple list could simplify and reduce the response body a bit. But it's not a big deal.
- The index endpoint [`/api/datatypes`](https://github.com/galaxyproject/galaxy/compare/dev...davelopez:fastapi-migration-datatypes?expand=1#diff-7d2152a780aa7c049cf34735fcc435ee1a462dbe4761856524bb87700cedbb17R61) returns a type `Union` of two types because the endpoint behaves differently depending on the optional query params. This kind of response is valid and works OK, but the swagger documentation does not support it. One possible solution could be removing the `extension_only` query parameter and adding a new `/api/datatypes/extensions/?upload_only=true` endpoint. This would change/break the API. It is not a big deal either, but the documentation gives the impression that there is an error when there is none:
image


lib/galaxy/webapps/galaxy/api/datatypes.py
```diff
@@ -2,14 +2,200 @@
...
+        response_model=Union[List[DatatypeDetails], List[str]],
+        response_description="List of data types")
+    async def index(self,
+                    extension_only: Optional[bool] = ExtensionOnlyQueryParam,
+                    upload_only: Optional[bool] = UploadOnlyQueryParam
+                    ) -> Union[List[DatatypeDetails], List[str]]:
+        """Gets the list of all available data types."""
+        try:
+            return self._index(extension_only, upload_only)
+        except Exception as e:
```
jmchilton: Can we do this without catching and rethrowing the exception? It is a lot of extra untested code - if we want all these exceptions logged like this we can do it at the framework level (https://github.com/galaxyproject/galaxy/pull/10823/files#diff-ef136e8d148bc95da2725c2cc8e228e60cffeeb0b84820f7152646d11ced9ff5R11).

jmchilton: I see - sometimes the legacy controllers do this also - it is bad when they do it also. Can you remove the exception handling down there also?

lib/galaxy/webapps/galaxy/api/datatypes.py
```diff
@@ -2,14 +2,200 @@
...
+    async def edam_data(self) -> Dict[str, str]:
+        """Gets a map of datatypes and their corresponding EDAM data."""
+        try:
+            return self.datatypes_registry.edam_data
+        except Exception as e:
+            log.exception('Could not get EDAM data mappings')
+            raise e
+
+    def _index(self, extension_only: Optional[bool] = True, upload_only: Optional[bool] = True) -> Union[List[DatatypeDetails], List[str]]:
+        if extension_only:
```
jmchilton: I would appreciate it if you can convert these to like view functions that can be reuse in the existing controller below.

I know we plan to eliminate those controllers but I don't think we should duplicate hundreds of lines of code even in the short term in case those plans don't pan out. I added a bunch of stuff to https://github.com/galaxyproject/galaxy/pull/10823 so that those controllers can return pydantic models and exceptions handling is synchronized.

If the old code and the new code are both as thin as possible - those are ideal controllers IMO and they will allow us to duplicate as little as possible.

jmchilton:
```
__________ ERROR collecting packages/data/galaxy/datatypes/schema.py ___________
galaxy/datatypes/schema.py:3: in <module>
    from pydantic import (
E   ModuleNotFoundError: No module named 'pydantic'
=============================== warnings summary ===============================
/tmp/gxpkgtestenvLFfwjN/lib/python3.6/site-packages/_pytest/config
```

I think pydantic needs to be add to ``packages/data/requirements.txt``.

lib/galaxy/webapps/galaxy/api/datatypes.py
davelopez: Sure!

lib/galaxy/webapps/galaxy/api/datatypes.py
davelopez: Of course! that was one of my concerns in the PR comment. Thanks for the suggestions, I will do it :)

davelopez:
Thanks @jmchilton! 

>I think pydantic needs to be add to packages/data/requirements.txt.

I ended up moving `/datatypes/schema.py` to `/managers/datatype.py` because I was feeling uneasy anyway about what I explained in my initial comment:
>The pydantic models have been placed in lib/galaxy/datatypes/schema.py as commented here. I am not completely sure if, in this case, this is the best place since that package mainly contains data type definitions (and in the future may be included a schema data type that could collide?)
---
I also added [here](https://github.com/galaxyproject/galaxy/pull/11098/commits/56d5822862d70d0d1dd3219dd27b63e791524a46) the view functions to be reused in the legacy and new routes. Do you think they should go to a manager class? I am still figuring out how everything should be laid out in the codebase.

Please let me know what else can be changed/improved :)

lib/galaxy_test/api/test_datatypes.py
```diff
@@ -84,6 +84,20 @@ def test_converter_present_after_toolbox_reload(self):
         self._assert_status_code_is(response, 200)
         assert converters == len(response.json())
 
+    def test_edamp_formats(self):
```
mvdbeek:
```suggestion
    def test_edam_formats(self):
```
I assume that's a typo ?

davelopez: Absolutely :) thanks for the catch

mvdbeek:
> I ended up moving `/datatypes/schema.py` to `/managers/datatype.py` because I was feeling uneasy anyway about what I explained in my initial comment:
> 
> > The pydantic models have been placed in lib/galaxy/datatypes/schema.py as commented here. I am not completely sure if, in this case, this is the best place since that package mainly contains data type definitions (and in the future may be included a schema data type that could collide?)

I think it's fine the way you have done it here now, but to clarify this, if we ever have a schema datatype, the datatype code can be loaded from any file. Having the pydantic models in the datatypes module would be useful if we wanted to validate datatype definitions in the registry itself ... we don't do that currently, so I think that's fine for now, and we can always change that later.
There was also the idea to move schema.py to _schema.py, maybe we could generalize that ?

> * but the swagger documentation does not support it.

Interesting, what do you see in the generated JSON spec ? https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/#oneof seems like it should handle this case

I do like `/api/datatypes/extensions` in principle, but I think we could also do that once we drop the old routes.

davelopez:
>There was also the idea to move schema.py to _schema.py, maybe we could generalize that ?

Yes, It would be awesome to have some document with all these ideas to unify and generalize them.

>Interesting, what do you see in the generated JSON spec ?

````json
"responses": {
    "200": {
      "description": "List of data types",
      "content": {
        "application/json": {
          "schema": {
            "title": "Response Index Api Datatypes Get",
            "anyOf": [
              {
                "type": "array",
                "items": {
                  "$ref": "#/components/schemas/DatatypeDetails"
                }
              },
              {
                "type": "array",
                "items": {
                  "type": "string"
                }
              }
            ]
          }
        }
      }
    },
    "422": {
      "description": "Validation Error",
      "content": {
        "application/json": {
          "schema": {
            "$ref": "#/components/schemas/HTTPValidationError"
          }
        }
      }
    }
  }
````
And the error in the console displays: `TypeError: Cannot read property 'anyOf' of undefined`

I will investigate further to see if I find out what is causing the problem.

**Comment by davelopez** [comment]
Sorry for the rebase mess 😅

File: lib/galaxy/webapps/galaxy/api/datatypes.py (hunk 2 above)
Comment (davelopez) [reply lib/galaxy/webapps/galaxy/api/datatypes.py]: Done :)

**Comment by davelopez** [comment]
I think this is ready for review again :)

**Comment by mvdbeek** [comment]
We'll need pydantic again in the data package :)

File: lib/galaxy/webapps/galaxy/api/datatypes.py [hunk 4]
```diff
@@ -2,14 +2,148 @@
...
+        response_model=Union[List[DatatypeDetails], List[str]],
+        response_description="List of data types",
+    )
+    async def index(
+        self,
+        extension_only: Optional[bool] = ExtensionOnlyQueryParam,
+        upload_only: Optional[bool] = UploadOnlyQueryParam
+    ) -> Union[List[DatatypeDetails], List[str]]:
+        """Gets the list of all available data types."""
+        return view_index(self.datatypes_registry, extension_only, upload_only)
```
jmchilton, line 76: These little one line methods shared with the older style controller are perfect!

jmchilton:
Wonderful, thanks for being so responsive to feedback.

PR #11101, opened 2021-01-12
More Sphinx Work

jmchilton:
- Redo the way API docs are generated - just link to regular pydocs instead of duplicating all of the pydocs.
- Add docstrings to api_asserts.py.
- Allow linking to requests to using intersphinx.
- Some new environment variables to speed up doc building.
- Fix a ton of sphinx warnings.

### DIFF
--- a/doc/source/admin/mq.md
+++ b/doc/source/admin/mq.md
@@ -1,3 +1,7 @@
+```eval_rst
+:orphan:
+```
+
 # Message queue
 
 Galaxy uses a message queue internally for communicating between processes. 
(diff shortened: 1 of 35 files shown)


doc/source/admin/mq.md
```diff
@@ -1,3 +1,7 @@
+```eval_rst
+:orphan:
```
nsoranzo, line 2: Should we add it admin or special topics index page?

mvdbeek:
Is there anything in here that isn't in https://github.com/galaxyproject/galaxy/pull/11103 ?

jmchilton:
Nope.

PR #11102, opened 2021-01-12
Update coverage documentation in run_tests.sh

davebx:


run_tests.sh
```diff
@@ -149,6 +149,9 @@ Extra options:
  --verbose_errors      Force some tests produce more verbose error reporting.
  --no_cleanup          Do not delete temp files for Python functional tests
                        (-toolshed, -framework, etc...)
+ --coverage            Generate a test coverage report. This option currently
+                       should work with every test that uses pytest, but the
+                       results may not be reliable for e.g. selenium tests.
```
jdavcs:
```suggestion
                       results may not be reliable, e.g. selenium tests.
```
Very minor. "for e.g." sounds redundant, I think (e.g. *is* "for example").

davebx: Good point.

jdavcs (approved):

ic4f:
I like the updated wording a lot more! (than the suggestion)

PR #11103, opened 2021-01-12
Progress toward failing CI on sphinx warnings.

jmchilton:
Builds on #11101 with a ton more sphinx fixes... in fact fixes all the warnings I think.

Conflicting this in some ways, I've got another branch that documents Pydantic models with like all their annotated data (https://github.com/jmchilton/galaxy/pull/new/pydantic_sphinx) but the plugin it uses seems a bit buggy and generates a bunch of new warnings. I guess we could bring that in and use a custom script to parse sphinx output the way Planemo does so we can filter specific problems.


.github/workflows/docs.yaml
```diff
@@ -1,5 +1,7 @@
 name: Build docs
 on: [push, pull_request]
+env:
+  SPHINXOPTS: '-W'
```
nsoranzo: Should we move this to `doc/Makefile` so that local doc builds are the same of the CI ones?

doc/source/releases/21.01_announce.rst
```diff
@@ -1,3 +1,4 @@
+:orphan:
```
nsoranzo, line 1: Or add it to doc/source/releases/index.rst (see comment below though) ?
Or add a stub of 21.01_announce_user that links to it (and maybe script this in `scripts/bootstrap_history.py` ).

doc/source/releases/index.rst
```diff
@@ -37,3 +37,15 @@ Releases
    13.02_announce
    13.01_announce
    older_releases
+
+User Announcements
+
+.. toctree::
+   :maxdepth: 1
+
+   20.09_announce_user
```
nsoranzo: I don't like the duplication of announcements, we could replace the XX.XX_announce with XX.XX_announce_user above for the releases that have them.

jmchilton: These are developer release notes right? Don't we intentionally not lead with the user release notes? If we wanted to lead with user release notes - we should just not produce developer release notes?

If you don't link these links here - what about another page with the user release notes?

.github/workflows/docs.yaml
jmchilton: I definitely don't love putting logic in this file. For development purposes - it is pretty frustrating that it only shows one warning at a time though. I wish it were implemented differently. For development purposes I think printing all the warnings is essential. I think it might be worth implementing this a different way than using -W. Hmm...

nsoranzo: I suppose that once this is merged, only new warnings introduced by someone's changes will fail the build. Not ideal but not too bad?

doc/source/releases/index.rst
nsoranzo: > These are developer release notes right?

By looking at the docs menu, that's the general (and only at the moment) release notes page, if it was for developer release notes it should go under "Developer Documentation", I think.

> Don't we intentionally not lead with the user release notes?

Maybe, I don't recall the discussion though, it could be a good time to rethink given the increasing effort on the user RNs.

> If we wanted to lead with user release notes - we should just not produce developer release notes?

I think this page should list the user release notes (`XX_announce` for 18.09 and earlier), and all user release notes already link to the detailed/dev-oriented release notes at the bottom.

Ping @hexylena

hexylena: I didn't push them as a replacement since they're different audiences, and I think devs like reading the dev notes for a rollup of the past N months work.

That said I can definitely see the duplication in the menu being suboptimal.

I guess the question is: are users (or in general people who would benefit from seeing these) finding their way to docs.galaxyproject.org? I'm not sure at all.  They are mentioned in the dev docs, but, argh.

nsoranzo: > are users (or in general people who would benefit from seeing these) finding their way to docs.galaxyproject.org? I'm not sure at all. They are mentioned in the dev docs, but, argh.

We are now directly linking to the user release notes from the Galaxy Help menu: https://github.com/galaxyproject/galaxy/pull/10960

.github/workflows/docs.yaml
jmchilton: Ummm... https://github.com/galaxyproject/galaxy/pull/11103/commits/9dd73c3590bab5e6948cd76282b55958e4d13009  maybe?

nsoranzo: Good for me!

doc/source/releases/index.rst
jmchilton: I don't like it in the abstract but I don't feel strongly enough to fight on this so https://github.com/galaxyproject/galaxy/pull/11103/commits/ba7286ef45c6bc9420d55f163ce126c45826119a. Also what I do feel strongly about is the last release's user release notes were shockingly good. I hadn't seen them before yesterday and I was blown away by the effort that must have been put into them - so I have like making them more prominent in the abstract.

doc/source/admin/mq.md
```diff
@@ -1,3 +1,7 @@
+```eval_rst
+:orphan:
```
nsoranzo: Should we add this to the admin or special topics index page instead? (comment moved from https://github.com/galaxyproject/galaxy/pull/11101 , probably easier to do all reviewing here).

jmchilton: Done.

doc/source/releases/21.01_announce.rst
nsoranzo: That's the only outstanding comment from me, I guess that also this line needs to be updated: https://github.com/galaxyproject/galaxy/blob/dev/scripts/bootstrap_history.py#L382

jmchilton:
Already a doc issue in the merge commit 😆... and of course from that careless @jmchilton

doc/source/releases/index.rst
jmchilton:
```
/Users/john/workspace/galaxy/doc/source/releases/19.05_announce.rst: WARNING: document isn't included in any toctree
/Users/john/workspace/galaxy/doc/source/releases/19.09_announce.rst: WARNING: document isn't included in any toctree
/Users/john/workspace/galaxy/doc/source/releases/20.01_announce.rst: WARNING: document isn't included in any toctree
/Users/john/workspace/galaxy/doc/source/releases/20.05_announce.rst: WARNING: document isn't included in any toctree
/Users/john/workspace/galaxy/doc/source/releases/20.09_announce.rst: WARNING: document isn't included in any toctree
```

Well that didn't even fix that problem 😓 . This PR got away from itself.

nsoranzo: So should we mark as orphan the XX.XX_announce.rst files for 19.01 and following ?

doc/source/releases/21.01_announce.rst
jmchilton: > Or add it to doc/source/releases/index.rst (see comment below though) ?

This is in a commit description but the whole reason I generate this page is to link it into the release index but then there was pushback against that maybe a year ago. I think we should either link it or stop pre-generating it - I have a vague preference but either are better than just tagging this as an orphan I think. Seems like maybe we should do that in a follow up PR specifically around these release page links though right? These seem to be more controversial than I had anticipated and I've already rolled back the strict checking so we have more freedom to address these issues in a follow up PR.

doc/source/releases/index.rst
jmchilton: Same comment as above - I think we should merge this while it is green and re-tackle release issue stuff in a clean PR focused on that.

PR #11111, opened 2021-01-12
really test variables in conditionals

bernt-matthias:
Was just wondering why this test was passing with wrong cheetah .. needed a while before I realized that actual test cases were missing. 

Edit: the tool is used here https://github.com/galaxyproject/galaxy/blob/3bed882331dd0d45b4a9a1c3f9516685fc22be78/lib/galaxy_test/api/test_tools.py#L1595 .. wondering why this was working...

Could remove the tests from the xml

### DIFF
--- a/test/functional/tools/identifier_in_conditional.xml
+++ b/test/functional/tools/identifier_in_conditional.xml
@@ -1,13 +1,6 @@
 <tool id="identifier_in_conditional" name="identifier_in_conditional" version="1.0.0">
     <command><![CDATA[
 #if str($outer_cond.multi_input) == 'true':
-    #for $input in $input1#
-        echo '$input.element_identifier' >> 'output1';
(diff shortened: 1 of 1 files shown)


test/functional/tools/identifier_in_conditional.xml
```diff
@@ -1,7 +1,11 @@
 <tool id="identifier_in_conditional" name="identifier_in_conditional" version="1.0.0">
     <command><![CDATA[
+## note access to input1 is supposed to work
+## - qualified ($outer_cond.input1)
+## - and unqualified ($input1)
+## both are tested here but the later should be avoided
```
nsoranzo:
```suggestion
## both are tested here but the latter should be avoided
```

nsoranzo (approved):

PR #11113, opened 2021-01-12
Establish a best practice for trans dependency management.

jmchilton:
As part of the ongoing effort to establish a best practice for modern backend components - I think this is important step. I could explain it all but I think the docstring at the top of the PR does a pretty good job explaining why this is a good direction.

### DIFF
--- a/lib/galaxy/managers/context.py
+++ b/lib/galaxy/managers/context.py
@@ -1,6 +1,40 @@
+"""Interfaces/mixins for transaction-like objects.
+
+These objects describe the context around a unit of work. This unit of work
+is very broad and can be anything from the response to a web request, the
+scheduling of a workflow, the reloading the toolbox, etc.. Traditionally,
+Galaxy has simply passed around a GalaxyWebTransaction object through
+all layers and large components of the Galaxy app. Having random backend
+components define explicit dependencies on this however is inappropriate
+because Galaxy may not be used in all sort of non-web contexts. The future
+use of message queues and web sockets as well as the decomposition of the
+backend into packages only further make this heavy reliance on
+GalaxyWebTransaction inappropriate.
+
+A better approach is for components to annotate their reliance on much
+narrower, typed views of the GalaxyWebTransaction. This allows explicit
+declaration of what is being required in the context of a method or class
+and allows the Python type system to ensure the transaction supplied to
+the method is appropriate for the context. For instance, an effective
+use of the type system in this way can prevent the backend work context
+used to schedule workflow from being supplied to a method that requires
+an older-style WSGI web response object.
+
+There are various levels of transactions defined in this file - these
+include :class:`galaxy.managers.context.ProvidesAppContext`,
+:class:`galaxy.managers.context.ProvidesUserContext`,
+and :class:`galaxy.managers.context.ProvidesHistoryContext`. Methods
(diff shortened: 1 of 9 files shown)


lib/galaxy/managers/context.py
```diff
@@ -1,6 +1,39 @@
...
+
+These objects describe the context around a unit of work. This unit of work
+is very broad and can be anything from the response to a web request, the
+scheduling of a workflow, the reloading the toolbox, etc.. Traditionally,
+Galaxy has simply passed around a GalaxyWebTransaction object through
+all layers and large components of the Galaxy app. Having random backend
+components define explicit dependencies on this however is inappropriate
+because Galaxy may not be used in all sort of non-web contexts. The future
+use of message queues and web sockets as well as the decomposition of the
+backend into packages all only further make this heavy reliance on
```
nsoranzo:
```suggestion
backend into packages only further make this heavy reliance on
```

lib/galaxy/managers/context.py
```diff
@@ -99,14 +140,18 @@ def install_model(self):
...
+    """ For transaction-like objects to provide Galaxy convenience layer for
     reasoning about users.
 
     Mixed in class must provide `user` and `app`
     properties.
     """
 
+    @abc.abstractproperty
+    def user(self):
+        """Provide access to the app object."""
```
nsoranzo:
```suggestion
        """Provide access to the user object."""
```

nsoranzo:
The py36-unit build failure seems relevant.

lib/galaxy/webapps/galaxy/api/roles.py
```diff
@@ -65,26 +63,26 @@ class FastAPIRoles:
     role_manager: RoleManager = Depends(get_role_manager)
 
     @router.get('/api/roles')
-    def index(self, trans: SessionRequestContext = Depends(get_trans)) -> RoleListModel:
+    def index(self, trans: ProvidesUserContext = Depends(get_trans)) -> RoleListModel:
```
jdavcs, line 66: Maybe move the `RoleListModel` definition (line 43) to `galaxy/manager/roles` and import it alongside the other pydantic models?

jmchilton: I get why that seems more wholistic and consistent. My counter argument is that I think the distinction is that RoleListModel exists only to adapt a list of roles to the web framework - the managers themselves are more useful and better when they return a list of role models. So it feels like moving that there is pushing "web app concerns" into "app" logic.

If you don't buy that though, I'll definitely refactor this to and push that down into the managers.

jdavcs: No, I agree, you're right. In fast, I had similar misgivings about the `TourList` model being part of the tours "schema". I think once we have a good sample of endpoints done, we may go back and refactor for consistency.

jdavcs (approved):

lib/galaxy/webapps/galaxy/api/roles.py
jmchilton: Finding the right distinction here is messy - I think we're asking the right questions and things are structured better than ever before inside the app - but we will need to work on this and accept a lot of fuzziness in these distinctions.

Thanks for the detailed review!

Issue #10884, opened 2020-12-10
Galaxy configuration mechanism for WSGI/ASGI apps


mvdbeek:
Most wsgi/asgi apps can be controlled by any number of servers (uvicorn/gunicorn/hypercorn), where typically you only need to indicate the object that is the actual asgi/uwsgi app (e.g. for the app instance in app.py:  `uvicorn app:app`).
https://github.com/galaxyproject/galaxy/pull/10823 adds a script that starts galaxy using `python scripts/fapi.py -c config/galaxy.yml`.

We should be able to run Galaxy with `GALAXY_CONFIG_FILE=/path/to/config uvicorn galaxy:app`, where GALAXY_CONFIG_FILE defaults to config/galaxy.yml

PR #11116, opened 2021-01-13
uvicorn-compatible factory function for fastapi.

jmchilton:
Implement https://github.com/galaxyproject/galaxy/issues/10884?

I'm sure we could drop the factory argument and just make app a global... but I do hate that style of code. I think everything in lib should be importable without side effects.

Update: I've rebased this with instructions for running via gunicorn which doesn't expose that factory argument to convince myself this can be done without globals. It seems to work.

### DIFF
--- a/lib/galaxy/main_config.py
+++ b/lib/galaxy/main_config.py
@@ -1,0 +1,36 @@
+"""Utilities for finding Galaxy's configuration file.
+
+This is for use by web framework code and scripts (e.g. scripts/galaxy_main.py).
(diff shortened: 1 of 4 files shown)


mvdbeek:
This is fantastic, I think that's even better than what I had in mind! Still WIP ?

jmchilton:
I think it needs some work in a few directions but I think it is ready to be merged, it would allow more people to hack on a more production style FastAPI setup I think.

PR #11124, opened 2021-01-14
Dynamic discovery of FastAPI routers.

jmchilton:
Reduce the burden/steps of implementing new controllers to match level of dynamism we have for existing controllers (web and API). 

Screen Shot 2021-01-13 at 7 00 28 PM

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/job_lock.py
+++ b/lib/galaxy/webapps/galaxy/api/job_lock.py
@@ -11,12 +11,12 @@ router = APIRouter(tags=['job_lock'])
 
 
 @router.get('/api/job_lock')
-def job_lock_status(job_manager=Depends(get_job_manager), admin_user=Depends(get_admin_user)) -> JobLock:
+async def job_lock_status(job_manager=Depends(get_job_manager), admin_user=Depends(get_admin_user)) -> JobLock:
     """Get job lock status."""
     return job_manager.job_lock()
 
 
 @router.put('/api/job_lock')
-def update_job_lock(job_manager=Depends(get_job_manager), admin_user=Depends(get_admin_user), job_lock: JobLock = Body(...)) -> JobLock:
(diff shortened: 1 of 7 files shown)


lib/galaxy/webapps/galaxy/api/job_lock.py
```diff
@@ -11,12 +11,12 @@
...
 @router.get('/api/job_lock')
-def job_lock_status(job_manager=Depends(get_job_manager), admin_user=Depends(get_admin_user)) -> JobLock:
+async def job_lock_status(job_manager=Depends(get_job_manager), admin_user=Depends(get_admin_user)) -> JobLock:
     """Get job lock status."""
     return job_manager.job_lock()
 
 
 @router.put('/api/job_lock')
-def update_job_lock(job_manager=Depends(get_job_manager), admin_user=Depends(get_admin_user), job_lock: JobLock = Body(...)) -> JobLock:
+async def update_job_lock(job_manager=Depends(get_job_manager), admin_user=Depends(get_admin_user), job_lock: JobLock = Body(...)) -> JobLock:
```
mvdbeek, line 20: These use the ORM, they should not be async routes.

mvdbeek:
from https://fastapi.tiangolo.com/async/:
> If you are using a third party library that communicates with something (a database, an API, the file system, etc) and doesn't have support for using await, (this is currently the case for most database libraries), then declare your path operation functions as normally, with just def, like:
and https://fastapi.tiangolo.com/tutorial/sql-databases/#about-def-vs-async-def

There is more background in https://docs.sqlalchemy.org/en/14/orm/extensions/asyncio.html.
I think we should only use async if there's really no blocking operation in what we do.

lib/galaxy/webapps/galaxy/api/job_lock.py
jmchilton: Apologies - https://github.com/galaxyproject/galaxy/pull/11128

PR #11126, opened 2021-01-14
FastAPI routes for configuration

jdavcs:
This is ready for review.

NOTES:
I am not at all happy with the "catch-all" nature of the configuration controller's endpoints. But I assume we need all or most of them, and for now, while we haven't converted the other controllers, maybe keeping them here is the simplest solution. I think it will be easier to reorganize endpoints once we have all the controllers done. (I kept `tool_lineages` for this reason too.)

I've added `common.py` for code shared by controllers. Since `api/__init__.py` is used for shared dependencies used by fastapi's DI system, I think a separate module is more appropriate. 

Ref: #10889.


mvdbeek:
> * made tours fastapi methods async (following [8c7d74c](https://github.com/galaxyproject/galaxy/commit/8c7d74cb10d2899bbce4779d11f65c9c915d4a43))

Everything that interfaces blocking methods should be regular sync function. That includes FS access, ORM access (so everything that (optionally) fetches a users object), anything using the requests library etc. See my comment in https://github.com/galaxyproject/galaxy/pull/11124#issuecomment-760129229

lib/galaxy_test/api/test_configuration.py
```diff
@@ -23,6 +24,11 @@
 
 class ConfigurationApiTestCase(ApiTestCase):
 
+    def test_whoami(self):
+        response = self._get("whoami")
+        self._assert_status_code_is(response, 200)
+        assert response.json()['email'] == DEFAULT_TEST_USER
```
jmchilton: This will pass on CI but isn't correct - maybe change this to TEST_USER not DEFAULT_TEST_USER at least.

jdavcs: Fixed, thanks!

ic4f:
@mvdbeek Thanks, corrected. Admittedly, I wasn't sure at all of the criteria for deciding on async vs regular. This makes sense now.

lib/galaxy/managers/users.py
```diff
@@ -48,6 +49,17 @@
 """
 
 
+class UserModel(BaseModel):
+    """User in a transaction context."""
+    id: int = Field(title='ID', description='User ID')
+    username: str = Field(title='username', description='User username')
```
davelopez: The first letter of the title should be capitalized to display it in the docs? Otherwise, I think the title will be redundant since it will get the name of the field by default.
Would be nice to keep the consistency in the OpenAPI docs, this is something that can go to the guidelines docs as well. Maybe is a good idea to do a documentation review and formatting once all routes are documented.

lib/galaxy/webapps/galaxy/api/configuration.py
```diff
@@ -2,28 +2,149 @@
...
+    @router.get(
+        '/api/configuration/decode/{encoded_id}',
+        dependencies=[AdminUserRequired],
+        summary="Decode a given id",
+        response_description="Decoded id"
+    )
+    def decode_id(
+        self,
+        trans: ProvidesAppContext = Depends(get_trans),
+        *,
```
davelopez: Is this `*` needed here?

lib/galaxy/webapps/galaxy/api/configuration.py
```diff
@@ -2,28 +2,149 @@
...
+        dependencies=[AdminUserRequired],
+        summary="Return tool lineages for tools that have them",
+        response_description="Tool lineages for tools that have them"
+    )
+    def tool_lineages(self) -> List[Dict[str, Dict]]:
+        """Return tool lineages for tools that have them."""
+        return self.configuration_manager.tool_lineages()
+
+    @router.put(
+        '/api/configuration/toolbox',
```
davelopez, line 144:
```suggestion
        '/api/configuration/toolbox',
        dependencies=[AdminUserRequired],
```

davelopez (commented):
Nice! :)

lib/galaxy/managers/users.py
jdavcs: Will change. And yes - definitely +1 on documentation review, formatting all meta, docstrings, etc. for consistency.

lib/galaxy/webapps/galaxy/api/configuration.py
jdavcs: Yep. Otherwise you'll get a syntax error: non-default argument (encoded_id) follows default argument (trans). It's a common thing, I think it's mentioned somewhere in fastapi docs, but here's the PEP: https://www.python.org/dev/peps/pep-3102/#specification

lib/galaxy/webapps/galaxy/api/configuration.py
jdavcs: Good catch, thanks!

lib/galaxy/webapps/galaxy/api/configuration.py
mvdbeek: I think we need to move the id encoding and decoding into the id fields. I don't think trans should be needed to decode an id. I started looking into that and it would be a hige simplification for all serialization/deserialization. In the meantime, is there a problem moving trans to the last argument ? That seems like it'd be a more conventional setup.

jmchilton: ConfigurationManager has a self.app - that can be used regardless of what cool type fixes we can do.

jdavcs: ^looking into this. Moving to WIP for now.

jdavcs: Replaced str with EncodedDatabaseIdField. Adding one `typing.cast` fixed the typing issue, hurray! 
Also, specifying the encoded_id as a Query object solved the "non-default arg after default arg" error.
Now looking into dropping trans in the decode endpoint.

Thank you, everyone, for the detailed reviews!

jdavcs: > I think we need to move the id encoding and decoding into the id fields. I don't think trans should be needed to decode an id. I started looking into that and it would be a hige simplification for all serialization/deserialization. In the meantime, is there a problem moving trans to the last argument ? That seems like it'd be a more conventional setup.

Done in 00b7ae8 .

lib/galaxy/webapps/galaxy/api/tours.py
```diff
@@ -38,17 +38,17 @@ class FastAPITours:
...
 
     @router.get('/api/tours')
-    def index(self) -> TourList:
+    async def index(self) -> TourList:
         """Return list of available tours."""
         return self.registry.get_tours()
 
     @router.get('/api/tours/{tour_id}')
-    def show(self, tour_id: str) -> TourDetails:
+    async def show(self, tour_id: str) -> TourDetails:
```
jmchilton: Doesn't this do synchronous I/O?

```python
            with open(tour_path) as f:
                tour = yaml.safe_load(f)
                load_tour_steps(tour)
```

I thought we were supposed to drop these asyncs.

jmchilton (approved):

lib/galaxy/webapps/galaxy/api/tours.py
jdavcs: I missed this, thank you! I made the change to async for config *and* tours, *then* we had that discussion and I changed it back - but only for config! Correcting now...

jmchilton:
Awesome, thanks so much!

Issue #11129, opened 2021-01-14
Add section for Conventions to API Design Guidelines


nsoranzo:
Wouldn't it be interesting to add a section for `Conventions` in the guidelines?
For example, naming conventions about the `FastAPI<ControllerName>` or `Async<ControllerName>` as @jmchilton suggested in Gitter, or module name conventions like `_schema.py` to contain the 'pydantic' models, etc.

Or maybe in a later PR when we discuss or figure it out a bit more on those conventions :)

_Originally posted by @davelopez in https://github.com/galaxyproject/galaxy/issues/11115#issuecomment-760113000_

PR #11132, opened 2021-01-14
Fix mypy error with Union in response_model

davelopez:
Apparently, in some versions of Python, `mypy` does not like the `Union` type in the FastAPI response_model declaration and will output this error:
````python
Argument "response_model" to "get" of "APIRouter" has incompatible type "object"; expected "Optional[Type[Any]]"
````
Since this endpoint needs to return different types of responses depending on its query parameters, this Union can not be removed.

Note: I had to 'unwrap' the line in order to use the `# type: ignore` otherwise no matter where I put it the error was still showing 😕

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/datatypes.py
+++ b/lib/galaxy/webapps/galaxy/api/datatypes.py
@@ -13,8 +13,8 @@ from fastapi import (
     Depends,
     Query,
 )
-from fastapi.routing import APIRouter
 from fastapi_utils.cbv import cbv
+from fastapi_utils.inferring_router import InferringRouter as APIRouter
 
 from galaxy.app import UniverseApplication
 from galaxy.datatypes.registry import Registry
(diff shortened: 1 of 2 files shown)


davelopez:
As @ic4f rightfully suggested in Gitter:

>I suggest dropping the response_model argument (and do that for all cases where we do not want the response model to be different from the annotated return type) and use InferringRouter. In cases where we do need a different response model, use a TypeVar, as per PEP484 https://www.python.org/dev/peps/pep-0484/#generics

So, I have replaced the regular `APIRouter` for `InferringRouter` and removed the `response_model` in all the endpoints for `datatypes` and `licenses`.

This seems to be another good example of convention/guideline that can be documented in #11129 :)

Many thanks to @ic4f!

jdavcs (approved):

ic4f:
Thanks, @davelopez! Also +1 on documenting such practices/conventions, etc. - I think that's an excellent idea!

Issue #11125, opened 2021-01-14
Galaxy requiring docker command before running mulled job containers?


afgane:
Ran into an issue with latest 20.09 where it seems any mulled job container is being looked up on the host by Galaxy running the `docker` command. This does not work for K8s deployments, especially when using container runtime other than Docker (which CloudMan/GVL now do), and prevents just about any job from running.

The following is a traceback showing up in the job handler:
```
urllib3.connectionpool DEBUG 2021-01-13 15:26:00,421 https://quay.io:443 "GET /api/v1/repository/biocontainers/porechop HTTP/1.1" 200 4033
galaxy.tool_util.deps.containers ERROR 2021-01-13 15:26:00,499 Could not get container description for tool 'toolshed.g2.bx.psu.edu/repos/iuc/porechop/porechop/0.2.3'
Traceback (most recent call last):
  File "/galaxy/server/lib/galaxy/tool_util/deps/containers.py", line 243, in find_best_container_description
    resolved_container_description = self.resolve(enabled_container_types, tool_info, **kwds)
  File "/galaxy/server/lib/galaxy/tool_util/deps/containers.py", line 265, in resolve
    container_description = container_resolver.resolve(enabled_container_types, tool_info, install=install, resolution_cache=resolution_cache, session=session)
  File "/galaxy/server/lib/galaxy/tool_util/deps/container_resolvers/mulled.py", line 410, in resolve
    if install and not self.cached_container_description(
  File "/galaxy/server/lib/galaxy/tool_util/deps/container_resolvers/mulled.py", line 385, in cached_container_description
    return docker_cached_container_description(targets, namespace, hash_func, resolution_cache)
  File "/galaxy/server/lib/galaxy/tool_util/deps/container_resolvers/mulled.py", line 202, in docker_cached_container_description
    cached_images = list_docker_cached_mulled_images(namespace, hash_func=hash_func, resolution_cache=resolution_cache)
  File "/galaxy/server/lib/galaxy/tool_util/deps/container_resolvers/mulled.py", line 71, in list_docker_cached_mulled_images
    images_and_versions = unicodify(subprocess.check_output(command)).strip().splitlines()
  File "/usr/lib/python3.8/subprocess.py", line 411, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.8/subprocess.py", line 489, in run
    with Popen(*popenargs, **kwargs) as process:
[...]

PR #11134, opened 2021-01-14
[20.09] Don't attempt to use docker cli if it is not available

mvdbeek:
This fixes https://github.com/galaxyproject/galaxy/issues/11125, which became a problem with https://github.com/galaxyproject/galaxy/pull/10814 and the recent removal of the docker runtime from kubernetes.

This is the quick fix, but I don't like the guessing and disabling that we do if docker is not available or not functional (but test are included, this should work™️ ).

I'm hesitating between adding a config option for the resolver (`docker_cli_available` or `docker_list_images` + `docker_pull_images`) and adding a `KubernetesContainerResolver` / `RemoteContainerResolver` class that doesn't assume docker is available for listing and pulling images.

Not wanting to list/pull images on the Galaxy node is probably a reasonable thing outside of kubernetes as well, so I tend to go with a `RemoteContainerResolver` class in dev ?

### DIFF
--- a/lib/galaxy/tool_util/deps/container_resolvers/mulled.py
+++ b/lib/galaxy/tool_util/deps/container_resolvers/mulled.py
@@ -325,6 +325,8 @@ def targets_to_mulled_name(targets, hash_func, namespace, resolution_cache=None,
 
 class DockerContainerResolver(ContainerResolver):
 
+    container_type = 'docker'
+
     def __init__(self, *args, **kwargs):
         self._docker_cli_available = bool(which('docker'))
         super().__init__(*args, **kwargs)
@@ -343,7 +345,6 @@ class DockerContainerResolver(ContainerResolver):
 class CachedMulledDockerContainerResolver(DockerContainerResolver):
 
     resolver_type = "cached_mulled"
-    container_type = "docker"
     shell = '/bin/bash'
 
     def __init__(self, app_info=None, namespace="biocontainers", hash_func="v2", **kwds):
@@ -389,7 +390,6 @@ class MulledDockerContainerResolver(DockerContainerResolver):
     """Look for mulled images matching tool dependencies."""
 
     resolver_type = "mulled"
-    container_type = "docker"
     shell = '/bin/bash'
     protocol = None
 
@@ -405,8 +405,8 @@ class MulledDockerContainerResolver(DockerContainerResolver):
         except subprocess.CalledProcessError:
             # We should only get here if a docker binary is available, but command quits with a non-zero exit code,
             # e.g if the docker daemon is not available
-            log.exception('An error occured while listing cached docker images. Disabling docker image cache.')
-            self.docker_cli_available = False
+            log.exception('An error occured while listing cached docker image. Docker daemon may need to be restarted.')
+            return None
 
     def pull(self, container):
         if self.docker_cli_available:
@@ -493,7 +493,6 @@ class BuildMulledDockerContainerResolver(DockerContainerResolver):
     """Build for Docker mulled images matching tool dependencies."""
 
     resolver_type = "build_mulled"
-    container_type = "docker"
     shell = '/bin/bash'
     builds_on_resolution = True
 
--- a/lib/galaxy/tool_util/deps/container_resolvers/mulled.py
+++ b/lib/galaxy/tool_util/deps/container_resolvers/mulled.py
@@ -9,6 +9,7 @@ import subprocess
 from galaxy.util import (
     string_as_bool,
     unicodify,
+    which,
 )
 from galaxy.util.commands import shell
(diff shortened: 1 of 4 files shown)


lib/galaxy/tool_util/deps/container_resolvers/mulled.py
```diff
@@ -488,7 +514,7 @@ def __init__(self, app_info=None, namespace="local", hash_func="v2", auto_instal
         self.auto_init = self._get_config_option("involucro_auto_init", True)
 
     def resolve(self, enabled_container_types, tool_info, install=False, **kwds):
-        if tool_info.requires_galaxy_python_environment or self.container_type not in enabled_container_types:
+        if not self.docker_cli_available or tool_info.requires_galaxy_python_environment or self.container_type not in enabled_container_types:
```
jmchilton: It would seem like it would be better to remove this from the list of default resolvers loaded if docker cli is unavailable - rather than loading it and just disabling it altogether? That way if it is explicitly loaded, it fails - like it should.

jmchilton:
Thanks for working on this fix!

lib/galaxy/tool_util/deps/container_resolvers/mulled.py
mvdbeek: Good idea, I'll do that.

lib/galaxy/tool_util/deps/containers.py
```diff
@@ -222,9 +222,15 @@ def __default_containers_resolvers(self):
                 CachedMulledSingularityContainerResolver(self.app_info, namespace="local"),
                 MulledDockerContainerResolver(self.app_info, namespace="biocontainers"),
                 MulledSingularityContainerResolver(self.app_info, namespace="biocontainers"),
-                BuildMulledDockerContainerResolver(self.app_info),
-                BuildMulledSingularityContainerResolver(self.app_info),
             ])
+            # BuildMulledDockerContainerResolver and BuildMulledSingularityContainerResolver both need the docker daemon to build images.
+            # If docker is not available, we don't load them.
+            build_mulled_docker_container_resolver = BuildMulledDockerContainerResolver(self.app_info)
```
jmchilton, line 228: Thanks!

jmchilton (approved):

lib/galaxy/tool_util/deps/container_resolvers/mulled.py
```diff
@@ -322,7 +323,24 @@ def first_tag_if_available(image_name):
...
+    def docker_cli_available(self, value):
+        if not value:
+            log.info('Docker CLI not available, cannot list or pull images in Galaxy process. Does not impact kubernetes.')
+        self._docker_cli_available = value
+
+
+class CachedMulledDockerContainerResolver(DockerContainerResolver):
 
     resolver_type = "cached_mulled"
     container_type = "docker"
```
nsoranzo: You could move this line to `DockerContainerResolver` and remove it from all its subclasses.

lib/galaxy/tool_util/deps/container_resolvers/mulled.py
```diff
@@ -382,11 +400,18 @@ def __init__(self, app_info=None, namespace="biocontainers", hash_func="v2", aut
...
 
     def cached_container_description(self, targets, namespace, hash_func, resolution_cache):
-        return docker_cached_container_description(targets, namespace, hash_func, resolution_cache)
+        try:
+            return docker_cached_container_description(targets, namespace, hash_func, resolution_cache)
+        except subprocess.CalledProcessError:
+            # We should only get here if a docker binary is available, but command quits with a non-zero exit code,
+            # e.g if the docker daemon is not available
+            log.exception('An error occured while listing cached docker images. Disabling docker image cache.')
+            self.docker_cli_available = False
```
nsoranzo: Since there is no method to re-enable `docker_cli_available` apart from a Galaxy server restart, can't we just return None without disabling this?
The docker daemon could simply be restarted by an admin that notices the error in the Galaxy logs.

mvdbeek: sounds good, will do that

nsoranzo (approved):

mvdbeek:
The package test failure looks unrelated, but we should probably do something about it ?

dannon:
I saw that failing elsewhere, too -- agreed it's unrelated but that someone should do something :)

mvdbeek:
I think the thing is that packages don't pin dependencies, so https://github.com/python-social-auth/social-core/releases/tag/4.0.3 depends on an dev release of pyjwt. We could just let it sit, or we pin python-social-auth to < 4.0.3

nsoranzo:
@dannon @mvdbeek I've got a fix, will open PR in a minute. Merging this one in the mean time.

PR #11121, opened 2021-01-13
Client testing/linting updates and followup

dannon:
I started working on this after https://github.com/galaxyproject/galaxy/pull/10904 as more followup.  It looks like a bunch of the new tests have issues like what I fixed in that PR.

We fixed a bunch of these, but some remain.  I would like to go ahead and get this in, though, to prevent more proliferation of this error on new code.  To that end, I used inline eslint-disables on the remaining few issues and made a note in the code.  I'm also going to open a separate issue for followup on those, so we don't lose it.  (https://github.com/galaxyproject/galaxy/issues/11636) 

Components with prop mutation issues:

- [x] client/src/components/Collections/ListDatasetCollectionElementView.vue
- [x] client/src/components/Collections/PairedElementView.vue
- [x] client/src/components/LibraryFolder/TopToolbar/FolderTopBar.vue
- [x] client/src/components/Masthead/Masthead.vue
- [ ] client/src/components/RuleBuilder/ColumnSelector.vue
- [x] client/src/components/RuleBuilder/RuleComponent.vue
- [ ] client/src/components/Tags/Tags.vue
- [ ] client/src/components/User/CloudAuth/CloudAuthItem.vue
- [x] client/src/components/Workflow/Editor/Index.vue

PR #11138, opened 2021-01-14
replaced global vue with test vue in unit tests

Nerdinacan:
Noticed a lot of client-side tests are continuing to use global Vue instead of our curated Vue instance.

Right now, it's just an annoyance with the eye-spam during test runs with the BootstrapVue components blithely trying to access a non-existent document.

But the problem is that component authors are not using [the standard mount function](https://github.com/galaxyproject/galaxy/blob/dev/client/src/utils/mountVueComponent.js) which means they are unit testing in a Vue environment that is unlikely to match the one in which the component actually runs.

The [helper we wrote](https://github.com/galaxyproject/galaxy/blob/bbd325527cf691258bab7989fbd4787f3160aa2d/client/tests/jest/helpers.js#L60) is intended to generate a Vue instance that matches the one in the client in terms of plugins, filters, directive overrides, etc.

### DIFF
--- a/client/src/components/DatasetInformation/DatasetInformation.test.js
+++ b/client/src/components/DatasetInformation/DatasetInformation.test.js
@@ -1,6 +1,7 @@
 import axios from "axios";
 import MockAdapter from "axios-mock-adapter";
 import { mount } from "@vue/test-utils";
+import { getLocalVue } from "jest/helpers";
 import DatasetInformation from "./DatasetInformation";
 import datasetResponse from "./testData/datasetResponse";
 import flushPromises from "flush-promises";
@@ -17,6 +18,8 @@ const mockDatasetProvider = {
         });
     },
 };
+
+const localVue = getLocalVue();
 
 describe("DatasetInformation/DatasetInformation.vue", () => {
     let wrapper;
@@ -42,6 +45,7 @@ describe("DatasetInformation/DatasetInformation.vue", () => {
             stubs: {
                 DatasetProvider: mockDatasetProvider,
             },
+            localVue,
         });
         datasetInfoTable = wrapper.find("#dataset-details");
         await flushPromises();
--- a/client/src/components/HistoryExport/Index.vue
+++ b/client/src/components/HistoryExport/Index.vue
@@ -25,20 +25,20 @@
 </template>
 
 <script>
-import Vue from "vue";
-import BootstrapVue from "bootstrap-vue";
+import { BCard, BTabs, BTab } from "bootstrap-vue";
 import ToLink from "./ToLink.vue";
 import ToRemoteFile from "./ToRemoteFile.vue";
 import { Services } from "components/FilesDialog/services";
 import LoadingSpan from "components/LoadingSpan";
-
-Vue.use(BootstrapVue);
 
 export default {
     components: {
         LoadingSpan,
         ToLink,
         ToRemoteFile,
+        BCard,
+        BTabs,
+        BTab,
     },
     props: {
         historyId: {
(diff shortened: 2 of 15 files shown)


dannon:
@Nerdinacan Thanks; I was going to take care of this in #11121, but not having to do it myself (again) is even better!

Nerdinacan:
Ahh crap sorry Dannon, didn't see that one.

dannon:
Totally fine -- again, one less thing I have to do, and I'm not upset about that :)

jmchilton:
If we're using that pattern over and over - does it make sense to have a wrapper around shallowMount in jest/helper that takes care of that detail?

dannon:
@jmchilton Yeah, when I added the helper my thinking it should be used sparingly, and was an extra indicator that a component was leaning on these extra directives.  I'll admit the behavior is more prevalent than I was thinking though.

Nerdinacan:
@jmchilton 
Mount renderless is for the renderless components (i.e. providers). The ones with just a single big slot and you still need to check the slotProps even though there's nothing to render.

I think the only reason I didn't make a helper as you suggested is that we should really be using shallowMount, and a lot of people are using mount(), and I'd have to rewrite a bunch of tests to make that fly.

The fact that people are using mount() instead of shallowMount() tells me that the components are too monolithic.
Presumably, nobody cares whether <b-form-checkbox> properly prints a string, right? What you care about is probably that variable XYZ is true or false. If all you care about is the variable you don't need a full mount(), you just check to see if vm.XYZ is what you want after you set up the test?

I think the real problem is people are building big-ass components instead of breaking them into logical chunks. It's a common problem with new component designers who aren't yet comfortable with props in / events out.

jmchilton:
> The fact that people are using mount() instead of shallowMount() tells me that the components are too monolithic.

While I've switch to ``shallowMount`` more and more for my recent tests, the thing I would read into this is that nearly all the examples in the official docs ( https://vue-test-utils.vuejs.org/api/wrapper/#emitted, https://vue-test-utils.vuejs.org/api/wrapper/#methods, etc...) use ``mount`` instead of ``shallowMount``.

dannon:
@Nerdinacan RefactorConfirmationModal is still spewing to info in my quick test

Nerdinacan:
@jmchilton 
No argument there.

The official docs, and every video I've ever seen are basically teaching you how to write a unit test that is testing rendering. They are showing you how to write a unit test that is really more of a selenium test.

While that's important if you are indeed writing a new kind of low-level rendering component, what's more important and flexible is making sure the business logic works right since presumably the rendering should be able to change when somebody makes it prettier without busting tests.

I agree with you that the availability of good advice about writing tests is thin.

Nerdinacan:
@dannon 
Hmmm yeah, flushPromises()?

jmchilton (approved):
I think we should just merge as is, seems atomic and working and just follow up with additional fixes in new PRs. I'll wait on the other requested reviewers though.

dannon:
It's most of the way there, let's give @Nerdinacan the chance to address the last bit in this PR before merging.

Nerdinacan:
@dannon 
The problem is not the test, it's the code.

The message itself is an indicator that something strange is happening, and sure enough if you look at one of the data services used in that component, it [references global Vue, seemingly to do the worrk of a Promise](https://github.com/galaxyproject/galaxy/blob/bbd325527cf691258bab7989fbd4787f3160aa2d/client/src/components/Workflow/Editor/modules/model.js#L18).

The Workflow editor needs to be fixed, not the test.

Nerdinacan:
I went ahead and removed the suppression of the messages so we remember to refactor that service.

dannon (approved):

PR #11139, opened 2021-01-14
Fix synchronization between legacy Galaxy currHistoryPanel and Vuex store

Nerdinacan:
Refining connection between legacy galaxy current history and beta history current history in the Vuex store.


client/src/components/Upload/UploadModal.vue
```diff
@@ -12,230 +17,56 @@
...
-            <b-tab title="Rule-based" id="rule-based" button-id="tab-title-link-rule-based" v-if="showRules">
-                <rules-input :app="this" />
-            </b-tab>
-        </b-tabs>
-        <div v-else>
-            <loading-span message="Loading required information from Galaxy server." />
-        </div>
+
+        <CurrentUser v-slot="{ user }">
+            <UserHistories v-if="user && user.id" v-slot="{ currentHistoryId }">
```
mvdbeek: This is nice! Can we do this outside of vue as well or can we wrap backbone components in vue components? I don't expect we'll be able to modernize the tool form before the release ...

mvdbeek:
I've pushed a commit that fixes the WorkflowRunSuccess component, feel free to drop it if that's not the right way to do it.
I think the last thing we really need to have fixed is https://github.com/Nerdinacan/galaxy/blob/db5d8c81c1f25e5057dadffec18c82c2b44b3448/client/src/mvc/tool/tool-form.js#L239 ... any hack is welcome there.

client/src/components/Upload/UploadModal.vue
Nerdinacan: @mvdbeek 
Hey I'm actually doing the upload stuff in a complete replacement of the Upload Modal. I'm going to close this PR and open 2 others.

As for "outside vue", one thing you could always do is go straight to the store to get config data you need, then provide that data as props to the Vue component you mount from backbone, if for intance, you're not able to modernize the tool-form.

Alternatively you could just redefine what you mount from Backbone as being a component that leverages these wrappers. I can show you how to pass attributes and listeners from a parent to a child to make it all completely transparent.

Nerdinacan: This PR was starting to balloon into a bigger one than I wanted.

Check this one out for addressing the bugs you originally pinged me for:
https://github.com/galaxyproject/galaxy/pull/11172

Nerdinacan:
That hack you have is fine unless the tool you're using explicitly needs the id, which won't be available. I've modified the HistoryPanelProxy to make sure that id will indeed be present in #11172

PR #11140, opened 2021-01-14
[21.01] Follow-up on best practice workflow linting user interface component

guerler:
This PR is a follow-up for #10988. It polishes the structure of the Vue component and revises how lint warnings are displayed. The collapsible sections have been replaced with fixed success and warning labels consistently for all issues. The smaller font notification sections have been removed such that only one font size is used throughout the component. Automatic fixing options have been reduced to fix all issues and not individual ones. Modules have been separated from the generic utilities module. Overall this PR attempts to simplify the appearance and structure of the component. The underlying API remains unchanged, except that `LegacyWorkflowParameter` has been renamed `UntypedWorkflowParameter`.

image


lib/galaxy/workflow/refactor/schema.py
```diff
@@ -217,7 +217,7 @@ class UpgradeToolAction(BaseAction):
     ConnectAction,
     DisconnectAction,
     ExtractInputAction,
-    ExtractLegacyParameter,
+    ExtractImplicitParameter,
```
jmchilton: This needs to target 21.01 if you're changing the API.

guerler: Yes, it should target 21.01. I wanted to replace the generic term `legacy` here, since it does not really convey any information and we are somewhat overusing it. I missed to comment on the previous PR. It could have been easily fixed before. Thanks a lot for looking into it. Overall it's really an amazing feature, I hope we can polish a little bit with this PR. I am definitely open to suggestions.

client/src/components/Workflow/Editor/Attributes.vue
```diff
@@ -105,7 +104,7 @@ export default {
             default: null,
         },
         parameters: {
-            type: LegacyParameters,
```
dannon, line 108: Seems like losing the type information here for 'Object' is a step backwards -- if we're doing this can you add a validator instead?

jmchilton:
I mean obviously this is amazingly more pretty than what was there. I'm truly impressed and in awe. I wish you would have done it without limiting what can happen in the lint sections though - they are over specified to the two kind of limited things the panel was doing so far. Most of the whole list of TODOs for this #10989 was more details and functionality within those sections. I'm really worried this is going to be like the side panel of the workflow nodes - where every time I want to implement custom functionality I hit a hard wall of a rigid framework and I land up incapable of doing what I need to.

client/src/components/Workflow/Editor/Attributes.vue
```diff
@@ -80,7 +79,7 @@ export default {
         },
         name: {
             type: String,
-            required: true,
```
dannon, line 83: Is it actually possible to have an attribute with an empty/null name?

guerler:
> I mean obviously this is amazingly more pretty than what was there. I'm truly impressed and in awe. I wish you would have done it without limiting what can happen in the lint sections though - they are over specified to the two kind of limited things the panel was doing so far. Most of the whole list of TODOs for this #10989 was more details and functionality within those sections. I'm really worried this is going to be like the side panel of the workflow nodes - where every time I want to implement custom functionality I hit a hard wall of a rigid framework and I land up incapable of doing what I need to.

The lint section is pretty straightforward, I think if you need a different type which does not fit into that lint section format you can just add a new one, or even write it directly into the lint component. I don't think it should hold your flexibility back in anyway.

lib/galaxy/workflow/refactor/schema.py
jmchilton: Implicit sounds nifty and exciting and new - and what it is is parameters that don't work on the backend or with the API. I too wish we would have argued about this sooner, I understand I did slip this in over the holidays which was shitty of me.

client/src/components/Workflow/Editor/Attributes.vue
guerler: Thanks for catching that.

client/src/components/Workflow/Editor/Attributes.vue
guerler: Yes, I noticed a console log error regarding this attribute. I have not dug much deeper yet, but it seems that this attribute is not available when the Attributes component is generated and then populated at a later time through an async request. Removing the required attribute avoids the console error.

client/src/components/Workflow/Editor/Options.vue
```diff
@@ -55,7 +55,7 @@
                 ><span class="fa fa-floppy-o mr-1" />Save As...</b-dropdown-item
             >
             <b-dropdown-item href="#" @click="$emit('onLint')"
-                ><span class="fa fa-magic mr-1" />Check for Best Practices</b-dropdown-item
+                ><span class="fa fa-magic mr-1" />Validate</b-dropdown-item
```
jmchilton: It isn't an invalid workflow because it doesn't have a license or an annotation - validate feels wrong? I think if you wanted to shorten this "Best Practices" might be a better shortening. I'm not going to die on that hill though.

client/src/components/Workflow/Editor/Attributes.vue
dannon: That smells like a bigger problem -- if it's a prop here that, data-wise is required, it should be available to the component from the start, not optional.

guerler: Well, yea I agree we should look more into it. The goal of this commit here is to avoid the console error, although it's not really part of this PR. Maybe we should create a separate issue for it and follow-up. It could make it into the release as bug fix.

client/src/components/Workflow/Editor/Attributes.vue
```diff
@@ -105,7 +105,7 @@ export default {
             default: null,
         },
         parameters: {
-            type: LegacyParameters,
+            type: ImplicitParameters,
```
mvdbeek: Call it LegacyWorkflowParamater maybe ? I don't think anything about this is implicit.

client/src/components/Workflow/Editor/Lint.vue
```diff
@@ -1,470 +1,201 @@
...
+                success-message="This workflow defines a license."
+                warning-message="This workflow does not specify a license. This is important metadata for workflows
+                    that will be published and/or shared to help workflow executors understand how it
+                    may be used."
+                attribute-link="Specify a License."
+                @onClick="onAttributes"
+            />
+            <LintSection
+                success-message="Workflow parameters are using formal inputs."
+                warning-message="This workflow uses implicit workflow parameters. They should be replaced with
```
mvdbeek: Again, I don't think there's anything implicit about them. An alternative would by untyped, but that seems too technical.

mvdbeek (commented):
> Automatic fixing options have been reduced to fix all issues and not individual ones.

That is a step back, can you please add this back? It's cool to be able to do all at once, but more often than not you'll want to decide what to do.

client/src/components/Workflow/Editor/Lint.vue
```diff
@@ -1,470 +1,201 @@
...
+                :okay="checkLicense"
+                success-message="This workflow defines a license."
+                warning-message="This workflow does not specify a license. This is important metadata for workflows
+                    that will be published and/or shared to help workflow executors understand how it
+                    may be used."
+                attribute-link="Specify a License."
+                @onClick="onAttributes"
+            />
+            <LintSection
+                success-message="Workflow parameters are using formal inputs."
```
mvdbeek: I think this should be formal parameters or formal input parameters?

jmchilton:
> It's cool to be able to do all at once, but more often than not you'll want to decide what to do.

In rare form I actually wrote this component with an intention on using it (for a set of workflows while wearing my contractor hat). I can confirm that I would definitely rather apply these operations one at a time and see what is going on.

dannon:
This is kind of an aside, but moving forward for this kind of change it's *a lot* easier to review changes if we separate all the renaming of things out into either a single commit or a separate PR.

jmchilton:
> I don't think it should prevent you from being flexible.

Well, clearly you write superior UIs and I totally appreciate this is your knowledgeable perspective - but I'm sharing my perspective. I'm significantly less likely to go invest in a bunch of detailed, custom UI helpers and contextualized help text work if the first thing I have to do is go in and undo all the consistency and conciseness you added and re-add a bunch of functionality & help text you dropped. It sounds frustrating both technically and socially to bump heads with such a different vision for what this should look like.

client/src/components/Workflow/Editor/Lint.vue
guerler: Renamed it to untyped parameters internally, we can name it whatever we we like in the ui.

lib/galaxy/workflow/refactor/schema.py
guerler: Renamed it `Untyped` internally following @mvdbeeks suggestion. Readded `legacy` to the UI help text.

guerler:
> > I don't think it should prevent you from being flexible.
> 
> Well, clearly you write superior UIs and I totally appreciate this is your knowledgeable perspective - but I'm sharing my perspective. I'm significantly less likely to go invest in a bunch of detailed, custom UI helpers and contextualized help text work if the first thing I have to do is go in and undo all the consistency and conciseness you added and re-add a bunch of functionality & help text you dropped. It sounds frustrating both technically and socially to bump heads with such a different vision for what this should look like.

I think the structure of the previous code was a reflection of the UI which they build. The changes here dramatically improve readability and maintainability of the code which is a priority even more important than the resulting UI itself while using 30% less code. If we do not start off right, we end up loosing control over modules and finally have to rewrite them in their entirety. Overall I think UI components should be written with the same care API's are written. We should not just implement font sizes, button styles, button behaviors and customized style elements which have never been used before without a good reason. I’m sorry that I did not review the previous PR earlier, this could avoided duplication of efforts. For the future, I offer to add new linting features to this UI as they become available. Breaking the consistency should rarely be needed particularly not right away just by assuming that it has to be broken in the future anyway for features which have not been developed yet.

The only element which I dropped from the previous PR are the gear-icons which allow fixing some lint issues but not others. These can be re-added if really required. Currently there are three linting issues which can be fixed automatically. I am not sure why users should individually fix these issues if they can fix them automatically unless there is a problem with automatically fixing them. In the latter case I first want to make sure that automatically fixing lint issues works reliably before expanding the options.

mvdbeek:
> The only element which I dropped from the previous PR are the gear-icons which allow fixing some lint issues but not others. These can be re-added if really required. Currently there are three linting issues which can be fixed automatically. I am not sure why users should individually fix these issues if they can fix them automatically

Because you may not want to fix them all and then undo the parts you don't like. Maybe just don't want to follow all of the best practices, or some of the unfixed inputs can be shared ? I have a hard time thinking that a workflow author that spent some time and thought in constructing a workflow would want to just click a button and cross fingers it did the right thing. It's also tricky then to see what exactly changed (plus I think there are more actions we could add, making that even more complicated).

If you don't want to restore the button, at least make it checkboxes you can apply ?

guerler:
@mvdbeek I added the option to automatically fix certain issues. If the issue is fixable a confirmation alert is shown which gives the user the option to fix the issue automatically. Depending on the preference I could also add a button or additional link into the form. However I thought it might be useful to show a confirmation dialog which would allow us to provide more details on the operation being performed.

client/src/components/Workflow/Editor/Attributes.test.js
```diff
@@ -10,15 +10,15 @@ const TEST_NAME = "workflow_name";
...
+        const untypedParameters = new UntypedParameters();
+        untypedParameters.getParameter("workflow_parameter_0");
+        untypedParameters.getParameter("workflow_parameter_1");
         const wrapper = mount(Attributes, {
             propsData: {
                 id: "workflow_id",
                 name: TEST_NAME,
                 tags: ["workflow_tag_0", "workflow_tag_1"],
-                parameters: legacyParameters,
+                parameters: untypedParameters,
```
Nerdinacan, line 21: This structural comment may fall outside the scope of your PR Sam, but passing a a class instance into Vue like this, has turned out to be problematic. I've done similar things in the past and I need to go fix them myself.

You don't want anything coming in as a prop that has the potential to be mutated unless it is one of the few types of javascript objects that Vue knows how to deal with. Immutable props go in, events come out.

Vue only knows how to make observable plain Objects, plain Arrays and primitives. Anything else it just tries to treat like a plain Object. It will add a ton of getters/setters to the object, which can collide with getters on your incoming legacyParameters class definition.

I've heard people can get away with this by doing an Object.freeze() on the object before binding it to the property. Then, Vue has no way to add its getters/setters and internally it is no longer internally observable. You can still do all your read operations inside the property value, but you won't have the ability to do writes or mutate the internal data structures. This way nobody can change it, and it's a properly immutable input. 

The way to trigger a component update in such a scenario is to hand the component an entirely new Object.freeze(legacyParameters) from the outside. Then you've got a genuine immutable input.

It seems like a minor complaint, but this can cause impossible-to-find bugs, often of the form "Why the hell isn't my screen updating?"

client/src/components/Workflow/Editor/LintSection.vue
```diff
@@ -1,22 +1,31 @@
...
+            <span>{{ warningMessage }}</span>
+            <div v-if="hasWarningItems" class="mt-2">
+                <div
+                    v-for="(item, idx) in warningItems"
+                    :key="idx"
+                    @mouseover="onMouseOver(item.stepId)"
+                    @mouseleave="onMouseLeave(item.stepId)"
+                    class="ml-2"
+                >
+                    <a href="#" @click="onClick(item.stepId, item)" class="scrolls">
```
jmchilton:
```vue
                                        <font-awesome-icon
                                            v-if="input.canExtract"
                                            icon="cog"
                                            v-b-tooltip.hover
                                            title="Attempt to create a workflow input connected to this step input."
                                            @click="extractWorkflowInput(input)"
                                        />
```

The previous version of this would only show an icon if it is auto-fixable - ``v-if="input.canExtract"``. Can you modify this LintSection interface to put an auto-fixable attribute on the items?

client/src/components/Workflow/Editor/LintSection.vue
```diff
@@ -1,22 +1,31 @@
...
+        </div>
+        <div v-else>
+            <font-awesome-icon icon="exclamation-triangle" class="text-warning" />
+            <span>{{ warningMessage }}</span>
+            <div v-if="hasWarningItems" class="mt-2">
+                <div
+                    v-for="(item, idx) in warningItems"
+                    :key="idx"
+                    @mouseover="onMouseOver(item.stepId)"
+                    @mouseleave="onMouseLeave(item.stepId)"
```
jmchilton: This would be a better abstraction IMO if emitted the item and not the step id - so we could theoretically highlight actual inputs and outputs. That isn't important in this PR, but I wanted to highlight a direction I would want to take this if development continues...

client/src/components/Workflow/Editor/LintSection.vue
```diff
@@ -1,22 +1,31 @@
...
-            <div class="lint-section-body">
-                <slot> </slot>
+    <div class="mb-2">
+        <div v-if="isOkay">
+            <font-awesome-icon icon="check" class="text-success" />
+            <span>{{ successMessage }}</span>
+        </div>
+        <div v-else>
+            <font-awesome-icon icon="exclamation-triangle" class="text-warning" />
+            <span>{{ warningMessage }}</span>
```
Nerdinacan: Something that's available to all components that use the standard mount is the localize filter and the v-localize directive. You could be using {{ warningMessage | localize }} here (and on every other message).

client/src/components/Workflow/Editor/Lint.vue
```diff
@@ -1,470 +1,237 @@
...
-                        </div>
-                    </span>
-                </LintSection>
-                <LintSection :okay="outputsOkay" title="Outputs" collapse-id="c-outputs">
-                    <span>
-                        <div v-if="outputsOkay">
-                            This workflow has outputs and they all have valid labels.
-                        </div>
-                        <div v-else-if="outputs.length == 0">
-                            This workflow has no labeled outputs, please select and label at least one output.
```
jmchilton, line 127: We're dropping this warning right? Can this be restored?

client/src/components/Workflow/Editor/Attributes.test.js
```diff
@@ -10,15 +10,15 @@ const TEST_NAME = "workflow_name";
 describe("Attributes", () => {
     it("test attributes", async () => {
```
Nerdinacan, line 11: I think we really need to start being more careful describing our tests. The tests are what the next victim uses to fix or modify this code and when it's poorly described, or in this case, not described at all, there is no indication of what it means to pass or fail a test.

It turns the tests into an obstacle instead of a guide.

Nerdinacan (commented):
I like some of the structural changes you've made to externalize data manipulation by leveraging events and to extract simple logic manipulation from the Vue definition. It's easier to test that way and easier to read. I think we could go even further with that idea, but probably can address that later.

client/src/components/Workflow/Editor/LintSection.vue
guerler: Added the localize option to the warningMessage.

client/src/components/Workflow/Editor/Attributes.test.js
guerler: Thanks for pointing this out. I totally agree. In this case here the PR is a mere refactoring so I do not intend to change it but I agree that we should avoid this in the future and maybe even refactor it here with a follow-up.

client/src/components/Workflow/Editor/LintSection.vue
guerler: Modified it such that item is parsed instead of item.stepId.

client/src/components/Workflow/Editor/LintSection.vue
guerler: Showing a 'magic' icon instead of 'search' if issue is autofixable.

client/src/components/Workflow/Editor/Lint.vue
guerler: Add the additional warning.

client/src/components/Workflow/Editor/LintSection.vue
Nerdinacan: I notice you did it on just the one message. Maybe we should schedule another PR that goes through all the components and just put in the localization filters when appropriate.

client/src/components/Workflow/Editor/Attributes.test.js
Nerdinacan: Awesome. That short description of what you're trying to test will help the next person. 

It looks like @jmchilton had some pretty descriptive tests he added in that file a month ago, maybe you need a rebase to catch them? Or are they no longer relevant after your refactor?

guerler:
Failing test is unrelated.

jmchilton:
I want to track this down but I'm running a little crazy today. Is it possible there was some regression int terms of distinguishing between legacy parameters and disconnected inputs?

Why does it think from is disconnected here?

Screen Shot 2021-02-02 at 2 17 52 PM

The same workflow on dev and it doesn't seem to?

Screen Shot 2021-02-02 at 2 18 23 PM

I also noticed that nested legacy parameters aren't working (parameters inside of conds and such), but they aren't working in dev either. I'm worried there has been some sort of regression already but that doesn't seem related to this PR.

guerler:
@jmchilton thank you for catching this, I fixed the condition to determine missing connections.

jmchilton:
Yes - this seems to work on some simple cases that it wasn't at my last review. Any chance I can convince you to retarget against 21.01?

jmchilton (approved):
I'm approving this because it is a lot cleaner and prettier and I do like the confirmation box it adds. I do still think some of the abstractions are a step backward and some of the input we were providing users is gone and that makes me sad but @guerler has been awesome and made the big structural changes @mvdbeek requested so I think it is important to get this into 21.01.

jmchilton:
I'm confused - this has dev test failures (?), any chance you can rebase and force push now that you have new target.

guerler:
> I'm confused - this has dev test failures (?), any chance you can rebase and force push now that you have new target.

Strange, I rebased it against the latest release_21.01 and had no conflicts after removing the 'craft' in front.

jmchilton:
I'll try re-running the failed tests then. Maybe we just confused Github Workflows somehow.

jmchilton (approved):
I'm approving this, but those tests are still running against some random dev branch commit. I think a rebase after the branch switch is warranted here but if you want to click the merge button go for it 😅.

mvdbeek:
Yeah, test failures are unrelated. I'll take a quick look and merge if it works for me.

mvdbeek:
Hmm, clicking on any of the issues doesn't activate the node anymore (it briefly flashes, instead of being "bold" on 21.01/dev), and the listed issues here are not correct (they all have labels and annotations) and working correctly on the dev branch and 21.01.
Screenshot 2021-02-23 at 17 20 07
I understand you didn't like the new buttons
Screenshot 2021-02-23 at 17 25 35
compare that with dev:
Screenshot 2021-02-23 at 17 28 02

I also think the explanation is really helpful, and the big green and orange boxes that group kinds of issues, while not pretty, are really nice for users. If you want to test this, https://github.com/iwc-workflows has a bunch of workflows that should be all green. I've used https://github.com/iwc-workflows/sars-cov-2-ont-artic-variant-calling/blob/main/ont-artic-variation.ga here.

mvdbeek:
OK, so I don't see the highlighting issue anymore, if we can fix the annotation and label  false positives I think we can merge this.

PR #11142, opened 2021-01-15
Start type signature for UniverseApplication.

jmchilton:
Builds on #11131. Provides stronger static checking, better docs, probably IDE stuff I should learn how to use 😅, etc...

### DIFF
--- a/lib/galaxy/tools/parameters/basic.py
+++ b/lib/galaxy/tools/parameters/basic.py
@@ -21,7 +21,6 @@ from galaxy.util import (
     unicodify,
     XML,
 )
-from galaxy.util.bunch import Bunch
 from galaxy.util.dictifiable import Dictifiable
 from galaxy.util.expressions import ExpressionContext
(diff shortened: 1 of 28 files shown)


lib/galaxy/tools/parameters/basic.py
```diff
@@ -38,7 +37,12 @@
 
 log = logging.getLogger(__name__)
 
-workflow_building_modes = Bunch(DISABLED=False, ENABLED=True, USE_HISTORY=1)
+
+class workflow_building_modes:
```
mvdbeek, line 41: Should this be an enum ?

mvdbeek (approved):

lib/galaxy/tools/parameters/basic.py
jmchilton: We don't iterate over the keys or use them as anything but constants and the types don't match. I guess we could get slightly better typing if we got this to work as an enum? I'm not sure.

PR #11144, opened 2021-01-15
[21.01] Maintain expanded state of JobStep during polling

mvdbeek:
The InvocationStepProvider updates the invocation state in 3 second
intervals, and with it the jobs array. The new jobs in the  array don't
have the _showDetails attribute, so all expanded view will be reset.
This fixes that by keeping a component local state for the _showDetails
attribute.

### DIFF
--- a/client/src/components/WorkflowInvocationState/JobStep.test.js
+++ b/client/src/components/WorkflowInvocationState/JobStep.test.js
@@ -1,0 +1,71 @@
+import BootstrapVue from "bootstrap-vue";
+import JobStep from "./JobStep";
+import { mount } from "@vue/test-utils";
+import jobs from "./test/json/jobs.json";
+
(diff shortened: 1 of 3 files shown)


jmchilton (approved):
I approve this the changes, you may want to ping Mason on the test cases - might be a good case for using getLocalVue from jest/helpers.

mvdbeek:
@Nerdinacan does this & the test cases look good to you ?

dannon:
The tests are fine using the standard createLocalVue.  They don't need to use the getLocalVue w/ directives.

mvdbeek:
I'm gonna roll with that approval then, but if you find anything that's not great I'll open a follow-up. Thanks @dannon @jmchilton!

dannon (commented):
:thumbsup:

PR #11149, opened 2021-01-15
[21.01] Unpin requirements for galaxy-data package

nsoranzo:


mvdbeek:
Thanks, I would have totally tried to pin social-auth instead. This is clearly better.

mvdbeek (approved):

mvdbeek:
Ah ... I think maybe just target 20.09, since it's the last stable release and it already dropped python 2 support ?

jmchilton (approved):

nsoranzo:
> Do we really support python 2 in 20.05 packages ?

It seems some packages still work, others don't (e.g. galaxy-objectore 20.5.0). To be sure I will target 20.09 in a backport.

nsoranzo:
> Thanks, I would have totally tried to pin social-auth instead. This is clearly better.

Actually that's needed on Python 3.5, i.e. for 20.09. Closing in favour of https://github.com/galaxyproject/galaxy/pull/11150 .

Issue #11059, opened 2021-01-06
Workflow extraction fails for list:paired inputs copied from another history


jmchilton:
Seems to not work regardless of mapping vs. not mapping.

mvdbeek:
I guess this should be fixed by https://github.com/galaxyproject/galaxy/pull/11151

PR #11151, opened 2021-01-16
[21.01] Fix certain classes of workflow extraction on copied objects.

jmchilton:
I believe this fixes https://github.com/galaxyproject/galaxy/issues/11059 reported by @nekrut.


mvdbeek (approved):
Now if the extraction wouldn't depend on the hid ... 🙏

PR #11153, opened 2021-01-16
Fix remaining issues reported by bugbear; add it to flake8 reqs

nsoranzo:


test/unit/util/test_sqlite_utils.py
```diff
@@ -37,19 +39,14 @@ def test_sqlite_exploits():
...
 
 def __assert_query_errors(connection, query):
-    exception = False
-    try:
-        for row in connection.cursor().execute(query):
-            pass
-    except Exception:
-        exception = True
-    assert exception
+    with pytest.raises(Exception):
```
jmchilton, line 48: This is only a small point but I don't think we shouldn't introduce pytest as a dependency to unit tests just to cleanup this pattern IMO. The cost of introducing the hard dependency on the library outweighs cleaning up the test a bit. I'm going merge this, but I'm going to keep resisting pytest.raises unless I have another reason to import pytest.

lib/galaxy/workflow/refactor/execute.py
```diff
@@ -135,7 +135,7 @@ def _apply_add_input(self, action: AddInputAction, execution: RefactorActionExec
         if input_type in ["data", "dataset"]:
             module_type = "data_input"
         elif input_type in ["data_collection", "dataset_collection"]:
-            module_type == "data_collection_input"
+            module_type = "data_collection_input"
```
jmchilton, line 138: A real bug! We should probably backport this fix.

jmchilton: https://github.com/galaxyproject/galaxy/pull/11154

PR #11155, opened 2021-01-16
Fix all E203, E741 and W504 flake8 errors

nsoranzo:


jmchilton (approved):

jmchilton:
Clicking approve because I think this is a big improvements. But I'll give some time before merging because some people prefer to line up their white spaces in sprawling dicts and such. I think that was mostly Carl that really cared but I've seen others do it.

lib/galaxy/config/script.py
```diff
@@ -154,7 +154,8 @@ def _handle_galaxy_yml(args, config_dir, data_dir):
 
     galaxy_config_template = []
     with open(GALAXY_CONFIG_TEMPLATE_FILE) as fh:
-        for line in [l.rstrip('\n') for l in fh.readlines()]:
+        for line in fh.readlines():
```
bgruening: Can we remove readlines here? Maybe a task a another PR?

bgruening (approved):

lib/galaxy/config/script.py
nsoranzo: Done, thanks!

mvdbeek (approved):

mvdbeek:
There's a conflict here, can you resolve @nsoranzo ?

nsoranzo:
> There's a conflict here, can you resolve @nsoranzo ?

Sure, rebased.

mvdbeek:
Thank you @nsoranzo!

PR #11159, opened 2021-01-17
[21.01] Small tweaks to invocations display

jmchilton:
Small change to backport from testing which I think I'll need to target at dev. The extra "View Report 1" on each invocation in a list doesn't make sense - only makes sense in context of run response I think. Also I think a bunch of ids added should be classes. id was added to a bunch a components that may appear multiple places on a page I believe and so should be class (https://www.w3schools.com/html/html_id.asp).


mvdbeek:
Thank you!

PR #11160, opened 2021-01-17
Selenium test case for invocation grid + display.

jmchilton:
Builds on #11156 and #11159.


mvdbeek:
Thanks so much @jmchilton!

PR #11161, opened 2021-01-17
Parallelize client and server builds in dockerfile

nuwang:
This PR modifies the k8s docker file so that client and server build stages are built in parallel, reducing total build time.

The uncached build time prior to this is ~18.5 minutes locally, which reduces to ~12.5 minutes, a ~33% reduction.
On dockerhub, there appears to be no benefit, as the builds don't seem to be parallelised in the free version.

This PR is dependent on: https://github.com/galaxyproject/ansible-galaxy/pull/120

### DIFF
--- a/.k8s_ci.Dockerfile
+++ b/.k8s_ci.Dockerfile
@@ -87,7 +87,6 @@ RUN git rev-parse HEAD > GITREVISION
 RUN rm -rf \
         .ci \
         .git \
-        .venv/bin/node \
         .venv/include/node \
         .venv/src/node* \
         doc \
@@ -108,7 +107,6 @@ WORKDIR $SERVER_DIR
 RUN rm -rf \
         .ci \
         .git \
-        .venv/bin/node \
         .venv/include/node \
         .venv/src/node* \
         doc \
--- a/.k8s_ci.Dockerfile
+++ b/.k8s_ci.Dockerfile
@@ -1,12 +1,18 @@
 # Stage 1:
 # - base: ubuntu (default) OR prebuilt image0
 # - install build tools
-# - run playbook (image0 avoids rerunning lengthy tasks)
+# - clone playbook
+# Stage 2: build galaxy client and server in parallel
+# Stage 2.1:
+# - run playbook with server build steps only
 # - remove build artifacts + files not needed in container
-# Stage 2:
+# Stage 2.2:
+# - run playbook with client build steps only
+# - remove build artifacts + files not needed in container
+# Stage 3:
 # - install python-virtualenv
 # - create galaxy user + group + directory
-# - copy galaxy files from stage 1
+# - copy galaxy files from stage 2.1 and 2.2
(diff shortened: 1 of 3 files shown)


.k8s_ci.Dockerfile
```diff
@@ -115,7 +153,8 @@ RUN set -xe; \
 WORKDIR $ROOT_DIR
 # Copy galaxy files to final image
 # The chown value MUST be hardcoded (see #35018 at github.com/moby/moby)
-COPY --chown=galaxy:galaxy --from=stage1 $ROOT_DIR .
+COPY --chown=galaxy:galaxy --from=server_build $ROOT_DIR .
+COPY --chown=galaxy:galaxy --from=client_build $SERVER_DIR/static ./server/static
```
nuwang, line 155: @dannon @mvdbeek This copies only the static files over from the client build. Anything else we should copy over?
I guess we do need node for the expression tools (see comment below)

.k8s_ci.Dockerfile
```diff
@@ -81,8 +96,31 @@ RUN rm -rf \
...
+        .ci \
+        .git \
+        .venv/bin/node \
+        .venv/include/node \
+        .venv/src/node* \
+        doc \
+        test \
+        test-data
+# Clean up *all* node_modules, including plugins.  Everything is already built+staged.
+RUN find . -name "node_modules" -type d -prune -exec rm -rf '{}' +
```
nuwang, line 116: @mvdbeek  You mentioned that the expression tools need node. Is it ok if we leave just the node binary or do we need some node_modules as well? The ideal scenario would be if we could separate the node_modules needed by the expression tools from the client node_modules, so that the final image size remains small.

dannon: No, they don't currently use any node modules.  Simply having node available as an execution context should be enough.  Down the road if we did have more complicated expression tools we could add a build process to avoid shipping all of node_modules, if necessary.

.k8s_ci.Dockerfile
dannon: Check visualization build artifacts (they get staged into `/static/plugins/` during the build), but without testing it this should probably work.

mvdbeek (approved):
Looks good to me, can you also verify https://github.com/galaxyproject/galaxy/pull/11161/files#r559193525 ?

.k8s_ci.Dockerfile
nuwang: Thanks, `/static/plugins` dir seems to be populated ok.

nuwang:
@mvdbeek @dannon I've added a docker build hook that should enable parellelization on hub. Have also preserved node. Thanks for reviewing.

nuwang:
It may make sense to include this in 21.01 as well, since it could significantly speed up build times?

mvdbeek:
No problem, I will backport it.

jmchilton (approved):

PR #11162, opened 2021-01-17
Python code formatting standard, automatic.

dannon:
Automatic formatting via `black` and import sorting with `isort` of the galaxy codebase.  To run the full thing locally, make sure you have python dev dependencies installed and then just run `make format`.

I tried to do this a few years ago, but recent PRs (including the one this builds on, #11155) make me hopeful that we're ready for it now.

~WIP, draft, I need to automate it and add github actions and actual dependencies, but~ I wanted to open this for discussion before putting in the extra legwork if this is going to be DOA.  I really hope it is not, though.  Prettier, even when it does weird stuff sometimes, has been a huge boon in having all the javascript code just look the same regardless of author, with no arguments or one person formatting one way, another formatting back a different way, etc.

Script example showing how to rebase pre-black branches easily here:  https://gist.github.com/ndevenish/a7a9eb93cc19abbfe80f2daab16d3367.  Note this uses 'master', and you'd likely be targeting 'dev', though.


dannon:
Hilariously enough, it looks like many of the E203 fixes in the parent PR are actually false positives that aren't inline with pep8?  (spaces in slices shouldn't trigger a violation -- https://black.readthedocs.io/en/stable/the_black_code_style.html#slices for more information)

Anyway, I'll dig more and we can certainly resolve configuration issues to get the style we want while passing basic lint.  I guess this also illustrates the point of using a tool like prettier/black/whatever -- you don't even have to think about e203 ever again, it just does it the one way for every bit of code and that's what we go with.

jmchilton:
I've now used black on real projects and my dislike it has only grown - for all the same reasons though. Prettier is great and it generates good code, I dislike the code black generates. black is collapsing those imports that will make merges and downstream forks harder, prettier was recently updated to ensure lists ends with commas, etc.. - really just a completely opposite mindset (within the spectrum of code formatting projects).

dannon:
@jmchilton Black collapses tuples without trailing commas.  For where we want the exploded import lists and easy merges of expected long lists?  Put a trailing comma.  It's actually *better* than prettier in this regard in that you can show intent and have *even cleaner* merges.

jmchilton:
After I posted that I did find some code I expected it to mangle and it indeed did not, probably because it is my code and I aggressively use trailing commas. Also upon further consideration the codebases I've worked with that used black had an unacceptably short max line lengths which resulted in a lot of ugly code, I suspect we'd not go that route.

I'm not going to merge because I have ten of thousand lines of outstanding Python in forks, but I can definitely understand the advantages and I'll retract my vote against. I would just plead that people merge outstanding green, non-WIP Python PRs before merging this 😅 .

dannon:
@jmchilton Hooray!  Yeah, I picked 120 chars to be consistent with what we use in JS (and, admittedly, what I generally set my terms to...).

And, yes, 100% -- I don't want to make extra work for anyone with outstanding changes.  There's no rush at all to merge this; I still need to do the automation/testing bits, etc., and I'm happy to keep rebuilding/rebasing over whatever we want to get in as we keep figuring things out.

I'll also see if I can figure out a simple way to update out-of-date branches to make merges easier.  Or I'll volunteer to just do it for folks, I really don't mind.

martenson:
The formatting itself seems okay-ish to me. My main concern is the increased complexity of reading the code history & git blame output. Yes, you can still get to it, but there is a hurdle in a way that you have to navigate around. Mainly for that reason I am not sure the switch is worth.

mvdbeek:
We've done a lot of this before (with prettier and flake8), I think that shouldn't stop us now.

mvdbeek:
We could also do this incrementally via `darker`: https://github.com/galaxyproject/galaxy/pull/13034

dannon:
Revisiting this for one more go.  I'll have an isolated commit for the mass application of black, and will add the run revision to a `.git-blame-ignore-revs` per https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame

pyproject.toml
```diff
@@ -91,6 +91,7 @@ zipstream-new = "*"
 
 [tool.poetry.dev-dependencies]
 cwltest = "2.2.20210901154959"
+black = "^22.1.0"
```
nsoranzo:
```suggestion
black = "^22.1.0"
cwltest = "2.2.20210901154959"
```

pyproject.toml
```diff
@@ -120,3 +121,14 @@ testfixtures = "*"
 tuspy = "*"
 twill = "*"
 watchdog = "*"
+
+[tool.black]
```
nsoranzo: I'd move this section to the top to keep the sections ordered alphabetically.

setup.cfg
```diff
@@ -7,9 +7,10 @@ universal = 1
 # E128 continuation line under-indented for visual indent
 # E402 module level import not at top of file # TODO, we would like to improve this.
 # E501 is line length
+# E203 is whitespace before ':'; we follow black's formatting here.  See https://github.com/psf/black/issues/315
 # W503 is line breaks before binary operators, which has been reversed in PEP 8.
 # D** are docstring linting - which we mostly ignore except D302. (Hopefully we will solve more over time).
-ignore = B008,E128,E501,E402,W503,D100,D101,D102,D103,D104,D105,D106,D107,D200,D201,D202,D204,D205,D206,D207,D208,D209,D210,D211,D300,D301,D400,D401,D402,D403,D412,D413
+ignore = B008,E128,E203,E501,E402,W503,D100,D101,D102,D103,D104,D105,D106,D107,D200,D201,D202,D204,D205,D206,D207,D208,D209,D210,D211,D300,D301,D400,D401,D402,D403,D412,D413
```
nsoranzo:
```suggestion
ignore = B008,E128,E203,E402,E501,W503,D100,D101,D102,D103,D104,D105,D106,D107,D200,D201,D202,D204,D205,D206,D207,D208,D209,D210,D211,D300,D301,D400,D401,D402,D403,D412,D413
```

setup.cfg
```diff
@@ -7,9 +7,10 @@ universal = 1
 # E128 continuation line under-indented for visual indent
 # E402 module level import not at top of file # TODO, we would like to improve this.
 # E501 is line length
+# E203 is whitespace before ':'; we follow black's formatting here.  See https://github.com/psf/black/issues/315
```
nsoranzo:
```suggestion
# E203 is whitespace before ':'; we follow black's formatting here. See https://black.readthedocs.io/en/stable/faq.html#why-are-flake8-s-e203-and-w503-violated
# E402 module level import not at top of file # TODO, we would like to improve this.
# E501 is line length
```

dannon:
I see what's going on with mypy, fixing.

pyproject.toml
```diff
@@ -120,3 +121,14 @@ testfixtures = "*"
...
+line-length = 120
+target-version = ['py37']
+include = '\.pyi?$'
+extend-exclude = '''
+/(
```
nsoranzo:
```suggestion
^/(
```

Otherwise directories like `lib/galaxy/tools` are excluded.

pyproject.toml
```diff
@@ -120,3 +121,14 @@ testfixtures = "*"
...
+  | packages
+  | tools
+)/
+'''
```
nsoranzo:
```suggestion
'''
force-exclude = 'lib/galaxy/util/jstree.py'
```

We also ignore this file for flake8 in `setup.cfg` . Changes to `lib/galaxy/util/jstree.py` should also be reverted, I think.

nsoranzo (commented):
Thanks @dannon ! What about integrating the changes from #13270 in here as well so we have a single commit to ignore that does both isort and black?

pyproject.toml
dannon: Will do!

dannon:
@nsoranzo I actually did originally tinker with isort in this branch back in January 2021, but Marius did a much better job setting it up in #13270.

I tested adding multiple commits to `.git-blame-ignore-revs` and it works great (including in vscode integrations/etc that use `git blame` internally)!   Given that, I don't think it's important to lump them together into a single commit, we can just easily add another entry for that change.

That said, if @mvdbeek wants to push those here for a single big PR I'm totally fine with it -- the tests that are still failing are indeed import related, which will be fixed by #13270 with 'black' style applied.

nsoranzo:
> That said, if @mvdbeek wants to push (or wants me to pull) those here for a single big PR I'm totally fine with it -- the tests that are still failing are indeed import related, which will be fixed by #13270 with 'black' style applied.

Actually the `I202` errors that you are seeing here are not going to be solved by #13270 with `profile=black` (I've tested it), but I think the solution for this is to drop `flake8-import-order` from `lib/galaxy/dependencies/lint-requirements.txt` (and its configuration from `setup.cfg`) and use the [isort GitHub action](https://pycqa.github.io/isort/docs/configuration/github_action.html).
Developers are probably better served by `darker` (through `make diff-format`) than a new tox environment for isort/black.

Makefile
```diff
@@ -41,6 +41,10 @@ setup-venv:
 diff-format:
 	$(IN_VENV) darker -r $(TARGET_BRANCH)
 
+format:
+	$(IN_VENV) black .
+	$(IN_VENV) isort .
```
nsoranzo:
```suggestion
	$(IN_VENV) isort .
	$(IN_VENV) black .
```

I feel black should have the last word?

pyproject.toml
```diff
@@ -88,8 +100,10 @@ uvloop = "*"
 WebOb = "*"
 Whoosh = "*"
 zipstream-new = "*"
+isort = "^5.10.1"
```
nsoranzo: Shouldn't `isort` go to the `[tool.poetry.dev-dependencies]` section instead?

Makefile
dannon: Sure, good idea.  I did test running them repeatedly in different orders and they're currently not in conflict (i.e. -- after they've both been run, there are no changes when running either, so order shouldn't currently matter), but it does feel like black should go last.

pyproject.toml
dannon: Good catch!  Wil fix and regenerate lock and dependencies.txts to make sure they're correct.

.git-blame-ignore-revs
```diff
@@ -1,5 +1,7 @@
 # Migrate code style to Black
 7ac461a3191fc56909da8ac3c8294f4dd4e0d543
+# Apply isort
+11207757b81157c13070c491ffa60822dcd33a57
```
mvdbeek: Maybe it'll be easier to rebase if this is one commit ? So once you hit the conflict, you checkout `ours` and then run both black and isort, otherwise you have to first check the conflicting commit to pick the right tool ?

dannon: Sure, I'll rebase those together and update the hash

mvdbeek: We should probably also have the big commit be after we setup the black and isort config, so that you already have our black and isort config in place when you hit the conflict ?

mvdbeek: (which I guess is the case anyway, otherwise it's hard to run the tools anyway. nevermind)

dannon: Yep, it'll be the big commit and then the hash update and that's the end

.isort.cfg
```diff
@@ -0,0 +1,14 @@
...
+# Override force_grid_wrap value from profile=black, but black is still happy
+force_grid_wrap=2
+# Same line length as for black
+line_length=120
+no_lines_before=LOCALFOLDER
+profile=black
+reverse_relative=true
+skip_gitignore=true
+# Make isort run faster by skipping database
+skip_glob=database/*
```
nsoranzo:
```suggestion
skip_glob=database/*
```

pyproject.toml
```diff
@@ -1,3 +1,15 @@
+[tool.black]
+line-length = 120
+target-version = ['py37']
+include = '\.pyi?$'
+extend-exclude = '''
+^/(
+  | packages
+  | tools
```
nsoranzo:
```suggestion
```

setup.cfg
```diff
@@ -5,16 +5,13 @@ universal = 1
 # These are exceptions allowed by Galaxy style guidelines:
 # B008 Do not perform function calls in argument defaults (for FastAPI Depends and Body)
 # E128 continuation line under-indented for visual indent
```
nsoranzo:
```suggestion
```

setup.cfg
```diff
@@ -5,16 +5,13 @@ universal = 1
...
+# E203 is whitespace before ':'; we follow black's formatting here. See https://black.readthedocs.io/en/stable/faq.html#why-are-flake8-s-e203-and-w503-violated
 # E402 module level import not at top of file # TODO, we would like to improve this.
 # E501 is line length
 # W503 is line breaks before binary operators, which has been reversed in PEP 8.
 # D** are docstring linting - which we mostly ignore except D302. (Hopefully we will solve more over time).
-ignore = B008,E128,E501,E402,W503,D100,D101,D102,D103,D104,D105,D106,D107,D200,D201,D202,D204,D205,D206,D207,D208,D209,D210,D211,D300,D301,D400,D401,D402,D403,D412,D413
+ignore = B008,E128,E203,E402,E501,W503,D100,D101,D102,D103,D104,D105,D106,D107,D200,D201,D202,D204,D205,D206,D207,D208,D209,D210,D211,D300,D301,D400,D401,D402,D403,D412,D413
```
nsoranzo:
```suggestion
ignore = B008,E203,E402,E501,W503,D100,D101,D102,D103,D104,D105,D106,D107,D200,D201,D202,D204,D205,D206,D207,D208,D209,D210,D211,D300,D301,D400,D401,D402,D403,D412,D413
```

dannon:
Checking through the codeql issues, it looks like it's just resolving and redetecting the same issues due to the formatting, nothing to worry about there.

.github/workflows/lint.yaml
```diff
@@ -43,3 +43,4 @@ jobs:
         run: tox -e lint_docstring_include_list
       - name: Run mypy checks
         run: tox -e mypy
+      - uses: psf/black@stable
```
nsoranzo, line 46:
```suggestion
      - uses: psf/black@stable
      - uses: jamescurtin/isort-action@master
```

pyproject.toml
dannon: @nsoranzo I was actually going to ask about this -- I left this out thinking that even though Galaxy is 3.7+ some of the tools might be deployed using older pythons.  If that's not a concern, then I'm happy to drop that exclude.

Makefile
```diff
@@ -41,6 +41,10 @@ setup-venv:
 diff-format:
```
nsoranzo:
```suggestion
diff-format:  ## Format Python code changes
```

To add this to the help print by `make` with no targe.

Makefile
```diff
@@ -41,6 +41,10 @@ setup-venv:
...
 	$(IN_VENV) darker -r $(TARGET_BRANCH)
 
+format:
```
nsoranzo:
```suggestion
format:  ## Format Python code base
```

pyproject.toml
nsoranzo: Mmmh, true, it may be that black changes the code in a 3.7+-only way.

dannon: So that's exactly what I was worried about, but it occurs to me I could apply black to tools with a 3.6 profile, run again with 3.7 and see if there are any changes.  Sound good?  Then we can maybe drop the exclude.

nsoranzo: There are many tools that don't specify any `python` requirement at all (which is clearly not good), and we could be faced with similar worries once we move to the next version of Python, I fear.

dannon: Sorry, I mean use black with the 3.6 profile (which I assume all tools are compatible with?), not the tool profile itself.

nsoranzo:
I've pushed a couple of commits to your branch, hope that's OK!

dannon:
@nsoranzo No worries, of course!  Thanks so much for the thorough review and help.  I did test toggling black's target profile from 3.6 to 3.7 and results were identical in /tools/, so I went ahead and applied it there and pushed a change including that.

nsoranzo (approved):

jmchilton (approved):
Much easier to approve with the trailing commas and with Marius and Nicola taking over the CWL branch lately. Thanks for all the work on this Dannon - really nice job.

mvdbeek (approved):
Awesome, thanks @dannon & @nsoranzo

dannon:
Thanks for the reviews, work, and approvals everyone, very happy to check this one off :)

Issue #11164, opened 2021-01-18
History import does not set `importing` attribute to False when finished


mvdbeek:
That makes imported histories not appear in histories/list, but they are present in history/view_multiple.
The column filter is here: https://github.com/galaxyproject/galaxy/blob/53b9d35863c43b7fb968bc13f6700c21aefb6738/lib/galaxy/webapps/galaxy/controllers/history.py#L148

mvdbeek:
Our friend, hidden exception placed on the stderr of the importing job:
```
Traceback (most recent call last):
  File "/Users/mvandenb/src/galaxy/lib/galaxy/tools/imp_exp/__init__.py", line 57, in cleanup_after_job
    model_store.perform_import(new_history, job=job, new_history=True)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/model/store/__init__.py", line 181, in perform_import
    self._import_jobs(object_import_tracker, history)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/model/store/__init__.py", line 692, in _import_jobs
    self._connect_job_io(imported_job, job_attrs, _find_hda, _find_hdca, _find_dce)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/model/store/__init__.py", line 931, in _connect_job_io
    imported_job.add_output_dataset_collection(output_name, output_hdca)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/model/__init__.py", line 974, in add_output_dataset_collection
    self.output_dataset_collection_instances.append(JobToOutputDatasetCollectionAssociation(name, dataset_collection_instance))
  File "<string>", line 4, in __init__
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.8/site-packages/sqlalchemy/orm/state.py", line 433, in _initialize_instance
    manager.dispatch.init_failure(self, args, kwargs)
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.8/site-packages/sqlalchemy/util/langhelpers.py", line 68, in __exit__
    compat.raise_(
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.8/site-packages/sqlalchemy/util/compat.py", line 182, in raise_
    raise exception
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.8/site-packages/sqlalchemy/orm/state.py", line 430, in _initialize_instance
    return manager.original_init(*mixed[1:], **kwargs)
[...]

PR #11165, opened 2021-01-18
[20.09] Fix import of histories with collections copied from another history

mvdbeek:
This looks like the right thing to do given what _find_hda does, and it
works for the case I encountered, but oddly I wasn't able to setup a
unit test that uses hdca_copied_from_sinks.

Fixes https://github.com/galaxyproject/galaxy/issues/11164


jmchilton:
There is a "Skipped" out test case that was dependent on this feature working I think https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy_test/api/test_workflow_extraction.py#L112

mvdbeek:
Ah, that almost passes ... the history import job is done before import is actually finished it seems, but that's another issue, I think.

jmchilton:
Yeah... I found something similar with https://github.com/galaxyproject/galaxy/blob/dev/client/src/components/HistoryExport/ToLink.vue#L157. The jobs are wrapping up before the operations are ready.

PR #11167, opened 2021-01-18
FastAPI routes for tool data tables

davelopez:
Ref #10889

## Summary
- Some additional functional tests to ensure a 404 error is returned as expected
- A test for the `reload` endpoint
- Created a [`ToolDataManager`](https://github.com/galaxyproject/galaxy/commit/372f6d6ce4dd3991a8ff736da25da71109216852#diff-37c16e129a9fcbfa5affe95b516f597fb0981ba345e5344d07dbd329a2bcaee7R10) to share as much logic as possible between the legacy and FastAPI controllers
- Added `aiofiles` as [dependency](https://github.com/galaxyproject/galaxy/commit/cdd9c7b6a777c440e01d236928a9343f2859f33a) to deal with `FileResponse` when downloading files using FastAPI

## Some additional comments
- Again, I am not 100% happy with the naming of classes, models, etc. that I have used. Please make suggestions if you feel the same :)
- The current `reload` action is modelled with a `GET`. Should it be a `POST`/`PUT` returning 200 instead of a `GET` operation since it may alter the data table? I am trying to figure out what is best practice, but changing it will break the legacy compatibility so probably not worth it.
- The pydantic `BaseModel` classes already have a `fields` property, so, an alias must be used to avoid conflicts with the `fields` attribute of [`ToolDataDetails`](https://github.com/galaxyproject/galaxy/compare/dev...davelopez:fastapi-migration-tool_data?expand=1#diff-01ebf253d063e9aaae129695c4ca04a8845deb143f290d78923a164f96190d1aR47) and [`ToolDataField`](https://github.com/galaxyproject/galaxy/compare/dev...davelopez:fastapi-migration-tool_data?expand=1#diff-01ebf253d063e9aaae129695c4ca04a8845deb143f290d78923a164f96190d1aR69) models. I have added some comments in the legacy endpoints that need to serialize the response explicitly using an alias with [`.dict(by_alias=True)`](https://github.com/galaxyproject/galaxy/compare/dev...davelopez:fastapi-migration-tool_data?expand=1#diff-db52a8a55c79b7d74d6dd77a5e6ddc3dbc5e51049067cf4fa247874c29e891b5R163)


lib/galaxy/webapps/galaxy/api/tool_data.py
```diff
@@ -1,21 +1,143 @@
...
+        field_name: str = ToolDataTableFieldName,
+        file_name: str = Path(
+            ...,  # Mark this field as required
+            title="File name",
+            description="The name of a file associated with this data table field",
+        ),
+    ):
+        """Download a file associated with the data table field."""
+        path = self.tool_data_manager.get_field_file_path(table_name, field_name, file_name)
+        return FileResponse(str(path))
```
mvdbeek, line 120: I think we'll want to do what we do in https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/web/framework/base.py#L484 somehow.
Maybe doing that as a middleware would be the best approach ? Or maybe as a dependency ?

lib/galaxy/webapps/galaxy/api/tool_data.py
```diff
@@ -24,12 +146,25 @@ def index(self, trans, **kwds):
...
     def show(self, trans, id, **kwds):
-        return self._data_table(id).to_dict(view='element')
+        """
+        GET /api/tool_data/{id}
+
+        Get details of a given tool data table.
+
+        :type   id:     str
+        :param  id:     the id of the data table
+        """
```
jmchilton: I wouldn't worry about adding docs to API endpoints we're immediately deprecating in the same PR 😁.

lib/galaxy/managers/tool_data.py
```diff
@@ -0,0 +1,81 @@
+
+from pathlib import Path
+from typing import Dict
+
+from galaxy import exceptions
+from galaxy.app import UniverseApplication
```
jmchilton: Just merged https://github.com/galaxyproject/galaxy/pull/11142/files to provide a better way of doing this I think. Any chance you can rebase and import ``galaxy.app.structured_app.StructuredApp`` instead for stronger typing and better dependency management.

lib/galaxy/managers/tool_data.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def __init__(self, app: UniverseApplication):
+        self._app = app
+
+    @property
+    def data_tables(self) -> Dict[str, TabularToolDataTable]:
+        return self._app.tool_data_tables.data_tables
+
+    @property
+    def queue_worker(self) -> GalaxyQueueWorker:
```
jmchilton: If it isn't meant to be part of the public interface of the class I guess we should prefix the name with ``_``. I think this should probably be ``_queue_worker``.

jmchilton: Actually if you just use ``self._app.queue_worker`` in the one place this is used you get all the typing benefit with the new StructureApp approach but you don't need to import ``GalaxyQueueWorker`` and make it explicit like this. I could go either way here though - this helper wouldn't be my preference but making this a little more explicit here might improve readability?

lib/galaxy/webapps/galaxy/api/tool_data.py
davelopez: You are right 😅

lib/galaxy/webapps/galaxy/api/tool_data.py
```diff
@@ -1,21 +1,143 @@
...
+    async def show(self, name: str = ToolDataTableName) -> ToolDataDetails:
+        """Get details of a given tool data table."""
+        return self.tool_data_manager.show(name)
+
+    @router.get(
+        '/api/tool_data/{name}/reload',
+        summary="Reloads a tool data table",
+        response_description="A description of the reloaded data table and its content",
+        dependencies=[
+            AdminUserRequired
```
jmchilton: Ooooo, that is kind of nice!

lib/galaxy/managers/tool_data.py
davelopez: Sure! I will do it. Let's see if this time I don't destroy the whole galaxy with my rebase attempt 😆

lib/galaxy/managers/tool_data.py
davelopez: Absolutely!, I missed it, thanks for the catch :)

davelopez: I like your suggestion of using `self._app.queue_worker` and remove the attribute and the import completely. It is used only once and it's quite readable and less code.

lib/galaxy/managers/tool_data.py
davelopez: Done!

lib/galaxy/webapps/galaxy/api/tool_data.py
```diff
@@ -1,22 +1,144 @@
...
 )
 from galaxy.webapps.base.controller import BaseAPIController
+from . import (
+    get_admin_user,
+    get_app,
+)
+
+router = APIRouter(tags=['tool data'])
+
+AdminUserRequired = Depends(get_admin_user)
```
jdavcs, line 33: This little change makes the code read so much more intuitive! `Depends=get_admin_user` - you assume you need the return value, then see it's in the route decorator - so you don't, and then finally see its purpose. `Depends=AdminUserRequired` simply doesn't raise any questions at all. Nice!

davelopez: Glad you like it :)

lib/galaxy_test/api/test_tool_data.py
```diff
@@ -44,6 +44,43 @@ def test_download_field_file(self):
...
+        self._assert_status_code_is(delete_response, 400)
+
+    def test_delete_with_wrong_values_raises_400(self):
+        payload = {"values": "wrong"}
+        delete_response = self._delete("tool_data/testbeta", data=payload, admin=True)
+        self._assert_status_code_is(delete_response, 400)
+
     # Following test case rendered invalid by the fix in
     # https://github.com/galaxyproject/galaxy/commit/48f77dc742acf01ddbafafcc4634e69378f1f020#diff-bfb557a99c1f7d646d4968d8d680b885R154.
     # TODO: Restore the test case when test framework allows actions from
```
davelopez: Any chance this can be tested now?
I have tried to uncomment the test and fix it a bit, but ended up with the error message: `Job 1 has no user and no session.` so I assume it is still not possible to run it.
Just to be sure I have tested the `DELETE` endpoint manually and it seems to work, but it would be great if we can test it properly :)

jmchilton: I'm working on it, should be possible.

jmchilton: There are two fixes in https://github.com/jmchilton/galaxy/pull/new/tool_data_fixes. Want to cherry-pick these into this branch?

davelopez: Done! Thank you so much @jmchilton

lib/galaxy/webapps/galaxy/api/tool_data.py
davelopez: @mvdbeek 
I have been playing around with something like:
````python
def add_send_file_middleware(app: FastAPI, gx_config=Depends(?)) -> None:
    @app.middleware("http")
    async def add_send_file_header(request: Request, call_next) -> Response:
        response = await call_next(request)
        if isinstance(response, StreamingResponse):
            if gx_config.nginx_x_accel_redirect_base:
                response.headers['X-Accel-Redirect'] = path_to_file
            elif gx_config.apache_xsendfile:
                response.headers['X-Sendfile'] = path_to_file
        return response
````
But I had some troubles figuring out how to get the config properly. Maybe when we merge #11180 we can look into it again? I will put it in my TODO list, promise :)

lib/galaxy_test/api/test_tool_data.py
davelopez: I will try to create the integration tests for this :)

davelopez: @jmchilton I moved the `test_delete_entry` to an integration test [here](https://github.com/galaxyproject/galaxy/pull/11167/commits/c11eca20d6ddcf6dc97b870280f0b167e1ec903a). Hope that is what you had in mind.

jmchilton: Beautiful!

lib/galaxy/webapps/galaxy/api/tool_data.py
jmchilton: The test data for these tools is relatively small so it is probably less important this works with this controller, but when we get to the datasets controller and such we will definitely need that working I think.

Issue #11168, opened 2021-01-18
History Import/Export Jobs Finish before the operations are complete


jmchilton:
It makes testing and automation tricky.

Some discussion here https://github.com/galaxyproject/galaxy/pull/11165#issuecomment-762298383.

mvdbeek:
On top of that I think they can mess up the user experience if you do manage to go to such a history (they're hidden in histories/list, but they do appear in the multiple history view):
```
DEBUG    urllib3.connectionpool:connectionpool.py:433 http://127.0.0.1:8793 "GET /api/histories/d9abeb98649a6a7e?key=0b94baa8125aeebabe6841ccffbd4e4b HTTP/1.1" 200 None
DEBUG    urllib3.connectionpool:connectionpool.py:226 Starting new HTTP connection (1): 127.0.0.1:8793
ERROR    galaxy.web.framework.decorators:decorators.py:307 Uncaught exception in exposed API method:
Traceback (most recent call last):
  File "/Users/mvandenb/src/galaxy/lib/galaxy/web/framework/decorators.py", line 294, in decorator
    rval = func(self, trans, *args, **kwargs)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/webapps/galaxy/api/history_contents.py", line 114, in index
    for content in history.contents_iter(**contents_kwds):
  File "/Users/mvandenb/src/galaxy/lib/galaxy/util/__init__.py", line 752, in merge_sorted_iterables
    yield from __merge_two_sorted_iterables(
  File "/Users/mvandenb/src/galaxy/lib/galaxy/util/__init__.py", line 770, in __merge_two_sorted_iterables
    if operator(next_2) < operator(next_1):
TypeError: '<' not supported between instances of 'int' and 'NoneType'
```
This happens when waiting for the history to get the correct length ... if we can rely on the job that'd be better, I think, because then the history gets its `importing` boolean set to false.

mvdbeek:
Fixed in https://github.com/galaxyproject/galaxy/pull/11170

PR #11170, opened 2021-01-18
[21.01] Fix history import job done before import complete

mvdbeek:
By moving history import from job wrapper cleanup to exec_after_process.
That seems like a better location overall, and in particular runs the
actual import before we set the final job state.

Fixes https://github.com/galaxyproject/galaxy/issues/11168

### DIFF
--- a/lib/galaxy/jobs/__init__.py
+++ b/lib/galaxy/jobs/__init__.py
@@ -1731,7 +1731,11 @@ class JobWrapper(HasResourceParameters):
         # Certain tools require tasks to be completed after job execution
         # ( this used to be performed in the "exec_after_process" hook, but hooks are deprecated ).
         param_dict = self.get_param_dict(job)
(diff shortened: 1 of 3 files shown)


lib/galaxy/jobs/__init__.py
```diff
@@ -1834,7 +1838,6 @@ def cleanup(self, delete_files=True):
                         if e.errno != errno.ENOENT:
                             raise
                 self.external_output_metadata.cleanup_external_metadata(self.sa_session)
-            galaxy.tools.imp_exp.JobImportHistoryArchiveWrapper(self.app, self.job_id).cleanup_after_job()
```
jmchilton, line 1837: I've always disliked this here, amazing!

jmchilton:
Such a simple change and such a massive fix, super impressive!

PR #11172, opened 2021-01-18
Fixed sync between legacy galaxy variable and Vuex current history

Nerdinacan:
Addresses flaws Marius noted with submitting tool forms and the new syncing to vuex feature. These functions are necessary to keep the Vuex betaHistoryStore in sync with the legacy Galaxy.currHistoryPanel variable so we can rewrite the legacy stuff in piecemeal.

Notably: when a tool form was used that accessed the Galaxy.currHistoryPanel, the ID was either uninitialized or the whole panel object was unavailable. That should be fixed now.

### Fix galaxy -> Vue synchronization

The goal of this sync function is to match the Vuex store to whatever is kept in galaxy.currHistoryPanel, but sometimes that variable is slow to load. This change just retries it for a while then gives up. 

On pages where the load is slower, this should catch it, but on pages where the "currHistoryPanel" variable never is initialized, it'll eventually just fail as it does now, showing with that error, indicating there's nothing to sync to.

I made the errors a little easier to consume so it wasn't so cryptic.

### Fix Vue -> galaxy synchronization

The HistoryPanelProxy facade creates a fake Backbone view for use in the legacy system. Its purpose is to look like what the rest of the legacy galaxy system wants to see when it accesses "Galaxy.currHistoryPanel". I've made some changes to keep it in sync with changes to the store.

The HistoryPanelProxy only runs when the beta panel is open (it's actually the mounter for the beta panel).

### DIFF
--- a/client/src/components/History/HistoryTopNav.vue
+++ b/client/src/components/History/HistoryTopNav.vue
@@ -35,7 +35,7 @@
                 key="use-legacy-history"
                 title="Return to legacy history panel"
                 icon="fas fa-exchange-alt"
-                @click="useLegacyHistoryPanel"
+                @click="switchToLegacyHistoryPanel"
             />
         </PriorityMenu>
     </div>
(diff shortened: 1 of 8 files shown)


mvdbeek:
Excellent, I haven't found a case where it's not working so far!

PR #11178, opened 2021-01-19
History beta poll decay time reset

Nerdinacan:
The beta history poll slows down over time so we're not asking the server for useless things during periods of inactivity, but it makes sense to speed up whenever we start manipulating content.

I've modified how the polling works to include an observable that looks at the browser XHR object and emits an event when a matching method / url are detected, this emission is used to reset the decay time on the polling.

### DIFF
--- a/client/src/components/History/providers/HistoryContentProvider/loadContents.js
+++ b/client/src/components/History/providers/HistoryContentProvider/loadContents.js
@@ -6,30 +6,30 @@ import { loadHistoryContents } from "../../caching";
 
 // prettier-ignore
 export const loadContents = (cfg = {}) => {
-    const {
-        windowSize,
-        initialInterval = 3 * 1000,
-        maxInterval = 10 * initialInterval,
-        disablePoll = false,
+    const { 
+        windowSize, 
+        initialInterval = 2 * 1000, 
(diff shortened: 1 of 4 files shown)


dannon:
This doesn't seem to be working like I'd expect.  I added debugging to the route matcher and that's firing, but it isn't resetting the decay in my hands:

image

Nerdinacan:
Thanks man, I'll beat on it some more. Was trying to have a global solution that didn't involve us touching every outgoing ajax POST.

dannon:
@Nerdinacan No worries at all -- I really like the approach, just need to get the restoration to $freshPoll working right.

Nerdinacan:
Yep got it. I think it was just misconfigured. I left a debug line in decay() so you can see it reset in the console.

dannon:
@Nerdinacan I'll see what I can figure out.  Agree it'd be great to just let prettier process these files (and notice issues), but putting all the piped stuff in a single line does indeed suck.

dannon:
@Nerdinacan Thanks.  This is working great now; I pushed a commit with minor cleanup and a typo fix but no functional changes.

dannon (approved):

PR #11180, opened 2021-01-19
Holistic approach to dependency injection.

jmchilton:
Dependency injection is cool - #10823 added the ability to write newer FastAPI controllers that can leverage typed dependency functions. This is a good way to start migrating away from ``trans`` and ``app`` anti-patterns (https://en.wikipedia.org/wiki/God_object). I love the approach, but It should not be confined to the top-layer of the application. Inversion of control (IoC) can help every layer and is a great way to de-couple various parts of the app.

This PR introduces the ability to use DI/IoC across the whole application.

## An Example

I wrote the dataset collection manager before the rest of the managers framework was filled in by Carl, and I wrote it with killing passing ``app`` around in mind. It doesn't save app and instead just grabs what it needs during object initialization. This is how all components should operate. I still needed to pass in ``app`` to get those dependencies cleanly though. Using the DI framework in this PR I'm finally able to do this without even passing ``app`` to the constructor:

```diff
@@ -37,16 +38,24 @@ class DatasetCollectionManager:
     """
     ELEMENTS_UNINITIALIZED = object()
 
-    def __init__(self, app):
+    def __init__(
+        self,
+        model: GalaxyModelMapping,
+        security: IdEncodingHelper,
+        hda_manager: HDAManager,
+        history_manager: HistoryManager,
+        tag_handler: GalaxyTagHandler,
+        ldda_manager: LDDAManager,
+    ):
         self.type_registry = DATASET_COLLECTION_TYPES_REGISTRY
         self.collection_type_descriptions = COLLECTION_TYPE_DESCRIPTION_FACTORY
-        self.model = app.model
-        self.security = app.security
+        self.model = model
+        self.security = security
 
-        self.hda_manager = hdas.HDAManager(app)
-        self.history_manager = histories.HistoryManager(app)
-        self.tag_handler = tags.GalaxyTagHandler(app.model.context)
-        self.ldda_manager = lddas.LDDAManager(app)
+        self.hda_manager = hda_manager
+        self.history_manager = history_manager
+        self.tag_handler = tag_handler
+        self.ldda_manager = ldda_manager
```

This code now doesn't need ``app`` and doesn't need to know how to construct its dependencies. We can decompose their dependencies likewise without any changes here.

## Implementation

This code brings in a framework (Lagom) to build a dependency injection framework and then adds a small function (``depends``) that adapts it to FastAPI's dependency injection mechanism ("Depends").

The advantages to this approach versus directly using FastAPI abstractions are numerous.

-  We don't need to write these little adapters for each component of the backend to adapt it FastAPI. For example look at the change to the roles API controller:

```diff
-def get_role_manager(app: UniverseApplication = Depends(get_app)) -> RoleManager:
-    return app.role_manager
-
-
 @cbv(router)
 class FastAPIRoles:
-    role_manager: RoleManager = Depends(get_role_manager)
+    role_manager: RoleManager = depends(RoleManager)
```

This is much less boilerplate. We don't need to implement & type that function get_role_manager and we don't need to bring in the import on UniverseApplication.

- We've got a clean abstraction that shields us from ``fastapi`` imports in all of our controllers. It should make it more possible to switch to new frameworks and such as the Python ecosystem matures.

- The same DI that is used to inject ``RoleManager`` into this controller is used to inject app into RoleManager when it is constructed during application initialization. This is demonstrated with the DatasetCollectionManager example above. Any component being managed by ``UniverseApplication`` can now rely on its constructor arguments to be injected if it wants. It is easy to see the cool examples on FastAPI and think it is just a technology for controllers, but it doesn't need to be.

- I don't think there is really a way to use FastAPI's dependency injection outside the context of that framework, but even if one could Lagom is superior IMO. All the auto-wiring is by type and requires zero framework and zero configuration (https://github.com/meadsteve/lagom#auto-wiring-with-zero-configuraton).

  Having a web framework provide these framework-bound extension points for injecting stuff into controllers is good and was the state of the art of Java like 15 years ago. Skipping that communities whole learning process however and using the type system and auto-wiring that isn't dependent on framework annotation really jumps us out to Java circa 8 years ago 🎆 !

- We are able to use this new depends framework to make our older controllers look a lot more like the newer ones relatively easily. 10 lines of tweaking our framework code have provided a very clear path to migrating in that direction by providing nice class dependency injection with existing controllers

```diff
-from . import BaseGalaxyAPIController
+from . import BaseGalaxyAPIController, depends
 
 log = logging.getLogger(__name__)
 
 
 class HistoriesController(BaseGalaxyAPIController, ExportsHistoryMixin, ImportsHistoryMixin, SharableMixin):
-
-    def __init__(self, app: StructuredApp):
-        super().__init__(app)
-        self.citations_manager = citations.CitationsManager(app)
-        self.user_manager = users.UserManager(app)
-        self.workflow_manager = workflows.WorkflowsManager(app)
-        self.manager = histories.HistoryManager(app)
-        self.history_export_view = histories.HistoryExportView(app)
-        self.serializer = histories.HistorySerializer(app)
-        self.deserializer = histories.HistoryDeserializer(app)
-        self.filters = histories.HistoryFilters(app)
+    citations_manager: citations.CitationsManager = depends(citations.CitationsManager)
+    user_manager: users.UserManager = depends(users.UserManager)
+    workflow_manager: workflows.WorkflowsManager = depends(workflows.WorkflowsManager)
+    manager: histories.HistoryManager = depends(histories.HistoryManager)
+    history_export_view: histories.HistoryExportView = depends(histories.HistoryExportView)
+    serializer: histories.HistorySerializer = depends(histories.HistorySerializer)
+    deserializer: histories.HistoryDeserializer = depends(histories.HistoryDeserializer)
+    filters: histories.HistoryFilters = depends(histories.HistoryFilters)
```

  These controllers are so much nicer, they don't need to use ``app`` or know how to build their dependencies. 

- Update: This is a fairly a large change and as I'm tweaking everything to get it working, I'm struck more than ever that injecting by type is nice in that it is forcing the correctness of type annotations. In Java, all the type annotations had to be correct at the outset - here making more use of them in strongly encouraging us to get them right 😆. This will have secondary benefits of more correct mypy static checking, more correct docs, and better IDE experiences.
  
  On my first pass through I had the user manager labeled as consuming the Galaxy app - but it doesn't - it needs a more generic type annotation because the tool shed also reuses all of that code. I finally understand why some places we actually do need to do ``app.model.<Class>`` instead of directly importing them (so that we can get the tool shed versions). It is a bit annoying to have to get this all typed correctly but it is going to have the secondary benefits listed above. 

## Why Lagom?

When researching DI frameworks, I couldn't find a clear winner but Lagom focus on type annotations versus annotating by name makes it clearly more modern than a lot frameworks by much bigger names (https://github.com/meadsteve/lagom/blob/master/docs/comparison.md). The other type-centric framework that had even an nicer interface that I found was punq (https://punq.readthedocs.io/en/latest/). The development just doesn't seem as active on punq.

While I didn't land up using the integration - Lagom seems to have async frameworks in mind (https://github.com/meadsteve/lagom/blob/master/lagom/integrations/fast_api.py), so that is another plus. Ultimately though I think I can swap between these two with like 10 lines of code switch, they do cool things with very simple interfaces and neither requires a bunch of investment in annotation on your components. With the latest changes to this PR, all lagom imports are confined to ``galaxy.di``, hopefully making it clear we could swap out a different dependency management system.


lib/galaxy/di.py
```diff
@@ -0,0 +1,37 @@
...
+from typing import Optional, Type, TypeVar
+
+from lagom import Container as LagomContainer
+from lagom.exceptions import UnresolvableType
+
+T = TypeVar("T")
+
+
+class Container(LagomContainer):
+    """Abstraction around lagom to provide an dependency injection context.
```
nsoranzo:
```suggestion
    """Abstraction around lagom to provide a dependency injection context.
```

lib/galaxy/di.py
```diff
@@ -0,0 +1,37 @@
...
+
+    Abstractions used by Galaxy should come through this interface so we can swap
+    out the backend as needed. For instance https://punq.readthedocs.io/en/latest/
+    containers look very nice and would allow us to also inject by name (e.g. for
+    config variables for instance).
+    """
+
+    def _register_singleton(self, dep_type: Type[T], instance: Optional[T] = None) -> T:
```
nsoranzo, line 19: This doesn't seem an internal method to me?

lib/galaxy/webapps/galaxy/api/__init__.py
```diff
@@ -4,64 +4,68 @@
...
 
 from fastapi import (
     Cookie,
-    Depends,
     Header,
     Query,
 )
-from sqlalchemy.orm import Session
+from fastapi.params import Depends
+from starlette.requests import Request
```
nsoranzo, line 23: I suppose that if we directly import this, `starlette` needs to be added to the `webapp` package and (together with `lagom`) to `pyproject.toml`.

lib/galaxy/di.py
jmchilton: It is only ever used by subclasses in their constructors or methods called by their constructors I think.

```python
lib/galaxy/app.py:        self._register_singleton(BasicApp, self)
lib/galaxy/app.py:        self._register_singleton(StructuredApp, self)
lib/galaxy/app.py:        self.config: Any = self._register_singleton(config.Configuration, config.Configuration(**kwargs))
lib/galaxy/app.py:        self.execution_timer_factory = self._register_singleton(ExecutionTimerFactory, ExecutionTimerFactory(self.config))
lib/galaxy/app.py:        self.application_stack = self._register_singleton(ApplicationStack, application_stack_instance(app=self))
lib/galaxy/app.py:        self.queue_worker = self._register_singleton(GalaxyQueueWorker, GalaxyQueueWorker(self))
lib/galaxy/app.py:        self._register_singleton(IdEncodingHelper, self.security)
lib/galaxy/app.py:        self._register_singleton(SharedModelMapping, self.model)
lib/galaxy/app.py:        self._register_singleton(GalaxyModelMapping, self.model)
lib/galaxy/app.py:        self._register_singleton(scoped_session, self.model.context)
lib/galaxy/app.py:        self.tag_handler = self._register_singleton(GalaxyTagHandler)
lib/galaxy/app.py:        self.user_manager = self._register_singleton(UserManager)
lib/galaxy/app.py:        self._register_singleton(GalaxySessionManager)
lib/galaxy/app.py:        self.hda_manager = self._register_singleton(HDAManager)
lib/galaxy/app.py:        self.history_manager = self._register_singleton(HistoryManager)
lib/galaxy/app.py:        self.job_search = self._register_singleton(JobSearch)
lib/galaxy/app.py:        self.dataset_collections_service = self._register_singleton(DatasetCollectionManager)
lib/galaxy/app.py:        self.workflow_manager = self._register_singleton(WorkflowsManager)
lib/galaxy/app.py:        self.workflow_contents_manager = self._register_singleton(WorkflowContentsManager)
lib/galaxy/app.py:        self.dependency_resolvers_view = self._register_singleton(DependencyResolversView, DependencyResolversView(self))
lib/galaxy/app.py:        self.test_data_resolver = self._register_singleton(TestDataResolver, TestDataResolver(file_dirs=self.config.tool_test_data_directories))
lib/galaxy/app.py:        self.library_folder_manager = self._register_singleton(FolderManager)
lib/galaxy/app.py:        self.library_manager = self._register_singleton(LibraryManager)
lib/galaxy/app.py:        self.role_manager = self._register_singleton(RoleManager)
lib/galaxy/app.py:        self.dynamic_tool_manager = self._register_singleton(DynamicToolManager)
lib/galaxy/app.py:        self.api_keys_manager = self._register_singleton(ApiKeyManager)
lib/galaxy/app.py:        self.file_sources = self._register_singleton(ConfiguredFileSources, ConfiguredFileSources.from_app_config(self.config))
lib/galaxy/app.py:        self.genomes = self._register_singleton(Genomes)
lib/galaxy/app.py:        self.data_provider_registry = self._register_singleton(DataProviderRegistry)
lib/galaxy/app.py:        self.job_metrics = self._register_singleton(JobMetrics, JobMetrics(self.config.job_metrics_config_file, app=self))
lib/galaxy/app.py:        self.error_reports = self._register_singleton(ErrorReports, ErrorReports(self.config.error_report_file, app=self))
lib/galaxy/app.py:        self.job_config = self._register_singleton(jobs.JobConfiguration)
lib/galaxy/app.py:        self.tool_cache = self._register_singleton(ToolCache)
lib/galaxy/app.py:        self.tool_shed_repository_cache = self._register_singleton(ToolShedRepositoryCache)
lib/galaxy/app.py:        self.watchers = self._register_singleton(ConfigWatchers)
lib/galaxy/app.py:        self.installed_repository_manager = self._register_singleton(InstalledRepositoryManager, InstalledRepositoryManager(self))
lib/galaxy/app.py:        self._register_singleton(Registry, self.datatypes_registry)
lib/galaxy/app.py:        self.data_managers = self._register_singleton(DataManagers)
lib/galaxy/app.py:        self.update_repository_manager = self._register_singleton(UpdateRepositoryManager, UpdateRepositoryManager(self))
lib/galaxy/app.py:        self.visualizations_registry = self._register_singleton(VisualizationsRegistry, VisualizationsRegistry(
lib/galaxy/app.py:        self.webhooks_registry = self._register_singleton(WebhooksRegistry, WebhooksRegistry(self.config.webhooks_dir))
lib/galaxy/app.py:        self.quota_agent = self._register_singleton(QuotaAgent, get_quota_agent(self.config, self.model))
lib/galaxy/app.py:        self.auth_manager = self._register_singleton(auth.AuthManager, auth.AuthManager(self.config))
lib/galaxy/app.py:        self.job_manager = self._register_singleton(manager.JobManager)
lib/galaxy/app.py:        self.trs_proxy = self._register_singleton(TrsProxy, TrsProxy(self.config))
lib/tool_shed/webapp/app.py:        self._register_singleton(idencoding.IdEncodingHelper, self.security)
lib/tool_shed/webapp/app.py:        self._register_singleton(SharedModelMapping, model)
lib/tool_shed/webapp/app.py:        self._register_singleton(mapping.ToolShedModelMapping, model)
lib/tool_shed/webapp/app.py:        self._register_singleton(scoped_session, self.model.context)
lib/tool_shed/webapp/app.py:        self._register_singleton(UserManager, UserManager)
lib/tool_shed/webapp/app.py:        self.auth_manager = self._register_singleton(auth.AuthManager, auth.AuthManager(self.config))
lib/tool_shed/webapp/app.py:        self.citations_manager = self._register_singleton(CitationsManager, CitationsManager(self))
lib/tool_shed/webapp/app.py:        self.quota_agent = self._register_singleton(QuotaAgent, NoQuotaAgent())
```

mvdbeek (approved):

davelopez:
I ran the selenium test locally and the timeout is triggered when trying to edit the contents of the HTML page:

image

I tried to create a page and edit it manually in a similar way but It didn't show the `not accessible error so I am not sure what can be the reason for the error :(

jmchilton:
Yeah - I was just working on it. I think it must have something to do with the way the local session interacts with the way I was fetching the app object before. I think my approach here clashed with the local session stuff.

jmchilton:
Yeah. ```trans.user != page.user``` any more even though they are the same user. 

```
INFO     galaxy.managers.base:base.py:77 item.user <galaxy.model.User(1) at 0x13ca714d0> and trans.user <galaxy.model.User(1) at 0x13cb63710>
INFO     galaxy.managers.base:base.py:78 are they != True
```

lib/galaxy/managers/session.py
```diff
@@ -6,16 +8,23 @@
...
+log = logging.getLogger(__name__)
+
 
 class GalaxySessionManager:
     """Manages GalaxySession."""
 
-    def __init__(self, model):
+    def __init__(self, model: SharedModelMapping):
         self.model = model
+        self.sa_session = model.context
```
mvdbeek, line 21: This gives you a single global session that is not going to be tied to a transaction.
```suggestion

    @property
    def sa_session(self):
        return self.model.session
```
This will give you the appropriate session.

mvdbeek: I will go through the codebase and replace all those occurences.

lib/galaxy/app.py
```diff
@@ -105,60 +119,68 @@ def __init__(self, **kwargs) -> None:
 
         # Security helper
         self._configure_security()
+        self._register_singleton(IdEncodingHelper, self.security)
+        self._register_singleton(SharedModelMapping, self.model)
+        self._register_singleton(GalaxyModelMapping, self.model)
+        self._register_singleton(scoped_session, self.model.context)
```
mvdbeek, line 125: We can only make this a singleton if we manage to update the scopting function during a request. That would be an elegant solution to the problem though.

lib/galaxy/managers/session.py
jmchilton: Yeah - this is better. I don't know why I didn't think of this. Thanks!

Here we actually use ``self.model.GalaxySession`` because we cannot import GalaxySession directly because the code is reused by the tool shed with a different class. But in other instances it would be cool to inject a ``Callable[] -> Session`` to not even require a dependency on ``ModelMapping``.

mvdbeek: I agree, and I think with a custom identity check as in 8ab79eec43cb992565fb7bf6da4026ede8916ce6 we can do this.

lib/galaxy/webapps/base/webapp.py
```diff
@@ -152,7 +153,18 @@ def _instantiate_controller(self, T, app):
...
         """
-        return T(app)
+        controller = None
+        if self.injection_aware:
+            controller = app.resolve_or_none(T)
+            if controller is not None:
+                for key, value in T.__dict__.items():
+                    if hasattr(value, "galaxy_type_depends"):
+                        value_type = value.galaxy_type_depends
+                        print(value_type)
```
mvdbeek: Is that a debug leftover ?

lib/galaxy/tools/data_manager/manager.py
```diff
@@ -16,7 +17,10 @@
 
 
 class DataManagers:
-    def __init__(self, app, xml_filename=None):
+    data_managers: Dict[str, 'DataManager']
```
mvdbeek, line 20: I haven't seen this before, does this mean the dict values need to be `DataManager` classes ?

mvdbeek (commented):
Is this still WIP @jmchilton ?

lib/galaxy/tools/data_manager/manager.py
jmchilton: Yeah - and typically one would just have ``Dict[str, DataManager]`` but the string representation is used to forward reference types that have yet to be defined in the file at the point where the type declaration is used.

lib/galaxy/webapps/base/webapp.py
jmchilton: Rebased without this.

jmchilton:
It is not, I've dropped the WIP tag.

mvdbeek (approved):

PR #11185, opened 2021-01-20
Early 21.05 dep update

dannon:
I updated these when checking for adding a new dependency.  There was a resolution conflict so I figured I'd sort it out and get it committed before it bit anyone else.


dannon:
I would guess that the the jwt-related unit test failure is likely legit; I'll take a look tomorrow and see what changed in the new version.

lib/galaxy/dependencies/pipfiles/default/Pipfile
```diff
@@ -12,7 +12,7 @@ name = "pypi"
 gunicorn = "*"
 lxml = "!=4.2.2"
 NoseHTML = "*"
-pygithub = "*"
+pygithub = "==1.54"
```
nsoranzo, line 15: Are we sure that pygithub 1.54 actually works correctly with pyjwt 2.x ?

dannon: I'll check -- been distracted for a few.

dannon:
Closing in favor of #11192, which pulls in the changes in 722260d as well.

Issue #11082, opened 2021-01-08
Option to disable tool cache


mvdbeek:
We do want to be able to disable the tool cache if it causes trouble. 21.01 has some improvements to xml macro parsing, so even without the cache loading is should be reasonably fast.

bgruening:
That would be nice. EU has still memory issues every week we update our tools, e.g. when a lot of tool installation happens.

mvdbeek:
Can't exclude it, but it would be a big surprise if the cache has any impact on memory.

PR #11186, opened 2021-01-20
[21.01] Add setting that lets admins enable/disable the tool document cache

mvdbeek:
Closes https://github.com/galaxyproject/galaxy/issues/11082

### DIFF
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -305,13 +305,14 @@ class ToolBox(BaseGalaxyToolBox):
         return self._tools_by_id
 
     def get_cache_region(self, tool_cache_data_dir):
-        if tool_cache_data_dir not in self.cache_regions:
-            self.cache_regions[tool_cache_data_dir] = ToolDocumentCache(cache_dir=tool_cache_data_dir)
-        return self.cache_regions[tool_cache_data_dir]
+        if self.app.config.enable_tool_document_cache:
+            if tool_cache_data_dir not in self.cache_regions:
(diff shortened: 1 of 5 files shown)


lib/galaxy/tools/__init__.py
```diff
@@ -305,13 +305,14 @@ def tools_by_id(self):
         return self._tools_by_id
 
     def get_cache_region(self, tool_cache_data_dir):
-        if tool_cache_data_dir not in self.cache_regions:
-            self.cache_regions[tool_cache_data_dir] = ToolDocumentCache(cache_dir=tool_cache_data_dir)
-        return self.cache_regions[tool_cache_data_dir]
+        if self.app.config.enable_tool_document_cache:
+            if tool_cache_data_dir not in self.cache_regions:
+                self.cache_regions[tool_cache_data_dir] = ToolDocumentCache(cache_dir=tool_cache_data_dir)
+            return self.cache_regions[tool_cache_data_dir]
```
bgruening, line 311: This function is now returning nothing if the tool cache is disabled. Should it return None?

nsoranzo: A function/method that doesn't explicitly return a value will return None. Explicit better than implicit, but shouldn't be a problem.

jdavcs: +1 on keeping `return None` implicit. It's a very common case, and it keeps it simple.

jdavcs (approved):

jmchilton (approved):

Issue #11174, opened 2021-01-18
Tools using singularity containers fail with low level errors if first used as a dataset collection execution


pcm32:
I'm testing the singularity support on 20.09. My current issue is that for tools that haven't been run, if the first run is a dataset collection, then multiple cluster nodes compete to get the container blobs and / or build the SIF.

Trying to get this sorted, I used the following container resolver conf file recommended by @mvdbeek :

```
<containers_resolvers>
  <explicit_singularity />
  <!-- explicit: resolves container URI for a job through explict container
       tags in the tool XML wrapper. -->

  <!-- All mulled flavors below only work if enable_beta_mulled_containers is
       set to true in the galaxy.yaml config file.

       automatic mapping, preferring cached images in the accessible docker
       engine. Requires docker engine. -->

  <cached_mulled_singularity />
  <mulled_singularity auto_install="False"/>
  <!-- cached_mulled_singularity: resolves container URI through
       bioconda to mulled automatic mapping, preferring cached singularity
       images to building local singularity images. Only works with
       enable_beta_mulled_containers set to true in the galaxy.yaml config file.
       -->

  <build_mulled_singularity auto_install="False"/>
  <!-- build_mulled_singularity: builds a singularity image locally for one
       or more package requirements listed as part of the tool's definition. -->
</containers_resolvers>
```

and I have tried using the UI admin/toolbox_dependencies to try to retrieve containers:

![image](https://user-images.githubusercontent.com/368478/104962877-47a58980-59d1-11eb-9677-9d03d65a90d5.png)

which on pressing "+ Attempt build" gives on the UI logs:

```
urllib3.connectionpool DEBUG 2021-01-18 21:04:02,014 [p:2105781,w:3,m:0] [uWSGIWorker3Core0] Starting new HTTPS connection (1): quay.io:443
urllib3.connectionpool DEBUG 2021-01-18 21:04:04,263 [p:2105781,w:3,m:0] [uWSGIWorker3Core0] https://quay.io:443 "GET /api/v1/repository?public=true&namespace=biocontainers HTTP/1.1" 200 1562773
urllib3.connectionpool DEBUG 2021-01-18 21:04:05,700 [p:2105781,w:3,m:0] [uWSGIWorker3Core0] Starting new HTTPS connection (1): quay.io:443
[...]

PR #11187, opened 2021-01-20
[20.09] Fix pulling of  singularity images if docker cli not available

mvdbeek:
That broke in https://github.com/galaxyproject/galaxy/pull/11134, since the singularity container resolve() inherits from docker container resolution it'd also check to see if `docker` was on path.

Should fix https://github.com/galaxyproject/galaxy/issues/11174 .


mvdbeek:
Found these (additional) issues and tested them during the admin training preparation.

hexylena (approved):

pcm32:
Thanks @mvdbeek , looking forward to try this!

PR #11188, opened 2021-01-20
Selenium test for ``library_import_dir``

OlegZharkov:
Implement Selenium test, that imports dataset from `library_import_dir`.

We're almost done with lib code testing, only `user_library_import_dir` left (going to work on that)

### DIFF
--- a/lib/galaxy/selenium/navigation.yml
+++ b/lib/galaxy/selenium/navigation.yml
@@ -563,10 +563,16 @@ libraries:
       download_button: '#download-btn'
       delete_btn: '.toolbtn-bulk-delete'
       toast_msg: '.toast-message'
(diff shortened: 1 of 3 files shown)


lib/galaxy/selenium/navigation.yml
```diff
@@ -563,10 +563,16 @@ libraries:
       download_button: '#download-btn'
       delete_btn: '.toolbtn-bulk-delete'
       toast_msg: '.toast-message'
+      select_import_dir_item: 'li[full_path="${name}"] .jstree-anchor'
```
dannon, line 566: Oof, I forgot jstree was here too and not just the toolshed -- one more thing to replace!

(this looks fine though)

OlegZharkov: yeah, it's obsolete code from old library. user_import_dir utilizes the same

PR #11192, opened 2021-01-20
Replace pipenv with poetry. Update all dependencies

nsoranzo:
with the updated `make update-dependencies` .


lib/galaxy/dependencies/lint-requirements.txt
```diff
@@ -0,0 +1,3 @@
+flake8
```
jmchilton, line 1: We've learned from experience we want to pin these right - otherwise we will have newer checks break older versions of Galaxy?

jmchilton:
I don't regret Pipenv - it was better than nothing, we're regularly upgrading our dependencies now. But it was really just barely better than nothing, I'm excited to see it go! Thanks so much.

jmchilton (approved):

lib/galaxy/dependencies/lint-requirements.txt
nsoranzo: Leaving these unpinned sometimes broke the CI lint jobs, yes.
I'm struggling to find a way to:
- don't have to install all dev requirements (which is a superset of the normal requirements) to run `tox -e py36-lint`
- automate flake8* pinning updates
- don't break CI lint jobs when flake8 or a plugin change the rules

all at the same time. Suggestions welcome!

jmchilton: I haven't seen anything in poetry to help :(. It is just three dependencies currently - we could just pin them at their current versions and manually modify that file?

That can still break if dependencies of dependencies break, a slightly more robust solution is we could keep it unpinned in dev and then build something into the release process to pip install that file and replace it with a locked version with ``pip freeze``.

lib/galaxy/dependencies/conda-file.sh
```diff
@@ -1,26 +0,0 @@
...
-    exit 1
-fi
-
-printf "Filtering out Galaxy requirements not available from Conda:" >&2
-egrep -iv $( \
-    conda create --override-channels -c conda-forge -c bioconda -c defaults \
-        -n _gx_test_env --dry-run --file <(sed 's/;.*//' $here/pinned-requirements.txt) python=2.7 2>&1 \
-        | grep '^\s*-' | grep -v https: | awk '{print $NF}' | paste -s -d'|' \
-) $here/pinned-requirements.txt | sed 's/;.*//'
-printf " done\n" >&2
```
jmchilton, line 26: Have you run this by Nate? I don't use it and it isn't tested so I'm okay with dropping it but I want to make sure he knows about that.

nsoranzo: I'm pretty sure that was a dead-end, we now use conda only to create an environment with Python3 and then install the requirements with pip. Any way ping @natefoo

lib/galaxy/dependencies/lint-requirements.txt
nsoranzo: I was also thinking about `pip freeze`, I'll give it a shot!
BTW, we could potentially have the same issue with mypy, should I add it to the lint requirements?

dannon:
@nsoranzo I tested a rerun of `make update-dependencies` and it looks good but I did notice the removal of the three flake8 deps.  Did something go wrong here, or were these inadvertently left in?   https://github.com/galaxyproject/galaxy/commit/54329f9cf239a22925668764dcadfadca11868dc

lib/galaxy/dependencies/lint-requirements.txt
jmchilton: Yeah - I'm sure you're right. That would be nice and very much appreciated.

nsoranzo:
> @nsoranzo I tested a rerun of `make update-dependencies` and it looks good but I did notice the removal of the flake8 and other linting deps from pinned-dev-requirements.txt in my rebuild. Did something go wrong here, or were these inadvertently left in? [54329f9](https://github.com/galaxyproject/galaxy/commit/54329f9cf239a22925668764dcadfadca11868dc)

Yes, my mistake there. I'm reworking that part as per suggestions above, will rebase and ping you. Thanks for the feedback!

lib/galaxy/dependencies/lint-requirements.txt
nsoranzo: @jmchilton Rebased with the requested changes (pinned lint requirements, including mypy).

nsoranzo:
Ready from my side.

dannon:
@nsoranzo Checking building one more time locally and then I'll merge, looks good to me.  There's no firm requirement for pyjwt2, so it's good to go ahead and be ready for that with tests still passing against 1.7

dannon (approved):

lib/galaxy/dependencies/conda-file.sh
natefoo: Yeah 👍 from me but thanks for the heads up

PR #11194, opened 2021-01-21
Fix bug: code expects dictionary, but value can be string

### LINKED ISSUE #11193: Bug in scripts/api/common.py
Accessing the `api/datatypes` endpoint via `/scripts/api/display.py` raises an AttributeError

```
(.venv) rivendell$ ./scripts/api/display.py [key] http://127.0.0.1:8080/api/datatypes
Collection Members
------------------
  File "./scripts/api/display.py", line 10, in <module>
    display(*sys.argv[1:3])
  File "/home/sergey/0dev/galaxy/_galaxy/dev/scripts/api/common.py", line 96, in display
    for k, v in i.items():  # TODO this is a str in api/datatypes route
AttributeError: 'str' object has no attribute 'items'
```

jdavcs:
Fixes #11193.

Is 20.09 the correct target?

### DIFF
--- a/scripts/api/common.py
+++ b/scripts/api/common.py
@@ -86,14 +86,17 @@ def display(api_key, url, return_formatted=True):
         print('Collection Members')
         print('------------------')
         for n, i in enumerate(r):
-            # All collection members should have a name in the response.
-            # url is optional
-            if 'url' in i:
-                print('#%d: %s' % (n + 1, i.pop('url')))
-            if 'name' in i:
-                print('  name: %s' % i.pop('name'))
-            for k, v in i.items():
-                print(f'  {k}: {v}')
+            if isinstance(i, str):
+                print('  %s' % i)
+            else:
+                # All collection members should have a name in the response.
+                # url is optional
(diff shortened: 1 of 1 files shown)


dannon:
20.09 is okay if that's what you want to target, but this bug has likely existed for all time.  Personally I don't think I'd bother with targeting anything beyond 21.01 for something like this, and if it were my PR I'd just target dev and move on.  Up to you though.  Want 20.09?  I'll merge this.

ic4f:
@dannon I'll retarget to dev.
Thank you for the explanation! I use this script daily while working on migrating endpoints to fastapi - that's how i discovered this. But I didn't know whether it was used in other contexts - so your note is very helpful. (git history can do only so much, especially for our pre-PR commits) Thanks!

ic4f:
OK, will resolve in a min..

mvdbeek:
Those scripts ... don't maybe belong into Galaxy ? A cursory look tells me they haven't been update in 10 years beyond flake8 usage.

PR #11197, opened 2021-01-21
be more verbose on output filter errors

bernt-matthias:
for `planemo test` where one/admin knows which tool is running
the message was sufficient, but not really helpful in production
settings. at least tool id or name should be included.


jmchilton (approved):

jmchilton:
Seems like a solid usability improvement, thanks so much!

Issue #11199, opened 2021-01-21
Extract dataset from list extracts all the tags


pvanheus:
If a dataset collection is created and the elements of the collection have tags, when the Extract Dataset tool is used, the extracted dataset element inherits *all* the tags of elements in the collection. This image illustrates the problem:

![image](https://user-images.githubusercontent.com/4154788/105343675-9d08b300-5bea-11eb-959a-0e4f494c6618.png)

mvdbeek:
If you want to change this you need to look at https://github.com/galaxyproject/galaxy/blob/7ffe71dd84ea67c48453d710272161db5d599124/lib/galaxy/tools/__init__.py#L2889. The right thing would be to ignore the passed in tags and copy them from the element object

pvanheus:
Thanks @mvdbeek . BTW I cannot reproduce this with the new (Beta) history view because Extract Dataset throws an error (see below) but that's a different problem.

```
The server could not complete the request. Please contact the Galaxy Team if this error persists. Error executing tool with id '__EXTRACT_DATASET__': 'str' object has no attribute 'copy'

{
    "history_id": "f597429621d6eb2b",
    "tool_id": "__EXTRACT_DATASET__",
    "tool_version": "1.0.0",
    "inputs": {
        "input": {
            "values": [
                {
                    "id": "f597429621d6eb2b",
                    "hid": 8,
                    "name": "datasets",
                    "src": "hdca",
                    "tags": []
                }
            ],
            "batch": false
        },
        "which|which_dataset": "first"
    }
}
```

nsoranzo:
Fixed in https://github.com/galaxyproject/galaxy/pull/11206 .

mmiladi:
Hello, I receive exactly the same error on the EU server and my WFs have become unfunctional. Is there any progress on this `copy` attribute error?

> Thanks @mvdbeek . BTW I cannot reproduce this with the new (Beta) history view because Extract Dataset throws an error (see below) but that's a different problem.
> 
> ```
> The server could not complete the request. Please contact the Galaxy Team if this error persists. Error executing tool with id '__EXTRACT_DATASET__': 'str' object has no attribute 'copy'
> 
> {
>     "history_id": "f597429621d6eb2b",
>     "tool_id": "__EXTRACT_DATASET__",
>     "tool_version": "1.0.0",
>     "inputs": {
>         "input": {
>             "values": [
>                 {
>                     "id": "f597429621d6eb2b",
[...]

PR #11206, opened 2021-01-22
Copy only dataset tags in Extract Dataset tool

pvanheus:
This addresses the bug reported in #11199. Instead of taking a list of tags as input, it only copies the tags from the dataset being extracted.


mvdbeek (commented):
Awesome, thank you @pvanheus. Closes https://github.com/galaxyproject/galaxy/issues/11199

PR #11210, opened 2021-01-22
Add support for the Mrc2014 image format

gregvonkuster:


gregvonkuster:
I don't know how to fix this PR so that the tests pass.  It seems that the pip requirement should not be contained within the same PR as the code that requires it.  Should it be in a different PR that gets merged first so that the tests will pass for this one?  Sorry if I've missed documentation that provides instructions for this.

gregvonkuster:
I'll close this and try separate PRs for the new requirement and these image datatypes

PR #11212, opened 2021-01-22
[21.01] Reduce authnz exception exposure

### LINKED ISSUE #11207: Unexpected authnz failures bleed through to the UI
https://github.com/galaxyproject/galaxy/blob/66e666b67833dd4f8da58f0b68292cfa7480a956/lib/galaxy/authnz/managers.py#L304 this is not optimal. I am gonna say we don't want to pass random exceptions out of the API and into the UI.

dannon:
Prevent leaking of raw exception, suggest contacting an admin (who can find the log if need be) in authnz controller.  
These definitely shouldn't go to the end user, as suggested.

fixes #11207


mvdbeek (approved):
Thank you!

PR #11213, opened 2021-01-22
[21.01] Pin mypy.

jmchilton:
Breaks the CI without this.

### DIFF
--- a/packages/test.sh
+++ b/packages/test.sh
@@ -14,7 +14,7 @@ TEST_ENV_DIR=${TEST_ENV_DIR:-$(mktemp -d -t gxpkgtestenvXXXXXX)}
 
 virtualenv -p "$TEST_PYTHON" "$TEST_ENV_DIR"
 . "${TEST_ENV_DIR}/bin/activate"
-pip install mypy
+pip install "mypy==0.790"
 
 # ensure ordered by dependency dag
 PACKAGE_DIRS=(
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -30,7 +30,7 @@ setenv =
 deps =
     lint,lint_docstring,lint_docstring_include_list: -rlib/galaxy/dependencies/pipfiles/flake8/pinned-requirements.txt
     unit: mock-ssh-server
-    mypy: mypy
+    mypy: mypy==0.790
```
nsoranzo, line 33: This line should be removed when merging forward.

packages/test.sh
```diff
@@ -14,7 +14,7 @@ TEST_ENV_DIR=${TEST_ENV_DIR:-$(mktemp -d -t gxpkgtestenvXXXXXX)}
 
 virtualenv -p "$TEST_PYTHON" "$TEST_ENV_DIR"
 . "${TEST_ENV_DIR}/bin/activate"
-pip install mypy
+pip install "mypy==0.790"
```
nsoranzo, line 17: Do we want to change this in dev to install from `lib/galaxy/dependencies/pinned-lint-requirements.txt` to keep the pinning all in the same file?

nsoranzo (approved):

packages/test.sh
jmchilton: That is a very good idea I think.

PR #11218, opened 2021-01-23
[21.01] Add ffindex and ffdata database file types

guerler:
Adds ffindex/ffdata datatypes to Galaxy. This is a simple index/database format for huge amounts of small files. The files are concatenated into one big data file (ffdata), the byte number at which the file is stored is tracked in the index file (ffindex) together with the size of the file.


bgruening:
@guerler how does an update of such a file look like? They both go together, left? A composite datatype would probably be better, similar to imzml?

guerler:
@bgruening they go together, however it is possible to use only one of them within a tool e.g. if only the identifiers are required or if a method operates on the entire database.

bgruening:
A composite datatype support to make the second file optional, so this would work. Or can both be optional?

guerler:
In principle each file can be used by itself.

bgruening:
@guerler can you target 21.01 please?

bgruening:
maybe we can also add https://github.com/soedinglab/ffindex_soedinglab as `description_url`

bgruening (approved):

bgruening:
Cool thanks!

PR #11219, opened 2021-01-25
Implement abstraction around InferringApiRouter tailored to typical Galaxy usage.

jmchilton:
The new router also allows require_admin as just an argument to the router verbs to avoid needing to import the require admin dependency.

Replaces three common imports with one cleaner one - including replacing two external imports with internal abstractions that protect us from lock in to some degree. I think the following diff demonstrates that this approach is a bit cleaner.

```diff
diff --git a/lib/galaxy/webapps/galaxy/api/roles.py b/lib/galaxy/webapps/galaxy/api/roles.py
index a7a415f6fcf..7275adecb7b 100644
--- a/lib/galaxy/webapps/galaxy/api/roles.py
+++ b/lib/galaxy/webapps/galaxy/api/roles.py
@@ -8,8 +8,6 @@
 from fastapi import (
     Body,
 )
-from fastapi_utils.cbv import cbv
-from fastapi_utils.inferring_router import InferringRouter as APIRouter
 from pydantic import (
     BaseModel,
 )
@@ -25,10 +23,10 @@
 from galaxy.schema.fields import EncodedDatabaseIdField
 from galaxy.webapps.base.controller import url_for
 from . import (
-    AdminUserRequired,
     BaseGalaxyAPIController,
     depends,
     DependsOnTrans,
+    Router,
 )
 
 log = logging.getLogger(__name__)
@@ -36,7 +34,7 @@
 
 # Empty paths (e.g. /api/roles) only work if a prefix is defined right here.
 # https://github.com/tiangolo/fastapi/pull/415/files
-router = APIRouter(tags=["roles"])
+router = Router(tags=["roles"])
 
 
 class RoleListModel(BaseModel):
@@ -53,7 +51,7 @@ def role_to_model(trans, role):
     return RoleModel(**item)
 
 
-@cbv(router)
+@router.cbv
 class FastAPIRoles:
     role_manager: RoleManager = depends(RoleManager)
 
@@ -68,7 +66,7 @@ def show(self, id: EncodedDatabaseIdField, trans: ProvidesUserContext = DependsO
         role = self.role_manager.get(trans, role_id)
         return role_to_model(trans, role)
 
-    @router.post("/api/roles", dependencies=[AdminUserRequired])
+    @router.post("/api/roles", require_admin=True)
     def create(self, trans: ProvidesUserContext = DependsOnTrans, role_definition_model: RoleDefinitionModel = Body(...)) -> RoleModel:
         role = self.role_manager.create_role(trans, role_definition_model)
         return role_to_model(trans, role)
```

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/__init__.py
+++ b/lib/galaxy/webapps/galaxy/api/__init__.py
@@ -16,6 +16,8 @@ from fastapi import (
     Query,
 )
 from fastapi.params import Depends
+from fastapi_utils.cbv import cbv
+from fastapi_utils.inferring_router import InferringRouter
 try:
     from starlette_context import context as request_context
 except ImportError:
@@ -147,3 +149,42 @@ class BaseGalaxyAPIController(BaseAPIController):
 
     def __init__(self, app: StructuredApp):
         super().__init__(app)
+
+
+class Router(InferringRouter):
+    """A FastAPI Inferring Router tailored to Galaxy.
+    """
+
+    def get(self, *args, **kwd):
+        """Extend FastAPI.get to accept a require_admin Galaxy flag."""
+        return super().get(*args, **self._handle_galaxy_kwd(kwd))
+
+    def put(self, *args, **kwd):
+        """Extend FastAPI.put to accept a require_admin Galaxy flag."""
+        return super().put(*args, **self._handle_galaxy_kwd(kwd))
+
+    def post(self, *args, **kwd):
+        """Extend FastAPI.post to accept a require_admin Galaxy flag."""
+        return super().post(*args, **self._handle_galaxy_kwd(kwd))
+
+    def delete(self, *args, **kwd):
+        """Extend FastAPI.delete to accept a require_admin Galaxy flag."""
+        return super().delete(*args, **self._handle_galaxy_kwd(kwd))
+
+    def _handle_galaxy_kwd(self, kwd):
+        require_admin = kwd.pop("require_admin", False)
+        if require_admin:
+            if "dependencies" in kwd:
+                kwd["dependencies"].append(AdminUserRequired)
+            else:
+                kwd["dependencies"] = [AdminUserRequired]
+        return kwd
+
+    @property
+    def cbv(self):
+        """Short-hand for frequently used Galaxy-pattern of FastAPI class based views.
+
+        Creates a class-based view for for this router, for more information see:
+        https://fastapi-utils.davidmontague.xyz/user-guide/class-based-views/
+        """
+        return cbv(self)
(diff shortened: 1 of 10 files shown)


lib/galaxy/webapps/galaxy/api/datatypes.py
```diff
@@ -9,14 +9,10 @@
     Union,
 )
 
-from fastapi import (
-    Depends,
-    Query,
-)
+from fastapi import Query
 from fastapi_utils.cbv import cbv
 from fastapi_utils.inferring_router import InferringRouter as APIRouter
```
davelopez: I think here those imports and the `APIRouter` can be replaced by the new `Router` :)

jmchilton: Rebased with a fix for this, thanks for the catch!

lib/galaxy/webapps/galaxy/api/__init__.py
```diff
@@ -104,3 +106,42 @@ class BaseGalaxyAPIController(BaseAPIController):
 
     def __init__(self, app: StructuredApp):
         super().__init__(app)
+
+
+class Router(InferringRouter):
+    """A FastAPI Inferring Router tailored to Galaxy.
+    """
+
+    def get(self, *args, **kwd):
```
jdavcs, line 158: The base router class also defines methods to handle patch, trace, head and options HTTP request methods. Although we might not be using all of those right now, we might at some point, and having them accept `require_admin` (or any other galaxy-specific keyword args) will keep them all consistent with the overridden get/post/put/delete. I think it'll be convenient and might save us from some future debugging. What do you think?

jmchilton: I tend toward not adding code we don't use and wouldn't have test coverage - and in fact I didn't have delete and put in here and Galaxy just failed to start with a very clear error message so it was pretty clear what the issue was so I wouldn't worry too much about debugging the problem. But I can definitely see implementing a complete solid, consistent interface as being important and I'll happily add those if you'd like.

jdavcs: Now that you mention this, even though both have merit (not adding code we don't use, and having a complete interface), I think you're right - let's not add them for now: it's not a framework after all. Thanks for considering this!

jmchilton: It is a very good point, thanks for bringing it up!

ic4f:
I'd like to approve this PR, but I can't do it without also approving the previous commit on top of which it's building - and that one if from #11180, which I am still studying (and I **do** like that one a lot, but it's a big move, and I'm, well, giving it due diligence.)

So, +1 for the router abstraction; and I'll comment on the DI PR today.

ic4f:
After #11180 is merged and this one rebased, I'll be happy to approve this. I'm sure it's all fine, but I'd rather not look through 128 changed files most of which will disappear after the rebase. (Unless you need this sooner)

PR #11220, opened 2021-01-25
Add mrc datatype (and Python dependency)

gregvonkuster:


gregvonkuster:
This mrcfille library is in preparation for supporting the Mrc2014 image file format - https://www.ccpem.ac.uk/mrc_format/mrc2014.php.  I assume the broken tests are not related to this PR, but please let me know if there is something I need to fix.  Thanks!

gregvonkuster:
@bgruening or others - when you get a chance can you let me know if this PR needs work or if it can be merged as is?  I've implemented support for 2 cryo-EM image data formats, one of which requires this library.  I'll submit PRs for them when this library is available in the installed Galaxy requirements.  Thanks!

mvdbeek:
We'll not merge this without the datatype added.

gregvonkuster:
Ah, ok thanks @mvdbeek!  Should I add it to this PR or a separate one?

mvdbeek:
yes, you can add it it to this PR

lib/galaxy/dependencies/dev-requirements.txt
```diff
@@ -89,6 +89,7 @@ markupsafe==1.1.1; (python_version >= "2.7" and python_full_version < "3.0.0") o
 mercurial==5.6.1
 mirakuru==2.3.0; python_version >= "3.6"
 mistune==0.8.4; python_version >= "3.6" and python_version < "4"
+mrcfile==1.2.0; python_version >= "3.6"
```
jmchilton, line 92: This probably doesn't belong in dev dependencies right?

I think it will need to be in packages/data/requirements.txt .

mvdbeek: Stumbled over this one too, but poetry writes out full requirements and we don't install both requirements.txt and dev-requirements.txt anymore, so this is correct, but the dependency also needs to go to packages/data/requirements.txt

gregvonkuster: Just so I'm sure, I should:
- remove the mrcfille entry from lib/galaxy/dependencies/dev-requirements.txt
- keep the current entry in lib/galaxy/dependencies/pinned-requirements.txt 
- add the entry into packages/data/requirements.txt
Thanks for the help!

mvdbeek: Only the third option, and you should add the requirement to https://github.com/galaxyproject/galaxy/blob/dev/pyproject.toml as well

gregvonkuster: Thanks a bunch @mvdbeek - I believe I've got the correct locations now.

lib/galaxy/dependencies/dev-requirements.txt
```diff
@@ -89,7 +89,6 @@ markupsafe==1.1.1; (python_version >= "2.7" and python_full_version < "3.0.0") o
 mercurial==5.6.1
 mirakuru==2.3.0; python_version >= "3.6"
 mistune==0.8.4; python_version >= "3.6" and python_version < "4"
-mrcfile==1.2.0; python_version >= "3.6"
```
mvdbeek: You need to keep that here.

lib/galaxy/dependencies/pinned-requirements.txt
```diff
@@ -79,7 +79,6 @@ markdown==3.3.3; python_version >= "3.6"
 markupsafe==1.1.1; (python_version >= "2.7" and python_full_version < "3.0.0") or (python_full_version >= "3.4.0")
 mercurial==5.6.1
 mistune==0.8.4; python_version >= "3.6" and python_version < "4"
-mrcfile==1.2.0; python_version >= "3.6"
```
mvdbeek: That also need to stay here

gregvonkuster: Ok, hopefully I have it correct now.  Thanks for the help!

lib/galaxy/config/sample/datatypes_conf.xml.sample
```diff
@@ -312,6 +312,7 @@
     <datatype extension="gii" auto_compressed_types="gz" type="galaxy.datatypes.images:Gifti" mimetype="application/octet-stream" display_in_upload="true" />
     <datatype extension="tck" type="galaxy.datatypes.images:Tck" mimetype="application/octet-stream" display_in_upload="true" />
     <datatype extension="trk" type="galaxy.datatypes.images:Trk" mimetype="application/octet-stream" display_in_upload="true" />
+    <datatype extension="mrc" type="galaxy.datatypes.images:Mrc2014" mimetype="application/octet-stream" display_in_upload="true" />
```
mvdbeek, line 315: The extension and the test dataset filename need to match.

gregvonkuster: Ah, sorry I missed that.  I started out with the longer extension and then found that the shorter extension is the one that is used.

mvdbeek:
These look like failures of the sniffer:
```
__________________ [doctest] galaxy.datatypes.sniff.guess_ext __________________
395     >>> guess_ext(fname, sniff_order)
396     'pdf'
397     >>> fname = get_test_fname('1.obo')
398     >>> guess_ext(fname, sniff_order)
399     'obo'
400     >>> fname = get_test_fname('1.arff')
401     >>> guess_ext(fname, sniff_order)
402     'arff'
403     >>> fname = get_test_fname('1.afg')
404     >>> guess_ext(fname, sniff_order)
Expected:
    'afg'
Got:
    Map ID string is incorrect: found b'CAAG', should be b'MAP '
    Invalid machine stamp: 0x41 0x41 0x43 0x54
    Invalid mode: 171601263
    Invalid axis mapping: found [857761653, 858857520, 976236858], should be [1, 2, 3]
    Error in header labels: nlabl is 1096042324 but 10 labels contain text
    File does not declare MRC format version 20140: nversion = 845229383
    Extended header type is undefined or unrecognised: exttyp = 'fish'
    Data block could not be read - file size not checked
    'afg'

/home/circleci/repo/packages/data/galaxy/datatypes/sniff.py:404: DocTestFailure
437     >>> guess_ext(fname, sniff_order, is_binary=True)
438     'hdt'
439     >>> fname = get_test_fname('1.phyloxml')
440     >>> guess_ext(fname, sniff_order)
441     'phyloxml'
442     >>> fname = get_test_fname('1.dzi')
443     >>> guess_ext(fname, sniff_order)
444     'dzi'
445     >>> fname = get_test_fname('1.tiff')
446     >>> guess_ext(fname, sniff_order)
Expected:
    'tiff'
Got:
    Map ID string is incorrect: found b'al),', should be b'MAP '
    Invalid machine stamp: 0x6d 0xb7 0x1d 0x2c
    Invalid mode: -1
    Header field 'nz' is negative
    Header field 'mx' is negative
    Header field 'my' is negative
    Header field 'mz' is negative
    Header field 'nlabl' is negative
    Cell dimension 'x' is negative
    Cell dimension 'z' is negative
    Invalid axis mapping: found [-434927847, 427215259, 946928498], should be [1, 2, 3]
    Error in header labels: nlabl is -445882299 but 10 labels contain text
    File does not declare MRC format version 20140: nversion = -1221555371
    'tiff'

/home/circleci/repo/packages/data/galaxy/datatypes/sniff.py:446: DocTestFailure
446     >>> guess_ext(fname, sniff_order)
447     'tiff'
448     >>> fname = get_test_fname('1.fastqsanger.gz')
449     >>> guess_ext(fname, sniff_order)  # See test_datatype_registry for more compressed type tests.
450     'fastqsanger.gz'
451     >>> fname = get_test_fname('1.mtx')
452     >>> guess_ext(fname, sniff_order)
453     'mtx'
454     >>> fname = get_test_fname('1imzml')
455     >>> guess_ext(fname, sniff_order)  # This test case is ensuring doesn't throw exception, actual value could change if non-utf encoding handling improves.
Expected:
    'data'
Got:
    Map ID string is incorrect: found b's="h', should be b'MAP '
    Invalid machine stamp: 0x74 0x74 0x70 0x3a
    Invalid mode: 824327534
    Invalid axis mapping: found [779581303, 791624304, 1865298807], should be [1, 2, 3]
    Error in header labels: nlabl is 1969761897 but 10 labels contain text
    File does not declare MRC format version 20140: nversion = 1701340019
    Extended header type is undefined or unrecognised: exttyp = 'xsi:'
    Data block could not be read - file size not checked
    'data'
```

gregvonkuster:
I'm a bit confused as to what is causing the errors.  The Mrc2014 sniffer shouldn't return an exception, should it?
```
    def sniff(self, filename):
         try:
             # An exception is thrown
             # if the file is not an
             # mrc2014 file.
             if mrcfile.validate(filename):
                 return True
         except Exception:
             return False
         return False
```
I placed this format at the top of the images sniff order since I thought it would be easiest to determine, but perhaps it should be in the last position?  It looks like the afg and tiff formats are not included in the sniff order.  Not sure if/how that would affect this though.

I ran the unit tests before submitting the PR and they all passed.  Are there additional tests that I should have run?

pyproject.toml
```diff
@@ -42,6 +42,7 @@ Mako = "*"
 Markdown = "*"
 MarkupSafe = "*"
 Mercurial = "*"
+mrcfile = "==1.2.0"
```
mvdbeek: Is there a reason for this pin ?

gregvonkuster: It's just the current version, and I noticed that other requirements were pinned.  I will remove it.

mvdbeek: These are only pinned if we know that's necessary. The toml file is used to update and generate the requirements.txt files that are always fully pinned.

mvdbeek:
The test failures are a weirdness of mrcfile:
https://github.com/ccpem/mrcfile/blob/master/mrcfile/validator.py#L88
combined with the bizarreness of python docstrings.
Nevertheless I don't think we want validation results on stdout, so I think you can pass an open `os.devnull` to the `print_file` argument

lib/galaxy/datatypes/images.py
```diff
@@ -285,6 +287,36 @@ def sniff_prefix(self, file_prefix):
...
+    >>> fname = get_test_fname('2.txt')
+    >>> Mrc2014().sniff(fname)
+    False
+    """
+    file_ext = 'mrc'
+
+    def sniff(self, filename):
+        # Handle the wierdness of mrcfile:
+        # https://github.com/ccpem/mrcfile/blob/master/mrcfile/validator.py#L88
+        with contextlib.redirect_stdout(None):
```
mvdbeek: Nice, I didn't know this one

mvdbeek (approved):

lib/galaxy/datatypes/images.py
nsoranzo: Should we worry about this caveat? "Note that the global side effect on `sys.stdout` means that this context manager is not suitable for use in library code and most threaded applications. It also has no effect on the output of subprocesses. However, it is still a useful approach for many utility scripts."

mvdbeek: Hmm good point, `global` and `most threaded applications` is as vague as it gets. Probably safer to pass something else in the print_file argument.

gregvonkuster: Should it be an open tmp file or something else?

mvdbeek: StringIO is probably fine.

lib/galaxy/datatypes/images.py
```diff
@@ -285,6 +288,36 @@ def sniff_prefix(self, file_prefix):
...
+
+    def sniff(self, filename):
+        # Handle the wierdness of mrcfile:
+        # https://github.com/ccpem/mrcfile/blob/master/mrcfile/validator.py#L88
+        with contextlib.redirect_stdout(StringIO()):
+            try:
+                # An exception is thrown
+                # if the file is not an
+                # mrc2014 file.
+                if mrcfile.validate(filename):
```
mvdbeek: Sorry about the misunderstanding, the idea is to use pass `StringIO()` to the `print_file` argument of  `mrcfile.validate`, to avoid the global `contextlib.redirect_stdout`

gregvonkuster: Ah, makes sense.  Sorry about this!

PR #11224, opened 2021-01-25
FastAPI route for tags (Disabled)

davelopez:
Ref #10889

## Summary
- The legacy controller logic has been refactored into a `TagManager` class shared with the new FastAPI controller.

## Additional comments
- I have placed the pydantic model in the manager module instead of a separate `_schemas.py` file because I wasn't sure where would be a better place.
- Apparently, the endpoint does not return anything. Would be better to return an empty response with a `204` status code so this is documented in the OpenAPI docs properly?
- The only test I found for this endpoint is [here](https://github.com/galaxyproject/galaxy/blob/8521a3afa0af5ff823f5dd1f1cce71136712347e/lib/galaxy_test/api/test_datasets.py#L123). It is currently failing when tested against the new FastAPI endpoint (`Bad Request`). I've been staring at it for a couple of hours but it doesn't fix itself neither I managed to figure out the reason, maybe another pair of trained eyes can help 😄

### DIFF
--- a/lib/galaxy/managers/tags.py
+++ b/lib/galaxy/managers/tags.py
@@ -10,7 +10,6 @@ from pydantic import (
 
 from galaxy.app import StructuredApp
 from galaxy.managers.context import ProvidesUserContext
-from galaxy.model.tags import GalaxyTagHandler
 
 
 class ItemTagsPayload(BaseModel):
@@ -41,22 +40,23 @@ class TagsManager:
             self,
             trans: ProvidesUserContext,
             payload: ItemTagsPayload,
-            tag_handler: GalaxyTagHandler
     ) -> None:
         """Apply a new set of tags to an item; previous tags are deleted."""
+        tag_handler = trans.app.tag_handler
         new_tags: Optional[str] = None
         if payload.item_tags and len(payload.item_tags) > 0:
             new_tags = ",".join(payload.item_tags)
-        item = self._get_item(trans, tag_handler, payload)
+        item = self._get_item(trans, payload)
         user = trans.user
         tag_handler.delete_item_tags(user, item)
         tag_handler.apply_item_tags(user, item, new_tags)
         trans.sa_session.flush()
 
-    def _get_item(self, trans: ProvidesUserContext, tag_handler: GalaxyTagHandler, payload: ItemTagsPayload):
+    def _get_item(self, trans: ProvidesUserContext, payload: ItemTagsPayload):
         """
         Get an item based on type and id.
         """
+        tag_handler = trans.app.tag_handler
         id = trans.security.decode_id(payload.item_id)
         item_class = tag_handler.item_tag_assoc_info[payload.item_class].item_class
         item = trans.sa_session.query(item_class).filter(item_class.id == id).first()
--- a/lib/galaxy/managers/tags.py
+++ b/lib/galaxy/managers/tags.py
@@ -36,6 +36,9 @@ class ItemTagsPayload(BaseModel):
         description="The list of tags that will replace the current tags associated with the item.",
     )
 
+    class Config:
+        use_enum_values = True
+
 
 class TagsManager:
     """Interface/service object shared by controllers for interacting with tags."""
--- a/lib/galaxy/managers/tags.py
+++ b/lib/galaxy/managers/tags.py
@@ -10,17 +10,13 @@ from pydantic import (
 )
 
 from galaxy.managers.context import ProvidesUserContext
+from galaxy.model import ItemTagAssociation
 from galaxy.schema.fields import EncodedDatabaseIdField
 
-
-class TaggableItemClass(str, Enum):
-    History = "History"
-    HistoryDatasetAssociation = "HistoryDatasetAssociation"
-    HistoryDatasetCollectionAssociation = "HistoryDatasetCollectionAssociation"
-    LibraryDatasetDatasetAssociation = "LibraryDatasetDatasetAssociation"
-    Page = "Page"
-    StoredWorkflow = "StoredWorkflow"
-    Visualization = "Visualization"
+taggable_item_names = {item: item for item in ItemTagAssociation.associated_item_names}
+# This Enum is generated dynamically and mypy can not statically infer it's real type
+# so it should be ignored. See:https://github.com/python/mypy/issues/4865#issuecomment-592560696
+TaggableItemClass = Enum('TaggableItemClass', taggable_item_names)  # type: ignore
 
 
 class ItemTagsPayload(BaseModel):
--- a/lib/galaxy/managers/tags.py
+++ b/lib/galaxy/managers/tags.py
@@ -8,7 +8,6 @@ from pydantic import (
     Field,
 )
 
-from galaxy.app import StructuredApp
 from galaxy.managers.context import ProvidesUserContext
 
 
@@ -33,14 +32,7 @@ class ItemTagsPayload(BaseModel):
 class TagsManager:
     """Interface/service object shared by controllers for interacting with tags."""
 
-    def __init__(self, app: StructuredApp):
-        self._app = app
-
-    def update(
-            self,
-            trans: ProvidesUserContext,
-            payload: ItemTagsPayload,
-    ) -> None:
+    def update(self, trans: ProvidesUserContext, payload: ItemTagsPayload) -> None:
         """Apply a new set of tags to an item; previous tags are deleted."""
         tag_handler = trans.app.tag_handler
         new_tags: Optional[str] = None
--- a/lib/galaxy/managers/tags.py
+++ b/lib/galaxy/managers/tags.py
@@ -1,3 +1,4 @@
+from enum import Enum
 from typing import (
     List,
     Optional,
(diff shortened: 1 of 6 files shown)


jmchilton:
> Would be better to return an empty response with a 204 status code so this is documented in the OpenAPI docs properly?

This sounds better than an empty 200.

davelopez:
## Update
- Return status 204 for the FastAPI endpoint (the legacy one will still return 200 with an empty JSON response). Also added an assert to the test to check the response was OK.
- To fix the `Bad Request` error in the new endpoint I had to change the test to send a JSON payload instead of a regular `dict`. Please let me know if there is a better way to deal with that issue.
- Unfortunately, the test is still failing (only for the new endpoint). This time the error is related to some kind of desynchronization in the database. Happening exactly when attaching the user to the `item_tag_assoc` [here](https://github.com/galaxyproject/galaxy/blob/fca1d13ffce17671da5d43ee41b60a40f223e51a/lib/galaxy/model/tags.py#L175).
````
sqlalchemy.exc.InvalidRequestError: Object '<User at 0x7fb0dc422198>' is already attached to session '17' (this is '19')
````
I don't understand where is the difference between the two endpoint calls (legacy and FastAPI) since they both share the same class. Maybe there is some more magic happening in the legacy endpoint decorator that doesn't break the database session?

**Comment by mvdbeek** [comment]
> * Unfortunately, the test is still failing (only for the new endpoint). This time the error is related to some kind of desynchronization in the database. Happening exactly when attaching the user to the `item_tag_assoc` [here](https://github.com/galaxyproject/galaxy/blob/fca1d13ffce17671da5d43ee41b60a40f223e51a/lib/galaxy/model/tags.py#L175).

This is a bigger issue, we need to always work with a SessionLocal database scope and we probably should separate this from the app object. Unfortunately I don't have a more concrete suggestion here, except that you could try tracking where those 2 different sessions come from. I need to fix something else urgently and then I will look into this (if noone beats me to it).

**Comment by jmchilton** [comment]
The difference between the end points if probably the implementation of ``trans`` - FastAPI isn't creating a full ``GalaxyWebTransaction``. Presumably something is happening in that constructor that results in a substantial difference. My guess would be this line:

https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/base/webapp.py#L170

I'm suspect based on Marius's comment this is not something we want to align between the endpoints though. So I'm not sure I have actionable advice without looking into it more.

**Comment by jmchilton** [comment]
Can you modify that whole tag pipeline so ``user`` can have type ``Union[User,str]`` and set ``user_id`` at that line instead of ``user`` if it is a string. That would be slightly more efficient and would bypass this problem. It sort of depends on us not doing something more interested with user in those functions - but we probably don't at first glace.

File: lib/galaxy/managers/tags.py [hunk 1]
```diff
@@ -0,0 +1,58 @@
+from typing import List, Optional
+
+from pydantic import (
+    BaseModel,
+    Field,
+)
+
+from galaxy.app import StructuredApp
+from galaxy.webapps.base.controller import UsesTagsMixin
+from galaxy.webapps.base.webapp import GalaxyWebTransaction
```
jmchilton: We really don't want webapps imports anywhere in app. This messes up the package dependency structure I'm trying to establish in ``packages/``. There are some slides on this here https://training.galaxyproject.org/training-material/topics/dev/tutorials/architecture/slides.html#61. 

Pretty much everything in from galaxy.webapps.base.controller predates the managers and the purpose of the managers was to replace those - so importing them here is problematic.

jmchilton: Unfortunately that mixin is doing a lot of work still it looks like:

```
ack UsesTagsMixin lib
lib/galaxy/webapps/galaxy/api/item_tags.py
9:    UsesTagsMixin
16:class BaseItemTagsController(BaseGalaxyAPIController, UsesTagsMixin):

lib/galaxy/webapps/galaxy/api/tags.py
8:from galaxy.webapps.base.controller import UsesTagsMixin
15:class TagsController(BaseGalaxyAPIController, UsesTagsMixin):

lib/galaxy/webapps/galaxy/api/users.py
47:    UsesTagsMixin
55:class UserAPIController(BaseGalaxyAPIController, UsesTagsMixin, BaseUIController, UsesFormDefinitionsMixin):

lib/galaxy/webapps/galaxy/api/history_contents.py
35:    UsesTagsMixin
42:class HistoryContentsController(BaseGalaxyAPIController, UsesLibraryMixin, UsesLibraryMixinItems, UsesTagsMixin):

lib/galaxy/webapps/galaxy/controllers/tag.py
11:from galaxy.webapps.base.controller import BaseUIController, UsesTagsMixin
16:class TagsController(BaseUIController, UsesTagsMixin):

lib/galaxy/webapps/base/controller.py
1514:class UsesTagsMixin(SharableItemSecurityMixin):
```

I would suggest a PR to replace it would a manager should precede this PR.

davelopez: Oh, I see, thanks for clarifying it!
I will try to refactor the manager to avoid importing those.

davelopez:
Thank you both for the help!
I will try the `Union[User,str]` solution proposed and see if it works.

lib/galaxy/managers/tags.py
davelopez: Is [this](https://github.com/galaxyproject/galaxy/pull/11224/commits/0eefb616a47cfd63146f0abb07d232599d1560c1) a good compromise? or maybe a bit hacky? 😅

lib/galaxy/webapps/galaxy/api/tags.py
```diff
@@ -31,7 +31,7 @@ def get_tags_manager(app: StructuredApp = Depends(get_app)) -> TagsManager:
 
 
 @cbv(router)
-class FastAPITags:
+class FastAPITags(UsesTagsMixin):
```
jmchilton: 😆 No, not this - we need to kill that mixin not expand its use. If you're not willing to bite off refactoring that mixin to become your manager and least just inject app and grab its copy of tag_handler.

jmchilton: Or - just have the manager consume ``app`` and grap the tag handler from that. This will be even cleaner down the road.

davelopez: >No, not this - we need to kill that mixin not expand its use.

Then I didn't understand it well the first time sorry😆

>Or - just have the manager consume app and grap the tag handler from that. This will be even cleaner down the road.

This brings up some pretty basic doubts I've been having about the `app` instance and its context/scope:
The `trans.app` is exactly the same instance as the `app` that gets injected in the controller? Can they be used interchangeably? or the `trans.app` is more specific to the current user making the request?

jmchilton: I could be wrong, but I think app is the same. I think trans is what is different and what it does with the thread bound database session.

jdavcs: Same here, it is my understanding that app == trans.app, at least in this context. I used this to get rid of one or the other at some refactoring step in the configuration controller.

jmchilton:
Actually I think I was a bit off:

```python
def get_trans(app: StructuredApp = depends(StructuredApp), user: Optional[User] = Depends(get_user),
              galaxy_session: Optional[model.GalaxySession] = Depends(get_session),
              ) -> SessionRequestContext:
    app.model.session.expunge_all()
    return SessionRequestContext(app=app, user=user, galaxy_session=galaxy_session)
```

It looks like in FastAPI the session is expunged after the user is loaded instead of before. Probably just moving that expunge_all line to the top of ``get_session`` would make FastAPI behave a lot more like the traditional controllers (for better or worse, I'm unsure).

lib/galaxy/webapps/galaxy/api/tags.py
davelopez: Thanks, I will get rid of one of them then :)

davelopez:
> It looks like in FastAPI the session is expunged after the user is loaded instead of before. Probably just moving that expunge_all line to the top of get_session would make FastAPI behave a lot more like the traditional controllers (for better or worse, I'm unsure).

I have tried to move that line, but it keeps giving the `sqlalchemy.exc.InvalidRequestError: Object '<User at 0x7fb0dc422198>' is already attached to session '17' (this is '19')` error. I'll try to investigate a bit further and see if I can bring some more clues.

davelopez:
The FastAPI router has been disabled until the database session issue is fixed.

lib/galaxy/managers/tags.py
```diff
@@ -0,0 +1,55 @@
...
+class ItemTagsPayload(BaseModel):
+    item_id: str = Field(
+        ...,  # This field is required
+        title="Item ID",
+        description="The `encoded identifier` of the item whose tags will be updated.",
+    )
+    item_class: str = Field(
+        ...,  # This field is required
+        title="Item class",
+        description="The name of the class of the item.",
```
mvdbeek: Can we make this an enum of all taggable item classes ?

lib/galaxy/managers/tags.py
```diff
@@ -0,0 +1,55 @@
...
+from pydantic import (
+    BaseModel,
+    Field,
+)
+
+from galaxy.managers.context import ProvidesUserContext
+
+
+class ItemTagsPayload(BaseModel):
+    item_id: str = Field(
```
mvdbeek: This should be a `EncodedDatabaseIdField`

davelopez: Good catch!

lib/galaxy/managers/tags.py
davelopez: Done! Great suggestion!
I've added to the enum the name of the classes that are listed in [GalaxyTagHandler](https://github.com/galaxyproject/galaxy/blob/6e71c75d9b4f32cb85fa179df9bfd9b94da03f06/lib/galaxy/model/tags.py#L350) but I'm not sure if those are all of them or I may have missed some.

mvdbeek: That looks complete to me. I've been thinking about maybe using a metaclass that will register new taggable classes (or https://stackoverflow.com/a/50099920), since they should all be subclass of ItemTagAssociation https://github.com/mvdbeek/galaxy/blob/5d1d29d838983748704cc89a246799e2224b2abc/lib/galaxy/model/__init__.py#L6348

But a simple enum here is fine for now.

davelopez: Thanks for the hint!
I changed the Enum and now it is dynamically generated based on the names of the subclasses of `ItemTagAssociation`. This is a much better approach since we already found that there were some classes missing:
image

However, this brings some issues:
- Mypy can not correctly detect the Enum type and the only workaround to avoid the error that I found was ignoring it.
- The OpenAPI documentation does not show the enum values as it does with the regular Enums and the JSON that represents the model does not show the `item_class` property anymore. I'm not sure why :/

mvdbeek (approved):

lib/galaxy/model/__init__.py
```diff
@@ -6350,6 +6350,11 @@ def __str__(self):
 class ItemTagAssociation(Dictifiable):
     dict_collection_visible_keys = ['id', 'user_tname', 'user_value']
     dict_element_visible_keys = dict_collection_visible_keys
+    associated_item_names: List[str] = []
+
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__(**kwargs)
+        cls.associated_item_names.append(cls.__name__.replace("TagAssociation", ""))
```
jmchilton, line 6357: I've not encountered this pattern before, very cool.

PR #11226, opened 2021-01-25
fix `test_user_library_import_dir_warning`

OlegZharkov:
This PR fixes the subsequent run of `test_user_library_import_dir.py`, sorry about that

In addition it asserts user corresponding `email` in permission field and waits for visible "saved" toast✌️

Also if we run different tests from the same file, the same instance is actually used for the both. For example `test/integration_selenium/test_user_library_import_dir.py` has 2 methods that create new library each. So when 2nd test runs, another library from the previous test still exists (didn't expect this behavior).

### DIFF
--- a/lib/galaxy/selenium/navigates_galaxy.py
+++ b/lib/galaxy/selenium/navigates_galaxy.py
@@ -914,9 +914,8 @@ class NavigatesGalaxy(HasDriver):
         if login:
             self.admin_login()
         self.libraries_open()
-        name = self._get_random_name(prefix="testcontents")
-        self.libraries_index_create(name)
-        return name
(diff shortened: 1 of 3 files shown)


OlegZharkov:
thanks @jmchilton and sorry about that test

jmchilton:
Thanks for the fix! I'm worried that toast assertion is going to fail from time to time if the timing is not right but we will see.

OlegZharkov:
> Thanks for the fix! I'm worried that toast assertion is going to fail from time to time if the timing is not right but we will see.

but we don't assert it anymore, we `wait_for_visible`. But yeah, we will see

PR #11227, opened 2021-01-25
Adding GALAXY_MEMORY_MB_PER_SLOT to k8s runner

almahmoud:
Some tools eg: https://github.com/galaxyproject/tools-iuc/blob/master/tools/rgrnastar/rg_rnaStarSolo.xml#L113 ignore `GALAXY_MEMORY_MB` and require `GALAXY_MEMORY_MB_PER_SLOT`. This is an attempt at adding both envs to each kubernetes job, rather than only the former.

### DIFF
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -307,16 +307,24 @@ class KubernetesJobRunner(AsynchronousJobRunner):
             envs = []
             if 'requests' in resources:
                 requests = resources['requests']
-                if 'memory' in requests:
-                    envs.append({'name': 'GALAXY_MEMORY_MB', 'value': str(ByteSize(requests['memory']).to_unit('M', as_string=False))})
                 if 'cpu' in requests:
-                    envs.append({'name': 'GALAXY_SLOTS', 'value': str(int(math.ceil(float(requests['cpu']))))})
+                    cpu_val = int(math.ceil(float(requests['cpu'])))
+                    envs.append({'name': 'GALAXY_SLOTS', 'value': str(cpu_val)})
+                if 'memory' in requests:
+                    mem_val = ByteSize(requests['memory']).to_unit('M', as_string=False)
+                    envs.append({'name': 'GALAXY_MEMORY_MB', 'value': str(mem_val)})
+                    if cpu_val:
+                        envs.append({'name': 'GALAXY_MEMORY_MB_PER_SLOT', 'value': str(math.floor(mem_val/cpu_val))})
             elif 'limits' in resources:
                 limits = resources['limits']
-                if 'memory' in limits:
-                    envs.append({'name': 'GALAXY_MEMORY_MB', 'value': str(ByteSize(limits['memory']).to_unit('M', as_string=False))})
                 if 'cpu' in limits:
-                    envs.append({'name': 'GALAXY_SLOTS', 'value': str(int(math.ceil(float(limits['cpu']))))})
+                    cpu_val = int(math.floor(float(limits['cpu'])))
+                    envs.append({'name': 'GALAXY_SLOTS', 'value': str(cpu_val)})
+                if 'memory' in limits:
+                    mem_val = ByteSize(limits['memory']).to_unit('M', as_string=False)
+                    envs.append({'name': 'GALAXY_MEMORY_MB', 'value': str(mem_val)})
+                    if cpu_val:
+                        envs.append({'name': 'GALAXY_MEMORY_MB_PER_SLOT', 'value': str(math.floor(mem_val/cpu_val))})
             k8s_container['resources'] = resources
             k8s_container['env'] = envs
 
--- a/lib/galaxy/jobs/runners/kubernetes.py
+++ b/lib/galaxy/jobs/runners/kubernetes.py
@@ -314,7 +314,7 @@ class KubernetesJobRunner(AsynchronousJobRunner):
                     mem_val = ByteSize(requests['memory']).to_unit('M', as_string=False)
                     envs.append({'name': 'GALAXY_MEMORY_MB', 'value': str(mem_val)})
                     if cpu_val:
-                        envs.append({'name': 'GALAXY_MEMORY_MB_PER_SLOT', 'value': str(math.floor(mem_val/cpu_val))})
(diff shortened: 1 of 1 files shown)


lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -307,16 +307,25 @@ def __get_k8s_containers(self, ajs):
...
+                    if cpu_val:
+                        envs.append({'name': 'GALAXY_MEMORY_MB_PER_SLOT', 'value': str(math.floor(mem_val/cpu_val))})
             elif 'limits' in resources:
                 limits = resources['limits']
-                if 'memory' in limits:
-                    envs.append({'name': 'GALAXY_MEMORY_MB', 'value': str(ByteSize(limits['memory']).to_unit('M', as_string=False))})
                 if 'cpu' in limits:
-                    envs.append({'name': 'GALAXY_SLOTS', 'value': str(int(math.ceil(float(limits['cpu']))))})
+                    cpu_val = int(math.floor(float(limits['cpu'])))
+                    cpu_val = cpu_val or 1
```
almahmoud, line 323: This used to be ceiling for both limits and requests, but figured we should ceil for requests but floor for limits so that there is never more being requested than the hard limit set. I don't think that we'll ever need to set a limit 0<x<1 but just in case, since requesting 0 doesn't make sense, if the limit is set to be under 1, it will still ceil it to 1.

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -307,16 +307,25 @@ def __get_k8s_containers(self, ajs):
...
-                if 'memory' in requests:
-                    envs.append({'name': 'GALAXY_MEMORY_MB', 'value': str(ByteSize(requests['memory']).to_unit('M', as_string=False))})
                 if 'cpu' in requests:
-                    envs.append({'name': 'GALAXY_SLOTS', 'value': str(int(math.ceil(float(requests['cpu']))))})
+                    cpu_val = int(math.ceil(float(requests['cpu'])))
+                    envs.append({'name': 'GALAXY_SLOTS', 'value': str(cpu_val)})
+                if 'memory' in requests:
+                    mem_val = ByteSize(requests['memory']).to_unit('M', as_string=False)
+                    envs.append({'name': 'GALAXY_MEMORY_MB', 'value': str(mem_val)})
+                    if cpu_val:
```
mvdbeek, line 317: `cpu_val` is possibly unbound here if there is no `cpu in requests`. I think you can set `cpu_val = None` before line 310.

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -307,16 +307,25 @@ def __get_k8s_containers(self, ajs):
...
-                    envs.append({'name': 'GALAXY_MEMORY_MB', 'value': str(ByteSize(limits['memory']).to_unit('M', as_string=False))})
                 if 'cpu' in limits:
-                    envs.append({'name': 'GALAXY_SLOTS', 'value': str(int(math.ceil(float(limits['cpu']))))})
+                    cpu_val = int(math.floor(float(limits['cpu'])))
+                    cpu_val = cpu_val or 1
+                    envs.append({'name': 'GALAXY_SLOTS', 'value': str(cpu_val)})
+                if 'memory' in limits:
+                    mem_val = ByteSize(limits['memory']).to_unit('M', as_string=False)
+                    envs.append({'name': 'GALAXY_MEMORY_MB', 'value': str(mem_val)})
+                    if cpu_val:
```
mvdbeek, line 328: This is also possibly unbound

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -307,16 +307,25 @@ def __get_k8s_containers(self, ajs):
...
                 if 'cpu' in requests:
-                    envs.append({'name': 'GALAXY_SLOTS', 'value': str(int(math.ceil(float(requests['cpu']))))})
+                    cpu_val = int(math.ceil(float(requests['cpu'])))
+                    envs.append({'name': 'GALAXY_SLOTS', 'value': str(cpu_val)})
+                if 'memory' in requests:
+                    mem_val = ByteSize(requests['memory']).to_unit('M', as_string=False)
+                    envs.append({'name': 'GALAXY_MEMORY_MB', 'value': str(mem_val)})
+                    if cpu_val:
+                        envs.append({'name': 'GALAXY_MEMORY_MB_PER_SLOT', 'value': str(math.floor(mem_val / cpu_val))})
             elif 'limits' in resources:
```
mvdbeek, line 319: That's a pre-existing issue, but are requests and limits completely exclusive ? I assume it should be possible to set a CPU limit and a memory request or vice-versa ?

almahmoud: The requests and limits are used for setting the actual resources of the pod in k8s, so the pod will "reserve" its request and get evicted if it goes over its limit. In general, we set both, but for the envs specified for the tool internally I think using just requests when they exist seems sensible to me

lib/galaxy/jobs/runners/kubernetes.py
almahmoud: Resolved in https://github.com/galaxyproject/galaxy/pull/11227/commits/c71db95d83e9673ca0dfa0a811be4611b0eb1c1d sorry for the oversight :)

lib/galaxy/jobs/runners/kubernetes.py
mvdbeek: So it's unlikely someone would mix requests and limits and the `elif` here is correct ?

mvdbeek (approved):

lib/galaxy/jobs/runners/kubernetes.py
almahmoud: I'm not sure tbh. I mean I think in theory requests being used when both are available is the way to go, and using limits if no requests are set is sensible. There is the more niche example of what if someone sets only cpu request and memory limit, and I guess in that scenario (assuming it is possible and we want the scenario to stay possible, but we could also potentially just limit it to be more restrictive and just not allow that?) what I would think to be ideal would be to use those individually rather than take only the CPU request and completely ignore the memory limit which is what would happen now, but I think that's a design decision that I do not have enough background to make myself I guess. I would be happy to implement something like that where request would take precedence but a combination would be possible, and it seems super easy like just a few minutes to change it, I just don't know if it's desirable or not... For our purposes, we always set all 4 for the resource sets we include by default in the chart iirc, so it would not affect us either way.

PR #11230, opened 2021-01-26
Remove `folder_up` button from data folder

### LINKED ISSUE #10800: Remove `folder_up` button from data folder
Initially suggested by @bgruening 

Would it make sense to remove `..` button from data library folder?
Since we have our great PathBar already, we don't really need `..`, do we?
image

OlegZharkov:
A minor cosmetic proposal. It removes `folder_up` button and moves breadcrumb path under the top bar
 Resolves https://github.com/galaxyproject/galaxy/issues/10800

After PR

image

Before:

image

### DIFF
--- a/lib/galaxy/selenium/navigation.yml
+++ b/lib/galaxy/selenium/navigation.yml
@@ -561,7 +561,9 @@ libraries:
       select_all: '#select-all-checkboxes'
       select_one: '.lib-folder-checkbox'
       empty_folder_message: '.empty-folder-message'
-      btn_open_upper_folder: '.btn_open_folder'
+      btn_open_parent_folder:
+        type: xpath
+        selector: '//a[contains(text(), "${folder_name}")]'
       edit_folder_btn: '.edit_folder_btn'
       description_field: '.description-field > div'
       description_field_shrinked: '.shrinked-description'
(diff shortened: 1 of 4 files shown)


lib/galaxy/selenium/navigation.yml
```diff
@@ -561,7 +561,9 @@ libraries:
       select_all: '#select-all-checkboxes'
       select_one: '.lib-folder-checkbox'
       empty_folder_message: '.empty-folder-message'
-      btn_open_upper_folder: '.btn_open_folder'
+      btn_open_parent_folder:
+        type: xpath
+        selector: '//a[contains(text(), "${folder_name}")]'
```
jmchilton, line 566: I've been using data- attributes to avoid this xpath pattern (i.e. https://github.com/galaxyproject/galaxy/pull/11160/files#diff-6383837499b568bc9eafea794ae063d6e8b74061c4bdf5501a4f7e25719f3b14R15). If we're going to use navigation.yml in tours I'm not sure if xpath will be usable (maybe that is not a problem though). I also just like annotated, structured DOM elements - but maybe that is unnecessary markup that could slow things down and shouldn't be a best practice.

OlegZharkov: I see, ok then, I will keep it in mind and thanks for the merge!

PR #11232, opened 2021-01-26
add ActionChains example in documentation of galaxy-selenium package within Jupyter

OlegZharkov:
A short example how to use ActionChains in Jupyter notebook with galaxy-selenium


jmchilton:
Thanks!

OlegZharkov:
@jmchilton thanks for this feature ;)

jmchilton:
So does the existing abstraction on ``HasDriver`` (i.e. ``action_chains()``) not work? If so, any clue why and can we create an issue for that?

PR #11234, opened 2021-01-27
Remove some seemingly unused cruft in GalaxyWebTransaction

jmchilton:
I was looking at trying to generalizing this for reuse with FastAPI, but there is a lot of functionality in there.

I don't think this stuff is used and it might be helpful to simplify before trying to pull the legacy templating stuff out for reuse in new contexts.

### DIFF
--- a/lib/galaxy/webapps/base/webapp.py
+++ b/lib/galaxy/webapps/base/webapp.py
@@ -15,7 +15,6 @@ import mako.lookup
 import mako.runtime
 from babel import Locale
 from babel.support import Translations
-from Cheetah.Template import Template
 from sqlalchemy import and_, true
 from sqlalchemy.orm.exc import NoResultFound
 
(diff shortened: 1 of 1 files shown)


jdavcs (approved):
..and +1 to "helpful to simplify before trying to pull the legacy templating stuff out"

lib/galaxy/webapps/base/webapp.py
```diff
@@ -921,12 +920,8 @@ def fill_template(self, filename, **kwargs):
         # call get_user so we can invalidate sessions from external users,
         # if external auth has been disabled.
         self.get_user()
-        if filename.endswith(".mako"):
-            return self.fill_template_mako(filename, **kwargs)
-        else:
-            template = Template(file=os.path.join(self.app.config.template_path, filename),
```
mvdbeek, line 927: Are you sure we don't need this for visualizations ? I think that might be used in https://github.com/galaxyproject/galaxy/blob/66e666b67833dd4f8da58f0b68292cfa7480a956/lib/galaxy/visualization/plugins/plugin.py#L218 ?

jmchilton:
```
# entry_point: how will this plugin render/load? mako, script tag, or static html file?
```

Totally fair question. I don't know for sure at all, but it didn't look like Cheetah was an option. @guerler do you happen to know if visualization plugins ever require Cheetah template evaluation - or is this only used for mako?

PR #11235, opened 2021-01-27
Sync x-frame-options handling across frameworks and test.

jmchilton:
Also cleanup some FastAPI cruft.

### DIFF
--- a/lib/galaxy/webapps/galaxy/buildapp.py
+++ b/lib/galaxy/webapps/galaxy/buildapp.py
@@ -31,9 +31,16 @@ class GalaxyWebApplication(galaxy.webapps.base.webapp.WebApplication):
     pass
 
 
-def app_factory(global_conf, load_app_kwds=None, **kwargs):
+def app_factory(*args, **kwargs):
     """
     Return a wsgi application serving the root object
+    """
+    return app_pair(*args, **kwargs)[0]
+
+
+def app_pair(global_conf, load_app_kwds=None, **kwargs):
+    """
+    Return a wsgi application serving the root object and the Galaxy application.
     """
     load_app_kwds = load_app_kwds or {}
     kwargs = load_app_properties(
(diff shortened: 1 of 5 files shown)


lib/galaxy/webapps/galaxy/buildapp.py
```diff
@@ -200,7 +207,7 @@ def app_factory(global_conf, load_app_kwds=None, **kwargs):
         if th.is_alive():
             log.debug("Prior to webapp return, Galaxy thread %s is alive.", th)
     # Return
-    return webapp
+    return webapp, app
```
jdavcs, line 210: Maybe I'm wrong, but can't we access `app` from `webapp`? And if we can, and if the pair as such is only used in `fast_app.py`, maybe give `fast_factory.py` a webapp and let its `factory` function extract the app object - and then pass both of them on to `fast_app.py`? (That way we postpone handling 2 different app objects until they are both needed)
I think the `app_factory` function as it is defined now is more straightforward: *it does one thing*: it builds a thing (a webapp) and returns it - i.e., it's simple. `app_pair`, on the other hand, returns 2 things that are not  meant to be used as a pair - it's just less straightforward, I think. 
I may be wrong though.

mvdbeek: The WSGI application (`webapp`) has no app, that's associated at each request cycle and app is only passed in for for the the security helper and some config values.

mvdbeek (approved):

lib/galaxy/webapps/galaxy/buildapp.py
jmchilton: > Maybe I'm wrong, but can't we access app from webapp? 

It is all wrapped up in the a bunch of middleware and I couldn't find anyway to extract app systematically from that. I definitely tried.

jdavcs: Oh, I see. Then it's fine. Thanks.

jdavcs (approved):

jmchilton:
xref https://github.com/galaxyproject/galaxy/issues/12503

PR #11239, opened 2021-01-27
[20.09] Resolve container_image_cache_path relative to data_dir

mvdbeek:
Since this is before we auto-create `container_image_cache_path` it is most likely never actually being resolved wrt data_dir.

### DIFF
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -615,7 +615,7 @@ class GalaxyAppConfiguration(BaseAppConfiguration, CommonConfigurationMixin):
             log.warning("preserve_python_environment set to unknown value [%s], defaulting to legacy_only")
             self.preserve_python_environment = "legacy_only"
         self.nodejs_path = kwargs.get("nodejs_path")
-        self.container_image_cache_path = self._in_data_dir(kwargs.get("container_image_cache_path", "container_images"))
+        self.container_image_cache_path = self._in_data_dir(kwargs.get("container_image_cache_path", "container_cache"))
         self.output_size_limit = int(kwargs.get('output_size_limit', 0))
         # activation_email was used until release_15.03
         activation_email = kwargs.get('activation_email')
--- a/lib/galaxy/config/__init__.py
+++ b/lib/galaxy/config/__init__.py
@@ -615,11 +615,7 @@ class GalaxyAppConfiguration(BaseAppConfiguration, CommonConfigurationMixin):
             log.warning("preserve_python_environment set to unknown value [%s], defaulting to legacy_only")
             self.preserve_python_environment = "legacy_only"
         self.nodejs_path = kwargs.get("nodejs_path")
-        # Older default container cache path, I don't think anyone is using it anymore and it wasn't documented - we
-        # should probably drop the backward compatiblity to save the path check.
         self.container_image_cache_path = self._in_data_dir(kwargs.get("container_image_cache_path", "container_images"))
-        if not os.path.exists(self.container_image_cache_path):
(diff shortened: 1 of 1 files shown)


lib/galaxy/config/__init__.py
```diff
@@ -615,11 +615,7 @@ def _process_config(self, kwargs):
             log.warning("preserve_python_environment set to unknown value [%s], defaulting to legacy_only")
             self.preserve_python_environment = "legacy_only"
         self.nodejs_path = kwargs.get("nodejs_path")
-        # Older default container cache path, I don't think anyone is using it anymore and it wasn't documented - we
-        # should probably drop the backward compatiblity to save the path check.
         self.container_image_cache_path = self._in_data_dir(kwargs.get("container_image_cache_path", "container_images"))
```
nsoranzo: I think that's what actually is the present default (at least it's the directory I have in my dev repo).

```suggestion
        self.container_image_cache_path = self._in_data_dir(kwargs.get("container_image_cache_path", "container_cache"))
```

nsoranzo (approved):

lib/galaxy/config/__init__.py
```diff
@@ -615,11 +615,7 @@ def _process_config(self, kwargs):
             log.warning("preserve_python_environment set to unknown value [%s], defaulting to legacy_only")
             self.preserve_python_environment = "legacy_only"
         self.nodejs_path = kwargs.get("nodejs_path")
-        # Older default container cache path, I don't think anyone is using it anymore and it wasn't documented - we
-        # should probably drop the backward compatiblity to save the path check.
-        self.container_image_cache_path = self._in_data_dir(kwargs.get("container_image_cache_path", "container_images"))
-        if not os.path.exists(self.container_image_cache_path):
-            self.container_image_cache_path = self._in_root_dir(kwargs.get("container_image_cache_path", self._in_data_dir("container_cache")))
+        self.container_image_cache_path = self._in_data_dir(kwargs.get("container_image_cache_path", "container_cache"))
```
martenson, line 618: Should `container_image_cache_path` go to galaxy.yml.sample? @mvdbeek @nsoranzo

mvdbeek: No, this is a legacy parameter and is overwritten by container_resolvers_conf.xml which gives you much more flexibility. See https://training.galaxyproject.org/training-material/topics/admin/tutorials/singularity/tutorial.html for details.

nsoranzo: Yes, I think it should be added to `lib/galaxy/webapps/galaxy/config_schema.yml`

martenson: After pulling this change our Galaxy failed to load because it wanted to write to `/mnt/volume/shared/galaxy/server/database/container_cache` that was never used before.

nsoranzo: @martenson Was your Galaxy using `/mnt/volume/shared/galaxy/server/database/container_images` ?

nsoranzo: > No, this is a legacy parameter and is overwritten by container_resolvers_conf.xml which gives you much more flexibility.

I don't agree, it's still functional AFAIU so it should be documented as legacy "please use `container_resolvers_conf.xml` (or removed).

martenson: @nsoranzo no, we are not using `/mnt/volume/shared/galaxy/server/database` at all.

nsoranzo: @martenson That can only come from your `data_dir` directory, I think. I've looked at this code a few times and it seems the standard way we treat similar config options.

PR #11240, opened 2021-01-28
FastAPI/ASGI CORS handling.

jmchilton:
- Some tests to clarify existing behavior (at least where they don't diverge). Including frameworks enhancements to allow integration tests to force on server type or the other.
- In WSGI-land return a 400 if an invalid origin is specified (to bring inline with CORSMiddleware).
- In ASGI-land, override CORSMiddleware to use Galaxy's logic for checking an origin against the config object.
- When building a WSGI webapp for FastAPI, skip CORS handling at that level so it can be handled by FastAPI app.

### DIFF
--- a/lib/galaxy/webapps/base/webapp.py
+++ b/lib/galaxy/webapps/base/webapp.py
@@ -9,6 +9,7 @@ import socket
 import string
 import time
 from http.cookies import CookieError
+from typing import Any, Dict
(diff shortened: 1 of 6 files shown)


mvdbeek:
Thank you!

jmchilton:
xref https://github.com/galaxyproject/galaxy/issues/12503

PR #2939, opened 2016-09-20
API access to genome fasta index and sequence

jj-umn:
Allows visualization components, e.g. IGV.js, to incrementally retrieve sequence as needed.

PR #11241, opened 2021-01-28
Migrate api/genomes to fastapi

jdavcs:
UPDATES: 
1. This is ready for review. 
2. I did not create Pydantic models because that requires modifying the underlying `visualization/genomes.py`, which I'm reluctant to do because its core functionality has no test coverage. Adding models in a minimally intrusive way may be possible, but I'm still weary of doing that without any tests. Adding tests is an option, if it's worth the effort (see https://github.com/galaxyproject/galaxy/pull/11241#discussion_r567140081)
3. Route and parameter metadata is minimalist. I included what I could infer from `genomes.py`, but didn't try to guess further.

THE FOLLOWING IS MOSTLY OUTDATED:
Contains some initial fixes to `api/genomes` before refactoring towards fastapi. Purpose: see if this breaks things.

I'd like some feedback on the following 2 items:

1. Following suggestion from @dannon, I've replaced the `@web` decorators in the controller with `@expose_api*`. Main reason: error handling. The `@web.json` decorator caused any error to be returned as an html document, sent as json in the response (`content-type: application/json`). Which has to be incorrect. See [discussion on gitter](https://gitter.im/galaxyproject/wg-backend?at=60121c8da0246860dc152a96) for more context. The tests pass, but I am not at all sure this won't break something else.

1. I haven't yet figured out how to manually verify the `api/genomes/{id}/indexes` and the `api/genomes/{id}/sequences` endpoints. I'm also not sure they are operational: they cause an internal server error on both usegalaxy.org and usegalaxy.eu. If they are supposed to be operational, I'll add proper error handling.

I've added additional error handling and tests for 2 endpoints (so far). I've also wrapped opening files in a context manager in `visualization/genomes.py` (there are no tests there, so I applied minimal edits with no further refactoring or cleanup).


lib/galaxy/webapps/galaxy/api/genomes.py
```diff
@@ -15,15 +18,14 @@ class GenomesController(BaseAPIController):
     RESTful controller for interactions with genome data.
     """
 
-    @web.legacy_expose_api_anonymous
+    @expose_api_anonymous
```
jmchilton, line 174: 🎆 

The other backend migration we're halfway through, any chance you can make an issue to track our remaining instances of legacy decorators? Backend team is rocking these checklists 😄.

lib/galaxy_test/api/test_genomes.py
```diff
@@ -0,0 +1,24 @@
...
+        response = self._get('genomes')
+        self._assert_status_code_is(response, 200)
+        assert response.json()[0][0] == 'unspecified (?)'
+
+    def test_show_invalid(self):
+        response = self._get('genomes/invalid')
+        self._assert_status_code_is(response, 404)
+        assert response.json()['err_code'] == ObjectNotFound.err_code.code
+
+    def test_show_missing_data(self):
```
jmchilton: I'm tempted to request making this an integration test? I don't think it is valid to assume the API tests will be run without this set - we can sort of guarantee this is the case with an integration test though.

While that would be the correct thing to do, I'm worried it is overkill for this API if the API isn't being used. Do you know if this API is being used by the GUI?

lib/galaxy/webapps/galaxy/api/genomes.py
jdavcs: Of course! Writing it up now.

lib/galaxy_test/api/test_genomes.py
jdavcs: The `api/genomes` and `api/genomes/{id}` endpoints are used by the client (for trackster and more), so I don't think it would be overkill. Besides, this test fails: it passes locally, so I think the failure is precisely for the reason you bring up: `anoGam3` doesn't exist on the test instance, so it's raising a different error (i.e., the test isn't testing what is should be testing). I'll write an integration test for this.

lib/galaxy/webapps/galaxy/api/genomes.py
jdavcs: @jmchilton, just to clarify, do you mean all instances of the `legacy*` decorators used in `galaxy/webapps`? (seems to be 88 instances of usage)?

jmchilton: Yup - thanks!

lib/galaxy_test/api/test_genomes.py
jmchilton: Sounds good! Thanks for the update.

jdavcs: If these integration tests pass remotely, I'll refactor them (wrap each in a context manager, at least). The tests use minimal patching and cover essential functionality only; they ensure that the new error handling works as expected (`MissingDataError` raised when ref data for a valid genome is missing, and `ObjectNotFound` when the provided genome id is invalid), and that data is returned as expected. (UPDATE: used integration tests with minimal mocking)

test/integration/test_genomes.py
```diff
@@ -0,0 +1,69 @@
+from galaxy.exceptions import (
+    MissingDataError,
+    ObjectNotFound,
+)
+from galaxy.visualization.genomes import Genome
+from galaxy_test.driver import integration_util
+
+
+class MockGenome:
```
jmchilton: I would feel better if instead of using a mock and breaking into app - you tweaked the actual configuration that Galaxy loads in order to produce this. I'll admit it is a personal preference though. I suspect we don't have any test coverage of `` app.genomes``.

jdavcs: No, you're right - it's an objectively better approach. I wasn't sure how difficult it would be to setup the required input data in this case - not so difficult after all. 

There is no test coverage of `app.genomes`, and, in my opinion, in its current form it is mostly untestable. Making it testable requires refactoring, which requires adding unit tests for its core functionality - and that may be a time-consuming task. Assuming this is not high priority, I'd suggest leaving it as is: the current integration tests cover the most general use cases of the available endpoints, including the corrected error handling. Although I do need to add tests for the 2 remaining endpoints which currently result in errors that are not properly handled - usegalaxy.[org|eu] included.  But if proper test coverage of this module *is* high priority, I'll take this on, of course.

ic4f:
I need some help understanding [this line](https://github.com/galaxyproject/galaxy/blob/release_20.09/lib/galaxy/webapps/galaxy/api/genomes.py#L74): why do we need `reference` here? We can, of course, do `api/genomes/indexes/[key]/sequences/?reference=true`, but why? It has no further meaning past the `assert` statement. @jmchilton - I've traced it back here: https://github.com/galaxyproject/galaxy/pull/2939 - can you help me understand this? (you fixed [this error](https://github.com/galaxyproject/galaxy/pull/2939/commits/97532515ea24325ba26877696991b24401733ee6#diff-26e6630e019a79bc864af56fe78cbe8ebcdeea9c44627dcab4eb3c281e0a01cfR81) in the initial PR)

Also, regarding the other endpoint: `api/genomes/[id]/indexes`: it returns [whatever it reads from an index file](https://github.com/galaxyproject/galaxy/blob/release_20.09/lib/galaxy/webapps/galaxy/api/genomes.py#L63) *without closing the file object*. Is that by design, or an oversight?

I'm inclined to:
- remove the `reference` keyword requirement in the first case
- raise a MissingDataError [here](https://github.com/galaxyproject/galaxy/blob/release_20.09/lib/galaxy/visualization/genomes.py#L372) instead of returning `None` - then it will be properly handled;
- close the file and return a string in the second case.

Is this a reasonable fix? I don't know the context in which this is used - I read the discussion in #2939, but that didn't help. Again, accessing these enpoints on both usegalaxy.org and eu results in a 500 server error.

ic4f:
I couldn't find a way to setup the proper test configuration to test the remaining endpoints (`sequences` and `indexes`), so I applied minimal patches instead (I'll rework these if there's a better way). The tests are not pretty, but they verify what the endpoints are supposed to do, given the instance has access to the required reference data. I've also added proper error handling. Since the underlying `genomes.py` is not covered by tests, these are necessary for any further refactoring of the api endpoints.

davelopez (approved):
Looks perfect to me :)

lib/galaxy/exceptions/__init__.py
```diff
@@ -221,6 +221,11 @@ class InvalidFileFormatError(MessageException):
     err_code = error_codes_by_name['INVALID_FILE_FORMAT']
 
 
+class ReferenceDataError(MessageException):
+    status_code = 500
+    err_code = error_codes_by_name['REFERENCE_DATA_ERROR']
```
mvdbeek, line 226: Sounds more like an invalid request ? 500 doesn't seem right here

lib/galaxy/managers/genomes.py
```diff
@@ -0,0 +1,78 @@
...
+        index_filename = self._get_index_filename(id, tbl_entries, ext, index_type)
+
+        try:
+            fh = open(index_filename, mode='r')
+        except OSError:
+            raise ReferenceDataError(f'Failed to load index file for {id}')
+        else:
+            return fh.read()
+        finally:
+            fh.close()
```
mvdbeek: Can you use a with statement here ? 

I'd also say this should just return a open file handle so our middleware can do the appropriate type of serving, BUT, I don't think this is worth doing for this endpoint in this PR. There are other endpoint though that must do this. I'm not sure how that will look like, but https://github.com/mvdbeek/galaxy/blob/12d27397659b28902e4aee2ab7f668fbaf160197/lib/galaxy/web/framework/base.py#L484 is how we do this in the old wsgi part.

jdavcs: Made it a with statement. Thanks for the reference to the code - I'll use it for other endpoints (which I can more easily test).

lib/galaxy/exceptions/__init__.py
jdavcs: I don't mind changing it, but, I thought it made more sense as a server-side error. So, just in case, to double check, here's my reasoning:
Here's an example of an invalid request: https://github.com/galaxyproject/galaxy/blob/16a420c9660819a3f7c15e7775bd470f42b80d1e/lib/galaxy/managers/genomes.py#L54 - i.e., the supplied index type is invalid.
But in the all cases where I use `ReferenceDataError` it is on the server end: the key is valid, but Galaxy can't find the reference data for that key, or can't find any reference data for that index type, etc. - which may happen for various reasons:
https://github.com/galaxyproject/galaxy/blob/16a420c9660819a3f7c15e7775bd470f42b80d1e/lib/galaxy/managers/genomes.py#L63 :point_up: also see line 70 and 71; 
and here:
https://github.com/galaxyproject/galaxy/blob/16a420c9660819a3f7c15e7775bd470f42b80d1e/lib/galaxy/visualization/genomes.py#L95 :point_up: and line 375.

If this can be treated as invalid requests, I'll change it to `RequestParameterInvalidException`.

mvdbeek: Alright, that makes sense. Thanks @ic4f!

mvdbeek (approved):

PR #11243, opened 2021-01-28
[21.01] Fix arguments in create_api_key

davelopez:
This was causing an error when trying to create a new API key:
error

````
TypeError: create_api_key() takes 2 positional arguments but 3 were given
````

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/users.py
+++ b/lib/galaxy/webapps/galaxy/api/users.py
@@ -761,7 +761,7 @@ class UserAPIController(BaseAPIController, UsesTagsMixin, BaseUIController, Uses
         """
         payload = payload or {}
         user = self._get_user(trans, id)
-        self.api_key_manager.create_api_key(trans, user)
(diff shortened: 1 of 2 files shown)


nsoranzo:
Broken in https://github.com/galaxyproject/galaxy/pull/10522 . Can you target `release_21.01` ?

So it looks like we don't have a API test for `PUT /api/users/{id}/api_key/inputs` ? https://github.com/galaxyproject/galaxy/blob/release_21.01/lib/galaxy/webapps/galaxy/buildapp.py#L771-L775

davelopez:
Ok @nsoranzo.
I will try to add a little test too.

davelopez:
I think this is ready for review now.
Sorry for the mess with the base branch... 😅

dannon:
Thanks for the fix, and for adding a test!

nsoranzo (approved):

dannon (approved):

Issue #11109, opened 2021-01-12
Improve clarity of dataset info page


simonbray:
The dataset info page located at `/datasets/{dataset_id}/show_params` could be presented more clearly - some ideas:

Paper-cut level:

- Make it clearer what items are links. For example, the colour of the link text could be made more distinct (I can barely tell the difference) and relevant icons added.
- Make more of the items links:
  - Tool ID can link to the toolshed
  - History ID can link back to the history (`/histories/view?id={history_id}`)
  - Job ID can link to the job rerun page
  - Dataset ID can link to the dataset
  - 'Name' and 'Format' could link to the 'Edit attributes' page
- Datasets can be part of multiple histories, all history IDs should be listed
- Better explanation of what some items mean, e.g. 'Number', 'Dbkey', 'Inheritance chain'. Maybe this can be shown as a mouseover text.

More advanced (need some simple changes on the backend):
- It could be stated if the dataset is part of a collection and if so the collection ID (again with a link)
- It could be stated if the job was created as part of a workflow invocation - if so, workflow and invocation IDs can be provided (with links)

PR #11244, opened 2021-01-28
Some minor changes to the dataset show_params page

simonbray:
Implemented some of the suggestions in https://github.com/galaxyproject/galaxy/issues/11109


client/src/components/DatasetInformation/DatasetInformation.vue
```diff
@@ -37,13 +37,25 @@
                     <tr v-if="dataset.id">
                         <td>History Content API ID</td>
                         <td>
-                            <div id="dataset-id">{{ dataset.id }} <decoded-id :id="dataset.id" /></div>
+                            <div id="dataset-id">
+                                {{ dataset.id }} <decoded-id :id="dataset.id" /> (<a
```
mvdbeek: This should go into the JobInformation.vue file and it probably needs to use the job id?

lib/galaxy/model/__init__.py
```diff
@@ -1136,6 +1136,12 @@ def to_dict(self, view='collection', system_details=False):
         rval = super().to_dict(view=view)
         rval['tool_id'] = self.tool_id
         rval['history_id'] = self.history_id
+        if self.workflow_invocation_step:
+            rval['created_by_workflow'] = True
+            rval['workflow_invocation_id'] = self.workflow_invocation_step.workflow_invocation.id
+            rval['workflow_id'] = self.workflow_invocation_step.workflow_invocation.workflow_id
+        else:
+            rval['created_by_workflow'] = False
```
mvdbeek: This is really muddying the waters here, can you please remove this section?

mvdbeek: While my initial reaction is that I don't like this, of course this would be useful, but this is a bit of a "details" view that we don't want to return whenever someone hits the job api endpoint.

I would suggest adding another view here and then serializing the workflow invocation step (`rval['workflow_invocation_step'] = self.workflow_invocation_step.to_dict() if self.workflow_invocation_step else {}`). You can to add the `workflow_invocation_id` in https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/model/__init__.py#L5670-L5671.

The job component can then fetch the additional details from the vuex store if it is appropriate.

client/src/components/DatasetInformation/DatasetInformation.vue
```diff
@@ -37,13 +37,25 @@
...
+                            </div>
                         </td>
                     </tr>
                     <tr v-if="dataset.history_id">
                         <td>History API ID</td>
                         <td>
-                            <div id="history_id">{{ dataset.history_id }} <decoded-id :id="dataset.history_id" /></div>
+                            <div id="history_id">
+                                {{ dataset.history_id }} <decoded-id :id="dataset.history_id" /> (<a
+                                    :href="'/histories/view?id=' + dataset.history_id"
```
mvdbeek: I think we want to get away from this direct navigation, but I'm not sure what the replacement is. @dannon is that how you would do this ? Might also be better to have a button for this ?

client/src/components/JobInformation/CodeRow.vue
```diff
@@ -28,9 +28,9 @@ export default {
         return { expanded: false };
     },
     methods: {
-        toggleExpanded() {
+        expandCode() {
             if (this.codeItem) {
-                this.expanded = !this.expanded;
+                this.expanded = true;
```
mvdbeek: Why this change ? I think it's quite useful being able to toggle long outputs ?

client/src/components/JobInformation/JobInformation.vue
```diff
@@ -5,12 +5,30 @@
...
+                    <td id="galaxy-tool-id">
+                        {{ job.tool_id }} (<a :href="'/root?tool_id=' + job.tool_id" target="_top">run</a>)
+                    </td>
                 </tr>
                 <tr v-if="job && job.tool_version">
                     <td>Galaxy Tool Version:</td>
                     <td id="galaxy-tool-version">{{ job.tool_version }}</td>
                 </tr>
+                <tr v-if="job">
+                    <td>Created by workflow?</td>
```
mvdbeek, line 24: We use this component also when displaying workflow invocations, so this would be redundant there. I think there's some logic for making this customizable with slots, I think that'd be a good use for parameterizing whether to display this or not.

mvdbeek: Or a `v-if` and a new prop would work too.

client/src/components/JobInformation/CodeRow.vue
simonbray: See https://github.com/galaxyproject/galaxy/pull/10980#issuecomment-765350030 and @OlegZharkov's suggestion underneath.

mvdbeek: Can you come up with another solution ? It should be possible to use a combination of on:click and on:mousedown to prevent the toggle. Alternatively you could add an expand button that is dedicated to toggling the expanded state.

client/src/components/JobInformation/JobInformation.vue
simonbray: Something like this @mvdbeek: https://github.com/galaxyproject/galaxy/pull/11244/commits/3d2a1fb3252ffd0067b49d26a8a92b5ad681dfd4 ?

It is not working currently, unfortunately.

mvdbeek: Yes, but you also need to set the prop. In client/src/components/WorkflowInvocationState/JobStep.vue  you don't need to create a new prop, you only need to pass `:showWorkflowInfo="false"` in <job-information/>

bgruening:
We just discussed in a paper-cuts call the stderr/stdout display. I think it would be nice if we have a more scalable way to display files in this view. Not only stderr/stdout but also inputs and outputs.

From my bug-hunting perspective, when I try to answer user-report-mails I usually see the job-info page. And most of the time I need to see stderr and the input sites. It would be nice if we could open a modal (or the scratchbook?) to show those datasets. This would make the debugging way easier.

This should not go into this PR, but if you like the idea @mvdbeek Oleg could work on it he said.

mvdbeek:
> We just discussed in a paper-cuts call the stderr/stdout display. I think it would be nice if we have a more scalable way to display files in this view. Not only stderr/stdout but also inputs and outputs.

Can we discuss that in the next call ? I joined late ... have you seen how this looks like in 21.01 ?

client/src/components/JobInformation/JobInformation.vue
simonbray: Got it, thanks.

client/src/components/DatasetInformation/DatasetInformation.vue
```diff
@@ -1,11 +1,11 @@
 <template>
     <DatasetProvider :id="hda_id" v-slot="{ item: dataset, loading }">
         <div v-if="!loading">
-            <h3>Dataset Information</h3>
+            <h3>Dataset Information (<a :href="'/datasets/edit?dataset_id=' + dataset.id" target="_top">edit</a>)</h3>
```
mvdbeek: Can you add a edit button here ?

client/src/components/DatasetInformation/DatasetInformation.vue
```diff
@@ -1,11 +1,11 @@
...
             <table id="dataset-details" class="tabletip info_data_table">
                 <tbody>
                     <tr>
-                        <td>Number</td>
+                        <td>Number in history</td>
```
mvdbeek: This is nice!
```suggestion
                        <td>Number in History</td>
```
Looks slightly better ?

client/src/components/JobInformation/JobInformation.vue
```diff
@@ -6,7 +6,7 @@
                 <tr v-if="job && job.tool_id">
                     <td>Galaxy Tool ID:</td>
                     <td id="galaxy-tool-id">
-                        {{ job.tool_id }}
+                        {{ job.tool_id }} (<a :href="'/root?tool_id=' + job.tool_id" target="_top">run</a>)
```
mvdbeek: Can you add a button here as well ?

client/src/components/JobInformation/JobInformation.vue
```diff
@@ -18,6 +18,24 @@
...
                 </tr>
+                <tr v-if="job && showWorkflowInfo">
+                    <td>Created by workflow?</td>
+                    <td id="created-by-workflow">{{ job.workflow_invocation_step != None }}</td>
+                </tr>
+                <tr v-if="job && job.workflow_invocation_step && showWorkflowInfo">
+                    <td>Workflow ID:</td>
+                    <td id="workflow-id">
+                        {{ job.workflow_invocation_step.id }} (<a
+                            :href="'/workflow/display_by_id?id=' + job.workflow_invocation_step.id"
```
mvdbeek: This is probably the workflow id, not the Stored Workflow id. Are you sure this is correct ?

client/src/store/jobStore.js
```diff
@@ -14,7 +14,7 @@ const getters = {
 
 const actions = {
     fetchJob: async ({ commit }, jobId) => {
-        const { data } = await axios.get(`${getAppRoot()}api/jobs/${jobId}?full=true`);
+        const { data } = await axios.get(`${getAppRoot()}api/jobs/${jobId}?full=true&view=workflow`);
```
mvdbeek: This is going to fetch the expensive details for every job here. You need to make the view a variable.

client/src/components/JobInformation/JobInformation.vue
```diff
@@ -47,7 +65,11 @@
                 </tr>
                 <tr v-if="job && job.id">
                     <td>Job API ID:</td>
-                    <td id="encoded-job-id">{{ job.id }} <decoded-id :id="job.id" /></td>
+                    <td id="encoded-job-id">
+                        {{ job.id }} <decoded-id :id="job.id" /> (<a :href="'/root?job_id=' + job.id" target="_top"
```
mvdbeek: Can you make the rerun a button (like it is in the history panel)

simonbray:
@mvdbeek I addressed your comments, with the exception of adding the buttons.

I understand @OlegZharkov is going to spend some time working on this page, he will do a better job of that UI stuff than me.

client/src/components/DatasetInformation/DatasetInformation.vue
OlegZharkov: addressed

OlegZharkov: addressed

client/src/components/DatasetInformation/DatasetInformation.vue
OlegZharkov: I guess, we kinda have to deal with it for now, not sure about global Vue router status

client/src/components/DatasetInformation/DatasetInformation.vue
OlegZharkov: addressed

client/src/components/JobInformation/JobInformation.vue
OlegZharkov: addressed

client/src/store/jobStore.js
OlegZharkov: addressed

client/src/components/JobInformation/JobInformation.vue
OlegZharkov: addressed

client/src/components/JobInformation/JobInformation.vue
OlegZharkov: addressed

lib/galaxy/model/__init__.py
OlegZharkov: this is also hopefully been addressed

lib/galaxy/model/__init__.py
```diff
@@ -856,10 +856,12 @@ class TaskMetricText(Base, BaseJobMetric, RepresentById):
     metric_value = Column(Unicode(JOB_METRIC_MAX_LENGTH))
     task = relationship('Task', back_populates='text_metrics')
 
+<<<<<<< HEAD
```
dannon: Looks like a bad merge here @simonbray

mvdbeek:
There's also a lot of different things in here, some more controversial than others. Could you break this up into one fix / improvement ? That'll increase the chance to get things merged

simonbray:
Just checking I understand correctly: the controversial part is exposing the workflow and invocation ID in the jobs API?

client/src/components/DatasetInformation/DatasetInformation.vue
```diff
@@ -1,11 +1,27 @@
...
     <DatasetProvider :id="hda_id" v-slot="{ item: dataset, loading }">
         <div v-if="!loading">
-            <h3>Dataset Information</h3>
+            <tr>
+                <td>
+                    <h3>Dataset Information</h3>
+                </td>
+                <td>
+                    <b-button
+                        class="edit-button"
```
mvdbeek: @simonbray there's more, this for instance probably doesn't make sense now that we display the dataset component in line, which already has the edit button

client/src/components/JobInformation/JobInformation.vue
```diff
@@ -7,6 +7,9 @@
                     <td>Galaxy Tool ID:</td>
                     <td id="galaxy-tool-id">
                         {{ job.tool_id }}
+                        <a title="run" :href="`${getAppRoot()}root?tool_id=${job.tool_id}`" target="_top">
+                            <font-awesome-icon icon="play" />
```
mvdbeek, line 11: This looks a bit odd without any spacing

mvdbeek:
> Just checking I understand correctly: the controversial part is exposing the workflow and invocation ID in the jobs API?

It's just generally a good strategy to create PRs that don't contain too many changes that don't depend on each other.

client/src/components/DatasetInformation/DatasetInformation.vue
simonbray: I removed the edit button as well as the job rerun button; both are already present in the newly embedded dataset / job components.

Issue #9981, opened 2020-07-10
OIDC case sensitivy


selten:
When users log in initially through OIDC, if their accounts on galaxy have email `firstname.lastname@test.com` and the OIDC account is `Firstname.Lastname@test.com`, clicking on the OIDC button will actually create a new account. I believe this should be made case-insensitive so the accounts are linked successfully without having to go through the 'user' menu.

nsoranzo:
I fixed the same issue for LDAP in https://github.com/galaxyproject/galaxy/pull/8631 .

selten:
Should be solved with #11251

PR #11251, opened 2021-01-28
Make OIDC email matching case-insensitive

selten:
OIDC email matching is currently case-sensitive, in order to not create duplicate accounts when the internal Galaxy database and the OIDC provider do not agree on the case-sensitivity of emails, they should be validated without case-sensitivity. See https://github.com/galaxyproject/galaxy/issues/9981.

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -6136,7 +6136,7 @@ class UserAuthnzToken(UserMixin, RepresentById):
 
     @classmethod
     def get_users_by_email(cls, email):
-        return cls.user_query().filter_by(email=email)
+        return cls.user_query().filter(func.lower(User.table.c.email) == email.lower())
 
     @classmethod
     def get_social_auth(cls, provider, uid):


nsoranzo (approved):

lib/galaxy/model/__init__.py
```diff
@@ -6136,7 +6136,7 @@ def get_user(cls, pk):
 
     @classmethod
     def get_users_by_email(cls, email):
-        return cls.user_query().filter_by(email=email)
+        return cls.user_query().filter(func.lower(User.table.c.email) == email.lower())
```
dannon, line 6139: It would make me more comfortable if this raised an exception here if multiple results were returned.  (I hope/assume callers currently do this validation?)

selten: This happens here: https://github.com/python-social-auth/social-core/blob/6bbad65832f311108f222aa5d6d36ad0bd8e8bc5/social_core/pipeline/social_auth.py#L78.

dannon: Ok, great.  It looks like the only other place this is used is https://github.com/galaxyproject/galaxy/blob/5e8264f9ff4b388bbd2ab990bd4074b638261d32/lib/galaxy/model/__init__.py#L6126, and it's also only to count > 0, so this is fine.  I was worried we were actually using the response from this in auth somewhere that N == 1 was expected.

selten: I definitely think it's a good idea to validate these concerns, could've caused a ton of issues down the line otherwise if people already have duplicate accounts.

Issue #11252, opened 2021-01-28
Migrate legacy_expose_api* decorators to expose_api* decorators


jdavcs:
We need to migrate from `legacy_*` decorators to their newer counterparts:
`legacy_expose_api` -> `expose_api` 
`legacy_expose_api_anonymous` -> `expose_api_anonymous`
`legacy_expose_api_raw` -> `expose_api_raw`
`legacy_expose_api_raw_anonymous` -> `expose_api_raw_anonymous`

The primary difference between the legacy and the new decorators is in error handling (see `expose_api` vs. `legacy_expose_api`). In theory, replacing one with the other should just work, as far as I can tell. Once all instances of a legacy decorator are removed, that decorator can be removed from `lib/galaxy/web/framework/decorators.py` and `lib/galaxy/web/__init__.py`.

There's also `legacy_tool_shed_endpoint` defined and used exclusively in `lib/galaxy/webapps/galaxy/controllers/admin_toolshed.py`

When you replace a legacy decorator, make sure there is a test to verify the endpoint works as expected. If there's no test, you'll need to write one (see [documentation on writing tests](https://docs.galaxyproject.org/en/latest/dev/writing_tests.html)).

Here's the list of controllers that need migrating:

- [ ] /galaxy/webapps/base/controller.py
  - [ ] legacy_expose_api 
- [ ] /galaxy/webapps/galaxy/api/datasets.py
  - [ ] legacy_expose_api_anonymous
  - [ ] legacy_expose_api_raw_anonymous
- [ ] /galaxy/webapps/galaxy/api/display_applications.py
  - [ ] legacy_expose_api
- [ ] /galaxy/webapps/galaxy/api/extended_metadata.py
  - [ ] legacy_expose_api 
- [ ] /galaxy/webapps/galaxy/api/forms.py
  - [ ] legacy_expose_api 
- [ ] /galaxy/webapps/galaxy/api/group_roles.py
  - [ ] legacy_expose_api 
- [ ] /galaxy/webapps/galaxy/api/group_users.py
  - [ ] legacy_expose_api 
- [ ] /galaxy/webapps/galaxy/api/groups.py
  - [ ] legacy_expose_api 
- [ ] /galaxy/webapps/galaxy/api/provenance.py
  - [ ] legacy_expose_api 
- [ ] /galaxy/webapps/galaxy/api/quotas.py
  - [ ] legacy_expose_api 
- [ ] /galaxy/webapps/galaxy/api/search.py
[...]

PR #11253, opened 2021-01-28
Get rid of legacy decorator in api/tours

jdavcs:
Ref: #11252 

Is this as simple as this?


jmchilton:
Looks good to me, it is that simple in this case - and probably in most cases. In some cases we should check how the client is handling these errors and make sure they are correct.

ic4f:
Thanks! I'll check the client each time.

PR #11255, opened 2021-01-29
Try out from_orm with nested models and id de/encoding

mvdbeek:


lib/galaxy/model/__init__.py
```diff
@@ -74,6 +75,7 @@
 log = logging.getLogger(__name__)
 
 _datatypes_registry = None
+_IdEncodingHelper: Optional[IdEncodingHelper] = None
```
jmchilton, line 78: I get that I'm going to lose on these globals but I hate them. Can you at least move this into schema/fields? That is where you use it.

mvdbeek: I'm still in favor of https://github.com/galaxyproject/galaxy/pull/11180, if that's what you mean. And sure.

jmchilton: This has nothing to do with #11180 - I'm just trying to figure out how to cleanly avoid globals in all these contexts. That datatypes registry for instance drives me crazy, I would love to find a way to avoid it (and avoid `Dataset.object_store setting` for that matter). I have a feeling pydantic is kind of young and going to encourage bad habits.

Overall though - this is wonderful, I'm not trying to be a downer. This is really cool stuff.

jmchilton: That isn't just me being polite - proper annotation of EncodedDatabaseIdField vs DatabaseIdField is huge and worth all the yuckiness of the globals. What we're doing in managers and not tracking these things systematically is bad and this could solve all of that nicely and provide a workable best practice I think.

mvdbeek: Okay, I need to play with #11180, but I thought we could create a singleton for an IdEncodingHelper instance and avoid the global here.

mvdbeek:
I've got an alternate version of this in one of my branches, opening that form there (if I can find it ...)

PR #11256, opened 2021-01-29
Copy Pulsar params so we don't modify the actual param dict in place

natefoo:
just a fix for #11254


jmchilton (approved):

natefoo:
Targeted the wrong branch... sigh.

Issue #10914, opened 2020-12-14
[20.09] Incorrect tool version saved in workflow


innovate-invent:
Open a workflow, click a tool, "Switch to" a different tool version, save, download. The downloaded JSON does not refer to the correct tool version despite the workflow referring to the correct tool version when run.

Looking at the json it does appear to have the correct changeset revision but the wrong tool id:
```
      "tool_id": "toolshed.g2.bx.psu.edu/repos/brinkmanlab/awkscript/awkscript/1.0",
      "tool_shed_repository": {
        "changeset_revision": "df0c8f33c8ea",
        "name": "awkscript",
        "owner": "brinkmanlab",
        "tool_shed": "toolshed.g2.bx.psu.edu"
      },
```

Should read `"tool_id": "toolshed.g2.bx.psu.edu/repos/brinkmanlab/awkscript/awkscript/1.1"`

wm75:
I just had to update a large, but rather old workflow to use all the latest tool versions, and I can confirm and add to @innovate-invent's observation.
Out of the four locations that mention specifics about the tool version, only

- `tool_shed_repository -> changeset_revision` and
- `tool_version`

do get updated when you're updating a tool version through the WF editor, while the paths in

- `content_id` and
- `tool_id`

are left untouched.

As the author of the WF, there is no way for you to know except by looking through the .ga file, but when you share such an out-of-sync WF with others, they seem to get switched back to the old tool version.

So this turns out to be a rather serious issue in terms of reproducibility :(

mvdbeek:
Should be fixed in https://github.com/galaxyproject/galaxy/pull/11257, thanks for the report @innovate-invent @wm75 !

PR #11257, opened 2021-01-29
[20.09] Fix tool version switch in workflow editor

mvdbeek:
Fixes https://github.com/galaxyproject/galaxy/issues/10914


martenson:
@hexylena I found it too (https://help.galaxyproject.org/t/server-error-when-opening-certain-shared-workflows/5571)

mvdbeek:
Yes, that's right.

PR #11262, opened 2021-01-29
FastAPI routes for Pages (Disabled)

davelopez:
Ref #10889

## Summary
- Legacy controller logic refactored into `PagesManager`.
- Added some more tests for edge cases.

## Additional comments
- Once more I placed the pydantic models in the manager module because I wasn't sure where to place the equivalent `_schema.py`.
- The empty JSON response in the `delete` endpoint now returns a `204`.
- The legacy controller was declared with some mixins (``SharableItemSecurityMixin``, ``UsesAnnotations``, ``SharableMixin``) but, apparently, they were not used so I removed them (the tests are passing), the same for the `PageSerializer`. Let me know if that was a bad idea 😅 
- When trying to download a PDF version of the page, I've added a method to raise an exception (early in the request) if `weasyprint` is not installed. I wonder if with some integration tests we can actually test if this works when the package is present.
- Forced the tests to send the payload as JSON to make them compatible with FastAPI.
- Most of the tests for the FastAPI endpoints are failing because of the issues with the database session 😓

### DIFF
--- a/lib/galaxy/managers/markdown_util.py
+++ b/lib/galaxy/managers/markdown_util.py
@@ -27,7 +27,11 @@ try:
 except Exception:
     weasyprint = None
 
-from galaxy.exceptions import MalformedContents, MalformedId
+from galaxy.exceptions import (
+    MalformedContents,
+    MalformedId,
+    MessageException,
+)
 from galaxy.managers.hdcas import HDCASerializer
 from galaxy.managers.jobs import (
     JobManager,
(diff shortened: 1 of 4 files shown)


jmchilton:
Maybe rebase this and tags PR without the FastAPI route - and have a branch with the new routes that don't yet work because of session issues?

davelopez:
The FastAPI routes have been temporarily removed and will be added in a separate PR when the database session issues are fixed.

jmchilton:
Awesome, thanks!

guerler:
The pages / markdown config contains optional attributes like e.g. history_datasets, history_dataset_collections, workflows, jobs and invocations which are not parsed anymore to the client. They need to be added as optional attributes to the schema.

guerler:
Yes, no worries. So the attributes are expected at: https://github.com/galaxyproject/galaxy/blob/dev/client/src/components/Markdown/Markdown.vue#L193. In order to test this, create a page and insert a workflow and attempt to view the result.

PR #11265, opened 2021-01-31
Workflow editor null/zero fixes

mvdbeek:
Both are type casting issues that I think can be avoided if we make our ToolParameters pydantic models.

### DIFF
--- a/lib/galaxy/tools/parameters/__init__.py
+++ b/lib/galaxy/tools/parameters/__init__.py
@@ -206,7 +206,7 @@ def params_to_strings(params, param_values, app, nested=False, use_security=Fals
     for key, value in param_values.items():
         if key in params:
             value = params[key].value_to_basic(value, app, use_security=use_security)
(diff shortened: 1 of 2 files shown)


dannon:
Is this the same root issue as https://github.com/galaxyproject/galaxy/pull/9484?

If so, I'm glad for the workaround leveraging typing on the ToolParameter side, but as a heads up there are likely still other issues related to the nested encoded json

mvdbeek:
Yes, I put that in the commit description. I did a fairly deep dive into how we build the tool form parameter dict on the client side.

> heads up there are likely still other issues related to the nested encoded json.

Yeah, for sure, it isn't even "native" json, it's just that the client casts all values to strings, which isn't optimal by a long shot. I really hope a more reasonable serialized format will fall out of vue-ifying the tool form parameters without much/any effort.

dannon (approved):

Issue #11270, opened 2021-02-01
Jobs fail due to underscore in username


code4dna:
We have 20.09 installed on a cluster. There are 60 students in our class and all were added to Galaxy on the same day. There are two who can login but their jobs fail. The only thing I can see is that these two students are the only ones that have underscores in their user names. Not sure if this is failing due to Galaxy or Slurm or both.

```
galaxy.jobs.runners.drmaa DEBUG 2021-02-01 10:39:18,065 [p:39975,w:1,m:0] [SlurmRunner.work_thread-8] (536) submitting file /scratch/user/galaxy/kaiser/jobs_directory/000/536/galaxy_536.sh
galaxy.jobs.runners.drmaa DEBUG 2021-02-01 10:39:18,065 [p:39975,w:1,m:0] [SlurmRunner.work_thread-8] (536) native specification is: --time=24:00:00 --nodes=1 --ntasks-per-node=1 --cpus-per-task=1 --mem=1200 --partition=knl
galaxy.jobs.runners ERROR 2021-02-01 10:39:18,069 [p:39975,w:1,m:0] [SlurmRunner.work_thread-8] (536) Unhandled exception calling queue_job
Traceback (most recent call last):
  File "lib/galaxy/jobs/runners/__init__.py", line 137, in run_next
    method(arg)
  File "lib/galaxy/jobs/runners/drmaa.py", line 204, in queue_job
    job_wrapper.change_ownership_for_run()
  File "lib/galaxy/jobs/__init__.py", line 2220, in change_ownership_for_run
    external_chown_script, description="working directory")
  File "lib/galaxy/util/path/__init__.py", line 360, in external_chown
    cmd.extend([path, pwent[0], str(pwent[3])])
TypeError: 'NoneType' object is not subscriptable
galaxy.jobs.runners.drmaa ERROR 2021-02-01 10:39:18,070 [p:39975,w:1,m:0] [SlurmRunner.work_thread-8] (536/None) User killed running job, but error encountered removing from DRM queue
Traceback (most recent call last):
  File "lib/galaxy/jobs/runners/drmaa.py", line 356, in stop_job
    assert ext_id not in (None, 'None'), 'External job id is None'
AssertionError: External job id is None
```

nsoranzo:
I guess you are [submitting job as the real user](https://docs.galaxyproject.org/en/master/admin/cluster.html#submitting-jobs-as-the-real-user) ?

code4dna:
[...]

PR #11272, opened 2021-02-01
Don't fail silently in `User.system_user_pwent()`

nsoranzo:
When the exception is not re-raised, `external_chown()` (in `lib/galaxy/util/path/__init__.py`) fails with:
```
  File "lib/galaxy/util/path/__init__.py", line 360, in external_chown
    cmd.extend([path, pwent[0], str(pwent[3])])
TypeError: 'NoneType' object is not subscriptable
```

xref https://github.com/galaxyproject/galaxy/issues/11270

### DIFF
--- a/lib/galaxy/tools/actions/upload_common.py
+++ b/lib/galaxy/tools/actions/upload_common.py
@@ -377,7 +377,7 @@ def create_paramfile(trans, uploaded_datasets):
             # TODO: This will have to change when we start bundling inputs.
             # Also, in_place above causes the file to be left behind since the
             # user cannot remove it unless the parent directory is writable.
-            if link_data_only == 'copy_files' and trans.user:
+            if link_data_only == 'copy_files' and trans.user and trans.app.config.external_chown_script:
                 external_chown(uploaded_dataset.path,
                                trans.user.system_user_pwent(trans.app.config.real_system_username),
                                trans.app.config.external_chown_script, description="uploaded file")
--- a/lib/galaxy/tools/imp_exp/__init__.py
+++ b/lib/galaxy/tools/imp_exp/__init__.py
@@ -23,11 +23,20 @@ class JobImportHistoryArchiveWrapper:
         self.sa_session = self.app.model.context
 
     def setup_job(self, jiha, archive_source, archive_type):
-        if archive_type != "url":
-            external_chown(archive_source, jiha.job.user.system_user_pwent(self.app.config.real_system_username),
-                           self.app.config.external_chown_script, "history import archive")
-        external_chown(jiha.archive_dir, jiha.job.user.system_user_pwent(self.app.config.real_system_username),
-                       self.app.config.external_chown_script, "history import archive directory")
+        if self.app.config.external_chown_script:
+            if archive_type != "url":
+                external_chown(
+                    archive_source,
+                    jiha.job.user.system_user_pwent(self.app.config.real_system_username),
+                    self.app.config.external_chown_script,
+                    "history import archive"
+                )
+            external_chown(
+                jiha.archive_dir,
+                jiha.job.user.system_user_pwent(self.app.config.real_system_username),
+                self.app.config.external_chown_script,
+                "history import archive directory"
+            )
 
     def cleanup_after_job(self):
         """ Set history, datasets, collections and jobs' attributes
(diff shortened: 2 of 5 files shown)


lib/galaxy/tools/actions/upload_common.py
```diff
@@ -377,7 +377,7 @@ def create_paramfile(trans, uploaded_datasets):
             # TODO: This will have to change when we start bundling inputs.
             # Also, in_place above causes the file to be left behind since the
             # user cannot remove it unless the parent directory is writable.
-            if link_data_only == 'copy_files' and trans.user:
+            if link_data_only == 'copy_files' and trans.user and trans.app.config.external_chown_script:
```
bernt-matthias, line 380:
```suggestion
            if link_data_only == 'copy_files' and trans.user and trans.app.config.external_chown_script is not None:
```

lib/galaxy/tools/imp_exp/__init__.py
```diff
@@ -23,11 +23,20 @@ def __init__(self, app, job_id):
         self.sa_session = self.app.model.context
 
     def setup_job(self, jiha, archive_source, archive_type):
-        if archive_type != "url":
-            external_chown(archive_source, jiha.job.user.system_user_pwent(self.app.config.real_system_username),
-                           self.app.config.external_chown_script, "history import archive")
-        external_chown(jiha.archive_dir, jiha.job.user.system_user_pwent(self.app.config.real_system_username),
-                       self.app.config.external_chown_script, "history import archive directory")
+        if self.app.config.external_chown_script:
```
bernt-matthias, line 26:
```suggestion
        if self.app.config.external_chown_script is not None:
```

lib/galaxy/tools/imp_exp/__init__.py
```diff
@@ -46,8 +55,13 @@ def cleanup_after_job(self):
         new_history = None
         try:
             archive_dir = jiha.archive_dir
-            external_chown(archive_dir, jiha.job.user.system_user_pwent(getpass.getuser()),
-                           self.app.config.external_chown_script, "history import archive directory")
+            if self.app.config.external_chown_script:
```
bernt-matthias, line 58:
```suggestion
            if self.app.config.external_chown_script is not None:
```

lib/galaxy/util/path/__init__.py
```diff
@@ -352,7 +352,7 @@ def external_chown(path, pwent, external_chown_script, description="file"):
     - True in case of success
     - False in case of failure
     """
-    if external_chown_script is None:
+    if not external_chown_script:
```
bernt-matthias: also check for None here?

bernt-matthias (commented):
The main change is in `lib/galaxy/model/__init__.py`, or? At which point(s) would this lead to an Exception? User creation?

Also I was thinking if we should do the check if `external_chown_script` is / is not None in the `external_chown` function or a new `external_chown_if_configured` function?

lib/galaxy/tools/actions/upload_common.py
nsoranzo: Why do you prefer checking for `None`? Is there a different meaning in an empty string?

bernt-matthias: Ah, I see - my mistake. Actually you make this more consistent. 

Question is if we want to / should assume that this comes as `None` from the configuration or empty string .. or if we don't care. 

Anyway I'm fine with the change.

nsoranzo:
> The main change is in `lib/galaxy/model/__init__.py`, or? At which point(s) would this lead to an Exception? User creation?

I've updated the PR description, I hope that clarifies the reason for the changes.

> Also I was thinking if we should do the check if `external_chown_script` is / is not None in the `external_chown` function or a new `external_chown_if_configured` function?

I've left the check:
```
    if not external_chown_script:
        return None
```
in the `external_chown` function because I was not sure if it was used somewhere else, but I am happy to drop it from there now that this is checked before calling it, if you like.

bernt-matthias:
> I've left the check:
> 
> ```
>     if not external_chown_script:
>         return None
> ```
> 
> in the `external_chown` function because I was not sure if it was used somewhere else, but I am happy to drop it from there now that this is checked before calling it, if you like.

I think having the check once should be sufficient. I don't care much if it's done inside the function or outside .. tend a bit to inside because simplifies the code a wee bit (but I guess having the check outside of the function may be more explicit which is good for readability).

nsoranzo:
> I think having the check once should be sufficient. I don't care much if it's done inside the function or outside .. tend a bit to inside because simplifies the code a wee bit (but I guess having the check outside of the function may be more explicit which is good for readability).

I've added it outside the call so we don't have to call `system_user_pwent()` when it's not necessary, since that can fail e.g. in a testing setup.

bernt-matthias (approved):
Looks good to me

PR #11273, opened 2021-02-01
Beta History content list item components

Nerdinacan:
More Beta Fixes:

Simplified component hierarchy on history list and collection list. The old system worked but was convoluted. I believe this new arrangement is easier to both understand and customize. In the process some common functions for editable elements have been consolidated.

**Motivation**
Most of what people think of as missing/broken history functionality can be distilled down into buttons or features that are or aren't present in one of the ContentItem components: either for a Dataset, a Collection or the collection-nested variants. I would like the intended usage of these components to be obvious.

**Issues Addressed**
Addresses flaws noted regarding the disparity between datasets that are direct history children and those that are more deeply nested inside collections. Notably collections and their contents are intended to be read-only.


Nerdinacan:
All done, was waiting for the ability to document this.

astrovsky01:
Some thoughts:
* If the filters are at default (hidden+deleted datasets invisible), list datasets as "x datasets in history" or something other than "Total Matches".
* The checkbox to activate multi-dataset select suggests to me it's a "select all" button rather than something to open the select dataset menu.
* For the execute drop down, add space above and a label for something like "History functions".
* The collapse all button remains greyed out until next click anywhere on screen, no matter mouse position.
* Add filter for error datasets or a hide/delete errored dataset function.
* I love the expanding menu of operations, but reorder the buttons? Might just be me, but I would think I would use 'extract history' and 'resume paused jobs' more often than duplicate or share, and based on how wide a user would generally keep their history panel, it makes sense to have the most-used buttons up front.
* Filter buttons (hidden/deleted) only show &&. I see no way to use an or. I can use a search, but the popup for filter keywords on the current history filter bar is gone.
* Select history button doesn't need ellipsis. It looks like something got cut off.

client/src/components/History/ContentItem/HistoryContentItem.vue
```diff
@@ -0,0 +1,106 @@
...
+        :is="contentItemComponent"
+        class="content-item p-1"
+        v-on="$listeners"
+        v-bind="$attrs"
+        :index="index"
+        :tabindex="index"
+        :item="localItem"
+        @update:item="onUpdate"
+        @delete="onDelete"
+        @undelete="onUndelete"
```
mvdbeek, line 14: This seems to not be defined now?
Screenshot 2021-04-08 at 12 07 05

If I remove both then I'm getting
Screenshot 2021-04-08 at 12 31 21

I guess this still needs some tweaking ?
Overall this looks like a nice improvement.

Nerdinacan: That's weird, I'm looking at my code right now and onUndelete is clearly defined in that component. I'll look at it closer after diner.

PR #11274, opened 2021-02-01
Generalise rule helper to detect whether tools can use any container

pcm32:
This PR adds the ability for the rule_helper to asses whether a tool can use not only docker containers but other types, such as singularity. This is towards dynamic destinations that could decide, like docker dispatch, whether you can dispatch a tool to a singularity destination.

### DIFF
--- a/lib/galaxy/jobs/rule_helper.py
+++ b/lib/galaxy/jobs/rule_helper.py
@@ -34,7 +34,7 @@ class RuleHelper:
 
         :param job_or_tool:
         :param container_type: either "docker" or "singularity" currently
-        :return:
+        :return: true if the tool supports the specified container type.
         """
         # Not a ton of logic in this method - but the idea is to shield rule
         # developers from the details and they shouldn't have to know how to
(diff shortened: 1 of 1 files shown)


pcm32:
I have tested this to work now... will check the linting issue later today.

lib/galaxy/jobs/rule_helper.py
```diff
@@ -43,10 +48,15 @@ def supports_docker(self, job_or_tool):
...
-        container_description = self.app.container_finder.find_best_container_description(["docker"], tool_info)
+        tool_info = ToolInfo(tool.containers, tool.requirements, tool.requires_galaxy_python_environment,
+                             tool.docker_env_pass_through)
+        container_description = self.app.container_finder.find_best_container_description([container_type], tool_info)
         return container_description is not None
 
+    def supports_docker(self, job_or_tool):
+        return self.supports_container(job_or_tool, container_type="docker")
+
+
```
mvdbeek, line 73:
```suggestion
```

pcm32:
I left the convenience call `supports_docker` with the same signature since this currently can be called from admin's written dynamic destinations in python, and so removing this could break installations.

pcm32:
I'm happy to add the dynamic destination I wrote for singularity_dispatcher based on this to the stock rules if you prefer, but didn't want to oblige.

Issue #8488, opened 2019-08-21
Rule based uploader fails all datasets if one url has an issue.


hexylena:
I ran the uploader with 6 URLs, one I did not realise was invalid.

All datasets are in error state, but unlike in https://github.com/galaxyproject/galaxy/issues/4021#issuecomment-523518510, they all bear the same error message, so there's no clue which one failed.

```
Traceback (most recent call last):
  File "/opt/galaxy/server/lib/galaxy/tools/data_fetch.py", line 329, in <module>
    main()
  File "/opt/galaxy/server/lib/galaxy/tools/data_fetch.py", line 39, in main
    galaxy_json = _request_to_galaxy_json(upload_config, request)
  File "/opt/galaxy/server/lib/galaxy/tools/data_fetch.py", line 49, in _request_to_galaxy_json
    fetched_target = _fetch_target(upload_config, target)
  File "/opt/galaxy/server/lib/galaxy/tools/data_fetch.py", line 167, in _fetch_target
    elements = elements_tree_map(_resolve_src, items)
  File "/opt/galaxy/server/lib/galaxy/tools/data_fetch.py", line 195, in elements_tree_map
    new_items.append(f(item))
  File "/opt/galaxy/server/lib/galaxy/tools/data_fetch.py", line 96, in _resolve_src
    name, path = _has_src_to_path(upload_config, item, is_dataset=True)
  File "/opt/galaxy/server/lib/galaxy/tools/data_fetch.py", line 223, in _has_src_to_path
    path = sniff.stream_url_to_file(url)
  File "/opt/galaxy/server/lib/galaxy/datatypes/sniff.py", line 57, in stream_url_to_file
    page = urlopen(path)  # page will be .close()ed in stream_to_file
  File "/usr/lib64/python2.7/urllib2.py", line 154, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib64/python2.7/urllib2.py", line 437, in open
    response = meth(req, response)
  File "/usr/lib64/python2.7/urllib2.py", line 550, in http_response
    'http', request, response, code, msg, hdrs)
  File "/usr/lib64/python2.7/urllib2.py", line 475, in error
    return self._call_chain(*args)
  File "/usr/lib64/python2.7/urllib2.py", line 409, in _call_chain
    result = func(*args)
  File "/usr/lib64/python2.7/urllib2.py", line 558, in http_error_default
    raise HTTPError(req.get_full_url(), code, msg, hdrs, fp)
[...]

PR #11276, opened 2021-02-02
Bugfix/8488/rule builder fails for one bad url

assuntad23:
URLs uploaded now use Data Fetch. 
Fix for #8488

To test: 
1. Try to upload these URLs using the Default or the Collection upload modal:
https://zenodo.org/api/files/55133323-b15b-45b4-98c9-dda00288b53f/nurse.bw
https://zenodo.org/api/files/55133323-b15b-45b4-98c9-dda00288b53f/nurse_Amel4.5_accepted_hits.bam

2.  On main/usegalaxy, they will both fail, because one URL is bad. On this branch, only the bad branch will fail. 
3. On Composite, the job still uses the upload1 tool, because if a URL fails, the dataset should be in an error state.


client/src/components/Upload/UploadModal.vue
```diff
@@ -331,6 +361,35 @@ const UploadModal = {
...
+            };
+
+            const urls = items[0].get("url_paste").split("\n");
+            for (var index in urls) {
+                var url = urls[index].trim();
+                if (url != "") {
+                    var element = {
+                        url: urls[index].trim(),
+                        src: "url",
+                        dbkey: "?",
```
jmchilton: dbkey and ext should be coming from the form in some way I think, these make sense as defaults but if they're overridden that value should be used I think.

jmchilton:
```
                        inputs[`${prefix}space_to_tab`] = (it.get("space_to_tab") && "Yes") || null;
                        inputs[`${prefix}to_posix_lines`] = (it.get("to_posix_lines") && "Yes") || null;
                        inputs[`${prefix}dbkey`] = it.get("genome", null);
                        inputs[`${prefix}file_type`] = it.get("extension", null);
```

These options are all available in the data fetch API, they are just a bit cleaner I think (e.g. not requiring the string "Yes").

client/src/components/Upload/Collection.vue
```diff
@@ -169,12 +170,20 @@ export default {
         this.initFtpPopover();
         // file upload
         this.initUploadbox({
-            url: this.app.uploadPath,
+            initUrl: (index) => {
```
jmchilton, line 173: Can you put this into the mixin these components shared, because this code is identical?

client/src/components/Upload/UploadModal.vue
```diff
@@ -262,12 +262,42 @@ const UploadModal = {
...
+            return data;
+        },
+        preprocess: function (items) {
+            var data = {
+                urls: [],
+                files: [],
+            };
+            for (var index in items) {
+                var it = items[index];
+                if (it.get("file_mode") != "new" || !it.get("url_paste").startsWith("http")) {
```
jmchilton: Can we synchronize this with the backend - we do a bit more aggressive checks for URIs:

https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/parameters/grouping.py#L359
https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/parameters/grouping.py#L25

client/src/components/Upload/UploadModalContent.vue
```diff
@@ -167,12 +168,50 @@ export default {
...
+                if (it.get("file_mode") != "new" || !this.itemIsURL(it)) {
+                    data.files.push(it);
+                } else {
+                    data.urls.push(it);
+                }
+            }
+            return data;
+        },
+        itemIsURL: function (item) {
+            for (var index in this.URI_PREFIXES) {
```
jmchilton: I'm pretty sure we don't want to use var here and we could simplify the loop by using ``of`` instead of ``in``.

```javascript
for(const prefix of this.URI_PREFIXES) {
```

client/src/components/Upload/UploadModalContent.vue
```diff
@@ -167,12 +168,50 @@ export default {
...
+                uploadRequest: null,
+            };
+            if (items && items.length > 0) {
+                var split = this.preprocess(items);
+                if (split.urls.length > 0) {
+                    data.fetchRequest = this.toFetchData(split.urls, history_id);
+                }
+                if (split.files.length > 0) {
+                    data.uploadRequest = this.toFileUploadData(split.files, history_id);
+                }
```
jmchilton, line 182: This looks like both data.fetchRequest and data.uploadRequest could be set right? But then other places files would be ignored if fetchRequest is set.

client/src/components/Upload/UploadModalContent.vue
mvdbeek: I like this over explicit loops as an alternative.
```javascript
return this.URI_PREFIXES.some((prefix) => item.get('url_paste').startsWith(prefix))
```

client/src/components/Upload/UploadModalContent.vue
```diff
@@ -167,12 +168,50 @@ export default {
         currentFtp: function () {
             return this.currentUserId && this.ftpUploadSite;
         },
-
+        toData: function (items, history_id) {
+            var data = {
```
mvdbeek: Shouldn't this be const ? I think generally speaking we'll want to use `const` if there is no reassignment or update and `let` if there is and avoid `var` as far as possible. In this case you're updating properties of data, which is fine with `const` as well

client/src/components/Upload/UploadModalContent.vue
assuntad23: @jmchilton whoops! I wrote this when I was still thinking I would need to use this for Composites, but since Composites should fail if one piece of data is no good, I'll convert this to a simple if/else.

client/src/components/Upload/UploadBoxMixin.js
```diff
@@ -332,5 +332,9 @@ export default {
             const models = allHids.map((hid) => Galaxy.currHistoryPanel.collection.getByHid(hid));
             return models;
         },
+        getRequestUrl: function (items, history_id) {
+            var data = this.app.toData(items, history_id);
+            return data.fetchRequest ? "/api/tools/fetch" : this.app.uploadPath;
```
jmchilton: This seems like it needs to have an appRoot prefix it - e.g. https://github.com/galaxyproject/galaxy/blob/dev/client/src/components/Upload/config.js#L9.

client/src/components/Upload/UploadModalContent.vue
```diff
@@ -238,6 +271,38 @@ export default {
...
+                    },
+                ],
+                auto_decompress: true,
+            };
+
+            // Composite does not use the fetch API, so we can just
+            // index into the first element of items
+            const urls = items[0].get("url_paste").split("\n");
+            const dbkey = items[0].get("genome", "?");
+            const ext = items[0].get("extension", "auto");
```
jmchilton: I assume these need to get added also:

```
                        inputs[`${prefix}space_to_tab`] = (it.get("space_to_tab") && "Yes") || null;
                        inputs[`${prefix}to_posix_lines`] = (it.get("to_posix_lines") && "Yes") || null;
```

They are used in data_fetch.py and they are probably just treated as booleans (instead of "Yes" or "" strings).

```
            to_posix_lines = upload_config.get_option(item, "to_posix_lines")
            space_to_tab = upload_config.get_option(item, "space_to_tab")
```

The defaults might be different also:

```
        self.to_posix_lines = request.get("to_posix_lines", False)
        self.space_to_tab = request.get("space_to_tab", False)
```

Update: If these advanced options are only shown for traditional uploads - you can ignore this comment.

jmchilton:
We discussed this in person - I was hoping you'd test uploading two sets of URLs with different options and make sure the options work and are distinctly handled in both cases. Thanks!

jmchilton:
I'll admit the code confuses me but it seems to work. I'm going to merge this - thanks so much @assuntad23!

PR #11277, opened 2021-02-02
[21.01] Fix error on Admin panel when creating new quota

davelopez:
When trying to create a new quota from the admin panel the following error was displayed: `Failed to load resource admin/create_quota.` and the form was not shown.

### DIFF
--- a/lib/galaxy/webapps/galaxy/controllers/admin.py
+++ b/lib/galaxy/webapps/galaxy/controllers/admin.py
@@ -677,8 +677,8 @@ class AdminGalaxy(controller.JSAppLauncher, AdminActions, UsesQuotaMixin, QuotaP
                                          .order_by(trans.app.model.Group.table.c.name):
                 all_groups.append((group.name, trans.security.encode_id(group.id)))
             default_options = [('No', 'no')]
-            for typ in trans.app.model.DefaultQuotaAssociation.types.__dict__.values():
-                default_options.append(('Yes, ' + typ, typ))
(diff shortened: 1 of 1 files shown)


mvdbeek (approved):

jdavcs (approved):
Definitely! 
We are also checking *all* the class attributes here: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/model/__init__.py#L2301 , for no good reason. Maybe correct that too?

jmchilton:
Sorry for breaking this, looks good to me! Thanks.

davelopez:
@ic4f 

>We are also checking all the class attributes here: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/model/__init__.py#L2301 , for no good reason. Maybe correct that too?

Should I change that too in this fix or make a separate PR?

ic4f:
> @ic4f
> 
> > We are also checking all the class attributes here: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/model/__init__.py#L2301 , for no good reason. Maybe correct that too?
> 
> Should I change that too in this fix or make a separate PR?

No difference, really. It belongs in this one (same type of fix), but the tests are already green, so the simplest/fastest thing to do is to merge this and just reference it in a separate PR. (in my personal experience, these "this too needs to be fixed" can turn a simple PR into days of lingering discussions :-)

ic4f:
And, @davelopez, thanks for finding and fixing this!

PR #11278, opened 2021-02-02
Fix printing non-dict output in api/display script

jdavcs:
I know it's an ancient script, but it's still useful for dev work testing api endpoints. The edit is to accommodate the `api/genomes` endpoint.


dannon:
This is fine, but I think we should probably clear these out of the repo at some point -- people stumble into trying old stuff that doesn't work or isn't recommended more often than they are useful, these days.

ic4f:
> This is fine, but I think we should probably clear these out of the repo at some point -- people stumble into trying old stuff that doesn't work or isn't recommended more often than they are useful, these days.

Thanks for merging! And sure, I don't mind if we remove this script (I'm only using it for simple GET requests).

PR #11281, opened 2021-02-03
[21.01] Revert "Make ``MetadataCollection`` a proper mapping"

mvdbeek:
This partially reverts commit 335a897.

The jbrowse tool iterates over metadata items in https://github.com/galaxyproject/tools-iuc/blob/master/tools/jbrowse/jbrowse.xml#L103 and the Mapping implementation behaves differently here.
I'm not sure what exactly is the difference, so there may be a better fix.

### DIFF
--- a/lib/galaxy/model/metadata.py
+++ b/lib/galaxy/model/metadata.py
@@ -11,7 +11,6 @@ import sys
 import tempfile
 import weakref
 from collections import OrderedDict
-from collections.abc import Mapping
 from os.path import abspath
 
 from sqlalchemy.orm import object_session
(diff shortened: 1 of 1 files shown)


nsoranzo:
After a day of debugging, I've found the real (stupid) issue with commit 335a89772012041538e86835a4e16f3a2e50d51f , opening an alternative PR in a sec.

mvdbeek:
Do you have time to dig further into https://github.com/galaxyproject/galaxy/pull/11293 ? This is failing jobs on main and I think this may or may not hide some other bugs. Maybe we can merge this and then merge https://github.com/galaxyproject/galaxy/pull/11293  when it's green ?

nsoranzo:
Superseded by https://github.com/galaxyproject/galaxy/pull/11293 .

PR #11288, opened 2021-02-03
[21.01] Use nested state tracking with sqlalchemy-mutable

mvdbeek:
That should fix AttributeError: 'MutationList' object has no attribute '_key' for good, hopefully.

In the process I also had to eliminate pickling of datasets, which fails with both sqlalchemy-mutable and sqlalchemy-json.
I also removed most uses of OrderedDict ... they could have been allowed as additional JSONType's, but since they don't offer anything in the current codebase (we don't seem to call `reversed` on them) I decided to just remove them.


lib/galaxy/model/custom_types.py
```diff
@@ -113,149 +113,15 @@ def compare_values(self, x, y):
...
-        def set(target, value, oldvalue, initiator):
-            if not isinstance(value, cls):
-                value = cls.coerce(key, value)
-            if isinstance(value, cls):
-                value._parents[target.obj()] = key
-            if isinstance(oldvalue, cls):
-                oldvalue._parents.pop(target.obj(), None)
-            return value
+class JSONType(BaseJSONType):
+    pass
```
nsoranzo: I'd move this after the definition of `SimpleJSONType` class, so closer to `Mutable.associate_with(JSONType)`.

Also, can you please add a class docstring?

lib/galaxy/managers/executables.py
```diff
@@ -1,6 +1,6 @@
 """Utilities for loading tools and workflows from paths for admin user requests."""
 
-from gxformat2.converter import ordered_load
+import yaml
```
jmchilton, line 3: I'm certain I had this in here for a reason. Is it just not needed on Python 3?

jmchilton (approved):

lib/galaxy/managers/executables.py
mvdbeek: Yep, on python 3.6 dicts became ordered, and since 3.7 that behavior is guaranteed

mvdbeek:
Hmm, one of the integration tests is timing out with a yaml tool, there's probably something subtly wrong here.

jmchilton:
I would suggest maybe just because ``{}`` is ordered now, doesn't mean ``yaml.safe_load`` is also?

mvdbeek:
It looks to be ordered so far, but I can dig up if that's documented.
The timeout is caused by this:
```
galaxy.jobs.runners ERROR 2021-02-05 19:04:35,130 [p:60074,w:1,m:0] [PulsarJobRunner.work_thread-1] (2) Unhandled exception calling finish_job
  File "/Users/mvandenb/src/galaxy/lib/galaxy/jobs/runners/pulsar.py", line 602, in finish_job
    client = self.get_client_from_state(job_state)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/jobs/runners/pulsar.py", line 572, in get_client_from_state
    return self.get_client(job_destination_params, job_id)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/jobs/runners/pulsar.py", line 596, in get_client
    return self.client_manager.get_client(job_destination_params, **get_client_kwds)
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.8/site-packages/pulsar/client/manager.py", line 212, in get_client
    destination_params.update(**kwargs)
TypeError: update() got an unexpected keyword argument 'files_endpoint'

  File "/Users/mvandenb/src/galaxy/lib/galaxy/jobs/runners/__init__.py", line 136, in run_next
    method(arg)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/jobs/runners/pulsar.py", line 628, in finish_job
    self.fail_job(job_state, message=GENERIC_REMOTE_ERROR, exception=True)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/jobs/runners/pulsar.py", line 654, in fail_job
    self.stop_job(job_state.job_wrapper)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/jobs/runners/pulsar.py", line 681, in stop_job
    client = self.get_client(job.destination_params, job.job_runner_external_id)
  File "/Users/mvandenb/src/galaxy/lib/galaxy/jobs/runners/pulsar.py", line 596, in get_client
    return self.client_manager.get_client(job_destination_params, **get_client_kwds)
  File "/Users/mvandenb/src/galaxy/.venv/lib/python3.8/site-packages/pulsar/client/manager.py", line 212, in get_client
    destination_params.update(**kwargs)
TypeError: update() got an unexpected keyword argument 'files_endpoint'
```
I did touch some pulsar thing, let me see what that could be

lib/galaxy/jobs/runners/pulsar.py
```diff
@@ -593,6 +593,8 @@ def get_client(self, job_destination_params, job_id, env=None):
             files_endpoint=files_endpoint,
             env=env
         )
+        # Turn MutableDict into standard dict for pulsar consumption
+        job_destination_params = dict(job_destination_params.items())
```
mvdbeek, line 597: So this could be a bugfix ... or not, but probably it is a bugfix. This used to be a a mutable column, so when in the pulsar client we do https://github.com/galaxyproject/pulsar/blob/68a52084468b48a447a4849498d30ac69b805900/pulsar/client/manager.py#L87 we actually change the job_destination_params in the database.

jmchilton: I did know it updated those, Nate and I hit that multiple times last week. I don't know if we're dependent on those database parameters being updated or not (past John didn't leave any clues why he did this xref https://github.com/galaxyproject/pulsar/commit/e4aeb22bca67ab3d20555b81fa73b33a8390d0f9).

I'm okay with merging this and seeing if something breaks. It should be pretty clear if files_endpoint or something needs to be populated initially for updates to work or something odd like that.

PR #11289, opened 2021-02-03
[21.01] Fix /u/username/* routes.

dannon:
Fix an issue in username routes -- when this was swapped from VALID_PUBLICNAME_RE to VALID_PUBLICNAME_SUB we lost the `+` which was causing route matching to fail for usernames longer than one character...  Oops.

The underlying route in the mapper also failed through to match `/{resource}/display_by_username_and_slug?args**`, so none of the testing noticed this either.

xref: https://gitter.im/galaxyproject/Lobby?at=601ad5e9ae4b9b27c1987f66  (@bwlang reported this on Gitter)


bwlang (approved):
looks safe to me

jdavcs (approved):

PR #11293, opened 2021-02-04
[21.01] Raise correct exceptions in `MetadataCollection`

nsoranzo:
Issue found by @mvdbeek when testing https://github.com/galaxyproject/tools-iuc/blob/master/tools/jbrowse/jbrowse.xml with `planemo test`.

Introduced in commit 335a89772012041538e86835a4e16f3a2e50d51f .

Alternative to https://github.com/galaxyproject/galaxy/pull/11281 .

Also 2 other small changes that helped me during the debugging.

### DIFF
--- a/lib/galaxy/model/metadata.py
+++ b/lib/galaxy/model/metadata.py
@@ -94,7 +94,7 @@ class MetadataCollection(Mapping):
         try:
             return self.__getattr__(key)
         except Exception:
-            return KeyError
+            raise KeyError
 
     def __len__(self):
         return len(self.spec)
@@ -113,6 +113,7 @@ class MetadataCollection(Mapping):
             return self.spec[name].wrap(self.spec[name].default, object_session(self.parent))
         if name in self.parent._metadata:
             return self.parent._metadata[name]
+        raise AttributeError
 
     def __setattr__(self, name, value):
         if name == "parent":
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -8,6 +8,7 @@ jobs:
     name: Test
     runs-on: ubuntu-18.04
     strategy:
+      fail-fast: false
       matrix:
         python-version: [3.7]
         subset: ['upload_datatype', 'extended_metadata', 'kubernetes', 'not (upload_datatype or extended_metadata or kubernetes)']
--- a/lib/galaxy/util/__init__.py
+++ b/lib/galaxy/util/__init__.py
@@ -1041,8 +1041,9 @@ def unicodify(value, encoding=DEFAULT_ENCODING, error='replace', strip_null=Fals
         if not isinstance(value, str):
             value = str(value, encoding, error)
     except Exception as e:
-        msg = "Value '{}' could not be coerced to Unicode: {}('{}')".format(value, type(e).__name__, e)
-        raise Exception(msg)
+        msg = "Value '{}' could not be coerced to Unicode: {}('{}')".format(repr(value), type(e).__name__, e)
+        log.exception(msg)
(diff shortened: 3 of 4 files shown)


mvdbeek (dismissed):
Awesome, thank you!

lib/galaxy/model/metadata.py
```diff
@@ -113,6 +113,7 @@ def __getattr__(self, name):
             return self.spec[name].wrap(self.spec[name].default, object_session(self.parent))
         if name in self.parent._metadata:
             return self.parent._metadata[name]
+        raise AttributeError
```
mvdbeek, line 116: I think this was supposed to return `None` silently.

mvdbeek:
Hmm, those test failure look real.

lib/galaxy/model/metadata.py
nsoranzo: No, see https://docs.python.org/3/reference/datamodel.html?highlight=__getattr__#object.__getattr__ : "This method should either return the (computed) attribute value or raise an `AttributeError` exception."

mvdbeek: But that's not how we use MetadataCollection. We're stuck with this behavior now afaict.

nsoranzo:
> Hmm, those test failure look real.

Fixed the unit tests in a rebase, the failures were a by-product of fixing `unicodify()`, it seems that Cheetah is catching the `NotFound` exception so we need to re-raise the same exception type.

lib/galaxy/model/metadata.py
nsoranzo: All tests pass (except for the broken kubernetes integration build), should we give it a try? Or do you have an example were this fails?

mvdbeek: Let's try. Thanks @nsoranzo!

mvdbeek: Yep, that's erroring now in sentry (https://sentry.galaxyproject.org/sentry/main/issues/1325196/):
```
AttributeError: null
  File "galaxy/web/framework/middleware/sentry.py", line 43, in __call__
    iterable = self.application(environ, start_response)
  File "/cvmfs/main.galaxyproject.org/venv/lib/python3.6/site-packages/paste/recursive.py", line 85, in __call__
    return self.application(environ, start_response)
  File "galaxy/web/framework/middleware/statsd.py", line 33, in __call__
    req = self.application(environ, start_response)
  File "/cvmfs/main.galaxyproject.org/venv/lib/python3.6/site-packages/paste/httpexceptions.py", line 640, in __call__
    return self.application(environ, start_response)
  File "galaxy/web/framework/base.py", line 136, in __call__
    return self.handle_request(environ, start_response)
  File "galaxy/web/framework/base.py", line 215, in handle_request
    body = method(trans, **kwargs)
  File "galaxy/webapps/galaxy/controllers/visualization.py", line 488, in display_by_username_and_slug
    visualization_config = self.get_visualization_config(trans, visualization)
  File "galaxy/webapps/base/controller.py", line 1035, in get_visualization_config
    tracks.append(pack_track(drawable_dict))
  File "galaxy/webapps/base/controller.py", line 996, in pack_track
    "filters": track_dict.get('filters', {'filters' : track_data_provider.get_filters()}),
  File "galaxy/visualization/data_providers/genome.py", line 249, in get_filters
    if self.original_dataset.metadata.viz_filter_cols:
  File "galaxy/model/metadata.py", line 116, in __getattr__
    raise AttributeError
```
That's code we have control over, so we could catch the attribute error here, but there's probably tool code does such checks that should just return None.

PR #11294, opened 2021-02-04
[21.01] Unpin pulsar-galaxy-lib in packages

dannon:
It's pinned to 0.14.1 in relevant files, so it needs to change from .0.dev4 in any event.

### DIFF
--- a/packages/app/requirements.txt
+++ b/packages/app/requirements.txt
@@ -8,7 +8,7 @@ galaxy-web-stack
 kombu
 Beaker
 pykwalify
-pulsar-galaxy-lib==0.14.0.dev4
+pulsar-galaxy-lib
 gxformat2
 Mako
 sqlitedict


packages/app/requirements.txt
```diff
@@ -8,7 +8,7 @@ galaxy-web-stack
 kombu
 Beaker
 pykwalify
-pulsar-galaxy-lib==0.14.0.dev4
+pulsar-galaxy-lib==0.14.1
```
mvdbeek: I think we don't need to pin released versions in packages (as long as we don't have infrastructure to sync pins). I think the pin was because `0.14.0.dev4` wasn't released yet?

dannon: That makes sense to me; in any event we don't want it pinned to `.0.dev4` I guess.  I'll drop the pin here.

Issue #11275, opened 2021-02-02
Composite dataset items fail to render in workflow report PDF


wm75:
https://github.com/galaxyproject/galaxy/pull/9769 brought support for referencing elements of composite datatypes in workflow reports - a great feature!
Currently though these references do not seem to work properly when exporting the report to pdf.

To reproduce the issue take a look at this public history:
https://usegalaxy.eu/u/wolfgang-maier/h/reportcompositetest

and try to run this workflow:
https://usegalaxy.eu/u/wolfgang-maier/w/test-wf-report-with-composite
with dataset 1 of the history.

wm75:
@OlegZharkov if you could come up with a fix for the image display at least, it would help me greatly.

mvdbeek:
I guess https://github.com/galaxyproject/galaxy/pull/11296 closes this ?

PR #11296, opened 2021-02-04
Fix images in workflow PDF report

OlegZharkov:
Addressing https://github.com/galaxyproject/galaxy/issues/11275. 
From now on, users would have to provide full path e.g. 

```galaxy
 history_dataset_as_image(history_dataset_id=a3ff968b01c9046c, path="GSM461178_untreat_paired_subset_1_fastq_fastqc/Images/per_sequence_quality.png")
```

If a user wants to know, what is the fullpath of composite dataset, he can either run 

```galaxy
history_dataset_index(output="{output_label}")
```
and get  
image

or just download dataset and see the full path himself

This PR fixes composite images in report PDF. Example:
[report.pdf](https://github.com/galaxyproject/galaxy/files/5927032/report.pdf)

### DIFF
--- a/lib/galaxy/managers/markdown_util.py
+++ b/lib/galaxy/managers/markdown_util.py
@@ -388,8 +388,7 @@ class ToBasicMarkdownDirectiveHandler(GalaxyInternalMarkdownDirectiveHandler):
     def handle_dataset_as_image(self, line, hda):
         dataset = hda.dataset
         name = hda.name or ''
-        filepath =  re.search(PATH_LABEL_PATTERN, line).group(2)
-
+        filepath = re.search(PATH_LABEL_PATTERN, line).group(2)
 
         if filepath is not None:
             file = os.path.join(hda.extra_files_path, filepath)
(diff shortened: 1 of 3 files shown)


lib/galaxy/managers/markdown_util.py
```diff
@@ -383,7 +384,14 @@ def _display_dataset_content(self, hda, header="Contents"):
     def handle_dataset_as_image(self, line, hda):
         dataset = hda.dataset
         name = hda.name or ''
-        with open(dataset.file_name, "rb") as f:
+        filepath = re.search(PATH_LABEL_PATTERN, line).group(2)
```
jmchilton: Wouldn't this break if the path is not present (a more typical use case)? Would something like:

```
path_match = re.search(PATH_LABEL_PATTERN, line)
if path_match:
   filepath = path_match.group(2)
   file = os.path.join(hda.extra_files_path, filepath)
else:
  file = dataset.file_name
```

Make more sense?

OlegZharkov: thanks, did it :)

jmchilton:
Feel free to open a PR to backport this to 21.01 (or even 20.09 if it is relevant).

OlegZharkov:
> Feel free to open a PR to backport this to 21.01 (or even 20.09 if it is relevant).

Yes. I should've done it right away

PR #11299, opened 2021-02-04
Drop legacy metadata strategy  

mvdbeek:
and fix a few shortcomings of the import store with LDDAs and DatasetCollections in mapped-over jobs


jmchilton (approved):

mvdbeek:
I've pulled this into https://github.com/galaxyproject/galaxy/pull/11288, couldn't get https://github.com/galaxyproject/galaxy/pull/11288 to work without this change.

PR #11300, opened 2021-02-04
Move docstring linting to GitHub workflows

nsoranzo:
Also:
- Fix pip cache key.
- Quote Python versions in YAML. Treating them as floats will break with Python 3.10 .

### DIFF
--- a/.github/workflows/maintenance_bot.yaml
+++ b/.github/workflows/maintenance_bot.yaml
@@ -10,6 +10,11 @@ jobs:
     env:
       MILESTONE_NUMBER: 19
     steps:
+      - name: Debug
+        run: |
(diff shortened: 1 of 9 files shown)


.github/workflows/maintenance_bot.yaml
```diff
@@ -10,6 +10,11 @@ jobs:
     env:
       MILESTONE_NUMBER: 19
     steps:
+      - name: Debug
```
nsoranzo, line 13: In case any one is wondering, I have been trying to understand why sometimes the bot removes manually-added labels, but I couldn't reproduce it reliably on my fork. So it's not a debugging leftover, I'd like to have these lines in for when it reappears.

mvdbeek:
Thank you @nsoranzo!

Issue #11302, opened 2021-02-05
Legacy endpoints not tested if fastapi endpoints exist


jdavcs:
The tests are run with unicorn by default (see [here](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy_test/driver/driver_util.py#L81) and [here](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy_test/driver/driver_util.py#L973)).
Galaxy webapp is [mounted](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/fast_app.py#L122) as a sub-application of the FastAPI app using WSGIMiddleware (see FastAPI docs [here](https://fastapi.tiangolo.com/advanced/wsgi/#using-wsgimiddleware)). However, since the sub-application is mounted onto the same path as the main application, if the same endpoint exists in both controllers, the main application shadows the sub-application. In other words, the legacy controller is *not* tested.

To replicate: pick any controller that has been migrated to FastAPI, e.g. `api/configuration`. The controller will have two sets of endpoints in two classes: `FastAPIConfiguration` (FastAPI) and `ConfigurationController` (legacy). Add a `raise` statement to an endpoint in the *legacy* controller. Run the corresponding tests: `pytest lib/galaxy_test/api/test_configuration.py` - the tests will pass. Now comment out the corresponding endpoint in the *FastAPI* controller and rerun the tests - they will fail because that endoint is no longer shadowed by the new controller.

Possible solution: mark relevant tests and rerun them as a separate github workflow under uwsgi (running the whole suite twice is wasteful, especially considering there will be integration tests as well). To test locally, run with `GALAXY_TEST_USE_UVICORN=0`.

ic4f:
Closing: tested via api_paste workflow.

PR #11303, opened 2021-02-05
Github workfow + run_tests.sh option to run marked tests with uwsgi instead of uvicorn

jdavcs:
Addresses #11302.
Will this work?
I've marked configuration, tours and licences only. If this is an acceptable solution, we'll need to mark all the other migrated controllers. (we can mark at the module level, but I think marking each controller class is better - more explicit).

### DIFF
--- a/.github/workflows/api_legacy.yaml
+++ b/.github/workflows/api_legacy.yaml
@@ -1,0 +1,43 @@
+name: API (legacy endpoints) tests
+on: [push, pull_request]
+env:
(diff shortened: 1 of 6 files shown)


mvdbeek:
Can we close this ? We run legacy tests with each PR using https://github.com/galaxyproject/galaxy/blob/dev/.github/workflows/api_paste.yaml#L5.

ic4f:
> Can we close this ? We run legacy tests with each PR using https://github.com/galaxyproject/galaxy/blob/dev/.github/workflows/api_paste.yaml#L5.

Of course! Good to know.

PR #11308, opened 2021-02-06
Temporarily turn off integration test

jdavcs:
This test appears to be consistently causing an error under the integration/kubernetes github workflow. The error happens due to the database being not available at test collection state. This has not happened until now and is *not* related to the content (what is tested) of the test.

Given the above, it seems reasonable to temporarily turn off the test and investigate it offline.

### DIFF
--- a/test/integration/test_config_defaults.py
+++ b/test/integration/test_config_defaults.py
@@ -283,8 +283,10 @@ def get_path_data():
 def get_key(option_data):
     return option_data.key
 
-
-@pytest.mark.parametrize('data', get_config_data(), ids=get_key)
+# TODO: commented out due to consistent errors at test collection stage
+# under integration/kubernetes github workflow. Needs to be resolved.
+#@pytest.mark.parametrize('data', get_config_data(), ids=get_key)
+@pytest.mark.skip(reason="Consistent errors due to db n/a at test collection stage under integration/k8s workflow")
 def test_config_option(data, driver):
     assert data.expected == data.loaded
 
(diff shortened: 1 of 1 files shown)


test/integration/test_config_defaults.py
```diff
@@ -283,8 +283,10 @@ def get_path_data():
 def get_key(option_data):
     return option_data.key
 
-
-@pytest.mark.parametrize('data', get_config_data(), ids=get_key)
+# TODO: commented out due to consistent errors at test collection stage
+# under integration/kubernetes github workflow. Needs to be resolved.
+#@pytest.mark.parametrize('data', get_config_data(), ids=get_key)
+@pytest.mark.skip(reason="Consistent errors due to db n/a at test collection stage under integration/k8s workflow")
```
mvdbeek, line 290: This fails during test collection, but skip doesn't skip test collection (unless the tests prove me wrong). Maybe just drop the test and re-add it when it's fixed ?

jdavcs: But the `parametrize` mark is commented out - so `get_config_data()` is not called. I added `skip` because without the data it's no longer functional, not to prevent test collection hitting the db. Maybe it's the other one - I've commented out that one, let's see if that works. If not, sure, we can just drop it.

jdavcs: Apparently, this test was not at fault - workflow is failing without it. I guess it was just the first one to hit the database. The db is n/a. I'll close this PR as it's not solving the issue.

ic4f:
Closing, as this test was not causing the error.

PR #11309, opened 2021-02-06
Allow element_identifiers or a manually specified name for datasets in build list tool

### LINKED ISSUE #7946: Allow specifying identifiers in Build List tool
Currently if I wan't my list identifiers to be specific values I have to embed a static file into the workflow using a tertiary tool and then use `Relabel List Identifiers`.

bernt-matthias:
allow also element_identifiers or a manually specified name

- [x] add tests to the tool
- [x] implement check for duplicate names?
  - I think it should be fine to allow duplicated names

would fix https://github.com/galaxyproject/galaxy/issues/7946

### DIFF
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -2955,16 +2955,15 @@ class BuildListCollectionTool(DatabaseOperationTool):
         new_elements = {}
 
         for i, incoming_repeat in enumerate(incoming["datasets"]):
-            if not incoming_repeat["input"]:
-                continue
-            new_dataset = incoming_repeat["input"].copy(copy_tags=tags)
-            if incoming_repeat["id_cond"]["id_select"] == 'idx':
-                identifier = str(i)
-            elif incoming_repeat["id_cond"]["id_select"] == 'identifier':
-                identifier = incoming_repeat["input"].element_identifier
-            elif incoming_repeat["id_cond"]["id_select"] == 'manual':
-                identifier = incoming_repeat["id_cond"]["identifier"]
-            new_elements[identifier] = new_dataset
+            if incoming_repeat["input"]:
+                if incoming_repeat["id_cond"]["id_select"] == 'idx':
+                    identifier = str(i)
+                elif incoming_repeat["id_cond"]["id_select"] == 'identifier':
+                    identifier = incoming_repeat["input"].name
+                elif incoming_repeat["id_cond"]["id_select"] == 'manual':
+                    identifier = incoming_repeat["id_cond"]["identifier"]
+                new_elements[identifier] = incoming_repeat["input"].copy(copy_tags=tags)
+
         self._add_datasets_to_history(history, new_elements.values())
         output_collections.create_collection(
             next(iter(self.outputs.values())), "output", elements=new_elements, propagate_hda_tags=False
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -2959,7 +2959,7 @@ class BuildListCollectionTool(DatabaseOperationTool):
                 if incoming_repeat["id_cond"]["id_select"] == 'idx':
                     identifier = str(i)
                 elif incoming_repeat["id_cond"]["id_select"] == 'identifier':
-                    identifier = incoming_repeat["input"].element_identifier
+                    identifier = getattr(incoming_repeat["input"], 'element_identifier', incoming_repeat["input"].name)
                 elif incoming_repeat["id_cond"]["id_select"] == 'manual':
                     identifier = incoming_repeat["id_cond"]["identifier"]
                 new_elements[identifier] = incoming_repeat["input"].copy(copy_tags=incoming_repeat["input"].tags)
--- a/lib/galaxy/tools/__init__.py
+++ b/lib/galaxy/tools/__init__.py
@@ -2962,7 +2962,7 @@ class BuildListCollectionTool(DatabaseOperationTool):
                     identifier = incoming_repeat["input"].name
                 elif incoming_repeat["id_cond"]["id_select"] == 'manual':
                     identifier = incoming_repeat["id_cond"]["identifier"]
-                new_elements[identifier] = incoming_repeat["input"].copy(copy_tags=tags)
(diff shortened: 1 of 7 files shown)


jmchilton:
This seems like a useful change, but I just placed that file in conflict by merging #11068 - can you rebase?

mvdbeek (approved):

bernt-matthias:
Could need a little help with the failing test. Apparently the modified tool is loaded and I guess that I need to specify the new input in the workflow? 

Also I see 

```
galaxy.workflow.modules INFO 2021-09-08 19:44:15,056 [pN:main,p:2225,tN:worker 57] Exact tool specified during workflow module creation for [__BUILD_LIST__] but couldn't find correct version [1.0.0].
galaxy.workflow.modules DEBUG 2021-09-08 19:44:15,107 [pN:main,p:2225,tN:worker 57] <span title="tool id '__BUILD_LIST__'">Using version '1.1.0' instead of version '1.0.0' specified in this workflow. 
```

mvdbeek:
You'll have to add the tool to https://github.com/galaxyproject/galaxy/blob/c357f5e4fef1e841860c18fd462405722297046a/lib/galaxy/tools/__init__.py#L217

lib/galaxy/tools/__init__.py
```diff
@@ -2956,7 +2956,13 @@ def produce_outputs(self, trans, out_data, output_collections, incoming, history
 
         for i, incoming_repeat in enumerate(incoming["datasets"]):
             if incoming_repeat["input"]:
-                new_elements["%d" % i] = incoming_repeat["input"].copy(copy_tags=incoming_repeat["input"].tags)
+                if incoming_repeat["id_cond"]["id_select"] == 'idx':
+                    identifier = str(i)
+                elif incoming_repeat["id_cond"]["id_select"] == 'identifier':
+                    identifier = incoming_repeat["input"].name
```
mvdbeek: Can you check if we have access to the element identifier here ? You can map this tool over a collection, so I think the element identifier would be the right attribute to use if it is there.

mvdbeek: Well, I guess that's a no:
```
galaxy.tools ERROR 2021-09-10 13:02:30,579 [pN:main,p:2140,tN:ThreadPoolExecutor-0_8] Exception caught while attempting to execute tool with id '__BUILD_LIST__':
  File "/home/runner/work/galaxy/galaxy/galaxy root/lib/galaxy/tools/__init__.py", line 1750, in handle_single_execution
    flush_job=flush_job,
  File "/home/runner/work/galaxy/galaxy/galaxy root/lib/galaxy/tools/__init__.py", line 1838, in execute
    return self.tool_action.execute(self, trans, incoming=incoming, set_output_hid=set_output_hid, history=history, **kwargs)
  File "/home/runner/work/galaxy/galaxy/galaxy root/lib/galaxy/tools/actions/model_operations.py", line 61, in execute
    self._produce_outputs(trans, tool, out_data, output_collections, incoming=incoming, history=history, tags=preserved_tags, hdca_tags=preserved_hdca_tags)
  File "/home/runner/work/galaxy/galaxy/galaxy root/lib/galaxy/tools/actions/model_operations.py", line 77, in _produce_outputs
    tool.produce_outputs(trans, out_data, output_collections, incoming, history=history, tags=tags, hdca_tags=hdca_tags, tag_handler=tag_handler)
  File "/home/runner/work/galaxy/galaxy/galaxy root/lib/galaxy/tools/__init__.py", line 2962, in produce_outputs
    identifier = incoming_repeat["input"].element_identifier
AttributeError: 'HistoryDatasetAssociation' object has no attribute 'element_identifier'
```

bernt-matthias: We can try/except .. but I guess when mapped over the tool will still get an HDA, or?

mvdbeek: Yes, but at that point we may have set the element_identifier ... I think that maybe be happening in https://github.com/jmchilton/galaxy/blob/6412363dcca1c005cd93d48d5c72354d46877b45/lib/galaxy/tools/parameters/meta.py#L252

mvdbeek: Also I'd suggest `getattr(incoming_repeat["input"], 'element_identifier', incoming_repeat["input"].name)` ... but only if the mapping over works :)

bernt-matthias: According to the nice figures that have been added recently to the help (https://github.com/galaxyproject/galaxy/pull/11068) mapping over should work. But I'm not really sure how to test (manually) if the element identifier is really used. Do you have an idea?

mvdbeek: I meant whether the element identifiers are really being added at this point. Yes, mapping over definitely works (and #11068 took some shortcuts to explain things that are more complex ...)

mvdbeek: You'd have to add an API test if you want to be sure, there's no way to test this with the tool testing framework.

bernt-matthias: > You'd have to add an API test if you want to be sure, there's no way to test this with the tool testing framework.

I just wanted to do a manual test and was wondering how I can produce a difference in name and identifier that I can observe in the output?

mvdbeek: You can create a list in the UI and rename some of the items.
Screenshot 2021-09-15 at 18 10 37

bernt-matthias:
Your suggestion (https://github.com/galaxyproject/galaxy/pull/11309/commits/6a06f8e1e882a35c0af8ef8391542e232f27fe94) seems to work.

For a test I created a collection and renamed the 2nd element. This is the result:

Screenshot from 2021-09-16 11-56-34

mvdbeek:
Nice, thank you!

bernt-matthias:
API test seems to fail because `lib/galaxy_test/base/data/test_subworkflow_with_integer_input.ga` uses the build list tool and does not define a value. 

Not sure why this fails. Output suggests that the default is used. I have not so much experience with the workflow format to change this.

mvdbeek:
Hmm, so that's tricky. I think I'm just gonna copy the old tool and we load both versions. That's probably more robust than what the alternative would be, which is to allow tool state changes for tools in WORKFLOW_SAFE_TOOL_VERSION_UPDATES

PR #11313, opened 2021-02-07
consider collections in all tests 

bernt-matthias:
In https://github.com/galaxyproject/galaxy/pull/10304 we realized that collections are ignored in all but the 1st test. @jmchilton came up with a fix. Unfortunately the PR is stalled. So I suggest to go on with this cherry pick...

Testing in IUC https://github.com/galaxyproject/tools-iuc/issues/3466

### DIFF
--- a/lib/galaxy/tool_util/verify/interactor.py
+++ b/lib/galaxy/tool_util/verify/interactor.py
@@ -1290,7 +1290,7 @@ class ToolTestDescription:
         self.error = processed_test_dict.get("error", False)
         self.exception = processed_test_dict.get("exception", None)
 
-        self.output_collections = list(map(TestCollectionOutputDef.from_dict, processed_test_dict.get("output_collections", [])))
+        self.output_collections = [TestCollectionOutputDef.from_dict(d) for d in processed_test_dict.get("output_collections", [])]
         self.command_line = processed_test_dict.get("command_line", None)
         self.command_version = processed_test_dict.get("command_version", None)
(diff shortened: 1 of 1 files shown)


lib/galaxy/tool_util/verify/interactor.py
```diff
@@ -1290,7 +1290,7 @@ def __init__(self, processed_test_dict):
         self.error = processed_test_dict.get("error", False)
         self.exception = processed_test_dict.get("exception", None)
 
-        self.output_collections = map(TestCollectionOutputDef.from_dict, processed_test_dict.get("output_collections", []))
+        self.output_collections = list(map(TestCollectionOutputDef.from_dict, processed_test_dict.get("output_collections", [])))
```
mvdbeek: How about 
```suggestion
        self.output_collections = [TestCollectionOutputDef.from_dict(d) for d in processed_test_dict.get("output_collections", [])]
```

mvdbeek (approved):

bernt-matthias:
CI on the IUC repo (excluding qiime) yields 21 failures with this change: https://github.com/bernt-matthias/tools-iuc/actions/runs/551052815

I guess the good message is that its not more :) 

Somehow I feel like suggesting to include this in 21.01 (its a bugfix) .. but we could also do this after the release if you prefer less noise during the release.

mvdbeek:
Yeah, let's backport it, the test failures are not too bad IMO.

PR #11315, opened 2021-02-08
Prepare FastAPI route for quotas

davelopez:
Ref #10889

## Overview
- Added some integration tests for the `quotas` API.
- Moved the `UserModel` from `lib/galaxy/managers/users.py` to `lib/galaxy/schema/schema.py`.
- Introduced the [ModelClassField](https://github.com/galaxyproject/galaxy/compare/dev...davelopez:prepare_fastapi_route_quotas?expand=1#diff-8f4ada8155446c24a5cc6d821b15a12c2fc7dedc4e8e615b187689ae08062708R53) to represent a particular `model_class` in pydantic models. See additional comments [*].
- Created pydantic models for quota operations.
- Prepare the interfaces for `QuotaManager` and `QuotasManager`.
- Added the FastAPI route with all endpoint operations (disabled)

## Additional Comments
- Some of the legacy controller actions return a raw string, which, sometimes, is a ';' separated list of messages. Should this be at least a JSON with an actual list of strings?
- [*] Now most of the legacy API payloads that involve database entities have an encoded id and an additional `model_class` field to indicate the type of entity. After modeling every payload with pydantic models, this `model_class` field will be redundant? or what is the real purpose of this field?

## Update

- [x] Refactor [AdminActions](https://github.com/galaxyproject/galaxy/blob/f6623d3a142149cdd5f0c563ace84dd52c691202/lib/galaxy/actions/admin.py#L12) and [UsesQuotaMixin](https://github.com/galaxyproject/galaxy/blob/f6623d3a142149cdd5f0c563ace84dd52c691202/lib/galaxy/webapps/base/controller.py#L1510) into a new [QuotaManager](https://github.com/galaxyproject/galaxy/compare/dev...davelopez:prepare_fastapi_route_quotas?expand=1#diff-d0ec4c74ff591699ce3812a6a6cbd036906c45c4b3b3616f36915fd1e1eb3e5cR49) class and remove the old mixins from [QuotaAPIController](https://github.com/galaxyproject/galaxy/blob/f6623d3a142149cdd5f0c563ace84dd52c691202/lib/galaxy/webapps/galaxy/api/quotas.py#L28) and [AdminGalaxy](https://github.com/galaxyproject/galaxy/blob/f6623d3a142149cdd5f0c563ace84dd52c691202/lib/galaxy/webapps/galaxy/controllers/admin.py#L508). This new `QuotaManager` will be used then (by composition) where needed.
- [x] Replace [QuotaParamParser](https://github.com/galaxyproject/galaxy/blob/f6623d3a142149cdd5f0c563ace84dd52c691202/lib/galaxy/web/params.py#L18) by the corresponding pydantic models in [_schema.py](https://github.com/galaxyproject/galaxy/compare/dev...davelopez:prepare_fastapi_route_quotas?expand=1#diff-a6ebb28a1ea10aac025dbf5658ad8a0f7d1b3d1bc90c378eb61c64476236d65eR164)
- [x] Refactor move the logic in `QuotaAPIController` to [QuotasManager](https://github.com/galaxyproject/galaxy/compare/dev...davelopez:prepare_fastapi_route_quotas?expand=1#diff-d0ec4c74ff591699ce3812a6a6cbd036906c45c4b3b3616f36915fd1e1eb3e5cR20).
- [x] Move `BaseAPIController.validate_in_users_and_groups` to QuotasManager since it is currently only used by quotas and that way it can be shared by both controllers.
- [x] Updated legacy_expose_api, xref #11252
- [x] In https://github.com/galaxyproject/galaxy/pull/11315/commits/aa6249f45d4ddfbc164c7babdee59900e2791f97 I assumed the id must be encoded for the UserModel. This affects the `api/whoami` in configuration.py which is also using schema.UserModel. Hope these changes are ok @ic4f 

Sorry for the long PR😅


davelopez:
@jmchilton 

What do you think would be the correct course of action here?
1. Leave this PR on hold until #10977 is merged and then rebase before starting to refactor the managers.
2. Merge this PR as it is (after fixing all checks and it is approved in review). And then opening a new one for the refactoring tasks after #10977 is completed.
3. Maybe a better approach?

mvdbeek:
> * After modeling every payload with pydantic models, this `model_class` field will be redundant? or what is the real purpose of this field?

We still need this if an endpoint returns a union of something and the client needs to know what it is dealing with. A common example is the history endpoint that returns either `HistoryDatasetAssociation` or `HistoryDatasetCollectionAssociation`.

> * Some of the legacy controller actions return a raw string, which, sometimes, is a ';' separated list of messages. Should this be at least a JSON with an actual list of strings?

I think that's a case-by-case decision to take. If we do change this we may also have to update whatever frontend component consumes this, and depending on the context (is this likely something external scripts are consuming ?) we would need to provide a route for these legacy consumers (i.e we could add a parameter that controls the response model).

lib/galaxy/webapps/galaxy/api/quotas.py
```diff
@@ -15,16 +19,154 @@
...
+    Depends,
+    get_admin_user,
+    get_trans,
+)
 
 log = logging.getLogger(__name__)
 
 
+# TODO: This FastAPI router is disabled. Please rename it to `router` when the database session issues are fixed.
+_router = APIRouter(tags=['quotas'])
```
jmchilton: Want to rebase this to actually use this now that it works?

davelopez: I am afraid the TODO comment here is misleading, is not disabled because of the session issues 😅 
The real reason to be disabled is that the manager classes are not implemented yet, [see this](https://github.com/galaxyproject/galaxy/pull/11315/files#diff-d0ec4c74ff591699ce3812a6a6cbd036906c45c4b3b3616f36915fd1e1eb3e5cR20).

I was waiting for some advice on how to proceed, [see here](https://github.com/galaxyproject/galaxy/pull/11315#issuecomment-776674134) :)

jmchilton: I'm going to regret saying it - but I don't know that you have to wait on  #10977. Probably better to merge this sooner. It is an easier thing to do.

davelopez: Ok, thanks @jmchilton, I will mark this as a draft again and continue working on it as soon as I can :)

davelopez:
Waiting for #11608

jmchilton:
Thank you - this looks really, really nice - sorry for the delay in reviewing it.

bgruening:
Cool @davelopez congrats! Thanks @jmchilton for reviewing!

davelopez:
No worries and many thanks @jmchilton!

PR #11316, opened 2021-02-08
Merge 20.09 -> 21.01 -> dev

mvdbeek:
Since this is a bit bigger I wanna see the tests


dannon:
Still sure you want to see those tests?  😅

mvdbeek:
Nah, all good here, nothing to see 😆

lib/galaxy/dependencies/pinned-requirements.txt
```diff
@@ -161,7 +161,7 @@ six==1.15.0; python_version >= "3.6" and python_full_version < "3.0.0" and pytho
 social-auth-core==3.3.0
 sortedcontainers==2.3.0
 sqlalchemy-migrate==0.13.0
-sqlalchemy-mutable==0.0.11
+sqlalchemy-mutable==0.0.11; python_version >= "3.6"
```
dannon, line 164: We don't support < 3.6 at all, anyway, right?  Can we drop the other pin for sqlalchemy, too, and update whatever was requiring this?

mvdbeek: Yeah, but since this is all auto-generated I guess there's no point in modifying this manually ?

dannon: Yeah, swapped to gitter for more discussion -- what I was getting at is wondering if we can just simplify all this by specifying 3.6+ at the top level (in pyproject.toml or wherever) and not have poetry worry with resolving for < 3.6.

Not manually un-doing the pins.

nsoranzo: We already specify the Python >=3.6 requirement: https://github.com/galaxyproject/galaxy/blob/dev/pyproject.toml#L57 , but this seems to be ignored when generating the `requirements.txt` files.

dannon: 👍 Yep, I looked through the poetry docs briefly and don't see a way to tell it to just ignore python < 3.6.

PR #11317, opened 2021-02-08
[20.09] Fix Workflow report markdown lost upon renaming WF

### LINKED ISSUE #11280: Workflow report markdown lost upon renaming WF
The title says it all, but exact steps to reproduce:

- add some reporting markdown to a WF and save
- go back to the workflow menu and rename the WF
- open the WF in the WF editor and inspect the markdown

=> it will have been reverted to the default markdown

davelopez:
Fixes #11280

When renaming a workflow from the popup-menu in the workflows list page...

image

The markdown report was not copied from the previous workflow version and the default markdown report was displayed instead.

I am not 100% sure this is the best way of keeping the markdown report from previous versions after a rename, but it seems to work :)

### DIFF
--- a/lib/galaxy/model/__init__.py
+++ b/lib/galaxy/model/__init__.py
@@ -4835,6 +4835,7 @@ class Workflow(Dictifiable, RepresentById):
         copied_workflow.name = self.name
         copied_workflow.has_cycles = self.has_cycles
         copied_workflow.has_errors = self.has_errors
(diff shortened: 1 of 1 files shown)


bgruening:
@davelopez can you maybe target 20.09? @mvdbeek would that be ok?

mvdbeek:
Sure, and yes, this looks like the right fix

mvdbeek (approved):

PR #11319, opened 2021-02-08
Fix transiently failing roles API test.

jmchilton:
I was able to get it to fail very rarely locally and I noticed some bugs (not returning a json response for instance even though it was annotated as such, decorators in the wrong order) - I don't know which error actually caused the connection reset error but these changes seem to maybe fix it.


mvdbeek (approved):
Can you fix the linting error and rebase against dev for the kubernetes testing fix ?

PR #11322, opened 2021-02-09
Custos `aud` handling fix

dannon:
Consolidate token decode handling in custos_authnz, fix `aud` audience validation to actually validate client id.

Add expiration time to token listing response and handle decode errors related to failed validation (provide error message alongside the galaxy-side info we have)

### DIFF
--- a/lib/galaxy/authnz/custos_authnz.py
+++ b/lib/galaxy/authnz/custos_authnz.py
@@ -42,6 +42,9 @@ class CustosAuthnz(IdentityProvider):
         elif provider == 'keycloak':
             self._load_config_for_keycloak()
 
+    def _decode_token(self, token):
+        return jwt.decode(token, audience=self.config['client_id'], options={"verify_signature": False})
+
     def authenticate(self, trans, idphint=None):
         base_authorize_url = self.config['authorization_endpoint']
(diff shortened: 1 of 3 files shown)


dannon:
I need to update the tests

lib/galaxy/authnz/custos_authnz.py
```diff
@@ -42,6 +42,9 @@ def __init__(self, provider, oidc_config, oidc_backend_config, idphint=None):
         elif provider == 'keycloak':
             self._load_config_for_keycloak()
 
+    def _decode_token(self, token):
```
mvdbeek: For the purpose of anyone accidentally calling this when they meant to verify the signature, can you push `verify_signature` to the function arguments and default it to True ? Or rename it to `_decode_token_no_verify_signature` ? This seems like something we need to be explicit about.

mvdbeek (approved):

lib/galaxy/authnz/custos_authnz.py
dannon: It should only be for internal usage here in the custos module, which only works the one way, but I can rename it.  Now that I'm thinking about it though, I'm going to revisit why we can't just validate the signature prior to proceeding here.

PR #11329, opened 2021-02-09
Drop docker testing image

mvdbeek:
We don't use this anymore and it's currently not building.

It served us well, but I don't think there's value in keeping this around.


jmchilton (approved):
We're locking ourselves into a wildly unethical corporation really rapidly, I'm committed to fighting that but I don't think preserving this is the way to do that unfortunately. Thanks for the cleanup.

PR #11332, opened 2021-02-09
[21.01] Submit input_format to tools API in payload, not as a parameter

simonbray:
A small fix to https://github.com/galaxyproject/galaxy/pull/10720.

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/tools.py
+++ b/lib/galaxy/webapps/galaxy/api/tools.py
@@ -563,7 +563,7 @@ class ToolsController(BaseAPIController, UsesVisualizationMixin):
         # as a regular tool parameter we accept both.
         use_cached_job = payload.get('use_cached_job', False) or util.string_as_bool(inputs.get('use_cached_job', 'false'))
 
-        input_format = payload.get('input_format', 'legacy')
+        input_format = str(payload.get('input_format', 'legacy'))
 
         vars = tool.handle_input(trans, incoming, history=target_history, use_cached_job=use_cached_job, input_format=input_format)
 
--- a/lib/galaxy/webapps/galaxy/api/tools.py
+++ b/lib/galaxy/webapps/galaxy/api/tools.py
@@ -486,8 +486,12 @@ class ToolsController(BaseAPIController, UsesVisualizationMixin):
         POST /api/tools
         Execute tool with a given parameter payload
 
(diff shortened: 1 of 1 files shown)


lib/galaxy/webapps/galaxy/api/tools.py
```diff
@@ -557,7 +554,7 @@ def _create(self, trans: GalaxyWebTransaction, payload, **kwd):
         # as a regular tool parameter we accept both.
         use_cached_job = payload.get('use_cached_job', False) or util.string_as_bool(inputs.get('use_cached_job', 'false'))
 
-        input_format = kwd.get('input_format', 'legacy')
+        input_format = payload.get('input_format', 'legacy')
```
simonbray:
```suggestion
        input_format = payload.get('input_format', '21.01')
```
A bit unrelated - this wasn't discussed in the original PR for some reason, but maybe it makes sense that '21.01' is actually the default value in Galaxy version 21.01?

nsoranzo: I guess the reason to keep `legacy` is default is that Galaxy API is generally very conservative in trying not to break backward-compatibility.

nsoranzo:
I've rebased to target `release_21.01`, which I think was the intention. Also I've restored the `input_format` documentation, I think it's fine even if it's passed through the payload.

nsoranzo (approved):

lib/galaxy/webapps/galaxy/api/tools.py
mvdbeek: That and the frontend doesn't submit in this format.

lib/galaxy/webapps/galaxy/api/tools.py
```diff
@@ -559,7 +563,7 @@ def _create(self, trans, payload, **kwd):
         # as a regular tool parameter we accept both.
         use_cached_job = payload.get('use_cached_job', False) or util.string_as_bool(inputs.get('use_cached_job', 'false'))
 
-        input_format = kwd.get('input_format', 'legacy')
+        input_format = str(payload.get('input_format', 'legacy'))
```
mvdbeek, line 566:
```suggestion
        input_format = payload.get('input_format', 'legacy')
```

mvdbeek: Ah, saw the commit message. We should fix that at the root.

PR #11333, opened 2021-02-09
Beta history docs

Nerdinacan:


bgruening:
Maybe you can link to https://github.com/nefe/You-Dont-Need-jQuery as well.

jmchilton:
Thanks for the docs!

Is there some client technology that will benefit from having this sort of docs in ``src/components`` (e.g. Vue style guide or ES6 doc)? If not I'd suggest starting a new client development section in Sphinx (maybe doc/source/dev/client) so it would be included in https://docs.galaxyproject.org/en/latest/dev/index.html.

I like the idea of the documentation being close to the code so even if there isn't a technical reason maybe I'm still wrong that it should be included in docs or maybe a symlink would be more appropriate if we did decide there was value in including it.

dannon:
The docs here creep a bit past just 'history', but I think the right place for component-specific documentation is the client styleguide.  It's intended to be a living document with documentation *and* live examples that someone can work against.

Nerdinacan:
I'm happy to move it someplace else. I just wanted to start writing it.

My main goal here is going to be to provide a real flowchart of the dependency chain for the caching layer and the history content provider observable chain.

dannon:
@Nerdinacan I'm not advocating moving this *to* anywhere else -- it should be with the code here.  The styleguide (already) just walks client directories looking for markdown and would include this.

And, yeah, the styleguide was broken by some of the init refactoring a little while back and someone needs to fix it.  And thn we need to test the build on PRs to keep it from breaking again.  It's on my list, but I'd welcome anyone else taking a look too.

Nerdinacan:
Cool. I have never even peeked in there enough to see how it runs. Dollars to donuts it was my bumbling that broke it though. I'll see if I can figure out why it doesn't stay running.

Nerdinacan:
> Maybe you can link to https://github.com/nefe/You-Dont-Need-jQuery as well.

Nice summary there. I will link it in.

My main problem with jQuery isn't even the library itself, it's the plugin structure which allows us to arbitrarily add functionality to a global object in a non-modular way with no control over what happens during those module initializations. jQuery is a catastrophe for testing and refactoring and it drags its problems with it wherever it is referenced.

But clearly this rant I went on doesn't belong in the history code documentation at all, maybe further up the tree? I could split this file into multiple docs.

dannon:
@Nerdinacan Yeah, if we could split this into history specific documentation (design, usage, etc) with the other bits included in the client readme or the like, that'd be great.

Nerdinacan:
I'm going to close this PR and use the content as a use-case sample for the fixed styleguide in this PR.

https://github.com/galaxyproject/galaxy/pull/11404

Nerdinacan:
I'm closing this because I put all this content in the style-guide fix PR as a test of the new section generator

PR #11338, opened 2021-02-10
Convert/standardize to fstrings across `lib`

dannon:
Testing this while revisiting my `black` branch.


OlegZharkov (commented):
this is really great step :)
Just a minor comment... What about https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/managers/markdown_util.py#L48? aren't those strings were also supposed to be converted?

dannon:
@OlegZharkov Those raw strings are going to take a closer look to convert, yeah.

dannon:
That test failure might be legit -- taking a look.

PR #11339, opened 2021-02-10
Submit input_format to tools API in payload, not as a parameter

simonbray:
In order for bioblend tests to pass, https://github.com/galaxyproject/galaxy/pull/11332 needs to be merged into dev as well as 21.01.


mvdbeek:
I've merged your PR into dev.

PR #11342, opened 2021-02-10
Resolve session problems, enabled pages and tags fastAPI routes

mvdbeek:


davelopez:
@mvdbeek 

I tried to debug a bit to see what is the cause of the API test failures here.

For the integration tests in [test/integration/test_page_revision_json_encoding.py](https://github.com/galaxyproject/galaxy/blob/aefa49e6a66a21d7bf8812b4b872be8c87224003/test/integration/test_page_revision_json_encoding.py#L16) the reason is the new FastAPI POST action in `/api/pages` does not like `dicts` as body and expects proper JSON. Simply adding `json=true` in line 29 and 53 should fix those tests.

In [lib/galaxy_test/api/test_datasets.py::test_tag_change](https://github.com/galaxyproject/galaxy/blob/aefa49e6a66a21d7bf8812b4b872be8c87224003/lib/galaxy_test/api/test_datasets.py#L124) the reason seems to be more tricky. I think it is that there are 2 different sessions conflicting in [TagsManager.update(...)](https://github.com/galaxyproject/galaxy/blob/aefa49e6a66a21d7bf8812b4b872be8c87224003/lib/galaxy/managers/tags.py#L46), one used by the `tag_handler` in [tag_handler.apply_item_tags](https://github.com/galaxyproject/galaxy/blob/aefa49e6a66a21d7bf8812b4b872be8c87224003/lib/galaxy/managers/tags.py#L55) (that is flushing by default for every tag associated) and the `trans.sa_session` which is trying to flush right after. But I am not so sure here...

mvdbeek:
> But I am not so sure here...

I think you were spot on with the diagnosis. I've created a new GalaxyTagHandlerSession instance in the manager, which takes in the current session. The global GalaxyTagHandler instance takes in a Session when constructed, but that is then not the correct session for a web request.

mvdbeek:
I think the startup test failures are unrelated timeouts when building the client.

jmchilton (approved):

lib/galaxy/webapps/galaxy/api/__init__.py
```diff
@@ -19,17 +20,32 @@
...
 from galaxy.web.framework.decorators import require_admin_message
 from galaxy.work.context import SessionRequestContext
 
 
-def get_app() -> UniverseApplication:
-    return cast(UniverseApplication, galaxy_app.app)
+def get_app() -> Generator[UniverseApplication, None, None]:
+    app = cast(UniverseApplication, galaxy_app.app)
+    try:
+        app.model.set_local_session()
```
jmchilton, line 41: I think we should do this as middleware instead so avoid circular dependency issues and such and needing this to be "depended on" before anything else. That approach would also leverage more standard Python webapp patterns and depends less of specifics of FastAPI dependency injection stuff also. Maybe I'm wrong though and that won't be a problem. I just hit a relatively subtle issue when rebasing #11180 on top of this.

mvdbeek: The reason I didn't use a middleware is that they run on every request, so would also run on future async routes or routes that don't need the database at all. Maybe those should be entirely separated apps, in that case we could omit the sync middleware? I don't mind making this a middleware for now but the dependency approach seems more flexible.

 I think the reason this doesn't feel awesome is that we don't have an isolated `get_db` that doesn't depend on app. Still a hack, but maybe better, would be moving this into `get_model` ?

Is this problem you saw https://github.com/galaxyproject/galaxy/pull/11180#issuecomment-780847940 ? In which case having a middleware would still give you the same issue, and one possible fix is https://github.com/galaxyproject/galaxy/pull/11180#discussion_r578278085

Issue #11321, opened 2021-02-08
Legacy endpoints not tested by integration tests if fastapi endpoint exists


jdavcs:
When migrating an endpoint to fastapi we may sometimes need to add an integration test instead of an api test (e.g. #11167, #11241, #11315). In this case, the test will be run against the fastapi endpoint only (see closed issue #11302 for context; this does *not* apply to api tests, hence issue was closed).

Possible solution: something similar to #11303: mark individual integration tests, have separate workflow to run those tests only. Another option is to run all integration tests with `GALAXY_TEST_USE_UVICORN: false`, but this seems wasteful.

PR #11345, opened 2021-02-11
Github workfow + run_tests.sh option to run marked integration tests on legacy apis

jdavcs:
Addresses #11321. Does this work as a solution?

### DIFF
--- a/.github/workflows/integration_legacy.yaml
+++ b/.github/workflows/integration_legacy.yaml
@@ -1,5 +1,7 @@
 name: Integration Legacy
-on: [push, pull_request]
+on:
+  schedule:
+    - cron: '0 3 * * *'
 env:
   GALAXY_TEST_DBURI: 'postgres://postgres:postgres@localhost:5432/galaxy?client_encoding=utf8'
   GALAXY_TEST_AMQP_URL: 'amqp://localhost:5672//'
--- a/.github/workflows/integration_legacy.yaml
+++ b/.github/workflows/integration_legacy.yaml
@@ -1,4 +1,4 @@
-name: Integration
+name: Integration Legacy
 on: [push, pull_request]
 env:
   GALAXY_TEST_DBURI: 'postgres://postgres:postgres@localhost:5432/galaxy?client_encoding=utf8'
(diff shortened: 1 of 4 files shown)


mvdbeek:
I guess it would, but this seems like unnecessary noise and overhead if we run that on each PR.
Can you change that to a cron job that runs at night?

ic4f:
> I guess it would, but this seems like unnecessary noise and overhead if we run that on each PR.
> Can you change that to a cron job that runs at night?

Yes - done, but commented out. First, I want to check that wrapping the run command works (need to change exit status from 5 to 0 if no tests are run). - and it seems it's not working..

ic4f:
So the new legacy tests fail when they don't find any tests to run - which is the case on this first PR (which is expected); but this will be also the case anytime there are no tests marked for execution. The reason for that is that pytest will exit with status 5 if no tests were run (which is the correct behavior). I don't see an obvious way to capture the exit status of the command within the workflow yaml file. I don't have a good solution yet. It's possible, of course, to capture the exit status inside run_tests.sh (and exit with 0 if it was 5) - but that would require modifying `run_tests.sh`. If that is the only option, I could add a flag to run_tests.sh that does that. 
But I'm not sure if this is a good solution..

.github/workflows/integration_legacy.yaml
```diff
@@ -0,0 +1,79 @@
...
+        if: matrix.subset == 'kubernetes'
+        run: |
+          . .ci/minikube-test-setup/start_services.sh
+          ./run_tests.sh -integration test/integration --legacy-api -- -k "${{ matrix.subset }}"
+        working-directory: 'galaxy root'
+      - uses: actions/upload-artifact@v2
+        if: failure()
+        with:
+          name: Integration test results (${{ matrix.python-version }}, ${{ matrix.subset }})
+          path: 'galaxy root/run_integration_tests.html'
```
mvdbeek, line 79: You can collapse this into one, that should solve the issue of no tests in these chunks.
We only split them to reduce runtime, which is not a concern here.

mvdbeek:
We're working towards removing the legacy API endpoints ASAP, ideally shortly after the 21.05 release, so maybe it's not worth pursuing this further ? I'll drop the milestone for now

ic4f:
> We're working towards removing the legacy API endpoints ASAP, ideally shortly after the 21.05 release, so maybe it's not worth pursuing this further ? I'll drop the milestone for now

I was just thinking about this! So yes, +1, closing it.

Issue #11250, opened 2021-01-28
Adding list:paired and paired DCs to the dataset library is not possible


hexylena:
And from list it seems to just add all of the children? That's so weird. I'd like to build up the DC how I want, add it to a library, and then for people to use that without having to rebuild the DC everytime?

![image](https://user-images.githubusercontent.com/458683/106186333-6c091f00-61a4-11eb-9bf5-8079aefc1a40.png)

PR #11346, opened 2021-02-11
Allow `GET /api/dataset_collections/{hdca_id}` without specifying `instance_type`

nsoranzo:
Also:
- raise `exceptions.NotImplemented` for `GET /api/dataset_collections`
- add/fix API docs

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/dataset_collections.py
+++ b/lib/galaxy/webapps/galaxy/api/dataset_collections.py
@@ -2,7 +2,7 @@ from logging import getLogger
 
 import routes
 
-from galaxy.exceptions import ObjectNotFound, RequestParameterInvalidException
+from galaxy import exceptions
 from galaxy.managers.base import decode_id
 from galaxy.managers.collections import DatasetCollectionManager
 from galaxy.managers.collections_util import (
@@ -31,8 +31,7 @@ class DatasetCollectionsController(
 
     @expose_api
     def index(self, trans, **kwd):
-        trans.response.status = 501
-        return 'not implemented'
+        raise exceptions.NotImplemented
 
     @expose_api
     def create(self, trans: ProvidesHistoryContext, payload: dict, **kwd):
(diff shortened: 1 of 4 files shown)


lib/galaxy/webapps/galaxy/api/dataset_collections.py
```diff
@@ -63,14 +62,18 @@ def create(self, trans: ProvidesHistoryContext, payload: dict, **kwd):
...
         dataset_collection_instance = self.__service.create(trans=trans, **create_params)
         return dictify_dataset_collection_instance(dataset_collection_instance,
                                                    security=trans.security, parent=create_params["parent"])
 
     @expose_api
-    def show(self, trans: ProvidesHistoryContext, instance_type, id, **kwds):
+    def show(self, trans: ProvidesHistoryContext, id, instance_type='history', **kwds):
+        """
+        GET /api/dataset_collections/{hdca_id}
+        GET /api/dataset_collections/{ldca_id}?instance_type=library
```
mvdbeek, line 75: ldca's are not really a galaxy type, these are still hdcas, but the parent is a library.
```suggestion
        GET /api/dataset_collections/{hdca_id}?instance_type=library
```
Seems more accurate

nsoranzo: We do have `self.model.LibraryDatasetCollectionAssociation` in `lib/galaxy/model/__init__.py` and it seems to be used when creating a collection with `POST /api/dataset_collections` and `instance_type=library` in the payload (see also `lib/galaxy/managers/collections.py`).

mvdbeek: Thanks, I did not know that ...

mvdbeek (approved):

lib/galaxy/webapps/galaxy/api/dataset_collections.py
nsoranzo: NP, to be fair [it is not yet possible to create or add collections to libraries in the UI](https://github.com/galaxyproject/galaxy/issues/11250), so...

PR #11259, opened 2021-01-29
Manage folder rewrite in Vue & VueRouter

OlegZharkov:
Most importantly, this PR brings _VueRouter_ to `AnalysisRouter.js`.
Since we do not reload page between  `library folder` and `manage permissions` anymore, it improves surfing speed dramatically. 

Also re-implementation of permission folder legacy backbone code. PR makes permissions view slightly more modern. It replaces paginated select2 with `vue-multiselect`, might be a good example how to replace paginated select2.

before: 
image

now: 
image

PR #11347, opened 2021-02-11
Dataset lib permissions Vue rewrite

OlegZharkov:
Continuation of https://github.com/galaxyproject/galaxy/pull/11259
This PR vuefies dataset library permissions and refreshes the look, also another entry in the router

old:
image

now:
image


jmchilton (approved):

dannon:
I think we can get rid of some of the color here in a future iteration -- those info/warning alertboxes should always be used sparingly on the page.  But that's not new.  This looks good to me.

PR #11353, opened 2021-02-12
[21.01] Return None on unknown metadata attribute access

mvdbeek:
The AttributeError was added in https://github.com/galaxyproject/galaxy/pull/11293, but I think `if data.metadata.anyAttribute:` is something tool authors probably use.

### DIFF
--- a/lib/galaxy/model/metadata.py
+++ b/lib/galaxy/model/metadata.py
@@ -113,7 +113,7 @@ class MetadataCollection(Mapping):
             return self.spec[name].wrap(self.spec[name].default, object_session(self.parent))
         if name in self.parent._metadata:
             return self.parent._metadata[name]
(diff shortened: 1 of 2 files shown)


lib/galaxy/model/metadata.py
```diff
@@ -113,7 +113,7 @@ def __getattr__(self, name):
             return self.spec[name].wrap(self.spec[name].default, object_session(self.parent))
         if name in self.parent._metadata:
             return self.parent._metadata[name]
-        raise AttributeError
+        return None
```
nsoranzo, line 117:
```suggestion
        # Instead of raising an AttributeError for non-existing metadata, we return None
        return None
```

nsoranzo (approved):

PR #11356, opened 2021-02-12
[20.09] Move startup test to github workflow

mvdbeek:
This times out a lot on circle-ci recently.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,17 +100,6 @@ jobs:
       - *install_tox
       - *install_ffprobe
       - run: tox -e py35-unit
-  py35_first_startup:
-    docker:
-      - image: circleci/python:3.5
-    <<: *set_workdir
-    steps:
-      - *restore_repo_cache
-      # Use this job to test the latest migrations
-      - run: wget -q https://github.com/jmchilton/galaxy-downloads/raw/master/db_gx_rev_0141.sqlite
-      - run: mv db_gx_rev_0141.sqlite database/universe.sqlite
-      - *install_tox
-      - run: tox -e py35-first_startup
   validate_test_tools:
     docker:
       - image: circleci/python:3.5
@@ -162,8 +151,6 @@ workflows:
       - py35_docstring:
           <<: *requires_get_code
       - py35_unit:
-          <<: *requires_get_code
-      - py35_first_startup:
           <<: *requires_get_code
       - test_galaxy_packages:
           <<: *requires_get_code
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -100,17 +100,6 @@ jobs:
       - *install_tox
       - *install_ffprobe
       - run: tox -e py35-unit
-  py35_first_startup:
```
nsoranzo, line 103: You need to remove this also from the workflow jobs at lines 155-156.

.github/workflows/first_startup.yaml
```diff
@@ -0,0 +1,36 @@
+name: first startup
+on: [push, pull_request]
+jobs:
+
+  test:
+    name: Startup test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.5']
```
nsoranzo, line 10: Need to change this to 3.6 when merging forward.

.github/workflows/first_startup.yaml
```diff
@@ -0,0 +1,36 @@
+name: first startup
+on: [push, pull_request]
+jobs:
+
+  test:
+    name: Startup test
+    runs-on: ubuntu-latest
```
nsoranzo, line 7: Does it make sense to merge this with the osx startup tests and use the matrix to test both OSs?

mvdbeek: Then they're connected and you can't restart them individually ... we could ?

nsoranzo: You cannot restart them individually (I'd call that a bug of GH workflows UI), but you can at least have the failure of one not stop the other with `fail-fast: false` in `strategy:` (see e.g. `.github/workflows/integration.yaml` ).

.github/workflows/first_startup.yaml
```diff
@@ -0,0 +1,36 @@
...
+        path: 'galaxy root'
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Cache pip dir
+      uses: actions/cache@v1
+      id: pip-cache
+      with:
+        path: ~/Library/Caches/pip
```
nsoranzo: This is a osx thing.

.github/workflows/first_startup.yaml
mvdbeek: I am very divided on that, I preferred fail-fast: true for the connected jobs (=integration tests) since you have to re-run them anyway in most cases

nsoranzo: On the other hand, we test Python-on-conda in the osx wf and migrations on Linux, so I'm fine keeping them separated.

nsoranzo (approved):

PR #9873, opened 2020-06-12
Show favourited tools independent of version

juleengraham:
Addresses #9623

PR #11360, opened 2021-02-12
[21.01] Toolbox search fixes

dannon:
This needs refactoring, but I'm trying to get a minimal fix in for the tool search issues currently seen on main.  The other option is backing out #9873

### DIFF
--- a/client/src/components/Panels/Common/ToolSearch.vue
+++ b/client/src/components/Panels/Common/ToolSearch.vue
@@ -16,11 +16,11 @@ export default {
     props: {
         query: {
             type: String,
-            default: ''
+            default: "",
         },
         placeholder: {
             type: String,
-            default: "search tools"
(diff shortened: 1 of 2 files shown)


mvdbeek (approved):
had to add
```
diff --git a/client/src/components/Panels/utilities.js b/client/src/components/Panels/utilities.js
index 7b2da177d8..555c16593f 100644
--- a/client/src/components/Panels/utilities.js
+++ b/client/src/components/Panels/utilities.js
@@ -79,7 +79,7 @@ function normalize_results(results) {
     results.forEach((result) => {
         norm_results.push(result);
         if (result.includes("/repos/")) {
-            norm_results.push(result.split("/")[3]);
+            norm_results.push(result.split('/repos/')[1].split('/')[2]);
         }
     });
     return norm_results;
```
locally and then the results were correct

PR #11366, opened 2021-02-13
[21.01] Revert sqlalachemy-json, add unit tests, fix 'MutationList' object has no attribute '_key'

mvdbeek:
Unfortunately there are still some places left that use non-dict/non-list values on JSONType columns, and that crashes currently.

We could fix those, but I'm not super-comfortable with making everything a `SimpleJSON` type column, even if we probably only need to keep the `_metadata` column as a nested mutable JSON column. So this reverts custom_types.py to the state in 20.09.
https://github.com/galaxyproject/galaxy/pull/11366/commits/66a5f4414c923d34bc303c200547c158ec8174ce then fixes `AttributeError: 'MutationList' object has no attribute '_key'` that I tried to fix by moving the sqlalchemy-json/sqlalchemy-mutable.
https://github.com/galaxyproject/galaxy/pull/11366/commits/af47800c4969914d6427728aa3160ee277a26b2d adds some unit tests that I used to figure out what is actually currently possible with our JSONType column (which is less than I was assuming ...)


mvdbeek:
The API test actually passed (`=========== 587 passed, 4 skipped, 6 warnings in 5563.30s (1:32:43) ============
`) but something's going wrong on shutdown

jmchilton (approved):

PR #11367, opened 2021-02-14
[20.09] Fix Mtx unbound error

mvdbeek:
If the dataset is empty ... not a big issue, but there was some refactoring in 21.01 where this became an actual bug. Will open another PR for that one.


mvdbeek:
Not really worth fixing now that we have the 21.01 bugfix in

PR #11369, opened 2021-02-14
[21.01] Fix setCurrentHistoryId in legacy history panel context

mvdbeek:
If we create a new history in the legacy history the store does not get
updated, so
https://github.com/galaxyproject/galaxy/blame/3f47effa087f4049132da65643396d58a2be51d1/client/src/components/History/model/historyStore.js#L48
will always return the most recently used history.
This is a little silly in that we will load the history twice, but we
don't switch the history so often that it should matter.

If you want to test this, open the upload modal, close it, and switch to a new history, then upload a file. The dataset will appear in the old history. This fixes that.

### DIFF
--- a/client/src/components/History/model/syncCurrentHistoryToGalaxy.js
+++ b/client/src/components/History/model/syncCurrentHistoryToGalaxy.js
@@ -30,7 +30,7 @@ export function syncCurrentHistoryToGalaxy(galaxy$, store, cfg = {}) {
     );
 
     return historyId$.subscribe(
-        (id) => store.commit("betaHistory/setCurrentHistoryId", id),
+        (id) => store.dispatch("betaHistory/setCurrentHistoryId", id),
         (err) => console.warn("syncCurrentHistoryToGalaxy error", err),
         () => console.log("syncCurrentHistoryToGalaxy complete")
     );


Nerdinacan:
Looks good to me, but let me check it out and play with a bit, see if I can find any hassles.

Nerdinacan:
This looks good, but I think we also want to remove the conditional gate in syncVuexToGalaxy because you will want the store monitor to work regardless of whether the beta panel is open or not, now that we are using the history store (via UserHistories) outside of the beta panel..

Nerdinacan:
https://github.com/mvdbeek/galaxy/blob/41dbd788a9dd4570efdfb9325ea4e77f9669f090/client/src/store/syncVuextoGalaxy.js#L30-L32

mvdbeek:
But I think that works correctly, if the beta history panel is open it sets the history on switch, and if it's not open it passes through the dispatch. I don't think there's another possibility ?

Nerdinacan:
Well what if something elsewhere in the legacy code accesses Galaxy.currentHistory to change the current history from outside of the beta panel interface? I thought that was your use case?

Nerdinacan:
I think maybe the list of histories might do something like this?
historylist

Here the old list changes the history with legacy code. I don't know how many other places might do the same, which is why I built the sync function.

mvdbeek:
Yeah, that's just what this fixes, except that switching histories in this legacy view also reloads the whole history panel, so it also worked without this fix. Switching the history even from legacy components nicely passes through the listener you wrote. 
I'm inclined to think about it if we notice this to be a problem and let this be a "future us" problem

Nerdinacan:
I'm sold.

dannon (approved):

PR #11373, opened 2021-02-15
Fix TypeError in WorkflowsAPIController.get_versions()

rikeshi:
Small fix for error

`
galaxy.web.framework.decorators ERROR 2021-02-15 10:36:14,965 [p:61509,w:1,m:0] [uWSGIWorker1Core2] Uncaught exception in exposed API method:
  File "lib/galaxy/web/framework/decorators.py", line 309, in decorator
  File "lib/galaxy/webapps/galaxy/api/workflows.py", line 262, in show_versions
    stored_workflow = self.workflow_manager.get_stored_accessible_workflow(trans, workflow_id, **kwds)
TypeError: get_stored_accessible_workflow() got an unexpected keyword argument 'key'
`

when making a GET request to

`/api/workflows/{encoded_workflow_id}/versions`

### DIFF
--- a/lib/galaxy/webapps/galaxy/api/workflows.py
+++ b/lib/galaxy/webapps/galaxy/api/workflows.py
@@ -258,7 +258,8 @@ class WorkflowsAPIController(BaseAPIController, UsesStoredWorkflowMixin, UsesAnn
 
         Lists all versions of this workflow.
         """
-        stored_workflow = self.workflow_manager.get_stored_accessible_workflow(trans, workflow_id)
+        instance = util.string_as_bool(kwds.get("instance", "false"))
+        stored_workflow = self.workflow_manager.get_stored_accessible_workflow(trans, workflow_id, by_stored_id=not instance)
         return [{'version': i, 'update_time': str(w.update_time), 'steps': len(w.steps)} for i, w in enumerate(reversed(stored_workflow.workflows))]
 
     @expose_api
--- a/lib/galaxy/webapps/galaxy/api/workflows.py
+++ b/lib/galaxy/webapps/galaxy/api/workflows.py
@@ -258,7 +258,7 @@ class WorkflowsAPIController(BaseAPIController, UsesStoredWorkflowMixin, UsesAnn
 
         Lists all versions of this workflow.
         """
-        stored_workflow = self.workflow_manager.get_stored_accessible_workflow(trans, workflow_id, **kwds)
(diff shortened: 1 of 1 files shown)


bgruening:
@rikeshi this seems to be a bug, can you target a stable Galaxy release. 21.01 for example or even 20.09 I the bug is also there?

rikeshi:
It exists back in 20.01, should i target that version with a new PR then? 19.09 does not have the bug.

davelopez:
Hi @rikeshi, welcome, and thanks for opening a PR!
Can you please elaborate a bit on the context of how did you get the error and help me reproduce it?
I tried to create a new simple workflow and also create a couple of versions of it but I didn't get any error when calling `/api/workflows/{encoded_workflow_id}/versions` neither in `dev` nor in `21.01`. Maybe some particularity of the workflow itself?

mvdbeek:
This happens if you set `key` in the request payload. See https://github.com/galaxyproject/galaxy/issues/11375

lib/galaxy/webapps/galaxy/api/workflows.py
```diff
@@ -258,7 +258,7 @@ def show_versions(self, trans: GalaxyWebTransaction, workflow_id, **kwds):
 
         Lists all versions of this workflow.
         """
-        stored_workflow = self.workflow_manager.get_stored_accessible_workflow(trans, workflow_id, **kwds)
+        stored_workflow = self.workflow_manager.get_stored_accessible_workflow(trans, workflow_id)
```
simonbray:
```suggestion
        instance = util.string_as_bool(kwds.get("instance", "false"))
        stored_workflow = self.workflow_manager.get_stored_accessible_workflow(trans, workflow_id, by_stored_id=not instance)
```

simonbray: I think this would be correct.

rikeshi: Is this change still relevant now that this should be fixed in bioblend?

simonbray: I think it is, the docstring documents an `instance` param, but this doesn't work currently. And any other `kwd` will also cause an error.

rikeshi: Okay, thanks for clarifying

mvdbeek:
Thank you @rikeshi, I will backport this to our supported releases.

PR #11377, opened 2021-02-15
unset dataset size needs to be checked against None

bernt-matthias:


lib/galaxy/model/__init__.py
```diff
@@ -2542,7 +2542,7 @@ def set_size(self, no_extra_files=False):
         calls to get_total_size or set_total_size - potentially avoiding both a database flush and check against
         the file system.
         """
-        if not self.file_size:
+        if self.file_size is not None:
```
mvdbeek, line 2685: I think this was intentional as the file size starts at 0 on job creation and needs to be updated after that. I think this would be a problem, but it's odd that no test failed.

bernt-matthias: For background: I was working on the `DatasetEmptyValidator` which uses `get_size()`. For the test it was easiest to just set `self.file_size = 0`. Without the change in this PR the size calculation starts and therefor a real data set would be needed which seemed unnecessary difficult in the doc test. 

> I think this was intentional as the file size starts at 0 on job creation and needs to be updated after that.

But, it could as well start with `None` or another constant (`-1` or `UNSET`) that would indicate that the size is not yet known. "Problem" with `0` is that for empty data sets size recalculation is started each time.

Also note that there is already one place where we check against `None`:

https://github.com/galaxyproject/galaxy/blob/595b85d0b8be945f2e0f885f709f16e384447694/lib/galaxy/model/__init__.py#L2560

I guess its again more complicated .. :)

mvdbeek: The dataset paths currently get pre-created, so the first time this will be called it will be set to 0.
I like the idea of setting it to unset though while the dataset state is not terminal, that would be the ideal solution, I think.

bernt-matthias: > I like the idea of setting it to unset though while the dataset state is not termina

Is your suggestion to add a test like `state in terminal_states` and return None if True

> The dataset paths currently get pre-created

Where in the code is this happening? You mean this happens before the job stated?

The `init` function uses a default `file_size=None`. The only place where this might be different is in `DatasetManager.create()` because it uses `kwargs` and I do not know if `file_size` might be in there.

For the use of the `get_size` function: This seems to happen after jobs finish.

- lib/galaxy/managers/hdas.py:
  - in `HDAManager.copy()` .. should be fine. Seems odd: why is it necessary to calculate the size for a copy .. its the same, isn't it?
- lib/galaxy/metadata/set_metadata.py:
    - in `set_metadata_portable` .. I guess this is also called after the job finishes
- lib/galaxy/webapps/galaxy/controllers/root.py:
    - in `RootController.history_add_to seems` fine, since the file is 'finished'
- lib/galaxy/job_execution/output_collect.py:
    - in `collect_extra_files` this should also be after job execution
- lib/galaxy/datatypes/registry.py:
    - in `Registry.change_datatype()` I assume that data type changes should only possible after the job finishes (even if the UI allows to trigger manual changes before the job finishes)
- lib/galaxy/model/__init__.py: should both be fine
- lib/galaxy/model/store/discover.py:
    - `ModelPersistenceContext.create_dataset`  when Move data from temp location to dataset location
    - `ModelPersistenceContext.update_object_store_with_datasets`
- lib/galaxy/jobs/__init__.py:
    - `JobWrapper._finish_dataset` and `JobWrapper.fail`  should both be fine

mvdbeek: All the copy functions can run at any point in the datasets' life cycle.
I don't know how the fix would look like.

mvdbeek: You can actually see this going wrong in one if the integration tests:
```
galaxy.metadata DEBUG 2021-02-17 10:16:33,017 setting metadata externally failed for HistoryDatasetAssociation 3: Traceback (most recent call last):
  File "/home/runner/work/galaxy/galaxy/galaxy root/lib/galaxy/metadata/set_metadata.py", line 238, in set_metadata_portable
    set_meta(dataset, file_dict)
  File "/home/runner/work/galaxy/galaxy/galaxy root/lib/galaxy/metadata/set_metadata.py", line 132, in set_meta
    set_meta_with_tool_provided(new_dataset_instance, file_dict, set_meta_kwds, datatypes_registry, max_metadata_value_size)
  File "/home/runner/work/galaxy/galaxy/galaxy root/lib/galaxy/metadata/set_metadata.py", line 96, in set_meta_with_tool_provided
    dataset_instance.datatype.set_meta(dataset_instance, **set_meta_kwds)
  File "/home/runner/work/galaxy/galaxy/galaxy root/lib/galaxy/datatypes/proteomics.py", line 553, in set_meta
    if tpe > 0:
UnboundLocalError: local variable 'tpe' referenced before assignment
```
Apart from a possible unbound error in this datatype this happens because `has_data()` doesn't update the size now.

bernt-matthias:
I found a way in https://github.com/galaxyproject/galaxy/pull/11043 to not require this change. Not sure if we still want/need this PR.

PR #11380, opened 2021-02-15
fix profile logic for configfile generation

bernt-matthias:
follow up fixes for

https://github.com/galaxyproject/galaxy/pull/9279/files
https://github.com/galaxyproject/galaxy/pull/9776/files

- wrong profile for the profile test tool (it used 20.01, but should
have used 20.05)
- the non-profile test tool used wrong assumptions. now fixed
  and also tested against 19.09
- in the profile version `value_wrapper.value` needs to be casted
  for `multiple==False` otherwise we gat `"None"` (which I do not
  understand 100%)

I guess we should backport this a few releases .. 20.01?


mvdbeek:
What exactly is the bug you're fixing here ? This is really difficult to understand unfortunately.

lib/galaxy/tools/parameters/wrapped_json.py
```diff
@@ -100,10 +100,13 @@ def _data_input_to_path(v):
         else:
             json_value = _cast_if_not_none(value_wrapper, bool, empty_to_none=input.optional)
     elif input_type == "select":
-        if input.multiple and packaging.version.parse(str(profile)) >= packaging.version.parse('20.05'):
-            json_value = [_ for _ in _cast_if_not_none(value_wrapper.value, list)]
-        else:
+        if packaging.version.parse(str(profile)) < packaging.version.parse('20.05'):
```
mvdbeek, line 103: This is just a longer version of what is already here, no ?
I'm really bad at this logic stuff, but if I rewrite it like this it still looks the same to me.

The new version is
```python
legacy_profile = packaging.version.parse(str(profile)) < packaging.version.parse('20.05')
if legacy_profile:
    json_value = _cast_if_not_none(value_wrapper, str)
else:
    if input.multiple:
        json_value = [str(_) for _ in _cast_if_not_none(value_wrapper.value, list)]
    else:
        json_value = _cast_if_not_none(value_wrapper.value, str)
```
While the old version is 
```python
if input.multiple and not legacy_profile:
     json_value = [str(_) for _ in _cast_if_not_none(value_wrapper.value, list)]
else:
    json_value = _cast_if_not_none(value_wrapper.value, str)
```
In both cases you can only get to the list comprehension when `input.multiple and not legacy_profile`, or am I missing something here ?

mvdbeek: I see, I focused at the wrong part of this, in the list comprehension you're casting to string ... 

Which seems like it would not be the right thing to do, unless this is needed in legacy tools.

bernt-matthias: The difference is that in the old version `value_wrapper.value` is casted (in the else branch) while in the new version `value_wrapper` is casted in the `legacy_profile` version. 

One can also see in the test files that the behavior was wrong: 

- In the legacy profile version: we had `None`, `[]` and `"a"`, `['a', 'b']` which is not the legacy behavior (I tested with 19.09)
- In the new profile version we had the opposite

So the two cases were actually switched (so at least the test cases are now correct).

Still you might be right that the logic was correct, but the bug is somewhere else, e.g. in `_cast_if_not_none`. I was also wondering about this line: https://github.com/galaxyproject/galaxy/blob/71dabed4170e392a009735d41676baf51a75fc3c/lib/galaxy/tools/parameters/wrapped_json.py#L89 which was recently added...

bernt-matthias: >  see, I focused at the wrong part of this, in the list comprehension you're casting to string ...

I think this cast is actually not strictly necessary.

mvdbeek: Hmm, I agree that the cases are switched ... but they probably have been switched since 20.05, so flipping this now could cause formerly working tools to fail. But this is only for json input, which is not such a common consumer, so maybe we just do it now ? Can you maybe refactor the if/else back to the original, which seemed slightly easier to digest to me ?

mvdbeek (approved):

lib/galaxy/tools/parameters/wrapped_json.py
bernt-matthias: > Can you maybe refactor the if/else back to the original, which seemed slightly easier to digest to me ?

I tried:

`inputs_as_json_profile` fails for:

```
        if input.multiple and packaging.version.parse(str(profile)) >= packaging.version.parse('20.05'):
            json_value = [_ for _ in _cast_if_not_none(value_wrapper.value, list)]
        else:
            json_value = _cast_if_not_none(value_wrapper, str)
```

and `inputs_as_json` fails for:

```
        if input.multiple and packaging.version.parse(str(profile)) >= packaging.version.parse('20.05'):
            json_value = [_ for _ in _cast_if_not_none(value_wrapper.value, list)]
        else:
            json_value = _cast_if_not_none(value_wrapper.value, str)
```

The difference is really if `value_wrapper.value` or `value_wrapper` is casted. 

Really wondering what thos value_wrappers are and what actually happens if they are caset, ie. what happens for int(some_object)? Is this essentially `int(str(some_object))`?

test/functional/tools/inputs_as_json_profile.xml
```diff
@@ -46,7 +46,7 @@ elif test_case == "2":
     assert_equals(as_dict["floattest"], 1.0)
     assert_equals(as_dict["radio_select"], "a_radio")
     assert_equals(as_dict["optional_select"], "a")
-    assert_equals(as_dict["optional_multiple_select"], 'a,b')
+    assert_equals(as_dict["optional_multiple_select"], ['a', 'b'])
```
jmchilton, line 49: Thanks for the fixes!

jmchilton: If this has been broken since 20.05 - should we just change the profile version we check to be 21.05?

bernt-matthias: > If this has been broken since 20.05 - should we just change the profile version we check to be 21.05?

If I'm not wrong:

- we introduced the change in 20.01
- then we decided in 20.05 to "hide" the change in a profile version .. and apparently I messed up that change :( 

My preference would be to backport it down to 20.01 .. then we have the documented behavior in all versions from 20.01. ... would be also good because the OpenMS tool suite depends on the correct export of `None`.

bernt-matthias: I created PRs for 20.05 to 20.01 .. because I was wrong about that the changes started in 20.01 https://github.com/galaxyproject/galaxy/blob/release_20.01/lib/galaxy/tools/parameters/wrapped_json.py

PR #11383, opened 2021-02-16
[21.05] ITs on Kubernetes

almahmoud:
Enabling running ITs with Service and Ingress without reverse proxy when Galaxy is running on k8s. Started with k8s runner cause more familiar with it but plan is to port to pulsar. Still very WIP, but the basics are there:
working_its


pcm32:
Great! So, this is not relying then on the GIE proxy mechanism to re-route things?

almahmoud:
Nope, no proxy mechanism, just requires an ingress controller in the cluster. Relies on the job runner to create and cleanup the Service/Ingress pair along with the IT job

lib/galaxy/config/__init__.py
```diff
@@ -817,6 +817,7 @@ def _process_config(self, kwargs):
         self.interactivetools_map = self._in_root_dir(kwargs.get("interactivetools_map", self._in_data_dir("interactivetools_map.sqlite")))
         self.interactivetools_prefix = kwargs.get("interactivetools_prefix", "interactivetool")
         self.interactivetools_proxy_host = kwargs.get("interactivetools_proxy_host", None)
```
jdavcs: This is not necessary: it's processed automatically

lib/galaxy/config/__init__.py
```diff
@@ -817,6 +817,7 @@ def _process_config(self, kwargs):
...
+        self.interactivetools_use_k8s_ingress = kwargs.get("interactivetools_use_k8s_ingress", False)
```
jdavcs: Could you, please, add this to the schema instead? Similar to `interactivetools_proxy_host`?

lib/galaxy/config/__init__.py
almahmoud: I didn't add them, they were here from the original IT work. If I'm understanding correctly, you're suggesting they should all be removed from here and just added to the schema? (at least `interactivetools_prefix` is not in the schema already)

lib/galaxy/config/__init__.py
almahmoud: I can. Was wondering if this should just be a runner param instead though, given that it really only works wit the k8s runner. I'll look into how easy it is to change it to that

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -136,15 +145,17 @@ def queue_job(self, job_wrapper):
...
-        job_id = job.metadata['name']
+        try:
+            job.create()
+            job_id = job.metadata.get('name')
+            # define job attributes in the AsyncronousJobState for follow-up
+            ajs.job_id = job_id
+            # store runner information for tracking if Galaxy restarts
+            job_wrapper.set_external_id(job_id)
+        except Exception as e:
+            log.debug(str(e))
```
jmchilton: You probably want to re-raise the exception right?

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -610,4 +798,18 @@ def finish_job(self, job_state):
             log.warning("More than one job matches selector. Possible configuration error"
                         " in job id '%s'", job_state.job_id)
         job = Job(self._pykube_api, jobs.response['items'][0])
+        if job_state.job_wrapper.tool.tool_type == "interactive":
```
jmchilton: I don't know for sure - but I think I was trying to use ``job_state.job_wrapper.tool.ports`` to determine if we need to expose ports instead of tool type being interactive.

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -18,18 +18,26 @@
...
+    delete_service,
     ensure_pykube,
+    find_ingress_object_by_name,
     find_job_object_by_name,
     find_pod_object_by_name,
+    find_service_object_by_name,
     galaxy_instance_id,
     is_pod_unschedulable,
+    Ingress,
+    ingress_object_dict,
```
```suggestion
    Ingress,
    ingress_object_dict,
    is_pod_unschedulable,
```
To make the linter happy.

lib/galaxy/jobs/runners/kubernetes.py
almahmoud: Done in https://github.com/galaxyproject/galaxy/pull/11383/commits/7bfa28dfa87952e941bbb4aec77c6ef63d2a2eb8

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -139,15 +148,18 @@ def queue_job(self, job_wrapper):
...
-        job_id = job.metadata['name']
+        try:
+            job.create()
+            job_id = job.metadata.get('name')
+            # define job attributes in the AsyncronousJobState for follow-up
+            ajs.job_id = job_id
+            # store runner information for tracking if Galaxy restarts
+            job_wrapper.set_external_id(job_id)
+        except Exception as e:
+            log.debug(str(e))
```
mvdbeek: If this is not supposed to happen if everything is working "normally" I would do this
```suggestion
            log.exception("Error occurred while creating kubernetes job object")
```
That'll include the traceback, and depending on the logging setup various additional things that might be helpful for debugging this further.

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -271,6 +283,101 @@ def __get_k8s_job_spec_template(self, ajs):
...
+                               "http": {
+                                   "paths": [{
+                                       "backend": {
+                                           "serviceName": "{}-{}".format(self.__produce_k8s_job_prefix(),
+                                                                         self.__force_label_conformity(ajs.job_wrapper.get_id_tag())),
+                                           "servicePort": int(ep["tool_port"])
+                                           # "service": {
+                                           #     "name": "job-{}-{}".format(self.__force_label_conformity(ajs.job_wrapper.get_id_tag()), p),
+                                           #     "port": { "number": int(p)}
+                                           # }
```
mvdbeek: Do you still need this ?
```suggestion
                                           # "service": {
                                           #     "name": "job-{}-{}".format(self.__force_label_conformity(ajs.job_wrapper.get_id_tag()), p),
                                           #     "port": { "number": int(p)}
                                           # }
```

almahmoud: Yes, forgot to change it back with the new api version. changed in https://github.com/galaxyproject/galaxy/pull/11383/commits/ec026579a5f6fafeb30c1212f52957586d484904

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -271,6 +283,99 @@ def __get_k8s_job_spec_template(self, ajs):
...
+                                                                      self.__force_label_conformity(ajs.job_wrapper.get_id_tag())),
+                                               "port": { "number": int(ep["tool_port"]) }
+                                           }
+                                       },
+                                       "path": ep.get("entry_path", '/'),
+                                       "pathType": "Prefix"
+                                   }]}} for ep in entry_points]
+                }
+            }
+            log.debug(k8s_spec_template)
```
```suggestion
            log.debug(f"Kubernetes spec template is {k8s_spec_template}")
```

mvdbeek: (or drop the debug, that's fine too)

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -271,6 +283,99 @@ def __get_k8s_job_spec_template(self, ajs):
...
+                               "targetPort": int(p)} for p in guest_ports],
+                    "selector": {
+                        "app.kubernetes.io/name": self.__force_label_conformity(ajs.job_wrapper.tool.old_id),
+                        "app.kubernetes.io/component": "tool",
+                        "app.galaxyproject.org/job_id": self.__force_label_conformity(ajs.job_wrapper.get_id_tag())
+                    },
+                    "type": "ClusterIP"
+                }
+            }
+            log.debug(k8s_spec_template)
```
mvdbeek: Same here, if you don't need it, to drop the debug, otherwise prefix this with a message that says what this is.

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -295,18 +400,18 @@ def __get_k8s_containers(self, ajs):
...
         k8s_container = {
             "name": self.__get_k8s_container_name(ajs.job_wrapper),
-            "image": self._find_container(ajs.job_wrapper).container_id,
+            "image": container.container_id,
             # this form of command overrides the entrypoint and allows multi command
             # command line execution, separated by ;, which is what Galaxy does
             # to assemble the command.
             "command": [ajs.job_wrapper.shell],
-            "args": ["-c", ajs.job_file],
+            "args": ["-c", ajs.job_file],  # "source /galaxy/server/.venv/bin/activate && {}".format(ajs.job_file)],
```
```suggestion
            "args": ["-c", ajs.job_file],
```

lib/galaxy/config/__init__.py
mvdbeek: I think that would make sense. I guess it's hypothetical, but in theory deployers might want to have more than one runner type that can run interactive tools, so that's another reason to put this in the runner parameters.

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -271,6 +283,96 @@ def __get_k8s_job_spec_template(self, ajs):
         k8s_spec_template["metadata"]["annotations"].update(extra_metadata.get('annotations', {}))
         return k8s_spec_template
 
+    def __get_k8s_service_spec(self, ajs):
+        """The k8s spec template is nothing but a Service spec, except that it is nested and does not have an apiversion
+        nor kind."""
+        container = self._find_container(ajs.job_wrapper)
+        guest_ports = container.tool_info.guest_ports
+        if not guest_ports:
```
```suggestion
        if not ajs.job_wrapper.guest_ports:
```

mvdbeek: Note that I haven't tested this, but we don't seem to need the container here.

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -271,6 +283,96 @@ def __get_k8s_job_spec_template(self, ajs):
...
+            }
+            return k8s_spec_template
+
+    def __get_k8s_ingress_spec(self, ajs):
+        """The k8s spec template is nothing but a Ingress spec, except that it is nested and does not have an apiversion
+        nor kind."""
+        container = self._find_container(ajs.job_wrapper)
+        guest_ports = container.tool_info.guest_ports
+        if not guest_ports:
+            return None
```
```suggestion
        if not ajs.job_wrapper.guest_ports:
            return None
```

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -487,6 +592,32 @@ def check_watched_item(self, job_state):
                 return None
             elif active > 0 and failed <= max_pod_retries:
                 if not job_state.running:
+                    job_state.running = True
+                    job_state.job_wrapper.change_state(model.Job.states.RUNNING)
+                    if job_state.job_wrapper.tool.tool_type == "interactive":
```
```suggestion
                    if job_state.job_wrapper.guest_ports:
```

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -487,6 +592,32 @@ def check_watched_item(self, job_state):
...
+                        k8s_ingress_obj = ingress_object_dict(
+                            self.runner_params,
+                            k8s_job_name,
+                            self.__get_k8s_ingress_spec(job_state)
+                        )
+                        service = Service(self._pykube_api, k8s_service_obj)
+                        try:
+                            service.create()
+                        except Exception as e:
+                            log.debug(f'Failed to create service k8s resource for IT due to: {str(e)}')
```
mvdbeek: Are these exception fatal for running an interactive tool ? In that case I'd re-raise the exception (or maybe not even catch them?)

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -554,32 +685,62 @@ def _handle_job_failure(self, job, job_state):
         job_state.running = False
         self.mark_as_failed(job_state)
         try:
+            if job_state.job_wrapper.tool.tool_type == "interactive":
+                k8s_job_prefix = self.__produce_k8s_job_prefix()
+                k8s_job_name = "{}-{}".format(k8s_job_prefix, self.__force_label_conformity(job_state.job_wrapper.get_id_tag()))
```
mvdbeek: You're using this pattern a couple of time ... does it make sense to move that into a utility function (get_k8s_job_name)?

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -643,4 +819,18 @@ def finish_job(self, job_state):
             log.warning("More than one job matches selector. Possible configuration error"
                         " in job id '%s'", job_state.job_id)
         job = Job(self._pykube_api, jobs.response['items'][0])
+        if job_state.job_wrapper.tool.tool_type == "interactive":
```
```suggestion
        if job_state.job_wrapper.guest_ports:
```

lib/galaxy/config/__init__.py
almahmoud: Done in https://github.com/galaxyproject/galaxy/pull/11383/commits/63510bea237eb1ef58aab8294e06e65fb3b743f3 (left default as if not added so in k8s destination have to specify: `require_monitor: False` but nothing will break)

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -603,10 +764,25 @@ def stop_job(self, job_wrapper):
             job_to_delete = find_job_object_by_name(self._pykube_api, job.get_job_runner_external_id(), self.runner_params['k8s_namespace'])
             if job_to_delete and len(job_to_delete.response['items']) > 0:
                 k8s_job = Job(self._pykube_api, job_to_delete.response['items'][0])
+                if job_wrapper.tool.tool_type == "interactive":
```
```suggestion
                if job_wrapper.guest_ports:
```

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -277,6 +289,91 @@ def __get_k8s_job_spec_template(self, ajs):
         k8s_spec_template["metadata"]["annotations"].update(extra_metadata.get('annotations', {}))
         return k8s_spec_template
 
+    def __get_k8s_service_spec(self, ajs):
+        """The k8s spec template is nothing but a Service spec, except that it is nested and does not have an apiversion
+        nor kind."""
+        if not ajs.job_wrapper.guest_ports:
```
```suggestion
        guest_ports = ajs.job_wrapper.guest_ports
        if not guest_ports:
```
I had missed that you're using them on line 313

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -277,6 +289,91 @@ def __get_k8s_job_spec_template(self, ajs):
...
+                    },
+                    "type": "ClusterIP"
+                }
+            }
+            return k8s_spec_template
+
+    def __get_k8s_ingress_spec(self, ajs):
+        """The k8s spec template is nothing but a Ingress spec, except that it is nested and does not have an apiversion
+        nor kind."""
+        if not ajs.job_wrapper.guest_ports:
```
```suggestion
        guest_ports = ajs.job_wrapper.guest_ports
        if not guest_ports:
```

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -442,10 +542,13 @@ def __get_k8s_container_name(self, job_wrapper):
             return cleaned_id
         return "job-container"
 
+    def __get_k8s_job_name(self, prefix, job_wrapper):
+      return "{}-{}".format(prefix, self.__force_label_conformity(job_wrapper.get_id_tag()))
```
```suggestion
        return "{}-{}".format(prefix, self.__force_label_conformity(job_wrapper.get_id_tag()))
```

lib/galaxy/jobs/runners/kubernetes.py
```diff
@@ -609,10 +762,25 @@ def stop_job(self, job_wrapper):
             job_to_delete = find_job_object_by_name(self._pykube_api, job.get_job_runner_external_id(), self.runner_params['k8s_namespace'])
             if job_to_delete and len(job_to_delete.response['items']) > 0:
                 k8s_job = Job(self._pykube_api, job_to_delete.response['items'][0])
+                if job_state.job_wrapper.guest_ports:
```
```suggestion
                if job_wrapper.guest_ports:
```

mvdbeek (commented):
Couple minor fixes here

mvdbeek (approved):

PR #11384, opened 2021-02-16
fix broken docs links

martenson:
This fixes the links for authentication details. That these are not accompanying the Galaxy codebase seems detrimental.


hexylena (approved):

hexylena:
Looks like an improvement for the short term, and then we come back later to migrate the whole batch? They really belong with the code, agreed.

PR #11385, opened 2021-02-16
Lint for optional or multiple selects with radio display

bernt-matthias:
select with `display="radio"` one can not be unselected once an options was chosen.
also radio style selects do not allow to select multiple options.

both might as well be considered bugs. but I think the idea of radio buttons could
as well be considered incompatible with multiple=true or optional
https://en.wikipedia.org/wiki/Radio_button

Are there already tests for this piece of code that might be extended? Or would is be
worth setting some up?

### DIFF
--- a/lib/galaxy/tool_util/linters/inputs.py
+++ b/lib/galaxy/tool_util/linters/inputs.py
@@ -45,11 +45,12 @@ def lint_inputs(tool_xml, lint_ctx):
                 message = "No options defined for select [%s]" % param_name
                 lint_ctx.warn(message)
 
-            if param_attrib["display"] == "radio":
-                if param_attrib["optional"] == "true":
-                    lint_ctx.error('Select [%s] display="radio" is incompatible with optional="true"', param_name)
-                if param_attrib["multiple"] == "true":
+            if param_attrib.get("display", "checkboxes") == "radio":
+                multiple = string_as_bool(param_attrib.get("multiple", False))
+                if multiple:
                     lint_ctx.error('Select [%s] display="radio" is incompatible with multiple="true"', param_name)
+                if string_as_bool(param_attrib.get("optional", multiple)):
+                    lint_ctx.error('Select [%s] display="radio" is incompatible with optional="true"', param_name)
         # TODO: Validate type, much more...
 
     conditional_selects = tool_xml.findall("./inputs//conditional")
--- a/lib/galaxy/tool_util/linters/inputs.py
+++ b/lib/galaxy/tool_util/linters/inputs.py
@@ -46,10 +46,9 @@ def lint_inputs(tool_xml, lint_ctx):
                 lint_ctx.warn(message)
 
             if param_attrib.get("display", "checkboxes") == "radio":
-                multiple = string_as_bool(param_attrib.get("multiple", False))
-                if multiple:
+                if string_as_bool(param_attrib.get("multiple", "false")):
                     lint_ctx.error('Select [%s] display="radio" is incompatible with multiple="true"', param_name)
-                if string_as_bool(param_attrib.get("optional", multiple)):
+                if string_as_bool(param_attrib.get("optional", "false")):
                     lint_ctx.error('Select [%s] display="radio" is incompatible with optional="true"', param_name)
         # TODO: Validate type, much more...
 
--- a/lib/galaxy/tool_util/linters/inputs.py
+++ b/lib/galaxy/tool_util/linters/inputs.py
@@ -44,6 +44,12 @@ def lint_inputs(tool_xml, lint_ctx):
             if dynamic_options is None and len(select_options) == 0:
                 message = "No options defined for select [%s]" % param_name
                 lint_ctx.warn(message)
+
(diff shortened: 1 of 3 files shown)


lib/galaxy/tool_util/linters/inputs.py
```diff
@@ -44,6 +44,12 @@ def lint_inputs(tool_xml, lint_ctx):
             if dynamic_options is None and len(select_options) == 0:
                 message = "No options defined for select [%s]" % param_name
                 lint_ctx.warn(message)
+
+            if param_attrib["display"] == "radio":
```
nsoranzo:
```suggestion
            if param_attrib.get("display", "checkboxes") == "radio":
```

lib/galaxy/tool_util/linters/inputs.py
```diff
@@ -44,6 +44,12 @@ def lint_inputs(tool_xml, lint_ctx):
...
+                if param_attrib["optional"] == "true":
+                    lint_ctx.error('Select [%s] display="radio" is incompatible with optional="true"', param_name)
+                if param_attrib["multiple"] == "true":
+                    lint_ctx.error('Select [%s] display="radio" is incompatible with multiple="true"', param_name)
```
nsoranzo:
```suggestion
                multiple = string_as_bool(param_attrib.get("multiple", False))
                if multiple:
                    lint_ctx.error('Select [%s] display="radio" is incompatible with multiple="true"', param_name)
                if string_as_bool(param_attrib.get("optional", multiple)):
                    lint_ctx.error('Select [%s] display="radio" is incompatible with optional="true"', param_name)
```

nsoranzo (commented):
These suggestions should fix also the unit tests.

lib/galaxy/tool_util/linters/inputs.py
```diff
@@ -44,6 +44,12 @@ def lint_inputs(tool_xml, lint_ctx):
             if dynamic_options is None and len(select_options) == 0:
                 message = "No options defined for select [%s]" % param_name
                 lint_ctx.warn(message)
+
+            if param_attrib.get("display", "checkboxes") == "radio":
```
nsoranzo: I had another look at the code and the default is actually `None`:
```suggestion
            if param_attrib.get("display") == "radio":
```

bernt-matthias:
Thanks for the suggestions @nsoranzo .. just found the unit test .. but having trouble to get it right.

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -2889,12 +2889,14 @@ the column index.
...
+boxes (``display="checkboxes"``) or radio buttons (``display="radio"``, note this
+is incompatible with ``multiple="true"`` and ``optional="true"``).
+Defaults to a drop-down menu select list.</xs:documentation>
           </xs:annotation>
         </xs:attribute>
         <xs:attribute name="multiple" type="PermissiveBoolean">
           <xs:annotation>
-            <xs:documentation xml:lang="en">Allow multiple valus to be selected.
+            <xs:documentation xml:lang="en">Allow multiple values to be selected.
 Valid with ``data`` and ``select`` parameters. ``select`` parameters with ``multiple="true"`` are optional by default.</xs:documentation>
```
```suggestion
            <xs:documentation xml:lang="en">Allow multiple values to be
selected. ``select`` parameters with ``multiple="true"`` are optional by
default. Used only when the ``type`` attribute value is ``data``, ``group_tag``,
or ``select``.</xs:documentation>
```

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -2889,12 +2889,14 @@ the column index.
           <xs:annotation>
             <xs:documentation xml:lang="en">This attribute is used only if
 ``type`` attribute value is ``select`` - render a select list as a set of check
```
nsoranzo: I am going to open a separate PR to standardise a bit these attribute descriptions, adding my changes for these ones here so we don't have conflicts.

```suggestion
            <xs:documentation xml:lang="en">Render a select list as a set of
checkboxes (``checkboxes``) or radio buttons (``radio``; note this is
```

lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -2889,12 +2889,14 @@ the column index.
...
-boxes or radio buttons. Defaults to a drop-down menu select list.</xs:documentation>
+boxes (``display="checkboxes"``) or radio buttons (``display="radio"``, note this
+is incompatible with ``multiple="true"`` and ``optional="true"``).
+Defaults to a drop-down menu select list.</xs:documentation>
```
```suggestion
incompatible with ``multiple="true"`` and ``optional="true"``). Defaults to a
drop-down menu select list. Used only when the ``type`` attribute value is
``select``.</xs:documentation>
```

lib/galaxy/tool_util/xsd/galaxy.xsd
nsoranzo: The mentioned PR is https://github.com/galaxyproject/galaxy/pull/11390 , if you want to take a look!

nsoranzo (approved):

PR #11386, opened 2021-02-16
Refactor display_applications API logic into DisplayApplicationsManager

davelopez:
## Overview
This refactoring is in preparation for the FastAPI migration.

- Added some API tests.
- Replaced the old legacy_expose_api decorators. See #11252. 
- Moved the controller logic to a manager class and added basic type hints.


jmchilton:
Awesome - thanks!

PR #11387, opened 2021-02-16
[21.01] Add tracking code to webhook for GTN stats

hexylena:


mvdbeek:
Do you want to target 21.01 ? I guess there was a recent change that is conflicting here?

hexylena:
Sure, can do. Not a huge priority but that fixes the conflict so, sounds good to me :) Need to double check that this works and then I can open for review.

dannon (approved):

hexylena:
Sorry for the delay, just wanted to be sure the proxy didn't need an update/new documentation. Happy with this now. It seems to be passing the ?s through nginx (despite what people say on the internet?) without changes.

PR #11389, opened 2021-02-16
[21.01] Fix rule-based screen-shooting

OlegZharkov:
a few fixes for https://github.com/galaxyproject/training-material/pull/2271
Fixing issues mentioned in https://github.com/galaxyproject/training-material/pull/2271#issuecomment-776794531.
Thanks a lot @hexylena for your review! It was quite helpful.

### DIFF
--- a/lib/galaxy/selenium/navigation.yml
+++ b/lib/galaxy/selenium/navigation.yml
@@ -583,6 +583,8 @@ upload:
       selector: '.selection-dialog-modal'
     rule_dataset_selector_row:
       selector: '.selection-dialog-modal [aria-rowindex="${rowindex}"]'
+    build_btn: '#rule-based #btn-build'
+    tooltip: '.b-tooltip'
 
 rule_builder:
   selectors:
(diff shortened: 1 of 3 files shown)


lib/galaxy/selenium/navigation.yml
```diff
@@ -583,6 +583,8 @@ upload:
       selector: '.selection-dialog-modal'
     rule_dataset_selector_row:
       selector: '.selection-dialog-modal [aria-rowindex="${rowindex}"]'
+    build_btn: '#rule-based #btn-build'
+    tooltip: '.b-tooltip'
```
mvdbeek: Selenium noob here, I didn't see where this is used here ? If it's not used we should remove it.

mvdbeek (approved):
Thank you @OlegZharkov !

lib/galaxy/selenium/navigation.yml
OlegZharkov: yes, you're right. I forgot to remove this `tooltip`

OlegZharkov:
@mvdbeek thanks!

PR #11390, opened 2021-02-16
Standardise tool param attribute docs

nsoranzo:
Also:
- Move `load_contents` to `data` param section.
- Move attribute list for `color` params to correct position.
- The `area` attribute is used only for `text` params.
- More correct explanation of default value of the ``optional`` param attribute.
- The `checked` attribute is used only for `boolean` params.
- Parse `load_contents` only in `DataToolParameter`.

### DIFF
--- a/lib/galaxy/tool_util/xsd/galaxy.xsd
+++ b/lib/galaxy/tool_util/xsd/galaxy.xsd
@@ -2508,7 +2508,7 @@ parameter being described. All the attributes for the ``param`` element are
 documented below for completeness, but here are the common ones for each
 type are as follows:
 
-$attribute_list:name,type,optional,label,help,argument,load_contents,refresh_on_change:4
+$attribute_list:name,type,optional,label,help,argument,refresh_on_change:4
 
 ### Parameter Types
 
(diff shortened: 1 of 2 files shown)


lib/galaxy/tool_util/xsd/galaxy.xsd
```diff
@@ -2803,57 +2803,56 @@ parameter.</xs:documentation>
...
         </xs:attribute>
         <xs:attribute name="data_ref" type="xs:string">
           <xs:annotation>
             <xs:documentation xml:lang="en"><![CDATA[
-
-Only valid if ``type`` attribute value is ``select``, ``data_column``, or
-``group_tag``. Used with select lists whose options are dynamically generated
+Used with select lists whose options are dynamically generated
 based on certain metadata attributes of the dataset or collection upon which
 this parameter depends (usually but not always the tool's input dataset).
```
bernt-matthias, line 2853: Wondering what this means: "(usually but not always the tool's input dataset)"?

bernt-matthias (approved):

lib/galaxy/tool_util/xsd/galaxy.xsd
nsoranzo: Not sure, to be honest. This text was originally added by @jmchilton in commit b059c8b592b0f45dd669f8e9cbca714ed292ccac .

jmchilton: I really don't know, sorry.

PR #11393, opened 2021-02-17
Read cgroups from cgroup filesystem mount rather than cgget, which is not always installed

natefoo:
The shell for these comes out to the following (reformatted for readability) and extra eyes here would be appreciated. For cpu and cpuacct, because they are comounted, I found that depending on the system cpu and cpuacct appeared in different orders (and even differed between their order in `/proc/$$/cgroup` and the entry in `/sys/fs/cgroup`:

```bash
if [ -e "/proc/$$/cgroup" -a -d "/sys/fs/cgroup" ]; then
    cgroup_path=$(cat "/proc/$$/cgroup" | awk -F':' '($2=="cpuacct,cpu")||($2=="cpu,cpuacct"){print $3}');
    for f in /sys/fs/cgroup/{cpu\,cpuacct,cpuacct\,cpu}$cgroup_path/{cpu,cpuacct}.*; do
        if [ -f "$f" ]; then
            echo "__$(basename $f)__" >> /tmp/foo; cat "$f" >> /tmp/foo 2>/dev/null;
        fi;
    done;
fi
```

Memory is thankfully simpler:

```bash
if [ -e "/proc/$$/cgroup" -a -d "/sys/fs/cgroup" ]; then
    cgroup_path=$(cat "/proc/$$/cgroup" | awk -F':' '$2=="memory"{print $3}');
    for f in /sys/fs/cgroup/memory$cgroup_path/memory.*; do
        echo "__$(basename $f)__" >> /tmp/foo; cat "$f" >> /tmp/foo 2>/dev/null;
    done;
fi
```

This will break metrics for any jobs running during upgrade, but that's not worth accounting for (the jobs will complete fine, they just won't have cgroup metrics, which plenty of jobs are already missing).

### DIFF
--- a/lib/galaxy/job_metrics/instrumenters/cgroup.py
+++ b/lib/galaxy/job_metrics/instrumenters/cgroup.py
@@ -25,6 +25,24 @@ CONVERSION = {
     "memory.oom_control.under_oom": lambda x: "Yes" if x == 1 else "No",
     "cpuacct.usage": lambda x: formatting.seconds_to_str(x / 10**9)  # convert nanoseconds
 }
+CPU_USAGE_TEMPLATE = """
+if [ -e "/proc/$$/cgroup" -a -d "{cgroup_mount}" ]; then
+    cgroup_path=$(cat "/proc/$$/cgroup" | awk -F':' '($2=="cpuacct,cpu") || ($2=="cpu,cpuacct") {{print $3}}');
+    for f in {cgroup_mount}/{{cpu\,cpuacct,cpuacct\,cpu}}$cgroup_path/{{cpu,cpuacct}}.*; do
+        if [ -f "$f" ]; then
+            echo "__$(basename $f)__" >> {metrics}; cat "$f" >> {metrics} 2>/dev/null;
+        fi;
+    done;
+fi
+""".replace("\n", " ").strip()  # noqa: W605
+MEMORY_USAGE_TEMPLATE = """
+if [ -e "/proc/$$/cgroup" -a -d "{cgroup_mount}" ]; then
+    cgroup_path=$(cat "/proc/$$/cgroup" | awk -F':' '$2=="memory"{{print $3}}');
+    for f in {cgroup_mount}/memory$cgroup_path/memory.*; do
+        echo "__$(basename $f)__" >> {metrics}; cat "$f" >> {metrics} 2>/dev/null;
+    done;
+fi
+""".replace("\n", " ").strip()
 
 
 Metric = namedtuple("Metric", ("key", "subkey", "value"))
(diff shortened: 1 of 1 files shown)


jmchilton (approved):

hexylena (approved):
looks fine to me. nice to get rid of the cgtools dep.

lib/galaxy/job_metrics/instrumenters/cgroup.py
```diff
@@ -72,64 +73,49 @@ def job_properties(self, job_id, job_directory):
         return metrics
 
     def __record_cgroup_cpu_usage(self, job_directory):
-        return """if [ `command -v cgget` ] && [ -e /proc/$$/cgroup ]; then cat /proc/$$/cgroup | awk -F':' '$2=="cpuacct,cpu"{{print $2":"$3}}' | xargs -I{{}} cgget -g {{}} > {metrics} ; else echo "" > {metrics}; fi""".format(metrics=self.__cgroup_metrics_file(job_directory))
+        # comounted cgroups (which cpu and cpuacct are on the supported Linux distros) can appear in any order (cpu,cpuacct or cpuacct,cpu)
+        return """if [ -e "/proc/$$/cgroup" -a -d "{cgroup_mount}" ]; then cgroup_path=$(cat "/proc/$$/cgroup" | awk -F':' '($2=="cpuacct,cpu")||($2=="cpu,cpuacct"){{print $3}}'); for f in {cgroup_mount}/{{cpu\,cpuacct,cpuacct\,cpu}}$cgroup_path/{{cpu,cpuacct}}.*; do if [ -f "$f" ]; then echo "__$(basename $f)__" >> {metrics}; cat "$f" >> {metrics} 2>/dev/null; fi; done; fi""".format(metrics=self.__cgroup_metrics_file(job_directory), cgroup_mount=self.cgroup_mount)  # noqa: W605
```
hexylena: what about just replacing the newlines for more readable code here?

```suggestion
        return """
        	if [ -e "/proc/$$/cgroup" -a -d "{cgroup_mount}" ]; then
        		cgroup_path=$(cat "/proc/$$/cgroup" | awk -F':' '($2=="cpuacct,cpu")||($2=="cpu,cpuacct"){{print $3}}'); 
        		for f in {cgroup_mount}/{{cpu\,cpuacct,cpuacct\,cpu}}$cgroup_path/{{cpu,cpuacct}}.*; do
        			if [ -f "$f" ]; then
        				echo "__$(basename $f)__" >> {metrics};
        				cat "$f" >> {metrics} 2>/dev/null;
        			fi; 
        		done; 
        	fi""".replace('\n', '').format(metrics=self.__cgroup_metrics_file(job_directory), cgroup_mount=self.cgroup_mount)  # noqa: W605
```

I know it doesn't change often but dunno, maybe readability is nice

natefoo: I honestly don't know why we tried to minify these for the job script. @jmchilton do you know?

natefoo:
@hexylena how about 652c7b8?

lib/galaxy/job_metrics/instrumenters/cgroup.py
natefoo: Oh I didn't notice your `.replace('\n', '')` in there which is what I ended up doing anyway in my version lol.

hexylena:
Don't even know if replacing the newlines is necessary? But yeah, looks
like what I was thinking, much easier to read the next time we need to edit
it!

Nate Coraor <notifications@github.com> schrieb am Mi., 17. Feb. 2021, 20:18:

natefoo:
> Don't even know if replacing the newlines is necessary?

I don't know either, I suspect it was just doing what other, less complicated metrics were doing.

natefoo:
I should also mention: the next time the `galaxy-job-metrics` is released (once I verify this works well with Pulsar destinations on test.galaxyproject.org I'll PR an update) then if you `pip install --upgrade galaxy-job-metrics` in your Pulsar venvs, Pulsar will produce cgroup metrics with the new method/format. Pulsar doesn't pin dependencies so this won't happen for you automatically.

PR #11400, opened 2021-02-17
Run maintainance bot on ready_for_review and edit

mvdbeek:


.github/workflows/maintenance_bot.yaml
```diff
@@ -1,7 +1,7 @@
 name: "Maintenance Bot"
 on:
   pull_request_target:
-    types: [opened, reopened]
+    types: [opened, reopened, edited, ready_for_review]
```
nsoranzo: Do we also need `unlabeled` for when the `status/WIP` label is removed? Not sure what `edited` covers.

mvdbeek: label and unlabel are actions for labels (like kind/bug), edited does cover changing the PR title

nsoranzo: So we do need `unlabeled` for when `merge` or `status/WIP` is removed.
```suggestion
    types: [opened, reopened, edited, ready_for_review, unlabeled]
```

nsoranzo (approved):

PR #11403, opened 2021-02-17
Fix SLURM runner for real user setups and avoid uneccesary chowns

bernt-matthias:
for finished jobs (state DONE) the SLURM runner modifies the job's stderr file. for real user setups this does not work because the job dir is still owned by the real user. 

solution: just add a reclaim ownership before writing

I observed that for one finished job chown is now called 3x. Therefore the 2nd commit.

The change survived 1 successful job and 1 successful job with resubmission

### DIFF
--- a/lib/galaxy/jobs/runners/slurm.py
+++ b/lib/galaxy/jobs/runners/slurm.py
@@ -153,6 +153,7 @@ class SlurmJobRunner(DRMAAJobRunner):
                     self.work_queue.put((self.fail_job, ajs))
                     return
             if drmaa_state == self.drmaa_job_states.DONE:
+                ajs.job_wrapper.reclaim_ownership()
                 with open(ajs.error_file) as rfh:
                     _remove_spurious_top_lines(rfh, ajs)
                 with open(ajs.error_file, 'r+') as f:
--- a/lib/galaxy/util/path/__init__.py
+++ b/lib/galaxy/util/path/__init__.py
@@ -32,6 +32,7 @@ from os.path import (
     relpath,
     sep as separator,
 )
+from pathlib import Path
 try:
     from pwd import getpwuid
 except ImportError:
@@ -352,6 +353,9 @@ def external_chown(path, pwent, external_chown_script, description="file"):
     try:
         if not external_chown_script:
             raise ValueError('external_chown_script is not defined')
+        if Path(path).owner() == pwent[0]:
+            return True
+
         cmd = shlex.split(external_chown_script)
         cmd.extend([path, pwent[0], str(pwent[3])])
         log.debug('Changing ownership of {} with: {}'.format(path, ' '.join(map(shlex.quote, cmd))))


mvdbeek (approved):

bgruening:
Thanks!

@bernt-matthias is the real-user setup for SLIRM somewhere documented? I had someone a while back, that ask for it.

bernt-matthias:
> is the real-user setup for SLIRM somewhere documented? I had someone a while back, that ask for it.

Should all be here https://docs.galaxyproject.org/en/master/admin/cluster.html#submitting-jobs-as-the-real-user and work the same for all runners (I did not need to change anything for the switch from UNIVA to SLURM).

code4dna:
We have 21.01 with Slurm and had to comment the following lines in lib/galaxy/tools/actions/upload_common.py which runs locally and not as a submitted job
```
#            if link_data_only == 'copy_files' and trans.user:
#               external_chown(uploaded_dataset.path,
#                              trans.user.system_user_pwent(trans.app.config.real_system_username),
#                              trans.app.config.external_chown_script, description="uploaded file")
```

mvdbeek:
Could you let us know what the error was ?
Was this perhaps fixed in https://github.com/galaxyproject/galaxy/pull/11818 and we need to backport this to 21.01 ?

code4dna:
```
  File "/sw/galaxy/tools/data_source/upload.py", line 339, in <module>
    __main__()
  File "/sw/galaxy/tools/data_source/upload.py", line 332, in __main__
    metadata.append(add_file(dataset, registry, output_path))
  File "/sw/galaxy/tools/data_source/upload.py", line 143, in add_file
    convert_spaces_to_tabs=dataset.space_to_tab,
  File "/sw/galaxy/datatypes/upload_util.py", line 34, in handle_upload
    is_binary = check_binary(path)
  File "/sw/galaxy/lib/galaxy/util/checkers.py", line 51, in check_binary
    temp = open(name, "rb")
PermissionError: [Errno 13] Permission denied: '/scratch/user/galaxy/galaxy/tmp/fec93676e7253918-1621632183537-4549'
```
permissions for that file upload:
```
8.0K -rw-r----- 1 john45 john45 4.5K May 21 16:24 fec93676e7253918-1621632183537-4549
4.0K -rw------- 1 galaxy galaxy    1 May 21 16:24 tmp67h74a66
8.0K -rw------- 1 galaxy galaxy 4.5K May 21 16:24 tmp9iqh1sl8
4.0K -rw------- 1 galaxy galaxy   35 May 21 16:24 tmpnqsq3ruh
```

bernt-matthias:
Seems to be different error from the one fixed in https://github.com/galaxyproject/galaxy/pull/11818 .. still it might be worth trying it since the `if` statement also checks for `trans.app.config.external_chown_script`.

But I guess you define `external_chown_script` in your `galaxy.yml`?

Is it correct that you are submitting jobs to a cluster as real user, but upload jobs are processed locally and run as the Galaxy user? If this is correct I guess it will be tricky to solve in general since the `external_chown_script` currently applies globally an not per destination. Could you run the upload jobs locally, but as real users?

code4dna:
we have the external_chown_script in galaxy.yml. Jobs are submitted as real user and uploads are local jobs as the user galaxy which is what we did with our earlier v15.07 Galaxy. When we installed 15.07, we did not have a way to have the compute nodes connect to the internet to upload from sources such as the UCSC browser and may have run into other issues for user processes running on the login nodes. 
I just tested running upload local from computer as real user as a job on the compute nodes and it is working on our v20.09 Galaxy. I do have to set the UCSC tool to run locally since the compute nodes are not connected to the internet and it works running as user galaxy. We reused some config settings when jumping from 15.07 to 20.09 but it looks like they are not needed anymore.

PR #11404, opened 2021-02-17
Repair styleguide

Nerdinacan:
Fixed styleguide generation, also some minor refactoring.

**Moved styleguide build folder to docs/dist.**
The new build requires a couple style-guide specific files so I needed a "src" folder. Figured I'd use docs for that and just move the default styleguilde foler to docs/dist.

**Cleaned up loading functions for galaxy style sections**
Basically just put the galaxy specific MD files in their own directory for cleanliness reasons.

**Created a recursive component tree walker**
Styleguide sections are built by walking the components file directory and first looking for all *.md files.

If that markdown doc is named "index" or "readme" it becomes a new section for the folder it's in and the components in that module become grouped under the new subsection.

If a loose MD file is present in a folder, it becomes its own subsection which is composed of just that file. Allows us to create documentation topics that will appear in the appropriate place in the tree, but aren't tied to specific component files.

Titles of subsections are created by humanizing the file names. "provider-strategy.md" will be "Provider Strategy".

**Fixed build Error**
The build was failing due to our [dynamic reassignment of the webpack public path variabl](https://github.com/Nerdinacan/galaxy/blob/2deaa23c29285d91c396319a997d2541ba66c73c/client/src/onload/publicPath.js#L12)e. I believe this was done to allow asynchronous module loading.... a strategy I'm not 100% convinced we need or want.

But if we DO want to do asynchronous module loading, I would prefer if we had a pathing solution that is defined inside the build configs rather than in the src, if at all possible. It was hard to figure out why this wasn't working, and webpack is no fun to debug in the first place. This was kind of a landmine.


client/src/onload/publicPath.js
```diff
@@ -5,6 +5,10 @@
  * dynamic bundle imports load, otherwise they will fail.
  */
 
+// TODO: See if this is neccessary. I'm not convinced we need dynamic loads at all, but if we do, I
```
dannon: @Nerdinacan and I talked about this a bit last night, but for anyone else catching up on this -- we do need this here to support serving Galaxy at a proxy prefix that isn't known at runtime.

Pushing it exclusively back into the webpackConfig (which is how it used to work) doesn't solve the problem sufficiently because that route requires a client rebuild at deploy time -- an example of where this was originally a problem and where that approach would not work is the anvil use case where we have a container that has a prebuilt client that gets served to the end user with a prefix.

client/docs/sections.test.js
```diff
@@ -0,0 +1,69 @@
+import { ensureParentSections, buildSectionTree, processDocumentFile } from "./sections";
```
dannon: Thanks for adding a test; this and actually publishing it on updates in the future will keep this thing functional for longer than a month or two this time :)

dannon:
It looks like this is no longer loading any of the vue components, and the assets are still not resolving quite right?  ` downloadable font: download failed (font-family: "Font Awesome 5 Free" style:normal weight:900 stretch:100 src index:1): status=2147746065 source: http://192.168.0.101:6060/dist/assets/9451d5fee89b51a20f8a44c56a4f6c1a.woff2`

client/src/onload/publicPath.js
Nerdinacan: @dannon I still have questions.

The reason that we have this file is because we do separate dynamic loads. Is that correct? Without that, requirement we'd just be loading your modules from the webpack generated chunks, and this dynamic pathing would be unnecessary. With it you.... load like a single component separately?

I'm not seeing a big up-side to the dynamic load unless there's a variable I'm missing. Seems to me you'll just end up duplicating a lot of base library code in a separate pack for the sake of using a fancy webpack feature?

Nerdinacan:
Yeah I'll see if I can get a few of them working. I want to try to add a couple features (like a doc graph) while we're here.

client/docs/sections.test.js
Nerdinacan: I don't think it actually runs where it is though, haha. I built it in a place where you can run jest, then just moved it there so we had it. We should probably figure out how to make it run where it is.

client/src/onload/publicPath.js
Nerdinacan: I just did a search for "import(" in the code base and here are the places we do dynamic loads. 

[ListCollectionCreatorModal.js](https://github.com/Nerdinacan/galaxy/blob/2deaa23c29285d91c396319a997d2541ba66c73c/client/src/components/Collections/ListCollectionCreatorModal.js#L9)
PairedCollectionCreatorModal.js
PairedListCollectionCreatorModal.js
RuleBasedCollectionCreatorModal.js
WorkflowInvocationStep.vue
Workflow/Editor/modules/layout

The only time this looks even vaguely appropriate is on the Workflow editor because it loads a large package (ELM) for layout of the workflow graph. But even THERE, that package exposes a standard module import scheme.

Why load these things in a separate chunk? What's the upside other than secret build complexity problems?

Can we not just have webpack make an "elm" chunk and completely bypass these issues? Or are you wanting to change our code-splitting strategy. Perhaps looking forward to native html modules?

If your anvil example is still the issue, I don't see why letting webpack package all the code doesn't solve that issue. I may need somebody to baby-steps me through this, because I'm clearly not seeing the problem.

client/docs/sections.test.js
Nerdinacan: Fixed. Jest now runs in docs too.

client/src/onload/publicPath.js
Nerdinacan: (I can sideline this code change into another PR too, it's straying away from "getting the docs to work".)

dannon: (sorry, forgot to reply to the first question you asked):

> The reason that we have this file is because we do separate dynamic loads. Is that correct?

Yes.

The size of the galaxy dependency bundle is absolutely gigantic, and we need to be working towards more aggressive splitting, not away from it.  I think the only two of these I am directly responsible for and familiar with adding are the rulebuilder and the workflow editor, which are both big chunks (multiple megabytes) that we load on demand instead of in the primary modules, because much of the time a user doesn't touch them at all.

Native modules will be amazing at some point, but that's probably a whole separate unrelated conversation where we can talk about potentially dumping webpack completely.

This is ripped from a separate project I was tinkering with sorting out the application structure overhaul, but in Q2 with the application entrypoint work/etc, I do want to pursue things like this, in the routing:

```

const routes = [
    {
        path: "/",
        name: "Home",
        component: Home,
    },
    {
        path: "/about",
        name: "About",
        component: () => import(/* webpackChunkName: "about" */ "../views/About.vue"),
    },
    {
        path: "/dashboard",
        name: "Dashboard",
        component: () => import(/* webpackChunkName: "dashboard" */ "../views/Dashboard.vue"),
    },
 ```

...  Which is so much cleaner than trying to manually add specific bundle rules to webpack and whatnot.

I guess my question is -- dynamically loaded bundles have been working great and significantly reduce client load time, other than this one minor hiccup in the styleguide.  Is there a compelling reason for *not* using them for now?

Comment (Nerdinacan) [reply client/src/onload/publicPath.js]: Although I can't think of any other complications yet, messing with an internal webpack variable at runtime is disruptive. I just saw it and wondered if you were in fact leaning away from the webpack splitting mechanism.

I will work on webpack to better enable your scheme. :) 
I still think we can get this path into the appropriate webpack variable by inserting a value from the galaxy configs, even if we just read the config directly from galaxy/config.

I am definitely cool with the idea of splitting the bundles more, not less, but I would like to make sure that common libs are only written once and referenced, in turn, by your asynchronously loaded chunks, without duplicating common code.

I think you may be inadvertently duplicating a lot of common libs by doing this, enlarging the size of galaxy even more. I'll peek at it. (in another PR).

Comment (dannon) [reply client/src/onload/publicPath.js]: The redundant entrypoints do duplicate a bunch of code, but these dynamic bundles should not.  That's the whole point.  I don't think I'd call it 'my scheme', though -- it's right there in the webpack docs, including setting the publicpath dynamically at runtime for situations just like this :)

(edit: refs)
https://webpack.js.org/guides/code-splitting/#dynamic-imports
and 
https://webpack.js.org/guides/public-path/#on-the-fly

This is all totally normal webpack functionality that I have not seen discouraged anywhere.

Comment (Nerdinacan) [reply client/src/onload/publicPath.js]: My concern is certainly not that it is an evil scheme, but that every time we use a new plugin, package, whatever, it will expect the internal pathInfo to jive with the configs.

Comment (Nerdinacan) [reply client/src/onload/publicPath.js]: > The redundant entrypoints do duplicate a bunch of code, but these dynamic bundles should not.

That's interesting! Maybe we should rewrite the entry files to do this then.

Comment (dannon) [reply client/src/onload/publicPath.js]: @Nerdinacan That is the whole point.  We'll have a single entrypoint that dynamically loads all this crap using this mechanism :)

Comment (Nerdinacan) [reply client/src/onload/publicPath.js]: @dannon Yeah but even now, we can take those 4 entry points and write them with a promise-based dynamic import and immediately get a payoff.

Comment (dannon) [reply client/src/onload/publicPath.js]: @Nerdinacan Yep!

**Comment by Nerdinacan** [comment]
@dannon I think this is all set. I moved some of the docs I had been writing in a different PR into here so we could see how the new sections work. I think this is good enough to start building onto.

**Comment by dannon** [comment]
I'm getting this when trying to run it:

```
[yoplait@incus:~/work/galaxy/client] yarn run styleguide                               fix-styleguide ✭
yarn run v1.22.10
$ vue-styleguidist server
Not rendering styleguide for the following components: [
  '**/HistoryView.vue',
  '**/admin/DataManager/*',
  '**/LibraryFolder/*'
]
ℹ ｢wds｣: Project is running at http://localhost:6060/
ℹ ｢wds｣: webpack output is served from
ℹ ｢wds｣: Content not from webpack is served from /Users/yoplait/work/galaxy/client

You can now view your style guide in the browser:

  Local:            http://localhost:6060/
  On your network:  http://192.168.0.101:6060/

 FAIL  Failed to compile

./src/components/Collections/PairedListCollectionCreator.vue?vue&type=script&lang=js& (./node_modules/babel-loader/lib??ref--3!./node_modules/vue-loader/lib??vue-loader-options!./src/components/Collections/PairedListCollectionCreator.vue?vue&type=script&lang=js&)
Module not found: Can't resolve 'splitpanes' in '/Users/yoplait/work/galaxy/client/src/components/Collections'
./src/components/Collections/PairedListCollectionCreator.vue?vue&type=script&lang=js& (./node_modules/babel-loader/lib??ref--3!./node_modules/vue-loader/lib??vue-loader-options!./src/components/Collections/PairedListCollectionCreator.vue?vue&type=script&lang=js&)
Module not found: Can't resolve 'splitpanes/dist/splitpanes.css' in '/Users/yoplait/work/galaxy/client/src/components/Collections'
./src/components/LibraryFolder/LibraryFolderPermissions/LibraryFolderDatasetPermissions.vue?vue&type=script&lang=js& (./node_modules/babel-loader/lib??ref--3!./node_modules/vue-loader/lib??vue-loader-options!./src/components/LibraryFolder/LibraryFolderPermissions/LibraryFolderDatasetPermissions.vue?vue&type=script&lang=js&)
Module not found: Can't resolve 'vue-observe-visibility' in '/Users/yoplait/work/galaxy/client/src/components/LibraryFolder/LibraryFolderPermissions'
./src/components/LibraryFolder/LibraryFolderPermissions/PermissionsInputField.vue?vue&type=script&lang=js& (./node_modules/babel-loader/lib??ref--3!./node_modules/vue-loader/lib??vue-loader-options!./src/components/LibraryFolder/LibraryFolderPermissions/PermissionsInputField.vue?vue&type=script&lang=js&)
Module not found: Can't resolve 'vue-observe-visibility' in '/Users/yoplait/work/galaxy/client/src/components/LibraryFolder/LibraryFolderPermissions'
./src/components/Collections/PairedListCollectionCreator.vue?vue&type=script&lang=js& (./node_modules/babel-loader/lib??ref--3!./node_modules/vue-loader/lib??vue-loader-options!./src/components/Collections/PairedListCollectionCreator.vue?vue&type=script&lang=js&)
Module not found: Can't resolve 'vuedraggable' in '/Users/yoplait/work/galaxy/client/src/components/Collections'
./src/components/Collections/ListCollectionCreator.vue?vue&type=script&lang=js& (./node_modules/babel-loader/lib??ref--3!./node_modules/vue-loader/lib??vue-loader-options!./src/components/Collections/ListCollectionCreator.vue?vue&type=script&lang=js&)
Module not found: Can't resolve 'vuedraggable' in '/Users/yoplait/work/galaxy/client/src/components/Collections'
```

dannon:
Looks like a nice step forward, thanks!

PR #11408, opened 2021-02-18
[21.01] backport #11380 (config files)

bernt-matthias:


nsoranzo:
This is not needed, the changes will reach 21.01 when https://github.com/galaxyproject/galaxy/pull/11410 is merged forward.

PR #11409, opened 2021-02-18
[20.09] backport #11380 (config files)

bernt-matthias:


nsoranzo:
This is not needed, the changes will reach 20.09 when #11410 is merged forward.

PR #11411, opened 2021-02-18
Use request id middleware to scope session

mvdbeek:
We still need to remove the session at the end of the request (https://docs.sqlalchemy.org/en/13/orm/contextual.html#using-thread-local-scope-with-web-applications), so maybe
a middleware that does this is the right thing.


jmchilton:
Exciting stuff. When I looked starlette_context, I think I thought contextvars could be used to clean it up a bit (https://docs.python.org/3/library/contextvars.html) and I did find a backport to Python 3.6 for that standard library. But I think this is totally fine. The import in the model layer is a bit ugly but it is conditional and I think this solution is very nice and very flexible.

mvdbeek:
starlette_context uses contextvars internally, that's why the library is python >= 3.7. I could clean up the import by having a middleware (or a dependency) set the request id on ModelMapping and then have the identity function read that.

test/unit/api/test_fastapi.py
```diff
@@ -0,0 +1,67 @@
+import uuid
```
nsoranzo: This file can probably go in `test/unit/webapps` , to be symlinked by the `webapps` package?

test/unit/webapps/test_request_scoped_sqlalchemy_sessions.py
```diff
@@ -0,0 +1,162 @@
+import asyncio
```
jmchilton, line 1: The intention here is for ``galaxy-webapps`` to depend on Python 3.7 or newer? I so, we should fine a way to encode that in its setup.py maybe?

mvdbeek: I think galaxy-webapps should work fine on Python 3.6, as long as you don't build an ASGI app with `initialize_fast_app`. I don't mind putting a python 3.7 restriction in setup.py, but it's not 100% necessary.

mvdbeek:
I tried implementing the sqlalchemy session management as a middleware, and https://github.com/galaxyproject/galaxy/pull/11411/commits/2545965425f007cfcf725be8286dc06bf9c1995d is as far as I got. The commit message has the details on why I abandoned that direction. I'm now only using `starlette_context` for `X-Request-ID`, which I could also just push into the path dependency (that dependency is not why we need python 3.7, the `contextvars` backport package is not asyncio compatible).

The key is now that the path dependency sets a ContextVar on ModelMapping, and that gets us a request scoped session, AND that in the `finally` part of the dependency we close and remove any created session, which runs before any of the outgoing middlewares, so it's not blocking the streaming response and it won't be skipped on unhandled exceptions.

PR #11412, opened 2021-02-18
Refactor groups API logic into GroupsManager

davelopez:
## Overview
More work to ease the transition to FastAPI.

- Added API tests for all endpoints
- Replaced the old legacy_expose_api decorators.
- Moved the controller logic to a manager class.
- Refactored the manager to handle errors raising the appropriate exception.

## Additional comments
- I removed some of the `log.info` entries that looked like debug leftovers, but if they are important I can put them back :)
- The `update` action was not checking that updating the name of the group could conflict with an existing name so I added the check.

### DIFF
--- a/lib/galaxy/managers/groups.py
+++ b/lib/galaxy/managers/groups.py
@@ -6,6 +6,13 @@ from typing import (
 from sqlalchemy import false
 
 from galaxy import model
+from galaxy.app import StructuredApp
+from galaxy.exceptions import (
+    Conflict,
+    ObjectAttributeMissingException,
+    ObjectNotFound,
+)
+from galaxy.managers.base import decode_id
 from galaxy.managers.context import ProvidesAppContext
 from galaxy.schema.fields import EncodedDatabaseIdField
 from galaxy.web import url_for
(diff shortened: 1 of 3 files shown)


lib/galaxy/managers/groups.py
```diff
@@ -0,0 +1,102 @@
...
+        """
+        name = payload.get('name', None)
+        if name is None:
+            raise ObjectAttributeMissingException("Missing required name")
+        self._check_duplicated_group_name(trans, name)
+
+        group = model.Group(name=name)
+        trans.sa_session.add(group)
+        user_ids = payload.get('user_ids', [])
+        users = [trans.sa_session.query(model.User).get(self._decode_id(i)) for i in user_ids]
```
mvdbeek: It'll be better to first decode the user ids and then to fetch them all in one query.
users = trans.sa_session.query(model.User).filter(model.User.id.in_(user_ids)).all()
The same applies to the roles.
Ideally we'd also create models for Groups and Roles.

davelopez: Done! Thanks Marius for the hint!
I was avoiding touching the database stuff but I need to learn SQLAlchemy at some point 😆

PR #11416, opened 2021-02-19
[20.09] Fix sniffers of interval datatypes for files starting with many comments

nsoranzo:
For files starting with >60 comment lines, the `sniff_prefix()` method of Bed, Gff, Gff3 and Gtf classes was returning `True` when actually all headers lines were skipped.

Also:
- Don't skip lines having more than one column just because the first column is empty.
- Properly skip comment lines in `is_column_based()`
- Add test file which would have been incorrectly sniffed as `gtf` before. It still isn't sniffed as `gff3`, but that's fine because it has too many comment lines for the default header prefix.
- Set Vcf metadata also for VcfGz files.

Reported on Gitter by @FredericBGA.


mvdbeek:
> * Add test file which would have been incorrectly sniffed as `gtf` before. It still isn't sniffed as `vcf`, but that's fine because it has too many comment lines for the default header prefix.

I was worried about that at first, but we check for https://github.com/mvdbeek/galaxy/blob/e82e23d08f2487a4eba387c89be474b9f15ad941/lib/galaxy/datatypes/tabular.py#L739 on the first line, which VCFs should do. The test data isn't actually valid vcf, is it ?
The specs say:
```
A single ‘fileformat’ field is always required, must be the first line in the file, and details the VCF format version
number. For example, for VCF version 4.2, this line should read:
##fileformat=VCFv4.2
```

mvdbeek (approved):

mvdbeek:
maybe change the name of the test data file to `header_invalid_vcf.tabular` ?

nsoranzo:
Good point! I've replaced the non-comment lines in the new test file to make it a GFF3 file instead of VCF, since our Gff3 filter does not strictly require an initial comment line. So now if some comment lines are removed from the test file, it is correctly sniffed as GFF3.

mvdbeek:
Thanks @nsoranzo and @FredericBGA!

PR #11423, opened 2021-02-19
fixed incorrect javascript import syntax

Nerdinacan:
I noticed some of this kind of thing while I was getting the style guide working. 

Including the specific file type like this is often considered bad form since we'll have to refactor in the event that a single-file module turns into a folder.

It's also hosing 3rd party tooling like the styleguide.

I can't tell whether the authors did this because they didn't understand how webpack and javascript modules work or if there is a legitimate but undocumented need for the overly-specific typing. I'm assuming the former unless somebody speaks up.

There's a lot more of this kind of thing, but I figured I'd let the tests run on this small sample to doublecheck that it's just a mistake and not a necessary convention.

### DIFF
--- a/client/src/components/LibraryFolder/LibraryFolderRouter.js
+++ b/client/src/components/LibraryFolder/LibraryFolderRouter.js
@@ -1,10 +1,9 @@
-<script>
 import { getAppRoot } from "onload/loadConfig";
 import Vue from "vue";
 import VueRouter from "vue-router";
-import LibraryFolderPermissions from "components/LibraryFolder/LibraryFolderPermissions/LibraryFolderPermissions.vue";
-import LibraryFolder from "components/LibraryFolder/LibraryFolder.vue";
-import LibraryFolderDatasetPermissions from "components/LibraryFolder/LibraryFolderPermissions/LibraryFolderDatasetPermissions.vue";
+import LibraryFolderPermissions from "components/LibraryFolder/LibraryFolderPermissions/LibraryFolderPermissions";
+import LibraryFolder from "components/LibraryFolder/LibraryFolder";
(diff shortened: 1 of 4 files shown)


mvdbeek (approved):

dannon:
How was this breaking the styleguide?

This should be valid import syntax for explicit components that (in many places) simply predates the package-style import handling we've used more recently and (to be fair) never really made an explicit style choice about -- I would not characterize the author's understanding one way or the other.

Nerdinacan:
It's bad form to explicitly name a javascript module file-extension, for the reasons I mentioned. I'm not saying the compiler won't do it in most cases, it's just bad strategy.

As for the styleguide, I'm pretty sure the script tag with no template block is what is confusing vue-styleguidist. I can dig deeper, but I didn't bother because again.... again it's just bad form. If you want me to figure out exactly why vue-styleguidist didn't anticipate us using poor form, I guess I can, Gimme a sec.

Nerdinacan:
Ok in the styleguide PR I re-included these compents as written and the following errors appear when trying to add them to the styleguide:

```
Warning: Error parsing src/components/admin/DataManager/DataManagerRouter.vue: Error: No suitable component definition "
Warning: Error parsing src/components/LibraryFolder/LibraryFolderRouter.vue: Error: No suitable component definition fo"
```

I think it doesn't like your script block with no template block. I didn't write vue-styleguidist, I'm just trying to make it work, but it's not going to work with the way these router files are defined. I could dig into the source of vue-styleguidist, or I can just remove the script tags and, export it as javascript file it really is, and work from there.

dannon:
I think the correct fix here is to not have a simple router object as an SFC without the template block -- this is the part that was incorrect from the start.

I still don't think there's a problem with specifically naming .vue components when it is in fact a vue component -- unlike what we have here.

Issue #11420, opened 2021-02-19
Binary files with long chunks of initial non-binary contents fail to be detected as binary.


pcm32:
Flow cytometry fcs files have a long text header before the binary portion starts (https://en.wikipedia.org/wiki/Flow_Cytometry_Standard), in particular an example there has the binary data starting at byte offset 4,000 aprox. Currently, Galaxy on upload tries to determine if a file is a binary through a check that encompasses only  the first 1024 bytes (https://github.com/galaxyproject/galaxy/blob/f15cc2a892e0bf0c4f19cdf49e9a48699e52eb68/lib/galaxy/util/checkers.py#L48), making fcs files failed to be detected as binary (and producing other side effects I think, as these files appeared corrupted after going through Galaxy).

pcm32:
I am not sure if this is to blame, but for instance a round trip file to a Galaxy instance, even setting the FCS datatype when uploading, corrupts the FCS file:

```
docker run -v /Users/pmoreno/Downloads:/data \
 -it quay.io/biocontainers/bioconductor-flowcore:1.42.0--r3.4.1_0 R
######## in R ##################
library(flowCore)
input_file<-"/data/FlowRepository_FR-FCM-Z2KP_files/export_COVID19 samples 21_04_20_ST3_COVID19_ICU_015_A ST3 210420_069_Live_cells.fcs"
wFile<-read.FCS(input_file, transformation = F)

# no error, then trying with the same file after uploading to Galaxy and downloading from Galaxy
input_nw<-"/data/Galaxy16-[export_COVID19_samples_21_04_20_ST3_COVID19_ICU_015_A_ST3_210420_069_Live_cells.fcs].fcs"
nwFile<-read.FCS(input_nw, transformation = F)
Error in read.FCS(input_nw, transformation = F) :
  file /data/Galaxy16-[export_COVID19_samples_21_04_20_ST3_COVID19_ICU_015_A_ST3_210420_069_Live_cells.fcs].fcs seems to be corrupted.
In addition: Warning message:
In readFCSdata(con, offsets, txt, transformation, which.lines, scale,  :
  Error in reading data stream for file '/data/Galaxy16-[export_COVID19_samples_21_04_20_ST3_COVID19_ICU_015_A_ST3_210420_069_Live_cells.fcs].fcs'/nData may be truncated!
```

nsoranzo:
I think the easiest solution here could be to open a PR increasing the 1024 to a larger value.

[...]

PR #11425, opened 2021-02-21
Check for binary files at the middle of the file instead of at the beginning

pcm32:
This PR aims to deal with #11420, where certain binary files are not recognised as such because the have long text headers (>1024 bytes). Here we read the same 1024 bytes but from the middle of the file. I'm not sure if this could produce a delay for when Galaxy processes very long files: I'm not sure how expensive it is to move the byte reader to half of the file, I presume that if this behaves like a random access reader, it should be fine.

I have checked all test datatype files available in Galaxy and seems to be fine telling apart all binary from non binary examples, left the result in this gist: https://gist.github.com/pcm32/856c86308b9ee71ac58b16e07737e4b7

### DIFF
--- a/lib/galaxy/util/checkers.py
+++ b/lib/galaxy/util/checkers.py
@@ -58,7 +58,8 @@ def check_binary(name, file_path=True):
         # Read 1024 from the middle of the file if this is not
         # a gzip or zip compressed file (bzip are indexed),
         # to avoid issues with long txt headers on binary files.
-        if not is_gzip(file_path) and not is_zip(file_path):
+        if file_path and not is_gzip(name) and not is_zip(name) and not is_bz2(name):
+            # file_path=False doesn't seem to be used in the codebase
             temp.seek(read_start)
         return util.is_binary(temp.read(1024))
     finally:
--- a/lib/galaxy/util/checkers.py
+++ b/lib/galaxy/util/checkers.py
@@ -60,14 +60,17 @@ def check_binary(name, file_path=True):
         read_start = int(length / 2)
         read_length = min(1024, int(length / 2) - 1)
     try:
+        if util.is_binary(temp.read(1024)):
+            return True
+        # Some binary files have text only within the first 1024
         # Read 1024 from the middle of the file if this is not
         # a gzip or zip compressed file (bzip are indexed),
         # to avoid issues with long txt headers on binary files.
         if file_path and not is_gzip(name) and not is_zip(name) and not is_bz2(name):
             # file_path=False doesn't seem to be used in the codebase
             temp.seek(read_start)
-            read_length = 1024
-        return util.is_binary(temp.read(read_length))
+            return util.is_binary(temp.read(read_length))
+        return False
     finally:
         temp.close()
 
--- a/lib/galaxy/util/checkers.py
+++ b/lib/galaxy/util/checkers.py
@@ -50,10 +50,15 @@ def check_binary(name, file_path=True):
     # Handles files if file_path is True or text if file_path is False
     if file_path:
         temp = open(name, "rb")
-        read_start = int(os.stat(name).st_size / 2)
+        file_size = os.stat(name).st_size
+        read_start = int(file_size / 2)
+        # avoid reading beyond the end of the file
+        read_length = min(1024, int(file_size / 2) - 1)
     else:
         temp = BytesIO(name)
-        read_start = int(len(name) / 2)
+        length = len(name)
+        read_start = int(length / 2)
+        read_length = min(1024, int(length / 2) - 1)
     try:
         # Read 1024 from the middle of the file if this is not
         # a gzip or zip compressed file (bzip are indexed),
@@ -61,7 +66,8 @@ def check_binary(name, file_path=True):
         if file_path and not is_gzip(name) and not is_zip(name) and not is_bz2(name):
             # file_path=False doesn't seem to be used in the codebase
             temp.seek(read_start)
-        return util.is_binary(temp.read(1024))
+            read_length = 1024
+        return util.is_binary(temp.read(read_length))
     finally:
         temp.close()
 
--- a/lib/galaxy/util/checkers.py
+++ b/lib/galaxy/util/checkers.py
@@ -1,5 +1,6 @@
 import bz2
 import gzip
+import os
 import re
 import tarfile
 import zipfile
@@ -49,9 +50,14 @@ def check_binary(name, file_path=True):
     # Handles files if file_path is True or text if file_path is False
     if file_path:
         temp = open(name, "rb")
+        read_start = int(os.stat(name).st_size/2)
     else:
         temp = BytesIO(name)
+        read_start = int(len(name)/2)
     try:
+        # Read 1024 from the middle of the file,
+        # to avoid issues with long txt headers on binary files.
+        temp.seek(read_start)
         return util.is_binary(temp.read(1024))
     finally:
         temp.close()
--- a/lib/galaxy/util/checkers.py
+++ b/lib/galaxy/util/checkers.py
@@ -50,17 +50,14 @@ def check_binary(name, file_path=True):
     # Handles files if file_path is True or text if file_path is False
     if file_path:
         temp = open(name, "rb")
-        file_size = os.stat(name).st_size
-        read_start = int(file_size / 2)
-        # avoid reading beyond the end of the file
-        read_length = min(1024, int(file_size / 2) - 1)
+        size = os.stat(name).st_size
     else:
         temp = BytesIO(name)
-        length = len(name)
-        read_start = int(length / 2)
-        read_length = min(1024, int(length / 2) - 1)
+        size = len(name)
+    read_start = int(size / 2)
+    read_length = 1024
     try:
-        if util.is_binary(temp.read(1024)):
+        if util.is_binary(temp.read(read_length)):
             return True
         # Some binary files have text only within the first 1024
         # Read 1024 from the middle of the file if this is not
--- a/lib/galaxy/util/checkers.py
+++ b/lib/galaxy/util/checkers.py
@@ -55,9 +55,11 @@ def check_binary(name, file_path=True):
         temp = BytesIO(name)
         read_start = int(len(name) / 2)
     try:
-        # Read 1024 from the middle of the file,
+        # Read 1024 from the middle of the file if this is not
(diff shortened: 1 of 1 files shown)


pcm32:
Credits to @bernt-matthias for the idea of checking in the middle of the file.

pcm32:
There is some discussion of the complexity of f.seek [here](https://stackoverflow.com/questions/51801213/complexity-of-f-seek-in-python). Apparently this should be fine O(1) for binary file and in-memory, but possibly not ideal for text files O(n) (worst case). Probably compressed files is O(n) as well. Open to better ideas!

lib/galaxy/util/checkers.py
```diff
@@ -49,9 +50,14 @@ def check_binary(name, file_path=True):
     # Handles files if file_path is True or text if file_path is False
     if file_path:
         temp = open(name, "rb")
+        read_start = int(os.stat(name).st_size/2)
```
nsoranzo:
```suggestion
        read_start = int(os.stat(name).st_size / 2)
```

lib/galaxy/util/checkers.py
```diff
@@ -49,9 +50,14 @@ def check_binary(name, file_path=True):
...
     else:
         temp = BytesIO(name)
+        read_start = int(len(name)/2)
```
nsoranzo:
```suggestion
        read_start = int(len(name) / 2)
```

lib/galaxy/util/checkers.py
pcm32: Thanks, sorry, I had added the fix before seeing this, otherwise I would have committed your suggestion.

lib/galaxy/util/checkers.py
pcm32: Same as the other one.

pcm32:
The case that motivates this fix should be handled here https://github.com/galaxyproject/galaxy/blob/f15cc2a892e0bf0c4f19cdf49e9a48699e52eb68/lib/galaxy/datatypes/sniff.py#L759, but the instances that I'm testing this in do have that datatype marked as binary, so the harm to the files is already done probably at this point.

pcm32:
Maybe instead of doing this we should be fixing this https://github.com/galaxyproject/galaxy/issues/11420#issuecomment-782942604.

jmchilton:
It looks like those same tests are still failing. You may want to rebase against the latest dev to make sure it isn't these changes - but I suspect it is.

pcm32:
Rebased now. But @jmchilton you are not bothered with this seeking all the way to the middle of the file? Couldn't that introduce degradation when sniffing lots of large files?

dannon:
I don't imagine a seek to a computed offset is going to substantially slow down sniffing.  If we had to *read* to that point, or had to do weird stuff to figure out the start of a line or something, sure, but just an offset?  Should not be significant.

pcm32:
This error https://github.com/galaxyproject/galaxy/pull/11425/checks?check_run_id=2026338642#step:6:987 seems to be rightly be a difference in size from files beyond the allowed limit (21 bytes of difference vs 20 bytes allowed in the delta). That is for the framework part. I struggle to understand how can this be related to us seeking bytes at the middle of a file. Happy to try and fix if someone can explain.

pcm32:
This second error seems to be more related: https://github.com/galaxyproject/galaxy/pull/11425/checks?check_run_id=2026338363#step:12:828

pcm32:
The second failure might be explained because gzip files are not easy to seek... https://stackoverflow.com/questions/25985645/about-the-use-of-seek-on-gzip-files

pcm32:
Apparently for bzip you need to use special libraries for seeking, so might need to add that as well.

pcm32:
The file in the test failing was very small, so I'm wondering if it was reading past the end of the file end...

pcm32:
@dannon or @jmchilton I have managed to fix one of the two failures, but I'm a bit clueless on what part of my PR could be generating the difference in size in the test that is failing (21 bytes of difference when 20 bytes allowed). Any ideas? Happy to try them. Thanks!

pcm32:
The problem with the test failing seems to be that the 1.sff file seems to have its binary portion at the beginning of the file, so it probably ends up being detected as text, which probably leads to some corruption and difference then in handling as I observed with the FCS originally that were not being detected properly as binaries. The approach of picking a region of the file to check for it to be binary is a bit ill fated as there will always be some weird combination of format/size that will end up having text where you search for binary evidence... 

I wonder if we should be trying to make this work instead:
https://github.com/galaxyproject/galaxy/issues/11420#issuecomment-782942604

The other option would be to check both for the first 1024 and the middle bytes, believing any one of them.

pcm32:
Tada! This is passing now all tests! Finally!

pcm32:
Any chance this could be reviewed? Thanks!

lib/galaxy/util/checkers.py
```diff
@@ -49,10 +50,27 @@ def check_binary(name, file_path=True):
...
         temp = open(name, "rb")
+        file_size = os.stat(name).st_size
+        read_start = int(file_size / 2)
+        # avoid reading beyond the end of the file
+        read_length = min(1024, int(file_size / 2) - 1)
     else:
         temp = BytesIO(name)
+        length = len(name)
+        read_start = int(length / 2)
+        read_length = min(1024, int(length / 2) - 1)
```
davelopez: Hi @pcm32 
This is a small refactoring suggestion:
```suggestion
    if file_path:
        temp = open(name, "rb")
        size = os.stat(name).st_size
    else:
        temp = BytesIO(name)
        size = len(name)
    read_start = int(size / 2)
    read_length = 1024
```

If I'm not mistaken, the `read([bytes])` function won't read beyond EOF so it's safe to fix the `read_length` to the value you wish to read at most.

lib/galaxy/util/checkers.py
```diff
@@ -49,10 +50,27 @@ def check_binary(name, file_path=True):
...
     try:
-        return util.is_binary(temp.read(1024))
+        if util.is_binary(temp.read(1024)):
```
davelopez:
```suggestion
        if util.is_binary(temp.read(read_length)):
```

lib/galaxy/util/checkers.py
pcm32: Sounds good.

pcm32:
Thanks for the suggestions @davelopez, hopefully they won't upset the tests and we can merge this. 

Should I add @nsoranzo a test case for FCS given that it motivated this PR? Could you point me in the right direction please? Thanks.

mvdbeek:
> Should I add @nsoranzo a test case for FCS given that it motivated this PR? Could you point me in the right direction please? Thanks.

You can put a FCS file in ljb/galaxy/datatypes/test, it'll be tested automatically then

pcm32:
Then it is done already @mvdbeek , I added an FCS file there as part of another PR in the past: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/datatypes/test/Accuri_C6_A01_H2O.fcs

ksuderman:
While I was performing the release 21.05 test I ran some simple `timeit` tests with the files in `lib/galaxy/datatypes/test` plus several large-ish (10-200MB) FCS files and a 183MB zip file.  It looks like the `seek` operation add about 10% to the run  time,  but the `os.stat` required to find the file size adds 30%, making the *seek to middle* approach about 40% slower. 

However, I wonder if it wouldn't be better to sniff the first few bytes of a file looking for a magic number before falling back to seeking?  Most binary file formats (including FCS) include a magic number.  In fact the sniffer in [flow.py](https://github.com/galaxyproject/galaxy/blob/37a15c5319c8a8643a9ebc59c35e4802d1fd7b66/lib/galaxy/datatypes/flow.py#L43) does just that.  For the timing tests below I included a trie based implemention to indentify magic numbers.  Interestingly, the trie based approach is able to identify 73 of the test files while the current `check_binary` implementation only recognizes 47.  This is mostly likely due to the trie  recognizing *files with magic numbers*, which may include some text formats like XML and HTML

Each test run consisted of 10 rounds of 1,000 iterations on a m1.medium Jetstream instance. Times are in seconds.

| Score          | Magic  | Seek | Stat | No seek |
| ------------- | ------- | ------- | ------ | ------- |
| **Min**       | 3.180   | 4.036   | 3.672  | 2.820   |
| **Avg**       | 3.228   | 4.078   | 3.770  | 2.885   |
| **Max**       | 3.311   | 4.176   | 3.993  | 2.967   |
| **Std Dev**  | 0.044   | 0.043   | 0.121  | 0.055   |
|               |         |         |        |         |
| **% of Best** | 111.883 | 141.352 | 130.68 | 100.000 |

The **Stat** column is for the current `check_binary` algorithm but does an `os.stat` on the file so we may see the runtime difference for just the stat operation.  The **No seek** column is the currrent `check_binary` algorithm that just reads from the start of the file.

pcm32:
I have come across some new FCS files that still get corrupted on a round trip to Galaxy (upload and then download). I'm assuming it is for the same reason as before (file treated as non binary after failing the binary file check). Maybe the magic number approach would be the best solution.

pcm32:
@ksuderman do you have somewhere the code where you relied on magic numbers to try it? In the meantime we could also delay the file size check after attempting the first binary check, so that this would be applied to less files at least. Thanks!

pcm32:
For the issue of FCS files, would be it acceptable to add

```python
BINARY_CHARS = [NULL_CHAR, b'FCS2.0', b'FCS3.0', b'FCS3.1']
```

at https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/util/__init__.py#L91 ?

That would at least immediately detect those files as binary and avoid the corruption issues I get. Thanks!

ksuderman:
@pcm32 I am working on the PR right now and will try to post a draft later tonight or tomorrow.  Unfortunately, I don't think adding those strings to `BINARY_CHARS` will work as I believe it would cause any file with an F, C, S, 1, 2, 3, or . (dot) in them to be classified as binary. https://github.com/galaxyproject/galaxy/blob/c1eae49615665aa01840e47e5c917d4496a34795/lib/galaxy/util/__init__.py#L131

Also, since not all binary file types have magic numbers the algorithm will look something like:
```
def check_binary(path):
    if magic_number.sniff(path):
        return True
    if utils.is_binary(path):  # Current method for checking binary files
        return True
    return False
```

PR #11427, opened 2021-02-22
[20.09] Make hash selection mandatory

mvdbeek:
@hexylena sent a bug report just now about this


mvdbeek:
This also begs the question, did the validator not work, or was this in a workflow ?

hexylena:
D'oh.

It was in a workflow.

mvdbeek:
This kind of stuff should be preventable at the workflow submission level too, we could definitely use some work on that area.

nsoranzo (approved):

PR #11428, opened 2021-02-22
Pin pytest-postgresql to prevent pytest error on dependency update

jdavcs:
Otherwise pytest-postgresql will be updated to 2.6 on next dependency update, and that release breaks pytest.

See details, discussion upstream:
Issue: https://github.com/ClearcodeHQ/pytest-postgresql/issues/369
PR with fix: https://github.com/ClearcodeHQ/pytest-postgresql/pull/370


mvdbeek:
Sounds like the next release will be fixed and we don't need the pin ? (Thanks @ic4f! )

ic4f:
> Sounds like the next release will be fixed and we don't need the pin ? (Thanks @ic4f! )

Yes, I'll unpin it as soon as it's fixed upstream.

ic4f:
and it's merged! :) Checking... if all is good, I"ll close this.

ic4f:
Yes, I think it's safe to close this and wait for the patched release.

PR #11429, opened 2021-02-22
Add UpgradeAllSteps workflow refactor action

simonbray:
An action to update all tools and subworkflows in a workflow, as discussed here: https://github.com/galaxyproject/galaxy/issues/11005. Also a button to run the upgrade through the UI.

@jmchilton could you have a look please?


client/src/components/Workflow/Editor/Index.vue
```diff
@@ -422,6 +423,9 @@ export default {
             this.$refs.lint.refresh();
             showLint();
         },
+        onUpgrade() {
+            this.attemptRefactor([{'action_type': 'upgrade_workflow'}]);
```
simonbray: One small problem is that the change isn't visible immediately, the page has to be refreshed. I tried a few things without success, hopefully this is fixable in about 5 seconds for someone who actually knows JS. :)

lib/galaxy/workflow/refactor/schema.py
```diff
@@ -211,6 +211,10 @@ class UpgradeToolAction(BaseAction):
     tool_version: Optional[str]
 
 
+class UpgradeWorkflowAction(BaseAction):
+    action_type: Literal['upgrade_workflow']
```
jmchilton: What about ``upgrade_all_steps``? I don't know, what you have here is a fine choice I'm just worried it is a little confusing with ``upgrade_subworkflow``.

If no one else feels strongly though, I'm okay with keeping it like it like you have it here I think.

client/src/components/Workflow/Editor/Options.vue
```diff
@@ -60,6 +60,9 @@
             <b-dropdown-item href="#" @click="$emit('onLayout')"
                 ><span class="fa fa-align-left mr-1" />Auto Layout</b-dropdown-item
             >
+            <b-dropdown-item href="#" @click="$emit('onUpgrade')"
+                ><span class="fa fa-recycle mr-1" />Upgrade All Workflow Steps</b-dropdown-item
+            >
```
jmchilton, line 65: What a clean UI implementation!

jmchilton:
Awesome stuff! There are no tests - does it work?

jmchilton (approved):

lib/galaxy/workflow/refactor/schema.py
simonbray: Sure, I agree that's clearer.

simonbray:
> does it work?

Haha - I was thinking of writing a test, but I wouldn't mind some guidance. Are there Galaxy test tools available which exist in multiple versions?

jmchilton:
Yes - there is a test at https://github.com/galaxyproject/galaxy/blob/dev/test/integration/test_workflow_refactoring.py#L483 for instances. So there is a tool with ID ``multiple_versions`` that gets loaded and used there. 

And another tool with ID ``multiple_versions_changes`` used at https://github.com/galaxyproject/galaxy/blob/dev/test/integration/test_workflow_refactoring.py#L510.

These tools are found in this directory (https://github.com/galaxyproject/galaxy/tree/dev/test/functional/tools) and added in this file (https://github.com/galaxyproject/galaxy/blob/dev/test/functional/tools/samples_tool_conf.xml) if you wanted to add some more tools or versions.

jmchilton:
This is just beautiful! Thanks so much.

PR #11436, opened 2021-02-22
[WIP] Bulk Operations panel for beta history

Nerdinacan:
This is the dynamic selection bulk-operation tool.

Users will be able to create a dynamic selection from the history, see a sample result in a dialog with a total row-count for matched history content elements, then pick an operation to run against all the selected content.

## New Selection Mode

I think the easiest and most concise way to implement this is going to be to allow the user to pick between selecting individual items and selecting items from history as a query. Conceptually, the user ALREADY selects items from the history as a query when they use the filter boxes and the deleted/hidden toggles, so my suggestion is that we reformulate all the "bulk operation" menu items to reflect the split in what it means to be a selection.

Users will make a selection using the normal filtering tools, pick an operation from a list, pass through a confirmation dialog and the UI will figure out whether it has to run on a selection on the server or the 10 rows the user may have selected in "item selection mode"

### The upside of this is we lose a lot of menu items.

Software with a lot of options can be confusing, but worse is software with a lot of similar options where you're not sure which one is right.

"Delete all Hidden Datasets" now just becomes the operation "Delete" which will 
* delete all hidden content if the user has selected the "hidden" toggle on the filters. 
* delete all content named "Dave" if they have text-filtered that same list.
* delete datasets tagged with a certain string if they're using the advanced filters
* or it can delete the 10 things they chose if "item selection" is toggled and they have explicitly check-boxed a bunch of items.

So we can remove the explicit delete buttons for those individual scenarios.

### The downside is that those operations now mean multiple things to the client code

Those operations must now conditionally figure out whether the user means the 10 items that were selected or a bulk operation defined by the search filters, but that's pretty easy since we already have model objects which represent both those situations.

I think, to the user, these distinctions should be invisible. The client software should be in charge of worrying about how to implement their type of selection. The user should only have to "delete the stuff I selected". 

### Unified confirm for all operations

In all cases the user should see a confirmation dialog that says "You are about to delete X items, you sure about that?"

### Surprise bonus

Turns out there's a way to use the query that returns the stats of a selection (total server-side matches) to increase the efficiency of the beta server-side query process. With the addition of one quick count it will no longer be necessary to do a bi-directional query server-side. We're still stuck with it in the client, but this means I can throw out roughly half of the SQL queries that the current history polling system performs.

We can trade one initial count query for about half of the subsequent SQL queries performed during history polling.


astrovsky01:
Some thoughts:

* If the filters are at default (hidden+deleted datasets invisible), list datasets as "x datasets in history" or something other than "Total Matches".
* The checkbox to activate multi-dataset select suggests to me it's a "select all" button rather than something to open the select dataset menu.
* For the execute drop down, add space above and a label for something like "History functions".
* The collapse all button remains greyed out until next click anywhere on screen, no matter mouse position.
Add filter for error datasets or a hide/delete errored dataset function.
* I love the expanding menu of operations, but reorder the buttons? Might just be me, but I would think I would use 'extract history' and 'resume paused jobs' more often than duplicate or share, and based on how wide a user would generally keep their history panel, it makes sense to have the most-used buttons up front.
* Filter buttons (hidden/deleted) only show &&. I see no way to use an or. I can use a search, but the popup for filter keywords on the current history filter bar is gone.
* Select history button doesn't need ellipsis. It looks like something got cut off.

PR #11440, opened 2021-02-22
interactivetools_proxy_host typo

almahmoud:
Pretty sure this is a typo since all the other ones are `interactivetools_`, not sure how it hasn't affected people in the past


dannon:
Yep, it looks like in the config it is `interactivetools_proxy_host` too: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/config_schema.yml#L1117

dannon:
(merging forward #11441)

PR #10804, opened 2020-11-25
Speedup xml replacement

mvdbeek:
This speeds up tool parsing significantly (if the tool isn't already in the tool cache).
I would like to thank John from February 2013 for the hint :).
Before:
```
*** PROFILER RESULTS ***
load_with_references (/Users/mvandenb/src/galaxy/lib/galaxy/util/xml_macros.py:10)
function called 2057 times

         37174705 function calls (26399913 primitive calls) in 159.561 seconds

   Ordered by: cumulative time, internal time, call count
   List reduced from 63 to 40 due to restriction <40>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     2057    0.019    0.000  159.561    0.078 xml_macros.py:10(load_with_references)
36510/2049   10.706    0.000  151.955    0.074 xml_macros.py:126(_expand_macros)
34461/23486    0.246    0.000  123.611    0.005 xml_macros.py:140(_expand_macro)
    34461  115.626    0.003  115.626    0.003 xml_macros.py:159(<dictcomp>)
5403373/79830   13.579    0.000   21.819    0.000 xml_macros.py:106(_expand_tokens_for_el)
5415134/73970    1.894    0.000   21.360    0.000 xml_macros.py:98(_expand_tokens)
     2049    0.007    0.000    6.373    0.003 xml_macros.py:63(_import_macros)
 11965416    5.378    0.000    6.353    0.000 xml_macros.py:119(_expand_tokens_str)
     3880    6.227    0.002    6.227    0.002 __init__.py:227(parse_xml)
3323/1500    0.011    0.000    5.820    0.004 xml_macros.py:187(_load_macros)
3323/1500    0.009    0.000    5.801    0.004 xml_macros.py:224(_load_imported_macros)
1823/1817    0.004    0.000    5.771    0.003 xml_macros.py:250(_load_macro_file)
159530/98533    2.130    0.000    2.492    0.000 copy.py:128(deepcopy)
    37859    0.292    0.000    1.612    0.000 xml_macros.py:263(_xml_replace)
    34461    0.060    0.000    1.058    0.000 copy.py:200(_deepcopy_list)
 12008626    0.957    0.000    0.957    0.000 {method 'items' of 'dict' objects}
    34461    0.124    0.000    0.737    0.000 xml_macros.py:163(_expand_yield_statements)
     2057    0.003    0.000    0.696    0.000 xml_macros.py:50(raw_xml_tree)
[...]

PR #11379, opened 2021-02-15
[21.01] Release Notes for 21.01

mvdbeek:

PR #11447, opened 2021-02-23
[21.01] Disable tool document cache by default

mvdbeek:
It's kind of brittle, and https://github.com/galaxyproject/galaxy/pull/10804
has improved the parsing so much that it doesn't add much benefit.
Instances on Anvil start up in ~ 32 seconds, with all of main's toolbox.

### DIFF
--- a/lib/galaxy/webapps/galaxy/config_schema.yml
+++ b/lib/galaxy/webapps/galaxy/config_schema.yml
@@ -770,9 +770,9 @@ mapping:
         default: false
         required: false
         desc: |
-          Set this to true to enable the tool document cache. This cache stores
-          expanded xml strings. Enabling the tool cache results in slightly faster startup
-          times. The tool cache is backed by sqlite database, which cannot
+          Whether to enable the tool document cache. This cache stores
+          expanded XML strings. Enabling the tool cache results in slightly faster startup
+          times. The tool cache is backed by a SQLite database, which cannot
           be stored on certain network disks. The cache location is configurable
           using the ``tool_cache_data_dir`` setting, but can be disabled completely here.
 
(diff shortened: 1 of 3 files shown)


lib/galaxy/webapps/galaxy/config_schema.yml
```diff
@@ -767,11 +767,11 @@ mapping:
...
         type: bool
-        default: true
+        default: false
         required: false
         desc: |
-          Set this to false to disable the tool document cache. This cache stores
-          expanded xml strings. Disabling the tool cache results in slower startup
+          Set this to true to enable the tool document cache. This cache stores
+          expanded xml strings. Enabling the tool cache results in slightly faster startup
           times. The tool cache is backed by sqlite database, which cannot
```
```suggestion
          Whether to enable the tool document cache. This cache stores
          expanded XML strings. Enabling the tool cache results in slightly faster startup
          times. The tool cache is backed by a SQLite database, which cannot
```

nsoranzo (approved):

nsoranzo:
Needs to be added to the Release Notes https://github.com/galaxyproject/galaxy/pull/11379 .

mvdbeek:
I will run the release note script before doing the release, that catches all of these PRs still coming in.

nsoranzo (approved):

nsoranzo:
> I will run the release note script before doing the release, that catches all of these PRs still coming in.

I meant in the "Configuration Changes" section that we introduced for 20.09: https://docs.galaxyproject.org/en/master/releases/20.09_announce.html#configuration-changes

It looks like that the script @hexylena added in https://github.com/galaxyproject/galaxy/pull/10258 for this is not part of the `bootstrap_history.py` script yet.

PR #11448, opened 2021-02-23
[20.09] always mount a job tmp rw

bernt-matthias:
Fixes: 

- In containerized tool runs `$_GALAXY_JOB_TMP_DIR` is created, the variable is "exported", but the directory **is not mounted writable**. 
- add `export _GALAXY_JOB_TMP_DIR` to make it usable in non-containerized tool runs
- `export GALAXY_SLOTS` as part of the cluster slot statement

### DIFF
--- a/test/functional/tools/job_environment_default.xml
+++ b/test/functional/tools/job_environment_default.xml
@@ -3,14 +3,16 @@
       <container type="docker">busybox:ubuntu-14.04</container>
     </requirements>
     <command detect_errors="exit_code"><![CDATA[
-      echo 'Writing environment properties to output files.';
-      (>&2 echo 'Example tool stderr output.');
-      echo `id -u` > '$user_id';
-      echo `id -g` > '$group_id';
-      echo `pwd` > '$pwd';
-      echo "\$HOME" > '$home';
-      echo "\$TMP"  > '$tmp';
-      echo "\$SOME_ENV_VAR" > '$some_env_var';
+      echo 'Writing environment properties to output files.' &&
+      (>&2 echo 'Example tool stderr output.') &&
+      echo `id -u` > '$user_id' &&
+      echo `id -g` > '$group_id' &&
+      echo `pwd` > '$pwd' &&
+      echo "\$HOME" > '$home' &&
+      echo "\$TMP"  > '$tmp' &&
+      echo "\$SOME_ENV_VAR" > '$some_env_var' &&
+      touch '\$_GALAXY_JOB_TMP_DIR/test' &&
+      touch '\$_GALAXY_JOB_HOME_DIR/test
     ]]></command>
     <inputs>
     </inputs>
--- a/test/functional/tools/job_environment_default.xml
+++ b/test/functional/tools/job_environment_default.xml
@@ -12,7 +12,7 @@
       echo "\$TMP"  > '$tmp' &&
       echo "\$SOME_ENV_VAR" > '$some_env_var' &&
       touch "\$_GALAXY_JOB_TMP_DIR/tmp_test" &&
-      touch "\$_GALAXY_JOB_HOME_DIR/home_test"
+      touch "\$HOME/home_test"
     ]]></command>
     <inputs>
     </inputs>
--- a/test/functional/tools/job_environment_default.xml
+++ b/test/functional/tools/job_environment_default.xml
@@ -3,7 +3,6 @@
       <container type="docker">busybox:ubuntu-14.04</container>
     </requirements>
     <command detect_errors="exit_code"><![CDATA[
-<<<<<<< HEAD
       echo 'Writing environment properties to output files.' &&
       (>&2 echo 'Example tool stderr output.') &&
       echo `id -u` > '$user_id' &&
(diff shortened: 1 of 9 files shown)


bernt-matthias:
Not sure if the tests check for tool error, i.e. if they catch if writing goes wrong.

mvdbeek:
Yes, they do. You can actually see that in the failing kubernetes runner job, which bypasses a lot of Galaxy job logic :(.

mvdbeek:
@nuwang @almahmoud  any chance I could convince you to look into how the kubernetes job runner wants to handle `$_GALAXY_JOB_TMP_DIR ` ?

bernt-matthias:
Do you think it would be a good idea to merge this PR with that https://github.com/galaxyproject/galaxy/pull/10820

test/functional/tools/job_environment_default.xml
```diff
@@ -3,14 +3,17 @@
...
+      echo 'Writing environment properties to output files.' &&
+      (>&2 echo 'Example tool stderr output.') &&
+      echo `id -u` > '$user_id' &&
+      echo `id -g` > '$group_id' &&
+      echo `pwd` > '$pwd' &&
+      echo "\$HOME" > '$home' &&
+      echo "\$TMP"  > '$tmp' &&
+      echo "\$SOME_ENV_VAR" > '$some_env_var' &&
+      ## test if job home and tmp are mounted rw
+      touch '\$_GALAXY_JOB_TMP_DIR/test' &&
```
mvdbeek:
```suggestion
      touch "\$_GALAXY_JOB_TMP_DIR/test" &&
```

test/functional/tools/job_environment_default.xml
```diff
@@ -3,14 +3,17 @@
...
+      touch '\$_GALAXY_JOB_HOME_DIR/test'
```
mvdbeek:
```suggestion
      touch "\$_GALAXY_JOB_HOME_DIR/test"
```

mvdbeek (commented):
You need to double quote environment variables

mvdbeek:
No, these are separately solvable issues that don't depend on each other.

almahmoud:
Unless i'm misunderstanding, I think in theory in k8s the container will be able to write at any path as ephemeral storage. We only need to mount a volume for it if we're worried about running out of space on the node and/or if we need it persisted in any way. This would only be a problem if the amount of data in ephemeral storage is huge and the tmp directory is not on the NFS. I thought it's relative to the jobs working dir (which is mounted NFS) by default, my memory might be wrong. @nuwang correct me if i'm wrong

mvdbeek:
The kubernetes runner skips all of the job setup where we're setting up things for the tool container. So `$_GALAXY_JOB_TMP_DIR` is never defined. I guess you'd need to set this up somewhere here: https://github.com/Nerdinacan/galaxy/blob/e212e89cf82e0640689f349d9d6595637e0d225c/lib/galaxy/jobs/runners/kubernetes.py#L306

mvdbeek:
@bernt-matthias actually yes, can you merge the PRs ? Is that why it fails ?

bernt-matthias:
> Is that why it fails ?

Could be. I did not have time to check the error log in detail. 

Note that I'm not confident of the last commit (https://github.com/galaxyproject/galaxy/pull/11448/commits/365fb066b0d8c86f4d45a3e626969756506c315a). It was a suggestion by @nsoranzo. See discussion here https://github.com/galaxyproject/galaxy/pull/10820#pullrequestreview-540138753

bernt-matthias:
> I guess you'd need to set this up somewhere here

How can I access the job_info from the ajs?

bernt-matthias:
Its really hard to debug this with all those `urllib3.connectionpool` messages

nsoranzo:
> Note that I'm not confident of the last commit ([365fb06](https://github.com/galaxyproject/galaxy/commit/365fb066b0d8c86f4d45a3e626969756506c315a)). It was a suggestion by @nsoranzo. See discussion here [#10820 (review)](https://github.com/galaxyproject/galaxy/pull/10820#pullrequestreview-540138753)

I have just added a lenghty explanation there, hope that helps.

bernt-matthias:
Thanks @nsoranzo. Then it might be reasonable to treat the temp for variable in the same way, or?

bernt-matthias:
Hi @nsoranzo I did some more experiments. In the current test we write into `$_GALAXY_JOB_HOME_DIR/` 

https://github.com/galaxyproject/galaxy/blob/de572726fbdbffdd27011ece76cc51450570a565/test/functional/tools/job_environment_default.xml#L16

This variable is not exported due to https://github.com/galaxyproject/galaxy/pull/11448/commits/365fb066b0d8c86f4d45a3e626969756506c315a .. but "only" `$HOME`

So I reverted the commit .. alternatively we could try to write into `$HOME/`, of course.

nsoranzo:
> Hi @nsoranzo I did some more experiments. In the current test we write into `$_GALAXY_JOB_HOME_DIR/`
> 
> https://github.com/galaxyproject/galaxy/blob/de572726fbdbffdd27011ece76cc51450570a565/test/functional/tools/job_environment_default.xml#L16
> 
> This variable is not exported due to [365fb06](https://github.com/galaxyproject/galaxy/commit/365fb066b0d8c86f4d45a3e626969756506c315a) .. but "only" `$HOME`
> 
> So I reverted the commit .. alternatively we could try to write into `$HOME/`, of course.

In my `$_GALAXY_JOB_HOME_DIR` should not be documented, it's an internal Galaxy thing, the test should try to write into `$HOME`, which is what tools normally (ab)use.

bernt-matthias:
> the test should try to write into $HOME

OK. Done. 

I think I also fixed one of the failing tests https://github.com/galaxyproject/galaxy/pull/11448/commits/5c272f76fa3b20b3a421ada6a66edd6a22f366ee ..

mvdbeek (approved):

innovate-invent:
I think this PR falsely assumes that the tmp dir is a host path.
When outputs_to_working_directory: True and the working directory is hosted in a docker volume a path on the host system is being created. This problem compounds when multiple instances of Galaxy are writing to the same host system jobs folder as a result of this.

mvdbeek:
Are you actually seeing a problem with this ? Because the kubernetes runner shouldn't be using any of this, since it doesn't run containers using the CLI. And $_GALAXY_JOB_TMP_DIR defaults to the job dir, which needs to be writable by the job handler in any case.

> This problem compounds when multiple instances of Galaxy are writing to the same host system jobs folder as a result of this.

How's that even possible ? Maybe I'm misunderstanding what you mean by multiple Galaxy instance, but if you configure multiple Galaxy instances to write to the same job directories that is absolutely not supported. A single instance with multiple handlers is only going to handle a single job per handler, so again I can't see how this would happen in that supported case.

innovate-invent:
This is running locally using docker. Kubernetes is not involved in any way.

The issue is that `$_GALAXY_JOB_TMP_DIR:$_GALAXY_JOB_TMP_DIR:rw` assumes that $_GALAXY_JOB_TMP_DIR exists on the host. In my case the path that $_GALAXY_JOB_TMP_DIR resolves to is in a docker volume.

The better solution is to provide a $tmp_dir variable in the [docker_volumes runner spec](https://github.com/galaxyproject/galaxy/blob/a0a274cb3a58daee7443973496eedcbacceb1e19/lib/galaxy/config/sample/job_conf.xml.sample_advanced#L464).

The $tmp_dir variable can then resolve to `self.job_info.tmp_directory or '$_GALAXY_JOB_TMP_DIR'`.
I can then omit $tmp_dir from the docker_volumes config as the data volume where the job folder exists is already in the spec and the $_GALAXY_JOB_TMP_DIR path remains valid.

PR #11452, opened 2021-02-23
Move unit tests to GitHub workflows, test also under Python 3.9

nsoranzo:
Also some random fixes to other GitHub workflows.

### DIFF
--- a/.github/workflows/unit.yaml
+++ b/.github/workflows/unit.yaml
@@ -1,4 +1,4 @@
-name: Python linting
+name: Unit tests
 on: [push, pull_request]
 jobs:
   test:
@@ -6,9 +6,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.6', '3.9']
+        python-version: ['3.6']
     steps:
       - uses: actions/checkout@v2
+        with:
+          path: 'galaxy root'
       - uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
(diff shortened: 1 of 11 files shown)


.github/workflows/unit.yaml
```diff
@@ -0,0 +1,28 @@
+name: Unit tests
+on: [push, pull_request]
+jobs:
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.6']
```
mvdbeek:
```suggestion
        python-version: ['3.6', '3.9']
```

.github/workflows/unit.yaml
```diff
@@ -0,0 +1,28 @@
+name: Unit tests
+on: [push, pull_request]
+jobs:
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
```
mvdbeek, line 9:
```suggestion
      fail-fast: false
      matrix:
```

mvdbeek (approved):
Should we test more python versions ?

.github/workflows/unit.yaml
nsoranzo: Maybe 3.6 and 3.9, like for lint workflow?

dannon:
Unit test failure looks real -- fractions.gcd import failing in 3.9?  (math.gcd now?)

nsoranzo:
> Unit test failure looks real -- `fractions.gcd` import failing in 3.9? (math.gcd now?)

Yes, unfortunately it's not in our code but in `networkx`, which comes in (pinned to 1.11) as a depedency of [isa-rwval](https://github.com/ISA-tools/isa-rwval). The `fractions.gcd` import is fixed in networkx >= 2.4 .

Not sure what's the best course of action here, any suggestion @djcomlab ?

djcomlab:
> depedency of [isa-rwval](https://github.com/ISA-tools/isa-rwval). The `fractions.gcd` import is fixed in networkx >= 2.4 .
> 
> Not sure what's the best course of action here, any suggestion @djcomlab ?

I believe `isa-rwval` can probably be updated quickly. I will check it today and let you know.

nsoranzo:
> Should we test more Python versions ?

@mvdbeek I was thinking of adding '3.9' also to `.github/workflows/first_startup.yaml`, what do you think?

mvdbeek:
Sounds good

nsoranzo:
All tests pass now! Thanks @djcomlab for the help!

mvdbeek (approved):

PR #11453, opened 2021-02-23
A new datatype cmap. 

jaidevjoshi83:
This PR includes changes in the tabular.py and conf file for the camp datatype.


jaidevjoshi83:
Please review and let me know if there is any suggestion.  Thank you.

lib/galaxy/config/sample/datatypes_conf.xml.sample
```diff
@@ -328,6 +328,7 @@
       <display file="ucsc/maf_customtrack.xml"/>
     </datatype>
     <datatype extension="mtx" type="galaxy.datatypes.tabular:MatrixMarket" display_in_upload="true"/>
+    <datatype extension="cmap" type="galaxy.datatypes.tabular:CMAP" display_in_upload="true"/>
```
bgruening, line 331: @jaidevjoshi83 I think you need to register the sniffer in this file below as well.

bgruening: Can you maybe add a description and a `description_url` to https://bionanogenomics.com/wp-content/uploads/2017/03/30039-CMAP-File-Format-Specification-Sheet.pdf

jaidevjoshi83:
A new updated PR (https://github.com/galaxyproject/galaxy/pull/11471 ) has been opened which is targeting "release_21.01". I included changes in the new PR as per @bgruening suggestions.

Issue #11432, opened 2021-02-22
[21.01] Select option no longer present when uploading dataset from workflow form.


jmchilton:
Tested on usegalaxy.org as part of the workflow tutorial.

jmchilton:
This is seeming very non-trivial to fix. I tried to pass selectable around more. Namely this:

```diff
+++ b/client/src/components/Upload/helpers.js
@@ -43,6 +43,10 @@ export const commonProps = {
         type: Function,
         default: null,
     },
+    selectable: {
+        type: Boolean,
+        default: false,
+    },
     auto: {
         type: Object,
         default: function () {
```

with the addition of initializing the option in a couple other places.

But then whether the button would appear or now depends on if the tool panel button was clicked first in a session or the tool form button. So maybe there is some state sharing - but it results in some Vue property issues.

<img width="607" alt="Screen Shot 2021-02-22 at 10 07 00 PM" src="https://user-images.githubusercontent.com/216771/108797898-5ea04400-755a-11eb-8b7f-86fd0e97f2d3.png">

mvdbeek:
@Nerdinacan any chance you could look at this ? I think that is related to mounting the upload modal just once ? Fwiw the callback should be passed in https://github.com/jmchilton/galaxy/blob/a12316de44cf663f03235d3496830eacad917ea6/client/src/components/DataDialog/DataDialog.vue#L138
We'd probably have to emit the `selectable` throught the eventHub ? This should make it to 21.01.

Nerdinacan:
@mvdbeek Will do. Hopefully it's just a missed prop definition in the wrappers that get it initialized

Nerdinacan:
https://github.com/galaxyproject/galaxy/pull/11457

mvdbeek:
Fixed in https://github.com/galaxyproject/galaxy/pull/11457, minor tweak in #11576

PR #11457, opened 2021-02-23
[21.01] Fix selecting datasets from workflow & tool form upload

Nerdinacan:
Addresses:
https://github.com/galaxyproject/galaxy/issues/11432

selectable and callback props were not capable of being re-assigned upon successive openings of the upload dialog, because the author imperatively re-assigned all properties to internal non-reactive variables instead of using Vue properties

Edit: Marius was nice enough to locate the issue with a lilttle video

second upload button

### DIFF
--- a/client/src/components/DataDialog/DataDialog.vue
+++ b/client/src/components/DataDialog/DataDialog.vue
@@ -49,7 +49,7 @@ import { UrlTracker } from "./utilities";
 import { Model } from "./model";
 import { Services } from "./services";
 import { getAppRoot } from "onload/loadConfig";
-import { openUploadModal } from "components/Upload";
+import { mountUploadModal } from "components/Upload";
 
 Vue.use(BootstrapVue);
 
@@ -134,8 +134,9 @@ export default {
                 format: this.format,
                 callback: this.callback,
                 modalShow: true,
+                selectable: true,
             };
-            openUploadModal(propsData);
+            mountUploadModal(propsData);
             this.modalShow = false;
         },
         /** Called when selection is complete, values are formatted and parsed to external callback **/
--- a/client/src/components/Upload/Collection.vue
+++ b/client/src/components/Upload/Collection.vue
@@ -43,7 +43,7 @@
         </template>
         <template v-slot:buttons>
             <b-button ref="btnClose" class="ui-button-default" id="btn-close" @click="$emit('dismiss')">
-                {{ btnCloseTitle }}
+                {{ btnCloseTitle | localize }}
             </b-button>
             <b-button
                 ref="btnReset"
@@ -157,7 +157,6 @@ export default {
             btnBuildTitle: _l("Build"),
             btnStopTitle: _l("Pause"),
             btnResetTitle: _l("Reset"),
-            btnCloseTitle: this.app.callback ? _l("Cancel") : _l("Close"),
         };
     },
     created() {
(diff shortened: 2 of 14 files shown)


jmchilton:
That fixes some serious issues!

There are still outstanding issues:
- If you click a tool form button that restricts the upload to one upload, then go back to the main uploader it is still restricted (cannot add new uploads).
- If ``upload_from_form_button`` is set to always on in the config, the resulting button doesn't have a select button (may be a pre-existing issue from 20.09 - I'm unsure).
- Also the state being shared between traditional upload and the tool form upload is a kind of broken user experience right? Like if you upload something, then click on the form to do an upload there is already junk in there and it will allow a single new upload but then select both the existing upload and the new upload you click the upload button.
-

Nerdinacan:
> * Also the state being shared between traditional upload and the tool form upload is a kind of broken user experience right? Like if you upload something, then click on the form to do an upload there is already junk in there and it will allow a single new upload but then select both the existing upload and the new upload you click the upload button.

I think we need to pick a paradigm. Is it an uploader you set up and walk away from (as is implied by the button/status indicator combo)? Or is it a thing you open and just let run?  A combination of the two?

If we need (and I think we do) multiple upload instances because of the select callback feature (where you opened the upload for a specific reason and intend to do something specialized with the results) --- then what that tells me is that you need a multi-layered design with core internal upload handling functionality that is exposed in multiple windows based on opener context, because the browser has limited upload resources we'll have to manage between multiple windows.


And that leads me to my next point.

The uploader is so broken it needs to be completely reworked. I'm not joking when I say working around the design of this single component has eaten a month of my time this just this year. It touches too many other things and I think part of the reason is the nebulous requirements you have illustrated with your comment about the weird user-experience.

My eyes bleed every time I look at the entire parent context being passed into each tab (as a property).  

```html
<b-tab title="Regular" id="regular" button-id="tab-title-link-regular" v-if="showRegular">
     <!-- note that the entire VM is passed in as a parameter to each tab -->
    <default :app="this" :lazy-load-max="50" :multiple="multiple" :selectable="selectable" v-on="$listeners" />
</b-tab>
```

Inside that component, we reference Global galaxy and a jQuery plugin that does the job of uploading that, to my knowledge, has no reason to be a jQuery plugin other than somebody tied it to $ at some point in the past. 

I do not want to fiddle with this butchered code any more than is absolutely necessary. 

I believe we should start assembling a complete set of requirements for the replacement. This particular piece of code has been a significant barrier to everything else I try to implement, without exception because it is somehow both spaghetti code and a turducken, which I think is absolutely amazing because I had naively previously believed those 2 programming flaws to be mutually exclusive.

I have already made significant headway into rebuilding a replacement, and would be eager to collect any additional requirements you might have.

Thanks man!

mvdbeek:
> I think we need to pick a paradigm.

We did, in the past these were completely separate instances with a separate lifetime. The tool form upload got destroyed when you click cancel / outside of the area. Seems like you can emulate that by dropping the callback when the modal is closed.

Nerdinacan:
So cancelling the modal stopped all the queued uploads for that instance? Was that the intended behavior? If so, why, have the progress bar / Upload button combo. Should I remove that?

Nerdinacan:
@jmchilton @mvdbeek 

I think you're all set. There's just 2 separate mount functions, one of which will re-use a global instance of the uploader. The only other change I made was to implement the callback as a proper event allowing the calling context to decide how to subscribe the the emitted result, which is really the more flexible way to approach getting data out of a component.

mvdbeek:
Excellent, this is working well. Thanks for following up @Nerdinacan!

Issue #11435, opened 2021-02-22
Demonstrate Implementing Galaxy Files Plugin for S3 Public Data


jmchilton:
@Delphine-L recommends starting by publishing https://registry.opendata.aws/genomeark/  | https://vgp.github.io/genomeark/Gallus_gallus/ to usegalaxy.org.

xref https://github.com/galaxyproject/galaxy/pull/10523

PR #11458, opened 2021-02-24
Allow anonymous access to S3 buckets via file sources plugins.

jmchilton:
The fs-s3fs PyFileSystem plugin (integrated in #10523) does not allow anonymous access to buckets as far as I can tell. s3fs (ironically not actually used by fs-s3fs) seems more mature and better maintained and does allow easy anonymous access to buckets.

I was able to access Genome Ark (for #11435) from my laptop without any configured credentials using the following definition and this PR.

```yaml
- type: s3fs
  label: Genome Ark
  id: genomeark
  doc: Access to Genome Ark open data on AWS.
  bucket: genomeark
  anon: true
```

### DIFF
--- a/lib/galaxy/dependencies/__init__.py
+++ b/lib/galaxy/dependencies/__init__.py
@@ -219,6 +219,14 @@ class ConditionalDependencies:
     def check_fs_webdavfs(self):
         return 'webdav' in self.file_sources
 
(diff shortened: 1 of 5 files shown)


mvdbeek:
Thanks @jmchilton!

PR #11471, opened 2021-02-24
[21.01] Add cmap datatype

jaidevjoshi83:
This PR includes changes in the tabular.py and conf file for the camp datatype.


ic4f:
@jaidevjoshi83  Does this replace #11453?

jaidevjoshi83:
> @jaidevjoshi83 Does this replace #11453?
I closed #11453. The new PR includes all the changes that #11453 has and a few more based on @bgruening suggestions.

ic4f:
> > I closed #11453. The new PR includes all the changes that #11453 has and a few more based on @bgruening suggestions.

Thanks! I'll review it tonight.

jaidevjoshi83:
> > > I closed #11453. The new PR includes all the changes that #11453 has and a few more based on @bgruening suggestions.
> 
> Thanks! I'll review it tonight.

Sounds great thank you..

lib/galaxy/config/sample/datatypes_conf.xml.sample
```diff
@@ -1014,6 +1015,7 @@
     <sniffer type="galaxy.datatypes.tabular:CSV"/>
     <sniffer type="galaxy.datatypes.tabular:TSV"/>
     <sniffer type="galaxy.datatypes.tabular:MatrixMarket"/>
+    <sniffer type="galaxy.datatypes.tabular:CMAP"/>
```
jdavcs: I think the sniffer for CMAP needs to appear prior to CSV and TSV (and, possibly, ConnectivityTable as well), as the format for cmap is more rigid (see docs: https://docs.galaxyproject.org/en/master/dev/data_types.html#step-2-sniffer - step 2). If you agree, please move it to line 1014, right before ConnectivityTable.

lib/galaxy/datatypes/tabular.py
```diff
@@ -1297,3 +1297,82 @@ def set_meta(self, dataset, overwrite=True, skip=None, max_data_lines=5, **kwd):
...
             dataset.metadata.column_types = ['int', 'int', 'float']
+
+
+@build_sniff_from_prefix
+class CMAP(TabularData):
+    MetadataElement(name="cmap_version", default='0.2', desc="version of cmap", readonly=True, visible=True, optional=False, no_value='0.2')
+    MetadataElement(name="label_channels", default=1, desc="the number of label channels", readonly=True, visible=True, optional=False, no_value=1)
+    MetadataElement(name="nickase_recognition_site_1", default=[], desc="comma separated list of label motif recognition sequences for channel 1", readonly=True, visible=True, optional=False, no_value=[])
+    MetadataElement(name="number_of_consensus_nanomaps", default=0, desc="the total number of consensus genome maps in the CMAP file", readonly=True, visible=True, optional=False, no_value=0)
+    MetadataElement(name="nickase_recognition_site_2", default=[], desc="comma separated list of label motif recognition sequences for channel 2 ", readonly=True, visible=True, optional=True, no_value=[])
```
jdavcs:
```suggestion
    MetadataElement(name="nickase_recognition_site_2", default=[], desc="comma separated list of label motif recognition sequences for channel 2", readonly=True, visible=True, optional=True, no_value=[])
```

lib/galaxy/datatypes/tabular.py
```diff
@@ -1297,3 +1297,82 @@ def set_meta(self, dataset, overwrite=True, skip=None, max_data_lines=5, **kwd):
...
+                    elif self.max_optional_metadata_filesize >= 0 and dataset.get_size() > self.max_optional_metadata_filesize:
+                        # If the dataset is larger than optional_metadata, just count comment lines.
+                        # No more comments, and the file is too big to look at the whole thing. Give up.
+                        dataset.metadata.data_lines = None
+                        break
+                    elif i == comment_lines + 1:
+                        number_of_columns = int(len(l.split('\t')))
+                if not (self.max_optional_metadata_filesize >= 0 and dataset.get_size() > self.max_optional_metadata_filesize):
+                    dataset.metadata.data_lines = i + 1 - comment_lines
+            dataset.metadata.comment_lines = int(comment_lines)
```
jdavcs:
```suggestion
            dataset.metadata.comment_lines = comment_lines
```
it's already an int

lib/galaxy/datatypes/tabular.py
```diff
@@ -1297,3 +1297,82 @@ def set_meta(self, dataset, overwrite=True, skip=None, max_data_lines=5, **kwd):
...
+            dataset.metadata.column_names = column_headers
+            dataset.metadata.column_types = cleaned_column_types
+            dataset.metadata.columns = int(number_of_columns)
```
jdavcs:
```suggestion
            dataset.metadata.columns = number_of_columns
```
same: already an int

lib/galaxy/config/sample/datatypes_conf.xml.sample
jaidevjoshi83: I moved it to line 1014.

lib/galaxy/datatypes/tabular.py
```diff
@@ -1297,3 +1297,82 @@ def set_meta(self, dataset, overwrite=True, skip=None, max_data_lines=5, **kwd):
...
+    file_ext = "cmap"
+
+    def sniff_prefix(self, file_prefix):
+        return file_prefix.startswith('# CMAP File Version:')
+
+    def set_meta(self, dataset, overwrite=True, skip=None, max_data_lines=7, **kwd):
+        if dataset.has_data():
+            with open(dataset.file_name) as dataset_fh:
+                comment_lines = 0
+                for i, l in enumerate(dataset_fh):
```
jdavcs: Please rename `l` >> `line`: the character `l` is not the best identifier: it looks very similar to the digit `1`, which makes the code less readable (also, see PEP 8). Also, unless it were a loop counter, using `line` makes the code easier to understand (vs. a single character identifier)

lib/galaxy/datatypes/tabular.py
jaidevjoshi83: Corrected.

lib/galaxy/datatypes/tabular.py
jdavcs: `column_headers`, `cleaned_column_types`, `number_of_columns` need to be initialized to their default values next to `comment_lines` (line 1332): Even if the cmap format implies that all these will be initialized, it is not obvious from the code, which suggests the possibility of a NameError.

lib/galaxy/datatypes/tabular.py
```diff
@@ -1297,3 +1297,82 @@ def set_meta(self, dataset, overwrite=True, skip=None, max_data_lines=5, **kwd):
...
+                        if l.startswith('#h'):
+
+                            column_headers = l.strip('\n').split("\t")[1:]
+                        elif l.startswith('#f'):
+                            cleaned_column_types = []
+                            for i in l.split('\t')[1:]:
+                                if i.strip("\n") == 'Hex':
+                                    cleaned_column_types.append('str')
+                                else:
+                                    cleaned_column_types.append(i)
```
jdavcs: Please rename this `i`: this variable overrides the outer loop variable (another `i`) at each step. Also, since it's not a counter, a descriptive name would be better (e.g. not `j`)

lib/galaxy/datatypes/tabular.py
```diff
@@ -1297,3 +1297,82 @@ def set_meta(self, dataset, overwrite=True, skip=None, max_data_lines=5, **kwd):
...
+                                if len(fields2) == 2:
+                                    dataset.metadata.channel_2_color = fields2[1]
+                                dataset.metadata.nickase_recognition_site_2 = fields2[0].split(',')
+                    elif self.max_optional_metadata_filesize >= 0 and dataset.get_size() > self.max_optional_metadata_filesize:
+                        # If the dataset is larger than optional_metadata, just count comment lines.
+                        # No more comments, and the file is too big to look at the whole thing. Give up.
+                        dataset.metadata.data_lines = None
+                        break
+                    elif i == comment_lines + 1:
+                        number_of_columns = int(len(l.split('\t')))
```
jdavcs:
```suggestion
                        number_of_columns = len(l.split('\t'))
```
no need: len() returns an int

ic4f:
@jaidevjoshi83 Looks great, thank you for putting so much thought into this! I'll check back tomorrow and will merge this once the remaining items have been addressed.

lib/galaxy/datatypes/tabular.py
```diff
@@ -1297,3 +1297,82 @@ def set_meta(self, dataset, overwrite=True, skip=None, max_data_lines=5, **kwd):
...
+    182 58474736.7  10235   1   1   58820.9 35.4    13.5    13.5    -1.00   -1.00   -1.00   3.63    0.00    0.00    -1.00   0
+    """
+    file_ext = "cmap"
+
+    def sniff_prefix(self, file_prefix):
+        return file_prefix.startswith('# CMAP File Version:')
+
+    def set_meta(self, dataset, overwrite=True, skip=None, max_data_lines=7, **kwd):
+        if dataset.has_data():
+            with open(dataset.file_name) as dataset_fh:
```
jdavcs, line 1331: It appears that this comment got overwritten somehow (so if this appears as a duplicate - just ignore). You have 7 instances of `strip('\n')`. This character appears only at the end of the line; you don't need to retain it, so you can just strip it as the first statement in the body of your main loop (`line = line.strip('\n')`) - and you're done! That'll make the code easier to read. More importantly, that will prevent from unintentionally *not* stripping it, like in line 1345.

lib/galaxy/datatypes/tabular.py
jaidevjoshi83: Whether should I initialize these with an empty list for the first two and a zero for "number_of_columns" or with the example provided in the spec sheet that I also included in the example file. As per the spec sheet, "Additional columns may be present but are not defined" which means this information could vary and didn't say anything about defaults.

jdavcs: You can initialize the list values to `None`, and the int (number_of_columns) to 0 - i.e., the natural values if the file were empty (so the spec is irrelevant).

jaidevjoshi83: Thank you.

lib/galaxy/datatypes/tabular.py
jaidevjoshi83: renamed `l >> line`

lib/galaxy/datatypes/tabular.py
jaidevjoshi83: Renamed as `i >> column_type`

lib/galaxy/datatypes/tabular.py
jaidevjoshi83: `int` removed from all the lines you mentioned.

lib/galaxy/datatypes/tabular.py
jaidevjoshi83: I removed the repeated `strip('\n')` and also incorporated/changed other parts as per your suggestions.  Thanks for the review and suggestions.  However, the test `Run yarn Jest` still failing and I'm figuring it out.

jdavcs: I've restarted that test - let's see if it passes. Might have been a flaky test.

jdavcs (approved):

ic4f:
@jaidevjoshi83 this is great! Thank you for taking care of all the proposed edits! Congratulations on your first commit to Galaxy's core repo!

PR #11478, opened 2021-02-25
Copy to Clipboard button next to the tool id in job information

### LINKED ISSUE #11473: Add a Copy to Clipboard button next to the tool id
Would be very convenient to add an icon next to the tool id in the dataset info page to simply copy the entire tool id. Also as part of #11472.

OlegZharkov:
fixes https://github.com/galaxyproject/galaxy/issues/11473

Clipboard button next to the tool id:

image

Additionaly it brings generic `copy-to-clipboard` component e.g.:

` <copy-to-clipboard message="Tool ID was copied to your clipboard" :text="job.tool_id" />`

### DIFF
--- a/client/src/components/JobInformation/JobInformation.vue
+++ b/client/src/components/JobInformation/JobInformation.vue
@@ -5,7 +5,14 @@
             <tbody>
                 <tr v-if="job && job.tool_id">
                     <td>Galaxy Tool ID:</td>
-                    <td id="galaxy-tool-id">{{ job.tool_id }}</td>
+                    <td id="galaxy-tool-id">
+                        {{ job.tool_id }}
+                        <font-awesome-icon
+                            title="Copy Tool ID"
+                            icon="copy"
+                            @click="copyToolId(job.tool_id)"
+                        ></font-awesome-icon>
+                    </td>
                 </tr>
                 <tr v-if="job && job.tool_version">
                     <td>Galaxy Tool Version:</td>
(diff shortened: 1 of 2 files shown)


mvdbeek (approved):
Maybe that's too much decomposing, but maybe the copy button can be its own vue component ? Otherwise this is great!

OlegZharkov:
> Maybe that's too much decomposing, but maybe the copy button can be its own vue component ? Otherwise this is great!

I mean, we could if it's useful, do we have another usecase for this?

mvdbeek:
You could make this generic, so that the text to copy is being passed in. Then we have the copy button (and the styling) in a single place. It also makes the parent component slimmer, so you don't have for instance the `copyToolId` method in the body of the parent component.

OlegZharkov:
> You could make this generic, so that the text to copy is being passed in. Then we have the copy button (and the styling) in a single place. It also makes the parent component slimmer, so you don't have for instance the `copyToolId` method in the body of the parent component. It's a style question, I don't want to hold up the PR.

no worries, I will do it now

OlegZharkov:
> You could make this generic, so that the text to copy is being passed in. Then we have the copy button (and the styling) in a single place. It also makes the parent component slimmer, so you don't have for instance the `copyToolId` method in the body of the parent component. It's a style question, I don't want to hold up the PR.

So... I changed it to `<copy-to-clipboard :text="{variabel to be copied}" message="{text on toast}"  />`

mvdbeek:
Thank you, that's great!

OlegZharkov:
should I change icon to this one? :thinking: 
image

dannon:
@OlegZharkov Good call -- yes, we should use the same icon consistently

client/src/components/JobInformation/JobInformation.vue
```diff
@@ -5,7 +5,14 @@
             <tbody>
                 <tr v-if="job && job.tool_id">
                     <td>Galaxy Tool ID:</td>
-                    <td id="galaxy-tool-id">{{ job.tool_id }}</td>
+                    <td id="galaxy-tool-id">
+                        {{ job.tool_id }}
+                        <copy-to-clipboard
+                            message="Tool ID was copied to your clipboard"
+                            :text="job.tool_id"
+                            title="copy Tool ID"
```
mvdbeek:
```suggestion
                            title="Copy Tool ID"
```
looks nicer, I think

OlegZharkov:
thanks @mvdbeek @dannon @afgane

PR #11483, opened 2021-02-25
Guess tool name when extracting workflows with imported datasets

### LINKED ISSUE #11449: When extracting a workflow from history, the tool for imported datasets appears as Unknown
When importing a dataset from a `Data Library` (instead of uploading your own file) and then trying to `Extract Workflow` from history, the <del>dataset</del> tool will appear as `Unknown`. Maybe something like `Import File` can be more consistent?

Example using both methods:

image

This may be just a papercut.

davelopez:
This should fix #11449
Although maybe not the most elegant solution :)

image

### DIFF
--- a/lib/galaxy/workflow/extract.py
+++ b/lib/galaxy/workflow/extract.py
@@ -167,15 +167,11 @@ class FakeJob:
 
     def _guess_name_from_dataset(self, dataset) -> Optional[str]:
         """Tries to guess the name of the fake job from the dataset associations."""
-        name = None
-        try:
-            if dataset.copied_from_history_dataset_association:
-                name = "Import from History"
-            elif dataset.copied_from_library_dataset_dataset_association:
-                name = "Import from Library"
-            return name
-        except Exception:
-            return None
+        if dataset.copied_from_history_dataset_association:
+            return "Import from History"
+        if dataset.copied_from_library_dataset_dataset_association:
+            return "Import from Library"
+        return None
 
 
 class DatasetCollectionCreationJob:
--- a/lib/galaxy/workflow/extract.py
+++ b/lib/galaxy/workflow/extract.py
@@ -2,6 +2,7 @@
 histories.
 """
 import logging
+from typing import Optional
 
 from galaxy import exceptions, model
 from galaxy.tool_util.parser import ToolOutputCollectionPart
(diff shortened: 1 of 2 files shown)


davelopez:
Oops didn't check the tests. I will fix it.

lib/galaxy/workflow/extract.py
```diff
@@ -161,7 +162,20 @@ class FakeJob:
...
+    def _guess_name_from_dataset(self, dataset) -> Optional[str]:
+        """Tries to guess the name of the fake job from the dataset associations."""
+        name = None
+        try:
+            if dataset.copied_from_history_dataset_association:
+                name = "Import from History"
+            elif dataset.copied_from_library_dataset_dataset_association:
+                name = "Import from Library"
+            return name
+        except Exception:
```
jdavcs: Is there a specific type of error that's expected here? If so, should we log it? If not, and since there's no error handling involved, do we need to wrap this in a try clause? `None` is returned implicitly, so it could be as simple as if [case 1] return x; elif [case 2] return y.

davelopez: Thank you @ic4f for checking.
> Is there a specific type of error that's expected here?

Probably an `AttributeError` but it is only expected in the unit tests.

> do we need to wrap this in a try clause?

This was an attempt to say, whatever happens here... just return the default value None. I definitely don't like the general catch.

In any case, it was a 'not very elegant' attempt to pass [this unit test](https://github.com/galaxyproject/galaxy/blob/6e59f9b919bd88efbbf894cb603c13ec5118171e/test/unit/workflows/test_extract_summary.py#L48) that was failing because the attribute `copied_from_library_dataset_dataset_association` was not defined in [MockHda here](https://github.com/galaxyproject/galaxy/blob/6e59f9b919bd88efbbf894cb603c13ec5118171e/test/unit/workflows/test_extract_summary.py#L129).
My first thought was to add the attribute to the MockHda (like `copied_from_history_dataset_collection_association` which is already there), but then I thought that this will require maintaining the mock with every other attribute that may be used to guess the name so I went for this ugly solution... 
Do you think would be better just to add the attribute to the mock and remove the try? Any suggestions are welcome :)

jmchilton: I would add the attributes to the MockHda instead of this catch I think.

jdavcs: Oh, I see. So, in my opinion, your first approach was spot on - defining it in the mock next to `copied_from_history_dataset_collection_association` makes perfect sense to me. (maintaining mocks shouldn't be an issue). 

You could also set that attribute on the mock *in the test itself*. In fact, that way you can add a separate unit test that would directly test the new functionality you've added! (you could create 3 HDAs in the same test method, mocking each one differently (you might need another `MockHDA` as the parent (`hda.copied_from_history_dataset_association = MockHda(job=UNDEFINED_JOB)`). And then test the 3 guessed names.
EDIT: I just saw John's comment above - github doesn't refresh these in real time. So yes, +1)

davelopez: Thanks for the advice! I have removed the catch and added a unit test for the `FakeJob` name guessing. I am happier now 😄

jdavcs: Thank you - looks great! And very good to have a unit test covering this.

jdavcs (approved):

ic4f:
Unit test failures unrelated (2 flaky tests)

PR #11485, opened 2021-02-25
Libraries rewrite

### LINKED ISSUE #11465: Data libraries search on enter triggers a refresh of the page
## Steps
- Go to `https://{your-galaxy}/library/list`.
- Type something in the search and hit enter.
- The page gets reloaded and the results are lost.

bug_search02

OlegZharkov:
Another round of Backbone removal. This PR introduces rewritten Data Libraries table in Vue.  Another entry in vue router, so we don't need to reload when we enter a folder, that makes libraries surfing will become (hopefully) smoother. We're almost out of backbone in libraries. Only a few little backbone-views are left.

It also fixes https://github.com/galaxyproject/galaxy/issues/11465 and few minor legacy bugs.
image


jmchilton:
Thanks so much! This all looks awesome to me.

OlegZharkov:
> Thanks so much! This all looks awesome to me.

thanks a lot @jmchilton

Issue #11004, opened 2020-12-28
UI: Available tool versions should be sorted from latest to oldest


nsoranzo:
In the tool form, the Versions button allow to select a different version of the tool to be used. The dropdown list of available tool versions is presently sorted from the oldest (on top) to the latest (at the bottom), I think it should be the inverse, so that newer versions are closer to the button.

cat-bro:
having a look at this for papercuts

mvdbeek:
Done in https://github.com/galaxyproject/galaxy/pull/11490

PR #11490, opened 2021-02-25
[21.01] Sort tool versions from newest to oldest in tool form

cat-bro:
For issue https://github.com/galaxyproject/galaxy/issues/11004

List alternative tool versions from newest to oldest.


mvdbeek:
Thank you @cat-bro!

PR #11492, opened 2021-02-26
Rebased anvil diff

almahmoud:
The only hard to figure out conflict was:

```
++<<<<<<< HEAD
 +    for el in root.xpath('//macro'):
 +        if el.get('type') != 'template':
 +            # Only keep template macros
 +            el.getparent().remove(el)
++=======
+     for el in root.xpath('//macros'):
+         el.getparent().remove(el)
++>>>>>>> 50746ebee2... Call (recursive) expand_token_for_el just once
```

here: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/util/xml_macros.py#L28

Otherwise this should hopefully make the divergence from Galaxy repo more transparent and easier to maintain.


dannon:
This PR is in the wrong repo, right?  It should be updating anvilproject:dev, not here.

almahmoud:
yes, oops i screwed up...

PR #11496, opened 2021-02-26
Mark some tool reload tests as expected to fail

davelopez:
The unit tests `ToolBoxTestCase::test_tool_reload_for_broken_tool` and `ToolBoxTestCase::test_tool_reload_when_macro_is_altered` were failing or passing transiently.
Marked as [expected to fail](https://docs.pytest.org/en/latest/skipping.html#xfail-mark-test-functions-as-expected-to-fail) as @mvdbeek suggested.

### DIFF
--- a/test/unit/tools/test_toolbox.py
+++ b/test/unit/tools/test_toolbox.py
@@ -190,7 +190,7 @@ class ToolBoxTestCase(BaseToolBoxTestCase):
         assert tool is not None
         assert len(tool._macro_paths) == 1
 
-    @pytest.mark.xfail
+    @pytest.mark.xfail(raises=AssertionError)
     def test_tool_reload_when_macro_is_altered(self):
         self._init_tool(filename="tool_with_macro.xml",
                         tool_contents=SIMPLE_TOOL_WITH_MACRO,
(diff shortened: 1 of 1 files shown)


test/unit/tools/test_toolbox.py
```diff
@@ -189,6 +190,7 @@ def test_record_macros(self):
         assert tool is not None
         assert len(tool._macro_paths) == 1
 
+    @pytest.mark.xfail
```
nsoranzo:
```suggestion
    @pytest.mark.xfail(raises=AssertionError)
```

test/unit/tools/test_toolbox.py
```diff
@@ -209,6 +211,7 @@ def check_tool_macro():
 
         self._try_until_no_errors(check_tool_macro)
 
+    @pytest.mark.xfail
```
nsoranzo:
```suggestion
    @pytest.mark.xfail(raises=AssertionError)
```

nsoranzo (approved):
We spent so much time with @mvdbeek trying to make these 2 tests work consistently, I guess it's a lost cause!

davelopez:
On the bright side, they sometimes pass! :)

PR #11497, opened 2021-02-26
xsd: use validator as child of param

bernt-matthias:
instead of child of options .. as documented


nsoranzo (approved):
I agree a validator makes more sense as child of `<param>`, so this improves the examples. The previous version is still valid though, a validator found as child of `<options>` is added to the parent param validator list (we have dozens of these in tools-iuc).

bernt-matthias:
Thanks for the feedback.

> I agree a validator makes more sense as child of <param>, so this improves the examples.  The previous version is still valid though, a validator found as child of <options> is added to the parent param validator list

Just wanted to increase consistency a bit.

> we have dozens of these in tools-iuc

Probably copy paste from those examples.

PR #11498, opened 2021-02-26
[WIP][20.09] Record explicit dataset collection instances as invocation outputs

mvdbeek:
This is needed for consuming outputs from tools that generate dataset collections that are not mapped over ... under certain conditions, that is why no tests are included here, but this is needed for https://github.com/galaxyproject/iwc/pull/23 to schedule successfully.


mvdbeek:
I think I was on the wrong track here and #11513 is the more correct fix.

PR #11499, opened 2021-02-26
[WIP][20.09] Record explicit HDCAs as invocation outputs

mvdbeek:
and set invocation outputs only when all jobs have been scheduled.

This is like https://github.com/galaxyproject/galaxy/pull/11498, but ensures that we also correctly set invocation workflow outputs.


mvdbeek:
And now we've lost the non-data parameters from the invocation outputs ...

mvdbeek:
Same here, #11513 should be the right fix.

PR #11507, opened 2021-02-27
add sshfs plugin to the new file infrastructure

bgruening:
This can be configured with something like in the files_sources_conf.yml

```
- type: ssh
  id: ssh-to-host-x
  doc: "SSH share on X"
  host: gruenings.eu
  user: bag
  passwd: None
  pkey: /home/user/.ssh/id_rsa
  timeout: 10
  port: 22
  keepalive: 10
  compress: False
  config_path: "~/.ssh/config"

- type: ssh
  id: local-ssh
  doc: "local ssh"
  host: localhost
  user: bag
```

@jmchilton should we add a files_sources_conf.yml.sample file with example configs?

fs.sshfs is needed to make it work.


jmchilton:
It would be great to have an example config file!

PR #11509, opened 2021-02-28
Add ready-to-go gitpod config

mvdbeek:
Once the workspace is pre-built it takes about 30-60 seconds to start the [Gitpod](https://www.gitpod.io/) environment, which gives you the option to start Galaxy within vscode in the browser, or run the tests, or do whatever you want in the terminal. We're also automatically starting `make client-watch`, so you can start changing the client and see the results in the browser.

If you start Galaxy in the debug configuration you'll be asked if you want to open the port, and then you can interact with a live Galaxy instance.

This is great for working on multiple branches, you don't need to switch branches, install dependencies and build the client ... just switch to a new tab in your browser.

Here's the Galaxy interface served in gitpod
Screenshot 2021-02-28 at 21 43 18
And here are the tests:
Screenshot 2021-02-28 at 21 45 39

To get this to work I had to fix the cookie parsing in https://github.com/galaxyproject/galaxy/commit/61cefb35525d399971b216d778f71740119c19fa

We could also setup cvmfs, and this might also be really cool for IUC and IWC.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,34 @@
+github:
+  prebuilds:
+    # enable for the master/default branch (defaults to true)
+    master: true
+    # enable for all branches in this repo (defaults to false)
+    branches: true
+    # enable for pull requests coming from this repo (defaults to true)
+    pullRequests: true
+    # enable for pull requests coming from forks (defaults to false)
+    pullRequestsFromForks: true
+    # add a "Review in Gitpod" button as a comment to pull requests (defaults to true)
+    addComment: false
+    # add a "Review in Gitpod" button to pull requests (defaults to false)
+    addBadge: false
+    # add a label once the prebuild is ready to pull requests (defaults to false)
+    addLabel: false
+image: gitpod/workspace-postgres
+tasks:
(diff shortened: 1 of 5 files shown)


dannon:
I can take a more in-depth look later, but I have to remind myself why we swapped to the more strict (but have to hope correct) standard lib cookie handling -- this has come up before.  Pretty cool though!

mvdbeek:
Thanks @dannon!
> We could also setup cvmfs, and this might also be really cool for IUC and IWC.

While that would be super cool the fuse module is not available (even of you opt in to the beta, which gives you access to docker and superuser). Maybe something to check with their support, maybe this could be enabled.

mvdbeek:
fwiw the parse_cookies function is here: https://github.com/Pylons/webob/blob/master/src/webob/cookies.py#L201 and it does some validation using the regexes here: https://github.com/Pylons/webob/blob/master/src/webob/cookies.py#L336-L344

dannon:
@mvdbeek Yeah, the cookie handling looks good there.  Test failure is unrelated.

PR #11511, opened 2021-03-01
[21.01] Fix DefaultQuotaAssociation access to Enum values

davelopez:
Follow up to #11277

Fixes some quota operation from the Admin panel:
- Set an existing quota as default
- Create a new quota as default with the `Yes` option and one of the supported types.


jmchilton:
Thanks!

Issue #11502, opened 2021-02-26
Error loading/executing workflow


kxk302:
@innovate-invent reported this error when executing his workflow:

"err_msg": {"cut_type_options|list": "parameter 'list': invalid options ('4,5,1') were selected (valid options: 1,4,5)"}, "err_code": 0, "err_data": {"5": {"cut_type_options|list": "parameter 'list': invalid options ('4,5,1') were selected (valid options: 1,4,5)"}}}

I re-produced the problem with the attached workflow and input files. Please remove the .txt extension from the workflow/input files (they were added as Github does not accept those file types as attachment).

The problem is that in ColumnListParameter's get_legal_values(), the column values is a string of comma separated integers (instead of being a list). Added code to handle this situation.

[example.gff3.txt](https://github.com/galaxyproject/galaxy/files/6050220/example.gff3.txt)

[Cluster_islands.ga.txt](https://github.com/galaxyproject/galaxy/files/6050218/Cluster_islands.ga.txt)

[example.fasta.txt](https://github.com/galaxyproject/galaxy/files/6050223/example.fasta.txt)

P.S. It seems a previous attempt was made to fix this issue: https://github.com/galaxyproject/galaxy/pull/11341

innovate-invent:
Great! Can you push your fix so I can test it out?

kxk302:
@innovate-invent I am creating a Pull Request next, so it can be reviewed, before its pushed.

PR #11514, opened 2021-03-01
Handle the case where column list is a comma-separated string

kxk302:
This is a fix for https://github.com/galaxyproject/galaxy/issues/11502

When the list of columns is a comma-separated string, it need to be broken down into a list. Also, we must check whether the set of columns is a subset of legal values.

### DIFF
--- a/lib/galaxy/tools/parameters/basic.py
+++ b/lib/galaxy/tools/parameters/basic.py
@@ -1362,9 +1362,9 @@ class ColumnListParameter(SelectToolParameter):
 
         value = other_values.get(self.name)
         if value is not None:
-            # There are cases where 'value' is a string of comma separated values
-            # This ensures that it is converted into a list
-            value = value if isinstance(value, list) else value.split(",")
+            # There are cases where 'value' is a string of comma separated values. This ensures 
+            # that it is converted into a list, with extra whitespace around items removed.
+            value = util.listify(value, do_strip=True)
             if not set(value).issubset(set(legal_values)) and self.is_file_empty(trans, other_values):
                 legal_values.extend(value)
 
--- a/lib/galaxy/tools/parameters/basic.py
+++ b/lib/galaxy/tools/parameters/basic.py
@@ -1362,7 +1362,7 @@ class ColumnListParameter(SelectToolParameter):
 
         value = other_values.get(self.name)
         if value is not None:
-            # There are cases where 'value' is a string of comma separated values. This ensures 
+            # There are cases where 'value' is a string of comma separated values. This ensures
             # that it is converted into a list, with extra whitespace around items removed.
             value = util.listify(value, do_strip=True)
             if not set(value).issubset(set(legal_values)) and self.is_file_empty(trans, other_values):
--- a/lib/galaxy/tools/parameters/basic.py
+++ b/lib/galaxy/tools/parameters/basic.py
@@ -1361,9 +1361,12 @@ class ColumnListParameter(SelectToolParameter):
         legal_values = self.get_column_list(trans, other_values)
 
         value = other_values.get(self.name)
-        if value is not None and value not in legal_values and self.is_file_empty(trans, other_values):
-            value = value if isinstance(value, list) else [value]
-            legal_values.extend(value)
+        if value is not None:
+            # There are cases where 'value' is a string of comma separated values
+            # This ensures that it is converted into a list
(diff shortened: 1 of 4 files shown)


lib/galaxy/tools/parameters/basic.py
```diff
@@ -1361,9 +1361,12 @@ def get_legal_values(self, trans, other_values):
         legal_values = self.get_column_list(trans, other_values)
 
         value = other_values.get(self.name)
-        if value is not None and value not in legal_values and self.is_file_empty(trans, other_values):
-            value = value if isinstance(value, list) else [value]
-            legal_values.extend(value)
+        if value is not None:
+            # There are cases where 'value' is a string of comma separated values
+            # This ensures that it is converted into a list
+            value = value if isinstance(value, list) else value.split(",")
```
jdavcs: Or you could call `galaxy.util.listify()` https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/util/__init__.py#L967 - it does just that, but also handles tuples, so is a bit more generic.

kxk302: Good idea. Changed accordingly.

jdavcs (approved):

lib/galaxy/tools/parameters/basic.py
jdavcs: Looks good to me, thank you!

lib/galaxy/tools/parameters/basic.py
```diff
@@ -1361,9 +1361,12 @@ def get_legal_values(self, trans, other_values):
         legal_values = self.get_column_list(trans, other_values)
 
         value = other_values.get(self.name)
-        if value is not None and value not in legal_values and self.is_file_empty(trans, other_values):
-            value = value if isinstance(value, list) else [value]
-            legal_values.extend(value)
+        if value is not None:
+            # There are cases where 'value' is a string of comma separated values. This ensures
+            # that it is converted into a list, with extra whitespace around items removed.
+            value = util.listify(value, do_strip=True)
```
innovate-invent, line 1367: is this a failure of the calling function to listify the values?
I feel like the tool input should be listified much earlier on

innovate-invent: I think the issue is actually here: 

https://github.com/galaxyproject/galaxy/blob/84647cbd0a05b8dd87a5ffe4662c033ad050b0ee/lib/galaxy/tools/parameters/basic.py#L1253-L1262

jdavcs: I'd rather not make that assumption about input; processing this collection of items as a list is just the method's implementation detail, which may or may not change in the future. Besides, accepting a list or a tuple, or a comma-separated string seems more flexible.

innovate-invent: The issue seems to be that the logic to convert the column spec to a list is very different than what the listify function does.
The column list spec seems to allow multiline and numbers prefixed with a 'c'.

mvdbeek:
@kxk302 Could you strip down the failing workflows as much as possible and recapitulate the error with one of the framework test tools ?
You can then add a workflow API test.

kxk302:
> @kxk302 Could you strip down the failing workflows as much as possible and recapitulate the error with one of the framework test tools ?
> You can then add a workflow API test.

@mvdbeek , I am writing 2 api tests to cover the empty input file and comma separated values for columns. However, I am having some issues. Would you or @jmchilton have 15 minutes to meet and discuss?

kxk302:
@mvdbeek Added another test, that fails without my fix for comma separated column string, and passes with the fix.

ic4f:
> @mvdbeek Added another test, that fails without my fix for comma separated column string, and passes with the fix.
> BTW, thanks for your help today!

Based on (a) these^^ results of running the new test, and (b) the characterization of a specific problem in this PR's description, I think it is safe to assume that this PR solves *the problem at hand*. What @innovate-invent has [described](https://github.com/galaxyproject/galaxy/pull/11514#discussion_r585154822) seems to be a related, but *different* problem, which, I believe, deserves its own issue and PR. The test demonstrates that there is indeed a problem that is fixed by the proposed solution.

jmchilton:
We should backport this to at least 21.01 I think. Seems like an important fix.

innovate-invent:
You might also want to add a test that includes newlines and 'c' before the column number
```
c1
c2,c3
```

PR #11520, opened 2021-03-02
[20.09] OIDC Logout+Redirect fix.

dannon:
All galaxy-app / transaction accessed cookies should start with 'galaxy'.  We intentionally toss anything else.


jmchilton (approved):

dannon:
I tested adding extra debugging to give hints to future-us for these types of things in https://github.com/galaxyproject/galaxy/commit/96d99a30c80bc7d45b73c33381e54d74ed9c92b7, but it's (as you'd expect for something that potentially logs on every request) super chatty.   I don't think we likely want it.

PR #11526, opened 2021-03-02
Fix clearing of search inputs

### LINKED ISSUE #11504: Clear Search button in History does not clear search text in some cases
I am not sure if this is intended behavior.
## Steps
- Write something in the History `search datasets` box (but do not press enter)
- Use the `clear search` button.
- Nothing happens. Here I will expect the input search text to be cleared.

image

Similar behavior happens in the `search histories` and `search all datasets` from the Histories panel. In these cases, after pressing enter and then use the clear search button, the search results are successfully cleared but the search text remains.

search

Also notice a little glitch when searching histories where the next history from current slightly moves behind the current history.

dannon:
The slightly more robust selector usage resolves this and will hopefully be more durable to changes moving
forward.

Fixes #11504

### DIFF
--- a/client/src/ui/search-input.js
+++ b/client/src/ui/search-input.js
@@ -37,7 +37,7 @@ function searchInput(parentNode, options) {
     // .................................................................... input rendering and events
     // visually clear the search, trigger an event, and call the callback
     function clearSearchInput(event) {
-        var $input = $(this).parent().children("input");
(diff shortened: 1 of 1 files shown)


jmchilton:
Probably should be backported to 21.01? @dannon any objection if I open a PR for that?

dannon:
@jmchilton None at all, go for it and I'll merge it in.  Thanks!

PR #11527, opened 2021-03-02
[21.01] Fix bug to prevent workflow refactor indices from being wrong.

jmchilton:
Don't let the workflow linting page build refactor actions with step indices that will be broken after the the prompt to save.

### DIFF
--- a/client/src/components/Workflow/Editor/Index.vue
+++ b/client/src/components/Workflow/Editor/Index.vue
@@ -342,7 +342,7 @@ export default {
         },
         async onRefactor(response) {
             await fromSimple(this, response.workflow);
-            this._loadEditorData(response);
+            this._loadEditorData(response.workflow);
         },
         onAdd(node) {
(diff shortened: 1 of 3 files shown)


mvdbeek (approved):

client/src/components/Workflow/Editor/Index.vue
```diff
@@ -346,6 +347,7 @@ export default {
         },
         onAdd(node) {
             this.nodes[node.id] = node;
+            this.requiresReindex = true;
```
mvdbeek, line 350: This is too aggressive and forces a reload even when you just loaded the workflow

jmchilton: Ahhh - I did not test that.

This would have been a great feature to have UI tests for 😅.

mvdbeek: Certainly looked fine on first glance 😆

